PR #107, opened 2021-03-02
Update vendored typescript compiler to v4.1.5 [FG-126]

amacneil:
We're currently using two versions of typescript. In this PR I updated the vendored version (in `app/package.json` used for compiling code in node playground) to match the version we use to compile studio itself.

We can't upgrade to typescript v4.2.x yet because of an apparent yarn bug. However, this brings us one step closer.


app/players/UserNodePlayer/nodeTransformerWorker/utils.ts
```diff
@@ -50,21 +50,8 @@ export const transformDiagnosticToMarkerData = (diagnostic: ts.Diagnostic): Diag
     character: endColumn,
   } = diagnostic.file.getLineAndCharacterOfPosition(diagnostic.start + diagnostic.length);
 
-  let messageText = "";
-  // Typescript sometimes builds a linked list of formatted diagnostic messages
-  // to be used as part of a multiline message.
```
amacneil, line 55: Instead of linked list, this is now passed as an array of linked lists for some reason.

The new API is not well documented but monaco-typescript was updated to consume it [here](https://github.com/microsoft/monaco-typescript/pull/36/files#diff-ea95b2efb3fd38940772369b9ec8a70e76d7a0f40d570f1056a4a054eccfb951).

PR #114, opened 2021-03-02
Add noUncheckedIndexAccess build flag to tsconfig.json

defunctzombie:
The flag is enabled at the top level tsconfig so vscode will highlight
the issues as you edit files. The build is configured to ignore known failing
files from UncheckedIndexAccess.json until they are fixed.


webpack.renderer.config.ts
```diff
@@ -194,7 +195,41 @@ export default (_: never, argv: WebpackArgv): Configuration => {
...
+            },
+          },
+        },
+        issue: {
+          exclude: (issue) => {
+            if (issue.file === undefined) {
+              return false;
+            }
+
+            const repoPath = issue.file.replace(`${__dirname}${path.sep}`, "");
```
jtbandes, line 228: Can this be something like `path.relative(__dirname, issue.file)`

PR #117, opened 2021-03-03
Fix git clone and yarn install on Windows

jhurliman:
- Update to latest rosbag.js to fix Windows yarn install error
- Update rosbag.js, set LF line endings with .gitattributes and vscode settings.json

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,10 @@
+* text=auto eol=lf
+
+*.afdesign filter=lfs diff=lfs merge=lfs -text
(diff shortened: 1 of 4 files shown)


.gitattributes
```diff
@@ -1,4 +1,10 @@
+* text=auto eol=lf
```
amacneil, line 1: remind me what this does again?

defunctzombie: Doesn't this make all files lfs? @jhurliman ?

defunctzombie: Sounds like this will ensure all line endings are consistent across platforms?

PR #123, opened 2021-03-03
Make Storybook work

jtbandes:
All stories now build successfully! Some still look broken at runtime though — we have more cleanup work to do.

- `yarn storybook` to open the storybook.
- Use the beta version of storybook that's compatible with Webpack 5 (https://github.com/storybookjs/storybook/pull/13808)
  - Most of our main renderer webpack config can be shared with storybook.
- Removed uses of `@dump247/storybook-state` and replace them with `useState` hooks.
- Run `build:storybook` in CI so we don't break it.

### DIFF
--- a/webpack.renderer.config.ts
+++ b/webpack.renderer.config.ts
@@ -187,6 +187,41 @@ export function makeConfig(_: unknown, argv: WebpackArgv): Configuration {
         resourceRegExp: /^\.[\\/]locale$/,
         contextRegExp: /moment$/,
       }),
+      // We use two ForkTsCheckers to ignore known files which fail noUncheckedIndexedAccess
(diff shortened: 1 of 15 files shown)


webpack.renderer.config.ts
```diff
@@ -236,4 +228,21 @@ export default (_: never, argv: WebpackArgv): Configuration => {
       __filename: true,
     },
   };
+}
+
+export default (env: unknown, argv: WebpackArgv): Configuration => {
```
defunctzombie, line 233: If this is the only plugin we don't want it might be cleaner to iterate the plugins array in storybook and slice this out - then we don't need any extra make stuff in here.

PR #128, opened 2021-03-03
Add Windows CI builder, build .appx

jhurliman:


.github/workflows/package.yml
```diff
@@ -36,7 +36,16 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-yarn-
 
+      # Work around yarn checksum mismatches for git repo dependencies on Windows
+      - name: YARN_CHECKSUM_BEHAVIOR=ignore yarn install --immutable
+        run: yarn install --immutable
+        if: matrix.os == 'windows-latest'
```
amacneil, line 42: fyi rather than conditionals, I think there is a slightly cleaner way to do this by setting a variable specific to that matrix:

https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#using-environment-variables-in-a-matrix

caveat I have not tried this, I just saw it in the docs.

PR #129, opened 2021-03-03
Add DropdownItem component

defunctzombie:
The DropdownItem component provides a way to set "values" for individual
dropdown items.

### DIFF
--- a/app/components/Dropdown/DropdownItem.tsx
+++ b/app/components/Dropdown/DropdownItem.tsx
@@ -1,0 +1,21 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { PropsWithChildren } from "react";
+
+type DropdownItemProps = {
+  value: unknown;
+};
+
+// DropdownItem creates an entry within a Dropdown. Use the _value_ prop to indicate
+// What value the dropdown will have when the item is selected.
+//
+// A childless DropdownItem will render as a <span>. Specify children to customize the appearence.
+export default function DropdownItem(props: PropsWithChildren<DropdownItemProps>) {
+  if (props.children === undefined) {
+    return <span>{String(props.value)}</span>;
+  }
+
+  return <>{props.children}</>;
+}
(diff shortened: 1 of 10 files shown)


app/components/PlaybackControls/PlaybackTimeDisplayMethod.tsx
```diff
@@ -181,14 +182,12 @@ const PlaybackTimeDisplayMethod = ({
...
-        {/* @ts-expect-error change <span> to DropdownItem since value is not a property of <span> */}
-        <span key="day" value="TOD">
-          Time of day (TOD)
-        </span>
-        {/* @ts-expect-error change <span> to DropdownItem since value is not a property of <span> */}
-        <span key="ros" value="ROS">
-          ROS time
-        </span>
+        <DropdownItem value="TOD">
+          <span key="day">Time of day (TOD)</span>
```
jtbandes, line 186: These should be able to be deleted. `key` is for React tree diffing and nothing else (I don't think it shows up in `props` at all).

app/components/Dropdown/DropdownItem.tsx
```diff
@@ -0,0 +1,21 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { PropsWithChildren } from "react";
+
+type DropdownItemProps = {
+  value: unknown;
```
jtbandes, line 8: I think this should always be... a string or a number, right?

app/panels/Plot/PlotLegend.tsx
```diff
@@ -130,14 +131,18 @@ export default function PlotLegend(props: PlotLegendProps) {
...
+            </DropdownItem>
+            <DropdownItem value="index">
+              <span>index</span>
+            </DropdownItem>
+            <DropdownItem value="currentCustom">
+              <span>msg path (current)</span>
+            </DropdownItem>
+            <DropdownItem value="custom">
+              <span>msg path (accumulated)</span>
+            </DropdownItem>
```
jtbandes, line 145: Can these become childless or maybe change to a `label="..."` prop in addition to `value`?

jtbandes (commented):
Seems like we could clean up further by just making `<span>` no longer a part of the API... just `<DropdownItem value="foo" label="Foo"/>`

PR #139, opened 2021-03-04
Fix ChildToggle state warning

jtbandes:
Ok, this grew a bit, but bear with me.

1. ChildToggle would setState inside its render function, which logged a React warning. I converted the component to use hooks and changed this to a useLayoutEffect so that it can still force an update after initial render (so that the `ref` gets hooked up) while no longer triggering a warning.

2. Added uncontrolled mode with optional `defaultIsOpen`. This cleans up some use cases a little bit. Controlled mode is still quite common because many places want custom button styling when the dropdown is open.


app/components/ChildToggle/index.tsx
```diff
@@ -67,78 +69,106 @@ type Props = {
...
+    };
+  }, [isOpen, tellAncestorAboutToggledChild]);
+
+  // add a document listener to hide the dropdown body if
+  // it is expanded and the document is clicked on
+  useEventListener(
+    document,
+    "click",
+    isOpen,
+    (e: MouseEvent) => {
```
defunctzombie, line 139: Prefer `ev` rather than single letter variables. Makes it slightly easier to grep for things.

PR #142, opened 2021-03-04
Add input source (player) selection to application toolbar

defunctzombie:


app/components/PlayerManager.tsx
```diff
@@ -242,107 +159,96 @@ function PlayerManager({
...
-          }
-        })
-        .catch((error) => {
-          sendNotification(
-            "Layout failed to load",
-            `Fetching remote file failed. ${corsError(layoutUrl)} ${error}`,
-            "user",
-            "error",
-          );
-        });
```
ruffsl, line 270: I would be handy to have something like this Foxglove Studio again.

PR #143, opened 2021-03-04
Replace node-sass with sass

jhurliman:


jhurliman:
Tracking webviz PR @ https://github.com/cruise-automation/webviz/issues/609

PR #145, opened 2021-03-04
Strongly type almost all of ChartJSManager.ts

jhurliman:


app/components/TimeBasedChart/index.stories.tsx
```diff
@@ -277,8 +278,8 @@ storiesOf("<TimeBasedChart>", module)
   })
   .add("can zoom and then update with new data without resetting the zoom", () => <ZoomExample />)
   .add("cleans up the tooltip when removing the panel", () => <CleansUpTooltipExample />)
-  .add("should call pauseFrame twice", () => <PauseFrameExample />)
+  .add("should call pauseFrame twice", () => <PauseFrameExample {...commonProps} />)
```
jhurliman, line 281: This was the actual fix

app/components/TimeBasedChart/index.stories.tsx
```diff
@@ -277,8 +278,8 @@ storiesOf("<TimeBasedChart>", module)
...
   .add(
     "should still call resumeFrame when removed in the middle of an update (shows `SUCCESS` message with no chart visible)",
-    () => <RemoveChartExample />,
+    () => <RemoveChartExample {...commonProps} />,
```
jhurliman, line 284: Here as well

app/components/ReactChartjs/ChartJSManager.ts
```diff
@@ -236,12 +263,16 @@ export default class ChartJSManager {
       ...otherDataProps,
     };
 
+    // We can safely replace the dataset array, as long as we retain the _meta property
+    // on each dataset.
+    chartInstance.config.data.datasets = datasets;
```
jhurliman, line 268: The order of operations had to be shuffled here a bit because there was no guarantee that `chartInstance.config.data` was defined before.

app/components/ReactChartjs/index.stories.tsx
```diff
@@ -53,12 +53,12 @@ const props = {
     annotations: { annotations: [] },
     plugins: {
       datalabels: {
-        anchor: "top",
+        anchor: "start",
```
jhurliman, line 56: TS enum strings are great. This was a real bug, that had no impact on anything.

app/components/TimeBasedChart/index.tsx
```diff
@@ -163,8 +164,8 @@ type DataSet = Readonly<{
 
 const scalePerPixel = (bounds: ScaleBounds | null | undefined): number | null | undefined =>
   bounds && Math.abs(bounds.max - bounds.min) / Math.abs(bounds.maxAlongAxis - bounds.minAlongAxis);
-const screenCoord = (value: any, valuePerPixel: any) =>
-  valuePerPixel == null ? value : Math.trunc(value / valuePerPixel);
+const screenCoord = (value: number, valuePerPixel: number | null | undefined) =>
+  !valuePerPixel ? value : Math.trunc(value / valuePerPixel);
```
jhurliman, line 168: Modified slightly to avoid a divide by zero. Previously, I think this would return `NaN` if `valuePerPixel=0`.

app/panels/Plot/PlotChart.tsx
```diff
@@ -264,16 +278,17 @@ function getDatasetAndTooltipsFromMessagePlotPath(
     }
   });
 
-  const dataset = {
-    borderColor: lineColors[index % lineColors.length],
+  const borderColor = lineColors[index % lineColors.length] ?? "#DDDDDD";
```
jhurliman, line 281: Had to add this extra fallback that should never happen. Maybe one day TS will know that `lineColors[index % lineColors.length]` has type `string`, not `string | undefined`

jhurliman: Never mind, if `lineColors.length=0` this would be undefined. Fallback is required unless the compiler can prove `lineColors.length > 0`

app/panels/Plot/PlotChart.tsx
```diff
@@ -286,13 +301,14 @@ function getDatasetAndTooltipsFromMessagePlotPath(
...
-function getAnnotationFromReferenceLine(path: PlotPath, index: number) {
+function getAnnotationFromReferenceLine(path: PlotPath, index: number): AnnotationOptions {
+  const borderColor = lineColors[index % lineColors.length] ?? "#DDDDDD";
   return {
     type: "line",
     drawTime: "beforeDatasetsDraw",
     scaleID: Y_AXIS_ID,
-    label: path.value,
-    borderColor: lineColors[index % lineColors.length],
+    label: { content: path.value },
```
jhurliman, line 310: Apparent bug, `label` is a complex object type and doesn't accept a string according to the TS definitions. It's possible that it was an undocumented feature, but this should be equivalent.

PR #148, opened 2021-03-05
Add reg-suit screenshot testing

amacneil:


package.json
```diff
@@ -16,20 +16,23 @@
...
+    "test": "jest",
+    "test:watch": "jest --watch",
     "package:win": "yarn package --win",
     "package:darwin": "yarn package --mac",
     "package:linux": "yarn package --linux",
     "package": "electron-builder",
     "storybook": "yarn workspace @foxglove-studio/app run start-storybook",
-    "build:storybook": "yarn workspace @foxglove-studio/app run build-storybook"
+    "storybook:build": "yarn workspace @foxglove-studio/app run build-storybook",
+    "storybook:storycap": "yarn workspace @foxglove-studio/app run storycap http://localhost:9001 --serverCmd \"yarn http-server storybook-static -p 9001\" --outDir \"storybook-screenshots\"",
```
amacneil, line 33: using `http-server` here rather than running `start-storybook` inline so that we get more explicit errors

.github/workflows/ci.yml
```diff
@@ -45,5 +45,15 @@ jobs:
           key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
           restore-keys: |
             ${{ runner.os }}-yarn-
+      - name: Configure Google Cloud
```
amacneil, line 48: this step currently wastes about 20s. I think I can find a faster way to do it, but will address in a follow up PR

PR #151, opened 2021-03-05
Add electron-updater

defunctzombie:


desktop/index.ts
```diff
@@ -187,6 +188,10 @@ async function createWindow(): Promise<void> {
 // initialization and is ready to create browser windows.
 // Some APIs can only be used after this event occurs.
 app.on("ready", async () => {
+  autoUpdater.checkForUpdatesAndNotify().catch((err) => {
+    console.error(err);
```
amacneil, line 192: report to sentry? or does `console.error()` trigger sentry?

defunctzombie: 👍 Will update to explicitly capture error. console.error doesn't report to sentry.

PR #155, opened 2021-03-05
Fix NotificationModal sticking around when switching between stories

jtbandes:
The biggest change here is from an imperative `showNotificationModal()` function to a `<NotificationModal />` component that can be rendered. The backing implementation is the same (uses renderToBody) but now we can auto-cleanup when the component is unmounted.


defunctzombie:
One day we should convert sendNotification to a context provider pattern with a useNotifications hook:

Inspiration: https://github.com/iamhosseindhv/notistack

defunctzombie:
LGTM

PR #158, opened 2021-03-05
reg-suit: don't set commit status on main branch

amacneil:
Two changes:

- The reg-suit snapshot is now uploaded under the merge commit sha (rather than PR branch head sha), which is more accurate since github actions actually run against a merge of the PR with main.
- Reg-suit is now configured to not set a commit status when it runs on main. I don't think it is helpful setting red status checks on commits to main since a change does not mean "failed" there. We will still get status checks and comments on PRs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,5 +55,5 @@ jobs:
       - run: yarn run ${{ matrix.config.command }}
         env:
           REG_SUIT_GITHUB_CLIENT_ID: ${{ secrets.REG_SUIT_GITHUB_CLIENT_ID }}
-          REG_SUIT_HEAD_SHA: ${{ github.event.pull_request.head.sha || github.sha }}
-          REG_SUIT_BASE_SHA: ${{ github.event.pull_request.base.sha || github.sha }}
+          REG_SUIT_HEAD_SHA: ${{ github.sha }}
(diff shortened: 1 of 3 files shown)


ci/storybook.sh
```diff
@@ -15,17 +15,23 @@ set -euo pipefail
 
 # If this is a PR, then by default github actions has us on a merge commit.
 # Checkout the feature branch and commit to make reg-suit happy.
-branch=${GITHUB_HEAD_REF:-}
```
amacneil, line 18: todo: everyone says I have to rewrite this in typescript

amacneil:
@jtbandes fyi

.github/workflows/ci.yml
```diff
@@ -55,5 +55,5 @@ jobs:
       - run: yarn run ${{ matrix.config.command }}
         env:
           REG_SUIT_GITHUB_CLIENT_ID: ${{ secrets.REG_SUIT_GITHUB_CLIENT_ID }}
-          REG_SUIT_HEAD_SHA: ${{ github.event.pull_request.head.sha || github.sha }}
-          REG_SUIT_BASE_SHA: ${{ github.event.pull_request.base.sha || github.sha }}
+          REG_SUIT_HEAD_SHA: ${{ github.sha }}
```
amacneil, line 58: using the merge commit sha here instead of the PR head sha

PR #160, opened 2021-03-05
Remove Storybook padding around body

jtbandes:
This caused some stories such as UnlinkGlobalVariables to behave differently than in the webviz repo due to different element positioning.


jtbandes:
image

PR #163, opened 2021-03-05
Windows in CI, switch to react-document-events 1.5.1, fix rosbag.js checksum mismatch

jhurliman:


.github/workflows/ci.yml
```diff
@@ -35,7 +35,7 @@ jobs:
       - name: Configure Node.js
         uses: actions/setup-node@v2.1.5
         with:
-          node-version: 15.x
+          node-version: ${{ matrix.node-version }}
```
amacneil, line 38: @jhurliman where is this getting set? I think this is evaluating to empty string now so node is unpinned

PR #164, opened 2021-03-05
Fix TimeBasedChart vertical bar CSS

jtbandes:
There was a CSS syntax error resulting from the removal of `styled.*.attrs` during TS conversion. We may need to do an audit of other places that used to use styled.attrs since it can be a performance optimization.


jhurliman:
Nice: https://storage.googleapis.com/foxglove-studio-reg-suit/37aaee52eb43fb8a73aa4105ec44ebf3e7559276/index.html#changed-TwoDimensionalPlot/zooms%20vertically%20to%20show%20reset%20view%20button.png

jtbandes:
Still looks like the plot isn't there...

PR #165, opened 2021-03-06
Add integration test

defunctzombie:


jest.integration.ts
```diff
@@ -0,0 +1,15 @@
...
+export default {
+  preset: "ts-jest",
+  globals: {
+    "ts-jest": {
+      tsconfig: "<rootDir>/tsconfig.json",
+    },
+  },
+  restoreMocks: true,
+  testPathIgnorePatterns: ["/node_modules/", "/app"],
+  modulePathIgnorePatterns: ["<rootDir>/.webpack", "<rootDir>/dist"],
```
jhurliman, line 14: This ignore patterns may or may not work in Windows.

defunctzombie: I've changed the config file and `test:integration` command to not need these.

desktop/index.ts
```diff
@@ -232,15 +232,11 @@ app.on("ready", async () => {
   // See: https://www.electronjs.org/docs/tutorial/security
   const contentSecurtiyPolicy: Record<string, string> = {
     "default-src": "'self'",
-    "script-src": `'self' 'unsafe-inline'`,
+    "script-src": `'self' 'unsafe-inline' 'unsafe-eval'`,
     "style-src": "'self' 'unsafe-inline'",
     "connect-src": "'self' ws: wss: http: https:", // Required for rosbridge connections
     "font-src": "'self' data:",
   };
-  if (!isProduction) {
```
defunctzombie, line 240: @amacneil fyi - this failed our integration test because the prod build still needs unsafe-eval. We have places where we dynamically build functions (rosbag, and bayerImage).

webpack.main.config.ts
```diff
@@ -8,7 +8,7 @@ import { Configuration, ResolveOptions, DefinePlugin, EnvironmentPlugin } from "
 
 import { WebpackArgv } from "./WebpackArgv";
 
-export default (_: never, argv: WebpackArgv): Configuration => {
+export default (_: unknown, argv: WebpackArgv): Configuration => {
```
amacneil, line 11: out of curiosity why would one use unknown vs never here?

jhurliman:
```
const compiler = webpack([
  mainConfig(undefined, webpackArgs), // Argument of type 'undefined' is not assignable to parameter of type 'never'.ts(2345)
  preloadConfig(undefined, webpackArgs), // Argument of type 'undefined' is not assignable to parameter of type 'never'.ts(2345)
  renderConfig(undefined, webpackArgs),
]);
```

defunctzombie: To expand on what John said - `never` in typescript instructs the compiler to "error" if, after resolving the various type arguments in a call, any of them are `never`. It is a way of deducing type arguments mostly for generics. I used it here without thinking much about it since we didn't call these function ourselves but the correct type is `unknown` since this function _can_ be called with an argument. `never` isn't synonymous with "we don't care what this type is" - it is synonymous with "this is not callable".

.github/workflows/ci.yml
```diff
@@ -91,3 +91,31 @@ jobs:
...
+        uses: actions/cache@v2.1.4
+        with:
+          path: |
+            .yarn/cache
+            **/node_modules
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-yarn-
+
+      - run: git config --global core.eol lf
```
defunctzombie, line 118: Remove this - it was not the issue

webpack.main.config.ts
amacneil: makes sense, thanks

PR #168, opened 2021-03-06
Document changes to joinTopics()

jhurliman:


app/util/topicUtils.ts
```diff
@@ -15,6 +15,10 @@ import { flatten } from "lodash";
 
 // Joins arrays of topics with proper slashes similar to node's path.join
 export const joinTopics = (...topics: string[]): string => {
+  // We can't use path.join() here because it uses OS-specific path delimiters
+  // and we always want "/"". This trims leading and trailing forward slashes off
+  // each topic component (so "/a/b/c/" becomes "a/b/c") then joins all topics
+  // together with "/"
   const parts = [...topics].map((t) => t.replace(/^\/+|\/+$/g, ""));
```
defunctzombie, line 22: Btw - this makes a new array of again with all the topics. This should be `topics.map(...` so you run map over the existing array

jtbandes: I think this comment could be move above the function since it’s documenting behavior and not implementation details. The part about not using the system join can be removed since there’s no reason to expect it would use that anyway

PR #169, opened 2021-03-06
Move jest.config.ts to app directory

defunctzombie:
The top level jest config is actually specific to app.


defunctzombie:
This paves the way for our integration tests which use their own jest config:

https://github.com/foxglove/studio/pull/165

PR #170, opened 2021-03-06
Add jest transform to support `?raw` import syntax

defunctzombie:
`?raw` indicates to webpack that an import statement should be treated as the raw content of the file. Jest doesn't know how to properly import this because it wants to resolve the entire module name (including the `?raw` string). This does not resolve to an actual file on disk and jest complains. Using `moduleNameMapper` in jest.config is also incorrect since it still has jest perform a resolution on the aliased string. Jest will try to transform it with the appropriate transformer for that file type. In the case of our raw imports in nodePlayground that file type is `.ts` and it will try the ts-jest loader. We do not want that - we want such an import to be loaded as raw content so we need to prevent jest from ever trying to resolve it to the typescript loader.

The more robust solution is to use the typescript ast rather than regex for import statements. The even more robust solution is to write a transform which uses webpack to compile the test file since all our code is written with the assumption it will run through a bundler. This solution is not currently possible since jest transforms must be sync and I think webpack compilation is async. There are some discussions within jest to add support for async transforms. If/when that does land this is a possible way forward.

### DIFF
--- a/app/jest.config.ts
+++ b/app/jest.config.ts
@@ -5,7 +5,7 @@
 export default {
   preset: "ts-jest",
   transform: {
-    "\\.tsx?$": "<rootDir>/test/transformers/webpacker",
+    "\\.tsx?$": "<rootDir>/test/transformers/rawImportPreprocessor.ts",
     "\\.ne$": "<rootDir>/test/transformers/neTransformer.js",
     "\\.(bin|template|wasm)$": "<rootDir>/test/transformers/rawTransformer.js",
   },
--- a/app/jest.config.ts
+++ b/app/jest.config.ts
@@ -2,9 +2,13 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
-// Shared configuration used by all jest projects
-const sharedConfig = {
+export default {
   preset: "ts-jest",
+  transform: {
+    "\\.tsx?$": "<rootDir>/test/transformers/webpacker",
+    "\\.ne$": "<rootDir>/test/transformers/neTransformer.js",
+    "\\.(bin|template|wasm)$": "<rootDir>/test/transformers/rawTransformer.js",
+  },
   globals: {
     "ts-jest": {
       tsconfig: "<rootDir>/tsconfig.jest.json",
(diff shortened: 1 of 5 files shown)


defunctzombie:
Not this is not a `1:1` replication of webpack raw loader since that loader actually stubs the imported module with `module.export = "string"`. I cannot do that here since I do not want any further resolution of the import statement. After performing the raw loading replacement it should be like there is no more import statement - the file is now part of the source code. Multiple import locations of the same raw file would result in multiple copies in their sources - but since this is only for testing rather than bundling it makes no difference.

defunctzombie:
@jhurliman there is still one failing windows test - I don't have a windows machine to iterate on so if you can take a look at why this particular test fails that would be helpful.

app/jest.config.ts
```diff
@@ -2,9 +2,13 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
-// Shared configuration used by all jest projects
-const sharedConfig = {
+export default {
   preset: "ts-jest",
+  transform: {
+    "\\.tsx?$": "<rootDir>/test/transformers/webpacker",
```
jtbandes, line 8: Can we pick a better name for this transform - maybe it should be named rawImportTransformer and the current rawTransformer.js should be renamed to `assetSourceStub` or something more clear?

app/players/UserNodePlayer/nodeTransformerWorker/typescript/rawUserUtils.ts
```diff
@@ -2,16 +2,12 @@
...
-import pointClouds from "@foxglove-studio/app/players/UserNodePlayer/nodeTransformerWorker/typescript/userUtils/pointClouds?raw";
-import readers from "@foxglove-studio/app/players/UserNodePlayer/nodeTransformerWorker/typescript/userUtils/readers?raw";
-import time from "@foxglove-studio/app/players/UserNodePlayer/nodeTransformerWorker/typescript/userUtils/time?raw";
-import types from "@foxglove-studio/app/players/UserNodePlayer/nodeTransformerWorker/typescript/userUtils/types?raw";
-import vectors from "@foxglove-studio/app/players/UserNodePlayer/nodeTransformerWorker/typescript/userUtils/vectors?raw";
-
-if (markers === undefined) {
-  throw new Error("markers?raw import failed");
-}
+import markers from "@foxglove-studio/app/players/UserNodePlayer/nodeTransformerWorker/typescript/userUtils/markers.ts?raw";
```
jtbandes, line 5: just curious: why was the `.ts` extension necessary here? Especially if `require.resolve` is being used in the transformer.

app/test/transformers/webpacker.js
```diff
@@ -0,0 +1,31 @@
...
+// re-write these into `const variable = "string source";`;
+const importRegEx = /^import (.*) from "(.*)\?raw";$/gm;
+const importReplacer = (_, p1, p2) => {
+  const resolved = require.resolve(p2);
+  const rawFile = fs.readFileSync(resolved, { encoding: "utf-8" });
+  return `const ${p1} = ${JSON.stringify(rawFile.toString())};`;
+};
+
+function rewriteSource(source) {
+  return importRegEx[Symbol.replace](source, importReplacer);
```
jtbandes, line 21: Maybe this is a philosophical question about the JS standard library, but why is Symbol.replace used instead of just `.replace`? Can this be changed to `source.replace(importRegEx, importReplacer)`?

app/test/transformers/webpacker.js
```diff
@@ -0,0 +1,31 @@
...
+
+/* eslint-disable @typescript-eslint/no-var-requires */
+const fs = require("fs");
+const tsjest = require("ts-jest");
+
+const transformer = tsjest.createTransformer();
+
+// look for `?raw` import statements
+// re-write these into `const variable = "string source";`;
+const importRegEx = /^import (.*) from "(.*)\?raw";$/gm;
```
jtbandes, line 13: Nit: might want to validate that p1 is a single identifier. We wouldn't support `import {a,b} from "...?raw"` and should probably throw an error in that case.

app/test/transformers/webpacker.js
```diff
@@ -0,0 +1,31 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+/* eslint-disable @typescript-eslint/no-var-requires */
+const fs = require("fs");
+const tsjest = require("ts-jest");
```
jtbandes, line 7: Does ts-jest not allow for our custom transformers to be written in TS?

app/players/UserNodePlayer/nodeTransformerWorker/typescript/rawUserUtils.ts
defunctzombie: It isn't - I felt this was clearer to indicate we are loading a file rather than importing.

app/jest.config.ts
defunctzombie: The existing transformer is also a rawTransformer since it does transform a file. I can rename webpacker to `rawImportPreprocessor` since that is spiritually what it does.

app/test/transformers/webpacker.js
defunctzombie: You know I didn't really try since I thought it would be circular. Like the transform in the jest config would try to transform itself with the same transform. But maybe it will somehow work out...

app/test/transformers/webpacker.js
defunctzombie: Yea - I can flip it to that. There is no real difference between them - source.replace is implemented by calling the Symbol.replace function like I do.

defunctzombie: Yea - I can flip it to that. There is no real difference between them - source.replace is implemented by calling the Symbol.replace function like I do.

PR #172, opened 2021-03-06
Add unsafe-eval to our script-src Content Security Policy

defunctzombie:


defunctzombie:
@amacneil fyi

desktop/index.ts
```diff
@@ -232,15 +232,11 @@ app.on("ready", async () => {
   // See: https://www.electronjs.org/docs/tutorial/security
   const contentSecurtiyPolicy: Record<string, string> = {
     "default-src": "'self'",
-    "script-src": `'self' 'unsafe-inline'`,
+    "script-src": `'self' 'unsafe-inline' 'unsafe-eval'`,
```
amacneil, line 235: maybe add a comment here about why it's needed and whether there is any hope of removing it in the future

PR #173, opened 2021-03-07
Add integration test

jhurliman:


.github/workflows/ci.yml
```diff
@@ -7,7 +7,7 @@ on:
     branches: [main]
 
 jobs:
-  build:
+  test:
```
defunctzombie, line 10: Why rename this?

.github/workflows/ci.yml
```diff
@@ -35,7 +35,7 @@ jobs:
       - name: Configure Node.js
         uses: actions/setup-node@v2.1.5
         with:
-          node-version: ${{ matrix.node-version }}
+          node-version: 15.x
```
defunctzombie, line 38: Fascinating - we didn't have a node-version in the matrix - what was this doing before?

.github/workflows/ci.yml
```diff
@@ -90,4 +104,8 @@ jobs:
   #         restore-keys: |
   #           ${{ runner.os }}-yarn-
   #     - run: yarn install --immutable
-  #     - run: yarn run ${{ matrix.config.command }}
+  #     - name: test:integration
```
defunctzombie, line 107: Lets remove the 3rd party action and put back `yarn run test:integration` along with changing the "runs-on" to `windows-latest`.

.github/workflows/ci.yml
jhurliman: I was breaking build out to a separate top-level item that ran on Windows. If we end up running integration tests on Windows this isn't necessary.

.github/workflows/ci.yml
jhurliman: :shrug:

PR #174, opened 2021-03-07
Improve PoseMarker type information

defunctzombie:
I also changed how `<PoseMarkers>` accepts the markers. Previously it used children whereas now it uses a property. I prefer conventionally using children for other react components rather than just another property of any type.

I've increased the screenshot delay on PostMarkers.stories with the hope of making it less flaky.

### DIFF
--- a/app/panels/ThreeDimensionalViz/commands/PoseMarkers.stories.tsx
+++ b/app/panels/ThreeDimensionalViz/commands/PoseMarkers.stories.tsx
@@ -12,10 +12,12 @@
 //   You may not use this file except in compliance with the License.
 
 import { storiesOf } from "@storybook/react";
-import * as React from "react";
+import { ComponentProps } from "react";
 import { Worldview, DEFAULT_CAMERA_STATE, Color } from "regl-worldview";
 
 import PoseMarkers from "./PoseMarkers";
+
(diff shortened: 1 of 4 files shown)


app/panels/ThreeDimensionalViz/commands/PoseMarkers.stories.tsx
```diff
@@ -85,17 +87,17 @@ function Example({
...
   );
 }
 
 storiesOf("<3DViz> / PoseMarkers", module)
   .addParameters({
     screenshot: {
-      delay: 3000,
+      // the car-model marker loads the car models asyncronously
+      // we delay screenshot until some time with the hope it has loaded
+      delay: 5000,
```
jtbandes, line 100: There is a `window.waitFor` technique we could try and use to explicitly allow screenshotting when loading is complete. But we would probably need to set up some management so that multiple components can add waitFor promises if needed without overwriting each other.

defunctzombie: I think even with this delay the car is missing from the screenshot. Which feels like there must be something else going on since it wouldn't take that long to load that car model via fetch - its basically a local load.

I need to find a sensible way to test that the delay is working.

PR #176, opened 2021-03-08
Add afterSign script to notarize macOS apps

defunctzombie:


defunctzombie:
@jtbandes do you know the universe of entitlements we might want? Are these things like "access to my documents" etc?

resources/mac/entitlements.plist
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+  <dict>
+    <key>com.apple.security.cs.allow-unsigned-executable-memory</key>
```
jtbandes, line 5: Why are we using this and not `com.apple.security.cs.allow-jit`? https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_security_cs_allow-jit

resources/notarize.cjs
```diff
@@ -0,0 +1,40 @@
+const { notarize } = require("electron-notarize");
+const builderUtil = require("builder-util");
+
+exports.default = async function notarizing(context) {
```
jtbandes, line 4: Any way to use TS for this file?

nit: default export name should match file name
```suggestion
exports.default = async function notarize(context) {
```

resources/mac/entitlements.plist
jtbandes: Just looking at the default: https://github.com/electron/electron-osx-sign/blob/master/default.entitlements.mas.plist

Maybe we should enable these?
- com.apple.security.app-sandbox
- com.apple.security.network.client
- com.apple.security.files.user-selected.read-only

defunctzombie: I was mostly following https://kilianvalkhof.com/2019/electron/notarizing-your-electron-application/ which was linked from the electron-builder docs

resources/notarize.cjs
defunctzombie: I think we would need to run electron-builder via `ts-node` so the _require_ hooks are injected. Maybe there is a _prescript_ I could tell electron-builder about. I'll investigate.

resources/mac/entitlements.plist
jtbandes: There is a list of available ones here: https://developer.apple.com/documentation/security/app_sandbox

I would hope the `allow-jit` is enough for JS and WASM to work but it does look like other people are using this `allow-unsigned-executable-memory`. But from the documentation it sounds like this one is not preferred and only for rare deprecated use cases.

defunctzombie: If I build the .app locally and run it to see if things work as expected is that sufficient for checking the entitlements or do those only affect the app once signed or notarized?

jtbandes: I think it may need to be actually signed. It might be possible to sign with a test/personal id that we wouldn't use for distribution.

defunctzombie: Updated entitlements

defunctzombie:
I've typescrit-ified the notarize afterSign script. That caught some bugs 👍 . Once I get a cert I will test this flow and then merge. We can iterate on the entitlements from there.

resources/mac/entitlements.plist
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+  <dict>
+    <key>com.apple.security.cs.allow-jit</key>
+    <key>com.apple.security.app-sandbox</key>
+    <key>com.apple.security.network.client</key>
```
jtbandes, line 7:
```suggestion
    <key>com.apple.security.cs.allow-jit</key>
    <true />
    <key>com.apple.security.app-sandbox</key>
    <true />
    <key>com.apple.security.network.client</key>
    <true />
```

PR #191, opened 2021-03-09
Replace Dimensions component with react-resize-detector

defunctzombie:
We already had some uses of react-resize-detector which is a better supported version of our internal Dimensions component.


app/panels/Plot/PlotChart.tsx
```diff
@@ -422,32 +422,30 @@ export default memo<PlotChartProps>(function PlotChart(props: PlotChartProps) {
...
-            saveCurrentView={saveCurrentView}
-            xAxisIsPlaybackTime={xAxisVal === "timestamp"}
-            scaleOptions={scaleOptions}
-            currentTime={currentTime}
-            defaultView={defaultView}
-            onClick={onClick}
-          />
-        )}
-      </Dimensions>
+      <TimeBasedChart // Force a redraw every time the x-axis value changes.
```
defunctzombie, line 431: This might be wrong since the dimensions are for the outer div so we may need to switch this to the ResizeDetector  component instead of the hook.

defunctzombie: It seems that useResizeDetector can leverage custom refs so you can point it at a specific component. Neat!

app/panels/Plot/PlotChart.tsx
```diff
@@ -422,32 +422,30 @@ export default memo<PlotChartProps>(function PlotChart(props: PlotChartProps) {
     tooltips,
     xAxisVal,
   } = props;
-  const annotations = getAnnotations(paths);
+
+  const annotations = useMemo(() => getAnnotations(paths), [paths]);
```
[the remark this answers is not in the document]
jtbandes: I dunno. The component is already wrapped in a memo, so not sure how important this is, but I think it's probably safe. You can test it by using the Plot panel :)

PR #192, opened 2021-03-09
ROS1 comms library

jhurliman:


packages/ros1/LICENSE
```diff
@@ -0,0 +1,202 @@
+
```
defunctzombie, line 1: @amacneil confirm we want to use apache v2 for this?

packages/ros1/sample-robot-docker/entrypoint.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+mkdir -p ~/catkin_ws/src
+cd ~/catkin_ws/
```
defunctzombie, line 4: oh catkin - how I [haven't] missed you

packages/ros1/src/ITcp.ts
```diff
@@ -0,0 +1,38 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
defunctzombie, line 1: File rename `Tcp.ts`? `TcpTypes.ts` or maybe just putting the bits in their own files.

packages/ros1/src/IXmlRpc.ts
```diff
@@ -0,0 +1,15 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
defunctzombie, line 1: `XmlRpcTypes` ? same thoughts as the tcp file comments

packages/ros1/src/RosMaster.ts
```diff
@@ -0,0 +1,103 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { URL } from "url";
+
+import { XmlRpcCreateClient, XmlRpcClient, XmlRpcResponse } from "./IXmlRpc";
+
+export default class RosMaster {
+  private _client: XmlRpcClient;
```
defunctzombie, line 10: lets start using the new `#` syntax for private fields (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields). This ensures true private behavior. You don't need the `private` keyword with them.

packages/ros1/src/RosMaster.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  constructor(options: {
+    xmlRpcCreateClient: XmlRpcCreateClient;
+    uri?: URL;
```
defunctzombie, line 14: I'd consider only supporting uri (via a string) so the user should always pass in something of the form of protocol://host:port and we decode ourselves - feels like fewer code paths and less variance at the use site.

packages/ros1/src/IXmlRpc.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export type XmlRpcValue = string | number | XmlRpcValue[];
+
+export type XmlRpcResponse = [code: number, msg: string, value: XmlRpcValue];
+
+export interface XmlRpcClient {
+  methodCall(method: string, args: XmlRpcValue[]): Promise<XmlRpcResponse>;
+}
+
+export interface XmlRpcCreateClient {
```
defunctzombie, line 13: Conventionally I've seen interfaces like this called `XmlRpcClientConstructor` and you should add a `new` in front of the function.

packages/ros1/src/RosMaster.ts
```diff
@@ -0,0 +1,103 @@
...
+
+import { URL } from "url";
+
+import { XmlRpcCreateClient, XmlRpcClient, XmlRpcResponse } from "./IXmlRpc";
+
+export default class RosMaster {
+  private _client: XmlRpcClient;
+
+  constructor(options: {
+    xmlRpcCreateClient: XmlRpcCreateClient;
```
defunctzombie, line 13: I think this should accept type `XmlRpc` rather than a create function and leave creation to the user

packages/ros1/src/node/xmlrpc-rosnodejs.d.ts
```diff
@@ -0,0 +1,108 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
defunctzombie, line 1: Custom typings customarily live in a `typings` folder - see how we've set this up for the root project.

packages/ros1/README.md
```diff
@@ -0,0 +1,15 @@
+# node-ros1
+
+node.js TypeScript library for interfacing with ROS1.
```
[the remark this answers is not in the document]
jhurliman: You're right. This started as a node library and became more generic as I went.

packages/ros1/LICENSE
amacneil: can we keep everything in monorepo MPL for now and relicense if/when we separate it? I would like to have a clear licensing story rather than files everywhere with different licenses

amacneil: either that or just start with a separate repo on day 1, but that sounds painful to iterate on to me.

also I think MIT is fine (rather than apache) for anything we decide not to use MPL for, it's a bit simpler / less overhead

packages/ros1-node/README.md
```diff
@@ -0,0 +1 @@
+# @foxglove/ros1-node
```
defunctzombie, line 1: consider `ros1-nodejs` just cause "node" is already such an overloaded term in rosland

packages/ros1/src/IXmlRpc.ts
jhurliman: Add a `new` where? When calling it? What happens if you don't?

defunctzombie:
```typescript
export interface XmlRpcConstructor {
   new (...): XmlRpc
}
```

This is an interface that describes a constructor. From what I saw in your stuff tho I would accept XmlRpc rather than building it - let the user build it.

packages/ros1/src/RosSlave.ts
```diff
@@ -0,0 +1,185 @@
...
+    if (Array.isArray(proto) && proto.length > 0) {
+      if (proto[0] === "TCPROS") {
+        return true;
+      }
+    }
+  }
+  return false;
+}
+
+export class RosSlave {
```
jtbandes, line 35: I know some of these things reference core ROS concepts (e.g. "ros master"), but I think it is worth the effort now to at least replace 'slave' with something like 'follower', 'client' etc. See some examples at https://en.wikipedia.org/wiki/Master/slave_(technology)#Terminology_concerns

packages/ros1-nodejs/README.md
```diff
@@ -0,0 +1,15 @@
+# @foxglove/ros1-nodejs
+
+node.js implementation for the @foxglove/ros1 library.
+
+### Usage
+
+TODO
```
defunctzombie, line 7: Remove `Usage` section - we can revisit later. TODO looks lame if someone opens the readme.

packages/ros1-nodejs/src/PlatformNode.ts
```diff
@@ -0,0 +1,80 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import os from "os";
+import { URL } from "whatwg-url";
+
+function IsPrivateIP(ip: string): boolean {
```
jtbandes, line 8: naming: capitals should be reserved for type/class/component names, lowerCamel for functions
```suggestion
function isPrivateIP(ip: string): boolean {
```

packages/ros1-nodejs/package.json
```diff
@@ -0,0 +1,42 @@
...
+  "private": true,
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/foxglove/studio.git"
+  },
+  "author": {
+    "name": "Foxglove Technologies",
+    "email": "support@foxglove.dev"
+  },
+  "homepage": "https://github.com/foxglove/studio",
```
defunctzombie, line 15: https://foxglove.dev

packages/ros1-nodejs/package.json
```diff
@@ -0,0 +1,42 @@
...
+  "main": "src/index.ts",
+  "module": "src/index.ts",
+  "files": [
+    "src"
+  ],
+  "engines": {
+    "node": ">=12"
```
defunctzombie, line 22: Maybe omit engines field? Top level package.json can specify it if we care.

packages/ros1-nodejs/package.json
```diff
@@ -0,0 +1,42 @@
...
+  },
+  "scripts": {},
+  "prettier": {
```
defunctzombie, line 25: Remove - we should use the prettier config in the top level of the repo to keep one consistent config.

packages/ros1-nodejs/src/PlatformNode.ts
```diff
@@ -0,0 +1,80 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
defunctzombie, line 1: Not sure what ties all the functions in this file together. Maybe a comment indicating what should be in this file so it doesn't become a dumping ground for _stuff_.

packages/ros1-nodejs/src/TcpMessageStream.ts
```diff
@@ -0,0 +1,65 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { Transform, TransformCallback } from "stream";
+
+export class TcpMessageStream extends Transform {
+  #inMessage = false;
+  #bytesNeeded = 4;
+  #chunks: Uint8Array[] = [];
+
+  _transform(chunk: Uint8Array, _encoding: BufferEncoding, callback: TransformCallback): void {
```
defunctzombie, line 12: Prefer `private` typescript keyword and no `_` prefix.

packages/ros1-nodejs/src/TcpMessageStream.ts
```diff
@@ -0,0 +1,65 @@
...
+      }
+
+      this.#inMessage = !this.#inMessage;
+      this.#chunks = [];
+    }
+
+    callback();
+  }
+
+  static ConcatData(chunks: Uint8Array[]): Uint8Array {
```
defunctzombie, line 47: Looks like this is unrelated to this class and belongs in a separate file.

packages/ros1-nodejs/src/TcpNode.ts
```diff
@@ -0,0 +1,149 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { TcpAddress, TcpServer, TcpSocket } from "@foxglove/ros1";
+import EventEmitter from "eventemitter3";
+import * as net from "net";
```
defunctzombie, line 7: `import net from "net";` should work, no?

packages/ros1-nodejs/src/TcpNode.ts
```diff
@@ -0,0 +1,149 @@
...
+    const family = this.#socket.remoteFamily; // There is no localFamily
+    const address = this.#socket.localAddress;
+    return port !== undefined && family !== undefined && address !== undefined
+      ? { port, family, address }
+      : undefined;
+  }
+
+  fd(): number | undefined {
+    // eslint-disable-next-line no-underscore-dangle
+    return ((this.#socket as unknown) as MaybeHasFd)._handle?.fd;
```
defunctzombie, line 63: What stopped the cast from this.#socket directly to `MaybeHasFd` ? Do sockets not always have `_handle`? Is there a way to specify that `this.#socket` should be a socket with a handle or we don't want that limitation?

packages/ros1-nodejs/src/TcpNode.ts
```diff
@@ -0,0 +1,149 @@
...
+          reject(err);
+          return;
+        }
+        resolve();
+      });
+    });
+  }
+}
+
+export class TcpServerNode extends EventEmitter implements TcpServer {
```
defunctzombie, line 87: Lets try to do one class per file. So `TcpServerNode.ts`, `TcpNode.ts`.

packages/ros1-nodejs/src/TcpNode.ts
```diff
@@ -0,0 +1,149 @@
...
+    }
+    return addr;
+  }
+
+  close(): void {
+    this.#server.close();
+  }
+}
+
+export function TcpListen(options: {
```
defunctzombie, line 114: Separate file.

packages/ros1-nodejs/src/TcpNode.ts
```diff
@@ -0,0 +1,149 @@
...
+    const server = net.createServer();
+    server.on("error", reject);
+    server.listen(options.port, options.host, options.backlog, () => {
+      server.removeListener("error", reject);
+      resolve(new TcpServerNode(server));
+    });
+  });
+}
+
+export function TcpConnect(options: { host: string; port: number }): Promise<TcpSocket> {
```
defunctzombie, line 129: Separate file

packages/ros1-nodejs/tsconfig.json
```diff
@@ -0,0 +1,36 @@
+{
```
defunctzombie, line 1: This file should extend our base tsconfig.json in repo root and override only a minimal set of this required for this project.

packages/ros1-turtlesim-test/tsconfig.json
```diff
@@ -0,0 +1,31 @@
+{
```
defunctzombie, line 1: This file should extend our base tsconfig.json

packages/ros1/package.json
```diff
@@ -0,0 +1,45 @@
...
+  "main": "src/index.ts",
+  "module": "src/index.ts",
+  "files": [
+    "src"
+  ],
+  "engines": {
+    "node": ">=12"
+  },
+  "scripts": {},
+  "prettier": {
```
defunctzombie, line 25: use prettier from our top level config - should not have a separate one

packages/ros1/src/TcpConnection.ts
```diff
@@ -0,0 +1,133 @@
...
+import { MessageReader, parseMessageDefinition, RosMsgDefinition } from "rosbag";
+import { TextDecoder, TextEncoder } from "web-encoding";
+
+import { Connection } from "./Connection";
+import { TcpAddress, TcpSocket } from "./TcpTypes";
+
+type Header = [string, string][];
+
+export class TcpConnection implements Connection {
+  _socket: TcpSocket;
```
defunctzombie, line 14: Prefer `#` for private fields.

packages/ros1/src/TcpConnection.ts
```diff
@@ -0,0 +1,133 @@
...
+    const addr = this._socket.remoteAddress();
+    const fd = this._socket.fd() ?? -1;
+    if (addr) {
+      const { address, port } = addr;
+      return `TCPROS connection on port ${localPort} to [${address}:${port} on socket ${fd}]`;
+    }
+    return `TCPROS not connected [socket ${fd}]`;
+  }
+
+  #handleClose = (): void => {
```
defunctzombie, line 61: Were you trying different approaches to making private methods? Should we be doing this or using `private` keyword for methods? Ideally we are consistent

packages/ros1/src/index.ts
```diff
@@ -0,0 +1,23 @@
...
+export * from "./RosSlave";
+export * from "./RosSlaveClient";
+export * from "./SubscriberLink";
+export * from "./Subscription";
+export * from "./TcpConnection";
+export * from "./TcpTypes";
+export * from "./XmlRpcTypes";
+
+// Mock backend
+export * as PlatformMock from "./mock/PlatformMock";
```
defunctzombie, line 21: Think it is useful to export the mocks?

packages/ros1/src/mock/TcpMock.ts
```diff
@@ -0,0 +1,67 @@
...
+  address(): TcpAddress | undefined {
+    return this.listening ? { address: "192.168.1.1", port: 20000, family: "IPv4" } : undefined;
+  }
+
+  close(): void {
+    this.listening = false;
+  }
+}
+
+export function TcpListen(_options: {
```
defunctzombie, line 57: I think you can omit the options arg altogether? Or does that fail typechecks when used in other places?

packages/ros1-nodejs/src/TcpNode.ts
```diff
@@ -0,0 +1,149 @@
...
+
+type MaybeHasFd = {
+  _handle:
+    | {
+        fd: number | undefined;
+      }
+    | undefined;
+};
+
+export class TcpSocketNode extends EventEmitter implements TcpSocket {
```
jtbandes, line 19: Does this "node" mean "node.js"? Maybe call it `TcpSocketNodeJS` to be clearer.

packages/ros1/tsconfig.json
```diff
@@ -0,0 +1,36 @@
+{
```
defunctzombie, line 1: Should extend the tsconfig.json at project root

packages/ros1/src/RosSlave.ts
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    const publications = this.#rosNode.publications.values();
+    const subscriptions = this.#rosNode.subscriptions.values();
+
+    const publishStats: XmlRpcValue[] = Array.from(publications, (pub, _) => pub.getStats());
+    const subscribeStats: XmlRpcValue[] = Array.from(subscriptions, (sub, _) => sub.getStats());
+    const serviceStats: XmlRpcValue[] = [];
+
+    return Promise.resolve([1, "test", [publishStats, subscribeStats, serviceStats]]);
```
jtbandes, line 84: is this "test" a TODO?

packages/ros1/src/RosNode.ts
```diff
@@ -0,0 +1,186 @@
...
+    // Register with each publisher
+    await Promise.all(
+      publishers.map(async (pubUrlStr) => {
+        const url = ToUrl(pubUrlStr as string);
+        if (url === undefined) {
+          return;
+        }
+
+        // Create an XMLRPC client to talk to this publisher
+        const xmlRpcClient = await this.#xmlRpcCreateClient({ url });
```
jtbandes, line 136: why does creation need to be an async call?

I wonder if it would be cleaner for RosSlaveClient to make the xmlrpc client itself, just given the url.

packages/ros1/src/TcpConnection.ts
```diff
@@ -0,0 +1,133 @@
...
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { MessageReader, parseMessageDefinition, RosMsgDefinition } from "rosbag";
+import { TextDecoder, TextEncoder } from "web-encoding";
+
+import { Connection } from "./Connection";
+import { TcpAddress, TcpSocket } from "./TcpTypes";
+
+type Header = [string, string][];
```
jtbandes, line 11: I think this would be nicer as an object/record, unless duplicate keys are needed? I also see you are using a Map elsewhere.

packages/ros1/src/TcpConnection.ts
```diff
@@ -0,0 +1,133 @@
...
+  static ParseHeader(data: Uint8Array): Map<string, string> {
+    const decoder = new TextDecoder();
+    const view = new DataView(data.buffer, data.byteOffset, data.byteLength);
+    const result = new Map<string, string>();
+
+    let idx = 0;
+    while (idx + 4 < data.length) {
+      const len = Math.min(view.getUint32(idx, true), data.length - idx - 4);
+      idx += 4;
+      const str = decoder.decode(new Uint8Array(data.buffer, data.byteOffset + idx, len)) as string;
```
jtbandes, line 120: why is `as string` needed here?

packages/ros1/src/TcpConnection.ts
```diff
@@ -0,0 +1,133 @@
...
+      let equalIdx = str.indexOf("=");
+      if (equalIdx < 0) {
+        equalIdx = str.length;
+      }
+      const key = str.substr(0, equalIdx);
+      const value = str.substr(equalIdx + 1);
+      result.set(key, value);
+      idx += len;
+    }
```
jtbandes, line 129: warn if leftover bytes?

packages/ros1/src/RosSlave.ts
jhurliman: I struggled with this. I initially had a different set of terminology and was getting lost in my translation from ROS concepts to my own. Now that the implementation is fleshed out and working I think renaming is more feasible. I'll do a pass now.

packages/ros1-nodejs/src/TcpMessageStream.ts
jhurliman: It's not my API :woman_shrugging:

packages/ros1-nodejs/src/TcpNode.ts
jhurliman: Sockets do not always have a handle, only one some platforms. It's also only documented through folklore and reading the internals of node.js source code, but this is the only way (in pure JS at least) to access the fd for a socket.

packages/ros1-nodejs/src/TcpMessageStream.ts
defunctzombie: ah yes I see that - happy times

packages/ros1/src/mock/TcpMock.ts
jhurliman: It no longer matches the interface definition so typechecking fails when used

packages/ros1-nodejs/src/PlatformNode.ts
```diff
@@ -0,0 +1,80 @@
...
+export function GetDefaultRosMasterUri(): Promise<URL> {
+  const urlStr = process.env["ROS_MASTER_URI"];
+  if (urlStr !== undefined && urlStr.length > 0) {
+    try {
+      const url = new URL(urlStr);
+      if (url.protocol === "http:" || url.protocol === "https:") {
+        return Promise.resolve(url);
+      }
+    } finally {
+      // Ignore malformed ROS_MASTER_URI env var
```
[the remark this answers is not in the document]
jhurliman: There are a handful of places in these libraries where dependency injecting a logger could be useful. I can create a followup ticket for that.

packages/ros1-nodejs/src/TcpNode.ts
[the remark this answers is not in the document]
jhurliman: `_handle?` doesn't assume `_handle` exists. This is the only way of accessing the file descriptor for a socket from pure JS, on platforms where sockets have file descriptors. This message isn't logged, it is sent over the network to ROS code which may be making terrible assumptions about the formatting of this string and `fd` being properly set on Linux. I'm trying to implement the ROS protocol as closely as possible to roscpp.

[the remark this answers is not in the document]
jtbandes: Is that even possible? What could someone on another host meaningfully do with information about which fd someone else's machine is using for their socket?

packages/ros1/src/RosNode.ts
jhurliman: Because the `node-electron` implementation will look like "send a message to a separate process requesting this thing be made, instantiate it in that separate process, return a message saying it has been instantiated with a handle referencing it"

packages/ros1/src/TcpConnection.ts
jhurliman: It would be nicer but less powerful, since you would not be able to control what the final serialization looked like. I did it this way to exactly match the roscpp serialization.

packages/ros1/src/TcpConnection.ts
jhurliman: Because `TextDecoder` is type `any` :(.

packages/ros1-nodejs/src/TcpNode.ts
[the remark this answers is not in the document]
jhurliman: It's a little unclear what the purpose of this data is at all, but I'm guessing it would be fetched and stored for some sort of logging system. Why the local socket fd matters, especially when that concept doesn't even exist on all platforms, is beyond me.

packages/ros1/src/RosSlave.ts
amacneil: Also it seems like ros themselves are moving more towards names like "roscore" over "rosmaster", although maybe they haven't fully committed to it because both seem to exist?

packages/ros1-nodejs/src/TcpNode.ts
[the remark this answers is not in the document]
jtbandes: So can we just remove this for now since we don't think it's being used for any important purpose? Would be nice to clean up the strange TS hackery unless we know it's valuable.

PR #195, opened 2021-03-09
Fix sentry sourcemap warnings in development

defunctzombie:
Use a better devTool source map which doesn't produce the warnings.


app/index.tsx
```diff
@@ -13,7 +13,7 @@ import installDevtoolsFormatters from "@foxglove-studio/app/util/installDevtools
 import overwriteFetch from "@foxglove-studio/app/util/overwriteFetch";
 import waitForFonts from "@foxglove-studio/app/util/waitForFonts";
 
-if (process.env.SENTRY_DSN) {
+if (typeof process.env.SENTRY_DSN === "string") {
```
amacneil, line 16: what if it is set to empty string?

defunctzombie: Then the init call will fail since there will be no DSN.

PR #206, opened 2021-03-10
More Storybook fixes

jtbandes:
- Increase delay yet again for State Transitions
- Exclude unintentionally exported function from new Diagnostics stories
- Set the User-Agent for screenshots so that existing inScreenshotTests() function works as expected
- Add a screenshot test for inScreenshotTests


package.json
```diff
@@ -31,8 +31,6 @@
     "package:linux": "yarn package --linux",
     "package": "NODE_OPTIONS='-r ts-node/register' electron-builder",
     "storybook": "yarn workspace @foxglove-studio/app run start-storybook",
-    "storybook:build": "yarn workspace @foxglove-studio/app run build-storybook",
-    "storybook:storycap": "yarn workspace @foxglove-studio/app run storycap http://localhost:9001 --serverCmd \"yarn http-server storybook-static -p 9001\" --outDir \"storybook-screenshots\"",
```
amacneil, line 35: had this here to make it easy for folks to run it locally

jtbandes: that wasn't working anyway.... would like to revisit and get it working

amacneil: although I guess your ts version is a little easier to read (especially diffs)

jtbandes: I think we may also want more configuration such as disabling headless mode. Maybe we can split out the storycap part into a separate script with more options

ci/storybook.ts
```diff
@@ -8,8 +8,27 @@ import exec from "./exec";
...
-await exec("yarn", ["run", "storybook:storycap"]);
+await exec("yarn", ["workspace", "@foxglove-studio/app", "run", "build-storybook"]);
+await exec("yarn", [
+  "workspace",
+  "@foxglove-studio/app",
+  "run",
+  "storycap",
+  "http://localhost:9001",
+  "--serverCmd",
+  "yarn http-server storybook-static -p 9001",
```
amacneil, line 19: while you're cleaning this up, you could add a comment above this line explaining that we use `http-server` instead of `start-storybook` so that we can more cleanly separate build errors rather than waiting several minutes for it to build inside the storycap command.

(iirc webviz uses the start-storybook command rather than separating these steps)

PR #207, opened 2021-03-10
Use esbuild-loader in-place of ts-loader

defunctzombie:
Using [esbuild](https://github.com/evanw/esbuild) via [esbuild-loader](https://github.com/privatenumber/esbuild-loader) decreases the prod and dev build time and bundle size.

| Build | esbuild | ts-loader |
| --- | --- | --- |
| build:prod | 35.2s | 43.6s |
| prod main.js | 5.94 mb | 23.2 mb |
| prod .webpack | 61mb | 64mb |
| build:dev |  23.4s | 26.3s |

There might be more gains if we also port main and preload over but renderer is the big one.

The biggest benefit of this is less the speedup but the smaller main.js size. A smaller size loads faster in electron.

I've also yet to try using esbuild just for the minification which is something we could try - tho I've not been able to get the minified builds to start typically.

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,6 +73,7 @@
     "electron-notarize": "1.0.0",
     "electron-squirrel-startup": "1.0.0",
     "electron-updater": "4.3.8",
+    "esbuild-loader": "2.9.2",
     "eslint": "7.21.0",
     "eslint-config-prettier": "8.1.0",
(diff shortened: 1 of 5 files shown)


defunctzombie:
The speed gains don't really exist - and the little that do are not worth the complexity.

PR #216, opened 2021-03-11
Enable Chromatic

amacneil:
Alternative to reg-suit, currently running in parallel.

- Chromatic job took 6m24s to build versus 9m15s for reg-suit
- Nice UI to review and specifically accept each changed story
- Able to review components (instead of just screenshots)
- I don't understand what is the difference between "UI Tests" and "UI Review" statuses that they put on Github. It looks like we can disable one of them, but need to figure out what their intention is first.

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -2,6 +2,8 @@
   "name": "@foxglove-studio/app",
   "private": true,
   "scripts": {
+    "chromatic": "chromatic --project-token $CHROMATIC_PROJECT_TOKEN --build-script-name storybook:build --exit-once-uploaded",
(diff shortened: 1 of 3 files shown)


app/package.json
```diff
@@ -132,6 +134,7 @@
     "@types/ws": "7.4.0",
     "@wojtekmaj/enzyme-adapter-react-17": "0.4.1",
     "argparse": "2.0.1",
+    "chromatic": "github:amacneil/chromatic-cli#detectyarn",
```
amacneil, line 137: https://github.com/chromaui/chromatic-cli/pull/296

PR #217, opened 2021-03-11
Add layout management

defunctzombie:
Add experimental feature for layout management. Layout management changes the "layout" menu to support:

* Adding new layouts
* Changing between layouts
* Rename layout
* Export/Import layouts to/from json files

See the Storybook LayoutsContextMenu story to play with the component.

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -58,7 +58,9 @@ rules:
   react/prop-types: off # Unnecessary with typescript validation
   react/no-unused-prop-types: error
   react-hooks/rules-of-hooks: error
-  react-hooks/exhaustive-deps: error
+  react-hooks/exhaustive-deps:
+    - error
+    - additionalHooks: "(useAsync)"
   no-new-func: error
   header/header:
     - error
--- a/app/App.tsx
+++ b/app/App.tsx
@@ -27,6 +27,7 @@ import GlobalKeyListener from "@foxglove-studio/app/components/GlobalKeyListener
 import GlobalVariablesMenu from "@foxglove-studio/app/components/GlobalVariablesMenu";
 import { WrappedIcon } from "@foxglove-studio/app/components/Icon";
 import LayoutMenu from "@foxglove-studio/app/components/LayoutMenu";
+import LayoutStorageReduxAdapter from "@foxglove-studio/app/components/LayoutStorageReduxAdapter";
 import { NativeFileMenuPlayerSelection } from "@foxglove-studio/app/components/NativeFileMenuPlayerSelection";
 import NotificationDisplay from "@foxglove-studio/app/components/NotificationDisplay";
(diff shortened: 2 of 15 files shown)


app/components/LayoutModal.tsx
```diff
@@ -1,70 +0,0 @@
...
-
-// TODO(JP): Use useSelector and useDispatch here, but unfortunately `loadLayout` needs
-// a `getState` function in addition to `dispatch`, so needs a bit of rework.
-// @ts-ignore look into errors for connect generic args
-const LayoutModal = connect<Props, OwnProps, _, _, _, _>(
-  (state: State) => ({ panels: state.persistedState.panels }),
-  { loadLayout },
-)(UnconnectedLayoutModal);
-
-export function openLayoutModal(history?: BrowserHistory) {
```
jtbandes, line 66: I think keeping this might be nice as a way to easily tweak configs for development/debugging

app/components/ReduxLayoutManager.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { v4 as uuidv4 } from "uuid";
+
+import { loadLayout } from "@foxglove-studio/app/actions/panels";
+import LayoutManagerContext from "@foxglove-studio/app/context/LayoutManagerContext";
+import { Layout, useLayoutStorage } from "@foxglove-studio/app/context/LayoutStorageContext";
+import { State } from "@foxglove-studio/app/reducers";
+import { PanelsState } from "@foxglove-studio/app/reducers/panels";
+import sendNotification from "@foxglove-studio/app/util/sendNotification";
+
+export default function ReduxLayoutManager(props: PropsWithChildren<unknown>) {
```
jtbandes, line 17: Maybe call this an "Adapter" to indicate its transitional status?

app/util/layout.ts
```diff
@@ -546,42 +532,10 @@ const layoutKeyMap = {
 };
 export const dictForPatchCompression = { ...layoutKeyMap, ...stateKeyMap };
 
-export function getUpdatedURLWithPatch(search: string, diff: string): string {
```
jtbandes, line 549: Can we move this stuff into something like a "LocationLayoutStorageProvider"?

desktop/index.ts
```diff
@@ -43,6 +44,10 @@ if (require("electron-squirrel-startup")) {
 const isMac = process.platform === "darwin";
 const isProduction = process.env.NODE_ENV === "production";
 
+ipcMain.handle("getUserDataPath", () => {
+  return app.getPath("userData");
+});
```
jtbandes, line 49: Soon we should create a typed wrapper around this so we aren't dealing with `any` so much

app/util/layout.ts
defunctzombie: I've re-worked this into LayoutStorageReduxAdapter. It is no longer a provider. It is a childless component that connects redux to a LayoutStorage.

PR #218, opened 2021-03-11
Convert StateTransitions panel to hooks

jtbandes:
Removes its use of MessageHistoryDEPRECATED.

Most of the logic from MessageHistoryDEPRECATED is moved into a new hook, `useMessagesByPath`. Some of the logic in the Plot panel is similar, but I think it's different because Plot can handle discontiguous ranges and preloading. Still, perhaps could be consolidated in the future.

### DIFF
--- a/app/PanelAPI/useDataSourceInfo.ts
+++ b/app/PanelAPI/useDataSourceInfo.ts
@@ -27,7 +27,7 @@ export type DataSourceInfo = {
   topics: ReadonlyArray<Topic>;
   datatypes: RosDatatypes;
   capabilities: string[];
-  startTime: Time | undefined; // Only `startTime`, since `endTime` can change rapidly when connected to a live system.
+  startTime?: Time; // Only `startTime`, since `endTime` can change rapidly when connected to a live system.
   playerId: string;
 };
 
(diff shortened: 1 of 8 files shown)


app/PanelAPI/useDataSourceInfo.ts
```diff
@@ -27,7 +27,7 @@ export type DataSourceInfo = {
   topics: ReadonlyArray<Topic>;
   datatypes: RosDatatypes;
   capabilities: string[];
-  startTime: Time | null | undefined; // Only `startTime`, since `endTime` can change rapidly when connected to a live system.
+  startTime: Time | undefined; // Only `startTime`, since `endTime` can change rapidly when connected to a live system.
```
defunctzombie, line 30: consider using `?:` instead of ` | undefined` if your intentention here is to allow someone making a `DataSourceInfo` to omit the `startTime` field.

https://www.typescriptlang.org/play?#code/C4TwDgpgBAYg9nKBeKBvAsAKClAzgQwFswAbCALj2ACcBLAOwHMoAfKAV3oBMIAzBiFwDcWAL5YsoSLAQB5MMFpx6+EsjRYcBYmQD8lXDQaMRmcZiwBjZYaiEQ8OJUfrUo01ZvAo+enGAAFhDUjs5yCkoqaihuQkA

Otherwise they have to specify it as `undefined` explicitly.

jtbandes: I think this type is only used as an output/return type, so the key will always be present

PR #223, opened 2021-03-11
Add fileAssociations for .bag files

defunctzombie:
This change allows users to double-click on a .bag file and have foxglove studio
launch and open their bag.


electron-builder.json
```diff
@@ -38,7 +44,30 @@
     "category": "public.app-category.developer-tools",
     "icon": "resources/icon/icon.icns",
     "entitlements": "resources/mac/entitlements.plist",
-    "entitlementsInherit": "resources/mac/entitlements.plist"
+    "entitlementsInherit": "resources/mac/entitlements.plist",
+    "extraResources": [{ "from": "resources/icon/BagIcon.png", "to": "BagIcon.png" }],
```
defunctzombie, line 48: @jtbandes do we need fileAssociations for mac?

desktop/index.ts
```diff
@@ -182,6 +189,53 @@ async function createWindow(): Promise<void> {
...
+        selector: `#${inputElementId}`,
+      });
+      await debug.sendCommand("DOM.setFileInputFiles", {
+        nodeId: queryRes.nodeId,
+        files: filesToOpen,
+      });
+
+      // clear the files once we've opened them
+      filesToOpen.splice(0, filesToOpen.length);
+    } finally {
```
amacneil, line 222: given that this is a "_hack_", it seems like you should catch any errors and report them to sentry

defunctzombie: I think they will be reported since I am not catching any exceptions here - they will either be unhandled within main or might even bubble up to user space so they know something went wrong.

electron-builder.json
```diff
@@ -9,19 +9,33 @@
   },
   "artifactName": "${name}-${version}-${os}-${arch}.${ext}",
   "afterSign": "resources/notarize.ts",
+  "icon": "resources/icon/icon.icns",
```
defunctzombie, line 12: @jtbandes fyi - I had to remove the "icon" entry in linux and use this top level icon entry with icns. The linux build wants a 32x32 icon for a png (otherwise it seemed to not be working). The linux build will generate one from the icns for us.

electron-builder.json
```diff
@@ -9,19 +9,33 @@
...
         "target": "deb",
         "arch": ["x64", "arm64"]
       }
     ],
-    "icon": "resources/icon/icon.png"
+    "fileAssociations": [
+      {
+        "ext": "bag",
+        "name": "ROS Bag File",
+        "mimeType": "application/octet-stream"
```
defunctzombie, line 24: mimeType is needed for linux - tested on my ubuntu vm and it opens bags from the file explorer

PR #233, opened 2021-03-12
Remove renderToBody

defunctzombie:


app/components/LayoutModal.tsx
```diff
@@ -55,16 +44,4 @@ function UnconnectedLayoutModal({
   );
 }
 
-// TODO(JP): Use useSelector and useDispatch here, but unfortunately `loadLayout` needs
-// a `getState` function in addition to `dispatch`, so needs a bit of rework.
```
jtbandes, line 59: weird - was this comment just outdated? I don't see a getState.

app/components/Modal.tsx
```diff
@@ -62,52 +59,35 @@ const StyledContent = styled.div`
...
-        </StyledContent>
-      </Container>
-    );
-  }
+          <CloseIcon />
+        </Icon>
+        {props.children}
+      </StyledContent>
+    </Container>,
+    document.body,
```
jtbandes, line 91: maybe need to create a new div for this purpose so we don't conflict with other things also rendering portals into body?

app/components/PanelToolbar/HelpButton.tsx
```diff
@@ -12,32 +12,31 @@
...
+export default function HelpButton(props: PropsWithChildren<Props>) {
+  const [showHelp, setShowHelp] = useState<boolean>(false);
+
+  return (
+    <Icon
+      tooltip="Help"
+      fade
+      onClick={() => {
+        setShowHelp(true);
+      }}
```
jtbandes, line 34: nit
```suggestion
      onClick={() => setShowHelp(true)}
```

app/components/PanelToolbar/index.tsx
```diff
@@ -136,98 +135,94 @@ function StandardMenuItems({ tabId, isUnknownPanel }: { tabId?: string; isUnknow
     [actions, mosaicActions, mosaicWindowActions, tabId],
   );
 
-  const onImportClick = useCallback(
-    (store, id) => {
-      if (!id) {
-        return;
-      }
-      const panelConfigById = store.getState().persistedState.panels.savedProps;
```
jtbandes, line 144: I guess this function may now update more frequently (e.g. as you drag to resize panels) since you moved savedProps to one of the useMemo deps. Might be worth checking the number of calls and getting the value from a useRef so it doesn't have to recreate the callback each time.

app/components/RenderToBodyComponent.tsx
```diff
@@ -0,0 +1,19 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+//
+// This file incorporates work covered by the following copyright and
+// permission notice:
```
jtbandes, line 6: I think at this point, it really doesn't

app/hooks/usePrompt.tsx
```diff
@@ -60,14 +60,18 @@ function ModalPrompt({ onComplete, placeholder }: ModalPromptProps) {
 
 function runPrompt(options?: PromptOptions): Promise<string | undefined> {
   return new Promise((resolve) => {
-    const modal = renderToBody(
+    const container = document.createElement("div");
+    document.body.appendChild(container);
```
jtbandes, line 64: do we wanna auto-remove this when the caller is unmounted as a follow up task?

app/players/RosbridgePlayer.ts
```diff
@@ -16,8 +16,6 @@ import { MessageReader, Time, parseMessageDefinition } from "rosbag";
 import roslib from "roslib";
 import { v4 as uuidv4 } from "uuid";
 
-import WssErrorModal from "@foxglove-studio/app/components/WssErrorModal";
```
jtbandes, line 19: This file is now sad and probably not used anywhere... can we have a follow up ticket to put it back (in a sane way rather than rendering UI from a player)?

app/players/RosbridgePlayer.ts
```diff
@@ -362,7 +356,8 @@ export default class RosbridgePlayer implements Player {
   }
 
   publish({ topic, msg }: PublishPayload): void {
-    if (!this._topicPublishers[topic]) {
+    const pubSub = this._topicSubscriptions.get(topic);
```
jtbandes, line 359: naming: `subscription`?

jtbandes (approved):

app/components/LayoutModal.tsx
defunctzombie: Not sure - but it definitely turned into a functional component without much pain.

app/components/Modal.tsx
defunctzombie: For better or worse this seems to work...

app/components/PanelToolbar/HelpButton.tsx
defunctzombie: Sometimes these don't work because it changes the return type of the function. I'll accept this suggestion and see if it builds.

app/components/PanelToolbar/index.tsx
defunctzombie: savedProps was already present above in the functional component so it would already cause renders - but I see what you mean about making the onImportClick function itself. since the store is more "stable" I'll make it the dependency to avoid making the function more times.

app/hooks/usePrompt.tsx
defunctzombie: yea we do - tho with what I've learned about portals I wonder if I can just create a portal here...

app/players/RosbridgePlayer.ts
defunctzombie: Added to internal issue tracker under an existing issue.

PR #234, opened 2021-03-12
Convert ExperimentalFeatures to a provider instead of GlobalHooks

jtbandes:
- Experimental feature configuration is now stored in `app/experimentalFeatures.ts`.
- Consolidates App and Root into one component that manages modal state.
- Also hooks up a preferences menu item to open the experimental features modal.
- Removes non-hooks imperative usage of experimental features.
- Hook up a help menu item to show keyboard shortcuts for good measure, moving the state into App instead of the history/router.
- Removes unused screenshot code/shortcuts.


.eslintrc.yaml
```diff
@@ -119,7 +119,6 @@ rules:
 
 overrides:
   - files:
-      - "app/App.tsx"
```
defunctzombie, line 122: 👏

app/App.tsx
```diff
@@ -1,35 +1,164 @@
...
+
+  .icon {
+    color: white;
+  }
+
+  // Allow interacting with buttons in the title bar without dragging the window
+  -webkit-app-region: no-drag;
+`;
+
+const connector = connect(
```
defunctzombie, line 69: Can we move this to hooks within the component?

app/components/ExperimentalFeatures.tsx
```diff
@@ -11,27 +11,27 @@
...
 import Storage from "@foxglove-studio/app/util/Storage";
 import logEvent, { getEventNames } from "@foxglove-studio/app/util/logEvent";
 
-// All these are exported for tests; please don't use them directly in your code.
+type ExperimentalFeaturesBackend = {
+  features: FeatureDescriptions;
+  settings: FeatureSettings;
+  changeFeature(id: string, value: FeatureValue): void;
+};
+const ExperimentalFeaturesContext = createContext<ExperimentalFeaturesBackend>({
```
defunctzombie, line 28: I've been putting contexts in `app/context` to keep the context separate from any concrete implementation of the context.

I've also included the `use...` function for the context in that file. This has been working decently for organizing.

app/context/ExperimentalFeaturesLocalStorageProvider.tsx
```diff
@@ -0,0 +1,66 @@
...
+          : { enabled: description[getDefaultKey()], manuallySet: false },
+      ),
+    [features, featureStorage],
+  );
+
+  const backend: ExperimentalFeaturesBackend = {
+    settings,
+    features,
+    changeFeature(id, value) {
+      const storage = new Storage();
```
defunctzombie, line 43: useLocalStorage?
https://github.com/streamich/react-use/blob/master/docs/useLocalStorage.md

I've already added the react-use package.

app/context/ExperimentalFeaturesLocalStorageProvider.tsx
```diff
@@ -0,0 +1,66 @@
...
+      const newStorage: FeatureStorage = { ...storage.getItem(EXPERIMENTAL_FEATURES_STORAGE_KEY) };
+
+      logEvent({
+        // @ts-ignore Event logging is not currently well typed
```
defunctzombie, line 47: change to @ts-expect-error so when we do add better types to logEvent then we get notified to remove the suppression by the compiler.

app/context/ExperimentalFeaturesLocalStorageProvider.tsx
```diff
@@ -0,0 +1,66 @@
...
+        name: getEventNames().CHANGE_EXPERIMENTAL_FEATURE,
+        tags: { feature: id, value },
+      });
+
+      if (value === "default") {
+        delete newStorage[id];
```
defunctzombie, line 53: should newStorage be a Map? Seems so based on usage?

app/experimentalFeatures.ts
```diff
@@ -0,0 +1,22 @@
...
+    productionDefault: false,
+  },
+  unlimitedMemoryCache: {
+    name: "Unlimited in-memory cache (requires reload)",
+    description:
+      "If you have a lot of memory in your computer, and you frequently have to play all the way through large bags, you can turn this on to fully buffer the bag into memory. However, use at your own risk, as this might crash the browser.",
+    developmentDefault: false,
+    productionDefault: false,
+  },
+} as FeatureDescriptions;
```
defunctzombie, line 22:
```
const experimentalFeatures: FeatureDescriptions = {...};
export default experimentalFeatures
```

This would give you type checking on the descriptions rather than unconditionally casing to FeatureDescriptions.

jtbandes: this also seems to give appropriate type errors if it is used incorrectly— I think I did it this way because I was doing `export default {}` and didn't have an intermediate variable

app/context/ExperimentalFeaturesLocalStorageProvider.tsx
jtbandes: This seems nice but there seems to be some logic in the Storage class that handles when localStorage gets full and automatically clear out the msg definition cache. So I would prefer not to touch this until we revisit that caching mechanism

app/context/ExperimentalFeaturesLocalStorageProvider.tsx
jtbandes: I think it hardly makes a difference here. Since it is going to/from localStorage it needs to be JSON serializable.

PR #237, opened 2021-03-13
Add value option for prompt to set initial value

defunctzombie:


app/hooks/usePrompt.test.tsx
```diff
@@ -0,0 +1,65 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { screen, fireEvent } from "@testing-library/react";
```
jtbandes, line 5: When should we use these vs ReactTestUtils or enzyme? I worry that we have too many different testing libraries available and it's not clear when to use them.

defunctzombie: Since we were already using `@testing-library/react-hooks` I figured I'd use the sibling library to fire events and query for my components. I guess I could use react-dom/test-utils directory to fire events (iirc this library uses the appropriate thing under the hood).

I don't see a comparable alternative to the screen query functions tho - I could document.querySelector instead but that was messier than using their nice higher level functions.

PR #244, opened 2021-03-14
Fix binaryObjects import cycle

defunctzombie:


app/util/binaryObjects/messageDefinitionUtils.ts
```diff
@@ -99,3 +99,22 @@ export const getDatatypes = (cls: any): [RosDatatypes, string] | null | undefine
 // player, and has gone missing in test fixture before as well. Best to just assume it doesn't
 // exist, because it's an annoying denormalization anyway.
 export const isComplex = (typeName: string) => !primitiveList.has(typeName);
+
+// True for "object bobjects" and array views.
+export const isBobject = (object: any | null | undefined): boolean => {
+  return object?.[deepParseSymbol] != null;
```
jtbandes, line 105: We should try typing deepParseSymbol as `unique symbol`, and then we can define a Bobject type as `{ [deepParseSymbol]: ... }` and this function can return `: is Bobject`.

defunctzombie: Going to save that for follow-up since it might uncover more nasties.

PR #251, opened 2021-03-14
Move ClearBagCache menu item to experimental modal

defunctzombie:


defunctzombie:
@jtbandes I need so guidance on where to put this button. To me it makes sense for it to live in the experimental features modal - since you reworked that recently maybe you have an idea of how best to introduce such a concept? All I know is that it shouldn't live in the _layout menu_.

amacneil:
Once we create a `Cmd+K` menu I think that would be a good place to store all these random features

PR #253, opened 2021-03-15
Move binaryTranslation to packages

defunctzombie:
This change moves binaryTranslation to a separate "module" under packages. A separate module helps define the boundary for modules. With this change, I've introduced a few changes to our jest setup which facilitate working with multiple projects.

A new top level _jest.config.ts_ file list _projects_ which are separate jest configurations. Running `yarn test` at the top level will run all the projects. Individual projects are run via `yarn test --project packages/bobject` for example.

I've updated the app tsconfig.json to use "module: es2020". This allows webpack to perform better tree-shaking and lowers the main.js prod file size to ~5MB down from ~25MB. The code parts are not "gone" just broken up into other files which does help with loading main.js

The updated tsconfig.json for app and bobject requires changes to ts-jest config. Since jest only understands commonjs, we use babel transforms to convert the es2020 module output of typescript compilation into a format jest understands and can run in node.


packages/bobject/src/BinaryMessageWriter.ts
```diff
@@ -14,58 +14,25 @@
...
-  }
-  return ModuleFactory({
-    locateFile: () => {
-      // get the path to the wasm file
-      // Webpack puts this file in the `dist` directory
-      return ModuleWasm;
-    },
-  });
-}
+const wasmUrl = new URL("../bin/translator.wasm", import.meta.url);
```
defunctzombie, line 20: Instead of using "loaders" this approach is more idiomatic of esm and how wasm ends up getting loaded. Webpack supports import.meta.url and will result in a proper url to the wasm file location in the bundle output directory.

In node, we need babel for jest to handle this feature.

jest.config.ts
```diff
@@ -0,0 +1,10 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+// Or top level jest config to reference all the projects under our monorepo
+export default {
+  // We exclude the integration test since it involves performing a webpack build and
+  // doesn't support "watch" or "debug" in the same way
+  projects: ["<rootDir>/app/jest.config.ts", "<rootDir>/packages/*/jest.config.ts"],
```
amacneil, line 9: so like, should `app` and `desktop` just live under `packages/` now? why are they special?

defunctzombie: Its up to us - I consider "packages" more like "potential yarn/npm modules" - rather than dumping everything into a utils folder, some things are large enough and separate enough that they could be "packages". It helps isolate iterating on them. app and desktop are special since _studio_ is the entire app. Think of packages as a way to organize larger module chunks.

packages/bobject/src/BinaryMessageWriter.ts
jtbandes: This depends on the webpack output path for the wasm file being correct, right? If it was output to e.g. `/translator[hash].wasm` without the `bin/` then this wouldn't work?

packages/bobject/src/index.test.ts
```diff
@@ -203,7 +203,7 @@ describe("definitions", () => {
         "fake_msgs/HasByteArray",
         definitions["fake_msgs/HasByteArray"],
       );
-      expect(definition).not.toBeNullOrUndefined();
```
jtbandes, line 206: is this because they are moved into a separate package? Would it make sense to move our jest config into another package that can be shared?

package.json
```diff
@@ -23,10 +23,10 @@
     "serve": "webpack serve --mode development --progress",
     "lint": "TIMING=1 eslint --fix .",
     "lint:nofix": "TIMING=1 eslint .",
-    "test": "yarn workspace @foxglove-studio/app test",
-    "test:debug": "yarn workspace @foxglove-studio/app test:debug",
-    "test:watch": "yarn workspace @foxglove-studio/app test:watch",
-    "test:integration": "jest --config jest.spectron.ts --rootDir test",
+    "test": "NODE_OPTIONS='-r ts-node/register' jest",
+    "test:debug": "NODE_OPTIONS='-r ts-node/register' --inspect-brk' jest",
```
jtbandes, line 27: this looks wrong, I think you have an extra `'`

app/test/setupTestFramework.ts
```diff
@@ -16,15 +16,15 @@ import { isEqual } from "lodash";
 
 import {
   mockSendNotification,
-  setNotificationHandler,
+  setNotificationHandler as mockSetNotificationHandler,
```
jtbandes, line 19: maybe just rename the export?

app/package.json
```diff
@@ -3,10 +3,7 @@
   "private": true,
   "scripts": {
     "chromatic": "chromatic --project-token $CHROMATIC_PROJECT_TOKEN --build-script-name storybook:build --exit-once-uploaded",
-    "storybook:build": "build-storybook",
-    "test": "NODE_OPTIONS='-r ts-node/register' jest",
-    "test:debug": "NODE_OPTIONS='-r ts-node/register --inspect-brk' jest",
-    "test:watch": "NODE_OPTIONS='-r ts-node/register' jest --watch"
+    "storybook:build": "TS_NODE_COMPILER_OPTIONS='{\"module\":\"commonjs\"}' build-storybook"
```
jtbandes, line 6: why wasn't this necessary before?

packages/bobject/.eslintrc.yaml
```diff
@@ -0,0 +1,2 @@
+parserOptions:
+  project: packages/bobject/tsconfig.json
```
jtbandes, line 2: naming: maybe call it `binary-objects`? since the source directory was binaryObjects

packages/bobject/src/BinaryMessageWriter.ts
defunctzombie: It would - webpack understands this whole URL string thing it seems and does the right thing. I'm gonna do some more tests to confirm what I've seen initially tho.

Even in the dev build I saw webpack rename the .wasm file. You can see what it actually loads in the _Network_ tab of devtools.

packages/bobject/src/index.test.ts
defunctzombie: It could - in this particular case I thought `isTruthy()` was adequate but maybe we should make a package for this jest matcher if we are to use it a lot.

app/package.json
defunctzombie: The `tsconfig.json` in /app used `module: commonjs`.

packages/bobject/.eslintrc.yaml
defunctzombie: As I read it more this particular code is more like "`ros-binary-translation` or some such.

PR #255, opened 2021-03-15
Add storage interface to OsContext

defunctzombie:
Expose storage as a LocalFileStorage instance to read/write local files.

### DIFF
--- a/preload/LocalFileStorage.ts
+++ b/preload/LocalFileStorage.ts
@@ -1,0 +1,97 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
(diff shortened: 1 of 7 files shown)


preload/LocalFileStorage.ts
```diff
@@ -0,0 +1,97 @@
...
+import path from "path";
+
+import type { Storage, StorageContent } from "@foxglove-studio/app/OsContext";
+
+export default class LocalFileStorage implements Storage {
+  #userDataPath = ipcRenderer.invoke("getUserDataPath");
+
+  async list(datastore: string): Promise<string[]> {
+    const datastoreDir = await this.ensureDatastorePath(datastore);
+    const result = new Array<string>();
```
jtbandes, line 16: I personally like `const result: string[] = []` 🤷

defunctzombie: /shrug

Will consider in follow-up. Maybe I'll like it better too.

PR #260, opened 2021-03-15
Move some package-specific tsconfig settings into base tsconfig, fix lots of unused React imports

jhurliman:


tsconfig.json
```diff
@@ -5,13 +5,19 @@
...
+    "downlevelIteration": true,
+    "importHelpers": true,
     "sourceMap": true,
     "moduleResolution": "node",
     "forceConsistentCasingInFileNames": true,
     "resolveJsonModule": true,
     "strict": true,
+    "noFallthroughCasesInSwitch": true,
+    "noImplicitAny": true,
+    "noImplicitReturns": false,
```
jhurliman, line 17: I can't enable this globally yet, the app makes heavy use of implicit returns

tsconfig.json
```diff
@@ -5,13 +5,19 @@
...
     "noUncheckedIndexedAccess": true,
+    "noUnusedLocals": true,
+    "noUnusedParameters": false,
```
jhurliman, line 20: We currently use the `after-used` setting for unused parameters which is much more lenient than `noUnusedParameters`. To fully migrate over we'd have to prepend `_` to all of the unused parameters in the codebase.

PR #265, opened 2021-03-16
Change eslint import/order to put @foxglove before relative imports

defunctzombie:
Stylistic organization to put @foxglove imports before relative path imports. @foxglove imports are between "external" and "local" since they could one day reference published packages. I find this import organization helps me when reading which imports are where.


amacneil:
Can you add PR description explaining why

PR #272, opened 2021-03-16
ros1-electron

jhurliman:


packages/ros1-electron/src/RpcMain.ts
```diff
@@ -0,0 +1,122 @@
...
+
+  handleMethod<K extends keyof RpcRendererMethodMap, V extends RpcRendererMethodMap[K]>(
+    method: K,
+    handler: (args: V["args"]) => Promise<V["out"]>,
+  ): void {
+    this.#serverHandlers.set(method, handler as RpcHandler);
+  }
+
+  static Create(channel: string, browser: BrowserWindow): RpcMain {
+    const a = new MessageChannelMain(); // renderer client, main server
```
defunctzombie, line 111: Lets give these variables a name thats more descriptive.

packages/ros1-electron/src/RpcRenderer.ts
```diff
@@ -0,0 +1,132 @@
...
+    this.#clientPort = clientPort;
+    this.#serverPort = serverPort;
+    this.#eventPort = eventPort;
+
+    // Listen for responses to RPC calls
+    this.#clientPort.addEventListener("message", (ev: MessageEvent<RpcResponsePayload>) => {
+      const [waitHandle, errMsg, out] = ev.data;
+      const callback = this.#callbacks.get(waitHandle);
+      if (!callback) {
+        // TODO: Log a warning
```
defunctzombie, line 36: Prefer `// fixme` rather than TODO since that will fail lint and make you fix it before merge.

jhurliman: Logging is a whole separate endeavor, I don't plan on tackling it this PR.

packages/ros1-electron/src/TcpSocketRenderer.ts
```diff
@@ -0,0 +1,78 @@
...
+  close(): void {
+    this._setClosed();
+    this.#rpc.call("TcpSocket_close", this.#socketId);
+  }
+
+  write(data: Uint8Array): Promise<void> {
+    return this.#rpc.call("TcpSocket_write", [this.#socketId, data]);
+  }
+
+  private _setClosed(): void {
```
jtbandes, line 73:
```suggestion
  #setClosed(): void {
```

packages/ros1-electron/src/RpcMain.ts
```diff
@@ -0,0 +1,146 @@
...
+  RpcCallPayload,
+  RpcEventPayload,
+  RpcHandler,
+  RpcMainEventMap,
+  RpcMainMethodMap,
+  RpcRendererMethodMap,
+  RpcResponsePayload,
+} from "./RpcTypes";
+
+export class RpcMain {
```
jtbandes, line 19: A few of these methods look substantially identical to RpcRenderer, can we share the implementations somehow?

packages/ros1-electron/src/RpcTypes.ts
```diff
@@ -0,0 +1,95 @@
...
+
+export type RpcValue =
+  | string
+  | number
+  | boolean
+  | Uint8Array
+  | RpcValue[]
+  | { [key: string]: RpcValue }
+  | undefined
+  | void;
```
jtbandes, line 15: Probably don't need void in addition to undefined

packages/ros1-electron/src/RpcTypes.ts
```diff
@@ -0,0 +1,95 @@
...
+  socketId: number;
+  remoteAddress: TcpAddress;
+  localAddress: TcpAddress;
+  fd?: number;
+};
+
+export interface RpcCall<Args extends RpcValue, Return extends RpcValue> {
+  args: Args;
+  out: Return;
+}
```
jtbandes, line 41: can we just use function types instead of defining what is basically a custom function type?

packages/ros1-electron/src/RpcTypes.ts
```diff
@@ -0,0 +1,95 @@
...
+    { serverId: number; address: HttpAddress }
+  >;
+  XmlRpcServer_close: RpcCall<number, void>;
+  XmlRpcServer_addMethod: RpcCall<{ serverId: number; method: string }, void>;
+}
+
+// RPC methods callable from the main process and received by the renderer
+export interface RpcMainMethodMap {
+  XmlRpcServer_onMethodCall: RpcCall<
+    { serverId: number; method: string; args: XmlRpcValue[] },
```
jtbandes, line 76: I'm a little confused by asymmetry here where it seems like serverId is part of the message being sent, but in the TCP events it is a separate argument such as `.on("TcpServer_onConnection", this.#serverId` ... why is the server id not part of the event itself in these cases?

PR #271, opened 2021-03-16
Move slow lint checks to CI-only

amacneil:
before

```
Rule                           | Time (ms) | Relative
:------------------------------|----------:|--------:
import/no-self-import          | 24012.591 |    15.4%
prettier/prettier              | 23319.138 |    15.0%
react/no-direct-mutation-state | 15868.722 |    10.2%
react/prop-types               | 14142.034 |     9.1%
react/no-deprecated            | 13845.635 |     8.9%
react/display-name             | 13580.431 |     8.7%
react/no-string-refs           | 13509.158 |     8.7%
react/require-render-return    | 13492.359 |     8.7%
react/no-unused-prop-types     | 13315.296 |     8.6%
✔ Lint done...
yarn lint  250.16s user 14.20s system 130% cpu 3:23.06 total
```

after

```
Rule                            | Time (ms) | Relative
:-------------------------------|----------:|--------:
import/order                    | 16906.047 |    15.2%
react/no-direct-mutation-state  | 13731.133 |    12.4%
react/prop-types                | 11963.317 |    10.8%
react/no-deprecated             | 11807.723 |    10.6%
react/display-name              | 11604.820 |    10.4%
react/no-string-refs            | 11533.362 |    10.4%
react/no-unused-prop-types      | 11512.345 |    10.4%
react/require-render-return     | 11503.888 |    10.4%
import/no-useless-path-segments |  2209.603 |     2.0%
✔ Lint done...
yarn lint  168.43s user 11.71s system 118% cpu 2:32.32 total
```

PR #273, opened 2021-03-16
Fix and enable `react/no-unused-prop-types` errors

amacneil:
Follow up work from https://github.com/foxglove/studio/pull/271


app/panels/NodePlayground/Sidebar.tsx
```diff
@@ -106,6 +106,7 @@ type NodesListProps = {
   deleteNode: (id: string) => void;
   collapse: () => void;
   selectedNodeId?: string;
+  // eslint-disable-next-line react/no-unused-prop-types
```
defunctzombie, line 109: delete userNodeDiagnostics property altogether?

PR #280, opened 2021-03-17
Fix panel settings json modal closing when clicked

jtbandes:
The modal was closing whenever its parent ChildToggle closed, since the RenderToBody was being hosted inside the menu (as of #233). Hoist the modal to the top level PanelToolbar so it is not removed when the menu closes.

### DIFF
--- a/app/components/PanelToolbar/index.tsx
+++ b/app/components/PanelToolbar/index.tsx
@@ -68,11 +68,11 @@ type Props = {
 function StandardMenuItems({
   tabId,
   isUnknownPanel,
-  setShowShareModal,
+  onEditPanelConfig,
 }: {
   tabId?: string;
   isUnknownPanel: boolean;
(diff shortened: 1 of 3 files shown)


app/components/PanelToolbar/index.tsx
```diff
@@ -333,6 +347,7 @@ export default React.memo<Props>(function PanelToolbar({
                 onDragStart={onDragStart}
                 onDragEnd={onDragEnd}
                 isUnknownPanel={!!isUnknownPanel}
+                setShowShareModal={setShowShareModal}
```
defunctzombie, line 350: I'd lean towards `showShareModalAction` to indicate that a user invoked the `showShareModalAction` on the PanelToolbar. How you make that (via setShowShareModal) is an implementation detail for the parent.

defunctzombie:
Small wording suggestion - LGTM

PR #286, opened 2021-03-18
Rename "Webviz Internals" panel to "Studio Internals"

defunctzombie:


defunctzombie:
@amacneil Probably makes sense to move this panel to the "debugging" section rather than the "utilities" section?

amacneil:
Yes, makes more sense under debugging

PR #288, opened 2021-03-18
Change default layout to 3d panel and 1x playback

defunctzombie:


defunctzombie:
On hold until https://github.com/foxglove/studio/pull/295 lands

amacneil:
@defunctzombie stale? or is there anything here we want to keep?

PR #292, opened 2021-03-18
electron-socket package for exposing TCP sockets to the renderer

jhurliman:


packages/electron-socket/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+initElectronSocket();
+```
+
+```
+<!-- index.html ------------------------------------------------------------->
+<script>
+import { initRendererSocket, createServer, createSocket } from "@foxglove/electron-socket/renderer";
+
+initRendererSocket();
```
defunctzombie, line 23: Does it make sense to have this return an instance that you then call "createServer", "createSocket" on? It would allow us to move away from createServer/createSocket globals. Is that useful to contain into an instance?

packages/electron-socket/src/electron/TcpServerElectron.ts
```diff
@@ -0,0 +1,103 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
defunctzombie, line 1: What does the `electron` folder mean here? The whole package is `electron`?

packages/electron-socket/preloader/index.ts
```diff
@@ -0,0 +1,105 @@
...
+import { TcpServerElectron } from "./TcpServerElectron";
+import { TcpSocketElectron } from "./TcpSocketElectron";
+import { getTransform, nextId, registerEntity } from "./registry";
+
+export { registerTransform } from "./registry";
+
+// A map of created `PreloaderSockets` instances
+const registeredSockets = new Map<string, PreloaderSockets>();
+
+export class PreloaderSockets {
```
defunctzombie, line 17: classes should live in their own files. Lets keep index for re-exporting.

packages/electron-socket/renderer/index.ts
```diff
@@ -0,0 +1,102 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { Cloneable, RpcCall, RpcResponse } from "../shared/Rpc";
+import { TcpServerRenderer } from "./TcpServerRenderer";
+import { TcpSocketRenderer } from "./TcpSocketRenderer";
+
+// A map of created `Sockets` instances, or a promise if creation is in progress
+const registeredSockets = new Map<string, Sockets | Promise<Sockets>>();
+
+export class Sockets {
```
defunctzombie, line 12: classes should be in their own files

PR #293, opened 2021-03-18
Add AppConfigurationContext

jtbandes:
Backed by [appDataDir]/settings/settings.json.

Not used yet in this PR – will be used for onboarding/welcome dialog state.

### DIFF
--- a/app/OsContext.ts
+++ b/app/OsContext.ts
@@ -17,7 +17,12 @@ export interface Storage {
   // get all the items in the datastore
   all(datastore: string): Promise<StorageContent[]>;
   // get a single item from the datastore
-  get(datastore: string, key: string): Promise<StorageContent | undefined>;
+  get(
+    datastore: string,
+    key: string,
+    options?: { encoding: undefined },
+  ): Promise<Uint8Array | undefined>;
+  get(datastore: string, key: string, options: { encoding: "utf8" }): Promise<string | undefined>;
   // put a single item into the datastore
   // This will replace any existing item with the same key
   put(datastore: string, key: string, value: StorageContent): Promise<void>;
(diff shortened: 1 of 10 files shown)


app/services/OsContextAppConfiguration.ts
```diff
@@ -0,0 +1,48 @@
...
+      .then((value) => JSON.parse(value ?? "{}"));
+  }
+
+  async get(key: string): Promise<unknown> {
+    return ((await this.#currentPromise) as Record<string, unknown>)[key];
+  }
+
+  async set(key: string, value: unknown): Promise<void> {
+    // Immediately set a new currentPromise so any future calls will wait for this one to complete.
+    this.#currentPromise = this.#currentPromise.finally(async () => {
```
jtbandes, line 29: This is my sketchy attempt at serializing potential overlapping `set` calls to avoid races. Another option is just to ignore the possibility of overlapping calls. Let me know what you think.

This `#currentPromise` also helps make `get` faster since it doesn't have to do `storage.get` (and read from disk) each time.

app/OsContext.ts
```diff
@@ -17,7 +17,12 @@ export interface Storage {
   // get all the items in the datastore
   all(datastore: string): Promise<StorageContent[]>;
   // get a single item from the datastore
-  get(datastore: string, key: string): Promise<StorageContent | undefined>;
+  get(
+    datastore: string,
+    key: string,
+    options?: { encoding: undefined },
```
defunctzombie, line 23: Should this limit the universe of allowed values?

defunctzombie: Oh I see - it will via the stuff below. Maybe `never` rather than `undefined` ?

jtbandes: I guess that would mean nobody can explicitly pass `{encoding: undefined}`?

PR #295, opened 2021-03-18
Welcome layout

jtbandes:
- Add a Welcome panel (that is not visible in the app menu).
- When the app is loaded, show the welcome layout & load the demo bag if it has not been shown before.
- Expose `setPlayerFromBagURL` in the player selection context, so the app can load the demo bag. Previously, this was done in a useEffect in the PlayerManager: https://github.com/foxglove/studio/blob/a99dbd613cdba5a6a05e55d22aa0b028c892ad49/app/components/PlayerManager.tsx#L271-L272

The demo bag is a filtered version of `NOISY_cut_off_2.bag` from [Udacity/Didi challenge dataset 1](https://github.com/udacity/didi-competition) and was created by running:
```
roscore
rosparam set use_sim_time true
rosrun image_proc image_proc
rosbag record -O out.bag  /diagnostics /image_color/compressed /radar/points /radar/tracks /radar/range /rosout /tf /velodyne_points
rosbag play --clock -s 9 NOISY_cut_off_2.bag
rosbag filter out.bag filtered-out.bag "(t.secs,t.nsecs)>(1490149580,100313663)"
```


app/onboarding/WelcomePanel.tsx
```diff
@@ -0,0 +1,95 @@
...
+const STerm = styled.span`
+  color: ${colors.accent};
+  font-weight: bold;
+`;
+
+function validateEmail(str: string): string | undefined {
+  if (str.length === 0) {
+    return undefined; // don't show an error when nothing has been entered
+  }
+  if (/^.+@.+\..+$/.test(str)) {
```
defunctzombie, line 26: Please comment crazy regex.

app/onboarding/WelcomePanel.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+import Button from "@foxglove-studio/app/components/Button";
+import Checkbox from "@foxglove-studio/app/components/Checkbox";
+import Icon from "@foxglove-studio/app/components/Icon";
+import Panel from "@foxglove-studio/app/components/Panel";
+import TextContent from "@foxglove-studio/app/components/TextContent";
+import TextField from "@foxglove-studio/app/components/TextField";
+import colors from "@foxglove-studio/app/styles/colors.module.scss";
+
+const STerm = styled.span`
```
defunctzombie, line 17: Drop the `S` prefix. I think this is not a useful indicator. Either call it `Term` or something more descriptive.

app/onboarding/welcomeLayout.ts
```diff
@@ -0,0 +1,130 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
defunctzombie, line 1: Do we need an onboarding folder?

app/onboarding/WelcomePanel.tsx
```diff
@@ -0,0 +1,95 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
defunctzombie, line 1: This should live in `app/panels` with all the other panels no?

app/panels/WelcomePanel.tsx
```diff
@@ -0,0 +1,137 @@
...
+          checked={slackInviteChecked}
+          onChange={setSlackInviteChecked}
+        />
+        <Button
+          isPrimary={subscribedState.value === false}
+          disabled={!submitEnabled}
+          onClick={submit}
+        >
+          {loading ? "Loading…" : "Sign Up"}
+        </Button>{" "}
```
defunctzombie, line 121: Did you mean to have the {" "} here? Maybe handle the spacing with styles?

desktop/index.ts
```diff
@@ -210,6 +211,11 @@ async function createWindow(): Promise<void> {
       } else {
         // "None"
       }
+    } else if (
+      channel === "onboarding.join-slack-clicked" &&
+      process.env.SLACK_INVITE_URL != undefined
+    ) {
+      shell.openExternal(process.env.SLACK_INVITE_URL);
```
defunctzombie, line 218: you might be able to avoid having this custom event if you look at:
https://stackoverflow.com/questions/32402327/how-can-i-force-external-links-from-browser-window-to-open-in-a-default-browser

Which will open urls in browser when _blank target is used.

webpack.renderer.config.ts
```diff
@@ -184,6 +184,7 @@ export function makeConfig(_: unknown, argv: WebpackArgv): Configuration {
       }),
       new EnvironmentPlugin({
         SENTRY_DSN: process.env.SENTRY_DSN ?? null, // eslint-disable-line no-restricted-syntax
+        SIGNUP_API_URL: "https://foxglove.dev/api/signup",
```
defunctzombie, line 187: We should scope this to prod builds and put the env var in our release process. I'd like to avoid other folks building the app and getting this injected by default.

webpack.main.config.ts
```diff
@@ -74,6 +74,7 @@ export default (_: unknown, argv: WebpackArgv): Configuration => {
         MAIN_WINDOW_WEBPACK_ENTRY: rendererEntry,
         // Should match webpack-defines.d.ts
         APP_NAME: JSON.stringify("Foxglove Studio"),
+        SLACK_INVITE_URL: "https://foxglove.dev/join-slack",
```
defunctzombie, line 77: What do we put in EnvironmentPlugin and what do we put into define plugin?

app/components/PlayerManager.tsx
```diff
@@ -276,6 +278,8 @@ function PlayerManager({
 
   const value: PlayerSelection = {
     selectSource,
+    setPlayerFromDemoBag: async () =>
```
defunctzombie, line 281: Add a comment here with some context on how this came to be and what it supports so we know what to do when we revisit later.

app/components/PlayerManager.tsx
```diff
@@ -49,6 +49,8 @@ import { State } from "@foxglove-studio/app/reducers";
 import { SECOND_SOURCE_PREFIX } from "@foxglove-studio/app/util/globalConstants";
 import { useShallowMemo } from "@foxglove-studio/app/util/hooks";
 
+export const DEMO_BAG_URL = "https://open-source-webviz-ui.s3.amazonaws.com/demo.bag";
```
defunctzombie, line 52: Maybe we should create our own bag?

app/App.tsx
```diff
@@ -102,6 +104,20 @@ function Root() {
...
   }, [isFullScreen]);
 
+  const appConfiguration = useAppConfiguration();
+
+  // Show welcome layout on first run
+  useEffect(() => {
+    (async () => {
+      const welcomeLayoutShown = await appConfiguration.get("onboarding.welcome-layout.shown");
+      if (!welcomeLayoutShown) {
+        dispatch(loadLayout(welcomeLayout));
```
defunctzombie, line 114: This needs gating with `isMounted()` - your component may have unmounted by the time your async call finished.

desktop/index.ts
jtbandes: I think @amacneil set up some help pages to intentionally open in-app when you click on a msg type name though?

webpack.main.config.ts
jtbandes: this was a mistake, good catch. However I don't have a good answer for this question - I guess I see EnvironmentPlugin as being used for things that are "configuration" values, but don't have a good definition of that

app/panels/WelcomePanel.tsx
jtbandes: intentional, leaving a space between button and adjacent span. Without a space the browser thinks these are part of the same word:
image

defunctzombie: I'd handle this with styling - To me the space is easy to mistake as an artifact of "auto-formatting" trying to be clever. I've had it insert space placeholders like that before.

desktop/index.ts
defunctzombie: Other way around - We have a ticket to avoid making such things open in the app since that gets leveraged into a full browser.

app/panels/WelcomePanel.tsx
jtbandes: I could add styling but on principle don't think it's correct to have the elements be adjacent with no space. A screen reader could read this as "Sign UpThanks for"

PR #307, opened 2021-03-18
Fix app.dock.setIcon() exception on platforms that do not have app.dock

jhurliman:


desktop/index.ts
```diff
@@ -298,8 +298,10 @@ app.on("ready", async () => {
 
     // In development, we run with the pre-packaged Electron binary, so we need to manually set the Dock icon.
     try {
-      // This fails when opening the app from a packaged DMG.
-      app.dock.setIcon("resources/icon/icon.png");
+      if (app.dock != undefined) {
```
defunctzombie, line 301: Bad typescript typing info if app.dock is sometimes undefined?

jtbandes: `app.dock?.setIcon(...)`

jtbandes: Also wait - what is going on here? We are already inside a `try` ....

jhurliman: or-undefined

:sigh:

PR #308, opened 2021-03-18
Remove reg-suit

amacneil:


ci/storybook.ts
```diff
@@ -1,116 +0,0 @@
-// This Source Code Form is subject to the terms of the Mozilla Public
```
jtbandes, line 1: R.I.P.

You wanna remove `exec.ts` as well, and those dependencies on `@actions/...`?

amacneil: What do you think? It might come in handy in the future. But I guess we can always resurrect it.

PR #328, opened 2021-03-20
ROS native player

jhurliman:
ros1-connector


app/components/PlayerManager.tsx
```diff
@@ -274,9 +292,14 @@ function PlayerManager({
     buildPlayer(buildPlayerFromFiles(usedFiles.current, buildPlayerOptions));
   }, [buildPlayer, filesToOpen, buildPlayerOptions]);
 
+  let availableSources = playerSources;
+  if (!useExperimentalFeature("ros1Native")) {
+    availableSources = availableSources.filter((src) => src.type !== "ros1-core");
```
defunctzombie, line 297: We can avoid providing the source in `playerSources` from the property rather than filtering here.

app/util/url.ts
```diff
@@ -0,0 +1,20 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import URL from "url-parse";
+
+export function parseInputUrl(str?: string): string | undefined {
```
defunctzombie, line 7: This needs to be documented clearly and assumptions outlined. It is far too generic of a filename and function to go undocumented.

app/util/url.ts
```diff
@@ -0,0 +1,20 @@
...
+  if (str == undefined || str.length === 0) {
+    return undefined;
+  }
+  if (str.indexOf("://") === -1) {
+    str = `http://${str}`;
```
defunctzombie, line 12: Why does it assume http?

app/util/url.ts
```diff
@@ -0,0 +1,20 @@
...
+  }
+  try {
+    const url = new URL(str);
+    return url.toString();
+  } catch (e) {
```
defunctzombie, line 17: Document why you ignore the error.

desktop/index.ts
```diff
@@ -78,6 +78,7 @@ async function createWindow(): Promise<void> {
       contextIsolation: true,
       preload: preloadPath,
       nodeIntegration: false,
+      webSecurity: process.env.NODE_ENV === "production",
```
defunctzombie, line 81: Document

package.json
```diff
@@ -101,6 +102,7 @@
     "ts-node": "9.1.1",
     "typescript": "4.2.2",
     "typescript-plugin-css-modules": "3.2.0",
+    "url-parse": "1.5.1",
```
defunctzombie, line 105: Don't we already have the whatwg-url module?

preload/index.ts
```diff
@@ -20,6 +25,9 @@ const menuClickListeners = new Map<string, IpcListener>();
 // Initialize the RPC channel for electron-socket
 PreloaderSockets.Create();
 
+// Register the native stream transform for TCPROS messages
+registerTransform("RosTcpMessageStream", () => new RosTcpMessageStream());
```
defunctzombie, line 29: I'd like to keep preload free of "ros-isms" if we can. The idea behind providing a socket to renderer was to allow it to handle such things.  Does this need to be in preload?

packages/ros1/src/nodejs/TcpSocketNode.ts
```diff
@@ -7,7 +7,7 @@ import net from "net";
 
 import { TcpAddress, TcpSocket } from "@foxglove/ros1";
 
-import { TcpMessageStream } from "./TcpMessageStream";
+import { RosTcpMessageStream } from "./RosTcpMessageStream";
```
defunctzombie, line 10: This file imports from "net" - when we originally spoke about `ros1` vs `ros1-node` we discussed that `ros1` would not use "node" modules and would only specify the interfaces other things should implement. Lets avoid falling back into that trap and keep concrete transport implementations separate from interface declarations.

packages/ros1/src/TcpConnection.test.ts
```diff
@@ -0,0 +1,105 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import net from "net";
```
defunctzombie, line 5: You should not need `net` here. The idea for `ros1` was to avoid any transport layer logic. You should be able to make your own interfaces and provide those rather than require `net`.

defunctzombie:
I'd like to go over the ros1/ros1-node and preloader aspects of this PR before it merges.

app/components/PlayerManager.tsx
jhurliman: You can’t. The experimental store doesn’t exist at that point.

package.json
jhurliman: No, and it doesn’t work across both node and browser

preload/index.ts
jhurliman: Yes this needs to be in preload. It’s a native stream transformer

packages/ros1/src/nodejs/TcpSocketNode.ts
jhurliman: This is native node.js code that is specific to ROS

packages/ros1/src/TcpConnection.test.ts
jhurliman: This is a test. It has to provide a concrete implementation in order to execute

app/components/PlayerManager.tsx
defunctzombie: Ah - yes I see looking at the code. Maybe the move is to put those initial sources under the experimental provider in the react tree. Then we can start to gate even that with experimental flags. Right now is a tad confusing to see us make a list of all the sources for the app and then hack it up here - feels like we could have avoided making the source in the first place.

package.json
defunctzombie: nothing can ever be easy

preload/index.ts
defunctzombie: I don't follow what a "native stream transformer" is. Transform streams are a concept on top of any stream interface.

packages/ros1/src/nodejs/TcpSocketNode.ts
defunctzombie: Is it native node.js code or is it platform agnostic "stream" code? If its platform agnostic and works on an underlying stream interface then I can see it living in ros1 and the user of ros1 is responsible for providing a viable _stream_.

packages/ros1/src/nodejs/platform.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export function getHostname(): string | undefined {
+  return os.hostname();
+}
+
+export function getNetworkInterfaces(): NetworkInterface[] {
+  const output: NetworkInterface[] = [];
+  const ifaces = os.networkInterfaces();
+  for (const name in ifaces) {
+    const iface = ifaces[name];
```
jtbandes, line 25: consider
```suggestion
  for (const [name, addresses] of Object.entries(os.networkInterfaces())) {
```

app/OsContext.ts
```diff
@@ -45,6 +58,13 @@ export interface OsContext {
   menuAddInputSource(name: string, handler: () => void): Promise<void>;
   menuRemoveInputSource(name: string): Promise<void>;
 
+  // Retrieve an environment variable
+  getEnvVar(envVar: string): string | undefined;
```
jtbandes, line 62: security-wise, it's probably better to just provide a `getRosMasterUri()`

packages/ros1/src/RosNode.ts
```diff
@@ -253,4 +241,61 @@ export class RosNode {
       }),
     );
   };
+
+  static GetRosHostname(
```
jtbandes, line 245: @defunctzombie do you have a preferred naming convention for static methods? I'm used to seeing lowercase in JS code, but we don't have many static methods in the code so we could change it — just think we should be consistent

packages/ros1/src/RosMasterClient.test.ts
```diff
@@ -2,10 +2,62 @@
...
+          "</data></array></value>" +
+          "<value><array><data>" +
+          "<value><string>/rosout_agg</string></value>" +
+          "<value><string>rosgraph_msgs/Log</string></value>" +
+          "</data></array></value>" +
+          "</data></array></value>" +
+          "</data></array></value>" +
+          "</param>" +
+          "</params>" +
+          "</methodResponse>";
```
jtbandes, line 44: Use backticks for a multiline string. Proper formatting of the XML would help readability

app/components/MessagePathSyntax/messagePathsForDatatype.ts
```diff
@@ -173,10 +173,9 @@ export function messagePathsForDatatype(
     }
   }
   const structureItem = messagePathStructures(datatypes)[datatype];
-  if (!structureItem) {
-    throw new Error(`datatype not found "${datatype}"`);
+  if (structureItem != undefined) {
+    traverse(structureItem, "");
   }
-  traverse(structureItem, "");
```
jtbandes, line 179: What is going on here? This seems like it could have unintended consequences and I'm surprised there are no tests that need to be updated for this...

app/components/PlayerManager.tsx
```diff
@@ -59,7 +62,7 @@ type BuiltPlayer = {
 function buildPlayerFromFiles(files: File[], options: BuildPlayerOptions): BuiltPlayer {
   if (files.length === 1) {
     return {
-      player: buildPlayerFromDescriptor(getLocalBagDescriptor(files[0]!), options),
+      player: buildPlayerFromDescriptor(getLocalBagDescriptor(files[0] as File), options),
```
jtbandes, line 65: why are these changing? Is it to avoid the warnings on `!`? I think using `as` is strictly less safe, so we should prefer to change our warning settings or fix the root of the problem -- Roman may have some advice as he's been doing a bunch of uncheckedIndexAccess fixes lately.

app/util/url.ts
```diff
@@ -0,0 +1,23 @@
...
+
+import URL from "url-parse";
+
+// Parse a user-input string as a URL using a forgiving parser that interprets
+// bare server names or server/port combos using an assumed default protocol
+export function parseInputUrl(str?: string, defaultProtocol = "https:"): string | undefined {
+  if (str == undefined || str.length === 0) {
+    return undefined;
+  }
+  if (str.indexOf("://") === -1) {
```
jtbandes, line 13:
```suggestion
  if (!str.includes("://")) {
```

app/util/url.ts
```diff
@@ -0,0 +1,23 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import URL from "url-parse";
```
jtbandes, line 5: We shouldn't need this; URL is a built-in type.

app/experimentalFeatures.tsx
```diff
@@ -26,6 +26,18 @@ export default {
     developmentDefault: false,
     productionDefault: false,
   },
+  ros1Native: {
+    name: "ROS1 Native Connector",
+    description: (
+      <>
+        Connect directly to a running ROS1 robot. The rosmaster instance and ROS nodes must be
+        addressable from this computer. If you need to specify a network interface to bind to, set
+        the ROS_HOSTNAME environment variable before launching Foxglove Studio.
```
jtbandes, line 35: - "ROS1" -> "ROS 1"
- "addressable" -> "reachable"?
- Mention `ROS_IP`?
- Link to documentation? Personally I don't understand when  you would want to set ROS_HOSTNAME
- Setting an env var before launching the program seems like an archaic way of using a GUI app — also I'm not even sure nowadays how to do it on macOS and I bet most users won't know how. We should explore allowing all this configuration to happen via GUI, although I understand wanting to keep env var support for those who have already configured their computers accordingly.

app/experimentalFeatures.tsx
```diff
@@ -26,6 +26,18 @@ export default {
     developmentDefault: false,
     productionDefault: false,
   },
+  ros1Native: {
+    name: "ROS1 Native Connector",
```
jtbandes, line 30:
```suggestion
    name: "ROS 1 Direct Connections",
```

app/players/Ros1Player.ts
```diff
@@ -0,0 +1,320 @@
...
+  constructor(url: string) {
+    this.#url = url;
+    this.#start = fromMillis(Date.now());
+    this.#open();
+  }
+
+  #open = async (): Promise<void> => {
+    const os = OsContextSingleton;
+    if (this.#closed || os == undefined) {
+      return;
```
jtbandes, line 62: should this throw an error?

app/players/Ros1Player.ts
```diff
@@ -0,0 +1,320 @@
...
+// Connects to `rosmaster` instance using `@foxglove/ros1`. Currently doesn't support seeking or
+// showing simulated time, so current time from Date.now() is always used instead.
+export default class Ros1Player implements Player {
+  #url: string; // rosmaster URL.
+  #rosNode?: RosNode; // Our ROS node when we're connected.
+  #id: string = uuidv4(); // Unique ID for this player.
+  #listener?: (arg0: PlayerState) => Promise<void>; // Listener for #emitState().
+  #closed: boolean = false; // Whether the player has been completely closed using close().
+  #providerTopics?: Topic[]; // Topics as advertised by rosmaster.
+  #providerDatatypes: RosDatatypes = {}; // All ROS message definitions received from subscriptions.
```
jtbandes, line 43: remove "provider" terminology, that was in reference to DataProvider interface

app/players/Ros1Player.ts
```diff
@@ -0,0 +1,320 @@
...
+import { wrapJsObject } from "@foxglove-studio/app/util/binaryObjects";
+import debouncePromise from "@foxglove-studio/app/util/debouncePromise";
+import { getTopicsByTopicName } from "@foxglove-studio/app/util/selectors";
+import sendNotification from "@foxglove-studio/app/util/sendNotification";
+import { fromMillis, TimestampMethod } from "@foxglove-studio/app/util/time";
+import { Sockets } from "@foxglove/electron-socket/renderer";
+import { RosNode, TcpSocket } from "@foxglove/ros1";
+import { HttpServer } from "@foxglove/xmlrpc/src";
+
+const capabilities = [PlayerCapabilities.advertise];
```
jtbandes, line 31: As long as we don't yet support publishing, we should remove this from capabilities

app/players/Ros1Player.ts
```diff
@@ -0,0 +1,320 @@
...
+  #closed: boolean = false; // Whether the player has been completely closed using close().
+  #providerTopics?: Topic[]; // Topics as advertised by rosmaster.
+  #providerDatatypes: RosDatatypes = {}; // All ROS message definitions received from subscriptions.
+  #start?: Time; // The time at which we started playing.
+  #requestedSubscriptions: SubscribePayload[] = []; // Requested subscriptions by setSubscriptions()
+  #parsedMessages: Message[] = []; // Queue of messages that we'll send in next #emitState() call.
+  #bobjects: BobjectMessage[] = []; // Queue of bobjects that we'll send in next #emitState() call.
+  #messageOrder: TimestampMethod = "receiveTime";
+  #requestTopicsTimeout?: ReturnType<typeof setTimeout>; // setTimeout() handle for #requestTopics().
+  #bobjectTopics: Set<string> = new Set();
```
jtbandes, line 50: consider
```suggestion
  #bobjectTopics = new Set<string>();
```

app/players/Ros1Player.ts
```diff
@@ -0,0 +1,320 @@
...
+// showing simulated time, so current time from Date.now() is always used instead.
+export default class Ros1Player implements Player {
+  #url: string; // rosmaster URL.
+  #rosNode?: RosNode; // Our ROS node when we're connected.
+  #id: string = uuidv4(); // Unique ID for this player.
+  #listener?: (arg0: PlayerState) => Promise<void>; // Listener for #emitState().
+  #closed: boolean = false; // Whether the player has been completely closed using close().
+  #providerTopics?: Topic[]; // Topics as advertised by rosmaster.
+  #providerDatatypes: RosDatatypes = {}; // All ROS message definitions received from subscriptions.
+  #start?: Time; // The time at which we started playing.
```
jtbandes, line 44:
```suggestion
  #startTime?: Time; // The time at which we started playing.
```

app/players/Ros1Player.ts
```diff
@@ -0,0 +1,320 @@
...
+const capabilities = [PlayerCapabilities.advertise];
+const NO_WARNINGS = Object.freeze({});
+
+// Connects to `rosmaster` instance using `@foxglove/ros1`. Currently doesn't support seeking or
+// showing simulated time, so current time from Date.now() is always used instead.
+export default class Ros1Player implements Player {
+  #url: string; // rosmaster URL.
+  #rosNode?: RosNode; // Our ROS node when we're connected.
+  #id: string = uuidv4(); // Unique ID for this player.
+  #listener?: (arg0: PlayerState) => Promise<void>; // Listener for #emitState().
```
jtbandes, line 40:
```suggestion
  #listener?: (state: PlayerState) => Promise<void>; // Listener for #emitState().
```

app/players/Ros1Player.ts
```diff
@@ -0,0 +1,320 @@
...
+    const os = OsContextSingleton;
+    if (this.#closed || os == undefined) {
+      return;
+    }
+
+    const net = await Sockets.Create();
+    const httpServer = await net.createHttpServer();
+    const tcpSocketCreate = (options: { host: string; port: number }): Promise<TcpSocket> => {
+      return net.createSocket(options.host, options.port);
+    };
```
jtbandes, line 69: I think if you inline this in the constructor below, it should be able to infer types and you can just do `async ({host, port}) => net.createSocket(host, port)`

app/players/Ros1Player.ts
```diff
@@ -0,0 +1,320 @@
...
+
+    if (this.#rosNode == undefined) {
+      this.#rosNode = new RosNode({
+        name: "/foxglovestudio",
+        hostname: RosNode.GetRosHostname(os.getEnvVar, os.getHostname, os.getNetworkInterfaces),
+        pid: os.pid,
+        rosMasterUri: this.#url,
+        httpServer: (httpServer as unknown) as HttpServer,
```
jtbandes, line 77: why do we need a force cast here?

packages/ros1/src/RosNode.ts
```diff
@@ -253,4 +241,61 @@ export class RosNode {
...
+    // Fall back to iterating network interfaces looking for an IP address
+    let bestAddr: NetworkInterface | undefined;
+    const ifaces = getNetworkInterfaces();
+    for (const name in ifaces) {
+      const iface = ifaces[name];
+      if (iface == undefined) {
+        continue;
+      }
+      if (
+        (iface.family !== "IPv4" && iface.family !== "IPv6") ||
```
jtbandes, line 271:
```suggestion
      if (
        iface == undefined ||
        (iface.family !== "IPv4" && iface.family !== "IPv6") ||
```

app/players/Ros1Player.ts
```diff
@@ -0,0 +1,320 @@
...
+    const httpServer = await net.createHttpServer();
+    const tcpSocketCreate = (options: { host: string; port: number }): Promise<TcpSocket> => {
+      return net.createSocket(options.host, options.port);
+    };
+
+    if (this.#rosNode == undefined) {
+      this.#rosNode = new RosNode({
+        name: "/foxglovestudio",
+        hostname: RosNode.GetRosHostname(os.getEnvVar, os.getHostname, os.getNetworkInterfaces),
+        pid: os.pid,
```
jtbandes, line 75: remind me why hostname and pid are needed?

app/players/Ros1Player.ts
```diff
@@ -0,0 +1,320 @@
...
+
+    const net = await Sockets.Create();
+    const httpServer = await net.createHttpServer();
+    const tcpSocketCreate = (options: { host: string; port: number }): Promise<TcpSocket> => {
+      return net.createSocket(options.host, options.port);
+    };
+
+    if (this.#rosNode == undefined) {
+      this.#rosNode = new RosNode({
+        name: "/foxglovestudio",
```
jtbandes, line 73:
```suggestion
        name: "/foxglove_studio",
```

app/players/Ros1Player.ts
```diff
@@ -0,0 +1,320 @@
...
+      const sortedTopics: Topic[] = sortBy(topics, "name");
+      if (isEqual(sortedTopics, this.#providerTopics)) {
+        return;
+      }
+
+      this.#providerTopics = sortedTopics;
+
+      // Try subscribing again, since we might now be able to subscribe to some new topics.
+      this.setSubscriptions(this.#requestedSubscriptions);
+      this.#emitState();
```
jtbandes, line 108: it's unfortunate that this kind of logic (as well as the whole emitState method, for example) needs to be duplicated between the rosbridge and ros1 direct players. It doesn't feel sustainable to copy a hundred lines here and there — would it make sense to create a common "ROS 1" player that they both inherit from, and only override the small pieces of code that actually interface with ROS? Or is Player the wrong level of abstraction to support these different variations (I see some of this code is also quite similar to the RandomAccessPlayer)

app/players/Ros1Player.ts
```diff
@@ -0,0 +1,320 @@
...
+        Object.assign(typesByName, this.#providerDatatypes);
+        Object.assign(typesByName, this.#getRosDatatypes(datatype, msgdef));
+        this.#providerDatatypes = typesByName;
+      });
+      subscription.on("message", (message, _data, _publisher) => {
+        if (this.#providerTopics == undefined) {
+          return;
+        }
+
+        const receiveTime = fromMillis(Date.now());
```
jtbandes, line 227: Receive time feels like something that should be provided by the RosNode subscription API, like it is in roscpp

app/players/Ros1Player.ts
```diff
@@ -0,0 +1,320 @@
...
+    // no-op
+  }
+  requestBackfill(): void {
+    // no-op
+  }
+  setGlobalVariables(): void {
+    // no-op
+  }
+
+  #getRosDatatypes = (datatype: string, messageDefinition: RosMsgDefinition[]): RosDatatypes => {
```
jtbandes, line 307: static?

app/players/Ros1Player.ts
```diff
@@ -0,0 +1,320 @@
...
+    const typesByName: RosDatatypes = {};
+    messageDefinition.forEach(({ name, definitions }, index) => {
+      // The first definition usually doesn't have an explicit name,
+      // so we get the name from the connection.
```
jtbandes, line 311: I see this is now the 3rd occurrence of this code in the codebase -- is it something we can reuse?

app/util/url.test.ts
```diff
@@ -0,0 +1,29 @@
...
+      expect(parseInputUrl("")).toBeUndefined();
+    });
+
+    it("accepts fully formed URL input", () => {
+      expect(parseInputUrl("http://server.com:11311/")).toEqual("http://server.com:11311/");
+      expect(parseInputUrl("https://server.com:11311/")).toEqual("https://server.com:11311/");
+    });
+
+    it("accepts shorthand URL inputs", () => {
+      expect(parseInputUrl("http://localhost:11311")).toEqual("http://localhost:11311/");
```
jtbandes, line 20: Looks like this function should be called `normalizeInputUrl`

packages/electron-socket/renderer/HttpServerRenderer.ts
```diff
@@ -32,10 +34,10 @@ export class HttpServerRenderer extends EventEmitter {
     ["error", (args) => this.emit("error", new Error(args[0] as string))],
   ]);
 
-  constructor(messagePort: MessagePort, requestHandler: HttpHandler) {
+  constructor(messagePort: MessagePort, requestHandler?: HttpHandler) {
     super();
     this.#messagePort = messagePort;
-    this.#handler = requestHandler;
+    this.handler = requestHandler ?? (() => Promise.resolve({ statusCode: 404 }));
```
jtbandes, line 40:
```suggestion
    this.handler = requestHandler ?? (async () => ({ statusCode: 404 }));
```

packages/electron-socket/renderer/HttpServerRenderer.ts
```diff
@@ -57,15 +59,28 @@ export class HttpServerRenderer extends EventEmitter {
...
+      return Promise.reject(new Error(err));
+    }
+
+    // Store the URL we are listening at
+    const addr = await this.address();
+    if (addr == undefined || typeof addr === "string") {
+      this.#url = addr;
+    } else {
+      this.#url = `http://${hostname ?? addr.address}:${addr.port}/`;
     }
```
jtbandes, line 84: What are the situations where we'd get an `addr` that is just a string, where we'd want to ignore the provided `hostname`?

packages/ros1/src/RosMasterClient.test.ts
```diff
@@ -2,10 +2,62 @@
...
+        res.write(data);
+        res.end();
+      })
+      .listen(undefined, "localhost", async () => {
+        const port = (server.address() as AddressInfo).port;
```
jtbandes, line 49: why cast?

packages/ros1/src/RosMasterClient.test.ts
```diff
@@ -2,10 +2,62 @@
...
 
+import http from "http";
+import type { AddressInfo } from "net";
+
 import { RosMasterClient } from "./RosMasterClient";
 
 describe("RosMasterClient", () => {
-  it("can be instantiated", async () => {
-    new RosMasterClient("http://localhost:11311/");
+  it("getPublishedTopics", async () => {
```
jtbandes, line 11: Nice test 👍  I like how it's kind of an end-to-end test but the setup all fits in one file. Are there other methods we can test like this too? getTopicTypes/getSystemState?

packages/ros1/src/RosTcpMessageStream.ts
```diff
@@ -36,13 +45,14 @@ export class TcpMessageStream extends Transform {
         this.emit("message", payload);
       } else {
         // Decoded the message length field and transition to reading a message
-        this.#bytesNeeded = new DataView(payload.buffer).getUint32(0, true);
+        this.#bytesNeeded = new DataView(
+          payload.buffer,
+          payload.byteOffset,
+          payload.byteLength,
+        ).getUint32(0, true);
```
jtbandes, line 52: yikes! nice fix

packages/ros1/src/Subscription.ts
```diff
@@ -27,51 +28,63 @@ type PublisherInfo = [
   connectionInfo: string,
 ];
 
-export class Subscription {
+export declare interface Subscription {
```
jtbandes, line 31: whoa, what's `export declare`? how is this different from just `export interface Subscription`?

packages/ros1/src/TcpConnection.test.ts
```diff
@@ -0,0 +1,105 @@
...
+
+import { TcpConnection } from "./TcpConnection";
+import { TcpSocket } from "./TcpTypes";
+import { TcpSocketNode } from "./nodejs/TcpSocketNode";
+
+////////////////////////////////////////////////////////////////////////////////
+
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
```
jtbandes, line 16: ?

packages/ros1/src/TcpConnection.test.ts
```diff
@@ -0,0 +1,105 @@
...
+
+function hexToUint8Array(hex: string): Uint8Array {
```
jtbandes, line 18: Can we use `Buffer.from(..., 'base64')` instead?

packages/ros1/src/TcpConnection.test.ts
```diff
@@ -0,0 +1,105 @@
...
+      connection.on("message", (msg, data: Uint8Array) => {
+        expect(data.byteLength).toEqual(3);
+        resolve(msg);
+      });
+      connection.on("error", reject);
+    });
+    await socket.connect();
+    const msg = await p;
+
+    expect(msg).toEqual({ b: 255, g: 86, r: 69 });
```
jtbandes, line 98:
```suggestion
    expect(msg).toEqual({ b: 0xff, g: 0x56, r: 0x69 });
```

packages/ros1/src/backoff.test.ts
```diff
@@ -0,0 +1,20 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { backoffTime } from "./backoff";
+
+describe("backoff", () => {
+  it("backoffTime works", () => {
```
jtbandes, line 8:
```suggestion
describe("backoffTime", () => {
  it("works", () => {
```

packages/ros1/src/mock/XmlRpcMock.ts
```diff
@@ -27,14 +20,14 @@ const TURTLESIM_SERVICES = new Set([
   "/rosout/set_logger_level",
 ]);
 
-export class XmlRpcClientMock implements XmlRpcClient {
-  readonly serverUrl: URL;
+export class XmlRpcClientMock {
```
jtbandes, line 23: naming nit: I think we usually use Mock as a prefix rather than suffix

packages/ros1/src/nodejs/TcpSocketNode.ts
```diff
@@ -19,55 +17,54 @@ export class TcpSocketNode extends EventEmitter implements TcpSocket {
...
-      return undefined;
+      return Promise.resolve(undefined);
     }
     const port = this.#socket.localPort;
     const family = this.#socket.remoteFamily; // There is no localFamily
     const address = this.#socket.localAddress;
-    return port !== undefined && family !== undefined && address !== undefined
-      ? { port, family, address }
-      : undefined;
+    return Promise.resolve(
```
jtbandes, line 50: just make these functions `async` and then you don't need the Promise.resolves

app/components/MessagePathSyntax/messagePathsForDatatype.ts
jhurliman: This previously made the assumption that all parsed message definitions were always available before any topic was subscribed to. In a direction ROS connection, you only get message definitions as topics are subscribed to. So the typing completion for the raw topic panel would hit this exception whenever it tried to do path completion on a topic that was observed in the rosgraph but not subscribed to, because it can't know at that point that `turtlesim/Color` has `r, g, b` fields. This is the only place in the code I've found where this assumption was strongly held, and loosening it doesn't seem to break anything (because the app wouldn't function if it was hitting this before).

PR #329, opened 2021-03-21
Select the initial value when displaying a prompt

defunctzombie:
The common use case for displaying a prompt with an initial value
is for the user to change that value. We pre-select the string
so they user can start typing their new value without having to first
select the pre-filled value.


amacneil:
I guess depending on the context we might want to either focus the text field, or pre-select the text. For renaming layouts pre-selecting text seems reasonable, but I'm not sure about what other modal use cases we have.

defunctzombie:
@amacneil if there is no initial value, the text box will be focused and a cursor placed at the beginning so the user can start typing an initial value. If there is an initial value it will be selected entirely. If we get a use case where we think the default action is to _append_ or tweak some part (I think rare in practice) then we can add those as options to the prompt invocation.

PR #337, opened 2021-03-23
Revert "Quit app when all windows closed"

amacneil:
Reverts foxglove/studio#161

Keeping app open when windows are closed is standard / expected behavior for mac apps.

### DIFF
--- a/desktop/index.ts
+++ b/desktop/index.ts
@@ -350,7 +350,11 @@ app.on("ready", async () => {
   });
 });
 
-// Quit when all windows are closed
+// Quit when all windows are closed, except on macOS. There, it's common
+// for applications and their menu bar to stay active until the user quits
+// explicitly with Cmd + Q.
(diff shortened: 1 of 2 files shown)


defunctzombie:
This was originally removed because `yarn test:integration` will leave apps open even when the test exists (or fails). If you can check or fix that behavior then I am ok with this being reverted.

Similarly we need a way to open a window again once you've closed one.

I also don't think its necessary for our app to "stay open" since there is only one window of the app ever open - but maybe its time to go multi-window 🥳

amacneil:
I will look into the `yarn test:integration` behavior. Seems like the test should clean up after itself.

The code to open a new window when clicking dock icon already exists in main, it wasn't deleted (https://github.com/foxglove/studio/blob/main/desktop/index.ts#L344-L350).

Adding support for multiple windows seems like a no brainer 😄

PR #346, opened 2021-03-25
Speed up BZ2 decompression with WebAssembly

jtbandes:
Adds a small library in the same vein as https://github.com/cruise-automation/wasm-lz4, for bzip2 decompression.

Time to fully preload data from a 62MB bag compressed with bz2:
- v0.2: ~14sec
- this branch with `yarn build:prod && yarn package:darwin`: ~5sec
(Tested by watching the clock, dropping a bag file and seeing how long it takes for the playback bar to fill up.)

Sadly this is still much slower than the same bag with lz4 compression, which takes ~1 sec or less to parse.


packages/wasm-bz2/src/module.d.ts
```diff
@@ -0,0 +1,17 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
defunctzombie, line 1: `.d.ts` files are typically in a typings folder

packages/wasm-bz2/test/index.test.ts
```diff
@@ -0,0 +1,21 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
defunctzombie, line 1: Our convention is for tests to live alongside the source files.

webpack.renderer.config.ts
```diff
@@ -149,16 +149,6 @@ export function makeConfig(_: unknown, argv: WebpackArgv): Configuration {
         { test: /\.scss$/, loader: "sass-loader", options: { sourceMap: true } },
         { test: /\.woff2?$/, type: "asset/inline" },
         { test: /\.(glb|bag|ttf|bin)$/, type: "asset/resource" },
-        {
```
defunctzombie, line 152: 👏

PR #349, opened 2021-03-26
Add player "presence" enum replacing showSpinner & showInitializing

jtbandes:
Add a new PlayerPresence enum to player state, and remove showSpinner and showInitializing.

Change player to "maybePlayer" in the MessagePipelineProvider API, allowing the PlayerManager to inform it about player construction status. The MessagePipelineProvider is responsible for merging this information with the player's own presence info (such as initializing/connected/reconnecting) to produce the final `playerState.presence`.

Add a spinner / error indicator into the top bar to show player connection status. Error messages are currently not displayed, but we should do this ASAP.

Also extract MockMessagePipelineProvider into its own file.

https://user-images.githubusercontent.com/14237/112567106-58cd9680-8d84-11eb-9b0e-1d6a04019d1f.mov


app/App.tsx
```diff
@@ -234,7 +265,7 @@ export default function App(): ReactElement {
       type: "ws",
     },
     {
-      name: "HTTP",
+      name: "Bag File (HTTP)",
```
amacneil, line 268: can you move this after `Bag File (local)`, the current menu sort order looks really weird

app/App.tsx
```diff
@@ -164,6 +171,28 @@ function Root() {
...
+    switch (playerPresence) {
+      case PlayerPresence.NOT_PRESENT:
+      case PlayerPresence.PRESENT:
+        return undefined;
+      case PlayerPresence.CONSTRUCTING:
+      case PlayerPresence.INITIALIZING:
+      case PlayerPresence.RECONNECTING:
+        return (
+          <Icon small style={{ paddingLeft: 10 }}>
+            <SpinningLoadingIcon />
```
amacneil, line 184: I'm not sure if this is the correct place my comment/request, but can you hide the spinner while the dialog is open (the file picker or connection URL dialog)? It's not really "loading" at that stage.

amacneil:
Once you get to the stage of displaying error messages (whether in this PR or a separate one), I think we could take clues from the gmail connecting UI (centered popup):

image

Alternatively, maybe a full width status bar could pop up between the window titlebar and the rest of the panel layout.

app/App.tsx
```diff
@@ -164,6 +171,28 @@ function Root() {
     })();
   }, [appConfiguration, openWelcomeLayout]);
 
+  const presenceIcon = (() => {
```
defunctzombie, line 174: useMemo?

jtbandes: 🤷 can do, but would've also just put this inline below - only factoring it out to make the switch easier

PR #350, opened 2021-03-26
@foxglove/velodyne-cloud

jhurliman:
This library transforms raw Velodyne packets into PointCloud classes that closely resemble the structure of PointCloud2 ROS messages. Decoding a full HDL32E scan in my Linux VM takes 5-6ms.


packages/velodyne/perf/transform.js
```diff
@@ -0,0 +1,22 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
defunctzombie, line 1: Please convert this to typescript

packages/velodyne/src/Calibration.ts
```diff
@@ -0,0 +1,171 @@
...
+    return Array(rows)
+      .fill(0)
+      .map((_row, x) =>
+        Array(cols)
+          .fill(0)
+          .map((_col, y) => fullFiring * block(x, y) + singleFiring * point(x, y) + offset),
+      );
+  }
+
+  static Data(model: Model): CalibrationData {
```
defunctzombie, line 127: `FromData` might be a more suitable name? Or maybe this belongs in a separate file on converters?

packages/velodyne/src/Calibration.ts
```diff
@@ -0,0 +1,171 @@
...
+import HDL32E_json from "./data/calibration/32db.json";
+import HDL64E_S21_json from "./data/calibration/64e_s2.1-sztaki.json";
+import HDL64E_S3_json from "./data/calibration/64e_s3-xiesc.json";
+import HDL64E_json from "./data/calibration/64e_utexas.json";
+import VLP16HiRes_json from "./data/calibration/VLP16_hires_db.json";
+import VLP16_json from "./data/calibration/VLP16db.json";
+import VLS128_json from "./data/calibration/VLS128.json";
+import VLP32C_json from "./data/calibration/VeloView-VLP-32C.json";
+
+export class Calibration {
```
defunctzombie, line 15: Class documentation please

packages/velodyne/src/RawPacket.ts
```diff
@@ -0,0 +1,71 @@
...
+  distance(index: number): number {
+    return this.view.getUint16(4 + 3 * index, true);
+  }
+
+  intensity(index: number): number {
+    return this.data[4 + 3 * index + 2] as number;
+  }
+}
+
+export class RawPacket {
```
defunctzombie, line 36: Split these two classes into files? Typically we try to keep 1 class per file - helps keep intent for each class clear.

packages/velodyne/src/RawPacket.ts
```diff
@@ -0,0 +1,71 @@
...
+      const blockSize = RawPacket.BLOCK_SIZE;
+      const blockData = new Uint8Array(data.buffer, data.byteOffset + blockSize * i, blockSize);
+      this.blocks.push(new RawBlock(blockData));
+    }
+
+    const view = new DataView(data.buffer, data.byteOffset, data.byteLength);
+
+    this.gpsTimestamp = view.getUint32(1200, true);
+    this.factoryField1 = data[1204] as number;
+    this.factoryField2 = data[1205] as number;
```
jtbandes, line 67: `as number` shouldn't be needed when indexing a Uint8Array?

packages/velodyne/src/Transformer.test.ts
```diff
@@ -0,0 +1,83 @@
...
+
+import { HDL32E_PACKET1 } from "../fixtures/packets";
+import { Calibration } from "./Calibration";
+import { PointCloud } from "./PointCloud";
+import { RawPacket } from "./RawPacket";
+import { Transformer } from "./Transformer";
+import { Model } from "./VelodyneTypes";
+
+function norm(x: number, y: number, z: number): number {
+  return Math.sqrt(x * x + y * y + z * z);
```
jtbandes, line 15:
```suggestion
  return Math.hypot(x, y, z);
```

defunctzombie:
Please use a different name - possibly something with _ros_ in the name if appropriate or `velodyne-to-pointcloud`. My thinking is that if we layer need to convert to other formats this will have taken a top level _velodyne_ namespace.

packages/velodyne/perf/transform.js
jhurliman: This is JS on purpose, you can't produce flame graphs for TS code.

packages/velodyne/src/RawPacket.ts
jhurliman: It is if you want a `number` and not a `number | undefined`.

packages/velodyne/perf/transform.js
defunctzombie: What prevents us from producing flame graphs for ts code? We can run node with a transformer that transpiles the code on-the-fly (which is what ts-node does). How is the js different?

amacneil:
👏 awesome

PR #354, opened 2021-03-26
Remove TwoDimensionalPlot

defunctzombie:
There are currently two plot components. 2D Plot is first in the list and someone who
selects it thinking it can plot values over time for their messages will discover it cannot.

For now focus on the value over time use case of the "Plot" panel which remains. In the future
we may expand its capability to plot bespoke x/y values from a single message instance (similar
to the table panel).


amacneil (approved):
Sure, it's definitely confusing having two Plots. I would like to revisit similar functionality within the main Plot panel though.

PR #360, opened 2021-03-27
Make parseInputUrl more robust and flexible, use it for all PlayerManager input URLs

jhurliman:


app/components/PlayerManager.tsx
```diff
@@ -164,11 +165,17 @@ async function getPlayerBuilderFromUserSelection(
...
+      });
+      const url = parseInputUrl(result, "ros:", {
+        "http:": { defaultPort: 80 },
+        "https:": { defaultPort: 443 },
+        "ros:": { protocol: "http:", defaultPort: 11311 },
       });
-      const url = parseInputUrl(result, "http:");
       if (url == undefined) {
-        return;
+        throw new AppError(
```
defunctzombie, line 176: Should we be throwing in such instances or using the notification interface? @jtbandes

app/util/url.ts
```diff
@@ -2,19 +2,49 @@
...
 // Parse a user-input string as a URL using a forgiving parser that interprets
-// bare server names or server/port combos using an assumed default protocol
-export function parseInputUrl(str?: string, defaultProtocol = "https:"): string | undefined {
+// bare server names or server/port combos using the passed in default protocol
+// and a map of protocols to default port numbers and a separate (optional)
+// protocol name to rewrite to
+export function parseInputUrl(
+  str: string | undefined,
+  defaultProtocol = "https:",
+  protocols: { [proto: string]: { protocol?: string; defaultPort: number } } = DEFAULT_PROTOS,
```
defunctzombie, line 21: I'd probably rename defaultPort to `port` since what the user is passing in is already the fallback

package.json
```diff
@@ -105,7 +104,7 @@
     "ts-node": "9.1.1",
     "typescript": "4.2.2",
     "typescript-plugin-css-modules": "3.2.0",
-    "url-parse": "1.5.1",
+    "universal-url": "2.0.0",
```
defunctzombie, line 107: I am amused by the progression of packages. First whatwg-url, url-parse, and now universal-url. Not sure we can go past universal-url that is pretty vast. Maybe "spiritual-url" to move into another realm.

app/components/PlayerManager.tsx
jhurliman: I changed it to throw under guidance from @jtbandes

PR #362, opened 2021-03-27
Use Node v14 for CI

amacneil:
Maintain support for LTS release


amacneil:
No performance reduction in CI (actually faster than the last few builds on main, but it might be a fluke).

PR #363, opened 2021-03-27
🔪 MessagePipelineConsumer [FG-206]

jtbandes:
This is an old way of consuming the MessagePipeline context from the pre-hooks era. It was used in the Publish panel and some tests. This required converting Publisher to a usePublisher hook, and the whole Publish panel to hooks, and the tests to use react-hooks/testing-library.


app/components/MessagePipeline/index.test.tsx
```diff
@@ -11,117 +11,133 @@
...
+        return ctx.playerState.presence;
+      }),
+    ).toEqual([
+      // Some unsightly duplication here - either an artifact of renderHook or a potential
+      // improvement by modifying useContextSelector.
+      PlayerPresence.NOT_PRESENT,
       PlayerPresence.NOT_PRESENT,
       PlayerPresence.CONSTRUCTING,
+      PlayerPresence.CONSTRUCTING,
+      PlayerPresence.ERROR,
```
jtbandes, line 140: Sad about this, though was not able to determine why the extra renders are happening now when they weren't before. I think FG-202 may address this since it's also backed by useMessagePipeline.

app/hooks/usePublisher.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { useMessagePipeline } from "@foxglove-studio/app/components/MessagePipeline";
+
+type Props = {
+  topic: string;
+  datatype: string;
+  name: string;
+};
+
+// Registers a publisher with the player and returns a publish() function to publish data.
+export default function usePublisher({ topic, datatype, name }: Props): (msg: unknown) => void {
```
jhurliman, line 26: This resolves https://github.com/cruise-automation/webviz/pull/563

PR #364, opened 2021-03-27
Allow unused locals during `yarn serve`

defunctzombie:
See comment in code

### DIFF
--- a/webpack.renderer.config.ts
+++ b/webpack.renderer.config.ts
@@ -190,7 +190,18 @@ export function makeConfig(_: unknown, argv: WebpackArgv): Configuration {
         // available options: https://github.com/Microsoft/monaco-editor-webpack-plugin#options
         languages: ["typescript", "javascript"],
       }),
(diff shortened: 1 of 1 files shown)


defunctzombie:
@jtbandes @jhurliman @amacneil let me know if you think this is useful or will be annoying to work locally and then have CI fail your build.

Note that `yarn build:dev` will still fail with unusedLocal errors. This is _only_ for `yarn serve`.

jtbandes:
Can they be warnings so they still show up in the console? I suppose TS doesn't usually have the option to make something into a warning, that'd only be possible if we were doing it via eslint?

defunctzombie:
> Can they be warnings so they still show up in the console? I suppose TS doesn't usually have the option to make something into a warning, that'd only be possible if we were doing it via eslint?

Yea - so I can make an "exclude" function like I had for forktschecker with unchecked access that will console.log these to the terminal where you ran `yarn serve`. I did that at first but figured it could easily be overlooked and decided against it for v1. Could recosider.

PR #365, opened 2021-03-27
Add a yarn plugin to enforce the node engine req during `yarn install` and disallow npm

jhurliman:
I need to compile `semver` into the `.cjs` file before this will actually work.

### DIFF
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -3,6 +3,8 @@ defaultSemverRangePrefix: ""
 nodeLinker: node-modules
 
 plugins:
+  - path: .yarn/plugins/@foxglove/node-version-check.cjs
+    spec: "@foxglove/node-version-check"
   - path: .yarn/plugins/@yarnpkg/plugin-interactive-tools.cjs
     spec: "@yarnpkg/plugin-interactive-tools"
   - path: .yarn/plugins/@yarnpkg/plugin-typescript.cjs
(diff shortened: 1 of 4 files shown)


.yarnrc.yml
```diff
@@ -3,6 +3,8 @@ defaultSemverRangePrefix: ""
 nodeLinker: node-modules
 
 plugins:
+  - path: .yarn/plugins/@foxglove/node-version-check.cjs
+    spec: "@foxglove/node-version-check"
```
jhurliman, line 7: The source code to `node-version-check.cjs` is not showing up since it is going into LFS. Here it is for easy viewing:

```
const semver = require("semver");
const package = require("../../../package.json");

module.exports = {
  name: "@foxglove/node-version-check",
  factory: (require) => ({
    hooks: {
      registerPackageExtensions() {
        var reqVersion = package.engines.node;
        if (!semver.satisfies(process.version, reqVersion)) {
          console.error("Node.js " + reqVersion + " required");
          process.exit(1);
        }
      },
    },
  }),
};
```

defunctzombie:
@jhurliman you still hoping to land this?

jhurliman:
@defunctzombie yes, I'm keeping it on the back burner for when I (or anyone) has free cycles.

defunctzombie:
@jhurliman I tried the npm please-use-yarn trick locally (with .npmrc) and it didn't work to stop `npm install`

PR #366, opened 2021-03-29
Hide dev tools under View->Advanced->... menu

jhurliman:


amacneil:
@jhurliman do you think we need both Reload and Force Reload? Seems like maybe only the latter is maybe necessary.

PR #367, opened 2021-03-29
Keep layout menu open after deleting a layout

defunctzombie:
Keeping the layout menu open after deleting a layout provides visual feedback
that the action succeeded. The user can then delete or rename other layouts.

### DIFF
--- a/app/App.tsx
+++ b/app/App.tsx
@@ -277,15 +277,15 @@ export default function App(): ReactElement {
       <Provider store={globalStore}>
         <ExperimentalFeaturesLocalStorageProvider features={experimentalFeatures}>
           <ErrorBoundary>
-            <PlayerManager playerSources={playerSources}>
-              <NativeFileMenuPlayerSelection />
-              <DndProvider backend={HTML5Backend}>
-                <OsContextLayoutStorageProvider>
+            <OsContextLayoutStorageProvider>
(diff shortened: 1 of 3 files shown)


app/reducers/panels.ts
```diff
@@ -1034,7 +1035,8 @@ const panelsReducer = function (state: State, action: ActionTypes): State {
       break;
 
     default:
-      break;
+      // avoid returning a copy of the state if we did not handle the action
+      return state;
```
defunctzombie, line 1039: @jtbandes this is the fix to modifying the state object identity when the action is not a panel action

app/App.tsx
```diff
@@ -277,15 +277,15 @@ export default function App(): ReactElement {
       <Provider store={globalStore}>
         <ExperimentalFeaturesLocalStorageProvider features={experimentalFeatures}>
           <ErrorBoundary>
-            <PlayerManager playerSources={playerSources}>
-              <NativeFileMenuPlayerSelection />
-              <DndProvider backend={HTML5Backend}>
-                <OsContextLayoutStorageProvider>
+            <OsContextLayoutStorageProvider>
```
defunctzombie, line 280: I changed the nesting order here to avoid LayoutStorage being influenced by PlayerManager, DND etc since it doesn't need those parents.

PR #373, opened 2021-03-29
Telemetry reporting to Amplitude for release builds

jhurliman:
This implements telemetry reporting for release builds where the `AMPLITUDE_API_KEY` environment variable is configured. Opt-out is supported via `settings.json` using the same key VSCode uses, `telemetry.telemetryEnabled` (defaults to true). Only basic app start / data source open / play / pause events are reported right now, the existing player metrics code was not an exact fit for reporting human interaction events.


app/util/Analytics.ts
```diff
@@ -0,0 +1,66 @@
...
+  private _storage = new Storage();
+
+  constructor(options: { amplitudeApiKey: string | undefined }) {
+    const amplitudeApiKey = options.amplitudeApiKey;
+    if (amplitudeApiKey != undefined && amplitudeApiKey.length > 0) {
+      const userId = this.getUserId();
+      const deviceId = this.getDeviceId();
+      const appVersion = this.getAppVersion();
+      // eslint-disable-next-line no-restricted-syntax
+      console.log(
```
defunctzombie, line 29: We have a logging library in utils - please use that

PR #374, opened 2021-03-29
Add support for nav_msg/Path

defunctzombie:
Builds on top of https://github.com/cruise-automation/webviz/pull/592
to support nav_msg/Path messages.


defunctzombie:
This also fixes https://github.com/cruise-automation/webviz/issues/610 which was an interesting edge case created by having an override color on a laser scan message. This behavior was introduced in the original nav_msg/Path PR again webviz. The fix was to move the override color logic to where the nav_msg/Path non-marker is created rather than handling override color for all non-markers.

PR #378, opened 2021-03-30
Fix CSS color inheritance by adding <!doctype html> to index.html

jtbandes:
Apparently tables were not inheriting text color from their parent div because we were missing `<!doctype html>`, which causes the browser to use quirks-mode rendering: https://stackoverflow.com/a/24259068/23649 Truly we live in an age of wonders.


amacneil:
image

PR #379, opened 2021-03-30
Default to port 9090 for all rosbridge websocket connections

amacneil:


jhurliman (approved):

app/components/PlayerManager.tsx
```diff
@@ -197,8 +197,8 @@ async function getPlayerBuilderFromUserSelection(
         value: "ws://localhost:9090",
         transformer: (str) => {
           const result = parseInputUrl(str, "http:", {
```
jtbandes, line 199: Should this be `"ws:"`? I guess it doesn't matter, since the `http` will end up mapping to `ws` anyway?

PR #380, opened 2021-03-30
Switch xmlrpc fetch() dep from cross-fetch to @foxglove/just-fetch

jhurliman:
cross-fetch, isomorphic-fetch, universal-fetch, and fetch-ponyfill﻿ libraries all strive to provide a uniform implementation of the `fetch()` method across as many browsers and environments as possible. This means we end up using an XmlHttpRequest wrapper from the electron renderer, which is not ideal. The `@foxglove/just-fetch` library is a paper-thin wrapper around the browser native window.fetch and the node-fetch library in node.js.


packages/panacea-fetch/.eslintrc.yaml
```diff
@@ -0,0 +1,6 @@
+parserOptions:
+  project: packages/panacea-fetch/tsconfig.json
+
+ignorePatterns:
+  - dist/
+  - dist.browser/
```
jtbandes, line 6: May want `\.` if this is a regex? (Not sure if it's regex or minimatch)

jtbandes (approved):
Nice...so why did you choose to remove `noEmit` and add a build script? This script isn't used during our app webpack build, right? (It just imports the .ts files directly?)

jtbandes:
Name suggestion: `just-fetch` 😛

PR #381, opened 2021-03-30
Add repeat toggle to playback controls

defunctzombie:
The repeat toggle controls end-of-playback behavior. If the toggle is active,
when playback reaches the end it will restart from the begginning. If the
toggle is inactive, playback will pause at the end.

https://user-images.githubusercontent.com/84792/113018012-ed454980-9134-11eb-95f7-45c71200a19b.mov


app/players/RandomAccessPlayer.ts
```diff
@@ -364,24 +361,8 @@ export default class RandomAccessPlayer implements Player {
...
-      if (toSec(subtractTimes(this._end, this._start)) < LOOP_MIN_BAG_TIME_IN_SEC) {
-        // Don't loop for short bags.
-        this.pausePlayback();
-        return;
-      }
-      // Wait a little bit before we loop back. This helps with extremely small bags; otherwise
-      // it looks like it's stuck at the beginning of the bag.
-      await delay(500);
-      if (this._isPlaying) {
-        this._playFromStart();
```
jtbandes, line 383: This method is now unused - but may be worth looking at its implementation since it subtracts 1ns before playing. If that is necessary to get messages at the beginning, then we should keep it - though note that it causes a bug when the first message is at time 0: FG-281 so maybe we need to handle this some other way.

defunctzombie: What would be a good test for me to perform here? Load a bag with markers at bag start? Plot message at bag start? Will that will be sufficient?

jtbandes: I think so, but not familiar with whatever bug caused this to be added originally.

PR #406, opened 2021-03-30
Lazy Message Reader

defunctzombie:
The lazy message reader provides on-demand access and deserialization
of fields within a serialized ROS message.

This PR introduces the LazyMessageReader within the @foxglove/rosmsg-deser package and updates ParseMessageDataProvider to use LazyMessageReader instead of the rosbag.js MessageReader.

Creating a LazyMessage from a Uint8Array incurs little overhead. A `new LazyMessage(buffer);` call occurs and the LazyMessage constructor performs no work. Only during field access does deserialization occur. Deserialization is limited to the accessed fields.

### DIFF
--- a/packages/rosmsg-deser/README.md
+++ b/packages/rosmsg-deser/README.md
@@ -6,7 +6,7 @@ Deserialize ArrayBuffers using [ROS Message serialization format](http://wiki.ro
 
 Lazy messages provide on-demand access and deserialization to fields of a serialized ROS message. Creating
 a lazy message from a buffer performs no de-serialization during creation. Only accessed fields are
-deserialized. Deserialization occurs at access time.
+deserialized; the deserialization occurs at access time.
 
 ```Typescript
 import { LazyMessageReader } from "@foxglove/rosmsg-deser";
@@ -15,7 +15,7 @@ import { LazyMessageReader } from "@foxglove/rosmsg-deser";
 const reader = new LazyMessageReader(messageDefinition);
 
 // build a new lazy message instance for our serialized message from the Uint8Array
-// Note: the array lifetime should be as long as the message lifetime
+// Note: since deserialization is lazy - avoid-reusing the array you provide for other messages
 const message = reader.readMessage([0x00, 0x00, ...]);
 
 // access message fields
--- a/packages/rosmsg-deser/README.md
+++ b/packages/rosmsg-deser/README.md
@@ -1,0 +1,23 @@
+# rosmsg-deser
+
+Deserialize ArrayBuffers using [ROS Message serialization format](http://wiki.ros.org/msg).
+
+## LazyMessage
+
+Lazy messages provide on-demand access and deserialization to fields of a serialized ROS message. Creating
+a lazy message from a buffer performs no de-serialization during creation. Only accessed fields are
+deserialized. Deserialization occurs at access time.
+
+```Typescript
+import { LazyMessageReader } from "@foxglove/rosmsg-deser";
+
+// message definition comes from rosbag.js
+const reader = new LazyMessageReader(messageDefinition);
+
+// build a new lazy message instance for our serialized message from the Uint8Array
+// Note: the array lifetime should be as long as the message lifetime
+const message = reader.readMessage([0x00, 0x00, ...]);
+
+// access message fields
+message.header.stamp;
+```
--- a/packages/rosmsg-deser/src/BuiltinDeserialize.ts
+++ b/packages/rosmsg-deser/src/BuiltinDeserialize.ts
@@ -1,0 +1,169 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+interface Indexable {
+  [index: number]: unknown;
+}
+
+interface TypedArrayConstructor<T> {
+  new (length?: number): T;
+  new (buffer: ArrayBufferLike, byteOffset?: number, length?: number): T;
+  BYTES_PER_ELEMENT: number;
+}
+
+// Given a TypeArray constructor (Int32Array, Float32Array, etc), create a deserialization function
+// This deserialization function tries to first use aligned access and falls back to iteration
+function MakeTypedArrayDeserialze<T extends Indexable>(
+  TypedArrayConstructor: TypedArrayConstructor<T>,
+  getter:
+    | "getInt8"
+    | "getUint8"
+    | "getInt16"
+    | "getUint16"
+    | "getInt32"
+    | "getUint32"
+    | "getBigInt64"
+    | "getBigUint64"
+    | "getFloat32"
+    | "getFloat64",
+) {
+  return (view: DataView, offset: number, len: number): T => {
+    const totalOffset = view.byteOffset + offset;
+    // aligned access provides for a fast path to typed array construction
+    if (totalOffset % TypedArrayConstructor.BYTES_PER_ELEMENT === 0) {
+      return new TypedArrayConstructor(view.buffer, totalOffset, len);
+    }
+
+    // unaligned access requires additional handling
+
+    // benchmarks indicate for len < ~10 doing each individually is faster than copy
+    if (len < 10) {
+      const arr = new TypedArrayConstructor(len);
+      for (let idx = 0; idx < len; ++idx) {
+        arr[idx] = view[getter](offset, true);
+        offset += TypedArrayConstructor.BYTES_PER_ELEMENT;
+      }
+      return arr;
+    }
+
+    // if the length is > 10, then doing a copy of the data to align it is faster
+    // using _set_ is slightly faster than slice on the array buffer according to today's benchmarks
+    const size = TypedArrayConstructor.BYTES_PER_ELEMENT * len;
+    const copy = new Uint8Array(size);
+    copy.set(new Uint8Array(view.buffer, totalOffset, size));
+    return new TypedArrayConstructor(copy.buffer, copy.byteOffset, len);
+  };
+}
+
+const deserializers = {
+  int8: (view: DataView, offset: number): number => {
+    return view.getInt8(offset);
+  },
+  uint8: (view: DataView, offset: number): number => {
+    return view.getUint8(offset);
+  },
+  bool: (view: DataView, offset: number): boolean => {
+    return view.getUint8(offset) !== 0;
+  },
+  int16: (view: DataView, offset: number): number => {
+    return view.getInt16(offset, true);
+  },
+  uint16: (view: DataView, offset: number): number => {
+    return view.getUint16(offset, true);
+  },
+  int32: (view: DataView, offset: number): number => {
+    return view.getInt32(offset, true);
+  },
+  uint32: (view: DataView, offset: number): number => {
+    return view.getUint32(offset, true);
+  },
+  int64: (view: DataView, offset: number): bigint => {
+    return view.getBigInt64(offset, true);
+  },
+  uint64: (view: DataView, offset: number): bigint => {
+    return view.getBigUint64(offset, true);
+  },
+  float32: (view: DataView, offset: number): number => {
+    return view.getFloat32(offset, true);
+  },
+  float64: (view: DataView, offset: number): number => {
+    return view.getFloat64(offset, true);
+  },
+  time: (view: DataView, offset: number): { sec: number; nsec: number } => {
+    const sec = view.getUint32(offset, true);
+    const nsec = view.getUint32(offset + 4, true);
+    return { sec, nsec };
+  },
+  string: (view: DataView, offset: number): string => {
+    const len = view.getInt32(offset, true);
+    const codePoints = new Uint8Array(view.buffer, view.byteOffset + offset + 4, len);
+    const decoder = new (global as any).TextDecoder("ascii");
+    return decoder.decode(codePoints);
+  },
+  int8Array: MakeTypedArrayDeserialze(Int8Array, "getInt8"),
+  uint8Array: MakeTypedArrayDeserialze(Uint8Array, "getUint8"),
+  int16Array: MakeTypedArrayDeserialze(Int16Array, "getInt16"),
+  uint16Array: MakeTypedArrayDeserialze(Uint16Array, "getUint16"),
+  int32Array: MakeTypedArrayDeserialze(Int32Array, "getInt32"),
+  uint32Array: MakeTypedArrayDeserialze(Uint32Array, "getUint32"),
+  int64Array: MakeTypedArrayDeserialze(BigInt64Array, "getBigInt64"),
+  uint64Array: MakeTypedArrayDeserialze(BigUint64Array, "getBigUint64"),
+  float32Array: MakeTypedArrayDeserialze(Float32Array, "getFloat32"),
+  float64Array: MakeTypedArrayDeserialze(Float64Array, "getFloat64"),
+  timeArray: (view: DataView, offset: number, len: number): { sec: number; nsec: number }[] => {
+    // Time and Duration are actually arrays of int32
+    // If the location of the TimeArray meets Int32Array aligned access requirements, we can use Int32Array
+    // to speed up access to the int32 values.
+    // Otherwise we fall back to individual value reading via DataView
+
+    const timeArr = new Array<{ sec: number; nsec: number }>(len);
+
+    const totalOffset = view.byteOffset + offset;
+    // aligned access provides for a fast path to typed array construction
+    if (totalOffset % Int32Array.BYTES_PER_ELEMENT === 0) {
+      const intArr = new Int32Array(view.buffer, totalOffset, len * 2);
+      for (let i = 0, j = 0; i < len; ++i, j = j + 2) {
+        timeArr[i] = {
+          sec: intArr[j]!,
+          nsec: intArr[j + 1]!,
+        };
+      }
+    } else {
+      for (let idx = 0; idx < len; ++idx) {
+        timeArr[idx] = {
+          sec: view.getInt32(offset, true),
+          nsec: view.getInt32(offset + 4, true),
+        };
+        offset += 8;
+      }
+    }
+
+    return timeArr;
+  },
+  fixedArray: (
+    view: DataView,
+    offset: number,
+    len: number,
+    elementDeser: (view: DataView, offset: number) => unknown,
+    elementSize: (view: DataView, offset: number) => number,
+  ): unknown[] => {
+    const arr = new Array<unknown>(len);
+    for (let idx = 0; idx < len; ++idx) {
+      arr[idx] = elementDeser(view, offset);
+      offset += elementSize(view, offset);
+    }
+    return arr;
+  },
+  dynamicArray: (
+    view: DataView,
+    offset: number,
+    elementDeser: (buff: DataView, offset: number) => unknown,
+    elementSize: (buff: DataView, offset: number) => number,
+  ): unknown[] => {
+    const len = view.getUint32(offset, true);
+    return deserializers.fixedArray(view, offset + 4, len, elementDeser, elementSize);
+  },
+};
+
+export default deserializers;
--- a/packages/rosmsg-deser/src/BuiltinDeserialize.ts
+++ b/packages/rosmsg-deser/src/BuiltinDeserialize.ts
@@ -30,12 +30,11 @@ function MakeTypedArrayDeserialze<T extends Indexable>(
 ) {
   return (view: DataView, offset: number, len: number): T => {
     const totalOffset = view.byteOffset + offset;
-    // aligned access provides for a fast path to typed array construction
+    // new TypedArray(...) will throw if you try to make a typed array on unaligned boundary
+    // but for aligned access we can use a typed array and avoid any extra memory alloc/copy
     if (totalOffset % TypedArrayConstructor.BYTES_PER_ELEMENT === 0) {
       return new TypedArrayConstructor(view.buffer, totalOffset, len);
     }
-
-    // unaligned access requires additional handling
 
     // benchmarks indicate for len < ~10 doing each individually is faster than copy
     if (len < 10) {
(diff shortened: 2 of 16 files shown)


packages/rosmsg-deser/README.md
```diff
@@ -0,0 +1,23 @@
...
+deserialized. Deserialization occurs at access time.
+
+```Typescript
+import { LazyMessageReader } from "@foxglove/rosmsg-deser";
+
+// message definition comes from rosbag.js
+const reader = new LazyMessageReader(messageDefinition);
+
+// build a new lazy message instance for our serialized message from the Uint8Array
+// Note: the array lifetime should be as long as the message lifetime
```
jtbandes, line 18: Not sure what this comment means - does the user need to do something to ensure this? It internally makes a DataView which will keep a reference to the array, right?

packages/rosmsg-deser/src/BuiltinDeserialize.ts
```diff
@@ -0,0 +1,169 @@
...
+    const copy = new Uint8Array(size);
+    copy.set(new Uint8Array(view.buffer, totalOffset, size));
+    return new TypedArrayConstructor(copy.buffer, copy.byteOffset, len);
+  };
+}
+
+const deserializers = {
+  int8: (view: DataView, offset: number): number => {
+    return view.getInt8(offset);
+  },
```
jtbandes, line 62: nit: remove `{}` for auto formatting to give shorter line count

packages/rosmsg-deser/src/BuiltinDeserialize.ts
```diff
@@ -0,0 +1,169 @@
...
+      return new TypedArrayConstructor(view.buffer, totalOffset, len);
+    }
+
+    // unaligned access requires additional handling
+
+    // benchmarks indicate for len < ~10 doing each individually is faster than copy
+    if (len < 10) {
+      const arr = new TypedArrayConstructor(len);
+      for (let idx = 0; idx < len; ++idx) {
+        arr[idx] = view[getter](offset, true);
```
jtbandes, line 44: hoist `view[getter]` out of the loop?

packages/rosmsg-deser/src/BuiltinDeserialize.ts
```diff
@@ -0,0 +1,169 @@
...
+    | "getFloat64",
+) {
+  return (view: DataView, offset: number, len: number): T => {
+    const totalOffset = view.byteOffset + offset;
+    // aligned access provides for a fast path to typed array construction
+    if (totalOffset % TypedArrayConstructor.BYTES_PER_ELEMENT === 0) {
+      return new TypedArrayConstructor(view.buffer, totalOffset, len);
+    }
+
+    // unaligned access requires additional handling
```
jtbandes, line 38: this comment could use expansion - what goes wrong if unaligned?

packages/rosmsg-deser/src/BuiltinDeserialize.ts
```diff
@@ -0,0 +1,169 @@
...
+  },
+  time: (view: DataView, offset: number): { sec: number; nsec: number } => {
+    const sec = view.getUint32(offset, true);
+    const nsec = view.getUint32(offset + 4, true);
+    return { sec, nsec };
+  },
+  string: (view: DataView, offset: number): string => {
+    const len = view.getInt32(offset, true);
+    const codePoints = new Uint8Array(view.buffer, view.byteOffset + offset + 4, len);
+    const decoder = new (global as any).TextDecoder("ascii");
```
jtbandes, line 101: I know we already used ascii in some places when decoding - but I question this choice... I think better to assume UTF8 if we have to assume something?

packages/rosmsg-deser/src/BuiltinDeserialize.ts
```diff
@@ -0,0 +1,169 @@
...
+    // If the location of the TimeArray meets Int32Array aligned access requirements, we can use Int32Array
+    // to speed up access to the int32 values.
+    // Otherwise we fall back to individual value reading via DataView
+
+    const timeArr = new Array<{ sec: number; nsec: number }>(len);
+
+    const totalOffset = view.byteOffset + offset;
+    // aligned access provides for a fast path to typed array construction
+    if (totalOffset % Int32Array.BYTES_PER_ELEMENT === 0) {
+      const intArr = new Int32Array(view.buffer, totalOffset, len * 2);
```
jtbandes, line 125: Does Int32Array guarantee little endian access or do we need to guard for this too?
edit: found it later ... perhaps we can just fall back to the other path if big endian?

packages/rosmsg-deser/src/DeserializerFactory.ts
```diff
@@ -0,0 +1,319 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+import { RosMsgDefinition, RosMsgField } from "rosbag";
+
+import deserializers from "./BuiltinDeserialize";
+
+// Sizes for builtin types - if a type has a fixed size, the deserializer is able to optimize
+const fixedSizeTypes = new Map<string, number>([
```
jtbandes, line 9: this is probably not a big deal, but if I had to guess I would think plain objects would be faster than Map for small amounts of data, since they have been heavily optimized for years?

packages/rosmsg-deser/src/DeserializerFactory.ts
```diff
@@ -0,0 +1,319 @@
...
+      const elementSize = typeSize(view, offset);
+      size += elementSize;
+      offset += elementSize;
+    }
+    return size;
+  },
+};
+
+function sanitizeName(name: string): string {
+  return name.replace("/", "_");
```
jtbandes, line 63:
```suggestion
  return name.replaceAll(/^[0-9]|[^a-zA-Z0-9_]/g, "_");
```

packages/rosmsg-deser/src/DeserializerFactory.ts
```diff
@@ -0,0 +1,319 @@
...
+export default function buildClass(types: RosMsgDefinition[]): SerializedMessageReader {
+  const classes = new Array<string>();
+
+  // Build the types in reverse order so the root message appears last
+  // Since the root message depends on custom types we want those to be defined
+  for (const type of types.reverse()) {
+    const name = sanitizeName(type.name ?? "__RootMsg");
+
+    const offsetFns = new Array<string>();
+    const fields = new Array<string>();
```
jtbandes, line 232: comment on what these represent? I was surprised to find fields is in `x: this.x` format

packages/rosmsg-deser/src/DeserializerFactory.ts
```diff
@@ -0,0 +1,319 @@
...
+      // they are the size of the offset of the previous field + size of previous field
+      // the first first field is at offset 0
+      if (!prevField) {
+        offsetFns.push(`
+          ${field.name}_offset(view, initOffset) {
+            return initOffset;
+          }`);
+      } else {
+        // offsets tell you where the raw data of your field starts (including any length bytes)
+        // they are the size of the offset of the previous field + size of previous field
```
jtbandes, line 253: duplicated comment

packages/rosmsg-deser/src/DeserializerFactory.ts
```diff
@@ -0,0 +1,319 @@
...
+        // ${field.type} ${field.name}
+        totalSize += ${fieldSize};
+        offset += ${fieldSize};
+      `;
+    }
+  }
+
+  return `
+    // ${field.type} ${field.name}
+    {
```
jtbandes, line 146: why braces for this one?

packages/rosmsg-deser/src/DeserializerFactory.ts
```diff
@@ -0,0 +1,319 @@
...
+
+    classes.push(messageSrc);
+  }
+
+  const src = classes.join("\n\n");
+
+  // close over our builtin deserializers and builtin size functions
+  // eslint-disable-next-line no-new-func
+  const wrapFn = new Function("readers", "sizes", `${src}\nreturn __RootMsg;`);
+  return wrapFn.call(undefined, deserializers, builtinSizes) as SerializedMessageReader;
```
jtbandes, line 318: nit: would help readability if the readers & sizes kept the same name inside the generated code

packages/rosmsg-deser/src/DeserializerFactory.ts
```diff
@@ -0,0 +1,319 @@
...
+      } else {
+        // offsets tell you where the raw data of your field starts (including any length bytes)
+        // they are the size of the offset of the previous field + size of previous field
+        offsetFns.push(`
+          ${field.name}_offset(view, initOffset) {
+            const prevOffset = this.${prevField.name}_offset(view, initOffset);
+            const totalOffset = prevOffset + ${name}.${prevField.name}_size(view, prevOffset);
+  
+            // future calls will avoid re-calculating the offset
+            this.${field.name}_offset = () => totalOffset;
```
jtbandes, line 260: wonder if you would get a speedup by not building a function here, just checking `typeof foo_offset === "number"` when accessing

packages/rosmsg-deser/src/BuiltinDeserialize.ts
```diff
@@ -0,0 +1,169 @@
...
+    const nsec = view.getUint32(offset + 4, true);
+    return { sec, nsec };
+  },
+  string: (view: DataView, offset: number): string => {
+    const len = view.getInt32(offset, true);
+    const codePoints = new Uint8Array(view.buffer, view.byteOffset + offset + 4, len);
+    const decoder = new (global as any).TextDecoder("ascii");
+    return decoder.decode(codePoints);
+  },
+  int8Array: MakeTypedArrayDeserialze(Int8Array, "getInt8"),
```
jtbandes, line 104: missing boolArray and durationArray? Or are these types never actually returned by rosbag.js since they are kind of aliases - in which case why do we list fixed sizes for them?

packages/rosmsg-deser/src/DeserializerFactory.ts
jtbandes: This will enforce that for each key `F` used in the Map, `deserializers.${F}Array` must exist. Should catch the missing `boolArray` and `durationArray` cases (assuming the Map constructor is typed correctly - if not, the same solution can be used with an object type)

```suggestion
type ArrayTypes<Keys> = Keys extends `${infer T}Array` ? T : never;
const fixedSizeTypes = new Map<ArrayTypes<keyof typeof deserializers>, number>([
```

packages/rosmsg-deser/src/DeserializerFactory.ts
```diff
@@ -0,0 +1,319 @@
...
+        return `
+          // ${field.type}[${arrLen}] ${field.name}
+          get ${field.name}() {
+            const offset = this.${field.name}_offset(this.#view, this.#offset);
+            return readers.${field.type}Array(this.#view, offset, ${arrLen});
+          }`;
+      } else {
+        // fixed size array of complex size items
+        return `
+        // ${field.type}[${arrLen}] ${field.name}
```
jtbandes, line 181:
```suggestion
          // ${field.type}[${arrLen}] ${field.name}
```

packages/rosmsg-deser/src/DeserializerFactory.ts
```diff
@@ -0,0 +1,319 @@
...
+      // total size is known, which means we should use a builtin array reader
+      if (fieldSize != undefined) {
+        return `
+          // ${field.type}[${arrLen}] ${field.name}
+          get ${field.name}() {
+            const offset = this.${field.name}_offset(this.#view, this.#offset);
+            return readers.${field.type}Array(this.#view, offset, ${arrLen});
+          }`;
+      } else {
+        // fixed size array of complex size items
```
jtbandes, line 179: I like this comment, could use a similar one in each of the 5 cases

packages/rosmsg-deser/src/DeserializerFactory.ts
```diff
@@ -0,0 +1,319 @@
...
+            return totalOffset;
+          }`);
+      }
+
+      fields.push(`${field.name}: this.${field.name}`);
+
+      prevField = field;
+    }
+
+    const messageSrc = `class ${name} {
```
jtbandes, line 270: nit: for readability & consistency with some other style
```suggestion
    const messageSrc = `
    class ${name} {
```

packages/rosmsg-deser/src/DeserializerFactory.ts
```diff
@@ -0,0 +1,319 @@
...
+//
+// The output is a set of classes - one for each custom message type. Only the root message
+// class is exposed.
+//
+// Each LazyMessage class consists of static _size_ functions, _offset_ methods, and property _getters.
+// The size functions calculate the size of fields within arrays.
+// The offset methods calculate the start byte of the field within the entire message buffer.
+// The getter de-serializes the field from the message buffer.
+export default function buildClass(types: RosMsgDefinition[]): SerializedMessageReader {
+  const classes = new Array<string>();
```
jtbandes, line 224: style nit
```suggestion
  const classes: string[] = [];
```

packages/rosmsg-deser/src/DeserializerFactory.ts
```diff
@@ -0,0 +1,319 @@
...
+  for (const type of types.reverse()) {
+    const name = sanitizeName(type.name ?? "__RootMsg");
+
+    const offsetFns = new Array<string>();
+    const fields = new Array<string>();
+
+    // getters need to "look back" at the previous field to create the offset function calls
+    let prevField: RosMsgField | undefined;
+
+    for (const field of type.definitions) {
```
jtbandes, line 237: I think this loop could stand to be a separate function - this function would become more readable

packages/rosmsg-deser/src/DeserializerFactory.ts
```diff
@@ -0,0 +1,319 @@
...
+      } else {
+        return `
+          // ${field.type}[] ${field.name}
+          get ${field.name}() {
+            const offset = this.${field.name}_offset(this.#view, this.#offset);
+            return readers.dynamicArray(this.#view, offset, ${readerFn}, ${sizeFn});
+          }`;
+      }
+    }
+  } else {
```
jtbandes, line 206: maybe put this in an early return - could be easier to follow

packages/rosmsg-deser/src/DeserializerFactory.ts
```diff
@@ -0,0 +1,319 @@
...
+        #view = undefined;
+        #offset;
+  
+        constructor(view, offset = 0) {
+          this.#view = view;
+          this.#offset = offset;
+        }
+
+        // return a json object of the fields
+        // this fully parses the message
```
jtbandes, line 300: so if I ask for a subfield `msg.f` that is a complex type, it fully recursively parses all the fields in f? it doesn't look like it to me, but maybe I have missed something

packages/rosmsg-deser/src/DeserializerFactory.ts
```diff
@@ -0,0 +1,319 @@
...
+// The size functions calculate the size of fields within arrays.
+// The offset methods calculate the start byte of the field within the entire message buffer.
+// The getter de-serializes the field from the message buffer.
+export default function buildClass(types: RosMsgDefinition[]): SerializedMessageReader {
+  const classes = new Array<string>();
+
+  // Build the types in reverse order so the root message appears last
+  // Since the root message depends on custom types we want those to be defined
+  for (const type of types.reverse()) {
+    const name = sanitizeName(type.name ?? "__RootMsg");
```
jtbandes, line 229: this requires that the outer type does not have a name, right? might want to document that explicitly or assert.

packages/rosmsg-deser/src/LazyMessageReader.test.ts
```diff
@@ -0,0 +1,205 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+import { parseMessageDefinition } from "rosbag";
+import { TextDecoder } from "util";
+
+import { LazyMessageReader } from "./LazyMessageReader";
+
+// Jest runs in a node environment, TextDecoder is not available globally.
+// We provide TextDecoder from the uitl module.
```
jtbandes, line 10:
```suggestion
// We provide TextDecoder from the util module.
```

jtbandes: We write this in other places as `global.TextDecoder = util.TextDecoder as typeof TextDecoder;`

packages/rosmsg-deser/src/LazyMessageReader.test.ts
```diff
@@ -0,0 +1,205 @@
...
+(global as any).TextDecoder = TextDecoder;
+
+const serializeString = (str: string): Uint8Array => {
+  const data = Buffer.from(str, "utf8");
+  const len = Buffer.alloc(4);
+  len.writeInt32LE(data.byteLength, 0);
+  return Uint8Array.from([...len, ...data]);
+};
+
+const Float32Buffer = (floats: number[]): Iterable<number> => {
```
jtbandes, line 20: not a class
```suggestion
const float32Buffer = (floats: number[]): Iterable<number> => {
```

packages/rosmsg-deser/src/LazyMessageReader.test.ts
```diff
@@ -0,0 +1,205 @@
...
+    [`uint16 sample # highest`, [0xff, 0xff], { sample: 65535 }],
+    [`int32 sample # lowest`, [0x00, 0x00, 0x00, 0x80], { sample: -2147483648 }],
+    [`int32 sample # highest`, [0xff, 0xff, 0xff, 0x7f], { sample: 2147483647 }],
+    [`uint64 sample # lowest`, [0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00], { sample: 0n }],
+    [
+      `uint64 sample # highest`,
+      [0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff],
+      { sample: 18446744073709551615n },
+    ],
+    [`float32 sample`, Float32Buffer([5.5]), { sample: 5.5 }],
```
jtbandes, line 42: maybe use the same float64 test number to show it gets truncated?

packages/rosmsg-deser/src/BuiltinDeserialize.ts
```diff
@@ -0,0 +1,169 @@
...
+  },
+  uint32: (view: DataView, offset: number): number => {
+    return view.getUint32(offset, true);
+  },
+  int64: (view: DataView, offset: number): bigint => {
+    return view.getBigInt64(offset, true);
+  },
+  uint64: (view: DataView, offset: number): bigint => {
+    return view.getBigUint64(offset, true);
+  },
```
jtbandes, line 86: Does our app actually support these or will it crash when trying to display them?

packages/rosmsg-deser/src/LazyMessageReader.test.ts
```diff
@@ -0,0 +1,205 @@
...
+    [`int8 sample # highest`, [0x7f], { sample: 127 }],
+    [`uint8 sample # lowest`, [0x00], { sample: 0 }],
+    [`uint8 sample # highest`, [0xff], { sample: 255 }],
+    [`int16 sample # lowest`, [0x00, 0x80], { sample: -32768 }],
+    [`int16 sample # highest`, [0xff, 0x7f], { sample: 32767 }],
+    [`uint16 sample # lowest`, [0x00, 0x00], { sample: 0 }],
+    [`uint16 sample # highest`, [0xff, 0xff], { sample: 65535 }],
+    [`int32 sample # lowest`, [0x00, 0x00, 0x00, 0x80], { sample: -2147483648 }],
+    [`int32 sample # highest`, [0xff, 0xff, 0xff, 0x7f], { sample: 2147483647 }],
+    [`uint64 sample # lowest`, [0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00], { sample: 0n }],
```
jtbandes, line 36: missing signed int64

packages/rosmsg-deser/src/LazyMessageReader.test.ts
```diff
@@ -0,0 +1,205 @@
...
+    const reader = new LazyMessageReader(parseMessageDefinition(msgDef));
+    const read = reader.readMessage(buffer);
+
+    // check that our reader expected size matches the buffer size
+    expect(reader.size(buffer)).toEqual(buffer.length);
+
+    // check that our message matches the object
+    expect(read.toJSON()).toMatchObject(expected);
+  });
+});
```
jtbandes, line 205: can you please add some tests for the source generation with `expect.toMatchSnapshot()` - I think it would be helpful to be able to read/review the generated source code and clearly see future changes in diffs

packages/rosmsg-deser/src/LazyMessageReader.ts
```diff
@@ -0,0 +1,46 @@
...
+  parser: ReturnType<typeof buildClass>;
+  definitions: RosMsgDefinition[];
+
+  constructor(definitions: RosMsgDefinition[]) {
+    this.parser = buildClass(definitions);
+    this.definitions = definitions;
+  }
+
+  // Return the size of our message within the buffer
+  size(buffer: Readonly<Uint8Array>): number {
```
jtbandes, line 34: Does this Readonly work? I think it goes on a per field basis - which would not prevent things like `.set()`

packages/rosmsg-deser/src/LazyMessageReader.ts
```diff
@@ -0,0 +1,46 @@
...
+// the browser makes typed arrays use the architecture endianness and ROS messages are little endian
+const isLittleEndian = !isBigEndian();
+if (!isLittleEndian) {
+  throw new Error("Only Little Endian architectures are supported");
+}
+
+type LazyMessage<T> = T & { toJSON: () => T };
+
+export class LazyMessageReader<T = unknown> {
+  parser: ReturnType<typeof buildClass>;
```
jtbandes, line 25: there are 3 names here: `parser`, `buildClass`, and `DeserializerFactory` - can we standardize? e.g. `import buildParser from "./ParserFactory"`

packages/rosmsg-deser/src/LazyMessageReader.ts
```diff
@@ -0,0 +1,46 @@
...
+  // Return the size of our message within the buffer
+  size(buffer: Readonly<Uint8Array>): number {
+    const view = new DataView(buffer.buffer, buffer.byteOffset);
+    return this.parser.size(view);
+  }
+
+  // Create a LazyMessage for the buffer
+  // We template on R here for call site type information if the class type information T is not
+  // known or available
+  readMessage<R = T>(buffer: Readonly<Uint8Array>): LazyMessage<R> {
```
jtbandes, line 42: Is this any better than the caller just doing `(reader as LazyMessageReader<T>).readMessage()`? I think you can remove `R`

defunctzombie: only in the sense that you can do `reader.readMessage<MessageType>()` which I find cleaner than `(reader as LazyMessageReader<MessageType>).readMessage()` tho one could argue that the cast of reader is the logically more correct action since you wouldn't be able to read different typed messages from one reader - which amusingly enough is also not true - for example you could `reader.readMessage<Header>(...)` if the header was the first field - and due to how messages are constructed and the fact that we lazy ready - you would be able to read the header out only...

packages/rosmsg-deser/src/LazyMessageReader.ts
defunctzombie: Probably not as we would expect. I guess I can remove it - its more of a signal of intent /shrug

packages/rosmsg-deser/src/BuiltinDeserialize.ts
defunctzombie: The offset is relative to the offset passed into the containing function - not relative to idx

packages/rosmsg-deser/src/BuiltinDeserialize.ts
defunctzombie: it does not - It uses the system endianness. This is why I guard the entire file:
https://github.com/foxglove/studio/pull/406/files#diff-6838f9d10b45a5f8a1e9b56272c89fac1274c08760ba56fe82caec79a21c4aa7R18

I figured a future enhancement would be to make this support big endian but for now everything we care about uses little during runtime.

packages/rosmsg-deser/src/DeserializerFactory.ts
defunctzombie: Breaks `fixedSizeTypes.get(field.type)` elsewhere in the code since field.type is a string (RosMsgField interface defines that).

PR #408, opened 2021-03-30
Always call `menuRemoveInputSource` before `menuAddInputSource`

jhurliman:
This should hopefully address Sentry errors about adding native file menu handlers multiple times

### DIFF
--- a/app/components/NativeFileMenuPlayerSelection.tsx
+++ b/app/components/NativeFileMenuPlayerSelection.tsx
@@ -13,6 +13,7 @@ export function NativeFileMenuPlayerSelection(): ReactElement {
 
   useEffect(() => {
     for (const item of availableSources) {
+      OsContextSingleton?.menuRemoveInputSource(item.name);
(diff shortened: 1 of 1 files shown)


app/components/NativeFileMenuPlayerSelection.tsx
```diff
@@ -13,6 +13,7 @@ export function NativeFileMenuPlayerSelection(): ReactElement {
 
   useEffect(() => {
     for (const item of availableSources) {
+      OsContextSingleton?.menuRemoveInputSource(item.name);
       OsContextSingleton?.menuAddInputSource(item.name, () => {
```
amacneil, line 17: should we just make `menuAddInputSource` always remove an existing item with the same name?

jhurliman: That would be a cleaner approach I think.

defunctzombie:
https://github.com/foxglove/studio/pull/409

PR #412, opened 2021-03-31
Don't attempt to publish empty topics, don't log 'not implemented' errors to Sentry

jhurliman:


app/players/Ros1Player.ts
```diff
@@ -274,12 +274,13 @@ export default class Ros1Player implements Player {
...
   setPublishers(publishers: AdvertisePayload[]): void {
     // TODO: Publishing
+    publishers = publishers.filter((p) => p.topic.length > 0);
     if (publishers.length > 0) {
       const topics = publishers.map((p) => p.topic).join(", ");
       sendNotification(
         "Publishing not supported",
         `Cannot publish to "${topics}", ROS publishing is not supported yet`,
-        "app",
+        "user",
```
jtbandes, line 283: 🤷 this kinda is a programming error — nobody should be allowed to register as a publisher if publishing is not supported.

PR #416, opened 2021-03-31
Add @foxglove/log package

defunctzombie:
@fogxlove/log provides a common logging facility for defining
logging channels.

This change adds an InternalLog panel to toggle logging channels.

---
Screen Shot 2021-03-30 at 10 39 57 PM
Screen Shot 2021-03-30 at 10 40 09 PM


app/dataProviders/IdbCacheWriterDataProvider.ts
```diff
@@ -46,7 +46,7 @@ import {
   PRIMARY_KEY,
 } from "./IdbCacheDataProviderDatabase";
 
-const log = new Logger(__filename);
+const log = Logger.getLogger(__filename);
```
amacneil, line 49: `Logger.get()` maybe to avoid stutter?

defunctzombie: get seemed too vague - I thought about _createLogger_ or _childLogger_. Settled on _getLogger_ which matches python (https://docs.python.org/3/library/logging.html) and deno (https://deno.land/std@0.92.0/log/README.md) so I figured decent prior art for consistent apis.

amacneil: I guess. I've spent too much time with Go where they are allergic to stutter like Foo.createFoo

PR #417, opened 2021-03-31
Add a unit test that checks dependency licenses against an allowlist

jhurliman:


amacneil:
Does this check packages inside yarn workspaces?

test/licenses.test.ts
```diff
@@ -0,0 +1,56 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
defunctzombie, line 1: Can we do this as a separate check/script instead of as a unit test?

jhurliman:
@amacneil our own workspace packages are currently ignored while they have `"private": true` enabled. All transitive dependencies (including dependencies of all packages in our workspace) should be checked since they all end up in the root `node_modules/` folder.

amacneil:
Ok cool. I was wondering about whether dependencies that ended up in nested `node_modules` folders would get checked, but I guess that would only happen for packages where we have multiple versions of the same package, and hopefully they have the same license.

PR #428, opened 2021-03-31
Add screenshot to README

jtbandes:


jhurliman:
Weird GitHub formatting, hopefully it’s just in the preview.

EDD9E5ED-85F9-45D8-B9B2-7CFF79498DE1

amacneil:
I think put the screenshot after the intro blurb?

Also see if it works if you commit it to the repo with lfs

jtbandes:
Hm, no @jhurliman I think you're right, it's just a problem only at very small widths.

jtbandes:
I made the screenshot full width. I think it takes up a lot of space, which is why I tried to right-align it, but I guess that had problems too since there's no way to do breakpoints for different screen sizes.

PR #429, opened 2021-03-31
Enable lint error on `await`ing non-Promises

jtbandes:
Enables `@typescript-eslint/await-thenable` and fixes outstanding errors.


defunctzombie:
Is there an error for the opposite as well? Failing to await on a promise or failing to capture the promise into a variable?

Is awaiting a non-promise bad?

jtbandes:
Yeah, there is [`no-floating-promises`](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-floating-promises.md).

Is it "bad"? Technically no, the behavior is well defined. But I think it is a potential footgun. You can always `await Promise.resolve(x)` instead of `await x` if you want.

PR #432, opened 2021-03-31
Open the app maximized for maximum experience

defunctzombie:


defunctzombie:
Seems I am alone in my desire for MAXIMUM SPACE.

PR #436, opened 2021-04-01
Auto hide window menu bar (linux and windows) unless Alt pressed.

defunctzombie:


jhurliman:
hidden

After pressing Alt key:

altkey

PR #440, opened 2021-04-01
Support all PointCloud2 field data types, perf boost for reading point cloud color from non-float fields

jhurliman:


app/panels/ThreeDimensionalViz/commands/PointClouds/buffers.ts
```diff
@@ -35,8 +38,8 @@ export const FLOAT_SIZE = Float32Array.BYTES_PER_ELEMENT;
 // Reinterpret a buffer of bytes as a buffer of float values
 export function reinterpretBufferToFloat(buffer: Uint8Array): Float32Array {
   return new Float32Array(
-    new Uint8Array(buffer).buffer,
-    0,
```
jtbandes, line 39: Was this just broken before?

jhurliman: It looks like it was always incorrect, but happened to work because we were only passing in `Uint8Array`s with a `byteOffset` of 0. This would break with more efficient message parsing, though.

PR #442, opened 2021-04-01
Create identity transforms for unseen TFs referenced by header.frame_id, always select a rootTf if we can

jhurliman:
Rendering is skipped entirely if there is no rootTf. Therefore, we should always select a rootTf unless there is absolutely no renderable data that could be shown

### DIFF
--- a/app/panels/ThreeDimensionalViz/Transforms.ts
+++ b/app/panels/ThreeDimensionalViz/Transforms.ts
@@ -27,6 +27,10 @@ function stripLeadingSlash(name: string) {
   return name.startsWith("/") ? name.slice(1) : name;
 }
 
+// REP 105 specifies a set of conventional root frame transform ids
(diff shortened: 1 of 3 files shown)


app/panels/ThreeDimensionalViz/Transforms.ts
```diff
@@ -146,20 +148,36 @@ class TfStore {
...
   has(key: string): boolean {
-    return this._storage.has(key);
+    return this._storage.has(stripLeadingSlash(key));
   }
 
   values(): Array<Transform> {
     return Array.from(this._storage.values());
   }
+
+  default(): Transform | undefined {
```
defunctzombie, line 158: comment this method to describe what the _default_ tf means

jhurliman:
This should resolve issues such as https://github.com/cruise-automation/webviz/issues/162

PR #443, opened 2021-04-01
Close webpack after compiling for integration testing

jhurliman:
This should fix the integration test not cleanly exiting in some cases


defunctzombie:
Good find!

defunctzombie (approved):

PR #444, opened 2021-04-01
Add support for tf/tfMessage

jhurliman:
In addition to tf2 messages, transforms can now be visualized for the older `tf/tfMessage` transforms.


app/panels/ThreeDimensionalViz/SceneBuilder/defaultHooks.ts
```diff
@@ -22,7 +22,7 @@ const sceneBuilderHooks: ThreeDimensionalVizHooks = {
   getTopicsToRender: () => new Set(),
   consumeBobject: (topic, datatype, msg, consumeMethods, { errors }) => {
     // TF messages are consumed by TransformBuilder, not SceneBuilder.
```
jtbandes, line 24: where does the consumption actually happen? It doesn't seem to be in this commit

jhurliman: The new message is identical to tf2 so the existing consumption, wherever it is, already works

jtbandes: I guess it wasn't predicated on datatype? 🤔

PR #457, opened 2021-04-03
Integrate @fluentui/react

jtbandes:
Integrate the `@fluentui/react` UI component library (FG-187). This library was chosen because of:
- being written in TypeScript
- visually appealing components without excessive animation
- comprehensive set of components including draggable modals, menus, and overlays
- active development by Microsoft (although the community is smaller than material-ui and others)
- pretty good documentation on type definition “headers”
- pretty good component examples at https://developer.microsoft.com/en-us/fluentui#/controls/web
- good themeability (with theme provider), with designer tool at http://aka.ms/themedesigner

This initial PR sets up some plumbing and uses the new components for the connection picker, layout menu, and prompt.

- Disable loading of default fonts from Microsoft servers. This is done by configuring an empty fontBaseUrl (can't be done programmatically: https://github.com/microsoft/fluentui/issues/3881; no sanctioned way to disable loading: https://github.com/microsoft/fluentui/issues/10363).
- Add a basic theme and ThemeProvider that also manages delivering SVG icons from `@fluentui/react-icons-mdl2` (MIT-licensed, unlike the default font icons). Expose the theme to styled-components as well.
- Add a `ModalContext` and `ModalHost` allowing us to render modals inside the ThemeProvider rather than attaching them to `document.body`. Modifies `usePrompt` to render into the modal context.

Screen Shot 2021-04-02 at 6 03 15 PM Screen Shot 2021-04-02 at 6 03 30 PM image 
image


app/components/TinyConnectionPicker.tsx
```diff
@@ -11,75 +11,69 @@
...
-  PlayerSourceDefinition,
-  usePlayerSelection,
-} from "@foxglove-studio/app/context/PlayerSelectionContext";
+import { usePlayerSelection } from "@foxglove-studio/app/context/PlayerSelectionContext";
 
 type TinyConnectionPickerProps = {
   defaultIsOpen?: boolean;
 };
 
+const RosIcon = createSvgIcon({
```
defunctzombie, line 25: Extract to component in components folder - might be nice to have around

app/components/LayoutMenu.tsx
```diff
@@ -132,26 +152,92 @@ export default function LayoutMenu(props: { isOpen?: boolean }) {
...
+
+    dispatch(loadLayout(newState));
+  }, [dispatch, store]);
+
+  const layoutItems: IContextualMenuItem[] = useMemo(() => {
+    if (loading || layouts === undefined) {
+      return [];
+    }
+
+    // Panel state may not have an ID yet. To make highlighting current layout work we need the ID.
```
defunctzombie, line 179: Is this comment no longer relevant?

defunctzombie:
I've pushes a change to the background darkness for modals. In the commit I've described why. Unfortunately there is no way to set this at the theme level - probably an enhancement we should propose.

Few other things after playing around with the UI and looking into some of the changes:
* Lets fix the icon/text menu alignment - right now it does not look polished to have a slight offset. Breaks the otherwise pleasant UI experience.
* Put back icons for rename/delete - there is too much menu nesting happening to rename or delete a layout.

How this framework loads icons is potentially problematic - it seems that you pass the _text_ of the icon name and some other layer resolves tis and loads it? Presumably that means it has all the icons pre-loaded? Most frameworks I'm use to in the react space have you pass either icon components or svg elements. What happens if you pass text of an icon that doesn't exist? Will we get a build-time error or a nice missing icon box during runtime for our users?

app/components/LayoutMenu.tsx
jtbandes: I think it is still relevant, may need to do some more testing

jtbandes:
The available icons are explicitly registered by us (see `useIcons()`). If one is missing it would currently be a console warning and an empty space. I suspect we can leverage either TS declaration merging or an ESLint rule to enforce that only known names are used.

It is possible to pass an icon component/rendering function instead of the name, but the name seems like the most natural API they have and is nice that you don’t have to clutter your imports with icons 🤷

Normally it would be used with an icon font but it seems only their SVG icons are MIT-licensed.

jtbandes:
Filed https://github.com/microsoft/fluentui/issues/17694 for theme override of isDark. However I think we can set this via the blackTranslucent40 palette entry as well.

PR #463, opened 2021-04-04
Remember url for roscore/rosbridge/remote bag sources

defunctzombie:
Automatically load the last source the user selected when the app opens. For
sources which support saving their user selected endpoint (i.e. url), restore
this value and try to connect the source on startup.

### DIFF
--- a/app/components/PlayerManager.tsx
+++ b/app/components/PlayerManager.tsx
@@ -145,7 +145,8 @@ type FactoryOptions = {
 async function localBagFileSource(options: FactoryOptions) {
   let file: File;
 
-  // local bag file sources cannot be stored yet
+  // future enhancement woudl be to store the fileHandle in indexeddb and try to restore
+  // fileHandles can be stored in indexeddb but not localstorage
   if (!options.skipRestore) {
     return;
   }
@@ -167,7 +168,7 @@ async function localBagFileSource(options: FactoryOptions) {
 }
 
 async function remoteBagFileSource(options: FactoryOptions) {
-  const storageCacheKey = "studio.source.remotebag";
+  const storageCacheKey = `studio.source.${options.source.name}`;
 
   // undefined url indicates the user canceled the prompt
   let maybeUrl;
@@ -204,7 +205,7 @@ async function remoteBagFileSource(options: FactoryOptions) {
 }
 
 async function rosbridgeSource(options: FactoryOptions) {
-  const storageCacheKey = "studio.source.rosbridge";
+  const storageCacheKey = `studio.source.${options.source.name}`;
 
   // undefined url indicates the user canceled the prompt
   let maybeUrl;
@@ -246,7 +247,7 @@ async function rosbridgeSource(options: FactoryOptions) {
 }
 
 async function roscoreSource(options: FactoryOptions) {
-  const storageCacheKey = "studio.source.roscore";
+  const storageCacheKey = `studio.source.${options.source.name}`;
 
   // undefined url indicates the user canceled the prompt
   let maybeUrl;
--- a/app/components/PlayerManager.tsx
+++ b/app/components/PlayerManager.tsx
@@ -135,188 +135,157 @@ async function buildPlayerFromBagURLs(
   throw new Error(`Unsupported number of urls: ${urls.length}`);
 }
 
-function useLocalBagFileSource() {
-  const build = useCallback(
-    async (options: BuildPlayerOptions, additionalOpt?: { skipRestore: boolean }) => {
-      let file: File;
+type FactoryOptions = {
+  source: PlayerSourceDefinition;
+  skipRestore: boolean;
+  prompt: ReturnType<typeof usePrompt>;
+  storage: Storage;
+};
 
-      // local bag file sources cannot be stored yet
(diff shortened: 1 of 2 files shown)


defunctzombie:
@jhurliman you had mentioned concern around the app loading a remote bag automatically. This implementation would do that - but I could tweak the "restore" functionality for remote bags to pre-populate the prompt with the user's value but not automatically start streaming the bag file.

app/components/PlayerManager.tsx
```diff
@@ -163,92 +157,166 @@ async function getPlayerBuilderFromUserSelection(
...
-            );
-          }
-          return result;
-        },
-      });
-      if (url == undefined) {
+
+      const url = maybeUrl;
+      storage.setItem(storageCacheKey, url);
+      return () => buildPlayerFromBagURLs([url], options);
```
jtbandes, line 208: Is there a way we can clean up by doing the `getItem`/`setItem` in one place rather than separately in each source hook?

app/components/PlayerManager.tsx
```diff
@@ -163,92 +157,166 @@ async function getPlayerBuilderFromUserSelection(
...
+    [],
+  );
+
+  return build;
+}
+
+function useRemoteBagFileSource() {
+  const prompt = usePrompt();
+
+  const storageCacheKey = "studio.source.remotebag";
```
jtbandes, line 172: Is there a need for this key to be different from `PlayerSourceDefinition.type`?

app/components/PlayerManager.tsx
```diff
@@ -131,26 +135,16 @@ async function buildPlayerFromBagURLs(
...
-  options.metricsCollector.setProperty("player", definition.type);
-
-  switch (definition.type) {
-    case "file": {
+function useLocalBagFileSource() {
+  const build = useCallback(
+    async (options: BuildPlayerOptions, additionalOpt?: { skipRestore: boolean }) => {
       let file: File;
+
+      // local bag file sources cannot be stored yet
```
jtbandes, line 143: FileSystemHandle can be saved in IndexedDB: https://web.dev/file-system-access/#storing-file-handles-or-directory-handles-in-indexeddb

app/components/PlayerManager.tsx
```diff
@@ -356,25 +423,90 @@ function PlayerManager({
...
+  );
+
+  const [selectedSource, setSelectedSource] = useState<PlayerSourceDefinition | undefined>(
+    previousSelectedSource,
+  );
+
+  const localBagFileSourceAction = useLocalBagFileSource();
+  const roscoreSourceAction = useRoscoreSource();
+  const rosbridgeSourceAction = useRosbridgeSource();
+  const remoteBagSourceAction = useRemoteBagFileSource();
```
jtbandes, line 437: Thoughts on rather than making these hooks, just passing the `prompt` and `storage` as arguments into the actions when called? Not everything always needs to be a hook IMO :)

app/components/PlayerManager.tsx
```diff
@@ -356,25 +423,90 @@ function PlayerManager({
...
+
+  // persist the selected source to storage
+  useEffect(() => {
+    saveSelectedSource(selectedSource);
+  }, [saveSelectedSource, selectedSource]);
```
jtbandes, line 442: The four `previousSelectedSource/saveSelectedSource/selectedSource/setSelectedSource` are somewhat confusing to me — why not just directly use the setter returned by useLocalStorage? It seems it takes care of triggering a component update for you.

app/components/PlayerManager.tsx
defunctzombie: Each source hook might want to restore state in a different way (as we know for local file - which requires indexeddb rather than localstorage). I did think about how three of these are all basically the same flow just with different prompt settings - so maybe a function for those rather than copying that would be useful.

app/components/PlayerManager.tsx
defunctzombie: Yes - I will improve the comment but I am not doing that as part of this change since it requires making more decisions around idb management.

app/components/PlayerManager.tsx
defunctzombie: Yea - I think that will be the path I go for these. Originally I had imagined some other flow or possibly passing these from a different location - but for keeping this change simple I think that will be good for now.

I thought about moving some of this out of player manager since I think having our list of supported sources grow each self-contained in their own file makes more sense but we need to revisit that.

app/components/PlayerManager.tsx
defunctzombie: To differentiate when a user selects a source and when we first load a source. I think now that I added the ref for gating the first source selection I can go back to the useLocalStorage being the source of truth. This went through a few iterations with useLocalStorage being the source of truth until I realized I needed to differentiate manually selecting an entry and showing a prompt (otherwise you would never be able to change the restored value).

app/components/PlayerManager.tsx
defunctzombie: The type is a more limited set - we could have different sources both use the same _type_ - tho I think thats actually not true with the way we have the _switch_ statement setup currently so probably should make it the type.

jtbandes (approved):

PR #464, opened 2021-04-05
Don't initialize telemetry until `telemetry.telemetryEnabled` setting has been read

jhurliman:
This prevents the APP_INIT telemetry event from being recorded if the user has set `telemetry.telemetryEnabled = false`


jhurliman:
Unnecessary now that `telemetryEnabled` is passed into argv of the BrowserWindow

PR #478, opened 2021-04-06
Replace GlobalConfig with PanelCatalogContext and BuiltinCatalogProvider

defunctzombie:
Rather than having all panels import global config, put the default configuration into each panel and provide the list of panels via a Panel Catalog Context. This moves us towards the goal of loading user panels and helps treat panels as separate components.


app/panels/PanelList/index.tsx
```diff
@@ -293,16 +280,29 @@ function PanelList(props: Props) {
...
-  const getFilteredItemsForCategory = React.useCallback(
-    (key: string) =>
-      searchQuery
+  const panelCategories = React.useMemo(() => {
+    return [
+      { label: "ROS", key: "ros" },
+      { label: "Utilities", key: "utilities" },
+      { label: "Debugging", key: "debugging" },
+    ];
+  }, []);
```
jtbandes, line 289: Shouldn't this come from the usePanelCatalog?

app/panels/ThreeDimensionalViz/CameraInfo/index.tsx
```diff
@@ -267,11 +266,7 @@ export default function CameraInfo({
                   </SValue>
                 </SRow>
               ) : (
-                <p>
-                  Locked to map (
-                  {(getGlobalHooks() as any).perPanelHooks().ThreeDimensionalViz.rootTransformFrame}
-                  )
-                </p>
+                <p>Locked to map</p>
```
jtbandes, line 269: Ideally this would say the correct frame name...not sure what that is at first glance

app/panels/ThreeDimensionalViz/WorldMarkers.tsx
```diff
@@ -231,7 +227,7 @@ export default function WorldMarkers({
         {pointcloud as any}
       </PointClouds>
       <Arrows layerIndex={layerIndex}>{arrow}</Arrows>
-      <Points layerIndex={layerIndex} useWorldSpaceSize={useWorldspacePointSize}>
+      <Points layerIndex={layerIndex} useWorldSpaceSize>
```
jtbandes, line 230: I suspect this should be false by default so that points remain a fixed pixel size rather than scaling up (just guessing what this prop does by its name) - unfortunately we are lacking in 3D panel stories right now

app/panels/ThreeDimensionalViz/commands/LaserScans.tsx
```diff
@@ -21,15 +21,15 @@ import {
   nonInstancedGetChildrenForHitmap,
 } from "regl-worldview";
 
-import { getGlobalHooks } from "@foxglove-studio/app/loadWebviz";
+import LaserScanVert from "@foxglove-studio/app/panels/ThreeDimensionalViz/LaserScanVert";
```
jtbandes, line 24: Feel free to inline this imported string below. It's not too long and would be nice to see along the fragment shader.

app/services/BuiltinPanelCatalog.ts
```diff
@@ -0,0 +1,98 @@
...
+
+    // allow hidden panels to be referenced by type but not appear in panels by category
+    panelsByCategory.set("hidden", hidden);
+
+    const allPanels = flatten([...panelsByCategory.values()]);
+    const panel = allPanels.find((item) => item.component.panelType === type);
+    return panel?.component;
+  }
+
+  getPanelsByType(): Map<string, PanelInfo> {
```
jtbandes, line 83: Should we just build these maps up front rather than providing getter functions to build them on demand?

app/panels/ThreeDimensionalViz/CameraInfo/index.tsx
defunctzombie: This is the fallback for when followTf is not present. Unfortunately the actual frame is not passed to this component.

app/panels/ThreeDimensionalViz/WorldMarkers.tsx
defunctzombie: The current version always returns true https://github.com/foxglove/studio/search?q=useWorldspacePointSize which aligns with the change here.

PR #481, opened 2021-04-06
Fix inScreenshotTests() for Chromatic

jtbandes:


defunctzombie:
The real fix is to delete `inScreenshotTests`

PR #484, opened 2021-04-06
Move GLSL shader code to separate .vert / .frag files

jhurliman:


jhurliman:
This would require a larger refactor to move the `${...}` template literals to shader uniforms. Closing for now.

PR #496, opened 2021-04-07
Modify sample-robot-docker to work on MacOS

jhurliman:
Instead of using host mode networking, we switch back to bridge mode and bind a fixed range of about 50 ports. TCPROS servers are forced to bind in this range by overriding the sysctl `net.ipv4.ip_local_port_range`


amacneil:
Clever. So there is no other way to tell the ros master which ports to use, besides `net.ipv4.ip_local_port_range`?

PR #509, opened 2021-04-07
Introduce a ScreenshotContext for stories to signal when the scene is ready.

defunctzombie:
Stories which test several interactions or have delayed rendering are not always
ready at a fixed delay from mounting. The scene ready signal provides the story author
more control over when a screenshot can be taken of the story.

### DIFF
--- a/app/.storybook/preview.tsx
+++ b/app/.storybook/preview.tsx
@@ -1,41 +1,14 @@
-import "@foxglove-studio/app/styles/global.scss";
-import "./styles.scss";
 import waitForFonts from "@foxglove-studio/app/util/waitForFonts";
 import { withScreenshot } from "storycap";
 import { withMockSubscribeToNewsletter } from "./__mocks__/subscribeToNewsletter";
 import { Story, StoryContext } from "@storybook/react";
 import ThemeProvider from "@foxglove-studio/app/theme/ThemeProvider";
-import signal from "@foxglove-studio/app/shared/signal";
-import ScreenshotContext from "@foxglove-studio/app/stories/ScreenshotContext";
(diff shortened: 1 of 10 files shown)


app/.storybook/preview.tsx
```diff
@@ -5,9 +5,38 @@ import { withScreenshot } from "storycap";
...
+import ScreenshotContext from "@foxglove-studio/app/stories/ScreenshotContext";
+import { useCallback, useRef, useState } from "react";
 
 let loaded = false;
 
+function withScreenshotSignal(Story: Story, { parameters }: StoryContext) {
+  const signalRef = useRef(signal());
+  const callCount = useRef(0);
+  const [error, setError] = useState<Error | undefined>(undefined);
+  parameters.screenshot.waitFor = signalRef.current;
```
jtbandes, line 18: error if waitFor was already set?

app/stories/ScreenshotContext.ts
```diff
@@ -0,0 +1,21 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { createContext, useContext } from "react";
+
+// indicate that screenshot is ready for capture
+type ScreenshotSignal = () => void;
+
+const ScreenshotContext = createContext<ScreenshotSignal | undefined>(undefined);
+
+export function useScreenshot(): ScreenshotSignal {
```
jtbandes, line 12: naming: useScreenshotReady? useScreenshotSignal?

app/.storybook/preview.tsx
defunctzombie: I decided to use the existing waitFor if it is set

PR #510, opened 2021-04-07
Disable asar packaging

jtbandes:
Workaround for https://github.com/electron/electron/issues/28572. Fixes SharedWorkers; may also help with source maps.

With asar disabled, a bug in `file` is bubbling up through `@electron/universal`, so we have to use a forked version with the fix from https://github.com/electron/universal/pull/16.

### DIFF
--- a/electron-builder.json
+++ b/electron-builder.json
@@ -2,6 +2,7 @@
   "$schema": "http://json.schemastore.org/electron-builder",
   "appId": "dev.foxglove.studio",
   "npmRebuild": false,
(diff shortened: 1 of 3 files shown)


electron-builder.json
```diff
@@ -2,7 +2,7 @@
   "$schema": "http://json.schemastore.org/electron-builder",
   "appId": "dev.foxglove.studio",
   "npmRebuild": false,
-  "asar": true,
+  "asar": false,
```
defunctzombie, line 5: Can we use `//` key/value trick and comment why we did this?

PR #511, opened 2021-04-08
Upgrade regl-worldview for webglcontextload/restore support

defunctzombie:
Fixes an issue when connecting/disconnecting an external display which
would result in the 3d panel becoming broken and requiring an app restart.


defunctzombie:
@amacneil plz merge once you have tested to your satisfaction

amacneil:
Current main after plugging and replugging external monitor (with laptop lid closed the entire time):

image

This branch:

image

🎉

PR #526, opened 2021-04-08
Revert "Upgrade regl-worldview for webglcontextload/restore support"

jhurliman:
Reverts foxglove/studio#511. This PR broke running `yarn` on Windows.


defunctzombie:
We should find a way to include the changes in this commit - not revert it.

PR #527, opened 2021-04-08
A map panel is born

defunctzombie:
Screen Shot 2021-04-10 at 9 42 03 AM

The map panel plots topics with datatype [sensor_msgs/NavSatFix](https://docs.ros.org/en/api/sensor_msgs/html/msg/NavSatFix.html) onto a zoomable/pannable street map.

For bag files, the panel plots all of the messages present in the bag file so the entire path is shown. For streaming, the panel saves the points currently seen so path history is shown.


amacneil:
Awesome! Tested with some local bags and works great.

image

amacneil:
Is there a topic selector? Or does it just assume there is only one topic with `NavSatFix` and use that?

defunctzombie:
> Is there a topic selector? Or does it just assume there is only one topic with `NavSatFix` and use that?

It plots all topics that are `NavSatFix`. A future enhancement will be to allow topic selection and color change (like in the 3d panel) but I did not tackle that for the MVP here.

PR #531, opened 2021-04-09
Re-enable ASAR packaging with NotActuallySharedWorker

jtbandes:
Disabling asar to fix SharedWorker has opened up a pandora's box of other issues (https://github.com/electron/universal/pull/16 and https://github.com/electron-userland/electron-builder/issues/460). I would like to address these so we can use SharedWorkers again, but for now we need a stopgap so we can make builds and not lose Node Playground functionality. So here it is: the worst code ever written

### DIFF
--- a/app/players/UserNodePlayer/index.ts
+++ b/app/players/UserNodePlayer/index.ts
@@ -592,7 +592,6 @@ export default class UserNodePlayer implements Player {
   }
 
   close = () => {
-    console.log("close usernodeplayer");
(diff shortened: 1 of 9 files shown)


app/players/UserNodePlayer/index.ts
```diff
@@ -586,6 +592,7 @@ export default class UserNodePlayer implements Player {
   }
 
   close = () => {
+    console.log("close usernodeplayer");
```
jhurliman, line 595: Change to `log.debug` or remove?

PR #532, opened 2021-04-09
Support ROS topics that do not start with `/`

jhurliman:
Simple grammar file fix for message path parsing


jhurliman:
This fixes another issue loading the Ouster sample bags linked to from https://github.com/cruise-automation/webviz/issues/162

PR #533, opened 2021-04-09
Move app/shared files to app/util

defunctzombie:
app/shared is not a concept for us


amacneil:
Everything is a util 🎉

PR #534, opened 2021-04-09
Split app/util/hooks.tsx into separate hook files

defunctzombie:
Separate hook files are easier to isolate and maintain


app/components/MessagePathSyntax/useMessagesByPath.ts
```diff
@@ -1,6 +1,16 @@
 // This Source Code Form is subject to the terms of the Mozilla Public
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
+//
```
defunctzombie, line 4: @jtbandes is this file studio license only?

PR #536, opened 2021-04-09
Fix some yarn missing package warnings

amacneil:
Nothing new here, just some missing peer dependencies.


app/package.json
```diff
@@ -127,11 +128,14 @@
...
     "@wojtekmaj/enzyme-adapter-react-17": "0.6.0",
     "argparse": "2.0.1",
+    "babel-loader": "8.2.2",
     "chromatic": "5.7.1",
     "enzyme": "3.11.0",
     "fetch-mock": "9.11.0",
     "http-server": "0.12.3",
+    "react-is": "17.0.2",
     "url-search-params": "1.1.0",
+    "webpack": "5.31.0",
```
defunctzombie, line 138: Why is this in app/package.json? We already list webpack in package.json at the root - we should not list it twice.

app/package.json
```diff
@@ -85,6 +85,7 @@
     "wasm-lz4": "1.0.0"
   },
   "devDependencies": {
+    "@babel/core": "7.13.15",
```
defunctzombie, line 88: This shouldn't need to be in app/package.json

amacneil:
```
➤ YN0002: │ @foxglove-studio/app@workspace:app doesn't provide @babel/core (pe45b9), requested by @storybook/addon-essentials
➤ YN0002: │ @foxglove-studio/app@workspace:app doesn't provide babel-loader (p2a7fd), requested by @storybook/addon-essentials
```

Would you prefer to move storybook to the top level?

app/package.json
amacneil:
```
➤ YN0002: │ @foxglove-studio/app@workspace:app doesn't provide webpack (p7bc32), requested by babel-loader
```

babel-loader is also being requested by storybook

app/package.json
defunctzombie: yea - I guess so for now - we should only have 1 storybook version for the monorepo

PR #539, opened 2021-04-09
Inform rosmaster when unsubscribing from a topic

jhurliman:
This will keep the rosmaster connectivity graph accurate

### DIFF
--- a/packages/ros1/src/RosNode.ts
+++ b/packages/ros1/src/RosNode.ts
@@ -36,6 +36,7 @@ export class RosNode extends EventEmitter {
   private _tcpSocketCreate: TcpSocketCreate;
   private _connectionIdCounter = 0;
   private _tcpServer?: TcpServer;
+  private _localApiUrl?: string;
 
   constructor(options: {
     name: string;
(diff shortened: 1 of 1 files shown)


packages/ros1/src/RosNode.ts
```diff
@@ -173,6 +186,25 @@ export class RosNode extends EventEmitter {
...
+        this.name,
+        topic,
+        callerApi,
+      );
+
+      if (status !== 1) {
+        throw new Error(`unregisterSubscriber() failed. status=${status}, msg="${msg}"`);
+      }
+    } catch (_err) {
+      // TODO: Log this warning
```
defunctzombie, line 204: can we do this TODO now? I am not a fan of TODOs since they largely go left undone.

defunctzombie: Or maybe throw this to the app and let it handle it?

jhurliman: There is no path back to the app since we do this asynchronously without waiting (it's an informative "by the way" call to rosmaster, not something we block on) and the context could get lost about whether this is something that should be surfaced to the user or not (IMO it should not). The real fix is to implement the logger dependency injection, which I can add to this PR.

jhurliman: I'm going to merge this now so I can start on logging as a clean PR

PR #540, opened 2021-04-09
Add Preferences window and time zone setting [FG-239]

jtbandes:
Move ExperimentalFeatures into Preferences window in a separate pane. Add time zone setting to preferences, which changes the time zone used in Raw Messages & 3D panel.

https://user-images.githubusercontent.com/14237/114233435-04e0b700-9919-11eb-8cb3-ba8791247a41.mov


app/components/Preferences.tsx
```diff
@@ -0,0 +1,156 @@
...
+  const theme = useTheme();
+  return (
+    <Pivot>
+      <PivotItem headerText="Settings" style={{ padding: theme.spacing.m }}>
+        <Stack.Item>
+          <TimezoneSettings />
+        </Stack.Item>
+      </PivotItem>
+      {/*
+      TODO(john): hook these up to actual app settings
```
jtbandes, line 132: FYI @jhurliman

PR #543, opened 2021-04-09
Add logging support to @foxglove/ros1

jhurliman:
Also fixes url() method for node.js HttpServer and unregisters subscribers during shutdown().


packages/ros1/src/LoggerService.ts
```diff
@@ -0,0 +1,10 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+export interface LoggerService {
+  warn(message: string, context?: string): void;
```
defunctzombie, line 6: A little weird that all the others are optional. Maybe they should all be optional.

PR #552, opened 2021-04-12
Conditionally call getUserDefault

jtbandes:
This function is only available on darwin.


jtbandes:
Proposed https://github.com/electron/typescript-definitions/pull/183 to help others avoid this issue.

PR #553, opened 2021-04-12
Replace PopperJS with FluentUI Tooltip

jtbandes:
Remove Tooltip.show()/hide() imperative APIs in favor of new props, `shown`+`targetPosition` which can be used to show the tooltip at a specific position by mounting a component. This was only used in the playback bar and for Plots.

Also removes ContextMenu.show which was a wrapper around Tooltip.show() and replace this with a real ContextualMenu in the image panel.

Introduce a `useTooltip` hook rather than using FluentUI's TooltipHost to manage state, because TooltipHost renders an extra `div` which messes up our layout and makes it impossible to attach tooltips in "fragile" DOM trees such as on a table cell.

Some cleanup to BaseButton allowing it to be used as `useTooltip().ref`. Remove `delay` functionality which was unused (originally introduced for deleting saved layouts).


app/components/PlaybackControls/index.tsx
```diff
@@ -41,8 +41,8 @@ import {
 } from "@foxglove-studio/app/components/PlaybackControls/sharedHelpers";
 import PlaybackSpeedControls from "@foxglove-studio/app/components/PlaybackSpeedControls";
 import Slider from "@foxglove-studio/app/components/Slider";
-import tooltipStyles from "@foxglove-studio/app/components/Tooltip.module.scss";
-import Tooltip from "@foxglove-studio/app/components/TooltipBase";
+// import tooltipStyles from "@foxglove-studio/app/components/Tooltip.module.scss";
```
defunctzombie, line 44: delete

app/components/PlaybackControls/index.tsx
```diff
@@ -235,6 +234,13 @@ export const UnconnectedPlaybackControls = memo<PlaybackControlProps>(
 
     return (
       <Flex row className={styles.container}>
+        {tooltip && (
```
defunctzombie, line 237: unrelated but I always felt a little let down that there wasn't a nicer way to conditionally render components /rant

app/components/Tooltip.tsx
```diff
@@ -11,50 +11,158 @@
 //   found at http://www.apache.org/licenses/LICENSE-2.0
 //   You may not use this file except in compliance with the License.
```
amacneil, line 12: should be fine to remove the cruise copyright if you completely rewrite 90%+ of a file like this

Issue #495, opened 2021-04-07
Can't load panels 


iceuw:
System: Windows 10
Describe: After launch program, all panels disappear after a flash
![屏幕截图 2021-04-07 121452](https://user-images.githubusercontent.com/23731407/113809523-e6ed4800-979a-11eb-9b41-0a96bf7cdd0f.png)

amacneil:
That's very strange. Which version are you using? Can you try v0.4.0 we released today?

iceuw:
I tried v0.4.0. And then there is nothing remains in the program window. And I tried the v0.3.0 with the Console in the Debug. It shows this error
![屏幕截图 2021-04-07 131121](https://user-images.githubusercontent.com/23731407/113813890-536c4500-97a3-11eb-81c5-89938484d134.png)

iceuw:
And I found this file called "app.asar". But the "app.asar" is a file ranther then a directory. Is there any wrong?

amacneil:
Ok thanks. No, the `app.asar` is just how we package the app, and those `failed to load SourceMap` warnings you see are also normal and not causing the error.

I think I have an idea what is causing this, let me get back to you with troubleshooting steps.

thoys:
I've got the same issue myself

amacneil:
Are either of you using a non-English language setup in Windows? We're seeing some `DWriteFont::Create` errors which might be related, but unclear what the cause of them is.

PR #557, opened 2021-04-13
Load the Roboto Mono font on chart.js worker thread explicitly

jhurliman:
A large portion of Windows users are affected by a rendering thread crash in [skia](https://skia.org/) that seems to happen when chart.js renders a custom font to a <canvas> element when the system display scaling is set >100%.

This PR directly fixes a related issue, the Roboto Mono font is specified for chart axis labels but is not actually available in the worker thread where `ctx.fillText()` is called. By waiting for the custom font to be loaded before rendering the chart, it also appears to fix the DirectWrite font crash in my testing so far.

Fixes https://github.com/foxglove/studio/issues/495

### DIFF
--- a/app/components/Chart/index.stories.tsx
+++ b/app/components/Chart/index.stories.tsx
@@ -71,7 +71,7 @@ const props: ComponentProps<typeof ChartComponent> = {
       y: {
         ticks: {
           font: {
-            family: `"Roboto Mono", Menlo, Inconsolata, "Courier New", Courier, monospace`,
+            family: `"Roboto Mono"`,
             size: 10,
           },
           color: "#eee",
@@ -86,7 +86,7 @@ const props: ComponentProps<typeof ChartComponent> = {
       x: {
         ticks: {
           font: {
-            family: `"Roboto Mono", Menlo, Inconsolata, "Courier New", Courier, monospace`,
+            family: `"Roboto Mono"`,
             size: 10,
           },
           color: "#eee",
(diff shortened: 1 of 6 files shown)


app/components/Chart/index.stories.tsx
```diff
@@ -71,7 +71,7 @@ const props: ComponentProps<typeof ChartComponent> = {
       y: {
         ticks: {
           font: {
-            family: `"Roboto Mono", Menlo, Inconsolata, "Courier New", Courier, monospace`,
```
defunctzombie, line 74: Should we leave "monospace" as a fallback?

app/components/Chart/worker/ChartJSManager.ts
```diff
@@ -40,26 +44,38 @@ function removeEventListener(emitter: EventEmitter) {
   };
 }
 
+// Explicitly load the "Roboto Mono" font, since custom fonts from the main renderer are not
+// inherited by web workers
```
defunctzombie, line 48: Consider copying the comment you have on the PR about the crash as additional reference here. To me it adds to the importance and severity around the fix.

app/components/Chart/worker/ChartJSManager.ts
```diff
@@ -40,26 +44,38 @@ function removeEventListener(emitter: EventEmitter) {
...
-  private _lastDatalabelClickContext?: DatalabelContext = undefined;
+  private _lastDatalabelClickContext?: DatalabelContext;
+
+  constructor(initOpts: InitOpts) {
+    log.info(`new ChartJSManager(id=${initOpts.id})`);
+    this.init(initOpts);
+  }
+
+  async init({ id, node, type, data, options, devicePixelRatio }: InitOpts): Promise<void> {
+    const font = await loadDefaultFont();
```
defunctzombie, line 76: This will load the font for every manager. We only need to load the font once for every worker. We can move this into ChartJSMux which sits above manager and dispatches to the different managers. Then we would load it once per worker rather than once per chart.

app/components/Chart/index.stories.tsx
jhurliman: We load `Roboto Mono` as inline base64, so if it fails to load it's either a programming error or something fairly serious. I can add a comment describing that plus the `monospace` fallback just to be defensive.

PR #569, opened 2021-04-13
Lint files in .storybook

jtbandes:
eslint ignores dot-files and folders by default.


amacneil:
why is this folder hidden anyway? it has genuine code in it

jtbandes:
That's the standard way of configuring storybook: https://storybook.js.org/docs/react/configure/overview

amacneil:
Seems kinda silly :)

PR #574, opened 2021-04-14
Send telemetry events to Sentry as breadcrumbs when crash reporting is enabled

jhurliman:
This change reports HTTP requests, responses, and user interaction events as Sentry.io breadcrumbs when crash reporting is enabled. It also uses consistent userIds and deviceIds across Amplitude and Sentry, which may improve the unique users affected per error metrics.


amacneil:
Should we also unify sentry breadcrumbs with amplify events? Or do you think they would have different uses.

PR #576, opened 2021-04-14
Opt in/out of telemetry and crash reporting from the UI

jhurliman:
Also fixes a pathing issue to make sure we read these preferences from the correct file

### DIFF
--- a/app/components/Preferences.tsx
+++ b/app/components/Preferences.tsx
@@ -2,12 +2,14 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 import {
+  Checkbox,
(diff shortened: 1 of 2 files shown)


app/components/Preferences.tsx
```diff
@@ -121,33 +123,40 @@ function TimezoneSettings(): React.ReactElement {
...
-          />
-          <Checkbox
-            checked={crashReportingEnabled}
-            onChange={(event, checked) => setCrashReportingEnabled(checked)}
-            label="Send anonymized crash reports"
-          />
-        </Stack>
-      </PivotItem>
-      */}
+      {
```
jtbandes, line 141: nit: these `{}` can be removed, were only necessary (or just added automatically) for the comment

PR #577, opened 2021-04-14
Restrict `allowDangerousHtml` to specific parts of the UI

jhurliman:
allowDangerousHtml was set to true recently to allow HTML formatting in Markdown help pages for panels. This PR rolls the setting back from being universally enabled to conditionally enabling in a few parts of the app, to prevent a future problem if we ever start rendering Markdown from user input (such as embedded in a bag).


amacneil:
We shouldn't need this for rendering markdown either - are we using https://github.com/remarkjs/react-markdown (I used it for public website)

jhurliman:
> We shouldn't need this for rendering markdown either - are we using https://github.com/remarkjs/react-markdown (I used it for public website)

We do need it. You can disable it unconditionally and see the screenshot test for the Topic Graph help page fail.

PR #585, opened 2021-04-14
Support TFs on all topics, not just /tf and /tf_static

jhurliman:
Previously we looked for magic topic names of `/tf` and `/tf_static` to collect all incoming transforms for the 3D panel. This switches from topic names to datatypes, subscribing to any topic with a supported TF datatype and extracting transforms each frame.

### DIFF
--- a/app/panels/ThreeDimensionalViz/Transforms.ts
+++ b/app/panels/ThreeDimensionalViz/Transforms.ts
@@ -149,6 +149,10 @@ class TfStore {
     return result;
   }
 
+  getMaybe(key: string): Transform | undefined {
+    return this._storage.get(stripLeadingSlash(key));
+  }
+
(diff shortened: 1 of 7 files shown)


app/panels/ThreeDimensionalViz/Transforms.ts
```diff
@@ -149,6 +149,10 @@ class TfStore {
     return result;
   }
 
+  getMaybe(key: string): Transform | undefined {
```
jtbandes, line 152: what if we call the other one getOrCreate?

jhurliman: That would be a helpful refactor. There's going to be a separate PR to actually fix all of these issues, this addresses one of who knows how many.

PR #601, opened 2021-04-16
Parameters panel

jhurliman:
parampanel

This PR implements a live view panel for rosparams and allows editing param values (calls `setParam` on the rosparam server).

### TODO
- [x] Make the list of params scrollable (is the global variables table scrollable?)
- [ ] Don't send param modifications on every keystroke, require the user to press a button
- [x] Confirm remote edits of rosparams are live updated in the panel with animation
- [x] Add a player capability for parameters, show a helpful message when the current player does not support params
- [ ] Sane handling for all possible param datatypes. Truncate long strings with a button to save to disk. Show a preview of UIn8tArray (like raw messages) with a button to save to disk.
- [x] Improve the help section a bit, mention that you can edit params (the UI does not make it obvious)


app/panels/RawMessages/index.tsx
```diff
@@ -269,10 +269,9 @@ function RawMessages(props: Props) {
           // sample output: Int8Array(331776) [-4, -4, -4, -4, -4, -4, -4, -4, -4, -4, -4, -4, -4, -4, -4, -4, -4, -4, -4, -4, ...]
           let smallNumberArrayStr = "";
           if (ArrayBuffer.isView(itemValue)) {
-            // @ts-expect-error whatever log is here is not correct for ArrayBufferView
-            const itemPart = itemValue.slice(0, DATA_ARRAY_PREVIEW_LIMIT).join(", ");
-            // @ts-expect-error whatever log is here is not correct for ArrayBufferView
-            const length = itemValue.length;
+            const array = itemValue as Uint8Array;
```
jtbandes, line 272: Seems sketchy too; the item may not be a Uint8Array (the example above shows an Int8Array)

app/panels/Parameters/index.tsx
```diff
@@ -0,0 +1,218 @@
...
+import { usePreviousValue } from "@foxglove-studio/app/hooks/usePreviousValue";
+import { ParameterValue, PlayerCapabilities } from "@foxglove-studio/app/players/types";
+import { colors as sharedColors } from "@foxglove-studio/app/util/sharedStyleConstants";
+
+import helpContent from "./index.help.md";
+
+// The minimum amount of time to wait between showing the parameter update animation again
+export const ANIMATION_RESET_DELAY_MS = 3000;
+
+const SScrollable = styled.div`
```
defunctzombie, line 36: Nit - I find the `S` hungarian notation prefix a bit "anti-pattern" esq. To me these are separate components `Scrollable`, `ParametersTable` and should live in their own files within your panel folder. The fact that they are mode using `styled.div` rather than `function()` is an implementation detail not relevant to the name of the component.

app/players/types.ts
```diff
@@ -31,6 +31,19 @@ export type ParsedMessageDefinitionsByTopic = {
...
+// Valid types for parameter data (such as rosparams)
+export type ParameterValue =
+  | undefined
+  | boolean
+  | number
+  | string
+  | Date
+  | Uint8Array
+  | ParameterValue[]
+  | ParameterStruct;
```
jtbandes, line 43:
```suggestion
  | Record<string, ParameterValue>;
```

app/panels/Parameters/index.tsx
```diff
@@ -0,0 +1,218 @@
...
+        <SParametersTable>
+          <table>
+            <thead>
+              <tr>
+                <th>Parameter</th>
+                <th>Value</th>
+              </tr>
+            </thead>
+            <tbody>
+              {parameterNames.map((name) => {
```
defunctzombie, line 184: Would be nice to hoist this to a component for readability

packages/ros1/src/RosNode.ts
```diff
@@ -61,10 +83,24 @@ export class RosNode extends EventEmitter {
...
+      const prevValue = this.parameters.get(key);
+      this.parameters.set(key, value);
+      this.emit("paramUpdate", key, value, prevValue, callerId);
+    });
+    this.rosFollower.on("publisherUpdate", (topic, publishers, callerId) => {
+      this.emit("publisherUpdate", topic, publishers, callerId);
+
+      // TODO: Subscribe/unsubscribe as necessary
+      // const sub = this.subscriptions.get(topic);
+      // if (sub != undefined) {}
```
jtbandes, line 102: is this important to do in this PR as a performance measure?

app/players/RandomAccessPlayer.ts
```diff
@@ -618,6 +619,15 @@ export default class RandomAccessPlayer implements Player {
     // no-op
   }
 
+  setParameter(key: string, _value: ParameterValue): void {
```
defunctzombie, line 622: Do players have a way of indicating they don't support setting parameters? Then the UI can disable editing the fields while still displaying the parameters.

packages/ros1/src/RosXmlRpcClient.ts
```diff
@@ -32,4 +32,39 @@ export class RosXmlRpcClient {
     }
     return res as RosXmlRpcResponse;
   };
+
+  protected _multiMethodCall = async (
```
jtbandes, line 36: could use a comment....and tests!

Is this a ROS-specific thing or should it be moved into XmlRpcClient?

packages/ros1/src/objectTests.ts
```diff
@@ -0,0 +1,41 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
defunctzombie, line 1: Please find a good npm module or make separate package? It would be tragic to end up with several copies of this same idea in our monorepo.

defunctzombie:
A few stories for the panel would be nice. Along with some tests for the newly added ros1 parameter code (to the extent that is possible).

packages/ros1/src/objectTests.ts
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  // Most likely a plain Object
+  return true;
+}
+
+// Returns true if an object was created by the Object constructor, Object.create(null), or {}, and
+// the object does not contain any enumerable keys.
+export function isEmptyPlainObject(o: unknown): boolean {
+  return isPlainObject(o) && Object.keys(o as Record<string, unknown>).length === 0;
```
jtbandes, line 40: Why do we care if `isPlainObject(o)`? What other kind of empty object would we get in the server response?

packages/ros1/src/RosNode.ts
```diff
@@ -61,10 +83,24 @@ export class RosNode extends EventEmitter {
...
+
+    this.rosFollower.on("paramUpdate", (key, value, callerId) => {
+      const prevValue = this.parameters.get(key);
+      this.parameters.set(key, value);
+      this.emit("paramUpdate", key, value, prevValue, callerId);
+    });
+    this.rosFollower.on("publisherUpdate", (topic, publishers, callerId) => {
+      this.emit("publisherUpdate", topic, publishers, callerId);
+
+      // TODO: Subscribe/unsubscribe as necessary
```
defunctzombie, line 100: I do not like TODOs in the code (they never get done) lets make a linear issue for any follow-up and for now update the code comment to indicate the tradeoff taken.

FWIW - if you use `fixme` instead of `todo` - lint will fail and be a good reminder to fix this before merge.

app/panels/RawMessages/index.tsx
jhurliman: It doesn't matter?

packages/ros1/src/RosNode.ts
jhurliman: No, this is topic handling which is orthogonal to rosparam handling. I added this as an assist for a future PR.

packages/ros1/src/RosXmlRpcClient.ts
jhurliman: Somewhere in between... it's an unofficial standard that sits on top of XML-RPC. A bit unclear whether it belongs in the XmlRpc library or not, although it seems like several other XML-RPC libs include it.

packages/ros1/src/objectTests.ts
jhurliman: This is needed to support a (IMO) poor decision in the rosparam server to return `{}` to signal "this parameter has not been set yet". Why they chose `{}` instead of `null` is difficult to understand, but that's how it is. So we need to check for this special value and reinterpret it as undefined for the end user to understand "this parameter has not been set yet".

jtbandes: How about:

```ts
(o.constructor === Object || o.constructor == undefined) && Object.keys(o).length === 0
```

PR #602, opened 2021-04-16
Remove magic app constants

amacneil:
Remove magic `APP_NAME` and `APP_VERSION` from Webpack DefinePlugin, since they were defined in 3 places (renderer/main/preload webpack configs) and sometimes causing production errors due to being defined in one file and not others.

Now they are normal typed constants imported directly.

### DIFF
--- a/app/constants.ts
+++ b/app/constants.ts
@@ -1,0 +1,9 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import packageJson from "../package.json";
+
+export const APP_NAME: string = packageJson.productName;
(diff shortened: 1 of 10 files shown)


app/constants.ts
```diff
@@ -0,0 +1,9 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import packageJson from "../package.json";
+
+export const APP_NAME = packageJson.productName;
+export const APP_VERSION = packageJson.version;
+export const APP_HOMEPAGE = packageJson.homepage;
```
jtbandes, line 9: are these typed as strings or `any`?

amacneil: strings. I'll add type to enforce

PR #608, opened 2021-04-18
Remove react-select

defunctzombie:
Update rosout panel to use fluentui components instead of react-select.


app/panels/Rosout/LogMessage.stories.tsx
```diff
@@ -0,0 +1,52 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import LogMessage from "./LogMessage";
+
+export default {
+  title: "panels/Rosout/LogMessage",
```
defunctzombie, line 8: This format for story titles creates a subheading and folder structure within storybook helping to organize the stories.

app/panels/Rosout/LogMessage.tsx
```diff
@@ -33,20 +33,29 @@ function Stamp(props: { stamp: Time }) {
   );
 }
 
+const classes = mergeStyleSets({
```
defunctzombie, line 36: @jtbandes is this the correct way to use merge styles?

app/theme/ThemeProvider.tsx
```diff
@@ -15,6 +15,7 @@ const icons: {
   [N in RegisteredIconNames]: React.ReactElement;
 } = {
   Add: <Icons.AddIcon />,
+  Cancel: <Icons.CancelIcon />,
```
defunctzombie, line 18: Until I added this the `x` icons didn't show up - it was kind of non-obvious which icon name I needed to add without looking at the fluent source code :/

jtbandes: Yep, I dunno if there’s any way to make this better unless we just go ahead and bundle/register *all* the icons.

The missing icon name should’ve been printed as a console warning though.

app/panels/Rosout/LogMessage.stories.tsx
jtbandes: Maybe we should make one big PR to reorganize all our stories? Our code has sort of a convention of including `<>` around the name but I don’t think there’s a good reason for it; we could remove those.

PR #609, opened 2021-04-18
Build typescript packages

amacneil:
- TypeScript packages under `packages/*` are individually compiled using `tsc`
- Correctly defined build graph using typescript [project references](https://www.typescriptlang.org/docs/handbook/project-references.html) (requires `tsc -b` flag which enables building dependencies with incremental compilation)
- Cleaned up `tsconfig` inheritance
- Cleaned up package.json files
- Pointed `"main"` to `.js` files in packages
- Enabled `projectReferences: true` in webpack config so that webpack knows to build dependencies via tsc (webpack watch mode behaves as expected when changing files in either packages or app)
- When importing packages, webpack uses the `.js` and `.d.ts`, which should speed up compilation (especially if more code moves into packages)

Future work
- Most packages are still using commonjs. I would like to attempt to move everything over to esm but will do so in a separate PR.


packages/README.md
```diff
@@ -0,0 +1 @@
+# packages directory
```
amacneil, line 1: TODO: explain how everything works here

webpack.renderer.config.ts
```diff
@@ -112,6 +112,7 @@ export function makeConfig(_: unknown, argv: WebpackArgv, options?: Options): Co
                 // https://github.com/TypeStrong/ts-loader#onlycompilebundledfiles
                 // avoid looking at files which are not part of the bundle
                 onlyCompileBundledFiles: true,
+                projectReferences: true,
```
amacneil, line 115: TODO: for some reason this correctly builds dependencies in main and preload, but not in renderer

preload/index.ts
```diff
@@ -35,6 +35,7 @@ type IpcListener = (ev: unknown, ...args: unknown[]) => void;
 const menuClickListeners = new Map<string, IpcListener>();
 
 // Initialize the RPC channel for electron-socket
+// FIXME: @jhurliman this returns a promise, should we have await here?
 PreloaderSockets.Create();
```
amacneil, line 39: @jhurliman should we be using `await` here?

packages/velodyne-cloud/src/perf/transform.ts
```diff
@@ -0,0 +1,22 @@
...
+import { Transformer } from "../Transformer";
+import { Model } from "../VelodyneTypes";
+import { HDL32E_PACKET1 } from "../fixtures/packets";
+
+const calibration = new Calibration(Model.HDL32E);
+const transform = new Transformer(calibration);
+
+for (let i = 0; i < 500; i++) {
+  const raw = new RawPacket(HDL32E_PACKET1);
+  const count = 0; // transform.validPoints(raw);
```
amacneil, line 17: @jhurliman when I converted this file to TS, it complained that this method doesn't exist. Can you confirm the correct behavior?

Also FYI I changed `count:` to `maxPoints:` in the line below, I assume that is correct?

amacneil:
Shelving for now, continued in https://github.com/foxglove/studio/pull/618

PR #617, opened 2021-04-20
URDF Viewer panel

jtbandes:
image

This is separate from the 3D panel because it takes advantage of existing `urdf-loader` library which depends on Three.js. We are inclined to work on converting the 3D panel to use Three.js in the future, so the URDF Viewer panel can be a testing ground for initial Three.js integration. Unfortunately this means that visualizing the robot model in world space with other markers, pointclouds, etc. is not yet possible.

Working:
- [x] Drag and drop a URDF file to view
  - [x] Look up mesh resources relative to the dropped file path
  - [x] Convert TIFF textures to PNG in main process since Chromium doesn't support TIFF
- [x] Load URDF from `/robot_description` param and resolve assets using `env.ROS_PACKAGE_PATH`
- [x] Sliders to change joint values (like joint_state_publisher_gui, although doesn't actually publish the values)
- [x] Some way of switching between models from `/robot_description` and dropped file(s?)
- [x] UI polish
- [x] Double-click to open URDF file

To do:
- [ ] (maybe a future PR) Buttons to reset / randomize joints (like joint_state_publisher)
- [ ] (maybe a future PR) draggable joints from https://github.com/gkjohnson/urdf-loaders/blob/master/javascript/src/URDFDragControls.js

Includes xacro-parser.d.ts as workaround for https://github.com/gkjohnson/xacro-parser/issues/54, and an `as any` as a workaround for https://github.com/gkjohnson/urdf-loaders/pull/204.


desktop/rosPackageResources.ts
```diff
@@ -0,0 +1,156 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+import { protocol } from "electron";
+import fs from "fs/promises";
```
defunctzombie, line 5: Is this the correct way to import this? I've always seen `import { promises as fs } from "fs"`. That keeps consistency with "fs" being the module.

desktop/index.ts
```diff
@@ -166,9 +175,9 @@ app.on("ready", async () => {
     "script-src": `'self' 'unsafe-inline' 'unsafe-eval'`,
     "worker-src": `'self' blob:`,
     "style-src": "'self' 'unsafe-inline'",
-    "connect-src": "'self' ws: wss: http: https:", // Required for rosbridge connections
+    "connect-src": "'self' ws: wss: http: https: x-foxglove-ros-package:", // Required for rosbridge connections
```
defunctzombie, line 178: `// required for rosbridge connections ` is no longer a clear comment on what is actually required.

jtbandes: I don't even know what it was originally referring to 😐

app/panels/URDFViewer/index.tsx
```diff
@@ -0,0 +1,383 @@
...
+  jointStatesTopic?: string;
+  customJointValues?: Record<string, number>;
+  opacity?: number;
+};
+type Props = {
+  config: Config;
+  saveConfig: SaveConfig<Config>;
+};
+
+function JointValueSliders({
```
defunctzombie, line 143: Separate into file - keep the file size manageable

PR #618, opened 2021-04-21
All packages inherit from tsconfig.base.json

amacneil:
- All tsconfig files inherit from new `tsconfig.base.json`
- Convert everything to ESM (except webpack.config.ts files in root which can't handle it)

Future work:
- Add missing `dependencies` and `devDepenedencies` and standardize `packages/*/package.json` files
- Add `clean`, `prepack`, and `test` scripts to `packages/*/package.json` files
- Point package.json `main` to compiled JS files and have webpack use typescript project references (if jest can support this)

### DIFF
--- a/app/tsconfig.json
+++ b/app/tsconfig.json
@@ -1,17 +1,26 @@
 {
-  "extends": "../tsconfig.json",
+  "extends": "../tsconfig.base.json",
+  "include": [
+    "./**/*",
+    "./**/*.json",
+    "./.storybook/**/*",
+    "../package.json",
+    "../typings/*.d.ts",
+    // Webpack config is imported by Storybook
+    "../webpack.*.config.ts",
+    "../WebpackArgv.ts"
+  ],
   "compilerOptions": {
-    "baseUrl": ".",
-    "module": "es2020",
+    "noEmit": true,
+    "rootDir": "../",
     "jsx": "react-jsx",
-    "lib": ["DOM", "DOM.Iterable", "es2020"],
+    "lib": ["dom", "dom.iterable", "es2020"],
     "paths": {
+      // Storybook breaks without this
       "micro-memoize": ["../typings/micro-memoize"],
-      "@foxglove-studio/app/*": ["*"]
-    }
-  },
-  "typeAcquisition": {
-    "exclude": ["node"]
-  },
-  "include": ["**/*", "../typings/*.d.ts", ".storybook/*"]
+      // This is only needed for vscode import suggestions
+      "@foxglove-studio/app/*": ["./*"]
+    },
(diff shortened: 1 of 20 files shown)


jtbandes:
> Removed unnecessary .eslintrc.yml files in packages where we are not overriding tsconfig compiler options

FWIW, I think it was necessary to have one in each project because otherwise eslint complained that the linted files didn't match the tsconfig `include`.

amacneil:
@jtbandes interesting. I think I avoided that by switching to the `tsconfig.base.json` which doesn't explicitly include or exclude any files. However, I'm seeing some different lint failures now which seem unrelated so will look into them.

desktop/tsconfig.json
```diff
@@ -1,4 +1,7 @@
 {
-  "extends": "../tsconfig.json",
-  "include": ["**/*", "../typings/*.d.ts"]
+  "extends": "../tsconfig.base.json",
+  "include": ["**/*", "../typings/*.d.ts"],
```
defunctzombie, line 3: Is this `**/*` correct? This includes every file regardless of extension. We should limit to ts files.

app/tsconfig.json
```diff
@@ -1,17 +1,25 @@
...
   },
   "typeAcquisition": {
     "exclude": ["node"]
   },
-  "include": ["**/*", "../typings/*.d.ts", ".storybook/*"]
+  "include": [
+    "./**/*",
+    "./**/*.json",
+    "./.storybook/**/*",
+    "../*.ts",
```
defunctzombie, line 21: Lets call out the webpack config files explicitly to be clear about why we reach into the parent directory.

packages/just-fetch/tsconfig.json
```diff
@@ -1,11 +1,9 @@
 {
-  "extends": "../../tsconfig.json",
-  "include": ["src/**/*.ts"],
+  "extends": "../../tsconfig.base.json",
+  "include": ["./src/**/*"],
```
defunctzombie, line 3: Do we not need `.ts` extensions?

desktop/tsconfig.json
amacneil: it includes only ts, tsx, js, jsx. we should have none of the latter. I can change if you want, but it seems less fragile to consistently include these file extensions

app/tsconfig.json
amacneil: done

packages/just-fetch/tsconfig.json
amacneil: no, see comment above. it only includes ts+js files by default, and there should be no js files under src

PR #619, opened 2021-04-21
Standardize packages folder structure

amacneil:
This is an even more minimal version of https://github.com/foxglove/studio/pull/618.

Basically, it just standardizes all packages to have a `src` directory (which is necessary so that we can run `tsc` to produce a nice tidy `dist` directory).

However, I left out all the tsconfig and other changes in this PR. Small PRs are easier to understand & review.


packages/velodyne-cloud/src/perf/transform.ts
```diff
@@ -0,0 +1,22 @@
...
+import { Model } from "../VelodyneTypes";
+import { HDL32E_PACKET1 } from "../fixtures/packets";
+
+const calibration = new Calibration(Model.HDL32E);
+const transform = new Transformer(calibration);
+
+for (let i = 0; i < 500; i++) {
+  const raw = new RawPacket(HDL32E_PACKET1);
+  const count = 0; // transform.validPoints(raw); // FIXME @jhurliman
+  const cloud = new PointCloud({ stamp: 0, maxPoints: count * 100 });
```
amacneil, line 18: @jhurliman two issues with this file I ran into:
- `transform.validPoints()` is not a function
- renamed `count` parameter to `maxPoints`

preload/index.ts
```diff
@@ -35,6 +35,7 @@ type IpcListener = (ev: unknown, ...args: unknown[]) => void;
 const menuClickListeners = new Map<string, IpcListener>();
 
 // Initialize the RPC channel for electron-socket
+// FIXME: @jhurliman this returns a promise, should we have await here?
 PreloaderSockets.Create();
```
amacneil, line 39: @jhurliman

packages/velodyne-cloud/src/perf/transform.ts
jhurliman: This should be:

```
  const maxPoints = RawPacket.MAX_POINTS_PER_PACKET * 100;
  const cloud = new PointCloud({ stamp: 0, maxPoints });
```

preload/index.ts
jhurliman: No, this is intentionally ran asynchronously.

jtbandes: There is a lint rule we can consider enabling for this but it's too noisy to turn on everywhere right now I think. Basically would require you to add a `.catch` or `void` or something.

PR #623, opened 2021-04-21
Load tsconfig.json files from root eslintrc

amacneil:
Remove unnecessary `parserOptions.project` in `.eslintrc.yaml` files. Instead, we specify all tsconfig files in the root `.eslintrc.yaml`.

I also deleted a bunch in https://github.com/foxglove/studio/pull/620, this is just remaining cleanup.

Fewer configuration files and overridden settings is easier to understand.

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -31,14 +31,9 @@ extends:
 
 parser: "@typescript-eslint/parser"
 parserOptions:
-  ecmaFeatures:
-    jsx: true
-
-  ecmaVersion: 12
+  ecmaVersion: 2020
   sourceType: module
-  project:
-    - "./tsconfig.json"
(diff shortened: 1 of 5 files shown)


.eslintrc.yaml
```diff
@@ -31,14 +31,9 @@ extends:
 
 parser: "@typescript-eslint/parser"
 parserOptions:
-  ecmaFeatures:
-    jsx: true
```
amacneil, line 35: According to https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/parser#parseroptionsecmafeaturesjsx this flag has no effect because (1) `*.tsx` files are always parsed with jsx support and (2) this flag is ignored if you set `project:` (which we do).

PR #627, opened 2021-04-22
Enable React.StrictMode for new panels

jtbandes:
Read more about strict mode at https://reactjs.org/docs/strict-mode.html

> StrictMode is a tool for highlighting potential problems in an application. …  It activates additional checks and warnings for its descendants.

> Strict mode checks are run in development mode only; they do not impact the production build.

The most notable behavior of strict mode is that it causes component render functions, useMemo callbacks, and other kinds of functions that should be stateless to be called **twice per render** in dev mode. This may cause problems in components that use these in side-effectful ways, so I didn't want to immediately enable it by default in all panels. I added a warning icon in the panel toolbar to prompt us to enable it in existing panels incrementally:

image

Also enable strict mode for URDFViewer, by removing usePreviousValue (and mark usePreviousValue as deprecated).

### DIFF
--- a/app/panels/URDFViewer/index.tsx
+++ b/app/panels/URDFViewer/index.tsx
@@ -11,7 +11,7 @@ import {
   Toggle,
   useTheme,
 } from "@fluentui/react";
-import { useCallback, useEffect, useLayoutEffect, useMemo, useState } from "react";
+import { useCallback, useEffect, useLayoutEffect, useMemo, useRef, useState } from "react";
 import { useResizeDetector } from "react-resize-detector";
 import { CameraStore, CameraListener, CameraState, DEFAULT_CAMERA_STATE } from "regl-worldview";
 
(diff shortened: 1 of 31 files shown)


app/panels/URDFViewer/index.tsx
```diff
@@ -59,16 +58,17 @@ function URDFViewer({ config, saveConfig }: Props) {
...
+    const prevAssetIds = new Set(prevAssets.current?.map(({ uuid }) => uuid));
+    prevAssets.current = assets;
     for (const asset of assets) {
       if (!prevAssetIds.has(asset.uuid) && asset.type === "urdf") {
         setSelectedAssetId(asset.uuid);
         return;
       }
     }
-  }, [assets, prevAssets, selectedAssetId]);
+  }, [assets]);
```
jtbandes, line 71: Example of how to fix a problem raised by strict mode. `usePreviousValue` would run twice in strict mode, so when the effect finally runs `prevAssets` would be the same as `assets`. Instead we move the tracking of the previous value to a place we control, and change it at the time we are using it (in an effect, which only runs once as the render is being committed).

PR #632, opened 2021-04-23
Fix TCP connections to mDNS hostnames

jhurliman:
This uses a JS-native mDNS implementation to work around inconsistent mDNS support at the OS level


defunctzombie:
```/// <reference types="node" />
/// <reference path="../../typings/dns-packet/index.d.ts" />
/// <reference path="../../typings/multicast-dns/index.d.ts" />```

Are these not picked up by typings? We should avoid this as it will pollute the entire namespace of the project I think.

File: packages/electron-socket/src/preloader/dns.ts [hunk 1]
```diff
@@ -0,0 +1,73 @@
...
+  const timer = setInterval(query, 1000);
+  query();
+
+  mdns.on("response", (res: packet.Packet) => {
+    if (res.answers == undefined) {
+      // Ignore this response, wait and see if another comes in
+      return;
+    }
+
+    for (const a of res.answers) {
```
jtbandes, line 50:
```suggestion
    for (const answer of res.answers) {
```

defunctzombie:
In your earlier testing it looked like `dns.lookup()` was able to lookup an mDNS value - are we not able to use that instead of these other js implementations?

jhurliman:
@defunctzombie it’s not consistent. On Linux for example, it depends on whether you have the avahi daemon installed and running.

defunctzombie:
> @defunctzombie it’s not consistent. On Linux for example, it depends on whether you have the avahi daemon installed and running.

😮 thats quite tragic

PR #640, opened 2021-04-24
Improve Jest performance

jtbandes:
On my M1 MacBook Air I can now run **all** the tests in about 16 seconds. Running a single small test file is down from 15sec to <2sec.

There were a few components to this:
- Jest would take several seconds to load `jest.config.ts` files (https://github.com/facebook/jest/issues/11337). Fixed by replacing them with `jest.config.json` files.
- Jest tries to use the native `find` if available over node fs, but it doesn't pass any exclude options to find, which means that it has to wait a few seconds for `find`, then filter a huge list of all .js files in app/node_modules. Upgrading jest lets us pass a config option to force it to use node fs instead (https://github.com/facebook/jest/pull/11264#issuecomment-825377579).
- `jsdom` is no longer the default environment in jest, and there is some overhead to jsdom. We are embracing this change by only enabling jsdom in certain test files that need it. Unfortunately Jest only recognizes the `/** @jest-environment jsdom */` comment if it is at the very beginning of the file, so I had to replace our eslint-plugin-header rule with a custom rule to enforce the presence of license headers.
- Converted the custom `?raw` transformer to .js instead of .ts, since we are not using ts-node anymore.
- Actually executing the test files with ts-jest was slow, because it has to run the TS compiler. Instead we can use babel-jest which doesn't do type checking. https://github.com/kulshekhar/ts-jest/issues/961#issuecomment-458294885 (This also requires enabling `isolatedModules` in tsconfig.) For better or worse, our regular webpack build already typechecks our test files.

Before:
image

After:
image


eslint-rules/foxglove-license-header.js
```diff
@@ -0,0 +1,37 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
defunctzombie, line 1: maybe throw this in packages?

packages/ros1/package.json
```diff
@@ -25,8 +25,8 @@
   },
   "devDependencies": {
     "@jest/types": "26.6.2",
-    "jest": "26.6.3",
-    "jest-circus": "26.6.3",
+    "jest": "27.0.0-next.8",
```
defunctzombie, line 28: drop this in favor of just having it listed in repo root?

jtbandes: @amacneil thoughts on this after your excursions in packaging setup?

amacneil: I haven't done my package.json cleanup yet, but jest and typescript should be listed in each package.json so that we can individually run `yarn test` (or `yarn jest`) from inside these packages.

amacneil:
Are files still being type checked in CI now?

jtbandes:
Some are, like the ones in `app`, but maybe not all. That is an important follow up item I can look at next week

PR #641, opened 2021-04-24
Use SentryWebpackPlugin to upload source maps for production builds

defunctzombie:
I've tested this change by uploading source maps for my prod build of a local `-dev` version.

### DIFF
--- a/webpack.renderer.config.ts
+++ b/webpack.renderer.config.ts
@@ -47,13 +47,14 @@ export function makeConfig(_: unknown, argv: WebpackArgv, options?: Options): Co
   if (
     !isDev &&
     process.env.SENTRY_AUTH_TOKEN != undefined &&
+    process.env.SENTRY_ORG != undefined &&
     process.env.SENTRY_PROJECT != undefined
   ) {
     plugins.push(
       new SentryWebpackPlugin({
         authToken: process.env.SENTRY_AUTH_TOKEN,
-        org: "foxglove",
-        project: "studio",
+        org: process.env.SENTRY_ORG,
+        project: process.env.SENTRY_PROJECT,
         release: `${process.env.SENTRY_PROJECT}@${packageJson.version}`,
 
         // Since the render config appears last in the list of webpack configs, we use it to upload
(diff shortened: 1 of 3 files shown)


webpack.renderer.config.ts
```diff
@@ -41,6 +43,28 @@ export function makeConfig(_: unknown, argv: WebpackArgv, options?: Options): Co
...
+    !isDev &&
+    process.env.SENTRY_AUTH_TOKEN != undefined &&
+    process.env.SENTRY_PROJECT != undefined
+  ) {
+    plugins.push(
+      new SentryWebpackPlugin({
+        authToken: process.env.SENTRY_AUTH_TOKEN,
+        org: "foxglove",
+        project: "studio",
+        release: `${process.env.SENTRY_PROJECT}@${packageJson.version}`,
```
defunctzombie, line 57: @amacneil This `release` field needs to match the `release` field in Sentry.init within our `app/index.tsx` and `preload/index.tsx` files. Importing `version.ts` which lives in `app` from the webpack config seemed _off_ so I imported package.json directly. Open to other thoughts.

amacneil: Yeah sounds fine. I’m not sure that there’s really much point to our version.ts file anyway. We should just import package.json directly everywhere. Typescript still enforces correctness for us.

webpack.renderer.config.ts
```diff
@@ -41,6 +43,28 @@ export function makeConfig(_: unknown, argv: WebpackArgv, options?: Options): Co
...
+  if (
+    !isDev &&
+    process.env.SENTRY_AUTH_TOKEN != undefined &&
+    process.env.SENTRY_PROJECT != undefined
+  ) {
+    plugins.push(
+      new SentryWebpackPlugin({
+        authToken: process.env.SENTRY_AUTH_TOKEN,
+        org: "foxglove",
+        project: "studio",
```
amacneil, line 56: You should set these to `env.SENTRY_ORG` and `env.SENTRY_PROJECT` (SENTRY_PROJECT already exists, we can add SENTRY_ORG).

Either that or we stop using env vars and just hard code them everywhere. But it seems sensible to keep it all in one place.

PR #642, opened 2021-04-25
Reintroduce storycap and reg-suit

amacneil:
With our current level of usage (\~370 stories, 4 committers) we are producing over 300k screenshots/month, which is unreasonably expensive on Chromatic (over $1500 in April). This would only grow as our team size and app complexity increases. Their "free for open source" plan would only save us $150/mo.

For now just adding storycap and reg-suit scripts back to the repo so we can run them locally enable them again in the future if needed. If we choose to enable and replace Chromatic I will do this in a separate PR (will require updating `chromatic` parameter everywhere).


defunctzombie:
You'll need to update all stories that use the `chromatic` parameter to use the `screenshot` parameter.

PR #645, opened 2021-04-26
Clean up packages package.json files

amacneil:
Mostly just consistency. At a minimum each package.json file has a `prepack` script to build. Packages with tests also have a `test` script.


amacneil:
Running `yarn test` inside the packages (for example `packages/wasm-bz2`) because it can't figure out how to parse typescript. I can't tell what I'm doing wrong.

@jtbandes any thougths?

PR #646, opened 2021-04-26
Fix unstable useMessagePipeline selectors

defunctzombie:
If the selector of useMessagePipeline returns a different object for
each invocation it will cause a state set and re-render. Which will
invoke the selector which would again return a different object and repeat
the cycle.

This change invokes the selector twice (in development) and compares the value.
If the value is different (reference equality fails), then a warning is shown
to fix this error during development.

### DIFF
--- a/app/PanelAPI/useDataSourceInfo.ts
+++ b/app/PanelAPI/useDataSourceInfo.ts
@@ -11,7 +11,7 @@
 //   found at http://www.apache.org/licenses/LICENSE-2.0
 //   You may not use this file except in compliance with the License.
 
-import { useCallback, useMemo } from "react";
+import { useMemo } from "react";
 import { Time } from "rosbag";
 
 import {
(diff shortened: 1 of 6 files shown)


app/PanelAPI/useMessageReducer.ts
```diff
@@ -126,6 +123,16 @@ type Params<T> = {
   preloadingFallback?: boolean;
 };
 
+const NO_MESSAGES = Object.freeze([]);
+function selectMessages(ctx: MessagePipelineContext) {
```
defunctzombie, line 127: This is broken - it causes state update on messages whenever any new message is received - not just when a message from our filter is received

app/PanelAPI/useMessageReducer.ts
```diff
@@ -189,31 +192,34 @@ export function useMessageReducer<T>(props: Params<T>): T {
   // Whenever `subscriptions` change, request a backfill, since we'd like to show fresh data.
   useEffect(() => requestBackfill(), [requestBackfill, subscriptions]);
 
-  // Keep a reference to the last messages we processed to ensure we never process them more than once.
-  // If the topics we care about change, the player should send us new messages soon anyway (via backfill if paused).
-  const lastProcessedMessagesRef = useRef<readonly Message[] | undefined>();
-  // Keep a ref to the latest requested topics we were rendered with, because the useMessagePipeline
-  // selector's dependencies aren't allowed to change.
-  const latestRequestedTopicsRef = useRef(requestedTopicsSet);
-  latestRequestedTopicsRef.current = requestedTopicsSet;
```
jtbandes, line 198: Is the theory that this was just suppressing the warning from useMessagePipeline previously, so it's safe to delete?

app/PanelAPI/useMessageReducer.ts
```diff
@@ -189,31 +192,34 @@ export function useMessageReducer<T>(props: Params<T>): T {
...
+          return prevFiltered.current;
+        }
+        prevMessageEventsRef.current = messageEvents;
+
+        const filtered = messageEvents.filter(({ topic }) => requestedTopicsSet.has(topic));
+        if (filtered.length === 0) {
+          return NO_MESSAGES;
+        }
+
+        prevFiltered.current = filtered;
```
jtbandes, line 218: This actually seems wrong since when you return NO_MESSAGES you have updated `prevMessageEventsRef`, but not `prevFiltered`. The first time it will return NO_MESSAGES and the second time (if messages have not changed) it will return the **old** value of prevFiltered.current rather than NO_MESSAGES again.

app/PanelAPI/useMessageReducer.ts
defunctzombie: Yep - there is a new warning within useContextSelector that detects unstable identities.

app/PanelAPI/useMessageReducer.ts
defunctzombie: yep - need to update this to set `prevFiltered.current = NO_MESSAGES`;

PR #666, opened 2021-04-26
Remove dead code for layout and panel actions

defunctzombie:
Found these while going through sendNotification calls.


app/components/ShareJsonModal.tsx
```diff
@@ -42,6 +42,7 @@ function encode(value: any): string {
   try {
     return JSON.stringify(value, undefined, 2);
   } catch (e) {
+    // fixme - this should not throw a notification app level, the component should handle it internally and display an error on the modal
     sendNotification("Error encoding value", e, "app", "error");
```
jtbandes, line 46: This one is weird, I don't really understand why it is here since the panel does already handle & display some other errors. Anyway, the whole component should probably be rewritten with hooks at some point.

jtbandes (approved):

app/components/ShareJsonModal.tsx
defunctzombie: I tried that once - maybe not that it is used in only one place I can have success again.

PR #669, opened 2021-04-26
Enable noUnusedParameters in app

amacneil:
This is the only remaining typescript syntax override in our repo. Outside of `app/` code was already enforcing `noUnusedParameters`.

I did not attempt to investigate whether any of these might be hiding real bugs, I only added underscore prefixes where necessary to make the compiler happy.

I also changed `any` and `never` to `unknown` in a few places where it made sense.

### DIFF
--- a/app/panels/ThreeDimensionalViz/commands/LaserScans.tsx
+++ b/app/panels/ThreeDimensionalViz/commands/LaserScans.tsx
@@ -79,30 +79,30 @@ const laserScan = (regl: Regl) =>
   `,
 
     uniforms: {
-      pointSize: (context: never, props: any) => props.settings?.pointSize || 4,
-      isCircle: (context: never, props: any) =>
+      pointSize: (_context: unknown, props: any) => props.settings?.pointSize || 4,
+      isCircle: (_context: unknown, props: any) =>
         (props.settings && props.settings.pointShape === "circle") || false,
       // Color is not included in the LaserScan message - it's only included if the color is added by
       // getChildrenForHitmap.
-      isHitmap: (context: never, props: any) => !!props.color,
+      isHitmap: (_context: unknown, props: any) => !!props.color,
 
       angle_min: regl.prop("angle_min"),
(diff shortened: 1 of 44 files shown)


app/panels/ThreeDimensionalViz/commands/LaserScans.tsx
```diff
@@ -79,30 +79,30 @@ const laserScan = (regl: Regl) =>
   `,
 
     uniforms: {
-      pointSize: (context: never, props: any) => props.settings?.pointSize || 4,
-      isCircle: (context: never, props: any) =>
+      pointSize: (_context: unknown, props: any) => props.settings?.pointSize || 4,
+      isCircle: (_context: unknown, props: any) =>
```
amacneil, line 83: Can someone confirm that changing `never` to `unknown` here is correct? My understanding is that `never` means "this function should never be called with this positional argument", which seems incorrect if we are simply ignoring a parameter.

@defunctzombie?

app/tsconfig.json
```diff
@@ -20,7 +20,6 @@
       "micro-memoize": ["../typings/micro-memoize"],
       // This is only needed for vscode import suggestions
       "@foxglove-studio/app/*": ["./*"]
-    },
-    "noUnusedParameters": false
```
amacneil, line 24: this is the actual config change

app/panels/ThreeDimensionalViz/commands/LaserScans.tsx
jtbandes: This is fine. Never would have been incorrect since this arg is passed in by regl. If we're not using it then unknown is correct. It looks like regl ships type definitions but we don't really use them anywhere that I'm aware of.

amacneil: I see. So it didn't cause any issues because this is only called by 3rd party code, but I assume if we tried to invoke this from our own codebase typescript would have been unhappy.

PR #670, opened 2021-04-26
Rework how players produce warnings and errors.

defunctzombie:
Previously, players used sendNotification whenever they encountered a
warning or an error. sendNotification is a one-way out-of-band reporting
mechanism. Once the player produced the error it would persist in the app
(and in the user's app bar) for the remainder of the app lifetime. This
led to a confusing user experience when a player had a transient error
that was resolved (i.e. reconnection after failed connection).

This change removes sendNotification from all players and replaces it
with a _problems_ entry in the player state. The player state is how
players communicate their updates to the outside world. Problems are another
form of player updates. By sending problems via the player state, a
player can add/remove/clear problems as necessary.

This change updates the player presence indicator ui to indicate the presence
of problems in addition to the player presence. The user can see the list
of problems by clicking the presence indicator and and further seeing any
details a problem may expose.

Here is an example of a player loading and indicating errors:
Screen Shot 2021-04-30 at 6 31 52 PM


app/components/NotificationDisplay.tsx
```diff
@@ -210,12 +214,60 @@ export function NotificationModal({
...
+  return ctx.playerState.error;
+}
+
 export default function NotificationDisplay(): React.ReactElement {
   const [notifications, setNotifications] = useState<NotificationMessage[]>([]);
   const [showMostRecent, setShowMostRecent] = useState(false);
   const [clickedNotification, setClickedNotification] = useState<NotificationMessage>();
   const hideTimeout = useRef<ReturnType<typeof setTimeout>>();
 
+  const playerNotification = useMessagePipeline(selectPlayerNotification);
```
jtbandes, line 227: I don't think this belongs in NotificationDisplay, which is otherwise just a generic component that displays notifications sent elsewhere... MessagePipeline itself might be a good place to do it

app/components/index.ts
```diff
@@ -0,0 +1,5 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+export * from "./PlayerStatusIndicator";
```
defunctzombie, line 5: I had the idea to import components from `@foxglove-studio/app/components` using `{}` brace syntax - but that may only be useful in limited places since other components in the components folder would not be able to do this as they would trigger circular dependencies.

app/panels/ThreeDimensionalViz/index.tsx
```diff
@@ -61,12 +69,7 @@ const BaseRenderer = (props: Props, ref: any) => {
   } = props;
   const { updatePanelConfig } = React.useContext(PanelContext) || {};
 
-  const currentTime = useMessagePipeline(
-    useCallback(
-      ({ playerState: { activeData } }) => activeData?.currentTime ?? { sec: 0, nsec: 0 },
```
defunctzombie, line 66: fixed up this selector return value being unstable when there is no currentTime - separate PR

app/util/RpcMainThreadUtils.ts
```diff
@@ -16,6 +16,9 @@ import sendNotification from "@foxglove-studio/app/util/sendNotification";
 
 import Rpc from "./Rpc";
 
+// fixme - this is confusing indirection - now we can't see easily what our
```
defunctzombie, line 19: These handlers should be in their individual files - separate PR tho

app/players/UserNodePlayer/index.ts
```diff
@@ -66,62 +68,43 @@ type UserNodeActions = {
   setUserNodeRosLib: SetUserNodeRosLib;
 };
 
-const rpcFromNewSharedWorker = (worker: SharedWorker, name: string) => {
```
defunctzombie, line 69: I dislike the disconnect from call site and error hookup - regardless of that stylistic preference - the error handling needs access to member field now so I found it easier to put inline since some of the workers are userspace node specific

app/players/UserNodePlayer/index.ts
```diff
@@ -66,62 +68,43 @@ type UserNodeActions = {
...
+  private _rosLibDatatypes?: RosDatatypes; // the datatypes we last used to generate rosLib -- regenerate if they change
+  private _globalVariables: GlobalVariables = {};
+  private _pendingResetWorkers?: Promise<void>;
+
+  // Player state changes when the child player invokes our player state listener
+  // we may also emit state changes on internal errors
+  private _playerState?: PlayerState;
+
+  // any errors to report into state we emit
+  //private _problems: Problem[] = [];
```
defunctzombie, line 103: delete

app/players/UserNodePlayer/index.ts
```diff
@@ -255,97 +239,197 @@ export default class UserNodePlayer implements Player {
...
-      nodeId,
-      nodeData,
-      inputs: inputTopics,
-      output: { name: outputTopic, datatype: outputDatatype },
-      processMessage: async (message: Message) => {
-        // We allow _resetWorkers to "cancel" the processing by creating a new signal every time we process a message
-        terminateSignal = signal<void>();
 
-        // Register the node within a web worker to be executed.
+    const problemKey = `node-id-${nodeId}`;
```
defunctzombie, line 243: add comment on what the problem key is

```problemKey is a unique identifier for the userspace node to surface a problem to the player alongside other userspace nodes or player problems. The userspace node clears its problem key when there is no longer a problem. This supports surfacing problems from multiple userspace nodes```

File: app/players/UserNodePlayer/index.ts [hunk 8]
```diff
@@ -255,97 +239,197 @@ export default class UserNodePlayer implements Player {
...
+
+        const { error, userNodeDiagnostics, userNodeLogs } = await rpc.send<RegistrationOutput>(
+          "registerNode",
+          {
+            projectCode,
+            nodeCode: transpiledCode,
+          },
+        );
+        if (error != undefined) {
+          this._setUserNodeDiagnostics(nodeId, [
```
defunctzombie, line 303: Do these diagnostics need to be surfaced as a problem? maybe the error needs to be surfaced as a problem and `setUserNodeDiagnostics` could be deleted

app/players/UserNodePlayer/index.ts
```diff
@@ -255,97 +239,197 @@ export default class UserNodePlayer implements Player {
...
+      }
 
-        const result = (await Promise.race([
-          rpc.send("processMessage", { message, globalVariables: this._globalVariables }),
-          terminateSignal,
-        ])) as any;
-        if (!result) {
-          return;
-        }
+      this._problemStore.delete(problemKey);
```
defunctzombie, line 354: add comment indicating that we received a message successfully and can clear any previous problem for this userspace node

app/players/UserNodePlayer/index.ts
```diff
@@ -490,9 +579,52 @@ export default class UserNodePlayer implements Player {
...
 
+      this._playerState = newPlayerState;
       this._lastPlayerStateActiveData = playerState.activeData;
-      return listener(newPlayerState);
-    });
+    } catch (err) {
+      log.error(err);
+
+      // if there is an error, add it to the player state
+      const newPlayerState: PlayerState = {
```
defunctzombie, line 588: change this to add the error to the problemStore? when would this error be cleared? right now this overwrites any problems from the child player which is wrong. emitState already handled augmenting child problems so we should avoid doing that in two places?

amacneil:
I am getting this (related?) error on current main branch. To produce, simply run `yarn clean && yarn build:prod && yarn start`.

image (3)

defunctzombie:
@amacneil I am not able to reproduce with your instructions on this branch.

app/components/PlayerStatusIndicator/PlayerPresenceIcon.tsx
```diff
@@ -0,0 +1,49 @@
...
+import AlertIcon from "@mdi/svg/svg/alert.svg";
+
+import Icon from "@foxglove-studio/app/components/Icon";
+import SpinningLoadingIcon from "@foxglove-studio/app/components/SpinningLoadingIcon";
+import { PlayerPresence, PlayerState } from "@foxglove-studio/app/players/types";
+import { colors } from "@foxglove-studio/app/util/sharedStyleConstants";
+
+type Props = {
+  presence: PlayerState["presence"];
+  problems?: boolean;
```
jtbandes, line 14:
```suggestion
  hasProblems?: boolean;
```

app/components/PlayerStatusIndicator/PlayerProblemModal.tsx
```diff
@@ -0,0 +1,56 @@
...
+  problem: Problem;
+  onRequestClose: ModalProps["onRequestClose"];
+};
+
+const useStyles = makeStyles({
+  modalBody: {
+    padding: "28px",
+    maxWidth: "600px",
+    minWidth: "300px",
+    maxHeight: "80hv",
```
jtbandes, line 23: vh?

app/components/PlayerStatusIndicator/PlayerProblemModal.tsx
```diff
@@ -0,0 +1,56 @@
...
+    overflow: "auto",
+  },
+  tip: {
+    paddingBottom: "8px",
+  },
+  error: {
+    fontFamily: "monospace",
```
jtbandes, line 30: Please use our normal monospace family, see `mixins.monospaceFont` elsewhere

app/components/PlayerStatusIndicator/PlayerProblemModal.tsx
```diff
@@ -0,0 +1,56 @@
...
+      <Title>{problem.message}</Title>
+      <hr />
+      <div className={classes.modalBody}>
+        {problem.tip != undefined && (
+          <div className={classes.tip}>
+            <span>{problem.tip}</span>
+          </div>
+        )}
+        {problem.error?.stack != undefined && (
+          <pre className={classes.error}>{problem.error?.stack}</pre>
```
jtbandes, line 51:
```suggestion
          <pre className={classes.error}>{problem.error.stack}</pre>
```

app/components/PlayerStatusIndicator/PlayerProblemModal.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { Problem } from "@foxglove-studio/app/players/types";
+
+type ModalProps = ComponentProps<typeof Modal>;
+
+type Props = {
+  problem: Problem;
+  onRequestClose: ModalProps["onRequestClose"];
+};
+
+const useStyles = makeStyles({
```
jtbandes, line 18: What's the difference between makeStyles/mergeStyles? Do you normally recommend using makeStyles?

app/components/PlayerStatusIndicator/PlayerStatusIndicator.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+function selectPlayerProblems(ctx: MessagePipelineContext) {
+  return ctx.playerState.problems;
+}
+
+/**
+ * PlayerStatusIndicator shows the player presence via an icon button. If there are player problems
+ * a badge or icon will appear indicating there are additional items for review.
+ *
+ * When a user clicks the button, a callbout with all player problems appears and the user can
```
jtbandes, line 32:
```suggestion
 * When a user clicks the button, a callout with all player problems appears and the user can
```

app/components/PlayerStatusIndicator/PlayerStatusIndicator.tsx
```diff
@@ -0,0 +1,111 @@
...
+ * review the details of the problems.
+ */
+export function PlayerStatusIndicator(): JSX.Element {
+  const playerPresence = useMessagePipeline(selectPlayerPresence);
+  const presenceTitle = PresenceToString(playerPresence);
+  const originalPlayerProblems = useMessagePipeline(selectPlayerProblems) ?? [];
+
+  // player problems reference might be changing when the actual problems are not
+  // or the player problems reference might be stable but new problems are added to the array
+  // the player api does not make guarantees about the _problems_ array identity when emitting state
```
jtbandes, line 42: please use some punctuation for long comments...if this gets wrapped differently it becomes hard to read (already hard to read at current length)

app/components/index.ts
jtbandes: maybe avoid this new pattern if there is only one thing using it? It could be nice but I agree we would have lots of circular dependency issues this way

app/players/RandomAccessPlayer.ts
```diff
@@ -130,6 +128,11 @@ export default class RandomAccessPlayer implements Player {
   _lastRangeMillis?: number;
   _parsedMessageDefinitionsByTopic: ParsedMessageDefinitionsByTopic = {};
 
+  // The problem store holds problems based on keys
+  // The overall player may be healthy, but individual topics may have warnings or errors.
+  // These are set/cleared in the store to track the current set of problems
+  _problemStore = new Map<string, Problem>();
```
jtbandes, line 134: Use an enum type for keys?

Any reason not to just call this `_problems`?

app/players/Ros1Player.ts
```diff
@@ -120,10 +122,14 @@ export default class Ros1Player implements Player {
...
 
     await this._rosNode.start();
-    this._requestTopics();
+    await this._requestTopics();
+    this._presence = PlayerPresence.PRESENT;
   };
 
   private _requestTopics = async (): Promise<void> => {
+    // clear problems before requesting topics again
+    this._problems = [];
```
jtbandes, line 131: comment doesn't really explain why?

app/players/UserNodePlayer/index.ts
jtbandes: ^

app/players/UserNodePlayer/index.ts
jtbandes: It seems like more error handling code is duplicated now; why did this need to be removed?

app/players/UserNodePlayer/nodeTransformerWorker/transformer.ts
```diff
@@ -385,26 +384,7 @@ export const extractDatatypes = (nodeData: NodeData): NodeData => {
...
-        {
-          severity: DiagnosticSeverity.Error,
-          message:
-            "Unknown error encountered. This is likely a problem with Node Playground itself.",
-          source: Sources.DatatypeExtraction,
-          code: ErrorCodes.DatatypeExtraction.UNKNOWN_ERROR,
-        },
-      ],
-    };
+    throw new Error(err);
```
jtbandes, line 387: What's `err`? Looks like there is `error`, which is already an Error

app/players/types.ts
```diff
@@ -96,6 +96,13 @@ export enum PlayerPresence {
   ERROR = "ERROR",
 }
 
+export type Problem = {
```
jtbandes, line 99: Problem is such a generic word, I think calling it at least PlayerProblem might be nice

app/players/UserNodePlayer/index.ts
defunctzombie: Error handling is different for the transform worker and the runtime worker. The runtime worker errors bucket by the nodeId so that we can surface errors from multiple userspace nodes and clear errors from individual userspace nodes. These errors happen when the worker itself may be running and we clear them if we get a successful message from the node. The error handling for the transform worker errors are not cleared and not tied to a specific node.

app/players/Ros1Player.ts
defunctzombie: Updated comment.

app/players/RandomAccessPlayer.ts
defunctzombie: The key may be a topic if the error pertains to a specific topic.

app/components/PlayerStatusIndicator/PlayerProblemModal.tsx
defunctzombie: Updated to use the mixin here. Quite tragic that I cannot provide fonts in the theme... typography is limited with fluent themes.

app/components/PlayerStatusIndicator/PlayerProblemModal.tsx
defunctzombie: makeStyles allows you to use a theme - and produces a _hook_ that you call in your component so access the classes (since they might change based on the theme). Since I don't have a theme here I can use mergeStyles which makes the classes once and doesn't need a hook.

PR #673, opened 2021-04-26
Set ROS_HOSTNAME in app preferences

jhurliman:
This allows Foxglove Studio to specify which hostname to explicitly announce itself as when connecting to a ROS1 native source. See http://wiki.ros.org/ROS/EnvironmentVariables#ROS_IP.2FROS_HOSTNAME for more information. `ROS_IP` could be added as well if there is a request for it.


app/players/Ros1Player.ts
```diff
@@ -46,6 +46,7 @@ const NO_WARNINGS = Object.freeze({});
 // showing simulated time, so current time from Date.now() is always used instead.
 export default class Ros1Player implements Player {
   private _url: string; // rosmaster URL.
+  private _hostname?: string; // ROS_HOSTNAME
```
defunctzombie, line 49: How is hostname different from the url?

app/components/PlayerManager.tsx
```diff
@@ -281,8 +282,11 @@ async function roscoreSource(options: FactoryOptions) {
 
   const url = maybeUrl;
   options.storage.setItem(storageCacheKey, url);
+
+  const hostname = useRosHostname();
+
   return async (playerOptions: BuildPlayerOptions) => ({
-    player: new Ros1Player(url, playerOptions.metricsCollector),
+    player: new Ros1Player(url, hostname, playerOptions.metricsCollector),
```
defunctzombie, line 289: Probably now is the time to make the constructor options an object rather than adding more fields?

app/components/Preferences.tsx
```diff
@@ -122,6 +132,30 @@ function TimezoneSettings(): React.ReactElement {
...
+        : "localhost",
+    [os],
+  );
+
+  return (
+    <TextField
+      label="ROS_HOSTNAME"
+      placeholder={rosHostnamePlaceholder}
+      value={rosHostname.value}
+      onChange={(_event, newValue) => setRosHostname(nonEmpty(newValue))}
```
defunctzombie, line 154: emptyToUndefined? I feel like there must be some conventional name for this? Maybe not cause :javascript:

PR #700, opened 2021-04-27
Improve useContextSelector efficiency, remove bailout feature

jtbandes:
Split from #646.

Update `createSelectableContext`/`useContextSelector` so it can run the selector (if it has changed) on each render, rather than needing to schedule a separate setState (the setState is still necessary when the subscriber is far away in a memoized part of the subtree).

Update `useMessageReducer` to do all the reducing inside its context selector, so we can remove the `BAILOUT` feature.

Remove some extra renders expected in tests, mostly added around the time of #363.

### DIFF
--- a/app/hooks/useContextSelector.test.tsx
+++ b/app/hooks/useContextSelector.test.tsx
@@ -132,17 +132,21 @@ describe("createSelectableContext/useContextSelector", () => {
     root.unmount();
   });
 
-  it("propagates value to multiple consumers", () => {
+  it("propagates value to multiple consumers, including memoized subtrees", () => {
     const C = createSelectableContext<{ one: number; two: number }>();
     const Consumer1 = createTestConsumer(C, ({ one }) => one);
     const Consumer2 = createTestConsumer(C, ({ two }) => two);
 
+    const Memoized = React.memo(function Memoized({ children }: PropsWithChildren<unknown>) {
+      return <div>{children}</div>;
+    });
+
     const root = mount(
       <C.Provider value={{ one: 1, two: 2 }}>
         <Consumer1 />
-        <div>
+        <Memoized>
           <Consumer2 />
-        </div>
+        </Memoized>
       </C.Provider>,
     );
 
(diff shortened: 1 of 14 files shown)


app/panels/ThreeDimensionalViz/FrameCompatibility.tsx
```diff
@@ -28,21 +28,19 @@ const useFrame = (
...
     restore: React.useCallback(() => {
       frame.current = {};
-      return Date.now();
-    }, [frame]),
-    addMessages: React.useCallback(
-      (time, messages: readonly Message[]) => {
-        for (const message of messages) {
-          (frame.current[message.topic] = frame.current[message.topic] ?? []).push(message);
-        }
-        return time;
```
jtbandes, line 42: I don't really understand how this was working before — the return value `time` would always be the same for each addMessages call, so how did the component actually end up updating at the right time?

app/hooks/useContextSelector.test.tsx
```diff
@@ -101,7 +109,7 @@ describe("createSelectableContext/useContextSelector", () => {
     expect(Consumer.selectorFn.mock.calls).toEqual([[{ num: 1 }], [{ num: 1 }], [{ num: 2 }]]);
     expect(Consumer.renderFn.mock.calls).toEqual([[1], [2]]);
 
-    // Selector returns BAILOUT, so no update should occur
+    // Selector returns the same value, so no update should occur
```
jtbandes, line 112: @defunctzombie Another thing I don't understand below this - the Consumer component is not memoized, so why doesn't it get rendered again during `root.setProps`?

jtbandes: I think there may be an implicit memoization happening somewhere in either React or enzyme.

jtbandes: Determined that this is a built-in magical behavior of Context.Providers.

app/panels/ThreeDimensionalViz/FrameCompatibility.tsx
jtbandes: Answer: it didn't, we just got lucky. Fixed and added a test.

PR #746, opened 2021-04-28
Remove explicit puppeteer-core dependency

defunctzombie:


defunctzombie:
@jtbandes the resolution was originally added in https://github.com/foxglove/studio/pull/746 - do you remember what the context for that was?

jtbandes:
I think it was needed in order to make it possible to run storycap/puppeteer locally on M1 with Chrome.app. Maybe this workflow is more usable with latest package versions now that things have had more time to bake?

PR #792, opened 2021-04-28
Fix resize detector in PlaybackBarHoverTicks

jtbandes:
`react-resize-detector` depends on the `ref`'d component being immediately available after the first render. In this case the resize detector was not properly attached because the initial render returned null. Instead, we can unconditionally render the measured element, and just not render the HoverBar if the bounds are unknown.

cc @defunctzombie


defunctzombie:
👍 thank you - my bad

Issue #343, opened 2021-03-24
Crash on launch when running studio inside docker container


goupil35000:
Hello,
I installed foxglove v0.2.0 in a docker container.
After installation of foxglove, I'm unable to use roslaunch or rosrun on every package.
Verified with a new container. Deinstallation of foxglove-studio doesn't resolve the problem.
Foxglove-studio seems to work correctly on the same computer directly outside the container.

$ rosrun turtlesim turtlesim_node
QStandardPaths: XDG_RUNTIME_DIR not set, defaulting to '/tmp/runtime-student'
[ INFO] [1616618619.246782155]: Starting turtlesim with node name /turtlesim
[ INFO] [1616618619.251715865]: Spawning turtle [turtle1] at x=[5.544445], y=[5.544445], theta=[0.000000]
dbus[6192]: The last reference on a connection was dropped without closing the connection. This is a bug in an application. See dbus_connection_unref() documentation for details.
Most likely, the application was supposed to call dbus_connection_close(), since this is a private connection.
  D-Bus not built with -rdynamic so unable to print a backtrace
Aborted (core dumped)

amacneil:
Hi @goupil35000 

Can you share your Dockerfile?

Also, I'm interested to hear why you want to run Foxglove inside a container. It's designed to run directly on your computer (although you can still connect to ROS running inside a container as long as you use `--network=host` mode to expose all the ports).

goupil35000:
Hi,
Thanks for your reply. 

Why using a container for us ? Because we all use the same container  and for security reason we are not allowed to install software outside our container. And in case of problem (for example this is the case after installing foxglove-studio), we launch a new container.

Our container is based at the beginning on osrf/ros:noetic-desktop-full, but many things were added to this without a Dockerfile. And the size of our container is now 25GB so impossible to send it.

So I retested it directly with osrf/ros:noetic-desktop-full. I updated softwares in container with apt and then add the foxglove deb file to the system. The problem is now different. I can launch turtlesim in this container. But foxglove is not working in the container.
$ foxglove-studio
Trace/breakpoint trap

amacneil:
I can reproduce this locally. We haven't tested running studio inside docker yet, so this isn't a fully supported workflow and several things could be at fault here.

```
ubuntu@d9938818b032:/$ foxglove-studio
Trace/breakpoint trap (core dumped)
```

[...]

PR #797, opened 2021-04-29
Add Dockerfile

amacneil:
Allows docker-based development per https://github.com/foxglove/studio/issues/343#issuecomment-828649467.

Currently this successfully runs `yarn serve` inside the container. I haven't tested running `yarn start` (electron) inside the container yet. I highly doubt it will work on macOS, but maybe it's not too hard to get it working on linux.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 *.log
 *.tsbuildinfo
+.docker/
 .DS_Store
 .webpack/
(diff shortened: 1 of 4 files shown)


defunctzombie:
I'm not sure this is a useful thing to maintain. This is a GUI app and GUI docker containers are not the most approachable things.

PR #799, opened 2021-04-29
Remove Message type in favor of explicit TypedMessage<unknown>

defunctzombie:


defunctzombie:
@jtbandes @jhurliman after this lands my plan is to rename `TypedMessage` to `MessageEvent` and attempt to update the variable names where a TypedMessage is called `message` to a `messageEvent`

PR #806, opened 2021-04-29
Rename TypedMessage to MessageEvent to convey intent

defunctzombie:


amacneil:
Follow up PR from https://github.com/foxglove/studio/pull/799

PR #807, opened 2021-04-29
Add mDNS caching and jitter to retries, broadcast mDNS requests on all network interfaces

### LINKED ISSUE #805: mDNS caching
There is no caching on mDNS responses right now, which means we can flood a server with mDNS requests when connecting to many nodes at once (such as subscribing to `/rosout`). This results in the occasional timeout which is at least caught and displayed to the user in 0.6.1-dev, but we should be able to prevent it from happening entirely by obeying the TTL field of the mDNS response. The retry loop would also benefit from the exponential truncated backoff with jitter algorithm we use elsewhere in the codebase (ros1 library) rather than a fixed one second loop which can cause a thundering herd problem.

jhurliman:
Fixes #805. With these changes, I can now consistently connect to ROS nodes running on Duckiebot from a Windows machine that has multiple network interfaces.


jtbandes:
Can we get some unit tests for this code?

Issue #802, opened 2021-04-29
DevTools error: Content Security Policy of your site blocks the use of 'eval' in JavaScript


jhurliman:
DevTools shows this warning in 0.6.0:

```
The Content Security Policy (CSP) prevents the evaluation of arbitrary strings as JavaScript to make it more difficult for an attacker to inject unathorized code on your site.

Source Location | Directive | Status
-- | -- | --
file:412466 | script-src | blocked
file:422950 | script-src | blocked
```

This happens in 0.6.0 production release every time the app starts. It points to *something* being broken, but it requires more investigation to find out what.

amacneil:
Happening for me on macOS too:

![image](https://user-images.githubusercontent.com/637671/116577563-6fa46300-a8c5-11eb-9db6-d40c1abc4249.png)

The strange thing is that we are explicitly allowing `unsafe-eval` in our content security policy:

https://github.com/foxglove/studio/blob/v0.6.0/desktop/index.ts#L173-L181

amacneil:
@jhurliman are you seeing any actual problems as a result of this or just the warning?

jhurliman:
@amacneil I haven't found the block(s) of code that are being disabled, it's possible they are unused and we can just delete some code to fix this. But I know we use `new Function()` for bayer decoding in the image panel and I haven't tested that yet.

amacneil:
We are definitely not blocking `eval` everywhere today. If you remove that from our CSP and run a production build there are a lot of devtools errors and the app refuses to load.

So, I think this error is coming from somewhere else.

amacneil:
I did some more testing of this and it feels like probably an electron bug (or undocumented behavior).

- Removing `unsafe-inline` from CSP causes our app to definitely not work.
- Opening up our CSP to be as [flexible as possible](https://stackoverflow.com/a/51067035/462702) still causes this error in devtools, although no noticeable problems in the app besides the error showing up.
- Moving our CSP to a `<meta http-equiv="">` tag still shows the error.
- Removing our CSP from every request except `renderer/index.html` still shows the error.
- Removing our CSP entirely fixes the error

Basically, I can't find any combination of CSP that avoids this error, other than no CSP at all. However, the error doesn't seem to point to any specific file that failed to load, and I can't see any negative side effects in the app.

defunctzombie:
I've removed the bug label. We've not identified anything that is broken as a result of this. Seems like an erroneous warning within electron.

defunctzombie:
[...]

PR #808, opened 2021-04-29
Remove CSP from data: requests

amacneil:
While investigating https://github.com/foxglove/studio/issues/802, I noticed that we unnecessarily set CSP on `data:` "requests". Typical browsers would not be loading `data:` URLs from a web server so they would not have CSP headers set on them.

Also moved CSP const back down closer to where we use it, and fixed an unrelated comment typo.

### DIFF
--- a/desktop/index.ts
+++ b/desktop/index.ts
@@ -237,18 +237,6 @@ app.on("ready", async () => {
     }
   }
 
-  // Content Security Policy
-  // See: https://www.electronjs.org/docs/tutorial/security
-  const contentSecurityPolicy: Record<string, string> = {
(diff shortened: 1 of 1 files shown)


jtbandes:
Does this fix any bugs or issues? Are data requests normally supposed to “inherit” the CSP from whatever loaded them?

amacneil:
No, it doesn't fix anything that I can tell, other than removing what I consider to be weird and unexpected behavior.

CSP still applies to decide where we are allowed to load `data:` URIs inside a larger page. Since CSP is an HTTP header I don't think it was ever designed it to apply to non-HTTP requests.

PR #809, opened 2021-04-29
Add .webpack to vscode search.exclude

defunctzombie:
.webpack is a build artifacts folder and should be excluded from vscode search by default

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -13,6 +13,7 @@
     "**/*.code-search": true,
     "**/bower_components": true,
     "**/node_modules": true,
+    ".webpack": true,
     ".yarn/**": true,
     "yarn.lock": true
   },


.vscode/settings.json
```diff
@@ -13,6 +13,7 @@
     "**/*.code-search": true,
     "**/bower_components": true,
     "**/node_modules": true,
+    ".webpack": true,
```
amacneil, line 16: does it need `/**` or what's up with the yarn line below this?

defunctzombie: I don't think it does - its unclear - but most node_modules examples I found didn't have `/**` at the end.

jtbandes:
This has never been a problem for me. The files are ignored by default.

Do you have this button enabled in your find panel?

image

defunctzombie:
@jtbandes how would they be ignored by default if nothing says to ignore them? Are dotfiles ignored by default?

defunctzombie:
Don't know what was going on with my vscode but I think its fine now without this.

PR #816, opened 2021-04-30
Sidebar

jtbandes:
Adds a new Sidebar to the app, and moves the Add Panel flow, Variables menu, and Settings into the sidebar.

The sidebar uses another react-mosaic-component to allow resizing and separate it from the rest of the app content, but it is not part of the actual saved layout.

Future work will be to move individual panel settings into the sidebar (initially part of this PR but temporarily reverted in 9ab66b9bf79046c0a85cf1ef347cb795d7b8b7ff) as well as the layout selector, and finally clean up the window title bar height and appearance.

image
image


app/Sidebar.tsx
```diff
@@ -0,0 +1,286 @@
...
+    flexGrow: 1,
+    "> div, > div > div": {
+      display: "flex",
+      flexDirection: "column",
+      flexGrow: 1,
+    },
+  },
+});
+
+function SidebarButton({
```
defunctzombie, line 60: separate file plz

PR #817, opened 2021-04-30
Fix restoring saved sources in Player Manager

### LINKED ISSUE #810: ROS TCP connection dialog pops up on page refresh
I'm already connected to a ROS TCP data source. When I refresh the page, the TCP connection dialog pops up again. The app **also** automatically reconnects to the source, so there's no reason popping up the dialog should be necessary.

Sometimes I get **two** copies of the prompt on top of each other.

image

defunctzombie:
useLocalStorage behavior changed in a recent release of react-use
(https://github.com/streamich/react-use/issues/1970) which led to
the saved source appearing to change during initial mount. This
triggered the source change logic multiple times and broke the fragile
first mount restore behavior.

This change re-works this behavior to restore the saved source once
on mount without any fragile refs traking whether to restore by
making an explicit selectSource function that accepts a restore argument.

Manually selecting a source via the PlayerSelection context avoids
restoring and shows prompts and dialogs.

Closes #810

### DIFF
--- a/app/components/PlayerManager.tsx
+++ b/app/components/PlayerManager.tsx
@@ -11,9 +11,17 @@
 //   found at http://www.apache.org/licenses/LICENSE-2.0
 //   You may not use this file except in compliance with the License.
 
-import { PropsWithChildren, useCallback, useEffect, useMemo, useRef, useState } from "react";
+import {
+  PropsWithChildren,
+  useCallback,
+  useEffect,
+  useLayoutEffect,
+  useMemo,
+  useRef,
+  useState,
+} from "react";
 import { connect, ConnectedProps } from "react-redux";
-import { useAsync, useLocalStorage, useMountedState } from "react-use";
+import { useLocalStorage, useMountedState } from "react-use";
 import { URL } from "universal-url";
 
 import { AppSetting } from "@foxglove-studio/app/AppSetting";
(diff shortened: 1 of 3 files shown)


defunctzombie:
@jtbandes fyi - I did not tackle a larger reworking of source selection - focused on fixing the issue at hand for now. Source selection needs more thought and we can handle separately.

app/components/PlayerManager.tsx
```diff
@@ -437,62 +444,67 @@ function PlayerManager({
...
-    }
+  const [savedSource, setSavedSource] = useLocalStorage<PlayerSourceDefinition>(
+    "studio.playermanager.selected-source",
+  );
 
-    log.debug(`Select Source: ${selectedSource.name} ${selectedSource.type}`);
-    try {
-      metricsCollector.setProperty("player", selectedSource.type);
+  const selectSource = useCallback(
+    async (selectedSource: PlayerSourceDefinition, restore: boolean = false) => {
```
jtbandes, line 457: A boolean param!?

app/components/PlayerManager.tsx
```diff
@@ -437,62 +444,67 @@ function PlayerManager({
...
-    metricsCollector,
-    prompt,
-    rosHostname,
-    selectedSource,
-    setPlayer,
-    storage,
-  ]);
+    // we only run the layout effect on first mount - never again even if the saved source changes
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
jtbandes, line 504: Never thought I would see the day you disabled react-hooks/exhaustive-deps... How about [`useMount`](https://github.com/streamich/react-use/blob/master/docs/useMount.md)? (Which is just as bad really)

app/components/PlayerManager.tsx
defunctzombie: I know right? You can tell how annoyed I became with trying to rework player selection by how bad the api I leave behind is.

app/components/PlayerManager.tsx
defunctzombie: I thought about abusing useState() for it - but it somehow seemed worse...

There is no winning in PlayerManager - only various degrees of badness.

PR #829, opened 2021-05-01
Research into a web build and repo organization

defunctzombie:
This PR is not meant to merge. It is provided as a basis for discussion/experiments. Smaller PRs will incrementally reach this final state.

PR #836, opened 2021-05-02
Move typings modules into packages/@types

defunctzombie:
Rather than using a typeRoots override in tsconfig which disables the
automatic @types node_modules detection, we instead leverage our yarn
workspace to provide @types packages for modules which don't have types
and are not yet in definitely typed.

See https://github.com/foxglove/studio/pull/829


amacneil:
Great idea 👏

PR #839, opened 2021-05-02
Separate renderer entry points from app.

defunctzombie:
Renderer entry points are specific to the desktop build of studio.
Move the electron specific entry elements to desktop/renderer.

To avoid top-level directory confusion, desktop now contains: main,
preload, and renderer. Each has the code for their equivalent electron
space. The webpack configurations are moved from the top level into desktop
since they are for the desktop build.

App exposes a "webpack" file with a function to build a partial webpack
configuration required to integrate app into renderer.

App exposes a "typings" file with global type information assumed by app
code and required to integrate app into renderer.

See https://github.com/foxglove/studio/pull/829

---
- [x] ~~Move some items from renderer/App.tsx back to app/Workspace (or some level above workspace). Only the OsContext (maybe rename to Platform?) need to be in the renderer~~ I've decided for now to leave Workspace and App unchanged. when I dig into the OsContext changes for https://github.com/foxglove/studio/pull/840 it will be clearer what should belong where).
- [x] figure out how to handle `File.path` which is not available outside electron


app/Workspace.tsx
```diff
@@ -159,7 +159,8 @@ export default function Workspace(): JSX.Element {
       const otherFiles: File[] = [];
       for (const file of files) {
         try {
-          if (!(await loadFromFile(file, { basePath: file.path }))) {
+          // fixme - file.path is an electron thing - not available in web
```
defunctzombie, line 162: fixme

app/components/Chart/index.tsx
```diff
@@ -2,6 +2,9 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
+/// <reference types="chartjs-plugin-datalabels" />
```
defunctzombie, line 5: Add comment explaining why - I don't _really_ know why other than this was needed to make typescript happy with the type merging these modules perform on the chartjs type definitions.

app/typings/index.d.ts
```diff
@@ -0,0 +1,5 @@
+import "./extensions";
```
defunctzombie, line 1: Add a comment explaining what this is for.

desktop/main/extensions.d.ts
```diff
@@ -0,0 +1,11 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
defunctzombie, line 1: This is here because main imports an .scss file from app - I'd probably like to just put that color variable in a js file and do away with this and the sass loader we had to add to the main webpack.

desktop/renderer/globals.d.ts
```diff
@@ -0,0 +1 @@
+import "@foxglove-studio/app/typings";
```
defunctzombie, line 1: add comment explaining what is going on

desktop/renderer/tsconfig.json
```diff
@@ -0,0 +1,19 @@
...
+  "include": ["./**/*", "./**/*.json"],
+  "compilerOptions": {
+    "noEmit": true,
+    "jsx": "react-jsx",
+    "lib": ["dom", "dom.iterable", "es2020"],
+    "paths": {
+      // micro-memoize types do not work with our strict compiler settings
+      // This overrides the module provided types with our @types package.
+      // We use several node_module fallbacks to be flexible to where the @types package is installed.
+      "micro-memoize": [
```
defunctzombie, line 12: We need to kill micro-memoize - there are 3 uses in the codebase and its debatable if they are good or hiding bugs. There are other memoize packages in life.

app/.eslintrc.yaml
```diff
@@ -1,4 +1,4 @@
 settings:
   import/resolver:
     webpack:
-      config: ./webpack.renderer.config.ts
+      config: ./app/webpack.ts
```
defunctzombie, line 4: @jtbandes what is this resolver config for? This is no longer a valid webpack config but eslint seems to have run fine (locally) even with this value here...

app/Workspace.tsx
defunctzombie, line 162: @jtbandes Thoughts on how you think it makes sense to resolve this? Should loadFromFile internally understand a file _may_ have a `path` ? Or we have some other wrapper for files? Another option might be to have a different type of asset loader for render vs web and we move creating the specific type out of workspace.

jest.config.json
```diff
@@ -1,8 +1,14 @@
 {
-  "//": "Our top level jest config to reference all the projects under our monorepo. We exclude the integration test since it involves performing a webpack build and doesn't support 'watch' or 'debug' in the same way",
+  "//": "Our top level jest config to reference all the projects under our monorepo. We exclude the desktop test since it involves performing a webpack build and doesn't support 'watch' or 'debug' in the same way",
   "projects": [
+    "<rootDir>/jest.config.json",
```
jtbandes, line 4: Why is this here? Is the file referencing itself?

licenses.test.ts
```diff
@@ -0,0 +1,54 @@
...
+            reject(err);
+            return;
+          }
+          resolve(modules);
+        },
+      );
+    });
+
+    // eslint-disable-next-line
+    console.log((checker as any).asSummary(packages));
```
jtbandes, line 52: Why?

PR #840, opened 2021-05-02
[RFC] Exploring OsContextSingleton 🔪 

defunctzombie:
I regret adding OsContextSingleton. It has become a place to jam things into and sprinkle globals throughout the app. Several uses would benefit from more react idiomatic approaches. Other uses should make the callers pass in the values. And there are uses which are tricky or will require a global. We should limit what is in that global tho rather than "anything on the context" as we have the singleton do today.

This PR is not meant for merge - only discussion.


defunctzombie:
https://github.com/foxglove/studio/pull/873

defunctzombie:
Remaining OsContextSingletons in app

```
// windowEvents? or some other form of toggle "fullscreen"
Workspace.tsx:        <Toolbar onDoubleClick={OsContextSingleton?.handleToolbarDoubleClick}>

// master uri should be provided elsewhere? tricky - this is where players are built!
// but we also don't want random env var fetching
components/PlayerManager.tsx:      value: value ?? OsContextSingleton?.getEnvVar("ROS_MASTER_URI") ?? "localhost:11311",

// deep links - these are the argv preload may provide to trigger commands
// this supports opening bag files with foxglove:// urls
// change this to invoke selectSource on the player and provide the url to open
// at least it will remove oscontextsingleton from this use in player manager
components/PlayerManager.tsx:    const links = OsContextSingleton?.getDeepLinks() ?? [];

// in getRosHostname, os.getEnv, getNetworkInterfaces, getHostname are all used to call
// RosNode.GetRosHostname(
// The same thing happens in ros1 player
components/Preferences.tsx:  const os = OsContextSingleton;
players/Ros1Player.ts:    const os = OsContextSingleton;

// could replace with static process.platform?
panels/WelcomePanel/index.tsx:    (OsContextSingleton && OsContextSingleton.platform === "darwin" && "⌘") ?? "^";

// app version we can get from version.ts
// machine id tho ... delete?
services/Analytics.ts:    return OsContextSingleton?.getAppVersion() ?? "0.0.0";
services/Analytics.ts:    return (await OsContextSingleton?.getMachineId()) ?? UUID_ZERO;

Issue #761, opened 2021-04-28
Consolidate memoization libs


amacneil:
via @defunctzombie:

We have both `memoize-weak` and `micro-memoize`. We do not need two. Find one that is modern, maintain, and meets our needs.

jtbandes:
`export getMemoize = memoizeWeak(microMemoize)`

defunctzombie:
🔪

PR #841, opened 2021-05-02
Remove micro-memoize

defunctzombie:
We have micro-memoize and memoize-weak packages which leads to the question: "which do I use"? Who knows but not both. memoize-weak is 4 years old - looks unmaintained - but is also short and simple? micro-memoize seems newer - is more complex - and has type definitions which don't work with our compiler settings (why they write their .d.ts file by hand when they use typescript I have no idea.

This change removes all uses of micro-memoize. In several places memoization should be done at another layer. I'm a fan of memoization when its obvious it is happening - maybe more explicit. Generally I expect my function calls to call the function.

Closes https://github.com/foxglove/studio/issues/761


app/players/OrderedStampPlayer.ts
```diff
@@ -37,10 +37,10 @@ import { clampTime, isTime, TimestampMethod } from "@foxglove-studio/app/util/ti
 // than their receive times.
 export const BUFFER_DURATION_SECS = 1.0;
 
-const getTopicsWithHeader = microMemoize((topics: Topic[], datatypes: RosDatatypes) => {
```
defunctzombie, line 40: I think memoizeWeak will drop the memoization if nothing hangs on to the input values - thats what it sells itself as. Is that ok here? Or will lead to the memoization not happening?

app/players/UserNodePlayer/index.ts
```diff
@@ -156,7 +157,7 @@ export default class UserNodePlayer implements Player {
 
   // When updating nodes while paused, we seek to the current time
   // (i.e. invoke _getMessages with an empty array) to refresh messages
-  _getMessages = microMemoize(
+  _getMessages = memoizeWeak(
```
defunctzombie, line 160: Something feels _off_ to me about this pattern of memoizing methods...but there are too many other things going on.

app/players/UserNodePlayer/index.ts
```diff
@@ -216,11 +239,21 @@ export default class UserNodePlayer implements Player {
...
   // Defines the inputs/outputs and worker interface of a user node.
   private _createNodeRegistration = async (
     nodeId: string,
     userNode: UserNode,
   ): Promise<NodeRegistration> => {
+    for (const cacheEntry of this._nodeRegistrationCache) {
+      if (nodeId === cacheEntry.nodeId && isEqual(userNode, cacheEntry.userNode)) {
+        return cacheEntry.result;
+      }
+    }
```
jtbandes, line 256: Changed to an explicit linear search of an array. 🤷

PR #843, opened 2021-05-03
Change AppConfigurationContext.get to sync

defunctzombie:
Load all settings on app startup and provide configuration values to
useAppSettings on first mount. This avoids additional renders and state
changes since we've loaded all the settings to launch the app.

---
When looking into `OsContextSingleton` uses in https://github.com/foxglove/studio/pull/840, I noticed AnalyticsProvider not using `useAppSettings`. When I switched over to `useAppSettings` the analytics provider would go through several state changes (and render passes) as a result of `useAsyncAppConfigurationValue` and AnalyticsContext getter behavior. Since we want to load all settings on app launch, I changed the OsContextAppSettings provider to wait until the settings have loaded and the `get` for each setting now returns the _current_ setting value.


defunctzombie:
@jtbandes When looking at this - I think it would be good to roll experimental features into app settings. The separate _tab_ in app settings should remain - the persisted values should be managed via app settings tho (in an `experimental` namespace). I will tackle that once this lands.

app/context/AppConfigurationContext.ts
```diff
@@ -4,12 +4,18 @@
 
 import { createContext, useContext } from "react";
 
+export type ChangeHandler = (newValue: unknown) => void;
+
 // Exposes an interface for reading and writing user-configurable options and other persistent application state.
 export interface AppConfiguration {
-  get(key: string): Promise<unknown | undefined>;
+  // get the current value for the key
+  get(key: string): unknown;
```
defunctzombie, line 12: Tragically there is not a way to say "Don't return a promise" - maybe the answer is to more strictly type this - getters should return `string | number` ?

app/context/AnalyticsProvider.tsx
```diff
@@ -4,23 +4,27 @@
...
 export default function AnalyticsProvider(
   props: PropsWithChildren<{ amplitudeApiKey?: string }>,
 ): React.ReactElement {
+  const enableTelemetry = useAppSetting<boolean>(AppSetting.TELEMETRY_ENABLED);
+  const enableCrashReporting = useAppSetting<boolean>(AppSetting.CRASH_REPORTING_ENABLED);
+
   const analytics = useMemo(() => {
     return new Analytics({
-      optOut: !(OsContextSingleton?.isTelemetryEnabled() ?? true),
+      optOut: !(enableTelemetry ?? true),
```
jtbandes, line 20: nit: move `?? true` to initialization

app/hooks/useAsyncAppConfigurationValue.ts
```diff
@@ -20,44 +20,41 @@ type OptimisticState<T> =
...
 ): [state: OptimisticState<T | undefined>, setter: (value?: T) => Promise<void>] {
   const appConfiguration = useAppConfiguration();
 
-  // async retry will start loading the value on mount
-  const getterState = useAsyncRetry(
-    async () => (await appConfiguration.get(key)) as T | undefined,
-    [appConfiguration, key],
+  const [optimisticValue, setOptimisticValue] = useState<T | undefined>(
+    appConfiguration.get(key) as T | undefined,
   );
```
jtbandes, line 33: This no longer correctly named as an "optimistic" value since it is unconditionally returned when the setterState is not loading/error. Consider whether you should remove the `optimistic` option entirely.

app/services/OsContextAppConfiguration.ts
```diff
@@ -4,35 +4,43 @@
...
-        OsContextAppConfiguration.STORE_KEY,
-        { encoding: "utf8" },
-      );
-      this._currentValue = JSON.parse(value ?? "{}");
-    });
+    this._currentValue = initialValue;
   }
 
-  async get(key: string): Promise<unknown | undefined> {
-    return await this._mutex.runExclusive(
```
jtbandes, line 33: So I know removing this is the whole point of the PR, but is it really safe? You could now have concurrent set() and get() calls. Gets will return the old value until the set finishes.

app/services/OsContextAppConfiguration.ts
```diff
@@ -4,35 +4,43 @@
...
+  ChangeHandler,
+} from "@foxglove-studio/app/context/AppConfigurationContext";
 
 export default class OsContextAppConfiguration implements AppConfiguration {
   static STORE_NAME = "settings";
   static STORE_KEY = "settings.json";
 
   private readonly _ctx: Pick<OsContext, "storage">;
-  private _listeners = new Map<string, Set<() => void>>();
+  private _listeners = new Map<string, Set<(newValue: unknown) => void>>();
```
jtbandes, line 17: ChangeHandler?

app/services/OsContextAppConfiguration.ts
```diff
@@ -4,35 +4,43 @@
...
-      this._currentValue = JSON.parse(value ?? "{}");
-    });
+    this._currentValue = initialValue;
   }
 
-  async get(key: string): Promise<unknown | undefined> {
-    return await this._mutex.runExclusive(
-      () => (this._currentValue as Record<string, unknown>)[key],
+  // create a new OsContextAppConfiguration
+  static async initialize(ctx: Pick<OsContext, "storage">): Promise<OsContextAppConfiguration> {
```
jtbandes, line 30: You and @jhurliman need to duke it out over how we are capitalizing static methods

app/hooks/useAppSetting.ts
```diff
@@ -5,8 +5,10 @@
 import { AppSetting } from "@foxglove-studio/app/AppSetting";
 import { useAsyncAppConfigurationValue } from "@foxglove-studio/app/hooks/useAsyncAppConfigurationValue";
 
+// Track a single app configuration setting
```
jtbandes, line 8: Prefer jsdoc comments for IDE integration -- and what does "track" mean? maybe explicitly mention that the component will re-render when the config value changes?

app/context/OsContextAppConfigurationProvider.tsx
```diff
@@ -2,25 +2,38 @@
...
     }
-    return new OsContextAppConfiguration(OsContextSingleton);
+
+    log.debug("Initializing app configuration");
+    return OsContextAppConfiguration.initialize(OsContextSingleton);
   }, []);
 
+  if (error) {
+    throw error;
+  }
```
jtbandes, line 29: Just pondering - seems like we are opening ourselves up to more bugs where a problem with filesystem errors or slowness leads to the app being completely unusable.

jtbandes:
> the analytics provider would go through several state changes

Why several? Is this really a big deal for top level providers?

app/services/OsContextAppConfiguration.ts
defunctzombie: Are we capitalizing static methods? I guess we are?

app/context/OsContextAppConfigurationProvider.tsx
defunctzombie: What do you think should happen in this case? It is problematic if we cannot open the user's settings since the app would be using completely unexpected settings if we can't open the settings?

app/services/OsContextAppConfiguration.ts
defunctzombie: We'd need to update OsContextAppConfiguration to optimistically have the new value. Practically tho if the user is using useAppSetting then their state is updated when the `set` completes and invokes listeners.

defunctzombie:
> Why several? Is this really a big deal for top level providers?

The getter was async and would set _loading_ and then unset loading. I think it is nice to lower the amount of re-renders happening for this since settings are something we have loaded once.

app/hooks/useAsyncAppConfigurationValue.ts
```diff
@@ -2,14 +2,12 @@
...
-  // If true, a newly set value will be returned from the hook even while the setter is still in progress.
-  optimistic?: boolean;
-};
+import {
+  AppConfigurationValue,
+  useAppConfiguration,
+} from "@foxglove-studio/app/context/AppConfigurationContext";
 
 // Like AsyncState<T>, but allowing `loading: true` & `value` to be present at the same time.
 type OptimisticState<T> =
```
jtbandes, line 13: maybe not needed?

app/hooks/useAppSetting.ts
```diff
@@ -14,10 +14,10 @@ import { useAsyncAppConfigurationValue } from "@foxglove-studio/app/hooks/useAsy
 export default function useAppSetting<T extends AppConfigurationValue>(
   setting: AppSetting,
 ): T | undefined {
-  const [entry] = useAsyncAppConfigurationValue<T>(setting);
+  const [entry] = useAppConfigurationValue<T>(setting);
   // treat empty strings as undefined (i.e. unset values)
-  if (typeof entry.value === "string" && entry.value.length === 0) {
+  if (typeof entry === "string" && entry.length === 0) {
     return undefined;
   }
```
jtbandes, line 21: We have a `nonEmptyOrUndefined` function for this now – also not convinced of the value of useAppSetting over useAppConfigurationValue. Can we just move this functionality in there?

app/panels/WelcomePanel/index.tsx
```diff
@@ -35,7 +35,7 @@ function validateEmail(str: string): string | undefined {
 }
 
 function WelcomePanel() {
-  const [subscribedState, setSubscribed] = useAsyncAppConfigurationValue<boolean>(
+  const [subscribed = false, setSubscribed] = useAppConfigurationValue<boolean>(
```
jtbandes, line 38: You'll need to update the stories for this component

Issue #848, opened 2021-05-03
Missing spinner UI when rosbridge connection is lost


amacneil:
1. Start FGS (latest main, post-error handling cleanup)
2. Connect to websocket. You get a nice spinner:
![image](https://user-images.githubusercontent.com/637671/116938982-68ac8600-ac20-11eb-8a10-7d0e7ac51325.png)

3. Start rosbridge (`roslaunch rosbridge_server rosbridge_websocket.launch`). Now we are connected:
![image](https://user-images.githubusercontent.com/637671/116939034-80840a00-ac20-11eb-9e57-33e7779be7fc.png)

4. Shut down rosbridge. FGS continues to attempt connections (per devtools), but we no longer show a spinner, just "Connection failed":
![image](https://user-images.githubusercontent.com/637671/116939097-9abde800-ac20-11eb-8b63-8376cfcebff1.png)

PR #850, opened 2021-05-04
Fix app crash when setPublishers called before Rosbridge connects

### LINKED ISSUE #679: Publish panel + websocket player = BSOD
- Connect to Websocket bridge
- Open a new Publish panel
- Whole app crashes irreparably

via @jtbandes

defunctzombie:
Rosbridge.setPublishers would check if it had a valid connection and
throw an Error if it did not. This resulted in a race between the publish
panel which invokes setPublishers (indirectly via usePublisher) on players
when a player is available and Rosbridge player not being connected when
setPublishers is invoked. This exposed an additional bug in Rosbridge
with publishers - publishers would not be re-created on reconnect.

This change fixes both issues. Instead of throwing in Rosbridge player if
there is no connection, store the advertise list. When a connection is established
the player creates the publishers from the advertise list. If a connection already
exists, we are able to make the publishers immediately. When a connection drops
and reconnects, the publishers are re-created.

An additional fix moves the ErrorBoundary in App.tsx to capture _all_ child elements
and display an error to the user if an exception is throw and unhandled anywhere in
the component tree. Prior to this move, the throw in setPublishers would result in
a blank app for the user since nothing visually surfaced the error.

Fixes #679,#848

### DIFF
--- a/app/players/RosbridgePlayer.ts
+++ b/app/players/RosbridgePlayer.ts
@@ -77,15 +77,15 @@ export default class RosbridgePlayer implements Player {
   private _parsedMessages: MessageEvent<unknown>[] = []; // Queue of messages that we'll send in next _emitState() call.
   private _messageOrder: TimestampMethod = "receiveTime";
   private _requestTopicsTimeout?: ReturnType<typeof setTimeout>; // setTimeout() handle for _requestTopics().
-  private _topicPublishers: {
-    [topicName: string]: roslib.Topic;
-  } = {};
+  // active publishers for the current connection
+  private _topicPublishers = new Map<string, roslib.Topic>();
+  // which topics we want to advertise to other nodes
+  private _advertisements: AdvertisePayload[] = [];
   private _parsedMessageDefinitionsByTopic: ParsedMessageDefinitionsByTopic = {};
   private _parsedTopics: Set<string> = new Set();
   private _receivedBytes: number = 0;
   private _metricsCollector: PlayerMetricsCollectorInterface;
   private _hasReceivedMessage = false;
-
   private _presence: PlayerPresence = PlayerPresence.NOT_PRESENT;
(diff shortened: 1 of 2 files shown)


app/players/RosbridgePlayer.ts
```diff
@@ -77,15 +77,17 @@ export default class RosbridgePlayer implements Player {
   private _parsedMessages: MessageEvent<unknown>[] = []; // Queue of messages that we'll send in next _emitState() call.
   private _messageOrder: TimestampMethod = "receiveTime";
   private _requestTopicsTimeout?: ReturnType<typeof setTimeout>; // setTimeout() handle for _requestTopics().
-  private _topicPublishers: {
-    [topicName: string]: roslib.Topic;
-  } = {};
+  // active publishers for the current connection
+  private _topicPublishers = new Map<string, roslib.Topic>();
```
jtbandes, line 81: one by one, they shall all be replaced

app/players/RosbridgePlayer.ts
```diff
@@ -77,15 +77,17 @@ export default class RosbridgePlayer implements Player {
...
   private _parsedMessageDefinitionsByTopic: ParsedMessageDefinitionsByTopic = {};
   private _parsedTopics: Set<string> = new Set();
   private _receivedBytes: number = 0;
   private _metricsCollector: PlayerMetricsCollectorInterface;
   private _hasReceivedMessage = false;
 
+  // which topics we want to advertise to other nodes
+  private _advertise: AdvertisePayload[] = [];
```
jtbandes, line 89: `_advertisements`? `_publishRequests`?

This should probably be next to topicPublishers since they are closely related

app/players/RosbridgePlayer.ts
defunctzombie: 🔪

PR #851, opened 2021-05-04
Add @foxglove/hooks package with useRethrow hook.

defunctzombie:
The useCatch hook surfaces errors thrown in a function handler
into the react render context. This allows these errors to bubble up
to an error boundary.

Update onPublishClick within the publish panel to useCatch when
publishing so publish errors are localized to the panel.

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -8,6 +8,7 @@
   "dependencies": {
     "@fluentui/react": "8.14.0",
     "@fluentui/react-icons-mdl2": "1.1.0",
+    "@foxglove/hooks": "workspace:packages/@foxglove/hooks",
     "@foxglove/velodyne-cloud": "workspace:packages/velodyne-cloud",
     "@mdi/svg": "5.9.55",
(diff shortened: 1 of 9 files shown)


packages/@foxglove/hooks/src/useCatch.ts
```diff
@@ -0,0 +1,35 @@
...
+export default function useCatch<Args extends unknown[], Ret>(
+  fn: Fn<Args, Ret>,
+): Fn<Args, Ret | void> {
+  const [_, setError] = useState(undefined);
+  return useCallback(
+    (...args: Args): Ret | void => {
+      try {
+        return fn(...args);
+      } catch (err) {
+        // throwing within a setError happens within a react render context
```
jtbandes, line 27: link to documentation on this?

defunctzombie: Is there documentation on this or are we just hoping for the best?

PR #857, opened 2021-05-04
Fix rosparam subscriptions when new parameters are added to a running system

jhurliman:


defunctzombie:
Test?

jhurliman:
@defunctzombie testing these methods is still blocked on an implementation of rosmaster/rosparam server (we can make a convincing mock, but it's about the same amount of work since it's ultimately just a key/value store that adheres to a particular API).

PR #859, opened 2021-05-04
10 breadcrumbs should be enough for anybody

defunctzombie:


amacneil:
Is this like “most recent 10” or “first 10”?

defunctzombie:
@amacneil https://github.com/getsentry/sentry-javascript/blob/a6f8dc26a4c7ae2146ae64995a2018c8578896a6/packages/hub/src/scope.ts#L341 most recent

PR #863, opened 2021-05-04
Fetch machineId asynchronously

amacneil:
`OsContextSingleton.getMachineId()` was taking around 2s on my machine, and since we call it twice (when initializing analytics) was leading to about a 4s startup delay for production builds.

Changed to call this method asynchronously, and kick it off at the start of preload, so that it is ready by the time analytics needs it.

Also, initializing analytics now happens async, and `logEvent()` will await until it is ready.

Paired with @defunctzombie

### DIFF
--- a/app/services/Analytics.ts
+++ b/app/services/Analytics.ts
@@ -2,7 +2,11 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
-import { addBreadcrumb, setUser, Severity } from "@sentry/electron";
+import {
+  addBreadcrumb as addSentryBreadcrumb,
+  setUser as setSentryUser,
+  Severity,
+} from "@sentry/electron";
 import amplitude from "amplitude-js";
 import { v4 as uuidv4 } from "uuid";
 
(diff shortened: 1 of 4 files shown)


app/services/Analytics.ts
```diff
@@ -80,13 +94,14 @@ export class Analytics {
     return userId;
   }
 
-  getDeviceId(): string {
-    return OsContextSingleton?.getMachineId() ?? UUID_ZERO;
+  // OsContextSingleton.getMachineId() can take 500-2000ms on macOS
+  async getDeviceId(): Promise<string> {
+    return (await OsContextSingleton?.getMachineId()) ?? UUID_ZERO;
```
jtbandes, line 99: Off topic for this PR, but why do we have a UUID_ZERO instead of generating a new, random UUID when we can't get one from the machine?

In fact we could persist our own unique ID in the app data directory instead of generating a new one each time. This might be preferable to using the "machine id" for privacy reasons.

amacneil: Yeah, definitely more could be cleaned up here. Looking at the [library source](https://github.com/automation-stack/node-machine-id/blob/master/index.js) I think we will actually reject if we can't find a machine ID, which means `UUID_ZERO` is only used when `OsContextSingleton` is undefined, which I don't understand why that is even an optional.

jhurliman: `UUID_ZERO` is a fallback for situations like unit testing or some sort of app embedding where we don't have an `OsContextSingleton`. This will never happen in the normal course of app usage.

PR #864, opened 2021-05-04
Type error in test doesn't fail build

defunctzombie:


defunctzombie:
Our CI pipeline has a `build:packages` step which runs `tsc` on all items within `packages`. A proposed solution is to move `app` into packages/studio or similar.

jtbandes:
Has this issue been resolved now?

defunctzombie:
Yep https://github.com/foxglove/studio/pull/1249

PR #868, opened 2021-05-04
Add params to selectSource

### LINKED ISSUE #865: Loading spinner comes back after refresh even if a bag file source was loaded
Steps to reproduce:

- Connect to a ROS or websocket source
- Drop a local bag file
- Refresh the page
- Loading spinner appears. Expected: no spinner

defunctzombie:
Source params add a way to provide additional options to the
selected source. These options instruct the player on how to build
the source.
    
Fixes #865

### DIFF
--- a/app/components/PlayerManager.tsx
+++ b/app/components/PlayerManager.tsx
@@ -71,8 +71,6 @@ import Logger from "@foxglove/log";
 
 const log = Logger.getLogger(__filename);
 
-const DEMO_BAG_URL = "https://storage.googleapis.com/foxglove-public-assets/demo.bag";
-
 type BuiltPlayer = {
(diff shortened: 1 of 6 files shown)


app/components/PlayerManager.tsx
```diff
@@ -507,6 +507,12 @@ function PlayerManager({
     selectSource,
     setPlayerFromFiles: useCallback(
       (files: File[], { append = false }: { append?: boolean } = {}) => {
+        // save that we last loaded a source from local bag file
+        setSavedSource({
+          name: "Bag File (local)",
```
jtbandes, line 512: Isn't this name used later for lookup? Seems brittle to hardcode this string

PR #874, opened 2021-05-05
Move experimental features to app settings

defunctzombie:
Removes the separate experimental features context in favor
of the more recent app configuration context. Experimental features
are a subset of app settings.


defunctzombie:
@jtbandes This drops the "development default" and "production default" features. Where we should put them back is defaults for app settings. Right now defaults are generally handled at the use site - which is ok for the most part - but there are some places where the default is duplicated. It would be nice to have one place for the app settings defaults (probably where we load them?)

defunctzombie:
Oh - I also think it could be ok to move these under the "General" tab - maybe an "experimental" sub-heading?

app/AppSetting.ts
```diff
@@ -7,4 +7,5 @@ export enum AppSetting {
   ROS1_ROS_HOSTNAME = "ros1.ros_hostname",
   TELEMETRY_ENABLED = "telemetry.telemetryEnabled",
   TIMEZONE = "timezone",
+  UNLIMITED_MEMORY_CACHE = "experimental.unlimited-memory-cache",
```
defunctzombie, line 10: I think we should change this to be an object of keys (and nested objects) with the values being the defaults?

PR #876, opened 2021-05-05
Move connection selection to the sidebar

defunctzombie:
Screen Shot 2021-05-05 at 11 37 23 AM

Screen Shot 2021-05-05 at 11 38 02 AM


defunctzombie:
@amacneil to play around with this and give some thoughts

defunctzombie:
@jtbandes one UX aspect that I didn't like with a connection sidebar is that once I select a connection the sidebar stays open. In this case I would like to have the sidebar go away once I select a connection.

Another thought @amacneil had was once a connection is selected - the display would have the connection info (rather than separate panel for that). So once you are connected you have the details in the sidebar. That does mean there would need to be a way to make a new connection tho.

amacneil:
+1 to my idea above - I think having the "connection" tab double as a "new connection" and "connection info" (data source info) would work well. data source info being a panel feels weird to me since it is metadata not data.

PR #883, opened 2021-05-05
Load new StudioWindows when creating via File -> New Window

defunctzombie:
new StudioWindow does not _load_ the content by default. This was done
so main startup could create the window and load the page once more
setup items were in place. Once the app is ready - future windows
should load on create.

### DIFF
--- a/desktop/main/StudioWindow.ts
+++ b/desktop/main/StudioWindow.ts
@@ -141,7 +141,7 @@ function buildMenu(browserWindow: BrowserWindow): Menu {
       {
         label: "New Window",
         click: () => {
-          new StudioWindow();
+          new StudioWindow().load();
         },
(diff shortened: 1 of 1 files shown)


jtbandes:
Can we add the New Window flow to the integration test?

defunctzombie:
> Can we add the New Window flow to the integration test?

We'll need to figure out how to access electron.app so we can get the app menu and trigger File -> New Window. Last I recall this wasn't available to us anymore due to some context/security changes.

defunctzombie:
@jtbandes I'm punting on the integration test for now - There are some limitations with spectron and newer electron versions that make this "not so simple". the main hurdle being no access to the electron _app_ so we can get the app menu and trigger the menu action.

PR #884, opened 2021-05-06
Fix handling of inverted min/max ranges from user input.

### LINKED ISSUE #875: ChartJSManager.update: cannot read property '...' of undefined
Steps:
- Open a Plot panel and plot some data
- Open the panel settings menu and type something like `-100` in the Y Max field (such that no data is visible)

image

defunctzombie:
If a user entered a max value which was smaller than any datapoint value,
chart.js would crash when computing linear scales. We fix this by
normalizing min/max within TimeBasedChart so min < max.

This change also makes it so that when the user removes their manual
override of the y-axis bounds, the chart returns back to dataset bounds.
Unfortunately we need to set the y-axis bounds manually in that case
possibly due to chartjs not re-calculating the bounds when we set min/max
back to undefined.

Fixes #875


jtbandes:
This seems to improve things but some of the screenshot changes look bad (missing points/y axes; missing dashed reference line).

app/components/TimeBasedChart/index.tsx
```diff
@@ -608,15 +608,53 @@ export default memo<Props>(function TimeBasedChart(props: Props) {
       padding: 0,
     };
 
+    hasUserPannedOrZoomed;
```
jtbandes, line 611: remove?

jtbandes (changes requested):

defunctzombie:
> This seems to improve things but some of the screenshot changes look bad (missing points/y axes; missing dashed reference line).

The degenerate (single point) case needs fixing. I'n this change I am calculating the ymin/max rather than using chartjs auto-axis and didn't handle single point. I tried using chartjs auto-axis for the y-axis but those do not reset when I set the min/max to undefined - thus having to do the bound calculation myself.

defunctzombie:
I've removed the y-axis bounds set to data in this PR as it breaks the automatic detection of the annotations (dashed lines) which is separate from the dataset.

Issue #877, opened 2021-05-05
Plot settings "Set to current view" button doesn't seem to do anything


jtbandes:
It looks like this button in the Plot settings menu is broken. Nothing happens when I click it.

<img width="301" alt="image" src="https://user-images.githubusercontent.com/14237/117195497-09b25280-ad81-11eb-928d-9652ace5aa18.png">

jtbandes:
I deleted this button in #917. Hopefully it's not too important 😅

Issue #793, opened 2021-04-28
"Status bar" to explain panel selection interactions


jtbandes:
Currently our panel selection (⌘/⌃-click) shows some explanation text when you hover on panels, but when you've just selected the first panel it's not very obvious why there's a blue outline or what you can do by selecting panels. Similarly, the `` ` `` "quick actions" key is not very discoverable.

<img align="top" width="281" alt="Screen Shot 2021-04-28 at 10 15 34 AM" src="https://user-images.githubusercontent.com/14237/116466819-acf7ea80-a80a-11eb-9165-2b6434fe260b.png"> <img width="479" alt="image" src="https://user-images.githubusercontent.com/14237/116466718-8f2a8580-a80a-11eb-88ee-ced8b8c16412.png">

Affinity Designer (and presumably other editing tools) have a nice status bar at the bottom that helps you understand what actions are available, which modifier keys to use, etc. Something like this might help our users better understand the available panel actions:

<img width="979" alt="Screen Shot 2021-04-28 at 10 11 46 AM" src="https://user-images.githubusercontent.com/14237/116466482-3e1a9180-a80a-11eb-9846-ee8c6ff79b3f.png">
<img width="979" alt="Screen Shot 2021-04-28 at 10 11 48 AM" src="https://user-images.githubusercontent.com/14237/116466489-407ceb80-a80a-11eb-9f6e-565ba1b1a5cb.png">
<img width="911" alt="image" src="https://user-images.githubusercontent.com/14237/116467185-1f68ca80-a80b-11eb-9e14-c831247fdc60.png">

jtbandes:
This would also be useful for #2719.

Issue #775, opened 2021-04-28
Joystick/gamepad/teleop panel


amacneil:
Feature request from @rgov on Slack:

Currently we have a Publish panel, but it is limited to only creating one button. It would be nice to create a panel to allow basic control of the robot. Idea would be to support both mouse/keyboard control via UI, and/or to support USB joysticks via HID.

We should publish `geometry_msgs/Twist` messages.

For example: https://github.com/WHOIGit/ros-web-gamepad.

Need to be able to customize the publish topic name. For example, duckiebot publishes to `/robotname/joy`.

jtbandes:
Example prototype implementation: https://github.com/cruise-automation/webviz/pull/561

PR #885, opened 2021-05-06
Add panel settings to sidebar

jtbandes:
Adds a new Panel Settings tab to the sidebar, which shows settings for the **currently selected** panel. The settings are pulled from a static `configSchema` which the panel type can export.

This PR migrates panel settings from the PanelToolbar menu to the new `configSchema` for Publish, Node Playground, and Image panels. The remaining panels with settings in the menu (search for `menuContent=`) are Plot and 3D panels. These both have more complicated settings (and Plot settings are broken #875 #877) so they are not being migrated for now.

TODO:
- [ ] Global Variable Slider settings
- [ ] single click to select panel when settings sidebar is open
- [ ] automatically open settings after adding a panel
- [ ] automatically deselect panel when sidebar closes
- [ ] link from panel toolbar menu to edit settings
- [ ] delete panel while viewing settings -> settings should close (just remove from selectedPanelIds?)


app/context/PanelIdContext.ts
```diff
@@ -0,0 +1,16 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { createContext, useContext } from "react";
+
+export const PanelIdContext = createContext<string | undefined>(undefined);
+
+export function usePanelId(): string {
```
defunctzombie, line 9: Feels to me like this is more of a ActivePanelContext? or possibly this is a property of the Layout? We don't have a layout context - I imagine that would handle the various tasks around panels - selecting -etc?

jtbandes: Maybe... does the notion of "selected panels" belong in the same place as "which panel am I rendering in"?

I originally added this because the settings sidebar code needs to know which panel it is rendering settings for. Now that panels only provide a settings schema rather than a settings component, it's actually only being used in the implementation of useConfig. I imagine we will prefer to migrate from our current props.config to useConfig, but I don't know if there are other use cases that will require the panel id, which is why it's not currently part of the PanelAPI.

app/panels/Publish/index.test.tsx
```diff
@@ -1,61 +0,0 @@
...
-          config={{
-            topicName: "/sample_topic",
-            datatype: "std_msgs/String",
-            buttonText: "Publish",
-            buttonTooltip: "",
-            buttonColor: "",
-            advancedView: true,
-            value: `{ "data": "hello" }`,
-          }}
-          saveConfig={saveConfig}
```
jtbandes, line 52: Removed this test because it passed in a custom saveConfig function, which I am removing from panel props in this PR. Currently there is no replacement other than providing a mock redux store, since this is not a common use case.

I believe the Publish panel will need to be reworked as we add ROS publishing support (#697 / #775) so I'm not too concerned with minor behavioral regressions like the one that was being tested for here.

app/components/PanelSettings/index.tsx
```diff
@@ -0,0 +1,139 @@
...
+      <ShareJsonModal
+        onRequestClose={() => setShowShareModal(false)}
+        value={panelConfigById[selectedPanelId] ?? {}}
+        onChange={(config) =>
+          dispatch(savePanelConfigs({ configs: [{ id: selectedPanelId, config, override: true }] }))
+        }
+        noun="panel configuration"
+      />
+    );
+  }, [selectedPanelId, showShareModal, store, dispatch]);
```
jtbandes, line 63: Moved from PanelToolbar

app/panels/ThreeDimensionalViz/stories/storyComponents.tsx
```diff
@@ -119,14 +128,11 @@ export class FixtureExample extends React.Component<FixtureExampleProps, Fixture
     return (
       <PanelSetup fixture={fixture} onMount={this.props.onMount}>
         <Flex col>
-          <ThreeDimensionalViz
-            config={this.state.config as any}
-            saveConfig={(config) => this.setState({ config: { ...this.state.config, ...config } })}
-          />
```
jtbandes, line 125: Attempted a fix here for the removal of saveConfig, but these stories are already broken anyway and need to be revisited at some point.

app/components/PanelSettings/index.tsx
```diff
@@ -0,0 +1,139 @@
...
+  const [config, saveConfig] = useConfigById<Record<string, unknown>>(
+    selectedPanelId,
+    panelInfo?.component.defaultConfig,
+  );
+
+  if (selectedPanelId == undefined) {
+    return (
+      <SidebarContent title={`Panel Settings`}>
+        <Text styles={{ root: { color: theme.palette.neutralTertiary } }}>
+          Select a panel to edit its settings.
```
esthersweon, line 74:
```suggestion
          Select a panel (`Ctrl` + click) to edit its settings.
```

app/components/PanelSettings/index.tsx
```diff
@@ -0,0 +1,139 @@
...
+                <SchemaEditor
+                  configSchema={panelInfo.component.configSchema}
+                  config={config}
+                  saveConfig={saveConfig}
+                />
+              </PanelIdContext.Provider>
+            </StrictMode>
+          ) : (
+            <Text styles={{ root: { color: theme.palette.neutralTertiary } }}>
+              No additional settings available.
```
esthersweon, line 102:
```suggestion
              No settings available.
```

esthersweon: Could include helpful messaging for when there is more than 1 panel selected? 

Screen Shot 2021-05-06 at 4 57 00 PM

jtbandes: This is a tough one – most panels are configurable **in the main panel UI**, but not many of them have these "extra" settings that would show up in the sidebar. Maybe we should move more settings into the sidebar ... but for now what language do you think we should use here to avoid confusing people?

app/components/PanelSettings/index.tsx
esthersweon: Add relevant "Click a panel to deselect" copy when a panel is selected.

app/components/PanelSettings/index.tsx
jtbandes: I think messaging for when multiple panels are selected would be great. I filed https://github.com/foxglove/studio/issues/793 to track that (with some screenshots showing a similar concept in other apps)

app/components/PanelSettings/index.tsx
jtbandes: Actually this PR makes it so that when the sidebar is open, you can just click the panel

esthersweon:
I feel uneasy about the fact that certain panel's settings are now fragmented to live in 2 different places – the panel's UI, or this sidebar.

Example below – `Diagnostics` panel has a `selectedHardwareId` setting that users can only specify in the panel itself.
Screen Shot 2021-05-06 at 5 02 51 PM

Should we include settings like this as read-only lines in the sidebar? Or should we allow users to edit these types of settings in both places? Or should the sidebar be the only place we edit settings, and the panel UI should only reflect those decisions with read-only text? 

Thinking out loud here – I personally like the 2nd option. I think the information in the sidebar should mirror the panel's settings JSON.

jtbandes:
> Or should we allow users to edit these types of settings in both places?

I like this option too. We discussed in Slack that there are some challenges to this, because many panels have dynamic settings: for instance, Diagnostics needs to parse messages on the `/diagnostics` topic to determine the list of available nodes, and Image needs to filter the currently available topics by their datatype. @defunctzombie suggested something like a function that allows panels to dynamically set their configSchema at runtime. I think we'll need to play with these ideas as follow-on work.

Issue #774, opened 2021-04-28
Plugin/extension support


amacneil:
This is a placeholder ticket. We want to support several types of extensions. At a minimum:

- Panel extensions
- Data provider extensions
- 3d panel extensions (render new things, similar to what is possible in Node Playground today)

https://code.visualstudio.com/api is a good reference for approaches

amacneil:
We now support custom panel extensions. We can create separate tickets to track data source extensions.

PR #886, opened 2021-05-06
Initial structure for creating and executing extensions

jhurliman:
Implements #774.

This is the initial phase of creating and packaging extensions independent of the `@foxglove/studio` codebase and dynamically loading them at runtime with the FGS application. The initial use case will be to load custom panels, starting by moving the Image panel into an extension, although for now this PR just contains the scaffolding for extension packaging and loading and nothing to do with UI or panels yet.

Core ideas:
* Extensions follow the convention of an npm package without actually publishing to the npm registry. The project skeleton is intended to be created by a code generator such as `yo fox myextension`
* Extensions import a single types-only package (or a package with minimal code plus types) called `@foxglove/studio` and hook into FGS by exporting particular named functions following a specific convention. This mirrors the practice of writing a VSCode extension
* Extensions are written in TypeScript and compiled with the provided webpack config. Other workflows are considered "at your own risk" while we continue to flesh out the API and decide between ESM + dynamic import, CommonJS + fetch + eval, etc
* Extensions are packaged into `.foxe` files similar to `.vsix` files in that they are just renamed zip files, but with 100% less XML (relying purely on the `package.json` files as the manifest)
* ~FGS should have a file handler for `.foxe` which unpacks the extension into `$HOME/.foxglove-studio/extensions/`~
* The `$HOME/.foxglove-studio/extensions/` directory is walked on app start to find all valid extensions. These are stored as a list of URIs in the form `x-foxglove-extension:<extensiondir>`
* Extensions are dynamically loaded into the renderer using dynamic imports: `import("x-foxglove-extension:foxglove.panel-image-0.0.1")` (this part is not implemented yet, hacked around by adding `file:` to CSP)
* Extensions expose various methods such as `activate(context)` that are called by FGS. The `activate` handler can in turn call `context.registerPanel(...)` to register one or more custom panels. This closely follows the convention of VSCode extensions


app/Extensions.ts
```diff
@@ -0,0 +1,38 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Logger from "@foxglove/log";
+import type { ExtensionContext } from "@foxglove/studio";
+
+import { Extension } from "./Extension";
+import OsContextSingleton from "./OsContextSingleton";
```
defunctzombie, line 9: Please no additional OsContextSingletons - if we need some behavior from the platform we should have a context that provides it.

app/Extensions.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export class Extensions {
+  loaded: Extension[] = [];
+
+  async load(): Promise<void> {
+    const uris = (await OsContextSingleton?.getExtensionUris()) ?? [];
+    for (const uri of uris) {
+      log.debug(`Importing extension from ${uri}`);
+      try {
+        const extension = (await import(/* webpackIgnore: true */ uri)) as Extension;
```
defunctzombie, line 21: One thing I've been wanting to move away from is loading all panels on startup - is this extension loading suppose to be the "cheap" import and the activate call the more "expensive" one? It would be nice to think of a way to not load lots of "js" for every possible extension until it is used.

app/context/ExtensionsProvider.tsx
```diff
@@ -0,0 +1,20 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
defunctzombie, line 1: The provider should probably be in components? I think thats were we've been putting providers recently. @jtbandes

packages/@foxglove/fox/src/fox
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env node
```
defunctzombie, line 1: should go in a `bin` folder - I don't like it when these are in the src folder

packages/@foxglove/panel-image/package.json
```diff
@@ -0,0 +1,31 @@
...
+  "description": "",
+  "version": "0.0.1",
+  "engines": {
+    "@foxglove/studio": "^0.7.0"
+  },
+  "main": "./dist/extension.js",
+  "scripts": {
+    "foxglove:prepublish": "yarn run package",
+    "compile": "webpack",
+    "watch": "webpack --watch",
```
defunctzombie, line 13: We should provide some scripts for these in our fox package (kinda like the react-scripts package) - so the user can just do `watch: fox watch` or something like that. Same with `package` - probably the direction you are going based on conversations - so this is more of a note on other packages I've seen have that behavior.

packages/@foxglove/panel-image/tsconfig.json
```diff
@@ -0,0 +1,17 @@
+{
```
defunctzombie, line 1: We could encourage the use of `@foxglove/tsconfig/tsconfig.recommended.json` or ship a recommended in the same package as the extension api type declarations. It would be nice to have a set of compiler flags on that we think help you write good typescript.

packages/@foxglove/panel-image/webpack.config.js
```diff
@@ -0,0 +1,47 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
defunctzombie, line 1: For basic extensions the user won't need a webpack config file? we will provide all this within `fox` ?

defunctzombie:
> * FGS should have a file handler for `.fgsx` which unpacks the extension into `$HOME/.foxglove-studio/extensions/`

This mirrors vscode for extensions. What is confusing to me is why the settings.json file isn't in a similar location for vscode - they keep that in the app data directory. I think we could keep extensions in the app data directory as well? Or maybe the correct move is to put settings in this .foxglove-studio directory as well. What goes in app data vs this new directory under home?

app/Extensions.ts
defunctzombie: I'm trying to kill them off to make a cleaner boundary for importing app into desktop (and later web): https://github.com/foxglove/studio/pull/840

Some "app" level things might remain - but I want to avoid as many as possible and leverage contexts for anything that needs to flow data into the app.

app/Extensions.ts
jhurliman: My idea is to have three levels of loading:

1. Scan the extensions directory and parse all `package.json` files. This is the "cheap" load that happens at startup.
2. This `load()` call where we dynamically import all extensions. Extension authors are encouraged not to do any work until the activate call so this should be relatively cheap (primarily the cost of loading and parsing JS), but ultimately we are parsing and executing third party code here so we can't make assumptions about load times. It should happen asynchronously while the rest of the app is user-interactable.
3. Calling `activate()` for all enabled (installed and not disabled) extensions. Activation will take an arbitrary amount of time so we should expect panel registrations, data source connector registrations, etc to trickle in while the user is interacting with the app. Ideally, this all happens before the user has a chance to start clicking around in the app but we can't make that guarantee. We need to consider the UX of the app starting up and loading a layout that contains extension-based panels, and what to show for those panels while extensions are still loading. And what happens when a layout references an extension that is missing or disabled.

packages/@foxglove/fox/src/fox
jhurliman: This is following the `vsce` project layout but I don't think there's any harm in deviating from theirs.

packages/@foxglove/panel-image/package.json
jhurliman: `fox package` and `fox install` are implemented now and used by `panel-image`. For these other scripts like compile or watch, `vsce` purposely omits them to avoid tooling lock-in. If we implement `fox watch`, extension authors will most likely be locked into jest and webpack tooling.

`react-scripts` is a slightly different ideology. They want to provide a fully managed tooling environment until you eject and then you're entirely on your own. `vsce` wants to provide a template project plus the minimal amount of custom tooling required to create and publish extensions. I'm following the `vsce` thinking so we have fewer lines of code to maintain and feature requests to field.

packages/@foxglove/panel-image/webpack.config.js
jhurliman: See my comment above about tooling ideology. I think leaving the compiler toolchain out of `vsce` was a conscious decision and one that we should mirror. The more toolchain we bake into `fox`, the more risk we have of never being able to change it without breaking all existing extensions and the more we're on the hook for supporting every team's unique snowflake CI and preferred workflow.

jhurliman:
> This mirrors vscode for extensions. What is confusing to me is why the settings.json file isn't in a similar location for vscode - they keep that in the app data directory. I think we could keep extensions in the app data directory as well? Or maybe the correct move is to put settings in this .foxglove-studio directory as well. What goes in app data vs this new directory under home?

They do put settings there, sort of. You'll notice an `argv.json` file which contains settings that are parsed by VSCode very early on during startup and set things like your crash ID (the equivalent of our deviceId or userId). There is a distinction between per-device configuration and per-user configuration. The app data folder is often backed up to the cloud so if you login with the same account on two computers you might see app data synchronized between them. That's why `"appData"` for electron on Windows points to a directory like `C:\Users\user\AppData\Roaming`. VSCode chooses to treat extensions as per-device instead of per-user, everything in `settings.json` as per-user, and everything in `argv.json` as per-device. I think there is an option to store a list of installed extensions per-user, but you still have to install them manually per-device (placing them in $HOME instead of your roaming folder).

app/context/ExtensionsProvider.tsx
jhurliman: I see a single provider in the `desktop/renderer/components` folder. It seems like this should stay in `app` instead of `desktop` because you could load extensions in a web context too, without the Electron desktop app.

packages/@foxglove/fox/src/package.ts
```diff
@@ -0,0 +1,267 @@
...
+  zip.file<"stream">(filename, createReadStream(filename), { createFolders: true });
+}
+
+function getPackageDirname(pkg: PackageManifest): string {
+  const pkgName = parsePackageName(pkg.name);
+  return `${pkg.namespaceOrPublisher}.${pkgName.name}-${pkg.version}`;
+}
+
+function parsePackageName(name: string): { namespace?: string; name: string } {
+  const res = /^@([^/]+)\/(.+)/.exec(name);
```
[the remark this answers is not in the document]
jtbandes: Unit tests please!

packages/@foxglove/panel-image/README.md
```diff
@@ -0,0 +1,58 @@
+# panel-image README
+
+This is the README for your extension "panel-image". After writing up a brief description, we recommend including the following sections.
```
jtbandes, line 3: Is there a way we can credit the VS Code project for their inspiration/templates etc.?

packages/@foxglove/fox/src/package.ts
```diff
@@ -0,0 +1,267 @@
...
+
+  for (const file of baseFiles) {
+    if (!(await pathExists(file, FileType.File))) {
+      throw new Error(`Missing required file ${file}`);
+    }
+    files.add(file);
+  }
+
+  if (pkg.files != undefined) {
+    for (const relFile of pkg.files) {
```
jtbandes, line 145: Should we allow globs? `dist/*`

packages/@foxglove/fox/src/package.ts
```diff
@@ -0,0 +1,267 @@
...
+import { createReadStream, createWriteStream } from "fs";
+import { mkdir, readFile, readdir, stat } from "fs/promises";
+import JSZip from "jszip";
+import ncp from "ncp";
+import { homedir } from "os";
+import { join, normalize, relative, sep } from "path";
+import rimraf from "rimraf";
+import { promisify } from "util";
+
+import { info } from "./log";
```
jtbandes, line 15: please `log.info`, my feeble mind can't handle the number of 3-7 character all lowercase identifiers in this node ecosystem

packages/@foxglove/fox/src/package.ts
```diff
@@ -0,0 +1,267 @@
...
+    switch (fileType) {
+      case FileType.File:
+        return finfo.isFile();
+      case FileType.Directory:
+        return finfo.isDirectory();
+      case FileType.FileOrDirectory:
+        return finfo.isFile() || finfo.isDirectory();
+    }
+  } catch {
+    // ignore
```
jtbandes, line 221: would be nice to surface the error details here - it looks like all callers end up throwing immediately anyway

packages/@foxglove/fox/src/package.ts
```diff
@@ -0,0 +1,267 @@
...
+  }
+  return false;
+}
+
+async function isDirectory(pathname: string): Promise<boolean> {
+  return (await stat(pathname)).isDirectory();
+}
+
+function rmdir(dirname: string): Promise<void> {
+  return new Promise<void>((c, e) => rimraf(dirname, (err) => (err != undefined ? e(err) : c())));
```
jtbandes, line 231: Does this not work with `promisify`?

packages/@foxglove/panel-image/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "package": "webpack --mode production --devtool hidden-source-map",
+    "package-extension": "fox package",
+    "install-extension": "webpack && fox install",
```
jtbandes, line 16:
```suggestion
    "install-extension": "yarn compile && fox install",
```

packages/@foxglove/panel-image/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "test-compile": "tsc -p ./",
+    "test-watch": "tsc -watch -p ./",
```
jtbandes, line 18: What does "test" mean here? Is this for tests or is it testing that the compile works? If the former, why `./` rather than `./test`?

packages/@foxglove/panel-image/tsconfig.json
```diff
@@ -0,0 +1,17 @@
...
+    "outDir": "out",
+    "lib": ["es2020"],
+    "sourceMap": true,
+    "rootDir": "src",
+    "moduleResolution": "node",
+    "strict": true /* enable all strict type-checking options */
+    /* Additional Checks */
+    // "noImplicitReturns": true, /* Report error when not all code paths in function return a value. */
+    // "noFallthroughCasesInSwitch": true, /* Report errors for fallthrough cases in switch statement. */
+    // "noUnusedParameters": true,  /* Report errors on unused parameters. */
```
jtbandes, line 14: What's the plan for these?

packages/@foxglove/studio/index.d.ts
```diff
@@ -0,0 +1,24 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+declare module "@foxglove/studio" {
+  export enum ExtensionMode {
+    Production = 1,
+    Development = 2,
+    Test = 3,
+  }
```
jtbandes, line 10: Our codebase usually uses all uppercase for enum cases

packages/@foxglove/studio/index.d.ts
```diff
@@ -0,0 +1,24 @@
...
+
+  export interface ExtensionContext {
+    readonly extensionMode: ExtensionMode;
+  }
+
+  export interface Extension<T> {
+    readonly id: string;
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    readonly packageJSON: any;
```
jtbandes, line 20: `Record<string, unknown>`? (Who will actually be accessing this field?)

packages/@foxglove/studio/index.d.ts
```diff
@@ -0,0 +1,24 @@
...
+    Development = 2,
+    Test = 3,
+  }
+
+  export interface ExtensionContext {
+    readonly extensionMode: ExtensionMode;
+  }
+
+  export interface Extension<T> {
+    readonly id: string;
```
jtbandes, line 17: Where does this come from? I don't see it referenced anywhere else

desktop/preload/index.ts
```diff
@@ -4,8 +4,10 @@
 
 import { init as initSentry } from "@sentry/electron";
 import { contextBridge, ipcRenderer } from "electron";
+import { readdir, readFile } from "fs/promises";
 import { machineId } from "node-machine-id";
 import os from "os";
+import { dirname, join as pathJoin } from "path";
```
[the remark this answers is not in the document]
jhurliman: I thought wildcard imports were discouraged these days due to tree shaking?

packages/@foxglove/panel-image/package.json
jhurliman: This was pulled in from the VSCode extension boilerplate. If it's not clear I'll delete it.

app/Extensions.ts
defunctzombie: 👍 Do you think `init` or `initialize` would be better instead of `load`? something to convey - "don't do expensive things here" - maybe load does convey that - or we just need to have a nice "warning" banner in the docs around this - possibly even perf measure slow loading extensions and have a devtools warning for extension authors.

app/context/ExtensionsProvider.tsx
defunctzombie: 👍 My thinking with `desktop/renderer/components` was to explore the separation between web/desktop/app - keeping app (and its dependencies) focused on the "core" re-usable parts across these two environments. App would provide the ExtensionContext interface indicating the api app wants to use and each platform would provide a concrete provider implementation appropriate for their platform (LocalstorageExtensionProvider, CloudExtensionProvider, etc) whatever makes sense within the context of the platform.

packages/@foxglove/panel-image/tsconfig.json
defunctzombie: It does seem nice tho for roboticists to not worry about random files/settings for the "guided path". We could use major version numbers liberally here. We could also detect the user having their own settings and allow those. We can also work towards or away from any of these approaches as we get a feel for how they are used. `next.js` is somewhat inspirational to me here. While it does lock you into some tooling - it also creates a pleasant out-of-box experience for newcomers (and seasoned folks too). They don't need to get a degree in webpack or typescript configuration to succeed (as we know this is not easy) and as the project evolves they get the learnings from upgraded tooling happening for them.

app/Extension.ts
```diff
@@ -0,0 +1,9 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
defunctzombie, line 1: I'd like us to find a better place for this file. Generally not a fan of adding more to the top level _app_ directory. Is this file consumed externally? If not then maybe _services_ folder is not a bad place? I know our folder structure is maybe lacking for where to put such a file.

app/ExtensionInstance.ts
```diff
@@ -0,0 +1,28 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { Extension } from "./Extension";
+
+export class ExtensionInstance {
```
defunctzombie, line 7: Instance feels redundant here (or somehow wrong) - since the class is - well a class - the instances are the variables you instantiate. What is the intent you wanted to convey with this? There is the Extension.ts file which the ExtensionInstance does not extend.

app/Extensions.ts
```diff
@@ -0,0 +1,50 @@
...
+          (instance.extension as { packageJson: unknown }).packageJson = instance.packageJson;
+        } catch (err) {
+          log.error(`Failed to import extension ${uri}: ${err}`);
+        }
+      }
+    }
+  }
+
+  activate(): void {
+    const extensionMode =
```
defunctzombie, line 35: What is an extension mode?

app/Extensions.ts
```diff
@@ -0,0 +1,50 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
defunctzombie, line 1: Is Extensions specific to environment and replaced by the context now? If not this file definitely feels like it would live in _services_.

app/OsContext.ts
```diff
@@ -81,6 +81,9 @@ export interface OsContext {
   // Get an array of deep links provided on app launch
   getDeepLinks: () => string[];
 
+  // Get an array of available extension URIs and parsed package.json files
```
defunctzombie, line 84: Is this still necessary with the ExtensionsContext? Is this related to the desktop environment and how extensions are loaded for studio desktop?

One path I wanted to explore was to expose several items over the electron context bridge - each specific to a high level function - rather than one global.

app/OsContext.ts
```diff
@@ -81,6 +81,9 @@ export interface OsContext {
...
+  getExtensions: () => Promise<{ uri: string; packageJson: unknown }[]>;
```
defunctzombie, line 85: Why do the uris and packageJson files travel separately? Is that due to the annoyances/limitations around sending objects over the context boundary?

app/context/ExtensionsContext.ts
```diff
@@ -0,0 +1,19 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { createContext, useContext } from "react";
+
+import { Extensions } from "@foxglove-studio/app/Extensions";
```
defunctzombie, line 7: Reading through this further - Seems to me that the typical term for what is called `Extensions` here would be `ExtensionsStore` or maybe `ExtensionCatalog` ? Tho I guess in ORM/data models land if you had an extensions table the model would be called "Extensions". Maybe a `models` folder is in order for such concepts.

desktop/main/index.ts
```diff
@@ -242,7 +241,8 @@ app.on("ready", async () => {
   // See: https://www.electronjs.org/docs/tutorial/security
   const contentSecurityPolicy: Record<string, string> = {
     "default-src": "'self'",
-    "script-src": `'self' 'unsafe-inline' 'unsafe-eval'`,
+    // We should use x-foxglove-extension: instead of file: (see https://github.com/foxglove/studio/issues/895)
```
defunctzombie, line 244: Let's avoid this tech-debt and see what we can do here as our first pass. I don't think this debt is necessary to take on in this situation and I want to be especially sure we are happy with extension loading.

desktop/preload/fileUrl.ts
```diff
@@ -0,0 +1,26 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { resolve } from "path";
+
+// Converts a filesystem path such as "/home/user/Documents/myfile" or
```
defunctzombie, line 7: What does this power? Extension loading? Are users able to "import" an extension from their downloads folder and have studio copy it to the correct location?

desktop/preload/index.ts
```diff
@@ -137,6 +140,33 @@ const ctx: OsContext = {
...
+          const packageData = await readFile(packagePath, { encoding: "utf8" });
+          const packageJson = JSON.parse(packageData);
+          const entryPoint = packageJson.main;
+          if (typeof entryPoint === "string" && entryPoint.length > 0) {
+            const entryPointPath = pathJoin(dirname(packagePath), entryPoint);
+            const url = fileUrl(entryPointPath);
+            uris.push(url);
+          }
+        } catch {
+          // ignore
```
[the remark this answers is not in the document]
defunctzombie: 👍 to logging to console. At least if an extension fails to load we should be able to ask the user for some basic debugging information.

desktop/preload/index.ts
```diff
@@ -137,6 +140,33 @@ const ctx: OsContext = {
...
   },
 
+  getExtensions: async (): Promise<{ uri: string; packageJson: unknown }[]> => {
+    const extensions: { uri: string; packageJson: unknown }[] = [];
+
+    const homePath = (await ipcRenderer.invoke("getHomePath")) as string;
+    const rootFolder = pathJoin(homePath, ".foxglove-studio", "extensions");
+    const rootFolderContents = await readdir(rootFolder, { withFileTypes: true });
+    for (const entry of rootFolderContents) {
+      if (entry.isDirectory()) {
```
defunctzombie, line 150: nit: early continue (in a flipped `if` statement) will reduce the nesting for the bulk of the business logic.

packages/@foxglove/fox/README.md
```diff
@@ -0,0 +1,3 @@
+# fox
+
+> _The Foxglove Studio Extension Manager_
```
defunctzombie, line 3: In what sense is it the extension manager?

jtbandes: https://github.com/microsoft/vscode-vsce/blame/22a3e7fe4f216341247459ca982d090c3d8e6799/README.md#L3

packages/@foxglove/panel-image/src/extension.ts
```diff
@@ -0,0 +1,11 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import type * as studio from "@foxglove/studio";
+
+// this method is called when your extension is activated
+export function activate(_context: studio.ExtensionContext): void {
+  // eslint-disable-next-line no-restricted-syntax
+  console.log("panel-image activate() called");
```
defunctzombie, line 10: Lets expose `studio.log` or `studio.logger` ? I think it would be good to get folks in the habit of logging via our api for their extensions

packages/@foxglove/panel-image/webpack.config.js
defunctzombie: I remain torn on this (see my comment elsewhere referencing next.js) - something to talk over. I see having users write their own webpack as another point of possible errors/edge cases but I also appreciate not taking on such responsibility. Maybe a middle ground is support for an "eject" mode where we don't use our default webpack config and you are on your own with build tooling.

packages/@foxglove/studio/src/index.ts
```diff
@@ -0,0 +1,21 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+export enum ExtensionMode {
```
defunctzombie, line 5: I'm guessing we will add documentation to this file as we expose apis to the end user?

What is the extension mode? I saw it else where in the PR but didn't follow the impact.

packages/@foxglove/studio/src/index.ts
```diff
@@ -0,0 +1,21 @@
...
+  PRODUCTION = 1,
+  DEVELOPMENT = 2,
+  TEST = 3,
+}
+
+export interface ExtensionContext {
+  readonly extensionMode: ExtensionMode;
+}
+
+export interface Extension<T> {
```
defunctzombie, line 15: Why do we expose "extension" to the user? Isn't that an internal studio construct and not an API the end user consumes?

app/Extensions.ts
jhurliman: I think `load` is the most descriptive, it is the call that literally fetches JS from disk and loads it into V8. The docs should definitely mention avoiding any expensive work in your extension in the top-level context (outside of a callback like activate).

packages/@foxglove/panel-image/tsconfig.json
jhurliman: Let's discuss on the call today. I didn't have to learn anything about webpack or build tooling to create my VSCode hello world extension, the tsconfig and webpack config it generated worked fine without any tweaking.

app/Extensions.ts
jhurliman: https://code.visualstudio.com/api/references/vscode-api#ExtensionMode

app/OsContext.ts
jhurliman: `ExtensionsContext` is just another level of indirection. It can't actually do anything useful. At the end of the day, someone has to call the node.js function `readdir` from either a preload script or the main process.

app/OsContext.ts
jhurliman: I'm not sure what you are asking. They are traveling as a struct with two members here, maybe a misread of the type signature?

packages/@foxglove/studio/src/index.ts
jhurliman: It's used in VSCode extensions so an extension can rely on and interact with another extension. We can drop that ability if we want.

desktop/preload/fileUrl.ts
jhurliman: This will change with `x-foxglove-extension:`

packages/@foxglove/fox/src/package.ts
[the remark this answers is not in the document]
jtbandes: @jhurliman We still need tests for these helper functions

PR #887, opened 2021-05-06
Add back check for __webviz_enum, in addition to new __foxglove_enum

esthersweon:
- Renamed `__studio_enum` (added in https://github.com/foxglove/studio/pull/847) to `__foxglove_enum`
- Added back check for `__webviz_enum`
- Added unit tests

_Note_: Will be adding to online docs in separate PR.

### DIFF
--- a/app/util/selectors.ts
+++ b/app/util/selectors.ts
@@ -92,11 +92,9 @@ export const constantsByDatatype = createSelector<any, any, any, any>(
 // Studio enum annotations are of the form: "Foo__foxglove_enum" (notice double underscore)
 // This method returns type name from "Foo" or undefined name doesn't match this format
 export function extractTypeFromStudioEnumAnnotation(name: string): string | undefined {
-  const matches = [/(.*)__foxglove_enum$/.exec(name), /(.*)__webviz_enum$/.exec(name)];
-  for (const match of matches) {
-    if (match) {
-      return match[1];
-    }
+  const match = /(.*)__(foxglove|webviz)_enum$/.exec(name);
+  if (match) {
+    return match[1];
   }
   return undefined;
 }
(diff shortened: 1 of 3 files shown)


app/util/selectors.ts
```diff
@@ -92,9 +92,11 @@ export const constantsByDatatype = createSelector<any, any, any, any>(
 // Studio enum annotations are of the form: "Foo__studio_enum" (notice double underscore)
 // This method returns type name from "Foo" or undefined name doesn't match this format
 export function extractTypeFromStudioEnumAnnotation(name: string): string | undefined {
-  const match = /(.*)__studio_enum$/.exec(name);
-  if (match) {
-    return match[1];
+  const matches = [/(.*)__studio_enum$/.exec(name), /(.*)__webviz_enum$/.exec(name)];
```
amacneil, line 95: I don't claim to fully understand what this is used for, but I think it's used in ros message definitions?

If so, maybe we should call it `__foxglove_enum` rather than `__studio_enum` since the latter is pretty generic?

esthersweon: Yep, it's for adding "enums" to ROS datatypes. Will change to `foxglove`.

esthersweon: Just important to note that we use "studio" to name foxglove studio-specific concepts in the codebase

i.e. `/studio_node`, `/studio_source_2`

Did this because Foxglove will eventually mean other products outside Studio, and wanted to keep it clear.

amacneil: Yeah, makes sense. I think it's probably worth a larger discussion with the rest of team before we change it.

My initial thought is that `/studio_node` makes sense because you only see that _inside_ studio, whereas `__studio_enum` might be a bit weird appearing inside someone's unrelated c++ codebase.

jhurliman: Don't forget the name of our app when connected as a ROS node, `/foxglovestudio` 😂

amacneil (approved):

PR #897, opened 2021-05-08
Add fallback for new global state key in local storage

esthersweon:
- Add fallback check for `webvizGlobalState` when getting initial persisted state
- Remove deprecated key from local storage going forward

### DIFF
--- a/app/reducers/panels.ts
+++ b/app/reducers/panels.ts
@@ -84,6 +84,7 @@ import {
 
 const storage = new Storage();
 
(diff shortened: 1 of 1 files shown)


defunctzombie:
Is this what caused the "unknown" layouts to proliferate? Or was this a bystander of the "remove all webviz" references change?

jtbandes:
The latter. I don't think we have looked into unknown layouts - will file a ticket for this

PR #899, opened 2021-05-10
Reduce width and min size of sidebar

amacneil:
Side note: We should find a way to set the sidebar width in pixels rather than percentage (e.g. if I increase the window size, I do not expect the sidebar to increase proportionally)

### DIFF
--- a/app/components/Sidebar/index.tsx
+++ b/app/components/Sidebar/index.tsx
@@ -75,7 +75,7 @@ export default function Sidebar<K extends string>({
           direction: "row",
           first: "sidebar",
           second: "children",
-          splitPercentage: 30,
+          splitPercentage: 23,
         });
       }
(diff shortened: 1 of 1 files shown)


app/components/Sidebar/index.tsx
```diff
@@ -75,7 +75,7 @@ export default function Sidebar<K extends string>({
           direction: "row",
           first: "sidebar",
           second: "children",
-          splitPercentage: 30,
+          splitPercentage: 23,
```
jtbandes, line 78: Oddly specific choice - maybe 25? Doesn't really matter though

jtbandes (approved):
Sadly I don't think react-mosaic provides support for pixel sizing. We may end up in a world where we need to make a new mosaic component to meet all our needs.

PR #903, opened 2021-05-10
Fix scrollbars hiding text on the error modal

jhurliman:
Before:

before

After:

after


app/components/PlayerStatusIndicator/PlayerProblemModal.tsx
```diff
@@ -29,7 +29,7 @@ const styles = mergeStyleSets({
   },
   error: {
     fontFamily: mixins.monospaceFont,
-    overflow: "scroll",
+    whiteSpace: "normal",
```
jtbandes, line 32: Does this disable text wrapping at newlines? What about just changing `overflow` to `auto`?

PR #906, opened 2021-05-10
Remove tabs from Preferences sidebar

jtbandes:
Now that preferences are in a sidebar, the tabbed Pivot UI doesn't really fit. Just stack all the items instead, since we don't have many settings.

image


amacneil:
lgtm

PR #925, opened 2021-05-12
ROS1 publishing support

jhurliman:
This adds publishing support to the `@foxglove/ros1` package. The primary new methods are `RosNode#advertise` and `RosNode#publish`. Advertised topics are described by topic, dataType, and any combination of messageDefinition, messageDefinitionText, and md5sum. This allows topics where we have the ROS msgdef to be republished, or new message types can by synthesized at runtime.

## TODO

 - [x] rosmaster server implementation
 - [x] integration testing using internal rosmaster server


defunctzombie:
https://github.com/foxglove/studio/pull/981

licenses.test.ts
```diff
@@ -28,27 +28,34 @@ jest.setTimeout(30 * 1000);
...
-            return;
-          }
-          resolve(modules);
-        },
-      );
-    });
-
-    // eslint-disable-next-line
-    console.log((checker as any).asSummary(packages));
+    // await printLicenses();
```
jtbandes, line 31: what's happening with this?

licenses.test.ts
```diff
@@ -28,27 +28,34 @@ jest.setTimeout(30 * 1000);
...
+    await checkLicenses(ALLOWED_LICENSES.join(";"));
   });
 });
+
+// async function printLicenses() {
```
defunctzombie, line 36: Dead code?

licenses.test.ts
jhurliman: Nothing, it's commented out

packages/ros1/src/Publication.ts
```diff
@@ -11,44 +15,95 @@ type SubscriberStats = [number, number, number, number, 0];
 // e.g. [2, "/listener", "o", "TCPROS", "/chatter", true, "TCPROS connection on port 55878 to [127.0.0.1:44273 on socket 7]"]
 type SubscriberInfo = [number, string, "o", string, string, number, string];
 
-export class Publication {
+export class Publication extends EventEmitter {
```
jtbandes, line 18: Is EventEmitter interface used? Doesn't look like it has been used in the class body – if it is used, please specify a type map here so the .on() methods are properly typed.

packages/ros1/src/RosMaster.ts
```diff
@@ -0,0 +1,507 @@
...
+  for (let i = 0; i < args.length; i++) {
+    if (expected[i] !== "*" && typeof args[i] !== expected[i]) {
+      return new Error(`Expected "${expected[i]}" for arg ${i}, got "${typeof args[i]}"`);
+    }
+  }
+
+  return undefined;
+}
+
+export class RosMaster extends EventEmitter {
```
jtbandes, line 28: Comment - what is this used for? How is it used?

packages/ros1/src/RosMaster.ts
```diff
@@ -0,0 +1,507 @@
...
+    this._server.close();
+  }
+
+  url(): string | undefined {
+    return this._url;
+  }
+
+  // <http://wiki.ros.org/ROS/Master_API> handlers
+
+  registerService = (_: string, args: XmlRpcValue[]): Promise<RosXmlRpcResponse> => {
```
jtbandes, line 84: what are these dangling string params? Can aid readability if the name is specified, so we know what is being ignored

packages/ros1-turtlesim-test/src/turtlesim.ts
```diff
@@ -0,0 +1,71 @@
...
+
+import { RosNode } from "@foxglove/ros1";
+import {
+  getEnvVar,
+  getHostname,
+  getNetworkInterfaces,
+  getPid,
+  TcpServerNode,
+  TcpSocketNode,
+} from "@foxglove/ros1/src/nodejs";
```
defunctzombie, line 13: I think it would be nice for the node interfaces to be separate package so users are not importing from "src" folder which is a bit of an anti-pattern? Can be follow-up PR since I know thats the status quo going into this PR.

packages/ros1/src/TcpPublisher.ts
```diff
@@ -0,0 +1,154 @@
...
+      topic: string,
+      connectionId: number,
+      destinationCallerId: string,
+      client: Client,
+    ) => void,
+  ): this;
+  on(event: string, listener: ListenerFn): this;
+}
+
+// Implements a subscriber for the TCPROS transport. The actual TCP transport is
```
jtbandes, line 27: publisher?

packages/ros1/src/nodejs/roscore.ts
```diff
@@ -0,0 +1,22 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
defunctzombie, line 1: Should this live in a bin folder? or maybe a separate ros1-core package? I like the idea of it being clear what are binaries and what are libraries.

packages/ros1/src/TcpPublisher.ts
```diff
@@ -0,0 +1,154 @@
...
+import { EventEmitter, ListenerFn } from "eventemitter3";
+
+import { Client } from "./Client";
+import { LoggerService } from "./LoggerService";
+import { Publication } from "./Publication";
+import { Publisher } from "./Publisher";
+import { PublicationLookup, TcpClient } from "./TcpClient";
+import { TcpAddress, TcpServer, TcpSocket } from "./TcpTypes";
+
+export declare interface TcpPublisher {
```
jtbandes, line 14: This should be done using the EventEmitter generic param instead of a separate interface.

packages/ros1/src/rosmsg.ts
```diff
@@ -0,0 +1,32 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { RosMsgDefinition } from "rosbag";
+
+export function rosMsgDefinitionText(msgDefs: RosMsgDefinition[]): string {
+  let output = "";
```
defunctzombie, line 8: random alternative suggestion: build up an array of strings and then `.join('\n')` at the end (can ignore if you want or do in another PR)

packages/ros1/src/mock/MockHttpServer.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  private _listening = false;
+  private _port?: number;
+  private _hostname?: string;
+
+  url(): string | undefined {
+    if (!this._listening) {
+      return undefined;
+    }
+    return `http://${this._hostname ?? "127.0.0.1"}:${this._port ?? 11311}/`;
```
jtbandes, line 18: seems like these defaults (at least the port) don't belong in a class generically called MockHttpServer – can we just require them to be passed in at construction?

packages/ros1/src/rosmsg.ts
```diff
@@ -0,0 +1,32 @@
...
+  for (let i = 0; i < msgDefs.length; i++) {
+    const msgDef = msgDefs[i] as RosMsgDefinition;
+    const constants = msgDef.definitions.filter((d) => d.isConstant);
```
defunctzombie, line 11: please avoid single letter vars - even in these cases - I'm a fan of words :D

packages/xmlrpc/src/XmlRpcServer.ts
```diff
@@ -8,7 +8,11 @@ import { Deserializer } from "./Deserializer";
 import { HttpRequest, HttpResponse, HttpServer } from "./HttpTypes";
 import { serializeFault, serializeMethodResponse } from "./Serializer";
 import { XmlRpcFault } from "./XmlRpcFault";
-import { XmlRpcMethodHandler } from "./XmlRpcTypes";
+import { XmlRpcMethodHandler, XmlRpcValue } from "./XmlRpcTypes";
+
+const APPLICATION_ERROR = -32500;
```
defunctzombie, line 13: Maybe an errors enum?

packages/ros1/src/RosNode.test.ts
```diff
@@ -0,0 +1,90 @@
...
+    const args = await paramUpdated;
+    expect(args.callerId).toEqual("/nodeA");
+    expect(args.key).toEqual("a");
+    expect(args.value).toEqual(42);
+    expect(args.prevValue).toBeUndefined();
+
+    nodeB.shutdown();
+    nodeA.shutdown();
+    await new Promise((r) => setTimeout(r, 100));
+    rosMaster.close();
```
jtbandes, line 88: should the close function be async so we can actually wait for teardown?

packages/xmlrpc/src/XmlRpcServer.ts
```diff
@@ -41,26 +45,63 @@ export class XmlRpcServer {
...
+    if (methodName === "system.multicall") {
+      if (!Array.isArray(args) || args.length !== 1 || !Array.isArray(args[0])) {
+        body = serializeFault(new XmlRpcFault("Invalid system.multicall", INVALID_PARAMS_ERROR));
+      } else {
+        const calls = args[0] as { methodName: string; params: XmlRpcValue[] }[];
+        const responses = await Promise.all(
+          calls.map((c) => this._methodCallHandler(c.methodName, c.params)),
         );
+        const res: XmlRpcValue[] = responses.map((r) => {
+          if (r instanceof XmlRpcFault) {
```
defunctzombie, line 67: Please avoid single letter var names - I have a preference for words when reading the code /shrug

packages/ros1/src/nodejs/roscore.ts
```diff
@@ -0,0 +1,22 @@
...
+import { RosMaster } from "../RosMaster";
+import { RosNode } from "../RosNode";
+
+async function main() {
+  const hostname = RosNode.GetRosHostname(getEnvVar, getHostname, getNetworkInterfaces);
+  const port = 11311;
+  const httpServer = new HttpServerNodejs();
+  const rosMaster = new RosMaster(httpServer);
+
+  await rosMaster.start(hostname, port);
```
jtbandes, line 17: should we gracefully close this in an atexit handler?

packages/ros1/src/TcpPublisher.ts
```diff
@@ -0,0 +1,154 @@
...
+    if (!this._shutdown) {
+      this._log?.warn?.(`tcp publisher error: ${err}`);
+    }
+  };
+}
+
+async function socketRemoteAddress(socket: TcpSocket): Promise<TcpAddress | undefined> {
+  try {
+    return await socket.remoteAddress();
+  } catch {
```
defunctzombie, line 151: Error thrown away? this could make debugging hard when it fails - why is the error discarded here? I'm tempted to say it should not be.

packages/ros1/src/rosmsg.ts
```diff
@@ -0,0 +1,32 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { RosMsgDefinition } from "rosbag";
+
+export function rosMsgDefinitionText(msgDefs: RosMsgDefinition[]): string {
```
jtbandes, line 7: Needs a comment. Also needs unit tests.

packages/ros1/src/TcpPublisher.ts
```diff
@@ -0,0 +1,154 @@
...
+export class TcpPublisher extends EventEmitter implements Publisher {
+  private _server: TcpServer;
+  private _nodeName: string;
+  private _getConnectionId: () => number;
+  private _getPublication: PublicationLookup;
+  private _pendingClients = new Map<number, TcpClient>();
+  private _shutdown = false;
+  private _log?: LoggerService;
+
+  constructor(
```
defunctzombie, line 42: Please use an options object - this is too many args to pass individually - makes call site reading hard(er).

packages/xmlrpc/src/XmlRpcServer.ts
jtbandes, line 13: Is there a link we can reference for where these standard values come from?

packages/ros1/src/TcpPublisher.ts
defunctzombie, line 27: I think we are using /** */ style comments so vscode can provide the documentation hints.

licenses.test.ts
jtbandes: Ok, please delete it then.

packages/ros1/src/rosmsg.ts
jtbandes: the zero-letter variable name: `filter(({ isConstant }) => isConstant)`

packages/ros1/src/TcpConnection.ts
```diff
@@ -144,7 +144,8 @@ export class TcpConnection extends EventEmitter implements Connection {
   }
 
   toString(): string {
-    return `tcpros://${this._address}:${this._port}`;
+    const host = this._address.includes(":") ? `[${this._address}]` : this._address;
```
defunctzombie, line 147: This could use a comment - I'm not following why this syntax is important in this case - I'm guessing something to do with tcpros:// assumptions but it would be nice to document.

packages/ros1/src/RosNode.ts
```diff
@@ -383,9 +462,13 @@ export class RosNode extends EventEmitter {
     return { port: protocol[2] as number, address: protocol[1] as string };
   }
 
-  private _newConnectionId(): number {
+  private _newConnectionId = (): number => {
```
defunctzombie, line 465: why did these change to property assignments rather than class functions? Does this make a new function on every instance of the class rather than making the function once? Or does it make no difference to the number of functions created?

packages/ros1/src/RosNode.ts
```diff
@@ -429,6 +566,34 @@ export class RosNode extends EventEmitter {
...
+    const callerApi = this._callerApi();
+
+    const [status, msg, subscribers] = await this.rosMasterClient.registerPublisher(
+      this.name,
+      publication.name,
+      publication.dataType,
+      callerApi,
+    );
+
+    if (status !== 1) {
```
defunctzombie, line 583: magic number `1` could use an enum home

packages/ros1/src/RosNode.ts
```diff
@@ -429,6 +566,34 @@ export class RosNode extends EventEmitter {
...
+      throw new Error(`registerPublisher() failed. status=${status}, msg="${msg}"`);
+    }
+
+    this._log?.debug?.(`registered publisher for ${publication.name} (${publication.dataType})`);
+    if (!Array.isArray(subscribers)) {
+      throw new Error(
```
defunctzombie, line 589: RosNode is an event emitter - does it make sense for this throw to instead emit an error? Then the creator of RosNode can handle "errors" it emits? Or is this call (_registerPublisher) the direct result of a user invoked function rather than background read/write?

packages/ros1/src/Subscription.ts
```diff
@@ -47,13 +47,15 @@ export class Subscription extends EventEmitter {
   readonly name: string;
   readonly md5sum: string;
   readonly dataType: string;
+  readonly tcpNoDelay: boolean;
   private _publishers = new Map<number, PublisherLink>();
 
-  constructor(name: string, md5sum: string, dataType: string) {
+  constructor(name: string, md5sum: string, dataType: string, tcpNoDelay: boolean) {
```
defunctzombie, line 53: Too many args - should be an options object now.

packages/ros1/src/TcpClient.ts
```diff
@@ -0,0 +1,220 @@
...
+  private _nodeName: string;
+  private _connected = true;
+  private _receivedHeader = false;
+  private _transportInfo: string;
+  private _stats: ClientStats = { bytesSent: 0, bytesReceived: 0, messagesSent: 0 };
+  private _getPublication: PublicationLookup;
+  private _log?: LoggerService;
+  private _transformer: RosTcpMessageStream;
+
+  constructor(
```
defunctzombie, line 35: Too many args - should be an options object now.

packages/ros1/src/TcpClient.ts
```diff
@@ -0,0 +1,220 @@
...
+    socket: TcpSocket,
+    address: string,
+    port: number,
+    nodeName: string,
+    getPublication: PublicationLookup,
```
defunctzombie, line 40: This option feels weird to me - could the caller announce they are publishing things be invoking a function on the TcpClient rather than having the client call out to functions passed in? There are times when I think this works - but here it feels like there is a more natural alternative.

packages/ros1/src/nodejs/roscore.ts
jhurliman: This isn't a real binary. The thing living in a bin folder would look like this:

```
#!/usr/bin/env node
require('./roscore.js')
```

defunctzombie: Its also not a library tho (doesn't export anything) - I dunno - just feels to me like module organization could be clearer here - or just make a ros1-core package for now that has the binary. I think its absolutely ok to separate out library from specific binaries you might run. Maybe this gets an arg parser later - now your ros1 package depends on an arg parser when no real reason for that. These are good examples for readers.

packages/ros1/src/RosNode.test.ts
jhurliman: That would be nice, but it would require rewriting a decent amount of code and should be a separate ticket+PR.

packages/ros1/src/nodejs/roscore.ts
jhurliman: What would the graceful closing do? Killing the process will already close any listening sockets. It could broadcast out empty publisher lists to all subscribers but I think that deviates from the `ros_comm` implementation.

packages/ros1/src/TcpPublisher.ts
jhurliman: This bug is being fixed https://github.com/microsoft/tsdoc/issues/160

packages/ros1/src/TcpConnection.ts
jhurliman: Adding a comment. This is required by RFC2732 for printing IPv6 addresses.

packages/ros1/src/TcpPublisher.ts
defunctzombie: My last reading of that issue did not give me the sense it is being fixed. Thats an open discussion. Have any of the ts core devs chimed in on it and indicated either way?

packages/ros1/src/RosNode.ts
jhurliman: It was required to make `getConnectionId: this._newConnectionId,` above work. I assume this makes a new function on every instance of the class unless there is some V8 optimization under the hood.

packages/ros1/src/RosNode.ts
jhurliman: This is only called directly from the user-invoked `advertise()` method so we want it to throw.

packages/ros1/src/RosNode.ts
defunctzombie: Oh I see - cause it is bound - I dunno - feels fragile - getConnectionId should maybe do ` = () => this._newConnectionId` so it is clear what is going on.

packages/ros1/src/TcpClient.ts
jhurliman: I'm not following the suggestion. `const pub = this._getPublication(topic);` needs to be called when a certain message is parsed on the TCP socket handler, or in some other place for a different transport. It's not something the caller controls the timing of. Maybe a code example in a separate PR would help.

packages/ros1/src/nodejs/roscore.ts
jhurliman: The TypeScript creators disagree with you that this is not a library. See:

https://github.com/microsoft/TypeScript/blob/master/lib/tsc.js
https://github.com/microsoft/TypeScript/blob/master/bin/tsc

That repo is also an example of disagreeing with the "no mixing binaries and libraries in one package" rule. Since there is other cleanup that needs to happen with package layout here like removing `/src/` from import statements I imagine we will follow up with one or more reorg PRs where this can be addressed.

packages/ros1/src/TcpPublisher.ts
jhurliman: My reading of that and the linked proposal PR suggested it was moving forward, but maybe that's overly optimistic. This style of comment is used extensively throughout our packages, I'll look at options for automating a conversion to `/** */`

PR #926, opened 2021-05-12
Fix a few `@typescript-eslint/no-explicit-any`

jtbandes:


desktop/main/StudioWindow.ts
```diff
@@ -282,6 +282,8 @@ class StudioWindow {
       if (Menu.getApplicationMenu() === this._menu) {
         const existingMenu = Menu.getApplicationMenu();
         const fileMenu = existingMenu?.getMenuItemById("fileMenu");
+        // https://github.com/electron/electron/issues/8598
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
         (fileMenu?.submenu as any)?.clear();
```
jtbandes, line 287: cc @defunctzombie - comments are helpful for future readers to know why some code has been added that seemingly would call a method that doesn't exist

defunctzombie: 👍 looking at this I definitely had deja-vu thinking I commented on this strangeness but I cannot find where that comment is so maybe it only existed in my mind.

Issue #907, opened 2021-05-10
Drag preview is incorrectly positioned for Rosout panel


jtbandes:
Somehow this only seems to occur, and occurs consistently, for the Rosout panel and not other panels (at least not the few others I tried).

https://user-images.githubusercontent.com/14237/117725624-17047e00-b181-11eb-9937-2917377efbb8.mov

jtbandes:
Issue still seems to occur, though not always on the Log panel. Blocked on https://bugs.chromium.org/p/chromium/issues/detail?id=1208623.

defunctzombie:
@jtbandes Are you still able to reproduce? I tried today by dragging the log panel around but didn't notice the same behavior. If you are able to reproduce please indicate the steps taken.

jtbandes:
There's definitely still some bugginess around this. This is the "all panels" layout:

https://user-images.githubusercontent.com/14237/188714258-63b44594-1765-4164-baef-0f9abfbc408d.mov

jtbandes:
However it doesn't seem like we can do much here and the issue is minor, so I guess we can close it

jtbandes:
Encountered this again today.

https://user-images.githubusercontent.com/14237/188952296-c6143119-5dcc-493d-889e-c734c1f3b6a0.mov

PR #927, opened 2021-05-12
Repro case for incorrect positioning

jtbandes:
https://bugs.chromium.org/p/chromium/issues/detail?id=1208623

Researching #907. It appears a sizing pseudo-element added by react-virtualized is incorrectly positioned relative to its parent element.

image

image


defunctzombie:
@jtbandes do we still need this open? The chrome issue has a separate storybook showing the problem now.

jtbandes:
The storybook was generated from this PR. That said I'm not sure there is any need to keep it open.

PR #932, opened 2021-05-13
Check for mounted state before invoking callbacks in Chart component

defunctzombie:
Chart async operations may complete after the chart has mounted. It is
no longer safe to call prop handlers or other state updates once unmounted
so we guard these calls with an isMounted check.

This fixes instances of "cannot find chart with id" errors which occur
when our chart component has been unmounted but the rpc still returned a final
result.

### DIFF
--- a/app/components/Chart/index.tsx
+++ b/app/components/Chart/index.tsx
@@ -10,7 +10,8 @@
 
 import { ChartOptions, ChartData, ScatterDataPoint } from "chart.js";
 import Hammer from "hammerjs";
-import { useCallback, useEffect, useMemo, useRef, useState } from "react";
+import { useCallback, useEffect, useRef, useState } from "react";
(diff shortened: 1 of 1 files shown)


app/components/Chart/index.tsx
```diff
@@ -72,9 +73,10 @@ function rpcMouseEvent(event: React.MouseEvent<HTMLCanvasElement>) {
 
 // Chart component renders data using workers with chartjs offscreen canvas
 function Chart(props: Props): JSX.Element {
-  const [id] = useState(uuidv4());
+  const idRef = useRef(uuidv4());
```
jtbandes, line 76: Can we avoid generating a new uuid and discarding it on each render pass?

defunctzombie: Changed this to use the function state initialization pattern to generate a uuid only once.

PR #940, opened 2021-05-13
Move desktop specific OsContext items into desktop

defunctzombie:
Rather than defining a single global object in app, move
the renderer/preload specific elements of OsContext storage
and native menu support into desktop. In desktop, create the native
providers for the expected contexts used within app.


desktop/renderer/App.tsx
```diff
@@ -65,27 +63,14 @@ export default function App(): ReactElement {
...
-  // We detect the full screen state and adjust our rendering accordingly
-  // Note: this does not removed the handlers so should be done at the highest component level
-  const [isFullScreen, setFullScreen] = useState(false);
-  useEffect(() => {
-    OsContextSingleton?.addIpcEventListener("enter-full-screen", () => setFullScreen(true));
-    OsContextSingleton?.addIpcEventListener("leave-full-screen", () => setFullScreen(false));
-  }, []);
-
-  const insetToolbar = OsContextSingleton?.platform === "darwin" && !isFullScreen;
-  const windowGeometry = useMemo(() => ({ insetToolbar }), [insetToolbar]);
```
jtbandes, line 78: Would prefer to keep these until we actually remove the top bar in favor of native title bar on macOS.

defunctzombie: I've added a Titlebar component in the theme and updated Tooltip to use the theme value for titlebar height when calculating bounds. I've verified that tooltips avoid display in the titlebar and thus don't appear under the traffic light.

desktop/renderer/components/NativeStorageLayoutStorageProvider.tsx
```diff
@@ -4,21 +4,20 @@
...
 import { PropsWithChildren, useMemo } from "react";
 
-import OsContextSingleton from "@foxglove-studio/app/OsContextSingleton";
 import LayoutStorageContext from "@foxglove-studio/app/context/LayoutStorageContext";
-import OsContextLayoutStorage from "@foxglove-studio/app/services/OsContextLayoutStorage";
+
+import { useNativeStorage } from "../context/NativeStorageContext";
+import NativeStorageLayoutStorage from "../services/NativeStorageLayoutStorage";
 
 // Provide an instance of the OsContextLayoutStorage
```
jtbandes, line 12: comment out of date

defunctzombie: https://github.com/foxglove/studio/pull/950

PR #943, opened 2021-05-13
Set hue-shifted icon for development mode

jtbandes:
Helps distinguish from production builds that may also be running/installed.

image


amacneil:
🤷  sure

PR #944, opened 2021-05-13
Fix table panel

### LINKED ISSUE #931: Table panel not showing data that shows up correctly in other panels
image

defunctzombie:
Table panel uses Object.keys which is not supported by lazy messages
since they do not have enumerable own properties. Fix by using toJSON()
on the lazy message which returns a plain object.

Update lazy message cache properties to be private so they don't show
up in Object.keys and create confusion as to why these weird properties
are present.

Fixes #931


app/panels/Table/sanitizeAccessor.ts
```diff
@@ -0,0 +1,16 @@
...
+// This file incorporates work covered by the following copyright and
+// permission notice:
+//
+//   Copyright 2020-2021 Cruise LLC
+//
+//   This source code is licensed under the Apache License, Version 2.0,
+//   found at http://www.apache.org/licenses/LICENSE-2.0
+//   You may not use this file except in compliance with the License.
+
+export function sanitizeAccessorPath(accessorPath: string): string {
```
jtbandes, line 14: nit: function name should match filename

PR #946, opened 2021-05-14
Remove isCrashReportingEnabled and isTelemetryEnabled from OsContext

defunctzombie:
isTelemetryEnabled was unused. isCrashReporting is specific to preloader/render interaction so it is moved as a single value exposed by the preloader and read by renderer. No longer relevant to app.


defunctzombie:
Related to https://github.com/foxglove/studio/pull/840

PR #947, opened 2021-05-14
Remove handleToolbarDoubleClick and getDeepLinks from OsContext

defunctzombie:
These are desktop specific functions exposed by preloader. Provide
them to the workspace via properties on the Workspace component rather than via a global singleton.


defunctzombie:
Related: https://github.com/foxglove/studio/pull/829

PR #948, opened 2021-05-14
Fix AnalyticsProvider default settings

defunctzombie:
The defaults were accidentally flipped in https://github.com/foxglove/studio/pull/843. If the setting is not explicitly false, crash reporting and telemetry should be enabled.


amacneil:
Oh deep. That explains why our usage was down this week.

PR #949, opened 2021-05-14
Add web build

defunctzombie:
Screen Shot 2021-05-13 at 9 13 07 PM

TODO

  - [x] Add package.json in web for web specific dependencies (useQueryParams?)
  - [ ] ~~store settings (localstorage? query params?) (maybe some settings don't make sense for web?)~~
  - [ ] ~~store layouts (localstorage? query params?) (probably layouts in localstorage - with current layout in query params)~~
  - [x] Publish (automatically via ci? or only releases?)
  - [x] Move more components into shared files (those that are not environment specific)
  - [x] Consider whether toolbar should be added manually by the user or part of the workspace (I like the idea of no top bar in the browser)

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,8 @@
     "packages": [
       "app",
       "desktop",
-      "packages/**"
+      "packages/**",
+      "web"
     ]
   },
   "devDependencies": {
--- a/web/webpack.config.ts
+++ b/web/webpack.config.ts
@@ -1,0 +1,114 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import ReactRefreshPlugin from "@pmmmwh/react-refresh-webpack-plugin";
+import SentryWebpackPlugin from "@sentry/webpack-plugin";
+import { CleanWebpackPlugin } from "clean-webpack-plugin";
+import CopyPlugin from "copy-webpack-plugin";
+import HtmlWebpackPlugin from "html-webpack-plugin";
+import path from "path";
+import { Configuration, EnvironmentPlugin, WebpackPluginInstance } from "webpack";
+
+import type { WebpackArgv } from "@foxglove-studio/app/WebpackArgv";
+import { makeConfig } from "@foxglove-studio/app/webpack";
+
+export default (env: unknown, argv: WebpackArgv): Configuration => {
+  const isDev = argv.mode === "development";
+  const isServe = argv.env?.WEBPACK_SERVE ?? false;
+
+  const allowUnusedLocals = isDev && isServe;
+
+  const plugins: WebpackPluginInstance[] = [];
+
+  if (isServe) {
+    plugins.push(new ReactRefreshPlugin());
+  }
+
+  // Source map upload if configuration permits
+  if (
+    !isDev &&
+    process.env.SENTRY_AUTH_TOKEN != undefined &&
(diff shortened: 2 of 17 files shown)


package.json
```diff
@@ -22,7 +22,9 @@
     "build:packages": "tsc --build --verbose packages/**/tsconfig.json",
     "build:prod": "webpack --mode production --progress --config desktop/webpack.config.ts",
     "build:dev": "webpack --mode development --progress --config desktop/webpack.config.ts",
+    "build:web": "webpack --mode production --progress --config web/webpack.config.ts",
     "serve": "webpack serve --mode development --progress --config desktop/webpack.config.ts",
+    "serve:web": "webpack serve --mode development --progress --config web/webpack.config.ts",
```
amacneil, line 27: can we make serve just serve everything for simplicity?

web/webpack.config.ts
```diff
@@ -0,0 +1,107 @@
...
+  const config: Configuration = {
+    ...appWebpackConfig,
+
+    target: "web",
+    context: path.resolve(__dirname, "./src"),
+    entry: "./index.tsx",
+    devtool: isDev ? "eval-cheap-module-source-map" : "source-map",
+
+    devServer: {
+      contentBase: path.resolve(__dirname, ".webpack"),
```
amacneil, line 58: would be nice to have a directory like `.webpack/public` for serving from. or is everything under `.webpack` public?

package.json
defunctzombie: yes and no - We certainly could run two webpack commands but since these are completely different entry points (web vs desktop) we would be building everything twice, need to use separate ports for the server, and different urls. I do not think it is a good idea to do that. The scripts here are getting a little much and we could re-organize. I was thinking to move some of the packaging related ones to `desktop` under a new package.json there. Web and Desktop really are different environments. We could even move the electron builder config etc under desktop. Keep top level clean.

web/webpack.config.ts
defunctzombie: Everything under .webpack is public. We don't have to use `.webpack` that was just some convention from electron-forge (builder?) and I copied it here. We can use whatever folder we want.

package.json
amacneil: oh interesting. can we not have webpack build the common files into a folder alongside the web + desktop files (e.g. for each)?

web/webpack.config.ts
amacneil: `.webpack` is fine

package.json
defunctzombie: I don't think it is worth doing that - the entry points are completely different. During development there are no "common" files on disk because everything is served from memory.

web/src/App.tsx
```diff
@@ -0,0 +1,92 @@
...
+    <AnalyticsProvider />,
+    <PlayerManager playerSources={playerSources} />,
+    <AssetsProvider loaders={assetLoaders} />,
+    /* eslint-enable react/jsx-key */
+  ];
+
+  return (
+    <ErrorBoundary>
+      <MultiProvider providers={providers}>
+        <LayoutStorageReduxAdapter />
```
jtbandes, line 80: Should this be in Workspace since it is always needed?

web/src/App.tsx
```diff
@@ -0,0 +1,92 @@
...
+        <NativeFileMenuPlayerSelection />
+        <DndProvider backend={HTML5Backend}>
+          <Suspense fallback={<></>}>
+            <BuiltinPanelCatalogProvider>
```
jtbandes, line 84: Several of these things look like they should be common regardless of web or desktop target

PR #952, opened 2021-05-14
Improve contrast in fluent theme

amacneil:
Before:
Screen Shot 2021-05-14 at 11 15 58 AM

After:
Screen Shot 2021-05-14 at 11 49 06 AM

(not sure why connection picker icons don't follow the sidebar icon colors, but that seems like a separate issue)


jtbandes:
Screenshot

defunctzombie:
I am all for improved contrast. My eyes are not what they use to be.

amacneil:
> Screenshot
you're too fast 😜

jtbandes:
> not sure why connection picker icons don't follow the sidebar icon colors

It looks like your "dark"/"light" direction has swapped - was that intentional?

amacneil:
> It looks like your "dark"/"light" direction has swapped - was that intentional?

Not sure exactly what you mean, but after you made this comment I slightly fixed the theme colors by triggering onchange in the theme designer.

jtbandes:
I meant that `themeLighter` used to be bright and `themeDarker` used to be dark, and now they have switched. Maybe that is correct because we use an inverted theme - I'm not sure.

amacneil:
Well, our `white` is dark and our `black` is light, so I think that's intentional. At least that is what the theme builder generated for me.

PR #967, opened 2021-05-15
Workaround for `webpack --watch` / fork-ts-checker issues

### LINKED ISSUE #893: `yarn serve` doesn't watch all files
Steps to reproduce:
- Run `yarn serve` on main.
- Edit the `PanelConfigSchemaEntry` type in `app/types/panels.ts`, or edit the `RegisteredIconNames` type in `app/typings/fluentui.d.ts`.

Webpack does not even attempt to rebuild any files when these two files are changed.

It seems likely that this is a regression caused by #829 / #839.

jtbandes:
This workaround forces fork-ts-checker to watch file dependencies in our packages dirs. Previously it would not rebuild when certain types-only files were changed (https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/issues/611).

Fixes #893


package.json
```diff
@@ -51,7 +51,7 @@
     "@babel/preset-typescript": "7.13.0",
     "@foxglove/electron-socket": "workspace:packages/electron-socket",
     "@foxglove/tsconfig": "workspace:packages/@foxglove/tsconfig",
-    "@pmmmwh/react-refresh-webpack-plugin": "0.5.0-beta.8",
+    "@pmmmwh/react-refresh-webpack-plugin": "0.5.0-beta.1",
```
defunctzombie, line 54: Fixed by https://github.com/foxglove/studio/pull/966 which adds the newer react-refresh dep - we don't need to downgrade this.

packages/@foxglove/tsconfig/tsconfig.base.json
```diff
@@ -43,6 +43,15 @@
...
     "noUnusedLocals": true,
     "noUnusedParameters": true,
-    "forceConsistentCasingInFileNames": true
+    "forceConsistentCasingInFileNames": true,
+
+    // These paths are added to ensure that fork-ts-checker correctly typechecks changed files.
+    // Using these resolutions will force TS to give us an error if we forgot to add these files to
+    // one of our tsconfig "include"s.
+    // https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/issues/611
+    "paths": {
```
defunctzombie, line 52: Base is suppose to be just the compile settings - not related to "app" because it is used by packages that do not depend on app. I don't think we should put this here.

jtbandes: This doesn't really add a dependency, just sets up an alias that if a package _were_ to include these, they would be included from the right place.

The alternative is duplicating it in a few other tsconfigs - adds a bit more mess, but if that's desired it's no problem to change

desktop/main/tsconfig.json
```diff
@@ -1,8 +1,24 @@
...
     "rootDir": "../../",
-    "noEmit": true
+    "noEmit": true,
+
+    // These paths are added to ensure that fork-ts-checker correctly typechecks changed files.
+    // Using these resolutions will force TS to give us an error if we forgot to add these files to
+    // one of our tsconfig "include"s.
+    // https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/issues/611
+    "paths": {
+      "@foxglove-studio/app/*": ["../../app/*"],
```
defunctzombie, line 20: Outdated

desktop/renderer/tsconfig.json
```diff
@@ -1,10 +1,33 @@
...
     "jsx": "react-jsx",
-    "lib": ["dom", "dom.iterable", "es2020"]
+    "lib": ["dom", "dom.iterable", "es2020"],
+
+    // These paths are added to ensure that fork-ts-checker correctly typechecks changed files.
+    // Using these resolutions will force TS to give us an error if we forgot to add these files to
+    // one of our tsconfig "include"s.
+    // https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/issues/611
+    "paths": {
+      "@foxglove-studio/app/*": ["../../app/*"],
```
defunctzombie, line 29: outdated

desktop/renderer/tsconfig.json
```diff
@@ -1,10 +1,33 @@
...
   "extends": "@foxglove/tsconfig/tsconfig.base.json",
-  "include": ["./**/*", "../common/**/*", "../../package.json"],
+  "include": [
+    "./**/*",
+    "../common/**/*",
+    "../../package.json",
+
+    // These includes are added to ensure that fork-ts-checker correctly typechecks changed files:
+    // https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/issues/611
+    "../../app/**/*",
```
defunctzombie, line 10: We should wait on this until we move app into packages

defunctzombie:
@jtbandes still doing anything with this? Was it ever resolved upstream?

jtbandes:
I don't think it was resolved upstream – we still experience issues with webpack not updating when types are changed. We've all just become used to restarting the dev server occasionally. I'm not sure this approach is desired so I'll close the PR.

PR #971, opened 2021-05-15
Remove overwriteFetch

defunctzombie:
We do not have cors issues.


jtbandes (changes requested):
Don't remove this. The default error from `fetch()` is extremely bad and makes it hard to understand what has gone wrong. Change the message if you wish to remove the CORS suggestion.

PR #979, opened 2021-05-15
Replace rc-color-picker's ColorPicker with FluentUI's

wkalt:


wkalt:
@defunctzombie @jtbandes this isn't ready to go yet as I need to update the storybook tests. I'm looking for input though on how I can manually validate the changes in "SchemaEntryEditor.tsx" previously made by @jtbandes , and also how I can get the colorpicker code in "renderStyleExpressionNodes.tsx" to trigger, here: https://github.com/foxglove/studio/blob/main/app/panels/ThreeDimensionalViz/TopicTree/renderStyleExpressionNodes.tsx#L322,L346 . I would like to change SchemaEntryEditor to use the new picker component.

There are a couple other items in here -
* Move ColorPicker to a component
* Replace usage of PICKER_SIZE with inline constants
* I'll post screenshots of the new picker, but the color button is now a rounded rectangle instead of a circle.

Please comment if you've got strong thoughts on those, or any other input on the code (probably missing some javascript idioms).

app/components/ColorPicker.tsx
```diff
@@ -0,0 +1,113 @@
...
+  return `rgba(${rgba.r}, ${rgba.g}, ${rgba.b}, ${rgba.a ? rgba.a / 100 : 1})`;
+}
+
+type ColorPickerProps = {
+  color?: Color;
+  onChange: (newColor: Color) => void;
+};
+
+// Returns a FluentUI ColorPicker that accepts our regl-worldview Color type.
+export function FGColorPicker({ color, onChange }: ColorPickerProps): JSX.Element {
```
wkalt, line 72: probably just call this ColorPicker

wkalt:
Oh, one additional question - I don't understand what effect I'm supposed to see when changing the color on the TF topics in the 3D panel's topics dropdown. On main and on this patch, it doesn't seem to have any effect outside the picker in the bags I've been testing. Maybe the storybook tests will clarify.

app/components/ColorPicker.tsx
```diff
@@ -0,0 +1,113 @@
...
+    b: rgba.b / 255,
+    a: rgba.a ? rgba.a / 100 : 1,
+  };
+}
+
+// Returns an RGB string for a regl-worldview color. The scale of the formatted
+// tuple is (255, 255, 255, 1).
+export function getDefaultedRGBStringFromColor(color?: Color): string {
+  const rgba = getIRGBFromColor(color);
+  return `rgba(${rgba.r}, ${rgba.g}, ${rgba.b}, ${rgba.a ? rgba.a / 100 : 1})`;
```
wkalt, line 63: prefix with space

defunctzombie:
There are some rc-color-picker things in global.scss that could be removed with this change. https://github.com/foxglove/studio/issues/986

app/components/ColorPicker.tsx
```diff
@@ -0,0 +1,115 @@
...
+  return `rgba(${rgba.r}, ${rgba.g}, ${rgba.b}, ${alpha / 100})`;
+}
+
+type ColorPickerProps = {
+  color?: Color;
+  onChange: (newColor: Color) => void;
+};
+
+// Returns a FluentUI ColorPicker that accepts our regl-worldview Color type.
+export function ColorPicker({ color, onChange }: ColorPickerProps): JSX.Element {
```
defunctzombie, line 74: I find it better when component files export only one component. You can make a folder for the component and split these into separate files. Same goes for some of the helper functions.

app/theme/index.ts
```diff
@@ -31,6 +32,22 @@ export default createTheme({
     warningBackground: colors.YELLOW1,
   },
   components: {
+    ColorPicker: {
+      styles: {
+        root: { maxWidth: 250 },
+        colorRectangle: { minWidth: 100, minHeight: 100 },
+        table: {
+          // We need to remove table styles from global.scss, but for now, changing them
```
defunctzombie, line 40: Related https://github.com/foxglove/studio/issues/986

Maybe a quick interlude is to remove those table styles now @jtbandes ?

app/components/ColorPicker.tsx
wkalt: :+1: I put the two color picker functions in separate components under the components/ColorPicker directory, and moved the helper functions into a new utils.tsx file under the same place.

app/theme/index.ts
jtbandes: I think if this is tracked in #986, we don't need to do it before this PR - this is just moving from a different file

app/panels/ThreeDimensionalViz/TopicTree/renderStyleExpressionNodes.tsx
```diff
@@ -339,7 +336,7 @@ function ColorPickerOverlay({
           flexDirection: "column",
         }}
       >
-        <ColorPickerSettingsPanel color={color} onChange={onChangeColor} />
+        <ColorPicker color={color} onChange={onChangeColor} />
       </Modal>
     </RenderToBodyComponent>
```
jtbandes, line 341: If there's no good reason for this to be in a modal while the others are callouts, maybe we should just change it to look the same as the others?

app/components/ColorPicker/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+//   found at http://www.apache.org/licenses/LICENSE-2.0
+//   You may not use this file except in compliance with the License.
+
+import { ColorPicker as Picker } from "@fluentui/react";
+import { Color } from "regl-worldview";
+
+import {
+  getColorFromIRGB,
+  getIColorFromColor,
+} from "@foxglove-studio/app/components/ColorPicker/utils";
```
jtbandes, line 20: Looks like maybe this file hasn't been committed yet?

app/components/ColorPicker/ColorPickerButtonCallout.tsx
```diff
@@ -0,0 +1,54 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+//
+// This file incorporates work covered by the following copyright and
+// permission notice:
+//
+//   Copyright 2019-2021 Cruise LLC
+//
+//   This source code is licensed under the Apache License, Version 2.0,
+//   found at http://www.apache.org/licenses/LICENSE-2.0
+//   You may not use this file except in compliance with the License.
```
jtbandes, line 12: If this new file is entirely Foxglove code then we don't need to regain the Apache license header.

app/components/ColorPicker/ColorPickerButtonCallout.tsx
```diff
@@ -0,0 +1,54 @@
...
+import ColorPicker from "@foxglove-studio/app/components/ColorPicker/index";
+import { getIColorFromColor } from "@foxglove-studio/app/components/ColorPicker/utils";
+
+type Props = {
+  color?: Color;
+  onChange: (newColor: Color) => void;
+};
+
+// Returns a button that pops out an ColorPicker in a fluent callout.
+export default function ColorPickerButtonCallout({ color, onChange }: Props): JSX.Element {
```
jtbandes, line 27: Naming wise, it seems like this is the main component that other places will be using, so I wonder if it should be called ColorPicker? That's a bit of deviation from the fluent naming scheme (since this includes the callout) so understand if you'd rather keep it as is, but I think "..ButtonCallout" is a bit of a confusing name too.

To avoid needing separate components, I wonder if we could just add something like `triggerStyle: "button" | "circle"` to Props, and render either the DefaultButton or the circle in here, so none of the usage sites need to bother with rendering their own triggers?

wkalt:
reopened from a new branch: https://github.com/foxglove/studio/pull/1005

app/components/ColorPicker/ColorPickerButtonCallout.tsx
wkalt: Yeah, I like this idea. I was hung up on this Friday because I wasn't easily able to figure out how to trigger the code in renderStyleExpressionNodes to validate the change. I haven't checked whether we have any storybooks that do though, yet.

jtbandes: One common pattern you might find is `data-test="...."` (or `dataTest=...` in some reusable components) & `querySelector("[data-test='...']")` as an explicit way to expose things for testing.

PR #980, opened 2021-05-15
Delete QUERY_KEY global constants

defunctzombie:
Studio does not use query keys to manage state.


app/middleware/updateUrlAndLocalStorage.ts
```diff
@@ -95,12 +94,6 @@ const updateUrlAndLocalStorageMiddlewareDebounced =
...
       }
       updateUrlTimer = setTimeout(async () => {
-        const shouldProcessPatch = getShouldProcessPatch();
-        if (!shouldProcessPatch) {
-          maybeSetPersistedStateInLocalStorage(store, skipSettingLocalStorage);
-          return result;
-        }
-
         maybeSetPersistedStateInLocalStorage(store, skipSettingLocalStorage);
         return result;
```
jtbandes, line 98: This code block is comical - must've been a side effect of previous removals that it looks this way.

app/reducers/panels.ts
```diff
@@ -156,55 +147,15 @@ export function getInitialPersistedStateAndMaybeUpdateLocalStorageAndURL(
...
     const newPersistedState = cloneDeep(defaultPersistedState) as PersistedState;
 
-    const { search: currentSearch, pathname } = history.location;
-    const currentSearchParams = new URLSearchParams(currentSearch);
     const oldFetchedLayoutState = oldPersistedState?.fetchedLayout;
-    const oldPersistedSearch = oldPersistedState?.search;
     const fetchedLayoutDataFromLocalStorage = oldFetchedLayoutState?.data;
 
-    let isInitializedFromLocalStorage = false;
+    const isInitializedFromLocalStorage = false;
```
jtbandes, line 153: This should also have been removed - doesn't do anything now

PR #981, opened 2021-05-15
Add rosmsg package

defunctzombie:


packages/rosmsg/README.md
```diff
@@ -0,0 +1,23 @@
+# rosmsg-deser
```
amacneil, line 1: Package name in readme doesn't align with actual package name.

I assume this will support serialization as well as deserialization in the near future?

defunctzombie: This is a copy/paste error. I copied the content from our existing rosmsg-deser package.

Regarding whether this will support serialization - I need to sync with @jhurliman on his thoughts. Right now this package is just parsing message definitions (and in the near future writing message definition strings). We have rosmsg-deser that has the lazy message parsing but could roll a rosmsg-serialization that has serialization (writing) and deserialization) reading.

PR #983, opened 2021-05-15
Replace util/logEvent with our Analytics service

### LINKED ISSUE #982: Switch uses of utils/logEvent to our analytics context

defunctzombie:
Needs decision around how to handle non-react components with analytics. We can follow the existing global pattern but should put that into a package and indicate how it is injected better.

Fixes #982


app/components/MessagePipeline/index.test.tsx
```diff
@@ -481,15 +480,17 @@ describe("MessagePipelineProvider/useMessagePipeline", () => {
...
         logger,
         { PAUSE_FRAME_TIMEOUT: "pause_frame_timeout" },
         { PANEL_TYPES: "panel_types" },
       );
+      */
     });
 
     afterEach(() => {
-      resetLogEventForTests();
+      //resetLogEventForTests();
```
jhurliman, line 493: ?

defunctzombie: Should be deleted - none of this log event stuff is hooked up to anything

PR #984, opened 2021-05-15
Change desktop and web imports to top level @foxglove-studio/app

### LINKED ISSUE #975: Change `@foxglove-studio/app/...` imports to `@foxglove-studio/app`
Desktop and Web still import directly from filesystem paths within studio/app. They should import from the top level package exports (index.ts) only.

defunctzombie:
Rather than import from a specific path within @foxglove-studio/app,
web and desktop should import from the publicly exposed interfaces.

Fixes #975

### DIFF
--- a/app/index.ts
+++ b/app/index.ts
@@ -8,12 +8,30 @@
 // See typings/index.d.ts for additional included references
 /// <reference types="./typings" />
 
-import App from "@foxglove-studio/app/App";
-import ErrorBoundary from "@foxglove-studio/app/components/ErrorBoundary";
-import MultiProvider from "@foxglove-studio/app/components/MultiProvider";
-import { PlayerSourceDefinition } from "@foxglove-studio/app/context/PlayerSelectionContext";
-import ThemeProvider from "@foxglove-studio/app/theme/ThemeProvider";
+import App from "./App";
+import { NetworkInterface, OsContext } from "./OsContext";
+import ErrorBoundary from "./components/ErrorBoundary";
+import MultiProvider from "./components/MultiProvider";
+import { PlayerSourceDefinition } from "./context/PlayerSelectionContext";
+import ThemeProvider from "./theme/ThemeProvider";
+import installDevtoolsFormatters from "./util/installDevtoolsFormatters";
+import { initializeLogEvent } from "./util/logEvent";
+import overwriteFetch from "./util/overwriteFetch";
+import waitForFonts from "./util/waitForFonts";
+import version from "./version";
 
-export { App, ErrorBoundary, MultiProvider, ThemeProvider };
+const pkgInfo = version;
 
-export type { PlayerSourceDefinition };
+export {
+  pkgInfo,
+  App,
+  ErrorBoundary,
+  MultiProvider,
+  ThemeProvider,
+  installDevtoolsFormatters,
+  initializeLogEvent,
+  overwriteFetch,
+  waitForFonts,
+};
+
+export type { PlayerSourceDefinition, OsContext, NetworkInterface };
--- a/app/version.ts
+++ b/app/version.ts
@@ -7,3 +7,9 @@ import packageJson from "../package.json";
 export const APP_NAME: string = packageJson.productName;
 export const APP_VERSION: string = packageJson.version;
 export const APP_HOMEPAGE: string = packageJson.homepage;
+
+export default {
+  name: APP_NAME,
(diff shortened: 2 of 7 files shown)


app/version.ts
```diff
@@ -7,3 +7,9 @@ import packageJson from "../package.json";
 export const APP_NAME: string = packageJson.productName;
 export const APP_VERSION: string = packageJson.version;
 export const APP_HOMEPAGE: string = packageJson.homepage;
+
+export default {
+  name: APP_NAME,
+  homepage: APP_HOMEPAGE,
+  version: APP_VERSION,
```
amacneil, line 14: If this is preferred then remove named exports?

app/index.ts
```diff
@@ -8,12 +8,30 @@
...
+import { NetworkInterface, OsContext } from "./OsContext";
+import ErrorBoundary from "./components/ErrorBoundary";
+import MultiProvider from "./components/MultiProvider";
+import { PlayerSourceDefinition } from "./context/PlayerSelectionContext";
+import ThemeProvider from "./theme/ThemeProvider";
+import installDevtoolsFormatters from "./util/installDevtoolsFormatters";
+import { initializeLogEvent } from "./util/logEvent";
+import overwriteFetch from "./util/overwriteFetch";
+import waitForFonts from "./util/waitForFonts";
+import version from "./version";
```
amacneil, line 21: How come these are relative?

web/src/index.tsx
```diff
@@ -25,7 +26,6 @@ if (typeof process.env.SENTRY_DSN === "string") {
   initSentry({
     dsn: process.env.SENTRY_DSN,
     autoSessionTracking: true,
-    release: `${process.env.SENTRY_PROJECT}@${APP_VERSION}`,
```
amacneil, line 28: We should probably still keep this but set app version to git sha on web?

app/version.ts
amacneil: Also, at this point we should probably just import packageJson directly? This whole file seems redundant.

defunctzombie: I think we should import the package.json within web and desktop then pass the version to the app or maybe set it on the oscontext. I didn't change this yet to see what direction we might want to go in. I agree - the purpose of this file does not seem useful now.

app/index.ts
defunctzombie: Not sure how best to tackle this - but `@foxglove-studio/app` is the "external" name - we also use it within the package since we have a tsconfig paths setting. I tried changing to use `@/` shorthand for internal (within the app package) with a paths setting - but because we build ts sources in web and desktop we would need the path setting there too and then its weird to have `@/`. Using something like `@/` to indicate "root of this package source" would also make it nice to move all the sources under `src` folder without changing all the imports if we wanted to do that later. But tragically it doesn't play nice with our "build all typescript from source".

I made these relative since this is the top most file and paths are always `./` relative to it. The reason we use `@foxglove-studio/app` elsewhere is to avoid `../../` all over the place between internal components. I was unrelated that I happened to pick the public name as the internal alias (path) name for these imports and it works (webpack and jest resolve) because yarn flattens the imports into a top level node_modules. If it didn't do that or stops doing that we would not be able to resolve the @foxglove-studio/app imports as written in the app sources today.

web/src/index.tsx
defunctzombie: Yes - this is for future work tho cause we will need to setup a separate sentry project for web (or maybe can do it with environments?) and add sentry keys to vercel (since the build happens there it needs the keys).

app/index.ts
amacneil: Seems like we could just add a tsconfig path in app to match the public package name? (actually, we already have it). Is there any problem with that approach?

defunctzombie: No problem other than it being a little a-typical (and ties internal code paths to a public path which isn't _strictly_ necessary. I think if we used project references we would be able to use a different internal _paths_ alias but I'm not sure.

amacneil: Makes sense. Maybe atypical for node.js folks... looks more normal to me having spent more time writing go. However, it's also the pattern we use in our repo today so that makes it look normal too :)

I think that approach is cleaner than relative imports, or some "local root" syntax like `@/foo`. The latter syntax seems fine in a single repo, but in a monorepo it seems like it would add more confusion since you're never being explicit about what exactly you are importing.

PR #985, opened 2021-05-16
Remove overflow hidden from ButtonBase

### LINKED ISSUE #718: "New Node" button is cut off in Node Playground UI
To reproduce (in v0.6.0) just add Node Playground panel:

image

defunctzombie:
Hiding overflow results in buttons cutting off their internal content
if they downsize vertically.

This change also removes zIndex: 0 from the ButtonBase in favor of
using the zindex of the container.

Fixes #718

### DIFF
--- a/app/components/ButtonBase.tsx
+++ b/app/components/ButtonBase.tsx
@@ -111,7 +111,7 @@ class ButtonBaseImpl extends React.Component<Props> {
         onFocus={onFocus}
         onMouseLeave={this.onMouseLeave}
         onMouseUp={this.onMouseUp}
-        style={{ position: "relative", zIndex: 0, overflow: "hidden", ...style }}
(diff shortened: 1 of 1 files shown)


defunctzombie:
I'm not 100% that this is the correct fix but wanted to throw something out there for folks to comment on.

jtbandes:
No idea why these styles were there in the first place... maybe it was to help the button truncate text correctly if it is squished horizontally? I think we should try to get rid of this ButtonBase anyway in favor of FluentUI.

defunctzombie:
> think we should try to get rid of this ButtonBase anyway in favor of FluentUI.

I agree. Are we ok with this as a short term fix?

jtbandes:
Fine by me - doesn't seem to affect any screenshot tests 🤷

PR #993, opened 2021-05-17
Memo player activeData.topics with useShallowMemo in MessagePipeline

### LINKED ISSUE #846: Provider topics should not change often
> If they do they are probably not memoized properly. Please let the Foxglove team know if you see this warning.

This happens when connecting to a websocket (just simple rosbridge+roscore running inside a VM) on latest main.

Not sure what it means but it is not actionable so should probably not be displayed to the user at all.

image

defunctzombie:
Rather than warning that player topics are changing frequently which
produces false positives during player changes, useShallowMemo to keep
a stable reference to topics as long as the list of topics within the
array is unchanged.

Fixes #846


app/components/MessagePipeline/index.tsx
```diff
@@ -240,16 +239,7 @@ export function MessagePipelineProvider({
...
-  useShouldNotChangeOften(topics, () => {
-    sendNotification(
-      "Provider topics should not change often",
-      "If they do they are probably not memoized properly. Please let the Foxglove team know if you see this warning.",
-      "app",
-      "warn",
-    );
-  });
-
+  const topics: Topic[] | undefined = useShallowMemo(playerState.activeData?.topics);
```
jtbandes, line 242: Seems like a good direction — if we are going to support loading players from extensions, the app should be robust to this kind of input.

Although this is some extra work we are now doing on every tick that provides a performance pessimization. More little papercuts...

PR #994, opened 2021-05-17
Remove warning for subscribing right after player loaded.

defunctzombie:


amacneil:
I have some context on this one but can you add PR descriptions so that I know what bugs are being fixed for the release notes

PR #997, opened 2021-05-17
Move app to packages/@foxglove/studio-base

defunctzombie:
Since app components are re-used across desktop and are used as a 'package'
by these environments, we move the app folder into our packages folder to
mirror where we store other packages in the workspace.


packages/@foxglove/studio-base/version.ts
```diff
@@ -2,7 +2,7 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
-import packageJson from "../package.json";
+import packageJson from "../../../package.json";
```
defunctzombie, line 5: @amacneil this is a strangeness since we are traversing up several levels to grab a package.json that isn't "our module".

We can make desktop/web provide these as properties or a context to the rest of the app (they might be the only users of this info since this is for sentry only?

defunctzombie: https://github.com/foxglove/studio/pull/998 for thoughts

defunctzombie:
https://github.com/foxglove/studio/pull/1002 going to land that first

PR #998, opened 2021-05-17
Remove app/version.ts and use pkgInfo.ts from desktop/common

defunctzombie:
The package info (version, name, etc) is specific to the build environment.
App is an environment agnostic package.


app/components/Preferences.tsx
```diff
@@ -192,12 +191,12 @@ export default function Preferences(): React.ReactElement {
           <SectionHeader>Privacy</SectionHeader>
           <Stack tokens={{ childrenGap: theme.spacing.s1 }}>
             <Text style={{ color: theme.palette.neutralSecondary }}>
-              Changes will take effect the next time {APP_NAME} is launched.
+              Changes will take effect the next time Studio is launched.
```
jtbandes, line 194: These changes are sad. Can we just move it to a constant somewhere under app?

defunctzombie: Yea - I could do that. We could also change this to "the app"? Or "Changes will take effect on app restart".

jtbandes: 🤷 I think avoiding the app name in copy writing will get awkward.

app/Workspace.tsx
```diff
@@ -387,7 +386,7 @@ export default function Workspace(props: WorkspaceProps): JSX.Element {
             <PlayerStatusIndicator />
           </SToolbarItem>
           <SToolbarItem>
-            <TruncatedText>{currentSourceName ?? APP_NAME}</TruncatedText>{" "}
+            <TruncatedText>{currentSourceName ?? "Studio"}</TruncatedText>{" "}
```
amacneil, line 389: Shouldn't this be `Foxglove Studio`

desktop/common/pkgInfo.ts
```diff
@@ -0,0 +1,7 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import pkgInfo from "../../package.json";
+
+export default pkgInfo;
```
amacneil, line 7: this file is ridiculous

app/panels/WelcomePanel/index.tsx
```diff
@@ -70,8 +69,8 @@ function WelcomePanel() {
       <TextContent style={{ padding: 12 }}>
         <h2 style={{ fontSize: "1.5em", marginBottom: "0.8em" }}>Welcome</h2>
         <p>
-          {APP_NAME} is an integrated visualization and debugging tool for robotics. It allows you
-          to quickly and easily understand what’s happening in real-time, and provides a unique
+          Studio is an integrated visualization and debugging tool for robotics. It allows you to
```
amacneil, line 72: I like this change. But make it say `Foxglove Studio`

app/components/Preferences.tsx
amacneil: Actually I think it's fine to just say `Foxglove Studio` in copy like this. I'm not sure what the expected use case is where we want to build a distribution of our app with a different name.

At some point we are going to want to make all these strings localizable, which will require moving everything to variables, but for now I think having the app name explicitly in copy is fine.

I would suggest using `Foxglove Studio` rather than just `Studio` in most places though, since `Studio` is a bit of an overloaded product name on a typical developer desktop.

desktop/common/pkgInfo.ts
defunctzombie: This file is 🔪

PR #1002, opened 2021-05-17
Rename @foxglove-studio/app to @foxglove/studio-base

defunctzombie:
Our org name for packages is @foxglove. @foxglove-studio/app was selected
during the initial fork prior to standardizing on the @foxglove
org name.


amacneil:
lgtm, merge it quickly!

PR #1005, opened 2021-05-18
Replace rc-color-picker with FluentUI's colorpicker

wkalt:


app/components/ColorPicker/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+import { Color } from "regl-worldview";
+
+import {
+  getIColorFromColor,
+  getColorFromIRGB,
+} from "@foxglove/studio-base/components/ColorPicker/utils";
+
+export enum ButtonShape {
+  Circle,
+}
```
wkalt, line 22: separate file? just use a flag?

wkalt:
@jtbandes @defunctzombie this is ready for a look

app/panels/Publish/index.tsx
```diff
@@ -35,7 +37,7 @@ type Config = {
   datatype: string;
   buttonText: string;
   buttonTooltip: string;
-  buttonColor: string;
+  buttonColor: Color;
```
jtbandes, line 40: Changing this will not be backwards-compatible with existing saved layouts – we don't have any migration story right now but you might want to either keep this as a string, or do something like `string | Color` so you can at least guard against old values

wkalt:
there's an issue with this around coloring unlabeled markers with the topic tree. Taking a look at that.

app/components/ColorPicker/index.tsx
jtbandes: or just `buttonShape?: "circle" | "default"`

jtbandes (approved):
Looks good! Just need to remove `rc-color-picker` now :) Thanks for taking the time to do some good cleanups and add tests where appropriate.

PR #1006, opened 2021-05-18
Upgrade electron to v13 beta27

defunctzombie:
electron-chromedriver (used by spectron) downloads a version of chromedriver from electron builds based on the version of the _electron-chromedriver_ package. So if electron-chromedriver is v12 it will download chromedriver for electron v12. Spectron was using electron-chromedriver v12 which didn't support the newer chrome in the recent v13 betas. We override electron-chromedriver for spectron so the correct chromedriver version is downloaded.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,10 @@
       "web"
     ]
   },
+  "resolutions": {
+    "@types/node": "15.3.0",
+    "spectron/electron-chromedriver": "foxglove/chromedriver#7edaf1852173776e0e0db364acad12d4a9a8d17f"
+  },
   "devDependencies": {
     "@actions/core": "1.2.7",
     "@actions/exec": "1.0.4",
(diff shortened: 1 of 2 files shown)


vladimiry:
I believe that a simpler solution for this specific problem is patching the module rather than maintaining a fork, https://github.com/vladimiry/ElectronMail/blob/1d5955b80d705ce2a789fafa6c40198f32637986/patches/patch-package/electron-chromedriver%2B12.0.0.patch.

defunctzombie:
@vladimiry I didn't think about using a patch when I was working on this but do like the idea. Once nice property of the fork is that it will work regardless of where in the filesystem the electron-chromedriver package gets installed.

vladimiry:
I don't think I clearly see the linked to the electron-chromedriver package filesystem issue. I just highlighted the alternative solution the helped me with the same issue. The https://github.com/ds300/patch-package helps with creating and applying of such patches (git patches).

PR #1009, opened 2021-05-18
Patch @storybook/react to support a screenshot signal for story ready

defunctzombie:
The screenshot signal delays the story ready event which delays taking
a screenshot. This allows for screenshots to wait until the story has
updated internal state and rendered.


patches/storybook-react.patch
```diff
@@ -0,0 +1,27 @@
...
++    });
+   });
+ };
+ 
+@@ -128,7 +130,7 @@ function _renderMain() {
+             }
+ 
+             _context.next = 7;
+-            return render(element, rootEl);
++            return render(element, rootEl, storyContext.parameters?.screenshot?.signal);
```
jtbandes, line 24: should we call this `waitFor`?

defunctzombie: probably - I also toyed with the idea of moving out of screenshot and under parameters directly as a "ready" field or some such. Technically it isn't screenshot specific - it delays the story rendered event - which happens to be a thing chromatic waits on.

PR #1015, opened 2021-05-18
Switch RosbridgePlayer to lazy message

defunctzombie:
Rosbridge player was not ported to lazy message when the rosbag player was ported


defunctzombie:
MessageReaderStore is unused

PR #1019, opened 2021-05-19
Move rosMsgDefinitionText() to `stringify` in @foxglove/rosmsg

jhurliman:


defunctzombie:
maybe call this stringify? ideally some verb - `rosMsgDefinitionText` doesn't sound "functiony" to me.

PR #1021, opened 2021-05-19
Strongly type all EventEmitters

jhurliman:


packages/electron-socket/src/renderer/HttpServerRenderer.ts
```diff
@@ -8,7 +8,12 @@ import { HttpHandler, HttpRequest, HttpResponse } from "../shared/HttpTypes";
 import { Cloneable, RpcCall, RpcEvent, RpcResponse } from "../shared/Rpc";
 import { TcpAddress } from "../shared/TcpTypes";
 
-export class HttpServerRenderer extends EventEmitter {
+export interface HttpServerRendererEvents {
```
defunctzombie, line 11: Do we need to export this interface? Its internal for the type only right?

PR #1022, opened 2021-05-19
Move Map Panel to a separate extension package

defunctzombie:


defunctzombie:
https://github.com/Swatinem/rollup-plugin-dts
https://github.com/timocov/dts-bundle-generator

app/package.json
```diff
@@ -19,14 +19,14 @@
     "classnames": "2.3.1",
     "cytoscape": "3.19.0",
     "cytoscape-dagre": "2.3.2",
+    "extension-map-panel": "workspace:packages/extension-map-panel",
```
amacneil, line 22: should this be `@foxglove/extension-map-panel`? (for the package name, not the folder)

defunctzombie: I actually called the package `extension-map-panel` but thats largely a placeholder. I've thinking of moving the builtin extensions out of _packages_ or maybe an "extensions" folder within packages.

This line can be removed now - this was from a prior experiment I had.

PR #1023, opened 2021-05-19
ROS1Player publishing

jhurliman:
Advertise the publish capability from Ros1Player and implements setPublishers() and publish().


packages/ros1/src/TcpTypes.ts
```diff
@@ -53,7 +53,7 @@ export interface TcpServerEvents {
 }
 
 export interface TcpServer {
-  address(): TcpAddress | undefined;
+  address(): Promise<TcpAddress | undefined>;
```
defunctzombie, line 56: guessing this is intentional - why the change in this public interface? Are addresses for servers not known immediately? I guess not if the server hasn't started?

app/players/Ros1Player.ts
```diff
@@ -342,10 +330,70 @@ export default class Ros1Player implements Player {
...
+    // Unadvertise any topics that were previously published and no longer appear in the list
+    for (const topic of this._rosNode.publications.keys()) {
+      if (!topics.has(topic)) {
+        this._rosNode.unadvertise(topic);
+      }
+    }
+
+    // Advertise new topics
+    for (const pub of publishers) {
+      if (!this._rosNode.publications.has(pub.topic)) {
```
defunctzombie, line 370: maybe flip this to early continue and less nesting?

app/players/Ros1Player.ts
```diff
@@ -342,10 +330,70 @@ export default class Ros1Player implements Player {
...
+        const topic = pub.topic;
+        let msgdef: RosMsgDefinition[];
+        try {
+          msgdef = rosDatatypesToMessageDefinition(this._providerDatatypes, pub.datatype);
+        } catch (error) {
+          this._problems.push({
```
defunctzombie, line 376: When does this problem get cleared?

app/package.json
```diff
@@ -68,7 +68,7 @@
     "regl-worldview": "github:foxglove/webviz#workspace=regl-worldview&commit=48979519bb7c067c7a0162255d1ef083fef49199",
     "rehype-raw": "5.1.0",
     "reselect": "4.0.0",
-    "rosbag": "github:foxglove/rosbag.js#fc61125fbc01ff55698c5ed48987b2d9e7599572",
+    "rosbag": "github:jhurliman/rosbag.js#c299e1ca95fc15b16e70465f10a28bcb1b350552",
```
jtbandes, line 71: 🤔 why do we need to switch from one fork to another?

app/players/Ros1Player.ts
```diff
@@ -342,10 +330,70 @@ export default class Ros1Player implements Player {
...
+
   setPublishers(publishers: AdvertisePayload[]): void {
-    publishers = publishers.filter((p) => p.topic.length > 0);
-    if (publishers.length > 0) {
-      throw new Error("Publishing not supported");
+    if (!this._rosNode || this._closed) {
+      return;
+    }
+
+    // ROS1 only supports publishing topics that begin with "/"
```
jtbandes, line 357: what does this mean exactly? I know we've seen .bag files with non-prefixed topics in them. I think name resolution will make them resolve to the same thing (if a node is not namespaced) so why is publishing them not supported? This seems like a non-obvious limitation that would need to be documented somewhere for users at least, if it can't be removed

packages/ros1/src/nodejs/TcpServerNode.ts
```diff
@@ -26,14 +26,14 @@ export class TcpServerNode extends EventEmitter<TcpServerEvents> implements TcpS
...
+  address(): Promise<TcpAddress | undefined> {
     const addr = this._server.address();
     if (addr == undefined || typeof addr === "string") {
       // Address will only be a string for an IPC (named pipe) server, which
       // should never happen in TcpServerNode
-      return undefined;
+      return Promise.resolve(undefined);
     }
-    return addr;
+    return Promise.resolve(addr);
```
jtbandes, line 36: Prefer `async` functions rather than just returning Promise.resolve() everywhere

packages/ros1/src/TcpTypes.ts
jhurliman: It was inconsistent across the different implementations before, the code didn't compile (pass type checking). All of these methods have to return promises because the electron implementation requires async communication over postMessage

PR #1027, opened 2021-05-19
ci: only run yarn install on windows

amacneil:
Everyone is having problems with windows CI not playing nicely with yarn workspaces and our caching.

For now let's revert to only running `yarn install` on windows and not actually trying to build things.


defunctzombie:
The future is now.

PR #1033, opened 2021-05-20
Fix markerKeyPath lookup in 3d panel

### LINKED ISSUE #1007: Markers not matching linked variable highlight are not shown – TypeError: item[key] is not a function
Steps to repro:
- Enable a marker topic
- Click on an object
- Link one of its fields to a global variable
- Open the topic tree and enable the colored entry for the linked variable:  
  image

- Expected: all markers should show up including the matching marker, and only the matching one should be blue
- Actual: "TypeError: item[key] is not a function"
   image

defunctzombie:
Looking up markers by key path was trying to invoke a key path as a function.
This was old behavior from when bobjects was around. Lazy messages and
regular js objects provide getters and properties for message fields rather
than functions.

Fixes #1007


defunctzombie:
Marker is blue as desired.

Screen Shot 2021-05-19 at 5 14 26 PM

Issue #758, opened 2021-04-28
Ensure std_msgs/Empty is supported


jtbandes:
https://github.com/foxglove/studio/blob/08e5c0c113ffb7cc3c3a7de24fd2836856e0b43d/app/dataProviders/BagDataProvider.ts#L202-L206

jtbandes:
Related: https://github.com/cruise-automation/webviz/issues/588

jtbandes:
@jhurliman did you check on the warning message I specifically linked to above, when a bag contains std_msgs/Empty?

defunctzombie:
Fixed by https://github.com/foxglove/studio/pull/1479

PR #1038, opened 2021-05-20
Add @foxglove/rosmsg-msgs-common package, expose all common ROS datatypes through UserNodePlayer

jhurliman:
publish-all-the-things

Fixes https://github.com/foxglove/studio/issues/758 and https://github.com/cruise-automation/webviz/issues/588


packages/rosmsg-msgs-common/msgdefs/actionlib_msgs/GoalStatusArray.msg
```diff
@@ -0,0 +1,58 @@
+# actionlib_msgs/GoalStatusArray
```
jtbandes, line 1: Please add a gitattributes file to mark these as linguist-vendored so they are hidden from PRs and language stats:

packages/rosmsg-msgs-common/dist/index.js
```diff
@@ -0,0 +1,1110 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
jtbandes, line 1: Is storing this in git lfs going to be asking for trouble?

jhurliman: Probably, since we wouldn't be able to review textual diffs. It's also not a very large file which is why I'm not bothering to store it in LFS.

jtbandes: It's a dist file...do we really want to review textual diffs?

Issue #814, opened 2021-04-29
"Zoom to fit" crashes the Image panel if the first image has not been received


jhurliman:
I'm having an issue where Duckiebot is very to slow to start streaming camera images sometimes. When the image panel is showing all black (before the first image frame is received) and I try to change the zoom, selecting "Zoom to fit", I get this error:

```
TypeError
Cannot read property 'zoomAbs' of undefined
Call Stack
 ImageCanvas.onZoomFit
  panels/ImageView/ImageCanvas.tsx:256:32
 HTMLUnknownElement.callCallback
  node_modules/react-dom/cjs/react-dom.development.js:3945:14
 Object.invokeGuardedCallbackDev
  node_modules/react-dom/cjs/react-dom.development.js:3994:16
 invokeGuardedCallback
  node_modules/react-dom/cjs/react-dom.development.js:4056:31
 invokeGuardedCallbackAndCatchFirstError
  node_modules/react-dom/cjs/react-dom.development.js:4070:25
 executeDispatch
  node_modules/react-dom/cjs/react-dom.development.js:8243:3
 processDispatchQueueItemsInOrder
  node_modules/react-dom/cjs/react-dom.development.js:8275:7
 processDispatchQueue
  node_modules/react-dom/cjs/react-dom.development.js:8288:5
 dispatchEventsForPlugins
  node_modules/react-dom/cjs/react-dom.development.js:8299:3
 eval
  node_modules/react-dom/cjs/react-dom.development.js:8508:12
```

jtbandes:
It’s possible I introduced a regression while fixing the strict-boolean-expressions lint errors…

Issue #866, opened 2021-05-04
zoomPercentage for the image panel was <N>, but must be between 0 and 150. It has been reset to 100.


jhurliman:
<img width="463" alt="zoom-percentage" src="https://user-images.githubusercontent.com/195374/117069390-43cc1780-ace1-11eb-906e-0c24c6d9de01.png">

I don't know what the exact repro is, but I think this happens sometimes when resizing the window.

jtbandes:
Related to #729

amacneil:
Fixed in https://github.com/foxglove/studio/pull/1046

Issue #729, opened 2021-04-28
Improve Image panel zooming behavior


amacneil:
via @jtbandes @jhurliman:

- Zooming with scroll wheel / trackpad pinch feels too slow
- Zoom percentage / menu looks strange – the percentage itself may not be that useful, and these menu items could probably be buttons instead for better UX
- Easy to get in a weird state where the image doesn't stay aligned as you resize the panel
- Max zoom of 150% is too small. It should be 200%

mwisniowski:
Requesting to increase the zoom limit to 5000%

For camera applications there are times when you want to zoom in to a pixel level, eg when manually focussing a lens

jhurliman:
@mwisniowski https://github.com/foxglove/studio/pull/1046 increases the default max zoom to 300% and makes it a panel setting, so you can turn it up to 5000%. It also switches the default image scaling from bilinear to nearest neighbor, although this can be toggled in the image panel settings as well.

PR #1046, opened 2021-05-20
Increase max zoom to 300% and make it configurable, default to nearest neighbor image scaling and make it configurable

jhurliman:
This also cleans up a user-facing error message and catches some edge case undefined errors. Fixes https://github.com/foxglove/studio/issues/814, https://github.com/foxglove/studio/issues/866, https://github.com/foxglove/studio/issues/729


app/panels/ImageView/ImageCanvas.tsx
```diff
@@ -304,14 +310,23 @@ export default class ImageCanvas extends React.Component<Props, State> {
     // context: https://stackoverflow.com/questions/37135417/download-canvas-as-png-in-fabric-js-giving-network-error
     // read the canvas data as an image (png)
     canvas.toBlob((blob) => {
+      if (blob == undefined) {
+        sendNotification(
+          `Image download failed`,
+          `Failed to create an image from ${canvas.width}x${canvas.height} canvas`,
+          "user",
```
jtbandes, line 317: app?

jtbandes (commented):
We still have class components? 🤦

app/panels/ImageView/ImageCanvas.tsx
```diff
@@ -367,11 +385,19 @@ export default class ImageCanvas extends React.Component<Props, State> {
       config: { minValue, maxValue },
       onStartRenderImage,
     } = this.props;
-    if (!topic || !image) {
+    if (topic == undefined || image == undefined) {
       return;
     }
 
-    const imageMessage = image?.message as any;
+    const imageMessage = image.message as {
```
jtbandes, line 392: Don't we have a typedef for this?

app/panels/ImageView/index.tsx
```diff
@@ -644,11 +646,25 @@ const defaultConfig: Config = {
...
+    title: "Bilinear smoothing",
+  },
+  {
+    key: "maxZoom",
+    type: "number",
+    title: "Maximum zoom %",
+    placeholder: `${DEFAULT_MAX_ZOOM}`,
+    allowEmpty: true,
+    validate: (value) => Math.max(100, value),
+  },
```
jtbandes, line 667: Why do we want max zoom to be configurable as opposed to just not limiting the zoom level (or limiting it large enough so that just a few pixels would be visible)?

Issue #959, opened 2021-05-15
Data Source Info panel eats an unnecessary amount of CPU


jhurliman:
The `Data Source Info` panel is refreshing at full framerate (60fps) and due to the timestamps, it redraws every frame at a cost of 1-1.5ms per frame on my machine. For me this translates into ~20% CPU usage and "Very High" power usage from our app. When I change that panel to something else, CPU usage drops to <5% and power usage goes to "Moderate".

This could probably be fixed by throttling updates to 5-10Hz, there's no reason that panel needs to update so quickly.

defunctzombie:
This surfaces an important question. Should players (or message pipeline) throttle down the message stream itself? Maybe to 30hz even? We aren't aiming for game engine refresh rates so that seems sensible when playing back live?

jtbandes:
> We aren't aiming for game engine refresh rates

Why not?

defunctzombie:
> Why not?

Is 60 better than 30 in a noticeable way for playback? Guess it depends who you ask :/

jtbandes:
I suppose latency is more noticeable than framerate, but do think the difference between 30 and 60 is noticeable. Especially if you are trying to look at something that's flickering.

jhurliman:
For users reading this issue, it can now be mitigated by reducing the "Message rate (Hz)" setting (click the settings cog in the bottom-right corner).

I'm leaving this ticket open so we can address the root cause of expensive re-renders in the Data Source Info panel. I have a suspicion this affects the Parameters panel too, but that panel redraws much less often.

defunctzombie:
@jhurliman is this during bag playback or during live playback?

amacneil:
@2metres @defunctzombie we should address this as part of moving this info to the Data Source sidebar (or at least ensure that the DS sidebar doesn't suffer from the same issue)

defunctzombie:
Yea - One idea here is that we should move all the components that need to render every frame "end time is a culprit here" into a nested component so it doesn't render everything else. Maybe that isn't true and the whole display renders ever frame (cause of the Hz readings). I think we could drop those from sidebar display and focus on the more static data. If you want to track Hz we can do that in a panel or some other data source viewer since it won't be a thing for live sources.

jtbandes:
FYI the Hz readings are not computed dynamically. They are just `topic.numMessages / duration`.

jhurliman:
This has been resolved by moving all the Data Source Info panel data into the connection sidebar, making this panel redundant.

PR #1052, opened 2021-05-21
App setting to reduce the message pipeline framerate

jhurliman:
This adds an app-wide configuration setting titled "Message rate (Hz)" (`messageRate`) to lower the framerate of the message pipeline. The application will continue to render at 60fps providing a smooth UI, but the message pipeline can run at a fraction of the redraw framerate which reduces the frequency of redraws for panels such as Raw Messages and Data Source Info.

With those two panels open, Windows 10 energy consumption for FGS goes from Very High at 60 to High at 30 and Medium at 15.

This is one approach to resolving https://github.com/foxglove/studio/issues/959, although we should separately investigate why that panel takes so long to redraw.

### DIFF
--- a/app/AppSetting.ts
+++ b/app/AppSetting.ts
@@ -4,6 +4,7 @@
 
 export enum AppSetting {
   CRASH_REPORTING_ENABLED = "telemetry.crashReportingEnabled",
+  MESSAGE_HZ = "messageHz",
   ROS1_ROS_HOSTNAME = "ros1.ros_hostname",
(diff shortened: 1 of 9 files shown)


defunctzombie:
How does this interact with the playback speed selection? That is kind of related to this. Maybe this indicates a need to bring back the playbar for live playback and expose this setting on the playbar?

I think this needs some more thought around impact to live vs bag playback and possibly the setting name (or whether it is a setting)

I agree we need to do something for panels that should not be consuming so much CPU - but I'd like to start with first looking at those panels.

PR #1053, opened 2021-05-21
Change studio_msgs/ImageMarkerArray to visualization_msgs/ImageMarkerArray

jhurliman:
This matches the rosmsg-msgs-common package that was recently added

### DIFF
--- a/app/panels/ImageView/index.tsx
+++ b/app/panels/ImageView/index.tsx
@@ -41,6 +41,10 @@ import { CameraInfo, StampedMessage } from "@foxglove/studio-base/types/Messages
 import { PanelConfigSchema, SaveConfig } from "@foxglove/studio-base/types/panels";
 import { nonEmptyOrUndefined } from "@foxglove/studio-base/util/emptyOrUndefined";
 import filterMap from "@foxglove/studio-base/util/filterMap";
+import {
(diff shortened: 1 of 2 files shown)


amacneil:
Do you think we should support a non-standard message under `visualization_msgs` before it is accepted upstream?

Ref https://github.com/ros/common_msgs/pull/172

jhurliman:
@amacneil now that we package our own `.msg` files in the repo I think we have more flexibility to run ahead of `ros/common_msgs`. We can tell end-users to vendor our copy of the `visualization_msgs/ImageMarkerArray` msg file into their repo for now. If we follow the model of browser vendors, we could support both `visualization_msgs/ImageMarkerArray` and the alias `foxglove_msgs/ImageMarkerArray` (presumably we would support this alias across more than just the Studio product in the future).

defunctzombie:
I do not think we should be using an upstream package name from ros for a message that does not exist in that upstream package. This is confusing since we are not the authority on the visualization_msgs package and it isn't clear this message will be accepted. If you now go lookup the visualization_msgs package you won't find this message there. We should avoid such strangeness and collisions.

amacneil:
Bump on this since it was merged but still needs discussing

defunctzombie:
@jhurliman I think its best to change back to studio_msgs

PR #1057, opened 2021-05-22
Revert "Initial structure for creating and executing extensions"

defunctzombie:
Reverts foxglove/studio#886

This PR was not ready to merge.


defunctzombie:
Instead of reverting the whole PR - which also reverts the foxe tool work which is separate from the app changes I am removing pieces we are not ready to merge.

https://github.com/foxglove/studio/pull/1062

Issue #895, opened 2021-05-07
Load extensions using x-foxglove-extension: URIs instead of file:


jhurliman:
We may be able to lock down the ability to import("file:///...") URLs by removing `file:` from the `script-src` CSP and replacing it with `x-foxglove-extension:`. This assumes each extension is compiled down to a single JS file, but that's how our extension template should be setup.

PR #1063, opened 2021-05-22
Remove "file:" from script-src in desktop.

defunctzombie:
This is counter to our security model which tries to _prevent_
having renderer code arbitrarily access any file on the user's machine.


defunctzombie:
Replaced by https://github.com/foxglove/studio/pull/1064 which tackles the broader change for https://github.com/foxglove/studio/issues/895

PR #1064, opened 2021-05-22
Change ExtensionsContext to ExtensionRegistryContext

defunctzombie:
* Move extension loading to desktop via ExtensionRegistryProvider
* Load builtin extensions and user extensions via desktop bridge

TODO
- ~~Make one "extension" package for all our builtin extensions~~
- ~~Does this approach work for _web_ build and builtin panels?~~


app/context/ExtensionsContext.ts
```diff
@@ -3,12 +3,16 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
 import { createContext, useContext } from "react";
+import { ExtensionContext } from "@foxglove/studio";
 
-import { Extensions } from "@foxglove/studio-base/Extensions";
+export type Extension = {
+  name: string;
```
jhurliman, line 9: Defined like this, it would require extension authors to export a `const name` from their extension. This would be a second source of truth for the name of a package beyond `package.json` name field. Speaking of which, is there still a wrapping class (ExtensionInstance or whatever we call it) that we can attach `packageJson` to or where should that object live?

defunctzombie: This is an internal representation rather than what the extension author's module exports. I'll probably have a separate interface ExtensionModule or some such that represents what we expect from the author's module. This is for a parsed representation of the extension for internal use - so we can read the package manifest once and populate various fields like this.

I'm not sure if this is the final form either - just the general direction I was taking this.

app/panels/HelloWorldExtension/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { useMemo } from "react";
+
+import Panel from "@foxglove/studio-base/components/Panel";
+import PanelToolbar from "@foxglove/studio-base/components/PanelToolbar";
+import { useExtensionRegistry } from "@foxglove/studio-base/context/ExtensionRegistryContext";
+
+// fixme - help should be loaded via fetch as well from the extension?
```
jhurliman, line 11: We can just add it as a second parameter to `registerPanel()`. That seems safer in a couple of different ways, most importantly compile time type checking.

defunctzombie: What would we provide in that parameter? The path to the markdown file? Relative to repo root? Maybe this is more appropriate for the extension's package.json and have our extension loader try to read the markdown from there?

defunctzombie: After discussing in slack - we decided to have a 'docs' folder with markdown files matching panel names. Importing markdown in the code requires more typescript setup (to educate typescript about importing markdown) and webpack configuration. To avoid adding complexity to both we will have extension loading handle reading docs folders.

desktop/common/types.ts
```diff
@@ -43,14 +43,20 @@ interface Storage {
   delete(datastore: string, key: string): Promise<void>;
 }
 
+type DesktopExtension = {
```
jhurliman, line 46: This type seems like it would generalize out to loading extensions in web too, not just desktop?

extensions/hello-world-panel/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "hello-world-panel",
+  "main": "index.js",
```
jhurliman, line 3: Not `dist/index.js`?

defunctzombie: Could make it dist/index - didn't matter too much for me since the target bundle folder is not co-located with the src folder.

desktop/renderer/Root.tsx
```diff
@@ -14,6 +14,7 @@ import {
 } from "@foxglove/studio-base";
 
 import { Desktop } from "../common/types";
+import ExtensionRegistryProvider from "./components/ExtensionRegistryProvider";
```
jtbandes, line 17: move to `providers` dir

desktop/webpack.extensions.config.ts
```diff
@@ -0,0 +1,113 @@
...
+
+export default (_: unknown, argv: WebpackArgv): Configuration => {
+  const isDev = argv.mode === "development";
+  const isServe = argv.env?.WEBPACK_SERVE ?? false;
+
+  const allowUnusedLocals = isDev && isServe;
+
+  return {
+    externals: {
+      "@foxglove/studio": "studio",
```
jtbandes, line 20: maybe FoxgloveStudio?

web/src/components/ExtensionRegistryProvider.tsx
```diff
@@ -0,0 +1,81 @@
...
+    // eslint-disable-next-line no-new-func
+    const fn = new Function(
+      "react",
+      "React",
+      "reactDom",
+      "studio",
+      `${unwrappedSource}; return __studio_extension_entrypoint`,
+    );
+
+    const wrappedExtensionModule = fn(React, React, ReactDom, {}) as ExtensionModule;
```
jtbandes, line 36: As an alternative, maybe consider webpack `options.library.type: 'this'` and `wrappedExtensionModule = fn.call({}, React, ...).__studio_extension_entrypoint`? Avoids needing to append to the module source. But I guess otherwise doesn't make much of a difference.

web/webpack.extensions.config.ts
```diff
@@ -0,0 +1,109 @@
...
+        },
+      ],
+    },
+
+    plugins: [
+      new ForkTsCheckerWebpackPlugin({
+        typescript: {
+          configOverwrite: {
+            compilerOptions: {
+              noUnusedLocals: !allowUnusedLocals,
```
jtbandes, line 94: I changed this to allowUnusedVariables and added another entry here for unused function params — please rebase on main

web/src/components/ExtensionRegistryProvider.tsx
jtbandes: Or make `exports` a parameter to the function and pass an initially empty array

extensions/hello-world-panel/package.json
defunctzombie: flattened these into one builtin extensions folder

desktop/renderer/Root.tsx
defunctzombie: There are other providers in this folder - I will move them all after merging this PR.

defunctzombie: nvm - decided to move in this PR

PR #1065, opened 2021-05-22
Skip reading extensions directory if it does not exist

wkalt:


desktop/preload/index.ts
```diff
@@ -113,7 +114,12 @@ const desktopBridge: Desktop = {
 
     const homePath = (await ipcRenderer.invoke("getHomePath")) as string;
     const rootFolder = pathJoin(homePath, ".foxglove-studio", "extensions");
-    const rootFolderContents = await readdir(rootFolder, { withFileTypes: true });
+    let rootFolderContents: Dirent[];
+    try {
+      rootFolderContents = await readdir(rootFolder, { withFileTypes: true });
+    } catch {
```
defunctzombie, line 120: Lets log the error in this catch so we can debug should there be some reading error.

wkalt: I changed it to check for directory existence instead. I think if it's an error other than that, throwing is probably appropriate.

PR #1068, opened 2021-05-23
Use a single tsconfig for eslint to reduce memory usage and speed up linting

jtbandes:
Because the typescript-eslint plugin does not receive information from eslint about the overall set of files to be linted, it's forced to hold all TypeScript projects in memory at once. Since our monorepo contains several separate projects, we run into memory limits when running eslint on all the projects in a single run.

Making eslint use just a single tsconfig means it doesn't need to hold as much in memory (recommendation from https://github.com/typescript-eslint/typescript-eslint/issues/1192 and [typescript-eslint monorepo docs](https://github.com/typescript-eslint/typescript-eslint/blob/master/docs/getting-started/linting/MONOREPO.md#important-note-regarding-large--10-multi-package-monorepos)). It also makes linting much faster (CI job completes in 2min vs 8min).

This approach works as long as our packages' tsconfigs don't vary much beyond the base config. Another alternative would be to run eslint separately on each project's files. This would be a bit trickier to maintain since it requires a script to make multiple eslint runs, and possibly a separate eslintrc file for each project to point it at the appropriate tsconfig (although it's not clear this part is necessary).

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -34,7 +34,7 @@ parser: "@typescript-eslint/parser"
 parserOptions:
   ecmaVersion: 2020
   sourceType: module
-  project: ./**/tsconfig.json
+  project: ./tsconfig.eslint.json
 
 settings:
   react:
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -182,7 +182,7 @@ rules:
 
 overrides:
   - rules:
-      "@typescript-eslint/no-explicit-any": warn
+      "@typescript-eslint/no-explicit-any": off
     files:
       - "app/components/Autocomplete.stories.tsx"
       - "app/components/Autocomplete.tsx"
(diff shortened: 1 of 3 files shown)


jtbandes:
typescript-eslint and tsc clearly disagree on some types here:

- [GridBuilder.ts](https://github.com/foxglove/studio/blob/main/app/panels/ThreeDimensionalViz/GridBuilder.ts) lint error without type annotation, no lint error after adding seemingly useless type annotation:
  image

  image

- [MarkerStory.tsx](https://github.com/foxglove/studio/blob/main/app/panels/ThreeDimensionalViz/stories/MarkerStory.tsx) lint error with cast, TS error without cast:
  image

  image

jtbandes:
Those errors turned out to be differences with the `jsx` and `lib` settings.

defunctzombie:
good find - maybe my vscode will be happier too

PR #1076, opened 2021-05-25
Map panel as builtin extension

defunctzombie:


defunctzombie:
I'm not sure this is the route to go for our builtin in-repo extensions. Many of them use other shared features that will not be exposed. We will re-visit as we travel further down the extensions path.

PR #1079, opened 2021-05-26
Account sidebar & Firebase login

jtbandes:
- Add AuthContext with authentication functions, FirebaseAuthProvider which implements auth by opening `https://foxglove.dev/auth` in an external browser, and FirebaseAppContext which manages the firebase config.
- Add an Account sidebar with UI for the above functionality.
- Handle `foxglove://auth/login-complete` and allow pasting auth token as fallback (protocol handler doesn't work in dev mode on all platforms).
- Convert the app to support a single running instance only, instead of multiple instances, so that the foxglove:// link is handled by the calling app instead of opening a new app.

image image image image


desktop/main/index.ts
```diff
@@ -28,272 +25,281 @@ import {
...
-});
-
-// works on osx - even when app is closed
-// tho it is a bit strange since it isn't clear when this runs...
-const openUrls: string[] = [];
-app.on("open-url", (ev, url) => {
-  if (!url.startsWith("foxglove://")) {
+
+  // If another instance of the app is already open, this call triggers the "second-instance" event
+  // in the original instance and returns true.
```
jtbandes, line 62:
```suggestion
  // in the original instance and returns false.
```

desktop/main/loadFilesToOpen.ts
```diff
@@ -0,0 +1,43 @@
...
+ * Our app has support for working with File instances in the renderer. This avoids extra copies
+ * while reading files and lets the renderer seek/read as necessary using all the browser
+ * primitives for File instances.
+ *
+ * Unfortunately Electron does not provide a way to create or send File instances to the renderer.
+ * To avoid sending the data over our context bridge, we use a hack.
+ * Via the debugger we inject a DOM event to set the files of an <input> element.
+ */
+const inputElementId = "electron-open-file-input";
+export default async function loadFilesToOpen(
```
defunctzombie, line 16: maybe "injectFilesToOpen" ?

desktop/preload/index.ts
```diff
@@ -101,6 +101,9 @@ const ctx: OsContext = {
 };
 
 const desktopBridge: Desktop = {
+  loginViaExternalBrowser: async (): Promise<string> => {
```
defunctzombie, line 104: could leave out browser and just call it "nativeAuth" or some such? Either way I'd change login to auth since thats what you seem to have settled on with the context name and url.

jtbandes:
We have some ideas around testing, but will tackle in separate PRs.

It seems like wrapping the test run in `firebase emulators:exec` is the most well-supported way of having a clean test environment. I don't think it would be easy to spin up separate emulators for each test, but a test can probably use a unique firebase project id (for example as [described here](https://timo-santi.medium.com/jest-testing-firebase-functions-with-emulator-suite-409907f31f39)) so that we are able to run tests in parallel.

PR #1081, opened 2021-05-26
Add foxglove_msgs package, move visualization_msgs/ImageMarkerArray -> foxglove_msgs/ImageMarkerArray

jhurliman:


packages/foxglove_msgs/dist/index.d.ts
```diff
@@ -0,0 +1,14 @@
+import { RosMsgDefinition } from "@foxglove/rosmsg";
```
amacneil, line 1: should dist be committed?

jhurliman: Yes, it has to because monorepo

PR #1082, opened 2021-05-26
Update PanelCatalogContext and providers to expose extension panels

defunctzombie:
The panel list receives panels from the panel catalog. Previously the
panel catalog only exposed builtin panels. This change introduces
extension panels via the panel catalog.

Panel categories are removed in favor of one list of panels. Studio internal panels are hidden behind an application setting.

image


defunctzombie:
An idea for `Studio Debug` panels might be to hide those from the add panel list under a flag?

amacneil:
why is `Hello` at the bottom in your sceenshot?

defunctzombie:
> why is `Hello` at the bottom in your screenshot?

it is part of our "extensions" for now and only shown in dev builds. https://github.com/foxglove/studio/blob/main/extensions/index.ts

I'm still unsure on how to proceed with "extensions" written within the monorepo so it exists as a placeholder for now.

PR #1083, opened 2021-05-27
Rename useConfig to usePanelConfig and add default param.

defunctzombie:
Rather than specifying a default as a static property on the function component,
pass it as an argument to usePanelConfig similar to useState initialization.


defunctzombie:
I'm gonna keep the useConfig name since I've namespaced it under `panel` https://github.com/foxglove/studio/pull/1080 and will do a follow-up PR to add the default config feature to the hook once 1080 merges.

PR #1085, opened 2021-05-27
Switch from ts-loader to esbuild-loader

jhurliman:


jhurliman:
This is all sadness. The integration test fails because webpack compilation runs out of JS heap space in CI (although it passes fine on my machine), and switching babel-jest to esbuild-jest encounters some sort of AST failure on some tests:

```
TypeError: ...\foxglove\studio\app\panels\ThreeDimensionalViz\SceneBuilder\index.ts: Property typeName of TSTypeReference expected node to be of a type ["TSEntityName"] but instead got "MemberExpression"
```

PR #1086, opened 2021-05-27
Remove Hippocratic license

amacneil:
We wish to make use of Foxglove Studio easy for consumers. To the extent possible, we should rely on popular and well-understood licenses approved by the OSI.

### DIFF
--- a/licenses.test.ts
+++ b/licenses.test.ts
@@ -20,7 +20,6 @@ const ALLOWED_LICENSES = [
   "WTFPL",
   "Unlicense",
   "OFL-1.1",
-  "Hippocratic-2.1",
 ];
 const EXCLUDED_PACKAGES = ["gl-vec3@1.1.3", "pngparse@2.0.1"];
 


jhurliman:
Look in `licenses.test.ts` at:

```
// Uncomment this and the printLicenses() call above to print the full list of
// found licenses. This can be helpful when debugging failures for this test
```

amacneil:
Yeah, I just spent a while digging around in this library. It seems that the `err` callback is only used for input validation, and any actual license issues result in a `process.exit(1)` call (which doesn't play nicely with jest). 

Maybe we should move this to a separate script in CI rather than jest test (or find an alternative that can actually return an error).

jhurliman:
@amacneil you can change the behavior by not passing in `onlyAllow`, but then you have to write your own license validation. It didn't seem worth it to me for these rare cases where you need to investigate a specific failure. The package you're looking for is:

```
@react-leaflet/core@1.1.0
react-leaflet@3.2.0
```

amacneil:
Still no way to avoid the `process.exit` call. What do you think of https://github.com/foxglove/studio/pull/1088

amacneil:
Replaced by https://github.com/foxglove/studio/pull/1088

PR #1088, opened 2021-05-27
Move license-check to CI script

amacneil:


jest.config.json
```diff
@@ -11,7 +11,6 @@
   "transform": {
     "\\.[jt]sx?$": ["babel-jest", { "rootMode": "upward" }]
   },
-  "testMatch": ["**/licenses.test.ts"],
```
amacneil, line 14: not sure if I can remove more of this file since this was the only test it covered, and everything else seems to be under `projects`?

jest.config.json
```diff
@@ -7,11 +7,7 @@
     "<rootDir>/packages/*/jest.config.json",
     "<rootDir>/packages/@foxglove/*/jest.config.json"
   ],
-  "testRunner": "jest-circus/runner",
-  "transform": {
-    "\\.[jt]sx?$": ["babel-jest", { "rootMode": "upward" }]
-  },
-  "testMatch": ["**/licenses.test.ts"],
+  "testMatch": ["none"],
```
amacneil, line 10: @defunctzombie does this make sense to you? I can't find any way to tell this root jest config to not scoop up any files itself (because everything is imported via `projects`). If I delete this lint entirely a bunch of jest tests start throwing errors.

jest.config.json
```diff
@@ -7,11 +7,7 @@
...
   "//": "Native find is slow because it does not exclude files: https://github.com/facebook/jest/pull/11264#issuecomment-825377579",
   "haste": { "forceNodeFilesystemAPI": true }
```
amacneil, line 12: @jtbandes I see you added this. Should it be in the root `jest.config.json` though? Since the projects don't inherit from this file, wouldn't it make more sense to have this in each project `jest.config.json`?

jest.config.json
defunctzombie: What do you mean by "not scoop up any files" ? If you remove anything that isn't the project configuration from this jest config it won't match any files.

defunctzombie: You can see what the file was like before:
https://github.com/foxglove/studio/commit/3f22edf7e466ef867ef42e11a63ba0d57d3da1de

amacneil: I tried removing everything except the project configuration, jest spews out millions of errors. Try it?

jest.config.json
amacneil: nevermind, I see it already exists in the other config files

jest.config.json
amacneil: fixed by removing the self reference from projects. thanks

PR #1090, opened 2021-05-27
Add default config to useConfig panel api

defunctzombie:
The default config is passed to useConfig and merges with any persisted
configuration.

### DIFF
--- a/app/PanelAPI/useConfig.test.tsx
+++ b/app/PanelAPI/useConfig.test.tsx
@@ -1,0 +1,119 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+/** @jest-environment jsdom */
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+//
+// This file incorporates work covered by the following copyright and
+// permission notice:
+//
+//   Copyright 2019-2021 Cruise LLC
+//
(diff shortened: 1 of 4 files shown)


app/PanelAPI/useConfig.test.tsx
```diff
@@ -0,0 +1,119 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+//
+// This file incorporates work covered by the following copyright and
+// permission notice:
+//
+//   Copyright 2019-2021 Cruise LLC
+//
+//   This source code is licensed under the Apache License, Version 2.0,
+//   found at http://www.apache.org/licenses/LICENSE-2.0
+//   You may not use this file except in compliance with the License.
```
jtbandes, line 17: something is wrong here

app/PanelAPI/useConfig.test.tsx
```diff
@@ -0,0 +1,119 @@
...
+          <PanelIdContext.Provider value="some-id">{children}</PanelIdContext.Provider>
+        </CurrentLayoutContext.Provider>
+      );
+    };
+
+    const { result } = renderHook(() => useConfig<{ foo: string }>({ foo: "bar" }), { wrapper });
+    expect(result.current[0]).toEqual({ foo: "bar" });
+
+    state.addPanelsStateListener((newState) => {
+      expect(newState.configById["some-id"]?.foo).toEqual("reset");
```
jtbandes, line 106: could also use a jest.fn with expect(f.mock.calls).toEqual, for more linear test flow (could then remove expect.assertions)

jtbandes:
Has this been superseded by other changes at this point?

defunctzombie:
Yea - This will need to be revisited in some other form because of how the panel settings sidebar wants to interact with panel settings. I'm not a fan of the current interactions and how we specify defaults today but this PR is stale and another approach will be needed.

PR #1093, opened 2021-05-27
Update to latest roslibjs

jhurliman:


ci/license-check.ts
```diff
@@ -25,12 +25,7 @@ const ALLOWED_LICENSES = [
   "OFL-1.1",
 ];
 
-const EXCLUDED_PACKAGES = [
-  "@react-leaflet/core@1.1.0",
-  "gl-vec3@1.1.3",
-  "pngparse@2.0.1",
-  "react-leaflet@3.2.0",
-];
+const EXCLUDED_PACKAGES = ["@react-leaflet/core@1.1.0", "gl-vec3@1.1.3", "react-leaflet@3.2.0"];
```
amacneil, line 28: nice

licenses.test.ts
```diff
@@ -1,63 +0,0 @@
-// This Source Code Form is subject to the terms of the Mozilla Public
```
amacneil, line 1: oops

jhurliman: We don't need this second copy, right?

amacneil: no, in fact I have no idea how it didn't get deleted in my PR, since I started that PR with a `git mv`. I guess it got restored in my rebasing.

PR #1099, opened 2021-05-28
Another round of warning fixes

jhurliman:
> ✖ 643 problems (0 errors, 643 warnings)


app/components/MessagePathSyntax/MessagePathInput.tsx
```diff
@@ -181,12 +182,12 @@ class MessagePathInputUnconnected extends React.PureComponent<
   _onChange = (event: React.SyntheticEvent<HTMLInputElement>, value: string) => {
     // When typing a "{" character, also  insert a "}", so you get an
     // autocomplete window immediately for selecting a filter name.
-    if ((event.nativeEvent as any).data === "{") {
-      const newCursorPosition = (event.target as any).selectionEnd;
+    if ((event.nativeEvent as InputEvent).data === "{") {
+      const target = event.target as HTMLInputElement;
+      const newCursorPosition = target.selectionEnd ?? 0;
```
jhurliman, line 187: I tested, and feeding either `0` or `null` into the `value.slice()` call below has the same effect so this is not a functionality change.

app/components/MessagePathSyntax/useLatestMessageDataItem.ts
```diff
@@ -34,11 +34,11 @@ export function useLatestMessageDataItem(path: string): MessageAndData | undefin
   const addMessages = useCallback(
     (
       prevMessageAndData: MessageAndData | undefined,
-      messages: readonly MessageEvent<unknown>[],
+      messages: Readonly<MessageEvent<unknown>[]>,
```
jtbandes, line 37: Hmm, just tested and I think this is identical to `readonly MessageEvent<unknown>[]` before the change – why this change?

app/util/layout.ts
```diff
@@ -336,9 +335,8 @@ export const removePanelFromTabPanel = (
   if (path.length === 0) {
     newTree = undefined;
   } else {
-    // eslint-disable-next-line no-restricted-syntax
-    const update = createRemoveUpdate(currentTabLayout ?? null, path);
-    newTree = updateTree<string>(currentTabLayout!, [update]);
+    const update = createRemoveUpdate(currentTabLayout ?? ReactNull, path);
```
jtbandes, line 338: 🤷 would've left this as eslint-disable since this function is not React related.

PR #1100, opened 2021-05-28
Minify production builds with ESBuildMinifyPlugin

jhurliman:
This reduces the CI integration test time from `7m 53s` to `2m 7s` (~3.7x speedup). As far as the outputs:

### TerserPlugin

```
  Assets:
    ts.worker.js (4.29 MiB)
    main.js (6.49 MiB)
    d6a81be6fe1edb49fdfd.glb (668 KiB)
    f4a990eab21714a17a87.bin (1.75 MiB)
    fa02fc7749c776e4c755.wasm (292 KiB)
    572.js (299 KiB)
    monaco-api.js (2.52 MiB)
    prettier.js (3.53 MiB)
    144.js (2.68 MiB)
```

### ESBuildMinifyPlugin

```
  Assets:
    ts.worker.js (4.41 MiB)
    main.js (6.53 MiB)
    d6a81be6fe1edb49fdfd.glb (668 KiB)
    f4a990eab21714a17a87.bin (1.75 MiB)
    fa02fc7749c776e4c755.wasm (292 KiB)
    572.js (300 KiB)
    monaco-api.js (2.5 MiB)
    prettier.js (3.63 MiB)
    419.js (2.81 MiB)
```

It produces a 0.6% larger main.js in a fraction of the time. Other files get larger or smaller (i.e. prettier.js is 2.8% larger) but nothing seems to change by more than a few percent.

## Other Changes

* `universal-url` was not being shimmed out of the web target, which added ~300KB to the minified build. We target ES2020 for web and Node.js >= 14 for the main process, so there's no need for a URL ponyfill. It has been removed in favor of using native URL directly everywhere
* We were setting environment variables directly in package.json scripts, which is not supported in Windows shells. `cross-env` is the solution for this, and has been prefixed to the relevant commands
* Our base tsconfig disabled require()ing tslib in favor of including it directly into output code. This is a cruel thing to do to anyone who tries to use one of our libraries in their own projects, and it's the reason our output main.js has 12 separate copies of tslib baked into it. We now follow the default tsc behavior around tslib usage


defunctzombie:
Can you share some details about the URL, cross-env, and base tsconfig changes? I'm guessing they have something to do with URL being auto-injected properly now but I want to avoid guessing.

app/webpack.ts
```diff
@@ -142,6 +143,11 @@ export function makeConfig(
             { loader: "css-loader", options: { sourceMap: true } },
           ],
         },
+        {
+          test: /\.s?css$/,
+          loader: "esbuild-loader",
+          options: { loader: "css", minify: true },
+        },
```
amacneil, line 150: do we need this in _addition_ to the other css/style/sass loaders?

also, should `minify` be true only for production?

jhurliman: Yes, it is an additional step on top of the other css/style/sass loaders and performs a different task. Good catch with toggling `minify`, fixing now.

app/webpack.ts
```diff
@@ -168,6 +174,10 @@ export function makeConfig(
         },
       ],
     },
+    optimization: {
+      removeAvailableModules: true,
+      minimizer: [new ESBuildMinifyPlugin({ target: "es2020" })],
+    },
```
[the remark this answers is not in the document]
jhurliman: This doesn't enable or disable minification, that's a separate `minify` param that we leave at the default (which is to enable in production and disable otherwise). This swaps out TerserPlugin for ESBuildMinifyPlugin, which is a no-op in dev builds.

app/webpack.ts
amacneil: out of curiosity, what is the `different task`?

jhurliman: Minification! Here's a breakdown, as I understand it:

* `sass-loader` converts SCSS into valid CSS
* `css-loader` loads CSS files from disk into strings. It understands CSS import and url() statements, and can resolve those to more files loaded from disk
* `esbuild-loader/css` minifies this CSS
* `style-loader` injects CSS into the DOM by writing `<style>` tags

The order that these execute in is important.

amacneil: I see. I didn't realize we weren't doing any minification of css today. 

I dream of a future where we don't need 4 loaders just for css. Maybe that's just the day we switch to esbuild entirely :smile:

PR #1109, opened 2021-05-28
Fix imports and add ForkTsChecker for web build

jtbandes:


amacneil:
This is incorrect because ForkTsCheckerWebpackPlugin already exists in the web build. Removing in https://github.com/foxglove/studio/pull/1144

PR #1124, opened 2021-06-01
Fix updatePanelConfig call in CameraInfo

defunctzombie:
Commit 7183c3f0 removed updatePanelConfig, however CameraInfo had an
_any_ that hid the type error resulting in updatePanelConfig not being
a function. This became a runtime error.

Surfaced via https://sentry.io/organizations/foxglove/issues/2430989186/?project=5649767&referrer=slack


defunctzombie:
@jtbandes handing this off to you to do a spot check/test that it behaves as expected.

jtbandes (approved):
Thanks. In case there was any question whether we should go fix all the `any`s...

PR #1084, opened 2021-05-27
Upgrade to TypeScript 4.3.2

jtbandes:
Waiting for the following fixes to be released:
- [x] ~https://github.com/yarnpkg/berry/pull/2889~ https://github.com/yarnpkg/berry/pull/2543
- [x] https://github.com/babel/babel/pull/13381
- [x] https://github.com/prettier/prettier/issues/10642 (fix available by overriding parser to `babel-ts`)

PR #1126, opened 2021-06-01
Add YARN_ENABLE_IMMUTABLE_INSTALLS=false for dependabot fix

jtbandes:
This action is currently failing with `The lockfile would have been modified by this install, which is explicitly forbidden.` However, the purpose of the action is to modify the lockfile.

This likely broke in the yarn upgrade (part of #1084).

### DIFF
--- a/.github/workflows/dependabot-fix.yml
+++ b/.github/workflows/dependabot-fix.yml
@@ -53,6 +53,10 @@ jobs:
           restore-keys: ${{ runner.os }}-yarn-
 
       - run: yarn install --skip-builds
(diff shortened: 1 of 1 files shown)


amacneil (approved):
I'm fairly sure this fix is correct. The yarn 3 [changelog](https://github.com/yarnpkg/berry/blob/master/CHANGELOG.md) says that immutable mode was changed to default on for CI environments in v3+.

This github action is explicitly intended to update the lockfile (that's the whole point of the action), so disabling immutable mode is what we want.

PR #1133, opened 2021-06-02
Remove separate extension bundle

defunctzombie:
The separate extension bundle breaks hot reloading (for yet unknown reasons),
and since we are not moving our builtin extensions to this package we don't
need to have this extra cruft around.


jtbandes:
I wouldn't say the separate bundle itself breaks hot reloading – the `injectClient` and `injectHot` settings break it.

web/src/providers/ExtensionLoaderProvider.tsx
```diff
@@ -10,24 +10,34 @@ import { ExtensionDetail, ExtensionLoader, ExtensionLoaderContext } from "@foxgl
...
-        source: source,
-      },
-    ];
-
-    const loader: ExtensionLoader = {
-      getExtensions: () => Promise.resolve(extensions),
-    };
-    return loader;
+    try {
+      const builtinExtensionFetch = await fetch("/builtinextensions.js");
```
jtbandes, line 19: should this be removed?

PR #1134, opened 2021-06-02
Migrate to @foxglove eslint config

amacneil:
Shared config lives at https://github.com/foxglove/eslint-plugin

- Common rules removed from `.eslintrc.yaml`
- Typescript-specific rules moved to override to avoid catching the small number of `.js` files in this repo


.eslintrc.yaml
```diff
@@ -15,125 +15,49 @@ ignorePatterns:
...
 
 settings:
-  react:
-    version: detect
   import/internal-regex: "^@foxglove"
 
 rules:
-  file-progress/activate: 1
+  "@foxglove/license-header": error
+  "@foxglove/strict-equality": off # will replace some no-restricted-syntax in the future
```
amacneil, line 30: intention is to enable this and remove the corresponding no-restricted-syntax entries after we finalize the recommendations around `!= undefined` vs `!== undefined` vs `!== null`

PR #1139, opened 2021-06-02
Revert "Upgrade electron to v13.0.1"

defunctzombie:
Reverts foxglove/studio#1131


defunctzombie:
Welp - that wasn't the problem.

PR #1017, opened 2021-05-18
Move app to packages/studio-base/src

amacneil:
There are two commits in this PR. The first one is mechanical (rename files) and can be easily recreated. The second commit contains actual code modifications.

Please keep this PR easy to rebase in the future, by amending any code changes to the second commit. Don't add additional commits.

How to rebase this PR (trash and recreate the first commit, then cherry pick the second commit):

```sh
# start from a fresh master
git status  # start with a clean local tree
git fetch
git reset --hard origin/main

# create the rename commit
mkdir -p packages/studio-base && \
git mv app packages/studio-base/src && \
git mv packages/studio-base/src/.eslintrc.yaml packages/studio-base/ && \
git mv packages/studio-base/src/*.json packages/studio-base/ && \
git mv packages/studio-base/src/webpack* packages/studio-base/ && \
git mv packages/studio-base/src/Webpack* packages/studio-base/ && \
git commit -m 'Rename files'

# cherry pick the actual code changes
git cherry-pick <sha>  # use the commit sha from the second commit in this PR
# fix any merge conflicts
```

PR #1142, opened 2021-06-02
Expose subscribeToNewsletter as a context

defunctzombie:
Allows for replacing within stories and tests via context rather than out of band mocks.
Remove the storybook mock and webpack alias that powered the mock.

---
Something with the mock was not working in https://github.com/foxglove/studio/pull/1017 so we opt for this approach to unblock 1017.


app/panels/WelcomePanel/SubscribeContext.ts
```diff
@@ -0,0 +1,19 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { createContext, useContext } from "react";
+
+import subscribeToNewsletter from "./subscribeToNewsletter";
+
+type SubscribeNewsletterFn = (email: string) => Promise<void> | void;
```
jtbandes, line 9: what's with `Promise<void> | void`?

defunctzombie: storybook `action("Foo")` returns a function that returns void.

PR #1154, opened 2021-06-03
Panel Extension API

defunctzombie:
Extension API

All extensions have an entry point source file. This file must export an activate function which accepts an extension context. Here you register the components of your extension (i.e. panels).

An example entry point file that registers one sample panel called "my-panel" and one called "my-react-panel". Both panels produce the same output.

index.ts
```typescript
import { ExtensionContext, PanelExtensionContext, RenderState, Topic, MessageEvent } from "@foxglove/studio";
import { initPanel } from "./MyPanel.ts";
import { initReactPanel } from "./MyReactPanel.tsx";

export function activate(extensionContext: ExtensionContext) {
  extensionContext.registerPanel({ name: "my-panel", initPanel: initPanel });
  extensionContext.registerPanel({ name: "my-react-panel", initPanel: initReactPanel });
}
```

Your panel entry point receives a `PanelExtensionContext` argument. This argument contains the root element for the panel and various methods to access data and interface with studio. Panels are framework agnostic. Here is an example using DOM primitives

MyPanel.ts
```typescript

export function initPanel(context: PanelExtensionContext) {
  const topicsListDiv = document.createElement("div");
  const messageCountDiv = document.createElement("div");

  context.panelElement.appendChild(topicsListDiv);
  context.panelElement.appendChild(messageCountDiv);

  // The render handler is run by the broader studio system during playback when your panel
  // needs to render because the fields it is watching have changed. How you handle rendering depends on your framework.
  // You can only setup one render handler - usually early on in setting up your panel.
  //
  // Without a render handler your panel will never receive updates.
  //
  // The render handler could be invoked as often as 60hz during playback if fields are changing often.
  context.onRender = (renderState: RenderState, done) => {
    topicsListDiv.innerText =  renderState?.topics.join(",");
    messageCountDiv.innerText = renderState?.currentFrame.length;

    // render functions receive a _done_ callback. You MUST call this callback to indicate your panel has finished rendering.
    // Your panel will not receive another render callback until _done_ is called from a prior render. If your panel is not done
    // rendering before the next render call, studio shows a notification to the user that your panel is delayed.
    done();
  };

  // After adding a render handler, you must indicate which fields from RenderState will trigger updates.
  // If you do not watch any fields then your panel will never render since the panel context will assume you do not want any updates.

  // tell the panel context that we care about any update to the _topic_ field of RenderState
  context.watch("topics");

  // tell the panel context we want messages for the current frame for topics we've subscribed to
  // This corresponds to the _currentFrame_ field of render state.
  context.watch("currentFrame");

  // subscribe to some topics, you could do this within other effects, based on input fields, etc
  // Once you subscribe to topics, currentFrame will contain message events from those topics (assuming there are messages).
  context.subscribe(["/some/topic"]);
}
```

Here we show an example using react. This example panel shows how to get updates for topics and messages on specific topics.

MyReactPanel.tsx
```typescript

function SamplePanel({ context }: { context: PanelExtensionContext}) {

  const [topics, setTopics] = useState<Topic[] | undefined>();
  const [messages, setMessages] = useState<MessageEvent<unknown>[] | undefined>();

  const doneRenderRef = useRef<(() => void) | undefined>();

  // We use a layout effect to setup render handling for our panel. We also setup some topic subscriptions.
  useLayoutEffect(() => {
    // The render handler is run by the broader studio system during playback when your panel
    // needs to render because the fields it is watching have changed. How you handle rendering depends on your framework.
    // You can only setup one render handler - usually early on in setting up your panel.
    //
    // Without a render handler your panel will never receive updates.
    //
    // The render handler could be invoked as often as 60hz during playback if fields are changing often.
    context.onRender = (renderState: RenderState, done) => {
       // render functions receive a _done_ callback. You MUST call this callback to indicate your panel has finished rendering.
       // Your panel will not receive another render callback until _done_ is called from a prior render. If your panel is not done
       // rendering before the next render call, studio shows a notification to the user that your panel is delayed.
       doneRenderRef.current = done;     
 
      // We may have new topics - since we are also watching for messages in the current frame, topics may not have changed
      // It is up to you to determine the correct action when state has not changed.
      setTopics(renderState.topics);
      
      // currentFrame has messages on subscribed topics since the last render call
      setMessages(renderState.currentFrame);
    };

    // After adding a render handler, you must indicate which fields from RenderState will trigger updates.
    // If you do not watch any fields then your panel will never render since the panel context will assume you do not want any updates.

    // tell the panel context that we care about any update to the _topic_ field of RenderState
    context.watch("topics");

    // tell the panel context we want messages for the current frame for topics we've subscribed to
    // This corresponds to the _currentFrame_ field of render state.
    context.watch("currentFrame");

    // subscribe to some topics, you could do this within other effects, based on input fields, etc
    // Once you subscribe to topics, currentFrame will contain message events from those topics (assuming there are messages).
    context.subscribe(["/some/topic"]);
  }, []);

  doneRenderRef.current?.();

  return (
    <>
      <div>{topics?.join(",")}</div>
      <div>{messages?.length}</div>
    </>
  );
}

export function initReactPanel(context: PanelExtensionContext) {
  ReactDOM.render(<SamplePanel context={context} />, context.panelElement);
}
```

### DIFF
--- a/packages/studio-base/src/PanelAPI/useConfig.ts
+++ b/packages/studio-base/src/PanelAPI/useConfig.ts
@@ -4,7 +4,6 @@
 
 import { useCallback, useMemo } from "react";
 
-import { panel } from "@foxglove/studio";
 import {
   useCurrentLayoutActions,
   useCurrentLayoutSelector,
@@ -15,9 +14,10 @@ import { SaveConfig } from "@foxglove/studio-base/types/panels";
 import { getPanelTypeFromId } from "@foxglove/studio-base/util/layout";
 
 /**
- * Mix partial panel config from savedProps with the panel type's `defaultConfig` to form the complete panel configuration.
+ * Load/Save panel configuration. This behaves in a manner similar to React.useState except the state
+ * is persisted with the current layout.
  */
-export const useConfig: typeof panel.useConfig = () => {
+export function useConfig<Config>(): [Config, (config: Partial<Config>) => void] {
   const panelId = usePanelId();
   const panelCatalog = usePanelCatalog();
   const panelComponent = useMemo(
@@ -31,7 +31,7 @@ export const useConfig: typeof panel.useConfig = () => {
     throw new Error(`Attempt to useConfig() with unknown panel id ${panelId}`);
   }
   return useConfigById(panelId, panelComponent?.defaultConfig);
-};
+}
 
 /**
  * Like `useConfig`, but for a specific panel id. This generally shouldn't be used by panels
--- a/packages/studio-base/src/PanelAPI/useDataSourceInfo.ts
+++ b/packages/studio-base/src/PanelAPI/useDataSourceInfo.ts
@@ -13,7 +13,7 @@
 
 import { useMemo } from "react";
 
-import { DataSourceInfo, panel } from "@foxglove/studio";
+import { DataSourceInfo } from "@foxglove/studio";
 import {
   useMessagePipeline,
   MessagePipelineContext,
@@ -39,7 +39,13 @@ function selectPlayerId(ctx: MessagePipelineContext) {
   return ctx.playerState.playerId;
 }
 
-export const useDataSourceInfo: typeof panel.useDataSourceInfo = () => {
+/**
+ * Data source info" encapsulates **rarely-changing** metadata about the source from which
+ * Studio is loading data.
+ *
+ * A data source might be a local file, a remote file, or a streaming source.
+ */
+export function useDataSourceInfo(): DataSourceInfo {
   const datatypes = useMessagePipeline(selectDatatypes);
   const topics = useMessagePipeline(selectTopics);
   const startTime = useMessagePipeline(selectStartTime);
@@ -56,4 +62,4 @@ export const useDataSourceInfo: typeof panel.useDataSourceInfo = () => {
       playerId,
     };
   }, [capabilities, datatypes, playerId, startTime, topics]);
-};
+}
--- a/packages/studio-base/src/PanelAPI/useMessagesByTopic.ts
+++ b/packages/studio-base/src/PanelAPI/useMessagesByTopic.ts
@@ -14,7 +14,6 @@
 import { groupBy } from "lodash";
 import { useCallback } from "react";
 
-import { panel } from "@foxglove/studio";
 import useDeepMemo from "@foxglove/studio-base/hooks/useDeepMemo";
 import { MessageEvent } from "@foxglove/studio-base/players/types";
 import concatAndTruncate from "@foxglove/studio-base/util/concatAndTruncate";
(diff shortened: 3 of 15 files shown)


packages/studio-base/src/components/PanelExtension/ExtendedMessageStore.ts
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  consume(blocks: BlockCache["blocks"]): void {
+    const events = [];
+    for (const block of blocks) {
+      if (!block) {
+        continue;
+      }
+
+      for (const messageEvents of Object.values(block.messagesByTopic)) {
```
jtbandes, line 47: better to check the topic from messagesByTopic rather than checking it on each message event?

packages/studio-base/src/components/PanelExtension/PanelExtensionAdapter.tsx
```diff
@@ -0,0 +1,169 @@
...
+          setSubscriptions(detail.id, []);
+          continue;
+        }
+        store.consume(playerState.activeData?.messages);
+      }
+    }
+
+    // see comment for prevBlocksRef on why extended message store updates are gated this way
+    const newBlocks = playerState.progress.messageCache?.blocks;
+    if (prevBlocksRef.current !== newBlocks) {
```
jtbandes, line 81: don't we also need this for `messages`?

packages/studio-base/src/components/PanelExtension/PanelExtensionAdapter.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+  const renderingRef = useRef<boolean>(false);
+
+  useMessagePipeline((ctx) => {
+    const playerState = ctx.playerState;
+    const prevState = previousPlayerStateRef.current;
+    previousPlayerStateRef.current = playerState;
+
+    if (playerState.activeData?.messages) {
+      for (const detail of latestMessagesStoreRef.current) {
```
jtbandes, line 68:
```suggestion
      for (const storeRef of latestMessagesStoreRef.current) {
```

packages/studio-base/src/components/PanelExtension/PanelExtensionAdapter.tsx
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    renderingRef.current = true;
+
+    // tell the panel to render and pause the frame until the render is complete
+    const resumeFrame = pauseFrame("fixme name here");
+    // what if there are no listeners for render?
+    eventEmitter.emit("render", () => {
+      renderingRef.current = false;
+      resumeFrame();
```
jtbandes, line 117: so I guess there can only be one render listener? do we have a way to enforce that?

packages/studio-base/src/components/PanelExtension/PanelExtensionAdapter.tsx
```diff
@@ -0,0 +1,169 @@
...
+    if (eventEmitter.listenerCount("render") === 0) {
+      return;
+    }
+
+    renderingRef.current = true;
+
+    // tell the panel to render and pause the frame until the render is complete
+    const resumeFrame = pauseFrame("fixme name here");
+    // what if there are no listeners for render?
+    eventEmitter.emit("render", () => {
```
jtbandes, line 115: pretty sketchy that this is happening inside a useMessagePipeline selector...should it be in an effect instead so it's more synced with react actually updating the dom?

defunctzombie: Yea - I think it should be placed somewhere that better aligns with the actual rendering and availability of new data. Those might not be the same thing tho.

defunctzombie:
Other things that will need apis:

Setting the config schema?
```
context.setConfigSchema({...});
```

Getting config updates (when set via the settings panel)
```
context.onConfig = () => {}
```

Getting notified when a seek happened (user jumped around rather than continuous playback)
```
config.onSeek = () => {}
```

Right now the renderState is about playback state - but there are events that happen outside of playback - i.e. config (panel settings) update, parameters changed, topics changed. Some of these (parameters, topics) come from players and might make sense to keep inline with render updates - or maybe it makes sense to separate those out as well.

PR #1160, opened 2021-06-03
Fix some no-explicit-any

jtbandes:


packages/studio-base/src/components/GlobalVariablesTable/index.tsx
```diff
@@ -311,7 +313,7 @@ function GlobalVariablesTable(): ReactElement {
           {unlinked.map((name, idx) => (
             <SAnimatedRow
               key={`unlinked-${idx}`}
-              skipAnimation={skipAnimation}
+              skipAnimation={skipAnimation.current}
```
jhurliman, line 316: :-o

jtbandes: I know right?

PR #1161, opened 2021-06-04
Fix no-explicit-any in studio-base/util

jtbandes:


packages/studio-base/src/util/CachedFilelike.test.ts
```diff
@@ -30,7 +30,7 @@ class InMemoryFileReader implements FileReader {
 
   fetch(offset: number, length: number): FileStream {
     return {
-      on: (type, callback) => {
+      on: (type: "data" | "error", callback: ((_: Buffer) => void) & ((_: Error) => void)) => {
```
jtbandes, line 33: This is technically wrong (should be | instead of & on the callback type), but TS accepts it, and I don’t know how to write it more correctly while still being accepted, so… :shipit:

PR #1175, opened 2021-06-07
Extensions sidebar

jhurliman:
Adds browsable lists of installed extensions and extensions that are available to be installed. The UI includes buttons for installing and uninstalling extensions that will need to be hooked up in a future PR.

extensions-list

extensions-details


packages/studio-base/src/components/Extensions.tsx
```diff
@@ -0,0 +1,117 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
defunctzombie, line 1: Name it ExtensionsSidebar maybe? Extensions is a rather generic name

jtbandes:
The `paddingLeft` change caused this issue with the `noPadding` setting: 
image

PR #1179, opened 2021-06-08
Change PanelToolbar to display when in edit layout mode.

defunctzombie:
Rather than have each panel add a panel toolbar, display the panel toolbar controls when editing a layout (adding/removing panels). This leaves the entire panel area for panel specific ui rather than earmarking a portion for our controls.


jtbandes:
Any plans to pursue this?

defunctzombie:
Not sure about this specific PR but the overall idea I think is still better than our current structure. I'm not a fan of the hover/floating toolbar within panels. It does provide a quick way to split a panel (without having to discover random shortcut keys). I suppose my main desire for dropping it is to give back that space to panels.

PR #1180, opened 2021-06-08
Fix more no-explicit-any

jtbandes:


packages/studio-base/src/dataProviders/types.ts
```diff
@@ -108,6 +108,7 @@ export interface MessageReader {
 export interface DataProviderConstructor {
   new (
     // The arguments to this particular DataProvider; typically an object.
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
     args: any,
```
defunctzombie, line 112: Could this be `args: unknown` ?

packages/studio-base/src/dataProviders/types.ts
```diff
@@ -197,6 +198,7 @@ export type Connection = {
 // means that you can describe a chain of data providers that includes a Worker or a WebSocket.
 export type DataProviderDescriptor = {
   name: string;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   args: any;
```
defunctzombie, line 202: Could this be `args: unknown` ?

jtbandes:
I tried these as unknown – they would eventually need to be cast upon usage because they are expected to be different things in different places. Also it breaks all the calls to createGetDataProvider because the various constructors expect specific arg types so they can't satisfy a `new (_: unknown ...)` requirement. I'm not sure the generics system is good enough to express what is actually going on and it doesn't seem worth it to try.

PR #1181, opened 2021-06-08
Show/hide panel toolbar using css instead of react state

defunctzombie:
CSS lets us show and hide child elements on hover. This is more reliable
than onMouseEnter/leave events which may not updated when the panel splits
or when child components capture the mouse events.

### DIFF
--- a/packages/studio-base/src/components/Panel.module.scss
+++ b/packages/studio-base/src/components/Panel.module.scss
@@ -6,6 +6,17 @@
   background-color: $dark;
   position: relative;
 
+  // To use css to hide/show toolbars on hover we use a global panelToolbar class
+  // because the PanelToolbar component is currently added within each panels render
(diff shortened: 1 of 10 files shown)


packages/studio-base/src/components/Panel.module.scss
```diff
@@ -1,11 +1,22 @@
 @import "@foxglove/studio-base/styles/colors.module.scss";
 @import "@foxglove/studio-base/styles/variables.module.scss";
 
+// To use css to hide/show toolbars on hover we use a global panelToolbar class
+// because the PanelToolbar component is currently added within each panels render
+// function rather than handling by the panel HOC
+:global(.floatingPanelToolbar) {
```
amacneil, line 7: this can move down to line 15 right before `&:hover :global(.floatingPanelToolbar)`

PR #1185, opened 2021-06-09
Fix chart component resize behavior

defunctzombie:
* Call update after resize
* Ensure width and height are whole numbers to avoid constant resizing

In some cases we were seeing the plot panel go blank after resize, clicking, or other interactions. The actual cause appears to be internal to chartjs but changing the order of resize/update calls reliably fixed the issue.


amacneil (approved):
Fixes the issue for me of plot going completely blank.

I still have issues with the plot disappearing for several seconds after adding or removing a panel (but not when resizing panels). However I am pretty sure that is a separate issue.

PR #1187, opened 2021-06-09
Add Quick Look preview for bag files

jtbandes:
Integrate with the macOS Quick Look extension API to provide a preview summary for bag files. The preview shows up in Finder and the Dock, and can be invoked by other applications.

This PR contains a new webpack entrypoint in the desktop build for the preview page, and an electron-builder afterPack script to bundle the app extension (beyond just copying files, this requires a codesign step to properly sandbox the extension, which means the build step will only work on macOS). The native code implementing the app extension is at https://github.com/jtbandes/quicklookjs.

https://user-images.githubusercontent.com/14237/121414564-7b426b00-c902-11eb-8088-81f22a8712d9.mov

### DIFF
--- a/desktop/quicklook/formatByteSize.test.ts
+++ b/desktop/quicklook/formatByteSize.test.ts
@@ -1,0 +1,28 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import formatByteSize from "./formatByteSize";
+
+describe("formatByteSize", () => {
+  it.each([
+    [0, "0 B"],
+    [1023, "1023 B"],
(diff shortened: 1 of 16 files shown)


desktop/quicklook/formatByteSize.test.ts
```diff
@@ -0,0 +1,28 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import formatByteSize from "./formatByteSize";
+
+describe("formatByteSize", () => {
```
jtbandes, line 7: Currently all the tests under desktop/ are part of the "integration" tests and only run in CI... this isn't a blocking issue for my PR, but in the future it would be nice to be able to put unit tests under desktop & web that are run as part of `yarn test`.

PR #1190, opened 2021-06-10
Change lazy message toJSON to fully deserialize the message

### LINKED ISSUE #1183: LazyMessages can't be transferred to Node Playground worker
Node Playground is not usable with lazy messages because of the `rpc.send` which does not transfer the fields of a lazy message:

https://github.com/foxglove/studio/blob/17d782db07a67b70a4cb48313a93f1ec248b908f/packages/studio-base/src/players/UserNodePlayer/index.ts#L345

Calling `toJSON()` on the message is also not a sufficient workaround, because it's only a "shallow toJSON" in that array fields may themselves contain more LazyMessages.

The Webviz equivalent solution to this was BobjectRpc: https://github.com/cruise-automation/webviz/blob/6a4226bc2959444704d650d8c55cea4f4220c75c/packages/webviz-core/src/util/binaryObjects/BobjectRpc.js#L160

defunctzombie:
Rather than deserializing only the first set of fields, the toJSON
method on a lazy message now deserializes the entire message including
nested fields and arrays. This makes toJSON useful for places where
you may want to entirely expand the message at the expense of deserializing.

fixes #1183

--

Tested node playground as shown below

Screen Shot 2021-06-09 at 7 26 48 PM

### DIFF
--- a/packages/rosmsg-deser/src/LazyMessageReaderFactory.ts
+++ b/packages/rosmsg-deser/src/LazyMessageReaderFactory.ts
@@ -271,7 +271,17 @@ export default function buildReader(types: readonly RosMsgDefinition[]): Seriali
           }`);
       }
 
-      fields.push(`${field.name}: this.${field.name}`);
+      // If the field is a simply type we use accessor notiation, this is sufficient to deserialized
+      // Builtin typed arrays fall into this category as well as builtin values
+      if (field.isComplex !== true) {
+        fields.push(`${field.name}: this.${field.name}`);
+      } else if (field.isArray === true) {
+        // The field is an array of complex values, we map each one using toJSON
+        fields.push(`${field.name}: this.${field.name}.map((item) => item.toJSON())`);
+      } else {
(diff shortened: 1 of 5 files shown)


packages/rosmsg-deser/src/LazyMessageReaderFactory.ts
```diff
@@ -271,7 +271,23 @@ export default function buildReader(types: readonly RosMsgDefinition[]): Seriali
...
+        }
+      } else {
+        // not an array
+        // if the field is basic types then getter access is sufficient to turn into a js object
+        if (field.isComplex === true) {
+          fields.push(`${field.name}: this.${field.name}.toJSON()`);
+        } else {
+          fields.push(`${field.name}: this.${field.name}`);
+        }
+      }
```
jtbandes, line 290: You could simplify this by flipping the order of checks. `!isComplex` -> case 1, `isArray` -> case 2, `else` -> case 3

PR #1191, opened 2021-06-10
Change Variable Slider panel to use fluentui slider

### LINKED ISSUE #1177: Sparklines (in playback performance) are blurry on retina displays
image

This was a regression probably caused by #555 (cc @defunctzombie) which removed a reference to `window.devicePixelRatio`.

Ideally this could auto update if the window is moved between screens, rather than reading the devicePixelRatio only once.

defunctzombie:
This removes the custom GlobalVariableSlider and SliderWithTicks components
as they are now unused.

Fixes #1177

Screen Shot 2021-06-09 at 8 20 17 PM


jtbandes:
Using the Fluent slider is nice, but now you can't see at a glance what the slider range is... maybe make it a `ranged` slider like these examples? https://developer.microsoft.com/en-us/fluentui#/controls/web/slider  I think showing tick marks would be nice too but would require more code.

defunctzombie:
I've added labels under the slider for min/max values.

Screen Shot 2021-06-15 at 5 23 41 PM

PR #1193, opened 2021-06-10
Added PR template to facilitate writing release changelog

esthersweon:


.github/pull_request_template.md
```diff
@@ -0,0 +1,14 @@
...
+  - e.g. `Update 3D panel's markers correctly on seek (#1000)`
+- [ ] Link to the relevant Github issue
+  - e.g. `Resolves #1000`
+- [ ] Describe the changes made
+  - Tradeoffs considered
+  - Solution and implementation details
+  - User impact
+  - e.g. `Fixed issue X in the Y component to do Z. Users should now be able to seek to any time and see updated 3D panel markers.`
+
+**Note**: `changelog.md` should be cleared after a release.
```
amacneil, line 14: Clearing changelog after release seems like a bit of a weird development process, have you seen any other repos doing that? We could just keep draft notes in the actual release, or stop using release notes and only use changelog.md, or start using something like https://www.conventionalcommits.org/en/v1.0.0/

I don't find it too hard to manually write the release notes though, and find there isn't always 1:1 mapping between PRs and release note entries.

esthersweon: Edited directions so devs just add entries directly to the release notes (if necessary, as not all commits will require an entry), and changed `changelog.md` to `release_template.md`:

`Add user-facing description of changes in a new release draft (if applicable)`
`If a draft does not already exist [here](https://github.com/foxglove/studio/releases), create one using the template at `release_template.md`

Issue #849, opened 2021-05-03
JavaScript Math Suffixes for Message Path Syntax Only Work in Plot Panel


esthersweon:
JavaScript Math suffixes currently only work in the Plot, even though other panels (like Raw Messages) does not complain when a topic message path using a JavaScript Math suffix is used

These suffixes should probably work everywhere message path syntax is used.

<img width="730" alt="Screen Shot 2021-05-03 at 3 38 16 PM" src="https://user-images.githubusercontent.com/6993359/116942008-9cd67580-ac25-11eb-8b91-e34345e1acf8.png">

jtbandes:
http://wiki.ros.org/Names

https://github.com/ros/ros_comm/blob/f5fa3a168760d62e9693f10dcb9adfffc6132d22/tools/rosgraph/src/rosgraph/names.py#L216

esthersweon:
Addressed in https://github.com/foxglove/studio/pull/1194.

PR #1194, opened 2021-06-10
Fix – Raise error in MessagePathInput for unsupported math modifiers

esthersweon:
- Addresses https://github.com/foxglove/studio/issues/849 
- Should only be valid in Plot panel's y-axis values

Example:
Screen Shot 2021-06-10 at 1 17 13 AM

Added story:
Screen Shot 2021-06-10 at 10 24 41 AM

Story for using math modifiers in Plot panel already exists.


defunctzombie (approved):
Add a story testing this feature

PR #1196, opened 2021-06-10
Remove react-leaflet

### LINKED ISSUE #1105: Current position is hidden by the buffered positions in Map panel
When using a bag file, the buffering places blue markers on the map panel. When I press play it is not possible to know where the current position is since the map is already populated by blue markers. I recommend to place a marker of different color or size for the current position and small blue markers for the buffered positions.

foxglove

Overall displaying markers for the buffered portion of the bag file is very focusing. If I click on a few different locations in the bag file, the buffering process creates unlinked graphics in plot and map panels.

foxglove2

### LINKED ISSUE #1182: Satellite view for Map panel
Requested in https://github.com/foxglove/studio/discussions/401#discussioncomment-711512

defunctzombie:
react-leaflet uses a non-standard OSS license. Since we use only a small set of leaflet
features we move over to using leaflet directly.

Fixes #1182
Fixes #1105

### DIFF
--- a/packages/studio-base/src/panels/Map/MapPanel.tsx
+++ b/packages/studio-base/src/panels/Map/MapPanel.tsx
@@ -2,15 +2,15 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
-import { Map as LeafMap } from "leaflet";
+import { Map as LeafMap, TileLayer, Control, LatLngBounds } from "leaflet";
 import { useEffect, useLayoutEffect, useRef, useState } from "react";
-import { MapContainer, TileLayer } from "react-leaflet";
+import { useDebouncedCallback } from "use-debounce";
 
 import { PanelExtensionContext, MessageEvent } from "@foxglove/studio";
 import EmptyState from "@foxglove/studio-base/components/EmptyState";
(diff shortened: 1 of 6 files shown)


packages/studio-base/src/panels/Map/MapPanel.tsx
```diff
@@ -49,14 +53,44 @@ function MapPanel(props: MapPanelProps): JSX.Element {
...
+      return;
+    }
+
+    const tileLayer = new TileLayer("https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png", {
+      attribution: '&copy; <a href="http://osm.org/copyright">OpenStreetMap</a> contributors',
+      maxNativeZoom: 18,
+      maxZoom: 24,
+    });
+
+    const satelite = new TileLayer(
```
jtbandes, line 68: typo: satellite

jtbandes (commented):
If this PR added features (satellite view) it should be called out more explicitly in the PR description - as is it's not clear that this PR contains anything changelog-worthy

packages/studio-base/src/panels/Map/MapPanel.tsx
defunctzombie: Will make a PR for seek preview (hover value) and fix that there. Spelling is hard for me.

jtbandes: There's a VS Code Extension For That™

PR #1197, opened 2021-06-10
internal: Add RemoteLayoutStorage and LayoutStorage interfaces

jtbandes:
The existing LayoutStorage is renamed to LocalLayoutStorage. It's still the only storage interface being used in the app right now, but will be combined in a future PR with remote storage to form a local cache implementation.

The new RemoteLayoutStorage interface demonstrates what operations a layout storage server with sharing features will be expected to implement. This may change a little in the near future, but has largely solidified after discussions and prototyping work on the server side.

The new LayoutStorage interface is what will power the forthcoming layout sidebar UI. It's a simplified version of RemoteLayoutStorage that doesn't require the caller to know about cached metadata for conflict resolution (a concrete "cached remote storage" implementation will automatically populate the ifUnmodifiedSince field by reading the locally cached metadata). Rather than returning errors from the method calls, conflict info will probably be available via a separate event emitter channel.

### User impact

None yet! 🔜

### DIFF
--- a/.vscode/snippets.code-snippets
+++ b/.vscode/snippets.code-snippets
@@ -25,5 +25,11 @@
     "prefix": "const",
     "body": "const {$2} = $1;",
     "description": "Destructure object"
+  },
+
+  "React component": {
+    "scope": "typescriptreact",
+    "prefix": "component",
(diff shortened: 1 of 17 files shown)


defunctzombie:
When reviewing https://github.com/foxglove/studio/pull/1206 I realized that I probably should have taken a closer look at this especially the proposed REST api surface. Would like to discuss this and 1206 again - My feeling looking at these is that we've made the mvp more involved than it needs to be.

PR #1199, opened 2021-06-11
Fix no-explicit-any in RawMessages

jtbandes:


packages/studio-base/src/panels/RawMessages/index.tsx
```diff
@@ -208,7 +208,8 @@ function RawMessages(props: Props) {
       ...keyPath: (number | string)[]
     ) => (
       <ReactHoverObserver className={styles.iconWrapper}>
-        {({ isHovering }: any) => {
+        {({ isHovering }) => {
+          // We make sure to always pass in a number at the end, but that's hard to express in the type system.
```
jtbandes, line 212: ported comment from https://github.com/cruise-automation/webviz/blob/6a4226bc2959444704d650d8c55cea4f4220c75c/packages/webviz-core/src/panels/RawMessages/index.js

jtbandes (commented):
@esthersweon do you have some context on these remaining errors? `valueRenderer` wants an `unknown[]` but the type of `diff` and `data` are both `unknown` here. Do we actually know these are arrays somehow?

cc'ing you for review since a lot of this touches diffing code which I believe you worked on.

packages/studio-base/src/panels/RawMessages/index.tsx
esthersweon: Maybe this deserves a refactor, but `lastKeyPath` will actually be a string in diff mode.

packages/studio-base/src/panels/RawMessages/index.tsx
```diff
@@ -201,20 +202,20 @@ function RawMessages(props: Props) {
...
         {({ isHovering }) => {
-          // We make sure to always pass in a number at the end, but that's hard to express in the type system.
-          const lastKeyPath = last(keyPath) as number;
+          // lastKeyPath is string in diff mode, number in regular mode
+          const lastKeyPath = last(keyPath) as number | string;
           let valueAction: ValueAction | undefined;
           if (isHovering && structureItem) {
             valueAction = getValueActionForValue(
-              data[lastKeyPath],
+              (data as unknown[])[lastKeyPath as number],
```
jtbandes, line 218: How do we know `data` isn't `DiffObject` here?

packages/studio-base/src/panels/RawMessages/index.tsx
esthersweon: Added clarifying comment in follow-up commit.

packages/studio-base/src/panels/RawMessages/index.tsx
esthersweon: Because of this line: https://github.com/foxglove/studio/pull/1199/commits/3725599912ae54dc2ebda8e3e98ccc4ee4cf20a6#diff-ca8a3a46914acc5156ff9c15ee6b85815754f4f4a02f2f2cd0e6d2c718cd2caaR380

esthersweon: In the code path where `data` is `DiffObject`, `structureItem` is always `undefined`.

esthersweon: Per my earlier comment, I think it's worth refactoring so that there's another `valueRenderer` function for diff inputs.

jtbandes: I see...though that line is also casting, assuming `diff` (return value from `getDiff`) is a `DiffObject`...

Maybe you could at least include that explanation in the comments? Refactoring at some point to avoid `as` would be great 😁  Though I know it's not always possible to avoid `as`, since `unknown` is pretty hard to refine into another type (`typeof x === "object"` doesn't cut it according to TS). So I'm _somewhat_ comfortable with `as Record<string, unknown>` in those cases we have already determined it's an object. Still there are lots of weird edge cases:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof
```ts
// Objects
typeof {a: 1} === 'object';

// use Array.isArray or Object.prototype.toString.call
// to differentiate regular objects from arrays
typeof [1, 2, 4] === 'object';

typeof new Date() === 'object';
typeof /regex/ === 'object'; // See Regular expressions section for historical results

// The following are confusing, dangerous, and wasteful. Avoid them.
typeof new Boolean(true) === 'object';
typeof new Number(1) === 'object';
typeof new String('abc') === 'object';
```

esthersweon: Just refactored in a follow-up commit - made it a lot simpler. No more weird type issues / type casting.

The existing screenshots for the Raw Messages panel (diff and regular mode) should cover these changes. Will double-check nothing looks off, and then merge.

PR #1204, opened 2021-06-11
Install and uninstall extensions

jhurliman:
This PR adds support for drag and drop installation of `.foxe` extension files, installing from the extension sidebar, and uninstalling extensions.


packages/studio-base/src/index.ts
```diff
@@ -33,3 +33,4 @@ export { default as ExtensionLoaderContext } from "./context/ExtensionLoaderCont
 export type { ExtensionLoader, ExtensionDetail } from "./context/ExtensionLoaderContext";
 export { default as AuthContext, useAuth } from "./context/AuthContext";
 export type { Auth, CurrentUser } from "./context/AuthContext";
+export * from "./util/extensions";
```
defunctzombie, line 36: I'd prefer we didn't do this - this exposes these free functions into the api space that we kinda are saying is the public api. I see it as a platform specific responsibility to handle how to identify the extensions - web might store in localstorage which doesn't have the same restrictions as disks.

PR #1205, opened 2021-06-12
Enable chromatic --only-changed flag (turbosnap)

amacneil:
Screenshot tests are only run based on changed files. Changed files are detected based on webpack dependency tree.

No user facing change.

Docs: https://www.chromatic.com/docs/turbosnap

### DIFF
--- a/packages/studio-base/package.json
+++ b/packages/studio-base/package.json
@@ -19,8 +19,8 @@
     "src"
   ],
   "scripts": {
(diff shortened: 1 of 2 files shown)


defunctzombie:
I read through the how it works and 👍 us trying this change. One gotcha here is our global css which is imported once rather than in every file that uses it - I think it might be imported by preview in a way that it still might rerun stories. Its minor and we also want to get rid of it so not a blocker.

PR #1206, opened 2021-06-12
Add OfflineLayoutStorage with sync logic

jtbandes:
- Rename LocalLayoutStorage → ILayoutCache to be more clear that the locally stored layouts are acting as a cache.
  - This change makes it so our current layout menu is using the "layout cache" as its primary storage. That's a bit strange, but the layout menu code will 🔜 be replaced with a new sidebar UI using the LayoutStorage anyway.
- Rename LayoutStorage → ILayoutStorage & RemoteLayoutStorage → IRemoteLayoutStorage to reduce confusion with concrete implementations.
- Add OfflineLayoutStorage class which implements ILayoutStorage, by composing an ILayoutCache with an IRemoteLayoutStorage. Changes are written first to the cache (except for "share" operations which go directly to the server). When the `syncWithRemote()` method is called they will be synchronized with the server. (No polling logic for now.)
- Add `@foxglove/den/async` with some async utilities that are in incubation.

Things remaining to do in future PRs:
- add functions that can be used to resolve conflicts (e.g. explicitly overwriting remote changes with local, or vice versa)
- add polling logic to automatically update cached metadata
- implement the UI!

User impact: none yet


packages/studio-base/src/services/CachedRemoteLayoutStorage/index.ts
```diff
@@ -0,0 +1,423 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
jtbandes, line 1: don't miss this giant file that github hides by default...

packages/studio-base/src/services/CachedRemoteLayoutStorage/LazilyInitialized.ts
```diff
@@ -0,0 +1,26 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
defunctzombie, line 1: Move to a package? Having a package and test for this seems nice rather than burying it under a folder in services

packages/studio-base/src/services/CachedRemoteLayoutStorage/MockLocalLayoutStorage.ts
```diff
@@ -0,0 +1,29 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
```
defunctzombie, line 2: This file should be co-located with LocalLayoutStorage? Others can discover it there and re-use?

packages/studio-base/src/services/CachedRemoteLayoutStorage/MutexLocked.ts
```diff
@@ -0,0 +1,20 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { Mutex } from "async-mutex";
+
+/**
+ * A wrapper around an object that ensures all accesses are guarded by a mutex lock.
```
defunctzombie, line 8: How does this ensure that all accesses are guarded? It doesn't behave as a proxy object - so all access is not guarded. Another candidate for the packages folder and probably an example or test?

This doesn't look to save much typing over using async-mutex directly - adds indirection.

defunctzombie (commented):
Maybe I missed this in another PR - what is the REST api look like? I'd like to take a look at that. My initial thoughts after looking at this is that we've made things more complicated than they need to be but I want to talk through that since I've easily missed something.

In my mind a v1 pass at this would be something like
- do a GET /layouts to get all the layouts available to the user
- save those layouts locally (with their layout ids - uuids from the server)
- periodically GET /layouts again to update the local copies (in a future there is a push notification so we don't have to use a timer)
- When a layout is changed, attempt to persist it to cloud - if conflict - user can overwrite with remote version or force push their version (we present the user with buttons to do so)

packages/studio-base/src/services/CachedRemoteLayoutStorage/MutexLocked.ts
jtbandes: It doesn't save much typing at all. But it prevents accidental unlocked usage of the object. The important part is `private value: T` and that `body` is `(value: T) => Result`. A plain `Mutex` does not hold a value and its function does not take an argument.

Without MutexLocked:
```ts
class T {
  cache = new Cache();
  cacheMutex = new Mutex();

  foo() {
    this.cacheMutex.runExclusive(() => this.cache.doSomething());
  }

  bar() {
    this.cache.doSomething(); // oops, forgot to use mutex
  }
}
```

With MutexLocked:
```ts
class T {
  cache = new MutexLocked(new Cache());

  foo() {
    this.cache.runExclusive((cache) => cache.doSomething()); // 👍
  }

  bar() {
    this.cache.doSomething(); // compile error
    this.cache.value.doSomething(); // compile error
  }
}
```

jtbandes: Sure it does not completely prevent someone from using it wrong. For example:

```ts
// this is like abusing unique_ptr - MutexLocked should be given sole ownership of the value
const cache = new Cache();
const lockedCache = new MutexLocked(cache);
lockedCache.runExclusive((c) => c.doSomething());
cache.doSomething();
```

```ts
// this is like abusing Swift.withUnsafePointer
const cache = new MutexLocked(new Cache());
let oops;
await cache.runExclusive((c) => (oops = c));
oops.doSomething();
```

I consider these to be pretty obvious programmer error so I don't mind the fact that they are still possible when using MutexLocked.

packages/studio-base/src/services/CachedRemoteLayoutStorage/LazilyInitialized.ts
jtbandes: Thoughts on package name? I assume we will want to group some of these _**utilities**_ into a package rather than having a separate package for each potentially reusable piece of logic we ever write?

packages/studio-base/src/services/CachedRemoteLayoutStorage/MockLocalLayoutStorage.ts
jtbandes: Thanks, I think this was an oversight from refactoring "move into a new file"

jtbandes:
> In my mind a v1 pass at this would be something like
> 
> * do a GET /layouts to get all the layouts available to the user
> * save those layouts locally (with their layout ids - uuids from the server)
> * periodically GET /layouts again to update the local copies (in a future there is a push notification so we don't have to use a timer)
> * When a layout is changed, attempt to persist it to cloud - if conflict - user can overwrite with remote version or force push their version (we present the user with buttons to do so)

This sounds simple but has a couple limitations we wanted to address in v1. Happy to discuss further

1. We have decided the `GET /layouts` only returns metadata. While a single layout json is not huge, it's feasible that a user will have access to dozens or hundreds of layouts (including team shared layouts) and we don't want to be downloading all of them on each request. So currently we only download and cache metadata, until the user actually clicks on a layout to inform us they want to load it.
2. With offline mode being a critical feature, we don't want "attempt to persist it to cloud" to be the very first action we take when the user saves or creates a layout. So I designed the cache around the idea that most actions can be taken while offline (except the "sharing" related actions). The majority of the logic is in reconciling the local cache with whatever is on the server next time we are able to GET /layouts. We also tabulate conflicts so we can present the user with appropriate buttons to resolve them like you suggested (no UI in this PR yet though).

packages/studio-base/src/services/CachedRemoteLayoutStorage/MutexLocked.ts
defunctzombie: Nice - I see how this is handy.

Could go in `@foxglove/async` for code that works with async things - could even export the async-mutex from there so we define the api surface if we want. Another option `@foxglove/helpers` with folders for specific concepts within that? I am not a fan of such things but I do like that slightly (only slightly) more the "utils" folder.

PR #1209, opened 2021-06-12
Update @foxglove/studio to v0.11.0

defunctzombie:
Bump @foxglove/studio version to v0.11.0

- Remove unused type definitions for DataSourceInfo

This will be the first version we publish to npm so we can start authoring extensions

### DIFF
--- a/packages/studio/index.d.ts
+++ b/packages/studio/index.d.ts
@@ -3,42 +3,17 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
 declare module "@foxglove/studio" {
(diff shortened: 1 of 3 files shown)


packages/studio/index.d.ts
```diff
@@ -3,42 +3,17 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
 declare module "@foxglove/studio" {
-  import { Time } from "rosbag";
```
defunctzombie, line 6: I am not importing from rosbag because we want to avoid having rosbag be a dependency of @foxglove/studio

Issue #986, opened 2021-05-16
Delete global.scss


defunctzombie:
Rather than overriding some components via a global.scss file, move styles into the component implementation and leverage theme where possible. Global css has a way of leaking into too many elements.

- [ ] rc-color-picker: https://github.com/foxglove/studio/blob/main/app/styles/global.scss#L19
- [ ] button: https://github.com/foxglove/studio/blob/main/app/styles/global.scss#L19, https://github.com/foxglove/studio/blob/main/app/styles/global.scss#L184
- [ ] tables: https://github.com/foxglove/studio/blob/main/app/styles/global.scss#L217

Injecting a css _reset_ or normalize and fonts should be provided via components - see https://material-ui.com/components/css-baseline/ This allows end users to opt in to such resets in a react friendly manner.

amacneil:
Completed in https://github.com/foxglove/studio/pull/1546

PR #1221, opened 2021-06-14
Remove th,tr,td from global.scss

defunctzombie:
Global css rules can have unintended effects on innocent html elements. We
prefer to scope our css to individual components.

Related https://github.com/foxglove/studio/issues/986

### DIFF
--- a/packages/studio-base/src/styles/global.scss
+++ b/packages/studio-base/src/styles/global.scss
@@ -195,36 +195,3 @@ button {
     padding: 4px 8px;
   }
 }
-
-th {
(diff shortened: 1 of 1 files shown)


jtbandes:
This is a nice thought but I don't think we should just delete the rules without replacing them / redesigning the places that they were having a useful effect.

defunctzombie:
@jtbandes converted to draft PR - this was meant to flush out those locations to the extent we can with our stories.

defunctzombie:
Replace by https://github.com/foxglove/studio/pull/1546

PR #1223, opened 2021-06-14
Remove paragraph styling from global.scss

defunctzombie:
Global css rules can have unintended effects on innocent html elements. We
prefer to scope our css to individual components.

Related #986


defunctzombie:
Replace by https://github.com/foxglove/studio/pull/1546

PR #1234, opened 2021-06-15
Fix: Set apt source architecture to amd64,arm64

### LINKED ISSUE #1231: Bug: Apt repo is not configured with desired architecture
# Current Behavior

When running `apt update` on Ubuntu, I see the warning

```
N: Skipping acquire of configured file 'main/binary-i386/Packages' as repository 'http://apt.foxglove.dev/studio stable InRelease' doesn't support architecture 'i386'
```

# Expected Behavior

I should not see this warning

# Steps to Reproduce

Install Foxglove Studio on Ubuntu, then run `apt update`

# Possible Solution

We should correctly add `[arch=amd64]` or `[arch=arm64]` as appropriate when we create `/etc/apt/sources.list.d/foxglove-studio.list`

defunctzombie:
If the architecture list is unspecified for an apt source entry, apt tries to fetch all
architectures configured under APT::Architectures for the system.

On x86 this results in warnings like the one below:
```
N: Skipping acquire of configured file 'main/binary-i386/Packages' as repository 'http://apt.foxglove.dev/studio stable InRelease' doesn't support architecture 'i386'
```

To avoid triggering this warning, we explicitly specify the architectures that could be fetched from our repo as amd64 and arm64.

Fixes #1231

---
**Changes**:
- User will no longer see the warning when doing `apt update` (once the new source line is in place).


amacneil:
I'm not an apt genius, but is it ok specifying both or would it be better to dynamically add the correct arch?

defunctzombie:
> I'm not an apt genius, but is it ok specifying both or would it be better to dynamically add the correct arch?

It is ok specifying both - tested it on my arm install and it was ok. This just tells apt what release files to download - not what to install. vscode specifies both.

PR #1235, opened 2021-06-15
Internal: switch to playwright for integration testing

defunctzombie:
This change switches us from [unmaintained spectron](https://github.com/electron-userland/spectron/issues/896) to [playwright](https://playwright.dev/) for our desktop integration tests. Playwright is actively maintained and will also support cross browser integration testing.

This change re-enables our integration tests which were previously skipped since spectron/chrome-driver were not working.

This change adds desktop to the root level jest projects so `yarn test` will run desktop unit tests. Integration tests are still run via `yarn test:integration` and are not run during `yarn test`.


desktop/integration-test/startup.test.ts
```diff
@@ -0,0 +1,43 @@
...
+import Logger from "@foxglove/log";
+
+import { appPath } from "./build";
+
+const log = Logger.getLogger(__filename);
+
+describe("startup", () => {
+  it("should start the application", async () => {
+    // In node.js the electron import gives us the path to the electron binary
+    // Our type definitions don't realize this so cast the variable to a string
```
jtbandes, line 17: Is this something we configure in our build setup or it's just a default behavior of the electron package? Maybe we can fix it with `declare module` instead of casting?

desktop/integration-test/jest.config.json
```diff
@@ -0,0 +1,8 @@
+{
+  "globalSetup": "<rootDir>/build.ts",
```
jtbandes, line 2: 👍 thanks for separating these.

desktop/integration-test/startup.test.ts
defunctzombie: Its not part of the build setup so much as a "feature" of the electron package. When imported/required in node it has one return value but when bundled with webpack it has another. It seems their types package focuses on the bundling use case rather than the nodejs use case. If we make a declare module it might override in the bundle case and we don't want that.

PR #1242, opened 2021-06-16
Partially revert "Enable chromatic --only-changed flag (turbosnap)"

amacneil:
Reverts foxglove/studio#1205

Broken on main: https://github.com/foxglove/studio/runs/2834196484

I'm keeping the `cross-env NODE_OPTIONS='--max-old-space-size=6144'` memory increase for now since that seems likely needed in the future anyway (and was already included when we run `start-storybook`).

### DIFF
--- a/packages/studio-base/package.json
+++ b/packages/studio-base/package.json
@@ -19,8 +19,8 @@
     "src"
   ],
   "scripts": {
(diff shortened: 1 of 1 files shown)


amacneil:
It looks like this is fixed on main in https://github.com/foxglove/studio/commit/adf36a98a0f738e93ff99d2d337f719499a95689 so this PR is no longer needed.

PR #1245, opened 2021-06-16
Internal: rework init panel handling in PanelExtensionAdapter

defunctzombie:
Create a ref callback to receive the panel element reference and
initialize the panel. The ref callback depends on the panel extension context
and init panel. If the ref callback changes, we re-create the
panel element div and re-initialize the panel.


amacneil:
Please add "why" to PR description in the future

PR #1246, opened 2021-06-16
Add Layouts tab to sidebar

jtbandes:
**User-facing changes**
Your layouts are now found in the app's sidebar instead of the top bar, with a new look.

**Description**
Replaces the old LayoutMenu with a new LayoutBrowser in the sidebar. The same actions are supported as the layout menu, but with a refreshed UI that will allow us to move towards a world with more interesting layout organization/sharing flows.

https://user-images.githubusercontent.com/14237/122159387-f972a600-ce09-11eb-8939-fce7837e1547.mov


packages/studio-base/src/theme/index.ts
```diff
@@ -138,15 +151,16 @@ export default createTheme({
...
-    themeLighterAlt: "#c0b7fd",
-    themeLighter: "#c8c0fd",
-    themeLight: "#d0c9fd",
-    themeTertiary: "#d7d2fe",
-    themeSecondary: "#dfdafe",
-    themeDarkAlt: "#e7e3fe",
-    themeDark: "#efecfe",
-    themeDarker: "#f6f5ff",
+    ...themePaletteColors({ inverted: true }),
+    // themePrimary: "#b8aefd",
```
defunctzombie, line 155: meant to be commented out?

jtbandes: Thanks, forgot to remove… when I started using Spinner I realized that our “theme” set of colors is kinda backwards in terms of which ones should be bright/dark (since we have an inverted theme). So this change will result in a bunch of small changes (and we should evaluate the contrast) but hopefully will fit better with the overall theme.

jtbandes: #1252

PR #1259, opened 2021-06-17
Fix renaming and deleting nodes in Node Playground

jtbandes:
**User-Facing Changes**
Fixed issues with renaming and deleting Node Playground nodes.

**Description**
Deleting: a bug introduced in #1026. By using `omit` instead of an actual `undefined` value for the node id, I prevented the node from being deleted.

Renaming: a bug caused by binding the initial value of the `save` function inside `editorDidMount`. The cmd-S action wasn't getting updated when the useCallback returned a newly bound function. Thanks @defunctzombie for debugging help.

### DIFF
--- a/packages/studio-base/src/panels/NodePlayground/index.tsx
+++ b/packages/studio-base/src/panels/NodePlayground/index.tsx
@@ -14,7 +14,7 @@
 import { Stack } from "@fluentui/react";
 import ArrowLeftIcon from "@mdi/svg/svg/arrow-left.svg";
 import PlusIcon from "@mdi/svg/svg/plus.svg";
-import { Suspense, useEffect } from "react";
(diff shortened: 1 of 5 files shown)


packages/studio-base/src/panels/NodePlayground/index.tsx
```diff
@@ -14,8 +14,7 @@
 import { Stack } from "@fluentui/react";
 import ArrowLeftIcon from "@mdi/svg/svg/arrow-left.svg";
 import PlusIcon from "@mdi/svg/svg/plus.svg";
-import { omit } from "lodash";
-import { Suspense } from "react";
+import { Suspense, useEffect } from "react";
```
jtbandes, line 17:
```suggestion
import { Suspense } from "react";
```

Issue #1256, opened 2021-06-17
3D panel goes blank when adjusting layouts


amacneil:
**Description**

When making certain changes to the layout (including splitting panels), the 3d panel goes blank. For example:
- If you are paused, and split the 3d panel, it goes blank. Starting playback fixes it.
- If you are playing, and split the 3d panel, it goes blank. Stopping and starting playback does not fix it. Deselecting and selecting topics doesn't fix it. Playing with the tf selector doesn't fix it. The only way I've found to reliably fix it is by pausing playback, deleting an adjacent panel, then restarting playback.

- Version: 0.11.0
- OS: Linux

**Steps To Reproduce**

See video:
- Add a 3d panel and configure a valid topic
- Start playback
- Split the 3d panel

https://user-images.githubusercontent.com/637671/122312120-6cdaed00-cec8-11eb-9c08-c984bebd4d82.mp4

**Expected Behavior**
Panel keeps showing data

**Actual Behavior**
Panel goes blank

amacneil:
Here's another video showing how enabling/disabling topics doesn't fix the issue:

https://user-images.githubusercontent.com/637671/122331307-76288180-cee9-11eb-8184-d4b5343ed805.mp4

amacneil:
This layout is still broken when testing with cal_loop.bag:

https://gist.github.com/amacneil/616d1911844a19cc7f031d8c169ee9cc


https://user-images.githubusercontent.com/637671/122822966-9e223700-d293-11eb-8882-db294a0d7f33.mp4

Issue #1255, opened 2021-06-17
Refactor `withTransforms` to use Hooks


jtbandes:
Getting the TS types to agree between 3 nested HOCs is difficult:

https://github.com/foxglove/studio/blob/393e557c8dcc7f0f955fbaa96a03113d517e8a9a/packages/studio-base/src/panels/ThreeDimensionalViz/index.tsx#L236-L238

However, FrameCompatibility is already implemented with a `useFrame` hook. If we refactor `withTransforms` to be a hook, then we can simply remove these two HOC wrappers and pass the renderer directly to Panel().

Issue #739, opened 2021-04-28
Remove FrameCompatibilityDEPRECATED


amacneil:
While not strictly harmful, the FrameCompatibility layer around 3D panel is a highly non-standard use of useMessageReducer, and should in principle be unnecessary.

The 3D panel is one of the oldest panels in the app, and was initially written when panels received a "frame" of new messages on each render. The SceneBuilder was written to collect these incoming messages into a stateful representation of the current scene, taking into account things like marker `lifetime`s, and custom "decay time" settings. Then on each render of the 3D panel (including simply moving the camera angle—not just when new messages are received), the SceneBuilder would use its state to render the currently visible messages. This rendering logic has complexity around handling scenarios like rendering while playback is paused and new frames are not being received, rendering things differently based on per-topic settings, etc.

Later, `MessageHistory` and eventually `useMessageReducer` became the primary ways to consume messages from the player, but the 3D panel was too complicated to adopt this paradigm, so FrameCompatibility was introduced as a translation layer, keeping compatibility for the `frame` input that the 3D panel knew how to handle, even after frames were no longer consumed anywhere else in the app.

By changing how frames are consumed in 3D panel and passed to SceneBuilder, we should be able to remove FrameCompatibility. Conceptually, SceneBuilder is acting like the reducer for useMessageReducer — although in practice the change may not be so simple.

PR #1264, opened 2021-06-18
Fix: keep frame state through renders in 3d panel

defunctzombie:
**User-Facing Changes**
When splitting the 3d panel during pause or playback the 3d panel no longer goes black.

**Description**
The existing useFrame hook within FrameCompatibilityDEPRECATED would clear the frame
    content after being invoked. Since react may invoke render functions multiple times
    this results in situations where the frame is cleared during remounts or multiple
    renders for a single frame. This would manifest as a blank 3d panel in situations
    like splitting the panel or changing the layout.
    
This change removes FrameCompatibilityDEPRECATED and withTransforms and replaces
with useFrame and useTransforms. useFrame collects messages from the current player
render frame and unlike before keeps this frame data around regardless of react render
frames.

useTransforms accumulates transforms until a reset (seek or player change)
    
Fixes: #1256
Fixes: #739
Fixes: #1255

----

https://user-images.githubusercontent.com/84792/122623050-8bb7bb80-d04f-11eb-8962-8255e671316b.mov


packages/studio-base/src/panels/ThreeDimensionalViz/FrameCompatibility.tsx
```diff
@@ -11,83 +11,32 @@
...
-      return ++counter.current;
+    restore: useCallback(() => {
+      return undefined;
     }, []),
-    addMessages: React.useCallback((_, messages: readonly MessageEvent<unknown>[]) => {
+    addMessages: useCallback((_, messages: readonly MessageEvent<unknown>[]) => {
+      const frame: Frame = {};
       for (const message of messages) {
-        (frame.current[message.topic] = frame.current[message.topic] ?? []).push(message);
+        (frame[message.topic] = frame[message.topic] ?? []).push(message);
```
jtbandes, line 35:
```suggestion
        (frame[message.topic] ??= []).push(message);
```

packages/studio-base/src/panels/ThreeDimensionalViz/useFrame.test.tsx
```diff
@@ -0,0 +1,138 @@
...
+        topic: "/some/topic",
+        receiveTime: { sec: 100, nsec: 0 },
+        message: { index: 0 },
+      },
+    ]);
+
+    // next render has cleared = false since the frame is no longer a new frame
+    rerender();
+    expect(result.current.cleared).toEqual(false);
+    expect(result.current.frame["/some/topic"]).toEqual([
```
jtbandes, line 84: maybe we can assert the frame identity is the same?

comment is misleading – cleared doesn't just mean it's a new frame, but a new player

defunctzombie: Doesn't always mean a new player - also happens on seek.

packages/studio-base/src/panels/ThreeDimensionalViz/FrameCompatibility.tsx
```diff
@@ -1,93 +0,0 @@
...
-
-const useFrame = (
-  topics: string[],
-): {
-  cleared?: boolean;
-  frame: Frame;
-} => {
-  // NOTE(JP): This is a huge abuse of the `useMessageReducer` API. Never use `useMessageReducer`
-  // in this way yourself!! `restore` and `addMessage` should be pure functions and not have
-  // side effects!
```
jtbandes, line 29: We should keep this comment. As you learned using useMessageReducer in this way has huge caveats that no mortals should be dealing with.

packages/studio-base/src/panels/ThreeDimensionalViz/useFrame.ts
```diff
@@ -0,0 +1,68 @@
...
+
+import * as PanelAPI from "@foxglove/studio-base/PanelAPI";
+import { Frame, MessageEvent } from "@foxglove/studio-base/players/types";
+
+/**
+ * useFrame returns the latest frame of messages.
+ *
+ * This hook is stateful. It will trigger a re-render when there are new frames.
+ *
+ * @returns a cleared field indicating if the frame is a new frame and the frame itself
```
jtbandes, line 24: Cleared does not mean the frame is new – frames are usually new. Cleared means the caller needs to go remove their cached state from previous frames.

packages/studio-base/src/panels/ThreeDimensionalViz/useFrame.ts
```diff
@@ -0,0 +1,68 @@
...
+  const frame = PanelAPI.useMessageReducer({
+    topics,
+    restore: useCallback(() => {
+      clearedRef.current = true;
+      returnedLastValueRef.current = false;
+      return {};
+    }, []),
+    addMessages: useCallback((prev, messages: readonly MessageEvent<unknown>[]) => {
+      if (returnedLastValueRef.current) {
+        prev = {};
```
jtbandes, line 44: a little strange to mutate prev - maybe just do `const frame = returnedLastValueRef.current ? {} : {...prev};` and then mutate?

packages/studio-base/src/panels/ThreeDimensionalViz/useFrame.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  const response = { cleared: clearedRef.current, frame };
+
+  // next time we update we no longer have a new frame (cleared)
+  clearedRef.current = false;
+
+  // indicate we've returned the previous state and can start building a new frame
+  returnedLastValueRef.current = true;
+
+  return response;
```
jtbandes, line 65:
```suggestion
  return { cleared: clearedRef.current, frame };
```

packages/studio-base/src/panels/ThreeDimensionalViz/FrameCompatibility.tsx
defunctzombie: The new use is more in line with a more acceptable use of useMessageReducer - the main thing the new code has are these refs because we believe that useMessageReducer _could_ call addMessages multiple times even tho I'm not sure thats a reality.

packages/studio-base/src/panels/ThreeDimensionalViz/useFrame.ts
defunctzombie: _reset_ might be a better name

packages/studio-base/src/panels/ThreeDimensionalViz/useFrame.ts
defunctzombie: returnedLastValueRef is not a frame - its a flag indicating we've exposed the frame to a caller.

packages/studio-base/src/panels/ThreeDimensionalViz/useFrame.ts
defunctzombie: response is created before we reset clearedRef.current so this is not the behavior we want

packages/studio-base/src/panels/ThreeDimensionalViz/useFrame.test.tsx
jtbandes: yep

packages/studio-base/src/panels/ThreeDimensionalViz/FrameCompatibility.tsx
jtbandes: The contract of useMessageReducer does not really specify the order of calls to restore/addMessage. I don't want other panels or 3rd party extensions to be writing code this way ... if we give them a frame-based API (even useFrame) to work with that is fine.

packages/studio-base/src/panels/ThreeDimensionalViz/useFrame.ts
jtbandes: 🤦

jtbandes: hence the `finally` in the old version...

jtbandes (approved):
Good changes – I would still argue that this is a highly non-standard usage of useMessageReducer that we do not want to encourage. We are only using it here because we aren't willing to put in the time to rewrite 3D panel in a stateless style.

PR #1271, opened 2021-06-18
Temporarily disable Quick Look extension

jtbandes:
**User-Facing Changes**
None - we never shipped this feature yet

**Description**
In case #1270 and other solutions do not work, this disables the Quick Look extension to get our notarize step working again.


jtbandes:
#1270 looks to be a viable solution.

PR #1286, opened 2021-06-22
Fix: memoize the currentTime field within RandomAccessPlayer

defunctzombie:
**User-Facing Changes**
Reduced CPU load when loading bag files (especially large bag files) by reducing renders on panels which should not be rendering.

During initial loading, the player will constant emit new state updates as the entire bag is loaded into memory. These updates would case panels using currentTime but not using the entire message range to re-render. This re-rendering isn't necessarily slow but it does run code paths that don't need to run thus slowing down the overall experience.

**Description**
Downstream consumers of activeData rely on fields maintaining reference
stability for change detection and memoization. Prior to this change
currentTime was being assigned a new reference on every state update
which caused unnecessary renders and code paths to run even tho the actual
value (sec, nsec) for currentTime remained unchanged.

This change memoizes currentTime and only updates it when the underlying
time value is different.


jtbandes:
Change seems fine – unit tests?

PR #1288, opened 2021-06-23
Update useConfirm to use FluentUI

jtbandes:
**User-Facing Changes**
None (useConfirm is not used in `main`).


defunctzombie:
If useConfirm is unused should we remove it?

jtbandes:
Planning to use it for the "delete layout" operation.

2metres:
@jtbandes can you close this issue?

PR #1290, opened 2021-06-23
Change rosmsg-deser to rosmsg-serialization, add MessageWriter

jhurliman:
This brings the MessageWriter class from rosbag.js into the @foxglove/rosmsg-serialization package so @foxglove/ros1 no longer has a dependency on rosbag.js


packages/ros1/package.json
```diff
@@ -29,10 +29,11 @@
     "typescript": "4.3.2"
   },
   "dependencies": {
+    "@foxglove/rosmsg": "workspace:packages/rosmsg",
+    "@foxglove/rosmsg-serialization": "workspace:packages/rosmsg-serialization",
```
amacneil, line 33: @jhurliman  I think these should all just say `workspace:*` now - see https://github.com/foxglove/studio/pull/1269

PR #1292, opened 2021-06-23
Remove version property from non-public packages

amacneil:
**User-Facing Changes**
N/A

**Description**
My intention is that everything in the studio repo will be versioned and released together (e.g. studio-base, studio, etc). However, it doesn't make sense to claim version numbers for our "incubation" packages, because when we eventually break these out they will have their own version numbers.

For now, npm/yarn don't care whether or not there is a version string, and it makes the release process slightly more sane if we don't have version numbers listed in the first place (so there is no version number for the release script to bump).

### DIFF
--- a/packages/den/package.json
+++ b/packages/den/package.json
@@ -1,7 +1,6 @@
 {
   "name": "@foxglove/den",
   "description": "Incubation area for code that may be released under future @foxglove packages",
-  "version": "0.12.0-dev",
   "license": "MPL-2.0",
   "private": true,
   "repository": {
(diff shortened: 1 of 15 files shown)


jhurliman (approved):

defunctzombie:
:shipit:

defunctzombie:
Related - @amacneil we need to release @foxglove/studio when you make studio releases

amacneil:
Good point, thanks. Published now, will automate it for next release.

PR #1294, opened 2021-06-23
Fix: Display bigint (uint64 or int64) values in RawMessage

defunctzombie:
**User-Facing Changes**
Prior to this change, the RawMessage panel would display <BigInt>
when viewing a message with int64/uint64 fields which were de-serialized
into a bigint. This change updates the value display to special case
bigint values to display the actual value.

image


jtbandes:
Alternate approach I had started working on: https://github.com/foxglove/studio/commit/de357c1ec4587e74ee33a42af3e16213ec2e57a6#diff-bfbb105b69d59e7363af305913bdaf2afccc16333a04a27e11248c61b07a4bd3R241-R243

defunctzombie:
@jtbandes I've adopted your approach and also added stories for diff mode (consecutive messages) since there was a fix to make diff mode work.

PR #1297, opened 2021-06-23
Update RawMessage panel to allow filtering on any field

defunctzombie:
**User-Facing Changes**

Previously the RawMessage panel would limit when a user could "pivot"
a field to certain conditions. This change removes those restrictions
and allows a user to select any field within an array of objects as
the filter value.

This change renames the "pivot" tooltip to "filter" as that better conveys
what the action does.

Screen Shot 2021-06-23 at 3 30 40 PM
Screen Shot 2021-06-23 at 3 27 59 PM

### DIFF
--- a/packages/studio-base/src/panels/RawMessages/getValueActionForValue.ts
+++ b/packages/studio-base/src/panels/RawMessages/getValueActionForValue.ts
@@ -11,23 +11,17 @@
 //   found at http://www.apache.org/licenses/LICENSE-2.0
 //   You may not use this file except in compliance with the License.
 
-import { last } from "lodash";
 import memoizeWeak from "memoize-weak";
 
 import { MessagePathStructureItem } from "@foxglove/studio-base/components/MessagePathSyntax/constants";
 import { isTypicalFilterName } from "@foxglove/studio-base/components/MessagePathSyntax/isTypicalFilterName";
 
-export type ValueAction =
-  | {
-      type: "pivot";
-      pivotPath: string; // Path to filter on using the current value, if the current value is an id.
-    }
-  | {
-      type: "primitive";
-      singleSlicePath: string; // Path that will only return one value per unit of time (for line charts).
-      multiSlicePath: string; // Path that might return multiple values per unit of time (for scatter plots).
-      primitiveType: string; // The ROS primitive type that these paths point at.
-    };
+export type ValueAction = {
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/panels/RawMessages/getValueActionForValue.ts
```diff
@@ -11,23 +11,18 @@
...
-      pivotPath: string; // Path to filter on using the current value, if the current value is an id.
-    }
-  | {
-      type: "primitive";
-      singleSlicePath: string; // Path that will only return one value per unit of time (for line charts).
-      multiSlicePath: string; // Path that might return multiple values per unit of time (for scatter plots).
-      primitiveType: string; // The ROS primitive type that these paths point at.
-    };
+export type ValueAction = {
+  type: "primitive";
```
jtbandes, line 20: type can be removed

packages/studio-base/src/panels/RawMessages/getValueActionForValue.ts
```diff
@@ -11,23 +11,18 @@
 //   found at http://www.apache.org/licenses/LICENSE-2.0
 //   You may not use this file except in compliance with the License.
 
-import { last } from "lodash";
 import memoizeWeak from "memoize-weak";
 
 import { MessagePathStructureItem } from "@foxglove/studio-base/components/MessagePathSyntax/constants";
-import { isTypicalFilterName } from "@foxglove/studio-base/components/MessagePathSyntax/isTypicalFilterName";
+//import { isTypicalFilterName } from "@foxglove/studio-base/components/MessagePathSyntax/isTypicalFilterName";
```
jtbandes, line 17: delete?

packages/studio-base/src/panels/RawMessages/getValueActionForValue.ts
```diff
@@ -85,25 +85,8 @@ export function getValueActionForValue(
           ? structureItem.next
           : { structureType: "primitive", primitiveType: "json", datatype: "" };
       multiSlicePath = `${singleSlicePath}[:]`;
-      // Ideally show something like `/topic.object[:]{id=123}` for the singleSlicePath, but fall
-      // back to `/topic.object[10]` if necessary.
```
jtbandes, line 89: I know you like deleting stuff but I don't think we should delete this. Is it broken?

PR #1298, opened 2021-06-23
Update Node Playground to target ES2020

defunctzombie:
**User-Facing Changes**
Users can write node playground nodes using bigint values.

**Description**
Adds support for using bigint within node playground.


packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/typescript/lib.ts
```diff
@@ -21,28 +58,60 @@ const lib_logger = `
...
...see the Intl.Collator object for details.\n      */\n    localeCompare(that: string, locales?: string | string[], options?: Intl.CollatorOptions): number;\n}\n\ninterface Number {\n    /**\n      * Converts a number to a string by using the current or specified locale.\n      * @param locales A locale string or array of locale strings that contain one or more language or locale tags. If you include more than one locale string, list them in descending order of priority so that the first entry is the preferred locale. If you omit this parameter, the default locale of the JavaScript runtime is used.\n      * @param options An object that contains one or more properties that specify comparison options.\n      */\n    toLocaleString(locales?: string | string[], options?: Intl.NumberFormatOptions): string;\n}\n\ninterface Date {\n    /**\n      * Converts a date and time to a string by using the current or specified locale.\n      * @param locales A locale string or array of locale strings that contain one or more language or locale tags. If you include more than one locale string, list them in descending order of priority so that the first entry is the preferred locale. If you omit this
```
jtbandes, line 109: Or maybe just `""`? Does the `<reference>` need to stay in?

packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/typescript/ast.ts
```diff
@@ -407,6 +407,14 @@ export const constructDatatypes = (
         return getRosMsgField(name, tsNode.elementType, true, true, typeMap, innerDepth + 1);
       }
 
+      case ts.SyntaxKind.BigIntKeyword:
+        return {
+          name,
+          type: "int64",
```
jtbandes, line 413: any thoughts on whether it matters that we arbitrarily chose int64 instead of uint64? I guess it's possible for someone to produce a bigint outside of the int64 range in node playground... the rest of the app probably won't care right? We would need to care if we were serializing these back to ROS binary.

Maybe worth a comment here.

jtbandes (approved):

PR #1301, opened 2021-06-24
Upgrade dependencies

defunctzombie:
**User-Facing Changes**
No intended changes

**Description**


packages/studio-base/src/components/ChildToggle/index.tsx
```diff
@@ -254,4 +254,4 @@ export default function ChildToggle(props: Props): ReactElement {
   );
 }
 
-ChildToggle.ContainsOpen = ChildToggleContainsOpen;
```
defunctzombie, line 257: This was causing a weird compilation issue in storybook where it turned this into "default.ChildToggle.ContainsOpen" which the js parser in chrome did not like. I don't like this pattern of monkey-patching an element onto another element and prefer to use export functionality.

PR #1306, opened 2021-06-25
@foxglove/cdr

jhurliman:
A Common Data Representation serde library. This is a building block toward ROS2 support


jhurliman:
Released at <https://github.com/foxglove/cdr>

PR #1309, opened 2021-06-26
Migrate more components to LayoutStorage, remove id/name from PanelsState

jtbandes:
**User-Facing Changes**
Minor tweaks to layouts sidebar: click to rename, and show actions in right click menu

**Description**

- Migrate some components that were still using LayoutCache onto LayoutStorage in preparation for shared layout support.
- Remove `id` and `name` from PanelsState (layout data); hold id in CurrentLayout. The "current layout" is now allowed to be `{selectedLayout: ... | undefined}`.
- Make `name`/`path` optional in `updateLayout` API so that only id and data need to be held in CurrentLayoutState.


packages/studio-base/src/context/CurrentLayoutContext/actions.ts
```diff
@@ -24,12 +24,8 @@ import {
 } from "@foxglove/studio-base/types/panels";
 
 export type PanelsState = {
-  id: string;
-  name: string;
```
jtbandes, line 28: FYI @defunctzombie this is a departure from your rework of native layout storage – the PanelsState is now (again) purely the layout data and other info is now stored in a higher level in [LayoutMetadata](https://github.com/foxglove/studio/blob/b80afb9964dac2e329ecc2e9eed161807b827b12/packages/studio-base/src/services/ILayoutStorage.ts#L20) / [CachedLayout](https://github.com/foxglove/studio/blob/b80afb9964dac2e329ecc2e9eed161807b827b12/packages/studio-base/src/services/ILayoutCache.ts#L8).

packages/studio-base/src/services/ILayoutStorage.ts
```diff
@@ -35,13 +35,17 @@ export interface ILayoutStorage {
...
+  saveNewLayout(params: {
     path: string[];
     name: string;
     data: PanelsState;
+  }): Promise<LayoutMetadata>;
+
+  updateLayout(params: {
     targetID: LayoutID;
+    name: string | undefined;
+    path: string[] | undefined;
```
jtbandes, line 47: Some callers want to update the name and the path and some don't.

packages/studio-base/src/context/CurrentLayoutContext/actions.ts
defunctzombie: Seems good to me. the future

PR #1310, opened 2021-06-26
Filter out DocgenPlugin from storybook webpack config

jtbandes:
**User-Facing Changes**
None

**Description**
The storybook build is already slow and the DocgenPlugin makes it slower. Just remove it since we don't use storybook's autogenerated docs.


amacneil:
Would be nice to add before & after performance to PR description.

jtbandes:
Yeah – I considered `time yarn storybook:build` but the initial build is always slower – incremental build speed is part of what I wanted to speed up and is a little harder to time. I can try and do it for future improvements though.

PR #1313, opened 2021-06-28
Fix some more no-explicit-any in 3D panel

jtbandes:
**User-Facing Changes**
None


packages/studio-base/src/panels/ThreeDimensionalViz/TopicTree/useTopicTree.ts
```diff
@@ -437,18 +439,20 @@ export default function useTree({
...
       }
     }
     return result;
   }, [defaultTopicSettings, settingsByKey]);
 
-  const onNamespaceOverrideColorChange = useCallback(
-    (newColor: string | undefined, prefixedNamespaceKey: string) => {
-      const newSettingsByKey = isNonEmptyOrUndefined(newColor)
+  const onNamespaceOverrideColorChange: OnNamespaceOverrideColorChange = useCallback(
+    (newColor: Color | undefined, prefixedNamespaceKey: string) => {
```
jtbandes, line 454: I really don't know what was going on here – I think @wkalt maybe ran into this one as well...seems like the string annotation was just wrong?

PR #1314, opened 2021-06-28
Extract @foxglove/just-fetch

amacneil:
**User-Facing Changes**
N/A

**Description**
Moved `@foxglove/just-fetch` package to https://github.com/foxglove/just-fetch and published to NPM.


amacneil:
I could use a sanity check here that the way I have packaged this is correct (`main` and `browser` entrypoints in package.json, etc). 

I tested locally with turtlesim ROS1 connection and everything seems to be working - I can see fetch requests happening in devtools.

jhurliman (approved):

PR #1326, opened 2021-06-29
Fix more no-explicit-any

jtbandes:
**User-Facing Changes**
None


defunctzombie:
WOW the end is in sight

PR #1329, opened 2021-06-29
Force any browser window navigation outside the app into external browser

defunctzombie:
**User-Facing Changes**
None. All links that used an explicit _blank target will continue to open in external browser.

**Description**
Rather than allowing the browser window to navigate away from the
studio app, handle the will-navigate event and send to external browser.

This prevents anchor tags or other navigation from inadvertently replacing
the studio window contents with another web page.

### DIFF
--- a/packages/studio-base/src/components/TextContent.tsx
+++ b/packages/studio-base/src/components/TextContent.tsx
@@ -35,7 +35,6 @@ export default function TextContent(props: PropsWithChildren<Props>): React.Reac
       return (
         <a
           href={linkProps.href}
-          target="_blank"
           rel="noopener noreferrer"
           onClick={(event) => handleLink(event, linkProps.href ?? "")}
         >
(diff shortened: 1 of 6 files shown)


defunctzombie:
I discovered this while working on https://github.com/foxglove/studio/pull/1323 which had a url to an external page and the default behavior was to open that in the studio browser window. This change fixes that default behavior so external links open in external windows.

packages/studio-base/src/components/TextContent.tsx
```diff
@@ -35,7 +35,6 @@ export default function TextContent(props: PropsWithChildren<Props>): React.Reac
       return (
         <a
           href={linkProps.href}
-          target="_blank"
           rel="noopener noreferrer"
```
amacneil, line 38: @defunctzombie you can remove these now too, they are only relevant in a browser context.

amacneil:
What is the expected behavior for web builds?

defunctzombie:
> What is the expected behavior for web builds?

Oh damn - that slipped my mind 🤦 I suppose we need the _blank back for those. It is less of an issue in web builds because the user has a url bar and back button. In studio once the page is swapped there is no hope beyond close and re-open.

PR #1304, opened 2021-06-25
Switch to external @foxglove/rosmsg package

jhurliman:

PR #1331, opened 2021-06-29
Revert "Switch to external @foxglove/rosmsg package"

jtbandes:
Reverts foxglove/studio#1304

Certain bags are not properly handled because message definitions cannot be parsed. Example failing definition: `time time_ref`


defunctzombie:
@jhurliman made the fixes in @foxglove/rosmsg - just need to upgrade studio to the new release https://github.com/foxglove/rosmsg

PR #1336, opened 2021-06-30
Refactor HelpModal to use @fluentui/react Modal

### LINKED ISSUE #1335: Convert HelpModal to FluentUI

2metres:
**User-Facing Changes**
| Before | After |
| ------ | ----- |
| Screen Shot 2021-06-30 at 1 25 52 pm | Screen Shot 2021-07-01 at 8 46 24 am |

Closes #1335

### DIFF
--- a/packages/studio-base/src/components/HelpModal.tsx
+++ b/packages/studio-base/src/components/HelpModal.tsx
@@ -11,35 +11,63 @@
 //   found at http://www.apache.org/licenses/LICENSE-2.0
 //   You may not use this file except in compliance with the License.
 
+import { IconButton, Modal, makeStyles, useTheme } from "@fluentui/react";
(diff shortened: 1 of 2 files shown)


2metres:
@jtbandes Ideally I would have the title sit in the header but presently the help text is rendered from a block of Markdown.

Could we either extract the title from the Markdown or pass in a title prop?

jtbandes (approved):

defunctzombie:
LGTM - @amacneil are we ok with the license change when we make such changes?

amacneil:
> @amacneil are we ok with the license change when we make such changes?

Yes, if a file is rewritten completely (as is the case here) then I think it is accurate to treat it as we would a new file and include our license header only.

Issue #1211, opened 2021-06-13
Move electron-socket to separate repo under foxglove org


defunctzombie:
This is a cool package that we can maintain separately from the studio codebase. I think we've proven out the idea to move it into its own repo and publish to npm.

https://github.com/foxglove/studio/issues/732 would be good to tackle with that.

PR #1343, opened 2021-06-30
Extract @foxglove/electron-socket package

amacneil:
**User-Facing Changes**
None

**Description**
Extract `@foxglove/electron-socket` into a separate npm package.

Moved dependency into `desktop` and `studio-base` rather than root `devDependencies`.

Please also sanity check modifications in https://github.com/foxglove/electron-socket/pull/1/files

Closes https://github.com/foxglove/studio/issues/1211


defunctzombie:
👍

PR #1346, opened 2021-06-30
Fix remaining no-explicit-any

defunctzombie:
**User-Facing Changes**
None


defunctzombie:
@jtbandes I'm stuck on last bit here with the Panel HOC

jtbandes:
Right, that's the one I think it may not be worth fixing - we could just suppress the lint error.

jtbandes:
One option is maybe to move the `childId` and `tabId` into context so that we don't need to pass them as props at all. How does this work for extension panels?

PR #1354, opened 2021-07-01
Extract @foxglove/xmlrpc

amacneil:
**User-Facing Changes**
None

**Description**
Moved @foxglove/xmlrpc to https://github.com/foxglove/xmlrpc


packages/ros1-turtlesim-test/src/index.ts
```diff
@@ -12,7 +12,7 @@ import {
   getPid,
   TcpSocketNode,
 } from "@foxglove/ros1/src/nodejs";
-import { HttpServerNodejs } from "@foxglove/xmlrpc/src/HttpServerNodejs";
+import { HttpServerNodejs } from "@foxglove/xmlrpc/dist/HttpServerNodejs";
```
amacneil, line 15: @jhurliman what do you want to do about these? I don't recall why they aren't exported from the main index.js.

amacneil: I'm gonna merge as is for now, and we can always export this separately in the package later

PR #1355, opened 2021-07-02
Use fuzzy time parsing in PlaybackTimeDisplayMethod

jtbandes:
Moved from https://github.com/foxglove/rostime/pull/1

**User-Facing Changes**
Timestamps typed into the playback bar can now be in milliseconds, microseconds, or nanoseconds.

https://user-images.githubusercontent.com/14237/124209830-20360b00-da9f-11eb-9326-2c3053bcdfc6.mov

**Description**

This logic will parse `31556937.600001`, `31556937600001`, and `31556937600001000` to the same value: `{ sec: 31556937, nsec: 600001000 }`.

### DIFF
--- a/packages/studio-base/src/util/formatTime.ts
+++ b/packages/studio-base/src/util/formatTime.ts
@@ -15,7 +15,8 @@ import momentDurationFormatSetup from "moment-duration-format";
 import moment from "moment-timezone";
 import { Time } from "rosbag";
 
-import { toDate, fromDate, getRosTimeFromString } from "./time";
+import parseFuzzyRosTime from "./parseFuzzyRosTime";
+import { toDate, fromDate } from "./time";
 
 // All time functions that require `moment` should live in this file.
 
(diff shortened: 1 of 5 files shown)


defunctzombie:
👍 tho it seems to me the user should be encouraged to use decimals - what was the previous behavior? did it require decimals or did it support no decimal and try to guess milliseconds or microseconds?

jtbandes:
Previous behavior supported decimals (seconds) only. The customer observed that certain time values, like +1000 years after 1970, don't really make sense for robotics (I will be very sad if this code is still being used 1000 years from now). So it's an opportunity to make the ergonomics a little better for those who have milliseconds, microseconds etc.

Btw, JS Date.valueOf() will give you milliseconds by default, so this is more compatible with that as well...

jtbandes:
So this only supports decimal points for seconds – only applies the fuzzy parsing if there is no decimal point (it's a large integer). I suppose we could use the fuzzy parsing if there is a decimal point too.... 🤷

PR #1369, opened 2021-07-02
Enable and fix `@typescript-eslint/promise-function-async`

jtbandes:
**User-Facing Changes**
None

**Description**
Initial fixes were made automatically, with some manual cleanup


amacneil:
should we configure it upstream in https://github.com/foxglove/eslint-plugin?

jtbandes:
Yeah, just wanted to play with it first here. I might make a few of these PRs and then migrate them all upstream at once.

One thing to note is that this rule is a bit on the slower side... but I think it's still reasonably fast.

PR #1370, opened 2021-07-02
Enable @foxglove/strict-equality eslint rule

amacneil:
**User-Facing Changes**
None

**Description**
- Enable `@foxglove/strict-equality` and remove corresponding `no-restricted-syntax`
- Remove eslint tests which are covered by https://github.com/foxglove/eslint-plugin/pull/21
- Replace `=== undefined` with `== undefined` everywhere (to check both `null` & `undefined`)

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -27,7 +27,6 @@ settings:
 
 rules:
   "@foxglove/license-header": error
-  "@foxglove/strict-equality": off # will replace some no-restricted-syntax in the future
 
   # show progress while linting
   file-progress/activate: warn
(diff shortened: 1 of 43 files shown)


eslint-test.ts
```diff
@@ -37,19 +29,22 @@ void (async () => {
...
+  wut === null; // eslint-disable-line no-restricted-syntax, @foxglove/strict-equality
 
   wut == undefined;
   wut != undefined;
-  undefined == wut; // eslint-disable-line no-restricted-syntax
-  undefined != wut; // eslint-disable-line no-restricted-syntax
-  undefined === wut;
-  wut === undefined;
+  undefined == wut;
+  undefined != wut;
```
jtbandes, line 38: do we need `yoda: never`?

.eslintrc.yaml
```diff
@@ -72,20 +71,11 @@ rules:
...
     - selector: "MethodDefinition[kind='set'], Property[kind='set']"
       message: "Property setters are not allowed; prefer function syntax instead."
+
     # We disable console methods here rather than using no-console so that it doesn't prohibit overrides such as "console.info = ..."
     - selector: "CallExpression[callee.object.name='console'][callee.property.name!=/^(warn|error|debug|assert)$/]"
       message: "Unexpected property on console object was called"
 
-    # Ban `null [!=]= x` and `x [!=]= null`
-    - selector: "BinaryExpression:matches([operator='=='], [operator='!=']):matches([left.type=Literal][left.raw=null], [right.type=Literal][right.raw=null])"
-      message: 'Prefer "x == undefined" or "x != undefined" to check for both null and undefined.'
```
amacneil, line 81: This error message implied that one should always use double equals undefined, but it never actually prevented triple equals undefined.

eslint-test.ts
amacneil: checking what? that we triple equals it?

PR #1374, opened 2021-07-02
Fix: use id for MosaicWindow key

defunctzombie:
**User-Facing Changes**
When a user splits a panel that has state but doesn't persist that state to the layout (i.e. the user has expanded fields in the RawMessage panel tree), the existing panel is kept mounted and the user's modifications are kept.

**Description**
When changing the layout, the mosaic path for a panel changes. The previous behavior used path as the key for mosaicWindow which results in re-mounting the panel (and losing any un-persisted state). The new behavior uses the id as the key which avoids re-mounting the panel.


jtbandes (approved):
🎉

No idea why it wasn't always this way. Hopefully we will not have to find out the hard way.

PR #1375, opened 2021-07-03
Enable and fix no-unused-expressions

jtbandes:
**User-Facing Changes**
None

### DIFF
--- a/packages/studio-base/src/components/AccountSettingsSidebar/SigninForm.tsx
+++ b/packages/studio-base/src/components/AccountSettingsSidebar/SigninForm.tsx
@@ -84,7 +84,6 @@ export default function SigninForm(): JSX.Element {
     if (!deviceResponse) {
       return;
     }
-    deviceResponse?.id_token;
   }, [deviceResponse]);
 
   const [{ value: session }, onOrgSelect] = useAsyncFn(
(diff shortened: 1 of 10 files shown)


defunctzombie:
nice!

packages/studio-base/src/components/MessagePathSyntax/useCachedGetMessagePathDataItems.ts
```diff
@@ -177,7 +177,6 @@ export function fillInGlobalVariablesInPath(
         return { ...messagePathPart, value };
       }
 
-      messagePathPart.type;
```
jtbandes, line 180: This was not random – in Flow it used to be a type assertion to validate the type refinement after the above conditional. Probably was removed at some point automatically by flow-to-ts or eslint.

packages/studio-base/src/components/AccountSettingsSidebar/SigninForm.tsx
```diff
@@ -84,7 +84,6 @@ export default function SigninForm(): JSX.Element {
     if (!deviceResponse) {
       return;
     }
-    deviceResponse?.id_token;
```
jtbandes, line 87: cc @defunctzombie - anything we should do here? actually I probably should've deleted the whole useEffect huh?

defunctzombie: yea - this must have been some code I left accidentally - doesn't seem used in this effect

jtbandes: But the effect also doesn’t do anything at all

PR #1389, opened 2021-07-06
Refactor Topic settings modal to use fluentui dialog

### LINKED ISSUE #1337: Convert 3d panel topic settings modal to FluentUI
Would likely use a [Contextual Menu](https://developer.microsoft.com/en-us/fluentui#/controls/web/contextualmenu) with a [Grouped List](https://developer.microsoft.com/en-us/fluentui#/controls/web/groupedlist) inside it

2metres:
I've added a Done button as the primary action, my monkey brain really wanted to push that reset changes button

|before|after|
|---|---|
|Screen Shot 2021-07-07 at 9 43 54 am|Screen Shot 2021-07-07 at 9 41 14 am|

Closes #1337

### DIFF
--- a/packages/studio-base/src/components/ShareJsonModal.tsx
+++ b/packages/studio-base/src/components/ShareJsonModal.tsx
@@ -63,9 +63,9 @@ export default function ShareJsonModal({
     <Dialog
       hidden={false}
       onDismiss={onRequestClose}
-      title={title}
       modalProps={{ isBlocking: true }}
       dialogContentProps={{
+        title,
         subText: `Paste a new ${noun} to use it, or copy this one to share it:`,
       }}
       maxWidth={`calc(100vw - ${theme.spacing.l2})`}
--- a/packages/studio-base/src/panels/ThreeDimensionalViz/TopicTree/TopicSettingsModal.tsx
+++ b/packages/studio-base/src/panels/ThreeDimensionalViz/TopicTree/TopicSettingsModal.tsx
@@ -64,7 +64,7 @@ function MainEditor({
         <DialogFooter>
           <DefaultButton onClick={() => onSettingsChange({})}>Reset to defaults</DefaultButton>
           <DefaultButton primary onClick={() => setCurrentEditingTopic(undefined)}>
-            Apply
+            Done
           </DefaultButton>
         </DialogFooter>
       </div>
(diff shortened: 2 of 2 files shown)


defunctzombie:
Pictures before/after?

packages/studio-base/src/components/ShareJsonModal.tsx
```diff
@@ -63,9 +63,9 @@ export default function ShareJsonModal({
     <Dialog
       hidden={false}
       onDismiss={onRequestClose}
-      title={title}
       modalProps={{ isBlocking: true }}
       dialogContentProps={{
+        title,
```
jtbandes, line 68: What's the difference between title and dialogContentProps.title?

2metres: Apparently `<Dialog title={} />` is [deprecated](https://github.com/microsoft/fluentui/blob/a712d19f8b73ea33406cfa5c51bb3e338599087a/packages/react/src/components/Dialog/Dialog.types.ts#L120-L125) and they recommend using `dialogContentProps.title instead`.

2metres:
> Pictures before/after?

@defunctzombie I've added some to the first comment

packages/studio-base/src/panels/ThreeDimensionalViz/TopicTree/TopicSettingsModal.tsx
```diff
@@ -97,23 +54,20 @@ function MainEditor({
...
-          }}
-        >
-          Reset to defaults
-        </Button>
-      </SEditorWrapper>
+        <DialogFooter>
+          <DefaultButton onClick={() => onSettingsChange({})}>Reset to defaults</DefaultButton>
+          <DefaultButton primary onClick={() => setCurrentEditingTopic(undefined)}>
+            Apply
+          </DefaultButton>
```
jtbandes, line 68: Not sure the apply button makes sense – it implies that if you don't click the button your changes won't be applied. But the changes are applied as you edit. Maybe call it Done?

2metres: Fair point, I just wanted something there

amacneil:
nice

Issue #1247, opened 2021-06-16
Feature request: Add seek to the extension api


defunctzombie:
Extension authors should be able to trigger a seek to a specific playback time.

An example use case is within the plot panel. When a user hovers over the plot they can click to seek playback to the time they were hovering. This provides a nice flow to identify an interesting point using the plot and seek to the time of the point to have other panels update.

PR #1398, opened 2021-07-06
Hover and click to seek in Map panel

### LINKED ISSUE #1394: Map panel should setPreviewTime on hover
Hovering on the playback bar highlights points on the map panel. The opposite should also be true (as originally mentioned in #801) — hovering on/near points in the map panel should use [setPreviewTime](https://github.com/foxglove/studio/blob/f40c30f070cd0ebc58cae67a8c80cfaadbf36d9d/packages/studio-base/src/components/PanelExtensionAdapter.tsx#L235) from Panel Extension API to highlight the corresponding point on the playback bar (and in other panels that listen to the previewTime).

jtbandes:
**User-Facing Changes**
Hovering a point in the Map panel now highlights the corresponding time in the playback bar, and clicking on the point seeks playback to that time.

**Description**
Fixes #1394 #1247. Fixes a bug where un-hovering the playback bar didn't remove the highlighted point in the Map panel.

### DIFF
--- a/packages/studio-base/src/components/PanelExtensionAdapter.tsx
+++ b/packages/studio-base/src/components/PanelExtensionAdapter.tsx
@@ -24,7 +24,7 @@ import {
   useHoverValue,
   useSetHoverValue,
 } from "@foxglove/studio-base/context/HoverValueContext";
-import { PlayerState } from "@foxglove/studio-base/players/types";
+import { PlayerCapabilities, PlayerState } from "@foxglove/studio-base/players/types";
 import { SaveConfig } from "@foxglove/studio-base/types/panels";
 import { fromSec, toSec } from "@foxglove/studio-base/util/time";
 
@@ -51,6 +51,14 @@ function selectRequestBackfill(ctx: MessagePipelineContext) {
   return ctx.requestBackfill;
 }
 
+function selectCapabilities(ctx: MessagePipelineContext) {
+  return ctx.playerState.capabilities;
+}
+
+function selectSeekPlayback(ctx: MessagePipelineContext) {
+  return ctx.seekPlayback;
+}
+
 /**
  * PanelExtensionAdapter renders a panel extension via initPanel
  *
@@ -64,6 +72,8 @@ function PanelExtensionAdapter(props: PanelExtensionAdapterProps): JSX.Element {
 
   const setSubscriptions = useMessagePipeline(selectSetSubscriptions);
   const requestBackfill = useMessagePipeline(selectRequestBackfill);
+  const capabilities = useMessagePipeline(selectCapabilities);
+  const seekPlayback = useMessagePipeline(selectSeekPlayback);
 
   const [subscriberId] = useState(() => uuid());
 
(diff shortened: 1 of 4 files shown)


packages/studio-base/src/components/PanelExtensionAdapter.tsx
```diff
@@ -232,6 +235,10 @@ function PanelExtensionAdapter(props: PanelExtensionAdapterProps): JSX.Element {
 
       saveState: saveConfig,
 
+      seekPlayback: (stamp: number) => {
+        latestPipelineContextRef.current?.seekPlayback(fromSec(stamp));
```
jtbandes, line 239: @defunctzombie How should we handle capabilities – e.g. for players that don't support seeking?

packages/studio/index.d.ts
```diff
@@ -82,6 +82,11 @@ declare module "@foxglove/studio" {
      */
     setPreviewTime: (time: number | undefined) => void;
 
+    /**
+     * Seek playback to the given time. Behaves as if the user had clicked the playback bar to seek.
+     */
+    seekPlayback: (time: number) => void;
```
jtbandes, line 88: @defunctzombie Why is previewTime exposed as `number` instead of `Time` – should we switch both of these to Time (or both, for backward compatibility)?

packages/studio-base/src/components/PanelExtensionAdapter.tsx
defunctzombie: Two ides:

1. The method should be a no-op for players that don't support seeking. Separately we should expose "data source capabilities" to plugins as a field.
2. The `seekPlayback` field should be marked optional - and if it is not present the panel extension author knows this functionality is not available and can adjust their UI accordingly.

I probably prefer #2 for simplicity and #1 for explicity /shrug. Probably stronger preference for #2 as an indicator this feature is not available.

packages/studio/index.d.ts
defunctzombie: I felt that `Time { sec, nsec }` was more heavyweight than necessary for the typical case of setting the preview time or seeking and that specifying seconds in floating point was sufficient - but I can see how we lose precision so maybe supporting Time could be useful. Need to decouple it from "rostime" nomenclature tho.

packages/studio-base/src/panels/Map/MapPanel.tsx
```diff
@@ -84,6 +87,7 @@ function MapPanel(props: MapPanelProps): JSX.Element {
 
     const map = new LeafMap(mapContainerRef.current, {
       layers: [tileLayer],
+      // renderer: new Canvas({ tolerance: 10 }),
```
defunctzombie, line 90: stale comment?

jtbandes: yep. This was an attempt to make it easier to hover on points by increasing the hit test radius. It kinda works, but hovering is a bit weird since it doesn't re-test for a new point until you move off the area for the currently hovered point.

PR #1416, opened 2021-07-07
Enable and fix `no-param-reassign`

jtbandes:
**User-Facing Changes**
None

**Description**
We require const for unmodified variables. Function parameters can't be annotated let/const, so it's better to treat them as const by default than non-const by default. (Ability to annotate them is tracked as a TypeScript feature request at https://github.com/microsoft/TypeScript/issues/18497)


packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/typescript/ast.ts
```diff
@@ -289,7 +289,7 @@ export const constructDatatypes = (
   depth: number = 1,
   currentTypeParamMap: TypeMap = {},
 ): { outputDatatype: string; datatypes: RosDatatypes } => {
-  if (++depth > MAX_DEPTH) {
```
jtbandes, line 292: This was an interesting one – the recursive calls already use `depth+1` so it seemed like this increment was pointless.

PR #1419, opened 2021-07-07
Use <NotificationModal /> to report player problems

### LINKED ISSUE #1409: Convert PlayerProblemModal to FluentUI

2metres:
Replace the `<PlayerProblemModal />` with the `<NotificationModal />` given that they are basically the same.

Closes #1409


packages/studio-base/src/components/NotificationModal.tsx
```diff
@@ -45,68 +22,47 @@ export default function NotificationModal({
...
-              marginLeft: theme.spacing.s1,
-              marginRight: theme.spacing.s1,
-              height: theme.spacing.l2,
-              lineHeight: theme.spacing.l2,
-              textAlign: "center",
-              flexShrink: 0,
+            field: {
+              color: theme.semanticColors.bodyText,
+              fontSize: "81.25%",
+              fontFamily: "'Ubuntu Mono', Menlo, Monaco, Courier, monospace !important",
```
defunctzombie, line 48: should we be overriding the font here or using font semantics from the theme?

packages/studio-base/src/components/NotificationModal.tsx
```diff
@@ -45,68 +22,47 @@ export default function NotificationModal({
...
-              verticalAlign: "top",
-              marginLeft: theme.spacing.s1,
-              marginRight: theme.spacing.s1,
-              height: theme.spacing.l2,
-              lineHeight: theme.spacing.l2,
-              textAlign: "center",
-              flexShrink: 0,
+            field: {
+              color: theme.semanticColors.bodyText,
+              fontSize: "81.25%",
```
defunctzombie, line 47: Thats a rather specific size - can we use theme sizing?

defunctzombie:
👍

packages/studio-base/src/components/NotificationModal.tsx
2metres: Ah yep, good catch.

PR #1425, opened 2021-07-08
Console API-based remote layout storage

jtbandes:
**User-Facing Changes**
None (feature flagged)

**Description**
Adds providers for layout storage backed by ConsoleApi, and various tweaks/fixes.


desktop/renderer/Root.tsx
```diff
@@ -58,7 +58,7 @@ export default function Root(): ReactElement {
     <StudioToastProvider />,
     <NativeStorageAppConfigurationProvider />,
     <NativeStorageLayoutCacheProvider />,
-    <LayoutStorageProviders />,
+    <FakeLayoutStorageProviders />,
```
defunctzombie, line 61: Why is this called a fake layout storage provider? Is this the actual provider now? Can we do away with this altogether?

packages/studio-base/src/App.tsx
```diff
@@ -46,11 +48,15 @@ export default function App(props: AppProps): JSX.Element {
...
 
+  // Allow platform to override layout storage, e.g. for debugging
+  const hasPlatformLayoutStorage = !!useContext(LayoutStorageContext);
+
   const providers = [
     /* eslint-disable react/jsx-key */
     <AnalyticsProvider />,
     <ConsoleApiContext.Provider value={api} />,
     <CurrentUserProvider />,
+    !hasPlatformLayoutStorage && <ConsoleApiLayoutStorageProvider />,
```
defunctzombie, line 59: A cleaner way to do this is maybe to allow the LayoutStorageProvider as a prop on app? Seems weird to me to do it this way.

Or we force the platform to provide the storage provider - and if one isn't present the features are disabled?

packages/studio-base/src/components/ExperimentalFeatureSettings.tsx
```diff
@@ -37,14 +37,19 @@ const features: Feature[] = [
     ? [
         {
           key: AppSetting.FAKE_REMOTE_LAYOUTS,
-          name: "Fake remote layout storage",
+          name: "Fake remote layout storage (desktop only)",
```
defunctzombie, line 40: Do we still want this option?

packages/studio-base/src/components/LayoutBrowser/index.tsx
```diff
@@ -70,23 +71,36 @@ export default function LayoutBrowser({
   }, [reloadLayouts]);
 
   const onSelectLayout = useCallback(
-    async (item: LayoutMetadata) => {
+    async (item: Pick<LayoutMetadata, "id">) => {
       const layout = await layoutStorage.getLayout(item.id);
       if (layout) {
-        setSelectedLayout(layout);
+        // Changing the layout causes React effects to run in panels, which may result in panels
```
defunctzombie, line 77: Does changing a layout not unmount event panel? Can we check at the location where we save the layout if the save is for a previous layout? It seems like this is patching over something else unpleasant that we should resolve.

packages/studio-base/src/services/ConsoleApiRemoteLayoutStorage.ts
```diff
@@ -0,0 +1,145 @@
...
+import { LayoutID, ISO8601Timestamp } from "@foxglove/studio-base/services/ILayoutStorage";
+import {
+  IRemoteLayoutStorage,
+  RemoteLayout,
+  RemoteLayoutMetadata,
+} from "@foxglove/studio-base/services/IRemoteLayoutStorage";
+
+const log = Logger.getLogger(__filename);
+
+export default class ConsoleApiRemoteLayoutStorage implements IRemoteLayoutStorage {
```
defunctzombie, line 17: Can we get away with just one "LayoutStorage" (i.e. remove IRemoteLayoutStorage) ? Maybe some of the concepts defined in this one should merge with our existing LayoutStorage and have that be able to express such things like conflict, etc? Feels like those concept could be relevant if someone wanted to make a smarter disk-based layout storage that also detected modified files.

web/webpack.config.ts
```diff
@@ -106,6 +106,8 @@ const mainConfig = (env: unknown, argv: WebpackArgv): Configuration => {
         AMPLITUDE_API_KEY: process.env.AMPLITUDE_API_KEY ?? null, // eslint-disable-line no-restricted-syntax
         SIGNUP_API_URL: "https://foxglove.dev/api/signup",
         SLACK_INVITE_URL: "https://foxglove.dev/join-slack",
+        OAUTH_CLIENT_ID: process.env.OAUTH_CLIENT_ID ?? "YICNZfXZp7Sx9bV0UsKkfqBT4CRYwOsY",
+        FOXGLOVE_API_URL: process.env.FOXGLOVE_API_URL ?? "http://api.localhost:3000",
```
defunctzombie, line 110: Should we default this to production url? That way when we are working on developing the app we can use our org accounts? Another thought - the API url should be an app setting?

packages/studio-base/src/services/ConsoleApiRemoteLayoutStorage.ts
jtbandes: Added to #1528

packages/studio-base/src/components/LayoutBrowser/index.tsx
jtbandes: Changed to use ref to prevent duplicate saves. Other issues may still arise as a result of non-batched updates.

packages/studio-base/src/components/ExperimentalFeatureSettings.tsx
jtbandes: Added removal to #1528

desktop/renderer/Root.tsx
jtbandes: still fake for debugging purposes, may remove in #1528

packages/studio-base/src/App.tsx
jtbandes: Changed to check the app setting instead of checking for an existing context provider

PR #1428, opened 2021-07-08
Internal: remove ts-nocheck from remaining source files

defunctzombie:
**User-Facing Changes**
None

**Description**
🔪 

Fixes https://github.com/foxglove/studio/issues/692

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -90,11 +90,7 @@ overrides:
     rules:
       "@typescript-eslint/ban-ts-comment":
         - error
-        # TODO: remove once we fix all TS issues
         - ts-expect-error: allow-with-description
-          ts-ignore: true
-          ts-nocheck: true
-          ts-check: true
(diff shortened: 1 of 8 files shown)


.eslintrc.yaml
```diff
@@ -93,7 +93,7 @@ overrides:
         # TODO: remove once we fix all TS issues
         - ts-expect-error: allow-with-description
           ts-ignore: true
-          ts-nocheck: false
+          ts-nocheck: true
           ts-check: true
```
jtbandes, line 97: These are almost the default options now, except for `ts-check` (which we probably don't really care about). Maybe remove the custom config? Upstream to @foxglove/eslint-plugin?

defunctzombie:
@jtbandes there are two places in ast.ts where I cheated with an _any_. If you want to have your hand at them go for it - I became too frustrated trying to figure out the intent.

PR #1429, opened 2021-07-08
Feature: Add Teleop Panel

2metres:
**User Facing**
Add a new panel called Teleop which has a 4 way control pad and can send `geometry_msgs/Twist`. This panel is able to control robots like Duckiebot and Turtlesim and other robots which support the `geometry_msgs/Twist` message.

Screen Shot 2021-08-16 at 12 28 25 PM

image

Fixes: #775


2metres:
@jtbandes any idea why [`makeStyles`](https://github.com/microsoft/fluentui/tree/master/packages/react/src/utilities/ThemeProvider#create-classes-for-react-components-based-on-theme) doesn't load the correct theme from the fluentui ThemeProvider? It looks like I'm getting the default them instead.

packages/studio-base/src/panels/Control/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+import { PanelExtensionContext } from "@foxglove/studio";
+import Panel from "@foxglove/studio-base/components/Panel";
+import PanelExtensionAdapter from "@foxglove/studio-base/components/PanelExtensionAdapter";
+import { SaveConfig } from "@foxglove/studio-base/types/panels";
+
+import ControlPanel from "./ControlPanel";
+import helpContent from "./index.help.md";
+
+import "leaflet/dist/leaflet.css";
```
defunctzombie, line 15: copy/paste error?

defunctzombie:
What's the plan for the control panel in terms of messages it will publish? Usually control panels are tried relatively closely to some specific type of robot or message format it expects.

packages/studio-base/src/panels/Control/index.tsx
2metres: Indeed.

2metres:
@amacneil and I were discussing tapping into the [Joystick sensor messages](http://docs.ros.org/en/api/sensor_msgs/html/msg/Joy.html)

2metres:
> @jtbandes any idea why [`makeStyles`](https://github.com/microsoft/fluentui/tree/master/packages/react/src/utilities/ThemeProvider#create-classes-for-react-components-based-on-theme) doesn't load the correct theme from the fluentui ThemeProvider? It looks like I'm getting the default them instead.

From what I can tell the panel is rendering outside the `ThemeProvider1 
Screen Shot 2021-07-08 at 3 54 54 pm

amacneil:
Can we call it either "teleop panel" or "joystick panel" - control panel sounds too generic

defunctzombie:
@amacneil The panel is called "Joystick" (in the panel list).

packages/studio-base/src/panels/JoystickPanel/index.help.md
```diff
@@ -0,0 +1,3 @@
+# Joystick
+
+<!-- FIXME: Write something about how it can be configured -->
```
jtbandes, line 3: Delete this for now?

packages/studio-base/src/panels/JoystickPanel/Settings.tsx
```diff
@@ -0,0 +1,234 @@
...
+          />
+        </Stack>
+      </StackItem>
+      <StackItem grow>
+        <Stack horizontal verticalAlign="end" tokens={{ childrenGap: theme.spacing.m }}>
+          <StackItem grow>
+            <Label>Up Button:</Label>
+          </StackItem>
+          <Dropdown
+            label="Field"
```
jtbandes, line 100: The word Field is unclear to me as a first time user. We need some explanatory text in the panel that describes what will happen when the buttons are clicked and what the settings mean.

packages/studio-base/src/panels/JoystickPanel/index.stories.tsx
```diff
@@ -0,0 +1,27 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { Story, StoryContext } from "@storybook/react";
+
+import PanelSetup from "@foxglove/studio-base/stories/PanelSetup";
+
+import JoystickPanel from "./index";
+
+export default {
+  title: "panels/Joystick/index",
```
jtbandes, line 12:
```suggestion
  title: "panels/Joystick",
```

packages/studio-base/src/panels/JoystickPanel/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+    <PanelExtensionAdapter
+      config={props.config}
+      saveConfig={props.saveConfig}
+      help={helpContent}
+      initPanel={initPanel}
+    />
+  );
+}
+
+JoystickPanelAdapter.panelType = "Control";
```
jtbandes, line 41: whatever we settle on for the name, this should be consistent

packages/studio-base/src/panels/JoystickPanel/types.ts
```diff
@@ -0,0 +1,27 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+export type DeepPartial<T> = {
+  [P in keyof T]?: DeepPartial<T[P]>;
+};
```
jtbandes, line 7: Since this is used for the extension context, it seems like something that should be part of the extension types rather than defined here

packages/studio-base/src/panels/index.ts
```diff
@@ -17,6 +17,7 @@ const builtin: PanelInfo[] = [
     module: async () => await import("./diagnostics/DiagnosticSummary"),
   },
   { title: "Image", type: "ImageViewPanel", module: async () => await import("./ImageView") },
+  { title: "Joystick", type: "Joystick", module: async () => await import("./JoystickPanel") },
```
jtbandes, line 20: no need to include the word "Panel" in the directory name

packages/studio-base/src/panels/JoystickPanel/index.stories.tsx
defunctzombie: I think this needs to be under Joystick cause we have other stories also under Joystick - I'll try this tho.

packages/studio-base/src/panels/JoystickPanel/types.ts
defunctzombie: Its used in the cast of the state - I could put it into the studio/index.d.ts file but nothing in there requires you use this. You could cast the initialState to something else?

jtbandes: Should we move it to den/types or just use the DeepPartial from some existing third party lib?

defunctzombie:
This panel is now called "Teleop"

PR #1430, opened 2021-07-08
Upgrade just-fetch

jtbandes:
**User-Facing Changes**
None

**Description**
https://github.com/foxglove/just-fetch/pull/6


amacneil:
Note: fixes a bug in main that never hit a release, so not needed in release notes

PR #1431, opened 2021-07-09
Rejig font stack

2metres:
@amacneil does this break on your Windows machine?

### DIFF
--- a/packages/studio-base/src/styles/fonts.module.scss
+++ b/packages/studio-base/src/styles/fonts.module.scss
@@ -12,7 +12,7 @@
 @mixin ff-sans-serif {
   font-family:
     // Apple
-    -apple-system, BlinkMacSystemFont,
+    -apple-system, ui-sans-serif,
     // Windows
     "Segoe UI",
     // Ubuntu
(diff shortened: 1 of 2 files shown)


amacneil:
I did something wrong my comments don't seem to appear inline with the actual PR diff

packages/studio-base/src/styles/fonts.module.scss
```diff
@@ -12,7 +12,7 @@
 @mixin ff-sans-serif {
   font-family:
     // Apple
-    -apple-system, BlinkMacSystemFont,
+    -apple-system, ui-sans-serif,
```
amacneil, line 15: I originally tried this, but from memory `ui-sans-serif` did not load `San Francisco` on mac+chrome and I had to use `BlinkMacSystemFont`. Is it actually working for you?

packages/studio-base/src/styles/fonts.module.scss
```diff
@@ -28,5 +28,18 @@
...
+    "Segoe UI Mono",
+    //  Ubuntu
+    "Ubuntu Mono",
+    // Chrome OS + Android
+    "Roboto Mono",
+    // Styled fallbacks
+    "Menlo",
+    "Monaco", "Courier",
+    // Fallback
+    "monospace";
```
amacneil, line 44: note that this is almost exactly reverting https://github.com/foxglove/studio/pull/933/files#diff-82b39cac5c90410eaf31e240139eccb2d4e217c99ecf0ad35ccde4efb3c15685R28-R31 which was set to fix a bug (per comments directly above this code block)

amacneil:
will test on windows+hidpi screen tomorrow

packages/studio-base/src/styles/fonts.ts
```diff
@@ -4,5 +4,36 @@
...
-export const SANS_SERIF = `-apple-system, BlinkMacSystemFont, "Segoe UI", "Ubuntu", "Roboto", sans-serif`;
-export const MONOSPACE = `"Roboto Mono"`;
+
+export const SANS_SERIF = [
+  // Apple
+  "-apple-system",
+  "BlinkMacSystemFont",
+  "ui-sans-serif",
+  // Windows
+  "Segoe UI",
```
amacneil, line 14: I don't know whether this will cause a problem, but by doing this you have removed quotes from the resulting string.

amacneil:
Moved to https://github.com/foxglove/studio/pull/1443

PR #1433, opened 2021-07-09
web: replace no-op storage providers with real localStorage implementations

jtbandes:
**User-Facing Changes**
None (web build is not public)

**Description**
Layouts now work properly in the web build.


web/src/components/LocalStorageAppConfigurationProvider.tsx
```diff
@@ -0,0 +1,58 @@
...
+const KEY_PREFIX = "studio.app-configuration.";
+
+export default function LocalStorageAppConfigurationProvider(
+  props: PropsWithChildren<unknown>,
+): JSX.Element {
+  const [ctx] = useState<AppConfiguration>(() => {
+    const changeListeners = new Map<string, Set<ChangeHandler>>();
+    return {
+      get(key: string): AppConfigurationValue {
+        const value = localStorage.getItem(KEY_PREFIX + key);
```
defunctzombie, line 23: can we use template strings here? Or do we do this for basic cases?

jtbandes: 🤷 I find this easier to read than `${KEY_PREFIX}${key}`

Issue #1432, opened 2021-07-09
"Copy msg" link in Raw Messages panel is broken


amacneil:
**Description**
Clicking "Copy msg" in Raw Messages causes the following error:

```
Uncaught TypeError: Cannot read private member #view from an object whose class did not declare it
    at __RootMsg.get header [as header] (eval at Ume (main.js:1), <anonymous>:530:48)
    at __RootMsg.toJSON (eval at Ume (main.js:1), <anonymous>:519:26)
    at JSON.stringify (<anonymous>)
```

Coming from https://github.com/foxglove/studio/blob/v0.12.0/packages/studio-base/src/panels/RawMessages/Metadata.tsx#L75

- Version: 0.12.0
- OS: Linux

PR #1437, opened 2021-07-09
Fix: RawMessage panel Copy msg to handle lazy messages, bigints, and typed arrays

defunctzombie:
**User-Facing Changes**
Users can now click `Copy Msg` in the Raw Message panel without the app erroring.

**Description**
Previously, the Copy msg feature of RawMessage would attempt to deep clone
the data before invoking JSON.stringify. This doesn't work for lazy messages as
they are not trivially deep clonable.

This change passes the data directory to JSON.stringify which understands `.toJSON`
functions on objects and uses that if available. Additionally, a replacer is provided
which handles bigint and typed arrays.

Fixes https://github.com/foxglove/studio/issues/1432

### DIFF
--- a/packages/studio-base/src/panels/RawMessages/Metadata.tsx
+++ b/packages/studio-base/src/panels/RawMessages/Metadata.tsx
@@ -10,16 +10,12 @@
 //   This source code is licensed under the Apache License, Version 2.0,
 //   found at http://www.apache.org/licenses/LICENSE-2.0
 //   You may not use this file except in compliance with the License.
-import ClipboardOutlineIcon from "@mdi/svg/svg/clipboard-outline.svg";
-import { cloneDeepWith } from "lodash";
-import React, { useCallback } from "react";
 import styled from "styled-components";
 
-import Icon from "@foxglove/studio-base/components/Icon";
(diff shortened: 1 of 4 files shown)


packages/studio-base/src/panels/RawMessages/Metadata.tsx
```diff
@@ -11,8 +11,9 @@
 //   found at http://www.apache.org/licenses/LICENSE-2.0
 //   You may not use this file except in compliance with the License.
 import ClipboardOutlineIcon from "@mdi/svg/svg/clipboard-outline.svg";
-import { cloneDeepWith } from "lodash";
-import React, { useCallback } from "react";
+import { TypedArray } from "esbuild-loader/node_modules/type-fest";
```
amacneil, line 14: what is this sourcery

jtbandes: i'm gonna guess this was an accidental vs code auto-import

defunctzombie: I dunno but I am removing it

PR #1439, opened 2021-07-09
Increase timeout for integration test

jtbandes:
**User-Facing Changes**
None

**Description**
I've had this test time out on me several times in CI. Re-running always makes it work. Hopefully doubling the timeout will reduce how often this happens.


defunctzombie:
`10_000` wow - the future

PR #1440, opened 2021-07-09
Topic Graph: add setting to hide topics without subscribers

jtbandes:
**User-Facing Changes**
The Topic Graph panel now allows hiding topics with no subscribers.

Showing all topics:
Screen Shot 2021-07-09 at 2 07 26 PM

Showing only topics with subscribers:
Screen Shot 2021-07-09 at 2 07 38 PM


defunctzombie:
The opposite ask is also useful - show me topics which don't have subscribers - useful to see what's not being subscribed to.

PR #1443, opened 2021-07-09
Back to system monospace fonts (again)

amacneil:
**User-Facing Changes**
Monospace fonts will use available system fonts instead of web font (Roboto Mono).

**Description**
Per request of @2metres in https://github.com/foxglove/studio/pull/1431

Reverts https://github.com/foxglove/studio/pull/933

In the past, system monospace fonts caused an electron crash on Windows machines with HiDPI screen. This no longer appears to be the case, likely due to us having upgraded electron.

**TODO**
- Bikeshed actual font choices
- More testing

@2metres let's avoid putting any changes to the sans serif fonts in this PR, keep things simple in case we need to revert it again

**Before** (Windows HiDPI)
Screenshot 2021-07-09 154429

**After** (Windows HiDPI)
Screenshot 2021-07-09 154007


2metres (approved):
Looks good to me

2metres:
@amacneil the only issue I saw in the visual diff is the ROS button is Monospace for some reason. I'll make the change now

PR #1453, opened 2021-07-14
Internal: upgrade dependencies

defunctzombie:
**User-Facing Changes**
None

**Description**
The grind continues.


defunctzombie:
Something has happened to the fonts which I do not yet understand.

PR #1457, opened 2021-07-14
Lower resumeCount to 6 in x-axes story

jtbandes:
This now times out occasionally, which is worse than capturing a different snapshot because I can't manually approve the change


defunctzombie:
LGTM - seems like the state transition stories will need love

PR #1463, opened 2021-07-15
Internal: add prepack command to studio-base to create module

defunctzombie:
**User-Facing Changes**
We can create and publish `@foxglove/studio-base` with minimal setup for end users.

**Description**


packages/studio-base/webpack.config.ts
```diff
@@ -0,0 +1,54 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+/* eslint-disable filenames/match-exported */
+
+import { CleanWebpackPlugin } from "clean-webpack-plugin";
```
amacneil, line 7: can you add a comment about why/when this webpack config is used

amacneil (approved):

PR #1466, opened 2021-07-15
Convert player components to use fluentui

2metres:
Replace the IconButtons and Dropdowns in the Player to use fluentui components.

**Before:**
Screen Shot 2021-07-15 at 4 07 02 pm

**After:**
Screen Shot 2021-07-15 at 4 07 06 pm


packages/studio-base/package.json
```diff
@@ -24,6 +24,7 @@
   },
   "dependencies": {
     "@fluentui/react": "8.22.0",
+    "@fluentui/react-icons": "1.1.135",
```
2metres, line 27: Introduces another [fluentui/react-icons](https://github.com/microsoft/fluentui-system-icons/tree/master/packages/react-icons) library, with more icons including filled in addition to outlined variants.

Why? Not sure. Description from the readme

> (Note: For those who were consuming @fluentui/react-icons v 0.x, we are releasing a new set of icons under @fluentui/react-icons v 1.x. You can still use 0.x version for the old icons, but if you major bump to 1.x, you will be using the new icons. The Fabric MDL2 SVG Icon set that is released with v 0.x has been rereleased in @fluentui/react-icons-mdl2.)
>
> Fluent System Icons are a collection of familiar, friendly, and modern icons from Microsoft.

packages/studio-base/src/components/HoverableIconButton.tsx
```diff
@@ -0,0 +1,31 @@
...
+      {...props}
+      iconProps={{
+        iconName: iconNameActive != undefined ? (hovered ? iconNameActive : iconName) : iconName,
+        ...rest,
+      }}
+      onMouseOver={() => setHovered(true)}
+      onMouseLeave={() => setHovered(false)}
+    />
+  );
+}
```
2metres, line 31: Introduces `<HoverableIconButton />` that extends the `<IconButton />` and swaps between outlined and filled version or any other swap you might want.

According to `@fluentui/react-icons` they have a way to do this with a helper util but I couldn't get it running do to sparse documentation.

packages/studio-base/src/components/LoopIcon.tsx
```diff
@@ -0,0 +1,42 @@
...
+          stroke="currentColor"
+          strokeWidth={strokeWidth}
+          strokeLinecap="round"
+          strokeLinejoin="round"
+          d="M13.308,20.121l-2.121,-2.121l2.121,-2.121"
+        />
+      </g>
+    </svg>
+  );
+}
```
2metres, line 42: Found the of the filled/outlined variants of `<RepeatAllIcon />` had a strange jump

Screen Recording 2021-07-15 at 12 27 45 pm

Decided to make my own Loop/Repeat Icon

packages/studio-base/src/components/PlaybackSpeedControls.tsx
```diff
@@ -57,28 +51,43 @@ export default function PlaybackSpeedControls(): JSX.Element {
...
+        gapSpace: 6,
+        items: SPEEDS.map(
+          (s: number): IContextualMenuItem => ({
+            canCheck: true,
+            key: `${s}`,
+            text: `${displaySpeed(s)}`,
+            isChecked: configSpeed === s,
+            onClick: () => setSpeed(s),
+          }),
+        ),
```
2metres, line 71: Could use some human typechecking here. Not sure if this is the right approach

packages/studio-base/src/components/PlaybackSpeedControls.tsx
```diff
@@ -10,32 +10,26 @@
...
 export default function PlaybackSpeedControls(): JSX.Element {
-  const configSpeed = useCurrentLayoutSelector(
-    (state) => state.selectedLayout?.data.playbackConfig.speed,
-  );
-  const speed = useMessagePipeline(
-    useCallback(({ playerState }) => playerState.activeData?.speed, []),
-  );
+  const theme = useTheme();
+  const configSpeed =
+    useCurrentLayoutSelector((state) => state.selectedLayout?.data.playbackConfig.speed) ?? 1;
```
2metres, line 30: Added a fallback of `1` here not sure that this is the right approach but makes sense to me

2metres (commented):
Overall most of the pain was dealing with iconsets but mostly looks the same only slightly tighter styles

packages/studio-base/package.json
jtbandes: I guess they are just different icon sets 🤷 Do they look good when mixed & matched? or are we going to replace all icons with this package?

packages/studio-base/src/components/MessageOrderControls.tsx
```diff
@@ -40,25 +37,58 @@ export default function MessageOrderControls(): JSX.Element {
...
+      calloutProps={{
+        styles: {
+          // TODO: this needs to be fixed in the theme
+          beak: { backgroundColor: "#fdfdfd" },
+          beakCurtain: { backgroundColor: "#fdfdfd" },
+        },
+      }}
+      content={`Order messages by ${orderText.toLowerCase()}`}
+      id={tooltipId}
+    >
```
jtbandes, line 52: You might be able to use the `useTooltip` that I set up to get better default tooltip styling

packages/studio-base/package.json
```diff
@@ -23,10 +23,11 @@
     "storybook:build": "cross-env NODE_OPTIONS='--max-old-space-size=6144' TS_NODE_COMPILER_OPTIONS='{\"module\":\"commonjs\"}' build-storybook --config-dir src/.storybook"
   },
   "devDependencies": {
-    "@fluentui/react": "8.23.2",
-    "@fluentui/react-icons-mdl2": "1.1.6",
+    "@fluentui/react": "8.22.0",
+    "@fluentui/react-icons": "1.1.135",
+    "@fluentui/react-icons-mdl2": "1.1.4",
```
jtbandes, line 28: why did this get downgraded?

packages/studio-base/src/components/PlaybackControls/PlaybackTimeDisplayMethod.tsx
```diff
@@ -173,20 +172,41 @@ const PlaybackTimeDisplayMethod = ({
...
       )}
-      <Dropdown
-        position="above"
-        value={timeDisplayMethod}
-        text={timeDisplayMethod}
-        onChange={(val) => setTimeDisplayMethod(val)}
-        btnClassname={styles.timeDisplayMethodButton}
+      <DefaultButton
+        menuProps={{
+          gapSpace: 6,
```
jtbandes, line 177: is this something we can set in the theme?

packages/studio-base/src/components/PlaybackSpeedControls.tsx
```diff
@@ -57,28 +51,43 @@ export default function PlaybackSpeedControls(): JSX.Element {
...
+        },
+      }}
     >
-      {SPEEDS.map((eachSpeed: string) => (
-        <DropdownItem key={eachSpeed} value={parseFloat(eachSpeed)}>
-          <span>{eachSpeed}x</span>
-        </DropdownItem>
-      ))}
-    </Dropdown>
+      {`${displaySpeed(configSpeed)}x` ?? "–"}
```
jtbandes, line 90: consider × instead of x

packages/studio-base/src/components/PlaybackSpeedControls.tsx
jtbandes: Would prefer not to have this fallback in the display component since we have some kind of speed fallback in some other places already. That's why the "–" was shown in case the speed was undefined (though in practice I don't think it would happen often)

packages/studio-base/src/components/PlaybackSpeedControls.tsx
jtbandes: this ?? is doing nothing here, might want to change to ``configSpeed == undefined ? "–" : `...` ``

packages/studio-base/package.json
2metres: They are largely the same, and have similar esthetics. I think we could likely replace the mdl2 version with this one.

Most cases where they differ, I tend to prefer the difference for readability reasons

**Example**

**Pause Icon from `fluentui/react-icons-mdl2`**
Screen Shot 2021-07-16 at 8 24 53 am

**Pause Icon from `fluentui/react-icons`**
Screen Shot 2021-07-16 at 8 25 23 am

packages/studio-base/package.json
2metres: Accidental, managed to bork my `yarn.lock` in the process. Reverting to prior version

packages/studio-base/src/components/PlaybackControls/PlaybackTimeDisplayMethod.tsx
2metres: Tried to set it on `DefaultButton` and `ContextualMenu` in the theme and doesn't seem to work for props just `ComponentStyles`

packages/studio-base/src/components/PlaybackSpeedControls.tsx
jtbandes: Which part here were you asking about specifically? I think it looks alright

packages/studio-base/src/components/PlaybackControls/PlaybackTimeDisplayMethod.tsx
```diff
@@ -143,51 +124,95 @@ const PlaybackTimeDisplayMethod = ({
...
+              },
+            }}
           >
             {currentTimeString}
-          </STimestamp>
+          </Text>
         )
       ) : (
         <span data-test="PlaybackTime-text">–</span>
       )}
```
2metres, line 179: @defunctzombie @jtbandes Is there a reason why this can't be a TextField that has subtle styles when unfocused?

packages/studio-base/src/components/PlaybackSpeedControls.tsx
```diff
@@ -57,28 +43,45 @@ export default function PlaybackSpeedControls(): JSX.Element {
...
+    <DefaultButton
+      data-test="PlaybackSpeedControls-Dropdown"
+      menuProps={{
+        calloutProps: {
+          calloutMaxWidth: 80,
+        },
+        directionalHint: DirectionalHint.topLeftEdge,
+        gapSpace: 6,
+        items: SPEEDS.map(
+          (s: number): IContextualMenuItem => ({
```
defunctzombie, line 58: nit: I don't like signal letter variable names - I prefer seeing the word so I don't have to think about what the letter stands for.

PR #1473, opened 2021-07-16
Cleanup .prettierrc.yaml

amacneil:
**User-Facing Changes**
None

**Description**
- Remove `babel-ts` config which links to an issue that seems to have been fixed in recent versions of prettier
- Remove redundant prettierrc configuration which specifies defaults
- Enable `files.insertFinalNewline` and `files.trimFinalNewlines` in vscode settings to catch files which prettier does not support

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,1 +1,3 @@
 .github/**/*.md
+.yarn/plugins
+.yarn/releases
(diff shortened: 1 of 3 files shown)


defunctzombie:
LGTM

PR #1474, opened 2021-07-16
Internal: Lazy load builtin panels

### LINKED ISSUE #747: Lazy load panel components on demand
Rather than importing all panels into the BuiltinPanelCatalog - make a catalog from instances (maybe of a Panel class) which will lazy load panel components on demand. This will lower the amount of code loading up-front to render the initial app containers.

defunctzombie:
**User-Facing Changes**
None.

**Description**
Rather than import all builtin panels, use dynamic imports to only
load the builtin panel module when added to the layout.

Using dynamic loading for panels decreases the size of the main bundle which helps
with initial load times.

Fixes #747


defunctzombie:
@jtbandes one change here is that useConfigById returns undefined configs if the config is not loaded from the current layout. I'm thinking that the Panel HOC should detect that config is undefined, and invoke saveConfig with the default config since it has the default config.

packages/studio-base/src/components/PanelLayout.tsx
```diff
@@ -108,16 +108,16 @@ export function UnconnectedPanelLayout(props: Props): React.ReactElement {
...
-        // If we haven't found a panel of the given type, render the panel selector
-        <Flex col center dataTest={id}>
-          <PanelToolbar floating isUnknownPanel />
-          Unknown panel type: {type}.
-        </Flex>
-      );
+      const panelInfo = panelCatalog.getPanelByType(type);
+      // If we haven't found a panel of the given type, render the panel selector
+      const PanelComponent = panelInfo
+        ? React.lazy(panelInfo.module)
```
jtbandes, line 114: Is it bad to be re-running React.lazy every time? Wouldn't that create a new component type and cause unnecessary remounts?

packages/studio-base/src/components/PanelSettings/index.tsx
```diff
@@ -87,19 +100,25 @@ export default function PanelSettings(): JSX.Element {
     );
   }
 
+  if (!config) {
+    return (
+      <SidebarContent title={`Panel Settings`}>
+        <Text styles={{ root: { color: theme.palette.neutralTertiary } }}>
+          loading panel settings...
```
jtbandes, line 107: maybe use an actual spinner component (fluentui has one)?

packages/studio-base/src/context/PanelCatalogContext.ts
```diff
@@ -5,11 +5,18 @@
...
 
 export type PanelInfo = {
   title: string;
-  component: ComponentType<{ childId?: string; tabId?: string }> &
-    PanelStatics<Record<string, unknown>>;
+  type: string;
+
+  // The panel module is a function to load the panel.
+  // This is to support our lazy built-in panels
+  module: () => Promise<{ default: PanelComponent }>;
```
jtbandes, line 19: prefer `/** */` style comments for new doc comments - VS Code picks it up to show when hovering on types and such

packages/studio-base/src/components/PanelLayout.tsx
defunctzombie: I think those code makes a new component every time regardless since it is a useCallback? I could make the lazy component once tho in a separate useMemo.

jtbandes: > I think those code makes a new component every time regardless since it is a useCallback?

hm, I guess it depends how renderTile is used internally in the mosaic. I think it's used like `renderTile()` and not `<renderTile />`.

The unnecessary remounts happen when the function used as the react tag/element changes identity. Though I guess the dependencies here are `[createTile, tabId, panelCatalog]` so it wouldn't change very often.

PR #1478, opened 2021-07-16
Internal: remove explicit chunk names from NodePlayground panel

defunctzombie:
**User-Facing Changes**
None

**Description**
We prefer to allow webpack to decide how to split the chunks rather
than forcing specific chunk names.


defunctzombie:
Blocked on https://github.com/foxglove/studio/pull/1478 to avoid bloating main.js further

defunctzombie:
Confirmed that main.js remains small after these changes - so the dynamically loaded modules are still in separate chunks.

PR #1479, opened 2021-07-16
Feature: allow bags with empty message definitions

defunctzombie:
**User-Facing Changes**
Bags containing empty message definitions (std_msgs/Empty) are loadable.

**Description**
An empty message definition is a valid message definition. This
change allows bags that have topics with empty message definitions
to load and play. Since the definition is empty, such topics are
not deserialized.

Alters the fix performed in webviz in response to https://github.com/cruise-automation/webviz/issues/373
where the app would previously crash when there was an empty message definition.

This change removes a flaky test for the notification when loading such
a bag since it is valid to load such a bag.


defunctzombie:
I'm on the fence with this one. Part of me thinks that an empty message definition should not be supported. It would mean that std_msgs/Empty would not be supported even tho it is a standard message type. It seems more likely that an empty message definition is an error with the bag and something to notify the user about rather than handle and display no data for the topic since the message definition says there is nothing to deserialize.

PR #1480, opened 2021-07-16
Add ROS 2 support to RosbridgePlayer

### LINKED ISSUE #716: ROS 2 rosbridge websocket connection support
Placeholder ticket to discuss if ROS 2 websocket connections are interesting to anyone.

In theory we could allow connections over websocket (similar to what we support with `rosbridge` in ROS 1) using https://github.com/RobotWebTools/ros2-web-bridge.

However, we have also discussed a native-connection-over-ssh option which might be more useful than websocket connections. We're open to feedback.

See also:
- https://github.com/foxglove/studio/issues/715 (ROS 2 native connections)
- https://github.com/foxglove/studio/issues/717 (ROS 2 bag support)
- https://github.com/RobotWebTools/ros2-web-bridge/issues/174
- https://github.com/RobotWebTools/rosbridge_suite/issues/345

jtbandes:
**User-Facing Changes**

**Description**
Fixes #716
Depends on https://github.com/RobotWebTools/rosbridge_suite/pull/574


desktop/renderer/Root.tsx
```diff
@@ -32,19 +32,23 @@ export default function Root(): ReactElement {
...
       type: "ros1-local-bagfile",
     },
     {
-      name: "ROS 1 Bag File (HTTP)",
+      name: "ROS 1 Bag (HTTP)",
       type: "ros1-remote-bagfile",
     },
     {
-      name: "ROS 2 Bag Folder (local)",
+      name: "ROS 2 Bag (local)",
```
defunctzombie, line 51: I think this requires a folder or can it also handle opening a single file? /cc @jhurliman

packages/studio-base/src/components/PlayerManager.tsx
```diff
@@ -490,8 +506,14 @@ export default function PlayerManager({
...
       case "ros1-socket":
         return roscoreSource;
-      case "ros-ws":
+      case "ros1-rosbridge-websocket":
+      case "ros2-rosbridge-websocket":
         return rosbridgeSource;
+        return async (options: FactoryOptions) =>
+          await rosbridgeSource({
+            ...options,
+            sourceOptions: { ...options.sourceOptions, rosVersion: 2 },
```
defunctzombie, line 515: Is this stale? This is a return under a return - so this code path won't run.

jtbandes: I thought eslint or typescript would catch things like this. But it sounds like it's turned off by default in TS: https://github.com/microsoft/TypeScript/issues/25738

desktop/renderer/Root.tsx
esthersweon: It should be a folder containing a `metadata.yaml` and `*.db3` files.

desktop/renderer/Root.tsx
```diff
@@ -48,7 +48,7 @@ export default function Root(): ReactElement {
       type: "ros1-remote-bagfile",
     },
     {
-      name: "ROS 2 Bag (local)",
+      name: "ROS 2 Bag Folder (local)",
```
amacneil: I suggested removing `File` / `Folder` in an earlier comment, because it seems redundant (now the PR only removes `File` and not `Folder` which is weird). For ROS 2 bags it's looking likely that we will support loading a file or a folder, since the db3 files are becoming self contained.

esthersweon: Removed all "file" and "folder" references.

amacneil:
this is ok for an experimental / feature flagged feature, but in a follow up PR could we make it just one rosbridge connection in the connection picker, and automatically detect whatever protocol differences exist between ROS 1/2?

PR #1485, opened 2021-07-17
Internal: simplify storybook ready signal use

defunctzombie:
**User-Facing Changes**
None

**Description**
* Change all stories to the useReadySignal hook
* Change screenshot.signal parameter to useReadySignal parameter

The useReadySignal parameter instructs storybook to create a suitable
ready signal function rather than the user having to provide a promise.

### DIFF
--- a/patches/storybook-react.patch
+++ b/patches/storybook-react.patch
@@ -2,12 +2,13 @@ diff --git a/dist/esm/client/preview/render.js b/dist/esm/client/preview/render.
 index 5883ed6c7e9db640cd8f209a3b840efb1d54c6a7..03c377000bdc7eb620c2f9107a2635e820d69247 100644
 --- a/dist/esm/client/preview/render.js
 +++ b/dist/esm/client/preview/render.js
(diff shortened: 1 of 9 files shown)


patches/storybook-react.patch
```diff
@@ -2,26 +2,27 @@ diff --git a/dist/esm/client/preview/render.js b/dist/esm/client/preview/render.
...
-+var render = function render(node, el, maybeSignal) {
++var render = function render(node, el, storyContext) {
++  var maybeSignal = storyContext.parameters?.storyReady;
    return new Promise(function (resolve) {
 -    ReactDOM.render(node, el, resolve);
 +    ReactDOM.render(node, el, () => {
 +      maybeSignal ? maybeSignal.then(resolve) : resolve();
 +    });
    });
  };
```
jtbandes, line 18: Here's a dik-dik for your diff-diff

image

defunctzombie: I don't know what happened here but I approve of this animal.

PR #1486, opened 2021-07-18
wip: attempt to remove global.scss

defunctzombie:
**User-Facing Changes**
TBD

**Description**
TBD


defunctzombie:
This did not work as easily as I anticipated - I had hoped to scope everything under the appContainer class - but something is really off with sidebar and other buttons.

PR #1497, opened 2021-07-19
Reduce install size by removing .map files from ASAR

jhurliman:
We already upload these files directly to Sentry, they do not need to be in the app installation images as well. Map files take up two thirds of the uncompressed ASAR disk space.

### DIFF
--- a/desktop/integration-test/build.ts
+++ b/desktop/integration-test/build.ts
@@ -18,9 +18,12 @@ export default async (): Promise<void> => {
   const compiler = webpack(
     webpackConfig.map((config) => {
       if (typeof config === "function") {
-        return config(undefined, { mode: "production" });
+        const cfg = config(undefined, { mode: "production" });
+        cfg.devtool = false;
+        return cfg;
       }
 
+      config.devtool = false;
       return config;
     }),
   );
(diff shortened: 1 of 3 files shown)


defunctzombie:
Does this work? I think the rimraf happens too late - the package has already been created.

desktop/integration-test/build.ts
```diff
@@ -18,9 +18,12 @@ export default async (): Promise<void> => {
   const compiler = webpack(
     webpackConfig.map((config) => {
       if (typeof config === "function") {
-        return config(undefined, { mode: "production" });
+        const cfg = config(undefined, { mode: "production" });
+        cfg.devtool = false;
```
defunctzombie, line 22: This needs a comment on what is happening. I'd probably have a slight preference for not changing devtool here and instead removing the map files in the buildSize test (or ignoring them there).

jtbandes:
Doesn't this make it harder to use the in-app debugger?

defunctzombie:
@jtbandes @jhurliman thoughts on pushing this over the finish line? Seems like we don't do a lot of stepping through the code with users and we already upload maps to sentry (as stated in the issue) so maybe its worth the decrease in bundle size to merge this?

PR #1502, opened 2021-07-19
Add banner for unsupported browsers + outdated Chrome versions

### LINKED ISSUE #1490: Warning banner for unsupported browsers
**Description**
studio.foxglove.dev is fussy about which browsers it works on. We are tracking cross-browser work in https://github.com/foxglove/studio/issues/1511.

In the meantime, we should display a banner across the top of the window for unsupported browsers encouraging them to try Chrome for now, and linking to a Github meta-ticket.

On my current gen iphone/iOS, I get nothing but a white screen. In this case we should display the same warning and ensure that it doesn't require our javascript bundle to load.

### LINKED ISSUE #1511: Meta: Improve Cross-Browser Compatibility
This is a meta-issue tracking work happening to improve cross-browser compatibility. Currently, we only support Chrome (v76+?).

- [ ] Edge: I don't believe anyone has tested, but since it uses Blink it might work?
- [ ] Safari, and all browsers on iOS/iPadOS (which are forced to use Webkit) (https://github.com/foxglove/studio/issues/1422)
- [ ] Firefox

esthersweon:
**User-Facing Changes**
Alert users to when their Chrome version is outdated, or when they're using an unsupported browser (i.e. not Chrome).

**Description**
Example of banner in Safari:

Screen Shot 2021-07-19 at 2 28 02 PM

Resolves #1511.

### DIFF
--- a/packages/studio-base/src/App.tsx
+++ b/packages/studio-base/src/App.tsx
@@ -25,8 +25,6 @@ import PanelCatalogProvider from "@foxglove/studio-base/providers/PanelCatalogPr
 import ConsoleApi from "@foxglove/studio-base/services/ConsoleApi";
 import URDFAssetLoader from "@foxglove/studio-base/services/URDFAssetLoader";
 
-import "./styles/global.scss";
-
 type AppProps = {
   availableSources: PlayerSourceDefinition[];
   demoBagUrl?: string;
--- a/packages/studio-base/src/App.tsx
+++ b/packages/studio-base/src/App.tsx
@@ -25,6 +25,8 @@ import PanelCatalogProvider from "@foxglove/studio-base/providers/PanelCatalogPr
 import ConsoleApi from "@foxglove/studio-base/services/ConsoleApi";
 import URDFAssetLoader from "@foxglove/studio-base/services/URDFAssetLoader";
 
+import "./styles/global.scss";
+
 type AppProps = {
   availableSources: PlayerSourceDefinition[];
   demoBagUrl?: string;
--- a/web/src/VersionBanner.tsx
+++ b/web/src/VersionBanner.tsx
@@ -1,0 +1,64 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+import { useState, ReactElement } from "react";
+import styled from "styled-components";
+
+import "@foxglove/studio-base/styles/global.scss";
+
+const MINIMUM_CHROME_VERSION = 91;
+const StyledBanner = styled.div`
+  text-align: center;
+  position: absolute;
+  top: 0px;
+  left: 0px;
+  width: 100vw;
+  padding: 10px;
+  background-color: rgba(99, 102, 241, 0.9);
+  z-index: 100;
+`;
+
+const VersionBanner = function ({
+  isChrome,
+  currentVersion,
+}: {
+  isChrome: boolean;
+  currentVersion: number;
+}): ReactElement | ReactNull {
+  const [showBanner, setShowBanner] = useState(true);
+
+  if (!showBanner || currentVersion >= MINIMUM_CHROME_VERSION) {
+    return ReactNull;
+  }
+
+  const prompt = isChrome
+    ? `Update Chrome to version ${MINIMUM_CHROME_VERSION}+ to continue.`
+    : `You're using an unsupported browser. Use Chrome ${MINIMUM_CHROME_VERSION}+ to continue.`;
+  const fixText = isChrome ? "Update Chrome" : "Download Chrome";
+  const continueText = isChrome
+    ? "Continue with unsupported version"
+    : "Continue with unsupported browser";
+
+  return (
+    <StyledBanner>
+      <div>
+        <p>{prompt} </p>
+        {isChrome ? undefined : (
+          <p>
+            Check out our browser support progress in{" "}
+            <a href="https://github.com/foxglove/studio/issues/1422">this GitHub issue</a>.
+          </p>
+        )}
+      </div>
+
+      <div style={{ paddingTop: "20px" }}>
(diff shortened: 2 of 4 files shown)


defunctzombie:
👍

packages/studio-base/src/App.tsx
```diff
@@ -25,8 +25,6 @@ import PanelCatalogProvider from "@foxglove/studio-base/providers/PanelCatalogPr
 import ConsoleApi from "@foxglove/studio-base/services/ConsoleApi";
 import URDFAssetLoader from "@foxglove/studio-base/services/URDFAssetLoader";
 
-import "./styles/global.scss";
```
defunctzombie, line 28: This needs to be here for our global styles. I explicitly moved it here so downstream users get these styles without additional imports.

web/src/index.tsx
```diff
@@ -3,14 +3,63 @@
...
 import { init as initSentry } from "@sentry/browser";
+import { useState } from "react";
 import ReactDOM from "react-dom";
+import styled from "styled-components";
 
 import Logger from "@foxglove/log";
-import { installDevtoolsFormatters, overwriteFetch, waitForFonts } from "@foxglove/studio-base";
 
-import Root from "./Root";
+import "@foxglove/studio-base/styles/global.scss";
```
defunctzombie, line 12: We don't put our css files into the studio-base package so we shouldn't expose these like this. At the least we could make a component that injects these global styles.

web/src/index.tsx
```diff
@@ -3,14 +3,63 @@
...
+  position: absolute;
+  top: 0px;
+  left: 0px;
+  width: 100vw;
+  padding: 10px;
+  background-color: rgba(99, 102, 241, 0.9);
+  z-index: 100;
+`;
+
+const VersionBanner = function ({ isChrome }: { isChrome: boolean }) {
```
defunctzombie, line 27: Should be in a separate file

web/src/index.tsx
```diff
@@ -30,22 +79,58 @@ if (typeof process.env.SENTRY_DSN === "string") {
...
+  installDevtoolsFormatters();
+  overwriteFetch();
   // consider moving waitForFonts into App to display an app loading screen
   await waitForFonts();
 
-  ReactDOM.render(<Root />, rootEl, () => {
-    // Integration tests look for this console log to indicate the app has rendered once
-    log.debug("App rendered");
-  });
+  const { Root } = await import("./Root");
```
defunctzombie, line 114: Should this import happen within the `if` ?

web/src/index.tsx
```diff
@@ -30,22 +79,58 @@ if (typeof process.env.SENTRY_DSN === "string") {
...
+  if (chromeVersion >= MINIMUM_CHROME_VERSION) {
+    ReactDOM.render(<Root />, rootEl, () => {
+      // Integration tests look for this console log to indicate the app has rendered once
+      log.debug("App rendered");
+    });
+    return;
+  }
+
+  ReactDOM.render(
```
defunctzombie, line 123: Can we cut down on the number of ReactDOM.render calls to avoid the duplication (especially the App rendered) duplication. Ideally we can have on ReactDOM render call for one top-level component and within that figure out what else to render.

web/src/index.tsx
esthersweon: No, since we use it on line 123 as well. Will reduce ReactDOM.render calls though.

web/src/index.tsx
esthersweon: I cut it down to 2, and used vars for the banner & the "App rendered" callback to reduce duplication.

For some reason, lazy-loading the App component from within the Root component isn't working as expected. It was also hairy trying to invoke the async `installDevtoolsFormatters`, `overwriteFetch`, and `waitForFonts` functions inside Root, so kept all that outside here.

Screen Shot 2021-07-19 at 6 13 00 PM

amacneil: given the various failure modes I can think of that we need to handle (e.g. browser failing to parse our JS bundle), would it make sense to render this banner completely outside of react? 

for example, if it just lived in our index.html file as static HTML, then a react effect hook just finds it with `document.getElementByID()` or something and removes that element on app load for a valid browser?

defunctzombie: > `document.getElementByID()` or something and removes that element on app load for a valid browser?

Need to be smarter than that because the banner will be present until the code loads and removes it - which means users on supported browsers would see it as well during loading which is undesirable.

amacneil: I guess the browser detection JS could also go in index.html inline then? It could be done with very simple `navigator.userAgent.includes()`, we're not doing anything particularly clever here. even the "chrome min version" probably isn't super necessary, although if we need it a regex would suffice

web/src/VersionBanner.tsx
```diff
@@ -0,0 +1,64 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+import { useState, ReactElement } from "react";
+import styled from "styled-components";
+
+import "@foxglove/studio-base/styles/global.scss";
+
+const MINIMUM_CHROME_VERSION = 91;
```
amacneil, line 9: @jhurliman says chrome v76+ should work. we should test but ideally support as many versions as practical.

web/src/VersionBanner.tsx
```diff
@@ -0,0 +1,64 @@
...
+    : "Continue with unsupported browser";
+
+  return (
+    <StyledBanner>
+      <div>
+        <p>{prompt} </p>
+        {isChrome ? undefined : (
+          <p>
+            Check out our browser support progress in{" "}
+            <a href="https://github.com/foxglove/studio/issues/1422">this GitHub issue</a>.
```
amacneil, line 49: Can you update the issue link to https://github.com/foxglove/studio/issues/1511 (just created to roll up various browser compat tickets)

amacneil: sorry I meant the issue link in the actual component

PR #1505, opened 2021-07-20
Add back ability to add a preconfigured panel (regular or Tab) to layout

esthersweon:
**User-Facing Changes**
- No user-facing changes

**Description**
- Added functionality to add a panel to a layout with its own config and related configs (for Tab panels with child panels)
- TODO in separate PR: Add UI for adding a preconfigured panel to your list, and add logic for persisting locally saved preconfigured panels


defunctzombie:
Should preconfigured panels appear in a separate section in the panel list? Maybe we need to bring back sections to separate out Builtin (studio), user (or org), and extensions?

defunctzombie:
This probably needs some test or storybooks injecting preconfigured panels to show how they are displayed?

esthersweon:
@defunctzombie -
> Should preconfigured panels appear in a separate section in the panel list? Maybe we need to bring back sections to separate out Builtin (studio), user (or org), and extensions?

> This probably needs some test or storybooks injecting preconfigured panels to show how they are displayed?

I think they should be in a separate section. 

Wanted to keep this PR separate from any UI changes, since that isn't set in stone yet (cc: @2metres), but added a fairly uncontroversial commit (with storybook tests) that places preconfigured panels in a list above the regular panels:

Screen Shot 2021-07-20 at 12 15 02 AM

packages/studio-base/src/components/PanelList/index.tsx
```diff
@@ -306,8 +335,13 @@ function PanelList(props: Props): JSX.Element {
         </div>
       </StickyDiv>
       <SScrollContainer>
+        {noPreconfiguredResults && allPreconfiguredPanels.length > 0 && (
```
defunctzombie, line 338: Is it worth separating preconfigured from builtin? I'd think treating it as one list for purposes of search is simpler.

packages/studio-base/src/context/PanelCatalogContext.ts
```diff
@@ -19,12 +19,14 @@ export type PanelInfo = {
...
   module: () => Promise<{ default: PanelComponent }>;
+  config?: PanelConfig;
+  relatedConfigs?: { [panelId: string]: PanelConfig };
 };
 
 /** PanelCatalog describes the interface for getting available panels */
 export interface PanelCatalog {
-  /** get a list of the available panels */
-  getPanels(): readonly PanelInfo[];
+  // get a list of the available panels
```
defunctzombie, line 28: Why the comment format change? We prefer the /** style since that will make it show up in vscode hover hints.

packages/studio-base/src/context/PanelCatalogContext.ts
```diff
@@ -19,12 +19,14 @@ export type PanelInfo = {
...
+  getPanels(): { regular: readonly PanelInfo[]; preconfigured: readonly PanelInfo[] };
```
defunctzombie, line 29: I'd probably make preconfigured a flag on the panel maybe? For a user point of view they are the same. Tho we will need to think about org preconfigured panels.

packages/studio-base/src/context/PanelCatalogContext.ts
esthersweon: This was unintentional - changing back.

packages/studio-base/src/context/PanelCatalogContext.ts
esthersweon: Added a boolean `preconfigured` flag to `PanelInfo` type. Maybe it can eventually be `preconfigured: "user" | "org"`?

packages/studio-base/src/components/PanelList/index.tsx
esthersweon: Now showing `allFilteredPanels` in one flat list.

PR #1507, opened 2021-07-20
Change: use fluentui spinner for lazy loaded panels

### LINKED ISSUE #1504: Follow-up tasks to lazy load builtin panels
Follow up from https://github.com/foxglove/studio/pull/1474

- [x] use fluentui spinner
- [x] Avoid making React.lazy every time?
- [x] prefer /** */ style comments

defunctzombie:
**User-Facing Changes**
When panel code is loading for the first time the user is briefly shown a spinner.

**Description**
Panels load lazily so we can keep the amount of code loaded on first render smaller. This helps with app startup perception.

Fixes #1504


defunctzombie:
Still need to figure out a way to avoid the React.lazy making a new component every time.

PR #1514, opened 2021-07-20
Display only the version banner for browser versions that crash on App render

### LINKED ISSUE #1511: Meta: Improve Cross-Browser Compatibility

esthersweon:
**User-Facing Changes**
User should be able to see partly working app on browsers outside of Chrome (supported) and Safari 14 (not functional). 

Firefox (not supported, but partly working app) vs. Safari 14 (app not functional):
Screen Shot 2021-07-20 at 2 47 57 PM

**Description**
Does a more generic check for any browser that does not recognize `BigInt64Array` / `BigUint64Array` to render just the banner (i.e. don't load the app at all, as whole page including banner will crash).

Resolve #1511.


amacneil:
Is `BigUint64Array` the only browser-compatibility issue we have at this point?

Maybe I am old school but it feels more sensible to whitelist a set of browsers we have tested with (only Chrome at this stage), given how many different things could potentially cause issues?

cc @jhurliman @defunctzombie @jtbandes @esthersweon

web/src/VersionBanner.tsx
```diff
@@ -51,11 +62,10 @@ const VersionBanner = function ({
         )}
       </div>
 
-      <div style={{ paddingTop: "20px" }}>
+      <div style={{ paddingTop: "10px" }}>
         <a href="https://www.google.com/chrome/" target="_blank" rel="noreferrer">
           <button>{fixText}</button>
```
jtbandes, line 67: can we use one of the FluentUI Button classes? our intent is to remove the global css that makes buttons look like they currently do

web/src/VersionBanner.tsx
```diff
@@ -32,16 +39,20 @@ const VersionBanner = function ({
...
   const fixText = isChrome ? "Update Chrome" : "Download Chrome";
-  const continueText = isChrome
-    ? "Continue with unsupported version"
-    : "Continue with unsupported browser";
 
   return (
     <StyledBanner>
       <div>
+        {isDismissable ? (
+          <StyledIconWrapper onClick={() => setShowBanner(false)}>
```
jtbandes, line 50: I think @2metres recently added a HoverableIconButton that you might want to use for this?

2metres: Currently wrapped up in my PR #1466 , hopefully merging today.

web/src/VersionBanner.tsx
defunctzombie: 👍 We should not use native elements.

2metres:
Would the: [MessageBar](https://developer.microsoft.com/en-us/fluentui#/controls/web/messagebar) from fluentui work here?

web/src/VersionBanner.tsx
esthersweon: Discussed; will not need.

esthersweon:
> Is `BigUint64Array` the only browser-compatibility issue we have at this point?
> 
> Maybe I am old school but it feels more sensible to whitelist a set of browsers we have tested with (only Chrome at this stage), given how many different things could potentially cause issues?

@amacneil - There are 3 categories of browsers that we want to have different UIs rendered for: 1) supported browsers (i.e. Chrome v76+), 2) not supported, but somewhat working, browsers (e.g. Chrome below v76, Safari 15, Firefox, etc.), and 3) completely broken browsers (e.g. Safari 14).

Discerning category 1 is easy enough; this check for `BigUint64Array` is an easy way to distinguish between categories 2 and 3 without having to parse the user agent for multiple browsers and to check against multiple corresponding version numbers (e.g. Safari - 15, Firefox - 91, etc.).

amacneil:
@esthersweon thanks for clarification, makes sense. When I wrote that comment earlier today I didn't realize we had two separate behaviors for "completely broken" vs "unsupported" browsers. I think the solution we landed on is good.

PR #1516, opened 2021-07-20
Change: call to action to use desktop version on extension installs

defunctzombie:
**User-Facing Changes**
When a user clicks "install" for an extension in the web version they see an error that indicates extensions are only supported in the desktop version.

**Description**
The web version cannot download extensions at this time so we make the
error message more clear that the user should use the desktop version.

Fixes https://github.com/foxglove/studio/issues/1515

image

### DIFF
--- a/web/src/providers/ExtensionLoaderProvider.tsx
+++ b/web/src/providers/ExtensionLoaderProvider.tsx
@@ -58,7 +58,7 @@ export default function ExtensionLoaderProvider(props: PropsWithChildren<unknown
 }
 
 async function downloadExtension(_url: string): Promise<Uint8Array> {
-  throw new Error("Please use the desktop version to use extensions");
+  throw new Error("Please download the desktop app to use extensions");
 }
 
 async function installExtension(_foxeFileData: Uint8Array): Promise<ExtensionInfo> {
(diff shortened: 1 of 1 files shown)


web/src/providers/ExtensionLoaderProvider.tsx
```diff
@@ -57,14 +57,12 @@ export default function ExtensionLoaderProvider(props: PropsWithChildren<unknown
   );
 }
 
-async function downloadExtension(url: string): Promise<Uint8Array> {
-  const res = await fetch(url);
-  return new Uint8Array(await res.arrayBuffer());
+async function downloadExtension(_url: string): Promise<Uint8Array> {
+  throw new Error("Please use the desktop version to use extensions");
```
defunctzombie, line 61:
```suggestion
  throw new Error("Please download the desktop app to use extensions");
```

web/src/providers/ExtensionLoaderProvider.tsx
```diff
@@ -57,14 +57,12 @@ export default function ExtensionLoaderProvider(props: PropsWithChildren<unknown
...
 }
 
 async function installExtension(_foxeFileData: Uint8Array): Promise<ExtensionInfo> {
-  // The web view can load extensions, but can't install them
-  throw new Error("Extensions cannot be installed from the web viewer");
+  throw new Error("Please use the desktop version to use extensions");
```
defunctzombie, line 65:
```suggestion
  throw new Error("Please download the desktop app to use extensions");
```

amacneil:
:+1: cool

PR #1518, opened 2021-07-21
web: add ?demo url param to force loading welcome layout

### LINKED ISSUE #1508: `?demo` URL param to restore welcome layout + demo connection
We want to link to studio.foxglove.dev from our homepage. We need a way to force it to re-load the demo layout and demo bag.

The most obvious solution would be a `?demo` URL parameter. It would also be nice to allow loading any arbitrary remote bag via a URL param, but we would still need the `?demo` option since we need this to restore the welcome layout too.

jtbandes:
**User-Facing Changes**
None (more for internal/marketing use)

**Description**
Fixes #1508 

If the `?demo` flag is present in the url, always load the welcome layout on first mount. Also do a `history.replaceState` to remove the demo flag.

(Note that this still means creating a _new_ layout each time.)

### DIFF
--- a/web/src/Root.tsx
+++ b/web/src/Root.tsx
@@ -19,7 +19,11 @@ import ExtensionLoaderProvider from "./providers/ExtensionLoaderProvider";
 
 const DEMO_BAG_URL = "https://storage.googleapis.com/foxglove-public-assets/demo.bag";
 
-export function Root(): JSX.Element {
+export function Root({
+  alwaysLoadWelcomeLayoutOnMount,
+}: {
+  alwaysLoadWelcomeLayoutOnMount: boolean;
+}): JSX.Element {
   const playerSources: PlayerSourceDefinition[] = [
     {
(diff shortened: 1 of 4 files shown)


web/src/Root.tsx
```diff
@@ -54,7 +58,11 @@ export function Root(): JSX.Element {
     <ThemeProvider>
       <ErrorBoundary>
         <MultiProvider providers={providers}>
-          <App demoBagUrl={DEMO_BAG_URL} availableSources={playerSources} />
+          <App
+            alwaysLoadWelcomeLayoutOnMount={alwaysLoadWelcomeLayoutOnMount}
```
defunctzombie, line 62: can this be `loadWelcomeLayout` to avoid the always and OnMount which seem not relevant?

jtbandes: always - means "ignore whether we've done it before"
onmount - means "don't expect changing this prop later to do anything"

PR #1520, opened 2021-07-21
Add a disabledReason field to PlayerSourceDefinition

defunctzombie:
**User-Facing Changes**
Web users are shown an alert that indicates they should download the desktop app when trying to access desktop only connections.

Screen Shot 2021-07-20 at 9 42 43 PM

**Description**
The disabledReason field allows components (like the ConnectionList)
to show an error if the particular player is not available on that
platform.

On web, this allows us to direct the user to download the desktop app
which does have the desired connection.

https://github.com/foxglove/studio/issues/1519


defunctzombie:
@jtbandes I refactored useConfirm to match usePrompt (and moved it into the hooks folder). Take a look.

amacneil:
Can you update the language:

> **Desktop app required**
> ROS 1 Native connections are only available in our desktop app. [Download it here](https://foxglove.dev/download).

PR #1521, opened 2021-07-21
Rehome global.scss into relevant places

2metres:


packages/studio-base/src/styles/useGlobalStyles.ts
```diff
@@ -0,0 +1,129 @@
...
+    },
+    ":global(ol, ul)": {
+      listStyle: "none",
+    },
+    ":global(blockquote, q)": {
+      quotes: "none",
+    },
+    ":global(blockquote:before, blockquote:after, q:before, q:after)": {
+      content: "none",
+    },
```
2metres, line 56: 🤔 Inlining `reset.scss`

2metres (commented):
Problems: The storybook still relies on the `global.scss` so we need to sort that out to use the same `makeStyles`

packages/studio-base/src/components/Autocomplete.module.scss
```diff
@@ -27,6 +28,27 @@ $row-height: 24px;
 }
 
 .input {
+  // Re-homed from global.scss
```
defunctzombie, line 31: Don't need this comment about global.scss - this doesn't give any useful context to future readers

packages/studio-base/src/components/Autocomplete.module.scss
```diff
@@ -27,6 +28,27 @@ $row-height: 24px;
...
+  &.disabled {
+    color: $text-input-disabled;
+    background-color: rgba(255, 255, 255, 0.3);
+  }
+  &:focus {
+    background-color: rgba(255, 255, 255, 0.075);
+    outline: none;
+  }
+
+  // styles for autocomplete
```
defunctzombie, line 51: All the styles in this module file are for autocomplete... 🙈

packages/studio-base/src/components/GlobalVariablesTable/index.tsx
```diff
@@ -292,7 +294,7 @@ function GlobalVariablesTable(): ReactElement {
 
   return (
     <SGlobalVariablesTable>
-      <table>
+      <LegacyGlobalTable>
```
defunctzombie, line 297: I would drop the word _Global_ from these components. Maybe `StyledTable` or `LegacyTable`. There is nothing _Global_ about this component. I also think it belongs in the components folder - we should be aiming to get rid of the _styles_ folder or at least keep it free of components.

packages/studio-base/src/util/sharedStyleConstants.ts
```diff
@@ -89,6 +87,43 @@ export const colors = {
   DIFF_MODE_SOURCE_2: robotStylesColors.TEAL,
   DIFF_MODE_SOURCE_BOTH: robotStylesColors.DARK7,
   CYAN: "#00ffff",
+
+  // Re-homed from global.scss
```
defunctzombie, line 91: same notes on global.scss comment as earlier

packages/studio-base/src/components/Autocomplete.module.scss
2metres: OHAY! Indeed they are. Leaving myself a little breadcrumb

2metres: Fixing up now

packages/studio-base/src/Workspace.tsx
```diff
@@ -148,6 +162,7 @@ type WorkspaceProps = {
 };
 
 export default function Workspace(props: WorkspaceProps): JSX.Element {
+  const classes = useStyles();
```
jtbandes, line 165: I think we call these `styles` instead of `classes` in many places – thoughts on which name we should prefer?

packages/studio-base/src/components/CssBaseline.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+import "@foxglove/studio-base/styles/reset.scss";
+
+import { MONOSPACE } from "@foxglove/studio-base/styles/fonts";
+
+export default function CssBaseline(): ReactNull {
+  const theme = useTheme();
+
+  mergeStyles({
+    ":global(body, html, #root)": {
```
jtbandes, line 15: I don't really see the point of making this change if we still have styles that apply globally.

We'd like for the app to be embeddable in another app (e.g. as a subtree (not iframe) inside another React or Angular app) without affecting the styles from outside – so we should strive to not apply any styles globally. I would think we can apply these styles onto our root element, and these selectors will nest inside e.g. we'll have styles for `.myRootElement code` instead of `:global(code)`. Is there a reason we can't do that?

packages/studio-base/src/components/LegacyStyledComponents.tsx
```diff
@@ -0,0 +1,172 @@
...
+
+import styled from "styled-components";
+
+import { MONOSPACE, SANS_SERIF } from "@foxglove/studio-base/styles/fonts";
+import { colors as sharedColors, spacing } from "@foxglove/studio-base/util/sharedStyleConstants";
+
+/**
+ * @deprecated The LegacyButton should not be used for new features. use fluentui/react instead
+ */
+export const LegacyButton = styled.button`
```
jtbandes, line 13: 👍 on pulling these out of global.scss and marking them deprecated

packages/studio-base/src/panels/ImageView/ImageCanvas.tsx
```diff
@@ -350,15 +351,15 @@ export default function ImageCanvas(props: Props): JSX.Element {
     return (
       <div className={styles.zoomContextMenu} data-zoom-menu>
         <div className={cx(styles.menuItem, styles.notInteractive)}>
-          Use mousewheel or buttons to zoom
+          Use mousewheel or button to zoom
```
jtbandes, line 354: I think this was buttons because it refers to the two buttons below. I don't love the wording in general though

packages/studio-base/src/components/LegacyStyledComponents.tsx
```diff
@@ -0,0 +1,172 @@
...
+  &.disabled {
+    color: ${sharedColors.TEXT_INPUT_DISABLED};
+    background-color: rgba(255, 255, 255, 0.3);
+  }
+`;
+
+/**
+ * @deprecated The LegacyGlobalTable should not be used for new features. use fluentui/react instead
+ */
+export const LegacyGlobalTable = styled.table`
```
jtbandes, line 138: why not just LegacyTable?

packages/studio-base/src/styles/reset.scss
```diff
@@ -1,8 +1,21 @@
+/*
+  Make sure everything uses box-sizing: border-box.
+  Per https://www.paulirish.com/2012/box-sizing-border-box-ftw/
+  When changing this, be aware that Mosaic also adds this by default.
```
jtbandes, line 4: sigh - are we getting a bunch of global css from react-mosaic as well?

PR #1522, opened 2021-07-21
Change: remove sendNotification from all randomAccessDataProviders

defunctzombie:
**User-Facing Changes**
Users will see errors surfaced from these data providers via the appropriate player. These errors will clear when the player changes.

**Description**
Rather than using sendNotification, we add problems to initializationResult
and throw where appropriate within individual methods.

Rather than assuming a global notification flow, these non-ui classes expose
their error and allow the callers to handle it for display. This also allows
the callers to clear the error if it is no longer relevant.

I'm also hoping this cuts down on sentry errors related to out-of-bounds bag reads or other possibly corrupt data reading. If users encounter such errors its better for them to see them and surface them to us since they will be able to provide the bag file or data source to help reproduce if it is a bug.


jtbandes:
Is this still something worth pursuing?

defunctzombie:
Yes but it gets stale quickly when I try to tackle all of them. I need to tackle them one by one. Its worth pursuing since it improves the error ux by using the existing data flows rather than the annoying toasts.

PR #1524, opened 2021-07-21
add build command for browser

hakuturu583:
Signed-off-by: Masaya Kataoka <ms.kataoka@gmail.com>

**User-Facing Changes**
I want to build website like https://studio.foxglove.dev/?, but I could not found command, so I add it.

**Description**
Adding build commands for building website.

**How to check**
please type commands below.
```
yarn web
```
connect to http://localhost:8080/

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "build:packages": "tsc --build --verbose packages/**/tsconfig.json",
     "build:prod": "webpack --mode production --progress --config desktop/webpack.config.ts",
     "build:dev": "webpack --mode development --progress --config desktop/webpack.config.ts",
+    "web": "webpack serve --mode development --progress --config web/webpack.config.ts",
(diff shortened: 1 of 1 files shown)


defunctzombie:
We've been doing this by using `yarn serve` in the _web_ directory.

@amacneil should we use this instead? If yes we should probably document it on the README?

amacneil:
I think either have both desktop and web scripts in the root package.json, or neither. 

If we use this approach, let’s align it with the desktop scripts (web:build:dev, web:build:prod, web:serve, etc).

If that feels like too many scripts, we could just remove all these build scripts from root and keep them under appropriate desktop or web packages only?

amacneil:
Thanks! I added the rest of the commands in https://github.com/foxglove/studio/pull/1525

PR #1536, opened 2021-07-22
Remove Layout.path

jtbandes:
**User-Facing Changes**
None

**Description**
Path is not supported in the current Console API, so just remove it.


defunctzombie:
🔪

PR #1543, opened 2021-07-22
Add Amplitude metrics for layout management

### LINKED ISSUE #1412: Add Amplitude metrics for layout management
Add Amplitude metrics for the following events:
- Creating new layout
- Importing / exporting a layout
- Adding a particular panel to a layout (via sidebar or panel splitting / changing)

These would be great for getting a sense of:
- how many layouts users are creating
- how often they're sharing these layouts
- how often certain panels are being used (most popular panels)

esthersweon:
**User-Facing Changes**
- None

**Description**

Added Amplitude metrics for the following events:

- Creating new layout
- Importing / exporting a layout
- Adding a panel to a layout (specify panel type)
- Removing a panel from a layout

Resolves #1412


packages/studio-base/src/Workspace.tsx
```diff
@@ -122,9 +125,22 @@ function Connection() {
 
 function AddPanel() {
   const addPanel = useAddPanel();
+
+  const analytics = useAnalytics();
+  const metricsCollector = useMemo(() => new AnalyticsMetricsCollector(analytics), [analytics]);
```
jtbandes, line 130: Will defer to @jhurliman as the author of the useAnalytics stuff, but I don't think we need the metrics collector here as you can just use the analytics object directly. The metrics collector looks like it's an adapter intended to match the Player interface (but just guessing from a cursory read.)

packages/studio-base/src/Workspace.tsx
```diff
@@ -122,9 +125,22 @@ function Connection() {
...
+
+  const onPanelSelect = useCallback(
```
jtbandes, line 132: There is at least 1 more place where we useAddPanel that this doesn't currently capture.

Consider instead tracking this at the CurrentLayoutProvider or CurrentLayoutState level — for instance CurrentLayoutState could accept the analytics object as a constructor argument

packages/studio-base/src/components/LayoutBrowser/index.stories.tsx
```diff
@@ -74,7 +75,9 @@ function WithSetup(Child: Story, ctx: StoryContext): JSX.Element {
         <CurrentLayoutContext.Provider value={currentLayout}>
           <LayoutCacheContext.Provider value={storage}>
             <CacheOnlyLayoutStorageProvider>
-              <Child />
+              <AnalyticsProvider>
```
jtbandes, line 78: Feel free to add AnalyticsProvider at the top level `preview.tsx` for all stories – I think that would be appropriate

packages/studio-base/src/Workspace.tsx
esthersweon: Commit here: https://github.com/foxglove/studio/pull/1543/commits/7735c08df86352d0096de2288c3864e0bca0c691

packages/studio-base/src/providers/CurrentLayoutProvider/CurrentLayoutState.ts
```diff
@@ -49,16 +52,23 @@ export const DEFAULT_LAYOUT_FOR_TESTS: LayoutState = {
 const LAYOUT_HISTORY_SIZE = 20;
 const LAYOUT_HISTORY_THROTTLE_MS = 1000;
 
-export default class CurrentLayoutState implements ICurrentLayout {
+const CurrentLayoutState = (initialState: LayoutState): CurrentLayoutStateWithAnalytics => {
+  const analytics = useAnalytics();
```
jtbandes, line 56: Would prefer not to introduce a hook call here since this file is pretty much non-React. This happens to be a non-standard use of hooks because normally they would only be used in a function component (something that returns JSX.Element) meaning that the callers would need to use `<CurrentLayoutState />` instead of `CurrentLayoutState()` to get the correct behavior. The `rules-of-hooks` lint rule doesn't complain because the function's name starts with a capital letter, so it assumes it is a function component — doesn't look at the call sites to see if they are actually using it as a component.

It's not safe to call this function from other React components because the nested hook call useAnalytics() may happen conditionally or out of order (since this function itself doesn't start with `use`, the lint rule won't prevent us from doing this elsewhere). And calling it from non-React code probably fails when it internally tries to do a useContext.

I think a simpler solution would be to just make the analytics param optional: `constructor(..., analytics?: Analytics)` and call useAnalytics wherever the CurrentLayoutState is being constructed in a useMemo, and don't pass in analytics for unit tests.

packages/studio-base/src/providers/CurrentLayoutProvider/CurrentLayoutState.ts
```diff
@@ -154,6 +158,7 @@ export default class CurrentLayoutState implements ICurrentLayout {
     createTabPanel: (payload: CREATE_TAB_PANEL["payload"]): void => {
       this.dispatch({ type: "CREATE_TAB_PANEL", payload });
       this.setSelectedPanelIds([]);
+      this.analytics!.logEvent(AppEvent.LAYOUT_ADD_PANEL, { type: "Tab" });
```
jtbandes, line 161: These will need to be `?.` rather than `!.` so they don't crash in tests where analytics is not defined. `!` should only be used when we really know for certain that the value is present but can't get that information into the TypeScript compiler (such as checking array lengths and then accessing items at an index).

PR #1546, opened 2021-07-22
Remove global styles as squashed commit

2metres:
Hoping that this fixes some of my chromatic diff baselines to compare against main


defunctzombie:
@2metres I pushed a commit which fixes the TimeBasedChartTooltip styling. It was largely a problem with conflicting styling in LegacyTable - another reason to do away with all these global evils.

2metres:
Massive amount of changes here, most of them just interim solutions until we can phase out <Dropdown>, <Button> and <TextField> to use fluentui components

defunctzombie:
Good stuff. LGTM!

amacneil:
Can you update PR title + description for PRs like this (I assume the actual description was in another PR somewhere). It's helpful for writing release notes and historical reference.

Issue #1491, opened 2021-07-19
Direct user to website to create orgs


defunctzombie:
During the sign-in flow a user must select their organization. If there are no organizations for the user to select, direct the user to the foxglove website to create their organization. This avoids us having to create org creation flows within the app.

This is not the ideal workflow since the user will have authorized their device before being kicked _back_ to the browser but this works around some auth0 limitations.

PR #1548, opened 2021-07-23
Change: signin flow uses a dialog

defunctzombie:
**User-Facing Changes**

Screen Shot 2021-07-22 at 6 08 49 PM
Screen Shot 2021-07-22 at 6 09 22 PM

**Description**
A dialog feels more natural for the device code flow by helping
the user focus on the signin process rather than hiding it in the sidebar.
    
This change also adjusts the org selection behavior. If the user does not have
any orgs, they are presented with a notice to have their org admin invite them
or to create an org on the website.
    
Fixes: #1491


packages/studio-base/src/components/AccountSettingsSidebar/DeviceCodeDialog.tsx
```diff
@@ -0,0 +1,211 @@
...
+                key={org.id}
+                onClick={async () => await onOrgSelect(org)}
+                className={classes.orgItem}
+              >
+                <Text variant="large">{org.display_name ?? org.slug}</Text>
+              </StackItem>
+            );
+          })}
+        </Stack>
+      );
```
jtbandes, line 162: Dropdown? Stack of links seems a bit non-standard

defunctzombie: I think the slack login flow presents a stack of links. Kinda like a nav menu/context menu list.

PR #1550, opened 2021-07-23
Improve ErrorBoundary appearance

jtbandes:
User-facing changes: not worth mentioning

Bring some FluentUI love to ErrorBoundary. Also integrate better with PanelToolbar for panel errors.

Still could use some spacing/alignment cleanup but I think this can wait until the global.scss changes settle down.


defunctzombie:
👍

PR #1558, opened 2021-07-23
Rename rosbag.js to @foxglove/rosbag

amacneil:
**User-Facing Changes**
None

**Description**
Consistent with @foxglove/rosbag2


defunctzombie:
Should be @foxglove/rosbag since that is what we will publish?

defunctzombie:
I see this is for your rename of the repo first. Also rename the package?

amacneil:
Ya will rename package. I could do that in the same PR I guess.

PR #1571, opened 2021-07-28
Internal: Provide a default NullAnalytics context

defunctzombie:
**User-Facing Changes**
None

**Description**
Allows for tests and stories to use components that have analytics
without having to inject analytics providers. Lowers the boilerplate
for writing the tests and stories.

### DIFF
--- a/packages/studio-base/src/services/NullAnalytics.ts
+++ b/packages/studio-base/src/services/NullAnalytics.ts
@@ -1,0 +1,9 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import IAnalytics from "@foxglove/studio-base/services/IAnalytics";
+
+export default class NullAnalytics implements IAnalytics {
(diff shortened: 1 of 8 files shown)


packages/studio-base/src/services/NullAnalytics.ts
```diff
@@ -0,0 +1,9 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import IAnalytics from "@foxglove/studio-base/services/IAnalytics";
+
+export default class NullAnalytics implements IAnalytics {
```
jtbandes, line 7: naming: NoOpAnalytics?

defunctzombie: I drew inspiration from /dev/null  - do you prefer NoOpAnalytics? Can make the change if you prefer that.

jtbandes: I guess we used to have some other NoOp things but they were removed as we implemented shared layouts, though we do still have a "NoopMetricsCollector".

PR #1572, opened 2021-07-28
Internal: use stable selector function for useConfigById

defunctzombie:
**User-Facing Changes**
None

**Description**
Avoid changing the selector function on every useConfigById invocation.


jtbandes:
Was this a performance issue?

defunctzombie:
@jtbandes no - just some random stuff I saw while trying to debug why we were seeing errors when changing layouts. Probably could have done without it but I had the change sitting around.

PR #1573, opened 2021-07-28
Internal: Fix update state after unmount errors when changing layouts

defunctzombie:
**User-Facing Changes**
None

**Description**
When changing layouts, the useCurrentLayoutSelector would fire state
updates (via the listener) into components that are unmounted. Removing
the listener on layout effect cleanup was happening too late
to prevent the listener from being invoked.

The errors no longer appear in dev console when changing layouts.

### DIFF
--- a/packages/studio-base/src/context/CurrentLayoutContext/index.ts
+++ b/packages/studio-base/src/context/CurrentLayoutContext/index.ts
@@ -120,7 +120,13 @@ export function useCurrentLayoutSelector<T>(selector: (layoutState: LayoutState)
     };
   }
   useLayoutEffect(() => {
+    let mounted = true;
     const listener = (layoutState: LayoutState) => {
+      // Note: Our removeLayoutStateListener is is too late if the layout state listeners are already
+      // being invoked. Our component might become unmounted during the state listener invocation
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/context/CurrentLayoutContext/index.ts
```diff
@@ -120,7 +120,13 @@ export function useCurrentLayoutSelector<T>(selector: (layoutState: LayoutState)
     };
   }
   useLayoutEffect(() => {
+    let mounted = true;
```
jtbandes, line 123: why not `useMountedState`?

jtbandes (commented):
Interesting, I assumed this error might have been on a per panel basis, but it's nice that there is a centralized fix.

packages/studio-base/src/context/CurrentLayoutContext/index.ts
defunctzombie: It didn't work. I couldn't quite figure out why but it seemed that the mounted state didn't realize it was unmounted in time and thus didn't actually work as expected. I think it suffers from effect execution order :(

PR #1575, opened 2021-07-29
Figure out which component handles layout deletions and new layouts

defunctzombie:
**User-Facing Changes**

**Description**

### DIFF
--- a/packages/studio-base/src/providers/CurrentLayoutProvider/index.tsx
+++ b/packages/studio-base/src/providers/CurrentLayoutProvider/index.tsx
@@ -4,7 +4,7 @@
 import { isEqual } from "lodash";
 import { useEffect, useLayoutEffect, useRef, useState } from "react";
 import { useToasts } from "react-toast-notifications";
-import { useAsync, useMountedState, useThrottle } from "react-use";
+import { useAsync, useThrottle } from "react-use";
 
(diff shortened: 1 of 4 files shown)


defunctzombie:
An annoyance I've run into when working on this PR is that it isn't clear which component is actually responsible for setting up a layout when one is deleted. The LayoutBrowser is fighting the CurrentLayoutProvider.

packages/studio-base/src/providers/CurrentLayoutProvider/index.tsx
```diff
@@ -54,20 +54,6 @@ function CurrentLayoutProviderWithInitialState({
...
   );
 
-  const isMounted = useMountedState();
-
-  // If the current layout is deleted, deselect it
-  useEffect(() => {
-    const listener = async () => {
-      const selectedId = stateInstance.actions.getCurrentLayoutState().selectedLayout?.id;
-      if (!(await layoutStorage.getLayouts()).some(({ id }) => id === selectedId) && isMounted()) {
-        stateInstance.actions.setSelectedLayout(undefined);
```
jtbandes, line 64: If we don't setSelectedLayout(undefined) then you end up seeing your panels in the mosaic with no selected layout in the sidebar – and a stale layout id in memory. And attempting to save any changes will fail because of the stale id.

Issue #1576, opened 2021-07-29
merge ILayoutStorage / IRemoteLayoutStorage?


defunctzombie:
Rather than two different types of storage. Enhance our layout storage apis to support the responses we need for remote layout storage layers.

jtbandes:
Closing for now, not clear that this is useful – it's valuable to have richer responses from the server and for ILayoutStorage to insert additional info that only it knows about (conflict info and unsynced changes)

Issue #1528, opened 2021-07-21
Team Layouts Private Beta


jtbandes:
Requirements:

- Personal layouts (when not signed in) track "baseline" and "dirty" versions
- Personal layouts (when signed in) can be read and created/renamed/updated/deleted while offline
  - Dirty layout is not synced to server
  - Multiple personal layouts can be dirty at once, indicated by some icon next to the layout
  - Sync to server is based on baseline last saved timestamp (using client time)
- All Team layouts can be read while offline, but must be online to create/rename/update/delete
  - Dirty version of team layouts is stored locally (not synced to server)
  - Changing away from a dirty team layout results in a prompt to save/discard changes

Tasks remaining to ship team layouts:

- [x] merge #1425 (feature flagged)
- [x] #1536
- [x] #1547
   - [x] ~#1576~
   - [x] https://github.com/foxglove/studio/pull/1613
   - [x] ~simplify or remove conflict types in favor of simpler "last write wins" strategy?~
- [x] #1607
- [x] #1544
- [x] #1412
- [x] #1491
- [x] #1054
- [x] #1575
- [x] #1545
- [x] Store layout baselines (and proactively fetch from server) to allow revert operation (should work when logged out too) - https://github.com/foxglove/studio/pull/1679
   - "save" operation updates baseline while offline, which can still lead to conflicts on next sync
   - Change `state` -> `data`
- [x] Migration from old data storage format
- [x] https://github.com/foxglove/studio/pull/1742
  - [x] When switching away from a shared layout, prompt to save as a copy / update template
  - [x] namespace the cache by userid/org for better signout experience?
  - [x] #1620
  - [x] https://github.com/foxglove/console/issues/106
  - [x] https://github.com/foxglove/console/issues/107
- [x] #1743
- [x] https://github.com/foxglove/studio/issues/1769
- [x] #1746
- [x] #1747
- [x] https://github.com/foxglove/studio/pull/1808
- [x] https://github.com/foxglove/studio/pull/1809
- [x] #1792
- [x] https://github.com/foxglove/studio/pull/1811
- [x] https://github.com/foxglove/studio/pull/1812
- [x] https://github.com/foxglove/studio/issues/1748
- [x] https://github.com/foxglove/studio/issues/1819
- [x] #1646
- [x] https://github.com/foxglove/studio/pull/1844
- [x] #1744

PR #1580, opened 2021-07-30
Consolidate layout interfaces

defunctzombie:
**User-Facing Changes**
TBD

**Description**
Rather than having three layout storage mechanisms (ILayoutCache, ILayoutStorage, IRemoteLayoutStorage), this change consolidates them into one ILayoutStorage. This reduces the indirection and simplifies the logic around cloud and local layouts.

- NativeStorageLayoutCache is now NativeStorageLayoutStorage and implements ILayoutStorage. It is backwards compatible with our existing on-disk layout storage methodology.
- CacheOnlyLayoutStorageProvider, ConsoleApiLayoutStorageProvider, FakeLayoutStorageProvider are removed in favor of ConsoleLayoutStorageProvider which determines if team layouts is enabled and a user is signed in to provide either an OfflineLayoutStorage instance or pass along the existing ILayoutStorage instance.
- OfflineLayoutStorage implements ILayoutStorage on top of an existing ILayoutStorage as an _offline_ storage and a ConsoleApi.
- Remove sync api from ILayoutStorage in favor of a "reset" option on `updateLayout` which indicates the layout storage should treat the update as an update on the "reset" state of the layout (I've not come up with a better term for this yet but this is the layout that a user could "go back to" even after making local changes. It is up to the ILayoutStorage whether it supports this.

Resolves some items from: #1528
- Remove fake layout storage
- Fixes: #1576
- Simplify or remove conflict types in favor of simpler "last write wins" strategy

Remaining tasks:
- various fixme comments
- OfflineLayoutStorage tests
- "startup/sync" for OfflineLayoutStorage that runs once when initializing and syncs offline with online


defunctzombie:
@jtbandes feel free to close and delete the branch this once it is no longer useful.

jtbandes:
Closing since many initial items from https://github.com/foxglove/studio/issues/1528 have been addressed and we have some new directions we need to go in – will continue thinking about this along the way to v1.

PR #1583, opened 2021-07-30
Use https for apt repo

amacneil:
**User-Facing Changes**
None

**Description**
Use https for apt repo (http currently returns a 301 anyway so this just avoids the redirect).


defunctzombie:
I thought we allowed for plain http on the apt repo? I know that historically https apt repo required more packages installed that were not present by default. Maybe these days they are present and it is assumed to work.

amacneil:
Interesting. That's probably true in a docker container or something. Don't think it's an issue on a desktop install. Most of my non-ubuntu apt repos are using https:

```
~ $ sudo apt-get update|grep https
Hit:3 https://cli.github.com/packages stable InRelease
Hit:4 https://dl.yarnpkg.com/debian stable InRelease
Hit:5 https://packages.microsoft.com/repos/ms-teams stable InRelease
Get:1 https://apt.foxglove.dev/studio stable InRelease [5,656 B]
Hit:7 https://deb.nodesource.com/node_16.x focal InRelease
Hit:8 https://download.docker.com/linux/ubuntu focal InRelease
Hit:6 https://packagecloud.io/github/git-lfs/ubuntu focal InRelease

```

amacneil:
I tested in an `ubuntu:20.04` docker container - it gives you an error suggesting to install `ca-certificates`, then after you do it works fine. So I think that is a reasonable dependency for anyone installing a gui tool.

PR #1590, opened 2021-08-03
Change: scope app styling with css baseline container

defunctzombie:
**User-Facing Changes**
None

**Description**
Rather than setting styles globally on elements, scope styling
to a top level app container. The CssBaseline component should be
used as this top level app container.


desktop/renderer/index.tsx
```diff
@@ -48,6 +49,21 @@ if (!rootEl) {
...
+    width: "100%",
+    display: "flex",
+    flexDirection: "column",
+    position: "relative",
+    flex: "1 1 100%",
+    outline: "none",
+    overflow: "hidden",
+  },
+});
+
```
2metres, line 66: We should probably consider name spacing our root selector something like `#foxglove-root`

2metres (changes requested):
The typography is currently broken in chromatic and the background is now included

desktop/renderer/index.tsx
defunctzombie: Any benefit to this since we don't reference it in studio-base?

defunctzombie:
TextContent has a line height of 1.65 - this causes too much space between lines.

PR #1592, opened 2021-08-04
Sync layouts periodically when logged in & online

### LINKED ISSUE #1544: run sync periodically when online and surface errors
When signed in, a user's layouts should sync with their online versions.

Considerations:
- What happens when the online version has changes and the offline version has changes?

jtbandes:
**User-Facing Changes**
None (behind feature flag)

**Description**
Resolves #1544

Sync periodically when all of these are true:
- logged in (currentUser is present)
- online (just using navigator.onLine as a rough status indicator – there may still be situations where the network connection is poor or the server is down)
- app is visible (document.visibilityState)

### DIFF
--- a/packages/studio-base/src/providers/ConsoleApiLayoutStorageProvider.tsx
+++ b/packages/studio-base/src/providers/ConsoleApiLayoutStorageProvider.tsx
@@ -4,8 +4,9 @@
 
 import { useCallback, useMemo } from "react";
 import { useToasts } from "react-toast-notifications";
+import { useInterval, useNetworkState } from "react-use";
 
-import { useShallowMemo } from "@foxglove/hooks";
+import { useShallowMemo, useVisibilityState } from "@foxglove/hooks";
 import Logger from "@foxglove/log";
 import { AppSetting } from "@foxglove/studio-base/AppSetting";
 import { useConsoleApi } from "@foxglove/studio-base/context/ConsoleApiContext";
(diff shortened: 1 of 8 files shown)


packages/studio-base/src/providers/ConsoleApiLayoutStorageProvider.tsx
```diff
@@ -20,6 +21,8 @@ import OfflineLayoutStorage from "@foxglove/studio-base/services/OfflineLayoutSt
 
 const log = Logger.getLogger(__filename);
 
+const SYNC_INTERVAL = 15_000;
```
defunctzombie, line 24: nice to expose as a prop on the component with a default value?

packages/studio-base/src/providers/ConsoleApiLayoutStorageProvider.tsx
```diff
@@ -40,15 +43,30 @@ export default function ConsoleApiLayoutStorageProvider({
...
   }, [addToast, offlineStorage]);
-  const debugging = useShallowMemo({ syncNow });
+
+  const { online = false } = useNetworkState();
+  const visibilityState = useVisibilityState();
+
+  // Sync periodically when logged in, online, and the app is not hidden
+  useInterval(
+    sync,
+    currentUser && online && visibilityState === "visible"
```
defunctzombie, line 64: I have a preference that such boolean and tertiary conditions be moved out of the inline function call and put into a variable. This tends to get out of hand and I've definitely struggled to read these in our codebase.

PR #1593, opened 2021-08-04
Move account sidebar under Team Shared Layouts feature flag & make feature flag available in prod

jtbandes:
**User-Facing Changes**
Adds a new feature flag for Team Shared Layouts.

**Description**
Related to #1528 

Previously the flag was only available in dev builds; now it is available in prod builds.

### DIFF
--- a/packages/studio-base/src/components/ExperimentalFeatureSettings.tsx
+++ b/packages/studio-base/src/components/ExperimentalFeatureSettings.tsx
@@ -48,6 +48,11 @@ const features: Feature[] = [
     name: "Legacy Plot Panel",
     description: <>Enable the legacy plot panel.</>,
   },
+  {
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/components/ExperimentalFeatureSettings.tsx
```diff
@@ -48,6 +48,11 @@ const features: Feature[] = [
     name: "Legacy Plot Panel",
     description: <>Enable the legacy plot panel.</>,
   },
+  {
+    key: AppSetting.ENABLE_CONSOLE_API_LAYOUTS,
+    name: "Team shared layouts",
```
defunctzombie, line 53: Team Layouts?

Other entries have capitalized each word so maybe Team Shared Layouts?

PR #1596, opened 2021-08-05
Fail CI on unused exports

jtbandes:
**User-Facing Changes**
None

**Description**
Unused exports often point to dead code. Linting these in CI will help us remove unused code as it becomes unused.


jtbandes:
Example error:
image

defunctzombie:
@jtbandes we should delete it - the extension api uses a different api for the config. I think we just never fully ported things over to the `useConfig` without a panel id.

amacneil:
Could this be done as an eslint rule rather than a separate script? That would be much cleaner for running it, using across projects, etc.

Similar request: https://github.com/typescript-eslint/typescript-eslint/issues/371

jtbandes:
https://github.com/typescript-eslint/typescript-eslint/pull/1905

PR #1598, opened 2021-08-05
Feature: add advertise and publish to the extension api

defunctzombie:
## User-Facing Changes
Panels authored via the extension api are able to advertise datatypes on topics.
Panels authored via the extension api are able to publish messages on advertised topics.

### Advertise
The advertise method on the panel extension context allows a panel to indicate an intent to publish a specific datatype on a topic. The panel would need to call advertise before being able to publish on the topic.

```typescript
context.advertise(topic: string, datatype: string, options?: Record<string, unknown>);
```

Options are specific to the current data source. Some data sources make use of specific options while others do not.

### Unadvertise 
Remove advertising for a topic/datatype pair.

### Publish
The publish function publishes a message on a previously advertised topic. If the topic is not advertised or otherwise malformed, the function will throw.

```typescript
context.publish(topic: string, message: Record<string, unknown>);
```

This function is fire-and-forget.

## Description
The user facing changes capture the scope of the PR.

### DIFF
--- a/packages/studio-base/src/components/PanelExtensionAdapter.tsx
+++ b/packages/studio-base/src/components/PanelExtensionAdapter.tsx
@@ -24,7 +24,11 @@ import {
   useHoverValue,
   useSetHoverValue,
 } from "@foxglove/studio-base/context/HoverValueContext";
-import { PlayerCapabilities, PlayerState } from "@foxglove/studio-base/players/types";
+import {
+  AdvertiseOptions,
+  PlayerCapabilities,
+  PlayerState,
+} from "@foxglove/studio-base/players/types";
 import { SaveConfig } from "@foxglove/studio-base/types/panels";
 import { fromSec, toSec } from "@foxglove/studio-base/util/time";
 
@@ -59,6 +63,8 @@ function selectSeekPlayback(ctx: MessagePipelineContext) {
   return ctx.seekPlayback;
 }
 
+type RenderFn = (renderState: Readonly<RenderState>, done: () => void) => void;
+
 /**
  * PanelExtensionAdapter renders a panel extension via initPanel
  *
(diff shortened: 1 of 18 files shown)


packages/studio-base/src/panels/Publish/PublishPanel.tsx
```diff
@@ -0,0 +1,202 @@
...
+
+    // fixme - is the info enough to advertise or do we need the list of datatypes?
+    // actually need the message definition in full
+    const info = datatypeInfo?.get(selectedDatatype);
+    if (!info) {
+      return;
+    }
+
+    const topicName = selectedTopic;
+    context.advertise(topicName, selectedDatatype, info);
```
jtbandes, line 151: why do we need to pass in info if it just comes from datatypeInfo which itself comes from the context?

packages/studio/index.d.ts
```diff
@@ -16,6 +16,14 @@ declare module "@foxglove/studio" {
     datatype: string;
   };
 
+  export type DatatypeInfo = {
+    /** The type of datatype info. This is specific to each data source. */
+    type: string;
+
+    /** Additional arguments based on the type */
+    [key: string]: unknown;
+  };
```
jtbandes, line 25: `{ type: "ros1", datatypes: RosDatatypes } | {type: string; [key: string]: unknown}` ?

packages/studio-base/src/components/PanelExtensionAdapter.tsx
```diff
@@ -287,11 +309,64 @@ function PanelExtensionAdapter(props: PanelExtensionAdapterProps): JSX.Element {
...
+        rosDatatypes[datatype] = datatypeInfo.detail as RosDatatype;
+
+        ctx.setPublishers(subscriberId, [
+          {
+            topic,
+            datatype,
+            datatypes: rosDatatypes,
+            advertiser: { type: "panel", name: subscriberId },
+          },
+        ]);
```
jtbandes, line 335: may need to be updated to handle multiple advertise calls per panel?

defunctzombie:
I've removed the DatatypeInfo and reworking Publish panel to extensions api aspects of this PR to focus on advertise and publish apis only. Those will be enough to support the new control panel which I'll update on top of this api once we merge.

esthersweon:
Friendly reminder to add `docs` label to PRs like this.

PR #1601, opened 2021-08-05
Upgrade rosbag

jtbandes:
**User-Facing Changes**
Fixed an error when loading message data with numeric characters in package names, for example `tf2_msgs/TFError`.

**Description**
Includes the fix from https://github.com/foxglove/rosmsg/pull/10. Also includes various refactors for compatibility with new Promise-based rosbag API and new package structure.


defunctzombie:
LGTM code wise - something we need to fix in rosbag repo for the types?

PR #1609, opened 2021-08-07
Ban use of any in conditionals

amacneil:
**User-Facing Changes**
None

**Description**
Set `allowAny: false` in `@typescript-eslint/strict-boolean-expressions`. Create list of overrides for offending files (unfortunately this will downgrade _all_ strict boolean violations in these files (not just allowAny issues) to warnings, but the override list shouldn't be around for long).


defunctzombie:
👍 might be good to follow this with a ticket to fixup these files so we don't forget to come back around to that work.

PR #1626, opened 2021-08-12
Checkout PR head sha

amacneil:
**User-Facing Changes**
None

**Description**
Alternative to https://github.com/foxglove/studio/pull/1625. Instead of checking out the merge commit, check out the PR head (specified by `sha` instead of `ref` as directed by the [actions/checkout docs](https://github.com/actions/checkout#Checkout-pull-request-HEAD-commit-instead-of-merge-commit)).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,7 +49,7 @@ jobs:
       - uses: actions/checkout@v2
         with:
           # Chromatic requires running build on the branch head, not PR merge commit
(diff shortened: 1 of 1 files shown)


jtbandes:
Didn't this cause problems with Chromatic?

amacneil:
This PR just changes from using ref to sha, so I think chromatic will be happy. The issues with chromatic stem from using the PR merge commit (which I tested removing in https://github.com/foxglove/studio/pull/1625 - and it did seem to cause chromatic to get confused about where the base is).

PR #1636, opened 2021-08-13
Fixes for ros2 message parsing and deserialization

jtbandes:
**User-Facing Changes**
Fixed some bugs with ROS 2 data sources, and improved error reporting.

**Description**
Pulls in the following fixes:
- https://github.com/foxglove/rosmsg/pull/12
- https://github.com/foxglove/cdr/pull/7

Also updates RosbridgePlayer to properly expose errors from message definition parsing and message deserialization as PlayerProblems:

image


packages/studio-base/src/players/PlayerProblemManager.ts
```diff
@@ -0,0 +1,49 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { PlayerProblem } from "@foxglove/studio-base/players/types";
+
+/**
+ * Manages a set of PlayerProblems keyed by ID.
```
defunctzombie, line 8: Maybe worth commenting here that this class aims to behave as an immutable - i.e. return same instance when unchanged, and new instances when changed and roughly why we do that.

defunctzombie (approved):

defunctzombie:
cool

PR #1642, opened 2021-08-16
Handle invalid point cloud byte offsets

jhurliman:
This adds a check to see if point cloud field offsets point past the end of the byte array. If so, an error is surfaced to the user and the point cloud is not rendered.

### DIFF
--- a/packages/studio-base/src/panels/ThreeDimensionalViz/commands/PointClouds/decodeMarker.ts
+++ b/packages/studio-base/src/panels/ThreeDimensionalViz/commands/PointClouds/decodeMarker.ts
@@ -18,7 +18,6 @@ import {
   ColorMode,
   PointCloudSettings,
 } from "@foxglove/studio-base/panels/ThreeDimensionalViz/TopicSettingsEditor/PointCloudSettingsEditor";
-import sendNotification from "@foxglove/studio-base/util/sendNotification";
 
 import {
   getFieldOffsetsAndReaders,
@@ -26,7 +25,6 @@ import {
   createColorBuffer,
   getVertexCount,
   getVertexValue,
-  FieldOffsetsAndReaders,
 } from "./buffers";
 import { PointCloudMarker, VertexBuffer } from "./types";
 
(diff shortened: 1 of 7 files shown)


packages/studio-base/src/panels/ThreeDimensionalViz/commands/PointClouds/readers.ts
```diff
@@ -120,6 +120,10 @@ export function getReader(
   datatype: number,
   offset: number,
 ): FieldReader | undefined {
+  if (offset >= data.length) {
+    throw new Error(`Point cloud data offset ${offset} starts past ${data.length} byte buffer`);
+  }
```
jtbandes, line 125: What's the point of this rethrow – just a nicer error message? Wouldn't the stack trace already be good enough to point us to the fact that it was a point cloud?

packages/studio-base/src/panels/ThreeDimensionalViz/commands/PointClouds/decodeMarker.ts
```diff
@@ -53,7 +55,14 @@ export function decodeMarker(marker: PointCloudMarker): DecodedMarker {
...
     data,
   } = marker;
-  const offsetsAndReaders = getFieldOffsetsAndReaders(data, fields);
+  let offsetsAndReaders: FieldOffsetsAndReaders | undefined;
+  try {
+    offsetsAndReaders = getFieldOffsetsAndReaders(data, fields);
+  } catch (err) {
+    sendNotification("Point cloud decoding failed", err, "user", "error");
+    return undefined;
+  }
```
jtbandes, line 64: It seems more appropriate for this function to throw an error than to `sendNotification`, since the function's job is purely decoding data and it shouldn't have external dependencies.

packages/studio-base/src/panels/ThreeDimensionalViz/commands/PointClouds/readers.ts
jhurliman: These error messages are surfaced to the user. A generic message about DataView being out of bands doesn't help the user troubleshoot their broken sensor_msgs/PointCloud2 payloads.

jtbandes: I see, I guess I assumed you had this issue crop up based on some bag/tcp parsing issue – but it was actually an incorrectly constructed PointCloud2 payload?

jtbandes (approved):
👍

PR #1644, opened 2021-08-16
Create README.md for @foxglove/studio-base

esthersweon:
**User-Facing Changes**
None in-app.

**Description**
- Added README for newly published `@foxglove/studio-base` package.
- Added correct GitHub URLs in package.json files for both `@foxglove/studio` and `@foxglove/studio-base`.

### DIFF
--- a/packages/studio-base/README.md
+++ b/packages/studio-base/README.md
@@ -1,0 +1,46 @@
+# @foxglove/studio-base &nbsp; [![npm version](https://img.shields.io/npm/v/@foxglove/studio-base.svg?style=flat)](https://www.npmjs.com/package/@foxglove/studio-base)
+
+This package contains core components used in the [Foxglove Studio](https://github.com/foxglove/studio) app.
+
+## Installation
+
+```
+npm install @foxglove/studio-base
+```
+
+or
+
+```
+yarn add @foxglove/studio-base
+```
+
+## Quick start
+
+This package includes the following exports:
+
+- Components – e.g. `ColorPicker`, `Dropdown`, `Menu`, etc.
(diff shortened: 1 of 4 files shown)


packages/studio-base/README.md
```diff
@@ -0,0 +1,5 @@
+# @foxglove/studio-base 
+
+This package contains core components used in the [Foxglove Studio](https://github.com/foxglove/studio) app.
+
+This includes React components (e.g. `App`), events (e.g. `NativeAppMenuEvent`), and contexts (e.g. `AppConfigurationContext`) with providers (e.g. `AppConfigurationContext.Provider`).
```
jtbandes, line 5: Not sure if events is the right thing to call out here – maybe hooks? (e.g. `useNativeAppMenuEvent`) cc @defunctzombie 

I'm not a huge fan of four `e.g.`s in the same sentence. I'm actually wondering if we need to call out these specific pieces at all — maybe it's better to either provide a high-level explanation, or an actual usage example, or both?

packages/studio-base/README.md
```diff
@@ -15,3 +15,8 @@ or
 ```
 yarn add @foxglove/studio-base
 ```
+
+## Stay in touch
+
+Join our [Slack channel](https://foxglove.dev/join-slack) to ask questions, share feedback, and stay up to date on what our team is working on.
```
jtbandes, line 21: nit: slack calls their orgs "workspaces", channel isn't quite the right word here

packages/studio-base/README.md
esthersweon: Should've opened this in draft mode - didn't know people would be so fast to review :) 

Added some sections that reflect our other package READMEs.

PR #1645, opened 2021-08-16
Don't send "ResizeObserver loop limit exceeded" errors to Sentry, increase breadcrumb limit back to 100

jhurliman:


desktop/renderer/index.tsx
```diff
@@ -36,7 +36,8 @@ if (isCrashReportingEnabled && typeof process.env.SENTRY_DSN === "string") {
         return integration.name !== "Breadcrumbs";
       });
     },
-    maxBreadcrumbs: 10,
+    beforeSend: (event, _hint?) =>
+      event.message === "ResizeObserver loop limit exceeded" ? ReactNull : event,
```
amacneil, line 40: looks like `ignoreErrors` is a more idiomatic way of doing this?

https://docs.sentry.io/platforms/javascript/guides/wasm/configuration/filtering/

jtbandes: ReactNull is meant to be used for React APIs but this is a Sentry API. If it has to be null rather than undefined you can just add an eslint suppression

amacneil:
@jhurliman how did you test this? still happening in v0.16

PR #1652, opened 2021-08-17
Internal: Don't report autoUpdater network errors to Sentry

jhurliman:
This checks the type of error caught during the auto-update version retrieval, and if it's network related just log a console warning instead of reporting to Sentry.

User-facing changes: None


defunctzombie:
Please remember to note that there are no user facing changes and write a description. Helps the release makers.

PR #1657, opened 2021-08-17
Feature: add dataSourceType to extension api RenderState

defunctzombie:
**User-Facing Changes**
Extension authors can subscribe to changes on dataSourceType to conditionally support different data sources.

**Description**
Some panels need to configure support for different data sources in different ways. For example the Teleop panel
needs to know whether the data source is ros1 or ros2 to select the correct datatype for publishing.

This change adds a `dataSourceType` field to RenderState. This field is a freeform string which the datasource is able
to populate with an arbitrary value. It is up to the data source to specify which values it provides and up to panels to
handle those values.

### DIFF
--- a/packages/studio-base/src/players/RosbridgePlayer.ts
+++ b/packages/studio-base/src/players/RosbridgePlayer.ts
@@ -90,6 +90,7 @@ export default class RosbridgePlayer implements Player {
   private _hasReceivedMessage = false;
   private _presence: PlayerPresence = PlayerPresence.NOT_PRESENT;
   private _problems = new PlayerProblemManager();
+  private _rosVersion: 1 | 2 = 1;
 
   constructor({
     url,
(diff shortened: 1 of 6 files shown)


packages/studio-base/src/players/RosbridgePlayer.ts
```diff
@@ -90,6 +90,7 @@ export default class RosbridgePlayer implements Player {
   private _hasReceivedMessage = false;
   private _presence: PlayerPresence = PlayerPresence.NOT_PRESENT;
   private _problems = new PlayerProblemManager();
+  private _rosVersion: 1 | 2 = 1;
```
jtbandes, line 93: what are the consequences of this changing partway through? can we just make it undefined until it is detected?

defunctzombie: Changing it part way through will result in the next emit state to have the new value. Yea - we could start this off as undefined instead.

PR #1659, opened 2021-08-17
Configure SentryWebpackPlugin to associate commits with the release

jtbandes:
**User-Facing Changes**
None

**Description**
Trying to get commits associated with our Sentry releases, as described in: https://docs.sentry.io/product/releases/suspect-commits/

### DIFF
--- a/desktop/webpack.renderer.config.ts
+++ b/desktop/webpack.renderer.config.ts
@@ -39,9 +39,10 @@ export default (env: unknown, argv: WebpackArgv): Configuration => {
         org: process.env.SENTRY_ORG,
         project: process.env.SENTRY_PROJECT,
         release: `${process.env.SENTRY_PROJECT}@${packageJson.version}`,
-        setCommits: process.env.SENTRY_CURRENT_COMMIT
-          ? { repo: "foxglove/studio", commit: process.env.SENTRY_CURRENT_COMMIT }
-          : undefined,
+        setCommits:
+          process.env.SENTRY_REPO && process.env.SENTRY_CURRENT_COMMIT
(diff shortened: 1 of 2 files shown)


desktop/webpack.renderer.config.ts
```diff
@@ -39,6 +39,9 @@ export default (env: unknown, argv: WebpackArgv): Configuration => {
         org: process.env.SENTRY_ORG,
         project: process.env.SENTRY_PROJECT,
         release: `${process.env.SENTRY_PROJECT}@${packageJson.version}`,
+        setCommits: process.env.SENTRY_CURRENT_COMMIT
+          ? { repo: "foxglove/studio", commit: process.env.SENTRY_CURRENT_COMMIT }
```
amacneil, line 43: maybe we should pass in SENTRY_REPO too? since we already created env vars for SENTRY_ORG and SENTRY_PROJECT

web/webpack.config.ts
```diff
@@ -76,6 +76,9 @@ const mainConfig = (env: unknown, argv: WebpackArgv): Configuration => {
         org: process.env.SENTRY_ORG,
         project: process.env.SENTRY_PROJECT,
         include: path.resolve(__dirname, ".webpack"),
+        setCommits: process.env.SENTRY_CURRENT_COMMIT
+          ? { repo: "foxglove/studio", commit: process.env.SENTRY_CURRENT_COMMIT }
```
amacneil, line 80: and here

amacneil (approved):

amacneil:
probably just `SENTRY_CURRENT_COMMIT` would be ok, but I already made you change it twice to do as you please :stuck_out_tongue:

jtbandes:
```
    /**
     * The current (last) commit in the release. Required if auto option is not true.
     */
    commit?: string;
```
🤷

PR #1669, opened 2021-08-18
Bump @foxglove/rosmsg-serialization, fixes ROS 1 msg decoding regression

jhurliman:


jtbandes:
It would be helpful if you can include a link to the relevant PRs in the other repos that this is pulling in

PR #1673, opened 2021-08-19
Fix: styling on time based chart hover bar

defunctzombie:
**User-Facing Changes**
The hover bar for plot and state transitions no longer has the chevrons above and below. Instead we have a simple vertical line.

**Description**
On some browsers the styling for the hover bar is broken due to how
css is handled for the border width and padding border. Instead we use
a simple 1px wide line for the hover bar.

### DIFF
--- a/packages/studio-base/src/components/TimeBasedChart/index.tsx
+++ b/packages/studio-base/src/components/TimeBasedChart/index.tsx
@@ -105,14 +105,10 @@ const SBar = styled.div<{ xAxisIsPlaybackTime: boolean }>`
   position: absolute;
   top: 0;
   bottom: 0;
-  width: 9px;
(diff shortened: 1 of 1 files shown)


jtbandes:
I don't understand this change. You said "On some browsers the styling for the hover bar is broken due to how css is handled for the border width and padding border." — but it sounds like this was a recent regression, not a fundamental problem with some browsers. Can't we just fix the regression?

defunctzombie:
@jtbandes probably some change in box styling (from a private conversation). Looking into it I saw little value in the chevrons so simplified the vertical bar to 1px which resolves the inconsistency in width at the same time. The change is to no longer have a bar that says it has a width of 9px - and instead have one that is 1px.

PR #1674, opened 2021-08-19
Internal: remove various uses of React.memo

defunctzombie:
**User-Facing Changes**
None

**Description**
With hooks memoization happens within the component so doing additional memoization
checks on the component will end up checking properties that would later be
checked for hook dependency arrays as well.

### DIFF
--- a/packages/studio-base/src/components/MessagePathSyntax/MessagePathInput.tsx
+++ b/packages/studio-base/src/components/MessagePathSyntax/MessagePathInput.tsx
@@ -162,9 +162,7 @@ type MessagePathInputBaseProps = {
   onTimestampMethodChange?: (arg0: TimestampMethod, index?: number) => void;
 };
 
-export default React.memo<MessagePathInputBaseProps>(function MessagePathInput(
-  props: MessagePathInputBaseProps,
-) {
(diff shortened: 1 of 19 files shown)


jtbandes (changes requested):
Most of these React.memos were intentionally added at some point in the past by someone who probably noticed a performance problem. Hypothetically-motivated changes like this are "performance theater" unless you want to do some performance testing and include results in the PR, or better yet, add scripts to test performance in a relatively reproducible way 😄

defunctzombie:
@jtbandes I am of the opinion that without comments around these memo's these should be removed in favor of memo's within hooks. I view the addition of these memo's as performance theater and don't assume there was a strict performance metric either way. I'd rather remove more code from being run and work from there as we build our tooling. We aren't going to gate every change because we don't have performance tests yet.

jtbandes:
I agree that an addition of React.memo without explanation would also be performance theater, however we are now in a position where their presence is the status quo. I personally think this change is likely to introduce either a small win (if the React.memo was added carelessly or erroneously), or a medium-to-large regression (if the React.memo was added intentionally to fix an issue), so the expected value is negative.

Issue #1540, opened 2021-07-22
Remove faux titlebar


defunctzombie:
**Description**
Rather than using the system titlebar for the MacOS desktop app, we use inset controls and create our own title bar. This Is a MacOS specific feature. On Linux and Windows this creates two bars on screen.

Removing the faux titlebar is blocked on:

- [x] #1553
- [ ] Moving the player problems/connection status indicator somewhere
- [ ] Displaying the active connection name somewhere

![image](https://user-images.githubusercontent.com/84792/126671992-b7413d2f-adf2-4e49-a6dc-a499c2e039a8.png)

PR #1677, opened 2021-08-20
Change: remove in-app titlebar

defunctzombie:
**User-Facing Changes**
The connection problems are now indicated by a badge on the sidebar connections icon and listed in the sidebar pane.

image

image

**Description**
On Windows and Linux we are not able to overlay native window controls
into the app and omit the system titlebar. This creates a "dual bar"
appearance with the system titlebar and our in-app toolbar. This dual
bar takes up screen space with little value - only the current connection string and problems are placed there.

This change removes the in-app toolbar and leaves the native window
titlebar and controls. It also moved the problems into the connection
sidebar pane and adds badges to the connection sidebar icon when there are problems with the connection.

Fixes: #1540


packages/studio-base/src/components/DocumentTitleAdapter.tsx
```diff
@@ -0,0 +1,23 @@
...
+ */
+export default function DocumentTitleAdapter(): JSX.Element {
+  const { currentSourceName } = usePlayerSelection();
+
+  useEffect(() => {
+    if (!currentSourceName) {
+      window.document.title = "Foxglove Studio";
+      return;
+    }
+    window.document.title = `Foxglove Studio - ${currentSourceName}`;
```
amacneil, line 20: can you change to `${currentSourceName} - Foxglove Studio` so it's consistent with other applications.

if you want to be more specific, macOS desktop should only show `currentSourceName` and we should append ` - Foxglove Studio` everywhere else

defunctzombie: https://github.com/foxglove/studio/pull/1678

desktop/main/StudioWindow.ts
```diff
@@ -115,7 +111,7 @@ function newStudioWindow(deepLinks: string[] = []): BrowserWindow {
   browserWindow.webContents.on("ipc-message", (_event: unknown, channel: string) => {
     if (channel === "window.toolbar-double-clicked") {
       const action: string =
-        systemPreferences.getUserDefault?.("AppleActionOnDoubleClick", "string") || "Maximize";
+        systemPreferences.getUserDefault?.("AppleActionOnDoubleClick", "string") ?? "Maximize";
```
amacneil, line 114: are you sure that `getUserDefault` returns undefined rather than empty string?

defunctzombie: https://github.com/foxglove/studio/pull/1680

Issue #776, opened 2021-04-28
Extract player input prompting from PlayerManager


jtbandes:
`PlayerSelectionContext` current exposes `setPlayerFromDemoBagURL` and `setPlayerFromFiles` functions, but also a `selectSource` that prompts the user for input (such as ROS master URL):

https://github.com/foxglove/studio/blob/08e5c0c113ffb7cc3c3a7de24fd2836856e0b43d/app/components/PlayerManager.tsx#L494-L524

It would be nicer for it to expose a setter that can build a player based on inputs the user has given. Currently though, the PlayerManager itself is the thing that takes user inputs, but that seems a bit backwards.

Some notes from prior Slack conversation:

- "exposing the actual setPlayer currently doesn’t work because the actual functions to build players properly from URLs are all internal to the PlayerManager" — @jtbandes 
- "I was thinking that setPlayerFromSource could have second argument that would be the sourceOptions … so you would still pass in the http source, and the second arg would be `{ url: ... }` or whatnot" — @defunctzombie

PR #1685, opened 2021-08-23
Internal: remove maybePlayer from MessagePipeline

### LINKED ISSUE #754: Investigate removing maybePlayer
via @jtbandes:

The CONSTRUCTING status was added to PlayerPresence because some player selection options cannot be immediately constructed. Specifically because of the IndexedDB bag caching feature, `buildPlayerFromBagURLs` requires fetching the "guid" of the remote bag before the data providers can be constructed. If we could make this part of the standard data provider initialization process then we could probably remove `maybePlayer` and make MessagePipeline the only vendor of player status.

defunctzombie:
**User-Facing Changes**
None

**Description**
Players already have a player state to indicate their initialization
    status. Rather than having another layer of player status, we defer
    to use the player's state.

Minor cleanup related to https://github.com/foxglove/studio/issues/776

Fixes #754

### DIFF
--- a/packages/studio-base/src/components/PlayerManager.tsx
+++ b/packages/studio-base/src/components/PlayerManager.tsx
@@ -20,16 +20,14 @@ import {
   useRef,
   useState,
 } from "react";
+import { useToasts } from "react-toast-notifications";
 import { useLocalStorage, useMountedState } from "react-use";
 
 import { useShallowMemo } from "@foxglove/hooks";
 import Logger from "@foxglove/log";
 import { AppSetting } from "@foxglove/studio-base/AppSetting";
 import OsContextSingleton from "@foxglove/studio-base/OsContextSingleton";
-import {
-  MaybePlayer,
-  MessagePipelineProvider,
-} from "@foxglove/studio-base/components/MessagePipeline";
+import { MessagePipelineProvider } from "@foxglove/studio-base/components/MessagePipeline";
 import { useAnalytics } from "@foxglove/studio-base/context/AnalyticsContext";
 import { useCurrentLayoutSelector } from "@foxglove/studio-base/context/CurrentLayoutContext";
 import PlayerSelectionContext, {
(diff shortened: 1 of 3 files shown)


jtbandes:
Fixes https://github.com/foxglove/studio/issues/754?

Are you going to remove the CONSTRUCTING enum value?

defunctzombie:
@jtbandes Wasn't going to remove that enum value yet cause I think some players use it for their status.

defunctzombie:
The bag guids are still fetched - and maybe they don't need to be - I plan to tackle that after landing this change. I also plan to make player imports dynamic so we don't load code for players we don't use.

jtbandes:
CONSTRUCTING was used during the async call to create the player object. It should not exist anymore if we are removing maybePlayer (as mentioned in the code comment)

packages/studio-base/src/components/PlayerManager.tsx
```diff
@@ -74,12 +72,12 @@ const DEFAULT_MESSAGE_ORDER = "receiveTime";
 const EMPTY_USER_NODES: UserNodes = Object.freeze({});
 const EMPTY_GLOBAL_VARIABLES: GlobalVariables = Object.freeze({});
 
-type BuiltPlayer = {
+type BuiltPlayerMeta = {
```
jtbandes, line 75: Name is unclear - what does Meta mean here? Is it supposed to be Metadata (doubt it since this contains the actual player as well)?

defunctzombie: I'm aiming to remove this (we'll see how it goes) the sources string should be part of the player state rather than separate which is what the "Meta" was used for here.

defunctzombie:
> CONSTRUCTING was used during the async call to create the player object. It should not exist anymore if we are removing maybePlayer (as mentioned in the code comment)

velodyne player sets constructing on initialization and then changes it once initialized

jtbandes:
That can be changed to INITIALIZING

packages/studio-base/src/components/PlayerManager.tsx
defunctzombie: https://github.com/foxglove/studio/pull/1688

PR #1694, opened 2021-08-24
Change: use org slug claim in id token for signin

defunctzombie:
**User-Facing Changes**
After clicking the "signin" button, the user is presented with the device code dialog and their browser automatically opened to the device code authorization page. The web flow takes over from there and adds an org slug claim to the id token. When the polling endpoint for studio returns an id token, studio is able to immediately use that token to sign in rather than asking the user to select an org.

This simplifies the number of actions the user needs to take within studio to sign in.

**Description**
The new browser sign-in flow adds an org slug claim to the id token.
This removes the need to select an org inside studio thus simplifying
the sign-in process for studio.

### DIFF
--- a/packages/studio-base/src/components/AccountSettingsSidebar/SigninForm.tsx
+++ b/packages/studio-base/src/components/AccountSettingsSidebar/SigninForm.tsx
@@ -54,6 +54,19 @@ export default function SigninForm(): JSX.Element {
     [setBearerToken],
   );
 
+  // open new window with the device code to facilitate user signin flow
+  useEffect(() => {
+    if (!deviceCode) {
+      return;
+    }
+
+    const url = new URL(deviceCode.verification_uri);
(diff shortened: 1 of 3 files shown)


defunctzombie:
Blocked on console sign-in flow adding an org slug to the id token.

packages/studio-base/src/components/AccountSettingsSidebar/SigninForm.tsx
```diff
@@ -54,6 +54,19 @@ export default function SigninForm(): JSX.Element {
     [setBearerToken],
   );
 
+  // open new window with the device code to facilitate user signin flow
+  useEffect(() => {
+    if (!deviceCode) {
+      return;
+    }
+
+    const url = new URL(deviceCode.verification_uri);
```
rdgifford, line 63: should check props on deviceCode?

PR #1695, opened 2021-08-24
Upgrade xmldom

amacneil:
**User-Facing Changes**
None

**Description**
Upgrade `xmldom` (by switching to `@xmldom/xmldom`) to fix [CVE-2021-32796](https://github.com/advisories/GHSA-5fg8-2547-mr8q). See https://github.com/xmldom/xmldom/issues/271 for background on why the package name changed.

Types are now included in the package (https://github.com/xmldom/xmldom/issues/191). This PR adds `"lib": ["dom"]` to hooks package tsconfig because the previous xmldom types were automatically adding this for us with a `<reference lib="dom" />` (https://github.com/xmldom/xmldom/issues/285).

### DIFF
--- a/packages/hooks/tsconfig.json
+++ b/packages/hooks/tsconfig.json
@@ -3,6 +3,7 @@
   "include": ["./src/**/*"],
   "compilerOptions": {
     "rootDir": "./src",
-    "outDir": "./dist"
+    "outDir": "./dist",
+    "lib": ["dom", "es2020"]
   }
 }
--- a/desktop/main/rosPackageResources.ts
+++ b/desktop/main/rosPackageResources.ts
@@ -1,12 +1,12 @@
 // This Source Code Form is subject to the terms of the Mozilla Public
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
+import { DOMParser } from "@xmldom/xmldom";
 import { protocol } from "electron";
 import { promises as fs } from "fs";
 import path from "path";
 import { PNG } from "pngjs";
 import UTIF from "utif";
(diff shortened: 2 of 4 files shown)


packages/hooks/tsconfig.json
```diff
@@ -3,6 +3,7 @@
   "include": ["./src/**/*"],
   "compilerOptions": {
     "rootDir": "./src",
-    "outDir": "./dist"
+    "outDir": "./dist",
+    "lib": ["dom", "es2020"]
```
amacneil, line 7: sanity check on this please

defunctzombie:
LGTM

karfau:
@amacneil (from the xmldom types perspective), I would really love to understand why the hooks package needed the change in tsconfig.json, since in the file you linked I do not see any reference to xmldom.
What exactly was the error message

defunctzombie:
@karfau The hooks package needed to specify "dom" in the lib config because the code in useVisibilityState (linked in an above comment) references the global "document". For that we need to tell typescript we want to bring in the dom library into our scope.

karfau:
Yes, that makes sense. I just don't understand why this change was required as part of this PR/how it is related to updating xmldom.

defunctzombie:
@karfau The previous xmldom types (@types/xmldom) had `<reference lib="dom" />`. This changes the typescript compiler configuration to add "dom" lib for any users of those types. The new types don't do this so the end user should add dom to their lib themselves.

amacneil:
Weirdly the new types do still contain that reference, but typescript definitely ignored it when I upgraded the package. 

I will try to make a minimal reproduction of this issue today.

amacneil:
Moving discussion to https://github.com/xmldom/xmldom/issues/285

PR #1698, opened 2021-08-24
Internal: debounce useResizeDetector changes

defunctzombie:
**User-Facing Changes**
None

**Description**
Avoid ResizeObserver loop limit exceeded errors by marking resize detector
events with a debounce refresh mode. This causes the react state update to
occur outside the resize observer handler.


defunctzombie:
Related https://github.com/maslianok/react-resize-detector/issues/45

web/src/index.tsx
```diff
@@ -12,6 +12,10 @@ import VersionBanner from "./VersionBanner";
 const log = Logger.getLogger(__filename);
 log.debug("initializing");
 
+window.onerror = (...args) => {
+  console.error(args);
```
jtbandes, line 16: probably `...args` here? no need to log it as an array

packages/studio-base/src/components/PanelToolbar/index.tsx
```diff
@@ -280,6 +280,8 @@ export default React.memo<Props>(function PanelToolbar({
 
   const { width, ref: sizeRef } = useResizeDetector({
     handleHeight: false,
+    refreshRate: 0,
+    refreshMode: "debounce",
```
jtbandes, line 284: drop a comment in these places to indicate why?

jtbandes (commented):
Seems fine – I feel like we still don't understand whether this change is doing something useful, but I guess suppressing a noisy error message that doesn't noticeably affect the app behavior is in itself useful 🤷

packages/studio-base/src/components/PanelToolbar/index.tsx
defunctzombie: yea - that seems like a good idea - I'll reference the resize detector issue in the comment

defunctzombie:
> @defunctzombie do you want to remove this from the sentry ignore list then? or keep it there?

https://github.com/foxglove/studio/issues/1790

PR #1707, opened 2021-08-27
Fix: limit the plot history size in follow mode

defunctzombie:
**User-Facing Changes**
No visual changes. The app should be better about memory management when streaming data and a follow window size is configured.

**Description**
When streaming data, the plot would keep an infinite history of
messages even when a fixed follow window of some seconds was set. Keeping
the app conntected would result in the memory usage continuing to grow
and more dataset processing to occur as the messages array grew.

With this change, if there is a follow window, messages older than the
window are discarded.


airfield20:
I am still having this issue running on version 1.0. Has this merge been included in the latest build?

defunctzombie:
@airfield20 This change has been included in 1.0. In what way are you seeing this issue manifest?

airfield20:
Foxglove studio goes from 100mb to 1.3gb ram usage over the course of 30 minutes when using the plotting tool with a 30 second x axis window set.

defunctzombie:
Which data source are you using?

defunctzombie:
@airfield20 Can you open a new issue and write up any notes you can on reproducing the issue? Helpful notes are: data source, operating system, which panels are involved. Anything that helps us isolate a reproducible case for this. That will help us track the problem and notify you of fixes.

airfield20:
sure thing https://github.com/foxglove/studio/issues/2868

PR #1718, opened 2021-08-29
Use full product name in public-facing text

amacneil:
**User-Facing Changes**
Update some strings from `Studio` to `Foxglove Studio`.

Just "Studio" alone looks awkward in sentences.


packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/typescript/templates/README.md
```diff
@@ -3,5 +3,3 @@
 ## Contributing
 
 Please feel free to add any template you like in this directory with a `*.ts` extension. Then reference that template in the `templates/index.js` file as is done with the other templates.
-
-Then open up a PR in the Studio repo. We'll make sure a team member looks at it in a timely manner! :)
```
amacneil, line 7: I also removed a few weird instructions from the Node Playground docs which were a hangover from Cruise internal use

PR #1720, opened 2021-08-30
Fix: interaction padding tweaks to align labels

defunctzombie:
**User-Facing Changes**
Labels in marker interaction view line up.

Before:
image

After: 
image

**Description**
The labels in marker interaction view had an incorrect left padding which made them appear offset from their expandable siblings. Now all the labels (expandable or not) line up.


jtbandes:
Can you fix this in Raw Messages as well?

PR #1724, opened 2021-08-30
Use readySignal in LegacyPlot stories

jtbandes:
**User-Facing Changes**
None

**Description**
Add a count feature to useReadySignal. Adjust the Panel HOC so extra props can be passed through, allowing onChartUpdate to be passed in to the LegacyPlot panel in stories.


defunctzombie:
LGTM - I tried to fix this with readySignal but found out that Panel() HOC wraps the actual panel so I didn't have a nice way to capture the "chart has rendered" event since LegacyPlot doesn't use resume/pause frame.

defunctzombie:
Looks like you've found the winning combination ♾️

PR #1725, opened 2021-08-30
Fix: avoid changing panel state for unchanged config

### LINKED ISSUE #1646: Changes appearing immediately after switching layouts
When switching between layouts, the layout sometimes immediately appears modified, even though the user didn't change anything. If you look at the json before/after, sometimes the panel configs that change are actually from panels that aren't present in the layout — this indicates a stale reference to old config being saved under the new layout.

I thought I already fixed this issue in #1425, so not sure if it regressed or just wasn't totally fixed. The logic for `savedDefaultConfig.current` here is supposed to prevent saving data from an old layout after switching.

https://github.com/foxglove/studio/blob/62d8b42e5817e00660015a5baf3144dc2d5f8d7f/packages/studio-base/src/components/Panel.tsx#L189-L198

defunctzombie:
**User-Facing Changes**
When the user changes layouts they would be shown that their layout is "dirty" and has unsaved changes even tho there were no actual changes to the layout. After this change the layout is not marked as dirty when just changing the layout.

**Description**
When a panel saves its config, this triggers a set of reducers
to run. If the config is unchanged from the previous config, the save
config reducer keeps the same state object to indicate there were
no changes to the state.

Fixes #1646

### DIFF
--- a/packages/studio-base/src/providers/CurrentLayoutProvider/index.tsx
+++ b/packages/studio-base/src/providers/CurrentLayoutProvider/index.tsx
@@ -1,6 +1,7 @@
 // This Source Code Form is subject to the terms of the Mozilla Public
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
+import { isEqual } from "lodash";
 import { useCallback, useEffect, useMemo, useRef, useState } from "react";
 import { getNodeAtPath } from "react-mosaic-component";
 import { useToasts } from "react-toast-notifications";
@@ -155,6 +156,14 @@ export default function CurrentLayoutProvider({
       ) {
         return;
       }
+      const oldData = layoutStateRef.current.selectedLayout.data;
+      const newData = panelsReducer(layoutStateRef.current.selectedLayout.data, action);
+
+      // the panel state did not change, so no need to perform layout state updates or layout manager updates
+      if (isEqual(oldData, newData)) {
+        return;
+      }
+
       const newLayout = {
         id: layoutStateRef.current.selectedLayout.id,
         data: panelsReducer(layoutStateRef.current.selectedLayout.data, action),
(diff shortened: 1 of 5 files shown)


defunctzombie:
@jtbandes if this approach makes sense I can add a test (I hope I can add a test) that ensures this behavior.

packages/studio-base/src/providers/CurrentLayoutProvider/reducers.ts
```diff
@@ -744,7 +756,9 @@ const endDrag = (panelsState: PanelsState, dragPayload: EndDragPayload): PanelsS
 };
 
 export default function (panelsState: PanelsState, action: PanelsActions): PanelsState {
-  let newPanelsState = { ...panelsState };
+  // start the newPanelsState as the current panels state so if there are no changes the identity
+  // of the panels state object remains the same
+  let newPanelsState = panelsState;
```
jtbandes, line 761: Do we need an isEqual in SET_GLOBAL_DATA and SET_USER_NODES as well?

jtbandes: Looks like OVERWRITE_GLOBAL_DATA mutates the state, so that needs to change

defunctzombie: Do we? Or do those eventually save into the state?

jtbandes: > do those eventually save into the state?

not sure what you mean – my point is they don't have isEqual checks so they may result in unnecessary updates

packages/studio-base/src/providers/CurrentLayoutProvider/index.tsx
```diff
@@ -155,6 +155,14 @@ export default function CurrentLayoutProvider({
       ) {
         return;
       }
+      const oldData = layoutStateRef.current.selectedLayout.data;
+      const newData = panelsReducer(layoutStateRef.current.selectedLayout.data, action);
+
+      // the panel state did not change, so no need to perform layout state updates or layout manager updates
```
defunctzombie, line 161: fixme - add isEqual check here as well

Issue #1620, opened 2021-08-11
Add Amplitude metrics for Team Layouts flows


esthersweon:
Add Amplitude metrics for the following events:

1. Actions after working with a team layout (i.e. a template)
   - Overwriting a team layout
   - Discarding changes to a team layout (i.e. keeping the template the same)
   - Saving an edited team layout as a personal layout
2. Conflicts
   - Personal layouts
   - Team layouts

These would be great for getting a sense of:

- how often shared layouts are edited vs. truly used as a template
- how often users are bothered to merge conflicts

PR #1735, opened 2021-08-31
Add analytics for layout overwrite/revert; provide event category to Sentry breadcrumbs

jtbandes:
**User-Facing Changes**
None

**Description**
Add analytics for new actions added in #1679. Add `type` and `category` to Sentry breadcrumbs instead of sending the event name as the category.

Related: #1620


packages/studio-base/src/services/IAnalytics.ts
```diff
@@ -23,13 +30,94 @@ enum AppEvent {
...
+}
+
+export function getEventBreadcrumbType(event: AppEvent): SentryBreadcrumbType {
+  switch (event) {
+    case AppEvent.APP_INIT:
+      return SentryBreadcrumbType.DEFAULT;
+
+    case AppEvent.PLAYER_CONSTRUCTED:
+    case AppEvent.PLAYER_INITIALIZED:
+      return SentryBreadcrumbType.TRANSACTION;
```
jhurliman, line 94: `NAVIGATION`?

packages/studio-base/src/services/IAnalytics.ts
```diff
@@ -23,13 +30,94 @@ enum AppEvent {
...
+
+    case AppEvent.PLAYER_PLAY:
+    case AppEvent.PLAYER_SEEK:
+    case AppEvent.PLAYER_SET_SPEED:
+    case AppEvent.PLAYER_PAUSE:
+      return SentryBreadcrumbType.USER;
+
+    case AppEvent.PLAYER_CLOSE:
+      return SentryBreadcrumbType.TRANSACTION;
```
jhurliman, line 103: `NAVIGATION`?

jhurliman (approved):

packages/studio-base/src/services/IAnalytics.ts
jtbandes: Not very clear to me the intended difference between `transaction` (a "tracing event") and `navigation` (a "UI transition") ... which one is a change in player? The UI doesn't really transition...the user has just made a selection.

jtbandes: Btw I'm not sure the `type` really does anything in the sentry UI other than change an icon.

PR #1742, opened 2021-09-01
Team Layouts II: Return of the Console API

jtbandes:
**User-Facing Changes**
None (feature flagged off)

**Description**
- Bring back IRemoteLayoutStorage; add new RemoteLayoutStorageContext.
  - Rename updatedAt → savedAt in Layout to match new server API.
  - Add syncInfo to Layout.
- Initialize LayoutManager with both local and remote storage.
  - Storage is namespaced by user id so logging out doesn't leave junk in your local layouts folder.
  - Create (share) / rename / update (overwrite) / delete operations _on shared layouts_ go directly to the server (will fail if offline).
- Prompt the user to save or revert changes when attempting to switch away from a shared layout that has local changes.
- Resolves https://github.com/foxglove/studio/issues/1620


packages/studio-base/src/services/ConsoleApi.ts
```diff
@@ -96,36 +99,55 @@ class ConsoleApi {
...
 
-  private async request<T>(url: string, config?: RequestInit): Promise<T> {
+  private async request<T>(
+    url: string,
+    config?: RequestInit,
+    {
+      allowedStatuses = [],
+    }: {
+      /** By default, status codes other than 200 will throw an error. */
+      allowedStatuses?: number[];
```
jtbandes, line 148: cc @defunctzombie

desktop/renderer/Root.tsx
```diff
@@ -58,10 +62,15 @@ export default function Root(): ReactElement {
...
 
+  const api = useMemo(() => new ConsoleApi(process.env.FOXGLOVE_API_URL!), []);
+
   const providers = [
     /* eslint-disable react/jsx-key */
-    <StudioToastProvider />,
     <NativeStorageAppConfigurationProvider />,
+    <ConsoleApiContext.Provider value={api} />,
+    <CurrentUserProvider />,
+    <ConsoleApiRemoteLayoutStorageProvider />,
```
jtbandes, line 72: cc @defunctzombie - hoisting all of these so the remote layout storage can be configured outside studio-base

packages/studio-base/src/hooks/useCallbackWithToast.ts
```diff
@@ -0,0 +1,32 @@
...
+import { useToasts } from "react-toast-notifications";
+
+import Logger from "@foxglove/log";
+
+const log = Logger.getLogger(__filename);
+
+/**
+ * A version of React.useCallback() displaying any errors thrown from the function as toast notifications.
+ */
+export default function useCallbackWithToast<Args extends unknown[]>(
```
jtbandes, line 15: cc @defunctzombie - may be useful in other places

packages/studio-base/src/services/LayoutManager/index.ts
```diff
@@ -53,148 +66,498 @@ class NamespacedLayoutStorage {
...
+  private emitter = new EventEmitter<LayoutManagerEventTypes>();
 
-  readonly supportsSharing = false;
+  private activityCount = 0;
+
+  /**
+   * A decorator to emit activity events before and after an async operation so the UI can show that
+   * the operation is in progress.
+   */
+  private static withActivity<Args extends unknown[], Ret>(
```
jtbandes, line 89: cc @defunctzombie for your consideration

desktop/renderer/Root.tsx
defunctzombie: :thumbsup:

packages/studio-base/src/hooks/useCallbackWithToast.ts
defunctzombie: useAsyncWithToast? to differentiate from useCallback which doesn't deal with async things?

I mocked up a version of this for console where I passed along the state from useAsync back down https://github.com/foxglove/console/pull/105/commits/2f85ec7f004e2be799c9009ecdbc3bfb2405d861#diff-e93b4eaa92ece2da93fb66e49c25ebcbfe7a6449a7fd17d5447b32c5044d81e3

packages/studio-base/src/services/LayoutManager/index.ts
defunctzombie: Can't the UI - which calls the async function do that itself? Maybe better handled by a hook? the useAsync hook for example already has a `isLoading` field to query for such things. I think the hook approach (or something at the async call site is more flexible than the decorator which we would need to add to various places.

packages/studio-base/src/services/ConsoleApi.ts
defunctzombie: I guess thats the future. Probably good to make the { status, json } thing a named type (rather than inline) - maybe RequestResult or some such. I'm sure Rust can provide some inspiration.

packages/studio-base/src/hooks/useConfirm.tsx
```diff
@@ -100,6 +100,7 @@ function ConfirmModal(props: ConfirmModalProps) {
       hidden={false}
       onDismiss={() => onComplete("cancel")}
       dialogContentProps={{ title: props.title }}
+      styles={{ main: { maxWidth: "unset !important" } }}
```
defunctzombie, line 103: What happened here?

defunctzombie:
I think what you are trying to do with the decorator is better done at the UI layer - needs more hooks :D

packages/studio-base/src/hooks/useCallbackWithToast.ts
jtbandes: But the function signature is most like useCallback - it just returns a function. useAsync immediately executes the function and returns a state, and useAsyncFn returns a state and function.

packages/studio-base/src/services/LayoutManager/index.ts
jtbandes: The thinking here is that the layout manager is the source of truth. Some of the actions are UI initiated and some are not, for example the periodic background sync. If the UI were to initiate an action and then reset (e.g. switching to a different sidebar tab and back) then the async state would be lost, whereas the layout manager always knows if it is doing an operation.

packages/studio-base/src/hooks/useConfirm.tsx
jtbandes: The dialog was cramping my style

packages/studio-base/src/services/LayoutManager/index.ts
defunctzombie: The periodic background sync is initiated in the provider right? I thought that is what we settled on to avoid a service having internal timers?

jtbandes: That's right. Though the provider and UI are different places – if you want the provider to track whether there's a current sync via useAsyncFn, then it would have to also provide that sync status down to the UI if the UI wants to show it. `<LayoutManagerProvider value={{ isLoading, layoutManager }}` seems kinda gross IMO.

jtbandes: discussed in slack; concluded that decorator pattern makes sense and the event/property are renamed to "busy" rather than "activity" to make intent more clear.

PR #1751, opened 2021-09-02
Fix live players updating timestamps too often, prefer network-sourced timestamps

jhurliman:
**User-Facing Changes**

Performance improvement when using a live data source.

### Read start/end times from live players when possible instead of local time

Live players were often publishing changes to the current timestamp faster than any data was being received from the network, causing spurious UI updates. The data was also misleading, because a robot could stop publishing and shut down, and the end timestamp would continue to increase.

This changes live player code to only update start/end timestamps when a message is received, and prefer to use timestamps from the data source instead of local time when possible.


packages/studio-base/src/players/Ros2Player.ts
```diff
@@ -547,13 +558,13 @@ export default class Ros2Player implements Player {
   }
 
   private _getCurrentTime(): Time {
-    const now = fromMillis(Date.now());
-    return now;
+    const lastTime = this._end ?? { sec: 0, nsec: 0 };
+    return lastTime;
     // if (this._clockTime == undefined) {
```
defunctzombie, line 563: stale code?

jhurliman: Not implemented yet

defunctzombie: I think better to open an issue or something - when I see commented out code in the source I wonder if it was a mistake - or more generally delete it figuring it was left behind on accident.

jtbandes:
What's the status of this PR?

Issue #1743, opened 2021-09-01
Team layouts UI polish (local layouts)


jtbandes:
Notes from #1528 / team discussion:

   - [ ] inconsistent ellipses in menus
   - [x] #1754
     - [x] show ... menu only on hover
     - [x] replace ... with filled dot or some other indicator of dirty state
     - [x] ~pressing `tab` during renaming a layout should loop between save, cancel, then back to textfield (within same row)~
     - [x] during layout rename, clicking (or right clicking) any element outside of the row should end the rename (saving changes)
- [ ] any other necessary polish

Shared layouts items tracked in:
- #1744

amacneil:
dot for dirty state:

![Screenshot from 2021-09-01 15-24-37](https://user-images.githubusercontent.com/637671/131753574-29b3986b-51d7-4a0e-bad6-4c6367bbd4e4.png)

my vote is to remove the "three dots" menu entirely and rely on right-click for context menu like every other app does

jtbandes:
Right click is rarely if ever the _only_ way to access some action. It is a nice to have for people who remember to use it but I don’t think we should lean on it.

amacneil:
Fair point - although typically the same actions are available via the traditional menubar as well (not via inline icons).

E.g. VSCode and Figma both follow this pattern. Most actions are available via right click menu, and also via the menubar. Both apps also expose their app menus under a single button in browser mode (i.e. github.dev for VSCode).

I'd encourage us to follow this pattern and act more like a native app, and expose the actions via right click and a `Layout` menu or something.

Our menubar is hidden by default on Linux+Windows, but can be easily accessed by hitting `Alt` (which most power users know about on win/mac). I would be in favor of showing it all the time though (VSCode shows menubar all the time, Figma desktop seems to rely on their same monomenu button as they use in browser).

amacneil:
I think "traditional" is changing over time. Both Gnome and Windows seem to be moving away from menu bars and dumping all menus under a hamburger menu somewhere, or removing it entirely. But power user tools tend to keep it.

But for some specific examples - on Linux, both Slack and VS Code show menu bar by default, but let you configure it. In Slack you can hide it via `Window`->`Always Show Menu Bar`. In VSCode you can hide it via `View`->`Appearance`->`Show Menu Bar`. You can always access it via `Alt` shortcut.

amacneil:
I don't use Windows as frequently, but I don't think the ribbon thing is very common outside of Office and Windows Explorer?

PR #1754, opened 2021-09-03
Layout Row styling

2metres:
### Changes

- Removes `✔ | ❌` buttons
- Only shows `•••` on hover
- More prominent selected state
 
https://user-images.githubusercontent.com/924528/131936339-7a374d38-9bcf-412a-8614-a3aef4fcf88f.mp4

Related to #1743 and #1528


packages/studio-base/src/components/LayoutBrowser/LayoutRow.tsx
```diff
@@ -5,13 +5,15 @@
...
   IconButton,
+  Text,
   TextField,
   ITextField,
   makeStyles,
   Stack,
   useTheme,
   IContextualMenuItem,
   ContextualMenu,
+  StackItem,
```
jtbandes, line 16: FYI I looked at the source and StackItem === Stack.Item. 🤷

packages/studio-base/src/components/LayoutBrowser/index.tsx
```diff
@@ -257,23 +257,39 @@ export default function LayoutBrowser({
       title="Layouts"
       noPadding
       trailingItems={[
-        layouts.loading && <Spinner />,
-        // eslint-disable-next-line react/jsx-key
+        layouts.loading && <Spinner key="spinner" />,
         <IconButton
+          key="add-layout"
```
jtbandes, line 262: this key ends up getting ignored by SidebarContent, that's why I disabled it – but doesn't really matter

PR #1755, opened 2021-09-03
Internal: move scrubber and time rendering to sub components

defunctzombie:
**User-Facing Changes**
Maybe the tiniest amount of CPU savings during playback.

**Description**
By moving the connected updates to scrubber and time display to sub
components we avoid re-rendering the broader playback control panel
during playback as the current time changes. Since the current time
may update at 60hz this saves us some rendering of non-essential components.


web/webpack.config.ts
```diff
@@ -100,7 +100,7 @@ const mainConfig = (env: unknown, argv: WebpackArgv): Configuration => {
       publicPath: "/",
 
       // Output filenames should include content hashes in order to cache bust when new versions are available
-      filename: "[name].[contenthash].js",
+      filename: isDev ? "[name].js" : "[name].[contenthash].js",
```
jtbandes, line 103: Why?

defunctzombie: Seems that in dev `contenthash` is undefined and `yarn web:serve` doesn't work 😭

packages/studio-base/src/components/PlaybackControls/RepeatAdapter.tsx
```diff
@@ -0,0 +1,50 @@
...
+      // if the user turns on repeat and we are at the end, we assume they want to play from start
+      // even if paused
+      play();
+    } else if (isPlaying) {
+      // no-repeat
+      // pause playback to toggle pause button state
+      // if the user clicks play while we are at the end, we go back to begginning
+      pause();
+    }
+  }
```
jtbandes, line 47: Generally an anti-pattern to have any side effects directly in render – put it in a useEffect.

I do think the message pipeline is where I would expect to find this code rather than some small unassuming React component in PlaybackControls 🤷

packages/studio-base/src/components/PlaybackControls/index.tsx
```diff
@@ -11,465 +11,255 @@
...
-    const clearHoverValue = useClearHoverValue();
-    const setHoverValue = useSetHoverValue();
-
-    // playerState is unstable, and will cause callbacks to change identity every frame. They can take
-    // a ref instead.
-    const playerState = useRef(player);
-    playerState.current = player;
+      [selector],
+    ),
+  );
```
jtbandes, line 63: Rather than a side-effectful selector, the way I'd expect this to be implemented is to just expose the currentValue() getter from here:

https://github.com/foxglove/studio/blob/3ceeb706842030d22cf07d367595ad6792f1dd35/packages/studio-base/src/hooks/useContextSelector.ts#L38

e.g.

```ts
function useSelectableContextGetter<T>(ctx: SelectableContext<T>) {
  return useContext(context._ctx).currentValue();
}

// in MessagePipeline:
function useMessagePipelineGetter() {
  return useSelectableContextGetter(ContextInternal)
}
// elsewhere
const getMessagePipelineState = useMessagePipelineGetter()
useCallback(()=>{
  ... use getMessagePipelineState().playerState ...
},[])
```

web/webpack.config.ts
jtbandes: wow wut

packages/studio-base/src/components/PlaybackControls/index.tsx
defunctzombie: I've added a useMessagePipelineGetter and selectableContextGetter as per the recommendation. I dunno if a getter is better than a ref but seems ok

packages/studio-base/src/components/PlaybackControls/RepeatAdapter.tsx
defunctzombie: I've put this into an effect. This was not in message pipeline since I didn't want to introduce the concept of repeat into pipeline at the time I added this.

PR #1762, opened 2021-09-03
Upgrade to TypeScript 4.4.2

jtbandes:
**User-Facing Changes**
None

**Description**
TypeScript 4.4.2 removes OffscreenCanvas types (https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/1029) so we use `@types/offscreencanvas` instead.

Disable `useUnknownInCatchVariables` for now because it would require a significant migration effort.


packages/studio-base/src/components/Chart/index.tsx
```diff
@@ -96,7 +96,7 @@ function Chart(props: Props): JSX.Element {
     async <T extends unknown>(
       topic: string,
       payload?: Record<string, unknown>,
-      transferables?: Transferable[],
+      transferables?: (Transferable | OffscreenCanvas)[],
```
defunctzombie, line 99: An OffscreenCanvas is not a Transferable anymore?

defunctzombie:
LGTM

packages/studio-base/src/components/Chart/index.tsx
jtbandes: Not in the new definition of Transferable from the TS bundled lib.

PR #1766, opened 2021-09-03
Expose filePath on PlayerState; setRepresentedFilename on macOS

jtbandes:
**User-Facing Changes**
On macOS, when a ROS 1 bag file is opened, the window title bar now references the file so you can see its location on the filesystem.

**Description**
- Add a `filePath` to player state and expose `File.path` (an electronism) for bag files. Unfortunately it only works for File, not FileSystemDirectoryHandle, so it doesn't work for ROS 2 bag folders. 🤷 
- Expose https://www.electronjs.org/docs/tutorial/represented-file via new NativeWindowContext and use it to show the player's filePath.

### DIFF
--- a/desktop/common/types.ts
+++ b/desktop/common/types.ts
@@ -48,6 +48,7 @@ type DesktopExtension = {
 };
 
 interface Desktop {
+  /** https://www.electronjs.org/docs/tutorial/represented-file */
   setRepresentedFilename(path: string | undefined): Promise<void>;
 
   // Get an array of deep links provided on app launch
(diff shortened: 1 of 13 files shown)


desktop/common/types.ts
```diff
@@ -48,6 +48,8 @@ type DesktopExtension = {
 };
 
 interface Desktop {
+  setRepresentedFilename(path: string | undefined): Promise<void>;
```
defunctzombie, line 52: Comment please

PR #1767, opened 2021-09-03
Move loading state into selectedLayout so id can be read even when layout is loading

jtbandes:
**User-Facing Changes**
None

**Description**
Reduces visual delay when selecting a layout. Now the clicked row will be highlighted while the layout is loading.


packages/studio-base/src/providers/CurrentLayoutProvider/index.tsx
```diff
@@ -68,7 +68,6 @@ export default function CurrentLayoutProvider({
   }, []);
 
   const [layoutState, setLayoutStateInternal] = useState<LayoutState>({
```
jtbandes, line 70: most significant change is in this file

amacneil:
:clap:

PR #1780, opened 2021-09-05
Replace PanelToolbar/index.module.scss with mergeStyleSets

2metres:
Part of larger style approach refactor, If done correctly should produce no user facing changes.


packages/studio-base/src/components/PanelToolbar/index.tsx
```diff
@@ -53,6 +52,75 @@ type Props = {
...
+      pointerEvents: "none",
+
+      "*": {
+        pointerEvents: "auto",
+      },
+      "&.hasChildren": {
+        left: 0,
+        backgroundColor: colors.TOOLBAR_FIXED,
+      },
+      "&:not(.hasChildren) > *": {
```
2metres, line 105: I don't love doing this but `makeStyles/mergeStyles` doesn't have a way to reference a selector it creates that I can find

defunctzombie:
LGTM

Issue #1691, opened 2021-08-23
Remove SCSS + CSS modules


amacneil:
Convert everything to `makeStyles`

2metres:
Closed by #1921

PR #1782, opened 2021-09-05
Refactor ToolbarTab to use fluentui style methods

2metres:
**User facing changes**
None

**Description**
Part of larger style approach refactors #1691 

- Change ToolbarTab to use CSS-in-JS and remove css modules


amacneil:
Please use the PR description template (i.e. put “none” under user facing changes section) so it’s easier to write release notes and debug future issues

PR #1785, opened 2021-09-07
Refactor Icon component to use CSS-in-JS

2metres:
**User-Facing Changes**
When using an Icon component use the `size` prop in favor of `xlarge | large | medium | small | xsmall | xxsmall` named size props:

```jsx
// Old API
<Icon small />
```
```jsx
// New API
<Icon size="small" />
```

This approach is more effective because in the old implementation allows for a situation like this: 
```jsx
<Icon xlarge large medium small xsmall xxsmall />
```
Creating a situation where whatever the last CSS selector applied wins

**Description**
Remove `Icon.modules.scss` in favor of CSS-in-JS. Part of larger style approach refactor #1691


2metres:
@defunctzombie I could use some help with Typescript and `mergeStyleSets`. 

It seems that fluentui style functions don't like programmatically generated selectors [FluentUI Issue 6060](https://github.com/microsoft/fluentui/issues/6060)

So currently the typechecking returns for each of the IconSizes: 
```sh
Property 'xlarge' does not exist on type 'IProcessedStyleSet<{...'.ts(2339)
```

defunctzombie:
@2metres I've fixed up the type issue

PR #1787, opened 2021-09-07
Added nginx-based dockerfile for web studio

ShepelIlya:
**User-Facing Changes**

**Description**

Added nginx-based dockerfile (like it was done in webviz app) for web build of Foxglove Studio.

**Docs**

For building this image use next commands:

```bash
cd studio
docker build . -f=docker/Dockerfile.StaticWebStudio -t <your_tag>
```

### DIFF
--- a/docker/Dockerfile.StaticWebStudio
+++ b/docker/Dockerfile.StaticWebStudio
@@ -1,0 +1,15 @@
+
+FROM node:latest AS builder
+
+WORKDIR /app
+
+COPY . /app
+
+RUN yarn install && yarn web:build:prod
+
+FROM nginx:latest
+
+RUN sed -i 's/listen  *80;/listen 8080;/g' /etc/nginx/conf.d/default.conf
+EXPOSE 8080
+
+COPY --from=builder /app/web/.webpack /usr/share/nginx/html


amacneil:
I think adding a Dockerfile is a good idea, but it might be easier to use npm's `http-server` rather than nginx.

We should probably also consider a 2 stage build so that the final container is small. We could then publish it on Github's registry (ghcr.io).

ShepelIlya:
I am not good enough in web development, so could you give some hints how can I rebase server to docker with npm?

defunctzombie:
@amacneil The current PR does use a 2 stage build.

A few thoughts on what I think we need to consider for this PR to land.

- Documentation on how to use this - probably on the README - at the least as a comment in the dockerfile itself since users will need to expose the right port from the dockerfile.
- The dockerfile should use explicitly versions of base images. Rather than `node:latest` we need to use `node:0.14.?` or `node:0.16.?` and the same with any second stage container. This is to avoid tooling compatibility issues if/when node makes a new release.
- `RUN yarn install && yarn web:build:prod` could be split into two RUN statements for readability. Related, the `COPY . /app` could try to copy only the yarn lockfile and perform the install step  then copy the other files. That tends to have better "rebuild" performance due to layer caching.
- I think this Dockerfile could live in the root of the repo for those that want a simple "build me a docker image to run this in browser" - not a strong preference

PR #1797, opened 2021-09-08
Remove sendNotification from SceneBuilder

defunctzombie:
**User-Facing Changes**
A user is able to open a bag which contains some markers and then some tf messages without being bombarded with global popup warnings about `Topic /foo has bad frame`. Instead the bag loads correctly and as frames are available markers are rendered. Any errors are localized to the 3d panel topic list.

**Description**
SceneBuilder would send a notification when a topic had a bad frame id. This notification
would popup for the user and persist even tho scene builder might receive more topics
and frames and internally clear the error.

This change removes the sendNotification and uses the already present per-topic
notification mechanism local to the 3d panel to inform the user if there are errors on
their topic. This mechanism properly clears itself when there are no errors on the topic.


amacneil:
🙏

jtbandes:
Does this actually work? It doesn't seem like the `topicsWithBadFrameIds` is accessed outside of Scene Builder. How do these errors end up appearing in the topic tree?

defunctzombie:
I will double check that with some fake data. It looked to me from the sample bag I had that frames appeared within the first loading so the errors cleared but the send notification persisted.

PR #1800, opened 2021-09-08
Don't create topic -> node edges for filtered topics

jhurliman:
**User-Facing Changes**

Fixes the "Error: Can not create edge `t:/a-n:/b` with nonexistant source `t:/a` error that would occur with certain filter settings in the Topic Graph panel

**Description**

A new topicPassesConditions() check was added but it was only used to filter topic nodes, not topic->rosnode edges.


esthersweon (approved):
🎉

PR #1805, opened 2021-09-08
Fix: remove sendNotification from PlaybackControls slider

### LINKED ISSUE #1804: AppError: Error: Slider component given invalid range: 281.03, 13.022
Sentry Issue: [STUDIO-G8](https://sentry.io/organizations/foxglove/issues/2633831966/?referrer=github_integration)

```
AppError: Error: Slider component given invalid range: 281.03, 13.022
    at Bs.render (file:///opt/Foxglove%20Studio/resources/app.asar/renderer/main.js:4093:1433)
    at ho (file:///opt/Foxglove%20Studio/resources/app.asar/renderer/main.js:335:67893)
    at ...
  at Bs.render (../packages/studio-base/src/components/PlaybackControls/Slider.tsx:136:19)
  at None (../packages/studio-base/src/components/PlayerManager.tsx:473:11)
  at u (../packages/studio-base/src/util/sendNotification.ts:97:24)
  at Bs.render (../packages/studio-base/src/components/PlaybackControls/Slider.tsx:137:7)
  at None (../packages/studio-base/src/components/PlayerManager.tsx:473:11)
...
(28 additional frame(s) were not displayed)
```

defunctzombie:
**User-Facing Changes**
Rather than seeing a global notification that the slider range is invalid on every render, the slider instead won't render.

For a user that experiences this, it will mean their scrubber is not usable as the range is wrong but other parts of their app should continue to function. The user will no longer receive an un-actionable message.

**Description**
Instead of notifying when the max/min range is inverted, the slider will
    render nothing and log a console error.
    
This change updates the slider component from a class component to a functional
    component. It also removes the document events module and listens to events on
    the slider div directly.

Fixes #1804


packages/studio-base/src/components/PlaybackControls/Slider.tsx
```diff
@@ -36,127 +40,115 @@ export const StyledRange = styled.div.attrs<{ width: number }>(({ width }) => ({
...
-        <DocumentEvents
-          target={window}
-          enabled={mouseDown && draggable}
-          onMouseUp={this._onMouseUp}
-        />
-        {renderSlider(value != undefined && min !== max ? (value - min) / (max - min) : undefined)}
-      </StyledSlider>
-    );
+  if (max < min) {
+    return <></>;
```
jtbandes, line 138: Rather than rendering nothing we could just render as normal and pass `undefined` to the renderSlider function.

Separately, the abstraction around `renderSlider` is probably not necessary any more and this component can be simplified.

defunctzombie: For now I've made the code call renderSlider(undefined) when the range is inverted. There was already support for passing undefined to the renderSlider function.

I agree there are some disconnected abstractions here within all the scrubber components. Something to tackle another time.

PR #1807, opened 2021-09-08
Internal: remove react-document-events

defunctzombie:
**User-Facing Changes**
None

**Description**
It is no longer used, has no types, and looks unmaintained.


defunctzombie:
Blocked by https://github.com/foxglove/studio/pull/1805

packages/studio-base/src/components/Panel.tsx
```diff
@@ -430,12 +430,16 @@ export default function Panel<
...
     );
 
-    const onBlurDocument = useCallback(() => {
-      exitFullScreen();
-      setCmdKeyPressed(false);
-      setShiftKeyPressed(false);
-      onReleaseQuickActionsKey();
-    }, [exitFullScreen, onReleaseQuickActionsKey]);
+    /* Ensure user exits full-screen mode when leaving window, even if key is still pressed down */
+    const visibilityState = useVisibilityState();
```
jtbandes, line 434: I think blur is different from visibilityState==="hidden"; is that change intentional?

defunctzombie: yea - you are right - I was trying stuff out. No this should stick with blur if we want this functionality. I'm just trying to rid us of this random module that has this only use.

defunctzombie: changed to use the tried and true useEffect

jtbandes (approved):

PR #1813, opened 2021-09-09
Fix: indicate topics with bad frame id

defunctzombie:
**User-Facing Changes**
TBD

**Description**
Surface scene builder errors for topics with bad frame ids to the topic tree.

Follow-up from https://github.com/foxglove/studio/pull/1797 to actually surface the errors for the bad frame id topics.


defunctzombie:
Closing. We've decided that this is not a user-actionable error and instead documented that we do not currently do interpolation in studio.

PR #1814, opened 2021-09-09
Clicking a chart in the State Transitions panel will seek to the corresponding timestamp

esthersweon:
**User-Facing Changes**
Clicking on a chart in the State Transitions panel will seek to the corresponding timestamp (similar to clicking on a chart in the Plot panel).

**Description**

- Used the `MessagePipelineContext`'s `seekPlayback` function to implement feature.
- Did some cleanup around the Plot panel's code for the same functionality (i.e. use `useMessagePipelineGetter` to ensure the `onClick` callback doesn't change unnecessarily).

### DIFF
--- a/packages/studio-base/src/panels/StateTransitions/index.tsx
+++ b/packages/studio-base/src/panels/StateTransitions/index.tsx
@@ -13,23 +13,29 @@
 
 import { ChartOptions, ScaleOptions } from "chart.js";
 import { uniq } from "lodash";
-import { ComponentProps, useMemo, useState } from "react";
+import { ComponentProps, useCallback, useMemo, useState } from "react";
 import { useResizeDetector } from "react-resize-detector";
 import stringHash from "string-hash";
 import styled, { css } from "styled-components";
 import tinycolor from "tinycolor2";
 
 import { subtract as subtractTimes, toSec } from "@foxglove/rostime";
+import { add as addTimes, fromSec } from "@foxglove/rostime";
 import * as PanelAPI from "@foxglove/studio-base/PanelAPI";
 import Button from "@foxglove/studio-base/components/Button";
 import MessagePathInput from "@foxglove/studio-base/components/MessagePathSyntax/MessagePathInput";
 import useMessagesByPath from "@foxglove/studio-base/components/MessagePathSyntax/useMessagesByPath";
+import {
+  MessagePipelineContext,
+  useMessagePipeline,
+} from "@foxglove/studio-base/components/MessagePipeline";
 import Panel from "@foxglove/studio-base/components/Panel";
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/panels/StateTransitions/index.tsx
```diff
@@ -389,6 +395,18 @@ const StateTransitions = React.memo(function StateTransitions(props: Props) {
     refreshMode: "debounce",
   });
 
+  const seek = useMessagePipeline((ctx: MessagePipelineContext) => ctx.seekPlayback);
```
defunctzombie, line 398: The selector needs to either be wrapped in a useCallback or hoisted to a stable function outside of the component similar to https://github.com/foxglove/studio/blob/main/packages/studio-base/src/components/PlaybackControls/index.tsx#L38. Otherwise this makes a new function on every render and useMessagePipeline will think it needs to run that selector again. In this case I think seek will remain stable so it would be ok - but still better to use that practice.

packages/studio-base/src/panels/StateTransitions/index.tsx
```diff
@@ -389,6 +395,18 @@ const StateTransitions = React.memo(function StateTransitions(props: Props) {
...
+  const onClick = useCallback(
+    ({ x: seekSeconds }: OnChartClickArgs) => {
+      if (seekSeconds == undefined) {
+        return;
+      }
+      const seekTime = addTimes(startTime, fromSec(seekSeconds));
+      seek(seekTime);
+    },
+    [seek, startTime],
```
defunctzombie, line 407: Consider the `useMessagePipelineGetter()` approach here since the onClick callback doesn't need to change when seek or startTime change. It only cares about those values when a click happens:

https://github.com/foxglove/studio/blob/main/packages/studio-base/src/components/PlaybackControls/index.tsx#L54

defunctzombie:
👍

packages/studio-base/src/panels/StateTransitions/index.tsx
esthersweon: Ah - will make that change in the Plot panel code as well (also changing the `onClick` callback every time `seek` and `startTime` change): https://github.com/foxglove/studio/blob/main/packages/studio-base/src/panels/Plot/index.tsx#L352

packages/studio-base/src/panels/StateTransitions/index.tsx
esthersweon: This makes sense; addressed.

esthersweon: Not doing this, as I'll be getting `seekPlayback` with the `useMessagePipelineGetter()` approach you mention in your next comment.

PR #1815, opened 2021-09-09
Fix: use Tooltip component for Playback Control buttons

### LINKED ISSUE #1770: Seek button tooltips get stuck easily
Tooltips on the seek button often get stuck when moving the mouse randomly or clicking on the buttons.

https://user-images.githubusercontent.com/14237/132081464-364ce0a0-b4a1-46f6-ad3b-52bb94ba7bd0.mov

- Version: main @ 170e38ab
- OS: macOS

defunctzombie:
**User-Facing Changes**
Tooltips for playback controls no longer get stuck when you leave the button with your cursor.

**Description**
Fixes an issue with the tooltip getting stuck open on the playback
controls. This was due to mouse events not firing for disabled buttons.

This change sets the allowDisabledFocus flag for the HoverableIconButton and
uses forwardRef to allow the HoverableIconButton to exist in a <Tooltip> component parent.

Additionally, this change wraps each <Tooltip> in PlaybackControls within a <StackItem>. Since the
<Tooltip> adds a <span> component for the "layer", the gap setting on our <Stack> would result in
the buttons shifting around when the tooltip was active.

Fixes #1770

### DIFF
--- a/packages/studio-base/src/components/Tooltip.tsx
+++ b/packages/studio-base/src/components/Tooltip.tsx
@@ -9,7 +9,7 @@ import {
   useTheme,
   ICalloutContentStyles,
 } from "@fluentui/react";
-import { Fragment, useCallback, useRef, useState } from "react";
+import { useCallback, Fragment, useRef, useState } from "react";
 
 type Contents = React.ReactNode | (() => React.ReactNode);
 
@@ -150,6 +150,7 @@ export default function Tooltip(
   const { ref, tooltip } = useTooltip(props);
   const child = React.Children.only(children);
   const host = React.cloneElement(child, { ref });
+
   return (
     // When studio-base is packaged for npm, we saw strange issues where React would warn about
     // missing keys on these children, so we add explicit keys.
--- a/packages/studio-base/src/components/HoverableIconButton.tsx
+++ b/packages/studio-base/src/components/HoverableIconButton.tsx
@@ -3,31 +3,53 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
 import { IconButton, IButtonProps, IIconProps } from "@fluentui/react";
-import { useState } from "react";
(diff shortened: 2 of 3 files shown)


defunctzombie:
@jtbandes I attempted to fix #1770 but this doesn't seem to fix the issue reliably either. Not sure if this is now a bug in fluent or in some other mouse event intercept we are doing elsewhere in the app. Prior to this fix I was able to get the tooltip stuck in the PlaybackControls story. With these changes I was not able to do that but am still able to get it stuck in the app.

packages/studio-base/src/components/Tooltip.tsx
```diff
@@ -146,16 +147,66 @@ export function useTooltip({
...
+  }[placement];
+
   return (
-    // When studio-base is packaged for npm, we saw strange issues where React would warn about
-    // missing keys on these children, so we add explicit keys.
-    <>
-      <Fragment key="host">{host}</Fragment>
-      <Fragment key="tooltip">{tooltip}</Fragment>
-    </>
+    <TooltipHost
```
jtbandes, line 196: A problem I found with using TooltipHost was that it inserted an extra element in the dom tree, so it would mess up various CSS/layout attributes set on the wrapped component. Hence the comment on L35-36.

defunctzombie: I wonder if that accounts for the story diff with the checkboxes.

Interestingly, when I used the non-TooltipHost version of <Tooltip> around the PlaybackControls HoverButtons that produced undesirable shift in the button on hover. This was not true of the TooltipHost version.

defunctzombie: Probably keeping the cloneElement approach is better - I just wanted to rule out that itself being the issue.

defunctzombie:
@jtbandes I found the root issue (events disabled on disabled buttons - there are some posts online about this). I've updated the PR description to highlight the changes made and why.

PR #1825, opened 2021-09-09
Display org name/slug in account pane

jtbandes:
**User-Facing Changes**
None (feature flagged)

**Description**
image


packages/studio-base/src/components/AccountSettingsSidebar/AccountInfo.stories.tsx
```diff
@@ -22,6 +22,9 @@ export const SignedIn = (): JSX.Element => {
   const me = {
     id: "fake-userid",
     orgId: "fake-orgid",
+    orgDisplayName: "Fake Org",
```
defunctzombie, line 25: Related: https://github.com/foxglove/console/issues/146

PR #1831, opened 2021-09-10
Refactor ImageView and Dropdown styles to use CSS-in-JS

2metres:
**User-Facing Changes**
None

**Description**
Refactor ImageView/ImageCanvas and Dropdown styles to use CSS-in-JS. 

Part of larger style approach simplification #1691


packages/studio-base/src/panels/ImageView/ImageCanvas.tsx
```diff
@@ -452,24 +513,22 @@ export default function ImageCanvas(props: Props): JSX.Element {
...
       <KeyListener keyDownHandlers={keyDownHandlers} />
-      {error && <SErrorMessage>Error: {error.message}</SErrorMessage>}
-      {renderError && <SErrorMessage>Error: {renderError.message}</SErrorMessage>}
+      {error && <div className={classes.errorMessage}>Error: {error.message}</div>}
+      {renderError && <div className={classes.errorMessage}>Error: {renderError.message}</div>}
       <canvas
         onContextMenu={onCanvasContextMenu}
         {...panZoomHandlers}
-        style={style}
+        style={{
```
defunctzombie, line 524: Why do we use classes for the divs and not for this?

defunctzombie:
LGTM

packages/studio-base/src/panels/ImageView/ImageCanvas.tsx
2metres: Excellent question. Will fix up now

Issue #1820, opened 2021-09-09
Add ability to set and display a title for Plot panels


esthersweon:
A plot title like "Kinematics" would be more useful at a glance than a hamburger menu (that contains long topic paths that run off the panel's edge and cover plot data).

The title can be displayed in a full-width bar at the top of each Plot panel. 

This solves several problems:
- Makes it easier for users to remember what a given plot in a Foxglove Studio layout is for
- Makes it possible to move the hamburger menu containing the topic paths into the sidebar
   - Saves precious layout real estate and avoids covering plot data on panel hover
- Creates an area where other state controls (not persisted in layout) can live (e.g. [syncing Plot panels in a layout](https://github.com/foxglove/studio/issues/1732))

Issue #1732, opened 2021-08-31
Confusing synchronization behavior in plot panels


defunctzombie:
**Current Experience**
Timeseries plots (and state transition panels) are pan/zoom synchronized. When a user pan/zooms one timeseries plot, other timeseries plots in the layout also pan/zoom by the same amount. This keeps the timeseries x min/max range the same across all plots. The intent for this feature is to easily allow a user to zoom into one plot to closer inspect one section of timeseries data and have the other plots zoom into the same data for comparison. Without this synchronization the user would need to pan/zoom each plot individually if they want to get a detail view on a specific section of the timeline.

When a user pan/zooms a plot - that plot is considered manually pan/zoomed by the user. The behavior is that when a user goes to pan/zoom another plot panel the first no longer synchronizes. This is shown below:

https://user-images.githubusercontent.com/84792/131562647-c89b275b-9f73-458a-8e25-d8e30ea79d32.mp4

This is a somewhat surprising user experience since just panning the plot doesn't necessarily imply you wanted to unlock the plot from other plots.

The "reset view" button does not make it so obvious that clicking it will "re-lock" the panel. The user may want to keep the current view but keep the panel locked for pan/zoom.

**Desired Experience**
The plot panels should remain locked for pan/zoom unless the user takes an explicit action. Maybe a control on the panel itself?

While it may add to the plot view clutter - I like the idea of a control on the panel rather than panel settings since it might be something you want to toggle as you use the panel. Changing in panel settings is a more cumbersome option.

We could also look at a right-click context menu for this setting. Right now we do not do a lot with right-click menus on panels and they raise questions of discoverability.

rgov:
As I understand it:

1. All timestamp-based plots start synchronized.
2. Panning or zooming a synchronized plot will also adjust the view of other synchronized plots, but
3. After completing a pan/zoom, a plot loses its synchronized status.

Here (3) violates my expectations; I would want a plot to remain synchronized after pan/zoom.

I would like to hold a modifier key while panning/zooming to desynchronize the plot from others. When desynchronized, a button should appear to reset. Note there are two meanings of "reset":

1. Reset the desynchronized view to match the synchronized views (if any exist).
2. Reset the view to the default range.

I think the existing "Reset View" button is too large and obscures too much of the plot.

defunctzombie:
This PR https://github.com/foxglove/studio/pull/1823 keeps plot panels always in-sync.

[...]

PR #1832, opened 2021-09-10
Add Plot panel setting for title

esthersweon:
**User-Facing Changes**
All Plot panels can now be given a human-readable title in the panel settings. Will default to being displayed as "Untitled".

Screen Shot 2021-09-10 at 7 51 55 AM
Screen Shot 2021-09-10 at 7 51 48 AM

**Description**
Empty titles will default to being displayed as "Untitled". In the future, this title bar can be used to include other state controls (e.g. https://github.com/foxglove/studio/issues/1732).

Resolves https://github.com/foxglove/studio/issues/1820.

### DIFF
--- a/packages/studio-base/src/panels/Plot/index.tsx
+++ b/packages/studio-base/src/panels/Plot/index.tsx
@@ -152,6 +152,7 @@ function selectEndTime(ctx: MessagePipelineContext) {
 function Plot(props: Props) {
   const { saveConfig, config } = props;
   const {
+    title,
     followingViewWidth,
     paths: yAxisPaths,
     minYValue,
@@ -355,6 +356,7 @@ function Plot(props: Props) {
   return (
     <Flex col clip center style={{ position: "relative" }}>
       <PanelToolbar helpContent={helpContent} floating />
+      <div>{title ?? "Untitled"}</div>
       <PlotChart
         isSynced={xAxisVal === "timestamp"}
         paths={yAxisPaths}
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/panels/Plot/index.tsx
```diff
@@ -393,6 +396,7 @@ const configSchema: PanelConfigSchema<PlotConfig> = [
 ];
 
 const defaultConfig: PlotConfig = {
+  title: "",
```
defunctzombie, line 399: lets leave this undefined in the default? Or does that not work?

defunctzombie:
👍

packages/studio-base/src/panels/Plot/index.tsx
esthersweon: Changed to `undefined`.

packages/studio-base/src/panels/Plot/index.tsx
```diff
@@ -355,6 +356,7 @@ function Plot(props: Props) {
   return (
     <Flex col clip center style={{ position: "relative" }}>
       <PanelToolbar helpContent={helpContent} floating />
+      <div>{title ?? "Untitled"}</div>
```
jtbandes, line 359: Seems a little wasteful of space to always show Untitled when the user hasn't configured a title? So there is no way to explicitly remove the title?

esthersweon: We were thinking that in the future, this title bar can be used to include other state controls (e.g. #1732).

Issue #1819, opened 2021-09-09
Cache CurrentUser information for use when offline


jtbandes:
Currently, the CurrentUserProvider calls the `/me` API on every app launch to fetch the user info. If the computer is offline, but the user was previously logged in, we should cache this info so the app doesn't appear to be logged out if the request fails.

defunctzombie:
One way to handle this might be to only look for the access token being present to determine if "signed in" rather than the presence of the user. The user information and presence is only needed on certain panels?

jtbandes:
The user id is needed since it's used to determine the cache location.

PR #1836, opened 2021-09-11
Fix: keep the user signed in when opening the app offline

defunctzombie:
**User-Facing Changes**
None - unreleased feature

**Description**
If the user is currently offline, studio will still open and show their
account information.

Fixes: #1819

### DIFF
--- a/packages/studio-base/src/providers/ConsoleApiCurrentUserProvider.tsx
+++ b/packages/studio-base/src/providers/ConsoleApiCurrentUserProvider.tsx
@@ -2,13 +2,13 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
-import { PropsWithChildren, useCallback, useState } from "react";
+import { PropsWithChildren, useCallback, useMemo } from "react";
 import { useAsync, useLocalStorage } from "react-use";
 
 import { useShallowMemo } from "@foxglove/hooks";
 import Logger from "@foxglove/log";
 import { useConsoleApi } from "@foxglove/studio-base/context/ConsoleApiContext";
-import CurrentUserContext from "@foxglove/studio-base/context/CurrentUserContext";
+import CurrentUserContext, { User } from "@foxglove/studio-base/context/CurrentUserContext";
 
 const log = Logger.getLogger(__filename);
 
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/providers/ConsoleApiCurrentUserProvider.tsx
```diff
@@ -22,31 +22,57 @@ export default function ConsoleApiCurrentUserProvider(
...
+    undefined,
+    {
+      raw: false,
+      serializer: (value: User) => JSON.stringify(value),
+      deserializer: (value: string) => JSON.parse(value) as User,
+    },
+  );
+
+  // We need to set the api auth header info before any other parts of the app can use the api
+  useMemo(() => {
```
defunctzombie, line 41: This is a useMemo with a side effect - but I couldn't come up with a nicer alternative to setting the bearer token if available. I suppose one option is setting it when making the console api - but that hasn't read local storage yet.

packages/studio-base/src/providers/ConsoleApiCurrentUserProvider.tsx
```diff
@@ -22,31 +22,57 @@ export default function ConsoleApiCurrentUserProvider(
...
+  }, [api, bearerToken, setCachedCurrentUser]);
 
   const signOut = useCallback(async () => {
     removeBearerToken();
+    removeCachedCurrentUser();
     await api.signout();
-  }, [api, removeBearerToken]);
+  }, [api, removeBearerToken, removeCachedCurrentUser]);
+
+  const currentUser = useMemo<User | undefined>(() => {
```
jtbandes, line 68: This useMemo has no effect

packages/studio-base/src/providers/ConsoleApiCurrentUserProvider.tsx
jtbandes: Why did this need to move out of the useAsync?

defunctzombie: Since we no longer conditionally render the provider on the `loading` state of the useAsync when we have a cached user record, we need to set the auth header for the api so downstream users wanting to make authenticated requests (i.e. layout api provider) are able to do so. When this is in the useAsync, it doesn't set the auth header before the downstream components make api requests.

packages/studio-base/src/providers/ConsoleApiCurrentUserProvider.tsx
defunctzombie: Huh - yea this went through a few iterations and I guess the final one ended up being a no-op :D

Issue #1837, opened 2021-09-11
Diagnostics Panel: Allow filtering diagnostics by level


jessicaaustin:
I'm interested in adding a new setting for the Diagnostics panel that allows filtering diagnostics by level. 

This would allow me to create a layout with two diagnostics panels: one that has only "bad" diagnostics, and one that has all diagnostics. This separation reduces cognitive load for the user -- they can look in one place for everything that's going wrong, and ignore the rest.

Note: This is similar to how rqt Robot Monitor works:

![rqt_robot_monitor_example](https://user-images.githubusercontent.com/1159930/132927692-365e11b1-bb3b-4ba9-8968-f590921a0ba7.png)

jessicaaustin:
See https://github.com/foxglove/studio/pull/1838 for an example of me playing around with this concept. 

Here, I've added a simple boolean to check if a diagnostic is OK or not. 

Probably having checkboxes to show/hide each level makes more sense?

jtbandes:
Thanks for your interest and contribution! We've implemented a more detailed level filter in https://github.com/foxglove/studio/pull/1877 that will be in the next release. Let us know what you think 😃

jessicaaustin:
@jtbandes Nice, I like it! I pulled down `main` and tried it out against one of our bag files, and was easily able to split out "bad" diagnostics from everything else.

PR #1838, opened 2021-09-11
Spike: add excludeOk setting for diagnostics summary panel

jessicaaustin:
See https://github.com/foxglove/studio/issues/1837

This PR is a spike; I was exploring how to add new settings to a panel. I haven't run tests, and this probably isn't the right approach anyway.

Example of how this looks. The top panel is only "bad" diagnostics. The second panel is all diagnostics.

Selection_689

### DIFF
--- a/packages/studio-base/src/panels/diagnostics/DiagnosticSummary.stories.tsx
+++ b/packages/studio-base/src/panels/diagnostics/DiagnosticSummary.stories.tsx
@@ -110,6 +110,22 @@ export function Filtered(): JSX.Element {
   );
 }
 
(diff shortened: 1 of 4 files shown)


defunctzombie:
As an alternative - on the Rosout panel the equivalent filter is done via a dropdown within the panel itself. Once nice property of that is that it is obvious that these diagnostics are only errors, and these are only warnings? Maybe it should be a multi-select dropdown.

image

esthersweon:
Closing in favor of https://github.com/foxglove/studio/pull/1877.

PR #1839, opened 2021-09-11
Internal: RandomAccessPlayer cleanup

defunctzombie:
**User-Facing Changes**
None

**Description**
- Mark some methods `private`
- Use an async mutex to prevent multiple _read calls


defunctzombie:
Some random little bits.

PR #1844, opened 2021-09-13
Add offline states to layout browser UI and disable menu items

jtbandes:
**User-Facing Changes**
None (feature flagged)

**Description**
Add `isOnline` property and `onlinechange` event to LayoutManager. The LayoutManagerProvider sets this based on the `useNetworkState` hook.

Add indicator to UI when network is offline:

image

Disable menu items that don't work offline:
image
image


defunctzombie:
👍 great UX enhancement

PR #1852, opened 2021-09-14
Refactor ExpandingToolbar component to use CSS-in-JS

2metres:
**User-Facing Changes**
None

**Description**
Refactor ExpandingToolbar to use CSS-in-JS, Part of larger styling approach simplification #1691


defunctzombie:
👍

Issue #1857, opened 2021-09-14
App re-mounts when signing in or out


jtbandes:
When signing in or out, the CurrentUserProvider re-mounts its children, which causes the sidebar to close rather than staying on the account pane.

This is probably a regression from https://github.com/foxglove/studio/pull/1836.

**Expected Behavior**
Account pane stays open for continuity of the login experience.

PR #1862, opened 2021-09-15
Fix: avoid remounting all children when signing out

defunctzombie:
**User-Facing Changes**
None - unreleased feature

**Description**
Only the first load through the ConsoleApiCurrentUserProvider should
avoid rendering children. Future changes to the provider state should
keep render the children with the updated value of the context.

This fixes a regression I introduced in #1836

Fixes: #1857


jtbandes (approved):
🎉

jtbandes:
FWIW this should be unit testable – use enzyme and make a child that detects how many times it has unmounted.

Issue #1861, opened 2021-09-15
State transitions panel is messed up


amacneil:
**Description**
State transitions panel is messed up (with welcome layout):

![image](https://user-images.githubusercontent.com/637671/133350687-3be35c5c-6b69-44f2-af34-07b509965a52.png)

- Version: v0.18
- OS: linux

PR #1827, opened 2021-09-10
Feature: add preloaded data to state transitions

defunctzombie:
**User-Facing Changes**
State transition panel now shows data from the entire bag.

**Description**
Display available preloaded data in state transitions.
    
Fixes: #689

PR #1863, opened 2021-09-15
Fix State Transitions y alignment for preloaded data

jtbandes:
**User-Facing Changes**
Fixes an issue with State Transitions where multiple preloaded datasets would overlap.

**Description**
Fixes https://github.com/foxglove/studio/issues/1861, introduced in #1827.


defunctzombie:
LGTM

Issue #1860, opened 2021-09-15
Plot tooltips are getting "stuck"


amacneil:
**Description**

When mousing around the plot, tooltips often get "stuck", leading to more tooltips overlapping them, and the tooltip not going away when my mouse leaves the panel.

https://user-images.githubusercontent.com/637671/133350547-afca4113-f606-4c42-8286-f323c57068ba.mp4

- Version: v0.18 and latest main
- OS: linux

PR #1864, opened 2021-09-15
Fix: prevent late hover events from triggering chart hover

defunctzombie:
**User-Facing Changes**
Tooltips appear and disappear as the user hovers over the plot and between plots.

**Description**
 Fix: prevent late hover events from triggering chart hover
    
 Ignore hover responses from the rpc endpoint if the mouse is no longer on the chart component. This prevents late hover events from triggering onHover and showing tooltips in downstream TimeBasedChart component.
    
Fixes: #1860

### DIFF
--- a/packages/studio-base/src/components/Chart/index.tsx
+++ b/packages/studio-base/src/components/Chart/index.tsx
@@ -342,22 +342,37 @@ function Chart(props: Props): JSX.Element {
     [rpcSend],
   );
 
+  // Since hover events are handled via rpc, we might get a response back when we've
+  // already hovered away from the chart. We gate calling onHover by whether the mouse is still
+  // present on the component
+  const mousePresetRef = useRef(false);
+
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/components/Chart/index.tsx
```diff
@@ -342,23 +342,38 @@ function Chart(props: Props): JSX.Element {
     [rpcSend],
   );
 
+  // Since hover events are handled via rpc, we might get a response back when we've
+  // already hovered away from the chart. We gate calling onHover by whether the mouse is still
+  // present on the component
+  const mousePresetRef = useRef(false);
```
jtbandes, line 348:
```suggestion
  const mousePresentRef = useRef(false);
```
```suggestion
  const mouseInsideRef = useRef(false);
```

amacneil (approved):
issue is fixed for me on this branch

PR #1867, opened 2021-09-15
Migrate old savedProps to configById when loading layouts

jtbandes:
**User-Facing Changes**
Fixed a crash when loading layouts created with older versions of Foxglove Studio or Webviz.

**Description**
Some migration logic was removed 2 releases ago, in https://github.com/foxglove/studio/pull/1679. Reinstate some migrations so that we can still support old layouts.

Tested manually by importing an old layout that was previously crashing.


defunctzombie:
👍

Issue #1870, opened 2021-09-15
Put all the team layout hazard zones behind a feature flag


amacneil:
Shouldn't be visible in a default dev build

PR #1872, opened 2021-09-15
Hide dev-only layout debug controls behind feature flag

jtbandes:
**User-Facing Changes**
None

**Description**
Fixes https://github.com/foxglove/studio/issues/1870

### DIFF
--- a/packages/studio-base/src/components/ExperimentalFeatureSettings.tsx
+++ b/packages/studio-base/src/components/ExperimentalFeatureSettings.tsx
@@ -54,6 +54,13 @@ const features: Feature[] = [
     description: <>Enable team layout sharing when signed in to Foxglove Studio.</>,
   },
 ];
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/components/ExperimentalFeatureSettings.tsx
```diff
@@ -54,6 +54,13 @@ const features: Feature[] = [
     description: <>Enable team layout sharing when signed in to Foxglove Studio.</>,
   },
 ];
+if (process.env.NODE_ENV === "development") {
```
amacneil, line 57: any reason to not make these features available in prod, since they will still be behind a flag? (e.g. if we need to debug something with prod build)

jtbandes: 🤷 just seems like we don't really want any users discovering this

Issue #1871, opened 2021-09-15
Investigate plot lag on prod build


amacneil:
More info coming soon

jtbandes:
The plot in question had ~50,000 points and part of the slowness was simply rendering all the arcs to draw the points. Perhaps the time has come to stop using 2D canvas to render plots and switch to WebGL:

- https://danchitnis.github.io/webgl-plot-examples/vanilla/index.html
- https://blog.scottlogic.com/2020/05/01/rendering-one-million-points-with-d3.html

defunctzombie:
Give this a try https://github.com/foxglove/studio/pull/1874

PR #1874, opened 2021-09-16
Fix: downsample all points within an interval for timeseries plots

defunctzombie:
**User-Facing Changes**
The plot panel is less laggy when viewing a dataset with high volatility within small intervals.

**Description**
Change the plot downsample algorithm to consider all points within an interval when performing downsampling. Previously, the downsample algorithm would only look at the previous point to determine if it should be downsampled away. For datasets where consecutive point distance is greater than the threshold no downsampling would occur. This resulted in plotting several tens of thousands of points and a slowdown of the plot.

Rather than only consider consecutive points, the new downsampling function is split into two. One for timeseries data and one for scatter data. The scatter data uses a sparse array to ignore previously seen x/y locations. The timeseries downsample
assumes time sorted data, and tries to preserve the shape of the data including min/max within an interval.

Fixes: #1871


packages/studio-base/src/components/TimeBasedChart/downsample.ts
```diff
@@ -39,14 +37,103 @@ export default function downsample(dataset: DataSet, bounds: DownsampleBounds):
...
+    }
+
+    // we need to keep the first points outside the set
+    if (datum.x < bounds.x.min || datum.x > bounds.x.max) {
+      continue;
+    }
+
+    if (datum.y < bounds.y.min || datum.y > bounds.y.max) {
+      continue;
+    }
```
jtbandes, line 64: - Multiply bounds.width by 1.5x or so to avoid seeing blank data when panning the plot.
- Always keep at least 1 point outside the plot bounds.

amacneil:
Simple layout with `cal_loop.bag`, plot `/imu/data.linear_acceleration.x`, and playback performance panel:

Screenshot from 2021-09-17 12-21-59

Latest main: ~1.5 fps
This branch: ~13.5 fps

So it's better, and can at least play back at 1x now. Plenty of room for improvement (webgl plots?) but this is still worth landing.

PR #1881, opened 2021-09-16
Do not display Plot panel's title, if not set

esthersweon:
**User-Facing Changes**
Do not display "Untitled" above a plot, if a Plot panel's title is not set


packages/studio-base/src/panels/Plot/index.tsx
```diff
@@ -356,7 +356,7 @@ function Plot(props: Props) {
   return (
     <Flex col clip center style={{ position: "relative" }}>
       <PanelToolbar helpContent={helpContent} floating />
-      <div>{title ?? "Untitled"}</div>
+      {title ? <div>{title}</div> : undefined}
```
defunctzombie, line 359: `{title && <div>{title}</div>}` is another common shorthand for this

PR #1883, opened 2021-09-17
Team layouts copy changes

amacneil:
**User-Facing Changes**
Minor updates to copy in team layouts feature

**Description**
See PR diff

### DIFF
--- a/packages/studio-base/src/components/LayoutBrowser/LayoutRow.tsx
+++ b/packages/studio-base/src/components/LayoutBrowser/LayoutRow.tsx
@@ -226,22 +226,21 @@ export default function LayoutRow({
       disabled: layoutIsShared(layout) && !isOnline,
       secondaryText: layoutIsShared(layout) && !isOnline ? "Offline" : undefined,
     },
-    {
+    // For shared layouts, duplicate first requires saving or discarding changes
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/components/LayoutBrowser/LayoutRow.tsx
```diff
@@ -226,37 +226,36 @@ export default function LayoutRow({
       disabled: layoutIsShared(layout) && !isOnline,
       secondaryText: layoutIsShared(layout) && !isOnline ? "Offline" : undefined,
     },
-    {
+    // For shared layouts, duplicate first requires saving or discarding changes
```
jtbandes, line 229: This seems like it could be a little confusing why the menu item isn’t available in certain circumstances. I guess being disabled can also be confusing 🤷

amacneil: the menu item is still there, it's just in a different place (above the separator). I guess I should check the logic, but I'm pretty sure you will always either see one or the other

amacneil: yeah just checked, one of these items will always show

PR #1885, opened 2021-09-17
Add mcap package with Message Capture parsing support

jtbandes:
**User-Facing Changes**
None

**Description**
Adds a workspace package `@foxglove/mcap` with parsing support for the nascent mcap file format. This PR contains no Studio integration, just adding the package since it's begun to stabilize.

The package includes a validate script which can be used to test with local .mcap files:

```sh
$ yarn workspace @foxglove/mcap validate /path/to/file.mcap [--deserialize [--dump]]
Reading gps.mcap
Read 886.10kiB in 89.54ms (9.66MiB/sec)
Record counts:
       1 Chunk
      22 ChannelInfo
   30445 Message
       1 Footer
```

Because the file format and especially the parser is still evolving, we'll keep it in the studio repo for now and later move it into a separate repo.


packages/mcap/package.json
```diff
@@ -0,0 +1,38 @@
+{
+  "name": "@foxglove/mcap",
```
defunctzombie, line 2: We might be able to get `mcap`

packages/mcap/src/McapReader.ts
```diff
@@ -0,0 +1,152 @@
...
+   * be processed after each chunk record itself.
+   */
+  includeChunks?: boolean;
+
+  /**
+   * When a compressed chunk is encountered, the entry in `decompressHandlers` corresponding to the
+   * compression will be called to decompress the chunk data.
+   */
+  decompressHandlers?: {
+    [compression: string]: (buffer: DataView, decompressedSize: bigint) => DataView;
```
defunctzombie, line 24: `buffer: Uint8Array` feels more conventional for "this is a buffer"

packages/mcap/src/McapReader.ts
```diff
@@ -0,0 +1,152 @@
...
+  };
+};
+
+/**
+ * A streaming reader for Message Capture files.
+ */
+export default class McapReader {
+  private buffer = new StreamBuffer(MCAP_MAGIC.length * 2);
```
defunctzombie, line 32: prefix these with `_` since they are still technically visible in the instance object to mark them as "don't touch"

packages/mcap/src/McapReader.ts
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  append(data: Uint8Array): void {
+    if (this.doneReading) {
+      throw new Error("Already done reading");
+    }
+    this.buffer.append(data);
+  }
+
+  nextRecord(): McapRecord | undefined {
```
defunctzombie, line 58: Comments for `nextRecord()` and `read` ?

Does this block? Example of how someone is meant to use this would be good in the comment as well.

packages/mcap/src/McapReader.ts
```diff
@@ -0,0 +1,152 @@
...
+  private includeChunks;
+  private doneReading = false;
+  private generator = this.read();
+
+  constructor({ includeChunks = false, decompressHandlers = {} }: McapReaderOptions = {}) {
+    this.includeChunks = includeChunks;
+    this.decompressHandlers = decompressHandlers;
+  }
+
+  done(): boolean {
```
defunctzombie, line 43: What's the benefit of marking the reader as done?

packages/mcap/src/McapReader.ts
```diff
@@ -0,0 +1,152 @@
...
+    return this.doneReading;
+  }
+
+  bytesRemaining(): number {
```
defunctzombie, line 47: Is this appended but unread bytes?

packages/mcap/src/McapReader.ts
```diff
@@ -0,0 +1,152 @@
...
+   */
+  decompressHandlers?: {
+    [compression: string]: (buffer: DataView, decompressedSize: bigint) => DataView;
+  };
+};
+
+/**
+ * A streaming reader for Message Capture files.
+ */
+export default class McapReader {
```
defunctzombie, line 31: This class could use a comment on usage - I generally like the minimal API surface tho 👍

packages/mcap/src/McapReader.ts
jtbandes: where are we on `#` support?

packages/mcap/src/McapReader.ts
jtbandes: yep

packages/mcap/src/McapReader.ts
jtbandes: largely useful for testing and asserting that there are no extraneous bytes in the stream

packages/mcap/src/McapReader.ts
defunctzombie: I think we need to change our targets to es2021 and we are good to go.

packages/mcap/src/McapReader.ts
defunctzombie: Other names I've seen for this in apis: `finalize`

jtbandes: this isn't a mutating method, it's a status check ... could rename it to `isDone`? or `get done()`?

defunctzombie: Oh I see - cause the done state is indicated by having read the end of the file. Ok.

Issue #1873, opened 2021-09-15
Could not find chart with id


amacneil:
**Description**
All I did was try to add a plot panel and I got this error.

![Screenshot from 2021-09-15 16-05-27](https://user-images.githubusercontent.com/637671/133524443-7932d052-48e3-41c9-9225-7fefd8800626.png)

amacneil:
Still happening on main

PR #1886, opened 2021-09-17
Fix: avoid sending chart rpc if the chart has unmounted

defunctzombie:
**User-Facing Changes**
Ideally none - if a user does encounter this error in production we will have a better indication of why.

**Description**

This change alters the logic around initializing, registering, and unregistering rpc workers.

Previously, the Chart logic would register itself in a memo, and unregister in a layout effect. In development, hot reloading would cause the effect to run which destroyed the id and unregistered the id. The component remained mounted and future rpc calls would fail because the ID was destroyed.

The new logic puts rpc registration and unregistration in the same effect. It also adds checks to see if the chart has already been initialized and throws an error indicating this. In production we should not expect to see this error as the effect should only run once on initial mount. If the error does happen, we know something caused the effect to run and will need to investigate further.

Fixes: #1873

### DIFF
--- a/packages/studio-base/src/components/Chart/index.tsx
+++ b/packages/studio-base/src/components/Chart/index.tsx
@@ -74,6 +74,12 @@ function rpcMouseEvent(event: React.MouseEvent<HTMLCanvasElement>) {
   };
 }
 
+type RpcSend = <T extends unknown>(
+  topic: string,
+  payload?: Record<string, unknown>,
+  transferables?: (Transferable | OffscreenCanvas)[],
+) => Promise<T>;
+
 // Chart component renders data using workers with chartjs offscreen canvas
 function Chart(props: Props): JSX.Element {
   const [id] = useState(() => uuidv4());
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/components/Chart/index.tsx
```diff
@@ -87,30 +91,44 @@ function Chart(props: Props): JSX.Element {
...
+
+  const rpcSendRef = useRef<RpcSend | undefined>();
+  const rpcSend = rpcSendRef.current;
+
+  useLayoutEffect(() => {
+    if (initialized.current) {
+      if (process.env.NODE_ENV === "development") {
+        throw new Error("Chart does not support hot-reloading - please reload the panel");
+      } else {
+        throw new Error("Chart has re-initialized expectedly");
```
jtbandes, line 108: unexpectedly?

packages/studio-base/src/components/Chart/index.tsx
```diff
@@ -87,30 +91,44 @@ function Chart(props: Props): JSX.Element {
...
-    return webWorkerManager.registerWorkerListener(id);
-  }, [id]);
+  type RpcSend = <T extends unknown>(
+    topic: string,
+    payload?: Record<string, unknown>,
+    transferables?: (Transferable | OffscreenCanvas)[],
+  ) => Promise<T>;
+
+  const rpcSendRef = useRef<RpcSend | undefined>();
+  const rpcSend = rpcSendRef.current;
```
jtbandes, line 101: It seems weird to have `[rpcSend]` as a dep in certain hooks below. Isn't the point of using a ref that the hook function bodies will access `rpcSendRef.current` to avoid unnecessary dependencies?

defunctzombie: Updated to use the ref directly

Issue #1622, opened 2021-08-11
Amplitude metrics not working on web build


esthersweon:
**Description**
Amplitude metrics are not working on studio.foxglove.dev.

- Version: 0.14.0
- OS: macOS

**Steps To Reproduce**

- Navigate to studio.foxglove.dev
- Look at Chrome console for errors
- See following Amplitude errors
<img width="813" alt="Screen Shot 2021-08-11 at 10 11 12 AM" src="https://user-images.githubusercontent.com/6993359/129073199-30afa436-0fef-4192-a6c6-c3d02059ce80.png">

**Expected Behavior**
No Amplitude errors in console; metrics correctly collected from web build.

**Actual Behavior**
Errors in console; metrics not being collected from web build.

jtbandes:
This bug is caused by https://github.com/webpack/webpack/issues/12960.

amacneil:
Pretty sure this is still not working

jtbandes:
One issue now is that in the web build, the app version is reported as 0.0.0:

https://github.com/foxglove/studio/blob/1d34d9672fbc8e4968f72c53c708cf9c03693c52/packages/studio-base/src/services/AmplitudeAnalytics.ts#L79-L81

(I don't think this is what's causing analytics to not show up at all though)

amacneil:
@esthersweon did this get fixed? If not, would be great if someone can look into this this week.

amacneil:
Pretty sure this is still not working

jtbandes:
Indeed, I don't think any work was done on this since my last comment (at least I didn't see any PRs go by)

PR #1887, opened 2021-09-18
Refactor Amplitude and fix web analytics

amacneil:
**User-Facing Changes**
None

**Description**
Amplitude was not working on web. Fixed and simplified analytics code.

Users can still opt out of analytics and/or crash reporting in the settings.

- Machine ID is no longer used, Amplitude already generates a Device ID
- Amplitude Device ID is passed to Sentry where relevant
- User ID is passed to Amplitude & Sentry where relevant
- Use Amplitude's existing opt out functionality instead of rolling our own
- Removed unnecessary logic around Sentry breadcrumbs (if crash reporting is opted out then Sentry is never initialized, but we can still call `Sentry.addBreadcrumb()` and it will be a no-op).

Closes https://github.com/foxglove/studio/issues/1622

### DIFF
--- a/desktop/preload/index.ts
+++ b/desktop/preload/index.ts
@@ -4,7 +4,6 @@
 
 import { init as initSentry } from "@sentry/electron";
 import { contextBridge, ipcRenderer } from "electron";
-import { machineId } from "node-machine-id";
 import os from "os";
 import { join as pathJoin } from "path";
 
@@ -73,10 +72,6 @@ window.addEventListener(
 
 const localFileStorage = new LocalFileStorage();
 
-// machineId() can sometimes take 500-2000ms on macOS
-// we fetch it early so that it is ready for Analytics.ts
-const machineIdPromise = machineId();
-
 const ctx: OsContext = {
   platform: process.platform,
   pid: process.pid,
@@ -97,9 +92,6 @@ const ctx: OsContext = {
       }
     }
     return output;
-  },
-  getMachineId: async (): Promise<string> => {
-    return await machineIdPromise;
   },
   getAppVersion: (): string => {
     return pkgInfo.version;
--- a/packages/studio-base/src/services/AmplitudeAnalytics.ts
+++ b/packages/studio-base/src/services/AmplitudeAnalytics.ts
@@ -2,116 +2,63 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
-import { addBreadcrumb as addSentryBreadcrumb, setUser as setSentryUser } from "@sentry/core";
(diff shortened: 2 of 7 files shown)


jtbandes:
What was the root cause of analytics not working on web?

How does amplitude generate their device id?

amacneil:
> What was the root cause of analytics not working on web?

I don't know exactly, but my theory is that setting `deviceId` to `00000000-0000-0000-0000-000000000000` had something to do with it.

> How does amplitude generate their device id?

Random string such as `r09x14Hu9rmVW-Fbdk1KR7`.

desktop/preload/index.ts
```diff
@@ -73,10 +72,6 @@ window.addEventListener(
 
 const localFileStorage = new LocalFileStorage();
 
-// machineId() can sometimes take 500-2000ms on macOS
-// we fetch it early so that it is ready for Analytics.ts
-const machineIdPromise = machineId();
```
defunctzombie, line 78: @jhurliman Do you remember why we added machineId in the first place?

packages/studio-base/src/services/AmplitudeAnalytics.ts
```diff
@@ -2,116 +2,64 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
-import { addBreadcrumb as addSentryBreadcrumb, setUser as setSentryUser } from "@sentry/core";
+import * as Sentry from "@sentry/core";
```
defunctzombie, line 5: Its nice not to import everything if we don't need everything from core. Its likely that sentry core is probably used in its entirely but also nice when modules are able to be used in parts. Keeps bundle sizes smaller (since they are already so large)

defunctzombie:
LGTM

defunctzombie:
LGTM

packages/studio-base/src/services/AmplitudeAnalytics.ts
amacneil: I did this for namespacing, but will revert

desktop/preload/index.ts
amacneil: I think it was just so that if someone cleared their electron cache (appdata) it wouldn't treat them as a different user. But this seems not worth the additional complexity.

If there is another reason happy to revert to previous behavior.

packages/studio-base/src/services/AmplitudeAnalytics.ts
jtbandes: FWIW I suspect the style of import (named or *) does not actually affect the bundle size. Webpack should be smarter than that

jtbandes:
For future reference how did you go about testing this? Make a prod build and look at the amplitude UI to ensure new events were actually showing up properly?

amacneil:
No need to make a prod build, I just set the api key env var while running `yarn serve` and verified in Amplify UI.

PR #1890, opened 2021-09-18
Report platform for desktop builds

amacneil:
**User-Facing Changes**
None

**Description**
Amplitude defaults to always tracking the platform as `Web`. Updated to report the OsContext platform.


jtbandes (commented):
LGTM — why bother capitalizing?

amacneil:
Only because their default `Web` (and other properties) were already capitalized.

PR #1900, opened 2021-09-22
Remove CSS modules from more components

2metres:
**User-Facing Changes**
None

**Description**
Refactor CSS modules out of `main/StudioWindow`, `components/TimeBasedChartTooltipContent`, and `components/TextContent` and into CSS-in-JS

Part of larger style approach refactor #1691


defunctzombie:
:shipit:

Issue #1901, opened 2021-09-22
Flag to disable account pane and sign in prompt in layout sidebar


jtbandes:
Once team layouts has shipped and feature flag removed, we still need a way in code to disable the login flow and shared layout features.

PR #1905, opened 2021-09-24
Don't show sign in prompt and account pane if ConsoleApi is not present

jtbandes:
**User-Facing Changes**
None (feature flagged)

**Description**
The Account tab of the sidebar and the sign-in prompt in the LayoutBrowser were previously shown when the Team Layouts feature flag was enabled. Now they are additionally gated on the presence of ConsoleApi, so when the feature flag is removed, these features will remain hidden if no ConsoleApi is provided.

Resolves https://github.com/foxglove/studio/issues/1901.

### DIFF
--- a/packages/studio-base/src/Workspace.tsx
+++ b/packages/studio-base/src/Workspace.tsx
@@ -11,7 +11,15 @@
 //   You may not use this file except in compliance with the License.
 
 import { makeStyles, Stack } from "@fluentui/react";
-import { useState, useEffect, useRef, useCallback, useMemo, useLayoutEffect } from "react";
+import {
+  useState,
+  useEffect,
+  useRef,
+  useCallback,
+  useMemo,
+  useLayoutEffect,
+  useContext,
+} from "react";
 import { useToasts } from "react-toast-notifications";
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/Workspace.tsx
```diff
@@ -137,6 +146,11 @@ export default function Workspace(props: WorkspaceProps): JSX.Element {
   const playerCapabilities = useMessagePipeline(selectPlayerCapabilities);
   const playerProblems = useMessagePipeline(selectPlayerProblems);
 
+  const [enableSharedLayouts = false] = useAppConfigurationValue<boolean>(
+    AppSetting.ENABLE_CONSOLE_API_LAYOUTS,
+  );
+  const supportsAccountSettings = useContext(ConsoleApiContext) != undefined && enableSharedLayouts;
```
jtbandes, line 152: @defunctzombie a consequence of our decision to make `useConsoleApi` (and other similar hooks) throw an error if the context is missing — here we have to explicitly `useContext(ConsoleApiContext)` because we want to check for it being undefined. Not really a problem, just want to call it out in case it sparks any ideas for cleaner or more consistent patterns.

Issue #1898, opened 2021-09-21
Amplitude warnings in dev builds


jtbandes:
In a dev build (yarn serve), when performing an action that would normally log an analytics event, I get the following warning in the console:

`[Amplitude] Invalid apiKey. Please set a valid apiKey with init() before calling logEvent()`

PR #1907, opened 2021-09-24
Skip logEvent if Amplitude API key is missing

amacneil:
**User-Facing Changes**
None

**Description**
Fix Amplitude console warning that was only happening during local development about calling `logEvent()` without calling `init()`. 

Fixes https://github.com/foxglove/studio/issues/1898

### DIFF
--- a/packages/studio-base/src/services/AmplitudeAnalytics.ts
+++ b/packages/studio-base/src/services/AmplitudeAnalytics.ts
@@ -17,12 +17,12 @@ import IAnalytics, {
 const os = OsContextSingleton; // workaround for https://github.com/webpack/webpack/issues/12960
 
 export class AmplitudeAnalytics implements IAnalytics {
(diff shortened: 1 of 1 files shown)


jtbandes (commented):
Maybe we should just make amplitudeApiKey non-optional for this class, and change the AnalyticsProvider to not instantiate the AmplitudeAnalytics if the api key is missing?

amacneil:
> Maybe we should just make amplitudeApiKey non-optional for this class, and change the AnalyticsProvider to not instantiate the AmplitudeAnalytics if the api key is missing?

Unfortunately this class is misnamed and does other stuff (Sentry breadcrumbs). It's unlikely that we have no Amplitude API key and still want sentry, but possible that someone making a custom build might want that?

jtbandes:
Ok, I guess this is fine then 🤷

PR #1913, opened 2021-09-28
Refactor PlotLegend.module.scss into CSS-in-JS

2metres:
**User-Facing Changes**
None

**Description**
Refactor PlotLegend.module.scss into CSS-in-JS, Part of larger styling approach refactoring #1691


defunctzombie:
LGTM

PR #1866, opened 2021-09-15
Update Console APIs to camelCase

jtbandes:
**User-Facing Changes**
None

**Description**
Fixes https://github.com/foxglove/studio/issues/1865

PR #1915, opened 2021-09-29
Migrate lowercase layout permission constants to uppercase

jtbandes:
**User-Facing Changes**
Team Layouts beta: fixed an issue where personal layouts created with older version of Studio would incorrectly become "team" layouts upon first login.

**Description**
These constants were changed from lowercase to uppercase in #1866, which was before we invited users to the team layouts beta. However, users may have created local personal layouts (which would still have the lowercase values) with older versions of the app before signing up for the beta. Due to an implementation detail, these old layouts were showing up under the "team" section and not getting uploaded to the server. Adding this migration ensures they are correctly treated as personal layouts and uploaded.

### DIFF
--- a/packages/studio-base/src/services/ILayoutStorage.ts
+++ b/packages/studio-base/src/services/ILayoutStorage.ts
@@ -129,7 +129,8 @@ export function migrateLayout(value: unknown): Layout {
   return {
     id: layout.id,
     name: layout.name ?? `Unnamed (${now})`,
-    permission: layout.permission ?? "CREATOR_WRITE",
+    permission:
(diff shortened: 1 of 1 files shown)


amacneil (approved):

amacneil:
Side note - if layouts have an unknown permission, should we not display them as team layouts? Just thinking for forwards compatibility in case we add more permission options in the future.

jtbandes:
Maybe so - that's just because of the `!==` here:

https://github.com/foxglove/studio/blob/acbdeb0e0142b871f8738d5ca454ba1e550a5ec6/packages/studio-base/src/services/ILayoutStorage.ts#L74-L78

PR #1916, opened 2021-09-29
Remove CSS Modules from components/Panel

2metres:
**User-Facing Changes**
None

**Description**
Remove CSS Modules from components/Panel, Part of larger style approach refactor #1691


defunctzombie:
LGTM

PR #1869, opened 2021-09-15
[WIP] FoxgloveDataPlatformPlayer

jhurliman:
**User-Facing Changes**

Yes

**Description**

TBD

PR #1919, opened 2021-09-29
Add data platform player with mcap streaming playback

jtbandes:
**User-Facing Changes**
None (for Data Platform private beta)

**Description**
Supersedes #1869 

- Adds FoxgloveDataPlatformPlayer and handler for `foxglove://open?type=foxglove-data-platform` URLs
- Player loads data in 15sec chunks, and parses the data as a stream. Loading stops at 15sec until playback reaches within 5sec of the next un-loaded range.
- New MessageMemoryCache class keeps track of pre-loaded ranges. Messages in the cache are fully parsed (or wrapped in lazy reader), not raw ArrayBuffers, for now. They are not currently segmented by topic, but if we want to support subscriber hints for preloading certain topics, we might want to change how this works (or just use 2 caches – one for preloaded topics and one for just-in-time topics).

Future work:
- [ ] Add correct schemaName (datatype) when it is available from /topics endpoint
- [ ] Evict data from memory cache as necessary
- [ ] Move message reader building out of `streamMessages()` so it can be done once up front
- [ ] Plumb total bytes received through to PlayerState
- [ ] Support seek backfill while paused
- [ ] https://github.com/foxglove/studio/pull/1919/files#r719059578


packages/studio-base/src/test/permutations.ts
```diff
@@ -0,0 +1,20 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+/**
+ * @returns An iterable producing all N! possible permutations of the items in `values`.
+ */
+export default function* permutations<T>(values: T[]): Iterable<T[]> {
```
jtbandes, line 8: moved

packages/studio-base/src/players/FoxgloveDataPlatformPlayer/streamMessages.ts
```diff
@@ -0,0 +1,136 @@
...
+          if (!isEqual(existingInfo.info, record)) {
+            throw new Error(`differing channel infos for for ${record.id}`);
+          }
+          break;
+        }
+        let parsedDefinitions;
+        let messageDeserializer;
+        if (record.encoding === "ros1") {
+          parsedDefinitions = parseMessageDefinition(new TextDecoder().decode(record.schema));
+          messageDeserializer = new LazyMessageReader(parsedDefinitions);
```
jtbandes, line 77: need to dedupe these deserializers rather than creating them for each /stream call – can happen in follow-up PR

defunctzombie:
> * They are not currently segmented by topic, but if we want to support subscriber hints for preloading certain topics, we might want to change how this works (or just use 2 caches – one for preloaded topics and one for just-in-time topics).

By topics makes sense to me since we don't want to request the same topic on the same range twice during a session. So if a user adds another panel, we want to avoid fetching any topics we already have and just fetch those that are missing.

packages/studio-base/src/Workspace.tsx
```diff
@@ -335,19 +336,47 @@ export default function Workspace(props: WorkspaceProps): JSX.Element {
...
+          },
+          { url: bagUrl },
+        );
+      } else if (type === "foxglove-data-platform") {
+        const start = url.searchParams.get("start") ?? "";
+        const end = url.searchParams.get("end") ?? "";
+        const seekTo = url.searchParams.get("seekTo") ?? undefined;
+        const deviceId = url.searchParams.get("deviceId");
+        if (!deviceId) {
+          log.warn(`Missing deviceId param in ${url}`);
```
defunctzombie, line 359: Probably should be toasts? Alternatively, could remove the error/warning handling here and pass the param strings to the selectSource options and let the player try reading them and reporting player problems. That feels more contained to the player which is generally a more favorable direction than parsing within Workspace.

packages/studio-base/src/components/NotificationModal.tsx
```diff
@@ -24,7 +24,7 @@ export default function NotificationModal({
 
   return (
     <Dialog
-      isOpen
+      hidden={false}
```
defunctzombie, line 27: 😭 reminds me of `isFailed: false` ♾️

packages/studio-base/src/players/FoxgloveDataPlatformPlayer/collateMessageStream.ts
```diff
@@ -0,0 +1,50 @@
...
+import { areEqual, Time } from "@foxglove/rostime";
+import { MessageEvent } from "@foxglove/studio-base/players/types";
+
+/**
+ * Given a stream of chunks of messages, track the changes in receiveTime and re-arrange the chunks
+ * so we can be certain each range is completely loaded when we add it to the cache.
+ *
+ * Assumes the incoming messages are in sorted order.
+ */
+export default async function* collateMessageStream(
```
defunctzombie, line 14: I dunno how I feel about more generators entering our lives.

packages/studio-base/src/players/FoxgloveDataPlatformPlayer/index.ts
```diff
@@ -0,0 +1,432 @@
...
+    end: string;
+    seek?: string;
+    deviceId: string;
+  };
+  metricsCollector: PlayerMetricsCollectorInterface;
+};
+
+const ZERO_TIME = Object.freeze({ sec: 0, nsec: 0 });
+
+const PRELOAD_THRESHOLD_SECS = 5;
```
defunctzombie, line 62: Lets make these member variables in the player - it might be nice to prepare for us providing these as player settings in a future

packages/studio-base/src/players/FoxgloveDataPlatformPlayer/index.ts
```diff
@@ -0,0 +1,432 @@
...
+    this._problemsById.set(id, problem);
+    this._problems = Array.from(this._problemsById.values());
+    if (!skipEmit) {
+      this._emitState();
+    }
+  }
+
+  // Potentially performance-sensitive; await can be expensive
+  // eslint-disable-next-line @typescript-eslint/promise-function-async
+  private _emitState = debouncePromise(() => {
```
defunctzombie, line 192: Can we remove this debouncePromise wrapping and use a different pattern for guarding emits? Something about this wrapper pattern has felt _off_ to me - like we should have a mutex to prevent multiple emit state but that the rate of emit state beyond that should not be gated by this wrapper. Its nicer to reason about a call to emit state emitting the state.

packages/studio-base/src/players/FoxgloveDataPlatformPlayer/index.ts
```diff
@@ -0,0 +1,432 @@
...
+
+  setSubscriptions(subscriptions: SubscribePayload[]): void {
+    log.debug("setSubscriptions", subscriptions);
+    this._requestedTopics = Array.from(new Set(subscriptions.map(({ topic }) => topic)));
+    this._clearPreloadedData();
+    this._startPreloadTaskIfNeeded();
+    this._emitState();
+  }
+
+  private _runPlaybackLoop = debouncePromise(async () => {
```
defunctzombie, line 252: Same comment here as above with debouncePromise. I'd probably like to see this with a mutex or similar lockout rather than the debounce. I think I worked on this in another PR and felt it was easier to read and reason about.

packages/studio-base/src/players/FoxgloveDataPlatformPlayer/index.ts
```diff
@@ -0,0 +1,432 @@
...
+
+      for await (const { messages, range } of collateMessageStream(stream, {
+        start: startTime,
+        end: endTime,
+      })) {
+        if (thisTask.signal.aborted) {
+          break;
+        }
+        log.debug("Adding preloaded chunk in", range, "with", messages.length, "messages");
+        this._preloadedMessages.insert(messages, range);
```
defunctzombie, line 351: This insert call feels backwards? range should be the first argument? Thinking similar to a map with key/values?

packages/studio-base/src/players/FoxgloveDataPlatformPlayer/collateMessageStream.ts
jtbandes: How else do write an emitter that can be consumed easily in an async function?

packages/studio-base/src/players/FoxgloveDataPlatformPlayer/index.ts
jtbandes: The rate of emitState has to be gated by the promise returned by the listener. debouncePromise is just abstracting the logic that holds onto that promise and waits for it before emitting the next state (and dedupes requests to emit state that happen while it’s waiting). We can chat about this but I have trouble envisioning an alternative that’s easier to read. If we determine a name for debouncePromise that captures its role better then we can happily rename it :)

packages/studio-base/src/players/FoxgloveDataPlatformPlayer/index.ts
jtbandes: Discussed in Slack – because of the backpressure built into this design (another emit doesn't start until the previous listener returns), we need some kind of queue or a way to track state. We could refactor it to keep the queue/state as a separate member of the class, but the wrapper approach is fine as is

packages/studio-base/src/components/NotificationModal.tsx
jtbandes: apparently isOpen is deprecated 🤷

PR #1921, opened 2021-10-01
Internal: remove sass packages and loaders

defunctzombie:
**User-Facing Changes**
None

**Description**

TODO
- [x] font scss file ported and removed
- [x] storybook style file (maybe can remain?)


2metres:
Looks good to me I've added a story for ShareJsonModal so that it… shows some JSON

Issue #1923, opened 2021-10-01
Expose datatype details in extension API


defunctzombie:
User Story
> As an extension author, I want to know the "shape" of the data on each topic. This might include which fields are in a message on that topic, are the fields arrays? primitive types? This information lets me build a panel that has topic autocomplete functionality similar to the builtin message path input for panels like Raw Message.

defunctzombie:
For additional context on what I see as the path forward to being comfortable exposing the "schemas" to the api in a way we can support.

https://github.com/foxglove/studio/pull/3091#issuecomment-1090598391

banisadr:
I think this is no longer relevant unless I'm misunderstanding the description?

defunctzombie:
I'm gonna close this out until we have more use cases. The one mentioned here for message path we've exposed to all panels via the _settings_ interface so panels don't need to build that themselves.

PR #1926, opened 2021-10-04
Feature: add control to map panel to toggle topics

defunctzombie:
**User-Facing Changes**
Users can click the layers control and toggle topics on/off on the map display.

**Description**
Add topics to the layer control so a user can toggle individual topics on/off in the map display.

Fixes: #1923

### DIFF
--- a/packages/studio-base/src/panels/Map/MapPanel.tsx
+++ b/packages/studio-base/src/panels/Map/MapPanel.tsx
@@ -2,8 +2,15 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
-import { Map as LeafMap, TileLayer, Control, LatLngBounds, CircleMarker } from "leaflet";
-import { useCallback, useEffect, useLayoutEffect, useRef, useState } from "react";
+import {
+  Map as LeafMap,
+  TileLayer,
+  Control,
+  LatLngBounds,
+  CircleMarker,
+  FeatureGroup,
+} from "leaflet";
+import { useCallback, useEffect, useLayoutEffect, useMemo, useRef, useState } from "react";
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/panels/Map/MapPanel.tsx
```diff
@@ -49,23 +56,54 @@ function MapPanel(props: MapPanelProps): JSX.Element {
...
         (topic) =>
           topic.datatype === "sensor_msgs/NavSatFix" ||
           topic.datatype === "sensor_msgs/msg/NavSatFix",
       )
       .map((topic) => topic.name);
+  }, [topics]);
 
+  // Subscribe to relevant topics
+  useEffect(() => {
     context.subscribe(eligibleTopics);
```
jtbandes, line 71: I think ideally the subscription would be made only when topics are selected by the user — it's wasteful to subscribe to all NavSatFix topics and just discard the ones that aren't selected later.

defunctzombie: Interesting - I hadn't thought about that since all the topics are subscribed to when the panel is added (since all topics are shown - unlike the 3d panel). I suppose I should track these within the panel settings. Then changing the logic to only subscribe to what is selected becomes more meaningful. I'll give that a think over and see if I layer that into this PR.

jtbandes: Btw, when multiple topics are enabled we should also use a different color for each topic (and provide a legend).

defunctzombie: Yea - I figure to do the color thing maybe another PR - I haven't figured out how to handle that yet cause the leaflet layer control doesn't have color selection. My hope was to at least provide folks with something given what is available in the control.

PR #1929, opened 2021-10-04
internal: change build target to ESNext

jtbandes:
**User-Facing Changes**
None

**Description**
Enables the use of `#` private identifiers natively.

### DIFF
--- a/packages/log/src/index.ts
+++ b/packages/log/src/index.ts
@@ -11,34 +11,34 @@ class Logger {
   // default logger has an empty name
   static default = new Logger("");
 
-  private _name: string;
-  private _enabled = true;
+  #name: string;
+  #enabled = true;
 
   // all new loggers are created from the default logger
   private constructor(name: string) {
-    this._name = name;
-    this._updateHandlers();
+    this.#name = name;
+    this.#updateHandlers();
 
     channels.set(name, this);
   }
 
   // fully qualified name for the logger
   name() {
-    return this._name;
+    return this.#name;
   }
 
   isEnabled() {
-    return this._enabled;
+    return this.#enabled;
   }
 
   enable() {
-    this._enabled = true;
-    this._updateHandlers();
+    this.#enabled = true;
+    this.#updateHandlers();
   }
 
   disable() {
-    this._enabled = false;
-    this._updateHandlers();
+    this.#enabled = false;
+    this.#updateHandlers();
   }
 
   debug(..._args: unknown[]) {}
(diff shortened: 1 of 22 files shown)


defunctzombie:
Do we want esnext or es2021?

packages/log/src/index.ts
```diff
@@ -11,34 +11,34 @@ class Logger {
   // default logger has an empty name
   static default = new Logger("");
 
-  private _name: string;
-  private _enabled = true;
+  #name: string;
```
defunctzombie, line 14: Fun performance side-effect. This code would previously be compiled out of the final build - leaving on the initializations in the constructor which are fast (https://bugs.chromium.org/p/v8/issues/detail?id=12242&hotlist_id=12935&sort=%20rank%20-ID) but when you use `#` it won't be compiled out which impacts performance. For things like this we probably don't care - and one day the chrome issue will be fixed - but it is worth knowing.

jtbandes:
> Do we want esnext or es2021?

TypeScript still uses the WeakMap when the target is set to es2021.

packages/log/src/index.ts
jtbandes: The removal seems like a TypeScript bug. The declaration has an observable effect: `"x" in new class {x;}` -> `true`. Do you know why TS made the apparently intentional choice to remove non-# property declarations?

amacneil:
Did you check with @jhurliman on this? I recall some performance issues we ran into.

How does this affect our web build?

packages/log/src/index.ts
defunctzombie: I don't know why typescript removes the non-# property declarations. I agree its weird there is a difference.

defunctzombie:
> I recall some performance issues we ran into.

Thats the _WeakMap_ stuff referenced above.

> How does this affect our web build?

Our web build will need to use a browser that supports `#` (private members spec). Since lazy message already used that in its generated class code that was already true and no change for our required browser support.

packages/log/src/index.ts
jtbandes: There is apparently a tsconfig setting `useDefineForClassFields` which causes TS to keep `x;` in esnext, or replace it with an Object.defineProperty call in earlier targets.

The [docs say](https://www.typescriptlang.org/tsconfig#useDefineForClassFields) this flag is true by default "for ES2022 and above, including ESNext", but when I try it on typescriptlang.org/play, it looks like I have to turn it on manually in order for `x;` to be kept.

amacneil:
>> I recall some performance issues we ran into.
> 
> Thats the WeakMap stuff referenced above.

John definitely already considered the option of using esnext target and not compiling them though. So just wondering if we can dig up the reason we chose not to do this in the first place.

jtbandes:
See https://github.com/foxglove/studio/pull/430

jtbandes:
Seems like with the current performance implications of field initializers, which are optional for non-private fields but required for private fields, there is no strong reason to push these changes through.

PR #1932, opened 2021-10-04
🔪 `parseMessageDefinitionsCache`

jtbandes:
**User-Facing Changes**
None

**Description**
We don't have any recent evidence that this cache provides much value for us, and it is implemented with direct localStorage access (well, via the Storage class) which is contrary to our newer design patterns.

Also snips out a couple of now-unused concepts: `messageDefinitionMd5SumByTopic` and `bustStorageFn`.


jtbandes:
@defunctzombie perhaps the Storage class itself could be removed with some small changes to PlayerManager?

defunctzombie:
SUCH EXCITE!

Issue #1551, opened 2021-07-23
Webfonts


2metres:
Given there seems to be an issue using system fonts on Windows. I figured it was best that we select a Web fonts to use for both our sans-serif and monospace fonts.

2metres:
### Inter & Roboto Mono
<img width="1312" alt="Inter V + Roboto Mono" src="https://user-images.githubusercontent.com/924528/126754382-d364d477-6f12-451b-8718-466c109b1a27.png">

2metres:
### PT Sans/Mono
<img width="1312" alt="PT Sans:Mono" src="https://user-images.githubusercontent.com/924528/126754412-31ada161-837a-44c2-83bb-d9915a36f9cc.png">

2metres:
### PT Sans/Mono
<img width="1312" alt="PT Sans:Mono" src="https://user-images.githubusercontent.com/924528/126754454-3aacb17e-2407-4b2a-bf26-e445835fac16.png">

2metres:
Roboto Sans/Mono
<img width="1312" alt="Roboto Sans:Mono" src="https://user-images.githubusercontent.com/924528/126754494-211eaa75-e4ee-406a-b2bd-4c710162844d.png">

2metres:
### Source Sans/Code Pro
<img width="1312" alt="Source Sans:Mono" src="https://user-images.githubusercontent.com/924528/126754701-aa10426b-9207-4df1-9897-78770e2c9186.png">

2metres:
### Space Grotesk/Mono
<img width="1312" alt="Space Grotesk:Mono" src="https://user-images.githubusercontent.com/924528/126754762-08ce39de-0f92-4774-a186-25800f0c3dba.png">

2metres:
Plex Sans/Mono
<img width="1312" alt="IBM Plex Sans:Mono" src="https://user-images.githubusercontent.com/924528/126754802-e7fd35aa-f58e-497e-b5b2-10d9e04f03e8.png">

amacneil:
Source Sans is nice. Good spacing and easy to read.

I prefer Roboto Mono over Code Pro. Code Pro looks to vertically squashed.

Something is wrong with your screenshots because the monospace in `Roboto Sans/Mono` is different from the monospace in `Inter & Roboto Mono`.

jtbandes:
The Source Sans screenshot looks like system-ui to me...

2metres:
[...]

PR #1933, opened 2021-10-05
Change font styles

2metres:
**User-Facing Changes**
sans-serif font now uses: 'Inter'
monospace font now uses: 'IBM Plex Mono'

**Description**

#1551


packages/studio-base/src/util/sharedStyleConstants.ts
```diff
@@ -151,6 +151,6 @@ export const rounded = {
 };
 
 export const fonts = {
-  SANS_SERIF: `-apple-system, BlinkMacSystemFont, "Segoe UI", "Ubuntu", "Roboto", sans-serif`,
-  MONOSPACE: `"Roboto Mono"`,
+  SANS_SERIF: `"Inter", -apple-system, BlinkMacSystemFont, "Segoe UI", "Ubuntu", sans-serif`,
```
amacneil, line 154: probably just kill the fallback fonts? if it looks terrible then it will be more obvious there a bug?

amacneil:
:tada:

PR #1939, opened 2021-10-05
Raw Messages: avoid expensive Object.keys call for large arrays

jtbandes:
**User-Facing Changes**
Improved performance of the Raw Messages panel when viewing large messages.

**Description**
The existing postprocessValue implementation would call Object.keys multiple times. For large objects, including TypedArrays, Object.keys can be expensive and return a large array:

```
> Object.keys(new Uint8Array(10))
[
  '0', '1', '2', '3',
  '4', '5', '6', '7',
  '8', '9'   <-- imagine a Uint8Array(1000) 😅 
]
```

This PR attempts to preserve the logic of checking the existence of diff-related keys, but does this before maybeDeepParse and without calling Object.keys.

### DIFF
--- a/packages/studio-base/src/panels/RawMessages/index.tsx
+++ b/packages/studio-base/src/panels/RawMessages/index.tsx
@@ -444,6 +444,9 @@ function RawMessages(props: Props) {
                 );
               }}
               postprocessValue={(rawVal: unknown) => {
+                if (rawVal == undefined) {
+                  return rawVal;
+                }
                 const idValue = (rawVal as Record<string, unknown>)[diffLabels.ID.labelText];
                 const addedValue = (rawVal as Record<string, unknown>)[diffLabels.ADDED.labelText];
                 const changedValue = (rawVal as Record<string, unknown>)[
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/panels/RawMessages/index.tsx
```diff
@@ -445,18 +444,27 @@ function RawMessages(props: Props) {
                 );
               }}
               postprocessValue={(rawVal: unknown) => {
-                const val = maybeDeepParse(rawVal);
+                if (rawVal == undefined) {
```
defunctzombie, line 447: Worth a comment around what you believe the logic should do any why deepParse is avoided until the end. It might be nice to hoist this function out of the inline jsx as well - starts to get rough to read.

packages/studio-base/src/panels/RawMessages/index.tsx
```diff
@@ -445,18 +444,27 @@ function RawMessages(props: Props) {
...
+                  return rawVal;
+                }
+                const idValue = (rawVal as Record<string, unknown>)[diffLabels.ID.labelText];
```
defunctzombie, line 450: Does this work if the rawValue is not an object? i.e. if you've filtered down to a single value? Or has that run a different code path?

jtbandes: As long as it's not undefined, accessing these keys won't raise an error so it works even without a `typeof rawVal==="object"` check... whether this is good or not is up for debate

Issue #1426, opened 2021-07-08
Convert class components to function components


defunctzombie:
Some components remain using classes. They should be assimilated into the function component borg.

- [x] packages/studio-base/src/components/Autocomplete.tsx
- [x] packages/studio-base/src/components/ButtonBase.tsx
- [ ] packages/studio-base/src/components/Dropdown/index.tsx https://github.com/foxglove/studio/pull/2141
- [ ] packages/studio-base/src/components/KeyListener.tsx https://github.com/foxglove/studio/pull/2142
- [x] packages/studio-base/src/components/Menu/Menu.tsx
- [ ] ~~packages/studio-base/src/components/Menu/SubMenu.tsx~~ We will replace this with Fluent or MUI
- [x] #1568
- [x] packages/studio-base/src/components/NotificationDisplay.tsx
- [x] packages/studio-base/src/components/Select/Option.tsx #1542
- [x] packages/studio-base/src/components/Select/Select.tsx #1542
- [x] packages/studio-base/src/components/Slider.tsx
- [ ] packages/studio-base/src/panels/ThreeDimensionalViz/DrawingTools/MeasuringTool.ts https://github.com/foxglove/studio/pull/2451
- [x] packages/studio-base/src/panels/diagnostics/DiagnosticStatus.tsx
- [x] packages/studio-base/src/panels/diagnostics/DiagnosticSummary.tsx

jtbandes:
Several of these should be removed and replaced with FluentUI:
- #703 
- #705 
- #702

defunctzombie:
Closing this out - The remaining items of interest have open PRs and we will track those individually.

PR #1986, opened 2021-10-18
components/autocomplete: wip transition from class to functional react (for #1426)

matthewmueller:
**User-Facing Changes**

Nothing should change.

**Description**

<s>Not quite working yet, but I wanted to commit early to make sure it's going in the right direction. </s>
Refactored the Autocomplete component from a class-based React component to a hook-based functional React component.

**Todos**

- [x] Code ported over to a functional component
- [x] Get Storybook to work the same across the class and functional variants
- [x] Update `<Autocomplete>` usage across the codebase
- [x] Cleanup variables now that it's working (_onBlur => onBlur)
- [x] Make the linter happy
-  Anything else?

**Notes**

I've left some notes in comments below

**References**

This is part of https://github.com/foxglove/studio/issues/1426

### DIFF
--- a/packages/studio-base/src/components/Autocomplete.tsx
+++ b/packages/studio-base/src/components/Autocomplete.tsx
@@ -193,6 +193,7 @@ export default React.forwardRef(function Autocomplete<T = unknown>(
 ): JSX.Element {
   // References
   const autocomplete = useRef<ReactAutocomplete>(ReactNull);
+  const ignoreBlur = useRef<boolean>(false);
 
   // Context
   const classes = useStyles();
@@ -201,7 +202,6 @@ export default React.forwardRef(function Autocomplete<T = unknown>(
   const [showAllItems, setShowAllItems] = useState<boolean>(false);
   const [stateValue, setValue] = useState<string | undefined>(undefined);
   const [focused, setFocused] = useState<boolean>(false);
-  const [stateIgnoreBlur, setIgnoreBlur] = useState<boolean>(false);
 
   const getItemText = useMemo(
     () => props.getItemText ?? defaultGetText("getItemText"),
@@ -247,11 +247,13 @@ export default React.forwardRef(function Autocomplete<T = unknown>(
   );
 
   const hasError = props.hasError ?? (autocompleteItems.length === 0 && value?.length);
+
   const open = focused && autocompleteItems.length > 0;
-  const ignoreBlur = open ? stateIgnoreBlur : false;
+  if (!open) {
+    ignoreBlur.current = false;
+  }
 
   const selectedItemValue = selectedItem != undefined ? getItemValue(selectedItem) : undefined;
-
   const setSelectionRange = useCallback((selectionStart: number, selectionEnd: number): void => {
     if (autocomplete.current?.refs.input) {
       (autocomplete.current.refs.input as HTMLInputElement).setSelectionRange(
@@ -272,7 +274,7 @@ export default React.forwardRef(function Autocomplete<T = unknown>(
     if (autocomplete.current?.refs.input) {
       (autocomplete.current.refs.input as HTMLInputElement).blur();
     }
-    setIgnoreBlur(false);
+    ignoreBlur.current = false;
     setFocused(false);
     if (onBlurCallback) {
       onBlurCallback();
@@ -298,7 +300,7 @@ export default React.forwardRef(function Autocomplete<T = unknown>(
   );
 
   // Make sure the input field gets focused again after selecting, in case we're doing multiple
-  // autocompletes. We pass an instance of an `AutocompleteController` to `onSelect` in case
+  // autocompletes. We pass an instance of an `IAutocomplete` to `onSelect` in case
   // the user of this component wants to call `blur()`.
   const onSelect = useCallback(
     (textValue: string, { item }: FzfResultItem<T>): void => {
@@ -325,7 +327,7 @@ export default React.forwardRef(function Autocomplete<T = unknown>(
   }, [clearOnFocus]);
 
   const onBlur = useCallback((): void => {
-    if (ignoreBlur) {
+    if (ignoreBlur.current) {
       return;
     }
     if (
@@ -488,7 +490,10 @@ export default React.forwardRef(function Autocomplete<T = unknown>(
           >
             {/* Have to wrap onMouseEnter and onMouseLeave in a separate <div>, as react-autocomplete
              * would override them on the root <div>. */}
-            <div onMouseEnter={() => setIgnoreBlur(true)} onMouseLeave={() => setIgnoreBlur(false)}>
+            <div
+              onMouseEnter={() => (ignoreBlur.current = true)}
+              onMouseLeave={() => (ignoreBlur.current = false)}
+            >
               {menuItemsToShow}
             </div>
           </div>
--- a/packages/studio-base/src/components/Autocomplete.tsx
+++ b/packages/studio-base/src/components/Autocomplete.tsx
@@ -301,12 +301,12 @@ export default React.forwardRef(function Autocomplete<T = unknown>(
   // autocompletes. We pass an instance of an `AutocompleteController` to `onSelect` in case
   // the user of this component wants to call `blur()`.
   const onSelect = useCallback(
-    (textValue: string, item: FzfResultItem<T>): void => {
+    (textValue: string, { item }: FzfResultItem<T>): void => {
       if (autocomplete.current?.refs.input) {
         (autocomplete.current.refs.input as HTMLInputElement).focus();
         setFocused(true);
         setValue(undefined);
-        onSelectCallback(textValue, item.item, { setSelectionRange, focus, blur });
+        onSelectCallback(textValue, item, { setSelectionRange, focus, blur });
       }
     },
     [onSelectCallback, blur, focus, setSelectionRange],
--- a/packages/studio-base/src/components/Autocomplete.tsx
+++ b/packages/studio-base/src/components/Autocomplete.tsx
@@ -15,7 +15,14 @@ import { makeStyles } from "@fluentui/react";
 import cx from "classnames";
 import { Fzf, FzfResultItem } from "fzf";
 import { maxBy } from "lodash";
-import React, { CSSProperties, PureComponent, RefObject, useCallback } from "react";
+import React, {
+  CSSProperties,
+  useRef,
+  useState,
+  useCallback,
+  useMemo,
+  useImperativeHandle,
+} from "react";
 import ReactAutocomplete from "react-autocomplete";
 import { createPortal } from "react-dom";
 import textMetrics from "text-metrics";
@@ -97,10 +104,10 @@ const useStyles = makeStyles((theme) => ({
 // strings like in the Plot panel.
 //
 // The multiple autocompletes doesn't work super well with react-autocomplete, so we have to
-// reimplement some of its behaviour to make things work properly, such as the `_ignoreBlur`
+// reimplement some of its behaviour to make things work properly, such as the `ignoreBlur`
 // stuff. Mostly, though, we can lean on react-autocomplete to do the heavy lifting.
 //
-// For future reference, the reason `<ReactAutocomplete>` (and we) has to do `_ignoreBlur`, is that
+// For future reference, the reason `<ReactAutocomplete>` (and we) has to do `ignoreBlur`, is that
 // when you select an item from the autocomplete menu by clicking, it first triggers a `blur` event
 // on the `<input>`, before triggering a `click` event. If we wouldn't ignore that `blur` event,
 // we'd hide the menu before the `click` event even has a chance of getting fired. So the `blur`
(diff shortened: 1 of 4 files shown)


packages/studio-base/src/panels/diagnostics/DiagnosticStatusPanel.tsx
```diff
@@ -55,11 +55,7 @@ function DiagnosticStatusPanel(props: Props) {
   } = config;
 
   const onSelect = useCallback(
-    (
-      _value: string,
-      entry: DiagnosticAutocompleteEntry,
-      autocomplete: Autocomplete<DiagnosticAutocompleteEntry>,
-    ) => {
+    (_value: string, entry: DiagnosticAutocompleteEntry, autocomplete: AutocompleteAPI) => {
```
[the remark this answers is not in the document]
matthewmueller: I started adjusting this, but it ended up being more code than keeping it as is. I'd suggest revisiting if onSelect needs more params.

packages/studio-base/src/panels/ThreeDimensionalViz/FollowTFControl.tsx
```diff
@@ -184,18 +186,26 @@ const FollowTFControl = memo<Props>((props: Props) => {
...
+    ],
   );
 
   const openFrameList = useCallback(() => {
     setForceShowFrameList(true);
-    if (autocomplete.current) {
-      autocomplete.current.focus();
+    if (autocomplete) {
+      autocomplete.focus();
     }
```
jtbandes, line 209: nit for future (not important enough to block this merging): can use TS optional chaining `autocomplete?.focus()`

packages/studio-base/src/panels/ThreeDimensionalViz/FollowTFControl.tsx
```diff
@@ -147,7 +149,7 @@ const FollowTFControl = memo<Props>((props: Props) => {
   const nodesWithoutDefaultFollowTfFrame = allNodes?.length;
   const newFollowTfFrame = allNodes?.[0]?.tf?.id;
 
-  const autocomplete = createRef<Autocomplete<TfTreeNode>>();
```
jtbandes, line 150: Was this change required? I think it's fine, just wondering why the change was made since I would expect it not to be necessary.

packages/studio-base/src/components/Autocomplete.tsx
```diff
@@ -181,308 +182,298 @@ const HighlightChars = (props: { str: string; indices: Set<number> }) => {
...
+  const [stateValue, setValue] = useState<string | undefined>(undefined);
+  const [focused, setFocused] = useState<boolean>(false);
+  const [stateIgnoreBlur, setIgnoreBlur] = useState<boolean>(false);
+
+  // Props
+  const {
+    autocompleteKey,
+    autoSize = false,
+    getItemText = defaultGetText("getItemText"),
+    getItemValue = defaultGetText("getItemValue"),
```
jtbandes, line 200: These are now being created on each render where previously they were created once when building the `static defaultProps`. Might be nice to do these with useMemo instead, e.g. `const getItemText = useMemo(() => props.getItemText ?? defaultGetText("getItemText"), [props.getItemText])`

packages/studio-base/src/components/Autocomplete.tsx
```diff
@@ -181,308 +182,298 @@ const HighlightChars = (props: { str: string; indices: Set<number> }) => {
...
+    sortWhenFiltering = true,
+    clearOnFocus = false,
+    minWidth = 100,
+    menuStyle = {},
+    inputStyle = {},
+    onChange: onChangeCallback,
+    onSelect: onSelectCallback,
+    onBlur: onBlurCallback,
+    disableAutoSelect,
+    ref = () => {},
```
jtbandes, line 215: Does this actually work? ref can't normally passed via props, and can only be accessed in functional components by wrapping with `forwardRef`. At least that's my understanding.

packages/studio-base/src/components/Autocomplete.tsx
```diff
@@ -181,308 +182,298 @@ const HighlightChars = (props: { str: string; indices: Set<number> }) => {
...
+  }: AutocompleteProps<T> = props;
+
+  const autocompleteItems: FzfResultItem<T>[] = filterText
+    ? new Fzf(items, {
+        fuzzy: filterText.length > 2 ? "v2" : false,
+        sort: sortWhenFiltering,
+        limit: MAX_ITEMS,
+        selector: getItemText as (_: unknown) => string, // Fzf selector TS type seems to be wrong?
+      }).find(filterText)
+    : items.map((item) => itemToFzfResult(item));
```
jtbandes, line 225: We'll want this wrapped in a useMemo so we don't call .find() unnecessarily if it's expensive. I'm not sure if the initialization of `new Fzf` is expensive either, but could consider memoizing that separately.

packages/studio-base/src/components/Autocomplete.tsx
```diff
@@ -181,308 +182,298 @@ const HighlightChars = (props: { str: string; indices: Set<number> }) => {
...
-  }
+  }, [onBlurCallback]);
 
-  private _onFocus = (): void => {
-    if (this._ignoreFocus) {
-      return;
-    }
-    const { clearOnFocus } = this.props;
+  // Give callers an opportunity to control autocomplete
+  ref({ setSelectionRange, focus, blur });
```
jtbandes, line 263: useImperativeHandle

packages/studio-base/src/components/Autocomplete.tsx
```diff
@@ -181,308 +182,298 @@ const HighlightChars = (props: { str: string; indices: Set<number> }) => {
...
+        onChangeCallback(event, (event.target as HTMLInputElement).value);
+      } else {
+        setValue((event.target as HTMLInputElement).value);
+      }
+    },
+    [onChangeCallback],
+  );
+
+  // Make sure the input field gets focused again after selecting, in case we're doing multiple
+  // autocompletes. We pass in `this` to `onSelect` in case the user of this component wants to call
```
jtbandes, line 277: nit: `this` comment is out of date

packages/studio-base/src/components/Autocomplete.tsx
```diff
@@ -109,21 +116,21 @@ const classes = mergeStyleSets({
 // it doesn't cause any problems.)
 type AutocompleteProps<T = unknown> = {
   items: T[];
-  getItemValue: (arg0: T) => string;
-  getItemText: (arg0: T) => string;
+  getItemValue?: (arg0: T) => string;
+  getItemText?: (arg0: T) => string;
```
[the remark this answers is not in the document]
jtbandes: Anything present in `static defaultProps` is allowed to be non-optional in Props.

packages/studio-base/src/panels/ThreeDimensionalViz/FollowTFControl.tsx
matthewmueller: I changed this because I wasn't able to figure out a way to manually write to a ref done here: https://github.com/foxglove/studio/pull/1986/files/49a7c781291d20572f93c7e76c62a6555c54b001#diff-0b248d633e53e623ebd68e8998751ce04b38d2801d3e2d64b2cd5602b25c39deR257

packages/studio-base/src/components/Autocomplete.tsx
matthewmueller: This does work, but I'm not sure it's the right way to do things.

My understanding is that ref is used exclusively to reference DOM nodes. The ref in this case is just a JS object that contains some methods for controlling autocomplete, allow other components to trigger re-renders. This is why I originally called this useAPI or something other than ref to avoid confusion.

Is there a way to use refs for non-DOM nodes? If so, is there some documentation I could take a look at or do we have any examples in the codebase?

packages/studio-base/src/components/Autocomplete.tsx
matthewmueller: Good catch! Fixed.

packages/studio-base/src/components/Autocomplete.tsx
matthewmueller: See https://github.com/foxglove/studio/pull/1986#discussion_r742531416. I'd like to clear that up before making any changes here.

packages/studio-base/src/components/Autocomplete.tsx
matthewmueller: Good call! Still learning all the various hooks. I appreciate for the reminders. I'll most likely need a few more of these 🙂

Regarding also memoizing `Fzf`. I don't think it makes much sense in this case because `Fzf` depends on `filterText` all the same, so every time `filterText` changes, you'd want to re-initialize `Fzf`. It's a rather bad use of a class actually.

packages/studio-base/src/components/Autocomplete.tsx
matthewmueller: Actually I just started seeing errors in development about passing a ref through. Odd that I didn't see it before. I just adjusted it to `controller` for the time-being. If there's a better way to pass non-DOM nodes as refs, please let me know!

packages/studio-base/src/panels/ThreeDimensionalViz/FollowTFControl.tsx
```diff
@@ -248,7 +254,7 @@ const FollowTFControl = memo<Props>((props: Props) => {
     >
       {showFrameList && (
         <Autocomplete
-          ref={autocomplete}
+          controller={(api) => (autocomplete = api)}
```
matthewmueller, line 257: renamed because ref is special

packages/studio-base/src/components/Autocomplete.tsx
matthewmueller: **Update** I just renamed it from ref to controller.

packages/studio-base/src/components/Autocomplete.tsx
```diff
@@ -181,336 +187,319 @@ const HighlightChars = (props: { str: string; indices: Set<number> }) => {
...
+            sort: sortWhenFiltering,
+            limit: MAX_ITEMS,
+            selector: getItemText as (_: unknown) => string, // Fzf selector TS type seems to be wrong?
+          }).find(filterText)
+        : items.map((item) => itemToFzfResult(item)),
+    [filterText, getItemText, items, sortWhenFiltering],
+  );
+
+  const hasError =
+    props.hasError === true ? props.hasError : autocompleteItems.length === 0 && value?.length;
```
jtbandes, line 250: I think this logic changed unintentionally. Looking at the storybook screenshots, I see red/error text in a lot more cases than before. The original behavior was that the computed value based on autocompleteItems.length would be used only if props.hasError was undefined.

packages/studio-base/src/components/Autocomplete.tsx
matthewmueller: Ah, TIL. Thanks!

packages/studio-base/src/components/Autocomplete.tsx
matthewmueller: Great catch, 

- in main: if `props.hasError` is false, then `hasError` would be false
- in this branch: if `props.hasError` is false, then `hasError` is computed from the autocomplete items.

Where are the storybook screenshots? 

I took a look at [this branch](https://www.chromatic.com/component?buildNumber=4416&historyLengthAtIndex=33&distanceToMoveBack=-22&appId=603ec8bf7908b500231841e2&name=components%2FAutocomplete&specName=with+a+long+truncated+path+%28and+autoSize%29&componentInspectorKey=6183fd02de17a7003afe1f56-1200-interactive-true) vs. [main](https://www.chromatic.com/component?appId=603ec8bf7908b500231841e2&name=components%2FAutocomplete&buildNumber=4446&componentInspectorKey=6185b774a34691003aa092dd-1200-interactive-true&historyLengthAtIndex=5&distanceToMoveBack=-1) but didn't notice a difference.

packages/studio-base/src/components/Autocomplete.tsx
```diff
@@ -246,9 +246,7 @@ export default React.forwardRef(function Autocomplete<T = unknown>(
     [filterText, getItemText, items, sortWhenFiltering],
   );
 
-  const hasError =
-    props.hasError === true ? props.hasError : autocompleteItems.length === 0 && value?.length;
-
+  const hasError = props.hasError ?? (autocompleteItems.length === 0 && value?.length);
```
matthewmueller, line 249: If props.hasError is null or undefined, then compute from autocompleteItems. Since hasError is typed only to undefined | boolean, it should be the same behavior as main.

I was tempted to just use the old school, `typeof props.hasError === 'undefined'`, but took this opportunity to refresh on the new syntax (I like it!)

matthewmueller:
FYI, incorporated https://github.com/foxglove/studio/pull/2078#pullrequestreview-798200186 into this branch.

packages/studio-base/src/components/Autocomplete.tsx
jtbandes: Not sure about the Autocomplete stories in particular but Autocomplete ends up getting used in lots of different stories. Here's an example of a build from this branch showing 88 changed stories; I think I was looking at something like this (but not exactly sure which build it was): https://www.chromatic.com/build?appId=603ec8bf7908b500231841e2&number=4416 But now that you've made fixes, the number of changes has gone down to 1: 
image

I think Chromatic keeps track of intermediate approvals so unfortunately it's not always possible to get a complete inventory of what's changed from main to the current branch head...

packages/studio-base/src/components/Autocomplete.tsx
```diff
@@ -181,336 +187,317 @@ const HighlightChars = (props: { str: string; indices: Set<number> }) => {
...
+            limit: MAX_ITEMS,
+            selector: getItemText as (_: unknown) => string, // Fzf selector TS type seems to be wrong?
+          }).find(filterText)
+        : items.map((item) => itemToFzfResult(item)),
+    [filterText, getItemText, items, sortWhenFiltering],
+  );
+
+  const hasError = props.hasError ?? (autocompleteItems.length === 0 && value?.length);
+  const open = focused && autocompleteItems.length > 0;
+  const ignoreBlur = open ? stateIgnoreBlur : false;
```
jtbandes, line 251: Not sure I totally get what this is for — is this a change from the original implementation?

jtbandes (approved):

packages/studio-base/src/components/Autocomplete.tsx
matthewmueller: Ah, I see that now. That's super nice. Okay so next time I'll check out the Unreviewed changes in Chromatic. Thanks!

packages/studio-base/src/components/Autocomplete.tsx
matthewmueller: Good catch! Yes, but it wasn't intentional :-) 

This translation would trigger re-renders, which I could see causing side-effects when you're considering event order (blur coming before click).

I adjusted the implemented to use references, which shouldn't trigger a re-render. I wasn't able to confirm the problem with the existing code though. It seems like it's most obvious with multiple autocompletes. I took a look through the codebase, but couldn't find an example of this. The comment mentions the Plot which doesn't seem to depend on Autocomplete as far as I can tell.

packages/studio-base/src/components/Autocomplete.tsx
```diff
@@ -181,336 +187,322 @@ const HighlightChars = (props: { str: string; indices: Set<number> }) => {
...
+  // References
+  const autocomplete = useRef<ReactAutocomplete>(ReactNull);
+  const ignoreBlur = useRef<boolean>(false);
 
-class AutocompleteImpl<T = unknown>
-  extends PureComponent<AutocompleteProps<T>, AutocompleteState>
-  implements IAutocomplete
-{
-  private _autocomplete: RefObject<ReactAutocomplete>;
-  private _ignoreFocus: boolean = false;
```
matthewmueller, line 195: Note that I removed `_ignoreFocus` because it seemed like unused logic. It's never set to true, so the conditional in `onFocus` always passes.

packages/studio-base/src/components/Autocomplete.tsx
matthewmueller: Maybe we add a story that demonstrates this problem?

jtbandes: If we can't easily figure out how to reproduce but we are keeping the existing behavior, I think it's probably not worth the effort to add a story for it right now. I think we are eventually due for a larger rewrite of this component because react-autocomplete is no longer maintained...

Issue #1998, opened 2021-10-20
Change default font size to 13px


amacneil:
**Description**
Studio currently uses a huge mix of base font sizes, from 12px for much of the original webviz design, to 14px for much of the new fluent stuff.

We should standardize on 13px as our base font size for majority of app text. Besides being a middle ground, it's also what vscode uses for most of its interface (besides the code editor).

2metres:
This will likely come with the switch over to Material-UI

![image](https://user-images.githubusercontent.com/924528/150865821-fb2bf6b4-0694-4717-a717-8dba4eec3279.png)

defunctzombie:
@amacneil @2metres is this still a thing or has our font size situation improved?

amacneil:
This is sufficiently done, font sizes are much more consistent now.

PR #1999, opened 2021-10-20
Use 13px base font size

amacneil:
**User-Facing Changes**
Font sizes are adjusted throughout the app to standardize on 13px for most text. See https://github.com/foxglove/studio/issues/1998 for rationale.

**Description**
- Various font sizes are configured in `src/theme/index.ts` and used throughout the app.
- `medium` corresponds to `13px` and used in most places
- Weird `xyzPlus` font variants are marked deprecated (some but not all are already marked deprecated in fluentui typedefs)
- Dialed back a few inconsistencies such as the Diagnostic Status panel heading font size

**Todo**
- Investigate: 3D panel buttons/icons got smaller for some reason

Closes https://github.com/foxglove/studio/issues/1998

**Before**
image

**After**
image


defunctzombie:
My general feeling is that maybe we should go smaller with the default font size - 👍 to going with 13 and being consistent to start.

2metres:
This appears to break all icons

2metres:
That’s been off for a while now, likely unrelated to your change

defunctzombie:
We want to switch to MUI and should revisit fonts and themes once we do that.

Issue #1344, opened 2021-06-30
Plot tooltip only allows viewing data from one dataset


lucabonamini:
**Description**
When more than one timeseries is loaded in a Plot panel, tooltip shows data of the first loaded only on hover.

- Version: 0.12.0
- OS: Ubuntu 18.04

**Steps To Reproduce**
![Screenshot from 2021-06-30 15-19-35](https://user-images.githubusercontent.com/24660508/123971034-fa830600-d9b9-11eb-815f-007fd3441898.png)

**Expected Behavior**
Tooltip should have data relatives to all the timeseries loaded.

**Actual Behavior**
Tooltip shows data of the first timeseries loaded only.

jtbandes:
Related: #1113

jetdillo:
I could really use a fix for this issue as I have my graphs configured to aggregate the data.* elements from a topic (like /imu/data.orientation.*) onto a single graph panel to more easily compare them against each other vs. split out into separate panels.

defunctzombie:
https://github.com/foxglove/studio/pull/1835 This allows for hovering near the other dataset points to see their values to fix parts of the poor UX described in this issue.

I think it could make sense to handle scatter and time series differently and display all the intersected x-axis values in one tooltip for time based plots so I will leave this issue open.

Issue #1173, opened 2021-06-07
Make Plot panel tooltip less intrusive


esthersweon:
# Feature request

From user:

"At the moment I use PlotJuggler and rviz together to give me the same functionality. One thing that is a real pain with Studio is the way it shows the current cursor value in a plot; PlotJuggler does this better imho."

<img width="495" alt="Screen Shot 2021-06-07 at 11 49 54 AM" src="https://user-images.githubusercontent.com/6993359/121074651-fda71f80-c788-11eb-83bc-e3c6f9765d93.png">

<img width="500" alt="Screen Shot 2021-06-07 at 11 51 42 AM" src="https://user-images.githubusercontent.com/6993359/121074656-00097980-c789-11eb-9c68-a2f599552db8.png">

As an aside, it's annoying to have the Plot panel's tooltip accidentally show up on hover, and finicky to have it show up **only** on line hover at that. 

# Potential solution(s)

- Use similar tooltip presentation to PlotJuggler – i.e. display x value in one place, and only show y value in tooltip.
- Use panel settings to show a more or less verbose version of the tooltip.
- Follow PlotJuggler's system of requiring manual click to show tooltip values.

PR #2014, opened 2021-10-22
Feature: simplify plot tooltip

defunctzombie:
**User-Facing Changes**
The tooltip for all plot panels shows only the series and series data. The tooltip can show data from multiple series. The tooltip follows the mouse y-axis so the user can move the tooltip out of the way.

The time fields are removed from the tooltip and shown above the playback control hover bar ticks as done in: https://github.com/foxglove/studio/pull/2134

| | Before | After |
| --- | --- | --- |
| Single Series | image | image |
| Multiple Series | image|image |
| State Transitions | image |image
 |

**Description**
The existing tooltip would cover up a lot of the plot. This change makes the tooltip show only the hovered value.
    
Fixes: #1173
Fixes: #1344

### DIFF
--- a/packages/studio-base/src/components/Chart/index.tsx
+++ b/packages/studio-base/src/components/Chart/index.tsx
@@ -488,7 +488,7 @@ function Chart(props: Props): JSX.Element {
       onMouseLeave={onMouseLeave}
       onMouseEnter={onMouseEnter}
       onMouseUp={onMouseUp}
-      style={{ width, height }}
+      style={{ width, height, cursor: "crosshair" }}
     />
   );
 }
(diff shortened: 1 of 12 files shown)


defunctzombie:
Some thoughts as I worked on this:

- We need a way to show the x axis value. For time plots that is the timestamp and for other plots that is the x value.
- Even for non-timeseries plots a user might still want to know _when_ the datapoint occurred (thus why receive time was shown)
- The receive time and header stamp may be different.
- When playback is header stamp order, knowing the header stamp time and how that relates to receive time might matter.

bclinkinbeard:
What about the concerns raised in your last comment? I don't think the X axis info can be removed entirely.

defunctzombie:
@bclinkinbeard Some of the x-axis information that was removed is displayed now on the playback scrubber when hovering over the plot panel. This change was done in https://github.com/foxglove/studio/pull/2134 which has some screenshots of what it looks like when you hover over the plot and what appears in the playback controls area.

bclinkinbeard:
Cool, that seems like a good approach.

Are there plans to make the tooltip customizable at all, even if in a follow up PR?

defunctzombie:
@bclinkinbeard no plans for customizability via panel settings. If that is something that is useful let's open an issue to discover the scope of what customizability should include.

bclinkinbeard:
Sorry, I meant customization by developers via some kind of API.

PR #2021, opened 2021-10-22
Cleanup ErrorBoundary appearance

defunctzombie:
**User-Facing Changes**
Visual changes to the error boundary for panels. The error details are default hidden and can be shown with a toggle.

The user can also remove the panel via a button on the boundary if the panel continues to misbehave.

Screen Shot 2022-01-13 at 12 27 22 PM
Screen Shot 2022-01-13 at 12 17 54 PM

**Description**
- Create a separate PanelErrorBoundary for panels. Allow the user to remove the panel within the boundary.
- Put error details for all boundaries under a toggle.


defunctzombie:
source line text is illegible

packages/studio-base/src/components/ErrorDisplay.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+function sanitizeStack(stack: string) {
+  return stack
+    .replace(/\s+\(.+\)$/gm, " (some location)")
+    .replace(/\s+https?:\/\/.+$/gm, " some location");
+}
+
+function ErrorStacktrace({ stack }: { stack: string }) {
+  const styles = useStyles();
+  const lines = sanitizeStack(stack)
```
jtbandes, line 48: It looks like this `sanitizeStack` function is now running all the time, where previously it was conditional. The purpose of this function was to make storybook screenshots deterministic (by hiding file hashes). But I don't think we should be hiding source locations from actual users, especially not replacing it with the string `"some location"`

packages/studio-base/src/components/ErrorDisplay.tsx
```diff
@@ -0,0 +1,144 @@
...
+        )}
+      </div>
+    );
+  }, [error, errorInfo, showErrorDetails, styles]);
+
+  return (
+    <div className={styles.root}>
+      <div className={styles.alertContainer}>
+        <Typography variant="h4">
+          {props.title ?? "The app encountered an unexpected error"}
```
jtbandes, line 115:
```suggestion
          {props.title ?? "This panel encountered an error"}
```

packages/studio-base/src/components/ErrorDisplay.tsx
```diff
@@ -0,0 +1,144 @@
...
+        </Typography>
+        <AlertTitle></AlertTitle>
+        <Typography variant="body1">
+          <Stack spacing={2}>
+            <div>
+              <p>Something went wrong in the panel. Click Dismiss to continue using the panel.</p>
+              <p>If the issue persists try resetting the panel or removing the panel entirely.</p>
```
jtbandes, line 122:
```suggestion
              <p>An unexpected error occurred in this panel. Click Dismiss to continue using the panel. If the issue persists, try resetting the panel.</p>
```

Telling them to remove it doesn't seem that helpful, they can decide whether or not they want to keep the panel.

I find it a little weird to say "Click Dismiss to continue ..." directing them to a button when that text could've just been clickable itself. Maybe we make this text a link that also does the dismiss action?

packages/studio-base/src/components/ErrorDisplay.tsx
```diff
@@ -0,0 +1,144 @@
...
+            </div>
+            <div>
+              {error?.message} -{" "}
```
jtbandes, line 125: We should make this a different font that is distinct from the explanatory text above.

Would rather put the link in a separate div so they are on two lines, the ` - ` thing is weird.

packages/studio-base/src/components/PanelLayout.stories.tsx
```diff
@@ -120,16 +119,14 @@ const PanelToolbarShown = createGlobalStyle`
 export const PanelWithError = (): JSX.Element => {
   return (
     <DndProvider backend={HTML5Backend}>
-      <HideErrorSourceLocations.Provider value={true}>
```
jtbandes, line 123: this hide source locations was meant to happen only in storybook

packages/studio-base/src/components/ErrorDisplay.tsx
defunctzombie: I'll make the text clickable that also does the dismiss action. I will remove the hint to remove the panel and keep the Remove panel button at the bottom.

packages/studio-base/src/components/ErrorDisplay.tsx
defunctzombie: 👍

packages/studio-base/src/components/PanelLayout.stories.tsx
defunctzombie: Gotcha - ok will make sure source locations are shown always. For storybook we actually write a manual stack to work around this.

jtbandes: restored in https://github.com/foxglove/studio/pull/2962 (not using context, just props)

PR #2042, opened 2021-10-28
PX4 ULog support

jhurliman:
**User-facing changes**

[PX4 ULogs](https://docs.px4.io/master/en/dev_log/ulog_file_format.html) can be displayed in Studio.

Screen Shot 2021-11-01 at 4 33 24 PM

**Description**

Using the [@foxglove/ulog](https://github.com/foxglove/ulog) library, PX4 ULog (`.ulg`) files can be loaded as a RandomAccessPlayer style data source. Two tweaks were made outside of adding the new player: <Timestamp>, used in the Data Source Info panel will heuristically toggle between rendering absolute and relative timestamps, and RandomAccessPlayer will now advertise the getParameters capability and allow data sources to initialize with a map of parameters.

This does not include any support for mapping well-known PX4 message types to well-known Studio message types, except for mapping Log meta-messages to the `rosgraph_msgs/Log` datatype.

**TODO**

- [ ] https://github.com/foxglove/studio/issues/2062 preloaded range (blocks) so we can plot using the data in the whole file. I think this makes it more compelling to users of ulog since they will be able to plot what their drone did after opening the file rather than playing the whole file to see the plot.
- [x] https://github.com/foxglove/studio/pull/1513 should merge first
- [x] rename `/rosout` to `/log` so that users of ulog don't need to learn "ros"-ism
- [x] update rosout panel to detect log topics by datatype rather than hard-coded (ok to have the data provider _fake_ the rosgraph_msgs for now)
- [ ] https://github.com/foxglove/studio/pull/2060 rename rosout panel to log? (or make a separate log panel) so ulog users don't need to learn "ros"-isms
- [x] performance improvements - opening the 88mb test file from the foxglove/ulog repo takes 1-2 minutes before studio allows playback. Feels rather high for 88mb file but maybe not?
- [x] Consider removing leading `/` from topic names if that isn't a ulog concept? tho maybe ok to leave it as it might break lots of things around the app in strange ways


defunctzombie:
Been playing with this. Few thoughts on things we should do before merging:

- preloaded range (blocks) so we can plot using the data in the whole file. I think this makes it more compelling to users of ulog since they will be able to plot what their drone did after opening the file rather than playing the whole file to see the plot.
- https://github.com/foxglove/studio/pull/1513 should merge first
- rename `/rosout` to `/log` so that users of ulog don't need to learn "ros"-ism
- update rosout panel to detect log topics by datatype rather than hard-coded (ok to have the data provider _fake_ the rosgraph_msgs for now)
- rename rosout panel to log? (or make a separate log panel) so ulog users don't need to learn "ros"-isms
- performance improvements - opening the 88mb test file from the foxglove/ulog repo takes 1-2 minutes before studio allows playback. Feels rather high for 88mb file but maybe not?
- Consider removing leading `/` from topic names if that isn't a ulog concept? tho maybe ok to leave it as it might break lots of things around the app in strange ways

I think we can get away with just the above for merging this PR and MVP ulog support. The alternative datatype work and map panel support can be done in a follow-up and would enhance the capabilities of the ulog reader but doesn't need to block landing initial support.

jtbandes:
> Consider removing leading / from topic names if that isn't a ulog concept? tho maybe ok to leave it as it might break lots of things around the app in strange ways

+1, if this breaks lots of things we need to learn that so we can fix it.

packages/studio-base/src/util/formatByteSize.ts
```diff
@@ -0,0 +1,14 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+export default function formatByteSize(size: number): string {
```
jtbandes, line 5: If this needs to be used across multiple packages let's move it to the `den` rather than just copy/pasting.

esthersweon:
@jhurliman - Reminder to add `docs` label for PRs that require doc updates. Thank you!

PR #2048, opened 2021-10-29
Attempt to cancel in-progress layout sync when remounting

jtbandes:
**User-Facing Changes**
None

**Description**
Thread AbortSignal through the layout sync logic to improve ability to interrupt an ongoing sync.

The root cause of certain sync errors is that the tree is remounting when the app initially loads, but this seemed like something we could do to be more defensive against problems.

See also: #2051

### DIFF
--- a/packages/studio-base/src/providers/LayoutManagerProvider.tsx
+++ b/packages/studio-base/src/providers/LayoutManagerProvider.tsx
@@ -51,12 +51,12 @@ export default function LayoutManagerProvider({
     if (!enableSyncing) {
       return;
     }
-    let stopped = false;
+    const controller = new AbortController();
     void (async () => {
       let failures = 0;
-      while (!stopped) {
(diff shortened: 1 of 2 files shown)


defunctzombie:
Is an alternative to mark when unmounted and avoid applying any async responses? Having the abort is nice so maybe both should be done? If the provider has unmounted then it shouldn't be applying any updates that arrive after.

jtbandes:
Not sure I understand how “mark when unmounted and avoid applying any async responses” is different from the AbortSignal? signal.aborted is how we mark when unmounted.

defunctzombie:
> Not sure I understand how “mark when unmounted and avoid applying any async responses” is different from the AbortSignal? signal.aborted is how we mark when unmounted.

I was thinking the Provider handled applying the results of fetching to the manager. But I see that the manager does this internally. 👍

PR #2060, opened 2021-11-01
Rename the "RosOut" panel to "Log" panel

jhurliman:
**User-facing changes**

The RosOut panel has been renamed "Log panel". The first topic with a datatype of `rosgraph_msgs/Log` or `rcl_interfaces/msg/Log` will be subscribed to instead of the hardcoded topic `/rosout`. Except for the supported datatype(s), ROS-specific language has been removed.


jtbandes:
Needs migration logic to avoid breaking old layouts.

defunctzombie:
Let's put the panel type back to `Rosout` and keep the user facing name `Log` along with the code renames. That way existing layouts will continue to work.

Add a comment for the panel type indicating it exists for legacy compatability.

packages/studio-base/src/panels/Log/LogMessage.stories.tsx
```diff
@@ -0,0 +1,52 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import LogMessage from "./LogMessage";
```
jtbandes, line 5: I deleted this file upstream

packages/studio-base/src/panels/Log/index.tsx
```diff
@@ -49,6 +52,17 @@ const RosoutPanel = React.memo(({ config, saveConfig }: Props) => {
...
 
+  const defaultTopicToRender = useMemo(
+    () =>
+      topics.find((topic) => topic.datatype === ROS1_LOG || topic.datatype === ROS2_LOG)?.name ??
+      "/rosout",
+    [topics],
+  );
+
+  if (config.topicToRender == undefined) {
+    config.topicToRender = defaultTopicToRender;
```
jtbandes, line 63: please don't mutate the config object (an input to this panel); this can make it hard to track where changes are happening and lead to surprising behavior

Issue #2061, opened 2021-11-01
Bag drag/drop opens file picker


jtbandes:
**Description**
Dragging and dropping a bag opens the file picker instead of loading the bag.
Probably a regression from #1513

- Version: aaacaa8
- OS: macOS

**Steps To Reproduce**
Open https://studio.foxglove.dev and drag a .bag file onto the Studio window.

**Expected Behavior**
Bag starts loading and data is displayed

**Actual Behavior**
File picker is opened

PR #2063, opened 2021-11-02
fix: drag & drop initializes a source with file arg

defunctzombie:
**User-Facing Changes**
Users can drag&drop files for any data source which supports file loading.

**Description**
The player manager failed to properly detect the files argument provided when selecting a data source via drag & drop. This was a regression introduced by https://github.com/foxglove/studio/pull/1513. This fixes the regression and also expands drag & drop support to any data source that supports files.

Removes support for "shift" drag&drop to append files. Support for this was recently removed for the ros1 bag data source and not supported for ros2 bag or ulog.
    
Fixes: #2061

### DIFF
--- a/packages/studio-base/src/Workspace.tsx
+++ b/packages/studio-base/src/Workspace.tsx
@@ -11,6 +11,7 @@
 //   You may not use this file except in compliance with the License.
 
 import { Link, makeStyles, Stack, Text, useTheme } from "@fluentui/react";
+import { extname } from "path";
 import {
   useState,
   useEffect,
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/Workspace.tsx
```diff
@@ -305,24 +310,31 @@ export default function Workspace(props: WorkspaceProps): JSX.Element {
       }
 
       if (otherFiles.length > 0) {
-        if (shiftPressed) {
-          previousFiles.current = previousFiles.current.concat(otherFiles);
```
jtbandes, line 309: I assume this removal is intentional? Have we decided we're ready to remove multi-file support?

jtbandes (approved):

packages/studio-base/src/Workspace.tsx
defunctzombie: Intentional. It's time for it to get the 🔪 .

PR #2065, opened 2021-11-02
Switch rosbag2 data source to load a file instead of a folder

jhurliman:


defunctzombie:
I asked about this in our community #lounge channel and the two responses I got back both indicated they work with folders rather than individual files. I think for now we keep the status quo and open folders until we get some data points the other way.

PR #2071, opened 2021-11-03
Use plot defaultConfig when destructuring config values

defunctzombie:
The default config is only applied when a panel is added to a layout. When we add new config options to the default config, these are not applied to existing panels. Any default options with "truthy" values are thus not actually set to those values for the panel.

This change moves the default value assignment into the panel logic so it is the responsibility of the panel to provide defaults for its options.

**User-Facing Changes**

**Description**

### DIFF
--- a/packages/studio-base/src/panels/Plot/index.tsx
+++ b/packages/studio-base/src/panels/Plot/index.tsx
@@ -211,15 +211,16 @@ function selectEndTime(ctx: MessagePipelineContext) {
 
 function Plot(props: Props) {
   const { saveConfig, config } = props;
+
   const {
-    title,
+    title = defaultConfig.title,
     followingViewWidth,
-    paths: yAxisPaths,
-    minYValue,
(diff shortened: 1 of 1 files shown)


defunctzombie:
@esthersweon @jtbandes I think my statement about the default config being applied only once is correct (will double check that tho before we make changes).

I recently made a change for Publish panel to make all the config items optional and set them in the panel itself (for similar reasons - I was seeing sentry errors that only made sense if the config the panel got was incomplete). I think having the panel logic set its default values is an approach we should take for our panels. Thoughts?

jtbandes:
Default config is automatically merged with panel config here: https://github.com/foxglove/studio/blob/10b8aa573c3cc70291c3a453cc57c62c228fb51d/packages/studio-base/src/components/Panel.tsx#L367-L370

defunctzombie:
Indeed. So this change shouldn't be necessary then.

PR #2074, opened 2021-11-03
Support for system color scheme in desktop build

jtbandes:
**User-Facing Changes**
Part of #2054 – desktop build now supports the "system" color setting.

**Description**
- The desktop app reads the colorScheme setting from settings.json at startup.
- The `desktopBridge` has a new function that the renderer can invoke when the setting is changed, to tell the desktop app to update from disk.


defunctzombie:
👍 Seems sensible.

PR #2076, opened 2021-11-03
Remove global body css

amacneil:
**User-Facing Changes**
WIP

**Description**


defunctzombie:
The body styles are used by a few components that append popup elements to the body. Our autocomplete component is an example. There might be a few more non-fluent components around that are similar.

amacneil:
Yes, looks like we need to update a few popups / dropdowns to use fluent:

image

Closing for now.

PR #2077, opened 2021-11-04
Remove RenameDataProvider and CombinedDataProvider

defunctzombie:
**User-Facing Changes**
None. We've previously disabled the dual rosbag1 loading for base/feature bags.

**Description**
We recently removed support for loading multiple ros1 bag files as the feature was not documented and not well supported across the app. Additionally no other data source behaved in the same way as ros1 bags when loading multiple files.
    
This change removes these two data providers.


packages/studio-base/src/panels/ThreeDimensionalViz/TopicTree/useTopicTree.ts
```diff
@@ -396,25 +356,18 @@ export default function useTopicTree({
   const derivedCustomSettingsByKey = useMemo((): DerivedCustomSettingsByKey => {
     const result: DerivedCustomSettingsByKey = {};
     for (const [topicKeyOrNamespaceKey, settings] of Object.entries(settingsByKey)) {
-      const isFeatureTopicOrNamespace = topicKeyOrNamespaceKey.includes(SECOND_SOURCE_PREFIX);
-      const columnIndex = isFeatureTopicOrNamespace ? 1 : 0;
+      const columnIndex = 0;
```
jtbandes, line 359: lol

packages/studio-base/src/panels/ThreeDimensionalViz/TopicTree/useTopicTree.ts
```diff
@@ -548,19 +489,15 @@ export default function useTopicTree({
...
+          ),
         ),
       );
 
       let newModififiedNamespaceTopics = [...modifiedNamespaceTopics];
       topicNames.forEach((topicName) => {
-        const prefixedTopicName = isFeatureColumn
-          ? `${SECOND_SOURCE_PREFIX}${topicName}`
-          : topicName;
+        const prefixedTopicName = topicName;
```
jtbandes, line 500: remove?

jtbandes (commented):
👏

defunctzombie:
> I've left the some of the "column" code in place to tackle that in a follow-up to keep this already growing PR contained. I also want to make sure that removing that column logic doesn't impact the saved layouts.

@jtbandes think its worth tackling in this PR? - I wasn't sure if it was better reviewed as a follow-up. If we think its not related to saved layouts then I can go through a pass and remove more of the column logic.

jtbandes:
I'm sure some of it is related to saved layouts but some is probably state-only. Separate PR is fine too.

defunctzombie:
Here's my plan:

- Remove the two providers in this PR so they are no longer around to maintain
- Make a PR to remove second source handling from RawMessages panel
- Make a PR to remove second source handling from 3d panel. That PR will also include cleaning up the column logic.

I think the work @jhurliman is doing with 3d panel won't conflict with the changes I make.

PR #2079, opened 2021-11-04
Remove "/index" from story names

jtbandes:
**User-Facing Changes**
None, internal only

**Description**
I'm not aware of any particular reason to keep `/index` in the name of any stories. It sometimes makes Chromatic hide the actual component name and display only "index".


defunctzombie:
👍

PR #2090, opened 2021-11-05
Rename "ROS" time display method to "SEC" (Seconds), default new layouts to "TOD" (Time of Day)

jhurliman:
**User-facing changes**

The "ROS (ROS time)" text in the playback bar dropdown has been changed to "SEC (Seconds)" to provide additional clarity, especially for non-ROS users. New layouts will default to showing "TOD (Time of Day)" for improved readability in the default experience.

### DIFF
--- a/packages/studio-base/src/components/PlaybackControls/PlaybackTimeDisplayMethod.tsx
+++ b/packages/studio-base/src/components/PlaybackControls/PlaybackTimeDisplayMethod.tsx
@@ -26,6 +26,7 @@ import {
   useCurrentLayoutActions,
   useCurrentLayoutSelector,
 } from "@foxglove/studio-base/context/CurrentLayoutContext";
+import { TimeDisplayMethod } from "@foxglove/studio-base/types/panels";
 import {
   formatDate,
   formatTime,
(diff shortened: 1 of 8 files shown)


packages/studio-base/src/components/PlaybackControls/PlaybackTimeDisplayMethod.tsx
```diff
@@ -222,10 +224,10 @@ const PlaybackTimeDisplayMethod = ({
             },
             {
               canCheck: true,
-              key: "ROS",
-              text: "ROS time",
-              isChecked: timeDisplayMethod === "ROS",
-              onClick: () => setTimeDisplayMethod("ROS"),
+              key: "SEC",
+              text: "Seconds",
```
jtbandes, line 228: Should we call this "epoch"? or unix/posix?

jhurliman: It is not necessarily an epoch timestamp. Some data sources will use time relative to the start of recording, others will use time relative to the epoch. The only constant across different data sources is that it will be displayed as a floating-point number of seconds.

amacneil:
I think we could eliminate the actual word from the playback bar, and only show the dropdown chevron icon next to the timestamp. It's obvious whether you are looking at seconds or human time so we don't need to repeat that fact beyond providing access to the dropdown.

Within the dropdown, I'd opt for more human readable words over short abbreviations.

packages/studio-base/src/components/PlaybackControls/PlaybackTimeDisplayMethod.tsx
jtbandes: In that case maybe "time of day" is also not a good phrase to describe the TOD setting, because while it will be formatted, it might represent a relative duration. Or maybe we need a third setting or to use the same heurstic we use elsewhere for relative timestamps.

defunctzombie:
> think we could eliminate the actual word from the playback bar, and only show the dropdown chevron icon next to the timestamp. It's obvious whether you are looking at seconds or human time so we don't need to repeat that fact beyond providing access to the dropdown.

👍 

> Within the dropdown, I'd opt for more human readable words over short abbreviations.

The dropdown items could actually be the current times displayed in the format rather than words. So you select the visually exactly what you will get.

PR #2094, opened 2021-11-06
Remove "TOD" / "SEC" label from playback bar, show actual time representations in dropdown

jhurliman:
**User-facing changes**

The "TOD" and "SEC" (previously "ROS") labels have been removed from the playback controls. The dropdown menu items now also show the actual representative time strings instead of descriptions, unless there is no open data source.

Screen Shot 2021-11-05 at 5 00 04 PM

### DIFF
--- a/packages/studio-base/src/components/PlaybackControls/PlaybackTimeDisplayMethod.tsx
+++ b/packages/studio-base/src/components/PlaybackControls/PlaybackTimeDisplayMethod.tsx
@@ -63,14 +63,18 @@ const PlaybackTimeDisplayMethod = ({
       setPlaybackConfig({ timeDisplayMethod: newTimeDisplayMethod }),
     [setPlaybackConfig],
   );
-  const currentTimeString = useMemo(() => {
-    if (currentTime) {
-      return timeDisplayMethod === "SEC"
-        ? formatTimeRaw(currentTime)
-        : formatTime(currentTime, timezone);
-    }
-    return undefined;
-  }, [currentTime, timeDisplayMethod, timezone]);
+  const timeRawString = useMemo(
+    () => (currentTime ? formatTimeRaw(currentTime) : undefined),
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/components/PlaybackControls/PlaybackTimeDisplayMethod.tsx
```diff
@@ -218,14 +224,14 @@ const PlaybackTimeDisplayMethod = ({
             {
               canCheck: true,
               key: "TOD",
-              text: "Time of day (TOD)",
+              text: timeOfDayString ? timeOfDayString : "Time of Day",
```
defunctzombie, line 227:
```suggestion
              text: timeOfDayString ?? "Time of Day",
```

packages/studio-base/src/components/PlaybackControls/PlaybackTimeDisplayMethod.tsx
```diff
@@ -218,14 +224,14 @@ const PlaybackTimeDisplayMethod = ({
...
               isChecked: timeDisplayMethod === "TOD",
               onClick: () => setTimeDisplayMethod("TOD"),
             },
             {
               canCheck: true,
               key: "SEC",
-              text: "Seconds",
+              text: timeRawString ? timeRawString : "Seconds",
```
defunctzombie, line 234:
```suggestion
              text: timeRawString ?? "Seconds",
```

packages/studio-base/src/components/PlaybackControls/PlaybackTimeDisplayMethod.tsx
jhurliman: This is not correct. It doesn't handle the case of `timeOfDayString === ""` which is what I am checking for, not a nullish value.

PR #2095, opened 2021-11-06
Remove feature branch support from 3d panel

defunctzombie:
**User-Facing Changes**
No new visual changes. The feature comparison feature was removed in the last release.

**Description**
This was a webviz/cruise specific feature which was not documented nor supported for anything but the bag workflow with the 3d panel and raw message panel.
    
We recently removed support for loading multiple ros1 bag files as the feature was not documented and not well supported across the app. This is cleanup from that for the various bits of code in 3d panel.
    
The config shape does change so a forward migration is introduced.


jhurliman (approved):

packages/studio-base/src/panels/ThreeDimensionalViz/TopicTree/renderTreeNodes.tsx
```diff
@@ -91,68 +87,39 @@ export function getNamespaceNodes({
...
-      hasFeatureColumn &&
-      availableNamespacesByTopic[`${SECOND_SOURCE_PREFIX}${topicName}`]) ||
-      [],
-  );
-  const uniqueNamespaces = uniq([
-    ...Array.from(baseNamespacesSet),
-    ...Array.from(featureNamespacesSet),
-  ]);
-  const columns = hasFeatureColumn ? [0, 1] : [0];
+  const uniqueNamespaces = uniq([...Array.from(baseNamespacesSet)]);
```
jtbandes, line 97: probably not needed

packages/studio-base/src/panels/ThreeDimensionalViz/TopicTree/useTopicTree.ts
```diff
@@ -35,23 +34,16 @@ import {
...
   name?: string;
   namespace?: string;
-  isFeatureColumn?: boolean;
 }): string {
-  const prefixedTopicName = topicName
-    ? isFeatureColumn
-      ? `${SECOND_SOURCE_PREFIX}${topicName}`
-      : topicName
-    : undefined;
+  const prefixedTopicName = topicName ? topicName : undefined;
```
jtbandes, line 46: nit: remove 'prefixed' wording

jtbandes (approved):
Any plan to migrate old layouts to the new setting names?

Tested what I could using my markers sample bag. It looks like color overrides, linking, and highlighting are working to some degree at least.

https://user-images.githubusercontent.com/14237/140812089-8fb825fc-6346-4ecf-88f6-342c4248f45d.mov

defunctzombie:
> Any plan to migrate old layouts to the new setting names?

I added migration logic here https://github.com/foxglove/studio/pull/2095/files#diff-82950c46eaab76f4e40b4055a2a5dab9547442c4f03675b15157672f1e825721R62 which I believe covers forward migration only.

jtbandes:
Ok cool. Would be nice to avoid mutation though if possible.

PR #2096, opened 2021-11-06
Demonstrate <Worldview backgroundColor=[...]> not working

jhurliman:


jtbandes:
See https://github.com/foxglove/studio/pull/2100

Issue #2093, opened 2021-11-05
3d panel should have lighter background in light mode


amacneil:
**Description**
Maybe something similar to rviz? Our black feels too dark when using light theme.

![image](https://user-images.githubusercontent.com/637671/140589192-95a8916f-ae55-480e-a1f0-749eee0712a4.png)

![image](https://user-images.githubusercontent.com/637671/140589348-8a426aaf-7712-4565-89ea-3dbde9359153.png)

PR #2100, opened 2021-11-08
Add setting and theme-based default for 3D panel background color

jtbandes:
**User-Facing Changes**
Added a setting to the 3D panel to choose the background color. By default, the background color will be based on which color scheme is used in Studio.

Fixes https://github.com/foxglove/studio/issues/2093

image
image
image


jhurliman (approved):

amacneil:
Decision: leave dark mode background black, light mode background `#303030` (or something near that)

PR #2102, opened 2021-11-08
Only allow one emitState setTimeout to be running in live players

jhurliman:
**User-Facing Changes**

Performance improvement when using a live data source.

### Only allow one emitState setTimeout to be running in live players

Previously, any received message in a live player would trigger a call to emitState (assuming it was not already concurrently running in which case it would be debounced). Each emitState call would in turn call setTimeout(), queuing up a new emitState call 100 ms later. This caused far more emitState calls than intended. The fix is to set and clear a timer handle so at most one setTimeout() call is scheduled.


defunctzombie:
Is this no longer desired? It sounds like a desired change.

PR #2104, opened 2021-11-08
Unify local file loading into a single "Open file" option

jhurliman:
**User-Facing Changes**

The various individual file type options in the sidebar have been replaced with a single "Open file" option, plus "Open folder" which currently only supports ROS 2 bags. The connection sidebar has been renamed to "Data Source".

Screen Shot 2021-11-09 at 11 41 02 AM


defunctzombie:
This requires the OpenFileDataSource to know about all the file types we support which is different from the prior behavior where we could easily add a file type (via a new DataSourceFactory) and have it be supported. The idea behind the data source factories is to move away from a hard-coded place for our supported data sources. The current DataSourceFactory style also allows for extensions to indicate they can open a file by registering their file extensions. This moves us away from the flexibility of that approach.

amacneil:
Visually the sidebar looks a bit messy in your screenshot. Maybe it's due to the text wrapping.

I think we need to think about changing `ROS 1 Bag (Remote)` to just `Open URL` or something (similar to VLC).

For `Rosbridge (ROS 1 + ROS 2)` I think we could drop the parentheses, just `Rosbridge` is ok. The reason for original naming was to make it discoverable (e.g. ROS 2 users would clearly see multiple options), but maybe there is a better way to solve that. Maybe we need an alternative flow where you select your framework and we show you the applicable options (e.g. we could have a dropdown at the top which lets you filter the connection types by framework?).

defunctzombie:
The future is now. We've done this with the Open Dialog work.

Issue #2058, opened 2021-11-01
Notify user when a panel tries to open an invalid sibling panel


amacneil:
My two cents as a user on figuring out what to put for `type`: just add an error log message if the type I pass in isn't valid ([currently it fails silently](https://github.com/foxglove/studio/blob/c36bb326d71d752ac78a115d44a03911bd9e9515/packages/studio-base/src/components/Panel.tsx#L382)). Someone calling this method is going to be a developer anyway, so if it doesn't work the first thing we'll do is open up the dev console and look for errors.

_Originally posted by @jessicaaustin in https://github.com/foxglove/studio/issues/1914#issuecomment-955127169_

PR #2107, opened 2021-11-08
Open an unknown panel when a sibling panel type cannot be found

defunctzombie:
**User-Facing Changes**
When the user takes an action that might open a new sibling panel, but that panel type does not exist, the newly opened sibling panel will say "Unknown panel type: ...".

**Description**
Rather than fail silently to open a panel via the openSiblingPanel api, the new panel is opened with the requested type. If there is no panel of the type in the panel catalog, the new panel falls back to the "Unknown panel type" empty state.

Fixes: #2058

### DIFF
--- a/packages/studio-base/src/components/Panel.tsx
+++ b/packages/studio-base/src/components/Panel.tsx
@@ -41,6 +41,7 @@ import {
   updateTree,
   MosaicNode,
 } from "react-mosaic-component";
+import { useToasts } from "react-toast-notifications";
 import { useMountedState } from "react-use";
 
 import { useShallowMemo } from "@foxglove/hooks";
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/components/Panel.tsx
```diff
@@ -387,9 +385,22 @@ export default function Panel<
           return;
         }
         const siblingPanel = panelCatalog.getPanelByType(panelType);
+        // No panel exists for the requested panel type.
+        // Make a panel anyway to show an unknown panel placeholder to inform the user.
         if (!siblingPanel) {
```
jtbandes, line 390: Can we merge these code paths now since they are both adding a panel? The only thing we can't do if the panel is not found is get the defaultConfig, right?

defunctzombie: Good suggestion. Less code this way :D

jtbandes:
Would be nice to add a story that exercises this behavior

PR #2116, opened 2021-11-09
Add link to website in Foxglove Studio app

esthersweon:
**User-Facing Changes**
- Fixed "dead-end" issue found by ahrefs.com

   Screen Shot 2021-11-09 at 1 38 15 AM

- Added link back to website from studio.foxglove.dev and the Foxglove Studio desktop app

  Screen Shot 2021-11-09 at 1 43 04 AM

### DIFF
--- a/packages/studio-base/src/components/Sidebar/index.tsx
+++ b/packages/studio-base/src/components/Sidebar/index.tsx
@@ -41,6 +41,7 @@ export type SidebarItem = {
   title: string;
   badge?: Badge;
   component?: React.ComponentType;
+  url?: string;
 };
 
 const useStyles = makeStyles({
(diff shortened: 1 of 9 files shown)


packages/studio-base/src/components/Sidebar/index.tsx
```diff
@@ -107,7 +107,18 @@ export default function Sidebar<K extends string>({
         throw new Error(`Missing sidebar item ${key}`);
       }
       const { title, iconName } = item;
-      return (
+      return key === "website" ? (
```
jtbandes, line 110: Key is supposed to be something the user chooses and Sidebar should not have any knowledge of specific key values. They are just meant to be used as keys in a Map / React keys.

jtbandes (commented):
Having this in the sidebar feels a little too prominent to me, and also breaks the pattern where the sidebar buttons are like "tabs" that just open up a panel on the left hand side.

PR #2122, opened 2021-11-09
Internal: Add PlotDataItem to Plot/internalTypes

defunctzombie:
**User-Facing Changes**
None

**Description**
Rather than using TooltipItem for plot data items, use PlotDataItem. I find this naming more representative of the information. The TooltipItem is used only in the context of tooltips rather than storing the plot data.


defunctzombie:
This is related to simplifying the plot tooltip. Trying to land a few basic changes separately from tooltip changes.

Issue #2099, opened 2021-11-08
Rosout panel settings disappears every time a new message appears in the panel


munzz11:
**Description**

- Version: 0.21.0
- OS: Ubuntu 20.04

**Steps To Reproduce**
1. Connect to ROS1 running on a separate machine
2. Add rosout panel to empty layout and start viewing rosout messages
3. Click the settings cog for the panel, every time there is a new message, the panel settings drop down disappears, meaning if you have a lot of log messages coming through, it can become almost impossible to change the panel settings  

**Expected Behavior**
Have the panel settings menu not disappear and be able to interact with it, when messages are coming through 

**Actual Behavior**
![Alt Text](https://i.ibb.co/CQYqj4G/ezgif-com-gif-maker.gif)

munzz11:
Source link of gif since is appears to freeze in github: https://i.ibb.co/CQYqj4G/ezgif-com-gif-maker.gif

foxymiles:
This looks like it's a consequence of fluentui's dismiss on scroll behavior:

https://github.com/microsoft/fluentui/issues/18080

I'm investigating some possible solutions.

PR #2123, opened 2021-11-09
Add workaround for fluent ui's dismiss on scroll behavior.

foxymiles:
**User-Facing Changes**
This is a somewhat hacky fix for FluentUI's default dismiss on scroll behavior. 

**Description**
FluentUI by default will dismiss any open Callout based widgets like dropdowns if any element in the window scrolls. This PR installs a window-level scroll event listener ahead of Fluent to prevent that from happening if the scroll event originates in a react-virtualized list like the lists we use for the log windows.

https://github.com/foxglove/studio/issues/2099


jtbandes (approved):
While hacky, this seems like a minimally invasive and targeted fix so I think it's worth it 👍

Issue #2047, opened 2021-10-29
Enable caching on studio.foxglove.dev


jtbandes:
When I refresh studio.foxglove.dev, I notice our various large JS assets are loaded with 200 responses rather than 304 (except main.js). It would be nice for these to be cached as much as possible, especially since content hashes are part of the resource names.

PR #2127, opened 2021-11-09
Attempt to make JS/WASM assets cacheable in Vercel config

jtbandes:
**User-Facing Changes**
None

**Description**
Attempting to address #2047 — it looks like Chrome does use its cache (see screenshot below...although confusingly it sometimes shows 200s even for responses that appear to be cached), but due to Vercel's default `public, max-age=0, must-revalidate` the browser always has to revalidate its caches with the server. Since these file paths include a content hash, it should be safe to allow them to be served from cache.

image

### DIFF
--- a/vercel.json
+++ b/vercel.json
@@ -4,7 +4,7 @@
   },
   "headers": [
     {
-      "source": "/((\\w+\\.)?[0-9a-f]{20}\\.(js|js\\.map|wasm))",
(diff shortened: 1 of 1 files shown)


defunctzombie:
I'm a bit surprised by this behavior from vercel. I'd have thought they would have a more obvious way for us to say which assets could be cached than this magical regex. 👍

PR #2129, opened 2021-11-10
Remove the Cruise car model, add a "404" mesh, support mesh marker alpha transparency

jhurliman:
**User-Facing Changes**

- The embedded Cruise car model has been removed. `PoseStamped` messages can only be visualized as arrows now
- If a mesh resource cannot be loaded, a "404" model will be shown instead of an error dialog
- Mesh markers support alpha transparency (via the mesh marker `color.a` field)

Screen Shot 2021-11-10 at 10 48 59 AM


defunctzombie:
Why would we load the car model rather than no mesh? Maybe a 3d foxglove logo mesh instead? Seems if the mesh resource is empty we should maybe show an error for the marker topic rather than fallback to the car model.

defunctzombie:
Yea - I am tempted to say we should maybe even drop the CarModel from the codebase?

amacneil:
Dropping car model entirely seems fine. We should add it to the demo bag via http url.

jhurliman:
> Yea - I am tempted to say we should maybe even drop the CarModel from the codebase?

What do we use for the Pose viz then?

defunctzombie:
> What do we use for the Pose viz then?

An arrow?

Or a horse sized duck. But arrow seems simpler.

amacneil:
Arrow seems fine. We could replace the current car model radio option with a banana.

Maybe in the future we let you select a local model file to use for pose, or link your org's github account so that a selected asset can be shared across team members?

That said, the way Cruise put the car at pose always seemed weird to me. Isn't that what transforms are for? Surely most people would use `robot_description` to place their robot rather than a pose topic?

jhurliman:
@amacneil it's all just nomenclature, but PoseStamped is commonly used for goal topics where you don't necessarily want to define a coordinate frame and you don't want to do any interpolation on your goal. The Cruise usage was somewhat odd  because the ego pose was double-published as a transform and a standalone topic, but that doesn't invalidate the concept of PoseStamped in general.

I can't tell what in this thread is a serious change request and what is just thinking out loud. There is a toggle option for visualizing a model for PoseStamped messages that is currently a car. Do we want to remove the option to render a model for PoseStamped (and if so, do we want to get rid of the concept of a default model entirely)? Do we want to keep the option but change the model? Is the new model supposed to be a banana, and if so which banana?

This is blocking the release of a new demo bag, which also holds up our next release and next blog post.

defunctzombie:
If we don't have a robot description to show then we should show an arrow/box/whatever. We should delete the car model from the repo and remove the toggle that displays a car. Our app is not specific to autonomous driving.

jtbandes:
Can we use some user-understandable text instead of the word "404"? Something like "mesh not found"?

defunctzombie:
@jhurliman why do we need a 404 mesh marker? We already have a mechanism for reporting errors for marker topics. Please use that mechanism.

Issue #2119, opened 2021-11-09
Log level filter dropdown jumps to the top left of the screen while the log is scrolling.


foxymiles:
**Description**
There seems to be a dropdown positioning glitch when the log level window window is scrolling.

**Steps To Reproduce**

1. Set up a workspace with an actively streaming Rosout window.
2. Click on the log level filter dropdown to open it.
3. Position your cursor over the error or fatal items in the dropdown but don't click it.
4. Wait for log events to scroll the underlying window.
5. You'll notice that the filter dropdown jumps back and forth from the correct position to the upper left of the entire workspace.

https://user-images.githubusercontent.com/93935560/140984277-bd621e33-5ef8-420c-8b66-31b359165a82.mp4

amacneil:
Fixed by https://github.com/foxglove/studio/pull/2131

PR #2131, opened 2021-11-10
Convert log window toolbar to non-floating to fix jumping menu issue.

foxymiles:
**User-Facing Changes**
As per @defunctzombie suggestion, this makes the toolbar at the top of the log window static instead of show on hover. This seems like a simple solution to https://github.com/foxglove/studio/issues/2119

**Description**
This simply removes the `floating` attribute from the toolbar.

https://github.com/foxglove/studio/issues/2119

### DIFF
--- a/packages/studio-base/src/panels/Log/index.tsx
+++ b/packages/studio-base/src/panels/Log/index.tsx
@@ -90,7 +90,7 @@ const LogPanel = React.memo(({ config, saveConfig }: Props) => {
 
   return (
     <Stack verticalFill>
-      <PanelToolbar floating helpContent={helpContent} additionalIcons={topicToRenderMenu}>
+      <PanelToolbar helpContent={helpContent} additionalIcons={topicToRenderMenu}>
         <FilterBar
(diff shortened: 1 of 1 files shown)


defunctzombie:
👍 we'll give this a go and see if anyone mentions they preferred the old way where it was hidden. If they do we can maybe turn it into a panel setting or find some way to show/hide via toggle rather than the mouse cursor behavior.

jtbandes:
Before merging you'll need to click through the "UI Tests" in the list of github status checks and manually approve/deny changes. Looking at those I'm seeing there is a padding above the log content that was intended to make sure the logs are always below the floating toolbar — and now that it's not floating it's just some extra whitespace that should be removed.

amacneil:
I'm +1 on this change. Showing the filter all the time is more consistent with how the Diagnostic Summary panel works too, which is nice.

defunctzombie:
Yep - we spoke with Andrew and this aligns with our desire to remove more randomly appearing/disappearing elements.

Issue #708, opened 2021-04-27
Panel disappears if dragged back and forth between Tab mosaics


amacneil:
via @jtbandes:

I can consistently repro this in the storybook scenario I newly added in https://github.com/foxglove/studio/pull/410.

This bug also exists in upstream webviz when tested with the same layout.

https://user-images.githubusercontent.com/637671/116322875-c862e700-a771-11eb-8a9d-f29f914489e0.mov

jtbandes:
There is another bug, probably related, which causes an error when a drag from a tab is **canceled** (press esc while dragging):


https://user-images.githubusercontent.com/14237/116940186-48bd9800-ac09-11eb-84a9-8177b672c62f.mov



[Mosaic drag testing2.json.zip](https://github.com/foxglove/studio/files/6418086/Mosaic.drag.testing2.json.zip)

defunctzombie:
@amacneil I'm no longer able to reproduce this with 0.20.1 (with the production app). Maybe the ticket @jtbandes referenced fixed this or other fixes we've made to id tracking. Going to close this as neither issue shown in the ticket I am able to reproduce. If you can reproduce with the app or current stories please re-open with new instructions.

foxymiles:
@jtbandes I think I have a fix for the crash adding a new panel. I'm looking into the panel disappearance issue now.

foxymiles:
I found an easier repro using the layout @jtbandes posted. If you drag the panel out of its original container, then try to drag it back but position the cursor so that no drop target is indicated and then drop the panel it just vanishes when you complete the drop.

https://user-images.githubusercontent.com/93935560/141116493-6e7640da-2e5d-43e0-9311-6a06d8c0b405.mp4

foxymiles:
https://github.com/foxglove/studio/pull/2132 Should fix the crash on adding a  sub panel to a fresh tab panel.

amacneil:
Noting here that we have not seen any issues here that don't involve the tab panel.

From @foxymiles: This issue is possibly related to garbage collection that runs at the end of the tab layout change process.

jtbandes:
The root cause of https://github.com/foxglove/studio/issues/708#issuecomment-965103972 is that `sourceTabChildConfigs` doesn’t account for when the target tab is itself a child of the source tab, so the source child config ends up overwriting the updated target tab config with its original value here:

![image](https://user-images.githubusercontent.com/14237/142548002-140b5601-4934-4abe-9350-66cb9d3cd03a.png)

PR #2132, opened 2021-11-10
Handle missing tab panel config gracefully.

foxymiles:
**User-Facing Changes**
This fixes an issue adding a sub panel to a freshly created tab panel.

**Description**
Looks like in some cases the tab panel is configured as an empty object. We should probably find the root cause of this but it's worth adding some logic to handle it here in case there are existing layouts with this issue.

https://github.com/foxglove/studio/issues/708

### DIFF
--- a/packages/studio-base/src/util/layout.ts
+++ b/packages/studio-base/src/util/layout.ts
@@ -60,6 +60,10 @@ export function isTabPanel(panelId: string): boolean {
   return getPanelTypeFromId(panelId) === TAB_PANEL_TYPE;
 }
 
+export function isTabPanelConfig(config: PanelConfig | undefined): config is TabPanelConfig {
+  return config != undefined && "tabs" in config && "activeTabIndex" in config;
+}
+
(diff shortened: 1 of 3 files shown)


jtbandes (approved):
Thanks for updating the tests! Not sure the isTabPanelConfig will be useful in general since a `Tab!...` id is really the best way to know that the panel is supposed to be a tab panel. That's why I suggested just `Partial<TabPanelConfig>`

packages/studio-base/src/util/layout.ts
```diff
@@ -60,6 +60,10 @@ export function isTabPanel(panelId: string): boolean {
   return getPanelTypeFromId(panelId) === TAB_PANEL_TYPE;
 }
 
+export function isTabPanelConfig(config: PanelConfig | undefined): config is TabPanelConfig {
```
esthersweon, line 63: There is already a `validateTabPanelConfig` util.

foxymiles: Oh thanks I didn't see that!

PR #2134, opened 2021-11-10
Add time label to playback scrubber hover ticks

defunctzombie:
**User-Facing Changes**
Hove time display above the hover ticks.

Screen Shot 2021-11-10 at 9 02 13 AM
Screen Shot 2021-11-10 at 9 02 03 AM

**Description**
When the hover time is set by panels a pair of tick marks shows up on the playback control scrubber to indicate the hover time location. This change adds a time label above the hover ticks.

The time label follows the time format selected in the playback controls.


defunctzombie:
This change spun out of UX review for simplifying plot tooltips. We will remove the time display from the tooltip and having this will allow the user to see the timestamp for where they are hovering.

PR #2136, opened 2021-11-10
Add setting to optionally show log timestamps in local time.

### LINKED ISSUE #724: Log panel should follow the user's timezone settings
via @defunctzombie:

Log panel current displays timestamps in unix time - we should have a setting to display the time in human readable format and follow the app's timezone settings

foxymiles:
**User-Facing Changes**
Allows the user to choose between unix timestamp and local time format for log messages.

**Description**
This adds a new user preference setting under the timezone selector. 

**Note that this also removes the per-layout setting for timeout and applies the global time format everywhere we display formatted times.**

Screen Shot 2021-11-12 at 11 58 27 AM

Fixes #724

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,9 @@ $ yarn start        # launch electron
 # To launch the browser app:
 $ yarn web:serve
 
+# To launch the storybook:
+$ yarn storybook
+
 # Advanced usage: running webpack and electron on different computers (or VMs) on the same network
 $ yarn serve --host 192.168.xxx.yyy         # the address where electron can reach the webpack dev server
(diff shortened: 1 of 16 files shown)


defunctzombie:
How does this differ from the "Display timestamps in" setting? If I select a timezone in that drop down. And then select "Local time" for timestamp format what format do I get for my timestamps?

Thoughts on using the playback controls time dropdown as the toggle for this setting rather than in the settings sidebar?

foxymiles:
@defunctzombie The idea here is to select a plain unix timestamp vs a localized time string. We could probably come up with better terminology.

foxymiles:
Note: this should synchronize with the setting on the playback controls.

Issue #1537, opened 2021-07-22
Add "Download App" CTA to web build; Allow visitors to easily navigate back to foxglove.dev


esthersweon:
Currently no obvious way to go from web build demo to an app download page.

Adding this button would be useful before web build has login flow.

amacneil:
I think this is a good idea but let’s not make it too distracting. A sidebar icon with red “unread messages” bubble or something unobtrusive like that?

@2metres probably has better suggestions

PR #2140, opened 2021-11-11
Add a "Help" section to the sidebar that contains panel docs & other resources

esthersweon:
**User-Facing Changes**
- Added a "Help" section to the app sidebar that contains panel docs (click help icon in panel to open) & other resources (links to website, etc.)

Screen Shot 2021-11-12 at 3 30 34 PM
Screen Shot 2021-11-12 at 3 31 06 PM

Resolves https://github.com/foxglove/studio/issues/1537

Follow-up PRs should cover:
- Only show "Download app" link when in web app
- Audit all other in-modal docs in the app (e.g. "Layouts", "Variables", "Message path syntax", "Keyboard shortcuts", etc.) – should they be in sidebar tab? Stay in modal? Be a collapsible section in the current tab?
- Desktop app's Help menu should reflect what is in the Help sidebar
- Migrate docs to mdx (like "Keyboard shortcuts" docs)


packages/studio-base/src/components/HelpSidebar/index.tsx
```diff
@@ -0,0 +1,100 @@
...
+  { text: "Foxglove Studio", url: "https://foxglove.dev/studio" },
+  { text: "Foxglove Data Platform", url: "https://foxglove.dev/data-platform" },
+];
+
+const legalLinks = [
+  { text: "License", url: "https://foxglove.dev/legal/studio-license" },
+  { text: "Privacy", url: "https://foxglove.dev/legal/privacy" },
+];
+
+export default function HelpSidebar(): React.ReactElement {
```
defunctzombie, line 33:
```suggestion
export default function HelpSidebar(): JSX.Element {
```

defunctzombie:
Does this work with extension panels? Or do we not have a way for those to provide help text yet?

esthersweon:
> Does this work with extension panels? Or do we not have a way for those to provide help text yet?

@defunctzombie - does not work with extension panels. We would probably want to keep that in the extensions tab, and maybe keep the UI in both tabs more consistent.

PR #2141, opened 2021-11-11
Convert dropdown to functional component.

foxymiles:
**User-Facing Changes**
This should be transparent to end users.

**Description**
This converts the `Dropdown` component from a class component to a functional component.

Part of https://github.com/foxglove/studio/issues/1426


defunctzombie:
👍 seems right

defunctzombie (approved):

defunctzombie:
LGTM - I think we can merge these in and later replace with MUI etc as the time comes. Having them be functional makes it more consistent with everything else we maintain in the codebase.

PR #2142, opened 2021-11-11
Convert KeyListener to function component.

foxymiles:
**User-Facing Changes**
This should be transparent to end users.

**Description**
This converts `KeyListener` into a functional component.

Part of https://github.com/foxglove/studio/issues/1426.

### DIFF
--- a/packages/studio-base/src/components/KeyListener.tsx
+++ b/packages/studio-base/src/components/KeyListener.tsx
@@ -11,92 +11,84 @@
 //   found at http://www.apache.org/licenses/LICENSE-2.0
 //   You may not use this file except in compliance with the License.
 
-type KeyHandlers = {
-  // Invoke handler for matcking key.
-  // By default, preventDefault() is invoked on the event. Return false to allow the default
-  [key: string]: (event: KeyboardEvent) => void | boolean | undefined;
-};
+import { ReactElement, useCallback, useEffect, useRef } from "react";
+
+// Invoke handler for matching key.
+// By default, preventDefault() is invoked on the event. Return false to allow the default.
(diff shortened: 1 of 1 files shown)


jtbandes:
It might be better to convert this to a hook like useKeyListener, or just replace it with existing hooks from `react-use` (or a plain useEffect + addEventListener) if it doesn't feel like it's used in enough places to be worth maintaining a custom hook

foxymiles:
@jtbandes It looks like this in some cases depends on rendering a new div and attaching a listener to that. Would we be able to duplicate that functionality in a hook? I guess it would if we changed it to require the parent element as a parameter.

jtbandes:
Right... the way I've usually seen this done with hooks is that the hook returns a ref which you can then attach to your desired element. For example: https://github.com/streamich/react-use/blob/master/docs/useScratch.md

foxymiles:
@jtbandes I wonder if it would be worth deprioritizing converting this component? That feels like a lot of changes for a small benefit.

defunctzombie:
LGTM

defunctzombie:
Still planning to merge this?

foxymiles:
> Still planning to merge this?

I was a little gun shy on this one since it seemed like the benefit was small and possible breakage was significant but I'll test it some more and merge it.

defunctzombie:
We can also punt on it if we want to. I agree its not critical but also seemed like it was done. I think the merge conflict is some fix I put into KeyListener around modifier keys.

foxymiles:
@defunctzombie I fixed the merge conflict and tidied this up a touch. It's working in my testing.

defunctzombie:
:shipit:

Issue #2091, opened 2021-11-05
Adding first panel to layout takes you to panel settings, additional panels do not


amacneil:
**Description**
- Version: v0.20.1
- OS: macOS, web

**Steps To Reproduce**
1. create a new layout
2. add a panel to it. observe: left sidebar changes to "panel settings"
3. go back to "add panel" sidebar and add another panel. observe: left sidebar stays in "add panel" mode

**Expected Behavior**
Consistency between these two actions.

I would prefer we never take you to the panel settings page after adding a panel. More often than not when building a layout I want to add multiple panels before configuring any settings, and it is jarring having the sidebar change after adding a panel.

jtbandes:
Seems to be a difference between drag/drop and just clicking an item in the list. Clicking works as expected.

jtbandes:
> I would prefer we never take you to the panel settings page after adding a panel. More often than not when building a layout I want to add multiple panels before configuring any settings, and it is jarring having the sidebar change after adding a panel.

The point of automatically opening settings was discoverability. For new users, it's probably not obvious how to configure panels so we wanted to start teaching them to use the sidebar. Any ideas for other ways we can improve discoverability?

amacneil:
> Seems to be a difference between drag/drop and just clicking an item in the list. Clicking works as expected.

Interesting. Clicking seems to always take you to the panel settings. Dragging only takes you to the panel settings for the first panel; subsequent panels do not take you to panel settings.

> The point of automatically opening settings was discoverability. For new users, it's probably not obvious how to configure panels so we wanted to start teaching them to use the sidebar. Any ideas for other ways we can improve discoverability?

We have a cog icon on every panel, and clicking it shows you the `Panel settings` option. That seems pretty discoverable?

Since many panel options (e.g. selected topic) are displayed directly in the panel, I don't find the panel settings something that is immediately needed while building a layout.

Curious to hear what others think of this behavior. cc @defunctzombie @jhurliman @2metres @esthersweon.

At the very least, we should fix the inconsistency, so that adding a panel either always opens the panel settings or never opens the panel settings.

amacneil:
per slack conversation - let's proceed with never automatically opening sidebar settings for now. we can revisit in the future once we are happier with how panel configuration works.

PR #2143, opened 2021-11-11
Don't open panel settings after adding a new panel via the add panel sidebar.

foxymiles:
**User-Facing Changes**
This changes the behavior of the add panel sidebar. Instead of switching to the panel settings sidebar after adding a new panel it will keep the add panel sidebar open.

**Description**
This keeps  the add panel sidebar open after adding a new panel via a click or drag. The goal is to make it easer and more ergonomic to add a series of panels in quick succession.

This is an attempt to address #2091

### DIFF
--- a/packages/studio-base/src/hooks/useAddPanel.ts
+++ b/packages/studio-base/src/hooks/useAddPanel.ts
@@ -8,20 +8,17 @@ import {
   useCurrentLayoutActions,
   useSelectedPanels,
 } from "@foxglove/studio-base/context/CurrentLayoutContext";
-import { useWorkspace } from "@foxglove/studio-base/context/WorkspaceContext";
 import { getPanelIdForType } from "@foxglove/studio-base/util/layout";
 
 export default function useAddPanel(): (selection: PanelSelection) => void {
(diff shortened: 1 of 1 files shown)


defunctzombie:
> This changes the behavior of the add panel sidebar.

Would be nice to call out what the change in behavior is.

foxymiles:
> > This changes the behavior of the add panel sidebar.
> 
> Would be nice to call out what the change in behavior is.

Should I elaborate on this?

`This changes that behavior to keep the add panel sidebar open instead.`

defunctzombie:
@foxymiles What you've written now is 👍 More clear what happens.

defunctzombie:
This seems right. It has the benefit that when you add a panel view the Add Panel sidebar you can continue to add more panels since that sidebar item remains open?

foxymiles:
> This seems right. It has the benefit that when you add a panel view the Add Panel sidebar you can continue to add more panels since that sidebar item remains open?

Yes that's correct.

foxymiles:
Note: this should synchronize with the setting on the playback controls.

PR #2145, opened 2021-11-11
Avoid showing the hover time above hover ticks when showing tooltip

defunctzombie:
**User-Facing Changes**
The hover ticks time is only shown when hovering over panels rather than the playback controls scrubber.

Screen Shot 2021-11-11 at 12 53 13 PM

**Description**
The hover time was added in https://github.com/foxglove/studio/pull/2134. However this presents the same information as the tooltip when a user hovers on the scrubber. This change avoids showing the hover time when a user hovers on the scrubber since it is shown in the tooltip.

### DIFF
--- a/packages/studio-base/src/components/PlaybackControls/PlaybackBarHoverTicks.tsx
+++ b/packages/studio-base/src/components/PlaybackControls/PlaybackBarHoverTicks.tsx
@@ -82,9 +82,13 @@ function displayMethodSelector(state: LayoutState): TimeDisplayMethod {
 
 type Props = {
   componentId: string;
+  // When true, this will display the hover time above the hover ticks
+  displayHoverTime: boolean;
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/components/PlaybackControls/PlaybackBarHoverTicks.tsx
```diff
@@ -82,9 +82,13 @@ function displayMethodSelector(state: LayoutState): TimeDisplayMethod {
 
 type Props = {
   componentId: string;
+  // When true, this will display the hover time above the hover ticks
+  displayHoverTime: boolean;
```
jtbandes, line 86: Why not use the componentId for this? Isn't the whole point of componentId for components to be able to identify themselves or each other?

PR #2146, opened 2021-11-11
Internal: change logic for scrubber hover time detection

defunctzombie:
**User-Facing Changes**
None

**Description**
Implement feedback from: https://github.com/foxglove/studio/pull/2145#discussion_r747817351

### DIFF
--- a/packages/studio-base/src/components/PlaybackControls/Scrubber.tsx
+++ b/packages/studio-base/src/components/PlaybackControls/Scrubber.tsx
@@ -237,10 +237,7 @@ export default function Scrubber(props: Props): JSX.Element {
           renderSlider={renderSlider}
         />
       </div>
-      <PlaybackBarHoverTicks
-        componentId={hoverComponentId}
-        displayHoverTime={tooltipState == undefined}
-      />
+      <PlaybackBarHoverTicks componentId={hoverComponentId} />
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/components/PlaybackControls/Scrubber.tsx
```diff
@@ -237,10 +237,7 @@ export default function Scrubber(props: Props): JSX.Element {
           renderSlider={renderSlider}
         />
       </div>
-      <PlaybackBarHoverTicks
-        componentId={hoverComponentId}
-        displayHoverTime={tooltipState == undefined}
-      />
+      <PlaybackBarHoverTicks componentId={hoverComponentId} />
```
jtbandes, line 240: Maybe consider renaming this prop to something that makes it more obvious this component id is the one that corresponds to the play back bar itself and should be ignored for showing the hover time.

jtbandes (approved):

packages/studio-base/src/components/PlaybackControls/Scrubber.tsx
defunctzombie: Have any name suggestions in mind? Happy to make the change but don't know what else I'd call it.

Issue #1868, opened 2021-09-15
Store and load current connection in URL


amacneil:
I should be able to bookmark the current page with a websocket connection, and come back to it later.

- When rosbridge connection is used, add this to the URL like `studio.foxglove.dev/?source=rosbridge&conn=wsblahblah`
- Loading this URL should load the rosbridge player
- https://github.com/foxglove/studio/pull/1919/files#r719059578

defunctzombie:
The user-story here is a desire to open the web version of studio with a specific player and parameters. This won't work for all players. We need to be careful about how we introduce/surface some of these concepts and trying to url/browser specific things deep within the app. This was something that I had to extract from PlayerManager and I'd like to avoid us putting it back in the same way since we have web and desktop.

This is also related to our data platform player and `foxglove://` url handling for desktop.

amacneil:
Good point - might be worth separating "layout in the url" from "connection details in the URL".

- Layout makes sense for shared layouts, and personal synced layouts, but unclear how we want to handle localstorage layouts? Maybe it is ok if you save a personal or localstorage layout URL and that URL only works on your computer, since other people don't have access to it?
- Connection details
  - Fairly straight forward for remote connections (ros1+2, rosbridge, http bag, foxglove DP)
  - Maybe also ok for file based connections? We can just have an encoded file path in the URL? Still nice that I can bookmark something and come back to it.

There is a separate discussion about how we can make these URLs "shareable" within the desktop build of Studio (Linear is a good example of prior art, Slack also has some "copy link" options although at per-message level rather than app level).

defunctzombie:
Related: https://github.com/foxglove/studio/issues/1848

defunctzombie:
Re-assigned to @miles who's actively working on this in https://github.com/foxglove/studio/pull/2171

miles:
How is it my username being used on this project? I keep getting notifications on it.

foxymiles:
@miles I think @defunctzombie meant to tag me, @foxymiles .

defunctzombie:
@miles Yea - sorry about that - the username autocomplete led me astray.

PR #2149, opened 2021-11-12
Change publicPath to "auto".

iwanders:
This allows hosting the files from any path on the webserver as opposed to just from the root.

**User-Facing Changes**
Should be none.

**Description**

This ensures the webpack output can be hosted from any publicPath instead of just from the `/` path on the server. The magic `"auto"` string feels a bit sketchy, but it is [what the documentation states](https://webpack.js.org/guides/public-path/#automatic-publicpath).

For context, I built the files using the provided Dockerfile (thanks for providing this!) using the following;
```
cd studio
docker build -t foxglove_build_static .
docker cp $(docker create --rm foxglove_build_static):/src/ ../build/
```
I wanted to serve these files form a subpath, like `https://hostname/foxglove/` because the root of this webserver is used for other things. I didn't want to host it on another instance or port because the bags I wanted to look at  are hosted on this webserver, hosting it from the same server avoids any CORS issues.

Upon testing I discovered that specifying the bag to load in the url isn't supported yet (#1868), but thought I might as well contribute this tiny change to make additional testing easier in the future.

fyi @jasonimercer

### DIFF
--- a/web/webpack.config.ts
+++ b/web/webpack.config.ts
@@ -97,7 +97,7 @@ const mainConfig = (env: unknown, argv: WebpackArgv): Configuration => {
     devtool: isDev ? "eval-cheap-module-source-map" : "source-map",
 
     output: {
-      publicPath: "/",
+      publicPath: "auto",
 
       // Output filenames should include content hashes in order to cache bust when new versions are available
       filename: isDev ? "[name].js" : "[name].[contenthash].js",


defunctzombie (approved):

iwanders:
Thanks for the quick :+1:  @defunctzombie 

This CLA requires me to (point 5 of [the CLA](https://cla-assistant.io/foxglove/studio?pullRequest=2149)):
> You represent that each of Your Contributions is Your original creation

There is no way I can claim this is an original creation, it's a direct copy paste from the webpack documentation. Which is licensed under [Creative Commons BY 4.0](https://webpack.js.org/license/) itself, so that alone makes this a no-go. Besides this CLA there's now the legal terms of `CLAassistant` [here](https://gist.github.com/CLAassistant/3a73e4cd729c9d0a6e30) which I would have to read and agree to as well (which comes in at 31 kilobytes of text).

I'm all for contributing and filing small quick PRs to help improve software, but I don't want to spend the weekend interpreting the CLA itself and the terms of CLAassistant. I hope everyone agrees that reading 31 kilobytes of text is a steep price to pay to get 5 character modified, especially for such a small QOL improvement. I'm not going to take the time to go through all of that, easiest solution would be to wait for the PR to go stale and be closed, but I thought I'd provide context and properly close the PR myself.

If there's a CLA, this should be stated upfront in the [contributing section of the readme](https://github.com/foxglove/studio/blob/48366d18c09432b43f3c0514192e62033d1b0ae7/README.md#contributing) instead of only when a PR is filed.

defunctzombie:
Thanks for the feedback regarding the README CLA link. I've added it via this PR https://github.com/foxglove/studio/pull/2150 so it is clear for new contributors prior to making a PR.

amacneil:
@iwanders thank you for taking time to share your feedback and frustration around this. 

For future reference - we use the [Apache CLA](https://www.apache.org/licenses/icla.pdf) which is fairly common in the industry, to ensure contributions are appropriately licensed.

Similar requirements also exist purely your use of GitHub, quoting the [GitHub Terms of Service](https://docs.github.com/en/github/site-policy/github-terms-of-service#6-contributions-under-repository-license):

> 6. Contributions Under Repository License
>
> Whenever you add Content to a repository containing notice of a license, you license that Content under the same terms, and you agree that you have the right to license that Content under those terms. If you have a separate agreement to license that Content under different terms, such as a contributor license agreement, that agreement will supersede.
>
> Isn't this just how it works already? Yep. This is widely accepted as the norm in the open-source community; it's commonly referred to by the shorthand "inbound=outbound". We're just making it explicit.

iwanders:
@amacneil , thanks for your followup. I'm not knowledgeable in the area of CLA's and whenever I see one I just move along, so appreciate you sharing the link that this is the Apache CLA, I had not identified this as a common document.

I'm a strong believer in the `inbound=outbound` situation and as you say, this is explicitly stated in Github's terms of service, which means everyone who files a PR agrees the original license applies to the work being contributed. As far as I know, this also means that requiring CLA's to be signed outside of Github is unnecessary, because everyone on Github already agreed that a contribution will have the same license as the original work it built on.

Requiring a CLA to be signed through a third party means that besides the CLA, I also have to agree to the [terms of use](https://gist.github.com/CLAassistant/3a73e4cd729c9d0a6e30) of that external platform used to collect those signatures. This requires significant time-investment, reading and interpreting that document would require more time than I'm willing to invest right now.

Either way, appreciate the quick improvement of the readme and the followup about the CLA itself. :+1:

PR #2154, opened 2021-11-15
Pass ROS parameters to extension panels.

foxymiles:
**User-Facing Changes**
This allows extension panels to opt in to receiving ROS parameters as well as messages. 

**Description**
This extends the `PanelExtensionContext` and `RenderState` APIs to include ROS parameters. Extension panels can request parameter values and updates by adding `parameters` to the render state fields they want to watch:

``` context.watch("parameters");```

Parameters are passed as a `Record<string, ParameterValue>` object.

Partially addresses #1988 

We'll need to update the extension panel API docs to include the new `parameters` field.

### DIFF
--- a/packages/studio-base/src/components/PanelExtensionAdapter.tsx
+++ b/packages/studio-base/src/components/PanelExtensionAdapter.tsx
@@ -3,7 +3,6 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
 import { useTheme } from "@fluentui/react";
-import { isEqual } from "lodash";
 import { CSSProperties, RefCallback, useCallback, useMemo, useRef, useState } from "react";
 import { v4 as uuid } from "uuid";
 
@@ -172,7 +171,7 @@ function PanelExtensionAdapter(props: PanelExtensionAdapterProps): JSX.Element {
 
     if (watchedFieldsRef.current.has("parameters")) {
       const parameters = playerState.activeData?.parameters ?? EmptyParameters;
-      if (!isEqual(parameters, renderState.parameters)) {
+      if (parameters !== renderState.parameters) {
         shouldRender = true;
         renderState.parameters = parameters;
       }
--- a/packages/studio-base/src/components/PanelExtensionAdapter.tsx
+++ b/packages/studio-base/src/components/PanelExtensionAdapter.tsx
@@ -3,6 +3,7 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
 import { useTheme } from "@fluentui/react";
+import { isEqual } from "lodash";
 import { CSSProperties, RefCallback, useCallback, useMemo, useRef, useState } from "react";
 import { v4 as uuid } from "uuid";
 
@@ -29,6 +30,7 @@ import {
 } from "@foxglove/studio-base/context/HoverValueContext";
 import {
   AdvertiseOptions,
+  ParameterValue,
   PlayerCapabilities,
   PlayerState,
 } from "@foxglove/studio-base/players/types";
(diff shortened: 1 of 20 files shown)

### CODE REVIEW

File: packages/studio/index.d.ts [hunk 1]
```diff
@@ -3,6 +3,8 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
 declare module "@foxglove/studio" {
+  export { ParameterValue } from "@foxglove/studio-base/players/types";
```
defunctzombie, line 6: This should be flipped around. index.d.ts should never include anything from studio-base since we publish the `@foxglove/studio` package without depending on `@foxglove/studio-base`.

defunctzombie:
Are we sure "parameters" is a generalizable enough concept? Seems like it probably is? Any better alternative names for it? Should it just be key/value pairs or do we think there's something else that we should be representing.

@jtbandes @jhurliman

packages/studio/index.d.ts
foxymiles: Thanks I've switched this around.

packages/studio/index.d.ts
```diff
@@ -69,6 +82,11 @@ declare module "@foxglove/studio" {
      */
     allFrames?: readonly MessageEvent<unknown>[];
 
+    /**
+     * Map of current parameter values.
+     */
+    parameters?: readonly Record<string, ParameterValue>;
```
defunctzombie, line 88: Should we use a `Map` here rather than Record? I'm always unsure/conflicted on what is the better way to expose such a key/value concept. I know for a long time there was no Map in javascript so everything key/value was always Objects but now that we have Map does it make more sense to use that?

Somewhat related - you could type this as `readonly ParameterStruct;` since that is the same type.

foxymiles: Good point on the typing. I was also unsure whether to provide this as a a `Map`. Right now I have this as a record instead since in my experience they tend to play better with state managers like Redux etc than Maps but I could easily be convinced to switch it to a `Map`.

packages/studio/index.d.ts
```diff
@@ -3,6 +3,19 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
 declare module "@foxglove/studio" {
+  // Valid types for parameter data (such as rosparams)
+  export type ParameterValue =
+    | undefined
+    | boolean
+    | number
+    | string
+    | Date
```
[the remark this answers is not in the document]
foxymiles: Good question. Looks like these are the types ROS considers native right?

32-bit integers
booleans
strings
doubles
iso8601 dates
lists
base64-encoded binary data

http://wiki.ros.org/Parameter%20Server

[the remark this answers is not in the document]
defunctzombie: Huh - I'm tempted to drop dates since it feels like thats a bit weird compared to the others and just pass those along as strings. Its also different from the ROS Time format (seconds/nanoseconds).

@jhurliman @jtbandes if we don't support deserializing iso8601 strings to Date objects is that ok? Do we handle doing that today?

packages/studio-base/src/components/PanelExtensionAdapter.tsx
```diff
@@ -166,6 +170,14 @@ function PanelExtensionAdapter(props: PanelExtensionAdapterProps): JSX.Element {
       }
     }
 
+    if (watchedFieldsRef.current.has("parameters")) {
+      const parameters = playerState.activeData?.parameters ?? EmptyParameters;
+      if (!isEqual(parameters, renderState.parameters)) {
```
defunctzombie, line 175: Can this isEqual be replaced with `===` check or does this check map key/values? Our typical pattern is to assume object reference equality (like react hooks).

foxymiles: I was trying to avoid unnecessary renders but it looks like you're right and reference equality also works here. I'll change it.

packages/studio/index.d.ts
foxymiles: I decided to go with a `Map` here since it avoids another conversion step.

PR #2155, opened 2021-11-15
Hide "Desktop app" link in Help sidebar in the desktop app

esthersweon:
**User-Facing Changes**
- N/A

**Description**
- Hide "Desktop app" link in Help sidebar in the desktop app

### DIFF
--- a/packages/studio-base/src/components/HelpSidebar/index.tsx
+++ b/packages/studio-base/src/components/HelpSidebar/index.tsx
@@ -18,8 +18,11 @@ const productLinks = [
   { text: "Foxglove Data Platform", url: "https://foxglove.dev/data-platform" },
 ];
 
+const userAgent = navigator.userAgent.toLowerCase();
+const isDesktopApp = userAgent.includes(" electron/");
+
 const resourceLinks = [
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/components/HelpSidebar/index.tsx
```diff
@@ -18,8 +18,11 @@ const productLinks = [
   { text: "Foxglove Data Platform", url: "https://foxglove.dev/data-platform" },
 ];
 
+const userAgent = navigator.userAgent.toLowerCase();
+const isDesktopApp = userAgent.includes(" electron/");
```
defunctzombie, line 22: Do we do this anywhere else? If we do maybe we should make this a util shared file?

jtbandes: Let's use the presence of the `desktopBridge`, or maybe `NativeWindowContext`, as an indicator of the desktop app rather than userAgent.

esthersweon: Resolved in https://github.com/foxglove/studio/pull/2162

Issue #2025, opened 2021-10-22
Meta: Improved URL parameters and link sharing


amacneil:
We want to improve URL support in Studio (GET parameters / deep links), so that users can bookmark or share the current URL and persist their current layout/connection/timestamp. We also want to make it easier for users to open a web link in their desktop app.

See: [Requirements doc](https://docs.google.com/document/d/1jULICvQaZbJM8Et4ylFCE7GIFwedq7YkB4NhvnpjW0k/edit)

- [x] https://github.com/foxglove/studio/issues/1868
- [x] https://github.com/foxglove/studio/issues/1911
- [x] https://github.com/foxglove/studio/issues/1960

itsderek23:
This would be great. I'd love to drop a link into Slack that opens a FoxGlove Studio window w/the same panels, bag (loaded via the data platform), and at the same point in time.

PR #2158, opened 2021-11-16
Add <link rel=canonical /> tag with canonical URL to web build

esthersweon:
**User-Facing Changes**
- None

**Description**
- In response to ahrefs.com site audit:
   Screen Shot 2021-11-15 at 6 23 50 PM
- Added <link rel=canonical /> tag with canonical URL to web build

### DIFF
--- a/web/webpack.config.ts
+++ b/web/webpack.config.ts
@@ -131,6 +131,7 @@ const mainConfig = (env: unknown, argv: WebpackArgv): Configuration => {
       <link rel="apple-touch-icon" sizes="180x180" href="/apple-touch-icon.png" />
       <link rel="icon" type="image/png" sizes="32x32" href="/favicon-32x32.png" />
       <link rel="icon" type="image/png" sizes="16x16" href="/favicon-16x16.png" />
+      <link rel="canonical" href="https://studio.foxglove.dev"/>
       <title>Foxglove Studio</title>
     </head>
     <script>


jtbandes:
What are the duplicate urls that it detected?

amacneil:
@esthersweon this doesn't seem like a good idea. When we add URL params support (https://github.com/foxglove/studio/issues/2025) this is going to be incorrect.

Per Jacob's question, what are the duplicate URLs? There shouldn't be any.

jtbandes:
> When we add URL params support (#2025) this is going to be incorrect.

Are you sure? I don't think we would really want search engines indexing different sets of url params as different resources, would we?

amacneil:
Hmm. Not sure - I don't think we should care about SEO for studio.foxglove.dev anyway. People might link to public datasets, but I don't know that we care if google indexes them or not.

The other thing to consider is that people might host studio on different domains, so it seems a bit weird hard coding our fqdn here? Maybe it's ok.

esthersweon:
@jtbandes @amacneil - Hm it seems like it's seeing the URL with and without query params as different URLs (i.e. `studio.foxglove.dev` and `studio.foxglove.dev/?demo`).

Screen Shot 2021-11-16 at 4 55 54 PM

According to [this resource](https://webmasters.stackexchange.com/questions/114622/should-query-strings-be-included-or-removed-from-the-canonical-tag), if different query params give you different content, their canonical URLs should include the query strings.

amacneil:
I see. I don't know whether we want to treat those as the same "content" then.

They have different content, even if a web crawler might not think so.

I don't know that this really matters either way, unless we start seeing search engines using the wrong URL as the "default".

esthersweon:
@amacneil - that makes sense; will revert this PR.

PR #2161, opened 2021-11-16
Add enter/exit fullscreen icon to panel toolbar

defunctzombie:
**User-Facing Changes**
Add an enter/exit fullscreen icon to the panel toolbar.

Remove the "exit fullscreen" floating button. This button covers up parts of the panel that are not otherwise covered when not in fullscreen mode.

Enter fullscreen icon:
Screen Shot 2021-11-16 at 10 19 00 AM

Exit fullscreen icon:
Screen Shot 2021-11-16 at 10 19 07 AM

**Description**
We changed _something_ with the styles which cased the "Exit fullscreen" floating button to move from its prior home in the lower right corner of panels to somewhere higher up in the panel.

Prior to this change when in fullscreen:
Screen Shot 2021-11-16 at 10 16 55 AM

This covers up other panel elements. Since we already have a component that covers up other panel elements (the panel toolbar), this change leverages that component for the enter/exit fullscreen function. This also has the benefit of making fullscreen easier to find.

The existing shortcut functions remain in place.

### DIFF
--- a/packages/studio-base/src/components/MockPanelContextProvider.tsx
+++ b/packages/studio-base/src/components/MockPanelContextProvider.tsx
@@ -33,6 +33,10 @@ const DEFAULT_MOCK_PANEL_CONTEXT: PanelContextType<PanelConfig> = {
   enterFullscreen: () => {
     // no-op
   },
+  exitFullscreen: () => {
+    // no-op
+  },
+  isFullscreen: false,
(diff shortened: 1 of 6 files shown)


defunctzombie:
@esthersweon I tagged this with `docs` but I'm not actually sure if there are explicit docs around this so feel free to remove that.

PR #2171, opened 2021-11-17
Enhance ability of studio to load state from external URLs.

### LINKED ISSUE #1868: Store and load current connection in URL

### LINKED ISSUE #1911: Store and load current layout ID in URL
I should be able to bookmark a URL and then open it later to see the same view.

There are many possible complexities we could introduce to this feature, but this ticket is an MVP that we can ship and iterate in the future.

Requirements:
- Current layout ID should be stored in the URL query params like `studio.foxglove.dev/?layout=123`
- Opening that URL should load the requested layout
- If the current user does not have permission to load that layout, we should load their default layout instead, and show a toast saying "layout not found".
- If I'm not logged in and layouts are only in localstorage, then we only support loading a layout ID that exists in your localstorage. Don't worry about loading local layouts across devices for v1.

### LINKED ISSUE #1960: Deep link landing page
Similar to Slack, Zoom, MS Teams, etc, we should have an http-based landing page to give users the option of opening a link in the desktop or browser apps.

URL would look something like `https://studio.foxglove.dev/layout/<layoutId>?connection=<playerType>&s=<connectionString>&t=<timestamp>`.

On this page you would see something like this:

image

### LINKED ISSUE #2025: Meta: Improved URL parameters and link sharing

foxymiles:
**User-Facing Changes**
This expands the list of parameters that can be specified in foxglove urls to bring up studio in a predefined state. It also gives the user a choice of opening a given URL in the web or desktop version of the app and to optionally remember that choice for future sessions.

**Description**
The goal here is threefold:
1. Encode the current state of the session in an always-sharable URL.
2. Reconstruct the state of the session given a URL.
3. Allow a user to select in which environment they want to open a URL.

Most of this logic is encapsulated in the `useStateLocationSynchronization` hook used in `Workspace.tsx`. It handles updating the URL as the session state changes and also restoring a session from a given URL.

Currently we support `http:` and `foxglove:` protocol URL formats. The `foxglove:` URLs are used to open the desktop app. The app state is entirely encoded in the URL search params; the host and path are not used to encode state. Datasource-specific parameters are prefixed with `ds.`.

Most URLs are a simple type/URL pair. This is the case for ros1, ros2, remote bag files and rosbridge. These take this form:

`{prefix}/?ds={type}&ds.url={url}&layoutId={layoutId}&time={time}`

Where time is an optional, RFC3339 encoded value and layoutId is an optional layoutId.

For example:

https://studio.foxglove.dev/?layoutId=7c6c0f02-b298-4ee4-9d1b-c37659a0659b&ds=ros1-remote-bagfile&ds.url=https%3A%2F%2Fstorage.googleapis.com%2Ffoxglove-public-assets%2Fdemo.bag&time=2017-03-22T02%3A26%3A20.202843111Z

Data platform URLs have several more parameters:

`{prefix}/?ds=foxglove-data-platform&ds.deviceId={deviceId}&ds.start={start}&ds.end={end}&layoutId={layoutId}&time={time}`

Where time is an optional RFC3339 time values and layoutId is an optional layoutId.

For example:

`foxglove://open/?ds=foxglove-data-platform&ds.deviceId=dev_DdVETMbZ9ikVxzdv&ds.start=2021-05-03T18%3A57%3A58.486Z&ds.end=2021-05-03T18%3A58%3A01.501Z&time=2021-05-03T18%3A57%3A59.993Z`

Some other changes were necessary to make this work:
1. Some changes to the environment config to make it possible to test signed in sessions in local dev.
2. Wrapping the `App` component for web sessions to present a web/desktop environment choice.
3. Changes to the `PlayerManager` and `CurrentLayoutProvider` to let the URL take precedence over localStorage in specifying session state.
4. Changes to the various players to include `urlState` in their `PlayerState` to pass through to the `useStateLocationSynchronization` hook.

__Note__: Because the browser doesn't give us the full path to local files it doesn't seem possible to encode their path in a bookmarkable/sharable URL.

Fixes #2025 
Fixes #1911 
Fixes #1868 
Fixes #1960

### DIFF
--- a/desktop/webpack.renderer.config.ts
+++ b/desktop/webpack.renderer.config.ts
@@ -10,6 +10,7 @@ import path from "path";
 import { Configuration, EnvironmentPlugin, WebpackPluginInstance } from "webpack";
 
 import type { WebpackArgv } from "@foxglove/studio-base/WebpackArgv";
+import { buildEnvironmentDefaults } from "@foxglove/studio-base/environment";
 import { makeConfig } from "@foxglove/studio-base/webpack";
 
 import packageJson from "../package.json";
@@ -79,17 +80,7 @@ export default (env: unknown, argv: WebpackArgv): Configuration => {
     plugins: [
       ...plugins,
       ...(appWebpackConfig.plugins ?? []),
-      new EnvironmentPlugin({
-        SENTRY_DSN: process.env.SENTRY_DSN ?? null, // eslint-disable-line no-restricted-syntax
-        SENTRY_PROJECT: process.env.SENTRY_PROJECT ?? null, // eslint-disable-line no-restricted-syntax
-        AMPLITUDE_API_KEY: process.env.AMPLITUDE_API_KEY ?? null, // eslint-disable-line no-restricted-syntax
-        SIGNUP_API_URL: "https://foxglove.dev/api/signup",
-        SLACK_INVITE_URL: "https://foxglove.dev/join-slack",
-        OAUTH_CLIENT_ID: process.env.OAUTH_CLIENT_ID ?? "oSJGEAQm16LNF09FSVTMYJO5aArQzq8o",
-        FOXGLOVE_API_URL: process.env.FOXGLOVE_API_URL ?? "https://api.foxglove.dev",
-        FOXGLOVE_ACCOUNT_DASHBOARD_URL:
-          process.env.FOXGLOVE_ACCOUNT_DASHBOARD_URL ?? "https://console.foxglove.dev/profile",
-      }),
+      new EnvironmentPlugin(buildEnvironmentDefaults(argv.env?.FOXGLOVE_BACKEND)),
       new HtmlWebpackPlugin({
         templateContent: `
   <!doctype html>
--- a/packages/studio-base/environment.ts
+++ b/packages/studio-base/environment.ts
@@ -1,0 +1,43 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+function serverURLsForBackend(backend: string): { api: string; console: string } {
+  if (backend === "local") {
+    return {
+      api: "http://localhost:3000/api",
+      console: "http://localhost:3000",
+    };
+  }
+
+  if (backend === "dev") {
+    return {
+      api: "https://dev-api.foxglove.dev",
+      console: "https://console-dev.foxglove.dev",
+    };
+  }
+
+  return {
+    api: process.env.FOXGLOVE_API_URL ?? "https://api.foxglove.dev",
+    console: process.env.FOXGLOVE_CONSOLE_URL ?? "https://console.foxglove.dev",
+  };
+}
+
+export function buildEnvironmentDefaults(
+  backend: string = "production",
+  // eslint-disable-next-line no-restricted-syntax
+): Record<string, string | null> {
+  const serverURLs = serverURLsForBackend(backend);
+  return {
+    AMPLITUDE_API_KEY: process.env.AMPLITUDE_API_KEY ?? null, // eslint-disable-line no-restricted-syntax
+    FOXGLOVE_API_URL: serverURLs.api,
+    FOXGLOVE_ACCOUNT_DASHBOARD_URL:
+      process.env.FOXGLOVE_ACCOUNT_DASHBOARD_URL ?? serverURLs.console + "/dashboard",
+    FOXGLOVE_CONSOLE_URL: serverURLs.console,
+    OAUTH_CLIENT_ID: process.env.OAUTH_CLIENT_ID ?? "oSJGEAQm16LNF09FSVTMYJO5aArQzq8o",
+    SENTRY_DSN: process.env.SENTRY_DSN ?? null, // eslint-disable-line no-restricted-syntax
+    SENTRY_PROJECT: process.env.SENTRY_PROJECT ?? null, // eslint-disable-line no-restricted-syntax
+    SIGNUP_API_URL: "https://foxglove.dev/api/signup",
+    SLACK_INVITE_URL: "https://foxglove.dev/join-slack",
+  };
+}
--- a/packages/studio-base/environment.ts
+++ b/packages/studio-base/environment.ts
@@ -12,7 +12,7 @@ function serverURLsForBackend(backend: string): { api: string; console: string }
 
   if (backend === "dev") {
     return {
-      api: "https://dev-api.foxglove.dev",
+      api: "https://api-dev.foxglove.dev",
       console: "https://console-dev.foxglove.dev",
     };
   }
--- a/packages/studio-base/src/App.tsx
+++ b/packages/studio-base/src/App.tsx
@@ -6,6 +6,7 @@ import { Suspense, useState } from "react";
 import { DndProvider } from "react-dnd";
 import { HTML5Backend } from "react-dnd-html5-backend";
 
+import { AppSetting } from "@foxglove/studio-base/AppSetting";
 import Workspace from "@foxglove/studio-base/Workspace";
 import DocumentTitleAdapter from "@foxglove/studio-base/components/DocumentTitleAdapter";
 import MultiProvider from "@foxglove/studio-base/components/MultiProvider";
@@ -18,13 +19,18 @@ import { HoverValueProvider } from "@foxglove/studio-base/context/HoverValueCont
 import ModalHost from "@foxglove/studio-base/context/ModalHost";
 import { IDataSourceFactory } from "@foxglove/studio-base/context/PlayerSelectionContext";
 import { UserNodeStateProvider } from "@foxglove/studio-base/context/UserNodeStateContext";
+import { useAppConfigurationValue } from "@foxglove/studio-base/hooks";
+import { useSessionStorageValue } from "@foxglove/studio-base/hooks/useSessionStorageValue";
 import CurrentLayoutProvider from "@foxglove/studio-base/providers/CurrentLayoutProvider";
 import ExtensionMarketplaceProvider from "@foxglove/studio-base/providers/ExtensionMarketplaceProvider";
 import ExtensionRegistryProvider from "@foxglove/studio-base/providers/ExtensionRegistryProvider";
 import HelpInfoProvider from "@foxglove/studio-base/providers/HelpInfoProvider";
 import LayoutManagerProvider from "@foxglove/studio-base/providers/LayoutManagerProvider";
 import PanelCatalogProvider from "@foxglove/studio-base/providers/PanelCatalogProvider";
+import { LaunchPreferenceScreen } from "@foxglove/studio-base/screens/LaunchPreferenceScreen";
+import { LaunchingInDesktopScreen } from "@foxglove/studio-base/screens/LaunchingInDesktopScreen";
 import URDFAssetLoader from "@foxglove/studio-base/services/URDFAssetLoader";
+import isDesktopApp from "@foxglove/studio-base/util/isDesktopApp";
 
 type AppProps = {
   /**
@@ -37,7 +43,7 @@ type AppProps = {
   deepLinks?: string[];
 };
 
-export default function App(props: AppProps): JSX.Element {
+function AppContent(props: AppProps): JSX.Element {
   const [assetLoaders] = useState(() => [new URDFAssetLoader()]);
 
   const providers = [
@@ -74,4 +80,28 @@ export default function App(props: AppProps): JSX.Element {
       </DndProvider>
     </MultiProvider>
   );
+}
+
+export default function App(props: AppProps): JSX.Element {
+  const isDesktop = isDesktopApp();
+  const [globalLaunchPreference] = useAppConfigurationValue<string>(AppSetting.LAUNCH_PREFERENCE);
+  const [sessionLaunchPreference] = useSessionStorageValue(AppSetting.LAUNCH_PREFERENCE);
+
+  // Session preferences take priority over global preferences.
+  const activePreference = sessionLaunchPreference ?? globalLaunchPreference;
+
+  if (isDesktop) {
+    return <AppContent {...props} />;
+  } else {
+    const url = new URL(window.location.href);
+    const hasParams = Array.from(url.searchParams.entries()).length > 0;
+    // Ask the user in which environment they want to open this session.
+    if (activePreference == undefined && hasParams) {
+      return <LaunchPreferenceScreen />;
+    } else if (activePreference === "desktop" && hasParams) {
+      return <LaunchingInDesktopScreen />;
+    } else {
+      return <AppContent {...props} />;
+    }
+  }
 }
--- a/packages/studio-base/src/Workspace.tsx
+++ b/packages/studio-base/src/Workspace.tsx
@@ -24,8 +24,6 @@ import {
 import { useToasts } from "react-toast-notifications";
 import { useMount, useMountedState } from "react-use";
 
-import Log from "@foxglove/log";
-import { fromRFC3339String } from "@foxglove/rostime";
 import AccountSettings from "@foxglove/studio-base/components/AccountSettingsSidebar/AccountSettings";
 import ConnectionList from "@foxglove/studio-base/components/ConnectionList";
 import connectionHelpContent from "@foxglove/studio-base/components/ConnectionList/index.help.md";
@@ -53,6 +51,7 @@ import Preferences from "@foxglove/studio-base/components/Preferences";
 import RemountOnValueChange from "@foxglove/studio-base/components/RemountOnValueChange";
 import Sidebar, { SidebarItem } from "@foxglove/studio-base/components/Sidebar";
 import { SidebarContent } from "@foxglove/studio-base/components/SidebarContent";
+import { URLStateSyncAdapter } from "@foxglove/studio-base/components/URLStateSyncAdapter";
 import { useAppConfiguration } from "@foxglove/studio-base/context/AppConfigurationContext";
 import { useAssets } from "@foxglove/studio-base/context/AssetsContext";
 import ConsoleApiContext from "@foxglove/studio-base/context/ConsoleApiContext";
@@ -73,8 +72,6 @@ import useElectronFilesToOpen from "@foxglove/studio-base/hooks/useElectronFiles
 import useNativeAppMenuEvent from "@foxglove/studio-base/hooks/useNativeAppMenuEvent";
 import welcomeLayout from "@foxglove/studio-base/layouts/welcomeLayout";
 import { PlayerPresence } from "@foxglove/studio-base/players/types";
-
-const log = Log.getLogger(__filename);
 
 const useStyles = makeStyles({
   container: {
@@ -377,55 +374,6 @@ export default function Workspace(props: WorkspaceProps): JSX.Element {
     }
   }, [filesToOpen, openFiles]);
 
-  useEffect(() => {
-    const firstLink = props.deepLinks?.[0];
-    if (firstLink == undefined) {
-      return;
-    }
-
-    try {
-      const url = new URL(firstLink);
-      // only support the open command
-
-      // Test if the pathname matches //open or //open/
-      if (!/\/\/open\/?/.test(url.pathname)) {
-        return;
-      }
-
-      const type = url.searchParams.get("type");
-      if (type === "rosbag") {
-        const bagUrl = url.searchParams.get("url");
-        if (!bagUrl) {
-          log.warn(`Missing rosbag url param in ${url}`);
-          return;
-        }
-        selectSource("ros1-remote-bagfile", { url: bagUrl });
-      } else if (type === "foxglove-data-platform") {
-        const start = url.searchParams.get("start") ?? "";
-        const end = url.searchParams.get("end") ?? "";
-        const seekTo = url.searchParams.get("seekTo") ?? undefined;
-        const deviceId = url.searchParams.get("deviceId");
-        if (!deviceId) {
-          log.warn(`Missing deviceId param in ${url}`);
-          return;
-        }
-        if (
-          !fromRFC3339String(start) ||
-          !fromRFC3339String(end) ||
-          (seekTo && !fromRFC3339String(seekTo))
-        ) {
-          log.warn(`Missing or invalid timestamp(s) in ${url}`);
-          return;
-        }
-        selectSource("foxglove-data-platform", { start, end, seekTo, deviceId });
-      } else {
-        log.warn(`Unknown deep link type ${url}`);
-      }
-    } catch (err) {
-      log.error(err);
-    }
-  }, [props.deepLinks, selectSource]);
-
   const dropHandler = useCallback(
     ({ files }: { files: FileList }) => {
       void openFiles(files);
@@ -516,6 +464,7 @@ export default function Workspace(props: WorkspaceProps): JSX.Element {
           <div className={classes.dropzone}>Drop a file here</div>
         </DropOverlay>
       </DocumentDropListener>
+      <URLStateSyncAdapter deepLinks={props.deepLinks ?? []} />
       <div className={classes.container} ref={containerRef} tabIndex={0}>
         <Sidebar
           items={sidebarItems}
--- a/packages/studio-base/src/components/PlayerManager.tsx
+++ b/packages/studio-base/src/components/PlayerManager.tsx
@@ -45,6 +45,7 @@ import UserNodePlayer from "@foxglove/studio-base/players/UserNodePlayer";
 import { Player } from "@foxglove/studio-base/players/types";
 import { UserNodes } from "@foxglove/studio-base/types/panels";
 import Storage from "@foxglove/studio-base/util/Storage";
+import { windowHasValidURLState } from "@foxglove/studio-base/util/appURLState";
 
 const log = Logger.getLogger(__filename);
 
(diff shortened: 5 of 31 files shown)


packages/studio-base/src/players/types.ts
```diff
@@ -94,6 +94,8 @@ export type PlayerProblem = {
   tip?: string;
 };
 
+export type PlayerURLState = { type: "rosbag"; url: string };
```
jtbandes, line 97: will this eventually be arbitrary key/value data? For example, the Data Platform source doesn't have a url, but it does have deviceid/start/end, right?

foxymiles: My thinking here is to have defined data fields for known types and an extra catch-all type for everything else so it will be a union that includes a final more generic type.

packages/studio-base/src/players/types.ts
```diff
@@ -94,7 +96,9 @@ export type PlayerProblem = {
   tip?: string;
 };
 
-export type PlayerURLState = { type: "rosbag"; url: string };
+export type PlayerURLState =
+  | { type: "foxglove-data-platform"; options: FoxgloveDataPlatformOptions }
+  | { type: "rosbag"; url: string };
```
foxymiles, line 102: @jtbandes This is what I had in mind.

desktop/webpack.renderer.config.ts
```diff
@@ -79,17 +80,7 @@ export default (env: unknown, argv: WebpackArgv): Configuration => {
...
-        SENTRY_PROJECT: process.env.SENTRY_PROJECT ?? null, // eslint-disable-line no-restricted-syntax
-        AMPLITUDE_API_KEY: process.env.AMPLITUDE_API_KEY ?? null, // eslint-disable-line no-restricted-syntax
-        SIGNUP_API_URL: "https://foxglove.dev/api/signup",
-        SLACK_INVITE_URL: "https://foxglove.dev/join-slack",
-        OAUTH_CLIENT_ID: process.env.OAUTH_CLIENT_ID ?? "oSJGEAQm16LNF09FSVTMYJO5aArQzq8o",
-        FOXGLOVE_API_URL: process.env.FOXGLOVE_API_URL ?? "https://api.foxglove.dev",
-        FOXGLOVE_ACCOUNT_DASHBOARD_URL:
-          process.env.FOXGLOVE_ACCOUNT_DASHBOARD_URL ?? "https://console.foxglove.dev/profile",
-      }),
+      new EnvironmentPlugin(buildEnvironmentDefaults(argv.env?.FOXGLOVE_BACKEND)),
```
foxymiles, line 83: There was some duplicated logic between the web & desktop environments so I consolidated these into a helper function and added some new overrides for local dev modes.

packages/studio-base/src/util/appURLState.ts
```diff
@@ -0,0 +1,155 @@
...
+
+  if (
+    type === "ros1-remote-bagfile" ||
+    type === "rosbridge-websocket" ||
+    type === "ros1" ||
+    type === "ros2"
+  ) {
+    const resourceUrl = url.searchParams.get("url");
+    if (!resourceUrl) {
+      return Error(`Missing resource url param in ${url}`);
```
jtbandes, line 95: why not just throw?

packages/studio-base/src/hooks/useStateLocationSynchronization.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  // To avoid spamming the time field in the URL we only update it when playback is stopped.
+  const { isPlaying, currentTime } = activeData ?? {};
+  const activeTime = isPlaying === true ? undefined : currentTime;
+
+  // Serialize changes of current app state into address bar.
+  useEffect(() => {
+    if (isDesktopApp()) {
+      return;
+    }
```
jtbandes, line 64: what happens if we try to run this logic for the desktop app?

Should we exit the whole function earlier if we are in the desktop environment? (probably will run into hook lint errors)

packages/studio-base/src/hooks/useStateLocationSynchronization.ts
```diff
@@ -0,0 +1,121 @@
...
+      if (appUrlState.type === "foxglove-data-platform") {
+        selectSource(appUrlState.type, appUrlState.options);
+      } else if (
+        appUrlState.type === "ros1" ||
+        appUrlState.type === "ros2" ||
+        appUrlState.type === "ros1-remote-bagfile" ||
+        appUrlState.type === "rosbridge-websocket"
+      ) {
+        selectSource(appUrlState.type, appUrlState);
+      }
```
jtbandes, line 100: why the difference between data platform and other sources? I guess this is how things already work today and the player related APIs are not changing in this PR?

packages/studio-base/src/players/RandomAccessPlayer.ts
```diff
@@ -359,6 +359,12 @@ export default class RandomAccessPlayer implements Player {
             publishedTopics,
             parsedMessageDefinitionsByTopic: this._parsedMessageDefinitionsByTopic,
           },
+      urlState: this._label
+        ? {
+            type: "ros1-remote-bagfile",
```
jtbandes, line 364: Previously I think these type strings were only known to the app / player builders, not to player classes themselves (because, like here, one player class can be used with multiple types). It seems like this might be tangling things a bit if the players now also have knowledge of / dependence on the type strings.

At least it might be nice to move them into an enum so we don't risk typos in the future

cc @defunctzombie

packages/studio-base/src/screens/LaunchPreferenceScreen.tsx
```diff
@@ -0,0 +1,80 @@
...
+  const [globalPreference, setGlobalPreference] = useAppConfigurationValue<string | undefined>(
+    AppSetting.LAUNCH_PREFERENCE,
+  );
+  const [_, setSessionPreference] = useSessionStorageValue(AppSetting.LAUNCH_PREFERENCE);
+  const [rememberPreference, setRememberPreference] = useState(globalPreference != undefined);
+
+  const cleanWebURL = new URL(window.location.href);
+  cleanWebURL.searchParams.delete("launch");
+
+  async function launchInWeb() {
```
jtbandes, line 34: nit: useCallback

packages/studio-base/src/screens/LaunchPreferenceScreen.tsx
```diff
@@ -0,0 +1,80 @@
...
+export function LaunchPreferenceScreen(): ReactElement {
+  const styles = useStyles();
+  const [globalPreference, setGlobalPreference] = useAppConfigurationValue<string | undefined>(
+    AppSetting.LAUNCH_PREFERENCE,
+  );
+  const [_, setSessionPreference] = useSessionStorageValue(AppSetting.LAUNCH_PREFERENCE);
+  const [rememberPreference, setRememberPreference] = useState(globalPreference != undefined);
+
+  const cleanWebURL = new URL(window.location.href);
+  cleanWebURL.searchParams.delete("launch");
```
jtbandes, line 32: nit: useMemo

packages/studio-base/src/screens/LaunchingInDesktopScreen.tsx
```diff
@@ -0,0 +1,47 @@
...
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { Link, Stack, Text } from "@fluentui/react";
+import { ReactElement, useEffect } from "react";
+
+import { AppSetting } from "@foxglove/studio-base/AppSetting";
+import { useSessionStorageValue } from "@foxglove/studio-base/hooks/useSessionStorageValue";
+
+export function LaunchingInDesktopScreen(): ReactElement {
```
jtbandes, line 11: I think the component name is a bit misleading, it sounds like it would just display a loading screen, but it actually triggers the desktop app to open as a side effect.

packages/studio-base/src/screens/LaunchingInDesktopScreen.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+  return (
+    <Stack
+      horizontalAlign="center"
+      verticalAlign="center"
+      verticalFill
+      tokens={{ childrenGap: "1rem" }}
+    >
+      <Text variant="xxLarge">Launching Foxglove Studio</Text>
+      <Text>We&apos;ve directed you to the desktop app.</Text>
```
jtbandes, line 41: nit: I think you can just type `'` or preferably `’` directly in the source code.

packages/studio-base/src/util/appURLState.test.ts
```diff
@@ -0,0 +1,110 @@
...
+import isDesktopApp from "@foxglove/studio-base/util/isDesktopApp";
+
+jest.mock("@foxglove/studio-base/util/isDesktopApp", () => ({
+  __esModule: true,
+  default: jest.fn(),
+}));
+
+const mockIsDesktop = isDesktopApp as jest.MockedFunction<typeof isDesktopApp>;
+
+function runCommonTests(urlBuilder: () => URL) {
```
jtbandes, line 18: consider using `describe.each` or `it.each` to achieve this rather than a custom helper function?

packages/studio-base/src/Workspace.tsx
```diff
@@ -249,6 +246,10 @@ export default function Workspace(props: WorkspaceProps): JSX.Element {
     }
   }, []);
 
+  // Sync session state to URL and restore session state from deeplinks if any
+  // are present.
+  useStateLocationSynchronization(props.deepLinks ?? []);
```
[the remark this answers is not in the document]
foxymiles: Possibly. The existing logic was in Workspace so it seemed a bit less disruptive to leave it there but encapsulated in a hook. But if we're game for a more extensive PR then it might make sense in app.

packages/studio-base/src/components/PlayerManager.tsx
```diff
@@ -275,8 +276,13 @@ export default function PlayerManager(props: PropsWithChildren<PlayerManagerProp
     ],
   );
 
-  // restore the saved source on first mount
+  // Restore the saved source on first mount unless our url specifies a source.
   useLayoutEffect(() => {
+    // The URL encodes a valid session state. Defer to the URL state.
+    if (windowHasValidURLState()) {
+      return;
+    }
```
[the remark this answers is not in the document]
foxymiles: Yeah this does feel a bit hacky but also doesn't require a more extensive reworking of the state bootstrapping logic. Do we want to try to take something like that on as part of this work?

packages/studio-base/src/hooks/useSessionStorageValue.ts
```diff
@@ -0,0 +1,51 @@
...
+ * This provides a convenience wrapper around sessionStorage and triggers
+ * a react state change when values change.
+ *
+ * TODO: We should probably refactor LocalStorageAppConfigurationProvider to
+ * handle localStorage & sessionStorage and replace this.
+ *
+ * @param key sessionStorage key to manage.
+ * @returns [value, setValue] tuple for that key.
+ */
+export function useSessionStorageValue(
```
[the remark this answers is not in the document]
foxymiles: Yeah I needed something that would provide a notification. I think longer term it would be better to abstract our existing localStorage context code to work with both and get rid of this.

packages/studio-base/src/hooks/useSessionStorageValue.ts
```diff
@@ -0,0 +1,51 @@
...
+  const [value, updateValue] = useState<string | undefined>(
+    sessionStorage.getItem(key) ?? undefined,
+  );
+
+  function setValue(newValue: string | undefined) {
+    // Hack a manual event for now. Unfortunately the browser only fires "storage"
+    // events when triggered outside our current tab.
+    if (newValue) {
+      sessionStorage.setItem(key, newValue);
+      window.dispatchEvent(new StorageEvent("storage", { key, newValue }));
```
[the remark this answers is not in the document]
foxymiles: Where would we specify this?

packages/studio-base/src/hooks/useStateLocationSynchronization.ts
```diff
@@ -0,0 +1,121 @@
...
+ * This handles both updating the current tab's URL to reflect the session state
+ * and also restoring the session state on first load from a URL.
+ */
+export function useStateLocationSynchronization(deepLinks: string[]): void {
+  const urlState = useMessagePipeline(selectUrlState);
+  const layoutId = useCurrentLayoutSelector((layout) => layout.selectedLayout?.id);
+  const stableUrlState = useDeepMemo(urlState);
+  const { selectSource } = usePlayerSelection();
+  const { setSelectedLayoutId } = useCurrentLayoutActions();
+  const activeData = useMessagePipeline(selectActiveData);
```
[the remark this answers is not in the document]
foxymiles: Thanks good catch. I will change this.

packages/studio-base/src/util/appURLState.ts
foxymiles: Could just as easily do that too I guess. It did seem a bit heavy handed to throw on a parse error.

packages/studio-base/src/hooks/useStateLocationSynchronization.ts
foxymiles: Yeah I think the hooks will be unhappy if we exit earlier. URL syncing in the desktop would have to follow a different method like using the hash part of the url maybe. My expectation was that we'd need some kind of explicit "share url" button for the destop.

packages/studio-base/src/hooks/useStateLocationSynchronization.ts
foxymiles: Right now all the resources are just a type/url pair with the exception of data platform that uses a bunch of other params.

packages/studio-base/src/players/RandomAccessPlayer.ts
foxymiles: Yeah we could use a different set of strings here and map them to the internal types. I'm not sure which is better.

I think typescript is smart enough to catch typos here since it's a bounded union type.

packages/studio-base/src/screens/LaunchPreferenceScreen.tsx
foxymiles: If this re-renders it's not really a big deal right?

packages/studio-base/src/screens/LaunchingInDesktopScreen.tsx
foxymiles: We want the desktop app to load directly from here don't we since the user has already clicked on the desktop button on the previous screen?

packages/studio-base/src/screens/LaunchingInDesktopScreen.tsx
foxymiles: I get an "unescaped-entities" lint warning if I do that.

packages/studio-base/src/util/appURLState.test.ts
foxymiles: Good idea I'll take a look at that thanks

packages/studio-base/src/components/PlayerManager.tsx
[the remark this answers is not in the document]
foxymiles: I don't think they'd get out of sync though since this just checks on startup of a new session.

packages/studio-base/src/hooks/index.ts
```diff
@@ -2,4 +2,7 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
+// These are exported from here to avoid circular imports via studio-base/index.
```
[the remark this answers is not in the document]
foxymiles: Yeah this is left over from a previous PR. Maybe I can fix it in a followup? I'm reluctant to drag more touched files into this one.

packages/studio-base/src/util/appURLState.test.ts
foxymiles: Updated tests to use `describe.each`. Definitely cleaner.

packages/studio-base/src/hooks/useStateLocationSynchronization.ts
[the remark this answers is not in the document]
foxymiles: Actually we do need this to trigger on currentTime changes so things like scrubbing will update the URL. I think this will be ok if we only do this when the app is not playing right?

packages/studio-base/src/util/appURLState.ts
foxymiles: I changed this to throw.

packages/studio-base/src/components/PlayerManager.tsx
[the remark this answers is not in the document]
jtbandes: probably fine for now, it might not be worth it to make this PR bigger.

packages/studio-base/src/hooks/useSessionStorageValue.ts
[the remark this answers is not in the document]
jtbandes: Not sure if you can specify it in the constructor – I just noticed that storageArea is a property of StorageEvent. This came to mind because I was wondering if there's a way for event listeners to know if it was sessionStorage or localStorage that changed.

packages/studio-base/src/screens/LaunchingInDesktopScreen.tsx
jtbandes: yeah I think so, this was just a comment about naming – might call it OpenDesktopApp or something? I dunno

packages/studio-base/src/screens/LaunchingInDesktopScreen.tsx
jtbandes: 😐

packages/studio-base/src/components/Preferences.tsx
```diff
@@ -188,6 +189,37 @@ function TimeFormat(): React.ReactElement {
...
 
+function LaunchDefault(): React.ReactElement {
+  const [preference = "unknown", setPreference] = useAppConfigurationValue<string | undefined>(
+    AppSetting.LAUNCH_PREFERENCE,
+  );
+
+  const entries: Array<{ key: string; text: string }> = [
+    { key: "unknown", text: "Ask Me" },
+    { key: "web", text: "Web" },
+    { key: "desktop", text: "Desktop" },
```
jtbandes, line 200: just some ideas, feel free to take some or none of these suggestoins
```suggestion
    { key: "unknown", text: "Ask each time" },
    { key: "web", text: "Web app (studio.foxglove.dev)" },
    { key: "desktop", text: "Desktop app" },
```

packages/studio-base/src/hooks/useSessionStorageValue.ts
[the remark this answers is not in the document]
foxymiles: Yeah looks like you can pass it to the event constructor like this:

```
new StorageEvent("storage", { key, newValue, storageArea: sessionStorage }),
```
I'll add it.

jtbandes (approved):
Overall it seems like the structure and functionality of url handling are improved a lot with this PR. Might be worth a little more bikeshedding on the actual url params since we're going to end up fairly locked in to this going forward – for example, thinking about whether we want to use prefixes like `d.deviceId=...`

packages/studio-base/src/components/Preferences.tsx
foxymiles: Good suggestions. I'll use these.

amacneil:
Decisions on url parameter bikeshedding:

- `layoutId`
- `time`
- `ds` (data source)
- `ds.*` (data source parameters)

examples:
```
https://studio.foxglove.dev/?layoutId=123&time=1234&ds=fgdata&ds.device=123&ds.start=456&ds.end=789
https://studio.foxglove.dev/?layoutId=123&time=1234&ds=rosbridge&ds.url=<urlencoded>
https://studio.foxglove.dev/?layoutId=123&time=1234&ds=ros1&ds.url=<ROS_MASTER_URI>&ds.roshostname=<ROS_HOSTNAME>
```

packages/studio-base/src/App.tsx
```diff
@@ -75,3 +81,27 @@ export default function App(props: AppProps): JSX.Element {
...
+  const isDesktop = isDesktopApp();
+  const [globalLaunchPreference] = useAppConfigurationValue<string>(AppSetting.LAUNCH_PREFERENCE);
+  const [sessionLaunchPreference] = useSessionStorageValue(AppSetting.LAUNCH_PREFERENCE);
+
+  // Session preferences take priority over global preferences.
+  const activePreference = sessionLaunchPreference ?? globalLaunchPreference;
+
+  if (isDesktop) {
+    return <AppContent {...props} />;
+  } else {
```
defunctzombie, line 95: nit: else not needed since the first if returns

packages/studio-base/src/hooks/useInitialDeepLinkState.ts
```diff
@@ -0,0 +1,99 @@
...
+        appUrlState.type === "rosbridge-websocket"
+      ) {
+        selectSource(appUrlState.type, appUrlState);
+      }
+
+      if (appUrlState.layoutId != undefined) {
+        setSelectedLayoutId(appUrlState.layoutId);
+      }
+    } catch (err) {
+      log.error(err);
```
defunctzombie, line 86: Probably need to inform the user some error happened or allow the throw to happen so we get sentry errors.

packages/studio-base/src/hooks/useInitialDeepLinkState.ts
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    try {
+      if (appUrlState.type === "foxglove-data-platform") {
+        selectSource(appUrlState.type, appUrlState.options);
+      } else if (
+        appUrlState.type === "ros1" ||
+        appUrlState.type === "ros2" ||
+        appUrlState.type === "ros1-remote-bagfile" ||
+        appUrlState.type === "rosbridge-websocket"
```
defunctzombie, line 77: can this list be built from the available sources list from usePlayerSelection?

packages/studio-base/src/players/FoxgloveDataPlatformPlayer/index.ts
```diff
@@ -259,6 +259,14 @@ export default class FoxgloveDataPlatformPlayer implements Player {
...
       playerId: this._id,
       problems: this._problems,
+      urlState: {
+        type: "foxglove-data-platform",
+        options: {
+          start: toRFC3339String(this._start ?? ZERO_TIME),
+          end: toRFC3339String(this._end ?? ZERO_TIME),
+          deviceId: this._deviceId,
+        },
+      },
```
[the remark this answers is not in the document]
defunctzombie: Having the type here is a disconnect from the factory where the type also exists. Probably better not to have the type here? Or to have a static type on the class so we can ensure the two are the same. Seems like type is a thing that should be outside the player to set in the url.

packages/studio-base/src/players/types.ts
```diff
@@ -94,6 +95,13 @@ export type PlayerProblem = {
   tip?: string;
 };
 
+export type PlayerURLState =
```
defunctzombie, line 98: One thing I've tried to move to over time is away from hard-coded lists of various players and instead read them from the available sources lists. For example ros1 isn't a valid data source in the browser - or if selected should inform the user it is not supported.

packages/studio-base/environment.ts
```diff
@@ -0,0 +1,43 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
defunctzombie, line 1: maybe the content of this file should live in the webpack.ts file (should be in the same studio-base folder) since it is specific to the webpack configuration use

packages/studio-base/src/players/FoxgloveDataPlatformPlayer/index.ts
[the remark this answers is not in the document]
foxymiles: Is there a better place in the code where we're sure we have a valid state to encode? I'll see if I can eliminate the `type` duplication here.

packages/studio-base/src/hooks/useInitialDeepLinkState.ts
foxymiles: Actually this and the type definitions were an attempt to get some typechecking on the different schemes we were using for DP vs other URLs. Since we've decided to make that more uniform maybe that's not helpful anymore and I should just pass generic connection info to PlayerManager and let it decide if the params make sense.

packages/studio-base/environment.ts
foxymiles: I was getting some very weird build errors when I did that. It seemed like webpack.ts was being built with some different constraints.

defunctzombie:
I've updated the PR description with some additional `Fixes` entries for issues I believe this PR also covers.

packages/studio-base/src/hooks/useInitialDeepLinkState.ts
foxymiles: Looks like `selectSource` will already show the user an error if it can't load the source.

PR #2172, opened 2021-11-17
[experiment] bag2proto

defunctzombie:
**User-Facing Changes**
None

**Description**
Make an protobuf mcap file from bag file

### DIFF
--- a/packages/mcap/scripts/bag2proto.ts
+++ b/packages/mcap/scripts/bag2proto.ts
@@ -1,0 +1,112 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+// convert a ROS1 .bag file to an mcap proto file
+
+import { program } from "commander";
+import { parse } from "protobufjs";
+
+import { Bag } from "@foxglove/rosbag";
+import { FileReader } from "@foxglove/rosbag/node";
+
+import { McapWriter, ChannelInfo, Message } from "../src";
+
+const protoSrc = `
+syntax = "proto3";
+
+package foxglove;
+
+// https://foxglove.dev/docs/datatypes/LocationFix
+message LocationFix {
+    double latitude = 1;
+    double longitude = 2;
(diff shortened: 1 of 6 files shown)


packages/mcap/scripts/bag2proto.ts
```diff
@@ -0,0 +1,112 @@
...
+    if (connection.type === "sensor_msgs/NavSatFix") {
+      navSatTopics.push(connection.topic);
+
+      const channelInfo: ChannelInfo = {
+        type: "ChannelInfo",
+        id: topicToChannelInfoMap.size,
+        topic: "/gps",
+        encoding: "protobuf",
+        schemaName: "foxglove.LocationFix",
+        schema: protoSrc,
```
jtbandes, line 54: Based on discussion with @amacneil and @wkalt we believe that it is reasonable for `encoding: "protobuf"` to imply the schema is a binary FileDescriptorSet (for mcap and the websocket protocol). To make this work here you should `import descriptor from "protobufjs/ext/descriptor"` and then you should be able to do something like:

```ts
base64.encode(descriptor.FileDescriptorSet.encode(MsgRoot.root.toDescriptor("proto3")).finish())
```

https://github.com/protobufjs/protobuf.js/blob/master/ext/descriptor/README.md

You will also want to move this PR over to the new [foxglove/mcap repo](https://github.com/foxglove/mcap).

packages/mcap/scripts/bag2proto.ts
```diff
@@ -0,0 +1,112 @@
...
+    const protoMsg = LocationFix.create({
+      latitude: rosMsg.latitude,
+      longitude: rosMsg.longitude,
+      altitude: rosMsg.altitude,
+      covariance: Array.from(rosMsg.position_covariance),
+    });
+
+    const protoMsgBuffer = LocationFix.encode(protoMsg).finish();
+
+    const timestamp = BigInt(result.timestamp.sec) * 1000000000n + BigInt(result.timestamp.nsec);
```
jtbandes, line 87: rostime has a toNanoSec util

defunctzombie:
Moved to mcap repo https://github.com/foxglove/mcap/pull/3

Issue #2170, opened 2021-11-17
Panel help content is rendering at top of Panel Settings


2metres:
**Description**
<img width="1215" alt="Screen Shot 2021-11-17 at 10 59 31 pm" src="https://user-images.githubusercontent.com/924528/142196721-8d9d7a0f-cc47-434e-bebd-1aea56f31330.png">

**Steps To Reproduce**
Click on 3D panel, Open Panel Settings

PR #2176, opened 2021-11-17
Fix: Do not display help content in the Panel Settings sidebar tab

esthersweon:
**User-Facing Changes**
- Fix: Do not display help content in the Panel Settings sidebar tab

Resolves https://github.com/foxglove/studio/issues/2170

### DIFF
--- a/packages/studio-base/src/components/PanelSettings/index.tsx
+++ b/packages/studio-base/src/components/PanelSettings/index.tsx
@@ -126,7 +126,7 @@ export default function PanelSettings(): JSX.Element {
     <SidebarContent title={`${panelInfo.title} panel settings`}>
       {shareModal}
       <Stack tokens={{ childrenGap: theme.spacing.m }}>
-        {panelInfo?.help ?? (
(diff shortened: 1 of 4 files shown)


packages/studio-base/src/components/PanelSettings/index.tsx
```diff
@@ -126,7 +135,7 @@ export default function PanelSettings(): JSX.Element {
     <SidebarContent title={`${panelInfo.title} panel settings`}>
       {shareModal}
       <Stack tokens={{ childrenGap: theme.spacing.m }}>
-        {panelInfo?.help ?? (
+        {panelInfo?.help != undefined ? (
```
defunctzombie, line 138: Could this be `panelInfo?.help != undefined &&` - then you can drop the `: undefined` in the ternary below. It seems the bug was using `??` rather than `&&` ?

defunctzombie: I made the change on this branch.

Issue #2159, opened 2021-11-16
3D Panel topic settings modal gets stuck to mouse cursor when moving it


2metres:
**Description**

https://user-images.githubusercontent.com/924528/141916419-2a1f7eea-be34-4cd9-a097-ceb16ff170b9.mov

- Version: Version 0.22.0-dev (darwin)
- OS: macOS 12.0 (21A344)

**Steps To Reproduce**
From demo.bag in the 3D panel open velodyne topic settings, drag window so that edge of modal is out of frame

jtbandes:
Missing an overflow:hidden somewhere?

2metres:
I think my recent change <https://github.com/foxglove/studio/pull/2128>
that removed the RenderToBody and thus made the Modal movable

>

jtbandes:
> and thus made the Modal movable

No, I made it draggable in https://github.com/foxglove/studio/pull/2054

PR #2178, opened 2021-11-17
Remove draggability from topic settings modal

jtbandes:
**User-Facing Changes**
Insignificant

**Description**
Draggable modals were causing problems with page layout when the modal went partly outside the page bounds.


defunctzombie:
Does this fix: https://github.com/foxglove/studio/issues/2159 ?

PR #2180, opened 2021-11-18
Implement MUI theme and provider

2metres:
**User-Facing Changes**
None, hopefully 🤞 

**Description**
Add mui/v5 theme and theme provider to start building off of


jtbandes (approved):
Needs to be merged/rebased, but looks like this was successfully implemented as a no-op so no harm in merging 👍

Issue #2169, opened 2021-11-17
Out of memory when playing back with '--clock'


zebrapants:
**Description**
I started playback of a set of bag files with '--clock' parameter set, then started Foxglove Studio with two image views, and after a short while, I got the message "Error: Failed to execute 'postMessage' on 'Worker': Data cannot be cloned, out of memory" in both image views. At this point, the memory consumption of Foxglove Studio has slowly climbed (not jumped) from 0.8 % to around 22 % of my entire system.

It appears the memory consumption also steadily increases with no image views, just not as fast.

The reluctance to release the memory only occurs when I use the '--clock' parameter.

The error log shows:
Error: failed to allocate 9463030 bytes to concatenate 157 chunks
    at Object.e [as concatData] (file:///opt/Foxglove%20Studio/resources/app.asar/renderer/main.js:6:93494)
    at h.addData (file:///opt/Foxglove%20Studio/resources/app.asar/renderer/main.js:6:79395)
    at bu._handleData (file:///opt/Foxglove%20Studio/resources/app.asar/renderer/main.js:6:87751)
    at bu.t.emit (file:///opt/Foxglove%20Studio/resources/app.asar/renderer/main.js:253:71633)
    at file:///opt/Foxglove%20Studio/resources/app.asar/renderer/main.js:3691:7379
    at MessagePort.x.onmessage (file:///opt/Foxglove%20Studio/resources/app.asar/renderer/main.js:3691:7625)

- Version: Foxglove Studio 0.22.0
- OS: Ubuntu 20.04
- ROS Noetic
- 32 GB memory

**Steps To Reproduce**

1. Start roscore
2. Start Foxglove Studio (connect to localhost, add panel on image topic)
3. Rosbag play a bunch of bags with '--clock'
4. Watch the memory consumption increase

![2021-11-17_11-24](https://user-images.githubusercontent.com/15982829/142184573-c6db052d-352e-42f4-a45a-6686003b356e.png)

**Expected Behavior**
- Show image and not error message
- Stable memory consumption

**Actual Behavior**
- Error message
- Monotonically increasing memory consumption

defunctzombie:
@zebrapants To aid with reproducing - does the error happen on the first play through of the bag file or subsequent plays?

zebrapants:
@defunctzombie Hi, thanks for picking this up. 
[...]

PR #2185, opened 2021-11-19
Clear incorrect messages from out-of-order messages after warning

defunctzombie:
**User-Facing Changes**
When replaying bag data with the `--clock` flag, Studio would continue to accumulate messages in memory until eventually crashing. This change fixes the memory accumulation behavior so the app does not crash.

**Description**
This change converts warnOnOutOfSyncMessages to a class so we can instantiate a new instance when players change. This resets message order tracking when changing players.
    
The new class fixes a memory issue with keeping all messages over the drift threshold in memory. The timeout which prints the log warning failed to clear the existing incorrect message array _and_ failed to clear the warningTimeout variable. The last bug prevented any new warning timeout from triggering so the array could be cleared.
    
Fixes: #2169


packages/studio-base/src/components/MessagePipeline/MessageOrderTracker.ts
```diff
@@ -0,0 +1,123 @@
...
+
+        if (currentTimeDrift > DRIFT_THRESHOLD_SEC) {
+          this.incorrectMessages.push(message);
+          if (!this.warningTimeout) {
+            this.warningTimeout = setTimeout(() => {
+              // timeout has fired, we need to clear so a new timeout registers if there are more messages
+              this.warningTimeout = undefined;
+              // reset incorrect message queue before posting warning so we never keep incorrectMessages around
+              const tempMessages = this.incorrectMessages;
+              this.incorrectMessages = [];
```
defunctzombie, line 83: These three lines (above) were the source of the unbounded memory growth. The incorrectMessages would forever get new messages, and nothing cleared the warningTimeout so it had no opportunity to run again as new messages were added.

defunctzombie:
@jtbandes Since I'll be out next week feel free to merge this after review so we can get it in the next release.

jtbandes (approved):

PR #2188, opened 2021-11-19
Remove redundant model selection from 3d panel pose editor.

### LINKED ISSUE #2184: Unselectable "Rendered Car" option in 3d panel Pose topic settings modal
**Description**
Load a bag with a pose, for example https://github.com/foxglove/studio/files/7325594/poses.bag.zip

Go to the pose topic settings. 

image

- Version: web
- OS: any

**Actual Behavior**
Unnecessary radio options show up, with a now-defunct `Rendered Car` option with no radio icon, and `Arrow` which is the default.

Since we only support Arrow now, we should just remove the radio options entirely.

foxymiles:
**User-Facing Changes**
This removes the extraneous model selection fields from the 3d panel pose editor dialog.

**Description**
Since we only support the arrow in this case we can remove the model selection radio & label.

Fixes #2184

### DIFF
--- a/packages/studio-base/src/panels/ThreeDimensionalViz/TopicSettingsEditor/PoseSettingsEditor.tsx
+++ b/packages/studio-base/src/panels/ThreeDimensionalViz/TopicSettingsEditor/PoseSettingsEditor.tsx
@@ -13,7 +13,6 @@
 
 import ColorPicker from "@foxglove/studio-base/components/ColorPicker";
 import Flex from "@foxglove/studio-base/components/Flex";
-import { LegacyInput } from "@foxglove/studio-base/components/LegacyStyledComponents";
 import { Color, PoseStamped } from "@foxglove/studio-base/types/Messages";
 import { colors } from "@foxglove/studio-base/util/sharedStyleConstants";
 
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/panels/ThreeDimensionalViz/TopicSettingsEditor/PoseSettingsEditor.tsx
```diff
@@ -37,66 +35,57 @@ export default function PoseSettingsEditor(
   const { message, settings, onFieldChange, onSettingsChange } = props;
 
   const settingsByCarType = React.useMemo(() => {
```
jtbandes, line 37: Probably could even get rid of this useMemo, as `settingsByCarType` wording isn't really relevant anymore.

jtbandes (approved):

packages/studio-base/src/panels/ThreeDimensionalViz/TopicSettingsEditor/PoseSettingsEditor.tsx
foxymiles: Good point. I've simplified this logic.

PR #2189, opened 2021-11-19
Support URDFs in the 3D panel

jhurliman:
**User-facing changes**

A new entry is shown in the 3D panel topic picker, "3D Model". ROS URDF files are currently supported. The URDF source URL will automatically be set if the data source supports parameters (live ROS 1 and ROS 2 connections) and the `/robot_description` parameter points at a valid URDF using either a `package://` or `http(s)://` URL.

Screen Shot 2021-11-19 at 10 01 29 AM

**Description**

ROS URDF files define links and joints, where each link can be thought of as a coordinate frame that optionally contains a visual representation of part of a robot, and each joint can be thought of as a transform defining a connection between parent and child coordinate frames, or links. When a URDF is parsed, each joint is turned into a transform internally. This is similar to receiving a single latched `/tf_static` message. During data playback, `/tf` messages can modify the initial transform poses set in the URDF to make the robot joints move.

Only a single URDF is currently supported per 3D panel. This is more of a UX limitation, where we don't currently have a flow for adding visualizations to a 3D scene other than observing incoming data during playback. Solving this is out of scope for the current PR but something that could be addressed in the future.

**TODO**

- [x] Monitor `/robot_description` parameter and set the URDF URL unless it is overridden in the "3D Model" settings
- [x] Sometimes the URDF model does not appear (even when the URDF transform axes are visible)
- [x] All of the transforms that only appear in the URDF are marked "Unavailable" in the topic picker and cannot be toggled to a hidden state, even though axes are rendered in the scene (that can't be hidden)

**Out of scope**

- Switch to a different tab then back to the current tab will lose the robot visualization and show the current frame as invalid until you change the camera to a different frame
- Support <material> color overrides for meshes (out of scope, requires changes to @foxglove/regl-worldview)
- Material support in COLLADA meshes
- xacro-parser fixes to support turtlebot
- xacro-parser fixes to support duckiebot

### DIFF
--- a/packages/studio-base/src/panels/ThreeDimensionalViz/UrdfBuilder.ts
+++ b/packages/studio-base/src/panels/ThreeDimensionalViz/UrdfBuilder.ts
@@ -85,12 +85,7 @@ export default class UrdfBuilder implements MarkerProvider {
   }
 
   setTransforms = (transforms: Transforms, rootTransformID: string | undefined): void => {
-    // Just checking transforms === this._transforms will sometimes return false
-    // even though they are the same object. The storage comparison consistently works
-    if (
-      transforms.storage === this._transforms?.storage &&
-      rootTransformID === this._rootTransformID
-    ) {
+    if (transforms === this._transforms && rootTransformID === this._rootTransformID) {
       return;
     }
     this._transforms = transforms;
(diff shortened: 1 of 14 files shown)


packages/studio-base/src/panels/ThreeDimensionalViz/UrdfBuilder.ts
```diff
@@ -0,0 +1,351 @@
...
+      const fetchUrl = rewritePackageUrl(url, { rosPackagePath });
+      const res = await fetch(fetchUrl);
+      return await res.text();
+    } catch (err) {
+      throw new Error(`Failed to fetch "${url}": ${err}`);
+    }
+  };
+}
+
+function eulerToQuaternion(rpy: Vector3): Quaternion {
```
jtbandes, line 320: gl-matrix?

jhurliman: I tried a few different iterations of gl-matrix quat and couldn't make it work. The rotations came out wrong each time, even with tweaks like flipping Y and Z. I haven't dug into the source code for gl-matrix yet to try and debug this.

PR #2187, opened 2021-11-19
Update account settings URL

bryfox:
This updates the URL used for Console's "Account Settings". These now live at `/organization`.

PR #2196, opened 2021-11-21
Fix account settings link

amacneil:
**User-Facing Changes**
Update "Account Settings" link for logged in users to correct URL.

cc @bryfox fyi


bryfox:
@amacneil sorry, I'd opened #2187 but not yet merged it. Currently the `/profile` page just has some UI preferences for Console. `/organization` has the team settings (though for non-admins, only a view of the team list). I thought the latter made more sense, but maybe we have other plans for `/profile`.

amacneil:
Oh right, sorry I didn’t see that. 

Yeah - I was also on the fence about user settings vs org settings. I don’t think it matters much. I went with user settings even though that page is largely empty today, only because it seemed like the most natural place to take someone after they click “Account settings” under their username, and not all users are going to have access to the org settings page.

Maybe in the future we will add more user settings (like updating your name) and the link will make more sense.

Issue #2144, opened 2021-11-11
mesh_resource marker with valid url results in 404 object in the scene


defunctzombie:
**Description**
Trying to play the referenced bag file results in a 404 displayed in the 3d panel and no other errors reported.

- Version: Studio as of 244b52f236d0f76602f756be7e138d837efc8808

**Steps To Reproduce**

Open the attached bag file. Add a 3d panel and enable the `/visualization_marker` topic.

[2021-11-12-00-43-31.bag.zip](https://github.com/foxglove/studio/files/7523090/2021-11-12-00-43-31.bag.zip)

**Expected Behavior**
Showing the model at the mesh_resource url or some error indicating why the model loading failed.

**Actual Behavior**
A 404 is shown in the 3d panel.

PR #2198, opened 2021-11-22
Improve error visibility for mesh markers; support capital .STL extension

jtbandes:
**User-Facing Changes**
Fixed loading mesh resource markers with capitalized file extensions such as `.STL`.

**Description**
Fixes https://github.com/foxglove/studio/issues/2144
- Pop up a toast notification when errors occur. This does not use the regular topic error reporting mechanism because the error occurs asynchronously.
   Screen Shot 2021-11-22 at 9 37 18 AM

- Adds an error message for unrecognized file types.
- Adds an error message when fetch() returns a non-200 status code.


defunctzombie:
Is there a way we could use the existing error reporting mechanism rather than a toast? Even if the operation is async it could be tracked and reported. It feels more intuitive to tie the error to the topic marker that produced it.

jtbandes:
I agree, though I think it will require a larger refactor. The errors are contained in SceneBuilder which runs synchronously during rendering. We would need something to collect async errors and merge them with those... and some way of making sure there is no race condition with async errors coming in after there is already a new frame (that would potentially fix those errors)

PR #2229, opened 2021-11-30
Clicking "View docs" should always open new tab

### LINKED ISSUE #2227: Clicking "View docs" from sidebar should open in a new tab
Clicking this link in the web build of studio navigates to a new page in the current tab, losing the active studio session. It should open in a separate tab.

image

esthersweon:
**User-Facing Changes**
- None (not yet merged)

**Description**
- Clicking "View docs" from Help sidebar's panel docs should open a new browser tab

Resolves #2227


jtbandes:
You might be able to do this by just changing the LinkRenderer in https://github.com/foxglove/studio/blob/jacob/playbackcontrol-optional/packages/studio-base/src/components/TextContent.tsx#L154 instead.

Issue #2067, opened 2021-11-02
Extension API: addPanel config is not applied


defunctzombie:
In terms of passing in the initial state, that partially worked for me with my extensions panel. That is, when it initially opens the sibling panel using `context.layout.addPanel`, the state is not applied. But if I restart Studio then it pulls the state I requested from disk. I'm guessing this is probably a problem on my end? I'm using this at the very start of my panel:

```
// DetailPanelState is my custom config type for this panel
  const [panelState, setPanelState] = useState<DetailPanelState>(
    context.initialState as DetailPanelState,
  );
```

_Originally posted by @jessicaaustin in https://github.com/foxglove/studio/issues/1914#issuecomment-953169873_

foxymiles:
Fixed by #2232

PR #2232, opened 2021-11-30
Initialize extension panels with the correct default state.

### LINKED ISSUE #1914: Extensions API: Add ability to open sibling panels
Originally posted at https://github.com/foxglove/fox/issues/28

Background:

I'm building an extension that shows data from a topic, and I would like to add the ability to open a sibling plot to show values over time, similar to the functionality in the Diagnostics panel and Raw Messages panel (they call `openSiblingPlotPanel` and `openSiblingStateTransitionsPanel` from `Plot`). 

Request:

Please add a function to the `PanelExtensionContext` to allow opening a sibling panel, with all the necessary props.

Thank you!

foxymiles:
**User-Facing Changes**
This fixes an issue in the extension panel API.

**Description**
This makes sure that when sibling panels are added via the extension panel API's `addPanel` function that they are initialized with the correct state.

Fixes #1914

### DIFF
--- a/packages/studio-base/src/components/Panel.tsx
+++ b/packages/studio-base/src/components/Panel.tsx
@@ -392,18 +392,21 @@ export default function Panel<
 
         // Otherwise, open new panel
         const newPanelPath = ownPath.concat("second");
-        void mosaicWindowActions.split({ type: panelType }).then(() => {
-          const newPanelId = getNodeAtPath(mosaicActions.getRoot(), newPanelPath) as string;
-          savePanelConfigs({
-            configs: [
-              {
-                id: newPanelId,
-                config: siblingConfigCreator(siblingDefaultConfig),
-                defaultConfig: siblingDefaultConfig,
-              },
-            ],
+        const newPanelConfig = siblingConfigCreator(siblingDefaultConfig);
+        void mosaicWindowActions
+          .split({ type: panelType, panelConfig: newPanelConfig })
+          .then(() => {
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/components/Panel.tsx
```diff
@@ -392,18 +392,21 @@ export default function Panel<
...
-            configs: [
-              {
-                id: newPanelId,
-                config: siblingConfigCreator(siblingDefaultConfig),
-                defaultConfig: siblingDefaultConfig,
-              },
-            ],
+        const newPanelConfig = siblingConfigCreator(siblingDefaultConfig);
+        void mosaicWindowActions
+          .split({ type: panelType, panelConfig: newPanelConfig })
```
defunctzombie, line 397: The bug was newPanelConfig was not passed in as the `panelConfig` property to split here?

foxymiles: Yeah the panel didn't receive the config on the initial `useLayoutEffect` render and then when the config is finally applied it doesn't cause the panel to update.

defunctzombie:
👍

jtbandes (approved):

jessicaaustin:
Tested this out. Works for me! This resolves https://github.com/foxglove/studio/issues/2067 for me

foxymiles:
@jessicaaustin Great! Thanks for testing and for the detailed bug report!

Issue #1958, opened 2021-10-11
Valid message path in Raw Message panel is highlighted in red


jhurliman:
**Description**

Open the attached bag, use a Raw Message panel and enter the following message path:

```
/example_markers.header.stamp.sec
```

It is highlighted in red, even though it works.

[poses.bag.zip](https://github.com/foxglove/studio/files/7325566/poses.bag.zip)

amacneil:
![image](https://user-images.githubusercontent.com/637671/142516203-6eca15d6-bfa6-4ab1-80de-d4eb4b21bb0f.png)

PR #2233, opened 2021-11-30
Fix message path input valid path detection for time and duration

defunctzombie:
**User-Facing Changes**
Writing a message path to .sec or .nsec fields within a field of type `time` or `duration` shows the path as valid.

**Description**
time and duration are considered primative types in ROS but are actually objects with two fields (sec, nsec). When specifying the sec or nsec field within a time field (i.e. header.stamp.sec) in the message path input, it would be marked as red indicating the path is invalid. This change fixes this behavior so paths accessing .sec or .nsec of time or duration fields is shown as valid.
    
The change produces a structure for time and duration with the message structure type and contains sec and nsec primitive fields. Previously the structure had time and duration primative fields with no knowledge of sec or nsec below.
    
Fixes: #1958

### DIFF
--- a/packages/studio-base/src/components/MessagePathSyntax/messagePathsForDatatype.test.ts
+++ b/packages/studio-base/src/components/MessagePathSyntax/messagePathsForDatatype.test.ts
@@ -81,9 +81,12 @@ describe("messagePathStructures", () => {
                 datatype: "std_msgs/Header",
               },
               stamp: {
-                primitiveType: "time",
-                structureType: "primitive",
-                datatype: "std_msgs/Header",
+                structureType: "message",
+                nextByName: {
+                  sec: { primitiveType: "uint32", structureType: "primitive", datatype: "" },
+                  nsec: { primitiveType: "uint32", structureType: "primitive", datatype: "" },
+                },
+                datatype: "time",
               },
             },
             structureType: "message",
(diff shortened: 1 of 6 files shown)


packages/studio-base/src/components/MessagePathSyntax/messagePathsForDatatype.ts
```diff
@@ -64,6 +64,25 @@ export function messagePathStructures(
...
   lastDatatypes = undefined;
   const structureFor = memoize((datatype: string): MessagePathStructureItemMessage => {
+    if (datatype === "time" || datatype === "duration") {
+      return {
+        structureType: "message",
+        nextByName: {
+          sec: {
+            structureType: "primitive",
+            primitiveType: "uint32",
+            datatype: "",
```
jtbandes, line 74: seems like datatype should maybe be `datatype` rather than empty string?

packages/studio-base/src/components/MessagePathSyntax/messagePathsForDatatype.test.ts
```diff
@@ -204,13 +220,17 @@ describe("messagePathsForDatatype", () => {
       ".header.frame_id",
       ".header.seq",
       ".header.stamp",
+      ".header.stamp.nsec",
+      ".header.stamp.sec",
```
jtbandes, line 224: Now that I think about it, maybe the reason this was treated as a primitive before is that we didn't want sec and nsec showing up in autocomplete all the time?

jtbandes (approved):
Seems ok – worth thinking about whether it's annoying to have sec and nsec showing up in autocomplete dropdowns. Any other ideas how the app user experience might differ with this change?

packages/studio-base/src/components/MessagePathSyntax/messagePathsForDatatype.ts
defunctzombie: I wasn't sure actually. I think for primitive types this field is not used. Maybe needs more typescript union type love.

packages/studio-base/src/components/MessagePathSyntax/messagePathsForDatatype.test.ts
defunctzombie: image

For whatever reason it doesn't seem like those show up in the autocomplete at the top level.

image

But they do show up when looking within stamp. I wonder if there's a depth thing at work here?

PR #2237, opened 2021-12-01
Specify in metrics when player was initialized with the demo bag

esthersweon:
**User-Facing Changes**
- None

**Description**
- Differentiate between players being initialized with demo bag vs. user data by passing a `isDemoBag` arg to the `PLAYER_INITIALIZED` metric
- Made sure all players log the `PLAYER_INITIALIZED` metric

### DIFF
--- a/packages/studio-base/src/players/FoxgloveDataPlatformPlayer/index.ts
+++ b/packages/studio-base/src/players/FoxgloveDataPlatformPlayer/index.ts
@@ -125,7 +125,6 @@ export default class FoxgloveDataPlatformPlayer implements Player {
       this._presence = PlayerPresence.ERROR;
       this._addProblem("open-failed", { message: error.message, error, severity: "error" });
     });
-    this._metricsCollector.initialized();
   }
 
   private _open = async (): Promise<void> => {
@@ -225,6 +224,7 @@ export default class FoxgloveDataPlatformPlayer implements Player {
     this._initialized = {
       preloadedMessages: new MessageMemoryCache({ start: this._start, end: this._end }),
     };
+    this._metricsCollector.initialized();
     this._emitState();
     this._startPreloadTaskIfNeeded();
   };
--- a/packages/studio-base/src/players/FoxgloveDataPlatformPlayer/index.ts
+++ b/packages/studio-base/src/players/FoxgloveDataPlatformPlayer/index.ts
@@ -125,7 +125,7 @@ export default class FoxgloveDataPlatformPlayer implements Player {
       this._presence = PlayerPresence.ERROR;
       this._addProblem("open-failed", { message: error.message, error, severity: "error" });
     });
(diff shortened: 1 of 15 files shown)


packages/studio-base/src/players/FoxgloveDataPlatformPlayer/index.ts
```diff
@@ -125,6 +125,7 @@ export default class FoxgloveDataPlatformPlayer implements Player {
       this._presence = PlayerPresence.ERROR;
       this._addProblem("open-failed", { message: error.message, error, severity: "error" });
     });
+    this._metricsCollector.initialized({ isDemoBag: false });
```
jhurliman, line 128: Having to set this to false everywhere feels like it is leaking a detail of one part of the app (the fact that we have a demo bag) to all data sources. Can this be made optional (and defaulted to false), so only the actual demo bag playback sets `isDemoBag: true`?

packages/studio-base/src/util/isDemoBagUrl.ts
```diff
@@ -0,0 +1,8 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+const DEMO_BAG_URL = "https://storage.googleapis.com/foxglove-public-assets/demo.bag";
```
jhurliman, line 5: This isn't exported so I assume we are defining this URL in two places in the app now. Can it be exported once and imported here?

jhurliman (changes requested):

packages/studio-base/src/players/FoxgloveDataPlatformPlayer/index.ts
jhurliman: I don't think this is the correct place to call `initialized`. PLAYER_CONSTRUCTED conveys when the data source constructor is called. PLAYER_INITIALIZED, in other players, happens when a connection is successfully established.

jhurliman (changes requested):

packages/studio-base/src/players/FoxgloveDataPlatformPlayer/index.ts
esthersweon: Moved to past `this._initialized = {  preloadedMessages: ... }` later in this file.

packages/studio-base/src/util/isDemoBagUrl.ts
esthersweon: There's currently a `DEMO_BAG_URL` in both the desktop and web root.

defunctzombie:
Looking back at this and I'm generally not a fan of how this hard-coded the plumbing for demo url detection. We should be passing properties to constructors if we want to influence actions like this rather than adding custom handling for constants in the player.

PR #2238, opened 2021-12-01
Fix a 3D panel crash with markers but no /tf topic

jhurliman:
**User-facing changes**

When playing back a data source that publishes markers but no transforms (no `/tf` topic or otherwise), the 3D panel will no longer occasionally crash.

### DIFF
--- a/packages/studio-base/src/panels/ThreeDimensionalViz/SceneBuilder/index.ts
+++ b/packages/studio-base/src/panels/ThreeDimensionalViz/SceneBuilder/index.ts
@@ -434,7 +434,7 @@ export default class SceneBuilder implements MarkerProvider {
   private _transformMarkerPose = (topic: string, marker: BaseMarker): MutablePose | undefined => {
     const frame_id = marker.header.frame_id;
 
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/panels/ThreeDimensionalViz/SceneBuilder/index.ts
```diff
@@ -434,7 +434,7 @@ export default class SceneBuilder implements MarkerProvider {
   private _transformMarkerPose = (topic: string, marker: BaseMarker): MutablePose | undefined => {
     const frame_id = marker.header.frame_id;
 
-    if (frame_id.length === 0) {
+    if (!frame_id) {
```
jtbandes, line 437: Are these two separate fixes? Does this somehow help with markers that have `frame_id: ""`? Would be helpful to update the PR description to include an explanation of what changes were necessary

jhurliman: This is an aesthetic change, not a fix.

PR #2243, opened 2021-12-01
Transform interpolation

jhurliman:
**User-Facing Changes**

Transforms are now interpolated. The `frame_locked` attribute of markers is now observed. Various display bugs are fixed (https://github.com/foxglove/studio/issues/2264, https://github.com/foxglove/studio/issues/2265, https://github.com/foxglove/studio/issues/2266, https://github.com/foxglove/studio/issues/2272) in the 3D panel where objects that could not be transformed into the current frame were erroneously rendered relative to the origin. A bug is fixed where the 3D panel topic picker would erroneously report missing transforms.

https://user-images.githubusercontent.com/195374/144687880-bbd703e4-0b70-4878-a617-12af854e1903.mov

**Description**

The details of ROS transform interpolation are described in [this paper](http://wiki.ros.org/Papers/TePRA2013_Foote?action=AttachFile&do=view&target=TePRA2013_Foote.pdf). The high level is that a buffer of every received transform in the last 10 seconds for each coordinate frame is held and transforms are always done at a point in time. If the requested time falls in between two transforms for any coordinate frame(s) between the child and parent, LERP+SLERP is used for interpolation. Extrapolation is disallowed. To translate between two sibling frames, a common parent is searched for and if found, a transform is done from A to parent followed by an inverse transform from parent to B.

To accomplish this, all coordinate transformations are now parameterized by time (instead of always using the latest transform). This adds `currentTime` as a data dependency to more React memoizations, but in testing, I have not observed a regression in app performance. I've taken extra care to make the storage and retrieval of transforms fast (using AVL trees) and the math fast by caching vec3+quat <-> mat4 transforms.

One difference from the ROS implementation is how we handle transform requests in the future. In ROS TF2, this will either block until new transform data is available or throw an exception. Since we cannot block rendering and we can't guarantee data will always arrive in the exact order it was received on a particular robot node (live connections, recording latency, etc), a slop factor is used to clamp transform requests up to one second in the future to the most recently received transform. This is a tradeoff between the accuracy of what is visualized and whether we visualize anything at all.


defunctzombie:
> A bug is fixed where the 3D panel topic picker would erroneously report missing transforms.

Under what conditions would this missing transforms report be erroneous?

> The `frame_locked` attribute of markers is now observed

What does this mean? Do we want to update docs with this info or is this something we want to keep internal for now? This is an rviz/marker feature?

Are any 3d panel settings we should be thinking about exposing? Maybe some of these slop arguments you mentioned?

packages/studio-base/src/panels/ThreeDimensionalViz/GridBuilder.stories.tsx
```diff
@@ -115,7 +115,7 @@ storiesOf("panels/ThreeDimensionalViz/GridBuilder", module)
     const topicSettings: TopicSettingsCollection = {};
     topicSettings[`t:${FOXGLOVE_GRID_TOPIC}`] = settings;
     gridBuilder.setSettingsByKey(topicSettings);
-    gridBuilder.renderMarkers(collector);
+    gridBuilder.renderMarkers(collector, { sec: 0, nsec: 0 });
```
defunctzombie, line 118: Is the time something specific to each renderMarkers call or something that should be separate on the builder? i.e. (builder.setTransformTime(...) or some such and all rendering is done using the last transform time?)

packages/studio-base/src/panels/ThreeDimensionalViz/SceneBuilder/index.ts
```diff
@@ -729,20 +750,21 @@ export default class SceneBuilder implements MarkerProvider {
...
+    const pose = this.transforms?.apply(
       sourcePose,
       sourcePose,
-      drawData.header.frame_id,
       this.rootTransformID,
+      frameId,
+      drawData.header.stamp,
     );
     if (!pose) {
       // Don't error on frame_id="", interpret it as an identity transform
```
defunctzombie, line 762: Is this comment still valid?

packages/studio-base/src/panels/ThreeDimensionalViz/threeDimensionalVizUtils.ts
```diff
@@ -46,18 +47,18 @@ function getZoomDistanceFromURLParam(): number | undefined {
...
-      return {
-        target: [px, py, pz],
-        targetOrientation: [ox, oy, oz, ow],
-      };
+  if (typeof followTf === "string" && followTf.length > 0) {
+    const frame = transforms.frame(followTf);
+    if (frame) {
+      const pose: MutablePose | undefined = emptyPose();
+      if (frame.root().apply(pose, pose, frame, time)) {
+        const p = pose.position;
```
defunctzombie, line 58: nit: more than one letter for variable name - easier to grok. These aren't math variable names right?

packages/studio-base/src/panels/ThreeDimensionalViz/transforms/CoordinateFrame.ts
```diff
@@ -0,0 +1,321 @@
...
+
+/**
+ * CoordinateFrame is a named 3D coordinate frame with an optional parent frame
+ * and a history of transforms from this frame to its parent. The parent/child
+ * hierarchy and transform history allow points to be transformed from one
+ * coordinate frame to another while interpolating over time.
+ */
+export class CoordinateFrame {
+  readonly id: string;
+  maxStorageTime: Duration;
```
defunctzombie, line 42: private?

packages/studio-base/src/panels/ThreeDimensionalViz/transforms/CoordinateFrame.ts
```diff
@@ -0,0 +1,321 @@
...
+
+    // Remove transforms that are too old
+    const endTime = this._transforms.maxKey()!;
+    const startTime = subtract(endTime, this.maxStorageTime);
+    while (this._transforms.size > 1 && isLessThan(this._transforms.minKey()!, startTime)) {
+      this._transforms.shift();
+    }
+  }
+
+  findClosestTransforms(
```
defunctzombie, line 112: Options object for the args here? 4 args starts to get hard to reason about at the call site.

packages/studio-base/src/panels/ThreeDimensionalViz/transforms/CoordinateFrame.ts
```diff
@@ -0,0 +1,321 @@
...
+    // Return the transforms closest to the requested time
+    const [gtTime, gtTf] = gt;
+    outLower[0] = lteTime;
+    outLower[1] = lteTf;
+    outUpper[0] = gtTime;
+    outUpper[1] = gtTf;
+    return true;
+  }
+
+  apply(
```
defunctzombie, line 177: Object options object here as well.

Separate from that - maybe a comment on what this does and the arguments.

packages/studio-base/src/panels/ThreeDimensionalViz/transforms/CoordinateFrame.ts
```diff
@@ -0,0 +1,321 @@
...
+      if (curFrame._parent == undefined) {
+        throw new Error(`Frame "${parentFrame.id}" is not a parent of "${childFrame.id}"`);
+      }
+      curFrame = curFrame._parent;
+    }
+
+    return true;
+  }
+
+  static Apply(
```
defunctzombie, line 273: Definitely needs an options object - this has too many args to be obvious to users.

packages/studio-base/src/panels/ThreeDimensionalViz/transforms/Transform.ts
```diff
@@ -0,0 +1,126 @@
...
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { mat4, vec3, quat, ReadonlyMat4, ReadonlyVec3, ReadonlyQuat } from "gl-matrix";
+
+import { MutablePose, Pose } from "@foxglove/studio-base/types/Messages";
+
+import { mat4Identity, quatIdentity, vec3Identity } from "./geometry";
+
+const tempMat = mat4Identity();
```
defunctzombie, line 11: TEMP_MAT? Other places in this PR you use that style.

packages/studio-base/src/panels/ThreeDimensionalViz/transforms/Transform.ts
```diff
@@ -0,0 +1,126 @@
...
+    out.orientation.y = this._rotation[1];
+    out.orientation.z = this._rotation[2];
+    out.orientation.w = this._rotation[3];
+  }
+
+  static Identity(): Transform {
+    return new Transform(vec3Identity(), quatIdentity());
+  }
+
+  static Interpolate(out: Transform, a: Transform, b: Transform, t: number): void {
```
defunctzombie, line 118: Docs comment plz.

What is `t` here?

packages/studio-base/src/panels/ThreeDimensionalViz/transforms/TransformTree.ts
```diff
@@ -0,0 +1,84 @@
...
+import { quatFromValues, vec3FromValues } from "./geometry";
+
+/**
+ * TransformTree is a collection of coordinate frames with convenience methods
+ * for getting and creating frames and adding transforms between frames.
+ */
+export class TransformTree {
+  private _frames = new Map<string, CoordinateFrame>();
+
+  addTransform(frameId: string, parentFrameId: string, time: Time, transform: Transform): void {
```
defunctzombie, line 19: Options args would make this easier to read at the call site. The arg order is not obvious to the caller otherwise.

packages/studio-base/src/panels/ThreeDimensionalViz/transforms/TransformTree.ts
```diff
@@ -0,0 +1,84 @@
...
+      this._frames.set(id, frame);
+    }
+    return frame;
+  }
+
+  frames(): ReadonlyMap<string, CoordinateFrame> {
+    return this._frames;
+  }
+
+  apply(
```
defunctzombie, line 65: Options arg

packages/studio-base/src/panels/ThreeDimensionalViz/GridBuilder.stories.tsx
jhurliman: I generally dislike imperative state changes in favor of functional approaches so I may be biased here, but I can't see the value of:

```
gridBuilder.setTransformTime(t);
gridBuilder.renderMarkers(collector);
```
over
```
gridBuilder.renderMarkers(collector, t);
```
When the timestamp is only relevant to marker rendering and doesn't need to be held as a class member for any other reason.

packages/studio-base/src/panels/ThreeDimensionalViz/SceneBuilder/index.ts
jhurliman: Yes, the next line still has (and needs) the frame_id length check because `""` is a valid JS string but it has a special meaning in the `frame_id` field (we shouldn't show the user an error saying `Transform "" is missing!)

packages/studio-base/src/panels/ThreeDimensionalViz/threeDimensionalVizUtils.ts
jhurliman: It's a setup to use math notation on the line below, approximating p<sub>x</sub> (et al) with `p.x`.

packages/studio-base/src/panels/ThreeDimensionalViz/transforms/CoordinateFrame.ts
jhurliman: It's left read/write and public on purpose. It's valid to modify `maxStorageTime` after construction if you'd like.

packages/studio-base/src/panels/ThreeDimensionalViz/GridBuilder.stories.tsx
defunctzombie: 👍 It came to mind since reading `renderMarkers(collector, t)` leaves the question what is `t` for this context? Am I rendering when time is `t`? Am I rendering only markers at `t`? I think in this case its interpolate at `t`? Maybe the name of the function could guide towards understanding what it will be doing with the arguments.

jhurliman: I thought about this and I think it makes sense to leave it as-is. It's a common pattern for render functions to be parameterized by time, see [requestAnimationFrame](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame).

packages/studio-base/src/panels/ThreeDimensionalViz/SceneBuilder/index.ts
jhurliman: I refactored this away since we decided to allow `""` as a valid frame_id.

jtbandes:
Can we please add some new story screenshots that show the correct behavior visually?

PR #2244, opened 2021-12-01
Remove URDF Viewer panel

esthersweon:
**User-Facing Changes**
- Remove URDF Viewer panel - all URDF viewing functionality now in 3D panel

**Description**
- Remove URDF Viewer panel - all URDF viewing functionality now in 3D panel


jtbandes:
The joint positioning functionality is not in the 3D panel as far as I know, so maybe we'd want to wait on this? cc @jhurliman

jtbandes:
We should also think about the migration story for users who had a URDF panel in their layout – can we show them something nicer than "Unknown panel"?

jhurliman:
No joint positioning and no support for opening a file or drag and dropping a `.urdf` file. The 3D panel URDF support only supports http(s):// and package:// URLs, handling files is challenging because reloading the browser window or app would break the `File` handle.

jtbandes:
> handling files is challenging because reloading the browser window or app would break the File handle.

isn't this true of the URDF Viewer panel too? or what did you mean?

defunctzombie:
Did we decide if we want to keep this panel or remove it?

PR #2247, opened 2021-12-01
WebSocket: Remove leading dot from Protobuf datatype name

jtbandes:
**User-Facing Changes**
None (feature flagged)

**Description**
This will be needed for the WebSocket player to take advantage of #2246.

I'm not sure if there are any caveats to doing this (like potential name conflicts). My guess is it's fine, since _all_ `fullName`s should have a leading dot, removing it should not cause any name collisions.

### DIFF
--- a/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
+++ b/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
@@ -60,7 +60,13 @@ function parseChannel(channel: Channel): ParsedChannel {
   const datatypes: RosDatatypes = new Map();
   protobufDefinitionsToDatatypes(datatypes, type);
 
-  return { channel, fullSchemaName: type.fullName, deserializer, datatypes };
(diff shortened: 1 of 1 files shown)


defunctzombie:
Should we be doing this here? Or should the sending side be responsible for this?

In the bag2proto tool I stripped the leading dots from the package fields of the descriptor set. That was motivated by not being able to load the descriptor set otherwise. Which was quite annoying since a round trip of `toDescriptor/fromDescriptor` did not work.

jtbandes:
When I was testing with eCAL HDF5 files I don't think it was the protobuf descriptorset that contained leading dots – it's just something the protobufjs library produces from fullName:

https://github.com/protobufjs/protobuf.js/blob/95b56817ef6fb9bdcb14d956c159da49d0889bff/src/object.js#L87

This code produces a leading dot because the top level parent has a `name` of `""` (at least in the descriptor I was using). Maybe these things vary between descriptors though?

PR #2252, opened 2021-12-02
Update mcap data provider protobuf handling for descriptor set

defunctzombie:
**User-Facing Changes**
Mcap is an experimental feature. This changes the requirements for mcap files containing protobuf encoded messages.

**Description**
Assume the _schema_ field for a protobuf encoding channel info contains a base64 encoded descriptor set. The descriptor set indicates how to decode binary protobuf messages into objects.


defunctzombie:
@jtbandes I tested this change using your proto names PR (https://github.com/foxglove/studio/pull/2246) and an mcap file generated via https://github.com/foxglove/mcap/pull/3 from the demo bag.

Screen Shot 2021-12-01 at 8 41 46 PM

PR #2254, opened 2021-12-02
Fix McapDataProvider ros1 and ros2 encoding detection

defunctzombie:
**User-Facing Changes**
None - mcap is not used publicly yet.

**Description**
MCapdDataProvider had incorrect logic for detecting ros1 and ros2 encoding. This fixes the logic.


jtbandes:
👍

PR #2256, opened 2021-12-02
Fixed regression where player type was not being reported with player initialization 

esthersweon:
**User-Facing Changes and Description**
- Fixed regression where player type was not being reported with player initialization 
- Also removed Foxglove Data Platform connection from the desktop app menu's list of connection options
- Also fixed formatting of ampersand in the desktop app's File menu ("Rosbridge (ROS 1 & 2)")

Screen Shot 2021-12-02 at 3 19 55 PM

### DIFF
--- a/desktop/main/StudioWindow.ts
+++ b/desktop/main/StudioWindow.ts
@@ -391,10 +391,14 @@ class StudioWindow {
   }
 
   addInputSource(name: string): void {
+    // A "Foxglove Data Platform" connection is triggered by opening a URL from console
+    // Not currently a connection that can be started from inside Foxglove Studio
     const unsupportedInputSourceNames = ["Foxglove Data Platform"];
     if (unsupportedInputSourceNames.includes(name)) {
       return;
     }
+
+    // Electron menus require a preceding & to escape the & char
     const formattedName = name
       .split("")
       .map((char) => (char === "&" ? "&&" : char))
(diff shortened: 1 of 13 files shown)


desktop/main/StudioWindow.ts
```diff
@@ -391,14 +391,22 @@ class StudioWindow {
   }
 
   addInputSource(name: string): void {
-    this._inputSources.add(name);
+    const unsupportedInputSourceNames = ["Foxglove Data Platform"];
```
jhurliman, line 394: Why is this unsupported? Comment?

desktop/main/StudioWindow.ts
```diff
@@ -391,14 +391,22 @@ class StudioWindow {
...
+    if (unsupportedInputSourceNames.includes(name)) {
+      return;
+    }
+    const formattedName = name
+      .split("")
+      .map((char) => (char === "&" ? "&&" : char))
```
jhurliman, line 400: I can see this is doubling the number of ampersands, but I'm not sure why.

esthersweon: https://github.com/electron/electron/issues/25028, which led me to https://www.electronjs.org/docs/latest/api/menu#static-methods.

Will add comment.

desktop/main/StudioWindow.ts
esthersweon: Adding the following:

```
 // A "Foxglove Data Platform" connection is triggered by opening a URL from console
 // Not currently a connection that can be started from inside Foxglove Studio
```

PR #2269, opened 2021-12-03
Upgrade webpack-dev-server

jtbandes:
**User-Facing Changes**
None

**Description**
Upgraded by following migration guide at https://github.com/webpack/webpack-dev-server/blob/master/migration-v4.md

Fixes another apparent recurrence of https://github.com/prettier/prettier/pull/11130. This was more annoying now because the warning overlay is on by default in the new dev server version.

### DIFF
--- a/packages/studio-base/webpack.ts
+++ b/packages/studio-base/webpack.ts
@@ -180,6 +180,11 @@ export function makeConfig(
                 replace:
                   "throw new Error('[Foxglove] This module is not supported in the browser.');",
               },
+              {
+                search: `getModuleResolver=function(e){let t;try{t=require(e)}`,
+                replace:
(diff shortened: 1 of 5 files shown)


packages/studio-base/webpack.ts
```diff
@@ -180,6 +180,11 @@ export function makeConfig(
                 replace:
                   "throw new Error('[Foxglove] This module is not supported in the browser.');",
               },
+              {
+                search: `getModuleResolver=function(e){let t;try{t=require(e)}`,
```
defunctzombie, line 184: Whats this about?

jtbandes: > Fixes another apparent recurrence of prettier/prettier#11130. This was more annoying now because the warning overlay is on by default in the new dev server version.

PR #2274, opened 2021-12-04
Combine same series hover values in timeseries tooltips

defunctzombie:
**User-Facing Changes**
When hovering on multiple values for a series all values are shown under a series label.

| before | after |
| --- | --- |
| image|image|

**Description**
Depending on the zoom level, the timeseries chart tooltip may show multiple values from a series. These values would each have the series label. This change groups values from the same series under one series label. This saves some vertical space by not repeating the series label and makes it easier to see that a series has multiple values at the hover location.

### DIFF
--- a/packages/studio-base/src/components/TimeBasedChart/TimeBasedChartTooltipContent.stories.tsx
+++ b/packages/studio-base/src/components/TimeBasedChart/TimeBasedChartTooltipContent.stories.tsx
@@ -21,7 +21,7 @@ export default {
   component: TimeBasedChartTooltipContent,
 };
 
-export function SingleItem(): React.ReactElement {
+export function SingleItemSingleDataset(): React.ReactElement {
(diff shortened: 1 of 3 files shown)


jtbandes:
Does the tooltip include all nearby items pre-downsampling, or does it only include items that survived the downsampling?

jtbandes:
I would think if we are downsampling, then the tooltip would only need to show 1 item at a time from the hovered series. I would expect multiple values at a time if we are not downsampling (but then it might be a really large number of values)

defunctzombie:
> I would think if the tooltip can only show downsampled values, then it would only need to show 1 item at a time from the hovered series. I would expect multiple values at a time if it is showing non-downsampled values (but then it might be a really large number of values)

The hover hit-region for chartjs is larger than 1px thats why we sometimes get adjacent values within the same series.

PR #2275, opened 2021-12-04
Make panel toolbars always visible

defunctzombie:
**User-Facing Changes**
Panel toolbars are always visibile regardless of user mouse position.

**Description**
Elements that change their visibility state based on the mouse hover on a panel draw attention as the user moves their cursor around the app or into a panel. They also make it harder to see where to move your cursor within a panel to use the actions on the toolbar.
    
By always showing the toolbar the user knows where the control icons are. The other elements do not shift when the toolbar appears/disappears since it is always visible.


defunctzombie:
@amacneil @2metres You can play with this in the preview version for this PR: https://studio-j0tkgvxwk-foxglove.vercel.app/

Overall I like that items are not constantly appearing/disappearing but I can also sympathize with how the toolbar constantly covers up panel space.

Issue #2248, opened 2021-12-01
Toolbar disappears in fullscreen mode if the panel is in a tab


bunnyc1986:
**Description**
Toolbar disappears in fullscreen mode if the panel is in a tab

- Version: The latest version on demo website
- OS: Mac OS 11.6, and Chrome 96.0.4664.55

**Steps To Reproduce**

![image](https://user-images.githubusercontent.com/1644274/144327052-14807ff0-4c6b-4f1a-9083-a843f81bad6c.png)

**Expected Behavior**
The toolbar should show up as if the panel is not in a tab.

**Actual Behavior**
Toolbar disappears in fullscreen mode if the panel is in a tab.

defunctzombie:
@bunnyc1986 could you elaborate on the steps to reproduce this?

I create a new tab panel and an image panel inside the tab panel. I can fullscreen both panels and continue to see the appropriate panel controls.

jtbandes:
Maybe this is a z-index issue? I see some problems when fullscreening a panel that's inside a tab if there are also other panels adjacent to the tab:

https://user-images.githubusercontent.com/14237/144651986-214ca4d3-b64a-41d2-a41d-fc38b3a0ce16.mov

bunnyc1986:
> @bunnyc1986 could you elaborate on the steps to reproduce this?
> 
> I create a new tab panel and an image panel inside the tab panel. I can fullscreen both panels and continue to see the appropriate panel controls.

@jtbandes 's screen recording has the exact reproduction steps. Thanks for posing it!

PR #2276, opened 2021-12-04
Toggle fullscreen state for parent panels

defunctzombie:
**User-Facing Changes**
Fullscreen toggle for panels works correctly for panels within tabs.

Removes the fullscreen option from the quick access overlay.

**Description**
When a panel within a tab enters fullscreen, it needs to tell the tab panel to enter fullscreen to properly appear on top of other panels and controls.
    
This change also removes the "fullscreen" feature from the quick access controls. There is already a fullscreen icon in the panel toolbar which is available without having to open any menus or special shortcut keys.
    
Fixes: #2248

### DIFF
--- a/packages/studio-base/src/components/CssBaseline.tsx
+++ b/packages/studio-base/src/components/CssBaseline.tsx
@@ -84,7 +84,7 @@ const useStyles = makeStyles((theme) => ({
       ".mosaic-tile": {
         // make room for splitters - unfortunately this means the background color will show
         // through even if the tile has its own background color set
-        margin: 1,
+        margin: 0,
       },
       ".mosaic-window": {
         boxShadow: "none",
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/components/CssBaseline.tsx
```diff
@@ -84,7 +84,7 @@ const useStyles = makeStyles((theme) => ({
       ".mosaic-tile": {
         // make room for splitters - unfortunately this means the background color will show
         // through even if the tile has its own background color set
-        margin: 1,
+        margin: 0,
```
jtbandes, line 87: This should not have been changed without changing the comment – with this changing to 0, now panel scrollbars appear cut off, and the comment is out of date

image

defunctzombie: 👍 It wasn't clear to me what impact that had since I did not notice any change in the splitters after making that change.

This margin creates a little black border around the panels (since the panel bg is different than the document bg) Maybe something that became noticeable after the color scheme changes.

Thoughts on what should be done here?

defunctzombie: https://github.com/foxglove/studio/pull/2281

Issue #2278, opened 2021-12-04
Point cloud does not render from demo bag


defunctzombie:
**Description**
When I open the demo bag and select the `/velodyne_points` topic in a new 3d panel I expect to see the point cloud show up. However nothing shows up when I play the file.

Regression from: https://github.com/foxglove/studio/pull/2243

**Steps To Reproduce**

1. Open a new layout
2. Add a 3d panel
3. Open the demo bag
4. Select `/velodyne_points` in the 3d panel
5. Play the file

**Expected Behavior**
After the above steps I expect to see point clouds rendered in the 3d scene.

**Actual Behavior**
Nothing is rendered in the 3d scene.

jhurliman:
I cannot repro using either latest `main` or https://studio.foxglove.dev/ and `demo.bag`. Screenshots:
<img width="1312" alt="Screen Shot 2021-12-05 at 2 38 46 PM" src="https://user-images.githubusercontent.com/195374/144766769-49514293-e712-4cff-b1ea-ae3d43225822.png">
<img width="1358" alt="Screen Shot 2021-12-05 at 2 39 54 PM" src="https://user-images.githubusercontent.com/195374/144766773-c78d37ae-b297-4dc8-8945-a35e9192c8bc.png">

defunctzombie:
The issue is that when adding a new 3d panel, then selecting the `velodyne_points` topic - I should see the points show up. That does not happen. In the above screenshots you are looking at the welcome layout which is already configured. Try adding a blank 3d panel and then enabling the topic.

The behavior I am describing should show the selected topic without the user having to configure the frame.

defunctzombie:
Seems like if this is a regression it isn't recent. Closing out the bug and we can separately discuss any UX around making it more obvious where to select the frame.

Issue #2277, opened 2021-12-04
Point cloud topic does not render when selected


defunctzombie:
**Description**
When I try to render the /radar/points and /velodyne_points topics from the demo bag one one of them shows up in the 3d scene with no indication on why the other does not show.

This is a regression from https://github.com/foxglove/studio/pull/2243

**Steps To Reproduce**

1. Open the demo bag
2. Enable `/radar/points` in the 3d panel
3. Select the "radar" frame
4. Enable `/velodyne_points` in the 3d panel

Notice that playin the bag does not show the velodyne points with no indication why.

**Expected Behavior**
An error notice that velodyne_points is missing a needed transform

Here is what it looks like in the last release of the studio desktop app:
![image](https://user-images.githubusercontent.com/84792/144717852-8835b6b4-c81b-4cae-aacf-68f2e763f004.png)

**Actual Behavior**
No velodyne points are displayed and no indication why.

PR #2279, opened 2021-12-06
Regression fixes for new TransformTree

jhurliman:
**User-facing changes**

None, regression fixes in between releases.

**Description**

- Don't accumulate pose transforms on URDF markers
- Avoid a repeated pose transform
- Remove unused topicsWithBadFrameIds
- Remove topicsMissingFrameIds and treat empty frame_id as valid
- Update non-marker message poses per-frame instead of once at consumption, report topic errors when any marker cannot be rendered due to transform failure


defunctzombie:
Tested out the preview deployment and can confirm that https://github.com/foxglove/studio/issues/2277 is fixed.

https://github.com/foxglove/studio/issues/2278 is unfixed.

defunctzombie:
Please add some stories for the 3d panel so we can see if this regresses in future refactors.

PR #2281, opened 2021-12-06
Use 1px margin for mosaic tiles next to a splitter.

defunctzombie:
**User-Facing Changes**
Panel elements are not obstructed when appearing within 1px of a splitter.

**Description**
react-mosaic renders tiles adjacent to each other and splitters on top of tiles. A tile without a margin is obstructed by any adjacent splitters. However, having a margin on all tiles results in unsightly gaps and borders. This change adds a margin to tiles only when there would be an adjacent splitter.


jtbandes (approved):
👾

Issue #2284, opened 2021-12-06
Map panel: add user-defined GeoJSON layer


jessicaaustin:
When viewing the Map Panel, it would be helpful to be to "annotate" the map, for example to mark a location of interest. 

One relatively quick way to do this would be to allow the user to specify a GeoJSON in the panel config. Then that could be loaded into the leaflet map. 

An example: 

Code: [add custom geojson layer ](https://github.com/jessicaaustin/studio/commit/a4349fda925d2d72e74b49856b9289466508ec7a)

In Studio:  ![http:--localhost:11311- – Foxglove Studio_999](https://user-images.githubusercontent.com/1159930/144927049-87f4913c-2314-427e-9b32-bcc2cb46a2f8.png)

jessicaaustin:
I attached a PR where I'm playing around with this idea.

If you guys are interested in this idea, and agree with this approach, I'd like to flesh this out and enable a user to specify the GeoJSON in panel settings. I'm not quite sure what the best way is to add panel settings for a panel that's using the PanelExtensionAdapter though -- could you point me in the right direction?

defunctzombie:
> I'm not quite sure what the best way is to add panel settings for a panel that's using the PanelExtensionAdapter though -- could you point me in the right direction?

I think we haven't exposed that to extensions - because we ourselves weren't too sold on how settings are setup today for non-extension panels.

Regarding displaying geoJSON data (or any overlays) is the use case to have a static file referenced or is the use case to have a topic that could produce geoJSON data and that data comes from the robot and could change over time?

jessicaaustin:
> Regarding displaying geoJSON data (or any overlays) is the use case to have a static file referenced or is the use case to have a topic that could produce geoJSON data and that data comes from the robot and could change over time?

The use case is static. Either a static file, or even have the user copy/paste in a block of geojson in the panel settings.

PR #2285, opened 2021-12-06
Add user-defined custom geojson layer to map

jessicaaustin:
See https://github.com/foxglove/studio/issues/2284

### DIFF
--- a/packages/studio-base/src/panels/Map/MapPanel.tsx
+++ b/packages/studio-base/src/panels/Map/MapPanel.tsx
@@ -11,6 +11,7 @@ import {
   FeatureGroup,
   LayersControlEvent,
   LayerGroup,
+  geoJSON,
 } from "leaflet";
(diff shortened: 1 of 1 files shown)


defunctzombie:
This is blocked on either exposing a _settings_ api to extensions panels (The builtin Map panel is authored via the extension api), or finding a UI within the panel to allow the user to setup the custom geojson layer.

defunctzombie:
@jessicaaustin This came up in another context recently. The use case there was suggesting reading the GeoJSON from a topic rather than hard-coded into the panel settings. Wondering if that would be compelling or meet your use case too?

jessicaaustin:
@defunctzombie Yes that would work as well, and probably a more flexible solution anyway for integrating into an existing system.

defunctzombie:
Closing this out since we landed https://github.com/foxglove/studio/pull/3078 which allows you to send geojson data within messages on topics.

Please give it a try and send feedback!

Thanks for the initial prototype here and feedback.

PR #2286, opened 2021-12-06
Deselect the active layout if it's deleted in another session.

### LINKED ISSUE #1996: Deselect layout when selected layout is deleted in the background
**Description**
When the selected layout is deleted on the server and a subsequent sync removes it from the cache, it's still available in the UI and attempts to update it result in `Your changes could not be saved. Error: Cannot update layout ... because it does not exist`

- Version: 0.19.0
- OS: Linux

**Steps To Reproduce**

1. Open two tabs
2. Select the same layout in both tabs
3. Delete the layout in one tab
4. Go back to your other tab. The layout has been deleted from the sidebar. But when you try to modify the current layout you get an annoying error. There is no way to "save" the state that you are in.

**Expected Behavior**
- Simple solution for now: Just deselect the layout (so no layout is selected) if the layout is deleted out from under you. Especially for personal layouts, if I deleted a layout then I presumably want it deleted.
- Maybe we could add some more advanced handling for team layouts in the future.

**Actual Behavior**
```
Your changes could not be saved. Error: Cannot update layout ef745db5-fabb-4cee-80a7-2569df0fdd84 because it does not exist
```
image

**Dev console logs**
```
index.ts?b144:474 [../../packages/studio-base/src/services/LayoutManager/index.ts] Starting layout sync
index.ts?b144:480 [../../packages/studio-base/src/services/LayoutManager/index.ts] Completed sync in 0.11s
LayoutManagerProvider.tsx?c674:70 [../../packages/studio-base/src/providers/LayoutManagerProvider.tsx] Waiting 4.81 sec for next sync {failures: 0}
index.tsx?9ba6:186 [../../packages/studio-base/src/providers/CurrentLayoutProvider/index.tsx] Error: Cannot update layout ef745db5-fabb-4cee-80a7-2569df0fdd84 because it does not exist
    at LayoutManager.updateLayout (index.ts?b144:284)
    at async LayoutManager.descriptor.value (index.ts?b144:140)
eval @ index.tsx?9ba6:186
Promise.catch (async)
eval @ index.tsx?9ba6:185
setTimeout (async)
eval @ index.tsx?9ba6:179
changePanelLayout @ index.tsx?9ba6:247
eval @ PanelLayout.tsx?06db:191
onChange @ PanelLayout.tsx?06db:170
MosaicWithoutDragDropContext._this.replaceRoot @ Mosaic.js?a997:68
MosaicWithoutDragDropContext._this.updateRoot @ Mosaic.js?a997:64
MosaicRoot._this.onResize @ MosaicRoot.js?7b63:44
onChange @ MosaicRoot.js?7b63:79
eval @ Split.js?47cd:71
invokeFunc @ debounce.js?44db:95
leadingEdge @ debounce.js?44db:105
debounced @ debounce.js?44db:172
Split._this.onMouseMove @ Split.js?47cd:66
index.ts?b144:474 [../../packages/studio-base/src/services/LayoutManager/index.ts] Starting layout sync
index.ts?b144:480 [../../packages/studio-base/src/services/LayoutManager/index.ts] Completed sync in 0.14s
LayoutManagerProvider.tsx?c674:70 [../../packages/studio-base/src/providers/LayoutManagerProvider.tsx] Waiting 3.54 sec for next sync {failures: 0}
index.ts?b144:474 [../../packages/studio-base/src/services/LayoutManager/index.ts] Starting layout sync
index.ts?b144:480 [../../packages/studio-base/src/services/LayoutManager/index.ts] Completed sync in 0.12s
LayoutManagerProvider.tsx?c674:70 [../../packages/studio-base/src/providers/LayoutManagerProvider.tsx] Waiting 7.10 sec for next sync {failures: 0}
```

foxymiles:
**User-Facing Changes**
This improves our handling of remote deletion of our active layout.

**Description**
This listens for notification that our active layout has been deleted and responds by showing a warning message and deselecting the layout.

Fixes #1996


packages/studio-base/src/services/LayoutManager/index.ts
```diff
@@ -591,6 +595,7 @@ export default class LayoutManager implements ILayoutManager {
                 return;
               }
               await local.delete(localLayout.id);
+              this.notifyDeleteListeners({ deletedLayoutId: localLayout.id });
```
jtbandes, line 598: should we also do this in other cases where we delete from the local store?

foxymiles: We seem to already handle the case where the user deletes the active layout in their current session. In what other cases could this happen?

jtbandes: tangentially: we've got some latent bugs around the theme of running multiple studio instances at once (multiple windows in the browser, multiple windows on macos or multiple app instances on other desktop platforms). The two instances don't really notify each other when layouts are deleted from disk.

foxymiles: I had an earlier approach to this that's a bit less targeted that might catch more cases:

https://github.com/foxglove/studio/pull/2286/commits/74a695a4725e75ad52d1347ef6af93be9cdae79a

jtbandes: > We seem to already handle the case where the user deletes the active layout in their current session. In what other cases could this happen?

I think the UI probably handles it ok, just was thinking it's best to be thorough with when we report these notifications so in case something else depends on them or the user gets the app in an unexpected state, we still handle it as well as we can.

packages/studio-base/src/providers/CurrentLayoutProvider/index.tsx
```diff
@@ -225,6 +225,22 @@ export default function CurrentLayoutProvider({
...
   }, [layoutManager, setLayoutState]);
 
+  // Make sure our layout still exists after changes. If not deselect it.
+  useEffect(() => {
+    async function listener() {
+      if (layoutStateRef.current?.selectedLayout?.id) {
+        const layout = await layoutManager.getLayout(layoutStateRef.current?.selectedLayout?.id);
+        if (!layout) {
+          await setSelectedLayoutId(undefined);
+          addToast("Your active layout was deleted.", { appearance: "warning" });
```
jtbandes, line 235: Do we have access to the layout name here? Might be nice to include that in the toast that for a slightly better user experience.

jtbandes (approved):

packages/studio-base/src/providers/CurrentLayoutProvider/index.tsx
foxymiles: Good question. As far as I can tell we don't have the name readily accessible here.

Issue #1261, opened 2021-06-17
Node Playground should support publishing common ROS datatypes even if they're not in the bag


jtbandes:
Split from #726 

The common datatypes introduced in #1038 should be exposed to Node Playground code so that it can use common ROS datatypes even if they are not in the bag/data source.

There is currently a hack specifically to allow publishing `visualization_msgs/MarkerArray`: 

https://github.com/foxglove/studio/blob/860b4df2cad85a39bca022749f80a0a48c6a8796/app/players/UserNodePlayer/nodeTransformerWorker/typescript/ast.ts#L322-L323

jtbandes:
I just discovered there is some logic that translates `a_msgs__B` to `a_msgs/B` in the transformation code: https://github.com/foxglove/studio/blob/ffe3c722d645ede0e7f81d69002261bd59762c60/packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/transformer.ts#L365-L368

bunnyc1986:
I found the fix it to add `basicDatatypes` into the `generateRosLib`'s `datatypes` Map

https://github.com/foxglove/studio/blob/main/packages/studio-base/src/players/UserNodePlayer/index.ts#L578-L581

to 
```
    const rosLib: string = await transformWorker.send("generateRosLib", {
      topics,
      datatypes: new Map([...basicDatatypes, ...datatypes]),
    });
```

PR #2289, opened 2021-12-06
Add basicDatatypes to UserNodePlayer roslib

defunctzombie:
**User-Facing Changes**
User node authors can output messages of any basic datatype even if the datatype is not present in the datasource.

**Description**
User nodes are only able to output known datatypes. Prior to this change, known datatypes would be read from the current data source. This limited the types of messages you can output from user nodes to the datatypes in the data source.
    
This change adds the basic datatypes to the known datatypes for user nodes. This allows a user node author to output any of the basic datatypes even if not present in the data source.
    
Fixes: #1261

### DIFF
--- a/packages/studio-base/src/players/UserNodePlayer/index.ts
+++ b/packages/studio-base/src/players/UserNodePlayer/index.ts
@@ -583,7 +583,9 @@ export default class UserNodePlayer implements Player {
     const transformWorker = this._getTransformWorker();
     const rosLib: string = await transformWorker.send("generateRosLib", {
       topics,
-      datatypes,
+      // Include basic datatypes along with any custom datatypes.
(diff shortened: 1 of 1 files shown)


defunctzombie:
Should I remove the hack on https://github.com/foxglove/studio/blob/860b4df2cad85a39bca022749f80a0a48c6a8796/app/players/UserNodePlayer/nodeTransformerWorker/typescript/ast.ts#L322-L323 ?

Whats a way I can test this? Write a node that outputs a type not in the datasource I suppose?

defunctzombie:
We've determined to leave the hack in and remove in another PR. And confirmed with manual writing of nodes that a datatype not present in the data source could still be used as an output type.

Issue #2133, opened 2021-11-10
Swapping out a panel produces panel with no settings


jtbandes:
**Steps To Reproduce**
- Add a 3D panel. Click ⚙️ > Panel Settings and observe defaults are present (2/3 switches are on by default).
- Click ⚙️ > Change panel > 3D. Look at the new panel's settings.

**Expected Behavior**
The swapped-in 3D panel should have the same defaults set.

**Actual Behavior**
The new panel has a config of `{}`, which for some reason does not get merged with the 3D panel's defaultConfig.

If swapping in a new Tab panel, this bug causes the newly added tab's "Pick a panel" to not work.

amacneil:
Let's also prevent users from changing their current panel to the same panel type.

defunctzombie:
Did this get looked at? I recall us looking at it but wasn't sure if this rolled into any of the various layout PRs that may have fixed this.

I tried reproducing by swapping in a new tab panel and was able to use the "pick a panel" feature of the tab panel and was able to pick a panel.

jtbandes:
That particular Tab issue was fixed by https://github.com/foxglove/studio/pull/2132

PR #2296, opened 2021-12-07
Set panel default config when config is empty

defunctzombie:
**User-Facing Changes**
When a user swaps a panel for another and then opens the panel settings they will see the correct defaults.

**Description**
Previously the Panel wrapper would set the default config only if the config was undefined. The config is sometimes defined but as an empty object (panel swap). In this instance the default config should also be set so the panel and panel settings receives the default values.
    
Fixes: #2133

### DIFF
--- a/packages/studio-base/src/components/Panel.tsx
+++ b/packages/studio-base/src/components/Panel.tsx
@@ -326,12 +326,15 @@ export default function Panel<
     const [savedConfig, saveConfig] = useConfigById<Config>(childId);
 
     // PanelSettings needs useConfigById to return a config
-    // If there is no saved config, we save the default config provided by the panel.
-    // This typically happens when a new panel is added and the layout does not yet have a config.
-    // Even if this effect gets run more than once, we only need to save the default config once.
+    // If there is no saved config (or it is an empty object), we save the default config provided
+    // by the panel. This typically happens when a new panel is added and the layout does not yet
+    // have a config. Even if this effect gets run more than once, we only need to save the default
+    // config once.
+    //
+    // An empty object can occur when swapping a panel
     const savedDefaultConfig = useRef(false);
     useLayoutEffect(() => {
-      if (!savedConfig && !savedDefaultConfig.current) {
+      if ((!savedConfig || Object.keys(savedConfig).length === 0) && !savedDefaultConfig.current) {
         savedDefaultConfig.current = true;
         saveConfig(defaultConfig);
       }


defunctzombie:
The reason the config is an empty object is because it is passed here: https://github.com/foxglove/studio/blob/main/packages/studio-base/src/components/PanelToolbar/index.tsx#L181 when the swap action is performed. Changing this to `undefined` has much broader impact on the code so I opted for the workaround fix.

packages/studio-base/src/components/Panel.tsx
```diff
@@ -326,12 +326,15 @@ export default function Panel<
...
+    // If there is no saved config (or it is an empty object), we save the default config provided
+    // by the panel. This typically happens when a new panel is added and the layout does not yet
+    // have a config. Even if this effect gets run more than once, we only need to save the default
+    // config once.
+    //
+    // An empty object can occur when swapping a panel
     const savedDefaultConfig = useRef(false);
     useLayoutEffect(() => {
-      if (!savedConfig && !savedDefaultConfig.current) {
+      if ((!savedConfig || Object.keys(savedConfig).length === 0) && !savedDefaultConfig.current) {
```
jtbandes, line 337: This fixes one particular instance of the problem but a more general issue is that _any key_ may be missing. Prior to some changes around panel loading earlier this year, the contract for panels with defaultConfig was that the defaultConfig would be shallowly merged into the saved config (like react's defaultProps). Since that contract was changed we have never really gotten back to true merging behavior. Is this something we can hope to achieve or are we just working with a new API contract nowadays (and recommend panels merge in their own default config — in which case I would advocate for completely removing this code block)?

defunctzombie: I think we need to move to a world where panels deal with their own configuration. The challenge we have around that is how panels specify the configuration value so settings can access it.

Whats interesting is that this is just for the saved config and by proxy the settings, the actual config the panel receives is handled below where default config is spread and savedConfig is spread over it.

jtbandes: Interesting, didn't realize the panel is already receiving the merged config. I thought it wasn't because we had that bug with Tab panels (fixed in https://github.com/foxglove/studio/pull/2132), but I now realize that was because the layout reducers didn't receive default panel config, not a problem with the Tab panel itself.

So the real request here is to do the same spreading of defaultConfig in settings that we do when rendering the panel. Why can't we spread defaultConfig there instead of saving it here? Is the argument just about extension panels and we don't want to make defaultConfig a first class concept in extensions?

defunctzombie: Settings doesn't have access to the defaultConfig last I remember. And yea - there is also a divide between extensions (which don't have any settings yet) and our built-in panels. We are starting to get requests that would benefit from settings for extensions.

I think we should merge this fix since we know that swapping a panel produces an empty config object currently. We separately need to think through settings for extensions and builtin panels and find a cohesive strategy.

jtbandes (approved):

Issue #2236, opened 2021-12-01
Plots reset when topic is added/removed to WebSocket player


jtbandes:
When using the WebSocket player, if data is being plotted and then a topic is dynamically added, the current state of the plot is cleared.

This bug might also affect the Rosbridge player and live connections.

It might be that the message pipeline is being reset (e.g. if lastSeekTime changes).

defunctzombie:
👍

PR #2298, opened 2021-12-07
Keep Plot data when topics are added/removed

jtbandes:
**User-Facing Changes**
Fixes an issue where Plot panels would reset when topics were dynamically added/removed.

**Description**
Fixes https://github.com/foxglove/studio/issues/2236


defunctzombie:
👍

PR #2305, opened 2021-12-07
Fix PoseStamped rendering

jhurliman:
**User-facing changes**

Fixes a crash when rendering `PoseStamped` topics in the 3D panel.

**Description**

Add a header to the generated Marker for PoseStamped topics


defunctzombie:
Can you make a story for this so we know if rendering regressed in the future?

Issue #2220, opened 2021-11-25
JSON encoding and JSON Schema support for websocket player


jtbandes:
The Foxglove WebSocket player should recognize `encoding: "json"` and interpret the `schema` as [JSON Schema](https://json-schema.org).

amacneil:
Bonus points if the schema can be optional. But if requiring a schema is easier then let's ship that first.

PR #2308, opened 2021-12-07
WebSocket: json encoding support

jtbandes:
**User-Facing Changes**
None (feature flagged)

**Description**
`encoding: "json"` is now supported in the WebSocket player, with the `schema` being interpreted as [JSON Schema](https://json-schema.org). Binary data is supported using `{ "type": "string", "contentEncoding": "base64" }`.

Fixes https://github.com/foxglove/studio/issues/2220


defunctzombie:
How hard will it be to change the schema to something else in the future if we go with encoding `json` referring to json schema?

jtbandes:
What do you mean by "change to something else"? Are you suggesting that we might want to choose a different schema representation for `json`? JSON Schema seems like a pretty solid choice given its widespread use, but I suppose there is always a chance we will run into some limitation.

defunctzombie:
Yea - maybe json schema isn't the best format (or the only format we want to consider). https://jsontypedef.com/ is another example that is simpler than json schema.

Issue #2309, opened 2021-12-07
List of recently opened files and connections


2metres:
As a user I would like to be able to quickly re-open recent files and connections.

Related to #2295

PR #2312, opened 2021-12-08
Add recents interface to PlayerSelectionContext

defunctzombie:
**User-Facing Changes**
None. No UI for recents is exposed yet.

**Description**
PlayerSelectionContext exposes the list of recents and a method to select a recent. Providers of PlayerSelectionContext implement these as appropriate.

PlayerManager implements PlayerSelectionContext. Recents are stored in indexeddb to support storing file and directory handles.

Fixes: #2309

### DIFF
--- a/desktop/main/StudioWindow.ts
+++ b/desktop/main/StudioWindow.ts
@@ -162,18 +162,20 @@ function buildMenu(browserWindow: BrowserWindow): Menu {
       submenu: [
         { role: "about" },
         { type: "separator" },
+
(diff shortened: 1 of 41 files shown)


2metres:
Can we not display the Recents in the connections for now given that we are planning a larger change? Maybe an item in the OS file menu Item?

defunctzombie:
Removed display of recents in the sidebar. The recents are tracked but not displayed yet.

PR #2317, opened 2021-12-08
Update React to 18.0.0-rc.0

jhurliman:


defunctzombie:
Per some recent discussions we've decided to hold off on upgrading to v18 until it officially releases and has had more time to bake.

defunctzombie:
We've decided to wait on this until react 18 as been out for a few months and has the initial kinks worked out in community adoption.

PR #2319, opened 2021-12-08
Render 3D panel in the selected frame instead of the root frame

jhurliman:
**User-Facing Changes**

The 3D scene world space can be set to a non-root frame. This allows rendering objects even when a transform to the root frame is missing, and fixes precision issues when a coordinate frame is far away (~1000km) from the root frame.

**Description**

Most of the work for this was done in the transform interpolation PR. This just replaces the concept of `rootTf` with `renderFrameId`, allowing any arbitrary frame to be used as the render frame.

### DIFF
--- a/packages/studio-base/src/panels/ThreeDimensionalViz/threeDimensionalVizUtils.ts
+++ b/packages/studio-base/src/panels/ThreeDimensionalViz/threeDimensionalVizUtils.ts
@@ -24,13 +24,11 @@ import {
   cameraStateSelectors,
   DEFAULT_CAMERA_STATE,
 } from "@foxglove/regl-worldview";
-import { Time } from "@foxglove/rostime";
 import { GlobalVariables } from "@foxglove/studio-base/hooks/useGlobalVariables";
 import { InteractionData } from "@foxglove/studio-base/panels/ThreeDimensionalViz/Interactions/types";
 import { LinkedGlobalVariables } from "@foxglove/studio-base/panels/ThreeDimensionalViz/Interactions/useLinkedGlobalVariables";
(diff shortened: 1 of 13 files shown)


packages/studio-base/src/panels/ThreeDimensionalViz/threeDimensionalVizUtils.ts
```diff
@@ -45,21 +43,12 @@ function getZoomDistanceFromURLParam(): number | undefined {
...
-    if (frame) {
-      const pose: MutablePose | undefined = emptyPose();
-      if (frame.root().apply(pose, pose, frame, time)) {
-        const p = pose.position;
-        const o = pose.orientation;
-        return { target: [p.x, p.y, p.z], targetOrientation: [o.x, o.y, o.z, o.w] };
-      }
-    }
+  if (typeof followTf === "string" && transforms.hasFrame(followTf)) {
+    return { target: [0, 0, 0], targetOrientation: [0, 0, 0, 1] };
```
jtbandes, line 51: This seems like a significant change. What does it do?

jhurliman: After this code refactor, this function would always evaluate to an identity transform (from `followTf` to `followTf` since `rootTf` is gone). This shortcuts some checks and explicitly returns an identity transform.

jtbandes: The comment just says:
```
// Get the camera target position and orientation
 function getTargetPose(
```

It's unclear what this means now that the function seems to be "getIdentityTransformButOnlySometimes". Can it be refactored a little to be more intuitive?

PR #2322, opened 2021-12-08
Use simulated time (/clock) for receive time in Ros1Player

jhurliman:
**User-facing changes**

Visualizations no longer flicker when connected to a live ROS1 source that publishes `/clock` (simulated time).

**Description**

Large disagreements between receive time and header stamps caused various data pipeline issues. Use `/clock` when available for setting receive time. This also fixes a "Bag went back in time" warning by rewriting the `receiveTime` of `/clock` messages after processing the new clock time.

### DIFF
--- a/packages/studio-base/src/players/Ros1Player.ts
+++ b/packages/studio-base/src/players/Ros1Player.ts
@@ -9,13 +9,7 @@ import { Sockets } from "@foxglove/electron-socket/renderer";
 import Logger from "@foxglove/log";
 import { RosNode, TcpSocket } from "@foxglove/ros1";
 import { RosMsgDefinition } from "@foxglove/rosmsg";
-import {
-  Time,
-  add as addTimes,
-  fromMillis,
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/players/Ros1Player.ts
```diff
@@ -563,7 +555,7 @@ export default class Ros1Player implements Player {
       }
 
       this._clockTime = time;
-      this._clockReceived = msg.receiveTime;
+      (msg as { receiveTime: Time }).receiveTime = this._getCurrentTime();
```
defunctzombie, line 558: Is this re-writing the receiveTime of the received message?

Mostly noticed this cause you had to cast `msg` ? I think its already a MessageEvent - is receiveTime readonly there?

jhurliman: Yes, it's overcoming a `readonly`

PR #2326, opened 2021-12-08
Remove path requirement for deep link.

### LINKED ISSUE #2299: load remote bag from url state fails
**Description**
The recently introduced feature (https://github.com/foxglove/studio/pull/2171, thank you @foxymiles) to represent the current state (bag+layout+time) in the url misbehaves in my first tests. I always get redirected to the welcome scenario at 
`https://example.org/foxglove/?ds=ros1-remote-bagfile&ds.url=https%3A%2F%2Fstorage.googleapis.com%2Ffoxglove-public-assets%2Fdemo.bag&layoutId=bff4eafa-77d2-4c0f-87a1-a19731fa887b&time=2017-03-22T02%3A26%3A20.202843112Z` instead of the desired remote bag. 

I use the docker container for a self hosted version and the only adjustion I made is to host at a subpath `/foxglove`.

- Version: 0.23.0, self hosted
- OS: ubuntu20.04

**Steps To Reproduce**
In the start up view, load a new remote bag via the panel -> bag loads successfully -> url gets replaced.
Then copy that url and paste it into a new chrome session -> bag and view shows up for a second but then the welcome layout + bag loads and overwrite the state. 

Do I use it wrong or what could be the issue here?
Thanks

foxymiles:
**User-Facing Changes**
This removes the path requirement from deep link state restoration.

**Description**
Currently we only restore state from URLs at the root path, `/`. This removes that restriction.

This should address #2299

### DIFF
--- a/packages/studio-base/src/util/appURLState.ts
+++ b/packages/studio-base/src/util/appURLState.ts
@@ -61,10 +61,6 @@ export function parseAppURLState(url: URL): AppURLState | undefined {
     throw Error("Unknown protocol.");
   }
 
-  if (!isDesktopApp() && url.pathname !== "/") {
-    throw Error("Unknown path.");
-  }
-
   const layoutId = url.searchParams.get("layoutId");
   const timeString = url.searchParams.get("time");
   const time = timeString == undefined ? undefined : fromRFC3339String(timeString);


esthersweon:
Friendly reminder to add `docs` label :)

foxymiles:
I don't think this particular one should affect any documentation.

PR #2333, opened 2021-12-09
Remove "ROS" text from <Timestamp>, use app setting for timestamp format

jhurliman:
**User-Facing Changes**

Data Source Info panel Start / End timestamps use the app setting for timestamp display format

Screen Shot 2021-12-08 at 8 44 39 PM
Screen Shot 2021-12-08 at 8 44 48 PM


2metres:
I've mostly refactored the JSX in #2332 any chance you want to make that change there instead?

jhurliman:
@2metres done: https://github.com/foxglove/studio/pull/2332/commits/fa9d8d086da81d14b9323d2aa70ad4f7581303de

PR #2344, opened 2021-12-10
Fix ImageMarker colors to use [0,1] range

jhurliman:
**User-facing changes**

`ImageMarker` colors now use the range [0,1] instead of [0,255] to align with color handling in `Marker` topics


defunctzombie:
This is a breaking change for those that previously used ImageMarkers?

Are there any docs on ImageMarkers that indicated we were doing the wrong thing? Its more conventional to specify rgb color in [0,255] is it not?

jtbandes:
Reminder to add a [docs] label for changes that require updating docs

PR #2345, opened 2021-12-10
Automatically select an available image topic when opening a new Image panel

jtbandes:
**User-Facing Changes**
A newly opened Image panel now automatically displays images if any are available.


amacneil:
👍 nice

PR #2346, opened 2021-12-10
Update actions/stale to v4

jtbandes:
This might make the actions bot better about removing the stale label when there have been new comments. See: https://github.com/actions/stale/pull/519


amacneil:
Cool - should update for rosbridge_suite too.

PR #2350, opened 2021-12-10
Patch to avoid storing recents in indexeddb

defunctzombie:
**User-Facing Changes**
Recents are not visually exposed to users.

**Description**
Currently File instances are stored in the database which is undesireable since it will store copies of potentially large files. This change always sets recents to an empty list until we figure out the strategy for recents.

### DIFF
--- a/packages/studio-base/src/util/IndexedDbRecentsStore.ts
+++ b/packages/studio-base/src/util/IndexedDbRecentsStore.ts
@@ -40,8 +40,11 @@ export class IndexedDbRecentsStore {
   }
 
   /** Save the recents into the store */
-  async set(recents: RecentRecord[]): Promise<void> {
-    await idbSet(this.key, recents, this.store);
+  async set(_recents: RecentRecord[]): Promise<void> {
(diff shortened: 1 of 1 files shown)


defunctzombie:
@amacneil We need to make a patch release with this fix. The indexeddb storage would incorrectly store File instances which are full copies of the files.

Related - I can't tell if the recents PR from before was in the patch release since the version bump commit is not present on main - can whatever process is used to make a release please ensure that happens? It's hard to know which commits have landed into a release.

amacneil:
> @amacneil We need to make a patch release with this fix. The indexeddb storage would incorrectly store File instances which are full copies of the files.

Building a release now.

> Related - I can't tell if the recents PR from before was in the patch release since the version bump commit is not present on main - can whatever process is used to make a release please ensure that happens? It's hard to know which commits have landed into a release.

Yes can automate the release branch merge. However, it's still not going to give you 100% reliability since git history is not linear. The best way to see which commits are in a release is with compare changes, e.g. https://github.com/foxglove/studio/compare/v0.23.1...main

PR #2351, opened 2021-12-11
Move coordinate frame selection into TopicTree

defunctzombie:
**User-Facing Changes**

**Description**
The coordinate frame determines how the 3d panel resolves and renders visual elements. The existing UX has frame selection in a different part of the panel than marker selection and hidden by default. This makes it non-obvious to users what steps to take to change their coordinate frame when they select different markers to visualize.
    
This change moves the coordinate frame selection into the topic tree near marker selection. The existing frame selection widget is simplified to only the three-way toggle button for selecting the camera behavior.

### DIFF
--- a/packages/studio-base/src/panels/ThreeDimensionalViz/FollowTFControl.tsx
+++ b/packages/studio-base/src/panels/ThreeDimensionalViz/FollowTFControl.tsx
@@ -12,114 +12,20 @@
 //   You may not use this file except in compliance with the License.
 
 import { IButtonStyles, IconButton, Stack, useTheme } from "@fluentui/react";
-import { sortBy, debounce } from "lodash";
-import { memo, useCallback, useState, useMemo, useRef } from "react";
-import shallowequal from "shallowequal";
+import { memo, useCallback, useMemo } from "react";
 
-import Autocomplete, { IAutocomplete } from "@foxglove/studio-base/components/Autocomplete";
 import { useTooltip } from "@foxglove/studio-base/components/Tooltip";
 import { colors } from "@foxglove/studio-base/util/sharedStyleConstants";
 
-import { TransformTree, CoordinateFrame } from "./transforms";
-
-type TfTreeNode = {
-  tf: CoordinateFrame;
-  children: TfTreeNode[];
-  depth: number;
-};
-
-type TfTree = {
-  roots: TfTreeNode[];
-  nodes: {
-    [key: string]: TfTreeNode;
-  };
-};
-
-const treeNodeToTfId = (node: TfTreeNode) => node.tf.id;
-
-const buildTfTree = (transforms: CoordinateFrame[]): TfTree => {
-  const tree: TfTree = {
-    roots: [],
-    nodes: {},
-  };
-  // Create treeNodes for all tfs.
-  for (const tf of transforms) {
-    if (tree.nodes[tf.id]) {
-      continue;
-    }
-    tree.nodes[tf.id] = {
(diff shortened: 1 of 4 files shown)


defunctzombie:
Some thoughts on field names:

- `Coordinate frame` doesn't convey enough context around what impact this selection has. In robotics Coordinate Frame is typically a term referring to each of the items in the list.
- `Render frame` sounds more like "here's the only coordinate frame to render"
- `Camera frame` could be confusing with "tell me what your camera's coordinate frame is"
- `Viewport frame` maybe also suffers from the same issue as `Render Frame`.
- `Fixed frame` a specific concept in robotics usually referring to the stationary frame. For our panel we don't really care about knowing this for the user's robot?
- `Follow frame` maybe conflicts with the three-way toggle

@jhurliman Does our transform interpolation support resolving between two frames with different parents but common grandparents? i.e with a tree of (a -> b - > c & a -> d -> e), if we select "e" as the follow frame, are markers in "c" and "b" resolvable and renderable since they have a common ancestor "a"?

amacneil:
What does 'auto' do?

jtbandes:
Seems like rviz uses "target frame"? http://wiki.ros.org/rviz/UserGuide#The_Target_Frame

jtbandes:
> The existing frame selection widget is simplified to only the three-way toggle button for selecting the camera behavior.

I don't think we should separate the follow toggle from the frame selector. Their behavior is inherently linked (clicking the button toggles the selected frame between "nothing" and its previously selected value)

defunctzombie:
> I don't think we should separate the follow toggle from the frame selector

I don't find it obvious that a user discovers this hidden/expanding button is an important aspect of the viewport display.

defunctzombie:
> (clicking the button toggles the selected frame between "nothing" and its previously selected value)

My understanding is that the states are:

- follow the selected frame
- don't follow the selected frame (the selected frame is still used to render tho)
- follow orientation of the selected frame

So having a specific selected frame is still important regardless of the follow state.

jtbandes:
> don't follow the selected frame (the selected frame is still used to render tho)

This doesn't fit with my understanding of what the button does (or at least how it was originally intended to work). "Not following the selected frame" means the camera has to be fixed to some other frame, i.e. the background/map/world frame. The action of toggling from following a frame to _not_ following a frame should adjust the camera offset (`targetOffset` in code) so that it points at the previously-followed frame, but if that frame moves in the future relative to the world frame, the camera position should no longer move. So in that sense there is no longer any frame selected.

The issue of _which_ frame to use as the fixed/world frame is also a difficult one, so maybe we need to just say that we need to scrap the 3-state toggle and only provide the dropdown, so you'd have to explicitly select which frame you want to use as the world frame. Then it might be weird to keep the "adjust the camera offset" behavior I described above by default ... or we could keep it as long as we provide a "reset to origin" button.

We do need to rethink "follow" and "follow orientation" in general because the way they works implies two frames: the explicitly selected frame and the implicitly chosen root/map frame. Maybe it's better to just give people explicit selectors for both, like rviz does.

defunctzombie:
From a discussion we also noted it would be interesting to be able to follow anything with a pose.

jhurliman:
Yes, transforms only need a common ancestor.

defunctzombie:
closing this. After discussion we decided that we will make some changes to the follow toggle UI first and infer the fixed frame from the follow frame (as we did before tf interpolation changes). We might revisit alternative follow UI approaches once we fix those outstanding issues.

PR #2353, opened 2021-12-13
Refactor ImageMarker rendering

jhurliman:
**User-facing changes**

Several improvements to `ImageMarker` handling:
- Remove proprietary `thickness` attribute from ImageMarker
- Support outline_colors for LINE_LIST
- Support both outlines and fill colors whenever possible
- Observe the `filled` attribute

**Description**

Only a limited number of image marker use cases were previously supported. Part of the issue was a lack of reference implementation since image markers are not supported by rviz. Another issue was the webviz proprietary attribute `thickness`, which always evaluated to undefined since there is no thickness field in `visualization_msgs/ImageMarker`. There is some amount of creative intrepretation in how scale, fill_color, and outline_color(s) are used for the different marker types but overall the approach is to be as forgiving as possible since image markers are poorly documented and not rendered by any other (known) application

Screen Shot 2021-12-12 at 6 00 24 PM

### DIFF
--- a/packages/studio-base/src/panels/ImageView/renderImage.ts
+++ b/packages/studio-base/src/panels/ImageView/renderImage.ts
@@ -405,13 +405,17 @@ function paintMarker(
     case ImageMarkerType.TEXT: {
       // TEXT (our own extension on visualization_msgs/Marker)
       const { x, y } = maybeUnrectifyPoint(cameraModel, marker.position);
+      const text = marker.text?.data ?? "";
+      if (!text) {
+        break;
+      }
 
       const fontSize = marker.scale * 12;
       const padding = 4 * marker.scale;
       ctx.font = `${fontSize}px sans-serif`;
       ctx.textBaseline = "bottom";
       if (marker.filled) {
-        const metrics = ctx.measureText(marker.text.data);
+        const metrics = ctx.measureText(text);
         const height =
           "fontBoundingBoxAscent" in metrics
             ? metrics.fontBoundingBoxAscent + metrics.fontBoundingBoxDescent
@@ -420,7 +424,7 @@ function paintMarker(
         ctx.fillRect(x, y - height, Math.ceil(metrics.width + 2 * padding), Math.ceil(height));
       }
       ctx.fillStyle = toRGBA(marker.outline_color);
-      ctx.fillText(marker.text.data, x + padding, y);
+      ctx.fillText(text, x + padding, y);
       break;
     }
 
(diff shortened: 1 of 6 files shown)


packages/studio-base/src/panels/ImageView/renderImage.ts
```diff
@@ -312,117 +306,194 @@ function paintMarker(
...
+        // to marker.outline_color as needed
+        const outlineColor = hasExactColors
+          ? marker.outline_colors[i / 2]!
+          : marker.outline_colors.length > i
+          ? marker.outline_colors[i]!
+          : marker.outline_color;
+        paintLine(
+          ctx,
+          marker.points[i]!,
+          marker.points[i + 1]!,
```
jtbandes, line 350: The `maybeUnrectifyPoint` calls were removed — seems like a regression?

packages/studio-base/src/panels/ImageView/renderImage.ts
```diff
@@ -312,117 +306,194 @@ function paintMarker(
...
+            ? marker.outline_color
+            : marker.fill_color;
+        paintCircle(ctx, point, marker.scale, marker.scale, undefined, fillColor, cameraModel);
       }
       break;
     }
 
-    case 5: {
+    case ImageMarkerType.TEXT: {
       // TEXT (our own extension on visualization_msgs/Marker)
```
jtbandes, line 406: why did we remove thickness but not text?

packages/studio-base/src/panels/ImageView/renderImage.ts
jtbandes: I see it's inside paintLine

jtbandes (commented):
👍

packages/studio-base/src/types/Messages.ts
```diff
@@ -373,12 +379,28 @@ export type TF = Readonly<
...
+  POLYGON = 3,
+  POINTS = 4,
+  // TEXT is not part of visualization_msgs/ImageMarker, but we include it to
+  // support existing frameworks that have extended this message definition
+  TEXT = 5,
+}
+
+export enum ImageMarkerAction {
+  ADD = 0,
+  REMOVE = 1,
```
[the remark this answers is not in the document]
jhurliman: I just assumed these were supported. If they're not, implementing support for them might be a better use of time than surfacing warnings but I'm not tuned in to how difficult it would be.

PR #2361, opened 2021-12-14
Style tweaks to connection sidebar based on feedback

2metres:
**User-Facing Changes**
Screen Shot 2021-12-14 at 3 34 16 pm

**Description**
Use plus button instead of button


defunctzombie (changes requested):
The new `plus` button is visible even when the open dialog feature is disabled. We should hide it unless the open dialog feature is enabled.

defunctzombie (approved):

PR #2362, opened 2021-12-14
Add sample data source

defunctzombie:
**User-Facing Changes**
Part of the new Open dialog experience and not enabled by default.

A new "Explore sample data" button appears on the open dialog. When clicked it loads the sample dataset and layout.

**Description**
A sample data source provides an initialization function to create a player and a layout. When selected, the player is initialized and the layout is loaded to the workspace.

### DIFF
--- a/packages/studio-base/src/Workspace.tsx
+++ b/packages/studio-base/src/Workspace.tsx
@@ -193,19 +193,10 @@ export default function Workspace(props: WorkspaceProps): JSX.Element {
   // when the user wants to select a new connection we track whether the sidebar item opened
   const userSelectSidebarItem = useRef(false);
 
-  const selectSidebarItem = useCallback(
-    (item: SidebarItemKey | undefined) => {
-      userSelectSidebarItem.current = true;
-      setSelectedSidebarItem(item);
-
-      // When activating the connection sidebar and no player is present we automatically trigger
-      // the open dialog.
-      if (item === "connection" && !isPlayerPresent) {
-        setShowOpenDialog(true);
-      }
-    },
-    [isPlayerPresent],
-  );
+  const selectSidebarItem = useCallback((item: SidebarItemKey | undefined) => {
+    userSelectSidebarItem.current = true;
+    setSelectedSidebarItem(item);
+  }, []);
 
   // When a player is activated, hide the open dialog. When a player is gone, show the open dialog.
   useLayoutEffect(() => {
--- a/packages/studio-base/src/Workspace.tsx
+++ b/packages/studio-base/src/Workspace.tsx
@@ -193,10 +193,24 @@ export default function Workspace(props: WorkspaceProps): JSX.Element {
   // when the user wants to select a new connection we track whether the sidebar item opened
   const userSelectSidebarItem = useRef(false);
 
-  const selectSidebarItem = useCallback((item: SidebarItemKey | undefined) => {
-    userSelectSidebarItem.current = true;
-    setSelectedSidebarItem(item);
-  }, []);
+  const selectSidebarItem = useCallback(
+    (item: SidebarItemKey | undefined) => {
+      userSelectSidebarItem.current = true;
+      setSelectedSidebarItem(item);
+
+      // When activating the connection sidebar and no player is present we automatically trigger
+      // the open dialog.
+      if (item === "connection" && !isPlayerPresent) {
+        setShowOpenDialog(true);
+      }
+    },
+    [isPlayerPresent],
+  );
+
+  // When a player is activated, hide the open dialog. When a player is gone, show the open dialog.
+  useLayoutEffect(() => {
+    setShowOpenDialog(!isPlayerPresent);
+  }, [isPlayerPresent]);
 
   // Automatically close the connection sidebar when a connection is chosen
   useLayoutEffect(() => {
@@ -214,15 +228,6 @@ export default function Workspace(props: WorkspaceProps): JSX.Element {
       setSelectedSidebarItem(undefined);
     }
   }, [selectedSidebarItem, playerPresence, enableOpenDialog]);
-
-  // When activating the connection sidebar and no player is present we automatically trigger
-  // the open dialog.
-  useLayoutEffect(() => {
-    if (selectedSidebarItem === "connection" && !isPlayerPresent) {
-      setShowOpenDialog(true);
-      return;
-    }
-  }, [isPlayerPresent, selectedSidebarItem]);
 
   const isMounted = useMountedState();
 
--- a/packages/studio-base/src/components/OpenDialog/OpenDialog.tsx
+++ b/packages/studio-base/src/components/OpenDialog/OpenDialog.tsx
@@ -4,10 +4,7 @@
 
 import { Dialog, Stack, useTheme } from "@fluentui/react";
 import { useCallback, useMemo, useState } from "react";
-import { useMountedState } from "react-use";
 
-import { useCurrentLayoutActions } from "@foxglove/studio-base/context/CurrentLayoutContext";
-import { useLayoutManager } from "@foxglove/studio-base/context/LayoutManagerContext";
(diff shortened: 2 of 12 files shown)


packages/studio-base/src/components/OpenDialog/OpenDialog.tsx
```diff
@@ -33,9 +47,28 @@ export default function OpenDialog(props: OpenDialogProps): JSX.Element {
         return;
       }
 
+      if (view === "demo" && firstDemoSource && firstDemoSource.layout) {
+        layoutStorage
```
defunctzombie, line 51: @jtbandes thoughts on this happening here vs in PlayerManager? As I adapted this feature into the existing data source factory concept, it started to feel more natural to probably have the PlayerManager setup the layout. The PlayerManager is a legacy name - and its actual function is PlayerSelectionProvider.

packages/studio-base/src/dataSources/DemoOneDataSourceFactory.ts
```diff
@@ -0,0 +1,190 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
defunctzombie, line 1: @jhurliman @jtbandes Does it make sense to treat demo data as a type of data source? Along with an optional _layout_ field?

I could alternatively create a separate DemoDataSource interface and have them be completely separate types and lists but this felt somewhat natural. As we add more demo data sources, the Open Dialog can populate them as cards.

packages/studio-base/src/Workspace.tsx
```diff
@@ -193,10 +193,24 @@ export default function Workspace(props: WorkspaceProps): JSX.Element {
...
-    userSelectSidebarItem.current = true;
-    setSelectedSidebarItem(item);
-  }, []);
+  const selectSidebarItem = useCallback(
+    (item: SidebarItemKey | undefined) => {
+      userSelectSidebarItem.current = true;
+      setSelectedSidebarItem(item);
+
+      // When activating the connection sidebar and no player is present we automatically trigger
+      // the open dialog.
```
jtbandes, line 202: sounds like a weird user interaction to click a sidebar tab and a modal pops up.

packages/studio-base/src/components/OpenDialog/OpenDialog.tsx
```diff
@@ -33,9 +47,28 @@ export default function OpenDialog(props: OpenDialogProps): JSX.Element {
         return;
       }
 
+      if (view === "demo" && firstDemoSource && firstDemoSource.layout) {
```
jtbandes, line 50: not clear to me as a first time reader what a "view" is. Is it like a "connection type"?

packages/studio-base/src/components/OpenDialog/Start.tsx
```diff
@@ -64,15 +64,13 @@ export default function Start(props: IStartProps): JSX.Element {
...
-      children: "Explore sample data",
-      secondaryText: "New to Studio? Have a play with demo data",
-      iconProps: { iconName: "BookStar" },
-      onClick: () => onSelectView("demo"),
-    },
-    */
+      {
+        id: "demo-data",
+        children: "Explore sample data",
+        secondaryText: "New to Studio? Have a play with demo data",
```
jtbandes, line 70: Maybe this is just an American thing but "have a play" sounds unusual to me.
```suggestion
        secondaryText: "New to Studio? Try out the app with some demo data",
```

packages/studio-base/src/components/PlayerManager.tsx
```diff
@@ -209,6 +209,17 @@ export default function PlayerManager(props: PropsWithChildren<PlayerManagerProp
       metricsCollector.setProperty("player", sourceId);
       setSelectedSource(() => foundSource);
 
+      // Demo sources don't need args or prompts to initialize
+      if (foundSource.type === "demo") {
```
jtbandes, line 213: I find it a little strange that we need special-case code based on source type both in the open dialog and in the PlayerManager. I would think the open dialog's job is to massage the user selection into a single piece of information the player manager will use to create the player. On the other hand I can see how maybe demo sources are special because they require no extra inputs

packages/studio-base/src/context/PlayerSelectionContext.ts
```diff
@@ -29,6 +30,7 @@ export interface IDataSourceFactory {
   disabledReason?: string | JSX.Element;
   badgeText?: string;
   hidden?: boolean;
+  layout?: PanelsState;
```
jtbandes, line 33: Seems kinda wrong, layout is something the user picks and not an attribute of data sources. I might feel less weird about this if it were called something like `sampleLayout` or `defaultLayout`.

If we wanted a feature where we auto-suggest panels when someone adds a new bag, would that also go through this property? In that case maybe it would make sense to have the layout be produced as part of the initialize() method so it can depend on the available topics/datatypes?

packages/studio-base/src/dataSources/DemoOneDataSourceFactory.ts
```diff
@@ -0,0 +1,190 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import {
+  IDataSourceFactory,
+  DataSourceFactoryInitializeArgs,
+} from "@foxglove/studio-base/context/PlayerSelectionContext";
+import { buildPlayerFromBagURLs } from "@foxglove/studio-base/players/buildPlayer";
+import { DEMO_BAG_URL } from "@foxglove/studio-base/util/isDemoBagUrl";
+
+class DemoOneDataSourceFactory implements IDataSourceFactory {
```
jtbandes, line 12: weird name, is it "one" because we plan to add more demo data? I think we should come up with a descriptive name for each demo dataset, e.g. "nuscenes-demo"

defunctzombie: Yea - we plan to add more demo data.

packages/studio-base/src/components/OpenDialog/Start.tsx
defunctzombie: `Try some demo data`?

Try our the app somehow sounds verbose... anyhow - can make this whatever we think is good language.

packages/studio-base/src/components/OpenDialog/OpenDialog.tsx
defunctzombie: view is which of the primary actions you've selected - and the dialog view changes based on the selected action.

packages/studio-base/src/dataSources/DemoOneDataSourceFactory.ts
defunctzombie: Whats your recommendation for a descriptive name for our existing demo bag?

packages/studio-base/src/Workspace.tsx
defunctzombie: Removed per our UX review

packages/studio-base/src/components/OpenDialog/Start.tsx
defunctzombie: Updated the language.

packages/studio-base/src/components/PlayerManager.tsx
defunctzombie: The open dialog special cases it until we have more sample data sources. Then they will appear as cards within a _sample_ view on that dialog. For now the open dialog selects the first sample source.

packages/studio-base/src/context/PlayerSelectionContext.ts
defunctzombie: renamed to `sampleLayout`

packages/studio-base/src/dataSources/DemoOneDataSourceFactory.ts
defunctzombie: Renamed to `SampleUdacityDataSourceFactory` and `Sample: Udacity` as the name.

PR #2372, opened 2021-12-14
Refactor pose handling for OccupancyGrid messages

jhurliman:
**User-facing changes**

- OccupancyGrids will render in the 3D panel with the correct pose regardless of what frame you are viewing
- Transform interpolation works for `PoseStamped` visualizations in the 3D panel

**Description**

OccupancyGrid messages with a frame_id different than the currently viewed frame_id in the 3D panel were not rendering with the correct pose. This is because the OG shader was reading pose from `.info.origin` but also using applyPose, which reads the `.pose` field that is being updated per frame. This removes special transform handling for OG messages and unifies on `.pose`.

PoseStamped markers just needed `frame_locked: true`.

### DIFF
--- a/packages/studio-base/src/panels/ThreeDimensionalViz/SceneBuilder/index.ts
+++ b/packages/studio-base/src/panels/ThreeDimensionalViz/SceneBuilder/index.ts
@@ -68,6 +68,7 @@ const buildSyntheticArrowMarker = (
   header: message.header,
   type: 103,
   pose,
+  frame_locked: true,
   scale: { x: 2, y: 2, z: 0.1 },
   color: getSyntheticArrowMarkerColor(topic),
   interactionData: { topic, originalMessage: message },
@@ -591,7 +592,7 @@ export default class SceneBuilder implements MarkerProvider {
       type,
       name,
       pose: clonePose(info.origin),
-      frame_locked: true,
+      frame_locked: false,
       interactionData: { topic, originalMessage: message },
     };
 
@@ -916,7 +917,7 @@ export default class SceneBuilder implements MarkerProvider {
       case 10: // MeshMarker
       case 11: // TriangleListMarker
       case 102: // PointCloud2
-      case 103: // (unknown!)
+      case 103: // PoseStamped
       case 108: // InstanceLineListMarker
       case 110: // ColorMarker
         marker = { ...marker, pose };
--- a/packages/studio-base/src/panels/ThreeDimensionalViz/SceneBuilder/index.ts
+++ b/packages/studio-base/src/panels/ThreeDimensionalViz/SceneBuilder/index.ts
@@ -186,6 +186,12 @@ function computeMarkerPose(
   return frame.apply(emptyPose(), marker.pose, srcFrame, time);
 }
 
+function poseClone(pose: Pose): MutablePose {
+  const p = pose.position;
+  const o = pose.orientation;
+  return { position: { x: p.x, y: p.y, z: p.z }, orientation: { x: o.x, y: o.y, z: o.z, w: o.w } };
+}
+
 export default class SceneBuilder implements MarkerProvider {
   topicsByName: {
     [topicName: string]: Topic;
@@ -573,10 +579,6 @@ export default class SceneBuilder implements MarkerProvider {
     const type = 101;
     const name = `${topic}/${type}`;
 
-    // set ogrid texture & alpha based on current rviz settings
-    // in the future these will be customizable via the UI
-    const [alpha, map] = this._hooks.getOccupancyGridValues(topic);
-
     const { header, info, data } = message;
     const mappedMessage = {
       header: {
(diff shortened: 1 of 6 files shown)


packages/studio-base/src/panels/ThreeDimensionalViz/SceneBuilder/index.ts
```diff
@@ -936,7 +927,9 @@ export default class SceneBuilder implements MarkerProvider {
       case 110: // ColorMarker
         marker = { ...marker, pose };
         break;
-      case 101: // OccupancyGridMessage - needs special handling
+      case 101: // OccupancyGridMessage
+        marker = { ...marker, pose };
+        break;
       case 104: // LaserScan - needs special handling
```
jhurliman, line 933: `LaserScan` is almost certainly broken. Something for another PR

defunctzombie:
Please add either tests or stories so we know this feature is working after future refactors.

packages/studio-base/src/panels/ThreeDimensionalViz/SceneBuilder/index.ts
defunctzombie: Is it broken by this PR or continues to be broken from other PRs?

jhurliman: I'm assuming it is already broken just by looking at this part of the code. We throw away the transformed `pose` in the LaserScan case so it's unlikely to have the correct pose when rendering a different frame.

packages/studio-base/src/panels/ThreeDimensionalViz/SceneBuilder/index.ts
```diff
@@ -592,25 +588,16 @@ export default class SceneBuilder implements MarkerProvider {
         origin: info.origin,
       },
       data,
-      alpha,
-      map,
       type,
       name,
-      pose: emptyPose(),
+      pose: clonePose(info.origin),
+      frame_locked: true,
```
jtbandes, line 594: Is this what rviz does? I thought the occupancy grid would usually be rendered in the "fixed frame" i.e. `frame_locked: false`. At least I thought that's how it worked at Cruise.

packages/studio-base/src/panels/ThreeDimensionalViz/SceneBuilder/index.ts
```diff
@@ -592,25 +588,16 @@ export default class SceneBuilder implements MarkerProvider {
         origin: info.origin,
       },
       data,
-      alpha,
-      map,
       type,
       name,
-      pose: emptyPose(),
+      pose: clonePose(info.origin),
```
jtbandes, line 593: so where/when does the TF from "render frame" to header.frame_id get applied to this pose?

jhurliman: `computeMarkerPose`

packages/studio-base/src/panels/ThreeDimensionalViz/SceneBuilder/index.ts
jhurliman: Good catch, this is a mistake. Fixing

packages/studio-base/src/panels/ThreeDimensionalViz/commands/OccupancyGrids.stories.tsx
```diff
@@ -18,13 +17,12 @@ export default {
...
 function makeGrid([px, py, pz]: vec3, [ox, oy, oz, ow]: quat) {
   return {
-    pose: emptyPose(),
+    pose: { position: { x: px, y: py, z: pz }, orientation: { x: ox, y: oy, z: oz, w: ow } },
     alpha: 0.8,
     info: {
       resolution: 0.2,
       width: 10,
       height: 10,
-      origin: { position: { x: px, y: py, z: pz }, orientation: { x: ox, y: oy, z: oz, w: ow } },
```
jtbandes, line 27: I don't think we should change this. The OccupancyGrid message does not have a `pose` field, it has an `info` field with an `origin`: https://docs.ros.org/en/noetic/api/nav_msgs/html/msg/OccupancyGrid.html

jhurliman: We don't render `OccupancyGrid` messages, though. We render `OccupancyGrid` messages that have been mutated into marker-like objects. Our pose interpolation code only operates on the `pose` field, which is why `origin` is cloned into `pose` on consumption.

jtbandes: I see, so this is necessary to make the test continue to work, but we don't have a test that shows that SceneBuilder would correctly clone/update the `info` field into the `pose` field...

jhurliman: That's right. We're missing integration tests that incorporate SceneBuilder, which is responsible for transforming incoming messages into "renderables", selecting which renderables will be rendered each time renderMarkers() is called, and doing any per-frame work like updating poses.

jtbandes: We don't have a lot of them, but we do have some. There is a [FixtureExample helper](https://github.com/foxglove/studio/blob/cd5ee2be52a84ff148190fe419f3f5583e42045c/packages/studio-base/src/panels/ThreeDimensionalViz/stories/MarkerStory.tsx#L66), and the PanelSetup can be used directly with a 3D panel. Please add stories or tests that cover this refactored code.

PR #2388, opened 2021-12-15
Add support for recent file

defunctzombie:
**User-Facing Changes**
None. The Open Dialog is feature flagged off by default.

Within the open dialog, the list of recent sources will contain recent files when opening a single file.

**Description**
When a file is opened via the open dialog, we record an entry into recents with the FileSystemFileHandle. This entry can later be clicked to re-open the existing file.
    
On web this feature is restricted to https and may prompt the user to confirm file access before the app is able to read the file again.

### DIFF
--- a/desktop/main/index.ts
+++ b/desktop/main/index.ts
@@ -28,6 +28,9 @@ import { getTelemetrySettings } from "./telemetry";
 
 const log = Logger.getLogger(__filename);
 
+// https://github.com/electron/electron/issues/28422#issuecomment-987504138
+app.commandLine.appendSwitch("enable-experimental-web-platform-features");
+
 /**
  * Determine whether an item in argv is a file that we should try opening as a data source.
  *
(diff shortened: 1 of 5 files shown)


desktop/main/index.ts
```diff
@@ -28,6 +28,9 @@ import { getTelemetrySettings } from "./telemetry";
 
 const log = Logger.getLogger(__filename);
 
+// https://github.com/electron/electron/issues/28422#issuecomment-987504138
+app.commandLine.appendSwitch("enable-experimental-web-platform-features");
```
defunctzombie, line 32: @jtbandes @jhurliman Do we think this is a bad idea? From what I can tell it seems to turn on all the experimental features? I can confirm that without this line present - re-using the FileSystemFileHandles stored in indexeddb is not possible in electron (dev nor packaged).

defunctzombie:
https://bugs.chromium.org/p/chromium/issues/detail?id=1080811

desktop/main/index.ts
jhurliman: Seems reasonable to me. This is a native app, after all.

PR #2390, opened 2021-12-15
Improve handling of point cloud render errors.

### LINKED ISSUE #2186: Point cloud decoding failed error on empty Pointcloud
**Description**
Pointcloud data with empty datafield generate error "Point cloud decoding failed"

- Version: Foxglove Studio v0.22.1
- OS: Ubuntu 18.04

**Steps To Reproduce**

Create a Pointcloud2 msg with the following content:

```
header: 
  seq: 1635729
  stamp: 
    secs: 1637311582
    nsecs: 506327942
  frame_id: "world"
height: 1
width: 0
fields: 
  - 
    name: "x"
    offset: 0
    datatype: 7
    count: 1
  - 
    name: "y"
    offset: 4
    datatype: 7
    count: 1
  - 
    name: "z"
    offset: 8
    datatype: 7
    count: 1
is_bigendian: False
point_step: 16
row_step: 0
data: []
is_dense: False
```
Got the following error on every received message

image

**Expected Behavior**
Reproduce Rviz behaviour: nothing displayed since pointcloud is empty but without errors

**Actual Behavior**
"Point cloud decoding failed" error displayed on every message received

### LINKED ISSUE #2225: PointCloud2 rendering should support `is_bigendian: true`
Currently, the `is_bigendian` field of point cloud topics ~is silently ignored~ throws an error as of #2390. We should add big-endian support, or at least log an error if we encounter a big-endian point cloud that we can't render.

- [x] Add an error to the topic tree when we encounter a big endian point cloud.
- [ ] Enhance our point cloud rendering to render big endian point clouds correctly.

foxymiles:
**User-Facing Changes**
This improves point cloud error handling.

**Description**
The solution is to inspect point cloud messages in `SceneBuilder`. If they are in a big-endian format, which we don't support yet, or have no point data, we register an error on the topic in the topic tree.

Fixes #2186 
Fixes #2225

### DIFF
--- a/packages/studio-base/src/panels/ThreeDimensionalViz/commands/PointClouds/decodeMarker.ts
+++ b/packages/studio-base/src/panels/ThreeDimensionalViz/commands/PointClouds/decodeMarker.ts
@@ -46,12 +46,6 @@ const DEFAULT_COLOR_FIELDS = ["intensity", "i"];
 // Decode a marker and generate position and color buffers for rendering
 // The resulting marker should be memoized for better performance
 export function decodeMarker(marker: PointCloudMarker): DecodedMarker {
-  // Throw an error on big endian data until we support rendering
-  // it correctly.
-  if (marker.is_bigendian) {
-    throw new Error("Unsupported big endian point cloud data.");
-  }
-
   const {
     fields = [],
     settings = {},
--- a/packages/studio-base/src/panels/ThreeDimensionalViz/commands/PointClouds/decodeMarker.ts
+++ b/packages/studio-base/src/panels/ThreeDimensionalViz/commands/PointClouds/decodeMarker.ts
@@ -46,6 +46,12 @@ const DEFAULT_COLOR_FIELDS = ["intensity", "i"];
 // Decode a marker and generate position and color buffers for rendering
 // The resulting marker should be memoized for better performance
 export function decodeMarker(marker: PointCloudMarker): DecodedMarker {
+  // Throw an error on big endian data until we support rendering
+  // it correctly.
+  if (marker.is_bigendian) {
+    throw new Error("Unsupported big endian point cloud data.");
+  }
+
   const {
     fields = [],
     settings = {},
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/panels/ThreeDimensionalViz/commands/PointClouds/decodeMarker.ts
```diff
@@ -46,6 +46,12 @@ const DEFAULT_COLOR_FIELDS = ["intensity", "i"];
 // Decode a marker and generate position and color buffers for rendering
 // The resulting marker should be memoized for better performance
 export function decodeMarker(marker: PointCloudMarker): DecodedMarker {
+  // Throw an error on big endian data until we support rendering
+  // it correctly.
+  if (marker.is_bigendian) {
```
defunctzombie, line 51: Can we detect this earlier (scene builder) and add this error to the errorsByTopic? Or does throwing here produce the same effect?

packages/studio-base/src/panels/ThreeDimensionalViz/commands/PointClouds/index.tsx
```diff
@@ -447,30 +451,56 @@ type Props = CommonCommandProps & {
...
+  const errorCache = useRef(new Set<string>());
+
+  // Clear our error cache when the player changes.
+  useEffect(() => {
+    errorCache.current.clear();
+  }, [playerId]);
+
+  // These errors can occur per-message so we avoid error spam by only showing
+  // the same error message once.
+  // TODO: Find a way to indicate these in the topic tree instead.
```
defunctzombie, line 473: TODO in the code are lost to time and never surfaced. If we can't surface these appropriately today - lets make an issue and link to the issue from the code.

packages/studio-base/src/panels/ThreeDimensionalViz/commands/PointClouds/decodeMarker.ts
foxymiles: Unfortunately it seems difficult to communicate this back to the scene builder at this point. I'll look into detecting it earlier.

PR #2396, opened 2021-12-15
Improve error handling for URDF loading failures

jhurliman:
**User-Facing Changes**

URDF errors will be surfaced as a toast notification instead of a full-screen error dialog

**Description**

Catch and surface URDF loading errors. Stop reporting to Sentry

Screen Shot 2021-12-15 at 12 59 25 PM


packages/studio-base/src/panels/ThreeDimensionalViz/UrdfBuilder.ts
```diff
@@ -140,7 +145,9 @@ export default class UrdfBuilder implements MarkerProvider {
       const res = await fetch(fetchUrl);
       text = await res.text();
```
amacneil, line 146: note from ☕ today: need to check response status code is 2xx

defunctzombie:
Is there a follow-up PR here for the discussions today?

Issue #2397, opened 2021-12-15
URL state sync is broken


amacneil:
**Description**
Loading a remote data source does not update the URL. Possibly broken in https://github.com/foxglove/studio/pull/2385

Repro:
1. Go to studio.foxglove.dev
2. Open a remote bag file
3. Nothing happens in the URL

PR #2406, opened 2021-12-16
Fix url state for data sources using RandomAccessPlayer

defunctzombie:
**User-Facing Changes**
Url state is correctly updated when using data sources which support url state. When switching to a data source that does not support url state, the data source fields are cleared.

**Description**
A recent change to remove the RandomAccessPlayerDescriptors broke url state handling which relied on a _label member of RandomAccessPlayer.
    
This is fixed by having data source factories initialize RandomAccessPlayer with appropriate options if providing a url state is important for the specific data source.
    
Update url state handling to unset data source fields if a data source does not provide state. Data sources which do not provide url state are those which cannot be restored from url state.
    
Fixes: #2397


foxymiles (approved):
Looks good to me!

Issue #2402, opened 2021-12-15
Static transforms being ignored


defunctzombie:
**Description**
With the recent TF interpolation changes, we now ignore tf frames outside of a hard-coded time bound. This results in static transforms (which are only published once) to be ignored.

PR #2407, opened 2021-12-16
Change CoordinateFrame maxDelta behavior to 0 seconds in the past and infinitely far into the future

jhurliman:
**User-Facing Changes**

- Infrequently published transforms such as static transforms no longer expire
- Transforms will never be used ahead of their publish time

**Description**

Previously, we used a `maxDelta` slop factor of one second when checking if a given input time fell within the history of a coordinate frame. This behavior is now changed to not allow any slop factor into the past (transforms must exist at or before the given `time` query parameter) and infinite slop into the future (searching for t = 1000 in a time window that spans [10-20] will return the latest transform at t = 20). Fixes https://github.com/foxglove/studio/issues/2402


defunctzombie:
@jhurliman Please add tests for changes like this

Issue #2395, opened 2021-12-15
Remove claims of supporting package:// URLs on web


amacneil:
**Description**
On web + desktop, for URDF support we claim to support `package://` URLs. This feature only works on desktop.

Using a `package://` URL results in an error silently being logged to the developer console:

https://user-images.githubusercontent.com/637671/146259462-60205d79-9b65-40d8-b075-ca1c88643710.mov

In this dialog, on desktop we can list support for both methods (package and http(s). On web, we should explain that only http(s) is supported and that they can get package support by installing the desktop app.

We should also explain in this dialog how `ROS_PACKAGE_PATH` is important for package URLs and maybe even display the file we are going to try to load.

cc @defunctzombie for thoughts on the ux

defunctzombie:
I don't recall if we have a good way to detect platform capabilities (i.e. whether we can support package:// paths or not). We might have some hard-coded util.js file which might be a decent(ish) place to start with the detection.

defunctzombie:
I've assigned this to myself since John has lots of `tf` and `urdf` related issues to look at already.

PR #2412, opened 2021-12-16
Omit package:// references from URDF settings editor in web

defunctzombie:
**User-Facing Changes**
Users are not shown a hint suggesting package:// urls are valid in the web app.

**Description**
Web does not support loading package:// urls so we do not display this as a hint to the user.

Fixes: #2395


jtbandes:
Might be best to mention that package urls are supported in the desktop app?

defunctzombie:
> Might be best to mention that package urls are supported in the desktop app?

I am torn on this. package:// urls are a ros-ism but it would be useful for ROS users that want to do that to know they can use the desktop app. At a high level I still feel like we should have an asset manager.

@2metres @amacneil thoughts on adding language here to tell people about package:// urls being available on desktop? If that is desired tell me the copy to add and I am happy to add it.

packages/studio-base/src/panels/ThreeDimensionalViz/TopicSettingsEditor/UrdfSettingsEditor.tsx
```diff
@@ -11,31 +11,37 @@
 //   found at http://www.apache.org/licenses/LICENSE-2.0
 //   You may not use this file except in compliance with the License.
 
+import { TextField } from "@fluentui/react";
+
 import Flex from "@foxglove/studio-base/components/Flex";
+import isDesktopApp from "@foxglove/studio-base/util/isDesktopApp";
```
amacneil, line 17: here we go - down the slippery slope

amacneil:
We have a feature that exists in our app, but for technical reasons is currently* only supported on desktop. So I agree with Jacob I think we should advertise the fact that they can use this feature (I guess if you want to get complicated we could show this feature only when you're using a ROS connection, i.e. not Foxglove Websocket? but it seems simpler to just handle the same for now).

So maybe something like:

> http(s) URL pointing to a Unified Robot Description Format (URDF) XML file
>
> [   text box   ]
>
> For ROS users, we also support package:// URLs (loaded from the local filesystem) in our [desktop app](https://foxglove.dev/download).

*I say currently because I still think we could add support for requesting assets over rosbridge at some point

gavanderhoorn:
> I say currently because I still think we could add support for requesting assets over rosbridge at some point

not specifically very fancy, but `ros-sharp` does this using a [file_server](https://github.com/siemens/ros-sharp/tree/6e91309b2447c0919d3dabaa0650bb71e3e490dd/ROS).

amacneil:
@gavanderhoorn interesting, thanks for the link!

packages/studio-base/src/panels/ThreeDimensionalViz/TopicSettingsEditor/UrdfSettingsEditor.tsx
```diff
@@ -11,33 +11,48 @@
...
 
 import { TopicSettingsEditorProps } from ".";
-import { SLabel, SDescription, SInput } from "./common";
 
 export type UrdfSettings = {
   urdfUrl?: string;
 };
 
+// One day we can think about using feature detection. Until that day comes we acknowledge the
+// realities of only having two playforms: web and desktop.
```
jtbandes, line 26: typo: playforms

defunctzombie: https://github.com/foxglove/studio/pull/2467

PR #2415, opened 2021-12-16
Fix unselectable color modes in point cloud settings editor.

### LINKED ISSUE #2373: Unable to change point cloud coloring in 3d panel
**Description**
- `Turbo` and `Rainbow` should be colored appropriately (currently the word `Turbo` uses rainbow coloring)
- When `Turbo` is selected by default, it won't let me select any other settings. The whole form is disabled. This resolves itself if you switch to `Flat` coloring and back again.

https://user-images.githubusercontent.com/637671/146083805-ba4ecd96-cbc8-48ea-b19e-a0afb2c14082.mov

foxymiles:
**User-Facing Changes**
This fixes an issue with color mode selection in the point cloud settings editor.

**Description**
The issue is that we can't offer the user any of the per-point point cloud coloring modes until we've received a message so we know what fields we can allow them to select to drive the coloring. 

The solution in this PR is twofold:
1. Don't allow users to select per-point coloring until we have the message data we need.
2. Add a message to explain to the user why per-point coloring modes are not available.

Fixes #2373


packages/studio-base/src/panels/ThreeDimensionalViz/TopicSettingsEditor/PointCloudSettingsEditor.tsx
```diff
@@ -90,6 +95,10 @@ export type PointCloudSettings = {
 
 const DEFAULT_COLOR_FIELDS = ["intensity", "i"];
 
+function isPointColorMode(mode: ColorMode | undefined): mode is PointColorMode {
+  return mode?.mode === "gradient" || mode?.mode === "rainbow" || mode?.mode === "turbo";
```
jhurliman, line 99: What about rgb?

packages/studio-base/src/panels/ThreeDimensionalViz/TopicSettingsEditor/PointCloudSettingsEditor.tsx
```diff
@@ -30,7 +31,7 @@ import CommonDecaySettings from "./CommonDecaySettings";
 import { SLabel, SInput } from "./common";
 import { turboColorString } from "./turboColor";
 
-export type ColorMode =
+type FlatColorMode =
```
jhurliman, line 34: This isn't the best name because it encompasses both flat coloring and per-point rgb coloring.

jhurliman (commented):
I think this needs to be tested with an RGB point cloud

packages/studio-base/src/panels/ThreeDimensionalViz/TopicSettingsEditor/PointCloudSettingsEditor.tsx
foxymiles: I was just trying to DRY up some logic here since that boolean test was repeated several times. Maybe it's better to call these "mapped" color modes instead of point color modes?

packages/studio-base/src/panels/ThreeDimensionalViz/TopicSettingsEditor/PointCloudSettingsEditor.tsx
```diff
@@ -232,9 +237,14 @@ export default function PointCloudSettingsEditor(
         </Flex>
       </Flex>
 
-      {(colorMode.mode === "gradient" ||
-        colorMode.mode === "rainbow" ||
-        colorMode.mode === "turbo") && (
+      {!pointModeAvailable && (
+        <Text style={{ color: theme.palette.neutralSecondary }}>
+          Point data coloring mode is not available until messages with field data have been
+          received.
```
jtbandes, line 243: I think we might've discussed letting you choose the mode before messages are received, but just not selecting which field to use, and if no field is selected then we'll auto-select one when the first message is received?

cc @amacneil @2metres cause I can't remember what we said in the meeting a couple weeks back

foxymiles:
@jhurliman @jtbandes Maybe this simpler, more targeted fix for this bug is better and we can handle improving the UX and default behavior in a followup?

jtbandes:
This change looks like it should be basically a no-op, how does it fix the bug?

packages/studio-base/src/panels/ThreeDimensionalViz/TopicSettingsEditor/PointCloudSettingsEditor.tsx
```diff
@@ -156,6 +144,15 @@ export default function PointCloudSettingsEditor(
       ? { mode: "turbo", colorField: defaultColorField }
       : { mode: "flat", flatColor: DEFAULT_FLAT_COLOR });
 
+  function onColorModeChange(
+    newValue: (ColorMode | undefined) | ((prevColorMode?: ColorMode) => ColorMode | undefined),
+  ) {
+    onSettingsChange((oldSettings) => ({
+      ...oldSettings,
+      colorMode: typeof newValue === "function" ? newValue(colorMode) : newValue,
```
foxymiles, line 152: @jtbandes The fix is to use the derived `colorMode` here instead of referencing the previous value from settings, which will be `{}` in a fresh panel.

foxymiles: It's `{}`. The main issue is this code:

```
         if (
                  prevColorMode?.mode === "gradient" ||
                  prevColorMode?.mode === "rainbow" ||
                  prevColorMode?.mode === "turbo"
                ) {
                  const { colorField, minValue, maxValue } = prevColorMode;
                  return id === "rainbow"
                    ? { mode: "rainbow", colorField, minValue, maxValue }
                    : id === "turbo"
                    ? { mode: "turbo", colorField, minValue, maxValue }
                    : {
                        mode: "gradient",
                        colorField,
                        minValue,
                        maxValue,
                        minColor: DEFAULT_MIN_COLOR,
                        maxColor: DEFAULT_MAX_COLOR,
                      };
                }
                return prevColorMode;
```

Which refuses to set the new mode if it doesn't have a previous mode, which it doesn't in the case of empty settings.

Yeah `onSettingsChange` can also take a plain value instead of a function. I can switch to that if we want to refactor this logic a bit more.

foxymiles:
@jtbandes I tidied up the color change logic here a bit. I left it taking a function just to keep most of the existing logic intact since that was how `onColorChange` was already being used in all but one case.

packages/studio-base/src/panels/ThreeDimensionalViz/TopicSettingsEditor/PointCloudSettingsEditor.tsx
```diff
@@ -156,6 +152,10 @@ export default function PointCloudSettingsEditor(
       ? { mode: "turbo", colorField: defaultColorField }
       : { mode: "flat", flatColor: DEFAULT_FLAT_COLOR });
 
+  function onColorModeChange(newValueFn: (prevColorMode: ColorMode) => ColorMode | undefined) {
```
defunctzombie, line 155: Why is newValueFn a function rather than the new color mode value?

foxymiles: I left it this way to avoid more extensive reworking of the somewhat involved existing color mode change logic. If we want to do a bigger overhaul of that then this can just take a value instead and I can update the logic in the event handlers.

defunctzombie: Sounds good. :shipit:

packages/studio-base/src/panels/ThreeDimensionalViz/TopicSettingsEditor/PointCloudSettingsEditor.tsx
```diff
@@ -102,6 +106,10 @@ const SegmentedControlWrapper = styled.div`
   align-items: center;
 `;
 
+function isMappedColorMode(mode: ColorMode): mode is MappedColorMode {
+  return mode.mode === "turbo" || mode.mode === "rainbow" || mode.mode === "gradient";
```
jtbandes, line 110: using `switch` might allow the TS compiler to warn us when we inevitably forget to add cases here.

jtbandes (commented):
👍

Issue #2270, opened 2021-12-03
Stepping through a bag results in a different end state than playing a bag


jhurliman:
[markertest009.bag.zip](https://github.com/foxglove/studio/files/7652913/markertest009.bag.zip)

End state after playing back the bag to the end:
<img width="930" alt="Screen Shot 2021-12-03 at 3 31 03 PM" src="https://user-images.githubusercontent.com/195374/144685851-ea25dc96-b9d5-4096-ab86-7e68e32bef68.png">

End state after pressing the right keyboard button until the end of the bag is reached:
<img width="930" alt="Screen Shot 2021-12-03 at 3 31 13 PM" src="https://user-images.githubusercontent.com/195374/144685861-e7004039-eafd-41b2-887d-1389e6293e14.png">

jtbandes:
This is kind of by design. Stepping is really a seek, and seeking to the end can result in a different end state as well.

I think the ticket should specifically call out TFs or markers, because those are some of the only things the app displays in a truly stateful manner (where to display them correctly you technically need to process every preceding message in the entire bag)

amacneil:
From live discussion: Let's scope this to supporting "step forward in time" only (e.g. for seeks <100ms).

Seeking forward large distances, or stepping backwards can continue to treat this as a seek and throw away existing visualization state.

PR #2418, opened 2021-12-16
Forward seek plays all messages until the seek time is reached

defunctzombie:
**User-Facing Changes**
Seeking (via the keyboard shortcut or the playback control button) triggers playback to start and automatically stop when the datasource time passes the desired seek location. This allows a user to repeatedly forward seek and achieve the same result with visualization as playing the datasource. This functionality impacts panels which rely on message state (i.e. coordinate frames)  for proper visualization.

**Description**
Forward seek plays forward rather than jump seek
    
Rather than jumping forward in the datasource a fixed amount, seeking forward plays the data until at least the additional amount of time has elapsed in the datasource.
    
This allows repeated forward seeking to achieve the same playback result as playing through the datasource start to finish.
    
Fixes: #2270

### DIFF
--- a/packages/studio-base/src/components/PlaybackControls/index.tsx
+++ b/packages/studio-base/src/components/PlaybackControls/index.tsx
@@ -56,7 +56,7 @@ export default function PlaybackControls({
   const [repeat, setRepeat] = useState(false);
   const stopAtTime = useRef<Time | undefined>(undefined);
 
-  // See comments below in ArrowRight handler for how seeking is handled
+  // See comments below in seekForwardAction for how seeking is handled
   useMessagePipeline(
     useCallback(
       (ctx) => {
--- a/packages/studio-base/src/components/PlaybackControls/index.tsx
+++ b/packages/studio-base/src/components/PlaybackControls/index.tsx
@@ -13,12 +13,13 @@
 
 import { Stack, IButtonStyles, useTheme, StackItem, makeStyles } from "@fluentui/react";
 import { merge } from "lodash";
-import { useCallback, useMemo, useState } from "react";
+import { useCallback, useMemo, useRef, useState } from "react";
 
 import { compare, Time } from "@foxglove/rostime";
 import HoverableIconButton from "@foxglove/studio-base/components/HoverableIconButton";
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/components/PlaybackControls/index.tsx
```diff
@@ -53,6 +54,25 @@ export default function PlaybackControls({
...
+      if (stopNextFrame.current) {
+        stopNextFrame.current = false;
+        setDoPause((old) => old + 1);
+      }
+    }, []),
+  );
+
+  useLayoutEffect(() => {
+    void doPause;
+    requestAnimationFrame(() => {
```
defunctzombie, line 72: queueMicrotask?

packages/studio-base/src/components/PlaybackControls/index.tsx
```diff
@@ -53,6 +54,25 @@ export default function PlaybackControls({
   const theme = useTheme();
   const styles = useStyles();
   const [repeat, setRepeat] = useState(false);
+  const stopNextFrame = useRef(false);
+
+  const [doPause, setDoPause] = useState(0);
+
+  useMessagePipeline(
+    useCallback(() => {
+      if (stopNextFrame.current) {
```
defunctzombie, line 63: change behavior to play at least some N time until pausing

defunctzombie:
@jtbandes @jhurliman I've implemented the behavior discussed: stop playback when the datasource current time has at least reached the seek to time.

jhurliman (approved):

packages/studio-base/src/components/PlaybackControls/index.tsx
```diff
@@ -53,11 +54,31 @@ export default function PlaybackControls({
   const theme = useTheme();
   const styles = useStyles();
   const [repeat, setRepeat] = useState(false);
+  const stopAtTime = useRef<Time | undefined>(undefined);
+
+  // See comments below in ArrowRight handler for how seeking is handled
```
defunctzombie, line 59:
```suggestion
  // See comments below in seekForwardAction for how seeking is handled
```

PR #2422, opened 2021-12-17
Performance improvements

jhurliman:
**User-facing changes**

Better performance when rendering complex scenes, especially in the 3D panel

**Description**

- Bump @foxglove/rosmsg-serialization to get faster lazy messages
- Remove unused hook, don't copy marker points
- Convert all incoming messages in SceneBuilder from lazy to fully parsed
- Speed up transform math by only working with unscaled matrices
- Add `points` to BaseMarker definition
- Cache and recycle the data structure in getMarkers instead of allocating new arrays each frame

### DIFF
--- a/packages/studio-base/src/panels/ThreeDimensionalViz/SceneBuilder/index.ts
+++ b/packages/studio-base/src/panels/ThreeDimensionalViz/SceneBuilder/index.ts
@@ -33,7 +33,6 @@ import {
   MutablePose,
   Pose,
   StampedMessage,
-  MutablePoint,
   BaseMarker,
   PoseStamped,
   VelodyneScan,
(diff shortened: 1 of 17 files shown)


packages/studio-base/src/panels/ThreeDimensionalViz/SceneBuilder/index.ts
```diff
@@ -810,6 +803,16 @@ export default class SceneBuilder implements MarkerProvider {
       return;
     }
 
+    // SceneBuilder fully parses all of the messages it consumes, and sometimes
```
defunctzombie, line 806: Its probably worth doing this toJSON in `useFrame` https://github.com/foxglove/studio/blob/main/packages/studio-base/src/panels/ThreeDimensionalViz/useFrame.ts#L44-L58 - so when no messages are arriving but the user is panning and scene builder is rendering you don't have to attempt toJSON() again.

packages/studio-base/src/panels/ThreeDimensionalViz/World.tsx
```diff
@@ -149,7 +128,35 @@ function World(
   }: Props,
   ref: typeof Worldview,
 ) {
-  const markersByType = getMarkers(markerProviders, currentTime);
+  const markersRef = useRef<InteractiveMarkersByType | undefined>(undefined);
```
defunctzombie, line 131: This could use a comment around why it exists and some usage notes.

PR #2413, opened 2021-12-16
Use correct casing for 'WebSocket URL' in connection dialog

esthersweon:
**User-Facing Changes**
- Use correct casing for 'WebSocket URL' in connection dialog
- Update Foxglove WebSocket default port to 8765

Fixes: #2420

PR #2426, opened 2021-12-17
Fix default value for FoxgloveWebsocketDataSourceFactory

defunctzombie:
**User-Facing Changes**
None unless Open Dialog is enabled

When connecting to Foxglove Websocket, the default value suggests port 8765.

**Description**
The default value for Websocket url should use port 8765 to match our example servers.

Fixes: #2420


defunctzombie:
Replaced by https://github.com/foxglove/studio/pull/2413#pullrequestreview-834643009

Issue #2417, opened 2021-12-16
Can't drag bag file on top of the Open dialog


jtbandes:
When the Open dialog is visible, if I try to drag a bag file into the app, the browser just downloads the file. Instead the app should show the drop target overlay, and load the file as a data source, the same way it does when I drag & drop if the open dialog is not up.

<img width="974" alt="image" src="https://user-images.githubusercontent.com/14237/146460017-eb4fdcb5-6348-4811-beac-bd540a3fdfcf.png">

PR #2429, opened 2021-12-17
Fix drag&drop when Open Dialog is open

defunctzombie:
**User-Facing Changes**
None unless Open Dialog feature flag is on.

Users can drag&drop files over the open dialog to load in studio.

**Description**
Enable event bubbling in open dialog to support drag&drop while the dialog is open.

Fixes: #2417

### DIFF
--- a/packages/studio-base/src/components/OpenDialog/OpenDialog.tsx
+++ b/packages/studio-base/src/components/OpenDialog/OpenDialog.tsx
@@ -114,6 +114,9 @@ export default function OpenDialog(props: OpenDialogProps): JSX.Element {
       maxWidth={800}
       minWidth={800}
       modalProps={{
+        layerProps: {
+          eventBubblingEnabled: true,
+        },
         styles: {
           main: {
             minHeight: 520,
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/components/OpenDialog/OpenDialog.tsx
```diff
@@ -114,6 +114,9 @@ export default function OpenDialog(props: OpenDialogProps): JSX.Element {
       maxWidth={800}
       minWidth={800}
       modalProps={{
+        layerProps: {
+          eventBubblingEnabled: true,
```
jtbandes, line 118: might want a comment on why this is useful – wonder if this is a problem with other modals too? Maybe not as big of a deal since most modals don't pop up immediately when you load the app?

defunctzombie: https://github.com/foxglove/studio/commit/b9a07286c113d077e5f1b8927fa53c8d26182502

PR #2430, opened 2021-12-17
URDF fixes

jhurliman:
**User-facing changes**

- More .urdf/.xacro files are loaded correctly
- More COLLADA .dae files are loaded correctly

**Description**

- Bump xacro-parser to pull in "=" vs ":=" fix
- Store a copy of the pose for URDF components, not a reference
- Handle leading/trailing whitespace in URDF vector attributes
- Show marker info when clicking on URDF markers
- Only render URDF markers that have a valid transform
- Fix URDF primitives not showing up, accidental Vector3 mutation by reference
- Catch GLTFExporter errors
- Fix collada loading for materials with normal/emissive/specular/light maps
- Fix THREE.js to ROS coordinate frame conversion
- Make Marker.points optional

Screen Shot 2021-12-16 at 8 47 14 PM
Screen Shot 2021-12-16 at 7 20 33 PM

### DIFF
--- a/packages/den/urdf/parser.ts
+++ b/packages/den/urdf/parser.ts
@@ -352,7 +352,7 @@ function parsePose(xml: Element): Pose {
 }
 
 function parseVec3Attribute(xml: Element, attribName: string): Vector3 | undefined {
-  const parts = xml.getAttribute(attribName)?.split(" ");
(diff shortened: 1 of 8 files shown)


jhurliman:
Ideally, we can pull in https://github.com/gkjohnson/xacro-parser/pull/76 as well and fix duckiebot.

PR #2432, opened 2021-12-17
Enable Open Dialog feature by default

defunctzombie:
**User-Facing Changes**

The open dialog is a new way to select data sources. Rather than a single list of different sources and types of sources (urls, files, native, etc), the open dialog presents the user with a simplified set of data source types. Different data sources have different modes of interaction for opening.

The dialog also introduces the concept of "sample data". These are pre-defined datasets and layouts for new users to learn aspects of the app while browsing sample datasets.

**Description**
This change also introduces a way to provide default values for app settings.

Fixes: #2431

### DIFF
--- a/desktop/renderer/Root.tsx
+++ b/desktop/renderer/Root.tsx
@@ -84,7 +84,7 @@ export default function Root(): ReactElement {
 
   const providers = [
     /* eslint-disable react/jsx-key */
-    <NativeStorageAppConfigurationProvider />,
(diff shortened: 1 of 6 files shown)


defunctzombie:
I have left the setting in experimental features so we can make a release where users _could_ disable it if they encounter a blocking issue to using the app with the open dialog. Once we make the release - I will remove the feature from the settings UI and it will remain as an override in the code only.

jtbandes:
Rather than the whole `defaults` concept can't you just use `?? true` in the place where you read the app setting?

Issue #2439, opened 2021-12-17
A layout renamed in another device does not update the UI


defunctzombie:
**Description**
When signed in to web desktop apps - if I rename a layout in one app, I do not see the updated name reflected in the other (even after several minutes).

**Steps To Reproduce**

- Sign into the same account in web and desktop
- Make a new layout with any panel
- Make sure the layout appears in both apps
- Rename the layout in desktop
- Observe the name does not update in web (even after several minutes).

Refreshing either app does not fix the issue. The name is never updated.

**Expected Behavior**
Updated name when renaming on either device.

**Actual Behavior**
No updated name.

PR #2442, opened 2021-12-18
Update savedAt when renaming layouts

jtbandes:
**User-Facing Changes**
Fixed a bug where renaming a layout would not take effect when signed in across multiple devices.

**Description**
Fixes https://github.com/foxglove/studio/issues/2439 (along with https://github.com/foxglove/console/pull/454).

Renames are different from other layout edits: they don't require a "save" step in order to be saved. Instead we directly update the name in the local cache, and mark the layout as `"updated"` so it will be uploaded with the next sync — during the sync, the `baseline` (local committed changes) is what we upload, so `baseline.savedAt` is used as the new savedAt time. Previously the `baseline.savedAt` was only being updated in the "overwrite" (when the user explicitly saves). This PR updates changes the `baseline.savedAt` when renaming a layout too.

### DIFF
--- a/packages/studio-base/src/services/LayoutManager/index.ts
+++ b/packages/studio-base/src/services/LayoutManager/index.ts
@@ -307,6 +307,11 @@ export default class LayoutManager implements ILayoutManager {
       this.notifyChangeListeners({ type: "change", updatedLayout: result });
       return result;
     } else {
+      const isRename =
+        this.remote != undefined &&
+        name != undefined &&
+        localLayout.syncInfo != undefined &&
+        localLayout.syncInfo?.status !== "new";
       const result = await this.local.runExclusive(
         async (local) =>
           await local.put({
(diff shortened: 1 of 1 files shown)


defunctzombie:
👍 

And a side-effect is that we also _save_ your changes to the layout when you rename?

jtbandes:
No, your `working` changes are not saved and stay as working changes. I’m actually not sure why the existing code updates `working.savedAt` when you make edits … I don’t know that `working.savedAt` is used for anything at all and could probably be removed.

packages/studio-base/src/services/LayoutManager/index.ts
```diff
@@ -307,21 +307,23 @@ export default class LayoutManager implements ILayoutManager {
...
-            // If the name is being changed, we will need to upload to the server
-            syncInfo:
-              this.remote &&
-              name != undefined &&
-              localLayout.syncInfo &&
-              localLayout.syncInfo?.status !== "new"
-                ? { status: "updated", lastRemoteSavedAt: localLayout.syncInfo?.lastRemoteSavedAt }
-                : localLayout.syncInfo,
+            // If the name is being changed, we will need to upload to the server with a new savedAt
+            baseline: isRename ? { ...localLayout.baseline, savedAt: now } : localLayout.baseline,
```
jtbandes, line 323: I feel like there might be some other strange side effects of updating baseline.savedAt when you rename a layout but I am not sure. In reality it probably only matters if you are editing a shared layout at the same time as someone else...

Issue #2371, opened 2021-12-14
Ignore Cmd modifier when toggling between 2d/3d in 3d panel


amacneil:
**Description**
Hitting `3` in the 3d panel apparently toggles between 2d and 3d mode.

But if I have Cmd held down, I would expect Cmd+3 to switch to the 3rd tab in Chrome. Instead, the 3d panel hijacks it.

We should ignore this keyboard shortcut if a modifier key is held down.

PR #2449, opened 2021-12-20
Ignore CMD modifier for 3d panel `3` hotkey

defunctzombie:
**User-Facing Changes**
Pressing `cmd+3` in the web app - properly changes you to your third tab (in chrome) works.

**Description**
The 3d panel sets up a keyboard shortcut on the `3` key. By default the keylister captures events with modifier keys and also prevents default. This makes this keyboard shortcut capture events for `cmd+3` as well as `3`. In chrome, cmd+3 is a shortcut for switching tabs. Since the 3d panel only needs the `3` keylistener, we avoid preventing default when the meta or ctrl key is also pressed.
    
Fixes: #2371

### DIFF
--- a/packages/studio-base/src/components/KeyListener.tsx
+++ b/packages/studio-base/src/components/KeyListener.tsx
@@ -12,7 +12,9 @@
 //   You may not use this file except in compliance with the License.
 
 type KeyHandlers = {
-  [key: string]: (event: KeyboardEvent) => void;
+  // Invoke handler for matcking key.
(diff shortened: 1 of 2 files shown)


defunctzombie:
A _better_ fix is probably to use https://www.npmjs.com/package/react-hotkeys or a hooks like version. This is a quick fix to address the immediate issue without re-working all uses of the key listener component.

foxymiles (approved):
Makes sense as an interim fix. 👍

Issue #2251, opened 2021-12-02
The `globalvariables` is not provided to node player until it's changed


bunnyc1986:
**Description**
The globalVars is empty in node's publisher call when the layout is loaded. The global variables are only set to the parameter when some change is made to global variables.

- Version: latest master
- OS: Chrome 96.0.4664.55

**Steps To Reproduce**

1. Create a node to log the globalVars and save it to a layout.
2. Create a global variables and see it's logged in node playground console.
3. Then refresh the page. Play the bag. The empty globalVars is logged.

**Expected Behavior**
The current global variables that are saved to the layout should be logged in the node playground console.

**Actual Behavior**
The empty global variables are logged.

I actually found the bug in the code in `PlayManager`: https://github.com/foxglove/studio/blob/bb806eaad6f41892a439dee2b395066092858787/packages/studio-base/src/components/PlayerManager.tsx#L95-L99
AND:
https://github.com/foxglove/studio/blob/bb806eaad6f41892a439dee2b395066092858787/packages/studio-base/src/components/PlayerManager.tsx#L117

`globalVariablesRef.current` actually doesn't change when `globalVariables` is changed. My fix is to remove L95 and change L99 to:
```
  const globalVariablesRef = useRef<GlobalVariables>(EMPTY_GLOBAL_VARIABLES);
  globalVariablesRef.current = useCurrentLayoutSelector(
    (state) => state.selectedLayout?.data?.globalVariables ?? EMPTY_GLOBAL_VARIABLES,
  );
```

PR #2453, opened 2021-12-21
Update player with global variables on initialization

defunctzombie:
**User-Facing Changes**
User nodes are initialized with the latest globalVariable values. A user does not need to manually change global variable values to get the correct varions in their user node code.

**Description**
During player initialization the global variables would be cached to the current value from the layout. If a layout was still loading these variables would be empty. When creating a player, the constructor would receive these empty cached values rather than the latest values.

This change updated the PlayerManager logic to use the latest global variable value when constructing a player. Care is taken to avoid constructing players when the variables change.

Fixes: #2251


jtbandes:
👍

PR #2456, opened 2021-12-21
helper hook that throws when an unstable function is provided

defunctzombie:
**User-Facing Changes**
None

**Description**
React hook dependencies use reference equality to determine if any dependencies are changed. For selector hooks we often want to ensure the caller is using a stable function identity so we are not running effects more than we need.

This change adds a new useThrowUnstableFunction hook which throws when it encounters an unstable function identity across multiple invocations.


defunctzombie:
Is this a good idea?

packages/studio-base/src/context/CurrentLayoutContext/index.ts
```diff
@@ -102,6 +102,19 @@ export type SelectedPanelActions = {
 
 const CurrentLayoutContext = createContext<ICurrentLayout | undefined>(undefined);
 
+const useThrowUnstableFunction =
```
defunctzombie, line 105: fixme - if we like this I'll move to separate file

foxymiles (approved):
This makes sense. It will be handy to have a tool to catch accidental use of unstable selector functions like this.

packages/studio-base/src/context/CurrentLayoutContext/index.ts
```diff
@@ -112,6 +112,9 @@ export function useCurrentLayoutSelector<T>(selector: (layoutState: LayoutState)
   const currentLayout = useGuaranteedContext(CurrentLayoutContext);
   const [_, forceUpdate] = useReducer((x: number) => x + 1, 0);
 
+  // Catch locations using unstable function selectors
+  useMustNotChange(selector);
```
jtbandes, line 116: It seems like this would make it impossible to write certain features/UIs, such as a dropdown list of panels in the current layout next to a code block showing the config of the panel you selected in the dropdown.

packages/hooks/src/useMustNotChange.ts
```diff
@@ -0,0 +1,28 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { useRef } from "react";
+
+const useMustNotChangeImpl = (value: unknown): void => {
+  const valueRef = useRef<unknown | undefined>(value);
+  if (valueRef.current !== value) {
+    throw new Error("Value must not change");
```
jtbandes, line 10: Maybe just copy this version which has a nicer message: https://github.com/cruise-automation/webviz/blob/633920b68bdcd951b79a0bc44d0a5cf2b35814d6/packages/webviz-core/src/util/hooks.js#L113-L120

packages/studio-base/src/context/CurrentLayoutContext/index.ts
defunctzombie: Should I relax this to trigger if the value changes before some number of renders? Or go with the current setup since we do not have any code that is using a changing selector?

packages/hooks/src/useMustNotChange.ts
defunctzombie: We could - when I was using this hook I found the stacktrace more informative than the value. Maybe cause that was specific to weeding out functions rather than changing values. I'll include the values in the message since it is strictly better to do so.

PR #2457, opened 2021-12-21
Introduce a fixed frame for 3D panel rendering

jhurliman:
**User-Facing Changes**

Non frame-locked renderables such as point clouds, occupancy grids, and markers with `frame_locked=false` will now correctly remain at their original rendered position in the root frame

**Description**

- Introduce a fixed frame for 3D panel rendering. The fixed frame is currently automatically set as the root frame of the current rendering frame. The rendering frame is the coordinate frame all marker poses are translated into before being sent to the GPU. Non frame-locked renderables such as point clouds, occupancy grids, and markers with `frame_locked=false` will now correctly remain at their original rendered position in the root frame. This works regardless of which frame is currently the render frame or which frame the renderables are in, as long as they share a common ancestor. These changes do not decouple the concepts of render frame and follow frame, but that could be something for a future PR (possibly using message path syntax to follow any pose in the scene?)
- Rename CoordinateFrame#apply() to CoordinateFrame#applyLocal(), and add an apply() function that takes an additional `rootFrame` parameter
- Add additional state to `renderMarkers()` arguments: `renderFrame`, `fixedFrame`, and `transforms`. This ensures marker providers always render with up to date coordinate frames and transforms, and reduces the amount of state each marker provider must hold

https://user-images.githubusercontent.com/195374/146882314-9e171887-f7ff-46f4-b299-1f42bd7962c4.mov


packages/studio-base/src/panels/ThreeDimensionalViz/SceneBuilder/index.ts
```diff
@@ -850,9 +853,9 @@ export default class SceneBuilder implements MarkerProvider {
           }
         }
 
-        const pose = computeMarkerPose(marker, this.transforms, this.renderFrameId, time);
+        const pose = computeMarkerPose(marker, transforms, renderFrame, fixedFrame, time);
         if (!pose) {
-          missingTfFrameIds.add(marker.header.frame_id);
+          this._missingTfFrameIds.add(marker.header.frame_id);
```
defunctzombie, line 858: We should also add to topic errors for this topic to indicate the specific topic is missing a frame_id.

How is the missingTfFrameIds exposed to the user?

packages/studio-base/src/panels/ThreeDimensionalViz/commands/PointClouds/memoization.ts
```diff
@@ -40,6 +42,11 @@ export function updateMarkerCache(
         hitmapColors: marker.hitmapColors,
       };
     }
+
+    // Update the marker pose unconditionally since it can change per-frame even
+    // when we don't need to do a full point cloud decode
+    (decoded.marker as { pose?: Pose }).pose = marker.pose;
```
defunctzombie, line 48: Does our marker type not include a pose field? Wondering why the cast is necessary here and if we can tighten the type information elsewhere?

Separately - should we be storing the rendering pose as a separate field on `decoded` so we don't mutate the marker? Seems like that might make it clearer reading the code that there is a marker that is immutable and separately we have a rendering pose which is updated.

packages/studio-base/src/panels/ThreeDimensionalViz/index.tsx
```diff
@@ -103,6 +107,31 @@ function BaseRenderer(props: Props): JSX.Element {
...
+  const orphanedFrame = useMemo(() => new CoordinateFrame("(empty)", undefined), []);
+
+  const renderFrame = useMemo<CoordinateFrame>(() => {
+    // If the user specified a followTf, do not fall back to any other (valid) frame
+    if (typeof followTf === "string") {
+      return transforms.frame(followTf) ?? orphanedFrame;
+    }
+
+    // Try the conventional list of transform ids
+    for (const frameId of DEFAULT_FRAME_IDS) {
```
defunctzombie, line 119: What would it look like to auto-select the follow frame based on the first marker selected (if no follow frame is selected)? Is that viable with these code paths? Don't need to do in this change (I think we have a separate issue for it) but I'm wondering if this makes it easier or harder to achieve that.

packages/studio-base/src/panels/ThreeDimensionalViz/transforms/CoordinateFrame.ts
```diff
@@ -261,6 +260,44 @@ export class CoordinateFrame {
...
+    rootFrame: CoordinateFrame,
+    srcFrame: CoordinateFrame,
+    dstTime: Time,
+    srcTime: Time,
+    maxDelta: Duration = INFINITE_DURATION,
+  ): MutablePose | undefined {
+    // perf-sensitive: function params instead of options object to avoid allocations
+
+    // Transform from the source frame to the root frame
+    if (rootFrame.applyLocal(out, input, srcFrame, srcTime, maxDelta) == undefined) {
```
defunctzombie, line 294: Is it feasible to compute the transformation from the source frame -> root frame -> this frame and then applyLocal once? Writing that out I guess thats no different than doing the two applyLocals since they have to compute those transformations. I had in mind that maybe the transformation is something that could be applied to more than one coordinate frame but that might not be true given all the paramers?

packages/studio-base/src/panels/ThreeDimensionalViz/SceneBuilder/index.ts
jhurliman:
```
if (this._missingTfFrameIds.size > 0) {
  const error = this._addError(this.errors.topicsMissingTransforms, topic.name);
  for (const frameId of this._missingTfFrameIds) {
    error.frameIds.add(frameId);
  }
}
```

packages/studio-base/src/panels/ThreeDimensionalViz/transforms/TransformTree.test.ts
```diff
@@ -126,20 +136,28 @@ describe("TransformTree", () => {
...
 
     // Too far in the past
-    expect(tree.apply(output, emptyPose(), "world", "base_link", TIME_ZERO)).toBeUndefined();
+    expect(
+      tree.apply(output, emptyPose(), "world", "world", "base_link", TIME_ZERO, TIME_ZERO),
+    ).toBeUndefined();
 
     // This works
-    expect(tree.apply(output, emptyPose(), "world", "base_link", T2)).toBeDefined();
+    expect(tree.apply(output, emptyPose(), "world", "world", "base_link", T2, T2)).toBeDefined();
```
jtbandes, line 160: Add some tests where srcTime and dstTime are different?

packages/studio-base/src/panels/ThreeDimensionalViz/transforms/CoordinateFrame.ts
```diff
@@ -261,6 +260,44 @@ export class CoordinateFrame {
...
+   * @param srcTime Time to transform from srcFrame into rootFrame
+   * @param maxDelta The time parameter can exceed the bounds of the transform
+   *   history by up to this amount and still clamp to the oldest or newest
+   *   transform
+   * @returns A reference to `out` on success, otherwise undefined
+   */
+  apply(
+    out: MutablePose,
+    input: Pose,
+    rootFrame: CoordinateFrame,
```
jtbandes, line 285: why name the param rootFrame instead of fixedFrame?

jhurliman: I want to package this as a separate library at some point, so I left the term "fixed frame" out of this part of the code to avoid introducing more terminology

jtbandes: I guess I was just a little confused by the rootFrame name because this frame isn't necessarily the root, right?

jhurliman: That's right. Maybe fixed frame is a better name, or another term?

PR #2463, opened 2021-12-21
Fix handling of launch preference change during web session.

### LINKED ISSUE #2383: Changing link preference to "Ask each time" immediately goes back to launch choice screen
1. Visit https://studio.foxglove.dev/?ds=foxglove-websocket&ds.url=ws://localhost:8765/
2. Select "remember my preference" and click "Web"
3. Open Preferences sidebar and select "Open links in: Ask each time"

**Expected:** no change for the current session; the **next** time I open a link, I will be asked whether to open it in web or desktop.

**Actual:** the app immediately reverts back to the screen where I originally clicked Web for the current session.

https://user-images.githubusercontent.com/14237/146097368-08d013b5-49e2-42d3-b00c-f12ef9c0d898.mov

### LINKED ISSUE #2511: Set session preference to web whenever we don't show launch page
**Description**
Bug:
- open incognito window, visit studio.foxglove.dev
- create a new empty layout
- hit refresh, now you see the launch (web vs desktop) modal

Solution:
Any time you open studio and _don't_ see the web vs desktop dialog, we should opt you in to web for that session. E.g. if I hit studio.foxglove.dev homepage my session should stay as web.

foxymiles:
**User-Facing Changes**
This fixes handling of launch preference changes in an active session.

**Description**
The solution here is to set a  temporary, `sessionStorage` preference for web if the user changes the launch preference to "ask me" during an active web session so that the existing session isn't disrupted.

Fixes #2383 
Fixes #2511

### DIFF
--- a/packages/studio-base/src/App.tsx
+++ b/packages/studio-base/src/App.tsx
@@ -2,7 +2,7 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
-import { Suspense, useState } from "react";
+import { Suspense, useEffect, useState } from "react";
 import { DndProvider } from "react-dnd";
 import { HTML5Backend } from "react-dnd-html5-backend";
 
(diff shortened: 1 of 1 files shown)


jtbandes (commented):
This doesn't really make sense to me, why would we want to set a value in sessionStorage but never read it? Does reading from localStorage somehow fall back to reading from sessionStorage?

foxymiles:
> This doesn't really make sense to me, why would we want to set a value in sessionStorage but never read it? Does reading from localStorage somehow fall back to reading from sessionStorage?

The value in session storage gets checked in `App.tsx`. We don't need to check the current value because our goal here is to set a temporary, session storage scoped value to keep the user from getting kicked out of their current session. Looking at this again though I realize we need to do the same thing when the user selects "open in desktop" in an active web session here so I've updated that logic.

jtbandes:
I'm not super familiar with how session storage works – if you click another studio link while your current page is still open, does the new page also use the same session? In that case, I think this implementation might not be quite right, because you might select "ask each time" and then click a link, and it would still use the newly set session preference of "web".

foxymiles:
> I'm not super familiar with how session storage works – if you click another studio link while your current page is still open, does the new page also use the same session? In that case, I think this implementation might not be quite right, because you might select "ask each time" and then click a link, and it would still use the newly set session preference of "web".

The difficulty is we also need to do the right thing on a refresh so we need to persist that value somehow. I think having that particular tab operate as a web session is a reasonable compromise but I'm open to better suggestions.

jtbandes (approved):

PR #2480, opened 2021-12-22
Update esbuild to 0.14.7

jhurliman:
**User-Facing Changes**

None

**Description**

Updates our JS and CSS minifier to the latest version

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
   },
   "resolutions": {
     "@types/node": "15.3.0",
-    "esbuild-loader/esbuild": "0.12.18",
+    "esbuild-loader/esbuild": "0.14.7",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -53,7 +53,7 @@
   },
   "resolutions": {
     "@types/node": "15.3.0",
-    "esbuild-loader/esbuild": "0.12.18",
+    "esbuild-loader/esbuild": "0.14.7",
```
jtbandes, line 56: are we able to remove this from "resolutions" now?

jhurliman: In one more release I imagine. It currently ships 0.14.6, and 0.14.7 includes a performance improvement for cross-module enum minification that was showing a 15% speedup in some apps.

PR #2487, opened 2021-12-22
Addressed copy feedback on panel tooltips in the panel list

esthersweon:
**User-Facing Changes**
- Addressed copy feedback on panel tooltips in the panel list


packages/studio-base/src/panels/index.ts
```diff
@@ -87,14 +87,14 @@ const builtin: PanelInfo[] = [
...
-    description: "Publish data over a life connection.",
+    description: "Publish messages to the data source.",
     help: PublishHelp,
     module: async () => await import("./Publish"),
   },
   {
     title: "Raw Messages",
     type: "RawMessages",
-    description: "Display topic messages as JSON.",
+    description: "Display topic messages as collapsible trees.",
```
defunctzombie, line 97: This sounds too robotic. The actual benefit to the user is that this panel allows them to inspect their message content.

packages/studio-base/src/panels/index.ts
```diff
@@ -38,21 +38,21 @@ const builtin: PanelInfo[] = [
...
-    description: "Display data for a given diagnostics node.",
+    description: "Display data for a given ROS diagnostics node.",
     help: DiagnosticStatusPanelHelp,
     module: async () => await import("./diagnostics/DiagnosticStatusPanel"),
   },
   {
     title: `Diagnostics – Summary`,
     type: "DiagnosticSummary",
-    description: "Display the status of diagnostics nodes.",
+    description: "Display the status of ROS diagnostics nodes.",
```
defunctzombie, line 48: We could be more explicit in these two panels and indicate they display specifically the DiagnosticStatus and DiagnosticArray (or whatever it is actually called).

http://docs.ros.org/en/melodic/api/diagnostic_msgs/html/msg/DiagnosticStatus.html

packages/studio-base/src/panels/index.ts
esthersweon: `Inspect the messages of incoming topics."?

packages/studio-base/src/panels/index.ts
```diff
@@ -38,21 +38,21 @@ const builtin: PanelInfo[] = [
   {
     title: `Diagnostics – Detail`,
     type: "DiagnosticStatusPanel",
-    description: "Display data for a given diagnostics node.",
+    description: "Display data for a given ROS diagnostics node.",
```
jtbandes, line 41: There is no such thing as a "diagnostics node"; any node can publish to the /diagnostics topic.
```suggestion
    description: "Display ROS diagnostics data for a given node.",
```

packages/studio-base/src/panels/index.ts
esthersweon: I personally want to leave those details to the actual docs (on website, Help sidebar).

These tooltips should be 5 - 10-word summaries of each panel, and just a gut-check for users when they are considering whether they should add a panel for a given data source.

packages/studio-base/src/panels/index.ts
```diff
@@ -38,21 +38,21 @@ const builtin: PanelInfo[] = [
...
-    description: "Display the status of diagnostics nodes.",
+    description: "Display the status of ROS diagnostics nodes.",
     help: DiagnosticSummaryHelp,
     module: async () => await import("./diagnostics/DiagnosticSummary"),
   },
   {
     title: "Image",
     type: "ImageViewPanel",
-    description: "Display camera feed images.",
+    description: "Display image topic messages.",
```
jtbandes, line 55: the words "topic" and "messages" are not particularly useful or unique to this panel

```suggestion
    description: "Display compressed or uncompressed images.",
```

packages/studio-base/src/panels/index.ts
defunctzombie: Maybe: `Inspect messages` - the spiritual tool in ROS that influenced this panel is `rostopic echo` if that helps think of a description.

defunctzombie: Not sure how much we need to repeat the word topics in these descriptions. Maybe `Inspect messages on a topic`. Incoming is implied since our app is all about viewing data. Topic might be nice to include but is somewhat also implied.

packages/studio-base/src/panels/index.ts
defunctzombie: Then maybe something like `Display ROS DiagnosticStatus messages`. I know our app displays lots of ROS specific messages - but many of those concepts are pretty generic ideas (images, point clouds, gps). Diagnostics is more likely to be framework specific.

packages/studio-base/src/panels/index.ts
esthersweon: Changed to "Inspect messages on a topic."

packages/studio-base/src/panels/index.ts
```diff
@@ -136,14 +136,14 @@ const builtin: PanelInfo[] = [
...
-    description: "View time and topic information for the data source.",
+    description: "View data source details like topics, datatypes, and duration.",
     help: SourceInfoHelp,
     module: async () => await import("./SourceInfo"),
   },
   {
     title: "Variable Slider",
     type: "GlobalVariableSliderPanel",
-    description: "Quickly update numerical variable values for a layout.",
+    description: "Update numerical variable values for a layout.",
```
defunctzombie, line 146:
```suggestion
    description: "Update global variables in the layout.",
```

packages/studio-base/src/panels/index.ts
```diff
@@ -38,21 +38,21 @@ const builtin: PanelInfo[] = [
...
-    description: "Display data for a given diagnostics node.",
+    description: "Display ROS diagnostics data for a given node.",
     help: DiagnosticStatusPanelHelp,
     module: async () => await import("./diagnostics/DiagnosticStatusPanel"),
   },
   {
     title: `Diagnostics – Summary`,
     type: "DiagnosticSummary",
-    description: "Display the status of diagnostics nodes.",
+    description: "Display the status of nodes outputting ROS diagnostics data.",
```
jtbandes, line 48:
```suggestion
    description: "Display a summary of ROS diagnostics data from all nodes.",
```

packages/studio-base/src/panels/index.ts
esthersweon: So something like this? 
```
  {
    title: `Diagnostics – Detail`,
    description: "Display ROS DiagnosticArray messages for a given node.",
   ...
  },
  {
    title: `Diagnostics – Summary`,
    description: "Display all ROS DiagnosticArray messages.",
   ...
  },
```

packages/studio-base/src/panels/index.ts
```diff
@@ -87,14 +87,14 @@ const builtin: PanelInfo[] = [
   {
     title: "Publish",
     type: "Publish",
-    description: "Publish data over a life connection.",
+    description: "Publish messages to the data source.",
```
jtbandes, line 90:
```suggestion
    description: "Publish messages to the data source (only available with live connections).",
```

packages/studio-base/src/panels/index.ts
```diff
@@ -38,21 +38,21 @@ const builtin: PanelInfo[] = [
   {
     title: `Diagnostics – Detail`,
     type: "DiagnosticStatusPanel",
-    description: "Display data for a given diagnostics node.",
+    description: "Display ROS diagnostics data for a given node.",
```
defunctzombie, line 41:
```suggestion
    description: "Display ROS diagnostics messages for a specific hardware_id.",
```

packages/studio-base/src/panels/index.ts
```diff
@@ -87,14 +87,14 @@ const builtin: PanelInfo[] = [
...
-    description: "Publish data over a life connection.",
+    description: "Publish messages to the data source.",
     help: PublishHelp,
     module: async () => await import("./Publish"),
   },
   {
     title: "Raw Messages",
     type: "RawMessages",
-    description: "Display topic messages as JSON.",
+    description: "Inspect messages on a topic.",
```
jtbandes, line 97: "on a topic" is basically implied for every panel.
```suggestion
    description: "View message data in its original, raw form.",
```

packages/studio-base/src/panels/index.ts
esthersweon: It can only be used for numerical variables (variables can be booleans, strings, etc.). Worth calling out.

We also want to move away from calling them global variables (even though we still reference them like that in the Node Playground), since they are layout-specific.

packages/studio-base/src/panels/index.ts
```diff
@@ -38,21 +38,21 @@ const builtin: PanelInfo[] = [
   {
     title: `Diagnostics – Detail`,
     type: "DiagnosticStatusPanel",
-    description: "Display data for a given diagnostics node.",
+    description: "Display ROS DiagnosticArray messages for a given node.",
```
jtbandes, line 41: If we are using the ros data type name here, should we use the full qualified name, `diagnostic_msgs/DiagnosticArray `?

packages/studio-base/src/panels/index.ts
esthersweon: Changed to "Display a summary of all ROS DiagnosticArray messages", per your and @defunctzombie 's feedback.

packages/studio-base/src/panels/index.ts
```diff
@@ -38,21 +38,21 @@ const builtin: PanelInfo[] = [
...
-    description: "Display the status of diagnostics nodes.",
+    description: "Display a summary of all ROS DiagnosticArray messages.",
     help: DiagnosticSummaryHelp,
     module: async () => await import("./diagnostics/DiagnosticSummary"),
   },
   {
     title: "Image",
     type: "ImageViewPanel",
-    description: "Display camera feed images.",
+    description: "Display compressed or uncompressed images.",
```
jtbandes, line 55: Maybe all descriptions should mention supported datatypes (if the list is not too horribly long like it would be for the 3D panel)?

We could say sensor_msgs/Image and sensor_msgs/CompressedImage here.

packages/studio-base/src/panels/index.ts
```diff
@@ -136,14 +136,14 @@ const builtin: PanelInfo[] = [
   {
     title: "Data Source Info",
     type: "SourceInfo",
-    description: "View time and topic information for the data source.",
+    description: "View data source details like topics, datatypes, and duration.",
```
jtbandes, line 139:
```suggestion
    description: "View metadata about the current data source, including topics and timestamps.",
```

packages/studio-base/src/panels/index.ts
esthersweon: I had "live connection", but @defunctzombie recommended adding "to the data source".

Best of both worlds: "Publish messages to the data source over a live connection."

packages/studio-base/src/panels/index.ts
esthersweon: I want to keep as is, for conciseness (https://github.com/foxglove/studio/pull/2487#discussion_r774197102). The website docs and Help sidebar have full names with corresponding links, if users want to learn more.

packages/studio-base/src/panels/index.ts
esthersweon: @defunctzombie - thoughts? I personally would prefer leaving datatype names out of the tooltip, and focus more on 1-line summaries of the kind of visualizations the panel provides.

packages/studio-base/src/panels/index.ts
jtbandes: I dunno, that makes it sound like a live connection can be established for any data source. In reality the data source determines whether publishing is available or not.

packages/studio-base/src/panels/index.ts
jtbandes: Not accurate, the panel can display all diagnostics for a hardware_id, or for a specific hardware_id/name combo.

packages/studio-base/src/panels/index.ts
jtbandes: In that case we can just say "Display images" 😛

jtbandes: We should probably also mention image markers

esthersweon: I wouldn't be opposed to `Display annotated markers`.

packages/studio-base/src/panels/index.ts
defunctzombie: `data in its original, raw form` sounds awkward. Do we think `inspect` is not obvious? Maybe it isn't and I am just used to it cause of the tools I use.

I'm still a fan of `Inspect messages on a topic.`

packages/studio-base/src/panels/index.ts
defunctzombie: `View details about the current data source, including topics and timestamps`.

packages/studio-base/src/panels/index.ts
defunctzombie: 👍 

numerical variable values sounded somewhat awkward but if you goal is to avoid saying global variables I guess it works.

We could be more explicit and say `Slider to change layout variables.`

packages/studio-base/src/panels/index.ts
esthersweon: Changed to `Inspect or find the diff between messages.`

packages/studio-base/src/panels/index.ts
```diff
@@ -136,14 +136,14 @@ const builtin: PanelInfo[] = [
...
-    description: "View time and topic information for the data source.",
+    description: "View metadata like topics and timestamps for the current data source.",
     help: SourceInfoHelp,
     module: async () => await import("./SourceInfo"),
   },
   {
     title: "Variable Slider",
     type: "GlobalVariableSliderPanel",
-    description: "Quickly update numerical variable values for a layout.",
+    description: "Update numerical variable values for a layout.",
```
defunctzombie, line 146:
```suggestion
    description: "Slider to change layout variables.",
```

packages/studio-base/src/panels/index.ts
```diff
@@ -136,14 +136,14 @@ const builtin: PanelInfo[] = [
   {
     title: "Data Source Info",
     type: "SourceInfo",
-    description: "View time and topic information for the data source.",
+    description: "View metadata like topics and timestamps for the current data source.",
```
defunctzombie, line 139:
```suggestion
    description: "View details like topics and timestamps for the current data source.",
```

defunctzombie:
Added two more ideas. We don't need to use them. I think this is 👍 . We can bikeshed more later.

defunctzombie (approved):

packages/studio-base/src/panels/index.ts
```diff
@@ -87,14 +87,14 @@ const builtin: PanelInfo[] = [
...
-    description: "Publish data over a life connection.",
+    description: "Publish messages to the data source (live connections only).",
     help: PublishHelp,
     module: async () => await import("./Publish"),
   },
   {
     title: "Raw Messages",
     type: "RawMessages",
-    description: "Display topic messages as JSON.",
+    description: "Inspect or find the diff between messages.",
```
defunctzombie, line 97: `find the diff` implies some searching. `Inspect or view the difference between messages.` would be clearer. I think introducing additional aspects like `view the difference` is unnecessary. Part of the `Inspect` functionality is plotting, diffing, etc.

As a user - what is the immediate thing I am gonna get out of this panel - inspecting messages.

jtbandes: Don't all panels inspect messages?

packages/studio-base/src/panels/index.ts
esthersweon: We use verb + purpose for all descriptions.

PR #2488, opened 2021-12-22
Axis visualization fixes

jhurliman:
**User-Facing Changes**

- Connecting lines from child to parent frames are rendered more accurately for long distances

**Description**

- Remove broken memoization in useGLText() causing text labels to not appear in 3D panel
- Limit child-to-parent line distance to 5000m

Screen Shot 2021-12-22 at 1 49 52 PM


packages/studio-base/src/panels/ThreeDimensionalViz/SearchText.tsx
```diff
@@ -75,36 +75,31 @@ export const useGLText = ({
 }: WorldSearchTextProps & {
   text: Interactive<TextMarker>[];
 }): Interactive<GLTextMarker>[] => {
-  const glText: Interactive<GLTextMarker>[] = React.useMemo(() => {
```
jtbandes, line 78: Why was this useMemo causing a problem?

PR #2490, opened 2021-12-22
Remove support for LinedConvexHulls

jhurliman:
**User-facing changes**

- Removes "Line marker click events override" toggle for marker topics in the 3D panel

**Description**

Remove support for toggling "Line marker click events override", which had the description: "Optionally allow treating line markers as polygons, so that clicking inside the lines in the line marker selects the marker. The default behavior for line markers requires the user to click exactly on the line to select the line marker. This option can reduce performance and will not work on instanced line markers (those with type: 108)"

Saves 140KB off the production build size.


defunctzombie:
Seems fine.

I find it useful as a reviewer when the description notes the reasoning behind removing the change.

PR #2496, opened 2021-12-23
Per data source seekback length, set to 0.5s for remote and 2.5s elsewhere

jhurliman:
**User-Facing Changes**

- Seeking and scrubbing will read messages up to 2.5 seconds before the playhead, improving the display of low frequency markers and transforms

**Description**

seekBackNs is a new constructor arg for RandomAccessPlayer, defaulting to 2.5s and overridden to 0.5s for Ros1RemoteBagDataSource. This could be user-configured in the future.

https://user-images.githubusercontent.com/195374/147195786-98dc3b3e-008b-4960-a586-4570df5e7662.mov

### DIFF
--- a/packages/studio-base/src/dataSources/Ros1RemoteBagDataSourceFactory.tsx
+++ b/packages/studio-base/src/dataSources/Ros1RemoteBagDataSourceFactory.tsx
@@ -56,6 +56,7 @@ class Ros1RemoteBagDataSourceFactory implements IDataSourceFactory {
     return new RandomAccessPlayer(messageCacheProvider, {
       metricsCollector: args.metricsCollector,
       seekToTime: getSeekToTime(),
+      seekBackNs: BigInt(0.5e9), // 500ms
       name: url,
       urlParams: {
         url,
--- a/packages/studio-base/src/dataSources/Ros1RemoteBagDataSourceFactory.tsx
+++ b/packages/studio-base/src/dataSources/Ros1RemoteBagDataSourceFactory.tsx
@@ -56,7 +56,9 @@ class Ros1RemoteBagDataSourceFactory implements IDataSourceFactory {
     return new RandomAccessPlayer(messageCacheProvider, {
       metricsCollector: args.metricsCollector,
       seekToTime: getSeekToTime(),
-      seekBackNs: BigInt(0.5e9), // 500ms
+      // Overridden to 500ms to limit the number of blocks that need to be
+      // fetched per seek from the potentially slow remote data source
+      seekBackNs: BigInt(0.5e9),
       name: url,
       urlParams: {
         url,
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/players/RandomAccessPlayer.test.ts
```diff
@@ -35,11 +35,13 @@ import { getSeekToTime, SEEK_ON_START_NS } from "@foxglove/studio-base/util/time
 
 import RandomAccessPlayer, {
   RandomAccessPlayerOptions,
-  SEEK_BACK_NANOSECONDS,
+  DEFAULT_SEEK_BACK_NANOSECONDS,
```
defunctzombie, line 38: Maybe we drop this default entirely and make callers provide one always?

packages/studio-base/src/players/RandomAccessPlayer.ts
```diff
@@ -58,36 +59,24 @@ import {
...
   metricsCollector?: PlayerMetricsCollectorInterface;
   seekToTime: SeekToTimeSpec;
 
+  // The number of nanoseconds to seek backwards to build context during a seek
+  // operation larger values mean more opportunity to capture context before the
+  // seek event, but are slower operations. We shouldn't make this number too big,
+  // otherwise we pull in too many unnecessary messages, making seeking slow. But
+  // we also don't want it to be too low, otherwise you don't see enough data when
+  // seeking.
+  seekBackNs?: bigint;
```
defunctzombie, line 78: Could be a regular number? Unlikely this is going to be such a large number that it requires bigint?

packages/studio-base/src/dataSources/Ros1RemoteBagDataSourceFactory.tsx
```diff
@@ -56,6 +56,7 @@ class Ros1RemoteBagDataSourceFactory implements IDataSourceFactory {
     return new RandomAccessPlayer(messageCacheProvider, {
       metricsCollector: args.metricsCollector,
       seekToTime: getSeekToTime(),
+      seekBackNs: BigInt(0.5e9), // 500ms
```
defunctzombie, line 59: Add comment around why we might not want this to be too big.

defunctzombie (approved):

packages/studio-base/src/players/RandomAccessPlayer.ts
jhurliman: It could, but this matches the type that is passed into the eventual `fromNanoSec()` call. I've been using bigint for nanoseconds, apart from sec/nsec `Time` objects.

defunctzombie:
One follow-up we could do is expose this as an app setting if we think its at all useful to tune it. Maybe if someone asks for it we can propose that option.

Issue #2486, opened 2021-12-22
If there is an automatically selected target frame, we don't show it in the TF dropdown


amacneil:
**Description**

We should remove the "choose a target frame" concept. Either no frame exists, or we automatically select a frame and show it in the dropdown.

![image](https://user-images.githubusercontent.com/637671/147155308-53483888-b82e-4569-a255-67f0fd6737c3.png)

amacneil:
Unrelated note: The follow icon should go "unblue" when you are not following.

defunctzombie:
Blue icon fixed here: https://github.com/foxglove/studio/pull/2494

PR #2497, opened 2021-12-23
Usability enhancements for FollowTFControl

defunctzombie:
**User-Facing Changes**
Disabled state when there are no transforms:

Screen Shot 2021-12-22 at 10 02 25 PM

The frame name of the current render frame is shown. This indicates to the user what frame the camera is following by default.
Screen Shot 2021-12-22 at 10 02 53 PM

The user can change the frame to any other valid frame.
Screen Shot 2021-12-22 at 10 03 28 PM

The selected frame is persisted to panel save state. If the user opens a data source missing this frame the text appears in red.
Screen Shot 2021-12-22 at 10 03 52 PM

The user can select a new frame if they no longer want to use the invalid frame.
Screen Shot 2021-12-22 at 10 04 40 PM

**Description**
- When there are no transforms, the control is disabled.
- If a follow frame is not found in the current transform tree, it is shown in red text in the text box.
- If a user has not manually selected a frame, the frame for the current renderFrame is shown.

Fixes: #2486

### DIFF
--- a/packages/studio-base/src/panels/ThreeDimensionalViz/index.tsx
+++ b/packages/studio-base/src/panels/ThreeDimensionalViz/index.tsx
@@ -119,8 +119,8 @@ function BaseRenderer(props: Props): JSX.Element {
 
   const renderFrame = useMemo<CoordinateFrame>(() => {
     // If the user specified a followTf, do not fall back to any other (valid) frame
-    if (typeof followTf === "string") {
-      return transforms.frame(followTf) ?? orphanedFrame;
+    if (followTf) {
+      return transforms.frame(followTf) ?? new CoordinateFrame(followTf, undefined);
     }
 
     // Try the conventional list of transform ids
@@ -242,7 +242,7 @@ function BaseRenderer(props: Props): JSX.Element {
       config={config}
       currentTime={currentTime}
       followMode={followMode}
-      followTf={followTf}
+      followTf={renderFrame.id}
       renderFrame={renderFrame}
       fixedFrame={fixedFrame}
       resetFrame={resetFrame}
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/panels/ThreeDimensionalViz/index.tsx
```diff
@@ -119,8 +119,8 @@ function BaseRenderer(props: Props): JSX.Element {
 
   const renderFrame = useMemo<CoordinateFrame>(() => {
     // If the user specified a followTf, do not fall back to any other (valid) frame
-    if (typeof followTf === "string") {
-      return transforms.frame(followTf) ?? orphanedFrame;
+    if (followTf) {
+      return transforms.frame(followTf) ?? new CoordinateFrame(followTf, undefined);
```
defunctzombie, line 123: @jhurliman I changed this to create a coordinate frame with the `followTf` name. This results in showing the frame name with red text in the FollowTFControl and better indicates to the user they have a followTf that was manually selected but is not currently found in the tree.

jhurliman: I removed the equivalent of this code previously. It has the side effect of creating phantom frames when you switch bags or data sources.

defunctzombie: How does it create phantom frames? The effect I saw was to actually show the user their selected string in the follow control even if we haven't seen that frame in the data yet. This was more useful and correct than showing `(empty)`

defunctzombie: Chatted in Slack. This does what we expect.

Issue #2495, opened 2021-12-23
Data Source Info panel shows topic name twice instead of datatype / schemaName


jtbandes:
I believe this is a recent regression. 

**Steps to reproduce:**

- Run `npx @foxglove/ws-protocol-examples@latest image-server`
- Visit https://studio.foxglove.dev/?ds=foxglove-websocket&ds.url=ws://localhost:8765/
- Open a Data Source Info panel

**Expected:** it shows `example_image` topic with `ros.sensor_msgs.CompressedImage` datatype

**Actual:** it shows `example_image` twice, and strangely truncated:
<img width="398" alt="image" src="https://user-images.githubusercontent.com/14237/147190613-cdc5aec1-1c1d-45cf-ae6d-d73540dcd96c.png">

PR #2498, opened 2021-12-23
Fix datatype column in DataSourceInfo panel

defunctzombie:
**User-Facing Changes**
The datatype column displays the datatype again - yay.

**Description**
The datatype column incorrectly displayed the topic name rather than the datatype.

Fixes: #2495

### DIFF
--- a/packages/studio-base/src/panels/SourceInfo/index.tsx
+++ b/packages/studio-base/src/panels/SourceInfo/index.tsx
@@ -54,6 +54,15 @@ function SourceInfo() {
     [theme],
   );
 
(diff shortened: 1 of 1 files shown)


jtbandes:
Nice – so since the screenshot tests cover this, I guess it was just inadvertently accepted in some older change?

Relatedly, we should consider making the Chromatic "UI Review" a required check.

defunctzombie:
> Nice – so since the screenshot tests cover this, I guess it was just inadvertently accepted in some older change

Yep.

Issue #1758, opened 2021-09-03
Bad link color on browser compatibility warning


jtbandes:
studio.foxglove.dev in Safari:

<img width="626" alt="image" src="https://user-images.githubusercontent.com/14237/132043642-a09fa1f7-3b49-4ec2-94fc-7d06981e0c3b.png">

The blue link is too similar in color to the purple background.

Issue #1541, opened 2021-07-22
Add on-screen canvas rendering as fallback


defunctzombie:
Safari 15 is able to display most aspects of the app. One area where it fails is panels which use offscreen canvas. We should add a fallback mode to these panels to use onscreen canvas. This fallback mode could display a notification on the panel that onscreen rendering may impact performance.

- [ ] ~Remove sendNotification from util/supportOffscreenCanvas~
- [x] Add fallback to Plot Panel
- [x] Add fallback to State Transition Panel
- [x] Add fallback to Image Panel

Related https://github.com/foxglove/studio/issues/1511 https://github.com/foxglove/studio/issues/1422

PR #2502, opened 2021-12-23
Chart: add main-thread fallback for browsers without OffscreenCanvas support

jtbandes:
Partially addresses https://github.com/foxglove/studio/issues/1541 (for Plot and StateTransitions panels).

Fixes https://github.com/foxglove/studio/issues/1758.

Adapted approach from the webviz code: https://github.com/cruise-automation/webviz/blob/633920b68bdcd951b79a0bc44d0a5cf2b35814d6/packages/webviz-core/src/panels/ImageView/ImageCanvas.js#L105-L109

image

### DIFF
--- a/packages/studio-base/src/components/Chart/index.tsx
+++ b/packages/studio-base/src/components/Chart/index.tsx
@@ -10,12 +10,15 @@
 
 import { ChartOptions, ChartData as ChartJsChartData, ScatterDataPoint } from "chart.js";
 import Hammer from "hammerjs";
-import { useCallback, useLayoutEffect, useRef, useState } from "react";
+import { useCallback, useEffect, useLayoutEffect, useRef, useState } from "react";
+import { useToasts } from "react-toast-notifications";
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/components/Chart/index.tsx
```diff
@@ -81,8 +84,21 @@ type RpcSend = <T>(
...
+
+const supportsOffscreenCanvas =
+  typeof HTMLCanvasElement.prototype.transferControlToOffscreen === "function";
+
 function Chart(props: Props): JSX.Element {
   const [id] = useState(() => uuidv4());
+  const { addToast } = useToasts();
+  useEffect(() => {
+    if (!supportsOffscreenCanvas) {
+      addToast(
```
defunctzombie, line 96: Should we be showing this in the panel? Thinking about how this would happen if we used the extension api for this panel. Do we want panels to be able to make toasts?

PR #2522, opened 2021-12-24
Copy tweaks for Open dialog

amacneil:
**User-Facing Changes**
Slight changes to the wording on Open dialog

### DIFF
--- a/packages/studio-base/src/components/OpenDialog/Start.tsx
+++ b/packages/studio-base/src/components/OpenDialog/Start.tsx
@@ -79,7 +79,7 @@ export default function Start(props: IStartProps): JSX.Element {
       {
         id: "open-url",
         children: "Open file from URL",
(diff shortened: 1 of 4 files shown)


packages/studio-base/src/components/OpenDialog/Start.tsx
```diff
@@ -93,7 +93,7 @@ export default function Start(props: IStartProps): JSX.Element {
       {
         id: "sample-data",
         children: "Explore sample data",
-        secondaryText: "New to Studio? View some sample data",
```
esthersweon, line 96: Strange without period at the end, since previous sentence has punctuation.

amacneil: agreed - I think I fixed this in a later commit because latest main has an exclamation point

PR #2526, opened 2021-12-24
Remove ROS_HOSTNAME from app settings

defunctzombie:
**User-Facing Changes**
The user cannot set ROS_HOSTNAME via app settings. Since open dialog flow was enabled we were not using the ROS_HOSTNAME setting from app settings anyway so this fixes the incorrect availability of this setting at the app level.

**Description**
The ROS_HOSTNAME option is now set in the ROS1 Connection settings within the open dialog flow.

### DIFF
--- a/packages/studio-base/src/AppSetting.ts
+++ b/packages/studio-base/src/AppSetting.ts
@@ -6,7 +6,6 @@ export enum AppSetting {
   CRASH_REPORTING_ENABLED = "telemetry.crashReportingEnabled",
   MESSAGE_RATE = "messageRate",
   LAUNCH_PREFERENCE = "launchPreference",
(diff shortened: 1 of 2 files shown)


amacneil:
Is this still stored as an app setting, or is it local to the connection? Like if I had two recent connections with different ROS_HOSTNAMES would it remember them separately?

defunctzombie:
It is local per connection. So you can have two different connections for different roscores and hostnames and they will be separate in the recents list. When you click each respective connection it will use the args you specified when you setup the connection.

amacneil:
> It is local per connection. So you can have two different connections for different roscores and hostnames and they will be separate in the recents list. When you click each respective connection it will use the args you specified when you setup the connection.

Nice 👌

PR #2528, opened 2021-12-24
Add isSampleDataSource to RandomAccessPlayerOptions

defunctzombie:
**User-Facing Changes**
None

**Description**
Rather than detecting a specific url for a demo bag via isDemoBag, this change adds a new constructor option to RandomAccessPlayer for a data source factory to specify if it is for a sample data set. This option updates the metrics collector to tag sample data vs non-sample data. This will support additional sample data sets.


jtbandes:
Looks like this param was never set to true 🤔  #2949

Issue #2532, opened 2021-12-25
Add checkbox "Show on startup" to the open dialog


defunctzombie:
This checkbox allows the user to toggle whether the open dialog should show on startup.

PR #2533, opened 2021-12-25
Add "Show on startup" checkbox to Open Dialog

defunctzombie:
**User-Facing Changes**
New "Show on startup" checkbox.

Screen Shot 2021-12-24 at 8 08 10 PM

**Description**
This checkbox allows the user to toggle whether the Open Dialog shows automatically on startup or requires the user to take an action to show the dialog.

Fixes: #2532

### DIFF
--- a/packages/studio-base/src/Workspace.tsx
+++ b/packages/studio-base/src/Workspace.tsx
@@ -170,10 +170,15 @@ export default function Workspace(props: WorkspaceProps): JSX.Element {
 
   const isPlayerPresent = playerPresence !== PlayerPresence.NOT_PRESENT;
 
+  const [enableOpenDialog] = useAppConfigurationValue(AppSetting.OPEN_DIALOG);
+
+  const [showOpenDialogOnStartup = true] = useAppConfigurationValue<boolean>(
+    AppSetting.SHOW_OPEN_DIALOG_ON_STARTUP,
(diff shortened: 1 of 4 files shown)


packages/studio-base/src/components/OpenDialog/OpenDialog.tsx
```diff
@@ -180,7 +180,6 @@ export default function OpenDialog(props: OpenDialogProps): JSX.Element {
           styles: {
             content: {
               overflow: "hidden",
-              height: 480,
```
defunctzombie, line 183: @2metres With this height present, the checkbox was cutoff in the dialog.

packages/studio-base/src/Workspace.tsx
```diff
@@ -194,9 +199,11 @@ export default function Workspace(props: WorkspaceProps): JSX.Element {
     setSelectedSidebarItem(item);
   }, []);
 
-  // When a player is activated, hide the open dialog. When a player is gone, show the open dialog.
```
defunctzombie, line 197: There is no need to open the dialog when a player goes away. It's not possible for a player to go away and even if it does - the user should initiate the open dialog when that happens. This change is needed so the effect doesn't run and open the dialog when the user does not want to open the dialog on startup.

Issue #2537, opened 2021-12-25
Avoid adding duplicate file entries to Recent list


defunctzombie:
If a user opens a file via "Open file" that appears in the Recent list, avoid adding the file again. The recent entry should still become first tho.

PR #2540, opened 2021-12-26
Avoid duplicate entires in recents

defunctzombie:
**User-Facing Changes**
When opening a file that exists in recents, avoid adding another duplicate entry to the recents list.

When selecting a recent from the recents list, the recent is moved to the start of the list.

**Description**
Rework PlayerManager recents management to avoid creating multiple recents entries for the same file or connection.

This change also fixes some ux issues around recents:
- Selecting a recent puts that recent at the start of the list
- Fix a bug that overrode saved recents on immediate source selection

Fixes: #2537


packages/studio-base/src/hooks/useIndexedDbRecents.ts
```diff
@@ -0,0 +1,163 @@
...
+  save: () => Promise<void>;
+}
+
+function useIndexedDbRecents(): IRecentsStore {
+  const idbStore = useMemo(() => idbCreateStore("foxglove-recents", "recents"), []);
+
+  const { value: initialRecents } = useAsync(async () => {
+    const untypedRecents = await idbGet(IDB_KEY, idbStore);
+    return untypedRecents as RecentRecord[];
+  }, [idbStore]);
```
foxymiles, line 63: You could make this a bit more concise right?

```suggestion
  const { value: initialRecents } = useAsync(
    async () => await idbGet<RecentRecord[]>(IDB_KEY, idbStore),
    [idbStore],
  );
```

packages/studio-base/src/hooks/useIndexedDbRecents.ts
```diff
@@ -0,0 +1,163 @@
...
+
+  // Add a new recent
+  addRecent: (newRecent: UnsavedRecentRecord) => void;
+
+  // Save changes
+  save: () => Promise<void>;
+}
+
+function useIndexedDbRecents(): IRecentsStore {
+  const idbStore = useMemo(() => idbCreateStore("foxglove-recents", "recents"), []);
```
foxymiles, line 58: Would `useState` be semantically closer to what we want here?

From `useMemo` docs:

```You may rely on useMemo as a performance optimization, not as a semantic guarantee. In the future, React may choose to “forget” some previously memoized values and recalculate them on next render, e.g. to free memory for offscreen components. Write your code so that it still works without useMemo — and then add it to optimize performance.```

foxymiles (approved):
Looks good. I added some minor react/TS comments.

packages/studio-base/src/hooks/useIndexedDbRecents.ts
defunctzombie: Yea - that or useRef (tho useRef has the annoyance that it runs the construction code every time and ignores it.

Really I could move this into a global since it is a constant I suppose.

Historically I've useMemo for stuff like this and treated it like a memoize rather than a traditional "hook".

PR #2547, opened 2021-12-27
Add setting for disabling auto-update

### LINKED ISSUE #2546: Setting for disabling auto-update
I think being able to disable auto-update is a very important setting to have access to. 

**Motivation**
Right now the new version breaks the ability to open .db3 files making it useless for reading ros2 bags. I want to keep on the v0.24 version to be able to open the logs but either I reinstall foxglove every time I restart it or I use the firewall to block the network access making it not possible to auto-update. It is very annoying not being able to disable auto-updates

bertaveira:
**User-Facing Changes**
Add check box in general settings to disable auto-updater

**Description**
The objective is to have the option to not update automatically. Ideally also being able to download older versions without having them autoupdate.

It is my first time working with Electron, Typescript or any javascript related language so I am not totally sure this works well or is done properly. Sorry in advance for rookie mistakes

I wonder if this works when you download and older version of the software and the first boot the default settings are to auto-update. Since it runs the first time with default settings it might make it impossible to use an older version when a more recent one exists except if you already had the setting to not update. Again I don't fully understand the Updater so sorry for the lack of knowledge from my part.

Closes #2546


bertaveira:
I made so that the first boot when installing the software does not update it. It is not the prettiest way of doing it but it works.

amacneil:
Thanks! We had some discussion about this today (specifically whether it's ok to never check for updates on first launch).

Assigning to @defunctzombie to decide and merge.

defunctzombie:
@bertaveira I've taken a slightly different approach to the update mode after looking at some other applications and landed on https://github.com/foxglove/studio/pull/2632. Lets continue the discussion there and make sure that meets your needs.

PR #2548, opened 2021-12-28
Add drag & drop file to recents

defunctzombie:
**User-Facing Changes**
When a user drag&drops a single file into the app - that file will appear in the recents list.

**Description**
To add file items to recents we need a FileSystemFileHandle instance. This change updates the DocumentDropListener component drop event to have either a Files array or a FileSystemFileHandle array.

For now I limited this to a single handle instance since that is all our app recents feature support.

### DIFF
--- a/packages/studio-base/src/Workspace.tsx
+++ b/packages/studio-base/src/Workspace.tsx
@@ -155,7 +155,7 @@ export default function Workspace(props: WorkspaceProps): JSX.Element {
   const allowedDropExtensions = useMemo(() => {
     const extensions = [".foxe", ".urdf", ".xacro"];
     for (const source of availableSources) {
-      if (source.supportedFileTypes) {
+      if (source.type === "file" && source.supportedFileTypes) {
         extensions.push(...source.supportedFileTypes);
       }
     }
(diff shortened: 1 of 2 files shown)


defunctzombie:
https://bugs.chromium.org/p/chromium/issues/detail?id=1080811

packages/studio-base/src/components/DocumentDropListener.tsx
```diff
@@ -37,6 +37,18 @@ export default function DocumentDropListener(props: Props): JSX.Element {
         return;
       }
 
+      const dataTransferItems = ev.dataTransfer.items;
+      if (dataTransferItems.length === 1) {
```
defunctzombie, line 41: Our _recents_ list only supports showing recent entries for single files. In writing up this PR I realized this filter logic is better in Workspace.tsx. The DocumentDropListener should remain agnostic to such app limitations.

packages/studio-base/src/Workspace.tsx
```diff
@@ -328,6 +328,48 @@ export default function Workspace(props: WorkspaceProps): JSX.Element {
 
   const extensionLoader = useExtensionLoader();
 
+  const openHandle = useCallback(
+    async (handle: FileSystemFileHandle) => {
+      const file = await handle.getFile();
+      // electron extends File with a `path` field which is not available in browsers
+      const basePath = (file as { path?: string }).path ?? "";
```
defunctzombie, line 335: I don't think this is set when I drag/drop with the handle approach. This might be an issue for urdf files (I haven't found a self-contained one to test yet).

As I looked through our asset loader I continue to think we've missed the mark here and should allow users to add and manage assets within the app more explicitly.

jtbandes: Most of the files in here should work for testing. https://github.com/ros/urdf_tutorial/tree/master/urdf

packages/studio-base/src/components/DocumentDropListener.tsx
jtbandes: Does that mean you plan to change this onDropProp to take an array of handles/files?

defunctzombie: yea

packages/studio-base/src/Workspace.tsx
defunctzombie: Thanks - helped me to some sanity checking.

packages/studio-base/src/components/DocumentDropListener.tsx
```diff
@@ -19,42 +19,64 @@ import { useToasts } from "react-toast-notifications";
...
-            }
-          } else if (entry.isDirectory) {
-            directories.push(entry as FileSystemDirectoryEntry);
+
+        // Attempt to grab the filesystem handle if the feature is available.
+        // A file system handle is more versatile than File instances.
+        // Note: awaiting on the fileSystemHandle function triggered a (bug?) where all other
+        // dataTransfer items were ignored
+        if ("getAsFileSystemHandle" in item) {
+          handlePromises.push(item.getAsFileSystemHandle());
```
bunnyc1986, line 53: for some reason, this will crash Chrome when the page is not opened on HTTPS. It might be a bug in Chrome because getAsFileSystemHandle shall not be available in `item` if the page is not under a secure context.

defunctzombie: Yes - this requires loading the page in a "secure context" This is either over `https` or on `localhost`

bunnyc1986: yes. I understand. But I think it's a bug that Chrome makes it available when the page is not in a secure context. I commented on the Chrome issue: https://bugs.chromium.org/p/chromium/issues/detail?id=1080811#c15

defunctzombie: Ah - ideally you want chrome to make the method undefined so we can do feature detection and avoid crashing.

bunnyc1986: yes. just wanted to leave a trace of what I've found. so if others ran into the same problem, they know that's not caused by the Foxglove code.

Issue #800, opened 2021-04-29
Map tiles should match dark theme


jtbandes:
Consider using a dark tile set such as https://openmaptiles.org/styles/#dark-matter

defunctzombie:
This dark map has really bad contrast - I find it hard to read and would not want this to be the default map view without a way for the user to go back to a light map.

jtbandes:
Reopening the issue because I still want this feature. Regardless of whether that specific tile set is good, I think it's a reasonable request for the dark theme to use a dark tile set.

amacneil:
https://github.com/foxglove/studio/issues/2046

jtbandes:
Unrelated, but SGTM :)

jhurliman:
I think we will need to register for an API key? https://api.maptiler.com/maps/darkmatter/style.json?key=get_your_own_key_rSw2Lu595oi7U6WngsFQ

defunctzombie:
Removing myself. I am not actively working on this feature. I've commented on the referenced PR with some outstanding issues.

defunctzombie:
https://github.com/foxglove/studio/pull/3364 Allows users to select their own tile theme. I think if we want to do this we can add light and dark options to the map tile layer dropdowns and default to match the theme. This lets users easily override if they prefer a different contrast.

jtbandes:
Is there any hope of completing this issue? Map panel sticks out as one that does not really "support" dark mode by default. What would it take to add dark tiles to the dropdown?

defunctzombie:
Yea - the path forward here is to add a dark tile option to the dropdown list (before we didn't have an easy way to select different tiles via a dropdown but we do now). I still think that this is blocked on finding a dataset that is widely usable without random IP restrictions. I also feel like we have added the option to bring your own tile set so if you really want to use something else you can.

PR #2550, opened 2021-12-29
Use dark tiles in map panel when dark theme is enabled

jhurliman:
**User-facing changes**

Map panel uses dark tiles when the dark theme is enabled

**Description**

Screen Shot 2021-12-29 at 12 13 53 AM
Screen Shot 2021-12-29 at 12 14 07 AM

Fixes: #800


packages/studio-base/src/panels/Map/MapPanel.tsx
```diff
@@ -147,17 +147,29 @@ function MapPanel(props: MapPanelProps): JSX.Element {
...
     if (!mapContainerRef.current) {
       return;
     }
 
-    const tileLayer = new TileLayer("https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png", {
-      attribution: '&copy; <a href="http://osm.org/copyright">OpenStreetMap</a> contributors',
-      maxNativeZoom: 18,
-      maxZoom: 24,
-    });
+    const tileLayer =
```
defunctzombie, line 158: The layout effect this is in I think is meant to run only once on startup. Did it work to change the color scheme without restarting the app?

jhurliman: It "works", but the map panel loses all of its data and says "Waiting for GPS..." until you playback for several seconds. I assumed this was a separate bug?

defunctzombie: Kinda. The nuance is that this effect as I wrote it was really only ever meant to run once at startup and initialize the map instances and initial tile layer. So if we want to change the tile layer we should do that without invoking the entire map initialization again. This is similar to what we do with the topic layers and message layers.

defunctzombie: I pushed a commit to this PR which fixes the state issue when changing color schemes. I think this PR is ready to go now.

defunctzombie:
Here is what the stories look like on chromatic:

image

Unless someone believes this is ok for some users to see randomly when the tile server decides their IP needs a subscription I don't see how we can proceed forward with these changes.

I would be in favor of allowing the user to provide their own tile server. Or if folks can file a tile server for dark tiles that they believe is more reliable we can consider using that.

@jhurliman @amacneil @jtbandes

amacneil:
Interesting. I assume that's happening in stories due to Chromatic use of AWS infrastructure or something, it probably takes a lot of API calls before a non-cloud IP hits a paywall?

Allowing the user to select their own tile server in the panel settings seems like a good idea either way.

defunctzombie:
@amacneil So are you wanting this merged and ok if some users experience the "No active subscription" fallback or want to defer until we have the settings for different tile set?

amacneil:
Oh, I missed the fact that this "no active subscription" was new behavior in this PR. I thought it was just something OSM was giving us.

In that case, it does seem suboptimal to introduce a dependency on a new service if we don't fully understand the rate limit terms of that service, and on from first appearances seems much less forgiving than OSM. I'm not sure what the solution is (if OSM don't offer dark tiles).

Your suggestion of allowing the user to choose a tile set specifically in the panel settings might be a good idea, e.g.:

```
[x] Use theme default
[ ] Open Street Maps
[ ] Stadia Maps (Adelaide Smooth Light)
[ ] Stadia Maps (Adelaide Smooth Dark)
[ ] Custom
```

jtbandes:
Did we end up making a call on what to do with this PR?

amacneil:
We should add these as options to the new "tile layer" option in the map panel settings.

I don't think we should enable this by default, given the rate limiting concerns.

defunctzombie:
Users can specify their own custom tile server now. I think that is sufficient for those that want to use a dark tile set and have one they prefer. Closing this out since the way we'd add this is different now that we have settings.

jtbandes:
I still think that all our built in panels should support both dark and light themes to the extent possible... If the user doesn't customize settings they would expect the panel to respond to the app level theme setting.

PR #2555, opened 2022-01-01
New message prototype: foxglove_msgs/MapTile and foxglove_msgs/MapTileArray

jhurliman:
**User-Facing Changes**

- Adds support for `foxglove_msgs/MapTile` and `foxglove_msgs/MapTileArray` message types

**Description**

This is a sandbox PR for experimenting with a new message type that describes a textured surface oriented in space, with an optional heightmap texture. Currently, the `MapTile` message looks like this:

```
{
  header: std_msgs/Header;
  pose: geometry_msgs/Pose;
  resolution: float32;
  width: uint32;
  height: uint32;
  albedo: { format: "png" | "jpeg", data: uint8[] },
  elevation: { format: "png" | "jpeg", data: uint8[] },
  elevation_scale: float32;
}
```

With a `MapTileArray` message that is a single array field of `MapTile`s. Some open questions and notes:

* Should map tiles be more like markers, uniquely identified by `ns`+`id` and with ADD/REMOVE/REMOVEALL actions? To give the publisher control of which tiles appear on the screen at which time, we can either use the marker style system or just use `MapTileArray` and have it replace all of the map tiles (on that topic) each time it is published. Similar to the difference between markers and occupancy grids / pointclouds / laserscans.
* Which image formats should `albedo` and `elevation` support? Raw rgb inline, PNG/JPEG inline, url. Do we need separate message types for some or all of these?
* What should be rendered after a message is received but before the images have loaded? A flat grey rectangle? Nothing?
* TODO: How do we make the scene update when a texture has finished asynchronously loading?
* TODO: Make albedo and heightmap optional, handle `albedo: []` or `elevation: []` (setting both to empty should be a topic error)
* TODO: Fix the caching logic, nothing is purging the cache so this will likely run out of memory quickly
* TODO: Can the error handling be migrated away from sendNotification()?

Screen Shot 2021-12-31 at 4 26 21 PM


defunctzombie:
Could you open an accompanying PR on our website repo documenting this message and its fields similar to how we documented the LocationFix message. We should use this PR and that documentation PR to work through your above questions.

We need to similarly add support for the protobuf message name variants. Going forward I'd like any new message support to always be accompanied by a documentation PR to website and support the protobuf names.

packages/studio-base/src/panels/ThreeDimensionalViz/commands/MapTiles.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+    attributes: {
+      point: (_context: unknown, props: MapTileMessage) => {
+        const vertices = new Float32Array(props.width * props.height * 18);
+        // Generate a triangle list for the props.width x props.height grid
+        for (let yi = 0; yi < props.height; yi++) {
+          for (let xi = 0; xi < props.width; xi++) {
+            const i = (yi * props.width + xi) * 18;
+            const x1 = xi / props.width;
+            const x2 = (xi + 1) / props.width;
```
jtbandes, line 93: So are the height values specifying the height of the x<sub>i</sub> vertices as opposed to the height of the whole [x<sub>i</sub>, x<sub>i+1</sub>] region? If there are only `width*height` quads and also `width*height` height values, what is the height of the vertices in the last row/column?

jhurliman: My intention was to replicate the semantics of `OccupancyGrid` with width and height.

jtbandes: But occupancy grids specify whole cells, whereas we need to set the height of each vertex, and if there are n cells per side then there are n+1 vertices per side. Or does OccupancyGrid address this somehow?

jhurliman: Good point, this will need to be addressed somehow.

jtbandes: I just remembered this issue (thanks stalebot) and it is not addressed by foxglove.Grid...

jtbandes: Can anyone figure out how grid_map_rviz_plugin is handling this issue? The code is a bit hard to read... https://github.com/ANYbotics/grid_map/blob/ba52294cbd28223dbbe9d1210dde7eb978944ee8/grid_map_rviz_plugin/src/GridMapVisual.cpp#L168

amacneil:
What's the plan for this PR? Is it going to land or is it going to get folded into new3d?

jtbandes:
But the plan is to use the new foxglove.Grid type from BYOD, right?

For height data we will still need to resolve this issue: https://github.com/foxglove/studio/pull/2555#discussion_r777622949

jtbandes:
Going to say this is closed by #4544.
Filed https://github.com/foxglove/studio/issues/4629 to track height support for Grid rendering.

PR #2445, opened 2021-12-18
Build docker image GitHub action

bertaveira:
**Motivation**
As we saw the necessity to have an image up to date with the foxglove I decided to make it automatically build the image and push it. It has worked for us (although in gitlab not github) so I thought it would be nice to make it official and make it public so everyone can have access to an official, automatically released and up to date image. 

The idea would be to have an official foxglove account on docker hub and replace my personal one in the CI so that on every release it builds the image and tags with the version number as well as replaces the latest tag.

**Description**
What this pull does is add a new Github action job where it builds the docker image for multi-architecture and then pushes it to docker hub. Right now it is pushing ot my personal account but if this is of interest to foxglove as much as it is to me the idea would be to make it push to an official docker hub account of foxglove.

Currently it is building for 2 platforms but more could be added:
- linux/amd64 - Regular Linux x86
- linux/arm64 - The ARM version supported by ROS official docker images (ARM v8 such as Apple M1 and latest Raspberry Pi's)

The images built are tagged with the version number as well as the `:latest` tag. This way the latest is always the most recent version of foxglove studio, but everyone can also choose a specific version if they so desire.

**Tests**
To test if the action was working i did some releases on my fork in a different branch. The reason I did them on a different branch is because the release CI also commits to the branch ("Bump dev" job). Since I didn't want this pull messing with versions due to my tests, they were done in a separate branch called ` bernardo/build-docker-action` (sorry for the confusing branch names). You can check the actions and releases I made to validate that it works as well as my docker hub repository ([here](https://hub.docker.com/repository/docker/bertaveira/foxglove/general)).

To note that building both images (amd64 and arm64) takes about 1h.

**Adaptations needed**

1. Create a foxglove docker account
2. Add docker hub credentials as secret variables on github (`DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN`) and replace name of docker image in the `post-release.yml` file. I can do this last step on my branch but the account needs to be created first.

PR #2556, opened 2022-01-01
Add note about self-hosting to readme

amacneil:
**User-Facing Changes**
Documenting Docker image from https://github.com/foxglove/studio/pull/2445

@esthersweon do you think we should document on our website too?


esthersweon:
@amacneil - will mention briefly in website docs, then link to GitHub.

PR #2558, opened 2022-01-03
Fix "Each child in a list should have a unique key prop" error

esthersweon:
**User-Facing Changes**
None

**Description**
- Got rid of unnecessary fragment in PanelList to get rid of "Each child in a list should have a unique key prop" error


defunctzombie:
👍

Issue #2539, opened 2021-12-25
Cannot double-click to open a bag file on MacOS when Studio is open.


defunctzombie:
**Description**
Double-clicking a .bag file in finder does not open the file in Studio if Studio is already open.

- Version: v0.25
- OS: MacOS

**Steps To Reproduce**

- Open Studio and close the window (this leaves Studio still open with no windows)
- Open Finder to a folder with a .bag file you want to open.
- Double click the bag file

Studio fails to open the bag file.

Note: If Studio is fully closed (i.e. File -> Quit) then double clicking the .bag file does launch Studio and open the bag.

**Expected Behavior**
Studio opens with the bag file.

PR #2559, opened 2022-01-03
Fix double-click file open on MacOS with no windows

defunctzombie:
**User-Facing Changes**
A user on MacOS can close all windows then double click a bag file to open it in a new window.

**Description**
With MacOS it is possible to close all active windows but keep the application open. If this happened, our desktop app would stop handling double-click file open actions since there was no window to handle the action.

This change fixes the behavior by opening a window if there are no windows.

Fixes: #2539

### DIFF
--- a/desktop/main/injectFilesToOpen.ts
+++ b/desktop/main/injectFilesToOpen.ts
@@ -1,7 +1,11 @@
 // This Source Code Form is subject to the terms of the Mozilla Public
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
-import { BrowserWindow } from "electron";
+import { Debugger, dialog } from "electron";
+
+import Logger from "@foxglove/log";
+
+const log = Logger.getLogger(__filename);
 
 /*
  * Our app has support for working with File instances in the renderer. This avoids extra copies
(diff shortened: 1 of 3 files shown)


desktop/main/injectFilesToOpen.ts
```diff
@@ -35,8 +38,12 @@ export default async function injectFilesToOpen(
       files: filesToOpen,
     });
 
+    log.debug(`Set input files #${inputElementId}: ${filesToOpen.join(",")}`);
+
     // clear the files once we've opened them
     filesToOpen.splice(0, filesToOpen.length);
+  } catch (err) {
+    log.error(err);
```
jtbandes, line 46: should we pop up an alert box so the user knows something went wrong instead of a silent failure?

defunctzombie: Using this? https://www.electronjs.org/docs/latest/api/dialog

jtbandes: yeah something like that. Since the user probably expected something to happen if they opened a file.

desktop/main/injectFilesToOpen.ts
```diff
@@ -35,8 +43,16 @@ export default async function injectFilesToOpen(
...
+    dialog.showErrorBox(
+      "Internal error",
+      `The app encountered an internal error trying to open: ${filesToOpen.join(",")}`,
```
jtbandes, line 54: shouldn't we include the error message as well?

defunctzombie: Yea - I guess it is hard to get the log otherwise for ore debugging. Tho not really sure what action these errors will present us with. Maybe we should be sending to sentry rather than showing user?

jtbandes: Yeah, probably both.

PR #2560, opened 2022-01-03
Fix chartjs-plugin-zoom checksum

amacneil:
**User-Facing Changes**
None

**Description**
`chartjs-plugin-zoom` was causing `yarn install` errors, locally and on Vercel since 1/1/2022. It would probably fail on CI too, except that CI had a cached yarn install.

Fixed by rerunning locally and committing the changed checksum. No idea why it actually changed though - we should probably stop using git dependencies.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -8751,7 +8751,7 @@ __metadata:
     hammerjs: ^2.0.8
   peerDependencies:
     chart.js: ^3.0.0-beta.13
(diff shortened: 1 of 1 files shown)


jtbandes:
> No idea why it actually changed though

https://github.com/foxglove/chartjs-plugin-zoom/blob/656541279943f00dce600288435f83c436293146/rollup.config.js#L16

defunctzombie:
> https://github.com/foxglove/chartjs-plugin-zoom/blob/656541279943f00dce600288435f83c436293146/rollup.config.js#L16

SUCH MAGIC

PR #2563, opened 2022-01-03
Fix OpenDialog height

defunctzombie:
**User-Facing Changes**
The OpenDialog does not change height when selecting different views.

**Description**
The OpenDialog height should be consistent across view changes so the back button and title remain in the same place on screen.


defunctzombie:
Regressed in https://github.com/foxglove/studio/pull/2533/files#r775094829

jtbandes:
This preview deployment exhibits some weird behavior with vertical resizing where the checkbox disappears/reappears suddenly, and is sometimes cut off without being scrollable:

https://user-images.githubusercontent.com/14237/147991934-e7ee1c2f-e2eb-4422-ba0e-149ec081ac13.mov

Issue #2551, opened 2021-12-29
Cannot advertise or publish from extensions api when using ROS1


ccpowers:
When using the extensions API to create an extension with a publisher, the publisher is never advertised, and nothing is published, when using a ROS 1 connection. Switching to using a rosbridge connection allows the publisher to be advertised and publish.

- Version: 0.19.0, Linux desktop app
- OS: Ubuntu 16.04

**Steps To Reproduce**
1. Create an example extension panel following the docs here: https://foxglove.dev/docs/extensions/getting-started
2. Add a button that publishes "Hello world!" with the following modifications to `ExamplePanel.tsx`:
```
import { PanelExtensionContext, RenderState } from "@foxglove/studio";
import { useLayoutEffect, useEffect, useState, useCallback } from "react";
import ReactDOM from "react-dom";

function PublisherExample({ context }: { context: PanelExtensionContext }): JSX.Element {
  const [renderDone, setRenderDone] = useState<(() => void) | undefined>();

  // We use a layout effect to setup render handling for our panel. We also setup some topic subscriptions.
  useLayoutEffect(() => {
    // The render handler is run by the broader studio system during playback when your panel
    // needs to render because the fields it is watching have changed. How you handle rendering depends on your framework.
    // You can only setup one render handler - usually early on in setting up your panel.
    //
    // Without a render handler your panel will never receive updates.
    //
    // The render handler could be invoked as often as 60hz during playback if fields are changing often.
    context.onRender = (renderState: RenderState, done) => {
      // render functions receive a _done_ callback. You MUST call this callback to indicate your panel has finished rendering.
      // Your panel will not receive another render callback until _done_ is called from a prior render. If your panel is not done
      // rendering before the next render call, studio shows a notification to the user that your panel is delayed.
      //

      console.log(`Got render state: ${renderState}`);
      // Set the done callback into a state variable to trigger a re-render.
      setRenderDone(done);
    };

    // After adding a render handler, you must indicate which fields from RenderState will trigger updates.
    // If you do not watch any fields then your panel will never render since the panel context will assume you do not want any updates.

    // tell the panel context that we care about any update to the _topic_ field of RenderState
    context.watch("topics");

    // tell the panel context we want messages for the current frame for topics we've subscribed to
    // This corresponds to the _currentFrame_ field of render state.
    context.watch("currentFrame");

    // subscribe to some topics, you could do this within other effects, based on input fields, etc
[...]

PR #2574, opened 2022-01-05
Queue ROS1 advertise requests until ROS1 node is ready.

foxymiles:
**User-Facing Changes**
This improves our handling of ROS1 `advertise` calls in extension panels.

**Description**
Currently if any `advertise` calls are made to the Ros1Player before our node is connected are silently ignored. The fix is to cache the advertise requests and try them again once the node is ready, in the same way we currently handle subscriptions.

Partially addresses #2551


defunctzombie:
👍

Issue #2577, opened 2022-01-05
Playback speed control not working when loading from data platform


ccpowers:
**Description**
When playing back a bag loaded from the data platform, the playback speed controls do nothing. If I load a bag stored locally, they do what I expect. 

- Version: Mac Desktop 0.25.0, also Chrome
- OS: MacOS, also happens in Chrome on MacOS

**Steps To Reproduce**
1. Load a bag from the timeline view of data platform
2. Attempt to change the playback speed

https://user-images.githubusercontent.com/477928/148280268-bc9db58b-ffa4-49db-ab11-2876e4b76f61.mp4

**Expected Behavior**
Playback changes speed, and UI updates to reflect this.

**Actual Behavior**
Playback appears to remain at 1x (time advancing 1 second per clock second) and the dropdown option does not change, even if you click on different speeds.

jtbandes:
Thanks for the report!

ccpowers:
Thanks for the quick follow up! Is there a specific release cadence that I can use to figure out when this will be available, or is it CI-d into the web version or anything?

amacneil:
The web version hosted at [studio.foxglove.dev](https://studio.foxglove.dev/) is always up to date with main, we deploy continuously.

Desktop releases happen every 1-2 weeks (sometimes sooner if we have high priority bugs to deploy). This PR was included in v0.26 - we keep release notes [here](https://github.com/foxglove/studio/releases) 🙂.

PR #2579, opened 2022-01-05
Add setSpeed capability to Data Platform Player; disable playback speed dropdown when setSpeed is not supported

jtbandes:
**User-Facing Changes**
Fixes an issue where setting playback speed did not work with data streamed from Foxglove Data Platform.

**Description**
Fixes https://github.com/foxglove/studio/issues/2577

### DIFF
--- a/packages/studio-base/src/players/FoxgloveDataPlatformPlayer/index.ts
+++ b/packages/studio-base/src/players/FoxgloveDataPlatformPlayer/index.ts
@@ -29,6 +29,7 @@ import {
   AdvertiseOptions,
   MessageEvent,
   Player,
+  PlayerCapabilities,
   PlayerMetricsCollectorInterface,
   PlayerPresence,
   PlayerProblem,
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/players/FoxgloveDataPlatformPlayer/index.ts
```diff
@@ -50,7 +51,7 @@ import streamMessages from "./streamMessages";
 
 const log = Logger.getLogger(__filename);
 
-const CAPABILITIES: string[] = ["playbackControl"];
+const CAPABILITIES = [PlayerCapabilities.playbackControl, PlayerCapabilities.setSpeed];
```
defunctzombie, line 54: Do we have a type for CAPABILITIES to catch the error in the future?

defunctzombie (approved):

packages/studio-base/src/players/FoxgloveDataPlatformPlayer/index.ts
jtbandes: The error was a missing value. This is not possible to catch with the type checker.

PR #2584, opened 2022-01-06
Add thumbnails to Panel List tooltip

2metres:
**User-Facing Changes**

https://user-images.githubusercontent.com/924528/148320226-a04a8e9f-bda1-42f2-a228-4c000d217f84.mov

**Description**
Show simplified thumbnails on hovering a panel in the panel list

### DIFF
--- a/packages/studio-base/src/panels/index.ts
+++ b/packages/studio-base/src/panels/index.ts
@@ -5,27 +5,45 @@ import { PanelInfo } from "@foxglove/studio-base/context/PanelCatalogContext";
 import { TAB_PANEL_TYPE } from "@foxglove/studio-base/util/globalConstants";
 
 import GlobalVariableSliderHelp from "./GlobalVariableSlider/index.help.md";
+import globalVariableSliderThumbnail from "./GlobalVariableSlider/thumbnail.png";
 import ImageViewHelp from "./ImageView/index.help.md";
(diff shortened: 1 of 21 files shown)


packages/studio-base/src/panels/index.ts
```diff
@@ -5,69 +5,93 @@ import { PanelInfo } from "@foxglove/studio-base/context/PanelCatalogContext";
...
+import diagnosticStatusThumbnail from "./diagnostics/thumbnails/diagnostic-status.png";
+import diagnosticSummaryThumbnail from "./diagnostics/thumbnails/diagnostic-summary.png";
 
 const builtin: PanelInfo[] = [
   {
     title: "3D",
     type: "3D Panel",
     description: "Display visualization markers and models in a 3D scene.",
     help: ThreeDimensionalVizHelp,
+    thumbnail: threeDimensionalVizThumbnail,
```
defunctzombie, line 54: Is this thumbnail field suppose to be the actual image or a url to the thumbnail image?

Might want to make sure it's only the url? Otherwise all those will be imported into this one index.ts file

jtbandes: It looks like it is being set as a `src` attribute later.

defunctzombie:
👍

Issue #2570, opened 2022-01-04
"Slowness" with loading remote bag


defunctzombie:
I noticed that our remote bag handling has gotten slow compared to _previous_ versions.

Open studio.foxglove.dev and click on "explore sample data". This loads our demo bag using the remote bag reader logic. Previously this was faster.

I've not researched when this changed or profiled the current code to identify what is slower.

The slowness also contributes to the app feeling like it has "stalled" at times when seeking the playback bar.

I've marked this as a bug since it looks like a regression to me and possibly something nefarious is happening with our remote loading.

PR #2586, opened 2022-01-06
Provide our own shim for process global

defunctzombie:
**User-Facing Changes**
Remote bag files load faster.

**Description**
Rather than use the npm process module, we use our own shim for process. This allows us to use a faster implementation for process.nextTick

Fixes: #2570

### DIFF
--- a/ci/lint-unused-exports.ts
+++ b/ci/lint-unused-exports.ts
@@ -11,7 +11,7 @@ import { execOutput } from "./exec";
 // An export is considered unused if it is never imported in any source file.
 //
 // Note: use the "// ts-prune-ignore-next" comment above an export if you would like to mark it
-// as used even tho it appears unused. This might happen for exports which are injected via webpack.
+// as used even though it appears unused. This might happen for exports which are injected via webpack.
 async function main(): Promise<void> {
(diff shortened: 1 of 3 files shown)


amacneil:
I feel like we’re nearly at a point where we should set up some perf test infrastructure to catch future regressions in this type of code.

defunctzombie:
> I feel like we’re nearly at a point where we should set up some perf test infrastructure to catch future regressions in this type of code.

Its not clear if this is a regression or we just never really noticed from day 1.

ci/lint-unused-exports.ts
```diff
@@ -6,6 +6,12 @@ import { info } from "@actions/core";
 
 import { execOutput } from "./exec";
 
+// Identify unused exports
+//
+// An export is considered unused if it is never imported in any source file.
+//
+// Note: use the "// ts-prune-ignore-next" comment above an export if you would like to mark it
+// as used even tho it appears unused. This might happen for exports which are injected via webpack.
```
jtbandes, line 14:
```suggestion
// as used even though it appears unused. This might happen for exports which are injected via webpack.
```

jtbandes (approved):

PR #2588, opened 2022-01-06
Fix startup state url handling

defunctzombie:
**User-Facing Changes**
When loading a data source via url, the url bar does not flicker and remove the source only to add it back.

**Description**

useInitialDeepLinkState is updated to only apply url state once. This is accomplished by starting with the initial url state and removing the items already applied. With this change, we also track whether the _time_ field should be applied. If the url state contains a time field but no data source, we should not apply the time field when the user selects a datasource manually.

useStateToURLSynchronization is updated to ignore startup undefined values which resulted in clearing the url bar values and quickly adding the values back. The logic is updated to remove values that do become undefined after having been defined before.

### DIFF
--- a/packages/studio-base/src/hooks/useInitialDeepLinkState.ts
+++ b/packages/studio-base/src/hooks/useInitialDeepLinkState.ts
@@ -2,8 +2,7 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
-import { useMemo } from "react";
-import { useMount } from "react-use";
+import { useLayoutEffect, useRef } from "react";
 
 import Log from "@foxglove/log";
 import { AppSetting } from "@foxglove/studio-base/AppSetting";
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/hooks/useInitialDeepLinkState.ts
```diff
@@ -26,32 +26,32 @@ const log = Log.getLogger(__filename);
...
-  }, [deepLinks]);
+  }
+
+  const shouldSeekTimeRef = useRef(false);
 
   // Set a sessionStorage preference for web if we have a stable URL state.
   // This allows us to avoid asking for the preference immediately on
   // launch of an empty session and makes refreshes do the right thing.
-  useEffect(() => {
+  useLayoutEffect(() => {
```
foxymiles, line 54: Does this need to be a `useLayoutEffect`?

defunctzombie: As opposed to a useEffect? I think the only difference between the two is semantics on when they run. I've trended towards useLayoutEffect tho maybe that is misguided. @jtbandes thoughts?

defunctzombie:
Updated to useEffect

PR #2591, opened 2022-01-06
Fix imageview flickering during zoom.

### LINKED ISSUE #2516: Flickering when zooming image panel
**Description**
- nuscenes bag
- add cam front + lidar annotations
- try zooming in and out

current web: lots of flickering
v0.24: no flickering

https://user-images.githubusercontent.com/637671/147296299-e638166c-898a-43b4-a6ff-969d28dd3085.mov

foxymiles:
**User-Facing Changes**
This fixes the flickering in the image view during zooming.

**Description**
The fix here is easy. We were trying to assign a fractional width & height to the canvas so we were effectively clearing the canvas unnecessarily on each render. All we need to do is make sure we pass integer values for the canvas viewport width and height.

Fixes #2516


defunctzombie:
🪄

PR #2592, opened 2022-01-06
Prevent zooming below a threshold in Image Panel.

### LINKED ISSUE #2514: Prevent image panel zooming out more than anything rendered in that image panel
If we have an image and some markers in the image panel, we don't really need the ability to zoom out more than the edge of that image + markers.

Today it's pretty easy to zoom far out and get lost in space.

One idea is to just prevent zooming out to a level where the image is less than 50% of the height of the panel (i.e. the image panel canvas could be 2x the size of the image).

image

foxymiles:
**User-Facing Changes**
This prevents zooming below a fixed factor in the Image Panel to make it harder for users to lose track of the content at tiny zoom levels.

**Description**
This simply hard caps the minimum zoom at 0.25.

Fixes #2514

### DIFF
--- a/packages/studio-base/src/panels/ImageView/ImageCanvas.tsx
+++ b/packages/studio-base/src/panels/ImageView/ImageCanvas.tsx
@@ -314,7 +314,7 @@ export default function ImageCanvas(props: Props): JSX.Element {
     setContainer,
     panZoomHandlers,
   } = usePanZoom({
-    minZoom: 0.25,
+    minZoom: 0.5,
     initialPan: config.pan,
     initialZoom: config.zoom,
   });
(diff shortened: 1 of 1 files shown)


amacneil:
Testing this locally I think even 0.5 is reasonable. I can't think of a reason why anyone needs to zoom out farther than that?

cc @jhurliman @defunctzombie @jtbandes @2metres for input

defunctzombie:
No strong thoughts. Easy enough to tweak later if we determine otherwise.

foxymiles:
> Testing this locally I think even 0.5 is reasonable. I can't think of a reason why anyone needs to zoom out farther than that?
> 
> cc @jhurliman @defunctzombie @jtbandes @2metres for input

I've increased it 0.5. What do you think?

defunctzombie:
LGTM - ship it.

PR #2596, opened 2022-01-06
Fix PointCloud2 rgb/rgba fields when is_bigendian = false

jhurliman:
**User-Facing Changes**

Visualizing PointCloud2 messages in the 3D panel with an rgb/rgba field will use ABGR order if is_bigendian = false, instead of BGRA

**Description**

The PointCloud2 rendering is currently not alpha-aware, even though color fields are four bytes that could represent red, green, blue, and alpha. A side effect of not paying attention to alpha is that we toggled between RGB and BGR based on the is_bigendian field, but this caused us to pause the four bytes of color fields as either RGBA or BGRA. The little endian equivalent of RGBA is ABGR, not BGRA. A story has been added to verify the parsing behavior.


packages/studio-base/src/panels/ThreeDimensionalViz/commands/PointClouds/buffers.ts
```diff
@@ -194,32 +186,19 @@ export function createColorBuffer({
...
     const rgbOffset = rgbField.offset ?? 0;
-    if (hasValidStride(FLOAT_SIZE * stride)) {
-      return {
-        // RGB colors are encoded in a single 4-byte tuple and unfortunately we cannot extract
-        // them in shaders by just reinterpreting the data buffer.
-        // In addition, the supported WebGL implementation constraints VBOs to be of type float,
-        // so we cannot send the data as is. Then, we're converting the data array into a float
-        // array, transforming each byte value into a float.
-        // We're definitely paying a memory cost here
-        buffer: expandBufferToFloat(data),
```
jtbandes, line 205: I don't think we should be deleting this, at least without updating other comments — `extractValues` says "This is a fallback function" but we now seem to be using it unconditionally.

packages/studio-base/src/panels/ThreeDimensionalViz/commands/PointClouds/buffers.ts
```diff
@@ -175,12 +165,14 @@ export function createColorBuffer({
...
   pointCount,
   stride,
+  isBigEndian,
 }: {
   data: Uint8Array;
   fields: FieldOffsetsAndReaders;
   colorMode: ColorMode;
   pointCount: number;
   stride: number;
+  isBigEndian: boolean;
```
jtbandes, line 175: Do other parts of this work with big endian data? If not, it seems weird to have it in a partially supported state. I would just assume little endian and track the big endian work in #2225.

packages/studio-base/src/panels/ThreeDimensionalViz/commands/PointClouds/buffers.ts
jhurliman: I just re-read the documentation for `extractValues` and compared it against the three places in the code where it's called. I think fallback is still a correct term, even though for the case of an "rgb" field it is unconditionally called.

packages/studio-base/src/panels/ThreeDimensionalViz/commands/PointClouds/buffers.ts
jhurliman: The previous code partially supported big-endian data, and that partial support was the root cause of this bug. I can tear out all references to the is_bigendian field, but that refactor will be a larger effort and move us further away from having big-endian support.

Issue #2593, opened 2022-01-06
Recent list isn't working for me


amacneil:
**Description**
😞 
![image](https://user-images.githubusercontent.com/637671/148452605-09158e74-021d-4b5e-9a63-71e0c4e19148.png)

- Version: web

PR #2597, opened 2022-01-06
Fix recent list for new users

defunctzombie:
**User-Facing Changes**
Recents list correctly saves recent entries for users that do not have any recent items.

**Description**
A logic error in useIndexedDbRecents prevented the recent list from saving to the database. This change fixes the error and adds tests to ensure recents are saved as expected.

Fixes: #2593


foxymiles (approved):
Ship it!

jtbandes (approved):

PR #2598, opened 2022-01-07
Support multiple Foxglove Studio processes in ROS1/2 native data sources

jhurliman:
**User-Facing Changes**

Support multiple Foxglove Studio processes in ROS1/2 native data sources

**Description**

New node name: `/foxglovestudio_${os.pid}`

### DIFF
--- a/packages/studio-base/src/players/Ros1Player.ts
+++ b/packages/studio-base/src/players/Ros1Player.ts
@@ -114,7 +114,7 @@ export default class Ros1Player implements Player {
 
     if (this._rosNode == undefined) {
       const rosNode = new RosNode({
-        name: "/foxglovestudio",
+        name: `/foxglovestudio_${os.pid}`,
         hostname,
         pid: os.pid,
         rosMasterUri: this._url,
--- a/packages/studio-base/src/players/Ros2Player.ts
+++ b/packages/studio-base/src/players/Ros2Player.ts
@@ -113,7 +113,7 @@ export default class Ros2Player implements Player {
 
     if (this._rosNode == undefined) {
       const rosNode = new RosNode({
-        name: "/foxglovestudio",
+        name: `/foxglovestudio_${os.pid}`,
         domainId: this._domainId,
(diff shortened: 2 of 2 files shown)


packages/studio-base/src/players/Ros1Player.ts
```diff
@@ -114,7 +114,7 @@ export default class Ros1Player implements Player {
 
     if (this._rosNode == undefined) {
       const rosNode = new RosNode({
-        name: "/foxglovestudio",
+        name: `/foxglovestudio_${os.pid}`,
```
defunctzombie, line 117: What happens when the user launches a new ROS1 connection with the same studio instance? Does that make advertising no longer work since the roscore will still have the old connection info (xmlrpc server ports, advertise ports, etc?)

jhurliman: Has that behavior been reported? This change won't affect launching new ROS1 connections with the same studio instance, so it will match the current behavior.

defunctzombie: Figured while you were changing this it's easy to test and add a random number to make that use case work since we are aware it can be an issue.

jhurliman: cc @foxymiles do you have context here?

foxymiles: @jhurliman Generating a unique node name isn't necessary for refresh/relaunch of studio to take over previously published topics. I'd need to check to see how this affects extension panels.

defunctzombie: @jhurliman another think to check. On macos multiple windows are supported for the same "app" instances. will all those windows have a different PID?

jhurliman: @defunctzombie that is a single PID with several Chromium instances. We don't currently support that AFAIK, but if we did this would need some additional thinking (along with sanity checking electron-socket works correctly).

PR #2600, opened 2022-01-07
Only show a single value + overflow message in plot tooltip.

### LINKED ISSUE #2367: Hovering a dense plot panel shows multiple points for a single series.
image

Users have noted this is counterintuitive to their experience and expectation when hovering plot panels. They expect to see just one point for each series - not multiple points.

My original thinking on this was to surface to the user that their plot data is dense and that their hover line is actually hovering over multiple points due to how dense the plot data is.

One approach is to show a single value for each series and then show some notice like _multiple values hovered_ or _zoom in to see additional values_ to indicate data is dense or downsampled.

A PR not yet released did make the tooltip more compact when multiple values are below the hover line: https://github.com/foxglove/studio/pull/2274

foxymiles:
**User-Facing Changes**
Currently when a user hovers over a position in the plot view corresponding to multiple data points we show all the values in the tooltip. This collapses that into a single value plus an overflow message.

Screen Shot 2022-01-07 at 8 08 38 AM

**Description**
The logic here is straightforward. We just show the first value only and an overflow message if needed.

Fixes #2367

### DIFF
--- a/packages/studio-base/src/components/TimeBasedChart/TimeBasedChartTooltipContent.tsx
+++ b/packages/studio-base/src/components/TimeBasedChart/TimeBasedChartTooltipContent.tsx
@@ -12,6 +12,7 @@
 //   You may not use this file except in compliance with the License.
 
 import { makeStyles } from "@fluentui/react";
+import { take } from "lodash";
 import { PropsWithChildren, useMemo } from "react";
 
 import { fonts } from "@foxglove/studio-base/util/sharedStyleConstants";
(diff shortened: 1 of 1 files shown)


defunctzombie:
👍  - maybe some note to zoom in to see more? Could also show this multiple values notice once at the bottom of the overall tooltip tho maybe it is useful to know there are multiple values for a specific series?

foxymiles:
> 👍 - maybe some note to zoom in to see more? Could also show this multiple values notice once at the bottom of the overall tooltip tho maybe it is useful to know there are multiple values for a specific series?

Yeah makes sense. It's possible to have multiple values for one series but not others right?

defunctzombie:
> It's possible to have multiple values for one series but not others right?

yep

Issue #2199, opened 2021-11-22
URDF viz disappears when toggling between tabs


jhurliman:
**Description**

Create a layout with two or more tabs, at least one containing a 3D scene. Set the 3D Model in that scene to a valid URDF URL so the robot model appears. Switch tabs, then switch back. The model and any transform visualizations loaded from the URDF will have disappeared until you change the camera coordinate frame.

- Version: `db448473a975cdd3da17d829a8e86d8de532368c `
- OS: All

defunctzombie:
@jhurliman sending your way - I think you added the URDF feature to the 3d panel so hoping you have some context on tackling this.

defunctzombie:
Spoke with @jhurliman - gonna grab this since it might be a strange react-ism we need to apply to resolve this.

PR #2602, opened 2022-01-07
Fix transform loading from URDF robot models in 3d panel

defunctzombie:
**User-Facing Changes**
The 3d scene updates with newly added transforms when setting the robot model.

**Description**

URDF models may contain joints. These joints represent transforms between different frames of the robot. 3D panel is able to add this frame information to the transform tree.

This change fixes update behavior in the 3d panel when loading URDF models with joints. Previously, the UrdfBuilder would add the joint transforms directly to the transform tree. This approach did not inform the react loop to run or for the transform tree to invalidate itself and create a new instance (as required by hook dependencies).

Instead, UrdfBuilder is now an event emitter that emits new transforms. The 3d panel components register event handlers and pass along the urdf transform information back through useTransforms. All other uses of the transform tree should be immutable. Only useTransforms can mutate the transform tree.

Fixes: #2199

### DIFF
--- a/packages/studio-base/src/panels/ThreeDimensionalViz/UrdfBuilder.ts
+++ b/packages/studio-base/src/panels/ThreeDimensionalViz/UrdfBuilder.ts
@@ -11,6 +11,7 @@
 //   found at http://www.apache.org/licenses/LICENSE-2.0
 //   You may not use this file except in compliance with the License.
 
+import { EventEmitter } from "eventemitter3";
 import { quat, vec3 } from "gl-matrix";
 import { isEqual } from "lodash";
 
@@ -29,9 +30,9 @@ import { rewritePackageUrl } from "@foxglove/studio-base/context/AssetsContext";
 import { TopicSettingsCollection } from "@foxglove/studio-base/panels/ThreeDimensionalViz/SceneBuilder";
 import { UrdfSettings } from "@foxglove/studio-base/panels/ThreeDimensionalViz/TopicSettingsEditor/UrdfSettingsEditor";
 import {
-  CoordinateFrame,
+  IImmutableCoordinateFrame,
+  IImmutableTransformTree,
   Transform,
-  TransformTree,
 } from "@foxglove/studio-base/panels/ThreeDimensionalViz/transforms";
 import {
   Color,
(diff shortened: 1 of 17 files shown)


packages/studio-base/src/panels/ThreeDimensionalViz/UrdfBuilder.ts
```diff
@@ -164,29 +163,29 @@ export default class UrdfBuilder implements MarkerProvider {
...
+        const transform = new Transform(translation, rotation);
+        const transformLink: TransformLink = {
+          parent: joint.parent,
+          child: joint.child,
+          transform,
+        };
+        return transformLink;
+      });
+
+      log.debug("Transforms from from urdf: ", transforms);
```
jhurliman, line 181: from from

packages/studio-base/src/panels/ThreeDimensionalViz/UrdfBuilder.ts
```diff
@@ -164,29 +163,31 @@ export default class UrdfBuilder implements MarkerProvider {
     try {
       log.debug(`Parsing ${text.length} byte URDF`);
       this._urdf = await parseRobot(text, fileFetcher);
-      this._needsUpdate = true;
+
+      const transforms = Array.from(this._urdf.joints.values()).map((joint) => {
```
jtbandes, line 167: Array.from takes a 2nd argument that is the map function, so you don't need a separate .map()

packages/studio-base/src/panels/ThreeDimensionalViz/index.tsx
```diff
@@ -337,6 +348,18 @@ function BaseRenderer(props: Props): JSX.Element {
     [autoSyncCameraState, saveCameraStateDebounced, updatePanelConfigs],
   );
 
+  const urdfBuilder = useMemo(() => new UrdfBuilder(), []);
+  useLayoutEffect(() => {
+    const handle = (newTransforms: TransformLink[]) => {
+      setUrdfTransforms((existing) => [...existing, ...newTransforms]);
```
jtbandes, line 354: wonder whether .concat() is faster than two spreads? I haven't tested.

jtbandes (approved):

PR #2603, opened 2022-01-07
Make reselecting the same panel type a no-op.

### LINKED ISSUE #2182: Prevent users from changing their current panel to the same panel type
Let's also prevent users from changing their current panel to the same panel type.

_Originally posted by @amacneil in https://github.com/foxglove/studio/issues/2133#issuecomment-973444397_

foxymiles:
**User-Facing Changes**
This makes reselecting the current panel type a no op instead of resetting the panel.

**Description**
This simply checks if the selected panel type is the same as the current panel and closes the menu.

Fixes #2182

### DIFF
--- a/packages/studio-base/src/components/PanelToolbar/index.tsx
+++ b/packages/studio-base/src/components/PanelToolbar/index.tsx
@@ -172,6 +172,12 @@ function PanelActionsDropdown({
   const swap = useCallback(
     (id?: string) =>
       ({ type, config, relatedConfigs }: PanelSelection) => {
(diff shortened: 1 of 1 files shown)


defunctzombie (approved):

jtbandes:
Could we instead make the dropdown item disabled so you can't click on it at all?

foxymiles:
> Could we instead make the dropdown item disabled so you can't click on it at all?

Would we grey it out in that case? Might be a bit weird visually to be greyed out but also have the highlight checkbox.

PR #2604, opened 2022-01-07
Render small points as rectangles in image view for better performance.

foxymiles:
**User-Facing Changes**
This improves render performance on large numbers of small `point` image markers.

**Description**
This renders small point image markers as rectangles instead of circles to improve performance on large point sets in the image panel.

### DIFF
--- a/packages/studio-base/src/panels/ImageView/renderImage.ts
+++ b/packages/studio-base/src/panels/ImageView/renderImage.ts
@@ -268,7 +268,12 @@ function render({
   ctx.scale(bitmap.width / originalWidth, bitmap.height / originalHeight);
 
   try {
-    paintMarkers(ctx, markers as MessageEvent<ImageMarker | ImageMarkerArray>[], cameraModel);
+    paintMarkers(
+      ctx,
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/panels/ImageView/renderImage.ts
```diff
@@ -397,7 +404,12 @@ function paintMarker(
             : marker.outline_color.a > 0
             ? marker.outline_color
             : marker.fill_color;
-        paintCircle(ctx, point, marker.scale, marker.scale, undefined, fillColor, cameraModel);
+        const size = marker.scale * panZoom.scale;
+        if (size <= 3) {
```
defunctzombie, line 408: Please add a comment on the reasoning to help us understand in the future why we made this performance enhancement. I see there is a comment for paintFastPoint 👍 might be nice to have one on why size <= 3.

defunctzombie:
👍 awesome

packages/studio-base/src/panels/ImageView/renderImage.ts
foxymiles: Done

PR #2608, opened 2022-01-07
Demonstrate a bug in arrow rendering when using two points to define the arrow

jhurliman:
**User-facing changes**

None

**Description**

This updates the Marker story to use a non-empty frame_id for all markers, confirming that frame_id is properly respected for every marker in its default configuration. A new story specific for arrows is also added, since arrows can be defined with either pose+scale or a points array of length 2. In the latter configuration, the pose field is ignored entirely which causes us to ignore any mutations applied by being in a different frame_id.

Screen Shot 2022-01-07 at 1 12 54 PM


defunctzombie:
It is somewhat confusing that this story is merged cause I would now assume it represents the correct state. Maybe nicer for the rest of us if it merged with a fix.

jhurliman:
@defunctzombie we have several 3D panel stories that currently demonstrate broken behavior. They are linked to GH issues and tracked here.

Issue #2476, opened 2021-12-22
Incorrect / mutated data is shown when clicking on markers in the 3D panel


jhurliman:
Clicking on a marker in the 3D panel brings up an overlay with a tab named "Selected object" that shows a view similar to the Raw Message panel for the selected marker. However, instead of showing the original raw marker message, we display the mutated "renderable" marker that is sent to Worldview. This is a confusing thing to see, since the header.frame_id might be `map` but if we're rendering in `base_link` the pose field will be relative to `base_link` and not `map`, and dynamically updating instead of showing the original message contents.

It looks like there is some leftover concept of `originalMessage` in various parts of the code, but it's not wired up to anything that I can see.

<img width="312" alt="Screen Shot 2021-12-22 at 10 08 49 AM" src="https://user-images.githubusercontent.com/195374/147136615-1efbff53-9363-47d0-ad95-b4f6faba56c0.png">

defunctzombie:
@jhurliman Are fields like "settings" also not meant to be present? What's a good way to identify that the pose (or other fields that should be present) are correct? Their values should match the same marker value from the Raw Message panel?

jhurliman:
@defunctzombie as a user, I would expect clicking on a marker to show the message that created this marker. Or maybe a subset of the message that created this marker, like a single Marker in a MarkerArray. Right now, it shows an object that is internal to Studio and not documented anywhere, with fields that don't exist on any message like "settings" and the value of pose mutating every frame (or a pose field injected even when the initial message didn't have a pose field).

PR #2617, opened 2022-01-08
Remove internal studio fields from 3d panel selected object details popup 

defunctzombie:
**User-Facing Changes**
Selected object data in the 3d panel shows information about the selected object or marker free of any internal implementation fields (i.e. settings, hitmap, etc).

Previously these Studio implementation specific fields would pollute the selected object display resulting in noise and confusion with the user's dataset.

| Before | After |
| --- | --- |
| image |image|

**Description**
Fixes: #2476


defunctzombie:
I am still working through the stories to make sure things are working as expected but wanted to open this up for some initial review and feedback.

After discussion with @jtbandes we've decided it is appropriate to remove the "instance object vs full object" dropdown. "Full object" is an internal implementation detail that does not tie into anything the user knows about.

PR #2619, opened 2022-01-08
Replace "stream" and "buffer" with "eventemitter3" and Uint8Array in FetchReader

jhurliman:
**User-facing changes**

Possibly faster downloads (needs testing)

### DIFF
--- a/packages/studio-base/src/randomAccessDataProviders/util.ts
+++ b/packages/studio-base/src/randomAccessDataProviders/util.ts
@@ -82,7 +82,7 @@ export const debounceReduce = <A, T>({
   };
 };
 
-const getLogThroughput = (extensionPoint: ExtensionPoint): ((arg0: Uint8Array) => void) => {
+const getLogThroughput = (extensionPoint: ExtensionPoint): ((buffer: Uint8Array) => void) => {
   return debounceReduce({
     action: (bytes) => extensionPoint.reportMetadataCallback({ type: "received_bytes", bytes }),
     wait: 10,
--- a/packages/studio-base/src/randomAccessDataProviders/util.ts
+++ b/packages/studio-base/src/randomAccessDataProviders/util.ts
@@ -94,7 +94,7 @@ const getLogThroughput = (extensionPoint: ExtensionPoint): ((buffer: Uint8Array)
 export const getReportMetadataForChunk = (
   extensionPoint: ExtensionPoint,
   stallThresholdMs: number = STALL_THRESHOLD_MS,
-): ((arg0: Uint8Array) => void) => {
+): ((buffer: Uint8Array) => void) => {
   const maybeLogStall = getMaybeLogStall(extensionPoint, stallThresholdMs);
   const logThroughput = getLogThroughput(extensionPoint);
   return (buffer: Uint8Array) => {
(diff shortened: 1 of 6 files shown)


defunctzombie:
👍 
I think you can also remove `readable-stream` from our package.json files (at least the studio-base one). I think this was the only user in our code.

jhurliman:
@defunctzombie we still reference `readable-stream` in our webpack config, which is why I assumed it needed to stay. Without that line in the webpack config, @foxglove/xmlrpc and some cipher-base library fail to build.

packages/studio-base/src/randomAccessDataProviders/util.ts
```diff
@@ -23,12 +23,12 @@ const STALL_THRESHOLD_MS = 2000;
 const getMaybeLogStall = (
   extensionPoint: ExtensionPoint,
   stallThresholdMs: number,
-): ((arg0: Buffer) => void) => {
+): ((arg0: Uint8Array) => void) => {
```
defunctzombie, line 26:
```suggestion
): ((buffer: Uint8Array) => void) => {
```

packages/studio-base/src/randomAccessDataProviders/util.ts
```diff
@@ -82,7 +82,7 @@ export const debounceReduce = <A, T>({
   };
 };
 
-const getLogThroughput = (extensionPoint: ExtensionPoint): ((arg0: Buffer) => void) => {
+const getLogThroughput = (extensionPoint: ExtensionPoint): ((arg0: Uint8Array) => void) => {
```
defunctzombie, line 85:
```suggestion
const getLogThroughput = (extensionPoint: ExtensionPoint): ((buffer: Uint8Array) => void) => {
```

packages/studio-base/src/randomAccessDataProviders/util.ts
```diff
@@ -94,10 +94,10 @@ const getLogThroughput = (extensionPoint: ExtensionPoint): ((arg0: Buffer) => vo
 export const getReportMetadataForChunk = (
   extensionPoint: ExtensionPoint,
   stallThresholdMs: number = STALL_THRESHOLD_MS,
-): ((arg0: Buffer) => void) => {
+): ((arg0: Uint8Array) => void) => {
```
defunctzombie, line 97:
```suggestion
): ((buffer: Uint8Array) => void) => {
```

packages/studio-base/src/util/FetchReader.ts
```diff
@@ -73,16 +70,14 @@ export default class FetchReader extends Readable {
...
-      });
+      this.emit("error", new Error(`GET <${this._url}> succeeded, but failed to stream`));
       return undefined;
     }
 
     return this._reader;
   }
 
-  override _read(): void {
+  read(): void {
```
defunctzombie, line 80: Maybe `.startStream()` ? To me `.read()` implies "perform the read".

Other ideas:
We could have read return the actual event emitter instance and the user attaches events to that.
We could invoke .read() automatically in the constructor (I think "streams" do that? Or maybe they wait for someone to register a "data" hander).

I dunno - not critical just something that came to mind while reading the new api.

defunctzombie (approved):
👍 I tested the preview and our sample data loads. Doesn't seem to have any speed impact but I like that we have Uint8Array and have dropped the stream stuff since we don't really need all the plumbing that provides.

packages/studio-base/src/util/FetchReader.ts
jhurliman: > To me .read() implies "perform the read".

Isn't that what this does? The purpose of this method is to call `getReader()` followed by `reader.read()`.

I often get bitten by classes that start performing work in their constructor, which is one of the problems with node streams API (IMO) that the WHATWG APIs don't have. Some form of `start*()` naming seems fine too.

Issue #2481, opened 2021-12-22
"Data sources" sidebar, Duration should show nanoseconds when nanosecond-level timestamps are enabled


jhurliman:
Studio allows toggling between "human-friendly" time and "whole_seconds.fractional_seconds" formatting. When the latter is selected, the "Duration" field of the Data sources sidebar should show full nanosecond precision instead of rounding to the nearest millisecond.

<img width="277" alt="Screen Shot 2021-12-22 at 11 07 14 AM" src="https://user-images.githubusercontent.com/195374/147142868-7b734139-5b2c-49ea-acb4-10a5df87f23b.png">

jtbandes:
I personally don't think it particularly makes sense to tie the time format directly to the seconds' precision. They could be two separate settings or we could just always show nanosecond precision.

The screenshot you showed is also weird because it shows "date > time", but the time is an absolute time rather than being relative to the date...

jtbandes:
Removing bug label – this issue doesn't really render the app broken.

PR #2628, opened 2022-01-08
Render DataSourceInfo Duration using app time format

defunctzombie:
**User-Facing Changes**
Duration values in the Data Source Sidebar and Data Source Info panel conform to the time format application setting. For _seconds_ the value is displayed as decimal seconds. For _time-of-day_ the value is displayed as h:mm:ss.sssssssss.

Screen Shot 2022-01-08 at 7 23 11 AM
Screen Shot 2022-01-08 at 7 23 24 AM
Screen Shot 2022-01-08 at 7 23 32 AM

**Description**
Rendering the duration on the DataSourceInfo sidebar or panel uses the global app time format. If the time format is seconds, then the duration is displayed as decimal seconds. If the time format is time-of-day, then the duration is displayed as h:mm:ss.sssssssss.

Fixes: #2481

### DIFF
--- a/packages/studio-base/src/components/ConnectionList/DataSourceInfo.tsx
+++ b/packages/studio-base/src/components/ConnectionList/DataSourceInfo.tsx
@@ -5,14 +5,13 @@
 import { ITextStyles, Stack, Text, useTheme } from "@fluentui/react";
 import { useMemo } from "react";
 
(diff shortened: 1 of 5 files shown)


2metres (approved):
Looks good

PR #2632, opened 2022-01-10
Add setting for controlling automatic update behavior for desktop

defunctzombie:
**User-Facing Changes**
A user can uncheck "Automatically install updates" to prevent the desktop app from automatically updating itself.
Screen Shot 2022-01-11 at 8 15 22 AM

**Description**
By default the application checks for updates periodically. The user can override this behavior to disable any automatic updates by unchecking this setting.

The default behavior delays checking for updates for a few minutes to give the user time to change the setting for first time app launch.

This setting only applies to the desktop app and is not available in the web version.

Fixes: #2546

### DIFF
--- a/desktop/main/StudioAppUpdater.ts
+++ b/desktop/main/StudioAppUpdater.ts
@@ -6,12 +6,17 @@ import { captureException } from "@sentry/electron";
 import { autoUpdater } from "electron-updater";
 
 import Logger from "@foxglove/log";
+import { AppSetting } from "@foxglove/studio-base/src/AppSetting";
+
+import { getAppSetting } from "./settings";
 
 const log = Logger.getLogger(__filename);
 
-type AppUpdateMode = "none" | "default" | "start";
+function isNetworkError(err: unknown) {
+  if (!(err instanceof Error)) {
+    return false;
+  }
 
-function isNetworkError(err: Error) {
   return (
     err.message === "net::ERR_INTERNET_DISCONNECTED" ||
     err.message === "net::ERR_PROXY_CONNECTION_FAILED" ||
@@ -23,8 +28,7 @@ function isNetworkError(err: Error) {
 }
 
 class StudioAppUpdater {
-  private updateMode: AppUpdateMode | undefined;
-
+  private started: boolean = false;
   // Seconds to wait after app startup to check and download updates.
   // This gives the user time to disable app updates for new installations
   private initialUpdateDelaySec = 60 * 10;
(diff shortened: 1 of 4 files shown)


defunctzombie:
Replaces https://github.com/foxglove/studio/pull/2547

packages/studio-base/src/components/Preferences.tsx
```diff
@@ -245,6 +246,41 @@ function MessageFramerate(): React.ReactElement {
...
+    { key: "none", text: "none - never check for updates" },
+  ];
+
+  if (!isDesktopApp()) {
+    return <></>;
+  }
+
+  return (
+    <Dropdown
+      label="App update mode"
```
jtbandes, line 266:
```suggestion
      label="Check for updates"
```

packages/studio-base/src/components/Preferences.tsx
```diff
@@ -245,6 +246,41 @@ function MessageFramerate(): React.ReactElement {
...
 
+function AutoUpdate(): React.ReactElement {
+  const [appUpdateMode, setAppUpdateMode] = useAppConfigurationValue<string>(
+    AppSetting.APP_UPDATE_MODE,
+  );
+
+  const entries: Array<IDropdownOption> = [
+    { key: "default", text: "default - periodically check for updates" },
+    { key: "start", text: "start - check for updates on startup" },
+    { key: "none", text: "none - never check for updates" },
```
jtbandes, line 257:
```suggestion
    { key: "default", text: "Periodically" },
    { key: "start", text: "On startup" },
    { key: "none", text: "Never" },
```

packages/studio-base/src/components/Preferences.tsx
defunctzombie: Not convinced. I looked at vscode and it calls this setting the update mode.

_Check for updates_ sounds like an action rather than a setting.

desktop/main/StudioAppUpdater.ts
```diff
@@ -0,0 +1,90 @@
...
+      throw new Error("Cannot start StudioAppUpdater again");
+    }
+
+    this.updateMode = updateMode;
+    log.info(`Starting automatic updates with mode: ${this.updateMode}`);
+
+    if (this.updateMode === "start") {
+      this.checkForUpdatesAndNotify();
+    } else {
+      setTimeout(() => {
```
jtbandes, line 53: If we are giving the user time to turn off updates, how does it work if we never cancel this timeout?

packages/studio-base/src/components/Preferences.tsx
defunctzombie: Do you not think we need more explanations on these items rather than one word?

desktop/main/StudioAppUpdater.ts
defunctzombie: We need a note above the app setting (in the app setting sidebar) that an app restart is required to change this setting. We read the app settings once on startup currently.

packages/studio-base/src/components/Preferences.tsx
jtbandes: Assuming the "Check for updates" change above was also included, this doesn't provide any less explanation than your version

packages/studio-base/src/components/Preferences.tsx
jtbandes: It's a dropdown, don't think the user is going to think they are doing some kind of action when they choose an item from the dropdown...

jtbandes:
What's the benefit of providing both periodic and startup-only options here?

defunctzombie:
> What's the benefit of providing both periodic and startup-only options here?

I dunno - could remove it. I copied it from vscode update mode settings.

defunctzombie:
@jtbandes I've re-worked the logic to support changing this setting without an app restart. I've also changed the UI to a checkbox per our discussion. I've left a label above the checkbox because the checkbox alone felt out of place.

jtbandes (approved):

PR #2633, opened 2022-01-10
Fix duplicate .bag item in Open Dialog's "Open local file" option

esthersweon:
**User-Facing Changes**
- Fix duplicate `.bag` item in Open Dialog's "Open local file" option

Screen Shot 2022-01-10 at 1 19 25 PM

### DIFF
--- a/packages/studio-base/src/components/OpenDialog/Start.tsx
+++ b/packages/studio-base/src/components/OpenDialog/Start.tsx
@@ -73,12 +73,17 @@ export default function Start(props: IStartProps): JSX.Element {
     [theme],
   );
 
+  const supportedLocalFiles = useMemo(
+    () => Array.from(new Set(supportedFileExtensions)).join(", "),
+    [supportedFileExtensions],
+  );
+
   const startItems: IButtonProps[] = useMemo(
     () => [
       {
         id: "open-local-file",
         children: "Open local file",
-        secondaryText: `Supports ${supportedFileExtensions.join(", ")} files`,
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/components/OpenDialog/Start.tsx
```diff
@@ -78,7 +78,9 @@ export default function Start(props: IStartProps): JSX.Element {
       {
         id: "open-local-file",
         children: "Open local file",
-        secondaryText: `Supports ${supportedFileExtensions.join(", ")} files`,
+        secondaryText: `Supports ${supportedFileExtensions
```
defunctzombie, line 81: `Array.from(new Set(supportedFileExtensions)).join(", ")` I believe will handle making a unique list for you

defunctzombie: Would be nice to hoist this up into the memo or wherever we make the list so we only make it once not on every render.

packages/studio-base/src/components/OpenDialog/Start.tsx
```diff
@@ -73,12 +73,17 @@ export default function Start(props: IStartProps): JSX.Element {
     [theme],
   );
 
+  const supportedLocalFiles = useMemo(
+    () => Array.from(new Set(supportedFileExtensions)).join(", "),
```
2metres, line 77: Nit: You can go even more compact using the spread operator

```suggestion
    () => […new Set(supportedFileExtensions)].join(", "),
```

PR #2634, opened 2022-01-10
Support image rotation and mirroring in the image panel.

### LINKED ISSUE #2523: Support custom image rotation in ImageView panel
### Discussed in https://github.com/foxglove/studio/discussions/2469

<div type='discussions-op-text'>

<sup>Originally posted by **marijapili** December 22, 2021</sup>
Is it possible to implement a feature for image rotation - to adjust the view if the camera is, for example, upside down?</div>

foxymiles:
**User-Facing Changes**
This implements rotation and both horizontal and vertical mirroring in the image panel.

**Description**
This exposes settings for rotation and mirroring in the image view settings panel. The implementation is straightforward. We just pass the additional parameters to the image worker that does the rendering. The only tricky consideration is that the calculation for the viewbox and zoom fit & fill modes needs to take rotation into account.

This also slightly refactors & tightens up the typing in the image rendering code.

Screen Shot 2022-01-10 at 12 47 44 PM

Fixes #2523

### DIFF
--- a/packages/studio-base/src/panels/ImageView/HitmapRenderContext.ts
+++ b/packages/studio-base/src/panels/ImageView/HitmapRenderContext.ts
@@ -163,6 +163,12 @@ export class HitmapRenderContext {
     this._hctx?.restore();
   }
 
+  rotate(angle: number): void {
+    const rads = (angle * Math.PI) / 180;
+    this._ctx.rotate(rads);
+    this._hctx?.rotate(rads);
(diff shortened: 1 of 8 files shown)


2metres:
Settings make sense to me, should we default to `0º`?

foxymiles:
> Settings make sense to me, should we default to `0º`?

Good idea.

PR #2635, opened 2022-01-10
Widen columns in data source info and allow resize

### LINKED ISSUE #2538: Datatype is cutoff in Data Source Info panel
**Description**
Datatype is cutoff in Data Source Info panel even tho there is plenty of room to display it.

Screen Shot 2021-12-25 at 11 57 28 AM

**Steps To Reproduce**
Open any bag file and view the data source info panel.

2metres:
**User-Facing Changes**
Widen columns in Data Source Info panel and allow resize

**Description**
Closes #2538

### DIFF
--- a/packages/studio-base/src/panels/SourceInfo/index.tsx
+++ b/packages/studio-base/src/panels/SourceInfo/index.tsx
@@ -115,7 +115,13 @@ function SourceInfo() {
           disableSelectionZone
           enableUpdateAnimations={false}
           items={detailListItems}
-          styles={{ root: { ".ms-DetailsHeader": { paddingTop: 0 } } }}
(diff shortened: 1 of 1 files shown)


defunctzombie:
👍 
Is it possible to have the columns automatically size like tables do to their content? So if the topic column doesn't need the space it can go to the datatype column.

2metres:
@defunctzombie might be possible the [demo in the docs](https://developer.microsoft.com/en-us/fluentui#/controls/web/detailslist) shows the ability to sort as well, which might be nice.

defunctzombie:
Meh - the fluent stuff is not long for this world so we can revisit when we port to mui.

PR #2640, opened 2022-01-11
Change cog icon to menu icon in panel toolbar

defunctzombie:
**User-Facing Changes**

Change the panel toolbar cog icon to a menu icon.

| Before | After |
| --- | --- |
|Screen Shot 2022-01-10 at 8 53 55 PM|Screen Shot 2022-01-10 at 8 54 04 PM|

**Description**
The menu icon is more appropriate for this action because it opens a context menu. This better aligns the icon with the action.

### DIFF
--- a/packages/studio-base/src/components/PanelLayout.stories.tsx
+++ b/packages/studio-base/src/components/PanelLayout.stories.tsx
@@ -95,7 +95,7 @@ export const PanelNotFound = (): JSX.Element => {
       <PanelSetup
         onMount={() => {
           setTimeout(() => {
-            (document.querySelectorAll("[data-test=panel-settings]")[0] as any).click();
+            (document.querySelectorAll("[data-test=panel-menu]")[0] as any).click();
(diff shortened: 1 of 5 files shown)


jhurliman (approved):

2metres:
Semantically I agree that the dropdown is a menu but I feel like the cog icon when next to the drag handle and the full screen icon is less ambiguous than the three lines icon.

defunctzombie:
> Semantically I agree that the dropdown is a menu but I feel like the cog icon when next to the drag handle and the full screen icon is less ambiguous than the three lines icon.

@2metres Do you have a recommendation for icons that would make more sense? I dislike the cog serving the purpose of a context menu.

defunctzombie:
Closing this per @2metres recommendation that the menu icon is too similar to the drag handle and other icons in the toolbar.

defunctzombie:
I've changed the tooltip name here: https://github.com/foxglove/studio/pull/2665

PR #2641, opened 2022-01-11
Fix state chart y axis zoom interaction.

### LINKED ISSUE #2623: State Transitions chart layout breaks when zooming/panning
When zooming/panning in the State Transitions plot, the series move to an incorrect position.

https://user-images.githubusercontent.com/14237/148629237-95c2ff16-b8c9-4d1b-a67f-2b02ce0d73cc.mov

foxymiles:
**User-Facing Changes**
This fixes an issue with zooming in the state transition chart.

**Description**
Looks like the logic here was incorrect for calculating y axis values when the user is zooming the chart. If the user had interacted with the view the y min/max were being set to `undefined`.

Fixes #2623

### DIFF
--- a/packages/studio-base/src/components/TimeBasedChart/index.tsx
+++ b/packages/studio-base/src/components/TimeBasedChart/index.tsx
@@ -533,22 +533,15 @@ export default function TimeBasedChart(props: Props): JSX.Element {
       padding: 0,
     };
 
-    let minY;
-    let maxY;
+    let { min: minY, max: maxY } = yAxes;
 
-    if (!hasUserPannedOrZoomed) {
-      // we prefer user specified bounds over dataset bounds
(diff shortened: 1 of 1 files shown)


defunctzombie:
The original intent there was to set to undefined to allow chartjs to manage the y bounds. It seems this is no longer working well (maybe it never was).

Does this break multiple plot pan/zoom sync? I don't think it should be something to confirm.

👍

foxymiles:
> The original intent there was to set to undefined to allow chartjs to manage the y bounds. It seems this is no longer working well (maybe it never was).
> 
> Does this break multiple plot pan/zoom sync? I don't think it should be something to confirm.
> 
> 👍

Good question @defunctzombie . As far as I can tell synced plots still work with this patch.

PR #2643, opened 2022-01-11
Fix panel toolbar hover to reveal.

### LINKED ISSUE #2377: All panel headers appear when hovering over any panel in a tab
**Description**

Create a top-level Tab panel. Put two Image panels inside of a single tab. Hover over one of the Image panels, and observe the Image panel header appearing on both Image panels.

foxymiles:
**User-Facing Changes**
This improves the hover interaction to hide & show toolbars in tab panels.

**Description**
This supplies a new `usePanelMousePresence` hook that tracks the presence of the mouse within the parent panel that can be used by child components to control their visibility.

Fixes #2377 

https://user-images.githubusercontent.com/93935560/149392145-6027c167-142f-4bb6-ab91-9a6aeea77c00.mov

### DIFF
--- a/packages/studio-base/src/components/PanelRoot.tsx
+++ b/packages/studio-base/src/components/PanelRoot.tsx
@@ -1,0 +1,38 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import styled from "styled-components";
+
+import { colors, spacing } from "@foxglove/studio-base/util/sharedStyleConstants";
+
+// This is in a separate file to prevent circular import issues.
+export const PanelRoot = styled.div<{ fullscreen: boolean; selected: boolean }>`
+  display: flex;
+  flex-direction: column;
+  flex: 1 1 auto;
+  overflow: hidden;
+  z-index: ${({ fullscreen }) => (fullscreen ? 10000 : 1)};
+  background-color: ${({ theme }) => (theme.isInverted ? colors.DARK : colors.LIGHT)};
+  position: ${({ fullscreen }) => (fullscreen ? "fixed" : "relative")};
+  border: ${({ fullscreen }) => (fullscreen ? "4px solid rgba(110, 81, 238, 0.3)" : "none")};
+  top: 0;
+  left: 0;
+  right: 0;
+  bottom: ${({ fullscreen }) => (fullscreen ? spacing.PLAYBACK_CONTROL_HEIGHT : 0)};
+
+  :after {
+    content: "";
+    top: 0;
+    left: 0;
+    right: 0;
+    bottom: 0;
+    opacity: ${({ selected }) => (selected ? 1 : 0)};
+    border: 1px solid ${colors.ACCENT};
+    position: absolute;
+    pointer-events: none;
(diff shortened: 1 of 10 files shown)


packages/studio-base/src/components/PanelRoot.tsx
```diff
@@ -0,0 +1,38 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import styled from "styled-components";
+
+import { colors, spacing } from "@foxglove/studio-base/util/sharedStyleConstants";
+
+// This is in a separate file to prevent circular import issues.
```
defunctzombie, line 9: Circular import issues with what?

packages/studio-base/src/hooks/usePanelHoverVisibility.tsx
```diff
@@ -0,0 +1,50 @@
...
+  ref: MutableRefObject<HTMLElement | ReactNull>,
+  options?: { forceVisible: boolean },
+): void {
+  const listener = useCallback(
+    (e: MouseEvent) => {
+      if (!ref.current) {
+        return;
+      }
+
+      if (e.type === "mouseenter") {
```
defunctzombie, line 24:
```suggestion
      if (ev.type === "mouseenter") {
```

packages/studio-base/src/hooks/usePanelHoverVisibility.tsx
```diff
@@ -0,0 +1,50 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { MutableRefObject, useCallback, useEffect } from "react";
+
+import { PanelRoot } from "@foxglove/studio-base/components/PanelRoot";
+
+/**
+ * Attaches a listener to the parent panel mouse enter & mouse leave events.
+ *
+ * @param ref The element to hide and show on panel hover.
```
defunctzombie, line 12: Should this return the visibility and allow the caller to show/hide their element themselves? So this hook exists to manage the enter/leave events.

packages/studio-base/src/hooks/usePanelHoverVisibility.tsx
```diff
@@ -0,0 +1,50 @@
...
+ */
+export function usePanelHoverVisibility(
```
defunctzombie, line 14:
```suggestion
export function usePanelMousePresence(
```

If you go with the approach of it returning a state vs changing the element directly

packages/studio-base/src/components/PanelRoot.tsx
foxymiles: For example the toolbar imports the hook which imports `PanelRoot` which imports the toolbar.

packages/studio-base/src/hooks/usePanelHoverVisibility.tsx
foxymiles: Good idea. I'll change it.

packages/studio-base/src/hooks/usePanelHoverVisibility.tsx
foxymiles: Yes good idea. I've reworked it to just return a boolean.

Issue #741, opened 2021-04-28
Enable React Strict Mode for all panels


jtbandes:
Explanation from #627:
> Read more about strict mode at https://reactjs.org/docs/strict-mode.html
> 
> > StrictMode is a tool for highlighting potential problems in an application. …  It activates additional checks and warnings for its descendants.
> 
> > Strict mode checks are run in development mode only; they do not impact the production build.
> 
> The most notable behavior of strict mode is that it causes component render functions, useMemo callbacks, and other kinds of functions that should be stateless to be called **twice per render** in dev mode. This may cause problems in components that use these in side-effectful ways, so I didn't want to immediately enable it by default in all panels. I added a warning icon in the panel toolbar to prompt us to enable it in existing panels incrementally:
> 
> <img width="402" alt="image" src="https://user-images.githubusercontent.com/14237/115669426-f49ae580-a2e3-11eb-9655-84f1889c4e79.png">

The panels below all have `supportsStrictMode` set to `false`. To complete one of these tasks, first remove the false value, then manually test the panel features to ensure strict mode doesn't cause any regressions. For an example of how to recognize and fix problems caused by strict mode, see https://github.com/foxglove/studio/pull/627#discussion_r618131836.

- [x] Internals
- [x] Diagnostic Detail
- [x] Diagnostic Summary
- [ ] Global Variables
- [x] Global Variable Slider
- [x] Image View
- [x] Internal Logs
- [x] Map
- [x] Node Playground
- [x] Parameters
- [x] Playback Performance
- [x] Plot
- [x] Publish
- [x] Raw Messages
- [x] Source Info
- [x] State Transitions
- [x] Tab
- [x] Table
- [x] ThreeDimensionalViz
- [x] Welcome

Finally, once all panels support strict mode:
- [ ] Delete the `supportsStrictMode` setting and always render panels in StrictMode.

jtbandes:
State Transitions and Plot blocked on: https://github.com/facebook/react/issues/23082

amacneil:
🎉

PR #2649, opened 2022-01-11
Enable strict mode for Welcome and Tab panels, remove supportsStrictMode boolean

jtbandes:
**User-Facing Changes**
None

**Description**
Closes https://github.com/foxglove/studio/issues/741

### DIFF
--- a/packages/studio-base/src/components/Panel.tsx
+++ b/packages/studio-base/src/components/Panel.tsx
@@ -230,7 +230,6 @@ type Props<Config> = {
 export interface PanelStatics<Config> {
   panelType: string;
   defaultConfig: Config;
-  supportsStrictMode?: boolean;
   configSchema?: PanelConfigSchema<Config>;
 }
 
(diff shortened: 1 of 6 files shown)


packages/studio-base/src/components/Panel.tsx
```diff
@@ -699,11 +697,7 @@ export default function Panel<
               </ActionsOverlay>
             )}
             <ErrorBoundary renderError={(errorProps) => <ErrorToolbar {...errorProps} />}>
-              {PanelComponent.supportsStrictMode ?? true ? (
-                <React.StrictMode>{child}</React.StrictMode>
-              ) : (
-                child
-              )}
+              <React.StrictMode>{child}</React.StrictMode>
```
amacneil, line 700: dumb question - can we just make the whole app one big `<React.StrictMode>` now?

jtbandes: We should work towards that but I haven’t tested all the sidebar and playback bar stuff in strict mode yet.

Issue #2651, opened 2022-01-11
Remote file does not support HTTP Range Requests when I got a file from url


yh1215853664:
**Description**
I'm using docker On my Server.
When I point the "Open File from URL",I got an error,Remote file does not support HTTP Range Requests
But I have confired that the server support range request.
The follow is the result on my client

PS C:\Users\Administrator> curl -i http://file.shui-cn.com/1.bag -H "range:bytes=1024-2000"
HTTP/1.1 206 Partial Content
Server: nginx/1.16.1
Date: Tue, 11 Jan 2022 20:58:35 GMT
Content-Type: application/octet-stream
Content-Length: 977
Last-Modified: Tue, 11 Jan 2022 20:44:03 GMT
Connection: keep-alive
ETag: "61ddec13-7b5bae1"
Accepr-Ranges: bytes
Access-Control-Allow-Origin: *
Access-Control-Allow-Methods: GET, POST, OPTIONS
Access-Control-Allow-Headers: DNT,X-Mx-ReqToken,Keep-Alive,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type,Authorization
Content-Range: bytes 1024-2000/129350369

And The button of Open local file is not avaliable when I using docker on my server
I will get the error from the brower console 

OpenDialog.tsx:52 
        
       TypeError: window.showOpenFilePicker is not a function
    at Uo (showOpenFilePicker.tsx:13:25)
    at useOpenFile.tsx:28:32
    at OpenDialog.tsx:50:7
    at Zu (react-dom.production.min.js:219:180)
    at oc (react-dom.production.min.js:259:160)
    at S.unstable_runWithPriority (scheduler.production.min.js:18:343)
    at eo (react-dom.production.min.js:122:325)
    at io (react-dom.production.min.js:252:279)
    at Es (react-dom.production.min.js:243:371)
    at react-dom.production.min.js:123:115
Thanks a lot

- Version:.026.0
- OS:Centos7

**Steps To Reproduce**

**Expected Behavior**

**Actual Behavior**

yh1215853664:
I try it both on Microsoft Edge and Google Chrome.The Version is 97.0.1072.55(Edge) and 96.0.4664.93(Chrome)

jtbandes:
`Accepr-Ranges: bytes`

Is this a typo or is the server really returning an `Accepr-Ranges: bytes` header? It should be `Accept-Ranges: bytes`.

[...]

PR #2653, opened 2022-01-12
Improve error message for failed remote bag requests

defunctzombie:
**User-Facing Changes**
When failing to load a remote url from a server, the user is shown an error message with more details on what might be misconfigured.

**Description**
In the web app, remote bag requests may fail due due to CORS errors. This change improves the error message to call out which parts of the CORS configuration should be verified to ensure compatability with studio.

Fixes: #2651

### DIFF
--- a/packages/studio-base/src/util/BrowserHttpReader.ts
+++ b/packages/studio-base/src/util/BrowserHttpReader.ts
@@ -13,6 +13,7 @@
 
 import { FileReader, FileStream } from "@foxglove/studio-base/util/CachedFilelike";
 import FetchReader from "@foxglove/studio-base/util/FetchReader";
+import isDesktopApp from "@foxglove/studio-base/util/isDesktopApp";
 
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/util/BrowserHttpReader.ts
```diff
@@ -35,15 +36,30 @@ export default class BrowserHttpReader implements FileReader {
       response = await fetch(this._url, { signal: controller.signal });
       controller.abort();
     } catch (error) {
-      throw new Error(`Fetching remote file failed. <${this._url}> ${error}`);
+      let errMsg = `Fetching remote file failed. ${error}`;
+
+      if (!isDesktopApp()) {
+        errMsg +=
+          "Sometimes this is due to a CORS configuration error on the server. Make sure CORS is enabled.";
```
jhurliman, line 43: Does this need a leading newline or space?

defunctzombie: yea it does - thanks!

Issue #2652, opened 2022-01-11
Text in "Log" panel is not infinitely scrollable


sherry-chef:
**Description**
Text in "Log" panel is not infinitely scrollable any more. Specifically, if the log content is longer than the Log panel size, the content before what can be contained in the Log panel will disappear when I attempted to scroll up. And then when I scrolled down the content outside of the screen sometimes disappears as well.

- Version: 0.24.0
- OS: Ubutn 18.04

**Steps To Reproduce**

Use any bag file with a super long rosout. Drag and drop it into Foxglove Studio. Let it play such that rosout content is longer than the Log panel. Then scroll up.
![image](https://user-images.githubusercontent.com/86276953/149029005-d3efab46-98e0-4a43-9e13-fd69b2b45bb6.png)

**Expected Behavior**
Infinite scroll - the content beyond the screen should still exist.

**Actual Behavior**
The content beyond the screen disappears.

defunctzombie:
Closing because we've decided to revert the fluent-ui removal until we can prepare a migration that does not undo this bugfix.

PR #2654, opened 2022-01-12
Switch to @fluentui/react list component for Log list

defunctzombie:
**User-Facing Changes**
Correctly show previous entries when scrolling back in the log panel.

**Description**
The @fluentui/list component handles virtualized lists so we don't need to maintain a separate virtualized list component.

This fixes a bug in the Log panel which prevented scrollback from displaying previous log items.

Fixes: #2652


defunctzombie:
There is a weird quirk with the fluent list that causes the list items to "jitter" as we append new items:

https://user-images.githubusercontent.com/84792/149152830-cea2b71c-d3bc-4f1d-a9c8-708f9a2b4283.mov

During manual scrolling this is not an issue but does break the smoothness of automatic scrolling for new items. If we think this is bad enough to avoid using the @fluentui/list component then we need to go back to the home grown implementation.

packages/studio-base/src/panels/Log/filterMessages.ts
```diff
@@ -26,6 +26,8 @@ export default function filterMessages(
     return messages;
   }
 
+  const searchTermsInLowerCase = searchTerms.map((term) => term.toLowerCase());
```
defunctzombie, line 29: This was needlessly inside each invocation of the filter function so I moved it out. It is unrelated to the bugfix.

packages/studio-base/src/panels/Log/index.tsx
```diff
@@ -72,8 +74,15 @@ const LogPanel = React.memo(({ config, saveConfig }: Props) => {
 
   // avoid making new sets for node names
   // the filter bar uess the node names during on-demand filtering
-  const seenNodeNames = useRef(new Set<string>());
-  messages.forEach((msg) => seenNodeNames.current.add(msg.message.name));
+  const seenNodeNamesCache = useRef(new Set<string>());
+
+  const seenNodeNames = useMemo(() => {
+    for (const msg of messages) {
+      seenNodeNamesCache.current.add(msg.message.name);
```
defunctzombie, line 81: @jtbandes have I used memo wrong here? :D

packages/studio-base/src/panels/Log/index.tsx
```diff
@@ -92,27 +103,86 @@ const LogPanel = React.memo(({ config, saveConfig }: Props) => {
...
-            </div>
-          )}
-        />
+      <Stack grow style={{ overflow: "hidden" }}>
+        <div ref={divRef} style={{ height: "100%", overflow: "scroll" }}>
+          {/* items property wants a mutable array but filteredMessages is readonly */}
+          <List
+            componentRef={listRef}
+            items={filteredMessages as ArrayElementType<typeof filteredMessages>[]}
+            onPagesUpdated={() => {
```
defunctzombie, line 162: tragically the fluent list component does not have any "autoscroll" behavior - so that is what I've implemented here with the scrollToBottom(), the scroll event listener, and the view reset button

jtbandes:
> There is a weird quirk with the fluent list that causes the list items to "jitter" as we append new items

This kinda looks like a font/positioning issue to me – like maybe the lines are being positioned at non-integer offsets and the renderer is rounding them?

defunctzombie:
> This kinda looks like a font/positioning issue to me – like maybe the lines are being positioned at non-integer offsets and the renderer is rounding them?

Maybe - don't know if I can do anything about it tho. Or are you thinking there is something to do?

packages/studio-base/src/panels/Log/index.tsx
jtbandes: Yes.

Issue #2647, opened 2022-01-11
Map panel previewTime indicator doesn't work while paused


jtbandes:
**Steps To Reproduce**
- Add a map panel
- Load a preloadable data source with GPS data
- Hover the playback bar to show the hovered time as a yellow dot

**Expected Behavior**
Hovered time is shown as a yellow dot whether playing or paused.

**Actual Behavior**
Yellow dot only appears when playing.

https://user-images.githubusercontent.com/14237/148999469-7cd04077-8b5a-4e84-94d0-de608e4c1a0a.mov

PR #2656, opened 2022-01-12
Invoke render for extension panels when the hover value changes.

defunctzombie:
**User-Facing Changes**
The map panel shows the current user hover location when a user hovers over the playback scrubber even when playback is paused.

**Description**
This change fixes a bug with extension panel rendering. Extension panels would not receive a new previewTime if the playback was paused since no logic invoked a panel render when paused. This change fixes the logic to invoke a panel render when the hover value changes as well as when message pipeline changes.

Fixes: #2647

### DIFF
--- a/packages/studio-base/src/components/PanelExtensionAdapter.tsx
+++ b/packages/studio-base/src/components/PanelExtensionAdapter.tsx
@@ -131,7 +131,11 @@ function PanelExtensionAdapter(props: PanelExtensionAdapterProps): JSX.Element {
 
   const colorScheme = useTheme().isInverted ? "dark" : "light";
 
-  const renderPanel = useCallback(() => {
+  // renderPanelImpl invokes the panel extension context's render function with updated
+  // render state fields.
+  //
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/components/PanelExtensionAdapter.tsx
```diff
@@ -256,18 +260,30 @@ function PanelExtensionAdapter(props: PanelExtensionAdapterProps): JSX.Element {
...
+      queueRender();
     },
-    [renderFn, renderPanel],
+    [queueRender],
   );
 
+  // Queue render on hover value changes which occur outside the message pipeline
+  useLayoutEffect(() => {
+    // No need to queue render if not interested in preview time
+    if (watchedFieldsRef.current.has("previewTime")) {
```
jtbandes, line 282: I guess technically this might render when queueRender changes, not just when hoverValue changes, but that's probably not a problem

jtbandes (approved):

PR #2659, opened 2022-01-12
Increase top margin of image panel topic selector icon.

### LINKED ISSUE #2658: Padding issues with image topic selector
**Description**
Topic selector is overlapping the buttons

image

foxymiles:
**User-Facing Changes**
This fixes an issue with margins on the new image view inspector icon.

**Description**
This just increases the top margin of the icon. It's hard to do a more intelligent fix since the toolbar wraps and hence has unpredictable height.

Screen Shot 2022-01-12 at 1 05 12 PM

Fixes #2658


amacneil:
👍

PR #2661, opened 2022-01-12
Enable Strict Mode for whole app, with ability to disable in settings

jtbandes:
**User-Facing Changes**
None

**Description**
- Wrap the whole app (including `Root`) in `<StrictMode>`.
  - Add an Experimental Feature setting to disable strict mode. This setting is only visible during development, since strict mode doesn't do anything in a production build. Strict mode will be **on** by default.
    image

  - Refactor web and desktop entry points to create the `LocalStorageAppConfiguration` and `NativeStorageAppConfiguration` outside of the Root component, so we can read the enableStrictMode setting before rendering Root.

- Patch `react-dom` so that console.log is not suppressed during double-renders in strict mode. This behavior was confusing and is being reverted in React 18, but we have not upgraded yet. https://github.com/facebook/react/issues/21783

- Change UserNodePlayer so that it doesn't call `setListener` on the underlying player in its constructor, but waits until setListener is called on it. The setListener function is a "side effect" that was happening during construction of the UserNodePlayer in a useMemo, and for remote bags this would trigger double preflight HTTP requests in strict mode.


packages/studio-base/src/AppSetting.ts
```diff
@@ -20,4 +20,5 @@ export enum AppSetting {
   OPEN_DIALOG = "ui.open-dialog",
   UPDATES_ENABLED = "updates.enabled",
   SHOW_OPEN_DIALOG_ON_STARTUP = "ui.open-dialog-startup",
+  ENABLE_REACT_STRICT_MODE = "enableReactStrictMode",
```
defunctzombie, line 23: `development.` prefix maybe to scope dev time only settings?

defunctzombie:
👍

packages/studio-base/src/AppSetting.ts
jtbandes: we have no consistency whatsoever among these different settings 😭

defunctzombie: not with that attitude we don't

jtbandes: at least I picked one of the existing styles rather than creating a new style

Issue #2644, opened 2022-01-11
ROS_HOSTNAME setting ignored when creating ROS 1 player


jtbandes:
When params are added, they show up in the Parameters panel. However, when an existing param is updated, the new value does not appear.

- Version: 0.26.0
- OS: macOS 12.1 host running Studio, Ubuntu 20.04 VM running ROS 1 noetic

**Steps To Reproduce**
- Run `roscore`
- Connect a ROS 1 data source in Studio
- Run `rosparam set /foo 'bar'` to add a param. The new param shows up in the Parameters panel.
- Run `rosparam set /foo 'baz'` to change the value of the existing param. The new value is **not** displayed in the Parameters panel.

Debugging note: it seems like the `paramUpdate` xmlrpc method is not being called?

It seems like the ROS_HOSTNAME text field is being ignored, and `localhost` is used as the hostname in the Ros1Player.

<img width="938" alt="image" src="https://user-images.githubusercontent.com/14237/149186412-25fc84c4-3189-48ab-9d8e-5b0c0010e301.png">

PR #2662, opened 2022-01-12
Fix parameter updating

defunctzombie:
**User-Facing Changes**
ROS1 Native connection users will receive parameter updates when connected to a remote machine (assuming their ROS_HOSTNAME is correct).

**Description**

This change fixes two issues. The first is with the Ros1DataSourceFactory. The factory was initializing the Ros1Player using the wrong argument name. The argument name is _hostname_ which matches the form field _id_ configured within the same factory.

The second change fixes JSONInput. JSONInput displays json to the user and allows them to edit the json. To accomplish this, the component has internal state that tracks the latest value of the value prop. It also had logic to update the state if the value was different from the prop. This set state was called in render producing a side effect.

Instead of setting the state in render and tracking the previous prop value with a ref, we track the _editing_ state and set the internal state value only when the user is not editing the component.

Fixes: #2644

### DIFF
--- a/packages/studio-base/src/components/input/JSONInput.tsx
+++ b/packages/studio-base/src/components/input/JSONInput.tsx
@@ -11,24 +11,42 @@
 //   found at http://www.apache.org/licenses/LICENSE-2.0
 //   You may not use this file except in compliance with the License.
 import { useTheme } from "@fluentui/react";
-import { useRef, useState } from "react";
+import { useEffect, useRef, useState } from "react";
 
 import { LegacyInput } from "@foxglove/studio-base/components/LegacyStyledComponents";
 
 const keyValMap: Record<string, number> = { ArrowDown: -1, ArrowUp: 1 };
 
+/**
+ * JSONInput displays a json value and allows the user to edit the value.
+ */
 export function JSONInput(props: {
   value: string;
   dataTest?: string;
-  onChange: (arg0: unknown) => void;
+  onChange: (newValue: unknown) => void;
 }): React.ReactElement {
   const theme = useTheme();
+
+  // The JSONInput is semi-controlled.
+  // We need to avoid updating the input text when the user is actively editing the input.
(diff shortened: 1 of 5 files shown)


packages/studio-base/src/components/input/JSONInput.tsx
```diff
@@ -37,7 +55,19 @@ export function JSONInput(props: {
...
+      }}
+      onBlur={() => {
+        editingRef.current = false;
+        // when we are done editing we might have an updated value to set
+        if (lastPropValueRef.current != undefined) {
+          setInternalValue(lastPropValueRef.current);
+        }
+      }}
       onChange={(e) => {
+        // we've updated out own value and don't want to use any intermediate value
```
jtbandes, line 69:
```suggestion
        // we've updated our own value and don't want to use any intermediate value
```

packages/studio-base/src/components/input/JSONInput.tsx
```diff
@@ -11,24 +11,42 @@
...
   const [internalValue, setInternalValue] = useState<string>(props.value);
-  const lastPropsValue = useRef<string>(props.value);
-  if (lastPropsValue.current !== props.value) {
-    lastPropsValue.current = props.value;
+  const editingRef = useRef(false);
+
+  // keep track of the last prop value we've seen so if we leave edit mode with no changes
+  // we can update the input field to the value
+  const lastPropValueRef = useRef<string | undefined>(props.value);
+  lastPropValueRef.current = props.value;
```
jtbandes, line 40: still not sure this is kosher, might want to move it into the useEffect

jtbandes (approved):

packages/studio-base/src/components/input/JSONInput.tsx
defunctzombie: The reason this is not ok is because the props.value here might never actually be committed?

Otherwise this seeks ok since the intent here is to capture the last value we've seen for props across any render call.

PR #2663, opened 2022-01-13
Add `displayName` to all React Contexts

jtbandes:
**User-Facing Changes**
None

**Description**
Improves component tree readability in the react dev tools.

before | after
--- | ---
Screen Shot 2022-01-12 at 4 03 20 PM | Screen Shot 2022-01-12 at 4 02 26 PM


defunctzombie:
😮

PR #2667, opened 2022-01-13
Remove fullscreen entry from panel toolbar context menu

defunctzombie:
**User-Facing Changes**
The fullscreen item is removed from the panel toolbar context menu.

**Description**
The panel toolbar already has a fullscreen icon for quick access. Having this in the context menu is redundant.

### DIFF
--- a/packages/studio-base/src/components/PanelToolbar/PanelActionsDropdown.tsx
+++ b/packages/studio-base/src/components/PanelToolbar/PanelActionsDropdown.tsx
@@ -159,18 +159,6 @@ export function PanelActionsDropdown({ isOpen, setIsOpen, isUnknownPanel }: Prop
     if (!isUnknownPanel) {
       items.push(
         {
-          key: "fullscreen",
-          text: "Fullscreen",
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/components/PanelToolbar/PanelActionsDropdown.tsx
```diff
@@ -158,18 +158,6 @@ export function PanelActionsDropdown({ isOpen, setIsOpen, isUnknownPanel }: Prop
     ];
     if (!isUnknownPanel) {
       items.push(
-        {
-          key: "fullscreen",
-          text: "Fullscreen",
-          onClick: () => {
-            panelContext?.enterFullscreen();
-          },
-          "data-test": "panel-menu-fullscreen",
```
jtbandes, line 167: FYI this will break some screenshot tests

defunctzombie: fixed

PR #2668, opened 2022-01-13
Fix app menu option to open a Rosbridge connection

esthersweon:
**User-Facing Changes**
- Fix "File > Open Connection > Rosbridge (ROS 1 & 2)" app menu option to open a Rosbridge connection correctly 

Screen Shot 2022-01-13 at 1 41 40 PM

### DIFF
--- a/desktop/main/StudioWindow.ts
+++ b/desktop/main/StudioWindow.ts
@@ -418,19 +418,14 @@ class StudioWindow {
       return;
     }
 
-    // Electron menus require a preceding & to escape the & char
-    const formattedName = name
-      .split("")
-      .map((char) => (char === "&" ? "&&" : char))
-      .join("");
-    this._inputSources.add(formattedName);
+    this._inputSources.add(name);
 
     const fileMenu = this._menu.getMenuItemById("fileMenu");
     if (!fileMenu) {
       return;
     }
 
-    const existingItem = fileMenu.submenu?.getMenuItemById(formattedName);
+    const existingItem = fileMenu.submenu?.getMenuItemById(name);
     // If the item already exists, we can silently return
     // The existing click handler will support the new item since they have the same name
     if (existingItem) {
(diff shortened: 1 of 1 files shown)


desktop/main/StudioWindow.ts
```diff
@@ -513,11 +513,15 @@ class StudioWindow {
     fileMenu.submenu?.append(
       new MenuItem({
         label: "Open Connection",
-        submenu: Array.from(this._inputSources).map((sourceName) => ({
-          label: sourceName,
+        submenu: Array.from(this._inputSources).map((formattedName) => ({
```
jtbandes, line 516: It would be nicer to iterate over the original names, and only "format" them right when they're being put into the `label` param. Then you can send the original unformatted name in to the .send() below.

Also if there is a simpler implementation like `.replace("&&, "&")`, would prefer that over split+map+join.

desktop/main/StudioWindow.ts
```diff
@@ -418,19 +418,14 @@ class StudioWindow {
...
-    this._inputSources.add(formattedName);
+    this._inputSources.add(name);
 
     const fileMenu = this._menu.getMenuItemById("fileMenu");
     if (!fileMenu) {
       return;
     }
 
-    const existingItem = fileMenu.submenu?.getMenuItemById(formattedName);
+    const existingItem = fileMenu.submenu?.getMenuItemById(name);
```
jtbandes, line 428: Hm, is this right?  Doesn't look like we are assigning an `id` to the MenuItems at all? cc @defunctzombie I think you wrote this code originally

jtbandes (approved):
Looks much nicer! Just a question about getMenuItemById

desktop/main/StudioWindow.ts
esthersweon: Weird. It seems like from the [source code](https://github.com/electron/electron/blob/8d0ed05c99a992cfb241329906e7fb20a001d228/lib/browser/api/menu.ts#L108) that it truly is looking for an `id` field.

@defunctzombie - did you maybe remove the `id` field at some point, and give `label` the intended `id` value?

defunctzombie: Possibly - tho it doesn't ring a bell for me that I did it explicitly. Things have certainly worked this way unchanged for a while? Or maybe not? I'd say tweak what needs to be done to make it work if it isn't since I can't recall anything about this _id_ field stuff.

defunctzombie:
Whats the status of this PR? Does it need additional review or are things working?

defunctzombie:
@esthersweon is this ready to merge?

desktop/main/StudioWindow.ts
esthersweon: The bug was fixed with my changes so far; this PR doesn't change any other functionality. Will merge now / investigate this separately.

PR #2672, opened 2022-01-14
Split up key handlers to stabilize handlers during user interactions.

### LINKED ISSUE #2666: Backticks are horribly broken
**Description**
1. Hold backtick to enter quick edit mode
2. Mouse over some panels
3. Let go of backtick. We don't let you out.
4. If you click around some panels it's possible to get into an even more broken state where a panel thinks it is permanently being moused over.

https://user-images.githubusercontent.com/637671/149413173-fcaa9eb7-e448-4f7d-a3fe-d48e1fc9e4fd.mov

foxymiles:
**User-Facing Changes**
This fixes an issue with the hotkey overlay in tab panels.

**Description**
The issue was that key handlers were being uninstalled and reinstalled during a user interaction and this prevented the panels from receiving coherent state updates. The solution here is to split out the frequently changing key handlers into a separate set of handlers.

Fixes #2666

### DIFF
--- a/packages/studio-base/src/components/Panel.tsx
+++ b/packages/studio-base/src/components/Panel.tsx
@@ -486,6 +486,8 @@ export default function Panel<
       [cmdKeyPressed, parentPanelContext],
     );
 
+    // We use two separate sets of key handlers because the panel context and exitFullScreen
+    // change often and invalidate our key handlers during user interactions.
     const { keyUpHandlers, keyDownHandlers } = useMemo(
       () => ({
         keyUpHandlers: {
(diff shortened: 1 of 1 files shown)


defunctzombie (approved):
👍 cool - I'd say we could even remove the useMemo and useCallback for each individual handler. The useMemo here seems unnecessarily trying to memo more things together than need to be.

foxymiles:
> 👍 cool - I'd say we could even remove the useMemo and useCallback for each individual handler. The useMemo here seems unnecessarily trying to memo more things together than need to be.

The main thing is that the keyhandler needs to not rerender during user interactions so I think the memo is helping us here in that respect. Maybe bundling all the keyhandlers into a single component makes it harder to keep it stable than we'd like it to be?

PR #2676, opened 2022-01-14
Wrap datasource timestamps in narrow views.

### LINKED ISSUE #2507: Data source start/end time are truncated without ellipsis
When the data source sidebar is narrow, the data source name has an ellipsis in the middle, but the start/end time are just cut off.

image

foxymiles:
**User-Facing Changes**
This improves rendering of datasource timestamps in narrow views.

**Description**
The approach here is to just wrap the text.

Screen Shot 2022-01-14 at 11 35 25 AM

Fixes #2507

### DIFF
--- a/packages/studio-base/src/components/ConnectionList/DataSourceInfo.tsx
+++ b/packages/studio-base/src/components/ConnectionList/DataSourceInfo.tsx
@@ -67,7 +67,7 @@ function DataSourceInfo(): JSX.Element {
       <Stack tokens={{ childrenGap: theme.spacing.s2 }}>
         <Text styles={subheaderStyles}>Start time</Text>
         {startTime ? (
-          <Timestamp horizontal time={startTime} />
+          <Timestamp horizontal time={startTime} wrap />
         ) : (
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/components/ConnectionList/DataSourceInfo.tsx
```diff
@@ -78,7 +78,7 @@ function DataSourceInfo(): JSX.Element {
       <Stack tokens={{ childrenGap: theme.spacing.s2 }}>
         <Text styles={subheaderStyles}>End time</Text>
         {endTime ? (
-          <Timestamp horizontal time={endTime} />
+          <Timestamp horizontal time={endTime} wrap />
```
2metres, line 81: Not sure if this needs to be a prop,  could just be the default behavior

2metres (changes requested):

packages/studio-base/src/components/ConnectionList/DataSourceInfo.tsx
foxymiles: I guess we only use this in one place now so you're right we don't really need it as a param.

PR #2680, opened 2022-01-14
Allow extensions to read and monitor application settings

jhurliman:
**User-facing changes**

None

**Description**

Adds `subscribeAppSettings(settings: string[])` to the extension API and an `appSettings?: ReadonlyMap<string, AppSettingValue>;` field to the extension `RenderState`.

### DIFF
--- a/packages/studio-base/src/components/PanelExtensionAdapter.tsx
+++ b/packages/studio-base/src/components/PanelExtensionAdapter.tsx
@@ -3,12 +3,21 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
 import { useTheme } from "@fluentui/react";
-import { CSSProperties, useCallback, useLayoutEffect, useMemo, useRef, useState } from "react";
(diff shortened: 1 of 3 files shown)


defunctzombie:
Note for the future. There **are** user facing changes here. The API is a user facing aspect of our app and should receive appropriate call-out and review.

packages/studio-base/src/components/PanelExtensionAdapter.tsx
```diff
@@ -407,6 +452,10 @@ function PanelExtensionAdapter(props: PanelExtensionAdapterProps): JSX.Element {
         subscribedTopicsRef.current.clear();
         setSubscriptions(panelId, []);
       },
+
+      subscribeAppSettings: (settings: string[]) => {
```
defunctzombie, line 456: We could send all app settings to all extensions. Its not changing often and doesn't matter if an extension gets all the settings?

Issue #2379, opened 2021-12-14
Diagnostic Detail panel dropdown hides all other options when one is selected


jhurliman:
The Diagnostic Detail panel has a dropdown that allows showing the key/values + OK/ERROR status for all diagnostics or only a single diagnostic. The "all diagnostics" option is just a blank dropdown entry which looks more like a bug than a purposeful choice, and the behavior of the dropdown is different after selecting the blank option vs one of the other options. Selecting a specific option and then opening the dropdown again, it looks like it's using the entry name as filter text so no other options appear.

<img width="475" alt="Screen Shot 2021-12-14 at 3 00 20 PM" src="https://user-images.githubusercontent.com/195374/146093253-c7e3908e-d9c2-4a6d-a32d-b0971e9dd9b3.png">
<img width="471" alt="Screen Shot 2021-12-14 at 3 00 32 PM" src="https://user-images.githubusercontent.com/195374/146093265-c0bf3650-eb04-4d88-bcff-3e2fce330e71.png">

jtbandes:
This seems to depend which kind of item you click. If you click an item that represents a whole hardware_id, the problem doesn't occur. If you select a hardware_id+name combo, it does. For example with this data (from demo.bag):

<img width="269" alt="image" src="https://user-images.githubusercontent.com/14237/146110990-142c18aa-2596-4b01-a8cb-d28b78093054.png">

If I choose the `piksi rtk` item, I don't experience the problem, but I do with the other two more specific piksi options.

amacneil:
Need to discuss desired behavior.

defunctzombie:
I re-read the original issue description and I do not understand how to reproduce. @jhurliman could you indicate what steps you took to get into the state where you see blank entries? I agree what I see in the screenshot with blank dropdown entries looks undesirable. What I don't yet understand is what the desired behavior is or the steps to reproduce.

jhurliman:
Here's a video demonstrating the problem. You can reproduce this by playing around with the dropdown in the fourth tab of the nuScenes layout.

https://user-images.githubusercontent.com/195374/149594924-db6f7497-b531-4f41-af6f-01d3c6245099.mov

Desired behavior:
 - Don't use `""` (empty strings) as a dropdown entry. It looks accidental and doesn't clue the user as to what will happen if you select it
 - Don't hide all of the other dropdown items when one is already selected

PR #2684, opened 2022-01-15
Improve usability for Diagnostic Detail panel

defunctzombie:
**User-Facing Changes**
The Autocomplete widget is replaced with a different autocomplete widget. The overall experience is similar to before (you can type to filter the list). The new widget resolves some UX strangeness from the previous widget described in more detail below.

We've also removed the code paths which strip away a leading "/" from the hardware id as well as the code paths which remove the hardware_id string from the _name_ field. Both of these looked like webviz "cruisisms" and don't seem to be something that should be done for all ROS and non-ros users.

**Description**
The autocomplete in the Diagnostic Detail panel would hide available options once the user selected one. To change to a different diagnostic the user has to click the input, realize they can delete their text, delete all the text and type again to find their entry.

This change switches the autocomplete input to the mui Autocomplete component. This component exhibits better UX behavior. When focusing on the input, the user is shown all available entries with their current entry highlighted. This also resolves some glitches with the dropdown.

This change also modifies the autocomplete options logic to ignore empty hardware ids with no name from showing in the autocomplete options list. These entries would create blanks in the dropdown list.

Fixes: #2379

### DIFF
--- a/packages/studio-base/src/theme/muiComponents.ts
+++ b/packages/studio-base/src/theme/muiComponents.ts
@@ -109,7 +109,7 @@ export default function muiComponents(theme: Theme): ThemeOptions["components"] 
     },
     MuiPaper: {
       defaultProps: {
-        elevation: 0,
+        elevation: 2,
         square: true,
       },
     },
--- a/packages/studio-base/src/panels/diagnostics/DiagnosticStatusPanel.tsx
+++ b/packages/studio-base/src/panels/diagnostics/DiagnosticStatusPanel.tsx
@@ -11,11 +11,11 @@
 //   found at http://www.apache.org/licenses/LICENSE-2.0
 //   You may not use this file except in compliance with the License.
 
+import { Autocomplete, TextField } from "@mui/material";
 import { sortBy } from "lodash";
-import { useCallback, useMemo } from "react";
+import { useMemo } from "react";
 
 import { useDataSourceInfo } from "@foxglove/studio-base/PanelAPI";
-import Autocomplete, { IAutocomplete } from "@foxglove/studio-base/components/Autocomplete";
 import EmptyState from "@foxglove/studio-base/components/EmptyState";
 import Flex from "@foxglove/studio-base/components/Flex";
 import Panel from "@foxglove/studio-base/components/Panel";
@@ -26,8 +26,9 @@ import { DIAGNOSTIC_TOPIC } from "@foxglove/studio-base/util/globalConstants";
 
 import DiagnosticStatus from "./DiagnosticStatus";
 import helpContent from "./DiagnosticStatusPanel.help.md";
-import useDiagnostics, { DiagnosticAutocompleteEntry } from "./useDiagnostics";
-import { DiagnosticInfo, getDisplayName, trimHardwareId } from "./util";
+import useAvailableDiagnostics from "./useAvailableDiagnostics";
+import useDiagnostics from "./useDiagnostics";
+import { getDisplayName } from "./util";
 
 export type Config = {
   selectedHardwareId?: string;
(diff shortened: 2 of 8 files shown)


defunctzombie:
@jhurliman can you give this PR a spin and see how it feels to use especially for the issues raised in #2379 . Once we are happy with the interactions we can get feedback from others on our changes.

jhurliman:
@defunctzombie yes, this is better. Ship it

jhurliman (changes requested):
I think the stories `"DiagnosticStatusPanel:Selected Hardware ID Only"`, `"DiagnosticStatusPanel:Selected Name"`, and `"DiagnosticStatusPanel:Moved Divider"` should be updated. It sets the `name` field to values such as `hardware_id1: name1` which we assume is a Cruise-ism, so the output looks off with the redundant hardware ids.

Another thing that appears to have changed is the dropdown is always visible, even when there are zero or one hardware ids. I prefer the previous behavior of hiding the dropdown in those cases (I generally lean toward minimal UI for Studio panels), but it's a good discussion point.

defunctzombie:
> I think the stories `"DiagnosticStatusPanel:Selected Hardware ID Only"`, `"DiagnosticStatusPanel:Selected Name"`, and `"DiagnosticStatusPanel:Moved Divider"` should be updated. It sets the `name` field to values such as `hardware_id1: name1` which we assume is a Cruise-ism, so the output looks off with the redundant hardware ids.

I looked at this and it's actually not because the story sets the name field to `hardware_id1: name1` but because the `<DiagnosticStatus>` component which renders each DiagnosticInfo item renders the displayName field from each item. We could detect that you've filtered down to a specific hardware_id or even a hardware_id and name and pass a prop to the DiagnositcStatus component to avoid showing the heading but its starting to feel like a bit too much scope creep for me on this specific fix.

> Another thing that appears to have changed is the dropdown is always visible, even when there are zero or one hardware ids. I prefer the previous behavior of hiding the dropdown in those cases (I generally lean toward minimal UI for Studio panels), but it's a good discussion point.

What if there are multiple names under a single hardware id? It would be weird UX to me for the panel to automatically select a diagnostic when you open a datasource just because that happened to be the first message seen in the data source. I actually find this behavior for the image panel somewhat annoying and another one of those areas where the app is trying to be too clever instead of making it easy and obvious how to configure the panel with your data.

defunctzombie:
@jhurliman I had accidentally removed a floating from the panel toolbar when I was iterating. I've put that back.

jtbandes:
This text should probably be disabled (grayed out), or just hide the dropdown when waiting for diagnostics because we already say it in the center of the panel:

image

We also need a drop shadow on the menu so it doesn't blend in with other UI elements:
image

compare with the Message Path Input autocomplete which has a shadow:
image

jhurliman (approved):

jtbandes (changes requested):
needs UI cleanup

defunctzombie:
I've updated the PR to disable the dropdown when there are no diagnostics - this prevents "waiting for diagnostics..." from being shown in the empty state.

I've updated the PR to set the default paper elevation to `2` which brings back a dropshadow.

@2metres Thoughts on why the default paper elevation was set to 0?

packages/studio-base/src/theme/muiComponents.ts
```diff
@@ -109,7 +109,7 @@ export default function muiComponents(theme: Theme): ThemeOptions["components"]
     },
     MuiPaper: {
       defaultProps: {
-        elevation: 0,
+        elevation: 2,
```
defunctzombie, line 112: @2metres is this ok to do here?

2metres: Yep looks good to me.

defunctzombie:
@jtbandes take a look and if you are 👍 on the changes please squash and merge.

jtbandes (approved):

PR #2685, opened 2022-01-16
Remove panel name from panel toolbar

defunctzombie:
**User-Facing Changes**
The panel name is no longer displayed in the panel toolbar

Before
Screen Shot 2022-01-15 at 4 40 06 PM

After
Screen Shot 2022-01-15 at 4 35 28 PM

**Description**
The panel name shows up in the panel toolbar when a panel is considered wide enough. Even with this logic, the panel name takes up valuable toolbar space when there are other controls in the toolbar.

Seeing the panel name is not important to visualization and diagnostic of the users dataset so we remove it in favor of more space for other UI elements.

### DIFF
--- a/packages/studio-base/src/components/PanelToolbar/PanelToolbarControls.tsx
+++ b/packages/studio-base/src/components/PanelToolbar/PanelToolbarControls.tsx
@@ -33,7 +33,6 @@ type PanelToolbarControlsProps = {
   // eslint-disable-next-line @foxglove/no-boolean-parameters
   setMenuOpen: (_: boolean) => void;
   showControls?: boolean;
-  showPanelName?: boolean;
 };
 
 const useStyles = makeStyles({
@@ -50,11 +49,6 @@ const useStyles = makeStyles({
     svg: {
       fontSize: 14,
     },
-  },
-  panelName: {
-    fontSize: 10,
-    opacity: 0.5,
-    marginRight: 4,
   },
   icon: {
     fontSize: 14,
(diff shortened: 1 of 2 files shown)


jtbandes:
Without the name in the toolbar, what’s the next easiest way to see what panel it is?

defunctzombie:
> Without the name in the toolbar, what’s the next easiest way to see what panel it is?

Good question. For some panels opening the panel settings. We need to discuss how important this is and maybe add a way to find the name. One idea is to show the name in the context menu. I am sure we can come up with other good ideas.

2metres:
Looks good to me, particualarly where the toolbar is like the one in the screenshot and has interactions in the same space as the panel toolbar. 

As a user of the app is it important to know the name of the panel?

jtbandes:
As a beginner user exploring layouts and learning how to create them, it seems useful to know what a panel is without having to scroll through the whole list of available panels and look at all the thumbnails.

defunctzombie:
I think the panel name provides more interference than value it provides. I'm willing to put the name somewhere but it should not be in the toolbar taking up visual UI space.

I'm willing to make the change in this PR to put the name elsewhere if a recommendation is given but beyond that I don't think we should hold up removing the name.

jtbandes:
Ok, then my recommendation is to put it in the menu. Maybe as a menu "header".

defunctzombie:
Each panel toolbar has a help icon. Clicking the help icon opens the sidebar with help for the panel. The heading on the help sidebar is the name of the panel.

I think thats a good mechanism for discovering panel names.

jtbandes:
👍

Issue #2646, opened 2022-01-11
Autocomplete in Node Playground `Input<>` throws an error


jtbandes:
**Description**
This is a regression compared with webviz.io/app.

**Steps To Reproduce**
- Add a new node in Node Playground
- Place cursor inside `Input<|>`, type quotes `Input<"|">`, and press ctrl+space to trigger autocomplete.

**Expected Behavior**
<img width="332" alt="image" src="https://user-images.githubusercontent.com/14237/148998831-4ad6d7a2-996f-4d1d-be71-f1b1bad599cd.png">

**Actual Behavior**
<img width="1110" alt="image" src="https://user-images.githubusercontent.com/14237/148998823-5eaa8bae-059e-49b7-badb-9704ab00dedd.png">

defunctzombie:
It seems that the generated roslib injection is causing monaco editor to receive a new roslib:

https://github.com/foxglove/studio/blob/main/packages/studio-base/src/panels/NodePlayground/Editor.tsx#L84-L89

I replaced the above code with:

```typescript
React.useEffect(() => {
    if (rosLib.includes("Placeholder")) {
      return;
    }
    const disposable = monacoApi.languages.typescript.typescriptDefaults.addExtraLib(
      rosLib,
      `file:///node_modules/@types/${projectConfig.rosLib.fileName}`,
    );

    return () => {
      disposable.dispose();
    };
  }, [rosLib]);
```

And that makes it more obvious that monaco editor (language server?) is not taking into account the new library code.

defunctzombie:
The above extra lib issue seems to be related to https://github.com/microsoft/monaco-editor/issues/2866

And may or may not be the issue described in this bug report. Still investigating.

PR #2687, opened 2022-01-16
Downgrade monaco-editor to 0.30.1

defunctzombie:
**User-Facing Changes**
Node playground editor correctly updates when a new ros library is generated.

**Description**

This fixes a bug with loading the generated rosLib into monaco editor. The monaco editor language server does not pick up the newly generated roslib.

This change adds a check to webpack.ts to alert us on future upgrades to confirm the feature works.

See: https://github.com/microsoft/monaco-editor/issues/2866

Fixes: #2646


packages/studio-base/webpack.ts
```diff
@@ -14,6 +15,21 @@ import webpack, { Configuration, WebpackPluginInstance } from "webpack";
 import { WebpackArgv } from "./WebpackArgv";
 import packageJson from "./package.json";
 
+if (monacoPkg.version !== "0.30.1") {
```
defunctzombie, line 18: I felt that stories were too easy to overlook - but I think a story will also show errors in the editor so we could remove this if we think the story is good enough.

packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/typescript/ros.ts
```diff
@@ -12,7 +12,37 @@
 //   You may not use this file except in compliance with the License.
 export const ros_lib_filename = "ros/index.d.ts";
 export const ros_lib_dts = `
-  // Placeholder module declaration.
-  // Once any bag-like data source is added this module will be autogenerated.
-  declare module "ros";
+  export declare interface Duration {
```
defunctzombie, line 15: This is a nicer experience for users that have not yet loaded a data source. They get the namespaces that will appear with nice comments on why they are not yet populated.

jtbandes (approved):

PR #2690, opened 2022-01-18
Use message path type to drive autocomplete.

### LINKED ISSUE #2669: Selecting some plot series adds an extra period to the message path
**Description**
Adding a plot series that ends with `.x` (and some other suffixes) causes the series to be added with an extra period, creating an invalid message path input that has to be fixed manually.

https://user-images.githubusercontent.com/637671/149415324-291c05c1-0ee7-48b7-909b-da1561cfd4a6.mov

foxymiles:
**User-Facing Changes**
This improves message path autocomplete behavior.

**Description**
The approach here is to calculate what the new path would be when an autocompletion is selected and determine if it would result in a path to a simple value. If so we don't append a `.` to help the user continue searching.

Fixes #2669

### DIFF
--- a/packages/studio-base/src/components/MessagePathSyntax/MessagePathInput.tsx
+++ b/packages/studio-base/src/components/MessagePathSyntax/MessagePathInput.tsx
@@ -22,6 +22,7 @@ import {
 import { flatten, flatMap, partition } from "lodash";
 import { CSSProperties, useCallback, useMemo } from "react";
 
+import { RosMsgField } from "@foxglove/rosmsg";
 import * as PanelAPI from "@foxglove/studio-base/PanelAPI";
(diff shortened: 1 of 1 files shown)


jtbandes (approved):
This solution seems ok. However, it indicates that `autocompleteType === "topicName"` is no longer a good name for this mode if some of the autocomplete entries with `topicName` type are not topic names...

foxymiles:
> This solution seems ok. However, it indicates that `autocompleteType === "topicName"` is no longer a good name for this mode if some of the autocomplete entries with `topicName` type are not topic names...

Yeah it's not really currently a good name either since it just defaults to that if it doesn't match a topic. Feels like this code could use a bit of a refactor at some point.

PR #2691, opened 2022-01-18
Upgrade @foxglove/ulog to fix truncated ULog reading

jhurliman:
**User-facing changes**

Truncated PX4 ULog files can now be read


jtbandes:
Do we display an error to the user if the file was truncated or just silently ignore it?

jhurliman:
@jtbandes we ignore it. The viewer at https://logs.px4.io/ does not treat partial messages as an error, it just discards the truncated message and declares the end of the previous message as the end of the file. I'm guessing unclean shutdowns are common in drone logs.

PR #2703, opened 2022-01-20
Fix plot panning.

### LINKED ISSUE #2701: Unable to pan Plot vertically
https://user-images.githubusercontent.com/14237/150219006-f3f1532f-dfa8-4458-8cb1-aa63d6efe59f.mov

foxymiles:
**User-Facing Changes**
This fixes panning in plots.

**Description**
We had special logic to not reset scales if the user has zoomed the chart but not if the user has panned the chart. So any user initiated panning was immediately overridden. This just adds logic to treat pans the same way we already treat zooms.

Fixes #2701


defunctzombie:
👍

jtbandes:
🎉

PR #2707, opened 2022-01-20
Display panel thumbnails in grid for empty layout

jtbandes:
**User-Facing Changes**

In an empty layout with no panels (or an empty tab), the list of available panels is displayed in a grid.

image


packages/studio-base/src/components/EmptyPanelLayout.tsx
```diff
@@ -77,35 +74,21 @@ export const EmptyDropTarget = ({ tabId }: Props): JSX.Element => {
...
-              lineHeight: "1.5",
-            },
-          }}
-        >
-          Nothing here yet.
-          <br />
-          <ChildToggle position="below" style={{ display: "inline-flex" }}>
-            <Link underline data-test="pick-a-panel">
-              Pick a panel
+        <Stack sx={{ paddingBottom: (theme) => theme.spacing(2) }}>
```
2metres, line 83: When using the `sx` prop for `padding` and `margin` the theme spacing is available without having to access the theme: `paddingBottom: 2`

Issue #2688, opened 2022-01-17
Raw message panel resets what parts of the message is collapsed when bag loops


clementchef:
Small bug in the raw message panel. Let say I want to see all details: I have to click all the drop down manually (which is fine to do once). However as soon as you loop the bag you lose whatever subtree you were seeing. Everything become collapsed again. You have to reopen the message tree manually everytime. Could this be saved somehow? even saved in layout if people always care abou the same subtree?

- Version:  0.25.1
- OS: ubuntu 18.04

**Steps To Reproduce**
Open raw message panel, expend a message subfiled, loop over the bag.

**Expected Behavior**
Panel should not change.

**Actual Behavior**
Panel collapses all fields.

defunctzombie:
@clementchef I'm working on a fix for this - just waiting on some updates to a library we use for the tree view.

As a workaround - I learned that clicking on the field name (rather than the arrow) to expand/collapse will remember the state when the bag loops. The bug comes from different handling of clicking the field name vs clicking arrow.

PR #2710, opened 2022-01-21
Track expand/collapse state for raw message fields

defunctzombie:
**User-Facing Changes**
Expanding a field in the raw message panel by clicking the arrow next to the field name behaves like clicking the label. The field is expanded (if collapsed), and the state of the field expansion is remembered through data source looping.

**Description**
When a user clicks on a field label, the raw message panel would remember the field was expanded. If the datasource re-loaded or looped, the panel would re-expand the previously expanded fields.

This behavior was not applied when clicking the arrow next to the label. This change fixes this behavior by registering onExpand and onCollapse handlers on the Tree. These fire when the user clicks on the arrow next to the label.

Fixes: #2688

### DIFF
--- a/packages/studio-base/package.json
+++ b/packages/studio-base/package.json
@@ -160,7 +160,7 @@
     "react-hover-observer": "2.1.1",
     "react-input-autosize": "3.0.0",
     "react-is": "17.0.2",
-    "react-json-tree": "0.15.1",
+    "react-json-tree": "patch:react-json-tree@npm:0.15.1#../../patches/react-json-tree.patch",
     "react-markdown": "7.1.2",
     "react-monaco-editor": "0.46.0",
     "react-mosaic-component": "5.0.0",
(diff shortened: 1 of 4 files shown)


foxymiles:
@defunctzombie Is there anything more specific I need to repro the original bug? I'm not seeing this on current main if I just load nuscenes and loop.


https://user-images.githubusercontent.com/93935560/152169764-72fc6160-b3b9-46c8-9049-298a910b1f92.mov

defunctzombie:
@foxymiles how did you expand the fields? Did you click on the field name or the arrows? If you expand by clicking the arrow, the fields will collapse on loop.

foxymiles:
> @foxymiles how did you expand the fields? Did you click on the field name or the arrows? If you expand by clicking the arrow, the fields will collapse on loop.

Just clicking on the arrow.


https://user-images.githubusercontent.com/93935560/152187303-ca4b8e09-f3a1-4111-b345-71f0740efc46.mov

defunctzombie:
Interesting that it doesn't reproduce the same way on main now - but I definitely reproduced it via looping (thats how I figured out the state wasn't tracked). I've found a more reliable way to reproduce - which is to seek in the playback controls while playing.

foxymiles (approved):
Logic looks good to me. 👍

PR #2713, opened 2022-01-21
Refactor timestamp method dropdown into a component

2metres:
**User-Facing Changes**
Refactor timestamp method dropdown into self contained component separate from the message path input

Screen Shot 2022-01-21 at 4 25 01 pm

Screen Shot 2022-01-21 at 4 25 06 pm

**Description**
Hoisting and refactoring mostly working except showing the current method


2metres:
Current issues: 
- [ ] Legend row loses focus on menu click
- [ ] checkmark doesn't show up on selected item

defunctzombie:
@2metres I pushed a commit to fix the legend row losing hover on menu click. We might want to make this a prop of the TimestampMethodDropdown if we plan to use it in other places?

I don't understand "checkmark doesn't show up on selected item". Does this refer to the same timestamp method dropdown? I was able to see checkmarks. Is this some other part of the legend?

2metres:
Screen Shot 2022-01-23 at 9 11 49 pm

Issue #2611, opened 2022-01-07
New users are prompted to create a layout before the app is usable


amacneil:
**Description**
New user experience is currently pretty clunky, since we no longer create your first layout, the user is forced to create an empty layout before they are able to select any visualization panels.

https://user-images.githubusercontent.com/637671/148610581-00503845-cfb8-4d23-aeb5-57b00e47601a.mov

Ideas:
- If you don't have any data source or any panels, prompt the user to select a data source
- If you start the app without any layouts, we automatically create your first layout
- Consider designing a generic default layout (e.g. 3d panel, image panel, raw message panel), so that we can guide new users to select some topics/data rather than wasting precious onboarding attention training them about layout management.

amacneil:
Proposed default layout:

[default_layout.json.txt](https://github.com/foxglove/studio/files/7997321/default_layout.json.txt)

PR #2716, opened 2022-01-21
Show panel thumbnails and descriptions in panel sidebar and menu

jtbandes:
**User-Facing Changes**
Panel thumbnails and descriptions are now shown in the "Add panel" sidebar and the "Change panel" menu.

**Description**
image

### DIFF
--- a/packages/studio-base/src/components/PanelList/index.tsx
+++ b/packages/studio-base/src/components/PanelList/index.tsx
@@ -22,7 +22,9 @@ import {
   List,
   ListItem,
   ListItemButton,
+  ListItemIcon,
   ListItemText,
+  Paper,
   Stack,
   Typography,
 } from "@mui/material";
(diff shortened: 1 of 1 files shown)


amacneil:
nit: we lost the padding at the bottom of this sidebar again. not sure why that keeps happening.

amacneil:
@defunctzombie @2metres thoughts?

I don't hate it. I do dislike that now I have to scroll the sidebar, whereas before I could always see the full list. However, it's nice having the descriptions and icons there. I worry that we're maybe too highly prioritizing first time users at the expense of our power users (rather than trying to consider both's needs)? But maybe that is not true.

It is pretty terrible if you have both the updated "add panel" sidebar open, as well as the "empty layout" state:

image

I think I'd be in favor of merging this PR and reverting/rethinking the empty layout state (potentially just a default layout, like we were discussing earlier).

Also I think we should merge `Diagnostic - Summary` and `Diagnostic - Details`, but that's a separate discussion.

jtbandes:
> It is pretty visually jarring if you have both the updated "add panel" sidebar open, as well as the "empty layout" state, in the same window

Probably an uncommon use case? Do you feel like you landed in this state naturally or were just poking around to test out the PR?

amacneil:
Hard to say. From an empty incognito window I opened a bag, clicked new layout, then instinctively clicked the add panel sidebar. I guess I’m not used to picking a panel from the empty layout panel grid yet.

jtbandes:
May revisit this after https://github.com/foxglove/studio/issues/2611, which is a higher priority.

PR #2719, opened 2022-01-24
Implement click to publish points & poses in 3d panel.

### LINKED ISSUE #605: Ability to publish pose to topic by clicking point in 3D panel
Is there any way to publish the pose msg by click the point in 3D panel, just like the "2D Pose Estimate" button in rviz

foxymiles:
**User-Facing Changes**
This adds a feature that allows users to click on the 3d panel to publish various point & pose messages.

**Description**
This slightly the existing state logic in the 3d panel toolbar to consolidate it into a shared reducer and adds new buttons and tools for publishing points & poses. It also adds new 3d panel settings to customize how these messages are formed.

Fixes #605


packages/studio-base/src/panels/ThreeDimensionalViz/MainToolbar.tsx
```diff
@@ -2,111 +2,294 @@
...
-  perspective: boolean;
   debug: boolean;
+  interactionState: InteractionState;
+  interactionStateDispatch: InteractionStateDispatch;
   onToggleCameraMode: () => void;
   onToggleDebug: () => void;
+  perspective: boolean;
+};
+
+const PublishClickIcons: Record<PublishClickType, ReactNode> = {
```
jtbandes, line 47: putting these in a separate file would be nice, don't need to clutter up layout logic with SVG data

packages/studio-base/src/panels/ThreeDimensionalViz/MeasureMarker.tsx
```diff
@@ -11,52 +11,50 @@
 //   found at http://www.apache.org/licenses/LICENSE-2.0
 //   You may not use this file except in compliance with the License.
 
-import { Lines, Spheres, Point } from "@foxglove/regl-worldview";
+import { ReactElement } from "react";
```
jtbandes, line 14: tip: you can use `JSX.Element` without importing anything

packages/studio-base/src/panels/ThreeDimensionalViz/index.tsx
```diff
@@ -407,24 +407,48 @@ const configSchema: PanelConfigSchema<ThreeDimensionalVizConfig> = [
...
 ];
 
 BaseRenderer.displayName = "ThreeDimensionalViz";
 BaseRenderer.panelType = "3D Panel";
 BaseRenderer.defaultConfig = {
+  autoSyncCameraState: false,
+  autoTextBackgroundColor: true,
+  cameraState: {},
   checkedKeys: ["name:Topics"],
+  clickToPublishGoalTopic: "/move_base_simple/goal",
```
jtbandes, line 437: let's rename these settings keys to `PoseEstimate` rather than `Goal`

packages/studio-base/src/panels/ThreeDimensionalViz/index.tsx
```diff
@@ -407,24 +407,48 @@ const configSchema: PanelConfigSchema<ThreeDimensionalVizConfig> = [
...
+  clickToPublishPointTopic: "/clicked_point",
+  clickToPublishPoseTopic: "/initialpose",
+  clickToPublishPoseXDeviation: 0.5,
+  clickToPublishPoseYDeviation: 0.5,
```
jtbandes, line 441: these ones should probably say `PoseEstimate` not just `Pose`

packages/studio-base/src/util/geometry.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export type Quaternion = { x: number; y: number; z: number; w: number };
+
+export type Vector3 = { x: number; y: number; z: number };
+
+export function distanceBetweenPoints(a: Point, b: Point): number {
+  return Math.hypot(b.x - a.x, b.y - a.y, b.z - a.z);
+}
+
+export function eulerToQuaternion(rpy: Vector3): Quaternion {
```
jtbandes, line 17: is this copied from UrdfBuilder? should we remove it from there and import this one?

packages/studio-base/src/util/geometry.ts
```diff
@@ -0,0 +1,57 @@
...
+  covariance[6 * 0 + 0] = Math.pow(x, 2);
+  covariance[6 * 1 + 1] = Math.pow(y, 2);
+  covariance[6 * 5 + 5] = Math.pow(theta, 2);
+  return covariance;
+}
+
+export function quaternionFromPoints(a: Point, b: Point): Quaternion {
+  const theta = Math.atan2(b.y - a.y, b.x - a.x);
+  return eulerToQuaternion({ x: 0, y: 0, z: theta });
+}
```
jtbandes, line 48: not sure this warrants a separate function, I think you can just do it with `quat.rotationTo` or `quat.rotateZ`? The function name `quaternionFromPoints` is also not super clear on how the quaternion is to be created from the points

packages/studio-base/src/util/geometry.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export function reglClickToPoint(click: ReglClickInfo): Point | undefined {
+  const intersection = click.ray.planeIntersection([0, 0, 0], [0, 0, 1]);
+  return intersection ? vec3ToPoint(intersection) : undefined;
+}
+
+export function vec3ToPoint(v: Vec3): Point {
+  return { x: v[0], y: v[1], z: v[2] };
+}
```
jtbandes, line 57: seems like worldview already exports this function?

jtbandes (approved):

packages/studio-base/src/util/geometry.ts
foxymiles: This was just an existing function moved over.

packages/studio-base/src/panels/ThreeDimensionalViz/index.tsx
foxymiles: Good idea.

packages/studio-base/src/util/geometry.ts
foxymiles: Good catch I'll use that one.

packages/studio-base/src/util/geometry.ts
foxymiles: Yes good idea.

PR #2730, opened 2022-01-25
Fix open url parsing logic to handle query params etc.

### LINKED ISSUE #2726: Data Source: open file from url(http/https) does't support searchParams
**Data Source: open file from url(http/https) does't support searchParams**

- Version: 0.27.0
- OS: macos Big Sur

**Steps To Reproduce**

image

**Expected Behavior**
can be opened regardless of whether url has searchParams or not

**Actual Behavior**
it has an error

----------------------------------------------------------------------------------------
I browse the code and I think the question may be in this:
 file location: foxglove/packages/studio-base/src/components/OpenDialog/Remote.tsx

```js
 const extension = path.extname(currentUrl);
```
if the variable currentUrl has searchParams which contains dot(.) will lead to above error.
Whether the pathname should be extracted as the path.extname() params:
```js
  const url = new URL(currentUrl);
  const extension = path.extname(url?.pathname);
```

foxymiles:
**User-Facing Changes**
This fixes an issue in the the open remote location dialog.

**Description**
This uses the `URL` class to parse the provided URL string instead of trying to extract the extension directly.

Fixes #2726

### DIFF
--- a/packages/studio-base/src/components/OpenDialog/Remote.tsx
+++ b/packages/studio-base/src/components/OpenDialog/Remote.tsx
@@ -20,6 +20,14 @@ type RemoteProps = {
   availableSources: IDataSourceFactory[];
 };
 
+function maybeParseURL(urlString: string): undefined | URL {
+  try {
+    return new URL(urlString);
+  } catch {
+    return undefined;
+  }
+}
+
 export default function Remote(props: RemoteProps): JSX.Element {
   const { onCancel, onBack, availableSources } = props;
 
@@ -32,8 +40,9 @@ export default function Remote(props: RemoteProps): JSX.Element {
       return;
     }
 
-    const extension = path.extname(currentUrl);
-    if (extension.length === 0) {
+    const parsedUrl = maybeParseURL(currentUrl);
+    const extension = parsedUrl ? path.extname(parsedUrl.pathname) : undefined;
+    if (extension == undefined || extension.length === 0) {
       setErrorMessage("URL must end with a filename and extension");
       return;
     }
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/components/OpenDialog/Remote.tsx
```diff
@@ -32,8 +40,9 @@ export default function Remote(props: RemoteProps): JSX.Element {
       return;
     }
 
-    const extension = path.extname(currentUrl);
-    if (extension.length === 0) {
+    const parsedUrl = maybeParseURL(currentUrl);
```
defunctzombie, line 43: What about catching the error here and using `setErrorMessage` indicating the url is not valid? I guess that happens somewhat by parsedUrl being _undefined_ and then showing the `URL must end with a filename and extension` error message. It might be nice to separate our parsing the url as an error message to indicate an issue prior to even detecting the extension. Don't feel strongly on it tho.

defunctzombie (approved):

packages/studio-base/src/components/OpenDialog/Remote.tsx
```diff
@@ -32,27 +32,32 @@ export default function Remote(props: RemoteProps): JSX.Element {
...
-        url: currentUrl,
-      },
-    });
+      selectSource(foundSource.id, {
+        type: "connection",
+        params: {
+          url: currentUrl,
+        },
+      });
+    } catch (error) {
```
defunctzombie, line 58: small nit - if `selectSource` throws during execution - the error will be caught as "current url is not a valid url" :D

foxymiles: Good point. I sometimes wish TS had `throws` annotations.

foxymiles: I think this is actually cleaner moving the parsing out and avoiding `try` blocks. But I'll add a check and message for an invalid url.

defunctzombie (approved):
🚢

PR #2731, opened 2022-01-25
Turn on @typescript-eslint/no-unnecessary-condition lint rule.

foxymiles:
**User-Facing Changes**
None

**Description**
This turns on the @typescript-eslint/no-unnecessary-condition  to provide stronger typechecking.


packages/studio-base/src/components/MessagePipeline/index.tsx
```diff
@@ -161,12 +161,8 @@ export function MessagePipelineProvider({
           await pauseFrameForPromises(promises);
 
           currentPlayerTickState.waitingForPromises = false;
-          // https://github.com/microsoft/TypeScript/issues/43781
-          // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
-          if (currentPlayerTickState.resolveFn) {
-            currentPlayerTickState.resolveFn();
-            currentPlayerTickState.resolveFn = undefined;
-          }
```
jtbandes, line 169: This change is wrong. See linked github issue.

packages/studio-base/src/panels/RawMessages/CopyMessageButton.tsx
```diff
@@ -21,7 +21,7 @@ export default function CopyMessageButton({ text, data }: Props): JSX.Element {
     (ev: MouseEvent<HTMLElement>) => {
       ev.stopPropagation();
       ev.preventDefault();
-      void clipboard.copy(JSON.stringify(data, copyMessageReplacer, 2) ?? "");
+      void clipboard.copy(JSON.stringify(data, copyMessageReplacer, 2));
```
jtbandes, line 24: I don't think this should be removed. stringify can return undefined...e.g. `JSON.stringify(undefined)` returns undefined.

packages/studio-base/src/panels/RawMessages/Value.tsx
```diff
@@ -57,7 +57,7 @@ export default function Value({
             fade
             className={classes.icon}
             onClick={() => {
-              void clipboard.copy(JSON.stringify(itemValue, copyMessageReplacer, 2) ?? "");
```
jtbandes, line 60: same here

packages/studio-base/src/util/typeUtils.ts
```diff
@@ -0,0 +1,8 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+/**
+ * Used to mark specific proprties in an object as optional.
+ */
+export type WithOptionals<T, Optionals extends keyof T> = Omit<T, Optionals> & Partial<T>;
```
jtbandes, line 8: Maybe `utility-types`? https://stackoverflow.com/a/60128290/23649

packages/studio-base/src/panels/ThreeDimensionalViz/TopicTree/NamespaceMenu.tsx
```diff
@@ -113,59 +113,57 @@ export default function NamespaceMenu({
...
-                        </Icon>
-                      )}
-                    </SColorPickerWrapper>
-                  </SItemContent>
-                </Item>
-              </>
-            )}
-          </>
-        )}
+        <>
```
jtbandes, line 116: can remove the `<></>` wrapper

packages/studio-base/src/panels/RawMessages/CopyMessageButton.tsx
foxymiles: Yeah you're right the type signature for `stringify` is just `string` but it can return undefined too.

packages/studio-base/src/panels/ThreeDimensionalViz/TopicTree/TreeNodeRow.tsx
```diff
@@ -232,7 +232,7 @@ export default function TreeNodeRow({
             <LeadPencilIcon />
           </Icon>
         )}
-        {showGroupError && errorTooltip && sceneErrors && (
+        {showGroupError && errorTooltip && (
```
jtbandes, line 235: Should this have been `sceneErrors.length > 0` instead?

packages/studio-base/src/components/MessagePipeline/index.tsx
foxymiles: Good catch I'll fix.

packages/studio-base/src/panels/ThreeDimensionalViz/TopicTree/renderNamespaceNodes.tsx
```diff
@@ -86,7 +86,7 @@ function NamespaceNodeRow({
   topicName: string;
   onNamespaceOverrideColorChange: OnNamespaceOverrideColorChange;
 }) {
-  const nodeVisibleInScene = visibleInScene ?? false;
+  const nodeVisibleInScene = visibleInScene;
```
jtbandes, line 89: maybe remove this variable?

packages/studio-base/src/util/clipboard.ts
```diff
@@ -12,7 +12,7 @@
 //   You may not use this file except in compliance with the License.
 
 function fallbackCopy(text: string) {
-  const { body } = document;
+  const body = document.body as undefined | HTMLElement;
   if (body == undefined) {
     throw new Error("Could not find body, failed to copy.");
   }
```
jtbandes, line 18: I think this conditional dance can be removed. I believe it was only here because in Flow `body` is typed as optional, but that's not true in lib.d.ts.

foxymiles: Makes sense I wasn't sure why that was there.

packages/studio-base/src/panels/RawMessages/CopyMessageButton.tsx
foxymiles: Maybe we should wrap it with our own utility function that includes `undefined` in the type signature?

jtbandes: Maybe we can override with a custom TS definition.

Specifically it looks like according to MDN,
> undefined, Functions, and Symbols are not valid JSON values. If any such values are encountered during conversion they are either omitted (when found in an object) or changed to null (when found in an array). JSON.stringify() can return undefined when passing in "pure" values like JSON.stringify(function(){}) or JSON.stringify(undefined).

So we could make a set of overloads for `stringify(data: undefined|Function|Symbol): undefined;` and `stringify(data:unknown): string`?

packages/studio-base/src/util/typeUtils.ts
foxymiles: Actually it looks like that package hasn't been updated since Oct. 2020. Maybe it's better to hold off on adding that as a new dependency until we have more need for it?

packages/studio-base/src/components/MessagePipeline/index.tsx
foxymiles: I've reverted this and added an eslint-disable to silence the warning.

packages/studio-base/src/panels/ThreeDimensionalViz/TopicTree/TreeNodeRow.tsx
foxymiles: Yes I think you're right. Fixed.

packages/studio-base/src/util/typeUtils.ts
jtbandes: ok. I think a better place for these utility types might be `packages/den`. I'm sure we have several similar types scattered around, e.g. I believe we also have a DeepPartial somewhere.

desktop/main/StudioWindow.ts
```diff
@@ -372,7 +372,7 @@ class StudioWindow {
         const existingMenu = Menu.getApplicationMenu();
         const fileMenu = existingMenu?.getMenuItemById("fileMenu");
         // https://github.com/electron/electron/issues/8598
-        (fileMenu?.submenu as ClearableMenu)?.clear();
+        (fileMenu?.submenu as ClearableMenu).clear();
```
jtbandes, line 375: seems like this might be wrong, probably more accurate to the original intent to use `(fileMenu?.submenu as ClearableMenu | undefined)?.clear()`

foxymiles: Yes I think you're right. Good catch.

packages/studio-base/src/panels/RawMessages/CopyMessageButton.tsx
foxymiles: This was addressed in https://github.com/foxglove/studio/pull/2736.

packages/studio-base/src/util/typeUtils.ts
foxymiles: Looks like we already had a transient dependency on https://github.com/ts-essentials/ts-essentials which has all the type utilities we need for this so I just made it an explicit dependency and used it instead.

packages/studio-base/src/PanelAPI/useConfigById.ts
```diff
@@ -25,7 +25,7 @@ export default function useConfigById<Config extends Record<string, unknown>>(
       if (panelId == undefined) {
         return undefined;
       }
-      return state.selectedLayout?.data?.configById?.[panelId] as Config | undefined;
+      return state.selectedLayout?.data?.configById[panelId] as Config | undefined;
```
defunctzombie, line 28: We might want to tighten this up - the configById could be loaded as empty from disk or over api request?

packages/studio-base/src/types/mightActuallyBeOptional.ts
```diff
@@ -0,0 +1,14 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
defunctzombie, line 1: The file and the single function name don't match. And the file is in a _types_ folder but isn't a type. Thats rather un-expected.

desktop/main/rosPackageResources.ts
```diff
@@ -17,9 +17,9 @@ const log = Logger.getLogger(__filename);
 /** Extract a package name from a ROS package.xml file. */
 function rosPackageName(packageXmlContents: string) {
   const doc = new DOMParser().parseFromString(packageXmlContents, "text/xml");
-  const packageName = Array.from(doc.documentElement?.childNodes ?? []).find(
-    (n) => n.nodeName.toLowerCase() === "name",
-  )?.textContent;
+  const packageName = Array.from(
+    (doc.documentElement as Partial<typeof doc.documentElement>).childNodes ?? [],
```
jhurliman, line 21: The `?` after `documentElement` was dropped. Is this safe to do?

foxymiles: Will `documentElement` always exist?

packages/studio-base/src/types/mightActuallyBeOptional.ts
foxymiles: Good point I've moved this over to `utils`.

desktop/main/rosPackageResources.ts
foxymiles: I moved this back to marking `doc` as optional and chaining the rest.

packages/studio-base/src/PanelAPI/useConfigById.ts
foxymiles: Yep good point I threw a `DeepPartial` on there.

PR #2736, opened 2022-01-26
Mark JSON.stringify as possibly returning undefined.

foxymiles:
**User-Facing Changes**
None

**Description**
The typescript return type for `JSON.stringify` is `string` but it can in fact return `undefined` as well. This overrides the typing for this function to indicate that this is possible and force users of the function to handle this case.

### DIFF
--- a/packages/studio-base/src/typings/overrides.d.ts
+++ b/packages/studio-base/src/typings/overrides.d.ts
@@ -1,0 +1,28 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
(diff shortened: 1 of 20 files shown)


packages/studio-base/src/typings/overrides.d.ts
```diff
@@ -0,0 +1,27 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
+/* eslint-disable no-restricted-syntax */
+
+// We extend the type definition of JSON.stringify because in some cases
+// it can return undefined. This isn't strictly correct but it at least
+// forces callers to handle the undefined case and prevent crashes.
```
jtbandes, line 10: maybe link to TS github issue?

jtbandes (approved):

Issue #2715, opened 2022-01-21
Cannot get data from rosbridge_suite v1.1.0


mhl787156:
**Description**
I just upgraded my container's rosbridge_suite from v1.0.7 to latest on apt v1.1.0 and foxglove seems to have stopped working... It registers as a client, but doesn't seem to receive anything as shown here:
![image](https://user-images.githubusercontent.com/10742598/150558624-ec9f14e9-3c44-4d89-8b95-8e840ec4c04c.png)

Previously when running v1.0.7 I would get something like the following. This also includes an extra topic I manually published over `ros2 topic pub /test_topic geometry_msgs/msg/Point {}`:
![image](https://user-images.githubusercontent.com/10742598/150558885-44f8b056-ae45-410e-b004-efaba50d9e5a.png)

I have double checked with my own frontend running roslibjs connecting to the same rosbridge suite seems to work fine for both the older and newer versions. 

For context, I have rosbridge_suite running as a Docker Container running this simple container: https://github.com/StarlingUAS/ProjectStarling/blob/master/system/rosbridge-suite/Dockerfile 

It is possible that I might have misunderstood something or provided the wrong configuration. But I have found that the configuration is reproducable on other machines (with the docker file below at least). I have not attempted to test outside of the dockerfile natively as I do not have the correct machines! 

I have attempted to look into it, but I have no idea where to start. I thought it might be that the rosbridge_server launch file may no longer be valid, or the topics_glob handling had changed, but providing my own launchfile did not appear to fix anything. Foxglove doesn't provide any indications as to what might be going wrong, and neither does rosbridge! 

Many Thanks in advance

- Version: 0.27.0
- OS: Windows & Linux & Chrome

**Steps To Reproduce**

Testing rosbridge_suite v1.1.0:
1. Run the following container: ` docker run -it --rm --network test_default -p 9090:9090 uobflightlabstarling/rosbridge-suite:latest`. This runs the dockerfile linked above. 
2. In a seperate terminal, setup a rostopic pub by running ` docker run -it --rm --network test_default  uobflightlabstarling/rosbridge-suite:latest bash`, Source ros (`. /opt/ros/foxy/setup.bash`) and run `ros2 topic pub /test_topic geometry_msgs/msg/Point {}`
3. Open Foxglove desktop or web and connect to rosbridge instance on `localhost:9090`. 
4. You should see that rosbridge_suite has a new client connected and the foxglove does not complain, but it is all waiting for data.

[...]

PR #2746, opened 2022-01-27
Add timeout to RosbridgePlayer getTopicsAndRawTypes

defunctzombie:
**User-Facing Changes**
There is additional user feedback if a Rosbridge connection is unable to lookup topics and raw types.

**Description**

Some versions of Rosbridge Suite would hang on getTopicsAndRawTypes requests. This change adds timeout logic and problems reporting so the user
has feedback on what went wrong. Previously this hang would provide no feedback and the user would be left with no indication on why studio is not displaying their data.

Fixes: #2715

### DIFF
--- a/packages/studio-base/src/players/RosbridgePlayer.ts
+++ b/packages/studio-base/src/players/RosbridgePlayer.ts
@@ -181,12 +181,27 @@ export default class RosbridgePlayer implements Player {
       return;
     }
 
+    // getTopicsAndRawTypes might silently hang. When this happens, there is no indication to the user
+    // that the connection is doing anything and studio shows no errors and no data.
+    // This logic adds a warning after 5 seconds (picked arbitrarily) to display a notice to the user.
+    const topicsStallWarningTimeout = setTimeout(() => {
+      this._problems.addProblem("topicsAndRawTypesTimeout", {
+        severity: "warn",
+        message: "Taking too long to get topics and raw types.",
+      });
+
+      this._emitState();
+    }, 5000);
+
     try {
       const result = await new Promise<{
         topics: string[];
         types: string[];
         typedefs_full_text: string[];
       }>((resolve, reject) => rosClient.getTopicsAndRawTypes(resolve, reject));
+
+      clearTimeout(topicsStallWarningTimeout);
+      this._problems.removeProblem("topicsAndRawTypesTimeout");
 
       const topicsMissingDatatypes: string[] = [];
       const topics = [];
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/players/RosbridgePlayer.ts
```diff
@@ -182,12 +183,24 @@ export default class RosbridgePlayer implements Player {
...
+      const topicsAndRawTypesPromise = new Promise<{
         topics: string[];
         types: string[];
         typedefs_full_text: string[];
       }>((resolve, reject) => rosClient.getTopicsAndRawTypes(resolve, reject));
 
+      // getTopicsAndRawTypes might silently hang
+      // This logic adds a 2 second (picked arbitrarily) timeout to display an erorr to the user.
+      // There is a _finally_ clause at the end of the containing _try_ block which retries the topic request
+      const result = await Promise.race([topicsAndRawTypesPromise, delay(1000 * 2)]);
```
jtbandes, line 195: 2 sec might be a bit small, what if the node is just slow but does work? This leaves the app in an unusable state.

defunctzombie: What number would you like me to use?

jtbandes: 🤷 10 sec? Is there a way we can display the warning but still leave the request running?

defunctzombie: Yes. Its async - so we can kick off the request by making the promise (not await it). Kickoff a setTimeout that adds a problem and emits. await the request and clearTimeout so the timeout is not triggered if the request came back.

Thinking more about this - if you are on a remote connection then I can see this taking some time so maybe doing a warning is the right direction.

defunctzombie:
@jtbandes I've updated the logic to display a warning after a timeout but allow the request to continue.

jtbandes (approved):

packages/studio-base/src/players/RosbridgePlayer.ts
```diff
@@ -181,13 +181,28 @@ export default class RosbridgePlayer implements Player {
       return;
     }
 
+    // getTopicsAndRawTypes might silently hang. When this happens, there is no indication to the user
+    // that the connection is doing anything and studio shows no errors and no data.
+    // This logic adds a warning after 5 seconds (picked arbitrarily) to display a notice to the user.
+    const topicsStallWarningTimeout = setTimeout(() => {
```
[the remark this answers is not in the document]
defunctzombie: ? The request blocks the rest of the logic - which includes any retry logic. Is there a situation where we are sending requests again?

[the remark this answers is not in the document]
jtbandes: I see, you're right, I was thinking of the old version which still had Promise.race.

PR #2748, opened 2022-01-28
Refactor inline styles to use MUI Box

2metres:
**User-Facing Changes**
None

**Description**

### DIFF
--- a/packages/studio-base/src/components/ConnectionList/index.tsx
+++ b/packages/studio-base/src/components/ConnectionList/index.tsx
@@ -46,7 +46,7 @@ export default function ConnectionList(): JSX.Element {
   return (
     <>
       <DataSourceInfo />
-      {playerProblems.length > 0 && <Divider />}
+      {playerProblems.length > 0 && <Divider sx={{ marginY: 1 }} />}
       {playerProblems.map((problem, idx) => {
         const iconName = problem.severity === "error" ? "Error" : "Warning";
         const color =
--- a/packages/studio-base/src/components/ConnectionList/index.tsx
+++ b/packages/studio-base/src/components/ConnectionList/index.tsx
@@ -2,7 +2,8 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
-import { Icon, makeStyles, useTheme } from "@fluentui/react";
+import { Icon, useTheme } from "@fluentui/react";
+import { Box, Divider } from "@mui/material";
 import { useCallback, useContext } from "react";
 
 import {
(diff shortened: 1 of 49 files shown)


packages/studio-base/src/components/ConnectionList/index.tsx
```diff
@@ -65,23 +46,24 @@ export default function ConnectionList(): JSX.Element {
   return (
     <>
       <DataSourceInfo />
-      {playerProblems.length > 0 && <hr className={classes.divider} />}
+      {playerProblems.length > 0 && <Divider sx={{ marginY: 1 }} />}
```
defunctzombie, line 49: Don't know how much we care but this is "worse" performance wise than what we had before. Previously we had a static class - now we have the _sx_ property which we know isn't the fastest thing. I think we're ok in this particular component cause we've separated the _fast_ rendered bit from the slower top level stuff. But it is something to think about.

packages/studio-base/src/components/JsonTree/useGetItemStringWithTimezone.tsx
```diff
@@ -25,7 +26,13 @@ function getArrow(x: number, y: number) {
     return;
   }
   return (
-    <span style={{ transform: `rotate(${Math.atan2(-y, x)}rad)`, display: "inline-block" }}>→</span>
+    <Box
+      component="span"
+      display="inline-block"
+      sx={{ transform: `rotate(${Math.atan2(-y, x)}rad)` }}
```
defunctzombie, line 32: why do we prefer `sx` here instead of `style` property? `sx` is useful when you need to do all that clever stuff with theme right? Is this for consistency? Why is display not in sx?

packages/studio-base/src/components/PanelExtensionAdapter.tsx
```diff
@@ -527,10 +528,10 @@ function PanelExtensionAdapter(props: PanelExtensionAdapterProps): JSX.Element {
   }
 
   return (
-    <div style={{ width: "100%", height: "100%", overflow: "hidden", zIndex: 0, ...style }}>
+    <Box width="100%" height="100%" overflow="hidden" zIndex={0} sx={style}>
```
defunctzombie, line 531: Why do some properties get put in `sx` while others as properties on the component?

packages/studio-base/src/components/TimeBasedChart/HoverBar.tsx
```diff
@@ -69,5 +60,19 @@ export default React.memo<Props>(function HoverBar({
     return { visibility: "visible", transform: `translateX(${positionX}px)` };
   }, [positionX]);
 
-  return <SWrapper style={{ visibility, transform }}>{children}</SWrapper>;
+  return (
+    <Box
```
defunctzombie, line 64: The timebased chart component is one that might be rendering a lot. We've now taken a styled div which would be styled once with a static class and are using _sx_. I remain concerned we are making the app slower by a thousand small cuts when we could use makeStyles. I think we need to be very mindful on the use of _sx_. This concept is one designed with CRUD interface building - not performance sensitive rendering apps.

defunctzombie (commented):
My general spidey-sense is that we should be using makeStyles and tending toward _div_ elements, not toward _Box_ with _sx_ property. We've seen that _sx_ does impact render performance - at least under some uses. For webpages that render a few times or when users interact this isn't an issue - but for our app render time matters. If we can avoid the thing we know to be slow and the thing that is faster isn't harder to do I think we should strive to do that.

PR #2751, opened 2022-01-28
Wrap message sorting & partitioning in safe time comparison.

### LINKED ISSUE #2750: Switching playback order from receive time to header stamp in nuscenes throws an error.
**Description**
Switching playback order from receive time to header stamp in nuscenes throws an error.

- Version: 0.28.0
- OS: macOS

**Steps To Reproduce**

1. Open the default nuscenes session.
2. Switch playback order from receive time to header stamp in the playback controls bar.

**Expected Behavior**
It should switch playback ordering.

**Actual Behavior**
It throws an error.

Screen Shot 2022-01-28 at 7 47 06 AM

foxymiles:
**User-Facing Changes**
This fixes an issue in message playback with message ordering set to `header stamp`.

**Description**
We have a wrapper function to extract the time from messages that carry a timestamp either in their header or in their `markers` subfield but the partitioning and comparison functions we use on them weren't using this function but trying to access the header directly.

Fixes #2750

### DIFF
--- a/packages/studio-base/src/players/OrderedStampPlayer.ts
+++ b/packages/studio-base/src/players/OrderedStampPlayer.ts
@@ -54,6 +54,24 @@ const getTopicsWithHeader = memoizeWeak((topics: Topic[], datatypes: RosDatatype
   });
 });
 
+function compareMessageEventsByTime(
+  a: MessageEvent<StampedMessage>,
+  b: MessageEvent<StampedMessage>,
+): number {
+  const timeA = getTimestampForMessage(a.message);
+  const timeB = getTimestampForMessage(b.message);
+  if (timeA == undefined && timeB == undefined) {
+    return 0;
+  }
+  if (timeA == undefined) {
+    return -1;
+  }
+  if (timeB == undefined) {
+    return 1;
+  }
+  return compare(timeA, timeB);
+}
+
 export default class OrderedStampPlayer implements Player {
   private _player: UserNodePlayer;
   private _messageOrder: TimestampMethod;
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/players/OrderedStampPlayer.ts
```diff
@@ -54,6 +54,24 @@ const getTopicsWithHeader = memoizeWeak((topics: Topic[], datatypes: RosDatatype
...
+    return 0;
+  }
+  if (timeA == undefined) {
+    return -1;
+  }
+  if (timeB == undefined) {
+    return 1;
+  }
+  return compare(timeA, timeB);
+}
```
jhurliman, line 73: If we assume having timestamps for both messages is the common case, here's a variant that uses a single branch for that case:

```suggestion
function compareMessageEventsByTime(
  a: MessageEvent<StampedMessage>,
  b: MessageEvent<StampedMessage>,
): number {
  const timeA = getTimestampForMessage(a.message);
  const timeB = getTimestampForMessage(b.message);
  if (timeA == undefined || timeB == undefined) {
    if (timeA == undefined && timeB == undefined) {
      return 0;
    } else if (timeA == undefined) {
      return -1;
    } else {
      return 1;
    }
  }
  return compare(timeA, timeB);
}
```

jhurliman (approved):

packages/studio-base/src/players/OrderedStampPlayer.ts
foxymiles: Good optimization. Thanks!

PR #2754, opened 2022-01-28
Disable minimization in storybook build to improve performance

jtbandes:
**User-Facing Changes**
None

**Description**
The storybook config used TerserPlugin for minimization which was slow. We don't really need minimization for storybook, so just disable it. Cuts `yarn storybook:build` time in half on my M1 Pro.


jtbandes:
😄 
image

Issue #2343, opened 2021-12-09
Newly subscribed panels never receive old messages when using live data sources


jtbandes:
**Description**
When Studio subscribes to a latched message, it immediately receives the message and displays it in any subscribed panels. However, if panels are added _later_ that subscribe to the same topic, they will not receive the old message even if it is still visible in other panels.

**Steps To Reproduce**
- Run `ros2 launch rosbridge_server rosbridge_websocket_launch.xml` (must use a version incorporating fix from https://github.com/RobotWebTools/rosbridge_suite/pull/690)
- Run `ros2 run tf2_ros static_transform_publisher 0 0 0 0 0 0 a b`
- Connect Studio via Rosbridge (ROS 2)
- Create a new layout in Studio.
- Add a 3D panel. Under the topic picker, the "TF" section now contains `a` and `b`.
- Remove all panels from the layout.
- Add a Raw Messages panel and subscribe to `/tf_static`. A message appears in the panel.
- Add a 3D panel. Under the topic picker, the "TF" section is empty.

**Expected Behavior**
The static TFs should be available to the 3D panel regardless of which order panels are added.

**Actual Behavior**
Static TFs are only visible if the 3D panel was the first panel to subscribe to `/tf_static`.

jtbandes:
Possible fix: Player or Message Pipeline could keep the most recent message on each topic in memory

defunctzombie:
I tried this with a local bag file and found that I could open an image panel then a raw message panel for the same topic. I'm thinking the current app behavior for message pipeline sends the subscription for new panels down to the player and if the player had the messages stored then it sends the messages to the new subscriber.

I bring that up to think about how any caching in message pipeline could interfere with subscribing to the player and getting messages from the player. Its interesting to think about handling this generically at the message pipeline layer tho I wonder if we should leave it up to the player.

jhurliman:
@defunctzombie thoughts on how to move forward with this? Caching at the message pipeline layer or per player is fine IMO (although the latter will require getting it right in many more places and eventually user-written data source extensions as well).

defunctzombie:
I've been trying to come up with a strategy and the only one I've been able to come up with is that its a _per player_ issue.

I started looking at "ok maybe I can make message pipeline hang on to the last message on any topic and when you subscribe you get that" but ran into the above note about interfering with what the player does. For example the random access player has the following logic:

[...]

PR #2758, opened 2022-01-31
Update rosbridge player to re-subscribe to all topics

defunctzombie:
**User-Facing Changes**
When connected to Rosbridge and opening multiple panels that need latching topics (like `/tf_static`), all panels will receive the message.

**Description**

Previously, new panel subscriptions to existing latching topics would
never receive the latched message. This happened because the RosbridgePlayer would ignore subscriptions to topics it was already subscribed to but would not send any previous messages on these topics to the new subscribers.

This change updates RosbridgePlayer logic to perform an unsubscribe/resubscribe cycle whenever the topic subscriptions change. This allows the underlying rosbridge server to process the subscription again and send any latched messages.

See: #2343

### DIFF
--- a/packages/studio-base/src/players/RosbridgePlayer.ts
+++ b/packages/studio-base/src/players/RosbridgePlayer.ts
@@ -383,6 +383,20 @@ export default class RosbridgePlayer implements Player {
       return;
     }
 
+    // Unsubscribe from all topics. When a panel subscribes to an existing topic, we want it to
+    // receive the last message on the topic. This is especially important for topics like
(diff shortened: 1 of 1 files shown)


jtbandes (approved):
Seems ok for now. Agree that it doesn't make sense for the players to handle backfill while message pipeline handles re-delivering latched/latest messages. So putting this in the player seems like the right thing to do. Each player will need a different implementation and this seems like an expedient one for rosbridge. An optimization would be to only resubscribe if studio internal subscriptions have changed for that topic.

PR #2764, opened 2022-01-31
Re-subscribe to all topics when changing subscribers for ros2

defunctzombie:
**User-Facing Changes**
When connected to ROS2 and opening multiple panels that need latching topics (like /tf_static), all panels will receive the message.

**Description**
Previously, new panel subscriptions to existing latching topics would never receive the latched message. This happened because the RosbridgePlayer would ignore subscriptions to topics it was already subscribed to but would not send any previous messages on these topics to the new subscribers.

This change updates Ros2Player to performan an unsubscribe/subsribe cycle whenever the topic subscriptions change. This establishes new subscriptions which re-send any latched messages.

See: #2343

### DIFF
--- a/packages/studio-base/src/players/Ros2Player.ts
+++ b/packages/studio-base/src/players/Ros2Player.ts
@@ -344,7 +344,19 @@ export default class Ros2Player implements Player {
       return;
     }
 
-    // Subscribe to additional topics used by Ros1Player itself
+    // Unsubscribe from all topics.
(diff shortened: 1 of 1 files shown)


defunctzombie:
See https://github.com/foxglove/studio/pull/2758

defunctzombie:
After discussion - we've determined unsubscribing/re-subscribing is not so healthy for a ros2 system and will try a different approach: https://github.com/foxglove/studio/pull/2778.

If that approach doesn't work we can revisit this one.

Issue #2759, opened 2022-01-31
Foxglove crashes when a PointCloud2 object does not contain `x`, `y` and `z` fields


KerstinKeller:
**Description**
If a PointCloud2 object, that is to be displayed in the 3D panel, is missing any of the `x`, `y` and `z` fields, the whole app crashes

- Version: 0.29.0
- OS: Windows

![image](https://user-images.githubusercontent.com/13848742/151845098-f3cd7d1d-180e-4a24-846c-dd704ed2023f.png)
This is ok, but if e.g. the third entry with `z` is not present, the crash occurs.

**Steps To Reproduce**

![image](https://user-images.githubusercontent.com/13848742/151844844-9ffed57d-3d00-41c7-b1f2-887acb03d434.png)

**Expected Behavior**
No crash, maybe an error indication next to the topic.
(like for the missing transforms)

![image](https://user-images.githubusercontent.com/13848742/151845513-2c3284a9-4fb7-4955-8e2a-dd4642125a23.png)

**Actual Behavior**
App Crashes

PR #2766, opened 2022-02-01
Update SceneBuilder to check for required PointCloud fields.

defunctzombie:
**User-Facing Changes**
Rather than crashing the app when fields are missing from the PointCloud message, the user is shown a topic error next to the malformed topic.

image

Any other rendering errors result in a panel crash.

**Description**
PointCloud rendering requires field descriptions for x, y, and z fields. This change updates SceneBuilder to check for these fields prior to adding the point cloud to the collector. This allows the UI to display a topic error for the malformed topic better informing the user to errors with their data.

Fixes: #2759

### DIFF
--- a/packages/studio-base/src/panels/ThreeDimensionalViz/SceneBuilder/index.ts
+++ b/packages/studio-base/src/panels/ThreeDimensionalViz/SceneBuilder/index.ts
@@ -995,7 +995,7 @@ export default class SceneBuilder implements MarkerProvider {
         if (missingFields.length > 0) {
           this._setTopicError(
             topic.name,
-            `Missing required fields entry for field names: ${missingFields}`,
+            `Point cloud is missing required fields: ${missingFields}`,
           );
           return;
         }
--- a/packages/studio-base/src/panels/ThreeDimensionalViz/SceneBuilder/index.ts
+++ b/packages/studio-base/src/panels/ThreeDimensionalViz/SceneBuilder/index.ts
@@ -974,8 +974,34 @@ export default class SceneBuilder implements MarkerProvider {
         return add.triangleList(marker);
       case 101:
         return add.grid(marker);
-      case 102:
+      case 102: {
+        // PointCloud decoding requires x, y, and z fields and will fail if all are not present.
+        // We check for the fields here so we can present the user with a topic error prior to decoding.
+        const fieldNames: { [key: string]: boolean } = {};
+        for (const field of marker.fields) {
+          fieldNames[field.name] = true;
+        }
+
+        let missingFields = "";
+        if (fieldNames.x == undefined) {
+          missingFields += "x";
+        }
+        if (fieldNames.y == undefined) {
+          missingFields += " y";
+        }
+        if (fieldNames.z == undefined) {
+          missingFields += " z";
+        }
+        if (missingFields.length > 0) {
+          this._setTopicError(
+            topic.name,
+            `Missing required fields entry for field names: ${missingFields}`,
+          );
+          return;
+        }
+
         return add.pointcloud(marker);
+      }
       case 103:
         return add.poseMarker(marker);
       case 104:
--- a/packages/studio-base/src/panels/ThreeDimensionalViz/commands/PointClouds/index.tsx
+++ b/packages/studio-base/src/panels/ThreeDimensionalViz/commands/PointClouds/index.tsx
@@ -31,7 +31,6 @@ import {
 } from "@foxglove/studio-base/panels/ThreeDimensionalViz/TopicSettingsEditor/PointCloudSettingsEditor";
 import { toRgba } from "@foxglove/studio-base/panels/ThreeDimensionalViz/commands/PointClouds/selection";
 import { mightActuallyBePartial } from "@foxglove/studio-base/util/mightActuallyBePartial";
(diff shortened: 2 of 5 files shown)


packages/studio-base/src/panels/ThreeDimensionalViz/SceneBuilder/index.ts
```diff
@@ -974,8 +974,34 @@ export default class SceneBuilder implements MarkerProvider {
         return add.triangleList(marker);
       case 101:
         return add.grid(marker);
-      case 102:
+      case 102: {
+        // PointCloud decoding requires x, y, and z fields and will fail if all are not present.
```
defunctzombie, line 978: Is this an acceptable price to pay? Probably no worse than all the other logic that happens for markers? I could also do this for the first message on each topic and assume follow-up messages are ok?

packages/studio-base/src/panels/ThreeDimensionalViz/SceneBuilder/index.ts
```diff
@@ -974,8 +974,34 @@ export default class SceneBuilder implements MarkerProvider {
...
+        // We check for the fields here so we can present the user with a topic error prior to decoding.
+        const fieldNames: { [key: string]: boolean } = {};
```
jtbandes, line 980: `Set`?

packages/studio-base/src/panels/ThreeDimensionalViz/SceneBuilder/index.ts
```diff
@@ -974,8 +974,34 @@ export default class SceneBuilder implements MarkerProvider {
...
+        if (fieldNames.y == undefined) {
+          missingFields += " y";
+        }
+        if (fieldNames.z == undefined) {
+          missingFields += " z";
+        }
+        if (missingFields.length > 0) {
+          this._setTopicError(
+            topic.name,
+            `Missing required fields entry for field names: ${missingFields}`,
```
jtbandes, line 998:
```suggestion
            `Point cloud is missing required fields: ${missingFields}`,
```

packages/studio-base/src/panels/ThreeDimensionalViz/commands/PointClouds/index.tsx
```diff
@@ -444,30 +443,14 @@ type Props = CommonCommandProps & {
...
-    : (children as PointCloudMarker[])
-        .map((m) => {
-          try {
-            return decodeMarker(m);
-          } catch (err) {
-            sendNotification("Point cloud decoding failed", err, "user", "error");
-            return undefined;
-          }
-        })
-        .filter((m) => m != undefined);
```
jtbandes, line 470: So by removing this, we now crash the whole panel when a point cloud can't be rendered?

packages/studio-base/src/util/sendNotification.ts
```diff
@@ -48,7 +48,7 @@ const defaultNotificationHandler: NotificationHandler = (
   type: NotificationType,
   severity: NotificationSeverity,
 ): void => {
-  if (!inWebWorker()) {
+  if (inWebWorker()) {
```
jtbandes, line 51: :philosorapter:

jtbandes (approved):

packages/studio-base/src/panels/ThreeDimensionalViz/commands/PointClouds/index.tsx
defunctzombie: Yes - I'm open to other ideas but IMO send notification is not the correct UX strategy.

packages/studio-base/src/panels/ThreeDimensionalViz/SceneBuilder/index.ts
defunctzombie: Significantly slower /shrug

packages/studio-base/src/util/sendNotification.ts
defunctzombie: Yea - I did not understand this. Its like this in [webviz](https://github.com/cruise-automation/webviz/blob/master/packages/webviz-core/src/util/sendNotification.js#L38) as well. What am I missing?

packages/studio-base/src/panels/ThreeDimensionalViz/commands/PointClouds/index.tsx
jhurliman: Should the PR description be updated to mention that all other pointcloud rendering errors have been converted into a panel crash?

packages/studio-base/src/util/sendNotification.ts
jtbandes: Probably nothing.

packages/studio-base/src/panels/ThreeDimensionalViz/commands/PointClouds/index.tsx
defunctzombie: Updated the PR description to note this. I'm also up to hearing how else to handle this that doesn't involve sending a notification to the user on every attempt to render a frame.

AndyYangjd:
The studio can't render the pointcloud. The error info is: non-identity pose is not yet supported.


Is there any config error?

PR #2770, opened 2022-02-02
Render grid using native GL lines, remove line width setting

### LINKED ISSUE #2742: 3D panel Grid - Line size scales with the grid size
**Description**

Screen Shot 2022-01-26 at 9 12 31 PM

**Steps To Reproduce**

Screen Shot 2022-01-26 at 9 15 50 PM

jtbandes:
**User-Facing Changes**
Improved 3D panel grid rendering for large grids; removed the line width setting for grids.

**Description**
Closes #2742.

In investigating this ticket, we learned that the problem with our scale-invariant line rendering is that the [two](https://github.com/foxglove/regl-worldview/blob/92df7aad8d28d15a811d5daaed672261c507e27a/src/commands/Lines.js#L173-L176) [places](https://github.com/foxglove/regl-worldview/blob/92df7aad8d28d15a811d5daaed672261c507e27a/src/commands/Lines.js#L144-L145) where we manually multiply/divide by `w` does not match the way the GPU handles vertex positions in extreme cases, such as `w < 0` (a vertex behind the camera). This is because the GPU [performs clipping](https://stackoverflow.com/a/41087445/23649) before perspective division. Additionally, [varying values may be incorrect](https://webglfundamentals.org/webgl/lessons/webgl-3d-perspective-correct-texturemapping.html) when implementing perspective division manually.

A complete fix to the instanced line renderer does not seem within reach in the short term, so to address the issue this PR converts the Grid feature to use native GL lines and removes the thickness setting. (Although the GL line width is configurable, the limits may vary by browser/platform and [MDN recommends](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/lineWidth) to only use the default thickness of 1.)

### DIFF
--- a/packages/studio-base/src/panels/ThreeDimensionalViz/commands/GlLineLists.tsx
+++ b/packages/studio-base/src/panels/ThreeDimensionalViz/commands/GlLineLists.tsx
@@ -62,11 +62,13 @@ const rawCommand = (regl: REGL.Regl) => {
 };
 
 type Props = CommonCommandProps & {
-  // TypeScript doesn't allow us to pass an array variable if `children` is set to an array type here
-  // https://github.com/microsoft/TypeScript/issues/30711#issuecomment-485013588
-  children: React.ReactNode;
+  glLineLists: GlLineListMarker[];
 };
 
 export default function GlLineLists(props: Props): JSX.Element {
-  return <Command getChildrenForHitmap={undefined} {...props} reglCommand={rawCommand} />;
+  return (
+    <Command getChildrenForHitmap={undefined} {...props} reglCommand={rawCommand}>
+      {props.glLineLists}
+    </Command>
+  );
 }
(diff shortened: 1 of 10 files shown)


packages/studio-base/src/panels/ThreeDimensionalViz/commands/GlLineLists.tsx
```diff
@@ -0,0 +1,72 @@
...
+    },
+
+    count: (_ctx, props) => props.points.length / 3,
+  });
+};
+
+type Props = CommonCommandProps & {
+  // TypeScript doesn't allow us to pass an array variable if `children` is set to an array type here
+  // https://github.com/microsoft/TypeScript/issues/30711#issuecomment-485013588
+  children: React.ReactNode;
```
defunctzombie, line 67: Since we are writing this command here - let's use a prop rather than this _children_ pattern. Our prop will be typed correctly.

defunctzombie (approved):

Issue #1589, opened 2021-08-03
First entry in Add Panel is highlighted


rgov:
**Description**

When I open the Add Panel sidebar, the first entry in the list, "3D", is highlighted by default. It stays highlighted even if you mouseover other entries. This makes it read like it is a grouping header instead of one of the panels. 

<img width="380" alt="image" src="https://user-images.githubusercontent.com/108767/128086894-5e27f7c2-742b-46c5-8c47-16da8110335b.png">

- Version: 0.14.0
- OS: macOS

jtbandes:
FYI, it's highlighted because the list supports keyboard navigation. You can press up/down arrow keys to select items. Not saying this is ideal, just an explanation of why it's like this.

rgov:
On Mac apps, such as the Finder, the first time you hit <kbd>↓</kbd> it highlights the first row. If you hit <kbd>↑</kbd> it highlights the last row.

amacneil:
> On Mac apps, such as the Finder, the first time you hit ↓ it highlights the first row. If you hit ↑ with no prior selection it highlights the last row.

Agreed, this would be better behavior than highlighting it by default

defunctzombie:
Should this also un-highlight the first item when you've used the mouse to hover over another item? Then keyboard keys would navigate from the hovered item?

amacneil:
Still a problem

PR #2771, opened 2022-02-02
Avoid automatically highlighting the first entry in the Add Panel List

defunctzombie:
**User-Facing Changes**

When showing the add panel list, no entries are highlighted by default. If a user types in the search box, the first entry is highlighted - allowing for quickly filtering and pressing "enter" to add the panel.

When the user clears the search box, highlighting is removed.

Keyboard navigation works as before - with "key down" moving down the list and "key up" moving up. Both wrapping around at each end of the list.

Fixes: #1589

### DIFF
--- a/packages/studio-base/src/components/PanelList/index.tsx
+++ b/packages/studio-base/src/components/PanelList/index.tsx
@@ -363,8 +363,7 @@ function PanelList(props: Props): JSX.Element {
           if (existing == undefined) {
             return allFilteredPanels.length - 1;
           }
-          const newIdx = (existing - 1) % (allFilteredPanels.length - 1);
-          return newIdx >= 0 ? newIdx : allFilteredPanels.length + newIdx;
+          return (existing - 1 + allFilteredPanels.length) % allFilteredPanels.length;
         });
       } else if (e.key === "Enter" && highlightedPanel) {
         onPanelSelect({
--- a/packages/studio-base/src/components/PanelList/index.tsx
+++ b/packages/studio-base/src/components/PanelList/index.tsx
@@ -360,6 +360,11 @@ function PanelList(props: Props): JSX.Element {
         });
       } else if (e.key === "ArrowUp") {
         setHighlightedPanelIdx((existing) => {
+          // nothing to highlight if there are no entries
+          if (allFilteredPanels.length <= 0) {
+            return undefined;
+          }
+
           if (existing == undefined) {
             return allFilteredPanels.length - 1;
           }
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/components/PanelList/index.tsx
```diff
@@ -347,11 +351,21 @@ function PanelList(props: Props): JSX.Element {
...
+          }
+          return (existing + 1) % allFilteredPanels.length;
+        });
+      } else if (e.key === "ArrowUp") {
+        setHighlightedPanelIdx((existing) => {
+          if (existing == undefined) {
+            return allFilteredPanels.length - 1;
+          }
+          const newIdx = (existing - 1) % (allFilteredPanels.length - 1);
+          return newIdx >= 0 ? newIdx : allFilteredPanels.length + newIdx;
```
jtbandes, line 367: Initially I was just going to suggest a cleanup, then I noticed the `% (length - 1)` which I think is kinda wrong, but maybe doesn't matter because `existing - 1` will never be less than `-(length - 1)` unless `length <= 1`.

Actually what happens if length<=1? Can we even get here?

```suggestion
          return (existing - 1 + allFilteredPanels.length) % allFilteredPanels.length;
```

packages/studio-base/src/components/PanelList/index.tsx
```diff
@@ -347,11 +351,20 @@ function PanelList(props: Props): JSX.Element {
       if (mode === "grid") {
         return;
       }
-      if (e.key === "ArrowDown" && highlightedPanelIdx != undefined) {
-        setHighlightedPanelIdx((highlightedPanelIdx + 1) % allFilteredPanels.length);
-      } else if (e.key === "ArrowUp" && highlightedPanelIdx != undefined) {
-        const newIdx = (highlightedPanelIdx - 1) % (allFilteredPanels.length - 1);
-        setHighlightedPanelIdx(newIdx >= 0 ? newIdx : allFilteredPanels.length + newIdx);
+      if (e.key === "ArrowDown") {
```
jtbandes, line 354: maybe we should bail out if no results here so we don't risk any negative indices?

jtbandes (approved):

packages/studio-base/src/components/PanelList/index.tsx
defunctzombie: Could do that too. We just need to pick something and go with it. We can always re-visit anytime since its internal logic.

jtbandes: No strong feeling, just trying to think through how this might break. I approved so if you wanna merge as is that's fine

defunctzombie: The old code did have a >= 0 check that is now gone after I accepted the suggestion. I'm not sure why we decided to rework this.

defunctzombie: I added a check for length 0

PR #2772, opened 2022-02-02
Fix logic for tooltip hover presence.

### LINKED ISSUE #2704: Tooltip value on plots is not working.
**Description**
Hovering over values on plots to display the values under the cursor is not working.

- Version: 0.27.0
- OS: macOS

**Steps To Reproduce**

1. Start with an empty plot panel & demo.bag.
2. Add a new series of `/radar/tracks.tracks[0].range`
3. Hover the cursor over the plot and no tooltip is shown.
4. Hover over the legend.
5. Hover over the plot again and now tooltips are shown.

**Expected Behavior**
Tooltip should be shown immediately.

**Actual Behavior**
Tooltip does not appear until other UI elements are hovered.

https://user-images.githubusercontent.com/93935560/150360343-7a753b34-aca5-47f8-8046-a5a8f72b2a91.mov

foxymiles:
**User-Facing Changes**
This fixes an issue that prevents tooltips from being shown on newly selected message paths in plots.

**Description**
The existing logic tracks a "mouse present" state and sets it to false when the mouse leaves the canvas. This seems to be incorrect when the mouse leaves in order to select a new path for the series and the mouse enter event when the user hovers the plot again doesn't register. This patch just assumes if we get a `onMouseMove` event then the cursor must be hovering the canvas.

Fixes #2704

### DIFF
--- a/packages/studio-base/src/components/Chart/index.tsx
+++ b/packages/studio-base/src/components/Chart/index.tsx
@@ -407,20 +407,23 @@ function Chart(props: Props): JSX.Element {
 
   const { onHover } = props;
   const onMouseMove = useCallback(
-    (event: React.MouseEvent<HTMLCanvasElement>) => {
-      mousePresentRef.current = true; // Mouse must be present if we get the move event.
+    async (event: React.MouseEvent<HTMLCanvasElement>) => {
+      mousePresentRef.current = true; // The mouse must be present if we're getting this event.
+
       if (onHover == undefined || rpcSendRef.current == undefined) {
         return;
       }
 
-      void rpcSendRef
-        .current<RpcElement[]>("getElementsAtEvent", { event: rpcMouseEvent(event) })
-        .then((elements) => {
-          // Check again if mouse has left the canvas while waiting for the RPC call.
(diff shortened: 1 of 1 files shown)


defunctzombie (approved):
I think the original idea that the comment alluded to was incorrectly implemented in the main version.

I don't think it's worth handling for now but here is what I was trying to avoid: The particular nuance is that the `await` might resolve (and onHover called) after the user has moved their mouse away. It's also an issue for moving the mouse within the display too.

foxymiles:
> I think the original idea that the comment alluded to was incorrectly implemented in the main version.
> 
> I don't think it's worth handling for now but here is what I was trying to avoid: The particular nuance is that the `await` might resolve (and onHover called) after the user has moved their mouse away. It's also an issue for moving the mouse within the display too.

I see what you mean. Maybe we can revisit this later to refine this logic?

defunctzombie:
> Maybe we can revisit this later to refine this logic?

👍

foxymiles:
@defunctzombie I tweaked the logic here to preserve the original intent of the mouse tracking code. Maybe this is better?

PR #2775, opened 2022-02-02
Fix warning in PlotLegendRow.

### LINKED ISSUE #2774: Unknown event handler property warning thrown from PlotLegendRow component
**Description**
The `PlotLegendRow` component throws this warning:

```Warning: Unknown event handler property `onTimestampMethodChange`. It will be ignored.```

- Version: 0.29.0
- OS: macOS

**Steps To Reproduce**

Load any session with a plot panel.

Screen Shot 2022-02-02 at 9 42 54 AM

### PR DESCRIPTION
**Pull request by foxymiles**
**User-Facing Changes**
None

**Description**
This fixes a warning in the PlotLegendRow component. The issue is that props were being passed to `IconButton` via a spread operator but this was including a new `onTimestampMethodChange` property that `IconButton` doesn't accept. The solution here is to remove that property from the `rest` prop before passing it on.

Fixes #2774

### CODE REVIEW

File: packages/studio-base/src/components/TimestampMethodDropdown.tsx [hunk 1]
```diff
@@ -47,7 +47,7 @@ export default function TimestampMethodDropdown(props: Props): JSX.Element {
     setAnchorEl(event.currentTarget);
   };
 
-  const { path, timestampMethod = "receiveTime", ...rest } = props;
+  const { path, timestampMethod = "receiveTime", onTimestampMethodChange, ...rest } = props;
```
defunctzombie, line 50: @2metres maybe the pattern is to avoid this type of rest props style and use an explicit `iconButtonProps` property. We've seen this before with other frameworks.

foxymiles: Yeah that's probably safer in general.

defunctzombie (approved):
While here we should move the icon props into an iconProps prop to avoid this happening in the future and give readers better code patterns to copy.

foxymiles:
> While here we should move the icon props into an iconProps prop to avoid this happening in the future and give readers better code patterns to copy.

Makes sense. @esthersweon @2metres Do either of you want to update this as per @defunctzombie's suggestion?

PR #2778, opened 2022-02-03
Inject the last message on a topic on new subscriptions

defunctzombie:
**User-Facing Changes**
When adding a panel which needs to subscribe to an infrequently publishing topic (i.e. /tf_static), the new panel will receive the message.

**Description**
This change updates the MessagePipelineProvider to add logic for injecting the last message on a topic to new subscribers of the topic. The provider tracks when new panels subscribe, via setSubscriptions calls, and makes a note on any new topics requested by the panel. When a new player state arrives, the pipeline checks any of these new topics and sends the last message on any of the topics to the newly subscribed panels.

To avoid sending the last message to previously subscribed panels, the provider adds a new context field `messageEventsBySubscriberId`. `useMessageReducer` and `PanelExtensionAdapter` read their message events from these per-subscriber queues rather than from a single `messages` firehose. This mechanism allows the provider to target message event updates to specific subscribers.

Once a new subscriber has received a backfilled message, it is cleared from the new subscriber list to avoid it receiving any further backfill messages. Only the last message on a topic is provided as backfill.

Fixes: #2343


packages/studio-base/src/players/RosbridgePlayer.ts
```diff
@@ -428,7 +434,11 @@ export default class RosbridgePlayer implements Player {
 
     // Subscribe to topics
     for (const topicName of topicNames) {
+      // New panels might subscribe to topics we've already seen.
+      // Add any topic we've already seen to the _newSubscriberTopics set.
+      // When emitting, we will inject the last message on any topic in this set.
       if (this._topicSubscriptions.has(topicName)) {
+        this._newSubscriberTopics.add(topicName);
```
jtbandes, line 441: seems like we maybe need to emitState here?

jtbandes (commented):
This approach doesn't seem crazy to me overall. My main question is, are we sure this doesn't break assumptions of the message pipeline / message reducer? If I `useMessagesByTopic({historySize: 2})` am I still guaranteed to never get 2 copies of the same message?

packages/studio-base/src/components/MessagePipeline/index.tsx
```diff
@@ -199,6 +205,77 @@ export function MessagePipelineProvider({
...
+      const topicToSubscriberIds = new Map<string, string[]>();
+      for (const [id, subs] of subsById) {
+        for (const subscription of subs) {
+          const topic = subscription.topic;
+
+          const ids = topicToSubscriberIds.get(topic) ?? [];
+          ids.push(id);
+          topicToSubscriberIds.set(topic, ids);
+        }
+      }
```
jtbandes, line 221: This seems like an expensive thing to do on every frame when it is unlikely to change (only changes when subscriptions change?)

packages/studio-base/src/components/MessagePipeline/index.tsx
```diff
@@ -199,6 +205,77 @@ export function MessagePipelineProvider({
...
+        const newTopics = newTopicsBySubscriberId.current.get(id);
+        if (!newTopics) {
+          continue;
+        }
+        for (const topic of newTopics) {
+          // If we had a message for this topic in the regular set of messages, we don't need to inject
+          // another message.
+          if (seenTopics.has(topic)) {
+            continue;
+          }
```
jtbandes, line 264: didn't we discuss wanting to _always_ include old messages if they are available for new subscribers?

packages/studio-base/src/components/MessagePipeline/index.tsx
```diff
@@ -230,22 +307,39 @@ export function MessagePipelineProvider({
...
   const setSubscriptions = useCallback(
     (id: string, subscriptionsForId: SubscribePayload[]) => {
       setAllSubscriptions((previousSubscriptions) => {
-        return { ...previousSubscriptions, [id]: subscriptionsForId };
+        const previousSubscriptionById = previousSubscriptions.get(id);
+
+        // Record any _new_ topics for this subscriber into newTopicsBySubscriberId
+        const newTopics = newTopicsBySubscriberId.current.get(id);
+        if (!newTopics) {
+          newTopicsBySubscriberId.current.set(
```
jtbandes, line 328: Side effects in setState function. We should avoid doing this. It feels like maybe you want to replace these useState+useRef with a `useReducer` so you can compute new values based on changes that happen.

packages/studio-base/src/panels/RawMessages/index.stories.tsx
```diff
@@ -64,7 +59,7 @@ storiesOf("panels/RawMessages", module)
       </PanelSetup>
     );
   })
-  .add("display big value – num", () => {
+  .add("display big value - num", () => {
```
jtbandes, line 62: 🤔

packages/studio-base/src/panels/RawMessages/index.tsx
```diff
@@ -156,7 +157,7 @@ function RawMessages(props: Props) {
   }, [datatypes, topic, topicRosPath]);
 
   // When expandAll is unset, we'll use expandedFields to get expanded info
-  const [expandAll, setExpandAll] = useState<boolean | undefined>(false);
+  const [expandAll, setExpandAll] = useState<boolean | undefined>(props.defaultExpandAll ?? false);
```
jtbandes, line 160: No longer can be undefined
```suggestion
  const [expandAll, setExpandAll] = useState(props.defaultExpandAll ?? false);
```

packages/studio-base/src/panels/ThreeDimensionalViz/useFrame.test.tsx
```diff
@@ -59,7 +59,8 @@ describe("useFrame", () => {
 
     const { result, rerender } = renderHook(
       () => {
-        return useFrame(["/some/topic"]);
+        const ret = useFrame(["/some/topic"]);
+        return ret;
```
jtbandes, line 63: clean up debug code
```suggestion
        return useFrame(["/some/topic"]);
```

packages/studio-base/src/players/types.ts
```diff
@@ -241,10 +241,6 @@ export type Progress = Readonly<{
   readonly messageCache?: BlockCache;
 }>;
 
-export type Frame = {
```
jtbandes, line 244: slowly the evil shall be contained.

packages/studio-base/src/components/MessagePipeline/index.tsx
defunctzombie: yea - I think it needs to move into the code that changes when topics/subscribers change. The original reason I was doing it here is cause I don't want to make a new callback for the setListener when those change - what I want is just the latest values of these things.

packages/studio-base/src/components/MessagePipeline/index.tsx
defunctzombie: did we? Why would we do that if the intent is to provide _at least_ one message on the topic?

packages/studio-base/src/components/MessagePipeline/index.tsx
defunctzombie: Yea - I agree - I want to move these out to a useMemo or useReducer.

packages/studio-base/src/panels/RawMessages/index.stories.tsx
defunctzombie: white space characters /shrug

packages/studio-base/src/panels/RawMessages/index.tsx
defunctzombie: There is code elsewhere in the file that sets expand all to undefined. That code could set it to false instead.

packages/studio-base/src/components/MessagePipeline/usePlayerState.ts
```diff
@@ -0,0 +1,197 @@
...
+      ids.push(id);
+      subscriberIdsByTopic.set(topic, ids);
+    }
+  }
+
+  return {
+    ...prevSubscriberState,
+    subscriptionsById: newSubscriptionById,
+    subscriberIdsByTopic,
+    newTopicsBySubscriberId: new Map(newTopicsBySubscriberId),
```
jtbandes, line 80: seems like this has already been copied, no need to copy again?

packages/studio-base/src/components/MessagePipeline/usePlayerState.ts
```diff
@@ -0,0 +1,197 @@
...
+// Update state with a subscriber. Any new topics for the subscriber are tracked in newTopicsBySubscriberId
+// to receive the last message on their newly subscribed topics.
+function updateSubscriberAction(
+  prevSubscriberState: InternalState,
+  action: UpdateSubscriberAction,
+): InternalState {
+  const previousSubscriptionsById = prevSubscriberState.subscriptionsById;
+  const newTopicsBySubscriberId = prevSubscriberState.newTopicsBySubscriberId;
+
+  const previousSubscriptionById = previousSubscriptionsById.get(action.id);
```
jtbandes, line 44:
```suggestion
  const previousSubscription = previousSubscriptionsById.get(action.id);
```

packages/studio-base/src/components/MessagePipeline/usePlayerState.ts
```diff
@@ -0,0 +1,197 @@
...
+  } else if (previousSubscriptionById) {
+    const prevTopics = new Set(previousSubscriptionById.map((sub) => sub.topic));
+    for (const { topic: newTopic } of action.payloads) {
+      if (!prevTopics.has(newTopic)) {
+        newTopics.add(newTopic);
+      }
+    }
+  }
+
+  const newSubscriptionById = new Map(previousSubscriptionsById);
```
jtbandes, line 60:
```suggestion
  const newSubscriptionsById = new Map(previousSubscriptionsById);
```

packages/studio-base/src/components/MessagePipeline/usePlayerState.ts
```diff
@@ -0,0 +1,197 @@
...
+        if (!subscriberMessageEvents) {
+          subscriberMessageEvents = [];
+          messagesBySubscriberId.set(id, subscriberMessageEvents);
+        }
+        subscriberMessageEvents.push(messageEvent);
+      }
+    }
+  }
+
+  // Inject the last message on a topic to all new subscribers of the topic
```
jtbandes, line 129: Feels like we would want to do this in `updateSubscriberAction` immediately when the player subscribes and not wait for the player to emit more messages (or some other state change)?

jtbandes (approved):

packages/studio-base/src/components/MessagePipeline/MockMessagePipelineProvider.tsx
```diff
@@ -144,6 +145,33 @@ export default function MockMessagePipelineProvider(props: {
...
 
+  // In the real pipeline, the messageEventsBySubscriberId only change
+  // on player listener callback - not on subscriber changes
+  //
+  // In tests, the first setSubscriptions call happens after we've already set props.messages
+  // So we have some special logic to detect the _first_ change of subscriptions
+  // and update messageEventsBySubscriberId.
+
+  const allSubsRef = useRef<typeof allSubscriptions>(allSubscriptions);
+  const firstChangeRef = useRef<boolean>(false);
```
jtbandes: nit: TS will figure out the type if you don't pass it an annotation in <>

PR #2781, opened 2022-02-03
Create and select a default layout if no layout is active. 

### LINKED ISSUE #2611: New users are prompted to create a layout before the app is usable

foxymiles:
**User-Facing Changes**
This prevents users landing in a state with no layout.

**Description**
The idea here is to create and select a default layout when no layout is active.

Fixes #2611

### DIFF
--- a/packages/studio-base/src/providers/CurrentLayoutProvider/defaultLayout.ts
+++ b/packages/studio-base/src/providers/CurrentLayoutProvider/defaultLayout.ts
@@ -3,7 +3,12 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
 import { PanelsState } from "@foxglove/studio-base/index";
+import { defaultPlaybackConfig } from "@foxglove/studio-base/providers/CurrentLayoutProvider/reducers";
 
+/**
+ * This is loaded when the user has no layout selected on application launch
+ * to avoid presenting the user with a blank layout.
+ */
 export const defaultLayout: PanelsState = {
   configById: {
     "3D Panel!18i6zy7": {
@@ -15,10 +20,7 @@ export const defaultLayout: PanelsState = {
   globalVariables: {},
   userNodes: {},
   linkedGlobalVariables: [],
-  playbackConfig: {
-    messageOrder: "receiveTime",
-    speed: 1,
-  },
+  playbackConfig: { ...defaultPlaybackConfig },
   layout: {
     first: "3D Panel!18i6zy7",
     second: {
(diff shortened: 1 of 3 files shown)


amacneil (approved):
Cool, I like this behavior much better than current behavior

packages/studio-base/src/providers/CurrentLayoutProvider/defaultLayout.ts
```diff
@@ -0,0 +1,33 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { PanelsState } from "@foxglove/studio-base/index";
+
+export const defaultLayout: PanelsState = {
```
jtbandes, line 7: maybe comment what this is used for?

packages/studio-base/src/providers/CurrentLayoutProvider/defaultLayout.ts
```diff
@@ -0,0 +1,33 @@
...
+    },
+    "RawMessages!os6rgs": {},
+    "ImageViewPanel!3mnp456": {},
+  },
+  globalVariables: {},
+  userNodes: {},
+  linkedGlobalVariables: [],
+  playbackConfig: {
+    messageOrder: "receiveTime",
+    speed: 1,
```
amacneil, line 20: would be nice to figure out where in the app the default playback speed of `0.2` comes from and fix it there, so we don't have to duplicate it

jtbandes (approved):

packages/studio-base/src/providers/CurrentLayoutProvider/defaultLayout.ts
foxymiles: Good idea.

packages/studio-base/src/providers/CurrentLayoutProvider/defaultLayout.ts
foxymiles: Looks like we're using 1.0 now but @defunctzombie pointed me to our default config for this so I've changed it to reuse that definition.

PR #2785, opened 2022-02-04
Fix `findDOMNode` warning from CSSTransition

jtbandes:
**User-Facing Changes**
None

**Description**
Fixes a React warning by using the `nodeRef` property of CSSTransition which was added as an alternative to the default (deprecated) findDOMNode behavior: https://github.com/reactjs/react-transition-group/releases/tag/v4.4.0

Also improves the animation by following guidelines from http://reactcommunity.org/react-transition-group/css-transition about only putting the `transition` property on the "active" classes.


defunctzombie (approved):
miracle

Issue #2780, opened 2022-02-03
Node Playground error when loading data platform link with specific layout


jtbandes:
When visiting [this link](https://studio.foxglove.dev/?ds=foxglove-data-platform&ds.deviceId=dev_mHH1Cp4gPybCPR8y&ds.end=2018-07-24T03%3A29%3A06.797000000Z&ds.start=2018-07-24T03%3A28%3A47.647951000Z&layoutId=96418452-e410-4062-8f9a-395453e6ddf3&time=2018-07-24T03%3A28%3A47.647951000Z), if another tab is already open on the same link, Studio throws an error:

<img width="723" alt="image" src="https://user-images.githubusercontent.com/14237/152402672-f8ca32f5-ee7a-496f-ab4b-fa0a164aea29.png">

In the NP shared worker, the following errors are visible:

<img width="853" alt="image" src="https://user-images.githubusercontent.com/14237/152402731-39218b57-c90a-4ad7-b1a6-4a64de20125a.png">

PR #2786, opened 2022-02-04
Fix same nodeTransformerWorker being used across tabs

jtbandes:
**User-Facing Changes**
Fixed a `no receiver registered for generateRosLib` error when using Node Playground in multiple tabs/windows at the same time.

**Description**
Fixes https://github.com/foxglove/studio/issues/2780. Basically the same fix as https://github.com/foxglove/studio/pull/2216, but for the nodeTransformerWorker: these workers were not designed to be actually shared between tabs (e.g. they do various logic in `onconnect` assuming that it'll only happen once, and they may be `close()`d by one tab when another could still be using it). Both of these have unique ids in upstream webviz and we accidentally removed them, but I only replaced one in #2216.


defunctzombie (approved):
Fascinating. I thought shared worker was about sharing data between things on the page rather than across tabs. This has opened my world to the crazy things we could do.

PR #2788, opened 2022-02-04
Fix useLatestMessageDataItem to keep last message

defunctzombie:
**User-Facing Changes**
Raw Message panel more reliably shows the latest message on startup for messages on latching topics.

**Description**
Previously useLatestMessageDataItem would lose the latest message on a topic when the queriedData was undefined (invalid path). The path is invalid until topics and datatypes are updated which happened after addMessages was invoked in useLatestMessageDataItem.

This change updates addMessages to keep the latest message event even when the path is invalid. When restore is called, the path becomes valid and the last message can be checked against the path.

### DIFF
--- a/packages/studio-base/src/components/MessagePathSyntax/useLatestMessageDataItem.test.tsx
+++ b/packages/studio-base/src/components/MessagePathSyntax/useLatestMessageDataItem.test.tsx
@@ -145,7 +145,7 @@ describe("useLatestMessageDataItem", () => {
     ]);
   });
 
-  it("updates when topics and datatypes becomes available", async () => {
+  it("restores previously received message when topics and datatypes becomes available", async () => {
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/components/MessagePathSyntax/useLatestMessageDataItem.test.tsx
```diff
@@ -144,4 +144,37 @@ describe("useLatestMessageDataItem", () => {
       },
     ]);
   });
+
+  it("updates when topics and datatypes becomes available", async () => {
```
jtbandes, line 148:
```suggestion
  it("restores previously received message when topics and datatypes becomes available", async () => {
```

jtbandes (approved):

Issue #2777, opened 2022-02-02
Schema/datatype names with dots don't work in Node Playground


jtbandes:
**Description**
When a schema name has dots (e.g. `example_pkg.Example`) the Node Playground lib generation produces this, which is invalid TS code:
```ts
    export interface example_pkg.stuff.ExampleMsg {
        msg: string;
        count: number;
    }
```

**Steps To Reproduce**
- Run the [protobuf example server](https://github.com/foxglove/ws-protocol/tree/main/python) but with the `ExampleMsg.proto` file modified to include `package example_pkg.stuff;` (re-running `protoc` to regenerate the python code and the `.bin`)
- Create a Node Playground node and try to compile it

**Expected Behavior**
Node Playground is usable

**Actual Behavior**
Importing from `"ros"` generates TS errors due to invalid lib code.

<img width="606" alt="image" src="https://user-images.githubusercontent.com/14237/152252970-22fd7be1-6376-446c-bbc4-214fa25ee6fd.png">

**Pointer to relevant code**
https://github.com/foxglove/studio/blob/514acda846cfdd7d0528fbeed84ab57a23f9eb3f/packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/generateRosLib.ts#L133

PR #2790, opened 2022-02-04
Add `./types` lib to node playground

defunctzombie:
**User-Facing Changes**
Add a new `./types` utility file for use in node playground scripts. This utility file is generated from the currently loaded data source and contains type definitions for the datatypes present in the data source.

Here's an example node playground script using the new data source utility file to access the types generated from the loaded data source.

```typescript

type Output = Message<"std_msgs/Header">;
type GlobalVariables = { id: number };

export const output = "/studio_node/gps-header";

const publisher = (
  msgEvent: Input<"/gps">,
  globalVars: GlobalVariables
): Output => {
  return msgEvent.message.header;
};

export default publisher;
```

**Description**

This new utility file differs from the existing `ros` utility module by using a different syntax to lookup datatypes. The `ros` utility exposed types on a `Messages` namespace with interfaces name like `std_msgs__Header`. These interface names were conversions of the `std_msg/Header` datatype into a valid typename for typescript. Now that we support datasources with `.` (protobuf package names), the interface name code was trying to generate a `std_msgs.Header` typename which is not valid.

Rather than trying to figure out all possible datatype characters, the new `Message` type follows the pattern of `Input` and lets the user specify a schema name as the generic type. This allows users to write `Message<"std_msgs.Header">` to match the datatype name they expect.

We additionally expose `Input`, similar to before, to allow for topic based lookup for the input message event.

Fixes: #2777


packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/generateStudioLib.ts
```diff
@@ -0,0 +1,122 @@
...
+  src += "\n}\n";
+
+  return src;
+};
+
+function generateTypesByTopicInterface(topics: Topic[]): string {
+  let src = "declare interface TypesByTopic {";
+
+  for (const topic of topics) {
+    src += `"${topic.name}": MessageEvent<Types["${topic.datatype}"]>;`;
```
jtbandes, line 87:
```suggestion
    src += `${JSON.stringify(topic.name)}: MessageEvent<Types[${JSON.stringify(topic.datatype)}]>;`;
```

packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/generateStudioLib.ts
```diff
@@ -0,0 +1,122 @@
...
+
+      if (isConstant === true) {
+        src += `\n // ${field.name} = ${field.valueText}`;
+      } else if (isArray === true) {
+        if (typedArray) {
+          src += `\n${field.name}: ${typedArray},`;
+        } else if (rosPrimitive) {
+          src += `\n${field.name}: ${rosPrimitive}[],`;
+        } else {
+          src += `\n${field.name}: Types["${type}"][],`;
```
jtbandes, line 64:
```suggestion
          src += `\n${field.name}: Types[${JSON.stringify(type)}][],`;
```

packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/generateStudioLib.ts
```diff
@@ -0,0 +1,122 @@
...
+        }
+      } else {
+        if (rosPrimitive) {
+          src += `\n${field.name}: ${rosPrimitive},`;
+        } else {
+          src += `\n${field.name}: Types["${type}"],`;
```
jtbandes, line 70:
```suggestion
          src += `\n${field.name}: Types[${JSON.stringify(type)}],`;
```

packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/generateStudioLib.ts
jtbandes: Probably should do this for all the `field.name` too. What if someone decides to put a `.` in their field name?

packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/generateStudioLib.ts
```diff
@@ -0,0 +1,122 @@
...
+  ["string", "string"],
+  ["bool", "boolean"],
+  ["time", "Time"],
+  ["duration", "Duration"],
+]);
+
+export const typedArrayMap = new Map<string, string>([
+  ["uint8", "Uint8Array"],
+  ["int8", "Int8Array"],
+]);
```
jtbandes, line 34: Do we need more entries here?

https://github.com/foxglove/rosmsg-serialization/blob/d1f3d3993edeecc0c0621a8d615c322d822580a8/src/BuiltinDeserialize.ts#L103

packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/generateStudioLib.ts
```diff
@@ -0,0 +1,122 @@
...
+
+  const src = `
+type Time = {
+  sec: number;
+  nsec: number;
+}
+
+type Duration = Time;
+
+declare interface MessageEvent<T> {
```
jtbandes, line 106:
```suggestion
interface MessageEvent<T> {
```

packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/generateStudioLib.ts
```diff
@@ -0,0 +1,122 @@
...
+  topic: T;
+  receiveTime: Time;
+  message: T;
```
jtbandes, line 109: topic and message can't both be `T` can they?

defunctzombie:
I'm not updating the default node template just yet. There is follow-up work to make Output types work better with DataSourceType. The current code to detect datatype names is tied to the `ros` library and `Messages` from the ros library. I want to move to deprecate the "ros" library and provide the builtin message types separate from the DataSourceTypes so users can output the pre-defined message types for display in 3d panel, etc.

packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/transform.ts
```diff
@@ -269,6 +270,8 @@ export const compile = (nodeData: NodeData): NodeData => {
   const sourceCodeMap = new Map<string, string>();
   sourceCodeMap.set(nodeFileName, sourceCode);
   sourceCodeMap.set(projectConfig.rosLib.filePath, rosLib);
+  sourceCodeMap.set("/studio_node/data-source.ts", dataSourceLib);
```
jtbandes, line 273: What's this `studio_node` dir for? Why is this hard coded while the `projectConfig.rosLib.filePath` is not?

packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/generateDataSourceLib.ts
```diff
@@ -0,0 +1,172 @@
...
+  return src;
+}
+
+let emptyLib: string | undefined;
+function generateEmptyDataSourceLib(): string {
+  if (emptyLib) {
+    return emptyLib;
+  }
+
+  return (emptyLib = generateDataSourceLib({ topics: [], datatypes: new Map() }));
```
jtbandes, line 169: `??=` is your friend

packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/generateDataSourceLib.ts
```diff
@@ -0,0 +1,172 @@
...
+     * MessageEventByTopic enumerates the MessageEvent types for all the topics in
+     * the current data source.
+     *
+     * Use subtype notation to access the MessageEvent type by topic name:
+     *
+     * \`\`\`
+     * type MsgEventFoo = MessageEventByTopic["/foo"];
+     * \`\`\`
+     */
+    export type MessageEventByTopic = {`;
```
jtbandes, line 122: We should try to be consistent with pluralization between DataSourceTypes & here.

```suggestion
    export type MessageEventsByTopic = {`;
```

packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/generateDataSourceLib.ts
```diff
@@ -0,0 +1,172 @@
...
+
+export type Duration = Time;
+
+/**
+ * MessageEvent contains a single message. The event has the topic, and the receive time.
+ */
+export interface MessageEvent<M, T = string> {
+  topic: T;
+  receiveTime: Time;
+  message: M;
```
jtbandes, line 152: Probably would be nice to add per-field doc comments, because they will show up in tooltips

packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/generateDataSourceLib.ts
```diff
@@ -0,0 +1,172 @@
...
+function generateTypesByTopicInterface(topics: Topic[]): string {
+  let src = `
+    /**
+     * MessageEventByTopic enumerates the MessageEvent types for all the topics in
+     * the current data source.
+     *
+     * Use subtype notation to access the MessageEvent type by topic name:
+     *
+     * \`\`\`
+     * type MsgEventFoo = MessageEventByTopic["/foo"];
```
jtbandes, line 119: this example seems a little contrived, maybe use it in a variable context, like `const point: MessageEventsByTopic["/points"] = {...}` ?

jtbandes (approved):

packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/transform.ts
defunctzombie: rosLib.filePath already has it iirc (or uses a different path).

The `/studio_node` folder is where all the node code "lives" from the perspective of the compiler. To make `import ./data-source` work, the data-source.ts needs to be in that folder. The utility files also get "put" in that folder. I don't see a reason for why we need to have this `/studio_node` folder vs having things at "root" but it wasn't something I wanted to change here.

I do think it would be good to move this into the projectConfig because Editor.tsx also needs the path iirc.

packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/generateDataSourceLib.ts
defunctzombie: I would probably change `DataSourceTypes` to `DataSourceType` then. The type from `MessageEventByTopic` is a single MessageEvent not an array. Maybe thats not a distinction that matters?

amacneil:
```typescript

export const inputs = ["/input_topic"];
export const output = "/studio_node/output_topic";

type Output = Message<"visualization_msgs/Marker">;
type GlobalVariables = { id: number };

// Populate 'Input' with a parameter to properly type your inputs, e.g. 'Input<"/your_input_topic">'
export default function process(
  input: Input<"/input_topic">,
  globalVars: GlobalVariables
) {
  return {};
}
```

packages/studio-base/src/players/UserNodePlayer/MemoizedLibGenerator.ts
```diff
@@ -0,0 +1,47 @@
...
+  private datatypes?: RosDatatypes;
+  private topics?: Topic[];
+  private fn: LibGeneratorFn;
+  private cached?: string;
+
+  constructor(fn: LibGeneratorFn) {
+    this.fn = fn;
+  }
+
+  async update(args: Args): Promise<[boolean, string]> {
```
jtbandes, line 30: this feels like a "naked" unlabeled boolean of the same type we try to prevent with our no-boolean-parameters rule... maybe put them in an object instead?

packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/generateTypesLib.ts
```diff
@@ -0,0 +1,154 @@
...
+  return src;
+}
+
+function generateTypesLib(args: Args): string {
+  const typesByTopic = generateTypesByTopicInterface(args.topics);
+  const typesBySchemaName = generateTypesInterface(args.datatypes);
+
+  const src = `
+// NOTE:
+// This file is generated from the current data source.
```
jtbandes, line 129: could include the data source name here?

packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/generateTypesLib.test.ts
```diff
@@ -0,0 +1,98 @@
...
+  projectCode: new Map<string, string>(),
+};
+
+describe("generateTypesLib", () => {
+  describe("generateDatatypesInterface", () => {
+    it("should generate basic Datatypes interface", async () => {
+      const src = generateTypesInterface(
+        new Map(
+          Object.entries({
+            "std_msgs.ColorRGBA": {
```
jtbandes, line 34: Test a name with a double quote and/or a backslash in it?

packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/transform.ts
```diff
@@ -269,6 +271,8 @@ export const compile = (nodeData: NodeData): NodeData => {
   const sourceCodeMap = new Map<string, string>();
   sourceCodeMap.set(nodeFileName, sourceCode);
   sourceCodeMap.set(projectConfig.rosLib.filePath, rosLib);
+  sourceCodeMap.set("/studio_node/generatedTypes.ts", typesLib ? typesLib : generatedTypesLibSrc);
```
jtbandes, line 274: I still find it weird that one of these paths is hard-coded and the others are not.

packages/studio-base/src/panels/NodePlayground/Editor.tsx
```diff
@@ -93,6 +95,16 @@ const Editor = ({
     };
   }, [rosLib]);
 
+  React.useEffect(() => {
+    const filePath = monacoApi.Uri.parse(`file:///studio_node/generatedTypes.ts`);
+    const model =
+      monacoApi.editor.getModel(filePath) ??
+      monacoApi.editor.createModel(typesLib, "typescript", filePath);
```
jtbandes, line 102: I see that this is where we tell monaco about `generatedTypes.ts`. Where do we tell it about `types.ts`?

jtbandes (approved):
I'm not sure about the name `./types` — if we think we will add a feature where user nodes can have more than one file, it feels like the `./` should be reserved for user imports. If we pick `./types` now, then when we add that feature we'll either force users to rewrite their imports, or we will have to prevent them from creating a file called `types.ts`...

packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/transform.ts
defunctzombie: I don't know yet the direction I am gonna go with this. I might change all of them to not be hard-coded.

packages/studio-base/src/panels/NodePlayground/Editor.tsx
defunctzombie: We already had `./types` as a util file. I added to the existing file.

packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/generateTypesLib.ts
defunctzombie: I don't think we plumb that data through currently.

defunctzombie:
We already had a `./types` utility file. I leveraged that file to add `Input` and `Message` types to it and import from `./generatedTypes`.

PR #2799, opened 2022-02-07
Disable tracking of out-of-order messages, which leaked memory

jtbandes:
**User-Facing Changes**
Fixed a memory leak when the app received many out-of-order messages due to the presence of the `/clock` topic.

**Description**
Out of order messages were being logged to the dev console. The memory debugger showed the console itself was responsible for retaining memory. So we disable logging of out of order messages unless explicitly enabled for debugging purposes.

This is a follow up to https://github.com/foxglove/studio/pull/2185. That fix prevented all the messages from being held in memory by the VM, but they were still held by the browser console when logged.

image

### DIFF
--- a/packages/studio-base/src/components/MessagePipeline/MessageOrderTracker.ts
+++ b/packages/studio-base/src/components/MessagePipeline/MessageOrderTracker.ts
@@ -35,6 +35,7 @@ class MessageOrderTracker {
   private lastMessageTopic?: string;
   private lastLastSeekTime?: number;
   private warningTimeout?: ReturnType<typeof setTimeout>;
+  private trackIncorrectMessages = false; // set this to true to debug out of order messages
   private incorrectMessages: MessageEvent<unknown>[] = [];
 
   update(playerState: PlayerState): void {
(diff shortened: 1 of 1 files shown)


foxymiles (approved):
Good catch!

defunctzombie:
Should this be an app setting? In what situation would we set this to true in the code and then debug the out of order messages?

jtbandes:
> In what situation would we set this to true in the code and then debug the out of order messages?

Probably someone who is writing a custom player / data source would do this. Or us if we find a bug with one of ours.

packages/studio-base/src/components/MessagePipeline/MessageOrderTracker.ts
```diff
@@ -35,6 +35,7 @@ class MessageOrderTracker {
   private lastMessageTopic?: string;
   private lastLastSeekTime?: number;
   private warningTimeout?: ReturnType<typeof setTimeout>;
+  private trackIncorrectMessages = false; // set this to true to debug out of order messages
```
defunctzombie, line 38: Add a comment on why this should not be true for production so it's more obvious why this flag exists. And a note on memory use.

defunctzombie (approved):
Improve the comment cause it has some nuances to turn this on.

PR #2802, opened 2022-02-08
Replace fluentui mergeStyleSets, mergeStyles and makeStyles to MUI makeStyles

2metres:
**Description**
Refactor low hanging fluentui makeStyles, mergeStyles and mergeStyleSets to use mui makeStyles

### DIFF
--- a/packages/studio-base/src/components/ErrorDisplay.tsx
+++ b/packages/studio-base/src/components/ErrorDisplay.tsx
@@ -3,39 +3,41 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
 import { Theme, Typography, Link, Stack, Divider } from "@mui/material";
-import { makeStyles } from "@mui/styles";
+import { createStyles, makeStyles } from "@mui/styles";
 import { ErrorInfo, useMemo, useState } from "react";
 
-const useStyles = makeStyles((theme: Theme) => ({
-  root: {
-    display: "flex",
-    flexDirection: "column",
-    height: "100%",
-    padding: theme.spacing(2),
-  },
-  alertContainer: {
-    display: "flex",
-    flexDirection: "column",
-    flexGrow: 1,
-    overflow: "hidden",
-  },
-  errorDetailHeader: {
-    fontWeight: "bold",
-  },
-  errorDetailStack: {
-    fontSize: theme.typography.body2.fontSize,
-    lineHeight: "1.3em",
-    paddingLeft: theme.spacing(2),
-  },
-  errorDetailContainer: {
-    flexGrow: 2,
-    overflowY: "auto",
-  },
-  actions: {
-    paddingTop: theme.spacing(2),
-    textAlign: "right",
-  },
-}));
+const useStyles = makeStyles((theme: Theme) =>
+  createStyles({
+    root: {
+      display: "flex",
+      flexDirection: "column",
+      height: "100%",
+      padding: theme.spacing(2),
+    },
+    alertContainer: {
+      display: "flex",
+      flexDirection: "column",
+      flexGrow: 1,
+      overflow: "hidden",
+    },
+    errorDetailHeader: {
+      fontWeight: "bold",
+    },
+    errorDetailStack: {
+      fontSize: theme.typography.body2.fontSize,
+      lineHeight: "1.3em",
+      paddingLeft: theme.spacing(2),
+    },
+    errorDetailContainer: {
+      flexGrow: 2,
+      overflowY: "auto",
+    },
+    actions: {
+      paddingTop: theme.spacing(2),
+      textAlign: "right",
+    },
+  }),
+);
 
 function ErrorStacktrace({ stack }: { stack: string }) {
   const styles = useStyles();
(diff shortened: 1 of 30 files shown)


packages/studio-base/src/components/ErrorDisplay.tsx
```diff
@@ -3,39 +3,41 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
 import { Theme, Typography, Link, Stack, Divider } from "@mui/material";
-import { makeStyles } from "@mui/styles";
+import { createStyles, makeStyles } from "@mui/styles";
```
jtbandes, line 6: What's the purpose of this?

> This function doesn't really "do anything" at runtime, it's just the identity function. Its only purpose is to defeat TypeScript's type widening when providing style rules to makeStyles/withStyles which are a function of the Theme.

Seems like it should only be added when there are type errors?

packages/studio-base/src/panels/ThreeDimensionalViz/DebugStats.tsx
```diff
@@ -11,7 +11,9 @@
 //   found at http://www.apache.org/licenses/LICENSE-2.0
 //   You may not use this file except in compliance with the License.
 
-import { makeStyles } from "@fluentui/react";
+// import { makeStyles } from "@fluentui/react";
```
jtbandes, line 14: leftover

packages/studio-base/src/typings/fluentui.d.ts
```diff
@@ -146,3 +147,8 @@ declare module "@fluentui/react" {
     styles?: IStyleFunctionOrObject<IIconStyleProps, IIconStyles>;
   }
 }
+
+// TODO: Move this into a mui.d.ts file
```
jtbandes, line 151: move before merging?

jtbandes (changes requested):
Looks like some contrast got worse (at least in light mode) on 3D view button:

image


This one seems like a real problem since the notification dot went away:

image

packages/studio-base/src/panels/diagnostics/DiagnosticStatus.tsx
```diff
@@ -171,6 +83,98 @@ function sanitize(value: string): { __html: string } {
...
+
+    collapsedSection: {
+      textAlign: "center",
+      color: theme.palette.error.main,
+    },
+    interactiveRow: {
+      cursor: "pointer",
+
+      "&:nth-child(odd)": {
+        backgroundColor: theme.palette.grey[50],
```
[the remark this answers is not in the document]
2metres: The idea with `grey` (at least IMO) is that is a consistent decay of neutral colors across themes, not nessicarily consistent like `common.white/black`

packages/studio-base/src/typings/fluentui.d.ts
2metres: What is the purpose of this? From what I can tell in the `makeStyles` usage it's not working.

2metres: I've copied it from here: https://mui.com/guides/migration-v4/#heading-mui-styles

packages/studio-base/src/panels/diagnostics/DiagnosticStatus.tsx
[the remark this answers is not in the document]
2metres: I managed to get the badge back there are some subtle differences between MUI and Fluent makeStyles mainly around psuedo and nested selectors.

```diff
const useStyles = makeStyles({
  root: {
-    ":hover": {
+    "&:hover": {
     }
-   ".child-selector": {
+   "& .child-selector": {
    }
  }
})
```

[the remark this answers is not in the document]
2metres: @jtbandes I can't see the contrast difference you're referrring to, can you provide a bit more detail?

defunctzombie (approved):

packages/studio-base/src/typings/mui.d.ts
```diff
@@ -0,0 +1,9 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { Theme } from "@mui/material/styles";
+
+declare module "@mui/styles/defaultTheme" {
```
jtbandes, line 7: maybe drop a comment with link to documentation on what this is for? It seems odd to me that we would import something from mui, and modify something from mui to re-export the thing that was already from mui... With the other theme libraries (styled-components) we've done overrides like this when we wanted to inject our custom theme object that the library was not already aware of. But maybe this is normal in mui?

jtbandes (approved):

packages/studio-base/src/typings/mui.d.ts
2metres: Just dove into this with @defunctzombie I don't think it really does anything. Removing for now.

PR #2814, opened 2022-02-09
Simplify findReturnType in node playground ast.ts

defunctzombie:
**User-Facing Changes**
None

**Description**
Rather than writing our own AST traversal to find the return type, leverage getTypeAtLocation from the TypeChecker to fully resolve the type. This provides wider syntax support for resolving valid types.

This change is required to support the new syntax used by https://github.com/foxglove/studio/pull/2790.

### DIFF
--- a/packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/transform.test.ts
+++ b/packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/transform.test.ts
@@ -1413,7 +1413,7 @@ describe("pipeline", () => {
           export default (msg: any): MyAny => {
             return () => 1;
           };`,
-        error: ErrorCodes.DatatypeExtraction.NO_TYPEOF,
+        error: ErrorCodes.DatatypeExtraction.NO_FUNCTIONS,
       },
       {
         description: "Nested typeof func",
@@ -1482,7 +1482,7 @@ describe("pipeline", () => {
           export default (msg: any): Pos[keyof Pos] => {
             return { pos: { x: 1, y: 2 } };
           };`,
-        error: ErrorCodes.DatatypeExtraction.INVALID_INDEXED_ACCESS,
+        error: ErrorCodes.DatatypeExtraction.LIMITED_UNIONS,
       },
       {
         description: "Indexed access type with non-string index",
@@ -1491,7 +1491,7 @@ describe("pipeline", () => {
           export default (msg: any): Pos[3] => {
             throw new Error();
           };`,
-        error: ErrorCodes.DatatypeExtraction.INVALID_INDEXED_ACCESS,
+        error: ErrorCodes.DatatypeExtraction.BAD_TYPE_RETURN,
       },
     ];
 
--- a/packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/transform.test.ts
+++ b/packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/transform.test.ts
@@ -1155,6 +1155,81 @@ describe("pipeline", () => {
         datatypes: baseDatatypes,
         outputDatatype: "std_msgs/ColorRGBA",
       },
+      {
+        description: "Should handle deep subtype lookup",
+        sourceCode: `
+          import { Input } from "ros";
+
+          type MessageBySchemaName = {
+            "std_msgs/Header": {
+              frame_id: string;
+              seq: number;
+              stamp: {
+                sec: number,
+                nsec: number,
+              };
+            };
+            "pkg/Custom": {
+              header: MessageBySchemaName["std_msgs/Header"];
+            };
+          };
+
+          type Message<T extends keyof MessageBySchemaName> = MessageBySchemaName[T];
+
+          export const inputs = ["/some_topic"];
+          export const output = "/studio_node/sample";
+
+          type PoseStamped = Message<"pkg/Custom">;
+          type Output = PoseStamped;
+
+          const publisher = (message: Input<"/some_topic">): Output => {
+            return {
+              header: { frame_id: "foo", seq: 0, stamp: { sec: 0, nsec: 0 } },
+            };
+          };
+          export default publisher;`,
+        datatypes: new Map(
+          Object.entries({
+            "/studio_node/main": {
+              definitions: [
+                {
+                  arrayLength: undefined,
+                  isArray: false,
+                  isComplex: true,
+                  name: "header",
+                  type: "/studio_node/main/header",
+                },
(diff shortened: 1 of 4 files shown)


packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/transform.test.ts
```diff
@@ -1482,7 +1557,7 @@ describe("pipeline", () => {
           export default (msg: any): Pos[keyof Pos] => {
             return { pos: { x: 1, y: 2 } };
           };`,
-        error: ErrorCodes.DatatypeExtraction.INVALID_INDEXED_ACCESS,
+        error: ErrorCodes.DatatypeExtraction.LIMITED_UNIONS,
```
jtbandes, line 1560: Test seems to have changed meaning, should it be renamed? Why is there a limited unions error here?

packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/transform.test.ts
```diff
@@ -1491,7 +1566,7 @@ describe("pipeline", () => {
           export default (msg: any): Pos[3] => {
             throw new Error();
           };`,
-        error: ErrorCodes.DatatypeExtraction.INVALID_INDEXED_ACCESS,
+        error: ErrorCodes.DatatypeExtraction.BAD_TYPE_RETURN,
```
jtbandes, line 1569: Rename or fix test?

packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/typescript/ast.ts
```diff
@@ -190,154 +189,70 @@ export const findDefaultExportFunction = (
...
-    }
+  // If there are multiple declarations for the symbol, filter down to only
+  // the interface ones. However, if there is only one, use that one to provide better errors
+  let declaration: ts.Declaration | undefined;
+  if (symbol.declarations.length === 1) {
+    declaration = symbol.declarations[0];
+  } else {
+    declaration = symbol.declarations.filter(
+      (decl) => decl.kind === ts.SyntaxKind.InterfaceDeclaration,
+    )[0];
```
jtbandes, line 232: does this mean there will be an error if there is no InterfaceDeclaration but there are multiple other declarations? (not sure exactly what that situation would be)

packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/transform.test.ts
defunctzombie: There's a limited unions error now because typescript expands `Pos[keyof Pos]` into `Pos["position"] | Pos["lala"]`

packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/typescript/ast.ts
```diff
@@ -190,154 +189,70 @@ export const findDefaultExportFunction = (
...
-      throw new DatatypeExtractionError(noTypeOfError);
+  if (!declaration) {
+    throw new DatatypeExtractionError(badTypeReturnError);
+  }
 
+  switch (declaration.kind) {
+    case ts.SyntaxKind.InterfaceDeclaration:
+      return declaration as ts.InterfaceDeclaration;
+    case ts.SyntaxKind.TypeLiteral:
+      return declaration as ts.TypeLiteralNode;
```
jtbandes, line 243: There are `isInterfaceDeclaration` and `isTypeLiteralNode` functions you can use to do this without casting

packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/transform.test.ts
defunctzombie: The test is correct. It shows failing when accessing with an integer and getting a bad return type. I think its fine.

packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/typescript/ast.ts
```diff
@@ -609,8 +513,17 @@ export const constructDatatypes = (
       case ts.SyntaxKind.AnyKeyword:
         throw new DatatypeExtractionError(noNestedAny);
 
-      default:
-        throw new Error(`Unhandled node kind (${tsNode.kind}) for field (${name})`);
+      default: {
+        const locationType = checker.getTypeAtLocation(tsNode);
```
jtbandes, line 517: Can't we just always do this? Why do we still need all the other cases in this switch?

packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/typescript/ast.ts
defunctzombie: This happens when you have the same type name as a variable name.

```
const Nested = "foo";
type Nested = { num: number };
```

Both of those are the same "symbol" and become declarations in the symbol.

packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/typescript/ast.ts
defunctzombie: There is not.

packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/typescript/ast.ts
defunctzombie: I tried - and failed. I welcome you to try with me if you'd like to experience it.

packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/typescript/ast.ts
jtbandes: Wut

jtbandes: image

defunctzombie: oh interesting - I was looking on the `node` instance cause that also has some `is` checks.

defunctzombie: done. It did mean I had to use `if` rather than the switch statement. I found the switch statement easier to read but ok.

jtbandes (approved):

Issue #2805, opened 2022-02-08
Alpha support for point clouds


joshabb:
It is necessary for point clouds to have alpha support as this is another way of visually encoding data. Can we please make add support for alpha coloring in point clouds?

amacneil:
@defunctzombie is this something we can include in https://github.com/foxglove/studio/issues/1482?

jtbandes:
Keep in mind that even if alpha is passed through to the GPU, the blending will probably not work correctly since correct transparency depends on the z-order, but we don't sort objects (especially points inside point clouds) by distance from camera before rendering them.

PR #2815, opened 2022-02-09
Add alpha support and byte order settings for RGB point clouds

jtbandes:
**User-Facing Changes**
Added support for rendering `rgba` fields of point cloud messages with embedded alpha.
Changed interpretation of `rgb`/`rgba` fields by default to use RGBA byte order. Added topic settings to select between ABGR (Studio/Webviz historical behavior) and BGRA (RViz/PCL behavior).

**Description**
Closes https://github.com/foxglove/studio/issues/2805

Screen Shot 2022-02-09 at 4 14 12 PM

Screen Shot 2022-02-09 at 4 16 50 PM


defunctzombie (approved):

jtbandes:
I think if we designed this feature from scratch we might want to use allow the field names to be `rgba`, `bgra`, etc. and just use that to determine the byte order. However, @jhurliman also pointed out that it can be useful to quickly switch between byte orders in the UI as a debugging technique and to help users get their data looking right as easily as possible (without code changes).

defunctzombie:
The field name does indicate the byte order. However its "easy" to muck it up and end up with data you still want to look at so these options exist as overrides incase you discovered that whoever write rgba data didn't understand byte order (looking at you PCL).

PR #2816, opened 2022-02-10
Fix hover-on-zero issue for Plot panel

esthersweon:
**User-Facing Changes**
- Fixed an issue where the Plot panel legend's format would be incorrect when hovering on a point without a valid current value (when `Show plot values in legend` setting is toggled on)

BEFORE:
Screen Shot 2022-02-09 at 7 29 20 PM

AFTER:
Screen Shot 2022-02-09 at 7 29 29 PM

### DIFF
--- a/packages/studio-base/src/panels/Plot/PlotLegendRow.tsx
+++ b/packages/studio-base/src/panels/Plot/PlotLegendRow.tsx
@@ -4,7 +4,8 @@
 
 import { useTheme as useFluentUITheme } from "@fluentui/react";
 import { Close as CloseIcon, Error as ErrorIcon, Remove as RemoveIcon } from "@mui/icons-material";
-import { Box, IconButton, Stack, Tooltip, Typography } from "@mui/material";
+import { IconButton, Theme, Tooltip, Typography } from "@mui/material";
+import { createStyles, makeStyles } from "@mui/styles";
 import { ComponentProps, useCallback, useMemo, useState } from "react";
 import { v4 as uuidv4 } from "uuid";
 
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/panels/Plot/PlotLegendRow.tsx
```diff
@@ -183,57 +183,58 @@ export default function PlotLegendRow({
...
             variant="body2"
             align="right"
             sx={{ color: currentDisplay.color, width: 150 }}
           >
-            {currentDisplay.value}
+            {currentDisplay.value ?? ""}
           </Typography>
         </Box>
       )}
+      <Box>
```
defunctzombie, line 198: @2metres where did we land on using `<Box>` without any args vs `<div>`?

2metres: Based on our recent discoveries about the sx prop both `Box` and `Stack` here should be refactored to use `makeStyles` instead. @esthersweon I can take this on.

packages/studio-base/src/panels/Plot/PlotLegendRow.tsx
```diff
@@ -105,39 +176,12 @@ export default function PlotLegendRow({
...
-          padding: 0.25,
-          position: "sticky",
-          left: 0,
-          // creates an opaque background for the sticky element
-          backgroundImage: `linear-gradient(${palette.background.paper}, ${palette.background.paper})`,
-          backgroundBlendMode: "overlay",
-        })}
-      >
+    <div className={classes.root}>
+      <div className={classes.listItcon}>
```
esthersweon, line 180:
```suggestion
      <div className={classes.listIcon}>
```

2metres (approved):

Issue #2809, opened 2022-02-09
Plot Panel: Adding new topic erase past data of other topics


clementchef:
**Description**
Adding new topic erase past data of other topics
![Peek 2022-02-08 14-39](https://user-images.githubusercontent.com/66749666/153096224-6e5459a2-0de6-4141-bac1-cb19630bb9dd.gif)

- Version: 0.25.1
- OS: Ubuntu 18.04

**Steps To Reproduce**
Live play with rosbag play. Plot one topic. Add a second topic after a while. 

**Current Experience**

**Desired Experience**

defunctzombie:
Confirming I am able to reproduce this with v0.29 and main.

PR #2822, opened 2022-02-10
Fix restore callback in Plot panel

defunctzombie:
**User-Facing Changes**
Adding new series in the plot panel does not clear data from the current series.

**Description**
The restore callback was ignoring the previous value. By not using the previous value, the plot panel would clear when changing series information. This change keeps the plot panel data when changing series.

fixes: #2809


foxymiles (approved):
Nice simple fix. 👍

packages/studio-base/src/panels/Plot/index.tsx
```diff
@@ -263,8 +263,8 @@ function Plot(props: Props) {
     return out;
   }, [allPaths]);
 
-  const restore = useCallback((): PlotDataByPath => {
-    return {};
+  const restore = useCallback((previous?: PlotDataByPath): PlotDataByPath => {
+    return previous ?? {};
```
jtbandes, line 267: You should probably filter out data for paths that are no longer included in `allPaths`. Otherwise I think the panel will accumulate data forever even as the user removes series from the UI.

Something I believe the plot panel used to support is changing the path for an existing topic, e.g. from `/foo.bar` to `/foo.baz`, without dropping old `/foo` messages. This can be achieved by holding the original messages by topic in the reducer state as well as the queried data by path. It looks like that has been removed so perhaps it was intentional, but just something to be aware of (and maybe worth comparing with webviz).

defunctzombie: The plot panel is full of traps. I will update restore to drop paths no longer present. I doubt it will be a problem in practice since the user is probably changing paths just a few times if that much but since we know about this I'll include that in this change.

I'll look at the topic message state idea but if it proves to be too much hassle I'll keep it our of this bugfix to focus on fixing the bug issue.

defunctzombie: @jtbandes updated with filtering logic on restore

PR #2828, opened 2022-02-11
Add MCAP v0 local file support

jtbandes:
- Refactor commonly reused channel/schema parsing and decompression functions into `parseChannel` into new `@foxglove/mcap-support` package
- Fix check for missing schemas to support schema id 0
- Add `Mcap0StreamedDataProvider` & `Mcap0IndexedDataProvider`
- Add `McapDataProvider` which detects the mcap version and delegates to `Macp0StreamedDataProvider`, `Mcap0IndexedDataProvider `, or `McapPre0DataProvider`

PR #2829, opened 2022-02-11
upgrade quicklook to latest mcap commit

jtbandes:
**User-Facing Changes**
None

**Description**
Minor upgrades from field renames.


jtbandes:
Will just do this in https://github.com/foxglove/studio/pull/2828

Issue #2784, opened 2022-02-04
Diff mode does not diff filtered messages correctly


defunctzombie:
**Description**
The raw message panel _diff_ mode shows an incorrect diff when using _filters_ in the topic path. Rather than seeing a diff from `WAITING -> SUCCESS`, I see `SUCCESS` as a new message indicating there was no previously message. I'd expect the raw message panel to remember the previously matched message and show me a diff against that with my new message.

```typescript
"/foo": [
      {
        topic: "/foo",
        receiveTime: { sec: 123, nsec: 1 },
        message: { type: 2, status: "WAITING" },
        sizeInBytes: 0,
      },
      {
        topic: "/foo",
        receiveTime: { sec: 123, nsec: 2 },
        message: { type: 1, status: "FAIL" },
        sizeInBytes: 0,
      },
      {
        topic: "/foo",
        receiveTime: { sec: 123, nsec: 3 },
        message: { type: 2, status: "SUCCESS" },
        sizeInBytes: 0,
      },
    ],
```

<img width="1042" alt="image" src="https://user-images.githubusercontent.com/84792/153676509-0c4fae65-1102-4d40-9c64-c3426fa1989a.png">

https://github.com/foxglove/studio/blob/main/packages/studio-base/src/panels/RawMessages/index.tsx#L163-L166

PR #2830, opened 2022-02-11
Fix diff mode for filtered raw message values

defunctzombie:
**User-Facing Changes**
Using raw message panel in "previous topic diff mode" with a filtered path shows you a diff between the previously filtered value and the current filtered value.

Before:

After:
image

**Description**
When using the raw message panel with a message path filter (i.e. `/foo{type=2}`), the raw message panel diff mode would compare the latest filtered value against the previous message on the topic, rather than the previous filtered message.

This change updates the raw message panel logic to track consecutive filtered messages rather than any message on the topic. For users who want to see diff between the current filtered message and any message on the topic, the can select the _custom_ option and enter a message path that only contains their topic.

Fixes: #2784

### DIFF
--- a/packages/studio-base/src/panels/RawMessages/index.tsx
+++ b/packages/studio-base/src/panels/RawMessages/index.tsx
@@ -41,7 +41,7 @@ import {
 } from "@foxglove/studio-base/components/MessagePathSyntax/messagePathsForDatatype";
 import parseRosPath from "@foxglove/studio-base/components/MessagePathSyntax/parseRosPath";
 import { MessagePathDataItem } from "@foxglove/studio-base/components/MessagePathSyntax/useCachedGetMessagePathDataItems";
-import { useLatestMessageDataItem } from "@foxglove/studio-base/components/MessagePathSyntax/useLatestMessageDataItem";
+import { useMessageDataItem } from "@foxglove/studio-base/components/MessagePathSyntax/useMessageDataItem";
 import Panel from "@foxglove/studio-base/components/Panel";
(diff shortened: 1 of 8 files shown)


packages/studio-base/src/panels/RawMessages/index.tsx
```diff
@@ -160,23 +157,29 @@ function RawMessages(props: Props) {
...
-    topics: [topicName],
-    historySize: 2,
-  })[topicName];
-  const cachedGetMessagePathDataItems = useCachedGetMessagePathDataItems([topicPath]);
-  const prevTickMsg = consecutiveMsgs?.[consecutiveMsgs.length - 2];
-  const prevTickObj = prevTickMsg && {
-    messageEvent: prevTickMsg,
-    queriedData: cachedGetMessagePathDataItems(topicPath, prevTickMsg) ?? [],
-  };
   const currTickObj = useLatestMessageDataItem(topicPath);
```
jtbandes, line 160: If `useLatestMessageDataItem` can be changed to support `historySize: 2`, that would be a more orthodox solution to this bug.

defunctzombie: Agree - will go that route. If it proves to be sadness I may revert to these sad ways.

defunctzombie: I've updated useLatestMessageDataItem to accept a `historySize` option. I've also renamed it to `useMessageDataItem`. Most of our message apis imply "latest" messages.

jtbandes (approved):

PR #2831, opened 2022-02-12
Add support to detect existing datatype from node playground

defunctzombie:
**User-Facing Changes**
In Node Playground, when using a Message from the `./types` utility as an output type results in publishing messages of that same datatype.

**Description**
When using the `./types` utility file for Message datatypes as output type, node playground detects the original datatype name and sets the node output datatype to the original datatype name. These datatype names matter to other parts of studio which rely on datatype name matching.

This behavior mirrors the existing behavior for the generated "ros" library.

### DIFF
--- a/packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/typescript/ast.ts
+++ b/packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/typescript/ast.ts
@@ -276,6 +276,19 @@ export const constructDatatypes = (
     };
   }
 
+  // In this case, we've detected that the return type comes from the generated types file.
+  // We can look up the datatype name by finding it in the file. The name will be the property name
+  // under which the type exists.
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/typescript/ast.ts
```diff
@@ -276,6 +276,19 @@ export const constructDatatypes = (
     };
   }
 
+  // In this case, we've detected that the return type comes from the generated types file.
+  // We can look up the datatype name by finding it in the file. The name will be the property name
+  // under which the type exists.
+  if (node.getSourceFile().fileName === "/studio_node/generatedTypes.ts") {
+    if (ts.isPropertySignature(node.parent) && ts.isStringLiteral(node.parent.name)) {
```
defunctzombie, line 283: This could be more robust over time but is a "good enough" starting point.

jtbandes (approved):

Issue #2825, opened 2022-02-10
Support /clock in ROS2 connection


bertaveira:
Would be nice to have this feature like in other ways of connecting. Specially when using gazebo and other simulators this is very much needed. 

For some reason some topics seem to work fine even with the /clock time as header stamp but markers don't work

defunctzombie:
@jhurliman assigning this to you per our planning discussion. The referenced PR implements the feature but needs protocol support.

PR #2833, opened 2022-02-12
Add /clock support to ROS2 data source

defunctzombie:
**User-Facing Changes**
When connecting to a running ROS2 system using the ROS2 connection, studio subscribes to the `/clock` topic. Any messages received on the clock topic put the player into a "simulation time" mode where the clock time is used as the currentTime and the receiveTime for messages. This allows Studio to mirror behavior of the ROS2 system when running with simulated time.

**Description**
Update the Ros2Player to handle /clock messages. A /clock message updates the internal time of the player. This time is the time assigned as the _receiveTime_ for new messages. It is also the "currentTime" the player exposes for player state.

Fixes: #2825

### DIFF
--- a/packages/studio-base/src/players/Ros2Player.ts
+++ b/packages/studio-base/src/players/Ros2Player.ts
@@ -11,7 +11,7 @@ import { RosNode } from "@foxglove/ros2";
 import { RosMsgDefinition } from "@foxglove/rosmsg";
 import { definitions as commonDefs } from "@foxglove/rosmsg-msgs-common";
 import { definitions as foxgloveDefs } from "@foxglove/rosmsg-msgs-foxglove";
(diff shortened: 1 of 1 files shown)


jhurliman:
I can't say for sure how gazebo publishes `/clock`, but it looks like we'll need to upgrade @foxglove/ros2 to support unreliable ("best-effort") publishers. As a test you could try modifying the @foxglove/ros2 code so it thinks a best-effort publisher is a reliable publisher and see what happens. It would probably confuse the publisher in some way to be receiving ACKs for an unreliable topic, but it might work for at least a short period of time.

jtbandes:
Related: #2871

jhurliman:
Replaced by https://github.com/foxglove/studio/pull/3106

PR #2835, opened 2022-02-14
Add foxglove.Log message support to Log panel

defunctzombie:
**User-Facing Changes**
Users can view foxglove.Log messages in the Log panel.

**Description**


packages/studio-base/src/panels/Log/FilterBar.tsx
```diff
@@ -18,17 +18,34 @@ import cx from "classnames";
...
+  },
+  {
+    text: ">= ERROR",
+    key: LogLevel.ERROR,
+  },
+  {
+    text: ">= FATAL",
+    key: LogLevel.FATAL,
+  },
+];
```
foxymiles, line 48: Maybe `as const` here?

foxymiles (approved):
👍

packages/studio-base/src/panels/Log/FilterBar.tsx
defunctzombie, line 48:
```suggestion
] as const;
```

Issue #2821, opened 2022-02-10
Deep link with time param may lose seek time state


bryfox:
**Description**

The `time` query param in a URL may change unexpectedly, and seek position lost, under some conditions.

- Version: web and 0.29
- OS: mac

**Steps To Reproduce**

1. Sign into a foxglove web session
2. Open a data-platform URL, with a valid `time`, but with `ds.end` and `ds.start` beyond the bounds of the available data

Example, using demo data: [this link](https://studio.foxglove.dev/?ds=foxglove-data-platform&ds.deviceId=dev_mHH1Cp4gPybCPR8y&time=2018-07-24T03%3A28%3A57.000Z&ds.start=2018-07-24T03%3A28%3A27.000Z&ds.end=2018-07-24T03%3A29%3A27.000Z) settles to [this rewritten link](https://studio.foxglove.dev/?ds=foxglove-data-platform&ds.deviceId=dev_mHH1Cp4gPybCPR8y&ds.end=2018-07-24T03%3A29%3A06.797517001Z&ds.start=2018-07-24T03%3A28%3A47.647951000Z&layoutId=66041df5-1514-494b-9c59-308c73b40031&time=2018-07-24T03%3A28%3A47.647951000Z) for me.

**Expected Behavior**

The requested seek time is maintained. `time` is unchanged in the URL even if start & end are clamped.

**Actual Behavior**

The seek position moves to start. The URL is rewritten with seek `time` set to the start time.

PR #2837, opened 2022-02-15
Clamp currentTime to start/end for FoxgloveDataPlatformPlayer

defunctzombie:
**User-Facing Changes**
When opening a data platform deep link, the player keeps the requested current time even if it has to resize the start/end range.

**Description**
When the player is given a start/end range that does not match available coverage, it will clamp the start/end range to a valid coverage range. Prior to this change, this would also unconditionally set the currentTime to the start time even if the currentTime fell within the new valid range. This change clamps the currentTime to the new valid range.

Fixes: #2821

### DIFF
--- a/packages/studio-base/src/players/FoxgloveDataPlatformPlayer/index.ts
+++ b/packages/studio-base/src/players/FoxgloveDataPlatformPlayer/index.ts
@@ -174,7 +174,7 @@ export default class FoxgloveDataPlatformPlayer implements Player {
       this._end = coverageEnd;
     }
 
-    // During startup, seekPlayback might get called to set the currentTime This might change the
+    // During startup, seekPlayback might get called to set the currentTime. This might change the
     // currentTime from the initial value set in the constructor. So we clamp the currentTime to the
     // new start/end range in the event.
     this._currentTime = clampTime(this._currentTime, this._start, this._end);
--- a/packages/studio-base/src/players/FoxgloveDataPlatformPlayer/index.ts
+++ b/packages/studio-base/src/players/FoxgloveDataPlatformPlayer/index.ts
@@ -168,12 +168,13 @@ export default class FoxgloveDataPlatformPlayer implements Player {
     if (isLessThan(this._start, coverageStart)) {
       log.debug("Reduced start time from", this._start, "to", coverageStart);
       this._start = coverageStart;
-      this._currentTime = this._start;
     }
     if (isGreaterThan(this._end, coverageEnd)) {
       log.debug("Reduced end time from", this._end, "to", coverageEnd);
       this._end = coverageEnd;
     }
+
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/players/FoxgloveDataPlatformPlayer/index.ts
```diff
@@ -168,13 +168,14 @@ export default class FoxgloveDataPlatformPlayer implements Player {
...
       log.debug("Reduced start time from", this._start, "to", coverageStart);
       this._start = coverageStart;
-      this._currentTime = this._start;
     }
     if (isGreaterThan(this._end, coverageEnd)) {
       log.debug("Reduced end time from", this._end, "to", coverageEnd);
       this._end = coverageEnd;
     }
 
+    this._currentTime = clampTime(this._currentTime, this._start, this._end);
```
jtbandes, line 177: This is an initialization function; the currentTime was just set in the constructor. With my current reading of the code I don't think there is any other change to currentTime since the constructor so there's no need for clamping logic here.

```suggestion
    this._currentTime = this._start;
```

defunctzombie: This code is putting back the bug I am fixing. `open` is async. During startup, the `urlInitialState` hook in the URLStateAdapter will call seekPlayback to set the currentTime to the time the user specified in the URL. If the start/end range need to clamp, the logic of setting currentTime to the start time overrides this.

I need to add a comment because changing the code back to what you have is _not_ the behavior we want.

packages/studio-base/src/players/FoxgloveDataPlatformPlayer/index.ts
```diff
@@ -168,13 +168,17 @@ export default class FoxgloveDataPlatformPlayer implements Player {
...
       log.debug("Reduced start time from", this._start, "to", coverageStart);
       this._start = coverageStart;
-      this._currentTime = this._start;
     }
     if (isGreaterThan(this._end, coverageEnd)) {
       log.debug("Reduced end time from", this._end, "to", coverageEnd);
       this._end = coverageEnd;
     }
 
+    // During startup, seekPlayback might get called to set the currentTime This might change the
```
jtbandes, line 177:
```suggestion
    // During startup, seekPlayback might get called to set the currentTime. This might change the
```

jtbandes (approved):
Separately: can we delete all this code that calls the coverage API? Can't the data platform just send us the right time range in the first place?

defunctzombie:
> Separately: can we delete all this code that calls the coverage API? Can't the data platform just send us the right time range in the first place?

This was an additional sanity check on that happening since the user could set arbitrary URL parameters and we still want the player to "work". I think we should keep it.

PR #2840, opened 2022-02-15
Fix seeking on plot and state transition panels pan.

### LINKED ISSUE #2836: Panning the plot panel causes seek on mouseup
**Description**
When I pan around the plot panel using click and drag, a seek occurs after lifting the mouse. I'd expect a seek to happen only when I've clicked without dragging the mouse.

- Version: Foxglove Studio version 0.29.0-dev
- OS: OSX 11.6

**Steps To Reproduce**
Click and drag on the plot panel. When you release, it will seek to wherever your mouse is.

**Expected Behavior**
No seek should happen when the user clicks and drags.

**Actual Behavior**
A seek happens on the mouseup event, regardless of whether the user has dragged the mouse or not.

foxymiles:
**User-Facing Changes**
This fixes an issue with the plot & state transition panels interpreting a pan gesture as a click that should result in a player seek.

**Description**
The fix here is to skip emitting the `onClick` event if the click was part of a pan gesture.

Fixes #2836

### DIFF
--- a/packages/studio-base/src/components/Chart/index.tsx
+++ b/packages/studio-base/src/components/Chart/index.tsx
@@ -175,6 +175,7 @@ function Chart(props: Props): JSX.Element {
     [isMounted],
   );
 
+  const hasPannedSinceMouseDown = useRef(false);
   const previousUpdateMessage = useRef<Record<string, unknown>>({});
 
   // getNewUpdateMessage returns an update message for the changed fields from the last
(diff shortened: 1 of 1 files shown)


defunctzombie:
Was this bug always present or is this the result of some new behavior that recently changed? I've not encountered this bug in the past when panning the plots.

defunctzombie (approved):

Issue #2838, opened 2022-02-15
Add `foxglove.RawImage` support to image panel


defunctzombie:
Related: https://github.com/foxglove/studio/issues/1482

Here is the proposed schema:

Field | Type | Required | Description
-- | -- | -- | --
timestamp | timestamp | ✓ | Timestamp of the image.
width | uint32 | ✓ | Image width
height | uint32 | ✓ | Image height
encoding | string | ✓ | Encoding of the raw image data
step | uint32 | ✓ | Byte length of a single row
data | bytes | ✓ | Raw image data

PR #2841, opened 2022-02-15
Add support for foxglove.RawImage to Image panel

defunctzombie:
**User-Facing Changes**
Users can visualize foxglove.RawImage via the image panel.

Field | Type | Required | Description
-- | -- | -- | --
timestamp | timestamp | ✓ | Timestamp of the image.
width | uint32 | ✓ | Image width
height | uint32 | ✓ | Image height
encoding | string | ✓ | Encoding of the raw image data
step | uint32 | ✓ | Byte length of a single row
data | bytes | ✓ | Raw image data

**Description**
foxglove.RawImage is our custom RawImage message. This allows non-ros users to publish images in the foxglove.RawImage message format and display them in the image panel.

Support is accomplished by normalizing all image message types into a NormalizedImageMessage. This union type supports raw and compressed images. Once normalized, other functions only need to understand the normalized message rather than each specific framework's datatypes.

Fixes: #2838

### DIFF
--- a/packages/studio-base/src/panels/ImageView/useCameraInfo.ts
+++ b/packages/studio-base/src/panels/ImageView/useCameraInfo.ts
@@ -1,0 +1,24 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { useCallback } from "react";
+
+import { MessageEvent } from "@foxglove/studio";
+import { useMessageReducer } from "@foxglove/studio-base/PanelAPI";
+import { CameraInfo } from "@foxglove/studio-base/types/Messages";
+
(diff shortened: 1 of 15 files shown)


packages/studio-base/src/panels/ImageView/useCameraInfo.ts
```diff
@@ -0,0 +1,24 @@
...
+import { useCallback } from "react";
+
+import { MessageEvent } from "@foxglove/studio";
+import { useMessageReducer } from "@foxglove/studio-base/PanelAPI";
+import { CameraInfo } from "@foxglove/studio-base/types/Messages";
+
+import { getCameraInfoTopic } from "./util";
+
+export function useCameraInfo(cameraTopic: string): CameraInfo | undefined {
+  // move lines below to useCameraInfo
```
foxymiles, line 14:
```suggestion
```

foxymiles (approved):
Looks good. Breaking up and refactoring like this is a good move.

Issue #2848, opened 2022-02-16
Add `foxglove.CompressedImage` to image panel


defunctzombie:
From: https://github.com/foxglove/website/pull/334

Field | Type | Required | Description
-- | -- | -- | --
timestamp | timestamp | ✓ | Timestamp of the image
format | string | ✓ | Image format
data | bytes | ✓ | Raw image data

PR #2851, opened 2022-02-16
Add foxglove.CompressedImage support to Image panel

defunctzombie:
**User-Facing Changes**
Users can visualize foxglove.CompressedImage messages in the image panel.

**Description**
Fixes: #2848

### DIFF
--- a/packages/studio-base/src/panels/ImageView/normalizeMessage.ts
+++ b/packages/studio-base/src/panels/ImageView/normalizeMessage.ts
@@ -2,6 +2,7 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
+import { fromNanoSec } from "@foxglove/rostime";
 import { CompressedImage, Image } from "@foxglove/studio-base/types/Messages";
 
 type RawImageMessage = {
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/panels/ImageView/normalizeMessage.ts
```diff
@@ -97,6 +105,20 @@ export function normalizeImageMessage(
         data: typedMessage.data,
       };
     }
+    case "foxglove.CompressedImage": {
+      const typedMessage = message as FoxgloveCompressedImageMessage;
+      const sec = typedMessage.timestamp / 1000000000n;
+      const stamp = {
+        sec: Number(sec),
+        nsec: Number(typedMessage.timestamp - sec * 1000000000n),
+      };
```
jhurliman, line 114: Recommend using @foxglove/rostime for this conversion which handles normalization and catches edge cases

defunctzombie: What method in rostime would do this? I wasn't aware that rostime did anything with bigint. And we haven't published any rostime docs so finding the API is a hunt through the source.

defunctzombie: Found this: https://github.com/foxglove/rostime/blob/main/src/timeUtils.ts#L295

Maybe we need to update the logic here. The Math.trunc is not necessary. I wonder if the `%` is worse than doing the multiply and subtract.

defunctzombie: Updated to use fromNanoSec.

foxymiles (approved):

Issue #2847, opened 2022-02-16
Add `foxglove.CameraCalibration` support to image panel


defunctzombie:
From: https://github.com/foxglove/website/pull/334

Field | Type | Required | Description
-- | -- | -- | --
timestamp | timestamp | ✓ | Timestamp for the image
width | uint32 | ✓ | Image width
height | uint32 | ✓ | Image height
distortion_model | string |   | Name of the distortion model
D | float64[] |   | Distortion parameters
K | float64[] |   | Intrinsic camera matrix (3x3 row-major matrix)
R | float64[] |   | Rectification matrix (3x3 row-major matrix)
P | float64[] |   | Projection/camera matrix (3x4 row-major matrix)

PR #2755, opened 2022-01-28
Visualize camera models and images in 3D panel

jhurliman:
WIP

Screen Shot 2022-01-11 at 3 16 59 PM

- Always return uint32 from rgba()
- Improve typing for CameraInfo
- Improve robustness of PinholeCameraModel, add rectifyPoint() and projectPixelTo3dRay()
- Remove redundant toJSON check
- Render CameraInfo in 3D panel, skeleton for Image/CompressedImage rendering in 3D panel

PR #2856, opened 2022-02-16
Add foxglove.CameraCalibration support to image panel

defunctzombie:
**User-Facing Changes**
foxglove.CameraCalibration messages are supported in the image panel and used to correct distortion when displaying markers

**Description**
Fixes: #2847

### DIFF
--- a/packages/studio-base/src/panels/ImageView/useCameraInfo.ts
+++ b/packages/studio-base/src/panels/ImageView/useCameraInfo.ts
@@ -2,22 +2,63 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
-import { useCallback } from "react";
+import { useCallback, useMemo } from "react";
 
 import { MessageEvent } from "@foxglove/studio";
-import { useMessageReducer } from "@foxglove/studio-base/PanelAPI";
+import { useMessageReducer, useDataSourceInfo } from "@foxglove/studio-base/PanelAPI";
 import { CameraInfo } from "@foxglove/studio-base/types/Messages";
 
+import type { FoxgloveCameraCalibration } from "./types";
 import { getCameraInfoTopic } from "./util";
 
+function normalizeCameraInfo(message: unknown, datatype: string): CameraInfo | undefined {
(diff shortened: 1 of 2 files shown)


defunctzombie:
I need some help on what to initialize the optional distortion parameters to when not provided in the foxglove message.

jtbandes:
I think the default distortion model should be an empty string, or stay as undefined. The way Studio currently uses this is an empty string will be recognized as "no distortion" and the `unrectifyPoint` function becomes a no-op. So that can be our default behavior.

https://github.com/foxglove/studio/blob/03952160ad6a487430e81d7ed2f5993db4f3d5fd/packages/studio-base/src/panels/ImageView/PinholeCameraModel.ts#L42-L46

packages/studio-base/src/panels/ImageView/useCameraInfo.ts
```diff
@@ -2,22 +2,66 @@
...
+    case "sensor_msgs/msg/CameraInfo":
+      return message as CameraInfo;
+    case "foxglove.CameraCalibration": {
+      const typedMessage = message as FoxgloveCameraCalibration;
+      return {
+        width: typedMessage.width,
+        height: typedMessage.height,
+        distortion_model: (typedMessage.distortion_model ?? "") as DistortionModel,
+        D: typedMessage.D ?? [],
+        K: typedMessage.K ?? [],
```
jhurliman, line 26: This doesn't look right

jtbandes (approved):

packages/studio-base/src/panels/ImageView/useCameraInfo.ts
jhurliman: https://github.com/foxglove/studio/pull/2755/files#diff-8fef9811da390ab3d255b559c083bd75c9ae7cb9300ff817564bc01cd9f63f09R37-R47

jhurliman: https://github.com/foxglove/studio/pull/2755/files#diff-8fef9811da390ab3d255b559c083bd75c9ae7cb9300ff817564bc01cd9f63f09R206-R222

Issue #2849, opened 2022-02-16
Add `foxglove.ImageAnnotations` support to image panel


defunctzombie:
## `foxglove.ImageAnnotations`

Field | Type | Required | Description
-- | -- | -- | --
circles | CircleAnnotation[] |   | Circle annotations
points | PointsAnnotation[] |   | Points annotations

## CircleAnnotation

Field | Type | Required | Description
-- | -- | -- | --
timestamp | timestamp | ✓ | Timestamp of the annotation.
id | string | ✓ | Unique id for the annotation
action | uint8 | ✓ | ADD=0REMOVE=1
position | Point2d |   | Center of the circle in 2d image coordinates
diameter | float64 |   | Circle diameter for circle annotations
filled | boolean |   | Fill the shape with color
fill_color | Color |   | Fill color
lifetime | timestamp |   | Lifetime of the annotation in nanoseconds. The annotation is considered stale after timestamp + lifetime

## PointsAnnotation

Field | Type | Required | Description
-- | -- | -- | --
timestamp | timestamp | ✓ | Timestamp of the annotation.
id | string | ✓ | Unique id for the annotation
type | enum | ✓ | POINTS=0POLYGON=1LINE_STRIP=2LINE_LIST=3
action | uint8 | ✓ | ADD=0REMOVE=1
points | Point2d[] |   | Array of points in the annotation
outline_colors | Color[] |   |  
filled | boolean |   | Fill the shape with color
fill_color | Color |   | Fill color
lifetime | timestamp |   | Lifetime of the annotation in nanoseconds. The annotation is considered stale after timestamp + lifetime

PR #2861, opened 2022-02-17
Add foxglove.ImageAnnotations support to image panel

defunctzombie:
**User-Facing Changes**
Users can visualize foxglove.ImageAnnotations messages within the image panel.

**Description**
Introduce useImagePanelMessages in the image panel. This hook handles image and annotation message handling. It outputs normalized images and annotations for rendering. If synchronization is on, it handles aggregating and outputs images and annotations when the synchronization conditions are met.
    
Fixes: #2849


packages/studio-base/src/panels/ImageView/useImagePanelMessages.ts
```diff
@@ -0,0 +1,178 @@
...
+    return out;
+  }, [annotationTopics, imageTopic]);
+
+  const shallowTopics = useShallowMemo(topics);
+
+  const datatypesByTopic = useDatatypesByTopic();
+
+  // fixme - when topics change should we clear the state?
+  // its weird to remove the image if the marker topic is toggled on/off
+  // but also weird to not if the image topic changes
```
jtbandes, line 127: I mostly agree with this comment (although it's not weird to remove the image if a marker is toggled on and synchronization is enabled). I think you can achieve this by storing the imageTopic in the state, and comparing the current imageTopic with state.imageTopic when restoring.

packages/studio-base/src/util/synchronizeMessages.ts
```diff
@@ -42,51 +42,6 @@ function allMessageStampsNewestFirst(
...
-      getHeaderStamp,
-    );
-    if (synchronizedMessagesByTopic != undefined) {
-      return synchronizedMessagesByTopic;
-    }
-  }
-  return undefined;
-}
-
 function getSynchronizedMessages(
```
jtbandes, line 45: aren't these all unused now?

packages/studio-base/src/panels/ImageView/useImagePanelMessages.ts
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  // Find the oldest entry where we have images and annotations
+  // We can display this to the user
+  let validEntry: [Time, ImagePanelMessages] | undefined = undefined;
+  for (const entry of state.tree.entries()) {
+    const messageState = entry[1];
+    if (messageState.image && messageState.annotations) {
+      validEntry = entry;
+    }
```
jtbandes, line 84: I think that when multiple marker topics are used, and the topics arrive out of sync, the behavior of rendering when _at least_ one marker is available could be undesirable. The image would appear with partial markers as soon as one becomes available, and then it would update with new markers as they become available (if I am understanding this code correctly).

In the worst case, if some of your markers take so long to compute that they always arrive after the _next image_ message, then you will actually **never** be able to visualize those markers, if other ones are also enabled that have lower latency.

So I think we should probably keep the current behavior as the default, though maybe making a panel setting (or per-topic setting) to use "at least one" mode would be fine.

packages/studio-base/src/panels/ImageView/util.test.ts
```diff
@@ -207,20 +200,20 @@ describe("ImageView", () => {
     it("requires cameraInfo if transformMarkers is true", () => {
```
jtbandes, line 200: This test change seems a bit strange. It's not immediately obvious that this function would be expected to return non-undefined if the markers array is empty – maybe that's just what it does when there are no markers to render. Not sure the logic that was supposed to be tested is still being tested.

defunctzombie: I fixed the test - it was testing that when markers are non-empty and you want to transform the markers you have to have camera calibration data - otherwise it returns undefined if you don't have that data.

packages/studio-base/src/panels/ImageView/useImagePanelMessages.ts
defunctzombie: I updated the logic to reset when the image topic changes.

packages/studio-base/src/panels/ImageView/useImagePanelMessages.ts
defunctzombie: I've updated the behavior to the current behavior where you need all the topics present before you get synchronized.

packages/studio-base/src/util/synchronizeMessages.ts
defunctzombie: These are gone 🔪

packages/studio-base/src/panels/ImageView/types.ts
```diff
@@ -59,8 +58,102 @@ export type RenderableCanvas = HTMLCanvasElement | OffscreenCanvas;
...
+type FoxgloveImageAnnotationCircleAnnotation = {
+  timestamp: bigint;
+  position: Point2D;
+  diameter: number;
+  thickness: number;
+  fill_color?: Color;
+  outline_color: Color;
+};
+
+type FoxgloveImageAnnotationPointsAnnotation = {
```
foxymiles, line 76: Maybe the last `Annotation` is overkill here?

packages/studio-base/src/panels/ImageView/useImagePanelMessages.ts
```diff
@@ -0,0 +1,257 @@
...
+  "studio_msgs/ImageMarkerArray",
+  "studio_msgs/msg/ImageMarkerArray",
+  "visualization_msgs/ImageMarkerArray",
+  "visualization_msgs/msg/ImageMarkerArray",
+  "ros.visualization_msgs.ImageMarkerArray",
+  // backwards compat with webviz
+  "webviz_msgs/ImageMarkerArray",
+  // foxglove
+  "foxglove.ImageAnnotations",
+];
```
foxymiles, line 60: Maybe good to mark things like this as `readonly`?

packages/studio-base/src/panels/ImageView/useImagePanelMessages.ts
```diff
@@ -0,0 +1,257 @@
...
+      };
+      item.annotationsByTopic.set(topic, annotations);
+      state.tree.set(sameStamp, item);
+    } else {
+      // Annotations have different timestamps. This means we have to group them by stamp.
+      // Then for each stamp, we update the annotations by topic at that stamp.
+
+      const groups = new Map<bigint, Annotation[]>();
+      for (const annotation of annotations) {
+        const key = toNanoSec(annotation.stamp);
```
foxymiles, line 121: Bigints tend to be slow. I wonder if using a string key would be faster here. Maybe there aren't enough items for it to matter?

foxymiles (approved):
Looks good. Having normalized messages cleans up the logic nicely.

packages/studio-base/src/panels/ImageView/useImagePanelMessages.ts
```diff
@@ -0,0 +1,257 @@
...
+
+  // Update annotations at the stamp time
+  if (annotations) {
+    // If we know all the annotations are the same timestamp, we can shortcut
+    // and set the annotations by topic directly for the single stamp
+    const sameStamp = annotations.reduce((prev, item) => {
+      // If any annotation stamps are different we return undefined
+      if (!prev || compareTime(item.stamp, prev) !== 0) {
+        return undefined;
+      }
```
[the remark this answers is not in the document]
defunctzombie: Fixed.

packages/studio-base/src/panels/ImageView/useImagePanelMessages.ts
defunctzombie: Fixed.

packages/studio-base/src/panels/ImageView/useImagePanelMessages.ts
defunctzombie: The tradeoff is that I then do `fromNanoSec` a little later in the code. /shrug Hard to say when it will "matter".

PR #2867, opened 2022-02-17
Improve error reporting for PointCloud2 x/y/z fields

jtbandes:
**User-Facing Changes**
Improved error reporting for certain invalid PointCloud2 messages.

**Description**
When a field datatype is not recognized, the `extractValues` function will return `NaN` for its values. This is undesirable for x/y/z fields because the point cloud will silently fail to display points rather than displaying an error to the user. To fix this I improved the error handling to throw earlier if the field datatype was not recognized.

There was also some special-case handling for x/y/z fields which uses a fast-path `reinterpretBufferToFloat` function, even if the field datatypes were not float32 😱  This could mean that if the user forgot to set the datatype field, the points might actually show up in the correct locations even though the data is technically incorrect. Added a check for the float32 datatype in the fast path to fix this.

### DIFF
--- a/packages/studio-base/src/panels/ThreeDimensionalViz/commands/PointClouds/buffers.ts
+++ b/packages/studio-base/src/panels/ThreeDimensionalViz/commands/PointClouds/buffers.ts
@@ -132,13 +132,35 @@ export function createPositionBuffer({
   stride: number;
 }): VertexBuffer {
   const { x: xField, y: yField, z: zField } = fields;
-  if (!xField || !yField || !zField) {
-    throw new Error("Cannot create a position buffer without x, y, and z fields");
+  if (!xField?.reader) {
+    throw new Error(
+      `${
+        xField ? `Unsupported datatype ${xField.datatype} for` : "Missing"
+      } x field in point cloud. Point clouds cannot be displayed without readable x, y, and z fields.`,
+    );
+  }
+  if (!yField?.reader) {
+    throw new Error(
+      `${
+        yField ? `Unsupported datatype ${yField.datatype} for` : "Missing"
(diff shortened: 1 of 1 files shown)


defunctzombie:
> Improved error reporting for certain invalid PointCloud2 messages.

How do we report these errors to the user? Could you show a screenshot.

packages/studio-base/src/panels/ThreeDimensionalViz/commands/PointClouds/buffers.ts
```diff
@@ -132,13 +132,35 @@ export function createPositionBuffer({
   stride: number;
 }): VertexBuffer {
   const { x: xField, y: yField, z: zField } = fields;
-  if (!xField || !yField || !zField) {
-    throw new Error("Cannot create a position buffer without x, y, and z fields");
+  if (!xField?.reader) {
+    throw new Error(
+      `${
+        xField ? `Unsupported datatype ${xField.datatype} for` : "Missing"
+      } x field in point cloud. Point clouds cannot be displayed without readable x, y, and z fields.`,
```
defunctzombie, line 139: Not sure I would know what a `readable` field is compared to a having the field.

defunctzombie (approved):
I guess this is an improvement in the messages we show users. I'd prefer if these were topic errors (maybe they are?) rather than send notifications (which I suspect they are).

jtbandes:
They are neither — you made these errors crash the panel in https://github.com/foxglove/studio/pull/2766

defunctzombie:
> They are neither — you made these errors crash the panel in #2766

Tragic. I think I remember cause this code can't modify the topic errors since its in a "command".

PR #2872, opened 2022-02-18
Add benchmark folder

defunctzombie:
- mcap benchmark player loads messages and plays them with no delay
- memory layout provides pre-defined layouts to benchmakr

Run
```
yarn benchmark:serve
```

Open a browser and load an mcap file.

### DIFF
--- a/benchmark/package.json
+++ b/benchmark/package.json
@@ -1,0 +1,12 @@
+{
+  "name": "benchmark",
+  "private": true,
+  "dependencies": {
+    "@foxglove/studio-base": "workspace:*"
+  },
+  "devDependencies": {
+    "@types/copy-webpack-plugin": "10.1.0",
+    "copy-webpack-plugin": "10.2.4",
+    "webpack": "5.68.0"
+  }
+}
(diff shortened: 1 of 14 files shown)


jtbandes (commented):
Is it necessary to duplicate the webpack config and index.tsx?

defunctzombie:
> Is it necessary to duplicate the webpack config and index.tsx?

"yes" - mostly cause I imagine taking benchmarks in a different build direction than _web_ folder goes. We might want to add some logs or other signals for automated tools. Its not strictly necessary but this is an easy thing to maintain separately.

defunctzombie:
> Is this meant to be used totally manually or with some automation? It would be nice to add some docs (benchmark/README.md?) around this cause today I see a benchmark folder but I don't know what to do with it.

I am hoping with some automation but right now it is manual - but not entirely sure yet. This is very much a WIP idea that I published to see what folks thought. I see now that somehow it is not a draft PR.

defunctzombie:
One thing I haven't figured out yet with this is why the 3d panel doesn't display anything  - or maybe it does and I don't see it fast enough.

Issue #1675, opened 2021-08-19
Image panel marker dropdown should only show compatible topics


jhurliman:
The image panel has a dropdown to select a topic for markers. However, the dropdown is populated with every known topic even though the majority of topics can't be visualized. It should restrict the list to only show topics with datatype `visualization_msgs/ImageMarker` or `foxglove_msgs/ImageMarkerArray`.

amacneil:
Is this related to https://github.com/foxglove/studio/pull/1584?

jtbandes:
Note that we do already have code to filter by datatype: https://github.com/foxglove/studio/blob/6ab76abd4436939053a6cc4978bf563312b31d7b/packages/studio-base/src/panels/ImageView/index.tsx#L291-L307

And this is used to filter the list of options in the dropdown: https://github.com/foxglove/studio/blob/6ab76abd4436939053a6cc4978bf563312b31d7b/packages/studio-base/src/panels/ImageView/util.ts#L53-L58

However, in addition to the main list, there is **also** a field that lets you "add" your own topic, and I assume that's what John is referring to in this ticket. It looks like that field includes all topics that aren't selected, not just the ones with matching datatypes:

https://github.com/foxglove/studio/blob/6ab76abd4436939053a6cc4978bf563312b31d7b/packages/studio-base/src/panels/ImageView/index.tsx#L460-L465

amacneil:
@jtbandes is this still an issue? I have only ever seen marker topics listed in the image marker dropdown.

jtbandes:
See the last paragraph of https://github.com/foxglove/studio/issues/1675#issuecomment-902133899

defunctzombie:
In addition to restricting the list - maybe it should be an action that opens a dialog to select the topic. Or the dropdown should act more like a multi-select combo box and show the other valid topics that are not selected. The current UI look/feel is strange.

PR #2877, opened 2022-02-18
Show all available annotation topics in the annotations dropdown

defunctzombie:
**User-Facing Changes**
Users can select any valid annotation topic using the annotation dropdown. The relevant annotation topics are still first.

Screen Shot 2022-02-18 at 1 52 04 PM

**Description**
This change updates the Image panel annotations dropdown to show all available annotation topics. The "related" topics appear first. Those are the ones that are under the same "camera namespace" (i.e same topic prefix). Then the remaining topics after.
    
This change also switches the Dropdown component to the MUI Select component for the dropdown.
    
Fixes: #1675


amacneil:
Should we hide the other topics under  "more" link or something? Do you think people will find it confusing seeing both related and unrelated annotation topics?

defunctzombie:
> Do you think people will find it confusing seeing both related and unrelated annotation topics?

Unrelated is somewhat of a "ros-ism" that you organize the topics in this way. But maybe enough folks adopt this convention that it works out. They are "unrelated" only in the sense that in ros you namespace the topics like this because the tooling does it by convention. Nothing "really" stops you from making topics like `/annotations/camera_front` that mirror `/camera_front/image` but I agree its somewhat contrived and weird.

We could make this a "more" sub-menu. We need to settle on what type of component this is tho for that. Right now its a "Select" component - that does not have "submenus" I think. We can make it have them tho.

defunctzombie:
@jtbandes @jhurliman for additional thoughts and usecases

amacneil:
Agreed - hard to say what is right/wrong.

Maybe - IF there are matching sub-topics we show those with the rest under a "more" link, and if there are none we show all topics by default?

jtbandes:
I think we have a lot of cleanup to do on the styling of the new dropdown, but the interaction looks ok to me. In the future I imagine we might want to make this look/work more like the 3D panel topic picker since it is kind of doing the same thing.

amacneil:
image

defunctzombie:
@2metres what's the status of this PR? I see that we've removed the disablePortal stuff. Is there some plan for dealing with the floating toolbar? It seems this PR is stuck in some limbo and we should figure out what we need to push it over the finish line.

2metres:
Talked to @foxymiles about it and he has some thoughts about how to implement the less glitchy hover and menu states. But wanted to do it in a separate PR as a fast follow.

defunctzombie:
Plan is to merge this PR without the disablePortal and figure out an approach for resolving the annotations dropdown glitch. This resolves the image topic glitch since the menu closes after selecting the topic.

packages/studio-base/src/components/PanelToolbar/index.tsx
```diff
@@ -58,11 +58,7 @@ const useStyles = makeStyles((theme) => ({
       width: "100%",
       zIndex: 5000,
       backgroundColor: "transparent",
-      pointerEvents: "none",
```
2metres, line 61: I think removing this could cause problems with clicking the underlying canvas

2metres (approved):
Just one minor css thing but looks good to me

packages/studio-base/src/components/PanelToolbar/index.tsx
defunctzombie: I am still able to click on objects in the canvas. Are you not?

defunctzombie: Indeed it did cause problems. https://github.com/foxglove/studio/pull/2989

Issue #2875, opened 2022-02-18
Ability to publish standard ROS messages from Node Playground


amacneil:
1. Build and run [ros1-bench-workspace](https://github.com/foxglove/ros1-bench-workspace)
2. Connect in Studio via rosbridge
3. Open node playground and load the "Markers" template node.

Studio does not know about `visualization_msgs/Marker`:

<img width="659" alt="image" src="https://user-images.githubusercontent.com/637671/154730035-7d25df28-2b5b-4088-a9da-b9a06058fe9e.png">

We should ship these standard messages by default even if they are not currently being published by the ROS source.

defunctzombie:
This should also include the foxglove types: `foxglove.LocationFix`

jtbandes:
Isn't this a duplicate of https://github.com/foxglove/studio/issues/1261? Why was that one closed?

defunctzombie:
@jtbandes This is for the new syntax `Message<...>`. That ticket was for the old syntax and we did implement that feature for the new syntax.

defunctzombie:
Relevant lines of code: https://github.com/foxglove/studio/blob/main/packages/studio-base/src/players/UserNodePlayer/index.ts#L150

PR #2878, opened 2022-02-18
Add builtin types to node playground `Message` type

defunctzombie:
**User-Facing Changes**
Using `Message<"visualization_msgs/Marker">` or `Message<"foxglove.LocationFix">` works even if your data source does not contain the datatype. This makes it possible to write nodes that emit these messages from messages of some other datatype.

**Description**

Fixes: #2875


packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/generateTypesLib.ts
```diff
@@ -20,8 +20,8 @@ const rosPrimitiveToTypescriptTypeMap = new Map<string, string>([
   ["int32", "number"],
   ["float32", "number"],
   ["float64", "number"],
-  ["int64", "number"],
-  ["uint64", "number"],
+  ["int64", "bigint"],
+  ["uint64", "bigint"],
```
jtbandes, line 24: Wouldn't we expect to see some snapshot tests changing with this?

defunctzombie: Yea - they will. I'll have to update those.

defunctzombie: Actually there were no snapshots that exposed this.

PR #2885, opened 2022-02-22
Change panel error boundary reset to also clear the error

defunctzombie:
**User-Facing Changes**
Clicking "Reset" on a crashed panel will clear the error before resetting the panel.

**Description**
When clicking the "Reset" button on a panel error boundary, the panel would reset but the error would not be cleared from the boundary. This change clears the error from the boundary and then resets the panel.


foxymiles (approved):

defunctzombie:
If I had known about that when I did the redesign maybe - but given that the error boundaries are now in a reasonable place I don't think I'll refactor that that. The code for the error boundaries is mostly style rather than substance.

Issue #2761, opened 2022-01-31
Support mcap v0 streaming from data platform


jtbandes:
Studio should request mcap v0 from data platform and migrate to the new v0 reader package, which we will publish on npm.

PR #2889, opened 2022-02-22
Update Data Platform player to request MCAP v0 data

jtbandes:
**User-Facing Changes**
Improved compatibility with MCAP data streamed from the Foxglove Data Platform.

**Description**
Closes https://github.com/foxglove/studio/issues/2761
Closes https://github.com/foxglove/studio/issues/2921
Blocked on https://github.com/foxglove/data-platform/pull/319

### DIFF
--- a/packages/studio-base/src/players/FoxgloveDataPlatformPlayer/index.ts
+++ b/packages/studio-base/src/players/FoxgloveDataPlatformPlayer/index.ts
@@ -3,12 +3,11 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
 import { captureException } from "@sentry/core";
+import { isEqual } from "lodash";
 import { v4 as uuidv4 } from "uuid";
 
 import Logger from "@foxglove/log";
-import { parse as parseMessageDefinition } from "@foxglove/rosmsg";
-import { LazyMessageReader } from "@foxglove/rosmsg-serialization";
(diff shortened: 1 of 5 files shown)


packages/studio-base/src/players/FoxgloveDataPlatformPlayer/index.ts
```diff
@@ -181,44 +177,38 @@ export default class FoxgloveDataPlatformPlayer implements Player {
...
+        messageEncoding,
+        schema: { name: schemaName, data: schema, encoding: schemaEncoding },
+      });
+
+      topics.push({ name: topic, datatype: parsedChannel.fullSchemaName });
+      parsedChannels.push({ messageEncoding, schemaEncoding, schema, parsedChannel });
+
+      // Final datatypes is an unholy union of schemas across all channels
+      for (const [name, datatype] of parsedChannel.datatypes) {
+        datatypes.set(name, datatype);
```
wkalt, line 211: datatypes aren't uniquely described by name, right? within a single recording they probably will be, but I think there's nothing that guarantees that? can't tell if we are assuming this here.

jtbandes: We are assuming this here and in every other player in studio, currently the app doesn't support different channels having different schemas with the same name at all.

bryfox (approved):

foxymiles (approved):

Issue #2886, opened 2022-02-22
Special characters in MCAP topic names cause issues.


jameskuszmaul-brt:
**Description**

I was converting some data to an MCAP file to try out Foxglove; I found that while MCAP appears to permit arbitrary UTF-8 strings for topic names, if I use a topic name with spaces, hyphens, or periods (or presumably other types of special characters), adding a field on that topic to a plot will result in the field simply not showing (I am able to add to the legend, but no data for that line actually displays on the plot).

- Version: 1.1.0 (ran `docker run --rm -p "8080:8080" ghcr.io/foxglove/studio:latest` on 2/17).
- OS: Using browser (Chrome)

**Steps To Reproduce**
Create an MCAP log with a topic that contains at least one space. I was noticing this with things of the format `/topic/name subtopic.Type` (for sundry reasons, the source data I'm working with uses a slightly different definition of "topic" where you need both a name and a type to uniquely resolve topics). If I replace all the spaces/`.`'s with `/`'s, then things work fine.

I can provide an exact mcap file for repro if needed, but that'll require me generating some non-private data.

**Expected Behavior**
Ideally, arbitrary topic names are allowed. At a minimum, the error is more obvious.

**Actual Behavior**
No obvious failure; instead data simply doesn't display.

jameskuszmaul-brt:
Ended up needing to make thing simpler to reproduce anyways, so I pulled some https://github.com/frc971/971-Robot-Code to play with.

The two below MCAP files are identical except for the topic names that they use (files are ~13 KB).
[good_topic_name.mcap.gz](https://github.com/foxglove/studio/files/8113781/good_topic_name.mcap.gz)
[bad_topic_name.mcap.gz](https://github.com/foxglove/studio/files/8113782/bad_topic_name.mcap.gz)

jtbandes:
Thanks for the report! This is a limitation with certain panels in the app, namely those which use [message path syntax](https://foxglove.dev/docs/studio/app-concepts/message-path-syntax). I agree the error should be made more obvious. Since supporting arbitrary topic names seems like a nice feature for the app to have I'm going to convert this to a feature request.

jameskuszmaul-brt:
Yeah, I don't have a pressing need to have arbitrary characters in the topic name (although it would be nice), but the lack of a better error message really threw me for a loop (especially since I was creating the MCAP logs myself and so was simultaneously trying to validate that I wasn't messing anything up on my end).

PR #2891, opened 2022-02-22
MessagePathInput: support wrapping topic names in quotes

jtbandes:
**User-Facing Changes**
Message path syntax now supports wrapping topic names in double quotes, to support topics with special characters. Quotes and backslashes in the string must be escaped with backslashes.

**Description**
Resolves https://github.com/foxglove/studio/issues/2886

image

### DIFF
--- a/packages/studio-base/src/components/MessagePathSyntax/constants.ts
+++ b/packages/studio-base/src/components/MessagePathSyntax/constants.ts
@@ -51,6 +51,7 @@ export type MessagePathPart =
 
 export type RosPath = {
   topicName: string;
+  topicNameRepr: string;
   messagePath: MessagePathPart[];
   modifier?: string;
 };
--- a/packages/studio-base/src/components/MessagePathSyntax/parseRosPath.ts
+++ b/packages/studio-base/src/components/MessagePathSyntax/parseRosPath.ts
@@ -21,6 +21,13 @@ import grammar from "./grammar.ne";
 
 const grammarObj = Grammar.fromCompiled(grammar);
 
+export function quoteIdentifierIfNeeded(name: string): string {
+  if (name.match(/^[a-zA-Z0-9_/]+$/)) {
+    return name;
+  }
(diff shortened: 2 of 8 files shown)


packages/studio-base/src/components/MessagePathSyntax/constants.ts
```diff
@@ -51,6 +51,7 @@ export type MessagePathPart =
 
 export type RosPath = {
   topicName: string;
+  topicNameRepr: string;
```
defunctzombie, line 54: What is this field? Lets add a comment while we're here for our future selves.

packages/studio-base/src/components/MessagePathSyntax/useCachedGetMessagePathDataItems.ts
```diff
@@ -374,7 +374,7 @@ export function getMessagePathDataItems(
   }
   const structure = structures[topic.datatype];
   if (structure) {
-    traverse(message.message, 0, filledInPath.topicName, structure);
+    traverse(message.message, 0, quoteIdentifierIfNeeded(filledInPath.topicName), structure);
```
defunctzombie, line 377: Should `traverse` do the quoting?

packages/studio-base/src/components/MessagePathSyntax/parseRosPath.ts
```diff
@@ -21,6 +21,13 @@ import grammar from "./grammar.ne";
 
 const grammarObj = Grammar.fromCompiled(grammar);
 
+export function quoteIdentifierIfNeeded(name: string): string {
```
defunctzombie, line 24: Comment please.

packages/studio-base/src/components/MessagePathSyntax/useCachedGetMessagePathDataItems.ts
jtbandes: Mostly it just appends things onto this path. It should probably do some quoting if we want to also add support for special chars in field names.

defunctzombie (approved):

PR #2903, opened 2022-02-24
Make useContextSelector concurrent mode safe

defunctzombie:
**User-Facing Changes**
None

**Description**
useContextSelector had side-effects in render. This is unsafe with react concurrent mode. This change removes the side-effects and simplifies the logic for updating the selected value.

### DIFF
--- a/packages/studio-base/src/hooks/useContextSelector.ts
+++ b/packages/studio-base/src/hooks/useContextSelector.ts
@@ -11,7 +11,7 @@
 //   found at http://www.apache.org/licenses/LICENSE-2.0
 //   You may not use this file except in compliance with the License.
 
-import { useRef, useLayoutEffect, useContext, useReducer } from "react";
+import { useLayoutEffect, useContext, useState } from "react";
 
 import { SelectableContext } from "@foxglove/studio-base/util/createSelectableContext";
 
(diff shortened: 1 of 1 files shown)


defunctzombie:
@jtbandes before I go and update all the tests related to this change - do you think this is the right change to make? It looks sensible to me.

jtbandes:
The change makes sense, but the previous implementation was intentionally designed to not require 2 renders for any update (which is why the tests are failing). Do you think there is some way we can keep that behavior while being concurrent safe? Maybe it's just not possible without https://github.com/reactjs/rfcs/pull/119 ? Or I wonder if they have solved this behavior in https://github.com/dai-shi/use-context-selector? (Haven't explored this library much)

packages/studio-base/src/hooks/useContextSelector.ts
```diff
@@ -36,19 +36,29 @@ export default function useContextSelector<T, U>(
...
+    if (shouldUpdate) {
+      state.current = {
+        contextValue,
+        selectedValue,
+        selector,
+      };
+    }
+  });
 
-  // Subscribe to context updates, and trigger a re-render when the selected value changes.
```
jtbandes, line 51: keep this comment?

jtbandes (approved):
This makes sense and feels like a more minimal change.

packages/studio-base/src/hooks/useContextSelector.ts
```diff
@@ -36,19 +36,29 @@ export default function useContextSelector<T, U>(
...
   useLayoutEffect(() => {
```
defunctzombie, line 62:
```suggestion
  // Subscribe to context updates, and trigger a re-render when the selected value changes.
  useLayoutEffect(() => {
```

PR #2908, opened 2022-02-24
Expose block cache from foxglove DP player.

### LINKED ISSUE #1964: Console player: Support preloading
Requires larger refactor to allow players other than RandomAccessPlayer to vend preloaded messages.

foxymiles:
**User-Facing Changes**
This will improve preloading behavior for the Foxglove Data Platform player.

**Description**
`FoxgloveDataPlatformPlayer` uses its own memory cache implementation and this cache doesn't expose cached messages in the blocks format that some panels expect. This patch just exposes the cached messages in blocks as well as the format used internally by `FoxgloveDataPlatformPlayer`.

Fixes #1964

### DIFF
--- a/packages/studio-base/src/players/FoxgloveDataPlatformPlayer/MessageMemoryCache.ts
+++ b/packages/studio-base/src/players/FoxgloveDataPlatformPlayer/MessageMemoryCache.ts
@@ -2,8 +2,14 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
+import { groupBy, sumBy } from "lodash";
+
 import { areEqual, isGreaterThan, isLessThan, Time, toString } from "@foxglove/rostime";
 import { MessageEvent } from "@foxglove/studio-base/players/types";
+import {
+  BlockCache,
+  MemoryCacheBlock,
+} from "@foxglove/studio-base/randomAccessDataProviders/MemoryCacheDataProvider";
 import { Range } from "@foxglove/studio-base/util/ranges";
 
 /** Represents a half-open time interval: [start, end) */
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/players/FoxgloveDataPlatformPlayer/MessageMemoryCache.ts
```diff
@@ -2,8 +2,14 @@
...
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
+import { groupBy, sumBy } from "lodash";
+
 import { areEqual, isGreaterThan, isLessThan, Time, toString } from "@foxglove/rostime";
 import { MessageEvent } from "@foxglove/studio-base/players/types";
+import {
+  BlockCache,
+  MemoryCacheBlock,
+} from "@foxglove/studio-base/randomAccessDataProviders/MemoryCacheDataProvider";
```
jtbandes, line 12: We should avoid importing from the random access data providers here. Let's move the type definitions up to a higher level at the Player API.

jtbandes (approved):
This looks good except the import from randomAccessDataProviders.

By the way, do you understand why the blockCache only has a startTime and not an endTime? Wondering what that's used for. I find it strange that the blocks just have an array of messages and not a start/end time.

packages/studio-base/src/players/FoxgloveDataPlatformPlayer/MessageMemoryCache.ts
foxymiles: Good idea I've refactored those types out.

foxymiles:
> This looks good except the import from randomAccessDataProviders.
> 
> By the way, do you understand why the blockCache only has a startTime and not an endTime? Wondering what that's used for. I find it strange that the blocks just have an array of messages and not a start/end time.

I'm not sure. It might be because it looks like it was originally used just to indicate preloading progress. I agree with your earlier comment that we should probably refactor this at some point.

Issue #2900, opened 2022-02-23
Dragging playback bar causes it to get stuck in seek mode


amacneil:
**Steps To Reproduce**

1. mousedown on the seek bar, then drag your mouse off the bar, then release it
2. now, every time you hover over the playback bar it will seek to the mouse cursor position
3. the only way to get out of this mode is to click on the playback bar

happening on current web + desktop

https://user-images.githubusercontent.com/637671/155410644-5282e9fd-28f3-490d-b3c3-56a69e231d04.mov

jtbandes:
This regressed in https://github.com/foxglove/studio/pull/1805.

PR #2912, opened 2022-02-25
Fix playback bar getting stuck in drag mode

jtbandes:
**User-Facing Changes**
Fixed an issue where scrubbing on the playback bar only worked as long as the mouse was actually kept on the playback bar, and could get "stuck" if the mouse was released when outside of the playback bar.

**Description**
Fixes https://github.com/foxglove/studio/issues/2900. This behavior was previously working and regressed in https://github.com/foxglove/studio/pull/1805. The mouseup and mousemove events are now tracked on the window instead of the slider element when the mouse is pressed, which improves the UX by allowing a drag to move outside the playback bar and continue working, and allows us to correctly track mouseup even when it happens outside the playback bar (or outside the window). Additionally, a mouseenter/mouseleave pair are used to track the "mouse inside" state, which allows us to hide the tooltip on mouse up, if the mouse was outside the slider.

https://user-images.githubusercontent.com/14237/155650472-8216f61d-0b98-4bc1-9f78-e25f624adca2.mov

### DIFF
--- a/packages/studio-base/src/components/PlaybackControls/Slider.tsx
+++ b/packages/studio-base/src/components/PlaybackControls/Slider.tsx
@@ -3,7 +3,15 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
 import { clamp } from "lodash";
-import { useCallback, useEffect, useRef, MouseEvent, ReactNode, useMemo } from "react";
+import {
+  useCallback,
+  useEffect,
+  useRef,
+  ReactNode,
+  useMemo,
+  useState,
+  useLayoutEffect,
+} from "react";
 import styled from "styled-components";
 
 import Logger from "@foxglove/log";
@@ -16,9 +24,8 @@ type Props = {
   max: number;
   disabled?: boolean;
   step?: number;
-  draggable?: boolean;
   onChange: (value: number) => void;
-  onHoverOver?: (ev: MouseEvent<HTMLDivElement>, value: number) => void;
+  onHoverOver?: (clientX: number, value: number) => void;
   onHoverOut?: () => void;
   renderSlider?: (value?: number) => ReactNode;
 };
@@ -53,7 +60,6 @@ export default function Slider(props: Props): JSX.Element {
     step = 0,
     max = 0,
     min = 0,
-    draggable = false,
     disabled = false,
     renderSlider = defaultRenderSlider,
     onHoverOver,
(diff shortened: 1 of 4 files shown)


packages/studio-base/src/components/PlaybackControls/Slider.tsx
```diff
@@ -138,15 +161,26 @@ export default function Slider(props: Props): JSX.Element {
     return value != undefined && max > min ? (value - min) / (max - min) : undefined;
   }, [max, min, value]);
 
+  useEffect(() => {
+    if (mouseDown) {
+      window.addEventListener("mouseup", onMouseUp);
+      window.addEventListener("mousemove", onMouseMove);
```
defunctzombie, line 167: Do you think there's an overhead of having mousemove events on the window? Seems like now we track all mouse move events and call the mousemove function. Maybe we should attach the listener only when you move into the playback controls area and then have logic in here to remove the listener when you've performed the action?

@foxymiles @2metres this gets me thinking that we could adopt the same tactic for the panel mouse state issue we are facing with the floating toolbar and image panel dropdowns.

packages/studio-base/src/components/PlaybackControls/Slider.tsx
```diff
@@ -82,49 +86,68 @@ export default function Slider(props: Props): JSX.Element {
...
+    setMouseDown(false);
+    if (!mouseInsideRef.current) {
+      onHoverOut?.();
+    }
+  }, [onHoverOut]);
+
   const onMouseMove = useCallback(
-    (ev: MouseEvent<HTMLDivElement>): void => {
+    (ev: React.MouseEvent | MouseEvent): void => {
       const val = getValueAtMouse(ev);
```
defunctzombie, line 121: If you keep the logic of having the mousemove event always attached (which I'd be in favor of attaching when needed). This logic should move below `disabled` or any other cheap check we can do to bail out of this function early since it is getting called on every move on window now.

packages/studio-base/src/components/PlaybackControls/Slider.tsx
foxymiles: @defunctzombie Are we having a similar issue with stuck mouse down in the floating toolbar? Isn't the issue there more that we get a mouseleave when the mouse hovers over the dropdown?

defunctzombie: Yea - the issue there is that we get mouseleave events - my thought was that we could use mousemove to detect if the mouse was still within the panel bounds. Tho it occurs to me now that for menus that go outside the panel bounds that wouldn't work.

jtbandes: > Maybe we should attach the listener only when you move into the playback controls area and then have logic in here to remove the listener when you've performed the action?

That’s exactly what this code does. Note the listener is conditionally attached to window when the mouse is down.

defunctzombie (approved):

packages/studio-base/src/components/PlaybackControls/Slider.tsx
jtbandes: However, the listener is always attached to the element itself, I kept that to make the tooltip work.

Issue #707, opened 2021-04-27
Local files: Support for latched or infrequently published topics when scrubbing


amacneil:
When scrubbing, we should intelligently find the right message to display for topics which are published infrequently.

- [x] Experimental play w/ support for .bag files
- [x] #3027
- [x] #3028
- [x] https://github.com/foxglove/studio/issues/2899
- [x] https://github.com/foxglove/studio/issues/1962
- [x] #3361

References:
- https://github.com/cruise-automation/webviz/issues/314
- https://github.com/fetchrobotics/webviz/commit/ab301ca317e816583db6e93d24b370253c5c2092

cc @cjds

rgov:
I'm not sure what you have in mind but it would be helpful to have a `PanelAPI` function that enumerates messages forward or backward in a given time window.

This be useful for things like video streams (#537), where to decode a given frame you need to find some earlier messages that provide context: the header(s) that describe the width, height, and other parameters; the last intra frame; and all subsequent inter frames that are deltas to the inter frame.

1. Search backwards from `currentTime` to find the header packet message.
2. Search backwards from `currentTime` to `headerPacket.header.stamp` to find the most recent intra frame.
3. Search forwards from `intraFrame.header.stamp` to `currentTime` to collect a list of inter frames.

jhurliman:
This may be related to <https://github.com/foxglove/studio/issues/2343>. Internally latching every topic in Studio would likely make seeking/scrubbing work a lot better in practice, especially with bags that have `/tf_static`.

amacneil:
Also created https://github.com/foxglove/studio/issues/2899 for the same feature in Data Platform

defunctzombie:
Re-opening because this issue is a meta-tracking issue which also needs to have support for mcap.

defunctzombie:
@rgov 

> I'm not sure what you have in mind but it would be helpful to have a `PanelAPI` function that enumerates messages forward or backward in a given time window.

This issue won't cover this functionality. This issue is scoped to only finding the last message on each topic.

PR #2915, opened 2022-02-25
Bag playback with forward and reverse iterators

defunctzombie:
**User-Facing Changes**
Add a new feature flag for "Experimental bag player".

When enabled, using a bag data source (remote or local) will use a new player with new seeking capabilities. Seeking to any location will load the last message on each subscribed topic into panels. This lets panels display the last message on a topic even if it occurred much earlier in the file.

**Description**
This change introduces a new IterablePlayer which uses iterators for message reading and preloading. When seeking, the player uses a reverse iterator to backfill the latest message on all subscribed topics and emit this to the message pipeline. When playing forward the player uses a forward iterator to grab latest messages for the _tick_ and also for the preloaded blocks. Similar to the RandomAccessPlayer, the player accepts pluggable instances which implement message iteration (IIterableSource) for an underlying data source. This change adds a BagIterableSource for handling local and remote bag files. 

The existing RandomAccessPlayer and `getMessages` interfaces were unable to support arbitrary reverse iteration and were designed around fetching messages for pre-defined start/end ranges.

Fixes: #707


packages/studio-base/src/players/BlockBagPlayer/BlockBagPlayer.ts
```diff
@@ -0,0 +1,943 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+//
+// This file incorporates work covered by the following copyright and
+// permission notice:
+//
+//   Copyright 2019-2021 Cruise LLC
```
amacneil, line 8: is there cruise code in this file?

defunctzombie: Not anymore.

packages/studio-base/src/util/overwriteFetch.ts
```diff
@@ -15,10 +15,11 @@
 // in our logs and want the logs to more fully reflect the error message.
 export default function overwriteFetch(): void {
   const originalFetch = global.fetch;
-  global.fetch = async (url: RequestInfo, init?: RequestInit) => {
+  // eslint-disable-next-line @typescript-eslint/promise-function-async
+  global.fetch = (url: RequestInfo, init?: RequestInit) => {
     // Use this replacement error instead of the original one, because this one will have the correct stack trace.
     const replacementError = new TypeError(`Failed to fetch: ${url}`);
-    return await originalFetch(url, init).catch((error) => {
+    return originalFetch(url, init).catch((error) => {
```
jtbandes, line 22: why bother with this change? Is this a performance optimization? When we disable the lint rule we should add a comment describing why it's disabled, if it's not obvious.

packages/studio-base/src/players/BlockBagPlayer/index.ts
```diff
@@ -0,0 +1,5 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+export * from "./BlockBagPlayer";
```
jtbandes, line 5: 🤷 just move the BlockBagPlayer.ts code to index.ts?

packages/studio-base/src/players/BlockBagPlayer/BlockBagPlayer.ts
```diff
@@ -0,0 +1,933 @@
...
+            await this._stateStartDelay();
+            break;
+          case "start-play":
+            await this._stateStartPlay();
+            break;
+          case "idle":
+            await this._emitState();
+            if (this._currentTime) {
+              const start = performance.now();
+              await this.startBlockLoad(this._currentTime);
```
jtbandes, line 446: It seems like this function does not "start" the load, but actually loads blocks, maybe it should be renamed.

packages/studio-base/src/players/BlockBagPlayer/BlockBagPlayer.ts
```diff
@@ -0,0 +1,933 @@
...
+            }
+            break;
+          case "seek-backfill":
+            await this._stateSeekBackfill();
+            break;
+          case "play": {
+            if (!this._currentTime) {
+              throw new Error("Tried to play before initialized");
+            }
+            const blockLoading = this.startBlockLoad(this._currentTime, { emit: false });
```
jtbandes, line 457: shouldn't this call be moved inside the `_statePlay` function? It feels like this is almost a pure state machine, but there is some extra logic inside the `_runState` function that makes it a bit confusing.

packages/studio-base/src/players/BlockBagPlayer/BlockBagPlayer.ts
```diff
@@ -0,0 +1,933 @@
...
+    this._currentTime = end;
+    this._messages = messages;
+    await this._emitState();
+  }
+
+  private async _statePlay() {
+    const subscriptions = this._subscriptions;
+
+    try {
+      while (this._isPlaying && !this._hasError && !this._nextState) {
```
jtbandes, line 640: Seems weird to have a loop inside `_statePlay`, since there is already an outer loop in the state machine; can't we just rely on the fact that `_statePlay` will continue being called as long as we are in the playing state?

packages/studio-base/src/players/BlockBagPlayer/BlockBagPlayer.ts
```diff
@@ -0,0 +1,933 @@
...
+    // Block caching works on the assumption that we are more likely to want the blocks in proximity
+    // to the _time_. This includes blocks ahead and behind the time.
+    //
+    // We build a _loadQueue_ which is an array of block ids to load. The load queue is
+    // organized such that we populate blocks outward from the requested load time. Blocks closest
+    // to the load time are loaded and blocks furthest from the load time are eligible for eviction.
+    //
+    // To build the load queue, two arrays are created. Pre and Post. Pre contains block ids before
+    // the desired start block, and post ids after. For the load queue, we reverse pre and alternate
+    // selecting ids from post and pre.
```
jtbandes, line 696: It seems like we would generally want to load more forwards than backwards, right? We only need to load backwards enough to be able to satisfy the reverse iteration. It's true that in some cases that means we need to load all the way back to the beginning, but not always.

defunctzombie: I thought about this for a while and came to the conclusion that this is actually necessarily true. Lets say you open some event at time _T_. Its actually more likely that you will want to look at data _prior_ to _T_ than after _T_. After thinking about that some more I came to this approach where I try loading data _around_ _T_ under the notion the _T_ is the interesting event and you need to see what happened before and after.

packages/studio-base/src/util/overwriteFetch.ts
defunctzombie: I will remove it from the PR - it was me trying to figure out another bug I was encountering but this was not the problem.

packages/studio-base/src/players/IterablePlayer/BagIterableSource.ts
```diff
@@ -0,0 +1,174 @@
...
+export class BagIterableSource implements IIterableSource {
+  private _source: BagSource;
+  private _bag: Bag | undefined;
+  private _readersByConnectionId = new Map<number, LazyMessageReader>();
+
+  constructor(source: BagSource) {
+    this._source = source;
+  }
+
+  async initialize(): Promise<Initalization> {
```
jtbandes, line 37: let's make this a static method so we don't have to mark the class properties as potentially undefined?

packages/studio-base/src/players/IterablePlayer/BagIterableSource.ts
```diff
@@ -0,0 +1,174 @@
...
+
+    const iterator = this._bag.messageIterator({
+      topics: opt.topics,
+      reverse: opt.reverse,
+      start: opt.start,
+    });
+
+    const readersByConnectionId = this._readersByConnectionId;
+    return {
+      async *[Symbol.asyncIterator](): AsyncIterator<IteratorResult> {
```
jtbandes, line 142: If you're just putting this in an object directly, you could instead replace this all with an IIFE `return (async function*(){ })()`

packages/studio-base/src/players/IterablePlayer/IIterableSource.ts
```diff
@@ -0,0 +1,45 @@
...
+    };
+
+export interface IMessageIterator {
+  [Symbol.asyncIterator](): AsyncIterator<Readonly<IteratorResult>>;
+}
+
+export interface IIterableSource {
+  initialize(): Promise<Initalization>;
+
+  messageIterator(args: MessageIteratorArgs): IMessageIterator;
```
jtbandes, line 44:
```suggestion
  messageIterator(args: MessageIteratorArgs): AsyncIterable<Readonly<IteratorResult>>;
```

packages/studio-base/src/players/IterablePlayer/BagIterableSource.ts
defunctzombie: That would require changing how PlayerManager deals with player initialization from the data source factory. The factory initialize call is sync.

packages/studio-base/src/players/IterablePlayer/BagIterableSource.ts
defunctzombie: We've learned that using an IIFE here is not correct and that it is correct to return an object with Symbol.asyncIterator because the IterablePlayer keeps the iterator around through multiple for-await-of calls. Using the `IFFE` did not work when the player used `break` in a for loop and then wanted to use the same iterator. The break cancelled the iterator.

packages/studio-base/src/players/BlockBagPlayer/BlockBagPlayer.ts
defunctzombie: The state machine works by invoking the state and waiting for the state to finish rather than constantly invoking the state again. I could change play to setState("play") at the end to keep itself playing but I think I'll keep it as is for now.

packages/studio-base/src/players/BlockBagPlayer/BlockBagPlayer.ts
defunctzombie: Moved these into their respective state handlers to keep the state machine _pure_.

packages/studio-base/src/players/IterablePlayer/IterablePlayer.test.ts
```diff
@@ -0,0 +1,152 @@
...
+      // before initialize
+      stateGen(),
+      // start delay
+      stateGen({ presence: PlayerPresence.INITIALIZING }),
+      // startPlay
+      stateGen({ presence: PlayerPresence.PRESENT }),
+      // idle
+      stateGen({ presence: PlayerPresence.PRESENT }),
+      // load blocks
+      stateGen({ presence: PlayerPresence.PRESENT }),
```
jtbandes, line 147: This stateGen thing is weird – just make a `const initialState` and use `{ ...initialState, presence: X }`

web/src/Root.tsx
```diff
@@ -43,24 +43,28 @@ import ExtensionLoaderProvider from "./providers/ExtensionLoaderProvider";
 // useAppConfiguration requires the AppConfigurationContext which is setup in Root
 // AppWrapper is used to make a functional component so we can use the context
 function AppWrapper() {
+  const [experimentalBagPlayer = false] = useAppConfigurationValue<boolean>(
```
jtbandes, line 46:
```suggestion
  const [useIterablePlayer = false] = useAppConfigurationValue<boolean>(
```

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
```diff
@@ -0,0 +1,846 @@
...
+    while (preIds.length > 0 || postIds.length > 0) {
+      const postId = postIds.shift();
+      if (postId != undefined) {
+        loadQueue.push(postId);
+      }
+      const preId = preIds.shift();
+      if (preId != undefined) {
+        loadQueue.push(preId);
+      }
+    }
```
jtbandes, line 643: Feels like these two loops could be combined with a little more math?

jtbandes (approved):
I haven't thoroughly reviewed all the player state implementations, nor the tests. But since this is under a feature flag I don't see a problem with it merging as is.

foxymiles (approved):

packages/studio-base/src/players/IterablePlayer/BagIterableSource.ts
```diff
@@ -0,0 +1,172 @@
...
+          // no-op?
+        },
+      });
+
+      // Call open on the remote reader to see if we can access the remote file
+      await remoteReader.open();
+
+      if (remoteReader.size() === 0) {
+        throw new Error("Cannot play bag file. File is 0 bytes in size.");
+      }
```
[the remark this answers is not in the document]
defunctzombie: Yea - the bag itself will fail. I can remove.

packages/studio-base/src/players/IterablePlayer/BagIterableSource.ts
```diff
@@ -0,0 +1,172 @@
...
+              },
+            };
+          } else {
+            yield {
+              connectionId,
+              msgEvent: undefined,
+              problem: {
+                severity: "error",
+                message: `Cannot deserialize message for missing connection id ${connectionId}`,
+                tip: `Check that your bag file is well-formed. It should have a connection record for every connection id referenced from a message record.`,
```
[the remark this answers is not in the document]
defunctzombie: Well - its what happened? They can take this to their syseng team and ask them maybe? Ideally a user doesn't see this but when they do I think this is fine. I guess the tip could also be "consult your bag manager for assistance" 🤣

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
defunctzombie: I did not have enough math to do so.

packages/studio-base/src/components/MessagePipeline/MessageOrderTracker.ts
```diff
@@ -78,34 +81,36 @@ class MessageOrderTracker {
...
-                  ? this.incorrectMessages
-                  : "not being tracked",
-              };
-              this.incorrectMessages = [];
-              log.warn(
-                `${messageOrder} very different from player.currentTime; without updating lastSeekTime`,
-                info,
-              );
-            }, WAIT_FOR_SEEK_SEC * 1000);
+        if (!didSeek) {
```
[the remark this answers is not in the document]
defunctzombie: I'll add a comment. This is needed because during seek the new player backfills data well before the currentTime which breaks this check of _DRIFT_.

PR #2917, opened 2022-02-25
Clear user session on API 401 response.

### LINKED ISSUE #1952: Detect expired sessions and signed out
**Description**
When a user signs out in another tab, the original tab spews 401 toasts:

image

We should intelligently handle sign out across all tabs, no toast necessary.

foxymiles:
**User-Facing Changes**
This improves our handling of stale authenticated sessions in studio.

**Description**
The idea here is to attach an observer to `ConsoleAPI` and listen for any 401 response. On receiving a 401 we clear the user session. 

Alternatively we could try a more comprehensive solution that involved custom error boundaries and adding custom error handler hooks to all the places we call the API but that would require much more significant changes.

Fixes #1952

### DIFF
--- a/desktop/renderer/components/ConsoleApiCurrentUserProvider.tsx
+++ b/desktop/renderer/components/ConsoleApiCurrentUserProvider.tsx
@@ -2,7 +2,7 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
-import { PropsWithChildren, useCallback, useMemo, useState } from "react";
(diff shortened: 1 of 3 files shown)


jtbandes (commented):
There's another CurrentUserProvider for the desktop app which we need to update as well. Which makes me wonder if there's a way we can avoid duplicating the logic... although it's a small amount of logic so maybe not a big deal.

foxymiles:
> There's another CurrentUserProvider for the desktop app which we need to update as well. Which makes me wonder if there's a way we can avoid duplicating the logic... although it's a small amount of logic so maybe not a big deal.

Good point. I think for now maybe I will just duplicate the logic since it's not very extensive.

jtbandes (approved):

PR #2919, opened 2022-02-25
Avoid finding the dataset value for PlotLegendRow when hidden

defunctzombie:
**User-Facing Changes**
None

When working with large dataset in the plot panel and using the existing legend without current value display, we avoid iterating the dataset to find a value that we won't display.

**Description**
Avoid iterating through the dataset to find the series value when the plot legend row is hidden since the value is not displayed. For large datasets, looping over the dataset results in a large runtime slowdown.

### DIFF
--- a/packages/studio-base/src/panels/Plot/PlotLegendRow.tsx
+++ b/packages/studio-base/src/panels/Plot/PlotLegendRow.tsx
@@ -4,7 +4,7 @@
 
 import { useTheme as useFluentUITheme } from "@fluentui/react";
 import { Close as CloseIcon, Error as ErrorIcon, Remove as RemoveIcon } from "@mui/icons-material";
-import { IconButton, Theme, Tooltip, Typography } from "@mui/material";
+import { IconButton, Theme, Tooltip, Typography, useTheme } from "@mui/material";
 import { createStyles, makeStyles } from "@mui/styles";
 import { ComponentProps, useCallback, useMemo, useState } from "react";
 import { v4 as uuidv4 } from "uuid";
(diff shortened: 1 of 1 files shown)


foxymiles (approved):

packages/studio-base/src/panels/Plot/PlotLegendRow.tsx
```diff
@@ -145,8 +145,19 @@ export default function PlotLegendRow({
       value,
       color: hoverValue?.value != undefined ? fluentUITheme.palette.yellowDark : "inherit",
     };
-  }, [hoverValue, correspondingData, currentTime, fluentUITheme]);
-  const classes = useStyles({ color: currentDisplay.color, enabled: path.enabled, index });
```
defunctzombie, line 149: @2metres @esthersweon I think this pattern results in making new classes. Not entirely 100% on that but I was seeing some emotion stuff in perf logs. Since I was making changes here I unrolled this as well. For small color tweaks on individual components like this I prefer to use `style` props on components and keep classes stable.

packages/studio-base/src/panels/Plot/PlotLegendRow.tsx
```diff
@@ -131,7 +123,15 @@ export default function PlotLegendRow({
   });
 
   const fluentUITheme = useFluentUITheme();
+  const theme = useTheme();
+
   const currentDisplay = useMemo(() => {
+    if (!showPlotValuesInLegend) {
```
defunctzombie, line 129: @esthersweon @2metres I'll make a separate ticket for this since this boolean is a temporary solution that only works if you don't turn on the values in legend feature. We'll need to remove the entire dataset iteration here and pass the latest values to the plot legend.

PR #2922, opened 2022-02-28
Improve preloading for data platform.

foxymiles:
**User-Facing Changes**
This improves the preloading behavior for the data platform player.

**Description**
Downstream code heavily memoizes the `blocks` returned by the player so we need to rebuild this array when we get new messages in order for that code to recognize we have new blocks.

### DIFF
--- a/packages/studio-base/src/players/FoxgloveDataPlatformPlayer/MessageMemoryCache.ts
+++ b/packages/studio-base/src/players/FoxgloveDataPlatformPlayer/MessageMemoryCache.ts
@@ -227,12 +227,14 @@ export default class MessageMemoryCache {
       messages: insertMessages,
     });
 
+    // Create new array to clear later memoizations.
+    this._blockCache.blocks = [...this._blockCache.blocks];
+
     // Build & insert new cache block for this range.
     this._blockCache.blocks.splice(spliceIdx, deleteCount, {
       messagesByTopic: groupBy(insertMessages, (m) => m.topic),
       sizeInBytes: sumBy(insertMessages, (m) => m.sizeInBytes),
     });
-    this._blockCache.blocks = [...this._blockCache.blocks]; // Needed to clear later memoizations.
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/players/FoxgloveDataPlatformPlayer/MessageMemoryCache.ts
```diff
@@ -232,6 +232,7 @@ export default class MessageMemoryCache {
       messagesByTopic: groupBy(insertMessages, (m) => m.topic),
       sizeInBytes: sumBy(insertMessages, (m) => m.sizeInBytes),
     });
+    this._blockCache.blocks = [...this._blockCache.blocks]; // Needed to clear later memoizations.
```
jtbandes, line 235: If the intent is to avoid mutating the blocks array, we should do this before the `splice` call

foxymiles: Actually the mutation approach here is pretty nice I think. Unfortunately this data is heavily memoized downstream so it doesn't catch the mutation.

jtbandes: Yeah, but since the downstream stuff depends on identity, I think we should avoid ever mutating the blocks array since that will break some downstream assumptions.

foxymiles: So you'd prefer to see this expressed as a mix of slice & concat or something like that?

foxymiles: Or you just mean make the copy first then mutate?

jtbandes: Yeah, just copy and then mutate seems fine

jtbandes (approved):

PR #2925, opened 2022-02-28
Render geometry_msgs/PoseArray in the 3D panel

jhurliman:
**User-Facing Changes**

- Added visualization of `geometry_msgs/PoseArray` messages in the 3D panel

**Description**

This reuses the nav_msgs/Path rendering code


jtbandes:
Does this approach work for the AMCL use case? It looks like the desired rendering (from the screenshot in #1630) is a bunch of separate arrows (an "arrow cloud") rather than a connected line?

jhurliman:
@jtbandes for AMCL I imagine you would want a path rather than a bunch of discrete poses, although it could be useful to toggle between the two methods as a topic setting.

jtbandes:
http://wiki.ros.org/amcl

image

This sounds like discrete arrows to me. A full video demo of the RViz behavior is shown at: http://wiki.ros.org/rviz/DisplayTypes/PoseArray

Issue #2926, opened 2022-02-28
Plot jitters due to format of x-axis label


rgov:
Inherited from Webviz but still present on Foxglove Studio 1.1.0 on macOS 12.2.1

From this issue: https://github.com/cruise-automation/webviz/issues/544

---

![ezgif-2-d72313466639](https://user-images.githubusercontent.com/108767/98702499-6db30b80-2348-11eb-9971-d868b66146ba.gif)

This is a plot of a value over time. The plot keeps jumping because the rightmost x-axis label sometimes switches from 1 digit of precision to 2.

jhurliman:
Still present:

https://user-images.githubusercontent.com/195374/165198320-21ff1973-86f9-4be7-99f4-61414b64e927.mp4

defunctzombie:
@jtbandes assigning to you. I think your PR is 👍 and we can go from there.

PR #2933, opened 2022-03-01
Fix jittery x axis tick formatting in Plot panel

jtbandes:
**User-Facing Changes**
Improved formatting of X-axis tick marks in the Plot panel to always use 2 decimal places, reducing jitter for live-updating data.

**Description**
Fixes https://github.com/foxglove/studio/issues/2926 by using a custom number formatter for the first/last tick marks.

This implementation is not ideal because it doesn't do any of the dynamic calculation of the necessary precision which the default formatter does. So if working in a very small or large time range, the format of the first/last ticks might have too much or too little precision.

The Chart.js default implementation is here, where you can see dynamic calculation of the number of digits (which is also the original source of this bug): https://github.com/chartjs/Chart.js/blob/7966227df36fdb3a79f35ac238ea8b5c84b3f3a8/src/core/core.ticks.js#L28-L54

Before:

https://user-images.githubusercontent.com/14237/156094429-0ed1583a-2062-4f08-b364-b0987a8e355a.mov

After: 

https://user-images.githubusercontent.com/14237/156094416-931399e0-3630-44a7-9567-080e093e8801.mov


packages/studio-base/src/components/Chart/worker/ChartJsMux.ts
```diff
@@ -83,6 +84,32 @@ Chart.register(
...
+ * width. See https://github.com/foxglove/studio/issues/2926
+ *
+ * Because this requires passing a `callback` function for the tick options, this has to be done in
+ * the worker, since functions can't be sent via postMessage.
+ */
+function fixTicks(args: RpcUpdateEvent): RpcUpdateEvent {
+  const xScale = args.options?.scales?.x;
+  const numberFormat = new Intl.NumberFormat(undefined, {
+    minimumFractionDigits: 2,
+    maximumFractionDigits: 2,
```
defunctzombie, line 98: What if the data source has more precision? Like when rendering x axis that is not time?

jtbandes: That was kinda my question above. Do you have any suggestions?

defunctzombie: Is there something we can patch some state onto? We could track the largest precision we've seen and always use that. So once we see 2 decimal places, we always use 2 decimal places. 3, etc. Another option is to track the last X items and use the max precision for those. This could still jitter but hopefully less so?

rgov:
I understand the precision is an unresolved issue, but this is probably my #1 source of irritation while monitoring a live system because the plots are constantly glitching. Any chance this incomplete solution could be adopted for now, maybe even only for live sources, or maybe only if the *x* axis is a timestamp where two digits of precision is fine, until the desired improvements can be made?

defunctzombie:
> _x_ axis is a timestamp where two digits of precision is fine

That sounds like something we could do. Even when you zoom in two digits is probably fine?

defunctzombie (approved):
I think if we remove maximumFractionDigits we are good to go - or set it to 3 which is what we determined the Chartjs default is.

jtbandes:
> I think if we remove maximumFractionDigits we are good to go - or set it to 3 which is what we determined the Chartjs default is.

These are not really viable options. In order to avoid jitter using this approach, we need minimumFractionDigits and maximumFractionDigits to be the same.

PR #2936, opened 2022-03-01
Use stable layout selector

### LINKED ISSUE #2932: "Value must not change" error in panel settings when changing selected panel
Steps to reproduce:
- Open the settings of a panel
- Click on another panel to switch to editing its settings

image

foxymiles:
**User-Facing Changes**
None.

**Description**
This converts the config selector we use in `useConfigsById` to a stable function.

Fixes #2932

### DIFF
--- a/packages/hooks/src/useMustNotChange.test.ts
+++ b/packages/hooks/src/useMustNotChange.test.ts
@@ -4,15 +4,24 @@
 
 import { renderHook } from "@testing-library/react-hooks";
 
+import Logger from "@foxglove/log";
+
 import { useMustNotChangeImpl } from "./useMustNotChange";
 
 describe("useMustNotChange", () => {
-  it("should throw when value changes", () => {
-    const { result, rerender } = renderHook((val) => useMustNotChangeImpl(val), {
+  it("should log an error when value changes", () => {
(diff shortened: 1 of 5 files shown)


defunctzombie:
> It will trigger the panel settings to re-render more often than it needs to

How will it do that? During rendering we will be getting the correct config will we not?

> @defunctzombie should we just remove the unstable selector error

Do you mean the `value must not change` error? I would be fine getting rid of it - the general idea behind the error is to help find places where the selector value is unstable (i.e. changing every render) since the selector is run when it is changed.

jtbandes:
> > It will trigger the panel settings to re-render more often than it needs to
>
> How will it do that? During rendering we will be getting the correct config will we not?

Yes, but it now renders whenever the `configById` changes, not just whenever `configById[selectedId]` changes.

> Do you mean the `value must not change` error? I would be fine getting rid of it - the general idea behind the error is to help find places where the selector value is unstable (i.e. changing every render) since the selector is run when it is changed.

Maybe we should use `useShouldNotChangeOften` here instead?

foxymiles:
Maybe we should make this just log an error instead of actually throwing an exception?

defunctzombie (approved):

PR #2938, opened 2022-03-01
Default to development backend routes in development

defunctzombie:
**User-Facing Changes**
None.

**Description**
Use the webpack _mode_ as the default value for determining which url routes to use. This changes the default routes in development to the development routes for web and desktop.

FOXGLOVE_BACKEND can still override the mode.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,9 +69,6 @@ $ yarn start        # launch electron
 # To launch the browser app:
 $ yarn web:serve
 
-# To launch the browser app using the dev backend server:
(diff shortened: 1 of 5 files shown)


defunctzombie:
One tradeoff here is that the desktop app also uses the `dev` endpoints rather than the production ones. The desktop app could use the production ones but for consistency it seems reasonable to have web and desktop development match.

foxymiles (approved):

Issue #2916, opened 2022-02-25
Node playground type checking failing on start when loading from data platform


ccpowers:
**Description**
 When loading a new bag from the data platform, node playground incorrectly reports errors on types for input topics. Below is an example using node playground to transform `geometry_msgs/PoseArray` messages on topic `/particlecloud` into `visualization_msgs/MarkerArray` messages on topic `/studio_node/particle_cloud`. 
 When loading a bag from disk with these messages, the node playground reports no errors. If the same bag is uploaded into data platform and opened in studio via data platform, the errors shown below are reported. If the node playground is edited and re-saved (adding trivial changes to comments, or other irrelevant changes), the errors disappear. 
 
![Screen Shot 2022-02-25 at 11 04 24 AM](https://user-images.githubusercontent.com/477928/155748240-fb0eeafa-74a5-484d-98e2-07ca25985968.png)

- Version: In browser (chrome)
- OS: The browser

**Steps To Reproduce**
1. Add a node playground panel with the following node included:
```
// The ./types module provides helper types for your Input events and messages.
import { Input, Message } from "./types";

// Your  node can output well-known message types, any of your custom message types, or
// complete custom message types.
//
// Use `Message` to access your data source types or well-known types:
// type Twist = Message<"geometry_msgs/Twist">;
//
// Conventionally, it's common to make a _type alias_ for your node's output type
// and use that type name as the return type for your node function.
// Here we've called the type `Output` but you can pick any type name.
type MarkerArray = Message<"visualization_msgs/MarkerArray">;
type Marker = Message<"visualization_msgs/Marker">;
type Pose = Message<"geometry_msgs/Pose">;
type PoseArray = Message<"geometry_msgs/PoseArray">;
type Header = Message<"std_msgs/Header">;

// These are the topics your node "subscribes" to. Studio will invoke your node function
// when any message is received on one of these topics.
export const inputs = ["/particlecloud"];

// Any output your node produces is "published" to this topic. Published messages are only visible within Studio, not to your original data source.
export const output = "/studio_node/particle_cloud";

// This function will transform one pose into one marker
const scale: Message<"geometry_msgs/Vector3"> = {
  x: 0.01,
  y: 0.01,
  z: 0.01,
};

const lifetime = { sec: 0.02, nsec: 0 };

const color: Message<"std_msgs/ColorRGBA"> = {
  r: 100,
  g: 0,
  b: 0,
  a: 100,
};
[...]

PR #2939, opened 2022-03-01
Reset node playground workers when topics change.

defunctzombie:
**User-Facing Changes**
Node playground nodes are updated with generated types as the datasource loads new topics and datatypes.

**Description**
Node playground nodes generate code for messages and inputs from the set of available topics and datatypes. This code generation would run once on startup. If the generation happened to run before the data source provided the topics and datatypes, the user would see compilation errors in their nodes and the nodes would not run.

This change re-generates the nodes when the data source topics change. This allows the node code to compile and the nodes to start running as the data source loads available topics and datatypes.

Fixes: #2916

### DIFF
--- a/packages/studio-base/src/players/UserNodePlayer/index.ts
+++ b/packages/studio-base/src/players/UserNodePlayer/index.ts
@@ -653,7 +653,7 @@ export default class UserNodePlayer implements Player {
           this._nodeRegistrationCache = [];
         }
 
-        this._lastPlayerStateActiveData = playerState.activeData;
+        this._lastPlayerStateActiveData = activeData;
         if (shouldReset) {
           await this._resetWorkers();
         }
--- a/packages/studio-base/src/players/UserNodePlayer/index.ts
+++ b/packages/studio-base/src/players/UserNodePlayer/index.ts
@@ -112,6 +112,12 @@ export default class UserNodePlayer implements Player {
   // a node may set its own problem or clear its problem
   private _problemStore = new Map<string, PlayerProblem>();
 
+  private _nodeRegistrationCache: {
+    nodeId: string;
+    userNode: UserNode;
+    result: NodeRegistration;
+  }[] = [];
+
   // exposed as a static to allow testing to mock/replace
   static CreateNodeTransformWorker = (): SharedWorker => {
     return new SharedWorker(new URL("./nodeTransformerWorker/index", import.meta.url), {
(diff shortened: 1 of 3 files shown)


defunctzombie:
I've removed the caching of node registrations as that was another layer that prevented re-building nodes when they should be rebuilt. It is true that nodes will be re-built when seeking or when topics change, but that seems like it happens quickly enough that its not a massive impact. We can introduce better "reset" behavior and caching if we do re-visit this code but it felt better to have things working as a tradeoff.

packages/studio-base/src/players/UserNodePlayer/index.ts
```diff
@@ -644,6 +623,17 @@ export default class UserNodePlayer implements Player {
         await this._resetWorkers();
         this.setSubscriptions(this._subscriptions);
         this.requestBackfill();
+      } else {
+        // Reset node state after seeking
+        let shouldReset = activeData.lastSeekTime !== this._lastPlayerStateActiveData.lastSeekTime;
+
+        // reset if topics change
+        shouldReset ||= activeData.topics !== this._lastPlayerStateActiveData.topics;
```
jtbandes, line 631: Should we also check `datatypes`?

jtbandes (approved):
> It is true that nodes will be re-built when seeking or when topics change, but that seems like it happens quickly enough that its not a massive impact.

What about when scrubbing quickly (dragging the slider)? Seems like rebuilding nodes on every frame is probably a huge perf impact.

defunctzombie:
I've thought about this and agree that re-compiling is probably not ideal. I'll bring back the caching and see if clearing it when topics changes is sufficient. If not, I might change the caching altogether since right now it looks like it stores old versions of code as well as new versions which seems unnecessary (rebuild on undo doesn't seem terrible).

defunctzombie:
@jtbandes I reworked the logic to clear the cache when topics and datatypes change. I've left the cache in place otherwise.

packages/studio-base/src/players/UserNodePlayer/index.ts
```diff
@@ -644,6 +640,23 @@ export default class UserNodePlayer implements Player {
...
+        // Reset node state after seeking
+        let shouldReset = activeData.lastSeekTime !== this._lastPlayerStateActiveData.lastSeekTime;
+
+        // When topics or datatypes change we also need to re-build the nodes so we clear the cache
+        if (
+          activeData.topics !== this._lastPlayerStateActiveData.topics ||
+          activeData.datatypes !== this._lastPlayerStateActiveData.datatypes
+        ) {
+          shouldReset ||= true;
+          this._nodeRegistrationCache = [];
```
jtbandes, line 653: It was initially confusing why this isn't under `shouldReset` but I see now that there are 2 levels of resetting: just resetting the node execution, vs re-compiling the nodes.

jtbandes (approved):

PR #2940, opened 2022-03-01
Fix ErrorDisplay scrolling and styling tweaks

### LINKED ISSUE #2931: Error boundary is not scrollable
When an error appears, if the height of the error boundary is too short, the error is not visible, but it's impossible to scroll down and see the error:

image

When you increase the height, the error becomes visible:

image

jtbandes:
**User-Facing Changes**

**Description**
- Fix #2931 by making error boundary scrollable
- Use monospace font for error message
- Add a background to the error stack
- Fix the `cursor` style on links
- Remove some extraneous styles & elements

### DIFF
--- a/packages/studio-base/src/.storybook/preview.tsx
+++ b/packages/studio-base/src/.storybook/preview.tsx
@@ -98,8 +98,15 @@ function WithContextProviders(Child: Story, ctx: StoryContext): JSX.Element {
       }
 
       {(colorScheme === "light" || colorScheme.startsWith("both")) && (
-        // Set a transform to make this the root for position:fixed elements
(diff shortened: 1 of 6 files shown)


foxymiles (approved):

jtbandes:
IMO the monospace font conveys that the text is coming from some internal part of the app, and isn't a regular part of the UI. Since the error display doesn't say something like "error message:" but does have a lot of other preamble text, it's a little hard to visually parse which part is the actual error, so that's why I changed the font. Gonna merge for now but don't feel too strongly, if anyone wants to make their own PR to tweak it in a different way that's fine too.

PR #2941, opened 2022-03-01
Fix crashes resulting from negative hover times

jtbandes:
**User-Facing Changes**
Fixed a crash when hovering or clicking on a negative timestamp in the Plot panel.

**Description**
When hovering on a negative x-axis value in the plot, if the hovered absolute time (start time + negative elapsed time) was <0, the app would crash when attempting to normalize the time value for display in the playback bar. Once that was fixed, it would crash when _clicking_ on the negative time.

This issue was particularly noticeable when using "simulated" timestamp values, i.e. data which starts from time 0, because you don't have to pan/zoom very far to be able to access negative plot ranges that trigger the crash.

image

### DIFF
--- a/packages/studio-base/src/panels/Plot/index.tsx
+++ b/packages/studio-base/src/panels/Plot/index.tsx
@@ -423,9 +423,10 @@ function Plot(props: Props) {
       if (!seekPlayback || !start || seekSeconds == undefined || xAxisVal !== "timestamp") {
         return;
       }
-      // The player validates and clamps the time.
-      const seekTime = addTimes(start, fromSec(seekSeconds));
-      seekPlayback(seekTime);
+      // Avoid normalizing a negative time if the clicked point had x < 0.
+      if (seekSeconds >= 0) {
+        seekPlayback(addTimes(start, fromSec(seekSeconds)));
+      }
     },
     [messagePipeline, xAxisVal],
   );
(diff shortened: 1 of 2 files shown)


foxymiles (approved):

packages/studio-base/src/panels/Plot/index.tsx
```diff
@@ -423,9 +423,10 @@ function Plot(props: Props) {
       if (!seekPlayback || !start || seekSeconds == undefined || xAxisVal !== "timestamp") {
         return;
       }
-      // The player validates and clamps the time.
-      const seekTime = addTimes(start, fromSec(seekSeconds));
-      seekPlayback(seekTime);
+      // Avoid normalizing a negative time if the clicked point had x < 0.
+      if (seekSeconds >= 0) {
+        seekPlayback(addTimes(start, fromSec(seekSeconds)));
```
defunctzombie, line 428: Another thought could be to use `clampTime` here so that clicking into negative seconds clamps you to the start time of 0.

And another thought - should the plot panel prevent pan/zoom that takes time negative?

jtbandes: I considered clamp, but addTimes throws an error before we get a chance to clamp.

> should the plot panel prevent pan/zoom that takes time negative

I don't feel like we should prevent people from panning so that negative times are _visible_ on the plot, even if there won't be any data there. The same way we allow zooming in/out on the image panel beyond the strictly necessary bounds. And once the negative times are visible, I think it's weird if the yellow hover line doesn't appear there, so I felt like it made sense to keep the negative hover value, and just fix the things that broke when it was negative.

Issue #2895, opened 2022-02-23
Support color by intensity for sensor_msgs/LaserScan


ccpowers:
I would like to display messages of type `sensor_msgs/LaserScan` with color determined by their intensity field, similar to RVIZ.
![Screen Shot 2022-02-23 at 8 30 49 AM](https://user-images.githubusercontent.com/477928/155330051-3336f4d9-e2dc-4b5b-aab1-48ae0707913b.png)

The specific use case i'm interested in is setting all points in the scan that are above a certain intensity threshold to one color, and all points below the threshold to another color. This can be achieved in RVIZ by setting the `min intensity` and `max intensity` to `threshold` and `threshold + 1`, and then choosing the min and max colors to display. 
A secondary use case is to display the intensity of each point in the scan transformed across a gradient, so that the relative intensity of each point can be visualized.

Thank you!

jhurliman:
Fixed in https://github.com/foxglove/studio/pull/2952

PR #2952, opened 2022-03-01
Render sensor_msgs/LaserScan by internally converting to sensor_msgs/PointCloud2

jhurliman:
**User-Facing Changes**

- `sensor_msgs/LaserScan` topics in the 3D panel have the same settings and rendering as `sensor_msgs/PointCloud2`

**Description**

Instead of using a separate rendering path for laser scans, we now internally convert laser scan messages to point clouds. This allows the settings editor and rendering code for point clouds to be reused.


packages/studio-base/src/panels/ThreeDimensionalViz/SceneBuilder/index.ts
```diff
@@ -611,6 +612,58 @@ export default class SceneBuilder implements MarkerProvider {
...
+      view.setFloat64(offset + 16, 0, true);
+      view.setFloat64(offset + 24, distance, true);
+      view.setFloat64(offset + 32, angle, true);
+      if (hasIntensity) {
+        view.setFloat64(offset + 40, intensity, true);
+      }
+    }
+
+    const fields: PointField[] = [
+      { name: "x", offset: 0, datatype: 8, count: 1 },
```
jtbandes, line 642: I think we have constants for these `datatype` values somewhere in the code, that would make it easier to read

packages/studio-base/src/panels/ThreeDimensionalViz/SceneBuilder/index.ts
```diff
@@ -611,6 +612,58 @@ export default class SceneBuilder implements MarkerProvider {
...
+    const pointStep = hasIntensity ? 48 : 40;
+
+    const data = new Uint8Array(pointStep * scan.ranges.length);
+    const view = new DataView(data.buffer);
+    let angle = scan.angle_min;
+    for (let i = 0; i < scan.ranges.length; i++) {
+      const offset = i * pointStep;
+      const distance = Math.min(scan.range_max, Math.max(scan.range_min, scan.ranges[i] ?? 0));
+      const intensity = scan.intensities[i] ?? Number.NaN;
+      angle = Math.min(scan.angle_max, angle + scan.angle_increment);
```
jtbandes, line 627: I usually prefer `angle = min + i * inc` rather than `angle += inc` to avoid compounding precision issues, though it probably doesn't matter much here

jtbandes (approved):
Seems like a decent approach to adding more features to LaserScan rendering without much work.

defunctzombie:
Curious to know what you learn about the performance implications of this. My understanding is that previously we had a regl command handle the laserscan message directly? This would be faster than doing a data conversion that we are doing in JS.

Is this PR in response to a bugfix or feature request? Maybe that has some sample data we can test with. On the surface I am hesitant of the impact of this on rendering for a large laser scan but if we can get a sense of the impact that would be good.

defunctzombie:
#2895

jhurliman:
This is a big improvement for anyone using `sensor_msgs/LaserScan`. I'm inclined to merge this since I would prefer the additional features at the expense of reduced performance if I was trying to debug a laser range-finding robot, and there are paths to improving performance if needed (build a vertex array and vertex attribute buffers instead of this intermediate format).

packages/studio-base/src/panels/ThreeDimensionalViz/SceneBuilder/index.ts
jhurliman: Oof, my implementation was incorrect since scan 0 was `angle_min + angle_increment` instead of `angle_min`.

PR #2954, opened 2022-03-02
Quick Look preview visual updates for small viewports

jtbandes:
**User-Facing Changes**
Improved macOS Quick Look preview when displayed at small sizes (such as in the Get Info window in Finder).

**Description**

- Added `yarn serve:quicklook` for faster development
- Added styled-components transformer (which we already using in our other webpack builds) for better CSS class names
- Added breakpoint for small sizes
- Remove extra table padding at small sizes
- Use some non-breaking spaces to avoid breaking in weird places
- Fixed filtering of `compression` values so that `""` displays as `Compression: none`

Before:

image

After:

Screen Shot 2022-03-01 at 4 59 03 PM Screen Shot 2022-03-01 at 4 59 13 PM

### DIFF
--- a/desktop/quicklook/FileInfoDisplay.tsx
+++ b/desktop/quicklook/FileInfoDisplay.tsx
@@ -2,7 +2,7 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
-import { useEffect } from "react";
+import { useEffect, useMemo } from "react";
 import styled from "styled-components";
 
 import Logger from "@foxglove/log";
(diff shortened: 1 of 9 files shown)


desktop/quicklook/FileInfoDisplay.tsx
```diff
@@ -50,6 +51,10 @@ const IconContainer = styled.div`
   display: flex;
   flex-direction: column;
   align-items: center;
+
+  @media (max-width: ${styleConstants.breakpoints.narrowMaxWidth}) {
```
defunctzombie, line 55: Quicklook should probably use its own set of breakpoints specific to its needs? Or a semantic name specific to quicklook?

defunctzombie (approved):

desktop/quicklook/FileInfoDisplay.tsx
jtbandes: Not sure I understand the suggestion; this styleConstants does live inside the quicklook folder.

defunctzombie: I missed that - somehow thought it was coming from a style constants in studio-base. I now see I confused that with `sharedStyleConstants.ts` 🤦

jtbandes: shared, but not _too_ shared

PR #2955, opened 2022-03-02
Build Stack component without use of sx prop

2metres:
**Description**
Hey the Stack component was great. But… `sx` prop. Here is an alternate version using the `styled` from material-ui without using the sx prop


foxymiles (approved):
I imagine we might need to tweak this a bit as we start using it but LGTM.

jtbandes:
[RIP Stack, long live Flex](https://github.com/foxglove/studio/pull/2725)

PR #2958, opened 2022-03-02
Add DialogHostIdContext for controlling modal attachment

bclinkinbeard:
Allows users to specify the id of a DOM element to be used as the parent element for modals.

**User-Facing Changes**
None.

**Description**
Enables the inclusion of a provider to control which DOM node to which modals will be appended. If not provided, the default behavior of being added to the end of the document is retained.

Sample: <DialogHostIdContext.Provider value={"my-cool-div"} />

### DIFF
--- a/packages/studio-base/src/context/DialogHostIdContext.ts
+++ b/packages/studio-base/src/context/DialogHostIdContext.ts
@@ -1,0 +1,15 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { createContext, useContext } from "react";
+
+const DialogHostIdContext = createContext<string | undefined>(undefined);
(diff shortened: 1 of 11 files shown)


packages/studio-base/src/context/DialogHostIdContext.ts
```diff
@@ -0,0 +1,15 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { createContext, useContext } from "react";
+
+const DialogHostIdContext = createContext<string | undefined>(undefined);
+DialogHostIdContext.displayName = "DialogHostIdContext";
+
+function useDialogHostId(): string | undefined {
+  return useContext(DialogHostIdContext);
+}
```
jtbandes, line 12: Can you add a comment around why we need this context?

bclinkinbeard: Updated

jtbandes (approved):

PR #2962, opened 2022-03-03
Fix nondeterministic ErrorBoundary stories

jtbandes:
**User-Facing Changes**
None

**Description**
This restores some logic that was removed in https://github.com/foxglove/studio/pull/2021. 

Since #2940, the storybook screenshots became flaky because file hashes and the chromatic URL would appear in screenshots ([example](https://www.chromatic.com/test?appId=603ec8bf7908b500231841e2&id=62201256deddcb003a991eb7)). The restored logic strips out these locations in storybook to make the screenshots deterministic.

### DIFF
--- a/packages/studio-base/src/components/ErrorBoundary.stories.tsx
+++ b/packages/studio-base/src/components/ErrorBoundary.stories.tsx
@@ -47,7 +47,7 @@ export const Default: Story = () => {
 export const ShowingDetails: Story = () => {
   return (
     <DndProvider backend={HTML5Backend}>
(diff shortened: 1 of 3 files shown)


defunctzombie:
In stories we throw errors with a modified stack so we can avoid this modification in the component. Did we stop doing that?

jtbandes:
If you look at the story being changed, you can see the modified stack is the "Error stack" portion of the stack trace. We do not control the "Component stack" information which is added automatically by React.

image

defunctzombie (approved):

Issue #2929, opened 2022-02-28
Misleading error when failing to load from the package path


defunctzombie:
**Description**
When a `package://` url fails to load, I get the following error. I've already set the `ROS_PACKAGE_PATH` so I'm not sure what else I should do. Ideally the error should tell me something more informative.

I also don't understand what `x-foxglove-ros-package:?targetPkg=cad&basePath=&relPath=/body_midres.stl` means since thats not a path I've put anywhere in my urdf files.

```
Failed to load robot_v1.urdf: Failed to load x-foxglove-ros-package:?targetPkg=cad&basePath=&relPath=/body_midres.stl. Loading assets from ROS packages requires the ROS_PACKAGE_PATH environment variable to be set.
```

PR #2970, opened 2022-03-03
Change package:// url handling to use package: protocol

defunctzombie:
**User-Facing Changes**
When loading a resource via `package://` urls fails, the user is shown an error that contains the failed `package://` url rather than a `x-foxglove-ros-package` url.

Loading 05-visual in Studio v1.4.0 from urdf_tutorials

image

With this PR

image

image

**Description**
This change removes the package:// url re-writing in favor of trying to load the package:// url using desktop protocol handlers. This preserves the urls as package:// urls which makes them show up as package:// urls in error messages.

For TIFF files (which are not supported by the Theejs ColladaLoader natively, we convert package:// urls into x-foxglove-converted-tiff:// urls. If there is an error loading the .tiff file we show the unconverted url path to the user to avoid showing the x-foxglove-converted-tiff urls.

Fixes: #2929

----
TODO

- [x] finish TIFF loader
- [x] Test urdf with TIFF files
- [x] Test loading other kinds of URDF
- [x] Test 3d panel model loading
- [x] Remove rewritePackageUrl


packages/studio-base/src/services/URDFAssetLoader.ts
```diff
@@ -13,11 +15,39 @@ import {
...
+  constructor(manager: THREE.LoadingManager) {
+    super(manager);
+  }
+
+  override async loadAsync(url: string): Promise<HTMLImageElement> {
+    const image = document.createElement("img");
+
+    const res = await fetch(url);
+    const blob = await res.blob();
+    image.src = URL.createObjectURL(blob);
```
jtbandes, line 34: I don't think this works with TIFF images in Chrome.

jtbandes (commented):
It seems like this PR is removing the `basePath` functionality. It's been a while since I worked on this (https://github.com/foxglove/studio/pull/617), but I believe this `basePath` was the reason for rewriting URLs. Specifically, the `basePath` itself can only be known in the renderer (it's the `.path` that Electron exposes on the File instance), but other files under sibling paths can only be _read_ from the preload/main process.

The `basePath` is another search location for ROS packages based on the location of the dragged/opened file. This improves the experience for users who have not configured a ROS_PACKAGE_PATH variable, such as those using Foxglove Studio on a separate machine from their ROS instance who have just downloaded some URDF files locally.

defunctzombie:
> The `basePath` is another search location for ROS packages based on the location of the dragged/opened file. This improves the experience for users who have not configured a ROS_PACKAGE_PATH variable, such as those using Foxglove Studio on a separate machine from their ROS instance who have just downloaded some URDF files locally.

I see two approaches:

1. I could re-write to append the basePath as a query argument to the package:// path (only if ROS_PACKAGE_PATH is not set maybe?)

2. Keep the `x-foxglove-*` urls and when a loading failure happens - re-write the url back to a `package://` for the error display.

I'm not convinced that we should have added this automatic lookup of package paths from the dragged file because `package://` paths expect some level of setup with ROS_PACKAGE_PATH and now we've created an additional lookup mechanism that doesn't exist in ROS tooling that understands package:// paths - at least that is my understanding.

jhurliman:
> I'm not convinced that we should have added this automatic lookup of package paths from the dragged file...

Dragged files (either .urdf or .xacro files in this discussion) often include references to other files. Those references can be relative file paths (`./meshes/turtle.dae`) or package URLs (`package://turtlebot/meshes/turtle.dae`). To make dragging and dropping a URDF successfully load all the referenced 3D models, we need to resolve package:// URLs to absolute filesystem paths. In ROS1, this is done with an environment variable called `ROS_PACKAGE_PATH`. Studio supports this environment variable for a "works out of the box" experience if a ROS1 environment has been configured in the shell that is launching Studio. Alternatively, since many users launch Studio from a desktop launcher that does not have a ROS1 environment configured or run Studio on a computer that does not have a ROS environment configured at all, we allow users to override `ROS_PACKAGE_PATH` as an application setting in Studio.

I don't think anyone is tied to the implementation details of how to make the above work, as long as we don't regress the existing functionality of:
 * Properly loading a URDF and all referenced resources without any configuration if the ROS_PACKAGE_PATH environment variable is correctly configured
 * Properly loading a URDF and all referenced resources if the ROS_PACKAGE_PATH app setting has been correctly configured

jhurliman:
On this PR, I don't think showing the user a `package://` URL is the best path. There are two flavors of error here:

1. `ROS_PACKAGE_PATH` is empty, meaning both the environment variable was unset and the Studio app setting was unset (or there is a bug in Studio). In this case, we can instruct the user exactly what they need to do (set the ROS_PACKAGE_PATH app setting).
2. An absolute filesystem path was constructed from the `package://` URL and `ROS_PACKAGE_PATH`, but we failed to read the file at that path (most likely the path does not exist or we don't have permission, less likely there was a read error like an external drive being unplugged or media failure). In this case, the user should be shown the absolute path that failed to load, and optionally the `package://` URL and `ROS_PACKAGE_PATH` value to help them troubleshoot whether the package path is wrong, the urdf file containing the package:// URL is wrong, or they are missing a required package or file from their computer.

packages/studio-base/src/services/URDFAssetLoader.ts
defunctzombie: Removed - had to go back to the special protocol handler for converting tiff files due to limitations with how loaders are configured in URDFLoader and Threejs with no ability to override specific file handling.

jtbandes (approved):

PR #2972, opened 2022-03-04
Fetch full topics in data platform player.

foxymiles:
**User-Facing Changes**
This improves loading behavior of plot data streaming from data platform.

**Description**
The idea here is to support two different kinds of subscriptions. One fetches the entire time range for use in panels like the plot panel. The other fetches data on demand for panels that show point in time data like the image panel.

The key change is to add a `range` field to `SubscriptionPayload` that can be either "full" or "partial". These are hints to the player that guide fetching behavior. A range of "full" indicates that a subscriber wants data for the full time range of the data source. "partial" indicates that data can be fetched as needed for the current time. Higher level looks like `useBlocksByTopic` or `useMessagesByTopic` will choose the appropriate value for `range` so most subscribers will not need to specify `range` explicitly.

### DIFF
--- a/packages/studio-base/src/PanelAPI/useBlocksByTopic.ts
+++ b/packages/studio-base/src/PanelAPI/useBlocksByTopic.ts
@@ -26,6 +26,7 @@ import {
   SubscribePayload,
   MessageEvent,
   MessageBlock as PlayerMessageBlock,
+  SubscriptionRange,
 } from "@foxglove/studio-base/players/types";
 
 export type MessageBlock = {
@@ -57,7 +58,10 @@ const filterBlockByTopics = memoizeWeak(
   },
 );
 
-const useSubscribeToTopicsForBlocks = (topics: readonly string[]) => {
+const useSubscribeToTopicsForBlocks = (
+  topics: readonly string[],
+  range: SubscriptionRange = "partial",
+) => {
   const [id] = useState(() => uuidv4());
   const { type: panelType = undefined } = useContext(PanelContext) ?? {};
 
@@ -69,9 +73,10 @@ const useSubscribeToTopicsForBlocks = (topics: readonly string[]) => {
     ),
   );
   const subscriptions: SubscribePayload[] = useMemo(() => {
-    const requester = panelType != undefined ? { type: "panel", name: panelType } : undefined;
-    return topics.map((topic) => ({ topic, requester } as SubscribePayload));
-  }, [panelType, topics]);
+    const requester: SubscribePayload["requester"] =
+      panelType != undefined ? { type: "panel", name: panelType } : undefined;
+    return topics.map((topic) => ({ topic, requester, range }));
+  }, [range, panelType, topics]);
   useEffect(() => setSubscriptions(id, subscriptions), [id, setSubscriptions, subscriptions]);
   useCleanup(() => setSubscriptions(id, []));
 };
@@ -89,10 +94,13 @@ const useSubscribeToTopicsForBlocks = (topics: readonly string[]) => {
 //   - Each block represents the same duration of time
 //   - The number of blocks is consistent for the data source
 //   - Blocks are stored in increasing order of time
-export function useBlocksByTopic(topics: readonly string[]): readonly MessageBlock[] {
+export function useBlocksByTopic(
+  topics: readonly string[],
+  range: SubscriptionRange = "partial",
+): readonly MessageBlock[] {
   const requestedTopics = useShallowMemo(topics);
 
-  useSubscribeToTopicsForBlocks(requestedTopics);
+  useSubscribeToTopicsForBlocks(requestedTopics, range);
 
   const allBlocks = useMessagePipeline<readonly (PlayerMessageBlock | undefined)[] | undefined>(
     useCallback((ctx) => ctx.playerState.progress.messageCache?.blocks, []),
--- a/packages/studio-base/src/PanelAPI/useBlocksByTopic.ts
+++ b/packages/studio-base/src/PanelAPI/useBlocksByTopic.ts
@@ -71,7 +71,7 @@ const useSubscribeToTopicsForBlocks = (topics: readonly string[]) => {
   const subscriptions: SubscribePayload[] = useMemo(() => {
     const requester: SubscribePayload["requester"] =
       panelType != undefined ? { type: "panel", name: panelType } : undefined;
-    return topics.map((topic) => ({ topic, requester, range: "full" }));
+    return topics.map((topic) => ({ topic, requester, preloadType: "full" }));
   }, [panelType, topics]);
   useEffect(() => setSubscriptions(id, subscriptions), [id, setSubscriptions, subscriptions]);
   useCleanup(() => setSubscriptions(id, []));
--- a/packages/studio-base/src/PanelAPI/useBlocksByTopic.ts
+++ b/packages/studio-base/src/PanelAPI/useBlocksByTopic.ts
@@ -26,7 +26,6 @@ import {
   SubscribePayload,
   MessageEvent,
   MessageBlock as PlayerMessageBlock,
-  SubscriptionRange,
 } from "@foxglove/studio-base/players/types";
 
 export type MessageBlock = {
@@ -58,10 +57,7 @@ const filterBlockByTopics = memoizeWeak(
   },
 );
 
-const useSubscribeToTopicsForBlocks = (
-  topics: readonly string[],
-  range: SubscriptionRange = "partial",
-) => {
+const useSubscribeToTopicsForBlocks = (topics: readonly string[]) => {
   const [id] = useState(() => uuidv4());
   const { type: panelType = undefined } = useContext(PanelContext) ?? {};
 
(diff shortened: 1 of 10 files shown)


packages/studio-base/src/PanelAPI/useBlocksByTopic.ts
```diff
@@ -86,10 +94,13 @@ const useSubscribeToTopicsForBlocks = (topics: readonly string[]) => {
 //   - Each block represents the same duration of time
 //   - The number of blocks is consistent for the data source
 //   - Blocks are stored in increasing order of time
-export function useBlocksByTopic(topics: readonly string[]): readonly MessageBlock[] {
+export function useBlocksByTopic(
+  topics: readonly string[],
+  range: SubscriptionRange = "partial",
```
defunctzombie, line 99: What does partial mean here? A user of useBlocksByTopic is already saying they want the full ranges. I don't think this should be an argument to the hook.

packages/studio-base/src/PanelAPI/useMessagesByTopic.ts
```diff
@@ -32,9 +32,10 @@ type UnknownMessageEventsByTopic = Record<string, readonly MessageEvent<unknown>
  */
 export function useMessagesByTopic(params: {
   topics: readonly string[];
+  range?: SubscriptionRange;
```
defunctzombie, line 35: This is implied by useMessagesByTopic so there's no need to expose this as an argument for the hook.

packages/studio-base/src/players/types.ts
```diff
@@ -241,13 +241,19 @@ export type Progress = Readonly<{
   readonly messageCache?: BlockCache;
 }>;
 
+export type SubscriptionRange =
+  | "full" // Fetch nessages for the entire content range.
+  | "partial"; // Fetch messages as needed.
```
defunctzombie, line 246: I think the two semantics we have are:

1. What you call _full_ here - which is to hint that the topic should be loaded into "blocks" and available via the blocks player state interface
2. What you call _partial_ here is actually more like "playback", "current frame" and just means you care to receive the messages during playback for the current playhead. I think we can find a better word than partial for this - or just default to that mode and have a boolean to indicate the other mode.

foxymiles: Yeah I'm not really happy with the naming here either. Open to better suggestions.

foxymiles: Although having this as an enum type instead of a boolean does make the logic in the player a bit more transparent IMO.

packages/studio-base/src/PanelAPI/useMessagesByTopic.ts
foxymiles: Right good point I'll change it.

packages/studio-base/src/players/FoxgloveDataPlatformPlayer/index.ts
```diff
@@ -355,36 +368,53 @@ export default class FoxgloveDataPlatformPlayer implements Player {
...
+    if (!this._caches || this._closed) {
+      return;
+    }
+
+    if (this._currentPreloadTasks[loadRange]) {
+      return;
+    }
+
+    const topics = chain(this._subscriptions)
+      .filter((s) => (loadRange === "full" ? s.range === "full" : s.range !== "full"))
```
jtbandes, line 392:
```suggestion
      .filter((s) => s.range === loadRange)
```

packages/studio-base/src/players/FoxgloveDataPlatformPlayer/index.ts
```diff
@@ -328,24 +336,29 @@ export default class FoxgloveDataPlatformPlayer implements Player {
...
+          await (this._loadedMoreMessages = signal());
+          if (this._lastSeekTime !== lastSeekTime) {
+            lastTickEndTime = undefined;
+            continue mainLoop;
+          }
         }
+        lastTickEndTime = performance.now();
+        this._nextFrame = messages;
+      } else {
+        this._nextFrame = [];
```
jtbandes, line 360: Would it be better to never create a new empty array (not triggering unnecessary updates downstream) if there are no "full" topics?

packages/studio-base/src/players/FoxgloveDataPlatformPlayer/index.ts
```diff
@@ -438,14 +468,25 @@ export default class FoxgloveDataPlatformPlayer implements Player {
...
 
     this._emitState();
   }
 
+  private _updateProgress() {
+    const noFullTopics = this._subscriptions.every((s) => s.range !== "full");
+    const cache = noFullTopics ? this._caches?.partial : this._caches?.full;
+    if (cache) {
+      this._progress = {
+        fullyLoadedFractionRanges: cache.fullyLoadedFractionRanges(),
```
jtbandes, line 484: Perhaps we should consider exposing both full and partial ranges on the progress bar? Might need some input from @2metres on how to do this. But we could start by exposing them both on `progress`.

packages/studio-base/src/players/types.ts
jtbandes: I think I might use an `enum` here instead of string literal types, just because it feels messy to be using the literals everywhere (like `startTaskIfNeeded("full")` looks a little weird to me)

I feel like we could come up with a better name but I'm not really sure. Maybe `SubscriptionPreloadType`?

packages/studio-base/src/players/FoxgloveDataPlatformPlayer/index.ts
foxymiles: The issue here I was trying to avoid is that range is optional and might be undefined.

packages/studio-base/src/players/FoxgloveDataPlatformPlayer/index.ts
foxymiles: Yeah I think eventually this is something we might want to expose to the user.

packages/studio-base/src/players/FoxgloveDataPlatformPlayer/index.ts
jtbandes: Should we make it non-optional? Or use `s.range == undefined || ...` or `(s.range ?? "default") === ...` ?

packages/studio-base/src/players/FoxgloveDataPlatformPlayer/index.ts
foxymiles: Yeah I was a little fuzzy on what should happen here. Should I just leave `nextFrame` alone in this case?

packages/studio-base/src/players/FoxgloveDataPlatformPlayer/index.ts
foxymiles: I tried making it non-optional but that required changing a lot of test code that doesn't put something in the subscription payload. Maybe comparison to `undefined` would be clearer?

foxymiles: Actually I rewrote this to partition the subscriptions at the time that they're set.

packages/studio-base/src/players/types.ts
foxymiles: > I think I might use an `enum` here instead of string literal types, just because it feels messy to be using the literals everywhere (like `startTaskIfNeeded("full")` looks a little weird to me)
> 
> I feel like we could come up with a better name but I'm not really sure. Maybe `SubscriptionPreloadType`?

The other term I considered for this was `strategy` or `loadStrategy`.

packages/studio-base/src/PanelAPI/useMessageReducer.test.tsx
```diff
@@ -279,8 +279,14 @@ describe("useMessageReducer", () => {
       root.unmount();
     });
     expect(setSubscriptions.mock.calls).toEqual([
-      [expect.any(String), [{ topic: "/foo" }]],
-      [expect.any(String), [{ topic: "/foo" }, { topic: "/bar" }]],
+      [expect.any(String), [{ topic: "/foo", range: "partial", requestor: undefined }]],
```
jtbandes, line 282: typo: requester vs requestor?

packages/studio-base/src/players/FoxgloveDataPlatformPlayer/index.ts
jtbandes: I would try that and see if anything breaks? Perhaps it will...

packages/studio-base/src/players/FoxgloveDataPlatformPlayer/index.ts
```diff
@@ -355,36 +374,48 @@ export default class FoxgloveDataPlatformPlayer implements Player {
...
+    const topics = uniq(this._subscriptions[loadRange].map((s) => s.topic));
+    if (topics.length === 0) {
       return;
     }
-    const preloadedMessages = this._initialized.preloadedMessages;
-    const nextRangeToLoad = preloadedMessages.nextRangeToLoad(this._currentTime);
+
+    const nextRangeToLoad = preloadedMessages.nextRangeToLoad(
+      loadRange === "full" ? this._start : this._currentTime,
+    );
```
jtbandes, line 404: So I guess if full preloading is requested, we start loading from the beginning rather than the current time? It still seems like it might be a good idea to prioritize loading near the current playback position, but I guess we haven't really polled users on this.

jtbandes (approved):

PR #2973, opened 2022-03-04
Move MessageBlock and BlockCache types to player/types

defunctzombie:
**User-Facing Changes**
None

**Description**
These types are used within the PlayerState and should exist in the player types file rather than the random access files so they can be used by players other than random access.


foxymiles (approved):
Good idea. @jtbandes also suggested doing this.

PR #2975, opened 2022-03-05
Remove READ_DELAY from CachedFileLike

defunctzombie:
**User-Facing Changes**
None

**Description**
Chrome dev tools as a builtin feature to slow down the network connection. The other downside of this implementation was that it resulted in a setTimeout delay for every call even if the delay was 0.


jhurliman (approved):

jtbandes (commented):
> Chrome dev tools as a builtin feature to slow down the network connection.

The change looks fine, but I don't think this dev tools feature is relevant to the change. It appears there is no networking going on in this code path.

PR #2976, opened 2022-03-05
Sort headerStamp series in the plot panel

defunctzombie:
**User-Facing Changes**
When viewing a "header stamp" series in the plot panel, the plot correctly visualizes the header data in header stamp order regardless of the order on the incoming messages.

**Description**
The contract studio has when providing messages in "receive time" order is that the messages are provided in increasing receive time order. However, within a single receive time, there are no further guarantees about message ordering. If there are multiple messages at a receive time with different header stamps, these could cause the plot to render lines "out-of-order" when an earlier header stamp appeared in the dataset after a later header stamp (both at the same receive time).

This change fixes the "out-of-order" header stamp issue by sorting the dataset when using the header stamp method for a series. This ensures that the data is visualized in increasing header stamp order.

### DIFF
--- a/packages/studio-base/src/components/TimestampMethodDropdown.tsx
+++ b/packages/studio-base/src/components/TimestampMethodDropdown.tsx
@@ -92,7 +92,7 @@ export default function TimestampMethodDropdown(props: Props): JSX.Element {
   );
 
   const timestampMethods = [
-    { label: "Receive time", value: "recieveTime" },
+    { label: "Receive time", value: "receiveTime" },
     { label: "header.stamp", value: "headerStamp" },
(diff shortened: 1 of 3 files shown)


jhurliman (approved):

foxymiles (approved):

jtbandes:
Was this sorting code previously present and was accidentally removed? Or did this feature never work correctly even in webviz?

defunctzombie:
> Was this sorting code previously present and was accidentally removed? Or did this feature never work correctly even in webviz?

Not sure because the code has undergone so many changes. What I did find was that loading the data using our RandomAccessPlayer was ok and it was an issue I saw happen with my iterator reader which produced the messages in a different order.

Issue #1790, opened 2021-09-07
Remove resize observer loop limit sentry suppressions


defunctzombie:
Since https://github.com/foxglove/studio/pull/1698 landed, we should remove the sentry suppressions for resize observer loop limit. These errors are useful as they indicate that we've put the browser in a resize detection loop. We want to avoid doing that and have only one resize detection fire.

PR #2980, opened 2022-03-07
Remove ResizeObserver error suppression for sentry

defunctzombie:
**User-Facing Changes**
None

**Description**
https://github.com/foxglove/studio/pull/1698 fixed instances of the resize observer errors. This change allows us to collect these errors in sentry again for monitoring if this error comes back.

Fixes: #1790


defunctzombie:
I'm not entirely sure we want to do this - but we can give it a try. I predict we might be putting these back 🤞

PR #2986, opened 2022-03-07
Always use header.stamp for marker lifetime expiration, add topic warning when receiving an already expired marker

jhurliman:
**User-Facing Changes**

- Markers with lifetimes will expire relative to header.stamp, not receive time
- A topic warning is shown when an already expired marker is received

**Description**

We previously had two separate places in the codebase that were checking marker expirations, so markers would not be rendered if currentTime > either header.stamp OR receiveTime. The code has been refactored to share a single implementation that uses header.stamp unless it is not available, then falls back to receiveTime (should not be the case for any currently renderable message types).

There is now an additional sanity check when first processing an incoming marker message. If the marker is expired before we can even process it, a topic warning is shown. This will hopefully help users troubleshoot why markers are not appearing, in the case where a lifetime is set but header.stamp is unset (0) or far back in the past.

### DIFF
--- a/packages/studio-base/src/panels/ThreeDimensionalViz/SceneBuilder/MessageCollector.ts
+++ b/packages/studio-base/src/panels/ThreeDimensionalViz/SceneBuilder/MessageCollector.ts
@@ -44,26 +44,6 @@ class MessageWithLifetime {
     this.receiveTime = receiveTime;
     this.lifetime = lifetime;
   }
-
-  isExpired(currentTime: Time) {
-    if (this.lifetime == undefined) {
-      // Do not expire markers if we can't tell what their lifetime is.
-      // They'll be flushed later if needed (see flush below)
-      return false;
-    }
-    const lifetime: Time = this.lifetime;
-
-    if (areEqual(lifetime, ZERO_TIME)) {
-      // Do not expire markers with infinite lifetime (lifetime == 0)
-      return false;
-    }
-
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/panels/ThreeDimensionalViz/SceneBuilder/MessageCollector.ts
```diff
@@ -151,12 +131,25 @@ export default class MessageCollector {
     const result: ObjectWithInteractionData[] = [];
     this.markers.forEach((marker, key) => {
       // Check if the marker has a lifetime and should be deleted
-      if (marker.isExpired(this.clock)) {
+      const messageStamp =
+        (marker.message as { header?: { stamp?: Time } }).header?.stamp ?? marker.receiveTime;
```
jtbandes, line 135: Could we just pass in the message to the markerIsExpired function, rather than separate lifetime+stamp, and move this casting in there?

jtbandes (approved):

packages/studio-base/src/panels/ThreeDimensionalViz/SceneBuilder/MessageCollector.ts
jhurliman: The other callers of this method don't have anything that looks like `marker` (with `receiveTime` on it).

Issue #2977, opened 2022-03-06
Ability to edit/save plot line parameters


valschmidt:
In an operational setting, where visibility of a plot is critical, one needs the ability to increase the width of the lines and size of points drawn in the plot windows. To some extent this is true of the plot labels, although one can usually fix the bounds and learn to interpret the plot quickly. We need to be able to see it clearly from a row of seats behind the robot operator sitting at the computer. Critically, specification of these formatting parameters need to be saved with the panel layout.

PR #2988, opened 2022-03-07
Add settings modal with line color setting to Plot legend

jtbandes:
**User-Facing Changes**
Added settings for manually configuring line colors in the Plot panel.

**Description**
Resolves  https://github.com/foxglove/nvidia-issue-tracker/issues/32
Relates to https://github.com/foxglove/studio/issues/2977

https://user-images.githubusercontent.com/14237/157137148-51c08179-935c-4a1b-95c0-0f6a0af5e530.mov

### DIFF
--- a/packages/studio-base/src/panels/Plot/PlotLegendRow.tsx
+++ b/packages/studio-base/src/panels/Plot/PlotLegendRow.tsx
@@ -3,7 +3,12 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
 import { useTheme as useFluentUITheme } from "@fluentui/react";
-import { Close as CloseIcon, Error as ErrorIcon, Remove as RemoveIcon } from "@mui/icons-material";
+import {
+  Close as CloseIcon,
+  Error as ErrorIcon,
+  Remove as RemoveIcon,
+  Settings as SettingsIcon,
+} from "@mui/icons-material";
 import { IconButton, Theme, Tooltip, Typography, useTheme } from "@mui/material";
 import { createStyles, makeStyles } from "@mui/styles";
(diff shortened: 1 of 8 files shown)


packages/studio-base/src/panels/Plot/PlotLegendRow.tsx
```diff
@@ -244,6 +263,9 @@ export default function PlotLegendRow({
         </div>
       )}
       <div className={classes.actions}>
+        <IconButton size="small" title="Edit settings" onClick={() => setSettingsModalOpen(true)}>
```
2metres, line 266: I think you can solve the sizing issue by applying the same className as the `removeIcon`. I would consider renaming it to something like `actionIcon` though.

```suggestion
        <IconButton classes={classes.actionIcon} size="small" title="Edit settings" onClick={() => setSettingsModalOpen(true)}>
```

packages/studio-base/src/panels/Plot/PlotLegendRow.tsx
```diff
@@ -3,7 +3,12 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
 import { useTheme as useFluentUITheme } from "@fluentui/react";
-import { Close as CloseIcon, Error as ErrorIcon, Remove as RemoveIcon } from "@mui/icons-material";
+import {
+  Close as CloseIcon,
+  Error as ErrorIcon,
+  Remove as RemoveIcon,
+  Settings as SettingsIcon,
```
2metres, line 10: We use the [MoreVertIcon](https://mui.com/components/material-icons/?query=more&selected=MoreVert) for a similar pattern on the 3D panel perhaps we should use that here too.

Issue #2987, opened 2022-03-07
Panel toolbar prevents clicking on top portion of panel body


jtbandes:
When the plot legend is in "top" mode (above the plot), the panel toolbar covers up the close button, so the legend can't be closed (or opened if it is already closed).

https://user-images.githubusercontent.com/14237/157129575-3069f5c6-e3e2-4b22-a12b-b8dcbac5416b.mov

The same problem affects State Transitions, making the first row unclickable.

https://user-images.githubusercontent.com/14237/157130220-67af45ab-c1f8-4ab7-a790-bf8c3f6c05de.mov

PR #2989, opened 2022-03-08
Fixes clicking in the upper area of a panel

defunctzombie:
**User-Facing Changes**
The user is able to click the upper area of panels with hidden floating toolbars.

**Description**
When using a floating toolbar, the hidden toolbar would prevent clicking the upper area of a panel. This regressed in https://github.com/foxglove/studio/pull/2877. This change fixes the bug by setting the width of the toolbar to 100% for toolbars with children and no set width for no children. This lets the panel toolbar container size to the controls rather than across the entire panel.

Fixes: #2987

### DIFF
--- a/packages/studio-base/src/components/PanelToolbar/index.tsx
+++ b/packages/studio-base/src/components/PanelToolbar/index.tsx
@@ -55,11 +55,14 @@ const useStyles = makeStyles((theme) => ({
       // leave some room for possible scrollbar
       paddingRight: 8,
       top: 0,
-      width: "100%",
       zIndex: 5000,
       backgroundColor: "transparent",
 
       "&.hasChildren": {
+        // If the toolbar has children - set the width to 100% to takeup the entire panel width
+        // If the toolbar does not have children, then the width should be only the controls
+        // so the div does not interfere with other panel elements.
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/components/PanelToolbar/index.tsx
```diff
@@ -55,11 +55,14 @@ const useStyles = makeStyles((theme) => ({
       // leave some room for possible scrollbar
       paddingRight: 8,
       top: 0,
-      width: "100%",
       zIndex: 5000,
       backgroundColor: "transparent",
 
       "&.hasChildren": {
+        // If the toolbar has children - set the width to 100% to takeup the entire panel width
```
jtbandes, line 62:
```suggestion
        // If the toolbar has children, set the width to 100% to take up the entire panel width.
```

jtbandes (approved):

2metres (approved):

packages/studio-base/src/components/PanelToolbar/index.tsx
defunctzombie: I've accepted your words. They were superior.

PR #2990, opened 2022-03-08
Node Playground: remove "/studio_node" prefix requirement

### LINKED ISSUE #2210: Remove requirement for node playground nodes to have `/studio_node/` prefix
Currently node playground forces you to have a `/studio_node/` prefix. This behavior is inherited from webviz, but seems unnecessarily restrictive. For example, people might want to write a NP node to publish a `/tf` topic if they don't already have one.

jtbandes:
**User-Facing Changes**
Removed the requirement for Node Playground node output topics to start with a "`/studio_node`" prefix.

**Description**
Resolves #2210

### DIFF
--- a/packages/studio-base/src/panels/NodePlayground/BottomBar/DiagnosticsSection.tsx
+++ b/packages/studio-base/src/panels/NodePlayground/BottomBar/DiagnosticsSection.tsx
@@ -42,7 +42,7 @@ const DiagnosticsSection = ({ diagnostics }: Props): ReactElement => {
   };
 
   return diagnostics.length > 0 ? (
-    <ul>
+    <ul style={{ listStyle: "none", padding: 0 }}>
       {diagnostics.map(({ severity, message, source, startColumn, startLineNumber }, i) => {
         const severityLabel =
           (invert(DiagnosticSeverity) as Record<string, keyof typeof DiagnosticSeverity>)[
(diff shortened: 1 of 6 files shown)


defunctzombie:
We should also delete `DEFAULT_STUDIO_NODE_PREFIX` from globalConstants. I find it misleading to have it there. I think if you delete it from there it exposes a few more usage sites.

jtbandes:
It's still the _default_ prefix, so why delete it?

defunctzombie:
> It's still the _default_ prefix, so why delete it?

Oh I see - so we don't require the prefix but still provide it in our examples and defaults. I noticed it in a projectConfig variable and wondered how it related to that.

defunctzombie (approved):

defunctzombie:
> It's still the _default_ prefix, so why delete it?

I'd probably advocate for moving it - globalConstants in util is rather a lofty place for something this specific.

jtbandes:
>  I noticed it in a projectConfig variable and wondered how it related to that.

I believe we tell typescript that all the files (including generated ones) are stored under the `/studio_node/` path. I changed the node code itself to be `index.ts` in this PR.

jtbandes:
Gonna leave that as is for now...I can make another PR later if we come up with a good home for the variable.

amacneil:
🎉

PR #2991, opened 2022-03-08
Refactor NodePlayground sidebar to use @mui tabs

### LINKED ISSUE #2854: Selected icon in Node Playground is invisible in light mode
**Description**
Selected icon in Node Playground is invisible in light mode:

image

2metres:
**User-Facing Changes**
Screen Shot 2022-03-08 at 12 59 29 pm

**Description**
Fixed up the whole sidebar while I was in here. Removed the `tooltip` prop in favor of `subheader` as rendering a code block inside a tooltip is not realistic. 

Closes #2854

### DIFF
--- a/packages/studio-base/src/panels/NodePlayground/Sidebar.tsx
+++ b/packages/studio-base/src/panels/NodePlayground/Sidebar.tsx
@@ -11,10 +11,11 @@
 //   found at http://www.apache.org/licenses/LICENSE-2.0
 //   You may not use this file except in compliance with the License.
 
-import ArrowLeftBoldIcon from "@mdi/svg/svg/arrow-left-bold.svg";
-import DeleteIcon from "@mdi/svg/svg/delete.svg";
-import FileMultipleIcon from "@mdi/svg/svg/file-multiple.svg";
-import HelpCircleIcon from "@mdi/svg/svg/help-circle.svg";
+import CloseIcon from "@mui/icons-material/Close";
+import ConstructionOutlinedIcon from "@mui/icons-material/ConstructionOutlined";
+import DeleteOutlineIcon from "@mui/icons-material/DeleteOutline";
+import TemplateIcon from "@mui/icons-material/PhotoFilter";
+import NoteIcon from "@mui/icons-material/StickyNote2Outlined";
 import {
   List,
   ListItem,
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/panels/NodePlayground/Sidebar.tsx
```diff
@@ -11,50 +11,54 @@
...
-  display: flex;
-  flex-direction: column;
-  align-items: center;
-  width: 40px;
-  background-color: ${({ theme }) => theme.palette.neutralLighterAlt};
-  & > * {
-    margin: 10px;
-  }
-`;
+const STab = muiStyled(Tab)(({ theme }) => ({
```
defunctzombie, line 42: No need to change here but I really dislike the whole `S` prefix convention throughout the codebase.

packages/studio-base/src/panels/NodePlayground/Sidebar.tsx
```diff
@@ -178,12 +206,22 @@ const Sidebar = ({
...
             collapse={() => updateExplorer(undefined)}
             title="Utilities"
-            tooltip={`You can import any of these modules into your node using the following syntax: 'import { .. } from "./pointClouds.ts".\n\nWant to contribute? Scroll to the bottom of the docs for details!`}
+            subheader={
+              <>
+                <Typography variant="body2" color="text.secondary">
+                  You can import any of these modules into your node using the following syntax:{" "}
+                  <pre>{`import { ... } from "./pointClouds.ts".`}</pre>
+                </Typography>
+                <Typography variant="body2" color="text.secondary">
```
defunctzombie, line 218:
```suggestion
```

packages/studio-base/src/panels/NodePlayground/Sidebar.tsx
```diff
@@ -178,12 +206,22 @@ const Sidebar = ({
...
+                  Want to contribute? Scroll to the bottom of the docs for details!
```
defunctzombie, line 219:
```suggestion
```

packages/studio-base/src/panels/NodePlayground/Sidebar.tsx
```diff
@@ -178,12 +206,22 @@ const Sidebar = ({
...
+                </Typography>
```
defunctzombie, line 220:
```suggestion
```

packages/studio-base/src/panels/NodePlayground/Sidebar.tsx
defunctzombie: We should remove this comment. Makes no sense to me.

defunctzombie (approved):
LGTM - remove the comment about "contributing".

maybe in the future we can have less STabbing in our components 🤣

packages/studio-base/src/panels/NodePlayground/Sidebar.tsx
jtbandes: I personally find it helps readability because it tells me at a glance whether a custom react component is actually a complex/compound component that includes its own logic, or just a styled-component.

packages/studio-base/src/panels/NodePlayground/Sidebar.tsx
2metres: Add a link to the repo instead?

packages/studio-base/src/panels/NodePlayground/Sidebar.tsx
2metres: Prior to Foxglove I tended to use `StyledTab` which is slightly more verbose but very clear what it is. 

@defunctzombie how does that make you feel?

defunctzombie: To me it's a `SidebarTab`. I don't find it useful to know its made via shorthand with muiStyled vs if I'd written a component for it /shrug. I don't like `S` or `Styled` because all it tells me is the component is styled - all components are styled.

packages/studio-base/src/panels/NodePlayground/Sidebar.tsx
defunctzombie: No. We don't need to say anything else here. This is the list of utilities. We don't call out other random places in the app to "open an issue with feedback" so I don't think we need to here either.

PR #2992, opened 2022-03-08
Set first seen cohort for new users

amacneil:
**User-Facing Changes**
None

**Description**
Add `first_seen_month` and `first_seen_year` properties to users.

### DIFF
--- a/packages/studio-base/src/services/AmplitudeAnalytics.ts
+++ b/packages/studio-base/src/services/AmplitudeAnalytics.ts
@@ -5,6 +5,7 @@
 import { setUser as setSentryUser, addBreadcrumb as addSentryBreadcrumb } from "@sentry/core";
 import { Severity } from "@sentry/types";
 import amplitude, { AmplitudeClient } from "amplitude-js";
+import moment from "moment";
 
 import Logger from "@foxglove/log";
 import OsContextSingleton from "@foxglove/studio-base/OsContextSingleton";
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/services/AmplitudeAnalytics.ts
```diff
@@ -49,6 +50,12 @@ export class AmplitudeAnalytics implements IAnalytics {
         this._amp.setVersionName(appVersion);
       }
 
+      // use setOnce() to set first seen cohort for new users
+      const identify = new amplitude.Identify();
+      identify.setOnce("first_seen_month", moment.utc().format("YYYY-MM"));
```
jhurliman, line 55: Is something expecting this format?

jtbandes: Do we know the user is a new user here, or does amplitude automatically discard new "identify" calls when it knows the user is not new?

amacneil: > Is something expecting this format?
it's just arbitrary key/value data so the key here is just something I made up.

> Do we know the user is a new user here, or does amplitude automatically discard new "identify" calls when it knows the user is not new?
Unfortunately no, there is no way to identify them as a new user. `setOnce()` will solve this going forward, future updates will be ignored. I've been emailing back and forth with Amplitude support but it appears there are no automatic features to group users by first seen periods, and there is no way to backfill this data for users. So it's just going to be a bit messy for the first couple months. I will wait on merging this PR until I hear back on my most recent question to them, but I'm not holding my breath.

Maybe there is some other way we can hook into "did amplitude create a new device ID" and only set this in those cases to avoid corrupting old data.

jtbandes: I guess my question is how does `setOnce` know whether the value needs to be overridden or not? How does it identify the "existing" user if there is one?

amacneil: `setOnce` is a server side operation ([docs](https://developers.amplitude.com/docs/identify-api#keys-for-the-identification-argument)).

I also heard back from their support and I think I have a path forward on the backfill, so will try that before merging.

PR #2993, opened 2022-03-08
Simplify sample layout perception tab

amacneil:
**User-Facing Changes**
Simplifies the nuscenes sample layout (use the deployment to test).

* Remove the row of 3 rear cameras
  Why: 3d view doesn't work well with a small height (topic selector and object inspector are impossible to use). We will fix this, but for now the sample layout gives a poor first user experience when they try to expand the topic picker and it is cut off. 

* Zoom out the perception 3d panel a bit
  Why: So the first time user can see what is going on better

### DIFF
--- a/packages/studio-base/src/dataSources/SampleNuscenesLayout.json
+++ b/packages/studio-base/src/dataSources/SampleNuscenesLayout.json
@@ -9,12 +9,7 @@
       "customMarkerTopicOptions": [],
       "transformMarkers": false,
       "synchronize": false,
-      "mode": "fit",
-      "zoom": 1.467091317737759,
-      "pan": {
-        "x": -18.766339672385854,
-        "y": 3.9962119425323834
-      }
+      "mode": "fit"
     },
     "ImageViewPanel!15kmj7r": {
       "cameraTopic": "/CAM_FRONT/image_rect_compressed",
@@ -25,12 +20,7 @@
       "customMarkerTopicOptions": [],
       "transformMarkers": false,
       "synchronize": false,
-      "mode": "fit",
-      "zoom": 1.1281780995019715,
-      "pan": {
-        "x": 9.510604159161662,
-        "y": -4.4608086865532295
-      }
+      "mode": "fit"
     },
     "ImageViewPanel!1jn49st": {
       "cameraTopic": "/CAM_FRONT_RIGHT/image_rect_compressed",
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/dataSources/SampleNuscenesLayout.json
```diff
@@ -9,12 +9,7 @@
       "customMarkerTopicOptions": [],
       "transformMarkers": false,
       "synchronize": false,
-      "mode": "fit",
-      "zoom": 1.467091317737759,
-      "pan": {
-        "x": -18.766339672385854,
-        "y": 3.9962119425323834
-      }
```
amacneil, line 17: removed these so studio will automatically "fit"

packages/studio-base/src/dataSources/SampleNuscenesLayout.json
```diff
@@ -310,20 +242,20 @@
...
             "second": {
               "first": "SourceInfo!7dc2bd",
               "second": "RawMessages!21ab95j",
               "direction": "column",
-              "splitPercentage": 59.81479888764657
+              "splitPercentage": 60
             },
             "direction": "row",
-            "splitPercentage": 48.80750465900287
+            "splitPercentage": 50
```
amacneil, line 258: manually set these percentages to be exact

packages/studio-base/src/dataSources/SampleNuscenesLayout.json
```diff
@@ -104,23 +41,32 @@
...
-      "settingsByKey": {},
+      "settingsByKey": {
+        "t:/pose": {
+          "size": {
+            "headLength": 0.5,
+            "headWidth": 2,
+            "shaftWidth": 2
+          }
+        }
+      },
```
amacneil, line 69: show the "pose" by default and make it slightly more arrow-like

packages/studio-base/src/dataSources/SampleNuscenesLayout.json
```diff
@@ -104,23 +41,32 @@
         "ns:/tf:CAM_BACK",
         "t:/map",
         "t:/LIDAR_TOP",
-        "t:/markers/annotations"
+        "t:/markers/annotations",
+        "t:/pose"
       ],
-      "expandedKeys": ["name:Topics", "t:/tf"],
+      "expandedKeys": ["name:Topics"],
```
amacneil, line 47: hide the "tf" layer selector by default

defunctzombie (approved):
I think you are somewhat skewed by using this on a smaller display? Larger display works fine with all the camera images and 3d panel maybe?

amacneil:
Even a large window size still results in 3d panel dropdowns being cut off in the existing layout, because the 3d panel only gets 30% of your window height:

Screen Shot 2022-03-08 at 2 48 43 PM

PR #2995, opened 2022-03-08
Fix data platform playback deadlock

jtbandes:
**User-Facing Changes**
Fixed an issue where playback from the Foxglove Data Platform would sometimes stall.

**Description**
Playback would stall when entering the `while` loop if the outstanding request was not large enough to ever satisfy the getMessages range. Moving the `_startPreloadTaskIfNeeded` inside the loop ensures that if the range is not satisfied by the current request, a new one will be made after it finishes. Then, adding another `this._loadedMoreMessages?.resolve();` in the `finally()` ensures the loop gets another chance to call `_startPreloadTaskIfNeeded` when it will actually have an effect (if it's called too early, before `this._currentPreloadTasks[preloadType] = undefined`, a new task would not be started).


foxymiles (approved):
Nice fix!

PR #2484, opened 2021-12-22
Move Plot panel's legend next to the plot itself

esthersweon:
**User-Facing Changes**
- Move Plot panel's legend next to the plot itself, so that the legend will not ever cover the plot data
- Make it possible to toggle the legend's visibility
- Legend cleanup
   - removed 'y' labels
   - condensed dropdown menu for message order options
   - removed help icon
   - moved download icon to panel toolbar
   - horizontally + vertically scrollable
- Made boundary between legend and Plot panel adjustable 
   - persist this value to the panel config
- Only show tooltips when legend is closed
   Screen Shot 2022-01-04 at 5 11 09 PM
   Screen Shot 2022-01-04 at 5 11 50 PM
- Added panel setting for toggling axes labels
   Screen Shot 2022-01-12 at 11 49 24 AM

PR #3001, opened 2022-03-09
Move plot series timestamp setting into settings modal

jtbandes:
**User-Facing Changes**
Moved Plot panel setting for receive-time/header-stamp ordering into the new per-series settings dialog.

**Description**
image

Screen Shot 2022-03-09 at 2 17 54 PM

Screen Shot 2022-03-09 at 2 17 59 PM

### DIFF
--- a/packages/studio-base/src/panels/Plot/PlotLegendRow.tsx
+++ b/packages/studio-base/src/panels/Plot/PlotLegendRow.tsx
@@ -16,10 +16,8 @@ import { v4 as uuidv4 } from "uuid";
 
 import MessagePathInput from "@foxglove/studio-base/components/MessagePathSyntax/MessagePathInput";
 import TimeBasedChart from "@foxglove/studio-base/components/TimeBasedChart";
-import TimestampMethodDropdown from "@foxglove/studio-base/components/TimestampMethodDropdown";
 import { useHoverValue } from "@foxglove/studio-base/context/HoverValueContext";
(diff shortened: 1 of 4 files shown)


packages/studio-base/src/panels/Plot/PlotLegendRow.tsx
```diff
@@ -166,12 +164,6 @@ export default function PlotLegendRow({
   const classes = useStyles();
 
   const isReferenceLinePlotPath = isReferenceLinePlotPathType(path);
-  let timestampMethod;
-  // Only allow chosing the timestamp method if it is applicable (not a reference line) and there is at least
-  // one character typed.
-  if (!isReferenceLinePlotPath && path.value.length > 0) {
-    timestampMethod = path.timestampMethod;
-  }
```
jtbandes, line 174: This code accidentally stopped working in https://github.com/foxglove/studio/pull/2484 when we moved the dropdown out of MessagePathInput. This PR fixes the behavior so the dropdown is disabled for reference lines.

defunctzombie (approved):
I think we should call this out in the release notes. It's a move of a UI element.

2metres:
Does this take into account the disccusion we had about inheriting from the "global" setting?

jtbandes:
> Does this take into account the disccusion we had about inheriting from the "global" setting?

No, purely moving the dropdown into the modal for now.

PR #3002, opened 2022-03-09
Add links to docs for each connection and remote data source in the Open Dialog

esthersweon:
**User-Facing Changes**
- Add links to docs for each connection and remote data source in the Open Dialog


jtbandes:
Can we put the data source description or some basic docs inline here as well? Seems like there is plenty of space to put at least a short paragraph about each one.

esthersweon:
> Can we put the data source description or some basic docs inline here as well? 

@jtbandes - Added in a follow-up commit; thoughts?

Screen Shot 2022-03-10 at 2 34 26 PM

packages/studio-base/src/context/PlayerSelectionContext.ts
```diff
@@ -25,6 +25,8 @@ export interface IDataSourceFactory {
   type: DataSourceFactoryType;
   displayName: string;
   iconName?: RegisteredIconNames;
+  docsIntro?: string;
+  docsLink?: string;
```
jtbandes, line 29: maybe rename `docsIntro` -> `helpText`, `description`, `summary`, something like that?

packages/studio-base/src/components/OpenDialog/Remote.tsx
```diff
@@ -80,6 +90,15 @@ export default function Remote(props: RemoteProps): JSX.Element {
             setCurrentUrl(newValue);
           }}
         />
+
+        {availableSources.map(
+          ({ displayName, docsLink }) =>
+            docsLink && (
+              <Link key={docsLink} href={`https://foxglove.dev/docs/studio/connection${docsLink}`}>
```
jtbandes, line 97: Might be nicer to make `docsLink` an absolute url so that if we need to update any of the urls, you don't have to piece together two different source files to figure out what url is being used.

packages/studio-base/src/dataSources/FoxgloveWebSocketDataSourceFactory.tsx
```diff
@@ -14,6 +14,8 @@ export default class FoxgloveWebSocketDataSourceFactory implements IDataSourceFa
   type: IDataSourceFactory["type"] = "connection";
   displayName = "Foxglove WebSocket";
   iconName: IDataSourceFactory["iconName"] = "Flow";
+  docsIntro = "Connect live to your non-ROS data via an encoding-agnostic WebSockets connection.";
```
jtbandes, line 17:
```suggestion
  docsIntro = "Connect live to your non-ROS data via an encoding-agnostic WebSocket connection.";
```

Not sure if we should say "non-ROS" here. If someone doesn't use ROS, then the fact that this is non-ROS might not be so relevant to them 😀   Maybe "custom data"?

I think it might be good to mention that in order to use this data source, you usually have to write a custom server.

packages/studio-base/src/dataSources/Ros2SocketDataSourceFactory.tsx
```diff
@@ -14,6 +14,9 @@ class Ros2SocketDataSourceFactory implements IDataSourceFactory {
   type: IDataSourceFactory["type"] = "connection";
   displayName = "ROS 2";
   iconName: IDataSourceFactory["iconName"] = "studio.ROS";
+  docsIntro =
+    "Connect to a running ROS 2 system via a native TCP connection that accesses your ROS nodes directly.";
+  docsLink = "/native";
 
   formConfig = {
     fields: [{ id: "url", label: "ROS_DOMAIN_ID", defaultValue: "0" }],
```
jtbandes, line 22: I think it would also be helpful to have a short description of what each field is for.

packages/studio-base/src/dataSources/RosbridgeDataSourceFactory.tsx
```diff
@@ -14,6 +14,9 @@ class RosbridgeDataSourceFactory implements IDataSourceFactory {
   type: IDataSourceFactory["type"] = "connection";
   displayName = "Rosbridge (ROS 1 & 2)";
   iconName: IDataSourceFactory["iconName"] = "Flow";
+  docsLink = "/rosbridge";
+  docsIntro =
+    "Connect to a running ROS 1 or ROS 2 system over WebSockets to access your ROS nodes directly.";
```
jtbandes, line 19:
```suggestion
    "Connect to a running ROS 1 or ROS 2 system over WebSockets using the rosbridge_suite package.";
```
Not married to this wording suggestion, but I think it is helpful to clarify that this connection is meant for use with an off-the-shelf server package.

jtbandes (approved):
I like having these descriptions. I think a short description of the form fields would also be helpful, at least for those that aren't immediately obvious like `ROS_DOMAIN_ID`.

packages/studio-base/src/dataSources/Ros2SocketDataSourceFactory.tsx
esthersweon: Added some follow-up commits to move docs link below description, and to add field-specific help text:

Screen Shot 2022-03-21 at 10 42 04 AM

PR #3007, opened 2022-03-10
Fix sensor_msgs/LaserScan hasIntensity check for typed arrays

jhurliman:
**User-Facing Changes**

- Fixed a bug where `sensor_msgs/LaserScan` messages with intensities would sometimes not allow coloring by intensity

**Description**

```
> Array.isArray(new Float32Buffer())
false
> ArrayBuffer.isView(new Float32Buffer())
true
```


jtbandes (approved):
Looks like `intensities` is typed as `number[]`, should we change it to `number[] | Float32Array`?

defunctzombie:
Please add a story to show we can color by intensity.

foxymiles (approved):

jhurliman:
@jtbandes probably. We have this problem in any message definition type that includes a `number[]`.

Issue #3017, opened 2022-03-11
Node playground is not working with experimental bag player


defunctzombie:
**Description**

> node playground complains that none of the custom message types are found on MessageTypeBySchemaName, and I see a few duplicate identifier: /tf errors as well. This doesn’t happen with the same bag with the other bag player.

PR #3022, opened 2022-03-11
Provide datatypes from IterablePlayer and BagIterableSource

defunctzombie:
**User-Facing Changes**
Node playground works when using the Experimental bag player.

**Description**
IterablePlayer did not expose the datatypes of its IterableSource. This change updates the player to expose the datatypes. This change also adds a check for duplicate topics (which Studio does not like) and filters those out.
    
Fixes: #3017

### DIFF
--- a/packages/studio-base/src/panels/NodePlayground/Sidebar.tsx
+++ b/packages/studio-base/src/panels/NodePlayground/Sidebar.tsx
@@ -31,7 +31,7 @@ import {
   Typography,
 } from "@mui/material";
 import * as monacoApi from "monaco-editor/esm/vs/editor/editor.api";
-import { ReactNode, useCallback, useMemo, useState } from "react";
+import { ReactNode, useCallback, useMemo } from "react";
 
 import { Explorer } from "@foxglove/studio-base/panels/NodePlayground";
 import { Script } from "@foxglove/studio-base/panels/NodePlayground/script";
@@ -159,7 +159,6 @@ const Sidebar = ({
   script,
   addNewNode,
 }: Props): React.ReactElement => {
-  const [activeExplorerTab, setActiveExplorerTab] = useState<string | undefined>(undefined);
   const nodesSelected = explorer === "nodes";
   const utilsSelected = explorer === "utils";
   const templatesSelected = explorer === "templates";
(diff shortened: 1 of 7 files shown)


packages/studio-base/src/panels/NodePlayground/Sidebar.tsx
```diff
@@ -210,7 +214,7 @@ const Sidebar = ({
             collapse={() => updateExplorer(undefined)}
             title="Utilities"
             subheader={
-              <Typography variant="body2" color="text.secondary">
+              <Typography variant="body2" color="text.secondary" component="div">
```
defunctzombie, line 217: @2metres FYI - this was producing errors in console saying the `<pre>` is now allowed as a parent of `<p>`. Typography is a <p> tag by default. Good to keep errors out of console.

packages/studio-base/src/panels/NodePlayground/Sidebar.tsx
```diff
@@ -282,7 +286,7 @@ const Sidebar = ({
   return (
     <Paper>
       <Stack direction="row" height="100%">
-        <STabs orientation="vertical" value={activeExplorerTab} onChange={handleExplorerTabChange}>
+        <STabs orientation="vertical" value={activeExplorerTab}>
```
defunctzombie, line 289: @2metres this was producing errors in console because the Tab components wants `false` used for the value when no tab is selected.

packages/studio-base/src/panels/NodePlayground/Sidebar.tsx
```diff
@@ -184,14 +183,19 @@ const Sidebar = ({
...
 
-  const handleExplorerTabChange = useCallback(
-    (_event: React.ChangeEvent<unknown>, newValue: string) => {
-      if (newValue !== activeExplorerTab) {
-        setActiveExplorerTab(newValue);
-      }
-    },
-    [activeExplorerTab],
-  );
+  const activeExplorerTab = useMemo(() => {
```
defunctzombie, line 186: @2metres updated this to compute the active tab from the explorer rather than its own state since the explorer is the active tab.

foxymiles (approved):

packages/studio-base/src/panels/NodePlayground/Sidebar.tsx
jtbandes: Seems like this could just be `const activeExplorerTab = explorer ?? false` ?

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
```diff
@@ -176,13 +176,31 @@ export class IterablePlayer implements Player {
...
         await this._iterableSource.initialize();
 
       this._start = this._currentTime = start;
       this._end = end;
-      this._providerTopics = topics;
       this._publishedTopics = publishersByTopic;
+      this._providerDatatypes = datatypes;
+
+      // Studio does not like duplicate topics or topics with different datatypes
+      // Check for duplicates or for miss-matched datatypes
```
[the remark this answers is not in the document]
jtbandes: Is this code actually checking for mismatched datatypes? I guess it just avoids that by avoiding duplicate topics?

(I don't think the MCAP and Data Platform players do a good job of avoiding this right now. The duplicate topics will probably just be passed through and cause issues downstream if they have non-matching datatypes. I think at some point we should try to get rid of the whole "datatypes" object and make the schema a field of Topic instead.)

jtbandes (approved):

Issue #3018, opened 2022-03-11
Topic setting to "frame lock" OccupancyGrid topics


defunctzombie:
**Description**

I have a topic which publishes an occupancy grid infrequently. This grid is published on a _map_ frame and represents a static map. As my robot moves around the world and localizes, it produces transforms from the _map_ frame to my _base_ frame.

If I set the 3d panel to follow the _base_ frame, and turn on the map topic, I am able to see the map as my robot moves. However, when I seek, the map disappears. Using the experimental bag player, even when I start playing the map never re-appears. This is because my map, which is published once and has a header stamp from many moments before the seek time. Once I seek, transforms appear for _base_ to _map_ at the seek time however the 3d panel attempts to lookup a transform for _map_ to _base_ for the map message header.stamp.

I'd like a topic setting which lets me ignore the `header.stamp` from the message and attempt to use the current time to lookup the transform between my frames. Turning this on for a topic would make the topic behave similar to the `frame_locked` marker field.

---

Idea: little icon next to frame locked topics

defunctzombie:
Related TODO

Update max storage to infinite duration
`const DEFAULT_MAX_STORAGE_TIME: Duration = INFINITE_DURATION;`

PR #3024, opened 2022-03-15
Add frame lock option for OccupancyGrid topics

defunctzombie:
**User-Facing Changes**
Add a new topic settings modal for OccupancyGrid topics. The modal introduces a `Frame lock` option. When the `Frame lock` option is enabled, occupancy grids are placed within the 3d scene using transformations from the _current time_ rather than the `header.stamp` time as is the default behavior.

image

**Description**
Adding a frame lock option to occupancy grid topics allows users who publish occupancy grids as static maps to publish their message infrequently while still having the grid display correctly relative to other visual elements as new transforms are received.

Fixes: #3018


packages/studio-base/src/panels/ThreeDimensionalViz/SceneBuilder/index.ts
```diff
@@ -197,6 +197,7 @@ function computeMarkerPose(
   }
   let srcTime: Time;
   let dstTime: Time;
+
```
defunctzombie, line 200:
```suggestion
```

packages/studio-base/src/panels/ThreeDimensionalViz/TopicSettingsEditor/GenericSettingsEditor.tsx
```diff
@@ -0,0 +1,47 @@
...
+        control={
+          <Checkbox
+            checked={settings.frameLocked === true}
+            onChange={(_ev, checked) => onFieldChange("frameLocked", checked)}
+          />
+        }
+        label="Frame lock"
+      />
+      <SDescription>
+        By default, topics are translated into the 3d scene using their header stamp time. Enabling
```
jtbandes, line 41:
```suggestion
        By default, topics are translated into the 3D scene using their header stamp time. Enabling
```

jtbandes: Not sure I like this explanation overall. I think this is clearer: https://answers.ros.org/question/290293/what-exactly-does-frame_locked-mean/?answer=290337#post-id-290337

jtbandes (approved):
Long live the unholy union

Issue #2824, opened 2022-02-10
Make "Data went back in time" less intrusive


bertaveira:
Would be nice to have a way to clear all warnings or even ignore future ones. When things go sideways I have to restart foxglove since it is impossible to clear all. It is even hard to debug the situation since it is covering part of the screen.

![image](https://user-images.githubusercontent.com/9644729/153476977-42c62243-ecb8-402c-82ba-394ba4be1b0a.png)

defunctzombie:
@amacneil Let's add this to an upcoming roadmap month. I think its not a big lift for us to make this at least less terrible for now. Longer term I'd like to not use this style of notification for this issue - but the architecture we have today makes that trickier.

amacneil:
A few thoughts:
- "clear all" button would definitely help, although with these errors it looks like they will keep happening if the data source is producing them
- grouping errors by text might be nice, although you still have to have the annoying popup over your screen the whole time (assuming it just comes back after you clear it)
- nicest option is probably to move this warning to the data source sidebar (similarly to how we display data source errors)

defunctzombie:
We can tackle this by changing the MessageOrderTracker `update` call to inject problems into the player state and return a new player state. This would remove send notification and allow the issue to display on the data source sidebar in a less intrusive manner.

https://github.com/foxglove/studio/blob/main/packages%2Fstudio-base%2Fsrc%2Fcomponents%2FMessagePipeline%2FMessageOrderTracker.ts#L48

PR #3030, opened 2022-03-16
Remove sendNotification from MessageOrderTracker

defunctzombie:
**User-Facing Changes**
When encountering message order warnings the user is no longer bombarded with toast noticiations. Instead, the warnings are added to player problems and displayed alongside the existing player problems display (currently in the Data Source sidebar).

**Description**
Rather than emitting a notification for each message order violation, the MessageOrderTracker builds up an array of player problems. These player problems are added to the existing player problems. Downstream consumers of the playerState are responsible for the display of the problems.

Fixes: #2824

### DIFF
--- a/packages/studio-base/src/components/MessagePipeline/MessageOrderTracker.ts
+++ b/packages/studio-base/src/components/MessagePipeline/MessageOrderTracker.ts
@@ -13,8 +13,7 @@
 
 import Logger from "@foxglove/log";
 import { Time, isLessThan, subtract as subtractTimes, toSec } from "@foxglove/rostime";
-import { PlayerState, MessageEvent } from "@foxglove/studio-base/players/types";
-import sendNotification from "@foxglove/studio-base/util/sendNotification";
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/components/MessagePipeline/MessageOrderTracker.ts
```diff
@@ -45,10 +44,13 @@ class MessageOrderTracker {
 
   private incorrectMessages: MessageEvent<unknown>[] = [];
 
-  update(playerState: PlayerState): void {
+  update(playerState: PlayerState): PlayerState {
```
jtbandes, line 47: I think the API would be cleaner if this just returned problems, and the caller decided what to do with those problems (e.g. merge them into player state)

jtbandes (approved):

packages/studio-base/src/components/MessagePipeline/MessageOrderTracker.ts
defunctzombie: 👍 good idea - will make that change

PR #3031, opened 2022-03-16
Fix url state and recents update issue

### LINKED ISSUE #2944: Recent data sources list shows "MCAP" label with `ws://` URL
**Description**
image

**Steps To Reproduce**
- Start a foxglove websocket server, e.g. `python3 -m foxglove_websocket.examples.json_server`
- Visit studio.foxglove.dev, connect the app to the foxglove websocket server
- Stop the server, so the app displays a "connection failed" error
- Refresh studio.foxglove.dev, so `?ds=foxglove-websocket&ds.url=...` is still in the address bar
- Drag a local .mcap file into the app, now `?ds=mcap-local-file&ds.url=...` is displayed in the address bar
- Refresh the page again
  - Now the data source recents list shows:
     image
  - Clicking on this entry logs out `Select Source: mcap-local-file` in the console, but doesn't show the open dialog or load a file.

**Expected Behavior**
It should never be possible to see a websocket url in the recents list with a "MCAP" label. Ideally the address bar would also never display `?ds=mcap-local-file&ds.url=...`, but we can't prevent people from typing in that url manually.

foxymiles:
**User-Facing Changes**
This fixes an issue with updates to the state url and the recents list.

**Description**
Existing logic for updating the URL and the recents list doesn't take the current state of the player into consideration and can in some cases garble state from two different sources. The solution here is to defer updating the URL until a newly selected source has successfully been loaded. This is a bit tricky since we need state from two different contexts to do this so some sequential `useEffect` steps are needed.

Fixes #2944

### DIFF
--- a/packages/studio-base/src/components/MessagePipeline/MockMessagePipelineProvider.tsx
+++ b/packages/studio-base/src/components/MessagePipeline/MockMessagePipelineProvider.tsx
@@ -29,6 +29,7 @@ import {
   PublishPayload,
   SubscribePayload,
   Topic,
+  PlayerURLState,
 } from "@foxglove/studio-base/players/types";
 import { RosDatatypes } from "@foxglove/studio-base/types/RosDatatypes";
 import naturalSort from "@foxglove/studio-base/util/naturalSort";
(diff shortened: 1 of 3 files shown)


defunctzombie (approved):
I think given how much we've changed this it might be time to make a test for this.

jtbandes (approved):

Issue #2303, opened 2021-12-07
Data platform streaming should fall back to just-in-time schema parsing


jtbandes:
If the /topics endpoint did not return a reader matching the streamed data, we currently throw an error: https://github.com/foxglove/studio/blob/e3fc661f7f2ae364d73fe71e6b17a9e9b0f8e5ba/packages/studio-base/src/players/FoxgloveDataPlatformPlayer/streamMessages.ts#L88

We should continue producing a warning here and log it to Sentry, but we should also add a fallback to parse the schema on demand so that we can continue streaming the data if possible.

PR #3037, opened 2022-03-17
Add just-in-time reader creation for data platform player

defunctzombie:
**User-Facing Changes**
Users playing data from data platform no longer see errors if they encounter previously unseen schema/channel combinations. This is an implementation detail that prevented the vieweing of their data.

**Description**
When data platform streaming encounters a previously unknown channel/schema combination, it creates a parser and stores it alongside the other pre-initialized parsers. This makes message streaming more resilient to to encountering semantically identical schemas (same serialization format), but physically different schema data (i.e. comments or other extraneous artifacts with no impact on serialization).

Fixes: #2303

### DIFF
--- a/packages/studio-base/src/players/FoxgloveDataPlatformPlayer/streamMessages.ts
+++ b/packages/studio-base/src/players/FoxgloveDataPlatformPlayer/streamMessages.ts
@@ -2,6 +2,7 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
+import { captureException } from "@sentry/core";
 import { isEqual } from "lodash";
 
 import Logger from "@foxglove/log";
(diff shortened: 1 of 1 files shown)


wkalt (approved):

foxymiles (approved):

packages/studio-base/src/players/FoxgloveDataPlatformPlayer/streamMessages.ts
```diff
@@ -117,17 +117,28 @@ export default async function* streamMessages({
...
+          schema: schema.data,
+          parsedChannel,
+        });
+
+        // This is tragic because we are mutating parsedChannelsByTopic which is an input
+        // To avoid mutating the input the overall flow needs reworking.
+        parsedChannelsByTopic.set(record.topic, parsedChannels);
+
+        channelInfoById.set(record.id, { channel: record, parsedChannel });
+        log.warn("No pre-initialized reader for", record);
```
jtbandes, line 140: Keep the `parsedChannels` in the output? would be helpful for debugging if this happens again...not clear why it is being removed

jtbandes (approved):

PR #3038, opened 2022-03-17
Fix destructing an undefined value for occupancy grids

defunctzombie:
**User-Facing Changes**
Fixes _cannot destructure_ error in 3d panel when using OccupancyGrid topics

**Description**
Bug introduced in https://github.com/foxglove/studio/pull/3024


jtbandes (approved):

foxymiles (approved):
This is probably a pattern we should be watching out for in general.

PR #3047, opened 2022-03-21
Allow number or string for foxglove Image message timestamp

defunctzombie:
**User-Facing Changes**

**Description**
When loading JSON data (via mcap), using `(u)int64` fields is awkward since there is no way to specify bigint types with JSONSchema. As a workaround, we allow the timestamp field to be a string or number and convert those to bigint within the image panel.

### DIFF
--- a/packages/studio-base/src/panels/ImageView/lib/normalizeMessage.ts
+++ b/packages/studio-base/src/panels/ImageView/lib/normalizeMessage.ts
@@ -23,6 +23,15 @@ export const NORMALIZABLE_IMAGE_DATATYPES = [
   "foxglove.CompressedImage",
 ];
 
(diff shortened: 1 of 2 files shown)


amacneil:
Also needs to be handled for CameraCalibration, ImageAnnotations, and Log

defunctzombie:
> Also needs to be handled for CameraCalibration, ImageAnnotations, and Log

Not yet sure if this is the approach. This PR is for discussion around approaches and to experiment. Missing all those cases is precisely why I am shying away from such manual handling.

defunctzombie:
After discussion, we've decided to go in a different direction for timestamps: https://github.com/foxglove/studio/issues/3070

Issue #2800, opened 2022-02-07
Support remote (http/https) playback of mcap0 files


amacneil:
Similar to https://github.com/foxglove/studio/issues/2760, except we can load the mcap file from http like we do with .bag files.

yjmade:
Any update on this feature? It's crutial for us to switch from rosbag to mcap.

jtbandes:
Opened a PR: https://github.com/foxglove/studio/pull/3049

PR #3049, opened 2022-03-21
Add support for remote MCAP files

jtbandes:
**User-Facing Changes**
Added support for loading .mcap files from a remote URL.

**Description**
Resolves https://github.com/foxglove/studio/issues/2800

### DIFF
--- a/packages/studio-base/src/components/OpenDialog/Remote.tsx
+++ b/packages/studio-base/src/components/OpenDialog/Remote.tsx
@@ -5,7 +5,7 @@
 import { TextField, Link, Text, useTheme } from "@fluentui/react";
 import { Stack } from "@mui/material";
 import path from "path";
-import { useCallback, useState } from "react";
+import { useCallback, useState, Fragment } from "react";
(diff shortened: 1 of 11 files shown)


packages/studio-base/src/components/OpenDialog/Remote.tsx
```diff
@@ -73,26 +73,23 @@ export default function Remote(props: RemoteProps): JSX.Element {
...
-            ),
-        )}
-
-        {availableSources.map(
-          ({ displayName, docsLink }) =>
-            docsLink && (
+            )}
+            {docsLink && (
               <Link key={docsLink} href={docsLink}>
                 View {displayName} docs.
```
defunctzombie, line 88: Please remove these doc links. I do not think they are useful to someone trying to connect to a remove data source.

defunctzombie (approved):
👍 

Remove the doc links on the connection page. If we want to say anything it should be below the URL bar but I don't think we need to say anything.

Issue #2899, opened 2022-02-23
Data platform: Support for infrequently published or latched topics while scrubbing


amacneil:
Similar to https://github.com/foxglove/studio/issues/707 (for local bag files).

When connecting to a data platform source, I should be able to seek and still see latched or infrequent topics such as tf_static, occupancy maps, etc.

defunctzombie:
Done in https://github.com/foxglove/studio/pull/3052

PR #3052, opened 2022-03-22
Add backfill method to IIterableSource, remove reverse param from mesageIterator

jtbandes:
**User-Facing Changes**
None - feature flagged

**Description**
Relates to https://github.com/foxglove/studio/issues/2899 for the experimental player.
The iterating-in-reverse logic is moved into BagIterableSource. The DataPlatformIterableSource implements backfill by passing a `replayPolicy` param.

### DIFF
--- a/packages/studio-base/src/players/IterablePlayer/BagIterableSource.ts
+++ b/packages/studio-base/src/players/IterablePlayer/BagIterableSource.ts
@@ -8,8 +8,9 @@ import { Bag, Filelike } from "@foxglove/rosbag";
 import { BlobReader } from "@foxglove/rosbag/web";
 import { parse as parseMessageDefinition } from "@foxglove/rosmsg";
 import { LazyMessageReader } from "@foxglove/rosmsg-serialization";
+import { compare } from "@foxglove/rostime";
 import { Topic } from "@foxglove/studio";
-import { PlayerProblem } from "@foxglove/studio-base/players/types";
+import { PlayerProblem, MessageEvent } from "@foxglove/studio-base/players/types";
 import { RosDatatypes } from "@foxglove/studio-base/types/RosDatatypes";
 import BrowserHttpReader from "@foxglove/studio-base/util/BrowserHttpReader";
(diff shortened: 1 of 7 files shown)


packages/studio-base/src/players/IterablePlayer/BagIterableSource.ts
```diff
@@ -141,7 +143,9 @@ export class BagIterableSource implements IIterableSource {
     };
   }
 
-  async *messageIterator(opt: MessageIteratorArgs): AsyncIterator<Readonly<IteratorResult>> {
+  async *messageIterator(
+    opt: MessageIteratorArgs & { reverse?: boolean },
```
defunctzombie, line 147: Why are we patching this onto message iterator rather than adding it to the IteratorArgs? Seems somewhat messy here when we control the types.

packages/studio-base/src/players/IterablePlayer/BagIterableSource.ts
```diff
@@ -182,4 +186,25 @@ export class BagIterableSource implements IIterableSource {
       }
     }
   }
+
+  async backfillMessages({ topics, time }: BackfillMessagesArgs): Promise<MessageEvent<unknown>[]> {
```
defunctzombie, line 190: Maybe call this `getBackfillMessages` or something that makes it clear its a call to get the messages? I don't know if we follow that pattern anywhere or just make the method call an active verb.

defunctzombie (approved):
Remove the `& { reverse` stuff in BagIterableSource. Rest looks fine and easy to change as we evolve this since its all internal.

wkalt (approved):
DP integration looks fine to me

packages/studio-base/src/players/IterablePlayer/BagIterableSource.ts
jtbandes: This does not apply to other iterators. It is an implementation detail of the bag source. I could separate it into a private method if that is preferable?

Issue #737, opened 2021-04-28
Create UI around the auto-updater


amacneil:
e.g. "check for updates" button, some kind of popup when an update is requires a restart (chrome/linear style), etc

defunctzombie:
I saw a notification on my mac recently when an update was available. We might want to do this in-app to be cross platform compatible.

jtbandes:
🤷 integrating with the native notification system is probably preferable for the user, and an API is [provided by electron](https://www.electronjs.org/docs/latest/tutorial/notifications)

jtbandes:
- https://github.com/foxglove/studio/pull/3057
- https://github.com/foxglove/studio/issues/1025

PR #3057, opened 2022-03-22
Add sidebar UI for Studio updates

jhurliman:
**User-Facing Changes**

Displays a sidebar icon when a new Studio update is available, has been downloaded, or the update check failed

**Description**

The electron-updater events are plumbed through to the Workspace component and used to conditionally add a new sidebar entry when there is app update activity. This could be extended to notify Linux .deb users when a new update is available as well.

**TODO**

- [ ] Finish the UpdaterSidebar component
- [ ] UpdaterSidebar storybook
- [ ] Finalize icon selection for the different states
- [ ] Test the various update events (does this need to be mocked?)
- [ ] Manually check for updates on Linux desktop, where electron-updater cannot auto-update

### DIFF
--- a/packages/studio-base/src/Workspace.tsx
+++ b/packages/studio-base/src/Workspace.tsx
@@ -56,6 +56,7 @@ import Sidebar, { SidebarItem } from "@foxglove/studio-base/components/Sidebar";
 import { SidebarContent } from "@foxglove/studio-base/components/SidebarContent";
 import Stack from "@foxglove/studio-base/components/Stack";
 import { URLStateSyncAdapter } from "@foxglove/studio-base/components/URLStateSyncAdapter";
+import UpdaterSidebar from "@foxglove/studio-base/components/UpdaterSidebar";
 import { useAssets } from "@foxglove/studio-base/context/AssetsContext";
 import ConsoleApiContext from "@foxglove/studio-base/context/ConsoleApiContext";
 import {
@@ -98,6 +99,14 @@ const useStyles = makeStyles({
   },
 });
 
+enum UpdaterState {
+  Idle,
+  Checking,
+  UpdateAvailable,
+  UpdateDownloaded,
+  Error,
+}
+
 type SidebarItemKey =
   | "connection"
   | "add-panel"
@@ -106,6 +115,7 @@ type SidebarItemKey =
   | "extensions"
   | "account"
   | "layouts"
+  | "updater"
   | "preferences"
   | "help";
 
@@ -312,6 +322,29 @@ export default function Workspace(props: WorkspaceProps): JSX.Element {
     }, []),
   );
 
+  const [updaterState, setUpdaterState] = useState(UpdaterState.Idle);
+
+  useNativeAppMenuEvent(
+    "checking-for-update",
+    useCallback(() => setUpdaterState(UpdaterState.Checking), []),
+  );
+  useNativeAppMenuEvent(
+    "update-available",
+    useCallback(() => setUpdaterState(UpdaterState.UpdateAvailable), []),
+  );
+  useNativeAppMenuEvent(
+    "update-not-available",
+    useCallback(() => setUpdaterState(UpdaterState.Idle), []),
+  );
+  useNativeAppMenuEvent(
+    "update-downloaded",
+    useCallback(() => setUpdaterState(UpdaterState.UpdateDownloaded), []),
+  );
+  useNativeAppMenuEvent(
+    "update-error",
+    useCallback(() => setUpdaterState(UpdaterState.Error), []),
+  );
+
   const nativeAppMenu = useNativeAppMenu();
 
   const connectionSources = useMemo(() => {
@@ -490,7 +523,7 @@ export default function Workspace(props: WorkspaceProps): JSX.Element {
   }, []);
 
   const sidebarItems = useMemo<Map<SidebarItemKey, SidebarItem>>(() => {
-    const SIDEBAR_ITEMS = new Map<SidebarItemKey, SidebarItem>([
+    const UPPER_SIDEBAR_ITEMS = new Map<SidebarItemKey, SidebarItem>([
       [
         "connection",
         {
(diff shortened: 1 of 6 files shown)


defunctzombie:
Related: https://github.com/foxglove/studio/issues/737

packages/studio-base/src/Workspace.tsx
```diff
@@ -312,6 +322,29 @@ export default function Workspace(props: WorkspaceProps): JSX.Element {
...
+    useCallback(() => setUpdaterState(UpdaterState.Idle), []),
+  );
+  useNativeAppMenuEvent(
+    "update-downloaded",
+    useCallback(() => setUpdaterState(UpdaterState.UpdateDownloaded), []),
+  );
+  useNativeAppMenuEvent(
+    "update-error",
+    useCallback(() => setUpdaterState(UpdaterState.Error), []),
+  );
```
jtbandes, line 346: How about we reduce it to 1 event called `updater-state`?

jhurliman: @jtbandes the way this event system was architected, we can't pass any arguments.

packages/studio-base/src/Workspace.tsx
```diff
@@ -510,25 +543,59 @@ export default function Workspace(props: WorkspaceProps): JSX.Element {
         { iconName: "SingleColumnEdit", title: "Panel settings", component: PanelSettings },
       ],
       ["variables", { iconName: "Variable2", title: "Variables", component: Variables }],
+    ]);
+    const LOWER_SIDEBAR_ITEMS = new Map<SidebarItemKey, SidebarItem>([
```
jtbandes, line 547: This is not what determines whether the items are on the top or bottom; there is a separate notion of `sidebarBottomItems` that the sidebar uses to position items differently. If you would like the item to be pinned to the bottom you should include it in `sidebarBottomItems`.

packages/studio-base/src/context/NativeAppMenuContext.ts
```diff
@@ -5,17 +5,24 @@
...
   | "open-extensions"
   | "open-help"
   | "open-account"
-  | "open-preferences";
+  // StudioAppUpdater events
+  | "checking-for-update"
+  | "update-available"
+  | "update-not-available"
+  | "update-downloaded"
+  | "update-error";
```
jtbandes, line 26: These don't belong in NativeAppMenuContext because they aren't related to the menus

packages/studio-base/src/Workspace.tsx
jtbandes: It looks like you added the `broadcastEvent` call which can accept args, so it seems like we should be able to plumb args through to the renderer as well?

packages/studio-base/src/context/NativeAppMenuContext.ts
jtbandes: The way I am reading the current code is: we have a `menuBridge` whose event listener methods happen to not automatically prefix their event names with `menu.`, so it is possible to abuse it for listening to any event. It would be best not to use the menuBridge at all for this feature though, because it's supposed to be about menus. Maybe add a method to `desktopBridge` instead? Perhaps @defunctzombie remembers the reasoning for the current separation into several different "bridge" objects rather than one bridge with all our APIs on it?

defunctzombie: Can confirm that `menuBridge` is for dealing only with the menus. This should be on the `desktopBridge`.

defunctzombie: The only things that have sent events up to now have been the menu items which is why it might look like `NativeAppMenu.on()` is our generic name for events from `browserWindow.webContents.send()` but thats only because nothing else has had to send events prior to these changes.

packages/studio-base/src/Workspace.tsx
kakaSheepA: > @jtbandes这个事件系统的架构方式，我们不能传递任何参数。

test

packages/studio-base/src/context/NativeAppMenuContext.ts
kakaSheepA: > 到目前为止，唯一发送事件的是菜单项，这就是为什么它看起来像`NativeAppMenu.on()`我们的事件的通用名称，`browserWindow.webContents.send()`但这只是因为在这些更改之前没有其他东西必须发送事件。

kakaSheepA: > > 项目，这些事件发送是为了，这就是为什么它看起来像是我们唯一`NativeAppMenu.on()`的事件的通用名称，这只是`browserWindow.webContents.send()`因为在此之前没有其他事件必须发送。

test

kakaSheepA: > > > 项目，这些事件发送是为了，这就是为什么它看起来是我们唯一`NativeAppMenu.on()`的事件的通用名称，这只是`browserWindow.webContents.send()`因为在此之前没有其他事件必须发送。
> 
> 测试

ceshi

kakaSheepA: > > > > 项目，这些事件发送是为了，这就是为什么它看起来是我们唯一`NativeAppMenu.on()`的事件的通用名称，这只是`browserWindow.webContents.send()`因为在此之前没有其他事件必须发送。
> > 
> > 
> > 测试
> 
> 切士

zhesojhgodfhjp;gosjdflkgj;dflhj

kakaSheepA: > > > 项目，这些事件发送是为了，这就是为什么它看起来是我们唯一`NativeAppMenu.on()`的事件的通用名称，这只是`browserWindow.webContents.send()`因为在此之前没有其他事件必须发送。
> 
> 测试
11111

Issue #3039, opened 2022-03-17
Image panel's annotations menu jumps around when selecting options


esthersweon:
**Description**
Image panel's annotations menu jumps around when selecting options

- Version: 1.4.0
- OS: macOS

**Steps To Reproduce**

- Open sample data layout
- Click to open annotations menu in right-most image panel
- Select a new annotation topic
- See the menu jump to the left side of the app viewport

https://user-images.githubusercontent.com/6993359/158891730-66d85875-b723-42f8-be18-eee9591f693a.mov

**Expected Behavior**
Annotations menu should stay where it is first displayed.

**Actual Behavior**
Annotations menu jumps around on the screen.

PR #3058, opened 2022-03-22
Use a stable anchor point for image annotations dropdown

defunctzombie:
**User-Facing Changes**
When selecting image annotations, the dropdown does not jump around between selections.

**Description**
When opening the image annotations dropdown, the floating panel toolbar disappears
because it detects the cursor is no longer hovering the panel and is instead hovering
the context menu which is outside the panel.

When the toolbar disappears, it also removes the anchor element for the menu. This
causes the menu to jump screen locations between selection. This change works around this
behavior by using an empty div to anchor the annotations dropdown. The empty div is
always present and keeps the dropdown location stable.

Fixes: #3039

### DIFF
--- a/packages/studio-base/src/panels/ImageView/index.tsx
+++ b/packages/studio-base/src/panels/ImageView/index.tsx
@@ -200,7 +200,9 @@ function ImageView(props: Props) {
     synchronize: shouldSynchronize,
   });
 
-  const markerDropdown = useMemo(() => {
+  const rootRef = useRef<HTMLDivElement>(ReactNull);
+
+  const annotationDropdown = useMemo(() => {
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/panels/ImageView/index.tsx
```diff
@@ -315,7 +307,24 @@ function ImageView(props: Props) {
 
   return (
     <Stack flex="auto" overflow="hidden" position="relative">
-      {toolbar}
+      {/*
+      HACK:
```
jtbandes, line 311: Is this a hack? It seems like a reasonable solution to the problem.

foxymiles (approved):
Seems like a reasonable solution to me.

PR #3060, opened 2022-03-22
Move data source info into Data Source Sidebar

### LINKED ISSUE #996: Migrate Data Source Info panel into Connection sidebar
Now that we have the Connection sidebar, it would be a logical place to show information about the current connection.

I propose removing the Data Source Info panel, and surfacing this information in the Connection sidebar for the active connection. It might need some layout adjustment to make sense given the sidebar width.

image

2metres:
**User-Facing Changes**
Move data source info into Data Source Sidebar once connection is live.

**Description**

https://user-images.githubusercontent.com/924528/159839105-24495525-bfa1-42fc-a9db-9647bea56d5c.mov

Closes #996


jtbandes:
Sidebar is scrollable even when the list of topics is short:

https://user-images.githubusercontent.com/14237/160035069-ae47189f-7879-455e-ba15-bda2681110f1.mov

jtbandes:
Consider adding a tooltip to the "copy to clipboard" button telling me what it's going to copy. I wasn't sure whether to expect the datatype or the topic name to be copied. In hindsight it's definitely more useful to copy the topic name, but the button is centered so that wasn't obvious.

image

jtbandes:
Search behavior is a bit weird — am I searching across the concatenation of topic+datatype?

image
image

jtbandes:
Sidebar jumps around as I type in the search field:

https://user-images.githubusercontent.com/14237/160035304-cf1f920d-4259-43c1-af46-76444bc103a2.mov

jtbandes:
Search field appearance is inconsistent with other search fields in the app:

image
image

packages/studio-base/src/components/DataSourceSidebar/ProblemsList.tsx
```diff
@@ -0,0 +1,67 @@
...
+      );
+    },
+    [modalHost],
+  );
+
+  if (problems.length === 0) {
+    return (
+      <Stack flex="auto" padding={2} fullHeight alignItems="center" justifyContent="center">
+        <Typography align="center" color="text.secondary">
+          No problems found
```
jtbandes, line 40: I think "no problems found" is a strange phrase since I didn't really come to the sidebar _looking_ for problems (unless I already knew there was one).

What do you think of completely hiding the Problems tab if there are no problems?

jtbandes:
Alignment of problems text, badge, and underline feels off:

image

Tab underline appearance is also inconsistent with other tabbed UIs in the app:

image


Long error messages make the sidebar scroll horizontally:

https://user-images.githubusercontent.com/14237/160035696-3d651084-8503-4bf8-84e6-bb94970a7053.mov

packages/studio-base/src/components/CssBaseline.tsx
```diff
@@ -21,15 +21,15 @@ const useStyles = makeStyles((theme) => ({
     },
     div: {
       "::-webkit-scrollbar": {
-        width: "4px",
-        height: "4px",
+        width: 4,
+        height: 4,
```
jtbandes, line 25: just curious, did this have an effect? I'd think it would be px by default.

jtbandes:
Spinner spins off-center:

https://user-images.githubusercontent.com/14237/160036042-a70d8427-516e-4913-9386-6ea51b79cc27.mov

packages/studio-base/src/components/DataSourceSidebar/TopicList.tsx
```diff
@@ -0,0 +1,261 @@
...
+  "&.MuiListItem-dense": {
+    ".MuiListItemText-root": {
+      marginTop: theme.spacing(0.5),
+      marginBottom: theme.spacing(0.5),
+    },
+  },
+  ".MuiListItemSecondaryAction-root": {
+    marginRight: theme.spacing(-1),
+  },
+  "@media (pointer: fine)": {
```
jtbandes, line 97: What's this for?

packages/studio-base/src/components/DataSourceSidebar/TopicList.tsx
```diff
@@ -0,0 +1,261 @@
...
+        <Typography align="center" color="text.secondary">
+          An error occurred
+        </Typography>
+      </Stack>
+    );
+  }
+
+  if (
+    playerPresence === PlayerPresence.INITIALIZING ||
+    playerPresence === PlayerPresence.RECONNECTING
```
jtbandes, line 145: We should think carefully about including "reconnecting" ... when I test with a WebSocket server (`npx @foxglove/ws-protocol-examples@latest sysmon`), if I stop the server, it goes back to the "Skeleton" state. In some ways this makes sense, but in some ways it's annoying because the UI now has this "shimmer" animation that makes it feel like an active loading state, while in reality it's never going to load until I start my server again.

image

Also the skeletons seem to move when I move my mouse over them:

https://user-images.githubusercontent.com/14237/160036573-a4d535ec-fd88-4248-b001-5c178a7f9650.mov

packages/studio-base/src/components/SidebarContent.tsx
```diff
@@ -2,130 +2,113 @@
...
       )}
-      <div
-        className={cx(classes.content, {
-          [classes.noPadding]: noPadding,
-        })}
-      >
-        {children}
-      </div>
-    </div>
+      <ContentWrapper {...{ disablePadding }}>{children}</ContentWrapper>
```
jtbandes, line 94: A little less duplication is not worth it here IMO
```suggestion
      <ContentWrapper disablePadding={disablePadding}>{children}</ContentWrapper>
```

jtbandes (changes requested):

jtbandes:
Maybe not something to tackle now...but we should probably remove "end time" and "duration" for live connections. Or at least rename "end time" to "current time"? Might need to add some `isLive` boolean to the players since I'm not sure we know that information otherwise.

image

2metres:
> Maybe not something to tackle now...but we should probably remove "end time" and "duration" for live connections. Or at least rename "end time" to "current time"? Might need to add some `isLive` boolean to the players since I'm not sure we know that information otherwise.

I've thought about this a few times opening an issue for further discussion: https://github.com/foxglove/studio/issues/3085

packages/studio-base/src/components/DataSourceSidebar/TopicList.tsx
2metres: This is me trying to acknoledge the fact that hide/show on hover is bad for touch devices so by putting `pointer: fine` the hover interaction only appears on devices with "fine" cursors.

packages/studio-base/src/components/CssBaseline.tsx
2metres: Yeah this should be no change just came across it in my travels

packages/studio-base/src/components/DataSourceSidebar/ProblemsList.tsx
2metres: Maybe just this instead? image

jtbandes (approved):
Discussed via huddle, all concerns were either addressed or deemed not important enough to block.

amacneil:
since it's a pretty big change, can we do a live review of this before merging? either tomorrow after team planning, or Wednesday (PT)

amacneil:
follow up items
- start/end time should be hidden when connected to live data source (separate ticket?)
- weird little square in bottom right
- can we auto-select the problems tab when problems appear?

esthersweon:
Still think we could save more space by having one "Time" heading that has a `startTime - endTime (duration)` that intelligently removes duplicate info.

e.g. the below could be displayed as 
```
TIME
2018-07-23 8:28:47.647 - 8:29:06.797 PM PDT (19.1 sec)
```
instead.

Screen Shot 2022-03-30 at 2 39 37 PM

PR #3067, opened 2022-03-23
Simplify platform specific app setup

defunctzombie:
**User-Facing Changes**
Using the App via @foxglove/studio-base is simpler. The user only needs to use one component. Features are setup via properties to the component.

**Description**
This change simplifies using Foxglove Studio as a "component" within an app. Rather than requiring setup of all the various context providers this exposes just one `<App>` component with properties to override platform specific handling.


packages/studio-base/src/ActualApp.tsx
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  return (
+    <AppConfigurationContext.Provider value={appConfiguration}>
+      <ColorSchemeThemeProvider>
+        <GlobalCss />
+        <CssBaseline>
+          <ErrorBoundary>
+            <MultiProvider providers={providers}>
+              <App availableSources={dataSources} deepLinks={[window.location.href]} />;
```
bclinkinbeard, line 86: Would also be nice if we could pass in `appProviders` or something similar, to be included at the end of the providers list in `packages/studio-base/src/App.tsx`.

jtbandes: Do you have providers that you need to be at the _end_ of the list? Otherwise, couldn't you just wrap them around the ActualApp at the top level?

packages/studio-base/src/ActualApp.tsx
```diff
@@ -0,0 +1,93 @@
...
+  dataSources: IDataSourceFactory[];
+  consoleApi: ConsoleApi;
+  currentUser: CurrentUser;
+  layoutStorage: ILayoutStorage;
+  extensionLoader: ExtensionLoader;
+  nativeAppMenu?: INativeAppMenu;
+  nativeWindow?: INativeWindow;
+};
+
+export function ActualApp(props: AppProps): JSX.Element {
```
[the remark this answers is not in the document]
defunctzombie: A better name is given.

defunctzombie:
Let's start with this change and then build from here on what else we want to make configurable.

packages/studio-base/src/App.tsx
```diff
@@ -2,55 +2,90 @@
...
+import HelpInfoProvider from "./providers/HelpInfoProvider";
+import LayoutManagerProvider from "./providers/LayoutManagerProvider";
+import PanelCatalogProvider from "./providers/PanelCatalogProvider";
+import UserProfileLocalStorageProvider from "./providers/UserProfileLocalStorageProvider";
+import { LaunchPreference } from "./screens/LaunchPreference";
+import ConsoleApi from "./services/ConsoleApi";
+import { ILayoutStorage } from "./services/ILayoutStorage";
+import URDFAssetLoader from "./services/URDFAssetLoader";
 
 type AppProps = {
```
bclinkinbeard, line 46: Can this be exported so consumers like `Root` can use it? I don't think `IAppConfiguration` alone is enough to make things configurable for all use cases.

defunctzombie:
@jtbandes Anything else you want to see here before we land the first pass.

desktop/renderer/Root.tsx
```diff
@@ -80,46 +68,39 @@ function AppWrapper() {
...
-    <ConsoleApiRemoteLayoutStorageProvider />,
-    <StudioToastProvider />,
-    <NativeStorageLayoutStorageProvider />,
-    <NativeAppMenuProvider />,
-    <NativeWindowProvider />,
-    <UserProfileLocalStorageProvider />,
-    <ExtensionLoaderProvider />,
-    /* eslint-enable react/jsx-key */
-  ];
+    appConfiguration.addChangeListener(AppSetting.COLOR_SCHEME, handler);
```
jtbandes, line 80: Should we also call the handler once at the beginning? Maybe it's not needed because the desktop app already reads the app config?

jtbandes (approved):

desktop/renderer/Root.tsx
defunctzombie: I think thats right - the desktop app reads the app config - but I am not well versed in this specific color scheme handling so maybe it is still useful to call once? I guess we can try it this way and adjust if we discover otherwise.

PR #3069, opened 2022-03-23
Experimental three.js renderer

jhurliman:
**User-facing changes**

None

**Description**

Adds a new 3D panel based on the [three.js](https://github.com/mrdoob/three.js/) WebGL renderer behind an experimental flag in dev mode. This panel is not ready for general use, and only serves as a development sandbox for now.

Screen Shot 2022-03-24 at 10 58 39 AM


packages/studio-base/src/AppSetting.ts
```diff
@@ -20,6 +20,7 @@ export enum AppSetting {
   // Experimental features
   UNLIMITED_MEMORY_CACHE = "experimental.unlimited-memory-cache",
   SHOW_DEBUG_PANELS = "showDebugPanels",
+  ENABLE_NEW_3D_PANEL = "enableNew3DPanel",
```
defunctzombie, line 23:
```suggestion
  EXPERIMENTAL_3D_PANEL = "experimental.3d-panel",
```

defunctzombie (approved):

siddancha:
This is cool! Just out of curiosity -- why a new 3D panel based on threejs? What use cases do you anticipate would be better served by this panel vs the current 3D panel based on regl-worldview?

I had [previously asked a question](https://foxglove.slack.com/archives/C028UEY858S/p1640647102347100) about Foxglove's design choice for 3D rendering. @amacneil suggested that you prefer regl-worldview or something [even more low-level](https://foxglove.slack.com/archives/C028UEY858S/p1640650618352500?thread_ts=1640647102.347100&cid=C028UEY858S). Curious about the turn towards a higher-level library.

jhurliman:
@siddancha there are a few motivations pushing us to look at alternative rendering approaches.

* Performance - regl-worldview couples the React paradigm with WebGL, which can give you great performance for some workloads but works against you for others. In Studio, new data is flowing in at framerate which causes a lot of temporary object creation. This can lead to frequent GC pauses. regl has also stated they do not intend to migrate to WebGL2, precluding the use of Vertex Array Objects, Non-Power-of-Two textures and signed integer textures, and bitwise operators in GLSL that could help with some of our specific rendering tasks.
* Correctness - regl-worldview uses an sRGB workflow instead of linear, making it difficult to match colors between markers, PNG/JPG textures, and DOM elements. Implementing a linear color workflow when working directly with WebGL is non-trivial, but becomes easier with a higher level engine. Rendering many transparent objects in a scene is another common issue we encounter, and advanced techniques like depth peeling are more feasible on top of a full-featured engine.
* Extensibility - Some developers that are running modified versions of Studio find the three.js API more familiar to work and integrate existing code into than building on top of regl-worldview. Support for a hierarchical scene graph and a large number of geometry builders, assets loaders, materials, etc. makes it easier to build complex custom scene extensions.

In addition to this prototyping, we're keeping an eye on new developments around WebGPU, WASM, more advanced use of Web Workers, and the ecosystems surrounding these technologies. We don't have the capacity right now to build a next-generation rendering engine and the related tooling from scratch, but as community-led efforts mature around these technologies we'll prototype more options.

jon-chuang:
Hello, I am attempting to create an experimental build of Foxglove running atop Tauri. However, we are currently in the brief interlude where the stable release of webkit2gtk (upon which Tauri's Linux WebView is based) does not support webgl2 whereas the upcoming unstable release does.

I have successfully gotten Tauri to run on the unstable release, but it is not a dependency that is easy to support for now.

Could I ask in what way `ThreeDeeRender` relies upong WebGL2-specific features, so that perhaps we may rely on a WebGL capability in the meantime?

Issue #1248, opened 2021-06-16
Feature request: watch playback time in extensions


defunctzombie:
Example use case - showing a line on a plot indicating where the current playback time is. As playback progresses this line would sweep across the plot providing a visual indicator to help the user relate other panel data to the location on the plot.

jhurliman:
@defunctzombie I thought this was added to the extension API?

defunctzombie:
We added _previewTime_ https://github.com/foxglove/studio/blob/main/packages/studio/index.d.ts#L103 which is the value when a user hovers over a plot or map panel location and we want to "click-to-seek" but we did not add currentTime yet.

PR #3071, opened 2022-03-23
Add currentTime to extension API RenderState

jhurliman:
**User-Facing Changes**

- Adds RenderState.currentTime and context.watch("currentTime") to the extension API


defunctzombie (approved):
LGTM - I don't know if we have any tests for the adapter calls - if we do copy/paste one - if not no worries.

defunctzombie:
Related: https://github.com/foxglove/studio/issues/1248

jhurliman:
@esthersweon I think https://foxglove.dev/docs/studio/extensions/panel-api-methods needs to be updated. Maybe avoid listing out the watchable fields and link to a `.ts` file instead?

PR #3078, opened 2022-03-25
Support GeoJSON messages

### LINKED ISSUE #2284: Map panel: add user-defined GeoJSON layer

foxymiles:
**User-Facing Changes**
This adds support for a new `foxglove.GeoJSON` message type that adds GeoJSON markers to maps.

**Description**
This upgrades the map panel to subscribe to topics with the `foxglove.GeoJSON` datatype and to render them alongside the existing marker types.

Screen Shot 2022-03-25 at 10 47 38 AM

Fixes #2284


packages/studio-base/src/panels/Map/MapPanel.tsx
```diff
@@ -290,65 +323,68 @@ function MapPanel(props: MapPanelProps): JSX.Element {
...
-      for (const messageEvent of navMessages) {
-        const point: Point = {
-          lat: messageEvent.message.latitude,
-          lon: messageEvent.message.longitude,
-        };
-
-        return point;
+      for (const messages of [mapMessages, allMapMessages]) {
+        for (const message of messages) {
+          if (!isGeoJSONMessage(message)) {
```
defunctzombie, line 328: Rather than doing the `isGeoJSONMessage` check in the various locations that handle messages, lets partition the messages when we receive them. That can also be a place to normalize. From there the rest of the panel works with defined types.

packages/studio-base/src/panels/Map/MapPanel.tsx
```diff
@@ -407,19 +446,15 @@ function MapPanel(props: MapPanelProps): JSX.Element {
...
-    for (const msgEvent of allNavMessages) {
-      const stamp = toSec(msgEvent.receiveTime);
-      const delta = previewTime - stamp;
-      if (delta < stampDelta && delta >= 0) {
-        stampDelta = delta;
-        event = msgEvent;
-      }
-    }
+    const prevNavMessages = allMapMessages.filter(
+      (m): m is MessageEvent<NavSatFixMsg> =>
```
defunctzombie, line 450: I prefer if we avoid single letter variable names even in small filters like this one.

packages/studio-base/src/panels/Map/types.ts
```diff
@@ -42,3 +48,5 @@ export type NavSatFixMsg = {
   position_covariance?: Matrix3x3;
   position_covariance_type?: NavSatFixPositionCovarianceType;
 };
+
+export type MapPanelMessage = MessageEvent<GeoJSONMessage> | MessageEvent<NavSatFixMsg>;
```
defunctzombie, line 52: I think its better if we do the normalization and partitioning on the inbound messages so anything downstream doesn't have to do dynamic type detection.

defunctzombie (approved):
LGTM - I think we should try normalizing and partition the messages prior to other parts of the panel consuming them so the rest of the code doesn't need to do dynamic detection and we can localize that to one place. This might still mean theres a discriminated union message or something - not sure but thats the only thing that really stood out to me.

packages/studio-base/src/panels/Map/types.ts
foxymiles: I've put some memoized pre-partitioning logic earlier in the component so the messages are already separated.

packages/studio-base/src/panels/Map/MapPanel.tsx
foxymiles: This is gone since we're pre-partitioning now.

packages/studio-base/src/panels/Map/types.ts
```diff
@@ -2,6 +2,12 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
+import { MessageEvent } from "@foxglove/studio";
+
+export type GeoJSONMessage = {
```
jtbandes, line 7: Can you put this in FoxgloveMessages.ts? I moved them in there to try and keep track of all our predefined schemas in one place.

jtbandes (approved):

PR #3081, opened 2022-03-25
[New3D] OccupancyGrid, axis arrows and lines

jhurliman:
**User-Facing Changes**

None

**Description**

This implements slow but correct OccupancyGrid rendering for the experimental 3D panel. The Int8Array grid is converted to a UInt8ClampedArray of RGBA values before uploading to the GPU.

This also extends the CoordinateFrame transform buffer to 60 seconds to mitigate the lack of a "reset" / "seek" event in extensions, and only renders the scene when new messages arrive or the camera moves.

Screen Shot 2022-03-25 at 3 22 57 PM
Screen Shot 2022-03-26 at 2 06 04 PM

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
@@ -15,8 +15,9 @@ import { MaterialCache } from "./MaterialCache";
 import { ModelCache } from "./ModelCache";
 import { FrameAxes } from "./renderables/FrameAxes";
 import { Markers } from "./renderables/Markers";
+import { OccupancyGrids } from "./renderables/OccupancyGrids";
 import { PointClouds } from "./renderables/PointClouds";
-import { Marker, PointCloud2, TF } from "./ros";
+import { Marker, OccupancyGrid, PointCloud2, TF } from "./ros";
 import { TransformTree } from "./transforms/TransformTree";
 
 const log = Logger.getLogger(__filename);
@@ -57,6 +58,7 @@ export class Renderer extends EventEmitter<RendererEvents> {
   renderFrameId: string | undefined;
 
   frameAxes = new FrameAxes(this);
+  occupancyGrids = new OccupancyGrids(this);
   pointClouds = new PointClouds(this);
   markers = new Markers(this);
 
@@ -98,6 +100,7 @@ export class Renderer extends EventEmitter<RendererEvents> {
 
     this.scene = new THREE.Scene();
     this.scene.add(this.frameAxes);
+    this.scene.add(this.occupancyGrids);
     this.scene.add(this.pointClouds);
     this.scene.add(this.markers);
 
@@ -133,6 +136,7 @@ export class Renderer extends EventEmitter<RendererEvents> {
 
   dispose(): void {
     this.frameAxes.dispose();
+    this.occupancyGrids.dispose();
     this.pointClouds.dispose();
     this.markers.dispose();
     this.gl.dispose();
@@ -147,6 +151,10 @@ export class Renderer extends EventEmitter<RendererEvents> {
 
   addTransformMessage(tf: TF): void {
     this.frameAxes.addTransformMessage(tf);
+  }
+
+  addOccupancyGridMessage(topic: string, occupancyGrid: OccupancyGrid): void {
+    this.occupancyGrids.addOccupancyGridMessage(topic, occupancyGrid);
   }
 
   addPointCloud2Message(topic: string, pointCloud: PointCloud2): void {
@@ -189,6 +197,7 @@ export class Renderer extends EventEmitter<RendererEvents> {
     this.materialCache.update(this.input.canvasSize);
 
     this.frameAxes.startFrame(currentTime);
+    this.occupancyGrids.startFrame(currentTime);
     this.pointClouds.startFrame(currentTime);
     this.markers.startFrame(currentTime);
 
--- a/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
@@ -25,6 +25,7 @@ const log = Logger.getLogger(__filename);
 export type RendererEvents = {
   startFrame: (currentTime: bigint, renderer: Renderer) => void;
   endFrame: (currentTime: bigint, renderer: Renderer) => void;
+  cameraMove: (renderer: Renderer) => void;
   renderableSelected: (renderable: THREE.Object3D, renderer: Renderer) => void;
   transformTreeUpdated: (renderer: Renderer) => void;
   showLabel: (labelId: string, labelMarker: Marker, renderer: Renderer) => void;
@@ -132,8 +133,9 @@ export class Renderer extends EventEmitter<RendererEvents> {
     this.camera.lookAt(new THREE.Vector3(0, 0, 0));
 
     this.controls = new OrbitControls(this.camera, this.gl.domElement);
+    this.controls.addEventListener("change", () => this.emit("cameraMove", this));
 
-    this.animationFrame(performance.now());
+    this.animationFrame();
   }
 
   dispose(): void {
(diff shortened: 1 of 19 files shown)


packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
```diff
@@ -252,45 +270,62 @@ export function ThreeDeeRender({ context }: { context: PanelExtensionContext }):
...
+    if (!renderer) {
       return;
     }
 
-    for (const message of messages) {
-      const datatype = topicsToDatatypes.get(message.topic);
-      if (!datatype) {
-        continue;
-      }
+    if (messages) {
```
jtbandes, line 288: reduce indentation with early return

packages/studio-base/src/panels/ThreeDeeRender/renderables/FrameAxes.ts
```diff
@@ -88,30 +128,169 @@ export class FrameAxes extends THREE.Object3D {
...
+    this.axesByFrameId.set(frameId, renderable);
   }
+
+  private _addChildParentLine(renderable: FrameAxisRenderable): void {
+    if (renderable.userData.parentLine) {
+      renderable.remove(renderable.userData.parentLine);
+    }
+    const line = new Line2(FrameAxes.LineGeometry(), this.lineMaterial);
+    line.castShadow = true;
+    line.receiveShadow = false;
```
jtbandes, line 245: do we actually render shadows right now? I'm wondering if it might be confusing to see shadows from these lines which aren't "real objects" in the scene...

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -43,6 +50,7 @@ export class Renderer extends EventEmitter<RendererEvents> {
   lod = DetailLevel.High;
```
jtbandes, line 50: Is it normal for this to be a setting on the renderer/scene as a whole? I would think we'd want to potentially adjust it for objects that are far away from the camera. Or is it going to be a user setting?

packages/studio-base/src/panels/ThreeDeeRender/renderables/FrameAxes.ts
```diff
@@ -88,30 +128,169 @@ export class FrameAxes extends THREE.Object3D {
...
+    // Create a scene graph object to hold the three arrows, a text label, and a
+    // line to the parent frame if it exists
+    const renderable = new THREE.Object3D() as FrameAxisRenderable;
+    renderable.name = frameId;
+    renderable.userData.frameId = frameId;
+    renderable.userData.pose = makePose();
+
+    // Create three arrow shafts
+    const arrowMaterial = standardMaterial(this.renderer.materialCache);
+    const shaftGeometry = FrameAxes.ShaftGeometry(this.renderer.lod);
```
jtbandes, line 174: Since this function caches and reuses a single geometry, do we risk it getting dispose()'d when the axes are disposed and causing a "use-after-free"?

packages/studio-base/src/panels/ThreeDeeRender/renderables/OccupancyGrids.ts
```diff
@@ -0,0 +1,297 @@
...
+  }
+
+  addOccupancyGridMessage(topic: string, occupancyGrid: OccupancyGrid): void {
+    let renderable = this.occupancyGridsByTopic.get(topic);
+    if (!renderable) {
+      renderable = new THREE.Object3D() as OccupancyGridRenderable;
+      renderable.name = topic;
+      renderable.userData.topic = topic;
+
+      // TODO: How do we fetch the stored settings for this topic?
```
jtbandes, line 75: Perhaps it should just be another input argument to this function?
Since these `add*` functions seem to involve adding and/or updating, maybe we should name them `addOrUpdate`?

packages/studio-base/src/panels/ThreeDeeRender/renderables/OccupancyGrids.ts
```diff
@@ -0,0 +1,297 @@
...
+  texture: THREE.DataTexture,
+  renderable: OccupancyGridRenderable,
+): THREE.MeshStandardMaterial | THREE.MeshBasicMaterial {
+  const transparent = occupancyGridHasTransparency(renderable.userData.settings);
+  const material = new THREE.MeshBasicMaterial({
+    map: texture,
+    side: THREE.DoubleSide,
+    // metalness: 0,
+    // roughness: 1,
+    dithering: true,
```
jtbandes, line 282: If we're rendering discrete cells why do we want dithering?

packages/studio-base/src/panels/ThreeDeeRender/renderables/OccupancyGrids.ts
```diff
@@ -0,0 +1,297 @@
...
+  });
+  material.name = `${renderable.userData.topic}:Material`;
```
jtbandes, line 284: what does this end up getting used for?

jtbandes (approved):

packages/studio-base/src/panels/ThreeDeeRender/renderables/FrameAxes.ts
jhurliman: This PR turns the shadow map off, but I set shadow settings meshes so it could be a one-click toggle if someone wanted shadows. The frame axes are also now rendered as 3D arrows with volume so the shadow settings make more sense.

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
jhurliman: It's normally a "max LOD" setting on the renderer as a whole, we just don't have THREE.LOD parent objects yet so this acts as the single scene-wide level of detail for now.

packages/studio-base/src/panels/ThreeDeeRender/renderables/FrameAxes.ts
jhurliman: Disposing a THREE.Mesh does not dispose() the referenced THREE.Geometry automatically for this reason.

packages/studio-base/src/panels/ThreeDeeRender/renderables/OccupancyGrids.ts
```diff
@@ -0,0 +1,297 @@
...
+    updateTexture(texture, occupancyGrid, renderable.userData.settings);
+
+    renderable.scale.set(resolution * width, resolution * height, 1);
+  }
+
+  static Geometry(): THREE.PlaneGeometry {
+    if (!OccupancyGrids.geometry) {
+      OccupancyGrids.geometry = new THREE.PlaneGeometry(1, 1, 1, 1);
+      OccupancyGrids.geometry.translate(0.5, 0.5, 0);
+      OccupancyGrids.geometry.computeBoundingSphere();
```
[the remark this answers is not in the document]
jhurliman: Internal frustum culling in three.js requires bounding spheres, which are not calculated for geometry automatically.

packages/studio-base/src/panels/ThreeDeeRender/renderables/OccupancyGrids.ts
jhurliman: Purely for debugging. It can be pulled out at some point but in early stages I've found it helpful for problems like "where are all these materials coming from? What is this material that is being updated every frame?"

packages/studio-base/src/panels/ThreeDeeRender/renderables/OccupancyGrids.ts
jhurliman: My idea with the naming was you add a message, and under the hood that may add or update (or remove!) a renderable. Maybe `ingestOccupancyGridMessage`? I'm not sure.

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
jhurliman: Changed to maxLod

packages/studio-base/src/panels/ThreeDeeRender/renderables/OccupancyGrids.ts
jhurliman: Fixed

packages/studio-base/src/panels/ThreeDeeRender/renderables/OccupancyGrids.ts
jtbandes: `ingest`, `handle`? `add` is also okay.

Issue #3064, opened 2022-03-22
Clicking seek or pause when using the experimental data platform player is laggy


defunctzombie:
When clicking seek or pause with the experimental data platform player there is a noticeable delay before the action is taken.

Issue #3063, opened 2022-03-22
Experimental data platform does not load my data


defunctzombie:
I have a layout with a plot panel and an image panel. When I open a data platform url nothing happens.

> http://localhost:8080/?ds=foxglove-data-platform&ds.deviceId=dev_2UQTNsiFX3jYbkz6&ds.end=2018-07-24T03%3A29%3A06.797000000Z&ds.start=2018-07-24T03%3A28%3A47.647951000Z&layoutId=5c73cbe5-6be1-468f-b3e2-c38ffdf55ffb&time=2018-07-24T03%3A29%3A06.093226585Z

No images load. No plot data is loaded.

I can see requests being made in the console but nothing visually happens.

PR #3082, opened 2022-03-26
Add abort signal to getBackfillMessages

defunctzombie:
**User-Facing Changes**
For users on Experimental Data Platform player, seeking behavior is improved.

**Description**
When in backfill mode and the user clicks to seek elsewhere, we need to abort the current request
and start a new backfill. This adds an abort controller to signal the abort for the fetch request.

This also adds an emitState prior to any backfill to give immediate feedback in the UI for the new
seek location even if the player needs to buffer some before it can play.

Fixes: #3064 #3063

### DIFF
--- a/packages/studio-base/src/players/FoxgloveDataPlatformPlayer/streamMessages.ts
+++ b/packages/studio-base/src/players/FoxgloveDataPlatformPlayer/streamMessages.ts
@@ -8,7 +8,7 @@ import { isEqual } from "lodash";
 import Logger from "@foxglove/log";
 import { Mcap0StreamReader, Mcap0Types } from "@foxglove/mcap";
 import { loadDecompressHandlers, parseChannel, ParsedChannel } from "@foxglove/mcap-support";
-import { fromNanoSec, isTimeInRangeInclusive, Time, toRFC3339String } from "@foxglove/rostime";
+import { fromNanoSec, Time, toRFC3339String } from "@foxglove/rostime";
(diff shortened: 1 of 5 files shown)


packages/studio-base/src/players/FoxgloveDataPlatformPlayer/streamMessages.ts
```diff
@@ -79,7 +79,8 @@ export default async function* streamMessages({
   if (controller.signal.aborted) {
     return;
   }
-  const response = await fetch(mcapUrl, { signal: controller.signal });
+
+  const response = await fetch(mcapUrl, { signal: controller.signal, cache: "no-cache" });
```
jtbandes, line 83: comment on no-cache?

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
```diff
@@ -372,7 +402,13 @@ export class IterablePlayer implements Player {
             await this._stateIdle();
             break;
           case "seek-backfill":
-            await this._stateSeekBackfill();
+            // We allow aborting requests when moving on to the next state
+            await this._stateSeekBackfill().catch((err) => {
+              if (this._nextState && err instanceof DOMException && err.name === "AbortError") {
+                return;
+              }
+              throw err;
```
jtbandes, line 410: I would move this try/catch into the stateSeekBackfill function.

jtbandes (approved):

PR #1985, opened 2021-10-15
Add Gauge panel

jtbandes:
**User-Facing Changes**
Adds a Gauge panel for displaying numerical values on a colored scale.

image

**Description**
~Blocked on https://github.com/foxglove/studio/issues/2450~

Open questions /cc @defunctzombie @2metres @amacneil @esthersweon:
- Should we allow the user to customize the colors, and if so how? (One option would be allowing them to specify a string that gets added to the CSS `conic-gradient` directive, however this would present challenges since the colors are unfortunately interspersed with angles that we don't want the user to be able to change when customizing colors.)
- Should we allow the min/max range to also be dynamic values based on a message path, rather than fixed values in the panel settings?
- Do we want any other display styles, such as a linear range (think battery meter image) / [level indicator](https://developer.apple.com/design/human-interface-guidelines/macos/indicators/level-indicators/)? Or would these be better as separate panels?

PR #3087, opened 2022-03-29
Layer list component

2metres:
**User-Facing Changes**
Start exposing layer properties in a more attribute editor style

**Description**
Screen Shot 2022-03-29 at 7 39 36 pm


2metres:
Discussion from today:

Start building out changes to Extension API by improving "simple" panels like Map #2550  panel or Gauge Panel #1985 

image

Issue #2974, opened 2022-03-04
Add dataType to MessageEvent in @foxglove/studio


jhurliman:
Incoming `MessageEvent`s: https://github.com/foxglove/studio/blob/1bfcfa924c7038050fcceb714d477921eb2e50d9/packages/studio/index.d.ts#L38-L43 
Would be easier to dispatch to different handlers if they included a `dataType` field in addition to `topic`. Otherwise, I have to manage my own topic->dataType mapping.

defunctzombie:
Yea - I agree - I got annoyed at this when I did the message normalization in Image and Log panel recently.

Kind of a duplicate of https://github.com/foxglove/studio/issues/1923

The way I handled it for now in the panels I needed to tie them together was to use the topics array of the `Topic` type and store that with the extension or panel. Then I used that to lookup the datatype. So it is possible to do today - but was more annoying that it needed to be.

jtbandes:
Done in https://github.com/foxglove/studio/pull/4554

Issue #2214, opened 2021-11-24
Change datatypes API to include enum information


jtbandes:
Currently we use `RosDatatypes` throughout the app. This data structure can include enum constants, but each message field is not annotated with a type indicating which set of constants it comes from. The `__foxglove_enum` hack can be used to add annotations to ROS msg fields matching them with a separate message containing only constants.

Protobuf, however, has native support for enums, so data sources using protobuf schemas will have information about which fields are associated with which enums. The type of `datatypes` and the `enumValuesByDatatypeAndField` will need to change to take full advantage of protobuf enums.

jtbandes:
It turns out Protobuf may not need so much special handling (see https://github.com/foxglove/studio/pull/3130), because `enumValuesByDatatypeAndField` depends on the **order** of the fields and constants. As long as the constants are added to the list immediately before the field itself, the names are correctly picked up at least for this simple case.

It would still be good to address this directly in the schema so that we don't have to rely on this hacky ordering-based check.

defunctzombie:
Sounds like we should look at addressing this with this work:
https://github.com/foxglove/studio/pull/3091

Or at least writing up a high level issue that describes the path from `RosDatatypes` to a ros-agnostic schema that is capable of conveying this information.

PR #3091, opened 2022-03-29
Add new StudioMsgDefinition with optional flag

### LINKED ISSUE #3045: Properties documented as optional are not optional in Node Playground
**Description**
`foxglove.LocationFix` is [documented](https://foxglove.dev/docs/studio/messages/location-fix) as only requiring `latitude` and `longitude`, but Node Playground nodes require all properties before they will compile.

Edit: This appears to apply to all `foxglove.*` schemas.

- Version: 1.4.0
- OS: mac

**Steps To Reproduce**

Nuscenes bag + NP node with the following content:

```typescript

type Output = Message<"foxglove.LocationFix">;

export const output = "/studio_node/gps";

export default function node(event: Input<"/gps">): Output {
  return {
    latitude: 0,
    longitude: 0,
  };
}

```

**Expected Behavior**
Successful node

**Actual Behavior**
```
Type '{ latitude: number; longitude: number; }' is missing the following properties from type '{ latitude: number; longitude: number; altitude: number; position_covariance: Float64Array; position_covariance_type: number; }': altitude, position_covariance, position_covariance_typeTypescript [26,3]
```

image

foxymiles:
**User-Facing Changes**
This adds a concept of optional fields to our internal datatype definitions.

**Description**
This is a minimalistic approach that just extends the existing ROS style definitions with an `optional` flag.

Fixes #3045


jtbandes:
This change (or something like it) will also be a blocker for https://github.com/foxglove/studio/issues/2214.

packages/studio-base/src/types/StudioDatatypes.ts
```diff
@@ -0,0 +1,45 @@
...
+  | string[]
+  | number[]
+  | bigint[]
+  | boolean[]
+  | undefined;
+
+export type StudioMsgField = {
+  type: string;
+  name: string;
+  isComplex?: boolean;
```
jtbandes, line 19: I think we should have a discussion around `type` and `isComplex` fields, and the structure of `StudioDatatypes` as a whole. Structuring it as a `Map<string, StudioMsgDefinition>`, where `name` is used as a lookup key in the map when `isComplex:true`, is a bit limiting.

It works for basic ROS and Protobuf types where there can be a unique name for each datatype (`package_name/Msg` in ROS, `package_name.Msg` in protobuf). One problem, though, is that it's technically possible to make a bag or mcap file that contains multiple versions of the same schema, in which case they'd have the same name, but different fields. Our app doesn't support that and Message Path Syntax will behave in a sub-optimal way (only one of the versions would be used for autocomplete, extracting values from messages for the Plot panel, etc).


A bigger problem (IMO—maybe debatable) is that the structure of this API forces code handling JSON Schemas to create fake `name` values for nested objects, even when there is not an obvious or unique name. So we use `.` to join field names into a key path: https://github.com/foxglove/studio/blob/ea2f7ea7e5c16e4bc1e1afaba4d7f2cc896ee32f/packages/mcap-support/src/parseJsonSchema.ts#L78

This is fragile, because it allows key collisions in the top-level map. For example, this perfectly valid, although adversarially crafted, json schema:

```js
{
    type: "object",
    properties: {
      key: {
        type: "object",
        properties: {
          nested: {
            type: "object",
            properties: {
              value: { type: "string" },
            },
          },
        },
      },
      "key.nested": {
        type: "object",
        properties: {
          value: { type: "number" },
        },
      },
    },
  }
```

Turns into these `datatypes`:

image

Notice that the `value: { type: "string" }` has been lost and overwritten with `value: { type: "number" }`.

So, in the same way that mcap schemas are self-contained, I would recommend that we stop using strings to refer to nested types by name, and instead move toward a richer schema representation that is more similar to json schema, where nested types are actually nested inside each other rather than referring back to the global namespace with a string.

Side note 1: while writing this comment I just learned that JSON Schema also [supports `$defs` and `$ref`](https://json-schema.org/understanding-json-schema/structuring.html) for named reusable types, but our messages don't use this today. And even if we supported this and used it to inform our generated type names, it would be optional, so we would still need to generate fake names in some cases.

Side note 2: even if we fix this datatype issue, we will still have problems if the different schema versions are published on the **same topic name**, because the app also assumes a one-to-one mapping of topic names to schema names.

defunctzombie:
Related: https://github.com/foxglove/studio/issues/1923 https://github.com/foxglove/studio/issues/2974

defunctzombie:
My thoughts on this are that we should move away from the RosDatatype structure and adopt a new `Schema` type which allows us to better represent a more diverse set of IDL capabilities which are missing in the RosDatatype approach yet desired when working with certain types of messages (https://github.com/foxglove/studio/issues/3128).

I've put together an example of the new schema format using an example protobuf file: https://github.com/foxglove/schema-research/blob/main/src/proto/AddressBook.proto and the resulting schema: https://github.com/foxglove/schema-research/blob/main/src/proto/AddressBook.proto.json

Here are the typescript types for the schema: https://github.com/foxglove/schema-research/blob/main/src/schema.ts

I also looked into using JSONSchema as the target schema output. I found JSONSchema lacking in the following ways:
- No way to indicate "maps" which have a distinct key type and a value/item type
- No _nice_ way to indicating enums mappings (which are common in binary formats). JSONSchema has enum values but considers those all one set of strings or numbers with no way to indicate numeric value -> label. There are some approaches that use `oneOf` or `anyOf` within the schema but I found this awkward and more work downstream to understand
- The _ref_ and _defs_ pattern does allow for "self-containing" the schema with any dependency types. However, JSONSchema has a predefined string pattern of indicating the refs which would require downstream users of the schema to unmangle to get the "original" types
- Lacks field ordering in objects. Field ordering is useful for some downstream use cases like autocomplete where we want to show fields in a specific order (x/y/z). This is not a strict requirement.
- A massive footprint. Even when using a subset for our needs we still introduce concepts like `anyOf` that mean all downstream consumers need to understand how to work with the schema.

Known downstream consumers:
- Node playground: Uses the datatype map to produce `Message<”datatype”>` type definitions.
- Message Path Syntax
- Raw message panel (enum values)
- Publishing panel: Creates a JSON skeleton from the message definition
- State transition panel: enum values

-----

In the interest of resolving the bug-fix without boiling the ocean I propose that we keep the existing name of `RosDatatype(s)` and add the `optional` field as this PR does. While not strictly a `rosmsg` capability, the RosDatatype is an internal concept for Studio to convey the schema. This will fix the bug and allow our immediate users to write node playground scripts without specifying the optional fields.

Separately, we should add the new Schema type and start introducing support for the it across the internal users. Once we feel happy with how it has evolved we would introduce it to the extension api surface.

foxymiles:
@jtbandes @defunctzombie I've changed this PR to just add an `optional` field to the existing ROS definitions. Is it better to do it like this as an override in studio or should I make this change in the `@foxglove/rosmsg` package instead?

defunctzombie (approved):

PR #3106, opened 2022-03-30
Set MessageEvent.publishTime, ROS2 /clock support

jhurliman:
**User-Facing Changes**

* ROS2 data source supports the /clock topic
* `MessageEvent.publishTime` is populated for ROS2 and MCAP data sources

**Description**

ROS2 and MCAP data sources now set the publishTime field for MessageEvents. This field is not used internally by Studio yet, but extensions can access it. In the future this could be a third option for the plot X-axis, incoming message sort order, and/or displayed in the Raw Message panel or elsewhere.

### DIFF
--- a/packages/studio-base/src/players/Ros1Player.ts
+++ b/packages/studio-base/src/players/Ros1Player.ts
@@ -88,7 +88,7 @@ export default class Ros1Player implements Player {
     this._metricsCollector = metricsCollector;
     this._url = url;
     this._hostname = hostname;
-    this._start = fromMillis(Date.now());
+    this._start = this._getCurrentTime();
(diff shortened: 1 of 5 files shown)


defunctzombie:
Can you de-couple these two changes - one for the change to the API surface and another for /clock support 🙏

defunctzombie:
LGTM otherwise.

jhurliman:
@defunctzombie to properly decouple it I think it would need to be three PRs: one introducing publishTime, one updating ROS2 player to use publishTime, and one updating MCAP players to use it. Or should I combine the latter two into a single PR? Or some other combination?

defunctzombie (approved):

PR #3107, opened 2022-03-30
Add MessageEvent.publishTime

jhurliman:
**User-Facing Changes**

* Extensions can access `MessageEvent.publishTime` for data sources that support it

**Description**

No data sources actually support `publishTime` yet. That will be added in a follow-up PR.

### DIFF
--- a/packages/studio/index.d.ts
+++ b/packages/studio/index.d.ts
@@ -36,9 +36,30 @@ declare module "@foxglove/studio" {
    * A message event frames message data with the topic and receive time
    */
   export type MessageEvent<T> = Readonly<{
+    /** The topic name this message was received on, i.e. "/some/topic" */
     topic: string;
+    /**
+     * The time in nanoseconds this message was received. This may be set by the
+     * local system clock or the data source, depending on the data source used
+     * and whether time is simulated via a /clock topic or similar mechanism.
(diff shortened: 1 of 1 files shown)


packages/studio/index.d.ts
```diff
@@ -36,9 +36,30 @@ declare module "@foxglove/studio" {
...
+     * The time in nanoseconds this message was originally published. This is
+     * only available for some data sources. The timestamp is often nanoseconds
+     * since the UNIX epoch, but may be relative to another event such as system
+     * boot time or simulation start time depending on the context.
+     */
+    publishTime?: Time;
+    /** The deserialized message as a JavaScript object. */
     message: T;
+    /**
+     * The approximate size of this message in its serialized form. This can be
```
jtbandes, line 60: I don't think this should ever be an approximation, it should be exact, right?

jtbandes (approved):

packages/studio/index.d.ts
jhurliman: This wording might be too conservative, but I wanted to leave a carveout for things like JSON serialization over a websocket where we are probably doing a string length estimation. Or protocols that have per-message overhead and it's not clear whether you count that overhead or not.

defunctzombie:
👍

PR #3114, opened 2022-03-31
Simplify platform specific app setup

defunctzombie:
Re-making the PR to simplify app setup for different platforms. There were reports of not seeing layouts when signed in so the original PR got reverted.


defunctzombie:
Blocked on fixing https://github.com/foxglove/studio/blob/main/web/src/components/ConsoleApiCookieCurrentUserProvider.tsx#L72 which renders the tree even when current user is not set.

packages/studio-base/src/App.tsx
```diff
@@ -61,47 +96,42 @@ function AppContent(props: AppProps): JSX.Element {
...
-      return <AppContent {...props} />;
-    }
-  }
+  return (
+    <AppConfigurationContext.Provider value={appConfiguration}>
+      <ColorSchemeThemeProvider>
+        <GlobalCss />
+        <CssBaseline>
+          <ErrorBoundary>
+            <MaybeLaunchPreference>
```
jtbandes, line 119: Can you just add these to the MultiProvider instead of increasing nesting?

web/src/Root.tsx
```diff
@@ -29,27 +16,24 @@ import {
...
-  );
-  const [enableExperimentalDataPlatformPlayer = false] = useAppConfigurationValue<boolean>(
-    AppSetting.EXPERIMENTAL_DATA_PLATFORM_PLAYER,
-  );
+export function Root({ appConfiguration }: { appConfiguration: IAppConfiguration }): JSX.Element {
+  const enableExperimentalBagPlayer: boolean =
+    (appConfiguration.get(AppSetting.EXPERIMENTAL_BAG_PLAYER) as boolean | undefined) ?? false;
+  const enableExperimentalDataPlatformPlayer: boolean =
+    (appConfiguration.get(AppSetting.EXPERIMENTAL_DATA_PLATFORM_PLAYER) as boolean | undefined) ??
+    false;
```
jtbandes, line 36: It's unfortunate that changing these settings now requires a full refresh.

jtbandes (approved):

packages/studio-base/src/App.tsx
defunctzombie: I guess? Originally that was for providers and this is not a provider. Suppose its become a more generic child list component 🤣 Maybe I'll do it in follow-up changes.

web/src/Root.tsx
defunctzombie: Yea - tho honestly I don't think it's that big a deal? In a way its good since it will force you to re-load  your data sources and ensure use of the player?

PR #3121, opened 2022-03-31
Fix start/end time from coverage calculation in data platform player

foxymiles:
**User-Facing Changes**
This fixes an issue in the data platform player start & end time calculation.

**Description**
This calculation was assuming that the coverage data coming from the API was sorted but it is in fact in random order.

### DIFF
--- a/packages/studio-base/src/players/FoxgloveDataPlatformPlayer/index.ts
+++ b/packages/studio-base/src/players/FoxgloveDataPlatformPlayer/index.ts
@@ -163,9 +163,9 @@ export default class FoxgloveDataPlatformPlayer implements Player {
 
     // Truncate start/end time to coverage range
     const coverageStart = minBy(coverage, (c) => c.start);
-    const converageEnd = maxBy(coverage, (c) => c.end);
-    const coverageStartTime = fromRFC3339String(coverageStart!.start);
-    const coverageEndTime = fromRFC3339String(converageEnd!.end);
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/players/FoxgloveDataPlatformPlayer/index.ts
```diff
@@ -162,22 +162,24 @@ export default class FoxgloveDataPlatformPlayer implements Player {
     }
 
     // Truncate start/end time to coverage range
-    const coverageStart = fromRFC3339String(coverage[0]!.start);
-    const coverageEnd = fromRFC3339String(coverage[coverage.length - 1]!.end);
-    if (!coverageStart || !coverageEnd) {
+    const coverageStart = minBy(coverage, (c) => c.start);
+    const coverageEnd = maxBy(coverage, (c) => c.end);
+    const coverageStartTime = coverageStart ? fromRFC3339String(coverageStart.start) : undefined;
+    const coverageEndTime = coverageEnd ? fromRFC3339String(coverageEnd.end) : undefined;
```
jtbandes, line 168: Length check was done above so these conditionals aren't strictly necessary. If you want to avoid `!` I would move the `coverageStart` & `coverageEnd` definitions above the `if` and then you can do `if (... || !coverageStart || !coverageEnd)` to satisfy TS

jtbandes (approved):

PR #3131, opened 2022-04-01
Prevent rendering CurrentUserProvider until first load

defunctzombie:
**User-Facing Changes**
None

**Description**
The ConsoleApiCookieCurrentUserProvider would prevent rendering children
until the async loading of the user was complete. However, since the currentUser
was actually set in an effect (to allow for unsetting on signout), this would
still result in a render where the currentUser would be undefined followed immediately
by a render where the currentUser is defined.

To avoid this dual rendering, we use a separate state variable to indicate we've
completed the loading. This state variable is set in the same cycle as the currentUser state.


jtbandes (approved):
What's old is new again a9fc932ce2ed878ec0a889ea37ef684e8bc2f887 fe0d96c257a31916b1dd9463f6f39d25aaae3959 bad280c89f30a4ad7522d4958327a08a80d47efc

PR #3132, opened 2022-04-01
[New3D] GPU picking

jhurliman:
**User-Facing Changes**

None

**Description**

Implements GPU picking by rendering a frame, re-rendering a small (9x) patch around the cursor position using a custom fragment shader for all objects that writes the objectId as the color, then extracts the color of the cursor's pixel. The `WebGLRenderer.renderLists` from the last drawn frame is used, simplifying the picking map render pass.

The selected object is highlighted by drawing the scene without it, followed by a dark screen overlay quad, then the selected object. This roughly matches the existing 3D panel's selection highlighting. Once the [three.js EffectComposer color issue](https://github.com/mrdoob/three.js/issues/23019) is resolved this can switch to an OutlinePass which doesn't suffer from the same usability issues.

**TODO**

- [x] Currently blocked on https://github.com/mrdoob/three.js/issues/23019, do our own bg/fg two-pass render to work around it for now
- [x] Custom picking shaders for PointCloud and OccupancyGrid
- [x] LineMaterial support
- [x] Track the currently selected object, trigger an event when selection changes

**Followup PRs**

- Multiple object picking, store a list of candidate selected objects
- Show the object inspector UI
- Second pass for instanced renderables, point clouds

Screen Shot 2022-04-06 at 7 57 26 PMScreen Shot 2022-04-06 at 7 57 21 PM


packages/studio-base/src/panels/ThreeDeeRender/Picker.ts
```diff
@@ -0,0 +1,176 @@
...
+  };
+
+  private processItem = (renderItem: THREE.RenderItem): void => {
+    const object = renderItem.object;
+    const objId = object.id;
+    const material = renderItem.material;
+    const geometry = renderItem.geometry;
+    if (
+      !geometry || // Skip if geometry is not defined
+      material.type === "LineBasicMaterial" || // Skip marker outlines
```
jtbandes, line 112: how does this work if we have some other renderable (e.g. a grid) that uses LineBasicMaterial? It's just not going to be pickable?

packages/studio-base/src/panels/ThreeDeeRender/Picker.ts
```diff
@@ -0,0 +1,176 @@
...
+    const doubleSide = material.side === THREE.DoubleSide ? 1 : 0;
+    const sprite = material.type === "SpriteMaterial" ? 1 : 0;
+    const sizeAttenuation =
+      (material as Partial<THREE.PointsMaterial>).sizeAttenuation === true ? 1 : 0;
+    const index =
+      (useInstancing << 0) |
+      (frontSide << 1) |
+      (doubleSide << 2) |
+      (sprite << 3) |
+      (sizeAttenuation << 4);
```
jtbandes, line 130: why do we use a different material when these properties change?

When using instanced rendering does every instance get its own id? (e.g. clicking on individual points in a point cloud)

packages/studio-base/src/panels/ThreeDeeRender/Picker.ts
jhurliman: Yeah, this would need to get smarter. Maybe we can set a `picking = false` attribute somewhere.

packages/studio-base/src/panels/ThreeDeeRender/Picker.ts
jhurliman: The idea is to use a smallish number of unique materials to render everything. We need a separate material for instanced geometry (just one) because the vertex shader looks different for instancing. And we need a separate one for front-side, back-side, and double-sided rendering (although I might simplify this and just render everything double-sided). Sprites and size-attenuated also need different vertex shaders.

jhurliman: RE: the question about instanced rendering, no it's just a single id. Right now this will select all of the instanced objects, which is not terrible since we don't share instancing across topics yet. If that ever changed, or we decide we want highlighting of single instances, it would need to get smarter.

jtbandes: Feature-parity-wise, the worldview-based 3D panel does use different ids per point in a point cloud:

https://github.com/foxglove/studio/blob/8c0b0e0cf95ceb831716604ddad38e8a13e6503d/packages/studio-base/src/panels/ThreeDimensionalViz/commands/PointClouds/index.tsx#L447

Similarly (though the code path is even more maze-like), CUBE_LIST (and other list) markers use a separate id per instance: https://github.com/foxglove/regl-worldview/blob/92df7aad8d28d15a811d5daaed672261c507e27a/src/commands/Cubes.js#L54

So I think this would be a good goal before New3D 1.0.

packages/studio-base/src/panels/ThreeDeeRender/Picker.ts
jhurliman: Added `picking = false`

packages/studio-base/src/panels/ThreeDeeRender/Picker.ts
jhurliman: Changed this to only support DoubleSided picking materials by default, or renderables can supply their own userData.pickingMaterial. Two-pass picking for instanced geometry and point clouds is on the TODO list now.

packages/studio-base/src/panels/ThreeDeeRender/Input.ts
```diff
@@ -37,17 +39,34 @@ export class Input extends EventEmitter<InputEvents> {
...
-    canvas.addEventListener("click", this.onClick, false);
+    document.addEventListener("keydown", this.onKeyDown);
+    canvas.addEventListener("mousedown", this.onMouseDown);
+    canvas.addEventListener("mousemove", this.onMouseMove);
+    canvas.addEventListener("click", this.onClick);
     canvas.addEventListener("touchstart", this.onTouchStart, { passive: false });
     canvas.addEventListener("touchend", this.onTouchEnd, { passive: false });
     canvas.addEventListener("touchmove", this.onTouchMove, { passive: false });
     canvas.addEventListener("touchcancel", this.onTouchCancel, { passive: false });
     canvas.addEventListener("touchendoutside", this.onTouchEndOutside);
```
jtbandes, line 50: I did a bit of googling and I don't think `touchendoutside` is a real event. It only shows up in results for pixijs and maybe some >10-year-old posts about MobileSafari.

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -237,15 +283,76 @@ export class Renderer extends EventEmitter<RendererEvents> {
...
+  object.layers.set(LAYER_SELECTED);
+  object.traverse((child) => {
+    child.layers.set(LAYER_SELECTED);
+  });
+}
+
+function deselectObject(object: THREE.Object3D) {
+  object.layers.set(0);
+  object.traverse((child) => {
+    child.layers.set(0);
```
jtbandes, line 356:
```suggestion
  object.layers.set(LAYER_DEFAULT);
  object.traverse((child) => {
    child.layers.set(LAYER_DEFAULT);
```

packages/studio-base/src/panels/ThreeDeeRender/renderables/OccupancyGrids.ts
```diff
@@ -284,6 +288,32 @@ function createMaterial(
...
+      void main() {
+        vec4 color = texture2D(map, vUv);
+        if (color.a == 0.0) {
+          discard;
+        }
+        gl_FragColor = objectId;
+      }
+    `,
+    side: THREE.DoubleSide,
+    uniforms: { map: { value: texture }, objectId: { value: [1.0, 1.0, 1.0, 1.0] } },
```
jtbandes, line 313: is it true that the various places where `objectId` is initialized to 1,1,1,1 it's just there so that its presence can be detected later and a real value filled in? In that case, maybe the initial value can be NaN,NaN,NaN,NaN just to be clear that it's never meant to be used?

packages/studio-base/src/panels/ThreeDeeRender/Input.ts
```diff
@@ -37,17 +39,34 @@ export class Input extends EventEmitter<InputEvents> {
     this.resizeObserver = new ResizeObserver(this.onResize);
     this.resizeObserver.observe(parentEl);
 
-    document.addEventListener("keydown", this.onKeyDown, false);
-    canvas.addEventListener("mousedown", this.onMouseDown, false);
-    canvas.addEventListener("mousemove", this.onMouseMove, false);
-    canvas.addEventListener("click", this.onClick, false);
+    document.addEventListener("keydown", this.onKeyDown);
```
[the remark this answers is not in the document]
jhurliman: Nothing yet, I can remove this

packages/studio-base/src/panels/ThreeDeeRender/renderables/markers/materials.ts
```diff
@@ -107,6 +108,21 @@ export function releaseLineMaterial(marker: Marker, materialCache: MaterialCache
   materialCache.release(LineVertexColor.id(lineWidth, transparent));
 }
 
+export function linePickingMaterial(
```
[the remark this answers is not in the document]
jhurliman: Agreed this makes more sense. I want to do a bulk rename of all the create methods in this file to be consistent.

packages/studio-base/src/panels/ThreeDeeRender/renderables/OccupancyGrids.ts
jhurliman: Yes, objectId is only instantiated here so it can have its value set or modified each picking render pass. NaN would be fine too.

PR #3136, opened 2022-04-01
Add support for foxglove.Grid, PointCloud, and LaserScan to 3D panel

jtbandes:
**User-Facing Changes**
Added basic support for `foxglove.Grid`, `foxglove.PointCloud`, and `foxglove.LaserScan` to 3D panel and Node Playground. More features will be added soon when support for these types is added to the experimental 3D panel.

**Description**

`foxglove.Grid`

field | type | description
--- | --- | ---
timestamp | Timestamp | Timestamp of grid
frame_id | string | Frame of reference
pose | Pose | The origin of the grid's corner relative to the frame of reference. The grid is positioned in the xy-plane relative to this origin.
column_count | integer | Number of columns in the grid
cell_size | Vector2D | The size of a single grid cell along x and y axes relative to `pose`
row_stride | integer | Number of bytes between rows in the `data`
cell_stride | integer | Number of bytes between cells within a row in the `data`
fields | Field[] | Fields in the `data`
data | bytes | Grid cell data, interpreted using `fields`, in row-major (y-major) order

`Field`

field | type | description
--- | --- | ---
name | string | Name of the field
offset | integer | Byte offset from the start of the `data` buffer
type | NumericType | Type of data in the field

`enum NumericType`
value | name
--- | ---
1 | UINT8
2 | INT8
3 | UINT16 (little-endian)
4 | INT16 (little-endian)
5 | UINT32 (little-endian)
6 | INT32 (little-endian)
7 | FLOAT32 (little-endian)
8 | FLOAT64 (little-endian)

`foxglove.PointCloud`

field | type | description
--- | --- | ---
timestamp | Timestamp | Timestamp of grid
frame_id | string | Frame of reference
pose | Pose | The origin of the grid relative to the frame of reference. The grid is positioned in the _xy_-plane relative to this origin.
point_stride | integer | Number of bytes between points in the `data`
fields | Field[] | Fields in the `data`
data | bytes | Grid cell data, interpreted using `fields`

`foxglove.LaserScan`

field | type | description
--- | --- | ---
timestamp | Timestamp | Timestamp of grid
frame_id | string | Frame of reference
pose | Pose | The origin of the scan relative to the frame of reference. The points in the scan are positioned in the _xy_-plane relative to this origin. Angles are interpreted as counterclockwise rotations around the _z_ axis with 0 rad being in the _+x_ direction.
start_angle | double | Bearing of the first point of the scan (radians)
end_angle | double | Bearing of the last point of the scan (radians)
ranges | double[] | Distance of detection from the origin. Detections are assumed to be at equally-spaced angles between `start_angle` and `end_angle`.
intensities | double[] | Intensity of detection

Example layout using NP:

```json
{
  "configById": {
    "NodePlayground!kba9ir": {
      "selectedNodeId": "2888af23-9b64-4855-a417-7d5ef20bd705",
      "autoFormatOnSave": true
    },
    "3D Panel!157ua3z": {
      "autoSyncCameraState": false,
      "autoTextBackgroundColor": true,
      "cameraState": {
        "distance": 12.05075138564216,
        "perspective": false,
        "phi": 0.48558613740095957,
        "targetOffset": [
          -0.16540173540685982,
          0.9398363822439408,
          0
        ],
        "thetaOffset": 6.75557945188287,
        "fovy": 0.7853981633974483,
        "near": 0.01,
        "far": 5000
      },
      "checkedKeys": [
        "name:Topics",
        "t:/livox_full_cloud_mapped",
        "t:/pose",
        "ns:/tf:base_link",
        "ns:/tf:RADAR_FRONT",
        "ns:/tf:RADAR_FRONT_LEFT",
        "ns:/tf:RADAR_FRONT_RIGHT",
        "ns:/tf:RADAR_BACK_LEFT",
        "ns:/tf:RADAR_BACK_RIGHT",
        "ns:/tf:LIDAR_TOP",
        "ns:/tf:CAM_FRONT",
        "ns:/tf:CAM_FRONT_RIGHT",
        "ns:/tf:CAM_BACK_RIGHT",
        "ns:/tf:CAM_BACK",
        "ns:/tf:CAM_BACK_LEFT",
        "ns:/tf:CAM_FRONT_LEFT",
        "t:/map",
        "t:/test_grid",
        "t:/test_pointcloud",
        "t:/test_laserscan"
      ],
      "clickToPublishPoseTopic": "/move_base_simple/goal",
      "clickToPublishPointTopic": "/clicked_point",
      "clickToPublishPoseEstimateTopic": "/initialpose",
      "clickToPublishPoseEstimateXDeviation": 0.5,
      "clickToPublishPoseEstimateYDeviation": 0.5,
      "clickToPublishPoseEstimateThetaDeviation": 0.2617993877991494,
      "customBackgroundColor": "#000000",
      "diffModeEnabled": true,
      "expandedKeys": [
        "name:Topics"
      ],
      "followMode": "follow",
      "followTf": "base_link",
      "modifiedNamespaceTopics": [
        "/tf"
      ],
      "pinTopics": false,
      "settingsByKey": {
        "t:/livox_full_cloud_mapped": {},
        "t:/test_grid": {
          "frameLocked": false
        },
        "t:/test_pointcloud": {
          "colorMode": {
            "mode": "rgb"
          },
          "pointSize": 10
        },
        "t:/test_laserscan": {
          "colorMode": {
            "mode": "turbo",
            "colorField": "intensity"
          },
          "pointSize": 10
        }
      },
      "useThemeBackgroundColor": false
    }
  },
  "globalVariables": {},
  "userNodes": {
    "74197db2-41b8-44d8-bf25-70c21f423851": {
      "sourceCode": "import { Input, Message } from \"./types\";\n\nexport const inputs = [\"/pose\"];\n\nexport const output = \"/test_grid\";\n\nexport default function node(event: Input<\"/pose\">): Message<\"foxglove.Grid\"> {\n  const width = 5;\n  const height = 3;\n  const offset = 2;\n  const row_stride = width + offset;\n  const data = new Uint8Array(row_stride * height);\n  const view = new DataView(data.buffer);\n  for (let r = 0; r < height; r++) {\n    for (let c = 0; c < width; c++) {\n      view.setInt8(\n        r * row_stride + c + offset,\n        (100 * (r * width + c)) / (width * height)\n      );\n    }\n  }\n  view.setInt8(offset + 0, -2);\n  view.setInt8(offset + 1, -3);\n  return {\n    frame_id: event.message.header.frame_id,\n    timestamp: event.message.header.stamp,\n    pose: {\n      position: { x: 0, y: 0, z: 1 },\n      orientation: { x: 0, y: 0, z: 0, w: 1 },\n    },\n    fields: [{ name: \"X\", type: 2, offset }],\n    data,\n    cell_size: { x: 1, y: 1 },\n    cell_stride: 1,\n    column_count: width,\n    row_stride,\n  };\n}\n",
      "name": "/studio_node/example grid"
    },
    "633c54b0-ebc6-41a1-9f3c-98fc04dc5deb": {
      "sourceCode": "import { Input, Message } from \"./types\";\n\nexport const inputs = [\"/pose\"];\n\nexport const output = \"/test_pointcloud\";\n\nexport default function node(\n  event: Input<\"/pose\">\n): Message<\"foxglove.PointCloud\"> {\n  const count = 5;\n  const point_stride = 8 + 8 + 8 + 4;\n\n  const data = new Uint8Array(point_stride * count);\n  const view = new DataView(data.buffer);\n  for (let i = 0; i < count; i++) {\n    const offset = i * point_stride;\n    view.setFloat64(offset, 0, true);\n    view.setFloat64(offset + 8, i, true);\n    view.setFloat64(offset + 16, 0, true);\n    const r = Math.round((i / count) * 0xff);\n    const g = 0xff;\n    const b = Math.round((1 - i / count) * 0xff);\n    const rgb = ((((r << 8) | g) << 8) | b) << 8;\n    view.setUint32(offset + 24, rgb);\n  }\n  return {\n    frame_id: event.message.header.frame_id,\n    timestamp: event.message.header.stamp,\n    pose: {\n      position: { x: 0, y: 0, z: 0 },\n      orientation: { x: 0, y: 0, z: 0, w: 1 },\n    },\n    fields: [\n      { name: \"x\", type: 8, offset: 0 },\n      { name: \"y\", type: 8, offset: 8 },\n      { name: \"z\", type: 8, offset: 16 },\n      { name: \"rgb\", type: 6, offset: 24 },\n    ],\n    data,\n    point_stride,\n  };\n}\n",
      "name": "/studio_node/example pointcloud"
    },
    "2888af23-9b64-4855-a417-7d5ef20bd705": {
      "sourceCode": "import { Input, Message } from \"./types\";\n\nexport const inputs = [\"/pose\"];\n\nexport const output = \"/test_laserscan\";\n\nexport default function node(\n  event: Input<\"/pose\">\n): Message<\"foxglove.LaserScan\"> {\n  const count = 10;\n  const ranges = new Array(10).fill(1);\n  const intensities = ranges.map((_, i) => i);\n  return {\n    frame_id: event.message.header.frame_id,\n    timestamp: event.message.header.stamp,\n    pose: {\n      position: { x: 0, y: 0, z: 0 },\n      orientation: { x: 0, y: 0, z: 0, w: 1 },\n    },\n    start_angle: 0,\n    end_angle: Math.PI / 2,\n    ranges: new Float64Array(ranges),\n    intensities: new Float64Array(intensities),\n  };\n}\n",
      "name": "/studio_node/example laserscan"
    }
  },
  "linkedGlobalVariables": [],
  "playbackConfig": {
    "speed": 1,
    "messageOrder": "receiveTime"
  },
  "layout": {
    "direction": "row",
    "first": "NodePlayground!kba9ir",
    "second": "3D Panel!157ua3z",
    "splitPercentage": 59.858657243816246
  }
}
```


packages/studio-base/src/types/FoxgloveMessages.ts
```diff
@@ -116,4 +127,46 @@ export type FoxgloveMessages = {
...
+    point_stride: number;
+    fields: Map<string, { offset: number; type: NumericType }>;
+    compression?: string;
+    data: Uint8Array;
+  };
+
+  "foxglove.AngularScan": {
+    timestamp: { sec: number; nsec: number };
+    frame_id: string;
+    pose: FoxgloveMessages["foxglove.Pose"];
```
jhurliman, line 165: Should this be `origin` to match the other messages, or is it semantically different?

jtbandes: Maybe they should all be `pose`? I saw `origin` in http://docs.ros.org/en/lunar/api/nav_msgs/html/msg/MapMetaData.html and it felt like it made sense for a map, but maybe pose is fine too. I realize now that the comment says "This is the real-world pose of the cell (0,0) in the map" but it doesn't specify which corner of that cell.

packages/studio-base/src/types/FoxgloveMessages.ts
```diff
@@ -116,4 +127,46 @@ export type FoxgloveMessages = {
     frame_id: string;
     poses: Array<FoxgloveMessages["foxglove.Pose"]>;
   };
+
+  "foxglove.Grid": {
```
defunctzombie, line 131: What's the thinking with a grid message or analog in the ROS space?

Regarding the fields:
- What does `compression` mean?
- What is `data`? Does this behave similar to a point cloud message where fields describes the packing?
- Why is fields a map rather than an array? It feels more natural to have an array (order matters) with field name, offset, and type?
- - Do we need row_stride and cell_stride when we have width, height? Why do we need cell_size if we have width and height?

packages/studio-base/src/types/FoxgloveMessages.ts
```diff
@@ -116,4 +127,46 @@ export type FoxgloveMessages = {
...
+
+  "foxglove.AngularScan": {
+    timestamp: { sec: number; nsec: number };
+    frame_id: string;
+    pose: FoxgloveMessages["foxglove.Pose"];
+
+    start_angle: number;
+    end_angle: number;
+
+    fields: Map<string, number[][]>;
```
defunctzombie, line 170: Similar to `Grid` I think an array is more natural. How come this message has fields and no data? Which is a different pattern from Grid or PointCloud.

packages/studio-base/src/types/FoxgloveMessages.ts
```diff
@@ -116,4 +127,46 @@ export type FoxgloveMessages = {
...
+  "foxglove.PointCloud": {
+    timestamp: { sec: number; nsec: number };
+    frame_id: string;
+    origin: FoxgloveMessages["foxglove.Pose"];
+
+    width: number;
+    height: number;
+
+    row_stride: number;
+    point_stride: number;
```
defunctzombie, line 156: Whats the benefit of a separate point_stride if we know the fields?

packages/studio-base/src/types/FoxgloveMessages.ts
jtbandes: > What does compression mean? What is data? Does this behave similar to a point cloud message where fields describes the packing?
Yes, and compression offers the ability to apply something like PNG compression to the data, or any compression, which can help if the data is sparse/repetitive.

> Why is fields a map rather than an array? It feels more natural to have an array (order matters) with field name, offset, and type?

packages/studio-base/src/types/FoxgloveMessages.ts
```diff
@@ -116,4 +127,46 @@ export type FoxgloveMessages = {
...
+    fields: Map<string, { offset: number; type: NumericType }>;
```
defunctzombie, line 157: I think an array is more natural here.

packages/studio-base/src/types/FoxgloveMessages.ts
jtbandes: > What does compression mean? What is data? Does this behave similar to a point cloud message where fields describes the packing?

Yes, and compression offers the ability to apply something like PNG compression to the data, or any compression, which can help if the data is sparse/repetitive.

> Why is fields a map rather than an array? It feels more natural to have an array (order matters) with field name, offset, and type?

Order does not matter if `offset` is specified.

> Do we need row_stride and cell_stride when we have width, height?

This allows for padding. Not sure how important that is.

> Why do we need cell_size if we have width and height?

width and height are number of cells.

packages/studio-base/src/types/FoxgloveMessages.ts
jtbandes: Padding

packages/studio-base/src/types/FoxgloveMessages.ts
jtbandes: > What's the thinking with a grid message or analog in the ROS space?

This is like `nav_msgs/OccupancyGrid`, but not limited to integer values in a specific range. One could imagine including height data in their grid, or RGBA colors.

packages/studio-base/src/types/FoxgloveMessages.ts
defunctzombie: Origin feels more natural when we expect other fields in the message to be relative to this origin and this origin relative to the frame itself.

Maybe `pose` conveys that same notion. We should be consistent unless there's math or stronger precedent for what someone might expect. ROS precedent is decent argument but to me isn't enough.

defunctzombie:
Grid:

- Are the x/y in frame space?
```
cell_size | Vector2D | The x and y size of a single grid cell
```

- Remove compression. We don't support it and we can come back to this feature request when there's an ask.

- Is this bytes _between_ rows or bytes for the entire row?
```
row_stride | integer | Number of bytes between rows in the data
```

- I think fields should be an array. I don't see the benefit of making it a map and we have precedent for array.

PointCloud:

- This is bytes for each "point" rather than between points?
```
point_stride | integer | Number of bytes between points in the data
```

- Remove compression

- Count is implied from point stride and size of data. Useful to re-state explicitly? Would we think to error if they don't match?

AngularScan:

- Point cloud can represent this data so why have another datatype for arrays of points with variable number of attributes (fields).

----

What made you pick `pose` instead of `origin` ?

jtbandes:
> Are the x/y in frame space?

Updated wording.

> [Grid/PointCloud] Is this bytes _between_ rows or bytes for the entire row? / This is bytes for each "point" rather than between points?

It is the offset from one row/point to the next, commonly known as "stride". https://en.wikipedia.org/wiki/Stride_of_an_array

> [PointCloud] Count is implied from point stride and size of data. Useful to re-state explicitly? Would we think to error if they don't match?

This could be an error, or we could just remove `count`, any preference?

> [AngularScan] Point cloud can represent this data so why have another datatype for arrays of points with variable number of attributes (fields).

The coordinate system of AngularScan is different (polar instead of cartesian).

The variable number of values per field is an attempt to subsume http://docs.ros.org/en/melodic/api/sensor_msgs/html/msg/MultiEchoLaserScan.html.

> What made you pick `pose` instead of `origin` ?

If I had to pick one, I went for `pose` because that's what we use in our other types.

defunctzombie:
> This could be an error, or we could just remove `count`, any preference?

The only thing that comes to mind is if you wanted to plot the number of points or surface this data through a method that doesn't understand how to compute this. I would say drop it and see what happens.

defunctzombie:
> The coordinate system of AngularScan is different (polar instead of cartesian).

👍

jtbandes:
Here is an alternative we could do that would not include MultiEcho features:



`foxglove.AngularScan`

field | type | description
--- | --- | ---
timestamp | Timestamp | Timestamp of grid
frame_id | string | Frame of reference
pose | Pose | The origin of the scan relative to the frame of reference. The points in the scan are positioned in the _xy_-plane relative to this origin. Angles are interpreted as counterclockwise rotations around the _z_ axis with 0 rad being in the _+x_ direction.
start_angle | double | Bearing of the first point of the scan (radians)
end_angle | double | Bearing of the last point of the scan (radians)
point_step | integer | Number of bytes between points in the `data`
fields | Field[] | Fields in the `data`
data | bytes | Scan data, interpreted using `fields`

defunctzombie (approved):

PR #3112, opened 2022-03-31
feat: handle qos like rosbag2 in ros2 player

wep21:
**User-Facing Changes**
handle qos like rosbag2 implementation in ros2 player
https://github.com/ros2/rosbag2/blob/e156eb9e38c377df33e5075ab7fa48f3b52eaa12/rosbag2_transport/src/rosbag2_transport/qos.cpp#L106-L173

**Description**
If all of endpoints has reliable reliability, subscriber uses reliable reliability, else subscriber uses best effort reliability.
If all of endpoints has transient_local durability, subscriber uses transient_local durability, else subscriber uses volatile durability.

PR #3137, opened 2022-04-01
Set ROS2 subscription QoS using the same rules as rosbag2

jhurliman:
**User-Facing Changes**

- Fixed ROS2 native subscriptions to multiple publishers with different QoS profiles

**Description**

Credit to @wep21 for the original PR (https://github.com/foxglove/studio/pull/3112).

Handle qos like rosbag2 implementation in ros2 player
https://github.com/ros2/rosbag2/blob/e156eb9e38c377df33e5075ab7fa48f3b52eaa12/rosbag2_transport/src/rosbag2_transport/qos.cpp#L106-L173

If all publishing endpoints for a topic have Reliable reliability, the subscriber uses Reliable, else subscriber uses BestEffort. If all endpoints have TransientLocal durability, the subscriber uses TransientLocal, else subscriber uses Volatile.


wep21:
@jhurliman Thank you for refactoring! It looks good to me.

PR #3144, opened 2022-04-04
Stabilize selector in topic list

foxymiles:
**User-Facing Changes**
None

**Description**
This stabilizes the message pipeline selector used in the `TopicList` component to silence warnings and avoid unnecessary rerenders.

Screen Shot 2022-04-04 at 9 34 50 AM


foxymiles:
> We should lift the function to be stable. Right now the actual selector function is new each render and I think that's going to run the selector tho it will product the same result which won't trigger re-render iirc so it is not as bad.
> 
> I'd follow the same pattern as the `selectPlayerPresence` right above it.

Good point. I've added a stable selector function.

PR #3150, opened 2022-04-06
Implement settings tree and editor

foxymiles:
**User-Facing Changes**
This adds a new `SettingsTree` panel state structure and automatic UI for editing it.

**Description**
This adds a new tree-like settings structure called `SettingsTree` and a data-driven UI to manipulate it. To opt-in to this new panel settings management approach panels can push UI descriptions in the form of a `SettingsTree` to the new `publishPanelSettingsTree` method in `PanelExtensionContext`.

```typescript
type SettingsTree = {
    actionHandler: (action: SettingsTreeAction) => void;
    settings: SettingsTreeNode
}
```

The action handler is implemented by each panel to apply actions from the editor to their internal state and to then publish and updated UI description.

- [x] Fix handling of undefined values.
- [x] Fix color picker.
- [ ] Fix gradient picker.
- [x] Add step to number input.
- [x] Remove from extension panel interface for now.

Screen Shot 2022-04-06 at 12 41 16 PM
Screen Shot 2022-04-06 at 12 41 26 PM


packages/studio/index.d.ts
```diff
@@ -96,6 +144,11 @@ declare module "@foxglove/studio" {
   }
 
   export interface RenderState {
+    /**
+     * The panel configuration.
+     */
+    configuration?: unknown;
```
defunctzombie, line 150: Guessing this is the configuration that comes back from the settings? Whats the general shape of this object? Would it be better for us to provide an interface for getting specific settings values? It feels maybe fragile to have this fully `unknown`? Wondering if we can do better for end users.

foxymiles: Handling the old & new formats is going to be one of the pain points of this transition. I think we're going to have to migrate settings from the old format to the new format on a per-panel basis. Seems like the best we can do is a `Record<string, unknown>` which is how this is typed internally in a few places.

Would doing better for end users mean bundling some kind of minimal panel SDK or utility library with the panel extension template project?

packages/studio/index.d.ts
```diff
@@ -166,6 +219,11 @@ declare module "@foxglove/studio" {
      */
     saveState: (state: Partial<unknown>) => void;
 
+    /**
+     * Register a change interceptor.
+     */
+    setSettingsActionInterceptor: (interceptor: SettingsTreeActionInterceptor) => void;
```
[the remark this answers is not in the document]
foxymiles: Yeah the interceptor can do whatever it wants with the incoming action. This is the place where a panel would do things like change the set of fields that are displayed when a user switches the coloring method of a point cloud or rejects an update that breaks the semantic constraints across multiple values.

packages/studio-base/src/panels/Teleop/TeleopPanel.tsx
```diff
@@ -90,13 +170,19 @@ function TeleopPanel(props: TeleopPanelProps): JSX.Element {
...
       if (renderState.colorScheme) {
         setColorScheme(renderState.colorScheme);
       }
     };
-  }, [context]);
+  }, [context, settingsActionHandler]);
+
+  useEffect(() => {
+    const tree = buildSettingsTree(config, topics);
+    context.publishPanelSettingsTree({ settings: tree, actionHandler: settingsActionHandler });
```
defunctzombie, line 183: Consider alternative names for this that don't use the word `publish`. `publish` is somewhat a loaded term in our domain.

packages/studio-base/src/panels/Teleop/TeleopPanel.tsx
```diff
@@ -2,25 +2,108 @@
...
+  publishRate: number;
+  upButton: { field: string; value: number };
+  downButton: { field: string; value: number };
+  leftButton: { field: string; value: number };
+  rightButton: { field: string; value: number };
+};
+
+function buildSettingsTree(config: Config, topics: readonly Topic[]): SettingsTreeNode {
+  return {
+    fields: {
```
defunctzombie, line 45: Should the root node have special handling?

packages/studio-base/src/panels/Teleop/TeleopPanel.tsx
```diff
@@ -2,25 +2,108 @@
...
+      publishRate: { label: "Publish Rate", input: "number", value: config.publishRate },
+      topic: {
+        label: "Topic",
+        input: "autocomplete",
```
defunctzombie, line 49: maybe we have a `topic` input type that handles this automatically and would be different from `messagepath`

defunctzombie: should this be `type: topic` rather than `input: "topic"`

packages/studio-base/src/panels/Teleop/TeleopPanel.tsx
```diff
@@ -70,17 +154,13 @@ function TeleopPanel(props: TeleopPanelProps): JSX.Element {
...
-    (newConfig: Config) => {
-      setConfig(newConfig);
-      saveState(newConfig);
-    },
-    [saveState],
-  );
+  const settingsActionHandler = useCallback((action: SettingsTreeAction) => {
+    setConfig((previous) => {
+      const newConfig = { ...previous };
+      set(newConfig, action.payload.path, action.payload.value);
```
defunctzombie, line 160: if (action.payload.input === "topic") ... then `action.payload.value` should narrow to `string`

defunctzombie, line 160: What forms does the `path` take? What are some examples? What would an example be for an array of fields (if thats a thing)?

packages/studio-base/src/panels/Teleop/TeleopPanel.tsx
```diff
@@ -2,25 +2,108 @@
...
+            input: "select",
+            value: config.upButton.field,
+            options: geometryMsgOptions,
+          },
+          value: { label: "Value", input: "number", value: config.downButton.value },
+        },
+      },
+      leftButton: {
+        label: "Left Button",
+        fields: {
```
defunctzombie, line 81: consider `attributes` ?

jhurliman:
The "see docs _here_" section seems out of place and takes up a lot of vertical real estate.

When I click _here_ then click the back button, it takes me to the root of the documentation. I can't get back to the settings.

jhurliman:
`undefined` values are rendered strangely. They show up as some predefined value determined by the UI control(?) such as black for color (with no hex code shown) or `0` for the number input. We need to come up with a way to signal "default" or "unset" for each control type, don't show a default (and probably incorrect) value.

Color picker is broken (nothing pops up when I click on it). Trying to type a hex code in crashes the sidebar with the first keystroke.

Gradient picker is incorrect. It shows a list of predefined list of color maps, not gradients, and doesn't allow me to choose start/end colors. The value returned by this input should be `[color, color]`.

jhurliman:
Numeric input needs an optional `step` (right now it is always 1).

packages/studio-base/src/components/SettingsTreeEditor/index.stories.tsx
```diff
@@ -0,0 +1,283 @@
...
+          label: "Frame lock",
+          input: "boolean",
+          value: false,
+          help: "When enabled, the map will not be updated when the robot moves.",
+        },
+      },
+    },
+    topics: {
+      label: "Topics",
+      children: {
```
defunctzombie, line 126: What would happen if I had `children` and `fields` for this node?

packages/studio-base/src/components/SettingsTreeEditor/index.stories.tsx
```diff
@@ -0,0 +1,283 @@
...
+
+import { SettingsTreeNode, SettingsTreeFieldValue, SettingsTreeAction } from "./types";
+
+export default {
+  title: "components/SettingsTreeEditor",
+  component: SettingsTreeEditor,
+};
+
+const DefaultSettings: SettingsTreeNode = {
+  children: {
```
defunctzombie, line 22: It feels somewhat strange that I immediately need to nest all my settings within a `children` field. I guess one thing this does allow for is a top level `fields` that are not in a collapsable group (i.e. always shown) ? I could see that being useful for something like the "follow frame" setting on the 3d panel.

foxymiles: Yeah that was the use case this was intended to address.

packages/studio-base/src/components/SettingsTreeEditor/index.stories.tsx
foxymiles: It would look like this:

Screen Shot 2022-04-08 at 12 15 10 PM

defunctzombie:
Not something we should do for this PR but I do want to understand if we would be able to implement this functionality in the future with the interfaces and types defined in this PR.

> Allow a user to re-arrange the order of items in the sidebar. In rviz, the sidebar is not pre-populated with topics but rather the user adds the topics they want to visualize. Our app works differently today since we show all visualizable topics, but maybe we want to allow for users to group these into their own collapse groups.

> With the Plot panel, maybe the user wants to re-arrange the order of the series in the sidebar (not sure its that useful to do), but if a panel did want to allow users to do such re-arranging how could we expose that event to the panels?

foxymiles:
> Not something we should do for this PR but I do want to understand if we would be able to implement this functionality in the future with the interfaces and types defined in this PR.
> 
> > Allow a user to re-arrange the order of items in the sidebar. In rviz, the sidebar is not pre-populated with topics but rather the user adds the topics they want to visualize. Our app works differently today since we show all visualizable topics, but maybe we want to allow for users to group these into their own collapse groups.
> 
> > With the Plot panel, maybe the user wants to re-arrange the order of the series in the sidebar (not sure its that useful to do), but if a panel did want to allow users to do such re-arranging how could we expose that event to the panels?

I think we could handle that with new action types. Currently we only have one action but we can add more to implement new editing features:

```typescript
export type SettingsTreeAction = {
  action: "update";
  payload: { path: readonly string[]; value: unknown };
};
```

packages/studio-base/src/panels/Teleop/TeleopPanel.tsx
foxymiles: I wanted to pick names that emphasize that this is describing a UI, not a data structure. Along those lines maybe it would make sense to change `fields` to `inputs`?

packages/studio-base/src/panels/Teleop/TeleopPanel.tsx
foxymiles: I've added some more typing constraints to make this narrowing possible.

foxymiles: `path` would be the series of string keys need to identify a value in the source panel's state. It's free to interpret this however it likes.

I haven't directly addressed the array case yet but it would be straightforward to use stringified integer keys like `["buttons", "1", "direction"]`.

packages/studio-base/src/panels/Teleop/TeleopPanel.tsx
foxymiles: I've renamed this to `updatePanelSettingsTree`.

packages/studio-base/src/panels/Teleop/TeleopPanel.tsx
foxymiles: Possibly. The current types allow us to add new top-level fields if we want to expand the semantics of the root node.

2metres (approved):

Issue #3088, opened 2022-03-29
Add .obj mesh support to mesh markers and robot descriptions


horverno:
**Description**

Connected to local ROS, the 3D marker topics and the URDF does not load.

- Version: Foxglove v1.4.0 / v1.5.1 / v1.6.0 and `ROS melodic` / `ROS noetic`
- OS: `Ubuntu 18.04` / `Ubuntu 20.04` 

**Steps To Reproduce**

3d topic:
```
cd ~/catkin_ws/src
git clone https://github.com/szenergy/rviz_markers
catkin build rviz_markers
source ~/catkin_ws/devel/setup.bash 
roscore
roslaunch rviz_markers demo1.launch
```
urdf:

```
cd ~/catkin_ws/src
git clone https://github.com/szenergy/nissan_leaf_ros
catkin build nissan_bringup
source ~/catkin_ws/devel/setup.bash 
roscore
roslaunch nissan_bringup 3d_nissan.launch 
roslaunch nissan_bringup tf_nissanleaf_statictf.launch
roslaunch nissan_bringup rviz00.launch
```

**Expected Behavior**

Display 3D data similar to RVIZ.

**Actual Behavior**

![foxglove_01_a](https://user-images.githubusercontent.com/11504709/160600912-44412c00-75b0-49c7-aadf-0992069146d9.png)

![foxglove_02](https://user-images.githubusercontent.com/11504709/160601884-8e81c7d5-010b-41a4-bf0b-649b1b532326.png)

i also tried to `/robot_description` to `package://nissan_bringup/param/nissan_leaf_robotdescription.ref.xml`, it did not solve the issue.

![foxglove_try_02_bb](https://user-images.githubusercontent.com/11504709/160601047-711962cf-bba7-4d84-a199-073a54be1c67.png)

Thank you!

jtbandes:
Hi, can you please share more details on the error messages from the developer console? (Press Alt key > View > Advanced)

It seems like the problem might be that the `nissan_bringup` package is not in your `ROS_PACKAGE_PATH` (that's why you get the "404" errors when trying to load the meshes). You can edit the `ROS_PACKAGE_PATH` by updating the environment variable before running Studio (e.g. launching Studio after `source ~/catkin_ws/devel/setup.bash`), or directly in the app settings (bottom button in the sidebar).

horverno:
Hi @jtbandes 

Thanks for the quick reply, i tried your suggestion sadly it did not worked.

The errors are the following:

```
[...]

PR #3157, opened 2022-04-07
Recursively search ROS_PACKAGE_PATH in findRosPackage

defunctzombie:
**User-Facing Changes**
Users can follow ROS conventions and set the ROS_PACKAGE_PATH to the root of their workspace. Studio (desktop only) will traverse this path to find all packages within it when resolving `package://` urls.

**Description**
ROS package:// urls follow the convention of `package://<package name>/<path to file>`. In ROS projects, it is common to have a monorepo with several packages folders. This folder would be in the ROS workspace. When setting the ROS_PACKAGE_PATH environment variable or app setting to point to the workspace, the correct behavior for finding packages is to look recursively at the ROS_PACKAGE_PATH location and any locations below it to find the package.

This change updates our package lookup logic to use a recursive search on the ROS_PACKAGE_PATH. This allows finding packages which are within parent folders under the ROS_PACKAGE_PATH.

Fixes: #3088

### DIFF
--- a/desktop/main/rosPackageResources.ts
+++ b/desktop/main/rosPackageResources.ts
@@ -40,30 +40,39 @@ export async function rosPackageNameAtPath(packagePath: string): Promise<string 
 }
 
 /**
- * Return a map of ROS package names to their absolute paths.
+ * Find ROS package named `pkg` within the `rootPath`
  *
- * rootPath is searched recursively to find all packages below the path
+ * rootPath is searched recursively
  */
-async function listRosPackages(rootPath: string): Promise<Map<string, string>> {
+export async function findRosPackageInRoot(
+  pkg: string,
+  rootPath: string,
+): Promise<string | undefined> {
   const packagePaths = await fs.readdir(rootPath, { withFileTypes: true });
   log.debug(`searching ${rootPath} for packages`);
 
-  const packages = new Map<string, string>();
-  for (const packagePath of packagePaths) {
+  const directories = packagePaths.filter((pkgPath) => pkgPath.isDirectory());
+
+  // Check any of our immediate folders for the package
+  for (const packagePath of directories) {
     const absolutePath = path.join(rootPath, packagePath.name);
 
     const packageName = await rosPackageNameAtPath(absolutePath);
-    if (packageName) {
-      packages.set(packageName, absolutePath);
-    } else if (packagePath.isDirectory()) {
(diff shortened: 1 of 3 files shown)


desktop/main/rosPackageResources.ts
```diff
@@ -41,29 +41,28 @@ export async function rosPackageNameAtPath(packagePath: string): Promise<string
...
-  for (const { name, absolutePath } of packagesArray) {
-    if (name != undefined) {
-      packages.set(name, absolutePath);
+  for (const packagePath of packagePaths) {
+    const absolutePath = path.join(rootPath, packagePath.name);
+
+    const packageName = await rosPackageNameAtPath(absolutePath);
+    if (packageName) {
+      packages.set(packageName, absolutePath);
+    } else if (packagePath.isDirectory()) {
```
jtbandes, line 58: Can there be packages that are valid (have a package.xml) but also contain other packages? If so, wouldn't you need to do this recursion regardless of whether rosPackageNameAtPath returned something?

defunctzombie: I thought about this for a moment but somehow don't think I've ever come across this in the wild. My inclination here is to go with the version as written here and if a user comes to us with an example we can revisit. I tried finding more details/spec on this recursive search concept but have not found anything definitive.

My main concern here is actually the cost of recursion. It could end up being quite expensive to do this search in this way. I think the change I'd want to make is to have `findPackage` do a breath first search and stop once it finds the package rather than doing a `listPackages` to get the whole list. This would tend to limit the search depth and filesystem traversal in the cases where the package does exist. If it doesn't exist then searching would still exhaust the tree.

defunctzombie:
I've updated the findRosPackage logic to stop once it has found the package. The check isn't really a breath first search since it recurses into the directories after checking the current level of directories at each step. Hard to know how much of a problem this is in practice without a large workspace to test on.

jhurliman (approved):

desktop/main/fixtures/packages/foo/package.xml
```diff
@@ -5,5 +5,5 @@
     This package provides foo capability.
   </description>
   <maintainer email="john@example.com">John Doe</maintainer>
-  <license>MOT</license>
+  <license>MIT</license>
```
jhurliman, line 8: 🤣

Issue #3158, opened 2022-04-07
Experimental data platform player is unusable with this url


defunctzombie:
https://studio.foxglove.dev/?ds=foxglove-data-platform&ds.deviceId=dev_jCuXYeFwCkZowpHs&ds.end=2021-03-22T15%3A09%3A18.000000000Z&ds.start=2021-03-22T15%3A03%3A38.381000000Z&layoutId=28cdd2dd-31de-4339-8386-ed0db0dfffde&time=2021-03-22T15%3A05%3A06.041352510Z

I can't click to seek.
I can't play.

From what I've been able to debug the player is making requests constantly.

defunctzombie:
Might be related? https://github.com/foxglove/studio/issues/3063

defunctzombie:
https://github.com/foxglove/studio/pull/3161

defunctzombie:
My local testing

http://localhost:8080/?ds=foxglove-data-platform&ds.deviceId=dev_K2Bh2eMjUVEs0FYO&ds.end=2021-03-22T15%3A09%3A18.000000000Z&ds.start=2021-03-22T15%3A03%3A38.381000000Z&layoutId=89d07096-bec3-4d65-8da6-e2b545c42e1d&time=2021-03-22T15%3A05%3A12.751213368Z

PR #3161, opened 2022-04-08
Use bounded iterator ranges with IterablePlayer

defunctzombie:
**User-Facing Changes**
Users of the experimental data platform player are able to use the Studio interface even when their data range does not have any data.

**Description**
Prior to this change the iterable player (what we publicly call the Experimental Player) had a behavior where it would appear to "stall" during initialization and seeking. This stall occurred even on player startup. The player would enter the stateStartPlay method which is responsible for playing a few milliseconds of data to put some messages into the pipeline. To play these messages the player created a message iterator with some start time.

The data platform source would produce a message iterator for this start time and start to stream messages. However, in some cases there might not be any messages to yield from the iterator. In that situation, the message iterator logic would finish the current stream, slide to the next time window and try streaming again. This would continue until the end of the entire requested dataset range was reached with not a single message emitted. Since no message is emitted the player has no opportunity to bail on any of these many stream requests. Even tho it only needed a few milliseconds of data, internally the message iterator went through the entire dataset.

This change updates the IIterableSource interface and player to provide an _end_ time for all messageIterator requests. This allows the player to tell the messageIterator that it need not continue trying to find messages beyond a certain time. The data platform source uses this to avoid making new stream requests after this time in the event no messages are returned.

Fixes: #3158
Fixes: #3063

### DIFF
--- a/packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
+++ b/packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
@@ -470,6 +470,7 @@ export class IterablePlayer implements Player {
     this._messages = [];
 
     const messageEvents: MessageEvent<unknown>[] = [];
+
     for (;;) {
       const result = await iterator.next();
       if (result.done === true) {
@@ -480,6 +481,7 @@ export class IterablePlayer implements Player {
       // This usually happens when seeking before the initial load is complete
       if (this._nextState) {
         log.info("Exit startPlay for new state");
+        void iterator.return?.();
         return;
       }
 
@@ -488,13 +490,15 @@ export class IterablePlayer implements Player {
         continue;
       }
 
+      // Just in case the iterator decides it is going to ignore our _end_ param
       if (compare(iterResult.msgEvent.receiveTime, stopTime) > 0) {
-        this._lastMessage = iterResult.msgEvent;
         break;
       }
 
       messageEvents.push(iterResult.msgEvent);
     }
+
+    void iterator.return?.();
 
     this._currentTime = stopTime;
     this._messages = messageEvents;
@@ -712,7 +716,11 @@ export class IterablePlayer implements Player {
 
           log.debug("Initializing forward iterator from", next);
 
-          const iteratorEnd = add(next, fromNanoSec(BigInt(this._iteratorDurationNanos)));
+          const iteratorEnd = clampTime(
+            add(next, fromNanoSec(BigInt(this._iteratorDurationNanos))),
+            next,
+            this._end,
+          );
           this._tickIterator = this._iterableSource.messageIterator({
             topics: Array.from(this._allTopics),
             start: next,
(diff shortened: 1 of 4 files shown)


packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
```diff
@@ -256,26 +445,33 @@ export class IterablePlayer implements Player {
...
       this._end,
     );
 
+    log.debug(`Playing from ${toString(this._start)} to ${toString(stopTime)}`);
+
+    // This iterator is setup to only read the start messages. For playback another iterator is used.
+    const iterator = this._iterableSource.messageIterator({
+      topics: Array.from(allTopics),
+      start: this._start,
+      end: stopTime,
```
jtbandes, line 466: The `compare(iterResult.msgEvent.receiveTime, stopTime) > 0` inside the loop below can probably be removed now, right?

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
```diff
@@ -256,26 +445,33 @@ export class IterablePlayer implements Player {
...
+
     this._lastMessage = undefined;
     this._messages = [];
 
-    log.debug(`Playing from ${toString(this._start)} to ${toString(stopTime)}`);
-
     const messageEvents: MessageEvent<unknown>[] = [];
     for (;;) {
-      const result = await this._forwardIterator.next();
+      const result = await iterator.next();
```
jtbandes, line 474: Since the iterator is now a local variable, can you just do `for (const iterResult of iterator)` ? The return/break below will cause the iterator `return()` to be called, but that seems ok now (in fact it is desirable, and the current implementation is missing it.)

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
```diff
@@ -542,9 +672,56 @@ export class IterablePlayer implements Player {
...
+          const next = add(lastMessageTime, { sec: 0, nsec: 1 });
+
+          // Next would be past our desired range so we stop reading from the iterators
+          if (compare(next, this._end) > 0) {
+            break;
+          }
+
+          log.debug("Initializing forward iterator from", next);
+
+          const iteratorEnd = add(next, fromNanoSec(BigInt(this._iteratorDurationNanos)));
```
jtbandes, line 715: Clamp to `this._end`?

jtbandes (approved):

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
defunctzombie: I removed the _lastMessage part but kept the check as a defensive measure for now. I also added end support to the BagIterablesource which didn't have it.

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
defunctzombie: I added a return.

Typescript won't let me use a for loop because `iterator` doesn't have a `[Symbol iterator]` /shrug

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
defunctzombie: clamped

esthersweon:
@defunctzombie - Will not include in changelog, as this is behind feature flag.

Issue #3172, opened 2022-04-11
Track message counts and frequencies for live connections


AndrewNolte:
**Description**
Message count and frequency values show "---" when using foxglove websocket

![image](https://user-images.githubusercontent.com/24322714/162806836-b460c4cc-f5dc-4361-a88e-d490ccc2cc27.png)

- Version: Latest webapp
- OS: Ubuntu 20.04

**Steps To Reproduce**
Connect to a foxglove websocket connection
Open a Data Source Info Panel

**Expected Behavior**

**Actual Behavior**

defunctzombie:
Also worth noting that these fields won't be available for topics which do not have active subscribers from other panels since studio does not subscribe to all topics.

Today this is by design for live connections.

Changing this over to a feature request for seeing message counts and frequencies for live connections.

AndrewNolte:
Actually it makes sense that it shouldn't work for all topics on live connections. I just noticed that it also isn't working on data platform files either.

PR #3175, opened 2022-04-12
Show message counts and frequency for live data sources

jhurliman:
**User-Facing Changes**

- Message counts and frequencies are shown in the Data Source sidebar
- Live data sources now show message counts and message frequency for subscribed topics

**Description**

This splits `numMessages` out of the `Topic` type and into a `TopicStats` type. `activeData.topicStats` is a new required field, a `Map<string, TopicStats>` for per-topic statistics (currently only numMessages and optional first/last message timestamps).

The Data Source sidebar has new UI to show message counts and message frequencies per topic.

All live data sources (except the Data Platform source, which doesn't support message counts yet) have been updated to publish message counts. This only works for subscribed topics, unsubscribed topics will continue to be blank.

Screen Shot 2022-04-20 at 4 25 27 PM


packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
```diff
@@ -325,7 +361,8 @@ export default class FoxgloveWebSocketPlayer implements Player {
         isPlaying: true,
         speed: 1,
         lastSeekTime: this._lastSeekTime,
-        topics: _topics,
+        // Always copy the topic array since message counts and timestamps are being updated
+        topics: _topics.slice(0),
```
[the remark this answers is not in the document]
jhurliman: Only a shallow copy to appease React. The players themselves mutate entries in `_topics`, downstream consumers of `activeData.topics` should not be mutating the data.

jhurliman:
@jtbandes `numMessages` has always been part of the `Topic` type, but refactoring it seems reasonable.

jtbandes:
Yep, it has been a part of the type, but it's never changed rapidly (as far as I know)

defunctzombie:
I also share the concern here with changing the topics array from the players. I know we use the stability of that array for things like re-computing message path input paths and re-deriving values from those (as one example).

I don't view this array as the place for this information. If we want to keep this information as something the players provide (I'm actually not convinced on that vs having a panel compute this) but assuming we do we should make a new entry in the player state (probably active data) with this info so only the panels that need to display this do so.

defunctzombie:
I also think it could be reasonable to say that we don't support such things for live connections as a response to #3172 and turn it into a feature request rather than a bug.

defunctzombie:
@jhurliman I changed the referenced issue to a feature request and explained some reasoning to the user. I think this functionality is better implemented in a `message timeline` panel.

jtbandes:
Not sure I agree that this belongs in a timeline panel. A timeline would be a useful view but I actually think the timeline and dynamically-calculated-rate features are almost mutually exclusive. The former mostly works with pre-recorded data and the latter mostly works with live data sources.

For a live data source, if I wanted to find out the current publish rate or data rate of a message, it does feel like I’d want to navigate first to the list of topics where I would find detailed info about each topic, such as its schema and publish rate.

2metres:
@jhurliman I've added the text to the sidebar but could use some help cleaning up some Type issues and undefined cases

defunctzombie:
We cannot merge this as-is by modifying the `topics`. As mentioned before we have lots of code paths that depend on `topics` being relatively "stable". This data does not belong in the `topics` array. If we want this data to come through the player it needs to be a separate entry in the active data.

packages/studio-base/src/components/DataSourceSidebar/TopicList.tsx
```diff
@@ -20,23 +20,26 @@ import {
...
+
+const topicToFzfResult = (item: TopicWithStats) =>
   ({
     item,
     score: 0,
     positions: new Set<number>(),
     start: 0,
     end: 0,
-  } as FzfResultItem<Topic>);
+  } as FzfResultItem<TopicWithStats>);
```
foxymiles, line 42: I think you can avoid a cast here:

```typescript
const topicToFzfResult = (item: TopicWithStats): FzfResultItem<TopicWithStats> => ({
  item,
  score: 0,
  positions: new Set<number>(),
  start: 0,
  end: 0,
});
```

foxymiles (approved):

packages/studio-base/src/players/IterablePlayer/DataPlatformIterableSource.ts
```diff
@@ -107,6 +112,9 @@ export class DataPlatformIterableSource implements IIterableSource {
     }
 
     const topics: Topic[] = [];
+    // TODO(jhurliman): Fill numMessages into topicStats per topic. Bonus points if we can get
```
defunctzombie, line 115: Please don't leave TODO in the code - make an issue and remove the TODO

PR #3184, opened 2022-04-13
Fix chart auto-y scaling

### LINKED ISSUE #3168: Plot y-axis is not scaling properly
The *y* axis of a Plot panel is not scaling to fit the data. Perhaps this is related to plotting an array of values?

- Version: 1.6.0
- OS: macOS

The Plot panel is plotting an array of values from a message, which is displayed in the upper left. You can see that almost all of the values are greater than 1, yet the *y*-axis scale only goes up to 0.98.

image

The scale *does* adjust with each incoming message but I can't figure out the logic to the values it picks.

foxymiles:
**User-Facing Changes**
This fixes an issue with plot auto-scaling on the Y axis.

**Description**
The issue is that the downsampling logic was also discarding points outside the current Y axis scale but this was preventing the chart from autoscaling as new data arrives.

Fix #3168

### DIFF
--- a/packages/studio-base/src/components/TimeBasedChart/downsample.test.ts
+++ b/packages/studio-base/src/components/TimeBasedChart/downsample.test.ts
@@ -94,7 +94,10 @@ describe("downsampleScatter", () => {
       bounds,
     );
     expect(result).toEqual({
-      data: [],
+      data: [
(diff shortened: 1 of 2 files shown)


defunctzombie (approved):
📿

Issue #3181, opened 2022-04-12
Panel context menu disappears


defunctzombie:
I've got a connection open to rosbridge and am running a node that logs at 1hz. When I have a layout that has a log panel and try to use the panel context menu on another panel the context menu closes whenever a new log occurs.

If I remove the log panel the issue goes away. I've also confirmed that the issue happens when the log panel "scrolls" to show new logs that appear at the bottom and scroll old logs off the screen at the top.

https://user-images.githubusercontent.com/84792/163062778-6f7fb286-8f40-4ed0-b804-c40e06da0c95.mov

PR #3189, opened 2022-04-14
Use column-reverse flex direction for log panel messages

defunctzombie:
**User-Facing Changes**
Fixes a bug where the panel context menu would close when the log panel received a new message and scrolled to display the message.

**Description**
Setting scrollTop to the scrollHeight causes a focus loss for open context menus. This change fixes the behavior by using column-reverse flex direction for the log message scroll area. column-reverse tells the browser to treat the bottom of the div as the scroll anchor. As new messages are appended to the div, the old messages scroll off the top.

Fixes: #3181


foxymiles (approved):
That's a clever solution. 👍

jtbandes (approved):

PR #3202, opened 2022-04-18
Persist desktop app state across refreshes

### LINKED ISSUE #3044: Desktop app doesn't retain state between refreshes
**Description**
Refreshing the desktop app with ⌘+R doesn't automatically restore the previous connection but requires the user to reselect it from the recents menu.

- Version: 1.4.0
- OS: macOS

**Steps To Reproduce**

1. Open a connection
2. Hit ⌘+R to refresh

**Current Experience**
You will have to reselect the connection from the recents menu.

**Desired Experience**
It would be nice if studio remembered the previous connection and tried to restore it automatically.

foxymiles:
**User-Facing Changes**
This improves state handling over refreshes for the desktop app.

**Description**
The approach here is to check on startup if we have parseable state in `window.location` and prioritize that over command line deep links since we will be updating the values in `window.location` dynamically in response to changes in data source, layout etc and thus this state should better represent the user's intentions for the session.

Fixes #3044

### DIFF
--- a/packages/studio-base/src/util/appURLState.ts
+++ b/packages/studio-base/src/util/appURLState.ts
@@ -6,7 +6,6 @@ import { isEmpty, omitBy } from "lodash";
 
 import { fromRFC3339String, toRFC3339String, Time } from "@foxglove/rostime";
 import { LayoutID } from "@foxglove/studio-base/index";
-import isDesktopApp from "@foxglove/studio-base/util/isDesktopApp";
 
 export type AppURLState = {
   ds?: string;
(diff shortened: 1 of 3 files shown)


jtbandes (commented):
I think this looks ok, but I'm not very familiar with the desktopBridge.getDeepLinks flow.

packages/studio-base/src/util/appURLState.ts
```diff
@@ -56,10 +55,6 @@ export function encodeAppURLState(url: URL, urlState: AppURLState): URL {
  * @throws Error if URL parsing fails.
  */
 export function parseAppURLState(url: URL): AppURLState | undefined {
-  if (isDesktopApp() && url.protocol !== "foxglove:") {
```
defunctzombie, line 59: Hows this change related?

defunctzombie (approved):
Makes sense

packages/studio-base/src/util/appURLState.ts
foxymiles: We want this function to work on any initial argv urls and also any localhost urls we get on refresh.

PR #3204, opened 2022-04-18
Fix stale messages in raw messages panel

### LINKED ISSUE #3151: Raw Messages shows old message and new topic when topic is changed on live connection
Steps to reproduce:
- ROS 1 live connection
- Select a topic in the Raw Messages panel, which will show some messages
- Type to select another topic, which is an available topic but does not have any messages published

=> The panel shows the old message, but with the new topic & datatype name.

Screenshot provided by @drldcsta:

image (1)

foxymiles:
**User-Facing Changes**
This fixes an issue with stale messages being shown in the raw messages panel.

**Description**
The existing logic holds on to previous messages in the message reducer if it can't find any new messages but it doesn't check to see if those older messages correspond to the current topic. This adds some extra logic to store and compare paths and only use cached messages if the topic is the same.

Fixes #3151

### DIFF
--- a/packages/studio-base/src/components/MessagePathSyntax/useMessageDataItem.test.tsx
+++ b/packages/studio-base/src/components/MessagePathSyntax/useMessageDataItem.test.tsx
@@ -191,6 +191,45 @@ describe("useMessageDataItem", () => {
     ]);
   });
 
+  it("clears previous messages on topic change", async () => {
+    const { result, rerender } = renderHook(
+      ({ path }) => useMessageDataItem(path, { historySize: 2 }),
+      {
+        initialProps: { path: "/topic.value" },
+        wrapper({ children }) {
+          return (
+            <MockCurrentLayoutProvider>
+              <MockMessagePipelineProvider
+                messages={fixtureMessages}
+                topics={topics}
+                datatypes={datatypes}
+              >
+                {children}
+              </MockMessagePipelineProvider>
+            </MockCurrentLayoutProvider>
+          );
+        },
+      },
+    );
+    expect(result.all).toEqual([
+      [],
+      [
+        {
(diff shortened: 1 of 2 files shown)


defunctzombie (approved):
I think we have tests for this file - if we do please try adding a test for this.

jtbandes:
Just looking at the diff, this change looks like a no-op because nothing is reading this variable. Is it being implicitly read somewhere else e.g. via `isEqual`?

jtbandes:
This fix is not what I expected, based on when I filed the issue. However after some more careful reading I think it is sufficient.

When I originally filed #3151 I found some relevant code that looked suspicious to me:

https://github.com/foxglove/studio/blob/ccec391d78bfa1132760779a6e9a67382d8b6076/packages/studio-base/src/panels/RawMessages/index.tsx#L393-L400

The reason this code looks problematic is that the datatype name comes from `topic`, and the `topic` variable doesn't have hysteresis that matches the 'latching' behavior of `useMessageDataItem` to keep old messages around when new ones do not match the filter. However, if I'm understanding your change correctly, you have updated `useMessageDataItem` so that whenever the input topic changes, its return value will also change, so it looks like it _is_ now safe for the calling code to assume that `topic` matches the displayed message. Let me know if I missed anything in that interpretation.

I agree we need to update the unit tests to cover this new logic.

foxymiles:
@jtbandes Yeah essentially the issue with the existing code is that when it tries to restore previously cached messages when the topic changes, if it can't find any old items that match the new path it just returns the old items anyway.

jtbandes (approved):
Ok, and you confirmed that this new test fails without the proposed fix?

foxymiles:
Yeah if you comment out `|| path !== prevValue.path` the test fails.

PR #3211, opened 2022-04-19
Wait until player is ready to seek to url time

### LINKED ISSUE #3178: Opening a url for data platform does not take me to my requested timestamp
When I open a data platform url and play the data I see the URL updates the `time` argument. However, when I refresh or share this url the datasource does not open to the requested seek time.

https://studio.foxglove.dev/?ds=foxglove-data-platform&ds.deviceId=dev_Wm1gvryKJmREqnVT&ds.end=2022-04-12T20%3A12%3A49.126000000Z&ds.start=2022-04-12T20%3A07%3A49.272000000Z&layoutId=ea800f2c-f854-49de-8ef1-a2f7e1f31719&time=2022-04-12T20%3A08%3A31.653456688Z

foxymiles:
**User-Facing Changes**
This fixes seeking and url times for the experimental data platform player.

**Description**
Currently we try to seek to the time specified in the url before the new data platform player is ready and the seek is just discarded. The new logic here waits until the player is ready before seeking. This also requires changing the new player to not report presence of `PRESENT` until it enters the `start-play` state.

In general the order of operations on starting up a new instance of studio is complex. Maybe player presence is too coarse a tool to handle this correctly?

Fixes #3178

### DIFF
--- a/packages/studio-base/src/hooks/useInitialDeepLinkState.ts
+++ b/packages/studio-base/src/hooks/useInitialDeepLinkState.ts
@@ -14,9 +14,11 @@ import { useCurrentLayoutActions } from "@foxglove/studio-base/context/CurrentLa
 import { usePlayerSelection } from "@foxglove/studio-base/context/PlayerSelectionContext";
 import useDeepMemo from "@foxglove/studio-base/hooks/useDeepMemo";
 import { useSessionStorageValue } from "@foxglove/studio-base/hooks/useSessionStorageValue";
+import { PlayerPresence } from "@foxglove/studio-base/players/types";
 import { AppURLState, parseAppURLState } from "@foxglove/studio-base/util/appURLState";
 import isDesktopApp from "@foxglove/studio-base/util/isDesktopApp";
 
+const selectPlayerPresence = (ctx: MessagePipelineContext) => ctx.playerState.presence;
 const selectSeek = (ctx: MessagePipelineContext) => ctx.seekPlayback;
 const selectUrlState = (ctx: MessagePipelineContext) => ctx.playerState.urlState;
 
@@ -33,6 +35,7 @@ export function useInitialDeepLinkState(deepLinks: string[]): void {
     AppSetting.LAUNCH_PREFERENCE,
   );
   const seekPlayback = useMessagePipeline(selectSeek);
+  const playerPresence = useMessagePipeline(selectPlayerPresence);
 
   const appUrlRef = useRef<AppURLState | undefined>();
   if (!appUrlRef.current) {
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/hooks/useInitialDeepLinkState.ts
```diff
@@ -85,6 +88,11 @@ export function useInitialDeepLinkState(deepLinks: string[]): void {
   }, [selectSource, setSelectedLayoutId]);
 
   useEffect(() => {
+    // Wait until player is ready before we try to seek.
+    if (playerPresence !== PlayerPresence.PRESENT) {
+      return;
+    }
```
jtbandes, line 94: Is there anything to prevent the below seek from running a second time when the player changes?

foxymiles: Yeah you can see later in the effect that once it seeks it clears the url time:

```
    seekPlayback(urlState.time);
    urlState.time = undefined;
```

jtbandes:
Would it be cleaner to somehow pass the initial state to the player on construction, rather than trying to wait and do a seek after it's initialized?

foxymiles:
@jtbandes Yeah I think so. Currently it's hard to handle start up state correctly since it's spread over multiple sources of truth and the player presence/state is a bit underspecified.

foxymiles:
I guess we could expand on this to also pass a `time` when the source is selected but that would be a more extensive change:

https://github.com/foxglove/studio/blob/3db35b228fe7d962a196aaedce2c8c17b2082446/packages/studio-base/src/context/PlayerSelectionContext.ts#L80

jtbandes:
There actually used to be some code for this: https://github.com/cruise-automation/webviz/search?q=seekToTime

Looks like some of it is still around: https://github.com/foxglove/studio/search?q=seekToTime

foxymiles:
I think that approach makes sense. Should we apply a quick fix now and then take some time to apply this as a more universal solution or just take the extra time now?

jtbandes (approved):
I suppose the quick fix for now doesn't hurt.

defunctzombie:
One annoyance with delaying the PRESENT state for the player is this makes the open dialog look feel unresponsive once you've clicked "open" because we have logic that checks for player state rather than player instance.

👍 To having _time_ be something we pass to player selection/construction. That will make the initial seek more robust. It also lets the player avoid loading data at the start only to immediately seek elsewhere.

esthersweon:
@foxymiles - will not include in changelog, as this is behind feature flag.

PR #3212, opened 2022-04-19
Fix url state for ros1 & ros2 players

foxymiles:
**User-Facing Changes**
This fixes url state encoding for ros1 & ros2 players.

**Description**
Both the ros1 & ros2 players were not emitting `urlState` and consequently weren't working for url state encoding. We didn't see this before because we weren't trying to encode url state for the desktop app,


jtbandes (approved):
Is there anything we could've done to catch this error at build time? Why is urlState an optional field?

foxymiles:
When I introduced it I made it optional so we didn't have to go change all the players at once but maybe now we can change it to be required?

PR #3214, opened 2022-04-19
add sentry tracing integration

jtbandes:
**User-Facing Changes**
None

**Description**
Add sentry BrowserTracing automatic integration: https://docs.sentry.io/platforms/javascript/performance/instrumentation/automatic-instrumentation/

### DIFF
--- a/web/src/index.tsx
+++ b/web/src/index.tsx
@@ -2,7 +2,8 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
-import { init as initSentry } from "@sentry/browser";
+import * as Sentry from "@sentry/browser";
+import { BrowserTracing } from "@sentry/tracing";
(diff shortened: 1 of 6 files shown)


web/src/index.tsx
```diff
@@ -2,7 +2,8 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
-import { init as initSentry } from "@sentry/browser";
+import * as Sentry from "@sentry/browser";
```
defunctzombie, line 5: No benefit in tree shaking only initSentry? Likely we use the whole thing?

defunctzombie (approved):

web/src/index.tsx
jtbandes: Seems to be a lot of FUD out there around tree shaking. https://stackoverflow.com/a/64569600/23649

PR #3219, opened 2022-04-20
feat(Three.js panel): Fix arrow markers rendering

dudv:
**User-Facing Changes**
- Fixed rendering of arrow markers
- Fixed crash when encountering a marker without points attribute

**Description**
- Fixed rendering of arrow markers
- Added Storybook story for three.js panel
- Changed FPS counter implementation to the standard three.js module

Screenshot 2022-04-20 at 16 16 33
Left: three.js panel, right: regl panel


dudv:
@jhurliman Could you please review it?

packages/studio-base/src/panels/ThreeDeeRender/ros.ts
```diff
@@ -142,9 +142,11 @@ addRosDataType(TF_DATATYPES, "tf2_msgs/TFMessage");
 
 export const MARKER_DATATYPES = new Set<string>();
 addRosDataType(MARKER_DATATYPES, "visualization_msgs/Marker");
+addRosDataType(MARKER_DATATYPES, "studio_msgs/Marker");
```
jtbandes, line 145: What are these for? The app doesn't currently support a datatype called `studio_msgs/Marker`.

dudv: We publish all our markers under `studio_msgs` namespace. I've just realized it's a non-standard one. Probably we need some API to make it configurable, any suggestions?

dudv: Done

packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
```diff
@@ -33,6 +33,8 @@ import {
 const SHOW_STATS = true;
 const SHOW_DEBUG = false;
 
+const statsInstance = Stats();
```
jhurliman, line 36: We should avoid shared globals for things that are tied to a specific instantiation of `<ThreeDeeRender>`, `Renderer`, etc. I'm not sure how this would behave with multiple (new) 3D panels.

packages/studio-base/src/panels/ThreeDeeRender/renderables/TopicMarkers.ts
```diff
@@ -131,6 +131,11 @@ export class TopicMarkers extends THREE.Object3D {
   }
 
   private _createMarkerRenderable(marker: Marker): RenderableMarker | undefined {
+    // make sure points array is always defined to simplify logic
+    // eslint-disable-next-line
+    if (!marker.points) {
+      marker.points = [];
+    }
```
jhurliman, line 138: We ran into lots of problems in the previous 3D panel by mutating incoming user data, so I think in this iteration we want to set a rule of never modifying `Marker` objects or any decoded messages. If this is coming in as undefined but we don't catch it, the type should be changed to `| undefined` so it can be handled wherever we deal with `.points`.

packages/studio-base/src/panels/ThreeDeeRender/Stats.tsx
```diff
@@ -1,249 +0,0 @@
-// This Source Code Form is subject to the terms of the Mozilla Public
```
jhurliman, line 1: This fork of the upstream stats UI adds a few extra debug panels, I'd prefer to keep it for now.

packages/studio-base/src/panels/ThreeDeeRender/renderables/TopicMarkers.ts
dudv: What if we make a copy of the marker object in `addMarkerMessage` instead of mutating it and pass that updated copy down? Changing type to make `points` optional complicates logic across many modules.

packages/studio-base/src/panels/ThreeDeeRender/Stats.tsx
dudv: I've reverted most of the changes to stats panel so that it both contains your custom panels and reuses the three.js implementation where possible.

packages/studio-base/src/panels/ThreeDeeRender/renderables/TopicMarkers.ts
dudv: I've updated the implementation in [dfa27e0](https://github.com/foxglove/studio/pull/3219/commits/dfa27e03924ae722311d45c2f1c88e7781ccd4bc), does it look better now?

jhurliman:
https://github.com/foxglove/studio/pull/3231

jhurliman:
Closing in favor of the refactored version at https://github.com/foxglove/studio/pull/3231

Issue #3148, opened 2022-04-06
Some node names break node playground typechecking


jtbandes:
<img width="379" alt="image" src="https://user-images.githubusercontent.com/14237/161878160-28719182-4ee6-4ac7-9712-0de5f3728679.png">

<img width="496" alt="image" src="https://user-images.githubusercontent.com/14237/161878174-813b57c8-6ee0-4c0e-8810-bbbc5404e2d4.png">

When a node name does not start with `/studio_node`, typechecking is broken in the editor. The node still works at runtime though.

defunctzombie:
This happens because the node name is also used as the "filepath" in the "virtual" project structure. Files like `types.ts` are hard-coded to have a file name like `/studio_node/types.ts`. When you change the node name, this becomes its new full path and `./types` is no longer resolved because it is no longer at the same "path".

jtbandes:
Yes, I think we need to stop using the node name as part of the filepath. I'm not sure if there was any valid reason to do that but it seems unnecessary.

PR #3223, opened 2022-04-20
Use the basename from node playground scripts in the editor

defunctzombie:
**User-Facing Changes**
User can name their node playground nodes whatever they want; they no longer need to keep the `/studio_node/` prefix in their name.

**Description**
The editor uses the script filePath to resolve additional imports. When a user renames their node and removes the /studio_node/ prefix, it causes the relative path of the file to change.

To allow users to name their node whatever they want, we use only the basename of the path and manually prefix the /studio_node/ path since this is an internal implementation detail.

This change also removes the /studio_node/ prefix from new node names since it is no longer a requirement to have this prefix.

Fixes: #3148


defunctzombie:
Practically I think it could be useful to cleanup more things around the `/studio_node/` fake path and how names are handled. There are also some gotchas with naming your name the name of an existing utility file but thats for another day.

This change stop breaking the editor when the user edits their node name and removes the prefix. Thats a win.

jtbandes (approved):

Issue #3203, opened 2022-04-18
Invalid WebSocket URL pops up toast behind the open dialog


jtbandes:
When entering an invalid url such as `test` or `test:` in the websocket url field, the error pops up behind the open dialog in a toast notification. It would be better to display any error inline in the dialog instead.

<img width="866" alt="Screen Shot 2022-04-18 at 11 31 00 AM" src="https://user-images.githubusercontent.com/14237/163840261-c3b8f063-7a72-47a3-9e1b-dc83d03195a8.png">

defunctzombie:
Picking up this issue as it has definitely annoyed me several times and I'd say is pretty terrible ux for what should otherwise be a simple text input error.

jtbandes:
Additionally, the user should not be required to enter `ws://`, we should assume `ws://` if no scheme is specified.

PR #3224, opened 2022-04-20
Add a validate property to IDataSourceFactory formConfig fields

defunctzombie:
**User-Facing Changes**

Rather than receive errors for an invalid url format after clicking "open" in the connection dialog, the user is shown an error under the text field and prevented from clicking "open" until they resolve the issue.

Screen Shot 2022-04-20 at 1 49 27 PM
Screen Shot 2022-04-20 at 1 49 17 PM
Screen Shot 2022-04-20 at 1 49 08 PM

**Description**
A data source factory can provide a validate function for each field. The validate function allows the factory to intercept new values and return an error if the value for the field is not valid. The open dialog uses this to determine whether to show an error for the field or allow the user to click the "open" button.

Fixes: #3203

----

Not sold on my approach with `setFieldErrors` and `setFieldValues` but maybe this is an ok solution for now since it is internal. The `validate` property for formConfig fields is arbitrary but served this specific goal.

### DIFF
--- a/packages/studio-base/src/components/OpenDialog/Connection.tsx
+++ b/packages/studio-base/src/components/OpenDialog/Connection.tsx
@@ -2,7 +2,7 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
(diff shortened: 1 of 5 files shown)


foxymiles (approved):
Looks good. Better UX I agree.

Issue #3100, opened 2022-03-29
Layout storage stops working when LocalStorage quota is exceeded


2metres:
**Description**
<img width="941" alt="Screen Shot 2022-03-30 at 9 31 39 am" src="https://user-images.githubusercontent.com/924528/160717637-3f6dfc55-0863-4820-a925-cae6adb643d1.png">

- Version: Dev
- OS: MacOS

**Steps To Reproduce**

Have a large number of layouts and load the example layout

**Expected Behavior**

**Actual Behavior**

jtbandes:
I think we might want to just switch to IndexedDB for the layout storage. While having lots of layouts is not ideal we shouldn't be imposing arbitrary limits on users adding lots of layouts if they really want to.

jtbandes:
> Why's it not ideal?

Because scrolling through a huge list is a bad UX. Maybe this is saying more about our layout UX than about the number of layouts.

https://web.dev/storage-for-the-web/#other

foxymiles:
Sounds like the strategy here is to switch to IndexedDB and do a one-time migration from localStorage?

defunctzombie:
> Sounds like the strategy here is to switch to IndexedDB and do a one-time migration from localStorage?

Yep.

PR #3225, opened 2022-04-21
Replace LocalStorageLayoutStorage with IdbLayoutStorage

jtbandes:
**User-Facing Changes**
Fixed an issue where creating too many layouts would cause the web app to run out of storage space.

**Description**
Fixes https://github.com/foxglove/studio/issues/3100

Migrate layout cache from localStorage to IndexedDB for web build.

Layouts are stored in one database with the primary key being the `[namespace, id]` tuple:
image


web/src/services/IdbLayoutStorage.ts
```diff
@@ -0,0 +1,134 @@
...
+    // no un-namespaced layouts to migrate.
+    void namespace;
+  }
+
+  /**
+   * Prior implementation (LocalStorageLayoutStorage) stored layouts in localStorage under a key
+   * prefix. This approach was abandoned due to small capacity constraints on localStorage.
+   * https://github.com/foxglove/studio/issues/3100
+   */
+  private async _migrateFromLocalStorage() {
```
foxymiles, line 104: I suppose this is fast enough that we don't need any UI state to indicate that it's in progress?

jtbandes: I sincerely hope so... :(

I do see that some folks call out indexeddb being slow particularly if you use separate transactions: https://rxdb.info/slow-indexeddb.html But I anticipate the number of layouts our users have will be much less than what they show in the charts...hopefully...

web/src/services/IdbLayoutStorage.ts
```diff
@@ -0,0 +1,134 @@
...
+      const store = db.createObjectStore(OBJECT_STORE_NAME, {
+        keyPath: ["namespace", "layout.id"],
+      });
+      store.createIndex("namespace", "namespace");
+    },
+  });
+
+  async list(namespace: string): Promise<readonly Layout[]> {
+    const results: Layout[] = [];
+    for (const record of await (
```
defunctzombie, line 44: I find the formatting here hard to read. Maybe make this a variable on the line above the for loop.

web/src/services/IdbLayoutStorage.ts
```diff
@@ -0,0 +1,134 @@
...
+      await this._db
+    ).getAllFromIndex(OBJECT_STORE_NAME, "namespace", namespace)) {
+      try {
+        results.push(migrateLayout(record.layout));
+      } catch (err) {
+        log.error(err);
```
defunctzombie, line 50: One thing that I have found somewhat lacking in our layout flow is the loss of errors to console. It might be useful in some future to attach errors to layouts so the user can see these. I don't think its a great experience for users to open the devtools to learn about what the app is doing if it is relevant to their workflow.

defunctzombie (approved):
I've forgotten what "namespaces" are for layouts but this seems 👍 . Its fully internal so we can iterate on it without impacting users.

web/src/services/IdbLayoutStorage.ts
jtbandes: Having something like "problems" that works at the whole app level might be nice. We can even call it sendNotification.

jtbandes:
Namespaces are how we separate layout storage for local only layouts vs logged-in users.

PR #3228, opened 2022-04-21
feat(Three.js panel): Custom background color settings

dudv:
**User-Facing Changes**
Custom background color settings option for three.js 3D panel

Screenshot 2022-04-20 at 14 59 47

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
@@ -34,6 +34,10 @@ export type RendererEvents = {
   removeLabel: (labelId: string, renderer: Renderer) => void;
 };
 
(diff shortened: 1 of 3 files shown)


dudv:
@jhurliman Could you please take a look?

dudv:
The new API seems much more verbose, especially calling private methods like `__updatePanelSettingsTree`.  Is it the right way of using it? @jhurliman

foxymiles:
@dudv Are you referring to this with your comment about casting?

```    (context as unknown as any).__updatePanelSettingsTree({
```

Hiding this as `__updatePanelSettingsTree` is just a temporary measure so we can use this in our extension panels and work out the details of the API before we commit to making this a public API we encourage people to use. Once we're happy with the details we'll make this a proper method on the extension panel context.

jhurliman:
This is implemented

PR #3232, opened 2022-04-21
Convert the image panel to use the new settings editor

foxymiles:
**User-Facing Changes**
This converts the image panel to the new settings editor.

**Description**
This converts the existing image panel properties to use the new settings editor. 

Screen Shot 2022-04-21 at 2 07 26 PM

It also improves the resize behavior of the settings editor to work better at smaller widths.

### Before:
Screen Shot 2022-04-21 at 2 06 52 PM

### After:
Screen Shot 2022-04-21 at 2 07 07 PM

### DIFF
--- a/packages/studio-base/src/components/PanelSettings/index.tsx
+++ b/packages/studio-base/src/components/PanelSettings/index.tsx
@@ -17,7 +17,6 @@ import {
   useCurrentLayoutSelector,
   useSelectedPanels,
 } from "@foxglove/studio-base/context/CurrentLayoutContext";
-import { useHelpInfo } from "@foxglove/studio-base/context/HelpInfoContext";
 import { usePanelCatalog } from "@foxglove/studio-base/context/PanelCatalogContext";
 import {
   ImmutableSettingsTree,
@@ -40,7 +39,7 @@ export default function PanelSettings({
   const selectedLayoutId = useCurrentLayoutSelector(selectedLayoutIdSelector);
   const { selectedPanelIds: originalSelectedPanelIds, setSelectedPanelIds } = useSelectedPanels();
   const selectedPanelIds = selectedPanelIdsForTests ?? originalSelectedPanelIds;
-  const { openLayoutBrowser, openHelp } = useWorkspace();
+  const { openLayoutBrowser } = useWorkspace();
   const selectedPanelId = useMemo(
     () => (selectedPanelIds.length === 1 ? selectedPanelIds[0] : undefined),
     [selectedPanelIds],
@@ -62,7 +61,6 @@ export default function PanelSettings({
     () => (panelType != undefined ? panelCatalog.getPanelByType(panelType) : undefined),
     [panelCatalog, panelType],
   );
-  const { setHelpInfo } = useHelpInfo();
 
   const [showShareModal, setShowShareModal] = useState(false);
   const shareModal = useMemo(() => {
@@ -157,22 +155,6 @@ export default function PanelSettings({
     <SidebarContent disablePadding={isSettingsTree} title={`${panelInfo.title} panel settings`}>
       {shareModal}
       <Stack gap={2} justifyContent="flex-start">
-        {panelInfo.help != undefined && (
-          <Stack paddingX={isSettingsTree ? 2 : 0}>
-            <Typography color="text.secondary">
-              See docs{" "}
-              <Link
-                onClick={() => {
-                  setHelpInfo({ title: panelInfo.type, content: panelInfo.help });
-                  openHelp();
-                }}
-              >
-                here
-              </Link>
-              .
-            </Typography>
-          </Stack>
-        )}
         <div>
           {settingsTree && <SettingsEditor settings={settingsTree} />}
           {!settingsTree && schema && (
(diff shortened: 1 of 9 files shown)


packages/studio-base/src/components/PanelSettings/index.tsx
```diff
@@ -157,22 +155,6 @@ export default function PanelSettings({
...
-                  setHelpInfo({ title: panelInfo.type, content: panelInfo.help });
-                  openHelp();
-                }}
-              >
-                here
-              </Link>
-              .
-            </Typography>
-          </Stack>
-        )}
```
jtbandes, line 175: Do we need a discussion on whether to keep or remove these? It seems unfortunate to remove a link to in-app help docs. However I do agree that the "See docs here" text is not useful. Maybe we want to rethink the design of the help button? cc @2metres @esthersweon

defunctzombie (approved):
Any migration we need to do with the previous config or are all the config keys the same?

packages/studio-base/src/components/PanelSettings/index.tsx
defunctzombie: This should be removed - I commented in some other ticket how I find it taking up more valuable space for a daily user. We can re-introduce a help icon but I generally don't like littering the app with "view docs" links like we have done in a few places.

packages/studio-base/src/panels/ImageView/index.tsx
```diff
@@ -96,6 +102,57 @@ const useStyles = makeStyles((theme: Theme) => ({
...
+      },
+      rotation: {
+        input: "select",
+        label: "Rotation",
+        value: String(config.rotation ?? 0),
+        options: [
+          { label: "0°", value: "0" },
+          { label: "90°", value: "90" },
+          { label: "180°", value: "180" },
+          { label: "270°", value: "270" },
```
jtbandes, line 137: These values used to be numbers and now they are strings. Intentional?

packages/studio-base/src/panels/ImageView/index.tsx
```diff
@@ -135,6 +194,23 @@ function ImageView(props: Props) {
...
   }, [allImageTopics, config, saveConfig]);
 
+  const actionHandler = useCallback(
+    (action: SettingsTreeAction) => {
+      const newConfig = { ...config };
+      set(newConfig, action.payload.path, action.payload.value);
+      saveConfig(newConfig);
+    },
+    [config, saveConfig],
+  );
```
jtbandes, line 204: Do we plan to make this the default behavior or do we need to copy this code into all the panels?

packages/studio-base/src/components/PanelSettings/index.tsx
jtbandes: You don't think it's useful to have any kind of way to jump from the panel to the panel's in-app docs?

packages/studio-base/src/panels/ImageView/index.tsx
foxymiles: My guess is that this is not going to be a universal enough solution to make it applicable as a default.

foxymiles: Yeah this is technically cheating although in this case I don't think it should have any bad consequences. Immer would be more correct.

packages/studio-base/src/panels/ImageView/index.tsx
foxymiles: Looks like the image panel still works this way (probably due to some JS coercion) but I'll fix this in the action handler.

jtbandes: Why the action handler and not here in the schema?

foxymiles:
> Any migration we need to do with the previous config or are all the config keys the same?

The actual config data is unchanged.

packages/studio-base/src/panels/ImageView/index.tsx
foxymiles: Currently the select is defined as string, string pairs but I guess we could make the value `string | number`.

```typescript
  | { input: "select"; value?: string; options: Array<{ label: string; value: string }> }
```

jtbandes: Ideally the `value` and `options.value` would be constrained to have the same type. One way to do this would be to have two separate union options for string & number.

foxymiles: Yeah. I've noticed typescript already struggles to type this correctly so I'm hesitant to make it a more complex type.

foxymiles: Actually we probably need to allow `number[]` and `string[]` here too for multi-selects.

jtbandes: does this work?

```ts
...
  | { input: "select"; value?: number | number[]; options: Array<{ label: string; value: number }> }
  | { input: "select"; value?: string | string[]; options: Array<{ label: string; value: string }> }
```

foxymiles: Hmm TS doesn't seem to like it:

```
ERROR in ./packages/studio-base/src/panels/ImageView/index.tsx 136:27-32
TS2322: Type 'number' is not assignable to type 'string'.
    134 |         options: [
    135 |           { label: "0°", value: 0 },
  > 136 |           { label: "90°", value: 90 },
        |                           ^^^^^
    137 |           { label: "180°", value: 180 },
    138 |           { label: "270°", value: 270 },
    139 |         ],
```

jtbandes: Even if you remove the String part from `value: String(config.rotation ?? 0),` above?

foxymiles: Oh right that did fix it. 👍

jtbandes (approved):

PR #3234, opened 2022-04-21
Convert publish panel to new settings editor

foxymiles:
**User-Facing Changes**
This converts the publish panel to the new settings editor.

**Description**
The publish panel's settings are simple so this is straightforward.

Screen Shot 2022-04-21 at 4 15 08 PM

Should we move the topic and datatype selectors into the settings as well?

Screen Shot 2022-04-21 at 4 16 23 PM

### DIFF
--- a/packages/studio-base/src/panels/Publish/index.tsx
+++ b/packages/studio-base/src/panels/Publish/index.tsx
@@ -12,7 +12,9 @@
 //   You may not use this file except in compliance with the License.
 
 import { Box, Stack, Typography } from "@mui/material";
-import { useCallback, useEffect, useMemo, useRef } from "react";
+import produce from "immer";
+import { set } from "lodash";
+import { useCallback, useContext, useEffect, useMemo, useRef } from "react";
 import styled from "styled-components";
 
 import { useRethrow } from "@foxglove/hooks";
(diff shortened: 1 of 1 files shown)


defunctzombie:
> Should we move the topic and datatype selectors into the settings as well?

I wouldn't bother for now. The main goal is to move the existing settings over rather than re-visit the UI of every panel.

packages/studio-base/src/panels/Publish/index.tsx
```diff
@@ -122,6 +143,25 @@ function Publish(props: Props) {
     prevDatatype.current = datatype;
   }, [saveConfig, datatype, datatypes]);
 
+  const actionHandler = useCallback(
+    (action: SettingsTreeAction) => {
+      saveConfig(
+        produce(props.config, (draft) => {
```
defunctzombie, line 149: What does this produce pattern do? Should all panels be doing this or can we find a way to handle the ideal case for panels? I'm concerned that if we push the burden of "doing the right thing" onto panel authors then there is more room for mistakes.

defunctzombie (approved):

packages/studio-base/src/panels/Publish/index.tsx
foxymiles: `produce` is from immer. It creates a writable "draft" copy that you can mutate directly but produces a new object that preserves as much of the state of the original object as possible. It accomplishes the same thing as the typical redux spread pattern with a lot less fuss.

We could build this into the API as the default behavior if a user doesn't supply an action handler to make the trivial cases easier. We'd probably want to document carefully what assumptions are built in if we do that.

PR #3235, opened 2022-04-21
Data platform player: don't continue trying to preload data if an error is encountered

jtbandes:
**User-Facing Changes**
Fixed an issue where errors loading data from Data Platform could lead to the app repeatedly requesting the same data.

**Description**
Put the player in a hard error state and break the playback loop that continued triggering preloading.

Fixes https://github.com/foxglove/studio/issues/3239


defunctzombie (approved):
So we can't remove the while loop because that is used to wait for more messages which might get loaded elsewhere - not in the getMessages call itself. We are gonna deprecate this player before too long but it does seem like having getMessage wait could have been nicer?

PR #3242, opened 2022-04-22
[New3D] Camera settings sidebar improvements

jhurliman:
**User-Facing Changes**

None

**Description**

Miscellaneous improvements to the camera settings sidebar


jtbandes (approved):

defunctzombie:
> The performance is not great, even with sidebar updates throttled to every 500ms.

👍 to merging this PR - but it does seem like it is worth having a discussion around this topic and how we see settings evolving to support such use cases

foxymiles:
You can also take a look at my branch which takes a slightly different approach and seems to have ok performance: https://github.com/foxglove/studio/tree/miles/new3d-settings

PR #3244, opened 2022-04-22
Convert diagnostic summary panel to new settings UI

foxymiles:
**User-Facing Changes**
This converts the diagnostic summary panel to use the new settings UI.

**Description**
This panel just has a single boolean setting so the conversion is simple.

Screen Shot 2022-04-22 at 7 19 59 AM

### DIFF
--- a/packages/studio-base/src/panels/diagnostics/DiagnosticSummary.tsx
+++ b/packages/studio-base/src/panels/diagnostics/DiagnosticSummary.tsx
@@ -23,8 +23,9 @@ import PinIcon from "@mdi/svg/svg/pin.svg";
 import { Stack, Theme, Menu, MenuItem } from "@mui/material";
 import { makeStyles } from "@mui/styles";
 import cx from "classnames";
+import produce from "immer";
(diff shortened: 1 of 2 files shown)


jtbandes (approved):

packages/studio-base/src/panels/diagnostics/DiagnosticSummary.tsx
```diff
@@ -207,6 +222,28 @@ function DiagnosticSummary(props: Props): JSX.Element {
     [pinnedIds, saveConfig],
   );
 
+  const actionHandler = useCallback(
```
jtbandes, line 225: I continue to feel like this is a lot of boilerplate for a simple behavior that should be default. I suppose after the initial  conversion is done we can come clean these up?

foxymiles: We could make the `actionHandler` optional and supply this behavior as its default in its absence. I think we'd want to carefully document what assumptions this makes if people want to use it that way.

PR #3252, opened 2022-04-25
Automatically select single panel layouts when opening the settings sidebar

jhurliman:
**User-Facing Changes**

* The settings sidebar works without interaction for single panel layouts

**Description**

If a layout only contains one panel, that panel is auto-selected when opening the settings sidebar.

### DIFF
--- a/packages/studio-base/src/components/PanelSettings/index.tsx
+++ b/packages/studio-base/src/components/PanelSettings/index.tsx
@@ -25,11 +25,16 @@ import {
 import { useWorkspace } from "@foxglove/studio-base/context/WorkspaceContext";
 import { PanelConfig } from "@foxglove/studio-base/types/panels";
 import { TAB_PANEL_TYPE } from "@foxglove/studio-base/util/globalConstants";
-import { getAllPanelIds, getPanelTypeFromId } from "@foxglove/studio-base/util/layout";
+import { getPanelTypeFromId } from "@foxglove/studio-base/util/layout";
 
 import SchemaEditor from "./SchemaEditor";
 
 const selectedLayoutIdSelector = (state: LayoutState) => state.selectedLayout?.id;
+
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/components/PanelSettings/index.tsx
```diff
@@ -37,8 +37,26 @@ export default function PanelSettings({
   selectedPanelIdsForTests?: readonly string[];
 }>): JSX.Element {
   const selectedLayoutId = useCurrentLayoutSelector(selectedLayoutIdSelector);
-  const { selectedPanelIds: originalSelectedPanelIds, setSelectedPanelIds } = useSelectedPanels();
+  const { getCurrentLayoutState } = useCurrentLayoutActions();
```
jtbandes, line 40: Rather than `useCurrentLayoutActions` + `useMemo`, just use `useCurrentLayoutSelector` with a custom selector that returns the panel id if the layout has a single panel, or undefined otherwise.

packages/studio-base/src/components/PanelSettings/index.tsx
```diff
@@ -37,8 +37,26 @@ export default function PanelSettings({
...
+  const savedProps = getCurrentLayoutState().selectedLayout?.data?.configById;
+  const allPanelIds = useMemo(
+    () => (selectedLayoutId && savedProps ? getAllPanelIds(selectedLayoutId, savedProps) : []),
```
jtbandes, line 43: The first argument to `getAllPanelIds` is supposed to be a layout object (selectedLayout.data), not a layout id. It is passed directly into the `getLeaves` function from react-mosaic.

jhurliman: > Argument of type 'PanelsState' is not assignable to parameter of type 'MosaicNode<string>'

jtbandes (approved):

PR #3253, opened 2022-04-25
Convert plot panel settings to new settings sidebar

foxymiles:
**User-Facing Changes**
This converts the plot panel settings to the new settings UI

**Description**
This is pretty straightforward since the settings are not complex.

Screen Shot 2022-04-25 at 6 56 54 AM


defunctzombie (approved):
Should we put the `X range` into a `Timeseries only` group? Maybe thats for another time when we re-visit the overall plot settings flow.

foxymiles:
@defunctzombie This is what it looks like with the follow width broken out into a timeseries only group.

Screen Shot 2022-04-26 at 12 07 17 PM

PR #3256, opened 2022-04-25
Add new toggle function to SettingsTreeNode

### LINKED ISSUE #3241: Clicking the "visibility" toggle for the Topic settings group does not hide the topics
**Description**
On the map panel, when I click the 👁️ icon to toggle visibility on the topics group the topics are still visible.

Screen Shot 2022-04-21 at 7 07 07 PM

Screen Shot 2022-04-21 at 7 07 11 PM

foxymiles:
**User-Facing Changes**
This removes the visibility toggle from panel settings where it's not useful.

**Description**
The visibility toggle we now show on all settings tree nodes doesn't make sense for all panels. It's up to each panel to decide if this is useful and what the semantics should be. So instead of showing it by default we make it opt in via the new `allowToggle` and `toggled` properties:

```typescript
export type SettingsTreeNode = {
  allowToggle?: boolean;
  children?: SettingsTreeChildren;
  fields?: SettingsTreeFields;
  label?: string;
  toggled?: boolean;
};
```

If panels want to opt-in to this feature they can set `allowToggle = true` and then handle the new `toggle` SettingsTreeAction:

```typescript
export type SettingsTreeAction =
  | { action: "toggle"; payload: { path: readonly string[] } }
...
```

Perhaps as this evolves we might want to also offer a `toggleIcon` property that allows users to choose from a predefined list of icons that can be shown in the toggle control position.

Fixes #3241

### DIFF
--- a/packages/studio-base/src/components/SettingsTreeEditor/types.ts
+++ b/packages/studio-base/src/components/SettingsTreeEditor/types.ts
@@ -38,9 +38,11 @@ export type SettingsTreeFields = Record<string, SettingsTreeField>;
 export type SettingsTreeChildren = Record<string, SettingsTreeNode>;
 
 export type SettingsTreeNode = {
+  allowToggle?: boolean;
   children?: SettingsTreeChildren;
   fields?: SettingsTreeFields;
   label?: string;
+  toggled?: boolean;
 };
 
 /**
@@ -53,13 +55,15 @@ type DistributivePick<T, K extends keyof T> = T extends unknown ? Pick<T, K> : n
  * Represents actions that can be dispatched to source of the SettingsTree to implement
  * edits and updates.
  */
-export type SettingsTreeAction = {
-  action: "update";
-  payload: { path: readonly string[] } & DistributivePick<
-    SettingsTreeFieldValue,
-    "input" | "value"
-  >;
-};
+export type SettingsTreeAction =
+  | { action: "toggle"; payload: { path: readonly string[] } }
+  | {
+      action: "update";
+      payload: { path: readonly string[] } & DistributivePick<
+        SettingsTreeFieldValue,
+        "input" | "value"
+      >;
+    };
 
 /**
  * A settings tree is a tree of panel settings that can be managed by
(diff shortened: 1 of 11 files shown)


packages/studio-base/src/components/SettingsTreeEditor/types.ts
```diff
@@ -53,13 +55,15 @@ type DistributivePick<T, K extends keyof T> = T extends unknown ? Pick<T, K> : n
...
  */
-export type SettingsTreeAction = {
-  action: "update";
-  payload: { path: readonly string[] } & DistributivePick<
-    SettingsTreeFieldValue,
-    "input" | "value"
-  >;
-};
+export type SettingsTreeAction =
+  | { action: "toggle"; payload: { path: readonly string[] } }
```
jtbandes, line 59: Why is toggle a separate kind of action and not simply an update that sets a boolean value?

foxymiles: We'd need to indicate somehow which boolean value contains the toggle state in that case right?

jtbandes: I think this is somewhat easy to change down the line so the first iteration doesn't matter that much. I am just wary of adding a feature/property called "toggled" which is a vague name and it's not clear what it should be used for (if the feature is supposed to be about visibility, maybe we should call it visibility).

Does the visible setting today get passed to the panels? In this PR it looks like the deleted code was just tracking it as a useState inside the tree itself, not hooked up to anything. In that case I don't see how adding an allowToggle field at this point is useful, if the feature doesn't work yet. Maybe I missed something though.

foxymiles: I've just turned this off for now in the 3d panel PR:

https://github.com/foxglove/studio/pull/3265

So maybe we can shelve this until we've had a chance to discuss the scope and implementation a bit more.

foxymiles: Also WRT `toggled` the idea was that visibility might not be the only thing it would make sense to toggle for a group and that it might be panel-dependent.

foxymiles:
Closed pending further discussion.

PR #3257, opened 2022-04-25
Make settings editor filter opt-in instead of opt-out

foxymiles:
**User-Facing Changes**
None

**Description**
This makes the top filter on the settings editor opt-in instead of opt-out. This makes more sense since the filter is not useful for most of our panels, which only have a few settings.

### DIFF
--- a/packages/studio-base/src/components/SettingsTreeEditor/index.stories.tsx
+++ b/packages/studio-base/src/components/SettingsTreeEditor/index.stories.tsx
@@ -287,8 +287,9 @@ export const Default = (): JSX.Element => {
 
   const settings = useMemo(
     () => ({
-      settings: settingsNode,
       actionHandler,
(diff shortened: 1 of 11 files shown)


foxymiles:
@jtbandes It would be easy enough for a panel to implement that heuristic on its own if it wanted to.

jtbandes:
But I'd rather not have several different panels implementing similar (but maybe slightly different) heuristics.

foxymiles:
@jtbandes A hardcoded # seems too rigid don't you think? It might depend on the content.

jtbandes:
I think we'll need to wait and see how it feels with our different panels.

foxymiles:
Right now it feels like the 3d panel is the only one that's likely to contain enough stuff for this to be useful but maybe also the plot panel if we move everything from the existing legend editor?

PR #3259, opened 2022-04-25
Fix ROS1 C++ subscribers connecting to a publishing Studio

jhurliman:
**User-Facing Changes**

* Fixed a bug with ROS1 C++ subscribers unable to connect to Studio publishing a topic

**Description**

ROS1 C++ XML-RPC client doesn't understand `<string/>`, we now send it as `<string></string>` so the `requestTopics` response can be parsed


defunctzombie (commented):
👍 ~~I'd word the user facing changes in terms of what studio now supports i.e. (Studio ROS1 publishing support c++ subscriber nodes) or some such variation.~~

defunctzombie (approved):

PR #3262, opened 2022-04-26
Don't emit undefined messages from ROS2 data source

jhurliman:
**User-Facing Changes**

* Fixes a crash when trying to visualize a ROS2 topic with a missing message definition

**Description**

The @foxglove/ros2 library emits message events even when messages are unparseable due to missing message definitions. Now we continue to count stats for these messages but don't emit the messages in Studio. A data source problem is already shown for these topics.

### DIFF
--- a/packages/studio-base/src/players/Ros2Player.ts
+++ b/packages/studio-base/src/players/Ros2Player.ts
@@ -482,9 +482,11 @@ export default class Ros2Player implements Player {
       this._metricsCollector.recordTimeToFirstMsgs();
     }
 
-    const msg: MessageEvent<unknown> = { topic, receiveTime, publishTime, message, sizeInBytes };
-    this._parsedMessages.push(msg);
-    this._handleInternalMessage(msg);
+    if (message != undefined) {
+      const msg: MessageEvent<unknown> = { topic, receiveTime, publishTime, message, sizeInBytes };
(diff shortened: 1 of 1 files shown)


foxymiles (approved):

packages/studio-base/src/players/Ros2Player.ts
```diff
@@ -482,9 +482,11 @@ export default class Ros2Player implements Player {
       this._metricsCollector.recordTimeToFirstMsgs();
     }
 
-    const msg: MessageEvent<unknown> = { topic, receiveTime, publishTime, message, sizeInBytes };
-    this._parsedMessages.push(msg);
-    this._handleInternalMessage(msg);
+    if (message != undefined) {
```
jtbandes, line 485: Is it desirable for the `on("message")` handler to be called with undefined? As a user of the ROS2 library I would be surprised to get undefined in this callback...

jhurliman: If you configure the library to not parse messages, that parameter will only be `undefined`. Currently, in parse mode you will either get a parsed message if the msgdef was available or `undefined` if no msgdef was available. The thinking is that missing message definitions in ROS2 is unfortunately common and there are still valid reasons to subscribe to topics even if we can't parse the messages (recording, topic stats, etc).

PR #3263, opened 2022-04-26
Improve plot panel message path interaction

### LINKED ISSUE #3146: Keyboard input sporadically disappears
**Description**
I've experienced this problem several times now (and my coworker has as well) where the characters we type in a plotting legend show up briefly, but then disappear, and then sometimes come back and sometimes not. It also is usually not all the characters I type - they are sporadic.

- Version: 1.5.1
- OS: Ubuntu 20.04

**Steps To Reproduce**

1. Open a live connection (I used turtlesim running on the same machine as foxglove in this example).
2. Create a plot panel subscribe to several topics (it seems that this bug is easier to reproduce when Foxglove is given a lot to do)
3. Type out some topic names

**Expected Behavior**
Typed words should appear on the screen as typed

**Actual Behavior**
Typed words get jumbled, with characters randomly dropped, or dropped and then re-added later in the string.

For example, I might try to type "hello/test/123/abc". Right after I type the "ab", they go away, so I'm left with "hello/test/123/c".

Or, the missing characters might come back. For example, if I type "this/is/a/test", the "th" in "this" disappears and doesn't come back, but also, "test" disappears, and _does_ come back after a second, leaving us with "is/is/a/test".

I captured a video of those two examples here, although its harder to understand if you don't see where my hands are typing.

https://user-images.githubusercontent.com/43035541/161804578-f8bd82a7-7024-46a0-a09e-2f15352a0ae6.mp4

Sometimes the characters come back later in the string, which is the worst of all. For example, "test/123/abc" might come back as "test/1/ab23c". I wasn't able to get an example of that on camera though.

foxymiles:
**User-Facing Changes**
Fixes a bug in message path inputs in the plot panel that could cause edits to be dropped or scrambled.

**Description**
Currently making any changes to one of the message paths in the plot legend, even changing a single character, forces a write to the panel config which forces a re-render of the entire plot panel and also sends a write to the layout and to storage. This creates a race condition that manifests when studio is busy and the user is rapidly editing the message path.

The solution in this PR is to create a stateful working copy of the panel config and commit edits to that working copy instead of pushing the edits all the way through to the layout. This working copy is used to render the plot. A debounced `useEffect` hook is used to flush state back to the layout so that the user's changes are saved.

Note - the performance of the plot panel accumulating messages from a live data source over time could probably be improved. We're doing a lot of work every time we receive new messages.

Fixes #3146

### DIFF
--- a/packages/studio-base/src/panels/Plot/index.tsx
+++ b/packages/studio-base/src/panels/Plot/index.tsx
@@ -17,7 +17,8 @@ import { Stack } from "@mui/material";
 import produce from "immer";
 import { compact, set, uniq } from "lodash";
 import memoizeWeak from "memoize-weak";
-import { useEffect, useCallback, useMemo, ComponentProps, useContext } from "react";
+import { useEffect, useCallback, useMemo, ComponentProps, useContext, useState } from "react";
+import { useDebouncedCallback } from "use-debounce";
 
 import { filterMap } from "@foxglove/den/collection";
 import {
@@ -171,7 +172,30 @@ function selectEndTime(ctx: MessagePipelineContext) {
 }
 
 function Plot(props: Props) {
-  const { saveConfig, config } = props;
+  const { saveConfig: persistConfig, config: savedConfig } = props;
+
+  // We make a working copy of our config for real time editing since changing the config
+  // forces the entire component to re-render and results in race conditions from
+  // the async write to layout storage.
+  const [config, setConfig] = useState(savedConfig);
+  const saveConfig = useCallback((newConfig: Partial<PlotConfig>) => {
+    setConfig((oldConfig) => ({ ...oldConfig, ...newConfig }));
+  }, []);
+
+  // Commit our working config to the layout with a debounce to prevent thrashing.
+  const debouncedSaveConfig = useDebouncedCallback(
+    (newConfig: Partial<PlotConfig>) => persistConfig(newConfig),
+    1000,
+  );
+  useEffect(() => {
+    debouncedSaveConfig(config);
+  }, [config, debouncedSaveConfig]);
+
+  // Sync our working config with the external config to support changes from settings etc.
+  useEffect(() => {
+    setConfig(savedConfig);
+  }, [savedConfig]);
+
   const {
     title,
     followingViewWidth,


jtbandes:
For user-facing changes, I think "This improves the user interaction with the message path input in the plot legend" is pretty vague. If an end user reads this in the release notes, it boils down to "something was improved" but they don't know what was improved. For this ticket in particular, I might reword it as a bugfix, like "Fixed a bug where [...problem...] would happen when [...condition...]" or "[...expected behavior...] now works correctly when [...condition...]"

jtbandes:
> …even changing a single character, forces a write to the panel config which forces a re-render of the entire plot panel…

this part may be true, but I don't see how this PR would change it – you are still triggering a render by calling setState?

> …and also sends a write to the layout and to storage

this part about the storage doesn't seem right — there is a debounce already in: https://github.com/foxglove/studio/blob/9581ab349efc97c88a08dd5adc55703b9820bd8a/packages/studio-base/src/providers/CurrentLayoutProvider/index.tsx#L154-L156

foxymiles:
> > …even changing a single character, forces a write to the panel config which forces a re-render of the entire plot panel…
> 
> this part may be true, but I don't see how this PR would change it – you are still triggering a render by calling setState?

You can try before and after this patch. You'll see tons of warnings of immediate rerenders in the behavior on main.

> > …and also sends a write to the layout and to storage
> 
> this part about the storage doesn't seem right — there is a debounce already in:
> 
> https://github.com/foxglove/studio/blob/9581ab349efc97c88a08dd5adc55703b9820bd8a/packages/studio-base/src/providers/CurrentLayoutProvider/index.tsx#L154-L156

Maybe this debounce may be responsible for the lost state at high frequency updates? Current behavior definitely drops updates.

packages/studio-base/src/panels/Plot/index.tsx
```diff
@@ -171,7 +172,30 @@ function selectEndTime(ctx: MessagePipelineContext) {
...
+  const debouncedSaveConfig = useDebouncedCallback(
+    (newConfig: Partial<PlotConfig>) => persistConfig(newConfig),
+    1000,
+  );
+  useEffect(() => {
+    debouncedSaveConfig(config);
+  }, [config, debouncedSaveConfig]);
+
+  // Sync our working config with the external config to support changes from settings etc.
+  useEffect(() => {
```
defunctzombie, line 195: Does this override any changes we might have in the working config but have not yet persisted to the saved config? Maybe we only use the savedConfig once on panel setup (like you have with useState(savedConfig) and treat it like state and wait for it to come back to the panel?

Another thought could be to make the layer that handles the `persistConfig` call do this debounce. So it immediately provides the updated config to the panel but internally queues it to persist to the underlying storage. This seems like a more robust solution since it would work across all panels rather than having each panel implement save debouncing. From what I can tell, panels treat persistConfig and savedConfig kind of like react state variables.

foxymiles: We had to implement the same perf workaround for the new 3d panel since writing changes all the way through to the panel config in realtime was too slow. Maybe there's a common solution to this problem?

jtbandes:
Fixed by #3281 #3290

foxymiles:
This is no longer necessary thanks to recent performance fixes.

PR #3271, opened 2022-04-27
Require only ds param for initial deep link state

foxymiles:
**User-Facing Changes**
This fixes an issue with deep linking to the sample data source using a link like https://studio.foxglove.dev/?ds=sample-nuscenes.

**Description**
The sample data source is a special data source that doesn't use `dsParams`. So we need to relax our requirement that both `ds` and `dsParams` are required to specify a deep link.


defunctzombie (approved):

defunctzombie:
I wonder why we had the `&& dsParams` logic. Guess we'll find out if something else breaks since we don't have any comment about it. At least now we have tests :D

PR #3275, opened 2022-04-27
feat(Three.js panel): API to add custom three.js objects

dudv:
**User-Facing Changes**
A new property for the three.js panel to inject three.js objects into the scene

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
@@ -88,6 +88,7 @@ export class Renderer extends EventEmitter<RendererEvents> {
   occupancyGrids = new OccupancyGrids(this);
   pointClouds = new PointClouds(this);
   markers = new Markers(this);
+  injectedObjectsByName = new Map<string, THREE.Object3D>();
 
   constructor(canvas: HTMLCanvasElement) {
     super();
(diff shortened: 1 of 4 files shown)


dudv:
@jhurliman Could you please review it?

jhurliman (changes requested):
The settings are intended to persist to disk, `THREE.Object3D` is not a good candidate for persistence.

3D Extensions should be introduced through the extension API, and it will require a proposal and agreement on the proposal before committing to an API. This is probably best left until after the new 3D panel is at feature parity with the previous one and we've made the transition over.

dudv:
Agreed that it should not be part of the settings then. How about passing it as a regular props? Or perhaps just exposing the Scene object in the meanwhile? I think it would be beneficial to experiment with injecting external three.js objects early on so that we get better understanding of the functionality we need to support it, even if the API will change in the future.

jhurliman:
@dudv that seems reasonable. I think the most straightforward way to enable that type of experimentation, for now, is to change these `Renderer` members to public:

```
private scene = new THREE.Scene();
private world = new THREE.Object3D();
private camera = new THREE.PerspectiveCamera();
```

dudv:
@jhurliman Could you please tag me on the PR with the new API? FWIW we've been able to integrate custom HD map rendering using a similar API to the one in this PR (i.e. injecting a named parent object into the scene).

PR #3278, opened 2022-04-28
Make URL state tracking more robust

### LINKED ISSUE #3188: URL's layoutId does not update correctly when there is no data loaded
**Description**
- URL's `layoutId` does not update correctly when there is no data loaded

- Version: 1.7
- OS: Web

**Steps To Reproduce**

- Open Studio app (studio.foxglove.dev) without any data loaded
- Click on any layout in the sidebar

**Expected Behavior**
- URL updates to have a `layoutId` param

**Actual Behavior**
- URL does not update to have a `layoutId` param

foxymiles:
**User-Facing Changes**
The `layoutId` url parameter (web app) updates when changing layouts and no data source is selected.

**Description**
This fixes an issue with layoutId not being tracked in the absence of a datasource. Currently `PlayerManager` has a separate notion of `selectedSource` and the actually active player. This makes managing URL state very difficult because these two things can be out of sync. This patch removes `selectedSource` as a separate state variable from `PlayerManager` and moves the `sourceId` that the url state hook needs into the live `PlayerState` so that it's always in sync with the other data the url state hook needs.

This patch also separates updates to the layoutId and time portions of the url from updates to the datasource related portions of the url. Overall the url state handling logic is much simpler and easier to understand with these changes.

Unfortunately this touches quite a few files but most of the changes are just plumbing through the `sourceId` to the players.

Also I should note that this _will_ show a `ds` value for local files like `?ds=mcap-local-file`. This makes the logic more consistent and I don't see any end-user value in making a special case for local files that shows no `ds`.

Fixes #3188

### DIFF
--- a/packages/studio-base/src/dataSources/FoxgloveDataPlatformDataSourceFactory.tsx
+++ b/packages/studio-base/src/dataSources/FoxgloveDataPlatformDataSourceFactory.tsx
@@ -53,6 +53,7 @@ class FoxgloveDataPlatformDataSourceFactory implements IDataSourceFactory {
       return new IterablePlayer({
         metricsCollector: args.metricsCollector,
         source,
+        sourceId: this.id,
         urlParams: {
           deviceId,
           start: toRFC3339String(startTime),
@@ -70,6 +71,7 @@ class FoxgloveDataPlatformDataSourceFactory implements IDataSourceFactory {
       },
       consoleApi: args.consoleApi,
       metricsCollector: args.metricsCollector,
+      sourceId: this.id,
     });
   }
 }
--- a/packages/studio-base/src/util/appURLState.ts
+++ b/packages/studio-base/src/util/appURLState.ts
@@ -21,23 +21,41 @@ export type AppURLState = {
  * @param urlState The player state to encode.
  * @returns A url with all app state stored as query pararms.
  */
-export function encodeAppURLState(url: URL, urlState: AppURLState): URL {
+export function updateAppURLState(url: URL, urlState: AppURLState): URL {
   const newURL = new URL(url.href);
 
-  // Clear all exisiting params first.
-  [...newURL.searchParams].forEach(([k, _]) => newURL.searchParams.delete(k));
-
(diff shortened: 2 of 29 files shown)


defunctzombie:
> Also I should note that this _will_ show a `ds` value for local files like `?ds=mcap-local-file`. This makes the logic more consistent and I don't see any end-user value in making a special case for local files that shows no `ds`.

I didn't experience this with the preview deployment. Not sure if thats a concern or not.

packages/studio-base/src/dataSources/FoxgloveDataPlatformDataSourceFactory.tsx
```diff
@@ -53,6 +53,7 @@ class FoxgloveDataPlatformDataSourceFactory implements IDataSourceFactory {
       return new IterablePlayer({
         metricsCollector: args.metricsCollector,
         source,
+        sourceId: this.id,
```
defunctzombie, line 56: The `sourceId` is more like a `sourceType` - not sure it is worth changing but figured I'd highlight that. To me `sourceId` implies a unique ID for the specific `source` we've instantiated.

packages/studio-base/src/util/appURLState.ts
```diff
@@ -21,23 +21,41 @@ export type AppURLState = {
...
+    } else {
+      newURL.searchParams.delete("ds");
+    }
   }
 
-  if (urlState.ds && urlState.dsParams) {
-    newURL.searchParams.set("ds", urlState.ds);
-    Object.entries(urlState.dsParams).forEach(([k, v]) => {
+  if ("dsParams" in urlState) {
+    [...newURL.searchParams].forEach(([k, _]) => {
```
defunctzombie, line 52: Any particular reason you copy the array and use `forEach` here rather than `for (const [k, _] of newURL.searchParams)` ?

defunctzombie (approved):
👍 on dropping _selectedSource_. That's some outdated stuff that I am happy to see go.

Double check the url state handling when changing sources (local to remote, etc).

PR #3279, opened 2022-04-28
Load blocks in ranges

defunctzombie:
**User-Facing Changes**
Users with experimental players enabled will see an improvement in message loading for full preload topics (i.e. plot panel, state transition panel, map panel).

**Description**
The block loading behavior in IterablePlayer tries to be too clever at the expense of poor loading performance for typical data sources. The current behavior tries to load blocks outward from the seek location - loading the block before, then after, then before - 1, then after + 1, etc. This is inefficient as it results in many requests and constantly making new iterators to load small bits of the data source from different regions of the data source. Most data sources prefer loading continuous regions from some time forward.

This change alters this behavior to favor continuous loading. Starting at the seek time, loading computes "spans" of continuous blocks. Continuous blocks are those that are missing required topics (to avoid fetching a block topic if we've already done so). These continuous blocks minimize iterator creation which speeds up loading data under various common situations.

This change refactors block loading out of IterablePlayer and into a new BlockLoader class to aid in testing block loading logic.

Fixes: #3740


packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
```diff
@@ -859,219 +842,40 @@ export class IterablePlayer implements Player {
...
-      const postId = postIds.shift();
-      if (postId != undefined) {
-        loadQueue.push(postId);
-      }
-      const preId = preIds.shift();
-      if (preId != undefined) {
-        loadQueue.push(preId);
-      }
+    if (this._abort) {
+      throw new Error("Invarient. Abort controller already defined");
```
foxymiles, line 848:
```suggestion
      throw new Error("Invariant. Abort controller already defined");
```

packages/studio-base/src/players/IterablePlayer/BlockLoader.ts
```diff
@@ -0,0 +1,388 @@
...
+    log.debug(`Block count: ${blockCount}`);
+    this.blocks = Array.from({ length: blockCount });
+  }
+
+  setTopics(topics: Set<string>): void {
+    log.debug("setTopics", topics);
+    this.topics = topics;
+  }
+
+  async load(args: LoadArgs): Promise<void> {
```
foxymiles, line 84: Would it be helpful to break any of this up into separately testable bits of logic?

foxymiles (approved):

packages/studio-base/src/players/IterablePlayer/BlockLoader.ts
defunctzombie: Probably. Tho since I'm actively iterating on this and it might change with caching work (or maybe not) I think I'll tackle in a follow-on PR.

I think the span computation and the repeated code for mutating the blocks are good candidates for better organization. The span computation could then be tested in isolation.

PR #3286, opened 2022-04-29
Tweak light mode background colors.

2metres:
**User-Facing Changes**
Adjust light mode background color heirarchy to be more consisitent with dark mode styles

**Description**


defunctzombie:
> Adjust light mode background color heirarchy to be more consisitent

Consistent with what?

2metres:
> Consistent with what?

Dark mode

jtbandes:
Looks like we are losing the grid on the legacy plot panel in light mode:

image

jtbandes:
You've got some test failures:

> Warning: React does not recognize the `shouldShow` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `shouldshow` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
[61](https://github.com/foxglove/studio/runs/6281133224?check_suite_focus=true#step:6:61)

2metres:
@amacneil will address the tabs in a separate PR

jtbandes (approved):

packages/studio-base/src/components/PanelToolbar/index.stories.tsx
```diff
@@ -41,9 +42,15 @@ class MosaicWrapper extends React.Component<{
             renderPreview={() => undefined as any}
           >
             <HelpInfoProvider>
-              <div style={{ width, height: 300, padding: 30, position: "relative" }}>
-                {id === "Sibling" ? "Sibling Panel" : this.props.children}
-              </div>
+              <Box
+                width="100%"
```
defunctzombie, line 46: Didn't we determine box is slow? we should be using divs with useStyle

packages/studio-base/src/components/PanelToolbar/index.stories.tsx
```diff
@@ -135,7 +142,7 @@ storiesOf("components/PanelToolbar", module)
       </Icon>
     );
     return (
-      <MosaicWrapper width={500}>
+      <MosaicWrapper width={468}>
```
defunctzombie, line 145: What do these have to do with light mode?

packages/studio-base/src/components/PanelToolbar/index.stories.tsx
2metres: For stories I feel like Box is fine.

packages/studio-base/src/components/PanelToolbar/index.stories.tsx
2metres: The spacing on the PanelToolbar changed this is attempting to keep the visual diff as similar as possible in the stoybook

PR #3292, opened 2022-04-29
Avoid addMessages in plot panel depending on blocks

defunctzombie:
**User-Facing Changes**
None

**Description**
During initial datasource loading, blocks are changing quickly. This resulted in the addMessages function constantly changing because it had logic that depended on ignoring paths that are present in blocks. useMessageReducer doesn't like that and produces a lot of warnings to the console because it keeps re-running the reducer.

This change memoizes the paths that exist in blocks and uses this variable for the addMessages logic rather than all the block data. These paths are more stable and do not change as blocks are loading.

While we no longer spam warning messages about "addMessages changing too frequently" to the console, we are now comparing the array of paths on every render. Practically this is no different than comparing all the dependency arrays for react hooks since the paths array is typically short.

### DIFF
--- a/packages/studio-base/src/panels/Plot/index.tsx
+++ b/packages/studio-base/src/panels/Plot/index.tsx
@@ -298,6 +298,13 @@ function Plot(props: Props) {
     [allPaths],
   );
 
+  // The addMessages function below is passed to useMessageReducer to handle new messages during
+  // playback. If we have messages for a specific path in _blocks_ then we ignore the messages in
+  // the reducer.
+  //
+  // To keep the addMessages function "stable" when loading new blocks we grab only the paths from
+  // the blocks and make addMessages depend on the paths. To keep paths referentially stable when
+  // the paths values haven't changed we use a shallow memo.
   const blockPaths = useMemo(() => Object.keys(plotDataForBlocks), [plotDataForBlocks]);
   const blockPathsMemo = useShallowMemo(Object.keys(blockPaths));
 
--- a/packages/studio-base/src/panels/Plot/index.tsx
+++ b/packages/studio-base/src/panels/Plot/index.tsx
@@ -306,7 +306,7 @@ function Plot(props: Props) {
   // the blocks and make addMessages depend on the paths. To keep paths referentially stable when
   // the paths values haven't changed we use a shallow memo.
   const blockPaths = useMemo(() => Object.keys(plotDataForBlocks), [plotDataForBlocks]);
-  const blockPathsMemo = useShallowMemo(Object.keys(blockPaths));
+  const blockPathsMemo = useShallowMemo(blockPaths);
 
   const addMessages = useCallback(
     (accumulated: PlotDataByPath, msgEvents: readonly MessageEvent<unknown>[]) => {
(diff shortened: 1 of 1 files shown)


defunctzombie:
If we think this is reasonable I can add a comment above the memo calls.

I am a bit sad that there is not a more intuitive way to accomplish this type of memoization given reacts reliance of reference identity.

packages/studio-base/src/panels/Plot/index.tsx
```diff
@@ -297,6 +298,9 @@ function Plot(props: Props) {
     [allPaths],
   );
 
+  const blockPaths = useMemo(() => Object.keys(plotDataForBlocks), [plotDataForBlocks]);
```
foxymiles, line 301: Could you just use a single instance of `useDeepMemo` here?

jtbandes: I think the intent is to only get the paths (keys), not the messages, so we don't want deep equality

packages/studio-base/src/panels/Plot/index.tsx
```diff
@@ -315,7 +319,7 @@ function Plot(props: Props) {
         for (const path of paths) {
           // Skip any paths we already service in plotDataForBlocks.
           // We don't need to accumulate these because the block data takes precedence.
-          if (path in plotDataForBlocks) {
+          if (blockPathsMemo.includes(path)) {
```
jtbandes, line 322: should we make it a Set so we can use `.has()`?

packages/studio-base/src/panels/Plot/index.tsx
defunctzombie: Right, I don't want to deep memo on the plotData, only the paths (which are the keys)

packages/studio-base/src/panels/Plot/index.tsx
defunctzombie: image

/shrug

packages/studio-base/src/panels/Plot/index.tsx
```diff
@@ -297,6 +298,16 @@ function Plot(props: Props) {
...
 
+  // The addMessages function below is passed to useMessageReducer to handle new messages during
+  // playback. If we have messages for a specific path in _blocks_ then we ignore the messages in
+  // the reducer.
+  //
+  // To keep the addMessages function "stable" when loading new blocks we grab only the paths from
+  // the blocks and make addMessages depend on the paths. To keep paths referentially stable when
+  // the paths values haven't changed we use a shallow memo.
+  const blockPaths = useMemo(() => Object.keys(plotDataForBlocks), [plotDataForBlocks]);
+  const blockPathsMemo = useShallowMemo(Object.keys(blockPaths));
```
jtbandes, line 309: Still think it could be worth the complexity to get a perf improvement by avoiding the Object.keys and shallow equality comparison if blockPaths has not changed...

On the other hand this currently looks wrong, you are doing Object.keys on blockPaths which is the result of Object.keys. Does this work and pass tests?

defunctzombie: This is wrong - its code from a different attempt. It should be `useShallowMemo(blockPaths)`

jtbandes (approved):

PR #3295, opened 2022-05-02
Fix numberinput step and add min/max

### LINKED ISSUE #3276: Pressing up/down arrows in settings sidebar number input does not respect `step`
Pressing the left/right arrows for a number input in the settings sidebar properly respects the `step` setting, but pressing up/down arrow keys on the keyboard always steps by `1.0`.

foxymiles:
**User-Facing Changes**
None

**Description**
This fixes an issue with the number inputs in the new settings UI. The `step` value now also controls increments initiated by pressing the up & down arrow keys.

This PR also adds support for a min & max value on number inputs that will be enforced for user edits.

Fixes #3276

### DIFF
--- a/packages/studio-base/src/components/SettingsTreeEditor/inputs/NumberInput.tsx
+++ b/packages/studio-base/src/components/SettingsTreeEditor/inputs/NumberInput.tsx
@@ -5,7 +5,7 @@
 import ChevronLeftIcon from "@mui/icons-material/ChevronLeft";
 import ChevronRightIcon from "@mui/icons-material/ChevronRight";
 import { IconButton, TextFieldProps, TextField, styled as muiStyled } from "@mui/material";
-import { ReactNode } from "react";
+import { ReactNode, useCallback } from "react";
 import { useKeyPress } from "react-use";
 
 import { fonts } from "@foxglove/studio-base/util/sharedStyleConstants";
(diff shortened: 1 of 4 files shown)


defunctzombie (approved):

packages/studio-base/src/components/SettingsTreeEditor/inputs/NumberInput.tsx
```diff
@@ -62,20 +64,36 @@ export function NumberInput(
...
       }
       type="number"
+      inputProps={{ max: props.max, min: props.min, step: stepAmount }}
       InputProps={{
         startAdornment: (
           <StyledIconButton
             size="small"
             edge="start"
-            onClick={() => value != undefined && onChange(value - stepAmount)}
+            onClick={() => updateValue((value ?? 0) - stepAmount)}
```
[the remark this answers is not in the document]
foxymiles: I guess we also don't want `min` because then we'll go to `min - stepAmount`? But you're right 0 is also not correct.

packages/studio-base/src/components/SettingsTreeEditor/inputs/NumberInput.tsx
```diff
@@ -62,20 +64,36 @@ export function NumberInput(
 
   const stepAmount = shiftPressed ? step * 10 : step;
 
+  const updateValue = useCallback(
+    (newValue: number) => {
+      if (props.max != undefined && newValue > props.max) {
+        return;
```
defunctzombie, line 70: Should this set the field to the max value?

foxymiles: Yeah I suppose that would make more sense. It would trigger another invocation of the actionHandler but that should be written to handle two actions with the same value anyway seems like.

Issue #3297, opened 2022-05-02
Seeking with new default bag player causes black screen flickering


amacneil:
**Description**
New experimental bag player shows black screen while seeking. This is quite jarring and significant regression from the old bag player.

- Version: main
- OS: mac/web

https://user-images.githubusercontent.com/637671/166329122-f970c13c-9dce-4713-9b7c-83e1e93c8244.mov

PR #3298, opened 2022-05-03
Allow 100 milliseconds for IterablePlayer to backfill

defunctzombie:
**User-Facing Changes**
When seeking or scrubbing, the panels no longer flash to an empty state. They display the previous state for a short window (100 milliseconds), and if the new seek location is not loaded by this time, _then_ the panels flash to an empty state indicating loading. If the new seek location _is_ loaded before the timeout, then the panels display the new message data.

**Description**
When seeking, the IterablePlayer would immediately emit a message with the new seek time. This would show an updated seek location in the scrub bar but would also clear all messages from the screen since there were no messages loaded for the seek time yet.

This change gives the backfill 100 milliseconds to emit state before emitting the new seek time. For local data this allows users to scrub continuously and see the scene or messages rather than a black screen.

Fixes: #3297

### DIFF
--- a/packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
+++ b/packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
@@ -541,12 +541,20 @@ export class IterablePlayer implements Player {
     // If the backfill does not complete within 100 milliseconds, we emit a seek event with no messages.
     // This provides feedback to the user that we've acknowledged their seek request but haven't loaded the data.
     const seekAckTimeout = setTimeout(async () => {
+      // fixme - we never emit anything now because we constantly keep scrubbing
+      // so do we want to allow some previous state to emit? and then start again?
+      // Why would we emit previous messages?
+      if (this._nextState) {
+        return;
+      }
+
       this._messages = [];
       this._currentTime = targetTime;
       this._lastSeekEmitTime = Date.now();
 
       // emit a state message for the seek time, we only do this if we haven't already emitted one
       await this._emitState();
+      // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition, @typescript-eslint/strict-boolean-expressions
       if (this._nextState) {
         return;
       }
--- a/packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
+++ b/packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
@@ -538,13 +538,19 @@ export class IterablePlayer implements Player {
     this._lastMessage = undefined;
     this._seekTarget = undefined;
 
-    this._messages = [];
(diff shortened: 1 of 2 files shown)


defunctzombie:
@amacneil Give this PR a try and see what you think for local bag files and for the sample data.

amacneil:
Still not as good as the original bag player. Dragging the seek bar in the old bag player was nearly as smooth as VLC. This PR has maybe 100ms delay when clicking the seek bar (fine), but when dragging it shows the old frame instead of updating in real time like the old player did.

amacneil:
Weird, I just looked at your video and it is much smoother. That is not what is happening for me. Are you sure you're using the experimental player in your video above?

Here's my experience - it doesn't redraw the 3d panel until my mouse stops moving:

https://user-images.githubusercontent.com/637671/166405513-36d229aa-f14f-4a29-ab02-e7108cd0a58d.mov

defunctzombie:
> Weird, I just looked at your video and it is much smoother. That is not what is happening for me. Are you sure you're using the experimental player in your video above?

I made the video before I made some additional tweaks to fix an issue. I will remove the video and look at this again to see if there's additional tweaks I can make.

One difference the new player will have over the old is that the new player uses a backfill method that is arguably more expensive than the previous player. The new player looks back to find the latest message on each topic rather than a fixed lookback window that the old player did. Without more aggressive caching this has a slower load time. You can notice this more for remote bag files - and even more for data platform. We need to make a tradeoff here on things looking like a video file vs loading the "right" data for the user. Its not as simple as video loading strategies which can do something basic like look back a small fixed amount.

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
```diff
@@ -538,13 +538,19 @@ export class IterablePlayer implements Player {
...
+
+    // If the backfill does not complete within 100 milliseconds, we emit a seek event with no messages.
+    // This provides feedback to the user that we've acknowledged their seek request but haven't loaded the data.
+    const seekAckTimeout = setTimeout(() => {
+      this._messages = [];
+      this._currentTime = targetTime;
+      this._lastSeekEmitTime = Date.now();
+
+      seekAckWait = this._emitState();
+    }, 100);
```
jhurliman, line 553: I suggest changing to 200 or 250

jhurliman (approved):

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
defunctzombie: Based on my scientific google search ™️ I'm gonna stick with 100ms here.

https://www.nngroup.com/articles/response-times-3-important-limits/

Issue #3282, opened 2022-04-28
Proposal: remove "header stamp" playback dropdown


amacneil:
No one can tell me what this dropdown is used for:

<img width="260" alt="image" src="https://user-images.githubusercontent.com/637671/165849920-f95f2fc6-45b2-4d38-a742-388c35afd33c.png">

The 3d panel seems to ignore it, and uses header stamp or receive time depending on the topic. The plot panel definitely ignores it and uses its own per-series settings. So I'm not clear what this dropdown actually affects and why someone would use it.

defunctzombie:
The dropdown would affect _playback_. Imagine "Header stamp" means "publish time" playback. The problem tho is that none of the data we typically play is indexed in a manner to support such a playback "correctly". Right now I believe the player logic buffers some window of data, re-orders it by header stamp, and then plays that.

PR #3311, opened 2022-05-04
Move header.stamp playback ordering behind an experimental flag

jhurliman:
**User-Facing Changes**

- The message playback ordering dropdown is moved behind an experimental flag until header.stamp playback is correctly implemented.

**Description**

The current implementation of header.stamp playback does not do what users may intuitively guess it does and fully order the playback stream by header.stamp fields. Instead, it keeps whatever coarse ordering the data source uses (receive time) and sorts messages by header.stamp within an internal lookahead window. If header.stamp is not available the message is skipped.


defunctzombie:
Related issue ticket: https://github.com/foxglove/studio/issues/3282

When there's a related issue ticket lets link them in the descriptions or comments so we have a nice trail for folks to follow either way.

packages/studio-base/src/components/ExperimentalFeatureSettings.tsx
```diff
@@ -114,6 +114,16 @@ if (process.env.NODE_ENV === "development") {
       </>
     ),
   });
+  features.push({
+    key: AppSetting.EXPERIMENTAL_MESSAGE_ORDER,
+    name: "Configurable message playback order",
+    description: (
+      <>
+        Allow messages to be sorted by either receive time or header.stamp during playback. Playback
+        with header.stamp mode is experimental and not fully implemented.
```
defunctzombie, line 123: It's not that it isn't fully implemented right? Its that it has certain limitations - i.e. if your header stamps are >X seconds away from receive they won't play?

jhurliman: I would describe that as a partial implementation in user-facing text. Saying it has certain limitations makes it sound final, but it could have better behavior with a more fleshed-out implementation.

defunctzombie:
@amacneil thoughts? I think I'm 👍 for this tho it does feel like if you have a _relatively_ close file with header stamps then things might generally work. It definitely won't work as well with data platform.

amacneil:
I think this is reasonable. Is it off by default?

"Playback by header stamp order" sounds like a good feature for us to have, but my understanding is that this setting does not currently do that, so having a broken feature is worse than having no feature imo.

jhurliman:
Yes, off by default

amacneil:
Sounds good. After we disable it we can see if anyone asks about it, and check telemetry to see how many folks are re-enabling it.

defunctzombie (approved):

PR #3314, opened 2022-05-04
Don't use header.stamp time to expire point markers

### LINKED ISSUE #3283: Adding decay time makes point cloud disappear
**Description**
I have a file with sensible receive times in a PointCloud2 message, and incorrect "header stamp" values.

Playing back this file works fine. However, if I add "decay" to the topic in 3d panel, the 3d panel goes blank.

image

Apparently this is due to the invalid header stamp times. If we are playing back by receive time, then we should decay by receive time, and header stamp time shouldn't factor in to it.

- Version: main
- OS: web

foxymiles:
**User-Facing Changes**
This fixes an issue with the display of point cloud messages in the 3d panel when a delay time is set.

**Description**
We were using the message's `header.stamp` time to determine the expiration time of the point cloud message. We should use the messages `receiveTime` in this case.

Fixes #3283

### DIFF
--- a/packages/studio-base/src/panels/ThreeDimensionalViz/SceneBuilder/MessageCollector.ts
+++ b/packages/studio-base/src/panels/ThreeDimensionalViz/SceneBuilder/MessageCollector.ts
@@ -131,8 +131,7 @@ export default class MessageCollector {
     const result: ObjectWithInteractionData[] = [];
     this.markers.forEach((marker, key) => {
       // Check if the marker has a lifetime and should be deleted
-      const messageStamp =
-        (marker.message as { header?: { stamp?: Time } }).header?.stamp ?? marker.receiveTime;
+      const messageStamp = marker.receiveTime;
       if (MessageCollector.markerIsExpired(marker.lifetime, messageStamp, this.clock)) {
         this.markers.delete(key);
       } else {
--- a/packages/studio-base/src/panels/ThreeDimensionalViz/SceneBuilder/MessageCollector.test.ts
+++ b/packages/studio-base/src/panels/ThreeDimensionalViz/SceneBuilder/MessageCollector.test.ts
@@ -110,7 +110,7 @@ describe("MessageCollector", () => {
     expect(collector.getMessages()).toHaveLength(1);
 
     collector.setClock({ sec: 100, nsec: 90 + lifetimeNanos + 1 });
-    expect(collector.getMessages()).toHaveLength(0);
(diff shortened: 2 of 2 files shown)


jtbandes:
Contradicts https://github.com/foxglove/studio/pull/2986

foxymiles:
> Contradicts #2986

See slack thread

jtbandes:
Slack threads are not publicly viewable and not persistent. Can you screenshot or summarize the relevant discussion here so that a future (or present) reader knows why this change is being made? The PR description currently explains what is changing, but not really _why_.

packages/studio-base/src/panels/ThreeDimensionalViz/SceneBuilder/MessageCollector.ts
```diff
@@ -131,8 +131,7 @@ export default class MessageCollector {
     const result: ObjectWithInteractionData[] = [];
     this.markers.forEach((marker, key) => {
       // Check if the marker has a lifetime and should be deleted
-      const messageStamp =
-        (marker.message as { header?: { stamp?: Time } }).header?.stamp ?? marker.receiveTime;
```
jtbandes, line 135: Doesn't this apply the change to all displayable objects and not just point clouds? Was that intentional?

foxymiles: @jhurliman Said we should just revert this change from #2986 so yeah it's my understanding that this is the intention.

foxymiles: This was our old behavior:

```typescript
  isExpired(currentTime: Time) {
    if (this.lifetime == undefined) {
      // Do not expire markers if we can't tell what their lifetime is.
      // They'll be flushed later if needed (see flush below)
      return false;
    }
    const lifetime: Time = this.lifetime;

    if (areEqual(lifetime, ZERO_TIME)) {
      // Do not expire markers with infinite lifetime (lifetime == 0)
      return false;
    }

    // we use the receive time (clock) instead of the header stamp
    // to match the behavior of rviz
    const expiresAt = add(this.receiveTime, lifetime);

    return isGreaterThan(currentTime, expiresAt);
  }
```

jhurliman (changes requested):
This isn't correct, it mixes the `receiveTime` and `header.stamp` time sources for some renderables.

Issue #3304, opened 2022-05-03
Move settings sidebar import/export and reset to defaults buttons into a context menu


foxymiles:
**Description**
These buttons are too prominent but the functions are still useful so let's move them into a context menu alongside the top heading in the panel settings sidebar.

- Version: 1.9.1
- OS: all

<img width="391" alt="Screen Shot 2022-05-03 at 1 00 46 PM" src="https://user-images.githubusercontent.com/93935560/166514086-f32b6599-9a6f-4ba8-8c5f-b1d3053a43c5.png">

PR #3316, opened 2022-05-05
Move settings panel actions buttons into action menu

foxymiles:
**User-Facing Changes**
This moves the action button at the bottom of the settings sidebar to an action menu in the title.

**Description**
These functions are still useful but the big buttons we currently have at the bottom of the settings sidebar are too prominent. This PR moves them to an action menu with a trigger button to the right of the title.

Screen Shot 2022-05-05 at 7 13 17 AM

Fixes https://github.com/foxglove/studio/issues/3304


2metres (approved):

defunctzombie:
Do we think having the `settings` word is necessary in the context menu?

foxymiles:
Feels like being specific here doesn't hurt? Import/export could refer to many things.

If anything maybe we should change `reset to defaults` -> `reset to default settings`?

PR #3317, opened 2022-05-05
Added support for running Foxglove Studio behind a proxy and under a subpath (not just `/`)

jpc:
**User-Facing Changes**

Hopefully none.

**Description**

Foxglove Studio is quite difficult to put behind a proxy server right now because it insists on loading all the assets from the root path of the server. Since it is a single-page application we only need to change the URLs for the favicons and the initial JS bundle to relative and everything else seems to work automatically.

### DIFF
--- a/web/webpack.config.ts
+++ b/web/webpack.config.ts
@@ -91,7 +91,7 @@ const mainConfig = (env: unknown, argv: WebpackArgv): Configuration => {
     devtool: isDev ? "eval-cheap-module-source-map" : "source-map",
 
     output: {
-      publicPath: "/",
+      publicPath: "./",
 
       // Output filenames should include content hashes in order to cache bust when new versions are available
       filename: isDev ? "[name].js" : "[name].[contenthash].js",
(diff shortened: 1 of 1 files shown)


web/webpack.config.ts
```diff
@@ -91,7 +91,7 @@ const mainConfig = (env: unknown, argv: WebpackArgv): Configuration => {
     devtool: isDev ? "eval-cheap-module-source-map" : "source-map",
 
     output: {
-      publicPath: "/",
+      publicPath: "./",
```
defunctzombie, line 94: I think we should use `auto`.

https://webpack.js.org/guides/public-path/#automatic-publicpath

web/webpack.config.ts
```diff
@@ -112,9 +112,9 @@ const mainConfig = (env: unknown, argv: WebpackArgv): Configuration => {
   <html>
     <head>
       <meta charset="utf-8">
-      <link rel="apple-touch-icon" sizes="180x180" href="/apple-touch-icon.png" />
-      <link rel="icon" type="image/png" sizes="32x32" href="/favicon-32x32.png" />
-      <link rel="icon" type="image/png" sizes="16x16" href="/favicon-16x16.png" />
+      <link rel="apple-touch-icon" sizes="180x180" href="./apple-touch-icon.png" />
```
defunctzombie, line 115: If you use publicPath: auto, I think you can remove the `./` from these and leave the filename.

web/webpack.config.ts
jpc: Sound good. We probably don’t care about IE so the caveat does not apply. I’ll change, test, and update the pull request, thanks.

defunctzombie: We definitely don't care about `IE`.

foxymiles, line 94:
```suggestion
      publicPath: "auto",
```

web/webpack.config.ts
```diff
@@ -112,9 +112,9 @@ const mainConfig = (env: unknown, argv: WebpackArgv): Configuration => {
...
+      <link rel="icon" type="image/png" sizes="32x32" href="./favicon-32x32.png" />
+      <link rel="icon" type="image/png" sizes="16x16" href="./favicon-16x16.png" />
```
foxymiles, line 117:
```suggestion
      <link rel="apple-touch-icon" sizes="180x180" href="apple-touch-icon.png" />
      <link rel="icon" type="image/png" sizes="32x32" href="favicon-32x32.png" />
      <link rel="icon" type="image/png" sizes="16x16" href="favicon-16x16.png" />
```

defunctzombie:
We landed these changes in https://github.com/foxglove/studio/pull/3460. Thanks for your contribution!

Issue #3277, opened 2022-04-28
Settings sidebar issues


jhurliman:
I've been implementing the settings sidebar in the 3d panel and overall it looks great! Here are the issues I've encountered:

- [x] 3D panel drops many frames when panning the camera (updating the sidebar) with all topic settings implemented
- [x] The default sidebar width is not wide enough for almost any of the topic settings labels. they are cut off with ellipses
- [x] Settings nodes need an option to display as rolled up by default (3D panel settings is overwhelming with everything unrolled)
- [x] `"gradient"` isn't usable yet, value should have type `[string, string]` and the control be a two-step gradient picker, not a color map dropdown
- [x] `"number"` would be much easier to implement with `min`, `max`
- [x] `"number"` could have better UX if the arrow buttons and arrow keys worked when there is only a placeholder value (parse as a number or fall back to 0)
- [x] `"number"` shows too many digits of precision
- [x] `"select"` setting value: `""` doesn't select the option entry with value set to `""`
- [x] Need a visibility toggle control and some field or event for visibility changes
- [x] The single `"color"` is confusing, i need separate `"rgb"` and `"rgba"` input types
- [x] `"color"` could benefit from a placeholder color option
- [x] We might need a validation step and error highlighting (color input red?). example: type only `e` in a number input
- [x] Need `step` on `vec3` input

Roman's stuff:

Working on input for custom map tile layers. I'd like to inform the user when they enter an invalid input into a field:

![image](https://user-images.githubusercontent.com/84792/168347877-2630faab-8cff-479f-82ef-fcf699261010.png)

Ideally the URL input field is in an error state and I can show the user a message on what is wrong.

foxymiles:
@jhurliman The number of digits shown seems like something we should leave up to panel implementers to implement when they build the settings tree.

foxymiles:
For rgba colors do you want them as css `rgba()` style strings or as 8-digit hex numbers?

foxymiles:
Actually it looks like our current color picker widget doesn't handle 8 digit hex numbers well so for now it's just going to be `rgba()`.

jhurliman:
@foxymiles where do I put that function call? If there was a format callback attribute on number fields I could use it, but I don't see how to incorporate that currently.

foxymiles:
Just put in in the code where you build the settings tree in the `value` attribute for the field.

[...]

PR #3319, opened 2022-05-05
Add basic gradient picker to settings UI

foxymiles:
**User-Facing Changes**
This adds a simple, two-point gradient color picker to the settings UI.

**Description**
This just uses two instances of the color picker and a gradient background to show the selected gradient.

https://user-images.githubusercontent.com/93935560/166979624-6da711ab-3fb7-42e4-9ce2-3a78b7a1c244.mov

Partially addresses https://github.com/foxglove/studio/issues/3277

### DIFF
--- a/packages/studio-base/src/components/SettingsTreeEditor/inputs/ColorGradientInput.tsx
+++ b/packages/studio-base/src/components/SettingsTreeEditor/inputs/ColorGradientInput.tsx
@@ -3,7 +3,7 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
 import { ColorPicker } from "@fluentui/react";
-import { Card, ClickAwayListener, TextField, styled as muiStyled } from "@mui/material";
+import { Popover, TextField } from "@mui/material";
 import { useState } from "react";
 import tinycolor from "tinycolor2";
 
@@ -12,21 +12,6 @@ import { fonts } from "@foxglove/studio-base/util/sharedStyleConstants";
 
 import { ColorSwatch } from "./ColorSwatch";
 
-const PickerWrapperLeft = muiStyled(Card)(({ theme }) => ({
-  position: "absolute",
-  zIndex: theme.zIndex.modal,
-  top: "100%",
-  left: 0,
-  transform: "translateX(-50%)",
-}));
-
-const PickerWrapperRight = muiStyled(Card)(({ theme }) => ({
-  position: "absolute",
-  zIndex: theme.zIndex.modal,
-  top: "100%",
-  right: 0,
-}));
-
 export function ColorGradientInput({
   colors,
   onChange,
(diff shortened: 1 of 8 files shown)


packages/studio-base/src/components/SettingsTreeEditor/inputs/ColorGradientInput.tsx
```diff
@@ -0,0 +1,118 @@
...
+  colors,
+  onChange,
+}: {
+  colors: undefined | readonly [string, string];
+  onChange: (colors: [string, string]) => void;
+}): JSX.Element {
+  const [leftAnchor, setLeftAnchor] = useState<undefined | HTMLDivElement>(undefined);
+  const [rightAnchor, setRightAnchor] = useState<undefined | HTMLDivElement>(undefined);
+
+  const handleLeftClick = (event: React.MouseEvent<HTMLDivElement>) => {
```
jtbandes, line 25: useCallback

packages/studio-base/src/components/SettingsTreeEditor/inputs/ColorGradientInput.tsx
```diff
@@ -0,0 +1,118 @@
...
+  return (
+    <Stack
+      direction="row"
+      style={{
+        position: "relative",
+        backgroundImage: `linear-gradient(to right, ${safeLeftColor}, ${safeRightColor})`,
+      }}
+    >
+      <ColorSwatch color={safeLeftColor} onClick={handleLeftClick} />
+      <TextField
```
jtbandes, line 54: What’s this for?

packages/studio-base/src/components/SettingsTreeEditor/inputs/ColorScalePicker.tsx
```diff
@@ -124,6 +124,7 @@ const ColorScale = muiStyled("div")<{ colorScale: string[] }>(({ theme, colorSca
   flex: "auto",
 }));
 
+// ts-prune-ignore-next
 export function ColorScalePicker(props: ColorScalePickerProps): JSX.Element {
```
jtbandes, line 128: Should we just delete this instead?

jtbandes (approved):

packages/studio-base/src/components/SettingsTreeEditor/inputs/ColorScalePicker.tsx
foxymiles: I think we're going to add this back as a `colormap` input or something like that.

packages/studio-base/src/components/SettingsTreeEditor/inputs/ColorGradientInput.tsx
foxymiles: Just to keep the layout consistent.

foxymiles: The TextField is useful because it has the same characteristics as the other input fields so we get a consistent layout with no extra effort. We could probably achieve the same thing with a spacer and manually setting the background color but it didn't seem worth spending the extra time on it.

PR #3320, opened 2022-05-05
Improve color picker positioning on small screens

foxymiles:
**User-Facing Changes**
None

**Description**
This improves the positioning of the color picker input in the settings UI on small screens where the sidebar is narrow. The solution here is to center-align instead of left-align and to set a min-width on the picker.

Screen Shot 2022-05-05 at 2 23 20 PM


packages/studio-base/src/components/SettingsTreeEditor/inputs/ColorPickerInput.tsx
```diff
@@ -43,6 +43,8 @@ const Root = muiStyled("div")({
 const PickerWrapper = muiStyled(Card)(({ theme }) => ({
   position: "absolute",
   zIndex: theme.zIndex.modal,
+  left: "0%",
+  transform: "translateX(-50%)",
```
jtbandes, line 47: Seems a bit strange that we do custom positioning on this popup. Does MUI not provide some kind of popup component we can use that will automatically put it in a separate layer above all other elements?

foxymiles: I wasn't sure if the fluent color picker would behave well in a MUI popup but it seems ok. Good call. I've changed it.

packages/studio-base/src/components/SettingsTreeEditor/inputs/ColorPickerInput.tsx
```diff
@@ -54,13 +43,22 @@ type ColorPickerInputProps = {
 
 export function ColorPickerInput(props: ColorPickerInputProps): JSX.Element {
   const { onChange, swatchOrientation = "start", value } = props;
-  const [showPicker, setShowPicker] = useState(false);
+
+  const [anchorElement, setAnchorElement] = useState<undefined | HTMLDivElement>(undefined);
+
+  const handleClick = (event: MouseEvent<HTMLDivElement>) => {
```
jtbandes, line 49: Since the color may change rapidly while dragging, probably want to wrap these in useCallback to avoid re-adding listeners on each render

jtbandes (approved):
Nice 👍

Does MUI also have a color picker we should switch to someday?

packages/studio-base/src/components/SettingsTreeEditor/inputs/ColorPickerInput.tsx
foxymiles: I'd be surprised if it's a significant performance hit but I've wrapped these in `useCallback` just in case.

PR #3328, opened 2022-05-06
Update tab panel tabs to look more like tabs

2metres:
**User-Facing Changes**
Update tab panel tabs to look more like tabs in light mode

**Description**
Screen Shot 2022-05-10 at 11 52 24 am

### DIFF
--- a/packages/studio-base/src/components/EmptyPanelLayout.tsx
+++ b/packages/studio-base/src/components/EmptyPanelLayout.tsx
@@ -11,15 +11,13 @@
 //   found at http://www.apache.org/licenses/LICENSE-2.0
 //   You may not use this file except in compliance with the License.
 
-import { useTheme } from "@fluentui/react";
-import { Link, Theme, Typography } from "@mui/material";
-import { makeStyles } from "@mui/styles";
-import cx from "classnames";
+import { Link, Typography, styled as muiStyled } from "@mui/material";
 import { useCallback } from "react";
 import { useDrop } from "react-dnd";
 import { MosaicDragType } from "react-mosaic-component";
 
 import PanelList, { PanelSelection } from "@foxglove/studio-base/components/PanelList";
+import Stack from "@foxglove/studio-base/components/Stack";
 import { useCurrentLayoutActions } from "@foxglove/studio-base/context/CurrentLayoutContext";
 import { MosaicDropResult } from "@foxglove/studio-base/types/panels";
 import { getPanelIdForType } from "@foxglove/studio-base/util/layout";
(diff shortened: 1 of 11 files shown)


packages/studio-base/src/components/EmptyPanelLayout.tsx
```diff
@@ -28,22 +26,18 @@ type Props = {
...
-  },
-  dropzoneOver: {
+const Root = muiStyled("div")(({ theme }) => ({
+  backgroundColor: theme.palette.background.paper,
+  width: "100%",
+  height: "100%",
+  overflowY: "auto",
+}));
+
+const DropTarget = muiStyled("div")<{ isOver: boolean }>(({ isOver, theme }) => ({
```
defunctzombie, line 36: Whats the thinking for preferring muiStyled over useStyles?

packages/studio-base/src/panels/Tab/DraggableToolbarTab.tsx
```diff
@@ -45,10 +40,18 @@ export function DraggableToolbarTab(props: Props): JSX.Element {
...
-  const [{ isOver }, dropRef] = useDrop<TabLocation, void, { isOver: boolean }>({
+  const [{ highlight }, dropRef] = useDrop<
+    TabLocation,
+    void,
+    { highlight: "before" | "after" | undefined }
+  >({
     accept: TAB_DRAG_TYPE,
-    collect: (monitor) => ({
-      isOver: monitor.isOver(),
+    collect: (monitor: DropTargetMonitor<TabLocation, void>) => ({
```
defunctzombie, line 49: Maybe this should be extracted to a separate function? I find the layers of nested ternary hard to read and reason about. Comments might help too.

packages/studio-base/src/panels/Tab/index.tsx
```diff
@@ -30,13 +30,13 @@ import { SaveConfig } from "@foxglove/studio-base/types/panels";
 import { TAB_PANEL_TYPE } from "@foxglove/studio-base/util/globalConstants";
 import { DEFAULT_TAB_PANEL_CONFIG, updateTabPanelLayout } from "@foxglove/studio-base/util/layout";
 
-const SPanelCover = styled.div`
+const SPanelCover = muiStyled("div")`
```
defunctzombie, line 33: Maybe separate component files help organize and keep files more focused?

defunctzombie (approved):
Looks nicer. Left some thoughts on code structure. Something we don't do so well today but I'd like to try towards is keeping 1 component per component file. I think that will help us re-use and structure our components in a cleaner way.

packages/studio-base/src/panels/Tab/DraggableToolbarTab.tsx
2metres: @jtbandes can you weigh in here?

packages/studio-base/src/components/EmptyPanelLayout.tsx
2metres: `makeStyles` has been marked as legacy by MUI. `muiStyled` solves the styling issue easily without introducing the sx prop which we all agree has performance issues. By using `muiStyled` here I reduced the style approaches in this file alone from 3 to 1.

2metres:
I think the idea of 1 component per file is not neccisarily susutainable. Especially in the world of styled-components

amacneil:
👍 big improvement

@2metres it could use a tiny bit more styling for the inactive tabs - right now they are just floating with no borders or hover state.

image

Issue #3119, opened 2022-03-31
Visiting a url with a layoutId that is not yet loaded (from remote sync) doesn't work.


defunctzombie:
If I make a layout and send a link to the layoutId to a team member, that layout does not load for them if they haven't sync'd their layouts.

Layout loading should work even if there hasn't been a sync.

----
* Sign in to your foxglove account.
* Open studio.foxglove.dev and let the layouts sync.
* Open https://studio.foxglove.dev/?layoutId=ea800f2c-f854-49de-8ef1-a2f7e1f31719 and the layout should open
* Delete all the localstorage items for cached remote layouts and refresh the page. The layout will not open and you'll see a `Select a layout in the sidebar to get started!` message.

This is a bad experience because when I make a new shared layout and send it to a team member to view they can't open the link.

foxymiles:
What approach should we take here? Force a sync and then re-try to load the layout? Should we have some kind of modal state for this?

defunctzombie:
My initial thought is that the CurrentLayoutContext/Provider needs to be enhanced to support a "loading" state. I believe right now we have a binary state: no layout and layout. With no layout we show the "select a layout from the sidebar" fallback, and with a layout we show the layout.

We need to update the state to a type that can represent four states: loading, no layout, layout, information. Now when we have the layout manager load a layout by id, we can update the layout context to indicate loading while the manager loads the layout. If there is no layout to load, then we use the no-layout state. If we have a layout we can show that, and finally if there is some permission error or load failure we can show that too (and maybe offer retry).

The **tl;dr;** is that we need to enhance the notion of the "current layout state" beyond binary present or not preset.

PR #3329, opened 2022-05-06
Add remote fallback to LayoutManager#getLayout

defunctzombie:
**User-Facing Changes**
When a user loads a url with a layoutId, Studio attempts to load the layout id from their account (if signed in). This creates a better link-sharing experience when working with team shared layouts.

**Description**
When a user visits a studio url (or deep link) with a specific layoutId, the expectation is that the app will display this layout to the user. The app attempts to load the layout by asking the LayoutManager to getLayout with the layout id. The layout manager consults its local cache and returns the layout with that id.

When working in a team setting, a user may receive a link from a co-worker with a layout id. When the user clicks this link, studio attempts to open the layout id but is unable to because the app has not yet sync'd the user's local cache with the team's remote layouts. This leaves the user with an undesirable message to "Select a layout to get started" rather than showing the user the available layout.

This change alters the behavior of LayoutManager#getLayout to fall-back to fetching the remote layout if a local one is not available. The newly fetched remote layout is inserted into the local cache for future lookups.

Additionally, this change ports the layout debugging controls to MUI to continue our ongoing effort to deprecate fluent. These changes have no user facing impact.

Fixes: #3119

### DIFF
--- a/packages/studio-base/src/services/LayoutManager/LayoutManager.ts
+++ b/packages/studio-base/src/services/LayoutManager/LayoutManager.ts
@@ -28,9 +28,9 @@ import {
   IRemoteLayoutStorage,
   RemoteLayout,
 } from "@foxglove/studio-base/services/IRemoteLayoutStorage";
-import computeLayoutSyncOperations, {
-  SyncOperation,
-} from "@foxglove/studio-base/services/LayoutManager/computeLayoutSyncOperations";
+
+import { NamespacedLayoutStorage } from "./NamespacedLayoutStorage";
+import computeLayoutSyncOperations, { SyncOperation } from "./computeLayoutSyncOperations";
 
 const log = Logger.getLogger(__filename);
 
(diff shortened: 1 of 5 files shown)


defunctzombie:
Draft until tests are written.

packages/studio-base/src/services/LayoutManager/LayoutManager.ts
```diff
@@ -57,55 +57,6 @@ async function updateOrFetchLayout(
   }
 }
 
-/**
- * A wrapper around ILayoutStorage for a particular namespace.
- */
-class NamespacedLayoutStorage {
```
defunctzombie, line 63: I find it easier to separate classes into files. Helps me keep individual files more manageable.

defunctzombie:
@jtbandes There are no tests for LayoutManager. Is that tested some other way?

jtbandes:
Nope: https://github.com/foxglove/studio/issues/1773

It originally had [a bunch of tests](https://github.com/foxglove/studio/blob/c1b820b9fa84d02061546d74fea63ec97e1118ba/packages/studio-base/src/services/OfflineLayoutStorage/index.test.ts) and then when we determined the UX was not as desired, the [total rework](https://github.com/foxglove/studio/pull/1679) did not include new tests for fear of wasting lots of time writing them only to rewrite everything again (like what happened the first time).

packages/studio-base/src/services/LayoutManager/LayoutManager.ts
```diff
@@ -209,9 +160,48 @@ export default class LayoutManager implements ILayoutManager {
...
+      // Layout sync may have happened while we fetched the remote layout.
+      // We see if we have the layout locally and use that before caching the fetched remote layout.
+      const localLayout = await local.get(id);
+      if (localLayout) {
+        log.debug(`Local layout loaded while fetching remote id:${id}`);
+        return localLayout;
+      }
+
+      log.debug(`Adding layout to cache from getLayout: ${remoteLayout.id}`);
+      return await local.put({
```
jtbandes, line 197: Why implement a custom get/put logic instead of just triggering a regular sync and trying again?

defunctzombie: To avoid fetching all the other layouts in the "hot" path of wanting to load this one layout.

defunctzombie:
> I am still not sure it would be worth it because we've identified some deficiencies in the UX and I can feel another rewrite coming down the pipeline...

So this is 👍 to merge without tests for now on the thinking we'll be re-visiting this flow in a more substantial capacity?

jtbandes (approved):

PR #3331, opened 2022-05-06
Convert help info context to zustand

foxymiles:
**User-Facing Changes**
None

**Description**
It looks like the `HelpInfo` context was another context we've hacked with a hand rolled listener system to avoid the performance pitfalls of React contexts. We can replace this with Zustand.


packages/studio-base/src/panels/NodePlayground/index.tsx
```diff
@@ -129,8 +129,10 @@ function buildSettingsStree(config: Config): SettingsTreeNode {
   };
 }
 
+const selectSetHelpInfo = (store: HelpInfoStore) => store.setHelpInfo;
```
jtbandes, line 132: Does zustand recommend hoisting these selectors so they don't change, or are they ok with them changing? Are the selectors allowed to depend on other props?

foxymiles: There's a marginal performance benefit to hoisting them this way. 

https://github.com/pmndrs/zustand/discussions/658

The selector can be any function you want but you'd probably want to `useCallback` if you want it to be as stable as possible.

jtbandes: We used to have self-memoizing selectors like that from reselect. But we have pretty much fully transitioned to using React's memoization hooks/primitives.

packages/studio-base/src/components/HelpSidebar/index.stories.tsx
```diff
@@ -44,30 +41,6 @@ class MockPanelCatalog implements PanelCatalog {
   }
 }
 
-const MockHelpInfoContext = createContext<IHelpInfo | undefined>(undefined);
-MockHelpInfoContext.displayName = "MockHelpInfoContext";
-function MockHelpInfoProvider({ children }: React.PropsWithChildren<unknown>): JSX.Element {
-  const helpInfo = useRef<HelpInfo>({ title: "Some title", content: <>Some help content</> });
```
jtbandes, line 50: Looks like this is removing the dummy/fixture data. Why didn't that result in a change in storybook?

foxymiles: Looks like that content was never actually rendered in the story.

jtbandes: Is the story providing any value if it doesn't actually show the help info? Maybe we should improve or delete it?

foxymiles: It is showing other content, just not this initial value in a mock provider which we don't need.

jtbandes (approved):

PR #3332, opened 2022-05-06
Fix Image panel error when transformMarkers is enabled

jtbandes:
**User-Facing Changes**
Fixes an Image panel error when the "transform markers" setting is enabled.

**Description**
Fixes this error:

### DIFF
--- a/packages/studio-base/src/panels/ImageView/index.tsx
+++ b/packages/studio-base/src/panels/ImageView/index.tsx
@@ -344,7 +344,8 @@ function ImageView(props: Props) {
     return {
       markers: annotations ?? [],
       transformMarkers,
-      cameraInfo,
+      // Convert to plain object before sending to web worker
+      cameraInfo: (cameraInfo as { toJSON?: () => unknown } | undefined)?.toJSON?.() ?? cameraInfo,
     };
   }, [annotations, cameraInfo, transformMarkers]);
 


packages/studio-base/src/panels/ImageView/index.tsx
```diff
@@ -344,7 +345,9 @@ function ImageView(props: Props) {
     return {
       markers: annotations ?? [],
       transformMarkers,
-      cameraInfo,
+      // Convert to plain object before sending to web worker
+      cameraInfo:
```
defunctzombie, line 349: cameraInfo comes from `useCameraInfo`. If we want to ensure it is always expanded I think we should do that in the hook rather than having these casts?

Another option is a function like `serializeForWorker` that does the casting and `toObject` call. This should use `toObject` I think rather than `toJSON` if thats available in the version of rosmsg-serialization we are using.

defunctzombie (approved):
Left some thoughts - somehow feels fragile so maybe a function to prepare for worker is better. Certainly the right move to fix this.

packages/studio-base/src/panels/ImageView/index.tsx
jhurliman: We have to do this unpacking in a few different places in the codebase, some unrelated to web workers. A generic method that handles the `toObject` vs `toJSON` confusion would be useful.

jtbandes: > cameraInfo comes from `useCameraInfo`. If we want to ensure it is always expanded I think we should do that in the hook rather than having these casts?

Not every user of useCameraInfo may care about declassifying the object...do you still feel like it's best to put it in the hook anyway?

defunctzombie: > Not every user of useCameraInfo may care about declassifying the object...do you still feel like it's best to put it in the hook anyway?

Probably? It feels like if you are using camera info then we might as well get you the fields from it? Alternatively, a generic unpack to POD could be useful. This comes up with protobuf as well since those messages are protobuf class instances.

jtbandes:
Merging for now, will follow up with cleanup later.

PR #3334, opened 2022-05-06
[New3D] Refactor settings code organization, enable visibility toggling for all rendered topics

jhurliman:
**User-Facing Changes**

None

**Description**

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
+++ b/packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
@@ -176,7 +176,6 @@ export function ThreeDeeRender({ context }: { context: PanelExtensionContext }):
 
     return {
       cameraState,
-      enableStats: partialConfig?.enableStats ?? true,
       followTf: partialConfig?.followTf,
       scene: partialConfig?.scene ?? {},
       topics: partialConfig?.topics ?? {},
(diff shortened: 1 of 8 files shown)


packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
```diff
@@ -297,6 +302,13 @@ export function ThreeDeeRender({ context }: { context: PanelExtensionContext }):
...
 
+  // Draw a new frame when followTf changes
+  useEffect(() => {
+    if (renderer && followTf != undefined) {
+      renderer.animationFrame();
+    }
+  }, [followTf, renderer]);
+
   // Save panel settings whenever they change
   const throttledSave = useDebouncedCallback(
```
foxymiles, line 313: I think we no longer need to debounce this save to config.

foxymiles (approved):

packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
jhurliman: I tried just now with and without the debounce, and qualitatively it feels smoother (open the new 3D panel settings sidebar, open the Camera section, and quickly orbit the camera to the left and right). I'll leave it for now and we can revisit this.

PR #3335, opened 2022-05-06
Allow panels to control node collapse state with new expanded attribute

foxymiles:
**User-Facing Changes**
None

**Description**
This allows panel authors to optionally manually control the expansion state of each node by setting the new `expanded` attribute to either true or false. Omitting this field delegates control of the node's expansion state to the settings editor. Panels that set this to a non-undefined value will receive update actions with a path ending in `expanded`.


jhurliman:
Merge latest `main` in, I think there was a semantic conflict.

foxymiles:
Closing this for now until the exact use case is more clear.

PR #3340, opened 2022-05-09
Fix image panel settings error

foxymiles:
**User-Facing Changes**
None

**Description**
This fixes an issue with the image panel settings. The issue is that when settings are changed in the new settings UI using immer the resulting config is a frozen object. The `ImageEmptyState` was using an in-place sort on the resulting topics and blowing up:

https://github.com/foxglove/studio/blob/6e2c8ca905e582d14dfeb95dd351bfc94698ec3d/packages/studio-base/src/panels/ImageView/components/ImageEmptyState.tsx#L54-L58

TS didn't catch this because the typings didn't reflect the frozen state. We should probably more rigorously define incoming configs to panels as `DeepReadonly`.

### DIFF
--- a/packages/studio-base/src/panels/ImageView/components/ImageEmptyState.tsx
+++ b/packages/studio-base/src/panels/ImageView/components/ImageEmptyState.tsx
@@ -17,7 +17,7 @@ import EmptyState from "@foxglove/studio-base/components/EmptyState";
 
 type Props = {
   cameraTopic: string;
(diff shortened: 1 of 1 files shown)


defunctzombie (approved):
fun learning I had with this - spread operator is much faster than `new Array(oldArray)` and seems to be the same speed as `Array.from(oldArray)`. JAVASCRIPT

PR #3342, opened 2022-05-09
Put image panel topic and annotation settings in settings sidebar

foxymiles:
**User-Facing Changes**
This puts the topic and maker selection UI for the image panel into the settings sidebar. For convenience the topic picker remains in the panel toolbar.

**Description**
This makes both the topic and marker selections available in the settings sidebar. Markers are controlled via node-level visibility toggles.

Screen Shot 2022-05-18 at 7 29 14 AM


defunctzombie:
I like the visible property version (with the eye). The one gotcha we need to think through is user-education around leveraging the new sidebar for things like this. We might need to add an icon to the panel toolbar to guide users.

2metres:
Can you restore the hide/show behavior on the visibility toggle?

2metres:
Can we add a visibility toggle on the Marker Node to hide/show all nodes?

2metres (approved):

PR #3343, opened 2022-05-10
[WIP] Fix for memory exhaustion issue

### LINKED ISSUE #3318: Foxglove Studio (1.9.1) Crashes - "Array buffer allocation failed"
When I load my ROS bag (duration ~6 min, ~5GB) into Foxglove Studio (version 1.9.1) I reproducibly get an "Array buffer allocation failed" when plotting a signal in a single plot widget. Sometimes the whole GUI just freezes without even showing the error. The signal I want to plot has a frequency of ~25 Hz (>9000 messages).

Is that a known issue? Any hint on how this can be traced down further?

foxymiles:
**User-Facing Changes**
None

**Description**
It looks like our lazy message strategy is biting us here. These individual messages are only 321 bytes but their DataView retains the entire 60mb buffer that we decompress from the bag and this exhausts memory as we fill up blocks.

Trimming the underlying DataView to only contain the data needed seems to fix the problem but I'm not sure this is the right place to fix this.

Screen Shot 2022-05-10 at 7 21 27 AM

Fixes #3318

### DIFF
--- a/packages/studio-base/src/players/IterablePlayer/BagIterableSource.ts
+++ b/packages/studio-base/src/players/IterablePlayer/BagIterableSource.ts
@@ -31,6 +31,25 @@ import {
 
 type BagSource = { type: "file"; file: File } | { type: "remote"; url: string };
 
+function messageHasLazyDataView(message: unknown): message is { _view: DataView } {
+  return message != undefined && typeof message === "object" && "_view" in message;
+}
+
+/**
+ * Trims the underlying DataView in a lazy message to avoid retaining large
+ * array buffers referenced by the DataView.
+ */
+function trimMessageDataView(message: unknown) {
+  if (messageHasLazyDataView(message)) {
+    // eslint-disable-next-line no-underscore-dangle
+    const dataView = message._view;
+    // eslint-disable-next-line no-underscore-dangle
+    message._view = new DataView(
+      dataView.buffer.slice(dataView.byteOffset, dataView.byteOffset + dataView.byteLength),
+    );
+  }
+}
+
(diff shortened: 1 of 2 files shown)


defunctzombie:
> These individual messages are only 321 bytes but their DataView retains the entire 60mb buffer that we decompress from the bag and this exhausts memory as we fill up blocks.

I think some idea here was to re-use underlying arrays rather than making copies but it seems like this is somehow backfiring. Since we make new iterators for block loading and for ticks, all these messages are referencing new 60mb arrays (presumably from the new chunks constantly loaded?).

I don't think this is the place to make this fix - but this does make me wonder if the individual raw `MessageEvent` records we are getting from our bag reading code (the iterators) is holding on to the underlying chunk arrays rather than individual message arrays. Maybe that code needs to be copying the message bytes into a smaller Uint8Array rather than holding onto the data from the larger underlying chunk. It does mean we incur another copy to grab the message bytes from the chunk rather than making a view onto the decompressed chunk bytes.

foxymiles:
I guess one advantage of making the fix in the player is that this is only likely to be a problem for blocks where we hold onto messages for a long time right? Otherwise we'll need to apply this fix in several places in our lower-level message reading code.

foxymiles:
It looks like the problem doesn't exist on the old bag player. I guess because of different read patterns?

foxymiles:
@defunctzombie I've moved the buffer trim code to `BagIterableSource`. If you think this is a better approach I can add it to the MCAP source too.

foxymiles:
Actually testing this on the original problematic bag converted to MCAP I don't see the same issue.

Issue #3055, opened 2022-03-22
ContextMenu panel API


2metres:
Introduce global ContextMenu that can have items pushed to it by panels/components

jtbandes:
Are we talking about context menus as in right-click detail menus? What is the motivation for a global context/provider for this?

PR #3347, opened 2022-05-11
Refactor ImagePanel zoom controls into a toolbar menu instead of floating button

2metres:
**User-Facing Changes**
Move the zoom control in the image panel into a button in the toolbar with a contextual menu instead of floating icon.

Screen Shot 2022-05-11 at 11 44 58 am

**Description**


2metres:
@foxymiles But could use help wiring it all up. It looks like most the ImagePanel zoom state is handled internally within the ImageCanvas component from what I can see.

defunctzombie:
Feels like the icon area is getting crowded - maybe we need a vertical divider between this control and the icons?

2metres:
@foxymiles Should the percentage update for fit and fill?

foxymiles:
> @foxymiles Should the percentage update for fit and fill?

The way the underlying lib we're using works is that zoom = 1 in both cases but the zoom mode is changed.

2metres:
@foxymiles the Image Canvas continues stops zooming out at 50% can we reflect that limit in the dropdown by disabling the Zoom out button at 50%

foxymiles (approved):

2metres:
I could use some help getting the fit and fill zoom level displaying the percentage correctly

foxymiles:
> I could use some help getting the fit and fill zoom level displaying the percentage correctly

This is actually not that easy because the "zoom to fit" & "zoom to fill" modes need to know the dimensions of the decoded bitmap in order to work so we'd need to somehow communicate that back from the worker to the UI.

amacneil:
Follow up from meeting today:

As is, I'm 👎 on merging this. I think it's not a clear win - removing the floating controls is nice, but taking up precious space from the title bar is not nice. For that reason, I'm not convinced the title bar is a good long term home for zoom controls (e.g. maybe context menu is a better place for this?), and we should err on the side of not changing features until we are sure they won't change again for a long time.

2metres:
Closing PR as team does not agree on wether we should use this approach

amacneil:
Just to be clear, I do think we should revisit this in the future, because the floating zoom controls are sub-optimal. However, I think we should consider other options besides using the panel titlebar. 

Let's discuss it again after we complete https://github.com/foxglove/studio/issues/3055 and/or when we implement the new 3d-based image panel.

PR #3351, opened 2022-05-11
Refactor panels to always show toolbar

2metres:
**User-Facing Changes**
Refactor all panels to have persistent toolbar of the same height. Reduce UI interaction flicker that happens when using panels and solve issue with menu losing reference to thier ref element.


2metres:
Outstanding issues
- [ ] When clicking Map Panel map covers toolbar
- [ ] TablePanel Small grey space to the right of toolbar
- [ ] DiagnosticSummary Database icon has menu issues

defunctzombie:
Can we break this up into smaller bits so we can tackle the blocking items without blocking other aspects of this merging?

defunctzombie:
We also need to tie this back to the larger roadmap item for context on why we are doing this. The description does not surface or link to where this is going and the discussions we've had around it which I think are important context.

packages/studio-base/src/components/PanelToolbar/ToolbarIconButton.tsx
```diff
@@ -0,0 +1,36 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { IconButton, IconButtonProps, styled as muiStyled } from "@mui/material";
+
+const StyledIconButton = muiStyled(IconButton)<{ subMenuActive: boolean }>(
```
foxymiles, line 7: I think we need a `shouldForwardProp` here.

packages/studio-base/src/panels/Log/FilterBar.tsx
```diff
@@ -107,10 +101,43 @@ export default function FilterBar(props: FilterBarProps): JSX.Element {
...
+          height: 22,
+          minHeight: 22,
+        },
+        dropdownItem: {
+          lineHeight: 22,
+          height: 22,
+          minHeight: 22,
+          fontSize: theme.fonts.small.fontSize,
+        },
+      } as Partial<IDropdownStyles>),
```
foxymiles, line 134: If we do this as:

```
const dropdownStyles: Partial<IDropdownStyles>
```

I think that allows TS to still catching typing errors instead of doing a hard cast.

amacneil:
If we show the panel toolbar by default, we should show the panel toolbar icons all the time too (instead of only on hover)

foxymiles (approved):

PR #3352, opened 2022-05-11
[New3D] Render PoseStamped and PoseWithCovarianceStamped

jhurliman:
**User-Facing Changes**

- None

**Description**

- Refactor standardMaterial() to take a ColorRGBA instead of a Marker
- Fix double-add of TF topics to subscription list
- Render PoseStamped and PoseWithCovarianceStamped

Screenshot from 2022-05-11 16-35-46

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/Poses.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/Poses.ts
@@ -1,0 +1,274 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import * as THREE from "three";
+
+import { SettingsTreeFields } from "@foxglove/studio-base/components/SettingsTreeEditor/types";
+
+import { Renderer } from "../Renderer";
+import { rgbaToCssString, stringToRgba } from "../color";
+import {
+  Pose,
+  rosTimeToNanoSec,
+  ColorRGBA,
+  Marker,
+  PoseWithCovarianceStamped,
+  PoseStamped,
+  POSE_WITH_COVARIANCE_STAMPED_DATATYPES,
+  MarkerAction,
+  MarkerType,
+} from "../ros";
+import { LayerSettingsPose, LayerType } from "../settings";
+import { makePose } from "../transforms/geometry";
+import { updatePose } from "../updatePose";
+import { RenderableArrow } from "./markers/RenderableArrow";
+import { RenderableSphere } from "./markers/RenderableSphere";
+import { missingTransformMessage, MISSING_TRANSFORM } from "./transforms";
+
+const DEFAULT_SCALE: THREE.Vector3Tuple = [1, 0.15, 0.15];
+const DEFAULT_COLOR = { r: 124 / 255, g: 107 / 255, b: 1, a: 1 };
+const DEFAULT_COVARIANCE_COLOR = { r: 198 / 255, g: 107 / 255, b: 1, a: 0.25 };
+
+const DEFAULT_COLOR_STR = rgbaToCssString(DEFAULT_COLOR);
+const DEFAULT_COVARIANCE_COLOR_STR = rgbaToCssString(DEFAULT_COVARIANCE_COLOR);
+
+const DEFAULT_SETTINGS: LayerSettingsPose = {
+  visible: true,
+  scale: DEFAULT_SCALE,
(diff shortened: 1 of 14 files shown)


defunctzombie:
@jhurliman Are there stories for all these renderables?

jhurliman:
@defunctzombie yes, that is the screenshot you're seeing in the PR description.

packages/studio-base/src/panels/ThreeDeeRender/renderables/Poses.ts
```diff
@@ -0,0 +1,274 @@
...
+    this.children.length = 0;
+    this.posesByTopic.clear();
+  }
+
+  addPoseMessage(topic: string, poseMessage: PoseStamped | PoseWithCovarianceStamped): void {
+    let renderable = this.posesByTopic.get(topic);
+    if (!renderable) {
+      renderable = new THREE.Object3D() as PoseRenderable;
+      renderable.name = topic;
+      renderable.userData.topic = topic;
```
jtbandes, line 104: is userData automatically initialized to an empty object? I think it might help readability to change this to `renderable.userData = { all the properties initialized below }`

packages/studio-base/src/panels/ThreeDeeRender/renderables/Poses.ts
```diff
@@ -0,0 +1,274 @@
...
+  // We currently only visualize position variance so extract the upper-left
+  // 3x3 matrix
+  // [X, -, -, -, -, -]
+  // [-, Y, -, -, -, -]
+  // [-, -, Z, -, -, -]
+  // [-, -, -, -, -, -]
+  // [-, -, -, -, -, -]
+  // [-, -, -, -, -, -]
+  const K = poseMessage.pose.covariance;
+  const scale = { x: Math.sqrt(K[0]), y: Math.sqrt(K[7]), z: Math.sqrt(K[14]) };
```
jtbandes, line 255: I suppose this only works if covariance is actually diagonal?

packages/studio-base/src/panels/ThreeDeeRender/renderables/Poses.ts
```diff
@@ -0,0 +1,274 @@
...
+  ): void {
+    const arrowMarker = createArrowMarker(poseMessage, renderable.userData.settings);
+    renderable.userData.arrow.update(arrowMarker);
+
+    if ("covariance" in poseMessage.pose) {
+      const poseWithCovariance = poseMessage as PoseWithCovarianceStamped;
+      const sphereMarker = createSphereMarker(poseWithCovariance, renderable.userData.settings);
+      if (sphereMarker) {
+        if (!renderable.userData.sphere) {
+          renderable.userData.sphere = new RenderableSphere(
```
jtbandes, line 199: I'm having trouble following the flow that happens in the update case ... are we creating a new sphere instance each time we get a new PoseWithCovariance message? Is there a way to reuse this / do we need to dispose the old sphere?

jtbandes (approved):

packages/studio-base/src/panels/ThreeDeeRender/renderables/Poses.ts
jhurliman: Yes, `userData` is set to `{}` by threejs. I can do a refactor pass for how userData is initialized in the renderables.

packages/studio-base/src/panels/ThreeDeeRender/renderables/Poses.ts
jhurliman: This will only render the standard deviation, `sqrt(variance)`, for each axis. The first three diagonal entries are the covariance of X and X, the covariance of Y and Y, and the covariance of Z and Z (wrt position). A more advanced approach would be to compute the eigendecomposition of the covariance matrix and use that to set both the scale and rotation of a spheroid.

packages/studio-base/src/panels/ThreeDeeRender/renderables/Poses.ts
jhurliman: No, the sphere instance is only created here if it has never been created before for this renderable. The only practical way to end up here is to publish the first message with a missing/invalid covariance matrix, followed by another matrix that has a valid matrix. Once the `RenderableSphere` has been created, it will only have its scale and color updated it won't create another one.

An alternative approach would be to unconditionally create the `RenderableSphere` and just toggle visibility based on whether we have a valid covariance matrix or not. That might simplify the code a bit, but I was building this under the assumption that the per-topic data is probably stable and you will either always have a covariance matrix or never have one.

PR #3356, opened 2022-05-12
Add custom map url config

r0gi:
**User-Facing Changes**
This extends the map panel configuration sidepanel to include a custom map layer which points to a custom map tile URL.
Screenshot from 2022-05-13 11-45-25

Users can now leverage custom URLs that may provide better imagery. For example, using a custom URL on the left:
Screenshot from 2022-05-13 11-54-01

**Description**
I had to add some basic validation to ensure the URL placeholders had the correct format (e.g. `{x}`), otherwise the leaflet js library would throw errors.

I'm not sure what the best way to handle image attribution is with custom map tile urls.

### DIFF
--- a/packages/studio-base/src/panels/Map/MapPanel.tsx
+++ b/packages/studio-base/src/panels/Map/MapPanel.tsx
@@ -150,11 +150,10 @@ function MapPanel(props: MapPanelProps): JSX.Element {
 
   const [customLayer] = useState(
     new TileLayer("https://example.com/{z}/{y}/{x}", {
-        attribution: "",
-        maxNativeZoom: 20,
-        maxZoom: 24,
-      },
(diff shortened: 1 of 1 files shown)


defunctzombie:
> I'm not sure what the best way to handle image attribution is with custom map tile urls.

I don't think custom urls need attribution from Studio's point of view since we aren't bundling those with our app and it's something you are hosting and configuring yourself. It's on you to attribute if you make a screenshot or something that maybe requires attribution.

defunctzombie:
@r0gi Thanks for making this PR. I've made a few tweaks to this in a new PR here (https://github.com/foxglove/studio/pull/3364). You can see what changed in the commit history.

Closing this PR and we will review and merge #3364.

PR #3360, opened 2022-05-13
Add new message expansion options in the raw messages panel

### LINKED ISSUE #3303: Preserve raw message expansion when switching tabs
If a raw messages panel is inside of a tab panel, and you switch away from that tab, the state of the raw messages panel is not preserved.

You can see this by:
1. Adding a tab panel to your layout
2. Adding a raw message panel in a tab
3. Expanding the message so that the whole header (or other piece of the message) is displayed
4. Click to another tab, and then click back to the raw message panel. The message will no longer be expanded.

Since messages are not expanded by default, this means that one must expand the message every time. I like to create layouts with raw message panels for several topics in a tab panel, but this makes switching between them cost more time.

foxymiles:
**User-Facing Changes**
This adds new auto-expansion behavior to the raw messages panel, giving users more control over when message fields are expanded.

**Description**
This adds a new "auto expand" setting to the raw messages panel with three modes:

1. Off - the current behavior
2. On - all message fields are automatically expanded
3. Auto - message fields under a certain size are expanded (currently this limits arrays to < 5 members)

Screen Shot 2022-05-13 at 10 39 09 AM

Fixes #3303

### DIFF
--- a/packages/studio-base/src/panels/RawMessages/index.tsx
+++ b/packages/studio-base/src/panels/RawMessages/index.tsx
@@ -19,6 +19,7 @@ import LessIcon from "@mdi/svg/svg/unfold-less-horizontal.svg";
 import MoreIcon from "@mdi/svg/svg/unfold-more-horizontal.svg";
 import { Theme } from "@mui/material";
 import { makeStyles } from "@mui/styles";
+import { Immutable } from "immer";
 // eslint-disable-next-line no-restricted-imports
 import { first, isEqual, get, last } from "lodash";
 import { useState, useCallback, useMemo, useEffect } from "react";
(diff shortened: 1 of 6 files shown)


packages/studio-base/src/panels/RawMessages/index.tsx
```diff
@@ -67,20 +70,24 @@ import {
...
 } from "./getValueActionForValue";
 import helpContent from "./index.help.md";
-import { DATA_ARRAY_PREVIEW_LIMIT, getItemStringForDiff } from "./utils";
+import { buildSettingsTree } from "./settings";
+import { DATA_ARRAY_PREVIEW_LIMIT, generateDeepKeyPaths, getItemStringForDiff } from "./utils";
 
 export const CUSTOM_METHOD = "custom";
 export const PREV_MSG_METHOD = "previous message";
+
+type AutoExpandMode = "auto" | "off" | "all";
```
defunctzombie, line 79: Maybe should live in settings.ts? Not sure what you've been doing for types related to settings.

defunctzombie (approved):

packages/studio-base/src/panels/RawMessages/index.tsx
foxymiles: I like the approach of having a `types.ts` in each panel's folder. Might be a good thing here too.

PR #3362, opened 2022-05-13
Add backfill messages to McapIterableSource

defunctzombie:
**User-Facing Changes**
None. This player is behind a feature flag that is off by default.

**Description**
When seeking using the "Experimental Mcap player" the last message on each topic is loaded. This matches the behavior of the experimental bag player and experimental data platform player

Fixes: #3361


defunctzombie:
Blocked by: https://github.com/foxglove/mcap/pull/391

defunctzombie:
@jtbandes This is ready for review. One downside of the approach taken here (and in the bag iterable source) is that each topic is a separate reverse read request which means a chunk might be deserialized and de-compressed multiple times if we could have gotten the messages from the same chunk. I think it's a future optimization to consider adding a single function to the mcap and bag readers to perform this last message lookup behavior. Then each reader can optimize this to load only the necessarily chunks once.

packages/studio-base/src/players/IterablePlayer/Mcap/McapIndexedIterableSource.ts
```diff
@@ -149,7 +152,40 @@ export class McapIndexedIterableSource implements IIterableSource {
...
-  async getBackfillMessages(_args: GetBackfillMessagesArgs): Promise<MessageEvent<unknown>[]> {
-    return [];
+  async getBackfillMessages(args: GetBackfillMessagesArgs): Promise<MessageEvent<unknown>[]> {
+    const { topics, time } = args;
+
+    const messages: MessageEvent<unknown>[] = [];
+    for (const topic of topics) {
+      // NOTE: An iterator is made for each topic to get the latest message on that topic.
+      // An single iterator for all the topics could result in iterating through many
+      // irrelevant messages to get to an older message on a topic.
```
jtbandes, line 162: Just wondering whether it would help at all to use `Promise.all()` rather than an outer for loop. Maybe it won't really improve performance at all? 🤷

packages/studio-base/src/players/IterablePlayer/Mcap/McapIndexedIterableSource.ts
```diff
@@ -149,7 +152,40 @@ export class McapIndexedIterableSource implements IIterableSource {
...
+      for await (const message of this.reader.readMessages({
+        endTime: toNanoSec(time),
+        topics: [topic],
+        reverse: true,
+      })) {
+        const channelInfo = this.channelInfoById.get(message.channelId);
+        if (!channelInfo) {
+          continue;
```
jtbandes, line 170: Should we log/throw an error here? When would this ever happen unless the file is corrupt?

jtbandes (approved):

packages/studio-base/src/players/IterablePlayer/Mcap/McapIndexedIterableSource.ts
defunctzombie: Will think about it as a future enhancement. Overall I still think it better to push this down a layer if we can.

PR #3363, opened 2022-05-13
Use zustand for hover value context

foxymiles:
**User-Facing Changes**
None

**Description**
This replaces the use of our custom selectable context with zustand in our hover value context.

### DIFF
--- a/packages/studio-base/src/context/HoverValueContext.tsx
+++ b/packages/studio-base/src/context/HoverValueContext.tsx
@@ -26,7 +26,7 @@ type HoverValueStore = Readonly<{
 
 const { Provider, useStore } = createContext<StoreApi<HoverValueStore>>();
 
-export function createHoverValueStore(): StoreApi<HoverValueStore> {
+function createHoverValueStore(): StoreApi<HoverValueStore> {
   return create((set) => {
     return {
       value: undefined,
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/context/HoverValueContext.tsx
```diff
@@ -12,34 +12,44 @@
...
-const Context = createSelectableContext<HoverValueContext>();
-Context.displayName = "HoverValueContext";
+const { Provider, useStore } = createContext<StoreApi<HoverValueStore>>();
 
-export function useClearHoverValue(): HoverValueContext["clearHoverValue"] {
-  return useContextSelector(
-    Context,
-    useCallback((ctx) => ctx.clearHoverValue, []),
-  );
+export function createHoverValueStore(): StoreApi<HoverValueStore> {
```
defunctzombie, line 29: Does this function need to be exported?

defunctzombie (approved):
👍 If we continue to use this library maybe we want to separate the stores and creation from the Context wrapper code. We'll see how it plays out.

packages/studio-base/src/context/HoverValueContext.tsx
foxymiles: No I guess not. I'll fix.

foxymiles:
> 👍 If we continue to use this library maybe we want to separate the stores and creation from the Context wrapper code. We'll see how it plays out.

In the typical case the `Provider` isn't very interesting so maybe there's not much benefit in breaking them up?

PR #3364, opened 2022-05-13
Add `Custom` map tile layer option to map panel

defunctzombie:
**User-Facing Changes**
* The `Layer` setting in Map panel is renamed to `Tile Layer`.
* A new `Custom` option is added to the Tile Layer selection. When the user selects `Custom` they can enter their own URL for tile layers.

Screen Shot 2022-05-13 at 11 56 20 AM

Screen Shot 2022-05-13 at 11 56 48 AM

Using a custom URL on the left compared to the default `Map` on the right:
Screenshot from 2022-05-13 11-54-01

Thank you @r0gi for the initial implementation in https://github.com/foxglove/studio/pull/3356

### DIFF
--- a/packages/studio-base/src/components/Panel.tsx
+++ b/packages/studio-base/src/components/Panel.tsx
@@ -48,7 +48,7 @@ import Icon from "@foxglove/studio-base/components/Icon";
 import KeyListener from "@foxglove/studio-base/components/KeyListener";
 import PanelContext from "@foxglove/studio-base/components/PanelContext";
 import PanelErrorBoundary from "@foxglove/studio-base/components/PanelErrorBoundary";
-import { PanelRoot, PANEL_ROOT_SELECTOR } from "@foxglove/studio-base/components/PanelRoot";
+import { PanelRoot } from "@foxglove/studio-base/components/PanelRoot";
 import {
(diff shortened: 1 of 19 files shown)


foxymiles (approved):

r0gi:
Thanks. The updates look good except for one small issue. With the custom map layer, I modified the TileLayer's `maxNativeZoom` [prop](https://leafletjs.com/reference.html#gridlayer-maxnativezoom) to `20`, as that seems to be the [maximum maptiles zoom level](https://wiki.openstreetmap.org/wiki/Zoom_levels). This got changed back in the tweaks commit to `18`, which makes my custom map lower in resolution than it could be. 

@defunctzombie can we change this back to `20`? Most maptile servers probably don't have imagery at that zoom level however, so we could make it another configuration setting?

defunctzombie:
@r0gi Yea - I see now with your link that 20 allows for a closer zoom level. I think our built-in tile sets couldn't support 20 so I had set it to 18. I do think it makes sense to set a min and max zoom level when using a custom tile set. We could default those to something sensible and allow the user to override.

Do you have time to tackle a PR for that based from the changes we've already landed in main? If not - let's make an issue so we remember the feature request.

PR #3368, opened 2022-05-13
Remove title attribute from internal LegacyButton within Button

defunctzombie:
**User-Facing Changes**
When hovering over some buttons that have tooltips, the app properly displays only a single tooltip rather than the tooltip and a button title.

Before:
image

After:
image

**Description**
The Button already creates a tooltip from the tooltip parameter. Passing a title prop to the LegacyButton creates a second popup with the same information. This removes the title property from the LegacyButton component within Button so only one tooltip is shown.

### DIFF
--- a/packages/studio-base/src/components/Button.tsx
+++ b/packages/studio-base/src/components/Button.tsx
@@ -118,7 +118,6 @@ export default function Button({
       onFocus={onFocus}
       {...eventHandlers}
       style={{ position: "relative", ...styles }}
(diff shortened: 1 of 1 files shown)


foxymiles (approved):

defunctzombie:
@2metres I know how much you hate tooltips - this one was already in place. If you think it makes more sense to remove tooltip and go with title only we can do that. Going to merge this to get the fix in tho.

Issue #3371, opened 2022-05-15
Add SubscriptionPreload options to extension subscriber api


defunctzombie:
The current subscribe api behavior for extensions uses the `full` preload option for all subscriptions (https://github.com/foxglove/studio/blob/main/packages/studio-base/src/players/types.ts#L270) since that is the default if nothing is specified (for backwards compat). This produces undesirable over-fetching of messages for panels which only need the current message (i.e. 3d panel).

Expose subscriber options to the subscribe api to allow extensions to select the preload mode for their topics.

PR #3372, opened 2022-05-15
Add subscriber options to extension api

defunctzombie:
**User-Facing Changes**
When subscribing to topics via the `subscribe` extension api the user can specify whether they would like the topic to preload data or not. There are two ways to subscribe to topics:

Using topic names will set the current subscribed topics to this list of topics (and assumes preloading for these topics). This is the existing behavior of the api.
```typescript
context.subscribe(['topic_a', 'topic_b']);
```

The new api allows a user to indicate whether they want to preload the topic:
```typescript
context.subscribe([{ topic: 'topic_a', preload: true }, { topic: 'topic_b' }]);
```

Now `topic_a` will preload and appear in `RenderState#allFrames` while `topic_b` will not. Both will appear in `RenderState#currentFrame`.

**Description**
Rather than assuming all subscriptions from the extension api should preload topics, this allows the extension author to decide if their panel should subscribe to a topic with preloading or not.

Preloading topics results in more data transfer and memory use and is recommended only for panels which display the entire dataset (i.e. plot, map are example panels) rather than displaying only the current frame (image, raw message, 3d, etc).

Fixes: #3371

---
I'd like to hear some alternative ideas for what to call the "preload" option. Internally we have `preload: "full" | "partial"` but I don't think `partial` is quite right. In this PR I've made `preload` a boolean but am not convinced of the term `preload`. To add to the confusion, the messages that we "preload" appear in the `allFrames` render state variable so there's not consistency there either.

I plan to deprecate the string array form for subscriptions because it defaults to preloading (backwards compatibility) and prefer use of the object array form that allows for subscriber options since I believe that most panels don't want preloading. Both forms _set_ the active list of topics; to unsubscribe from a topic call subscribe with an array that does not contain that topic. Probably the method could have been called `setSubscriptions` or something better to indicate that behavior.

Another option would be to see if the panel has called `context.watch('allFrames')` and change the string array form of subscribe to preload topics only in that case - but that has gotchas with when the user might watch allFrames.


defunctzombie:
@jhurliman I've added a fixme in the ThreeDeeRender where it subscribes so we could use that as a discussion point. My plan is to update that subscribe logic so ThreeDeeRender doesn't force preloading of the topics a user enables since those tend to be heavy-weight topics.

Map panel is another panel we can look to for how it is managing subscriptions to topics.

amacneil:
I don’t think we need to solve this immediately, but it’s also worth thinking about how the API might evolve in the future if we wanted to support server side downsampling. E.g. plot panel could request preload of all data but specify sampling rate based on how many pixels are available to render.

packages/studio-base/src/components/PanelExtensionAdapter.tsx
```diff
@@ -414,8 +415,13 @@ function PanelExtensionAdapter(props: PanelExtensionAdapterProps): JSX.Element {
         watchedFieldsRef.current.add(field);
       },
 
-      subscribe: (topics: string[]) => {
-        subscribedTopicsRef.current.clear();
+      subscribe: (topics: string | string[], options?: SubscribeOptions) => {
```
jhurliman, line 418: Can we make SubscribeOptions per-topic? With this API, it doesn't seem possible to preload one topic and do a normal subscription to another.

defunctzombie: Changed the api as discussed.

defunctzombie:
@foxymiles @jhurliman @jtbandes this is ready for review. I've updated the PR description so please re-read that since items have changed.

@jhurliman This PR contains a change to the new3d panel to leverage this new API. If you'd rather I make that a separate follow-on PR I am happy to do so. I wanted to make sure the interface worked as expected for the 3d panel using this PR and I believe it does.

defunctzombie:
> Should we support panels explicitly requesting to preload a custom time range?
> 
> For example, if I am rendering a plot, but the plot viewport does not span the entire file, it would be beneficial to request only the time range necessary to render the current viewport.

I'm leaning towards no. I can definitely see how this _could_ be a thing but it also seems tricky to reason about when you start to think about panning/zooming the plot.

With the new players I've written we no longer load all subscribed topics into blocks (the old players would do this and use more memory). Since the new players don't load topics which the user doesn't want preloaded, that leaves more memory space for loading lightweight topics for plotting on the plot panel.

I think we will need to see how our data usage evolves and extension api use evolves to really answer this question. One nice property of the API here is that if we do want to add more options to a topic subscription we are able to do so.

packages/studio-base/src/components/PanelExtensionAdapter.tsx
```diff
@@ -103,7 +105,10 @@ function PanelExtensionAdapter(props: PanelExtensionAdapterProps): JSX.Element {
 
   const [error, setError] = useState<Error | undefined>();
   const watchedFieldsRef = useRef(new Set<keyof RenderState>());
-  const subscribedTopicsRef = useRef(new Set<string>());
+  // When subscribing to preloaded topics we use this array to filter the raw blocks to include only
+  // the topics we subscribed to in the allFrames render state. Otherwise the panel would receive
+  // messages in allFrames for topics the panel did not subscribe to.
+  const subscribedTopicsRef = useRef<string[]>([]);
```
defunctzombie, line 111: I've used an array here because we've learned the truth about arrays vs sets for small-ish sized datasets. I cry.

packages/studio-base/src/components/PanelExtensionAdapter.tsx
```diff
@@ -414,20 +419,30 @@ function PanelExtensionAdapter(props: PanelExtensionAdapterProps): JSX.Element {
...
         // created, we unset the blocks ref which will force re-creating allFrames.
         prevBlocksRef.current = undefined;
 
-        const subscribePayloads = topics.map((topic) => ({ topic }));
+        const subscribePayloads = topics.map<SubscribePayload>((item) => {
+          if (typeof item === "string") {
+            subscribedTopicsRef.current.push(item);
+            // For backwards compatability with the topic-string-array api `subscribe(["/topic"])`
+            // results in a topic subscription with full preloading
+            return { topic: item, preloadType: "full" };
```
defunctzombie, line 436: I could make this smarter by only using `full` when `context.watch('allFrames')` was called.

jhurliman (approved):

PR #3373, opened 2022-05-16
Display an error icon and message tooltip for settings fields with errors

foxymiles:
**User-Facing Changes**
This allows the settings UI to indicate field-level errors to the user

**Description**
This adds a new `error` attribute at the field level in the settings tree. In the UI this is represented as a red error icon with a hover tooltip containing the error message as well as a red outline on the field input.

Screen Shot 2022-05-16 at 9 10 52 AM

Screen Shot 2022-05-16 at 9 11 02 AM


foxymiles:
This works on all input types not just text fields since we apply the error at the level of the `FieldEditor` wrapper component.

The MUI error state wouldn't work on our custom input types without some extra work but we could certainly display errors in that style if we're willing to incur the layout shifts to make room for the messages inline.

PR #3382, opened 2022-05-17
Fix react uncontrolled component warnings

foxymiles:
**User-Facing Changes**
None

**Description**
This fixes the uncontrolled component warnings we're getting from React on some settings inputs by rendering an empty string instead of `undefined` for the number & textfield inputs. This makes these always "controlled" components in React's semantics.

It also sets a react `key` equal to the selected panel id on the `SettingsEditor` to avoid any inadvertent UI sharing between different instances of settings trees for different instances of the same panel type.

```
Warning: A component is changing a controlled input to be uncontrolled. This is likely caused by the value changing from a defined to undefined, which should not happen. Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://reactjs.org/link/controlled-components
    at input
    at eval (webpack-internal:///../../node_modules/@emotion/react/dist/emotion-element-699e6908.browser.esm.js:54:66)
    at div
    at eval (webpack-internal:///../../node_modules/@emotion/react/dist/emotion-element-699e6908.browser.esm.js:54:66)
    at InputBase (webpack-internal:///../../node_modules/@mui/material/InputBase/InputBase.js:240:83)
    at FilledInput (webpack-internal:///../../node_modules/@mui/material/FilledInput/FilledInput.js:182:82)
    at div
```

### DIFF
--- a/packages/studio-base/src/components/PanelSettings/index.tsx
+++ b/packages/studio-base/src/components/PanelSettings/index.tsx
@@ -162,7 +162,7 @@ export default function PanelSettings({
       {shareModal}
       <Stack gap={2} justifyContent="flex-start">
         <div>
-          {settingsTree && <SettingsEditor settings={settingsTree} />}
+          {settingsTree && <SettingsEditor key={selectedPanelId} settings={settingsTree} />}
           {!settingsTree && (
             <Typography color="text.secondary">No additional settings available.</Typography>
           )}
--- a/packages/studio-base/src/components/SettingsTreeEditor/FieldEditor.tsx
+++ b/packages/studio-base/src/components/SettingsTreeEditor/FieldEditor.tsx
@@ -159,7 +159,7 @@ function FieldInput({
           variant="filled"
           size="small"
           fullWidth
-          value={field.value}
+          value={field.value ?? ""}
           placeholder={field.placeholder}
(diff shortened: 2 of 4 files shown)


jtbandes:
> It also sets a react key equal to the selected panel id on the SettingsEditor to avoid any inadvertent UI sharing between different instances of settings trees for different instances of the same panel type.


Can you explain this more? My impression is that React `key`s are interpreted within the scope of the parent & sibling elements. In this case there is at most 1 sibling element (the element isn't immediately inside a .map() or array), and the default behavior of using the index as a key seems like it would be sufficient.

jtbandes (approved):
Controlled/uncontrolled changes look good, not convinced that the `key` change has any effect.

foxymiles:
> Controlled/uncontrolled changes look good, not convinced that the `key` change has any effect.

You can see the effect of the key change switching back between two instances of the same type of panel, although the specific effect is masked by fixing this uncontrolled input issue.

jtbandes:
Would there be any situations, other than incorrectly switching between controlled & uncontrolled mode, where remounting the tree would matter?

foxymiles:
None come specifically to mind but this seems like a reasonable hygiene guarantee.

PR #3387, opened 2022-05-17
[New3D] TRIANGLE_LIST Markers

jhurliman:
**User-Facing Changes**

None

**Description**

Adds support for rendering TRIANGLE_LIST markers (user-supplied triangle vertices and colors). No EdgesGeometry for now since this can be slow to compute for a long triangle list that is changing frequently.

This also changes the React flow in ThreeDeeRender to avoid redundant React updates and/or multiple calls to `renderer.animationFrame()` in a single update.


packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
```diff
@@ -415,8 +410,20 @@ export function ThreeDeeRender({ context }: { context: PanelExtensionContext }):
...
+    renderer?.setCameraState(cameraState);
+    renderRef.current.needsRender = true;
+  }, [cameraState, renderer]);
+
+  // Render a new frame if requested
+  if (renderer && renderRef.current.needsRender) {
     renderer.animationFrame();
-  }, [cameraState, messages, renderer, topicsToDatatypes]);
+    renderRef.current.needsRender = false;
+  }
```
jtbandes, line 426: This should probably move into the useEffect right before renderDone

packages/studio-base/src/panels/ThreeDeeRender/renderables/markers/RenderableTriangleList.ts
```diff
@@ -2,5 +2,139 @@
...
+    const prevMarker = this.userData.marker;
+    super.update(marker);
+
+    const vertexCount = marker.points.length;
+    const count = vertexCount * 3;
+    if (vertexCount === 0) {
+      this._renderer.layerErrors.addToTopic(
+        this.userData.topic,
+        NOT_DIVISIBLE_ERR,
+        `TRIANGLE_LIST: points[0] is empty`,
```
jtbandes, line 64:
```suggestion
        `TRIANGLE_LIST: points is empty`,
```

packages/studio-base/src/panels/ThreeDeeRender/renderables/markers/RenderableTriangleList.ts
```diff
@@ -2,5 +2,139 @@
...
+      );
+      this.geometry.setAttribute("position", new THREE.BufferAttribute(EMPTY_FLOAT32, 3));
+      return;
+    }
+    if (vertexCount % 3 !== 0) {
+      this._renderer.layerErrors.addToTopic(
+        this.userData.topic,
+        EMPTY_ERR,
+        `TRIANGLE_LIST: points[${vertexCount}] is not divisible by 3`,
```
jtbandes, line 73: Should this be just a warning and we render up to `floor(vertexCount / 3)` vertices?

packages/studio-base/src/panels/ThreeDeeRender/renderables/markers/RenderableTriangleList.ts
```diff
@@ -2,5 +2,139 @@
...
+      colors[i * 4 + 1] = tempColor.g;
+      colors[i * 4 + 2] = tempColor.b;
+      colors[i * 4 + 3] = tempColor.a;
+    }
+
+    if (dataChanged) {
+      this.geometry.setAttribute("position", new THREE.BufferAttribute(vertices, 3));
+      this.geometry.setAttribute("color", new THREE.BufferAttribute(colors, 4));
+      this.geometry.attributes.position!.needsUpdate = true;
+      this.geometry.attributes.color!.needsUpdate = true;
```
jtbandes, line 135: what does this do? In the docs I see needsUpdate on Material, but these aren't materials, right?

jtbandes (approved):
I imagine this is going to require some significant changes to support triangle strip & triangle fan, but that can come later

packages/studio-base/src/panels/ThreeDeeRender/renderables/markers/RenderableTriangleList.ts
jhurliman: These buffers are the CPU-based mirror of buffers that also exist on the GPU. This tells 3js to sync CPU->GPU. I think they could already be set by the previous method calls, but those kind of internal details seem to change in 3js from version to version.

jhurliman:
> I imagine this is going to require some significant changes to support triangle strip & triangle fan, but that can come later

My opinion is that we shouldn’t support these. 3js dropped support for them a while ago and just provides util methods to convert to triangle lists. There would be some bandwidth savings (not to the GPU, just disk or network) but compression would eliminate a lot of that savings.

packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
jhurliman: After making this change, camera input does not cause redraws while the player is paused.

PR #3388, opened 2022-05-17
Change Diagnostics panel names to "ROS Diagnostics"

jtbandes:
**User-Facing Changes**
Changed the names of "Diagnostics – Summary" and "Diagnostics – Detail" panels to reflect that they are ROS-specific.

**Description**
We don't intend to create Foxglove equivalent messages under @foxglove/message-schemas for these panels in their current form, because they are so ROS-specific. In the future, if there is user demand, we may consider a redesign of these panels so that they are more widely applicable.

Relates to https://github.com/foxglove/roadmap/issues/10

### DIFF
--- a/packages/studio-base/src/panels/index.ts
+++ b/packages/studio-base/src/panels/index.ts
@@ -56,7 +56,7 @@ const builtin: PanelInfo[] = [
     module: async () => await import("./ThreeDimensionalViz"),
   },
   {
-    title: `ROS Diagnostics – Detail`,
+    title: `Diagnostics – Detail (ROS)`,
     type: "DiagnosticStatusPanel",
     description: "Display ROS DiagnosticArray messages for a specific hardware_id.",
     help: DiagnosticStatusPanelHelp,
@@ -64,7 +64,7 @@ const builtin: PanelInfo[] = [
     module: async () => await import("./diagnostics/DiagnosticStatusPanel"),
   },
   {
-    title: `ROS Diagnostics – Summary`,
+    title: `Diagnostics – Summary (ROS)`,
     type: "DiagnosticSummary",
     description: "Display a summary of all ROS DiagnosticArray messages.",
     help: DiagnosticSummaryHelp,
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/panels/index.ts
```diff
@@ -56,15 +56,15 @@ const builtin: PanelInfo[] = [
     module: async () => await import("./ThreeDimensionalViz"),
   },
   {
-    title: `Diagnostics – Detail`,
+    title: `ROS Diagnostics – Detail`,
```
esthersweon, line 59: Can we do `Diagnostics - Detail (ROS)` instead?

Would be nice to have the name still lead with the panel function, vs the framework.

packages/studio-base/src/panels/index.ts
```diff
@@ -56,15 +56,15 @@ const builtin: PanelInfo[] = [
...
     type: "DiagnosticStatusPanel",
     description: "Display ROS DiagnosticArray messages for a specific hardware_id.",
     help: DiagnosticStatusPanelHelp,
     thumbnail: diagnosticStatusThumbnail,
     module: async () => await import("./diagnostics/DiagnosticStatusPanel"),
   },
   {
-    title: `Diagnostics – Summary`,
+    title: `ROS Diagnostics – Summary`,
```
esthersweon, line 67: Same comment here.

If there will be anything else in this category (ROS-specific panels), maybe we can create an entirely new "ROS" section in the panels list.

esthersweon (approved):
Just left minor comments around panel name.

packages/studio-base/src/panels/index.ts
jtbandes:
```suggestion
    title: `Diagnostics – Detail (ROS)`,
```

packages/studio-base/src/panels/index.ts
jtbandes:
```suggestion
    title: `Diagnostics – Summary (ROS)`,
```

Issue #3374, opened 2022-05-16
Update map panel settings to show error when custom url is malformed


defunctzombie:
**Description**
When a user enters a custom tile url in map panel settings, we should show an error if we consider the URL malformed.

Blocked on: https://github.com/foxglove/studio/pull/3373

PR #3390, opened 2022-05-18
Add error field to customTileUrl setting when invalid

defunctzombie:
**User-Facing Changes**
User sees an error indicator when they enter an invalid custom tile url field in map panel settings.

**Description**
When the customTileUrl setting value for the map panel is invalid, add an error message to the field to inform the user on what is invalid.

Fixes: #3374


packages/studio-base/src/panels/Map/MapPanel.tsx
```diff
@@ -85,10 +85,22 @@ function buildSettingsTree(config: Config, eligibleTopics: string[]): SettingsTr
 
   // Only show the custom url input when the user selects the custom layer
   if (config.layer === "custom") {
+    let error: string | undefined;
+    if (config.customTileUrl.length > 0) {
+      const placeholders = config.customTileUrl.match(/\{.+?\}/g) ?? [];
```
defunctzombie, line 90: @foxymiles This duplicates logic lower in the file where I also check for this before using assigning the value. I wanted to present this PR to get thoughts on how better to organize the logic around this to avoid duplicating this check and introducing inconsistencies over time.

Maybe for my use case the settings tree needs rebuilding when the user makes changes to the input?

foxymiles (approved):
Looks good to me. Maybe refactor validation into a function?

PR #3405, opened 2022-05-19
Add editable title to Plot panel

2metres:
**User-Facing Changes**
Screen Shot 2022-05-19 at 1 06 19 pm

**Description**

### DIFF
--- a/packages/studio-base/src/panels/Plot/index.tsx
+++ b/packages/studio-base/src/panels/Plot/index.tsx
@@ -17,7 +17,7 @@ import { InputBase, Typography } from "@mui/material";
 import produce from "immer";
 import { compact, set, uniq } from "lodash";
 import memoizeWeak from "memoize-weak";
-import { useEffect, useCallback, useMemo, ComponentProps, ChangeEvent } from "react";
+import { useEffect, useCallback, useMemo, useState, ComponentProps, ChangeEvent } from "react";
 
 import { filterMap } from "@foxglove/den/collection";
 import { useShallowMemo } from "@foxglove/hooks";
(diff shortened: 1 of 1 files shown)


2metres:
@foxymiles having trouble updating the panel settings for the Input `onChange` event

foxymiles:
@2metres All you have to do is change the config and the settings tree will automatically be rebuilt by the `useEffect` above the change handler:

```
  const handleTitleChange = useCallback(
    (event: ChangeEvent<HTMLInputElement>) => {
      saveConfig({ title: event.target.value });
    },
    [saveConfig],
  );
```

```
  useEffect(() => {
    updatePanelSettingsTree(panelId, {
      actionHandler,
      roots: buildSettingsTree(config),
    });
  }, [actionHandler, config, panelId, updatePanelSettingsTree]);
```

2metres:
Summarizing offline feedback from @amacneil:
> Plot title in toolbar should be only editable on double-click

amacneil:
> Summarizing offline feedback from @amacneil:
> 
> > Plot title in toolbar should be only editable on double-click

Yeah, the "always editable" title feels off to me. I think we should either make it only editable in the settings, or double click to edit (or right click in the future once we sort out our context menu story). See what feels the most natural.

foxymiles:
What if we put a little pen icon next to it that you click to go into edit mode?

foxymiles:
@2metres @amacneil The problem with using double click to edit is that double click is already a standard interaction to select text. What if we show a pencil icon on hover that the user can click on to edit the title?

amacneil:
Can we change the default name back to `Plot`?

I still find this interaction a little weird. Agree with @foxymiles point about double click already having an action, although selecting panel titles probably isn't very common.

I would be in favor of either just removing the inline edit (if you want to change the title you do it in the sidebar), or if we really want to keep double click to edit, then I think when you double click the title it should turn into something more obviously resembling a text input box (rather than just a tiny cursor appearing)

amacneil (approved):

PR #3420, opened 2022-05-20
[New3D] PointCloud fixes

jhurliman:
**User-Facing Changes**

None

**Description**

This includes a collection of fixes, additional implementation, and perf improvements for New3D point clouds that should address all point cloud issues except decay time.

- Fix divide by zero for pointcloud fields where min==max
- Make row_step errors non-critical since we currently ignore row_step
- Do per-point sRGB -> linear conversion on the GPU
- Render points as circles or squares
- Fix XBGR auto-selection for little-endian "rgb" fields
- Handle "rgb" fields misreported as FLOAT32
- Switch labels from sans-serif to monospace font
- Initialize Renderer with a config so it is never undefined, fix settings updates for point clouds when the first message is received

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/Labels.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/Labels.ts
@@ -176,7 +176,7 @@ export class Labels extends THREE.Object3D {
 
     const color = rgbaToLinear(makeRgba(), label.color ?? fgColor);
     const fontSize = label.fontSize ?? 14;
-    const fontFamily = fonts.SANS_SERIF;
+    const fontFamily = fonts.MONOSPACE;
     const fontWeight = 400;
     // const padding = label.padding ?? 2;
     const backgroundColor = rgbaToLinear(makeRgba(), label.backgroundColor ?? bgColor);
--- a/packages/studio-base/src/panels/ThreeDeeRender/MaterialCache.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/MaterialCache.ts
@@ -18,6 +18,37 @@ type MaterialCacheEntry = {
   disposer: DisposeMaterial;
 };
 
+// Fragment shader chunk to convert sRGB to linear RGB. This is used by some
+// PointCloud materials to avoid expensive per-point colorspace conversion on
+// the CPU. Source: <https://github.com/mrdoob/three.js/blob/13b67d96/src/renderers/shaders/ShaderChunk/encodings_pars_fragment.glsl.js#L16-L18>
+const FS_SRGB_TO_LINEAR = /* glsl */ `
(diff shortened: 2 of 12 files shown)


packages/studio-base/src/panels/ThreeDeeRender/Labels.ts
```diff
@@ -176,7 +176,7 @@ export class Labels extends THREE.Object3D {
 
     const color = rgbaToLinear(makeRgba(), label.color ?? fgColor);
     const fontSize = label.fontSize ?? 14;
-    const fontFamily = fonts.SANS_SERIF;
+    const fontFamily = fonts.MONOSPACE;
```
jhurliman, line 179: Point clouds look better when nearby labels have the correct font

packages/studio-base/src/panels/ThreeDeeRender/MaterialCache.ts
```diff
@@ -18,6 +18,37 @@ type MaterialCacheEntry = {
...
+// Fragment shader chunk to convert sRGB to linear RGB. This is used by some
+// PointCloud materials to avoid expensive per-point colorspace conversion on
+// the CPU. Source: <https://github.com/mrdoob/three.js/blob/13b67d96/src/renderers/shaders/ShaderChunk/encodings_pars_fragment.glsl.js#L16-L18>
+const FS_SRGB_TO_LINEAR = /* glsl */ `
+vec4 sRGBToLinear(in vec4 value) {
+	return vec4(mix(
+    pow(value.rgb * 0.9478672986 + vec3(0.0521327014), vec3(2.4)),
+    value.rgb * 0.0773993808,
+    vec3(lessThanEqual(value.rgb, vec3(0.04045)))
+  ), value.a);
```
jtbandes, line 30: can this just be `return vec4(sRGBToLinear(value.rgb), value.a)`?

packages/studio-base/src/panels/ThreeDeeRender/MaterialCache.ts
```diff
@@ -176,18 +207,44 @@ export const PointsVertexColor = {
...
       size: scale,
       sizeAttenuation: false,
     });
-    material.name = PointCloudColor.id(scale, transparent);
+    material.name = PointCloudColor.id(shape, encoding, scale, transparent);
     material.transparent = transparent;
     material.depthWrite = !transparent;
+    // Tell three.js to recompile the shader when `shape` or `encoding` change
+    material.customProgramCacheKey = () => `${shape}-${encoding}`;
+    material.onBeforeCompile = (shader) => {
```
jtbandes, line 233: With all this hackery of the built-in PointsMaterial shaders, should we just write our own shaders from scratch for point clouds?

jtbandes (approved):

PR #3421, opened 2022-05-20
[New3D] Show coordinate frames in the settings sidebar

jhurliman:
**User-Facing Changes**

None

**Description**

Adds a new top-level settings group called "Transforms" with each coordinate frame listed below. Only visibility toggling is currently supported, but all the code is in place to add more per-frame settings.

Screen Shot 2022-05-20 at 2 22 19 PM

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/settings.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/settings.ts
@@ -24,6 +24,10 @@ import {
   TRANSFORM_STAMPED_DATATYPES,
 } from "./ros";
 
+export type LayerSettingsTransform = {
+  visible: boolean;
+};
+
 export type ThreeDeeRenderConfig = {
   cameraState: CameraState;
   followTf: string | undefined;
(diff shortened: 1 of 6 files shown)


packages/studio-base/src/panels/ThreeDeeRender/settings.ts
```diff
@@ -154,8 +176,24 @@ export function buildSettingsTree(options: SettingsTreeOptions): SettingsTreeRoo
...
-  const topicsChildren: SettingsTreeChildren = {};
+  // Build the settings tree for transforms
+  const transformsChildren: SettingsTreeChildren = {};
+  const tfSettingsNodeProvider = settingsNodeProviders.get(LayerType.Transform);
+  if (tfSettingsNodeProvider == undefined) {
+    throw new Error(`Missing required settings node provider for LayerType.Transform`);
+  }
+  for (const { value: frameId } of options.coordinateFrames) {
+    // We key our memoized function by the first argument. Since the config
+    // maybe be undefined we use the config or the topic name.
```
jtbandes, line 187:
```suggestion
    // may be undefined we use the config or the topic name.
```

jtbandes (approved):

defunctzombie:
Nice! One thought I had while seeing the screenshot was that it could be cool to have a way of clicking each transform and making that the follow frame?

Issue #2038, opened 2021-10-27
Extensions API: support calling ROS services


jessicaaustin:
What the title says. 

Possible duplicate of https://github.com/foxglove/studio/issues/790 , though I wanted to be explicit about being able to do this within the context of an Extension.

amacneil:
@jessicaaustin are you still blocked on this or do you have a workaround for now?

jessicaaustin:
Not blocked at all. We moved to a model where Studio publishes on a "commands" topic, which is then picked up by a node on the robot that handles calling the actual services. For a number of reasons we'd be moving this direction anyway. 

So I could see this functionality being useful for someone else who uses services and is evaluating moving from rqt-->Studio, but we don't require it anymore.

defunctzombie:
@jessicaaustin The merged PR adds the interface for calling services. The interface is only available if the connected data source supports services. Services are implemented for the rosbridge data source only with the merged PR.

Issue #790, opened 2021-04-28
Support calling ROS services


amacneil:
Would be good to call services like you can do in rqt

defunctzombie:
> Related: Hide/show services in Topic Graph panel

Is blocked by adding an API to enumerate services

PR #3422, opened 2022-05-22
WIP Allow extensions to call services

rgov:
Related to #790 and #2038.

This plumbs through support for ROS service calls. My main interest is in supporting service calls with the Rosbridge data source, so I've prototyped that as well.

Would like to know if I've committed atrocities across the codebase.

### DIFF
--- a/packages/studio-base/src/components/PanelExtensionAdapter.tsx
+++ b/packages/studio-base/src/components/PanelExtensionAdapter.tsx
@@ -477,6 +477,19 @@ function PanelExtensionAdapter(props: PanelExtensionAdapterProps): JSX.Element {
           }
         : undefined,
 
+      callService: capabilities.includes(PlayerCapabilities.callServices)
+        ? async (service, request): Promise<unknown> => {
+            const ctx = latestPipelineContextRef.current;
+            if (!ctx) {
+              throw new Error("Unable to call service. There is no active connection.");
+            }
+            return await ctx.callService({
+              service,
+              request: request as Record<string, unknown>,
+            });
+          }
+        : undefined,
+
       unsubscribeAll: () => {
         subscribedTopicsRef.current.clear();
         setSubscriptions(panelId, []);
--- a/packages/studio-base/src/players/types.ts
+++ b/packages/studio-base/src/players/types.ts
@@ -59,6 +59,8 @@ export interface Player {
   // If the Player supports publishing (i.e. PlayerState#capabilities contains
   // PlayerCapabilities.advertise), publish a message.
   publish(request: PublishPayload): void;
+  // Make a service call. Available if `capabilities` contains PlayerCapabilities.callServices),
+  callService(request: ServiceCall): ServiceCallResult;
   // Basic playback controls. Available if `capabilities` contains PlayerCapabilities.playbackControl.
   startPlayback?(): void;
   pausePlayback?(): void;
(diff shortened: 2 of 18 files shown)


packages/studio-base/src/players/types.ts
```diff
@@ -298,11 +300,20 @@ export type AdvertiseOptions = {
 // The actual message to publish.
 export type PublishPayload = { topic: string; msg: Record<string, unknown> };
 
+// A service call with arguments.
+export type ServiceCall = { service: string, request: Record<string, unknown> };
+
+// FIXME: Maybe use Response instead of Result for consistency?
+export type ServiceCallResult = Promise<Record<string, unknown>>;
```
rgov, line 307: Should it be `ServiceCallResult`? Or just `ServiceResponse` (matching Roslib.js)?

packages/studio-base/src/components/PanelExtensionAdapter.tsx
```diff
@@ -477,6 +477,19 @@ function PanelExtensionAdapter(props: PanelExtensionAdapterProps): JSX.Element {
           }
         : undefined,
 
+      callService: capabilities.includes(PlayerCapabilities.callServices)
+        ? async (service: any /* FIXME */, request: any /* FIXME */): Promise<unknown> => {
```
rgov, line 481: Why do I have to explicitly say `any` here while the `publish` method right above doesn't have to?

packages/studio-base/src/players/types.ts
```diff
@@ -298,11 +300,20 @@ export type AdvertiseOptions = {
 // The actual message to publish.
 export type PublishPayload = { topic: string; msg: Record<string, unknown> };
```
rgov, line 301: I don't understand why this is a single parameter of `publish()` rather than two separate parameters.

packages/studio-base/src/players/types.ts
```diff
@@ -59,6 +59,8 @@ export interface Player {
   // If the Player supports publishing (i.e. PlayerState#capabilities contains
   // PlayerCapabilities.advertise), publish a message.
   publish(request: PublishPayload): void;
+  // Make a service call. Available if `capabilities` contains PlayerCapabilities.callServices),
+  callService(request: ServiceCall): ServiceCallResult;
```
rgov, line 63: This is the *only* method defined on `Player` that returns any value. I think it should return a Promise that gets resolved with the service response object. Is this violating the design in some way, and it should pass in a callback instead perhaps?

packages/studio/index.d.ts
```diff
@@ -222,6 +222,15 @@ declare module "@foxglove/studio" {
...
     publish?(topic: string, message: unknown): void;
 
+    /**
+     * Call a service.
+     *
+     * @param service The name of the service to call
+     * @param request The service request
+     * @returns A promise that will be resolved when the result is available
+     */
+    callService?(service: string, request: unknown): Promise<unknown>;
```
defunctzombie, line 232: `callService<T>?(service: string, request: unknown): Promise<T>`

defunctzombie: Promise<ServiceResonse<T>>

defunctzombie: We can use the promise rejection for error handling.

jhurliman: Can `request` be narrowed down to something like the `XmlRpcValue` type?

rgov: What is the benefit for a more narrow `request` type if `publish()`'s `message` is also `unknown`? They seem similar to me.

defunctzombie:
Thanks for your initial work here. Per our discussion, I've picked up this PR and created a new one in our repo to make it easier for foxglove collaborators to iterate and leave your branch as you have it

Closing this PR in favor of the new one here: https://github.com/foxglove/studio/pull/3552

PR #3423, opened 2022-05-23
Improve settings tree node header styling

foxymiles:
**User-Facing Changes**
None

**Description**
This makes two changes:
1. Node headers are only bolded at indent level 0 and 1
2. The visibility toggles are only visible when the node is hovered

Screen Shot 2022-05-23 at 9 14 25 AM

### DIFF
--- a/packages/studio-base/src/components/SettingsTreeEditor/NodeEditor.tsx
+++ b/packages/studio-base/src/components/SettingsTreeEditor/NodeEditor.tsx
@@ -34,9 +34,17 @@ const NodeHeader = muiStyled("div")(({ theme }) => {
     gridColumn: "span 2",
     paddingRight: theme.spacing(1.5),
 
+    ".MuiCheckbox-root": {
+      visibility: "hidden",
+    },
+
     "&:hover": {
       outline: `1px solid ${theme.palette.primary.main}`,
       outlineOffset: -1,
+
+      ".MuiCheckbox-root": {
+        visibility: "visible",
+      },
     },
   };
 });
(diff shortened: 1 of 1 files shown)


jhurliman:
I’m curious what @2metres thinks of the visibility icons only on hover, I thought we were steering away from UX that requires hover to work.

foxymiles:
This is something I discussed with @2metres because we felt the repetition of the always visible eye icons was too noisy but it would be good to get his final signoff on this implementation.

packages/studio-base/src/components/SettingsTreeEditor/NodeEditor.tsx
```diff
@@ -34,9 +34,17 @@ const NodeHeader = muiStyled("div")(({ theme }) => {
...
+    },
+
     "&:hover": {
       outline: `1px solid ${theme.palette.primary.main}`,
       outlineOffset: -1,
+
+      ".MuiCheckbox-root": {
+        visibility: "visible",
+      },
     },
```
2metres, line 48:
```suggestion
  "@media (pointer: fine)": {
      ".MuiCheckbox-root": {
        visibility: "hidden",
      },
  
      "&:hover": {
        outline: `1px solid ${theme.palette.primary.main}`,
        outlineOffset: -1,
  
        ".MuiCheckbox-root": {
          visibility: "visible",
        },
      },
    },
```

2metres (changes requested):
Looks good to me. just need to wrap the hover interaction in a media query so that the actions always appear on touchscreens.

2metres (approved):

PR #3424, opened 2022-05-23
Improve handling of invalid color values in settings editor

### LINKED ISSUE #3415: Settings sidebar crashes when typing invalid input in color field
When I type `123` into a color field, the settings sidebar crashes:

image

foxymiles:
**User-Facing Changes**
None

**Description**
This uses `tinycolor` to validate & format color values before they are passed to the `ColorSwatch` component.

Fixes #3415

### DIFF
--- a/packages/studio-base/src/components/SettingsTreeEditor/inputs/ColorPickerInput.tsx
+++ b/packages/studio-base/src/components/SettingsTreeEditor/inputs/ColorPickerInput.tsx
@@ -49,7 +49,7 @@ export function ColorPickerInput(props: ColorPickerInputProps): JSX.Element {
   const open = Boolean(anchorElement);
 
   const isValidColor = value != undefined && tinycolor(value).isValid();
-  const swatchColor = isValidColor ? value : "#00000044";
+  const swatchColor = isValidColor ? tinycolor(value).toHex8String() : "#00000044";
 
   return (
     <Root>


packages/studio-base/src/components/SettingsTreeEditor/inputs/ColorPickerInput.tsx
```diff
@@ -49,7 +49,7 @@ export function ColorPickerInput(props: ColorPickerInputProps): JSX.Element {
   const open = Boolean(anchorElement);
 
   const isValidColor = value != undefined && tinycolor(value).isValid();
-  const swatchColor = isValidColor ? value : "#00000044";
+  const swatchColor = isValidColor ? tinycolor(value).toHex8String() : "#00000044";
```
jtbandes, line 52: Seems like it would be better to invoke tinycolor() only once so we don’t waste CPU on parsing it twice?

jtbandes (approved):
So does this mean that MUI’s idea of a valid color is more restrictive than tinycolor’s? Because I see we already had a validity check.

packages/studio-base/src/components/SettingsTreeEditor/inputs/ColorPickerInput.tsx
foxymiles: It seems unlikely this would be expensive enough to affect performance.

PR #3439, opened 2022-05-24
Prompt for login when opening data platform urls

### LINKED ISSUE #3411: Automatically redirect to login for data platform player
**Description**
If someone shares a data platform link with me, but I am not signed in to Studio, I get an error:

image

Instead of this unhelpful error, we should just redirect the user to the signin page (or on desktop, pop up the signin dialog).

foxymiles:
**User-Facing Changes**
None

**Description**
This prompts the user to sign in and then continues to open the URL if the user tries to open a data platform URL before signing in to studio. This touches on some complex and fragile aspects of the studio session startup process and is a bit of a bandaid solution given that the state we're concerned with here spans several independent React contexts each with their own async transitions and error states.

The approach here depends on the following points:
1. `Workspace` now accesses the `useInitialDeepLinkState` hook directly, which now returns a value indicating that a user session is required for the datasource in the URL.
2. If the datasource requires a user session and the user isn't logged in, `Workspace` will present a sign in modal to the user.
3. The `useInitialDeepLinkState` now tracks the various pieces of URL state to apply more explicitly and in separate `useEffect` instances, in the case of the datasource waiting for a user session to try to select it.
4. The web vs desktop launch preference default has been separated into a separate, unrelated hook to reduce the complexity of `useInitialDeepLinkState`.

I suspect at some point we're going to have to commit to a larger refactor of our session bootstrapping code, including user sessions, layout syncing, datasource loading and timestamp seeking but this PR should address the immediate issue of guiding users to sign in upon opening a datasource URL.

Fixes #3411

### DIFF
--- a/packages/studio-base/src/hooks/useDefaultWebLaunchPreference.ts
+++ b/packages/studio-base/src/hooks/useDefaultWebLaunchPreference.ts
@@ -1,0 +1,35 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { useEffect } from "react";
+
+import { AppSetting } from "@foxglove/studio-base/AppSetting";
+import {
+  MessagePipelineContext,
+  useMessagePipeline,
+} from "@foxglove/studio-base/components/MessagePipeline";
+import { useSessionStorageValue } from "@foxglove/studio-base/hooks/useSessionStorageValue";
+import isDesktopApp from "@foxglove/studio-base/util/isDesktopApp";
+
+const selectHasUrlState = (ctx: MessagePipelineContext) => ctx.playerState.urlState != undefined;
+
+export function useDefaultWebLaunchPreference(): void {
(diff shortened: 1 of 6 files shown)


packages/studio-base/src/hooks/useDefaultWebLaunchPreference.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export function useDefaultWebLaunchPreference(): void {
+  const hasUrlState = useMessagePipeline(selectHasUrlState);
+  const [launchPreference, setLaunchPreference] = useSessionStorageValue(
+    AppSetting.LAUNCH_PREFERENCE,
+  );
+
+  // Set a sessionStorage preference for web if we have a stable URL state.
+  // This allows us to avoid asking for the preference immediately on
+  // launch of an empty session and makes refreshes do the right thing.
```
jtbandes, line 25: I'm confused by this comment. In an empty session, would we have a url state?

foxymiles: To be honest this entire hook may have been obsoleted by our introduction of the open datasource dialog.

amacneil:
🎉

Issue #3430, opened 2022-05-23
Ensure Protobuf timestamps (google.protobuf.Timestamp) are parsed correctly


jtbandes:
@foxglove/schemas has changed to use `google.protobuf.Timestamp` instead of `foxglove.Time` for protobuf message timestamps. We need to ensure these are correctly parsed & handled in Studio.

PR #3443, opened 2022-05-24
Handle google.protobuf.Timestamp/Duration in protobuf messages

jtbandes:
**User-Facing Changes**
Protobuf messages using `google.protobuf.Timestamp` and `Duration` are now treated as though `seconds` and `nanos` were named `sec` and `nsec`, consistent with other timestamp values in the app.

**Description**
Fixes https://github.com/foxglove/studio/issues/3430


defunctzombie:
What would be an alternative to modifying the user's input?

jtbandes:
All the code in the app that uses `sec`/`nsec` would need to learn to also handle `seconds`/`nanos`. Or we would need to keep both the original and modified copies of the message threaded through everywhere.

PR #3450, opened 2022-05-25
Animate panels' transitions into/out of fullscreen

jtbandes:
**User-Facing Changes**
Added animation effect when panels enter or exit fullscreen mode.

**Description**
Uses [`<Transition>`](https://reactcommunity.org/react-transition-group/transition) to manage enter/exit transition states.

Was able to remove the need for making ancestor panels fullscreen at the same time (added in https://github.com/foxglove/studio/pull/2276) by removing the `z-index: 1` when not in fullscreen mode.

https://user-images.githubusercontent.com/14237/170348201-4387f9ce-6ce5-4c9f-b867-5c4029cac805.mov


foxymiles:
I like the idea 👍. Should we make the transition a little faster?

foxymiles (approved):

PR #3451, opened 2022-05-25
Clear topic subscription errors when new messages are received

### LINKED ISSUE #3293: When a connection to a topic is reestablished, it should clear the errors
**Description**
When a node that publishes on a topic is killed or dies and a custom extension panel is subscribed to this topic, an error message is displayed on foxglove. But when the node is re-run and the topic is being used again, the error is not cleared even though the whole system is valid.

- Version: 20.04.4
- OS: Ubuntu

**Steps To Reproduce**
1. Subscribe to a topic on a custom extension published by a ros node
2. Kill the node that publishes on the topic
3. re-run the node 

image

**Expected Behavior**
When the connection to a topic is reestablished, the errors should be cleared

**Actual Behavior**
When the connection to a topic is reestablished, the errors are still there even though the topic is being used

foxymiles:
**User-Facing Changes**
None

**Description**
Currently once we encounter a subscription error on a topic on a ROS connection we display a player error but we don't clear this error once the subscription to that topic is working again. The solution here is to clear the error once we successfully receive a message on that topic again.

Fixes #3293

### DIFF
--- a/packages/studio-base/src/players/Ros1Player.ts
+++ b/packages/studio-base/src/players/Ros1Player.ts
@@ -393,9 +393,11 @@ export default class Ros1Player implements Player {
         const newDatatypes = this._getRosDatatypes(datatype, msgdef);
         this._providerDatatypes = new Map([...this._providerDatatypes, ...newDatatypes]);
       });
-      subscription.on("message", (message, data, _pub) =>
-        this._handleMessage(topicName, message, data.byteLength, true),
-      );
+      subscription.on("message", (message, data, _pub) => {
+        this._handleMessage(topicName, message, data.byteLength, true);
+        // Clear any existing subscription problems for this topic if we're receiving messages again.
+        this._clearProblem(`subscribe:${topicName}`, { skipEmit: true });
+      });
       subscription.on("error", (error) => {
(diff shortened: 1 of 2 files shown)


jhurliman (approved):

jhurliman:
Would the ROS2 player benefit from the same fix?

foxymiles:
> Would the ROS2 player benefit from the same fix?

Yeah looks like it. I'll add it there as well.

defunctzombie:
Since there can be multiple publishers to a single topic - what happens when a subscriber to one of the publishers fails but the other is working? Would we still surface an error to the user? If we have a subscription id we should include that in the error key so we can clear only the errors for the subscription that is working again.

foxymiles:
Good question. Looks like right now we don't know the id of the publisher that caused the error?

https://github.com/foxglove/studio/blob/5680d79476a20e4ccc1889a1c52d06ddfcce80d4/packages/studio-base/src/players/Ros1Player.ts#L401

defunctzombie:
@foxymiles I see yes the subscription is for any publisher. We'd need to update our ros1 library to provide that additional information in the `error`. Something we'll have to stash away for later.

PR #3453, opened 2022-05-25
[New3D] Render images in the 3D view

jhurliman:
**User-Facing Changes**

None

**Description**

Renders `sensor_msgs/Image` and `sensor_msgs/CompressedImage` topics in New3D. The PinholeCameraModel is used to create a rectified plane projected a user-configurable distance from the camera frame origin. Image messages are converted to texture data and uploaded to the GPU as they arrive.

**TODO**

- [ ] Finish support for `sensor_msgs/Image`
- [x] Increase the deformation grid from 1x1 to something larger (10x10? User-configurable?) and test with distortion
- [x] Fix colorspace

https://user-images.githubusercontent.com/195374/170363483-3687558e-218a-40c5-aac6-1cc8595680ae.mov


foxymiles (approved):
New topic-level icons look good.

packages/studio-base/src/panels/ThreeDeeRender/normalizeMessages.ts
```diff
@@ -30,6 +32,39 @@ export function normalizeTime(time: Partial<Time> | undefined): Time {
   return { sec: time.sec ?? 0, nsec: time.nsec ?? 0 };
 }
 
+export function normalizeByteArray(byteArray: unknown): Uint8Array {
+  if (byteArray == undefined) {
+    return new Uint8Array(0);
+  } else if (byteArray instanceof Uint8Array) {
+    return byteArray;
+  } else if (Array.isArray(byteArray) || byteArray instanceof ArrayBuffer) {
```
jtbandes, line 40: is there a real use case for these? aren't users expected to provide a foxglove.CompressedImage which will have a Uint8Array? when would we get an Array or ArrayBuffer?

packages/studio-base/src/panels/ThreeDeeRender/renderables/Images.ts
```diff
@@ -0,0 +1,450 @@
...
+  renderable.userData.material = texture ? createMaterial(texture, renderable) : undefined;
+
+  // Destroy the mesh, it needs to be rebuilt
+  if (renderable.userData.mesh) {
+    renderable.remove(renderable.userData.mesh);
+    renderable.userData.mesh = undefined;
+  }
+}
+
+async function createBitmap(image: CompressedImage): Promise<ImageBitmap> {
```
jtbandes, line 314: Did you mean to remove the `async` too?

packages/studio-base/src/panels/ThreeDeeRender/renderables/Images.ts
```diff
@@ -0,0 +1,450 @@
...
+      vertices[vOffset + 0] = p.x;
+      vertices[vOffset + 1] = p.y;
+      vertices[vOffset + 2] = p.z - EPS;
+
+      uvs[uvOffset + 0] = ix / gridX;
+      uvs[uvOffset + 1] = iy / gridY;
+    }
+  }
+
+  geometry.setAttribute("position", new THREE.BufferAttribute(vertices, 3));
```
jtbandes, line 385: how does this end up turning into a plane? I guess the geometry uses indices?

packages/studio-base/src/panels/ThreeDeeRender/renderables/Images.ts
```diff
@@ -0,0 +1,450 @@
...
+          renderable.userData.mesh = undefined;
+        }
+      }
+    }
+
+    // Create or update the bitmap texture
+    if ((image as Partial<CompressedImage>).format) {
+      const compressed = image as CompressedImage;
+      createBitmap(compressed)
+        .then((bitmap) => {
```
jtbandes, line 242: what if a new image was received before the createBitmap task finished and we need to drop a frame?

packages/studio-base/src/panels/ThreeDeeRender/normalizeMessages.ts
jhurliman: An array can come from JSON decoding

jtbandes: it shouldn't — the json schemas use `{ type: "string", contentEncoding: "base64" }` (documented at: https://foxglove.dev/docs/studio/connection/foxglove-websocket) and parse the base64 string into a Uint8Array

packages/studio-base/src/panels/ThreeDeeRender/renderables/Images.ts
jhurliman: I think it's ok if another frame starts decoding while the previous frame decoding is still in progress, as long as they don't decode out of order. Should a counter be used to avoid out-of-order decodes?

PR #3455, opened 2022-05-26
Extension sidebar cleanup

2metres:
**User-Facing Changes**
Clean up extension styling

Before
Screen Shot 2022-05-26 at 3 32 27 pm

After
Screen Shot 2022-05-26 at 3 32 24 pm

**Description**
Refactor Extension Sidebar to use MUI styles, one less instance of fluentui, makeStyles and @foxglove/Button

### DIFF
--- a/packages/studio-base/src/components/ExtensionDetails.tsx
+++ b/packages/studio-base/src/components/ExtensionDetails.tsx
@@ -3,7 +3,16 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
 import ChevronLeftIcon from "@mui/icons-material/ChevronLeft";
-import { IconButton, Button, Link, Tab, Tabs, Typography, Divider } from "@mui/material";
+import {
+  IconButton,
+  Button,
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/components/ExtensionDetails.tsx
```diff
@@ -25,31 +34,11 @@ type Props = {
...
-
-const Description = styled.div`
-  margin-top: 16px;
-  margin-bottom: 16px;
-`;
+const StyledButton = muiStyled(Button)({ minWidth: 100 });
 
 export function ExtensionDetails({ extension, onClose, installed }: Props): React.ReactElement {
-  const [isInstalled, setIsInstalled] = useState(installed);
+  const [isInstalled, setIsInstalled] = useState<boolean>(installed);
```
foxymiles, line 40: Typescript can infer the type of the state hook from the argument so in cases like this you can leave out the `<boolean>`.

foxymiles (approved):

PR #3460, opened 2022-05-26
Configure webpack to allow foxglove to be served at a URL other than the root

foxymiles:
**User-Facing Changes**
None

**Description**
This uses the webpack `auto` publicPath feature to make it possible for foxglove to be served at a URL other than the root and still load assets correctly.


amacneil (approved):
seems to work

Issue #2909, opened 2022-02-24
Support flatbuffers in MCAP


jameskuszmaul-brt:
In studio, support flatbuffer schemas and messages.

I can generate a sample MCAP log that matches the spec I'm thinking of, but for now I'm thinking:

New schema encoding `flatbuffer`. This will be a flatbuffer `reflection.Schema` type (definition: https://github.com/google/flatbuffers/blob/master/reflection/reflection.fbs), encoded as a flatbuffer.

New message encoding `flatbuffer`. This will be a serialized flatbuffer of the type represented by the Schema.

Useful references:

Flatbuffer Serialization Spec: https://google.github.io/flatbuffers/flatbuffers_internals.html
Crude typescript reflection library for flatbuffers: https://github.com/frc971/971-Robot-Code/blob/master/aos/network/www/reflection.ts (this is pretty self-contained, but is also missing some features, and I haven't really vetted it against, e.g., malformed flatbuffers).
C++ libraries that convert between JSON <-> flatbuffers: https://github.com/frc971/971-Robot-Code/blob/master/aos/json_to_flatbuffer.cc

If we end up using foxglove going forwards, probably going to want to use this, but for the time being using JSON gets us most of the way there and lets us test things out.

For working with logs offline, I'm actually not sure how much performance benefit we'd get out of this. It'll make MCAP file sizes *much* smaller, but I'm not sure how much faster typescript is really able to parse flatbuffers using reflection than straight JSON.

PR #3465, opened 2022-05-27
Flatbuffers support in @foxglove/mcap-support

jameskuszmaul-brt:
**User-Facing Changes**

Adds support for flatbuffers as an MCAP encoding in studio.

**Description**

This copies in a flatbuffers reflection library that I wrote
(originally for https://github.com/frc971/971-Robot-Code, the main 
modifications being to bring it up to snuff with the typescript checks
and linters being run here) and uses it to implement reading of
flatbuffers in the `mcap-support` library.

Caveats:
* Pulling in the flatbuffers reflection library is a bit messy right
  now. I have a PR up upstream with the contents of the package
  (https://github.com/google/flatbuffers/pull/7301), but haven't yet
  sorted out how to actually deal with the package (naming it
  flatbuffers_reflection seems suboptimal, but since I'm used to
  just dealing with Bazel I haven't needed to worry about that before).
* Performance still isn't as great as I'd like, but I need to
  profile things better. I think a lot of the losses are in
  deserializing the entire flatbuffer message, rather than
  just deserializing the fields that we care about.

Solves #2909


defunctzombie:
@jameskuszmaul-brt Are you looking for some guidance on how to proceed or think you've got a good sense of the remaining items to land this? We're definitely 👍 to adding flatbuffer decoding support to studio so let us know if there's anything blocking you on this.

jameskuszmaul-brt:
@defunctzombie Thanks for checking in. I've been a bit busy lately. I don't think there's too much left here (main blockers are ensuring it passes CI and then I need to write appropriate tests for the new code)--I'll try to get back to it some time next week.

I'd also sort of been hoping that the code I've pulled in as `packages/flatbuffers_reflection` would get merged to upstream flatbuffers, but I haven't had the time to push on that PR either, so I think I'll keep that as-is for now.

jameskuszmaul-brt:
It is non-obvious to me why the CI is failing here, help would be appreciated; other than that I think this is good for review. @defunctzombie

jtbandes:
I think it would be better to make the flatbuffers_reflection package a separate repo. If you need a starting point for a TS package you can use [foxglove/template-typescript](https://github.com/foxglove/template-typescript).

packages/mcap-support/src/parseChannel.ts
```diff
@@ -74,12 +75,26 @@ export function parseChannel(channel: Channel): ParsedChannel {
...
-      deserializer = (data) =>
+      deserializer = (data) => {
         postprocessValue(JSON.parse(textDecoder.decode(data)) as Record<string, unknown>);
+      };
     }
     return { fullSchemaName: channel.schema.name, deserializer, datatypes };
   }
 
+  if (channel.messageEncoding === "flatbuffer") {
+    if (channel.schema?.encoding !== "flatbuffer") {
```
jtbandes, line 86:
```suggestion
  if (channel.messageEncoding === "flatbuffers") {
    if (channel.schema?.encoding !== "flatbuffers") {
```

Seems like the official name is "flatbuffers" and I don't see this shortened version anywhere?

packages/mcap-support/src/parseFlatbufferSchema.ts
```diff
@@ -0,0 +1,202 @@
...
+      return "float32";
+    case BaseType.Double:
+      return "float64";
+    case BaseType.String:
+      return "string";
+    case BaseType.Vector:
+    case BaseType.Obj:
+    case BaseType.Union:
+    case BaseType.Array:
+      return "";
```
jtbandes, line 43: Is the expectation that this will never be called for these types? Then maybe we should throw an error instead of returning `''`?

packages/mcap-support/src/parseFlatbufferSchema.ts
```diff
@@ -0,0 +1,202 @@
...
+  return "";
+}
+
+// eslint-disable-next-line no-restricted-syntax
+function flatbufferString(unchecked: string | Uint8Array | null | undefined): string {
+  if (typeof unchecked === "string") {
+    return unchecked;
+  }
+  throw new Error("Not a string.");
+  return "";
```
jtbandes, line 57: nit: unreachable

packages/mcap-support/src/parseFlatbufferSchema.ts
```diff
@@ -0,0 +1,202 @@
...
+  }
+  return "";
+}
+
+// eslint-disable-next-line no-restricted-syntax
+function flatbufferString(unchecked: string | Uint8Array | null | undefined): string {
+  if (typeof unchecked === "string") {
+    return unchecked;
+  }
+  throw new Error("Not a string.");
```
jtbandes, line 56: More helpful error message might be nice, e.g.
```suggestion
  throw new Error(`Expected string, found ${typeof unchecked}`);
```

packages/mcap-support/src/parseFlatbufferSchema.ts
```diff
@@ -0,0 +1,202 @@
...
+    case BaseType.Double:
+    case BaseType.String:
+    case BaseType.None: {
+      const simpleType = typeForSimpleField(field.type.baseType);
+      // Enums have magic logic--the constants definitions for the enum values
+      // have to go right before the enum itself.
+      if (field.type.index !== -1) {
+        const enums = schema.enums[field.type.index]?.values;
+        if (enums == undefined) {
+          throw new Error("Invalid schema");
```
jtbandes, line 83:
```suggestion
          throw new Error(`Invalid schema, missing enum values for field type ${field.type.index}`);
```

packages/mcap-support/src/parseFlatbufferSchema.ts
```diff
@@ -0,0 +1,202 @@
...
+      throw new Error(
+        `Type "${fullSchemaName}" is not available in the schema for "${schema.rootTable?.name}".`,
+      );
+    }
+  }
+  const deserializer = (buffer: ArrayBufferView) => {
+    const byteBuffer = new ByteBuffer(
+      new Uint8Array(buffer.buffer, buffer.byteOffset, buffer.byteLength),
+    );
+    const parser = new Parser(rawSchema);
```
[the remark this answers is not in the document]
jameskuszmaul-brt: It can be, although I don't believe it should be that expensive (if you look at the constructor for `Parser`, it is empty and just taking the schema as a member). Moved out on principle, though.

packages/mcap-support/src/parseChannel.ts
jameskuszmaul-brt: Oops, my MCAP file generator was using the wrong name....

jameskuszmaul-brt:
> I think it would be better to make the flatbuffers_reflection package a separate repo, and publish the package to NPM. If you need a starting point for a TS package you can use [foxglove/template-typescript](https://github.com/foxglove/template-typescript).

I am pretty aggressively trying to avoid being responsible for package management. I'll push again on upstream flatbuffers to see about getting this code in there.

packages/mcap-support/src/parseChannel.ts
jtbandes: We should probably change that, the flatbuffers docs seem to call them flatbuffers. cc @amacneil

amacneil: I don't feel strongly. I can find places in the docs where they say things like `the FlatBuffer compiler`. Protobuf is also called `Protocol buffers` in most places. As long as the mcap docs match our behavior.

amacneil: In [this list](https://github.com/foxglove/mcap/blob/main/docs/specification/appendix.md#well-known-message-encodings) of `well known message encodings`, they are all singular so it seems like making `flatbuffers` randomly plural would look a bit out of place. But I don't feel strongly about it.

jtbandes: > In [this list](https://github.com/foxglove/mcap/blob/main/docs/specification/appendix.md#well-known-message-encodings) of well known message encodings, they are all singular so it seems like making flatbuffers randomly plural would look a bit out of place

didn’t we just make up this list recently though?

amacneil: they have all existed at least as long as the mcap spec being finalized. before that I can't remember.

jtbandes: Ok, I see the docs do sometimes use "flatbuffer" so this is fine.

https://google.github.io/flatbuffers/flatbuffers_guide_tutorial.html

> - Write a FlatBuffer schema file.
> - Use the flatc FlatBuffer compiler.
> - Parse [JSON](http://json.org/) files that conform to a schema into FlatBuffer binary files.

packages/flatbuffers_reflection/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "flatbuffers_reflection",
+  "description": "Typescript library for doing reflection on flatbuffers.",
+  "license": "Apache 2",
```
defunctzombie, line 4: Different license from the rest of the code in our studio repo is a good reason for publishing this as a separate npm package.

@jameskuszmaul-brt We'll discuss internally and make a decision on whether to include in our repo or make a separate one this week.

defunctzombie:
This is great! We'll have a chat internally about the additional package and how we want to handle that and get this moved across the finish line.

packages/flatbuffers_reflection/package.json
jameskuszmaul-brt: Ah, that can get relicensed if need be. I wrote the relevant code in that package in my personal capacity (originally for the frc team) and am fine with importing it here and relicensing under MPL-2.0--Apache 2 is what flatbuffers and the frc971 repositories use so this was almost more habit than anything else.

packages/mcap-support/src/parseFlatbufferSchema.ts
```diff
@@ -0,0 +1,204 @@
...
+    case BaseType.Union:
+    case BaseType.Array:
+    case BaseType.MaxBaseType:
+    case undefined:
+      throw new Error("Unions and Arrays are not supported in mcap-support currently.");
+  }
+  return fields;
+}
+
+// TODO(jkuszmaul): It should be reasonably trivial to support a lazy message reader in the same
```
defunctzombie, line 153: I prefer not leaving TODO in the code since these items rarely get re-visited. If you want to leave a note for future readers of a tradeoff thats fine - otherwise I'd remove this.

jameskuszmaul-brt: Done.

jameskuszmaul-brt: #3734

Issue #3437, opened 2022-05-24
Plot Panel Y-limit bugs


jackiescanlon:
**Description**
There are a few bugs in the Y-limits settings for Plot panels. I grouped them altogether in this issue but am happy to separate them if that would be better.

- Version: 1.12.0
- OS: Ubuntu 20.04
- Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native): Seems to be data source agnostic - I am able to replicate the issues even when not connected to a data source. For my example below I used a ROS 1 live connection.
- ROS distro (if applicable): Noetic

**Steps To Reproduce and Expected Behavior**

1. Create a new layout and add a Plot panel
2. Plot something
3. Open the panel settings and look at Y-min and Y-max.

**Auto-scale**

**Expected result:** I would've expected the defaults to be `auto` and `auto`, with the plot auto-scaling.
**Actual result:** The plot is auto-scaling as expected, but the Y-min and Y-max are reading back `0` and `0` instead of `auto` and `auto`. This is confusing, especially because if you backspace the `0`s they will turn to `auto`. 
**Workaround:** The only way to set an actual limit of 0 is to backspace the 0 and retype it in.

![image](https://user-images.githubusercontent.com/43035541/170037530-898045c8-c99c-4641-b8d3-4fa8b6de166b.png)

![image](https://user-images.githubusercontent.com/43035541/170037573-537e0d2e-bb56-4eec-a271-41b165e53b99.png)

**Negative values**

4. Now, in the Y-min box, backspace the 0 and attempt to type -10.

**Expected result:** As I type `-10`, it appears in the textbox, and the minimum on the plot goes to `-10`.
**Actual result:** A `-` is not allowed in the textbox when the textbox is empty. 
**Workaround:** Type `10`, and then move the cursor to the front to type the `-`. It seems to accept the `-` as long as the textbox already has other text in it and isn't empty.

**Reversed values**

5. Attempt to set the Y-min to 10 and the Y-max to -10 (using the above workaround to type the `-`). I've accidentally done this many times, reversing the Y-min and the Y-max (I think my brain wants to have the Y-max on the top because its setting the value for the top of the graph).

**Expected result:** Studio throws an error at me, because my Y-max is less than my Y-min. Or even better, Studio figures out that i just reversed them and plots [10, 10].
**Actual result:** Foxglove does figure out that I reversed them and plots [-10, 10], which is really helpful. However, for some reason it removes all the horizontal lines on my plot :/
[...]

PR #3467, opened 2022-05-31
Show error when plot settings has ymin > ymax

foxymiles:
**User-Facing Changes**
The plot panel settings UI will now indicate an error when a y min value > y max value is specified.

**Description**
This just sets the `error` state of the ymax field if both ymax and ymin are entered and ymin > ymax.

Screen Shot 2022-05-31 at 7 43 43 AM

Partially addresses https://github.com/foxglove/studio/issues/3437


packages/studio-base/src/panels/Plot/settings.ts
```diff
@@ -2,11 +2,18 @@
...
+
 import { SettingsTreeRoots } from "@foxglove/studio-base/components/SettingsTreeEditor/types";
 
 import { PlotConfig } from "./types";
 
 export function buildSettingsTree(config: PlotConfig): SettingsTreeRoots {
+  const maxYError =
+    isNumber(config.minYValue) && isNumber(config.maxYValue) && config.minYValue >= config.maxYValue
+      ? "Y max must be greater than Y min."
+      : undefined;
```
jtbandes, line 15: Just musing out loud... I guess this logic needs to be kept in sync with the plot rendering logic itself, which needs some fallback values to use in this case? I guess that will be a general challenge with keeping settings logic in sync with display logic.

jtbandes (approved):

packages/studio-base/src/panels/Plot/settings.ts
foxymiles: Seems like chartjs does its best to make sense of nonsensical values at least in this case and just swaps the min and max. Not ideal but at least the user can see something.

PR #3468, opened 2022-05-31
Map log level severity values by message datatype

### LINKED ISSUE #3456: The Log message can be shown rightly, but the filter still can not be used.
Thanks for fixing the problem.The Log message can be shown rightly, but the filter still can not be used.

filter1

_Originally posted by @xyzxyzxyzxyzxz in https://github.com/foxglove/studio/issues/3417#issuecomment-1136612711_

foxymiles:
**User-Facing Changes**
None

**Description**
This allows the log panel message filtering logic to handle datasources that use different numeric values to indicate different message severity levels. 

This implementation uses the message datatype to determine if log level mapping is necessary.

Fixes #3456

### DIFF
--- a/packages/studio-base/src/panels/Log/conversion.tsx
+++ b/packages/studio-base/src/panels/Log/conversion.tsx
@@ -25,7 +25,7 @@ export function getNormalizedMessage(logMessage: LogMessageEvent["message"]): st
   return "";
 }
 
-function getNormalizedLevel(datatype: string, raw: LogMessageEvent["message"]) {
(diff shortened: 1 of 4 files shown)


foxymiles:
@jtbandes Good catch I missed that. I've reworked it to use the existing log level conversion logic.

jtbandes (approved):
Thanks, this looks better. Still a bit strange that this panel filters messages and _then_ normalizes them, seems like it would be simpler to normalize up front (but maybe a little less performant)

foxymiles:
Yeah I considered just normalizing the entire message array and this does make the downstream logic simpler but I wasn't sure of the performance impact.

PR #3469, opened 2022-05-31
Fix pose publication

ayrton04:
**User-Facing Changes**
The pose that gets published using the 3D panel's pose tool will now use the position of the first click (start of the pose arrow), rather than the position of the second click (end of the pose arrow).

**Description**
When setting a pose, two mouse clicks are required: one to specify the point, and a second one to provide direction. However, the current code is sending the position of the _second_ click (i.e., the end of the pose arrow) to the robot. This makes sending a robot to a goal pose very challenging.

end_of_arrow


foxymiles (approved):
This makes sense. Thanks for the fix!

PR #3470, opened 2022-05-31
Prototype plot series settings

### LINKED ISSUE #3432: Refactor plot panel to use settings sidebar
Now that settings nodes can have actions ( #3419 )  we should refactor the plot panel to make use of the settings sidebar

foxymiles:
**User-Facing Changes**
This moves the UI for adding new data series in the plot panel from the current legend into the settings sidebar UI.

**Description**
This is a prototype of how this might work. A couple of issues present themselves:

1. The UI is less compact, especially with all series expanded.
2. The message path input is narrow which is going to be awkward for the long message paths that are common in the plot UI. Maybe we should consider the full-width / 2 row input type we discussed earlier.
3. Because each series has to exist as a node in the tree it has to have a name. Just naming them `Series X` sequentially doesn't seem very helpful to the user.
4. It will be particularly important that we direct users that are used to the current legend to the sidebar so they understand how to configure plot series with this new approach.

Screen Shot 2022-05-31 at 1 05 29 PM

Fixes #3432


defunctzombie:
I really like the UI here. I think this has to pair with showing a legend in the plot and allowing the user to set a name for each series (default to the message path input).

Is it reasonable to land this as a feature flag so we can iterate on the experience without the logic going stale?

foxymiles:
Yeah it wouldn't be hard to feature flag this.

PR #3471, opened 2022-05-31
Remove panel error boundary showing details story

foxymiles:
**User-Facing Changes**
None

**Description**
This removes the `PanelErrorBoundary` story with the error details shown. Since the error details contain checksums that vary from build to build this story gets erroneously flagged as changed on every PR.


jhurliman (approved):

defunctzombie:
@jtbandes thanks for catching this and reverting - 👍 to keeping this story cause its useful for both dev work on the error boundary and regression detection.

PR #3475, opened 2022-06-01
Fix imageview hitmap detection

### LINKED ISSUE #3473: Image panel "selected point" seems to pop up when clicking on pixels
**Description**
Only seems to happen when there is an overlay on the image. Clicking anywhere on the image (not on the annotations) seems to cause the select point dropdown to cover up the whole image.

foxymiles:
**User-Facing Changes**
This fixes an issue with marker click to inspect in the image panel.

**Description**
The issue was that the `HitmapRenderContext` was using a context with no alpha but we need the alpha value to signify the absence of a marker.

This PR also hides the inspection UI when the user clicks away from a point containing a marker.

Fixes #3473

### DIFF
--- a/packages/studio-base/src/panels/Image/components/ImageCanvas.tsx
+++ b/packages/studio-base/src/panels/Image/components/ImageCanvas.tsx
@@ -495,6 +495,8 @@ export function ImageCanvas(props: Props): JSX.Element {
       .then((r) => {
         if (r?.marker) {
           props.setActivePixelData(r);
+        } else {
+          props.setActivePixelData(undefined);
         }
       });
   }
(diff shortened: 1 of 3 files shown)


jtbandes (approved):
> we need the alpha value to signify the absence of a marker

Can you explain more about how this works and why alpha is needed? Is it because this function expects to be able to read the alpha value? https://github.com/foxglove/studio/blob/c5a8364ff1321b63a72847947d7e4e0ed8c86a12/packages/studio-base/src/panels/Image/lib/util.ts#L61-L63

foxymiles:
> > we need the alpha value to signify the absence of a marker
> 
> Can you explain more about how this works and why alpha is needed? Is it because this function expects to be able to read the alpha value?

The way the hitmap works is to render the markers into a separate canvas with the color = the integer index of the marker and with alpha = 255. I haven't found a way to disable anti-aliasing/blending in this offscreen canvas so markers get rendered with alpha values and interpolated colors that don't correspond to the color we want to use to encode the marker index so we only use pixels with max alpha which have not been blended.

jtbandes:
But then how did it ever work with `alpha: false`?

foxymiles:
It never did with with alpha=false. 

I think I inadvertently broke it trying to improve performance and we didn't notice it right away since it seems that in electron the canvas always has an alpha layer so the bug didn't manifest on desktop.

PR #3478, opened 2022-06-01
Fix image view empty state logic

foxymiles:
**User-Facing Changes**
Fix fixes an erroneous waiting state in the image panel UI

**Description**
The new experimental players have different message loading behavior and that has exposed an error in the logic we use to determine whether or not we should show an empty state in the image panel where the empty state is sometimes shown on top of an image.

This PR changes the logic to be consistent between determining what to render and showing the empty state and also clears the cached last image on topic change.

Screen Shot 2022-06-01 at 12 18 26 PM

### DIFF
--- a/packages/studio-base/src/panels/Image/index.tsx
+++ b/packages/studio-base/src/panels/Image/index.tsx
@@ -285,7 +285,7 @@ function ImageView(props: Props) {
     return <TopicDropdown multiple={false} title={title} items={items} onChange={onChange} />;
   }, [cameraTopic, allImageTopics, onChangeCameraTopic]);
 
-  const showEmptyState = !image;
+  const showEmptyState = !imageMessageToRender;
 
   return (
     <Stack flex="auto" overflow="hidden" position="relative">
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/panels/Image/index.tsx
```diff
@@ -229,6 +230,10 @@ function ImageView(props: Props) {
   // Improve perf by hiding the ImageCanvas while seeking, instead of unmounting and remounting it.
   const imageMessageToRender = image ?? lastImageMessageRef.current;
 
+  useUpdateEffect(() => {
+    lastImageMessageRef.current = undefined;
+  }, [cameraTopic]);
```
jtbandes, line 235: This code above should also be moved into an effect:
```ts
  const lastImageMessageRef = useRef(image);
  if (image) {
    lastImageMessageRef.current = image;
  }
```

jtbandes: Add a comment saying why we clear this out?

jtbandes (approved):

jtbandes (approved):

PR #3486, opened 2022-06-03
Cleanup mismatched styling in node playground chrome

2metres:
**User-Facing Changes**
Before
Screen Shot 2022-06-03 at 1 43 21 pm

After
Screen Shot 2022-06-09 at 10 45 18 am

**Description**
Cleanup nodeplayground chrome to use MUI components. See storybook for all changes to log styling


foxymiles (approved):

packages/studio-base/src/panels/NodePlayground/BottomBar/index.tsx
```diff
@@ -1,33 +1,24 @@
 // This Source Code Form is subject to the terms of the Mozilla Public
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
-//
```
defunctzombie, line 4: I don't think we can remove this copyright just because we changed some of the elements here. We don't know what code remains from Cruise.

defunctzombie (approved):
make sure to put relevant copyright/license back before merge

PR #3488, opened 2022-06-03
Handle empty string in node paths in settings

### LINKED ISSUE #3483: Settings sidebar action callback returns the incorrect path for child nodes with the key `""`
**Description**

https://github.com/foxglove/studio/blob/main/packages/studio-base/src/components/SettingsTreeEditor/NodeEditor.tsx#L128 is using `""` as a reserved key to refer to the parent, which causes a key collision if there is a child settings node with the key `""`.

foxymiles:
**User-Facing Changes**
None

**Description**
This fixes an issue with paths in nodes in the settings tree that have the empty string `""` as a key.

Fixes #3483

### DIFF
--- a/packages/studio-base/src/components/SettingsTreeEditor/NodeEditor.tsx
+++ b/packages/studio-base/src/components/SettingsTreeEditor/NodeEditor.tsx
@@ -123,9 +123,10 @@ function NodeEditorComponent(props: NodeEditorProps): JSX.Element {
   const { fields, children } = settings;
   const hasProperties = fields != undefined || children != undefined;
 
-  // Provide stable subpaths so that memoization works.
+  // Provide stable subpaths so that memoization works. We force a dependency on path
+  // here to make sure we reset the cache if the path changes.
   const stablePaths = useMemo<Record<string, readonly string[]>>(
-    () => ({ "": props.path }),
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/components/SettingsTreeEditor/NodeEditor.tsx
```diff
@@ -123,9 +123,10 @@ function NodeEditorComponent(props: NodeEditorProps): JSX.Element {
   const { fields, children } = settings;
   const hasProperties = fields != undefined || children != undefined;
 
-  // Provide stable subpaths so that memoization works.
+  // Provide stable subpaths so that memoization works. We force a dependency on path
+  // here to make sure we reset the cache if the path changes.
   const stablePaths = useMemo<Record<string, readonly string[]>>(
-    () => ({ "": props.path }),
+    () => ({ [Symbol()]: props.path }),
```
jtbandes, line 129: What is going on here? Could we produce the desired values inside the useMemo instead of mutating below?

foxymiles: We need to memoize for each child path so we'd need some kind of `useMemo` in a loop but you can't do that with hooks.

jhurliman (approved):

PR #3489, opened 2022-06-03
Delay layout syncs when the app is under heavy load

### LINKED ISSUE #3274: Loading layouts sometimes takes a long time
**Description**

When I open the Layouts side panel, sometimes it takes a few seconds for the layouts to load. Today, things were especially bad, with loading taking 30s or more at times.

https://user-images.githubusercontent.com/1159930/165590102-fe5833f9-9f52-4180-b7ee-24e55a058df3.mp4

This also happens after the Layout panel loads for the first time -- it looks like it gets refreshed periodically, and while it's refreshing, I can't switch to another layout.

- Version: 1.7.0
- OS: Ubuntu

**Steps To Reproduce**

I don't have nice clean steps to replicate here with a dummy setup, so I'll try to describe my real setup and what I'm seeing as best as possible.

* I am logged in with an account, and I have about ~50 personal layouts and about ~10 team layouts. Some of these layouts are fairly complicated.
* It looks like there's a request for `/layouts` that happens periodically, all of those are coming through quickly so it doesn't seem like a delay due to a remote server
   * Selection_999(511)
* I'm not seeing any obvious errors in the Console when this happens, though I do see some warnings, dunno if they're related
   * Selection_999(513)

The one thing I think is influential is the complexity of the currently loaded layout, and (possibly) the overall load on the computer too. For example, if I load a complex layout with lots of plots, then in general Studio becomes sluggish, including the processing of Layouts. Here's one example where i took the demo layout and duplicated a bunch of panels and added a plot panel. I'm able to get Studio processes to use a lot of CPU this way, and then when I open the Layouts side panel, it takes ~5s to load.

Selection_999(519)

I understand that high load on the running machine is out of your hands, but this can lead to frustration because if I made a layout "too complicated" then I can't switch out of it without having to reload everything. Perhaps one "fix" might be to have a setting about how often to refresh layouts, so I could lower that refresh rate in this situation? Or to not freeze the layouts panel and allow you to switch while it's being loaded? For example I tried logging out of my account so it would stop loading, but then all my layouts were gone. :shrug: 

I have a few videos of this behavior on the real setup if it's helpful, DM me in slack if you'd like to see them.

foxymiles:
**User-Facing Changes**
This improves layout syncing behavior when the app is under heavy load.

**Description**
The idea here is to monitor the time it takes to do a layout sync and make sure the app doesn't spend more than a certain fraction of runtime in layout syncs.

Fixes #3274

### DIFF
--- a/packages/studio-base/src/providers/LayoutManagerProvider.tsx
+++ b/packages/studio-base/src/providers/LayoutManagerProvider.tsx
@@ -56,12 +56,12 @@ export default function LayoutManagerProvider({
     const controller = new AbortController();
     void (async () => {
       let failures = 0;
-      let syncDuration = 0;
+      let lastSyncDuration = 0;
       while (!controller.signal.aborted) {
         try {
           const startTime = performance.now();
           await layoutManager.syncWithRemote(controller.signal);
-          syncDuration = performance.now() - startTime;
+          lastSyncDuration = performance.now() - startTime;
           failures = 0;
         } catch (error) {
           log.error("Sync failed:", error);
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/providers/LayoutManagerProvider.tsx
```diff
@@ -54,9 +56,12 @@ export default function LayoutManagerProvider({
     const controller = new AbortController();
     void (async () => {
       let failures = 0;
+      let syncDuration = 0;
```
jtbandes, line 59: rename to `lastSyncDuration`?

jtbandes (approved):

packages/studio-base/src/providers/LayoutManagerProvider.tsx
```diff
@@ -65,10 +70,21 @@ export default function LayoutManagerProvider({
...
-          Math.random() * Math.min(SYNC_INTERVAL_MAX_MS, SYNC_INTERVAL_BASE_MS * 2 ** failures);
-        log.debug("Waiting", (duration / 1000).toFixed(2), "sec for next sync", { failures });
-        await delay(duration);
+        const backoffDuration = Math.random() * SYNC_INTERVAL_BASE_MS * 2 ** failures;
+
+        // In high CPU load scenarios syncing can be slow to complete. We make sure that
+        // we're not spending more than a certain fraction of run time doing syncs.
+        const idleDuration = lastSyncDuration * SYNC_INTERVAL_IDLE_RATIO;
+
+        const timeToDelay = clamp(
```
jtbandes, line 79: I think this clamp kind of defeats some of the purpose of the jitter (it makes it more likely that the final delay time is base_ms). But I think it's not a big deal since this is only in particularly slow sync scenarios.

foxymiles: Yeah good point.

I'd like to see us move this from a pull to a push mechanism which would eliminate this issue entirely.

foxymiles: Yeah I think we'd need some kind of push + fallback scheme.

jessicaaustin:
@foxymiles 

To test this, I pulled this branch locally, run a clean `yarn install` then `yarn serve` + `yarn start`.

I'm not sure how I can log in with the dev version -- it sends me to http://console-dev.foxglove.dev/ and my google credentials don't work there -- so I just tested without any remote layouts to sync. Interestingly, it still takes 5-10s to load the layouts panel every time for my "lots of plots" layout, even though I'm not logged in so in theory it's not hitting this updated block of code at all. This is the same as if I download and run 1.11.0 of Studio.

foxymiles:
Looks like the original issue is not actually related to syncing. Closing this for now.

PR #3494, opened 2022-06-03
Only wait to apply layout id for authorized datasources

### LINKED ISSUE #3492: Layout doesn't load until data source is loaded
When I have a layout & local file source loaded, and I reload the app, my layout in the background is blank until I select a new data source.

image

If I exit out of the open dialog instead of selecting a source, no layout is selected, even though I had one selected before reloading:

image

foxymiles:
**User-Facing Changes**
None

**Description**
This tweaks the deep link logic state to only delay loading the specified layout if for data sources that require authorization.

Fixes #3492

### DIFF
--- a/packages/studio-base/src/hooks/useInitialDeepLinkState.ts
+++ b/packages/studio-base/src/hooks/useInitialDeepLinkState.ts
@@ -71,23 +71,21 @@ export function useInitialDeepLinkState(deepLinks: readonly string[]): {
 
   // Select layout from URL.
   useEffect(() => {
-    if (!unappliedUrlState) {
+    if (!unappliedUrlState?.layoutId) {
       return;
     }
 
-    // If we have a target datasource then wait until the player has loaded to select the layout.
-    // This also handles the case where the datasource requires a logged in user
-    // since the player won't start until we have a current user.
-    if (targetUrlState?.ds && playerPresence !== PlayerPresence.PRESENT) {
+    // If our datasource requires a current user then wait until the player is
+    // available to load the layout since we may need to sync layouts first and
+    // that's only possible after the user has logged in.
+    if (currentUserRequired && playerPresence !== PlayerPresence.PRESENT) {
       return;
     }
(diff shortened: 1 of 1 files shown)


jtbandes:
This doesn't really make sense since the layout is not related to the data source. I wonder if instead we should wait to load the layout if the data source is being automatically loaded. In the bug I filed the local file source does not get restored when the page loads, so it's easy to close the open dialog and see the bug. But with a source that would automatically connect, the open dialog should close on its own (or never appear).

foxymiles:
The layout is related to the datasource in the sense that we have special logic to queue up and retry loading a layout after a sync and the sync will only kick off after a user has logged in.

I agree it would be better to make all these dependencies explicit and manage them in a controlled, unified procession of states.

jtbandes:
> The layout is related to the datasource in the sense that we have special logic to queue up and retry loading a layout after a sync and the sync will only kick off after a user has logged in.

What does the user logging in have to do with the data source? 😀 

I know the answer is that in the specific case of the data platform source, the user _must_ be logged in to use it. But other sources can be used whether or not the user is logged in.

foxymiles:
@jtbandes Yeah that's why this patch only waits to apply the layoutId if the data source is one that requires a user.

foxymiles:
@jtbandes You can see here that if we get a layoutId for a layout we don't have the layoutmanager will try to load it after a sync but it does this immediately. So for this to work we have to delay until a sync is possible:

https://github.com/foxglove/studio/blob/5e9cdbc7e70f42eb1ab74495653fe2222f1e2c32/packages/studio-base/src/services/LayoutManager/LayoutManager.ts#L180-L207

The idea of waiting until a player is present is to delay as long as possible before loading the layout to give the other parts of studio enough time to set up:

```typescript
    if (currentUserRequired && playerPresence !== PlayerPresence.PRESENT) {
      return;
    }
```
If you don't like this approach can you propose an alternative?

PR #3498, opened 2022-06-06
Restyle action overlay using muiStyled

2metres:
**User-Facing Changes**
Before
Screen Shot 2022-06-08 at 11 55 58 am

After
Screen Shot 2022-06-08 at 12 03 01 pm

**Description**
Restyle action overlay using muiStyled

### DIFF
--- a/packages/studio-base/src/components/Panel.tsx
+++ b/packages/studio-base/src/components/Panel.tsx
@@ -11,12 +11,11 @@
 //   found at http://www.apache.org/licenses/LICENSE-2.0
 //   You may not use this file except in compliance with the License.
 
-import { makeStyles } from "@fluentui/react";
-import BorderAllIcon from "@mdi/svg/svg/border-all.svg";
-import ExpandAllOutlineIcon from "@mdi/svg/svg/expand-all-outline.svg";
-import GridLargeIcon from "@mdi/svg/svg/grid-large.svg";
-import TrashCanOutlineIcon from "@mdi/svg/svg/trash-can-outline.svg";
-import { styled as muiStyled } from "@mui/material";
+import CancelPresentationOutlinedIcon from "@mui/icons-material/CancelPresentationOutlined";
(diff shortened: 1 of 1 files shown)


foxymiles (approved):

packages/studio-base/src/components/Panel.tsx
```diff
@@ -69,114 +66,56 @@ import {
...
-  perfInfo: {
-    position: "absolute",
-    whiteSpace: "pre-line",
-    bottom: 2,
-    left: 2,
-    fontSize: "9px",
-    opacity: 0.7,
-    userSelect: "none",
-    mixBlendMode: "difference",
+const ActionsOverlay = muiStyled("div")(({ theme }) => ({
```
defunctzombie, line 70: Does muiStyled use stable classes under the hood or is this somehow turning into another `sx` everywhere?

defunctzombie (approved):

packages/studio-base/src/components/Panel.tsx
2metres: Seems fairly stable from what I can tell. None of the keymatching that happens in sx happens in muiStyles

packages/studio-base/src/components/Panel.tsx
```diff
@@ -648,18 +586,30 @@ export default function Panel<
...
-                      </Icon>
-                      Group in tab
-                    </Button>
-                    <Button className={classes.tabActionsOverlayButton} onClick={createTabs}>
-                      <Icon size="small" style={{ marginBottom: 5 }}>
-                        <ExpandAllOutlineIcon />
-                      </Icon>
-                      Create {numSelectedPanelsIfSelected} tabs
-                    </Button>
+                    <Grid container spacing={1} padding={2} maxWidth={300}>
```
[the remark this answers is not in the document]
2metres: Long answer spacing is not using the `sx` prop from what I can tell. Padding might be.

PR #3513, opened 2022-06-07
Put related annotation topics at top of annotation list in image panel settings

### LINKED ISSUE #3505: Image panel - fix image annotation topics sort order
**Description**
When the image annotation topics were in the toolbar as a dropdown we would sort "relevant" topics before other topics so they appear first in the list. Relevant topics are those matching the image topic prefix.

image

With the new annotation topics list in panel settings this list is unsorted making the relevant topics harder to find.

foxymiles:
**User-Facing Changes**
This sorts the annotation topics in the image panel settings so that topics related to the camera topic are at the top of the list.

**Description**
This also changes the styling of the in-panel camera topic picker to a standard select since it's no longer a multi-valued select.

Screen Shot 2022-06-07 at 9 42 25 AM

Note that this does not implement annotation separation / segmentation as described in #3505 since it seems we need to think through the right approach in more detail first.

Fixes #3505

### DIFF
--- a/packages/studio-base/src/panels/Image/components/TopicDropdown.tsx
+++ b/packages/studio-base/src/panels/Image/components/TopicDropdown.tsx
@@ -44,6 +44,7 @@ export function TopicDropdown(props: Props): JSX.Element {
         disabled={items.length === 0}
         displayEmpty
         title={title}
+        renderValue={(_selected) => title}
         size="small"
         variant="filled"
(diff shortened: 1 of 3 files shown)


defunctzombie (approved):
👍 cool - TopicDropdown itself looks like it could almost go too. I guess it does simplify the parent component use site.

PR #3519, opened 2022-06-07
[New3D] Fix Line2 picking

belopot:
**User-Facing Changes**

None

**Description**
Add picking value to userData of line mesh

- hitmap before change
Screenshot 2022-06-08 012715

- hitmap after change
Screenshot 2022-06-08 012818

As you can see through above screenshots, Line2 meshes appeared


jhurliman:
This turned out to be related to a race condition I just fixed where we were trying to modify `ShaderLib.line` while three.js was internally doing the same.

jhurliman:
Closing in favor of https://github.com/foxglove/studio/pull/3524

PR #3521, opened 2022-06-07
[WIP] Memoize 3d panel settings elements

foxymiles:
**User-Facing Changes**
None

**Description**
This uses memoization to speed up the construction of the new 3d panel settings.


jhurliman:
Using an immer-based approach instead

Issue #3481, opened 2022-06-01
Experimental bag player breaks image panel backfill


jtbandes:
**Description**
When turning on/off an image annotation with the experimental bag player, the image disappears and doesn't reappear. With the old bag player, the image reappears when the missing annotation is turned off again.

https://user-images.githubusercontent.com/14237/171507123-2dc190a9-9eba-4611-93d9-6bd3bc789bf5.mov

PR #3522, opened 2022-06-07
Implement requestBackfill for IterablePlayer

defunctzombie:
**User-Facing Changes**
When using any of the experimental players and subscribing to a topic which already has a subscription panels correctly receive their requested message.

**Description**
The IterablePlayer setSubscriptions implementation would ignore any calls which did not result in actually changing the subscriptions. This behavior resulted in panels which want to subscribe to a topic that is already subscribed from getting the message. The way message pipeline handles this is by invoking requestBackfill after a setSubscriptions call but since requestBackfill was not implemented in IterablePlayer (since setSubscriptions itself changed the state to load when needed), this meant panels did not get messages they needed.

This change fixes this behavior by moving the "backfill" state change logic from setSubscriptions into requestBackfill for the IterablePlayer. In a future world we can remove requestBackfill and have message pipeline understand when subscriptions have not changed since it also has the last player state but for now we adhere to the expected pattern of setSubscriptions followed by requestBackfill.

Fixes: #3481

### DIFF
--- a/packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
+++ b/packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
@@ -259,7 +259,9 @@ export class IterablePlayer implements Player {
 
     this._allTopics = allTopics;
     this._partialTopics = partialTopics;
+  }
 
+  requestBackfill(): void {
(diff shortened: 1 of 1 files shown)


jtbandes (approved):

defunctzombie:
@esthersweon 👍 no need to mention in changelog

PR #3524, opened 2022-06-07
[New3D] Fix Line2 picking shader, support world units

jhurliman:
**User-Facing Changes**

None

**Description**

Line lists and line strips are now clickable in New3D. Previously, we were using the wrong vertex shader and didn't support lines in world units.

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/MaterialCache.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/MaterialCache.ts
@@ -311,11 +311,12 @@ export const LineVertexColor = {
 };
 
 export const LineVertexColorPicking = {
-  id: (lineWidth: number): string => `LineVertexColorPicking-${lineWidth.toFixed(4)}`,
+  id: (lineWidth: number, worldUnits: boolean): string =>
+    `LineVertexColorPicking-${lineWidth.toFixed(4)}${worldUnits ? "-w" : ""}`,
 
-  create: (lineWidth: number): THREE.ShaderMaterial => {
+  create: (lineWidth: number, worldUnits: boolean): THREE.ShaderMaterial => {
     return new THREE.ShaderMaterial({
-      vertexShader: THREE.ShaderLib["line"]!.vertexShader,
+      vertexShader: THREE.ShaderLib["foxglove.line"]!.vertexShader,
       fragmentShader: /* glsl */ `
         uniform vec4 objectId;
         void main() {
@@ -325,7 +326,6 @@ export const LineVertexColorPicking = {
       clipping: true,
       uniforms: {
         objectId: { value: [NaN, NaN, NaN, NaN] },
-        worldUnits: { value: 0 },
         linewidth: { value: lineWidth },
         resolution: { value: new THREE.Vector2(1, 1) },
         dashOffset: { value: 0 },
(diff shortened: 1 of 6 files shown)


belopot (approved):

packages/studio-base/src/panels/ThreeDeeRender/renderables/markers/materials.ts
```diff
@@ -131,17 +131,24 @@ export function releaseLineMaterial(marker: Marker, materialCache: MaterialCache
...
-    () => LineVertexColorPicking.create(lineWidth),
+    LineVertexColorPicking.id(lineWidth, worldUnits),
+    () => LineVertexColorPicking.create(lineWidth, worldUnits),
     LineVertexColorPicking.dispose,
   );
 }
 
-export function releaseLinePickingMaterial(lineWidth: number, materialCache: MaterialCache): void {
-  materialCache.release(LineVertexColorPicking.id(lineWidth));
+export function releaseLinePickingMaterial(
```
defunctzombie, line 145: How come this release function needs worldUnits but the other release functions don't? The inconsistency in release calls made me wonder this.

packages/studio-base/src/panels/ThreeDeeRender/MaterialCache.ts
```diff
@@ -311,11 +311,12 @@ export const LineVertexColor = {
...
 export const LineVertexColorPicking = {
-  id: (lineWidth: number): string => `LineVertexColorPicking-${lineWidth.toFixed(4)}`,
+  id: (lineWidth: number, worldUnits: boolean): string =>
+    `LineVertexColorPicking-${lineWidth.toFixed(4)}${worldUnits ? "-w" : ""}`,
 
-  create: (lineWidth: number): THREE.ShaderMaterial => {
+  create: (lineWidth: number, worldUnits: boolean): THREE.ShaderMaterial => {
     return new THREE.ShaderMaterial({
-      vertexShader: THREE.ShaderLib["line"]!.vertexShader,
+      vertexShader: THREE.ShaderLib["foxglove.line"]!.vertexShader,
```
jtbandes, line 319: Don't we need a corresponding change in LineMaterial.ts?

packages/studio-base/src/panels/ThreeDeeRender/renderables/markers/materials.ts
jhurliman: The `release*()` calls release references using the material `id()`, which is parameterized by a different list of arguments for each material type. For most of our materials, we either always use world units or never use them, so there's no ambiguity. Our other custom line materials (currently) always use world units. LinePickingMaterial is a bit different since it gets used with both our custom line materials and the default THREE.js line material with worldUnits=false, so we need to specify whether we're using worldUnits or not. That's the essence of this bugfix.

Overall, I'm not that happy with the current MaterialCache implementation. It achieves the desired effect of reducing total material instances (which can sometimes mean fewer GL commands), but several different places in the codebase have to be changed to add a new cacheable material or change the parameterization of an existing one. The code is confusing to navigate if you're not familiar with it.

This is something I'd like to revisit in a cleanup pass once we're at feature parity.

packages/studio-base/src/panels/ThreeDeeRender/MaterialCache.ts
jhurliman: Yes, that change already merged into main.

PR #3530, opened 2022-06-08
Refactor image panel download code

foxymiles:
**User-Facing Changes**
None

**Description**
This extracts the download image logic from the image canvas component into a separate function. This is just a refactor to keep the upcoming context menu PR more manageable.

### DIFF
--- a/packages/studio-base/src/panels/Image/lib/downloadImage.ts
+++ b/packages/studio-base/src/panels/Image/lib/downloadImage.ts
@@ -1,0 +1,59 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { Topic } from "@foxglove/studio";
+import { downloadFiles } from "@foxglove/studio-base/util/download";
+
+import { Config, NormalizedImageMessage } from "../types";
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/panels/Image/lib/downloadImage.ts
```diff
@@ -0,0 +1,59 @@
...
+    },
+  });
+
+  if (!dimensions) {
+    return;
+  }
+
+  // context: https://stackoverflow.com/questions/37135417/download-canvas-as-png-in-fabric-js-giving-network-error
+  // read the canvas data as an image (png)
+  tempCanvas.toBlob((blob) => {
```
jtbandes, line 45: to make this async function actually wait, we should wrap this in `new Promise()` and resolve at the end of the callback

jtbandes (approved):

packages/studio-base/src/panels/Image/lib/downloadImage.ts
defunctzombie:
```
const blob = new Promise((resolve) => tempCambas.toBlob(resolve));

// rest of the code that was in the closure before
```

defunctzombie (approved):
👍 small suggestion to keep the download function async throughout the entire download. Before it didn't matter cause the callback wasn't async but now that you have cleaned this up this seems nicer

PR #3534, opened 2022-06-08
Add new PanelContextMenu component

### LINKED ISSUE #3055: ContextMenu panel API

foxymiles:
**User-Facing Changes**
None

**Description**
This adds a simple `PanelContextMenu` component that makes it easy for panels to offer a right-click context menu of actions, including a set of actions common to all panels.

In this implementation a panel needs to explicitly add a `<PanelContextMenu>` child component to offer a context menu. The component takes both a function to generate menu items, optionally based on a click position, and a handler function to respond to menu clicks:

```typescript
type PanelContextMenuProps = {
  /**
   * Function that returns a list of menu items, optionally dependent on the x,y
   * position of the click.
   */
  itemsForClickPosition: (position: {
    x: number;
    y: number;
  }) => DeepReadonly<PanelContextMenuItem[]>;
};
```

Menu items can either be a label with an optional icon or a divider:

```typescript
/**
 * Types of items that can be included in a context menu. Either a clickable item
 * or a divider.
 */
export type PanelContextMenuItem =
  | {
      /** Type of selectable menu items. */
      type: "item";

      /** True if the item should be shown but disabled. */
      disabled?: boolean;

      /** Label shown for the menu item. */
      label: string;

      /** Callback triggered by clicking the item. */
      onclick: () => void;
    }
  | {
      /** Type of item dividers. */
      type: "divider";
    };
```

Screen Shot 2022-06-09 at 6 48 54 AM

This implementation adds a new context menu to the image panel as a reference.

Currently this only supports the "remove" common option to keep this PR more manageable but we should also support the for splitting the panel and replacing it with a new panel of a different type in a followup PR.

Fixes #3055


jtbandes:
Do we have a plan for exposing this in the extension API? I guess that API will have to look pretty different if the extensions aren't required to use React?

packages/studio-base/src/components/PanelContextMenu.tsx
```diff
@@ -0,0 +1,170 @@
...
+      return;
+    }
+
+    const parent: HTMLElement | ReactNull = element.closest(PanelRoot);
+    parent?.addEventListener("contextmenu", listener);
+
+    return () => {
+      parent?.removeEventListener("contextmenu", listener);
+    };
+  }, [listener]);
```
jtbandes, line 132: Did you test what happens when the panel gets moved to a new location in the mosaic? Is there a chance the listener doesn't get reconfigured on the new parent and is left on the old parent? Might be a good thing to add a comment on since this is going a bit outside the normal React flow by doing direct DOM traversal.

foxymiles:
> Do we have a plan for exposing this in the extension API? I guess that API will have to look pretty different if the extensions aren't required to use React?

We'd have to add some new functions on the extension `context` to support this. The two callbacks for generating items and for handling clicks would have to be registered with the context perhaps like the `onRender` callback is now.

packages/studio-base/src/components/PanelContextMenu.tsx
foxymiles: That's a good question I'll have to check that.

foxymiles: It looks since this remains within the same `PanelRoot` the listeners are still attached to the right element.

defunctzombie:
Do we plan to expose this to the extension API? Or is this not exposing to the extension api yet? Wondering how we will handle icons for that.

foxymiles:
Yeah we will definitely expose this to the extension API. My plan was to use the same approach for icons that we use in the settings API which we also expose to extensions. In that case icons are specified by name as a string drawn from the set in https://github.com/foxglove/studio/blob/miles/panel-context-menu/packages/studio-base/src/components/CommonIcons.tsx

PR #3537, opened 2022-06-08
Add function style panel saveConfig

foxymiles:
**User-Facing Changes**
None

**Description**
This expands on the panel's `saveConfig` method to allow a function form similar to the react form of `setState` that takes the old state as an argument to a function that returns a new derived state. 

This makes it easier to write stable callbacks that update a panel config as a derived value of the old config, as commonly seen in the `actionHandler` in the panel settings API.

As a concrete example, you can see we can now effectively memoize part of the image panel's settings API which previously had no effect because the `actionHandler` was rebuilt on every config change.


defunctzombie (approved):
👍 Don't see any changes that impact public APIs so this should be good to try out internally and see what we think

PR #3539, opened 2022-06-08
Reorganize panel toolbar buttons

### LINKED ISSUE #3515: Settings are hard to find/show
As the settings sidebar is gaining functionality, opening it continues to be a two-click process. The user has to click the settings cog in the panel and then panel settings. We should make opening the panel settings a 1-click action.

Related items:

The panel settings navbar icon does not match the panel settings icon in the dropdown panel actions.
The dropdown panel actions need a place to live. They could live in a panel context menu or a hamburger menu (or other icon).
Discoverability for new users that settings is an important thing to open. Maybe when adding a new panel open the settings?

foxymiles:
**User-Facing Changes**
This reorganizes the panel toolbar buttons to make them more ergonomic.

**Description**
This makes the following changes:
1. The cog icon now directly opens settings.
2. There is now a more options button that opens the menu that used to be attached to the cog icon.
3. The drag handle has been removed and the entire panel toolbar is now draggable.

Screen Shot 2022-06-09 at 9 00 08 AM

Fixes #3515

### DIFF
--- a/packages/studio-base/src/components/PanelLayout.stories.tsx
+++ b/packages/studio-base/src/components/PanelLayout.stories.tsx
@@ -169,7 +169,10 @@ export const FullScreen = (): JSX.Element => {
         omitDragAndDrop
         onMount={() => {
           setTimeout(() => {
-            (document.querySelectorAll("[data-test=panel-toolbar-fullscreen]")[0] as any).click();
+            (document.querySelectorAll("[data-test=panel-menu]")[0] as any).click();
+          }, DEFAULT_CLICK_DELAY);
+          setTimeout(() => {
+            (document.querySelectorAll("[data-test=panel-menu-fullscreen]")[0] as any).click();
           }, DEFAULT_CLICK_DELAY);
         }}
       >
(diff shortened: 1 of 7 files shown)


amacneil:
> I think we should still link the the panel help content in the sidebar next to the panel title

where?

we could add `Help` to the `:` menu, if that's what you mean?

foxymiles:
@amacneil @2metres https://github.com/foxglove/studio/pull/3549

defunctzombie (approved):

amacneil:
I love it - looks so much cleaner!

Some assorted thoughts:

- Having the "grab" cursor appear for tab titles seems weird. Maybe we should make tab headers non-draggable? (i.e. the dead space for the overall tab group would still be draggable, but the individual tab titles would be non-draggable)
- The right hand icons being clickable and also draggable is a nice migration path, but feels like we should probably remove this eventually (as @jtbandes pointed out, this is uncommon UX). Maybe in the short term we make the settings icon non-draggable and only let you drag the `:` menu?
- For raw messages, it would be nice if the message path autocomplete didn't take up the full width by default (i.e. expand to text length). Not sure how difficult that is, but it could be a separate PR.
- We should make a separate ticket to go through and clean up the few remaining panels that have a database icon (move it to the panel settings).

PR #3540, opened 2022-06-08
Add link to Foxglove website from web app

esthersweon:
**User-Facing Changes**
- Add link to Foxglove website from web app
- Fixes the following ahrefs issue: 
  Screen Shot 2022-06-08 at 3 10 30 PM


defunctzombie (changes requested):
The user is already using the app and the help section has links to the website. Maybe add a link to the website in the help section. I do not think this needs to take up sidebar space. Let's be mindful of using our app space for app features not to appease SEO.

PR #3541, opened 2022-06-09
[New3D] Fix config mutations for Renderer#config

jhurliman:
**User-Facing Changes**

None

**Description**

Fixes the occasional crash when initializing New3D.


amacneil:
Separately I'm curious why sometimes I got this crash and the dead panel display didn't show up. I couldn't reproduce it consistently, but happened to me more than a few times. Probably not panel specific - this is just the first crash I've encountered in quite a while.

foxymiles (approved):
I feel like we can find a better solution to this but let's go with it for now to fix the crash.

packages/studio-base/src/typings/web.d.ts
```diff
@@ -0,0 +1,9 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+interface WindowOrWorkerGlobalScope {
+  structuredClone<T>(value: T, options?: StructuredSerializeOptions): T;
+}
+
+declare function structuredClone<T>(value: T, options?: StructuredSerializeOptions): T;
```
defunctzombie, line 9: Apparently available in typescript 4.7 once we upgrade to that. FYI.

PR #3548, opened 2022-06-09
Hide floating arrow button on image panel when nothing is selected

foxymiles:
**User-Facing Changes**
This hides the floating arrow button on the image panel unless an annotation has been selected.

**Description**
The floating arrow button is distracting and takes up space. We can hide it until the user clicks on an annotation.


jhurliman (approved):

2metres:
I'm not a fan of this change. I feel like it makes finding the tool nearly impossible.

If we did want to remove the icon possibly having a keyboard modifier that changes the cursor would be better way to expose this also having some sort of "zoomed" view of the pixels would make point selection easier.

amacneil:
If this is contentious, let's wait until we are ready to build the new 3d-based image panel, and figure out a paradigm that works for both of them.

PR #3549, opened 2022-06-09
Add help to panel context menu

foxymiles:
**User-Facing Changes**
This adds a help item to the panel context menu.

**Description**
We removed the help button from the panel toolbar in https://github.com/foxglove/studio/pull/3538 to reduce clutter and make room in the toolbar for more important items. But we can use the context menu to give the user a convenient shortcut to the panel help.

Screen Shot 2022-06-09 at 9 24 13 AM

### DIFF
--- a/packages/studio-base/src/components/PanelToolbar/PanelActionsDropdown.tsx
+++ b/packages/studio-base/src/components/PanelToolbar/PanelActionsDropdown.tsx
@@ -13,7 +13,7 @@
 
 import { ContextualMenu, IContextualMenuItem, useTheme } from "@fluentui/react";
 import SettingsIcon from "@mui/icons-material/Settings";
-import { useCallback, useContext, useMemo, useRef } from "react";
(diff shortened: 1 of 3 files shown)


defunctzombie:
I'm probably slightly -1 on this change. Maybe I've tuned it out - but I don't think many apps I use put as much help icons all over the place as we've been doing. Users can access the docs through a sidebar icon. We've gone through and removed the obvious ones now which is good so I'm questioning if we really want to be adding them back. I think we should keep app space focused on the day-to-day actions a user will take rather than constantly reminding them of help.

jhurliman (changes requested):
Requested change: fewer help icons

foxymiles:
@defunctzombie I guess one advantage of putting it here is that it takes you directly to the specific panel docs instead of forcing you to manually navigate to that page in the sidebar. Not sure if that justifies a permanent place in the context menu or not.

PR #3552, opened 2022-06-09
Add callService api to the panel extension interface

defunctzombie:
**User-Facing Changes**
The callService api is available to extension panels when a data source that supports service calls is active. The data source determines how to process the callService call. The call returns a promise that resolves with the response from the call. The response is data source specific.

Fixes: #2038


packages/studio-base/src/players/RosbridgePlayer.ts
```diff
@@ -51,6 +51,10 @@ function isClockMessage(topic: string, msg: unknown): msg is { clock: Time } {
   return topic === "/clock" && maybeClockMsg.clock != undefined && !isNaN(maybeClockMsg.clock.sec);
 }
 
+function isRecord(value: unknown): value is Record<string, unknown> {
```
jtbandes, line 54: Seems identical to lodash isObjectLike. There is also isObject and isPlainObject.

packages/studio-base/src/players/RosbridgePlayer.ts
```diff
@@ -580,6 +584,41 @@ export default class RosbridgePlayer implements Player {
...
+    }
+
+    // Query the type name for this service.
+    const serviceType = await new Promise<string>((resolve, reject) => {
+      this._rosClient!.getServiceType(
+        service,
+        (type: string) => resolve(type),
+        (error: Error) => reject(error),
+      );
+    });
```
jtbandes, line 603: Could be quite slow to do this on each call. Do we want to cache this? Or is it safer to do each time in case it changes?

packages/studio-base/src/players/types.ts
```diff
@@ -297,6 +300,9 @@ export const PlayerCapabilities = {
   // Publishing messages. Need to be connected to some sort of live robotics system (e.g. ROS).
   advertise: "advertise",
 
+  // Calling ROS services
```
jtbandes, line 303:
```suggestion
  // Calling services
```

jtbandes (approved):
Seems like a good, not too invasive start for this feature

packages/studio-base/src/players/RosbridgePlayer.ts
defunctzombie: 👍 Will change over to those if they have the right typescript narrowing types

packages/studio-base/src/players/RosbridgePlayer.ts
defunctzombie: The thinking for now was that service calls happen on things like user input but I could see this being somewhat slow. I will add a comment above this with the tradeoff for now and we can come along and adjust this once there are some users.

packages/studio-base/src/players/RosbridgePlayer.ts
```diff
@@ -580,6 +584,41 @@ export default class RosbridgePlayer implements Player {
...
+
+    // Create a proxy object for dispatching our service call
+    const proxy = new roslib.Service({
+      ros: this._rosClient,
+      name: service,
+      serviceType,
+    });
```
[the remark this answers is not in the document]
defunctzombie: I'll double check - I think this is a wrapper that is cheap.

defunctzombie:
> Seems like a good, not too invasive start for this feature

My thinking as well - it is minimal surface area and as there are users we can tweak the internals as needed

packages/studio-base/src/players/RosbridgePlayer.ts
jhurliman: I think this will make a remote call to the `/rosapi/service_type` service, so it wouldn't be fast enough to keep up with user input if they are typing / adjusting a slider / etc.

packages/studio-base/src/players/RosbridgePlayer.ts
defunctzombie: I looked at those and they don't do the typescript narrowing thing I need done here with `value isRecord` because the API for rosbridge we've defined to require a `Record<string, unknown>`. We could write more AJV validation on the value but that seems maybe overkill for now?

packages/studio-base/src/players/RosbridgePlayer.ts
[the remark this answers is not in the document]
defunctzombie: Looks like a cheap wrapper https://github.com/RobotWebTools/roslibjs/blob/develop/src/core/Service.js

packages/studio-base/src/players/RosbridgePlayer.ts
defunctzombie: I've added a caching layer for service type lookups.

rgov:
I implemented a panel that uses this to communicate with `topic_tools/mux` today and it worked great!

PR #3554, opened 2022-06-09
Switch from @foxglove/mcap to @mcap/core

defunctzombie:
**User-Facing Changes**
None

**Description**
@mcap/core is the released package on npm.


defunctzombie:
> m👏

m🧢

PR #3563, opened 2022-06-10
Add support for cyclical types

defunctzombie:
**User-Facing Changes**

**Description**

### DIFF
--- a/packages/mcap-support/src/protobufDefinitionsToDatatypes.ts
+++ b/packages/mcap-support/src/protobufDefinitionsToDatatypes.ts
@@ -44,6 +44,7 @@ export function protobufDefinitionsToDatatypes(
   type: protobufjs.Type,
 ): void {
   const definitions: RosMsgField[] = [];
+  datatypes.set(stripLeadingDot(type.fullName), { definitions });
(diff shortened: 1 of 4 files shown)


defunctzombie:
> Change seems straightforward – a little unfortunate that we have to fix it in 3-4 separate places but I guess that's just how these things go.

MessagePathInput is a dark place. But yea at the end of the day not _that_ surprising that a fundamental change to what was previously an invariant required a few tweaks.

packages/mcap-support/src/protobufDefinitionsToDatatypes.ts
```diff
@@ -44,6 +44,7 @@ export function protobufDefinitionsToDatatypes(
   type: protobufjs.Type,
 ): void {
   const definitions: RosMsgField[] = [];
+  datatypes.set(stripLeadingDot(type.fullName), { definitions });
```
defunctzombie, line 47:
```suggestion
  // The empty list reference is added to the map so a `.has` lookup below can prevent infinite recursion on cyclical types
  datatypes.set(stripLeadingDot(type.fullName), { definitions });
```

PR #3564, opened 2022-06-10
Allow Stack component to have forwardRef

2metres:
**Description**
Allow Stack to have forwardRef


foxymiles (approved):
I needed this just the other day in fact.

PR #3565, opened 2022-06-10
Fix image panel zoom to fit / zoom to fill

### LINKED ISSUE #3560: Image view zoom to fit doesn't always work.
**Description**
The "zoom to fit" button in the image panel zoom menu doesn't always work.

- OS: macOS
- Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native): nuscenes
- ROS distro (if applicable): -

**Steps To Reproduce**
1. Open sample data
2. Create a new layout
3. Add an image panel
4. Zoom out as far as possible
5. Select "zoom to fit" from the zoom menu

**Expected Behavior**
The image should zoom to fit.

**Actual Behavior**
Nothing happens.

Note that if you zoom slightly after doing zoom to fit it seems to be zooming from a full zoom which suggests that the zoom is being set but it's not triggering a re-render of the image.

foxymiles:
**User-Facing Changes**
This fixes the image panel zoom to fit / zoom to fill function.

**Description**
Our recent conversion of the image panel `ZoomMenu` to MUI removed a state variable that was used to track the open state of the zoom menu:

```typescript
  const [openZoomContext, setOpenZoomContext] = useState(false);
```

Without this state variable which changed as the menu was opened and closed, changing zoom and pan alone didn't change component state in `ImageCanvas` so it didn't force a re-render.

The solution here is just to force a re-render on `resetPanZoom`.

Fixes #3560

### DIFF
--- a/packages/studio-base/src/panels/Image/components/ImageCanvas.tsx
+++ b/packages/studio-base/src/panels/Image/components/ImageCanvas.tsx
@@ -327,6 +327,8 @@ export function ImageCanvas(props: Props): JSX.Element {
   const resetPanZoom = useCallback(() => {
     setPan({ x: 0, y: 0 });
     setZoom(1);
+    // We have to force an update here becase pan & zoom are not state
+    // variables and setting them alone will not trigger a re-render.
     forceUpdate();
   }, [setPan, setZoom]);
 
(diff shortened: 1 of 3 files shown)


defunctzombie:
Why does the action of zooming to fit / fill not cause the panel to re-render? Is there no variable or config tracking the target zoom state that causes a re-render?

defunctzombie (approved):
Let's add a comment about the forceUpdate call since its a bit of an anti-pattern in react to force an update.

FYI - I think there's a `forceUpdate` in react-use. If there isn't maybe a time to add one to our `@foxglove/den` package.

foxymiles:
I added a comment to explain why the forced update is necessary.

I agree this is something we should generally try to avoid. Maybe making it into a hook would only encourage something we consider an anti-pattern?

PR #3557, opened 2022-06-09
[WIP] Cache layouts for better layout browser user experience

foxymiles:
**User-Facing Changes**
None

**Description**
This loads layouts on startup and then maintains a cache so that we don't have to reload them all every time the user switches to the layout browser.

We may also want to revive https://github.com/foxglove/studio/pull/3489 as a companion to this.

Fixes https://github.com/foxglove/studio/issues/3274

PR #3571, opened 2022-06-10
Add cache for layout storage

### LINKED ISSUE #3274: Loading layouts sometimes takes a long time

jtbandes:
**User-Facing Changes**
Improved performance of the Layouts sidebar with more caching.

**Description**
Alternative to #3557
Fixes #3274
Revival of/successor to code added in #1206 & removed in #1679

Marking as draft because we need to discuss the implications on using multiple Studio windows/tabs/instances. Currently the cache will not be reloaded during the whole session, which means changes in the underlying storage from another instance of the app will not be picked up. Is this acceptable?


foxymiles (approved):

defunctzombie:
This is a draft PR - is this the final form or ready for review?

defunctzombie:
Do we currently handle changing across multiple instances of the app?

I'm tempted to say that it is acceptable for now in the interest of improving this experience. I think our general strategy for data-sync needs some thought but we don't need to tackle that here.

jtbandes:
Currently changes across multiple instances are not explicitly handled, but because we occasionally reload the whole cache, changes will eventually be picked up (not changes to your currently selected layout, but if you switch layouts)

defunctzombie:
> Currently changes across multiple instances are not explicitly handled, but because we occasionally reload the whole cache, changes will eventually be picked up (not changes to your currently selected layout, but if you switch layouts)

Could we still re-load the whole cache in the background? Just not make that a blocker to interactive use?

PR #3572, opened 2022-06-10
Align settings node action menu items without icons

foxymiles:
**User-Facing Changes**
None

**Description**
This aligns settings node action menu items with & without icons consistently and also modifies the default story so that the action menu is displayed.

Screen Shot 2022-06-10 at 1 38 07 PM

### DIFF
--- a/packages/studio-base/src/components/SettingsTreeEditor/index.stories.tsx
+++ b/packages/studio-base/src/components/SettingsTreeEditor/index.stories.tsx
@@ -3,6 +3,7 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
 import { Box } from "@mui/material";
+import { fireEvent } from "@testing-library/dom";
 import produce from "immer";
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/components/SettingsTreeEditor/index.stories.tsx
```diff
@@ -525,6 +527,10 @@ export function Basics(): JSX.Element {
   return <Wrapper roots={BasicSettings} />;
 }
 
+Basics.play = () => {
```
defunctzombie, line 530: Nifty

defunctzombie (approved):

amacneil:
How do we feel about icons in the menu at all? Seems like we could just remove them altogether.

foxymiles:
> How do we feel about icons in the menu at all? Seems like we could just remove them altogether.

Feels like having them as an option makes sense right? The uses I've seen for them so far seem to be positive.

PR #3573, opened 2022-06-10
Update demo bag & range support check

bryfox:
**User-Facing Changes**

- Serves the demo bag from our CDN
- Changes the check for range support in the browser reader

**Description**

The second change here requires some explanation and review.

The remote reader's setup ensures that a server advertises support for range requests:

https://github.com/foxglove/studio/blob/bf6aed9ce5b2ddbe522dac8b419bc0f1fdb6d029/packages/studio-base/src/util/BrowserHttpReader.ts#L56

The behavior I observe is that (1) once the bag is in its HTTP cache, Chrome will add `range` and `if-range` headers to this initial request; (2) the proxy server (outside of our control) omits the `accept-ranges` header iff the GET request includes a `range` header, causing bag loading to fail. (Compare the behavior between the dev deployments on the commits here.)

`no-store` forces a remote request, but in this case ensures that we receive the desired header. `no-store` also prevents storing the response, but we cancel the request immediately.

### DIFF
--- a/packages/studio-base/src/dataSources/SampleNuscenesDataSourceFactory.ts
+++ b/packages/studio-base/src/dataSources/SampleNuscenesDataSourceFactory.ts
@@ -30,8 +30,7 @@ class SampleNuscenesDataSourceFactory implements IDataSourceFactory {
   }
 
   initialize(args: DataSourceFactoryInitializeArgs): ReturnType<IDataSourceFactory["initialize"]> {
-    const bagUrl =
-      "https://storage.googleapis.com/foxglove-public-assets/nuScenes-v1.0-mini-scene-0061.bag";
+    const bagUrl = "https://assets.foxglove.dev/nuScenes-v1.0-mini-scene-0061.bag";
 
     if (this.enableIterablePlayer) {
       const bagSource = new BagIterableSource({ type: "remote", url: bagUrl });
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/dataSources/SampleNuscenesDataSourceFactory.ts
```diff
@@ -30,8 +30,7 @@ class SampleNuscenesDataSourceFactory implements IDataSourceFactory {
   }
 
   initialize(args: DataSourceFactoryInitializeArgs): ReturnType<IDataSourceFactory["initialize"]> {
-    const bagUrl =
-      "https://storage.googleapis.com/foxglove-public-assets/nuScenes-v1.0-mini-scene-0061.bag";
+    const bagUrl = "https://assets.foxglove.dev/nuScenes-v1.0-mini-scene-0061.bag";
```
defunctzombie, line 33: How do we feel about changing out sample data source to an mcap file rather than a bag file?

@amacneil

defunctzombie: Not that we don't need to do that in this PR - I just thought of it while looking at this PR.

defunctzombie (approved):

packages/studio-base/src/dataSources/SampleNuscenesDataSourceFactory.ts
amacneil: I think once @jhurliman finishes making our updated nuscenes dataset it will all be mcap. We can change it then.

PR #3575, opened 2022-06-10
Support dividers in settings action menus

foxymiles:
**User-Facing Changes**
None

**Description**
This adds support for dividers in settings node action menus.

Screen Shot 2022-06-14 at 6 54 40 AM


defunctzombie:
Would be good to also track aligning the text for icon and no-icon entries if we don't have a ticket for that yet.

2metres (approved):

PR #3578, opened 2022-06-12
Remove unused welcome panel

defunctzombie:
**User-Facing Changes**
Layouts which contained the welcome panel will instead say "Unknown panel type: welcome" and the user will need to remove the empty panel or change it to another.

Since the welcome panel is not important for day-to-day use of anyon's debugging I don't think its a bad breakage for any layouts that might still reference it.

**Description**
So long - you served us well.


amacneil (approved):

packages/studio-base/src/panels/ThreeDimensionalViz/CameraInfo/validate.ts
```diff
@@ -0,0 +1,102 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { ValidationResult } from "@foxglove/studio-base/util/validators";
```
amacneil, line 5: Bad merge? Can’t tell why this is here

defunctzombie: After removing the WelcomePanel, there was an unused method in utils/validation (isEmail). I removed that method and realized that most of the other stuff in there was only used by CameraInfo so I moved it to the CameraInfo folder. Once we delete the old 3d panel it will get swept up in that cleanup.

amacneil: 🔪  utils

PR #3584, opened 2022-06-13
[New3D] Add 2d view mode

belopot:
**User-Facing Changes**
In order to active 3D viewport mode, user need to off the perspective of Camera in panel setting.
image


packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -476,24 +504,40 @@ export class Renderer extends EventEmitter<RendererEvents> {
...
+      this.perspectiveCamera.fov = THREE.MathUtils.radToDeg(cameraState.fovy);
+      this.perspectiveCamera.near = cameraState.near;
+      this.perspectiveCamera.far = cameraState.far;
+
+      this.camera = this.perspectiveCamera;
+    } else {
+      //2D view mode
+      this.orthographicCamera.position
+        .setFromSpherical(tempSpherical.set(cameraState.distance, 0, 0))
+        .applyAxisAngle(UNIT_X, PI_2);
```
jtbandes, line 533: I don't think we really need spherical coordinates here, you should just be able to set x=distance (or something like that)

jtbandes: Does rotation work in 2D mode (right-click + drag)? I think you need to apply `cameraState.thetaOffset` to make this work (rotation in the plane of the screen).

Similar to this existing code:
```ts
      this.perspectiveCamera.quaternion.setFromEuler(
         tempEuler.set(cameraState.phi, 0, -cameraState.thetaOffset, "ZYX"),
       );
```
but without `cameraState.phi`

belopot: Like 3D view, should we allow to rotate camera by right mouse drag in 2d mode as well?

jtbandes: Yes, you can compare with the old panel to see an example of the expected behavior

packages/@types/foxglove__regl-worldview/index.d.ts
```diff
@@ -140,6 +141,12 @@ declare module "@foxglove/regl-worldview" {
     fovy: number;
     near: number;
     far: number;
+    orthoDistance: number;
+    orthoTargetOffset: Vec2;
+    orthoThetaOffset: number;
+    orthoFrustum: number;
+    orthoNear: number;
+    orthoFar: number;
```
jhurliman, line 149: You modified the type definition for `CameraState` from the `@foxglove/regl-worldview` package, but haven't modified that package. All of these properties will be undefined.

packages/studio-base/src/dataSources/SampleNuscenesLayout.json
```diff
@@ -45,19 +45,36 @@
...
+        "far": 5000,
+        "orthoDistance": 1000,
+        "orthoTargetOffset": [
+          0,
+          0
+        ],
+        "orthoThetaOffset": 0,
+        "orthoFrustum": 100,
+        "orthoNear": 0.01,
+        "orthoFar": 5000
```
jhurliman, line 73: Instead of adding a full separate set of config options for the second camera, please share settings where possible like the 2D camera in the old 3D panel.

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -476,24 +504,42 @@ export class Renderer extends EventEmitter<RendererEvents> {
...
+    } else {
+      console.log(cameraState);
+      //2D view mode
+      this.orthographicCamera.position.add(
+        tempVec.set(
+          cameraState.orthoTargetOffset[0],
+          cameraState.orthoTargetOffset[1],
+          cameraState.orthoDistance,
+        ),
+      );
```
jhurliman, line 538: It looks like this currently supports panning, but not zooming or rotating.

defunctzombie:
Let's avoid changing the sample layout with our new3d work if we can.

packages/studio-base/src/dataSources/SampleNuscenesLayout.json
```diff
@@ -45,19 +45,36 @@
...
+          3.552551884097018,
+          0.17283719439514628,
+          0
+        ],
         "thetaOffset": 0.8495002225312027,
         "fovy": 0.7853981633974483,
         "near": 0.01,
-        "far": 5000
+        "far": 5000,
+        "orthoDistance": 1000,
```
jhurliman, line 65: `SampleNuscenesLayout.json` only contains old 3D panels, it does not have any new 3D panels (`ThreeDeeRender`). Changing this layout will have no effect.

packages/studio-base/src/components/SettingsTreeEditor/inputs/Vec2Input.tsx
```diff
@@ -0,0 +1,57 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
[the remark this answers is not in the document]
belopot: Here is the PR
https://github.com/foxglove/studio/pull/3620

packages/@types/foxglove__regl-worldview/index.d.ts
belopot: I removed the items and add only one item : frustum. How to update the package?

packages/studio-base/src/dataSources/SampleNuscenesLayout.json
belopot: Yes

Issue #3586, opened 2022-06-13
Migrate database icon to sidebar settings


amacneil:
**Description**
Follow up from https://github.com/foxglove/studio/pull/3539:

Remove all remaining database icons from panel titlebars, and move this configuration to the settings sidebar
<img width="276" alt="image" src="https://user-images.githubusercontent.com/637671/173405120-43f33e0f-368b-41b8-91d5-03f015d71fd3.png">

- [x] Diagnostic Summary
- [x] Diagnostic Details
- [x] Log

PR #3587, opened 2022-06-13
Move log panel topic selection to settings

foxymiles:
**User-Facing Changes**
This moves the topic selection dropdown in the log panel toolbar into the settings sidebar and removes the item count from the toolbar.

**Description**
This moves the topic selection dropdown in the log panel toolbar into the settings sidebar. 

Screen Shot 2022-06-14 at 7 07 23 AM

Screen Shot 2022-06-13 at 12 52 16 PM

Partially addresses #3586


defunctzombie:
What does "not available"? I did a double take and am not sure what I'm suppose to do with that information.

foxymiles:
> What does "not available"? I did a double take and am not sure what I'm suppose to do with that information.

Just trying to preserve the existing behavior. I guess the idea was to show the chosen topic even if it's not available in the current datasource?

defunctzombie:
> Just trying to preserve the existing behavior. I guess the idea was to show the chosen topic even if it's not available in the current datasource?

Yea.

Here's what I think would be good behavior.

* When there's no data source loaded show `/rosout` (or whatevever topic the user configured)
* When a data-source is loaded but the user's topic is not available show a field error (red dot) with text indicating such

The dropdown itself should not contain any extra words like `(not available)` and should only have the topic.

defunctzombie (approved):
Before merge update the PR description or screenshots since the UX changed so we get a nice trail for the changelog.

PR #3590, opened 2022-06-13
Don't show drag cursor on single panel layout

### LINKED ISSUE #3585: Hide drag cursor for single root panel
**Description**
Follow up from https://github.com/foxglove/studio/pull/3539:

If there is only one top level panel, it is already non draggable. Therefore, we should hide the drag cursor.

foxymiles:
**User-Facing Changes**
None

**Description**
This only shows the grab cursor on panel toolbars if they're draggable, i.e. not the only panel in the layout.

Fixes #3585


jtbandes (approved):

amacneil:
Possibly related: https://github.com/foxglove/studio/issues/3632

PR #3595, opened 2022-06-13
Don't run lipo during afterPack for universal arch

jtbandes:
**User-Facing Changes**
None

**Description**
`yarn package:darwin --universal` no longer fails.
Fix for an issue introduced by https://github.com/electron-userland/electron-builder/pull/6887


defunctzombie (approved):
Maybe a comment in the code referencing the electron-builder issue

jtbandes:
I think the code is fairly self explanatory – we can't run `lipo -extract universal` because that command is invalid.

PR #3600, opened 2022-06-14
Allow undefined elements in settings trees

foxymiles:
**User-Facing Changes**
None

**Description**
This allows undefined elements in settings tree nodes & fields. This just makes life a little bit easier for panel authors that need to optionally exclude elements of the tree. Undefined elements will just be ignored by the settings tree editor.

### DIFF
--- a/packages/studio-base/src/components/SettingsTreeEditor/NodeEditor.tsx
+++ b/packages/studio-base/src/components/SettingsTreeEditor/NodeEditor.tsx
@@ -28,7 +28,7 @@ import Stack from "@foxglove/studio-base/components/Stack";
 import { FieldEditor } from "./FieldEditor";
 import { NodeActionsMenu } from "./NodeActionsMenu";
 import { VisibilityToggle } from "./VisibilityToggle";
-import { SettingsTreeAction, SettingsTreeNode } from "./types";
+import { SettingsTreeAction, SettingsTreeField, SettingsTreeNode } from "./types";
 
 export type NodeEditorProps = {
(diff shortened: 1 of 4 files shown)


packages/studio-base/src/components/SettingsTreeEditor/NodeEditor.tsx
```diff
@@ -145,23 +145,29 @@ function NodeEditorComponent(props: NodeEditorProps): JSX.Element {
   const { fields, children } = settings;
   const hasProperties = fields != undefined || children != undefined;
 
-  const fieldEditors = Object.entries(fields ?? {}).map(([key, field]) => {
-    const stablePath = makeStablePath(props.path, key);
-    return <FieldEditor key={key} field={field} path={stablePath} actionHandler={actionHandler} />;
-  });
+  const fieldEditors = Object.entries(fields ?? {})
+    .filter((kv): kv is [string, SettingsTreeField] => kv[1] != undefined)
```
defunctzombie, line 149: FYI we have a `filterMap` function in `@foxglove/den/collection` that will do a filter and map with one iteration of an array.

defunctzombie (approved):

packages/studio-base/src/components/SettingsTreeEditor/NodeEditor.tsx
foxymiles: Switched to `filterMap` for these. As a bonus it seems to have better type narrowing.

PR #3602, opened 2022-06-14
feat(MapPanel): Follow GPS on map update

dudv:
**User-Facing Changes**
Screenshot 2022-06-15 at 09 58 36

Allow users to follow a particular topic

**Description**
When the ego vehicle moves fast, it becomes hard to follow its position on the map. This feature makes the map easier to use for such use cases.


defunctzombie:
Which topic is followed? Seems like the `Follow` setting should let the user pick a topic to follow if there are multiple topics.

NIT: you can drop the `GPS` label from `Follow` since the entire panel is about GPS (navsat) data.

defunctzombie:
> It uses the same logic to get the center as previously - i.e. use the first encountered GPS message.

That was ok for getting the center once but would not work so well if there are two NavSat topics in different locations and the map is constantly bouncing around because the "first message" is a different topic each render cycle.

dudv:
Thanks, @defunctzombie, I've added a way to follow a particular topic.

defunctzombie (approved):

defunctzombie:
Thank you! This is a great feature addition.

PR #3605, opened 2022-06-14
Tidy up styling of diagnostic status panel

foxymiles:
**User-Facing Changes**
None

**Description**
This does some minor cosmetic polishing of the diagnostic status panel. Since we include this in the sample data layout it makes sense to make it look tidy.

Screen Shot 2022-06-14 at 12 21 01 PM
Screen Shot 2022-06-14 at 12 21 07 PM

### DIFF
--- a/packages/studio-base/src/components/CssBaseline.tsx
+++ b/packages/studio-base/src/components/CssBaseline.tsx
@@ -45,13 +45,6 @@ const useStyles = makeStyles((theme) => ({
     "b, strong": {
       fontWeight: 700,
     },
-    table: {
-      borderCollapse: "collapse",
-      borderSpacing: 0,
-    },
-    "th, td": {
-      verticalAlign: "top",
-    },
 
     // container styling
     height: "100%",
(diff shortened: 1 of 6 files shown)


defunctzombie:
Don't know if this was the behavior before but when I select a name in the toolbar dropdown, I still see the heading for that hardware id. Maybe that's cause our sample data only has name and no hardware id.

image

Something we can address in a follow-up if we think it useful - just wanted to mention it for reference.

defunctzombie (approved):
👍 nice

2metres:
> Don't know if this was the behavior before but when I select a name in the toolbar dropdown, I still see the heading for that hardware id. Maybe that's cause our sample data only has name and no hardware id.
> 
> image
> 
> Something we can address in a follow-up if we think it useful - just wanted to mention it for reference.

That still works, it's just our sample data that doesn't have a hardware ID see [storybook](https://603ec8bf7908b500231841e2-wbbxfyqmmx.chromatic.com/?path=/story/panels-diagnostics-diagnosticstatuspanel--moved-divider)

2metres (approved):

Issue #3143, opened 2022-04-04
[New3D] Remaining Tasks


jhurliman:
Remaining tasks to reach feature parity with the existing `ThreeDimensionalViz` panel.

- [ ] Picking, multiple selection UI control (https://github.com/foxglove/studio/pull/3132)
- [ ] Single-pass object highlighting
- [ ] Fix horizontal camera panning to match cursor position
- [ ] Coordinate frame UI control
- [ ] Labels (coordinate frames, text markers)
- [ ] Layer settings editing, topic/ns layer hierarchy
- [ ] Persist settings
- [ ] Topic/namespace visibility toggling
- [ ] Grid layer
- [ ] Search UI control
- [ ] 2D (orthographic) mode
- [ ] Measurement tool
- [ ] Existing debug UI control
- [ ] Camera UI control

defunctzombie:
Does picking include publishing?

PR #3607, opened 2022-06-14
Add didSeek to extension API

jtbandes:
**User-Facing Changes**
Added `didSeek` to RenderState in the extension API.

**Description**
Required for #3143  and #2003


packages/studio-base/src/components/PanelExtensionAdapter.tsx
```diff
@@ -175,6 +177,16 @@ function PanelExtensionAdapter(props: PanelExtensionAdapterProps): JSX.Element {
     // The render state stats with the previous render state and changes are applied as detected
     const renderState: RenderState = prevRenderState.current;
 
+    if (watchedFieldsRef.current.has("didSeek")) {
```
defunctzombie, line 180: As use of the extension api has evolved I wonder if we should limit watch fields to things that would otherwise cause frequent updates (i.e. watching the currentTime) and allow fields like didSeek to always be present?

defunctzombie (approved):

PR #3609, opened 2022-06-14
Add `simpleGetMessagePathDataItems` which doesn't require topics & datatypes

jtbandes:
**User-Facing Changes**
None

**Description**
Adds a `simpleGetMessagePathDataItems()` which performs the same path-execution functionality as `getMessagePathDataItems()` but without requiring topics and datatypes as input. This will be usable from extension panels (although we should remove the "RosPaths" stuff before we actually publish it with `@foxglove/studio`).

This is required for #2003 / #1985 in order to use the `input: 'messagepath'` from extension panels.

### DIFF
--- a/packages/studio-base/src/components/MessagePathSyntax/simpleGetMessagePathDataItems.ts
+++ b/packages/studio-base/src/components/MessagePathSyntax/simpleGetMessagePathDataItems.ts
@@ -7,13 +7,13 @@ import { MessageEvent } from "@foxglove/studio-base/players/types";
 import { RosPath } from "./constants";
 import { filterMatches } from "./filterMatches";
 
-export function simpleGetMessagePathDataItem(
+export function simpleGetMessagePathDataItems(
   message: MessageEvent<unknown>,
   filledInPath: RosPath,
-): unknown | undefined {
+): unknown[] {
   // We don't care about messages that don't match the topic we're looking for.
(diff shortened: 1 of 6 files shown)


packages/studio-base/src/components/MessagePathSyntax/simpleGetMessagePathDataItems.ts
```diff
@@ -0,0 +1,60 @@
...
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { MessageEvent } from "@foxglove/studio-base/players/types";
+
+import { RosPath } from "./constants";
+import { filterMatches } from "./filterMatches";
+
+// ts-prune-ignore-next
+export function simpleGetMessagePathDataItems(
```
defunctzombie, line 11: Comment here would be nice.

defunctzombie:
If this new function can perform the same path finding as `getMessagePathDataItems` why do we keep `getMessagePathDataItems` around?

jtbandes:
Not sure if it's totally necessary, but I don't want to tackle the refactor/deletion as part of #2003.

One thing that `getMessagePathDataItems` does is tell you which path a result came from when you have multiple results. The Plot panel uses this for tooltips (see `values[6]` etc):

image

defunctzombie (approved):

PR #3618, opened 2022-06-15
Enable studio extensions in the web app

foxymiles:
**User-Facing Changes**
This makes it possible to use install and use studio extensions in the web app.

**Description**
The implementation here rests on a few key points.
1. Instead of a single extension loader there is now a set of them, each managing an extension namespace. Currently the only namespaces are `local`, which corresponds to our current implementation, and `private` which will be used by the private extension registry.
2. The desktop app will continue to store local extensions in the local filesystem, to avoid requiring any migrations for desktop users.
3. The web app will store extensions as binary bundles in namespaced indexdb databases, with a separate db per namespace.
4. Panels loaded from extensions will now be referenced in layouts with a globally unique id consisting of the extension namespace, the extension's publisher, and the extension's package name: `{namespace}/{publisher}/{packageName}`.
5. To avoid breaking existing layouts using extension panels the layout id of extension panels loaded from the local namespace by the desktop app will continue to use the existing scheme of `{packageName}.{panelName}`.


foxymiles:
Here are some prebuilt extension bundles to test with.

[extension_bundles.zip](https://github.com/foxglove/studio/files/8912470/extension_bundles.zip)

defunctzombie:
For users trying to install public extensions in the web version we should avoid showing this error:

```
Failed to download extension foxglove.studio-extension-turtlesim. Failed to fetch: https://github.com/foxglove/studio-extension-turtlesim/releases/download/v0.0.1/foxglove.studio-extension-turtlesim-0.0.1.foxe
```

And instead show our current error which tells you that you can't install extensions on the web.

packages/studio-base/src/components/ExtensionsSidebar/index.stories.tsx
```diff
@@ -47,6 +48,7 @@ const marketplaceExtensions: ExtensionInfo[] = [
   {
     id: "publisher.storyextension",
     name: "storyextension",
+    qualifiedName: "storyextension",
```
jtbandes, line 51: this doesn't look like a qualified name?

packages/studio-base/src/services/IdbExtensionStorage.ts
```diff
@@ -0,0 +1,91 @@
...
+    log.debug("Storing extension", { extension });
+
+    const transaction = (await this.#db).transaction(
+      [METADATA_STORE_NAME, EXTENSION_STORE_NAME],
+      "readwrite",
+    );
+    await Promise.all([
+      transaction.db.put(METADATA_STORE_NAME, extension.info),
+      transaction.db.put(EXTENSION_STORE_NAME, extension),
+      transaction.done,
```
jtbandes, line 72: Hm, I guess it's probably fine, but it strikes me as weird to include tx.done here when the other tasks were just barely started. It means the order of items in the array matters, which usually isn't the case when using Promise.all. I would just move await tx.done after the Promise.all.

jtbandes (approved):

packages/studio-base/src/services/IdbExtensionStorage.ts
foxymiles: I was just following the example from the idb lib:
https://github.com/jakearchibald/idb#idbtransaction-enhancements

Maybe it's incorrect?

packages/studio-base/src/components/ExtensionsSidebar/index.stories.tsx
foxymiles: For a story it doesn't matter right?

packages/studio-base/src/services/IdbExtensionStorage.ts
jtbandes: I think it's correct, just depends on the order of evaluation of array literal elements 🤷

packages/studio-base/src/components/ExtensionsSidebar/index.stories.tsx
jtbandes: sure, I was just confused

defunctzombie:
Thoughts:

Going forward all installs use: `namespace?id` where namespace is `local`, `private`, `mktplace`. In the case where we sync with a registry, we use the _id_ the registry provides.

For `local` namespace. The _id_ comes from the package.json fields publisher and package name.

packages/studio-base/src/services/IdbExtensionLoader.ts
```diff
@@ -0,0 +1,94 @@
...
+      throw new Error("Invalid extension: missing package.json");
+    }
+
+    const rawInfo = validatePackageInfo(JSON.parse(pkgInfoText) as Partial<ExtensionInfo>);
+    const normalizedPublisher = rawInfo.publisher.toLowerCase().replace(/[\W_]+/g, "_");
+    const info: ExtensionInfo = {
+      ...rawInfo,
+      id: `${normalizedPublisher}.${rawInfo.name}`,
+      namespace: this.namespace,
+      qualifiedName: [this.namespace, normalizedPublisher, rawInfo.name].join("/"),
```
bryfox, line 78: In the case of scoped packages, `name` here differs semantically from desktop. e.g., if the package name was `@foxglove/foo`, then the name as parsed by desktop would be `foo` (and `rawInfo.publisher` need not be specified since it's scoped to `foxglove`). We should decide if this is what we want, since this will affect how a remote registry enforces uniqueness.

foxymiles: That's a good catch. I think I need to make both use the same parsing logic here.

foxymiles: Well taking a closer look we're actually currently putting the extension's `displayName` as the identifier in the layout:

https://github.com/foxglove/studio/blob/0ffb8aa473e8673d9630dda7db05a1d26026e5a2/desktop/renderer/services/DesktopExtensionLoader.ts#L65-L68

https://github.com/foxglove/studio/blob/0ffb8aa473e8673d9630dda7db05a1d26026e5a2/packages/studio-base/src/providers/ExtensionRegistryProvider.tsx#L51-L55

I'm duplicating this for the local namespace for desktop & web for now to preserve compatibility with existing layouts.

PR #3619, opened 2022-06-15
Fix crash on layouts with missing extension panels

foxymiles:
**User-Facing Changes**
None

**Description**
This fixes a crash in the `PanelToolbarControls` component when studio tries to load a layout with a missing extension panel.

### DIFF
--- a/packages/studio-base/src/components/PanelToolbar/PanelToolbarControls.tsx
+++ b/packages/studio-base/src/components/PanelToolbar/PanelToolbarControls.tsx
@@ -12,9 +12,9 @@
 //   You may not use this file except in compliance with the License.
 
 import SettingsIcon from "@mui/icons-material/Settings";
-import { useCallback } from "react";
+import { useCallback, useContext } from "react";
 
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/components/PanelToolbar/PanelToolbarControls.tsx
```diff
@@ -42,20 +42,23 @@ export const PanelToolbarControls = React.memo(function PanelToolbarControls({
...
     [panelId],
   );
 
   const hasSettings = usePanelSettingsEditorStore(hasSettingsSelector);
 
   const openSettings = useCallback(() => {
-    setSelectedPanelIds([panelId]);
-    openPanelSettings();
+    if (panelId) {
+      setSelectedPanelIds([panelId]);
```
defunctzombie, line 59: If `panelId` could be undefined how come the type for this function call didn't catch trying to pass an undefined item of an array?

defunctzombie (approved):

PR #3623, opened 2022-06-16
Move low hanging styled-components over to muiStyled

2metres:
**User-Facing Changes**
None

**Description**
Move less complicated styled-components over to use muiStyled

### DIFF
--- a/packages/studio-base/src/components/Tooltip.stories.tsx
+++ b/packages/studio-base/src/components/Tooltip.stories.tsx
@@ -11,13 +11,9 @@
 //   found at http://www.apache.org/licenses/LICENSE-2.0
 //   You may not use this file except in compliance with the License.
 
-import styled from "styled-components";
+import { Box } from "@mui/material";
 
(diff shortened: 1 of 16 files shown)


packages/studio-base/src/components/Tooltip.stories.tsx
```diff
@@ -11,45 +11,36 @@
...
   } as const;
+
   return (
-    <div style={containerStyle}>
+    <div style={{ padding: 100, display: "flex", width: 400, height: 300 }}>
       <Tooltip contents="Top" placement="top" shown>
         <div style={style} />
       </Tooltip>
-      <Spacer />
+      <Box width={10} />
```
defunctzombie, line 35: Didn't we determine Box was sad times?

defunctzombie (approved):

packages/studio-base/src/components/Tooltip.stories.tsx
2metres: Not in storybooks

PR #3629, opened 2022-06-16
Use local paths state in plot legend to improve interactivity

### LINKED ISSUE #3146: Keyboard input sporadically disappears

foxymiles:
**User-Facing Changes**
This improves interactivity of the plot legend message path input.

**Description**
Currently we immediately flush every edit to a plot message path input to  the panel config, and this results in studio doing a lot of work including updating subscriptions, rebuilding callbacks etc and re-rendering the entire plot on every keystroke.

The approach here is for the plot legend to maintain its own local copy of the paths and use a debounced flush to push this back to the underlying panel config.

It also adds a few missing memoizations.

Fixes #3146

### DIFF
--- a/packages/studio-base/src/panels/Plot/PlotLegend.tsx
+++ b/packages/studio-base/src/panels/Plot/PlotLegend.tsx
@@ -23,7 +23,8 @@ import { Button, IconButton, Theme, alpha } from "@mui/material";
 import { makeStyles } from "@mui/styles";
 import cx from "classnames";
 import { last } from "lodash";
-import { ComponentProps, useCallback, useMemo, useRef } from "react";
+import { ComponentProps, useCallback, useMemo, useRef, useState } from "react";
+import { useDebouncedCallback } from "use-debounce";
 
 import Dropdown from "@foxglove/studio-base/components/Dropdown";
 import DropdownItem from "@foxglove/studio-base/components/Dropdown/DropdownItem";
@@ -31,6 +32,7 @@ import MessagePathInput from "@foxglove/studio-base/components/MessagePathSyntax
 import { PANEL_TOOLBAR_MIN_HEIGHT } from "@foxglove/studio-base/components/PanelToolbar";
 import TimeBasedChart from "@foxglove/studio-base/components/TimeBasedChart";
 import PlotLegendRow from "@foxglove/studio-base/panels/Plot/PlotLegendRow";
+import { PanelConfig, SaveConfig } from "@foxglove/studio-base/types/panels";
 
 import { PlotPath, BasePlotPath, isReferenceLinePlotPathType } from "./internalTypes";
 import { plotableRosTypes, PlotConfig, PlotXAxisVal } from "./types";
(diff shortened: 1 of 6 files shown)


packages/studio-base/src/panels/Plot/PlotLegend.tsx
```diff
@@ -228,26 +230,42 @@ const useStyles = makeStyles((theme: Theme) => ({
...
+  const [localConfig, setlocalConfig] = useState({ paths: props.paths });
+
+  const debouncedSaveConfig = useDebouncedCallback((newConfig: Partial<PanelConfig>) => {
+    props.saveConfig(newConfig);
+  }, 500);
+
+  const saveConfig = useCallback(
+    (newConfig: Partial<PlotConfig>) => {
+      setlocalConfig((prevConfig) => ({ ...prevConfig, ...newConfig }));
+      debouncedSaveConfig(newConfig);
```
defunctzombie, line 262: Technically we are saving a different thing than what we are setting localConfig to here since `newConfig` might not equal `{ ...prevConfig, ...newConfig}`.

Maybe the call to debouncedSaveConfig should happen in a useEffect that depends on `localConfig` ?

defunctzombie (approved):

packages/studio-base/src/panels/Plot/PlotLegend.tsx
foxymiles: I think we're ok in that case because the semantics of `saveConfig` are to merge not replace.

But the other issue with this approach is that all updates to config are debounced. I think I'll rework this to just wrap and debounce the `paths` element of config because that's the only part that is affected by this issue.

PR #3636, opened 2022-06-17
Force panel to update when moved from root to non-root

### LINKED ISSUE #3632: After a panel stops being the root panel, it loses drag functionality
**Description**
1. Create a layout with single root panel. The panel is non-draggable, as expected per https://github.com/foxglove/studio/pull/3590
2. Add another panel.
3. The newly added panel is draggable, but the first panel is still non-draggable for some reason.

https://user-images.githubusercontent.com/637671/174187875-2806ccf9-fe89-458d-994a-5b09983bb915.mov

foxymiles:
**User-Facing Changes**
This fixes an issue with undraggable panels.

**Description**
The issue is that the panel component is memoized and when it changes from being the root panel to one panel of several in a layout nothing changes in the panel props so it doesn't update. The solution here is to put the path length in the panel key which forces it to update as its position in the layout hierarchy changes.

Fixes #3632

### DIFF
--- a/packages/studio-base/src/components/PanelLayout.tsx
+++ b/packages/studio-base/src/components/PanelLayout.tsx
@@ -153,7 +153,8 @@ export function UnconnectedPanelLayout(props: Props): React.ReactElement {
               </EmptyState>
             }
           >
-            <Panel childId={id} tabId={tabId} />
+            {/* We put the path length in the key here to force an update when the panel moves from root to non-root. */}
+            <Panel childId={id} tabId={tabId} key={`${id}${tabId}${path.length}`} />
           </Suspense>
         </MosaicWindow>
       );
--- a/packages/studio-base/src/components/PanelLayout.tsx
+++ b/packages/studio-base/src/components/PanelLayout.tsx
@@ -138,6 +138,10 @@ export function UnconnectedPanelLayout(props: Props): React.ReactElement {
         panelCompoentCache.current.set(type, Panel);
       }
 
+      // When a panel changes from being the only panel to one of many in a layout and
+      // is no longer the top level panel we need to force it to update to recalculate
+      // whether it should be draggable or not. Since the panel component is memoized we use
+      // a key to break through the memoization when the panel's layout path changes.
       const mosaicWindow = (
         <MosaicWindow
           title=""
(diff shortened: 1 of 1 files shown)


defunctzombie (approved):

jtbandes (commented):
A more direct way to solve this issue would've been to fix the Panel.tsx code so that `isTopLevelPanel` is actually a dependency that causes the `connect` functions to re-run. Currently the code is structured like:

```ts
const isTopLevelPanel = mosaicWindowActions.getPath().length === 0 && tabId == undefined;

<Something ...
    ref={(el) => {
      panelRootRef.current = el;
      // disallow dragging the root panel in a layout
      if (!isTopLevelPanel) {
        connectOverlayDragPreview(el);
        connectToolbarDragPreview(el);
      }
    }}
>
```

Because this code is inside a `ref={}` function, it only runs when the element is mounted/unmounted. The connect calls could be moved into a useEffect with `isTopLevelPanel` as a dependency, which would cause them to re-run at the right times.

jtbandes:
Though that's not quite right either – need to check if `useContext(MosaicWindowContext)` & `mosaicWindowActions.getPath()` are enough to trigger the panel to re-render.

foxymiles:
> A more direct way to solve this issue would've been to fix the Panel.tsx code so that `isTopLevelPanel` is actually a dependency that causes the `connect` functions to re-run. Currently the code is structured like:
> 
> ```ts
> const isTopLevelPanel = mosaicWindowActions.getPath().length === 0 && tabId == undefined;
> 
> <Something ...
>     ref={(el) => {
>       panelRootRef.current = el;
>       // disallow dragging the root panel in a layout
>       if (!isTopLevelPanel) {
>         connectOverlayDragPreview(el);
>         connectToolbarDragPreview(el);
>       }
>     }}
> >
> ```
> 
> Because this code is inside a `ref={}` function, it only runs when the element is mounted/unmounted. The connect calls could be moved into a useEffect with `isTopLevelPanel` as a dependency, which would cause them to re-run at the right times.

I tried to make this via a few different approaches but couldn't make it work.

jtbandes:
Fixed in https://github.com/foxglove/studio/pull/4410 by adding a new context provider to trigger a re-render with the mosaic path changes.

PR #3637, opened 2022-06-17
Automatically select newly swapped panels

### LINKED ISSUE #3625: Changing panel type results in "Loading panel settings"
**Description**
1. Open panel settings sidebar
2. Change the panel type
3. Panel settings gets stuck in "Loading panel settings" state

https://user-images.githubusercontent.com/637671/173993879-c025c2d7-ab81-4d56-b37d-9a9cc1d3e3e6.mov

foxymiles:
**User-Facing Changes**
None

**Description**
The issue is that when we swap a panel we don't update the selected panel id so the context's selected panel still points to the id of the previous panel that no longer exists and the settings sidebar interprets this as no settings. 

The solution in this PR is to automatically make the newly swapped panel the selected panel. Due to the way the reducer is set up we don't know the id of the new panel so we have to do a slightly hacky before and after diff to figure it out.

Fixes #3625

### DIFF
--- a/packages/studio-base/src/providers/CurrentLayoutProvider/index.tsx
+++ b/packages/studio-base/src/providers/CurrentLayoutProvider/index.tsx
@@ -1,7 +1,7 @@
 // This Source Code Form is subject to the terms of the Mozilla Public
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
-import { isEqual } from "lodash";
+import { difference, isEqual } from "lodash";
 import { useCallback, useEffect, useMemo, useRef, useState } from "react";
 import { getNodeAtPath } from "react-mosaic-component";
 import { useToasts } from "react-toast-notifications";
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/providers/CurrentLayoutProvider/index.tsx
```diff
@@ -308,7 +308,16 @@ export default function CurrentLayoutProvider({
       },
       splitPanel: (payload: SplitPanelPayload) => performAction({ type: "SPLIT_PANEL", payload }),
       swapPanel: (payload: SwapPanelPayload) => {
+        // Select the new panel. We don't know what the new panel id will be so we diff
+        // the panelIds of the old and new layout.
```
defunctzombie, line 312:
```suggestion
        // the panelIds of the old and new layout so we can select the new panel.
```

defunctzombie (approved):
Should we only select if the outgoing panel was selected? Or do we assume triggering the context menu selects a panel?

amacneil:
If I understand the question correctly, it is whether to show the new panel's settings sidebar _always_ or whether to only show them _if the outgoing panel had settings sidebar open_?

I feel like we should make minimal changes to the user's UI, so we should only select the new panel if the old panel was selected.

foxymiles:
> If I understand the question correctly, it is whether to show the new panel's settings sidebar _always_ or whether to only show them _if the outgoing panel had settings sidebar open_?
> 
> I feel like we should make minimal changes to the user's UI, so we should only select the new panel if the old panel was selected.

The old panel will always be selected in this case because the act of clicking on the "change panel" menu also happens to select it.

amacneil:
I think that makes sense. But if I have a different sidebar open (e.g. Variables or whatever), we shouldn't change that.

foxymiles:
> I think that makes sense. But if I have a different sidebar open (e.g. Variables or whatever), we shouldn't change that.

Unfortunately this code has no access to the state that controls the active sidebar item.

amacneil:
If the settings sidebar is not open, then the current panel is not "selected", right? At least there is no blue border in the UI.

Therefore if the old panel doesn't have a blue border, the new one also should not. Is that possible?

foxymiles:
> If the settings sidebar is not open, then the current panel is not "selected", right? At least there is no blue border in the UI.
> 
> Therefore if the old panel doesn't have a blue border, the new one also should not. Is that possible?

Yeah with this PR if the settings sidebar isn't already open then it won't open it.

PR #3640, opened 2022-06-17
Allow explicit sort order on settings nodes

foxymiles:
**User-Facing Changes**
None

**Description**
This allows panel authors to optionally override the natural sort order of settings tree nodes.

### DIFF
--- a/packages/studio-base/src/components/SettingsTreeEditor/types.ts
+++ b/packages/studio-base/src/components/SettingsTreeEditor/types.ts
@@ -152,6 +152,15 @@ export type SettingsTreeNode = {
   renamable?: boolean;
 
   /**
+   * Optional sort order to override natural object ordering. All nodes
+   * with a sort order will be rendered before nodes all with no sort order.
+   *
+   * Nodes without an explicit order will be ordered according to ES2015
+   * object ordering rules.
+   */
+  order?: number;
+
+  /**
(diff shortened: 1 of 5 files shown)


jhurliman (approved):

packages/studio-base/src/components/SettingsTreeEditor/types.ts
```diff
@@ -151,6 +151,15 @@ export type SettingsTreeNode = {
    */
   renamable?: boolean;
 
+  /**
+   * Optional sort order to override natural object ordering. All nodes
+   * with a sort order will be rendered before nodes all with no sort order.
+   *
+   * Nodes without an explicit order will be ordered according to ES2015
```
defunctzombie, line 158: Do we plan to describe this on our public api like this too? I'd like to see us come up with a better fall-back sort order cause this is not a very user-friendly way to describe the default sort order.

defunctzombie (approved):
For the public api we will need to re-visit how we describe the fallback sort order.

packages/studio-base/src/components/SettingsTreeEditor/types.ts
foxymiles: Yeah we should put this in more user friendly terms in the documentation. The short version is that they iterate in order of insertion unless they are numbers or stringified numbers in which case they're ordered by the numeric value of the key.

PR #3641, opened 2022-06-17
[New3D] Refactor renderables to use SceneExtension base class

jhurliman:
**User-Facing Changes**

None

**Description**

This is a major rewrite of how renderables are added to the scene in New3D. Two new base classes are added, `Renderable` and `SceneExtension`, to represent individual renderable entities and the managing class that receives data of some form and creates renderables. This groups all of the code necessary to subscribe to a topic, normalize incoming messages, create renderable entities, add sidebar nodes, receive sidebar edit events, update the config file, sidebar, and live renderable settings, into one place.


packages/studio-base/src/panels/ThreeDeeRender/settings.ts
```diff
@@ -2,375 +2,18 @@
...
-  distance: number;
-  color: string;
+  frameLocked?: boolean;
 };
 
-export type CustomLayerSettings = {
-  visible: boolean;
+export type CustomLayerSettings = BaseSettings & {
+  instanceId: string;
+  layerId: string;
```
jtbandes, line 14: Terminology is confusing me a bit – what is a "layer"? Is it the same as a "renderable"? Should we name this `renderableId`?

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -559,30 +587,65 @@ export class Renderer extends EventEmitter<RendererEvents> {
...
+    const instanceId = actionId.slice(-36);
+
+    const entry = this.customLayerActions.get(layerId);
+    if (!entry) {
+      throw new Error(`No custom layer action found for "${layerId}"`);
+    }
+
+    // Regenerate the action menu entry with a new instanceId
+    const { label, icon } = entry.action;
+    this.addCustomLayerAction({ layerId, label, icon, handler: entry.handler });
```
jtbandes, line 618: It kind of looks like we are adding a new custom action each time a custom action is invoked. That doesn't seem right, what am I missing?

packages/studio-base/src/panels/ThreeDeeRender/SceneExtension.ts
```diff
@@ -0,0 +1,127 @@
...
+export type RawMessage<T> = DeepPartial<T>;
+
+export type RawMessageEvent<T> = MessageEvent<DeepPartial<T>>;
+
+export class SceneExtension<
+  TRenderable extends Renderable<BaseUserData> = Renderable<BaseUserData>,
+> extends THREE.Object3D {
+  readonly extensionId: string;
+  readonly renderer: Renderer;
+  readonly renderables = new Map<string, TRenderable>();
```
jtbandes, line 27: What is the idea of the string keys here? Is this for when a user wants to add multiple instances of the scene extension's renderable (such as multiple grids)?

Should we introduce a new "brand" type for the key so that it can be clear which strings are supposed to be this "renderable id" (whatever it is)?

packages/studio-base/src/panels/ThreeDeeRender/SceneExtension.ts
```diff
@@ -0,0 +1,127 @@
...
+
+  // create settings nodes at arbitrary paths (or just under topics / layers?)
+
+  // handle user input at or beneath the registered settings node paths
+
+  // [x] startFrame, make it easy to update poses for all renderables
+
+  // [x] how do we track all renderables? traverse children?
+
+  // picking: enable/disable, provide custom shader(s), handle picking subcomponents
```
jtbandes, line 126: are these useful todo items or should they be deleted?

packages/studio-base/src/panels/ThreeDeeRender/SettingsManager.ts
```diff
@@ -0,0 +1,184 @@
...
+
+      curNode.error = this.errors.errors.errorAtPath(path);
+      return draft;
+    });
+
+    this.emit("update");
+  };
+}
+
+function removeNodeAtPath(root: SettingsTreeNode, path: Path): boolean {
```
jtbandes, line 119: Are these things going to be useful in other panels using the settings tree? Maybe they should be put in a common location that's not inside ThreeDeeRender

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
jhurliman: Yes, updating the comment to explain further

packages/studio-base/src/panels/ThreeDeeRender/color.ts
```diff
@@ -74,3 +74,21 @@ export function rgbaEqual(a: ColorRGBA, b: ColorRGBA): boolean {
     approxEquals(a.a, b.a)
   );
 }
+
+/**
+ * Computes a gradient step from colors `a` to `b` using pre-multiplied alpha to
+ * match CSS linear gradients.
+ */
+export function rgbaGradient(output: ColorRGBA, a: ColorRGBA, b: ColorRGBA, t: number): void {
```
jtbandes, line 82: Is it correct to say that this function takes in two colors which _do not_ have premultiplied alpha and returns a color that _does_ have premultiplied alpha? Hard to tell because the types are the same but the semantics are different.

packages/studio-base/src/panels/ThreeDeeRender/SceneExtension.ts
```diff
@@ -0,0 +1,127 @@
...
+import { Path } from "./LayerErrors";
+import { BaseUserData, Renderable } from "./Renderable";
+import type { Renderer } from "./Renderer";
+import type { SettingsTreeEntry } from "./SettingsManager";
+import { missingTransformMessage, MISSING_TRANSFORM } from "./renderables/transforms";
+import { updatePose } from "./updatePose";
+
+export type RawMessage<T> = DeepPartial<T>;
+
+export type RawMessageEvent<T> = MessageEvent<DeepPartial<T>>;
```
jtbandes, line 20: The name "RawMessage" doesn't convey much meaning here (and conflicts with the name of a panel), rename these something like PartialMessage or MaybePartialMessage?

packages/studio-base/src/panels/ThreeDeeRender/renderables/Cameras.ts
```diff
@@ -32,146 +52,126 @@ const CAMERA_MODEL = "CameraModel";
...
+    this._updateCameraInfoRenderable(
+      renderable,
+      cameraInfo,
+      receiveTime,
+      renderable.userData.settings,
+    );
+  };
 
-  _updateCameraInfoRenderable(
+  protected _updateCameraInfoRenderable(
```
jtbandes, line 170: why protected and not private?

packages/studio-base/src/panels/ThreeDeeRender/SceneExtension.ts
```diff
@@ -0,0 +1,127 @@
...
+
+export class SceneExtension<
```
jtbandes, line 22: It would be helpful to have documentation describing what needs to be overridden/extended in a subclass to successfully implement a new scene extension.

packages/studio-base/src/panels/ThreeDeeRender/SceneExtension.ts
jhurliman: Added comments

packages/studio-base/src/panels/ThreeDeeRender/SettingsManager.ts
jhurliman: It's not clear yet if there's a generally useful pattern here, or an optimization that is more tailored to the 3D panel architecture. Worth discussing.

packages/studio-base/src/panels/ThreeDeeRender/renderables/Images.ts
```diff
@@ -264,7 +281,7 @@ export class Images extends THREE.Object3D {
       const compressed = image as CompressedImage;
       const bitmapData = new Blob([image.data], { type: `image/${compressed.format}` });
       self
-        .createImageBitmap(bitmapData)
+        .createImageBitmap(bitmapData, { resizeWidth: DEFAULT_IMAGE_WIDTH })
```
jtbandes, line 284: I suppose this means larger images will lose some detail even when you zoom in on them in 3D?

packages/studio-base/src/panels/ThreeDeeRender/color.ts
jhurliman: Made the comment more verbose. The answer is yes, matching the behavior of CSS `linear-gradient`.

packages/studio-base/src/panels/ThreeDeeRender/renderables/Poses.ts
```diff
@@ -41,70 +54,127 @@ const DEFAULT_SETTINGS: LayerSettingsPose = {
...
+      }
+    }
+    return entries;
   }
 
-  dispose(): void {
-    for (const renderable of this.posesByTopic.values()) {
-      renderable.userData.arrow.dispose();
-      renderable.userData.sphere?.dispose();
+  handleSettingsAction = (action: SettingsTreeAction): void => {
```
jtbandes, line 135: I see a lot of renderable updating code happening in handleSettingsAction. Does this mean the only way a setting update will be reflected in the current scene is if it comes from a settings action? Supposing config changes can happen other ways (such as the user hand-editing config json – although maybe we removed that – or clicking some button in the panel that makes a config change), how do those get applied? Or does that just not happen today?

jhurliman: I had this thought earlier today. I think there's likely a bug/missing feature related to this. I need to come up with a failing test/story for this, because I currently don't understand what events to listen for (or hooks to call etc).

packages/studio-base/src/panels/ThreeDeeRender/settings.ts
jhurliman: It's a namespace or type name for a type of renderable the user can add to the scene, such as a Grid.

jhurliman: Added comments describing layers and these fields.

jtbandes (approved):

PR #3653, opened 2022-06-21
Only select swapped panel if original panel was selected

### LINKED ISSUE #3625: Changing panel type results in "Loading panel settings"

foxymiles:
**User-Facing Changes**
None

**Description**
This fixes a secondary issue caused by the fix for #3625. The change here is to only select the new panel if the panel we're swapping out was already selected.

Fixes #3625


amacneil (approved):
behavior seems good on this branch. there is a separate issue that happens only during screen recording, will file separate issue

PR #3654, opened 2022-06-21
Make settings api public and move to @foxglove/studio/settings package

foxymiles:
**User-Facing Changes**
None

**Description**
This moves the settings API types to the core `@foxglove/studio/settings` submodule and makes the `updatePanelSettingsTree` function in the extension panel API public.

I've put the settings related types in the submodule `@foxglove/studio/settings` submodule instead of directly in `@foxglove/studio` to reduce clutter in our top level module.


packages/studio-base/src/components/PanelExtensionAdapter.tsx
```diff
@@ -520,6 +517,8 @@ function PanelExtensionAdapter(props: PanelExtensionAdapterProps): JSX.Element {
       subscribeAppSettings: (settings: string[]) => {
         setSubscribedAppSettings(settings);
       },
+
+      updatePanelSettingsTree: updateSettings,
```
jtbandes, line 521: Maybe just call it updatePanelSettings? The fact that it's a tree isn't really relevant to other parts of the API.

packages/studio/index.d.ts
```diff
@@ -306,4 +312,256 @@ declare module "@foxglove/studio" {
...
+    /**
+     * True if the editor should show the filter control.
+     */
+    enableFilter?: boolean;
+
+    /**
+     * The actual settings tree roots. Updates to these will automatically be reflected in the
+     * editor UI.
+     */
+    roots: SettingsTreeRoots;
```
jtbandes, line 565: I don't like the naming of this field/type – most trees in the software world don't have multiple roots. It kinda feels like this should either be singular (`root: SettingsTreeRoot`) or just `tree: SettingsTree` because it _is_ the tree. The other fields here, `actionHandler` and `enableFilter`, are not part of the tree and this could be renamed to SettingsConfiguration or something like that.

packages/studio-base/src/components/PanelExtensionAdapter.tsx
foxymiles: The thing I didn't like about `updatePanelSettings` is that it suggests you're updating settings when in fact you're updating a representation of the settings. `updatePanelSettingsTree` does seem a bit clunky though.

packages/studio/index.d.ts
jtbandes: But I agree that SettingsConfiguration isn't much better than SettingsSettings 😅

jtbandes: yeah, you're right, we probably shouldn't introduce a third term for the same thing in our public API. I would vote for changing them all to "settings".

foxymiles: Yeah we need names for these three things:
1. the actual stored state
2. the mapping of that into the form that the editor users
3. the editor itself

foxymiles: And another wrinkle is that some panels use the `Config` from their props directly and others maintain their own copy and `saveState` when needed.

jtbandes: Settings, SettingsTree, SettingsEditor?

We plan to eventually remove Config from props, right? Since we will transition panels to using the public API.

foxymiles: Renamed `roots` ➡️  `nodes`.

jtbandes (approved):

jhurliman (approved):

packages/studio/index.d.ts
```diff
@@ -306,4 +312,256 @@ declare module "@foxglove/studio" {
...
+    | {
+        action: "perform-node-action";
+        payload: { id: string; path: readonly string[] };
+      };
+
+  export type SettingsTreeNodes = Record<string, undefined | SettingsTreeNode>;
+
+  /**
+   * A settings tree is a tree of panel settings that can be managed by
+   * a default user interface in Studio.
```
defunctzombie, line 548: What is a default user interface?

packages/studio/index.d.ts
```diff
@@ -306,4 +312,256 @@ declare module "@foxglove/studio" {
...
+   */
+  export type SettingsTree = {
+    /**
+     * Handler to process all actions on the settings tree initiated by the UI.
+     */
+    actionHandler: (action: SettingsTreeAction) => void;
+
+    /**
+     * True if the editor should show the filter control.
```
defunctzombie, line 557: Which editor?

packages/studio-base/src/components/PanelExtensionAdapter.tsx
foxymiles: Renamed this to `updatePanelSettingsEditor`, which feels more descriptive.

PR #3660, opened 2022-06-21
Improve panel key event handling for selection

### LINKED ISSUE #3657: macOS screen record causes Foxglove to get stuck in panel multi-select mode
**Description**
Today in "really weird bugs"

- open a layout
- hit cmd+shift+5 to show the screen record options
- hit esc to close
- clicking in the studio browser window is now stuck in multi-select mode
- the only way you can get out of this mode is by clicking out of the chrome window and then going back

In this video, about half way through I pull up the screen record options, and you can see we get stuck in panel multi-select mode. Then at the end I fix it by clicking away.

https://user-images.githubusercontent.com/637671/174845653-3836ff32-723d-4bc1-803e-00f509b48ded.mov

jtbandes:
**User-Facing Changes**
Fixed an issue where the app would get stuck in panel selection mode in certain situations (such as after taking a screenshot) on macOS.

**Description**
Fixes #3657
The issue is caused by a Chrome/Safari bug (https://bugs.chromium.org/p/chromium/issues/detail?id=1194004) which causes shift/cmd keyup events not to be delivered. However, it was easy to switch to e.shiftKey/e.metaKey instead of keeping track of the down/up state manually.

### DIFF
--- a/packages/studio-base/src/components/Panel.tsx
+++ b/packages/studio-base/src/components/Panel.tsx
@@ -199,8 +199,6 @@ export default function Panel<
     } = useCurrentLayoutActions();
 
     const [quickActionsKeyPressed, setQuickActionsKeyPressed] = useState(false);
-    const [shiftKeyPressed, setShiftKeyPressed] = useState(false);
-    const [cmdKeyPressed, setCmdKeyPressed] = useState(false);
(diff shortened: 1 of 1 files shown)


amacneil:
💜

PR #3667, opened 2022-06-22
[New3D] Call Renderer#flushState() + SceneExtension#flushState() on didSeek

jhurliman:
**User-Facing Changes**

None

**Description**

The UX becomes worse but we no longer show incorrect data after seeking in New3D. In the future, TF and Marker topics will be preloaded and cached to avoid the UX regression.


packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -274,6 +274,17 @@ export class Renderer extends EventEmitter<RendererEvents> {
     this.gl.dispose();
   }
 
+  /**
+   * Resets internal state such as the TransformTree and removes Renderables from SceneExtensions.
+   * This is useful when seeking to a new playback position or when a new data source is loaded.
+   */
+  flushState(): void {
```
jtbandes, line 281: Maybe call it something like `clear()` or `removeAllRenderables()`? "Flush" to me usually implies that items are being fully processed but not necessarily removed.

jtbandes (approved):

PR #3674, opened 2022-06-22
Animate sidebar opening/closing

jtbandes:
**User-Facing Changes**
Added animation when opening/closing the sidebar.

**Description**
The sidebar opening/closing is a major UI shift. An animation helps the eye track what is changing and reduce confusion.

https://user-images.githubusercontent.com/14237/175134214-666d2071-cce8-414b-8c90-6ce4bb31b507.mov

### DIFF
--- a/packages/studio-base/src/components/Sidebar/index.tsx
+++ b/packages/studio-base/src/components/Sidebar/index.tsx
@@ -10,10 +10,11 @@ import {
   ResizeGroup,
   ResizeGroupDirection,
 } from "@fluentui/react";
-import { Theme, useTheme } from "@mui/material";
+import { Theme } from "@mui/material";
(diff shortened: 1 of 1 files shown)


foxymiles:
Kind of mixed feelings about this. It looks nice but I wonder if it might become annoying for heavy users. I think I'd reduce the duration at least.

defunctzombie:
What's the status of this? Do we intend to push this along or do we want to close this out as not planned?

jtbandes:
I would like to pursue it. Now that we have moved off the old 3d panel, maybe that has resolved some of the flickering issues that were raised.

jtbandes:
That said, maybe with mixed feedback and more changes needed it is not worth the effort. 🤷

PR #3680, opened 2022-06-23
Allow string ordering values and sort new3d topics by topic name

foxymiles:
**User-Facing Changes**
None

**Description**
This allows sorting settings tree nodes by numeric or string values and uses string values to sort topics alphabetically in the new 3d panel settings.

Partially addresses #3663


jhurliman (approved):
Innovation

PR #3698, opened 2022-06-24
Refactor and centralize extension management

foxymiles:
**User-Facing Changes**
None

**Description**
This is a refactor in preparation for the private extension registry, with two goals:
1. Refactor extension management logic.
2. Route all extension installation and uninstallation through the `ExtensionRegistryProvider` so we can update studio as extensions are changed without doing a brute force reload of the entire app.

Since the `ExtensionRegistryProvider` is now the only direct client of the loaders the `ExtensionLoaderContext` has been subsumed and is no longer a separate context.

https://user-images.githubusercontent.com/93935560/175566810-01887d65-5922-4afa-bcfb-60a0e90cc7ab.mov


amacneil:
While we are refactoring all extension management logic, is it worth moving desktop extensions to use browser storage too (localstorage/indexeddb/whatever we are doing)? At this point in time we have relatively few desktop extension users and they probably would not mind migrating manually, and I assume it would simplify our code.

foxymiles:
@amacneil Yeah if we're willing to ask users to migrate then having one codebase would make things simpler. We'd need to adjust the way the `local-install` command works  for extensions but that shouldn't be difficult. 

Maintaining the existing filesystem implementation for desktop users isn't a lot of extra work right now either but I suppose over time it could be as we add features and make changes.

defunctzombie:
> At this point in time we have relatively few desktop extension users and they probably would not mind migrating manually, and I assume it would simplify our code.

We don't know that since most extension users are using private extensions. I would prefer we start with new desktop installs using the same storage as web and visit deprecating (or porting) existing installs as a follow-up.

defunctzombie:
> We'd need to adjust the way the `local-install` command works for extensions but that shouldn't be difficult.

What would we do for that? It is convenient to run that command when developing your extension.

packages/studio-base/src/providers/ExtensionRegistryProvider/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { ExtensionInfo, ExtensionNamespace } from "@foxglove/studio-base/types/Extensions";
+
+import { useExtensionLoaders } from "./loaders";
+import { useExtensionPanels } from "./panels";
+
+const log = Logger.getLogger(__filename);
+
+const NO_EXTENSIONS: ExtensionInfo[] = [];
+
+export default function ExtensionRegistryProvider(
```
defunctzombie, line 20: nit: I've got a small preference that we stop putting our classes in `index.ts` files and use files with the name of the class (like we do for other files that are not index.ts. index.ts can re-export if thats useful.

packages/studio-base/src/providers/ExtensionRegistryProvider/panels.ts
```diff
@@ -2,32 +2,28 @@
...
 const log = Logger.getLogger(__filename);
 
-export default function ExtensionRegistryProvider(props: PropsWithChildren<unknown>): JSX.Element {
-  const extensionLoader = useExtensionLoader();
-
-  const { value: registry, error } = useAsync(async () => {
-    const extensionList = await extensionLoader.getExtensions();
-    log.debug(`Found ${extensionList.length} extension(s)`);
-
+export function useExtensionPanels(
```
defunctzombie, line 17: should the filename by `useExtensionPanels` or does this file do something beyond provide this hook? Is it useful for this hook to live in our hooks folder or is this more a code-org within the RegistryProvider?

packages/studio-base/src/services/ExtensionLoader.ts
```diff
@@ -0,0 +1,22 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { ExtensionInfo, ExtensionNamespace } from "@foxglove/studio-base/types/Extensions";
+
+export interface ExtensionLoader {
```
defunctzombie, line 7: Worth documenting what an ExtensionLoader is suppose to be.

defunctzombie (approved):

amacneil:
> We don't know that since most extension users are using private extensions. I would prefer we start with new desktop installs using the same storage as web and visit deprecating (or porting) existing installs as a follow-up.

Yeah, I'm also fine with it being a follow up task. Just thought if it is easy to do now we could do so.

packages/studio-base/src/providers/ExtensionRegistryProvider/panels.ts
foxymiles: This is purely a code-org thing. I didn't name it `useX` to discourage other people from thinking they should use it outside the context of the RegistryProvider but I can change it if you think it's more clear.

packages/studio-base/src/providers/ExtensionRegistryProvider/index.tsx
foxymiles: I'm also not a fan of the `index.ts{x}` convention. I find it makes it harder to navigate code.

PR #3700, opened 2022-06-24
Fix bug with node playground scripts and custom datatypes

defunctzombie:
**User-Facing Changes**
Writing node playground scripts using the `./types` module and custom datatypes works even if the user has not separately subscribed to the input topic in another panel.

**Description**
Node playground scripts that used the ./types module and referenced custom datatypes did not work. They would fail to compile because the custom datatypes were missing from the Messages generic (i.e. Message<"my_msgs/Foo"> would be a compile error). Node playground player and transform worker logic would exclude any nodes that failed to compile when building a list of topics to subscribe to. Without subscribing to the input topics of the node script, the compilation step could never get the custom datatypes and the node would never compile successfully.

This change alters the behavior of UserNodePlayer to subscribe to the related input topics of an output topic from node scripts even if the node does not compile (but is able to identify the input topics). By subscribing to the input topics, the underlying player can provide the datatype information for the custom datatype which rebuilds the ./types module and allows the node script to compile successfully.

Prior to this change, a node which did not compile would not cause subscriptions to occur to its input topics since a node that doesn't compile cannot run. After this change, the underlying player will subscribe to "broken" nodes which may result in subscriptions to messages that are not processed by the node since the node is broken. We assume the desired state for node scripts is to be "working", we consider this an acceptable tradeoff to supporting custom datatypes.

### DIFF
--- a/packages/studio-base/src/panels/NodePlayground/BottomBar/index.tsx
+++ b/packages/studio-base/src/panels/NodePlayground/BottomBar/index.tsx
@@ -110,7 +110,6 @@ const BottomBar = ({ nodeId, isSaved, save, diagnostics, logs }: Props): ReactEl
           paddingRight={1}
         >
           <StyledTabs
-            disableRipple
             textColor="inherit"
             value={bottomBarDisplay !== "closed" ? bottomBarDisplay : false}
             onChange={handleChange}
--- a/packages/studio-base/src/players/UserNodePlayer/index.ts
+++ b/packages/studio-base/src/players/UserNodePlayer/index.ts
@@ -69,6 +69,12 @@ type UserNodeActions = {
   setUserNodeTypesLib: (lib: string) => void;
 };
 
+type NodeRegistrationCacheItem = {
+  nodeId: string;
+  userNode: UserNode;
+  result: NodeRegistration;
+};
+
 function maybePlainObject(rawVal: unknown) {
   if (typeof rawVal === "object" && rawVal && "toJSON" in rawVal) {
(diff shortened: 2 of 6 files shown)


packages/studio-base/src/panels/NodePlayground/BottomBar/index.tsx
```diff
@@ -110,7 +110,6 @@ const BottomBar = ({ nodeId, isSaved, save, diagnostics, logs }: Props): ReactEl
           paddingRight={1}
         >
           <StyledTabs
-            disableRipple
```
jhurliman, line 113: Bring back Ripple

packages/studio-base/src/players/UserNodePlayer/index.ts
```diff
@@ -677,6 +702,7 @@ export default class UserNodePlayer implements Player {
           this._nodeRegistrationCache = [];
         }
 
+        //console.log("new data", { shouldReset, activeData });
```
jhurliman, line 705: log.debug()?

foxymiles (approved):

packages/studio-base/src/players/UserNodePlayer/index.ts
defunctzombie: Actually meant to delete 🔪

packages/studio-base/src/panels/NodePlayground/BottomBar/index.tsx
defunctzombie: Amusingly this property isn't valid with this component and causes console log spew with react errors.

Issue #3645, opened 2022-06-20
bug: datatype not found: "builtin_interfaces/Time"


xmfcx:
**Description**
I get following error when I try the ROS2 connection.

![Screenshot from 2022-06-20 14-50-07](https://user-images.githubusercontent.com/10751153/174596383-106aa562-968d-4d0b-b813-82e792bff3d5.png)

**Steps To Reproduce**

Following the instructions on the https://github.com/foxglove/studio#contributing page,

- I've run: 
```
# To launch the desktop app (run both scripts concurrently):
$ yarn serve        # start webpack
$ yarn start        # launch electron
```
on different terminals. I've also sourced `ros2 galactic` in both terminals beforehand just to be safe.
- On the get started screen, selected `Open connection` option.
- Select ROS 2
- Enter Ros Domain ID (was 30 for me) (bug happens with 0 too)
- Click Open button

**Version:**
- OS: Ubuntu 20.04
- Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native): ROS2 Native
- ROS distro (if applicable): `galactic`

**Expected Behavior**
Error doesn't occur.

**Actual Behavior**

I get error: `datatype not found: "builtin_interfaces/Time"`

```
Call Stack
 eval
  packages/studio-base/src/components/MessagePathSyntax/messagePathsForDatatype.ts:108:19
 memoized
  node_modules/lodash/lodash.js:10621:27
 eval
  packages/studio-base/src/components/MessagePathSyntax/messagePathsForDatatype.ts:123:19
 eval
  packages/studio-base/src/components/MessagePathSyntax/messagePathsForDatatype.ts:110:33
 memoized
  node_modules/lodash/lodash.js:10621:27
 messagePathStructures
  packages/studio-base/src/components/MessagePathSyntax/messagePathsForDatatype.ts:135:36
 eval
  packages/studio-base/src/components/MessagePathSyntax/MessagePathInput.tsx:197:177
 updateMemo
  node_modules/react-dom/cjs/react-dom.development.js:15869:19
 Object.useMemo
  node_modules/react-dom/cjs/react-dom.development.js:16415:16
```

PR #3702, opened 2022-06-24
Add builtin_interfaces/Time and builtin_interfaces/Duration types to Ros2Player

defunctzombie:
**User-Facing Changes**
Ros2 native connections work again.

**Description**
https://github.com/foxglove/studio/pull/3532 updated foxglove/schemas. These new schemas reference the builtin_interfaces/Time type within the schema datatypes. This complex type was never provided explicitly in the Ros2Player provider datatypes. As a result, the messagePathsForDatatypes would error when trying to create a structure for fields having the builtin_interfaces/Time type.

This change fixes this behavior by manually adding these two builtin interfaces to the datatypes within Ros2Player.

Fixes: #3645

### DIFF
--- a/packages/studio-base/src/players/Ros2Player.ts
+++ b/packages/studio-base/src/players/Ros2Player.ts
@@ -131,6 +131,23 @@ export default class Ros2Player implements Player {
       name: "std_msgs/msg/Header",
       definitions,
     });
+
+    // Add Time and Duration builtin interfaces
+    this._providerDatatypes.set("builtin_interfaces/Time", {
+      name: "builtin_interfaces/Time",
+      definitions: [
+        { name: "sec", type: "int32" },
+        { name: "nanosec", type: "uint32" },
+      ],
+    });
+
+    this._providerDatatypes.set("builtin_interfaces/Duration", {
+      name: "builtin_interfaces/Duration",
+      definitions: [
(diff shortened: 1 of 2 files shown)


jhurliman (approved):

packages/studio-base/src/players/Ros2Player.ts
```diff
@@ -131,6 +131,23 @@ export default class Ros2Player implements Player {
       name: "std_msgs/msg/Header",
       definitions,
     });
+
+    // Add Time and Duration builtin interfaces
```
jhurliman, line 135: Should this be done in some upstream library?

defunctzombie: I could make a case that `generateRosMsgDefinition` in `@foxglove/schemas` should do it.

https://github.com/foxglove/studio/blob/main/packages/studio-base/src/players/Ros2Player.ts#L104

Possibly a similar case for our rosmsg-msg-common tho for those we'd need separate sets of RosDatatypes.

defunctzombie:
I am gonna merge this PR since ROS 2 native connections don't work if you have any panel that needs message path input until we land this bugfix.

If we decide that @foxglove/schemas should be the place to inject these datatype I can do that as a follow-on.

PR #3703, opened 2022-06-24
[New3D] Promote New3D to beta

jhurliman:
**User-Facing Changes**

- A new panel type, `3D (Beta)`, is available for testing

**Description**

Remove the experimental name and flag from New3D

### DIFF
--- a/packages/studio-base/src/panels/index.ts
+++ b/packages/studio-base/src/panels/index.ts
@@ -57,6 +57,15 @@ const builtin: PanelInfo[] = [
     module: async () => await import("./ThreeDimensionalViz"),
   },
   {
+    title: "3D (Beta)",
+    type: "3D",
+    description:
+      "Display visualization markers and models in a 3D scene using a new rendering engine.",
+    help: ThreeDeeRenderHelp,
+    thumbnail: threeDimensionalVizThumbnail,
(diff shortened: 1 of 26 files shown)


packages/studio-base/src/panels/index.ts
```diff
@@ -56,6 +56,15 @@ const builtin: PanelInfo[] = [
     thumbnail: threeDimensionalVizThumbnail,
     module: async () => await import("./ThreeDimensionalViz"),
   },
+  {
+    title: "3D (Beta)",
+    type: "3D",
+    description:
+      "Display visualization markers and models in a 3D scene using a new rendering engine.",
```
amacneil, line 63: At first glance this text is very similar to the old 3d panel (except the last few words). While it is in beta, it's probably best to get straight to the point, like:

```suggestion
      "Test our next-generation 3D panel, using a new rendering engine.",
```

amacneil:
👍 lgtm. I think we should disable `render stats` by default

defunctzombie (approved):
Agree with @amacneil comment to drop the render stats hover widget. You could make it a sidebar setting :D

amacneil:
It already is a sidebar setting :)

defunctzombie:
> It already is a sidebar setting :)

😮

Issue #3689, opened 2022-06-23
[New3D] `sensor_msgs/LaserScan` support


jhurliman:
Add a `LaserScans.ts` SceneExtension, LaserScanRenderable, and `SensorMsgs_LaserScan.stories.tsx`. `PointClouds.ts` can be used as a starting point.

PR #3712, opened 2022-06-25
[New3D] Support sensor_msgs/LaserScan

jtbandes:
**User-Facing Changes**
None (feature parity)

**Description**
Add support for LaserScan, sharing some code with PointClouds
Closes https://github.com/foxglove/studio/issues/3689


packages/studio-base/src/panels/ThreeDeeRender/renderables/PointCloudsAndLaserScans.ts
```diff
@@ -63,52 +70,62 @@ const POINT_SHAPE_OPTIONS = [
...
 
   override settingsNodes(): SettingsTreeEntry[] {
     const configTopics = this.renderer.config.topics;
     const handler = this.handleSettingsAction;
     const entries: SettingsTreeEntry[] = [];
     for (const topic of this.renderer.topics ?? []) {
-      if (POINTCLOUD_DATATYPES.has(topic.datatype)) {
-        const config = (configTopics[topic.name] ?? {}) as Partial<LayerSettingsPointCloud2>;
+      const isPointCloud = POINTCLOUD_DATATYPES.has(topic.datatype);
+      const isLaserScan = LASERSCAN_DATATYPES.has(topic.datatype);
```
jhurliman, line 120:
```suggestion
      const isLaserScan = !isPointCloud && LASERSCAN_DATATYPES.has(topic.datatype);
```

jhurliman (approved):

jtbandes:
@jhurliman Any idea what could be wrong with the hitmap that causes it to use the wrong point size?

jhurliman:
> @jhurliman Any idea what could be wrong with the hitmap that causes it to use the wrong point size?

No, but anecdotally it feels like the hitmap behavior is not always correct. I wonder if it has to do with depth sorting, although that wouldn't explain this point size issue. spector.js can capture offscreen buffers which could be helpful for troubleshooting.

PR #3715, opened 2022-06-25
[New3D] Customization options for transforms

jhurliman:
**User-Facing Changes**

None

**Description**

Adds customization options to labels and transforms in New3D. This improves usability for robots with many transforms, or scenes with larger or smaller scales.

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/color.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/color.ts
@@ -42,6 +42,15 @@ export function stringToRgb<T extends ColorRGB | THREE.Color>(output: T, colorSt
   return output;
 }
 
+// eslint-disable-next-line @foxglove/no-boolean-parameters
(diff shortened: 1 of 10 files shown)


packages/studio-base/src/panels/ThreeDeeRender/color.ts
```diff
@@ -42,6 +42,15 @@ export function stringToRgb<T extends ColorRGB | THREE.Color>(output: T, colorSt
   return output;
 }
 
+// eslint-disable-next-line @foxglove/no-boolean-parameters
+export function rgbToThreeColor(output: THREE.Color, rgb: ColorRGB, convert = true): THREE.Color {
```
jtbandes, line 46: Rather than disabling the lint rule, just make two functions that are named separately, rgbToThreeColor{Linear,SRGB}, or alternatively change call sites passing true to chained calls `rgbToThreeColor(...).convertSRGBToLinear()`.

PR #3717, opened 2022-06-27
Add PanelExtensionContext.dataSourceProfile, New3D strips "/" from coordinate frames for ROS data sources

jhurliman:
**User-Facing Changes**

- Visualization works correctly in `3D (Beta)` when connected to a ROS data source where some coordinate frames are prefixed with "/" and some are not
- [Extension API] Adds `dataSourceProfile` string to `PanelExtensionContext`

**Description**

This fixes rendering for ROS data sources where some `frame_id`s are prefixed with "/" and some are not. An example is the `taurob_tracker_simulation.bag` open dataset. This fix follows the principle of not mutating parsed messages, so Raw Message and other panels still correctly show the original data, but the New3D panel will check if the current data source is ROS and normalize coordinate frames that are inserted into or queried against the TransformTree.


packages/studio/index.d.ts
```diff
@@ -181,6 +181,9 @@ declare module "@foxglove/studio" {
     /** Actions the panel may perform related to the user's current layout. */
     readonly layout: LayoutActions;
 
+    /** Set to true if the data source is Robot Operating System (ROS) based. */
+    readonly ros: boolean;
```
defunctzombie, line 185: I'm so-so on this. If we have this flag what does it mean for panels? Are we specifying what this means or are panels suppose to do what they consider correct? Do we need to expose 1 vs 2?

Looking at the larger issue - I know that rviz will render the frame ids as being the same but is this actually true across ROS robots? Is it true for ros1 and ros2 and a documented behavior that a transform frame at an absolute path is equivalent to one with no leading absolute path? There are always going to be public datasets that we might consider "malformed" so I'd like to understand the documented behavior and what we expect panels to do.

packages/studio/index.d.ts
```diff
@@ -181,6 +181,14 @@ declare module "@foxglove/studio" {
     /** Actions the panel may perform related to the user's current layout. */
     readonly layout: LayoutActions;
 
+    /**
+     * Identifies the semantics of the data being played back, such as which topics or parameters
+     * are semantically meaningful or normalization conventions to use. This typically maps to a
+     * shorthand identifier for a robotics framework such as "ros1", "ros2", or "ulog". See the MCAP
+     * profiles concept at <https://github.com/foxglove/mcap/blob/main/docs/specification/appendix.md#well-known-profiles>.
+     */
+    readonly profile: string;
```
defunctzombie, line 190: I think we should have this field be undefined (rather than an empty string) if no profile is available.

```suggestion
    readonly profile?: string;
```

packages/studio-base/src/players/types.ts
```diff
@@ -116,6 +116,14 @@ export type PlayerState = {
   // See `const PlayerCapabilities` for more details.
   capabilities: typeof PlayerCapabilities[keyof typeof PlayerCapabilities][];
 
+  /**
+   * Identifies the semantics of the data being played back, such as which topics or parameters are
+   * semantically meaningful or normalization conventions to use. This typically maps to a shorthand
+   * identifier for a robotics framework such as "ros1", "ros2", or "ulog". See the MCAP profiles
+   * concept at <https://github.com/foxglove/mcap/blob/main/docs/specification/appendix.md#well-known-profiles>.
+   */
+  profile: string;
```
defunctzombie, line 125: Make this optional? I have a slight preference for that over empty strings.

```suggestion
  profile?: string;
```

jtbandes:
Is the PR title wrong? This doesn't seem like a "capability", just a property of the player.

defunctzombie (approved):
Please make sure to update the PR description and merge commit message with the latest field names since we did go through a few iterations.

Issue #3714, opened 2022-06-25
Node playground topics should work in Plot panel


amacneil:
Node playground does not currently preload/backfill data when it is viewed in the plot panel. We should make it work.

<img width="1118" alt="image" src="https://user-images.githubusercontent.com/637671/175757911-c5c9a677-7cb0-42ed-b31e-4e1491fd1060.png">

jtbandes:
Related https://github.com/foxglove/studio/issues/2797

jameskuszmaul-brt:
I was about to ask for this! I look forward to whatever y'all figure out...

defunctzombie:
@jameskuszmaul-brt I made a POC you can try from the preview deployment of this PR: https://github.com/foxglove/studio/pull/3721 you can give it a try on local files.

I noted a possible gotcha in the PR that does alter what messages are fed into the scripts. I'll circulate this internally and see if we are ok with the change.

jameskuszmaul-brt:
I haven't actually had a chance to try this, but just in terms of some of the subtleties about how to handle non-monotonic preloading of data:
Supporting purely stateless node playgrounds is extremely valuable on its own, so I'll try to play with it in that context; however, I know a good chunk of the things I want to implement do require maintaining some state. E.g.:
1. Performing simple causal filtering on data.
   * E.g., calculating a simple first-order filter, or a basic window filter, or whatever.
   * Doing something like calculating the derivative of a signal would also fall into this (yes, I know the `.@derivative` already covers this particular case).
2. Non-causal filtering/processing on data (i.e., when you generate the output from your playground, you are able to either view future messages or specify a timestamp different than that of the input).
   * This would be a new feature relative to the existing playground.
   * This allows, e.g., doing a moving average that has no phase delay (or calculating an actually centered derivative).
3. Combining data between different topics published at similar rates (to, say, average a couple of different sensor values).
4. Cross-referencing data from a latched topic (or similarly data in the attachment section of an MCAP file) that can be used to interpret data in a periodic message (e.g., you have transform data for the position of a sensor in a latched topic and want to use it to reference sensor data to a common origin).

[...]

PR #3721, opened 2022-06-27
Process preloaded messages with node playground

defunctzombie:
**User-Facing Changes**
Adds support for plotting the outputs of node playground scripts.

**Description**
Node playground scripts are also run on preloaded messages. Panels like plot benefit from showing the entire loaded dataset when possible. Prior to this change, node playground scripts would only run "during playback" which meant that a plot panel could not show the output of a node playground script against the entire range of data.

This change alters this behavior by running the scripts on preloaded messages and playback messages.

This implementation works by creating two instances of the shared worker for each node and running blocks through the second instance. In order for the block instance to see messages in the same order we have to flatten and re-sort the block messages. Hopefully memoization will minimize the performance hit.

Fixes: #3714

### DIFF
--- a/packages/studio-base/src/players/UserNodePlayer/index.ts
+++ b/packages/studio-base/src/players/UserNodePlayer/index.ts
@@ -10,6 +10,7 @@
 //   This source code is licensed under the Apache License, Version 2.0,
 //   found at http://www.apache.org/licenses/LICENSE-2.0
 //   You may not use this file except in compliance with the License.
+
 import { isEqual } from "lodash";
 import memoizeWeak from "memoize-weak";
 import shallowequal from "shallowequal";
@@ -272,45 +273,39 @@ export default class UserNodePlayer implements Player {
       return this._lastBlockRequest.result;
     }
 
-    const outputBlocks: (MessageBlock | undefined)[] = [];
+    // If no downstream subscriptions want blocks for our output topics we can just pass through
+    // the blocks from the underlying player.
+    const fullRegistrations = nodeRegistrations.filter(
+      (reg) => this._nodeSubscriptions[reg.output.name]?.preloadType === "full",
+    );
+    if (fullRegistrations.length === 0) {
+      return blocks;
+    }
 
+    const outputBlocks: (MessageBlock | undefined)[] = [];
     for (const block of blocks) {
       if (!block) {
         outputBlocks.push(block);
         continue;
       }
 
+      // Flatten and re-sort block messages so that nodes see them in the same order
+      // as the non-block nodes.
       const messagesByTopic = { ...block.messagesByTopic };
-
-      // for each node registration, iterate the input topics
-      // for each input topic process the messagesByTopic
-      for (const nodeRegistration of nodeRegistrations) {
-        const topic = nodeRegistration.output.name;
-        const subscription = this._nodeSubscriptions[topic];
-        if (subscription == undefined || subscription.preloadType !== "full") {
-          continue;
-        }
-
-        const outputMessages: MessageEvent<unknown>[] = [];
-        for (const inputTopic of nodeRegistration.inputs) {
-          const inputMessages = block.messagesByTopic[inputTopic];
-          if (!inputMessages) {
-            continue;
-          }
-
-          for (const inputMessage of inputMessages) {
-            const outputMessage = await nodeRegistration.processBlockMessage(
-              inputMessage,
-              globalVariables,
-            );
-            if (!outputMessage) {
-              continue;
+      const blockMessages = Object.values(messagesByTopic)
+        .flat()
+        .sort((a, b) => compare(a.receiveTime, b.receiveTime));
+      for (const nodeRegistration of fullRegistrations) {
+        const outTopic = nodeRegistration.output.name;
+        for (const message of blockMessages) {
+          if (nodeRegistration.inputs.includes(message.topic)) {
+            const outputMessage = await nodeRegistration.processMessage(message, globalVariables);
+            if (outputMessage) {
+              messagesByTopic[outTopic] ??= [];
+              messagesByTopic[outTopic]?.push(outputMessage);
             }
-            outputMessages.push(outputMessage);
           }
         }
-
-        messagesByTopic[topic] = outputMessages;
       }
 
       outputBlocks.push({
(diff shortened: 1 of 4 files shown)


defunctzombie:
A gotcha here is that any node playground scripts that manage "state" are going to have a trickier time since we won't be invoking the script with continuous message ranges or time ranges.

defunctzombie:
Some ideas from our slack conversations:

- Make a separate instance of the user script for running against preloaded ranges. This keeps the state of scripts run on preloaded ranges separate from those run during playback.
- Only run scripts on preload ranges if their output topic is subscribed to with "full" preload option. If the layout doesn't need the output topic in a full preload then there's no value in running the script on the preload data.

packages/studio-base/src/players/UserNodePlayer/index.ts
```diff
@@ -247,6 +248,72 @@ export default class UserNodePlayer implements Player {
...
+      // for each node registration, iterate the input topics
+      // for each input topic process the messagesByTopic
+      for (const nodeRegistration of nodeRegistrations) {
+        if (!this._nodeSubscriptions.has(nodeRegistration.output.name)) {
+          continue;
+        }
+
+        const outputMessages: MessageEvent<unknown>[] = [];
+        for (const inputTopic of nodeRegistration.inputs) {
+          const inputMessages = block.messagesByTopic[inputTopic];
```
foxymiles, line 288: @defunctzombie If we iterate blocks this way then it seems like the node is still going to see messages in a different order than the node rendering the non-block messages because we're sending all messages of each topic before moving on to the new topic.

packages/studio-base/src/players/UserNodePlayer/index.ts
```diff
@@ -247,6 +258,72 @@ export default class UserNodePlayer implements Player {
...
+              messagesByTopic[outTopic] ??= [];
+              messagesByTopic[outTopic]?.push(outputMessage);
+            }
+          }
+        }
+      }
+
+      outputBlocks.push({
+        messagesByTopic,
+        sizeInBytes: block.sizeInBytes,
```
defunctzombie, line 315: Worth a comment here that the block size does not include any newly generated messages.

packages/studio-base/src/players/UserNodePlayer/index.ts
```diff
@@ -457,7 +536,8 @@ export default class UserNodePlayer implements Player {
       nodeData,
       inputs: inputTopics,
       output: { name: outputTopic, datatype: outputDatatype },
-      processMessage,
+      processMessage: buildMessageProcessor(),
+      processBlockMessage: buildMessageProcessor(),
```
jtbandes, line 540: Why do we need two separate processors?

packages/studio-base/src/players/UserNodePlayer/index.ts
```diff
@@ -247,6 +258,75 @@ export default class UserNodePlayer implements Player {
...
+      if (!block) {
+        outputBlocks.push(block);
+        continue;
+      }
+
+      // Flatten and re-sort block messages so that nodes see them in the same order
+      // as the non-block nodes.
+      const messagesByTopic = { ...block.messagesByTopic };
+      const blockMessages = Object.values(pick(messagesByTopic, allInputTopics))
+        .flat()
```
jtbandes, line 298: Could this be replaced with `allInputTopics.flatMap((topic) => messagesByTopic[topic])` ? Might be slightly more efficient (fewer intermediate arrays/objects)?

packages/studio-base/src/players/UserNodePlayer/index.ts
```diff
@@ -247,6 +258,75 @@ export default class UserNodePlayer implements Player {
...
+        .sort((a, b) => compare(a.receiveTime, b.receiveTime));
+      for (const nodeRegistration of fullRegistrations) {
+        const outTopic = nodeRegistration.output.name;
+        for (const message of blockMessages) {
+          if (nodeRegistration.inputs.includes(message.topic)) {
+            const outputMessage = await nodeRegistration.processMessage(message, globalVariables);
+            if (outputMessage) {
+              messagesByTopic[outTopic] ??= [];
+              messagesByTopic[outTopic]?.push(outputMessage);
```
jtbandes, line 307: I think this can be
```suggestion
              (messagesByTopic[outTopic] ??= []).push(outputMessage);
```

jtbandes (approved):

packages/studio-base/src/players/UserNodePlayer/index.ts
```diff
@@ -716,8 +796,26 @@ export default class UserNodePlayer implements Player {
...
+      const playerProgress = {
+        ...playerState.progress,
+      };
+
+      if (playerProgress.messageCache) {
+        const newBlocks = await this._getBlocks(
+          playerProgress.messageCache.blocks,
+          globalVariables,
+          this._nodeRegistrations,
+        );
```
jtbandes, line 808: Does this mean we are running nodes on _all_ messages in the input blocks each time the blocks change? That sounds like it would be really slow during preloading.

Also, do we need to reset the nodes' state (by restarting the worker?) each time we process new blocks? how does statefulness play with preloading?

packages/studio-base/src/players/UserNodePlayer/index.ts
foxymiles: See discussion in comments. The goal is to have processors see the same messages in the same order whether they're processing blocks or individual messages. Since this precludes sending the same message to the same process more than once we need one node to process the blocks and single messages separately.

packages/studio-base/src/players/UserNodePlayer/index.ts
foxymiles: Yes we are not currently caching per-block. Perhaps we could try to do that as an optimization in a followup PR?

I guess the statefulness of the nodes is potentially going to be different from the per message workers no matter what we do since we dynamically eject blocks when we exceed our memory limit and they can be recalculated in any order.

packages/studio-base/src/players/UserNodePlayer/index.ts
foxymiles: I wouldn't expect this to be performance critical but it is slightly nicer to read. 👍

PR #3722, opened 2022-06-27
Always show visibility toggle on untoggled settings nodes

### LINKED ISSUE #3509: Unselected topics in settings sidebar look disabled
**Description**
In the new settings sidebar, unselected topics are greyed out (in addition to the "eye" button that appears on hover):

image

To me, this looks disabled, i.e. not something that I have the option to enable. Maybe we should consider a checkbox or some other styling to draw attention to the fact that these topics are in fact enable-able.

foxymiles:
**User-Facing Changes**
None

**Description**
The visibility toggle is always shown for nodes that have been toggled off to make it clearer to the user that they're not disabled.

Personally I think the meaning of the closed eye icon is not obvious and this might be better:

https://mui.com/material-ui/material-icons/?query=vis&selected=VisibilityOff

Screen Shot 2022-06-27 at 10 40 21 AM

Fixes #3509

### DIFF
--- a/packages/studio-base/src/components/SettingsTreeEditor/NodeEditor.tsx
+++ b/packages/studio-base/src/components/SettingsTreeEditor/NodeEditor.tsx
@@ -49,7 +49,9 @@ const EditButton = muiStyled(IconButton)(({ theme }) => ({
   padding: theme.spacing(0.5),
 }));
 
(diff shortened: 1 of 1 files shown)


amacneil:
Or just a tried and true checkbox

defunctzombie (approved):

jhurliman (approved):

Issue #3713, opened 2022-06-25
Plot panel x path autocomplete forces a leading slash


amacneil:
**Description**
The attached mcap file has a single topic named `sample_topic` (no leading slash).

[plot_example.mcap.zip](https://github.com/foxglove/studio/files/8983849/plot_example.mcap.zip)

Our plot panel x path forces you to have a leading slash in the topic name. The y path seems happy without leading slash.

<img width="634" alt="image" src="https://user-images.githubusercontent.com/637671/175757731-9bf83c23-4220-4bcb-b16f-a0fb21541c27.png">

It is possible to work around this by typing a topic, then going back to delete the leading slash. But it is not possible to clear the slash before you start typing, leading to a broken autocomplete experience.

- Version: main
- OS: web

PR #3724, opened 2022-06-27
Allow user to delete the initial slash in plot x path

foxymiles:
**User-Facing Changes**
None

**Description**
This is a partial fix for https://github.com/foxglove/studio/issues/3713. It allows the user to manually delete the initial `/` for the x-path in the plot panel. Unfortunately clearing the subsequent "mismatched messages" error described in https://github.com/foxglove/studio/issues/3713 is going to require a more significant reworking of the messaging logic here.

Partially fixes https://github.com/foxglove/studio/issues/3713

### DIFF
--- a/packages/studio-base/src/panels/Plot/PlotLegend.tsx
+++ b/packages/studio-base/src/panels/Plot/PlotLegend.tsx
@@ -394,7 +394,7 @@ export default function PlotLegend(props: PlotLegendProps): JSX.Element {
           />
           {(xAxisVal === "custom" || xAxisVal === "currentCustom") && (
             <MessagePathInput
-              path={xAxisPath?.value ? xAxisPath.value : "/"}
+              path={xAxisPath?.value ?? "/"}
               onChange={(newXAxisPath) =>
                 saveConfig({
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/panels/Plot/PlotLegend.tsx
```diff
@@ -394,7 +394,7 @@ export default function PlotLegend(props: PlotLegendProps): JSX.Element {
           />
           {(xAxisVal === "custom" || xAxisVal === "currentCustom") && (
             <MessagePathInput
-              path={xAxisPath?.value ? xAxisPath.value : "/"}
+              path={xAxisPath?.value ?? "/"}
```
defunctzombie, line 397: Can we remove the `/` fallback altogether?

defunctzombie (approved):

packages/studio-base/src/panels/Plot/PlotLegend.tsx
foxymiles: Sure I can just take it out.

PR #3736, opened 2022-06-28
Graceful chart component unmounting

defunctzombie:
**User-Facing Changes**
None

**Description**
In dev the Chart component was constantly throwing rpc errors when loading data sources or changing panels. The Chart component sends an rpc command to destroy when shutting down but doesn't wait for the response before removing the rpc id from the worker pool.

This change fixes this the crash behavior by removing the rpc id in a finally that happens after the rpc call completes.

### DIFF
--- a/packages/studio-base/src/components/Chart/index.tsx
+++ b/packages/studio-base/src/components/Chart/index.tsx
@@ -144,10 +144,11 @@ function Chart(props: Props): JSX.Element {
     return () => {
       log.info(`Unregister chart ${id}`);
       rpcSendRef.current = undefined;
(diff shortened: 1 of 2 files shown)


jtbandes (approved):
Seems like there could still be an uncaught exception, but only if the worker actually has some error? Still might be nice to do something other than leave it uncaught

defunctzombie:
> Seems like there could still be an uncaught exception, but only if the worker actually has some error? Still might be nice to do something other than leave it uncaught

I think for now I'll let that go up to the app and see what comes of it. Maybe a sentry error? Will re-visit in that case.

PR #3737, opened 2022-06-28
Remove duplicate nodeSubscriptions filter in UserNodePlayer

defunctzombie:
**User-Facing Changes**
None

**Description**
UserNodePlayer setSubscriptions creates a set of subscriptions for its child player. These set of subscriptions is a merger of the subscriptions from the setSubscriptions call and the required topics from user node scripts.

The logic it used would limit each topic to one subscription entry. Unfortunately, since we support "partial" and "full" subscriptions on topics this meant that subscription information was lost.

This change removes the topic filtering and passes through all requested subscriptions to the underlying player.

### DIFF
--- a/packages/studio-base/src/components/MessagePipeline/usePlayerState.ts
+++ b/packages/studio-base/src/components/MessagePipeline/usePlayerState.ts
@@ -58,7 +58,13 @@ function updateSubscriberAction(
   }
 
   const newSubscriptionsById = new Map(previousSubscriptionsById);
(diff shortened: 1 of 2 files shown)


defunctzombie:
@jtbandes @foxymiles Alternatively - if we think it is unwise to pass along all subscriptions to the underlying player and allow duplicate topics (which the interface technically allows), I think it is more correct to filter this at the MessagePipeline layer.

PR #3741, opened 2022-06-28
Add Immutable definition to studio-base

foxymiles:
**User-Facing Changes**
None

**Description**
Immutability is a very useful concept for code correctness but `DeepReadonly` is syntactically awkward and a proliferation of imports from `ts-essentials` all over the codebase seems undesirable.  `Immutable` from immerjs is syntactically cleaner but we don't want to add dependencies on immer just for this definition to code that doesn't otherwise use immer.

This adds an `Immutable` definition to `@foxglove/studio-base` that for now is just an alias to the `ts-essentials` `DeepReadonly` definition but is nicer to read and more convenient to import.

### DIFF
--- a/packages/studio-base/src/typings/immutable.d.ts
+++ b/packages/studio-base/src/typings/immutable.d.ts
@@ -1,0 +1,7 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+declare module "@foxglove/studio-base" {
+  export type Immutable<T> = import("ts-essentials").DeepReadonly<T>;
+}
--- a/packages/studio-base/src/typings/immutable.d.ts
+++ b/packages/studio-base/src/typings/immutable.d.ts
@@ -5,3 +5,5 @@
 declare module "@foxglove/studio-base" {
   export type Immutable<T> = import("ts-essentials").DeepReadonly<T>;
 }
+
+export {};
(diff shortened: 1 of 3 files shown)


jhurliman (approved):

packages/studio-base/src/typings/immutable.d.ts
```diff
@@ -0,0 +1,9 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+declare module "@foxglove/studio-base" {
```
jtbandes, line 5: Why do we need this `declare module` thing? Can't we just add the export to `studio-base/src/index.ts`?

jtbandes (commented):
I personally don't find Immutable any clearer than DeepReadonly and don't feel strongly that importing from ts-essentials is bad.

foxymiles:
Do you have an opinion @defunctzombie ?

defunctzombie (commented):
We don't really have a habit of importing from studio-base root within files that are in studio-base since that could lead to circular imports. 

What do you think about putting this in `foxglove/den/types`?

defunctzombie:
Beyond my opinion that I'd prefer to see this in `den` rather than `studio-base` LGTM to have this in our toolbelt.

defunctzombie:
I don't have a preference over Immutable vs DeepReadonly. Arguably `DeepReadonly` mirrors the builtin `Readonly` by name better.

Can we modify `Readonly` to be deep by default?

jhurliman:
> Can we modify `Readonly` to be deep by default?

Modifying built-in language types to have different behavior would likely increase the learning curve to working on this codebase.

foxymiles:
Yeah modifying existing definitions seems likely to cause more problems than it solves.

PR #3748, opened 2022-06-28
Add a memory use indicator option to experimental settings

defunctzombie:
**User-Facing Changes**
Users can enable a memory use indicator in the app settings. The indicator shows the used javascript memory as a percent and used vs free as bytes in a tooltip.

image

**Description**
The memory use indicator shows used js heap memory relative to total js heap memory. This is useful in understand the memory impact of various caching strategies.


amacneil:
I would argue for allowing this in production under a "show debug information" flag, or a keyboard shortcut, or something. I think it would be helpful for debugging occasionally, and I and other users are not typically running the dev build.

defunctzombie:
Updated to be available in production releases as well. There's probably a better place for this but for now it can live in the navbar.

packages/studio-base/src/AppSetting.ts
```diff
@@ -30,4 +30,5 @@ export enum AppSetting {
   // Dev only
   ENABLE_LAYOUT_DEBUGGING = "enableLayoutDebugging",
   ENABLE_REACT_STRICT_MODE = "enableReactStrictMode",
+  ENABLE_MEMORY_USE_INDICATOR = "dev.memory-use-indicator",
```
jtbandes, line 33: our naming scheme in this file is a disaster

packages/studio-base/src/components/ExperimentalFeatureSettings.tsx
```diff
@@ -53,6 +53,11 @@ const features: Feature[] = [
     name: "Latching",
     description: <>Enable message latching for bag, mcap, and data platform sources.</>,
   },
+  {
+    key: AppSetting.ENABLE_MEMORY_USE_INDICATOR,
+    name: "Memory use indicator",
+    description: <>Show the app memory use in the navbar.</>,
```
jtbandes, line 59: What's a navbar? Never seen that term in our app before. Maybe avoid this abbreviation in user facing text?

packages/studio-base/src/components/Sidebar/MemoryUseIndicator.tsx
```diff
@@ -0,0 +1,50 @@
...
+    textAlign: "left",
+    padding: theme.spacing(1),
+  },
+}));
+
+function toMB(bytes: number): number {
+  return bytes / 1024 / 1024;
+}
+
+function MemoryUseIndicator(): JSX.Element {
```
jtbandes, line 23: Can we get this (with fake data) in a screenshot test?

packages/studio-base/src/components/Sidebar/index.tsx
```diff
@@ -77,6 +80,9 @@ export default function Sidebar<K extends string>({
   selectedKey: K | undefined;
   onSelectKey: (key: K | undefined) => void;
 }>): JSX.Element {
+  const [enableMemoryUseIndocator = false] = useAppConfigurationValue<boolean>(
```
jtbandes, line 83:
```suggestion
  const [enableMemoryUseIndicator = false] = useAppConfigurationValue<boolean>(
```

packages/studio-base/src/hooks/useMemoryInfo.ts
```diff
@@ -0,0 +1,49 @@
...
+
+interface Performance {
+  memory?: MemoryInfo;
+}
+
+type UseMemoryInfoOptions = {
+  refreshIntervalMs: number;
+};
+
+const performance = window.performance as Performance;
```
jtbandes, line 28: why is this casting dance necessary? isn't performance already treated as a global var (on window) of type Performance? Or maybe there is some `@types` package we can install to get this?

jtbandes (approved):

packages/studio-base/src/components/Sidebar/MemoryUseIndicator.tsx
jtbandes: Kinda weird for this component to live in the Sidebar folder.

packages/studio-base/src/components/ExperimentalFeatureSettings.tsx
defunctzombie: What do I call the icons? Its not the sidebar since the sidebar is the thing that opens? Or do we call it the sidebar?

packages/studio-base/src/components/Sidebar/MemoryUseIndicator.tsx
defunctzombie: This one is pretty specific to how I'm wanting to display it in the Sidebar so I put it in the Sidebar folder for now.

packages/studio-base/src/hooks/useMemoryInfo.ts
defunctzombie: Apparently it is not in our types. I'll check again if there's a type I can install to get this.

defunctzombie: performance.memory is specific to Chrome and not present in typings. So this cast is about adding the optional memory field for access in the hook.

packages/studio-base/src/components/Sidebar/MemoryUseIndicator.tsx
defunctzombie: Not seeing a nice way to do this without making an un-connected component and a connected one. There's no dependency injection in this hook.

PR #3750, opened 2022-06-29
Remove one more duplicate iterator instantiation from IterablePlayer

defunctzombie:
**User-Facing Changes**
None for regular player users.

Experimental player users will have a slight performance improvement by re-using the forward iterator rather than creating a new one. This performance improvement is unlikely to be noticable in practice for most data sources.

**Description**
Re-using the forward iterator from start-play as the playback iterator removes another instance of iterator creation which imroves startup performance since the iterator has already loaded the data we want to start playing.


defunctzombie:
Waiting on https://github.com/foxglove/studio/pull/3279

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
```diff
@@ -522,16 +522,13 @@ export class IterablePlayer implements Player {
         continue;
       }
 
-      // Just in case the iterator decides it is going to ignore our _end_ param
+      messageEvents.push(iterResult.msgEvent);
```
jtbandes, line 525: why does this move earlier? seems like we would not want to emit a message that comes after our stopTime?

defunctzombie: Good catch because in fact this means that we might have a message after stopTime but we set currentTime to stop time and thats a no-no to the message pipeline.

fixing.

jtbandes (approved):

PR #3752, opened 2022-06-29
Demo only: Update electron-builder.json for building snaps

defunctzombie:
**User-Facing Changes**
None

**Description**
This changes allow for electron-builder to produce snaps for foxglove studio.

* Update `electron-builder.json` snap configuration to use `strict` confinement. `strict` confinement is a requirement for streamlining acceptance into the snap store.
* Patch `app-builder-lib` to use `core20` as the base core for snap files.

### DIFF
--- a/.github/workflows/snap.yml
+++ b/.github/workflows/snap.yml
@@ -1,0 +1,62 @@
+name: CI
+
+on:
+  push:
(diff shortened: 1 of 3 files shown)


defunctzombie:
Closing - the demo served its purpose.

PR #3755, opened 2022-06-29
[New3D] PointCloud and LaserScan decay

### LINKED ISSUE #3283: Adding decay time makes point cloud disappear

jhurliman:
**User-Facing Changes**

- Implements the decay setting for PointCloud and LaserScan topics in `3D (Beta)` panel

**Description**

Fixes #3283. Implements decay for point clouds and laser scans. The original messages are not kept around (a consideration for per-point picking when that is implemented), only the GPU buffers.

Other fixes:

- Option to override setUsage() for DynamicBufferGeometry
- Don't expire transforms after 60s
- Dispose renderables in removeAllRenderables()
- Fix transform lookup when all TFs are after the search time but within maxDelta

Screen Shot 2022-06-29 at 1 50 04 PM


packages/studio-base/src/panels/ThreeDeeRender/renderables/PointCloudsAndLaserScans.ts
```diff
@@ -272,58 +345,131 @@ export class PointCloudsAndLaserScans extends SceneExtension<PointCloudAndLaserS
     );
   };
 
+  // eslint-disable-next-line @foxglove/no-boolean-parameters
+  _createGeometry(topic: string, dynamic: boolean): DynamicFloatBufferGeometry {
```
jtbandes, line 349: Remove lint suppression and pass in Usage enum directly for clarity

jtbandes (approved):
Do you think there's a way we can reduce the number of different places in this code where we have to initialize objects? It just gets confusing to read when there are 4 or so calls to createPoints() and other initializations. When I was adding LaserScans I might have made it worse 😬

jhurliman:
@jtbandes I think it could be refactored to remove the create calls when initializing renderables and in the _update*() methods, check if `pointsHistory.length === 0`. That would get it down to two places, one for point clouds and one for laser scans.

Issue #3730, opened 2022-06-27
"Invariant: uninitialized" with experimental mcap player


jtbandes:
Drop the following file into studio. The player displays "Invariant: uninitialized" along with the actual error. If this is actually an invariant then we must have some logic error.

[bad.mcap.zip](https://github.com/foxglove/studio/files/8996241/bad.mcap.zip)

![image](https://user-images.githubusercontent.com/14237/176053676-c4302767-6463-4884-97ca-26d1592a4618.png)

PR #3759, opened 2022-06-30
Gracefully handle un-indexed mcap files with experimental player

defunctzombie:
**User-Facing Changes**
Loading an un-indexed mcap file with the experimental player no longer shows an un-actionable "invariant" error.

**Description**
Rather than throwing an invariant error for un-indexed files, throw in initialize() which puts the IterablePlayer in an error state and avoids trying to read messages which triggered the invariant throws.

Fixes: #3730

### DIFF
--- a/packages/studio-base/src/players/IterablePlayer/Mcap/McapIterableSource.ts
+++ b/packages/studio-base/src/players/IterablePlayer/Mcap/McapIterableSource.ts
@@ -40,6 +40,8 @@ async function tryCreateIndexedReader(readable: Mcap0Types.IReadable) {
   return reader;
 }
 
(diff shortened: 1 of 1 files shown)


jtbandes:
What would happen if we just made the initialize() method throw instead, and make the player show that failure as a problem, rather than requiring initialize to return a no-op source?

defunctzombie:
> What would happen if we just made the initialize() method throw instead, and make the player show that failure as a problem, rather than requiring initialize to return a no-op source?

I think the player doesn't "unset" the source so it would still try to call the methods on the source. We could change that behavior so the player catches initialization throws and treats that as an end to any further playback.

esthersweon:
@defunctzombie - I don't think we should list changes to off-by-default experimental players as "User-facing changes".

PR #3763, opened 2022-06-30
Unbreak JSON MCAP/ws support

jameskuszmaul-brt:
Accidentally made a flyby "improvement" to the syntax in #3465 that
was actually behavioral.

Should really have tests for this....

**User-Facing Changes**

MCAP files using JSON encoding and foxglove websocket connectors using JSON will actually work again.

**Description**

In #3465, I incidentally added some braces in a spot that caused the JSON deserializer to no longer have a return value.

Don't have the bandwidth to write tests or add more explicit type checking for this right this instant, but should probably do so.


jtbandes:
I would've hoped for TypeScript to catch this kind of error, but I think the deserializer functions return `unknown`, and `undefined` satisfies `unknown` :(

PR #3764, opened 2022-06-30
Use full size color input in settings

### LINKED ISSUE #3684: Show #ffffffff in settings tree rgba fields instead of rgba(255, 255, 255, 1)
**Description**

Switching back and forth between hex strings and CSS-style rgba(...) strings based on what alpha value is used is a surprising result, and the rgba(...) strings are usually too long to fit on screen in the sidebar. We should consistently display eight-character hex codes for rgba inputs (keeping six-character hex for rgb inputs). We can still accept user input of the form rgba(...) but it should be reformatted as eight-character hex.

foxymiles:
**User-Facing Changes**
This improves the color input in the settings editor.

**Description**
This fills the entire background of the color field with the selected color and requires the user to open the color picker to manually enter a value. 

This has a few benefits:

1. `rgba` values are no longer clipped.
2. We no longer have the problem of handing text input in arbitrary format.s
3. The chosen color is more visible than it is in the small swatch we have now.
4. The resulting input looks more like the gradient input.

Screen Shot 2022-06-30 at 11 34 08 AM

Fixes #3684

### DIFF
--- a/packages/studio-base/src/components/SettingsTreeEditor/index.stories.tsx
+++ b/packages/studio-base/src/components/SettingsTreeEditor/index.stories.tsx
@@ -611,6 +611,19 @@ const FilterSettings: SettingsTreeNodes = {
   },
 };
 
+const ColorSettings: SettingsTreeNodes = {
+  colors: {
+    fields: {
+      hex6: { label: "Hex 6", input: "rgb", value: "#ffaa00" },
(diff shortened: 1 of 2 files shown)


defunctzombie:
I'm not sure I'd call this an "improvement". I personally do not like reading text over color backgrounds.

I am 👎 this change.

PR #3766, opened 2022-06-30
Always show color values as hex

### LINKED ISSUE #3684: Show #ffffffff in settings tree rgba fields instead of rgba(255, 255, 255, 1)

foxymiles:
**User-Facing Changes**
This improves the settings editor color input.

**Description**
This is an alternative solution to https://github.com/foxglove/studio/pull/3764.

The solution here is to always show the hex value of the color and require the user to use the color picker to manually enter values.

Screen Shot 2022-06-30 at 2 11 43 PM

Fixes #3684


jhurliman (approved):
I haven't reviewed the code change, but I much prefer this solution over the alternative.

2metres:
I prefer this version as well

PR #3767, opened 2022-06-30
Fix erroreneous "&:" causing body background to have hover state

2metres:
**User-Facing Changes**
None

**Description**
Removes body background hover state introduced in #3650


jtbandes (approved):
How did this result in the body having a hover background? Why does the `&` end up being the body?

2metres:
> How did this result in the body having a hover background? Why does the `&` end up being the body?

Differences between MUI makeStyles and FluentUI makeStyles syntax

Issue #3731, opened 2022-06-27
Incorrect error message for mcap file with no messages using experimental mcap player


jtbandes:
When dropping this file into Studio, I get "The mcap file is unindexed. Only indexed files are supported."

[bad.mcap.zip](https://github.com/foxglove/studio/files/8996255/bad.mcap.zip)

However, the file is in fact indexed, so the error message is wrong.

![image](https://user-images.githubusercontent.com/14237/176053933-3d2ee10c-4cf6-4f04-bb7f-03b980324999.png)

defunctzombie:
When I try loading this same file in the non-experimental player I see this warning:

<img width="716" alt="image" src="https://user-images.githubusercontent.com/84792/176573404-11470d1f-e732-4228-9e40-443650a45731.png">

It also thinks the file is un-indexed.

Here is how we are identifying indexed files right now:

https://github.com/foxglove/studio/blob/main/packages/studio-base/src/players/IterablePlayer/Mcap/McapIterableSource.ts#L25

Is there something wrong with this logic? I copied it from the non-experimental player.

jtbandes:
Yeah, probably both are wrong. I guess they don't account for _empty_ files as a special case.

defunctzombie:
I can make a fix here if you think there is a fix we should make? Should we add logic to detect what we think to be an empty file? Or do we keep the existing error message?

PR #3768, opened 2022-06-30
Distinguish between un-indexed and empty mcap files

defunctzombie:
**User-Facing Changes**
Users of the experimental mcap player are given a more detailed error message if their mcap file cannot be opened with the indexed reader.

**Description**
Mcap files may have summary records but be missing chunkindex records (as a result of not having any messages). We distinguish between files with no summary information and those that might have some summary information but are missing the summary information we need.

Fixes: #3731

----
I'm open to other suggestion in the language here. A more robust check would be for the Mcap0IndexedReader to report from `Initialize()` on whether it will make an attempt at reading the file (i.e. whether initialize succeeds) rather than the caller of Initialize having to guess at the failure modes since those are internal to the reader.

### DIFF
--- a/packages/studio-base/src/players/IterablePlayer/Mcap/McapIterableSource.ts
+++ b/packages/studio-base/src/players/IterablePlayer/Mcap/McapIterableSource.ts
@@ -25,9 +25,9 @@ async function tryCreateIndexedReader(readable: Mcap0Types.IReadable) {
 
   if (reader.chunkIndexes.length === 0 || reader.channelsById.size === 0) {
     if (reader.summaryOffsetsByOpcode.size > 0) {
-      throw new Error("The mcap file is empty or has an incomplete summary section.");
+      throw new Error("The MCAP file is empty or has an incomplete summary section.");
     } else {
-      throw new Error("The mcap file is unindexed. Only indexed files are supported.");
+      throw new Error("The MCAP file is unindexed. Only indexed files are supported.");
     }
   }
   return reader;
--- a/packages/studio-base/src/players/IterablePlayer/Mcap/McapIterableSource.ts
+++ b/packages/studio-base/src/players/IterablePlayer/Mcap/McapIterableSource.ts
@@ -4,7 +4,6 @@
 
 import { Mcap0IndexedReader, Mcap0Types } from "@mcap/core";
 
-import Logger from "@foxglove/log";
 import { loadDecompressHandlers } from "@foxglove/mcap-support";
 import { FileReadable } from "@foxglove/studio-base/players/IterablePlayer/Mcap/FileReadable";
 import { MessageEvent } from "@foxglove/studio-base/players/types";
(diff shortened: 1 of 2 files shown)


defunctzombie:
Not saying we should go with this wording or approach but we can find a more robust way to indicate to the user why we are rejecting their file.

packages/studio-base/src/players/IterablePlayer/Mcap/McapIterableSource.ts
```diff
@@ -18,25 +17,19 @@ import {
...
-  let hasMissingSchemas = false;
-  for (const channel of reader.channelsById.values()) {
-    if (channel.schemaId !== 0 && !reader.schemasById.has(channel.schemaId)) {
-      hasMissingSchemas = true;
-      break;
+  if (reader.chunkIndexes.length === 0 || reader.channelsById.size === 0) {
+    if (reader.summaryOffsetsByOpcode.size > 0) {
+      throw new Error("The mcap file is empty or has an incomplete summary section.");
+    } else {
+      throw new Error("The mcap file is unindexed. Only indexed files are supported.");
```
jtbandes, line 30:
```suggestion
      throw new Error("The MCAP file is empty or has an incomplete summary section.");
    } else {
      throw new Error("The MCAP file is unindexed. Only indexed files are supported.");
```

packages/studio-base/src/players/IterablePlayer/Mcap/McapIterableSource.ts
```diff
@@ -18,25 +17,19 @@ import {
...
 async function tryCreateIndexedReader(readable: Mcap0Types.IReadable) {
   const decompressHandlers = await loadDecompressHandlers();
   const reader = await Mcap0IndexedReader.Initialize({ readable, decompressHandlers });
 
-  let hasMissingSchemas = false;
-  for (const channel of reader.channelsById.values()) {
-    if (channel.schemaId !== 0 && !reader.schemasById.has(channel.schemaId)) {
-      hasMissingSchemas = true;
-      break;
+  if (reader.chunkIndexes.length === 0 || reader.channelsById.size === 0) {
```
jtbandes, line 26: Maybe add a check that's also based on summary.messageCount? Or are people allowed to leave that blank but still have the file be indexed?

jtbandes (approved):
Could we copy this logic to the old player's mcap data source at the same time so they stay in sync?

packages/studio-base/src/players/IterablePlayer/Mcap/McapIterableSource.ts
defunctzombie: > Or are people allowed to leave that blank but still have the file be indexed?

Technically yes? That was my hesitation with putting that check in.

esthersweon:
@defunctzombie - FYI, will not include in release notes.

PR #3769, opened 2022-07-01
Use MUI Dialog in favor of Fluent UI Dialog

2metres:
**User-Facing Changes**
Minimal visual changes (see storybook diffs for full differences)

**Description**
Replace fluentui Dialog components with @mui equivalents

### DIFF
--- a/packages/studio-base/package.json
+++ b/packages/studio-base/package.json
@@ -106,7 +106,7 @@
     "@types/uuid": "8.3.4",
     "@types/wicg-file-system-access": "2020.9.5",
     "@types/ws": "8.5.2",
-    "@wojtekmaj/enzyme-adapter-react-17": "0.6.6",
+    "@wojtekmaj/enzyme-adapter-react-17": "0.6.7",
     "amplitude-js": "8.16.1",
     "argparse": "2.0.1",
     "async-mutex": "0.3.2",
--- a/packages/studio-base/src/components/HoverableIconButton.tsx
+++ b/packages/studio-base/src/components/HoverableIconButton.tsx
@@ -8,10 +8,12 @@ import { forwardRef, useCallback, useEffect, useState } from "react";
 type Props = {
   icon: React.ReactNode;
   activeIcon?: React.ReactNode;
-} & Omit<IconButtonProps, "children">;
+  color?: IconButtonProps["color"];
+  activeColor?: IconButtonProps["color"];
+} & Omit<IconButtonProps, "children" | "color">;
 
 const HoverableIconButton = forwardRef<HTMLButtonElement, Props>((props, ref) => {
-  const { icon, activeIcon, ...rest } = props;
+  const { icon, activeIcon, color, activeColor, ...rest } = props;
 
   const [hovered, setHovered] = useState(false);
 
@@ -39,6 +41,7 @@ const HoverableIconButton = forwardRef<HTMLButtonElement, Props>((props, ref) =>
       component="button"
       onMouseEnter={onMouseOver}
       onMouseLeave={onMouseLeave}
+      color={activeColor != undefined ? (hovered ? activeColor : color) : color}
     >
       {activeIcon != undefined ? (hovered ? activeIcon : icon) : icon}
     </IconButton>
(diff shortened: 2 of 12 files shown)


packages/studio-base/package.json
```diff
@@ -106,7 +106,7 @@
     "@types/uuid": "8.3.4",
     "@types/wicg-file-system-access": "2020.9.5",
     "@types/ws": "8.5.2",
-    "@wojtekmaj/enzyme-adapter-react-17": "0.6.6",
+    "@wojtekmaj/enzyme-adapter-react-17": "0.6.7",
```
jtbandes, line 109: what's this about?

packages/studio-base/src/components/HoverableIconButton.tsx
```diff
@@ -8,23 +8,37 @@ import { forwardRef, useCallback, useEffect, useState } from "react";
...
   icon: React.ReactNode;
   activeIcon?: React.ReactNode;
-} & Omit<IconButtonProps, "children">;
+  color?: IconButtonProps["color"];
+  activeColor?: IconButtonProps["color"];
+} & Omit<IconButtonProps, "children" | "color">;
 
 const HoverableIconButton = forwardRef<HTMLButtonElement, Props>((props, ref) => {
-  const { icon, activeIcon, ...rest } = props;
+  const { icon, activeIcon, color, activeColor, onMouseLeave, onMouseEnter, ...rest } = props;
```
jtbandes, line 16: What's the point of accepting enter/leave handlers as props? Looks like that is not being used anywhere?

packages/studio-base/src/components/LayoutBrowser/UnsavedChangesPrompt.tsx
```diff
@@ -105,62 +86,63 @@ export function UnsavedChangesPrompt({
...
+        <DialogContent>
+          <Stack gap={2} style={{ minHeight: 180 }}>
+            <RadioGroup defaultValue="discard" onChange={handleChoiceGroupChange}>
+              <FormControlLabel value="discard" label="Discard changes" control={<Radio />} />
+              <FormControlLabel
+                value="overwrite"
+                label={[
+                  `Update team layout “${layout.name}”`,
+                  !isOnline && "(unavailable while offline)",
+                ].join(" ")}
```
jtbandes, line 101: Maybe add filter(Boolean)? As is, this change adds a trailing space in the case where isOnline=true.

jtbandes (approved):
- Feels pretty low contrast with no shadow, especially in dark mode:
  image
  image

- Looks like the Open dialog has not been converted yet?

- Some text alignment issue in the MarkerArray settings: 
  image

- Are we planning to use some kind of mui tooltip component in the future? The native `title` feels out of place because it's slow to appear and doesn't match our color scheme:
  image

packages/studio-base/package.json
2metres: There is a very specific enzyme error that happens when trying to register a click event on a MUI (emotion) button that seems to be resolved by this version bump

packages/studio-base/src/components/HoverableIconButton.tsx
```diff
@@ -8,23 +8,37 @@ import { forwardRef, useCallback, useEffect, useState } from "react";
...
+      if (onMouseEnter != undefined) {
+        onMouseEnter(event);
+      }
+      if (props.disabled === true) {
+        return;
+      }
+      setHovered(true);
+    },
+    [onMouseEnter, props.disabled],
+  );
```
2metres, line 31: @jtbandes it's used here

2metres:
> Looks like the Open dialog has not been converted yet?

Yes, I've got a [separate PR](https://github.com/foxglove/studio/pull/3774) specificallty targeting the OpenDialog

2metres:
@jtbandes 

> Are we planning to use some kind of mui tooltip component in the future? The native title feels out of place because it's slow to appear and doesn't match our color scheme

Eventually yes, but for now i'm removing the use of useTooltip as it relies on Fluent UI. I'd be open to PRs that implement something similar using the MUI tooltip component.

packages/studio-base/src/components/HoverableIconButton.tsx
jtbandes: Yeah but are any users of HoverableIconButton actually passing these in?

Issue #3749, opened 2022-06-28
Add playback caching to iterable player


defunctzombie:
The iterable player does not perform any caching. This creates a worse user experience in the following situations:

- When streaming data from data platform, topics already streamed are re-streamed when the list of subscribed topics changes.
- When reading local files, new iterators are created when the list of subscribe topics changes. This leads to de-compressing chunks again.

The upside to not caching is that memory use is lower. Instead of keeping blocks in memory for all topics (like the random access player family would do), the iterable player only keeps blocks with preloaded topic messages. This keeps memory use down when most of the topics are not pre-loaded. However, when we have available memory we should use it to improve playback performance.

PR #3772, opened 2022-07-01
Add buffering and caching for IterablePlayer playback

defunctzombie:
**User-Facing Changes**
When playing using the experimental players, the playback will cache and buffer leading to a smoother and faster playback experience.

The two videos below show before and after. Both videos show a connection to our dev data platform deployment. In each instance, I am opening the same stream and layout.

In the without-buffering video, playback pauses every 5 seconds as the player waits to start another 5 second stream. Seeking back requires re-loading the stream again.

In the with-buffering video, playback does not pause because a buffering process is reading ahead of where we are. We see an indicator for this buffering and what data is cached. We can seek back quickly because the data is cached.

https://user-images.githubusercontent.com/84792/179742804-a05c9e49-d534-4827-80f2-05e5ac6b2c6c.mov

https://user-images.githubusercontent.com/84792/179742903-4ca82235-28f8-4537-bcb1-0ae95b59fb2b.mov

**Description**
Buffering and caching are implemented via two wrapper classes. Each class implements IIterableSource and exposes a messageIterator and getBackfillMessages.

The BufferingIterableSource does read-ahead buffering of an IIterableSource. This read-ahead happens in parallel to the message reading. Messages are produced into a read-ahead buffer by the producer "process" and consumed in the messageIterator.

The CachingIterableSource does memory caching on messages produces by an IIterableSource. It's messageIterator tries first to read from its memory cache, and when there are no more cached messages, falls-back to the underlying IIterableSource. When it falls-back to reading from the source, these results are cached in memory for future reading. These results are also available to getBackfillMessages.

CachingIterableSource organizes the iterator result cache into "blocks". A block has an array of cached iterator result items within a start and end time. The start and end time are inclusive of the time span the block covers. When a block grows too large in size (based on tracking MessageEvent#sizeInBytes), a new block is started. When the entire cache is grows too large, the oldest used blocks are purged.

Fixes: #3749


packages/studio-base/src/players/IterablePlayer/CachingIterableSource.ts
```diff
@@ -0,0 +1,458 @@
...
+  GetBackfillMessagesArgs,
+  IIterableSource,
+  Initalization,
+  IteratorResult,
+  MessageIteratorArgs,
+} from "./IIterableSource";
+
+const log = Log.getLogger(__filename);
+
+// An individual cache item represents a continues range of CacheIteratorItems
```
foxymiles, line 22:
```suggestion
// An individual cache item represents a continuous range of CacheIteratorItems
```

foxymiles (approved):

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
```diff
@@ -773,6 +775,37 @@ export class IterablePlayer implements Player {
...
+    });
+
+    for (;;) {
+      this._progress = {
+        fullyLoadedFractionRanges: this._bufferedSource.loadedRanges(),
+        messageCache: this._progress.messageCache,
+      };
+      await this._emitState();
+
+      await Promise.race([delay(1000), aborted]);
```
defunctzombie, line 803: add comment

packages/studio-base/src/players/IterablePlayer/CachingIterableSource.ts
```diff
@@ -0,0 +1,458 @@
...
+  }
+
+  async *messageIterator(args: MessageIteratorArgs): AsyncIterator<Readonly<IteratorResult>> {
+    if (!this.initResult) {
+      throw new Error("Invariant: uninitialized");
+    }
+
+    let sourceMessageIterator: AsyncIterator<Readonly<IteratorResult>> | undefined;
+    try {
+      const newTopics = [...args.topics].sort();
```
defunctzombie, line 135: should we unique this?

packages/studio-base/src/players/IterablePlayer/CachingIterableSource.ts
```diff
@@ -0,0 +1,458 @@
...
+            cachedItem = block.items[idx];
+            if (!cachedItem) {
+              break;
+            }
+
+            // update the last time this block was accessed
+            block.lastAccess = Date.now();
+
+            // Increment to the next cache item
+            cacheIndex += 1;
```
defunctzombie, line 192: don't need this anymore

packages/studio-base/src/players/IterablePlayer/CachingIterableSource.ts
```diff
@@ -0,0 +1,458 @@
...
+        }
+
+        let cacheIndex = -1;
+        if (block) {
+          cacheIndex = findCacheItem(block.items, toNanoSec(readHead));
+          if (cacheIndex < 0) {
+            cacheIndex = ~cacheIndex;
+          }
+        }
+        let cachedItem = block?.items[cacheIndex];
```
defunctzombie, line 176: maybe cachedItem doesn't need to exist outside of the loop?

packages/studio-base/src/players/IterablePlayer/CachingIterableSource.ts
```diff
@@ -0,0 +1,458 @@
...
+          // next read can start at 1 nanosecond after the end
+          readHead = add(block.end, { sec: 0, nsec: 1 });
+          continue;
+        }
+
+        let sourceReadStart = readHead;
+        let sourceReadEnd = this.initResult.end;
+
+        // We have a block that is not done, we continue from where we left off and read until the next block starts
+        if (block) {
```
defunctzombie, line 206: Why would a block overlap with our time but not actually contain anything at that time

packages/studio-base/src/players/IterablePlayer/CachingIterableSource.ts
```diff
@@ -0,0 +1,458 @@
...
+          }
+
+          // Store the latest message in pending results and flush to the block when time moves forward
+          pendingIterResults.push([lastTime, iterResult.value]);
+
+          // update the last time this block was accessed
+          block.lastAccess = Date.now();
+
+          this.totalSizeBytes += sizeInBytes;
+          block.size += sizeInBytes;
```
defunctzombie, line 343: This is not the right place to add the size because we haven't added the pending items yet

packages/studio-base/src/players/IterablePlayer/CachingIterableSource.ts
```diff
@@ -0,0 +1,458 @@
...
+
+    const out: MessageEvent<unknown>[] = [];
+    const needsTopics = new Set(args.topics);
+
+    const cacheBlock = this.cache[cacheBlockIndex];
+    if (cacheBlock) {
+      let readIdx = findCacheItem(cacheBlock.items, toNanoSec(args.time));
+
+      // If readIdx is negative then we don't have an exact match, but readIdx does tell us what that is
+      // See the binarySearch documentation for how to interpret it.
```
defunctzombie, line 390: comment is wrong on binarySearch

packages/studio-base/src/players/IterablePlayer/CachingIterableSource.ts
```diff
@@ -0,0 +1,458 @@
...
+    // then we can't know that we will load the backfill messages correctly
+    const cacheBlockIndex = this.cache.findIndex((item) => {
+      return compare(item.start, args.time) <= 0 && compare(item.end, args.time) >= 0;
+    });
+
+    const out: MessageEvent<unknown>[] = [];
+    const needsTopics = new Set(args.topics);
+
+    const cacheBlock = this.cache[cacheBlockIndex];
+    if (cacheBlock) {
```
defunctzombie, line 386: If there is a block before ours we can scan it too - if there is a gap then we must fall back

packages/studio-base/src/players/IterablePlayer/CachingIterableSource.test.ts
```diff
@@ -0,0 +1,266 @@
...
+  }
+
+  async *messageIterator(_args: MessageIteratorArgs): AsyncIterator<Readonly<IteratorResult>> {}
+
+  async getBackfillMessages(_args: GetBackfillMessagesArgs): Promise<MessageEvent<unknown>[]> {
+    return [];
+  }
+}
+
+describe("BufferedIterableSource", () => {
```
defunctzombie, line 37: fixme - name is wrong

packages/studio-base/src/players/IterablePlayer/CachingIterableSource.test.ts
```diff
@@ -0,0 +1,266 @@
...
+      }
+
+      {
+        const iterResult = await messageIterator.next();
+        expect(iterResult).toEqual({
+          done: true,
+        });
+      }
+
+      expect(bufferedSource.loadedRanges()).toEqual([
```
defunctzombie, line 260: Should this be one continuous range rather than two small ranges

packages/den/collection/minIndexBy.ts
```diff
@@ -0,0 +1,24 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+function minIndexBy<T>(collection: Array<T>, compare: (itemA: T, itemB: T) => number): number {
```
jtbandes, line 5: Comment?

packages/studio-base/src/players/IterablePlayer/BufferedIterableSource.ts
```diff
@@ -0,0 +1,199 @@
...
+
+    // Clear the cache and start producing into an empty array, the consumer removes elements from
+    // the start of the array.
+    this.cache.length = 0;
+
+    const sourceIterator = this.source.messageIterator({
+      topics: args.topics,
+      start: args.start,
+      end: args.end,
+    });
```
jtbandes, line 80: Ideally we would pass an AbortSignal all the way into this messageIterator call so that we can abort the message request when this.aborted becomes true

packages/studio-base/src/players/IterablePlayer/CachingIterableSource.ts
```diff
@@ -0,0 +1,496 @@
...
+
+  let left = 0;
+  let right = list.length - 1;
+
+  while (left <= right) {
+    const mid = (left + right) >> 1;
+    const midKey = list[mid]![0];
+
+    if (midKey === key) {
+      return mid;
```
jtbandes, line 81: Is this guaranteed to return the _first_ item at the given timestamp if there are more than one? Does it need to?

jtbandes (approved):

packages/studio-base/src/players/IterablePlayer/CachingIterableSource.ts
defunctzombie: I believe the answer is yes - and it definitely needs to otherwise you wouldn't have read a message at the first index. I am gonna move this to den and write a test for it - can't have random binary search functions floating around with no tests. bad form.

jtbandes: Might be able to just use sortedIndexBy from lodash.

defunctzombie: maybe this is not ready for den yet ... naming is hard

jtbandes: Needs more than 1 test.
image (1)

esthersweon:
Will not include "user-facing changes" in release notes, as these experimental players are all off by default.

PR #3773, opened 2022-07-02
[New3D] Performance improvements, fix followTf dropdown

jhurliman:
**User-facing changes**

- Performance improvements in the `3D (Beta)` panel

**Description**

The major change here is replacing the use of `@foxglove/avl` with a new `SortedList<K, V>` class that maintains a flat, sorted array of key/value pairs with unique keys. The array size is capped at 10,000 entries to prevent OOM problems when connected to a live data source. Full fix list:

- Fix followTf in settings sidebar
- Enable frustum culling for LaserScan
- Dispose geometry when disabling PointCloud or LaserScan topic
- Topic error when we are not rendering in the user's desired frame
- Add caching for _updateMaterials()
- Performance optimizations for TransformTree, remove AVL usage, trim by count in addition to time
- Various fixes for Grid
- Show tf/topic/layer counts in the sidebar


packages/studio-base/src/panels/ThreeDeeRender/transforms/SortedList.ts
```diff
@@ -0,0 +1,84 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+export class SortedList<K, V> {
```
defunctzombie, line 5: Maybe you wanna let this hang out in here for a bit before sharing but this could be a good candidate for `@foxglove/den/collection` - we have some stuff in there already.

Don't know how far along this draft is but it would be good to add tests and comments for this class if you do end up sharing it to den.

packages/den/collection/SortedList.ts
```diff
@@ -0,0 +1,112 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+/**
+ * A list of values with type `V` sorted by numeric keys of type `K`
+ * (either number or BigInt). Keys are unique, so setting a value with the same
+ * key as an existing entry is a replacement operation.
+ */
+export class SortedList<K, V> {
```
jtbandes, line 10: I wouldn't call this a List, since lists usually have a single type parameter... How about SortedMap or OrderedMap? Or maybe ListMap / ArrayMap (borrowing names from Java)

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -794,10 +796,24 @@ export class Renderer extends EventEmitter<RendererEvents> {
...
+        FRAME_NOT_FOUND,
+        `Frame "${this.followFrameId}" not found, rendering in "${this.renderFrameId}"`,
+      );
+    } else {
+      this.settings.errors.clearPath(FOLLOW_TF_PATH);
+    }
   }
 
-  private _updateMaterials(resolution: THREE.Vector2): void {
+  private _updateMaterials(): void {
```
jtbandes, line 810: nit: maybe rename to updateResolution since that's all it does?

packages/studio-base/src/panels/ThreeDeeRender/renderables/PointCloudsAndLaserScans.ts
```diff
@@ -196,6 +198,13 @@ export class PointCloudsAndLaserScans extends SceneExtension<PointCloudAndLaserS
       renderable.visible = renderable.userData.settings.visible;
       if (!renderable.visible) {
         this.renderer.settings.errors.clearPath(path);
+        const pointsHistory = renderable.userData.pointsHistory;
+        for (let i = 0; i < pointsHistory.length - 1; i++) {
+          const entry = pointsHistory[i]!;
+          entry.points.geometry.dispose();
+          renderable.remove(entry.points);
+        }
+        pointsHistory.splice(0, pointsHistory.length - 1);
```
jtbandes, line 207: splice returns the deleted elements – you could do `for (const entry of pointsHistory.splice(...))`

jtbandes: Just curious, how did you track down the fact that this dispose() was missing?

jtbandes (approved):

defunctzombie (changes requested):
SortedList needs some more thought if its gonna go into den

packages/studio-base/src/panels/ThreeDeeRender/renderables/PointCloudsAndLaserScans.ts
jhurliman: It was either the draw count or geometry count, one of those counters was not dropping when I toggled the visibility of a LaserScan topic with decay set.

defunctzombie (approved):

PR #3774, opened 2022-07-05
Move Open Dialog over to use MUI Components

2metres:
**User-Facing Changes**
Minimal visual differences to OpenDialog

**Description**
Move Open Dialog over to use MUI Components


defunctzombie:
image

The balance of the left and right sides here feels worse to me than the existing dialog. I haven't tried opening some files but there is also less space for the recents list which sometimes has longer file names.

2metres:
> The balance of the left and right sides here feels worse to me than the existing dialog. I haven't tried opening some files but there is also less space for the recents list which sometimes has longer file names.

The internal grid is unchanged, so should behave the same as before.
Screen Shot 2022-07-15 at 9 46 27 am

defunctzombie (approved):
Code changes seem fine. Visually I think something feels slightly off.

PR #3536, opened 2022-06-08
Added RGBA8 Support for Image Display

osholsonel20:
**User-Facing Changes**
Added support for images with `rgba8` format in the Image panel.

**Description**
Partially addresses https://github.com/foxglove/studio/issues/1273

Issue #1273, opened 2021-06-18
Support rgba8/bgra8 images


jtbandes:
Requested in https://github.com/foxglove/studio/discussions/1268#discussioncomment-889978 and https://github.com/cruise-automation/webviz/issues/739

The Image panel currently supports rgb8/bgr8, but not rgba8/bgra8 images. Some relevant references:

* http://docs.ros.org/en/jade/api/sensor_msgs/html/namespacesensor__msgs_1_1image__encodings.html
* http://wiki.ros.org/cv_bridge/Tutorials/ConvertingBetweenROSImagesAndOpenCVImagesPython

WaelKarkoub:
Hello, will you guys be supporting bgra8 in the near future? Having the same issues

jtbandes:
This isn't currently on our roadmap, but we would be happy to review a pull request if anyone is interested in contributing. It should be relatively straightforward to add/modify the code here:

https://github.com/foxglove/studio/blob/main/packages/studio-base/src/panels/ImageView/lib/decodings.ts

https://github.com/foxglove/studio/blob/735fff89303a526ea10af595c3cdf0725fab46ba/packages/studio-base/src/panels/ImageView/lib/renderImage.ts#L139-L145

norbertmarko:
Hi! 
I've seen that you have added `rgba8` support and thank you for this! Unfortunately, our `.bag` files contain Images in `bgra8` format. Is it possible to add support for this type with a little alteration of the new commit? We would be very grateful for it!

defunctzombie:
@norbertmarko We'd be happy to review a PR if you want to implement bgra8 support. It should be easy to add by copying the `rgba8` decode function and changing the fields around.

jhurliman:
Fixed in https://github.com/foxglove/studio/pull/3796

PR #3780, opened 2022-07-06
Add BGRA8 support for Image Panel

OleKet:
**User-Facing Changes**
Added support for images with `bgra8` format in the Image panel.

**Description**
Adresses the part of #1273, which #3536 has not adressed.

### DIFF
--- a/packages/studio-base/src/panels/Image/lib/decodings.ts
+++ b/packages/studio-base/src/panels/Image/lib/decodings.ts
@@ -88,6 +88,28 @@ export function decodeRGBA8(
   }
 }
 
+export function decodeBGRA8(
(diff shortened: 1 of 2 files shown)


jtbandes (commented):
Thanks! Could you also add it to this file so we get support in the 3D (Beta) panel? https://github.com/foxglove/studio/blob/main/packages/studio-base/src/panels/ThreeDeeRender/renderables/Images.ts#L523

jhurliman:
Thank you for the contribution! I've ported this code into the PR at https://github.com/foxglove/studio/pull/3796 which implements the change for both Image panel and 3D (Beta) panel.

PR #3781, opened 2022-07-06
Delete unused Sparkline component

defunctzombie:
**User-Facing Changes**
None

**Description**
Delete unused Sparkline component


defunctzombie:
Hm - guess it is used. I missed that in search somehow.

Issue #3676, opened 2022-06-22
Popup to notify existing users about new panel settings location


amacneil:
Some users are confused because we didn't tell them where we moved panel settings (e.g. https://github.com/foxglove/studio/issues/3270).

Image panel, Map panel, and New3D especially now rely heavily on the panel settings. We should show existing users a one-time tooltip popup to notify them of this change.

amacneil:
@foxymiles @2metres can we ship something for this this week?

I suggest a small popup that shows up once per person, grab the first panel that has a settings cog in their layout, and show the popup next to that. New users should not see the popup at all.

A quick way I can think of to differentiate new vs existing users is check the analytics ([first_seen_month](https://github.com/foxglove/studio/blob/main/packages/studio-base/src/services/AmplitudeAnalytics.ts#L56)). There might be a cleaner way to do this but it would work fine.

Not sure where/how you want to store the "dismissed" state.

foxymiles:
@defunctzombie @amacneil Would we should this even for users that use the app for the first time a year from now?

foxymiles:
@amacneil I don't see a way to fetch data back from analytics to check if this is a user we've already logged.

amacneil:
> @defunctzombie @amacneil Would we show this even for users that use the app for the first time a year from now?

My vote is no. As I said above "New users should not see the popup at all".

Although @foxymiles you're right I also can't see any way to read user properties in analytics. So I guess we need to track in localstorage alongside some "first seen" timestamp so we can avoid showing it to new users. You could probably initialize "first seen" from the creation dates of layouts or something.

defunctzombie:
What's wrong with showing an indicator to new users that they can find panel settings by clicking the settings cog? Do we think we won't get it right and it will be annoying?

amacneil:
I think the tooltip should say "FYI we have moved settings here". I find it very annoying when I sign up for an app as a new user and the first thing I have to do is click through a bunch of "hey we have a new feature, hey we moved xyz" when I have never used the app before. We should solve first-run experience separately.

defunctzombie:
We've decided the window for informing users of this change has passed. In the future we will provide a transition notice alongside landing sweeping UI changes.

PR #3783, opened 2022-07-07
Show one time tooltip informing users that settings have migrated to the sidebar

2metres:
**User-Facing Changes**
Screen Shot 2022-07-07 at 10 33 00 am

**Description**
Show users that things have changed recently on Map, Image, 3D and URDF panel settings.

Fixes https://github.com/foxglove/studio/issues/3676

### DIFF
--- a/.storybook/main.ts
+++ b/.storybook/main.ts
@@ -12,6 +12,11 @@ module.exports = {
   core: {
     builder: "webpack5",
   },
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  env: (config: any) => ({
+    ...config,
+    STORYBOOK_MODE: true,
+  }),
 
   // Carefully merge our main webpack config with the Storybook default config.
   // For the most part, our webpack config has already been designed to handle
--- a/packages/studio-base/src/AppSetting.ts
+++ b/packages/studio-base/src/AppSetting.ts
@@ -30,6 +30,7 @@ export enum AppSetting {
   HIDE_SIGN_IN_PROMPT = "hideSignInPrompt",
   LAUNCH_PREFERENCE = "launchPreference",
   SHOW_OPEN_DIALOG_ON_STARTUP = "ui.open-dialog-startup",
+  SHOWN_PANEL_CHANGE_WARNINGS = "ui.shown-panel-change-warnings",
 
   // Dev only
   ENABLE_LAYOUT_DEBUGGING = "enableLayoutDebugging",
--- a/packages/studio-base/src/components/PanelToolbar/PanelToolbarControls.tsx
+++ b/packages/studio-base/src/components/PanelToolbar/PanelToolbarControls.tsx
@@ -14,7 +14,8 @@
 import SettingsIcon from "@mui/icons-material/Settings";
 import { useCallback, useContext } from "react";
 
-import PanelContext from "@foxglove/studio-base/components/PanelContext";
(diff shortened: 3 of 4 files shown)


2metres:
@foxymiles I could use some help wiring up the logic so that:

- Only one tooltip shows at a time
- Once dismissed it is no longer shown
- When clicking away the tooltip is dismissed

packages/studio-base/src/components/PanelToolbar/PanelToolbarControls.tsx
```diff
@@ -42,7 +48,13 @@ export const PanelToolbarControls = React.memo(function PanelToolbarControls({
   menuOpen,
   setMenuOpen,
 }: PanelToolbarControlsProps) {
+  const [shownChangeWarnings, setShownChangeWarnings] = useLocalStorage<Record<string, boolean>>(
+    AppSetting.SHOWN_PANEL_CHANGE_WARNINGS,
```
jtbandes, line 52: Doesn't really make sense to be using AppSetting with `useLocalStorage` because we have a separate AppSettings mechanism that this enum was meant for.

.storybook/main.ts
```diff
@@ -12,6 +12,11 @@ module.exports = {
   core: {
     builder: "webpack5",
   },
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  env: (config: any) => ({
+    ...config,
+    STORYBOOK_MODE: true,
+  }),
```
jtbandes, line 19: Let's not add an env for this purpose, can we use a regular React context to disable the change warnings? cc @defunctzombie for ideas on how to control this

defunctzombie: Agreed. Just like I don't believe code should know it is running under test - I don't believe code should know it is running in storybook unless it is a last resort.

packages/studio-base/src/AppSetting.ts
```diff
@@ -30,6 +30,7 @@ export enum AppSetting {
   HIDE_SIGN_IN_PROMPT = "hideSignInPrompt",
   LAUNCH_PREFERENCE = "launchPreference",
   SHOW_OPEN_DIALOG_ON_STARTUP = "ui.open-dialog-startup",
+  SHOWN_PANEL_CHANGE_WARNINGS = "ui.shown-panel-change-warnings",
```
defunctzombie, line 33: These are definitely not app settings tho since the user is not going to be controlling these as a preference.

Are these warnings? These are just notices right? Arguably guides in the app.

defunctzombie:
How does this avoid showing itself for new users?

I'm still not convinced that we should be taking users down a path of callouts and guides over these alternatives. Maybe I am the lone dissenter but I do not believe we need to do this change at all right now.

On the wording side, I'm not a fan of "We're making some changes" - we're always making some changes. I'd prefer more direct language like you have at the end "Click ⚙️ to configure additional panel settings".

I remain un-convinced that this is the right start to guiding users around our changes. I believe the ship for notifying folks about this change has largely sailed (so to speak) and we can think about the future where we maybe show a user some "whats new" page.

.storybook/main.ts
foxymiles: I couldn't think of another way to exclude these without touching the 30+ stories this affects. Any ideas?

2metres:
Abandoning this PR following a conversation with @defunctzombie and @foxymiles today.

PR #3788, opened 2022-07-07
Scope global styles to the #root container

bclinkinbeard:
**User-Facing Changes**

None

**Description**

Scopes the global styles to the #root container so embedding Studio within another app doesn't affect things outside its parent node.


defunctzombie:
Closing this out in favor of https://github.com/foxglove/studio/pull/3802. We should be able to land 3802 once we remove fluentui. I don't have an exact timeline for that removal because we want to give folks time to migrate off the 3d (legacy) panel themselves.

PR #3791, opened 2022-07-07
New3D font atlas-based labels

jtbandes:
**User-Facing Changes**
Improved performance of text markers in the 3D (Beta) panel.

**Description**
Resolves https://github.com/foxglove/studio/issues/3742

Adds `@foxglove/three-text` package to the monorepo with SDF atlas-based label drawing, replacing existing `Labels` with a new `LabelPool`. All label instances share the same texture.

References/inspiration:
- https://github.com/Jam3/three-bmfont-text
- https://github.com/protectwise/troika/tree/main/packages/troika-three-text
- https://github.com/Shags/react-msdf-text
- https://github.com/cruise-automation/webviz/blob/master/packages/regl-worldview/src/commands/GLText.js, https://github.com/cruise-automation/webviz/pull/291, https://github.com/cruise-automation/webviz/pull/363, https://github.com/cruise-automation/webviz/pull/687
- https://github.com/thi-ng/umbrella/tree/develop/packages/webgl-msdf
- https://github.com/visgl/deck.gl/blob/master/modules/layers/src/text-layer/font-atlas-manager.ts
- https://css-tricks.com/techniques-for-rendering-text-with-webgl/


packages/studio-base/src/panels/ThreeDeeRender/renderables/FrameAxes.ts
```diff
@@ -138,13 +146,46 @@ export class FrameAxes extends SceneExtension<FrameAxisRenderable> {
...
+    colorScheme: "dark" | "light",
+    backgroundColor: THREE.Color | undefined,
+  ): void {
+    const foreground = colorScheme === "dark" ? 1 : 0;
+    this.labelForegroundColor = foreground;
+    this.labelBackgroundColor.setRGB(1 - foreground, 1 - foreground, 1 - foreground);
+    if (backgroundColor) {
+      this.labelForegroundColor =
+        getLuminance(backgroundColor.r, backgroundColor.g, backgroundColor.b) > 0.5 ? 0 : 1;
+      this.labelBackgroundColor.copy(backgroundColor);
```
jhurliman, line 159: This works because we only use pure black or pure white, but if any other value was used it would need sRGB -> linear conversion

packages/studio-base/webpack.ts
```diff
@@ -244,6 +244,7 @@ export function makeConfig(
         Buffer: ["buffer", "Buffer"],
         process: ["@foxglove/studio-base/util/process", "default"],
         setImmediate: ["@foxglove/studio-base/util/setImmediate", "default"],
+        THREE: "three",
```
jhurliman, line 247: What does this do?

jhurliman (approved):

packages/studio-base/webpack.ts
jtbandes: This might not be necessary anymore, I’ll check. It’s used to work around packages in node_modules that use THREE without actually importing it (assuming it’s a global), by telling webpack to inject the import automatically.

packages/studio-base/src/panels/ThreeDeeRender/renderables/FrameAxes.ts
jtbandes: Do you think it would be better to put that in the setBackgroundColor method or leave it to the callers?

jhurliman: So far I've been following a convention that `THREE.Color` objects are in linear space (they've already been converted) and `ColorRGBA`, hex strings, or any other representation are in sRGB space (they need to be converted when turning it into a `THREE.Color`).

PR #3792, opened 2022-07-07
Abort panel drag within a tab with no destination

### LINKED ISSUE #3631: Dropping a panel between two panels causes it to replace entire tab
**Description**
1. Create a tabbed layout
2. Drag and drop one of your panels exactly between two other panels
3. The entire tab contents is replaced with that panel 😱 

I tried this with a non-tabbed layout and it behaves correctly, the drop is rejected and the dragged panel stays in its original position.

https://user-images.githubusercontent.com/637671/174185737-026e5cc1-6a26-49ee-ab51-b8162f0ed754.mov

foxymiles:
**User-Facing Changes**
This fixes an issue with dragging panels within a tab.

**Description**
The fix here is to check if we're dragging within the same tab. If so and we have no position or destination for the drag we simply abort the drop.

Fixes #3631

### DIFF
--- a/packages/studio-base/src/providers/CurrentLayoutProvider/reducers.ts
+++ b/packages/studio-base/src/providers/CurrentLayoutProvider/reducers.ts
@@ -692,6 +692,11 @@ const endDrag = (panelsState: PanelsState, dragPayload: EndDragPayload): PanelsS
     return config ? { id, config } : undefined;
   });
 
+  // If dragging within the same tab without position & destination just cancel the drag.
+  if (withinSameTab && position == undefined && destinationPath == undefined) {
+    return { ...panelsState, layout: originalLayout, configById: originalSavedProps };
+  }
+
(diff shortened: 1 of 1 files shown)


jtbandes:
Does a missing position/destination ever result in a successful drag? Or why do we only check this if the drag is within a tab?

foxymiles:
> Does a missing position/destination ever result in a successful drag? Or why do we only check this if the drag is within a tab?

I'm not sure. I was just trying to make this as targeted a fix as possible since the problem was only reported dragging within a tab and this logic is complex.

foxymiles:
If you look further down we do have actions that don't require position & destination. It's the second one here that's causing the bad behavior in the originally reported issue:

https://github.com/foxglove/studio/blob/53600da21fb63c5b5944406f48703f94bb87b010/packages/studio-base/src/providers/CurrentLayoutProvider/reducers.ts#L719-L745

jtbandes (approved):

PR #3796, opened 2022-07-08
Add BGRA8 decoding support

jhurliman:
**User-Facing Changes**

- `"bgra8"` encoding is now supported in `Image` panel and `3D (Beta)` panel

**Description**

- Add decodeBGR8() and use it
- Move shared code between New3D and Image to den
- Document and attribute the source for two-pass line rendering technique


packages/den/image/PinholeCameraModel.ts
```diff
@@ -11,22 +11,49 @@
...
+  number, number, number, number,
+];
 
 type Vec8 = [number, number, number, number, number, number, number, number];
 
+type FloatArray = number[] | Float32Array | Float64Array;
+
+type DistortionModel = "plumb_bob" | "rational_polynomial" | "";
+
+export type CameraInfo = Readonly<{
```
defunctzombie, line 33: Would be nice to organize this into its own file. CameraInfo.ts? Or should we just have a types.ts in image folder?

packages/den/package.json
```diff
@@ -11,19 +11,5 @@
     "name": "Foxglove Technologies",
     "email": "support@foxglove.dev"
   },
-  "homepage": "https://foxglove.dev/",
```
defunctzombie, line 14: Why the changes here?

packages/den/image/PinholeCameraModel.ts
```diff
@@ -11,22 +11,49 @@
...
 
-import {
-  CameraInfo,
-  Matrix3,
-  Matrix3x4,
-  MutablePoint,
-  MutablePoint2D,
-  Point2D,
-} from "@foxglove/studio-base/types/Messages";
+type Vector2 = { x: number; y: number };
```
defunctzombie, line 14: Should these live in some math package? Or is the idea to keep these internal to image?

defunctzombie (commented):
👍 on movement to den (or packages generally) - I prefer that over dumping into utils folder in studio-base for sure.

I don't have strong thoughts on whether this goes in den or some new package in packages; I think we did have in mind that den would be some generally applicable libs rather than domain specific things but thats only a loose thought. @jtbandes maybe feels differently, if not then into den it goes.

I do think that putting something in den should get it a few more comments tho it seems like a number of the things moved had some level of comments already but might be worth a quick scan to see if a word or two could help with understanding anything that was moved.

packages/den/package.json
jhurliman: The information below this line is all wrong, and didn't actually do anything.

packages/den/image/PinholeCameraModel.ts
jhurliman: I wasn't sure if it was worth moving from a standalone package to having a dependency for simple internal types like this.

PR #3802, opened 2022-07-08
Make GlobalCss usable outside of <App>

defunctzombie:
**User-Facing Changes**
Using `<App>` from `@foxglove/studio-base` no longer takes over your html and body styling.

**Description**
The styling for the `<App>` component is now fully scoped to the container the app runs in. The `<GlobalCss>` component is removed from App and applied within `desktop` and `web` respectively to remove some padding and apply fallback colors and fonts for the body - tho these are not strictly necessary.

We are able to do this now that fluentui is removed from our UI. MUI properly places context menu and tooltip elements within the component tree of our <App> and those elements receive the correct styling.

### DIFF
--- a/packages/studio-base/src/components/CssBaseline.tsx
+++ b/packages/studio-base/src/components/CssBaseline.tsx
@@ -13,6 +13,38 @@ import { fonts } from "@foxglove/studio-base/util/sharedStyleConstants";
 
 const useStyles = makeStyles()(({ palette, typography }) => ({
   root: {
+    // container styling
+    height: "100%",
+    width: "100%",
+    display: "flex",
+    flexDirection: "column",
+    position: "relative",
+    flex: "1 1 100%",
+    overflow: "hidden",
+    background: palette.background.default,
+    color: palette.text.primary,
+    font: "inherit",
+    fontSize: typography.body2.fontSize,
+    fontFeatureSettings: fonts.SANS_SERIF_FEATURE_SETTINGS,
+    fontFamily: typography.body2.fontFamily,
+    fontWeight: typography.body2.fontWeight,
+    zIndex: 0,
+    boxSizing: "border-box",
+
+    // Prevent scroll "bouncing" since the app workspace is not scrollable. Allows individual
+    // scrollable elements to be scrolled without the whole page moving (even if they don't
+    // preventDefault on scroll events).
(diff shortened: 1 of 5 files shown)


defunctzombie:
@2metres is my analysis on why we need the fonts in the body correct here? It is a fluent thing and that for mui we are ok cause mui puts its context menus under a theme provider?

2metres:
@defunctzombie I believe that is correct MUI puts it's poppers just inside the root element whereas Fluent UI puts it in the body

defunctzombie:
This PR's time will come. It waits as a reminder to me.

defunctzombie:
@2metres I'm struggling with the padding on the navbar buttons. That's the only thing that looks out of place to me compared to current main.

defunctzombie:
I could trim down GlobalCss some more if we desire.

foxymiles (approved):
🎉

packages/studio-base/src/components/CssBaseline.tsx
```diff
@@ -13,6 +13,38 @@ import { fonts } from "@foxglove/studio-base/util/sharedStyleConstants";
...
+    overflow: "hidden",
+    background: palette.background.default,
+    color: palette.text.primary,
+    font: "inherit",
+    fontSize: typography.body2.fontSize,
+    fontFeatureSettings: fonts.SANS_SERIF_FEATURE_SETTINGS,
+    fontFamily: typography.body2.fontFamily,
+    fontWeight: typography.body2.fontWeight,
+    zIndex: 0,
+    boxSizing: "border-box",
```
2metres, line 32: I think messing with the `box-sizing` is doing more harm than good

packages/studio-base/src/components/CssBaseline.tsx
```diff
@@ -13,6 +13,38 @@ import { fonts } from "@foxglove/studio-base/util/sharedStyleConstants";
...
+    overscrollBehavior: "none",
+
+    // https://github.com/necolas/normalize.css/blob/master/normalize.css#L12
+    lineHeight: 1.15,
+
+    /// --- child and element styling follows ---
+
+    "*, *:before, *:after": {
+      boxSizing: "inherit",
+    },
```
2metres, line 46: I think messing with the `box-sizing` is doing more harm than good

PR #3803, opened 2022-07-08
Display buffering indicator in scrubber

defunctzombie:
**User-Facing Changes**
When experimental players are loading data the playback bar shows a loading animation.

**Description**
Add a `loading` property to the ProgressPlot component. When `loading` is true, the component shows a moving stripe animation in the un-filled ranges to indicate loading is happening.

Add a new `BUFFERING` PlayerPresence for players to indicate they are buffering. A BUFFERING player is already initialized but may not be able to complete playback because it is waiting for data.

Refactor the ProgressPlot to remove canvas. There's no need to involve canvas to display what is typically a handful of rectangles. This is accomplished easier with some divs. The browser scales and renders these divs without having to use autosizing logic.

Fixes: #3777


2metres:
Colors need some tweaks, I'll take a look

packages/studio-base/src/components/PlaybackControls/Scrubber.tsx
```diff
@@ -13,94 +12,60 @@ import {
...
+  }),
 }));
 
 const selectStartTime = (ctx: MessagePipelineContext) => ctx.playerState.activeData?.startTime;
 const selectCurrentTime = (ctx: MessagePipelineContext) => ctx.playerState.activeData?.currentTime;
 const selectEndTime = (ctx: MessagePipelineContext) => ctx.playerState.activeData?.endTime;
-const selectProgress = (ctx: MessagePipelineContext) => ctx.playerState.progress;
+const selectRanges = (ctx: MessagePipelineContext) =>
+  ctx.playerState.progress.fullyLoadedFractionRanges;
+const selectPrecense = (ctx: MessagePipelineContext) => ctx.playerState.presence;
```
foxymiles, line 68: Should be `selectPresence`.

foxymiles (approved):
Looks nice!

Feels like at some point we're going to need more fined grained player status than the single `presence` state.

defunctzombie:
Jacob noted that one of the stories (https://www.chromatic.com/test?appId=603ec8bf7908b500231841e2&id=62ce5a06a4689a0136e0d41a) looks incorrect. The range has changed from being "filled" to being "empty". Needs investigation before merging.

Additional feedback is that the PR for buffering https://github.com/foxglove/studio/pull/3772 should land first to avoid some of the "flickering" sometimes present when the player is working with local files and triggering the timeouts constantly during seek or playback.

2metres:
> Jacob noted that one of the stories (https://www.chromatic.com/test?appId=603ec8bf7908b500231841e2&id=62ce5a06a4689a0136e0d41a) looks incorrect. The range has changed from being "filled" to being "empty". Needs investigation before merging.

That change is not a change from filled to empty but an overal contrast bump from the MUI greys being slight lighter

2metres:
No more muiStyled namaste 🧘

jtbandes:
> That change is not a change from filled to empty but an overal contrast bump from the MUI greys being slightly lighter

I don't think so – there are some that are just a change in grays:

image

But some stories look like they're actually changing from filled to empty. If it were just the same contrast change as the other stories, the result would be darker, not lighter:

image

2metres (approved):

esthersweon:
Will not add these "user-facing changes" to release notes, as experimental players are off by default.

PR #3814, opened 2022-07-11
Add warning to ros2 native connection option

defunctzombie:
**User-Facing Changes**
Screen Shot 2022-07-11 at 8 22 07 AM

**Description**
The ROS 2 native connection does not work with custom messages and generally has more gotchas due to the complexities of DDS. Show a warning and recommend the user use rosbridge instead.


defunctzombie:
Let's start with this warning for now and consider dropping ROS 2 socket support entirely in a future release. If we do drop ROS 2 socket support entirely, we should enhance the Rosbridge connection panel to have a quicker path to getting started for ROS 1 and ROS 2 users (maybe link directly to rosbridge getting-start docs in our docs portal)

Issue #3784, opened 2022-07-07
Node Playground code doesn't work with custom message types


jackiescanlon:
**Description**
My node playground panel won't work when there are custom data types present in the live data that aren't subscribed to somewhere in the config.

- Version: 1.16.0
- OS: Ubuntu 20.04
- Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native): ROS1 connection. Bag file loaded directly into Foxglove does not seem to have this issue.
- ROS distro (if applicable): Noetic

**Steps To Reproduce**

[sine_wave.zip](https://github.com/foxglove/studio/files/9063854/sine_wave.zip) contains bag file and foxglove config file referenced below.

1. Play back the bag file on loop to simulate what it would look like if this data was being published from a live robot
* `roscore`
* `rosbag play sine_wave_bool_stamped.bag -l`
* Contains two logged messages:
  * `/sine_wave_stamped`, of custom message type `merlin_common/Float32Stamped`, which looks like this:
```
Header header
float32 value
```
  * `/bool_stamped`, of custom message type 'merlin_common/BoolStamped`, which looks like this:
```
Header header
bool value
```
2. Open Studio and start a ROS 1 connection
3. Import this foxglove config
* Second tab contains a node playground panel that reads in `/sine_wave_stamped` and converts it from meters to feet and republishes it as `/studio_node/sine_wave_stamped_ft`
```
import { Input, Message } from "./types";

export const M_TO_FT = 3.28084;

type Output = Message<"merlin_common/Float32Stamped">;

export const inputs = ["/sine_wave_stamped"];
export const output = "/studio_node/sine_wave_stamped_ft";

export default function node(event: Input<"/sine_wave_stamped">): Output {
  return {
    header: event.message.header,
    value: event.message.value * M_TO_FT,
  };
}
```
* First tab contains a plot of `/studio_node/sine_wave_stamped_ft`

**Expected Behavior**
The node playground panel in the second tab works without errors and causes the plot on the first tab to fill with the converted data

**Actual Behavior**
1. I receive an error in the Node playground panel:
![image](https://user-images.githubusercontent.com/43035541/177779161-475cabdb-802c-4133-8b34-6a15e2138b29.png)

2. The plot on the first tab always shows a bad topic:
![image](https://user-images.githubusercontent.com/43035541/177778997-4f33267d-27b5-46af-ac1f-551451f3c09f.png)

**Workaround**
[...]

PR #3818, opened 2022-07-11
NodePlayground: Use empty datatype definition for unknown datatypes

defunctzombie:
**User-Facing Changes**
Node playground scripts work in more situations where custom datatype are involved.

**Description**
When a topic references a datatype that is not yet known to the data source, the types library would fail to compile and run. The user would need to manually add a subscriber (some panel) for a topic of the missing datatype to unblock node playground.

This changes fixes this requirement by using empty datatype placeholders for unknown datatypes so the user script compiles and runs.

Fixes: #3784


foxymiles (approved):

jtbandes:
Why was the datatype unknown/missing until a subscription was added?

PR #3821, opened 2022-07-11
Update sidebar settings icons to be consistent with Panel Settings button

2metres:
**User-Facing Changes**

** Before/After** 
Screen Shot 2022-07-12 at 9 50 34 am

**Description**
Update cog icon on 1st and 4th icon to use 6 toothed cog


defunctzombie (approved):
Should the icon in each panel match the sidebar icon? We don't have to do that in this PR - since this PR is already an improvement but not having these icons match does seem strange.

2metres:
@defunctzombie not sure what you mean when you say

> Should the icon in each panel match the sidebar icon?

I think using the cog on the panel and then the Panel + Cog icon is fine.

defunctzombie:
> I think using the cog on the panel and then the Panel + Cog icon in the sidebar is fine.

Thats what I was referring to - that the icon you click on the sidebar is not the same one you click on the panel.

On your new panel setting icon w/cog - maybe the cog should go in the upper right corner since that is where it is in the panel.

Issue #3770, opened 2022-07-01
New3D panel can't display marker with no /tf topic


jtbandes:
Load the following bag into Studio: [output.db3.zip](https://github.com/foxglove/studio/files/9024961/output.db3.zip)

There is only one topic with markers, and no /tf:
<img width="705" alt="image" src="https://user-images.githubusercontent.com/14237/176820889-25217758-de89-4e36-915f-91645b47ad57.png">

The New3D panel's frame list is empty and nothing appears in the scene: 
<img width="386" alt="image" src="https://user-images.githubusercontent.com/14237/176820807-56296f2e-a04b-4927-852b-fba8d49e0c8c.png">

But it works in the old 3d panel:

<img width="585" alt="image" src="https://user-images.githubusercontent.com/14237/176820849-a369b494-6495-4863-9ff4-9029b0197d13.png">

PR #3823, opened 2022-07-12
[New3D] Scrape frame_id from MarkerArray

jhurliman:
**User-Facing Changes**

- `3D (Beta)` panel can now visualize `MarkerArray` topics when no TF messages are published

**Description**

Fixes https://github.com/foxglove/studio/issues/3770 using the same approach the previous 3D panel used

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
@@ -53,6 +53,7 @@ import { Poses } from "./renderables/Poses";
 import { MarkerPool } from "./renderables/markers/MarkerPool";
 import {
   Header,
+  MarkerArray,
   Quaternion,
   TFMessage,
   TF_DATATYPES,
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -551,16 +552,23 @@ export class Renderer extends EventEmitter<RendererEvents> {
...
-    // If this message has a top-level frame_id, scrape it
-    const maybeHasFrameId = message as Partial<{ frame_id: string }>;
-    if (typeof maybeHasFrameId.frame_id === "string") {
+    } else if (Array.isArray(maybeHasMarkers.markers)) {
+      // If this message has an array called markers, scrape the frame_id from the first marker
+      const marker = maybeHasMarkers.markers[0];
+      if (marker) {
+        const frameId = marker.header?.frame_id ?? "";
+        this.addCoordinateFrame(frameId);
+      }
```
jtbandes, line 569: Shouldn't we scrape from all markers? What if the markers are in different frames?

jtbandes (approved):

PR #3827, opened 2022-07-12
Treat layout sync errors as offline instead of showing an error toast

### LINKED ISSUE #3809: Repeating error of failing to connect to foxglove website
**Description**
When connected to a wifi without internet a warning of failure to connect to a foxglove domain keeps appearing every few minutes.

Probably trying to synchronise the layout or something like that. The message disappears very quickly but it happens every few minutes. Not critical but when using it a full day ends up becoming a bit annoying. 

Maybe after one failure to connect to network keep trying but not needed to replay the same error message.

We connect the PC to a private network that includes the cars/robots we want to monitor. This network does not include internet most of the times since it is in deserted or very unpopulated areas.

foxymiles:
**User-Facing Changes**
Makes layout sync failure notifications less intrusive.

**Description**
Instead of showing an error toast on layout syncs we just register an error and consider studio to be offline for syncing purposes. When a layout sync fails we show the broken cloud icon.

Screen Shot 2022-07-12 at 10 54 18 AM

Fixes #3809


defunctzombie:
I don't know if we already do this - but it would be nice to show the last error (if there is one) in a popover when the user hovers over the broken cloud icon. It would provide a convenient way to learn why I am seeing the broken cloud icon.

foxymiles:
> I don't know if we already do this - but it would be nice to show the last error (if there is one) in a popover when the user hovers over the broken cloud icon. It would provide a convenient way to learn why I am seeing the broken cloud icon.

We'd also need to plumb through a more informative error message than the vanilla fetch failure I think but yeah this would be a good addition.

PR #3828, opened 2022-07-12
Fix ros2 getMessages end time calculation

### LINKED ISSUE #3826: Last message from ROS2 bag is not seekable
**Description**
The attached ROS2 bag contains 8 messages on the `/rt_topo_occupy` topic, but seeking or playing to the end of the file stops at the 7th message.

- Version: web
- OS: macOS

**Steps To Reproduce**

[output.zip](https://github.com/foxglove/studio/files/9094763/output.zip)

1. Load this bag file. 
2. Seek or play to the end.
3. Observe the `/rt_topo_occupy` topic in the raw messages panel.

The last message should be the message with timestamp 1657629576529131008 but the last message shown has timestamp 1657629576224429056.

foxymiles:
**User-Facing Changes**
This fixes an issue with ros2 bag message reading.

**Description**
The fix here is to add 1 nsec to our calculation of the time range of the messages to fetch since the rosbag2-web implementation of `readMessages` returns messages with end time < the specified time, not <=.

Fixes #3826

### DIFF
--- a/packages/studio-base/src/randomAccessDataProviders/Rosbag2DataProvider.ts
+++ b/packages/studio-base/src/randomAccessDataProviders/Rosbag2DataProvider.ts
@@ -107,13 +107,9 @@ export default class Rosbag2DataProvider implements RandomAccessDataProvider {
       parsedMessageDefinitionsByTopic[topicDef.name] = fullParsedMessageDefinitions;
     }
 
-    // Add 1 nsec to the end time because rosbag2 treats the time range as non-inclusive
-    // of the exact end time.
-    const inclusiveEndTime = { sec: end.sec, nsec: end.nsec + 1 };
-
     return {
       start,
-      end: inclusiveEndTime,
+      end,
       topics,
       topicStats,
       connections,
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/randomAccessDataProviders/Rosbag2DataProvider.ts
```diff
@@ -139,8 +139,15 @@ export default class Rosbag2DataProvider implements RandomAccessDataProvider {
       return {};
     }
 
+    // Add 1 nsec to the end time because rosbag2 treats the time range as non-inclusive
+    // of the exact end time.
+    const inclusiveEndTime = { sec: end.sec, nsec: end.nsec + 1 };
```
jtbandes, line 144: What if end.nsec was 999999999? Do we need to fixTime or will the readMessages function do that internally?

jhurliman: It doesn't matter, because https://github.com/foxglove/rosbag2-web/blob/main/src/SqliteSqljs.ts#L117 calls https://github.com/foxglove/rostime/blob/main/src/timeUtils.ts#L252 which just returns integer nanoseconds.

jhurliman (approved):

Seraphli:
I test this change. It only works when clicking the forward button. It doesn't work when dragging the timeline.

PR #3825, opened 2022-07-12
[New3D] Bug fix: Create/remove axes/arrows on every `PoseArray` update

siddancha:
**User-Facing Changes**
Previously, rendering `PoseArrays` in the 3D (beta) panel crashed for the `Axis` or `Arrow` display types when the length of the `PoseArray` changed. Rendering now works.

**Description**
Previously, new `Axis`/`RenderableArrow` were only created when the settings changed, not on every `PoseArray` update. Therefore, whenever the length of the `PoseArray` increased, elements out of bounds of the `Axis`/`RenderableArrow` buffers were accessed, throwing an error. Now, `Axis`/`RenderableArrow` objects are added (or extra objects removed) from the buffers on every `PoseArray` update.

**Before:**

ROS bag containing `nav_msgs/Path` messages with increasing length:  [path.bag.zip](https://github.com/foxglove/studio/files/9094447/path.bag.zip)

Using "Axis": TypeError -- `Cannot read properties of undefined (reading 'position')`
Using "Arrow": TypeError -- `Cannot read properties of undefined (reading 'position')`

```
Call Stack
 setObjectPose
  packages/studio-base/src/panels/ThreeDeeRender/renderables/PoseArrays.ts:300:12
 PoseArrays._updatePoseArrayRenderable
  packages/studio-base/src/panels/ThreeDeeRender/renderables/PoseArrays.ts:275:21
 PoseArrays.addPoseArray
  packages/studio-base/src/panels/ThreeDeeRender/renderables/PoseArrays.ts:195:14
 PoseArrays.handleNavPath
  packages/studio-base/src/panels/ThreeDeeRender/renderables/PoseArrays.ts:125:18
 Renderer.addMessageEvent
  packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts:537:17
 eval
  packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx:306:22
 invokePassiveEffectCreate
  node_modules/react-dom/cjs/react-dom.development.js:23484:20
 HTMLUnknownElement.callCallback
  node_modules/react-dom/cjs/react-dom.development.js:3947:14
 Object.invokeGuardedCallbackDev
  node_modules/react-dom/cjs/react-dom.development.js:3996:16
 invokeGuardedCallback
  node_modules/react-dom/cjs/react-dom.development.js:4058:31
```

**After:**
Using "Axis":
image

Using "Arrow":
image

PR #3830, opened 2022-07-12
[New3D] Bug fix: Create axes/arrows as needed on every PoseArray update

jhurliman:
**User-Facing Changes**

- Fixed a crash in `3D (Beta)` when rendering PoseArrays that grew over time as axes or arrows

**Description**

This is a variation of https://github.com/foxglove/studio/pull/3825 (authored by [siddancha](https://github.com/siddancha)) that only grows arrays instead of growing+shrinking to avoid dispose+reallocate in situations where pose arrays are frequently changing lengths. Also includes two minor fixes to the followTf sidebar control.

- [New3D] Bug fix: Create axes/arrows as needed on every PoseArray update
- Fix continuously clearing/resetting followTf sidebar errors
- Show the currently rendered frame_id in fallback cases

178526172-9692e544-b490-4070-9126-7736b0e148c7

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/PoseArrays.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/PoseArrays.ts
@@ -260,6 +260,76 @@ export class PoseArrays extends SceneExtension<PoseArrayRenderable> {
     );
   }
 
+  _createAxesToMatchPoses(
+    renderable: PoseArrayRenderable,
+    poseArray: PoseArray,
+    topic: string,
+  ): void {
+    const scale = renderable.userData.settings.axisScale * (1 / AXIS_LENGTH);
+
+    // Update the scale and visibility of existing AxisRenderables as needed
+    const existingUpdateCount = Math.min(renderable.userData.axes.length, poseArray.poses.length);
+    for (let i = 0; i < existingUpdateCount; i++) {
+      const axis = renderable.userData.axes[i]!;
+      axis.visible = true;
+      axis.scale.set(scale, scale, scale);
+    }
+
+    // Create any AxisRenderables as needed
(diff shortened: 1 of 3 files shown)


jtbandes (approved):

packages/studio-base/src/panels/ThreeDeeRender/renderables/PoseArrays.ts
```diff
@@ -260,6 +260,76 @@ export class PoseArrays extends SceneExtension<PoseArrayRenderable> {
...
+    for (let i = renderable.userData.axes.length; i < poseArray.poses.length; i++) {
+      const axis = new Axis(topic, this.renderer);
+      renderable.userData.axes.push(axis);
+      renderable.add(axis);
+
+      // Set the scale for each new axis
+      axis.scale.set(scale, scale, scale);
+    }
+
+    // Hide any AxisRenderables as needed
```
siddancha, line 288: Hi @jhurliman @jtbandes 

If the pose array grows and then shrinks, lines 289-292 could lead to unnecessarily setting the visibility of a large number of already-invisible-axes to false. Wonder if this can be made more efficient by breaking the loop as soon as the first invisible axis is encountered e.g. by adding `if (!axis.visible) {.break; }` between lines 290 and 291.

siddancha: Feel free to ignore if you think the gains from this optimization will be marginal; pointing it out just in case.

jhurliman: It might be interesting to come up with a stress test bag and test this out. My initial hunch is the additional branch test per loop iteration will make it not worth it, but it's difficult to reason about modern CPU micro-behavior like this without testing.

PR #3836, opened 2022-07-13
Implement private registry syncing

foxymiles:
### User-Facing Changes
This implements extension syncing with a private registry.

### Description
This implements syncing of private registry extensions. The implementation depends on a few points:
1. All private extensions are installed into a separate private extension namespace.
2. All available private extensions are installed automatically.
3. Any existing private extensions that are not listed in the registry are uninstalled.
4. We sync all extensions on app startup and again every ten minutes.
5. Extensions installed via the private registry can't be manually uninstalled by the user.


packages/studio-base/src/services/ConsoleApi.ts
```diff
@@ -39,6 +39,16 @@ type DeviceCodeResponse = {
   interval: number;
 };
 
+type ExtensionResponse = {
+  activeVersion: string;
+  description: string;
```
bryfox, line 44: `description` is nullable.

`activeVersion` and `sha256Sum` are technically/documented as nullable as well, though you won't encounter that today.

packages/studio-base/src/services/IdbExtensionLoader.ts
```diff
@@ -13,6 +13,14 @@ import { IdbExtensionStorage } from "./IdbExtensionStorage";
...
 
+function parsePackageName(name: string): { publisher?: string; name: string } {
+  const res = /^@([^/]+)\/(.+)/.exec(name);
+  if (res == undefined) {
+    return { name };
+  }
+  return { publisher: res[1], name: res[2] as string };
+}
+
 function qualifiedName(namespace: ExtensionNamespace, info: ExtensionInfo): string {
```
bryfox, line 24: Should qualified name now include publisher? The registry API is unique on a combination of lower-cased publisher + name.

foxymiles: Seems like we don't need it since these are all going into a namespace unique to the org?

packages/studio-base/src/services/ConsoleApi.ts
foxymiles: Thanks will change. Can we exclude extensions with no `activeVersion` from the response so I can treat it as non-nullable? I can't do anything with extensions with no active version anyway.

bryfox: We've talked about adding more granular control over versions, which (depending on management UI requirements) could introduce the possibility for an extension to not have an active version. Other clients (again, for extension management) would still need to see the extension information in this case.

I'd prefer to leave these nullable to support those possible cases in the future without having to version the endpoint. Today, you'll always receive a string for version & shasum. If we don't want to check for an activeVersion in the response today, we can deal with it when it becomes needed in practice.

foxymiles: That makes sense. It's easy enough for me to ignore them.

packages/studio-base/src/services/IdbExtensionLoader.ts
bryfox: I'm not sure how this is used, but an org may use arbitrary publisher values in their private registry. The following (partial) packages are considered different extensions.

```json
[
{ "publisher": "a", "name": "extn" },
{ "publisher": "b", "name": "extn" }
]
```

foxymiles: I've incorporated the publisher into the qualified name. Well spotted.

packages/studio-base/src/types/Extensions.ts
```diff
@@ -2,6 +2,11 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
+/**
+ * Extensions are installed into separate namespaces enumerated here.
```
defunctzombie, line 6: Worth adding some comments about what org and local are. Its true that this type enumerates the namespaces so the existing comment doesn't really tell me much that I can't see from the code but if there was some comment that told me what local vs org are or pointed me at something that could educate me that might be more useful.

packages/studio-base/src/services/ExtensionLoader.ts
```diff
@@ -21,6 +21,5 @@ export interface ExtensionLoader {
   installExtension(foxeFileData: Uint8Array): Promise<ExtensionInfo>;
 
   // uninstall extension with id
-  // return true if the extension was found and uninstalled, false if not found
```
defunctzombie, line 24: We don't need to return true here anymore?

packages/studio-base/src/services/IdbExtensionLoader.ts
```diff
@@ -13,27 +13,41 @@ import { IdbExtensionStorage } from "./IdbExtensionStorage";
 
 const log = Log.getLogger(__filename);
 
-function qualifiedName(namespace: ExtensionNamespace, info: ExtensionInfo): string {
+function parsePackageName(name: string): { publisher?: string; name: string } {
```
defunctzombie, line 16: I thought we get the publisher from the package.json file rather than the package name?

packages/studio-base/src/components/OrgExtensionRegistrySyncAdapter.tsx
```diff
@@ -0,0 +1,123 @@
...
+    async (id: string) => {
+      const metaData = await api.getExtension(id);
+      const dataResponse = await fetch(metaData.foxe);
+      const data = new Uint8Array(await dataResponse.arrayBuffer());
+      await installExtension("org", data);
+    },
+    [api, installExtension],
+  );
+
+  // Use polling for now. To be replaced with some kind of notification mechanism.
```
defunctzombie, line 47: `To be` is getting awfully close to a `TODO` comment and I don't like TODO comments in the code. nitpick semantics but I'd feel better if this was reworded to something like `A notification mechanism could remove the need for polling.`. This way when we read this in two years we don't see `To be replaced` that never happened :D Or just remove the comment altogether and make an issue to track adding notifications if you want to remember tho I doubt we will forget if it starts to matter.

packages/studio-base/src/components/PanelLayout.tsx
```diff
@@ -215,22 +229,25 @@ export default function PanelLayout(): JSX.Element {
     },
     [changePanelLayout],
   );
+
+  if (registeredExtensions == undefined) {
+    return <LoadingState />;
```
defunctzombie, line 234: Does this mean we block displaying any panels to the user (even installed ones) until we've heard back from the extension registry? What happens if API is down or network is down? Does `state.registeredExtensions` have a timeout or some initial state where it is an empty array rather than undefined?

I think it is important that we can show the user the layout immediately rather than blocking on getting the API response or potentially installing the extensions. The individual panel that might need loading should itself show the loading indicator.

packages/studio-base/src/context/ExtensionRegistryContext.ts
```diff
@@ -22,22 +23,19 @@ export type ExtensionRegistry = {
   ) => Promise<ExtensionInfo>;
   loadExtension(id: string): Promise<string>;
   refreshExtensions: () => Promise<void>;
-  registeredExtensions: ExtensionInfo[];
-  registeredPanels: Record<string, RegisteredPanel>;
-  uninstallExtension(id: string): Promise<boolean>;
+  registeredExtensions: undefined | ExtensionInfo[];
+  registeredPanels: undefined | Record<string, RegisteredPanel>;
+  uninstallExtension: (namespace: ExtensionNamespace, id: string) => Promise<void>;
 };
```
defunctzombie, line 29: I can't comment on line 17 (cause github) but I think `type ExtensionRegistry` would be better as `interface IExtensionRegistry` since it has "methods" and is moving beyond a basic struct description and instead is meant to represent an interface that something can implement.

packages/studio-base/src/providers/ExtensionRegistryProvider.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+export default function ExtensionRegistryProvider({
+  children,
+  loaders,
+}: PropsWithChildren<{ loaders: readonly ExtensionLoader[] }>): JSX.Element {
+  const [store] = useState(createExtensionRegistryStore(loaders));
+
+  return (
+    <ExtensionRegistryContext.Provider value={store}>
+      <InitialRefreshAdapter />
```
defunctzombie, line 161: Why is the OrgExtensionRegistrySync not here? Or generally a function that is within the provider?

Maybe I've mis-understood what the ExtensionRegistryProvider is - does it provide the extension registry that is installed locally? I think its worth a comment for ExtensionRegistryProvider to describe its scope cause originally I'd thought this is where there would also be calls out to the API.

Maybe ExtensionRegistry is too generic/overloaded a term since we also have a "registry" for the organization's extensions?

defunctzombie (approved):
👍 left some thoughts on a few things that jumped out at me. I don't think any of them are a blocker or can't be tackled as follow-on. If they make sense then might be worth tackling in this PR or getting this merged so we can start testing it.

packages/studio-base/src/services/IdbExtensionLoader.ts
bryfox: This matches API logic, and I think desktop parsing as well. You may define a publisher in the `publisher` field or as a scoped package name (e.g. `@foxglove/turtlesim`). The former trumps the latter if both are used.

foxymiles: That's right the `publisher` is the primary field but we allow the `@publisher` prefix in the package name too.

packages/studio-base/src/services/ExtensionLoader.ts
foxymiles: Uninstall can't really be done transactionally so we're introducing a race condition here by trying to return a success status. Feels like an exception or just subscribing to the list of installed extensions (which is what we currently do) is more robust.

packages/studio-base/src/components/OrgExtensionRegistrySyncAdapter.tsx
foxymiles: I've updated the comment.

packages/studio-base/src/components/PanelLayout.tsx
foxymiles: No this is only shown while we're scanning already installed extensions. This prevents us from flashing a screenful of "missing panel" messages which are immediately replaced with panels. This doesn't require any interaction with the API.

packages/studio-base/src/context/ExtensionRegistryContext.ts
foxymiles: Yeah maybe we should change that in a followup PR? We've had `ExtensionRegistry` for 14 months.

packages/studio-base/src/providers/ExtensionRegistryProvider.tsx
foxymiles: Yeah I think we should rename `ExtensionRegistry` since it's only concerned with the locally installed extensions and this name is confusingly overloaded with the new "org" remote registry.

packages/studio-base/src/components/PanelLayout.tsx
foxymiles: I'm going to rename this to `installedExtensions` in a followup to make what's happening here more clear.

Issue #3692, opened 2022-06-23
Hide all topics by default (matches behavior of old3d)


jhurliman:
For now let's focus on feature parity rather than changing the default topic behavior of the 3d panel. It is not obvious that we should be turning on all topics by default.

defunctzombie:
@jhurliman can you please make this change this week - it's an easy change and been sitting on the backlog.

defunctzombie:
Re-opening this issue because I am still seeing transforms on the screen when I open a new 3d (Beta) panel.

Here's a comparison of the current 3d panel and 3d (beta). I've opened the nuscenes sample dataset and added both panels to a new empty layout.

![image](https://user-images.githubusercontent.com/84792/179270823-e8c3c57b-4da2-4fa5-b7b8-03b4e096e330.png)

defunctzombie:
Spoke in slack. We agreed to leave these visible for now since 3d (beta) unconditionally subscribes to transform messages so there's no additional "cost" in showing these by default.

We also mentioned that having a grid in the scene by default would be a nice visual to orient the display and make it obvious its not broken if there's nothing else to render.

PR #3839, opened 2022-07-14
[New3D] Make topics invisible by default

### LINKED ISSUE #3832: [New3D] Image topics do not render unless associated CameraInfo topic is also marked visible

jhurliman:
**User-Facing Changes**

- `3D (Beta)` will hide all topics by default. "Show All" and "Hide All" menu actions have been added for convenience
- `3D (Beta)` fixed default value handling for Grid settings
- `3D (Beta)` now renders Image topics even when the associated CameraInfo topic is hidden
- `3D (Beta)` only allows objects to be selected when the Object Selection dialog is open
- `3D (Beta)` shows info in the Object Selection dialog for all renderables
- `3D (Beta)` selection backdrop matches the active color scheme

**Description**

A collection of fixes. Fixes #3832 and #3692.


packages/studio-base/src/panels/ThreeDeeRender/ScreenOverlay.ts
```diff
@@ -2,34 +2,46 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
+import type { vec4 } from "gl-matrix";
```
jtbandes, line 5: maybe we should try to avoid gl-matrix in three.js code? I guess we might need to define our own version of some of these types if three.js doesn't provide them

jtbandes (approved):

Issue #3812, opened 2022-07-11
Incorrect plot panel lines for mcap file using experimental mcap player


defunctzombie:
**Description**
Loading a plot with the experimental mcap player produces incorrectly connected lines for this mcap file:

https://drive.google.com/file/d/19kZ8pIMZT6Bxx1YxPiDB0l-Zg9d-zmnz/view?usp=sharing

<img width="586" alt="Screen Shot 2022-07-11 at 7 20 09 AM" src="https://user-images.githubusercontent.com/84792/178286180-ec80fe61-cdd4-4cea-98c5-49fd7e333593.png">

When loading this same data using the non-experimental mcap player the plot line looks continuous as expected

foxymiles:
Extra context - after changing the timestamp method to `header.stamp` the points seem properly connected:

<img width="734" alt="Screen Shot 2022-07-12 at 8 24 16 AM" src="https://user-images.githubusercontent.com/93935560/178513684-30a5c9c5-aff3-4ab7-ba04-a665b8cb4585.png">

PR #3852, opened 2022-07-15
Upgrade @mcap/core

defunctzombie:
**User-Facing Changes**
Certain mcap files would have their message data read out of order. These files are now read in-order.

**Description**
Fixes: #3812


jtbandes (approved):
🐞

defunctzombie:
Screenshot of the plot panel using the experimental player to read an mcap file that previously produced an incorrect looking plot panel.

image

PR #3853, opened 2022-07-15
[New3D] Move measure tool distance into 3D scene

jtbandes:
**User-Facing Changes**
Measured distance in the 3D (Beta) panel is now shown along the measured line instead of next to the measurement tool button.

https://user-images.githubusercontent.com/14237/179317065-ea82f6ef-cf2a-4be5-9f8a-8db94f6cc1fd.mov

**Description**
Added sizeAttenuation=false support to `@foxglove/three-text`


jtbandes:
FYI @2metres I went with middle of line instead of endpoint because I was having trouble offsetting the text to always be _above_ the cursor independent of the camera angle.

jhurliman (approved):

PR #3855, opened 2022-07-15
Attempt to fix flaky PublishClickTool stories

jtbandes:
**User-Facing Changes**
None

**Description**
Stories introduced in #3837 were apparently flaky. After much debugging it turned out that some of the events were being sent too early, while the canvas still had height=0. This caused the worldSpaceCursorCoords calculation to fail.


foxymiles (approved):
Instead of hardcoded delays do you think we could create an async helper that polls the size of the canvas?

jtbandes:
That is effectively what the for loop is doing, because the canvas size is in the loop condition.

PR #3859, opened 2022-07-16
Refactor playback performance panel to use MUI styling

2metres:
**User-Facing Changes**
Minimal

**Description**
Remove uses of fluentui styles and use MUI typography component

### DIFF
--- a/packages/studio-base/src/components/Sparkline.tsx
+++ b/packages/studio-base/src/components/Sparkline.tsx
@@ -11,7 +11,7 @@
 //   found at http://www.apache.org/licenses/LICENSE-2.0
 //   You may not use this file except in compliance with the License.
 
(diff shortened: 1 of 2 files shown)


jtbandes (approved):
Looks ok, might want to improve the contrast in dark mode. Also I reopened #1177 because I noticed it while testing this PR.

defunctzombie (changes requested):
No more muiStyled please 🙏

defunctzombie (approved):

2metres:
@jtbandes a better indication of contrast issues is represented here: https://603ec8bf7908b500231841e2-qzcwbxmbbh.chromatic.com/?path=/story/components-sparkline--standard this PR increasees the contrast between the chart background and the line itself

PR #1880, opened 2021-09-16
Disable minification of identifiers in prod builds

jtbandes:
**User-Facing Changes**
None

**Description**
While we do have source maps sent to Sentry, it can be frustrating when an error boundary or other stack trace shows minified names instead of real function names.

<table>
<tr><th>Before</th><th>After</th></tr>
<tr><td>

WARNING in asset size limit: The following asset(s) exceed the recommended size limit (244 KiB).
This can impact web performance.
Assets: 
  ts.worker.905cdde0893ccfe4c3d3d52f504bbf5c.js (4.58 MiB)
  main.js (5.33 MiB)
  9691.js (523 KiB)
  4048.js (3.38 MiB)
  f4a990eab21714a17a87.bin (1.75 MiB)
  1271.js (361 KiB)
  78e0e3ce0190f1d1017d.wasm (279 KiB)
  5634.js (328 KiB)
  7185.js (389 KiB)
  6079.js (2.76 MiB)
  8411.js (559 KiB)
  5562.js (257 KiB)
  6974.js (423 KiB)
  8101.js (271 KiB)
  e46cedbe043b399854ef.wasm (1.06 MiB)
  6952.js (2.58 MiB)
  144.js (3.13 MiB)

WARNING in entrypoint size limit: The following entrypoint(s) combined asset size exceeds the recommended limit (244 KiB). This can impact web performance.
Entrypoints:
  main (5.33 MiB)
      main.js

</td>
<td>

WARNING in asset size limit: The following asset(s) exceed the recommended size limit (244 KiB).
This can impact web performance.
Assets: 
  editor.worker.a85ce25ec6767b7d9d2d3d88291dd6eb.js (313 KiB)
  ts.worker.905cdde0893ccfe4c3d3d52f504bbf5c.js (8.88 MiB)
  main.js (10 MiB)
  4112.js (346 KiB)
  9691.js (776 KiB)
  4048.js (5.29 MiB)
  8871.js (431 KiB)
  f4a990eab21714a17a87.bin (1.75 MiB)
  8243.js (583 KiB)
  1271.js (758 KiB)
  78e0e3ce0190f1d1017d.wasm (279 KiB)
  5634.js (639 KiB)
  7185.js (397 KiB)
  6079.js (5.68 MiB)
  8411.js (1.21 MiB)
  5562.js (701 KiB)
  4131.js (435 KiB)
  6974.js (1010 KiB)
  2392.js (256 KiB)
  8101.js (545 KiB)
[...]

PR #3860, opened 2022-07-18
Replace esbuild-loader with TerserPlugin and swc

defunctzombie:
**User-Facing Changes**
Faster web app load time because the main bundle is reduced in size from 10mb to 5mb.

**Description**
esbuild-loader and ESBuildMinifyPlugin were not working! Our builds were not minified. This changes actually has our builds minified now and uses swc for minification.

On my machine the web production build using swc took ~50 seconds while the default terser (non-swc) took ~63.

### DIFF
--- a/packages/studio-base/webpack.ts
+++ b/packages/studio-base/webpack.ts
@@ -3,7 +3,6 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
 import CircularDependencyPlugin from "circular-dependency-plugin";
-import { ESBuildMinifyPlugin } from "esbuild-loader";
 import ForkTsCheckerWebpackPlugin from "fork-ts-checker-webpack-plugin";
 import MonacoWebpackPlugin from "monaco-editor-webpack-plugin";
 import monacoPkg from "monaco-editor/package.json";
@@ -163,11 +162,6 @@ export function makeConfig(
           loader: "css-loader",
           options: { sourceMap: true },
         },
-        {
-          test: /\.css$/,
-          loader: "esbuild-loader",
-          options: { loader: "css", minify: !isDev },
-        },
         { test: /\.woff2?$/, type: "asset/inline" },
         { test: /\.(glb|bag|ttf|bin)$/, type: "asset/resource" },
         {
@@ -221,15 +215,6 @@ export function makeConfig(
             strict: true,
           },
         },
-      ],
-    },
-    optimization: {
-      removeAvailableModules: true,
-      minimizer: [
-        new ESBuildMinifyPlugin({
-          target: "es2020",
-          minifyIdentifiers: false, // readable error stack traces are helpful for debugging
-        }),
       ],
     },
     plugins: [
--- a/desktop/webpack.main.config.ts
+++ b/desktop/webpack.main.config.ts
@@ -2,9 +2,9 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
-import { ESBuildMinifyPlugin } from "esbuild-loader";
 import ForkTsCheckerWebpackPlugin from "fork-ts-checker-webpack-plugin";
 import path from "path";
+import TerserPlugin from "terser-webpack-plugin";
 import { Configuration, ResolveOptions, DefinePlugin, EnvironmentPlugin } from "webpack";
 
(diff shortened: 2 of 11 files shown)


packages/studio-base/webpack.ts
```diff
@@ -223,15 +217,6 @@ export function makeConfig(
         },
       ],
     },
-    optimization: {
```
defunctzombie, line 226: This shouldn't be here. It should be in the downstream webpack config files (i.e. web/webpack.config, renderer/webpack.config, etc).

jhurliman:
Did you test this against a working `ESBuildMinifyPlugin`?

foxymiles (approved):

defunctzombie:
@jhurliman I couldn't get ESBuildMinifyPlugin to work. The latest versions of it and esbuild did not look like they were minifying anything. I don't know if this is because of our identifier setting or not - I didn't muck around with this plugin and instead opted for terser. Terser does have an esbuild variant as well but I have no real reason to prefer esbuild over swc (swc is the default bundler for nextjs these days).

defunctzombie:
I tried the TerserPlugin with swc vs esbuild:

| Variant | build time | main.js size (mb) |
| --- | --- | --- |
| swc | 50ms | 5.27 |
| esbuild w/minifyIdentifiers:false | 40ms | 5.52 |

Up to us what tradeoff we want.

jtbandes:
> ```diff
> -  minifyIdentifiers: false, // readable error stack traces are helpful for debugging
> ```

Are we going back to unreadable stack traces in prod? Is there any equivalent setting in swc?

amacneil:
> Are we going back to unreadable stack traces in prod? Is there any equivalent setting in swc?

Since the app is open source, can we publish source maps on the public deployment and make use of them when we display stack traces?

jtbandes:
See discussion on https://github.com/foxglove/studio/pull/1880. The library John linked could help if we want to un-minify stack traces programmatically using the source map. But it is simpler to just disable some kinds of minification/mangling to make the stack traces more readable.

defunctzombie:
> Are we going back to unreadable stack traces in prod? Is there any equivalent setting in swc?

There is no equivalent setting in SWC.

The option we have is to use esbuild (still with Terser as I mention in the performance numbers above).

I didn't catch this at the time in #1880 but that is when our build went from 5mb -> 10mb and the minification stopped working. That was not an acceptable tradeoff.

@jtbandes What is the method you used to test what a stack trace looks like if we use this feature from esbuild? The referenced PR only showed a before/after of the build size but not what a stacktrace looks like before/after.

> Apparently it doesn't?

It has modes to also use swc or esbuild.

jtbandes:
> I didn't catch this at the time in https://github.com/foxglove/studio/pull/1880 but that is when our build went from 5mb -> 10mb and the minification stopped working.

I think that was the whole point of the PR, and clearly shown in the PR description 😅  But minification wasn't entirely disabled, just minification of identifiers, because those are what show up in stack traces.

> What is the method you used to test what a stack trace looks like if we use this feature from esbuild?

Here's something you can do to test:

- add a `throw` to a panel somewhere (e.g. Raw Messages)
- `yarn web:build:prod`
- `cd web/.webpack && python3 -m http.server && open http://localhost:8000`

On main (notice the RawMessages panel name is visible):
image

On roman/remove-esbuild-minify:

image

It's also the same in the app's UI.
image


In both cases the react-dom function names are unreadable, because we're using a pre-minified version of it (react-dom.production.min.js). But we do control whether our own functions show up with real or minified names. I think this matters because being able to open the error details / developer console on prod builds and see useful information is very helpful in debugging. It also helps when we can ask our users who encounter problems to open up the console and take screenshots of it.

defunctzombie:
Closing in favor of https://github.com/foxglove/studio/pull/3877.

I think full minification is probably better overall but will accept this stepping stone.

jon-chuang:
>> Are we going back to unreadable stack traces in prod? Is there any equivalent setting in swc?
>
> There is no equivalent setting in SWC.

This is, at least as of today, false. See `keepClassNames`, `keepFnNames`, `keepPrivateProps` in https://swc.rs/docs/configuration/minification.

I think we can explore using SWC-based build again if it makes the build faster.

jon-chuang:
> On my machine the web production build using swc took ~50 seconds while the default terser (non-swc) took ~63.

It does seem that ESBuild plugin is faster than SWC in most cases. Did you take note of the speed in this case?

Issue #3851, opened 2022-07-15
serializeStyles in emotion-serialize.browser.esm.js is called many times per frame


jhurliman:
**Description**

This method consistently shows up in profiling reports, and setting a breakpoint during playback (`3D (Beta)` is the only panel in the layout) shows it is called many times per frame. Is there something we should do differently to reduce the number of times this method needs to be called?

2metres:
Ok deep dive with @defunctzombie  on this one today. I looks as if @emotion-js (being used by @mui) rerenders very easily when the parent state updates. 

See example sandbox https://codesandbox.io/s/customstyles-demo-material-ui-tiyk49?file=/demo.tsx

Eventhough the props of the component are not changing the checkbox is re-rendering it's style function. 

Not sure what we can do right now. Long term answer is to ditch the use of @mui/styled and use am @emotion-js port of makeStyles and withStyles similar to @MUI v4's API which doesn't cause nearly as many re-renders.

Example using tss-react's makeStyles and withStyles functions which do not re-render when the parent state updates: 
https://codesandbox.io/s/customstyles-demo-material-ui-forked-fxlylz?file=/demo.tsx

defunctzombie:
@2metres can we consider this done given the merged PR #3868 or is there additional items you want to tackle in this specific bug report?

defunctzombie:
https://github.com/foxglove/studio/pull/4119

We need to remove instances of emotion styled from mui. We are aware of this now and have added some lint rules to warn us.

PR #3868, opened 2022-07-19
Move FoxgloveStack to use TSS-react instead of muiStyled

2metres:
**User-Facing Changes**
Better performance

**Description**
Move away from uses of muiStyled related to #3851

### DIFF
--- a/packages/studio-base/package.json
+++ b/packages/studio-base/package.json
@@ -195,6 +195,7 @@
     "ts-key-enum": "2.0.11",
     "ts-loader": "9.2.7",
     "tsconfig-paths": "3.13.0",
+    "tss-react": "3.7.1",
     "ttypescript": "1.5.13",
     "typescript": "4.6.2",
     "typescript-plugin-styled-components": "2.0.0",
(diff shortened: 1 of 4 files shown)


defunctzombie (approved):
What do you think that makeStyles is doing to avoid calling the styling function on every render?

I added a console.log to the styling function of Stack on `main` and indeed it is called every render like we saw in codesandbox. When I checkout this branch and add the same console.log the styling function is called maybe 5 times and then no more during render. How does that work?

2metres:
Not really sure because it still calls: [serializeStyles](https://github.com/garronej/tss-react/blob/11b4c5884ead610f9f31249b21518647ea69008d/src/cssAndCx.tsx#L32) just like styled but just not as often. Maybe something to do with the fact that it's not a component. But then again `withStyles` does essentially the same thing without the problem

garronej:
Hey,  
I was having a look at the modules depending on tss-react that had the most star and I found your project.  
I am very please to see such a cool project using TSS, and using it well!  

I know this is an old question but, I compute a fingerprint of the param object passed to the useStyles function:  

https://github.com/garronej/tss-react/blob/cf81afaffa55993b1c3fc50b214b871fa89cfedc/src/tools/getDependencyArrayRef.ts#L3-L21  

And it's used here:  

https://github.com/garronej/tss-react/blob/2b8cb6dee1fdf86f49fd1db635e14d58747dda62/src/makeStyles.tsx#L238  

I'm glad you noticed.

PR #3875, opened 2022-07-19
Automated performance benchmarking

jhurliman:
**User-Facing Changes**

None

**Description**

Adds a `benchmark/` folder and `yarn run benchmark:serve` target that serves up a benchmarking test for Studio in a browser.

Fixes #3726 

***TODO***

- [x] Preload the entire .mcap before beginning playback
- [x] Measure frame render times and print summary stats
- [x] Create comparable benchmark layouts for Old3D and New3D

PR #3876, opened 2022-07-19
[New3D] Performance improvements

jhurliman:
**User-Facing Changes**

- Performance improvements in `3D (Beta)` panel

**Description**

The intent is to collect miscellaneous performance improvements to New3D and see if we can use https://github.com/foxglove/studio/pull/3875 to detect an improvement.

- Don't call updateMatrix() on renderables in updatePose()
- Use uint8[4] for color attributes instead of float32[4]
- Simplify checks for skipping pointcloud min/max color calculations


packages/studio-base/src/panels/ThreeDeeRender/renderables/markers/RenderablePoints.ts
```diff
@@ -65,12 +65,12 @@ export class RenderablePoints extends RenderableMarker {
...
+    const rgbaData = attribute.array as Uint8Array;
     this._markerColorsToLinear(marker, pointsLength, (color, i) => {
-      rgbaData[4 * i + 0] = color[0];
-      rgbaData[4 * i + 1] = color[1];
-      rgbaData[4 * i + 2] = color[2];
-      rgbaData[4 * i + 3] = color[3];
+      rgbaData[4 * i + 0] = (color[0] * 255) | 0;
+      rgbaData[4 * i + 1] = (color[1] * 255) | 0;
+      rgbaData[4 * i + 2] = (color[2] * 255) | 0;
+      rgbaData[4 * i + 3] = (color[3] * 255) | 0;
```
jtbandes, line 73: I think this might need to be `* 256) | 0` (the `|0` does a floor, right?)

jtbandes: Reasoning $\lfloor (2^{\text{bits}}-1)\times (x\in[0,1))\rfloor<(2^{\text{bits}}-1)$, but we only need $\dotsm<2^{\text{bits}}$

jhurliman: I didn't realize `color[i]` was in the range `[0, 1)`. `color[0]` will never be `1`?

jtbandes: Oh yeah you're right, never mind 🤦

jtbandes: Still seems a bit weird that setting a field to `0.99999` will give you a value of `254`.

jtbandes (approved):

PR #3877, opened 2022-07-19
Enable minifySyntax and minifyWhitespace in ESBuildMinifyPlugin config

jtbandes:
**User-Facing Changes**
Reduced app bundle size.

**Description**
Alternative to #3860. Apparently when `minifyIdentifiers: false` was added in #1880, it caused the other minify options to also default to false. We need to set them to true explicitly.


defunctzombie (approved):

defunctzombie:
7.2mb main size. An improvement.

I am not convinced that we don't want to fully minify. Could even get some runtime gains from tiny names if thats still a thing for js engines.

PR #3909, opened 2022-07-21
Remove disabled message order feature

defunctzombie:
**User-Facing Changes**
None. We have already disabled this feature with no way for users to re-enable it.

**Description**
In https://github.com/foxglove/studio/pull/3311 we moved the playback message order setting behind a feature flag and turned it off by default. However, we didn't notice that we only made the setting available in dev mode. So no one could turn it back on in production.

Given that we've had no requests for this feature and our reasons for disabling it remain true, it is time to fully remove it.

This change removes the feature flag and all support code related to the header stamp playback feature.


amacneil (approved):
I think it's possible we may wish to revisit this feature in the future. However, the previous implementation of it was fundamentally broken so I support removing it entirely for now.

jtbandes (approved):

jhurliman (approved):

defunctzombie:
> I think it's possible we may wish to revisit this feature in the future. However, the previous implementation of it was fundamentally broken so I support removing it entirely for now.

If we revisit this in the future I think it will be under the concept of "publish time" rather than header.stamp (which in ros1 was the publish time but for mcap we have that as a first class concept).

PR #3912, opened 2022-07-21
Add alpha in point cloud settings for old 3D panel

siddancha:
**User-Facing Changes**
Add an option called "Alpha" in user settings for point clouds in the old 3D panel. This represents the opacity of the point cloud. It defaults to 1, but the user can set it to any value between 0 and 1.

**Description**
Transparency is very useful for point clouds, especially when the scene is cluttered with many renderables. There is currently no way to make point clouds transparent in the user settings. This PR adds transparency.

**Before**:

**After**:


jtbandes:
Thanks for the contribution. At this point, we’ve deprecated the old 3D panel and won’t be accepting new features for it since we plan to delete it soon (#4450), so I’m going to close this PR. Please try out the new 3D panel and let us know how it goes!

PR #3915, opened 2022-07-21
Fix message iterator range in CachingIterableSource

defunctzombie:
**User-Facing Changes**
None.

This fixes a bug not currently exposed by the way CachingIterableSource is used.

**Description**
The CachingIterableSource did not respect the _end_ option for messageIterator. This change fixes the iterable source to respect _end_ when reading from the cache or the underlying data source.


defunctzombie:
Split out these fixes from https://github.com/foxglove/studio/pull/3913

jtbandes (approved):

Issue #3615, opened 2022-06-15
Bulk save, revert, export, and delete layouts


jackiescanlon:
It would be awesome if we could bulk save, revert, export and delete layouts. Right now, the user must click the dot or three dots next to each layout and perform the action one at a time.

![image](https://user-images.githubusercontent.com/43035541/173833924-2a4bbc14-4546-4a55-b09e-e14606d29aa1.png)

amacneil:
+1

Especially bulk delete would be super helpful

jtbandes:
cc @2metres – we would need some designs around how the selection should work. Or better, refactor this whole thing to use some Mui off the shelf component?

foxymiles:
@jackiescanlon All of these are implemented now except for bulk export which is a bit more complex. How important is bulk export to your workflow?

jackiescanlon:
Thanks so much! It would be useful at some point but having the other bulk actions is already a huge improvement.

amacneil:
Thanks for the feedback!

Closing this ticket since the most important bulk actions have been implemented.

Bulk export would require us to zip the exported layouts client side (since browsers can't reliably trigger multiple downloads at once), and I'm not sure that is worth the additional technical complexity, especially since exporting layouts is really only used for debugging.

PR #3920, opened 2022-07-22
Update useConfirm and usePrompt Dialogs to submit if Enter key is pressed

### LINKED ISSUE #3858: Hitting enter in "Delete layer" dialog no longer works
**Description**
Pre-MUI migration, hitting enter would confirm the delete action in this dialog. Deleting layouts is already pretty painful so I rely on this to get through it slightly faster.

image

2metres:
**User-Facing Changes**
useConfirm Dialog submit's when the 'Enter' key is pressed

**Description**
Resolves #3858

### DIFF
--- a/packages/studio-base/src/hooks/useConfirm.tsx
+++ b/packages/studio-base/src/hooks/useConfirm.tsx
@@ -13,6 +13,7 @@
 
 import { Dialog, DialogContent, DialogTitle, DialogActions, Button } from "@mui/material";
 import { useCallback, useEffect, useContext, useRef } from "react";
+import { useKeyPressEvent } from "react-use";
 
 import ModalContext from "@foxglove/studio-base/context/ModalContext";
 
@@ -50,6 +51,8 @@ function ConfirmModal(props: ConfirmModalProps) {
     },
     [originalOnComplete],
   );
+
+  useKeyPressEvent("Enter", () => onComplete("ok"));
 
   // Ensure we still call onComplete(undefined) when the component unmounts, if it hasn't been
   // called already
--- a/packages/studio-base/src/hooks/useConfirm.tsx
+++ b/packages/studio-base/src/hooks/useConfirm.tsx
@@ -12,7 +12,7 @@
 //   You may not use this file except in compliance with the License.
 
 import { Dialog, DialogContent, DialogTitle, DialogActions, Button } from "@mui/material";
-import { useCallback, useEffect, useContext, useRef } from "react";
+import { useCallback, useContext, useRef } from "react";
 import { useKeyPressEvent } from "react-use";
 
 import ModalContext from "@foxglove/studio-base/context/ModalContext";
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/hooks/useConfirm.tsx
```diff
@@ -51,6 +52,8 @@ function ConfirmModal(props: ConfirmModalProps) {
     [originalOnComplete],
   );
 
+  useKeyPressEvent("Enter", () => onComplete("ok"));
+
   // Ensure we still call onComplete(undefined) when the component unmounts, if it hasn't been
```
defunctzombie, line 57: this comment doesn't match the code in the useEffect :D

defunctzombie (approved):
Are we sure we want "enter" to confirm? Even for delete operations which are destructive? Is that the common pattern in dialogs and prompts?

packages/studio-base/src/hooks/useConfirm.tsx
2metres: Don't know what that comment means. You wrote it.

2metres: I think this might be left over from fluentui's dialog. The MUI dialog explicitly requires an onClose event that is called when "Esc" is pressed

2metres: @jtbandes Removing this useEffect causes the first test to fail: https://github.com/foxglove/studio/blob/c82b0bdfc1db4c542212e8ae47dd3d96cec40134/packages/studio-base/src/hooks/usePrompt.test.tsx#L14-L27

Should I remove the test or update the test to trigger a keyboard event instead of unmounting?

jtbandes: I don't really understand the question – I think we should keep the test because it's important for the dialog to be removed when the parent component is unmounted.

Maybe we should change our prompt/confirm APIs to return a component that you have to render in your subtree, rather than dynamically adding something to the modal host. That would avoid this problem.

defunctzombie: I like the idea of giving you something to mount in the tree - it does make the api a little more cumbersome to use. Would we use a portal since we want the modal to be free of whatever place in the tree it lives?

jtbandes: Yea but I think that's already how fluent/mui dialogs work right? You just render the <Dialog> somewhere and they internally take care of where to actually mount it in the DOM?

amacneil:
> Are we sure we want "enter" to confirm? Even for delete operations which are destructive? Is that the common pattern in dialogs and prompts?

Not sure what is "common". Certainly hitting enter usually takes at least one action in a dialog (confirm or reject - whichever is highlighted). I tested macOS file delete and they highlight "cancel" by default.

We don't yet support bulk layout delete (https://github.com/foxglove/studio/issues/3615), so hitting enter to submit the dialog is important to me to delete layouts faster.

So I'm 👍 on this feature

PR #3922, opened 2022-07-22
Refactor global variables sidebar

2metres:
**User-Facing Changes**
Redesign, rename and refactor Variable Sidebar

| Before | After |
| --- | --- |
| Screen Shot 2022-08-02 at 4 28 40 pm | Screen Shot 2022-08-02 at 4 30 16 pm|


2metres:
Need help wiring up:

- naming new rows on create + renaming unlined vars
- JSON form validation and onChange handling to setGlobalVariables 
@foxymiles

foxymiles:
@2metres For renaming what UX did you have in mind? Pencil icon like settings? Rename action from the context menu similar to the layout browser?

2metres:
@foxymiles the current functionality is that the variable name is an. open text field but I think that  `••• > Rename` is a lot more consistent with other places in the UI

packages/studio-base/src/components/SettingsTreeEditor/NodeEditor.tsx
```diff
@@ -70,8 +70,7 @@ const NodeHeader = muiStyled("div", {
       },
 
       "&:hover": {
-        outline: `1px solid ${theme.palette.primary.main}`,
-        outlineOffset: -1,
```
2metres, line 74: Updating the NodeEditor hover state to be a highlighted background, same as the variable sidebar

defunctzombie:
I tried to set a variable value and got hit with this error. It took me a moment to figure out that I need to quote my value with strings. The current variable editor adds the quotes for me automatically.

Screen Shot 2022-08-02 at 10 28 06 PM

defunctzombie (approved):
Definitely looks much better. The UX around entering basic values should be a bit more friendly (i.e. not make me figure out I need to quote strings).

packages/studio-base/package.json
```diff
@@ -109,6 +109,7 @@
     "@types/uuid": "8.3.4",
     "@types/wicg-file-system-access": "2020.9.5",
     "@types/ws": "8.5.2",
+    "@uiw/react-textarea-code-editor": "2.0.3",
```
defunctzombie, line 112: Do we still need this or did you switch to the existing json editor?

foxymiles:
> Definitely looks much better. The UX around entering basic values should be a bit more friendly (i.e. not make me figure out I need to quote strings).

Dynamically adding quotes as the user types is pretty tricky. What if we preserve the behavior of the old UI and set new variables to `""`?

defunctzombie:
@foxymiles whatever seems to make sense for simplifying the entry of basic values. Right now its a bit unfriendly in that most folks just want to enter a single value rather than learn nuances of json.

amacneil:
1/ the linked variables are about 10 pixels smaller than regular variables. The top two look better than the bottom one imo.
image

2/ `Delete variable` might be simpler as a trash icon rather than a dropdown menu with one option? Although it's not terrible, not sure if other thoughts went into this? Side note - why so much padding in our context menus?
image

3/ The temporary purple flash when a variable changes is a nice touch, but it should be shorter and more of a fade effect imo. I can't think where I have seen this recently, but I would expect it to highlight and quickly fade out (probably less than 1s total):

https://user-images.githubusercontent.com/637671/183128462-fa5eed29-ea32-464d-bc4e-c822b0b4d2dd.mov

4/ The purple flash works when setting a variable from 3d panel selection, but doesn't work when setting a variable from Variable Slider panel

https://user-images.githubusercontent.com/637671/183130198-d2c0610a-9d0a-4145-a446-0b36fbf4d1a6.mov

5/ This error is kinda silly. Imo it would be better to only treat variables starting with `{` as json, and everything else as string (although that might require a migration to unquote string variables in existing layouts). A potentially easier solution would be to automatically insert quotes if I clear a variable and start typing any non-numeric character.
image

foxymiles:
@amacneil I've tried various approaches to dynamically adding quotes around the input as the user types but it's not easy to make this work as expected in general cases. I'm sure it's doable if we're willing to spend more time on this but this current implementation mirrors the existing implementation on main.

amacneil:
Makes sense. I'm fine with dropping (5) since the behavior is better than what we have on main today (where we seem to just delete unquoted strings).

2metres:
Not sure why but the flash on change works here the same as before. The variable slider does not cause a flash opening an issue to resolve seperately. #4090

PR #3924, opened 2022-07-22
Fix plot message path scrolling on linux chrome

### LINKED ISSUE #3743: Cant access begining of path if very long path exists in plot
**Description**
Small UI glitch. Can go further left than this in this long topic path  (diagnostics one). Im trying to go at the beginning to add /old/ but I cant.
image

Path was `/diagnostics.status[:]{hardware_id=="none"}{name=="m01/conveyor_speed_node: Overall health"}.values[:]{key=="Elaspsed (s): conveyor_speed_estimation"}.value`

- Version:1.13.2
- OS:ubuntu
- ROS distro (if applicable):

**Steps To Reproduce**

**Expected Behavior**

**Actual Behavior**

foxymiles:
**User-Facing Changes**
This fixes an issue with long message path inputs in plots on Linux Chrome.

**Description**
This is a somewhat hacky JS fix that synchronizes the plot legend content scroll position with the cursor position on nested message path inputs. At least on Linux Chrome the nested input's scroll container and the parent scroll container don't quite sync up. I tried a variety of CSS solutions but couldn't find any that didn't require functional changes to the plot legend.

This also adds a CSS fix for the sticky positioning of the `Add line` button.

Fixes #3743

### DIFF
--- a/packages/studio-base/src/panels/Plot/PlotLegend.tsx
+++ b/packages/studio-base/src/panels/Plot/PlotLegend.tsx
@@ -121,14 +121,12 @@ const useStyles = makeStyles((theme: Theme) => ({
     backgroundColor: "transparent !important",
     padding: "4px !important",
   },
-  footer: ({ legendDisplay }: StyleProps) => ({
+  footer: ({ showPlotValuesInLegend = false }: StyleProps) => ({
     padding: theme.spacing(0.5),
-    gridColumn: "span 4",
-    ...(legendDisplay !== "floating" && {
-      position: "sticky",
-      right: 0,
-      left: 0,
-    }),
+    gridColumn: showPlotValuesInLegend ? "span 4" : "span 3",
+    position: "sticky",
+    right: 0,
+    left: 0,
   }),
   floatingWrapper: {
     overflow: "hidden",
--- a/packages/studio-base/src/panels/Plot/PlotLegend.tsx
+++ b/packages/studio-base/src/panels/Plot/PlotLegend.tsx
@@ -24,7 +24,15 @@ import { Button, IconButton, Theme, alpha, MenuItem, Menu } from "@mui/material"
 import { makeStyles } from "@mui/styles";
 import cx from "classnames";
 import { last } from "lodash";
-import { ComponentProps, useCallback, useEffect, useMemo, useRef, useState } from "react";
+import {
+  ComponentProps,
+  useCallback,
+  useEffect,
+  useLayoutEffect,
+  useMemo,
+  useRef,
+  useState,
+} from "react";
 import { useLatest } from "react-use";
 import { useDebouncedCallback } from "use-debounce";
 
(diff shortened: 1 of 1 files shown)


defunctzombie:
What does _nested input_ mean here? Is this an issue that impacts all <input> fields in chrome or is this something else?

foxymiles:
> What does _nested input_ mean here? Is this an issue that impacts all fields in chrome or is this something else?

I think the issue here is that we have three nested scroll containers. One for a wrapper, one for the content, and then the input itself is a scroll container. We also have a `position: sticky` button on the left. All this seems to defeat Chrome's logic for synchronizing scroll positions automatically, at least on Linux.

defunctzombie:
> I think the issue here is that we have three nested scroll containers.

Can we drop any of them? Maybe we only need the input to scroll? What else do we need to scroll?

foxymiles:
> > I think the issue here is that we have three nested scroll containers.
> 
> Can we drop any of them? Maybe we only need the input to scroll? What else do we need to scroll?

I tried a whole bunch of these kinds of approaches but couldn't come up with any that didn't change how this component works. If we're willing to make some slight functional changes we can probably find a pure CSS solution. More specifically I couldn't find a way in CSS only to preserve the ability to scroll horizontally on all inputs at once like in the attached recording.

https://user-images.githubusercontent.com/93935560/180481950-4258a166-642b-429d-842f-160c652c945f.mov

foxymiles:
@defunctzombie I also have a CSS only solution in https://github.com/foxglove/studio/tree/miles/legend-fix-noscroll but we lose the group horizontal scrolling and the toggle button on the left is no longer sticky.

defunctzombie:
Do we think the group horizontal scrolling is that important? I find the scrolling behavior in the video a bit strange - like why does the `Add line` button scroll off the screen? Also the way the context menu and remove buttons appear over the text is another one of those ongoing annoyances I have with elements that appear on hover without a specific place allocated for them - but thats a battle for another day I suppose.

foxymiles:
> Do we think the group horizontal scrolling is that important? I find the scrolling behavior in the video a bit strange - like why does the `Add line` button scroll off the screen? Also the way the context menu and remove buttons appear over the text is another one of those ongoing annoyances I have with elements that appear on hover without a specific place allocated for them - but thats a battle for another day I suppose.

Yeah I agree that the UX here seems weird. If we're willing to make some functional changes I think we could improve it.

foxymiles:
> https://github.com/foxglove/studio/tree/miles/legend-fix-noscroll

The approach I took in https://github.com/foxglove/studio/tree/miles/legend-fix-noscroll seems cleaner to me but not sure how much people are depending on the existing scroll behavior.

2metres:
I think having both rows scroll together feels more intuitve.

defunctzombie (approved):
Andrew likes having both rows scroll so lets go with this fix. I dream of the days of panel settings making all this mess go away.

PR #3929, opened 2022-07-23
[New3D] Fix COLLADA asset loading with Lambertian materials

### LINKED ISSUE #3908: .dae files no longer loading in 3D (Beta)
.dae files that loaded in the previous 3D viewer do not load in the Beta version and throw the following error: 'Error loading mesh from "[filepath here]": Cannot read properties of undefined (reading 'x').' A failing .dae file is attached.
[logo.zip](https://github.com/foxglove/studio/files/9160036/logo.zip)

jhurliman:
**User-Facing Changes**

- Fixes a parsing error in `3D (Beta)` for some COLLADA meshes

**Description**

THREE.js doesn't have any defensive checks in Material.copy() so we need to ensure all the right fields exist when copying properties from a phong or lambert material into a standard material. Fixes #3908

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/ModelCache.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/ModelCache.ts
@@ -266,6 +266,13 @@ function toStandard(
   const standard = new THREE.MeshStandardMaterial({ name: material.name });
   const shininess = (material as Partial<THREE.MeshPhongMaterial>).shininess ?? 0; // [0-100]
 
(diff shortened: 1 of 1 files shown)


defunctzombie:
@jhurliman Is there a way to detect or catch the error and show it on the topic? Or was that already happening. I couldn't tell from the referenced bug report whether it was a panel crash or a 🔴

jhurliman:
@defunctzombie there's no topic error to display here, it was a valid COLLADA file that we didn't handle properly.

jtbandes (approved):

Issue #3863, opened 2022-07-18
3D beta frame error is stuck


defunctzombie:
I was looking at a data source with a base_link frame. I then opened a different data source that did not have a base_link and turned on a topic to show. The 3d panel automatically changed my frame to the topic's frame and is showing an error.

<img width="398" alt="Screen Shot 2022-07-18 at 9 21 45 AM" src="https://user-images.githubusercontent.com/84792/179566198-2e4aa915-598c-4782-86bc-bb38559eab9c.png">

I've tried re-selecing the new frame in the drop down but the error persists.

**Desired Behavior**

It is un-intuitive to have to "re-select" a frame manually if I see it is already selected. I'd rather have seen a topic error telling me that there is no transform from `base_link -> vertical` and realized I need to change my frame than have whatever auto-magic switch the frame for me when I toggle topics.

jhurliman:
> I'd rather have seen a topic error telling me that there is no transform from base_link -> vertical

This is not the problem that is encountered, the problem is you manually selected a frame named `base_link` at some point and then you loaded a data source that does not have a frame with that name. At that point we pretty much have two options:

1. Show a blank 3D panel, set the Frame dropdown to no selection (blank), and show a sidebar error saying `Frame "base_link" not found`
2. Use our default auto-frame selection logic to choose a frame in the current data source, but warn the user with a sidebar error that the frame they manually selected is not available and another frame is being used for rendering instead (the current behavior).

Which option should we go with?

amacneil:
What about a variation on (1): Show a blank 3D panel, set the Frame dropdown value to `base_link`, show a sidebar error saying `Frame "base_link" not found`.

That way it's clear that the currently selected frame doesn't exist. I can hit the dropdown and choose a different frame (to be clear, assuming MUI allows this, I would expect the field to show `base_link`, but `base_link` would not be in the list of dropdown options when I go to select a new frame).

jhurliman:
@amacneil that's not allowed with the settings sidebar API (I assume due to MUI internals) today. You get a console warning and a blank selection. The closest we can do is add "base_link" to the dropdown, but selecting it continues to show the sidebar error. And as soon as you select a different frame, "base_link" would disappear from the options. It's a bit weird (IMO) UX-wise but doable.

amacneil:
Can we confirm if that's a mui limitation or just something we enforced? 

[...]

PR #3930, opened 2022-07-23
[New3D] Don't auto-select a different frame if the user-configured render frame doesn't exist

jhurliman:
**User-Facing Changes**

- `3D (Beta)` panel will no longer auto-select a different frame if the user-selected frame does not exist

**Description**

Fixes https://github.com/foxglove/studio/issues/3863

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
@@ -495,8 +495,8 @@ export class Renderer extends EventEmitter<RendererEvents> {
     }
 
     // Top priority is the followFrameId
-    if (this.followFrameId != undefined && this.transformTree.hasFrame(this.followFrameId)) {
-      return this.followFrameId;
+    if (this.followFrameId != undefined) {
+      return this.transformTree.hasFrame(this.followFrameId) ? this.followFrameId : undefined;
     }
 
     // Prefer frames from [REP-105](https://www.ros.org/reps/rep-0105.html)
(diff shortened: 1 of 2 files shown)


amacneil:
Seems to work.

There is a strange behavior that I guess the panel defaults to "auto" frame or something? Because if I start with a fresh new3D panel, load a bag file with `base_link` frame, then load a different bag file with no `base_link`, it will happily switch over to `world` frame automatically. But if I specifically change the frame then it will stay on what I selected.

The behavior seems reasonable, just want to check that it is intended.

defunctzombie:
> There is a strange behavior that I guess the panel defaults to "auto" frame or something? Because if I start with a fresh new3D panel, load a bag file with `base_link` frame, then load a different bag file with no `base_link`, it will happily switch over to `world` frame automatically. But if I specifically change the frame then it will stay on what I selected.

I find this behavior still somewhat odd. I think the panel starting out in _auto_ mode is fine but once a frame is selected (and I see it in the `Display frame` it seems sensible that frame remains selected. I wonder if we should have an _auto_ option be explicit in the dropdown? You basically let the panel decide?

With no auto option the behavior that seems most "stable" to me is when you add a new 3d panel it is in frame discovery mode. Once you open some data the panel selects a suitable frame and that is what you see in the frame selection drop down. When this happens the panel does not go back to auto-discovery mode.

My reasoning is as follows: let's say I open a file for my robot. I want to view things in base_link and the 3d panel auto-selected that. I see `base_link` as the display frame and take no action to re-select it. I open another file and for some reason this file doesn't have the right transform information. Rather than being notified that some items can't render because we don't have `base_link`, I've been thrown into another frame - one that I don't view data in or want to use day-to-day.

amacneil:
I actually found the current behavior pretty helpful. It means when you open a new file you have a good chance of the 3d panel working.

If we want to keep the current behavior but make it more explicit, we could add a menu entry like `Auto (base_link)` so we can show both the fact that you're in auto mode, and the current frame you're looking at?

defunctzombie:
> I actually found the current behavior pretty helpful. It means when you open a new file you have a good chance of the 3d panel working.

When you open a file and it selects a strange frame (like camera front) and you are now inverted in the scene - "working" isn't quite the feeling. And I've definitely seen it happen where the panel puts you into a strange frame and view. Maybe we go with this for now and if I run across such examples I can remember to file issues for evidence over anecdotes.

jhurliman:
> When you open a file and it selects a strange frame (like camera front)

Please file this as a bug with repro steps if possible. The auto-frame selection should only select a CAMERA_FRONT frame if there are no well-known frames and CAMERA_FRONT is the root frame of the largest tree of coordinate frames.

defunctzombie (approved):

PR #3934, opened 2022-07-24
Improve full preload handling for iterable player

defunctzombie:
**User-Facing Changes**
Better handling of preloading topics for users on experimental players

**Description**
The BlockLoader is responsible for loading subscriptions with the "full" subscriber option. These are topics which a panel has indicated it wants all the messages across the entire data source range. When there is not enough memory to hold the entire dataset, the loader attempts to prioritize showing the data at and after the current time.

The existing behavior of BlockLoader had some drawbacks:

- When switching from idle -> play (and vice versa) the loader would stop and restart.
- When loading blocks the loader would not consider the position of the playhead for which blocks to evict vs keep which resulted in evicting blocks relevant to the current playhead.

This change addresses these issues by making the loader a "background" task of the iterable player. Rather than starting/stopping the loader during state changes, the iterable player updates the loader with the current time and the current full subscriber topic set. The loader uses these inputs to manage the full preload. When these inputs allow, the loader keeps loading rather than starting and stopping. When these inputs change in a way that the loader needs to start loading another region (i.e. seeking to an un-loaded region), the loader stops and restarts itself to always ensure it is loading the region with the playhead.

This change adds additional tests to verify that the loader loads the expected regions under playback and seeking.

### DIFF
--- a/packages/den/async/Condvar.test.ts
+++ b/packages/den/async/Condvar.test.ts
@@ -1,0 +1,42 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { Condvar } from "./Condvar";
+
+describe("Condvar", () => {
+  it("should notifyAll waiting", async () => {
+    const condvar = new Condvar();
+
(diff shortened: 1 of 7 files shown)


foxymiles (approved):
Looks good. I'm surprised there isn't an existing JS condition variable library but I looked around a bit and didn't find anything.

Issue #3786, opened 2022-07-07
Support simple GeoJSON styling in the map view


efoley:
It would be nice if GeoJSON rendering in the map view could support simple styling (at least coloring.)

This is easy to implement, but I don't think there is any standardized way of including style in the GeoJSON itself. One way to do this is to pull out `feature.properties.color` (see example diff [here](https://github.com/foxglove/studio/compare/main...efoley:studio:efoley/map_geojson_style)), as this follows the example in the leaflet docs.

Coloring topics (https://github.com/foxglove/studio/issues/3753) would be a partial workaround for this, but it would still be nicer to style from the GeoJSON as that supports more complex cases.

PR #3935, opened 2022-07-25
Allow overrides for map panel topic colors

### LINKED ISSUE #3753: Allow control over color of topics displayed on Map panel
We find the Map panel very useful for displaying multiple robots. The change to show different topics in different colors was a big improvement, but having more control over the color would be very helpful in our case. Currently, the color seems to be chosen based on the order the topic appears, which can vary and takes some additional time each run to figure out which point corresponds with which vehicle. Furthermore, The points we care the most about are consistently given a pale white color that blends in perfectly with the roads we test on so that we can't see them at all.

foxymiles:
**User-Facing Changes**
This allows users to override the topic colors in the map panel on a per-topic basis.

**Description**
This expands the settings UI for the map panel to allow users to choose between automatic and custom coloring for each topic and to select a color for each custom topic.

Screen Shot 2022-07-25 at 8 42 37 AM

Fixes #3753 
Partially addresses #3786

### DIFF
--- a/packages/studio-base/src/panels/Map/config.ts
+++ b/packages/studio-base/src/panels/Map/config.ts
@@ -14,6 +14,7 @@ export type Config = {
   layer: string;
   zoomLevel?: number;
   followTopic: string;
+  topicColors: Record<string, string>;
 };
 
 export function validateCustomUrl(url: string): Error | undefined {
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/panels/Map/index.stories.tsx
```diff
@@ -120,26 +120,17 @@ SinglePoint.parameters = {
...
 SinglePointWithSettings.parameters = {
-  chromatic: {
-    delay: 1000,
-  },
+  ...SinglePoint.parameters,
+  includeSettings: true,
+};
+
+export const SinglePointWithSettingsOverride = (): JSX.Element => {
+  return <MapPanel overrideConfig={{ layer: "custom", topicColors: { "/gps": "pink" } }} />;
```
defunctzombie, line 129: Can "pink" ever be selected by the user?

packages/studio-base/src/panels/Map/config.ts
```diff
@@ -14,6 +14,7 @@ export type Config = {
   layer: string;
   zoomLevel?: number;
   followTopic: string;
+  topicColors: Record<string, string>;
```
defunctzombie, line 17: worth making as `?` here since it won't exist in previous layouts? Or does `initialConfig` already do a full partial on the config it gives the panel?

defunctzombie (approved):

packages/studio-base/src/panels/Map/config.ts
foxymiles: I'm fixing this up in the panel itself so it's never undefined after initialization.

packages/studio-base/src/panels/Map/index.stories.tsx
foxymiles: No but it gets converted to hex for display. But I'll change this to a hex value just to be as consistent with user input as possible.

PR #3937, opened 2022-07-25
Make panel titles editable

### LINKED ISSUE #3806: Add title to indicator panel
We use a lot of these panel to indicate status of systems but there is no way to distinguish them despite changing for example the "False" label to "System1 - False" which is not ideal.

Would be nice to be able to edit the title of the panel like you can with the plot panel.

foxymiles:
**User-Facing Changes**
This implements editable panel titles for all standard panels.

**Description**
This makes panel titles editable. This implementation adds a pencil icon to put the titlebar in editing mode. If we feel this adds too much clutter to the titlebar then another option is a new item in the context menu that toggles title editing mode.

Fixes #3806

### DIFF
--- a/packages/studio-base/src/components/PanelToolbar/index.tsx
+++ b/packages/studio-base/src/components/PanelToolbar/index.tsx
@@ -14,7 +14,7 @@
 import CheckIcon from "@mui/icons-material/Check";
 import EditIcon from "@mui/icons-material/Edit";
 import FullscreenExitIcon from "@mui/icons-material/FullscreenExit";
-import { IconButton, InputBase, styled as muiStyled, Typography } from "@mui/material";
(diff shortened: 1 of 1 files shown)


2metres:
I feel like there are a couple problems I see with this

The placement of the pencil icon is very disconnected to the the actual panel title on wider panels, essentially have to click a button on the right of your layout then look to the left to edit the text.

For panels where the toolbar has children passed in (about half of the current set) the edit button doesn't do anything. So we can do 1 of 2 things, neither of which I think is great
  - Hide the toolbar children when there is a custom title
  - Remove places where toolbars have children in favor of an internal "toolbar" provided by the panel

jtbandes:
What's the plan for this feature? Is there going to be a replacement for this PR?

PR #3939, opened 2022-07-25
Add keyboard shortcut to close the sidebar

### LINKED ISSUE #3873: Make the sidebar easier/quicker to close
**Description**
The only way to close the sidebar is to click on the sidebar tab you're currently on. This is especially frustrating when it is accidentally opened, say, by fat-fingering the kebab icon next to the panel settings icon in panel headers. Having a quicker, more natural way to close the sidebar would streamline my usage of Studio.

- Version: Latest
- OS: Any
- Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native): Any
- ROS distro (if applicable): Any

**Steps To Reproduce**
See description.

**Current Experience**
See description.

**Desired Experience**
Proposed ideas, non-exclusive:

- Create a keyboard shortcut, e.g. 'esc' or 'ctrl-b', to close the sidebar. This could be the default keybinding to close any sort of menu within the application.
- When clicking on the panel settings icon in the panel header, the settings menu opens up. Clicking on the same settings button again should close the sidebar. With the sidebar open, clicking on a different panel's settings icon would open up the settings for that panel.

foxymiles:
**User-Facing Changes**
This adds a global keyboard shortcut for closing the sidebar, mapped to `ctrl-b`.

**Description**
This just adds a global key handler to close the sidebar. It's not a toggle and doesn't reopen the sidebar its previous state but just mirrors the process of clicking on sidebar button. We could expand on this to maintain or restore the previous sidebar state when the shortcut is used when the sidebar is closed but that would require more extensive changes.

Fixes #3873

### DIFF
--- a/packages/studio-base/src/Workspace.tsx
+++ b/packages/studio-base/src/Workspace.tsx
@@ -36,6 +36,7 @@ import variablesHelpContent from "@foxglove/studio-base/components/GlobalVariabl
 import HelpSidebar, {
   MESSAGE_PATH_SYNTAX_HELP_INFO,
 } from "@foxglove/studio-base/components/HelpSidebar";
+import KeyListener from "@foxglove/studio-base/components/KeyListener";
 import LayoutBrowser from "@foxglove/studio-base/components/LayoutBrowser";
 import {
   MessagePipelineContext,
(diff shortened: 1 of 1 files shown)


amacneil:
If we're going for vs code consistency, can we map it to `Cmd + B` on macOS and `Ctrl + B` on Windows/Linux?

foxymiles:
> If we're going for vs code consistency, can we map it to `Cmd + B` on macOS and `Ctrl + B` on Windows/Linux?

Looks like I'll have to add some OS sniffing to make the distinction but otherwise easy to do.

foxymiles:
@amacneil I've updated this to use `Cmd+B` on Macs.

2metres (approved):

amacneil:
Awesome. Can we make the same shortcut show the sidebar (rather than just hide)?

foxymiles:
> Awesome. Can we make the same shortcut show the sidebar (rather than just hide)?

I mentioned that at the end of the description. We'd need a more extensive PR to preserve and restore state to make that work but it's doable.

PR #3941, opened 2022-07-26
Read and write global variables from extension panels

jhurliman:
**User-Facing Changes**

- [Extension API] Add global variable support via `RenderState#variables`, `PanelExtensionContext#setVariable(name, value)`

**Description**

This adds global variable support to the extension panel API with a new `setVariable()` method and a `variables` map on RenderState. This pattern mimics the parameters API, although the types are slightly different.


foxymiles (approved):
Code looks good. It would be nice to have some tests or stories around this.

defunctzombie (approved):
👍 seems legit. Agree that writing a test or two would be good before merging. I think we have some tests for PanelExtensionAdapter that you can reference for how we've tested other items here.

PR #3945, opened 2022-07-26
fix(New3D): Fix topic subscriptions logic

dudv:
**Description**

After the latest rebase, our custom `SceneExtension` stopped working. On a closer look, it turned out the subscription messages were not going through even though `SceneExtension` subscribed through `addTopicSubscription`. The culprit was that `ThreeDeeRender` requires _both_ `topicHandlers` and `datatypeHandlers` to be present while we only had a single `topicHandler`. This PR fixes the logic so that a single handler is sufficient.

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
+++ b/packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
@@ -569,9 +569,9 @@ export function ThreeDeeRender({ context }: { context: PanelExtensionContext }):
       } else if (
         // prettier-ignore
         (topicHandlers.get(topic.name)?.length ?? 0) +
-        (datatypeHandlers.get(topic.datatype)?.length ?? 0) > 1
+        (datatypeHandlers.get(topic.datatype)?.length ?? 0) > 0
       ) {
-        // Subscribe if there are multiple handlers registered for this topic
+        // Subscribe if there are handlers registered for this topic
         subscriptions.add(topic.name);
       }
     }


jhurliman (changes requested):
This won’t work exactly as written, because it would cause every potentially renderable topic to be subscribed to even if topic visibility is toggled off for everything. I think it can be fixed with a little logic rework though.

dudv:
> it would cause every potentially renderable topic to be subscribed
It would cause only the topics which have active handlers to be subscribed. AFAIK the topic visibility wasn't taken into account in the previous variation of the logic as well.

jhurliman:
@dudv here's a walkthrough of the logic:

```typescript
    for (const topic of topics) {
      if (
        FRAME_TRANSFORM_DATATYPES.has(topic.datatype) ||
        TF_DATATYPES.has(topic.datatype) ||
        TRANSFORM_STAMPED_DATATYPES.has(topic.datatype)
      ) {
        // Subscribe to all transform topics
        subscriptions.add(topic.name);
```

All transform topics are unconditionally subscribed to so we can ingest the full transform hierarchy for the scene.

```typescript
      } else if (config.topics[topic.name]?.visible === true) {
        // Subscribe if the topic is visible
        subscriptions.add(topic.name);
```

Every topic marked as `visible: true` in the layout is subscribed to. Topics are marked visible in the layout when a SceneExtension has created a settings sidebar node for it and the user has enabled the visibility toggle. Example: PointCloudsAndLaserScans creates a settings node for all topics that have a point cloud or laserscan datatype, the user marks one of these topics as visible, then the above logic creates a subscription for it. If the topic is not marked as visible, we don't subscribe to it. Otherwise, we would be subscribing to every topic the 3D scene could possibly render, which may overload the network connection if there are many high-bandwidth topics.

```typescript
      } else if (
        // prettier-ignore
        (topicHandlers.get(topic.name)?.length ?? 0) +
        (datatypeHandlers.get(topic.datatype)?.length ?? 0) > 1
      ) {
        // Subscribe if there are multiple handlers registered for this topic
        subscriptions.add(topic.name);
      }
    }
```

This is a special exception, and it should probably be done in a cleaner way. Some SceneExtensions share data between each other, such as `Cameras` and `Images` where `Cameras` creates camera models for CameraInfo topics and `Images` uses those camera models to render images in 3D space. Because it's valid to only have visibility enabled for an image and not the associated camera info topic, this block of code creates a subscription to all CameraInfo topics regardless of visibility toggles.

A cleaner way of doing this would be a flag or separate method for scene extensions to create "always-on" subscriptions, where a topic is subscribed to even if visibility is toggled off for a topic or there is no settings sidebar / layout entry for the topic. Would this satisfy your use case? I'm assuming that's what you're trying to achieve and why the "subscribe if topic is marked visible" logic is not meeting your needs.

dudv:
Hmm, I guess my description of the problem was not clear. The issue I was trying to fix was that none of the topics from my custom SceneExtension got subscribed to after a recent rebase. Here's the setup:

1. `NvCustomExtension.ts constructor`:
```
renderer.addTopicSubscription("av/SomeTopic", this.handleSubscriptionMessages)
```
Note the `av/SomeTopic` is not renderable i.e. it's not a marker or something, just some raw message coming from the back-end. So it's not in the list of renderable topics.

2. `ThreeDeeRenderer.ts`
```
        (topicHandlers.get(topic.name)?.length ?? 0) +
        (datatypeHandlers.get(topic.datatype)?.length ?? 0) > 1
      ) {
        // Subscribe if there are multiple handlers registered for this topic
        subscriptions.add(topic.name);
      }
    }

    const newTopics = Array.from(subscriptions.keys()).sort();
```
Here,  `subscriptions` do not include `av/SomeTopic` because it's not visible it has only one `topicHandler` (created above in `NvCustomExtension.ts`).

So this PR was an attempt to fix that. The logic you've described above seems correct for _renderable_ topics but not for subscriptions to some input data. To reiterate, our `NvCustomExtension` uses some arbitrary messages as input and produces visualization. It actually worked just fine before `1.18.0` so I thought such a use-case is supported but now it seems the logic is designed for renderable topics and not just any random topics.

jhurliman:
@dudv I will look at adding the ability to subscribe to a topic or datatype unconditionally. That should clean up the existing code a bit and enable your use case.

PR #3946, opened 2022-07-26
Select entire layout name on rename

foxymiles:
**User-Facing Changes**
None

**Description**
This fixes the logic to select the entire layout name for convenience when the user chooses `rename` from the layout context menu. Setting the `disableAutoFocus` and `disableRestoreFocus` on the menu fixes the focus interference of the menu and input.


defunctzombie:
Amazing - I've been annoyed by this recently 👏

2metres (approved):

PR #3950, opened 2022-07-26
[New3D] Fix topic links from selected object dialog

jhurliman:
**User-Facing Changes**

- Fixed `3D (Beta)` selection dialog topic link to use the correct topic name

**Description**

Only put topic names in the topic links, don't show topic links for renderables that do not have a topic name


defunctzombie (approved):
What type of renderable doesn't have a topic?

foxymiles (approved):

jhurliman:
@defunctzombie URDFs are one example

PR #3951, opened 2022-07-26
[New3D] Set "selected_id" global variable when selecting a marker

jhurliman:
**User-Facing Changes**

- `3D (Beta)` panel sets a global variable named `selected_id` when selecting a marker in the 3D scene

**Description**

This also wires the map of global variables into `Renderer` instances, which is not currently used but could be used in the future to make object selection propagate across all 3d panels or expose variables to 3d panel scene extensions.


foxymiles (approved):

amacneil:
Feedback:
- No need to implement support for changing the selected variable ID (wait and see if anyone asks for it)
- Set `$selected_id` to `null` after I deselect an item (so that it doesn't cause message path syntax errors in my layout)
- Update Node Playground examples from `id` to `selected_id`

PR #3955, opened 2022-07-27
Support batch deleting layouts in the layout browser

foxymiles:
**User-Facing Changes**
This makes it possible to batch delete layouts in the layout browser.

**Description**
This implementation allows the user to select multiple layouts by holding the `Ctrl/Cmd` key while selecting layouts. When more than one layout is selected a delete icon appears in the layout browser header that will delete all selected layouts when clicked. 

The implementation here deletes one layout at a time. We could expand this to support transactional deletes of multiple layouts but that would require much more extensive changes.

Another enhancement would be supporting selecting ranges of layouts via `shift` clicking, similar to the Mac finder or Windows explorer.

Partially addresses #3615

### DIFF
--- a/packages/studio-base/src/components/LayoutBrowser/index.tsx
+++ b/packages/studio-base/src/components/LayoutBrowser/index.tsx
@@ -215,16 +215,7 @@ export default function LayoutBrowser({
         }
         void analytics.logEvent(AppEvent.LAYOUT_SELECT, { permission: item.permission });
       }
-      const selectedIsShared =
-        layouts.value?.shared.some((layout) => layout.id === item.id) === true;
-      const currentIsShared =
-        currentLayoutId != undefined &&
-        layouts.value?.shared.some((layout) => layout.id === currentLayoutId) === true;
-      if (
-        (event?.ctrlKey === true || event?.metaKey === true) &&
-        !selectedIsShared &&
-        !currentIsShared
-      ) {
+      if (event?.ctrlKey === true || event?.metaKey === true) {
         if (item.id !== currentLayoutId) {
           dispatch({ type: "toggle-selected", id: item.id });
         }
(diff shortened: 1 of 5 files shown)


packages/studio-base/src/components/LayoutBrowser/index.tsx
```diff
@@ -113,7 +152,7 @@ export default function LayoutBrowser({
 
   // Start loading on first mount
   useEffect(() => {
-    void reloadLayouts();
+    reloadLayouts().catch((err) => log.error(err));
```
defunctzombie, line 155: Do we need to surface this to the user as offline mode or anything?

packages/studio-base/src/components/LayoutBrowser/index.tsx
```diff
@@ -161,22 +200,47 @@ export default function LayoutBrowser({
...
+      { selectedViaClick = false, event }: { selectedViaClick?: boolean; event?: MouseEvent } = {},
+    ) => {
       if (selectedViaClick) {
         if (!(await promptForUnsavedChanges())) {
           return;
         }
         void analytics.logEvent(AppEvent.LAYOUT_SELECT, { permission: item.permission });
       }
-      setSelectedLayoutId(item.id);
+      const selectedIsShared =
```
defunctzombie, line 218: Comment that we decided to now allow bulk delete of shared layouts

defunctzombie (approved):

packages/studio-base/src/components/LayoutBrowser/index.tsx
foxymiles: Yeah probably. I just wanted to at least log it instead of swallowing it as we do on main. Should I try to address that in this PR?

packages/studio-base/src/components/LayoutBrowser/index.tsx
foxymiles: Ok. I'll remove that restriction.

PR #3957, opened 2022-07-27
Add remote mcap file support for experimental mcap player.

defunctzombie:
**User-Facing Changes**
When the experimental mcap player flag is on, remote mcap files are handled via the new player.

**Description**

Fixes: #3949


defunctzombie:
I tested this by setting up a local file server.

foxymiles (approved):

PR #3959, opened 2022-07-27
Refactor layout browser stories

foxymiles:
**User-Facing Changes**
None

**Description**
This is an internal refactor of the stories for the `LayoutBrowser` component to use the `play` and `fireEvent` APIs of storybook and react-testing-library. This is just setting the stage for more extensive stories covering upcoming multi-select actions.


defunctzombie (approved):
exciting! I wonder how many uses of useReadySignal we can 🔪

PR #3964, opened 2022-07-27
Fix setRenderDone call for new3d

defunctzombie:
**User-Facing Changes**
None

**Description**
Calling a react `setState` function with a function argument will call that function argument immediately. The behavior we want here is to set `renderDone` to a function we can later call to indicate rendering is done.


amacneil:
This is why I don't write react

foxymiles (approved):

jtbandes:
I'm pretty sure when we first discovered this issue in a different panel, we said "oh yeah we should probably fix this in the New3D panel too". I guess we never did 😄

PR #3970, opened 2022-07-28
Avoid constantly subscribing when typing in raw message input

defunctzombie:
**User-Facing Changes**
TBD

**Description**
TBD

### DIFF
--- a/packages/studio-base/src/components/MessagePathSyntax/MessagePathInput.tsx
+++ b/packages/studio-base/src/components/MessagePathSyntax/MessagePathInput.tsx
@@ -161,8 +161,10 @@ function getExamplePrimitive(primitiveType: RosPrimitive) {
 type MessagePathInputBaseProps = {
   supportsMathModifiers?: boolean;
   path: string; // A path of the form `/topic.some_field[:]{id==42}.x`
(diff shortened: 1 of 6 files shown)


packages/studio-base/src/components/MessagePathSyntax/MessagePathInput.tsx
```diff
@@ -161,8 +161,10 @@ function getExamplePrimitive(primitiveType: RosPrimitive) {
 type MessagePathInputBaseProps = {
   supportsMathModifiers?: boolean;
   path: string; // A path of the form `/topic.some_field[:]{id==42}.x`
+  // fixme - wtf is this and why is making anonymous functions bad?
```
jtbandes, line 164: it's like 

```ts
items.map((foo, i) => <Input onChange={() => doSomething(i)} />)
```
vs
```ts
const onChange = (i) => doSomething(i);
items.map((foo) => <Input index={i} onChange={onChange} />)
```

to avoid creating new functions and thrashing the onChange prop. That is my guess.

defunctzombie:
I'll get there one day. Its a reminder to me for now.

jtbandes:
cc @achim-k – this is similar to the publish panel advertising when you type in the topic field.

PR #3971, opened 2022-07-28
Create quaternion2rpy

ljburtz:
**User-Facing Changes**
New utility code snippet to make plotting human readable IMU messages

**Description**
Utility function to convert from ROS Quaternion message type to roll, pitch, yaw (in degrees)
this is useful to make human readable plots of a robot's IMU or Pose (any message with a 'quaternion' or 'orientation' field)
inspired by plotjuggler feature that does this behind the scenes automatically (and better)
https://github.com/facontidavide/PlotJuggler/blob/f5f4bdc2176ebec2505c55f21bdf711f86f79cdc/plotjuggler_app/resources/default.snippets.xml

**Caveats**
1. manual user intervention to specify the topic
2. need one node for each topic to be plotted
3. would be great if the topic name could be a ["variable" ](https://foxglove.dev/docs/studio/panels/node-playground#using-global-variables). But I could only get these variables inside the body of the node function, not in the definition of input/outputs

Feedback welcome

### DIFF
--- a/packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/typescript/userUtils/quaternion2rpy
+++ b/packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/typescript/userUtils/quaternion2rpy
@@ -1,0 +1,36 @@
+// utility function to convert from ROS Quaternion message type to roll, pitch, yaw (in degrees)
+// this is useful to make human readable plots of a robot's IMU or Pose (any message with a 'quaternion' or 'orientation' field)
+// inspired by plotjuggler feature that does this behind the scenes automatically
+// https://github.com/facontidavide/PlotJuggler/blob/f5f4bdc2176ebec2505c55f21bdf711f86f79cdc/plotjuggler_app/resources/default.snippets.xml
+
+import { Input } from "./types";
+type Output = {
+  roll_deg: number;
+  pitch_deg: number;
+  yaw_deg: number;
+};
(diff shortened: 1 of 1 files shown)


foxymiles:
Thanks for the PR!

I'd like to write a test for this. What should the output be for `{x: 1, y: 2, z: 3, w: 1}`?

ljburtz:
great to hear!
Note that by convention quaternions must be unit normalized
For example the following input:

x: 0.03813457647485015
y: 0.189307857412
z: 0.2392983377447303
w: 0.9515485246437886

should output:
roll: 10.0 degrees
pitch: 20.0 degrees
yaw: 30.0 degrees

would that work for the unit test?

ljburtz:
and I had forgotten to attach it but quick screenshot of the node in action
and making sure the roll/pitch/yaw angles are correct (with a real robot / very noisy IMU)
Screenshot from 2022-07-28 17-40-01

foxymiles:
@ljburtz Yes thanks that will work. I'll add it as a test case.

foxymiles:
@ljburtz Would it make sense to call the conversion function `quaternionToEuler`?

ljburtz:
@foxymiles absolutely quaternionToEuler is a better name!

foxymiles:
@ljburtz I have merged these changes with some testing enhancements in this PR with you listed as a co-author: https://github.com/foxglove/studio/pull/4009

Thanks for the PR!

Issue #3962, opened 2022-07-27
Support deleting multiple layouts from context menu


amacneil:
When multiple layouts are selected:
- Remove the trash can icon
- Implement "Delete" context menu item to delete all selected layouts (not just one)
- Disable any context menu items that don't apply to all selected layouts

Optional extras
- Implement bulk "save changes"
- Implement bulk "duplicate"

PR #3972, opened 2022-07-28
Use context menu for multiple deletes

foxymiles:
**User-Facing Changes**
This replaces the multi-delete icon in the layout browser header with the context menu delete item.

**Description**
Instead of having a special icon for deleting multiple layouts this just uses the existing context menu delete item to delete whichever layouts are selected.

Partially addresses #3962

### DIFF
--- a/packages/studio-base/src/components/LayoutBrowser/LayoutRow.tsx
+++ b/packages/studio-base/src/components/LayoutBrowser/LayoutRow.tsx
@@ -223,15 +223,15 @@ export default React.memo(function LayoutRow({
   }, []);
 
   const confirmDelete = useCallback(() => {
+    const layoutWarning =
+      !multiSelection && layoutIsShared(layout)
+        ? "Team members will no longer be able to access this layout. "
+        : "";
+    const prompt = `${layoutWarning}This action cannot be undone.`;
+    const title = multiSelection ? "Delete selected layouts?" : `Delete “${layout.name}”?`;
     void confirm({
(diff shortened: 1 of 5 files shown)


packages/studio-base/src/components/LayoutBrowser/LayoutRow.tsx
```diff
@@ -223,18 +224,22 @@ export default React.memo(function LayoutRow({
 
   const confirmDelete = useCallback(() => {
     void confirm({
-      title: `Delete “${layout.name}”?`,
-      prompt: `${
-        layoutIsShared(layout) ? "Team members will no longer be able to access this layout." : ""
-      } This action cannot be undone.`,
+      title: multiSelection ? "Delete selected layouts?" : `Delete “${layout.name}”?`,
```
defunctzombie, line 227: Can we un-ternary the nested ternary code any here? I find one level of ternary about as much as I can sensibly reason through.

defunctzombie (approved):

packages/studio-base/src/components/LayoutBrowser/LayoutRow.tsx
foxymiles: These are separate ternaries but I can replace them with temp vars if you want.

foxymiles: Agree about nested ternaries and readability.

amacneil:
Works great 😄

PR #3973, opened 2022-07-28
Persist raw messages panel expansion state

### LINKED ISSUE #3963: Change raw message expansion behavior
- Remove "Auto expand" setting
- Automatically decide on default expand state only when the very first message is seen
- Store the full expand state in the layout
- All expand/collapse decisions are up to me after that

Results:
- Changing tabs and returning should keep the same expanded items
- Hitting play/pause should keep the same expanded items

foxymiles:
**User-Facing Changes**
This persists the expansion state of nodes in the raw messages panel.

**Description**
This replaces the auto expansion setting with persistence of the user's interactions with the nodes. Nodes that have not been manually expanded or collapsed are still subject to the auto-expansion rule that anything other than arrays > 5 members are expanded by default.

Fixes #3963


packages/studio-base/src/panels/RawMessages/index.tsx
```diff
@@ -236,27 +201,35 @@ function RawMessages(props: Props) {
...
-      } else {
-        expandedFieldsCopy.add(key);
-        setExpandedFields(expandedFieldsCopy);
-      }
-      setExpandAll(undefined);
+      setExpansion((old) => {
+        if (old === "all") {
+          return { [key]: "c" };
+        } else if (old === "none") {
+          return { [key]: "e" };
```
amacneil, line 214: since there are only three states (expanded/collapsed/undefined), why don't you store it in the layout as a bool (true/false/undefined)?

packages/studio-base/src/panels/RawMessages/index.tsx
```diff
@@ -660,12 +644,12 @@ function RawMessages(props: Props) {
         </IconButton>
         <IconButton
           className={classes.iconButton}
-          title={expandAll ?? false ? "Collapse all" : "Expand all"}
+          title={expansion === "all" ? "Collapse all" : "Expand all"}
```
amacneil, line 647: The default action of this button isn't right. When I create a new raw message panel, load a topic, then click this button, nothing happens, because its default action is to "expand all", but all my fields are already expanded.

I think the button action needs to be something like:
- If at least one field is collapsed, expand all
- Otherwise, collapse all

You could argue for inverting it (collapse all if at least one field is expanded, otherwise expand all). That would probably be fine too.

But what we have right now is weird because I can get in a state where all fields are already expanded, and the button does nothing.

packages/studio-base/src/panels/RawMessages/index.tsx
foxymiles: I'm trying to save space in the layout by omitting keys that haven't been clicked and c/e is more compact than true/false.

packages/studio-base/src/panels/RawMessages/index.tsx
foxymiles: Ok I'll see if I can improve this.

packages/studio-base/src/panels/RawMessages/index.tsx
amacneil: Omitting keys that haven't been clicked makes sense. But for the keys we have set, bool is more compact than string right?

At least in memory... I guess if it's serialized as a json string `true` is one byte longer than `"e"` 😭

amacneil:
FYI I did some testing to make sure this didn't break existing panels, and it seemed to handle things fine.

packages/studio-base/src/panels/RawMessages/index.tsx
foxymiles: This is not straightforward to fix without compromising performance. I'd have to generate the full set of possible expansion paths including all arrays and then compare that to what the user has clicked.

foxymiles: FWIW the old implementation before we added the auto expand feature didn't handle this correctly either.

amacneil: How about a slight tweak on my proposal then:

- If at least one field has been manually collapsed, expand all
- Otherwise, collapse all

foxymiles: Ok I've updated it to use this heuristic.

amacneil:
Discussion from live call: We should reset back to "auto" expand state when changing topic.

amacneil (approved):
cool, this is much better than current behavior

PR #3978, opened 2022-07-28
Replace signal with Condvar

defunctzombie:
**User-Facing Changes**
None

**Description**
Replace some instances of signal() with Condvar

- UserNodePlayer
- CurrentLayoutProvider
- storybook/preview
- MessagePipeline

### DIFF
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -6,7 +6,7 @@ import { Story, StoryContext } from "@storybook/react";
 import { useMemo, useRef } from "react";
 import { ToastProvider } from "react-toast-notifications";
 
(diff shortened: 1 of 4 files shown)


defunctzombie:
@foxymiles @jtbandes thoughts on this? In some places the condvar feels clear - in others maybe not so much?

@foxymiles btw - the use of signal in UserNodePlayer with terminate I think was not doing what we thought but with Condvar it is now?

defunctzombie:
The main _gotcha_ with some of the uses of Condvar is that you have to call _wait_ before _notifyAll_ whereas with signal - the signal is the promise.

foxymiles (approved):
Seems like a pretty straightforward translation.

PR #3993, opened 2022-07-29
Save current user script before creating a new one

### LINKED ISSUE #3986: Creating a new User Script discards unsaved changes
**Description**
1. Create a user script (node playground)
2. Make some changes, but do not save it yet
3. Create a new user script (either using the + icon or from a template)
4. Your unsaved changes are discarded :(

Strangely, if you switch to a different node, or switch to viewing one of the utility files, we auto-save your changes. This only seems to happen when you create a new node.

**Solution**
Either auto-save the changes (like we do when you switch nodes), or prompt with a modal to save changes.

https://user-images.githubusercontent.com/637671/181659975-37efa128-e6be-4da2-a99b-30f5e5c10b2a.mov

foxymiles:
**User-Facing Changes**
This fixes an issue with the user script editor.

**Description**
The fix is simply to use the same logic from saving before switching nodes when creating new nodes.

Fixes #3986

### DIFF
--- a/packages/studio-base/src/panels/NodePlayground/index.tsx
+++ b/packages/studio-base/src/panels/NodePlayground/index.tsx
@@ -263,8 +263,22 @@ function NodePlayground(props: Props) {
     });
   }, [currentScript]);
 
+  const saveCurrentNode = useCallback(() => {
+    if (
+      selectedNodeId != undefined &&
+      selectedNode &&
+      currentScript &&
+      isCurrentScriptSelectedNode
+    ) {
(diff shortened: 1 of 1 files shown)


defunctzombie (approved):
probably we can start to move away from the "node" name. `saveCurrentScript` ? :D

foxymiles:
> probably we can start to move away from the "node" name. `saveCurrentScript` ? :D

Wasn't sure what to do with that. There are loads of code references to `node` so it might be more confusing to have a mix?

PR #3994, opened 2022-07-29
Search on panel description as well as title in panel list

### LINKED ISSUE #3985: Searching "node playground" in the panel list should show User Scripts
We renamed `Node Playground` to `User Scripts` in https://github.com/foxglove/studio/pull/3883.

In order to make this transition more seamless for users, we should return this in search results for `Node Playground` in the empty layout, `Add Panel` sidebar and the `Change Panel` context menu.

One idea for achieving this is by adding something like "Previously known as Node Playground" to the panel description, and including panel descriptions in the search text.

foxymiles:
**User-Facing Changes**
Search by panel description as well as title in the panel list.

**Description**
This expands search matching to panel descriptions and also adds `Previously known as Node Playground` to the user scripts panel so users searching for `node playground` will still get a match.

Fixes #3985

### DIFF
--- a/packages/studio-base/src/components/PanelList/index.stories.tsx
+++ b/packages/studio-base/src/components/PanelList/index.stories.tsx
@@ -47,6 +47,7 @@ const allPanels: PanelInfo[] = [
   {
     title: "Preconfigured Panel AAA",
     type: "Sample1",
+    description: "Panel description",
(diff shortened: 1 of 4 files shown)


defunctzombie (approved):

amacneil (approved):
Cool, works well in my testing

PR #3996, opened 2022-07-29
Add RosDb3WorkerIerableSource

defunctzombie:
**User-Facing Changes**
None

**Description**
The RosDb3WorkerIterableSource implements an IterableSource interface for reading ros2 .db3 files.

Reading ros2 .db3 files must be done in a worker to leverage the FileReaderSync interface for sqlite. We need this interface to read files larger than can fit in memory. To support using a worker, this change uses the comlink library and a custom transfer handler for serializing/deserializing the async message iterator.

Fixes: #3983


packages/comlink-transfer-handlers/src/iterableTransferHandler.ts
```diff
@@ -0,0 +1,61 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { transferHandlers, proxy, TransferHandler } from "comlink";
+
+const isObject = (val: unknown): val is object => typeof val === "object" && val != undefined;
```
foxymiles, line 7: Can you use lodash `isObject` here?

foxymiles (approved):

packages/comlink-transfer-handlers/src/iterableTransferHandler.ts
defunctzombie: Will switch to lodash `isObject` if it works

PR #4001, opened 2022-07-30
Fix plot panel auto-complete from incomplete topic

jameskuszmaul-brt:
Previously, if you created a plot panel, and e.g. wanted to plot a topic
named `"/drivetrain frc971.control_loops.drivetrain.Output".left_high`
and then typed `output` into the legend, the aforementioned field might
be listed in the auto-complete options, but if you selected it then
it would append a `.` and auto-complete to `"/drivetrain frc971.control_loops.drivetrain.Output".left_high.`.
This was a consequence of the logic checking against this exact scenario
being broken--in particular, at least the JSON and flatbuffer
parsers for MCAP files set `isComplex` to undefined rather than to false.
Given that multiple tests seem to duplicate this assumption, I am
assuming that this is the fault of the MessagePathInput and not the MCAP
parser.

Added a small extra story to sanity check for regressions in this code.

**User-Facing Changes**

Fixes autocomplete in certain scenarios.

### DIFF
--- a/packages/studio-base/src/components/MessagePathSyntax/MessagePathInput.stories.tsx
+++ b/packages/studio-base/src/components/MessagePathSyntax/MessagePathInput.stories.tsx
@@ -13,6 +13,7 @@
 
 import { Stack } from "@mui/material";
 import { storiesOf } from "@storybook/react";
+import TestUtils from "react-dom/test-utils";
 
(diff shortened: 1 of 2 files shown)


jameskuszmaul-brt:
@foxymiles Tests should be functional; I assume someone needs to approve the initial story baselines.

foxymiles (approved):
Looks good! Thanks for the PR and for adding the stories.

Issue #3967, opened 2022-07-28
Db3: Data source topic list should show all topics, including  unsupported ones


amacneil:
**Description**
I downloaded an open source db3 file from [here](https://autowarefoundation.github.io/autoware-documentation/main/tutorials/ad-hoc-simulation/rosbag-replay-simulation/).

When I open it in Foxglove Studio, the "Topics" list is missing a few topics:

<img width="608" alt="image" src="https://user-images.githubusercontent.com/637671/181411919-d09c9fb1-f0d4-45e4-b033-86dff6620fdc.png">

They do show up in the "Problems" tab:

<img width="522" alt="image" src="https://user-images.githubusercontent.com/637671/181412216-10ff5a8c-fca9-453f-88e1-99385d7b13f8.png">

I think we should show all topics in the Topics list, even if we are missing a message definition. We could still list the topic name, schema name, and publish frequency. We could also show an error inline next to the topic (similar to sidebar settings errors).

amacneil:
cc @defunctzombie for feedback

defunctzombie:
I think it's up to the data source to emit the topics it wants to make available. In this case the db3 source I probably not adding these topics to its list since it doesn't have definitions for them.

Its a simple change to switch up the logic here: https://github.com/foxglove/studio/blob/main/packages/studio-base/src/players/IterablePlayer/rosdb3/RosDb3IterableSource.worker.ts#L74-L80

I don't have super strong feelings either way but seems fine to show the user the topic and datatype and change the warnings to something more actionable.

PR #4003, opened 2022-07-30
Update RosDb3IterableSource to list all topics

defunctzombie:
**User-Facing Changes**
Data source info shows all topics found in .db3 files - even those we cannot deserialize.

**Description**
Even if a topic cannot be deserialized (due to missing message definitions), we still show it in the topics results so the user knows it is present in their file.

Fixes: #3967

### DIFF
--- a/packages/studio-base/src/players/IterablePlayer/rosdb3/RosDb3IterableSource.worker.ts
+++ b/packages/studio-base/src/players/IterablePlayer/rosdb3/RosDb3IterableSource.worker.ts
@@ -70,24 +70,25 @@ export class RosDb3IterableSource implements IIterableSource {
     const parsedMessageDefinitionsByTopic: ParsedMessageDefinitionsByTopic = {};
 
     for (const topicDef of topicDefs) {
-      const parsedMsgdef = ROS2_TO_DEFINITIONS.get(topicDef.type);
-      if (parsedMsgdef == undefined) {
-        problems.push({
-          severity: "warn",
-          message: `Topic "${topicDef.name}" has unrecognized datatype "${topicDef.type}"`,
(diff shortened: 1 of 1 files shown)


defunctzombie:
Alternative guide for converting .db3 to mcap: https://github.com/foxglove/mcap/tree/main/go/cli/mcap#examples Maybe a more direct link.

I think this item is common enough that we should have a dedicated docs page to it.

packages/studio-base/src/players/IterablePlayer/rosdb3/RosDb3IterableSource.worker.ts
```diff
@@ -70,24 +70,25 @@ export class RosDb3IterableSource implements IIterableSource {
...
+      }
+
       const parsedMsgdef = ROS2_TO_DEFINITIONS.get(topicDef.type);
       if (parsedMsgdef == undefined) {
         problems.push({
           severity: "warn",
-          message: `Topic "${topicDef.name}" has unrecognized datatype "${topicDef.type}"`,
-          tip: "ROS 2 bags don't contain full message definitions, so only well-known ROS types are supported in Studio. As a workaround, you can try using a Rosbridge WebSocket connection. For more information, see: https://github.com/ros2/rosbag2/issues/782",
+          message: `Topic "${topicDef.name}" has unsupported datatype "${topicDef.type}"`,
+          tip: "ROS 2 .db3 files do not contain message definitions, so only well-known ROS types are supported in Studio. As a workaround, you can convert the db3 file to mcap using the mcap CLI. For more information, see: https://foxglove.dev/docs/studio/connection/local-file",
```
amacneil, line 85:
```suggestion
          tip: "ROS 2 .db3 files do not contain message definitions, so only well-known ROS types are supported in Foxglove Studio. As a workaround, you can convert the db3 file to mcap using the mcap CLI. For more information, see: https://foxglove.dev/docs/studio/connection/local-file",
```

amacneil (approved):

PR #4007, opened 2022-08-01
Fix renaming unsaved layouts

### LINKED ISSUE #3966: Renaming an unsaved layout does not work
**Description**

https://user-images.githubusercontent.com/84792/181398961-754511fc-0521-47c2-b709-fe7c06ba7aad.mov

Once I saved the layout I could rename it.

I think this might have regressed in: https://github.com/foxglove/studio/commit/5ab7425bbd418d44e173b288f6007f45d8aaca37 or https://github.com/foxglove/studio/commit/adaeaf08c3242067736ebffc903897e060de9dfd ? since those touched layout logic recently.

foxymiles:
**User-Facing Changes**
This fixes renaming of unsaved layouts.

**Description**
For some reason the input is being immediately submitted after being focused. Dynamically inserting and removing the input from the DOM causes a lot of complexity in the event handling so this fix just keeps the input in the DOM and toggles its `display` property when needed.

Fixes #3966

### DIFF
--- a/packages/studio-base/src/components/LayoutBrowser/LayoutRow.tsx
+++ b/packages/studio-base/src/components/LayoutBrowser/LayoutRow.tsx
@@ -17,7 +17,16 @@ import {
   TextField,
   styled as muiStyled,
 } from "@mui/material";
-import { useCallback, useContext, useLayoutEffect, useMemo, useState, MouseEvent } from "react";
+import {
+  useCallback,
+  useContext,
+  useLayoutEffect,
+  useMemo,
+  useState,
+  MouseEvent,
+  useEffect,
+  useRef,
+} from "react";
 import { useMountedState } from "react-use";
 
 import { useLayoutManager } from "@foxglove/studio-base/context/LayoutManagerContext";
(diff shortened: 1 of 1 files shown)


defunctzombie (approved):
Can we use the TextField as the display as well rather than in edit only mode? Something like toggling its editable state if thats a thing?

foxymiles:
> Can we use the TextField as the display as well rather than in edit only mode? Something like toggling its editable state if thats a thing?

Yeah I was considering that. I'd need to do some more styling work since the text field has a different background but I can do that if you prefer.

defunctzombie:
> Yeah I was considering that. I'd need to do some more styling work since the text field has a different background but I can do that if you prefer.

It seemed cleaner to me rather than changing the underlying content but maybe its not so common to do this. Having a re-usable component that encapsulates this would let us avoid this error in other places (if we have any such places).

Seems fine to get this bugfix out if you don't think custom component is worth it.

foxymiles:
> > Yeah I was considering that. I'd need to do some more styling work since the text field has a different background but I can do that if you prefer.
> 
> It seemed cleaner to me rather than changing the underlying content but maybe its not so common to do this. Having a re-usable component that encapsulates this would let us avoid this error in other places (if we have any such places).
> 
> Seems fine to get this bugfix out if you don't think custom component is worth it.

I think I'll get this fix out and then try a custom component separately since there are a few other places where we could use such a thing.

PR #4009, opened 2022-08-01
Add quaternion to euler utility to user scripts utilties

foxymiles:
**User-Facing Changes**
This adds a new quaternion utility library for user scripts.

**Description**
This adds typedefs and a `quaternionToEuler` utility to the user scripts utility library. This is based on https://github.com/foxglove/studio/pull/3971


packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/typescript/userUtils/quaternions.ts
```diff
@@ -0,0 +1,41 @@
...
+export type Euler = {
+  x: number;
+  y: number;
+  z: number;
+  roll: number;
+  pitch: number;
+  yaw: number;
+};
+
+export function quaternionToEuler(quaternion: Quaternion): Euler {
```
jtbandes, line 21: Can we add a doc comment describing what this does and the axis order conventions used?

Where did this algorithm come from? Can we link to an original source, or if we derived it ourselves let's add some comments to explain it.

packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/typescript/userUtils/quaternions.ts
```diff
@@ -0,0 +1,41 @@
...
+  const roll = toDegrees * Math.atan2(dcm21, dcm22);
+  const pitch = toDegrees * Math.asin(-dcm20);
+  const yaw = toDegrees * Math.atan2(dcm10, dcm00);
+  return {
+    x: roll,
+    y: pitch,
+    z: yaw,
+    roll,
+    pitch,
+    yaw,
```
jtbandes, line 39: If these three are the same, why do we need all six?

packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/typescript/userUtils/quaternions.ts
foxymiles: This is based on a PR from user @ljburtz.

packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/typescript/userUtils/quaternions.ts
foxymiles: It seems like euler angles are described either as `x, y, z` or `roll, pitch, yaw` so I put both in for convenience.

The algorithm here uses them interchangeably:
https://en.wikipedia.org/wiki/Conversion_between_quaternions_and_Euler_angles

jhurliman (approved):
LGTM, the method could use a jsdoc comment explaining what it does and that the output is RPY degrees

jtbandes:
Is it normal to represent euler angles in degrees or should we provide a separate function/constant to do that conversion?

foxymiles:
> Is it normal to represent euler angles in degrees or should we provide a separate function/constant to do that conversion?

I wondered about that too but @jhurliman says degrees are more commonly used in robotics applications.

packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/typescript/userUtils/quaternions.ts
KerstinKeller: As a user, it would greatly irritate me to have a struct that contains values twice.
In your place I would stick to one of them (we usually use `roll, pitch, yaw` - then `psi, theta, phi`  or `alpha, beta, gamma`. But never heard of `x, y, z`).

foxymiles: Interesting. Thanks for your feedback. @jhurliman Should we remove the x,y,z fields here?

ljburtz: my 2 cents -> only keep roll/pitch/yaw

agreed with Kerstin, i don't understand the added convenience of duplicating. less values is actually more user friendly when we are selecting fields to plot from the drop down (which is the use case here)

also x,y,z are ambiguous -> usually reserved for spatial coordinates / axis themselves

packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/typescript/userUtils/quaternions.ts
ljburtz: axis convention used: static frame, x, y, z (it's the ROS default)
see https://github.com/ros/geometry/blob/d51887c976a45ba01db9c61361e94076d3bcf3f1/tf/src/tf/transformations.py
(euler_from_quaternion() )

algorithms is also available in the ROS standard Plotjuggler
https://github.com/facontidavide/PlotJuggler/blob/f5f4bdc2176ebec2505c55f21bdf711f86f79cdc/plotjuggler_app/resources/default.snippets.xml

packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/typescript/userUtils/quaternions.ts
jhurliman: I agree with the above, remove x,y,z in favor of the more descriptive roll,pitch,yaw

jhurliman:
Clarification: degrees are more common for visualization (in robotics applications). Meaningful positions like 90 or 180 degrees align with grid lines in a plot using degrees, but drawing grid lines at `Pi` or `Tau` is difficult.

foxymiles:
I've removed the extra fields in this PR: https://github.com/foxglove/studio/pull/4027

PR #4014, opened 2022-08-02
Wait on current emit promise before emitting for tick

defunctzombie:
**User-Facing Changes**
Messages would be dropped in some situations when using the new latching players. This ensures all messages are delivered.

**Description**
We must wait on the current emit to complete before emitting a tick in the IterablePlayer. If we don't wait for the tick to complete we might drop messages since the emitState call might get debounced.

### DIFF
--- a/packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
+++ b/packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
@@ -790,6 +790,11 @@ export class IterablePlayer implements Player {
       return;
     }
 
+    // Wait on any active emit state to finish as part of this tick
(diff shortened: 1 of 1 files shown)


defunctzombie:
This bug was introduced when I moved the IterablePlayer to use debouncePromise for the emit state calls. Prior to that I was awaiting on each emitState within the tick which produced the desired behavior of ensuring all messages were delivered. With debounce promise the tick call must wait for the previous emit to finish so it can be sure its emit state call won't be dropped.

jhurliman (approved):

foxymiles (approved):

jtbandes:
👏

PR #4021, opened 2022-08-03
[New3D] Patch THREE.js to improve custom shader performance

jhurliman:
**User-Facing Changes**

- Improved `3D (Beta)` TEXT_VIEW_FACING rendering performance

**Description**

For custom shaders, THREE.js uses the internal WebGLShaderCache class which contains a `Map<string, WebGLShaderStage>` where the keys are vertex and fragment shader source code. This change patches THREE.js (via `yarn patch`) to use `Material#vertexShaderKey` and/or `Material#fragmentShaderKey` properties as map keys when available. The cache lookup methods have also been rewritten to avoid unnecessary lookups.

These new properties are used to speed up shader execution for TEXT_VIEW_FACING markers (`RenderableTextViewFacing`). Performance tracing shows a ~30% reduction in frame times in the synthetic label benchmark on my machine after this change.


defunctzombie:
> Performance tracing shows a ~30% reduction in frame times in the synthetic label benchmark on my machine after this change.

Were you able to use the benchmark player to see the same improvements?

.yarn/patches/three-npm-0.140.2-944041dff4
```diff
@@ -0,0 +1,417 @@
+diff --git a/build/three.cjs b/build/three.cjs
```
jtbandes, line 1: please add a `.patch` to the filename so github highlights it properly

packages/three-text/src/LabelPool.ts
```diff
@@ -137,6 +137,16 @@ void main() {
...
+
+    // We use a patched version of THREE.js where the internal WebGLShaderCache class has been
+    // modified to allow caching based on `vertexShaderKey` and/or `fragmentShaderKey` instead of
+    // using the full shader source as a Map key
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    (this as any).vertexShaderKey = "LabelMaterial-VertexShader";
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    (this as any).fragmentShaderKey = `LabelMaterial-FragmentShader${
+      params.picking === true ? "-picking" : ""
+    }`;
```
jtbandes, line 149: This class is in a separate package that we should move out to a new repo in the future. Can we keep it free of assumptions about the patched version of three.js? For example would it work to move this to the studio-base package and change it to `LabelMaterial.prototype.vertexShaderKey = ...` ?

jtbandes (approved):

packages/three-text/src/LabelPool.ts
jhurliman: I don't know how to conditionally set `fragmentShaderKey` based on the `picking` parameter when using the prototype approach.

jtbandes: We could expose `picking` as a property on this object and use that? I am not in favor of the current `(this as any)` because it introduces a dependency of studio on this hidden feature, and it's also just confusing code because the type of `this` is always known.

jtbandes:
Have we searched for existing github issues on three.js about this performance concern? I think we should file one if there isn't one already, so that they are aware of the problem.

Issue #3989, opened 2022-07-29
[feature from plotjuggler] simple X/Y plot


ljburtz:
Hello, migrating from plotjuggler to foxglove but a currently blocking use case is comparing robot odometry calculated by the robot to a ground truth

for this I need to display an x/y plot -> top down map
(the x axis is the odometry x, the y axis is the odometry y. there is no timestamp indexing as in a regular time series)

this works great in foxglove for a single plot by using "msg path accumulated) as an index (see screenshot, blue line)
![image](https://user-images.githubusercontent.com/26211495/181671710-42dea49e-f3a3-4ef4-a49e-dd72cd94a027.png)

However, I can't find out how to display two (or three) overlapping robot trajectories on the same plot = to compare the robot's odometry estimate with the ground truth

See screenshot of this feature in plotjuggler -> x/y plots make it very intuitive to compare estimated trajectories to ground truth
![image](https://user-images.githubusercontent.com/26211495/181671773-de1c027b-d0e7-4c7b-927f-02aeaf0d9a09.png)
two different trajectory estimates, drifting away from the ground truth as the robot travels

Thanks!

ljburtz:
Maybe in foxglove parlance these types of x/y plots are a separate panel type?
Something closer to the Map panel? But not using a NavSatFix message (longitude / latitude) but simple x and y  coordinates pairs from selectable topic's message field?

basically the Map panel is overkill for many use cases (that don't use longitude/latitude) and the current plot panel is best for timeseries where the plot's x axis is a timestamp, not a coordinate

ljburtz:
this issue is similar to #1393 of 1 year ago!

PR #4022, opened 2022-08-03
[3D panel] Add utility to create markers for trajectory plots

ljburtz:
**User-Facing Changes**
new user script utility to create markers

**Description**
example plotting several x/y coordinates on the 3D panel
use case: plot robot trajectory (estimated vs ground truth etc)

trigger: workaround suggested on Slack until the Plot panel can support several x/y pairs on the same plot
related feature request: #3989
but create_marker utility function is generic for other use cases as well

### DIFF
--- a/packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/typescript/userUtils/trajectoryPlot.ts
+++ b/packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/typescript/userUtils/trajectoryPlot.ts
@@ -1,0 +1,114 @@
+// example plotting several x/y coordinates on the 3D panel
+// use case: robot trajectory (estimated vs ground truth etc)
+
+import { Input, Point, Message, Pose, RGBA, Header } from "./types";
+type Marker = Message<"visualization_msgs/Marker">;
+
+export const inputs = [
+  "/small_scout_1/rover_odometry_gt",
+  "/small_scout_1/rover_odometry",
+  "/small_scout_1/rover_odometry_naive",
+];
+export const output = "/small_scout_1/trajectory2dplot";
+
+export function create_marker(
+  position: Point,
+  header: Header,
+  ns: string,
+  color: RGBA
+): Marker {
+  // Prepare Marker-specific attributes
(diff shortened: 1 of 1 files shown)


ljburtz:
I committed the entire script I created for my use case. I understand that this creates several unsuccessful checks. I guess the intent is that only the utility function create_marker() is included in the PR? let me know if you prefer that / or seing the entire script in action and then cherry picking yourself!

screenshot of the resulting 'plot' on the 3D panel (beta) :
image

defunctzombie:
@ljburtz Sorry this slipped through and no one got assigned to review or provide feedback. If this is still something you think we should include in the app, my initial feedback is that we'd want to find a more generic script that avoids specific topics like `/small_scout_1/rover_odometry_gt`. Also it appears that this utility is actually a full user script rather than a utility and should be re-worked into a utility unless I miss-understand the PR title and description.

ljburtz:
Thanks @defunctzombie for following up.


> it appears that this utility is actually a full user script rather than a utility 

absolutely. the generic utility that I think would benefit people is just the create_marker() function
-> useful factory function for the 3d plot panel in several use cases

the above script is, as you said, not something to merge directly. I thought that you being able to see a full example would have a more clear idea of what to cherry pick / the use of that create_marker() function

> If this is still something you think we should include in the app, 

Yes this is the workaround our team is using to visualize robot trajectories, until the plot panel can support several x/y message pairs. (we may have talked about it together during Roscon in Kyoto)

jtbandes:
Thanks for the contribution. It is interesting to see the ideas sketched out here, but this PR is not something we could accept in its current form, so I'm going to close it out. I see there's already an open feature request at https://github.com/foxglove/studio/issues/3989, but for other things such as user script templates or functions, it would be more appropriate to create a separate feature request, or a PR that proposes the specific changes so they can be reviewed. You might be interested in a feature we recently released, [message converter extensions](https://foxglove.dev/blog/announcing-message-converter-extensions-in-foxglove-studio) which can provide an alternative to user scripts that are packaged as an extension (which can be automatically installed for all users in your org).

PR #4024, opened 2022-08-03
Convert components/sidebar to mui

defunctzombie:
**User-Facing Changes**
None

**Description**
Convert `components/Sidebar` to mui from fluentui.

Fixes: #4020
Fixes: #4030
Fixes: #4031


jtbandes:
When the window is small, the items at the bottom don't seem to be pinned anymore. What happened to that behavior?


https://user-images.githubusercontent.com/14237/184689348-0b7c654b-8a4f-45a8-a2e0-66a9950f3e44.mov

We also lost the hover state on the sidebar buttons. Can we get that back?

defunctzombie:
@jtbandes I can bring the bottom items back - its a tradeoff of whether to use one set of tabs or two sets of tabs. I don't know that I feel that strongly since either way something is gonna have to scroll off screen so just having all the tabs look like one and scroll together seems equally usable.

jtbandes (approved):

2metres:
> @2metres there's some issue with the account icon. This is what I see in the signed-in state.
> 
> image

Looks like an issue with the SVG?

2metres (approved):

amacneil:
Any particular reason for reordering the bottom icons, or was that a mistake? I don't see any reference to it in the PR description.

defunctzombie:
> Any particular reason for reordering the bottom icons, or was that a mistake? I don't see any reference to it in the PR description.

Looks like a mistake. We don't intend to re-order. I will fix that before merging.

jtbandes:
>  We also lost the hover state on the sidebar buttons. Can we get that back?

No takers on this issue before merging I guess, so I filed a separate ticket: https://github.com/foxglove/studio/issues/4142

PR #4026, opened 2022-08-03
Remove fluent from preferences sidebar

foxymiles:
**User-Facing Changes**
None

**Description**
This removes all uses of fluentUI from the preferences sidebar.

Fixes https://github.com/foxglove/studio/issues/4029


defunctzombie (approved):
👍 looks good

Please reference any tickets from the fluentui epic so we can close those out.

PR #4028, opened 2022-08-03
Fix panel toolbar inputs drag interaction

### LINKED ISSUE #4019: Click + Drag on top of Raw Message panel moves panel instead of selecting text
**Description**

If you click + drag in the text box for defining what message/fields to show in the `Raw Message` panel (where the cursor is in the below image), it results in the whole panel getting dragged round, instead of selecting and highlighting text. This is somewhat irritating (if understandable) behavior. I do not know the correct resolution while still maintaining consistency with being able to move the panel.
image

- Version: whatever is at studio.foxglove.dev on 2022.08.02
- OS: Web Browser

**Steps To Reproduce**
Hold down left-click in the topic/field name text box in the `Raw Message` panel, and while left-click is still held, move your mouse left or right.

**Current Experience**

The entire panel is rearranged, potentially messing up an existing layout.

**Desired Experience**

Highlight the text between where left-click was pressed and the cursor currently is (as works, e.g., in the legend for the `Plot` panel). I should still have access to some mechanism to allow me to drag the panel around.

foxymiles:
**User-Facing Changes**
This fixes interaction conflicts with inputs embedded in the panel toolbars and drag and drop.

**Description**
The fix here is to restrict the drag and drop area to the controls at the end of the toolbar if we have any children embedded in the toolbar.

Fixes #4019

### DIFF
--- a/packages/studio-base/src/components/PanelToolbar/index.tsx
+++ b/packages/studio-base/src/components/PanelToolbar/index.tsx
@@ -85,12 +85,20 @@ export default React.memo<Props>(function PanelToolbar({
     );
   }, [additionalIcons, isFullscreen, exitFullscreen]);
 
+  // If we have children then we limit the drag area to the controls. Otherwise the entire
+  // toolbar is draggable.
+  const rootDragRef =
+    isUnknownPanel || children != undefined ? undefined : panelContext?.connectToolbarDragHandle;
+
+  const controlsDragRef =
+    isUnknownPanel || children == undefined ? undefined : panelContext?.connectToolbarDragHandle;
+
   return (
     <PanelToolbarRoot
       backgroundColor={backgroundColor}
       data-testid="mosaic-drag-handle"
       enableDrag={panelContext?.connectToolbarDragHandle != undefined}
-      ref={isUnknownPanel ? undefined : panelContext?.connectToolbarDragHandle}
+      ref={rootDragRef}
     >
       {children ??
         (panelContext != undefined && (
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/components/PanelToolbar/PanelToolbarControls.tsx
```diff
@@ -34,46 +34,48 @@ type PanelToolbarControlsProps = {
...
-export const PanelToolbarControls = React.memo(function PanelToolbarControls({
-  additionalIcons,
-  isUnknownPanel,
-  menuOpen,
-  setMenuOpen,
-}: PanelToolbarControlsProps) {
-  const panelId = useContext(PanelContext)?.id;
-  const { setSelectedPanelIds } = useSelectedPanels();
-  const { openPanelSettings } = useWorkspace();
+const PanelToolbarControlsComponent = forwardRef<HTMLDivElement, PanelToolbarControlsProps>(
```
defunctzombie, line 37: Does this make sense in its own file?

defunctzombie (approved):
Works great!

packages/studio-base/src/components/PanelToolbar/PanelToolbarControls.tsx
foxymiles: You mean just separating out the `forwardRef` part?

defunctzombie: I meant just moving it to its own file. I like 1 component per file (unless the component is like 2 or 3 lines which this one isn't). I find 1 component per file helps keep the files more maintainable.

foxymiles: There's really only one component here I just had to introduce this intermediate component so TS could understand the props.

amacneil:
A bunch of the panels are now showing drag cursors where they are not draggable. E.g. plot panel titlebar, and some small gaps in between controls on diagnostics (plus maybe other panels, I didn't test them all).

https://user-images.githubusercontent.com/637671/182680441-933d5ccf-0b96-4c45-9d0d-da4d23cfd667.mov

foxymiles:
@amacneil Good catch. I've disabled the drag cursor for toolbars that aren't draggable.

amacneil:
Ok so I'm probably venturing into nit picky land now, but the diagnostics panel titlebar still has some weirdness. Now there is actually a little region (next to setting icon) where dragging does work, but we don't show the drag cursor:

https://user-images.githubusercontent.com/637671/182710216-ed19bff1-6af4-4576-8d06-e07275868c43.mov

If this is too hard to fix, just leave it. I eventually want to enforce a consistent pattern for panel titlebars.

foxymiles:
Yeah this will happen if the panel is putting custom content into the titlebar but not completely filling it up. Fixable but more work.

defunctzombie:
:shipit:

amacneil (approved):
great improvement. did not review code.

PR #4039, opened 2022-08-03
Fix timezone preferences

### LINKED ISSUE #4036: Timezone data not found in time formats
**Description**
momentjs timezone data seems to be unavailable to the app.

- OS: mac/chrome
- Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native): nuscenes

**Steps To Reproduce**
1. Change your timezone in the preferences panel to any specific timezone.
2. Load nuscenes.
3. Hover over the playback bar.

**Expected Behavior**
The timezone data should be found and used.

**Actual Behavior**
Console logs numerous errors like this:

```
Moment Timezone has no data for zone:Africa/Algiers.
```

foxymiles:
**User-Facing Changes**
None

**Description**
This fixes an issue with the timezone selector in the preferences sidebar and adds stories to cover timezone selection and formatting.

Fixes #4036

### DIFF
--- a/packages/studio-base/src/components/PlaybackControls/index.stories.tsx
+++ b/packages/studio-base/src/components/PlaybackControls/index.stories.tsx
@@ -57,7 +57,13 @@ function getPlayerState(): PlayerState {
 }
 
 const mockAppConfiguration: IAppConfiguration = {
-  get: () => undefined,
+  get: (key: string) => {
+    if (key === "timezone") {
+      return "America/Los_Angeles";
+    } else {
+      return undefined;
+    }
+  },
   set: async () => {},
   addChangeListener: () => {},
   removeChangeListener: () => {},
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/components/PlaybackControls/index.stories.tsx
```diff
@@ -57,7 +57,13 @@ function getPlayerState(): PlayerState {
...
 
 const mockAppConfiguration: IAppConfiguration = {
-  get: () => undefined,
+  get: (key: string) => {
+    if (key === "timezone") {
+      return "America/Los_Angeles";
+    } else {
+      return undefined;
+    }
+  },
```
jtbandes, line 66: What's the purpose of this change?

jtbandes (approved):

packages/studio-base/src/components/PlaybackControls/index.stories.tsx
foxymiles: This forces the playback bar to render with a timezone set which should fail if the timezone data can't load for some reason.

jtbandes: Ok, but I would think we also want to ensure that it works if the user has never changed the setting (so it's undefined by default) right?

foxymiles: Yeah that's a good point. Maybe I'll do a followup to expand on these stories to cover both cases.

Issue #3898, opened 2022-07-21
3D Beta: Improve camera info topic matching


amacneil:
**Description**
When displaying a camera in the 3D panel, we require an associated camera info topic. I'm not sure what our algorithm is for guessing this.

Currently, it does not guess an association between these topics - I think we could detect it.
- `/camera/camera_info`
- `/camera/image_raw/compressed`

Similarly, @jhurliman suggested these topics which we should probably also detect:

- `/camera/image_raw/compressed/camera_info`
- `/camera/image_raw/compressed`

This topic matching also applies to the Image panel, but I suggest that we focus on 3D Beta for now, since we plan to rewrite the Image panel using the same rendering engine in the future.

Broken out from https://github.com/foxglove/studio/issues/3891 for further discussion.

PR #4041, opened 2022-08-03
[New3D] Make camera_info topic matching more lenient

jtbandes:
**User-Facing Changes**
Improved the automatic matching between camera images and camera info topics in the 3D (Beta) panel.

**Description**
Fixes https://github.com/foxglove/studio/issues/3898

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/color.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/color.ts
@@ -8,7 +8,11 @@ import tinycolor from "tinycolor2";
 import { lerp } from "./math";
 import { ColorRGB, ColorRGBA } from "./ros";
 
-export const SRGBToLinear = THREE.SRGBToLinear;
+// From https://github.com/mrdoob/three.js/blob/dev/src/math/ColorManagement.js
+// which is not exported
+export function SRGBToLinear(c: number): number {
+  return c < 0.04045 ? c * 0.0773993808 : Math.pow(c * 0.9478672986 + 0.0521327014, 2.4);
+}
 
 export function stringToRgba(output: ColorRGBA, colorStr: string): ColorRGBA {
   const color = tinycolor(colorStr);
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/panels/ThreeDeeRender/color.ts
```diff
@@ -65,9 +64,9 @@ export function rgbaToCssString(color: ColorRGBA): string {
 }
 
 export function rgbaToLinear(output: ColorRGBA, color: ColorRGBA): ColorRGBA {
-  output.r = SRGBToLinear(color.r);
-  output.g = SRGBToLinear(color.g);
-  output.b = SRGBToLinear(color.b);
+  output.r = THREE.SRGBToLinear(color.r);
+  output.g = THREE.SRGBToLinear(color.g);
+  output.b = THREE.SRGBToLinear(color.b);
```
amacneil, line 69: unrelated or accidental?

packages/studio-base/src/panels/ThreeDeeRender/color.ts
```diff
@@ -2,14 +2,13 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
-import { SRGBToLinear } from "three/src/math/ColorManagement";
-import { clamp } from "three/src/math/MathUtils";
+import * as THREE from "three";
```
amacneil, line 5: Why not `import { SRGBToLinear, MathUtils } from "three"` ?

amacneil (approved):
cool, works with my test bag

packages/studio-base/src/panels/ThreeDeeRender/color.ts
jtbandes: Had to change how the imports are done to make it work in jest

packages/studio-base/src/panels/ThreeDeeRender/color.ts
jtbandes: 🤷  we could, but I think THREE is a pretty standard way of using three.js

defunctzombie: Kinda weird that this import is special and given all caps we use for file level constants.

PR #4043, opened 2022-08-04
[New3D] Revert camera settings back to degrees, remove @foxglove/regl-worldview dep from New3D, cleanup

jhurliman:
**User-Facing Changes**

- Fixed: Revert `3D (Beta)` camera settings from radians back to degrees

**Description**

This fixes a regression where the user-input camera settings inadvertently changed from degrees to radians, but the input controls still used precision and step values meant for degrees. Users only deal in degrees now (even though the under the hood math uses radians). The DEFAULT_CAMERA_STATE object now lives in this codebase so the values are discoverable by reading the source code and we avoid additional rad2deg confusion, and with this change we also drop all imports of `@foxglove/regl-worldview` from `ThreeDeeRender/*`.

The _updateCameras() method has been streamlined a bit and more comments added.

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
@@ -10,7 +10,6 @@ import { DeepPartial } from "ts-essentials";
 import { v4 as uuidv4 } from "uuid";
 
 import Logger from "@foxglove/log";
-import { CameraState } from "@foxglove/regl-worldview";
 import { toNanoSec } from "@foxglove/rostime";
 import type { FrameTransform } from "@foxglove/schemas/schemas/typescript";
 import {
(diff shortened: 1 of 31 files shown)


packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -851,13 +842,14 @@ export class Renderer extends EventEmitter<RendererEvents> {
   }
 
   frameHandler = (currentTime: bigint): void => {
-    const camera = this.activeCamera();
-    this.emit("startFrame", currentTime, this);
-
+    this.currentTime = currentTime;
```
jtbandes, line 845: what did this fix?

jtbandes (approved):

jtbandes:
> Users only deal in degrees now (even though the under the hood math uses radians)

I would mildly prefer storing radians in the config, so the code is mostly dealing with radians, except when needed...not going to block the PR on this though

defunctzombie:
> I would mildly prefer storing radians in the config, so the code is mostly dealing with radians, except when needed...not going to block the PR on this though

Is it weird to store the config as one unit type when the users deal with another? Does it make it strange to reason about the config?

jhurliman:
> Is it weird to store the config as one unit type when the users deal with another? Does it make it strange to reason about the config?

That was my reasoning for storing degrees in the config. Especially for the default camera values, I would find it surprising to see `45` in the sidebar, and when I click export I see `0.7853981633974483`.

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
jhurliman: This was preparation for the upcoming follow modes work, to make sure `this.currentTime` is up to date before calling `_updateFrames()`

PR #4044, opened 2022-08-04
Use ToggleButtons for color scheme, timestamp format, and launch preference selectors

jtbandes:
**User-Facing Changes**
Not worth including in release notes

**Description**
image


amacneil:
Cool. Couple thoughts:
- I feel like Dark, Light, System is a more natural order than having System in the middle?
- I would probably change it to `Use System` or `Follow System` rather than just `System`?

foxymiles:
Looks nice! 👍 

Are there other places in the app where it would make sense to use a segmented control? It feels a bit odd that this is the only instance of it.

jtbandes:
image

Only problem is it doesn't respond well to smaller sizes:

image

jtbandes:
> Are there other places in the app where it would make sense to use a segmented control? It feels a bit odd that this is the only instance of it.


Stuff like this probably, where there are a small fixed number of choices:

image


I would also say we use it already in Settings...maybe we should make these look more consistent? I don't see a reason why the color scheme should be so different between Preferences and Panel Settings.

image

foxymiles:
> > Are there other places in the app where it would make sense to use a segmented control? It feels a bit odd that this is the only instance of it.
> 
> Stuff like this probably, where there are a small fixed number of choices:
> 
> image
> 
> I would also say we use it already in Settings...maybe we should make these look more consistent? I don't see a reason why the color scheme should be so different between Preferences and Panel Settings.
> 
> image

Yeah I think the selected state of that toggle is also not clear with that coloring.

foxymiles:
> image
> 
> Only problem is it doesn't respond well to smaller sizes:
> 
> image

Yeah responsiveness is tricky there.

jtbandes:
🤦 didn't realize this is a thing: https://mui.com/material-ui/react-toggle-button/

foxymiles:
> 🤦 didn't realize this is a thing: https://mui.com/material-ui/react-toggle-button/

I guess we could flip to the vertical mode on small screens?

jtbandes:
That might be ideal but I don't want to add a ResizeDetector just for this purpose. I just enabled flex-wrap.
image

foxymiles:
Should we make this full width to align with the other inputs?

defunctzombie:
I am mixed on this. It looks ok but the dropdown seems to look more consistent in more situations. On narrow sidebars it remains a single dropdown and still obvious to interact with.

jtbandes:
Updated to include toggle for timestamp format selection too.

amacneil:
Waiting for @2metres to return before final 👍/👎

2metres:
I like the Color Scheme toggle but no sure about the timestamp toggle

jtbandes:
How about these ones too?

image

jtbandes:
Also looks nice if we use the primary color to better match the rest of the app: 
image

2metres (approved):

Issue #3980, opened 2022-07-28
Arrow right auto play at low replay speed on bag


clementchef:
**Description**
Pressing the right arrow on keyboard triggers auto play (instead of just moving right).
This only happen at low replay speed (eg: 0.05x) on bag, when paused.

- Version:
- OS:
- Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native):
- ROS distro (if applicable):

**Steps To Reproduce**

**Expected Behavior**

**Actual Behavior**

defunctzombie:
Confirming I can reproduce by selecting a slow speed and hitting the right arrow. Will look into this.

Issue #3872, opened 2022-07-19
"Play when seeking forward" functionality is confusing UX


DavidXenakis:
**Description**
The "play when seeking forwards" functionality is confusing and breaks the assumption that clicking backwards then forwards will land the play head back at the same timestamp. IIRC, this change was made a few months ago to fix a bug in stateful visualizations, where skipped messages would result in incorrect state, and thus incorrect visualizations.

Could this feature be easily configurable, so that it could be turned off at build time?

- Version: Latest
- OS: All 
- Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native): Any
- ROS distro (if applicable): Any

**Steps To Reproduce**
Click the "Seek Forward" button, then the "Seek Backward" button. After seeking backwards, the time stamp is different than before.

https://user-images.githubusercontent.com/3076080/179814549-339db9d8-a03f-43ba-b4f1-622602f26c6c.mov

**Current Experience**
This behavior is confusing and not needed by any of our visualizations. 

**Desired Experience**
Seeking forward should jump the time once, rather than playing and stopping.

Issue #2283, opened 2021-12-06
Playback breakpoints


jtbandes:
Currently the app provides no way to "find" instances of messages on a certain topic other than by seeking back/forward until the desired message is displayed. If the messages on the desired topic are very high-frequency (such as /tf) it can be very difficult to find and display the desired message in the app.

The app could support a "breakpoints" feature where playback would automatically stop when some kind of event occurs. Such an event could be:
- A certain receiveTime is reached (this could be used to implement forward/back arrow keys as per #2270)
- A message is received on a certain topic
- A message is received on a certain topic matching a user-defined predicate (e.g. `header.frame_id="parent" AND child_frame_id="child"`)

PR #4045, opened 2022-08-04
Add optional playUntil method to Player interface

defunctzombie:
**User-Facing Changes**
Forward seek and backwards seek step forward and backwards the same amount of time.

**Description**
Our current behavior for forward seeking (via arrow key or seek button), is to "play" the data forward at least the seek amount. The purpose of this behavior is to ensure replay of stateful messages (like tf and markers) so pressing "seek" from the start of the file results in the same playback as playing from the start of the file.

Unfortunately, the way we implemented this feature was to start playback and attempt to pause playback after the seek target was reached. This behavior would result in a forward seek moving forward different and variable amounts of time since the "pause" would be a best effort invocation after some frames of messages.

This change removes this workaround play/quick-pause behavior in favor of first-class support within the player. The player can optionally implement a `playUntil` method. If available, seeking forward will invoke this method instead of regular `seek`. The player can implement `playUntil` however it wants. For IterablePlayer, the implementation emits all the messages in one "tick" to make forward seeking a small amount appear like a jump.

Fixes: #3980
Fixes: #3872
Fixes: #4059


jtbandes (commented):
I'm not sure we should make this argument optional in the API – if the calling code is expecting the player to play until a specific time, it would be bad if the player just started playing and never paused

defunctzombie:
> I'm not sure we should make this argument optional in the API – if the calling code is expecting the player to play until a specific time, it would be bad if the player just started playing and never paused

Yea - I think I landed here as well. My main hesitation was all the code I'd need to update 😭 Maybe its not that much or maybe I can wait until the great RAP purge.

I also thought about making it a separate player api method that could be optional. Maybe that is a nicer route to go. `startPlaybackUntil` ?

defunctzombie:
@jtbandes I've re-worked this to add a new optional `playUntil` method on the player interface. After writing the description I'm wondering if it should be instead a `seekUntil`. I'll wait for you to review and give any thoughts on naming or implementation.

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
```diff
@@ -718,13 +732,12 @@ export class IterablePlayer implements Player {
       throw new Error("Invariant: Tried to play with no current time.");
     }
 
+    // If we have an until time, we will read all the messages until that time in one tick
+    const targetTime = this._untilTime ?? add(this._currentTime, fromMillis(rangeMillis));
+
```
jtbandes, line 737: Are you assuming that the untilTime is not far in the future? That seems dangerous.

packages/studio-base/src/players/UserNodePlayer/index.ts
```diff
@@ -915,16 +915,47 @@ export default class UserNodePlayer implements Player {
...
-  callService = async (service: string, request: unknown): Promise<unknown> =>
-    await this._player.callService(service, request);
-  startPlayback = (): void => this._player.startPlayback?.();
-  pausePlayback = (): void => this._player.pausePlayback?.();
-  setPlaybackSpeed = (speed: number): void => this._player.setPlaybackSpeed?.(speed);
-  seekPlayback = (time: Time, backfillDuration?: Time): void =>
-    this._player.seekPlayback?.(time, backfillDuration);
-  requestBackfill = (): void => this._player.requestBackfill();
+  setPublishers(publishers: AdvertiseOptions[]): void {
+    return this._player.setPublishers(publishers);
```
jtbandes, line 919: Remove all these `return`s, the method is `void`

packages/studio-base/src/players/types.ts
```diff
@@ -65,6 +65,8 @@ export interface Player {
   startPlayback?(): void;
   pausePlayback?(): void;
   seekPlayback?(time: Time, backfillDuration?: Time): void;
+  // Advanced playback controls.
```
jtbandes, line 68: This comment is not helping anyone

jtbandes (approved):
Looks ok, would avoid the assumption that untilTime is in the near future. This feature can eventually evolve into breakpoints: https://github.com/foxglove/studio/issues/2283

packages/studio-base/src/players/types.ts
defunctzombie: It's helping my line count 🤣

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
defunctzombie: is there a reasonable assumption to clamp here? Maybe if <= 100ms then we can emit in one batch? Or maybe don't bother?

defunctzombie:
Maybe I need to guard against untilTime being set to some time behind the current playhead?

jtbandes:
That would also be a good thing to chec[](url)

PR #4047, opened 2022-08-04
Fix crash in raw messages panel

### LINKED ISSUE #4040: Message diff module crashes with array.slice is not a function
When trying to view the diffs between an array within a message, the module itself crashes. It crashes when:

- I attempt to view the element in the array
- When the array goes from 1 item length to 0 item length

- Version: 1.20.1 
- OS: Darwin
- Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native): bag
- ROS distro (if applicable):

**Steps To Reproduce**
Replayable bag sent over Slack.

**Expected Behavior**
Not to crash the panel.

**Actual Behavior**
Panel crashes

foxymiles:
**User-Facing Changes**
Fixes a crash in the raw messages panel.

**Description**
The logic here was trying to cast a `DataView` to an array. We need to check if the object is an array or a `DataView` and if it's a `DataView` wrap it in a sliceable array.

Fixes #4040

### DIFF
--- a/packages/studio-base/src/panels/RawMessages/fixture.ts
+++ b/packages/studio-base/src/panels/RawMessages/fixture.ts
@@ -63,6 +63,7 @@ export const fixture = {
         receiveTime: { sec: 122, nsec: 456789011 },
         message: {
           some_array: ["a", "b", "c"],
+          some_data_view: new DataView(new ArrayBuffer(10)),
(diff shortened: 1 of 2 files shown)


foxymiles:
Closed in favor of https://github.com/foxglove/studio/pull/4070.

PR #4051, opened 2022-08-04
Remove strict mode toggle from experimental features

jtbandes:
**User-Facing Changes**
None

**Description**
This experimental was only enable in dev mode (because strict mode operates in dev mode). We have been running with strict mode for a while and the issues have been ironed out; there's no reason to keep this toggle around.


defunctzombie (commented):
I've been running with this feature off. I am not a fan of strict mode, what it did to console logs (maybe that was fixed), and what it does to how my code runs even when dependencies don't change. I would not be in favor of removing this and always enabling strict mode.

jtbandes:
If you liked strict mode you’re gonna love react 18

defunctzombie:
:shipit:

defunctzombie (approved):

PR #4052, opened 2022-08-05
Upgrade to React 18.2.0

jtbandes:
**User-Facing Changes**
None

**Description**
- Upgrade react, react-dom, and related packages
- Switch from @testing-library/react-hooks to @testing-library/react (see https://github.com/foxglove/studio/pull/4063)
- Update a handful of test assertions to account for slight differences in render count/order
- Tweak some stories, and disable a few pending debugging with Chromatic support (FG-1083)


jtbandes:
Some day.

packages/studio-base/src/components/MessagePipeline/index.test.tsx
```diff
@@ -38,7 +38,8 @@ import { MAX_PROMISE_TIMEOUT_TIME_MS } from "./pauseFrameForPromise";
 
 jest.setTimeout(MAX_PROMISE_TIMEOUT_TIME_MS * 3);
 
-// We require two state updates for each player emit() to take effect, because we  React 18 / @testing-library/react,
+// We require two act() calls for each player emit() to take effect because of some quirks with
```
defunctzombie, line 41: Do we have any notion of what these quirks are? This is a tough comment to address in the future - maybe by fixing the player emit? if we don't know the quirks.

jtbandes: Nope. What type of "addressing" are you hoping to see in the future? The app works, and the tests work.

snosenzo:
Seems to be no lingering `IterablePlayer` leaks anymore from the scrubber after some cursory testing 👍

packages/studio-base/src/components/MessagePipeline/index.test.tsx
defunctzombie: I don't think it is a crazy ask here that this comment do more to explain why we need two act calls when initially we would imagine we only need one? Is it because we update state more than we should? Is it because of strict mode running things twice? Right now this comment tells me we do something because of "react 18" but we went into this imagining we'd only need one act call?

defunctzombie (changes requested):
I don't follow how the removal of downsampling throttle is part of this PR. If we are using react wrong lets make a separate PR to address that before we upgrade. I'd prefer if this PR focused on the minimal diff required to go from 17->18. Anything we can and should do today with 17 we should merge in preparation for 18.

```
const throttledDownsample = useThrottle(() => downsampleDatasets, 100);
```

packages/studio-base/src/components/MessagePipeline/index.test.tsx
jtbandes: Unfortunately I don't know the reason. I could write that in the comment if that's helpful.

defunctzombie (approved):

Issue #4053, opened 2022-08-05
Cannot unset color fields


jhurliman:
**Description**

- Version: dev
- OS: All
- Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native): All
- ROS distro (if applicable): N/A

**Steps To Reproduce**

1. Create a "rgba" or "rgb" `SettingsTreeField` with an initial value of `undefined`. Observe the field correctly shows an empty color input field
2. Set the field to some color value. (This will require some code to properly re-render the settings tree with the color field value filled in.)
3. Attempt to set the color field back to its original value of `undefined`. Observe that there is no way for the user to do this.

**Expected Behavior**

It should be possible to delete the text or press an "X" clear button or some other interaction to return this value to its original state. Note that a "Reset" action menu is not sufficient since there may be several fields in this settings tree node.

**Actual Behavior**

The color input cannot be reset to `undefined`, it can only be changed to a different color.

PR #4054, opened 2022-08-05
[New3D] Color override setting for marker topics

### LINKED ISSUE #3943: [3D Beta] Override marker colors
In the current `3D` panel, I can override marker colors (both regular and mesh markers):

image

In `3D Beta`, this does not seem possible (no arrow to expand settings).

image

We should implement this for feature parity.

jhurliman:
**User-Facing Changes**

- Color can be overridden for marker topics in the `3D (Beta)` panel

**Description**

This PR introduces a concept of "render marker" vs "original marker" for marker topics, allowing a mutated marker to be created based on user settings. Currently only color override is supported, but this approach would allow for color blend modes or modifying other attributes such as scale/pose.

A quirk is also fixed: when a single marker topic contains both empty namespace markers (ns="") and one or more non-empty namespace. In that case, the empty namespace will appear as `""`.

Note that we may want to block this on #4053. Otherwise, marker color overrides cannot be unset after setting.

Fixes #3943.

Screen Shot 2022-08-05 at 10 14 14 AM


packages/studio-base/src/panels/ThreeDeeRender/renderables/TopicMarkers.ts
```diff
@@ -90,6 +90,14 @@ export class TopicMarkers extends Renderable<MarkerTopicUserData> {
     }
   }
 
+  update(): void {
```
jtbandes, line 93: Comment? It's unclear what this update() is supposed to do since it seems to just be referencing the original marker – is this only for updating transforms?

packages/studio-base/src/panels/ThreeDeeRender/renderables/markers/RenderableArrow.ts
```diff
@@ -89,8 +85,9 @@ export class RenderableArrow extends RenderableMarker {
     super.dispose();
   }
 
-  override update(marker: Marker, receiveTime: bigint | undefined): void {
-    super.update(marker, receiveTime);
+  override update(newMarker: Marker, receiveTime: bigint | undefined): void {
+    super.update(newMarker, receiveTime);
+    const marker = this.userData.marker;
```
jtbandes, line 90: when will this be different from the `newMarker`?

packages/studio-base/src/panels/ThreeDeeRender/renderables/markers/RenderableMarker.ts
```diff
@@ -89,4 +93,21 @@ export class RenderableMarker extends Renderable<MarkerUserData> {
       callback(tempTuple4, i);
     }
   }
+
+  private _renderMarker(marker: Marker): Marker {
+    const topicName = this.userData.topic;
+    const settings = this.renderer.config.topics[topicName] as
+      | Partial<LayerSettingsMarker>
+      | undefined;
```
jtbandes, line 101: Is there a way we can pass the settings in to each renderable instead of having them reach out and pull it from the renderer? Seems like this pattern (if copied around to other renderables) could lead to a lot of deeply entangled code between the renderer and renderables.

packages/studio-base/src/panels/ThreeDeeRender/renderables/markers/RenderableMarker.ts
```diff
@@ -89,4 +93,21 @@ export class RenderableMarker extends Renderable<MarkerUserData> {
...
+    const colorStr = settings?.color;
+
+    if (colorStr == undefined) {
+      return marker;
+    }
+
+    // Create a clone of the marker with the color overridden
+    const color = stringToRgba(makeRgba(), colorStr);
+    const newMarker = { ...marker, color, colors: [] };
```
jtbandes, line 110: Did you consider also just overriding the color attribute data using settings at render time? How did you settle on the approach of cloning the marker to override the color?

jtbandes (approved):
Looks fine, but raised some architectural questions for me.

packages/studio-base/src/panels/ThreeDeeRender/renderables/markers/RenderableArrow.ts
jhurliman: `this.userData.marker` is assigned in the super.update() call via `this._renderMarker(marker)`. It's not the original marker, it's the synthetic marker created from the original marker + user settings.

packages/studio-base/src/panels/ThreeDeeRender/renderables/markers/RenderableMarker.ts
jhurliman: That's a good idea, I'll look at a refactor in a followup.

packages/studio-base/src/panels/ThreeDeeRender/renderables/markers/RenderableMarker.ts
jhurliman: I've tried to avoid per-frame work wherever possible, since doing something on user input is effectively free in comparison to per-frame where we have a finite budget. I liked the flexibility of modifying any marker property using this approach (text labels could be overridden, for example).

PR #4055, opened 2022-08-05
Allow data: URLs for connect-src

jhurliman:
**User-Facing Changes**

- Unclear, let's go with "none"

**Description**

If we allow `file:`, there should be no concern with `data:`. During local development I encountered a CSP error about a `data:` URL. I wasn't able to track down the source of the fetch request, but this seems like an innocuous change.


defunctzombie (approved):
seems no worse than `blob` 🤞

PR #4060, opened 2022-08-05
Allow users to clear settings color fields

### LINKED ISSUE #4053: Cannot unset color fields

foxymiles:
**User-Facing Changes**
Adds a button to clear the value of a color field in the settings UI.

**Description**
This adds a clear button to the right of the color value.

Fixes #4053

### DIFF
--- a/packages/studio-base/src/components/SettingsTreeEditor/inputs/ColorPickerInput.tsx
+++ b/packages/studio-base/src/components/SettingsTreeEditor/inputs/ColorPickerInput.tsx
@@ -4,7 +4,7 @@
 
 import { ColorPicker } from "@fluentui/react";
 import ClearIcon from "@mui/icons-material/Clear";
-import { TextField, Popover, Button, IconButton } from "@mui/material";
+import { TextField, Popover, IconButton } from "@mui/material";
 import { useCallback, MouseEvent, useState } from "react";
 import tinycolor from "tinycolor2";
 import { makeStyles } from "tss-react/mui";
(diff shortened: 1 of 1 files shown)


jhurliman (approved):

amacneil:
The button doesn't do anything in some cases. E.g. I have 4 `x` buttons for this map layer and none of them do anything when I click them. Could we hide the button if the field is already set to the default?

image

It also feels too busy having 4 `x` icons next to each other in this example. But I think if we hide the button when the field is already set to the default, then it will be less of an issue.

foxymiles:
@amacneil At the settings level we don't have a concept of a "default" value but we could add a flag to the input to show or hide the clear button.

amacneil:
If we don't have a concept of "default", then how come I can change the min color, hit `x`, and it resets to the default?

foxymiles:
> If we don't have a concept of "default", then how come I can change the map min color, hit `x`, and it resets to the default?

It's up to the panel implementor. They'll receive an incoming value of `undefined` but can process that however they want. In the case of that panel it's reverting it to the panel's own concept of the default. But at the level of the settings ui there's no "default" value.

amacneil:
I see. I will create a ticket to discuss this then. I'm not sure how hard it is to implement, but maybe it's worth adding a declarative default (could still be `undefined` for e.g. colors that are set per-message) so that we can hide these buttons.

amacneil:
https://github.com/foxglove/studio/issues/4122

PR #4061, opened 2022-08-05
Fix style injection order

foxymiles:
**User-Facing Changes**
None

**Description**
This fixes the order of style injection so that our customizations in `makeStyles` are loaded after the stock MUI styles and take precedence.

See https://docs.tss-react.dev/readme-1


foxymiles:
cc @2metres

jtbandes (approved):

PR #4064, opened 2022-08-06
Update Chart component to support unmounting/remounting for React 18

jtbandes:
**User-Facing Changes**
None

**Description**
Split from #4052 

Updates the Chart component with changes that will make it work in React 18's Strict Mode, where effects are double-invoked (read more about this at https://github.com/reactwg/react-18/discussions/19). This means the component has to survive mount → unmount (useEffect cleanups) → remount, which is similar to what happens during hot reloading.

Changed to dynamically create a new canvas element on each mount, which was much easier than getting the OffscreenCanvas _back_ from the worker when unmounting.


defunctzombie (approved):
What's a few more refs between friends. This component is a deathtrap. 📿

PR #4066, opened 2022-08-06
[New3D] Add "Opacity" setting for PointCloud's `colormap` and `rgb` modes

siddancha:
**User-Facing Changes**
An "Opacity" topic setting now appears when using the `colormap` or `rgb` modes for PointClouds in the 3D (Beta) panel. This lets the user set the opacity (alpha) for the PointCloud.

**Description**
Transparency is very useful for point clouds, especially when the scene is cluttered with many renderables. There are 5 modes for viewing PointClouds in the 3D (Beta) panel. Previously, these were the ways that the user could control their opacity (alpha):

| *Mode* | *Could set alpha?* | *How to set alpha* |
| --- | --- | --- |
| `flat` | ✅ | Using the colorpicker. |
| `gradient` | ✅ | Using the colorpickers of the two gradient endpoints. |
| `rgba` | ✅ | From the alpha field in the PointCloud's user data. |
| `rgb` | ❌ | _Impossible to set alpha_ |
| `colormap` | ❌ | _Impossible to set alpha_ |

This PR adds an "Opacity" setting in the PointCloud's topic settings when the user is in either the `colormap` or `rgb` mode.
- "Opacity" defaults to 1 i.e. the previous behavior (fully opaque / no transparency).
- The explicit "Opacity" setting disappears when the user is not in either the `colormap` or `rgb` modes.

**Before**
image

**After**
image

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/PointCloudsAndLaserScans.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/PointCloudsAndLaserScans.ts
@@ -1306,7 +1306,7 @@ function settingsNode(
           ],
           value: colorMap,
         };
-        fields.opacity = {
+        fields.colorMapAlpha = {
           label: "Opacity",
           input: "number",
           step: 0.1,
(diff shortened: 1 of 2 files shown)


jtbandes (approved):
Thanks!

PR #4071, opened 2022-08-08
Fix selection of default timezone

foxymiles:
**User-Facing Changes**
Fixes an issue with the auto-detected timezone in preferences.

**Description**
The code was setting the selected timezone to the string `"undefined"`, including quotes, instead of removing the value from local storage.


jtbandes (approved):
How did this manifest to the user? Is there a test or story we should add?

foxymiles:
Moment timezone would treat it as missing timezone data. We don't have any interactive stories for preferences now but maybe we should add some?

foxymiles:
Added some stories for changing timezone and time format.

PR #4081, opened 2022-08-10
[New3D] fix crash when updating line strip to be empty

### LINKED ISSUE #3954: 3D (Beta) Panel error when MarkerArray points array becomes empty
**Description**
The 3D (Beta) Panel stops working and becomes un-responsive if a visualised Marker Array topic transitions from having a non-empty points array -> empty points array (Issue does not occur if marker type does not use the points array e.g. CUBE instead of LINE_STRIP as demonstrated below)

- Version:
- OS: Ubuntu 20.04
- Data source: ROS 1 Rosbag
- ROS distro: ROS Noetic

**Steps To Reproduce**
1. In a Node Playground, produce a marker array of a marker type that requires an array of points (i.e. MarkerTypes.LINE_STRIP)
2. Setup a 3D (Beta) panel that displays that marker array topic
3. In the Node Playground, set it up so the marker array list of points displayed is non-zero (>1 for viz to work)
4. Modify the Node Playground code so mid-ROSbag the list of points becomes empty [I tried to reproduce this easily by a Global Variable but that did not reproduce this issue. Best is to make the point array change live given a trigger from data in the ROSbag].
5. Play the rosbag, make sure visualisation topic is being displayed.

Here are some example pictures showing a line strip marker array (The one reproducing this issue) and another marker for reference (To note when the 3D Beta panel stops working). On the right is the input marker array msg reproducing this issue.
- Before points array becomes empty
Screenshot from 2022-07-27 12-12-24
- Points array becomes empty (Note 3D Beta panel stops working)
Screenshot from 2022-07-27 12-12-45
- If a new 3D (Beta) panel is created now, this error is shown
Screenshot from 2022-07-27 12-13-35
- Points array is non-empty again (Note 3D Beta panel has not recovered, whereas the 3D panel is still working as expected)
Screenshot from 2022-07-27 12-13-04

**Expected Behavior**
3D (Beta) panel should behave like the 3D panel and not stop working. It should flag that the marker array contains no points and thus cannot be displayed.

**Actual Behavior**
3D (Beta) panel stops working and has to be Reset. If reset while the empty points marker array msg is still selected it will display

jtbandes:
**User-Facing Changes**
Fixed a crash in the 3D (Beta) when a line strip marker was updated to have 0 points.

**Description**
Fixes #3954.

THREE.LineGeometry.setPositions does not handle an empty array: https://github.com/mrdoob/three.js/blob/38273fb9b2acde3f6046301757d9f1e5428f7d9d/examples/jsm/lines/LineGeometry.js#L15-L20

Avoid the crash by hiding the renderables and adding an early return when the new length is zero.

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/markers/RenderableLineStrip.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/markers/RenderableLineStrip.ts
@@ -66,6 +66,14 @@ export class RenderableLineStrip extends RenderableMarker {
 
   override update(marker: Marker, receiveTime: bigint | undefined): void {
     const pointsLength = marker.points.length;
+    if (pointsLength === 0) {
+      this.linePrepass.visible = false;
+      this.line.visible = false;
+      return;
+    } else {
+      this.linePrepass.visible = true;
+      this.line.visible = true;
+    }
 
     const prevMarker = this.userData.marker;
     super.update(marker, receiveTime);
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/markers/RenderableLineStrip.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/markers/RenderableLineStrip.ts
@@ -67,6 +67,8 @@ export class RenderableLineStrip extends RenderableMarker {
   override update(marker: Marker, receiveTime: bigint | undefined): void {
     const pointsLength = marker.points.length;
     if (pointsLength === 0) {
+      // THREE.LineGeometry.setPositions crashes when given an empty array:
+      // https://github.com/foxglove/studio/issues/3954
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/panels/ThreeDeeRender/renderables/markers/RenderableLineStrip.ts
```diff
@@ -66,6 +66,16 @@ export class RenderableLineStrip extends RenderableMarker {
 
   override update(marker: Marker, receiveTime: bigint | undefined): void {
     const pointsLength = marker.points.length;
+    if (pointsLength === 0) {
+      // THREE.LineGeometry.setPositions crashes when given an empty array:
```
defunctzombie, line 70: Worth opening an issue on Three.js?

defunctzombie (approved):

Issue #4075, opened 2022-08-09
One black pixel in the scrub bar for data platform


amacneil:
**Description**
<img width="194" alt="image" src="https://user-images.githubusercontent.com/637671/183559020-43c4d612-dbed-4e9f-b637-2609b3cd6bc9.png">

- Version: main
- OS: web / macos
- Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native): data platform
- ROS distro (if applicable):

PR #4087, opened 2022-08-10
Merge loaded ranges for CachingIterableSource

defunctzombie:
**User-Facing Changes**
No more erroneous single black pixel gaps in the playback scrubber when all the data is actually present and there is no gap.

**Description**
When converting the underlying cache blocks to the normalized range [0, 1], each start/end time would be converted to float and normalized. This left "gaps" between consecutive blocks (those separated by only 1 nanosecond).

This change considers consecutive blocks as part of the same range and grows the range to avoid gaps. The final float conversion happens once the merged consecutive ranges are calculated.

Fixes: #4075

### DIFF
--- a/packages/studio-base/src/players/IterablePlayer/CachingIterableSource.ts
+++ b/packages/studio-base/src/players/IterablePlayer/CachingIterableSource.ts
@@ -437,9 +437,32 @@ class CachingIterableSource implements IIterableSource {
       return;
     }
 
-    this.loadedRangesCache = this.cache.map((block) => {
-      const start = Number(toNanoSec(block.start) - sourceStartNs) / rangeNs;
-      const end = Number(toNanoSec(block.end) - sourceStartNs) / rangeNs;
+    // Merge continuous ranges (i.e. a block that starts 1 nanosecond after previous ends)
+    // This avoids float rounding errors when computing loadedRangesCache and produces
+    // continuous ranges for continuous spans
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/players/IterablePlayer/CachingIterableSource.ts
```diff
@@ -437,9 +437,32 @@ class CachingIterableSource implements IIterableSource {
...
+    }
+    if (prevRange) {
+      ranges.push(prevRange);
+    }
+
+    this.loadedRangesCache = ranges.map((item) => {
+      const start = Number(item.start - sourceStartNs) / rangeNs;
+      const end = Number(item.end - sourceStartNs) / rangeNs;
       return { start, end };
     });
```
jtbandes, line 467: Since you are already looping over the array, maybe this is an opportunity to remove the extra `map()` for a small speedup.

jtbandes (approved):

amacneil:
I can't seem to find a DP link that reproduces the original issue any more. Do you have one?

defunctzombie:
> I can't seem to find a DP link that reproduces the original issue any more. Do you have one?

You can reproduce with a local file - doesn't require the DP link. When it happens will depend on how the cache is loading. It won't be reproduced anymore because there is no more float rounding issue.

amacneil:
Cool, managed to repro (on main) with nuscenes demo. Seems to be fixed on this branch 👍

Issue #4079, opened 2022-08-09
migrate panels/log List to MUI


foxymiles:
Convert the log panel List to MUI.

Whomever tackles this make sure to test this behavior so we do not regress!
https://github.com/foxglove/studio/issues/2652

2metres:
Closed by #4094

defunctzombie:
@2metres Is this done? What I've understood from discussion is that we've only partially ported panels/log to MUI. Is there another ticket tracking the remaining work?

foxymiles:
This is the PR that converted the Log panel to fluent: https://github.com/foxglove/studio/pull/2654

PR #4092, opened 2022-08-10
Convert log panel toolbar to MUI

foxymiles:
**User-Facing Changes**
None

**Description**
This converts the log panel toolbar to MUI.

https://user-images.githubusercontent.com/93935560/183912835-9583fde9-4587-4b69-99c7-5f9555b4d532.mov

Partially addresses #4079

### DIFF
--- a/packages/studio-base/src/panels/Log/FilterBar.tsx
+++ b/packages/studio-base/src/panels/Log/FilterBar.tsx
@@ -11,13 +11,14 @@
 //   found at http://www.apache.org/licenses/LICENSE-2.0
 //   You may not use this file except in compliance with the License.
 
-import { TagPicker } from "@fluentui/react";
 import CopyAllIcon from "@mui/icons-material/CopyAll";
-import { MenuItem, Select, useTheme as useMuiTheme } from "@mui/material";
+import { MenuItem, Select, Typography } from "@mui/material";
 import { useCallback } from "react";
+import { makeStyles } from "tss-react/mui";
 
 import ToolbarIconButton from "@foxglove/studio-base/components/PanelToolbar/ToolbarIconButton";
 import Stack from "@foxglove/studio-base/components/Stack";
+import { FilterTagInput } from "@foxglove/studio-base/panels/Log/FilterTagInput";
 import useLogStyles from "@foxglove/studio-base/panels/Log/useLogStyles";
 import clipboard from "@foxglove/studio-base/util/clipboard";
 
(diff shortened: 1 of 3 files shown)


amacneil:
It is slightly larger than other toolbars:

image

Do you have a sample data source to test it with? Seems like nuscenes bag doesn't have any log messages.

foxymiles:
I don't have a bag with log messages to test. I've just been running a ros1 system via parallels.

foxymiles:
And this script to generate messages:

```python
import rospy
from std_msgs.msg import String

def talker():
    pub = rospy.Publisher('chatter', String, queue_size=10)
    rospy.init_node('talker', anonymous=True)
    rate = rospy.Rate(1) # 1hz
    while not rospy.is_shutdown():
        hello_str = "hello world %s" % rospy.get_time()
        rospy.loginfo(hello_str)
        rospy.logwarn(hello_str)
        rospy.logerr(hello_str)
        rospy.logfatal(hello_str)
        pub.publish(hello_str)
        rate.sleep()

if __name__ == '__main__':
    try:
        talker()
    except rospy.ROSInterruptException:
        pass
```

foxymiles:
> It is slightly larger than other toolbars:

Good catch. Should be fixed now.

amacneil:
lgtm. tested using this user script:

```ts

type Output = Message<"rosgraph_msgs/Log">;
export const inputs = ["/imu"];
export const output = "/log";
const people = ["alice", "bob", "caitlin"];

export default function script(ev: Input<"/imu">): Output {
  let level = 1;
  const accel = ev.message.linear_acceleration.x;

  if (accel > 0.15) {
    level = 16;
  } else if (accel > 0) {
    level = 8;
  } else if (accel > -0.5) {
    level = 4;
  } else if (accel > -1) {
    level = 2;
  }

  const person = people[Math.floor(Math.random() * people.length)];

  const out: Output = {
    header: {
      seq: 0,
      stamp: ev.message.header.stamp,
      frame_id: "",
    },
    level,
    name: "foo",
    msg: `${person}'s acceleration is ${accel.toFixed(3)}`,
    file: "",
    function: "",
    line: 0,
    topics: [],
  };

  return out;
}
```

amacneil (approved):

amacneil:
I do think the search bar in this panel is really weird. After we come up with a logical operator search syntax for console, we should backport it to this log panel.

foxymiles:
> I do think the search bar in this panel is really weird. After we come up with a logical operator search syntax for console, we should backport it to this log panel.

I agree. I think we could come up with some better UX here.

PR #4094, opened 2022-08-10
Port log panel content to MUI (except List component)

foxymiles:
**User-Facing Changes**
None

**Description**
Ports the log panel content to MUI.

Partially addresses: #4079


foxymiles:
We should look for a virtualized list component as part of porting this.

packages/studio-base/src/panels/Log/index.tsx
```diff
@@ -194,40 +210,19 @@ const LogPanel = React.memo(({ config, saveConfig }: Props) => {
...
-              const normalizedLog = normalizedLogMessage(datatype, item["message"]);
-              return (
-                <LogMessage
-                  value={normalizedLog}
-                  timestampFormat={timeFormat}
-                  timeZone={timeZone}
-                />
-              );
-            }}
-          />
```
2metres, line 221: Should we switch over to use the List component from "react-virturalized"?

foxymiles: I've tried a couple of different libraries to replace the fluent list but so far haven't found one that performs as well.

2metres:
what about https://virtuoso.dev ?

foxymiles:
> what about https://virtuoso.dev ?

Tried this and didn't get great results either.

foxymiles:
Here's another one we could try https://tanstack.com/virtual/v3

2metres (approved):

PR #4119, opened 2022-08-12
Add deprecation warning for fluentui and styled-components

2metres:
**User-Facing Changes**
None

**Description**
Add deprecation warning for fluentui and styled-components


defunctzombie (approved):
Seems 👍 - we don't really pay attention to our warnings. I'd maybe argue for this to be an error and suppress the various locations where it exists to prevent new locations from being added.

PR #4124, opened 2022-08-12
Expand state transition color palette

### LINKED ISSUE #4118: State Transitions doesn't try very hard to use unique colors
**Description**
State Transitions panel effectively uses `require("string-hash")(value) % colors.length` to determine which color corresponds to a string value. When the user gets unlucky, or purposely chooses strings with the same string-hash, the results are pretty useless.

[string_enum.mcap.zip](https://github.com/foxglove/studio/files/9312724/string_enum.mcap.zip)

**Current Experience**

image

**Desired Experience**

Unique colors (when the number of unique values is small enough that it's feasible for them to all be unique).

foxymiles:
**User-Facing Changes**
Improve state transition pane's color selection.

**Description**
This expands on the set of colors the state transition panel uses to color items to reduce the likelihood of clashes.

Before:
Screen Shot 2022-08-12 at 1 21 30 PM

After:
Screen Shot 2022-08-12 at 1 22 05 PM

It looks like our hash function does pretty well on single char values so it seems like it should be enough to just make sure we have enough colors to accommodate them:

Screen Shot 2022-08-12 at 1 23 57 PM

Fixes #4118

### DIFF
--- a/packages/studio-base/src/panels/StateTransitions/index.stories.tsx
+++ b/packages/studio-base/src/panels/StateTransitions/index.stories.tsx
@@ -17,6 +17,7 @@ import TestUtils from "react-dom/test-utils";
 import { BlockCache } from "@foxglove/studio-base/players/types";
 import PanelSetup from "@foxglove/studio-base/stories/PanelSetup";
 import { useReadySignal } from "@foxglove/studio-base/stories/ReadySignalContext";
+import { expandedLineColors } from "@foxglove/studio-base/util/plotColors";
 
 import StateTransitions from "./index";
 
@@ -75,6 +76,7 @@ const fixture = {
   ),
   topics: [
     { name: "/some/topic/with/state", datatype: "msgs/SystemState" },
+    { name: "/some/topic/with/string_state", datatype: "msgs/SystemState" },
     { name: "/blocks", datatype: "msgs/SystemState" },
   ],
   activeData: {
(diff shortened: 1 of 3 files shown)


jtbandes (approved):
I think this change is okay, although it looks like it yields some new colors that are quite similar to each other (compare `j` and `n` in your screenshot).

However, I don't think it "fixes" the linked issue. If I create a mcap file that alternates between two string values, `a` and `l` (based on your screenshot), I would only see a single color in the state transitions panel. I know that a perfect solution might introduce a performance issue, but I think it's likely that the perf would be acceptable, so it's worth trying for a complete fix.

jtbandes:
image
[string_enum-2.mcap.zip](https://github.com/foxglove/studio/files/9329430/string_enum-2.mcap.zip)

foxymiles:
The trick with basing it off blocks is that we may not have all blocks in memory so we may not know the complete set of distinct values and we need to figure out what value equality looks like for each message type in order to bucket them correctly.

jtbandes:
I guess that also means that if we wanted to choose the colors as new blocks load in, then the colors may end up being different depending on which order the blocks were loaded into memory. 😢

jtbandes:
I’m only aware of the one user report we got recently. Maybe we merge this and wait to see if there’s any user feedback or further reports?

Issue #4091, opened 2022-08-10
Sometimes some topics get "stuck" while others keep playing


amacneil:
**Description**

- Version: main
- OS: web
- Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native): sample nuscenes

Sometimes during playback, my map and markers freeze up, but the lidar keeps playing. Happens on both old3d + new3d. Maybe it is related to transform frame locking?

https://user-images.githubusercontent.com/637671/183826194-7562a6fe-ecd9-4636-92c7-c00c8a5d731f.mov

amacneil:
Here is a better video with some raw message panels shown:

https://user-images.githubusercontent.com/637671/183827313-0e57ff3c-541f-48cd-914f-f34c2ac74a9e.mov

There doesn't seem to be any obvious reason why this starts happening, but it does "fix" itself when the new tf message arrives.

Note this was on https://github.com/foxglove/studio/pull/4087 branch.

defunctzombie:
I can confirm I am also seeing a behavior here that looks like missing messages. What I am seeing is that during playback, I can seek forward up-to and past a time when I know there is a _/tf_ message yet I won't see this message in the raw message panel.

There should be messages at:

* 1532402929.697797000
* 1532402930.152601000
* 1532402930.648325000

But I can definitely play forward in a way that causes one of these to be "skipped" which makes the scene look "stuck" because the /tfs are old. The annotations are in _map_ frame but there's no update of _map_ to _base-link_ so the annotations follow along with base link when instead they should appear further behind.

amacneil:
This is still happening for me on latest main.

defunctzombie:
@amacneil I am unable to reproduce this issue on studio.foxglove.dev. What topics are getting "stuck" for you? The behavior in the original video is no longer happening for me.

amacneil:
Sample nuscenes, fresh layout with only 3D beta panel, all topics enabled.

Always, the lidar topic keeps moving, and every other topic gets stuck:

https://user-images.githubusercontent.com/637671/185271017-9a75629d-d4d8-4990-90dd-c89cbbe600dc.mov

Lmk how I can help debug.

defunctzombie:
[...]

PR #4126, opened 2022-08-14
Fix missing messages from CachingIterableSource

defunctzombie:
**User-Facing Changes**
Fixes messages drops when playing data from data sources which contain multiple messages at the exact same log time (or receive time).

**Description**
When reading from the cache, the CachingIterableSource uses a binary search to deterline where to start reading in a cache block. If a block contained multiple messages at the exact same time as the search parameter, the search algorithm would return a valid index for any time matching the search parameter not necessarily the _first_ occurrence in the block. By not returning an index for the _first_ occurrence, some messages would be skipped during playback.

This change updates `FindStartCacheItemIndex` to handle this case. If an exact match is found in the cache block for the target time, the new logic iterates the block backwards until we find the first occurrence of the target time.

Fixes: #4091

### DIFF
--- a/packages/studio-base/src/players/IterablePlayer/CachingIterableSource.ts
+++ b/packages/studio-base/src/players/IterablePlayer/CachingIterableSource.ts
@@ -52,11 +52,6 @@ type Options = {
   maxTotalSize?: number;
 };
 
-function findStartCacheItemIndex(items: [bigint, IteratorResult][], key: bigint) {
-  const idx = sortedIndexByTuple(items, key);
-  return idx < 0 ? ~idx : idx;
-}
-
 /**
  * CachingIterableSource proxies access to IIterableSource through a memory buffer.
  *
@@ -148,7 +143,10 @@ class CachingIterableSource implements IIterableSource {
 
       // We've found a block containing our readHead, try reading items from the block
       if (block) {
-        const cacheIndex = findStartCacheItemIndex(block.items, toNanoSec(readHead));
+        const cacheIndex = CachingIterableSource.FindStartCacheItemIndex(
+          block.items,
+          toNanoSec(readHead),
+        );
 
         // We have a cached item, we can consume our cache until we've read to the end
         for (let idx = cacheIndex; idx < block.items.length; ++idx) {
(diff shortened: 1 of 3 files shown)


foxymiles (approved):

packages/studio-base/src/players/IterablePlayer/CachingIterableSource.ts
```diff
@@ -304,8 +302,6 @@ class CachingIterableSource implements IIterableSource {
           block.size += pendingSizeInBytes;
         }
 
-        pendingIterResults.length = 0;
```
defunctzombie, line 307: There is no need to reset pendingIterResults here since the value is never used again.

jtbandes:
It would have been better to fix the `sortedIndexByTuple` function. This is a bug I pointed out when the function was first introduced: https://github.com/foxglove/studio/pull/3772#discussion_r927037465 

Also, the `sortedIndexByTuple` function is used in **two** places in CachingIterableSource. Does the other one need a fix too?

defunctzombie:
> Also, the `sortedIndexByTuple` function is used in **two** places in CachingIterableSource. Does the other one need a fix too?

Yes - tho in the opposite direction. I need to update that use to find the _last_ occurrence of the timestamp that matched. I actually had that code in one version for this bugfix but somehow convinced myself that was not important since any message at that time would be valid to consider as the "last" message - however upon further thought I actually do want the last occurrence so I can make best use of the cache.

> It would have been better/cleaner to fix the `sortedIndexByTuple` function. This is a bug I pointed out when the function was first introduced: [#3772 (comment)](https://github.com/foxglove/studio/pull/3772#discussion_r927037465)

I decided to leave that function alone and instead linearly iterate from the found location. The alternative is to have two functions - one to find the first occurrence and one for the last (see above). When I thought about the tradeoff of continuing to do a binary search for first occurrence vs linear scan forwards or backwards - it seemed to me like a linear scan would yield the result quicker.

jtbandes:
> it seemed to me like a linear scan would yield the result quicker.

Depends on the data. Imagine all items are the same — you have to traverse N/2 items linearly.

jtbandes:
> The alternative is to have two functions - one to find the first occurrence and one for the last

In C++ these are called `upper_bound` and `lower_bound`.

defunctzombie:
https://github.com/foxglove/studio/pull/4144

PR #4127, opened 2022-08-15
Run rosbridge topic graph setup queries in parallel

### LINKED ISSUE #4123: Long time to establish data source connection
**Description**
When loading Foxglove Studio on our robots it takes an average of 1 minute 40 seconds to establish a data source connection and we think this relates to the rosbridge player topic initialization.

Screenshot from 2022-08-12 13-16-50

When reverting [this change](https://github.com/foxglove/studio/pull/3185/files#diff-502298948b3339b723b05e46807ad8b121945b7753732c899253f8aa37aceb6b) we noticed that the data source connection is established almost immediately. The PR notes that this change fixes an issue with the topic graph panel, but we don't find this feature useful. A typical environment for us consists of ~160 nodes, 980 topics, and 620 services.

Is there a workaround or another solution you could recommend?

- Version: 1.20.1 (also tested on 1.18.0)
- OS: Ubuntu
- Data source: rosbridge
- ROS distro (if applicable):

**Steps To Reproduce**
Load a production build of Foxglove Studio in an environment with a large number of nodes, topics, and services. 

**Expected Behavior**
The data source connection is established in a reasonable time.

**Actual Behavior**
The data source connection takes approximately 1 minute 40 seconds to establish. 

@jasonimercer

foxymiles:
**User-Facing Changes**
Improves startup time of rosbridge sessions with large topic graphs.

**Description**
This runs the topic graph queries in parallel to speed them up.

Fixes #4123

### DIFF
--- a/packages/studio-base/src/players/RosbridgePlayer.ts
+++ b/packages/studio-base/src/players/RosbridgePlayer.ts
@@ -11,7 +11,7 @@
 //   found at http://www.apache.org/licenses/LICENSE-2.0
 //   You may not use this file except in compliance with the License.
 
-import { isEqual, sortBy } from "lodash";
+import { isEqual, sortBy, transform } from "lodash";
 import roslib from "roslib";
 import { v4 as uuidv4 } from "uuid";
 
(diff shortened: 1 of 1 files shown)


defunctzombie (approved):
Might be easier to read if the Promise.then chain was in an async function and you used `async` instead - I know you want to wrap the whole thing in a finally but you can do that using a `try/finally` in the async function.

PR #4131, opened 2022-08-15
Update ImageCanvas to support React 18

jtbandes:
**User-Facing Changes**
None

**Description**
Updates to Image panel to make it run under React 18 Strict Mode, where the component will run a mount->unmount->mount cycle when it first mounts (read more about this at https://github.com/reactwg/react-18/discussions/19). Similar to #4064, we dynamically create the canvas element on each mount so we don't have to worry about calling `transferControlToOffscreen()` more than once on the same canvas.


defunctzombie (approved):
👍 Does feel like there's an opportunity here to cleanup some state and effect logic into a reducer or some simpler hook separate from the component but maybe something for a different cleanup.

PR #4134, opened 2022-08-15
Upgrade typescript

defunctzombie:
**User-Facing Changes**
None

**Description**
Upgrade typescript to 3.7.4 to stay current.


defunctzombie:
@jtbandes there's a failing build issue in the three-text package. I'm hoping you can take a look or provide some context on what an appropriate fix is.

```
Error: packages/three-text/src/LabelPool.ts(401,5): error TS2741: Property 'colorSpace' is missing in type '{ data: Uint8ClampedArray; width: number; height: number; }' but required in type 'ImageData'.
```

jtbandes (approved):

PR #4139, opened 2022-08-16
Disable context menu for the entire app

### LINKED ISSUE #3961: Disable browser context menu for entire application
It gets weird when I have a context menu open and the browser context menu happens over top of it.

foxymiles:
**User-Facing Changes**
Disables the right click context menu for the entire app.

**Description**
This just catches the `onContextMenu` event of the root workspace element.

Fixes #3961

### DIFF
--- a/packages/studio-base/src/App.tsx
+++ b/packages/studio-base/src/App.tsx
@@ -2,7 +2,7 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
-import { useState, Suspense, Fragment } from "react";
+import { useState, Suspense, Fragment, useEffect } from "react";
 import { DndProvider } from "react-dnd";
 import { HTML5Backend } from "react-dnd-html5-backend";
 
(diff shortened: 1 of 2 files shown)


jtbandes:
I don't love this change because it also breaks the normal, expected behavior of right-clicking in a text field to copy or paste. Unfortunately I doubt there is a way to disable some of the internal context menus but leave the ability to copy/paste by context menu.

foxymiles:
@jtbandes I've adjusted the logic to allow right click on inputs and textareas.

jtbandes:
Thanks, that is nicer. I'm still able to right-click on modal dialogs on this build, probably because they are outside Workspace.

defunctzombie:
> Thanks, that is nicer. I'm still able to right-click on modal dialogs on this build, probably because they are outside Workspace.

We might need to provide a specific target element for mui modals - not even sure if we can do that globally for all mui modals.

foxymiles:
I moved the handler to the `App` component and installed it at the `document` level. This seems to catch modals too.

jtbandes:
Next complaint: it's annoying not being able to right click on a link to open it in a new tab (at least in the web build).

jtbandes (approved):
Pretty usable though 👍

foxymiles:
> Next complaint: it's annoying not being able to right click on a link to open it in a new tab (at least in the web build).

I guess we could add that back with our own context menu? Or we could show the native one but it seemed like the goal of the original ticket was to avoid that as much as possible.

amacneil:
👍 

The next step after this is to start adding back our own context menu where it makes sense. https://vscode.dev/ is a good example of this done well imo (browser context menu is disabled globally, but most things are right-clickable).

Issue #3990, opened 2022-07-29
[quick feature] 


ljburtz:
feature 1. plot setting to lock the x and y axis scales (aspect ratio) -> Important for "map" type applications
(when chossing the x axis is "msg path (accumulated)" for example
a square robot trajectory should be a square/not a rectangle! 

feature 2. fix xmin and xmax as well as ymin and ymax (important for consistent/comparable plots)

simple additional settings in the 'plot panel settings' would address this
![image](https://user-images.githubusercontent.com/26211495/181675171-ac966812-106c-43e3-bb93-51fa10d5b7a5.png)

amacneil:
@ljburtz let's keep each feature request in a separate ticket. I split the xmin/xmax request into https://github.com/foxglove/studio/issues/3991

PR #4145, opened 2022-08-16
Add manual limits for x axis

### LINKED ISSUE #3991: Plot: Option to set "x min" and "x max"
Ability to set "x min" and "x max" in addition to "y min"m and "y max" (already supported).

This is important for consistent/comparable plots.

It could be added to the plot settings:

image

Extracted from https://github.com/foxglove/studio/issues/3990 (@ljburtz fyi)

### LINKED ISSUE #4251: Plot with auto x range is very narrow
- Version: 1.23.0
- OS: macOS
- Data source: ROS1 bag
- ROS distro: Melodic

I am using `rosbag record` with a split option that produces an hourly file. The system runs for a long time, potentially for weeks. When I load one of these slices, a Plot panel that plots a value vs `header.stamp` with auto *x* range looks like this, with all the data squeezed to the right:

image

I assume it's setting the minimum *x* limit to the entire lifetime of the ROS master, and not to just what this bag contains. Is there a better solution?

(Bad title, sorry, feel free to edit.)

foxymiles:
**User-Facing Changes**
This adds new plot panel settings to manually specify the min and max X axis values.

**Description**
This adds new settings and new logic to clamp the X axis when both min and max are specified.

It also separates X & Y related settings into their own nodes in the settings sidebar.

Screen Shot 2022-08-16 at 3 07 33 PM

Fixes #3991
Fixes #4251

### DIFF
--- a/packages/studio-base/src/components/TimeBasedChart/index.tsx
+++ b/packages/studio-base/src/components/TimeBasedChart/index.tsx
@@ -477,18 +477,10 @@ export default function TimeBasedChart(props: Props): JSX.Element {
       max = datasetBounds.x.max;
     }
 
-    // if we are syncing and have global bounds there are two possibilities
-    // 1. the global bounds are from user interaction, we use that unconditionally
-    // 2. the global bounds are min/max with our dataset bounds
-    if (globalBounds) {
-      if (globalBounds.userInteraction) {
-        min = globalBounds.min;
-        max = globalBounds.max;
-      } else if (defaultView?.type === "following") {
-        // If following and no user interaction - min/max with globalBounds.
-        min = Math.min(min ?? globalBounds.min, globalBounds.min);
-        max = Math.max(max ?? globalBounds.max, globalBounds.max);
-      }
+    // If the global bounds are from user interaction, we use that unconditionally.
+    if (globalBounds?.userInteraction === true) {
+      min = globalBounds.min;
+      max = globalBounds.max;
     }
 
     // if the min/max are the same, use undefined to fall-back to chart component auto-scales
--- a/packages/studio-base/src/components/TimeBasedChart/index.tsx
+++ b/packages/studio-base/src/components/TimeBasedChart/index.tsx
@@ -484,7 +484,7 @@ export default function TimeBasedChart(props: Props): JSX.Element {
       if (globalBounds.userInteraction) {
         min = globalBounds.min;
         max = globalBounds.max;
-      } else if (defaultView?.type !== "following") {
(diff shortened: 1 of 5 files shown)


foxymiles:
`Range` here for X axis seems a bit ambiguous. Maybe we should call it `Follow range`?

amacneil:
Looks nice. @ljburtz FYI this was your request.

X axis settings currently seem pretty broken in this PR:

https://user-images.githubusercontent.com/637671/185045022-2f6592d2-5f07-4c90-a9be-861e70896928.mov

amacneil:
@foxymiles 

> `Range` here for X axis seems a bit ambiguous. Maybe we should call it `Follow range`?

It feels to me like maybe there are two modes - I don't know what to call them, but let's say "live" and "static".

Live mode is when x axis represents time, you are following the playhead, and you set a `range` on how much time you want to see prior to your playhead.
- This should be the default for live data sources

Static mode (needs a better name) is when you see the entire window of time, and can set a min/max x axis values.
- This should be the default for historical data sources (is that even possible to change the default based on data source?)
- If you are using a live data source, this option is still available, e.g. if you want to accumulate all data since time=0.

This needs more thinking through, but I just wanted to put some thoughts out there.

foxymiles:
> X axis settings currently seem pretty broken in this PR:

I've fixed the toggle button and changed the logic to apply the min max if at least one of them is specified instead of requiring both.

defunctzombie:
How does the `Range` setting interact with `Min/Max` for the x-axis? It feels to me like we need a dropdown that lets you select whether to use the `min/max` or the offset from current options to control time on the x-axis.

foxymiles:
> How does the `Range` setting interact with `Min/Max` for the x-axis? It feels to me like we need a dropdown that lets you select whether to use the `min/max` or the offset from current options to control time on the x-axis.

Currently setting one clears the other but yeah it might be better to make it explicit. I guess that will mean dirtying the layout if we add a new field to control that setting?

ljburtz:
great to see the work happening for this feature, much appreciated.

to contribute to the thinking -> an important use case related to the xmin / xmax is when plotting using msg.accumulated as the x axis index (= sometimes the x axis is not time). might be good to test this use case during this xmin/xmax development.

also: these two use cases are relevant #3989 and #3990

foxymiles:
> The Range setting didn't have any effect on my plot. This is our sample dataset.

Looks like we don't have a story covering that setup. I'll add one.

defunctzombie (approved):

PR #4146, opened 2022-08-16
render order being disabled causing transparency fighting

### LINKED ISSUE #4082: Map layer opacity affects objects in front of it
**Description**
When I adjust the opacity of my map layer, it affects the visibility of objects in front of the map.

https://user-images.githubusercontent.com/637671/183789037-f47c198c-e8d2-4133-a862-9b3f88c79372.mov

- Version: main
- OS: web / macos
- Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native): sample nuscenes

snosenzo:
Fixes #4082 - Enable threejs object sorting

**User-Facing Changes**
 - Transparent object ordering in the new 3D view has been fixed so that their colors can be properly stacked in order away from the camera
 
**Description**
 - enable sortObjects on the threejs renderer
 - doesn't seem to affect the render order or lines and text markers/labels

https://github.com/foxglove/studio/issues/4082

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
@@ -295,7 +295,7 @@ export class Renderer extends EventEmitter<RendererEvents> {
     this.gl.info.autoReset = false;
     this.gl.shadowMap.enabled = false;
     this.gl.shadowMap.type = THREE.VSMShadowMap;
-    this.gl.sortObjects = false;
+    this.gl.sortObjects = true;
     this.gl.setPixelRatio(window.devicePixelRatio);
 
     let width = canvas.width;


defunctzombie:
Is there a story we can make to surface this regressing in the future?

I imagine this impacts performance somewhere in the pipeline. Is that something we will notice?

snosenzo:
Yeah I can try to make a story that captures this so that we don't regress
Yeah it looks like it does have a slightly significant performance impact: the average latency on the each frame appears to have risen by ~50% or more (from , but it's still performant. I think this is something we probably need to be doing anyway how many transparent objects we have. This also muddies the color of the annotation boxes when they're overlapping with each other because they have random order otherwise. 
these are the stats graphs:
my change:
image

master: 
image

I can do a full google perf monitor session if you want more info too

defunctzombie:
@snosenzo Are those benchmark runs both on the production build?

jtbandes (approved):

snosenzo:
I used my local build for the first one and the second I just used studio.foxglove.dev. Is that what I should be using?

defunctzombie:
> I used my local build for the first one and the second I just used studio.foxglove.dev. Is that what I should be using?

Some notes on the various builds involved:

`yarn web:serve` produces a "debug" build (as much as there can be in web). This is non-minified and uses the debug build of react (which does additional checks at runtime).

`yarn build:prod` will produce a "production" build locally (minified and env is set to production for react).

`studio.foxglove.dev` runs a "production" build.

Each PR has a "preview" deployment with a "production" build:
image

----

production builds will definitely produce different perf results from dev builds. So it is useful to run the perf check against equal "types" of builds. So when you say "local build" that could be production or dev depending on what commands you ran.

snosenzo:
Thanks for that explanation! Good to know going forward.
I tested against the prod vercel build of this PR and the performance seems to be nearly identical before and after the fix.

I'm looking into understanding storybook a bit more now, but it might take me a bit to get my head around making a test case for this. Do you want me to do the regression story as part of this PR or a subsequent PR?

amacneil:
There is also this thing John created: https://github.com/foxglove/studio/tree/main/benchmark

I don't know how to use it, but it seems like this is a perfect use case.

defunctzombie:
> I'm looking into understanding storybook a bit more now, but it might take me a bit to get my head around making a test case for this. Do you want me to do the regression story as part of this PR or a subsequent PR?

@jtbandes is this change something we can storybook with our existing setup or outside the scope of any of our existing stories? If it is outside the scope I think we can move along with this bugfix as is.

snosenzo:
@amacneil thanks for the suggestion! looks like the benchmarks are pretty much identical from main this morning to my branch.
main: 
[players/BenchmarkPlayer.ts] Frame time (filtered) average: 16.89984374998603, median: 16.69999998807907, P90: 17.200000002980232
[players/BenchmarkPlayer.ts] Frame time (filtered) average: 16.91374999997206, median: 16.69999998807907, P90: 17.200000002980232
[players/BenchmarkPlayer.ts] Frame time (filtered) average: 16.88328125004191, median: 16.700000002980232, P90: 17.299999997019768
this branch:
[players/BenchmarkPlayer.ts] Frame time (filtered) average: 16.836875000060537, median: 16.69999998807907, P90: 17.100000008940697
[players/BenchmarkPlayer.ts] Frame time (filtered) average: 16.887812500074507, median: 16.69999998807907, P90: 17.200000002980232
[players/BenchmarkPlayer.ts] Frame time (filtered) average: 16.879218749958092, median: 16.700000002980232, P90: 17.200000002980232

snosenzo:
@defunctzombie   @jtbandes  brought up a good point last night to me that something like this will be hard to make a story for because the render order with `sortObjects = false` is unspecified, it would be hard to guarantee a regression failure with a story

defunctzombie:
:shipit:

PR #4149, opened 2022-08-16
Hide Safari chrome when launched from Home screen on iOS

rgov:
**User-Facing Changes**

When the user saves the web-based Studio to the Home Screen, this causes the shortcut to launch as a separate "app" without Safari chrome, removing the address bar and the user's other tabs for a cleaner look.

**Description**

See here: https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/ConfiguringWebApplications/ConfiguringWebApplications.html#//apple_ref/doc/uid/TP40002051-CH3-SW2

To take this further you could specify the status bar color.

```
<meta name="apple-mobile-web-app-status-bar-style" content="black">
```

I understand Safari and touch-based devices are not officially supported but this would be a nicety for users who are experimenting with Studio as, for example, a read-only dashboard.

### DIFF
--- a/web/webpack.config.ts
+++ b/web/webpack.config.ts
@@ -112,6 +112,7 @@ const mainConfig = (env: unknown, argv: WebpackArgv): Configuration => {
   <html>
     <head>
       <meta charset="utf-8">
+      <meta name="apple-mobile-web-app-capable" content="yes">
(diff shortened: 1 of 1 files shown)


defunctzombie (approved):

jtbandes (approved):

jtbandes:
Thanks!

Issue #2834, opened 2022-02-13
Publish pose does not work when connected to a ROS2 stack via rosbridge


samiamlabs:
Hi folks!

I noticed that the tools for publishing poses and points do not work for ROS2.
![image](https://user-images.githubusercontent.com/5682888/153764250-5e248ef7-7c1b-4f34-86aa-d2f9d1bd38d6.png)

The issue seems to be that the header fields in PoseStamped/PoinsStamped differ somewhat between the versions of ROS.

I get this error from rosbridge when I try to publish a pose:
```bash
[rosbridge_websocket-2] [ERROR] [1644769470.882365620] [rosbridge_websocket]: [Client 3] [id: publish:/goal_pose:114] publish: Message type geometry_msgs/PoseStamped does not have a field header.seq
[rosbridge_websocket-2] [ERROR] [1644769515.979927426] [rosbridge_websocket]: [Client 3] [id: publish:/clicked_point:125] publish: Message type geometry_msgs/PointStamped does not have a field header.seq
``` 

It would be nice to have an option to send ROS2 compatible messages with the tools.

jtbandes:
Yes, we have a separate ticket for ROS 2 native publishing: https://github.com/foxglove/studio/issues/1661

However this issue is largely about the incompatibility between ROS 1 & 2 versions of PoseStamped, namely that header.seq was removed in ROS 2.

karanchahal:
we have this problem as well, would be great to see support for ROS2 message types on foxglove

jtbandes:
Yes, I suppose the player could expose some info about whether it is ros1/ros2/other, and the panel could use that to automatically select the message format. I'm not sure what we should do for non-ROS players to be able to use this feature in the future, though (related: https://github.com/foxglove/ws-protocol/issues/38). I guess another option is that the panel could send the message to the player in a common format (like our [Foxglove message types](https://foxglove.dev/docs/studio/messages/introduction)), and the player could take care of translating it to whatever format is required for publishing.

Thoughts @defunctzombie?

defunctzombie:
> Yes, I suppose the player could expose some info about whether it is ros1/ros2/other, and the panel could use that to automatically select the message format.

Maybe this could be the "profile" concept that we have with mcap headers.

> I'm not sure what we should do for non-ROS players to be able to use this feature in the future, though

I think we would document the 3d panel as publishing a `foxglove.Pose` (or appropriately named message) for any "custom" framework (or profile).

[...]

PR #4151, opened 2022-08-16
Remove `seq` from ThreeDeeRender and ThreeDimensionalViz publish messages

defunctzombie:
**User-Facing Changes**
Publishing via the 3d and 3d (beta) panel works for ROS2 data sources if the connection supports publishing (rosbridge supports publishing to both ros1 and ros2).

**Description**
When publishing messages to a ROS 2 source (rosbridge), the `seq` field would cause the rosbridge to fail because headers in ROS2 do not have a `seq` field.

This change removes the `seq` field from ThreeDeeRender and ThreeDimensionalViz so those messages are valid to publish.

Fixes: #2834


defunctzombie:
I tested this manually against ros1 rosbridge and ros2 rosbridge. I also tested this against `ros1 native` in the desktop app and that continues to work even without the `seq` field.

jtbandes (approved):

PR #4157, opened 2022-08-17
[New3D] Update settings sidebar when toggling 2D/3D mode via panel

jtbandes:
**User-Facing Changes**
Fixed a bug in the 3D (Beta) panel where the settings sidebar would not update when toggling between 2D/3D mode via the panel button or keyboard shortcut.
Fixed an issue with numeric step buttons in 3D (Beta) panel settings not working for Phi, Theta, and Y-Axis FOV.

**Description**
For performance reasons, the New3D panel does not call `updatePanelSettingsEditor` any time its configuration changes, but only when the panel deems necessary (usually in a settings action handler, or when a scene extension notices a change in the underlying data source, after a debounced waiting period). The 2D/3D toggle button & keyboard shortcut were implemented via directly changing the config, and the config change triggers a re-render via useEffect, but that did not trigger an update of the settings tree. (The CoreSettings object updates its settings subtree when it detects a cameraMove event from the panel, but in this case the change did not come from the panel.) The fix is to manually tell the CoreSettings object to update the tree after the setting is toggled.

The `ONE_DEGREE` -> `1` change fixes a regression from https://github.com/foxglove/studio/pull/4043 where we forgot to update the step values in settings tree when the values changed from radians to degrees.

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
+++ b/packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
@@ -463,7 +463,8 @@ export function ThreeDeeRender({ context }: { context: PanelExtensionContext }):
     });
   }, [actionHandler, context, settingsTree]);
 
-  // Update the renderer's reference to `config` when it changes
+  // Update the renderer's reference to `config` when it changes. Note that this does *not*
+  // automatically update the settings tree.
   useEffect(() => {
     if (renderer) {
       renderer.config = config;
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
```diff
@@ -793,7 +794,8 @@ export function ThreeDeeRender({ context }: { context: PanelExtensionContext }):
       ...prevConfig,
       cameraState: { ...prevConfig.cameraState, perspective: !prevConfig.cameraState.perspective },
     }));
-  }, []);
+    setTimeout(() => renderer?.updateCoreSettings(), 0);
```
foxymiles, line 797: I'm assuming this doesn't work without the timeout?

defunctzombie: Relatedly - should this be `setImmediate` ?

This should get a comment since it does look a bit like a gotcha.

jtbandes: I tried queueMicrotask and that didn't work. We need to wait more for the setConfig to actually take effect.

foxymiles (approved):

PR #4164, opened 2022-08-17
Enable and fix `@typescript-eslint/explicit-member-accessibility`

jtbandes:
**User-Facing Changes**
None

**Description**
Rationale: With `public` being the default accessibility level, TypeScript makes it easy to _accidentally_ expose methods and members that were intended to be internal to other files/classes. In a large codebase this can lead to unintentional entangling of components over time. By making an explicit accessibility level required, we encourage developers to think about the accessibility level of each method/field, which will hopefully steer them towards more `private` or `#`-private methods/fields.


defunctzombie:
I don't know what to do about this PR. I'd be more inclined to 👍 if you could point out an area where having this would have prevented us from making a mistake or helped us catch a bug sooner.

jtbandes:
I would say that public-vs-private is not so much about _mistakes_, but about leaky abstractions. It is probably rare to cause a bug by accidentally accessing a variable that was intended to be private, but it is easy to be lazy about it and slowly make the architecture worse over time.

We could also try marking as many of these `private` as possible, but that would be a bit more work up front.

jtbandes:
PR has been updated to mark things as `private` where possible. The things left `public` are actually being used.

defunctzombie (approved):

Issue #3899, opened 2022-07-21
Manual looping is broken in experimental player


amacneil:
**Description**
1. Enable experimental player
2. Load sample dataset
3. Ensure "loop" is disabled
3. Play to the end of file
4. Hit "play" again

Playback does not loop back to the start, it gets stuck at the end.

Expected behavior: When experimental player is disabled, playback will pause at the end of the file, and hitting play will restart playback from the beginning.

- Version: main
- OS: web
- Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native): sample data
- ROS distro (if applicable): n/a

https://user-images.githubusercontent.com/637671/180128660-d5e9e722-c330-4ca4-89d2-63b05e4e9c59.mov

amacneil:
@defunctzombie this is broken again

PR #4165, opened 2022-08-17
Fix manual looping when using the latching player.

defunctzombie:
**User-Facing Changes**
Playing a file when already at the end restarts playback at the start.

When a file reaches the end of its playback the "pause" icon changes back to a "play" icon to indicate that the player is paused.

**Description**
PR https://github.com/foxglove/studio/pull/4045 removed the resumePlay function which was responsible for seeking to the start and playing when the user wanted to continue playing from the end. This change adds that functionality back so when a user wants to "play" and playback is at the end, we seek to the start.

Unfortunately, this logic was being intercepted by logic in RepeatAdapter which would immediately try to pause playback when repeat was off and `isPlaying` was still true. This change updates the IterablePlayer to indicate it is no longer playing when it has reached the end and updates the RepeatAdapter to remove the `isPlaying` case since the player should be responsible for that condition.

Fixes: #3899


defunctzombie:
I could also be convinced that we need to add a new method to message pipeline (toggle repeat) or something to that effect and move the selector logic to message pipeline.

packages/studio-base/src/components/PlaybackControls/index.tsx
```diff
@@ -145,15 +153,34 @@ export default function PlaybackControls(props: {
     [seekBackwardAction, seekForwardAction, togglePlayPause],
   );
 
+  const repeatSelector = useCallback(
```
foxymiles, line 156: What's the advantage of putting the `seek` and `play` in the selector instead of just selecting values and acting on them in a separate `useEffect`?

defunctzombie: I could remove them from the selector and use the values from the `ctx`. I used the ones passed in via props since they were available.

Acting on them in a useEffect was what the adapter did and I found that somewhat strange because react doesn't need to be involved in this. This is a message pipeline action and it is something that can happen on the message pipeline context.

jtbandes (approved):

packages/studio-base/src/components/PlaybackControls/index.tsx
defunctzombie: Fixed this up per discussion in slack.

Issue #4084, opened 2022-08-10
"User Script did not produce a message" but no error details


jtbandes:
**Description**

- Version: latest web
- OS: Chrome / macos
- Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native): https://librealsense.intel.com/rs-tests/TestData/stairs.bag

**Steps To Reproduce**
- Load this data source: https://librealsense.intel.com/rs-tests/TestData/stairs.bag
- Add a user script with this code:
```ts
import { Input, Message } from "./types";
type Output = Message<"foxglove_msgs/ImageAnnotations">;
type PointsAnnotationMsgType = Message<"foxglove_msgs/PointsAnnotation">;

export const inputs = ["/device_0/sensor_1/Color_0/image/data"];
export const output = "/studio_script/point_annotations";

export default function node(
  event: Input<"/device_0/sensor_1/Color_0/image/data">
): Output {
  const out = {} as Output;
  out.points = new Array();
  const line_points = {} as PointsAnnotationMsgType;
  line_points.points = new Array();
  line_points.outline_colors = new Array();
  (line_points as any).outline_color = { r: 1, g: 1, b: 1, a: 1 };
  let x0: number = 0;
  let x1: number = 1920;
  let y0: number = 500;
  let step: number = 10;
  let i: number = 0;
  while (x0 + i * step <= x1) {
    let xi: number = i * step;
    line_points.points.push({ x: xi, y: y0 });
    i++;
  }
  line_points.type = 3;
  line_points.timestamp = event.message.header.stamp;
  if (line_points.points.length % 2 !== 0) {
    line_points.points.length -= 1;
    line_points.outline_colors.length -= 1;
  }
  out.points.push(line_points);
  return out;
}
```

**Expected Behavior**
Some error should show me why my user script didn't produce a message.

**Actual Behavior**
<img width="741" alt="image" src="https://user-images.githubusercontent.com/14237/183792003-381a579c-2ba4-4a51-aad6-ba6284406b65.png">

jtbandes:
You also have to enable the annotation topic in an Image panel:

<img width="387" alt="image" src="https://user-images.githubusercontent.com/14237/185223359-27b18f13-3b1f-4120-8c85-7367941d4eb6.png">

defunctzombie:
<img width="506" alt="image" src="https://user-images.githubusercontent.com/84792/185227378-a20e24af-acf5-454a-85b0-fb09e7192d50.png">

[...]

PR #4178, opened 2022-08-18
Improve error message when user script errors during runtime

defunctzombie:
**User-Facing Changes**
When a user script errors, the new message displayed to the user is the following:

Screen Shot 2022-08-17 at 6 09 36 PM

Additionally, the `Problems` tab diagnostics list is fixed to removed the `undefined` text that would show up after `Runtime`.

Screen Shot 2022-08-17 at 5 45 12 PM

**Description**
When a user script errors during runtime and fails to produce a message, the previous error displayed would only indicate that a script failed to produce a message. This warning did not make it obvious that the user script in fact errored during runtime.

This change improves the error to be clear that the script errored during runtime when there is an error rather than only a missing message.

Fixes: #4084

### DIFF
--- a/packages/studio-base/src/players/UserNodePlayer/index.ts
+++ b/packages/studio-base/src/players/UserNodePlayer/index.ts
@@ -485,26 +485,34 @@ export default class UserNodePlayer implements Player {
           return;
         }
 
-        const diagnostics =
-          result.error != undefined
-            ? [
-                {
-                  source: Sources.Runtime,
-                  severity: DiagnosticSeverity.Error,
-                  message: result.error,
-                  code: ErrorCodes.RUNTIME,
-                },
-              ]
-            : [];
-        if (diagnostics.length > 0) {
-          this._setUserNodeDiagnostics(nodeId, diagnostics);
+        const allDiagnostics = result.userNodeDiagnostics;
+        if (result.error) {
+          allDiagnostics.push({
+            source: Sources.Runtime,
+            severity: DiagnosticSeverity.Error,
+            message: result.error,
+            code: ErrorCodes.RUNTIME,
+          });
         }
+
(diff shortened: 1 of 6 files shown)


amacneil:
Screenshot looks ok for now. One day we can figure out a better place to expose playback errors like this.

defunctzombie:
> Screenshot looks ok for now. One day we can figure out a better place to expose playback errors like this.

We could add dots on the timeline if we want - just seems generally more invasive /shrug

packages/studio-base/src/players/UserNodePlayer/index.ts
```diff
@@ -485,26 +485,34 @@ export default class UserNodePlayer implements Player {
...
+            code: ErrorCodes.RUNTIME,
+          });
         }
+
         this._addUserNodeLogs(nodeId, result.userNodeLogs);
 
+        if (allDiagnostics.length > 0) {
+          this._problemStore.set(problemKey, {
+            severity: "error",
+            message: `User Script ${nodeId} suffered an error.`,
```
jtbandes, line 503:
```suggestion
            message: `User Script ${nodeId} encountered an error.`,
```

jtbandes (approved):

packages/studio-base/src/players/UserNodePlayer/index.ts
defunctzombie: I accepted this change but I do feel like suffered is better :D encountered is like the script ran into the error but the error is something that was along the scripts journey. Suffered is more like the script itself became ill with the error 🤣

defunctzombie:
> I think we should get rid of "No details provided". It doesn't help add any context for the user.

Sure - I can make that change in the dialog. If there is a `tip` but no `error` I can remove the no details provided. I think it is a useful fallback when there is no tip _and_ no error.

Issue #4168, opened 2022-08-17
Help Button in Data Source Panel links to 404


snosenzo:
**Description**
Help Button in Data Source Panel links to 404
https://foxglove.dev/docs/studio/app-concepts/data-sources

- Version:
- OS: web/macos
- Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native): n/a
- ROS distro (if applicable): n/a

**Steps To Reproduce**

 - explore sample data
 - open data source panel
 - click help button in top right of panel

**Expected Behavior**
 - maybe link to this page? https://foxglove.dev/docs/studio/panels/data-source-info

**Actual Behavior**
 - links to this page and it's a 404: https://foxglove.dev/docs/studio/app-concepts/data-sources

amacneil:
This was the intended destination: https://foxglove.dev/docs/studio/connection/data-sources

@esthersweon can you add a redirect on the website repo too

PR #4179, opened 2022-08-18
Fix outdated links to docs

defunctzombie:
**User-Facing Changes**
Clicking the updated links takes users to the correct docs pages.

**Description**
Fixes: #4168


defunctzombie:
I'd personally petition to remove all the `?` icons we've littered on the sidebars as they are not relevant to my day-to-day use of the app.

jtbandes (approved):
I thought we generally tried to keep redirects for old links when we move pages around on the website?

PR #4181, opened 2022-08-18
Dark Sidebar Icon Strip on light mode

2metres:
**User-Facing Changes**
Screen Shot 2022-08-18 at 8 29 33 pm

**Description**
Change to dark sidebar strip on light mode a la VS Code. @amacneil


defunctzombie:
Is it a light strip in dark mode? What's the thinking behind this? Is this a common pattern across apps or just a thing vscode does?

jtbandes:
Meh – I personally feel that the dark strip sticks out too much visually. Just because VS Code does it doesn't mean it looks nice.

jtbandes:
Perhaps we could explore custom themes like Slack:

image


image

2metres:
Sounds like the consensus from conversations today is NO. Closing for further consideration

Issue #4142, opened 2022-08-16
No hover state on sidebar buttons


jtbandes:
Regression from https://github.com/foxglove/studio/pull/4024

<table><tr><th>Before</th><th>After</th></tr>
<tr>
<td>

https://user-images.githubusercontent.com/14237/184940341-5e36deba-2229-4a65-af06-7f041148e22d.mov

</td><td>

https://user-images.githubusercontent.com/14237/184940366-6de2470b-7c0a-4caf-990e-ff64f91f65ca.mov

</td></table>

defunctzombie:
@2metres should we have a hover state?

defunctzombie:
Marking as a _bug_ since it is a regression - we can separately determine whether we want this behavior.

PR #4185, opened 2022-08-18
Change tab style to have opacity 1 when hovering

defunctzombie:
**User-Facing Changes**
Visual feedback when hovering a tab.

**Description**
To provide visual indicator to a user when hovering a tab, set the opacity on hover to 1.

Fix a bug with setting the selected item's opacity to 1. Previously it was 0.8 because the style override was not correctly matching the needed css rule.

Fixes: #4142

### DIFF
--- a/packages/studio-base/src/theme/muiComponents.ts
+++ b/packages/studio-base/src/theme/muiComponents.ts
@@ -425,8 +425,13 @@ export default function muiComponents(theme: Theme): ThemeOptions["components"] 
         root: {
           opacity: 0.8,
 
-          "&$selected": {
+          "&.Mui-selected": {
             opacity: 1,
+          },
+
+          "&:not(.Mui-selected):hover": {
+            opacity: 1,
+            color: theme.palette.action.active,
           },
         },
         selected: {},


defunctzombie:
@2metres One thing I've not figured out in this change is how to keep the `selected` icon color purple when I hover over it. Right now any hovering changes the color to the `palette.active` including the selected item.

Another thing to decide is whether we want this change app-wide or scoped to the navbar.

jtbandes:
> One thing I've not figured out in this change is how to keep the selected icon color purple when I hover over it. Right now any hovering changes the color to the palette.active including the selected item.

I think you might just need a separate rule for both selected+hover.

defunctzombie:
Updated the PR not change the color of the selected item.

amacneil (approved):
lgtm

jtbandes:
I think this is better than current main, but have a couple of minor issues with it:
- In dark mode, the new hover state is nice and bright. However, in light mode, it's way too subtle of a change.
- This hovered state is very different from almost all other hoverable controls in the app. Most of our hoverable buttons change their background color, rather than changing the icon/text color.

defunctzombie:
@2metres thoughts on the feedback from @jtbandes?

snosenzo:
I agree with @jtbandes on the light mode hover not being prominent enough.
As for the background color vs icon/text color, the vscode ui has a mix of both for what it's worth. Looks like they have text/icon color change for larger buttons/tabs and background change for smaller, then they do both for files. This sidebar behavior appears to align with what they have.

2metres (approved):

Issue #4162, opened 2022-08-17
Remove `Latching` feature and have it always on with no way to turn it off


defunctzombie:
PR that merged latching default: https://github.com/foxglove/studio/pull/3902

This feature flag default was in the v1.20.0 release on Aug 1st.

PR #4192, opened 2022-08-18
Graduate the experimental latching feature to general availability

defunctzombie:
**User-Facing Changes**
The latching feature is removed from experimental and now a default part of app playback.

**Description**
This feature is now the default and uses the new iterable players for playback on all supported data sources.

Fixes: #4162


defunctzombie:
This PR blocks my other PRs which remove the dead code paths. It makes more sense to remove this feature from experimental settings first as the user facing change so the remaining PRs have no user facing changes.

amacneil (approved):

PR #4195, opened 2022-08-19
After switching org, studio should select last current user layout

### LINKED ISSUE #4100: After switching org, studio should select first available personal layout
**Description**
After switching org, Studio forgets that we have an active layout.

We should select the first personal layout (if there is one) if the app starts with no layout and personal layouts are available.

image

- Version: main
- OS: web
- Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native): data platform

snosenzo:
Fixes #4100

**User-Facing Changes**
 - will auto select the user profile's current layout if the one in the url doesn't exist or if it belongs to another organization

**Description**
 - added logic to check that the layoutId in the url exists before setting it as the layout in CurrenLayoutProvider initialization
 - removed url layoutId initialization from useInitialDeepLinkState because it will happen in the CurrentLayoutProvider apon initialization

#4100


snosenzo:
Let me know if there's a preference for having this logic in the DeepLinkState hook or the CurrentLayoutProvider. I settled on the CurrentLayoutProvider because it already had some good fallbacks set up.

snosenzo:
Going to table this for now given the reasons described here:
https://github.com/foxglove/studio/issues/4100#issuecomment-1224750397

defunctzombie:
@snosenzo Do you want to close this our since it has been tabled?

snosenzo:
yeah sounds good

PR #4201, opened 2022-08-19
Implement event fetching and sidebar

foxymiles:
**User-Facing Changes**
Fetches events from data platform and displays them in the datasource sidebar in a new tab.

**Description**
This sets the stage for further integration of events into studio.

Screen Shot 2022-08-19 at 10 54 43 AM


packages/studio-base/src/components/DataSourceSidebar/EventsList.tsx
```diff
@@ -0,0 +1,232 @@
...
+import Stack from "@foxglove/studio-base/components/Stack";
+import { useEvents } from "@foxglove/studio-base/context/EventsContext";
+import {
+  useClearHoverValue,
+  useSetHoverValue,
+} from "@foxglove/studio-base/context/HoverValueContext";
+import { useAppTimeFormat } from "@foxglove/studio-base/hooks";
+import { ConsoleEvent } from "@foxglove/studio-base/services/ConsoleApi";
+
+const useStyles = makeStyles<void, "eventMetadata">()((theme, _params, classes) => ({
```
jtbandes, line 25: Cool, TIL about the `classes` param. Reminds me of the original [Y combinator](https://medium.com/@ayanonagon/the-y-combinator-no-not-that-one-7268d8d9c46)

packages/studio-base/src/components/EventsSyncAdapter.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+const log = Logger.getLogger(__filename);
+
+export function EventsSyncAdapter(): ReactNull {
+  const { currentUser } = useCurrentUser();
+  const urlState = useMessagePipeline((s) => s.playerState.urlState);
+  const consoleApi = useConsoleApi();
+  const setEvents = useEvents((store) => store.setEvents);
+
+  useAsync(async () => {
```
jtbandes, line 21: Did you consider having this logic live in the EventsProvider?

The `setEvents({loading:true})` etc. feels a bit redundant because `useAsync` returns an object that has something like `{ loading: boolean, error?: Error, value?: T }`. So if you moved this into the Provider then you might be able to use that return value directly, rather than setting it via the setEvents function.

jtbandes (approved):

packages/studio-base/src/components/EventsSyncAdapter.tsx
foxymiles: I initially did bundle the sync adapter with the provider but it resulted in a lot of what felt like incidental coupling with this and other contexts.

defunctzombie:
Some thoughts from using this feature:

- I was somewhat surprised that the _Events_ tab was automatically selected when opening the data source sidebar but that maybe makes sense and was only surprising because I am so use to seeing "topics" there. It is a bit strange when I open a local file I get one set of displayed items but if I open a similar file from data platform my workflow has to adjust because I'm looking at events by default. I am not sure how much I like that.
- Stylistically I found having "No events" floating in the center a bit odd (mostly because I read top of bottom and this makes me scan more to see there are no events). Tho it seems to match with our "No problems" display?
- I found the display of each event and purple hover complete different than anything else we have in the app. I'm generally of the opinion we should have consistent patterns for similar UI concepts. The events list seems like it should match the settings key/value display or we define a clearer list view for events. Right now it feels like its own bespoke UI element for no clear reason other than to be different.
- It's cool that when I hover over an event I am shown the corresponding time in the scrubber, when I hover over the scrubber I am not shown the corresponding event.

foxymiles:
> * I was somewhat surprised that the _Events_ tab was automatically selected when opening the data source sidebar but that maybe makes sense and was only surprising because I am so use to seeing "topics" there. It is a bit strange when I open a local file I get one set of displayed items but if I open a similar file from data platform my workflow has to adjust because I'm looking at events by default. I am not sure how much I like that.

I've changed the logic here so that the events tab is only displayed by default if we have a `eventId` in the url.

defunctzombie:
@2metres thoughts on my styling/UI comments?

amacneil:
- formatting duration should follow the rules console uses (seconds, us, ns)

defunctzombie (approved):

Issue #4173, opened 2022-08-17
Performance degrades over time with certain datasets


foxymiles:
**Description**
Performance degrades over time recording certain topics.  Setting a follow window for the plot doesn't seem to solve the problem.

- Version: 1.21.0
- OS: Ubuntu 20.04
- Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native): bag
- ROS distro (if applicable): Noetic

**Steps To Reproduce**
Performance degrades over time recording certain topics. This was reported on a live session but we have a bag capturing the same topics.

See ilia-baranov-2022-08-17-10-08-24.bag and https://foxglove.slack.com/archives/C028UEY858S/p1660695147452789

iliabaranov:
Thanks for opening this! Happy to reply here with any additional detail, you guys rock!

iliabaranov:
https://user-images.githubusercontent.com/90713890/185204051-c1985caa-cf90-4ff2-a9aa-ac560f753732.mp4

Screen recording of the behavior, note that Foxglove is lagging behind both rqt and Formant interface

iliabaranov:
Interesting data point: seems to lose the lag momentarily when resizing any of the windows. Perhaps this is related to the fact that resizing wipes the graph

foxymiles:
@iliabaranov Yeah that could be. Thanks for the extra detail!

foxymiles:
@iliabaranov Can you please try one more thing? Remove the plot from your layout and run the same session with no plots and see if you see the same performance degradation?

iliabaranov:
Absolutely, trying now, will see the result in ~10 mins

iliabaranov:
Issue hasn't happened since! So seems to be plot related.

foxymiles:
Ok that's very helpful. Thanks for running that test.

defunctzombie:
Thanks for reporting this. After our call we fixed up the bad performance with having the data source sidebar open. The plot panel remains something that you need to be mindful to set a window size when viewing live data.

We have ongoing work to track performance in the app across devices and panels so that will further help track how performance changes in the app over time. I don't think there's anything else actionable from this bug report but if you do find yourself able to reproduce specific and egregious performance slow downs we are always happy to investigate.

PR #4202, opened 2022-08-19
Improve datasource sidebar performance

foxymiles:
**User-Facing Changes**
This improves performance of the datasource sidebar.

**Description**
This uses a combination of debouncing updates to 2fps for the info panel and 1fps for the topics and memoizing some components to improve the performance of the datasource info sidebar. On my machine this brings the fps of the app from about 28 on the sinewave benchmark to ~58.

Would could possibly make this a little bit smarter and make the debounce time for the topics a function of the number of topics. For the sinewave benchmark reducing the debounce from 1000ms to 500ms costs about 10fps overall for the app but we could afford to debounce for a shorter time with fewer topics.

Partially addresses #4173

### DIFF
--- a/packages/studio-base/src/components/DataSourceSidebar/TopicList.tsx
+++ b/packages/studio-base/src/components/DataSourceSidebar/TopicList.tsx
@@ -18,7 +18,9 @@ import {
   TypographyProps,
 } from "@mui/material";
 import { Fzf, FzfResultItem } from "fzf";
+import { cloneDeep } from "lodash";
 import { useMemo, useState } from "react";
+import { useDebounce } from "use-debounce";
 
 import { areEqual, subtract as subtractTimes, Time, toSec } from "@foxglove/rostime";
 import {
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/components/DataSourceSidebar/TopicList.tsx
```diff
@@ -151,6 +153,54 @@ const messageFrequency = (topic: TopicWithStats, duration: Time | undefined) =>
   return `${value.toFixed(digits)} Hz`;
 };
 
+function TopicListItem({
```
defunctzombie, line 156: Separate file plz since we already have a nice folder to organize these into.

packages/studio-base/src/components/DataSourceSidebar/TopicList.tsx
```diff
@@ -168,19 +218,30 @@ export function TopicList(): JSX.Element {
     [topics, topicStats],
   );
 
-  const duration =
-    endTime != undefined && startTime != undefined ? subtractTimes(endTime, startTime) : undefined;
+  // Clone deep is necessary here because players mutate the stats directly and
+  // break memoization.
+  const [debouncedData] = useDebounce(
+    {
+      items: cloneDeep(items),
```
defunctzombie, line 225: Were we doing this close deep before?

defunctzombie:
Could we debounce the info faster? `2fps` doesn't look very "realtime" when connected to a live data source. I'm mostly thinking about the `end time` and `duration` display UX which should update relatively quickly since the user can watch those to see the data source time given we don't have a playback bar in live mode.

In addition to this debounce, I'm thinking that the live players should avoid emitting this data every frame. Right now the live players don't even use a sliding window. I was using a connection over the weekend where I tweaked the rate of a message and because I remained connected via rosbridge, the rate didn't actually change because the player calculates the rate across "all" messages rather than "recent" messages. For me it was more relevant to see recent rate - arguably even for playback of recorded data.

More broadly - this makes me lean towards an approach where we take the stats calculation our of players and into the message pipeline. But that would be a separate discussion.

packages/studio-base/src/components/DataSourceSidebar/TopicList.tsx
foxymiles: We didn't need to before because we were just passing the stats along and not trying to memoize them. Memoizing without the clone doesn't work because the players are sort of violating the expectation here that the stats object won't be mutated directly.

defunctzombie (approved):
I'm generally of the opinion that 1 component per file leads to easier to maintain and swap code since you focus on each component doing its specific task. Even when components are maybe coupled its nice to reason (and read) about each component on its own.

Changes seem good - hopefully soon we can have perf tracking numbers to avoid regressing some enhancements like this.

foxymiles:
> I'm generally of the opinion that 1 component per file leads to easier to maintain and swap code since you focus on each component doing its specific task.

Yeah I agree. I can separate this in a followup that also addresses the data source info panel.

PR #4217, opened 2022-08-19
Automatically add names to makeStyles calls from tss-react

jtbandes:
**User-Facing Changes**
None

**Description**

Adds a custom transformer to `ts-loader` that transforms calls to `makeStyles()(...)` into `makeStyles({name: "a useful name"})(...)`. This makes our `tss-react` class names human readable so you can find a pointer back to the code that generated a style.

I also added an explicit name to our Stack component, because it's so commonly used that having a shorter name is better.

Before | After
--- | ---
Screen Shot 2022-08-19 at 10 50 58 AM | Screen Shot 2022-08-19 at 11 46 26 AM


defunctzombie (approved):
I'd like us to enable this only for dev builds. For prod builds lets keep the shorter names.

packages/studio-base/webpack.ts
```diff
@@ -44,6 +45,54 @@ type Options = {
   allowUnusedVariables?: boolean;
 };
 
+/**
+ * Add `name` automatically to `makeStyles()` calls.
```
defunctzombie, line 49: Can we move this to a package in packages? It seems sensible to me to use packages as our mechanism for such organization. (Better than utils)

PR #4226, opened 2022-08-22
Persist map zoom level

### LINKED ISSUE #4225: Map panel does not remember pan and zoom
**Description**
The map panel doesn't remember pan and zoom between sessions in studio.

- OS: macOS
- Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native): nuscenes.mcap
- ROS distro (if applicable): -

**Steps To Reproduce**
1. Open nuscenes.
2. Create a new layout and add a map panel.
3. Pan and zoom to a new position in the map panel.
4. Refresh the browser.

**Expected Behavior**
The map panel should remember pan and zoom.

**Actual Behavior**
Pan and zoom are reset to defaults on refresh.

foxymiles:
**User-Facing Changes**
Remembers changes in pan and zoom for the map panel.

**Description**
We were already saving zoom level to the panel config but not applying it when loading the panel. This also extends the same logic to the map center.

Fixes #4225

### DIFF
--- a/packages/studio-base/src/panels/Map/config.ts
+++ b/packages/studio-base/src/panels/Map/config.ts
@@ -9,7 +9,7 @@ import { SettingsTreeFields, SettingsTreeNodes } from "@foxglove/studio";
 
 // Persisted panel state
 export type Config = {
-  center: undefined | { lat: number; lon: number };
+  center?: { lat: number; lon: number };
   customTileUrl: string;
   disabledTopics: string[];
   followTopic: string;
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/panels/Map/config.ts
```diff
@@ -9,12 +9,13 @@ import { SettingsTreeFields, SettingsTreeNodes } from "@foxglove/studio";
 
 // Persisted panel state
 export type Config = {
+  center: undefined | { lat: number; lon: number };
```
snosenzo, line 12: nit -- or is there a reason you don't do this?
```suggestion
  center?: { lat: number; lon: number };
```

snosenzo (approved):

defunctzombie (approved):

PR #4231, opened 2022-08-22
[New3D] Fix OccupancyGrid switching between opaque/transparent colors, enable alpha clipping, set color defaults to opaque

jhurliman:
**User-Facing Changes**

- Fixes a bug in `3D (Beta)` when switching OccupancyGrid settings between opaque and translucent colors
- Improved `3D (Beta)` rendering of opaque OccupancyGrids

**Description**

Previously, OccupancyGrids only called createMaterial() on initial mesh creation. This adds an extra check when modifying settings for transitioning from a fully opaque renderable to a renderable with (potential) transparency and recreates the material. The default color scheme has changed from 50% alpha colors to opaque to improve rendering, and opaque OccupancyGrids now use alpha clipping to cut out `100` values in the same way for both transparent and opaque renderables.

Fixes https://github.com/foxglove/studio/issues/4082

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/OccupancyGrids.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/OccupancyGrids.ts
@@ -140,21 +140,9 @@ export class OccupancyGrids extends SceneExtension<OccupancyGridRenderable> {
       // Check if the transparency changed and we need to create a new material
       const newTransparent = occupancyGridHasTransparency(renderable.userData.settings);
       if (prevTransparent !== newTransparent) {
-        renderable.userData.material.dispose();
-        renderable.userData.pickingMaterial.dispose();
-        renderable.remove(renderable.userData.mesh);
-
-        const newSettings = renderable.userData.settings;
-        const mesh = createMesh(topicName, renderable.userData.texture, newSettings);
-        const material = mesh.material as THREE.MeshBasicMaterial;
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/panels/ThreeDeeRender/renderables/OccupancyGrids.ts
```diff
@@ -131,10 +131,32 @@ export class OccupancyGrids extends SceneExtension<OccupancyGridRenderable> {
...
         | Partial<LayerSettingsOccupancyGrid>
         | undefined;
       renderable.userData.settings = { ...renderable.userData.settings, ...settings };
+
+      // Check if the transparency changed and we need to create a new material
+      const newTransparent = occupancyGridHasTransparency(renderable.userData.settings);
+      if (prevTransparent !== newTransparent) {
+        renderable.userData.material.dispose();
+        renderable.userData.pickingMaterial.dispose();
+        renderable.remove(renderable.userData.mesh);
```
jtbandes, line 145: If all that's being changed is `transparent` and `depthWrite`, I think you can get away with just setting `material.needsUpdate = true`.

jtbandes (approved):

packages/studio-base/src/panels/ThreeDeeRender/renderables/OccupancyGrids.ts
jhurliman: Transparency is explicitly listed as something that can't be changed after material creation: https://threejs.org/docs/#manual/en/introduction/How-to-update-things

PR #4233, opened 2022-08-23
Call pauseFrame/resumeFrame in PanelExtensionAdapter

defunctzombie:
**User-Facing Changes**
Extension panels wait to receive new frames until they are done processing their existing frames resulting in fewer message drops for extension panels when the computer is missing animation frames.

**Description**
When handling context updates it is important to call pauseFrame to indicate that the panel will perform some work to render the frame and resumeFrame when the panel's work is complete. Without these calls, the message pipeline will assume the panel has completed the frame and will feed another frame to the panel to render. This additional frame will override the previous and result in dropped messages for the panel.

This change adds pauseFrame/resumeFrame logic to the PanelExtensionAdapter. As part of this change I've re-organized the rendering flow logic. The `queueRender` function is replaced with a `useLayoutEffect` that triggers when various inputs for renderState change. A new function `initRenderStateBuilder` returns a builder function that creates RenderState given these various inputs. This new builder functions is invoked inside of the layout effect and the panel's render function is called.

Fixes: #4091

---

When I wrote the PanelExtensionAdapter I wanted to avoid having a panel be able to _stall_ the app by rendering slowly. The tradeoff is it does mean panels could drop frames - which is what we were seeing with the 3d panel. This change adds pause/resume which allows extension panels to stall the app. I've not yet figured out an approach to surface which panel has caused the app to stall (which is what I was doing before with the slow render state).


packages/studio-base/src/components/PanelExtensionAdapter.tsx
```diff
@@ -309,91 +259,201 @@ function PanelExtensionAdapter(props: PanelExtensionAdapterProps): JSX.Element {
...
+
+    // tell the panel to render and lockout future renders until rendering is complete
+    renderingRef.current = true;
+    try {
+      setError(undefined);
+      let doneCalled = false;
+      renderFn(renderState, () => {
+        // ignore any additional done calls from the panel
+        if (doneCalled) {
+          return;
```
jtbandes, line 434: should this log a warning/error?

defunctzombie: added a warn

jtbandes (approved):

PR #4235, opened 2022-08-23
Replace react-toast-notification with notistack and remove @fluentui/react

### LINKED ISSUE #4206: migrate components/StudioToastProvider to MUI

2metres:
**User-Facing Changes**

Before
Screen Shot 2022-08-23 at 2 50 44 pm

After
Screen Shot 2022-08-24 at 2 33 38 pm

**Description**
Remove use of @fluentui/react MessageBar in favor of @mui/Snackbar and Alert via `notistack`

Closes #4206


2metres:
My first thoughts is the I'd like to tweak the red to be more like the fluent red

2metres:
😮‍💨 

Even though that notistack uses the MUI components for some of it's internals it does not use the theme for colors

https://github.com/iamhosseindhv/notistack/blob/2160b28e2359d3c70d9a65df7e76341e478f4a03/src/SnackbarItem/SnackbarItem.tsx#L54-L69

defunctzombie:
What happened to the _see details_ text? Do we provide users a way to see the additional details of the error some other way?

2metres:
> What happened to the _see details_ text? Do we provide users a way to see the additional details of the error some other way?

Still there on the [SendNotificationToastAdapeter](https://www.chromatic.com/test?appId=603ec8bf7908b500231841e2&id=63047074e29c152db13bd013) i've added a test for the Toasts without sendNotification. That looks similar but is the underlying toast component.

Not 100% sure why we both

defunctzombie:
In your before/after screenshots I see `see details` in before but not in `after`. Does that still show up when there are details?

defunctzombie (approved):

defunctzombie:
@2metres if you merge main it will fix your remaining lint error

PR #4236, opened 2022-08-23
Moves legacy styled-components table and input under Old 3d panel for easier deletion

2metres:
**User-Facing Changes**
None

**Description**
Moves LegacyStyledComponents under old 3d panel and out of common components


2metres:
@jhurliman teeing up some more deletions for you:

alleyoop

defunctzombie (approved):

PR #4239, opened 2022-08-23
Improve performance of Autocomplete

bclinkinbeard:
Avoids unnecessarily recreating the Fzf instance and the full list of results. Significantly improves UI responsiveness for large topic lists.

**User-Facing Changes**
Better performance when searching for topic names in components like MessagePathInput


foxymiles (approved):
Makes sense. Thanks for the PR!

jhurliman:
Just a note that this PR also changes the fzf fuzzy matching algorithm used for one and two character queries. That's not necessarily good or bad, I just noticed it wasn't mentioned in the description.

Issue #4237, opened 2022-08-23
Upstream and release roslibjs changes


amacneil:
**Description**
We currently pin a specific commit in roslib:

https://github.com/foxglove/studio/blob/f67dc22be9ab51389d12333c94fe181f2dd9b5bb/packages/studio-base/package.json#L189

We should upstream our patches and release the changes.

Requested by @qdhuxp in our Slack community channel.

jtbandes:
Upstreaming these changes would be tricky, we have made a bunch of changes to the build, see: https://github.com/RobotWebTools/roslibjs/compare/develop...foxglove:roslibjs:develop

defunctzombie:
I think we should take a different tactic - which is to grab just the websocket bits of roslibjs, port them to typescript, and make that a package:

https://github.com/foxglove/studio/pull/4240

jtbandes:
Should that package be in a separate repo (perhaps a new ROS JS monorepo which we can migrate our other packages to?) It doesn’t seem exclusive to Studio and would surely be useful for other people in the community who’ve struggled with the same import/bundling issues we did with roslibjs.

defunctzombie:
> Should that package be in a separate repo (perhaps a new ROS JS monorepo which we can migrate our other packages to?) It doesn’t seem exclusive to Studio and would surely be useful for other people in the community who’ve struggled with the same import/bundling issues we did with roslibjs.

Yea - I am using that PR to get a working version first and then the idea would be a separate repo.

jtbandes:
Something to keep in mind and possibly add in our new package: https://github.com/foxglove/studio/issues/4005

jtbandes:
Moved to Linear

PR #4240, opened 2022-08-23
Add @foxglove/rosbridge-client package

defunctzombie:
**User-Facing Changes**
None

**Description**
Port of only the rosbridge communication parts of roslibjs.

Fixes: #4237


defunctzombie:
I'm setting this work aside in favor of the ros-foxglove-bridge work happening to create a new ROS node that speaks the foxglove websocket protocol. This allows us to reduce the surface area of websocket protocols we need to actively maintain.

PR #4249, opened 2022-08-24
Refactor OpenDialog/Connection to use BuiltinIcon

### LINKED ISSUE #4210: migrate components/OpenDialog/Connection to MUI
packages/studio-base/src/components/OpenDialog/Connection.tsx:import { Icon } from "@fluentui/react";

2metres:
**User-Facing Changes**
None

**Description**
Use BuiltinIcon instead of fluentui/Icon in OpenDialog/Connection

Closes #4210

### DIFF
--- a/packages/studio-base/src/components/BuiltinIcon.tsx
+++ b/packages/studio-base/src/components/BuiltinIcon.tsx
@@ -5,10 +5,13 @@
 import ICONS from "@foxglove/studio-base/theme/icons";
 
 type BuiltinIconProps = {
-  name: keyof typeof ICONS;
+  name?: RegisteredIconNames;
 };
 
 function BuiltinIcon(props: BuiltinIconProps): JSX.Element {
+  if (props.name == undefined) {
+    return <></>;
+  }
   return ICONS[props.name];
 }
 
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/components/BuiltinIcon.tsx
```diff
@@ -5,10 +5,13 @@
 import ICONS from "@foxglove/studio-base/theme/icons";
 
 type BuiltinIconProps = {
-  name: keyof typeof ICONS;
+  name?: RegisteredIconNames;
 };
 
 function BuiltinIcon(props: BuiltinIconProps): JSX.Element {
+  if (props.name == undefined) {
```
defunctzombie, line 12: Should this throw if no name is provided? In what situation do we want this to silently produce nothing?

maybe in dev mode we should throw here to catch instances of this during development but be silent in prod?

packages/studio-base/src/components/OpenDialog/Connection.tsx
```diff
@@ -118,10 +118,16 @@ export default function Connection(props: ConnectionProps): JSX.Element {
             {enabledSourcesFirst.map((source, idx) => {
               const { id, iconName, displayName } = source;
               return (
-                <Tab value={idx} key={id} icon={<Icon iconName={iconName} />} label={displayName} />
+                <Tab
+                  value={idx}
+                  key={id}
+                  icon={<BuiltinIcon name={iconName} />}
```
defunctzombie, line 124: Instead of allowing `name` to be optional (which I think feels weird), you could look at optionally setting icon property here? or maybe using `iconName ?? 'placeholder'` or some name that indicates a placeholder icon.

defunctzombie (approved):

PR #4252, opened 2022-08-24
[New3D] Do not load lights from COLLADA files

jhurliman:
**User-Facing Changes**

- Fixed visualization of COLLADA files containing lights in `3D (Beta)` panel

**Description**

If a COLLADA (.dae) mesh included its own lights, those were getting loaded into the scene which would generally blow out the lighting since we already include two scene lights. This change removes any embedded lights in COLLADA scene graphs

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/markers/RenderableMeshResource.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/markers/RenderableMeshResource.ts
@@ -98,9 +98,12 @@ export class RenderableMeshResource extends RenderableMarker {
       return;
     }
 
-    const mesh = opts.useEmbeddedMaterials
-      ? cachedModel.clone(true)
-      : replaceMaterials(cachedModel.clone(true), this.material);
+    const mesh = cachedModel.clone(true);
+    removeLights(mesh);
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/panels/ThreeDeeRender/renderables/markers/RenderableMeshResource.ts
```diff
@@ -111,7 +114,21 @@ export class RenderableMeshResource extends RenderableMarker {
...
+  // Remove lights from the model
+  const lights: THREE.Light[] = [];
+  model.traverse((child: THREE.Object3D) => {
+    const maybeLight = child as Partial<THREE.Light>;
+    if (maybeLight.isLight === true) {
+      lights.push(maybeLight as THREE.Light);
+    }
+  });
+  for (const light of lights) {
+    light.removeFromParent();
```
jtbandes, line 127: any `dispose` needed?

jtbandes (approved):

packages/studio-base/src/panels/ThreeDeeRender/renderables/markers/RenderableMeshResource.ts
jhurliman: Yes, good catch

PR #4254, opened 2022-08-24
Implement settings number field scrubbing

foxymiles:
**User-Facing Changes**
This allows users to click and drag on number values in the settings editor to set the value interactively.

**Description**
This is a pretty straightforward implementation that relies on pointer capture.

https://user-images.githubusercontent.com/93935560/186478245-949641ad-30a6-4894-8d8f-c8d097e64427.mov

### DIFF
--- a/packages/studio-base/src/components/SettingsTreeEditor/inputs/NumberInput.tsx
+++ b/packages/studio-base/src/components/SettingsTreeEditor/inputs/NumberInput.tsx
@@ -4,7 +4,7 @@
 
 import ChevronLeftIcon from "@mui/icons-material/ChevronLeft";
 import ChevronRightIcon from "@mui/icons-material/ChevronRight";
-import { IconButton, TextFieldProps, TextField } from "@mui/material";
+import { IconButton, TextFieldProps, TextField, darken, lighten } from "@mui/material";
 import { clamp, isFinite } from "lodash";
 import { ReactNode, useCallback, useState } from "react";
 import { useKeyPress, useLatest } from "react-use";
@@ -27,7 +27,6 @@ const useStyles = makeStyles()((theme) => ({
   },
 
   textField: {
-    backgroundColor: theme.palette.background.paper,
     cursor: "auto",
     ".MuiInputBase-formControl.MuiInputBase-root": {
       paddingTop: 0,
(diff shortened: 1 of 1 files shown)


jtbandes:
This is cool, but makes it impossible to use the regular drag to select text in the field...

foxymiles:
> This is cool, but makes it impossible to use the regular drag to select text in the field...

That's a tradeoff blender makes too. You can still double click to select all.

jtbandes:
Two more bugs: doesn't work when field is empty, and is very susceptible to floating point precision issues. This is an Image panel settings:

https://user-images.githubusercontent.com/14237/186486642-42dd40a5-db47-4477-99b9-320a0cba6113.mov

foxymiles:
@jtbandes Thanks I'll fix the empty case. For the precision issue I recommend using the `precision` field of the number input.

jtbandes:
>  For the precision issue I recommend using the `precision` field of the number input.

Maybe we need a default precision then? I don't think this should be the default behavior.

wimagguc:
When the user is done with dragging the input, the cursor is where they left off:

https://user-images.githubusercontent.com/349687/186524006-84811fbd-c188-4318-bdcd-8b71938b58ad.mov

amacneil:
- need to fix step / rounding issue
- need to either show cursor, or keep it hidden and enable pointer lock while dragging (so that the cursor stays in place when you release)
- dragging inputs in 3d panel seems completely broken

foxymiles:
> * need to fix step / rounding issue

Changed this to set a default precision of 2 if none is specified.

> * need to either show cursor, or keep it hidden and enable pointer lock while dragging (so that the cursor stays in place when you release)

Changed this to just show the <-> cursor instead of trying to lock the pointer.

> * dragging inputs in 3d panel seems completely broken

We may need to disable this for new3d for now.

foxymiles:
> In Blender, when the user is done dragging, the cursor stays over the input:

Thanks for posting this. Given the downsides of pointer lock in the browser I think I'll just show the <-> cursor instead.

amacneil:
Read only fields should hide the scrub icon and arrows
image

foxymiles:
> Read only fields should hide the scrub icon and arrows

This should be fixed now.

defunctzombie (approved):

PR #4256, opened 2022-08-24
[New3D] Labels should be above (+Z) coordinate frame axes in the fixed frame, not the local frame

### LINKED ISSUE #3716: [New3D] Labels should be above (+Z) coordinate frame axes in the fixed frame, not the local frame
We currently set the `FrameAxisRenderable` label child to <0, 0, 0.4>. This works in simple cases, but if the child->parent transform has a rotation, the label will not be 0.4 units above the `Axis`. The proper solution is likely more involved since it would require updating per frame.

The 0.4 constant is worth reconsidering as well. Now that labels and frame axes can be resized, maybe it should be based on the frame axis scale or label size.

Screen Shot 2022-06-25 at 5 27 30 PM

snosenzo:
Fixes #3716

**User-Facing Changes**
 - FrameAxes labels will always appear above (+Z) coordinate frame axes in the fixed frame, not the local frame

**Description**
 - set label position in renderFrame and translate back to local
 - added comments for FrameAxes.startFrame for readability

#3716


jhurliman:
@snosenzo I made some aesthetic tweaks to use named variables instead of `tempVec*` so the code is a little easier to follow. Otherwise, LGTM.

I think there's a follow-up nice-to-have where `LABEL_OFFSET` is dynamically calculated based on label scale. Maybe multiplying <0, 0, 0.4> by label scale is enough.

jhurliman (approved):

jtbandes (approved):

Issue #4212, opened 2022-08-19
migrate panels/URDFViewer to MUI


defunctzombie:
And all files in that panel

2metres:
IMO the the joint states should move into Panel settings as Number Inputs

PR #4260, opened 2022-08-24
Move all URDF panel controls to settings

foxymiles:
**User-Facing Changes**
Moves all controls in the URDF panel to the settings sidebar.

Screen Shot 2022-08-31 at 5 53 59 PM

Fixes https://github.com/foxglove/studio/issues/4212


amacneil:
- set all inputs to 2dp (same as what we do on main)
- load min/max range from urdf

defunctzombie (approved):

PR #4262, opened 2022-08-24
[New3D] Position frame axis labels based on axis length and label size

jhurliman:
**User-Facing Changes**

- Improve coordinate frame label positioning in `3D (Beta)` panel

**Description**

Compute the label offset based on the axis length and label size. We want the label to float a little above the up axis arrow, proportional to the height of the label. This makes the label position look consistent regardless of user scale settings

Before:
Screen Shot 2022-08-24 at 1 29 20 PM
After:
Screen Shot 2022-08-24 at 1 28 01 PM

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/FrameAxes.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/FrameAxes.ts
@@ -18,7 +18,7 @@ import { SettingsTreeEntry } from "../SettingsManager";
 import { getLuminance, stringToRgb } from "../color";
 import { BaseSettings, fieldSize } from "../settings";
 import { Duration, Transform, makePose, CoordinateFrame, MAX_DURATION } from "../transforms";
-import { Axis } from "./Axis";
+import { Axis, AXIS_LENGTH } from "./Axis";
 import {
   DEFAULT_AXIS_SCALE,
   DEFAULT_LINE_COLOR_STR,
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/panels/ThreeDeeRender/renderables/FrameAxes.ts
```diff
@@ -226,7 +233,7 @@ export class FrameAxes extends SceneExtension<FrameAxisRenderable> {
       }
 
       // Add the label offset in "world" coordinates (in the render frame)
-      worldPosition.add(LABEL_OFFSET);
+      worldPosition.add(labelOffset);
```
jtbandes, line 236: nit: don't really need a whole vector3 for this, could just do `worldPosition.z += axisLength + labelSize * 1.5`

jtbandes (approved):

snosenzo:
I was wondering where Label Size was haha. Do we want to factor label scale  in? Is there a reason there's both label size and label scale?

jhurliman:
> I was wondering where Label Size was haha. Do we want to factor label scale in? Is there a reason there's both label size and label scale?

Ah you're right. The label scale is global across all labels in the scene (such as TEXT_VIEW_FACING markers), where label size is "size of the labels to use for frame axes". I'll fix this to incorporate both settings in the calculation.

snosenzo:
sounds good, I'll close my PR since this will cover it then.

Issue #4222, opened 2022-08-21
"Data went back in time" problem appears while scrubbing


rgov:
Foxglove Studio 1.22.0

If I scrub rapidly back and forth through a ROS1 bag file, a red (1) appears on the data source sidebar icon for a fraction of a second. The problem appears to be "data went back in time."

I think this may be related to which parts of the file have been loaded in, and maybe it appears when new data has to be read from disk? Not sure.

defunctzombie:
Adding a bug label because I don't think this should be happening and might be an indicator the message pipeline is not clearing as expected.

If you can reproduce this reliably could you record a video of it happening?

rgov:
https://user-images.githubusercontent.com/108767/185796292-b397f44a-8e29-4675-b12a-13817caef774.mov

amacneil:
It’s not wrong… the data did go back in time 😄

I assume Latching setting is enabled?

rgov:
I didn't know of this setting but it is enabled, although the issue seems to occur regardless of this setting.

PR #4266, opened 2022-08-24
Set _this.messages alongside _lastSeekEmitTime in stateSeekBackfill

defunctzombie:
**User-Facing Changes**
Quickly scrubbing back and forth does not cause a "Data went back in time" error to appear in data source problems.

**Description**
`stateSeekBackfill` would set `this._messages` after loading the backfill messages separately from setting `this._lastSeekEmitTime` which results in emitting a set of messages with no seekEmitTime making it look like the emit is a regular message playback rather than a seek. This would trip the message order tracker when a batch of older messages from a seek backfill (seeking to an older time) would be emitted without the seek emit time (because the user seeks again immediately and there's a next state which prevents seek backfill from setting the emit time. Since we have a "background" job for preloading data, this job would call `queueEmitState` which is what ultimately resulted in emitting a batch of messages from a seek backfill without their last emit time.

The fix is to move the logic in `stateSeekBackfill` to set messages after the `_nextState` check to ensure outgoing messages are only set if also setting the `lastSeekEmitTime`.

Fixes: #4222

### DIFF
--- a/packages/studio-base/src/components/MessagePipeline/MessageOrderTracker.ts
+++ b/packages/studio-base/src/components/MessagePipeline/MessageOrderTracker.ts
@@ -115,7 +115,6 @@ class MessageOrderTracker {
             `Processed a message on ${message.topic} at ${formattedTime} which is earlier than ` +
             `last processed message on ${this.lastMessageTopic} at ${lastMessageTime}.`;
 
-          log.error(errorMessage);
           problems.push({
             severity: "warn",
             message: "Data went back in time",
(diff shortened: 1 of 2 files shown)


jtbandes (approved):

jtbandes:
Is this something we can unit test?

defunctzombie:
> Is this something we can unit test?

Tricky but I think the answer should be yes because we can control the inputs and preloading.

defunctzombie:
> Is this something we can unit test?

Tragically I have been unable to reproduce with a test using the existing api. The wall I hit was that I need to trigger a state change but there's no api for me to do that. I can't "play" because there's logic preventing that until backfill finishes and there's no way to trigger idle.

I could expose another api specifically to do the test. I am gonna merge this to get the bug fix and re-visit a test pending thoughts on what we want to do here.

Issue #4216, opened 2022-08-19
Delete variable doesn't work as expected


snosenzo:
**Description**
There's a few different issues here:
1. The delete variable prompt doesn't close after a variable has been deleted (not sure if this is the desired outcome)
2. The delete variable action visually removes the last variable in the list rather than the one the prompt was opened from
3. Then when a new variable is created with the same name as the bottom variable that was visually removed, it throws an error saying it already exists. So presumably the second variable was deleted actually but the UI did not update to reflect that.

- Version:  1.22.0-dev
- OS: web/macos

**Steps To Reproduce**

![delete variables bug](https://user-images.githubusercontent.com/10187776/185681455-f87fecfe-d585-4f93-95d1-6b79b494e4c7.gif)

PR #4272, opened 2022-08-25
Fix variable sidebar

defunctzombie:
**User-Facing Changes**
* Close the context menu when a variable is deleted
* Fix issue with adding a new variable but being told it already exists (when it isn't in the list)
* Fix issue with removing the correct deleted variable from the list

**Description**

The core of these issues is the `key` field within each <Variable> element in the Variable sidebar react component. This key is the list index which meant that when a variable in the middle of the list was removed and the render was re-creating the components from the `map` over the variables, the index made react think the particular item is the same key.

It would pass in the new name and props but because the editableName state was only set to the `name` prop once, the element displayed the wrong (stale) name.

This change updates the keys to key off the variable names and only updates the variable name once the user has blurred from the input. This allows the user to freely iterate on the variable name even through duplicate names and only updates the name at the end if it isn't in conflict.

Fixes: #4216

### DIFF
--- a/packages/studio-base/src/components/VariablesSidebar/index.tsx
+++ b/packages/studio-base/src/components/VariablesSidebar/index.tsx
@@ -79,7 +79,7 @@ export default function VariablesSidebar(): ReactElement {
         <Divider />
         {linked.map((name, idx) => (
           <Variable
-            key={`linked.${idx}`}
+            key={`linked.${name}`}
             name={name}
             selected={!skipAnimation.current && changedVariables.includes(name)}
             linked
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/components/VariablesSidebar/index.tsx
```diff
@@ -79,7 +79,7 @@ export default function VariablesSidebar(): ReactElement {
         <Divider />
         {linked.map((name, idx) => (
           <Variable
-            key={`linked.${idx}`}
+            key={`linked.${name}`}
```
foxymiles, line 82: I wonder if we should put both the idx and name in the key?

foxymiles (approved):

packages/studio-base/src/components/VariablesSidebar/index.tsx
defunctzombie: I was always under the impression that react recommended against using the `idx` but we certainly can do that here. The new logic with editing the input means that neither the name nor idx are updated until "saving". Previously using _name_ would result in losing focus on the input.

I think _name_ is ok here now given that they are unique.

Issue #4269, opened 2022-08-24
Missing end time and duration


jtbandes:
Steps to reproduce:
- visit studio.foxglove.dev
- load example data

<img width="437" alt="image" src="https://user-images.githubusercontent.com/14237/186530899-f2e71aaf-c7f5-4a99-b156-8e25faee3e77.png">

defunctzombie:
@jtbandes do you happen to know when this regressed? This seems recent.

jtbandes:
I don't know. I didn't notice it until today, but I noticed styling issues (#4130) a little while ago (maybe unrelated?)

As a quick way to check old builds, I usually go back to random PRs from a few days/weeks ago and look at the vercel preview builds.

PR #4238, opened 2022-08-23
Improve data source info panel performance

foxymiles:
**User-Facing Changes**
This improves the performance of the data source info panel.

**Description**
This uses refs to bypass react entirely for rapidly updating message counts and frequencies.

Partially addresses #4173

PR #4273, opened 2022-08-25
Fix datasource info view end time and duration

foxymiles:
**User-Facing Changes**
This fixes an issue with the display of end time and duration for non-live data sources.

**Description**
This fixes a regression from https://github.com/foxglove/studio/pull/4238. The issue is that non-live data sources don't send multiple updates for the session end time so for these data sources we also need to update when the player presence changes in order to display the correct end time and duration.

Fixes https://github.com/foxglove/studio/issues/4269

### DIFF
--- a/packages/studio-base/src/components/DataSourceInfoView.tsx
+++ b/packages/studio-base/src/components/DataSourceInfoView.tsx
@@ -21,15 +21,11 @@ import { fonts } from "@foxglove/studio-base/util/sharedStyleConstants";
 
 import { MultilineMiddleTruncate } from "./MultilineMiddleTruncate";
 
-const useStyles = makeStyles()((theme) => ({
+const useStyles = makeStyles()({
   numericValue: {
-    color: theme.palette.secondary.main,
     fontFamily: fonts.MONOSPACE,
   },
-  value: {
-    color: theme.palette.secondary.main,
-  },
-}));
+});
 
 const selectStartTime = (ctx: MessagePipelineContext) => ctx.playerState.activeData?.startTime;
 const selectEndTime = (ctx: MessagePipelineContext) => ctx.playerState.activeData?.endTime;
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/components/DataSourceInfoView.tsx
```diff
@@ -74,7 +87,7 @@ function DataSourceInfoContent(props: {
         {playerPresence === PlayerPresence.INITIALIZING ? (
           <Skeleton animation="wave" width="50%" />
         ) : (
-          <Typography variant="inherit" ref={endTimeRef}>
+          <Typography className={classes.numericValue} variant="inherit" ref={endTimeRef}>
```
snosenzo, line 90: doesn't seem related to this issue but how is this updated? why doesn't it have a  `<Timestamp time={endTime}...` tag?

packages/studio-base/src/components/DataSourceInfoView.tsx
```diff
@@ -16,9 +17,16 @@ import { useAppTimeFormat } from "@foxglove/studio-base/hooks";
 import { subtractTimes } from "@foxglove/studio-base/players/UserNodePlayer/nodeTransformerWorker/typescript/userUtils/time";
 import { PlayerPresence } from "@foxglove/studio-base/players/types";
 import { formatDate, formatDuration } from "@foxglove/studio-base/util/formatTime";
+import { fonts } from "@foxglove/studio-base/util/sharedStyleConstants";
 
 import { MultilineMiddleTruncate } from "./MultilineMiddleTruncate";
 
+const useStyles = makeStyles()({
+  numericValue: {
+    fontFamily: fonts.MONOSPACE,
```
defunctzombie, line 26: Can we get this from the mui theme instead? I think @2metres was trying to move away from referencing the sharedStyleConstants directly.

defunctzombie (approved):

packages/studio-base/src/components/DataSourceInfoView.tsx
snosenzo: oh I see in the parent component that it's updating the DOM directly -- and I gather it's because the end time for live datasources is constantly changing 👍

packages/studio-base/src/components/DataSourceInfoView.tsx
foxymiles: Yeah probably. I'd like to get this fix out since it's a release blocker but it seems like we shouldn't need to reference fonts this way.

packages/studio-base/src/components/DataSourceInfoView.tsx
foxymiles: Yeah this is a performance hack so we can render this fast.

PR #4278, opened 2022-08-25
Add filePath to IIterableSource Initialization result

defunctzombie:
**User-Facing Changes**
On macOS the "proxy icon" updates to show the currently open local file.

**Description**
The desktop app uses filePath to show the "proxy icon" when working with a local file. This change updates PlayerManager to set the representedFilename when a user selects a single file.

### DIFF
--- a/packages/studio-base/src/components/DocumentTitleAdapter.tsx
+++ b/packages/studio-base/src/components/DocumentTitleAdapter.tsx
@@ -8,18 +8,14 @@ import {
   MessagePipelineContext,
   useMessagePipeline,
 } from "@foxglove/studio-base/components/MessagePipeline";
-import { useNativeWindow } from "@foxglove/studio-base/context/NativeWindowContext";
 
 const selectPlayerName = (ctx: MessagePipelineContext) => ctx.playerState.name;
-const selectPlayerFilePath = (ctx: MessagePipelineContext) => ctx.playerState.filePath;
 
 /**
  * DocumentTitleAdapter sets the document title based on the currently selected player
  */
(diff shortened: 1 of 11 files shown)


defunctzombie:
@jtbandes I wasn't sure exactly at what level to put filePath support - at first I was thinking it _could_ be an option to IterablePlayer but eventually landed on IIterableSource#initialize since that keeps the sources self-contained and lets you pass a source to the player without it having to know more about it.

jtbandes (approved):

packages/studio-base/src/components/DocumentTitleAdapter.tsx
```diff
@@ -8,18 +8,14 @@ import {
...
 
 const selectPlayerName = (ctx: MessagePipelineContext) => ctx.playerState.name;
-const selectPlayerFilePath = (ctx: MessagePipelineContext) => ctx.playerState.filePath;
 
 /**
  * DocumentTitleAdapter sets the document title based on the currently selected player
  */
 export default function DocumentTitleAdapter(): JSX.Element {
-  const nativeWindow = useNativeWindow();
   const playerName = useMessagePipeline(selectPlayerName);
```
jtbandes, line 18: Do you want to move this too? It is very similar in purpose to the representedFilename thing.

defunctzombie:
https://github.com/electron/electron/issues/33647 😭

packages/studio-base/src/components/DocumentTitleAdapter.tsx
defunctzombie: Maybe another day ♾️

PR #4288, opened 2022-08-26
Make variable slider settings less strict

### LINKED ISSUE #4215: Slider Step setting too strict for typing
**Description**
The slider panel's step setting is too strict when the user is typing into it. It doesn't allow the user to type in a number starting with zero or allow for hanging decimal points. 
A workaround for this is typing the decimal you want elsewhere and pasting it in. 

- OS: web/macos
- Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native): n/a
- ROS distro (if applicable): n/a

**Steps To Reproduce**

**Expected Behavior**
 - allow me to delete the default 1 value and replace it with 0.1
 - allow me to delete the "1" in "0.1" and type "2" to make "0.2"

**Actual Behavior**
 - deleting or otherwise changing the value into an invalid input does not register the text change. If the user needs to pass through an invalid state to get to a valid one, they must write and copy the value elsewhere to paste into the input.

foxymiles:
**User-Facing Changes**
This makes the variable slider panel's settings less strict.

**Description**
This permits fractional or undefined values for variable slider settings and uses defaults when they are absent. It also refactors the settings code.

Fixes #4215

### DIFF
--- a/packages/studio-base/src/panels/VariableSlider/index.tsx
+++ b/packages/studio-base/src/panels/VariableSlider/index.tsx
@@ -12,108 +12,45 @@
 //   You may not use this file except in compliance with the License.
 
 import { Slider, Typography, useTheme } from "@mui/material";
-import produce from "immer";
-import { set } from "lodash";
-import { useCallback, useEffect } from "react";
+import { useCallback } from "react";
 
-import { SettingsTreeAction, SettingsTreeNodes } from "@foxglove/studio";
 import Panel from "@foxglove/studio-base/components/Panel";
-import { usePanelContext } from "@foxglove/studio-base/components/PanelContext";
 import PanelToolbar from "@foxglove/studio-base/components/PanelToolbar";
 import Stack from "@foxglove/studio-base/components/Stack";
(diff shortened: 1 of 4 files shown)


packages/studio-base/src/panels/VariableSlider/index.tsx
```diff
@@ -129,9 +66,9 @@ function VariableSliderPanel(props: Props): React.ReactElement {
         paddingX={3}
       >
         <Slider
-          min={sliderProps.min}
-          max={sliderProps.max}
-          step={sliderProps.step}
+          min={min}
+          max={max}
+          step={step}
```
snosenzo, line 71: not an issue but maybe a warning comment : if the marks prop is ever the default (marks = true) then `step` cannot be less than or equal to 0

snosenzo (approved):

packages/studio-base/src/panels/VariableSlider/index.tsx
foxymiles: Good to point that out.

PR #4291, opened 2022-08-26
[New3D] Fix URDF transforms disappearing on seek or loop

jhurliman:
**User-Facing Changes**

- Fixes URDF transforms disappearing in `3D (Beta)` when seeking or looping playback

**Description**

Renderer#clear() is called on any playback discontinuity (seeking, looping) which clears the transform trees and most renderables. The URDF SceneExtension was already overriding removeAllRenderables() to prevent the URDF model from disappearing, but Renderer.transformTree gets cleared as well which made the URDF-loaded transforms disappear. This change adds them back to the scene when removeAllRenderables() is called

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
@@ -107,6 +107,9 @@ export class UrdfRenderable extends Renderable<UrdfUserData> {
 }
 
 export class Urdfs extends SceneExtension<UrdfRenderable> {
+  private frames: string[] = [];
+  private transforms: TransformData[] = [];
+
   public constructor(renderer: Renderer) {
     super("foxglove.Urdfs", renderer);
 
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
```diff
@@ -185,7 +193,8 @@ export class Urdfs extends SceneExtension<UrdfRenderable> {
   }
 
   public override removeAllRenderables(): void {
-    // no-op
+    // Re-add coordinate frames and transforms since the scene has been cleared
+    this._loadTransforms(this.frames, this.transforms);
```
snosenzo, line 197: is the tree always cleared before `removeAllRenderables` has been called? Should there be a note on order there?

snosenzo: looks like it is, but there should probably be a comment where `transformTree.clear()` is called in `Renderer.ts` saying that it should be called before `removeAllRenderables`

snosenzo (approved):

Issue #4293, opened 2022-08-26
Debounce setSubscriptions in message pipeline


defunctzombie:
Rather than calling the player setSubscriptions immediately on every subscription update, debounce the calls so that during a burst of updates each call does not make the player update it's subscriptions.

PR #4294, opened 2022-08-26
Add debounce to setSubscriptions in message pipeline

defunctzombie:
**User-Facing Changes**
None

**Description**
Previously the requestBackfill method debounced telling a player to update itself with new subscriptions. This method was removed from the pipeline in https://github.com/foxglove/studio/pull/4292. This change brings back the debounce logic on setSubscriptions to batch subscriber updates before they get to the player. This helps avoid the player starting to fetch topics only to immediately have to fetch other topics.

Fixes: #4293

### DIFF
--- a/packages/studio-base/src/components/MessagePipeline/index.tsx
+++ b/packages/studio-base/src/components/MessagePipeline/index.tsx
@@ -11,7 +11,7 @@
 //   found at http://www.apache.org/licenses/LICENSE-2.0
 //   You may not use this file except in compliance with the License.
 
-import { flatten } from "lodash";
+import { debounce, flatten } from "lodash";
 
 import { Condvar } from "@foxglove/den/async";
 import { useShallowMemo } from "@foxglove/hooks";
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/components/MessagePipeline/index.tsx
```diff
@@ -125,7 +125,28 @@ export function MessagePipelineProvider({
...
+
+  // Debounce the subscription updates for players. This batches multiple subscribe calls
+  // into one update for the player which avoids fetching data that will be immediately discarded.
+  //
+  // The delay of 0ms is intentional as we only want to give one timeout cycle to batch updates
+  const debouncedPlayerSetSubscriptions = useMemo(() => {
+    return debounce((subs: SubscribePayload[]) => {
+      player?.setSubscriptions(subs);
+    });
+  }, [player]);
```
foxymiles, line 137: Maybe a bit cleaner to use the `useDebouncedCallback` hook here?

foxymiles (approved):

packages/studio-base/src/components/MessagePipeline/index.tsx
defunctzombie: I went that route at first and then didn't like that their implementation had so many hooks internally - it was basically the same thing as using the lodash debounce with more overhead.

PR #4295, opened 2022-08-28
[New3D] Per-point point cloud and laser scan picking

### LINKED ISSUE #3707: Support picking of individual items in instanced renderables & point clouds

jhurliman:
**User-Facing Changes**

- Individual points in point clouds and laser scans can be selected in `3D (Beta)`

**Description**

Screen Shot 2022-09-07 at 11 27 14 AM

This adds a pickInstance() method that is called to render a hitmap of instance indices for an instanced renderable. This allows individual points, lines, and primitives to be selected. Renderables can set `public override pickableInstances = true` and optionally specify a custom `userData.instancePickingMaterial` that should output `gl_FragColor` (or equivalent) equal to the 32-bit integer `instanceId` of the picked instance. If this is not specified, a default shader is used that will use `gl_InstanceID`. Point clouds and laser scans currently set a custom shader that uses `gl_VertexID`.

Renderables also need to implement `instanceDetails(instanceId)` returning a POJO of per-instance details that will be visualized in the UI.

Fixes #3707


packages/studio-base/src/panels/ThreeDeeRender/Picker.ts
```diff
@@ -157,53 +246,119 @@ export class Picker {
...
+    return renderMaterial;
+  }
+}
+
+function setObjectId(material: THREE.ShaderMaterial, objectId: number): void {
+  const iObjectId = material.uniforms.objectId;
+  if (!iObjectId) {
+    throw new Error(`objectId uniform not found in picking material`);
+  }
+  iObjectId.value = [
```
snosenzo, line 355: should we extract this out into `fromPixelBuffer` and `toPixelBuffer` methods that we can test? or is this robust enough

jhurliman: I think the wrapping `setObjectId()` is already extracted enough. The specific math is going to show up a few times because it has multiple language implementations (JS + GLSL).

packages/studio-base/src/panels/ThreeDeeRender/renderables/PointCloudsAndLaserScans.ts
```diff
@@ -1118,6 +1265,39 @@ function createPickingMaterial(
   });
 }
 
+function createInstancePickingMaterial(
```
snosenzo, line 1268: could this be used for other instance renderables as well?  or is it specific to Point clouds and laser scans?

jhurliman: This code is very specific to point clouds. It won't work for laser scans or any other renderable

packages/studio-base/src/panels/ThreeDeeRender/Picker.ts
snosenzo: is it worth centralizing that so that whenever it's changed, all the changes can happen in one place? like having a file with both the JS/TS and GLSL (as just GLSL string template of the function) implementations

jhurliman: I don't see how that would work without a more complex include system, the full vertex and fragment shaders are specific to point clouds and laser scans.

The first time this needs to be changed we can figure out if a refactoring makes sense.

snosenzo (approved):

PR #4296, opened 2022-08-28
Add "todo" to list of disallowed words in comments

defunctzombie:
Similar to "fixme", this change adds "todo" to the list of disallowed words in comments.

Rather than use TODO comments in the code, we create tickets for any follow-up work that needs to happen. Tickets are how we review outstanding work and ensure it is prioritized and completed.

If there is an implementation tradeoff, a code comment should describe the comment and and references (which could be tickets too) but it should not do this in the form of a TODO.

**User-Facing Changes**

**Description**

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -59,7 +59,7 @@ rules:
 
   no-warning-comments:
     - error
-    - terms: ["fixme", "xxx"]
+    - terms: ["fixme", "xxx", "todo"]
       location: anywhere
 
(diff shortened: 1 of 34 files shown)


defunctzombie:
@jhurliman I've deleted a few TODO items in the ThreeDeeRender codebase. If these are important for GA release please ticket them and add to the epic for tracking.

foxymiles (approved):

jtbandes:
🔪

PR #4299, opened 2022-08-29
Implement events in the timeline

foxymiles:
**User-Facing Changes**
Adds support for display of user events on the timeline including hover interactions.

**Description**
Renders events on the timeline and adds logic to detect when a user has hovered within an event's time bounds.

https://user-images.githubusercontent.com/93935560/187263757-e32e057f-d7e0-4f4c-9dcd-648d3f21c08f.mov

Fixes https://github.com/foxglove/studio/issues/4150


2metres:
- When hovering sidebar only the event you are hovering should be emphasized in the playback bar
- When hovering playback bar all events under the cursor/playhead should highlight in the sidebar
- When event is selected in sidebar clicking event in list to deslect

2metres:
Overall, Looking good. One outstanding issue the sidebar hover still not working exactly i would expect:

https://user-images.githubusercontent.com/924528/188022489-603bee6a-f34a-4837-975c-fe2fd993ae8f.mov

foxymiles:
> Overall, Looking good. One outstanding issue: the sidebar hover still not working exactly i would expect. I would expect the small event to be highlighted in the progress bar and the larger event to not/
> 

I think we've created some somewhat confusing hover semantics for ourselves here. Hovering over an event is treated like a hover at that point in time elsewhere in the UI. If you were to hover at that time point in a plot panel then all events would be hovered. Maybe we need a different state in the events list for "mouse is over this single event" vs "these events overlap this point in time"?

foxymiles:
@2metres I've created a distinction in the timeline state between "events that overlap the current hover value" and "an event directly hovered by the user in the events list".

packages/studio-base/src/components/EventsSyncAdapter.tsx
```diff
@@ -2,41 +2,132 @@
...
       urlState.parameters != undefined
     ) {
       const queryParams = urlState.parameters as { deviceId: string; start: string; end: string };
       setEvents({ loading: true });
       try {
-        const events = await consoleApi.getEvents(queryParams);
-        setEvents({ loading: false, value: events });
+        const fetchedEvents = await consoleApi.getEvents(queryParams);
+        const positionedEvents = positionEvents(fetchedEvents, startTime, endTime);
+        setEvents({ loading: false, value: positionedEvents });
```
jtbandes, line 98: I still find it weird that we use this setState thing inside useAsync. The useAsync hook itself returns a `{loading, value}` object that we could use directly, so the function here would just have to `return positionedEvents`.

packages/studio-base/src/providers/EventsProvider.tsx
```diff
@@ -4,20 +4,58 @@
...
-      selectEvent: (id: undefined | string) => set({ selectedEventId: id }),
-      setEvents: (events: AsyncState<ConsoleEvent[]>) => set({ events }),
-    })),
-  );
+  const [store] = useState(createEventsStore);
 
   return <EventsContext.Provider value={store}>{children}</EventsContext.Provider>;
 }
+
+const selectFilteredEvents = createSelector(
```
jtbandes, line 36: What benefit are we getting from createSelector here? I ask because we are pretty close to being able to get rid of the reselect dependency. It's only used in 2-3 other places.

packages/studio-base/src/services/ConsoleApi.ts
```diff
@@ -54,7 +56,11 @@ export type ConsoleEvent = {
   createdAt: string;
   deviceId: string;
   durationNanos: string;
+  endTime: Time;
+  endTimeInSeconds: number;
```
jtbandes, line 60: why keep both of these next to each other? Our app usually deals in one or the other (mostly Time)

jtbandes (approved):
🎉

packages/studio-base/src/components/EventsSyncAdapter.tsx
foxymiles: You mean add a `useEffect` to respond to the return from the `useAsync` to call `setEvents`?

packages/studio-base/src/providers/EventsProvider.tsx
foxymiles: Actually I'd like to make a lot more use of `createSelector` since it's a nice way to build up memoized derived state but we've made it hard to use by spreading our state out among independent contexts.

packages/studio-base/src/services/ConsoleApi.ts
foxymiles: The timeline stuff is all in seconds so doing the conversion once on fetch eliminates a lot of downstream conversions.

packages/studio-base/src/components/EventsSyncAdapter.tsx
jtbandes: Ah I see, I was missing that setEvents is coming from the zustand store. I do think the useEffect approach you mentioned would be a little less error-prone for concurrent mode, but this is alright.

Issue #4310, opened 2022-08-30
User script panel does not resize properly


KerstinKeller:
**Description**
When you open a measurement, the measurement control opens in the bottom. However, other panels are not resized accordingly, so the control is on top of the other panel.

Resizing the whole Foxglove application fixes the problem.

- Version: 1.24.0
- OS: Windows

**Steps To Reproduce**
- Open Foxglove with a layout
- drop / open a measurement
- See how the measurement control is displayed on top of the widgets

https://user-images.githubusercontent.com/13848742/187442592-43340585-14ba-4eee-8bf5-5bfaaf634e61.mp4

foxymiles:
Thanks for the bug report. Have you see this happen with any panels besides the user scripts panel?

KerstinKeller:
I just tried the plot panel, which resizes ok. Maybe it's just the user scripts panel....

defunctzombie:
Renamed the issue because I am _reasonably_ sure that this is specific to the `user scripts` panel and cutting off the bottom bar in that panel. We'll get this fixed up.

@KerstinKeller as a related note - what you've called the "measurement control" we refer to as the "playback controls".

PR #4311, opened 2022-08-30
Fix user scripts panel underflow

foxymiles:
**User-Facing Changes**
Fixes an overflow in the user scripts panel that causes it to underflowing the timeline.

**Description**
Constrains the container stack of the user scripts panel to full height to prevent underflowing the timeline.

Fixes https://github.com/foxglove/studio/issues/4310

### DIFF
--- a/packages/studio-base/src/panels/NodePlayground/index.tsx
+++ b/packages/studio-base/src/panels/NodePlayground/index.tsx
@@ -335,7 +335,7 @@ function NodePlayground(props: Props) {
     <Stack fullHeight>
       <PanelToolbar helpContent={helpContent} />
       <Divider />
-      <Stack direction="row" fullHeight>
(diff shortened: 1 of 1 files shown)


defunctzombie (approved):

defunctzombie:
@foxymiles The description says this constrains the Stack to full height - but it already had the `fullHeight` attribute. Is it that it was allowing overflow? Won't this hidden overflow be equally cut-off even after this change?

foxymiles:
> @foxymiles The description says this constrains the Stack to full height - but it already had the `fullHeight` attribute. Is it that it was allowing overflow? Won't this hidden overflow be equally cut-off even after this change?

We need both `height: 100%` and `overflow: hidden` for the height limiting to have an effect. At some point I'd like to refactor all these nested stacks to use a a grid.

PR #4321, opened 2022-08-30
Support plot panel data series configuration in settings sidebar

foxymiles:
**User-Facing Changes**
None. This is under a feature flag for now.

**Description**
This is a proof of concept. I think in order to make this viable we need to find a way of making editing long message paths more ergonomic.

_Note_: Due to some optimizations in the plot panel legend, changes you make in the settings sidebar will not show up in the plot legend. This is fixable but I thought it wasn't worth spending the time until we decided if we like this approach or not.

Screen Shot 2022-08-31 at 1 53 16 PM

Fixes https://github.com/foxglove/studio/issues/3432

### DIFF
--- a/packages/studio-base/src/AppSetting.ts
+++ b/packages/studio-base/src/AppSetting.ts
@@ -20,6 +20,7 @@ export enum AppSetting {
   // Experimental features
   SHOW_DEBUG_PANELS = "showDebugPanels",
   ENABLE_LEGACY_PLOT_PANEL = "enableLegacyPlotPanel",
+  ENABLE_PLOT_PANEL_SERIES_SETTINGS = "enablePlotPathSeriesSettings",
 
   // Miscellaneous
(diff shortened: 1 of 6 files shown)


defunctzombie:
We should keep `user facing changes` to `None` in this PR. For now let's add an app setting to enable this for the plot panel so we can experiment with some ideas. Maybe we can graduate that into a panel setting once we are relatively happy and then transition away from the in-panel legend.

amacneil:
before merging this PR
- move X and Y axis above and collapse by default
- keep feature flag and hide all current legend functionality (legend position, show value) when the flag is enabled

foxymiles:
> before merging this PR
> 
> * move X and Y axis above and collapse by default
> * keep feature flag and hide all current legend functionality (legend position, show value) when the flag is enabled

I've made these changes.

defunctzombie (approved):

PR #4325, opened 2022-08-31
add url param &openIn=web/desktop to override launch preference

### LINKED ISSUE #3129: Add query param to force "open in desktop" or "open in web"
Sometimes users want to make links that force the desktop app to open. An example would be to open a websocket connection to a non-localhost domain (which won't work with https pages in the browser).

https://studio.foxglove.dev/?ds=foxglove-websocket&ds.url=ws://myurl:8765

Having a query argument that can bypass the launch preference and open the desktop app would allow for sharing such links.

jtbandes:
**User-Facing Changes**
Added a URL param `&openIn=[web/desktop]` to override the default behavior of asking the user whether to open urls in the desktop app.

**Description**
Resolves #3129

### DIFF
--- a/packages/studio-base/src/screens/LaunchPreference.tsx
+++ b/packages/studio-base/src/screens/LaunchPreference.tsx
@@ -4,6 +4,8 @@
 
 import { PropsWithChildren } from "react";
 
+import { LaunchPreferenceValue } from "@foxglove/studio-base/types/LaunchPreferenceValue";
+
 import { AppSetting } from "../AppSetting";
 import { useAppConfigurationValue } from "../hooks";
 import { useSessionStorageValue } from "../hooks/useSessionStorageValue";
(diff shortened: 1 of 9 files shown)


packages/studio-base/src/screens/LaunchingInDesktopScreen.tsx
```diff
@@ -13,7 +13,7 @@ export function LaunchingInDesktopScreen(): ReactElement {
   const [, setLaunchPreference] = useSessionStorageValue(AppSetting.LAUNCH_PREFERENCE);
 
   const cleanWebURL = new URL(window.location.href);
-  cleanWebURL.searchParams.delete("launch");
+  cleanWebURL.searchParams.delete("openIn");
```
jtbandes, line 16: Seems like we had an old version of this feature once?

amacneil:
- keep functionality in this PR to allow link author to force web vs desktop
- URL argument `open`
- rename `unknown` to `ask`
- change default user preference to `web` rather than `ask`

packages/studio-base/src/screens/LaunchPreference.tsx
```diff
@@ -4,29 +4,37 @@
...
+  const [globalLaunchPreference] = useAppConfigurationValue<string>(AppSetting.LAUNCH_PREFERENCE);
   const [sessionLaunchPreference] = useSessionStorageValue(AppSetting.LAUNCH_PREFERENCE);
 
   const url = new URL(window.location.href);
 
   // Session preferences take priority over global preferences.
-  const activePreference =
+  let activePreference =
     url.searchParams.get("openIn") ?? sessionLaunchPreference ?? globalLaunchPreference;
+  switch (activePreference) {
```
defunctzombie, line 25: Probably worth a quick comment that this sanitizes the value to a known set. I did a double take on the purpose of this switch to finally understand what it was doing.

defunctzombie (approved):
What's the benefit of having a `LaunchPreferenceValue` enum over union strings in typescript? We already get type validation if we pass an incorrect string. The updated use sites are longer now with no benefit over their string counterparts /shrug.

I am reminded of the discussion around constants for the various datatypes. Maybe this is different somehow from that.

foxymiles (approved):
I also prefer string unions over enums. Is there a specific advantage here of an enum?

jtbandes:
> We already get type validation if we pass an incorrect string.

Doesn't apply to cases where the parameter/return value is a `string`. So jest tests, `useSessionStorageValue`, `searchParams.get`...

> The updated use sites are longer now with no benefit over their string counterparts /shrug.

Slightly easier to find in the code?

jtbandes:
I can change it if we are 2 to 1 on this issue.

Issue #3710, opened 2022-06-24
Position-only follow mode


jhurliman:

jtbandes:
Do we have an idea of how important this is to our users?

defunctzombie:
Let's drop it - if we get requests to add it we can come back to it.

defunctzombie:
We got a request to add this so we've brought it back in scope.

defunctzombie:
Related: https://github.com/foxglove/studio/pull/3926

defunctzombie:
John: search for "unfollow pose snapshot"

PR #4329, opened 2022-08-31
Add Follow Modes To New 3D Panel Settings

snosenzo:
**User-Facing Changes**
Add follow modes: stationary , follow and follow-position follow modes to new 3D panel. This gives the users the ability to detach the camera from the current frame, and have it remain static as the frame changes.
 - "Pose" mode follows the orientation and positional changes of the frame
 - "Position" mode follows the frame position change, but the orientation remains static relative to the render frame
 - "None" follow mode keeps the camera position and orientation static

**Description**
 - created a `cameraGroup` on the renderer to hold the transformation back to the snapshot in the render frame. 
 - To make this work I had to remove the call to `controls.update()` while the stationary/follow-position modes are active because it fights the transforms. Doing this doesn't seem to affect any of the existing behavior that I've seen.

https://github.com/foxglove/studio/issues/3710


packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -255,8 +259,12 @@ export class Renderer extends EventEmitter<RendererEvents> {
 
   private perspectiveCamera: THREE.PerspectiveCamera;
   private orthographicCamera: THREE.OrthographicCamera;
+  private cameraGroup: THREE.Group;
```
jtbandes, line 262: Since the implementation of this feature is pretty finicky and hard to understand, I'd like to see a lot more comments on this variable and the various updated functions below explaining how it works. Will help future maintainers make changes!

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -893,6 +908,39 @@ export class Renderer extends EventEmitter<RendererEvents> {
...
+        fixedFrame,
+        currentTime,
+      );
+      if (!snapshotInRenderFrame) {
+        return;
+      }
+      if (this.followMode === "follow-position") {
+        // only make orientation static/stationary in this mode
+        // the position still follows the frame
+        this.cameraGroup.position.set(0, 0, 0);
```
jtbandes, line 928: How does this work? The comment says we are only making orientation static, but then the code is changing the position?

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -1140,6 +1188,22 @@ export class Renderer extends EventEmitter<RendererEvents> {
     }
 
     this.settings.errors.clearPath(FOLLOW_TF_PATH);
+
+    if (this.followMode !== "follow" && !this.unfollowPoseSnapshot) {
+      // Snapshot the current pose of the render frame in the fixed frame
```
jtbandes, line 1193: This works well while using the app, but since we don't remember this value across reloads, the behavior is a bit weird when you reload the app and the camera is no longer in the same position.

The way this used to work in the old3d panel was that in unfollow mode, `targetOffset` would simply be the camera's position in the fixed frame, so that was actually remembered in the layout... Not sure if doing something equivalent here would make sense?

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
snosenzo: Will update the comment, but essentially the application of the unfollowPoseSnapshot position and orientation components makes their position static relative to the fixed frame. So when the frame changes the angle of the camera relative to the scene will not change because only the snapshotPose orientation is applied

amacneil:
Can you add a present tense PR title - see `main` branch commit history for examples

defunctzombie:
Some things I noticed in the preview deployment:

- Switching from `Follow Position Only` to `Stationary` jumps my viewport to some different part of the scene. I would expect no change in my viewport when moving between these follow modes.
- Consider dropping "Only" from the `Follow Position Only` option
- The `Follow Mode` label should be `Follow mode` to follow the sentence case style we use throughout the app. (Display Frame is also incorrect and should be `Display frame`. https://github.com/foxglove/studio/issues/4335
- `Follow Position` should be `Follow position`
- Consider naming the modes `Pose`, `Position`, `Off`. I am not sure that this is better than `Follow`, `Follow position`, `Stationary` but worth considering.

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
snosenzo: Hope what I said is enough?

snosenzo: Hope what I said is enough?

snosenzo:
So where we've landed with this:
- When switching to any of the "follow" modes the targetOffset will be reset so that the user will be centered on the frame center. This is so that the user won't be thrown somewhere relative to the stationary follow mode target offset when switching to a followmode. 
- When switching to the stationary mode the user's target offset and camera center will stay the same
- In stationary mode upon reload it will only keep the same viewstate if playback has not occurred after switching to the current followmode. From talking with Jacob this seemed acceptable since it's not a huge use case and it's very hard to account for without a bunch more work.

packages/studio-base/src/dataSources/SampleNuscenesLayout.json
```diff
@@ -64,7 +64,7 @@
         "t:/tf"
       ],
       "expandedKeys": ["name:Topics"],
-      "followMode": "follow",
```
defunctzombie, line 67: The sample layout should not change since that uses the current 3d panel and not 3d beta.

benchmark/src/layouts.ts
```diff
@@ -180,7 +180,7 @@ const LAYOUTS = new Map<string, Layout>([
               customBackgroundColor: "#000000",
               diffModeEnabled: true,
               expandedKeys: ["name:Topics"],
-              followMode: "follow",
+              followMode: "follow-pose",
```
defunctzombie, line 183: Is this a layout for the old 3d panel? If so it shouldn't change?

defunctzombie (changes requested):
Double check the changes to sample layout and benchmark layouts. I would not have expected those to have any changes.

benchmark/src/layouts.ts
snosenzo: ah sorry I didn't realize this was pertaining to the old 3D panel. will change

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -739,7 +756,9 @@ export class Renderer extends EventEmitter<RendererEvents> {
   public setCameraState(cameraState: CameraState): void {
     this._isUpdatingCameraState = true;
     this._updateCameras(cameraState);
-    this.controls.update();
+    if (this.followMode === "follow-pose") {
+      this.controls.update();
```
jtbandes, line 760: comment? why do we only need this in follow-pose mode?

packages/studio-base/src/panels/ThreeDeeRender/renderables/CoreSettings.ts
```diff
@@ -281,6 +295,20 @@ export class CoreSettings extends SceneExtension {
...
         this.renderer.settings.errors.clearPath(["general", "followTf"]);
+      } else if (path[1] === "followMode") {
+        const followMode = value as FollowMode;
+        // Update the configuration. This is done manually since followMode is at the top level of
+        // config, not under `general`
+        this.renderer.updateConfig((draft) => {
+          // any follow -> stationary no clear
+          // stationary -> any follow clear offset (center on frame)
+          if (draft.followMode === "follow-none") {
+            draft.cameraState.targetOffset = [0, 0, 0];
```
jtbandes, line 306: also reset thetaOffset? And reset only thetaOffset when changing from position -> pose mode?

jtbandes (approved):

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -916,6 +935,51 @@ export class Renderer extends EventEmitter<RendererEvents> {
...
+
+    // If in stationary or follow-position modes
+    if (
+      this.followMode !== "follow-pose" &&
+      this.unfollowPoseSnapshot &&
+      renderFrame &&
+      fixedFrame
+    ) {
+      const snapshotInRenderFrame = renderFrame.applyLocal(
+        makePose(),
```
jhurliman, line 949: Avoid allocating a new pose every frame. Declare `const snapshotInRenderFrame = makePose();` once globally in this file, since it's a temporary that isn't used across multiple function calls.

jhurliman (approved):

PR #4336, opened 2022-09-01
Restack and style playback bar so Scrubber is above controls

2metres:
**User-Facing Changes**
Screen Shot 2022-09-01 at 3 42 31 pm

Screen Shot 2022-09-01 at 3 42 45 pm

**Description**
Moves scrubber bar above playback controls, removes background color from buttons and styles Hover timestamp like other tooltips.

### DIFF
--- a/packages/studio-base/src/components/LoopIcon.tsx
+++ b/packages/studio-base/src/components/LoopIcon.tsx
@@ -2,10 +2,13 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
-export default function LoopIcon({ strokeWidth = 1 }: { strokeWidth: number }): JSX.Element {
+import { SvgIcon, SvgIconProps } from "@mui/material";
+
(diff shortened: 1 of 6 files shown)


defunctzombie:
I like it.

Using it in the preview deployment felt natural and not a massive break from today's setup. One thought that I had while actually using it with some data is whether the _time_ should go in the center of the screen under the scrubber (where the play and seek buttons are). My high level ux idea here is that the current playback time is actually an important part of the debug and diagnostic process and by putting it in the middle we allow a user to have easier focus on it without other elements distracting.

2metres:
@defunctzombie Not sure I 100% understand what you asking for here. You get a timecode just above your cursor when hovering the the scrubber similar to what you are describing. IMO Adding the number in the middle will increase the height of the bar to be far too tall.

I worked through several arrangements and I think having the playback in the bottom left and the controls in the middle works best of all the variations I tried.

defunctzombie:
@2metres 

image

amacneil:
👌

defunctzombie (approved):

PR #4337, opened 2022-09-01
remove focus outline from canvas elements

foxymiles:
**User-Facing Changes**
None

**Description**
This globally removes the focus outline from canvas elements.


defunctzombie (approved):
Maybe we should do this for all <canvas> elements within our CSSBaseline? Image panel also suffers from this.

jtbandes (approved):

PR #4338, opened 2022-09-01
Remove zustand/context for panel settings editor context

foxymiles:
**User-Facing Changes**
None

**Description**
As of v4 it's no longer necessary to use the wrapped `zustand/context`. The store can be used directly as a value for a vanilla react context.

### DIFF
--- a/packages/studio-base/src/providers/PanelSettingsEditorContextProvider.tsx
+++ b/packages/studio-base/src/providers/PanelSettingsEditorContextProvider.tsx
@@ -2,25 +2,19 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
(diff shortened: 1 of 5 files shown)


packages/studio-base/src/providers/PanelSettingsEditorContextProvider.tsx
```diff
@@ -61,5 +53,11 @@ export function PanelSettingsEditorContextProvider({
 }: {
   children?: ReactNode;
 }): JSX.Element {
-  return <Provider createStore={createSettingsEditorStore}>{children}</Provider>;
+  const [store] = useState(createSettingsEditorStore());
```
jtbandes, line 56:
```suggestion
  const [store] = useState(() => createSettingsEditorStore());
```

jtbandes (approved):

Issue #4341, opened 2022-09-01
Inconsistent time formatting of start & end times


jtbandes:
When viewing this file: [arrows.mcap.zip](https://github.com/foxglove/studio/files/9473795/arrows.mcap.zip)

The sidebar shows the start time in seconds, but the end time as a date:

<img width="290" alt="image" src="https://user-images.githubusercontent.com/14237/188025252-8037ad57-d6e0-455c-930f-80b92290e3a2.png">

Expected behavior: both should be shown as seconds, because the file was produced using 0-based time and the total duration is <1 min.

Related: #4130

osaro-imohe:
@jtbandes Can I get this?

osaro-imohe:
Also, can I get some help setting up Foxglove Studio. I'm running into a generic network error when I try to run Fox Glove Studio web. It seems can't find some UI fonts.  

<img width="1440" alt="Screen Shot 2022-09-02 at 1 43 53 AM" src="https://user-images.githubusercontent.com/44953690/188035766-9a43dc6e-205d-4e21-b77e-79f04f2b8f4e.png">

foxymiles:
Closed by https://github.com/foxglove/studio/pull/4346

PR #4343, opened 2022-09-02
Fix issue #4341 -- Inconsistent time formatting of start & end times

osaro-imohe:
**User-Facing Changes**
Screen Shot 2022-09-02 at 10 30 42 AM

**Description**
I took a closer look and discovered that the start-time was being passed into the `Timestamp` component which contains a function for properly formatting absolute time:
```
if (!isAbsoluteTime(time)) {
    return (
      </Stack>
    );
  }
  ```

Plus it seemed inconsistent to use a ref for the endTime but not for startTime, so I scrapped that also.


osaro-imohe:
Closing this PR, found a better solution :)

Issue #4130, opened 2022-08-15
Inconsistent font size in data source sidebar


jtbandes:
Font sizes are inconsistent for start/end time and duration in the "Data source" sidebar:

<img width="154" alt="image" src="https://user-images.githubusercontent.com/14237/184696240-bfd88398-8ff2-4608-ac06-78b48db9ba13.png">

- Version: latest web
- OS: macOS Chrome
- Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native): local mcap file

jtbandes:
Note that sometimes the font family also differs. Maybe some styles are not getting applied depending on the order they change in. Perhaps a precedence issue with our style libraries

foxymiles:
Fixed in https://github.com/foxglove/studio/pull/4273

jtbandes:
Still occurs on main:

<img width="242" alt="image" src="https://user-images.githubusercontent.com/14237/187986255-c11c1795-530a-427b-8f0b-8504c0625d2f.png">

foxymiles:
@jtbandes Do you have any repro steps? I'm not seeing it here.

jtbandes:
Seems to only happen when the start time is shown as a duration. Try this file:

[arrows.mcap.zip](https://github.com/foxglove/studio/files/9473795/arrows.mcap.zip)

By the way, it seems like a bug that we show the start time as 0.0 but the end time as a date in 1969.

osaro-imohe:
My pending PR resolves this issue.

foxymiles:
Closed by https://github.com/foxglove/studio/pull/4346

PR #4344, opened 2022-09-02
Fix issues #4341 and #4130

osaro-imohe:
**User-Facing Changes**

Screen Shot 2022-09-02 at 11 57 27 AM

**Description**
I took a closer look and discovered that the start time was being passed into the Timestamp component which contains a function for properly formatting nonabsolute time:

```
if (!isAbsoluteTime(time)) {
    return (
      </Stack>
    );
  }
```

Initially, I scrapped the ref completely for endTime and used the `<TimeStamp />` component to keep it uniform with startTime. I noticed that would introduce performance loss in relation to @foxymiles improvements in #4273. So instead I decided to format it manually and append it to the ref's innerHTML as a walk-around.

```
endTimeRef.current.innerText = !isAbsoluteTime(endTime)
          ? `${formatTimeRaw(endTime)}`
          : `${date} ${formatTime(endTime)}`;
          
```

On further digging, I noticed that issue #4130 was also caused by the duration and endTime ref not being passed into the `<TimeStamp />` component. Applying the same Font Family to the duration and endTime ref fixed that.

```
<Typography
            className={classes.numericValue}
            fontFamily={fonts.MONOSPACE} // this was missing
            variant="inherit" // this was also missing
            ref={endTimeRef}
>
```

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
@@ -34,7 +34,7 @@ import { SceneExtension } from "./SceneExtension";
 import { ScreenOverlay } from "./ScreenOverlay";
 import { SettingsManager, SettingsTreeEntry } from "./SettingsManager";
 import { CameraState } from "./camera";
-import { stringToRgb } from "./color";
+import { DARK_OUTLINE, LIGHT_OUTLINE, stringToRgb } from "./color";
 import { FRAME_TRANSFORM_DATATYPES } from "./foxglove";
 import { DetailLevel, msaaSamples } from "./lod";
 import {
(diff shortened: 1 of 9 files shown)


packages/studio-base/src/components/DataSourceInfoView.tsx
```diff
@@ -74,7 +75,12 @@ function DataSourceInfoContent(props: {
         ) : startTime ? (
           <Timestamp horizontal time={startTime} />
         ) : (
-          <Typography className={classes.numericValue} variant="inherit" color="text.secondary">
+          <Typography
+            className={classes.numericValue}
+            fontFamily={fonts.MONOSPACE}
```
jtbandes, line 80: These fontFamily changes shouldn't be necessary, because classes.numericValue already includes the font setting (line 26)... Maybe we have some bug with CSS rule precedence?

osaro-imohe: Yup, looks like you're right. I just realized the reason for the different fonts was the `inherit` mission on line 39. NO need for the font family addition

PR #4345, opened 2022-09-02
3D beta: fix some settings bugs

### LINKED ISSUE #4284: New3d: Unsetting decay time does not work
**Description**

* Open a data set with point cloud data and enable the point cloud topic.
* Turn the decay time up to some number of seconds
* Play the data and let points accumulate
* Edit the decay time back to 0

Notice that the decay time does not reset to 0 and we are still seeing accumulated points. I would expect to no longer be accumulating points during playback.

jtbandes:
**User-Facing Changes**

**Description**
- Fixes #4284, which occurred because the old settings were being spread in with the new settings, rather than falling back to the defaults.
- Fixes a crash when scrubbing on the decay time field. Adds missing `min` to decay time setting.
- Fixes issues with some settings not updating the UI immediately (because the scene was re-rendering before the update was fully applied).

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
+++ b/packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
@@ -435,6 +435,10 @@ export function ThreeDeeRender({ context }: { context: PanelExtensionContext }):
   // Handle user changes in the settings sidebar
   const actionHandler = useCallback(
     (action: SettingsTreeAction) =>
+      // Wrapping in unstable_batchedUpdates causes React to run effects _after_ the handleAction
+      // function has finished executing. This allows scene extensions that call
+      // renderer.updateConfig to read out the new config value and configure their renderables
+      // before the render occurs.
       ReactDOM.unstable_batchedUpdates(() => renderer?.settings.handleAction(action)),
     [renderer],
   );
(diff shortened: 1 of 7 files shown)


packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
```diff
@@ -434,7 +434,12 @@ export function ThreeDeeRender({ context }: { context: PanelExtensionContext }):
 
   // Handle user changes in the settings sidebar
   const actionHandler = useCallback(
-    (action: SettingsTreeAction) => renderer?.settings.handleAction(action),
+    (action: SettingsTreeAction) =>
+      // Wrapping in unstable_batchedUpdates causes React to run effects _after_ the handleAction
+      // function has finished executing. This allows scene extensions that call
+      // renderer.updateConfig to read out the new config value and configure their renderables
+      // before the render occurs.
```
defunctzombie, line 441: Worth looking at a world as we evolve this code where data flows one way or where the parent (renderer) invokes renderables rather than both calling each other.

defunctzombie (approved):
Any tests for this stuff?

jtbandes:
I think we have tests for 3d panel settings that show what it's rendered with, but none that include user interaction around clicking the fields in the settings sidebar...

PR #4346, opened 2022-09-02
Bugfix 4341 4130 inconsistent time formatting

osaro-imohe:
**User-Facing Changes**

Screen Shot 2022-09-02 at 11 57 27 AM

**Description**
I took a closer look and discovered that the start time was being passed into the Timestamp component which contains a function for properly formatting nonabsolute time:

```
if (!isAbsoluteTime(time)) {
    return (
      </Stack>
    );
  }
```

Initially, I scrapped the ref completely for endTime and used the `<TimeStamp />` component to keep it uniform with startTime. I noticed that would introduce performance loss in relation to @foxymiles improvements in #4273. So instead I decided to format it manually and append it to the ref's innerHTML as a walk-around.

```
endTimeRef.current.innerText = !isAbsoluteTime(endTime)
          ? `${formatTimeRaw(endTime)}`
          : `${date} ${formatTime(endTime)}`;
          
```

On further digging, I noticed that issue #4130 was also caused by the `<TimeStamp />` component returning different typography for absolute time. Adding variant="inherit" to the absolute time typography fixed that.

```
<Typography
            className={classes.numericValue}
            variant="inherit" // this was also missing
            ref={endTimeRef}
>
```

### DIFF
--- a/packages/studio-base/src/components/DataSourceInfoView.tsx
+++ b/packages/studio-base/src/components/DataSourceInfoView.tsx
@@ -152,7 +152,7 @@ export function DataSourceInfoView(): JSX.Element {
       endTimeRef={endTimeRef}
       playerName={playerName}
       playerPresence={playerPresence}
-      startTime={endTime}
+      startTime={startTime}
     />
   );
 }
--- a/packages/studio-base/src/components/DataSourceInfoView.tsx
+++ b/packages/studio-base/src/components/DataSourceInfoView.tsx
@@ -75,12 +75,7 @@ function DataSourceInfoContent(props: {
         ) : startTime ? (
           <Timestamp horizontal time={startTime} />
         ) : (
-          <Typography
-            className={classes.numericValue}
-            fontFamily={fonts.MONOSPACE}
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/components/DataSourceInfoView.tsx
```diff
@@ -34,7 +35,7 @@ const selectPlayerPresence = ({ playerState }: MessagePipelineContext) => player
 
 function DataSourceInfoContent(props: {
   durationRef: MutableRefObject<ReactNull | HTMLDivElement>;
-  endTimeRef: MutableRefObject<ReactNull | HTMLDivElement>;
+  endTimeRef?: MutableRefObject<ReactNull | HTMLDivElement>;
```
foxymiles, line 38: Why does this need to be made an optional?

osaro-imohe: Good point, I've made it mandatory

foxymiles (approved):
Looks good. Thanks for the fix!

Issue #4088, opened 2022-08-10
Playhead is not at the beginning when loading a bag


jtbandes:
**Description**
The playhead appears slightly offset from the beginning of the bag when loading a bag, although the player starts paused and no data is displayed.

<img width="146" alt="image" src="https://user-images.githubusercontent.com/14237/183790462-0a4648c9-f0d6-4441-a83f-bc1d543d5fa9.png">

- Version: latest web
- OS: Chrome macOS
- Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native): https://librealsense.intel.com/rs-tests/TestData/stairs.bag

jtbandes:
Downgrading to UX. Sounds like the time displayed on the playhead is accurate.

defunctzombie:
This behavior is by design. The intent of this feature is to make an effort to read some messages from the file so the various panels in the UI display some data rather than sitting empty and waiting for the user to start playing before any data is loaded since rarely are all the desired messages at the exact start time of the file (usually only one or two are exactly at the start).

I'm going to close this as not planned. I'll keep an eye out for any user's that surface this as confusing behavior and if we see that we can consider revisiting.

PR #4347, opened 2022-09-04
Fix bug #4088 set Playhead to start on .bag load

osaro-imohe:
**User-Facing Changes**

https://user-images.githubusercontent.com/44953690/188307222-da2f6183-f6e8-4baa-99a0-181539311c35.mov

**Description**
It seems this was a `bug` not a `UX` issue. 

When Studio loads a bag file for the first time it reads a small amount of data from the data source with the hope of producing a message or two.
Without an initial read, the user would be looking at a blank layout since no messages have yet been delivered.
 

Studio was setting the currentTime to the stopTime from the preload, which was causing the slider to jump ahead a little.

You can read more about it at IterablePlayer.ts on line 519

### DIFF
--- a/packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
+++ b/packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
@@ -576,7 +576,6 @@ export class IterablePlayer implements Player {
       clearTimeout(tickTimeout);
     }
 
-    this._currentTime = stopTime;
     this._messages = messageEvents;
     this._presence = PlayerPresence.PRESENT;
     this._queueEmitState();


defunctzombie:
Which issue is this trying to fix? The behavior we have when opening a new bag file and reading a few milliseconds into the file is by design and allows us to have some data to display for the user. If we did not do this loading for local files then we would not have any messages to display for the user in their panels.

We might reconsider this default one day but for now we are ok with the tradeoff.

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
```diff
@@ -576,7 +576,6 @@ export class IterablePlayer implements Player {
       clearTimeout(tickTimeout);
     }
 
-    this._currentTime = stopTime;
```
defunctzombie, line 579: This violates playback assumptions. The _currentTime should never be before the time of messages sent to the message pipeline. Think of the currentTime as the playhead time - it would be incorrect to produce messages ahead of the playhead time since that would make it appear like those messages were received before the playhead when actually they were not.

jtbandes:
> Which issue is this trying to fix?

It's in the title. #4088.

Issue #4234, opened 2022-08-23
Sync light/dark mode preference across studio and console


amacneil:
Foxglove Studio and Console currently have independent light/dark modes. We should sync this state on web.

Perhaps this is also the time to make "follow system" the default across both tools. We should do that after new 3D panel reaches GA.

jtbandes:
For whoever works on this, reminder that the place to store this kind of info in Studio is the User Profile: https://github.com/foxglove/studio/blob/main/packages/studio-base/src/providers/UserProfileLocalStorageProvider.tsx

Right now it’s stored in localStorage only; we’d probably want to keep the localStorage piece as a cache when we add syncing, so the app doesn’t flash the wrong color scheme every time you load it.

Issue #4232, opened 2022-08-23
Rename new panel to 3D and rename old panel to 3D (Legacy)


defunctzombie:
We should also update both the _default layout_ and _sample nuscenes layout_ to use the new 3d panel in the same PR - to avoid showing sample data with a panel that says "legacy".

PR #4357, opened 2022-09-06
Use system color scheme instead of dark theme by default

jtbandes:
**User-Facing Changes**
The system's color scheme is now used by default instead of dark mode. The color scheme can be changed in the Preferences sidebar.

**Details**
Related: https://github.com/foxglove/studio/issues/4234


amacneil (approved):
Wait until after https://github.com/foxglove/studio/issues/4232 merges

PR #4358, opened 2022-09-07
3D Beta: add onboarding tooltip on settings button

### LINKED ISSUE #4354: Help text for moved topic selection list
Once we rename 3d (beta) -> 3d and 3d -> 3d (legacy), users that were previously familiar with adding a 3d panel will be surprised to learn their panel is now missing the “topic tree” icon and selection. We experienced the same feedback when we originally moved the map and image panel items under the settings cog; users were wondering what happened to the settings.

The 3d panel is a central workflow for many of our users we should provide a better transition so they don’t feel that something is broken or missing in their 3d panel.

jtbandes:
**User-Facing Changes**
None (3d beta, self-explanatory feature)

**Description**
Adds a tooltip to help users discover topic settings, which are in a significantly different location from the old panel.

The tooltip is shown on any new 3D panel, until the first time the user clicks the settings button. The clicked state is stored in the user profile (in localStorage).

Also updates thumbnail image and description for 3D beta panel

Closes #4354

Screen Shot 2022-09-06 at 4 21 50 PM

### DIFF
--- a/packages/studio-base/src/components/PanelToolbar/index.stories.tsx
+++ b/packages/studio-base/src/components/PanelToolbar/index.stories.tsx
@@ -16,7 +16,6 @@ import { Box } from "@mui/material";
 import { storiesOf } from "@storybook/react";
 import { Mosaic, MosaicWindow } from "react-mosaic-component";
 
-import ChildToggle from "@foxglove/studio-base/components/ChildToggle";
 import MockPanelContextProvider from "@foxglove/studio-base/components/MockPanelContextProvider";
 import ToolbarIconButton from "@foxglove/studio-base/components/PanelToolbar/ToolbarIconButton";
 import MockCurrentLayoutProvider from "@foxglove/studio-base/providers/CurrentLayoutProvider/MockCurrentLayoutProvider";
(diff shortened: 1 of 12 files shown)


defunctzombie (approved):

defunctzombie:
Tried this on the preview. 👏

2metres:
Tweaked to "info" blue: 
Screen Shot 2022-09-07 at 8 32 06 pm

packages/studio-base/src/components/PanelToolbar/index.stories.tsx
```diff
@@ -84,29 +83,12 @@ class PanelToolbarWithOpenMenu extends React.PureComponent {
...
           </div>
-          <KeepToolbarVisibleHack />
         </PanelToolbar>
       </div>
     );
   }
 }
 
-// Keep PanelToolbar visible by rendering an empty ChildToggle inside the toolbar
-function KeepToolbarVisibleHack() {
```
snosenzo, line 95: what in this PR causes this to no longer be an issue? or is this just no longer relevant

jtbandes: A while ago we removed the behavior of auto hiding toolbars.

packages/studio-base/src/panels/index.ts
```diff
@@ -61,10 +62,11 @@ const builtin: PanelInfo[] = [
   {
     title: "3D (Beta)",
     type: "3D",
-    description: "Test our next-generation 3D panel, using a new rendering engine.",
+    description: "Display markers, camera images, meshes, URDFs, and more in a 3D scene.",
     help: ThreeDeeRenderHelp,
-    thumbnail: threeDimensionalVizThumbnail,
+    thumbnail: threeDeeRenderThumbnail,
     module: async () => await import("./ThreeDeeRender"),
+    settingsOnboardingTooltip: "Configure topics to visualize in the panel settings.",
```
amacneil, line 69:
```suggestion
    settingsOnboardingTooltip: "Open settings to configure topics and layers.",
```

amacneil:
Feedback from product review:
- add `x` to dismiss popper
- need to consider interaction with welcome dialog: probably it should be hidden until welcome dialog is dismissed (or as a fallback, only show after the 3d panel is clicked on somehow)
- consider white bg or something more neutral

PR #4372, opened 2022-09-08
hide settings tooltip under modal

jtbandes:
**User-Facing Changes**
None

**Description**
Follow up to #4358


jtbandes:
Whoever gets around to reviewing this, feel free to merge.

amacneil:
The sibling selector doesn't seem to be working for me (even when popper comes after modal):

image

Works in safari as expected. It would be nice if we can get the sibling selector working in at least one direction (it covers the most common case when you first open the app). If we can't figure that out then just remove it because it's not doing anything.

defunctzombie (approved):

PR #4376, opened 2022-09-08
Update default 3d camera position

amacneil:
**User-Facing Changes**
Zoom in the 3d camera and angle it more by default. Personal preference based on some tweaks and assuming that many common robots are even smaller than cars.

Bonus: it makes the TF labels possible to read.

**Before**
image
image

**After**
image
image

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/camera.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/camera.ts
@@ -30,13 +30,13 @@ export type CameraState = {
 };
 
 export const DEFAULT_CAMERA_STATE: CameraState = {
-  distance: 75,
+  distance: 20,
   perspective: true,
-  phi: 45,
+  phi: 60,
   target: [0, 0, 0],
   targetOffset: [0, 0, 0],
   targetOrientation: [0, 0, 0, 1],
-  thetaOffset: 0,
+  thetaOffset: 45,
   fovy: 45,
   near: 0.01,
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/panels/ThreeDeeRender/camera.ts
```diff
@@ -30,13 +30,13 @@ export type CameraState = {
...
-  distance: 75,
+  distance: 20,
   perspective: true,
-  phi: 45,
+  phi: 60,
   target: [0, 0, 0],
   targetOffset: [0, 0, 0],
   targetOrientation: [0, 0, 0, 1],
-  thetaOffset: 0,
+  thetaOffset: 45,
```
snosenzo, line 39: I don't think there's that much added benefit setting this to anything because it's just the rotation around the central axis and is going to change a lot across different files.

jtbandes: I guess from the screenshots shown, it does show a bit more depth because it makes the target frame's x axis point into the screen rather than just to the right

jtbandes (approved):

packages/studio-base/src/panels/ThreeDeeRender/camera.ts
amacneil: exactly - I felt the rotated perspective was more obviously a 3d scene (especially when adding a grid, which we want to do by default)

Issue #4352, opened 2022-09-06
Enabling loop doesn't restart playback...sometimes


jtbandes:
**Description**

https://user-images.githubusercontent.com/14237/188711205-ecf74896-15f9-4b80-9ee5-cc1ebb33d528.mov

- Version: latest web
- OS: macos
- Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native): `nuScenes-v1.0-mini-scene-0061.bag`

**Steps To Reproduce**
- Create an empty layout with a 3D Beta panel (no topics shown, only TF)
- Load the bag file
- Play to the end, and observe playback pauses automatically when the end is reached
- Click the 🔁 button to enable looping, observe playback restarts at the beginning of the bag
- Enable all topics in the 3D panel
- Disable loop and wait for playback to pause again
- Click 🔁 button to enable loop and playback does **not** restart at the beginning, and the ⏯️ button does not work

**Expected Behavior**
Enabling 🔁 should restart playback from the beginning if it previously auto-paused by reaching the end of the bag.

**Actual Behavior**
Playback only restarts sometimes (e.g. if all topics are turned on)

PR #4379, opened 2022-09-08
Fix click-to-loop playback behavior

defunctzombie:
**User-Facing Changes**
Looping at the end of a file is more reliable.

**Description**
Looping behavior is handled by the RepeatAdapter component. This component has a layoutEffect that runs when active data is changed and checks if playback needs to restart. To restart playback it calls seek followed by play.

The iterable player receives the seek request and begins a seek backfill to load messages for the seek time. While it does this, if the layout effect in the adapter runs again (possibly due to a previous state emit finishing), then the logic to seek to the start is run again. The iterable player is again told to seek to the start which interrupts the previous unfinished seek and the cycle repeats. This leads to the player never actually seeking to the start.

This change updates the IterablePlayer to check if the requested seek time matches the current targetted seek time and avoid entering another seek state and letting the current one finish.

Fixes: #4352


defunctzombie:
@jtbandes give this a spin and see if it resolves the loop issue you reported.

jtbandes (approved):
Can no longer reproduce the issue on the preview deployment 👍

PR #4383, opened 2022-09-08
Enable source map generation in ts-loader compilerOptions

### LINKED ISSUE #4382: Source maps not working properly
Something has regressed about our source maps. In a dev build, sources in the chrome dev tools look like this:

image

On studio.foxglove.dev they look like this:

image

Neither of these actually looks like the original source file.

jtbandes:
**User-Facing Changes**
None

**Description**
Our packages/studio-base/tsconfig.json has source maps disabled. This is probably because we don't want to bother outputting source maps in the published studio-base package. However, we do want them enabled for Studio builds.

Fixes #4382


desktop/integration-test/buildSize.test.ts
```diff
@@ -7,7 +7,7 @@ import * as path from "path";
 
 // Adjust this byte size as needed if the app is growing for a valid reason.
 // This check is here to catch unexpected ballooning of the build size
-const MAX_BUILD_SIZE = 150_000_000;
+const MAX_BUILD_SIZE = 160_000_000;
```
defunctzombie, line 10: :(

package.json
```diff
@@ -57,7 +57,6 @@
   },
   "resolutions": {
     "@types/node": "15.3.0",
-    "esbuild-loader/esbuild": "0.14.7",
```
defunctzombie, line 60: Do we no longer want to force a specific esbuild anymore?

defunctzombie (approved):

desktop/integration-test/buildSize.test.ts
jtbandes: :)

package.json
jtbandes: Unclear why this was here? Things seem to work when I remove it and upgrade packages.

jtbandes: https://github.com/foxglove/studio/pull/2480

PR #4385, opened 2022-09-08
update Raw Message to link to foxglove docs for foxglove datatypes

jtbandes:
**User-Facing Changes**
When using Foxglove schemas, Raw Messages panel now includes a link to the schema documentation.

**Description**
I thought we had a ticket for this but I can't find it...

### DIFF
--- a/packages/studio-base/src/panels/RawMessages/utils.ts
+++ b/packages/studio-base/src/panels/RawMessages/utils.ts
@@ -11,12 +11,13 @@
 //   found at http://www.apache.org/licenses/LICENSE-2.0
 //   You may not use this file except in compliance with the License.
 
-import { first, last } from "lodash";
+import { first, kebabCase, last } from "lodash";
 
+import { foxgloveMessageSchemas } from "@foxglove/schemas";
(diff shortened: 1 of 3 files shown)


defunctzombie (approved):
👍 I'd love to see utils.ts renamed to something better but that might be too much to ask :D

packages/studio-base/src/panels/RawMessages/utils.ts
```diff
@@ -98,11 +99,27 @@ export function getChangeCounts(
...
+  if (pkg != undefined && ROS_COMMON_MSG_PACKAGES.has(pkg)) {
+    return `https://docs.ros.org/api/${pkg}/html/msg/${filename}.html`;
+  }
+
+  const foxgloveDocsLink = foxgloveDocsLinksByDatatype.get(datatype);
+  if (foxgloveDocsLink != undefined) {
+    return foxgloveDocsLink;
+  }
+
+  return `https://www.google.com/search?q=${datatype}`;
```
defunctzombie, line 124: I think we should return undefined and not have any link in this case. It doesn't seem so useful to do a google search on people's custom message types.

PR #4386, opened 2022-09-08
[3D] Preload transform topics

jhurliman:
**User-Facing Changes**

- [3D] All transforms are preloaded from local and remote file sources

**Description**

Renderer#addDatatypeSubscriptions() and Renderer#addTopicSubscriptions() has been changed to optionally accept a configuration object specifying forced and/or preload. This simplifies the subscription handling code and allows the preload boolean to be passed through to the extension API. Transform (i.e. "/tf") topics are now preloaded, and Renderer#transformTree is not flushed on seek/scrub. This should improve the usability of preload-enabled data sources by always loading the full set of coordinate frames and transforms into the UI.

Note that for most files, this will have the side effect of loading and decompressing every chunk from disk or network since `/tf` messages tend to be small and frequent, interspersed through every chunk in the recording.

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
@@ -157,8 +157,15 @@ export type RendererConfig = {
 export type MessageHandler<T = unknown> = (messageEvent: MessageEvent<T>) => void;
 
 export type RendererSubscription<T = unknown> = {
+  /** Preload the full history of topic messages as a best effort */
   preload?: boolean;
+  /**
+   * By default, topic subscriptions are only created when the topic visibility
+   * has been toggled on by the user in the settings sidebar. Enabling forced
+   * will unconditionally create the topic subscription(s)
+   */
   forced?: boolean;
+  /** Callback that will be fired for each matching incoming message */
   handler: MessageHandler<T>;
 };
 
--- a/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
@@ -68,7 +68,7 @@ import {
   TRANSFORM_STAMPED_DATATYPES,
   Vector3,
 } from "./ros";
-import { BaseSettings, CustomLayerSettings, SelectEntry, SubscriptionType } from "./settings";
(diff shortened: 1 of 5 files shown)


packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -248,14 +254,10 @@ export class Renderer extends EventEmitter<RendererEvents> {
...
-  // datatype -> handler[], always active
-  public forcedDatatypeHandlers = new Map<string, MessageHandler[]>();
-  // topicName -> handler[], only active when visibility is toggled on
-  public topicHandlers = new Map<string, MessageHandler[]>();
-  // topicName -> handler[], always active
-  public forcedTopicHandlers = new Map<string, MessageHandler[]>();
+  // datatype -> RendererSubscription[]
+  public datatypeHandlers = new Map<string, RendererSubscription[]>();
+  // topicName -> RendererSubscription[]
+  public topicHandlers = new Map<string, RendererSubscription[]>();
```
jtbandes, line 260: Is it true that there are no forced topic subscriptions right now? What would it mean to force a topic subscription?

Also I don't really get what a _non-forced_ datatype subscription is. Wouldn't a datatype subscription mean the same thing as subscribing to all topics of a given datatype?

jhurliman: There are no public forced topic subscriptions currently, only in private user extensions.

Datatype subscriptions are expanded out into topic subscriptions before considering what to do with them, so yes a datatype subscription means subscribing to all topics of a given datatype (without having to monitor what that set of topics is).

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -154,7 +154,13 @@ export type RendererConfig = {
 };
 
 /** Callback for handling a message received on a topic */
-export type MessageHandler = (messageEvent: MessageEvent<unknown>) => void;
+export type MessageHandler<T = unknown> = (messageEvent: MessageEvent<T>) => void;
+
+export type RendererSubscription<T = unknown> = {
+  preload?: boolean;
+  forced?: boolean;
```
jtbandes, line 161: add some comments to explain what these do

jtbandes (approved):

Issue #4389, opened 2022-09-09
Support opening unindexed mcap files


defunctzombie:
This regressed when I moved over to the new player architecture.

PR #4391, opened 2022-09-09
Add support for reading unindexed mcap files

defunctzombie:
**User-Facing Changes**
Users can open unindexed mcap files.

**Description**
This brings back the same level of support for unindexed mcap files that we had in the previous player architecture. The unindexed mcap file is fully read into memory. For now this is the only way to process an mcap file with no index or summary section because the IIterableSource can only provide topic information and datatypes during initialization not during message reading.

Future enhancements could include:
- Reading the summary section if it contains schemas and channels only thus avoiding the need to read the message data into memory.
- Adding a way for IIterableSource to provide updates to topics and datatypes as messages are read.

Fixes: #4389

### DIFF
--- a/packages/studio-base/src/players/IterablePlayer/Mcap/McapStreamingIterableSource.ts
+++ b/packages/studio-base/src/players/IterablePlayer/Mcap/McapStreamingIterableSource.ts
@@ -1,0 +1,241 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { Mcap0StreamReader, Mcap0Types } from "@mcap/core";
+import { isEqual } from "lodash";
+
+import { loadDecompressHandlers, parseChannel, ParsedChannel } from "@foxglove/mcap-support";
+import {
+  Time,
+  isLessThan,
+  isGreaterThan,
+  isTimeInRangeInclusive,
+  fromNanoSec,
+  subtract,
+  toSec,
+} from "@foxglove/rostime";
+import { Topic, MessageEvent } from "@foxglove/studio";
+import {
+  GetBackfillMessagesArgs,
+  IIterableSource,
+  Initalization,
+  IteratorResult,
+  MessageIteratorArgs,
+} from "@foxglove/studio-base/players/IterablePlayer/IIterableSource";
+import { PlayerProblem, TopicStats } from "@foxglove/studio-base/players/types";
+import { RosDatatypes } from "@foxglove/studio-base/types/RosDatatypes";
+
+type Options = { size: number; stream: ReadableStream<Uint8Array> };
+
+export class McapStreamingIterableSource implements IIterableSource {
+  private options: Options;
+  private msgEventsByChannel?: Map<number, MessageEvent<unknown>[]>;
+  private start?: Time;
+  private end?: Time;
+
+  public constructor(options: Options) {
+    this.options = options;
+  }
+
+  public async initialize(): Promise<Initalization> {
+    if (this.options.size > 1024 * 1024 * 1024) {
+      // This provider uses a simple approach of loading everything into memory up front, so we
+      // can't handle large files
+      throw new Error("Unable to stream MCAP file; too large");
+    }
+    const decompressHandlers = await loadDecompressHandlers();
+
+    const streamReader = this.options.stream.getReader();
+
+    const problems: PlayerProblem[] = [];
+    const channelIdsWithErrors = new Set<number>();
+
+    const messagesByChannel = new Map<number, MessageEvent<unknown>[]>();
+    const schemasById = new Map<number, Mcap0Types.TypedMcapRecords["Schema"]>();
+    const channelInfoById = new Map<
+      number,
+      { channel: Mcap0Types.Channel; parsedChannel: ParsedChannel }
+    >();
+
+    let startTime: Time | undefined;
+    let endTime: Time | undefined;
+    let profile: string | undefined;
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/players/IterablePlayer/Mcap/McapStreamingIterableSource.ts
```diff
@@ -0,0 +1,246 @@
...
+        topicStats.set(channel.topic, { numMessages });
+      }
+      // Final datatypes is an unholy union of schemas across all channels
+      for (const [name, datatype] of parsedChannel.datatypes) {
+        datatypes.set(name, datatype);
+      }
+    }
+
+    this.start = startTime ?? { sec: 0, nsec: 0 };
+    this.end = endTime ?? { sec: 0, nsec: 0 };
```
jtbandes, line 180: looks like this would only happen if the file has no messages. What happens in the case of an empty file? Do we allow the user to load the file or do we display some error?

packages/studio-base/src/players/IterablePlayer/Mcap/McapStreamingIterableSource.ts
```diff
@@ -0,0 +1,246 @@
...
+
+    const fileDuration = toSec(subtract(this.end, this.start));
+    if (fileDuration > 2.628e6) {
```
jtbandes, line 183: What is this number? Prefer more readable constants such as `20*365*24*60*60` (or use `DURATION_20_YEARS_SEC` which we already export somewhere in the codebase)

packages/studio-base/src/players/IterablePlayer/Mcap/McapStreamingIterableSource.ts
```diff
@@ -0,0 +1,246 @@
...
+      problems.push({
+        message: "This file has an abnormally long duration.",
+        tip: "Check the start and end time.",
```
jtbandes, line 186: Include more details in the error message, i.e. the actual start and end time, to aid in debugging

packages/studio-base/src/players/IterablePlayer/Mcap/McapStreamingIterableSource.ts
```diff
@@ -0,0 +1,246 @@
...
+        severity: "warn",
+      });
+    }
+
+    problems.push({
+      message: "This file unindexed. Unindexed files may have degraded performance.",
```
jtbandes, line 192: Not a super actionable error message, can we point users to the `mcap reindex` command (or some docs)?

packages/studio-base/src/players/IterablePlayer/Mcap/McapStreamingIterableSource.ts
```diff
@@ -0,0 +1,246 @@
...
+    const topicsSet = new Set(topics);
+
+    for (const msgEvents of this.msgEventsByChannel.values()) {
+      for (const msgEvent of msgEvents) {
+        if (
+          isTimeInRangeInclusive(msgEvent.receiveTime, start, end) &&
+          topicsSet.has(msgEvent.topic)
+        ) {
+          yield {
+            connectionId: undefined,
```
jtbandes, line 232: why undefined? we should be able to pass the connection id from the key of msgEventsByChannel, right? (I don't remember what this gets used for in the app?)

packages/studio-base/src/players/IterablePlayer/Mcap/McapStreamingIterableSource.ts
```diff
@@ -0,0 +1,246 @@
...
+        }
+      }
+    }
+  }
+
+  public async getBackfillMessages(
+    _args: GetBackfillMessagesArgs,
+  ): Promise<MessageEvent<unknown>[]> {
+    return [];
+  }
```
jtbandes, line 245: Are we planning to improve this? Maybe file a feature request ticket? If we are loading all messages into memory we should be able to support backfill.

jtbandes (approved):

packages/studio-base/src/players/IterablePlayer/Mcap/McapStreamingIterableSource.ts
defunctzombie: Updated with the channelId

packages/studio-base/src/players/IterablePlayer/Mcap/McapStreamingIterableSource.ts
defunctzombie: Added a basic implementation. For some portion of the file this is already handled by the caching iterator layer above but still good to have a fallback.

packages/studio-base/src/players/IterablePlayer/Mcap/McapStreamingIterableSource.ts
defunctzombie: We don't have `mcap reindex` but I did link to the summary section of the docs.

packages/studio-base/src/players/IterablePlayer/Mcap/McapStreamingIterableSource.ts
defunctzombie: Added the start/end time.

packages/studio-base/src/players/IterablePlayer/Mcap/McapStreamingIterableSource.ts
defunctzombie: Updated

packages/studio-base/src/players/IterablePlayer/Mcap/McapStreamingIterableSource.ts
defunctzombie: In the case of no messages we don't do anything special other than still tell the user the file is unindexed.

Issue #4207, opened 2022-08-19
migrate components/Tooltip to MUI


defunctzombie:

defunctzombie:
@2metres to confirm that this is done with the exception of old3d panel which we will get for free when we delete it

PR #4392, opened 2022-09-09
Convert playback controls tooltip to MUI

foxymiles:
**User-Facing Changes**
Minor visual changes.

**Description**
Convert the playback controls tooltip to the MUI tooltip.

Screen Shot 2022-09-09 at 7 58 53 AM

Note: This looks a bit different. We can of course style it more like the existing tooltip but maybe it's better to stick to the standard MUI tooltip styling?

Partially addresses https://github.com/foxglove/studio/issues/4207

### DIFF
--- a/packages/studio-base/src/components/PlaybackControls/PlaybackControlsTooltipContent.tsx
+++ b/packages/studio-base/src/components/PlaybackControls/PlaybackControlsTooltipContent.tsx
@@ -1,0 +1,122 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { Divider, Typography } from "@mui/material";
+import { isEmpty } from "lodash";
+import { Fragment } from "react";
+import { makeStyles } from "tss-react/mui";
+
+import { subtract as subtractTimes, toSec, fromSec } from "@foxglove/rostime";
+import {
+  MessagePipelineContext,
+  useMessagePipeline,
+} from "@foxglove/studio-base/components/MessagePipeline";
+import {
+  TimelineInteractionStateStore,
+  useTimelineInteractionState,
+} from "@foxglove/studio-base/context/TimelineInteractionStateContext";
+import { useAppTimeFormat } from "@foxglove/studio-base/hooks";
+import { fonts } from "@foxglove/studio-base/util/sharedStyleConstants";
+
+export type PlaybackControlsTooltipItem =
+  | { type: "divider" }
+  | { type: "item"; title: string; value: string };
+
+const useStyles = makeStyles()((theme) => ({
+  tooltipDivider: {
+    gridColumn: "span 2",
+    marginBlock: theme.spacing(0.5),
+    opacity: 0.5,
+  },
+  tooltipWrapper: {
+    fontFeatureSettings: `${fonts.SANS_SERIF_FEATURE_SETTINGS}, "zero"`,
+    fontFamily: fonts.SANS_SERIF,
+    whiteSpace: "nowrap",
+    columnGap: theme.spacing(0.5),
+    display: "grid",
+    alignItems: "center",
+    gridTemplateColumns: "auto 1fr",
+    flexDirection: "column",
+  },
+  itemKey: {
+    fontSize: "0.7rem",
+    opacity: 0.7,
+    textAlign: "end",
+    textTransform: "lowercase",
+  },
+}));
+
+const selectHoveredEvents = (store: TimelineInteractionStateStore) => store.eventsAtHoverValue;
+const selectStartTime = (ctx: MessagePipelineContext) => ctx.playerState.activeData?.startTime;
+
+export function PlaybackControlsTooltipContent(params: {
+  hoverXPosition: number;
+}): ReactNull | JSX.Element {
+  const { hoverXPosition } = params;
+  const { formatTime, timeFormat } = useAppTimeFormat();
+  const hoveredEvents = useTimelineInteractionState(selectHoveredEvents);
+  const startTime = useMessagePipeline(selectStartTime);
+  const { classes } = useStyles();
+
+  if (!startTime) {
+    return ReactNull;
+  }
+
+  const stamp = fromSec(hoverXPosition);
+  const timeFromStart = subtractTimes(stamp, startTime);
+
+  const tooltipItems: PlaybackControlsTooltipItem[] = [];
+
+  if (!isEmpty(hoveredEvents)) {
+    Object.values(hoveredEvents).forEach(({ event }) => {
+      tooltipItems.push({
+        type: "item",
+        title: "Start",
+        value: formatTime(event.startTime),
+      });
+      tooltipItems.push({
+        type: "item",
+        title: "End",
+        value: formatTime(event.endTime),
+      });
+      if (!isEmpty(event.metadata)) {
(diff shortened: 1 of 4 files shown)


packages/studio-base/src/components/PlaybackControls/Scrubber.tsx
```diff
@@ -208,54 +115,72 @@ export default function Scrubber(props: Props): JSX.Element {
...
-          onHoverOver={onHoverOver}
-          onHoverOut={onHoverOut}
-          onChange={onChange}
-          renderSlider={renderSlider}
-        />
+      <Stack
+        ref={hoverElRef}
+        direction="row"
+        flexGrow={1}
+        onPointerMove={handlePointerMove}
```
defunctzombie, line 159: The Tooltip component doesn't handle this for us around the stack? Or is it that it does not update as the user moves the cursor and we want it to update in this case?

Maybe worth making a component like <ContinuousTooltip> ? My main thought is around how to make it more explicit what the intent here was and why the existing tooltip was insufficient.

defunctzombie (approved):

packages/studio-base/src/components/PlaybackControls/Scrubber.tsx
foxymiles: Yeah the MUI tooltip out of the box doesn't support dynamic positioning like this. It wants to be in a fixed position relative to the anchor as dictated by the `placement` property. This kind of tracking tooltip seems rare in our app and a bit difficult to extract without passing a lot of refs around. Do you think it's worth trying to generalize?

packages/studio-base/src/components/PlaybackControls/PlaybackControlsTooltipContent.tsx
```diff
@@ -0,0 +1,122 @@
...
+  },
+  itemKey: {
+    fontSize: "0.7rem",
+    opacity: 0.7,
+    textAlign: "end",
+    textTransform: "lowercase",
+  },
+}));
+
+const selectHoveredEvents = (store: TimelineInteractionStateStore) => store.eventsAtHoverValue;
```
snosenzo, line 50: probably beyond the scope of this PR but Any way we can better differentiate between a DOM `Event` and a Timeline Event? or maybe rename Events to Marks or Tags or something b/c not all of these things are necessarily events right?

foxymiles: The term `Event` is used pretty heavily through the UI and the documentation so I think it would probably be more confusing to call it something completely different. I've labeled as a `ConsoleEvent` in the API code so it doesn't directly clash with the native DOM `Event`. We could include that prefix in all these but that would lead to some pretty unwieldy names like `selectHoveredConsoleEvent` and `TimelinePositionedConsoleEvent`.

packages/studio-base/src/components/PlaybackControls/Scrubber.tsx
defunctzombie: I wouldn't say it needs to generalize a lot but I think maybe having a component in the PlaybackControls folder might help readability. Your call.

packages/studio-base/src/components/PlaybackControls/Scrubber.tsx
```diff
@@ -208,54 +115,72 @@ export default function Scrubber(props: Props): JSX.Element {
...
+  const popperRef = React.useRef<Instance>(ReactNull);
+
+  const positionRef = React.useRef({ x: 0, y: 0 });
+
+  const handlePointerMove = (event: React.PointerEvent) => {
+    positionRef.current = { x: event.clientX, y: event.clientY };
+
+    if (popperRef.current != undefined) {
+      void popperRef.current.update();
+    }
```
jtbandes, line 134: Why do we change the positionRef and manually call update() vs just passing in a different `anchorEl`? Does that not work?

foxymiles: Not sure. This is the method they recommend in the documentation.

foxymiles:
> edit: actually I think what I'm seeing is also a gap between the arrow and main tooltip body? Maybe we need to round() some pixel coordinates?

I'll check.

> The tooltip also runs into the edge of the window. Would be nice to enforce some padding: image

👍 

> I'd also advocate for removing the appear/disappear animations, it makes the app feel a bit more sluggish.

Yeah I agree for this use case it's an annoyance.

jtbandes:
It looks like the playhead moves in coarser increments than the tooltip, so they're not always perfectly aligned. It would be nice if they stayed perfectly aligned:

https://user-images.githubusercontent.com/14237/189421158-7eb65bd7-3650-42a1-b306-7f9dc6171c7e.mov

packages/studio-base/src/components/PlaybackControls/PlaybackControlsTooltipContent.tsx
snosenzo: yeah that's fair. I usually  lean toward more verbose namings, but I guess the separate type helps

foxymiles:
> It looks like the playhead moves in coarser increments than the tooltip, so they're not always perfectly aligned. It would be nice if they stayed perfectly aligned:

This seems to be an issue on main too and is a bit tricky to solve since they rely on completely different formulas for calculating placement so I'd be inclined to defer that to a followup PR.

jtbandes:
Also noticing that the fact that I can hover on the tooltip leads to this weird-feeling behavior:

https://user-images.githubusercontent.com/14237/189780863-915e7e56-2dd9-4458-a4ac-7461f78f31ed.mov

foxymiles:
> Also noticing that the fact that I can hover on the tooltip leads to this weird-feeling behavior:

It looks like we need to set the `disableInteractive` property to prevent this. Thanks @2metres for the pointer.

jtbandes:
> It looks like we need to set the `disableInteractive` property to prevent this.

On the other hand I could see it being useful to be able to copy/paste things out of the event metadata. Though maybe if you can easily click to see the event in the sidebar, it's not necessary for the tooltip to be interactive.

packages/studio-base/src/components/PlaybackControls/Scrubber.tsx
```diff
@@ -208,54 +115,90 @@ export default function Scrubber(props: Props): JSX.Element {
...
+      disableInteractive
+      TransitionComponent={Fade}
+      TransitionProps={{ timeout: 0 }}
+      PopperProps={{
+        popperRef,
+        modifiers: [
+          {
+            name: "computeStyles",
+            options: {
+              gpuAcceleration: false, // Fixes hairline seam on arrow in chrome.
```
jtbandes, line 150: First time I've seen a gpu acceleration setting in a CSS library 🤣

foxymiles:
> On the other hand I could see it being useful to be able to copy/paste things out of the event metadata. Though maybe if you can easily click to see the event in the sidebar, it's not necessary for the tooltip to be interactive.

As you pointed out it's annoying when you want to hover the pointer over something above the timeline and the tooltip interferes. @2metres and I discussed having some kind of modifier key interaction that let's you interact with the content of the tooltip and maybe add a "create event" button inside the tooltip when doing this.

packages/studio-base/src/components/PlaybackControls/Scrubber.tsx
foxymiles: I think all it does is use `left` & `top` positioning instead of a `transform`. Unfortunately I couldn't find another fix for this but performance still seems ok.

Issue #4395, opened 2022-09-10
Switching tabs on the sample data does not work


defunctzombie:
**Description**

When switching tabs (or really any new change in topics), the player will fail to start loading the new topics because it will try to "seek" to the same time it is already at and this is a no-op as of https://github.com/foxglove/studio/pull/4379.

Regressed in https://github.com/foxglove/studio/pull/4379 because the new logic avoid seeking when the currentTime is the same as seekTime. This breaks other logic which uses "seeking" to reset iterators and load backfill messages.

PR #4397, opened 2022-09-10
Fix topic subscriptions for iterable player

defunctzombie:
**User-Facing Changes**
Subscribing to new topics while paused is fixed.

The regression did not appear in a desktop release but did appear in the web release.

**Description**
PR https://github.com/foxglove/studio/pull/4379 broke subscribing to topics when paused. This behavior broke because 4379 added a check to seekPlayback that would avoid seeking if the requested seek time was already the current time. Unfortunately, the IterablePlayer relied on this behavior when subscriptions changed to trigger a backfill and reset the message iterators by calling seekPlayback with the currentTime.

This change fixes the setSubscriptions behavior to instead directly request the seek-backfill state rather than routing through seekPlayback which is a public-facing api.

Fixes: #4395

### DIFF
--- a/packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
+++ b/packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
@@ -293,11 +293,16 @@ export class IterablePlayer implements Player {
     this._partialTopics = partialTopics;
     this._blockLoader?.setTopics(this._partialTopics);
 
-    // We only seek playback if the player is not playing. If the player is playing, the
-    // playing state will detect any subscription changes and emit new messages.
+    // If the player is playing, the playing state will detect any subscription changes and adjust
+    // iterators accordignly. However if we are idle or already seeking then we need to manually
+    // trigger the backfill.
     if (this._state === "idle" || this._state === "seek-backfill" || this._state === "play") {
       if (!this._isPlaying && this._currentTime) {
-        this.seekPlayback(this._currentTime);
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
```diff
@@ -293,11 +293,16 @@ export class IterablePlayer implements Player {
...
+    // iterators accordignly. However if we are idle or already seeking then we need to manually
+    // trigger the backfill.
     if (this._state === "idle" || this._state === "seek-backfill" || this._state === "play") {
       if (!this._isPlaying && this._currentTime) {
-        this.seekPlayback(this._currentTime);
+        this._seekTarget = this._currentTime;
+        this._untilTime = undefined;
+
+        // Trigger a seek backfill to load any missing messages and reset the forward iterator
+        this._setState("seek-backfill");
```
jtbandes, line 305: Does this still also happen in the seekPlayback function? It’s just that we can’t use that function here because of the check for currentTime that it does?

jtbandes (approved):

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
defunctzombie: Yep. The seekPlayback function also queues the `seek-backfill` state. The `seek-backfill` state is what actually performs all the logic around resetting, loading, etc.

PR #4402, opened 2022-09-12
Improve horizontal scaling of settings editor fields

foxymiles:
**User-Facing Changes**
Improves horizontal scaling of settings editor fields.

**Description**
Improves horizontal scaling of settings editor fields and replaces muiStyled with makeStyles for the settings editor fields, improving UX for interactions with wide fields like message paths. Right align field labels for better layout when the sidebar is wide.

Screen Shot 2022-09-12 at 8 31 07 AM
Screen Shot 2022-09-12 at 8 31 28 AM
Screen Shot 2022-09-12 at 8 31 38 AM

### DIFF
--- a/packages/studio-base/src/components/SettingsTreeEditor/FieldEditor.tsx
+++ b/packages/studio-base/src/components/SettingsTreeEditor/FieldEditor.tsx
@@ -9,7 +9,6 @@ import {
   ToggleButton,
   ToggleButtonGroup,
   Typography,
-  styled as muiStyled,
   List,
   MenuItem,
   Select,
@@ -18,6 +17,7 @@ import {
   ListProps,
 } from "@mui/material";
 import { DeepReadonly } from "ts-essentials";
+import { makeStyles } from "tss-react/mui";
 import { v4 as uuid } from "uuid";
 
 import { SettingsTreeAction, SettingsTreeField } from "@foxglove/studio";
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/components/SettingsTreeEditor/FieldEditor.tsx
```diff
@@ -29,83 +29,91 @@ import { ColorPickerInput, ColorGradientInput, NumberInput, Vec3Input, Vec2Input
...
+      fontSize: "0.75em",
+      backgroundColor: inputBackgroundColor,
+
+      input: {
+        height: "1.4375em",
+      },
+      "&:hover": {
+        backgroundColor: prefersDarkMode ? "rgba(255, 255, 255, 0.13)" : "rgba(0, 0, 0, 0.09)",
+        // Reset on touch devices, it doesn't add specificity
+        "@media (hover: none)": {
```
jtbandes, line 78: TIL about `hover` media. Any reason to use `:hover ...` and `(hover: none)` rather than just `@media (hover: hover) { :hover ... }` ?

jtbandes (approved):
How about we discuss the right-aligning in product review (or slack) before merging?

packages/studio-base/src/components/SettingsTreeEditor/FieldEditor.tsx
foxymiles: Not sure.  That's @2metres' CSS wizardry.

foxymiles:
> How about we discuss the right-aligning in product review (or slack) before merging?

Yeah this is the kind of change that should have broad consensus.

packages/studio-base/src/components/SettingsTreeEditor/FieldEditor.tsx
2metres: What are you trying to do here? Not show hover on touch devices?

2metres: I think ths would do it?

```js
    "@media (hover: hover)": {
      "&:hover": {
        backgroundColor: prefersDarkMode ? "rgba(255, 255, 255, 0.13)" : "rgba(0, 0, 0, 0.09)",
      },
    },
```

packages/studio-base/src/components/SettingsTreeEditor/index.tsx
```diff
@@ -4,28 +4,30 @@
...
+const useStyles = makeStyles()((theme) => ({
+  appBar: {
+    top: -1,
+    zIndex: theme.zIndex.appBar - 1,
+    borderBottom: `1px solid ${theme.palette.divider}`,
+    padding: theme.spacing(1),
+  },
+  fieldGrid: {
+    display: "grid",
+    gridTemplateColumns: "minmax(0, 1fr) minmax(0, 2fr)",
```
2metres, line 28: Is there any reason why we wouldn't want to do this?

```suggestion
    gridTemplateColumns: "minmax(0, max-content) minmax(0, 2fr)",
```

packages/studio-base/src/components/SettingsTreeEditor/FieldEditor.tsx
```diff
@@ -29,83 +29,91 @@ import { ColorPickerInput, ColorGradientInput, NumberInput, Vec3Input, Vec2Input
...
+          backgroundColor: inputBackgroundColor,
+        },
+      },
+      "&:focus-within": {
+        backgroundColor: inputBackgroundColor,
+      },
     },
-  },
-}));
+    styledToggleButtonGroup: {
```
2metres, line 86:
```suggestion
    toggleButtonGroup: {
```

packages/studio-base/src/components/SettingsTreeEditor/FieldEditor.tsx
```diff
@@ -29,83 +29,91 @@ import { ColorPickerInput, ColorGradientInput, NumberInput, Vec3Input, Vec2Input
...
+      backgroundColor: theme.palette.action.hover,
+      gap: theme.spacing(0.25),
```
2metres, line 88:
```suggestion
      gap: theme.spacing(0.25),
      maxWidth: 200,
```

amacneil:
I like the right aligned labels. 90% of them work great but we might need to rethink the copy for a small number of them, e.g.

image

The dragging is definitely better than what we had, but feels like we could put a max-width on the labels. If I keep dragging the sidebar super wide there is no reason the labels should keep expanding indefinitely.

packages/studio-base/src/components/SettingsTreeEditor/index.tsx
foxymiles: Yeah this means long labels will be shown in full width without any truncation. Maybe that's ok and we just ask people to be reasonable with the length of labels?

foxymiles:
@amacneil @2metres I tweaked the grid rules to give what feels like a better compromise with scaling and label size. Let me know what you think.

amacneil:
Seems to work decently well, I'm fine merging as is.

Perhaps this is a separate ticket/discussion, but IMO we need to make the default sidebar width fixed in pixels (probably with two options, slightly larger default if you have a wide screen). Having the sidebar be a % of screen space makes no sense. I believe the only reason we did that originally was to save time on the implementation. @2metres for thoughts.

defunctzombie:
Small nit:

The text and error dot could use some spacing tweaks to appear less jammed together. Might be worth considering putting the error dot in front of the label to avoid UI "jumpiness" with errors appearing/disappearing.

image

defunctzombie:
Is the loss of the red outline for the input intentional?

image

foxymiles:
> Is the loss of the red outline for the input intentional?

No. I'll investigate.

amacneil:
product review feedback:
- color preview resizes while its outer box does not
- move error icons to the left of text

Issue #4368, opened 2022-09-07
Plot panel settings series sidebar should save input when I click away


amacneil:
Believe it or not, this is in edit mode:

<img width="383" alt="image" src="https://user-images.githubusercontent.com/637671/188990011-b2170ee4-105f-4fce-8b99-45ea5258f4e3.png">

When I click away, it should save the value. Ideally it would look and work consistently with renaming layouts.

<img width="386" alt="image" src="https://user-images.githubusercontent.com/637671/188990164-7560a2e4-8731-452f-b7c7-0cf90787399c.png">

foxymiles:
A clarification: the value is being saved on every keystroke. The issue is that it's not leaving the editing state on blur.

defunctzombie:
> A clarification: the value is being saved on every keystroke. The issue is that it's not leaving the editing state on blur.

Should we save on every keystroke or go into an edit mode that you can "esc" out of to undo?

foxymiles:
> Should we save on every keystroke or go into an edit mode that you can "esc" out of to undo?

Not sure which is better from UX point of view but in general all settings updates are immediate currently. We can certainly handle this differently of course.

amacneil:
If we save on every keystroke then why do we need the checkbox thing at all?

I don't particularly care whether we save on every keystroke or not (seems like an implementation detail), but I agree with Roman that it would be nice to hit Esc to cancel/revert the changes.

foxymiles:
Currently the purpose of the checkbox is to exit edit mode. I agree it's probably a bit confusing and suggests that your changes aren't saved until you click it.

defunctzombie:
A) If we go the immediate save route then lets remove the checkbox since that to me is something I see when I need to confirm my changes not when they've already been saved.

B) If we go the route of matching layout rename workflow then we should avoid saving until the change is accepted so the user can 'esc' to cancel the change.

I am 👍 on us going with A for now since it is visually the least disruptive to the existing workflow and we can explore B as an enhancement pending more feedback.

foxymiles:
If we go with A then how should the user exit edit mode? Only on enter?

defunctzombie:
Played with this feature some more and I think my only real complaint is that it should exit edit mode of "blur" or loss of focus on the input. If it did that I think it would behave mostly as expected.

foxymiles:
@defunctzombie It should be exiting edit mode on blur in that PR. Was that not working for you?

defunctzombie:
[...]

PR #4405, opened 2022-09-12
Improve settings node rename interactions

foxymiles:
**User-Facing Changes**
Minor visual changes.

**Description**
Improve settings node rename interactions. Make the editing state more visually distinct and exit editing state on blur. 

Also convert `NodeEditor` styling to `makeStyles`.

https://user-images.githubusercontent.com/93935560/189973291-415f2cff-d6f3-41d6-a70b-919727dc108d.mov

Fixes https://github.com/foxglove/studio/issues/4368

### DIFF
--- a/packages/studio-base/src/components/SettingsTreeEditor/NodeEditor.tsx
+++ b/packages/studio-base/src/components/SettingsTreeEditor/NodeEditor.tsx
@@ -10,8 +10,8 @@ import ErrorIcon from "@mui/icons-material/Error";
 import {
   Divider,
   IconButton,
-  InputBase,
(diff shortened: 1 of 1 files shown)


2metres:
Can we make it so that the check looks like it's inside the editing input?

snosenzo:
do we want people to be able to enter an empty string? looks like that's possible right now

foxymiles:
> do we want people to be able to enter an empty string? looks like that's possible right now

Yeah that's possible on main and this PR does nothing to change it. It would be up to the panel author to reject empty strings.

PR #4407, opened 2022-09-12
MessagePipeline: replace createSelectableContext with zustand

jtbandes:
**User-Facing Changes**
None

**Description**
- Removes createSelectableContext/useContextSelector and replaces it with a zustand store.
- Re-combines the pieces of state that had been factored out as `usePlayerState` back into the top-level state, because with zustand all our state needs to be in one store.
- The parts of state which can be consumed using useMessagePipeline selectors have been moved under a `public` sub-object.


packages/studio-base/src/components/MessagePipeline/index.tsx
```diff
@@ -169,119 +132,25 @@ export function MessagePipelineProvider({
...
+      dispatch({
         type: "update-player-state",
         playerState: defaultPlayerState(),
         renderDone: undefined,
       });
     };
-  }, [player, updateState]);
-
-  const topics: Topic[] | undefined = useShallowMemo(state.playerState.activeData?.topics);
-  const sortedTopics = useMemo(() => (topics ?? []).sort(), [topics]);
```
jtbandes, line 187: significant change: this used to mutate the `topics` array

packages/studio-base/src/components/MessagePipeline/store.ts
```diff
@@ -0,0 +1,359 @@
...
+      const msgEvent = lastMessageEventByTopic.get(topic);
+      if (msgEvent) {
+        const subscriberMessageEvents = messagesBySubscriberId.get(id) ?? [];
+        // the injected message is older than any new messages
+        subscriberMessageEvents.unshift(msgEvent);
+        messagesBySubscriberId.set(id, subscriberMessageEvents);
+      }
+    }
+    // We've processed all new subscriber topics into message queues
+    newTopics.clear();
```
jtbandes, line 261: significant change: this used to be `newTopicsBySubscriberId.delete(id)`, which would forget that the topic is no longer 'new'...how did this ever work?

packages/studio-base/src/components/MessagePipeline/store.ts
```diff
@@ -0,0 +1,359 @@
...
+  const newTopicsBySubscriberId = new Map(prevState.newTopicsBySubscriberId);
+
+  const previousSubscription = previousSubscriptionsById.get(action.id);
+
+  // Record any _new_ topics for this subscriber into newTopicsBySubscriberId
+  const newTopics = newTopicsBySubscriberId.get(action.id);
+  if (!newTopics) {
+    const actionTopics = action.payloads.map((sub) => sub.topic);
+    newTopicsBySubscriberId.set(action.id, new Set(actionTopics));
+  } else if (previousSubscription) {
```
jtbandes, line 150: need to port the fix from #4410 after that merges

packages/studio-base/src/components/MessagePipeline/store.ts
```diff
@@ -0,0 +1,359 @@
...
+    newTopicsBySubscriberId: new Map(),
+    lastMessageEventByTopic: new Map(),
+    lastCapabilities: [],
+
+    public: {
+      playerState: defaultPlayerState(),
+      messageEventsBySubscriberId: new Map(),
+      subscriptions: [],
+      publishers: [],
+      sortedTopics: [],
```
defunctzombie, line 97: Why do we hoist this out of active data but keep other things in active data?

packages/studio-base/src/components/MessagePipeline/MockMessagePipelineProvider.tsx
```diff
@@ -11,18 +11,17 @@
 //   found at http://www.apache.org/licenses/LICENSE-2.0
```
defunctzombie, line 11: I think its worth moving the Mock into the providers folder while you are in here making these changes? Or maybe it is useful to keep it in the MessagePipeline folder but thats not what we do for other providers.

foxymiles (approved):
Looks good. I think we can build on this taking advantage of Zustand's transient update feature to get good performance out of things updating in realtime without having to resort to the ref hacks I used for the data source info performance fix.

defunctzombie (approved):
Re-posting my comments from slack:

- I have a preference for avoiding two "topics" fields in the context
- Should the sorted topics be in active data?
- Should we drop active data from the context altogether and keep a flatter structure?

jtbandes:
> * I have a preference for avoiding two "topics" fields in the context
> * Should the sorted topics be in active data?
> * Should we drop active data from the context altogether and keep a flatter structure?

These are good ideas, but IMO out of scope for this PR since the design of activeData and sortedTopics have been around for a long time.

PR #4410, opened 2022-09-13
Fix 3D panel clearing on split

### LINKED ISSUE #4373: 3d (beta) panel goes blank when split
**Description**

Splitting the 3d (beta) panel causes the panel to go blank.

https://user-images.githubusercontent.com/84792/189009801-396a0ea6-3e0a-4b6d-9e77-e64bc61dd247.mov

jtbandes:
**User-Facing Changes**

**Description**
Fixes #4373
- Reduce unnecessary re-renders when splitting or moving panels (see comment in https://github.com/foxglove/studio/pull/3636#pullrequestreview-1012677374)
- Fix logic in MessagePipeline so latest messages are delivered to new subscribers

### DIFF
--- a/packages/studio-base/src/components/MessagePipeline/usePlayerState.ts
+++ b/packages/studio-base/src/components/MessagePipeline/usePlayerState.ts
@@ -41,15 +41,14 @@ function updateSubscriberAction(
   const previousSubscriptionsById = prevSubscriberState.subscriptionsById;
   const newTopicsBySubscriberId = new Map(prevSubscriberState.newTopicsBySubscriberId);
 
-  const previousSubscription = previousSubscriptionsById.get(action.id);
-
   // Record any _new_ topics for this subscriber into newTopicsBySubscriberId
   const newTopics = newTopicsBySubscriberId.get(action.id);
   if (!newTopics) {
     const actionTopics = action.payloads.map((sub) => sub.topic);
     newTopicsBySubscriberId.set(action.id, new Set(actionTopics));
-  } else if (previousSubscription) {
-    const prevTopics = new Set(previousSubscription.map((sub) => sub.topic));
+  } else {
+    const previousSubscription = previousSubscriptionsById.get(action.id);
+    const prevTopics = new Set(previousSubscription?.map((sub) => sub.topic) ?? []);
     for (const { topic: newTopic } of action.payloads) {
       if (!prevTopics.has(newTopic)) {
         newTopics.add(newTopic);
--- a/packages/studio-base/src/components/MosaicPathContext.ts
+++ b/packages/studio-base/src/components/MosaicPathContext.ts
@@ -5,4 +5,9 @@
 import { createContext } from "react";
 import { MosaicPath } from "react-mosaic-component";
 
+/**
+ * Exposes the mosaic path at which a panel is located. Unlike calling
+ * `mosaicWindowActions.getPath()` during render, subscribing to this context will trigger a
+ * re-render when the path changes.
+ */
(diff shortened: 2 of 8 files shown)


packages/studio-base/src/components/MosaicPathContext.ts
```diff
@@ -0,0 +1,8 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { createContext } from "react";
+import { MosaicPath } from "react-mosaic-component";
+
+export const MosaicPathContext = createContext<MosaicPath | undefined>(undefined);
```
defunctzombie, line 8: This needs a comment helping us understand its purpose and what gotcha it helps us avoid.

packages/studio-base/src/components/MessagePipeline/usePlayerState.ts
```diff
@@ -41,15 +41,14 @@ function updateSubscriberAction(
...
-  const previousSubscription = previousSubscriptionsById.get(action.id);
-
   // Record any _new_ topics for this subscriber into newTopicsBySubscriberId
   const newTopics = newTopicsBySubscriberId.get(action.id);
   if (!newTopics) {
     const actionTopics = action.payloads.map((sub) => sub.topic);
     newTopicsBySubscriberId.set(action.id, new Set(actionTopics));
-  } else if (previousSubscription) {
-    const prevTopics = new Set(previousSubscription.map((sub) => sub.topic));
+  } else {
```
defunctzombie, line 49: Is there a test to write here for asserting that the new subscriber receives the messages?

packages/studio-base/src/providers/CurrentLayoutProvider/index.tsx
```diff
@@ -97,10 +98,12 @@ export default function CurrentLayoutProvider({
   const getSelectedPanelIds = useCallback(() => selectedPanelIds.current, []);
   const setSelectedPanelIds = useCallback(
     (value: readonly string[] | ((prevState: readonly string[]) => readonly string[])): void => {
-      selectedPanelIds.current =
-        typeof value === "function" ? value(selectedPanelIds.current) : value;
-      for (const listener of [...selectedPanelIdsListeners.current]) {
-        listener(selectedPanelIds.current);
+      const newValue = typeof value === "function" ? value(selectedPanelIds.current) : value;
```
defunctzombie, line 101: Is this related to the changes? What's this guarding against with the shallowequal?

defunctzombie (approved):

packages/studio-base/src/providers/CurrentLayoutProvider/index.tsx
jtbandes: After fixing the panel reset I found that there was an extra render happening that I didn't expect. I tracked it down to a no-op change in the selectedPanelIds, so I fixed it by adding a shallowequal comparison before calling the listeners.

packages/studio-base/src/components/MessagePipeline/usePlayerState.ts
jtbandes: I think this is tricky because we do have tests for new subscribers receiving existing messages, but I think this has to do with the exact order of player state and subscribe events, like if the subscribers are happening in an async fashion. I'll make a ticket to add a test because I'd like to get this fix in.

jtbandes: https://github.com/foxglove/studio/issues/4415

PR #4411, opened 2022-09-13
Replace URDF Message Bar with MUI version

2metres:
**User-Facing Changes**
None

**Description**
Replaces URDF warnings with MUI version


2metres:
Ideally we wouldn't export this as a node but instead call `enqueueSnackbar()` so this renders inside the ToastProvider

jtbandes (approved):

PR #4420, opened 2022-09-13
Fix shallowequal in MessagePipeline topics/datatypes comparison

jtbandes:
**User-Facing Changes**
None

**Description**
Fixes an issue where datatypes from a bag file would not be vended out to panels. (regression from #4407)
topics is `Topic[]` and datatypes is a `Map`, neither of which are really appropriate for `shallowequal`. Just use object identity.


defunctzombie (approved):
Seems like we should have a test that caught this so lets add one since this behavior has wide impact when it breaks.

jtbandes:
Will add test in a follow up PR

PR #4425, opened 2022-09-14
3D: support foxglove Image types

### LINKED ISSUE #4323: 3D Beta support for Foxglove image types
Add support for RawImage, CompressedImage, and CameraCalibration.
~Blocked on https://github.com/foxglove/schemas/issues/59~

jtbandes:
**User-Facing Changes**
Added support for Foxglove RawImage, CompressedImage, and CameraCalibration schemas to the new 3D panel.

**Description**
Closes #4323

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/Cameras.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/Cameras.ts
@@ -5,6 +5,7 @@
 import { PinholeCameraModel } from "@foxglove/den/image";
 import Logger from "@foxglove/log";
 import { toNanoSec } from "@foxglove/rostime";
+import { CameraCalibration } from "@foxglove/schemas/schemas/typescript";
 import { SettingsTreeAction, SettingsTreeFields } from "@foxglove/studio";
 import type { RosValue } from "@foxglove/studio-base/players/types";
 import { MutablePoint } from "@foxglove/studio-base/types/Messages";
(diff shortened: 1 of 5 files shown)


packages/studio-base/src/panels/ThreeDeeRender/renderables/Cameras.ts
```diff
@@ -125,17 +132,25 @@ export class Cameras extends SceneExtension<CameraInfoRenderable> {
...
+      const { cameraInfo, receiveTime, originalMessage } = renderable.userData;
       if (cameraInfo) {
         const settings = this.renderer.config.topics[topicName] as
           | Partial<LayerSettingsCameraInfo>
           | undefined;
-        this._updateCameraInfoRenderable(renderable, cameraInfo, receiveTime, settings);
+        this._updateCameraInfoRenderable(
+          renderable,
+          cameraInfo,
+          originalMessage,
```
snosenzo, line 143: just curious, I know this doesn't seem to have to do with the foxglove schema types. why is the originalMessage now being passed down? Is it just so it can be shown in the picker?

snosenzo (commented):
So this doesn't add support for the image scene entity right? maybe I'm mistaken but I thought the foxglove types were also the scene entities?

packages/studio-base/src/panels/ThreeDeeRender/renderables/Cameras.ts
jtbandes: Yep, wanted to make sure the picker shows the actual message that the user provided, otherwise it might be confusing

jtbandes:
SceneEntities doesn't include an image type at this point. Meaning images don't support deletion, lifetime, frame_locked etc. We adapted the set of entities from the ROS Marker types and there is no image Marker (well there's an ImageMarker, but that's the equivalent of our ImageAnnotations)

snosenzo (approved):
gotchya, thanks for the clarification!

Issue #2624, opened 2022-01-08
State Transitions should have a setting to disable x axis syncing


jtbandes:
The State Transitions panel syncs its x axis like Plot panels, but unlike Plot panels, it provides no setting to disable syncing.

https://user-images.githubusercontent.com/14237/148629377-357c90cd-6052-4d63-b617-7ff51de16fcc.mov

PR #4428, opened 2022-09-14
Add isSynced toggle to State Transitions' Settings panel

aschimmichanga:
**User-Facing Changes**
This adds an isSynced toggle to State Transitions' Settings panel that lets users turn syncing with other panels on or off.
image

**Description**
In order to sync with other panels, the isSynced toggle must be turned on for all panels involved (for example, the State Transitions and Plot panels can be synced). Syncing is on by default.

[Relevant issue](https://github.com/foxglove/studio/issues/2624)


defunctzombie:
There's something weird with this changeset. It seems to be including changes from other PRs. Maybe some bad merge into your branch?

aschimmichanga:
it should be fixed now, was an error with main earlier

jtbandes (approved):
This looks great! I tried out the feature and it seems to work as intended. The only thing I would recommend is to make the setting on by default (to minimize user disruption, since today it's always on, and to match the Plot panel's default behavior). There is a `defaultConfig` object in StateTransitions/index.tsx where you can set isSynced to true (and it looks like the type of that variable should be changed to `StateTransitionConfig`).

PR #4430, opened 2022-09-15
Stream by importId

wimagguc:
**User-Facing Changes**

Deep links to Studio can now look like `?&ds.importId=A8cx32CJKVCt383S` rather than `?&ds.deviceId=dev_zWbczQJndSAHOZOh&ds.start=2022-08-28T17%3A48%3A16.047000000Z&ds.end=2022-08-28T17%3A48%3A35.948000000Z`

**Description**

Introduces the option to stream a file from data platform by its `importId` directly (rather all of `deviceId, start, end`).

This will work with the latest version of data platform, where the `/coverage`, `/topics` and other endpoints all support streaming by importId. These APIs will still take start/end as an optional parameter with imports, in case that's required by buffering or any other use case.


packages/studio-base/src/dataSources/FoxgloveDataPlatformDataSourceFactory.ts
```diff
@@ -28,31 +28,36 @@ class FoxgloveDataPlatformDataSourceFactory implements IDataSourceFactory {
...
+    const endTime = end ? fromRFC3339String(end) : undefined;
+
     const source = new DataPlatformIterableSource({
       api: args.consoleApi,
-      start: startTime,
-      end: endTime,
       deviceId,
+      importId,
+      ...(startTime && { start: startTime }),
+      ...(endTime && { end: endTime }),
```
jtbandes, line 45: why not just `deviceId, importId, start: startTime, end: endTime`?

packages/studio-base/src/players/IterablePlayer/foxglove-data-platform/DataPlatformIterableSource.ts
```diff
@@ -280,8 +296,11 @@ export class DataPlatformIterableSource implements IIterableSource {
         }
       }
 
-      localStart = clampTime(localStart, streamStart, streamEnd);
-      localEnd = clampTime(addTime(localStart, { sec: 5, nsec: 0 }), streamStart, streamEnd);
+      localStart = streamStart ? clampTime(localStart, streamStart, streamEnd) : undefined;
+      localEnd =
+        localStart && streamStart
+          ? clampTime(addTime(localStart, { sec: 5, nsec: 0 }), streamStart, streamEnd)
+          : undefined;
```
jtbandes, line 303: It seems weird that these could ever become undefined after we start streaming messages. For importId streaming, wouldn't we have determined the start/end time based on the coverage range above? Maybe I haven't read carefully enough but it feels like it should be impossible to get here with localStart/streamStart being undefined

packages/studio-base/src/screens/LaunchingInDesktopScreen.tsx
```diff
@@ -36,6 +36,9 @@ export function LaunchingInDesktopScreen(): ReactElement {
           case "ds.deviceId":
             desktopURL.searchParams.set("deviceId", v);
             break;
+          case "ds.importId":
+            desktopURL.searchParams.set("importId", v);
```
jtbandes, line 40: did you test that this "launching in desktop app" flow transfers the url correctly to the desktop app and actually works? I don't know how all this logic works for desktop app urls so just want to make sure we test it...

wimagguc: Yess! Launching a desktop app with `importId` (and also with `importId` plus any combination of `{deviceId, start, end}` worked well locally).

packages/studio-base/src/players/IterablePlayer/foxglove-data-platform/DataPlatformIterableSource.ts
wimagguc: So technically, if you log the `localStart` and `streamStart` parameters, they will never be null in practice. This is because none of those parameters will be null after the source has been initialized — but nothing in `messageIterator` checks whether the source is ready, and so in theory `streamStart = args.start ?? this._start` _could_ be `undefined`.

Two things we can do to suppress the linter's warning is to either add `!` to `streamStart = args.start ?? this._start!`, or, be more defensive and check whether both localStart and streamStart has a defined value. I generally like the second option more but happy to be convinced otherwise?

packages/studio-base/src/players/IterablePlayer/foxglove-data-platform/DataPlatformIterableSource.ts
```diff
@@ -51,17 +51,19 @@ export type DataPlatformInterableSourceConsoleApi = Pick<
...
 export class DataPlatformIterableSource implements IIterableSource {
   private readonly _consoleApi: DataPlatformInterableSourceConsoleApi;
 
-  private _start: Time;
-  private _end: Time;
-  private _deviceId: string;
+  private _start: Time | undefined;
+  private _end: Time | undefined;
+  private _deviceId: string | undefined;
+  private _importId: string | undefined;
```
foxymiles, line 66: I think we can mark these as `readonly`.

packages/studio-base/src/players/IterablePlayer/foxglove-data-platform/streamMessages.ts
```diff
@@ -50,9 +50,10 @@ export async function* streamMessages({
 
   /** Parameters indicating the time range to stream. */
   params: {
-    deviceId: string;
-    start: Time;
-    end: Time;
+    importId?: string;
+    deviceId?: string;
+    start?: Time;
+    end?: Time;
```
foxymiles, line 56: Could we have some stronger typing around this and maybe some other places instead of a bunch of optional params and then some logic to check them? Maybe we can have a union type like:

```typescript
type Params = { type: 'by-device', device, start, end } | { type: 'by-import', id }
```

wimagguc: Data platform will currently pass on all these values (deviceId, importId, start and end) for backwards compatibility, and from what I've seen in Studio, we'd need to allow the player to change start and end timestamps as needed.

So the check today would be more like "at least one of importId or deviceId is required; start/end timestamps are always allowed, but only required if importId is _not_ present". Is it worth trying to encode this? Would make quite a long union type I think.

foxymiles: Wouldn't this do it?

```typescript
type SourceSpec =
  | { importId: string; start?: Time; end?: Time }
  | { deviceId: string; start: Time; end: Time };
```

packages/studio-base/src/players/IterablePlayer/foxglove-data-platform/DataPlatformIterableSource.ts
jtbandes: I see, so maybe we should add a check at the beginning of messageIterator to make sure the source has been initialized? That might simplify the logic here (and avoid `!`) if you rule out some things being undefined.

If it gets too annoying to check both `if (!this._start || !this._end)` then you could wrap them in one object, like `private _initializationResult?: { start, end }` and then just check `if (!this._initializationResult)`

foxymiles (approved):
I think we could tighten up the types further here by defining something like a `ConsoleStreamSpec` type that limits the possible ways you can request data from DP but we can defer that to a followup PR.

wimagguc:
> I think we could tighten up the types further here by defining something like a `ConsoleStreamSpec` type that limits the possible ways you can request data from DP but we can defer that to a followup PR.

It probably should be a separate PR: Console is currently more forgiving than it could be — because it sends both deviceId and importId together for backwards compatibility, it also needs to accept both of those for now. We'll need another Console update before we can introduce more accurate types here.

PR #4432, opened 2022-09-15
Make eslint config intellij friendly

foxymiles:
**User-Facing Changes**
None

**Description**
This converts our root eslintrc from yaml to JSON so we can use the `__dirname` variable. This is necessary for the sub packages to find tsconfig.eslint.json in the correct root path.

Related issue https://github.com/typescript-eslint/typescript-eslint/issues/251


snosenzo (approved):
I'm all for json over yaml

defunctzombie (dismissed):
We have comments in our .eslintrc file and this change loses those. To me the comments outweight the benefit of json.

snosenzo:
this is actually just a js file, so we should be able to transfer those comments over

foxymiles:
I'll add the comments back.

foxymiles:
Manually setting the eslint working directory to `./;` as described in one of the comments here seems to be a workaround if we prefer to keep the eslint rules in yaml.

https://youtrack.jetbrains.com/issue/WEB-47201/WebStorm-incorrectly-builds-the-path-to-eslints-parserOptionsproject-file-since-20202

amacneil:
What's old is new again https://github.com/foxglove/template-typescript/pull/28

foxymiles:
Closing this since it seems like there is a workaround and people like the yaml syntax. We can revisit if it seems to be a problem for people.

Issue #4434, opened 2022-09-15
Topics still subscribed after removing 3D panel (all panels)


achim-k:
After removing all panels, some topics are still subscribed even though there is no panel left where they are displayed. 

- Version: `Foxglove Studio version 1.25.0-dev` (e92ba0b892dac94cb2392c14e773d65c84111219)
- OS: `Ubuntu 22.04.1 LTS`
- Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native): `rosbridge`
- ROS distro (if applicable): `melodic`

**Steps To Reproduce**

Server side e.g. with `docker run -it --rm --net=host ros:melodic`:

```
sudo apt update
sudo apt install -y wget ros-melodic-rosbridge-suite

wget https://assets.foxglove.dev/nuScenes-v1.0-mini-scene-0061.bag  # ~490 mb

roscore&
sleep 1
roslaunch rosbridge_server rosbridge_websocket.launch&
rosbag play nuScenes-v1.0-mini-scene-0061.bag --loop
```

Client side (studio): 

1. Launch studio app with `yarn desktop:start`
2. Connect to rosbridge datasource
3. Inspect the websocket traffic and confirm that the default 3D view subscribes to messages
4. Remove all panels
5. [OK] Observe that topics are unsubscribed
6. Re-add a 3D panel
7. [OK] Observe that topics are re-subscribed
8. Remove the 3D panel again
9. [KO] Observe that topics are still subscribed

[Screencast from 15.09.2022 16:50:53.webm](https://user-images.githubusercontent.com/9250155/190499195-d230bc43-f9b2-4dda-be97-cba8fe47fecb.webm)

**Expected Behavior**
Topics should be unsubscribed when there are no active panels

**Actual Behavior**
Some topics are still subscribed even after all panels and layouts have been removed

foxymiles:
Thanks for the detailed repro steps. I used this to verify https://github.com/foxglove/studio/pull/4579

I used `rosnode info` to determine what topics studio was subscribed to.

PR #4436, opened 2022-09-15
PanelExtensionAdapter: check panelId before performing state changes

jtbandes:
**User-Facing Changes**
Fixed a bug where an extension panel would be able to asynchronously subscribe to new topics or otherwise change application state even after the panel was removed from the layout.

**Description**
Fixes https://github.com/foxglove/studio/issues/4434


defunctzombie (approved):
We have an `isMounted` hook that returns a function you can call to see if the panel is still mounted. Using that hook would be clearer intent than the panel id trying to do the same thing.

jtbandes:
The panelId can change even if the PanelExtensionAdapter does not get unmounted.

Issue #4431, opened 2022-09-15
Developer console logs are very noisy by default


amacneil:
**Description**
Developer console should not show info-level logs by default, only errors + warnings.

<img width="1685" alt="image" src="https://user-images.githubusercontent.com/637671/190329557-e57f3d8a-c8b7-4e99-aa4d-65515269179d.png">

User impact is that this makes extension development difficult due to all the noise.

jtbandes:
> should not show info-level logs by default, only errors + warnings

Seems like this isn't something we have control over? Did you mean that we should downgrade some of our 'info' logs to 'debug' logs?

defunctzombie:
My plan is to default disable all logs except _errors_ and _warnings_ in production. The new `Studio Logs Settings` panel will have a toggle to turn on the other log levels in production for inspection and debugging.



> Seems like this isn't something we have control over? Did you mean that we should downgrade some of our 'info' logs to 'debug' logs?

We do have control over this. All our logging happens through @foxglove/log and I will add a way for that to filter out specific levels. It already has a way to disable specific log categories.

PR #4439, opened 2022-09-16
Set minimum log level to _warn_ for production builds

defunctzombie:
**User-Facing Changes**
None.

**Description**
Add a new dropdown to the Studio Logs Settings sidebar (sidebar is under a feature flag), which allows setting the global log level.

The level sets the minimum severity for writing logs to console.

image

This change updates the default minimum severity to "warn" for production builds.

Fixes: #4431

### DIFF
--- a/packages/studio-base/src/providers/StudioLogsSettingsProvider/store.ts
+++ b/packages/studio-base/src/providers/StudioLogsSettingsProvider/store.ts
@@ -4,7 +4,7 @@
 
 import { createStore, StoreApi } from "zustand";
 
-import Log, { Logger } from "@foxglove/log";
+import Log, { Logger, LogLevel, toLogLevel } from "@foxglove/log";
 import {
   IStudioLogsSettings,
   StudioLogConfigChannel,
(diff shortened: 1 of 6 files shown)


packages/studio-base/src/providers/StudioLogsSettingsProvider/store.ts
```diff
@@ -14,12 +14,16 @@ import { LocalStorageSaveState } from "./types";
...
+// fixme - override anything in saved settings? reason would be to prevent user from keeping debug logs on...
+const defaultGlobalLevel: LogLevel = process.env.NODE_ENV === "development" ? "debug" : "warn";
+
 function createStudioLogsSettingsStore(
   initialState?: LocalStorageSaveState,
 ): StoreApi<IStudioLogsSettings> {
-  log.debug(
-    `Initializing log Config. ${initialState?.disabledChannels.length ?? 0} disabled channels.`,
-  );
+  const globalLevel = toLogLevel(initialState?.globalLevel ?? defaultGlobalLevel);
```
defunctzombie, line 23: Not sure what strategy to use here. On one hand it might be nice to override the default and have this persist across page restarts, on the other it would be nice to reset this to the default if a user forgets to set this back.

jhurliman: Persisting the setting seems preferable to me

jhurliman (approved):

PR #4441, opened 2022-09-16
Add typography catalog story

foxymiles:
**User-Facing Changes**
None

**Description**
Add a new story demonstrating all variants of the `Typography` component. This is useful as a reference for styling work and also to catch any unintended theme changes.

Screen Shot 2022-09-19 at 6 38 41 AM


defunctzombie:
👍 I'd find it helpful to put each within a div that has a border so I can also see what the padding/margin is for each.

foxymiles:
> 👍 I'd find it helpful to put each within a div that has a border so I can also see what the padding/margin is for each.

Good idea. I've added a border.

defunctzombie (approved):

PR #4442, opened 2022-09-16
[3D] Add a setting for ignoring COLLADA <up_axis> tags

jhurliman:
**User-Facing Changes**
- Correctly handle COLLADA <up_axis> tags for loaded .dae files
- Added a setting to 3D panel for ignoring COLLADA <up_axis> tags to match rviz behavior

**Description**
This matches the behavior and configurability of the legacy 3D panel

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/ModelCache.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/ModelCache.ts
@@ -42,7 +42,7 @@ export class ModelCache {
   private _models = new Map<string, Promise<LoadedModel | undefined>>();
   private _edgeMaterial: THREE.Material;
 
-  public constructor(private options: ModelCacheOptions) {
+  public constructor(public readonly options: ModelCacheOptions) {
     this._edgeMaterial = options.edgeMaterial;
   }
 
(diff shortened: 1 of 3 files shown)


jtbandes:
What would it take to make the setting work without a reload?

jtbandes (approved):

snosenzo (approved):

jhurliman:
> What would it take to make the setting work without a reload?

Quite a lot, unfortunately. We throw away all the original XML on ingestion including the `<up_axis>` tag, so we would instead need to keep that tag and apply a transform instead that could be modified when the setting changes. But we'd have to parse `<up_axis>` ourselves and handle turning all possible values into the correct transform for our scene.

defunctzombie:
> Quite a lot, unfortunately. We throw away all the original XML on ingestion including the `<up_axis>` tag, so we would instead need to keep that tag and apply a transform instead that could be modified when the setting changes. But we'd have to parse `<up_axis>` ourselves and handle turning all possible values into the correct transform for our scene.

Would this be easier to do if the original xml was saved and you re-processed it? I could see someone being confused when they toggle this setting and nothing happens. Do we have any indicator dot to tell them to reload?

jtbandes:
https://github.com/foxglove/studio/pull/4442/files#diff-51bd08c4a7f13682c9171452f1b7f1c31be57cd961f4d94666466fc5ca80ea8eR144

PR #4445, opened 2022-09-19
[3D] Transform editor

jhurliman:
**User-Facing Changes**
- [3D] Transform poses can be manually edited

**Description**

A new setting toggle, `Transforms`->`Settings`->`Editable` has been added (defaults to false). When this toggle 
is enabled, six additional fields appear in the settings for each coordinate frame, allowing XYZRPY offsets to 
be manually applied on top of the normal data-driven TF interpolation.

https://user-images.githubusercontent.com/195374/191135453-b1c6f494-d484-4390-b4de-2efc74a8beb0.mp4

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/FrameAxes.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/FrameAxes.ts
@@ -480,15 +480,8 @@ export class FrameAxes extends SceneExtension<FrameAxisRenderable> {
     }
 
     const frameKey = `frame:${renderable.userData.frameId}`;
-    const xyzOffset = this.renderer.config.transforms[frameKey]?.xyzOffset;
-    const rpyOffset = this.renderer.config.transforms[frameKey]?.rpyOffset;
-
-    frame.offsetPosition = xyzOffset
-      ? [xyzOffset[0] ?? 0, xyzOffset[1] ?? 0, xyzOffset[2] ?? 0]
-      : undefined;
-    frame.offsetEulerDegrees = rpyOffset
-      ? [rpyOffset[0] ?? 0, rpyOffset[1] ?? 0, rpyOffset[2] ?? 0]
-      : undefined;
+    frame.offsetPosition = getOffset(this.renderer.config.transforms[frameKey]?.xyzOffset);
+    frame.offsetEulerDegrees = getOffset(this.renderer.config.transforms[frameKey]?.rpyOffset);
   }
 
   private static LineGeometry(): LineGeometry {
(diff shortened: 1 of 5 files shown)


defunctzombie:
> (defaults to true)

Proposing we default this to _false_ to make transforms non-editable by default as that is the typical use case.

packages/studio-base/src/panels/ThreeDeeRender/renderables/FrameAxes.ts
```diff
@@ -542,3 +620,22 @@ function abs(x: bigint): bigint {
...
+
+function getOffset(
+  maybeOffset: Readonly<[number | undefined, number | undefined, number | undefined]> | undefined,
+): THREE.Vector3Tuple | undefined {
+  if (!maybeOffset) {
+    return undefined;
+  }
+  const x = maybeOffset[0] ?? 0;
+  const y = maybeOffset[1] ?? 0;
+  const z = maybeOffset[2] ?? 0;
```
jtbandes, line 636: consider:
```suggestion
  const [x = 0, y = 0, z = 0] = maybeOffset;
```

packages/studio-base/src/panels/ThreeDeeRender/transforms/CoordinateFrame.ts
```diff
@@ -470,3 +493,53 @@ function copyPose(out: Pose, pose: Readonly<Pose>): void {
   out.orientation.z = o.z;
   out.orientation.w = o.w;
 }
+
+// Compute XYZ Euler angles in degrees from an unscaled rotation matrix. This
+// method is adapted from THREE.js Euler#setFromRotationMatrix()
```
jtbandes, line 498: why don't we just use `Euler#setFromRotationMatrix()`?

Is there a common place we should be putting these generic math helpers rather than inside CoordinateFrame.ts?

packages/studio-base/src/panels/ThreeDeeRender/renderables/FrameAxes.ts
```diff
@@ -440,6 +458,30 @@ export class FrameAxes extends SceneExtension<FrameAxisRenderable> {
...
+    const editable = this.renderer.config.scene.transforms?.editable ?? DEFAULT_EDITABLE;
+    if (!editable) {
+      frame.offsetPosition = undefined;
+      frame.offsetEulerDegrees = undefined;
+      return;
+    }
+
+    const frameKey = `frame:${renderable.userData.frameId}`;
+    frame.offsetPosition = getOffset(this.renderer.config.transforms[frameKey]?.xyzOffset);
+    frame.offsetEulerDegrees = getOffset(this.renderer.config.transforms[frameKey]?.rpyOffset);
```
jtbandes, line 484: Is there a way we could implement this without the scene extension reaching into the transform tree and mutating it? This kind of direct mutation could lead to displaying incorrect data (e.g. if we forget to reset a value when the user reverts after it's mutated) and makes it hard to understand the flow of data and settings through the app.

jhurliman: The user edits need to be accounted for during TF interpolation, the math doesn't work out the same to just apply another transform after interpolation has happened. That's why these are direct properties of `CoordinateFrame` and part of the transforms library rather than living in an abstraction layer somewhere else (I started with that approach until I realized it wouldn't work out).

packages/studio-base/src/panels/ThreeDeeRender/transforms/CoordinateFrame.ts
jhurliman: The `transforms` library doesn't depend on three.js. That will become more strictly enforced when this spins out into a standalone library (precursor to my upcoming tutorial blog post about transforms).

PR #4449, opened 2022-09-20
Fix DataPlatformIterableSource partial consumption

defunctzombie:
**User-Facing Changes**
Fixes streaming data from data platform when there are gaps in coverage.

**Description**
When streaming from data platform using the "partial" consumption type, the iterable source would incorrectly reset the next window. This happened when trying to use coverage information to skip gaps but failing to detect that the window was already in a valid coverage zone.

Regressed in: https://github.com/foxglove/studio/pull/4398


jtbandes (approved):

defunctzombie:
Tested manually on data set with small gaps and large gaps.

Issue #4446, opened 2022-09-20
User Scripts do not run if foxglove opened by shareable link


RileyEv:
**Description**

When attempting to open remote rosbags with a shareable link and a saved layout with user scripts, the user scripts do not run. The visualisations still work, it just affects anything that depends on user scripts. You can get the scripts to run by deleting and re-importing them. If opening foxglove without a direct link and then opening a remote rosbag from a url, there are no issues. 

- Version: 1.25.0
- OS: I've been able to reproduce this on the web version and on local versions on Linux. 
- Data source: Remote rosbag opened with shareable link.

**Steps To Reproduce**

1. Add the example layout from below and ensure that it is the layout that opens by default.
2. Open this link: `https://studio.foxglove.dev/?ds=ros1-remote-bagfile&ds.url=https%3A%2F%2Fassets.foxglove.dev%2FnuScenes-v1.0-mini-scene-0061.bag`. This will open the sample foxglove data in the browser, the layout should then open.
3. User scripts do not run. I get the error message in the user scripts window `Input "/gps" is not yet available InputTopicsChecker`

**Expected Behavior**
In the ros message inspector panel on the right, I'd expect to see `"Hello World"` produced at the same Hz as the `/gps` message.

**Actual Behavior**
The panel shows nothing as the message is not being produced by the user scripts.

**Example Layout**
(sorry it wouldn't let me upload json! :cry:)
```json
{
  "configById": {
    "NodePlayground!1nieorr": {
      "selectedNodeId": "394d2b33-eba0-4d22-baac-3615ee5f0d29",
      "autoFormatOnSave": true
    },
    "RawMessages!25axhjx": {
      "diffEnabled": false,
      "diffMethod": "custom",
      "diffTopicPath": "",
      "showFullMessageForDiff": false,
      "topicPath": "/studio_script/output_topic"
    },
    "3D!1z1k64b": {
      "cameraState": {
        "distance": 19.999999999999982,
        "perspective": true,
        "phi": 59.99999999999997,
        "target": [
          0,
          0,
          0
        ],
        "targetOffset": [
          0.027861423906154804,
          0.027861423906154807,
          0
        ],
        "targetOrientation": [
          0,
          0,
          0,
          1
        ],
        "thetaOffset": 44.99999999999999,
        "fovy": 45,
        "near": 0.01,
        "far": 5000
      },
      "followMode": "follow-pose",
      "scene": {},
      "transforms": {},
      "topics": {
[...]

PR #4455, opened 2022-09-20
Fix race conditions in UserNodePlayer

jtbandes:
**User-Facing Changes**
Fixed an issue where user scripts would sometimes fail to run after loading a data source.

**Description**
Fixes https://github.com/foxglove/studio/issues/4446
The root cause was that setUserNodes could be called while user node registrations were in progress from a previous PlayerState listener, and they would stomp on each other's state. This PR fixes the problem by moving _most_ of the state mutation under a mutex to avoid races.

### DIFF
--- a/packages/studio-base/src/players/UserNodePlayer/index.ts
+++ b/packages/studio-base/src/players/UserNodePlayer/index.ts
@@ -11,6 +11,7 @@
 //   found at http://www.apache.org/licenses/LICENSE-2.0
 //   You may not use this file except in compliance with the License.
 
+import { captureException } from "@sentry/core";
 import { isEqual, uniq } from "lodash";
 import memoizeWeak from "memoize-weak";
 import shallowequal from "shallowequal";
@@ -84,6 +85,7 @@ function maybePlainObject(rawVal: unknown) {
   return rawVal;
 }
 
+/** Mutable state protected by a mutex lock */
 type ProtectedState = {
   nodeRegistrationCache: NodeRegistrationCacheItem[];
   nodeRegistrations: readonly NodeRegistration[];
@@ -885,7 +887,10 @@ export default class UserNodePlayer implements Player {
       .runExclusive(async (state) => {
         this._setSubscriptionsUnlocked(subscriptions, state);
       })
-      .catch((err) => log.error(err));
+      .catch((err) => {
+        log.error(err);
+        captureException(err);
+      });
   }
 
   private _setSubscriptionsUnlocked(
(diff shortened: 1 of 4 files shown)


packages/studio-base/src/players/UserNodePlayer/index.ts
```diff
@@ -84,29 +84,39 @@ function maybePlainObject(rawVal: unknown) {
   return rawVal;
 }
 
+type ProtectedState = {
```
defunctzombie, line 87: Comment about the purpose of protected state and how one should protect it (maybe this comment belongs on the field in the class - that I leave to you to decide).

packages/studio-base/src/players/UserNodePlayer/index.ts
```diff
@@ -879,15 +881,25 @@ export default class UserNodePlayer implements Player {
 
   public setSubscriptions(subscriptions: SubscribePayload[]): void {
     this._subscriptions = subscriptions;
+    this._protectedState
+      .runExclusive(async (state) => {
+        this._setSubscriptionsUnlocked(subscriptions, state);
+      })
+      .catch((err) => log.error(err));
```
defunctzombie, line 888: Send to sentry?

packages/studio-base/src/players/UserNodePlayer/index.ts
```diff
@@ -913,9 +925,11 @@ export default class UserNodePlayer implements Player {
   }
 
   public close = (): void => {
-    for (const nodeRegistration of this._nodeRegistrations) {
-      nodeRegistration.terminate();
-    }
+    void this._protectedState.runExclusive(async (state) => {
```
defunctzombie, line 928: Why is the function call to runExclusive async here? There's no async action in it.

defunctzombie (approved):

packages/studio-base/src/players/UserNodePlayer/index.ts
jtbandes: What if there is already an action in progress under the mutex lock?

defunctzombie: ? That doesn't make the function given to the runExclusive call have to be async tho. It only needs to be async if there's something async in that function.

jtbandes: Oh I see, at first you said "the function call to runExclusive", but the question makes more sense for "the function given to the runExclusive call". The answer is that's just how the types are currently set up, could probably change it 🤷

PR #4456, opened 2022-09-20
[WIP] Data handler extensions

jhurliman:
**User-Facing Changes**
- TBD

**Description**

This PR adds a new `registerDataHandler()` method to ExtensionContext, allowing extensions to register data 
handlers in addition to panels. A data handler is essentially a headless panel; it has all of the non-rendering 
capabilities of an extension panel (PanelExtensionContext is an extension of DataExtensionContext), plus an 
onData callback similar in form to onRender.

This opens up some new possibilities:

1. The same workflow of message transforming that is done with user scripts today could also be done in an extension. We can provide an extension template that implements the following method: `registerMessageTransformer<T>(inputTopics: string[], outputTopic: Topic, messageHandler: (messageEvent: MessageEvent, globalVars: Record<string, VariableValue>) => Message<T> | Message<T>[]);` and have something that ergonomically operates like a user script, but as part of an installed extension instead of persisting in a layout.
2. Another extension template could be provided for transforming any topics matching a given datatype to a set of Primitive messages, making it easy to build scene extensions that add support for rendering new datatypes. Example: `registerSceneExtension<T>(datatype: string, messageHandler: (messageEvent: MessageEvent<T>, globalVars: Record<string, VariableValue>) => PrimitiveMessage<T> | PrimitiveMessage<T>[]);`
3. Customize the behavior of Studio without having to add a dummy panel to every layout. Since (almost) the full panel API is offered to these headless extension points that execute as long as the extension is installed, users don't have to add a panel to their layout to run the extension.

**TODO**

- [ ] Figure out how/where to inject `ExtensionCatalog#installedDataHandlers` into Studio's React tree
- [ ] Call onData?.(state) methods for all data handlers each frame
- [ ] Working proposal for `registerMessageHandler()` using the existing extension panel API
- [ ] Tests

### DIFF
--- a/packages/studio-base/src/components/DataExtensionAdapter/DataExtensionAdapter.tsx
+++ b/packages/studio-base/src/components/DataExtensionAdapter/DataExtensionAdapter.tsx
@@ -2,10 +2,7 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
-import { useTheme } from "@mui/material";
-import { isEqual } from "lodash";
-import { CSSProperties, useEffect, useLayoutEffect, useMemo, useRef, useState } from "react";
-import { useUpdateEffect } from "react-use";
+import { useEffect, useLayoutEffect, useMemo, useRef, useState } from "react";
 import { v4 as uuid } from "uuid";
 
 import { useValueChangedDebugLog } from "@foxglove/hooks";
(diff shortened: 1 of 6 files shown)


defunctzombie:
Closing this out in favor of https://github.com/foxglove/studio/pull/4507 which implements our proposed extension api for transforming messages. We might re-visit an extension that allows for transforming based on topics but that is not the current ask.

PR #4460, opened 2022-09-21
3D settings: filter topics by visibility

### LINKED ISSUE #4433: 3D settings: filter topics by visibility
New 3D panel is missing this feature:

snosenzo:
**User-Facing Changes**
 - add filter field to topics settings node with the ability to filter the topics list by visibility

**Description**
 - broke out some functions to enable updating of settings without adding customLayerActions
 - topics and custom layer settings nodes now synced in debounce update

Fixes #4433

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
@@ -17,6 +17,7 @@ import {
   ParameterValue,
   SettingsIcon,
   SettingsTreeAction,
+  SettingsTreeField,
   SettingsTreeNodeActionItem,
   SettingsTreeNodes,
   Topic,
@@ -43,7 +44,7 @@ import {
   normalizeTransformStamped,
 } from "./normalizeMessages";
 import { Cameras } from "./renderables/Cameras";
-import { CoreSettings } from "./renderables/CoreSettings";
+import { CoreSettings, TopicsFilterSelect } from "./renderables/CoreSettings";
 import { FrameAxes, LayerSettingsTransform } from "./renderables/FrameAxes";
 import { Grids } from "./renderables/Grids";
 import { Images } from "./renderables/Images";
@@ -99,6 +100,7 @@ export type RendererEvents = {
 };
 
 export type FollowMode = "follow-pose" | "follow-position" | "follow-none";
+export type TopicsFilter = "list-all" | "list-visible" | "list-not-visible";
 
 export type RendererConfig = {
   /** Camera settings for the currently rendering scene */
@@ -107,6 +109,8 @@ export type RendererConfig = {
   followTf: string | undefined;
   /** Camera follow mode */
   followMode: FollowMode;
+  /** Filter Mode for Showing Topics in Settings Panel */
+  topicsFilter: TopicsFilter;
   scene: {
     /** Show rendering metrics in a DOM overlay */
     enableStats?: boolean;
@@ -287,6 +291,7 @@ export class Renderer extends EventEmitter<RendererEvents> {
   private aspect: number;
   private controls: OrbitControls;
   public followMode: FollowMode;
+  private topicSettingsFilter: TopicsFilter;
   // The pose of the render frame in the fixed frame when following was disabled
   private unfollowPoseSnapshot: Pose | undefined;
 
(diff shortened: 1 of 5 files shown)


packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -99,6 +100,7 @@ export type RendererEvents = {
 };
 
 export type FollowMode = "follow-pose" | "follow-position" | "follow-none";
+export type TopicsFilter = "list-all" | "list-visible" | "list-not-visible";
```
jtbandes, line 103: remove "list-" prefix? Suppose I could've made the same comment about follow modes, oh well, too late 😅

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -291,6 +295,7 @@ export class Renderer extends EventEmitter<RendererEvents> {
   private aspect: number;
   private controls: OrbitControls;
   public followMode: FollowMode;
+  private topicSettingsFilter: TopicsFilter;
```
jtbandes, line 298: is there a reason for the naming difference of topicsFilter vs topicSettingsFilter?

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -623,8 +627,27 @@ export class Renderer extends EventEmitter<RendererEvents> {
...
+        defaultExpansionState: "expanded",
+        actions: [
+          { id: "show-all", type: "action", label: "Show All" },
+          { id: "hide-all", type: "action", label: "Hide All" },
+        ],
+        fields: {
+          topicsFilter: {
+            ...TopicsFilterSelect,
+            value: this.topicSettingsFilter,
+          } as SettingsTreeField,
```
jtbandes, line 646: Would be nicer to add a strong type to TopicsFilterSelect itself rather than using `as` here. `as` can hide type errors

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -702,10 +725,30 @@ export class Renderer extends EventEmitter<RendererEvents> {
...
-      // Rebuild the settings nodes for all scene extensions
-      for (const extension of this.sceneExtensions.values()) {
-        this.settings.setNodesForKey(extension.extensionId, extension.settingsNodes());
-      }
+      this._rebuildSceneExtensionNodes();
+    }
+  }
+
+  private _rebuildSceneExtensionNodes(): void {
+    const listAllFilter = (_: SettingsTreeEntry) => true;
```
jtbandes, line 733: nit: could save a little CPU by not running .filter() if the mode is list-all.

packages/studio-base/src/panels/ThreeDeeRender/renderables/CoreSettings.ts
```diff
@@ -30,6 +30,18 @@ export const DEFAULT_PUBLISH_SETTINGS: RendererConfig["publish"] = {
...
 };
 
+const TopicsFilterOptions = [
+  { label: "All", value: "list-all" },
+  { label: "Visible", value: "list-visible" },
+  { label: "Not Visible", value: "list-not-visible" },
+];
+export const TopicsFilterSelect = {
+  label: "Filter Topics",
+  help: "Filter Topics By Visibility",
```
jtbandes, line 40: I don't think we're consistent everywhere but I'm pretty sure we decided that we want to use sentence case
```suggestion
  label: "Filter topics",
  help: "Filter topics by visibility",
```

jtbandes (approved):
Nice!

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
snosenzo: oop no, I think I just forgot to change this in my renaming to `TopicsFilter`

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
snosenzo: ah yeah, the input type was being read as a `string` rather than `select` so it wasn't narrowing the scope how I wanted. -- should be fixed now

PR #4461, opened 2022-09-21
Rename tsx dataSources files

defunctzombie:
**User-Facing Changes**
None

**Description**
These files don't need to be tsx. This renames them to ts for consistency with other non tsx files.


jtbandes (approved):
Is there any performance improvement that we get from doing this? Should we make a linter that requires files to be .ts if they don't use any jsx syntax?

defunctzombie:
> Is there any performance improvement that we get from doing this? Should we make a linter that requires files to be .ts if they don't use any jsx syntax?

Not something I have measured. I only noticed this by chance rather than seeing something speed up by changing these files.

PR #4481, opened 2022-09-26
Clear tab index on number input step buttons

### LINKED ISSUE #4451: Cannot tab between settings sidebar fields
**Description**

- Version: main
- OS: all
- Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native): all
- ROS distro (if applicable): n/a

**Steps To Reproduce**

1. Select an editable field in the 3D panel settings sidebar that is followed by another editable field, such as "Translation Offset (X)" for a coordinate frame (with editing enabled).
2. Press tab

**Expected Behavior**

The following field in the sidebar becomes focused.

**Actual Behavior**

The following field in the sidebar does not gain focus.

foxymiles:
**User-Facing Changes**
Improve tab behavior on settings editor input fields.

**Description**
Set tab index for the stepper buttons on the number input to -1 so that hitting tab on the keyboard moves directly to the next input field.

Fixes #4451

### DIFF
--- a/packages/studio-base/src/components/SettingsTreeEditor/inputs/NumberInput.tsx
+++ b/packages/studio-base/src/components/SettingsTreeEditor/inputs/NumberInput.tsx
@@ -168,7 +168,7 @@ export function NumberInput(
             className={classes.iconButton}
             size="small"
             edge="start"
-            tabIndex={-1}
+            tabIndex={-1} // Disable tabbing to the step buttons.
             onClick={(event: React.MouseEvent) =>
               updateValue((value ?? placeHolderValue ?? 0) - (event.shiftKey ? step * 10 : step))
             }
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/components/SettingsTreeEditor/inputs/NumberInput.tsx
```diff
@@ -168,6 +168,7 @@ export function NumberInput(
             className={classes.iconButton}
             size="small"
             edge="start"
+            tabIndex={-1}
```
jtbandes, line 171: Include a comment explaining why this is here? Otherwise someone might be inclined to delete it later.

jtbandes (approved):

packages/studio-base/src/components/SettingsTreeEditor/inputs/NumberInput.tsx
foxymiles: I've added a comment.

PR #4482, opened 2022-09-26
Disable sign in sidebar tab for docker instances

foxymiles:
**User-Facing Changes**
Disable sign in sidebar tab for docker hosted instances.

**Description**
Sign in doesn't currently work for instances hosted from docker so hide the sidebar tab.

Fixes #4404

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,6 +3,7 @@ FROM node:16 as build
 WORKDIR /src
 COPY . ./
 
+ENV FOXGLOVE_DISABLE_SIGN_IN=true
 RUN yarn install --immutable
 RUN yarn run web:build:prod
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,8 +3,9 @@ FROM node:16 as build
 WORKDIR /src
 COPY . ./
 
-ENV FOXGLOVE_DISABLE_SIGN_IN=true
 RUN yarn install --immutable
+
+ENV FOXGLOVE_DISABLE_SIGN_IN=true
 RUN yarn run web:build:prod
 
 # Release stage
--- a/packages/studio-base/src/Workspace.tsx
+++ b/packages/studio-base/src/Workspace.tsx
@@ -185,7 +185,8 @@ export default function Workspace(props: WorkspaceProps): JSX.Element {
     return extensions;
   }, [availableSources]);
 
-  const supportsAccountSettings = useContext(ConsoleApiContext) != undefined;
+  const supportsAccountSettings =
+    useContext(ConsoleApiContext) != undefined && props.disableSignin !== true;
 
   // We use playerId to detect when a player changes for RemountOnValueChange below
   // see comment below above the RemountOnValueChange component
@@ -535,7 +536,7 @@ export default function Workspace(props: WorkspaceProps): JSX.Element {
       ["help", { iconName: "QuestionCircle", title: "Help", component: HelpSidebar }],
     ]);
 
-    if (supportsAccountSettings && props.disableSignin !== true) {
+    if (supportsAccountSettings) {
       bottomItems.set("account", {
         iconName: currentUser != undefined ? "BlockheadFilled" : "Blockhead",
         title: currentUser != undefined ? `Signed in as ${currentUser.email}` : "Account",
@@ -555,7 +556,6 @@ export default function Workspace(props: WorkspaceProps): JSX.Element {
     playerProblems,
     enableStudioLogsSidebar,
     supportsAccountSettings,
-    props.disableSignin,
     currentUser,
   ]);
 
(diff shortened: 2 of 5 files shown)


packages/studio-base/src/Workspace.tsx
```diff
@@ -534,7 +535,7 @@ export default function Workspace(props: WorkspaceProps): JSX.Element {
       ["help", { iconName: "QuestionCircle", title: "Help", component: HelpSidebar }],
     ]);
 
-    if (supportsAccountSettings) {
+    if (supportsAccountSettings && props.disableSignin !== true) {
```
jtbandes, line 538: We already have this `supportsAccountSettings` bool, which comes from the presence/absence of the ConsoleApi object. Should we just use that instead of adding a new flag? The ConsoleApi is created here: https://github.com/foxglove/studio/blob/877896a47651a768bceca732a0c2e6009fe258ba/web/src/Root.tsx#L56 Maybe we change this to undefined if the env var is not present?

foxymiles: It looks like we have a lot of places that are not prepared for this to be undefined. That feels like a riskier change.

jtbandes: Then let's remove the supportsAccountSettings variable. My main complaint is that we have two pieces of code here trying to do the same thing and one of them is not being used.

foxymiles: The console api stuff is used all over the codebase. I'd prefer not to touch it for what amounts to a targeted fix for a niche issue. The risk/benefit doesn't seem to be there IMO.

Dockerfile
```diff
@@ -3,6 +3,7 @@ FROM node:16 as build
 WORKDIR /src
 COPY . ./
 
+ENV FOXGLOVE_DISABLE_SIGN_IN=true
```
defunctzombie, line 6: This should be below the yarn install for better docker layer caching. When this value changes we only need to build not install.

foxymiles: I have yet to get a successful docker layer cache since the previous `COPY` breaks cache when any change anywhere in the entire codebase is made. But I can move this.

defunctzombie (commented):
I probably have a preference for flipping this around and explicitly enabling features in our production builds rather than disabling in non-prod builds. For someone going through the Dockerfile I don't think it's useful to be bogged down with noise like these env flags since these are features specific to our production builds. We will need a workflow that enables these for our local development tho.

I agree with @jtbandes that we should have 1 flag for identifying whether we have "account" functionality.

Dockerfile
defunctzombie: Ah right - indeed that COPY is a killer and we need it because of the monorepo workspace :(

foxymiles:
> I probably have a preference for flipping this around and explicitly enabling features in our production builds rather than disabling in non-prod builds. For someone going through the Dockerfile I don't think it's useful to be bogged down with noise like these env flags since these are features specific to our production builds. We will need a workflow that enables these for our local development tho.
> 
> I agree with @jtbandes that we should have 1 flag for identifying whether we have "account" functionality.

I guess approaching this ticket I was just generally reluctant to make deeper changes to something that could potentially break logins in order to fix something that seems like a minor inconvenience for a small minority of users. If you think it's worth digging deeper on this and making it semantically an enable vs a disable and to make the required production and dev environment changes I can do that.

defunctzombie (approved):
Ok lets go with this env var for this change. We might revisit extracting console or making it generally optional throughout the app.

packages/studio-base/src/Workspace.tsx
```diff
@@ -184,7 +185,8 @@ export default function Workspace(props: WorkspaceProps): JSX.Element {
     return extensions;
   }, [availableSources]);
 
-  const supportsAccountSettings = useContext(ConsoleApiContext) != undefined;
+  const supportsAccountSettings =
+    useContext(ConsoleApiContext) != undefined && props.disableSignin !== true;
```
jtbandes, line 189: Did you mean `||`?

But is there any reason to keep the `useContext(ConsoleApiContext) != undefined` part? I thought we determined it's never undefined today.

jtbandes (approved):

packages/studio-base/src/Workspace.tsx
foxymiles: It just happens to never be undefined today but there's nothing in the API that constrains this to be true. I assume this check was installed for a reason at some point?

jtbandes: This constrains it: https://github.com/foxglove/studio/blob/d1381913a21d1910a95cdc5db493f0d1893e7982/packages/studio-base/src/App.tsx#L53

foxymiles: Yes that's currently the case but it's an implicit constraint. `useContext` can return undefined.

PR #4486, opened 2022-09-26
Hides Clear button when color is cleared or when flag is set to true

### LINKED ISSUE #4122: Hide `x` buttons in settings color picker when control is already set to the default
**Description**
https://github.com/foxglove/studio/pull/4060 introduced the ability to "clear" color fields, resetting them to the default.

Unfortunately, this leads to some instances where I have an `x` button that does nothing when clicked, because the field is already at the default.

image

I think we should hide the `x` button when a field is already set to its default value.

According to https://github.com/foxglove/studio/pull/4060#issuecomment-1213346122, "default" is not actually a concept, we just send `undefined` to the panel and let it figure out how it wants to handle things.

> It's up to the panel implementor. They'll receive an incoming value of undefined but can process that however they want. In the case of that panel it's reverting it to the panel's own concept of the default. But at the level of the settings ui there's no "default" value.

Therefore, to fix this we may need to introduce a concept of "default" for controls like this. Opening up for discussion.

cc @foxymiles @2metres

aschimmichanga:
**User-Facing Changes**
The Clear button is now hidden whenever there is no color defined in the color picker or when the panel author chooses to hide it.

**Description**
This adds a new flag for color pickers, `hideClearButton`, which hides the clear button when set to true by the panel author. 

On top of covering the case where the color value is cleared, this allows flexibility for different use-cases, such as if different color values are being generated dynamically or if the author conditionally wants or doesn't want users to change the color picker (such as hiding when a user-defined default value is chosen).

Fixes #4122

### DIFF
--- a/packages/studio-base/src/components/SettingsTreeEditor/index.stories.tsx
+++ b/packages/studio-base/src/components/SettingsTreeEditor/index.stories.tsx
@@ -632,7 +632,14 @@ const ColorSettings: SettingsTreeNodes = {
     fields: {
       undefined: { label: "Undefined", input: "rgb", value: undefined, placeholder: "placeholder" },
       invalid: { label: "Invalid", input: "rgb", value: "invalid" },
+      hiddenClearButton: {
+        label: "Hidden Clear Button",
+        input: "rgb",
+        value: "#00ffbf",
+        shouldHideClearButton: true,
+      },
       hex6: { label: "Hex 6", input: "rgb", value: "#ffaa00" },
+      hex7: { label: "Hex 6", input: "rgb", value: "#ffaa00" },
       hex8: { label: "Hex 8", input: "rgb", value: "#00aaff88" },
       rgb: { label: "RGB", input: "rgb", value: "rgb(255, 128, 0)" },
       rgba: { label: "RGBA", input: "rgba", value: "rgba(255, 0, 0, 0.5)" },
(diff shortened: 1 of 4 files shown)


packages/studio-base/src/components/SettingsTreeEditor/inputs/ColorPickerInput.tsx
```diff
@@ -88,7 +90,11 @@ export function ColorPickerInput(props: ColorPickerInputProps): JSX.Element {
           readOnly: true,
           startAdornment: <ColorSwatch color={swatchColor} onClick={handleClick} />,
           endAdornment: (
-            <IconButton onClick={clearValue} size="small" disabled={disabled}>
+            <IconButton
+              onClick={clearValue}
+              size="small"
+              style={{ visibility: shouldHideClearButton ? "hidden" : "visible" }}
+            >
```
foxymiles, line 97: I think we lost the `disabled` property here.

packages/studio-base/src/components/SettingsTreeEditor/index.stories.tsx
```diff
@@ -632,7 +632,14 @@ const ColorSettings: SettingsTreeNodes = {
...
       undefined: { label: "Undefined", input: "rgb", value: undefined, placeholder: "placeholder" },
       invalid: { label: "Invalid", input: "rgb", value: "invalid" },
+      hiddenClearButton: {
+        label: "Hidden Clear Button",
+        input: "rgb",
+        value: "#00ffbf",
+        hideClearButton: true,
+      },
       hex6: { label: "Hex 6", input: "rgb", value: "#ffaa00" },
+      hex7: { label: "Hex 6", input: "rgb", value: "#ffaa00" },
```
jtbandes, line 642:
```suggestion
      hex7: { label: "Hex 7", input: "rgb", value: "#ffaa00" },
```

packages/studio-base/src/components/SettingsTreeEditor/inputs/ColorPickerInput.tsx
```diff
@@ -88,7 +90,12 @@ export function ColorPickerInput(props: ColorPickerInputProps): JSX.Element {
           readOnly: true,
           startAdornment: <ColorSwatch color={swatchColor} onClick={handleClick} />,
           endAdornment: (
-            <IconButton onClick={clearValue} size="small" disabled={disabled}>
+            <IconButton
+              onClick={clearValue}
+              size="small"
+              disabled={disabled}
+              style={{ visibility: shouldHideClearButton ? "hidden" : "visible" }}
+            >
```
jtbandes, line 98: I think it might be slightly better to completely remove the button rather than setting `visibility: hidden` when it is hidden. `visibility: hidden` hides an element visually, but the element remains in the DOM, and the layout of other elements on the page may still be affected by it (i.e., it would take up space in the text field). I would recommend doing something like: `endAdornment: !shouldHideClearButton && <IconButton>...</IconButton>` which will completely remove the button when it's hidden.

However, visibility:hidden is not bad and if you have a particular reason to prefer it which I wasn't aware of, then that's fine too!

jtbandes (approved):

packages/studio-base/src/components/SettingsTreeEditor/index.stories.tsx
jtbandes: looks like this still says 6

jtbandes (approved):

PR #4489, opened 2022-09-27
Editing a marker color hides the color input as it is still being edited

### LINKED ISSUE #4487: Editing a marker color hides the color input as it is still being edited
In the 3D panel I've expanded the settings for a marker message and started to edit the color. After a second the color picker disappears and the settings are collapsed. I am not able to continue selecting the color I want.

https://user-images.githubusercontent.com/84792/192557977-1ed5401a-b23e-4d33-8b4f-db0427bbf98e.mov

snosenzo:
Fix Color Editors closing in 3D panel

**User-Facing Changes**
- n/a

**Description**
 - was rebuilding topics nodes each time and it was causing the continous editors from staying open

Fixes #4487

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
@@ -644,6 +644,7 @@ export class Renderer extends EventEmitter<RendererEvents> {
             value: this.topicsFilter,
           },
         },
+        children: this.settings.tree()["topics"]?.children,
         handler: this.handleTopicsAction,
       },
     };


jtbandes:
Why does setting this `children` prevent the tree from closing?

snosenzo:
because the topics tree was updated with the topics children as empty, and then the scene extension nodes were added when the nodes are rebuilt with the filter. with this change the existing settings tree nodes are consistently in the settings and only removed when the visibility changes

foxymiles (approved):

jtbandes (approved):

Issue #4478, opened 2022-09-25
Unintended placement of UI elements in panel full-screen mode


esthersweon:
Likely related to @jtbandes' work in https://github.com/foxglove/studio/pull/4459.

**Description**
- Strange and likely unintentional placement of UI components in panel full-screen mode (see playback bar and panel settings below):

<img width="1512" alt="Screen Shot 2022-09-25 at 12 45 36 PM" src="https://user-images.githubusercontent.com/6993359/192162372-0732ff0e-2531-4d19-a7bf-149df1b3944d.png">

- Version: 1.25
- OS: MacOS Monterey 12.0
- Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native): Sample data
- ROS distro (if applicable): N/A

**Steps To Reproduce**

- Open Studio with sample data
- Full-screen the 3D panel

**Expected Behavior**
- Going into full-screen mode should close sidebar or at least completely cover it
- Going into full-screen mode should probably cover all or none of the sidebar menu as well (one visible cog icon is strange)
- Playback bar should take up full width of app if sidebar is closed

**Actual Behavior**
- Parts of sidebar menu and panel are visible
- Playback bar is weirdly floating on top of panel UI

defunctzombie:
Related to this - I'm tempted to say that maybe "full screen" should mean "panel area" rather than sidebar? I think a useful workflow is to "full screen" a panel to configure it. For this we need sidebar visibility.

foxymiles:
We should add a story for this too as part of the fix.

jhurliman:
> Related to this - I'm tempted to say that maybe "full screen" should mean "panel area" rather than sidebar? I think a useful workflow is to "full screen" a panel to configure it. For this we need sidebar visibility.

As long as there is a "full full screen" option available.

PR #4490, opened 2022-09-27
Fix playback bar overlapping fullscreen panel

jtbandes:
**User-Facing Changes**
Fixed a visual artifact with fullscreen panels.

**Description**
Fixes https://github.com/foxglove/studio/issues/4478
Removed the behavior where we automatically exit fullscreen when the window loses focus, because we no longer have the "lock fullscreen" feature or the ability to fullscreen from the quick actions overlay.

### DIFF
--- a/packages/studio-base/src/Workspace.stories.tsx
+++ b/packages/studio-base/src/Workspace.stories.tsx
@@ -1,0 +1,80 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { fireEvent, screen } from "@testing-library/dom";
+
+import MultiProvider from "@foxglove/studio-base/components/MultiProvider";
+import Panel from "@foxglove/studio-base/components/Panel";
+import PanelToolbar from "@foxglove/studio-base/components/PanelToolbar";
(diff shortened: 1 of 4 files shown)


defunctzombie:
When there are no playback controls there's this weird gap in fullscreen mode. This would be present for live connections where we don't have a playback bar. We can consider this a separate issue if its not related to the fix for the bug since the bug has more impact because it covers the panel and prevents viewing it.

image

How do we make this fullscreen feature more robust to changes in playback control size or presence of playback controls?

jtbandes:
That is unrelated to the bug fix.

defunctzombie (approved):
My take on the bugfix is that maybe we should address the core deficiency that requires us to tweak a magic constant in a file whenever we change the playback control height.

jtbandes:
Any suggestions for how to do that? I'm not sure how to make some element track the position of another arbitrary element on screen. I suppose we could make a react context provider for the toolbar height. Also not sure how this should interact with the sidebar.

defunctzombie:
> Any suggestions for how to do that? I'm not sure how to make some element track the position of another arbitrary element on screen. I suppose we could make a react context provider for the toolbar height. Also not sure how this should interact with the sidebar.

My take would be that our mosaic or whatever manages the tile layout needs the "fullscreen" feature as a first class concept. Then any tile becomes fullscreen in the space of the mosaic.

PR #4494, opened 2022-09-28
[3D] Fix orientation for glTF, COLLADA, STL, and OBJ meshes

jhurliman:
**User-Facing Changes**
- Fixed orientation for meshes in the `3D` panel

**Description**

This adds a storybook that renders the same source model of labeled axis arrows converted into four different formats and stored as data URIs. Before applying fixes, the glTF (.glb) model successfully appeared with the "+y" facing up, but the "+z" faced in the wrong direction. The other three formats did not show "+y" facing up. Fixes have been applied to all model loaders to ensure the orientation matches Blender and each other.

Two separate storybooks have been added specifically for COLLADA `<up_axis>` handling, to demonstrate the behavior difference between enabling or disabling the "ignore up-axis" setting.

Screen Shot 2022-09-27 at 11 03 37 PM


jtbandes (approved):

jtbandes:
Can we include a collada model with a different value of `<up_axis>` and have a story that demonstrates the behavior of the "ignore `<up_axis>`" feature?

defunctzombie:
From the slack discussion I thought we wanted z to be "up"? Is the story showing that we've correctly handled z as up and then applied some transformation?

jhurliman:
> From the slack discussion I thought we wanted z to be "up"? Is the story showing that we've correctly handled z as up and then applied some transformation?

The original model was a COLLADA file with y-up, converted into all the mesh formats we support. The storybook demonstrates that we can take a y-up model and make it visually appear in Studio the same way it appeared in the original DCC tool and other model viewers. Imagine the "+y" is the head of a humanoid model: the head is correctly placed upright here (after the relevant fixes landed).

A separate storybook demonstrating the "ignore <up_axis>" feature is working correctly would also be useful.

packages/studio-base/src/panels/ThreeDeeRender/ModelCache.ts
```diff
@@ -207,6 +224,10 @@ async function loadOBJ(
   const group = objLoader.parse(text);
   manager.itemEnd(url);
 
+  // THREE.js uses Y-up, while Studio follows the ROS
+  // [REP-0103](https://www.ros.org/reps/rep-0103.html) convention of Z-up
+  group.rotateX(Math.PI / 2);
```
snosenzo, line 229: do we do this to all objects that act this way? why don't we use something like: https://threejs.org/docs/#api/en/core/Object3D.DefaultUp

jhurliman: > do we do this to all objects that act this way?

We do this to all imported meshes from `THREE.*Loader` classes, so yes.

> why don't we use something like `Object3D.DefaultUp`

W do use `Object3D.DefaultUp`; see the global side effect in `Renderer.ts`. That doesn't fix `THREE.*Loader` mesh importing.

If you think about how geometry gets into the 3D scene, there are two main paths. We either procedurally generate it for markers, occupancy grids, camera image planes, etc. or we import user meshes. In the procedural generation cases we are either generating directly in a Z-up orientation, or we use THREE.js helper methods and apply rotations. For user imported meshes, we need to consider what the up orientation is for the source data (generally Y-up), and apply a transformation if needed to match our scene.

snosenzo: got it, thanks for the explanation!

PR #4504, opened 2022-09-28
registerMessageTransformer with pseudo topic

defunctzombie:
**User-Facing Changes**
TBD

**Description**
TBD


defunctzombie:
We've decided to go with the existing topic approach since that is more natural from the Studio end-user perspective

PR #4506, opened 2022-09-29
Refactor TextMiddleTruncate to use MUI Tooltip

2metres:
**User-Facing Changes**
None

**Description**
Also cleans up makeStyles code slightly


packages/studio-base/src/components/TextMiddleTruncate.tsx
```diff
@@ -45,18 +40,15 @@ type Props = {
...
   testShowTooltip?: boolean;
 };
 
 export default function TextMiddleTruncate({
   tooltips,
   text,
   endTextLength,
-  style,
   testShowTooltip,
+  ...rest
```
defunctzombie, line 51: There are no other rest parameters in `Props` so why capture them here?

packages/studio-base/src/components/TextMiddleTruncate.tsx
```diff
@@ -68,13 +60,13 @@ export default function TextMiddleTruncate({
   const endText = text.substring(startTextLen);
 
   const elem = (
-    <div className={classes.STextMiddleTruncate} style={style}>
-      <div className={classes.SStart}>{startText}</div>
-      <div className={classes.SEnd}>{endText}</div>
-    </div>
+    <Stack direction="row" justifyContent="flex-start" {...rest}>
```
defunctzombie, line 63: The `Props` type does not extend any Stack props so there is no need for rest here since a caller would not have any props to pass in that were not consumed.

2metres: Ah sorry meant to extend the props with StackProps

2metres: I think it's mainly for passing classes/styles through, might do that instead

defunctzombie (approved):

PR #4507, opened 2022-09-29
Add registerMessageConverter to the extension api

defunctzombie:
**User-Facing Changes**
A one-liner for the release notes:
> Add a new kind of extension for registering message converters to convert from one schema to another.

The user-facing changes are described in detail within this website PR: https://github.com/foxglove/website/pull/850

**Description**
Adds a new function `registerMessageConverter` to the `ExtensionContext`. This function allows a for creating extensions to tell Studio how to convert one schema to another. Such extensions are useful for adapting studio visualizations to display custom messages. For example converting a custom GPS message into `foxglove.LocationFix` so it can be visualized in the Map panel.

```typescript
registerMessageConverter({
  fromSchemaName: "my.CustomGpsMessage",
  toSchemaName: "foxglove.LocationFix",
  converter: (message) => { ... }
});
```

Registering a message converter alters what is possible for panels to subscribe to. Assuming the above message converter is registered a few new things are available to extension panels:

A new `additionalSchemaNames` field is available for any topic having the `my.CustomGpsMessage` schema. So if we have a topic `some/topic` that has the type `my.CustomGpsMessage`, we'd have an entry for `foxglove.LocationFix` in the `additionalSchemaNames` array for the topic.

```typescript
{
  name: "some/topic",
  schemaName: "my.CustomGpsMessage",
  additionalSchemaNames: ["foxglove.LocationFix"],
}
```

A panel can then subscribe to the topic and request the _converted_ messages by using the `convertTo` subscriber option.

```
context.subscribe([{ topic: "some/topic", convertTo: "foxglove.LocationFix" }]);
```

And finally, any `MessageEvent` the panel receives as a result of a convertTo subscription, either in `currentFrame` or `allFrames`, would have a the converted message in the `message` field and a new field `originalMessageEvent` which contains the original un-converted message event.

The logic for these additions is implemented entirely within the `PanelExtensionAdapter` `buildRenderState` function and available only to panels built using the extension api.

Fixes: #4730


defunctzombie:
image

Not sure if this is new to this PR or not

packages/studio-base/src/panels/Map/MapPanel.tsx
```diff
@@ -171,7 +176,20 @@ function MapPanel(props: MapPanelProps): JSX.Element {
...
-    return topics.filter(topicMessageType).map((topic) => topic.name);
+    return filterMap(topics, (topic) => {
+      if (isSupportedSchema(topic.schemaName)) {
+        return topic;
+      }
+
+      if (topic.additionalSchemaNames) {
+        for (const schemaName of topic.additionalSchemaNames) {
+          if (isSupportedSchema(schemaName)) {
+            return { name: topic.name, schemaName, datatype: schemaName };
```
jtbandes, line 187: Doubt we need both datatype and schemaName here? Maybe we should just form the Subscription here instead of making a list of topics and then iterating it again to make subscriptions? Or at least call the property `convertTo` here

packages/studio-base/src/players/types.ts
```diff
@@ -268,7 +268,7 @@ export type SubscriptionPreloadType =
 
 // Represents a subscription to a single topic, for use in `setSubscriptions`.
 export type SubscribePayload = {
-  // The topic name to subscribe to.
+  // The topic name to subscribe to
```
jtbandes, line 271: Using `/** ... */` comments will provide better VS Code integration

packages/studio/index.d.ts
```diff
@@ -55,11 +55,26 @@ declare module "@foxglove/studio" {
      * i.e. `package.Message` in protobuf-like serialization or `pkg/Msg` in ROS systems.
      */
     schemaName: string;
+
+    /**
+     * Lists any additional schema names available for subscribers on the topic. When subscribing to
+     * a topic, the panel can request messages be automatically converted from schemaName into one
+     * of the additionalSchemaNames.
```
jtbandes, line 62:
```suggestion
     * of the additionalSchemaNames using the convertTo option.
```

packages/studio/index.d.ts
```diff
@@ -55,11 +55,26 @@ declare module "@foxglove/studio" {
...
+     */
+    additionalSchemaNames?: string[];
```
jtbandes, line 64: Naming suggestion: `convertibleToSchemaNames`?

packages/studio-base/src/providers/ExtensionCatalogProvider.tsx
```diff
@@ -19,14 +19,21 @@ import { ExtensionInfo, ExtensionNamespace } from "@foxglove/studio-base/types/E
 
 const log = Logger.getLogger(__filename);
 
-async function registerExtensionPanels(
+type ContributionPoints = {
+  panels: Record<string, RegisteredPanel>;
+  messageConverters: RegisterMessageConverterArgs[];
```
jtbandes, line 24: If this were `RegisterMessageConverterArgs<any, unknown>` you wouldn't need the `as` cast below

packages/studio/index.d.ts
```diff
@@ -342,11 +364,21 @@ declare module "@foxglove/studio" {
     initPanel: (context: PanelExtensionContext) => void;
   };
 
+  export type RegisterMessageConverterArgs<Src = unknown, Dest = unknown> = {
+    fromSchemaName: string;
+    toSchemaName: string;
+    converter: (msg: Src) => Dest;
```
jtbandes, line 370: naming idea: `convert`?

packages/studio/index.d.ts
```diff
@@ -342,11 +364,21 @@ declare module "@foxglove/studio" {
...
+
   export interface ExtensionContext {
     /** The current _mode_ of the application. */
     readonly mode: "production" | "development" | "test";
 
     registerPanel(params: ExtensionPanelRegistration): void;
+
+    registerMessageConverter<Src = unknown, Dest = unknown>(
+      args: RegisterMessageConverterArgs<Src, Dest>,
+    ): void;
```
jtbandes, line 381: I still don't really get why this is a generic function. It seems like it would be fine to _always_ use `Src = any` and `Dest = unknown`, so you could just bake those into the type definition. What benefit is this providing?

packages/studio-base/src/components/PanelExtensionAdapter/renderState.test.ts
```diff
@@ -0,0 +1,72 @@
...
+    });
+
+    expect(state).toEqual({
+      topics: [
+        { name: "test", schemaName: "schema", datatype: "schema", additionalSchemaNames: ["more"] },
+      ],
+    });
+  });
+
+  it("should avoid conversion if the topic schema is already the desired convertTo schema", () => {
```
jtbandes, line 36: Is it allowed to subscribe to the original schema _and_ converted schema for the same topic? Should we have some tests around this (either that it produces some kind of error, or that it works in whatever way we want it to work)?

packages/studio-base/src/components/PanelExtensionAdapter/renderState.test.ts
```diff
@@ -0,0 +1,72 @@
...
+      currentFrame: undefined,
+      colorScheme: undefined,
+      globalVariables: {},
+      hoverValue: undefined,
+      sortedTopics: [{ name: "test", schemaName: "schema" }],
+      subscriptions: [],
+      messageConverters: [
+        {
+          fromSchemaName: "schema",
+          toSchemaName: "more",
```
jtbandes, line 23: Would be nice to add tests for collisions inside the convertersByKey map, e.g. `{ fromSchemaName: "sche", toSchemaName: "mamore" }`

packages/studio-base/src/components/PanelExtensionAdapter/renderState.test.ts
```diff
@@ -0,0 +1,72 @@
...
+          converter: () => {},
+        },
+      ],
+    });
+
+    expect(state).toEqual({
+      topics: [
+        { name: "test", schemaName: "schema", datatype: "schema", additionalSchemaNames: ["more"] },
+      ],
```
jtbandes, line 32: Can we test that the converter is actually run and produces the correct new message with the originalMessageEvent?

packages/studio/index.d.ts
jtbandes:
```suggestion
    additionalSchemaNames?: readonly string[];
```

defunctzombie: I've renamed to `convertibleTo`

packages/studio-base/src/components/PanelExtensionAdapter/renderState.ts
```diff
@@ -113,17 +141,105 @@ function initRenderStateBuilder(): BuildRenderStateFn {
...
+          topics.push(newTopic);
+        }
+
+        renderState.topics = topics;
         prevSortedTopics = sortedTopics;
       }
     }
 
+    // Update the mapping of converters.
+    // This needs to happen _after_ the above topics processing which re-runs if the message converters have changed,
```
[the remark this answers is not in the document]
defunctzombie: Added more comments to this to clarify. It is because setting `prevMessageConverters` would ruin the detection topics uses to see if the message converters have changed. Arguably this is somewhat fragile and a boolean like `messageConvertersChanged` should be used

packages/studio-base/src/components/PanelExtensionAdapter/renderState.test.ts
defunctzombie: 😮

packages/studio-base/src/components/PanelExtensionAdapter/renderState.test.ts
defunctzombie: One of the new tests checks this.

packages/studio-base/src/components/PanelExtensionAdapter/renderState.test.ts
defunctzombie: added a test and fixed the behavior. Amazing potential bug.

packages/studio/index.d.ts
defunctzombie: Don't feel strongly - haven't changed it.

packages/studio-base/src/components/PanelExtensionAdapter/renderState.ts
```diff
@@ -113,17 +141,105 @@ function initRenderStateBuilder(): BuildRenderStateFn {
...
         shouldRender = true;
-        renderState.topics = sortedTopics.map(({ name, schemaName }) => ({
-          name,
-          datatype: schemaName,
-          schemaName,
-        }));
+
+        const topics: Topic[] = [];
+
+        for (const topic of sortedTopics) {
```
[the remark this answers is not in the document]
defunctzombie: Which structure? The `sortedTopics` variable? Could you clarify the ask here.

packages/studio-base/src/components/PanelExtensionAdapter/renderState.ts
```diff
@@ -113,17 +141,105 @@ function initRenderStateBuilder(): BuildRenderStateFn {
...
+
+            // find any converters that can convert _from_ the schema name of the topic
+            // the _to_ names of the converter become additional schema names for the topic entry
+            for (const converter of messageConverters) {
+              if (converter.fromSchemaName === topic.schemaName) {
+                if (!additionalSchemaNames.includes(converter.toSchemaName)) {
+                  additionalSchemaNames.push(converter.toSchemaName);
+                }
+              }
+            }
```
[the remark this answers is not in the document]
defunctzombie: I think we did a benchmark recently and found that small sets and string arrays were the same tho I imagine memory layout wise the small array is a simpler structure. If we imagine these are always size 1 or 2 my intuition would tell me the array would win?

packages/studio/index.d.ts
defunctzombie: I want the user to be able to write an extension that looks like this:

```
type MyCustomType = { a: number };
context.registerMessageConverter({
  from: "",
  to: "",
  converter: (msg: MyCustomType) => {
  },
});
```

Specifically they should be able to tell the compiler that `msg` is type `MyCustomType` rather than having to cast `msg` within their converter body.

A future enhancement that I would consider would be a way we could detect a foxglove message type in the "toSchemaName" value and provide return value checking. For now folks can install our @foxglove/schemas package.

packages/studio-base/src/providers/ExtensionCatalogProvider.tsx
defunctzombie: I want to avoid `any` in the typings file and use `unknown` as that requires the user to acknowledge the lack of type information.

packages/studio/index.d.ts
jtbandes: Doesn’t that code still work fine with the change I suggested?

packages/studio-base/src/providers/ExtensionCatalogProvider.tsx
jtbandes: If you want them to be able to write `converter: (msg: MyCustomType) => …` as stated below, then it sounds like you **don’t** want them to acknowledge the lack of type information.

packages/studio-base/src/components/PanelExtensionAdapter/renderState.ts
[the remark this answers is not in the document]
jtbandes: Just pointing out that it used to be `topics = sortedTopics.map(...)` and is now `const topics = []; for (const topic of sortedTopics) { ... topics.push() }`. The former feels more readable IMO.

packages/studio-base/src/components/PanelExtensionAdapter/renderState.test.ts
```diff
@@ -0,0 +1,258 @@
...
+            receiveTime: { sec: 0, nsec: 0 },
+            sizeInBytes: 1,
+          },
+        },
+      ],
+    });
+  });
+
+  // Test that the correct converter is run when fromSchema + toSchema produce the same string
+  // for two different converters.
```
jtbandes, line 183: Not really seeing how the problem was fixed here – if the key is `from + "." + to` then `from: "a.", to: "b"` and `from: "a", to: ".b"` produce the same key. Would be nice to actually check for collisions.

packages/studio-base/src/components/PanelExtensionAdapter/renderState.ts
```diff
@@ -113,17 +142,111 @@ function initRenderStateBuilder(): BuildRenderStateFn {
...
+    // run again if the message converters have changed.
+    //
+    // And if this runs after the currentFrame and allFrames handling then convertersByKey will be
+    // from the previous converters.
+    if (messageConverters !== prevMessageConverters) {
+      convertersByKey.clear();
+
+      if (messageConverters) {
+        for (const converter of messageConverters) {
+          const key = converter.fromSchemaName + "." + converter.toSchemaName;
```
jtbandes, line 197: still want this to be in a helper function so the key formatting isn't done in multiple places

jtbandes (approved):
Looking pretty good. Left more comments about the outstanding issue with key collisions and some refactoring suggestions, will leave it up to you how to handle those.

packages/studio-base/src/components/PanelExtensionAdapter/renderState.test.ts
defunctzombie: So any character that is valid in a schema name would be a problem to use for concat (tho I had hoped that the use of `.` would mean it would not appear in practice in the schema start or finish. What should happen when there is a collision? There's no good way to surface this error.

packages/studio-base/src/components/PanelExtensionAdapter/renderState.ts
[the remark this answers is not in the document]
defunctzombie: Ah yea. 👍 I can make that change to use `map`.

packages/studio-base/src/providers/ExtensionCatalogProvider.tsx
defunctzombie: @jtbandes and advice for avoiding `any` in the type definition file?

defunctzombie:
@jtbandes Responded to feedback. Take a look and merge if you are good with the current state. We can continue to make improvements to the implementation if we are happy with the public API stability.

packages/studio-base/src/panels/Map/MapPanel.tsx
defunctzombie: I missed this comment on the PR. Yea I don't think we need datatype here especially since this is a legacy name. I'll make a follow-up PR.

PR #4519, opened 2022-09-30
Revert corepack

jhurliman:
**User-Facing Changes**
- None

**Description**
Creating a revert branch so I can continue developing Studio

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,7 +51,6 @@ jobs:
       - uses: actions/setup-node@v3.5.0
         with:
           node-version: 16.17
-      - run: corepack enable yarn
 
       - name: Restore cache
         uses: actions/cache@v3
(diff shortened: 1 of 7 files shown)


defunctzombie:
We've decided to revert John's computer instead.

jhurliman:
The issue hasn't been resolved yet. It looks like installing both `node` and `yarn` with homebrew on MacOS results in the Studio dev environment not working.

defunctzombie:
> The issue hasn't been resolved yet

What is the issue you want to resolve? There are many ways users can end up with weird installs. It seems sensible that macOS users install `node` and `yarn` via brew so we can amend the warning message to reference such a setup and recommend folks remove any installed yarn. The recommended install instructions for yarn say to use `corepack` (https://yarnpkg.com/getting-started/install#install-corepack) and not brew. In fact no instructions say to use brew.

We are not reverting the corepack stuff but I am happy to review a PR to improve the warning message.

Issue #4509, opened 2022-09-29
Extension in organisation disappears when foxglove opens without internet connection


bertaveira:
**Description**
I have noticed that sometimes the extension that we have installed in our organisation does not show up on the desktop version of foxglove. This seems to happen sometimes when we open foxglove without internet on a computer that previously had used the extension and is still logged in.

The quick fix is always to connect to a hotspot, open foxglove and then disconnect the hotspot.

- Version: 1.26
- OS: MacOS

foxymiles:
When you say it doesn't show up do you mean you get this error message or that you're unable to use panels from that extension in layouts?

<img width="344" alt="Screen Shot 2022-09-29 at 8 12 04 AM" src="https://user-images.githubusercontent.com/93935560/193040708-90d13084-3962-4a05-97d7-9e659783ceb9.png">

bertaveira:
No I get an error like this:
<img width="400" alt="image" src="https://user-images.githubusercontent.com/9644729/193043233-3dfcae39-cd9a-42d3-9bc6-1f800d382185.png">

And on the panel sidebar the extension panel is no longer there. Local installed extensions are there and work but organisations extensions not

foxymiles:
Ok thanks for the clarification. When you don't have a connection what do you see in the extension sidebar?

bertaveira:
Oh sorry. Yes I see the error you shared in the screenshot

foxymiles:
I'm trying this to reproduce:

1. Start a new session using a layout with a panel from an org extension.
2. Turn off the network on my machine.
3. Reload.

In this case the layout and panel still load ok.

When you say you open it without internet, does that mean that you have a live network connection but it's not able to talk to the internet or is the network completely disabled?

bertaveira:
If I open it with internet it load the extension properly. Then if I disable wifi and open again it still works. Then a day later I open again still without internet and the extension is gone. It does not happen every time but it seems that after a while the extension gets removed and unless it has internet it is no longer able to load.

foxymiles:
Ok thanks for the extra details. I'm looking into it now.

bertaveira:
Ok today something even weirder happened. I opened foxglove with internet connection and it loaded the extension fine for 5 seconds and then it refreshed and the extension and the panel was gone. Restarting foxglove fixed it.

bertaveira:
[...]

PR #4521, opened 2022-09-30
Use case insensitive name comparisons for org extension sync operations

foxymiles:
**User-Facing Changes**
Fixes an issue with org extension syncing.

**Description**
Use case insensitive name comparisons when syncing org extensions.

We lowercase the extension name when we use it as an identifier in layouts to make it easy to do a fast case insensitive lookup for loading panels. But for syncing purposes we need to do a case insensitive comparison of the extension name in case the original name had any uppercase characters.

Fixes https://github.com/foxglove/studio/issues/4509

### DIFF
--- a/packages/studio-base/src/components/OrgExtensionRegistrySyncAdapter.test.tsx
+++ b/packages/studio-base/src/components/OrgExtensionRegistrySyncAdapter.test.tsx
@@ -58,6 +58,7 @@ describe("Private registry sync adapter", () => {
     getExtensions.mockReturnValue([
       { name: "id1" },
       { name: "id2" },
+      { name: "mixedCaseName", activeVersion: "1" },
       { name: "private-installed-1", activeVersion: "2" },
       { name: "private-installed-2", activeVersion: "1" },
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/components/OrgExtensionRegistrySyncAdapter.test.tsx
```diff
@@ -70,6 +71,7 @@ describe("Private registry sync adapter", () => {
       getExtensions: jest
         .fn()
         .mockResolvedValue([
+          fakeExtension({ namespace: "org", name: "mixedcasename", version: "1" }),
```
defunctzombie, line 74: What is mixed case about `mixedcasename` ? That is all lower case.

defunctzombie (approved):
Should we have a check in the API that requires lowercase extension names?

foxymiles:
> Should we have a check in the API that requires lowercase extension names?

Not sure. Do we want to allow people to use them? Mixed case is perhaps slightly more descriptive.

PR #4522, opened 2022-09-30
Fix PanelList keyboard interactions

### LINKED ISSUE #4500: Change panel dropdown doesn't let me type some characters
image

I can't type `c`, `s`, `f`, or `r` as the first character in the search box, because it highlights menu entries in the top level menu.

foxymiles:
**User-Facing Changes**
Fixes an issue with type to search in the PanelList component

**Description**
Prevents keyboard events from propagating out of the PanelList text field and triggering menu events in the parent menu.

Also fixes a React forwardRef error.

Fixes #4500

### DIFF
--- a/packages/studio-base/src/components/PanelList/index.tsx
+++ b/packages/studio-base/src/components/PanelList/index.tsx
@@ -291,7 +291,7 @@ function verifyPanels(panels: readonly PanelInfo[]) {
   }
 }
 
-function PanelList(props: Props): JSX.Element {
+const PanelList = React.forwardRef<HTMLDivElement, Props>((props: Props, ref) => {
   const [searchQuery, setSearchQuery] = React.useState("");
   const [highlightedPanelIdx, setHighlightedPanelIdx] = React.useState<number | undefined>();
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/components/PanelList/index.tsx
```diff
@@ -395,6 +395,11 @@ function PanelList(props: Props): JSX.Element {
 
   const onKeyDown = React.useCallback(
     (e: React.KeyboardEvent) => {
+      // Prevent key down events from triggering the parent menu, if any.
+      if (e.key !== "Escape") {
```
snosenzo, line 399: does this interrupt browser shortcuts with ctrl and shift?

snosenzo: ah wait I tested, no it doesn't 👍

snosenzo (approved):

jtbandes (approved):

packages/studio-base/src/components/PanelList/index.tsx
foxymiles: Good thinking to test that.

PR #4523, opened 2022-09-30
Make copybutton text only created on click to avoid perf implications 

snosenzo:
**User-Facing Changes**
 - raw message panel being open will no longer hang when loading new large messages

**Description**
 - changed `value: string` to `copyText: () => string` so that JSON.stringify isn't called if the value isn't copied.

### DIFF
--- a/packages/studio-base/src/components/CopyButton.tsx
+++ b/packages/studio-base/src/components/CopyButton.tsx
@@ -19,7 +19,7 @@ import clipboard from "@foxglove/studio-base/util/clipboard";
 
 export default function CopyButton(
   props: PropsWithChildren<{
-    copyText: () => string;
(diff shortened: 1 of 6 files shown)


packages/studio-base/src/components/CopyButton.tsx
```diff
@@ -19,7 +19,7 @@ import clipboard from "@foxglove/studio-base/util/clipboard";
 
 export default function CopyButton(
   props: PropsWithChildren<{
-    value: string;
+    copyText: () => string;
```
jtbandes, line 22: nit: since copy is a verb, it sounds like this function is supposed to copy the text. maybe `getText`?

jtbandes (approved):
Thanks!

PR #4525, opened 2022-09-30
Add settings toggle to allow disabling TF preloading

jtbandes:
**User-Facing Changes**
Added a setting to 3D panel to allow disabling transform preloading.

**Description**
We don't currently handle it well when a dataset reaches the CoordinateFrame capacity limit (currently 10k TF messages). This change allows the user to opt out of preloading with a setting.

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
@@ -129,6 +129,8 @@ export type RendererConfig = {
       lineWidth?: number;
       /** Color of the connecting line between child and parent frames */
       lineColor?: string;
+      /** Enable transform preloading */
+      enablePreloading?: boolean;
     };
     /** Toggles visibility of all topics */
     topicsVisible?: boolean;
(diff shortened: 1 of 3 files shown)


defunctzombie:
I'm inclined to think this is a useful toggle generally since this question does come up even when we debug things and want to see what happens when we turn off tf preloading. One area where we need to give some thought is whether it is worth showing a status notice or some hint in the panel itself when preloading is off. I can see myself ending up in a world where I've turned off preloading and forget about it.

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -129,6 +129,8 @@ export type RendererConfig = {
       lineWidth?: number;
       /** Color of the connecting line between child and parent frames */
       lineColor?: string;
+      /** Enable transform preloading */
+      enablePreloading?: boolean;
```
defunctzombie, line 133: Worth calling this `enableTransformPreloading` and similarly for the user-facing label making it clear this is for preloading.

Maybe worth putting these data loading settings in their own section under settings.

jtbandes: This is in a section underneath Transforms.

jhurliman (approved):

PR #4533, opened 2022-10-01
Add WorkerIterablePlayer

defunctzombie:
**User-Facing Changes**
None

**Description**
Add a new WorkerIterablePlayer which runs the player loop in a web worker. This moves all data source handling (decompression, reading, deserialization) into a worker. The player state messages cross the worker boundary.

Only .bag and .mcap files are run in workers in this PR.


defunctzombie:
My experience with playing the nuscenes mcap file using this PR is that there is more jitter in the playback.

defunctzombie:
While this does remove the decompression from the main thread, this still keeps decompression on the same thread as the playback loop which means deserializing and decompressing impacts the playback performance.

PR #4538, opened 2022-10-03
Add WorkerIterableSource

defunctzombie:
**User-Facing Changes**
None

**Description**
Runs a source in a web worker

### DIFF
--- a/packages/mcap-support/src/parseChannel.ts
+++ b/packages/mcap-support/src/parseChannel.ts
@@ -169,7 +169,7 @@ export function parseChannel(channel: Channel): ParsedChannel {
             `Buffer not large enough: expected ${size} bytes, got ${data.byteLength}`,
           );
         }
-        return reader.readMessage(data);
+        return reader.readMessage(data).toObject();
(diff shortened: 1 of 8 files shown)


jhurliman:
👍🏼 major speedup for high bandwidth rosbags

packages/studio-base/src/players/IterablePlayer/foxglove-data-platform/DataPlatformIterableSource.test.ts
```diff
@@ -88,7 +88,6 @@ describe("DataPlatformIterableSource", () => {
       [
         expect.objectContaining({
           params: {
-            type: "by-device",
```
defunctzombie, line 91: @wimagguc Confirming that the api does not actually need a `type` argument in the API request.

defunctzombie:
Cookie auth needs testing - since cookie auth is not used during local development I've only tested device code auth.

foxymiles (approved):
Nice work!

PR #4542, opened 2022-10-03
Transform benchmark

jhurliman:
**User-Facing Changes**
- [3D] Improve performance for high volume transforms

**Description**
- Add a benchmark for transforms
- Fast path in binarySearch for `key` being outside the bounds of the ArrayMap

Screen Shot 2022-10-03 at 11 14 42 AM

### DIFF
--- a/packages/den/collection/ArrayMap.ts
+++ b/packages/den/collection/ArrayMap.ts
@@ -94,6 +94,13 @@ export class ArrayMap<K, V> {
     let left = 0;
     let right = list.length - 1;
 
+    // Quick checks to see if key is outside the bounds of the list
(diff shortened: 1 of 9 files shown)


jtbandes (approved):

packages/den/collection/ArrayMap.ts
```diff
@@ -94,6 +94,13 @@ export class ArrayMap<K, V> {
     let left = 0;
     let right = list.length - 1;
 
+    // Quick checks to see if key is outside the bounds of the list
```
defunctzombie, line 97:
```suggestion
    // If key is outside the bounds of the list we can avoid looping
```

PR #4544, opened 2022-10-03
Support foxglove.Grid

### LINKED ISSUE #4465: Support foxglove.Grid
Add support for `foxglove.Grid` non-ROS schemas within the 3D Panel.
https://foxglove.dev/docs/studio/messages/grid

snosenzo:
**User-Facing Changes**
 - enable visualization of Foxglove.Grid topic data types in the 3D panel

**Description**
 - Existing issues with this implementation: 
    - ~~converted "/map" OccupancyGrid message takes much more time to update the exture because of the CPU color converters. Do we want this implemented with shaders for this initial pass, sounds like it's a TODO for right now for Pointclouds.~~ Should be done in shaders in future -- similar to point clouds
    - ~~There's some pretty bad z-fighting without `logarithmicDepthbuffer` enabled -- not sure what we want the solution to be here. I don't know why it's acting differently from the OccupancyGrid~~ This might be added as a setting in the future
 - Adapted a lot of code from the `OccupancyGrid` and `PointCloudsAndLaserScans` renderables
 - tried to genericize and move ColorMode/field utility functions. I did not make the pointclouds use the settings nodes because I didn't want to break anything in this already large PR, so I'll do it in a followup.
 - created `Grid`s stories

Fixes #4465


packages/studio-base/src/panels/ThreeDeeRender/renderables/FoxgloveGrid.ts
```diff
@@ -0,0 +1,587 @@
...
+): void {
+  if (SKIP_MIN_MAX.includes(settings.colorMode)) {
+    return;
+  }
+
+  const [numericMin, numericMax] = NumericTypeMinMaxValueMap[numericType];
+  let minColorValue = settings.minValue ?? numericMin;
+  let maxColorValue = settings.maxValue ?? numericMax;
+  if (settings.minValue == undefined || settings.maxValue == undefined) {
+    for (let i = 0; i < cellCount; i++) {
```
snosenzo, line 563: Point clouds was doing this, but it might be overkill -- happens on every new message, not too intensive to minmax though

jhurliman: Having to min/max incoming data in point clouds is a (the?) major barrier to improving the performance of point clouds. If it wasn't for that requirement, and heterogeneous data types were supported for interleaved buffers in three.js, we could simply upload the whole `data` blob for incoming point clouds to the GPU and do all of the work in a shader.

We should avoid introducing this performance bottleneck for other renderable types if at all possible.

jhurliman: PointCloud field values generally come from sensors, where the range is unknown. For a Grid, it's more reasonable to expect the user to know the min/max values and to default to [0.0-1.0].

jtbandes: Maybe the default should be based on the field type, e.g. 0-1 for float, but 0-255 for uint8, etc.

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -342,6 +343,7 @@ export class Renderer extends EventEmitter<RendererEvents> {
       canvas,
       alpha: true,
       antialias: true,
+      logarithmicDepthBuffer: true,
```
jhurliman, line 346: This is likely what broke most of the 3D panel stories

packages/studio-base/src/panels/ThreeDeeRender/renderables/FoxgloveGrid.ts
snosenzo: Yeah, sounds good I'll update the NumericMin/max map to have better defaults for floats

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
snosenzo: Is it fine to keep this enabled then? or should I explore other options?

jhurliman: We definitely can't keep it enabled with the current state of the storybooks. Axis labels have strange depth ordering problems and some artifacts appeared on other renderables. We could introduce this as a user setting defaulting to off in a separate PR.

jtbandes:
Sorry I just gave you some merge conflicts with #4562 😬  Hopefully nothing too major. I can help resolve as well.

snosenzo:
yeah, nothing major, just with something I took out and put in `colors.ts`

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
snosenzo: good point, removed it

packages/studio-base/src/panels/ThreeDeeRender/renderables/FoxgloveGrid.ts
```diff
@@ -0,0 +1,523 @@
...
+            : undefined;
+        colorReader = getReader(field, stride, forceType);
+        if (!colorReader) {
+          const typeName = NumericType[type];
+          const message = `Grid field "${field.name}" is invalid. type=${typeName}, offset=${field.offset}, stride=${stride}`;
+          invalidFoxgloveGridError(this.renderer, renderable, message);
+          return false;
+        }
+      }
+    }
```
jtbandes, line 259: wonder if there's a way we can share all this logic with PointClouds...maybe it is different enough that it's not worth it?

packages/studio-base/src/panels/ThreeDeeRender/renderables/FoxgloveGrid.ts
```diff
@@ -0,0 +1,523 @@
...
+
+  const [numericMin, numericMax] = NumericTypeMinMaxValueMap[numericType];
+  const minColorValue = settings.minValue ?? numericMin;
+  const maxColorValue = settings.maxValue ?? numericMax;
+  output[0] = minColorValue;
+  output[1] = maxColorValue;
+}
+
+const NumericTypeMinMaxValueMap: Record<NumericType, [number, number]> = {
+  [NumericType.UNKNOWN]: [Number.NEGATIVE_INFINITY, Number.POSITIVE_INFINITY],
```
jtbandes, line 514: Does this ever actually get used? It seems like -∞ to ∞ would never produce any useful color values. Maybe just go with `[0, 1]` for unknown or omit it if it never gets used?

jtbandes (approved):

packages/studio-base/src/panels/ThreeDeeRender/stories/foxglove.Grid.stories.tsx
```diff
@@ -0,0 +1,522 @@
...
+import PanelSetup from "@foxglove/studio-base/stories/PanelSetup";
+
+import ThreeDeeRender from "../index";
+import { TransformStamped } from "../ros";
+import { QUAT_IDENTITY, rad2deg } from "./common";
+import useDelayedFixture from "./useDelayedFixture";
+
+export default {
+  title: "panels/ThreeDeeRender",
+  component: ThreeDeeRender,
```
jtbandes, line 17:
```suggestion
  component: ThreeDeeRender,
  parameters: { colorScheme: "light" },
```
removes the double-rendering of each story with two color schemes because the color scheme isn't really relevant to most 3d panel stories.

packages/studio-base/src/panels/ThreeDeeRender/renderables/FoxgloveGrid.ts
snosenzo: yeah I think, it's a bit hard w/ the Numeric vs PointField Types. it could be  consolidated, but I think I'd rather tackle that consolidation in the same PR that I make the Grid, VelodyneScans, and PointClouds use the same base colorMode settings node. I wanted to be careful not to break point clouds in this PR

packages/studio-base/src/panels/ThreeDeeRender/renderables/FoxgloveGrid.ts
snosenzo: hmm yeah, I'll go with 0, 1 for now, but If we want to omit it in the future then we'll probably want to throw an error for the unknown field. Not sure when the numericType would be unknown from the data creation perspective.

jhurliman:
PR needs a screenshot

snosenzo:
oop will update

PR #4545, opened 2022-10-04
Fix post-release workflow

defunctzombie:
**User-Facing Changes**
Dockerfile build works

**Description**
- Remove `cache: yarn` from setup-node. This does not work because corepack is not enabled and our check fails
- Add `corepack enable` to Dockerfile

### DIFF
--- a/.github/workflows/post-release.yml
+++ b/.github/workflows/post-release.yml
@@ -64,7 +64,6 @@ jobs:
       - uses: actions/setup-node@v3.5.0
         with:
           node-version: 16.17
-          cache: yarn
       - run: corepack enable yarn
 
       - run: yarn install --immutable
@@ -124,7 +123,6 @@ jobs:
         with:
           node-version: 16.17
           registry-url: https://registry.npmjs.org
-          cache: yarn
       - run: corepack enable yarn
 
       - run: yarn install --immutable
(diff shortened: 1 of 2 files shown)


jtbandes (approved):

.github/workflows/post-release.yml
```diff
@@ -64,7 +64,6 @@ jobs:
       - uses: actions/setup-node@v3.5.0
         with:
           node-version: 16.17
-          cache: yarn
```
amacneil, line 67: why no more caching? that's gonna make the release even slower

defunctzombie: We don't have caching on the studio builds and things still cache. The caching doesn't work because it tries to run yarn but our yarn wrapper fails when it detects its not run under corepack.

amacneil: That is because we are explicitly using `actions/cache` in the main builds, rather than the built in caching of `actions-setup-node`:

https://github.com/foxglove/studio/blob/main/.github/workflows/ci.yml#L57

You need one or the other if you want caching. I suppose caching is not critical for post-release workflow tho since it is not time sensitive.

defunctzombie: 👍 Do you think we should add `actions/cache` or avoid the cache to make sure we can build without it? If we think the cache is nice here I can make a PR to add it back - not hard.

amacneil: I think it's fine leaving the cache out. If we run into some timeout issues or something we can revisit. But speed is not critical for the post-release actions.

PR #4550, opened 2022-10-04
Use new event search API in studio

foxymiles:
**User-Facing Changes**
Connects the event search feature in studio to the new API in console, allowing for more flexible searches.

**Description**
Replace the built-in event search with a call to the new search API.

https://user-images.githubusercontent.com/93935560/193822906-3a284f86-eb67-4368-be36-a835a86f1a14.mov

### DIFF
--- a/packages/studio-base/src/components/DataSourceSidebar/EventsList.tsx
+++ b/packages/studio-base/src/components/DataSourceSidebar/EventsList.tsx
@@ -23,7 +23,6 @@ import {
   useTimelineInteractionState,
 } from "@foxglove/studio-base/context/TimelineInteractionStateContext";
 import { useAppTimeFormat } from "@foxglove/studio-base/hooks";
-import { EventsSelectors } from "@foxglove/studio-base/providers/EventsProvider";
 
 import { EventView } from "./EventView";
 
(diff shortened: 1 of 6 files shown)


packages/studio-base/src/components/DataSourceSidebar/EventsList.tsx
```diff
@@ -114,43 +112,15 @@ export function EventsList(): JSX.Element {
...
-    <Stack className={classes.root}>
+    <Stack className={classes.root} fullHeight>
       <AppBar className={classes.appBar} position="sticky" color="inherit" elevation={0}>
         <TextField
           variant="filled"
           fullWidth
           value={filter}
           onChange={(event) => setFilter(event.currentTarget.value)}
-          placeholder="Filter event metadata"
+          placeholder="Search by key:value"
```
snosenzo, line 123: is this the only way people can search now? not by just a value?

snosenzo (approved):

packages/studio-base/src/components/DataSourceSidebar/EventsList.tsx
foxymiles: In fact we currently treat a string without a colon as matching either key or value but I wasn't sure how to express this in a clear way in this tiny placeholder.

snosenzo: maybe just `Search by key, value, or key:value`? should still fit reasonably?

Issue #4548, opened 2022-10-04
Data platform `ds.importId` URLs cause studio to freeze browser tab


amacneil:
**Description**
Example URL from demo org:

https://studio.foxglove.dev/?ds=foxglove-data-platform&time=2018-07-24T03%3A28%3A47.647Z&ds.importId=g5kSXSVEysv9oKjj&ds.deviceId=dev_SUQ41s35DqPfGAoo&ds.start=2018-07-24T03%3A28%3A47.647Z&ds.end=2018-07-24T03%3A29%3A06.797Z

Clicking this link causes my tab to become unresponsive.

PR #4551, opened 2022-10-04
Fix browser tab freeze

wimagguc:
**User-Facing Changes**

Fixes an issue where the new `ds.importId` parameter could cause the browser tab to freeze.

**Github issue**

Fixes https://github.com/foxglove/studio/issues/4548


defunctzombie:
I don't follow how this fixes the freeze? It sounds like if I make the same request using a deviceId (removing importId) then I can still freeze my browser tab?

packages/studio-base/src/players/IterablePlayer/foxglove-data-platform/DataPlatformIterableSource.ts
```diff
@@ -239,7 +239,7 @@ export class DataPlatformIterableSource implements IIterableSource {
         api,
         parsedChannelsByTopic,
         params: {
-          ...(deviceId ? { deviceId } : { importId: importId! }),
+          ...(importId ? { importId } : { deviceId: deviceId! }),
```
defunctzombie, line 242: I would prefer if we extracted `params` to a variable and set this field correctly on it. I find this spread operator ternary hard to read.

packages/studio-base/src/players/IterablePlayer/foxglove-data-platform/streamMessages.ts
```diff
@@ -76,11 +75,19 @@ export async function* streamMessages({
 
   log.debug("streamMessages", params);
   const startTimer = performance.now();
+
+  const startString = params.start ? toRFC3339String(params.start) : undefined;
+  const endString = params.end ? toRFC3339String(params.end) : undefined;
+
+  const deviceId = "deviceId" in params ? params.deviceId : undefined;
```
defunctzombie, line 82: This looks a-typical to me. We know that `deviceId` is in `params` because the params type is  `DataPlatformSourceParameters` and has the deviceId. Is that field suppose to be optional in `DataPlatformSourceParameters`?

defunctzombie (approved):
👍

The `"foo" in bar` code looks suspicious to me (we don't have a lot of that in Studio and where we do have it, it is for user-defined types). Probably a nicer way to write that to avoid the `in` unless I misread the intent.

It would be good to add a comment to the streamMessages call highlighting that if importId is provided that is used instead of deviceId (if that's the intended behavior).

wimagguc:
> 👍
> 
> The `"foo" in bar` code looks suspicious to me (we don't have a lot of that in Studio and where we do have it, it is for user-defined types). Probably a nicer way to write that to avoid the `in` unless I misread the intent.
> 
> It would be good to add a comment to the streamMessages call highlighting that if importId is provided that is used instead of deviceId (if that's the intended behavior).

@foxymiles had a good suggestion on introducing `{ type: "by-device" ... }` on these discriminating union types, so we can use `x.type === "by-device" ? ...` instead of `"a" in foo ? x.a ...`

foxymiles:
@wimagguc I think we can take the typing further to clean up this logic even more. Please have a look at this branch:

https://github.com/foxglove/studio/tree/miles/stream-request-types

foxymiles:
Specifically this commit: https://github.com/foxglove/studio/commit/91a19441751bb5af0fa697fce9baca40f75048e1

foxymiles (changes requested):
Going in the right direction but I think we can still get more mileage out of the typing here.

wimagguc:
> Going in the right direction but I think we can still get more mileage out of the typing here.

Haha well, thanks for the review and essentially pulling it across the finish line. Everything nice and clean on your branch, I merely had to change one test after those changes... Thanks again

foxymiles (approved):
Code looks good now I think.

jtbandes:
Was there an answer to https://github.com/foxglove/studio/pull/4551#issuecomment-1267195608? What was actually the root cause of the freeze?

PR #4553, opened 2022-10-04
Retain most recent messages on each topic in map panel

foxymiles:
**User-Facing Changes**
Improves message handling behavior in the map panel.

**Description**
Keep the most recent messages on each topic in the map panel, only replacing them when we receive new messages on that topic.

### DIFF
--- a/packages/studio-base/src/panels/Map/MapPanel.tsx
+++ b/packages/studio-base/src/panels/Map/MapPanel.tsx
@@ -487,42 +487,46 @@ function MapPanel(props: MapPanelProps): JSX.Element {
       return;
     }
 
-    const groupedNavMessages = groupBy(currentNavMessages, (msg) => msg.topic);
-    for (const [topic, messages] of Object.entries(groupedNavMessages)) {
+    const navByTopic = groupBy(currentNavMessages, (msg) => msg.topic);
+    for (const [topic, messages] of Object.entries(navByTopic)) {
       const topicLayer = topicLayers.get(topic);
-      if (topicLayer) {
-        topicLayer.currentFrame.clearLayers();
-        const [fixEvents, noFixEvents] = partition(messages, hasFix);
-
-        const pointLayerNoFix = FilteredPointLayer({
-          map: currentMap,
-          navSatMessageEvents: noFixEvents,
-          bounds: filterBounds ?? currentMap.getBounds(),
-          color: darkColor(topicLayer.baseColor),
-          hoverColor: darkColor(topicLayer.baseColor),
-          showAccuracy: true,
-        });
-
-        const pointLayerFix = FilteredPointLayer({
-          map: currentMap,
-          navSatMessageEvents: fixEvents,
-          bounds: filterBounds ?? currentMap.getBounds(),
-          color: topicLayer.baseColor,
-          hoverColor: darkColor(topicLayer.baseColor),
-          showAccuracy: true,
-        });
-
-        topicLayer.currentFrame.addLayer(pointLayerNoFix);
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/panels/Map/MapPanel.tsx
```diff
@@ -482,36 +486,43 @@ function MapPanel(props: MapPanelProps): JSX.Element {
...
 
-      currentGeoMessages
-        .filter((message) => message.topic === topic)
-        .forEach((message) => addGeoJsonMessage(message, topicLayer.currentFrame));
+    const groupedGeoMessages = groupBy(currentGeoMessages, (msg) => msg.topic);
+    for (const [topic, messages] of Object.entries(groupedGeoMessages)) {
+      const topicLayer = topicLayers.get(topic);
+      if (topicLayer) {
+        topicLayer.currentFrame.clearLayers();
+        messages.forEach((msg) => addGeoJsonMessage(msg, topicLayer.currentFrame));
```
defunctzombie, line 524: Since there's no more `filter` call here, I think it would be nicer to change this `forEach` to a regular `for` loop.

packages/studio-base/src/panels/Map/MapPanel.tsx
```diff
@@ -482,36 +486,43 @@ function MapPanel(props: MapPanelProps): JSX.Element {
...
-        hoverColor: darkColor(topicLayer.baseColor),
-        showAccuracy: true,
-      });
-
-      topicLayer.currentFrame.addLayer(pointLayerNoFix);
-      topicLayer.currentFrame.addLayer(pointLayerFix);
+    const groupedNavMessages = groupBy(currentNavMessages, (msg) => msg.topic);
+    for (const [topic, messages] of Object.entries(groupedNavMessages)) {
+      const topicLayer = topicLayers.get(topic);
+      if (topicLayer) {
```
defunctzombie, line 492: nit: you can reduce the overall nesting of the loop by reversing the `if` condition here and using `continue`. The rest of the loop body will get one indent level back. Up to you if you find that more readable or not. Same for the loop below for geo messages.

defunctzombie (approved):
The main business logic change is to have a topmost loop over the messages by topic rather than each topic layer so you only clear the topic layer currentFrame corresponding to the topics that have new messages. 👍

packages/studio-base/src/panels/Map/MapPanel.tsx
foxymiles: Yeah I tend to like that style too, especially when you allow that to be on one line:

```typescript
if (!topicLayer) continue;
```

It's very idiomatic in swift code.

```swift
guard let name = getName() else { return }
```

packages/studio-base/src/panels/Map/MapPanel.tsx
```diff
@@ -342,7 +342,11 @@ function MapPanel(props: MapPanelProps): JSX.Element {
       setPreviewTime(renderState.previewTime);
 
       if (renderState.topics) {
-        setTopics(renderState.topics);
+        // Changing the topic list clears all map layers so we try to preserve reference identity
+        // if the contents of the topic list haven't changed.
+        setTopics((oldTopics) => {
+          return isEqual(oldTopics, renderState.topics) ? oldTopics : renderState.topics ?? [];
```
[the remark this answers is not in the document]
foxymiles: I wasn't able to make the story work without this. If I send a second fixture to PanelSetup without topics it gets treated as a new, empty topic list. If I send the same topic list it gets treated as a new topic list since it goes through a `sort` before it hits the panel.

PR #4554, opened 2022-10-04
Add datatype to MessageEvent

defunctzombie:
**User-Facing Changes**
Panel Extension API MessageEvents have a new `datatype` field.

**Description**
Prior to this change each MessageEvent only had a topic name. If a panel author wanted to determine the type of message within the MessageEvent they would need to keep a mapping of topic -> datatype.

This change adds the `datatype` field to the MessageEvent. This field contains the datatype of the message within the message event.

Fixes: https://github.com/foxglove/studio/issues/4561


jtbandes (approved):

jtbandes:
We currently have `datatype` (the string name) and `RosDatatypes` (the structure describing the fields and field types). I argued these should be called `schemaName` and `schemas` but was outvoted.

amacneil:
I see, we already have those in the extension API. I thought they only existed as internal tech debt. 

Seems like we should consolidate these and standardize on `schema`, although it would require some non breaking deprecations in the extension API.

defunctzombie:
> Seems like we should consolidate these and standardize on `schema`, although it would require some non breaking deprecations in the extension API.

`schema` is also overloaded - does it refer to the full descriptor (like the Schema record in mcap) or is it the name? IMO datatype is a good and appropriate term for the name of the type of data. `type` is a word that I believe is more commonly used in places where the value is a _string_ name for the type.

Some random links exploring this space:
- https://en.wikipedia.org/wiki/Data_type
- https://en.wikipedia.org/wiki/XML_schema
- https://github.com/rticommunity/ros-data-types
- https://www.ibm.com/docs/en/iis/9.1?topic=connector-message-schemas
- https://cloud.google.com/pubsub/docs/schemas
- https://en.wikipedia.org/wiki/Media_type

jtbandes:
> schema is also overloaded - does it refer to the full descriptor (like the Schema record in mcap) or is it the name?

We use schemaName to refer to the name elsewhere.

defunctzombie:
image

Our Studio docs use Datatype.

And schema name in other places.

image

And then "types" again in other places.

image

Schema Name is a concept that was introduced in mcap.

PR #4556, opened 2022-10-04
3D settings: filter topics by visibility

### LINKED ISSUE #4433: 3D settings: filter topics by visibility

snosenzo:
**User-Facing Changes**
 - enable filtering of topics list by visibility in the 3D panel

**Description**
 - made this generic in the NodeEditor, so anything in the settings tree can enable this to filter it's children by visibility
 - should be much faster than previous implementation because it doesn't rebuild the settings nodes whenever the settings change, just filters child nodes at the tree level

Fixes #4433

### DIFF
--- a/packages/studio-base/src/components/SettingsTreeEditor/NodeEditor.tsx
+++ b/packages/studio-base/src/components/SettingsTreeEditor/NodeEditor.tsx
@@ -134,9 +134,34 @@ function ExpansionArrow({ expanded }: { expanded: boolean }): JSX.Element {
 
 const makeStablePath = memoizeWeak((path: readonly string[], key: string) => [...path, key]);
 
+type SelectVisibilityFilterValue = "all" | "visible" | "invisible";
+const SelectVisibilityFilterOptions: { label: string; value: SelectVisibilityFilterValue }[] = [
+  { label: "List all", value: "all" },
+  { label: "List visible", value: "visible" },
+  { label: "List invisible", value: "invisible" },
+];
+function showVisibleFilter(child: DeepReadonly<SettingsTreeNode>): boolean {
+  // want to show children with undefined visibility
+  return child.visible !== false;
+}
+function showInvisibleFilter(child: DeepReadonly<SettingsTreeNode>): boolean {
+  // want to show children with undefined visibility
+  return child.visible !== true;
+}
+const SelectVisibilityFilterField = {
+  input: "select",
+  label: "Filter list",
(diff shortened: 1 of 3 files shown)


jhurliman (approved):

packages/studio-base/src/components/SettingsTreeEditor/NodeEditor.tsx
```diff
@@ -170,18 +204,27 @@ function NodeEditorComponent(props: NodeEditorProps): JSX.Element {
...
-      />
-    );
-  });
+  const filterFn =
+    state.visibilityFilter === "visible"
+      ? showVisibleFilter
+      : state.visibilityFilter === "invisible"
+      ? showInvisibleFilter
+      : undefined;
+  const childNodes = prepareSettingsNodes(children ?? {}).reduce((agg, [key, child]) => {
```
jtbandes, line 213: Instead of reduce with a conditional push, try using our filterMap function instead. It tends to be faster than this functional chaining kind of approach since it uses a raw for loop under the hood.

packages/studio/index.d.ts
```diff
@@ -597,6 +597,11 @@ declare module "@foxglove/studio" {
      * to the action handler.
      **/
     visible?: boolean;
+
+    /**
+     * Filter Children by visibility status
+     */
+    selectVisibilityFilterEnabled?: boolean;
```
jtbandes, line 604: What's the "select" part of this mean? Would just `visibilityFilterEnabled` make sense? Note that we also have an `enableFilter` on the top level tree. So maybe we should call it `enableVisibilityFilter`. cc @foxymiles for thoughts

foxymiles: I like `enableVisibilityFilter`.

snosenzo: just was kinda hinting that it enables the _selection_ of the visibility filter, but yeah it's probably too verbose

snosenzo: I usually like formatting config variables along the lines of <descriptor><valueword> because it filters down autocomplete sooner and in this case sounds less like an action, but not a big deal and I'll move forward with the suggestion.

jtbandes:
I think you'll need to merge latest main in order to fix the vercel error

snosenzo:
yeah something went weird w/ git confusing my old branch when merging, should be good now.

jtbandes (approved):

PR #4564, opened 2022-10-04
[3D] Don't preload future transforms

### LINKED ISSUE #4605: Preloaded transforms from the "future" should not be used to display data in 3D
I am seeing incorrect positions for my scene entities when viewing them in Studio. I expect the entities to be static until a new frame information arrives but one of my entities continues to update in the scene even tho no new frame information is published.

My setup has two topics: a `transforms` topic and an `objects`. On the `transforms` topic I publish a `foxglove.FrameTransform` message between "map" and "robot" frames. On the `objects` topic I publish two sphere entities. The red sphere is centered on the "map" frame and the green sphere is centered on the "robot" frame. I publish the scene entity update message every second.

My robot does not publish `transform` messages frequently and my robot's decisions are based on the last published transform. When I play back my recorded data in Studio I am seeing my robot entity continue moving relative to the map entity even when I've not published any transform updates.

Below are two screenshots showing the 3D panel and Raw message panel. You can see that there's been no new transform published for map -> robot yet the green sphere has moved! This results in me looking at a scene that does not match what my robot is doing. I've explicitly opted to publish transform messages matching the frame information my robot uses but I'm seeing a different visualization in Studio.

image

image

Mcap and layout:
[output.mcap.zip](https://github.com/foxglove/studio/files/9742650/output.mcap.zip)
[Default.json.zip](https://github.com/foxglove/studio/files/9742652/Default.json.zip)

jhurliman:
**User-Facing Changes**
- [3D] Transforms from the future are no longer used for TF interpolation

**Description**
The goal of preloading TFs is to get the full back history of TF messages in case one or a few TF messages were 
published early on and not republished or republished infrequently (such as static TFs). Preloading TFs from past the 
current playback time produces unexpected interpolation that deviates from how a live robot would behave.

Resolves #4605

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
+++ b/packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
@@ -21,7 +21,7 @@ import { DeepPartial } from "ts-essentials";
 import { useDebouncedCallback } from "use-debounce";
 
 import Logger from "@foxglove/log";
-import { isLessThan, Time, toNanoSec } from "@foxglove/rostime";
+import { fromNanoSec, isLessThan, Time, toNanoSec } from "@foxglove/rostime";
 import {
   LayoutActions,
   MessageEvent,
@@ -676,10 +676,17 @@ export function ThreeDeeRender({ context }: { context: PanelExtensionContext }):
       return;
     }
 
+    const timeLimit = fromNanoSec(currentTime ?? 0n);
+
     for (const message of preloadedMessages) {
       // Skip preloaded messages before the last receiveTime we've previously processed
       if (isLessThan(message.receiveTime, lastPreloadedMessageTimeRef.current)) {
         continue;
+      }
+
+      // Don't ingest future messages
+      if (!isLessThan(message.receiveTime, timeLimit)) {
+        break;
       }
 
       const datatype = topicsToDatatypes.get(message.topic);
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
```diff
@@ -676,12 +676,19 @@ export function ThreeDeeRender({ context }: { context: PanelExtensionContext }):
       return;
     }
 
+    const timeLimit = fromNanoSec(currentTime ?? 0n);
+
     for (const message of preloadedMessages) {
```
jtbandes, line 681: I don't think we should be iterating through the whole preloadedMessages array every time the currentTime changes

defunctzombie:
John and I had a brainstorm on this and we came up with an idea to try:

Preload only the first N transform messages and preload M transform messages before the current time. As more transforms arrive cycle out old ones but always keeping the first N transforms from start and M before current time active. The thinking was that this solves the common case of ensuring static transforms are present (but loading the first N transform messages) and being able to transform recent messages by having the M transforms before current time present.

If a message needs transforming that falls before the M oldest transform time show a topic error that transforms for that time are expired (or something like that). We can see how often this happens. If necessary we could load a small batch of transforms for this new transform time.

We would do the above _iff_ the total amount of transforms exceeds some upper bound we determine. The idea is that past this upper bound - working with the entire array no longer becomes practical for every frame rendering.

@jhurliman can double check what I've written and correct me if I've stated something incorrect

jtbandes:
Can this PR be closed since we've backed out TF preloading? We have an issue tracking it at: 
- https://github.com/foxglove/studio/issues/4657

Issue #4568, opened 2022-10-05
Filter list should be hidden when there are no child nodes


jhurliman:
<img width="359" alt="Screen Shot 2022-10-04 at 7 29 48 PM" src="https://user-images.githubusercontent.com/195374/193966663-5269f742-e220-474e-add4-250e1806959e.png">

jhurliman:
cc @snosenzo

PR #4569, opened 2022-10-05
[3D] Don't enable the visibility filter if the topic list is empty

jhurliman:
**User-Facing Changes**
- [Minor][3D] The visibility filter for topics is not shown if the topic list is empty

**Description**
A panel-specific fix for https://github.com/foxglove/studio/issues/4568

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
@@ -601,17 +601,18 @@ export class Renderer extends EventEmitter<RendererEvents> {
     this.customLayerActions.set(options.layerId, { action, handler });
 
     // "Topics" settings tree node
+    const topicsChildren = this.settings.tree()["topics"]?.children;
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -601,17 +601,18 @@ export class Renderer extends EventEmitter<RendererEvents> {
     this.customLayerActions.set(options.layerId, { action, handler });
 
     // "Topics" settings tree node
+    const topicsChildren = this.settings.tree()["topics"]?.children;
     const topics: SettingsTreeEntry = {
       path: ["topics"],
       node: {
-        enableVisibilityFilter: true,
+        enableVisibilityFilter: Object.keys(topicsChildren ?? {}).length > 0 ? true : undefined,
```
jtbandes, line 608: Seems like we should probably hide this for all nodes where enableVisibilityFilter is true and children is empty, not just in the 3D panel?

PR #4572, opened 2022-10-05
[3D] Avoid synchronously calling SceneExtension methods that may be overridden from the base class constructor

jhurliman:
**User-Facing Changes**
- None

**Description**
This fixes a potential bug that could cause undefined behavior in SceneExtension subclasses, by calling the subclass 
`settingsNodes()` method before the constructor ran.


jtbandes (approved):

jtbandes:
Did we encounter this bug actually occurring?

jhurliman:
@jtbandes in another branch

PR #4573, opened 2022-10-05
[3D] URDF joint viewer, manually modify joint positions

jhurliman:
**User-Facing Changes**
- [3D] URDFs display full joint information in the sidebar. Joint positions can be manually modified

**Description**

This updates the `Urdfs` SceneExtension to display a list of all joints for a URDF and all parsed information for each 
joint. A "manual" field for each joint controls the angle or position by modifying the offsetPosition or 
offsetEulerDegrees for the relevant frame.

Support for JointState is partially added, but the subscription won't work 
until we have a better way for scene extensions to signal whether a particular subscription should be on or off (in the 
case of Urdfs, we want to enable the subscription when there is at least one loaded URDF).

Screen Shot 2022-10-05 at 1 54 12 PM

https://user-images.githubusercontent.com/195374/194161621-6db8700d-b7fb-47fd-9aa1-ad43b7c7b0c8.mov

https://user-images.githubusercontent.com/195374/194161636-9652ebd3-ddb1-40c7-9d10-697dedbf0413.mov

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
@@ -148,6 +148,9 @@ export class Urdfs extends SceneExtension<UrdfRenderable> {
     super("foxglove.Urdfs", renderer);
 
     renderer.addTopicSubscription(TOPIC_NAME, this.handleRobotDescription);
+    // Note that this subscription will never happen because it does not appear as a topic in the
+    // topic list that can have its visibility toggled on. The ThreeDeeRender subscription logic
+    // needs to become more flexible to make this possible
     renderer.addDatatypeSubscriptions(JOINTSTATE_DATATYPES, this.handleJointState);
     renderer.on("parametersChange", this.handleParametersChange);
     renderer.addCustomLayerAction({
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
@@ -2,11 +2,14 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
+import { vec3 } from "gl-matrix";
 import { maxBy } from "lodash";
+import * as THREE from "three";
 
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
```diff
@@ -289,6 +367,23 @@ export class Urdfs extends SceneExtension<UrdfRenderable> {
...
+    const positions = msg.position ?? [];
+    const timestamp = toNanoSec(messageEvent.receiveTime);
+
+    for (let i = 0; i < names.length; i++) {
+      const name = names[i]!;
+      const position = positions[i] ?? 0;
+
+      const prevTimestamp = this.jointStates.get(name)?.timestamp;
+      if (prevTimestamp == undefined || timestamp >= prevTimestamp) {
+        this.jointStates.set(name, { timestamp, position });
```
jtbandes, line 382: Where does this end up actually modifying the 3D scene? Or does it not currently because of the subscription issue you described?

jhurliman: It doesn't right now, it's a read-only view. We would need additional UI to opt-in to JointState -> TF (effectively `robot_state_publisher`).

packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
```diff
@@ -269,16 +311,52 @@ export class Urdfs extends SceneExtension<UrdfRenderable> {
...
+      const joint = transformData.joint;
+      const frame = this.renderer.transformTree.getOrCreateFrame(transformData.child);
+      const frameKey = `frame:${frame.id}`;
+      const isAngular = joint.jointType === "revolute" || joint.jointType === "continuous";
+      const axis = new THREE.Vector3(joint.axis.x, joint.axis.y, joint.axis.z);
+
+      if (isAngular) {
+        const degrees = action.payload.value as number;
+        const quaternion = new THREE.Quaternion().setFromAxisAngle(axis, degrees * DEG2RAD);
+        const euler = new THREE.Euler().setFromQuaternion(quaternion);
```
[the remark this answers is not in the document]
jhurliman: I haven't been worrying about small allocations in handleSettingsUpdate() since these trigger off user input, although it's not a bad idea to be consistent. I'll do a pass on allocations.

2metres:
Can we write a user-script to make it dance for the announcement?

jtbandes (approved):

PR #4577, opened 2022-10-06
[New3D] Add option to ignore "up axis" for STL models

v-kiniv:
**User-Facing Changes**
Add `Ignore STL <up_axis>` option to New3D panel->Settings->Scene

**Description**
In my case, after this [commit](https://github.com/foxglove/studio/commit/09f46a498e9431f95f090839932ba74ee9a9e509#diff-1e01cf8f145a7673bda300ba7644ba3b3e911d63d2cc880e4afc85b8a3f81f67R163), my robot model was totally broken. So I  added this option, similar to the existing one  - `Ignore COLLADA <up_axis>`. 
P.S. My robot model displayed ok in RVIZ and previous versions of Foxglove(prior to mentioned commit).

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/CoreSettings.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/CoreSettings.ts
@@ -6,6 +6,7 @@ import { cloneDeep, round, set } from "lodash";
 
 import { SettingsTreeAction } from "@foxglove/studio";
 
+import { DEFAULT_STL_UP_AXIS } from "../ModelCache";
 import { FollowMode, Renderer, RendererConfig } from "../Renderer";
 import { SceneExtension } from "../SceneExtension";
 import { SettingsTreeEntry } from "../SettingsManager";
@@ -149,14 +150,18 @@ export class CoreSettings extends SceneExtension {
                   ? "This setting requires a restart to take effect"
                   : undefined,
             },
-            ignoreStlUpAxis: {
-              label: "Ignore STL <up_axis>",
+            stlUpAxis: {
+              label: "STL up axis",
               help: "Match the behavior of rviz by ignoring the <up_axis> tag in STL files",
-              input: "boolean",
-              value: config.scene.ignoreStlUpAxis,
+              input: "select",
+              value: config.scene.stlUpAxis ?? DEFAULT_STL_UP_AXIS,
+              options: [
+                { label: "Y axis", value: "y_up" },
+                { label: "Z axis", value: "z_up" },
+              ],
               error:
-                (config.scene.ignoreStlUpAxis ?? false) !==
-                this.renderer.modelCache.options.ignoreStlUpAxis
+                (config.scene.stlUpAxis ?? DEFAULT_STL_UP_AXIS) !==
+                this.renderer.modelCache.options.stlUpAxis
                   ? "This setting requires a restart to take effect"
                   : undefined,
             },
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/CoreSettings.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/CoreSettings.ts
@@ -152,7 +152,7 @@ export class CoreSettings extends SceneExtension {
             },
             meshUpAxis: {
               label: "Mesh up axis",
-              help: "The direction to use as “up” when loading meshes without orientation info(STL, OBJ)",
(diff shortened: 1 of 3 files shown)


jhurliman:
Thanks for the PR! Functionally this is a good improvement, since STL files (like OBJ files) are ambiguous as to what the up-axis is. We currently assume Y-up and apply a 90 degree rotation when loading the mesh in our Z-up scene. This aligns with the behavior of Blender and correctly renders many STL files "in the wild". However, Z-up STL files are perfectly valid as well.

The only thing I would change here is the user interface portion. There is no `<up_axis>` in STL files (it's not XML based, and there is no way of specifying up-axis in the file format), so what you're actually doing is telling Studio what the up-axis is in your STL files. I think it should be a dropdown labeled "STL up axis" with two choices, Y-up and Z-up (defaulting to Y-up).

You can make this change yourself, or I can get to it later this week.

v-kiniv:
I changed it to select, but I'm not sure about two things:
- I don't see where the default values for CoreSettings are kept, so I added `DEFAULT_STL_UP_AXIS` constant to ModelCache so there is single source of default value;
- as a type for `StlUpAxis` I would naturally choose enum, but throughout the project I see that unions are used, so I leaned towards union type.

packages/studio-base/src/panels/ThreeDeeRender/renderables/CoreSettings.ts
```diff
@@ -149,6 +150,21 @@ export class CoreSettings extends SceneExtension {
                   ? "This setting requires a restart to take effect"
                   : undefined,
             },
+            stlUpAxis: {
+              label: "STL up axis",
```
jtbandes, line 154: We discussed this and we think the setting should apply to both STL and OBJ files, since both of them are ambiguous (don't include embedded orientation info)

packages/studio-base/src/panels/ThreeDeeRender/renderables/CoreSettings.ts
```diff
@@ -149,6 +150,21 @@ export class CoreSettings extends SceneExtension {
...
+              help: "Match the behavior of rviz by ignoring the <up_axis> tag in STL files",
```
jtbandes, line 155:
```suggestion
              help: "The direction to use as “up” when loading meshes without orientation info",
```

packages/studio-base/src/panels/ThreeDeeRender/renderables/CoreSettings.ts
v-kiniv: I changed the code to use the option for OBJ as well and updated label to `Mesh up axis`, and added `(STL, OBJ)` to your proposed description

packages/studio-base/src/panels/ThreeDeeRender/renderables/CoreSettings.ts
```diff
@@ -150,18 +150,18 @@ export class CoreSettings extends SceneExtension {
                   ? "This setting requires a restart to take effect"
                   : undefined,
             },
-            stlUpAxis: {
-              label: "STL up axis",
-              help: "The direction to use as “up” when loading meshes without orientation info",
+            meshUpAxis: {
+              label: "Mesh up axis",
+              help: "The direction to use as “up” when loading meshes without orientation info(STL, OBJ)",
```
jtbandes, line 155:
```suggestion
              help: "The direction to use as “up” when loading meshes without orientation info (STL and OBJ)",
```

packages/studio-base/src/panels/ThreeDeeRender/renderables/CoreSettings.ts
```diff
@@ -6,6 +6,7 @@ import { cloneDeep, round, set } from "lodash";
 
 import { SettingsTreeAction } from "@foxglove/studio";
 
+import { DEFAULT_MESH_UP_AXIS } from "../ModelCache";
```
jtbandes, line 9: I believe there are some other DEFAULT_* things in CoreSettings.ts. You can put it with those.

v-kiniv: It's also used in `Renderer.ts`, so it looks more logical to include it there from `ModelCache.ts` then `CoreSettings`, there's is also similar approach for camera:

```import { DEFAULT_CAMERA_STATE } from "../camera";```

If you think it should be changed, just add commit to the branch.

jhurliman (approved):

Issue #4241, opened 2022-08-23
Improve empty layout state


foxymiles:
**Description**
Our empty layout state could be improved.

- Version: studio.foxglove.dev
- OS: any

**Steps To Reproduce**
There are various ways to get into this empty layout state but it's a bit frustrating for the user because you have to go open the layout sidebar and select a layout or create a new one. Instead we could either add a "Create a New Layout" button here or just load the default layout.

<img width="552" alt="Screen Shot 2022-08-23 at 4 30 36 PM" src="https://user-images.githubusercontent.com/93935560/186260552-22d63b12-e979-41d9-b87e-4b1267a757bc.png">

snosenzo:
related to this issue?: https://github.com/foxglove/studio/issues/4100

defunctzombie:
I think it's reasonable to default to a layout with a single 3d panel, raw message panel, and logs panel. Between these three panels there is _some_ action you can take to see some data.

osaro-imohe:
I can't see that on studio desktop. Has this been resolved?

jtbandes:
No, it has not been resolved. One way to get to this empty state is to visit a url with `&layoutId=some-invalid-id`.

osaro-imohe:
I took a look at the code and have some thoughts. It seems there are three ways to see the empty layout state:

1. Not putting in a layout Id
2. Inputting an invalid layoutId
3. Not being able to retrieve the layout for a valid layoutId

The first situation was fixed by creating a new default layout when no layoutId was provided, then setting the new default  layoutId as the selectedLayoutId.

```
 const layout = currentLayoutId ? await layoutManager.getLayout(currentLayoutId) : undefined;
 if (layout) {
      await setSelectedLayoutId(currentLayoutId, { saveToProfile: false });
 } else {
      const newLayout = await layoutManager.saveNewLayout({
        name: "Default",
        data: defaultLayout,
        permission: "CREATOR_WRITE",
   });
   await setSelectedLayoutId(newLayout.id);
```

I considered applying the same solution to the other two instances that lead to the empty layout state. The issue is there's no way to know which layoutId is the default layoutId after you've already created the first one. For instance, if studio is loaded twice with no layoutId, it'd create two new default layoutIds. 

I think only allowing one default layout might be the way to go here. Once a default layoutId is created we could store that in localStorage and use that whenever a user runs into one of the three instances above.

So it'd be something like:
* load studio with no layoutId or invalid layoutId
[...]

PR #4578, opened 2022-10-06
Feat #4241 Improve empty layout state

osaro-imohe:
**User-Facing Changes**

**Description**
Add's `Create new layout` button as referenced in issue foxglove/community#489 

Screen Shot 2022-10-06 at 2 42 46 PM

### DIFF
--- a/packages/studio-base/src/components/PanelLayout.tsx
+++ b/packages/studio-base/src/components/PanelLayout.tsx
@@ -40,8 +40,8 @@ import {
   LayoutState,
   useCurrentLayoutSelector,
   usePanelMosaicId,
+  useCurrentLayoutActions,
 } from "@foxglove/studio-base/context/CurrentLayoutContext";
-import { useCurrentLayoutActions } from "@foxglove/studio-base/context/CurrentLayoutContext";
 import { PanelsState } from "@foxglove/studio-base/context/CurrentLayoutContext/actions";
 import "react-mosaic-component/react-mosaic-component.css";
 import { useExtensionCatalog } from "@foxglove/studio-base/context/ExtensionCatalogContext";
--- a/packages/studio-base/src/components/PanelLayout.tsx
+++ b/packages/studio-base/src/components/PanelLayout.tsx
@@ -276,7 +276,7 @@ export default function PanelLayout(): JSX.Element {
   return (
     <EmptyState>
       <Typography display="block" variant="body1" color="text.primary">
-        You currently don&apos;t have a selected layout.
+        You don&apos;t currently have a layout selected.
       </Typography>
       <Link onClick={selectExistingLayout} underline="hover" color="primary" variant="body1">
         Select an existing layout
--- a/packages/studio-base/src/components/PanelLayout.tsx
+++ b/packages/studio-base/src/components/PanelLayout.tsx
@@ -12,7 +12,6 @@
 //   You may not use this file except in compliance with the License.
 
 import { Button, CircularProgress, styled as muiStyled } from "@mui/material";
-import moment from "moment";
 import React, {
   useCallback,
   useMemo,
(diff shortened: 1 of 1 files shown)


amacneil:
Thanks for the PR!

This is an improvement over what we have currently. I think it would be better still if:

- If no layout is selected, automatically select the first personal layout
- If there are no personal layouts, automatically create the "default layout" (functionality already exists)

@foxymiles thoughts?

osaro-imohe:
> Thanks for the PR!
> 
> This is an improvement over what we have currently. I think it would be better still if:
> 
> * If no layout is selected, automatically select the first personal layout
> * If there are no personal layouts, automatically create the "default layout" (functionality already exists)
> 
> @foxymiles @banisadr thoughts?

That sounds better. I've made the changes.

foxymiles:
@amacneil Selecting the first personal layout is a lot harder because those can be in a variety of different syncing states.

osaro-imohe:
@amacneil I think it'd be best  to revert to this commit (https://github.com/foxglove/studio/pull/4578/commits/064d24e0d63f8e7eb2880ce7029e10dc2382e709) in response to @foxymiles comment on syncing states

amacneil:
What's hard about selecting a layout that is in a syncing state?

osaro-imohe:
From foxymiles explanation, my understanding is that `getLayouts` returns all layouts, but Studio is constantly fetching layouts from the backend whenever they're available, there could be an instance where a user has a layout that isn't yet in `getLayouts` or where the last layout in `getLayouts` isn't actually the most recent layout

amacneil:
That doesn't seem like a major problem - worst case we select a different layout, but that is still better than showing them no layout.

osaro-imohe:
Yes, that's a valid point, but shouldn't the `Create a new layout` button always lead to the layout panel selector? That's what the button says. It seems inconsistent for the button to create a new layout sometimes and not create a new layout other times.

amacneil:
I don't think we need the button at all. If no layout is selected, just select one for me. If there are no available layouts, re-create the default layout.

Apparently "just select one for me" is difficult but I'm trying to understand why.

foxymiles:
> Apparently "just select one for me" is difficult but I'm trying to understand why.

It's not difficult to just select one of whatever we may have managed to sync at that point in time but that doesn't strike me as great UX. I can't think of any other apps that will just show me one of my existing documents picked at random.

defunctzombie:
I'd advocate for this screen to be more informative. Looking at the original issue here are the things we should be doing on this screen:

- Telling you that we couldn't select some layout with some layout id (maybe asking the user to check that they have permissions to access the layout)
- Showing you the list of layouts available to select
- Showing you a button to create a new empty layout

@amacneil I don't think making a default layout automatically is good UX for the flow here - we want to inform the user their desired layout is not one can load and then have them take an appropriate action as a next step.

amacneil:
Let's talk about what would cause you to get into this state. In regular use, studio doesn't allow you to have no layout selected.

- If you open the app with no layouts, we create a default layout for you
- If you open the app with an invalid `layoutId` query param, you get this page (we should instead show you an error, like you suggested
- If you delete the last available layout, we show this page. Makes sense, although I think it would be equally sensible to just re-create the default layout, as we would do when you next open the app.
- If you sign out, we delete all your layouts, but keep the active layout. That's weird - we should either delete all layouts or not, but deleting them and keeping one active is strange.
- If you sign out then hit refresh, we create a new Default layout for you (because the existing `layoutId` no longer exists). Makes sense, because this is the same as hitting out app with an empty state.
- If you sign out and immediately back into a new org from console, you get `Your active layout was deleted`, then you see this page. That's silly, we should have cleared the layout state when you signed out. When I sign into a new org, you should show me a layout. Ideally show me the last layout I had selected when I was signed in to this org, but if we don't want to do this, then I would settle for just showing me the first personal layout you find in that org.

Anecdotally, the main reason I see this page is when switching orgs, and I would much prefer to have you select a layout for me rather than always seeing this page. I realize most users do not switch orgs often (or ever), which is why I haven't pushed on prioritizing this ticket, but here we are with a PR so I may as well share thoughts on it.

amacneil:
Discussed internally at product review. There are a lot of things we want to do to improve layout workflow, but your original idea here is a good improvement worth merging.

The only change we should make is provide both options (select existing layout + create layout). For example

```
You don't currently have a layout selected.
_Select an existing layout_ or _Create new layout_
```

(probably two buttons, but you could try two links instead)

cc @2metres for thoughts on the UI

2metres:
I think 2 links are my preference

Screen Shot 2022-10-13 at 12 33 00 pm

osaro-imohe:
@amacneil @foxymiles

PR #4579, opened 2022-10-06
Fixes copying settings between 3d panels

### LINKED ISSUE #4498: Unable to copy panel settings from one 3d panel to another 3d panel
See the below screencast:

https://user-images.githubusercontent.com/84792/192817866-0682a3c8-cbc5-4a23-98be-d49cc80d07f1.mov

Expected behavior:

When I pasted new settings into the 3D panel "import settings" dialog I expected to see the `Follow mode` setting update to `position` (to match the setting I copy/pasted). However the setting did not update after I clicked "apply".

When I refresh the page (after applying the setting) it does look like the new setting is applied.

foxymiles:
**User-Facing Changes**
Fixes copying settings between 3d panels.

**Description**
Some panels, like extension panels and the new 3d panel, only read their config from the layout once when the panel is mounted and maintain their own working copy of the config after that. This means that a new config pasted in from another panel is effectively ignored. We have an existing mechanism for handling this in the case where the panel config is reset but this approach won't work for copying settings since we don't have a simple way to distinguish a copied config from a merely updated config the way we do for a panel reset which results in an empty `{}` config.

The fix here is to add a new sequence number concept to panel settings and increment this when a change is made to a panel's config that should result in a remount. Arguably this should go into the CurrentLayout context but this is already a very complex context with a bespoke notification system and persistence features. So for the sake of simplicity I've opted to put this in the PanelSettingsEditorContext for now. I think it would make sense at some point to refactor the CurrentLayout context to use Zustand and to layer the persistence features on top of the store in a separate sync component, at which point these sequence numbers could be moved to this new store.

Fixes #4498


packages/studio-base/src/components/PanelExtensionAdapter/PanelExtensionAdapter.tsx
```diff
@@ -89,7 +88,7 @@ function PanelExtensionAdapter(props: PanelExtensionAdapterProps): JSX.Element {
 
   const { openSiblingPanel } = usePanelContext();
 
-  const [panelId, setPanelId] = useState(() => uuid());
+  const [panelId] = useState(() => uuid());
   const latestPanelId = useRef<string | undefined>(panelId);
```
jtbandes, line 92: latestPanelId is probably not needed anymore, then?

packages/studio-base/src/providers/PanelSettingsEditorContextProvider.tsx
```diff
@@ -16,7 +16,20 @@ import {
 function createSettingsEditorStore(): StoreApi<PanelSettingsEditorStore> {
   return createStore((set) => {
     return {
+      sequenceNumbers: {},
       settingsTrees: {},
+
+      incrementSequenceNumber: (panelId: string) => {
```
jtbandes, line 22: We should provide a way to delete a sequence number instead, otherwise the object grows larger indefinitely.

foxymiles: These are transient and not stored in the layout. It's hard to imagine these could ever contribute significantly to memory use since there's one ID per panel.

packages/studio-base/src/components/PanelExtensionAdapter/PanelExtensionAdapter.tsx
foxymiles: Yeah right we can delete that.

foxymiles: I guess we still need some kind of check to retain the fix from https://github.com/foxglove/studio/pull/4436.

Can we use `isMounted` for that?

packages/studio-base/src/providers/PanelSettingsEditorContextProvider.tsx
foxymiles: If we really think this is an issue I suppose we could add a function to clear it when a new layout is selected.

Of course it's also possible to move this into the CurrentLayout context in which case it would automatically be cleared on a new layout but that's significantly more complex code to change.

packages/studio-base/src/components/PanelExtensionAdapter/PanelExtensionAdapter.tsx
defunctzombie: IIRC we saw behavior that made it look like the panel was actually still mounted but maybe an `isMounted` would be sufficient. I don't recall if we tried that when working on the fix in #4436

jtbandes: Didn’t use that approach because of: https://github.com/foxglove/studio/pull/4436#issuecomment-1249723152

however, that is no longer true with this PR as I understand it, so maybe isMounted is sufficient now. It would need some testing.

foxymiles:
After adding the checks for mounted state I tried the repro steps from https://github.com/foxglove/studio/issues/4434 and was not able to observe the bug again so I think it's working.

One wrinkle - I had to write a local definition of the `useMountedState` hook that depends on `useLayoutEffect` instead of `useEffect` because we do most of our work in `useLayoutEffect` in the extension adapter and that runs before `useEffect` so the `isMounted` state returned false even though the panel was mounted.

packages/studio-base/src/components/PanelExtensionAdapter/PanelExtensionAdapter.tsx
```diff
@@ -48,6 +55,26 @@ import { initRenderStateBuilder } from "./renderState";
 
 const log = Logger.getLogger(__filename);
 
+/**
+ * We have to use useLayoutEffect to track our mounted state because we use layout effects
+ * to initialize the panel and render. If we use useEffect this runs later in the react
+ * cycle and isMounted returns false for layout effects.
+ */
+function useSynchronousMountedState(): () => boolean {
```
defunctzombie, line 63: This should go into the hooks package: https://github.com/foxglove/studio/tree/main/packages/hooks and get a test

foxymiles:
> Do we have tests for any of the changed behavior? PanelExtensionAdapter has a large downstream impact if we mess things up so I'd like to push a bit harder on making sure we have test coverage for it.

I'm not quite sure how to write more PanelExtensionAdapter tests for this change, since it's mainly moving functionality out. We do have pretty good tests for the expected behaviors of this component like mounting & unmounting & subscribing/unsubscribing and those pass after the change.

jtbandes:
There was a unit test added in #4436 so that should be decent coverage I think.

packages/studio-base/src/hooks/useSynchronousMountedState.tsx
```diff
@@ -0,0 +1,28 @@
...
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { useCallback, useLayoutEffect, useRef } from "react";
+
+/**
+ * Registers and unregisters mounted state during useLayoutEffect instead of useEffect.
+ *
+ * If you're using useLayoutEffect in your component and want to track your component's
+ * mounted state you need to use this instead of useMountedState.
```
jtbandes, line 11: Makes me wonder if we should just recommend always using this version, and ban use of useMountedState. I can't think of a reason you'd want to prefer the useEffect version over this version.

jtbandes (approved):

packages/studio-base/src/hooks/useSynchronousMountedState.tsx
foxymiles: The only reason I can conceive of is a possible performance hit from adding a `useLayoutEffect` but it's hard to imagine it would be noticeable even if so.

foxymiles:
@defunctzombie Do you still have reservations about this? I'd like to get two approvals on this since, as you correctly point out, we need to be careful changing PanelExtensionAdapter.

packages/studio-base/src/components/PanelLayout.tsx
```diff
@@ -94,6 +100,28 @@ function TabMosaicWrapper({ tabId, children }: PropsWithChildren<{ tabId?: strin
   );
 }
 
+/**
+ * Wrapper component used to force-remount the panel when key properties like the tabId
+ * or settings sequence number change.
+ */
+function PanelRemounter({
```
defunctzombie, line 107: I prefer seeing stuff like this in its own file even if its a small component. Helps promote the idea of component is a file and maintained as a file. Not a hard ask tho since its largely my preference.

packages/studio-base/src/context/PanelSettingsEditorContext.ts
```diff
@@ -12,7 +12,24 @@ import useGuaranteedContext from "@foxglove/studio-base/hooks/useGuaranteedConte
 export type ImmutableSettingsTree = DeepReadonly<SettingsTree>;
 
 export type PanelSettingsEditorStore = {
+  /**
+   *  Used for forcing remounts on panels when fundamental properties change.
```
defunctzombie, line 16: Describe what a "fundamental" property is and how we determine valid fundamental properties vs other properties.

packages/studio-base/src/hooks/useSynchronousMountedState.test.tsx
```diff
@@ -0,0 +1,20 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { renderHook } from "@testing-library/react-hooks";
+
+import { useSynchronousMountedState } from "@foxglove/studio-base/hooks/useSynchronousMountedState";
+
+describe("useSynchronousMountedState", () => {
+  it("tracks mounted state", () => {
```
defunctzombie, line 10: Do we need a test showing the value remains true through renders but not remounts?

packages/studio-base/src/hooks/useSynchronousMountedState.tsx
defunctzombie: 👍 to banning `useMountedState`

Maybe we should require all hooks from `react-use` to go through our hooks package so it is clear which hooks are ok to use given that package is not maintained anymore.

foxymiles:
@defunctzombie Do you have any other ideas for forcing panels to recognize changes in their configs that should supersede whatever local state they're maintaining? I couldn't think of any since currently panels can do whatever they want for local state management and don't report back through the API that they're ignoring config changes so a panel reboot seemed like the only way.

packages/studio-base/src/hooks/useSynchronousMountedState.test.tsx
foxymiles: Good idea. I've added this.

packages/studio-base/src/context/PanelSettingsEditorContext.ts
foxymiles: Changed this wording to specifically refer to force-reloading configs since that's the only case where it's used now.

foxymiles:
@defunctzombie I don't see a way around this kind of forced-reload as long as panels are free to ignore changes to their configs.

If we're willing to risk a more disruptive but perhaps slightly less hacky change I would suggest we:
1. Move this concept of panel settings sequence numbers/stamps into the CurrentLayoutProvider.
2. Convert the CurrentLayoutProvider to zustand instead of a custom listener system.
3. Extract the persistence features of the CurrentLayoutProvider into a separate component we layer on top of the provider.

The sequence number could just as easily be a uuid or any other sequentially unique value. We just need some mechanism to generate a new react key and force the panel to reload.

PR #4581, opened 2022-10-06
RenderableTriangles does not handle colors.length < points.length

### LINKED ISSUE #4559: RenderableTriangles does not handle colors.length < points.length
This bug may be present in other scene entities as well. This should result in a topic error and fall back to `triangles.color` and then some default color (white?).

https://github.com/foxglove/studio/blob/main/packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableTriangles.ts#L87

snosenzo:
**User-Facing Changes**
 - n/a

**Description**
 - renderableTriangles now shows error when the colors array is shorter than the points array and displays a green debug color in its place
 - an error message will also be attached to the related topic
 - topic error tooltips are now scrollable and have a max size

Screen Shot 2022-10-06 at 4 30 27 PM

Fixes #4559

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableTriangles.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableTriangles.ts
@@ -91,7 +91,7 @@ export class RenderableTriangles extends RenderablePrimitive {
         vertices.setXYZ(i, point.x, point.y, point.z);
 
         if (!singleColor && colors && primitive.colors.length > 0) {
-          const color = primitive.colors[i] ? primitive.colors[i]! : missingColor;
+          const color = primitive.colors[i] ?? missingColor;
           // only trigger on last point index
           if (i === primitive.points.length - 1 && color === missingColor) {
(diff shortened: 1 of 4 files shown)


snosenzo:
example data: https://drive.google.com/file/d/1Uyzv24HUHiKwhioKEEa549A4nB_2I5El/view?usp=sharing

packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableTriangles.ts
```diff
@@ -88,7 +91,15 @@ export class RenderableTriangles extends RenderablePrimitive {
         vertices.setXYZ(i, point.x, point.y, point.z);
 
         if (!singleColor && colors && primitive.colors.length > 0) {
-          const color = primitive.colors[i]!;
+          const color = primitive.colors[i] ? primitive.colors[i]! : missingColor;
```
jtbandes, line 94:
```suggestion
          const color = primitive.colors[i] ?? missingColor;
```

jtbandes (approved):

PR #4582, opened 2022-10-06
rename Topic.datatype to schemaName

jtbandes:
**User-Facing Changes**
[Extension API] `Topic#datatype` has been renamed to `schemaName`. `datatype` will still be available for backwards compatibility, but is now deprecated.

**Description**
Within Studio code, we use a Topic type that has only name and schemaName. The deprecated `datatype` key is added by the ExtensionAdapter for extension panels.


packages/studio/index.d.ts
```diff
@@ -281,11 +285,11 @@ declare module "@foxglove/studio" {
     subscribeAppSettings(settings: string[]): void;
 
     /**
-     * Indicate intent to advertise on a specific topic and datatype.
+     * Indicate intent to advertise on a specific topic.
```
defunctzombie, line 288: Do we want to mention and `schemaName` ? The `schemaName` part is important for advertisement.

jtbandes: This felt awkward because you’re not really advertising “on a datatype/schemaName”… any suggestions for better wording?

defunctzombie: I think what you've done in your latest change is 👍 

I was going to suggest adding `topic with messages conforming to schemaName` - and you have basically that.

defunctzombie (approved):
👏

Issue #2999, opened 2022-03-09
Switch "explore sample data" to use the nuscenes mcap file instead of the bag file.


amacneil:
**Description**
Our sample bag is currently not particularly user-friendly with some topics such as `/imu` which have infrequent receive times. This leads to unexpected rendering in the Raw Message and Plot panels which default to showing messages by receive time.

We should update the sample bag to pin receive time to header stamp, so that it leads to a more predictable experience for new users.

defunctzombie:
@amacneil do you still want this? I see it's gone through some assignment changes but we've never actually thrown it into a project or planning session.

foxymiles:
I've made some changes to the https://github.com/foxglove/python-mcap-ros1-support library to make writing a script to do this easier but @jhurliman said there are some other obstacles in the data to making this happen.

jhurliman:
Yeah, this is still in progress. The script has to handle MarkerArray and ImageMarkerArray, and I haven't had time this week to touch this.

defunctzombie:
Sounds good. I've thrown it on the Q2+ project board so we make some time to push this over the finish line within the quarter.

amacneil:
I still think this is important, our sample dataset currently gives a bad plotting experience out of the box.

foxymiles:
Note from @jhurliman:

```
if you have time today, could you modify it to also check messages 
for a .markers field that is a list, and if the list is non-empty 
use .header.stamp from the first list entry? that will cover 
MarkerArray and ImageMarkerArray which are the last two holdouts
```

foxymiles:
Sent convert script to @jhurliman

defunctzombie:
What's the status of this? Its been assigned for a while with no activity.

foxymiles:
Here's the current script. When I convert nuscenes and play the converted file it looks a bit strange to me like parts of the scene are out of sync with each other. @jhurliman Does this look like it should be doing what we expect?

```
from mcap_ros1.writer import Writer as Ros1Writer
from std_msgs.msg import String  # type: ignore
import rosbag  # type: ignore
import sys

bag = sys.argv[1]
mcap = open("converted.mcap", "w+b")
writer = Ros1Writer(mcap)
for topic, message, msg_time in rosbag.Bag(bag).read_messages():
    if topic == "/tf" and message.transforms:
        message_time = message.transforms[0].header.stamp.to_nsec()
    elif hasattr(message, "markers") and message.markers[0].header.stamp:
[...]

PR #4584, opened 2022-10-06
Update sample dataset to use nuscenes2mcap

jtbandes:
**User-Facing Changes**
Updated the dataset used for "Explore sample data". The file size is reduced, and the dataset demonstrates more features of Studio, such as foxglove.SceneEntity and foxglove.Grid types, and support for MCAP files with mixed encodings (ROS, Protobuf, and JSON). The time-ordering of messages in the sample data has also been improved.

**Description**
Closes https://github.com/foxglove/studio/issues/2999


amacneil:
Something is horribly wrong...

https://user-images.githubusercontent.com/637671/194462049-dbc490e5-c542-4435-9fb9-db72cd0bb470.mp4

jtbandes:
Does it still do that after preloading has finished?

amacneil:
Yes*

*but we only preload half the file, so I have to lag-play through it before I can get sane performance.

https://user-images.githubusercontent.com/637671/194463202-71f9bd08-abe7-440c-a890-2f8e469b64db.mp4

snosenzo:
the drivable area isn't on by default but it should have these defaults instead of colormap: turbo
image

jtbandes:
I've removed the re-publishing of the map, so that should help reduce stuttering once preloaded, but it looks like preloading is bottlenecked on ImageAnnotation deserialization. #4538 should help.

amacneil:
Suggestions
- Hide `map` transform
- Show `/CAM_FRONT/image_rect_compressed` and `/CAM_FRONT/camera_info` at 2m distance

amacneil:
Blocked on making sure we are not going to revert https://github.com/foxglove/studio/pull/4538

cc @defunctzombie

defunctzombie (approved):

PR #4591, opened 2022-10-07
Fix race condition in Marker mesh_resource loading

jtbandes:
**User-Facing Changes**
[3D] Fixed an issue where ROS mesh_resource markers would sometimes overwrite each other due to a race condition.

**Description**
Also added dispose() calls to RenderableMeshResource

Test data:
[models.mcap.zip](https://github.com/foxglove/studio/files/9736873/models.mcap.zip)
[layout.json.zip](https://github.com/foxglove/studio/files/9736875/layout.json.zip)

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/dispose.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/dispose.ts
@@ -1,0 +1,39 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import * as THREE from "three";
+
+export function disposeMaterial(material: THREE.Material): void {
+  if (material instanceof THREE.MeshStandardMaterial) {
+    material.map?.dispose();
+    material.lightMap?.dispose();
+    material.aoMap?.dispose();
+    material.emissiveMap?.dispose();
+    material.bumpMap?.dispose();
+    material.normalMap?.dispose();
+    material.displacementMap?.dispose();
+    material.roughnessMap?.dispose();
+    material.metalnessMap?.dispose();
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/panels/ThreeDeeRender/dispose.ts
```diff
@@ -0,0 +1,39 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import * as THREE from "three";
+
+export function disposeMaterial(material: THREE.Material): void {
+  if (material instanceof THREE.MeshStandardMaterial) {
+    material.map?.dispose();
```
snosenzo, line 9: I think you might be able to do:
```
for (const value of Object.values(material)) {
        if (value instanceof THREE.Texture) {
          value.dispose();
        }
      }
```

snosenzo: should you check the uniforms on the material too? 
https://r105.threejsfundamentals.org/threejs/lessons/threejs-cleanup.html

snosenzo (approved):

packages/studio-base/src/panels/ThreeDeeRender/dispose.ts
jtbandes: Will let @jhurliman weigh in since he wrote the original version of this material dispose block iirc.

jtbandes: Going to merge for now since this fixes an annoying bug, and can revisit in a separate PR if desired.

snosenzo: sounds good

PR #4593, opened 2022-10-07
more datatype→schemaName renaming

jtbandes:
**User-Facing Changes**
None

**Description**
Rename datatype in AdvertiseOptions and some local variables


benchmark/src/players/SinewavePlayer.ts
```diff
@@ -95,7 +95,7 @@ class SinewavePlayer implements Player {
 
     for (let i = 0; i < sinewaveCount; ++i) {
       const topicName = `sinewave_${i}`;
-      topics.push({ name: topicName, datatype: "Sinewave" });
```
defunctzombie, line 98: How did these get past the type change? Do we have something downstream that ensures we continue setting `datatype` even when a player only sets `schemaName` ?

defunctzombie (approved):
As long as we have something downstream that sets `datatype` for extension panels 👍

benchmark/src/players/SinewavePlayer.ts
jtbandes: I don't think the benchmark players are being typechecked in CI.

jtbandes:
https://github.com/foxglove/studio/blob/9e768524cb65b393087ffbeb5f88444b9eb06e8a/packages/studio-base/src/components/PanelExtensionAdapter/renderState.ts#L118-L122

jtbandes:
As discussed in the https://github.com/foxglove/studio/pull/4582 description.

> Within Studio code, we use a Topic type that has only name and schemaName. The deprecated datatype key is added by the ExtensionAdapter for extension panels.

PR #4600, opened 2022-10-09
Improve yarn-wrapper warning

defunctzombie:
**User-Facing Changes**
Better instructions for developers running into yarn install issues when getting started on contributing to studio.

**Description**
Even after running `corepack enable` some developers report continuing to see the error
message for enabling corepack. We've tracked this down to instances where the user installed
yarn at the system level in a way that would take precedence over the version corepack installs.

This change adds language to the error message guiding users to uninstall their global version
so the corepack version runs instead.

### DIFF
--- a/.yarn/yarn-wrapper.js
+++ b/.yarn/yarn-wrapper.js
@@ -10,8 +10,8 @@ if (process.env.COREPACK_ROOT == undefined) {
   console.error("");
   console.error("If you have run `corepack enable` and still see this error, you have likely");
   console.error("installed yarn globally or using your system package manager. Your installed");
-  console.error("version of yarn is superceding corepack's version. Delete or uninstall your");
-  console.error("version of yarn to use the corepack version");
+  console.error("version of yarn is superseding corepack's version. Delete or uninstall your");
+  console.error("version of yarn to use the corepack version.");
   process.exit(1);
 }
 
--- a/.yarn/yarn-wrapper.js
+++ b/.yarn/yarn-wrapper.js
@@ -7,6 +7,11 @@ const path = require("path");
 if (process.env.COREPACK_ROOT == undefined) {
   console.error("This repository uses corepack. Enable corepack by running `corepack enable`");
   console.error("Learn more at: https://nodejs.org/api/corepack.html");
+  console.error("");
+  console.error("If you have run `corepack enable` and still see this error, you have likely");
+  console.error("installed yarn globally or using your system package manager. Your installed");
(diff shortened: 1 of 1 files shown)


defunctzombie:
bump @jtbandes @jhurliman

.yarn/yarn-wrapper.js
```diff
@@ -7,6 +7,11 @@ const path = require("path");
 if (process.env.COREPACK_ROOT == undefined) {
   console.error("This repository uses corepack. Enable corepack by running `corepack enable`");
   console.error("Learn more at: https://nodejs.org/api/corepack.html");
+  console.error("");
+  console.error("If you have run `corepack enable` and still see this error, you have likely");
+  console.error("installed yarn globally or using your system package manager. Your installed");
+  console.error("version of yarn is superceding corepack's version. Delete or uninstall your");
+  console.error("version of yarn to use the corepack version");
```
jtbandes, line 14:
```suggestion
  console.error("version of yarn is superseding corepack's version. Delete or uninstall your");
  console.error("version of yarn to use the corepack version.");
```
```suggestion
  console.error("version of yarn is superseding corepack's version. Delete or uninstall your");
```
needs more cspell

jtbandes: Why not just print out 1 long line?

jtbandes (approved):

.yarn/yarn-wrapper.js
defunctzombie: Doesn't look as nice /shrug I guess I should let the terminal wrap it? I don't feel strongly. If we think that's better I can change it.

.yarn/yarn-wrapper.js
```diff
@@ -7,6 +7,11 @@ const path = require("path");
 if (process.env.COREPACK_ROOT == undefined) {
   console.error("This repository uses corepack. Enable corepack by running `corepack enable`");
   console.error("Learn more at: https://nodejs.org/api/corepack.html");
+  console.error("");
+  console.error("If you have run `corepack enable` and still see this error, you have likely");
+  console.error("installed yarn globally or using your system package manager. Your installed");
+  console.error("version of yarn is superseding corepack's version. Delete or uninstall your");
+  console.error("version of yarn to use the corepack version.");
```
jhurliman, line 14: This printout will look very strange if your terminal width is less than 76 characters.

defunctzombie: I've removed the manual wrapping. We will let nature takes its course.

PR #4609, opened 2022-10-11
Move overrides out of a single file and into individual files

2metres:
**User-Facing Changes**
None

**Description**
Move overrides out of a single file and into individual files as it was becoming overgrown.


2metres:
Not sure if this is better. Open to suggestions as to how we could reduce the boilerplate

defunctzombie:
Personally I like the single file for this. It's not so long and serves the single purpose of overriding specific keys of the theme object. I can quickly see how we do overrides for other mui objects. I do understand how this could become a more involved set of overrides or a more bespoke theme over time but it doesn't feel like its gotten there yet.

foxymiles:
I guess I don't have a strong opinion on this either way.

One thing it might be worth thinking about while making changes to this stuff is how best to keep Studio & Console looking consistent.

PR #4610, opened 2022-10-11
Replace Promise + setTimeout wait with delay helper

osaro-imohe:
**User-Facing Changes**
NIL

**Description**
Replace Promise + setTimeout wait with delay helper

### DIFF
--- a/desktop/main/simulateUserClick.ts
+++ b/desktop/main/simulateUserClick.ts
@@ -4,6 +4,8 @@
 
 import { BrowserWindow } from "electron";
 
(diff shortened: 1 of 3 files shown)


defunctzombie (approved):

defunctzombie:
The build is failing because the import is incorrect. You don't need `/src/` in the import from `@foxglove/studio-base`.

Otherwise this PR looks 👍  to me.

osaro-imohe:
Nice catch! I've updated it.

PR #4617, opened 2022-10-12
Add `playbackSpeed` to `PanelExtensionAdapter.RenderState`

codewithpassion:
**User-Facing Changes**

Adding `playbackSpeed` to `@foxglove/studio/RenderState` so that extensions can react to changes to the playback speed.

**Description**

For certain extensions, it might be required to know the current playback speed. For example my [experimental H64 extension](https://github.com/codewithpassion/foxglove-studio-h264-extension). 

This PR is adding a `playbackSpeed` to the extension context render state.

### DIFF
--- a/packages/studio-base/src/components/PanelExtensionAdapter/renderState.ts
+++ b/packages/studio-base/src/components/PanelExtensionAdapter/renderState.ts
@@ -193,6 +193,14 @@ function initRenderStateBuilder(): BuildRenderStateFn {
       }
     }
 
+    if (watchedFields.has("playbackSpeed")) {
+      const playbackSpeed = activeData?.speed;
+      if (playbackSpeed !== renderState.playbackSpeed) {
+        renderState.playbackSpeed = playbackSpeed;
+        shouldRender = true;
+      }
+    }
+
     if (!shouldRender) {
       return undefined;
     }
--- a/packages/studio/index.d.ts
+++ b/packages/studio/index.d.ts
@@ -193,6 +193,9 @@ declare module "@foxglove/studio" {
 
     /** Application settings. This will only contain subscribed application setting key/values */
     appSettings?: ReadonlyMap<string, AppSettingValue>;
+
+    /** The speed of playback currently set on the player. */
+    playbackSpeed?: number;
   }
 
   export type PanelExtensionContext = {


jhurliman (approved):

defunctzombie:
@codewithpassion to confirm again - this is required for your extension to work or you were able to find a way to feed messages that does not require this? My understanding from my playing with h264 things is that the libraries can be made to produce a frame given some NAL inputs and don't depend on knowing some external playback rate.

Because this is a user-facing API that we will need to avoid breaking here are some high level thoughts we should answer before merging:

- Is the playback speed sufficient or do you also need to know the target framerate (i.e. 30fps, 60fps)?
- Should any delays in playback affect the playback speed reporting?
- What are the units for the playbackSpeed value? Should it be Hz? ms? the multiplier against realtime? Right now this exposes a field from the playerState that's always been an internal property with little thought given to changing it breaking something.

@jhurliman you had indicated this might be required for some other kinds of extensions. Could you elaborate on what those might be? What libraries we might use and the inputs those would expect?

codewithpassion:
Hi @defunctzombie. Thanks for considering this change. 


I did do extensive testing with jMuxer and I can make playback up to `1x` speed work without knowing the playback speed.
The frame rate can be encoded in the H264 units and be read from there. 

If the playback speed is `< 1x` then it's ok, as the the messages should now come in at the right speed and will be rendered.
At very low speeds it might be possible to see the individual frames, but I deem this accpetable.

The problem I have is when the playback speed is > `1x`. 
What is happening then is, we get more frames and push them into the decoding pipeline (jmuxer to the media extensions). And they are played back at the FPS encoded in the NALUs. So, assuming my video source is `30fps, and the playback speed sits at '2x', the video will still play at '30fps' rather then '60fps`. If I then press pause, the video will continue to play until the buffer runs dry. 

To you questions:
- For my purpose, it looks like the `playbackSpeed`, as a number value of `0.01` to `5` (or more) is sufficient. I can calculate the actual framerate from that value.
- not sure if you can even supply the target framerate in Hz. If `Studio` knows that rate, that could be passed instead. But we probably would need to know the base (aka `1x`) framerate too, otherwise it could cause issues if the source video frame rate is `60fps` but the data in the MCAP/ROS file is at a different frequency.
- Can you elaborate on `delays in playback`? How would those delays manifest?
- Regarding units: I'm quite happy with the value being a multiplier against realtime. Hz or ms again would be depended on the  message frequency and the FPS of the video source.

Hope that helps
Dom

jhurliman:
`playbackRate` is a more precise term for this, and aligns with existing web APIs (see https://developer.mozilla.org/en-US/docs/Web/Guide/Audio_and_video_delivery/WebAudio_playbackRate_explained). It would require changing in a handful of places in the codebase, but I think it's a worthwhile change to make if we're exposing this in the public API.

Currently, extension panels have a callback that is triggered as new data comes in at some nominal maximum frequency (configurable in Studio) and the wallclock time between callbacks can be measured with web APIs. This is sufficient for taking chunks of streaming data with a known source rate and playing them back at a real-time rate. However, Studio has a user-configurable toggle for playback rate (as a unit ratio of real-time, 1.0=real-time and 0.5=half speed). Without knowing what the target playback rate is, any playback process that relies on buffered playback won't know how fast to "render".

One example is the web audio API. The typical way to play back a topic containing audio data would be to construct a https://developer.mozilla.org/en-US/docs/Web/API/AudioBufferSourceNode from incoming audio sample data and communicate the sampling frequency of this data through a side channel. The `AudioBufferSourceNode#playbackRate` would be set to Studio's `playbackRate` and playback would be scheduled using the start timestamp of each chunk of samples.

Video is another example. While it would be possible in some circumstances to decode and present video frames as they're received by an extension with no buffering, this will behave poorly under load and may not work when the target `playbackRate` is faster than the rendering loop.

codewithpassion:
> > And they are played back at the FPS encoded in the NALUs
> 
> If the video plays back with the FPS encoded in the NALU how does knowing the playback speed help resolve this?

If we pass the duration or FPS, we can override the handling of the FPS. So jMuxer is using the passed fps rather the default one (30fps) or the one read from the stream.

Knowing the `playback speed`, we can calculate the proper FPS. So that a `30fps` clip would be played at `60fps` if the playback speed is set to `2x`. 

> In your extension you are are assuming 60fps for message playback? Or some fps for the video?
> 
> https://github.com/codewithpassion/foxglove-studio-h264-extension/blob/main/src/H264Panel.tsx#L76
> 

At the moment, I do default to `60fps` as this is the fps my video source is producing. 
On the [next line](https://github.com/codewithpassion/foxglove-studio-h264-extension/blob/main/src/H264Panel.tsx#L77) I then calculate the proper time for the `playback speed` if it's known.

Since writing this, I also realized that the media source API also supports playback speed - called `playbackRate`.
While I haven't experimented with that - it would lend itself to be used exactly for the use case of fast-forward. 
`playbackRate` in the media source API is a floating point number. `1` being the native playback speed of the source, `0.5` being half speed, `2` being double speed. Same as in the studio at the moment.

> Is there any aspect of jMuxer internals that is preventing this? I want to make sure we understand whether this is some limitation of jmuxer api decisions and is something that is changeable or if there's some fundamental reason that prevents rendering to the canvas at faster rates.

To be honest, I need to look into this more next week to answer this question.

> 
> For example - I have a viewer repo where I've been tinkering with h264 playback from mcap files via jmuxer and I was able to construct JMuxer with a `flushingTime: 1` and feed data with `duration: 2`. With these values I could feed in messages slowly or quickly and did not adjust duration separately for slow vs fast.

Have you tried feeding in at `2x` speed?
What happens if you stop the playback? In my experience, the playback of the media is actually still `1x` and when I hit pause, the playback keeps going until the buffer runs dry.


I agree with @jhurliman  that `playbackRate` would be a better name.

jhurliman:
https://github.com/foxglove/studio/pull/4645

codewithpassion:
Thanks @defunctzombie for looking into this. 

I've tried the [above approach](https://github.com/codewithpassion/foxglove-studio-h264-extension/blob/a20ac5251659d5ed7e1d126e6acd5d80e192c70a/src/H264Panel.tsx#L85) but the video turned out to be chunky and not play properly at 1x speed.

At >1x it doesn't show as much as the video plays quicker. 

Now, I've been digging through the code of `jMuxer` to look at how things are handled on that side. 
I found that:
- If the `fps` are not specified, it reverts to `30` as a default unless `readFpsFromTrack` is set. Then it reads the fps from the video data.
- If the duration is omitted in the `feed()` function call, it calculates a `duration` for every frame in the `feed()` payload. This calculation is based on the FPS from the configuration stage, or the FPS read from the video. This duration is then used to create a MP4 stream that then is passed to the `MediaSource API` for decoding.

So, if we call `feed({video, duration: 1});` what happens is, that the frame gets played at speed of `1ms` (or at least as fast as it can) which leads to jumpy video.

This leaves me with the problem that I don't see how I can push video frames into the pipeline at the right speed without knowing the `playbackSpeed/playbackRate`. 

Even if we would not use `jMuxer` by using the `MediaSource API` directly, and create the MP4 stream ourself, I'm not sure how to calculate a frame duration without knowing the playback speed.
Happy to jump on a call and discuss options and work through the problem if you have an idea. 

- Dom

codewithpassion:
I'll look into that.

codewithpassion:
So, To try and not be depending on `jMuxer`, I went and added my own implementation to set up a media source and then create fragmented MP4 from the h264 NALU.

You can't just feed h264 NALUs into the `MediaSourceAPI`, you have to send it boxed in an MP4. 

The creation of the `MediaSource` is done here https://github.com/codewithpassion/foxglove-studio-h264-extension/blob/custom-mse/src/H264Video.tsx

And the MP4 boxing here: https://github.com/codewithpassion/foxglove-studio-h264-extension/blob/custom-mse/src/lib/remuxer.ts

Part of the MP4 creation is a part where I need to give the MP4 fragment a duration. 
https://github.com/codewithpassion/foxglove-studio-h264-extension/blob/custom-mse/src/lib/remuxer.ts#L107

If this duration is not set to a reasonable value, then the video does not play nicely. 
At the moment, I'm taking the FPS that is part of the video stream, and calculate the duration based off that. 

That leaves us still with the problem of the video only ever playing at 1x speed max. Like with jMuxer. 

@defunctzombie  If you get a chance, you can check out that branch and see if you can get things to run better. 
I'm a bit a loss on what else to try. 
Short of running a software decoder and converting frames to JPEG and then render them on a canvas.

codewithpassion:
Alright. After some more digging I came across `[WebCodecs](https://w3c.github.io/webcodecs/)`. The API is currently a draft and while Chrome (and derived like Edge) browsers support it, Firefox does not at this point. 

With this API its possible to decode raw H264 into video frames and then render them onto a canvas. 
This approach renders 'as fast as possible' and does not require a duration or playback rate. 

While this is using an API that is still in development, it does what I need it to do at the moment and therefore, I think **this PR** can be closed.

I've updated [my extension](https://github.com/codewithpassion/foxglove-studio-h264-extension) on the `main` branch to use a WebWorker and the WebCodecs.

PR #4620, opened 2022-10-12
Incorrect use of "Render Frame" for grid "Frame" selection.

### LINKED ISSUE #4619: Incorrect use of "Render Frame" for grid "Frame" selection.
image

When selecting the grid frame, the dropdown lists a `<Render Frame>` option. We have no such concept elsewhere in the 3D panel settings. This option refers to the grid rendering in the `Display frame`. This option text should match the setting label.

snosenzo:
**User-Facing Changes**
 n/a

**Description**
 - changed `Render Frame` to `Display frame`
 - also updated a log that was using similar terminology

Fixes #4619

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
@@ -1255,7 +1255,7 @@ export class Renderer extends EventEmitter<RendererEvents> {
         this.fixedFrameId = undefined;
         return;
       } else {
-        log.debug(`Setting display frame to ${this.renderFrameId}`);
+        log.debug(`Setting render frame to ${this.renderFrameId}`);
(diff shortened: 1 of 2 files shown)


defunctzombie (approved):
👍  do we wan to use the `< >` as before or have without?

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -1255,7 +1255,7 @@ export class Renderer extends EventEmitter<RendererEvents> {
         this.fixedFrameId = undefined;
         return;
       } else {
-        log.debug(`Setting render frame to ${this.renderFrameId}`);
+        log.debug(`Setting display frame to ${this.renderFrameId}`);
```
jtbandes, line 1258: I'd omit this change, it's making the debug log message inconsistent with the code. The variable is named `renderFrameId`

PR #4621, opened 2022-10-12
3D Panel will error on cyclical transforms within the TransformTree

### LINKED ISSUE #4618: 3D Panel can cause page freeze
**Description**
When I connect foxglove studio to my robot the page instantly freeze. So I record a bag file and it still doesn't work. However my rviz is working normally (https://github.com/KisonHe/bug-report-bag-files Got a screenshot)

So I made a ROS bag file and used https://studio.foxglove.dev/ to open it, still it freezes. I am not good at js stuff but I think the 3D panel should not cause browser freeze in any case. I tried opening on many devices including iPad/Chrome, windows/edge,chrome they all freeze. The bag file is here https://github.com/KisonHe/bug-report-bag-files 

After some try i think it might be caused by the tf topic, not sure though

- Version:
- OS: Many OS tried. Manjaro linux, ubuntu ,windows, iPadOS 's browers. All freeze
- Data source bag file at https://github.com/KisonHe/bug-report-bag-files 
- ROS distro (if applicable): melodic

**Steps To Reproduce**

Download the bag file from https://github.com/KisonHe/bug-report-bag-files and open it with foxglove studio.(With a 3d panel)

**Expected Behavior**
It should not freeze

**Actual Behavior**
It freeze

Thx, have a good day :)

snosenzo:
**User-Facing Changes**
 - 3D panel will throw an error when a file attempts to add a cyclical transform to the tree. The error tells them what parent/child frame pair first caused the cycle.
 - user must fix file in order to load transforms

**Description**
 - cycles in the transform tree will throw an error instead of freezing the chrome tab.
image

Fixes #4618

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/transforms/CoordinateFrame.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/transforms/CoordinateFrame.ts
@@ -27,7 +27,6 @@ const tempVec4: vec4 = [0, 0, 0, 0];
 const tempTransform = Transform.Identity();
 const tempMatrix = mat4Identity();
 const temp2Matrix = mat4Identity();
-const visitedNodes = new Set<string>();
 
 /**
  * CoordinateFrame is a named 3D coordinate frame with an optional parent frame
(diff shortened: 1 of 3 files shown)


jhurliman (changes requested):
Good fix, but a few changes would improve this:

- Cycles are not supported in the tf tree architecture, so it would be nice to detect this at a lower level in the transforms/ code and fail out.
- We should add a test for catching cycles in the transforms/ code
- The user facing changes section of the PR should not be empty, this does affect the end user experience for certain datasets
- The PR title should reflect what is being changed instead of the problem

jhurliman (approved):
This is definitely an improvement over freezing the app. We may want to surface this error in a more friendly way in the future, but I propose merging this now, and we can follow up with a second PR if that's desirable.

defunctzombie:
Good stuff. Does the `throw` added in this PR show up as a topic error?

snosenzo:
It does not, it comes up as a closeable page error. Just opened a thread in slack to discuss where this error should be

PR #4622, opened 2022-10-12
Fix BufferedIterableSource readUntil behavior

defunctzombie:
**User-Facing Changes**
Better buffering during playback

**Description**
The `BufferedIterableSource` is responsible for reading the data source ahead of the current playhead to buffer messages for playback. Prior to this change, the `BufferedIterableSource` would only update the `readUntil` time when it was done buffering and was waiting for the consumer to read more messages. This could result in a state where the buffer would not update the `readUntil` time while playback was still playing through buffered data.

This change fixes this behavior by always updating the `readUntil` time from the latest `readHead` during each loop iteration. This ensures it is always one `readAheadDuration` ahead of the `readHead` and allows the producer to continue filling the buffer.


jtbandes (approved):
Is there a unit test we can add for this?

defunctzombie:
> Is there a unit test we can add for this?

I have some so I should be able to modify for this behavior. I'll give it a try before merging.

PR #4623, opened 2022-10-12
Allow ros1 encoded messages for foxglove websocket

achim-k:
**User-Facing Changes**
None

**Description**
Allows subscribing to ros1 encoded messages that are send using the [foxglove websocket protocol](https://github.com/foxglove/ws-protocol/blob/87cae3504c560c65503e24c43620d75d54999bb8/docs/spec.md).


jtbandes (approved):
Mind blown. 😮 

I believe the reason we didn't support this already was that we assumed ROS folks would be using rosbridge server. But there's no reason we shouldn't support all the same encodings as we do for MCAP files.

Issue #4601, opened 2022-10-09
the self-host web app crashed after i drag my extension panel to web App while i using dockerfile


hql420621:
**Description**

- Version: latest
- OS: linux
- Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native): 
- ROS distro (if applicable):

**Steps To Reproduce**

step 1 ,run a docker image by using `docker run --rm -p "8080:8080" ghcr.io/foxglove/studio:latest`

step 2. open chrome then access the server 

![image](https://user-images.githubusercontent.com/57355203/194746463-8b089517-dd5c-46dc-a65e-84fc0fd2d3d6.png)

step3  drag my custom extension into web app,then it crashed!

**Expected Behavior**
the web app can install custom extension correctly ,

**Actual Behavior**
![image](https://user-images.githubusercontent.com/57355203/194746849-b04e103e-c130-424f-9ee5-105f23ea417a.png)

is there another way to install my custom extension while i use self-host docker excepting drag
can i specify the location where i install my custom extension ,does its stored in the cache,if so ,it will missing after i clear browser cache?

foxymiles:
So far I am not able to reproduce this. Does your extension work if you try to install it on https://studio.foxglove.dev?

hql420621:
> So far I am not able to reproduce this. Does your extension work if you try to install it on https://studio.foxglove.dev?

yes,it works if i install it on https://studio.foglove.dev? but not if i use docker in self-host way, and is there another way to install extension package into web app except drag the extension?

foxymiles:
Could you please try this again in the docker hosted version but this time have the chrome developer tools open and report any error you see there when the installation fails?

So far I am not able to reproduce this locally.

hql420621:
this is what developer tools report as below
![image](https://user-images.githubusercontent.com/57355203/195227993-1662109a-6866-4647-bd96-eabacdbbb200.png)

hql420621:
this is what developer tools report as below
![image](https://user-images.githubusercontent.com/57355203/195227993-1662109a-6866-4647-bd96-eabacdbbb200.png)

LJL0511:
I've had this problem. Studio must be deployed behind an https domain.

defunctzombie:
> way to install extension package into web app except drag the extension?

The other way we support installing extensions is via our automatic extension provisioning when you are signed into your Foxglove account. This requires using the hosted version of Studio or the Foxglove packaged Studio app.

[...]

PR #4630, opened 2022-10-13
Avoid calling `getAsFileSystemHandle` outside of a secure context

defunctzombie:
**User-Facing Changes**
Dragging a file into studio does not crash the browser when viewing Studio from an unsecure context.

**Description**
The `getAsFileSystemHandle` api is only available in a secure context. Calling it outside of a secure context crashes the tab and is un-catchable. This change adds a check to verify there is a secure context before trying to use the api.

Fixes: #4601


defunctzombie:
Tested by loading the app in a 192.168.x.x:8080 url and localhost:8080. I can drag/drop an mcap file in both apps without crashing. In the localhost variant the recents list is updated. In the 192.168 variant it is not since we do not have access to file system handles to save in indexeddb.

PR #4400, opened 2022-09-10
Listen on all interfaces for Ros1Player

defunctzombie:
**User-Facing Changes**
When publishing messages from studio using the ROS1 native connection, the internal ROS node within Studio will listen on all network interfaces.

**Description**
This change updates the Ros1Player to listen on all interfaces for the rosnode internal to studio. This mirrors the ros_comm c++ behavior.

Fixes: #3182

PR #4631, opened 2022-10-13
Revert ROS1 player to bind on ROS_HOSTNAME

### LINKED ISSUE #4598: publish panel can not publish the topic in version 1.27
**Description**
I can not publish the topic through the publish-panel, but it works in version 1.19

- Version:1.27
- OS:Ubuntu 18
- ROS distro (if applicable):

**Steps To Reproduce**

**Expected Behavior**
When I edit the topic to disarm the drone ,(and i swear I didn't get the topic wrong as well its datatype!) the console of drone should show this change.

**Actual Behavior**
Nothing happened/(ㄒoㄒ)/~~
I think it could be some bug in different version, please help me

foxymiles:
**User-Facing Changes**
Fixes an issue in publishing to a ROS1 system from studio.

**Description**
Fixes regression from https://github.com/foxglove/studio/pull/4400.

The issue is that the `hostname` we pass to the TCP server `listen` call is returned in the `requestTopic` call:

https://github.com/foxglove/ros1/blob/73839e31c678749042651ff852fa39a1e4634f29/src/RosFollower.ts#L175-L198

This needs to be an actual IP address that other subscribers can reach so returning something like `localhost` or `0.0.0.0` is nonsensical and breaks subscribers.

Fixes #4598

### DIFF
--- a/packages/studio-base/src/players/Ros1Player.ts
+++ b/packages/studio-base/src/players/Ros1Player.ts
@@ -114,15 +114,7 @@ export default class Ros1Player implements Player {
     };
     const tcpServer = await net.createServer();
 
-    // Mirror the ros_comm c++ library behavior when setting up the tcp server listener.
-    // ros_comm listens on all interfaces unless the hostname is explicity set to 'localhost'
-    // https://github.com/ros/ros_comm/blob/noetic-devel/clients/roscpp/src/libros/transport/transport_tcp.cpp#L393-L395
-    // https://github.com/ros/ros_comm/blob/f5fa3a168760d62e9693f10dcb9adfffc6132d22/clients/roscpp/src/libros/transport/transport.cpp#L67-L72
-    let listenHostname = undefined;
-    if (hostname === "localhost") {
-      listenHostname = "localhost";
-    }
-    await tcpServer.listen(undefined, listenHostname, 10);
+    await tcpServer.listen(undefined, hostname, 10);
 
     if (this._rosNode == undefined) {
       const rosNode = new RosNode({


defunctzombie:
I'm in favor of exploring this a bit more.

- Thoughts on changing the listenHostname to `0.0.0.0`
- And using https://nodejs.org/api/dns.html#dnslookuphostname-options-callback to get the IP address we should return to the `requestTopics` call?

foxymiles:
I think trying to make smarter guesses about the network environment is just going to make things mysteriously work and fail in ways that are highly dependent on the particular network environment studio is running on.

IMO we're better off just spelling out extremely clearly that ROS_HOSTNAME needs to be either a hostname that resolves consistently on all nodes in the network or a literal IP address that's reachable by all nodes.

defunctzombie (approved):
More specifically - when passing `undefined` the behavior is:

> If host is omitted, the server will accept connections on the [unspecified IPv6 address](https://en.wikipedia.org/wiki/IPv6_address#Unspecified_address) (::) when IPv6 is available, or the [unspecified IPv4 address](https://en.wikipedia.org/wiki/0.0.0.0) (0.0.0.0) otherwise.

We do want to listen on ANY (0.0.0.0) but need to make that explicit here. However when we listen on 0.0.0.0 our implementation of ROS1 returns the same IP in the `requestTopics` response and nodes are not able to connect. Our implementation here diverges from rosc++ which has different semantics around ROS_HOSTNAME and ROS_IP handling.

We should merge this change to get a fix out for _most_ cases and make a follow-up that correctly handles ROS semantics for ROS_HOSTNAME and ROS_IP or consider automatic identification of the request IP (which we can also get from the socket and might be more robust).

PR #4632, opened 2022-10-13
Renaming event dialog variables + fixing add row bug

aschimmichanga:
**User-Facing Changes**
When the add button is pressed for an event metadata entry, a new empty row is added after the row, not before.

**Description**
Fixing bug
Also renaming certain variables for code clarity and consistency across both studio and console implementations.


bryfox (commented):
I don't have strong feelings on the variable renaming ("entries" might imply an array of tuples to me, but then again "metadata" might imply the resulting object), so I'll let others review & sign off. But I'm +1 on the `index + 1` to insert below the clicked row.

package.json
```diff
@@ -135,5 +135,8 @@
     "webpack-dev-server": "4.11.0",
     "webpack-hot-middleware": "2.25.2"
   },
-  "packageManager": "yarn@3.1.0"
+  "packageManager": "yarn@3.1.0",
+  "dependencies": {
+    "use-immer": "0.7.0"
```
foxymiles, line 140: Let's add this to dev dependencies in packages/studio-base/package.json instead.

foxymiles (changes requested):
Looks fine we just need to move the use-immer dependency into the studio-base package.json instead of top-level.

foxymiles (approved):

PR #4634, opened 2022-10-13
Improve ROS1 multi-host connection establishment

defunctzombie:
**User-Facing Changes**
Improve handling of subscription connections from ROS1 nodes when using Studio and ROS nodes on different hosts.

**Description**
Upgrade the `@foxglove/electron-socket` and `@foxglove/ros1` packages to incorporate improvements for subscriber connectivity.

The ROS1 change is explained here: https://github.com/foxglove/ros1/pull/26.

The @foxglove/electron-socket change adds the new fields to the proxied http request from electron socket so it can be used by our ros1 client within Studio desktop which sits on top of electron-socket.


foxymiles:
I'm getting a new error on this branch:

```
"Cannot read properties of undefined (reading 'parse')"

"TypeError: Cannot read properties of undefined (reading 'parse')
    at ipaddr.process (webpack-internal:///../../node_modules/ipaddr.js/lib/ipaddr.js:949:27)
    at RosFollower.requestTopic (webpack-internal:///../../node_modules/@foxglove/ros1/dist/RosFollower.js:136:64)
    at async XmlRpcServer._methodCallHandler (webpack-internal:///../../node_modules/@foxglove/xmlrpc/dist/XmlRpcServer.js:65:29)
    at async XmlRpcServer._requestHandler [as handler] (webpack-internal:///../../node_modules/@foxglove/xmlrpc/dist/XmlRpcServer.js:49:29)
    at async eval (webpack-internal:///../../node_modules/@foxglove/electron-socket/dist/renderer/HttpServerRenderer.js:21:31)"
```

If I revert to your previous e451f1143782449d0dc97c1c8d4355f55d322fc8 commit it still works.

foxymiles:
I think this was working in your original commit because you patched it like this:

https://github.com/foxglove/studio/blob/e451f1143782449d0dc97c1c8d4355f55d322fc8/patches/ros1.patch

defunctzombie:
@foxymiles Thanks for finding that. That is crazy that they use `this` for a free function call :(. I've updated the ros1 library  with the fix and updated this branch to point at a new commit of ros1. Please give it another go.

foxymiles:
Works now. 👍 

One thing we should probably address in a followup -  studio defaults my ROS_HOSTNAME to just `anax` (my local machine name), which doesn't work. I have to manually change it to `anax.local`. We should improve the logic for guessing a default ROS_HOSTNAME.

jhurliman (approved):

foxymiles:
> @foxymiles that sounds like a DNS configuration issue specific to your network or host. Using mDNS (.local) instead of the DNS protocol is not always the right choice.

I think we must have changed this behavior since it used to default to `.local`. I suspect on Macs at least this is likely to work more often than the bare hostname.

PR #4642, opened 2022-10-14
Add IteratorCursor and getMessageCursor to iterable player architecture

defunctzombie:
**User-Facing Changes**
Faster preloading of large datasets when reading from data platform, local file, and remote file sources.

**Description**
With IterablePlayer we changed the architecture for how we read messages from data sources. We introduced a new IIterableSource interface to expose a _messageIterator_. The message iterator provides the caller the ability to read messages from the data source one by one. This provides flexibility to the caller and allows for receiving a message as soon as the data source is able to provide a messages.

In a recent change, we moved the data source message reading into web workers. This provided a UX improvement by moving expensive CPU decompression and deserialization work off the main thread; data sources can deserialize the upcoming messages in parallel to panels rendering the current frame. However, this change came with a tradeoff. Since the interface for reading messages from the data source was a message iterator, getting each message from the data source required an RPC call to the webworker. To understand how this can impact loading we need to understand the two ways data is loaded from sources.

There are two consumers of messages within the player: playback and preloading. Playback reads messages for all subscribed topics when the user wants to be "playing" the data. It reads small time slices of message data from the source (on the order of 16ms) and produces state updates with these messages sometimes called "frames". On the other hand, preloading wants to read all the messages in the data source for preloaded subscriptions. Here is where we see the downsides of the move to webworkers.

For large datasets (many hundred thousand) messages, preloading the data (i.e. to plot a signal) would result in several hundred thousand RPC calls. The overhead of making these calls add up and negatively impact the preloading experience. When viewing a plot panel the UX of watching the data load would be worse for some datasets than before the move to webworkers.

This change introduces a new interface on top of message iterators: cursors. A source can optionally provide a method to return an IMessageCursor instance. The message cursor interface has a `readUntil` method. This method allows the caller to make one _read_ call and receive all the messages as a response. For web workers, the same amount of data is sent over the RPC channel but it is done in far fewer calls.

The improvements from this approach are shown in the below benchmarks. Specifically the _single update_ benchmark shows the time to preload ~700k messages decrease from ~54 seconds to ~13 seconds. And when including plot panel updates (which introduces main thread overhead) we see the time decrease from ~77 seconds to ~30 seconds.

## Benchmarks

To benchmark the user impact from this change I ran two sets of benchmarks using two different files against main and this PR.

In both sets I had a single plot panel with the following series:

```
/diagnostics.status[:].values[:]{key=="throttle_sensor"}.value
```

And in both sets I used the nuscenes sample data mcap with ~22k `/diagnostics` messages and another sample over several nuscenes drives with ~708k `/diagnostics` messages.

Both sets instrument the total duration taken to load the messages in `BlockLoader`. I instrumented block loader to record the time before and after the load cycle and recorded the difference.

All benchmarks were run using a production build.

### Updates during pre-loading

This benchmark uses the default logic which updates progress (which sends new preload data to the plot panel for rendering) as new _blocks_ are loaded in BlockLoader.

| Variant | Message Count | Duration (ms)
| --- | --- | --- |
| main | 22,019 |4728 |
| main | 708,777 | 77776|
| readUntil | 22,019 |1078 |
| readUntil | 708,777 | 30144|

### Single update at end of pre-loading

This benchmark comments out the intermediate progress updates and only sends one progress update after the data is loaded in BlockLoader. This takes plot rendering out of the timing path and shows the time taken to read the underlying data with BlockLoader.

| Variant | Message Count | Duration (ms)
| --- | --- | --- |
| main  | 22,019|2429 |
| main |708,777 |54800 |
| readUntil | 22,019 |630 |
| readUntil | 708,777 | 13773|


packages/studio-base/src/players/IterablePlayer/IIterableSource.ts
```diff
@@ -73,6 +73,30 @@ export type GetBackfillMessagesArgs = {
   abortSignal?: AbortSignal;
 };
 
+export interface IMessageCursor {
```
jtbandes, line 76: Top level comment on what a message cursor is?

packages/studio-base/src/players/IterablePlayer/IIterableSource.ts
```diff
@@ -73,6 +73,30 @@ export type GetBackfillMessagesArgs = {
...
+
+  /**
+   * End the cursor
+   *
+   * Release any held resources by the cursor.
+   *
+   * Calls to next() and readUntil() should return `undefined` after a cursor is ended as if the
+   * cursor reached the end of its messages.
+   */
+  end(): Promise<void>;
```
jtbandes, line 97: maybe call this close(), destroy(), or abort()?

packages/studio-base/src/players/IterablePlayer/IteratorCursor.ts
```diff
@@ -0,0 +1,82 @@
...
+
+import { compare } from "@foxglove/rostime";
+import { Time } from "@foxglove/studio";
+
+import type { IMessageCursor, IteratorResult } from "./IIterableSource";
+
+/// IteratorCursor implements a IMessageCursor interface on top of an AsyncIterable
+class IteratorCursor implements IMessageCursor {
+  private _iter: AsyncIterableIterator<Readonly<IteratorResult>>;
+  private _lastIteratorResult?: IteratorResult;
```
jtbandes, line 13: comment on this variable to describe why we need to track it

packages/studio-base/src/players/IterablePlayer/WorkerIterableSource.ts
```diff
@@ -82,6 +86,38 @@ export class WorkerIterableSource implements IIterableSource {
...
   }
 
+  public getMessageCursor(args: MessageIteratorArgs & { abort?: AbortSignal }): IMessageCursor {
+    if (this._worker == undefined) {
+      throw new Error(`WorkerIterableSource is not initialized`);
+    }
+
+    // An AbortSignal is not clonable, so we remove it from the args and send it as a separate argumet
+    // to our worker getBackfillMessages call. Our installed Comlink handler for AbortSignal handles
+    // making the abort signal available within the worker.
```
jtbandes, line 96: so the comlink handler is only dispatched for top-level args, not objects nested inside of other objects?

packages/studio-base/src/players/IterablePlayer/WorkerIterableSource.ts
```diff
@@ -82,6 +86,38 @@ export class WorkerIterableSource implements IIterableSource {
...
+
+      async readUntil(end: Time) {
+        const messageCursor = await messageCursorPromise;
+        return await messageCursor.readUntil(end);
+      },
+
+      async end() {
+        const messageCursor = await messageCursorPromise;
+        await messageCursor.end();
+        messageCursor[Comlink.releaseProxy]();
```
jtbandes, line 114: maybe put this in finally?

jtbandes (commented):
Overall makes sense. I see that this mostly seems to affect BlockLoader code paths and not other stuff in the IterablePlayer, because the benefit is probably best for large preload tasks. But do we also plan to eventually use readUntil for regular playback when we are loading a frame of messages?

Looks like you still have some fixmes to resolve.

defunctzombie:
> But do we also plan to eventually use readUntil for regular playback when we are loading a frame of messages?

Yes - I wanted to keep this change focused on block loader to minimize potential disruption and the diff if there is some issue. I think it would help regular playback just a pinch.

> Looks like you still have some fixmes to resolve.

Indeed. My plan was to resolve those after the first round of comments.

packages/studio-base/src/players/IterablePlayer/WorkerIterableSource.ts
defunctzombie: Correct. Learned that through some painful exercises.

packages/studio-base/src/players/IterablePlayer/IIterableSource.ts
defunctzombie: Don't like `end`? Its pretty common for nodejs streams or request/responses.

jtbandes: https://developer.mozilla.org/en-US/docs/Web/API/TimeRanges/end - not a destructive action

🤷 end is fine if that's commonly used elsewhere.

jtbandes (approved):

PR #4643, opened 2022-10-14
Improve Cyclical Transform Error state

### LINKED ISSUE #4636: Improve Cyclical Transform Error state
**Description**
 - Right now tfMessages that cause a cyclical transform cause a Panel-wide error in the 3D panel we want to improve this behavior and make it a topic error belonging to the child
 - 3D panel can render incorrectly but should not crash
 - for URDF's with cyclical transforms we'll want to show it on the URDF entry

cc @jhurliman

snosenzo:
**User-Facing Changes**
 - cyclical transforms will no longer error 3D Panel
 - settings panel errors will appear next to relevant childFrame nodes, and, if relevant, responsible URDF nodes.

**Description**
 - add transform returns whether a cycle was detected or not
 - child frames are still added and can be viewed in UI but are not re-parented to avoid creating cycle
 - responsible URDFs will also get a settings node error

My one issue with this is that it is a little hidden to the user unless they open the Transforms dropdown, since it's on the settings node itself rather than a top-level issue. If it's a subtle mistake in the 3D scene it will not be evident to them.

image
image

Fixes #4636

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
@@ -950,13 +950,13 @@ export class Renderer extends EventEmitter<RendererEvents> {
       this.settings.errors.add(
         ["transforms", `frame:${childFrameId}`],
         CYCLE_DETECTED,
-        `Transform tree cycle detected: Received transform with parent "${parentFrameId}" and child "${childFrameId}", but "${childFrameId}" is already an ancestor of "${parentFrameId}".`,
+        `Transform tree cycle detected: Received transform with parent "${parentFrameId}" and child "${childFrameId}", but "${childFrameId}" is already an ancestor of "${parentFrameId}". Transform message dropped.`,
       );
       if (errorSettingsPath) {
         this.settings.errors.add(
           errorSettingsPath,
           CYCLE_DETECTED,
-          `Attempted to add cyclical transform: Frame "${parentFrameId}" cannot be the parent of frame "${childFrameId}". Frame was not re-parented`,
+          `Attempted to add cyclical transform: Frame "${parentFrameId}" cannot be the parent of frame "${childFrameId}". Transform message dropped.`,
         );
       }
     }
--- a/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
@@ -210,6 +210,7 @@ const FOLLOW_TF_PATH = ["general", "followTf"];
 const NO_FRAME_SELECTED = "NO_FRAME_SELECTED";
 const FRAME_NOT_FOUND = "FRAME_NOT_FOUND";
 const TF_OVERFLOW = "TF_OVERFLOW";
+const CYCLE_DETECTED = "CYCLE_DETECTED";
 
 // An extensionId for creating the top-level settings nodes such as "Topics" and
 // "Custom Layers"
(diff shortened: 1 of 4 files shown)


jtbandes:
> My one issue with this is that it is a little hidden to the user unless they open the Transforms dropdown, since it's on the settings node itself rather than a top-level issue.

Could we just put the error on the Transforms section header instead of the specific frame's node?

snosenzo:
oh we can we do that? I just wasn't sure that was a thing.  is there an example of where this is done elsewhere?

packages/studio-base/src/panels/ThreeDeeRender/transforms/TransformTree.ts
```diff
@@ -31,21 +31,23 @@ export class TransformTree {
...
-      updated = true;
+      if (!cycleDetected) {
+        frame.setParent(this.getOrCreateFrame(parentFrameId));
+        updated = true;
+      }
     }
 
     frame.addTransform(time, transform);
-    return updated;
+    return { updated, cycleDetected };
```
jtbandes, line 50: Is it right that there can only be `updated: true, cycleDetected: false`, `updated: false, cycleDetected: true`, and `updated: false, cycleDetected: false`? Makes me wonder if we should just return a 3-state string enum like `"not-updated" | "updated" | "cycle"`. Guess it doesn't make a huge difference.

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -932,18 +933,39 @@ export class Renderer extends EventEmitter<RendererEvents> {
...
     if (updated) {
       this.coordinateFrameList = this.transformTree.frameList();
       this.emit("transformTreeUpdated", this);
     }
 
+    if (cycleDetected) {
+      this.settings.errors.add(
+        ["transforms", `frame:${childFrameId}`],
+        CYCLE_DETECTED,
+        `Transform tree cycle detected: Frame "${parentFrameId}" cannot be the parent of frame "${childFrameId}". Frame was not re-parented`,
```
jtbandes, line 958:
```suggestion
        `Transform tree cycle detected: Received transform with parent "${parentFrameId}" and child "${childFrameId}", but "${childFrameId}" is already an ancestor of "${parentFrameId}".`,
```

cc @banisadr for wording suggestions

jtbandes (approved):

packages/studio-base/src/panels/ThreeDeeRender/transforms/TransformTree.ts
snosenzo: yeah it could, I kinda prefer having the two boolean values to be more explicit about the status of things that callers of the method might care about. Also makes it a bit more extensible for possible future changes

jhurliman:
> > My one issue with this is that it is a little hidden to the user unless they open the Transforms dropdown, since it's on the settings node itself rather than a top-level issue.
> 
> Could we just put the error on the Transforms section header instead of the specific frame's node?

Errors being buried is a larger UX issue with Studio that I don't think we should be hacking around as one-offs. Attaching the error to the specific transform provides more information to the user by showing which frame has an ambiguous parent. Taking this in the far opposite direction, we would surface all topic errors in one place at the top level of the app. This is the previous design we moved away from.

jhurliman (changes requested):
Overall structure looks good, but `addTransform()` is a performance-sensitive method. We can't create a heap allocation for the return value.

packages/studio-base/src/panels/ThreeDeeRender/transforms/TransformTree.ts
```diff
@@ -31,21 +32,23 @@ export class TransformTree {
     parentFrameId: string,
     time: Time,
     transform: Transform,
-  ): boolean {
-    let updated = !this.hasFrame(frameId);
+  ): { updated: boolean; cycleDetected: boolean } {
```
jtbandes, line 35: I think it's weird to use a global variable for this without documenting it in the comment above. Would help clarify things if it were accepted as an out parameter as well: `outStatus: {updated:boolean;cycleDetected:boolean}`

All in all, I think I would find a simple string union or `enum AddTransformResult { NOT_UPDATED, UPDATED, CYCLE_DETECTED }` to be clearer code style than returning a reference to a global object.

defunctzombie:
> Errors being buried is a larger UX issue with Studio that I don't think we should be hacking around as one-offs. Attaching the error to the specific transform provides more information to the user by showing which frame has an ambiguous parent. Taking this in the far opposite direction, we would surface all topic errors in one place at the top level of the app. This is the previous design we moved away from.

Agreed. Let's stick this on the correct node and separately look at how the settings tree or panel need to make users aware of errors when the settings are closed.

/cc @foxymiles @2metres for visibility since yall have talked about this before and want to give you more data points to think about - not trying to go into that discussion here

defunctzombie:
`Frame was not re-parented` sounds awkward. Is the behavior that we ignore the transform trying to re-parent the frame?

If that is the behavior is to drop the transform let's be clear that is what happens. If the behavior is to include the transform (but not re-parent the frame) what does it mean to keep this transform? If we didn't re-parent the frame what is the purpose of the transform?

snosenzo:
> Frame was not re-parented sounds awkward. Is the behavior that we ignore the transform trying to re-parent the frame?

I updated to say that `Transform message was dropped.` and also am no longer adding the transform to the child frame when the parent isn't added because I don't think it makes sense to add it. Previously I was just to have as little intervention on the logic as possible, since it's going to show up wrong no matter what. But I think this makes a clearer cut that the message will be dropped and no change will be done.

jhurliman (approved):

PR #4649, opened 2022-10-17
Allow cdr (ros2) encoded messages for foxglove websocket

achim-k:
**User-Facing Changes**
None

**Description**
Similar to #4623, this PR allows subscribing to `cdr` encoded messages that are sent using the [foxglove websocket protocol](https://github.com/foxglove/ws-protocol/blob/87cae3504c560c65503e24c43620d75d54999bb8/docs/spec.md).


jtbandes (approved):
Probably when we make the next revision to the websocket protocol, we should separate message encoding & schema encoding to match MCAP.

defunctzombie:
@jtbandes can you kick the CI system for this and merge this - I think it is stuck in GH actions hell.

jtbandes:
@achim-k can you push another commit to kick CI?

PR #4651, opened 2022-10-17
Additional validation for Foxglove.Grid and Support Grids with row padding

### LINKED ISSUE #4650: Validate Grid Rows is integer and properly support padding for rows
**Description**
 - currently the logic of iterating through the data array does not support padding at the end of rows. It iterates through the data assuming that there is no padding at the end of a row of data. 
 - Also the row count is calculated by dividing the bytelength by the row_stride without validating that it is an integer. If the value is not an integer it should add an error to the topic

snosenzo:
**User-Facing Changes**
 - additional validation to foxglove grid
- support grids with row padding (row_stride > column_count * cell_stride)

**Description**
 - added another story to make sure grids with row padding work

Fixes #4650

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/FoxgloveGrid.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/FoxgloveGrid.ts
@@ -283,11 +283,26 @@ export class FoxgloveGrid extends SceneExtension<FoxgloveGridRenderable> {
       );
       return;
     }
-    const { cell_stride } = foxgloveGrid;
-    const { cols, rows } = getFoxgloveGridDimensions(foxgloveGrid);
-    const size = cols * rows * foxgloveGrid.cell_stride;
+    const { cell_stride, row_stride, column_count: cols } = foxgloveGrid;
+    const rows = foxgloveGrid.data.byteLength / row_stride;
+
+    if (Math.floor(cols) !== cols || Math.floor(rows) !== rows) {
+      const message = `FoxgloveGrid column count (${foxgloveGrid.column_count}) or row count (${rows} = data.byteLength / row_stride) is not an integer.`;
+      invalidFoxgloveGridError(this.renderer, renderable, message);
+      return;
+    }
+
+    const size = row_stride * rows;
     if (foxgloveGrid.data.length !== size) {
-      const message = `FoxgloveGrid data length (${foxgloveGrid.data.length}) is not equal to cols ${cols} * rows ${rows} * cell_stride ${cell_stride}`;
+      const message = `FoxgloveGrid data length (${foxgloveGrid.data.length}) is not equal to cols (${cols}) * rows (${rows}) * cell_stride ${cell_stride}`;
+      invalidFoxgloveGridError(this.renderer, renderable, message);
+      return;
+    }
+
+    if (cell_stride * cols > row_stride) {
+      const message = `FoxgloveGrid row_stride (${row_stride}) does not fit requisite column_count (${cols}) with cell stride (${cell_stride}) = (${
+        cols * cell_stride
+      }) `;
       invalidFoxgloveGridError(this.renderer, renderable, message);
       return;
     }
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/panels/ThreeDeeRender/renderables/FoxgloveGrid.ts
```diff
@@ -283,11 +283,26 @@ export class FoxgloveGrid extends SceneExtension<FoxgloveGridRenderable> {
...
       return;
     }
-    const { cell_stride } = foxgloveGrid;
-    const { cols, rows } = getFoxgloveGridDimensions(foxgloveGrid);
-    const size = cols * rows * foxgloveGrid.cell_stride;
+    const { cell_stride, row_stride, column_count: cols } = foxgloveGrid;
+    const rows = foxgloveGrid.data.byteLength / row_stride;
+
+    if (Math.floor(cols) !== cols || Math.floor(rows) !== rows) {
+      const message = `FoxgloveGrid column count (${foxgloveGrid.column_count}) or row count (${rows} = data.byteLength / row_stride) is not an integer.`;
```
jtbandes, line 290:
```suggestion
      const message = `Grid column count (${cols}) or row count (${rows} = data length ${data.byteLength} / row_stride ${row_stride}) is not an integer.`;
```

packages/studio-base/src/panels/ThreeDeeRender/renderables/FoxgloveGrid.ts
```diff
@@ -283,11 +283,26 @@ export class FoxgloveGrid extends SceneExtension<FoxgloveGridRenderable> {
...
+      invalidFoxgloveGridError(this.renderer, renderable, message);
+      return;
+    }
+
+    const size = row_stride * rows;
     if (foxgloveGrid.data.length !== size) {
-      const message = `FoxgloveGrid data length (${foxgloveGrid.data.length}) is not equal to cols ${cols} * rows ${rows} * cell_stride ${cell_stride}`;
+      const message = `FoxgloveGrid data length (${foxgloveGrid.data.length}) is not equal to cols (${cols}) * rows (${rows}) * cell_stride ${cell_stride}`;
```
jtbandes, line 297:
```suggestion
      const message = `Grid data length (${foxgloveGrid.data.length}) is not equal to cols (${cols}) * rows (${rows}) * cell_stride ${cell_stride}`;
```

packages/studio-base/src/panels/ThreeDeeRender/renderables/FoxgloveGrid.ts
```diff
@@ -283,11 +283,26 @@ export class FoxgloveGrid extends SceneExtension<FoxgloveGridRenderable> {
...
+    const rows = foxgloveGrid.data.byteLength / row_stride;
+
+    if (Math.floor(cols) !== cols || Math.floor(rows) !== rows) {
+      const message = `FoxgloveGrid column count (${foxgloveGrid.column_count}) or row count (${rows} = data.byteLength / row_stride) is not an integer.`;
+      invalidFoxgloveGridError(this.renderer, renderable, message);
+      return;
+    }
+
+    const size = row_stride * rows;
     if (foxgloveGrid.data.length !== size) {
```
jtbandes, line 296: This condition will always be true the way it's currently written. You defined `size` as `row_stride * rows`, and `rows` as `foxgloveGrid.data.byteLength / row_stride`.

Also, the error message doesn't really match what is being checked (we aren't checking cell_stride).

jtbandes (changes requested):

packages/studio-base/src/panels/ThreeDeeRender/renderables/FoxgloveGrid.ts
snosenzo: Ah yeah, thanks for catching that. Do we even need this check then? It's not possible to check for the size of the grid to correspond with the length of the array since the length determines the number of rows.

snosenzo: I removed it since the other checks will better validate the data.

jtbandes (approved):

PR #4653, opened 2022-10-17
rosbridge: Readvertise published topics on reconnection.

### LINKED ISSUE #4499: Reconnection to Rosbridge does not reliably re-advertise topics
This comes from a report from a user of my extension, not sure what version they're using but it's fairly recent and on macOS.

My panel has a button that publishes a `/bladescript/execute` message. The panel is written as a React functional component like so:

```jsx
function BladescriptPanel(
    { context }: { context: PanelExtensionContext }
): JSX.Element {
...
    // Set up rendering for the panel
    useLayoutEffect(() => {
        ...

        // Advertise that we publish scripts
        context.advertise?.("/bladescript/execute", "std_msgs/String");
    }, [context]);

    ...
}
```

Occasionally the user gets a complaint from Rosbridge when pressing the button that it was not expecting the publication of the `/bladescript/execute` message. My theory is that this has to do with when the connection is lost and then the connection gets re-established.

Am I misusing the API here?

image

achim-k:
**User-Facing Changes**
Fixed an issue where Rosbridge connections would not advertise when re-connecting

**Description**
Fixes #4499 by actively re-advertising published topics when reconnecting with the rosbridge server.

### DIFF
--- a/packages/studio-base/src/players/RosbridgePlayer.ts
+++ b/packages/studio-base/src/players/RosbridgePlayer.ts
@@ -160,7 +160,7 @@ export default class RosbridgePlayer implements Player {
       this._problems.removeProblem("rosbridge:connection-failed");
       this._rosClient = rosClient;
 
-      this._setupPublishers({ advertisePublishers: true });
+      this._setupPublishers();
       void this._requestTopics({ forceUpdate: true });
     });
 
@@ -579,7 +579,7 @@ export default class RosbridgePlayer implements Player {
     }
     this._topicPublishers.clear();
     this._advertisements = publishers;
-    this._setupPublishers({ advertisePublishers: false });
+    this._setupPublishers();
   }
 
   public setParameter(_key: string, _value: ParameterValue): void {
@@ -662,7 +662,7 @@ export default class RosbridgePlayer implements Player {
     // no-op
   }
 
-  private _setupPublishers(opt?: { advertisePublishers: boolean }): void {
+  private _setupPublishers(): void {
     // This function will be called again once a connection is established
     if (!this._rosClient) {
       return;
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/players/RosbridgePlayer.ts
```diff
@@ -160,7 +160,7 @@ export default class RosbridgePlayer implements Player {
       this._problems.removeProblem("rosbridge:connection-failed");
       this._rosClient = rosClient;
 
-      this._setupPublishers();
+      this._setupPublishers({ advertisePublishers: true });
```
jtbandes, line 163: Why do we want to advertise here but not in setPublishers?

jtbandes:
Please update the PR description to include a "user-facing changes" section with a quick explanation of the bug being fixed, to be included in the release notes.

packages/studio-base/src/players/RosbridgePlayer.ts
achim-k: Main motivation was to avoid sending unnecessary advertise/unadvertise operations to the backend as noticed in https://github.com/foxglove/studio/issues/4499#issuecomment-1276346285. However, your question is valid and I think you are right in that we should stay consistent and in both cases actively advertise the topics. The question is then, should we avoid sending intermediate advertise/unadvertise operations and if so, how?

achim-k: We decided to always advertise published topics.

jtbandes (approved):
Looks good. But this change does have user-facing changes: we can say something like “Fixed an issue where Rosbridge connections would not advertise when re-connecting”.

Issue #4652, opened 2022-10-17
Empty "Selected object" inspector popup is unnecessary


amacneil:
**Description**
User feedback: when clicking the cursor button to enter object inspect mode (in the 3D panel and Image panel) results in this popup on your screen, even when no object is selected.

<img width="411" alt="image" src="https://user-images.githubusercontent.com/637671/196243343-14b655be-d2df-414f-bb71-92ed9a020c17.png">

These instructions and popup aren't really necessary: the cursor button turning blue would be sufficient to indicate you are in Inspect mode.

amacneil:
cc @2metres @banisadr for feedback

banisadr:
I disagree that the pop-up and instructions aren't necessary. For a new user, it is not at all obvious what the arrow does, what object inspector is, or how to use them - the pop-up and instructions solve that.

That said, yes, the empty pop-up is annoying, particularly if using a small panel:
<img width="428" alt="image" src="https://user-images.githubusercontent.com/12058745/196260249-1a061d0f-e854-459a-80ce-af3cf9ff59a6.png">

With a larger panel, however, It's not really in the way and is a good reminder that you're in inspector mode, even for an experienced user:
<img width="1302" alt="image" src="https://user-images.githubusercontent.com/12058745/196259601-aa1a6e86-a04a-4ce0-b0ff-9bc62a9955d0.png">

I would suggest we keep the default behavior but allow the users to collapse or dismiss the empty pop-up.

defunctzombie:
Proposing a middle ground between being able to keep the selection mode on and the dialog hidden when nothing is selected:

- When a user toggles the pointer we open the popup to display the "Click an object" help text. We can make the help text go away after a few moments.
- When a user de-selects an object we hide the popup but keep the selection mode enabled.
- When a user clicks the "collapse" icon and an object is selected, we unselect the object and hide the dialog (but keep the selection mode on).
- When a user clicks another object we re-show the popup.
- When a user clicks the pointer when selection mode is enabled we turn off selection mode.

amacneil:
@defunctzombie I think it is potentially confusing if clicking the collapse icon keeps it in selection mode.

Maybe we just show the popup when you first enter selection mode, and dismiss it as soon as you click on empty space? (clicking on "something" will cause the popup to populate with data, clicking on empty space will hide it - at least in 3d panel).

defunctzombie:
Proposed operation mode per @amacneil feedback:

[...]

PR #4654, opened 2022-10-17
Hide "Selected object" popup when no object is selected

defunctzombie:
**User-Facing Changes**

**Description**

Fixes: #4652


defunctzombie:
I still believe we should hide the popup when you click away from an object but show it on first click of the cursor. I don't have bandwidth for this change tho so am closing this.

PR #4656, opened 2022-10-17
[3D] Disable TF preloading

jhurliman:
**User-Facing Changes**
- [3D] Preloading of transform topics is disabled until performance and correctness issues are resolved

**Description**
With the current implementation of preloaded topics in the extension API, whenever a new `allFrames` is received we must 
iterate over the entire list looking for new messages. Further complicating things, message events in Studio do not have 
unique keys so we must re-ingest every TF each time `allFrames` changes.

One solution to this would be to provide a delta API that incrementally provides preloaded message events similar to how 
`currentFrame` works.

### DIFF
--- a/packages/den/collection/ArrayMap.ts
+++ b/packages/den/collection/ArrayMap.ts
@@ -57,7 +57,7 @@ export class ArrayMap<K, V> {
   public removeAfter(key: K): void {
     const index = this.binarySearch(key);
     const greaterThanIndex = index >= 0 ? index + 1 : ~index;
-    this._list.splice(greaterThanIndex);
+    this._list.length = greaterThanIndex;
   }
 
   /** Access the first key/value tuple in the list, without modifying the list. */
--- a/packages/den/collection/ArrayMap.ts
+++ b/packages/den/collection/ArrayMap.ts
@@ -53,6 +53,13 @@ export class ArrayMap<K, V> {
     return this._list.pop();
   }
 
+  /** Removes all elements with keys greater than the given key. */
+  public removeAfter(key: K): void {
+    const index = this.binarySearch(key);
+    const greaterThanIndex = index >= 0 ? index + 1 : ~index;
+    this._list.splice(greaterThanIndex);
+  }
+
   /** Access the first key/value tuple in the list, without modifying the list. */
   public minEntry(): [K, V] | undefined {
     return this._list[0];
--- a/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
@@ -521,8 +521,9 @@ export class Renderer extends EventEmitter<RendererEvents> {
    * This is useful when seeking to a new playback position or when a new data source is loaded.
    */
   public clear(): void {
-    // This must be cleared before calling `SceneExtension#removeAllRenderables()` to allow
-    // extensions to add errors back afterward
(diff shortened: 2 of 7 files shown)


packages/den/collection/ArrayMap.ts
```diff
@@ -53,6 +53,13 @@ export class ArrayMap<K, V> {
     return this._list.pop();
   }
 
+  /** Removes all elements with keys greater than the given key. */
+  public removeAfter(key: K): void {
+    const index = this.binarySearch(key);
+    const greaterThanIndex = index >= 0 ? index + 1 : ~index;
+    this._list.splice(greaterThanIndex);
```
jtbandes, line 60: Since splice returns the removed elements, it might be faster to use `this._list.length = greaterThanIndex`

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -442,17 +442,20 @@ export class Renderer extends EventEmitter<RendererEvents> {
     this.addDatatypeSubscriptions(FRAME_TRANSFORM_DATATYPES, {
       handler: this.handleFrameTransform,
       forced: true,
-      preload: config.scene.transforms?.enablePreloading ?? true,
+      // Disabled until we can efficiently preload transforms
```
jtbandes, line 445: Is there a ticket number we can reference in these comments?

packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
```diff
@@ -393,21 +392,20 @@ export function ThreeDeeRender({ context }: { context: PanelExtensionContext }):
...
-  useEffect(
-    () => setRenderer(canvas ? new Renderer(canvas, configRef.current) : undefined),
-    [canvas, configRef, config.scene.transforms?.enablePreloading],
-  );
+  const rendererRef = useRef<Renderer | undefined>(undefined);
+  useEffect(() => {
+    const newRenderer = canvas ? new Renderer(canvas, configRef.current) : undefined;
+    setRenderer(newRenderer);
+    rendererRef.current = newRenderer;
+    return () => rendererRef.current?.dispose();
```
jtbandes, line 400: for completeness, set rendererRef.current = undefined in the cleanup

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -521,8 +524,9 @@ export class Renderer extends EventEmitter<RendererEvents> {
    * This is useful when seeking to a new playback position or when a new data source is loaded.
    */
   public clear(): void {
-    // This must be cleared before calling `SceneExtension#removeAllRenderables()` to allow
-    // extensions to add errors back afterward
+    // These must be cleared before calling `SceneExtension#removeAllRenderables()` to allow
+    // extensions to add transforms and errors back afterward
+    this.transformTree.clearAfter(this.currentTime);
```
jtbandes, line 529: why not clear the whole tree?

jtbandes (approved):

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
jhurliman: The UX is quite bad when the whole tree is cleared. The TF tree in the sidebar disappears for a frame, then reappears as a partially connected tree, then fully rebuilds as data is loaded. Any open TF sidebar nodes and your position in the sidebar is lost.

PR #4658, opened 2022-10-18
[Testing] Throw an error if allFrames receiveTime goes backwards

jhurliman:
Starting playback in the middle of a bag then randomly seeking around while playing back should trigger this after a short while. There's probably a more precise set of repro steps once we understand exactly why this happens.

Screen Shot 2022-10-18 at 10 57 39 AM


defunctzombie:
I've not been able to get this branch to crash while using one of the more recent nuscenes mcap files. Please share a file that reproduces the crash.

defunctzombie:
@jhurliman follow up on my above comment. Can you point me at a file that I can use to reproduce this?

jhurliman:
> @jhurliman follow up on my above comment. Can you point me at a file that I can use to reproduce this?

`dev_RCt8o7KdWco4ErdY_2022-05-24-15-38-23.mcap`

defunctzombie:
Closing this. If you can provide the file I can use to reproduce this we can revisit. I don't know what to do with a posted file name.

PR #4659, opened 2022-10-18
Revert "[3D] Remove TF preloading"

defunctzombie:
Reverts foxglove/studio#4656

Getting this ready in case we want to bring this feature back.


defunctzombie:
Reading the PR description I don't see what is incorrect about preloading or what we are fixing by reverting it. We already included a setting to disable preloading for users to manually enable.

The mention of receiveTime being out of order in https://github.com/foxglove/studio/pull/4658 I've not been able to repro this with my sample files.

When reverting a feature we should be more clear on the revert what the situations are that fail and how those compare to working situations.

PR #4661, opened 2022-10-18
 Prevent LinePrimitive scene entities from crashing with empty `points`

snosenzo:
**User-Facing Changes**
 - prevent LinePrimitive scene entities from crashing when `points` is empty and not a loop

**Description**
 - prevent LinePrimitive scene entities from crashing when `points` is empty and not a loop
- add test case to SceneEntities story


jtbandes (approved):
Thanks!

Issue #4647, opened 2022-10-17
diagonal and blocky artifacts with overlapping foxglove.Grid instances


james-rms:
**Description**
A continuation of #4641 - I tried shortening the far clipping plane of the camera by a factor of 50 (which should increase the bit depth of the Z-buffer by the same factor) but this made no improvement on the number of block artifacts that I previously attributed to z-fighting. There are also some diagonal lines appearing in the artifacts which shouldn't happen as a result of Z-fighting. Some screenshots:

![image](https://user-images.githubusercontent.com/18162835/196089491-3cd51fa5-0db8-4a87-8181-74d3dcc40321.png)
![image](https://user-images.githubusercontent.com/18162835/196089521-380f0534-1d58-4755-a336-cf07a9cc8900.png)

- Version: studio.foxglove.dev
- OS: Chrome Version 106.0.5249.103 (Official Build) (arm64)
- Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native): check the NuScenes sample data for a good example
- ROS distro (if applicable):

**Steps To Reproduce**

**Expected Behavior**

**Actual Behavior**

snosenzo:
Would a possible solution to this to be to change the example file to put the driveable_area on a coordinate frame as a child of `map` that has a translation like .05 above the map? (that way it can also be adjustable to users)

jtbandes:
I played with this a bit more and I saw that adjusting the near plane seemed to have the desired effect of removing the artifacts. So I do believe that we are seeing "regular" z-fighting here. I'm making a PR to allow enabling logarithmic depth: #4662

https://user-images.githubusercontent.com/14237/196517231-d46ffcbf-cf64-491b-9389-ccad8302b5bd.mov

Issue #4641, opened 2022-10-14
Grid z-fighting is annoying to work around


james-rms:
**Description**
Currently when generating demo data from NuScenes datasets, we place the driveable area grid 1cm above the map grid. Sometimes, z-fighting still occurs between these two planes:
https://user-images.githubusercontent.com/18162835/195738181-64b06ce8-e1d8-45f1-92c2-704c81d00eed.mov

Whether you see z-fighting depends on:
1. the distance between planes
2. the Z-buffer bit depth
3. the distance of the camera from the planes

Indeed when you zoom in close to the car, the flickering goes away.

I believe this represents a usability issue for users trying to make their visualisation messages look right. To compensate for this now, they have to modify their source data - this is painful if you have a long-running pipeline producing the source data.

### Have a safe distance

We might be able to solve this by documenting a known "safe distance" between planes to prevent z-fighting, based on whatever maximum zoom level we expect to place the camera at. I don't think we have a defined max zoom level though. Also, if this safe distance is too far (say 1m) it prevents users from actually representing their environment accurately in visualization.

### Polygon offset topic setting

One proposed solution is to add a "polygon offset" topic setting to studio, so that users can tweak this offset manually until z-fighting goes away. I think this might be a nice feature, but there would still be a usability problem, because they have no way to fix the experience for all users at once, or for all source data at once.

### Jack up the Z-buffer bit depth

Not sure how feasible this is.

### Fix it in the renderer

It may be possible to put in special logic to determine the polygon offset `foxglove.Grid` and `OccupancyGrid` objects automatically, which would allow users to use an arbitrarily small gap between planes to declare their order. I'm not sure how tough this would be.

## Details

- Version: studio.foxglove.dev, Chrome  106.0.5249.103 (Official Build) (arm64)
- OS: Mac OS 12.4
- Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native): Check the demo data here https://studio.foxglove.dev/?ds=foxglove-data-platform&ds.deviceId=dev_SUQ41s35DqPfGAoo&ds.end=2018-11-21T11%3A47%3A47.447Z&ds.importId=GtECnrokPxX8s4XM&ds.start=2018-11-21T11%3A47%3A27.948Z&layoutId=76bde952-b9e4-4eaf-be25-6c901e7fc099
- ROS distro (if applicable):

jtbandes:
Reading material about polygon offset:
[...]

PR #4662, opened 2022-10-18
[3D] add user setting to enable logarithmic depth buffer

jtbandes:
**User-Facing Changes**
[3D] Added a setting to allow using a [logarithmic depth buffer](https://threejs.org/examples/#webgl_camera_logarithmicdepthbuffer), which can improve rendering quality for certain data.

**Description**
Relates to #4641, #4647

https://user-images.githubusercontent.com/14237/196515291-4e172183-407e-4402-bdff-253e6810682b.mov

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
@@ -257,7 +257,13 @@ class InstancedLineMaterial extends THREE.LineBasicMaterial {
  */
 export class Renderer extends EventEmitter<RendererEvents> {
   private canvas: HTMLCanvasElement;
-  public readonly gl: THREE.WebGLRenderer;
+
+  // eslint-disable-next-line no-restricted-syntax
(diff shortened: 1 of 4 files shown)


jtbandes:
Hmm, why does it degrade the quality of labels? Is that expected or something we're doing wrong with our label rendering?

snosenzo (approved):
works great! the only thing I kinda wish was that the logarithmic depth buffer toggle second or third in the list of camera settings next to perspective. just seems a little tacked on at the bottom, but I know we might not want to be advertising it too much.

jhurliman (changes requested):
This is a good idea, but we need to either disallow raw shader usage (preferable, so we don't run into this with other three.js engine features) or provide helper GLSL so raw shaders can take use a logarithmic depth buffer in three.js.

This is blocked until labels work, but it's a lower priority since we fixed the demo layout z-fighting by changing the near plane.

jtbandes:
Closing since we've found other workarounds to the original issues and this was blocked anyway on https://github.com/foxglove/three-text/issues/63.

PR #4664, opened 2022-10-18
Tweak NuScenes demo layout near plane

james-rms:
This pushes the near plane out to 1 meter, resulting in significantly better Z-fighting behavior for close planes.

Having a near plane at 1m means one only sees clipping at pretty tight distances, as shown here:
Screen Shot 2022-10-19 at 8 05 27 am

This may be inappropriate for applications like robot manipulation, but the setting is still configurable for those situations.
**User-Facing Changes**

**Description**

### DIFF
--- a/packages/studio-base/src/dataSources/SampleNuscenesLayout.json
+++ b/packages/studio-base/src/dataSources/SampleNuscenesLayout.json
@@ -52,7 +52,7 @@
         "targetOrientation": [0, 0, 0, 1],
         "thetaOffset": 45,
         "fovy": 45,
-        "near": 0.01,
+        "near": 1.0,
         "far": 5000
       },
       "scene": {},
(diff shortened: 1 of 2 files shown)


james-rms:
Did a little find-and-replace here, one question I have is - why is the near plane distance defined in `SampleNuScenesLayout` at all if it's the default?

jhurliman (approved):
I think 0.5m would be tight enough for even smaller robots like desktop arms. Can we use that as the default and leave nuScenes at 1.0?

jtbandes:
The full state is serialized in the layout so we probably copy/pasted it in and didn't bother deleting it. Omitting it might also cause the layout to appear "modified" soon after it's loaded even if the user doesn't make changes.

jtbandes (approved):

PR #4671, opened 2022-10-19
Fix Gauge/Indicator skipping messages on high-frequency topics

jtbandes:
**User-Facing Changes**
Fixed a bug in the Gauge and Indicator panels where messages on high-frequency topics would sometimes be skipped when using a message path filter.

**Description**
These panels only processed the _last_ message on each `currentFrame`. That's incorrect, because if the message path contains a filter, one of the intermediate messages in the frame might match the filter even if the last one does not. To fix the issue, we process all messages in the currentFrame on the requested topic. Also added stories with fixture data containing multiple messages per frame to prevent regressions.

Thanks to @2metres for noticing the bug.

### DIFF
--- a/packages/studio-base/src/panels/Gauge/Gauge.tsx
+++ b/packages/studio-base/src/panels/Gauge/Gauge.tsx
@@ -2,6 +2,7 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
+import { last } from "lodash";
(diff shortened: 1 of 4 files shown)


defunctzombie (commented):
Is it right to "dispatch" every message rather than dispatch the `currentFrame` and handle the filtering logic in the reducer? This reads to me like it will now call `dispatch` many times whereas before it only called it once per render.

defunctzombie (approved):

Issue #4626, opened 2022-10-12
Perform value to color conversion in shaders


jtbandes:
The Grid and PointCloud color conversion currently happens on the CPU, requiring a slow, serial iteration over all the field data before uploading to the GPU. Instead, we should perform this conversion from an int/float value to a rgba value on the GPU — this means giving the shaders access to the `colorMode`, `colorMap`, custom `gradient`, and other settings values that are needed to perform the conversion.

The only step that is hard to perform on the GPU is determining the **min & max values** automatically from the data, which requires a full scan of the data. We have discussed that replacing this "auto" feature with a button to **explicitly** grab the min/max values from the latest message (only when clicked) would be a reasonable compromise. (It is technically possible to compute min/max on the GPU, but doing so may be [difficult & complex](https://gist.github.com/jhurliman/997cae2d942ed6b841d63326ed971b4f), or require new APIs in WebGPU, and we believe the per-message auto min/max feature is not terribly important to keep.)

jtbandes:
Moved to Linear

PR #4673, opened 2022-10-19
Shader Performance Boost for Foxglove.Grid Messages

snosenzo:
**User-Facing Changes**
 - performance upgrade for foxglove.Grid messages

**Description**
 - color conversion done in shaders vs on CPU
 - moved functions for updating material and texture onto renderable
  
Things that aren't quite working yet:
 - ~~colors aren't the same as before (can be seen in stories).  I have a feeling it has to do with the srgb/linear colorspace. I'll need some help navigating and debugging it.~~
   - This was solved by applying `LinearTosRGB` in the shader to colors that were being defined by the CPU (RGBA and gradient values). These were applied to maintain existing color behavior.
 - ~~picking is broken. will also need some tips or assistance as to why my current solution doesn't work. I think it might be because non of the uniforms or texture for the picking material aren't updated? But I'm also confused how it was being updated before? What was updating the map texture uniform to the picking texture previously if the size changed?~~
   - picking has been fixed with the `syncPickingMaterial` method and usage which copies the uniforms from the other material

Partially addresses #4626


packages/studio-base/src/panels/ThreeDeeRender/renderables/FoxgloveGrid.ts
```diff
@@ -429,7 +420,10 @@ export class FoxgloveGrid extends SceneExtension<FoxgloveGridRenderable> {
         });
       }
 
-      const texture = createTexture(foxgloveGrid);
+      // Check color
+      const texture = floatTextureColorModes.has(settings.colorMode)
```
defunctzombie, line 415: Fun fact - in our wonderful javascript, a lookup for a string in small arrays is faster than a set lookup.

packages/studio-base/src/panels/ThreeDeeRender/renderables/FoxgloveGrid.ts
```diff
@@ -237,7 +237,7 @@ export class FoxgloveGridRenderable extends Renderable<FoxgloveGridUserData> {
     const maxColor = stringToRgba(tempColor, settings.gradient[1]);
     rgbaToLinear(maxColor, maxColor);
     maxGradientColor.value.set(maxColor.r, maxColor.g, maxColor.b, maxColor.a);
-    // material.needsUpdate = true uneccessary because all uniforms are sent to GPU every frame
+    // material.needsUpdate = true unnecessary because all uniforms are sent to GPU every frame
```
defunctzombie, line 240:
```suggestion
```

We should not leave dead code in the codebase without a comment because it isn't clear whether it is accidentally commented out or with some purpose

snosenzo: yeah I didn't intend for it to look like commented out code. just wanted to say material needsUpdate doesn't need to be true and why. Reworded to be a bit more clear

packages/studio-base/src/panels/ThreeDeeRender/renderables/FoxgloveGrid.ts
```diff
@@ -398,10 +537,33 @@ function createTexture(foxgloveGrid: Grid): THREE.DataTexture {
   return texture;
 }
 
+function createFloatTexture(foxgloveGrid: Grid): THREE.DataTexture {
```
jtbandes, line 540: Maybe dedupe these two functions? You could pass in the TypedArrayConstructor.

packages/studio-base/src/panels/ThreeDeeRender/renderables/FoxgloveGrid.ts
```diff
@@ -398,10 +537,33 @@ function createTexture(foxgloveGrid: Grid): THREE.DataTexture {
...
+  );
+  texture.generateMipmaps = false;
+  return texture;
+}
+
 function createMesh(topic: string, texture: THREE.DataTexture): THREE.Mesh {
   // Create the texture, material, and mesh
-  const pickingMaterial = createPickingMaterial(texture);
-  const material = createMaterial(texture, topic);
+  const material = createMaterial(texture, topic) as GridShaderMaterial;
```
jtbandes, line 565: `as` makes me nervous – can the `createMaterial` be changed to return GridShaderMaterial?

packages/studio-base/src/panels/ThreeDeeRender/renderables/FoxgloveGrid.ts
```diff
@@ -42,6 +44,19 @@ const DEFAULT_MIN_COLOR = { r: 100 / 255, g: 47 / 255, b: 105 / 255, a: 1 };
...
+  RGB: 1,
+  RGBA: 2,
+  GRADIENT: 3,
+  COLORMAP: 4,
+};
+
+const COLOR_MAP_TO_GLSL: Record<string, number> = {
+  TURBO: 0,
+  RAINBOW: 1,
+};
```
jtbandes, line 58: I'd prefix these keys just to aid clarity and avoid potential name conflicts with other shader defines.

```suggestion
const COLOR_MODE_TO_GLSL: Record<string, number> = {
  COLOR_MODE_FLAT: 0,
  COLOR_MODE_RGB: 1,
  COLOR_MODE_RGBA: 2,
  COLOR_MODE_GRADIENT: 3,
  COLOR_MODE_COLORMAP: 4,
};

const COLOR_MAP_TO_GLSL: Record<string, number> = {
  COLOR_MAP_TURBO: 0,
  COLOR_MAP_RAINBOW: 1,
};
```

packages/studio-base/src/panels/ThreeDeeRender/renderables/FoxgloveGrid.ts
```diff
@@ -56,16 +71,41 @@ const DEFAULT_SETTINGS: LayerSettingsFoxgloveGrid = {
...
   foxgloveGrid: Grid;
   mesh: THREE.Mesh;
   texture: THREE.DataTexture;
-  material: THREE.MeshBasicMaterial;
+  material: GridShaderMaterial;
   pickingMaterial: THREE.ShaderMaterial;
 };
 
+const tempFieldReader = {
+  fieldReader: zeroReader as FieldReader,
```
jtbandes, line 105: avoid `as` since it disables some TS errors, prefer either changing the definition of zeroReader or putting the annotation on tempFieldReader e.g. `const tempFieldReader: {fieldReader: FieldReader} = ...`

packages/studio-base/src/panels/ThreeDeeRender/renderables/FoxgloveGrid.ts
```diff
@@ -76,12 +116,208 @@ export class FoxgloveGridRenderable extends Renderable<FoxgloveGridUserData> {
...
   }
+
+  public syncPickingMaterial(): void {
+    const { pickingMaterial, material } = this.userData;
+    pickingMaterial.uniforms = material.uniforms;
+    pickingMaterial.needsUpdate = true;
+  }
+
+  private _getFieldReader(
+    output: { fieldReader: FieldReader },
```
jtbandes, line 127: why return output via an object + boolean rather than just returning `FieldReader | undefined`?

packages/studio-base/src/panels/ThreeDeeRender/renderables/FoxgloveGrid.ts
jtbandes: Furthermore if we are relying on the fact that each of these corresponds to one of `ColorModeSettings["colorMode"]`, then we should leverage the type system to enforce it. For example, ``const COLOR_MODE_TO_GLSL: { [K in ColorModeSettings["colorMode"] as `COLOR_MODE_${Uppercase<K>}`]: number }``

packages/studio-base/src/panels/ThreeDeeRender/renderables/FoxgloveGrid.ts
```diff
@@ -429,43 +591,144 @@ function numericTypeWidth(type: NumericType): number {
...
-          discard;
+        if(colorMode == RGBA || colorMode == RGB || colorMode == FLAT) {
+          // colors that are coming from CPU need to be converted
+          gl_FragColor = LinearTosRGB(color);
+        } else {
+          float delta = max(maxValue - minValue, 0.00001);
+          float colorValue = color.r;
+          float normalizedColorValue = (colorValue - minValue) / delta;
+          if(colorMode == GRADIENT) {
+            // weight each color by its alpha
```
jtbandes, line 695: Maybe copy the comment from here to keep some context for why this is being done?

https://github.com/foxglove/studio/blob/3089eadac185f934ea6bc37c890fe11080cefd5b/packages/studio-base/src/panels/ThreeDeeRender/color.ts#L87-L89

packages/studio-base/src/panels/ThreeDeeRender/renderables/FoxgloveGrid.ts
```diff
@@ -429,43 +591,144 @@ function numericTypeWidth(type: NumericType): number {
...
+      uniform float colorMapOpacity;
+
+      uniform vec4 minGradientColor;
+      uniform vec4 maxGradientColor;
+
       varying vec2 vUv;
+
+      // fifth-order polynomial approximation of Turbo based on:
+      // https://observablehq.com/@mbostock/turbo
+      vec3 turbo(float x) {
```
jtbandes, line 643: Maybe use this instead since we already have it in the old3d panel? https://github.com/foxglove/studio/blob/main/packages/studio-base/src/panels/ThreeDimensionalViz/commands/PointClouds/index.tsx#L132

packages/studio-base/src/panels/ThreeDeeRender/renderables/FoxgloveGrid.ts
```diff
@@ -429,43 +591,144 @@ function numericTypeWidth(type: NumericType): number {
...
+        float r = 0.1357 + x * ( 4.5974 - x * ( 42.3277 - x * ( 130.5887 - x * ( 150.5666 - x * 58.1375 ))));
+        float g = 0.0914 + x * ( 2.1856 + x * ( 4.8052 - x * ( 14.0195 - x * ( 4.2109 + x * 2.7747 ))));
+        float b = 0.1067 + x * ( 12.5925 - x * ( 60.1097 - x * ( 109.0745 - x * ( 88.5066 - x * 26.8183 ))));
+        return vec3(r,g,b);
+      }
+
+      vec3 rainbow(float pct) {
```
jtbandes, line 650: comment on where this came from? I see the existing version we have in the old3d panel links to http://docs.ros.org/jade/api/rviz/html/c++/point__cloud__transformers_8cpp_source.html

packages/studio-base/src/panels/ThreeDeeRender/renderables/FoxgloveGrid.ts
```diff
@@ -429,43 +591,144 @@ function numericTypeWidth(type: NumericType): number {
...
+        } else {
+          float delta = max(maxValue - minValue, 0.00001);
+          float colorValue = color.r;
+          float normalizedColorValue = (colorValue - minValue) / delta;
+          if(colorMode == GRADIENT) {
+            // weight each color by its alpha
+            vec4 weightedMinColor = vec4(minGradientColor.rgb * minGradientColor.a, minGradientColor.a);
+            vec4 weightedMaxColor = vec4(maxGradientColor.rgb * maxGradientColor.a, maxGradientColor.a);
+            vec4 finalColor = mix(weightedMinColor, weightedMaxColor, normalizedColorValue);
+            // gradient computation takes place in linear colorspace and must be translated to sRGB
```
jtbandes, line 699: might help to name the uniforms `minGradientColorLinear` etc. to make it easier to read and understand which variables are linear and which are srgb.

packages/studio-base/src/panels/ThreeDeeRender/renderables/FoxgloveGrid.ts
```diff
@@ -76,12 +116,208 @@ export class FoxgloveGridRenderable extends Renderable<FoxgloveGridUserData> {
...
+      this.userData.texture = texture;
+      this.userData.material.uniforms.map.value = texture;
+    }
+
+    if (floatMode) {
+      // FLOAT texture handling
+      // type of image.data is Uint8ClampedArray, but it is in fact the raw texture data even thought it's
+      // meant to be used as an RGBA image data array
+      const valueData = texture.image.data as unknown as Float32Array;
+      const r32fView = new DataView(valueData.buffer, valueData.byteOffset, valueData.byteLength);
```
jtbandes, line 277: What does `r32f` mean?

Why use DataView instead of just updating `valueData[i] = colorValue`?

jtbandes (approved):
@jhurliman could you please double-check the linear/srgb conversions?

packages/studio-base/src/panels/ThreeDeeRender/renderables/FoxgloveGrid.ts
snosenzo: yeah I wanted it to be type enforced but didn't know how to do it properly. Thanks!
good point on the collisions, will update

packages/studio-base/src/panels/ThreeDeeRender/renderables/FoxgloveGrid.ts
snosenzo: r32f is the format and type of the texture data. RedFormat FloatType.

I wanted to make sure that the number was properly converted to float32 before being written to the buffer. Is that not a concern when just setting the valuedata?

jtbandes: if the array is actually a Float32Array, then using `[]` to set the value should be fine

packages/studio-base/src/panels/ThreeDeeRender/renderables/FoxgloveGrid.ts
snosenzo: :/ .toUpperCase doesn't keep the base string literal type. would it be alright to use the workaround outlined here to change the global string interface for uppercase and lowercase to maintain these types through this method? https://github.com/microsoft/TypeScript/issues/44268

snosenzo:
I just don't want to lose track of this, but we can further improve the speed of RGB(A) colors by potentially a significant margin by doing something like this: https://web.archive.org/web/20190117084110/https://jsperf.com/canvas-pixel-manipulation
which I found here: https://hacks.mozilla.org/2011/12/faster-canvas-pixel-manipulation-with-typed-arrays/

packages/studio-base/src/panels/ThreeDeeRender/renderables/FoxgloveGrid.ts
snosenzo: will float64 numbers be automatically converted if I don’t use the setFloat32 on the data view and instead set the index on the typed array directly? is that something I should account for?

packages/studio-base/src/panels/ThreeDeeRender/renderables/FoxgloveGrid.ts
snosenzo: I don't think `TypedArrayConstructor` is enough because the RGBA values need 4 times the size of the type of array. I'll pass the `colorModeTextureType` and use that

packages/studio-base/src/panels/ThreeDeeRender/renderables/FoxgloveGrid.ts
snosenzo: ah yeah I adapted it from `colors.ts` and forgot this attribution, thanks

packages/studio-base/src/panels/ThreeDeeRender/renderables/FoxgloveGrid.ts
snosenzo: ah yeah I took this is a vestige of taking this logic from the points clouds that I forgot to clean up previously. thanks for catching

PR #4675, opened 2022-10-20
Fix velodyne open dialog

defunctzombie:
**User-Facing Changes**
Fix opening a connection to a velodyne in the desktop app.

**Description**
The VelodyneDataSourceFactory incorrectly queried for the _portStr_ parameter from the form config. The actual id of the parameter is _port_. This fixes the parameter lookup and allows the player to be built.

### DIFF
--- a/packages/studio-base/src/dataSources/VelodyneDataSourceFactory.ts
+++ b/packages/studio-base/src/dataSources/VelodyneDataSourceFactory.ts
@@ -23,7 +23,7 @@ class VelodyneDataSourceFactory implements IDataSourceFactory {
   };
 
   public initialize(args: DataSourceFactoryInitializeArgs): Player | undefined {
-    const portStr = args.params?.portStr;
+    const portStr = args.params?.port;
     if (portStr == undefined) {
       return;
     }


jhurliman (approved):

packages/studio-base/src/dataSources/VelodyneDataSourceFactory.ts
```diff
@@ -23,7 +23,7 @@ class VelodyneDataSourceFactory implements IDataSourceFactory {
   };
 
   public initialize(args: DataSourceFactoryInitializeArgs): Player | undefined {
-    const portStr = args.params?.portStr;
+    const portStr = args.params?.port;
```
jhurliman, line 26: Needs more TypeScript

jtbandes: +1

defunctzombie: Where would I put more typescript? Params is a generic key/value object from the form fields. Guess we could add some type to the overall factory and the form fields and initialize both have to adhere to that.

jtbandes: Some trips down memory lane:

https://github.com/foxglove/studio/pull/1513/files#diff-a1f2844d25ced6ce2c9d64210f38a6d68f8309729796738b54553cb79f3f1a92L47-L56

https://github.com/foxglove/studio/pull/1513/files#diff-9c3683eeee282330415f8ecd494d9f4b74e67f88f140829fe905876dd25b9ab7R30

defunctzombie: I do not want to go back to this. If I move towards a direction where we can provide these as extensions this level of type setup is not available.

The schema validation would be a good step. Another thing that would have helped here is throwing in the return - right now before this fix the dialog would hang open with no further debug info.

Issue #4615, opened 2022-10-11
Style uplift on Studio Readme


2metres:
Feedback from @amacneil:
> Really nice readme hero image, badges and links. we could do something similar for mcap + studio repos: https://github.com/prisma/prisma#readme

<img width="934" alt="Screen Shot 2022-10-12 at 7 38 46 am" src="https://user-images.githubusercontent.com/924528/195193690-0b541b3b-d9b9-4c1d-969b-2386350378b8.png">

2metres:
Might be worth reconsidering the content of the Readme at this momemnt as well

amacneil:
Don't overthink it, just a nice header image, centered title, and links is a great improvement.

ocin-rye:
Uploading header image here to reference GitHub image URL in the readme file.

![readme-header](https://user-images.githubusercontent.com/14011012/195717422-bfcc5a12-0281-4f30-b310-9322f6fdd660.png)

PR #4676, opened 2022-10-20
Update repo header with markdown image to resolve visibility issue

ocin-rye:
**User-Facing Changes**

Increases consistency of header image visibility.

**Description**

HTML  header image replaced markdown image syntax to resolve inconsistent image visibility.

Addresses image quirk introduced with https://github.com/foxglove/studio/issues/4615 - Style uplift on Studio Readme


jtbandes:
For future reference it seems like this was caused by a GitHub bug. When the user's theme is set to "single theme" mode rather than "sync with system", they have some theming code that stomps on the `media` attribute we chose for our `<source>` elements. I couldn't find a consistent way to work around it.

PR #4682, opened 2022-10-20
Improve performance of state transition panel

foxymiles:
**User-Facing Changes**
Improves performance of the state transition panel.

**Description**
Currently we map each state transition to a new dataset. For data with a large number of transitions this results in very bad interactive performance in ChartJS. The fix here is to use a different dataset for each distinct value in the data instead of for each transition and to create gaps at each state transition point.

Longer term we will probably want to rework this to put all points in a single dataset and use the ChartJS per-segment coloring features to delineate changes but this is difficult in the current architecture because the chart config needs to be serializable to pass to the worker and this precludes passing a function for the `segments` parameter:

https://www.chartjs.org/docs/latest/samples/line/segments.html

### DIFF
--- a/packages/studio-base/src/panels/StateTransitions/messagesToDatasets.ts
+++ b/packages/studio-base/src/panels/StateTransitions/messagesToDatasets.ts
@@ -2,6 +2,7 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
+import { ScatterDataPoint } from "chart.js";
 import stringHash from "string-hash";
 
 import { Time, toSec, subtract as subtractTimes } from "@foxglove/rostime";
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/panels/StateTransitions/messagesToDatasets.ts
```diff
@@ -31,21 +35,23 @@ type Args = {
 };
 
 export default function messagesToDatasets(args: Args): DatasetInfo {
-  const { path, startTime, y, pathIndex, blocks } = args;
+  const { path, startTime, y, blocks } = args;
```
jtbandes, line 38: should pathIndex be removed from the Args typedef?

jtbandes (approved):

packages/studio-base/src/panels/StateTransitions/messagesToDatasets.ts
foxymiles: Actually this is incorrect. The `label` later should still be `pathIndex`.

Issue #4483, opened 2022-09-26
Add ARGB (RViz-compatible) point cloud byte order option


defunctzombie:
Add an ARGB byte order option to point clouds in the 3d Panel.

<img width="420" alt="image" src="https://user-images.githubusercontent.com/84792/192318419-88627d8c-b830-44c7-939b-6c850fb3f042.png">

defunctzombie:
https://github.com/foxglove/studio/blob/58455bbe367d9e2b8067082e0178312916d2824a/packages/studio-base/src/panels/ThreeDeeRender/renderables/pointClouds/colors.ts#L81

jtbandes:
For reference, this ARGB mode seems to be what rviz does by default:

https://github.com/ros-visualization/rviz/blob/a60b334fd10785a6b74893189fcebbd419d468e4/src/rviz/default_plugin/point_cloud_transformers.cpp#L400-L404

https://github.com/ros-visualization/rviz/blob/6a234b4404c5ee71051f0b338ab8d5d2d931c27b/src/rviz/default_plugin/point_cloud_common.cpp#L189-L196

defunctzombie:
@jtbandes I am changing this issue to a bug. It seems we overlooked porting this mode to the new 3d panel and have regressed this functionality.

Below is the dialog from the legacy 3d panel point cloud settings:

<img width="452" alt="image" src="https://user-images.githubusercontent.com/84792/197293391-18be820d-d15c-45f3-a617-4213fe5ab991.png">

jtbandes:
Looks like we had 3 modes in the old 3d panel too, and they were named in opposite ways. I guess when they were ported over, we ended up with a different set of 3 modes and lost the rviz compatible mode.

This is a screenshot with old3d in "RGBA" mode and new3d in "ABGR" mode, and they match.

<img width="719" alt="image" src="https://user-images.githubusercontent.com/14237/197295919-e8a294c1-8325-422d-8e8e-a7bb20da012a.png">

jtbandes:
There is a branch to implement this 4th option: https://github.com/foxglove/studio/compare/aschimmichanga/issue4483. It will need to be rebased on top of #4687 before it can merge.

defunctzombie:
@jtbandes I've re-assigned this to you. Let's get this regression fixed asap.

PR #4687, opened 2022-10-21
[3D] Rename "RGB byte order" setting to "RGB wire format", clarify dropdown options

jtbandes:
**User-Facing Changes**
Improved the clarity of RGB byte order settings for point clouds and grids in the 3D panel.

**Description**
Related: #4483

This replaces `rgbByteOrder` with `rgbWireFormat`. The dropdown labels and raw config values have been byte-swapped, so what used to be called "RGBA" and stored in the config as `{ rgbByteOrder: "rgba" }` is now called "ABGR" and stored in the config as `{ rgbWireFormat: "abgr" }`.

This corrects a misleading inconsistency where the setting was called "byte order", but the order of bytes in the pointcloud/grid `data` was actually the opposite from what was shown in the settings UI and stored in the config.

Previously, the "RGB byte order" setting was interpreted by Studio as though it actually referred to the order of color components in the human-readable **positional notation** (where the most significant digits are on the left). That is, a setting of "RGBA" meant that the 32-bit numeric value of 0x11223344 (287454020 in base 10) would be interpreted as red=0x11, green=0x22, blue=0x33, alpha=0x44. Where this becomes confusing is that Studio also requires 32-bit numeric values to be encoded with **little-endian byte order**. This means that the actual order of bytes in the `data` would be equivalent to `new Uint8Array([0x44, 0x33, 0x22, 0x11])`.

To address this confusion, we replace "byte order" with "wire format" to be more clear that we are talking about the encoded bytes in the buffer. We also swap the labels of the options, so that the option labeled "RGBA" option now actually refers to a wire format of `new Uint8Array([0xRR, 0xGG, 0xBB, 0xAA])` (which is equivalent to `new Uint32Array([0xAABBGGRR])` on a little-endian system).

### Proposed docs update
This is a proposal for some additions to the 3D panel documentation to explain packed RGB support and help users work through these inconsistencies:

----
Foxglove Studio allows PointClouds and Grids to contain embedded color information. Rather than using a numeric field (for example, the `intensity` of a lidar return), and choosing point or cell colors based on the range of those `intensity` values, the RGB or RGBA values can be **packed** into a single 4-byte field, where each byte represents one channel (red, green, blue, or alpha).

The values may be packed into any 4-byte field supported by [PackedElementField](https://foxglove.dev/docs/studio/messages/packed-element-field), but `uint32` is recommended. To color a point cloud or grid using packed RGB data, choose the **RGB** or **RGBA** options in the "**Color mode**" dropdown. (RGBA reads all four channels from the packed data, while RGB ignores the packed alpha channel and allows you to configure an explicit alpha value in the topic settings.)

Different robotics frameworks or custom software stacks may choose to pack the red, green, blue, and alpha values into byte buffers in different ways. To accommodate the most common encodings, the 3D panel also provides a topic setting called "**RGB wire format**". This setting tells Studio what to expect when it reads the packed 4-byte field. The setting values refer to the **order of bytes "on the wire"** (meaning in the `data` buffer, or in your computer's memory). So, if the wire format is set to **ABGR**, then the first byte in the buffer would represent the alpha value, and the fourth would represent the red value.

> Note: when writing code to pack your own RGBA values, be aware that your computer likely represents 32-bit numbers in **little-endian byte order** (least significant bytes come first in memory). This means that a 32-bit numeric value of `0x12345678` (305419896 in base 10) in your source code would actually be encoded in memory as the 4-byte sequence `[0x78, 0x56, 0x34, 0x12]`.

The following table demonstrates Foxglove Studio's different wire format settings and how they map to common code patterns.

"RGB wire format" setting | Bytes in memory | 32-bit unsigned integer value (little-endian)
--- | --- | ---
RGBA | <pre>buffer[offset+0] = 0xRR<br>buffer[offset+1] = 0xGG<br>buffer[offset+2] = 0xBB<br>buffer[offset+3] = 0xAA</pre> | `0xAABBGGRR` or <code>(0xAA << 24) \| (0xBB << 16) \| (0xGG << 8) \| 0xRR</code>
ARGB | <pre>buffer[offset+0] = 0xAA<br>buffer[offset+1] = 0xRR<br>buffer[offset+2] = 0xGG<br>buffer[offset+3] = 0xBB</pre> | `0xBBGGRRAA` or <code>(0xBB << 24) \| (0xGG << 16) \| (0xRR << 8) \| 0xAA</code>
ABGR | <pre>buffer[offset+0] = 0xAA<br>buffer[offset+1] = 0xBB<br>buffer[offset+2] = 0xGG<br>buffer[offset+3] = 0xRR</pre> | `0xRRGGBBAA` or <code>(0xRR << 24) \| (0xGG << 16) \| (0xBB << 8) \| 0xAA</code>

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/FoxgloveGrid.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/FoxgloveGrid.ts
@@ -14,6 +14,8 @@ import {
   getColorConverter,
   autoSelectColorField,
   NEEDS_MIN_MAX,
+  DEFAULT_RGB_WIRE_FORMAT,
+  getRgbWireFormat,
 } from "@foxglove/studio-base/panels/ThreeDeeRender/renderables/pointClouds/colors";
 import type { RosValue } from "@foxglove/studio-base/players/types";
 
@@ -40,7 +42,6 @@ const DEFAULT_COLOR_MAP = "turbo";
 const DEFAULT_FLAT_COLOR = { r: 1, g: 1, b: 1, a: 1 };
 const DEFAULT_MIN_COLOR = { r: 100 / 255, g: 47 / 255, b: 105 / 255, a: 1 };
 const DEFAULT_MAX_COLOR = { r: 227 / 255, g: 177 / 255, b: 135 / 255, a: 1 };
-const DEFAULT_RGB_BYTE_ORDER = "rgba";
 
 const DEFAULT_SETTINGS: LayerSettingsFoxgloveGrid = {
   visible: false,
@@ -53,8 +54,19 @@ const DEFAULT_SETTINGS: LayerSettingsFoxgloveGrid = {
   gradient: [rgbaToCssString(DEFAULT_MIN_COLOR), rgbaToCssString(DEFAULT_MAX_COLOR)],
   colorMap: DEFAULT_COLOR_MAP,
   explicitAlpha: 1,
-  rgbByteOrder: DEFAULT_RGB_BYTE_ORDER,
+  rgbWireFormat: DEFAULT_RGB_WIRE_FORMAT,
 };
+
+function mergeSettingsWithDefaults(
+  settings: Partial<LayerSettingsFoxgloveGrid> | undefined,
+): LayerSettingsFoxgloveGrid {
+  return {
+    ...DEFAULT_SETTINGS,
+    ...settings,
+    rgbByteOrder: undefined,
+    rgbWireFormat: settings ? getRgbWireFormat(settings) : DEFAULT_SETTINGS.rgbWireFormat,
+  };
+}
 
 export type FoxgloveGridUserData = BaseUserData & {
   settings: LayerSettingsFoxgloveGrid;
@@ -138,7 +150,7 @@ export class FoxgloveGrid extends SceneExtension<FoxgloveGridRenderable> {
       const settings = this.renderer.config.topics[topicName] as
         | Partial<LayerSettingsFoxgloveGrid>
         | undefined;
-      renderable.userData.settings = { ...DEFAULT_SETTINGS, ...settings };
+      renderable.userData.settings = mergeSettingsWithDefaults(settings);
 
       this._updateFoxgloveGridRenderable(
         renderable,
@@ -160,7 +172,7 @@ export class FoxgloveGrid extends SceneExtension<FoxgloveGridRenderable> {
       const userSettings = this.renderer.config.topics[topic] as
         | Partial<LayerSettingsFoxgloveGrid>
         | undefined;
-      const settings = { ...DEFAULT_SETTINGS, ...userSettings };
+      const settings = mergeSettingsWithDefaults(userSettings);
       if (settings.colorField == undefined) {
         autoSelectColorField(settings, foxgloveGrid.fields);
         // Update user settings with the newly selected color field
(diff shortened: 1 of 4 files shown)


defunctzombie:
For additional reference, here is what the legacy 3d panel point settings dialog displayed:

image

@amacneil raised an interesting question which I did not know the answer

> Can we auto-select the rviz compatible wire format when we see the user has a `PointCloud2` message? Users with ROS messages are coming from a world where they previously used RVIZ and things "just worked" with their message since they would have had to pack the rgb color field in an rviz compatible way.

jtbandes:
> Can we auto-select the rviz compatible wire format when we see the user has a `PointCloud2` message?

Once #4483 is implemented, that should be possible. We do have access to the schema name in the place where we're auto-selecting the color settings.

defunctzombie:
Who implemented the point cloud settings in the new 3d panel? @jhurliman ?

Our settings labels do not match the legacy settings dialog (they are flipped around) and we should identify what happened here or why we changed them from what we had previously.

jtbandes:
Cross-posting https://github.com/foxglove/studio/issues/4483#issuecomment-1287485417 because there are too many threads going on.

> Looks like we had 3 modes in the old 3d panel too, and they were named in opposite ways. I guess when they were ported over (#3069), we ended up with a **different** set of 3 modes and lost the rviz compatible mode. 
> 
> This is a screenshot with old3d in "RGBA" mode and new3d in "ABGR" mode, and they match.
> 
> image
>

amacneil:
My thoughts from product review:

Foxglove `foxglove.PointCloud`:
* Delete this feature (configurable byte order) entirely, require that the schema contains fields named `red`, `green`, `blue`, `alpha` if you want to use colored point clouds. Might need to hand hold some existing customers to migrate.

ROS `sensor_msgs/PointCloud2`
* We should just always use the same byte order that RViz expects, and not support configuration

Need feedback from @defunctzombie @jhurliman

jhurliman:
We can do whatever we want with `foxglove.PointCloud` since no (public) codebases are publishing that message type today. For `sensor_msgs/PointCloud2` there is an existing ecosystem of ROS driver nodes and `.bag` files in the wild and there is no universal agreement on byte ordering for colored point clouds. Configurable byte ordering is required for rviz parity.

jtbandes:
> there is no universal agreement on byte ordering for colored point clouds. Configurable byte ordering is required for rviz parity.

I don't think that's right... RViz has a hard-coded byte ordering and that's about as close to "agreement" as I could imagine it getting.

defunctzombie:
> We can do whatever we want with `foxglove.PointCloud` since no (public) codebases are publishing that message type today.

We have customers using this type today.

> For `sensor_msgs/PointCloud2` there is an existing ecosystem of ROS driver nodes and `.bag` files in the wild and there is no universal agreement on byte ordering for colored point clouds. Configurable byte ordering is required for rviz parity.

Agreed.

---
I like what is being proposed in this PR. `wire` is a better word for what we are describing and removes some of the ambiguity. Coupled with improvements to the documentation (also proposed in the PR) we are in a place where I can link a user to a docs page and they can understand what code to write to produce a point cloud.

I am in favor of moving forward with the changes proposed.

defunctzombie (approved):

amacneil:
We looked at rviz today and could not find any such configurable byte order. I think we are adding complexity to our app for a feature that is not needed by anyone (the current "three dropdowns" approach to coloring my point cloud is unnecessarily complicated).

Restating my proposal:
- `sensor_msgs/PointCloud2` should _always_ use the byte order encoding that rviz does
- `foxglove.PointCloud` should require authors to explicitly specify the byte order using named `PackedElementField`s. We can work with the very small number of existing users on a migration plan. I do not understand why we created our own schema with with ambiguous data order during write, then require the data reader/viewer to remember and configure the correct order every time they select a topic. We have already had customers confused by this and require hand holding to get the correct byte order.

This approach would require decoupling how we display settings between `sensor_msgs/PointCloud2` vs `foxglove.PointCloud`, which I think is sensible in this instance since our own schema can be more explicit.

jtbandes:
> since our own schema can be more explicit

To be clear it doesn’t sound like we want to change anything about the schema. We are just changing conventions for how studio handles certain fields. On another note we still need to find the right place to document these conventions.

Issue #4690, opened 2022-10-21
Studio uses incorrect `profile` when streaming from Data Platform


jtbandes:
When streaming ROS 1 data from Data Platform, Studio does not use the `ros1` profile, so some features do not work the same way they do when dropping a .bag file into the app.

jtbandes:
Temp workaround: https://github.com/foxglove/studio/pull/4691

PR #4691, opened 2022-10-21
Use ros1 profile if all topics from Data Platform use ros1 encoding

jtbandes:
**User-Facing Changes**
Improved handling of TF frame names in ROS 1 data stored in Data Platform.

**Description**
This is a temporary workaround for #4690. A more appropriate fix would be to directly get the `profile` info from Data Platform, but that will require more extensive changes in both DP and Studio.

### DIFF
--- a/packages/studio-base/src/players/IterablePlayer/foxglove-data-platform/DataPlatformIterableSource.ts
+++ b/packages/studio-base/src/players/IterablePlayer/foxglove-data-platform/DataPlatformIterableSource.ts
@@ -186,6 +186,17 @@ export class DataPlatformIterableSource implements IIterableSource {
       }
     }
 
+    let profile: string | undefined;
+
+    // Workaround for https://github.com/foxglove/studio/issues/4690.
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/players/IterablePlayer/foxglove-data-platform/DataPlatformIterableSource.ts
```diff
@@ -186,14 +186,25 @@ export class DataPlatformIterableSource implements IIterableSource {
...
+    let profile: string | undefined;
+
+    // Workaround for https://github.com/foxglove/studio/issues/4690.
+    // If all topics use ros1 schemas and message encodings, assume we are working with the ros1 profile data.
+    if (
+      rawTopics.length > 0 &&
+      rawTopics.every((topic) => topic.encoding === "ros1" && topic.schemaEncoding === "ros1msg")
+    ) {
+      profile = "ros1";
+    }
```
jhurliman, line 198: We could do the same for ros2msg || ros2idl here

PR #4693, opened 2022-10-21
Convert log panel from fluent list to react-window

foxymiles:
**User-Facing Changes**
None

**Description**
Convert the log panel from the fluentui list component to react-window. In my testing this seems to have comparable performance and interaction behavior to our existing fluentui implementation.

I wasn't able to get react-virtualized to work in this case so maybe we should switch the DiagnosticSummary panel to use react-window as well and then we can drop react-virtualized as a dependency.

Fixes https://github.com/foxglove/studio/issues/4079


jtbandes:
Looks like the layout is not being updated properly when resizing the panel if messages aren't actively coming in.

https://user-images.githubusercontent.com/14237/197283000-e5440148-f73d-4aff-ab5e-c13ae997a3ac.mov

jtbandes:
Do you have a good example file to test with?

foxymiles:
I've been using this:

https://github.com/foxglove/ros1-bench-workspace

With the sinewave.py script tweaked to emit log messages of different sizes:

```python
#!/usr/bin/env python

import math
import queue
import time
import rospy
from std_msgs.msg import Float32, String


def main():
    rospy.init_node("sinwave", anonymous=True)

    pub1 = rospy.Publisher("/sample1", Float32, queue_size=100)
    pub2 = rospy.Publisher("/sample2", Float32, queue_size=100)
    pub3 = rospy.Publisher("/chatter", String, queue_size=100)

    rate = rospy.Rate(10)  # Hz
    count = 0
    while not rospy.is_shutdown():
        val = math.sin(time.time())
        pub1.publish(val)
        if count % 20 == 0:
            pub2.publish(val * 0.5)
        if count % 2 == 0:
            rospy.loginfo(f"hello world {count}")
        else:
            rospy.loginfo(f"Long\nmessage\nneeds\nwrapping {count}")
        rate.sleep()
        count += 1


if __name__ == "__main__":
    try:
        main()
    except rospy.ROSInterruptException:
        pass
```

foxymiles:
> Looks like the layout is not being updated properly when resizing the panel if messages aren't actively coming in.

Good catch. I've pushed a small change that seems to handle this case.

packages/studio-base/src/panels/Log/LogList.tsx
```diff
@@ -133,7 +133,7 @@ function LogList({ items }: Props): JSX.Element {
     <AutoSizer>
       {({ width, height }) => {
         return (
-          <div style={{ position: "relative", width, height }}>
+          <div style={{ position: "relative", width, height }} key={width}>
```
jtbandes, line 136: Seems aggressive to completely remount when the width changes. Is there a way to tell react-window to recompute the layout?

foxymiles: In theory clearing the item size cache and calling `resetFromIndex(0)` should do it but it's not working for me.

jtbandes: Not sure if this is helpful but it does sound like react-window is missing some support for dynamic resizing. https://github.com/bvaughn/react-window/issues/6

foxymiles: Yeah the libraries that do claim to support it have failed for me. 

I've pushed another fix that seems to do better.

jtbandes (approved):

defunctzombie:
Things different in the log panel behavior in this PR vs main:

- Log entries start at the top of the panel and grow down until reaching the bottom, then the old ones are pushed out of the top. On `main` the entries start at the bottom and grow up. I think the behavior on this PR matches other log apps and terminal behavior so this change seems 👍 
- When I scroll down manually to the bottom the "scroll to bottom" icon is still present. I have to manually click it. On `main` I can scroll to the bottom and this puts be panel back into "follow logs" mode.

I don't think the scroll to bottom behavior has to block this PR but it is worth seeing if it is easy to replicate before merging this change.

jtbandes:
> On `main` the entries start at the bottom and grow up. I think the behavior on this PR matches other log apps and terminal behavior so this change seems 👍

You did this in #3189. If you have more context on why that change was made and whether it's important, please share.

defunctzombie:
> You did this in #3189. If you have more context on why that change was made and whether it's important, please share.

My only recollection is that I probably thought it was nice to see the latest entry consistently at the bottom of the panel rather than only at the bottom when the scroll buffer filled up. In retrospect I can't think of many logging tools that do this - most start at the top and grow down.

foxymiles:
@defunctzombie I spent a lot of time trying and failing to replicate the scroll from bottom behavior with this and other libraries.

I think we can easily address your second point with a little extra logic in the `onScroll` handler.

defunctzombie:
> @defunctzombie I spent a lot of time trying and failing to replicate the scroll from bottom behavior with this and other libraries.

👍 let's ignore it then. I don't have a strong justification for why I did that beyond what I recalled above and I think the new behavior is fine too.

jtbandes:
This context from the old PR seems relevant:

> Setting scrollTop to the scrollHeight causes a focus loss for open context menus. This change fixes the behavior by using column-reverse flex direction for the log message scroll area. column-reverse tells the browser to treat the bottom of the div as the scroll anchor. As new messages are appended to the div, the old messages scroll off the top.

> Fixes: https://github.com/foxglove/studio/issues/3181

Is this still a concern? Or maybe it was a Fluent issue that we don't need to work around anymore?

foxymiles:
@jtbandes Thanks for bringing that up. I checked and this doesn't seem to be an issue with the panel context menu and the log panel after this change.

Issue #4681, opened 2022-10-20
Parameters panel has strange escaping and cuts off parameters at small sizes.


defunctzombie:
**Description**

Connect to a ROS1 data source and open the parameters panel.

<img width="774" alt="image" src="https://user-images.githubusercontent.com/84792/196957069-2cc53caf-ee53-41a0-97d2-6ce94734c8a2.png">

- All the string entries have weird escaping characters `"\"`. This is unexpected for string parameters.
- All the number parameters are quoted as if they are string. This is unexpected for number parameters.

PR #4701, opened 2022-10-24
Fix value escaping in parameter panel

foxymiles:
**User-Facing Changes**
Fixes an issue with the display of values in the parameter panel.

**Description**
The parameter panel was applying `JSON.stringify` to all parameter values. We should only do this in cases where the value isn't a type we can edit directly.

Fixes https://github.com/foxglove/studio/issues/4681

### DIFF
--- a/packages/studio-base/src/panels/Parameters/index.tsx
+++ b/packages/studio-base/src/panels/Parameters/index.tsx
@@ -185,7 +185,7 @@ function Parameters(): ReactElement {
                   {canSetParams ? (
                     <TableCell padding="none">
                       <JsonInput
-                        dataTestId={`parameter-value-input-${editValue}`}
+                        dataTestId={`parameter-value-input-${displayValue}`}
                         value={editValue}
                         onChange={(newVal) => {
                           setParameter(name, newVal as ParameterValue);
--- a/packages/studio-base/src/panels/Parameters/index.tsx
+++ b/packages/studio-base/src/panels/Parameters/index.tsx
@@ -79,7 +79,7 @@ const useStyles = makeStyles<void, "copyIcon">()((_theme, _params, classes) => (
   },
 }));
 
-function displayValue(value: unknown): string | number | boolean | unknown[] | object {
+function editableValue(value: unknown): string | number | boolean | unknown[] | object {
   if (
     typeof value === "string" ||
     typeof value === "number" ||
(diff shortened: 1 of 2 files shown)


jtbandes (approved):

packages/studio-base/src/panels/Parameters/index.tsx
```diff
@@ -79,6 +79,20 @@ const useStyles = makeStyles<void, "copyIcon">()((_theme, _params, classes) => (
   },
 }));
 
+function editableValue(value: unknown): string | number | boolean | unknown[] | object {
```
defunctzombie, line 82: Please add a comment for this function to explain why we stringify certain values and the desired ux. I can see us wanting to tweak this in some future and it will help to have this context.

packages/studio-base/src/panels/Parameters/index.tsx
```diff
@@ -152,7 +166,8 @@ function Parameters(): ReactElement {
           </TableHead>
           <TableBody>
             {parameterNames.map((name) => {
-              const value = JSON.stringify(parameters.get(name)) ?? "";
+              const displayValue = JSON.stringify(parameters.get(name)) ?? "";
```
defunctzombie, line 169: Why do we stringify the name? This seems weird to me if I was on a framework that has no problems with me using strings like `foo"bar"`, I would see this string as `"foo\"bar\""` in studio?

foxymiles: If we don't stringify things like arrays or objects react throws an error. We could have another function that does this selectively but the extra quotes don't get shown anyway if we're just displaying the value. You can see this in the storybook for the non-editable parameters.

packages/studio-base/src/panels/Parameters/index.tsx
foxymiles: Sorry I already merged this to try to get a quick fix out. Should I do a followup with a comment here?

defunctzombie: 👍 Getting the quick fix out is the right move. Yea it would be nice to get something recorded in the code for this.

packages/studio-base/src/panels/Parameters/index.tsx
defunctzombie: Similar to the value, we should selectively stringify the name if it is an array or object. This allows us to display regular strings without any extra decoration.

PR #4702, opened 2022-10-24
Fix crash in state transition panel on large datasets

foxymiles:
**User-Facing Changes**
Fix crash in the state transition panel on large datasets.

**Description**
Fix a crash in large datasets in the state transition panel. The panel uses `array.push(...data)` to append new items to the accumulated message array but when data is large this overflows the stack since the elements in data have to be pushed onto the stack to make the function call. In my testing this occurs somewhere above 300k elements.

Further work will be necessary to make the state transition panel performant for datesets with this many messages but this will at least address the crash.

### DIFF
--- a/packages/studio-base/src/panels/StateTransitions/index.tsx
+++ b/packages/studio-base/src/panels/StateTransitions/index.tsx
@@ -235,8 +235,8 @@ const StateTransitions = React.memo(function StateTransitions(props: Props) {
   const { datasets, tooltips, minY } = useMemo(() => {
     let outMinY: number | undefined;
 
-    const outTooltips: TimeBasedChartTooltipData[] = [];
-    const outDatasets: ChartData["datasets"] = [];
+    let outTooltips: TimeBasedChartTooltipData[] = [];
+    let outDatasets: ChartData["datasets"] = [];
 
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/panels/StateTransitions/index.tsx
```diff
@@ -264,8 +264,8 @@ const StateTransitions = React.memo(function StateTransitions(props: Props) {
           blocks: blocksForPath,
         });
 
-        outDatasets.push(...newDataSets);
-        outTooltips.push(...newTooltips);
+        outDatasets = outDatasets.concat(newDataSets);
```
jtbandes, line 267: did you try replacing this with `for (const dataset of newDataSets) outDatasets.push(dataset);`? I wonder if it would be even faster.

jtbandes (approved):

packages/studio-base/src/panels/StateTransitions/index.tsx
foxymiles: I didn't experiment with optimizing this bit since we have perf issues here that dwarf any optimizations we can make at this level.

PR #4705, opened 2022-10-25
Fix Cube List markers not updating

### LINKED ISSUE #4699: Visualization_msgs/Marker cube_list not updating correctly
When using a cube_list as visualization_msgs/Marker, the cubes do not get updated correctly in the 3D panel. They keep the position of the first position that is published. When looking at the raw data, the cubes do get different coordinates.

- Version: 1.29.1
- OS: Ubuntu 20.04
- Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native): both native ROS1 and rosbag
- ROS distro (if applicable): Noetic

**Steps To Reproduce**

- Download [this bag](https://drive.google.com/file/d/1QTNDHmvmkDr-Y06nWjY3xnajivH0knNQ/view?usp=sharing)
- Subscribe to the `/conestreamer/ground_truth_track`
- Show this topic in the 3D panel

**Expected Behavior**
Constantly updating positions of the cubes

**Actual Behavior**
Cubes not moving, some are disappearing and coming back in the same position.

snosenzo:
**User-Facing Changes**
 - fix cube list (type = 6) marker primitives not updating

**Description**
 - attributebuffers need `needsUpdate = true` after they're changed to be sent to the GPU again. should also fix similar issue with spheres and other `DynamicInstanceMesh` renderables

Fixes #4699

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/DynamicInstancedMesh.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/DynamicInstancedMesh.ts
@@ -49,6 +49,10 @@ export class DynamicInstancedMesh<
       tempColor.setRGB(color.r, color.g, color.b);
       this.setColorAt(i, tempColor);
     }
+    this.instanceMatrix.needsUpdate = true;
+    if (this.instanceColor) {
+      this.instanceColor.needsUpdate = true;
+    }
   }
 
   private _setCount(count: number) {


snosenzo:
Yeah everything that wasn't a CubeList was working perfectly fine from the example file they provided. And it was clear that the points were getting updated but the renderer just wasn't updated. Looks like CubeList and SphereList use this. I checked the other primitives that use `setMatrixAt` and this appears to be the only place where the `needsUpdate` on the BufferAttributes isn't being set. `Axis.ts` renderable also doesn't do this but its update method is private and only used on initialization (before being passed to GPU).

Issue #4694, opened 2022-10-21
Log panel caches messages from previous data source


jtbandes:
The Log panel keeps some stale data around when switching data sources. In this screen recording, the first bag has log messages, but the second does not. Yet when I drop the second bag, I still see log messages from the first bag. 😱 

https://user-images.githubusercontent.com/14237/197283491-0ad1c939-a4f3-4297-abb8-2ed42f9e8ada.mov

foxymiles:
I think this might be player related. If I log what messages the panel receives from `useMessagesByTopic` during the transition it first receives an empty array, then a fresh array containing one message from the previous datasource.

It also seems as if the message reducer is being called at regular intervals even though the data source is a static file and playback is stopped.

Any ideas @defunctzombie? Is that the behavior you'd expect for a static source?

jtbandes:
> It also seems as if the message reducer is being called at regular intervals even though the data source is a static file and playback is stopped.


I've seen this before in other contexts; I believe this is an issue with the iterable player:

https://github.com/foxglove/studio/blob/cccc3a9eca94b91db146bc7418c6b80892ff006a/packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts#L888-L891

PR #4707, opened 2022-10-25
Don't cache message events from previous player when switching players

jtbandes:
**User-Facing Changes**
Fixed a bug where data from a previous data source would be visible in the Log panel after switching data sources.

**Description**
Fixes https://github.com/foxglove/studio/issues/4694 by clearing `lastMessageEventByTopic` and `messageEventsBySubscriberId` when switching players.

### DIFF
--- a/packages/studio-base/src/components/MessagePipeline/index.test.tsx
+++ b/packages/studio-base/src/components/MessagePipeline/index.test.tsx
@@ -366,6 +366,97 @@ describe("MessagePipelineProvider/useMessagePipeline", () => {
     ]);
   });
 
+  // https://github.com/foxglove/studio/issues/4694
+  it("does not inject the last message when the player changes", async () => {
+    const player = new FakePlayer();
+    const { Hook, Wrapper, setPlayer } = makeTestHook({ player });
+    const { result, rerender } = renderHook(Hook, { wrapper: Wrapper });
+    await doubleAct(
+      async () =>
+        await player.emit({
+          activeData: {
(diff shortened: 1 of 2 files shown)


snosenzo:
makes me wonder what should be maintained from the previous player state when the player switches and why is it anything? Can we add some comments clarifying some of that?

foxymiles (approved):

jtbandes:
> makes me wonder what should be maintained from the previous player state when the player switches and why is it anything? Can we add some comments clarifying some of that?

> Why not just recreate the store?

We were also pondering this while working on the PR. I think recreating the store might be a good idea. Subscriptions are one thing that doesn't get reset, but in reality since we also reset all panels when the player changes, they do effectively get reset shortly after: https://github.com/foxglove/studio/blob/39bbe37e16e95af38814c5c671f7b6e987194b24/packages/studio-base/src/Workspace.tsx#L619-L620

Currently the MessagePipeline doesn't require that this is done because it does keep track of subscriptions between players. But if we had already planned to reset the whole state when it was first written, maybe we would be recreating the store from scratch already.

Another way to reset would be to remount the MessagePipeline itself, but I'm not sure if that might cause React to actually remount the _entire_ app which seems less than ideal (for example, it would lose your selected tab and scroll position in the sidebar):
https://github.com/foxglove/studio/blob/39bbe37e16e95af38814c5c671f7b6e987194b24/packages/studio-base/src/components/PlayerManager.tsx#L330

Issue #4704, opened 2022-10-25
Is internet connection needed during ROS2 initial connection?


felixf4xu:
When I tried to open a ROS2 connection, if the PC has no internet connection, the data source side bar will display in-progress state forever. `ros2 topic list` and `ros2 topic echo` works as expected at that moment.

If the pc has internet connection, it works ok.

- Version: 1.25
- OS: Ubuntu 20
- Data source : ROS 2
- ROS distro (if applicable): Galactic

jtbandes:
Are you using the desktop app or the web app? You will need the desktop app for offline use.

jhurliman:
This is caused by https://github.com/foxglove/rtps/issues/17

felixf4xu:
Nice to know that, I will come back if the walkaround (disabling all other network adapters) works and come back.

jtbandes:
Going forward, we plan to deprecate and remove the ROS 2 native connection (https://github.com/foxglove/studio/issues/4900). We are working on a new [foxglove_bridge](https://github.com/foxglove/ros-foxglove-bridge), which will provide improved performance and reliability over the Rosbridge and native connections. You can already use the new bridge today (see the README). Stay tuned for upcoming announcements about the new bridge!

Issue #4698, opened 2022-10-22
ros2 MarkerArray failed to parse, invalid typed array length


mrkbac:
**Description**

- Version: 1.29.0
- OS: ubuntu 22.04
- Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native): ros2 native
- ROS distro (if applicable): humble

**Steps To Reproduce**

When running the following python code foxglove reports "Error receiving messages on "/markers"

```
Error: Failed to parse 2328 byte message on "rt/markers": Invalid typed array length: 1072693247
    at Participant.<anonymous> (file:///opt/Foxglove%20Studio/resources/app.asar/renderer/main.js:6:115248)
    at EventEmitter.emit (file:///opt/Foxglove%20Studio/resources/app.asar/renderer/main.js:1202:12257)
    at Participant.handleUserData (file:///opt/Foxglove%20Studio/resources/app.asar/renderer/main.js:211:24763)
    at Participant.handleData (file:///opt/Foxglove%20Studio/resources/app.asar/renderer/main.js:211:19783)
    at Participant.handleDataMsg (file:///opt/Foxglove%20Studio/resources/app.asar/renderer/main.js:211:18309)
    at Participant.handleUdpMessage (file:///opt/Foxglove%20Studio/resources/app.asar/renderer/main.js:211:14684)
    at EventEmitter.emit (file:///opt/Foxglove%20Studio/resources/app.asar/renderer/main.js:1202:12314)
    at file:///opt/Foxglove%20Studio/resources/app.asar/renderer/main.js:4749:32078
    at UdpSocketRenderer.messagePort.onmessage (file:///opt/Foxglove%20Studio/resources/app.asar/renderer/main.js:4749:32418)
```

```python3
import sys

import rclpy
from rclpy.node import Node
from visualization_msgs.msg import Marker, MarkerArray

class Visu_Node(Node):
    def __init__(self):
        super().__init__('visu_node')
        self.idx = 0

        self.markers = MarkerArray()
        self.markers.markers.clear()

        for x in range(10):
            m = Marker()
            m.header.stamp = self.get_clock().now().to_msg()
            m.header.frame_id = "/map"
            m.id = self.idx
            m.action = Marker.ADD
            m.type = Marker.CYLINDER

            m.scale.x = 1.0
            m.scale.y = 1.0
            m.scale.z = 1.0

            m.pose.position.x = x * 1.0

            self.markers.markers.append(m)

[...]

Issue #3272, opened 2022-04-27
ROS2 node crash in CycloneDDS when connecting with Foxglove Studio


BorisKontorovich:
**Description**
My ROS2 package crashes when running with FoxGlove

- Version: v1.9.0
- OS: Ubuntu 20.01

**Steps To Reproduce**
1. Connecr 1 Intel RealSense D435 camera
2. Run Intel RealSense ROS2 Wrapper
3. Open Foxglove studio

**Expected Behavior**
Foxglove can display published images

**Actual Behavior**
Intel RealSense ROS2 Wrapper crashes a few seconds after FoxGlove is started.

Error:
[cam-2] python3: /tmp/binarydeb/ros-galactic-cyclonedds-0.8.0/src/core/ddsi/src/ddsi_entity_index.c:411: entidx_lookup_proxy_reader_guid: Assertion `is_reader_entityid (guid->entityid)' failed.
[ERROR] [cam-2]: process has died [pid 3848, exit code -6, cmd '/home/sh/ws/flir-wrapper_ws/install/flir_wrapper/lib/flir_wrapper/cam --ros-args -r __node:=cam_727 -r __ns:=/cam_727 --params-file /tmp/launch_params_5fvheszt'].

defunctzombie:
@BorisKontorovich This looks like an error with your ROS2 node and not an error from Studio. I'd recommend looking into why the node is crashing as this would be specific to the node's logic. Maybe open an issue with the authors of the Intel RealSense ROS2 Wrapper with your above trace.

BorisKontorovich:
Can you give me some ideas how I can try foxglove with a camera?

BorisKontorovich:
Anything?

jtbandes:
Are you able to view raw data with `ros2 topic echo /my_camera_topic`?


@jhurliman Could you look into this as it appears to be DDS related? Is there a chance our RTPS implementation is generating some kind of id wrong, or doing something that triggers a bug in cyclonedds?

Looks like the assertion is here:

https://github.com/eclipse-cyclonedds/cyclonedds/blob/3c81d2856aa088e48d2e9f2365c53c2591ee6fdb/src/core/ddsi/src/ddsi_entity_index.c#L411

And the `is_reader_entityid` function:

https://github.com/eclipse-cyclonedds/cyclonedds/blob/3c81d2856aa088e48d2e9f2365c53c2591ee6fdb/src/core/ddsi/src/q_entity.c#L211-L221

[...]

Issue #2614, opened 2022-01-07
Support reading and setting ROS params in ROS2 native connector


jhurliman:

bertaveira:
+1 would love to see this feature as well

jtbandes:
Going forward, we plan to deprecate and remove the ROS 2 native connection (https://github.com/foxglove/studio/issues/4900). We are working on a new [foxglove_bridge](https://github.com/foxglove/ros-foxglove-bridge), which will provide improved performance and reliability over the Rosbridge and native connections. You can already use the new bridge today (see the README). Stay tuned for upcoming announcements about the new bridge!

PR #4708, opened 2022-10-25
Remove ROS 2 native connector

jhurliman:
**User-Facing Changes**
- The ROS 2 native connector is removed in the desktop version of Studio. Connecting via rosbridge is the preferred option for ROS 2 robots

**Description**
Large message support over DDS/RTPS is significantly more complex than messages that fit in a single UDP datagram since RTPS relies heavily on tuning knobs and heuristics to prevent retry floods and achieve real-time frame rates for image or pointcloud-sized messages. It's unlikely the @foxglove/rtps library will achieve the level of performance needed to reliably receive sensor message streams in the near future, so this PR removes the option from Studio for now.

Addresses #4704, #4698, #3272, #1661, #2614.


jtbandes:
Rather than just removing the feature suddenly, we should probably follow a deprecation and feature flag process like we are doing with the old3d panel.

amacneil:
Not sure whether this was discussed in Slack or this PR is just coming out of the blue. But cc @banisadr to think about because we have quite a few users of ROS 2 native connections.

defunctzombie:
> RTPS relies heavily on tuning knobs and heuristics to prevent retry floods and achieve real-time frame rates for image or pointcloud-sized messages.

@jhurliman could you elaborate on this here so we can understand some of the nuances and gotchas involved in updating our RTPS library to support large messages. Some things I would find helpful are links to code or docs that show the complexity of the knobs or heuristics that we'd have to replicate or the RTPS documentation/spec for sending large messages.

banisadr:
To be clear, you're not saying that the ROS 2 native connection route is a broken one right? Just that it's less performant with large messages (& what constitutes a large message in this case)?

jhurliman:
@banisadr I am saying it is broken. You might see one or two camera frames, then we never receive 100% of the packets to constitute an additional full frame so the connection appears broken while your network is being flooded with packets.

jhurliman:
> > RTPS relies heavily on tuning knobs and heuristics to prevent retry floods and achieve real-time frame rates for image or pointcloud-sized messages.
> 
> @jhurliman could you elaborate on this here so we can understand some of the nuances and gotchas involved in updating our RTPS library to support large messages. Some things I would find helpful are links to code or docs that show the complexity of the knobs or heuristics that we'd have to replicate or the RTPS documentation/spec for sending large messages.

Here is the [official spec](https://www.omg.org/spec/DDSI-RTPS/2.2/PDF) where large message sending is described; see the sections describing `DataFrag`, `HeartbeatFrag`, and `NackFrag`. The description is fairly bare-bones and @foxglove/rtps I think implements what is described in the spec. When receiving a large message, FastDDS will send some percentage of the packets making up a large `sensor_msgs/Image` message but then pause as if it's waiting for ACKs. We are sending ACKs and they do work sometimes, but maybe one in five attempts. I've tried many different patterns of how to coax the sender to deliver more packets, some more successful than others. Still, eventually, it will fall behind and I have not figured out how to get the data flow back on track once a packet has been dropped or we fall behind. I don't have any pointers to FastDDS or CycloneDDS code on how to "make it work" (otherwise I might be further along!).

banisadr:
From discussion today: Plan forward is to hold off on this 🔪 until we have a 1st class bridge which will act as our happy path. We'll 🔪 the native connector once that's available.

defunctzombie:
Closing this PR until it is time to get around to it. The plan is to deprecate in stages with warnings and a feature flag.

PR #4712, opened 2022-10-26
Configurable numeric precision for Diagnostics panel values

jhurliman:
**User-Facing Changes**
- Diagnostics Details panel now supports configurable numeric precision

**Description**
Screen Shot 2022-10-26 at 4 08 40 PM

### DIFF
--- a/packages/studio-base/src/panels/diagnostics/DiagnosticStatus.tsx
+++ b/packages/studio-base/src/panels/diagnostics/DiagnosticStatus.tsx
@@ -42,6 +42,7 @@ type Props = {
   splitFraction: number | undefined;
   onChangeSplitFraction: (arg0: number) => void;
   topicToRender: string;
+  precision: number | undefined;
   openSiblingPanel: OpenSiblingPanel;
 };
 
@@ -169,6 +170,7 @@ export default function DiagnosticStatus(props: Props): JSX.Element {
     onChangeSplitFraction,
     info,
     topicToRender,
+    precision,
     openSiblingPanel,
     splitFraction = 0.5,
   } = props;
@@ -219,6 +221,15 @@ export default function DiagnosticStatus(props: Props): JSX.Element {
       if (html) {
         return <HTMLTableCell dangerouslySetInnerHTML={html} />;
       }
+
+      let strToRender = str;
+      if (precision != undefined) {
+        const maybeFloat = parseFloat(strToRender);
+        if (!isNaN(maybeFloat)) {
+          strToRender = maybeFloat.toFixed(precision);
+        }
+      }
+
       return (
         <>
           <TableCell padding="checkbox">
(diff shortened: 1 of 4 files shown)


packages/studio-base/src/panels/diagnostics/DiagnosticStatus.tsx
```diff
@@ -42,6 +42,7 @@ type Props = {
   splitFraction: number | undefined;
   onChangeSplitFraction: (arg0: number) => void;
   topicToRender: string;
+  precision: number | undefined;
```
defunctzombie, line 45: Let's change this variable to `numericPrecision` to be more specific and match the configuration label.

defunctzombie (commented):
The default, auto, and 0 behavior feels awkward.

Here's what I see when I opened the preview deployment for the first time:

image

I think this means "auto"? But I don't know.

After changing the precision and then going back to `0` I see the following:

image

I have no way to go back to "auto" now.

packages/studio-base/src/panels/diagnostics/DiagnosticStatus.tsx
```diff
@@ -219,6 +221,15 @@ export default function DiagnosticStatus(props: Props): JSX.Element {
       if (html) {
         return <HTMLTableCell dangerouslySetInnerHTML={html} />;
       }
+
+      let strToRender = str;
+      if (precision != undefined) {
+        const maybeFloat = parseFloat(strToRender);
```
jtbandes, line 227: So if someone has an entry with the string “0x12”, and they set numeric precision, that’ll turn into “0.00”?

jhurliman:
> I have no way to go back to "auto" now.

@defunctzombie you go back to "auto" by deleting the input value. This allows you to differentiate between "auto" and `0` which have different behavior (full precision vs integer rounding).

defunctzombie:
> @defunctzombie you go back to "auto" by deleting the input value. This allows you to differentiate between "auto" and `0` which have different behavior (full precision vs integer rounding).

Makes sense. Would be good for the display to show "auto" when that happens or when there is no value. I believe we do this in other panels.

jhurliman:
@defunctzombie comments addressed, can you take another look?

packages/studio-base/src/panels/diagnostics/DiagnosticStatus.tsx
```diff
@@ -368,3 +378,14 @@ export default function DiagnosticStatus(props: Props): JSX.Element {
...
+// Returns true if the input string can be parsed as a float or an integer using
+// parseFloat(). Hex and octal numbers will return false.
+function isFloatOrInteger(n: string): boolean {
+  if (n[0] === "0" && n.length > 1) {
+    if (n[1] === "x" || n[1] === "X" || n[1] === "o" || n[1] === "O") {
+      return false;
+    }
+  }
+  return !isNaN(parseFloat(n)) && isFinite(Number(n));
+}
```
jtbandes, line 391: 👍 Unit tests please

defunctzombie (approved):
Feels much better! Please add the requested unit test and LGTM.

PR #4713, opened 2022-10-26
Adds Stixel View rendering option to point cloud renderer 

DavidXenakis:
**User-Facing Changes**
Adds an option on the point cloud renderer to visualize points as stixels. These stixels match the color of the point and extend from the point's Z location to 0.

CC @jhurliman -- This MR isn't exactly ready to be merged, so I'd love your feedback. The one functional issue which I've noticed is that it appears that the pose of the stixel geometry is wrong, even though I'm feeding it the same pose as the corresponding point cloud. Another issue is just around code organization. The PointCloudsAndLaserScans file could use some refactoring, imo, and I'm adding to the tech debt there with this change.

**Description**
Screen Shot 2022-10-26 at 10 31 25 AM

https://user-images.githubusercontent.com/3076080/198083130-60c6031c-e2f4-49a5-bdd3-6f398767494d.mov


jtbandes:
> The one functional issue which I've noticed is that it appears that the pose of the stixel geometry is wrong

That would probably be because the pose is updated here and it doesn't look like the stixels are being updated: https://github.com/foxglove/studio/blob/505b255e4191fc65e9f188cb9db78688186cbedf/packages/studio-base/src/panels/ThreeDeeRender/renderables/PointCloudsAndLaserScans.ts#L803-L811

jtbandes:
> Another issue is just around code organization. The PointCloudsAndLaserScans file could use some refactoring, imo, and I'm adding to the tech debt there with this change.

The overall structure of this change seems alright to me – it's mostly adding things that mirror the existing geometries/materials and it's not changing the organization at all. When you were working on it, which parts did you feel were most in need of organizational improvement?

packages/studio-base/src/panels/ThreeDeeRender/renderables/PointCloudsAndLaserScans.ts
```diff
@@ -577,10 +625,18 @@ export class PointCloudAndLaserScanRenderable extends Renderable<PointCloudAndLa
         (tempColor.b * 255) | 0,
         (tempColor.a * 255) | 0,
       );
+      if (settings.stixel) {
+        stixelPositionAttribute.setXYZ(i * 2, x, y, z);
+        stixelPositionAttribute.setXYZ(i * 2 + 1, x, y, 0);
+        stixelColorAttribute.setXYZW(i * 2, tempColor.r, tempColor.g, tempColor.b, tempColor.a);
+        stixelColorAttribute.setXYZW(i * 2 + 1, tempColor.r, tempColor.g, tempColor.b, tempColor.a);
+      }
```
jhurliman, line 633: Should this be `if (settings.stixel) { ... } else { ... }`? Right now it is updating four buffers (positionAttribute, colorAttribute, stixelPositionAttribute, and stixelColorAttribute) in stixel mode.

packages/studio-base/src/panels/ThreeDeeRender/renderables/PointCloudsAndLaserScans.ts
```diff
@@ -95,6 +98,8 @@ type PointCloudAndLaserScanUserData = BaseUserData & {
   material: Material;
   pickingMaterial: THREE.ShaderMaterial | LaserScanMaterial;
   instancePickingMaterial: THREE.ShaderMaterial | LaserScanMaterial;
+  stixelsHistory: StixelsAtTime[];
+  stixelMaterial: THREE.LineBasicMaterial;
```
jhurliman, line 102: Ideally these wouldl be optionals (`?`) and only instantiated if stixels are being used, so there is no cost of additional buffer manipulation when they're turned off. I'm not sure if that would make the code significantly more complex.

packages/studio-base/src/panels/ThreeDeeRender/renderables/PointCloudsAndLaserScans.ts
DavidXenakis: What would be in the else? It makes sense that it would be updating four buffers in stixel mode, since it need to draw both the position and color of both the stixels and the points themselves. The stixels aren't meant to replace the points.

DavidXenakis:
> The overall structure of this change seems alright to me – it's mostly adding things that mirror the existing geometries/materials and it's not changing the organization at all. When you were working on it, which parts did you feel were most in need of organizational improvement?

Hey Jacob, I just felt I was having to repeat myself quite a lot. It's not a big deal, but perhaps a little error prone.

defunctzombie:
@jhurliman Any further feedback on this or are we good with this to merge?

DavidXenakis:
Hey guys, I've been chipping away at this MR, trying to clean up, refactor, improve the quality of the code, and fix the pose issue. One thing I've realized is that my solution will not work for LaserScans yet. Does studio.foxglove.dev offer any example laser scan data I could test with? Or would it be okay to make this feature only supported for PointClouds?

jtbandes:
I think supporting it for both would probably be preferable. This bag file has some example LaserScan data: https://github.com/ROBOTIS-GIT/bags/blob/master/TB3_WAFFLE_SLAM.bag

jhurliman:
Laser scans don't have variable height, so it's not clear to me what the value of drawing all equal-height lines would be. Am I missing something?

jtbandes:
Hm you're right, I forgot they also don't have a `pose` field, so it would have to show the deviation from the fixed frame's xy plane. I guess it's not desirable to support this for LaserScans.

jtbandes:
I think it would be best to tackle this issue first, in a separate PR:
- https://github.com/foxglove/studio/issues/4803

snosenzo:
So I've updated this to fit into the refactors that have been done:

I've refactored the `pointsHistory` functionality to be encapsulated in a class that can be instantiated and stored on the renderable. This has been made more generic to accomodate for the stixels as well as points. So now there are two separate histories for stixels and points. To accommodate for this `createPoints` and `createStixels` have been moved to the `Renderable` constructor. 
I noticed that we weren't being consistent with updating the buffer usage so also add some logic to make sure they're properly set for decay/non-decay modes.

jtbandes (commented):
@snosenzo Not entirely sure what the problem is, but this test file seems to show stixels misaligned from corresponding points. Maybe a problem with pose handling?

image

[gh4892_decay_time.mcap.zip](https://github.com/foxglove/studio/files/10362692/gh4892_decay_time.mcap.zip)


<details><summary>Script:</summary>

```py
import math
import struct
from os import path
from tqdm import tqdm
from foxglove_schemas_protobuf.LaserScan_pb2 import LaserScan
from foxglove_schemas_protobuf.PointCloud_pb2 import PointCloud
from foxglove_schemas_protobuf.FrameTransform_pb2 import FrameTransform
from foxglove_schemas_protobuf.PackedElementField_pb2 import PackedElementField
from pyquaternion import Quaternion


    writer = Writer(stream)

    duration_ns = 10_000_000_000
    nframes = 100
        t = int(frame * duration_ns / nframes)

        if frame%10==0:
            q = Quaternion(axis=[0, 0, 1], angle=math.pi/4 * frame/nframes)
            scan = LaserScan()
            scan.frame_id = "sensor1"
            scan.timestamp.FromNanoseconds(t)
            scan.start_angle = 0
            scan.end_angle = 0.2
            scan.pose.position.x = 1 * frame/nframes
            scan.pose.orientation.x = q.x
            scan.pose.orientation.y = q.y
            scan.pose.orientation.z = q.z
            scan.pose.orientation.w = q.w
            scan.ranges[:] = [1, 1, 1, 1, 1, 1, 1]
            scan.intensities[:] = [1, 2, 3, 4, 5, 6, 7]
            writer.write_message("scan", scan, t)

            cloud = PointCloud()
            cloud.frame_id = "sensor2"
            cloud.timestamp.FromNanoseconds(t)
            cloud.fields.add(
                name = "x",
                offset = 0,
                type = PackedElementField.FLOAT32
            )
            cloud.fields.add(
                name = "y",
                offset = 4,
                type = PackedElementField.FLOAT32
            )
            cloud.fields.add(
                name = "z",
                offset = 8,
                type = PackedElementField.FLOAT32
            )
            cloud.fields.add(
                name = "i",
                offset = 12,
                type = PackedElementField.UINT8
            )
            data = bytearray(13 * 7)
            cloud.point_stride = 13
            for i in range(0, 7):
                struct.pack_into("<fffB", data, i * 13, 1, 0.3*i/7, math.sin(i+frame/nframes), i)
            cloud.pose.position.x = 1 * frame/nframes
            q = Quaternion(axis=[0, 0, 1], angle=-math.pi/4 * frame/nframes)
            cloud.pose.orientation.x = q.x
            cloud.pose.orientation.y = q.y
            cloud.pose.orientation.z = q.z
            cloud.pose.orientation.w = q.w
            cloud.data = bytes(data)
            writer.write_message("cloud", cloud, t)

        tf = FrameTransform()
        tf.parent_frame_id = "base"
        tf.child_frame_id = "sensor1"
        tf.timestamp.FromNanoseconds(t)
        tf.translation.y = 3 * frame/nframes
        tf.rotation.w = 1
        writer.write_message("tf", tf, t)

        tf = FrameTransform()
        tf.parent_frame_id = "base"
        tf.child_frame_id = "sensor2"
        tf.timestamp.FromNanoseconds(t)
        tf.translation.y = -3 * frame/nframes
        tf.rotation.w = 1
        writer.write_message("tf", tf, t)

        tf = FrameTransform()
        tf.parent_frame_id = "world"
        tf.child_frame_id = "base"
        tf.timestamp.FromNanoseconds(t)
        tf.translation.x = frame/nframes
        tf.rotation.w = 1
        writer.write_message("tf", tf, t)

    writer.finish()
    stream.close()
```

</details>

jtbandes (approved):
Thanks for all the collaborative work here!

Issue #2792, opened 2022-02-04
URL param to allow loading layout from a remote URL


jtbandes:
Related: https://github.com/foxglove/studio/issues/2787

I'd like to embed Studio in my website using an `<iframe>`. I need to be able to specify a layout, but I can't control the localStorage or the user's account, so I'd like to do something like `&layout.data=<base64...>` or `&layout.url=https://example.org/mylayout.json`.

defunctzombie:
Related - loading a set of "team layouts" or "layout templates" from a url? Maybe a separate issue?

jhurliman:
Also related to #2782

jtbandes:
Hm, maybe a duplicate, although that other ticket seems to be trying to cover multiple things in one.

jameskuszmaul-brt:
+1 to this. Not sure what the reference to "team layouts" is, but being able to load in with multiple useful layouts available to the user would be great as well. But just getting a single default layout via remote url + deep linking would be amazing.

jasonimercer:
A nice blend between the old webviz feature of 1 hosted layout and the current Foxglove abilities would be to allow us to load a set of named layouts from a URL/json file.

In our webviz use-case we can launch the viewer using different layouts for realtime technical observations of the state of the robots at "dark sites" where there is no internet access. We'd like to transition to Foxglove and the ability to have curated layouts hosted on the robot and switchable at runtime would be very appealing.

amacneil:
Importing comment from https://github.com/foxglove/studio/issues/2117#issue-1048474869

> In my firm we build ROS1 and ROS2 based robots. We plan to use FoxGlove to propose some maintenance pages for our robots. Most of the time, when users will use these maintenance pages, they will be offline (without any internet connection, and we can't assume that they already had one). I would like to serve foxglove web on a webserver hosted on the robot. And I would like to have some predefined layouts hosted on the robot. These predefined layouts will be designed by the robots development teams. They haven’t vocation to be modifiable.
>
> Today I do this with webviz by using the URL parameter :"layout-url", and having some JSON layout hosted on my robot. That works like a charm. I read the shared gdoc document “FoxGlove studio deeplink”, and my use case doesn't seem to be addressed. (https://docs.google.com/document/d/1jULICvQaZbJM8Et4ylFCE7GIFwedq7YkB4NhvnpjW0k/edit#heading=h.ic2oeiqyrxid)
>
[...]

PR #4716, opened 2022-10-27
Adds support for layoutUrl deep-link

RileyEv:
**User-Facing Changes**
Adds a `layoutUrl` parameter for deep links. e.g.
`https://studio.foxglove.dev/?layoutUrl=%2Flayout.json`

**Description**
Loads the layout from the URL, saves it to storage, and opens the layout.
Can be relative or absolute URL, as long as absolute URLs have correct CORS setup

Resolves foxglove/community#192


jk-ethz:
Is this working already? I'd love to see this feature merged :tada:

RileyEv:
> Is this working already?

The branch works if you build locally, I've just messed up a few tests, which I'm planning to fix today.

> Do we have control over what happens when the layout was modified locally and the URL is opened again? Like by an layoutUrlOverwrite=true/false URL parameter?

No there isnt anything like that at the moment. It just creates a new layout with the json of the url. Maybe something to consider, however, the URL is not currently stored so it would not know which layout to overwrite, unless I changed the name of the layout to be the full URL

jk-ethz:
> The branch works if you build locally, I've just messed up a few tests, which I'm planning to fix today.

Great!

> No there isnt anything like that at the moment. It just creates a new layout with the json of the url. Maybe something to consider, however, the URL is not currently stored so it would not know which layout to overwrite, unless I changed the name of the layout to be the full URL

We are planning to make this happen by a reverse proxy (i.e. the user opens `foxglove.local/` and our proxy changes that to `foxglove.local/?layoutUrl=...`, so every time we open the page we would get a new layout? What about another `layoutName` param in the URL instead of using the full URL as a name? Or using some name param in the JSON file as a name, but that's probably more involved.

jtbandes:
Thanks for the PR! It would be nice to fix the issue with duplicate layouts being created each time the page loads. We fixed a similar issue with the "explore sample data" feature here: https://github.com/foxglove/studio/pull/3339 The workaround was to avoid creating a layout with the same name.

RileyEv:
I agree, it'd be nice to solve that. I guess the difference here is that you'd probably want the layout to be updated every time you open it from the url. I'll have a go at doing that

jtbandes:
FYI, we'll need to have some internal discussions about this before we can provide a decision on this PR — we haven't yet reached consensus on how we want to address the use cases discussed in #2792 and don't necessarily want to expose the layout json format as part of the app's "public API". Thanks for your patience.

jk-ethz:
> [...] and don't necessarily want to expose the layout json format as part of the app's "public API". Thanks for your patience.

Aren't you already exposing it at this point since you allow to export/import the layout json files?

packages/studio-base/src/hooks/useInitialDeepLinkState.test.tsx
```diff
@@ -152,4 +170,32 @@ describe("Initial deep link state", () => {
...
+    );
+
+    // Required to wait for the async callback to have completed
+    await act(async () => {
+      await new Promise((resolve) => process.nextTick(resolve));
+    });
+
+    expect(fetch).toHaveBeenCalled();
+    expect(saveNewLayout).toHaveBeenCalledWith({
+      name: "layout.json",
```
amacneil, line 193: I suggest we remove the suffix from the layout name. Also, can we add a test demonstrating expected behavior if there is a folder path?

I propose that `lahoutUrl=/layouts/debug.json` should result in creating or updating a layout named `debug`.

amacneil:
👍 I'm on board with this feature.

I think if there is an existing layout with that name, we should overwrite it. For the demo layout we do not override it, but that is a different situation imo. If someone passes `layoutUrl` they will expect it to load the app using the exact layout they specified.

>> [...] and don't necessarily want to expose the layout json format as part of the app's "public API".
>
> Aren't you already exposing it at this point anyway since you allow to export/import the layout json files?

@jk-ethz Layout JSON is an internal representation, and despite it being exportable, we don't currently provide any guarantee about it being portable between different versions. That said, we can document `layoutUrl` as an advanced feature, and recommend that people re-test their layout when updating the app.

packages/studio-base/src/hooks/useInitialDeepLinkState.test.tsx
RileyEv: I've added a test which runs through some different cases :smile:

RileyEv:
> I think if there is an existing layout with that name, we should overwrite it.

One thing I have thought is, what happens if a user has a local layout that happens to have the same name. Is it worth prefixing the layout with something to indicate that its remote and reduce the risk of collisions. For example `Remote: debug` instead of just `debug`

amacneil:
I don't think we should worry about overwriting layouts. The primary use case for this feature I have heard is that people are hosting Foxglove Studio on their robot, and want to open a specific layout. In that scenario you are less likely to have personal layouts, and if the naming conflict is a problem then the admin setting up the layout URL can choose a unique name.

RileyEv:
@jtbandes @amacneil Any chance this can get merged? I've just given it a rebase as there were conflicts.

packages/studio-base/src/hooks/useInitialDeepLinkState.ts
```diff
@@ -70,32 +79,104 @@ function useSyncSourceFromUrl(
...
+        layoutId?: LayoutID;
+        layoutUrl?: string;
+      }
+    | undefined
+  >(
+    targetUrlState
+      ? { layoutId: targetUrlState.layoutId, layoutUrl: targetUrlState.layoutUrl }
+      : undefined,
+  );
+  const [fetchingLayout, setFetchingLayout] = useState(false);
```
snosenzo, line 101: I think using a ref here might be better since we don't need to trigger a re-render after setting it either true or false

packages/studio-base/src/hooks/useInitialDeepLinkState.ts
```diff
@@ -70,32 +79,104 @@ function useSyncSourceFromUrl(
...
+        newLayout = await layoutManager.updateLayout({
+          id: sourceLayout.id,
+          name,
+          data,
+        });
+      }
+
+      setSelectedLayoutId(newLayout.id);
+      setUnappliedLayoutArgs((oldState) => ({ ...oldState, layoutUrl: undefined }));
+      setFetchingLayout(false);
```
snosenzo, line 141: The setting of this variable (whether it be in a ref or useState) should probably be done in `fetchLayout.finally(() => {...})`, just in case there is an error and `setFetchingLayout` is never set to false

RileyEv: Good point :smile: I've add a finally case to `fetchLayout`

snosenzo (approved):

RileyEv:
@snosenzo could this get merged?

jhurliman (approved):

banisadr:
@RileyEv we owe you an apology for going back and forth on this one so many times and leaving it hanging for this long. Sorry about that - we can do better and we absolutely plan to going forward.

We've made the decision to not merge this PR. At the very minimum, we owe you an explanation - there are a few factors that went into this. First, we don't consider the JSON schema that defines a layout to be a stable API, so supporting this as a feature could age poorly for users unless we also commit to that. Second, we have a few features we are working on landing in the first quarter of this year that overlap with this feature's use cases but in a way that enable us to ensure the stability and build additional feature's we're targeting.

That said, we are in agreement with your intent here that both the ability to load a layout definition while offline as well as the ability to publicly share and load a layout have significant user value so please rest assured that the feature work I mentioned we have planned does enable these capabilities as well.

maxpolzin:
@banisadr Great to hear that you are planning to enable offline loading of layout definition. Looking forward to have it integrated in the main repo.

siddancha:
Any ETA on when the new feature you mentioned is going to land?

P. S. IMHO being able to publicly share and load layouts would significantly increase Foxglove's adoption in academic research. This is because how academia is structured; there exists a class of managers (aka professors) who would really appreciate interactive visualizations and debugging with a single click but would prefer not to install software or organize project-specific layouts since they're involved in multiple projects.

banisadr:
@siddancha we're aiming to have layout sharing available in the next month or so. The first iteration of this will be private to people within your org, but I completely agree with the value of being able to share layouts (and data) publicly. We want to build on this feature to provide both as options.

mfahem:
@banisadr Hi, any news on the ability to share layouts publicly ? This feature would be a great addition and useful for many people IMO.

banisadr:
@mfahem its still on our roadmap but likely not in the next 3-6 months. I don't have a timeline that we can share yet.

PR #4735, opened 2022-11-01
Add Toggle for Auto-Calculating Min/Max to Datatypes with Fields

### LINKED ISSUE #4720: Add Button to calculate Min/Max field values for Current Message when using value color modes
For renderables that use PackedElementFields and offer value colorMode options (gradient, color mapping), the user should be able to click a button in the settings panel to calculate the min and max value of the currently shown message.
For point clouds it will cut down on this needing to be done every new message, and for grids it will give an easier way to calculate it if the user doesn't know what the min/max values are. 
This will also enable the coloring of point clouds to be done in the shader in a future PR.

snosenzo:
**User-Facing Changes**
 - can now toggle the finding of min/max field values for topics with datatypes that have fields and colorMap modes
 - will find the new min/max values every message and set them in the UI when active. the min/max value fields are disabled while this is active
 - leaving the min/max value field as auto/undefined will remain the same as before while the Calculate Min/Max toggle is disabled

**Description**
 - Move more logic into renderable for foxglove grids
 - Tried to separate things out more in foxglove grid to optimize updates.
 - with point clouds I tried to change as little as possible while maintaining the same overall pattern as the grids
 - Point clouds now hold the min/max color values on the instance. This is read after reading/handling the message to set the UI by calling `updateSettingsMinMax`/`_updateMinMaxColorValues`. Hopefully it's fine to call the `saveSetting` twice like that because of the debounce

https://user-images.githubusercontent.com/10187776/199346755-fe2d40ea-5bf6-47c8-9455-b8187ea109a9.mov

Fixes #4720

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/FoxgloveGrid.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/FoxgloveGrid.ts
@@ -471,12 +471,12 @@ export class FoxgloveGrid extends SceneExtension<FoxgloveGridRenderable> {
       renderable.handleSettingsChange(fullSettings);
 
       if (settings?.calcMinMax === true) {
-        this.updateSettingsMinMaxFromRenderable(renderable);
+        this._updateMinMaxValueSettings(renderable);
       }
     }
   };
 
-  private updateSettingsMinMaxFromRenderable(renderable: FoxgloveGridRenderable) {
+  private _updateMinMaxValueSettings(renderable: FoxgloveGridRenderable): void {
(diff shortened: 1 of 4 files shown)


amacneil:
Product review feedback:

This conflicts with the existing "auto" setting. Also, calculating range per frame is not great for performance, so we probably want to evaluate different approaches to setting the default colorization range.

cc @defunctzombie @jhurliman @2metres

snosenzo:
Going to close this and the issue for now. It's not necessary for Point Cloud shader performance improvements, and it doesn't seem like there's much of a value add.

PR #4738, opened 2022-11-01
Move unavailable data source error messages into alerts

### LINKED ISSUE #4727: All unavailable data source errors should be in a yellow alert container
**Description**
- All unavailable data source errors (`Ros1UnavailableDataSourceFactory`, `Ros2UnavailableDataSourceFactory`, `VelodyneUnavailableDataSourceFactory`) should be in a 2nd yellow alert container at the top of the Data Source modal (below any existing ones, like below):

Screen Shot 2022-10-31 at 4 14 51 PM

Unavailable data source errors (currently below input fields):

Screen Shot 2022-11-01 at 10 38 59 AM
Screen Shot 2022-11-01 at 10 39 03 AM
Screen Shot 2022-11-01 at 10 39 07 AM

esthersweon:
**User-Facing Changes**
- Move unavailable data source error messages into alerts

Screen Shot 2022-11-01 at 4 13 04 PM
Screen Shot 2022-11-01 at 4 13 01 PM
Screen Shot 2022-11-01 at 4 13 08 PM

Resolves #4727

### DIFF
--- a/packages/studio-base/src/components/OpenDialog/Connection.tsx
+++ b/packages/studio-base/src/components/OpenDialog/Connection.tsx
@@ -130,7 +130,9 @@ export default function Connection(props: ConnectionProps): JSX.Element {
           </Tabs>
         </Stack>
         <Stack key={selectedSource?.id} flex="1 0" gap={2}>
-          {selectedSource?.warning && <Alert severity="warning">{selectedSource.warning}</Alert>}
+          {selectedSource?.disabledReason == undefined && selectedSource?.warning && (
+            <Alert severity="warning">{selectedSource.warning}</Alert>
+          )}
           {selectedSource?.disabledReason != undefined && (
             <Alert severity="warning">{selectedSource.disabledReason}</Alert>
           )}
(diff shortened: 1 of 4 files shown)


defunctzombie (approved):
Awesome.

Some suggestions if they are easy:

1. Hide the "limitations of ros 2" notice from the web version since we already tell you ros 2 native is not available it is a bit silly to tell you the limitation of it
2. Hide the form content altogether and only show the warning text for the unavailable sources. No sense in adding noise to the dialog for fields you can't use.

esthersweon:
@defunctzombie - 1st item is easy enough to do. For the 2nd, I think it might be confusing and look a little strange if we don't show the disabled inputs, as users wouldn't be aware of what that flow might look like. cc: @2metres

defunctzombie:
> For the 2nd, would we also hide the "Open" button then?

Disabling the button would be good but we might not have an easy way to do that.

> think it would be confusing if we don't show the disabled inputs, as it'd look like that workflow doesn't even exist

We've already told the user they need to use the desktop app. The workflow exists there. These exist as placeholders for discovery that you need to use the desktop app if you want to use this connection type. As a user you can't configure it.

If the above are annoying or you disagree - I am 👍 with this change as-is since it is better than what we have now with no aspect of it being worse to my experience.

jtbandes:
> Hide the form content altogether and only show the warning text for the unavailable sources. No sense in adding noise to the dialog for fields you can't use.

to provide a dissenting opinion, I think showing the fields is nice because it helps to give people an idea of what the feature is. The whole form should be disabled if the source is unavailable though.

PR #4742, opened 2022-11-02
Fix Flaky screenshot test: Plot:in a line graph with multiple plots, x-axes are synced

### LINKED ISSUE #4733: Flaky screenshot test: Plot:in a line graph with multiple plots, x-axes are synced
The `Plot:in a line graph with multiple plots, x-axes are synced` screenshot test is flaky. This test constantly shows up as a diff in PRs that do not have anything to do with plot panel changes. Flaky tests introduce uncertainty to the review and approval process. Even worse, over time we become desensitized to them and may miss a legitimate regression.

[Plot](https://www.chromatic.com/component?appId=603ec8bf7908b500231841e2&csfId=panels-plot&buildNumber=10307):[in a line graph with multiple plots, x-axes are synced](https://www.chromatic.com/component?appId=603ec8bf7908b500231841e2&csfId=panels-plot&specName=in%20a%20line%20graph%20with%20multiple%20plots,%20x-axes%20are%20synced&buildNumber=10307)

snosenzo:
**User-Facing Changes**
 - n/a

**Description**
 - I don't really know a better way to test this other than running it a few times in CI
 - since this is really hard to test locally I'm just going to add a minHeight css via child selector to the PanelSetup component to assign the minHeight for each panel to be 50%. this at least prevents the test from being flaky (so far)

Fixes #4733

### DIFF
--- a/packages/studio-base/src/panels/Plot/index.stories.tsx
+++ b/packages/studio-base/src/panels/Plot/index.stories.tsx
@@ -463,7 +463,7 @@ export function InALineGraphWithMultiplePlotsXAxesAreSynced(): JSX.Element {
   const pauseFrame = useCallback(() => readySignal, [readySignal]);
 
   return (
-    <PanelSetup fixture={fixture} pauseFrame={pauseFrame} style={{ flexDirection: "column" }}>
+    <PanelSetup fixture={fixture} pauseFrame={pauseFrame}>
       <Plot
         overrideConfig={{
           ...exampleConfig,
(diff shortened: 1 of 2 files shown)


jtbandes (commented):
Instead of CSS hackery, can you try setting the colorScheme for this story so we don't get the side-by-side view? I have been wondering if something about having them side by side is causing issues.

Another thing to try before that would be increasing `count` to 7.

snosenzo:
I'll try the count because that could be it, but I think getting rid of the side-by-side isn't really a better solution since I feel like it doesn't tell us what's going on. I also don't want to have to wait for this build to finish and kick off another commit another 10 times 😅

snosenzo:
looks like neither `count: 7` nor `count: 8` worked. I'll try removing the side-by-side

snosenzo:
removing the dark scheme also didn't work. Going to revert to my original solution

jtbandes (approved):
Doesn't make any sense, but I guess it's better than having a flaky story

PR #4745, opened 2022-11-02
Add MCAP mdimporter to macOS app bundle

jtbandes:
**User-Facing Changes**
Added a Spotlight importer for MCAP files to the macOS desktop app. The names of topics, schemas, attachments, and metadata in MCAP files can now be used as search terms in Finder and Spotlight.

**Description**
Added https://github.com/foxglove/MCAPSpotlightImporter to app bundle via afterPack script.

Also made some updates to Studio's Info.plist:
- Make Studio the "owner/exporter" of .mcap and .foxe types
- Rename MCAP UTI from `dev.foxglove.mcap` to `dev.mcap.mcap` now that mcap.dev is a registered domain
- Added McapIcon.png which was missing from the app bundle
- Add more conformances to "functional" type hierarchy for mcap and bag files: `public.log` and `public.composite-content`
- Change .bag description from "ROS Bag File" to "ROS 1 Bag File"
- Added reference URLs for all the file types

Finder Get Info window
Finder search
Finder search by topics and schemas
Spotlight search

### DIFF
--- a/desktop/afterPack.ts
+++ b/desktop/afterPack.ts
@@ -60,7 +60,7 @@ async function copySpotlightImporter(context: AfterPackContext) {
   const zipPath = path.join(outDir, "MCAPSpotlightImporter.mdimporter.zip");
   await fs.unlink(zipPath).catch(() => {});
   await downloadTool(
-    "https://github.com/foxglove/MCAPSpotlightImporter/releases/download/v1.0.1/MCAPSpotlightImporter.mdimporter.zip",
+    "https://github.com/foxglove/MCAPSpotlightImporter/releases/download/v1.0.2/MCAPSpotlightImporter.mdimporter.zip",
     zipPath,
   );
   try {
@@ -114,7 +114,7 @@ async function configureQuickLookExtension(context: AfterPackContext) {
     NSExtension: {
       ...(originalInfo.NSExtension as PlistObject),
       NSExtensionAttributes: {
-        QLSupportedContentTypes: ["org.ros.bag", "dev.foxglove.mcap"],
+        QLSupportedContentTypes: ["org.ros.bag", "dev.mcap.mcap"],
         QLSupportsSearchableItems: false,
       },
     },
--- a/desktop/afterPack.ts
+++ b/desktop/afterPack.ts
@@ -3,6 +3,7 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
 import { exec } from "@actions/exec";
+import { downloadTool, extractZip } from "@actions/tool-cache";
 import type MacPackager from "app-builder-lib/out/macPackager";
 import { log, Arch } from "builder-util";
 import { AfterPackContext } from "electron-builder";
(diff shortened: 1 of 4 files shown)


desktop/afterPack.ts
```diff
@@ -21,6 +22,61 @@ async function getKeychainFile(context: AfterPackContext): Promise<string | unde
...
+  if (electronPlatformName !== "darwin") {
+    return;
+  }
+
+  const appName = context.packager.appInfo.productFilename;
+  const appPath = path.join(appOutDir, `${appName}.app`);
+  const spotlightDirPath = path.join(appPath, "Contents", "Library", "Spotlight");
+  const zipPath = path.join(outDir, "MCAPSpotlightImporter.mdimporter.zip");
+  await fs.unlink(zipPath).catch(() => {});
+  await downloadTool(
```
james-rms, line 62: can we verify the hash of the downloaded file here?

desktop/afterPack.ts
```diff
@@ -21,6 +22,61 @@ async function getKeychainFile(context: AfterPackContext): Promise<string | unde
...
+async function copySpotlightImporter(context: AfterPackContext) {
+  const { electronPlatformName, outDir, appOutDir } = context;
+  if (electronPlatformName !== "darwin") {
+    return;
+  }
+
+  const appName = context.packager.appInfo.productFilename;
+  const appPath = path.join(appOutDir, `${appName}.app`);
+  const spotlightDirPath = path.join(appPath, "Contents", "Library", "Spotlight");
+  const zipPath = path.join(outDir, "MCAPSpotlightImporter.mdimporter.zip");
```
james-rms, line 60: why not unzip from memory rather than downloading the zip into the FS first?

jtbandes: Mainly because there was a convenient `extractZip` function available from @actions/tool-cache

james-rms: I wonder if this would cause an exciting bug if the extract dir was in `~/Downloads`, and Mac OS went ahead and applied the ` com.apple.quarantine` xattr to the bundled executable.

jtbandes: Is this a thing that happens automatically for arbitrary files? This file is being created by some JS code, not downloaded from a browser or by extracting a dmg. I don't think I've heard of the quarantine attribute being added for files created directly by programs that aren't browsers.

desktop/afterPack.ts
```diff
@@ -21,6 +23,69 @@ async function getKeychainFile(context: AfterPackContext): Promise<string | unde
...
+  const zipURL =
+    "https://github.com/foxglove/MCAPSpotlightImporter/releases/download/v1.0.2/MCAPSpotlightImporter.mdimporter.zip";
+  const zipSHA = "26cafa3e3069fcbd294864ceeee1bc9899e94456e0d28079f966787b6f05c7a2";
+  await downloadTool(zipURL, zipPath);
+  const actualSHA = crypto
+    .createHash("sha256")
+    .update(await fs.readFile(zipPath))
+    .digest("hex");
+  if (actualSHA !== zipSHA) {
+    throw new Error(`SHA mismatch for ${zipURL}: expected ${zipSHA}, got ${actualSHA}`);
```
snosenzo, line 73: Is there an action that the user can take here to fix this? Or will the solution vary too much to mention an action?

jtbandes: This indicates the download was corrupt or the url/sha is simply wrong. It's not within the scope of this code to do something about it, the user will have to figure out what went wrong (e.g. a network error)

Hopefully 99.99% of the time nobody will encounter this :)

defunctzombie (commented):
I'm all for this feature.

I'd feel happier about the deployment story if we could package as part of the release process (even if it is built separately) using the `extraResources` feature: https://www.electron.build/configuration/contents.html

This would let us avoid an additional download step and potential for failure. We can limit this to the macos package so Windows and Linux users are not paying with larger bundle sizes. 

If you looked into that and there were other complications with that approach then we can use this one and I'll approve the PR.

desktop/afterPack.ts
jtbandes: Btw, "user" here means "studio release engineer"

defunctzombie (approved):
I now understand that none of the download is happening on the user's machine. These download steps are in afterPack and doing exactly what I imagined - downloading the file for inclusion into the package bundle.

Issue #4746, opened 2022-11-02
add support to load pcl data with field count > 1


oosthoek:
i have rosbag with pointfield "timestamp" that has count > 1.
when trying to visualize, i get `PointCloud field "timestamp" has invalid count 8. Only 1 is supported`
i am not trying to visualize this field, but want to visualize the xyz fields also in the pcl. 
can support be added to still load the pcl?

i have attached sample bag 

[sample.zip](https://github.com/foxglove/studio/files/9924628/sample.zip)

jtbandes:
Thanks for the report. I'm upgrading this to a bug because the app should still be usable as long as you are not using this field.

Out of curiosity, why does the field have count `8`? Does it really contain 8 values? How did you create this point cloud?

PR #4747, opened 2022-11-02
Hide PointCloud2 fields having `count` ≠ 1

jtbandes:
**User-Facing Changes**
3D panel can now load PointCloud2 messages with fields whose `count` is unsupported (i.e. ≠ 1), and these fields are hidden. Previously, the whole point cloud topic was unusable.

**Description**
Resolves https://github.com/foxglove/studio/issues/4746

Hides these multi-valued fields from the "Color by" settings and the "selected object details" popup. A future feature could be to actually support multiple values in the selected object popup (displaying them as an array). For now we will leave it as-is until we receive user feedback that supporting multi-valued fields is desired.

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/PointCloudsAndLaserScans.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/PointCloudsAndLaserScans.ts
@@ -4,6 +4,7 @@
 
 import * as THREE from "three";
 
+import { filterMap } from "@foxglove/den/collection";
 import { Time, toNanoSec } from "@foxglove/rostime";
 import {
   LaserScan as FoxgloveLaserScan,
@@ -220,7 +221,9 @@ export class PointCloudAndLaserScanRenderable extends Renderable<PointCloudAndLa
       for (const field of pointCloud.fields) {
         const pointOffset = instanceId * pointStep;
         const reader = getReader(field, stride);
-        details[field.name] = reader?.(view, pointOffset);
+        if (reader) {
+          details[field.name] = reader(view, pointOffset);
+        }
       }
 
       return details;
@@ -431,9 +434,8 @@ export class PointCloudAndLaserScanRenderable extends Renderable<PointCloudAndLa
           : (field as PointField).datatype;
 
       if (count != undefined && count !== 1) {
-        const message = `PointCloud field "${field.name}" has invalid count ${count}. Only 1 is supported`;
-        invalidPointCloudOrLaserScanError(this.renderer, renderable, message);
-        return false;
+        // Skip this field, we don't support counts other than 1
+        continue;
       } else if (field.offset < 0) {
         const message = `PointCloud field "${field.name}" has invalid offset ${field.offset}. Must be >= 0`;
         invalidPointCloudOrLaserScanError(this.renderer, renderable, message);
@@ -996,7 +998,10 @@ export class PointCloudsAndLaserScans extends SceneExtension<PointCloudAndLaserS
     // Update the mapping of topic to point cloud field names if necessary
     let fields = this.pointCloudFieldsByTopic.get(topic);
     if (!fields || fields.length !== pointCloud.fields.length) {
-      fields = pointCloud.fields.map((field) => field.name);
+      // Omit fields with count != 1
+      fields = filterMap(pointCloud.fields, (field) =>
+        field.count === 1 ? field.name : undefined,
+      );
       this.pointCloudFieldsByTopic.set(topic, fields);
       this.updateSettingsTree();
     }
@@ -1380,6 +1385,9 @@ export function autoSelectColorField(
 ): void {
   // Prefer color fields first
   for (const field of pointCloud.fields) {
+    if ("count" in field && field.count !== 1) {
+      continue;
+    }
     const fieldNameLower = field.name.toLowerCase();
     if (COLOR_FIELDS.has(fieldNameLower)) {
       output.colorField = field.name;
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/panels/ThreeDeeRender/renderables/PointCloudsAndLaserScans.ts
```diff
@@ -431,9 +434,8 @@ export class PointCloudAndLaserScanRenderable extends Renderable<PointCloudAndLa
           : (field as PointField).datatype;
 
       if (count != undefined && count !== 1) {
-        const message = `PointCloud field "${field.name}" has invalid count ${count}. Only 1 is supported`;
-        invalidPointCloudOrLaserScanError(this.renderer, renderable, message);
-        return false;
+        // Skip this field, we don't support counts other than 1
```
defunctzombie, line 437: Previously we informed the user about this and that's how they surfaced the issue to us. Now we are silent on this. Is there some sensible way we can still indicate this to the user or do we think it will be sufficient for them to see that the field they expect is missing?

packages/studio-base/src/panels/ThreeDeeRender/renderables/PointCloudsAndLaserScans.ts
```diff
@@ -1380,6 +1385,9 @@ export function autoSelectColorField(
 ): void {
   // Prefer color fields first
   for (const field of pointCloud.fields) {
+    if ("count" in field && field.count !== 1) {
```
defunctzombie, line 1388: Do we need to check `in` here? should our type definition be updated to make count optional?

packages/studio-base/src/panels/ThreeDeeRender/renderables/pointClouds/fieldReaders.ts
```diff
@@ -45,6 +45,9 @@ export function getReader(
   stride: number,
   forceType?: PointFieldType | NumericType,
 ): FieldReader | undefined {
+  if ("count" in field && field.count !== 1) {
```
defunctzombie, line 48: This filtering is repeated several times - should we be filtering the fields earlier or are all these places operating on the original message vs a cleaned up version? Maybe a common function `isSupportedField(field)` worth introducing?

defunctzombie (commented):
Provided some feedback as an observer. I'll let sam or john take a look and provide the final approval since they do more work with the code here.

packages/studio-base/src/panels/ThreeDeeRender/renderables/PointCloudsAndLaserScans.ts
jtbandes: This is a `PackedElementField | PointField`, so TS doesn't let you access properties that aren't present in both

packages/studio-base/src/panels/ThreeDeeRender/renderables/PointCloudsAndLaserScans.ts
jtbandes: Do we want a topic error to be displayed? I considered that but thought it might be a bit annoying.

Another option would be to show it in the dropdown, but make it disabled? This would require adding a disabled flag to the extension settings API.

packages/studio-base/src/panels/ThreeDeeRender/renderables/PointCloudsAndLaserScans.ts
```diff
@@ -996,7 +998,10 @@ export class PointCloudsAndLaserScans extends SceneExtension<PointCloudAndLaserS
     // Update the mapping of topic to point cloud field names if necessary
     let fields = this.pointCloudFieldsByTopic.get(topic);
     if (!fields || fields.length !== pointCloud.fields.length) {
-      fields = pointCloud.fields.map((field) => field.name);
+      // Omit fields with count != 1
```
snosenzo, line 1001: Won't this update the fields more often than necessary? should it not filterMap the pointCloud.fields before comparing their lengths?

snosenzo (commented):
left a comment about the length comparison. I like Roman's suggestion of having a function for this in the `fieldReaders.ts` file

packages/studio-base/src/panels/ThreeDeeRender/renderables/PointCloudsAndLaserScans.ts
snosenzo: Not sure where we want to land on this. I agree that showing the error would probably be annoying. Maybe making them disabled with a tooltip would be best but might be beyond the scope of this bugfix

packages/studio-base/src/panels/ThreeDeeRender/renderables/pointClouds/fieldReaders.ts
```diff
@@ -40,11 +40,23 @@ export function float64Reader(fieldOffset: number): FieldReader {
   return (view: DataView, pointOffset: number) => view.getFloat64(pointOffset + fieldOffset, true);
 }
 
+export function isSupportedField(field: PackedElementField | PointField): boolean {
```
jtbandes, line 43: Would it make sense/work for this function to accept a PointField since all PackedElementFields are supported?

snosenzo: most of its occurrences are used on the `|` typed `field`. It would casting it to a PointField wherever it wasn't already narrowed to that type (everywhere except `handleRosPointCloud`)

jtbandes (commented):
Looks good, thanks for the help!

snosenzo (approved):

Issue #4744, opened 2022-11-02
DP playback issuing many requests and not proceeding under some condition


wkalt:
Repro steps:

1. Open this link in studio (must be signed into Foxglove HQ):
https://studio.foxglove.dev/?ds=foxglove-data-platform&ds.deviceId=dev_Wm1gvryKJmREqnVT&ds.end=2022-08-22T00%3A02%3A11.940Z&ds.start=2022-08-21T19%3A59%3A22.201Z&layoutId=9c6018d1-2b91-42ff-af34-4f6e4df6f68b&time=2022-08-21T19%3A59%3A53.022858460Z
2. Configure a single raw message panel and set it to display /imu
3. Hit play

Playback will make no progress. Observe in the network tab that we will make hundreds of tiny requests receiving no data back. There is legitimately no data in the small time window being requested.

Set the topic instead to /rosout and see that playback works.

It seems like we are hitting some condition where we enter a retry loop if no messages exist in the requested window. This can happen if your request falls between messages, and is more likely if the number of topics being requested is few.

defunctzombie:
Url for myself to test in hq dev: http://localhost:8080/?ds=foxglove-data-platform&ds.deviceId=dev_GLQsznQJcUbZvyf1&ds.end=2022-08-21T20%3A19%3A55.540Z&ds.start=2022-08-21T19%3A57%3A49.954Z&layoutId=a7ddb3bd-1bb1-4469-9ec0-a899b690b055&time=2022-08-21T20%3A02%3A49.738824292Z

PR #4750, opened 2022-11-03
Introduce a "stamp" IteratorResult

defunctzombie:
**User-Facing Changes**
Reduce the number of network requests when subscribed to sparsely populated topics and paused.

**Description**
The BufferedIterableSource is responsible for keeping a time-based read-ahead buffer of messages for playback. It does this using a producer and consumer pattern where the consumer reads from the buffer and updates the `readHead` location while the producer keeps messages buffered based on the read head. To keep the buffer filled, the producer reads messages from the underlying data source message iterator until some `readUntil` time which is a fixed offset ahead of the `readHead`. The producer stops when it encounters a message occurring after the `readUntil` time and waits for the consumer to move the `readHead` forward before attempting to read more messages.

The DataPlatformIterableSource implements two modes for message iteration. `partial` and `full`. In `partial` mode (the mode used for playback buffering), the implementation reads messages from the API using fixed-time request windows (currently 5 seconds). This is because we cannot control back-pressure on the request and want to avoid downloading large quantities of data prior to the user needing to play this data.

When opening a data platform stream and subscribing to a sparsely populated topic, the above behaviors result in many (hundreds to thousands) of requests to fetch each next 5 seconds of data from the api. This happens because the buffered source continues to read from the underlying source (data platform) until it can encounter a message _after_ the `readUntil` time. For a sparsely populated dataset it could be minutes or hours before the next such message exists and we get there in 5 second request increments.

This change provides a workaround to this behavior by introducing a new type of IteratorResult. Unlike the `messageEvent` or `problem` IteratorResult, the _stamp_ iterator result contains only a timestamp (no message or problem). The underlying source uses the stamp result to indicate it has loaded data through the stamp time as a signal to upstream message iterator consumers. The BufferedIterableSource is updated with logic to check the `stamp` time against the `readUntil` time to identify if the underlying source has been sufficiently buffered for playback. The DataPlatformIteraboeSource is updated to emit `stamp` results after each 5 second stream ends. Even if there are no messages in the stream, the stamp result indicates to callers that the specific timestamp has been reached.

Fixes: #4744

### DIFF
--- a/packages/studio-base/src/players/IterablePlayer/IIterableSource.ts
+++ b/packages/studio-base/src/players/IterablePlayer/IIterableSource.ts
@@ -54,6 +54,18 @@ export type MessageIteratorArgs = {
   consumptionType?: "full" | "partial";
 };
 
+/**
+ * IteratorResult represents a single result from a message iterator or cursor. There are three
+ * types of results.
+ *
+ * - message-event: the result contains a MessageEvent
+ * - problem: the result contains a problem
+ * - stamp: the result is a timestamp
+ *
+ * Note: A stamp result acts as a marker indicating that the source has reached the specified stamp.
+ * The source may return stamp results to indicate to callers that it has read through some time
+ * when there are no message events available to indicate the time is reached.
+ */
 export type IteratorResult =
   | {
       type: "message-event";
--- a/packages/studio-base/src/players/IterablePlayer/IIterableSource.ts
+++ b/packages/studio-base/src/players/IterablePlayer/IIterableSource.ts
@@ -59,11 +59,19 @@ export type IteratorResult =
       connectionId: number | undefined;
       msgEvent: MessageEvent<unknown>;
       problem: undefined;
+      stamp: undefined;
     }
   | {
       connectionId: number | undefined;
       msgEvent: undefined;
       problem: PlayerProblem;
+      stamp: undefined;
+    }
+  | {
+      stamp: Time;
+      connectionId: undefined;
+      msgEvent: undefined;
+      problem: undefined;
     };
 
 export type GetBackfillMessagesArgs = {
--- a/packages/studio-base/src/players/IterablePlayer/IIterableSource.ts
+++ b/packages/studio-base/src/players/IterablePlayer/IIterableSource.ts
@@ -56,22 +56,18 @@ export type MessageIteratorArgs = {
 
 export type IteratorResult =
   | {
-      connectionId: number | undefined;
+      type: "message-event";
+      connectionId?: number;
       msgEvent: MessageEvent<unknown>;
-      problem: undefined;
-      stamp: undefined;
     }
   | {
-      connectionId: number | undefined;
-      msgEvent: undefined;
+      type: "problem";
+      connectionId?: number;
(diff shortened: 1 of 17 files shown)


packages/studio-base/src/players/IterablePlayer/IIterableSource.ts
```diff
@@ -56,14 +56,18 @@ export type MessageIteratorArgs = {
...
   | {
-      connectionId: number | undefined;
-      msgEvent: undefined;
+      type: "problem";
+      connectionId?: number;
       problem: PlayerProblem;
+    }
+  | {
+      type: "stamp";
+      stamp: Time;
```
jtbandes, line 70: Need comment on what this means. What does the time represent?

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
```diff
@@ -776,13 +785,32 @@ export class IterablePlayer implements Player {
     const targetTime = add(this._currentTime, fromMillis(rangeMillis));
     const end: Time = clampTime(targetTime, this._start, this._untilTime ?? this._end);
 
+    if (this._lastStamp) {
+      // If the last message we saw is still ahead of the tick end time, we don't emit anything
+      if (compare(this._lastStamp, end) >= 0) {
```
jtbandes, line 790: Does `_lastStamp` correspond to the "last message we saw"? I thought a stamp result was different from a message result.

jtbandes (approved):
Gave the player stuff a cursory read. Can't say I fully understand it but am glad we have new tests being added 👍

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
```diff
@@ -776,13 +785,32 @@ export class IterablePlayer implements Player {
...
+        this._messages = [];
+        this._queueEmitState();
+
+        if (this._untilTime && compare(this._currentTime, this._untilTime) >= 0) {
+          this.pausePlayback();
+        }
+        return;
+      }
+
+      this._lastStamp = undefined;
```
snosenzo, line 804: this whole sequence is a little unclear to me, could use a few more comments. I just don't understand why these things are happening outside of not emiting state when the lastStamped > endTIme

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
defunctzombie: I've removed this comment and added a new one above the parent `if` statement.

packages/studio-base/src/players/IterablePlayer/IIterableSource.ts
defunctzombie: Added a comment for the `IteratorResult` type.

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
defunctzombie: I've added a comment to the parent `if` statement describing the logic and why this bail-out of the tick exists.

PR #4752, opened 2022-11-03
Fix Value Layout Flashing in Raw Message Panel on Hover

### LINKED ISSUE #4749: Weird UI when hovering over message in RawMessagePanel
**Description**

I saw a strange flickering when hovering over messages in the RawMessagePanel.
It occurs when the message panel is just barely wide enough to show the data.
When you then hover the mouse over any data element, it just goes back and forth.

I guess what actually happens, when you hover some additional stuff is displayed on the item, which introduces a line break and expands, but then as soon as it is expanded, the mouse is no longer hovering, so it collapses again. and this just repeats infinitely.

The video does not quite capture the actual speed at which it flickers.

Foxglove version 1.30.0

**Expected Behavior**
No jumping back and forth

**Actual Behavior**

https://user-images.githubusercontent.com/13848742/199702196-075e7a33-3ef0-46e8-aae4-3095c4a54a71.mp4

snosenzo:
**User-Facing Changes**
 - Fix Value Layout Flashing in Raw Message Panel on Hover when buttons cause wrapping

**Description**
 - Probably spent a little too much time trying to do this with just css, but was ultimately unable to. The parent `ReactHoverObserver` breaks to the next line before the child `Stack` has a chance to wrap. Do let me know if there might be a solution here beyond my knowledge. 
 - Final solution: use placeholder action icon buttons with a maximum of 4. Avoids slight perf impact from always running the `getValueActionForValue` for every value each new message.

Fixes #4749


jtbandes:
Please make the PR title/commit message something that describes what change is being made – if the title is "Weird UI when hovering over message in RawMessagePanel" then it sounds like this PR is introducing some weird UI when hovering on a message.

snosenzo:
my bad 😅 , I really need to remember to change the title in vscode before submitting

packages/studio-base/src/panels/RawMessages/Value.tsx
```diff
@@ -37,8 +37,14 @@ const StyledIconButton = withStyles(HoverableIconButton, (theme) => ({
     },
   },
 }));
+const useStyles = makeStyles<{ isHovering?: boolean }>()((_theme, { isHovering }) => ({
```
foxymiles, line 40: It's fine to use a param like this but you can also use the `cx` function returned by `useStyles` to apply classes conditionally. I tend to prefer the latter but I think it's mostly just a stylistic difference.

foxymiles (approved):
Seems like a reasonable solution. 👍

packages/studio-base/src/panels/RawMessages/Value.tsx
snosenzo: updated, I should be using this to be consistent. It adds a bit more readability in the css styles in the DOM. 
 - using classNames: only difference between using the hidden classname and non-hidden is a random id (`d4dwh-actionContainer` vs `abcde-actionContainer`) though it does reuse that for the same css values, so each element doesn't have a different one
 - using cx also reuses (what I wanted to make sure) and shows something like `ab4de-actionContainer` vs `ab4de-actionContainer-hidden` which is definitely nicer and more clear

defunctzombie:
> * So I ultimately went for rendering the buttons always but setting their visiblity to 'hidden' to make the elements resize as if they were always there. This does however force `getValueActionForValue` to be called for all values regardless of their hover state.

Is this expensive in the current implementation? If it is - you can use a placeholder/skeleton action icon until the actual action icon is known during hover. If this is not expensive then maybe ok not to do? Generally I think it is nice if we can avoid doing work we don't need in the panels since they are already asked to do a lot during rendering.

snosenzo:
yeah it looks like it's adding about 10ms onto render time (14ms -> 25ms). Going to add skeleton/placeholder, but unfortunately I'll have to default to the largest number of action buttons for placeholder spacing since that information isn't available until hover.

packages/studio-base/src/panels/RawMessages/Value.tsx
```diff
@@ -158,6 +177,18 @@ export default function Value(props: ValueProps): JSX.Element {
     valueAction,
   ]);
 
+  // need to keep space to prevent resizing and wrapping on hover
+  const placeholderActionsForSpacing = useMemo(() => {
+    const actions: ValueActionItem[] = [];
+    let i = availableActions.length;
```
defunctzombie, line 183: I think a `for` loop feels a bit more natural here:

`for (let i=availableActions.length; i<MAX_ACTION_ITEMS; ++i)`

defunctzombie (approved):

PR #4764, opened 2022-11-06
Raise dependabot.yml limit to 5

defunctzombie:
**User-Facing Changes**
None

**Description**
Raise the number of allowed dependabot updates to 5 so we can feel out the waters with a few more dependabot PRs.


jtbandes (approved):
🙏

Issue #4763, opened 2022-11-05
Map panel breaks if robot loses GPS fix


wojtekbacza-coco:
**Description**
If robot loses GPS fix at any point in time, the map panel will go blank when it loads this data point, making us unable to plot any robot path where GPS was lost even for 1 second.
Here is an example `sensor_msgs/NavSatFix` message that causes the problem:
```yaml
header: 
  seq: 1
  stamp: 
    secs: 1667575031
    nsecs: 618148088
  frame_id: "/gps"
status: 
  status: -1
  service: 1
latitude: nan
longitude: nan
altitude: nan
position_covariance: [20.25, 0.0, 0.0, 0.0, 20.25, 0.0, 0.0, 0.0, 81.0]
position_covariance_type: 1
```
This message is coming from [https://wiki.ros.org/nmea_navsat_driver](https://wiki.ros.org/nmea_navsat_driver) GPS driver. When we filtered out all messages with `header.status.service==-1` then the rest of the path got loaded and displayed correctly.

- Version: 1.30.0
- OS: Ubuntu (both in studio and web)
- Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native): Bag file
- ROS distro (if applicable): noetic

**Steps To Reproduce**

Load any bag with aforementioned message, the Map panel will go blank.
Attached are 2 bags, [map_panel_not_working.zip](https://github.com/foxglove/studio/files/9942714/map_panel_not_working.zip) - shows short GPS path that causes map panel not to load, [no_fix_messages_filtered.zip](https://github.com/foxglove/studio/files/9942715/no_fix_messages_filtered.zip) - same path, but with messages with `header.status.service==-1` filtered out - on this bag the panel loads correctly.

**Expected Behavior**
Path is displayed correctly, ignoring messages without GPS fix.

**Actual Behavior**
Map panel goes blank

foxymiles:
Thanks for the detailed bug report. It looks like for those lost GPS messages the position currently comes through as `NaN`, which the panel isn't prepared to handle.

<img width="795" alt="Screen Shot 2022-11-07 at 8 06 22 AM" src="https://user-images.githubusercontent.com/93935560/200200351-3d0bb48b-854f-4809-bf7f-ba6372a6dc19.png">

From my reading of the ROS spec it sounds like technically this is not allowed but I think we can add a check in the Map panel to handle this.

http://docs.ros.org/en/melodic/api/sensor_msgs/html/msg/NavSatFix.html

PR #4765, opened 2022-11-07
Make map panel NavSatFix message handling more robust

foxymiles:
**User-Facing Changes**
Make message handling in the map panel more robust.

**Description**
Add extra validation to incoming map panel messages to discard messages with missing lat/long values or without a position fix.

Fixes https://github.com/foxglove/studio/issues/4763

### DIFF
--- a/packages/studio-base/src/panels/Map/MapPanel.tsx
+++ b/packages/studio-base/src/panels/Map/MapPanel.tsx
@@ -4,14 +4,14 @@
 
 import produce from "immer";
 import {
-  Map as LeafMap,
-  TileLayer,
-  LatLngBounds,
   CircleMarker,
   FeatureGroup,
-  LayerGroup,
   geoJSON,
+  LatLngBounds,
   Layer,
+  LayerGroup,
+  Map as LeafMap,
+  TileLayer,
 } from "leaflet";
 import { difference, groupBy, isEqual, minBy, partition, union } from "lodash";
 import { useCallback, useEffect, useLayoutEffect, useMemo, useRef, useState } from "react";
@@ -19,7 +19,7 @@ import { useResizeDetector } from "react-resize-detector";
 import { useDebouncedCallback } from "use-debounce";
 
 import { toSec } from "@foxglove/rostime";
-import { PanelExtensionContext, MessageEvent, SettingsTreeAction } from "@foxglove/studio";
+import { MessageEvent, PanelExtensionContext, SettingsTreeAction } from "@foxglove/studio";
 import Stack from "@foxglove/studio-base/components/Stack";
 import FilteredPointLayer, {
   POINT_MARKER_RADIUS,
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/panels/Map/MapPanel.tsx
```diff
@@ -47,6 +47,25 @@ function isGeoJSONMessage(msgEvent: MessageEvent<unknown>): msgEvent is GeoJsonM
   );
 }
 
+/**
+ * Verify that the message is either a GeoJSON message or a NavSatFix message with a
+ * position fix and finite latitude and longitude so we can actually display it.
+ */
+function isValidMapMessage(msgEvent: MessageEvent<unknown>): msgEvent is MapPanelMessage {
```
snosenzo, line 54: do we want to be able to add a topic error for any of these to notify the user?

snosenzo (approved):

packages/studio-base/src/panels/Map/MapPanel.tsx
foxymiles: Good question. At least on the criteria we're filtering on here we wouldn't consider these messages in error, just not messages we can display.

snosenzo: yeah that makes sense. should be fine for now

packages/studio-base/src/panels/Map/MapPanel.tsx
```diff
@@ -47,6 +47,25 @@ function isGeoJSONMessage(msgEvent: MessageEvent<unknown>): msgEvent is GeoJsonM
...
+    return true;
+  }
+
+  const message = msgEvent.message as Partial<NavSatFixMsg>;
+  return (
+    message.latitude != undefined &&
+    isFinite(message.latitude) &&
+    message.longitude != undefined &&
+    isFinite(message.longitude) &&
+    message.status?.status !== NavSatFixStatus.STATUS_NO_FIX
```
defunctzombie, line 65: We don't have a `status` field on the `foxglove.LocationFix` message: https://foxglove.dev/docs/studio/messages/location-fix. I suppose this `status?` syntax will still be ok with this.

One approach we could take here is to normalize the incoming messages (like we do for the log panel). This allows us to tailor the handling to each specific type of message. So for a ROS message we would be checking the status but for foxglove messages we would not. Right now this logic assumes the message is type `NavSatFixMsg` but that assumption is not right and generally we should move away from "assume ROS" in our panels.

foxymiles: Makes sense. Should I do a message normalization followup?

defunctzombie: > Makes sense. Should I do a message normalization followup?

I think it would put us in a better place yea. If it feels like too much of a lift then maybe we wait for the next related bugfix.

foxymiles: No I don't think it's too big an effort. I'll open a followup ticket.

Issue #4484, opened 2022-09-26
Cannot toggle visibility for `/robot_description` topic


jhurliman:
**Description**

From a user report, playing `/robot_description` over a ROS2 native connection shows the topic in the 3D panel but the visibility icon does nothing when clicked (it does not toggle from hidden to visible).

https://user-images.githubusercontent.com/195374/192342352-ddab1832-36bd-46c8-942e-9c030bead5cd.mp4

defunctzombie:
@jhurliman Do we have an mcap file or bag file with a `/robot_description` topic we can use to reproduce this? Or is this specific to live connections?

jhurliman:
@defunctzombie I'm not aware of any recordings with that topic. The user report was about a live connection.

defunctzombie:
John: We support receiving urdf over a parameter and a topic. If you receive it as a parameter we show it in the UI as a "topic" but that topic can't be toggled because there is no way to unsubscribe from a parameter. This results in "broken" visibility toggle because there is nothing to toggle.

For the parameter we should show the URDF in some other location and not in the topic list.

amacneil:
Product review feedback:

- Specifically look for `/robot_description` parameter, and if it exists just show it as a Topic named `/robot_description (parameter)`.
- Fix the bug where visibility toggle is not working.

PR #4779, opened 2022-11-08
[3D] Show /robot_description parameter separately from topic

jtbandes:
**User-Facing Changes**
[3D] Fixed some issues with settings for URDFs from the `/robot_description` parameter and/or topic.

**Description**
Fixes https://github.com/foxglove/studio/issues/4484

- Parameter is shown under Topics, displayed as `/robot_description (parameter)`, with settings stored under a `param:/robot_description` key
- Make sure `renderable.userData.settings` is updated
- Fix toggling param visibility off and back on

Example URDF on /robot_description topic (ROS 2) following this tutorial: https://docs.ros.org/en/galactic/Tutorials/Intermediate/URDF/Using-URDF-with-Robot-State-Publisher.html

https://user-images.githubusercontent.com/14237/200459036-651dce51-8076-4d33-9712-b2880e567acd.mov

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
@@ -48,6 +48,7 @@ const TOPIC_NAME = "/robot_description";
 
 /** ID of fake "topic" used to represent the /robot_description parameter */
 const PARAM_KEY = "param:/robot_description";
+/** Standard parameter name used for URDF data in ROS */
 const PARAM_NAME = "/robot_description";
 const PARAM_DISPLAY_NAME = "/robot_description (parameter)";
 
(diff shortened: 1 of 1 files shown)


snosenzo (approved):
couple questions for context:
I assume there can't be more than one URDF topic or parameter in a file?
Is it bad to have a `/robot_description` parameter and topic? should there be a warning?

not sure I have total context of parameters vs topic, hence the questions, but looks good otherwise!

jtbandes:
> I assume there can't be more than one URDF topic or parameter in a file?

Params are only supported on live connections (for now). But yes, like topics, they are assumed to have unique names.

> Is it bad to have a /robot_description parameter and topic? should there be a warning?

From my brief research it seems like most people use the param for ROS 1, and the topic for ROS 2 (because params aren't global in ROS 2). I don't think there is anything technically wrong with having both, so this code tries to support it

PR #4784, opened 2022-11-09
Add Amplitude event tracking for data source selection

esthersweon:
**User-Facing Changes**
- None

**Description**
- Add metrics for data source selections in the Welcome dialog: Local, remote, live connection (Rosbridge, FG WebSocket, ROS 1, ROS 2, Velodyne), or sample data

### DIFF
--- a/packages/studio-base/src/components/OpenDialog/Start.tsx
+++ b/packages/studio-base/src/components/OpenDialog/Start.tsx
@@ -15,8 +15,10 @@ import { useMemo } from "react";
 import { AppSetting } from "@foxglove/studio-base/AppSetting";
 import Stack from "@foxglove/studio-base/components/Stack";
 import TextMiddleTruncate from "@foxglove/studio-base/components/TextMiddleTruncate";
+import { useAnalytics } from "@foxglove/studio-base/context/AnalyticsContext";
 import { usePlayerSelection } from "@foxglove/studio-base/context/PlayerSelectionContext";
 import { useAppConfigurationValue } from "@foxglove/studio-base/hooks";
+import { AppEvent } from "@foxglove/studio-base/services/IAnalytics";
 
 import ActionList, { ActionListItem } from "./ActionList";
 import { OpenDialogViews } from "./types";
@@ -95,6 +97,7 @@ export default function Start(props: IStartProps): JSX.Element {
     onSelectView,
   } = props;
   const { recentSources, selectRecent } = usePlayerSelection();
+  const analytics = useAnalytics();
 
   const [showOnStartup = true, setShowOnStartup] = useAppConfigurationValue<boolean>(
     AppSetting.SHOW_OPEN_DIALOG_ON_STARTUP,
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/components/OpenDialog/Start.tsx
```diff
@@ -118,7 +121,10 @@ export default function Start(props: IStartProps): JSX.Element {
             <path d="M1955 1533l-163-162v677h-128v-677l-163 162-90-90 317-317 317 317-90 90zM256 1920h1280v128H128V0h1115l549 549v475h-128V640h-512V128H256v1792zM1280 512h293l-293-293v293z" />
           </SvgIcon>
         ),
-        onClick: () => onSelectView("file"),
+        onClick: () => {
+          onSelectView("file");
+          void analytics.logEvent(AppEvent.DATA_LOCAL);
```
amacneil, line 126: We need a strategy for naming our events (and document it somewhere).

Most/all of the current ones are using "NOUN_VERB" I don't know why we decided they should be all caps, but here we are. Anyway, `DATA_LOCAL` is meaningless and does not follow the current pattern.

esthersweon: I see - figured "DATA_LOCAL" was preferable to "DATA_CLICK_LOCAL".

Maybe a better alternative is to combine all these under one "DATASOURCE_CLICK" event and pass in the type (local, remote, live, or sample) with other relevant data as necessary (i.e. "rosbridge", "ros1-socket", etc. for the live option).

Will document in Notion once we confirm.

defunctzombie:
We already track player creation events: https://analytics.amplitude.com/foxglove/chart/37jd2mx what does this give us that this tracking is not providing?

defunctzombie (approved):
👍 to adding the tracking code in Start.tsx (seems to be tracking which data sources we are selecting), 👎 to adding the code in Connection.tsx. I think we already have that covered so we can remove that from this PR.

packages/studio-base/src/components/OpenDialog/Start.tsx
```diff
@@ -130,7 +136,10 @@ export default function Start(props: IStartProps): JSX.Element {
           </SvgIcon>
         ),
         iconProps: { iconName: "FileASPX" },
-        onClick: () => onSelectView("remote"),
+        onClick: () => {
+          onSelectView("remote");
+          void analytics.logEvent(AppEvent.DATASOURCE_CLICK, { type: "remote" });
```
defunctzombie, line 141: I'd name this something like `OPEN_DIALOG_TYPE_SELECT` or something more clear than `DATASOURCE_CLICK`.

esthersweon: I'll rename to DIALOG_SELECT_DATASOURCE, as we may want to add more "DIALOG_" prefixed metrics around other actions a user can take in that UI.

defunctzombie: Not that "DATASOURCE" is not what is selected here. This is only selecting the view. `DIALOG_SELECT_VIEW` would be more appropriate.

packages/studio-base/src/components/OpenDialog/Start.tsx
jtbandes: > I don't know why we decided they should be all caps, but here we are.

I think it came from our conventions on enum naming in TS code, but if you want the actual values sent to amplitude to be different, that's fine

PR #4785, opened 2022-11-09
Remove "Open file by URL" from Welcome dialog

### LINKED ISSUE #4759: Add "Open from Data Platform" to Studio Welcome Dialog
Image

Would open Console's "Recordings" page.

esthersweon:
**User-Facing Changes**
- Move "Open file by URL" in Welcome dialog to under "Open connection" --> "Remote file"
- Replace with callout to load data from Data Platform (open Foxglove console directly)

Screen Shot 2022-11-15 at 1 11 51 PM
Screen Shot 2022-11-15 at 1 11 43 PM

Resolves #4759


jtbandes:
Are we trying to prevent users from opening remote http files? I don't see anywhere to do that in the UI anymore after this change.

esthersweon:
@jtbandes - That was my understanding when reading the issue and discussing with @amacneil and @ocin-rye – i.e. we want to make the limits of Studio without Data Platform clearer to the user (and more pressing for them to create an account).

jhurliman:
For removing the remote loading of bag/mcap over HTTP(S), I propose we follow a deprecation policy similar to other features in the app:

1. Add a warning banner in the Open remote URL dialog that this feature will be removed in a future release, and here's what you should be doing instead: ...
2. Remove Open remote URL from the UI
3. Remove opening remote files via URL
4. Delete the dead code

jtbandes:
From discussion, we are going to try moving the url flow into the "connection" screen.

amacneil:
There is no plan to remove "open from URL" functionality. It is still accessible by forming the correct URL, but is not a recommended approach (remote URLs are inherently a bad user experience because many visualizations require downloading the entire file).

esthersweon:
@jtbandes @amacneil - Updated PR description with latest.

packages/studio-base/src/components/OpenDialog/Start.tsx
```diff
@@ -128,16 +151,16 @@ export default function Start(props: IStartProps): JSX.Element {
...
             <path d="M256 1920h512v128H128V0h1115l549 549v91h-640V128H256v1792zM1280 512h293l-293-293v293zm128 256q133 0 249 50t204 137 137 203 50 250q0 133-50 249t-137 204-203 137-250 50q-133 0-249-50t-204-137-137-203-50-250q0-133 50-249t137-204 203-137 250-50zm0 1152q21 0 37-14t28-38 21-53 15-57 9-53 6-41h-230q2 14 5 39t10 53 16 58 21 52 27 39 35 15zm126-384q1-32 1-64t1-64q0-63-3-128h-250q-3 65-3 128 0 64 3 128h251zm-638-128q0 32 4 64t12 64h243q-3-64-3-128 0-63 3-128H912q-8 32-12 64t-4 64zm512-512q-19 0-34 15t-27 39-21 53-15 57-10 53-6 39h225q-2-13-6-37t-11-53-16-58-20-54-27-39-32-15zm253 384q3 65 3 128v64q0 32-2 64h242q8-32 12-64t4-64q0-32-4-64t-12-64h-243zm190-128q-43-75-108-131t-145-88q21 52 32 107t19 112h202zm-637-218q-78 32-142 88t-107 130h200q13-111 49-218zm-249 730q42 73 106 129t142 88q-21-51-31-106t-17-111H965zm642 215q77-32 139-87t105-128h-198q-5 51-15 109t-31 106z" />
           </SvgIcon>
         ),
         iconProps: { iconName: "FileASPX" },
+        href: "https://console.foxglove.dev/recordings",
         onClick: () => {
-          onSelectView("remote");
           void analytics.logEvent(AppEvent.DIALOG_SELECT_VIEW, { type: "remote" });
```
2metres, line 164: Should we update this tracking metric to { type: "data-platform" }

packages/studio-base/src/components/OpenDialog/Start.tsx
```diff
@@ -128,16 +151,16 @@ export default function Start(props: IStartProps): JSX.Element {
       },
       {
         key: "open-url",
-        text: "Open file from URL",
-        secondaryText: `Load a file via HTTP(S).\nSupports ${supportedRemoteFiles} files.`,
+        text: "Open remote data...",
```
jtbandes, line 154: What's the thinking with "…" here and not on the other items? I am used to "…" meaning "something will pop up when you click this instead of immediately performing the action". However we've been pretty consistently _not_ using it in our app as far as I know.

esthersweon: Unlike the other options, this will open a separate window outside of the app (to the Console recordings page). Wanted to make that clear.

esthersweon: And I agree with consistency. These instances don't make much sense:

Screen Shot 2022-11-17 at 2 14 34 PM

We also don't use it for the docs links. 

Maybe I can remove it here, and then we can revisit the other instances in another PR?

jtbandes (changes requested):
The remote bag feature doesn't work. I tested with these urls:

https://assets.foxglove.dev/NuScenes-v1.0-mini-scene-0061-f4fbf7b.mcap
https://assets.foxglove.dev/demo.bag

The MCAP file works, but the bag doesn't (for the reason I described in another comment—second player is overwriting the first one). I think you will need logic in the factory that checks the file extension (like how Ros1RemoteBagDataSourceFactory had `supportedFileTypes = [".bag"]`, etc.) and decides which player to create.

esthersweon:
@jtbandes - Confirmed both remote URLs work (bag and mcap), with most recent changes.

packages/studio-base/src/dataSources/RemoteDataSourceFactory.ts
```diff
@@ -0,0 +1,78 @@
...
+  };
+
+  public initialize(args: DataSourceFactoryInitializeArgs): Player | undefined {
+    const url = args.params?.url;
+    if (!url) {
+      return;
+    }
+
+    const urlParts = url.split(".");
+    const fileExtension = urlParts[urlParts.length - 1];
```
jtbandes, line 62: This is implemented differently than the code above on line 33 – let's try to make them the same since they are both trying to split off the path extension.

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
```diff
@@ -438,7 +440,7 @@ export class IterablePlayer implements Player {
         publishersByTopic,
         datatypes,
         name,
-      } = await this._bufferedSource.initialize();
+      } = result;
```
jtbandes, line 443: Irrelevant changes in this file – remove them?

packages/studio-base/src/dataSources/RemoteDataSourceFactory.ts
```diff
@@ -0,0 +1,78 @@
...
+
+class RemoteDataSourceFactory implements IDataSourceFactory {
+  public id = "remote-file";
+  public type: IDataSourceFactory["type"] = "connection";
+  public displayName = "Remote file";
+  public iconName: IDataSourceFactory["iconName"] = "FileASPX";
+  public supportedFileTypes = [".bag", ".mcap"];
+  public description =
+    "Fetch and load pre-recorded ROS 1 (.bag) or MCAP (.mcap) files from a remote location.";
+  public docsLink = "https://foxglove.dev/docs/studio/connection/ros1-bag";
```
jtbandes, line 22: Is this the right docs link now? Doesn't involve mcap – will we need to update the docs?

jtbandes (approved):
Added some final cleanup suggestions but the code looks a lot better!  cc @defunctzombie for a quick review as one of the original authors of the Open Dialog stuff.

jtbandes:
Can we make this a placeholder value in the field, instead of actually pre-filling the value? Because it never makes sense for someone to click "Open" with the default value here (like it does with the rosbridge & websocket sources)

image

packages/studio-base/src/dataSources/RemoteDataSourceFactory.ts
esthersweon: Revised `docsLink` to be `docsLinks` and potentially include more than one URL.

amacneil:
Per Slack thread, I think we should hold off on merging this until we have a clearly defined goal of what we want users to achieve here.

esthersweon:
Merging now, based on project meeting conclusion.

foxymiles:
I reverted the change to the mui theme and just targeted the new fields to reduce the number of visual diffs but otherwise I think this is ok to merge.

packages/studio-base/src/components/OpenDialog/FormField.tsx
```diff
@@ -14,9 +15,18 @@ type Props = {
...
 };
 
+const useStyles = makeStyles()({
+  field: {
+    "& legend": {
+      display: "none",
+    },
+  },
+});
+
```
2metres, line 25: You can also achieve this by `InputProps={{ notched: false }}`

2metres (approved):

packages/studio-base/src/components/OpenDialog/FormField.tsx
esthersweon: "Property 'notched' does not exist on type 'Partial<InputProps>'."

esthersweon:
@foxymiles - Thanks for the UI cleanup – I think this is ready to merge too. The only last TODO was slightly revamping the copy, per @amacneil's feedback – just adjusted slightly to include the words "Foxglove Data Platform". 

I added those edits in a final commit:

Screen Shot 2022-11-29 at 7 50 37 PM

foxymiles:
The rightmost column looks pretty scrunched now. Should we try to give it a bit more space?

esthersweon:
@foxymiles - Revised copy to use "team" that we use in other callouts (instead of "organization"):

Screen Shot 2022-12-05 at 10 14 43 AM

2nd column width now looks good.

PR #4786, opened 2022-11-09
Remove player initialized event

amacneil:
**User-Facing Changes**
None

**Description**
We don't need both "Player Constructed" and "Player Initialized". Of the two, initialized seems to happen about 50% less, so I assume the data is bad. Therefore, keeping "constructed" and removing "initialized".

### DIFF
--- a/packages/studio-base/src/players/AnalyticsMetricsCollector.ts
+++ b/packages/studio-base/src/players/AnalyticsMetricsCollector.ts
@@ -35,12 +35,6 @@ export default class AnalyticsMetricsCollector implements PlayerMetricsCollector
     this.logEvent(AppEvent.PLAYER_CONSTRUCTED);
   }
 
-  public initialized(args?: { isSampleDataSource: boolean }): void {
-    this.logEvent(AppEvent.PLAYER_INITIALIZED, {
-      isSampleDataSource: args?.isSampleDataSource === true,
-    });
-  }
-
   public play(speed: number): void {
     this.logEvent(AppEvent.PLAYER_PLAY, { speed });
   }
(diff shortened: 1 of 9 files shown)


packages/studio-base/src/players/AnalyticsMetricsCollector.ts
```diff
@@ -35,12 +35,6 @@ export default class AnalyticsMetricsCollector implements PlayerMetricsCollector
     this.logEvent(AppEvent.PLAYER_CONSTRUCTED);
   }
 
-  public initialized(args?: { isSampleDataSource: boolean }): void {
-    this.logEvent(AppEvent.PLAYER_INITIALIZED, {
-      isSampleDataSource: args?.isSampleDataSource === true,
```
defunctzombie, line 40: This is the only place we were tracking whether it was sample data. So these will show up like mcap remote file but we won't know that means sample data.

amacneil: No, this is outdated. Since around a year ago the sample data is tracked through a separate `sample-nuscenes` player.

defunctzombie: Not a player but a data source id: https://github.com/foxglove/studio/blob/9916289c4f26b38ae4bf16c5e6f37a6ae4a919e5/packages/studio-base/src/dataSources/SampleNuscenesDataSourceFactory.ts#L14

👍

defunctzombie (approved):

esthersweon:
> Of the two, initialized seems to happen about 50% less, so I assume the data is bad. Therefore, keeping "constructed" and removing "initialized".

@amacneil - I put in some logs, and found that in the case of an unsuccessful player initialization (example below: bad URL to remote MCAP file), the player is still "constructed":

Screen Shot 2022-11-15 at 4 08 36 PM

Our "PLAYER_CONSTRUCTED" metric only tracks user attempts to playback data, not actual successes. It seems like we still need the "PLAYER_INITIALIZED" metric for that.

Issue #4667, opened 2022-10-19
Text Input on Settings Gradient Picker doesn't submit on blur or return key


2metres:
**Description**
Text Input on Settings Gradient Picker doesn't submit on blur or return key, Observed in the Indicator panel

- Version:
- OS: Web, MacOS, Chrome 106
- Data source: Demo Bag

**Steps To Reproduce**

https://user-images.githubusercontent.com/924528/196569728-59d2ada9-b830-4d80-b810-a19fcc8a8b96.mov

**Expected Behavior**
Blur or enter should set value

osaro-imohe:
I'll get this

jtbandes:
@osaro-imohe Since there's been no update on this issue I'm going to remove your assignment for now

snosenzo:
Able to only reproduce partially. Maybe this behavior has changed in the past month?
This is the experience that I am seeing:
 - On changing the value in the gradient picker, the value in the settings is immediately reflective of the change. The gradient picker remains open while the value in the settings is changing.
 - Closing of the gradient picker onBlur is working, but on enter it is not closing (can address in PR)
 - changing the value via the textbox in the gradient picker only updates the settings, but does not update the panel (will address as part of PR). Clicking and dragging over the colorspace to choose a color fixes this.

Let me know if this aligns with your current experience

PR #4810, opened 2022-11-14
Fix color picker text input not updating values

snosenzo:
**User-Facing Changes**
 - Fix color picker text input not updating panels

**Description**
- close color picker on enter key text input
- fix text input values being invalid for consumption by panels by adding `#` to beginning of value string for consumption elsewhere in UI

Fixes: #4667

### DIFF
--- a/packages/studio-base/src/components/SettingsTreeEditor/inputs/ColorPickerControl.tsx
+++ b/packages/studio-base/src/components/SettingsTreeEditor/inputs/ColorPickerControl.tsx
@@ -41,6 +41,7 @@ type ColorPickerInputProps = {
   alphaType: "none" | "alpha";
   value: undefined | string;
   onChange: (value: string) => void;
+  onEnterKey?: () => void;
 };
 
 function isValidHexColor(color: string, alphaType: "none" | "alpha") {
@@ -48,7 +49,7 @@ function isValidHexColor(color: string, alphaType: "none" | "alpha") {
 }
 
 export function ColorPickerControl(props: ColorPickerInputProps): JSX.Element {
-  const { alphaType, onChange, value } = props;
+  const { alphaType, onChange, value, onEnterKey } = props;
 
   const { classes } = useStyles();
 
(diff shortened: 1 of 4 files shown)


packages/studio-base/src/components/SettingsTreeEditor/inputs/ColorPickerControl.tsx
```diff
@@ -106,6 +107,7 @@ export function ColorPickerControl(props: ColorPickerInputProps): JSX.Element {
         }}
         placeholder={alphaType === "alpha" ? "RRGGBBAA" : "RRGGBB"}
         value={editedValue}
+        onKeyDown={(event) => event.key === "Enter" && onEnterKey && onEnterKey()}
```
jtbandes, line 110: nit
```suggestion
        onKeyDown={(event) => event.key === "Enter" && onEnterKey?.()}
```

packages/studio-base/src/components/SettingsTreeEditor/inputs/ColorPickerControl.tsx
```diff
@@ -70,7 +71,7 @@ export function ColorPickerControl(props: ColorPickerInputProps): JSX.Element {
       setEditedValue(newValue);
 
       if (isValidHexColor(newValue, alphaType)) {
-        onChange(newValue);
+        onChange(`#${newValue}`);
```
jtbandes, line 74: This seems like it could be a risky change... Do we have tests covering the places where this gets used elsewhere in the codebase? Is this just making it more consistent between the color picker & editing the text field?

cc @2metres for review

jtbandes (approved):

packages/studio-base/src/components/SettingsTreeEditor/inputs/ColorPickerControl.tsx
snosenzo: yeah the colorpicker already passes a string formatted this way, it's just that the textfield wasn't and passing it to the outside wasn't working because the consumers were expecting a `#` formatted color.

Issue #4806, opened 2022-11-12
Diagnostic detail error color swapped with warning color


wouter-heerwegh:
The error messages are yellow and warnings are red.

- Version: 1.31.0
- OS: Windows 10 and Ubuntu 20.04
- Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native): ROS 1 native and bagfile
- ROS distro (if applicable): noetic

**Steps To Reproduce**
- open a bag file with diagnostics
- add a Diagnostic detail panel and look at a message with error level

![image](https://user-images.githubusercontent.com/20578313/201483370-fc1b18a4-7169-4b44-bf93-13f622e9700c.png)

**Expected Behavior**
Error messages should be red, warnings should be yellow

**Actual Behavior**
Error messages are yellow, warnings are red.

jtbandes:
Thanks for the report! Looks like this may have been broken in #3605.

PR #4820, opened 2022-11-14
Fix diagnostic status message coloring

foxymiles:
**User-Facing Changes**
Fixes message coloring in the diagnostic status panel.

**Description**
Fixes message coloring by level and adds a new story to catch regressions.

Fixes https://github.com/foxglove/studio/issues/4806


jtbandes (approved):
Thanks!

PR #4827, opened 2022-11-15
Add Amplitude metric tracking when users close the Welcome Dialog

esthersweon:
**User-Facing Changes**
None

**Description**
Add Amplitude metrics tracking:
- when users close the Welcome Dialog
- the view they are dismissing when they do so

### DIFF
--- a/packages/studio-base/src/components/OpenDialog/OpenDialog.tsx
+++ b/packages/studio-base/src/components/OpenDialog/OpenDialog.tsx
@@ -8,10 +8,12 @@ import { useCallback, useLayoutEffect, useMemo, useState } from "react";
 import { useMountedState } from "react-use";
 
 import Stack from "@foxglove/studio-base/components/Stack";
+import { useAnalytics } from "@foxglove/studio-base/context/AnalyticsContext";
 import {
   IDataSourceFactory,
   usePlayerSelection,
 } from "@foxglove/studio-base/context/PlayerSelectionContext";
+import { AppEvent } from "@foxglove/studio-base/services/IAnalytics";
 
 import Connection from "./Connection";
 import Remote from "./Remote";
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/services/IAnalytics.ts
```diff
@@ -17,6 +17,7 @@ enum AppEvent {
 
   // Dialog events
   DIALOG_SELECT_VIEW = "DIALOG_SELECT_VIEW",
+  DIALOG_CLOSE = "DIALOG_CLOSE",
```
jtbandes, line 20: "dialog" sounds like it could be any dialog in the app, do we want to name these something more like "open dialog" (which matches the name of the component)?

packages/studio-base/src/components/OpenDialog/OpenDialog.tsx
```diff
@@ -53,6 +55,15 @@ export default function OpenDialog(props: OpenDialogProps): JSX.Element {
     setActiveView(view);
   }, []);
 
+  const analytics = useAnalytics();
+
+  const onModalClose = useCallback(() => {
+    if (onDismiss) {
+      onDismiss();
+      void analytics.logEvent(AppEvent.DIALOG_CLOSE, { view: activeView });
```
jtbandes, line 63:
```suggestion
      void analytics.logEvent(AppEvent.DIALOG_CLOSE, { activeView });
```

jtbandes (approved):

packages/studio-base/src/services/IAnalytics.ts
esthersweon: Unfortunately already shipped the `DIALOG_SELECT_VIEW` metric on Mon, so would rather keep consistent.

For what it's worth, these metrics would ideally go away with our new Welcome screen (provided they prove that it is more successful than our current dialog).

PR #4450, opened 2022-09-20
Remove legacy 3D panel

jhurliman:
**User-Facing Changes**
- The 3D (Legacy) panel has been removed.

**Description**
Resolves #4493 
Resolves #4207
Resolves https://github.com/foxglove/issues/issues/62
Completely removes code related to the "linked global variables" feature.

PR #4829, opened 2022-11-16
Add (disabled) permanent migration of all Legacy 3D panels

jtbandes:
**User-Facing Changes**
None

**Description**
Adds a permanent auto-migration of all Legacy 3D panels to new 3D panels.
This migration is commented out for now. It will be enabled in https://github.com/foxglove/studio/pull/4450.

### DIFF
--- a/packages/studio-base/src/services/LayoutManager/LayoutManager.ts
+++ b/packages/studio-base/src/services/LayoutManager/LayoutManager.ts
@@ -29,6 +29,7 @@ import {
   RemoteLayout,
 } from "@foxglove/studio-base/services/IRemoteLayoutStorage";
 
+import { migratePanelsState } from "../migrateLayout";
 import { NamespacedLayoutStorage } from "./NamespacedLayoutStorage";
 import WriteThroughLayoutCache from "./WriteThroughLayoutCache";
 import computeLayoutSyncOperations, { SyncOperation } from "./computeLayoutSyncOperations";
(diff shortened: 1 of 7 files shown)


snosenzo (approved):
other than the question looks good to me. But I can't say I'm all that familiar with what the caveats would be so might want a more familiar pair of eyes on it

packages/studio-base/src/services/LayoutManager/LayoutManager.ts
```diff
@@ -216,13 +217,14 @@ export default class LayoutManager implements ILayoutManager {
   @LayoutManager.withBusyStatus
   public async saveNewLayout({
     name,
-    data,
+    data: unmigratedData,
```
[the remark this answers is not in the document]
jtbandes: It's already run [here](https://github.com/foxglove/studio/blob/7dd7d8a4b17a27d3a2a649689deaeff6a0df2ce0/web/src/services/IdbLayoutStorage.ts#L59:L59) and [here](https://github.com/foxglove/studio/blob/7d6eaac598012b9135ed73f1bbb9007ad5ea0d35/desktop/renderer/services/NativeStorageLayoutStorage.ts#L53:L53) but I was trying to test by importing a layout from a json file and it wasn't being run immediately. That's why I added this one.

packages/studio-base/src/services/migrateLayout.ts
```diff
@@ -0,0 +1,67 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
defunctzombie, line 1: Arguably not a service? Is there a better place for this to live? Maybe services are generic functionality?

jtbandes: I dunno. Didn't feel right to put it in utils.

defunctzombie: It never does. Oh well. Maybe on the next pass these can go into a folder like `layout` or some such.

Issue #4800, opened 2022-11-10
Add iframe panel


jk-ethz:
I see Foxglove Studio becoming a powerful tool not only for visualization, but for providing a general dashboard functionality.
To that extent, it would be awesome if we could add external sites as dockable "iframe" panels.

As an example, we are integrating [WeTTY](https://github.com/butlerx/wetty) as a web terminal solution that allows us to start and stop our nodes from the web. This would fit nicely as an embedded iframe panel/window within Foxglove Studio.

Just like that, many other applications become possible using iframes. Since we are working locally on the robot, we can fully control all required CORS headers (or even serve it from the same domain via a reverse proxy).
Is a feature like that planned?

jtbandes:
This can be done by creating an extension: https://foxglove.dev/docs/studio/extensions/getting-started

PR #4831, opened 2022-11-16
Add website iframe panel

jk-ethz:
Display external websites as an embedded iframe.

Solves #4800

Allows to embed great things, such as terminals or code editors.

image

### DIFF
--- a/packages/studio-base/src/panels/Website/config.ts
+++ b/packages/studio-base/src/panels/Website/config.ts
@@ -1,0 +1,39 @@
+// (c) jk-ethz, https://github.com/jk-ethz
+// All rights reserved.
+
+import { SettingsTreeNodes } from "@foxglove/studio";
+
+// Persisted panel state
+export type WebsiteConfig = {
+  url: string;
+};
+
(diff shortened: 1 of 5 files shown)


jk-ethz:
> @jk-ethz Thanks for putting this together. I think we should add this to the public extension marketplace rather than a built-in panel in the Studio source tree.

Why do you think so? Isn't an iframe panel something very basic that doesn't add much weight to the core bundle size?

defunctzombie:
@jk-ethz yep! If you make a PR to: https://github.com/foxglove/studio-extension-marketplace your extension will show up in the extension marketplace sidebar.

jk-ethz:
> @jk-ethz yep! If you make a PR to: https://github.com/foxglove/studio-extension-marketplace your extension will show up in the extension marketplace sidebar.

Cool, will do that once I have the time!

jk-ethz:
When I try to install an extension, it tells me `Download the desktop app to use marketplace extensions.`
Is that something on the todo to get it working in the web app as well?

PR #4833, opened 2022-11-16
Support parsing Uint8Arrays from flatbuffer encoded messages

snosenzo:
**User-Facing Changes**
- support 'flatbuffers' encoding (in addition to 'flatbuffer') to be consistent with `flatbuffers` library

**Description**
 - flatbuffers and flatbuffer are used mostly interchangably in code, but to be consistent with the library name and prevent headaches in the future it is probably wise for us to support both

### DIFF
--- a/packages/mcap-support/package.json
+++ b/packages/mcap-support/package.json
@@ -34,7 +34,7 @@
     "@foxglove/wasm-bz2": "0.1.0",
     "@protobufjs/base64": "1.1.2",
     "flatbuffers": "2.0.6",
-    "flatbuffers_reflection": "0.0.1",
+    "flatbuffers_reflection": "patch:flatbuffers_reflection@0.0.1#../../patches/flatbuffers_reflection.patch",
     "protobufjs": "7.1.2",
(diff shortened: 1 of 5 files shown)


jtbandes:
Hm, the website also says "FlatBuffer Internals" and "A FlatBuffer is a binary file and in-memory format...". So maybe it's not so clear-cut which one is right.

snosenzo:
Yeah, good point. Our original in @foxglove/schemas was to keep it consistent with the import `flatbuffers`, but I think having it be singular (referring to flatbuffer as a file that is in this binary format rather than the library) as the encoding makes sense. I'll leave it as only supporting 'flatbuffers', and if there's an issue in the future we can update across the board.

jtbandes:
So this PR is now about some uint8array support rather than encoding names?

snosenzo:
okay actually going to close this out for now, but I'll reopen a PR for bumping the version of `flatbuffers_reflection` when this lands and is deployed to npm: https://github.com/jkuszmaul/flatbuffers_reflection/pull/24

Issue #4834, opened 2022-11-16
ws-protocol player loses topic subscriptions made when the socket is disconnected


jhurliman:
Compare the ROS 1 native player vs the ws-protocol player. ROS 1:

```
  public setSubscriptions(subscriptions: SubscribePayload[]): void {
    this._requestedSubscriptions = subscriptions;

    if (!this._rosNode || this._closed) {
      return;
    }
```

If the connection is currently closed, it will early exit but first it stores the list of requested subscriptions so they can be made upon reconnect. Compare with the ws-protocol player:

```
  public setSubscriptions(subscriptions: SubscribePayload[]): void {
    if (!this._client || this._closed) {
      return;
    }
```

If the socket is currently closed, the `subscriptions` parameter is discarded and the method early returns. You can reproduce this bug by opening a socket connection, killing the server, setting a Raw Messages panel to a new topic, then restarting the server. The socket will open but data won't come in on that topic until you take a user action to resubscribe.

foxymiles:
A fast way to repro this:

1. create a layout with a single datasource info panel
2. create a layout with a single raw messages panel subcribed to `/example_msg`
3. run the python json example server from https://github.com/foxglove/ws-protocol/tree/main/python and connect to it in studio
4. select the first layout
5. kill the python server
6. select the second layout
7. restart the python server

You'll see that the raw messages panel doesn't resume showing messages even though the connection has been restored.

PR #4836, opened 2022-11-17
Retry subscription when websocket connection is established

foxymiles:
**User-Facing Changes**
Fixes resubscribing to topics when a ws-protocol connection is broken and reestablished.

**Description**
Currently when we receive subscription requests while the player is disconnected we simply discard them. This makes an effort to remember the requested topics and re-subscribe once the connection is established, in a similar manner to the ROS1 player.

Fixes https://github.com/foxglove/studio/issues/4834


jtbandes (approved):
Makes sense, and kinda matches the rosbridge player

jhurliman (approved):

Issue #4788, opened 2022-11-09
ws-protocol connector displays an error when a channel is advertised more than once


jhurliman:
**Description**

There shouldn't be anything wrong with a ws-protocol server sending the same advertisement more than once in a session. Disallowing this means the server has to do additional bookkeeping.

**Steps To Reproduce**

1. Connect Studio to a recent build of https://github.com/foxglove/ros-foxglove-bridge
2. Do something to change the topic graph, such as publishing a new topic via `ros2 topic pub`
3. If the issue hasn't triggered yet, change the topic graph again by stopping the publisher

**Expected Behavior**

No error reported by Studio.

**Actual Behavior**

<img width="713" alt="Screen Shot 2022-11-09 at 2 19 59 PM" src="https://user-images.githubusercontent.com/195374/200954344-5aca0b64-303b-47fe-905d-f77253a2ba93.png">

PR #4837, opened 2022-11-17
Fix erroneous duplicate channel warning in websocket player

foxymiles:
**User-Facing Changes**
Fixes an extraneous warning shown for some websocket datasources.

**Description**
The existing logic adds a player problem if we see an advertised channel twice. We really only need to show this if two different channels are advertised on the same topic. The change here is to compare the new and previous channels to catch any changes in the channel id or schema name. 

I didn't include the content of the schemas in this equality check since they could be larger objects but maybe they should be included for correctness?

Fixes https://github.com/foxglove/studio/issues/4788

### DIFF
--- a/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
+++ b/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
@@ -3,23 +3,24 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
 import * as base64 from "@protobufjs/base64";
+import { isEqual } from "lodash";
 import { v4 as uuidv4 } from "uuid";
 
 import { debouncePromise } from "@foxglove/den/async";
 import Log from "@foxglove/log";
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
```diff
@@ -36,6 +36,14 @@ const ZERO_TIME = Object.freeze({ sec: 0, nsec: 0 });
 
 type ResolvedChannel = { channel: Channel; parsedChannel: ParsedChannel };
 
+function areChannelsEqual(channelA: Readonly<Channel>, channelB: Readonly<Channel>) {
```
jtbandes, line 39: I'd say either compare the schemas or call this `areChannelsMostlyEqual` 😄

jtbandes (approved):

foxymiles:
@jtbandes I switched it to use lodash `isEqual` instead, which still seems to resolve the issue in my ros2 testing.

PR #4855, opened 2022-11-18
Remove InitialRefreshAdapter from ExtensionCatalogProvider

defunctzombie:
**User-Facing Changes**
None

**Description**
Roll this into the ExtensionCatalogProvider component logic directly. I find this a bit cleaner to reason about.

### DIFF
--- a/packages/studio-base/src/providers/ExtensionCatalogProvider.tsx
+++ b/packages/studio-base/src/providers/ExtensionCatalogProvider.tsx
@@ -12,7 +12,6 @@ import {
   ExtensionCatalog,
   ExtensionCatalogContext,
   RegisteredPanel,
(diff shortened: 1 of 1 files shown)


defunctzombie:
@foxymiles tagging you for review since you wrote this code. If you feel strongly that the adapter approach is cleaner I'm happy to drop this PR. I personally found this easier to understand because it removed a layer of indirection around the initial store setup.

foxymiles:
There were some tricky issues around startup and panels being loaded twice or not at all related to this code. I'll need to double check that all that still works.

foxymiles:
I did some extra testing and this seems to work fine. I agree it's simpler this way.

foxymiles (approved):

Issue #4854, opened 2022-11-18
Map panel incorrectly calls setState on another component while rendering


defunctzombie:
**Description**
<img width="670" alt="image" src="https://user-images.githubusercontent.com/84792/202329915-5a87fef0-464c-4c21-bfeb-8300787cf768.png">

This occurs when the resizeDetector fires the `onResize` handler:
https://github.com/foxglove/studio/blob/main/packages/studio-base/src/panels/Map/MapPanel.tsx#L156-L167

Which results in invalidating the lealflet map and triggering the move/zoom handlers:
https://github.com/foxglove/studio/blob/main/packages/studio-base/src/panels/Map/MapPanel.tsx#L600-L611

These handlers call `saveState` which under-the-hood invokes functions on the layout provider.

PR #4858, opened 2022-11-20
Fix react render error in map panel

foxymiles:
**User-Facing Changes**
None

**Description**
Fixes a react render warning in the map panel by using a useEffect that depends on the size returned by the resize observer to invalidate the map size instead of trying to write the new size into the layout directly in the resize handler.

Fixes https://github.com/foxglove/studio/issues/4854

### DIFF
--- a/packages/studio-base/src/panels/Map/MapPanel.tsx
+++ b/packages/studio-base/src/panels/Map/MapPanel.tsx
@@ -157,8 +157,8 @@ function MapPanel(props: MapPanelProps): JSX.Element {
   // an existing resize observation.
   // https://github.com/maslianok/react-resize-detector/issues/45
   const {
-    width: mapWidth,
-    height: mapHeight,
+    width: panelWidth,
+    height: panelHeight,
     ref: sizeRef,
   } = useResizeDetector({
     refreshRate: 0,
@@ -166,11 +166,11 @@ function MapPanel(props: MapPanelProps): JSX.Element {
   });
 
   useEffect(() => {
-    // We depend on changes in the resized dimensions to tell the Leaflet map to
+    // We depend on changes in the resized panel dimensions to tell the Leaflet map to
     // recalculate its size.
-    void { mapWidth, mapHeight };
+    void { panelWidth, panelHeight };
     currentMap?.invalidateSize();
-  }, [mapWidth, mapHeight, currentMap]);
+  }, [panelWidth, panelHeight, currentMap]);
 
   // panel extensions must notify when they've completed rendering
   // onRender will setRenderDone to a done callback which we can invoke after we've rendered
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/panels/Map/MapPanel.tsx
```diff
@@ -153,19 +153,25 @@ function MapPanel(props: MapPanelProps): JSX.Element {
...
+  } = useResizeDetector({
     refreshRate: 0,
     refreshMode: "debounce",
-    onResize,
   });
 
+  useEffect(() => {
+    // We depend on changes in the resized panel dimensions to tell the Leaflet map to
+    // recalculate its size.
+    void { panelWidth, panelHeight };
```
defunctzombie, line 171: As a reader I don't understand why this effect depends on `panelWidth` and `panelHeight` only then to not do anything with those variables. I suspect what you are going for here is that we need to invoke `invalidateSize()` after the _render_ of our component rather than during which seems to be happening with the existing `onResize` behavior from `useResizeDector`. If we want to use this workaround the comment should be more clear on why we are avoiding `onResize` in `useResizeDetector`.

I also looked into the code for useResizeDetector and it _looks_ like it calls onResize inside of a `setState` call: https://github.com/maslianok/react-resize-detector/blob/master/src/useResizeDetector.ts#L41 https://github.com/maslianok/react-resize-detector/blob/432c334ce0744cccb32dc825f7f4ac30696d9e77/src/utils.ts#L30 which I would have thought runs at a different time than the rendering of a component?

defunctzombie (approved):
I'm ok with this workaround if the comment is updated to be more clear around the gotchas we are avoiding and why the weird looking effect with a no-op variable is used.

It does look to me like this should already be handled properly in the library but maybe it's not running when I think it should be with the setState

packages/studio-base/src/panels/Map/MapPanel.tsx
jtbandes: Having a side effect inside of a setState callback does seem wrong...the callback is not called after the new state takes effect, it's called in order to compute the new state. Those callbacks are supposed to be idempotent and may be called multiple times in strict mode.

foxymiles: Another solution here is to put `panelWidth` and `panelHeight` in the dependency list for the hook, but then you have to use an eslint override and lose the benefit of eslint tracking that the dependencies are correct. IMO this is ok because we're sycing two pieces of state, the width of the panel and the map size, which is what useEffect is for. It just so happens that Leaflet doesn't require the width and height as parameters to the `invalidate` call but it does implicitly depend on them.

jtbandes (approved):

foxymiles:
@defunctzombie I've updated the comment to make it more clear why this is needed.

Issue #4846, opened 2022-11-17
Link 3D Panels


banisadr:
Ability to sync or "lock" multiple 3D panels together in a layout such that pan/zoom in one does the same in the other linked panels. Needs both the ability to do 1-time sync of panels as well as ability to lock panels together.

foxymiles:
Do we want to sync the complete camera state or just pan zoom? If we only sync pan and zoom that might not make sense for different views with different initial camera states right?

foxymiles:
Also if there is a mix of 3d and 2d views in a layout how should we sync them?

foxymiles:
Screenshot of the feature in the old 3d panel:

<img width="358" alt="image" src="https://user-images.githubusercontent.com/93935560/202965004-415c4f0e-1370-4892-902b-9afef78e2927.png">

jtbandes:
Assigning to @snosenzo to get #4859 over the finish line with the changes requested in https://github.com/foxglove/studio/pull/4859#issuecomment-1322801204

Issue #2081, opened 2021-11-04
Allow orthographic projection for more than just z axis


clementchef:
Currently the 2d/3d button in 3d pannel only works for z axis. 
Sometime the user would love to project along x, -x, y, -y.
Might be nice to provide some choice for the projection view.

PR #4859, opened 2022-11-21
Sync 3d panel camera states

foxymiles:
**User-Facing Changes**
Adds a new panel setting to sync camera states across multiple 3d panels.

**Description**
This adds a new setting and new panel extension context features to sync the camera states of multiple 3d panels. This is accomplished via a new `sharedPanelState` property in the `PanelExtensionContext` `RenderState`.

Panels that wish to use shared state can watch `sharedPanelState` in the render state and call `setSharedPanelState` to update the shared state, which is typed as an opaque record type.

```typescript
    setSharedPanelState: (state: undefined | Readonly<string, unknown>) => void;
```

The shared state is scoped to the panel type, so, for example, the 3d panel will only see shared state for the 3d panel.

Screenshot 2022-11-21 at 12 08 13 PM

https://user-images.githubusercontent.com/93935560/202955310-91eeb167-b6b5-4bd9-a21c-74d410e781d7.mov

Fixes https://github.com/foxglove/studio/issues/4846

### DIFF
--- a/packages/studio-base/src/components/PanelExtensionAdapter/renderState.ts
+++ b/packages/studio-base/src/components/PanelExtensionAdapter/renderState.ts
@@ -25,16 +25,17 @@ const log = Log.getLogger(__filename);
 const EmptyParameters = new Map<string, ParameterValue>();
 
 type BuilderRenderStateInput = {
-  watchedFields: Set<string>;
-  playerState: PlayerState | undefined;
   appSettings: Map<string, AppSettingValue> | undefined;
-  currentFrame: MessageEvent<unknown>[] | undefined;
   colorScheme: RenderState["colorScheme"] | undefined;
+  currentFrame: MessageEvent<unknown>[] | undefined;
   globalVariables: GlobalVariables;
   hoverValue: HoverValue | undefined;
+  messageConverters?: RegisterMessageConverterArgs<unknown>[];
+  playerState: PlayerState | undefined;
+  sharedPanelState: Record<string, unknown> | undefined;
   sortedTopics: readonly PlayerTopic[];
   subscriptions: Subscription[];
-  messageConverters?: RegisterMessageConverterArgs<unknown>[];
+  watchedFields: Set<string>;
 };
 
 type BuildRenderStateFn = (input: BuilderRenderStateInput) => Readonly<RenderState> | undefined;
@@ -64,6 +65,7 @@ function initRenderStateBuilder(): BuildRenderStateFn {
   let prevSubscriptions: BuilderRenderStateInput["subscriptions"];
   let prevSortedTopics: BuilderRenderStateInput["sortedTopics"] | undefined;
   let prevMessageConverters: BuilderRenderStateInput["messageConverters"] | undefined;
+  let prevSharedPanelState: BuilderRenderStateInput["sharedPanelState"];
 
   // Topics which we are subscribed without a conversion, these are topics we want to receive the original message
   const topicNoConversions: Set<string> = new Set();
@@ -78,16 +80,17 @@ function initRenderStateBuilder(): BuildRenderStateFn {
 
   return function buildRenderState(input: BuilderRenderStateInput) {
     const {
-      playerState,
-      watchedFields,
       appSettings,
-      currentFrame,
       colorScheme,
+      currentFrame,
       globalVariables,
       hoverValue,
-      sortedTopics,
       messageConverters,
+      playerState,
+      sharedPanelState,
+      sortedTopics,
       subscriptions,
+      watchedFields,
     } = input;
 
     // If the player has loaded all the blocks, the blocks reference won't change so our message
@@ -141,6 +144,14 @@ function initRenderStateBuilder(): BuildRenderStateFn {
       if (parameters !== renderState.parameters) {
         shouldRender = true;
         renderState.parameters = parameters;
+      }
+    }
+
+    if (watchedFields.has("sharedPanelState")) {
+      if (sharedPanelState !== prevSharedPanelState) {
+        shouldRender = true;
+        prevSharedPanelState = sharedPanelState;
+        renderState.sharedPanelState = { ...sharedPanelState };
       }
     }
 
--- a/packages/studio-base/src/components/PanelExtensionAdapter/renderState.ts
+++ b/packages/studio-base/src/components/PanelExtensionAdapter/renderState.ts
@@ -151,7 +151,7 @@ function initRenderStateBuilder(): BuildRenderStateFn {
       if (sharedPanelState !== prevSharedPanelState) {
         shouldRender = true;
         prevSharedPanelState = sharedPanelState;
-        renderState.sharedPanelState = { ...sharedPanelState };
+        renderState.sharedPanelState = sharedPanelState;
       }
     }
 
--- a/packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
+++ b/packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
@@ -711,7 +711,7 @@ export function ThreeDeeRender({ context }: { context: PanelExtensionContext }):
 
   // Sync camera with shared state, if enabled.
   useEffect(() => {
-    if (!renderer || sharedPanelState == undefined || config.scene.syncCamera === false) {
+    if (!renderer || sharedPanelState == undefined || config.scene.syncCamera !== true) {
       return;
     }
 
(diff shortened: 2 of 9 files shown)


packages/studio/index.d.ts
```diff
@@ -136,6 +136,19 @@ declare module "@foxglove/studio" {
...
+    perspective: boolean;
+    phi: number;
+    target: readonly [number, number, number];
+    targetOffset: readonly [number, number, number];
+    targetOrientation: readonly [number, number, number, number];
+    thetaOffset: number;
+    fovy: number;
+    near: number;
+    far: number;
+  };
```
jtbandes, line 150: I'm not a fan of this being part of the public extension API. This couples the 3D panel implementation to the extension API and prevents us from making changes in the 3D panel camera system later (for example, https://github.com/foxglove/studio/issues/2081). We should strive to make the extension API powerful enough that panels as complex as 3D could be built as third-party extensions.

jtbandes:
> 1. Should we sync the entire camera state or just pan/zoom?
> 2. What should we do if we have a mix of perspective and orthographic views? Sync them separately?

Entire state is how it worked in the old panel so I think that's fine for now. Perspective/ortho mode is part of the camera state, so that should be synced as well.

Sam has a good point regarding follow modes. We should probably sync the follow mode too, or disallow syncing the camera state if a different follow mode/display frame is used, because the syncing wouldn't really make sense.

packages/studio/index.d.ts
foxymiles: Do you have any alternative suggestions for exposing this in a way that other panels can consume it? Mark it as unknown instead and the panels cast it to the known type?

jtbandes:
From Slack discussion, we have consensus on exposing a key-value-like API, similar to globalVariables, but where the variables are internal-only and not displayed in the sidebar. This allows panels to communicate, allows newly appearing panels to read current state, and avoids exposing specific 3D panel types in the extension API.

packages/studio-base/src/providers/PanelStateContextProvider.tsx
```diff
@@ -61,8 +73,43 @@ export function usePanelSettingsTreeUpdate(): (newTree: ImmutableSettingsTree) =
...
+
 export function PanelStateContextProvider({ children }: { children?: ReactNode }): JSX.Element {
   const [store] = useState(createPanelStateStore());
 
+  const currentLayoutId = useCurrentLayoutSelector(selectCurrentLayoutId);
+
+  // clear shared panel state on layout change
+  useEffect(() => {
+    void currentLayoutId;
+    store.setState({ sharedPanelState: {} });
```
snosenzo, line 111: would it also be possible to clear this if there are no longer panels using it? I'm a little worried about it keeping references to panels around if extension panels should choose to somehow store a reference to itself, either intentionally or unintentionally

defunctzombie:
Watching the video I think it might be nice to add an icon into the viewport to indicate which viewports have sync lock turned on?

packages/studio/index.d.ts
defunctzombie: Marking as resolved as this is no longer part of the public extension interface.

packages/studio/index.d.ts
```diff
@@ -261,6 +267,11 @@ declare module "@foxglove/studio" {
      */
     setParameter: (name: string, value: ParameterValue) => void;
 
+    /**
+     * Set the transient state shared by panels of the same type as the caller of this function.
+     */
```
defunctzombie, line 272: Is it worth noting that this will _not_ be saved?

packages/studio-base/src/providers/PanelStateContextProvider.tsx
foxymiles: Good question. I think I can add this without too much extra complexity.

defunctzombie:
These APIs make sense to me.

Once we are in a place where we are happy with the APIs I think it would be good to open a website PR to add the docs so we get a sense of how it reads.

jtbandes:
> `    setSharedPanelState: (state: undefined | Readonly<string, unknown>) => void;`

Should we offer a functional form of `set((old) => new)`? Or maybe this doesn't provide much real world benefit?

foxymiles:
> Should we offer a functional form of `set((old) => new)`? Or maybe this doesn't provide much real world benefit?

Generally I like having that as an option although it does seem a bit weird to use it for this particular case. It's easy enough to add.

packages/studio/index.d.ts
foxymiles: Calleing this out specifically in the comment now.

packages/studio-base/src/providers/PanelStateContextProvider.tsx
foxymiles: I've added some garbage collection logic around shared state to clear it out once there is no longer a panel of the corresponding type using it.

packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
```diff
@@ -819,13 +886,16 @@ export function ThreeDeeRender({ context }: { context: PanelExtensionContext }):
...
 
   const onTogglePerspective = useCallback(() => {
-    setConfig((prevConfig) => ({
-      ...prevConfig,
-      cameraState: { ...prevConfig.cameraState, perspective: !prevConfig.cameraState.perspective },
-    }));
-    // Wait for the setConfig to propagate to the renderer before updating the settings tree
-    setTimeout(() => renderer?.updateCoreSettings(), 0);
-  }, [renderer]);
+    const currentState = renderer?.getCameraState().perspective ?? false;
```
snosenzo, line 889: This does look like a less hacky solution without the timeout, but why this change?
just unclear on when to use `actionHandler` vs `setConfig` and when to use which

snosenzo (approved):
looks good to me, but might want more seasoned eyes on it for some of the `renderState` stuff since I'm a little less familiar

packages/studio-base/src/components/PanelExtensionAdapter/renderState.ts
```diff
@@ -144,6 +147,14 @@ function initRenderStateBuilder(): BuildRenderStateFn {
       }
     }
 
+    if (watchedFields.has("sharedPanelState")) {
+      if (sharedPanelState !== prevSharedPanelState) {
+        shouldRender = true;
+        prevSharedPanelState = sharedPanelState;
+        renderState.sharedPanelState = { ...sharedPanelState };
```
jtbandes, line 154: why shallow copy?

packages/studio-base/src/panels/ThreeDeeRender/renderables/CoreSettings.ts
```diff
@@ -150,6 +150,12 @@ export class CoreSettings extends SceneExtension {
                   ? "This setting requires a restart to take effect"
                   : undefined,
             },
+            syncCamera: {
+              label: "Sync camera",
```
jtbandes, line 154: Can we add help text that briefly describes how the setting works?

packages/studio-base/src/providers/PanelStateContextProvider.tsx
```diff
@@ -61,8 +75,101 @@ export function usePanelSettingsTreeUpdate(): (newTree: ImmutableSettingsTree) =
...
+const selectCurrentLayoutId = (state: LayoutState) => state.selectedLayout?.id;
+
+const selectLayoutConfigById = createHasSameKeySelector(
+  (state: LayoutState) => state.selectedLayout?.data?.configById,
+  (config) => config,
+);
+
+const selectPanelTypesInUse = createSelector(selectLayoutConfigById, (config) => {
+  return uniq(Object.keys(config ?? {}).map(getPanelTypeFromId));
+});
```
jtbandes, line 156: Isn't it possible that there are panel types in use that don't have any saved config, but are using the shared state API?

jtbandes (approved):

packages/studio-base/src/providers/PanelStateContextProvider.tsx
foxymiles: It looks like we still create a (empty) key in `configById` for panels with no settings.

packages/studio-base/src/components/PanelExtensionAdapter/renderState.ts
foxymiles: You mean instead of just passing it through? I guess you're right it's not really necessary. I'll take it out.

PR #4860, opened 2022-11-21
Performance boost for flatbuffer schemas that use Uint8 arrays

snosenzo:
**User-Facing Changes**
 - performance boost for flatbuffer schemas that use Uint8 arrays

**Description**
 - update flatbuffers_reflection to new version that returns Uint8Arrays that are a slice into the buffer
 - prevents normalizing the message from an array to a Uint8Array in panels


packages/mcap-support/src/parseFlatbufferSchema.test.ts
```diff
@@ -355,4 +356,35 @@ describe("parseFlatbufferSchema", () => {
...
+     * }
+     * root_type ByteVector;
+     */
+    const data = ByteVector.createDataVector(builder, [1, 2, 3]);
+    ByteVector.startByteVector(builder);
+    ByteVector.addData(builder, data);
+    const byteVector = ByteVector.endByteVector(builder);
+    builder.finish(byteVector);
+    /** the underlying buffer for the builder is larger than the uint8array of the data
+     * this needs to be cleared so that the reading from the buffer by the parser doesn't use the wrong offsets
```
jtbandes, line 375: which code would end up using the wrong offsets? Is that a bug we can fix? Uint8Array includes offset and length information so it should not be necessary to clone the data in order to set offset=0.

packages/mcap-support/src/parseFlatbufferSchema.test.ts
```diff
@@ -355,4 +356,35 @@ describe("parseFlatbufferSchema", () => {
...
+     * table ByteVector {
+     *   data:[uint8];
+     * }
+     * root_type ByteVector;
+     */
+    const data = ByteVector.createDataVector(builder, [1, 2, 3]);
+    ByteVector.startByteVector(builder);
+    ByteVector.addData(builder, data);
+    const byteVector = ByteVector.endByteVector(builder);
+    builder.finish(byteVector);
```
jtbandes, line 373: Rather than dealing with the generated code and imports you could consider just hard-coding a Uint8Array with the encoded message here. But I guess that wouldn't save you from needing the bfbs file, so maybe not worth it.

jtbandes (approved):

packages/mcap-support/src/parseFlatbufferSchema.test.ts
snosenzo: the code is in the `flatbuffers_reflection` repo, because I encountered it in my test there as well. I couldn't quite pinpoint what part of the offset wasn't right
you can take a look here: https://github.com/jkuszmaul/flatbuffers_reflection/pull/24

PR #4875, opened 2022-11-22
Remove URDF panel

jhurliman:
**User-Facing Changes**
- URDF panel has been removed. URDFs can be visualized and manipulated in the 3D panel

**Description**

This also removes the ability to drag and drop a `.urdf` or `.xacro` file into Studio. That functionality was partially broken; the app would do nothing unless a URDF panel existed in the current tab of the current layout. Bringing drag-and-drop URDF functionality into the 3D panel will require extension API work.

Closes FG-1116


defunctzombie:
Some things to consider:

- This panel is the only one supporting JointState (which we don't have a foxglove message yet)
- Drag/drop URDF only works with this panel but also isn't a workflow we've exposed to any extension api or a workflow we use for loading data other than message data into the app.

jhurliman:
We have `private handleJointState = (messageEvent: PartialMessageEvent<JointState>): void => {` but I don't recall if this is fully hooked up to something the user can see happening or not. It needs testing.

defunctzombie:
We wouldn't have to do this issue if we delete this panel: https://github.com/foxglove/studio/issues/4882 🤣

defunctzombie:
Closing this. We have a project tracking the work to remove the panel and the steps we need to take to get there.

PR #4877, opened 2022-11-22
Make the welcome dialog take up the whole screen

### LINKED ISSUE #4861: Avoid stranding users in a confusing and not-useful no-data state
**Feature**: 
- Avoid stranding users in a confusing and not-useful state in Studio by making the Welcome dialog take up the whole app background (no X to dismiss).
- Users should still be able to dismiss this welcome screen by actively selecting an existing layout or creating a new one.

**Roll out**: 
- We want to A/B test this new UI with our existing dialog. We should do a random split between our users for experiencing the original vs new UI.

**Design**:
Screen Shot 2022-11-22 at 3 40 38 PM

esthersweon:
**User-Facing Changes**
- Make the Welcome dialog take up the whole screen
- Remove x icon for exiting out of screen
- Revise UI to more closely match [Nico's designs](https://www.figma.com/file/4QttcTEtbt70D5GjN0ddcT/Untitled?node-id=0%3A1&t=Y54woQTmWgCVp1ul-0)

Screen Shot 2022-11-22 at 3 49 36 PM

Resolves #4861

------

**TODO**: 
- Figure out instances in which a user should be able to dismiss this view (cc: @ocin-rye @foxymiles)
   - Just selecting layout? Creating a new layout?
   - How about adding a panel? 
- Do we want to preserve the original "Show on startup" app setting?


foxymiles:
It sounds like we still need to make the sidebar accessible while this screen is shown so users can select a layout or data source directly right?

packages/studio-base/src/components/WelcomeScreen/WelcomeScreen.tsx
```diff
@@ -152,37 +156,19 @@ export default function OpenDialog(props: OpenDialogProps): JSX.Element {
...
-        flexGrow={1}
-        flexBasis={450}
-        fullHeight
-        justifyContent="space-between"
-        gap={2}
-        paddingX={5}
-        paddingY={3}
-      >
-        {view.component}
+    <div className={classes.root}>
```
2metres, line 163: Can we use the mui Container compoment here to limit the width?

esthersweon:
Closing out, based on our project meeting. 

We will focus instead on #4862

Issue #4801, opened 2022-11-10
Plot panel memory usage rapidly increases over time, leading to long GC cycles for dense layouts


DavidXenakis:
**Description**
We have been experiencing poor performance and crashes related to runaway memory use. One of the culprits, we believe, is the plot panel. For particularly dense layouts including many plots and series, chrome's performance profile shows memory usage rapidly increasing during playback, even after all data has been loaded and cached. 

Sawtooth memory usage pattern.
<img width="1726" alt="Screen Shot 2022-10-25 at 11 37 13 AM" src="https://user-images.githubusercontent.com/3076080/201206665-09a1b617-692a-4398-bf0c-87b613a294e4.png">

Please investigate whether or not the plot panel's memory footprint can be reduced.

- Version:
- OS: All
- Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native): Any? It repros from the studio.foxglove.dev sample dataset.
- ROS distro (if applicable): n/a

**Steps To Reproduce**
1. Visit https://studio.foxglove.dev/?ds=sample-nuscenes&layoutId=0daa2893-d4b6-4146-a4b7-9285c9637358&time=2018-07-24T03%3A28%3A49.082513669Z
2. Load all the data by playing through the log.
3. Run a performance profile during playback to capture memory usage.

**Expected Behavior**

**Actual Behavior**

defunctzombie:
Layout:

<details>
<pre>
{
  "configById": {
    "Plot!39p1t8k": {
      "title": "Plot",
      "paths": [
        {
          "value": "/gps.latitude",
          "enabled": true,
          "timestampMethod": "receiveTime"
        },
        {
          "value": "/gps.longitude",
          "enabled": true,
          "timestampMethod": "receiveTime"
        },
        {
          "value": "/gps.altitude",
          "enabled": true,
          "timestampMethod": "receiveTime"
        },
        {
          "value": "/gps.position_covariance_type",
          "enabled": true,
          "timestampMethod": "receiveTime"
        },
        {
          "value": "/drivable_area.pose.position.x",
          "enabled": true,
          "timestampMethod": "receiveTime"
        },
        {
          "value": "/drivable_area.pose.position.y",
          "enabled": true,
          "timestampMethod": "receiveTime"
        },
        {
          "value": "/drivable_area.pose.position.z",
          "enabled": true,
          "timestampMethod": "receiveTime"
        },
        {
          "value": "/drivable_area.pose.orientation.x",
[...]

PR #4884, opened 2022-11-23
Plot perf improvements

jtbandes:
**User-Facing Changes**
Improved performance of the Plot panel.

**Description**
- Move the current playback position (vertical bar) out of the canvas, similar to HoverBar, so we don't have to re-render the plot constantly during playback
- Add useMemo in several places to reduce re-rendering during playback and when updating hover time

Relates to #4801

### DIFF
--- a/packages/studio-base/src/components/TimeBasedChart/index.tsx
+++ b/packages/studio-base/src/components/TimeBasedChart/index.tsx
@@ -47,6 +47,7 @@ import { fonts } from "@foxglove/studio-base/util/sharedStyleConstants";
 
 import HoverBar from "./HoverBar";
 import TimeBasedChartTooltipContent from "./TimeBasedChartTooltipContent";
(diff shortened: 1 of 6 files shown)


packages/studio-base/src/components/TimeBasedChart/index.tsx
```diff
@@ -452,9 +441,9 @@ export default function TimeBasedChart(props: Props): JSX.Element {
         },
       },
       ...props.plugins,
-      annotation: { annotations },
+      annotation: { annotations: props.annotations },
     } as ChartOptions["plugins"];
-  }, [currentTime, props.annotations, props.plugins, props.zoom, zoomMode]);
```
jtbandes, line 457: Removing `currentTime` from dependencies here is a major perf win.

foxymiles (approved):

PR #4885, opened 2022-11-23
Plot perf improvement for seeking

jtbandes:
**User-Facing Changes**
Improved Plot panel performance when seeking through preloaded data.

**Description**
Prevent re-rendering the chart while seeking if data is all preloaded in blocks.

Split from #4884 
Relates to #4801 

https://user-images.githubusercontent.com/14237/204933498-b4aadbf4-4baa-4733-bc8b-c562cac9f524.mov

</td><td>

https://user-images.githubusercontent.com/14237/204933514-2cf86af4-d4d2-4d72-8b9d-8dc3b09bd400.mov

</td></tr></table>

### DIFF
--- a/packages/studio-base/src/panels/Plot/index.tsx
+++ b/packages/studio-base/src/panels/Plot/index.tsx
@@ -419,12 +419,17 @@ function Plot(props: Props) {
     ],
   );
 
-  const plotDataByPath = useMessageReducer<PlotDataByPath>({
-    topics: subscribeTopics,
-    preloadType: "full",
-    restore,
-    addMessages,
-  });
+  // The extra useShallowMemo is useful when seeking. Without it, the reduced value will be reset,
+  // and trigger the plot to re-render even if the old value and new value were both empty ({})
+  // which happens for fully pre-loaded data coming from blocks.
+  const plotDataByPath = useShallowMemo(
+    useMessageReducer<PlotDataByPath>({
+      topics: subscribeTopics,
+      preloadType: "full",
+      restore,
+      addMessages,
+    }),
+  );
 
   // Keep disabled paths when passing into getDatasets, because we still want
   // easy access to the history when turning the disabled paths back on.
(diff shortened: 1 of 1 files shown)


jtbandes:
@defunctzombie This should be a no-op change, but it's breaking storybook. Any ideas would be appreciated...

packages/studio-base/src/panels/Plot/index.tsx
```diff
@@ -419,12 +422,17 @@ function Plot(props: Props) {
...
-  const plotDataByPath = useMessageReducer<PlotDataByPath>({
-    topics: subscribeTopics,
-    preloadType: "full",
-    restore,
-    addMessages,
-  });
+  // The extra useShallowMemo is useful when seeking. Without it, the reduced value will be reset,
+  // and trigger the plot to re-render even if the old value and new value were both empty ({})
+  // which happens for fully pre-loaded data coming from blocks.
+  const plotDataByPath = useShallowMemo(
```
jtbandes, line 428: This is the perf improvement. The other changes above may actually be a bit of a perf regression, but they fix side effects (mutation) that were incorrectly happening in the reducer.

foxymiles (approved):

snosenzo (approved):

defunctzombie (approved):
Hard to tell from the structure but does the shallow memo _prevent_ updates in situations where it should not?

Further thinking on this makes me wonder why this isn't the responsibility of the _restore_ function to return the same reference if there is no change? The behavior of `useMessageReducer` is to pass along the return value of `restore` and `addMessage` and leave the logic of whether there is "equality" up to those.

Maybe this has already been considered in the implementation which is why I am approving the PR. I do find it unexpected that the extra layer had to be added. We don't do that at other reducer use sites.

jtbandes:
> why this isn't the responsibility of the restore function to return the same reference if there is no change?

The restore function doesn't get access to the previous data during a seek, to prevent panels from accidentally displaying data from the wrong timestamp.

Issue #4857, opened 2022-11-18
Poor buffering performance when loading a file with extremely high frequency topics


defunctzombie:
I've created this synthetic file to reproduce an issue with playback of recordings that contain very high frequency topics. The sample file here has a topic with ~10Khz messages.

When playing this data we see that the buffering ahead of playback does not do a great job of keeping data buffered. When pausing we see the playback buffering is able to buffer quicker.

https://user-images.githubusercontent.com/84792/202816892-9b4059c7-0be9-4014-8a10-b295bdacb5ae.mov

Looking into this at a high level seems to point to the way messages are loaded from the underlying file. Each _message_ requires an async iterator and post message (because the actual reader is in another worker) to fetch the message. This creates overhead for buffering each message.

[output.mcap.zip](https://github.com/foxglove/studio/files/10045373/output.mcap.zip)

PR #4891, opened 2022-11-27
Use batch requests to fetch messages from WorkerIterableSourceWorker

defunctzombie:
**User-Facing Changes**
Faster buffering of data sources with high topic rates.

**Description**
The data source logic runs in a web worker and exposed via the WorkerIterableSource class. When using the messageIterator method of the IterableSource this results in a worker RPC call to fetch each next message. With high rate sources (multi khz) this creates a lot of overhead to read each message and drastically slows down the buffering performance.

This change improves the performance substantially by introducing a new `nextBatch` method to the cursor interface. The `nextBatch` method allows fetching a batch of messages rather than a single next message. The `nextBatch` method receives a hint on how many messages to batch. Since the fastest Studio renders is 60 hz, this means we typically work in batches of 16 millisecond message ranges. This change updates the `messageIterator` call for the `WorkerIterableSource` to request batches of 16 milliseconds from the worker. This reduces the number of RPC calls from potentially several thousand to ~60 to fetch 1 second of data.

Fixes: #4857


defunctzombie:
@jtbandes @snosenzo Let me know what you think of this approach to solve the performance problem noted in the referenced issue. On my computer this leads to a substantial improvement with buffering and allows the buffering to keep up with playback on the sample bag that surfaced the performance problem.

packages/studio-base/src/players/IterablePlayer/WorkerIterableSource.ts
```diff
@@ -51,24 +51,19 @@ export class WorkerIterableSource implements IIterableSource {
...
-        const iterResult = await iter.next();
-        if (iterResult.done === true) {
-          return iterResult.value;
+        const results = await cursor.nextBatch(16);
+        if (!results || results.length === 0) {
+          break;
+        }
+        for (const result of results) {
+          yield result;
         }
```
jtbandes, line 63:
```suggestion
        yield* results;
```

jtbandes (approved):

PR #4903, opened 2022-11-30
Rename TimelineInteractionState

foxymiles:
**User-Facing Changes**
None

**Description**
This is a preliminary PR to rename the `TimelineInteractionState` store to `AppInteractionState` in anticipation of it taking on broader responsibilities like inter-panel communication for https://github.com/foxglove/studio/issues/4846.

### DIFF
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -12,7 +12,7 @@ import MultiProvider from "@foxglove/studio-base/components/MultiProvider";
 import StudioToastProvider from "@foxglove/studio-base/components/StudioToastProvider";
 import AppConfigurationContext from "@foxglove/studio-base/context/AppConfigurationContext";
 import { UserNodeStateProvider } from "@foxglove/studio-base/context/UserNodeStateContext";
-import TimelineInteractionStateProvider from "@foxglove/studio-base/providers/TimelineInteractionStateProvider";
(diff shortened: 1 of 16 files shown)


jtbandes (approved):
I'm not sure about the "Interaction" part of the name - makes sense for hovering, but doesn't really make sense for arbitrary key/value data sent by panels?

foxymiles:
> I'm not sure about the "Interaction" part of the name - makes sense for hovering, but doesn't really make sense for arbitrary key/value data sent by panels?

The immediate goal of syncing camera states also pertains to interactions but yeah it's more general than that. Any better suggestions?

jtbandes:
I guess my answer is that I don't really feel like these belong together under one logical unit. I get the desire to merge separate useState-based contexts into one zustand-based context — if several of all of the other contexts were merged (as they might be in the future), what would we ideally call that zustand context? AppState?

foxymiles:
> I guess my answer is that I don't really feel like these belong together under one logical unit. I get the desire to merge separate useState-based contexts into one zustand-based context — if several of all of the other contexts were merged (as they might be in the future), what would we ideally call that zustand context? AppState?

Personally I favor fewer contexts so I'd be happy with a more general AppState. Not sure if others would agree.

foxymiles:
Closing in favor of https://github.com/foxglove/studio/pull/4904

PR #4904, opened 2022-11-30
Rename PanelSettingsEditorStore to PanelStateStore

foxymiles:
**User-Facing Changes**
None

**Description**
Rename `PanelSettingsEditorStore` to `PanelStateStore` in preparation for adding more state to this store to implement inter-panel data sharing to achieve things like syncing 3d panel camera states.


jtbandes (approved):
LGTM. Note that we've got a TS type called `PanelsState` — that should probably be renamed to avoid confusion with this store...

PR #4898, opened 2022-11-29
Fix various memory leaks across changing tabs and opening new sources

snosenzo:
**User-Facing Changes**

**Description**

Fixes: #4824

PR #4906, opened 2022-11-30
Fix initial player memory leak for layouts with no tabs

snosenzo:
**User-Facing Changes**
 - Fix initial player memory leak for layouts with no tabs and no 3D panel

**Description**
 - tested with two raw message panels over multiple loading of the same source local file. 
 - useInitialDeeplinkState was keeping the initial player around in the context of the useEffect functions via `seekPlayback` since the context of those functions would not update when the player changes. To solve this, I moved each `useEffect` sync into their own functions so that only the `useEffect` that actually needs `seekPlayback` will be exposed to it. 

Part of https://github.com/foxglove/studio/pull/4898

### DIFF
--- a/packages/studio-base/src/hooks/useInitialDeepLinkState.ts
+++ b/packages/studio-base/src/hooks/useInitialDeepLinkState.ts
@@ -14,7 +14,7 @@ import { useCurrentUser } from "@foxglove/studio-base/context/CurrentUserContext
 import { EventsStore, useEvents } from "@foxglove/studio-base/context/EventsContext";
 import { usePlayerSelection } from "@foxglove/studio-base/context/PlayerSelectionContext";
 import { PlayerPresence } from "@foxglove/studio-base/players/types";
-import { parseAppURLState } from "@foxglove/studio-base/util/appURLState";
+import { AppURLState, parseAppURLState } from "@foxglove/studio-base/util/appURLState";
 
 const selectPlayerPresence = (ctx: MessagePipelineContext) => ctx.playerState.presence;
 const selectSeek = (ctx: MessagePipelineContext) => ctx.seekPlayback;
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/hooks/useInitialDeepLinkState.ts
```diff
@@ -88,9 +93,16 @@ export function useInitialDeepLinkState(deepLinks: readonly string[]): {
...
+    setUnappliedUrlState({ layoutId: undefined });
   }, [currentUserRequired, playerPresence, setSelectedLayoutId, unappliedUrlState?.layoutId]);
+}
 
+function useSyncTimeFromUrl(targetUrlState: AppURLState | undefined) {
+  const seekPlayback = useMessagePipeline(selectSeek);
+  const playerPresence = useMessagePipeline(selectPlayerPresence);
+  const [unappliedUrlState, setUnappliedUrlState] = useState(
+    targetUrlState ? { time: targetUrlState.time } : undefined,
+  );
```
jtbandes, line 105: Should we just split all these into separate states now, like `const [unappliedTime, setUnappliedTime] = useState(targetUrlState?.time)`?

packages/studio-base/src/hooks/useInitialDeepLinkState.ts
```diff
@@ -14,42 +14,31 @@ import { useCurrentUser } from "@foxglove/studio-base/context/CurrentUserContext
...
 
 const log = Log.getLogger(__filename);
 
-/**
- * Restores our session state from any deep link we were passed on startup.
+/*
+ * Sync functions necessary to keep separate to prevent memory leak from context kept in
+ * useEffect closures. Notably `seekPlayback` being attached to the context of the `useEffect` closures
+ * and that keeping the initial player in memory because the context is never updated because it doesn't depend
+ * on the player.
```
jtbandes, line 29: Trying to read and understand this comment I don't find it very clearly worded. I think using full sentences and not omitting words for brevity would help. We could also include a link back to the gh issue that spawned this work so it's easier to discover.

foxymiles (approved):

PR #4907, opened 2022-11-30
Fix DraggableToolbarTab causing react dnd to hold onto refs of detached elements causing old players to stick around

snosenzo:
**User-Facing Changes**
 - Fix memory leak from `DraggableToolbarTab` causing old players to stick around

**Description**
 - `DraggableToolbarTab` was connecting its outgoing `innerRef` to the larger `react-dnd` context (attached to window), causing the whole tab and connected player to remain in memory. The solutions was passing a callback function to connect the ref to the dnd context in order to connect only when the inner `Tab` renders. This will also make it properly clean up after unmounting.
 
Heap trace:
image (3)

formerly part of: https://github.com/foxglove/studio/pull/4898


packages/studio-base/src/panels/Tab/DraggableToolbarTab.tsx
```diff
@@ -63,16 +63,14 @@ export function DraggableToolbarTab(props: Props): JSX.Element {
...
-
   const tabProps = {
     tabTitle,
     tabIndex,
     isActive,
     tabCount,
     actions,
     isDragging,
-    innerRef: ref,
+    innerRef: () => connectDragRef(connectDropRef(ref)), // Combine drag and drop refs
```
jtbandes, line 73:
```suggestion
    innerRef: (el) => { connectDragRef(el); connectDropRef(el); }, // Combine drag and drop refs
```

packages/studio-base/src/panels/Tab/DraggableToolbarTab.tsx
```diff
@@ -32,15 +32,15 @@ export function DraggableToolbarTab(props: Props): JSX.Element {
   const { moveTab } = useCurrentLayoutActions();
 
   const ref = useRef<HTMLDivElement>(ReactNull);
```
jtbandes, line 34:
```suggestion
```

jtbandes (changes requested):
Dragging and dropping tabs doesn't work.

jtbandes (approved):

PR #4913, opened 2022-12-01
fix: Remove styling from global html,body

david-mancuso:
**User-Facing Changes**
As a user who embeds Foxglove studio into a larger suite of applications, we face the issue of global styling escaping the scope of the GlobalCss component in the studio-base package as it applies to global html and body tags.

**Description**
This change aims to reduce the impact of these global styles by consolidating them onto the #root element while maintaining the "full-bleed" design as originally intended.

### DIFF
--- a/packages/studio-base/src/components/GlobalCss.tsx
+++ b/packages/studio-base/src/components/GlobalCss.tsx
@@ -12,19 +12,26 @@ export default function GlobalCss(): JSX.Element {
     <GlobalStyles
       styles={{
         "html, body": {
-          boxSizing: "border-box",
           margin: 0,
           padding: 0,
-          height: "100%",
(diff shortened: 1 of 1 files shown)


defunctzombie:
Thanks for taking the time to contribute. This styling was preciously required on the _body_ due to a UI component framework we were using. We've recently stopped using that framework and use one that is smarter about how it handles styling so we are in a position to remove these from _body_.

I had previously opened a PR: https://github.com/foxglove/studio/pull/3802 trying a different approach to letting users opt-into some global styles. I think we can reconcile this PR with your PR and get something merged.

defunctzombie:
Closing this in favor of https://github.com/foxglove/studio/pull/3802

PR #4914, opened 2022-12-01
Add OnUnmount to panel extension API

snosenzo:
**User-Facing Changes**
- add `onUnmount` to panel extension api
- update `ThreeDeePanel.tsx` to use it

**Description**
Add settable `onUnmount` callback to PanelExtensionContext. This will allow the children of the panel element to call cleanup methods before the `panelElement` is removed. This will be called for each new `panelElement`. 
Update `ThreeDeePanel`'s `initPanel` to set the `onUnmount` to remove the React component from the panel element, causing the renderer to be disposed and cleanup functions within `ThreeDeePanel.tsx` to be called.


packages/studio-base/src/components/PanelExtensionAdapter/PanelExtensionAdapter.tsx
```diff
@@ -470,6 +471,9 @@ function PanelExtensionAdapter(props: PanelExtensionAdapterProps): JSX.Element {
 
     // Reset local state when the panel element is mounted or changes
     setRenderFn(undefined);
+    renderingRef.current = false;
+    setSlowRender(false);
```
snosenzo, line 475: Doing this solved the issue of the 3D panel initializing to a blank state because of the unMounting after initial mount.

defunctzombie:
Was there a drawback to using the "return function" from `initPanel` approach (similar to how useEffect/useLayoutEffect) work in react?

packages/studio-base/src/components/PanelExtensionAdapter/PanelExtensionAdapter.tsx
```diff
@@ -487,9 +491,17 @@ function PanelExtensionAdapter(props: PanelExtensionAdapterProps): JSX.Element {
       set onRender(renderFunction: RenderFn | undefined) {
         setRenderFn(() => renderFunction);
       },
+      // eslint-disable-next-line no-restricted-syntax
+      set onUnmount(onUnmountFn: () => void | undefined) {
```
defunctzombie, line 495: If I call this twice my previous function function is lost.

defunctzombie: With the return function approach we avoid this issue since there is only one thing you can return.

packages/studio-base/src/panels/ThreeDeeRender/index.tsx
```diff
@@ -13,13 +13,18 @@ import { SaveConfig } from "@foxglove/studio-base/types/panels";
...
+}
+
 function initPanel(context: PanelExtensionContext) {
   ReactDOM.render(
     <StrictMode>
       <ThreeDeeRender context={context} />
     </StrictMode>,
     context.panelElement,
   );
+  return unmountPanelFn(context.panelElement);
```
jtbandes, line 27: nit: would just put the closure inline here

jtbandes: Can we add this in all the other extension panels as well?

packages/studio-base/src/components/PanelExtensionAdapter/PanelExtensionAdapter.tsx
jtbandes: cc @defunctzombie for review

jtbandes:
"3D panel updated to use it" is not a user-facing change – these are meant to be things we can include in [release notes](https://github.com/foxglove/studio/releases).

packages/studio/index.d.ts
```diff
@@ -367,8 +360,13 @@ declare module "@foxglove/studio" {
     // your panel.
     name: string;
 
-    // This function is invoked when your panel is initialized
-    initPanel: (context: PanelExtensionContext) => void;
+    /**
+     * This function is invoked when your panel is initialized
+     * @return: (optional) Put necessary cleanup code here. Called before removing parent panelElement.
+     *   NOTE: useEffect cleanup functions will not activate in react extension panels by default.
```
defunctzombie, line 366: I would remove the `useEffect` note since that is specific to react extensions. As someone new to react I would also find this note somewhat confusing because it doesn't actually tell me how to activate those via this return value.

packages/studio-base/src/panels/ThreeDeeRender/index.tsx
snosenzo: > nit: would just put the closure inline here

I just wanted to be extra careful and put it outside so that it wouldn't capture the `context` in its closure in case the return function was kept around by react in the `useEffect` memoized context. I might be too paranoid 😅

> Can we add this in all the other extension panels as well?

We can, but I worry about time to test for unintended side-effects and it will take more time toward fixing the 3D panel memory leak if there are more caveats with it. I feel like it might be better to spend more time on after these memory leaks have been fixed. Happy to make another issue for it.

snosenzo: actually, after testing, the other extension panels certainly do have this problem and keep the old players around. will update

packages/studio-base/src/components/PanelExtensionAdapter/PanelExtensionAdapter.tsx
defunctzombie: 👍

packages/studio/index.d.ts
```diff
@@ -372,8 +372,11 @@ declare module "@foxglove/studio" {
     // your panel.
     name: string;
 
-    // This function is invoked when your panel is initialized
-    initPanel: (context: PanelExtensionContext) => void;
+    /**
+     * This function is invoked when your panel is initialized
+     * @return: (optional) Put necessary cleanup code here. Called before removing parent panelElement.
```
defunctzombie, line 377:
```suggestion
     * @return: (optional) A function which is called when the panel is removed. Typically intended for cleanup logic to gracefully teardown your panel.
```

defunctzombie (approved):

packages/studio-base/src/components/PanelExtensionAdapter/PanelExtensionAdapter.tsx
```diff
@@ -188,7 +189,7 @@ function PanelExtensionAdapter(props: PanelExtensionAdapterProps): JSX.Element {
   // updates.
   const renderingRef = useRef<boolean>(false);
   useLayoutEffect(() => {
-    if (!renderFn) {
+    if (!renderFn || !isPanelInitializedRef.current) {
```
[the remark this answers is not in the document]
snosenzo: Will add a comment, but apparently the `onRender` function is being called between the `initPanel`'s `useLayoutEffect` cleanup and `initPanel` being called again even if `setRenderFn(undefined)` is called in the cleanup function. This causes the old `renderFn` to be called in this effect and `pauseFrame` to happen, but it is never resumed, thus causing a 5 second delay in all panels in the layout to be loaded.

Issue #4918, opened 2022-12-01
Sidebar stuck saying "Loading panel settings..." after deleting a panel


jhurliman:
**Steps To Reproduce**
1. Create a layout with two panels
2. Open the sidebar settings and select a panel
3. Delete the selected panel using `~`
4. Observe the settings sidebar says "Loading panel settings..." but nothing is loading. It will stay in this state until closing the sidebar or manually selecting the remaining panel

**Expected Behavior**

Automatically select another panel and show its settings. If there are no panels, display a call to action or a blank sidebar. Don't say "loading" unless something is loading.

PR #4923, opened 2022-12-04
Show the correct settings state when the last selected panel is deleted

foxymiles:
**User-Facing Changes**
Fix an issue with the settings sidebar state after deleting the selected panel.

**Description**
We have code to deselect the panel when it's deleted:
https://github.com/foxglove/studio/blob/a95135e7f85c51809d212783e182dcc821a36856/packages/studio-base/src/providers/CurrentLayoutProvider/index.tsx#L305

But this code was firing *after* that update, reselecting a panel that is no longer in the layout:
https://github.com/foxglove/studio/blob/a95135e7f85c51809d212783e182dcc821a36856/packages/studio-base/src/components/Panel.tsx#L354-L369

The fix here is to stop event propagation from the click on the menu item so that the `onPanelRootClick` does not fire when a menu item is selected.

Fixes https://github.com/foxglove/studio/issues/4918

### DIFF
--- a/packages/studio-base/src/components/PanelToolbar/PanelActionsDropdown.tsx
+++ b/packages/studio-base/src/components/PanelToolbar/PanelActionsDropdown.tsx
@@ -246,7 +246,10 @@ export function PanelActionsDropdown({ isUnknownPanel }: Props): JSX.Element {
           ) : (
             <MenuItem
               key={item.key}
-              onClick={item.onClick}
+              onClick={(event) => {
+                event.stopPropagation();
+                item.onClick?.();
+              }}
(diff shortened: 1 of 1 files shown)


snosenzo:
i think this is still happening for the `~`-> remove panel  behavior described in the issue. 
Might have to address the remove panel `onClick` here as well? https://github.com/foxglove/studio/blob/main/packages/studio-base/src/components/Panel.tsx#L634

foxymiles:
> i think this is still happening for the `~`-> remove panel behavior described in the issue. Might have to address the remove panel `onClick` here as well? https://github.com/foxglove/studio/blob/main/packages/studio-base/src/components/Panel.tsx#L634

Good catch. Fixing.

Issue #4824, opened 2022-11-15
Memory leak when switching layouts containing 3D panel


asyncanup:
Seeing memory leaks when switching back and forth between layouts containing 3D panel, and when unmounting/remounting the app inside a host app.

The root cause for the memory leak is likely the following: the new 3D panel is implemented as an extension, and there does not appear to be any mechanism implemented in FG Studio to clean up after extension panels, or the way the 3D panel is wrapped as extension does not detect unmounting of the extension wrapper. Almost all objects that linger around as leak are allocated in the 3D panel.

We observed that the 3D panel gets initialized actually twice during a single bringup (in live mode), but never cleared. And I am still measuring about 10-20MB leak per cycle (of unmounting and remounting the app)

- Version: v1.31.0
- OS:
- Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native):
- ROS distro (if applicable):

**Steps To Reproduce**

**Expected Behavior**

**Actual Behavior**

sarbjit-longia:
I have worked on reproducing the issue with app mount/unmount and was able to reproduce the issue. I do understand that this might not be the normal use case for foxglove and might never happen with foxglove studio but its a big show stopper for us so would appreciate if we can spend time on this and fix it.

I have created a document that can be used to follow the steps in order to reproduce the issue. Document contains following:

- Steps to reproduce
- Drive session data data used
- Layout used for reproducing
- Video of the steps I used along with final Ah-Snap happening.
- Heap snapshots that I used to get some data.
- Code analysis for what I found out.

Please keep in mind that the code analysis is done to the best of my knowledge and is possible that I am looking in wrong direction so please use your own judgement and knowledge to fix this issue. 

**Steps:**

- Started with foxglove latest code from github.
- Added the code to add On-Off button([Patch](https://drive.google.com/file/d/1jomPy7No6bewfD_Uern-PSFhhA2pU2mv/view?usp=share_link)).
- Used data source from [link.](https://drive.google.com/file/d/1L9rmnlRcAGQIKwdcTK16YUMMm-NtX-Vt/view?usp=share_link)
- Used [layout](https://drive.google.com/file/d/1MEGE5ARAWT5yKbHbzxkMmaSuiRaS8vZT/view?usp=share_link).
- Used on-off button few times

[Google Doc](https://docs.google.com/document/d/1X9g6VuHUyoquzfQ3NX9pqgYYCmzbPqvuXo-xeL6arWI/edit?usp=sharing)

[...]

PR #4924, opened 2022-12-05
Fix `selectRecent` memory leak keeping old players in memory

snosenzo:
**User-Facing Changes**
 - fix frequent memory leak where `selectRecent` would keep the players in memory after selecting a new source

**Description**

Because the callback is exposed to the larger PlayerManager closure context it captures the player in it. This function is used in memoized state in `Start.tsx`, but the `OpenDialog` that contains it isn't cleaned up by react on unmount for an unknown reason. `selectSource` does not seem to have the same problem with its usage. As a workaround, moving the definition of the `selectRecent` callback to an outside function will prevent it from keeping the player in its closure context this way.

The `stateNode` holding onto the detached element is the mui-modal in `OpenDialog` which is unmounted, but it's still stored in react internal state functions (ie: `stateNode` and `lastEffect`), which makes me think this might be a react issue. When we update react we should probably check to see whether this is still the case because apparently cleanup in react 18 is more aggressive. 

Retainer stack:
image

To reproduce prior to fix:
(with memory panel open in chrome devtools)
1. use studio default layout
2. Refresh the page and select a local file source
3. Wait for buffering to finish
4. Go to data source tab and select the same local source
5. Take a snapshot in the memory panel
6. See two iterable players in memory, the old IterablePlayer will be the one that has a greater distance, click to inspect its retainers
7. It's primary retainers should resemble the screenshot above

### DIFF
--- a/packages/studio-base/src/components/PlayerManager.tsx
+++ b/packages/studio-base/src/components/PlayerManager.tsx
@@ -41,7 +41,7 @@ import PlayerSelectionContext, {
 } from "@foxglove/studio-base/context/PlayerSelectionContext";
 import { useUserNodeState } from "@foxglove/studio-base/context/UserNodeStateContext";
 import { GlobalVariables } from "@foxglove/studio-base/hooks/useGlobalVariables";
-import useIndexedDbRecents from "@foxglove/studio-base/hooks/useIndexedDbRecents";
+import useIndexedDbRecents, { RecentRecord } from "@foxglove/studio-base/hooks/useIndexedDbRecents";
 import useWarnImmediateReRender from "@foxglove/studio-base/hooks/useWarnImmediateReRender";
 import AnalyticsMetricsCollector from "@foxglove/studio-base/players/AnalyticsMetricsCollector";
 import UserNodePlayer from "@foxglove/studio-base/players/UserNodePlayer";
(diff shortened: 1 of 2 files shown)


snosenzo:
part of fixing: https://github.com/foxglove/studio/issues/4824

defunctzombie:
If we move everything in this component to a zustand store do we think that would solve this situation more broadly rather than picking off one-by-one?

packages/studio-base/src/components/PlayerManager.tsx
```diff
@@ -334,3 +313,36 @@ export default function PlayerManager(props: PropsWithChildren<PlayerManagerProp
     </>
   );
 }
+
+// fixes memory leak across > 2 data sources without reloading
```
defunctzombie, line 317: This comment does not tell how the memory leak is fixed or what I should avoid doing to re-introduce the memory leak. I'd recommend either improving the comment or moving the structure of state management towards an approach that prevents this kind of _leak_ generally.

foxymiles: Is there another word we can use besides `context` to describe closed over variables? Just skimming this I'm going to think it means a react context.

jtbandes: closure context....clontext?

snosenzo:
> If we move everything in this component to a zustand store do we think that would solve this situation more broadly rather than picking off one-by-one?

That's a good question. Potentially? It's hard to say without really testing. It's probably be worth attempting in the future, but I worry that we still might run into the same problem that the `createPlayerListener` function has in the message pipeline.

packages/studio-base/src/components/PlayerManager.tsx
snosenzo: I added more info to the PR description and updated the comment to be a little more clear. let me know if it's still confusing

PR #4928, opened 2022-12-06
Remove "Open file by URL" and add DP callout to Welcome dialog; with local MCAP file loading fix

esthersweon:
**User-Facing Changes**
- Remove "Open file by URL" and add DP callout to Welcome dialog
- Fixed bug in previous PR (#4785) with loading local MCAP files 
- 
image


jtbandes (approved):
I think this should resolve the issue. Could you please confirm that we've manually tested at least these four different source types:
- local .bag
- local .mcap
- remote .bag
- remote .mcap

esthersweon:
@jtbandes - have tested all 4 successfully (used sample dataset for the remote bag and https://assets.foxglove.dev/NuScenes-v1.0-mini-scene-0061-8c50124.mcap for the remote MCAP).

jtbandes:
FYI I think the sample dataset is now a .mcap file, but we do have .bags in the assets bucket as well: https://assets.foxglove.dev/demo.bag

esthersweon:
@jtbandes - Confirmed that remote bag file URL also works.

PR #4929, opened 2022-12-06
Fix source selection in useOpenFile

defunctzombie:
**User-Facing Changes**
None

**Description**
useOpenFile would select any source which supported the extension of the file we were trying to load. This resulted in it select _remote_ sources if they appeared before _file_ in the sources list.

This change fixes this behavior to only consider _file_ sources.


jtbandes (approved):
Would it make sense to assert that filter(...).length <= 1?

defunctzombie:
> Would it make sense to assert that filter(...).length <= 1?

Done

Issue #4934, opened 2022-12-06
Rogue Roboto font?


jtbandes:
<img width="318" alt="image" src="https://user-images.githubusercontent.com/14237/205818068-ba1f5ace-dce6-4204-89bd-f19c636b868d.png">

It appears we are now using the Roboto font for some text, e.g. in the Autocomplete dropdown. I believe this is a regression, since we normally only use Inter as our proportional font and Roboto Mono as our monospace font.

May be related to:
- https://github.com/foxglove/studio/pull/3802

jtbandes:
This was not the case in 1.33.0:

<img width="404" alt="image" src="https://user-images.githubusercontent.com/14237/205818544-145eaa2c-7c86-474c-8dfd-c09fbcb6d9b9.png">

jtbandes:
Another one from the Open dialog:

<img width="292" alt="image" src="https://user-images.githubusercontent.com/14237/206009378-dc7e82d7-c713-471f-9daa-32a17715c334.png">

PR #4940, opened 2022-12-07
Move GlobalCss back into <App> under a prop

defunctzombie:
**User-Facing Changes**
Styling fixes for some modal text elements.

**Description**
The MuiModal root appears outside of the #root component directly within body. This means it does not inherit the styling we apply on #root (such as font family and sizes) for some components (Link seems to be one of them).

As a workaround we move <GlobalCss> back into <App> under a property to allow users of <App> to specify if they want to turn these global values on. Without these the styling might be wrong on certain elements if the user does not set the same font on the body that they use within the app.

Additionally, GlobalCss must exist within the ThemeProvider and was not before so its setting of the background color and body elements was incorrect when used outside of <App>

Fixes: #4934

### DIFF
--- a/desktop/renderer/Root.tsx
+++ b/desktop/renderer/Root.tsx
@@ -106,7 +106,7 @@ export default function Root({
         extensionLoaders={extensionLoaders}
         nativeAppMenu={nativeAppMenu}
         nativeWindow={nativeWindow}
-        enableGlobalCss={true}
+        enableGlobalCss
       />
     </>
   );
--- a/desktop/renderer/Root.tsx
+++ b/desktop/renderer/Root.tsx
@@ -10,7 +10,6 @@ import {
   ConsoleApi,
   FoxgloveDataPlatformDataSourceFactory,
   FoxgloveWebSocketDataSourceFactory,
-  GlobalCss,
   IAppConfiguration,
   IDataSourceFactory,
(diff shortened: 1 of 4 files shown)


defunctzombie:
An additional safety net we can apply is to add a style override for `MuiModal` root that sets the typography css properties to those that mirror what we set on the root element elsewhere.

desktop/renderer/Root.tsx
```diff
@@ -108,6 +106,7 @@ export default function Root({
         extensionLoaders={extensionLoaders}
         nativeAppMenu={nativeAppMenu}
         nativeWindow={nativeWindow}
+        enableGlobalCss={true}
```
jtbandes, line 109:
```suggestion
        enableGlobalCss
```

web/src/Root.tsx
```diff
@@ -74,6 +72,7 @@ export function Root({ appConfiguration }: { appConfiguration: IAppConfiguration
         layoutStorage={layoutStorage}
         consoleApi={consoleApi}
         extensionLoaders={extensionLoaders}
+        enableGlobalCss={true}
```
jtbandes, line 75:
```suggestion
        enableGlobalCss
```

jtbandes (approved):

PR #4942, opened 2022-12-07
Fix Windows deep link handling by encoding with base64

### LINKED ISSUE #4896: Opening Timeline Recording to Desktop on Windows Fails
**Description**
Opening a recording from timeline to desktop on Windows results in a blank studio client.

- Version: 1.33.0
- OS: Windows

**Steps To Reproduce**
1. Set Studio to open links in desktop 
image
2. Select recording time-range in Console
3. Attempt to open time-range

**Expected Behavior**
Recording range opened in studio

**Actual Behavior**
Blank window

cc: @amacneil @rrryanc

jtbandes:
**User-Facing Changes**
Fixed an issue on Windows where `foxglove://` links would open a blank Foxglove Studio window.

**Description**
Fixes #4896 

It seems like passing arbitrary arguments to the renderer is not safe on windows for some reason. We work around the problem by base64-encoding.

See: https://github.com/electron/electron/issues/31168, https://github.com/electron/electron/issues/32064

### DIFF
--- a/desktop/preload/index.ts
+++ b/desktop/preload/index.ts
@@ -13,6 +13,7 @@ import Logger from "@foxglove/log";
 import { NetworkInterface, OsContext } from "@foxglove/studio-base/src/OsContext";
 
 import pkgInfo from "../../package.json";
+import { decodeRendererArg } from "../common/rendererArgs";
 import { Desktop, ForwardedMenuEvent, NativeMenuBridge, Storage } from "../common/types";
 import LocalFileStorage from "./LocalFileStorage";
 import { getExtensions, loadExtension, installExtension, uninstallExtension } from "./extensions";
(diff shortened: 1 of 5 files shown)


desktop/preload/index.ts
```diff
@@ -104,7 +104,11 @@ const desktopBridge: Desktop = {
     await ipcRenderer.invoke("updateNativeColorScheme");
   },
   getDeepLinks(): string[] {
-    return window.process.argv.filter((arg) => arg.startsWith("foxglove://"));
+    const deepLinksArg = window.process.argv
```
defunctzombie, line 107: Please extract this logic to a function. I'm thinking one you can pass an argument array to that finds the argument you want and then returns the deserialized deep links. Add a unit test for it. 🙏

defunctzombie:
Good stuff

PR #4944, opened 2022-12-07
[3D] Customize "selected_id" variable for SceneEntity and Marker topics

jhurliman:
**User-Facing Changes**
- [3D] Set another global variable in addition to "selected_id" when selecting a SceneEntity or Marker

**Description**

SceneEntity and Marker topic settings now have an additional field to customize the global variable that is updated when 
selecting an entity or marker. This global variable is sticky, and is set in addition to `selected_id` to preserve the 
existing behavior.

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/Renderable.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/Renderable.ts
@@ -10,6 +10,8 @@ import type { Renderer } from "./Renderer";
 import type { BaseSettings } from "./settings";
 import type { Pose } from "./transforms";
 
+export const SELECTED_ID_VARIABLE = "selected_id";
+
 export type BaseUserData = {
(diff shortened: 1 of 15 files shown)


snosenzo (approved):
looks good to me, but I'm wondering if it makes sense to keep these selection variables across different data sources? or are global variables already cleared when a new source comes in and then reinstantiated by the panels in layout?

jhurliman:
> looks good to me, but I'm wondering if it makes sense to keep these selection variables across different data sources? or are global variables already cleared when a new source comes in and then reinstantiated by the panels in layout?

I believe global variables are a layout concept, not a player concept. Layouts contain topic names, user scripts, global variables, and settings that may or may not make sense across different data sources.

Clearing all global variables when opening a data source is worth discussing, but I think orthogonal to this PR.

PR #4946, opened 2022-12-08
Significantly reduce memory impact of leaked players across sources

snosenzo:
**User-Facing Changes**
 - Significantly reduce memory impact of leaked players across sources

**Description**
In the event that part of application is holding onto old IterablePlayers, now on `close()` we will clear the buffer caches to eliminate the majority of the memory leak. This will significantly reduce the impact of this kind of memory leak in the future to a near negligible size.
Retained sizes before -> after: 308428529 bytes -> 5024 bytes 
image

### DIFF
--- a/packages/studio-base/src/players/IterablePlayer/BufferedIterableSource.ts
+++ b/packages/studio-base/src/players/IterablePlayer/BufferedIterableSource.ts
@@ -158,6 +158,11 @@ class BufferedIterableSource implements IIterableSource {
     log.debug("producer done");
   }
 
+  public async terminate(): Promise<void> {
+    this.cache.clear();
+    await this.source.terminate();
+  }
+
   public async stopProducer(): Promise<void> {
     this.aborted = true;
     this.writeSignal.notifyAll();
--- a/packages/studio-base/src/players/IterablePlayer/CachingIterableSource.ts
+++ b/packages/studio-base/src/players/IterablePlayer/CachingIterableSource.ts
@@ -92,6 +92,11 @@ class CachingIterableSource implements IIterableSource {
     return this.initResult;
   }
 
+  public async terminate(): Promise<void> {
+    this.cache.length = 0;
+    this.cachedTopics.length = 0;
+  }
+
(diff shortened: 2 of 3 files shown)


packages/studio-base/src/players/IterablePlayer/BufferedIterableSource.ts
```diff
@@ -158,6 +158,11 @@ class BufferedIterableSource implements IIterableSource {
     log.debug("producer done");
   }
 
+  public async terminate(): Promise<void> {
```
jhurliman, line 161: We use `dispose()` throughout the 3D panel; I'm not sure if `terminate()` is already used elsewhere in Studio though.

packages/studio-base/src/players/IterablePlayer/CachingIterableSource.ts
```diff
@@ -92,6 +92,11 @@ class CachingIterableSource implements IIterableSource {
     return this.initResult;
   }
 
+  public async terminate(): Promise<void> {
+    this.cache.length = 0;
+    this.cachedTopics.length = 0;
+  }
```
jhurliman, line 98: Is there a use case for making this async? I haven't seen any resource disposal in JavaScript that would require a round-trip through the event loop.

packages/studio-base/src/players/IterablePlayer/BufferedIterableSource.ts
snosenzo: this is overriding the definition in `IIterableSource` which expects it to be async, wanted to keep to that signature. 
```
  /**
   * Optional method a data source can implement to cleanup resources. The player will call this
   * method when the source will no longer be used.
   */
  terminate?: () => Promise<void>;
  ```

packages/studio-base/src/players/IterablePlayer/CachingIterableSource.ts
snosenzo: see above, though open to hear thoughts for setting it apart.

Issue #4909, opened 2022-12-01
Studio uses the incorrect schema name for protobuf Schema MCAP records


jhurliman:
**Description**

Studio data source sidebar shows the wrong schema name in the case of a disagreement between a protobuf descriptor and it's encompassing Schema record's `schema_name`.

**Steps To Reproduce**

Create an MCAP file with a protobuf Schema record using:

```
syntax = "proto3";

message foo.Bar {
  optional x string = 1;
}
```

And set the MCAP Schema record `schema_name` to `"Bar"`. Open this file in Studio and observe `foo.Bar` is shown as the schema name instead of `Bar`.

jtbandes:
Relevant code:

https://github.com/foxglove/studio/blob/fc20bc0d271b48e7ddf6e91e1f8e1ade8f48440e/packages/mcap-support/src/parseChannel.ts#L142-L143

Perhaps the `fullSchemaName` concept can go away?

Note that there may be existing WebSocket (or MCAP) users working with Protobuf who are depending on this behavior, although we never really documented it.

Issue #4732, opened 2022-11-01
FrameTransform not detected properly when using live websocket connection


pickledgator:
**Description**
- Version: 1.29 and 1.30
- OS: MacOS
- Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native): Live websocket stream using protobuf API
- ROS distro (if applicable): 

**Steps To Reproduce**
Send a `foxglove.FrameTransform` protobuf message via websocket API

**Expected Behavior**
Client detects the transform properly, as it did in version 1.28
![image](https://user-images.githubusercontent.com/966598/199322829-d9ff4f83-d1d6-4e76-9699-411732e515dd.png)

**Actual Behavior**
3D panel (and potentially others) does not detect the transform
Some panels (raw data) detect transform correctly

![image](https://user-images.githubusercontent.com/966598/199322994-67d69310-0488-4c7d-9586-0f7a57630a95.png)
![image](https://user-images.githubusercontent.com/966598/199322879-bac8eaad-76ff-48fb-a7cb-8137730c40f2.png)

jtbandes:
I tested FrameTransform over WS with a simple test server: https://gist.github.com/jtbandes/9e738c11bb8ec9ba36536837e7990f17

It seems to work okay, so waiting for more user feedback.

https://user-images.githubusercontent.com/14237/200989163-cc62f716-a45c-4635-81ab-8b1dddb714c0.mov

pickledgator:
I’m traveling for the next week or so and then will jump back on this.
Sorry for the delay in feedback.

defunctzombie:
@pickledgator No worries. We'll keep this alive waiting for your response.

maxgaw:
@defunctzombie @jtbandes  I can jump in since I have the same issue. It seems like the 3D panel is not able to use the FrameTransform when using a Websocket connection.
In the video you can see the live connection first. At 00:21 you can see the exact same data loaded from the data-portal. 
Besides the message missing frame to frame, nothing else gets reported. In this example the frame hierarchy should be root > lane_frame > host_frame - mov_obj_frame as seen in the example at 00:21
When using build 1.28 everything works as expected.
https://user-images.githubusercontent.com/73784937/202665785-b73260a6-2543-42c6-86b0-d43f6aa93cfc.mp4

acowley:
[...]

PR #4948, opened 2022-12-08
remove fullSchemaName, adhere to name from data source

jtbandes:
**User-Facing Changes**
- Changed: Studio no longer modifies or normalizes schema names from the underlying data source. When using foxglove schemas, schema names must be fully-qualified (`foxglove.X` rather than just `X`)

**Description**
Removes `fullSchemaName` (introduced in https://github.com/foxglove/studio/pull/2167) in favor of just using the schema name from the data source.
Fixes https://github.com/foxglove/studio/issues/4909

### DIFF
--- a/packages/mcap-support/src/parseChannel.ts
+++ b/packages/mcap-support/src/parseChannel.ts
@@ -11,7 +11,7 @@ import { MessageReader as ROS2MessageReader } from "@foxglove/rosmsg2-serializat
 
 import { parseFlatbufferSchema } from "./parseFlatbufferSchema";
 import { parseJsonSchema } from "./parseJsonSchema";
-import { protobufDefinitionsToDatatypes, stripLeadingDot } from "./protobufDefinitionsToDatatypes";
+import { protobufDefinitionsToDatatypes } from "./protobufDefinitionsToDatatypes";
 import { RosDatatypes } from "./types";
 
 type Channel = {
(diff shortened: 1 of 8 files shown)


packages/mcap-support/src/parseChannel.ts
```diff
@@ -138,12 +137,7 @@ export function parseChannel(channel: Channel): ParsedChannel {
     const datatypes: RosDatatypes = new Map();
     protobufDefinitionsToDatatypes(datatypes, type);
```
snosenzo, line 138: this method is still using stripLeading dot and setting to the `datatypes` not sure if that's desirable

jtbandes: Not sure either. That is probably related to why it was originally done this way though. Might need some deeper investigation...

jtbandes:
Need to investigate this before merging: https://github.com/foxglove/studio/issues/4732#issuecomment-1343629020

jhurliman (approved):

jtbandes:
Update on #4732: Confirmed that the behavior unintentionally changed in this commit since it switched from using the original schema name to the inferred full name: https://github.com/foxglove/studio/pull/4656/files#diff-a9c3d412d9905d1721d44be8b2625582b9167e30f824540bcf61e17464cc1743L882

PR #4905, opened 2022-11-30
Restyled Open Dialog

2metres:
**User-Facing Changes**
| Dark | Light |
|---|---|
|Screen Shot 2022-12-02 at 12 32 15 pm|Screen Shot 2022-12-02 at 12 32 04 pm|

**Description**
Restyle Open Dialog

Resolves #4862

PR #4951, opened 2022-12-08
Adjust Data Platform CTA copy for unauthenticated users

esthersweon:
**User-Facing Changes**
None (part of earlier commit's changes - #4905)

**Description**
- Mention both products in right panel's headers for stronger brand presence
- Revise header text to not just focus on collaboration, as we don't want to discourage single users from creating an account
- Cover all categories of benefits laid out on [the Console signup page](https://console.foxglove.dev/signin) 
- Avoid having so many pairs of words for Data Platform CTAs ("Upload and share", "Store and collaborate") – weakens messaging around DP's value props

Before:
Screen Shot 2022-12-08 at 1 42 30 PM

After:
Screen Shot 2022-12-08 at 3 00 16 PM

### DIFF
--- a/packages/studio-base/src/components/OpenDialog/Start.tsx
+++ b/packages/studio-base/src/components/OpenDialog/Start.tsx
@@ -304,22 +304,24 @@ function SidebarItems(props: { onSelectView: (newValue: OpenDialogViews) => void
 
   const sidebarItems: SidebarItem[] = useMemo(() => {
     switch (currentUserType) {
-      case "unauthenticated":
+      case "unauthenticated": {
+        const ulStyle = { padding: "0px" };
+        const liStyle = { margin: "0px 0px 5px 10px" };
         return [
           ...freeUser,
           {
             id: "collaborate",
-            title: "Accelerate   development with Foxglove Data Platform",
+            title: "Accelerate development with Foxglove Data Platform",
             text: (
-              <ul style={{ padding: "0px" }}>
-                <li>Securely store petabytes of ROS or custom data</li>
-                <li>
-                  Use a convenient web interface to search and retrieve data at lightning speed
+              <ul style={ulStyle}>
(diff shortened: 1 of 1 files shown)


jtbandes:
- bullet points look weird protruding out the left side
- maybe add a little spacing between list items?
- 4 bullets is a lot of text, maybe cut to 3?
- do we want one of the CTAs to be something like "learn more"? Having the two options be create an account or sign in feels a bit heavy weight – some users might not want to invest in creating an account before they learn more about DP.

esthersweon:
@jtbandes - 

> * bullet points look weird protruding out the left side
> * maybe add a little spacing between list items?
> * 4 bullets is a lot of text, maybe cut to 3?

Implemented all of the above.

> * do we want one of the CTAs to be something like "learn more"? Having the two options be create an account or sign in feels a bit heavy weight – some users might not want to invest in creating an account before they learn more about DP.

Both current options take you to a page that teaches you more about the benefits of an account, so I don't think so. @2metres wanted to prioritize CTAs that welcome in both users without accounts and users with accounts who haven't yet signed in.

packages/studio-base/src/components/OpenDialog/Start.tsx
```diff
@@ -304,22 +304,24 @@ function SidebarItems(props: { onSelectView: (newValue: OpenDialogViews) => void
 
   const sidebarItems: SidebarItem[] = useMemo(() => {
     switch (currentUserType) {
-      case "unauthenticated":
+      case "unauthenticated": {
+        const ulStyle = { padding: "0px" };
+        const liStyle = { margin: "0px 0px 5px 10px" };
```
jtbandes, line 309: nit: maybe put these inside the `makeStyles` block at the top and use them that way?

jtbandes (approved):

PR #4952, opened 2022-12-08
experimental toolbar API for extension panels

jtbandes:
**User-Facing Changes**

**Description**


defunctzombie:
We've decided to go a different direction and not add this to the API for now. Feel free to keep open if you want to keep experimenting.

jtbandes:
I considered closing but I noted that we still have some panels (like Log) with custom toolbars that we don't have a good solution for yet.

defunctzombie:
Probably the experiment ended and we can close this? Do you want to save this work in some form? Maybe a branch or notes in some document?

PR #4953, opened 2022-12-09
[3D] Fix projected image dimensions, configurable planar projection factor

### LINKED ISSUE #4852: Image and View Frustum seem overly stretched in 3D view
**Description**
When using the CameraCalbiration Proto and a RawImage proto to draw an image in a 3D view, the images look stretched. Square images seem to stay square, but non-square get warped in some odd way.

**Steps To Reproduce**

Run the following code
```python
# Install deps:
#   pip3 install foxglove-websocket foxglove-schemas-protobuf
# Run:
#   python3 test_img.py

import math
import asyncio
import time
from base64 import b64encode
from typing import Set, Type
from foxglove_websocket import run_cancellable
from foxglove_websocket.server import FoxgloveServer, FoxgloveServerListener
from foxglove_websocket.types import ChannelId
from foxglove_schemas_protobuf.RawImage_pb2 import RawImage
from foxglove_schemas_protobuf.CameraCalibration_pb2 import CameraCalibration
import google.protobuf.message
from google.protobuf.descriptor_pb2 import FileDescriptorSet
from google.protobuf.descriptor import FileDescriptor


def build_file_descriptor_set(
    message_class: Type[google.protobuf.message.Message],
) -> FileDescriptorSet:
    """
    Build a FileDescriptorSet representing the message class and its dependencies.
    """
    file_descriptor_set = FileDescriptorSet()
    seen_dependencies: Set[str] = set()

    def append_file_descriptor(file_descriptor: FileDescriptor):
        for dep in file_descriptor.dependencies:
            if dep.name not in seen_dependencies:
                seen_dependencies.add(dep.name)
                append_file_descriptor(dep)
        file_descriptor.CopyToProto(file_descriptor_set.file.add())  # type: ignore

    append_file_descriptor(message_class.DESCRIPTOR.file)
    return file_descriptor_set


async def main():
    class Listener(FoxgloveServerListener):
        def on_subscribe(self, server: FoxgloveServer, channel_id: ChannelId):
            print("First client subscribed to", channel_id)

        def on_unsubscribe(self, server: FoxgloveServer, channel_id: ChannelId):
            print("Last client unsubscribed from", channel_id)

    async with FoxgloveServer("0.0.0.0", 8765, "example server") as server:
        server.set_listener(Listener())
        img_chan_id = await server.add_channel(
            {
                "topic": "example_msg",
                "encoding": "protobuf",
                "schemaName": RawImage.DESCRIPTOR.full_name,
                "schema": b64encode(
                    build_file_descriptor_set(RawImage).SerializeToString()
                ).decode("ascii"),
            }
        )
        calib_chan_id = await server.add_channel(
            {
                "topic": "example_calib",
                "encoding": "protobuf",
                "schemaName": CameraCalibration.DESCRIPTOR.full_name,
                "schema": b64encode(
                    build_file_descriptor_set(CameraCalibration).SerializeToString()
                ).decode("ascii"),
            }
        )

        i = 0
        while True:
            i += 1
            await asyncio.sleep(0.1)
            now = time.time_ns()

            img = RawImage()
            img.timestamp.FromNanoseconds(now)
            img.frame_id = "img"
            img.width = 100
            img.height = 100
            img.encoding = "bgra8"
            data = bytearray()
            for y in range(img.height):
                for x in range(img.width):
                    data.extend([int(y/img.height * 255), int(x/img.width * 255), 127, 255])
            img.data = bytes(data)
            await server.send_message(img_chan_id, now, img.SerializeToString())

            calib = CameraCalibration()
            calib.timestamp.FromNanoseconds(now)
            calib.frame_id = "img"
            calib.width = img.width
            calib.height = img.height
            calib.distortion_model = "plumb_bob"

            fx = 30
            fy = 30
            cx = img.width / 2
            cy = img.height / 2
            tx = 0
            ty = 0
            calib.K[:] = [fx, 0, cx, 0, fy, cy, 0, 0, 1]
            calib.P[:] = [fx, 0, cx, tx, 0, fy, cy, ty, 0, 0, 1, 0]

            await server.send_message(calib_chan_id, now, calib.SerializeToString())


if __name__ == "__main__":
    run_cancellable(main())
```
And connect to the websocket 

**Actual Behavior**
A width height ratio of 1:1 results in square. This is expected.
Screenshot 2022-11-17 at 4 56 37 PM

With a width height ratio of 2:1 the side lengths look like 4:1
Screenshot 2022-11-17 at 4 54 50 PM

Notice that with a ratio of sqrt(2):1 the side lengths appear 2:1
Screenshot 2022-11-17 at 4 54 24 PM

The squashing is evident in real data as well. In the Foxglove samples, I saw the same behavior. Look how the W is stretched as compared with the actual image.
Screenshot 2022-11-16 at 5 14 53 PM (1)

jhurliman:
**User-Facing Changes**
- [3D] Fixed aspect ratio for camera frustums and images
- [3D] Configurable "Planar Projection Factor" for camera frustums and images

**Description**
This PR makes a few changes to the math in PinholeCameraModel to make the results match the ROS PinholeCameraModel as closely as possible. Tests have been added with ground truth from ROS Noetic. Rectification correctly normalizes by focal length instead of image dimensions. Five Jacobi iterations are used instead of three, with a bypass if the distortion model is empty, to match OpenCV behavior. The output of projectPixelTo3dRay is normalized to match the ROS version, and a separate projectPixelTo3dPlane is added to preserve the previous behavior. The planar projection factor is a new setting for cameras and images to control how curved or flat the projection into 3D space is.

Fixes #4852

Screenshot 2022-12-09 at 8 56 48 AM

https://user-images.githubusercontent.com/195374/206807548-be24e914-947e-4965-8e89-a7dea841df58.mp4

### DIFF
--- a/packages/den/image/PinholeCameraModel.ts
+++ b/packages/den/image/PinholeCameraModel.ts
@@ -1,15 +1,6 @@
 // This Source Code Form is subject to the terms of the Mozilla Public
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
-//
-// This file incorporates work covered by the following copyright and
-// permission notice:
-//
-//   Copyright 2018-2021 Cruise LLC
-//
-//   This source code is licensed under the Apache License, Version 2.0,
-//   found at http://www.apache.org/licenses/LICENSE-2.0
-//   You may not use this file except in compliance with the License.
 
 import type { CameraInfo } from "./CameraInfo";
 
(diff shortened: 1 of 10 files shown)


packages/den/image/PinholeCameraModel.ts
```diff
@@ -114,10 +154,25 @@ export class PinholeCameraModel {
     out.x = (pixel.x - cx - tx) / fx;
     out.y = (pixel.y - cy - ty) / fy;
     out.z = 1.0;
+
+    // Normalize the ray direction
+    const invNorm = 1.0 / Math.sqrt(out.x * out.x + out.y * out.y + out.z * out.z);
```
snosenzo, line 159: Why is it important to normalize this now?

packages/den/image/PinholeCameraModel.ts
```diff
@@ -156,20 +217,28 @@ export class PinholeCameraModel {
 
     const x0 = x;
     const y0 = y;
-    for (let i = 0; i < iterations; i++) {
-      const r2 = x ** 2 + y ** 2; // squared distance in the image projected by the pinhole model
+    const count = k1 !== 0 || k2 !== 0 || p1 !== 0 || p2 !== 0 || k3 !== 0 ? iterations : 1;
+    for (let i = 0; i < count; i++) {
+      const r2 = x * x + y * y; // squared distance in the image projected by the pinhole model
```
snosenzo, line 222: is there a difference between `x ** 2` and `x * x`? or was this changed just to use a more predictable operator?

packages/den/image/PinholeCameraModel.ts
```diff
@@ -156,20 +217,28 @@ export class PinholeCameraModel {
 
     const x0 = x;
     const y0 = y;
-    for (let i = 0; i < iterations; i++) {
-      const r2 = x ** 2 + y ** 2; // squared distance in the image projected by the pinhole model
+    const count = k1 !== 0 || k2 !== 0 || p1 !== 0 || p2 !== 0 || k3 !== 0 ? iterations : 1;
```
snosenzo, line 220: why are these checks of nonzero variables necessary?

snosenzo: I think putting the comment about matching cv behavior here would be helpful

jhurliman: The line reads: "If distortion parameters are non-zero, use `iterations` iterations to approximate a solution. If all distortion parameters are zero, this reduces to a closed-form equation so we can use a single iteration".

packages/den/image/PinholeCameraModel.ts
jhurliman: I hope an optimizing JIT like v8 will turn `x ** 2` into `x * x` instead of `Math.pow(x, 2)`, this is just guaranteeing it does the more efficient thing.

packages/den/image/PinholeCameraModel.ts
jtbandes, line 159: Did this result in any visible bug? Or is it part of the same bug?

packages/den/image/PinholeCameraModel.ts
snosenzo: ah gotchya, should you replace the remaining `**` operators on the next line then?

packages/den/image/PinholeCameraModel.ts
jtbandes: Ah, found it in the PR description

> A side effect of normalizing the projected ray: images in 3D are no longer planar (they have a planar projection factor of 0).

jtbandes (approved):
Can we add a story screenshot using the sample data from https://github.com/foxglove/studio/issues/4852 with a head-on view that makes it more obvious that the image has the correct aspect ratio? The change is visible in the current screenshot tests, but it's not particularly evident that one is right and one is wrong.

image

PR #4957, opened 2022-12-10
Add cbor mcap support

therishidesai:
**User-Facing Changes**

**Description**
Foxglove studio can now parse CBOR messages and display them in the UI. The one requirement is that a json schema is also given to make for easy search of topics/message types in the UI.

https://github.com/foxglove/community/issues/330

To test this you can run the UI from this branch and use the mcap file generated by this repo: https://github.com/therishidesai/mcap_experiments

### DIFF
--- a/package.json
+++ b/package.json
@@ -122,5 +122,8 @@
     "webpack-dev-server": "4.11.0",
     "webpack-hot-middleware": "2.25.2"
   },
-  "packageManager": "yarn@3.1.0"
+  "packageManager": "yarn@3.1.0",
+  "dependencies": {
+    "cbor-js": "0.1.0"
+  }
 }
--- a/packages/mcap-support/src/parseChannel.ts
+++ b/packages/mcap-support/src/parseChannel.ts
@@ -14,7 +14,7 @@ import { parseJsonSchema } from "./parseJsonSchema";
 import { protobufDefinitionsToDatatypes, stripLeadingDot } from "./protobufDefinitionsToDatatypes";
 import { RosDatatypes } from "./types";
 
-import cbor from "cbor";
+var CBOR = require("cbor-js");
 
 type Channel = {
   messageEncoding: string;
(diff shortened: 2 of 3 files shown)


defunctzombie:
@jtbandes I've tagged you as a reviewer and assigned it to you as well. I suspect there will be some related work around adding items to mcap spec (if we decide to go that route) and documentation. I suggest making a project for this in our linear team and we can figure out how best to schedule the additional work and whether and of that work needs to block this PR merging. Probably a good item to discuss at the team planning.

jtbandes:
One question that comes to mind with this change is how to handle `bytes`. Byte strings can be encoded directly in CBOR, however I don't know that there is a jsonschema equivalent to represent them. For json encoded messages, we support `bytes` fields via `{ type: "string", contentEncoding: "base64" }` in json schema. If we're using the exact same json schema representation for CBOR support, then would someone using CBOR have to put `contentEncoding: "base64"` in their json schema even though they are not using base64 encoding? Will the `postprocessValue` function crash when it receives a Uint8Array when it was expecting a string?

therishidesai:
That’s a good question. I wonder if there is a better way to infer the schema from the first processed CBOR message instead of assuming a jsonschema is attached.

defunctzombie:
> I wonder if there is a better way to infer the schema from the first processed CBOR message instead of assuming a jsonschema is attached.

The downside of this approach is that you don't get autocomplete on any messages until it is seen. For some lower frequency messages you would not know what might even be in the message until it is seen.

> { type: "string", contentEncoding: "base64" }

Can we use `contentEncoding` of `octet-stream` or whatever the appropriate "this is binary data" value would be? or possibly `{type: string, format: binary}` ?

jtbandes:
Per https://github.com/json-schema-org/json-schema-spec/issues/6, the most appropriate way of doing this may be with a custom keyword. It does not fit within a standard `type` offered by JSON Schema and it does not make sense to use `{"type":"string", "contentEncoding":"base64"}` since the data is not a `string`.

therishidesai:
Just bumping this PR again, was wondering if the foxglove team was able to come up with some good ideas on supporting CBOR in the UI? [CDDL](https://datatracker.ietf.org/doc/rfc8610/) may be something worth looking at instead of using jsonschema

jtbandes:
CDDL seems interesting, but also quite complex and given the state of [the JS implementation](https://www.npmjs.com/package/cddl) is officially "don't use this for anything yet!" I'm hesitant to say we should start using it. It looks like there is [a webassembly implementation](https://github.com/anweiss/cddl) but there aren't many docs/examples on how to use it from JS. It seems like JSON Schema with a custom keyword is probably a simpler path. We could support CDDL in the future as a non-breaking change.

jon-chuang:
> If we're using the exact same json schema representation for CBOR support, then would someone using CBOR have to put contentEncoding: "base64" in their json schema even though they are not using base64 encoding?

Can't we just add a "bytes" type, extending the JSON schema?

jtbandes:
My feeling is that we should probably not accept/rely on a custom derivative of JSON Schema and still call it a `jsonschema` encoding, although others are welcome to debate this. As I read it the JSON Schema spec does not allow custom `type`s, but it does allow custom keywords.

jon-chuang:
It would be nice to have a schema that could simultaneously handle json, avro, CBOR and msgpack.

therishidesai:
Should we close this PR in favor of https://github.com/foxglove/studio/pull/5574

defunctzombie:
Closing in favor of https://github.com/foxglove/studio/pull/5574

Thanks for doing some early work on this!

Issue #4908, opened 2022-11-30
Support for `/clock` topic when using foxglove_bridge


jtbandes:

jtbandes:
Moved to Linear

PR #4984, opened 2022-12-14
Foxglove websocket: Fix flickering when receiving messages with interleaved timestamps

### LINKED ISSUE #3194: Allow interleaving times in topics with websocket protocol
I would like to be able to have Studio subscribe to two websocket topics, e.g. `/a` and `/b`, and then have the following messages come in in this order, and without having the display cleared:

```
/a: timestamp 1.0 sec
/b: timestamp 0.5 sec
/a: timestamp 2.0 sec
```

And then if I have a 2-D plot or just am looking at raw messages of `/a` and `/b` I would like to have everything still displayed.

Essentially, I am happy to have the server guarantee that an individual topic will send all of its messages across the websocket in order (and I think that's consistent with most pub-sub frameworks out there), but it is much more obnoxious to guarantee that messages across *all* topics are sorted before sending them across.

### LINKED ISSUE #4917: Extension panels repeatedly seeing didSeek=true when connected to a data source that does not support seeking (ws-protocol)
1. Connect to `foxglove_bridge` and subscribe to at least two topics, one lightweight and one heavyweight (such as an image stream).
2. Observe 3D panel and Image panel repeatedly flashing/strobing

This is due to many frames coming in with `didSeek=true`. Any time a message is received on the WebSocket with a message timestamp less than the previous message timestamp, this happens:

```typescript
// If time goes backwards, increment lastSeekTime and discard unemitted messages from before
// the discontinuity. This prevents us from queueing an unbounded number of messages when
// servers loop over the same recorded data multiple times. However, for now the queue can
// still grow unboundedly in a live system if the listener is not processing messages (such
// as when the app is hidden/backgrounded).
if (this._currentTime && isLessThan(receiveTime, this._currentTime)) {
  ++this._lastSeekTime;
  this._parsedMessages = [];
}
```

achim-k:
**User-Facing Changes**
- Foxglove websocket: Fix flickering when receiving messages with interleaved timestamps

**Description**
With this PR, we ignore timestamps on incoming messages and instead set the current time to
- the last server published time message (if the server advertised the `time` capability)
- the current system time (`new Date()`)

This prevents seeks from being detected when messages with interleaving timestamps (from multiple topics) are received

Fixes foxglove/community#239
Fixes foxglove/studio#4917
Fixes foxglove/studio#4908

### DIFF
--- a/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
+++ b/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
@@ -63,6 +63,8 @@ export default class FoxgloveWebSocketPlayer implements Player {
 
   /** Earliest time seen */
   private _startTime?: Time;
+  /** Latest time seen */
+  private _endTime?: Time;
   /* The most recent published time, if available */
   private _clockTime?: Time;
   private _serverPublishesTime = false;
@@ -126,6 +128,7 @@ export default class FoxgloveWebSocketPlayer implements Player {
       log.info("Connection closed:", event);
       this._presence = PlayerPresence.RECONNECTING;
       this._startTime = undefined;
+      this._endTime = undefined;
       this._clockTime = undefined;
       this._serverPublishesTime = false;
 
@@ -299,19 +302,18 @@ export default class FoxgloveWebSocketPlayer implements Player {
     });
 
     client.on("time", ({ timestamp }) => {
-      if (this._serverPublishesTime) {
-        const time = fromNanoSec(timestamp);
-
-        if (this._clockTime == undefined) {
-          this._startTime = time;
-        } else if (isLessThan(time, this._clockTime)) {
-          ++this._lastSeekTime;
-          this._parsedMessages = [];
-        }
-
-        this._clockTime = time;
-        this._emitState();
+      if (!this._serverPublishesTime) {
+        return;
       }
+
+      const time = fromNanoSec(timestamp);
+      if (this._clockTime != undefined && isLessThan(time, this._clockTime)) {
+        ++this._lastSeekTime;
+        this._parsedMessages = [];
+      }
+
+      this._clockTime = time;
+      this._emitState();
     });
   };
 
@@ -364,6 +366,13 @@ export default class FoxgloveWebSocketPlayer implements Player {
     }
 
     const currentTime = this._getCurrentTime();
+    if (this._startTime == undefined) {
+      this._startTime = currentTime;
+    }
+    if (this._endTime == undefined || isGreaterThan(currentTime, this._endTime)) {
+      this._endTime = currentTime;
+    }
+
     const messages = this._parsedMessages;
     this._parsedMessages = [];
     return this._listener({
@@ -382,8 +391,8 @@ export default class FoxgloveWebSocketPlayer implements Player {
       activeData: {
         messages,
         totalBytesReceived: this._receivedBytes,
-        startTime: this._startTime ?? ZERO_TIME,
-        endTime: currentTime,
+        startTime: this._startTime,
+        endTime: this._endTime,
         currentTime,
         isPlaying: true,
         speed: 1,
@@ -488,6 +497,6 @@ export default class FoxgloveWebSocketPlayer implements Player {
   public setGlobalVariables(): void {}
 
   private _getCurrentTime(): Time {
-    return this._clockTime ?? fromMillis(Date.now());
+    return this._serverPublishesTime ? this._clockTime ?? ZERO_TIME : fromMillis(Date.now());
   }
 }
--- a/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
+++ b/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
@@ -9,7 +9,7 @@ import { v4 as uuidv4 } from "uuid";
 import { debouncePromise } from "@foxglove/den/async";
 import Log from "@foxglove/log";
 import { parseChannel, ParsedChannel } from "@foxglove/mcap-support";
-import { fromNanoSec, isGreaterThan, isLessThan, Time } from "@foxglove/rostime";
+import { fromMillis, fromNanoSec, isGreaterThan, isLessThan, Time } from "@foxglove/rostime";
 import PlayerProblemManager from "@foxglove/studio-base/players/PlayerProblemManager";
 import {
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
```diff
@@ -303,6 +297,22 @@ export default class FoxgloveWebSocketPlayer implements Player {
       }
       this._emitState();
     });
+
+    client.on("time", ({ timestamp }) => {
+      if (this._serverPublishesTime) {
```
achim-k, line 302: This check is not necessary as the server is only allowed to publish time messages when it advertises the `time` capability

packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
```diff
@@ -145,6 +150,7 @@ export default class FoxgloveWebSocketPlayer implements Player {
 
     client.on("serverInfo", (event) => {
       this._name = `${this._url}\n${event.name}`;
+      this._serverPublishesTime = event.capabilities.includes(ServerCapability.time);
```
achim-k, line 153: @jhurliman I think we do not have to check the capability here. Just the reception of the first time message is enough, no?

jhurliman: We need to pick one mode and stick with it for the entire session. So waiting for the first time message won't work, because the spec doesn't guarantee the time message will be sent before any other messages.

packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
```diff
@@ -236,7 +242,8 @@ export default class FoxgloveWebSocketPlayer implements Player {
       this._emitState();
     });
 
-    client.on("message", ({ subscriptionId, timestamp, data }) => {
+    client.on("message", ({ subscriptionId, data }) => {
+      this._receivedBytes += data.byteLength;
```
jhurliman, line 246: Good catch

packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
jhurliman: We should keep this check because we can't switch the behavior of time handling mid-session. I would change it to an `if (!this._serverPublishesTimes) { return; }` though to avoid indenting the whole callback.

packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
```diff
@@ -372,8 +383,8 @@ export default class FoxgloveWebSocketPlayer implements Player {
         messages,
         totalBytesReceived: this._receivedBytes,
         startTime: this._startTime ?? ZERO_TIME,
-        endTime: this._endTime ?? ZERO_TIME,
-        currentTime: this._currentTime ?? ZERO_TIME,
+        endTime: currentTime,
```
jhurliman, line 386: I think we lost the logic for "`/clock`" mode where endTime can be > currentTime

packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
```diff
@@ -475,4 +486,8 @@ export default class FoxgloveWebSocketPlayer implements Player {
   }
 
   public setGlobalVariables(): void {}
+
+  private _getCurrentTime(): Time {
+    return this._clockTime ?? fromMillis(Date.now());
```
jhurliman, line 491: This doesn't look right. If we are in "clock mode" but _clockTime hasn't been set yet, it will return the current timestamp and then switch to a different time source mid-session once the first time message is received. One option is to remove the `_serverPublishesTime` boolean and instead set `_clockTime` to 0.

jhurliman (changes requested):

defunctzombie:
From the point of view of clients this is breaking behavior in studio. We no longer use the receiveTime they send whereas before we did. I don't have intuition one way or the other but I do want to call that out so we are deliberate in what we are doing here and prepared to communicate accordingly if questions come up around this.

@jtbandes

jtbandes:
Is there some way we could keep the current behavior as default for a while but get people to switch, or change the default but provide some backwards compatible mode?

packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
```diff
@@ -303,6 +300,21 @@ export default class FoxgloveWebSocketPlayer implements Player {
...
     });
+
+    client.on("time", ({ timestamp }) => {
+      if (!this._serverPublishesTime) {
+        return;
+      }
+
+      const time = fromNanoSec(timestamp);
+      if (this._clockTime != undefined && isLessThan(time, this._clockTime)) {
+        ++this._lastSeekTime;
```
jhurliman, line 311: Calling this variable `Time` when it's really an incrementing counter is confusing. We should either rename the variable or use it to store actual timestamps.

packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
```diff
@@ -353,6 +365,14 @@ export default class FoxgloveWebSocketPlayer implements Player {
       });
     }
 
+    const currentTime = this._getCurrentTime();
+    if (this._startTime == undefined) {
+      this._startTime = currentTime;
```
jhurliman, line 370: It looks like this does not correctly handle the case where you connect part-way through bag playback, then the time loops back to a timestamp earlier than what the client has observed.

jtbandes:
Decisions and context from team meeting today:
- We do not believe there are a large number of users depending on the old behavior, so it's okay to break it by switching to use the system clock instead of the receiveTime sent by the server.
  - However, we need to provide a smooth migration path, so we should add support for the new clock op first, talk about it and the planned change in release notes, and then switch to use the system clock by default after another 1 or 2 releases (rather than doing them at the same time as currently done in this PR).
- We do not believe there are any significant app behaviors that will work differently when using the system clock instead of the receiveTime.

achim-k:
> However, we need to provide a smooth migration path, so we should add support for the new clock op first, talk about it and the planned change in release notes, and then switch to use the system clock by default after another 1 or 2 releases (rather than doing them at the same time as currently done in this PR).

I will close this PR and open two separate PRs to impoement the proposed migration strategy

Issue #4799, opened 2022-11-10
Misleading message frequency statistics for live connections


jhurliman:
Message frequencies shown in the "Data source" sidbar are based on message count over the full duration from start time to end time. This makes sense for static data sources like recording files or data platform. For a live connection, if you switch tabs or put the computer to sleep or do something that temporarily pauses or slows message delivery while the wall clock duration of playback continues to increase, the reported message frequency will be unexpectedly low and remain lower than the actual frequency messages are coming in at until the connection is restarted.

Showing the frequency of recent data for live connections would be less surprising. For example, an exponential moving average of a sliding window of message times sampled at a fixed interval.

jhurliman:
Things get even weirder for a looping bag playback. The frequency (Hz) of all subscribed topics steadily increases as more messages are received with no apparent upper bound.

PR #5007, opened 2022-12-21
Improve message frequency display for live connections

foxymiles:
**User-Facing Changes**
Improves calculation of topic publishing frequencies for live connections.

**Description**
This replaces the static calculation of n / time for live messages with a moving average of frequency for topics of live connections. This gives better results for cases where messages are missed for any period of time, at the cost of some jitter in the actual displayed values.

We don't seem to explicitly mark a player as live so I'm inferring that from the absence of the `playbackControl` player capability.

Fixes https://github.com/foxglove/studio/issues/4799

### DIFF
--- a/packages/studio-base/src/components/DirectTopicStatsUpdater.tsx
+++ b/packages/studio-base/src/components/DirectTopicStatsUpdater.tsx
@@ -51,7 +51,7 @@ function calculateStaticItemFrequency(
   }
 }
 
-function calculateLiveitemFrequency(numMessages: number, duration: Time) {
+function calculateLiveItemFrequency(numMessages: number, duration: Time) {
   const durationSec = toSec(duration);
 
   return durationSec > 0 ? numMessages / durationSec : undefined;
@@ -95,7 +95,7 @@ export function DirectTopicStatsUpdater({ interval = 1 }: { interval?: number })
   const latestStats = useLatest(topicStats);
   const updateCount = useRef(0);
   const rootRef = useRef<HTMLDivElement>(ReactNull);
-  const samples = useRef<Record<string, StatSample>>({});
+  const samplesByTopic = useRef<Record<string, StatSample>>({});
 
   const playerIsStaticSource = useMemo(
     () => playerCapabilities.includes(PlayerCapabilities.playbackControl),
@@ -143,11 +143,11 @@ export function DirectTopicStatsUpdater({ interval = 1 }: { interval?: number })
           }
         } else {
           // For a live source we calculate a running average of frequency.
-          const sample = samples.current[topic];
+          const sample = samplesByTopic.current[topic];
           if (stat == undefined) {
             field.innerText = EM_DASH;
           } else if (sample == undefined) {
-            samples.current[topic] = {
+            samplesByTopic.current[topic] = {
               time: thisCurrentTime,
               count: stat.numMessages,
               frequency: undefined,
@@ -156,7 +156,7 @@ export function DirectTopicStatsUpdater({ interval = 1 }: { interval?: number })
             const messageDelta = stat.numMessages - sample.count;
             if (messageDelta > 0) {
               const timeDelta = subtractTimes(thisCurrentTime, sample.time);
-              const newFrequency = calculateLiveitemFrequency(messageDelta, timeDelta);
+              const newFrequency = calculateLiveItemFrequency(messageDelta, timeDelta);
               if (newFrequency != undefined) {
                 const smoothedFrequency = smoothValues(sample.frequency, newFrequency);
                 sample.frequency = smoothedFrequency;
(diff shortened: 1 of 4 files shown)


packages/studio-base/src/components/DirectTopicStatsUpdater.tsx
```diff
@@ -50,53 +82,107 @@ const selectTopicStats = (ctx: MessagePipelineContext) =>
...
   const latestDuration = useLatest(duration);
   const latestStats = useLatest(topicStats);
   const updateCount = useRef(0);
   const rootRef = useRef<HTMLDivElement>(ReactNull);
+  const samples = useRef<Record<string, StatSample>>({});
+
+  const playerIsStaticSource = useMemo(
+    () => playerCapabilities.includes(PlayerCapabilities.playbackControl),
+    [playerCapabilities],
+  );
```
jtbandes, line 103: > We don't seem to explicitly mark a player as live so I'm inferring that from the absence of the playbackControl player capability.

Another way we could do this would be to just add an explicit property to the players. Do you think that would be better? This way is probably fine too, it would just break down if we ever wanted a player to support playback control but still use the moving average stats calculation (or if we wanted a static player to not allow playback control).

packages/studio-base/src/components/DirectTopicStatsUpdater.tsx
```diff
@@ -11,35 +11,67 @@ import {
...
   const topicDurationSec = toSec(subtractTimes(lastMessageTime, firstMessageTime));
-  return `${((numMessages - 1) / topicDurationSec).toFixed(2)} Hz`;
+  if (topicDurationSec > 0) {
+    return (numMessages - 1) / topicDurationSec;
+  } else {
+    return undefined;
+  }
+}
+
+function calculateLiveitemFrequency(numMessages: number, duration: Time) {
```
jtbandes, line 54:
```suggestion
function calculateLiveItemFrequency(numMessages: number, duration: Time) {
```

packages/studio-base/src/components/DirectTopicStatsUpdater.tsx
```diff
@@ -50,53 +82,107 @@ const selectTopicStats = (ctx: MessagePipelineContext) =>
...
+  const playerCapabilities = useMessagePipeline(selectPlayerCapabilities);
+  const playerId = useMessagePipeline(selectPlayerId);
   const duration = endTime && startTime ? subtractTimes(endTime, startTime) : { sec: 0, nsec: 0 };
 
+  const latestCurrentTime = useLatest(currentTime);
   const latestDuration = useLatest(duration);
   const latestStats = useLatest(topicStats);
   const updateCount = useRef(0);
   const rootRef = useRef<HTMLDivElement>(ReactNull);
+  const samples = useRef<Record<string, StatSample>>({});
```
jtbandes, line 98: I find things like this easier to read if the key is part of the name:
```suggestion
  const samplesByTopic = useRef<Record<string, StatSample>>({});
```

jtbandes (approved):
Would it be possible to add a story or unit tests to show the expected frequencies are calculated? I'm imagining the test would include a hardcoded list of player updates in the form of `(time, numMessages)` pairs, and show that the resulting frequency calculation is sensible.

packages/studio-base/src/components/DirectTopicStatsUpdater.tsx
foxymiles: I was reluctant to touch every player for this but that would certainly be a more robust approach.

jtbandes: I think it's fine, just something to keep in mind if we start fleshing out the websocket player with more features like playback control for pre-recorded data.

packages/studio-base/src/hooks/useTopicPublishFrequences.test.tsx
```diff
@@ -0,0 +1,91 @@
...
+      currentTime: { sec: 2, nsec: 0 },
+      endTime: { sec: 10, nsec: 0 },
+      startTime: { sec: 0, nsec: 0 },
+      topicStats: new Map([
+        ["topic_a", { numMessages: 10 }],
+        ["topic_b", { numMessages: 20 }],
+      ]),
+    };
+
+    (useMessagePipeline as jest.Mock).mockImplementation(
```
jtbandes, line 68: Instead of mocking out this function can you use a wrapper for renderHook which uses MockMessagePipelineProvider? Example: https://github.com/foxglove/studio/blob/58077209d8eecef43a50f596c47f1d43b3a9fa0f/packages/studio-base/src/PanelAPI/useBlocksByTopic.test.tsx#L37

This seems like a slightly more stable way to do it since it's overriding via the context rather than mocking an implementation detail of the hook. Well, maybe both are kinda implementation details. But at least using the wrapper would match our other tests in studio :)

PR #5008, opened 2022-12-21
Foxglove websocket: Use server published time as message receive timestamps

achim-k:
**User-Facing Changes**
- Foxglove websocket: Use server published time as message receive timestamps, if available

**Description**
Since https://github.com/foxglove/ws-protocol/pull/299, the foxglove websocket server may send time messages to clients. If this is the case, Studio will use the server-published time as the message receive timestamp instead of the actual message receive time. This helps in preventing undesired “clearing” issues caused by messages being out of order across channels. The player behavior remains the same if the server does not provide time information to clients.

Supersedes #4984 

Related issues:
- https://github.com/foxglove/community/issues/239
- https://github.com/foxglove/studio/issues/4917
- https://github.com/foxglove/studio/issues/4908

---

**No time published - noticeable flickering**

[No time published - noticeable flickering](https://user-images.githubusercontent.com/9250155/208963660-568147f8-bf86-44c6-bb53-c261fed7643c.webm)

**Server publishes time - flickering is gone**

[Server publishes time - flickering is gone](https://user-images.githubusercontent.com/9250155/208964276-831b8bc1-cb88-423f-b9bd-58bc5b0f0d6f.webm)

### DIFF
--- a/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
+++ b/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
@@ -268,7 +268,9 @@ export default class FoxgloveWebSocketPlayer implements Player {
       }
 
       try {
-        const receiveTime = fromNanoSec(timestamp);
+        const receiveTime = this._serverPublishesTime
+          ? this._clockTime ?? ZERO_TIME
+          : fromNanoSec(timestamp);
         const topic = chanInfo.channel.topic;
         // If time goes backwards, increment lastSeekTime and discard unemitted messages from before
         // the discontinuity. This prevents us from queueing an unbounded number of messages when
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
```diff
@@ -303,6 +318,21 @@ export default class FoxgloveWebSocketPlayer implements Player {
...
+      }
+
+      const time = fromNanoSec(timestamp);
+      if (this._clockTime != undefined && isLessThan(time, this._clockTime)) {
+        ++this._lastSeekTime;
+        this._parsedMessages = [];
+      }
+
+      this._clockTime = time;
+      this._emitState();
```
jtbandes, line 334: There's some extra logic in handling the receiveTime above:

```
        if (!this._startTime || isLessThan(receiveTime, this._startTime)) {
          this._startTime = receiveTime;
        }
        if (!this._endTime || isGreaterThan(receiveTime, this._endTime)) {
          this._endTime = receiveTime;
        }
```
I think we should use this logic for the server `time` if present, and `this._currentTime` should be updated with the server time, so that the data source sidebar matches the server time.

achim-k: I see what you mean. Right now `this._currentTime` would only be updated when receiving new messages but we also want to do that for new `time` messages. I implemented that in #5012 but I can move parts of it already in this PR (or rather #5013)

PR #5016, opened 2022-12-21
Re-enable LayoutBrowser stories

jtbandes:
**User-Facing Changes**
None

**Description**
Follow-up from #4052
Fixes FG-1083


defunctzombie:
@jtbandes Something you want to throw in linear and get done?

jtbandes:
React 18 upgrade was reverted so I guess this isn't an issue....for now.

PR #5023, opened 2022-12-21
Update readme badges

jtbandes:
**User-Facing Changes**
None

**Description**
Before

image

After

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,8 +4,8 @@
 
 <div align="center">
     <h1>Foxglove Studio</h1>
+    <a href="https://github.com/foxglove/studio/releases"><img src="https://img.shields.io/github/v/release/foxglove/studio?label=version" /></a>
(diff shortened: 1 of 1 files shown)


ocin-rye (approved):
Looks good!

amacneil:
the `ROS 1 & 2 supported` one sort of implies to me that we are ros-specific. maybe drop that?

you could also add one for `version` if you want

jtbandes:
I was thinking of making it more attractive to ROS users with the ROS badge, but if you think it makes it appear too ROS-specific, I'll drop that

banisadr:
Looks good to me. Not strictly necessary, but is there a SO overflow badge? I really want to get that public knowledge base going.

banisadr (approved):

jtbandes:
I'll let you make a PR if you desire to add another badge 😄

PR #5043, opened 2022-12-22
Add ci for catching broken web deployments

snosenzo:
**User-Facing Changes**

-n/a

**Description**
Created an integration test in much of the same way as the electron integration test. The notable differences here would be that this ci needs to wait until after the vercel deployment to test against it. This check should happen with each vercel deployment

FG-1088


.github/workflows/web-integration.yml
```diff
@@ -0,0 +1,37 @@
...
+  deployment_status:
+
+jobs:
+  run-e2es:
+    if: github.event_name == 'deployment_status' && github.event.deployment_status.state == 'success'
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-20.04]
```
jtbandes, line 13: maybe remove matrix since we're not really using it

.github/workflows/web-integration.yml
```diff
@@ -0,0 +1,37 @@
+name: Web Integration Tests
+
+on:
+  deployment_status:
+
+jobs:
+  run-e2es:
```
jtbandes, line 7:
```suggestion
  test:
```

.github/workflows/web-integration.yml
```diff
@@ -0,0 +1,37 @@
+name: Web Integration Tests
+
+on:
+  deployment_status:
```
jtbandes, line 4: maybe a comment about how this relates to vercel?

jtbandes (approved):

Issue #4851, opened 2022-11-18
Image rendering does not update when CameraCalibration changes


jtbandes:
**Steps To Reproduce**
- Run this script: https://gist.github.com/jtbandes/b549f560679708507b13fb1633c4fac0
- Connect Studio using Foxglove WebSocket connection, open a 3D panel, turn on both the RawImage and CameraCalibration topics
- Edit the script to change `fx` and `fy`
- Kill the script and run it again, allowing studio to auto-reconnect

**Expected Behavior**
Image and camera frustum should always match

**Actual Behavior**
Image does not update but frustum does.
<img width="1114" alt="image" src="https://user-images.githubusercontent.com/14237/202587484-f6002ddd-6d95-4dc4-87f6-356dee08c297.png">

danielgordon10:
If you change the distance then the image also goes back to working as expected.
<img width="349" alt="Screenshot 2022-11-17 at 4 41 10 PM" src="https://user-images.githubusercontent.com/7245472/202590184-dda00f3f-2c0d-4b25-b7e4-b14875d14666.png">

PR #5044, opened 2022-12-22
[3D] Replace forced subscription param with shouldSubscribe callback, separate Images and Cameras SceneExtensions

### LINKED ISSUE #4393: Avoid subscribing to unnecessary topics in 3d panel.
From a slack converstation:

The 3D panel always subscribes to camera info topics even when no camera image or camera info is turned on. If a user is not visualizing any camera stuff in the 3d panel we don’t want to be streaming the camera information back to the app. Just like we don’t want to be streaming camera images or other random topics that are not being put towards visualizing the scene.

We've agreed that the 3D panel should avoid subscribing to these topics when it does not need to use them.

jhurliman:
**User-Facing Changes**
- [3D] CameraInfo topics are no longer automatically subscribed
- [3D] Fixes a bug where new CameraInfo messages would not update Image topics in the 3D view

**Description**
The boolean `forced` parameter has been replaced with a `shouldSubscribe()` callback, giving SceneExtensions more flexibility on when subscriptions occur. The Images 
SceneExtension builds its own PinholeCameraModel, decoupling it from the Cameras SceneExtension and fixing a bug with the camera model not updating.

Fixes #4393, #4851


packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
```diff
@@ -616,7 +616,9 @@ export function ThreeDeeRender({ context }: { context: PanelExtensionContext }):
       rendererSubscription: RendererSubscription,
       convertTo?: string,
     ) => {
-      if (rendererSubscription.forced === true || config.topics[topic]?.visible === true) {
+      const shouldSubscribe =
+        rendererSubscription.shouldSubscribe ?? ((t) => config.topics[t]?.visible === true);
+      if (shouldSubscribe(topic)) {
```
jtbandes, line 621: This API seems a bit unclear and bug-prone. If I might not want to subscribe, why am I calling a function named `addSubscription`? What are the expectations and guarantees around when the `shouldSubscribe` function will be called? If some internal state changes so that the `shouldSubscribe` function would return a different result, how do I notify the renderer of that change so it actually takes effect?

jhurliman: `addSubscription` does not automatically create a subscription, it registers that a SceneExtension would like a subscription to be created when the correct conditions have been met (and the subscription to be removed when those conditions no longer hold true). Previously, the condition was the `forced` parameter being true or topic visibility enabled in the settings sidebar for the associated topic. After this PR, the condition can be a SceneExtension-provided callback instead.

The `shouldSubscribe` method(s) will be called when the list of topics changes or any topic configuration changes. There is no ability to force `shouldSubscribe` to be called, other than making a modification to topic configuration.

packages/studio-base/src/panels/ThreeDeeRender/renderables/Images.ts
```diff
@@ -711,3 +757,42 @@ function normalizeCompressedImage(message: PartialMessage<CompressedImage>): Com
     data: normalizeByteArray(message.data),
   };
 }
+
+class BiMap<TKey1, TKey2> {
```
jtbandes, line 761: BiMap should be in its own file (under den/collection?) and have some unit tests.

packages/studio-base/src/panels/ThreeDeeRender/renderables/Images.ts
```diff
@@ -104,6 +104,7 @@ export class ImageRenderable extends Renderable<ImageUserData> {
 
 export class Images extends SceneExtension<ImageRenderable> {
   private cameraInfoTopics = new Set<string>();
+  private cameraInfoToImageTopics = new BiMap<string, string>();
```
jtbandes, line 107: comment on what this is used for? I'm not sure I understand why it was necessary by glancing at this commit.

packages/studio-base/src/panels/ThreeDeeRender/renderables/Images.ts
```diff
@@ -711,3 +757,42 @@ function normalizeCompressedImage(message: PartialMessage<CompressedImage>): Com
...
+  }
+
+  public getByValue(key: TKey2): TKey1 | undefined {
+    return this.map2.get(key);
+  }
+
+  public set(key1: TKey1, key2: TKey2): void {
+    this.map1.set(key1, key2);
+    this.map2.set(key2, key1);
+  }
```
jtbandes, line 776: If there was already an entry for key1 with an oldKey2, what should happen to the oldKey2->key1 association in map2? It seems a bit weird to keep it if it only existed because of the `set(key1, oldKey2)` call which has been overridden with `set(key1, key2)`.

jhurliman: Yes, it should check for that and delete the dangling entry. This is supposed to enforce a 1:1 bi-directional mapping between two keys.

packages/studio-base/src/panels/ThreeDeeRender/renderables/Images.ts
```diff
@@ -191,16 +191,39 @@ export class Images extends SceneExtension<ImageRenderable> {
...
+    const image = renderable?.userData.image;
+    const cameraModel = renderable?.userData.cameraModel;
+    if (image && cameraModel) {
+      const receiveTime = renderable.userData.receiveTime;
+
+      this._updateImageRenderable(renderable, image, cameraModel, receiveTime, settings);
+    }
+  };
+
+  private shouldSubscribe = (cameraInfoTopic: string): boolean => {
```
snosenzo, line 217: nit: since there's a few different topics this extension deals with would it be better to be more specific and name is `cameraInfoShouldSubscribe` or `nonImageShouldSubscribe` or something along those lines?

packages/studio-base/src/panels/ThreeDeeRender/renderables/Images.ts
```diff
@@ -354,20 +379,14 @@ export class Images extends SceneExtension<ImageRenderable> {
...
-        // log.debug(
-        //   `Constructing geometry for ${cameraModel.width}x${cameraModel.height} camera image on "${topic}"`,
-        // );
-        const geometry = createGeometry(cameraModel, renderable.userData.settings);
-        renderable.userData.geometry = geometry;
-        if (renderable.userData.mesh) {
-          renderable.remove(renderable.userData.mesh);
-          renderable.userData.mesh = undefined;
-        }
+      // log.debug(
```
snosenzo, line 382: what's the reason for keeping this around?

snosenzo (approved):
couple of nits, but looks good to me otherwise

packages/den/collection/MultiMap.ts
```diff
@@ -0,0 +1,46 @@
...
+  #map = new Map<TKey, TValue[]>();
+
+  public get(key: TKey): TValue[] | undefined {
+    return this.#map.get(key);
+  }
+
+  public set(key: TKey, value: TValue): void {
+    const values = this.#map.get(key);
+    if (values != undefined) {
+      if (!values.includes(value)) {
```
defunctzombie, line 18: These could have been one `if` statement `values?.includes(value) == false` or `values != undefined && !values.includes(value)` to reduce the nesting

packages/den/collection/MultiMap.ts
```diff
@@ -0,0 +1,46 @@
...
+        values.push(value);
+      }
+    } else {
+      this.#map.set(key, [value]);
+    }
+  }
+
+  public delete(key: TKey, value: TValue): void {
+    const values = this.#map.get(key);
+    if (values != undefined) {
```
defunctzombie, line 28: Flipping this around to return early would reduce the nesting. For me that makes it easier to read

packages/den/collection/MultiMap.ts
jhurliman: That's not an equivalent logic flow. Luckily, the unit tests will catch this programming error if you try.

defunctzombie: Indeed - I was curious enough to propose a change anyway https://github.com/foxglove/studio/pull/5128 🤣

jtbandes: If we are doing `includes` each time, should we just use a Set?

Issue #4989, opened 2022-12-16
User script changes do not update until another message arrives


defunctzombie:
**Description**
I've created a layout with two raw message panels and a user script panel. I've written a basic user script using the default "new script" template which outputs `hello: world`. My script has one input topic and one output topic. One raw message panel is subscribed to my input topic and one to my output topic.

I now load a data source that contains messages on my input topic. I play through the data some and pause. The raw message panel shows me the expected output `hello: world`. I update the user script to change the string `world` to something else and click `save`. My output raw message panel does not update with the new message. If I start to play the output raw message panel finally updates.

**Expected Behavior**
The output raw message panel updates with the new message. I believe this behavior use to work (it does seem to work with webviz).

Maybe we've changed something in our user script panel or in our message pipeline that does not allow us to re-transmit the same "frame" again but it is worth investigating why we regressed.

jtbandes:
It appears that webviz was triggering a backfill when variables changed in the OrderedStampPlayer:

https://github.com/cruise-automation/webviz/blob/4ed917d47e57d4286d7aa4b5c783e3f26420446e/packages/webviz-core/src/players/OrderedStampPlayer.js#L183-L188

jtbandes:
Related: https://github.com/foxglove/studio/issues/1910

PR #5047, opened 2022-12-22
Fix IterablePlayer and UserNodePlayer bugs with seeking

jtbandes:
**User-Facing Changes**
- Fixed an issue where the app would sometimes not respect the `time` URL parameter.

**Description**
- Fixes IterablePlayer to support seek events that happen during initialization (such as the seek from `useSyncTimeFromUrl`), and to not use `{sec:0,nsec:0}` as a placeholder time before initialization is completed.
- Deletes a line from UserNodePlayer which was sometimes triggering a seek with the incorrect timestamp (if another seek had been requested after `lastPlayerStateActiveData` was captured). The code being removed was ostensibly supposed to update the user script output when the script was edited, but that's not working (https://github.com/foxglove/studio/issues/4989 / FG-1019) so it's simpler to delete this line. The way script updates are handled will need to be re-worked to solve https://github.com/foxglove/studio/issues/4989 and FG-1071.

### DIFF
--- a/packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
+++ b/packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
@@ -33,6 +33,7 @@ import {
   PlayerPresence,
   PlayerCapabilities,
   TopicStats,
+  PlayerStateActiveData,
 } from "@foxglove/studio-base/players/types";
 import { RosDatatypes } from "@foxglove/studio-base/types/RosDatatypes";
 import delay from "@foxglove/studio-base/util/delay";
@@ -112,8 +113,8 @@ export class IterablePlayer implements Player {
   private _isPlaying: boolean = false;
   private _listener?: (playerState: PlayerState) => Promise<void>;
   private _speed: number = 1.0;
-  private _start: Time = { sec: 0, nsec: 0 };
-  private _end: Time = { sec: 0, nsec: 0 };
+  private _start: Time | undefined;
+  private _end: Time | undefined;
   private _enablePreload = true;
 
   // next read start time indicates where to start reading for the next tick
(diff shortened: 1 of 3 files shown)


defunctzombie:
> * Fixed an issue where data would sometimes fail to pre-load in the Plot panel and other panels.

Do we think this is actually fixing this issue? My understanding is that this fixes a failure to _seek_ from the time specified in the URL. We _hope_ it is also going to fix the preload issue but we are not sure?

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
```diff
@@ -112,8 +113,8 @@ export class IterablePlayer implements Player {
   private _isPlaying: boolean = false;
   private _listener?: (playerState: PlayerState) => Promise<void>;
   private _speed: number = 1.0;
-  private _start: Time = { sec: 0, nsec: 0 };
-  private _end: Time = { sec: 0, nsec: 0 };
+  private _start: Time | undefined;
```
defunctzombie, line 116: Do we prefer this to `?`

defunctzombie (approved):

jtbandes:
> We _hope_ it is also going to fix the preload issue but we are not sure?

That would be more accurate. Would you like us to leave that bug open for a while and see if it comes up again, or close for now?

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
jtbandes: Not sure it makes a difference in this context. I can change to `?` if you prefer.

defunctzombie:
> Would you like us to leave that bug open for a while and see if it comes up again, or close for now?

I'd say we can merge this and give it a test with the layout we know reproduces the issue. If we cannot reproduce close out the bug.

PR #5050, opened 2022-12-22
Remove no-restricted-imports overrides from eslintrc

defunctzombie:
**User-Facing Changes**
None

**Description**
No files remain that use the restricted imports so we don't need these overrides.


defunctzombie:
:( I had such dreams. I thought all these were gone

PR #5052, opened 2022-12-22
Team -> Org

amacneil:
**User-Facing Changes**
Terminology in the sidebar is consistent with the rest of Foxglove


jtbandes (commented):
There are a handful of other places in the app where we have things like "Share a copy with your team", "Update team layout", etc.

amacneil:
@jtbandes good catch, updated. There are a few remaining uses of the word "team" where I felt it was justified.

banisadr (approved):
:shipit:

PR #5054, opened 2022-12-23
Revert "Upgrade to React 18.2.0 (#4052)"

jtbandes:
**User-Facing Changes**
None

**Description**
Aside from issues stemming from Strict Mode changes, we're seeing some panels completely freeze in dev mode in React 18 and triggering an infinite re-render loop. Just disabling Strict Mode is not enough to fix these issues – need to dig in further.


defunctzombie (approved):
Approving if you decide to merge but I'll leave it to you to determine if there are fixes we can make without revert if this only impacts dev.

PR #5055, opened 2022-12-23
Add robotics stack exchange badge

banisadr:
I desired another badge: https://github.com/foxglove/studio/pull/5023#issuecomment-1362371826


jtbandes:
Seems like it will end up with a lot of questions being created incorrectly using the [mcap] tag...

banisadr:
> Seems like it will end up with a lot of questions being created incorrectly using the [mcap] tag...

Good point, removed that.

jtbandes (approved):
Seems fine – did you consider linking to https://robotics.stackexchange.com/questions/tagged/foxglove-studio instead?

banisadr:
What's the better stack exchange practice here - linking to create a new question with appropriate tags, or listing existing questions with the tag? I'll defer to you as a member of the SO inner sanctum.

jtbandes:
I don't have any real guidance. Would look toward other sites that are linking to stack exchange for help — not sure I can think of them off the top of my head though.

As the first example I found, React docs link to both pages 🤦 https://reactjs.org/community/support.html#stack-overflow

amacneil:
I think it would make more sense to point people to https://github.com/orgs/foxglove/discussions/categories/studio, which we already do from https://github.com/foxglove/studio/issues/new/choose

PR #5056, opened 2022-12-23
Fix setState hook in PanelExtensionAdapter for React 18

jtbandes:
**User-Facing Changes**
None

**Description**
This fixes a freeze when using the 3D panel with React 18.
I'm not sure why the `.has()` check is necessary — it should be fine if multiple calls to this setter function return different Set instances. But in practice it is necessary 🤷


defunctzombie (approved):
probably worth a comment that we noticed bad behavior in react 18 constantly calling this setter and this workaround resolved that

PR #5013, opened 2022-12-21
Foxglove websocket: Use server published time as message receive timestamps

achim-k:
**User-Facing Changes**
- Foxglove websocket: Use server published time as message receive timestamps, if available

**Description**
Since https://github.com/foxglove/ws-protocol/pull/299, the foxglove websocket server may send time messages to clients. If this is the case, Studio will use the server-published time as the message receive timestamp instead of the actual message receive time. This helps in preventing undesired “clearing” issues caused by messages being out of order across channels. The player behavior remains the same if the server does not provide time information to clients.

Supersedes #4984 #5008 

Fixes FG-1113

Related issues:
- https://github.com/foxglove/community/issues/239
- https://github.com/foxglove/studio/issues/4917
- https://github.com/foxglove/studio/issues/4908

---

**No time published - noticeable flickering**

[No time published - noticeable flickering](https://user-images.githubusercontent.com/9250155/208963660-568147f8-bf86-44c6-bb53-c261fed7643c.webm)

**Server publishes time - flickering is gone**

[Server publishes time - flickering is gone](https://user-images.githubusercontent.com/9250155/208964276-831b8bc1-cb88-423f-b9bd-58bc5b0f0d6f.webm)

PR #5057, opened 2022-12-23
Foxglove websocket: Use system time if server does not publish time messages

### LINKED ISSUE #4908: Support for `/clock` topic when using foxglove_bridge

### LINKED ISSUE #4917: Extension panels repeatedly seeing didSeek=true when connected to a data source that does not support seeking (ws-protocol)

achim-k:
**User-Facing Changes**
- Foxglove websocket: Use system time if server does not publish time messages

**Description**
*For some reason #5012 got automatically closed and I can't reopen it. This PR replaces it*

This PR is based on #5013 and only to be merged after 1-2 releases after the release that contains #5013. It is the second part of the migration strategy proposed in https://github.com/foxglove/studio/pull/4984#issuecomment-1360366496

From #5013:

> Since https://github.com/foxglove/ws-protocol/pull/299, the foxglove websocket server may send time messages to clients. If this is the case, Studio will use the server-published time as the message receive timestamp instead of the actual message receive time. This helps in preventing undesired “clearing” issues caused by messages being out of order across channels. The player behavior remains the same if the server does not provide time information to clients.

This PR changes the behavior in the following way:
- The system time is used if the server does not provide time information to clients

Fixes foxglove/community#239
Fixes #4917
Fixes #4908
Fixes FG-1114

### DIFF
--- a/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
+++ b/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
@@ -305,7 +305,7 @@ export default class FoxgloveWebSocketPlayer implements Player {
 
       const time = fromNanoSec(timestamp);
       if (this._clockTime != undefined && isLessThan(time, this._clockTime)) {
-        ++this._lastSeekTime;
+        this._lastSeekTime = time.sec;
         this._parsedMessages = [];
       }
 
--- a/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
+++ b/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
@@ -9,7 +9,7 @@ import { v4 as uuidv4 } from "uuid";
 import { debouncePromise } from "@foxglove/den/async";
 import Log from "@foxglove/log";
 import { parseChannel, ParsedChannel } from "@foxglove/mcap-support";
-import { fromNanoSec, isGreaterThan, isLessThan, Time } from "@foxglove/rostime";
+import { fromMillis, fromNanoSec, isGreaterThan, isLessThan, Time } from "@foxglove/rostime";
 import PlayerProblemManager from "@foxglove/studio-base/players/PlayerProblemManager";
 import {
   AdvertiseOptions,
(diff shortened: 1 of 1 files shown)


jhurliman (approved):

packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
```diff
@@ -320,11 +305,11 @@ export default class FoxgloveWebSocketPlayer implements Player {
 
       const time = fromNanoSec(timestamp);
       if (this._clockTime != undefined && isLessThan(time, this._clockTime)) {
-        ++this._lastSeekTime;
+        this._lastSeekTime = time.sec;
```
achim-k, line 308: @jhurliman maybe better to use `time.nsec` here? Otherwise multiple seeks in the same second might not be detected?

```suggestion
        this._lastSeekTime = time.nsec;
```

jhurliman: `toNanoSec(time)` and make _lastSeekTime a bigint

jtbandes: I don't think this change will behave as intended. If the clock time resets from >0 to =0 multiple times, the `_lastSeekTime` will not change, which means that panels throughout the app will not be reset. If the concern is with the name of the variable, I'd suggest renaming the variable instead.

achim-k: That's a valid point, I will rename the variable and revert back to incrementing it on every seek

jtbandes (approved):

PR #5059, opened 2022-12-25
Add welcome dialog backdrop

amacneil:
**User-Facing Changes**
New welcome dialog backdrop effect

**Description**
Inspired by #2561, ported to Three.js + MUI

### DIFF
--- a/packages/studio-base/src/components/OpenDialog/OpenDialog.tsx
+++ b/packages/studio-base/src/components/OpenDialog/OpenDialog.tsx
@@ -8,6 +8,7 @@ import { useCallback, useLayoutEffect, useMemo, useState } from "react";
 import { useMountedState } from "react-use";
 import { makeStyles } from "tss-react/mui";
 
+import Snow from "@foxglove/studio-base/components/OpenDialog/Snow";
 import Stack from "@foxglove/studio-base/components/Stack";
 import { useAnalytics } from "@foxglove/studio-base/context/AnalyticsContext";
 import {
@@ -21,9 +22,10 @@ import Start from "./Start";
 import { OpenDialogViews } from "./types";
 import { useOpenFile } from "./useOpenFile";
 
-type OpenDialogProps = {
-  activeView?: OpenDialogViews;
+export type OpenDialogProps = {
   activeDataSource?: IDataSourceFactory;
+  activeView?: OpenDialogViews;
+  backdropAnimation?: boolean;
   onDismiss?: () => void;
 };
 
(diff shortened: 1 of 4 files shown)


packages/studio-base/src/components/OpenDialog/OpenDialog.tsx
```diff
@@ -21,9 +22,10 @@ import Start from "./Start";
 import { OpenDialogViews } from "./types";
 import { useOpenFile } from "./useOpenFile";
 
-type OpenDialogProps = {
-  activeView?: OpenDialogViews;
+export type OpenDialogProps = {
   activeDataSource?: IDataSourceFactory;
+  activeView?: OpenDialogViews;
+  backdropAnimation?: "disabled";
```
defunctzombie, line 28:
```suggestion
  disableBackdropAnimation?: boolean;
```

packages/studio-base/src/components/OpenDialog/OpenDialog.tsx
```diff
@@ -87,6 +89,19 @@ export default function OpenDialog(props: OpenDialogProps): JSX.Element {
...
 
+  const backdrop = useMemo(() => {
+    const now = new Date();
+    if (backdropAnimation === "disabled") {
+      return;
+    } else if (now >= new Date(now.getFullYear(), 11, 25)) {
+      return <Snow effect="snow" />;
+    } else if (now < new Date(now.getFullYear(), 0, 2)) {
+      return <Snow effect="confetti" />;
+    } else {
```
defunctzombie, line 100: You don't need this extra else

packages/studio-base/src/components/OpenDialog/OpenDialog.tsx
defunctzombie: Alternatively, flip this around and use `backdropAnimation?: boolean` and explicitly enable the backdrop in our components rather than disable in tests

defunctzombie (approved):

packages/studio-base/src/components/OpenDialog/OpenDialog.tsx
amacneil: I needed a final return to make the useMemo linter happy. Removed the `else {}` block to shave a line off.

packages/studio-base/src/components/OpenDialog/OpenDialog.tsx
amacneil: Changed to boolean but will keep it as `backdropAnimation` to avoid double negative. It can still default to true.

PR #5062, opened 2023-01-01
Analytics: update capitalization

amacneil:
**User-Facing Changes**
None

**Description**
Consistent with console


foxymiles (approved):
This won't invalidate existing events right?

defunctzombie (approved):
I'm tempted to saw we should lowercase all event names to remove any mental overhead in trying to format event names

esthersweon (approved):

Issue #5041, opened 2022-12-22
useConfirm doesn't work in dev mode


jtbandes:
Double-invoke of effects in Strict Mode causes the prompt to dismiss prematurely due to this code:

https://github.com/foxglove/studio/blob/341f69488fd6fb5e4c2af969966a69fbcafdfd26/packages/studio-base/src/hooks/useConfirm.tsx#L57-L61

The same issue probably affects usePrompt.

It would be better to get rid of ModalHost and set both of these up to return an element that the caller has to render in the react tree themselves.

foxymiles:
Do we even need that `useEffect`? I think the MUI dialog will still call `onClose` if the user closes via ESC key.

jtbandes:
I think the original reason for adding it was to handle if the parent component unmounts without the user pressing esc.

foxymiles:
In that case is it still important to invoke the callback?

PR #5066, opened 2023-01-04
Remove ModalHost/ModalContext, fix useConfirm in dev mode

jtbandes:
**User-Facing Changes**
None

**Description**
Update `useConfirm`, `usePrompt`, and `ProblemsList` to keep the modal element in `useState` rather than in the `ModalHost`. The caller of `useConfirm`/`usePrompt` is now passed a modal element in addition to a confirm/prompt function and is responsible for rendering that element in the tree. This allows for the deletion of `ModalHost` & `ModalContext`.

Fixes https://github.com/foxglove/studio/issues/5041
Fixes FG-1140


foxymiles (approved):
I don't think I've seen JSX elements in `useState` before. Is that an encouraged practice?

jtbandes:
Not sure I’ve ever heard it encouraged or discouraged — do you think it could be problematic?

foxymiles:
> Not sure I’ve ever heard it encouraged or discouraged — do you think it could be problematic?

It's certainly unconventional. Did you consider using a React portal for this? That's the most common solution I've seen for this kind of thing.

defunctzombie (commented):
How does this prevent modals from inheriting the z-index and other styling elements of the container you put them in? I suppose that would be true of any place you use a MUI `<Modal>` and it is something that the modal itself handles internally?

jtbandes:
> Did you consider using a React portal for this? That's the most common solution I've seen for this kind of thing.

I don’t think this comment is relevant to the useState issue? Isn’t it common to render mui Modals without a portal?

> How does this prevent modals from inheriting the z-index and other styling elements of the container you put them in? I suppose that would be true of any place you use a MUI `<Modal>` and it is something that the modal itself handles internally?

Same

Issue #4956, opened 2022-12-10
Move ROS 2 Native connection behind feature flag


jtbandes:

jtbandes:
Moved to Linear

Issue #4900, opened 2022-11-29
🌉 Remove ROS 2 native connector


jtbandes:
We intend to deprecate & remove the ROS 2 native connector in favor of the rosbridge or foxglove_bridge connections.

- [ ] foxglove_bridge feature parity with native connector
  - [ ] #4908
- [ ] #4901 
   - also to docs: FG-909, FG-910
- [ ] wait 1 month
- [ ] https://github.com/foxglove/studio/issues/4956
- [ ] wait 1 month
- [ ] #4902

jtbandes:
Closing in favor of tracking in Linear: https://linear.app/foxglove/project/replace-ros-2-native-connector-0a8c287407b4

PR #5017, opened 2022-12-21
Mark ROS2 native connection as deprecated

achim-k:
**User-Facing Changes**
- Mark ROS2 native connection as deprecated

**Description**
- Adds deprecation notice to ROS 2 connector
- Fixes the docs URL to be redirected directly to the ROS2 connection docs

#4901
FG-1042

PR #5070, opened 2023-01-04
Hide ROS 2 native data source behind feature flag

achim-k:
**User-Facing Changes**
- Hide ROS 2 native data source behind feature flag

**Description**
The ROS 2 native connection will be removed in the future, see #4900.
With this PR, the ROS 2 native connection will be hidden by default in the connection dialog. However, the user has the option to enable the ROS 2 connection via an additional feature flag in the app settings

Original issue: https://github.com/foxglove/studio/issues/4956
Fixes [FG-1043](https://linear.app/foxglove/issue/FG-1043/move-ros-2-native-connection-behind-feature-flag-%5Bfoxglovestudio-4956%5D)
Fixes FG-2114


jtbandes (commented):
Looks good, however we shouldn't merge until after #5017 has been released for a while (see rough timeline in https://github.com/foxglove/studio/issues/4900)

jtbandes:
#5017 was released today, so let's say we merge this after Fri Jan 27? cc @banisadr

banisadr:
I'm aligned with that.

jhurliman (approved):

defunctzombie:
The target date to merge this is ~March 10th.

Also - for the implementation - you don't need to make changes to the factory. The source can be excluded one layer higher in the `Root` component of desktop by excluding it from the list. The app configuration is available at that time: https://github.com/foxglove/studio/blob/main/desktop/renderer/Root.tsx#L66

packages/studio-base/src/i18n/zh/preferences.ts
```diff
@@ -33,6 +33,8 @@ export default {
   newNavigation: "新版导航栏",
   newNavigationDescription: "使用新版导航栏。",
   restartTheAppForChangesToTakeEffect: " 重新启动应用程序以使更改生效。",
+  ros2NativeConnection: "ROS 2 原生连接",
+  ros2NativeConnectionDescription: "启用已弃用的 ROS 2 本机实时连接。",
```
achim-k, line 37: From google translate, probably not correct. Who can review this?

defunctzombie: @Woodii1998 ^

jtbandes: I don't think you should worry about adding the chinese translation in this PR. It's probably better to have un-translated english rather than poorly translated chinese.

achim-k: > You should be able to add the strings to en but omit them from zh without getting any errors, right?

Let's try that now

```suggestion
```

jtbandes: See also: #5416

achim-k:
I have changed it now a little to follow @defunctzombie's request in FG-2114:

When the feature flag is disabled, we still show the connection entry in the list but disabled:

| Feature flag enabled                                                                                                                         | Feature flag disabled                                                                                                                        |
|----------------------------------------------------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------|
| Screenshot from 2023-03-10 10-55-34 | Screenshot from 2023-03-10 10-56-20 |

packages/studio-base/src/i18n/en/preferences.ts
```diff
@@ -38,6 +38,8 @@ export default {
   newNavigation: "New navigation",
   newNavigationDescription: "New top navigation bar.",
   restartTheAppForChangesToTakeEffect: " Restart the app for changes to take effect.",
+  ros2NativeConnection: "ROS 2 native connection",
+  ros2NativeConnectionDescription: "Enable the deprecated ROS 2 native live connection.",
```
defunctzombie, line 42:
```suggestion
  ros2NativeConnectionDescription: "Enable the deprecated ROS 2 native connector.",
```

defunctzombie (approved):

achim-k:
@defunctzombie I believe we can merge this one now?

packages/studio-base/src/i18n/zh/preferences.ts
jtbandes: I thought the plan was to remove these?

achim-k: > I thought the plan was to remove these?

Seems like I didn't pull and rebased the version without the commit that removes these lines.

```suggestion
```

defunctzombie:
@achim-k duplicating our Slack discussion here; we'll wait to merge this until 1.44.0 is released.

PR #5071, opened 2023-01-04
Saving User scripts updates subscribed panels

### LINKED ISSUE #4989: User script changes do not update until another message arrives

snosenzo:
**User-Facing Changes**

- Saving a user script will update subscribed panels with new output messages

**Description**

 Added state to remember last of each topic's message in the userNodePlayer and every time `setUserNodes` is called as an update to an existing node, that node will be added to a set of nodes that need to have their outputs refreshed. In the next `_onPlayerState` call this will read in the previous messages for each input topic and add them to the messages to be processed in `_getMessages` and then remove them from the list of nodes to update. `IterablePlayer` triggers this during `idle` state. This avoids an additional `_emitState` call and issues from my previous solutions with race conditions where previous playerStates were being overridden and old sourceCode was being shown.

Fixes #4989
FG-1019


defunctzombie (changes requested):
Please let's avoid adding a requestBackfill method to the player API. I felt pretty strongly that this was a weird anti-pattern for the player interface and worked to purge it from our code in favor of the player updating when it gets subscriptions.

If the user script needs to re-process the current frame I think it should do that within its own logic.

snosenzo:
> Please let's avoid adding a requestBackfill method to the player API. I felt pretty strongly that this was a weird anti-pattern for the player interface and worked to purge it from our code in favor of the player updating when it gets subscriptions.
> 
> If the user script needs to re-process the current frame I think it should do that within its own logic.

Ah okay, good to know. I refactored to make the `UserNodePlayer` keep track of last previous input messages and to reprocess them when the UserNodes are updated. Let me know if this isn't desirable, the last option I have in mind is to unsubscribe and re-subscribe to the input topic, but that felt a little hacky

packages/studio-base/src/players/UserNodePlayer/index.ts
```diff
@@ -796,8 +812,37 @@ export default class UserNodePlayer implements Player {
...
+          }
+        }
+        this._userNodeIdsNeedUpdate.clear();
+
+        for (const message of messages) {
+          this._lastMessageByInputTopic.set(message.topic, message);
+        }
+
+        const messagesToBeParsed =
+          messagesForRecompute.length > 0 ? messages.concat(messagesForRecompute) : messages;
```
defunctzombie, line 843: Does this not end up with some duplicate messages? Is the assumption here that `messages` is empty? If any items in `messages` match any items in `_lastMessageByInputTopic` then we end up with duplicate messages into the _getMessages call?

defunctzombie (commented):
Does this also run for preload messages nodes? So if I plot the output of a user script and then update that script, does my plot update?

snosenzo:
@defunctzombie 
> does my plot update

confirmed that it does work with plot view

> duplicate messages?

good catch -- I added some logic to make sure input topics to get messages from were unique and that it wouldn't use old messages if a message with a topic to be updated was in active state.

defunctzombie (approved):
Check that this also works on preloaded messages and I think this is good to go.

PR #5072, opened 2023-01-04
Fix extension id collisions across multiple loaders

### LINKED ISSUE #4967: Unable to load local extension when org extension installed/cached
hello again! has anyone experienced caching old code after publishing an extension, or know how I can load new local updates to an extension? Here is what I tried:
1. publish an extension to org
2. make new (local) changes to extension
3. attempt to load new local code by running `yarn local-install` + refresh Studio. (old code (version published to org) is still running when I open the [local] version)
4. try bumping `package.json` version, deleting everything in `.foxglove-studio/extensions/` , log out, quit and reopen Foxglove Studio app, and retry step 3 but changes are still not loaded
before this, yarn local-install + refresh studio had worked to reload changes, so not sure what happened, or if I’m missing a step. thank you in advance!

[Slack Message](https://foxglove.slack.com/archives/C028UEY858S/p1670873520807189?thread_ts=1670873520.807189&cid=C028UEY858S)

jtbandes:
**User-Facing Changes**
Fixed a bug that prevented the use of org-wide extensions and locally-installed extensions with the same ID.

**Description**
We were calling `getExtensions()` and then calling `loadExtension(id)` on a different loader.
This bug seems to have been present since the introduction of multiple loaders (https://github.com/foxglove/studio/pull/3836).

[View diff ignoring whitespace](https://github.com/foxglove/studio/pull/5072/files?diff=split&w=1)

Fixes #4967 
Fixes FG-934

### DIFF
--- a/packages/studio-base/src/providers/ExtensionCatalogProvider.tsx
+++ b/packages/studio-base/src/providers/ExtensionCatalogProvider.tsx
@@ -23,72 +23,68 @@ type ContributionPoints = {
   messageConverters: RegisterMessageConverterArgs<unknown>[];
 };
 
-async function activateExtensions(
-  extensions: ExtensionInfo[],
-  loadExtension: ExtensionLoader["loadExtension"],
-): Promise<ContributionPoints> {
+function activateExtension(
+  extension: ExtensionInfo,
+  unwrappedExtensionSource: string,
+): ContributionPoints {
   // registered panels stored by their fully qualified id
   // the fully qualified id is the extension name + panel name
   const panels: Record<string, RegisteredPanel> = {};
 
   const messageConverters: RegisterMessageConverterArgs<unknown>[] = [];
 
-  for (const extension of extensions) {
-    log.debug(`Activating extension ${extension.qualifiedName}`);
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/providers/ExtensionCatalogProvider.tsx
```diff
@@ -116,37 +112,35 @@ export function createExtensionRegistryStore(
...
     refreshExtensions: async () => {
       if (loaders.length === 0) {
         return;
       }
 
-      const extensionList = (
-        await Promise.all(loaders.map(async (loader) => await loader.getExtensions()))
-      )
-        .flat()
-        .sort();
```
jtbandes, line 140: I removed `sort()`. The default sort behavior would convert the items to strings, which would probably just be `[object Object]` for all of the items, so it wouldn't do much.

foxymiles (approved):

defunctzombie (commented):
👍 For future readability I think it could be nice to reduce the nesting on the business logic by extracting some of the loops or loop logic into functions. I think it would make it easier to follow by reading what each step is suppose to do as a function.

PR #5075, opened 2023-01-05
topnav: enable unified title bar on macOS

jtbandes:
**User-Facing Changes**
None (feature flagged)

**Description**
Return of https://github.com/foxglove/studio/pull/15
Resolves FG-1259

macOS:
image

Ubuntu:
image

### DIFF
--- a/desktop/common/types.ts
+++ b/desktop/common/types.ts
@@ -16,6 +16,8 @@ export type ForwardedMenuEvent =
   | "open-help"
   | "open-account";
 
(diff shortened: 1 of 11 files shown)


defunctzombie (approved):
I think we've done a disservice to our ubuntu users (by far the largest userbase) with these changes if we can't find a way to better leverage this space or use the titlebar space in the desktop app.

Approving since this is behind a feature flag and meant for experimentation.

jtbandes:
We can do similar work on Linux & Windows, it just means making our own custom window buttons since I believe the OS doesn't provide re-positioning like macOS does.

PR #5095, opened 2023-01-06
Add parameter support to foxglove websocket connection

achim-k:
**User-Facing Changes**
- Add parameter support to foxglove websocket connection

**Description**
This PR adds parameter support to the foxglove websocket connection according to the ws-protocol specification update in https://github.com/foxglove/ws-protocol/pull/288

[FG-1222](https://linear.app/foxglove/issue/FG-1222/add-parameter-support-to-foxglove-websocket-player)

### DIFF
--- a/packages/studio-base/package.json
+++ b/packages/studio-base/package.json
@@ -56,7 +56,7 @@
     "@foxglove/ulog": "2.1.2",
     "@foxglove/velodyne-cloud": "1.0.1",
     "@foxglove/wasm-bz2": "0.1.1",
-    "@foxglove/ws-protocol": "0.2.0",
+    "@foxglove/ws-protocol": "0.3.1",
     "@foxglove/xmlrpc": "1.3.0",
     "@mcap/core": "0.3.0",
     "@mdi/svg": "7.1.96",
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/panels/Parameters/index.tsx
```diff
@@ -196,7 +196,7 @@ function Parameters(): ReactElement {
                 <TableRow
                   hover
                   className={classes.tableRow}
-                  key={`parameter-${name}`}
+                  key={`parameter-${name}-${displayValue}`}
```
achim-k, line 199: I want the parameter panel to always show the latest value coming from the server. This change was necessary to force re-rendering when the value changes. However, the drawback is, that the value might get updated while still typing. 

IMO, we should not set parameters in the `onChange` event, but instead set them when <kbd>⏎</kbd> is pressed or on blur.

achim-k: image

achim-k: :point_up: I have pushed a preliminary version of the parameter panel panel that only sets the new parameter value when the submit button is pressed. It works, but it is far from perfect.

jtbandes (approved):

jhurliman (approved):

PR #5101, opened 2023-01-06
[Draft] TF Preloading up to current time 

snosenzo:
**User-Facing Changes**

 - add TF preloading toggle

**Description**
TF preloading spike approach

The approach is outlined as follows:

ThreeDeePanel was updated to read `preloaded` data from `allFrames` via a cursor that represents up-to where in `allFrames` data has been handled by the renderer. Whenever `allFrames` or the `currentTime` is updated the cursor will read from its starting place up to when either it reaches a message with `receiveTime` past the `currentTime` (where it will trigger a re-render) or it reaches the last message. If the `currentTime` was not reached it will wait for `allFrames` to be updated, at this time it will continue in the spot that the cursor left off until either of the two conditions have been met and so on. There is the case of block `eviction` that it needs to take into account however. If block eviction has happened then that means that blocks have been taken away

The assumptions made of `allFrames` are as follows (copying the comment:)
```
Assumptions about allFrames needed by allFramesCursor:
     *  - always sorted by receiveTime
     *  - preloaded topics/schemas are only ever all removed or all added at once, 
     *    otherwise it is not stable and would need to be reset
       */
```

The cases for `allFrames` updates are covered as so (copied from comment): 
How allFrames can change that this accounts for:

-  [new blocks being loaded after cursor (main case)] - new messages are added to end of array. The cursor iterates over the new part of the array
- [blocks being evicted from beginning] - messages before cursor have been removed and cursor is no longer pointing at last read message. cursor looks back for last read message and sets cursor to that index, or if last read message receiveTime is less than the first message it resets the cursor index
- [blocks being evicted from end] - messages after cursor have been removed. cursor is still pointing at last read message and does not need to look back
- [new blocks being loaded before the cursor] - This can only happen if the cache eviction has occurred on blocks at the beginning while the cursor was toward the end of the file, the user seeks to earlier in the file that hasn't been evicted. The cursor is reset to 0 before the blockloader has been able to re-load in the prior blocks and starts reading from an allFrames that doesn't begin at the end of the file up to the currentTime (it will render a possibly incorrect frame if a necessary FrameTransform from the beginning of the file is missing). allFrames will update and the cursor will see that there is a `lastReadMessage` mismatch and will look back until it finds the message, but the message will be after the cursor, so the cursor index will be set back to -1 and all of the `allFrames` messages will be read in again until the currentTime. This will happen for each missing block between the beginning of the file and the seek time. This is not ideal, but it should be a rare case. 

BlockLoader (responsible for all preload behavior) was changed so that its loading behavior will load from the beginning of the source to the currentTime (activeBlock) and then from the currentTime to the end of the file. Previously it loaded from the `activeBlock` to the end of the source and then from the beginning to the `activeBlock`.  This is necessary so that we can make assumptions about `allFrames` which is created in `renderState` by the 3D Panel Adapter from the messages within the blocks. This unfortunately does require all of the file up to the currentTime to have been read into memory for the renderFrame to be displayed.

Another thing I did to guarantee the assumption that preloading is all or nothing for the tf panel was filter out preloaded topics from being read into allFrames that weren’t specifically requested as `preload=true` for the given panel. Not sure if there’s a use-case to have a panel receive updates for messages from a topic that it isn’t subscribed to. 

The change to decrease the size of the `TransformTree` frame cache greatly decreases the time it takes to seek because most of that time (after the capacity has been reached) is taken performing `shift` operations. Decreasing the size of the cache greatly decreases its cost. Also arguably the cache doesn’t need to exist at all in this situation since the transform tree with capacity 1 for each frame should represent the state at the current time because the transforms should always have been read in from the beginning.

### Current Behavior and Improvements:

The current behavior I think is pretty good. It can read the end of an hour long file with ~250Hz transforms in less than ~2 seconds once the file has been loaded (also given that there’s no other data than transforms). For shorter files the seektime is essentially instantaneous. The longes times are from the beginning to end of a file or seeking slightly backward from the end. 

#### Definitely needs fixed before landing:
 - don’t want to show warnings for the transform cache, might just want to make the transform cache only hold 1 transform since we’ll never not have evaluated transforms up to a certain point and can assume all past transforms will be the last added to the transform tree
 - occasionally I’ve run into max loop depth errors in development while testing eviction of blocks with a smaller cache, I’m not sure if that’s because of some of my intermediate logic, but would want to confirm that this doesn’t exist in the final product. 
 - want to test a LARGE (>10GB) file with more things than transforms in it 
 - add test for render state not reading things into allFrames that it didn’t subscribe to

### Possible further improvements:
 - Don’t feel good about the allFrames assumption that it can only consist of FrameTransforms for the 3D panel — I’m wondering if this logic would be better transferred to the transform tree. We could have some kind of ingestion pipeline that queues messages before render and keeps track of the last messages of each frame. In this way we could avoid the creation of unnecessary or unused FrameTransforms at all because the Transform tree would only keep the state of the currentFrame. I think this has the potential to be much more performant possibly as well. 
 - would be great to enter a “loading” state for the panel after seeking if it takes longer than a certain amount of time


jhurliman:
We may want to move transforms out into a standalone library first so the separation of concerns is more clear.

defunctzombie:
@snosenzo Is this draft still needed or did you land these changes?

snosenzo:
all these changes were landed

PR #5102, opened 2023-01-06
Upgrade electron & related packages

jtbandes:
**User-Facing Changes**
None

**Description**
- Update electron to v22
  - disable sandbox which is now on by default, but will require changes in our webpack config and electron-socket to work


desktop/main/StudioWindow.ts
```diff
@@ -141,11 +142,9 @@ function newStudioWindow(deepLinks: string[] = []): BrowserWindow {
   });
 
   // Open all new windows in an external browser
-  // Note: this API is supposed to be superseded by webContents.setWindowOpenHandler,
-  // but using that causes the app to freeze when a new window is opened.
```
jtbandes, line 145: this bug no longer seems to happen, probably fixed in https://github.com/electron/electron/pull/29851

defunctzombie (approved):

PR #5109, opened 2023-01-07
Enable ESM-only packages in jest by transpiling node_modules

### LINKED ISSUE #1854: Configuration changes to allow ESM-only dependencies
Upgrading our `react-markdown` and `rehype-raw` dependencies brings them to new versions that are [ESM-only](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c). This is expected to happen with more other packages in the future. This currently causes test failures when imported in node/jest, so we will need to adopt configuration changes to address this without breaking our other packaging steps.

jtbandes:
**User-Facing Changes**
None

**Description**
Using Node's native ESM support with jest still has some problems (see https://github.com/facebook/jest/issues/13739). However, it is possible to use ESM-only packages by enabling transpilation (with babel) for `node_modules` by overriding the default `transformIgnorePatterns`.

Fixes #1854

### DIFF
--- a/packages/studio-base/jest.config.json
+++ b/packages/studio-base/jest.config.json
@@ -2,10 +2,13 @@
   "//": "Note: we use babel-jest rather than ts-jest for performance reasons.",
   "testMatch": ["<rootDir>/src/**/*.test.ts(x)?"],
   "transform": {
+    "\\.jsx?$": "babel-jest",
     "\\.tsx?$": "<rootDir>/src/test/transformers/typescriptTransformerWithRawImports.js",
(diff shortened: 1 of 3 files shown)


defunctzombie:
This random PR took 1m25s to run tests while this PR took 2m40s https://github.com/foxglove/studio/actions/runs/3839833170/jobs/6538130158 - I don't know if that's enough to care but wanted to point that out

jtbandes:
Test is still one of the fastest jobs so I am tempted to say it does not matter enough to tackle in this PR.

defunctzombie (approved):

PR #5113, opened 2023-01-07
Add i18n support and some Chinese translations

Woodii1998:
**User-Facing Changes**
Added a setting to select the app language. Some Chinese translations are now available.

**Description**
Add internationalization module, and add some Chinese translation


banisadr:
@Woodii1998 thanks for the patience, we've been having a few internal discussion to educate ourselves on the latest in i18n, make sure the direction of this PR aligns with our goals, and looking at the various framework options. The only think we'd like to request updated is following [i18next's documentation here](https://www.i18next.com/overview/typescript) to make the t function fully type safe and then we're happy to merge the PR.

Let us know if you've got any questions or concerns there.

Woodii1998:
> @Woodii1998 thanks for the patience, we've been having a few internal discussion to educate ourselves on the latest in i18n, make sure the direction of this PR aligns with our goals, and looking at the various framework options. The only think we'd like to request updated is following [i18next's documentation here](https://www.i18next.com/overview/typescript) to make the t function fully type safe and then we're happy to merge the PR.
> 
> Let us know if you've got any questions or concerns there.

Updated Please help review

packages/studio-base/src/locales/en/preferences.ts
```diff
@@ -0,0 +1,29 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+export default {
+  preferences: "Preferences",
+  colorScheme: "Color scheme",
+  dark: "Dark",
+  light: "Light",
+  followSystem: "Follow system",
+  en: "English",
+  zh: "中文",
```
geoffliu, line 12: Nit: In this case you actually want to hard code the strings in the target languages, bypassing the i18n process. It's bit of unnecessary effort?

banisadr:
@jtbandes could you take a look at this PR? I think it ought to be ready to merge based on our discussions.

packages/studio-base/src/App.tsx
```diff
@@ -44,6 +44,7 @@ import ConsoleApi from "./services/ConsoleApi";
 import { ExtensionLoader } from "./services/ExtensionLoader";
 import { ILayoutStorage } from "./services/ILayoutStorage";
 import URDFAssetLoader from "./services/URDFAssetLoader";
+import "./util/i18n";
```
jtbandes, line 47: Rather than depending on these side effects from import, maybe you can wrap them in an `initI18n()` function which we explicitly call?

packages/studio-base/src/locales/en/preferences.ts
jtbandes: agree, `t(option.text)` seems a bit weird above.

packages/studio-base/src/components/Preferences.tsx
```diff
@@ -419,7 +467,7 @@ export default function Preferences(): React.ReactElement {
                   onChange={(_event, checked) => void setTelemetryEnabled(checked)}
                 />
               }
-              label="Send anonymized usage data to help us improve Foxglove Studio"
+              label={t("sendAnonymizedUsageData")!}
```
jtbandes, line 470: Why do these ones need a `!`?

jtbandes (commented):
These changes in stories are undesirable: https://www.chromatic.com/test?appId=603ec8bf7908b500231841e2&id=63dbc9c06c8d515a8923669b

Maybe we need to also add the initialization stuff to `storybook/preview.tsx`?

It would be nice to also add at least one story variant that uses the chinese localization.

Woodii1998:
@jtbandes Ok thanks for your review I overlooked the change in the storybook section, I will fix the above problem later

packages/studio-base/src/locales/index.ts
```diff
@@ -0,0 +1,17 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { enPreferences, enGeneral } from "./en";
+import { zhPreferences, zhGeneral } from "./zh";
```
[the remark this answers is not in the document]
Woodii1998: @jtbandes This way of writing will work fine but seems to cause lint errors
https://github.com/foxglove/studio/actions/runs/4173389392/jobs/7225651429#step:7:64

Woodii1998:
@jtbandes All have been fixed, any comments on the above question about lint?

.storybook/preview.tsx
```diff
@@ -54,6 +55,10 @@ function StudioContextProviders({
   children,
   ctx,
 }: React.PropsWithChildren<{ ctx: StoryContext }>): JSX.Element {
+  useEffect(() => {
+    initI18n();
```
jtbandes: I'd put this under `loaders` below (next to `await waitForFonts`)

packages/studio-base/src/util/sharedStyleConstants.ts
```diff
@@ -12,8 +12,9 @@ export const fonts = {
...
   SANS_SERIF: "'Inter'",
-  SANS_SERIF_FEATURE_SETTINGS:
-    // enable font features https://rsms.me/inter/lab
-    // This style affects Chinese encoding and should be disabled when the current language is Chinese.
-    i18next.language === "zh" ? "" : "'cv08', 'cv10', 'tnum'",
+  // enable font features https://rsms.me/inter/lab
+  // This style affects Chinese encoding and should be disabled when the current language is Chinese.
+  SANS_SERIF_FEATURE_SETTINGS: (): string => {
+    return i18next.language === "zh" ? "" : "'cv08', 'cv10', 'tnum'";
+  },
```
jtbandes: It seems like changing the language wouldn't trigger components to re-render and get the new sans-serif feature settings. Any ideas for how to solve that?

@2metres would it work to move `font-feature-settings` to the `@font-face` blocks in inter.css, so that it would only apply when the Inter font is actually being used? (just guessing)

packages/studio-base/src/locales/index.ts
```diff
@@ -2,16 +2,7 @@
...
-export const translations = {
-  en: {
-    preferences: enPreferences,
-    general: enGeneral,
-  },
-  zh: {
-    preferences: zhPreferences,
-    general: zhGeneral,
-  },
-};
```
jtbandes: Sorry, I guess my suggestion caused errors with the "unused export" linter. You can revert back to this version.

packages/studio-base/src/components/Preferences.tsx
```diff
@@ -123,6 +129,7 @@ export function TimezoneSettings(): React.ReactElement {
 
   const { classes } = useStyles();
 
+  const { t } = useTranslation("preferences");
   const [timezone, setTimezone] = useAppConfigurationValue<string>(AppSetting.TIMEZONE);
   const detectItem: Option = useMemo(
```
jtbandes, line 134: In the screenshot I noticed `Detect from system` below is not translated. Maybe you can add a translation for that one so it's more consistent?

.storybook/preview.tsx
Woodii1998: @jtbandes Did you not update the code? The version you reviewed seems to be an intermediate version in my revision process, in the final version I have put this line in loaders

packages/studio-base/src/util/sharedStyleConstants.ts
Woodii1998: @jtbandes This also seems to be the modified version, and I removed the code here later, because this line of style affects all Chinese on the page (possibly also Japanese and Korean, but I haven't verified it), even the Chinese entered in the input field, so I don't think this issue has anything to do with the internationalization feature itself, and a separate PR should be proposed to fix this issue, and I will continue to I will continue to follow up on this issue

In fact, I've tried a lot for this problem, and placing it in @font-face is not feasible, because only firefox supports this writing method ([CSS at-rule: @font-face: font-feature-settings](https://caniuse.com/mdn-css_at-rules_font-face_font-feature-settings))

PS: If you want to reproduce the problem you can copy this paragraph to any input box: 显示时间戳在, you can see the difference between the content you copy and the content displayed on the page

Woodii1998: 显示时间戳在:
WeChate3ad9e2778a3588f54d33489850a1222

packages/studio-base/src/locales/index.ts
Woodii1998: it's ok I will restore this part of the code

packages/studio-base/src/components/Preferences.tsx
Woodii1998: Okay, I'll make that part consistent.

Woodii1998:
@jtbandes Updated Please help review

jtbandes (approved):

jtbandes:
@defunctzombie please merge if you are satisfied

jtbandes:
@Woodii1998 Thanks for working on this!

PR #5114, opened 2023-01-09
Revise sidebar icons for readability

esthersweon:
**User-Facing Changes**
- Revised "Panel settings" and "Extensions" sidebar icons for easier differentiation
Screen Shot 2023-01-09 at 12 16 52 PM

**Description**
- Based on user and team feedback about rectangular-ish icons being hard to differentiate


2metres:
We currently use the gear icon to indicate the panel settings in the panel toolbar, I think the pencil icon makes this more confusing not less.

2metres:
Longer term, I think the plan is to remove these Icons entirely and find better ways to access these features

PR #5116, opened 2023-01-09
Parameter panel: Add button to submit parameter changes

achim-k:
**User-Facing Changes**
- Parameter panel: Add button to submit parameter changes

**Description**
In #5095 I had several issues with the parameter panel:
- When the server rejects parameter updates, the parameter panel should reflect that. However, the last user-set value is shown
- There is no dedicated submit button and new parameter values are send while typing, often leading to invalid or incomplete parameter values being send to the server

This PR adds a dedicated submit button to send paramter updates only when the user is presses it. Additionally, the parameter row is re-rendered if the value changes.

image

### DIFF
--- a/packages/studio-base/src/panels/Parameters/index.tsx
+++ b/packages/studio-base/src/panels/Parameters/index.tsx
@@ -21,6 +21,7 @@ import {
   TableContainer,
   TableHead,
   TableRow,
+  Tooltip,
   Typography,
 } from "@mui/material";
 import { isEqual, isObject, union } from "lodash";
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/panels/Parameters/index.tsx
```diff
@@ -116,6 +120,36 @@ function displayableValue(value: unknown): string {
...
+        <Tooltip key="submit" title="Submit change">
+          <IconButton onClick={() => props.onSubmit(value)}>
+            <CheckIcon />
+          </IconButton>
+        </Tooltip>,
+        <Tooltip key="reset" title="Reset">
+          <IconButton key="reset" onClick={() => setValue(editableValue(props.value))}>
+            <ClearIcon />
+          </IconButton>
+        </Tooltip>,
```
achim-k, line 147: Ideally we can connect that to shortcuts (e.g. ctrl+enter / escape)

packages/studio-base/src/panels/Parameters/index.tsx
```diff
@@ -128,7 +162,7 @@ function Parameters(): ReactElement {
       (name: string, value: ParameterValue) => setParameterUnbounced(name, value),
       [setParameterUnbounced],
     ),
-    200,
+    500,
```
achim-k, line 165: This change is not relevant anymore. We can also remove the debounce (I think)

foxymiles (approved):
This seems good to me. The only downside I can see here is it slows down making a bunch of quick edits in succession.

achim-k:
> This seems good to me. The only downside I can see here is it slows down making a bunch of quick edits in succession.

I think this wouldn't be an issue anymore if we support shortcuts (e.g. <kbd>ctrl + enter</kbd> to submit, <kbd>esc</kbd> to cancel) or proper tab support (at the moment a tab is inserted in the jsoninput)

PR #5139, opened 2023-01-12
[Top nav] Use custom title bar buttons on windows & linux

jtbandes:
**User-Facing Changes**
None (feature flagged)

**Description**
Fixes FG-1293

Use custom window controls on Linux, and [window controls overlay](https://www.electronjs.org/docs/latest/tutorial/window-customization#window-controls-overlay-macos-windows) on Windows.

image

### DIFF
--- a/packages/studio-base/src/components/AppBar/index.tsx
+++ b/packages/studio-base/src/components/AppBar/index.tsx
@@ -2,6 +2,10 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
+import CheckBoxOutlineBlankIcon from "@mui/icons-material/CheckBoxOutlineBlank";
+import CloseIcon from "@mui/icons-material/Close";
+import FilterNoneIcon from "@mui/icons-material/FilterNone";
+import MinimizeIcon from "@mui/icons-material/Minimize";
 import { AppBar as MuiAppBar, Button, IconButton, Toolbar, Typography } from "@mui/material";
 import { MouseEvent, useCallback, useContext, useState } from "react";
 import { makeStyles } from "tss-react/mui";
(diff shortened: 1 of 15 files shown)


packages/studio-base/src/components/AppBar/index.tsx
```diff
@@ -164,66 +200,103 @@ export function AppBar(props: AppBarProps): JSX.Element {
...
+                      aria-haspopup="true"
+                      aria-expanded={userMenuOpen ? "true" : undefined}
+                      onClick={handleUserMenuClick}
+                      size="small"
+                    />
+                  ) : (
+                    <Button
+                      variant="contained"
+                      color="primary"
+                      onClick={() => {
```
defunctzombie, line 251: I've found myself thinking we should push more on keeping inline functions beyond a single line out of the react/jsx part of the render code. I think it makes it easier to read and maintain the jsx markup

packages/studio-base/src/components/AppBar/index.tsx
```diff
@@ -164,66 +200,103 @@ export function AppBar(props: AppBarProps): JSX.Element {
...
+                            cta: "sign-in",
+                          });
+                        }
+                      }}
+                    >
+                      Sign in
+                    </Button>
+                  ))}
+                {showCustomWindowControls && (
+                  <>
```
defunctzombie, line 265: Possibly worth turning these controls into their own component and using that within the app bar?

amacneil:
How do other electron apps on linux+windows handle the icon problem? Is there no way to get more native looking icons?

Haven't tested myself, but the screenshot looks nice. Low risk since it is behind a feature flag.

jtbandes:
We could try this controls overlay thing instead: https://www.electronjs.org/docs/latest/tutorial/window-customization#window-controls-overlay-macos-windows
But that’s not supported on Linux. Anyway aren’t there multiple Linux window managers that would have different-looking buttons anyway? Do we want to try and detect the window manager and use custom icons that match more closely?

defunctzombie (approved):

PR #5141, opened 2023-01-13
Color plot paths by series color in tooltip

foxymiles:
**User-Facing Changes**
Use series colors to color paths in the plot panel tooltips.

**Description**
This makes a few changes:
1. Includes the dataset index in the value returned in the rpc hover message.
2. Lifts the construction of the point -> data tooltip lookup one level and includes the dataset index in the lookup key.
3. Passes a datsetindex -> color lookup to the tooltip content to color the path.

This is an alternate approach to https://github.com/foxglove/studio/pull/4955 that requires some extra effort to avoid another object allocation in the tooltip array which could  be expensive since we currently construct a tooltip for every point in the potentially large dataset. 

This also allows us to always display the values in the tooltip in the same order as the series definitions, which is not currently the case on main. By lifting the construction of the lookup map up one level we also skip the construction of the intermediate concatenated array of all tooltips.

Screenshot 2023-01-13 at 9 58 47 AM

### DIFF
--- a/packages/studio-base/src/components/TimeBasedChart/TimeBasedChartTooltipContent.tsx
+++ b/packages/studio-base/src/components/TimeBasedChart/TimeBasedChartTooltipContent.tsx
@@ -11,7 +11,7 @@
 //   found at http://www.apache.org/licenses/LICENSE-2.0
 //   You may not use this file except in compliance with the License.
 
-import { take } from "lodash";
+import { sortBy, take } from "lodash";
 import { PropsWithChildren, useMemo } from "react";
 import { makeStyles } from "tss-react/mui";
 
@@ -20,6 +20,7 @@ import { fonts } from "@foxglove/studio-base/util/sharedStyleConstants";
 import { TimeBasedChartTooltipData } from "./index";
 
 type Props = {
+  colorsByDatasetIndex?: Record<string, undefined | string>;
   content: TimeBasedChartTooltipData[];
   // Flag indicating the containing chart has multiple datasets
   multiDataset: boolean;
@@ -39,7 +40,7 @@ const useStyles = makeStyles()((theme) => ({
     fontStyle: "italic",
   },
   path: {
-    opacity: 0.6,
+    opacity: 0.9,
     whiteSpace: "nowrap",
   },
 }));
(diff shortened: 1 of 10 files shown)


jtbandes (approved):

packages/studio-base/src/components/TimeBasedChart/TimeBasedChartTooltipContent.tsx
```diff
@@ -102,12 +103,24 @@ export default function TimeBasedChartTooltipContent(
     );
   }
 
+  const sortedItems = sortBy(
```
defunctzombie, line 106: A comment on the intended sort order/why would help us maintain this across refactors.

packages/studio-base/src/components/TimeBasedChart/TimeBasedChartTooltipContent.tsx
```diff
@@ -102,12 +103,24 @@ export default function TimeBasedChartTooltipContent(
...
+
   return (
     <div className={classes.root} data-testid="TimeBasedChartTooltipContent">
-      {Array.from(itemsByPath.out.entries(), ([path, items], idx) => {
+      {sortedItems.map(([path, items], idx) => {
+        const firstItem = items[0];
+        const color =
+          firstItem?.datasetIndex != undefined
+            ? colorsByDatasetIndex?.[firstItem.datasetIndex]
+            : "auto";
```
defunctzombie, line 118: If you want a one-liner. Uses the trick that `-1` results in an undefined lookup for js arrays.

```
// Find the color of the dataset of this item or fallback to 'auto'
const color = colorsByDatasetIndex?.[items[0]?.datasetIndex ?? -1] ?? 'auto'
```

packages/studio-base/src/panels/Plot/index.tsx
```diff
@@ -124,6 +124,16 @@ const ZERO_TIME = { sec: 0, nsec: 0 };
 
 const performance = window.performance;
 
+function buildTooltipLookupMap(datasets: DataSet[]): Map<string, TimeBasedChartTooltipData> {
```
defunctzombie, line 127: comment on the key -> value relationship (i.e. what are the keys/why)

packages/studio-base/src/panels/StateTransitions/index.tsx
```diff
@@ -344,6 +344,14 @@ const StateTransitions = React.memo(function StateTransitions(props: Props) {
 
   useStateTransitionsPanelSettings(config, saveConfig);
 
+  const tooltipLookup = useMemo(() => {
```
defunctzombie, line 347: I tend to like putting the key->value relationship in the name of map variables. i.e. `tooltipKeyToTooltip`. Up to you.

defunctzombie (approved):

PR #5142, opened 2023-01-13
Improves empty layout state

foxymiles:
**User-Facing Changes**
Improves empty layout state.

**Description**
This fixes remaining issues with https://github.com/foxglove/studio/pull/4578
Closes https://github.com/foxglove/studio/pull/4578

### DIFF
--- a/packages/studio-base/src/Workspace.stories.tsx
+++ b/packages/studio-base/src/Workspace.stories.tsx
@@ -2,18 +2,23 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
+import { Story } from "@storybook/react";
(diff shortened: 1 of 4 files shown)


jtbandes (approved):

defunctzombie:
"Create a new layout" creates a new default layout for me. Is the intent to have that happen rather than drop the user into the panel selection grid I see when I click "+" in the layout panel?

defunctzombie (approved):

foxymiles:
@defunctzombie I'm not sure. It would of course be easy to create an empty layout instead if you think that's better UX.

defunctzombie:
@foxymiles probably worth doing a demo at upcoming product review and seeing what the beehive thinks and then merging based on that feedback.

amacneil:
When I hit "create a new layout" let's create an empty one rather than default layout.

PR #5143, opened 2023-01-13
[Top nav] move extensions settings into a tab in modal

jtbandes:
**User-Facing Changes**
None (feature flagged)

**Description**

Fixes FG-1214

image
image


defunctzombie (approved):
The list UI and individual detail view styling feels _off_ when put into a dialog environment but I think that is easy to iterate on separately from this migration since this re-uses the same underlying components.

PR #5150, opened 2023-01-14
Improve description for new navigation flag

amacneil:
**User-Facing Changes**
None

**Description**
Improve name & description in settings for new top nav feature flag

### DIFF
--- a/packages/studio-base/src/components/ExperimentalFeatureSettings.tsx
+++ b/packages/studio-base/src/components/ExperimentalFeatureSettings.tsx
@@ -62,12 +62,12 @@ const features: Feature[] = [
     description: <>Allow editing plot panel data series in the sidebar.</>,
   },
   {
-    key: AppSetting.ENABLE_NEW_UI,
-    name: "Enable new user interface",
+    key: AppSetting.ENABLE_NEW_TOPNAV,
+    name: "New navigation UI",
     description: (
       <>
-        Try our redesigned navigation and layout.
-        {isDesktopApp() && " Close and reopen the app for changes to take effect."}
+        Enable experimental new user interface.
+        {isDesktopApp() && " Restart the app for changes to take effect."}
       </>
     ),
   },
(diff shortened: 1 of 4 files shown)


packages/studio-base/src/components/ExperimentalFeatureSettings.tsx
```diff
@@ -62,12 +62,12 @@ const features: Feature[] = [
...
-    key: AppSetting.ENABLE_NEW_UI,
-    name: "Enable new user interface",
+    key: AppSetting.ENABLE_NEW_TOPNAV,
+    name: "New navigation UI",
     description: (
       <>
-        Try our redesigned navigation and layout.
-        {isDesktopApp() && " Close and reopen the app for changes to take effect."}
+        Enable experimental new user interface.
+        {isDesktopApp() && " Restart the app for changes to take effect."}
```
jtbandes, line 70: This doesn't feel like an improvement 🤷
"Experimental" is redundant because this is in a section called experimental features. And I'm surprised you went with "new user interface" given that you don't seem to like `NEW_UI` as part of the flag name in code...

defunctzombie:
I'm gonna remove myself from reviewers. I am confident you two can come to a mutual agreement on words.

packages/studio-base/src/components/ExperimentalFeatureSettings.tsx
amacneil: Yeah, it seemed at least slightly better in the description rather than the title?

Dunno, it just seems really silly having a feature flag called "Enable new user interface" - most of our experimental features are new user interfaces. I don't care about the const we usually internally, I updated it to match but could also leave that out.

jtbandes: If it's silly to have a flag called "Enable new user interface", why are you proposing the description should be "Enable experimental new user interface"? That's why I'm confused 😅

jtbandes (approved):
Gonna approve this because realistically I don't care what we call the flag.

packages/studio-base/src/components/ExperimentalFeatureSettings.tsx
amacneil: fixed :D

jtbandes:
FWIW we still use `enableNewUI` as a variable name in a few places: https://github.com/foxglove/studio/blob/7fc2727e1300aa3b86e3c4973062515fff79f466/desktop/main/StudioWindow.ts#L104

PR #5156, opened 2023-01-17
Bypass debounce for interactive settings edits in 3d

foxymiles:
**User-Facing Changes**
Improves interactivity of edits to 3d panel settings fields.

**Description**
Makes the following changes:
1. Adds a new `interaction` property to settings tree action update payloads for number, vec2, and vec3 inputs that can be either `direct` for direct user interactions with the input or `scrubbing` for rapid scrubbing interactions.
2. Lifts the debounced settings update out of individual scene extensions to the 3d panel itself.
3. For non-scrubbing interactions, the 3d panel action handler immediately flushes settings tree updates to the sidebar to prevent writing debounced updates to a react controlled input, which results in loss of caret position.

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/SceneExtension.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/SceneExtension.ts
@@ -2,7 +2,7 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
-import { debounce, set, unset } from "lodash";
+import { set, unset } from "lodash";
 import * as THREE from "three";
 import { DeepPartial } from "ts-essentials";
 
@@ -18,8 +18,6 @@ import { updatePose } from "./updatePose";
 export type PartialMessage<T> = DeepPartial<T>;
 
 export type PartialMessageEvent<T> = MessageEvent<DeepPartial<T>>;
-
-const SETTINGS_DEBOUNCE_MS = 100;
 
 /**
  * SceneExtension is a base class for extending the 3D scene. It extends THREE.Object3D and is a
@@ -54,12 +52,6 @@ export class SceneExtension<
    * renderable per topic.
    */
   public readonly renderables = new Map<string, TRenderable>();
-
-  private _settingsUpdateDebounced = debounce(
-    () => this.renderer.settings.setNodesForKey(this.extensionId, this.settingsNodes()),
-    SETTINGS_DEBOUNCE_MS,
-    { leading: true, trailing: true, maxWait: SETTINGS_DEBOUNCE_MS },
-  );
 
   /**
    * @param extensionId A unique identifier for this SceneExtension, such as `foxglove.Markers`.
@@ -123,7 +115,7 @@ export class SceneExtension<
    * `settingsNodes()` method will be called to retrieve the latest nodes.
    */
   public updateSettingsTree(): void {
-    this._settingsUpdateDebounced();
+    this.renderer.settings.setNodesForKey(this.extensionId, this.settingsNodes());
   }
 
   /**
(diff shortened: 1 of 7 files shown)


packages/studio-base/src/panels/ThreeDeeRender/SceneExtension.ts
```diff
@@ -122,15 +122,22 @@ export class SceneExtension<
...
 
   /**
    * Persists a value to the panel configuration at the given path. The base class implementation
    * calls `renderer.updateConfig()` and `updateSettingsTree()`.
    */
-  public saveSetting(path: Path, value: unknown): void {
+  public saveSetting(
+    path: Path,
+    value: unknown,
+    write: "immediate" | "debounced" = "immediate",
```
jtbandes, line 139: why change the default for all settings saves? Maybe the default should be debounced and we use immediate in a few places we need it? Or maybe there should be no default so the caller is forced to choose one?

foxymiles: From what I can tell `saveSettings` is mainly called by the action handlers, so these will be coming from user-initiated edits, all of which we'd want to flush immediately to avoid the caret positioning problem.

@jhurliman does that sound right?

jhurliman: My understanding is user-initiated edits can happen at vsync rate if you are dragging a slider back and forth. I think those need to be debounced.

jhurliman:
> If the interactive scrubbing is the only issue maybe we should just debounce the updates from that before they hit the action handlers?

That could work

defunctzombie:
My first glance at this is that it is not a great devx for the extension author. Why do I need to select “direct” or “scrubbing” - both are interactions with the input. The description on direct says “direct user interactions with the input” - how is that different than when I scrub? That is a direct interaction.

I am 👎 based on the current PR description. I would push for us to handle this better or find a simpler way for the extension author to ensure proper behavior. The proposal here _feels_ like it introduces more confusion.

defunctzombie:
@foxymiles What is the difference in behavior I would observe as a panel author for `direct` vs `scrubbing`?

foxymiles:
@defunctzombie The only difference is the speed of updates. `scrubbing` updates can arrive at `requestAnimationFrame` rates, so 60+hz. For most panels this is immaterial but for panels with very complex settings like 3d you may want to debounce only those high frequency messages.

Maybe there's better terminology than `direct` vs `scrubbing`? I also considered just having an `isScrubbing` boolean.

defunctzombie:
> Maybe there's better terminology than `direct` vs `scrubbing`? I also considered just having an `isScrubbing` boolean.

I think my hangup is that it feels like maybe the move here is to expose the extension author to the debounce concept here any maybe let them configure some delay? Then as an author I can be clear that what I am doing is selecting the delay for updates on this field. I personally did not find `direct` or `scrubbing` very obvious terms that I could translate into how it would impact my settings updates.

Maybe see if others share my feelings or are ok with direct vs scrubbing (or have other terms that are even more clear). I don't mind the concept of allowing an extension author to control a delay here but I think we should consider the terms we want to use.

foxymiles:
I considered allowing panel authors to configure this on a per-field basis but the way that 3d implements settings would require quite a bit of extra work to set this on all the various renderables that can contribute fields to the rendered tree. It also seems to me that it's something we'd want to debounce for any numeric input in the 3d panel.

Maybe @jhurliman has some better ideas?

jhurliman (approved):

defunctzombie (changes requested):
I want to take a look at this because I am not a fan of the proposed terms.

foxymiles:
Closed in favor of https://github.com/foxglove/studio/pull/5419

PR #5157, opened 2023-01-17
Remove Add Series button from plot legend

foxymiles:
**User-Facing Changes**
Removes the add series button from the plot legend when the `Plot panel series in settings` experimental feature is enabled.

**Description**
We believe that this is redundant since users can add a new series via the context menu in the settings sidebar.

### DIFF
--- a/packages/studio-base/src/panels/Plot/NewPlotLegend.tsx
+++ b/packages/studio-base/src/panels/Plot/NewPlotLegend.tsx
@@ -2,13 +2,12 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
-import AddIcon from "@mui/icons-material/Add";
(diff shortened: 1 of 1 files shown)


jtbandes (approved):

banisadr:
Agree with 🔪 ing the redundancy, but adding a new series is now buried two additional clicks deeper under the kebab menu of the series settings. 
image

To retain the intuitive-ness of the old UI, but also gain the consistency of the settings, could put the `+ Add line` button in the series settings context (and rename it `+ Add series`)?
image

defunctzombie:
Since this is under a feature flag I think we should merge this and figure out how we want to surface "add series". Right now its awkward to me having it take up space in the panel and when you click it you have to go configure it elsewhere. Maybe others feel differently but my UI hope is to turn the legend into an actual legend with no more frills and keep the editing contained to settings.

defunctzombie (approved):

PR #5159, opened 2023-01-17
Run foxglove websocket connection in web worker

achim-k:
**User-Facing Changes**
None

**Description**
If supported, a dedicated web worker will be created for managing the foxglove websocket connection, to reduce the load on the main thread. 

Summary:
- Make sure that we call `close()` on the `FoxgloveClient`, otherwise the worker will not be destroyed (memory leak)
- Create `WorkerSocketAdaptor` which creates a regular websocket connection in a webworker. It implements the [IWebSocket interface](https://github.com/foxglove/ws-protocol/blob/f6c95b059b7ef87cbf1058252f5cb32a61ccf22c/typescript/ws-protocol/src/types.ts#L143-L156) and can therefore be passed to `FoxgloveClient` instead of a regular websocket

FG-1250

### DIFF
--- a/packages/studio-base/src/players/FoxgloveWebSocketPlayer/WorkerSocketAdaptor.ts
+++ b/packages/studio-base/src/players/FoxgloveWebSocketPlayer/WorkerSocketAdaptor.ts
@@ -1,0 +1,70 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { IWebSocket } from "@foxglove/ws-protocol";
+
+import { FromWorkerMessage, ToWorkerMessage } from "./worker";
+
+export default class WorkerSocketAdaptor implements IWebSocket {
+  private worker: Worker;
+  public binaryType: string = "";
+  public protocol: string = "";
+  public onerror: ((event: unknown) => void) | undefined = undefined;
+  public onopen: ((event: unknown) => void) | undefined = undefined;
+  public onclose: ((event: unknown) => void) | undefined = undefined;
+  public onmessage: ((event: unknown) => void) | undefined = undefined;
+
+  public constructor(wsUrl: string, protocols?: string[] | string) {
+    this.worker = new Worker(new URL("./worker", import.meta.url));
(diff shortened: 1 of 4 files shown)


achim-k:
[FG-1250](https://linear.app/foxglove/issue/FG-1250)

packages/studio-base/src/players/FoxgloveWebSocketPlayer/WorkerSocketAdaptor.ts
```diff
@@ -0,0 +1,70 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { IWebSocket } from "@foxglove/ws-protocol";
+
+import { FromWorkerMessage, ToWorkerMessage } from "./worker";
+
+export default class WorkerSocketAdaptor implements IWebSocket {
```
jtbandes, line 9: I think it's not wrong, but we use the `Adapter` spelling in several other places in the studio code, so let's change this one to match

packages/studio-base/src/players/FoxgloveWebSocketPlayer/WorkerSocketAdaptor.ts
```diff
@@ -0,0 +1,70 @@
...
+      }
+    };
+  }
+
+  public close(): void {
+    this.sendToWorker({
+      type: "close",
+      data: undefined,
+    });
+    this.worker.terminate();
```
jtbandes, line 60: I think terminate() might shut down the worker before the "close" message is received. Not sure if that is really a problem, but wanted to mention it since I think the close code might not have any effect. Needs testing.

jtbandes (approved):
Seems pretty simple.

packages/studio-base/src/players/FoxgloveWebSocketPlayer/WorkerSocketAdaptor.ts
achim-k: I think I will leave it as it is for now. We only ever call `close()` if the underlying websocket `close` event is emitted, so in otherwords the websocket is already closed when this function is called

PR #5168, opened 2023-01-18
Record app name with events

amacneil:
**User-Facing Changes**
None

**Description**
Record app name (`studio`) in all analytics events for easier differentiation.


jtbandes:
Aren't all the events already under a studio project?

amacneil:
No, our 1 year startup trial expired and we got bumped back to the amplitude free plan, which doesn't support multiple projects.

jtbandes (approved):

Issue #5115, opened 2023-01-09
Messages on output topics from user scripts are repeated for every block update (for preload='full')


snosenzo:
**Description**
There’s a bug where multiple user_script message are being repeatedly added to `blocks` they’ve already been added as more blocks come in (see `_getBlocks` in `UserNodePlayer`).  With a plot reading the output topic, I made a user script that takes in camera info and outputs a message and after buffering is done it has 12 of the same message in the block under the output topic when there’s only one camera_info input. 

- Version: 1.36.0
- Platform: web

**Steps To Reproduce**

Data: nuscenes demo
Layout: user script, plot with line for output topic (.now) from user script
UserScript code:
```
import { Input, Message } from "./types";
type Output = {
  now: string;
};
export const inputs = ["/CAM_FRONT_LEFT/camera_info"];
export const output = "/studio_script/output_topic";
export default function script(
  event: Input<"/CAM_FRONT_LEFT/camera_info">
): Output {
  return {
    now: "" + Date.now(),
  };
}
```
See this on the plot: 
<img width="565" alt="image" src="https://user-images.githubusercontent.com/10187776/211413741-122d9add-0677-487d-80df-243dd6b4c661.png">

Showing multiple output messages for single time values.

**Expected Behavior**
I would expect that these would not aggregate and there would be a 1:1 input->output message present in blocks and in the plot.

PR #5172, opened 2023-01-18
Fix creation of duplicate user script messages for preloaded output topics

snosenzo:
**User-Facing Changes**

 - fix bug where plots would contain old messages from user scripts

**Description**
 - fix duplicate output message in blocks

#5115
FG-1302


jtbandes:
Can we add a unit test for this?

jtbandes (approved):

PR #5173, opened 2023-01-18
Update userscript messages when global variables change

snosenzo:
**User-Facing Changes**

- userscript messages recomputed on global variables change

**Description**
Utilized `lastMessagesByInputTopic` to recalculate user script messages for the current frame. 
Will capture functionality with a test.

FG-1071

### DIFF
--- a/packages/studio-base/src/players/UserNodePlayer/index.ts
+++ b/packages/studio-base/src/players/UserNodePlayer/index.ts
@@ -134,6 +134,7 @@ export default class UserNodePlayer implements Player {
   // keep track of last message on all topics to recompute output topic messages when user nodes change
   private _lastMessageByInputTopic = new Map<string, MessageEvent<unknown>>();
   private _userNodeIdsNeedUpdate = new Set<string>();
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/players/UserNodePlayer/index.ts
```diff
@@ -350,6 +351,7 @@ export default class UserNodePlayer implements Player {
 
   public setGlobalVariables(globalVariables: GlobalVariables): void {
     this._globalVariables = globalVariables;
+    this._globalVariablesChanged = true;
```
jtbandes, line 354: It seems like this depends on a new player state coming in, but if the player is not emitting new state then we won't update the variables, is that right? Seems like we would want to update them immediately if possible.

snosenzo: Ah yeah I was operating under the assumption that the underlying player was the `IterablePlayer` that emits every 1 second in its idle state. Will update so it works with other players

jtbandes (approved):

PR #5177, opened 2023-01-19
NewUI -> NewTopNav

amacneil:
**User-Facing Changes**
None

**Description**
Cleanup from https://github.com/foxglove/studio/pull/5150#issuecomment-1397502308

### DIFF
--- a/packages/studio-base/src/AppSetting.ts
+++ b/packages/studio-base/src/AppSetting.ts
@@ -21,7 +21,7 @@ export enum AppSetting {
   SHOW_DEBUG_PANELS = "showDebugPanels",
   ENABLE_LEGACY_PLOT_PANEL = "enableLegacyPlotPanel",
   ENABLE_PLOT_PANEL_SERIES_SETTINGS = "enablePlotPathSeriesSettings",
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/AppSetting.ts
```diff
@@ -21,7 +21,7 @@ export enum AppSetting {
   SHOW_DEBUG_PANELS = "showDebugPanels",
   ENABLE_LEGACY_PLOT_PANEL = "enableLegacyPlotPanel",
   ENABLE_PLOT_PANEL_SERIES_SETTINGS = "enablePlotPathSeriesSettings",
-  ENABLE_NEW_TOPNAV = "enableNewUI",
```
jtbandes, line 24: Changing this is unfortunate because it means the old setting won't be remembered. Not sure if we care about that – it's still pretty experimental so maybe not a big deal.

jtbandes (approved):

jtbandes (approved):
test

packages/studio-base/src/AppSetting.ts
amacneil: oh I see. oh well, should only be FG employees using it.

PR #5180, opened 2023-01-20
Convert Image panel to extension API

jtbandes:
**User-Facing Changes**
Changed: Image panel topic selection is now exclusively located in the panel settings.

**Description**
Converts image panel to use the extension API.
- Reimplement useImagePanelMessages with zustand for state management

Resolves FG-1648


amacneil:
👍

packages/studio-base/src/panels/Image/ImageView.tsx
```diff
@@ -0,0 +1,384 @@
...
+      subs.push(cameraInfoTopic);
+    }
+    return subs;
+  }, [cameraTopic, cameraInfoTopic, enabledMarkerTopics]);
+
+  const [colorScheme, setColorScheme] = useState<"dark" | "light">("light");
+
+  useEffect(() => {
+    context.saveState(config);
+    context.setDefaultPanelTitle(config.cameraTopic === "" ? undefined : config.cameraTopic);
```
amacneil, line 108: What happens if you set the default panel title to empty string? Presumably we would still fall back to `Image`?

jhurliman (approved):

packages/studio-base/src/panels/Image/ImageView.tsx
```diff
@@ -0,0 +1,384 @@
...
+  padding: theme.spacing(0.25, 0.5),
+  userSelect: "all",
+
+  ".mosaic-window:hover &": {
+    opacity: "1",
+  },
+  ...(screenshotTest && {
+    opacity: 1,
+  }),
+}));
```
2metres, line 67: Contemplate not using `muiStyled` here for better CSS karma

packages/studio-base/src/panels/Image/components/ImageCanvas.tsx
```diff
@@ -42,7 +42,7 @@ type Props = {
   image?: NormalizedImageMessage;
   rawMarkerData: RawMarkerData;
   config: Config;
-  saveConfig: SaveImagePanelConfig;
+  saveConfig: (config: Partial<Config>) => void;
```
foxymiles, line 45: Can you use our `SaveConfig` typedef here? https://github.com/foxglove/studio/blob/c680b4b9669633c755160d13bfd8fa3019070dd4/packages/studio-base/src/types/panels.ts#L39-L41

foxymiles (approved):

packages/studio-base/src/panels/Image/ImageView.tsx
jhurliman: @2metres what should be used instead?

defunctzombie: tss-react https://github.com/foxglove/studio/blob/2ba5d3ceff1e88cf86f29750612a7c1b3a3434e4/packages/studio-base/src/panels/Log/FilterTagInput.tsx#L7

defunctzombie: You want to avoid muiStyled because it has bad re-render behavior.

packages/studio-base/src/panels/Image/components/ImageCanvas.tsx
jhurliman: Doesn't look like it:

> Type 'Partial<Config> | ((oldConfig: Config) => Partial<Config>)' is not assignable to type 'Partial<Config>'.

amacneil:
👏

PR #5185, opened 2023-01-20
Run dependabot daily, bump limit to 10

jtbandes:
**User-Facing Changes**
None

**Description**
Seems like dependabot has proven itself to be working.


defunctzombie (approved):
📿

PR #5187, opened 2023-01-20
 Fix freeze after receiving tf message with identical parent and child ids

snosenzo:
**User-Facing Changes**

 - fix freeze after receiving tf message with identical parent and child ids

**Description**
 - cycle detection wasn't checking for frameId being the same as parentId

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/transforms/TransformTree.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/transforms/TransformTree.ts
@@ -160,6 +160,9 @@ export class TransformTree {
     return output;
   }
   private _checkParentForCycle(frameId: string, parentFrameId: string): boolean {
+    if (frameId === parentFrameId) {
+      return true;
+    }
     // walk up tree from parent Frame to check if it eventually crosses the frame
     let frame = this.frame(parentFrameId);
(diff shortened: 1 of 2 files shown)


jhurliman (approved):

packages/studio-base/src/panels/ThreeDeeRender/transforms/TransformTree.ts
```diff
@@ -160,6 +160,9 @@ export class TransformTree {
     return output;
   }
   private _checkParentForCycle(frameId: string, parentFrameId: string): boolean {
+    if (frameId === parentFrameId) {
+      return true;
+    }
```
jtbandes, line 165: Could there be a way to fold this into the logic below, rather than adding a new `if`? Such as changing the logic inside the while to be `if (frame.id === frameId) { return true }`?

snosenzo: I don't think so because the condition for the loop is whether the frame has a parent, and in this case the frame doesn't have a parent defined yet because `parent` hasn't been set. There might be a more clever way, but I'm happy with the readability of this as is

jtbandes (approved):

PR #5194, opened 2023-01-21
add non-dismissible deprecation banner to URDF Viewer panel

jtbandes:
**User-Facing Changes**
The URDF Viewer panel is now deprecated and will be removed in a future release. [Use the 3D panel](https://foxglove.dev/docs/studio/panels/3d#add-unified-robot-description-format-urdf) to visualize URDFs instead.

**Description**
image
Resolves FG-1118

### DIFF
--- a/packages/studio-base/src/panels/URDFViewer/index.tsx
+++ b/packages/studio-base/src/panels/URDFViewer/index.tsx
@@ -186,7 +186,7 @@ function URDFViewer({ config, saveConfig }: Props) {
     <div className={classes.root}>
       <PanelToolbar />
       <Alert severity="warning">
-        The URDF Viewer panel is now deprecated. See the{" "}
+        The URDF Viewer panel is deprecated. See the{" "}
         <Link
           href="https://foxglove.dev/docs/studio/panels/3d#add-unified-robot-description-format-urdf"
           target="_blank"
--- a/packages/studio-base/src/panels/URDFViewer/index.tsx
+++ b/packages/studio-base/src/panels/URDFViewer/index.tsx
@@ -2,7 +2,7 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
-import { Button } from "@mui/material";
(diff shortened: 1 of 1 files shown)


amacneil:
Rather than linking to docs, how about a link that switches the panel out for 3d? Like we did with old3d.

defunctzombie:
> Rather than linking to docs, how about a link that switches the panel out for 3d? Like we did with old3d.

The transition here is going to be more involved for users that the 3d -> 3d one. If we switch the panel and no docs link to the relevant section that's a rougher transition path IMO.

amacneil (approved):
lgtm

Are there any features in the URDF panel not supported in the 3D panel?

defunctzombie:
> Are there any features in the URDF panel not supported in the 3D panel?

Yes

amacneil:
What are they? Do we have a migration plan?

banisadr:
@jtbandes and I did a deep dive on this last week and made some decisions:

* Primary feature not supported & with no deprecation plan is support for JointState messages. @jhurliman has mentioned several times, and myself/Jacob/Roman are in alignment that it does not provide value and is not required. Omitting it is a two-way door, so if we turn out to be wrong, we can plan the work to add it to the 3D panel.
* [This project](https://linear.app/foxglove/project/remove-urdf-viewer-panel-5537a017e10f) lists the required bug & doc updates before we can remove the URDF panel. Note we cancelled the 'Drag and drop' support issue. 
* [This project](https://linear.app/foxglove/project/improve-urdf-usability-6983f92dae27) lists all of the items we noted should be addressed to make using URDFs in Studio more user friends. This is applies equally to the current panel and using URDFs with the 3D panel, but does not block deprecation & removal of the existing panel (mentioning for completeness).

banisadr:
Happy to explain the reasoning behind ✂️ Drag & drop if its a concern. 

I'm not certain why [this ticket](https://linear.app/foxglove/issue/FG-716/deprecate-the-urdf-panel) was moved to canceled though. @jtbandes?

jtbandes:
I think because it got converted to a project.

PR #5196, opened 2023-01-22
Add a yarn cache clean step to Dockerfile

asymingt:
Simply adds one line to the Foxglove studio `Dockerfile` to clean the yarn cache prior to installing, which gets around an error I had encountered when trying to build and run foxglove with `docker compose up`,

Signed-off-by: Andrew Symington <andrew.c.symington@gmail.com>

**User-Facing Changes**

None

**Description**

This is the error I had encountered:

```
asymingt@mars:~/development/grids$ docker compose run foxglove
[+] Building 4.2s (12/14)                                                                                                                                                           
 => [internal] load build definition from Dockerfile                                                                                                                           0.0s
 => => transferring dockerfile: 32B                                                                                                                                            0.0s
 => [internal] load .dockerignore                                                                                                                                              0.0s
 => => transferring context: 34B                                                                                                                                               0.0s
 => [internal] load metadata for docker.io/library/caddy:2.5.2-alpine                                                                                                          0.4s
 => [internal] load metadata for docker.io/library/node:16                                                                                                                     0.4s
 => [internal] load build context                                                                                                                                              0.1s
 => => transferring context: 3.07MB                                                                                                                                            0.1s
 => [build 1/6] FROM docker.io/library/node:16@sha256:27fab5920246070cf13449cf44c25bc4f5adef18ca7482b2bda90b7cf9e64481                                                         0.0s
 => [stage-1 1/3] FROM docker.io/library/caddy:2.5.2-alpine@sha256:b31ff95e98737b849d6af1fb9d9cb54a66ba3684564b3310541f60b12b1dd619                                            0.0s
 => CACHED [stage-1 2/3] WORKDIR /src                                                                                                                                          0.0s
 => CACHED [build 2/6] WORKDIR /src                                                                                                                                            0.0s
 => [build 3/6] COPY . ./                                                                                                                                                      1.1s
 => [build 4/6] RUN corepack enable                                                                                                                                            0.4s
 => ERROR [build 5/6] RUN yarn install --immutable                                                                                                                             2.1s
------
 > [build 5/6] RUN yarn install --immutable:
#0 2.047 Internal Error: Unable to deserialize cloned data due to invalid or unsupported version.
#0 2.047     at Object.deserialize (node:v8:380:7)
#0 2.047     at Ke.restoreInstallState (/root/.cache/node/corepack/yarn/3.1.0/yarn.js:447:965)
#0 2.047     at processTicksAndRejections (node:internal/process/task_queues:96:5)
#0 2.047     at async wC.execute (/root/.cache/node/corepack/yarn/3.1.0/yarn.js:486:14746)
#0 2.047     at async wC.validateAndExecute (/root/.cache/node/corepack/yarn/3.1.0/yarn.js:337:620)
#0 2.047     at async oo.run (/root/.cache/node/corepack/yarn/3.1.0/yarn.js:351:1846)
#0 2.047     at async oo.runExit (/root/.cache/node/corepack/yarn/3.1.0/yarn.js:351:2013)
#0 2.047     at async i (/root/.cache/node/corepack/yarn/3.1.0/yarn.js:448:12377)
#0 2.047     at async r (/root/.cache/node/corepack/yarn/3.1.0/yarn.js:448:10617)
------
failed to solve: executor failed running [/bin/sh -c yarn install --immutable]: exit code: 1
```
The addition of the new line fixes it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,6 +4,7 @@ WORKDIR /src
 COPY . ./
 
 RUN corepack enable
+RUN yarn cache clean
 RUN yarn install --immutable
 
 ENV FOXGLOVE_DISABLE_SIGN_IN=true


defunctzombie:
Can you explain why this is needed? The docker build starts with no yarn packages installed so there's nothing to clean prior to install.

asymingt:
> Can you explain why this is needed? The docker build starts with no yarn packages installed so there's nothing to clean prior to install.

I'm not super familiar with how yarn caching works, but a clean checkout on my machine to folder `studio` and the simple `docker-compose.yml` file below used with `docker compose up` yields an error. 

```yml
version: '3'
services:
  foxglove:
    container_name: foxglove
    network_mode: host
    build: ./studio
```

I suspect that the `COPY . ./` command in the studio Dockerfile includes the `.yarn` folder, which has an old cache in it (I might be wrong about this) and the `yarn cache clean` remove this before the build.

defunctzombie:
> I suspect that the `COPY . ./` command in the studio Dockerfile includes the `.yarn` [folder](https://github.com/foxglove/studio/tree/main/.yarn), which has an old cache in it (I might be wrong about this) and the `yarn cache clean` fixes this before the build.

Your comment says that a clean checkout of Studio yields this error. How would a clean checkout have a cache in the `.yarn` folder?

I suspect there's more to this story - like maybe incompatible node versions used to run yarn on your local machine vs the docker environment.

asymingt:
> > I suspect that the `COPY . ./` command in the studio Dockerfile includes the `.yarn` [folder](https://github.com/foxglove/studio/tree/main/.yarn), which has an old cache in it (I might be wrong about this) and the `yarn cache clean` fixes this before the build.
> 
> Your comment says that a clean checkout of Studio yields this error. How would a clean checkout have a cache in the `.yarn` folder?
> 
> I suspect there's more to this story - like maybe incompatible node versions used to run yarn on your local machine vs the docker environment.

You have much better knowledge of how yarn works. It could very well have been that I attempted to first build outside of Docker and then inside. But a containerized environment should have isolation from any persisted build products. Is there anything from the `COPY . ./` that might move intermediary build products into the container. Can we find a way to omit these?

defunctzombie:
> Can we find a way to omit these?

We do use a .dockerignore file (https://github.com/foxglove/studio/blob/main/.dockerignore) which by-design ignores `node_modules` but allows `.yarn`. The purpose is to avoid re-downloading these packages in your container if you've already downloaded them into the yarn cache locally for the project.

I'd not have thought this to be a problem since the same version of yarn would be run outside (on your host) and inside the container. Which version of node is on your host that you may have run `yarn` with the first time?

asymingt:
> > Can we find a way to omit these?
> 
> We do use a .dockerignore file (https://github.com/foxglove/studio/blob/main/.dockerignore) which by-design ignores `node_modules` but allows `.yarn`. The purpose is to avoid re-downloading these packages in your container if you've already downloaded them into the yarn cache locally for the project.
> 
> I'd not have thought this to be a problem since the same version of yarn would be run outside (on your host) and inside the container. Which version of node is on your host that you may have run `yarn` with the first time?

Here's the yarn installed on system by default
```sh
asymingt@mars:~/development/adehome$ yarn --version
0.32+git
```

defunctzombie:
> Here's the yarn installed on system by default

That's not the version of yarn that would be used in the Studio repo. When you `cd` to the studio repo and run `yarn --version` what do you see? Did you run yarn somehow outside of that repo on the repo itself? We have a specific version of yarn that we use in the studio repo.

asymingt:
> > Here's the yarn installed on system by default
> 
> That's not the version of yarn that would be used in the Studio repo. When you `cd` to the studio repo and run `yarn --version` what do you see? Did you run yarn somehow outside of that repo on the repo itself? We have a specific version of yarn that we use in the studio repo.

Doing it in the studio directory doesn't change the result:

```
asymingt@mars:~/development/grids/studio$ git status
On branch main
Your branch is up to date with 'origin/main'.

nothing to commit, working tree clean
asymingt@mars:~/development/grids/studio$ yarn --version
0.32+git
asymingt@mars:~/development/grids/studio$ 
```

asymingt:
> yarn --version

Switching directories to studio doesn't appear to change the result.

```
asymingt@mars:~/development/grids/studio$ yarn --version
0.32+git
```


> > > Here's the yarn installed on system by default
> > 
> > 
> > That's not the version of yarn that would be used in the Studio repo. When you `cd` to the studio repo and run `yarn --version` what do you see? Did you run yarn somehow outside of that repo on the repo itself? We have a specific version of yarn that we use in the studio repo.
> 
> Doing it in the studio directory doesn't change the result:
> 
> ```
> asymingt@mars:~/development/grids/studio$ git status
> On branch main
> Your branch is up to date with 'origin/main'.
> 
> nothing to commit, working tree clean
> asymingt@mars:~/development/grids/studio$ yarn --version
> 0.32+git
> asymingt@mars:~/development/grids/studio$ 
> ```

So, I went back to try and repeat the instructions to work out why I had an issue on my machine (Ubuntu 22.04 x86_64). Following the README install steps I hit this. 

```
asymingt@mars:~/development/grids/studio$ corepack enable
Internal Error: EINVAL: invalid argument, readlink '/usr/bin/yarn'
Error: EINVAL: invalid argument, readlink '/usr/bin/yarn'
```

So, I must have done something bad to get yarn / corepack as a result of this.  I believe corepack might now be enabled by default in nodejs, so perhaps this step doesn't need to be run?

If I do these steps I get a different result...

```sh
asymingt@mars:~/development/grids/studio$ curl -fsSL https://deb.nodesource.com/setup_current.x | sudo -E bash -
asymingt@mars:~/development/grids/studio$ sudo apt install nodejs
...
asymingt@mars:~/development/grids/studio$ yarn install
...
asymingt@mars:~/development/grids/studio$ yarn --version
3.1.0
```

Is that what you'd expect?

asymingt:
Proof of repeatability -- copy this to a `Dockerfile`.

```Dockerfile
# Base image and shell
FROM ubuntu:22.04
SHELL ["/bin/bash", "-c"]

# Dependencies
RUN apt-get update && apt-get install -q -y --no-install-recommends \
    ca-certificates \
    curl \
    dirmngr \
    git \
    git-lfs \
    gnupg2 \
    lsb-release \
    && rm -rf /var/lib/apt/lists/*
RUN curl -fsSL https://deb.nodesource.com/setup_current.x | bash -
RUN apt-get update && apt-get install -q -y --no-install-recommends \
    nodejs \
    && rm -rf /var/lib/apt/lists/*

# Add an 'ubuntu' user with dialout and plugdev access
RUN useradd -ms /bin/bash ubuntu && echo "ubuntu:ubuntu" | chpasswd
USER ubuntu
WORKDIR /home/ubuntu

# Run the steps
RUN git clone https://github.com/foxglove/studio.git
RUN corepack enable
RUN yarn install
```

Then run `docker build --tag foxglove:problem .`

And you will see:

```
...
Internal Error: EACCES: permission denied, symlink '../lib/node_modules/corepack/dist/pnpm.js' -> '/usr/bin/pnpm'
Error: EACCES: permission denied, symlink '../lib/node_modules/corepack/dist/pnpm.js' -> '/usr/bin/pnpm'
The command '/bin/bash -c corepack enable' returned a non-zero code: 1
```

defunctzombie:
> I believe corepack might now be enabled by default in nodejs, so perhaps this step doesn't need to be run?

It is still experimental: https://nodejs.org/dist/latest-v19.x/docs/api/corepack.html#enabling-the-feature

Somehow you ended up with yarn installed globally. That will indeed cause issues and is not the documented way to install yarn: https://yarnpkg.com/getting-started/install

It does seem that `https://deb.nodesource.com/setup_current.x` prints a command at the end to tell users to `apt-get install yarn` which ultimately is at odds with corepack (as you experienced above).

defunctzombie:
I can confirm that your above script in the dockerfile does fail on the `corepack enable` stage.

defunctzombie:
I've given this some more thought and am going to close this out. I think with the example Dockerfile the issue comes from some unhappiness with the useradd and how that interacts with the node install. The only supported workflow for Docker we have is with the Dockerfile in our repo which continues to work (to my knowledge). If you find that you can reproduce a failure with _that_ dockerfile we can take a look. Otherwise there are many ways to make bespoke Dockerfiles that don't work for some various reason or sadness with corepack or installations (as shown in this thread).

PR #5198, opened 2023-01-23
3D Panel TransformTree addTransform performance optimization

snosenzo:
**User-Facing Changes**

- 3D panel performance improvement after transform history capacity has been exceeded.

**Description**
 Optimize addTransform performance by amortizing trimming of arraymap. CoordinateTransform takes optional parameter specifying percentage of overfill the history has before trimming it to size. This makes it so that the transform history is trimmed every few thousand transforms instead of every transform, which was building up time-cost with high-frequency transforms. 

Optimization work needed for transform preloading. 

FG-1492

### DIFF
--- a/packages/den/collection/ArrayMap.ts
+++ b/packages/den/collection/ArrayMap.ts
@@ -60,6 +60,12 @@ export class ArrayMap<K, V> {
     this._list.length = greaterThanIndex;
   }
 
+  /** Removes all elements with keys less than the given key. */
+  public removeBefore(key: K): void {
+    const index = this.binarySearch(key);
+    this._list.splice(0, index);
+  }
+
   /** Access the first key/value tuple in the list, without modifying the list. */
   public minEntry(): [K, V] | undefined {
     return this._list[0];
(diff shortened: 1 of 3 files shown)


jhurliman (approved):

packages/den/collection/ArrayMap.ts
```diff
@@ -60,6 +60,12 @@ export class ArrayMap<K, V> {
     this._list.length = greaterThanIndex;
   }
 
+  /** Removes all elements with keys less than the given key. */
+  public removeBefore(key: K): void {
+    const index = this.binarySearch(key);
+    this._list.splice(0, index);
+  }
```
jtbandes, line 67: Doesn't seem to handle negative return values from `binarySearch`. Add unit tests?

packages/studio-base/src/panels/ThreeDeeRender/transforms/CoordinateFrame.ts
```diff
@@ -39,6 +39,8 @@ export class CoordinateFrame {
   public readonly id: string;
   public maxStorageTime: Duration;
   public maxCapacity: number;
+  // The percentage of maxCapacity that can be exceeded before overfilled frames in history are cleared
```
jtbandes, line 42: maybe note here about performance motivation for adding this?

packages/den/collection/ArrayMap.ts
snosenzo: good catch, added for both removeAfter and removeBefore

PR #5199, opened 2023-01-23
Bump yarn to 3.3.1

jtbandes:
**User-Facing Changes**
None

**Description**
Attempt to address build failure on #5182. May be related to yarn's built-in TS patch which was supposedly updated in 3.3.1 (https://github.com/yarnpkg/berry/issues/5146 / https://github.com/yarnpkg/berry/issues/5125).


defunctzombie (approved):
What's some more lockfiles between friends.

We did figure out that the resolution to the other issue was the inlineHunks right? So maybe we don't care about this now? Or maybe we upgrade anyway. I leave this to you.

jtbandes:
An upgrade in the hand is worth two in the backlog.

PR #5200, opened 2023-01-23
Use local timestamps for message frequency calcs

### LINKED ISSUE #5140: Data Source Info shows impossibly high message frequency for rosbridge
In the attached video, the camera message count starts at 33 and ends at 82 over 14 seconds. This is a 3.5Hz message frequency. However, the topic frequency stays consistently over 20Hz. When streaming the same topic using the ws-protocol player and foxglove_bridge, the message frequency is reported as ~3.5Hz.

https://user-images.githubusercontent.com/195374/212196434-7e9b7894-d606-4fcd-8880-c7ab053d0d6f.mp4

jhurliman:
**User-Facing Changes**
- Fixed message frequency statistics when connected to a live data source that publishes clock messages

**Description**
Don't use server-provided timestamps such as from a /clock topic when calculating message receive frequencies. Fixes #5140


packages/studio-base/src/players/RosbridgePlayer.ts
```diff
@@ -535,12 +535,6 @@ export default class RosbridgePlayer implements Player {
               this._providerTopicsStats.set(topicName, stats);
             }
             stats.numMessages++;
-            stats.firstMessageTime ??= receiveTime;
-            if (stats.lastMessageTime == undefined) {
-              stats.lastMessageTime = receiveTime;
-            } else if (isGreaterThan(receiveTime, stats.lastMessageTime)) {
-              stats.lastMessageTime = receiveTime;
-            }
```
jtbandes, line 543: Seems like you added these in https://github.com/foxglove/studio/pull/3175 — was there something we got wrong the first time?

It also looks like FoxgloveWebSocketPlayer sets `firstMessageTime` and `lastMessageTime` — why do we want to set these in one live player but not the other?

jhurliman: Setting these in a live player is redundant as of <https://github.com/foxglove/studio/pull/5007>. I'll fix FoxgloveWebSocketPlayer as well.

jtbandes: Should we add a doc comment on the type definition for these that explains they shouldn't be set for live players?

jtbandes (approved):

PR #5212, opened 2023-01-24
Change block loading behavior to load from beginning to end of source

snosenzo:
**User-Facing Changes**

 - topics that require preloading (like for plots) will load from the beginning of the file to the end
 - remove preloaded topic block cache eviction and replace with error.

**Description**
Change block preloading behavior to load from beginning to end of the file instead of from the activeBlock to the end and then the beginning to `activeBlock`. Replaces block eviction behavior (evicting blocks not in previously active range in front of currentTime) with removing unused preloaded topics from blocks and show error in source panel when the cache is full. So block eviction will not occur and preloading will stop for preloaded topics when the cache is full. The user however can and is prompted lower the number of preloaded topics in panels that preload them (e.g. plot panel) to preload more desired topics. 
See behavior here:
https://user-images.githubusercontent.com/10187776/215120423-29a887ab-b9da-4283-9b23-cd600c0fe102.mov

FG-1493


snosenzo:
Still need to fix tests and address:
 >  In this state, if eviction happens before we get to the activeblock it never loads to the active block, which means preloading doesn’t really happen and only that portion in the beginning will be loaded on a plot. A solution might be to make the blockloader have more specific actions for seek and playback such that whenever there’s a seek it snakes (remove from front, add to end) its way to and past the activetime, then playback just snakes along with the activeblock, essentially making it act like a cursor that guarantees its been iterated through to the activeblock at each new frame. or we can just take the approach where we don’t load more after the cache is full and that’s that (with an error message)

snosenzo:
> 

this has now been addressed, and the main PR description has been updated to reflect the new behavior with no cache eviction.

packages/studio-base/src/players/IterablePlayer/BlockLoader.ts
```diff
@@ -148,7 +168,7 @@ export class BlockLoader {
 
       // The active block id is the same as when we started.
```
defunctzombie, line 169: outdated comment

packages/studio-base/src/players/IterablePlayer/BlockLoader.ts
```diff
@@ -135,7 +156,6 @@ export class BlockLoader {
     while (!this.stopped) {
       this.abortController = new AbortController();
 
-      const activeBlockId = this.activeBlockId;
       const topics = this.topics;
 
       // Load around the active block id, if the active block id changes then bail
```
defunctzombie, line 161: Outdated comment

packages/studio-base/src/players/IterablePlayer/BlockLoader.ts
```diff
@@ -167,12 +187,8 @@ export class BlockLoader {
...
     }
 
-    // Load the active block to end first, then go back and load first block to active block
-    await this.loadBlockRange(this.activeBlockId, this.blocks.length, args.progress);
-
-    if (this.activeBlockId > 0) {
-      await this.loadBlockRange(0, this.activeBlockId, args.progress);
-    }
+    // Load blocks from beginning to end of the file
+    await this.loadBlockRange(0, this.blocks.length, args.progress);
```
defunctzombie, line 191: Arguably there's no need for a range anymore and the block loading itself knows it will always go from 0 to length.

packages/studio-base/src/players/IterablePlayer/BlockLoader.ts
```diff
@@ -306,23 +322,21 @@ export class BlockLoader {
...
-              startId: this.activeBlockId,
-              endId: currentBlockId,
-            });
-            // If we could not evict any blocks to bring our size down, then we stop loading more data
-            if (evictedSize === 0) {
-              log.debug("could not evict more blocks", {
-                totalBlockSizeBytes,
-                messageSizeInBytes,
-                maxCache: this.maxCacheSize,
+          if (totalBlockSizeBytes > this.maxCacheSize) {
```
defunctzombie, line 325: I'd find this easier to read if the `if` condition was flipped, and the removeProblem moved into this first condition followed by a `continue`.

Then the remaining logic could be un-indented and would represent what to do if the cache size is full.

defunctzombie (approved):
👏

PR #5213, opened 2023-01-24
`renderState.allFrames` no longer includes topics that the panel didn't request to be preloaded

snosenzo:
**User-Facing Changes**
 - [extension-panels] `renderState.allFrames` no longer includes topics that the panel didn't request to be preloaded

**Description**
Filter topics that the panel doesn't want preloaded out of `allFrames`. This allows extension panels to not have to iterate through messages that they didn't ask to have preloaded. Previously `allFrames` was constructed from all topics that the panel subscribed to that had loaded blocks regardless of whether they were preloaded. 

FG-1625

### DIFF
--- a/packages/studio-base/src/components/PanelExtensionAdapter/renderState.ts
+++ b/packages/studio-base/src/components/PanelExtensionAdapter/renderState.ts
@@ -2,6 +2,7 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
+import { filterMap } from "@foxglove/den/collection";
 import Log from "@foxglove/log";
 import { toSec } from "@foxglove/rostime";
 import {
@@ -279,12 +280,11 @@ function initRenderStateBuilder(): BuildRenderStateFn {
         shouldRender = true;
         const frames: MessageEvent<unknown>[] = (renderState.allFrames = []);
         // only populate allFrames with topics that the panel wants to preload
-        const topicsToPreloadForPanel = new Set<string>();
-        for (const sub of subscriptions) {
-          if (sub.preload === true) {
-            topicsToPreloadForPanel.add(sub.topic);
-          }
-        }
+        const topicsToPreloadForPanel = new Set<string>(
+          Array.from(
+            filterMap(subscriptions, (sub) => (sub.preload === true ? sub.topic : undefined)),
+          ),
+        );
         for (const block of newBlocks) {
           if (!block) {
             continue;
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/components/PanelExtensionAdapter/renderState.ts
```diff
@@ -278,12 +278,23 @@ function initRenderStateBuilder(): BuildRenderStateFn {
...
+            topicsToPreloadForPanel.add(sub.topic);
+          }
+        }
         for (const block of newBlocks) {
           if (!block) {
             continue;
           }
 
-          for (const messageEvents of Object.values(block.messagesByTopic)) {
+          for (const topic of topicsToPreloadForPanel) {
```
jhurliman, line 293: Is it a problem if allFrames has an unstable sort order due to the `Set<>`?

snosenzo: ah good catch I'll take that into account for when I implement the allFrames sorted guarantee, but I don't think that's an issue for now since there are no guarantees of allFrames order across adding/removing preloaded topics.

snosenzo: also as long as `subscriptions` has a stable order the set should as well.

defunctzombie:
> * [extension-panels] `renderState.allFrames` no longer includes topics that the panel didn't request to be preloaded

Would this happen when another panel _did_ want those topics preloaded?

packages/studio-base/src/components/PanelExtensionAdapter/renderState.ts
```diff
@@ -278,12 +278,23 @@ function initRenderStateBuilder(): BuildRenderStateFn {
       if (newBlocks && prevBlocks !== newBlocks) {
         shouldRender = true;
         const frames: MessageEvent<unknown>[] = (renderState.allFrames = []);
+        // only populate allFrames with topics that the panel wants to preload
+        const topicsToPreloadForPanel = new Set<string>();
```
defunctzombie, line 282: A potential one-liner. Don't know if clearer but I figured I'd share.
`new Set<string>(Array.from(subscriptions).filter(sub => sub.preload).map(sub.topic))`

defunctzombie (approved):

packages/studio-base/src/components/PanelExtensionAdapter/renderState.ts
jtbandes: filterMap would be faster than Array.from().filter().map() 😃

PR #5225, opened 2023-01-25
Add setDefaultPanelTitle to extension API, and ability to edit panel titles

jtbandes:
**User-Facing Changes**
- Added `setDefaultPanelTitle()` to `PanelExtensionContext` API
- Added ability to edit the title of most panels in panel settings.

**Description**
- Adds `setDefaultPanelTitle` for panel to override its default title (e.g. based on configured message path)
- Inject a text field in settings sidebar to edit a hard-coded settings key, `foxglovePanelTitle` for a custom title override
- Use the overridden title if set, otherwise the default title if set, otherwise fall back to the panel name (as we already do today)

Resolves https://linear.app/foxglove/issue/FG-1488/allow-user-to-rename-any-panel

https://user-images.githubusercontent.com/14237/214679560-350d5fbe-4687-4aeb-8458-1bd2ea012694.mov

### DIFF
--- a/packages/studio-base/src/components/PanelExtensionAdapter/PanelExtensionAdapter.tsx
+++ b/packages/studio-base/src/components/PanelExtensionAdapter/PanelExtensionAdapter.tsx
@@ -112,15 +112,7 @@ function PanelExtensionAdapter(props: PanelExtensionAdapterProps): JSX.Element {
   const isPanelInitializedRef = useRef(false);
 
   const [slowRender, setSlowRender] = useState(false);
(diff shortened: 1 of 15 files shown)


amacneil:
👏 much excite

amacneil:
Review feedback:
- People are happy with the extension API
- Plot panel needs to be fixed
- Camel case in layout json
- Can we hide the `Title` input for non-extension panels (since it is broken for them) - image, publish, tab, etc

foxymiles (approved):
Nice work!

PR #5226, opened 2023-01-25
`renderState.allFrames` made sorted by `msg.receiveTime`

snosenzo:
**User-Facing Changes**

 - [extension-panels] - `renderState.allFrames` will be sorted by message `receiveTime`

**Description**
Add function to merge sorted topics arrays to form sorted `allFrames`. This allows extension panels to have some more stability in allFrames and make assumptions about it that could allow for performance optimizations. This does scale the time-complexity by the number of preloaded topics, but in most cases that should be fairly small. 

FG-1631


packages/studio-base/src/components/PanelExtensionAdapter/renderState.ts
```diff
@@ -381,3 +381,43 @@ function initRenderStateBuilder(): BuildRenderStateFn {
 }
 
 export { initRenderStateBuilder };
+
+function forEachSortedArrays<Item>(
```
defunctzombie, line 385: This type of data-structure wrangling code definitely needs a comment describing what the intent of this function, inputs/outputs, relevant implementation comments (any gotchas or guidance); and finally it needs tests.

Guessing you checked lodash and ramda for having anything similar to this that we could levarage?

defunctzombie:
Hows the runtime or big-o change with this? Blocks tends to be the more expensive stuff to wrangle - I'd like to better understand how many iterations we are going from/to for this effort and how many new arrays or maps we are making in the process. My initial feeling is that the approach with cursors is right but that we can do it maybe without initializing more intermediate data structures or copies?

snosenzo:
> Hows the runtime or big-o change with this?

The big O change with this is as I stated in the description. It scales this function by (at max) the number of topics that the panel is requesting to preload, which should be relatively small. So the big O runtime is O(t * n) where t is the number of topics and n is the number of messages. 

> we can do it maybe without initializing more intermediate data structures or copies

The only intermediate data structures is the array of arrays going into the function, which is only the length of preloaded topics, and the array of cursors, also the length of the topics. The biggest cost in this is creating a new array for allFrames every update. If we just added/subtracted to it based off of the blocks it would be ideal: not having to sort all of the blocks of it each time, and wouldn't have to create a new huge allFrames array each time. However we would lose the referential difference between different allFrames that causes updates (at least in react) to downstream consumers and would probably need another variable to track changes to allFrames. Open to exploring that more though

packages/studio-base/src/components/PanelExtensionAdapter/renderState.ts
snosenzo: yeah,  I wasn't able to find anything similar in either of those places.

defunctzombie (approved):
Probably doesn't matter but `Array.from(topicsToPreloadForPanel)` could be done outside of the loop.

PR #5227, opened 2023-01-25
Fix type definitions in studio/index.d.ts

jtbandes:
**User-Facing Changes**
Fixed some invalid TypeScript types in the `@foxglove/studio` package.

**Description**
Use an `index.ts` source file with a build step, rather than hand-writing `index.d.ts`, to ensure TypeScript actually checks for errors in CI.


packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
```diff
@@ -586,7 +587,10 @@ export default class FoxgloveWebSocketPlayer implements Player {
     }
 
     log.debug(`FoxgloveWebSocketPlayer.setParameter(key=${key}, value=${value})`);
-    this._client.setParameters([{ name: key, value }], uuidv4());
+    this._client.setParameters(
+      [{ name: key, value: value as WebSocketParameter["value"] }],
+      uuidv4(),
+    );
```
jtbandes, line 593: @jhurliman @achim-k it seems like the incoming value can be `undefined`, but the `@foxglove/ws-protocol` package doesn't accept `undefined` params. What should we do here?

packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
```diff
@@ -503,9 +503,9 @@ export function ThreeDeeRender({ context }: { context: PanelExtensionContext }):
       const id = selection?.renderable.idFromMessage();
       const customVariable = selection?.renderable.selectedIdVariable();
       if (customVariable) {
-        context.setVariable(customVariable, id ?? ReactNull);
+        context.setVariable(customVariable, id);
       }
-      context.setVariable(SELECTED_ID_VARIABLE, id ?? ReactNull);
```
jtbandes, line 508: @jhurliman any idea why null was used here explicitly?

packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
achim-k: The spec currently does not allow setting parameter values to `undefined` (I guess that means removing them?). Should we add that?

packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
jhurliman: Using `undefined` was causing the variable to be removed entirely, where it should be set to empty here but not removed.

packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
jhurliman: One option is to update `ws-protocol` to allow `undefined` for params, and delete params that are set to `undefined`.

achim-k: Yes we could do that. It's not a breaking change. Will create a ticket for that.

packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
jtbandes: Looks like this isn't working on main now, so I think it's ok to proceed with this change.

packages/studio-base/src/players/Ros1Player.ts
```diff
@@ -139,7 +139,7 @@ export default class Ros1Player implements Player {
 
       rosNode.on("paramUpdate", ({ key, value, prevValue, callerId }) => {
         log.debug("paramUpdate", key, value, prevValue, callerId);
-        this._parameters = new Map(rosNode.parameters);
+        this._parameters = new Map(rosNode.parameters as Map<string, ParameterValue>);
```
jtbandes, line 142: @jhurliman @achim-k Let me know if you can think of better ideas here. Does the ros 1 param server actually support these types (Date, Uint8Array, Record<string,XmlRpcValue>)?

defunctzombie: We have learned that it does (from what I read in slack). So why do we need the cast here?

packages/studio-base/src/components/VariablesSidebar/Variable.tsx
```diff
@@ -137,7 +138,7 @@ export default function Variable(props: {
 
   const onChangeValue = useCallback(
     (newVal: unknown) => {
-      setGlobalVariables({ [name]: newVal });
+      setGlobalVariables({ [name]: newVal as VariableValue });
```
defunctzombie, line 141: Should we add a check that `newVal` conforms to `VariableValue` types? In recent times I've started to prefer type narrowing functions rather than `as` casts since these tend to surface as errors later when we are wondering why something downstream is failing.

defunctzombie (changes requested):
I don't understand this change in `studio-base/tsconfig.json`. This does not align with how we've been using the package specific `tsconfig.json` files.
```
"rootDir": "..",
```

packages/studio-base/src/components/VariablesSidebar/Variable.tsx
jtbandes: This comes from a JSON.parse. We treat that as unknown so it needs a cast. I am not sure what would be an appropriate level of validation but I feel like somehow doing something like a full ajv schema is overkill. Writing a validator by hand that appropriately handles arrays and nested objects would also feel like overkill.

packages/studio-base/tsconfig.json
```diff
@@ -2,14 +2,15 @@
   "extends": "@foxglove/tsconfig/base",
   "include": ["./src/**/*", "./src/**/*.json", "../package.json"],
   "compilerOptions": {
-    "rootDir": "./src",
+    "rootDir": "..",
```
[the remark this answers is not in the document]
jtbandes: Are these a concern too?

image

defunctzombie (approved):

PR #5228, opened 2023-01-25
Use sentence case for menu items, panel settings, etc.

esthersweon:
**User-Facing Changes**

- App-wide audit of using sentence case for menu items, panel settings, and other copy
- Use "Min" / "Max" consistently in panel settings (vs. "Minimum" / "Maximum")


jhurliman:
Horizontal space is at a premium for sidebar settings, we've tried to find succinct titles for each setting so it's readable (otherwise you see something like "Synchronize Ti..."). I recommend using "Min" and "Max" consistently in the settings instead of "Minimum" and "Maximum".

packages/studio-base/src/panels/Image/settings.ts
```diff
@@ -50,37 +50,37 @@ export function buildSettingsTree({
       label: "General",
       fields: {
         cameraTopic: {
-          label: "Camera Topic",
+          label: "Camera topic",
```
jtbandes, line 53: This is changing in another PR, right?

jtbandes (approved):

packages/studio-base/src/panels/Image/settings.ts
esthersweon: Ah yes, will just be "Topic". Will edit here.

PR #5233, opened 2023-01-25
Remove cog from all headings in the panel settings

esthersweon:
**User-Facing Changes**
- Remove cog from all 'General' sections in panel settings

### DIFF
--- a/packages/studio-base/src/components/StudioLogsSettingsSidebar/useStudioLogsSettingsTree.ts
+++ b/packages/studio-base/src/components/StudioLogsSettingsSidebar/useStudioLogsSettingsTree.ts
@@ -30,7 +30,6 @@ function useStudioLogsSettingsTree(): SettingsTree {
     const settingsRoot: SettingsTreeNodes = {
       Settings: {
         label: "Settings",
-        icon: "Settings",
         fields: {
           level: {
             label: "level",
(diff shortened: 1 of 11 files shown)


amacneil:
I haven't gone through to audit, but do any of these make sense?

E.g. if there is a list of sections and each section has an icon, then we should keep the icon. If none of the other sections have an icon then we should remove the icon.

esthersweon:
@amacneil - they all make sense to remove (they're either the only General section, or the only section with a cog). Most are covered by [the screenshots here](https://www.chromatic.com/pullrequest?appId=603ec8bf7908b500231841e2&number=5233&activeElementId=comparison-test-key-62a7947f6331ea695db36ee3-1200&view=changes).

List of changes:
- Studio logs sidebar pane
- Log, Map, Plot, Publish, State Transitions, User Scripts, URDF, Variable Slider, both Diagnostics panels
- Settings underneath the 3D panel's "Transforms" section

I can see the last one being the only controversial one. Maybe we can keep that cog, to differentiate from the frames?

Screen Shot 2023-01-25 at 4 06 14 PM

amacneil:
Sounds good.

Re the contentious transform one, it was already removed in https://github.com/foxglove/studio/pull/5230

amacneil (approved):

jtbandes:
Hm, now that I think of it, maybe it is better to revert #5230 to keep it distinct from the frame names themselves? What if you have a frame named Settings? 😅

amacneil:
@jtbandes we can still discuss reverting that one if you want, regardless of this PR.

I feel like the transform settings feel weird both with and without it.

PR #5241, opened 2023-01-26
Improve plot panel settings API integration

foxymiles:
**User-Facing Changes**
This improves the existing integration of the plot panel and the settings API.

**Description**
This makes the following changes to the plot panel's use of the settings API:

* Add Label to series, default to empty string, if empty use Path as label
* Remove existing "Rename" pencil icon from series settings
* Remove legend series X
* Make the series name grey in the legend when hidden
* Remove legend hamburger menu and adds a separate `Show legend` toggle setting.
* Rename Show plot values to Show values 

Note that I have opted to make `Show legend` a separate boolean instead of a new `None` option in the position setting because this maps directly to the existing plot panel config schema and allows us to seamlessly switch back and forth between the existing UI and the new experimental UI without any conversion or migration.

Screenshot 2023-01-26 at 9 01 07 AM

### DIFF
--- a/packages/studio-base/src/panels/Plot/NewPlotLegendRow.tsx
+++ b/packages/studio-base/src/panels/Plot/NewPlotLegendRow.tsx
@@ -5,7 +5,6 @@
 import CircleIcon from "@mui/icons-material/Circle";
 import CircleOutlinedIcon from "@mui/icons-material/CircleOutlined";
 import CircleTwoToneIcon from "@mui/icons-material/CircleTwoTone";
-import CloseIcon from "@mui/icons-material/Close";
 import ErrorIcon from "@mui/icons-material/Error";
 import { IconButton, Tooltip, Typography } from "@mui/material";
 import { ComponentProps, useMemo, useState } from "react";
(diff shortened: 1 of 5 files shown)


packages/studio-base/src/panels/Plot/NewPlotLegendRow.tsx
```diff
@@ -32,11 +31,23 @@ type PlotLegendRowProps = {
...
+/**
+ * Coalesces null, undefined and empty string to undefined.
+ */
+function presence<T>(value: undefined | T): undefined | T {
+  if (value === "") {
+    return undefined;
+  }
+
+  return value == undefined ? undefined : value;
+}
```
jtbandes, line 43: the return of `nonEmptyOrUndefined` :smile:  https://github.com/foxglove/studio/commit/00e18ed0045ed2ac5f71bb8a328f06a298e975e6#diff-3c8c5d3ac062f89d7eda65fb08dc06fcc4055563cd6813cc52fe5837d744d8fe

jtbandes (approved):

PR #5243, opened 2023-01-26
Update local file help text

amacneil:
**User-Facing Changes**
Update text below `Open local file` in welcome dialog.

image

**Description**
We had some feedback from users that they were confused whether `Open local file` would upload the file to Foxglove. This is an attempt at improving the text to reassure folks.

FG-989

### DIFF
--- a/packages/studio-base/src/components/OpenDialog/Start.tsx
+++ b/packages/studio-base/src/components/OpenDialog/Start.tsx
@@ -18,7 +18,6 @@ import { AppEvent } from "@foxglove/studio-base/services/IAnalytics";
 import { OpenDialogViews } from "./types";
 
 export type IStartProps = {
-  supportedLocalFileExtensions?: string[];
   onSelectView: (newValue: OpenDialogViews) => void;
 };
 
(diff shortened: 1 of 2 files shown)


amacneil:
Looking for feedback on whether people like this wording, or have better suggestions.

packages/studio-base/src/components/OpenDialog/Start.tsx
```diff
@@ -410,21 +409,17 @@ function SidebarItems(props: { onSelectView: (newValue: OpenDialogViews) => void
...
-    const formatter = new Intl.ListFormat("en-US", { style: "long" });
-    const supportedLocalFiles = formatter.format(
-      Array.from(new Set(supportedLocalFileExtensions)).sort(),
-    );
     return [
       {
         key: "open-local-file",
         text: "Open local file",
-        secondaryText: `Supports ${supportedLocalFiles} files.`,
+        secondaryText: "Visualize directly from your local filesystem.",
```
esthersweon, line 422:
```suggestion
        secondaryText: "Visualize data directly from your local filesystem.",
```

esthersweon (approved):

packages/studio-base/src/components/OpenDialog/Start.tsx
esthersweon, line 422:
```suggestion
        secondaryText: "Visualize data directly from your local filesystem.",
```

I think having in-app documentation about what the supported file formats are is still useful. Could we combine both?

ocin-rye (approved):
Looks good. It should reduce some of the confusion experienced by users.

defunctzombie:
Bowing out of the review here. I am confident the remaining wordsmiths will produce a good result.

amacneil:
> I think having in-app documentation about what the supported file formats are is still useful. Could we combine both (e.g. "Visualize ${supportedLocalFiles} files directly from your local filesystem")?

@esthersweon we do link to the docs from the welcome dialog, and the "open file" picker will already highlight supported file types. Do you think we need more than that?

PR #5254, opened 2023-01-27
Fix new preference dialog height and scrolling issues

foxymiles:
**User-Facing Changes**
Fixes the height & scrolling issues with the new preferences dialog.

**Description**
This makes two changes to the preferences dialog:
1. Sets a fixed 70vh height and "md" fixed width.
2. Confines scrolling to the right hand pane.

https://user-images.githubusercontent.com/93935560/215014496-d272d33c-8bbc-410c-9522-8069306ff037.mov


amacneil (approved):
Seems to work well in my testing 👍

PR #5255, opened 2023-01-27
Support inline actions on settings nodes

foxymiles:
**User-Facing Changes**
Make it possible for panel authors to request that node actions be rendered inline instead of in the context menu.

**Description**
Expands the `SettingsTreeNodeActionItem` to include an optional `display` property, which can be set to `inline` or `menu`. If `inline` it will be rendered inline alongside the node label instead of in the context menu. This should make frequently used actions more efficient and discoverable.

https://user-images.githubusercontent.com/93935560/215614093-e5ab70f4-efbd-4b67-a597-45f468c9ea8a.mov

Some open questions:
1. Should these always be shown or only on hover?
2. Should we render them in the primary color?


amacneil:
Do we think we will every want to support _both_ inline and menu for "more actions"?

foxymiles:
> Do we think we will ever want to support _both_ inline actions and a menu for "more actions"?

That works with this API if you specify the same action twice, once as an inline action and once as a menu action.

defunctzombie (approved):

PR #5256, opened 2023-01-27
Foxglove websocket: Fix error when server sends invalid `serverInfo` message

achim-k:
**User-Facing Changes**
- Foxglove websocket: Fix error when server sends invalid `serverInfo` message

**Description**
This PR handles the special case when the server sends a [serverInfo](https://github.com/foxglove/ws-protocol/blob/e2ad9af97694d00bbd50dc2aba91740e486f5c9b/docs/spec.md#server-info) message with `capabilities` being `null` or `undefined` (or anything else that is not an array).

### DIFF
--- a/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
+++ b/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
@@ -189,7 +189,7 @@ export default class FoxgloveWebSocketPlayer implements Player {
       if (!Array.isArray(event.capabilities)) {
         this._problems.addProblem("ws:invalid-capabilities", {
           severity: "warn",
-          message: `Server capabilities communicated by server must be an array (server sent '${event.capabilities}')`,
+          message: `Server sent an invalid or missing capabilities field: '${event.capabilities}'`,
         });
       }
       this._name = `${this._url}\n${event.name}`;
--- a/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
+++ b/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
@@ -186,9 +186,15 @@ export default class FoxgloveWebSocketPlayer implements Player {
     });
 
     this._client.on("serverInfo", (event) => {
+      if (!Array.isArray(event.capabilities)) {
+        this._problems.addProblem("ws:invalid-capabilities", {
+          severity: "warn",
+          message: `Server capabilities communicated by server must be an array (server sent '${event.capabilities}')`,
+        });
+      }
       this._name = `${this._url}\n${event.name}`;
-      this._serverCapabilities = event.capabilities;
-      this._serverPublishesTime = event.capabilities.includes(ServerCapability.time);
+      this._serverCapabilities = Array.isArray(event.capabilities) ? event.capabilities : [];
+      this._serverPublishesTime = this._serverCapabilities.includes(ServerCapability.time);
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
```diff
@@ -186,9 +186,15 @@ export default class FoxgloveWebSocketPlayer implements Player {
     });
 
     this._client.on("serverInfo", (event) => {
+      if (!Array.isArray(event.capabilities)) {
+        this._problems.addProblem("ws:invalid-capabilities", {
+          severity: "warn",
```
defunctzombie, line 191: error?

defunctzombie: Maybe even fail the entire connection? This is a server that is violating the procol?

defunctzombie (approved):

packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
```diff
@@ -186,9 +186,15 @@ export default class FoxgloveWebSocketPlayer implements Player {
     });
 
     this._client.on("serverInfo", (event) => {
+      if (!Array.isArray(event.capabilities)) {
```
jhurliman, line 189: It seems like there is a deeper problem with typing here. Before this change, we assumed somewhere that `event.capabilities` exists and is an array. But this is coming from a parsed external payload, so we can't make any strong assumptions about the shape of the data or the existence of fields. Do we need a `DeepPartial<...>` somewhere?

packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
jhurliman: The "warn" log level contradicts the "must be" statement in the warning message. I recommend either making it an error and bail out of the connection, or keep it a warning and change the message to `"Server sent an invalid or missing capabilities field: '${event.capabilities}'"`

packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
```diff
@@ -186,9 +186,15 @@ export default class FoxgloveWebSocketPlayer implements Player {
...
+        this._problems.addProblem("ws:invalid-capabilities", {
+          severity: "warn",
+          message: `Server capabilities communicated by server must be an array (server sent '${event.capabilities}')`,
```
achim-k, line 192:
```suggestion
          message: `Server sent an invalid or missing capabilities field: '${event.capabilities}'`,
```

packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
achim-k: Should we handle this in `@foxglove/ws-protocol` instead and emit an `error` event when the `serverInfo` message is invalid?

jhurliman: I think that's the right place to catch a protocol error. We can merge this first, then follow up with catching this in `@foxglove/ws-protocol`

achim-k: -> https://github.com/foxglove/ws-protocol/issues/348

PR #5264, opened 2023-01-30
Fix `web-integration` check to run for forks

snosenzo:
**User-Facing Changes**

 - n/a

**Description**

 Remove github actions cache to allow for forks to use this web integration test. I believe that this is the issue since this is the only part of the job that depends on a GITHUB_REF. There is open issue about this on actions/cache. When that is closed we should be able to add this back. 
https://github.com/actions/cache/issues/319

FG-1472


defunctzombie:
> Remove github actions cache to allow for forks to use this web integration test.

I don't understand how this helps. For example regular CI uses actions cache and has no problem running for forks. Or do our regular CI jobs also fail for forks?

I am ok giving this a try and seeing what the impact is on this workflow's runtime since without a cache we will be downloading all of the packages for the workspace.

defunctzombie (approved):
I don't understand how the cache figures into forks (since forks still can run CI and that uses cache) but we can give it a try.

PR #5267, opened 2023-01-30
Safer fallback for legacy subscribe behavior

defunctzombie:
**User-Facing Changes**

**Description**

### DIFF
--- a/packages/studio-base/src/components/PanelExtensionAdapter/PanelExtensionAdapter.tsx
+++ b/packages/studio-base/src/components/PanelExtensionAdapter/PanelExtensionAdapter.tsx
@@ -366,11 +366,12 @@ function PanelExtensionAdapter(props: PanelExtensionAdapterProps): JSX.Element {
         if (!isMounted()) {
           return;
         }
+        const watchAllFrames = watchedFields.has("allFrames");
         const subscribePayloads = topics.map<SubscribePayload>((item) => {
           if (typeof item === "string") {
             // For backwards compatability with the topic-string-array api `subscribe(["/topic"])`
             // results in a topic subscription with full preloading
-            return { topic: item, preloadType: "full" };
+            return { topic: item, preloadType: watchAllFrames ? "full" : "partial" };
           }
 
           return {


packages/studio-base/src/components/PanelExtensionAdapter/PanelExtensionAdapter.tsx
```diff
@@ -366,11 +366,12 @@ function PanelExtensionAdapter(props: PanelExtensionAdapterProps): JSX.Element {
         if (!isMounted()) {
           return;
         }
+        const watchAllFrames = watchedFields.has("allFrames");
         const subscribePayloads = topics.map<SubscribePayload>((item) => {
           if (typeof item === "string") {
             // For backwards compatability with the topic-string-array api `subscribe(["/topic"])`
             // results in a topic subscription with full preloading
-            return { topic: item, preloadType: "full" };
+            return { topic: item, preloadType: watchAllFrames ? "full" : "partial" };
```
jtbandes, line 374: Not sure exactly what motivated this change but it seems like this would require the panel to watch allFrames _before_ calling subscribe. Historically we haven't required those calls to happen in a certain order.

defunctzombie: We ran into a bug where the image panel was subscribing to all image topics with `preloadType: full` because it was using the shorthand `subscribe(["topic"])` api syntax rather than the object based syntax. This shorthand syntax defaulted to subscribing the topic with full preloading since that was the behavior we had before we could allow you to specify this as a panel author. We figured a nice way to avoid this gotcha for some panels was to only use the "full" preload type if the author was also subscribed to "allFrames". If they are not then there's no reason to preload full.

But yes - you have noted the defect in this simple code change which is that it makes a dependency on the fields and we don't want that. This is why we have not merged this.

jtbandes:
Is this still something we want to pursue?

defunctzombie:
@jtbandes practically yes - I think this would help extension authors who are using the `string[]` form of subscribe but not watching "allFrames".

See: https://github.com/foxglove/studio/pull/5977 for additional ways we can nudge folks in a better direction.

PR #5272, opened 2023-01-31
Allow for TF Preloading

snosenzo:
**User-Facing Changes**

 - [3D Panel] TF preloading will now be enabled by default in the `Transform` Settings. This will guarantee that the TransformTree is accurate up to the current time in the file after preloading has completed. 

**Description**
There is an up-front performance cost, but it should result in low-frequency transforms always being included without the need for playback. After initial preloading, for most cases (<250Hz TFs, and < 1hr long files) seek experience won't appear degraded. However as these files reach outside of this range, they might experience longer seek times toward the end of the file in addition to longer preloading time. Outside of preloading, normal playback performance should be the same. Most costly operation outside of preloading is seeking to slightly before the cursor at the end of the file or from the very beginning to very end of file. This feature can be disabled in the Transform settings in the 3D Panel settings if a user wishes to return to the previous behavior. 

Also this PR addressed an issue where the Display Frame wasn't being updated when errors were added or removed. 

FG-1494

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
+++ b/packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
@@ -733,7 +733,7 @@ export function ThreeDeeRender({ context }: { context: PanelExtensionContext }):
 
     // if a seek occurred and the new time is before the current cursor time, reset the cursor for this read
     if (didSeek) {
-      if (cursorTimeReached && isGreaterThan(currentTime, cursorTimeReached)) {
+      if (cursorTimeReached && isGreaterThan(cursorTimeReached, currentTime)) {
         cursorTimeReached = undefined;
         cursor = -1;
       }
--- a/packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
+++ b/packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
@@ -21,7 +21,7 @@ import { DeepPartial } from "ts-essentials";
 import { useDebouncedCallback } from "use-debounce";
 
 import Logger from "@foxglove/log";
-import { Time, toNanoSec } from "@foxglove/rostime";
+import { Time, compare, isGreaterThan, isLessThan, toNanoSec } from "@foxglove/rostime";
 import {
   LayoutActions,
   MessageEvent,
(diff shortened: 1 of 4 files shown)


packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
```diff
@@ -693,7 +722,84 @@ export function ThreeDeeRender({ context }: { context: PanelExtensionContext }):
     }
   }, [backgroundColor, colorScheme, renderer]);
 
+  // Handle preloaded messages and render a frame if new messages are available
+  // Should be called before `messages` is handled
+  useEffect(() => {
```
defunctzombie, line 727: Something feels off to me about there being two useEffects that deal with the allFramesCursor. Not sure what yet.

packages/studio-base/src/panels/ThreeDeeRender/renderables/FrameAxes.ts
```diff
@@ -173,12 +173,11 @@ export class FrameAxes extends SceneExtension<FrameAxisRenderable> {
...
-          // Disabled until we can efficiently preload transforms
-          // enablePreloading: {
-          //   label: "Enable preloading",
-          //   input: "boolean",
-          //   value: config.scene.transforms?.enablePreloading ?? true,
-          // },
+          enablePreloading: {
+            label: "Enable preloading",
+            input: "boolean",
+            value: config.scene.transforms?.enablePreloading ?? true,
```
defunctzombie, line 179: Gave this some more thought. Let's start with this _off_ for the initial merge which will let the team test it with some more datasets. I might also ask some of our customers to flip this on and test it. Once we are ready - I think we will remove this option altogether from being something the user can override.

defunctzombie (commented):
Did a cursory (no pun intended) pass on this. LGTM - please have @jhurliman do a review of the logic as well.

Left a note about how we should aim to land this to minimize chances of revert or negative impact.

packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
snosenzo: Consolidated it into pretty much a single useEffect hook except for checking to see that the cursor is up to the current time before requesting an animation frame.

packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
```diff
@@ -742,7 +837,14 @@ export function ThreeDeeRender({ context }: { context: PanelExtensionContext }):
 
   // Render a new frame if requested
   useEffect(() => {
-    if (renderer && renderRef.current.needsRender) {
+    const { cursorTimeReached } = allFramesCursorRef.current;
+    if (
```
defunctzombie, line 841: Please add a comment here describing a bit of the _why_ in this logic. Maybe extract the inner ternary to a variable? This condition is hard for me to read.

defunctzombie (approved):

jhurliman (approved):

PR #5276, opened 2023-02-01
Move Variables and Events into right sidebar

2metres:
**User-Facing Changes**
None

**Description**
Introduces Right Sidebar and moves Variables and Events into it.

Screen Shot 2023-02-01 at 12 06 39 pm


2metres:
Todo 
- [ ] Make right sidebar resizeable/collapsible
- [ ] Connect selected object to sidebar

packages/studio-base/src/components/TabPanel.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { PropsWithChildren } from "react";
+import { makeStyles } from "tss-react/mui";
+
+const useStyles = makeStyles()({
+  root: {
+    flex: "auto",
+  },
+});
+
+export function TabPanel(
```
jtbandes, line 14: This is an extremely confusing component name since we also have the `Tab` component, which is a Panel.

jtbandes: I also don't get what this component is for. Its main functionality seems to be wrapping children in a `<div>` and conditionally hiding them. Doesn't really sound like it does much tab-related stuff?

packages/studio-base/src/components/Sidebar/index.stories.tsx
```diff
@@ -72,16 +73,18 @@ function Story({
...
-        <Sidebar
-          items={ITEMS}
-          bottomItems={BOTTOM_ITEMS}
-          selectedKey={selectedKey}
-          onSelectKey={setSelectedKey}
-        >
-          Main content
-        </Sidebar>
-      </div>
+      <MockMessagePipelineProvider>
```
jtbandes, line 76: I haven't read through yet to see why this is added, but with no other context, it seems wrong that we would need the MessagePipeline context in a story that's just about the Sidebar component (which handles tabs & layout of other components)

packages/studio-base/src/components/SecondarySidebar.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+        <IconButton className={classes.iconButton} size="small" onClick={toggleCollapsed}>
+          <ArrowRightIcon fontSize="inherit" />
+        </IconButton>
+      </Stack>
+      <Divider />
+      {!collapsed && (
+        <>
+          <TabPanel value={activeTab} index={0}>
+            <VariablesSidebar />
```
jtbandes, line 102: Seems like maybe this component is due to be renamed given that it is not a sidebar in itself?

packages/studio-base/src/components/SecondarySidebar.tsx
```diff
@@ -0,0 +1,113 @@
...
+  iconButton: {
+    fontSize: 20,
+    borderRadius: 0,
+  },
+}));
+
+const selectPlayerSourceId = ({ playerState }: MessagePipelineContext) =>
+  playerState.urlState?.sourceId;
+
+export function SecondarySidebar({
```
jtbandes, line 57: Not sure about "secondary" naming... maybe "RightSidebar"? Are we planning to also have one of these on the left as well, eventually, or totally remove the left one?

packages/studio-base/src/components/SecondarySidebar.tsx
```diff
@@ -0,0 +1,113 @@
...
+}: {
+  collapsed: boolean;
+  toggleCollapsed: () => void;
+  activeTab: number;
+  setActiveTab: (newValue: number) => void;
+}): JSX.Element {
+  const { classes } = useStyles();
+  const { currentUser } = useCurrentUser();
+
+  const playerSourceId = useMessagePipeline(selectPlayerSourceId);
```
jtbandes, line 71: (related to my comment in Sidebar stories) - let's refactor this component so it handles layout only, and leave the decision on what content to show in the Workspace component, like we do for the left sidebar. It's confusing to have the two sidebars be so different, and it also clutters up the tests/stories with providers that seem irrelevant to the concept of a sidebar.

packages/studio-base/src/components/Sidebar/index.tsx
```diff
@@ -268,24 +324,76 @@ export default function Sidebar<K extends string>(props: SidebarProps<K>): JSX.E
...
+          <IconButton
+            title="Variables"
+            className={classes.iconButton}
+            size="small"
+            onClick={() => {
+              setRightBarShown((old) => !old);
+              setActiveRightTab(0);
+            }}
+          >
+            <BracesVariable20Filled />
```
jtbandes, line 380: From a usability perspective, I find it confusing that icons are shown on the right, but when I click one, it disappears and is replaced with text.

packages/studio-base/src/components/Sidebar/index.tsx
```diff
@@ -124,32 +162,50 @@ export default function Sidebar<K extends string>(props: SidebarProps<K>): JSX.E
...
-          second: "children",
+          first: "leftbar",
+          second: {
+            direction: "row",
+            first: "children",
+            second: "rightbar",
+            splitPercentage: 80,
+          },
           splitPercentage:
             // keep previous splitPercentage when changing from one tab to another
```
jtbandes, line 180: I would think we probably want similar functionality for the sidebar on the right, right? I can't actually test this out easily because the preview deployment doesn't let me log in...

packages/studio-base/src/components/Sidebar/index.tsx
```diff
@@ -124,32 +162,50 @@ export default function Sidebar<K extends string>(props: SidebarProps<K>): JSX.E
...
+        };
+      });
+    } else if (leftBarShown && !rightBarShown) {
+      setMosaicValue((oldValue) => {
+        return {
+          direction: "row",
+          first: "leftbar",
+          second: "children",
+          splitPercentage:
+            // keep previous splitPercentage when hiding the right bar
```
jtbandes, line 193: not sure what this comment means but it does appear that the left bar changes width when I hide/show the right bar, so I think something is broken

jtbandes:
This PR's "user-facing changes" should take into account that this is all behind a feature flag. I would write "None (feature flagged)" for user-facing changes.

packages/studio-base/src/components/Sidebar/index.tsx
foxymiles: We only have one "tab" on the right side right now.

jtbandes: Ah I saw two in the screenshot, thought Events would be there too? I guess I'll pull and test locally.

foxymiles: There is an issue though with toggling the right bar changing the width of the left. Working on it.

jtbandes: There are two tabs when you have a DP source loaded. However, it does seem to somehow remember the splitPercentage of the right sidebar...not sure how that's working 🤷   And yeah I did see the issue you mentioned with the right sidebar changing the left width and vice versa.

image

foxymiles: I think the term "tab" here is becoming a bit confusing here. The splitPercentage logic for the left side was to keep the sidebar width the same when changing between different items on the vertical sidebar. These are internal to the right sidebar.

In any case the current logic isn't right so I'll work on it.

jtbandes (changes requested):
Various UX and architecture issues raised in prior comments

amacneil:
Getting this crash in 3d panel, not sure if related to this PR
image

amacneil:
Closing and opening the left sidebar takes it to take up 80%

foxymiles:
> Closing and opening the left sidebar takes it to take up 80%

Good catch. Should be fixed now.

defunctzombie:
There's a good number of eyeballs on this from technical and product so I will remove myself as a reviewer of this. I'll definitely take a poke at it once it has merged and I can enable the feature flag to see it.

packages/studio-base/src/components/TabPanel.tsx
2metres: This is largely based on [examples from MUI](https://mui.com/material-ui/react-tabs/#basic-tabs), they even export a version in `mui/lab`

packages/studio-base/src/components/SecondarySidebar.tsx
2metres: @foxymiles can you address this request?

foxymiles: The `Sidebar` component now would be more accurately called something like `SidebarsLayout` since it now contains both left & right sidebars and the content of the layout itself. Not sure if we want to expand the scope of the PR to make that change.

foxymiles: I've lifted the message pipeline stuff up into the workspace.

packages/studio-base/src/components/Sidebar/index.tsx
2metres: I feel like this pattern is fairly inuitive. 

When in the collapsed state they are minimized they are icons when epanded they are tabs. You can jump to the expanded item by clicking the icon.

2metres:
> Getting this crash in 3d panel, not sure if related to this PR image

I am seeing this crash pretty regularly on localhost: 

```
Error
Unsupported color mode gradient
Call Stack
 paletteColorCached
  packages/studio-base/src/panels/ThreeDeeRender/renderables/OccupancyGrids.ts:415:15
 updateTexture
  packages/studio-base/src/panels/ThreeDeeRender/renderables/OccupancyGrids.ts:305:13
 OccupancyGrids._updateOccupancyGridRenderable
  packages/studio-base/src/panels/ThreeDeeRender/renderables/OccupancyGrids.ts:227:9
 OccupancyGrids.handleOccupancyGrid [as handler]
  packages/studio-base/src/panels/ThreeDeeRender/renderables/OccupancyGrids.ts:128:18
 handleMessage
  packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts:1038:26
 Renderer.addMessageEvent
  packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts:836:9
 eval
  packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx:562:22
 invokePassiveEffectCreate
  node_modules/react-dom/cjs/react-dom.development.js:23484:20
 HTMLUnknownElement.callCallback
  node_modules/react-dom/cjs/react-dom.development.js:3947:14
 Object.invokeGuardedCallbackDev
  node_modules/react-dom/cjs/react-dom.development.js:3996:16
```

foxymiles:
> I am seeing this crash pretty regularly on localhost:

It's hard to see how the changes in this PR might have triggered that. Do you see it on main?

2metres:
Yeah, fair seeing something similar on Production

packages/studio-base/src/components/Sidebar/index.tsx
2metres: Following @jtbandes and I's discussion opting for toggle in the Top Nav to open close the sideflaps

packages/studio-base/src/components/TabPanel.tsx
2metres: I have renamed TabContent to try and clarify that it's not related to Foxglove "Panels"

jtbandes (approved):
Basically a self-approval at this point
image

PR #5279, opened 2023-02-01
Remove MockPanelContextProvider from stories which do not need it

defunctzombie:
**User-Facing Changes**
None

**Description**
Over-use of this provider hid a bug with PanelLayout which caused crashes in real-world use.


snosenzo (approved):
is there a way to detect multiple of the same context provider so this doesn't happen again?

defunctzombie:
> is there a way to detect multiple of the same context provider so this doesn't happen again?

Fascinating. Since the `Panel` HOC provides the PanelContext itself, the HOC component could `useContext(...)` And see if a context is already defined. If it is - fail. I think that would work and we could do it in dev only.

defunctzombie:
Tragically the answer is no: https://github.com/foxglove/studio/blob/main/packages/studio-base/src/components/Panel.tsx#L213

We already have a valid use case for having parent PanelContext.

Issue #5262, opened 2023-01-30
Plots titles resets when saving layout


iftahnaf:
**Description**
When saving shared layout, the plot titles resets to default (`Plot`).

- Version: 1.39.0
- Platform: Ubuntu 22.04

**Steps To Reproduce**
1. Set `Title` to the `Plot` panel.
2. Save the organization layout.
3. The `Title` resets.

**Expected Behavior**
The title should not be reset after saving the layout.

snosenzo:
should be fixed in: https://github.com/foxglove/studio/pull/5282

PR #5282, opened 2023-02-02
Don't override existing custom plot title in migration

foxymiles:
**User-Facing Changes**
Fixes an issue with custom plot panel titles in shared layouts.

**Description**
Fix the migration logic for the plot panel title to check for an existing custom tile before overwriting it with the legacy title.

### DIFF
--- a/packages/studio-base/src/panels/Plot/index.tsx
+++ b/packages/studio-base/src/panels/Plot/index.tsx
@@ -231,10 +231,11 @@ function Plot(props: Props) {
     xAxisVal,
     xAxisPath,
     sidebarDimension = config.sidebarWidth ?? defaultSidebarDimension,
+    [PANEL_TITLE_CONFIG_KEY]: customTitle,
   } = config;
 
   useEffect(() => {
-    if (legacyTitle) {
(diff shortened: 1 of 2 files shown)


defunctzombie (approved):

snosenzo:
This should fix: https://github.com/foxglove/studio/issues/5262

packages/studio-base/src/panels/Plot/index.tsx
```diff
@@ -231,18 +231,19 @@ function Plot(props: Props) {
...
+    [PANEL_TITLE_CONFIG_KEY]: customTitle,
   } = config;
 
   useEffect(() => {
-    if (legacyTitle) {
+    if (legacyTitle && (customTitle == undefined || customTitle === "")) {
       // Migrate legacy Plot-specific title setting to new global title setting
       // https://github.com/foxglove/studio/pull/5225
       saveConfig({
         title: undefined,
```
jtbandes, line 242: This shouldn't really have been necessary since the `title` is being cleared at the same time as the migration... 🤷 oh well

PR #5283, opened 2023-02-02
Remove duplicate layout types from ConsoleApi

defunctzombie:
**User-Facing Changes**
None

**Description**
The LayoutID and ISO8601Timestamp were duplicated in ConsoleApi.ts. This change updates ConsoleApi to use the eixsting types from ILayoutStorage.

### DIFF
--- a/packages/studio-base/src/services/ConsoleApi.ts
+++ b/packages/studio-base/src/services/ConsoleApi.ts
@@ -5,7 +5,7 @@
 import * as base64 from "@protobufjs/base64";
 
 import { add, fromNanoSec, Time, toRFC3339String, toSec } from "@foxglove/rostime";
-import { LayoutID, ISO8601Timestamp } from "@foxglove/studio-base";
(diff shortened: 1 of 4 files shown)


packages/studio-base/src/services/ConsoleApi.ts
```diff
@@ -5,6 +5,7 @@
 import * as base64 from "@protobufjs/base64";
 
 import { add, fromNanoSec, Time, toRFC3339String, toSec } from "@foxglove/rostime";
+import { LayoutID, ISO8601Timestamp } from "@foxglove/studio-base";
```
snosenzo, line 8: shouldn't this be coming from `@foxglove/studio-base/services/IlayoutStorage`? Shouldn't they all be coming from the same place if they're the same?

snosenzo (approved):

packages/studio-base/src/services/ConsoleApi.ts
defunctzombie: `studio-base` re-exports those types so these are the same types. But I will update this import to be more clear.

PR #5285, opened 2023-02-02
Add DataSourceEvent type to EventContext

defunctzombie:
**User-Facing Changes**
None

**Description**
The EventContext file should define the event type. ConsoleApi can then produce results of that type.

### DIFF
--- a/packages/studio-base/src/context/EventsContext.ts
+++ b/packages/studio-base/src/context/EventsContext.ts
@@ -7,15 +7,32 @@ import { AsyncState } from "react-use/lib/useAsyncFn";
 import { DeepReadonly } from "ts-essentials";
 import { StoreApi, useStore } from "zustand";
 
+import { Time } from "@foxglove/rostime";
 import useGuaranteedContext from "@foxglove/studio-base/hooks/useGuaranteedContext";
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/context/EventsContext.ts
```diff
@@ -7,15 +7,32 @@ import { AsyncState } from "react-use/lib/useAsyncFn";
...
 import { StoreApi, useStore } from "zustand";
 
+import { Time } from "@foxglove/rostime";
 import useGuaranteedContext from "@foxglove/studio-base/hooks/useGuaranteedContext";
-import { ConsoleEvent } from "@foxglove/studio-base/services/ConsoleApi";
+
+/**
+ * A single event
+ */
+export type Event = {
```
snosenzo, line 16: I think part of naming it `ConsoleEvent` previously was to make sure it didn't collide with other `Event` types. if we can keep some specificity about this I think that would be helpful

foxymiles: Right. `Event` is already in the JS namespace.

defunctzombie:
Suggestion from the slack: `DataSourceEvent`

foxymiles (approved):

PR #5287, opened 2023-02-02
Make new plot legend & settings the default

foxymiles:
**User-Facing Changes**
Replace the old plot legend with the new legend and settings driven configuration.

**Description**
Removes the old plot legend and replaces it with the new legend and moves plot configuration into the settings sidebar.


defunctzombie (approved):
Do we have an easy path to having the legend container size itself down to the text?

image

Looking at this component it feels like we can't be that far off from having the background adjust to the size of the legend rather than taking up all this space.

This LGTM otherwise.

defunctzombie:
:shipit:

amacneil (approved):
Seems to work great

PR #5289, opened 2023-02-03
Split single WorkerIterableSourceWorker into separate workers

defunctzombie:
**User-Facing Changes**
None

**Description**
Rather than having a single worker implementation which supports all the different types of data sources, this change splits into separate worker implementations for each data source.

Since each worker needs to property wrap and expose the source functions to the main thread. For this the WorkerIterableSourceWorker is re-purposed to provide this wrapper for each worker implementation to use to wrap their source implementations.

### DIFF
--- a/packages/studio-base/src/dataSources/RemoteDataSourceFactory.ts
+++ b/packages/studio-base/src/dataSources/RemoteDataSourceFactory.ts
@@ -23,7 +23,7 @@ const initWorkers: Record<string, () => Worker> = {
   ".mcap": () => {
     return new Worker(
       new URL(
-        "@foxglove/studio-base/players/IterablePlayer/mcap/McapIterableSourceWorker.worker",
+        "@foxglove/studio-base/players/IterablePlayer/Mcap/McapIterableSourceWorker.worker",
         import.meta.url,
       ),
     );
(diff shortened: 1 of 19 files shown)


packages/studio-base/src/dataSources/RemoteDataSourceFactory.ts
```diff
@@ -11,6 +11,25 @@ import {
 import { IterablePlayer, WorkerIterableSource } from "@foxglove/studio-base/players/IterablePlayer";
 import { Player } from "@foxglove/studio-base/players/types";
 
+const initWorkers: Record<string, () => Worker> = {
+  ".bag": () => {
+    return new Worker(
+      new URL(
+        "@foxglove/studio-base/players/IterablePlayer/BagIterableSourceWorker.worker",
```
snosenzo, line 18: I feel like it might be worth it to have these strings centralized in a singular record by-source that all of these can use? or even just having all of the initialization callbacks themselves centralized somewhere

snosenzo (approved):
looks good to me for the most part, but I'm less familiar with these worker changes or it's implications, might want to get another pair of eyes on it

packages/studio-base/src/dataSources/RemoteDataSourceFactory.ts
defunctzombie: > I feel like it might be worth it to have these strings centralized in a singular record by-source that all of these can use? or even just having all of the initialization callbacks themselves centralized somewhere

The purpose of the PR is actually to move away from that :D

PR #5290, opened 2023-02-03
TF Preloading benchmark

snosenzo:
**User-Facing Changes**

 - n/a
**Description**
- updated benchmarkplayer to seek backwards from end (keeping this as an additional test for benchmark player)
- added tf preloading benchmark layout
- Created synthetic player for transform preloading that benchmarks forward and backwards seeking

Output looks like:
```
Number of messages: 150000
Seek frame average times from beginning to end of playtime. Should remain generally constant:
 21.73ms, 13.79ms, 14.73ms, 12.66ms, 12.78ms, 15.71ms, 13.35ms, 13.22ms, 13.20ms, 13.09ms
 Seek frame average times from end to beginning of playtime. Should start high and decrease:
 71.03ms, 68.06ms, 59.42ms, 53.88ms, 40.60ms, 40.02ms, 28.62ms, 21.97ms, 16.70ms, 11.34ms
```
I'm using such a large number of messages to trigger the transform tree capacity at 50000 messages for each frame. 
Because I'm testing with a large number of message events there is major garbage collection happening during that benchmark which can add variation to results. 

FG-1495


jhurliman:
The timings suggest intermittent stalls. How frequent are the stalls and how bad compared to some sort of baseline?

snosenzo:
Yeah I'm not sure those stats were that representative from a single run so I ran it across multiple tries and am just calculating the average seek times to given parts of a file. 
This is what I got from a larger file with 250000 transform messages
```
Seek frame times (from end to beginning of playtime): 116.37ms, 94.85ms, 84.14ms, 70.51ms, 66.69ms, 52.40ms, 42.00ms, 30.48ms, 23.12ms, 11.11ms
```

defunctzombie:
@snosenzo lets change this to a player that produces transforms rather than requiring a file to load? You would then add an entry here: https://github.com/foxglove/studio/blob/main/benchmark/benchmarks.txt so that we are able to run these benchmarks and track how the performance profile changes over time. This is more robust than having 1-off file tests that we are not tracking.

defunctzombie (approved):

PR #5292, opened 2023-02-03
Adding support for YUYV(YUV2) aside from YUV422(UYVY) along with UYVY alias

FordoA455:
**User-Facing Changes**

**Description**
This commit is to add support for YUYV(YUV2) encoding aside from YUV422 which is encoded in the UYVY format. Our cameras take pictures in the YUYV format and this commit is to make it directly compatible with our images without having to do conversions with the data. The name might need to be changed in the future as YUV422 is supposed to be YUYV, not UYUV, but we did not want to mess up other people's programs through this change.

### DIFF
--- a/packages/studio-base/src/panels/Image/lib/renderImage.ts
+++ b/packages/studio-base/src/panels/Image/lib/renderImage.ts
@@ -14,6 +14,7 @@
 import {
   PinholeCameraModel,
   decodeYUV,
+  decodeYUYV,
   decodeRGB8,
   decodeRGBA8,
   decodeBGRA8,
@@ -137,6 +138,10 @@ function decodeMessageToBitmap(
       switch (encoding) {
         case "yuv422":
           decodeYUV(rawData as unknown as Int8Array, width, height, image.data);
+          break;
+        // change name in the future
+        case "yuyv":
+          decodeYUYV(rawData as unknown as Int8Array, width, height, image.data);
           break;
         case "rgb8":
           decodeRGB8(rawData, width, height, image.data);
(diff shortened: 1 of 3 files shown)


jkuszmaul:
@jhurliman FYI this change is from one of our students on the FRC team.

@FordoA455 , please add details to the main PR description (where you currently have "User-Facing Changes" and "Description").

packages/studio-base/src/panels/Image/lib/renderImage.ts
```diff
@@ -138,6 +139,10 @@ function decodeMessageToBitmap(
         case "yuv422":
           decodeYUV(rawData as unknown as Int8Array, width, height, image.data);
           break;
+        // change name in the future
```
jkuszmaul, line 142: The existing terminology being used for "yuv422" is dubious. E.g., the v4l2 documentation refers to the new format being added as YUV422 (https://www.kernel.org/doc/html/v4.8/media/uapi/v4l/pixfmt-yuyv.html) and the format being used in `main` is https://www.kernel.org/doc/html/v4.8/media/uapi/v4l/pixfmt-uyvy.html . @jhurliman do you have a preference? I don't want to break existing users, but it's a bit misleading...

jhurliman: They are both YUV 4:2:2 formats, just with different ordering. I think we need to keep “yuv422” mapping to uyvy, but we should also introduce a “uyvy” alias and this “yuyv” support.

jhurliman (approved):

PR #5295, opened 2023-02-06
Remove ConsoleApiContext from AppBar

defunctzombie:
**User-Facing Changes**
None

**Description**
AppBar detects signIn support via signIn being defined.

### DIFF
--- a/packages/studio-base/src/components/AppBar/index.tsx
+++ b/packages/studio-base/src/components/AppBar/index.tsx
@@ -7,7 +7,7 @@ import CloseIcon from "@mui/icons-material/Close";
 import FilterNoneIcon from "@mui/icons-material/FilterNone";
 import MinimizeIcon from "@mui/icons-material/Minimize";
 import { AppBar as MuiAppBar, Button, IconButton, Toolbar, Typography } from "@mui/material";
-import { MouseEvent, useCallback, useContext, useState } from "react";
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/components/AppBar/index.tsx
```diff
@@ -286,13 +284,11 @@ export function AppBar(props: AppBarProps): JSX.Element {
...
                       color="primary"
                       onClick={() => {
-                        if (signIn) {
-                          signIn();
-                          void analytics.logEvent(AppEvent.APP_BAR_CLICK_CTA, {
-                            user: "unauthenticated",
-                            cta: "sign-in",
-                          });
-                        }
+                        signIn();
```
jtbandes, line 287: How did you get away with removing the `if()` around this?

jtbandes (approved):

packages/studio-base/src/components/AppBar/index.tsx
defunctzombie: This UI is in a block after `{supportsAccountSettings &&` and typescript was smart enough to figure out the `const supportsAccountSettings = signIn != undefined && ` logic. I too was surprised.

defunctzombie:
@jtbandes The SignIn button is gone from the stories now with this change. Should I add it back to the stories or make a separate story with the button present so we know that works?

PR #5296, opened 2023-02-06
Implement focused path in settings API

foxymiles:
**User-Facing Changes**
Allows extension panel authors to focus and highlight a particular node in their settings UI.

**Description**
Adds a new focusedPath property to `SettingsTree`. When set the settings UI will expand all nodes up to and including the node at the path, scroll the node at the path into view, and flash the header in yellow.

The immediate need for this feature is improving ergonomics and discoverability for the new plot settings, https://github.com/foxglove/studio/pull/5287.

https://user-images.githubusercontent.com/93935560/216885715-ff9578fb-8b7c-44db-a2df-fd5a74d0284d.mov

### DIFF
--- a/packages/studio/index.d.ts
+++ b/packages/studio/index.d.ts
@@ -702,6 +702,35 @@ declare module "@foxglove/studio" {
   /**
    * A settings tree is a tree of panel settings that can be displayed and edited in
    * the panel settings sidebar.
+   *
+   * Nodes and fields in the tree can be referred to by a string path, which collects
+   * the keys of each node on the path from the root to the child node or field.
+   *
+   * For example, for the following tree:
+   *
+   *  root: {
+   *    children: {
+   *      a: {
+   *        children: {
+   *          b: {
+   *            fields: {
+   *              toggleMe: {
+   *                label: "Toggle me",
+   *                input: "boolean",
+   *                value: false,
+   *              },
+   *            },
+   *          },
+   *        },
+   *      },
+   *    },
+   *  },
+   *
+   * the path to the node at b would be ["a", "b"] and the path to the toggleMe
+   * field would be ["a", "b", "toggleMe"]. These paths are used in the
+   * actionHandler, which responds to updates to values in the tree, and also in
+   * the focusedPath, which is used to focus the editor UI at a particular node
+   * in the tree.
    */
   export type SettingsTree = {
     /**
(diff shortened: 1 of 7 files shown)


amacneil:
👍 works great.

Can we make the flash a single fade in/out and the same purple color as when variables update?

foxymiles:
> Can we make the flash a single fade in/out and the same purple color as when variables update?

Done

packages/studio/index.d.ts
```diff
@@ -714,6 +714,11 @@ declare module "@foxglove/studio" {
      */
     enableFilter?: boolean;
 
+    /**
+     * If set the editor will scroll to and highlight the node at this path.
```
defunctzombie, line 718: I'd find it helpful to call out here that the caller should un-set this before the next update? Or does the settings tree update ignore if the path is the same? Basically, looking for a bit more documentation on guidance for our extension authors on how to use this correctly.

packages/studio/index.d.ts
```diff
@@ -714,6 +714,11 @@ declare module "@foxglove/studio" {
...
+     */
+    focusedPath?: readonly string[];
```
defunctzombie, line 720: What's the thinking behind a string array here? Can you set more than one path? Or is a path defined as a string array? Maybe an example path could clarify things if we don't have one already described in the settings api. Remember that users of this are _not_ foxglove developers so they need guidance on how to build with this.

defunctzombie (approved):

packages/studio/index.d.ts
foxymiles: The `path` already exists as a concept in the API:

https://github.com/foxglove/studio/blob/ca36c94a3c07db0467d36085415c07520a50550f/packages/studio/index.d.ts#L687-L698

I can add a small example of a path if you think it would help illuminate both of these instances for users.

packages/studio/index.d.ts
foxymiles: It doesn't need to be unset, it's a fire & forget value and won't have any effect beyond the initial `useEffect` it triggers downstream. I'll expand on the comment to make that more clear.

jtbandes (approved):

packages/studio/index.d.ts
foxymiles: I've expanded the comments to explain what a path is and where it is used and also to include a small example of a settings tree and paths to elements of that tree.

packages/studio/index.d.ts
foxymiles: I've expanded this comment to make it clear it's not necessary to unset the focusedPath after setting it.

amacneil:
👏 this is awesome

PR #5304, opened 2023-02-08
Fix cursor jump in user scripts editor

### LINKED ISSUE #5303: User script panel resets cursor to start of editor
**Description**
After switching to a new script if I delete something, the cursor gets reset to the start of the script. See attached video for details.

- Version: `1.39.1 (darwin)`
- Platform: `MacOS Monterey 12.5.1` on Apple M1 Max

**Steps To Reproduce**

1. Create 2 new user scripts
2. Switch to a different script
3. Delete something
The cursor resets to the beginning of the file

**Expected Behavior**
Cursor stays where it is.

https://user-images.githubusercontent.com/25373098/217401006-4001c6f1-c70e-4b16-8c7a-c724f83046db.mov

foxymiles:
**User-Facing Changes**
Fixes an issue with the loss of cursor position in the user scripts editor during edits.

**Description**
The `onChange` handler transitively depended on the script content and was being reset during edits. Changing this to use a stable ref to the update function instead of depending directly on `setScriptCode` fixes this behavior.

Fixes #5303

### DIFF
--- a/packages/studio-base/src/panels/NodePlayground/Editor.tsx
+++ b/packages/studio-base/src/panels/NodePlayground/Editor.tsx
@@ -286,8 +286,9 @@ const Editor = ({
     });
   }, []);
 
+  // Refer to setScriptCode by reference so that the onChange callback isn't invalidated
+  // on every edit.
   const latestSetScriptCode = useLatest(setScriptCode);
-
   const onChange = React.useCallback(
     (srcCode: string) => latestSetScriptCode.current(srcCode),
     [latestSetScriptCode],
--- a/packages/studio-base/src/panels/NodePlayground/Editor.tsx
+++ b/packages/studio-base/src/panels/NodePlayground/Editor.tsx
@@ -19,6 +19,7 @@ import * as path from "path";
 import { ReactElement, useCallback, useRef } from "react";
 import MonacoEditor, { EditorDidMount, EditorWillMount } from "react-monaco-editor";
 import { useResizeDetector } from "react-resize-detector";
+import { useLatest } from "react-use";
 
 import getPrettifiedCode from "@foxglove/studio-base/panels/NodePlayground/getPrettifiedCode";
 import { Script } from "@foxglove/studio-base/panels/NodePlayground/script";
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/panels/NodePlayground/Editor.tsx
```diff
@@ -285,7 +286,13 @@ const Editor = ({
     });
   }, []);
 
-  const onChange = React.useCallback((srcCode: string) => setScriptCode(srcCode), [setScriptCode]);
+  // Refer to setScriptCode by reference so that the onChange callback isn't invalidated
+  // on every edit.
+  const latestSetScriptCode = useLatest(setScriptCode);
+  const onChange = React.useCallback(
```
defunctzombie, line 292:
```suggestion
  const onChange = useCallback(
```

defunctzombie (approved):
Does it even matter if onChange is invalidated on every edit? React makes me sad that action handling is so hard. `onChange` as a handler only needs initializing once but react makes this annoyingly hard.

foxymiles:
> Does it even matter if onChange is invalidated on every edit? React makes me sad that action handling is so hard. `onChange` as a handler only needs initializing once but react makes this annoyingly hard.

In general I don't think it's a problem but the Monaco editor is a very complex component and doesn't seem to handle this well.

foxymiles:
Probably a better way to avoid this kind of thing in general is to avoid having a callback depend on the value it's updating and to use the functional form of `setState`:

https://github.com/foxglove/studio/blob/4303d40e02ada0d4d1b409c463df0cdd3d5c4573/packages/studio-base/src/panels/NodePlayground/index.tsx#L309-L321

More like:

```typescript
  const setScriptCode = React.useCallback((code: string) => {
    // update code at top of backstack
    setScriptBackStack((oldStack) => {
      const backStack = [...oldStack];
      if (backStack.length > 0) {
        const script = backStack.pop();
        if (script && !script.readOnly) {
          return [...backStack, { ...script, code }];
        }
      }
      return oldStack;
    });
  }, []);
```

PR #5307, opened 2023-02-08
Replace storybook-patch with Story.play

defunctzombie:
**User-Facing Changes**
None

**Description**
Rather than patching storybook to support a ready signal via the react rendering callback, this change leverages the `play` functionality to wait on the signal and delay the story ready event.


defunctzombie:
Somehow even tho I removed the patch and have not yet updated all the other stories that useReadyState to have a `.play` function, there are no changes to the UI stories?!

packages/studio-base/src/components/TimeBasedChart/index.stories.tsx
```diff
@@ -212,7 +212,11 @@ export function CleansUpTooltipOnUnmount(_args: unknown): JSX.Element | ReactNul
       </MockMessagePipelineProvider>
     </div>
   );
-}
+};
+CleansUpTooltipOnUnmount.play = async (ctx) => {
+  await ctx.parameters.storyReady;
```
jtbandes, line 217: Would it make sense to clean up by creating the storyReady signal explicitly rather than having the useReadySignal boolean? Or do we still need that?

defunctzombie: I wanted to keep a smaller diff. I do think having the boolean parameter to make the signal is still nice as it ultimately provides something for the react hook in the Story to use and the play function to await on. There's not an easier way to share "stuff" between the story and the play function.

jtbandes (approved):
Seems fine for now (if it works).

PR #5310, opened 2023-02-08
Construct mesh_resource URLs relative to a base URL

jhurliman:
**User-Facing Changes**
- Fixes relative paths in URDFs loaded from HTTP(S)

**Description**
`mesh_resource` URLs previously used the filename specified in URDFs directly, such as `"../meshes/robot.stl"` which would resolve relative to the loaded webpage (i.e. https://studio.foxglove.dev/). Now, the relative path is combined with the base URL of the URDF to construct the `mesh_resource` URL.

Fixes FG-1544

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
@@ -587,7 +587,8 @@ export class Urdfs extends SceneExtension<UrdfRenderable> {
     renderable.removeChildren();
 
     const createChild = (frameId: string, i: number, visual: UrdfVisual): void => {
-      const childRenderable = createRenderable(visual, robot, i, frameId, renderer);
+      const baseUrl = renderable.userData.url;
+      const childRenderable = createRenderable(visual, robot, i, frameId, renderer, baseUrl);
       // Set the childRenderable settingsPath so errors route to the correct place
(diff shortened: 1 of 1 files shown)


defunctzombie (approved):
👍 It might be helpful to note for our future selves that `new URL("foo.ext", "https://example.com/file.urdf").toString();` produces `http://example.com/foo.ext`. I did a double take because it wasn't clear to me that `baseUrl` which would actually be the path to the urdf file rather than only a baseUrl would work as intended.

packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
```diff
@@ -773,7 +776,7 @@ function createMeshMarker(
     points: [],
     colors: [],
     text: "",
-    mesh_resource: mesh.filename,
+    mesh_resource: new URL(mesh.filename, baseUrl).toString(),
```
jtbandes, line 779: Does this correctly ignore the base URL if the mesh.filename is an absolute URL?

jtbandes (approved):
What thoughts do you have on how we might be able to add regression tests for this behavior?

packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
defunctzombie: image

Seems to

PR #5311, opened 2023-02-08
Improve consistency of renderable disposing

jhurliman:
**User-Facing Changes**
- None

**Description**
Consistently call `super.dispose()` from renderables, and do a more exhaustive disposing of children from `Renderable.dispose()` to ensure the default behavior of renderables is to clean up after themselves. This probably won't fix anything in the current codebase, but will set up future renderables to not leak memory by default.

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/dispose.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/dispose.ts
@@ -4,8 +4,6 @@
 
 import * as THREE from "three";
 
-import { Renderable } from "./Renderable";
-
 export function disposeMaterial(material: THREE.Material): void {
   if (material instanceof THREE.MeshStandardMaterial) {
     material.map?.dispose();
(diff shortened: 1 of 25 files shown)


packages/studio-base/src/panels/ThreeDeeRender/dispose.ts
```diff
@@ -22,18 +22,25 @@ export function disposeMaterial(material: THREE.Material): void {
 }
 
 export function disposeMeshesRecursive(object: THREE.Object3D): void {
-  object.traverse((child) => {
-    if (child instanceof THREE.Mesh) {
-      child.geometry.dispose();
-      if (Array.isArray(child.material)) {
-        for (const material of child.material) {
+  const disposeAny = (obj: THREE.Object3D) => {
```
snosenzo, line 25: do we have tests for this? would it be easy to add some?
also should this account for textures as well?

snosenzo (approved):

packages/studio-base/src/panels/ThreeDeeRender/dispose.ts
jhurliman: > do we have tests for this?

I couldn't say, I didn't write this method. But I assume no, because:

> would it be easy to add some?

I don't know how to ask three.js if something has been disposed or not.

jhurliman: > also should this account for textures as well?

I don't think meshes own textures? They can own one or more materials, which are disposed by this method.

PR #5312, opened 2023-02-08
Point unsupported browser banner at discussions

wkalt:
Previously this pointed at a closed github issue that had been moved to this discussion.

### DIFF
--- a/web/src/VersionBanner.tsx
+++ b/web/src/VersionBanner.tsx
@@ -85,9 +85,9 @@ const VersionBanner = function ({
 
             {!isChrome && (
               <Typography align="center" variant="subtitle1">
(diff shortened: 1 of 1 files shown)


defunctzombie:
I think we should be more decisive here and make a docs page on the website describing what we use that is chrome specific (or new browser specific) and that we do not support other browsers at this time. I don't think there's anything a community discussion is gonna do for folks here. We know we don't support the other browsers in all cases and should document those. The link in this banner should point to those documented limitations.

jtbandes:
Sending to @esthersweon to figure out the desired website updates and banner wording.

defunctzombie:
Updated the PR here: https://github.com/foxglove/studio/pull/5708

PR #5315, opened 2023-02-09
Leave label of new plot series blank by default

foxymiles:
**User-Facing Changes**
Leave the `Label` of new plot panel series blank by default.

**Description**
When adding a new series, leave the label blank instead of giving it a default like `Series 1`.


jtbandes (approved):
Thanks!

PR #5317, opened 2023-02-10
Enable help button on app sidebar

foxymiles:
**User-Facing Changes**
Re-enable the help button on the app sidebar.

**Description**
The help button on the app sidebar should be enabled whether or not the new app bar is enabled.

### DIFF
--- a/packages/studio-base/src/components/Sidebar/index.stories.tsx
+++ b/packages/studio-base/src/components/Sidebar/index.stories.tsx
@@ -2,6 +2,8 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
+import { screen } from "@testing-library/dom";
+import userEvent from "@testing-library/user-event";
 import { useEffect, useState } from "react";
 import { DndProvider } from "react-dnd";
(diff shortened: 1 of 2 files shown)


2metres (approved):

jtbandes (approved):

defunctzombie:
> The help button on the app sidebar should be enabled whether or not the new app bar is enabled.

Why's that? If we have a help button in the new app bar why do we need one in the app sidebar?

foxymiles:
@defunctzombie We only show the new app bar *or* the sidebar, not both. So the help button needs to work in both.

defunctzombie:
@foxymiles This is not true. When I enable the new navigation experimental feature the sidebar is still shown.

foxymiles:
> @foxymiles This is not true. When I enable the new navigation experimental feature the sidebar is still shown.

Yes you're right. My mistake.

I guess an alternative fix would be to hide the help button in the sidebar if the topnav is enabled? The old code showed the button but disabled the menu.

defunctzombie:
> I guess an alternative fix would be to hide the help button in the sidebar if the topnav is enabled?

👍 Hide seems sensible to me since we don't need _two_ help icons visible on the screen at once.

PR #5319, opened 2023-02-10
Foxglove websocket player: Set ROS profile for tf frame normalization

achim-k:
**User-Facing Changes**
- Foxglove websocket player: Set ROS profile for tf frame normalization

**Description**
From FG-1853: 

> When the foxglove websocket connection is established to ros foxglove bridge we need to detect the ROS distro and set the ROS profile (in the player) accordingly. This is important as it triggers some downstream panels to alter their behavior for handling certain types of messages (i.e. TF frames in ROS1 with leading / and without).

### DIFF
--- a/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
+++ b/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
@@ -77,6 +77,7 @@ export default class FoxgloveWebSocketPlayer implements Player {
   private _presence: PlayerPresence = PlayerPresence.NOT_PRESENT;
   private _problems = new PlayerProblemManager();
   private _numTimeSeeks = 0;
+  private _profile?: string;
 
   /** Earliest time seen */
   private _startTime?: Time;
@@ -143,6 +144,7 @@ export default class FoxgloveWebSocketPlayer implements Player {
       this._channelsById.clear();
       this._channelsByTopic.clear();
       this._setupPublishers();
+      this._profile = undefined;
     });
 
     this._client.on("error", (err) => {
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
```diff
@@ -199,9 +201,12 @@ export default class FoxgloveWebSocketPlayer implements Player {
...
 
       if (event.metadata != undefined && "ROS_DISTRO" in event.metadata) {
-        // Add common ROS message definitions
         const rosDistro = event.metadata["ROS_DISTRO"] as string;
-        const rosDataTypes = ["melodic", "noetic"].includes(rosDistro)
+        const isRos1 = ["melodic", "noetic"].includes(rosDistro);
+        this._profile = isRos1 ? "ros1" : "ros2";
+
+        // Add common ROS message definitions
+        const rosDataTypes = isRos1
```
jhurliman, line 209: What happens in the case that the websocket server is not a ROS system at all?

achim-k: Then `ROS_DISTRO` won't be set and the if-condition above `false` -> `this._profile = undefined`

defunctzombie: That's not true. If there is no metadata for ROS_DISTRO then `isRos1` will evaluate to false and the profile will be set to `ros2` (based on my reading).

Also the `as string` casts need to be removed. If those were removed I bet typescript would have given a better indicator that something is wrong here.

defunctzombie: I originally overlooked this as well but the `&& "ROS_DISTRO"` is the important bit of the _if_ statement that makes this code only run when ROS_DISTRO is set. This was not obvious to me.

Another way maybe to write this would be:

```typescript
// comment here about what we are trying to accomplish
const maybeRosDistro = event.metadata?.["ROS_DISTRO"]`;
if (maybeRosDistro) {...}

File: packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts [hunk 2]
```diff
@@ -199,9 +201,12 @@ export default class FoxgloveWebSocketPlayer implements Player {
       this._datatypes = new Map();
 
       if (event.metadata != undefined && "ROS_DISTRO" in event.metadata) {
```
achim-k, line 203:
```suggestion
      const maybeRosDistro = event.metadata?.["ROS_DISTRO"]`;
      if (maybeRosDistro) {
```

packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
achim-k: changed it to Roman's proposal

PR #5324, opened 2023-02-12
Hide sidebar help button when top nav is enabled

foxymiles:
**User-Facing Changes**
Hide the sidebar help button if the experimental top navigation is enabled.

**Description**
There's a help button in the top nav bar so the sidebar help button is redundant if top nav is enabled.

### DIFF
--- a/packages/studio-base/src/components/Sidebar/index.stories.tsx
+++ b/packages/studio-base/src/components/Sidebar/index.stories.tsx
@@ -2,11 +2,12 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
-import { screen } from "@testing-library/dom";
-import userEvent from "@testing-library/user-event";
 import { useEffect, useState } from "react";
 import { DndProvider } from "react-dnd";
 import { HTML5Backend } from "react-dnd-html5-backend";
+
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/components/Sidebar/index.stories.tsx
```diff
@@ -35,13 +36,22 @@ const BOTTOM_ITEMS = new Map<string, SidebarItem>([
...
   clickKey?: string;
   defaultSelectedKey?: string | undefined;
+  enableAppBar?: boolean;
   height?: number;
 }) {
   const [selectedKey, setSelectedKey] = useState<string | undefined>(defaultSelectedKey);
+  const [_, setAppBarEnabled] = useAppConfigurationValue<boolean>(AppSetting.ENABLE_NEW_TOPNAV);
+
+  useEffect(() => {
+    if (enableAppBar === true) {
```
defunctzombie, line 51: Why is this setting the configuration value? I thought we had a way to setup the providers with the values we wanted for config?

defunctzombie (approved):

jtbandes (approved):

packages/studio-base/src/components/Sidebar/index.stories.tsx
foxymiles: We do but this requires less mocking and also tests functionality across a config change.

PR #5326, opened 2023-02-13
Use configured timezone in data source info start/end time

foxymiles:
**User-Facing Changes**
Fixes an issue with the start and end date displayed in the data source sidebar.

**Description**
Respects the configured timezone when calculating start and end times of a data source in the sidebar.

### DIFF
--- a/packages/studio-base/src/components/DataSourceInfoView.tsx
+++ b/packages/studio-base/src/components/DataSourceInfoView.tsx
@@ -7,13 +7,14 @@ import { MutableRefObject, useEffect, useRef } from "react";
 import { makeStyles } from "tss-react/mui";
 
 import { Time } from "@foxglove/rostime";
+import { AppSetting } from "@foxglove/studio-base/AppSetting";
 import {
   MessagePipelineContext,
   useMessagePipeline,
 } from "@foxglove/studio-base/components/MessagePipeline";
 import Stack from "@foxglove/studio-base/components/Stack";
 import Timestamp from "@foxglove/studio-base/components/Timestamp";
-import { useAppTimeFormat } from "@foxglove/studio-base/hooks";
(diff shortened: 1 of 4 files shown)


packages/studio-base/src/components/DataSourceInfoView.tsx
```diff
@@ -122,6 +124,7 @@ export function DataSourceInfoView(): JSX.Element {
   const durationRef = useRef<HTMLDivElement>(ReactNull);
   const endTimeRef = useRef<HTMLDivElement>(ReactNull);
   const { formatTime } = useAppTimeFormat();
+  const [timezone] = useAppConfigurationValue<string>(AppSetting.TIMEZONE);
```
snosenzo, line 127: instead of pulling in `timezone` like this can we add the `formatDate` to the return `useAppTimeFormat` and create it like `formatTime`? I feel like it makes sense to have `formatDate` be part of that hook as well

packages/studio-base/src/components/DataSourceInfoView.tsx
```diff
@@ -73,7 +75,7 @@ function DataSourceInfoContent(props: {
         {playerPresence === PlayerPresence.INITIALIZING ? (
           <Skeleton animation="wave" width="50%" />
         ) : startTime ? (
-          <Timestamp horizontal time={startTime} />
+          <Timestamp horizontal time={startTime} timezone={timezone} />
```
snosenzo, line 78: yeah if we made `formatDate` in `TimeStamp` also come from `useAppTimeFormat` we'd be able to remove this as an optional prop.

foxymiles: Makes sense. I've updated the PR to incorporate this and your other suggestion.

jtbandes (approved):

PR #5330, opened 2023-02-13
Foxglove websocket: Fix datatype updates not being communicated

achim-k:
**User-Facing Changes**
- Foxglove websocket: Fix datatype updates not being communicated

**Description**
This PR fixes an issue with messages of new channels not being visible in the Raw message panel. 

Fixes FG-1856

### DIFF
--- a/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
+++ b/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
@@ -630,7 +630,7 @@ export default class FoxgloveWebSocketPlayer implements Player {
         topics: _topics,
         // Always copy topic stats since message counts and timestamps are being updated
         topicStats: new Map(this._topicsStats),
-        datatypes: _datatypes,
+        datatypes: new Map(_datatypes),
         parameters: new Map(this._parameters),
         services: new Map(this._services),
       },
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
```diff
@@ -630,7 +630,7 @@ export default class FoxgloveWebSocketPlayer implements Player {
         topics: _topics,
         // Always copy topic stats since message counts and timestamps are being updated
         topicStats: new Map(this._topicsStats),
-        datatypes: _datatypes,
+        datatypes: new Map(_datatypes),
```
jtbandes, line 633: Seems like we wouldn't want to _always_ copy this, but only when datatypes actually changed.

jtbandes:
You can fix the CI failure by rebasing on #5329

packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
achim-k: > Seems like we wouldn't want to _always_ copy this, but only when datatypes actually changed.

Yes, that would be the best. Do you have an idea on how I could do that?

achim-k:
> You can fix the CI failure by rebasing on #5329

I think the test strategy is to test the merged branch (into main). So an empty commit should have also worked, not sure why vercel deployment still failed. Anyway, will rebase

packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
jtbandes: I would recommend making _datatypes a ReadonlyMap to prevent any mutations. Mutations can be done by producing a new map (e.g. using immer)

jhurliman (approved):

packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
achim-k: Making it a `ReadonlyMap` makes it incompatible with the PlayerInterface. What I did instead, was to leave it a normal Map, and only copy assign it after potential changes (rather than on every state update)

PR #5341, opened 2023-02-14
Clear timing events on component unmount to prevent memory leaks

Arooba-git:
Hello 👋 
As part of our project, we are using Facebook's new Memlab tool to detect memory leaks in SPA applications and their libraries. While running the tool and analyzing the code of foxglove's studio, we saw that your project does a very good job of ensuring that all async operations are cancelled when components destroy. However, as per Memlab execution results, we found some uncleared timers, causing the memory to leak (screenshots below).

[before]
before Screen Shot 2023-02-14 at 1 30 36 AM

Hence we added the fix by clearing the timers on component unload, and you can see the count of leaks reducing noticeably:
 <br />

after both Screen Shot 2023-02-14 at 1 52 54 AM

Note. There are still some uncleared timers in other files, but our fix for them was having typing conflicts [1] when defining timerID variables, (using one kind of typing would pass linting but fail the build and vice versa), so we would leave it to you to clear timers in those files.

We executed the unit test suite after the fixes in this PR and it successfully passed 120 of 120 total (in watch + non-watch mode); we also ensured all the changes passed the lint.

You can analyze this and other potential leak sources, if you like, by running [Memlab](https://facebook.github.io/memlab/) with a scenario file covering maximum count of use cases.

Following is a sample of the scenario file we used (it needs to be a .js file but attaching here in .txt form):
[foxglove-scenario-memlab.txt](https://github.com/foxglove/studio/files/10727459/foxglove-scenario-memlab.txt)

Note that some other reported leaks (in Memlab) originated from React's internal objects, and hence were ignored.

References:
[1] https://github.com/Microsoft/TypeScript/issues/30128

### DIFF
--- a/packages/studio-base/src/components/AutoSizingCanvas/index.stories.tsx
+++ b/packages/studio-base/src/components/AutoSizingCanvas/index.stories.tsx
@@ -28,7 +28,7 @@ function Example({
   const [width, setWidth] = useState(300);
   const [pixelRatio, setPixelRatio] = useState(devicePixelRatio);
   useEffect(() => {
-    setTimeout(() => {
+    const timeOutID = setTimeout(() => {
       if (changeSize) {
         setWidth(150);
       }
@@ -36,6 +36,10 @@ function Example({
         setPixelRatio(2);
       }
     }, 10);
+
+    return () => {
+      clearTimeout(timeOutID);
+    };
   }, [changePixelRatio, changeSize]);
 
   return (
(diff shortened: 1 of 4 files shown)


jtbandes (approved):
Thanks!

Most of these are in `.stories.tsx` files, which are used for screenshot testing, not in the production app, so I don't expect much of an impact in real-world usage. However, I agree that the changes are technically correct / good practice.

> Note. There are still some uncleared timers in other files, but our fix for them was having typing conflicts [1] when defining timerID variables, (using one kind of typing would pass linting but fail the build and vice versa), so we would leave it to you to clear timers in those files.

Feel free to prepare and submit a draft PR with these changes. We can help fix the typing issues.

PR #5343, opened 2023-02-14
Cancel animation frame requests on component unmount to prevent memory. leaks

Arooba-git:
Hello 👋 
As part of our project, we are using Facebook's new Memlab tool to detect memory leaks in SPA applications and their libraries. While running the tool and analyzing the code of foxglove's studio, we saw that your project does a very good job of ensuring that all async operations are cancelled when components destroy. However, as per Memlab execution results, we found some uncancelled animation frame requests, causing the memory to leak (screenshots below).

[before]
before Screen Shot 2023-02-14 at 1 30 36 AM

Hence we added the fix by cancelling the requests on component unload, and you can see the count of leaks reducing noticeably:
 <br />

after anim Screen Shot 2023-02-14 at 1 35 47 AM

We executed the unit test suite after the fixes in this PR and it successfully passed 120 of 120 total (in watch + non-watch mode); we also ensured all the changes passed the lint.

You can analyze this and other potential leak sources, if you like, by running [Memlab](https://facebook.github.io/memlab/) with a scenario file covering maximum count of use cases.

Following is a sample of the scenario file we used (it needs to be a .js file but attaching here in .txt form):
[foxglove-scenario-memlab.txt](https://github.com/foxglove/studio/files/10727459/foxglove-scenario-memlab.txt)

Note that some other reported leaks (in Memlab) originated from React's internal objects, and hence were ignored.

### DIFF
--- a/packages/studio-base/src/components/TimeBasedChart/index.tsx
+++ b/packages/studio-base/src/components/TimeBasedChart/index.tsx
@@ -141,6 +141,7 @@ export type Props = {
 // component. Uses chart.js internally, with a zoom/pan plugin, and with our
 // standard tooltips.
 export default function TimeBasedChart(props: Props): JSX.Element {
+  const requestID = useRef<number>(0);
   const {
     datasetId,
     type,
@@ -195,7 +196,7 @@ export default function TimeBasedChart(props: Props): JSX.Element {
 
     if (current) {
       // allow the chart offscreen canvas to render to screen before calling done
-      requestAnimationFrame(current);
+      requestID.current = requestAnimationFrame(current);
     }
   }, []);
 
@@ -203,6 +204,7 @@ export default function TimeBasedChart(props: Props): JSX.Element {
     // cleanup paused frames on unmount or dataset changes
     return () => {
       onFinishRender();
+      cancelAnimationFrame(requestID.current);
     };
   }, [pauseFrame, onFinishRender]);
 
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/components/TimeBasedChart/index.tsx
```diff
@@ -196,13 +197,15 @@ export default function TimeBasedChart(props: Props): JSX.Element {
     if (current) {
       // allow the chart offscreen canvas to render to screen before calling done
       requestAnimationFrame(current);
+      requestID.current = requestAnimationFrame(current);
```
jtbandes, line 200: This seems wrong, it's adding a 2nd call to rAF.

Arooba-git: Hmmm, can you elaborate on this a bit? 🤔

jtbandes:
```
      requestAnimationFrame(current);
      requestID.current = requestAnimationFrame(current);

```

There are now 2 calls to `requestAnimationFrame` when previously there was only 1.

Arooba-git: Oh right, this is a mistake, let me fix it up!

Arooba-git: fixed :)

jtbandes (approved):

jtbandes:
@defunctzombie please merge

PR #5344, opened 2023-02-14
Cap plot legend size at 80% of container

foxymiles:
**User-Facing Changes**
Limit the size of the plot legend to prevent it from completing displacing the plot.

**Description**
Cap plot legend size at 80% of container size to guarantee there is always some room to display the plot itself.


jtbandes (approved):
Is this fixing a bug we can cover with a screenshot test?

foxymiles:
@jtbandes Yeah I think I can cover this in a story. I'll add one.

PR #5345, opened 2023-02-14
add analytics for URDFAssetLoader

jtbandes:
**User-Facing Changes**
None

**Description**
Adds analytics event for loading a local .urdf file, and automatically loading one from the robot_description parameter.
Closes FG-1549
Relates to #4875

### DIFF
--- a/packages/studio-base/src/services/URDFAssetLoader.ts
+++ b/packages/studio-base/src/services/URDFAssetLoader.ts
@@ -30,12 +30,27 @@ export default class URDFAssetLoader implements AssetLoader {
       source: "param" | "local_file";
     },
   ): Promise<Asset | undefined> {
-    void options.analytics.logEvent(
-      options.source === "param" ? AppEvent.URDF_LOAD_LOCAL_FILE : AppEvent.URDF_LOAD_LOCAL_FILE,
-    );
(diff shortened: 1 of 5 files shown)


packages/studio-base/src/services/URDFAssetLoader.ts
```diff
@@ -21,7 +22,17 @@ const URDF_ROOT = "$URDF_ROOT";
...
-  public async load(file: File): Promise<Asset | undefined> {
+  public async load(
+    file: File,
+    options: {
+      basePath: string | undefined;
+      analytics: IAnalytics;
+      source: "param" | "local_file";
+    },
+  ): Promise<Asset | undefined> {
+    void options.analytics.logEvent(
```
defunctzombie, line 33: Shouldn't this happen after the regex test on the next line? Otherwise we are calling this for any file load?

defunctzombie: Maybe also worth tracking successful loads - since ones that don't work don't lead to anything meaningful for the user either.

defunctzombie (approved):
A little weird that `loadFile` has to accept the analytics as argument but I don't see a better way.

defunctzombie (approved):

PR #5347, opened 2023-02-14
Clear timers when components unmount to prevent memory leak

Arooba-git:
Disclaimer: This PR has not passed linting as it makes use of `any` type.

This is a followup of PR # [5341](https://github.com/foxglove/studio/pull/5341#pullrequestreview-1296725784)
That PR fixes uncleared timers in some files, while also passing test cases and linting checks. This PR is created on the suggestion of @jtbandes to clear timers in the remainder of the files (however linting issues need to be manually addressed in this one). 

Note. Typescript is currently logging the warning `Argument of type 'Timeout | undefined' is not assignable to parameter of type 'number | undefined'.` I believe this is because it is wrongly assuming the environment of Node when running Storybook in the browser. [1]

Reference: 
[1] https://github.com/Microsoft/TypeScript/issues/30128


jtbandes:
@defunctzombie please merge

jtbandes (approved):

PR #5348, opened 2023-02-14
3D: reuse renderables for models with embedded data

### LINKED ISSUE #5197: ModelPrimitive with embedded data flickers or doesn't appear even if the data is not changing
**Description**

ModelPrimitives with embedded data are re-loaded each frame even if the data has not changed. Since the loading is async this results in flickering, and if the playback speed is significantly faster than the loading speed, they don't appear at all until playback is paused:

https://user-images.githubusercontent.com/14237/214116498-be4cfb54-c263-4367-b006-15c237b316d0.mov

### Sample data

<details><summary>Script to create example data (file is too large to upload):</summary>

```py
import math
from os import path
import requests
from tqdm import tqdm
from pyquaternion import Quaternion
from foxglove_schemas_protobuf.SceneUpdate_pb2 import SceneUpdate
from foxglove_schemas_protobuf.FrameTransform_pb2 import FrameTransform

avocado_data = requests.get("https://assets.foxglove.dev/Avocado.glb").content

    writer = Writer(stream)

    frame_interval_ns = 500_000_000
    nframes = 10
    nshapes = 20
        t = frame * frame_interval_ns

        nshapes_visible = int(nshapes * (1 + math.sin(2 * math.pi * 4 * frame / nframes)) / 2)

        scene_update = SceneUpdate()
        entity = scene_update.entities.add()
        entity.timestamp.FromNanoseconds(t)
        entity.frame_id = "entity"
        entity.id = "mymodel"

        for i in range(3):
            axis = [1, 0, 0]
            q = Quaternion(axis=axis, angle=2 * math.pi * (i / nshapes + frame / nframes))
            model = entity.models.add()
            model.pose.position.x = 1 + i*0.5
            model.pose.position.y = math.sin(i * 2 * math.pi * frame / nframes)
            model.pose.orientation.x = q.x
            model.pose.orientation.y = q.y
            model.pose.orientation.z = q.z
            model.pose.orientation.w = q.w
            model.scale.x = 5
            model.scale.y = 5
            model.scale.z = 5
            model.data = avocado_data
            model.media_type = "model/gltf-binary"

        writer.write_message("scene update", scene_update, t)

        q = Quaternion(axis=[0, 0, 1], angle=2 * math.pi * t / 1_000_000_000 / 30)
        frame_transform = FrameTransform()
        frame_transform.timestamp.FromNanoseconds(t)
        frame_transform.parent_frame_id = "scene"
        frame_transform.child_frame_id = "entity"
        frame_transform.translation.x = 0
        frame_transform.translation.y = 0
        frame_transform.translation.z = 0
        frame_transform.rotation.x = q.x
        frame_transform.rotation.y = q.y
        frame_transform.rotation.z = q.z
        frame_transform.rotation.w = q.w
        writer.write_message("transforms", frame_transform, t)

    writer.finish()
    stream.close()
```

</details>

**Expected Behavior**
The models should not be re-loaded each frame

jtbandes:
**User-Facing Changes**
Improved 3D panel performance when `ModelPrimitives` contain embedded `data` that is equal between consecutive messages.

**Description**
Reuse `RenderableModel` for models with embedded `data`, if the data & media_type match, rather than always re-loading them.

`crc32` is used as a hash function, and then we compare to confirm equality before reusing the renderable.

The underlying loading is still async because some loaders require wasm (i.e. DracoLoader)

Fixes #5197 
Fixes FG-1593

Perf note: on my M1 Pro, it seems like an 8MB model (avocado) takes about 4ms to crc32, and 9ms to do a full equality check.

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableModels.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableModels.ts
@@ -4,6 +4,7 @@
 
 import * as THREE from "three";
 
+import { crc32 } from "@foxglove/crc";
 import { toNanoSec } from "@foxglove/rostime";
 import { ModelPrimitive, SceneEntity } from "@foxglove/schemas";
 import { emptyPose } from "@foxglove/studio-base/util/Pose";
@@ -27,9 +28,26 @@ type RenderableModel = {
   model: THREE.Group;
   /** Reference to the original model before modification so it can be re-cloned if necessary. */
   cachedModel: LoadedModel;
+  /** Reference to the original message for checking whether this renderable can be reused */
+  primitive: ModelPrimitive;
 };
 
+function byteArraysEqual(a: Uint8Array, b: Uint8Array): boolean {
+  if (a.length !== b.length) {
+    return false;
+  }
+  for (let i = 0; i < a.length; i++) {
+    if (a[i] !== b[i]) {
+      return false;
+    }
+  }
+  return true;
+}
+
 export class RenderableModels extends RenderablePrimitive {
+  /** Renderables loaded from embedded data */
+  private renderablesByDataCrc = new Map<number, RenderableModel[]>();
+  /** Renderables loaded from URLs */
   private renderablesByUrl = new Map<string, RenderableModel[]>();
   private updateCount = 0;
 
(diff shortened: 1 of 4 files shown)


packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableModels.ts
```diff
@@ -53,60 +111,77 @@ export class RenderableModels extends RenderablePrimitive {
...
+        if (originalUpdateCount !== this.updateCount) {
+          // another update has come in, bail before doing any mutations
+          return;
+        }
+        if (renderable) {
+          newRenderables.push(renderable);
+          this.add(renderable.model);
+
+          // Render a new frame now that the model is loaded
+          this.renderer.queueAnimationFrame();
```
snosenzo, line 183: why shouldn't this be in the `finally` after models are removed?

snosenzo (approved):
other than one question looks good to me

packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableModels.ts
jtbandes: Hm, it does seem like maybe it should happen there. But I think it's also nice to do it immediately when each model is loaded in case some of them take longer than others.

PR #5350, opened 2023-02-14
Fix foxglove.Grid default auto-select colorMode behavior

snosenzo:
**User-Facing Changes**

 - Fix foxglove.Grid default auto-select colorMode behavior to use separate RGBA-fields when available

**Description**
`autoSelectColorField` formerly did not account for separate RGBA fields. Now it should. 

Fixes FG-1718

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/pointClouds/colors.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/pointClouds/colors.ts
@@ -4,7 +4,6 @@
 
 import * as THREE from "three";
 
-import { PackedElementField } from "@foxglove/schemas";
 import { SettingsTreeFields, SettingsTreeNode, Topic } from "@foxglove/studio";
 import { BaseSettings } from "@foxglove/studio-base/panels/ThreeDeeRender/settings";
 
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/panels/ThreeDeeRender/renderables/pointClouds/colors.ts
```diff
@@ -205,6 +205,11 @@ export function autoSelectColorField<Settings extends ColorModeSettings>(
     }
   }
 
+  if (hasSeparateRgbaFields(fields.map((f) => f.name))) {
+    output.colorMode = "rgba-fields";
```
jtbandes, line 209: does this affect Grid only? We have some logic to disable the separate fields mode for ROS point clouds and only enable it for foxglove point clouds: https://github.com/foxglove/studio/blob/8b42c7631867be1879695aaa33730e75e67175fb/packages/studio-base/src/panels/ThreeDeeRender/renderables/pointExtensionUtils.ts#L91

snosenzo: It's only used by the grid right now. looks like point clouds use an identically named function in `pointExtensionUtils` 😆 -- should I consolidate them?

jtbandes: Up to you, I guess separate is fine if they have different behavior, but maybe consolidating would bring some more sanity?

jtbandes (approved):

PR #5351, opened 2023-02-14
Support schemaless channels

jtbandes:
**User-Facing Changes**
Added support for loading data from MCAP channels with no schema.

**Description**
Allow schema/schemaName to be undefined in several places. Remove some error messages around missing schemas.

Resolves FG-1441

### DIFF
--- a/packages/mcap-support/src/parseChannel.ts
+++ b/packages/mcap-support/src/parseChannel.ts
@@ -49,33 +49,31 @@ function parsedDefinitionsToDatatypes(
  */
 export function parseChannel(channel: Channel): ParsedChannel {
   if (channel.messageEncoding === "json") {
-    if (channel.schema?.encoding !== "jsonschema") {
+    if (channel.schema != undefined && channel.schema.encoding !== "jsonschema") {
       throw new Error(
-        `Message encoding ${channel.messageEncoding} with ${
-          channel.schema == undefined
-            ? "no encoding"
-            : `schema encoding '${channel.schema.encoding}'`
-        } is not supported (expected jsonschema)`,
+        `Message encoding ${channel.messageEncoding} with schema encoding '${channel.schema.encoding}' is not supported (expected jsonschema or no schema)`,
       );
     }
     const textDecoder = new TextDecoder();
-    const schema =
-      channel.schema.data.length > 0
-        ? JSON.parse(textDecoder.decode(channel.schema.data))
-        : undefined;
     let datatypes: RosDatatypes = new Map();
     let deserializer = (data: ArrayBufferView) => JSON.parse(textDecoder.decode(data));
-    if (schema != undefined) {
-      if (typeof schema !== "object") {
-        throw new Error(`Invalid schema, expected JSON object, got ${typeof schema}`);
+    if (channel.schema != undefined) {
+      const schema =
+        channel.schema.data.length > 0
+          ? JSON.parse(textDecoder.decode(channel.schema.data))
+          : undefined;
+      if (schema != undefined) {
+        if (typeof schema !== "object") {
+          throw new Error(`Invalid schema, expected JSON object, got ${typeof schema}`);
+        }
+        const { datatypes: parsedDatatypes, postprocessValue } = parseJsonSchema(
+          schema as Record<string, unknown>,
+          channel.schema.name,
+        );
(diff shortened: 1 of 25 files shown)


packages/mcap-support/src/parseChannel.ts
```diff
@@ -49,33 +49,31 @@ function parsedDefinitionsToDatatypes(
  */
 export function parseChannel(channel: Channel): ParsedChannel {
   if (channel.messageEncoding === "json") {
-    if (channel.schema?.encoding !== "jsonschema") {
+    if (channel.schema != undefined && channel.schema.encoding !== "jsonschema") {
```
jhurliman, line 52: I thought our eslint rules would enforce the `?.encoding` variant here. Was this change done by linting or manually?

packages/studio-base/src/players/IterablePlayer/Mcap/McapIndexedIterableSource.ts
```diff
@@ -191,7 +184,7 @@ export class McapIndexedIterableSource implements IIterableSource {
             publishTime: fromNanoSec(message.publishTime),
             message: channelInfo.parsedChannel.deserializer(message.data),
             sizeInBytes: message.data.byteLength,
-            schemaName: channelInfo.schemaName,
+            schemaName: channelInfo.schemaName ?? "",
```
jhurliman, line 187: :( maybe it should be `"-"` to match the UI 🤣

jhurliman (approved):

packages/mcap-support/src/parseChannel.ts
jtbandes: Manually but the linter didn't complain. I'm always a bit surprised about when that rule fires and when it doesn't 🤷

packages/studio-base/src/components/PanelExtensionAdapter/renderState.ts
```diff
@@ -171,8 +171,8 @@ function initRenderStateBuilder(): BuildRenderStateFn {
         const topics = sortedTopics.map<Topic>((topic) => {
           const newTopic: Topic = {
             name: topic.name,
-            datatype: topic.schemaName,
-            schemaName: topic.schemaName,
+            datatype: topic.schemaName ?? "",
+            schemaName: topic.schemaName ?? "",
```
snosenzo, line 175: what's the general rule for where this needs to be a string vs string | undefined?
Should they still not show externally as `string | undefined`?

snosenzo (approved):
just a question about when they shouldn't be able to be undefined

packages/studio-base/src/components/PanelExtensionAdapter/renderState.ts
jtbandes: @defunctzombie suggested that we should not change the public APIs in @foxglove/studio which means we can't change the type of MessageEvent.

packages/mcap-support/src/parseChannel.ts
```diff
@@ -49,33 +49,31 @@ function parsedDefinitionsToDatatypes(
...
       throw new Error(
-        `Message encoding ${channel.messageEncoding} with ${
-          channel.schema == undefined
-            ? "no encoding"
-            : `schema encoding '${channel.schema.encoding}'`
-        } is not supported (expected jsonschema)`,
+        `Message encoding ${channel.messageEncoding} with schema encoding '${channel.schema.encoding}' is not supported (expected jsonschema or no schema)`,
```
defunctzombie, line 54: Where does the check happen that schemaless uses a schemId of `0`? Or are we saying that users can still provide a schema record with no schema for json encoded messages? Technically not a valid mcap file (or maybe it is?) but we want to be flexible about it in studio?

defunctzombie (approved):

packages/mcap-support/src/parseChannel.ts
jtbandes: Does this answer your question? https://github.com/foxglove/studio/pull/5351/files/d42e95b3e69bcfa577ffd96b589c4e2b968bc6e0#diff-2a8df5991c9755a3c42c61399d45b12e2e04c804b97b780053696725b984f7d8R55

defunctzombie: Yep 👍

PR #5352, opened 2023-02-14
Make the signIn and signOut fields optional for CurrentUserContext

defunctzombie:
**User-Facing Changes**
None

**Description**
Whether these fields are present or not will indicate if any such functionality is available.

### DIFF
--- a/packages/studio-base/src/components/AppBar/User.tsx
+++ b/packages/studio-base/src/components/AppBar/User.tsx
@@ -62,6 +62,11 @@ export function UserMenu({
 
   const beginSignOut = useCallback(async () => {
     try {
+      // There are checks in the parent component that should prevent showing this to users if there
+      // is no user context available. We assert here to detect miss-use.
+      if (!signOut) {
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/components/AppBar/User.tsx
```diff
@@ -62,6 +62,11 @@ export function UserMenu({
 
   const beginSignOut = useCallback(async () => {
     try {
+      // There are checks in the parent component that should prevent showing this to users if there
+      // is no user context available. We assert here to detect miss-use.
+      if (!signOut) {
+        throw new Error("Invariant: CurrentUserProvider must provider signOut");
```
jtbandes, line 68:
```suggestion
        throw new Error("Invariant: CurrentUserProvider must provide signOut");
```

jtbandes (approved):
I've seen some places have a boolean for supportsSignIn, should we get rid of that with this change?

defunctzombie:
Rolling this change into https://github.com/foxglove/studio/pull/5361

PR #5353, opened 2023-02-14
Add AppModule context

defunctzombie:
**User-Facing Changes**
None

**Description**
The AppModule context provides a way of replacing or injecting certain app components.

### DIFF
--- a/packages/studio-base/src/components/CreateEventDialog.tsx
+++ b/packages/studio-base/src/components/CreateEventDialog.tsx
@@ -33,6 +33,7 @@ import {
   useMessagePipeline,
 } from "@foxglove/studio-base/components/MessagePipeline";
 import Stack from "@foxglove/studio-base/components/Stack";
+import { useAppModule } from "@foxglove/studio-base/context/AppModuleContext";
 import { useConsoleApi } from "@foxglove/studio-base/context/ConsoleApiContext";
 import { EventsStore, useEvents } from "@foxglove/studio-base/context/EventsContext";
 import { useAppTimeFormat } from "@foxglove/studio-base/hooks";
@@ -124,6 +125,7 @@ export function CreateEventDialog(props: { deviceId: string; onClose: () => void
   );
 
   const { formatTime } = useAppTimeFormat();
+  const { createEvent: appModuleCreateEvent } = useAppModule();
 
   const countedMetadata = countBy(event.metadataEntries, (kv) => kv.key);
   const duplicateKey = Object.entries(countedMetadata).find(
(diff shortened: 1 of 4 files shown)


packages/studio-base/src/components/CreateEventDialog.tsx
```diff
@@ -142,19 +144,33 @@ export function CreateEventDialog(props: { deviceId: string; onClose: () => void
...
+        timestamp: event.startTime.toISOString(),
+        durationNanos: toNanoSec(
+          event.durationUnit === "sec"
+            ? { sec: event.duration, nsec: 0 }
+            : { sec: 0, nsec: event.duration },
+        ).toString(),
+        metadata: keyedMetadata,
+      });
+    } else {
+      await consoleApi.createEvent({
```
jtbandes, line 160: Why keep `else`? I thought the goal was to remove the consoleApi.

packages/studio-base/src/context/AppModuleContext.ts
```diff
@@ -0,0 +1,26 @@
...
+
+interface IAppModule {
+  providers?: readonly JSX.Element[];
+  syncAdapters?: readonly JSX.Element[];
+  createEvent?: (args: {
+    deviceId: string;
+    timestamp: string;
+    durationNanos: string;
+    metadata: Record<string, string>;
+  }) => Promise<void>;
```
jhurliman, line 15: `createEvent` seems like a very specific API. Is the idea to expand the `IAppModule` API surface over time as more functionality is added to Studio?

packages/studio-base/src/components/CreateEventDialog.tsx
defunctzombie: Indeed that is the goal but I want to have things always in a working state during the transition.

packages/studio-base/src/context/AppModuleContext.ts
jtbandes: return of the `GlobalHooks`/`GlobalConfig` https://github.com/foxglove/studio/blob/05f25e2698cb289c3f91696872b6e1a8e9f2e6f3/app/GlobalConfig.ts

defunctzombie: Possibly - at this stage this was the only thing I identified as needing a "shim" to be able to extract ConsoleApi. Maybe IAppModule expands or maybe it stays more opaque via the providers array.

defunctzombie:
I will re-work this PR to see if I can simplify the top level providers out.

defunctzombie:
Replacing this with: https://github.com/foxglove/studio/pull/5361

PR #5354, opened 2023-02-14
Add lint rule enforcing all links have an explicit target

jtbandes:
**User-Facing Changes**
None

**Description**
Add a simple rule to enforce that any JSX element with a `href` attribute must also have a `target` attribute.


packages/studio-base/src/components/OpenDialog/Start.tsx
```diff
@@ -119,10 +119,11 @@ type DataSourceOptionProps = {
   icon: JSX.Element;
   onClick: () => void;
   href?: string;
+  target: "_blank";
```
jtbandes, line 122: Had to add a new prop here since this component wraps a link. Seems like this is not super common so not a big deal that the new prop is needed? 🤷

eslint-rules/foxglove-link-target.js
```diff
@@ -0,0 +1,34 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
defunctzombie, line 1: Can we put this in `packages` as an actual package and have it as a dependency in package.json? That will make it easier to consume in our other repo since we probably want this there too?

defunctzombie (approved):

eslint-rules/foxglove-link-target.js
jtbandes: I could - but how will it make it easier to consume anywhere? The eslint setting points at a folder.

defunctzombie: If its a package doesn't eslint pick it up if we add it to the plugins or rules section? There's something about naming the package starting with `eslint-` and some magic stuff happens? Then we don't need the extra CLI arg change you added?

defunctzombie (approved):

PR #5357, opened 2023-02-15
Fixed `nav_msgs/OccupancyGrid` negative color values

mica-angeli:
**User-Facing Changes**

**Description**

When using the costmap palette for `nav_msgs/OccupancyGrid` , the negative colors are reversed.

This is what all the `nav_msgs/OccupancyGrid` colors look like in Rviz.  It's a map generated with values going from -128 to 127 from left to right on each column.

Here's what it looks like in Foxglove before the fix.  Notice the red-yellow gradient is flipped.  This is because the negative values are being miscalculated:

original

I corrected that.  Here's what it looks like now:

foxglove-fixed


mica-angeli:
@snosenzo @jhurliman

snosenzo (approved):
thanks so much for fixing this

PR #5360, opened 2023-02-15
Refactor VersionBanner to avoid muiStyled

2metres:
**User-Facing Changes**
None

**Description**
Remove muiStyled from VersionBanner


jtbandes:
@2metres would you consider also moving this component `studio-base` as part of this change? That way we could use it in the closed source repo as well so we don't have to maintain 2 copies.

2metres:
Closing PR as I'm finding diminishing returns on fixing this

PR #5362, opened 2023-02-15
Fix `nav_msgs/OccupancyGrid` costmap transparency

mica-angeli:
**User-Facing Changes**

**Description**

Cells of value 0 of the `nav_msgs/OccupancyGrid` type were not getting the transparency rendered at all.  See the image below.  The grey cells in the costmap should be fully transparent:

transparency_before

Here's how it looks like in Rviz (at a different time point in the sample data):

transparency_rviz

With this fix, this is how that same time point looks like in Foxglove.  Notice the grey cells are now transparent:

transparency_fixed


mica-angeli:
@snosenzo @jhurliman @defunctzombie

snosenzo (approved):
another good catch, really appreciate the time on these!

PR #5366, opened 2023-02-16
3D: compute vertex normals when overriding mesh color

jtbandes:
**User-Facing Changes**
Fixed color override functionality for certain mesh entities in the 3D panel.

**Description**
Meshes without normals (e.g. https://github.com/mrdoob/three.js/raw/dev/examples/models/gltf/Flamingo.glb) would appear black. Compute vertex normals when replacing the mesh's material.

Fixes FG-1935

image


foxymiles (approved):
This is a one-time calculation right? Shouldn't be any performance issues?

jhurliman (approved):

snosenzo (approved):

jtbandes:
> This is a one-time calculation right?

It's done once per model when the setting is changed to add an override color. If you remove and re-add the override color then it will happen again. But if you just change the color then it won't (because we just change the color of the material).

PR #5378, opened 2023-02-17
Switch from `zstd-codec` to `@foxglove/wasm-zstd`

jhurliman:
**User-Facing Changes**
- None

**Description**
This switches from the zstd-codec module, which could only decompress with the zstd streaming API, to @foxglove/wasm-zstd which has an API similar to @foxglove/wasm-lz4 and can decompress given a byte buffer and target decompressed size. A few related packages were bumped along the way.


packages/mcap-support/src/decompressHandlers.ts
```diff
@@ -23,36 +23,12 @@ async function _loadDecompressHandlers(): Promise<McapTypes.DecompressHandlers>
...
-          }
-        })(),
-        Number(decompressedSize),
-      );
-      if (!result) {
-        throw new Error("Decompression failed");
-      }
-      return result;
-    },
+    zstd: (buffer, decompressedSize) => decompressZstd(buffer, Number(decompressedSize)),
```
jtbandes, line 32: 🎉

yarn.lock
```diff
@@ -2384,7 +2383,17 @@ __metadata:
...
+  version: 0.8.2
+  resolution: "@foxglove/schemas@npm:0.8.2"
+  dependencies:
+    "@foxglove/rosmsg-msgs-common": ^2.0.0
+    tslib: ^2
+  checksum: a2f085a94b168ee3874f76d9505702df0b7e8c9ba5eddd9721de04f5a199462e2afb8d0ce534c0e9bc38c619822b76652f604c4a95cd3a892d03cfac2dd96c13
+  languageName: node
+  linkType: hard
+
+"@foxglove/schemas@npm:^0.7.0":
```
jtbandes, line 2396: I have been thinking of adding a lint rule to prevent this. It seems sketchy if our app is built with multiple different versions of our own packages. These are all things we control and we can probably avoid it. That said it's probably not a big deal for schemas in particular.

jtbandes (approved):

yarn.lock
defunctzombie: Do we know where this dependency difference comes from?

defunctzombie: Which packages tho are the source of the differing schema requirements? I thought the only package that depends on @foxglove/schemas was the studio repo. What other package depends on schemas for runtime?

jtbandes: Looks like @foxglove/rosbag2 does.

I disagree that this is "caused by not following semantic versioning". This duplication could absolutely still happen with versions that don't start with 0.

PR #5379, opened 2023-02-17
Add support for remaining Rviz color modes to `nav_msgs/OccupancyGrid`

mica-angeli:
**User-Facing Changes**

**Description**

This PR adds support for the "Map" and "Raw" `nav_msgs/OccupancyGrid` color modes from Rviz.  It also fixes a bug with how the grid values were being converted that was producing slightly wrong colors.  It also adds support for an "Alpha" field that modifies the alpha color values when any non-custom color mode is selected.

Below is a screenshot showing a sample with the background map set to the "Map" color mode and the obstacle grid set to the "Costmap" color mode:

screenshot1

Below is another screenshot showing a sample with the background map set to the default "Custom" color mode and the obstacle grid set to the "Raw" color mode:

screenshot2


mica-angeli:
@snosenzo @jhurliman @defunctzombie

jhurliman (approved):

PR #5383, opened 2023-02-17
Reset player state when player changes

defunctzombie:
**User-Facing Changes**
Fix an issue that prevented the open dialog from closing after re-selecting a data source in certain situations.

**Description**
This change updates the message pipeline to re-set the player state to the default when there is no player (or when a player becomes undefined which happens when selecting a new player). Resetting the state ensures it moves back to not preset.

This change updates the Connection dialog `Open` action to explicitly close the dialog rather than relying on upstream guessing when a player becomes active to close the dialog.

### DIFF
--- a/packages/studio-base/src/components/MessagePipeline/index.test.tsx
+++ b/packages/studio-base/src/components/MessagePipeline/index.test.tsx
@@ -108,6 +108,30 @@ describe("MessagePipelineProvider/useMessagePipeline", () => {
         setParameter: expect.any(Function),
         pauseFrame: expect.any(Function),
       },
+      {
+        playerState: {
+          activeData: undefined,
+          capabilities: [],
+          presence: PlayerPresence.NOT_PRESENT,
+          playerId: "",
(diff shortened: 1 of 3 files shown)


jhurliman (approved):

jtbandes (approved):

jtbandes:
Does this affect any scenarios where we previously would've kept data visible when you disconnect, but now it disappears instantly? I know we've had feedback from users in the past that when their server/simulation disconnects it is helpful to keep some data visible on screen. But maybe that's not relevant to this scenario because the player still _exists_ in that case?

defunctzombie:
> But maybe that's not relevant to this scenario because the player still _exists_ in that case?

Correct. The player still exists. This is explicitly around when a player stops existing. Technically I think this is still not "perfect" since nothing says that the player cycle has to go through having a player undefined. But I think with the various async things that happen this does occur. Even if it doesn't it felt sensible to reset if the player ever does become undefined.

PR #5386, opened 2023-02-17
Add `listAssets` and `fetchAsset` to extension API

jhurliman:
**User-Facing Changes**
- Extension API has new methods `listAssets()` and `fetchAsset()` for listing and retrieving assets exposed by the current data source, such as attachments from MCAP files.

**Description**
This adds `listAssets()` and `fetchAsset()` methods to the extension panel API. The definition of an Asset closely resembles an MCAP attachment, except asset names are treated as unique keys. The MCAP iterable source handles this by always returning the first attachment with a matching name.


jtbandes:
> It can be used to fetch() remote content, or retrieve attachments from MCAP files.

How does the caller choose which one to do? Should these be separate APIs?

An extension can already use the built-in fetch(), right? So why does it need a separate function for fetching remote content?

jtbandes:
I swear I already posted this comment but I can't find it now so writing again.

> Extension API has new methods listAssets() and fetchAsset() for listing and retrieving remote content or accessing attachments from MCAP files.

When I call fetchAsset("https://...") how do I know whether I am going to get the contents of that url, or an MCAP attachment with the same name? Should remote content and MCAP attachments be separate APIs?

Can't extension panels already use the built-in fetch() function?

jtbandes:
Oh, I found it 😀  
image

jhurliman:
> I swear I already posted this comment but I can't find it now so writing again.
> 
> > Extension API has new methods listAssets() and fetchAsset() for listing and retrieving remote content or accessing attachments from MCAP files.
> 
> When I call fetchAsset("https://...") how do I know whether I am going to get the contents of that url, or an MCAP attachment with the same name? Should remote content and MCAP attachments be separate APIs?
> 
> Can't extension panels already use the built-in fetch() function?

`fetchAsset()` has nothing to do with `global.fetch()`. If you want an asset from the data source, use `fetchAsset()`. If you want an HTTP(S) URL's contents, use `fetch()`.

jtbandes:
I see...so what does "retrieving remote content" refer to? Maybe the wording should be clarified a bit

jtbandes:
We still plan to work on this feature, but will need to make changes to ensure the asset fetching API can interoperate properly with the Data Platform.

Issue #5244, opened 2023-01-26
Studio resets start time to 0 after websocket reconnect


neilisaac:
I've seen a minor issue in studio a couple times where the start time is reset to 0 (unix epoch) after the server is restarted and the client reconnects to the websocket.  This results in plots being compressed and unusable.

- Version: 1.31.1 (darwin)
- Platform: mac desktop app

**Steps To Reproduce**
1. connect to foxglove websocket server
2. start time is set to timestamp of first message
3. restart server
4. client reconnects
5. start time is reset to 0

**Expected Behavior**

start timestamp is reset to first message received after reconnecting

PR #5387, opened 2023-02-17
Avoid clearing state on Foxglove Websocket disconnect

defunctzombie:
**User-Facing Changes**
Fix an issue where disconnects in the Foxglove Websocket connection type would result in resetting the current time and ultimately clear certain panels like the plot panel.

Fix an issue where reconnection attempts in the Foxglove Websocket connection type would cause flickering on some panels.

When a Foxglove Websocket connection is lost, state is maintained so panels like plot continue showing their data.

**Description**
This change updates the FoxgloveWebsocketPlayer to avoid clearing certain state elements when a connection is lost. Rather the player keeps the state and waits for the connection to re-establish before deciding which state elements to clear.

This change also stops updating the clock when a player is disconnected. This makes the time update behavior the same for server supplied clocks and wall clocks - both stop on disconnect.

Fixes FG-1836
Fixes FG-1658
Fixes GH-5244


packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
```diff
@@ -647,11 +654,16 @@ export default class FoxgloveWebSocketPlayer implements Player {
...
   public close(): void {
     this._closed = true;
-    if (this._client) {
-      this._client.close();
-    }
+    this._client?.close();
     this._metricsCollector.close();
     this._hasReceivedMessage = false;
+    if (this._openTimeout != undefined) {
+      clearTimeout(this._openTimeout);
```
jtbandes, line 661:
```suggestion
      clearTimeout(this._openTimeout);
      this._openTimeout = undefined;
```

jtbandes (approved):

PR #5388, opened 2023-02-17
Expose MCAP metadata and MCAP/rosbag attributes in the Parameters panel

jhurliman:
**User-Facing Changes**
- MCAP `Metadata` records can be viewed in the Parameters panel, along with `MCAP_LIBRARY`, `MCAP_PROFILE` parameters for MCAP files and `ROSBAG_CHUNK_COUNT`, `ROSBAG_CONNECTION_COUNT` parameters for rosbag files

**Description**
Pipe MCAP metadata through to `parameters` in the message pipeline and add the `getParameters` capability for MCAP and rosbag files. This also adds `MCAP_LIBRARY`, `MCAP_PROFILE`, `ROSBAG_CHUNK_COUNT`, and `ROSBAG_CONNECTION_COUNT` parameters

screenshot_2023-02-17_at_3 38 41_pm

### DIFF
--- a/packages/studio-base/src/players/IterablePlayer/BagIterableSource.ts
+++ b/packages/studio-base/src/players/IterablePlayer/BagIterableSource.ts
@@ -7,6 +7,7 @@ import { BlobReader } from "@foxglove/rosbag/web";
 import { parse as parseMessageDefinition } from "@foxglove/rosmsg";
 import { MessageReader } from "@foxglove/rosmsg-serialization";
 import { compare } from "@foxglove/rostime";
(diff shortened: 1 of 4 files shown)


jtbandes (changes requested):
My initial reaction is that parameters is the wrong place for this info — I think we should have more detailed internal discussions first so blocking the PR for now :)

defunctzombie:
Closing this since we are not sure parameters is the right place for this. It could be interesting to expose a specific metadata record called "parameters" into the panel tho.

PR #5395, opened 2023-02-18
Move plot series message path above label

amacneil:
**User-Facing Changes**
Move Plot panel series `Message path` text input above `Label`.

**Description**
Since we merged the new plot panel settings, I find myself constantly clicking the top text input of the series expecting to edit the series data. `Label` is an optional field so it feels weird having it above `Message path` - I propose moving it below.

After:
image


foxymiles (approved):
Makes sense. We should make sure to update the screenshots in the docs as part of this change. cc @esthersweon

PR #5397, opened 2023-02-19
Support ZFP-compressed images in the Image and 3D panel

jhurliman:
**User-Facing Changes**
- Adds support for [ZFP](https://computing.llnl.gov/projects/zfp)-compressed images to the Image and 3D panels

**Description**
ZFP image support is a generic solution for depth images or other two-dimensional data requiring >8-bit depth in situations that benefit from message-level compression, such as live streaming or lossy storage. Lossy and lossless modes are supported. The ZFP data must be packaged in a `foxglove.CompressedImage` or `sensor_msgs/CompressedImage` message with `format="zfp"`. Only two-dimensional data is supported, and the compressed stream must be prefixed with a full ZFP header (which describes the shape and format of the data).

Website documentation PR is at <https://github.com/foxglove/website/pull/997> and example code for creating ZFP CompressedImage messages is at <https://github.com/foxglove/mcap/blob/main/cpp/examples/protobuf/write_zfp_images.cpp>.

https://user-images.githubusercontent.com/195374/219933766-a984858a-5aaa-44a9-8af9-0c09469cfe01.mov

### DIFF
--- a/packages/den/image/decodings.ts
+++ b/packages/den/image/decodings.ts
@@ -19,7 +19,7 @@ import Zfp, { ZfpType, ZfpBuffer, ZfpResult } from "wasm-zfp";
 // https://github.com/ros-perception/image_pipeline/blob/42266892502427eb566a4dffa61b009346491ce7/image_view/src/nodes/image_view.cpp#L80-L88
 // https://github.com/ros-visualization/rqt_image_view/blob/fe076acd265a05c11c04f9d04392fda951878f54/src/rqt_image_view/image_view.cpp#L582
 // https://github.com/ros-visualization/rviz/blob/68b464fb6571b8760f91e8eca6fb933ba31190bf/src/rviz/image/ros_image_texture.cpp#L114
-const DEFAULT_MAX = 10000;
+const DEFAULT_INT_MAX = 10000;
 
 // The set of image formats that can natively be rendered in the browser
 export const BROWSER_IMAGE_FORMATS = new Set([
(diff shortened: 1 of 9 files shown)


defunctzombie:
With `format="zfp"` how do you know what they underlying image/data format? From what I've been able to read about `zfp` this would be similar to seeing `format="gzip"` which is not an image format by itself.

Do you have corresponding schema repo or docs updates (or drafts) that we can read about how we would present this to the user and how we'd need the user to format their data?

packages/den/image/decodings.ts
```diff
@@ -11,6 +11,31 @@
...
+  "gif",
+  "jpeg",
+  "png",
+  "svg+xml",
+  "webp",
+]);
+
+// Global buffer for decompressing zfp images. This is a pointer to a malloc()
+// in the wasm-zfp WASM heap for a ZfpBuffer struct
+let zfpBuffer: ZfpBuffer | undefined;
```
snosenzo, line 37: is it possible for this to be used by two different asynchronous decodings at the same time? Could this be an issue?

defunctzombie (changes requested):
Holding until we can do more research into alternative approaches and get feedback from the customer: https://github.com/foxglove/website/pull/997#issuecomment-1441214313

packages/den/image/decodings.ts
jhurliman: Yes, it's possible because the `wasm-zfp` module is single-threaded.

defunctzombie:
Closing. See comment: https://github.com/foxglove/website/pull/997#issuecomment-1497624043

jtbandes:
Copying the linked comment here for public visibility:

> Closing this. I do not think "zfp" is correct as a "format" peer to "png" or "jpeg". I could see us adding a compression field to RawImage and using "zfp" there as generic compression format for certain kinds of byte data. That seems more aligned with what is happening here with the `mono16` data.
> 
> We've also learned that png itself can store mono16 data.

PR #5399, opened 2023-02-21
[DRAFT] H.264 video support in 3D panel

jhurliman:
**User-Facing Changes**
- TBD

**Description**
This is a research spike for native video decoding in Studio. The goal is to define a set of encoding requirements compatible with MCAP and Foxglove WebSocket protocol for H.264 video streams (other codecs TBD) and provide a UX for video streams in Studio that is similar to compressed image streams such as MJPEG.

- Add `timeRange` to panel extension API
- Add "Example CameraInfo" dropdown option to image renderables that synthesizes a plausible CameraInfo
- Add h264 video rendering to Image panel (corrupted video due to `useImagePanelMessages.ts`)
- Add h264 video rendering to 3D panel

***TODO***
- [x] Encoding writeup
- [x] Decoding writeup
- [ ] Break out non-video PRs
- [ ] Solve 2D panel dropped message problem
- [ ] Storybooks

https://user-images.githubusercontent.com/195374/220274087-1df74030-1a95-46a7-ba58-76002c1f3ad4.mov

---

## Goal

The purpose of supporting video playback in Studio is to allow MCAP files, data platform connections, and live robot connections to stream compressed video to Studio directly. This reduces storage costs, bandwidth, and CPU usage in workflows where a robot is already producing compressed video. The underlying sensor data is still the same: a stream of discrete timestamped image frames. Ideally, the user experience should not change due to a different compression approach. The image panel and 3D panel should still be able to render discrete, timestamped images. The correct image should be shown on screen at any given timestamp. Seeking should continue to work[1].

## Encoding

This research spike only covers H.264 video data. However, nothing in the approach precludes HEVC (H.265), VP9, AV1, etc.

To enable the above goals, we need the ability to address individual images and associate a user-provided timestamp with each frame. MCAP messages provide a framing for each image. This is accomplished by:

* Serialize the H.264 bitstream into an MCAP channel comprising discrete messages. The bitstream must be in Annex B or RTP packet format.
* Each message must contain exactly one NALU describing a single frame, either an IDR (I-frame or keyframe) or NDR (P-frame or predictive frame). Every IDR frame must be preceded by SPS and PPS NAL packets within the same message.
* Only IDR and NDR frames are supported. B-frames are not supported.
* Encoding parameters cannot change in a single recording. The SPS/PPS packets must be equivalent throughout the entire stream.
* Each message must have a timestamp in absolute time corresponding to the time that message's video frame should be displayed. In other words, the difference between the current frame's timestamp and the first frame's timestamp is the Presentation Timestamp (PTS).
* Video message timestamps must be strictly monotonic. No two messages can share the same timestamp, and time cannot go backward (receive_time and video message time must have identical ordering).
* The stream must be associated with a `frame_id` to place it into the robotics hierarchy and render in the 3D scene, composited with image annotations, etc.

This raises the question of how to serialize the video bitstream and how to serialize the timestamp and frame_id. Using MCAP, this could be accomplished by directly storing the NALU chunks described above as Message records and setting the channel encoding to h264 with no schema (schema_id=0). However, this would require storing the frame_id as channel metadata and overloading the usage of either log_time or publish_time. In all other camera sensor-based topics, a wrapping message is used to hold the timestamp and frame_id. This keeps the "application-level" timestamps (sometimes referred to as header stamps from the ROS convention of `.header.stamp`) separate from "recording-level" timestamps. Using MCAP Message records directly would require reinterpreting one of the recording-level timestamps as an application-level timestamp, a practice that is generally avoided.

Looking at the wrapping approach, `foxglove.CompressedImage` seems like a natural fit. The data we store in a single CompressedImage message corresponds to a single decompressed image, but with the added caveat that it's not always possible to render this single image without processing additional messages that came before it. An alternative could be a `foxglove.CompressedVideo` message that has an identical structure to `foxglove.CompressedImage` and is otherwise treated as an alias.

Reference C++ code will likely need to be made available to demonstrate encoding video into an MCAP file following these constraints. Otherwise, it is easy to miss a small detail like not repeating SPS/PPS before each IDR, writing multiple P-frames to a single message, using the wrong timestamp, etc.

## Decoding

Given the above encoding constraints, decoding in a modern browser becomes very straightforward using [WebCodecs](https://www.w3.org/TR/webcodecs/#video-decoder-config). We look for the first observed SPS packet and parse it to build a `VideoDecoderConfig`. Then we wait for the first message that contains an IDR frame (which could be the message that was just parsed, or it may come later if the message was a SPS/PPS/NDR combo) and feed that into the `VideoDecoder`. Each (and every!) following message is fed into the VideoDecoder, until playback reaches the end or a seek is performed. At either of these conditions, `reset()` is called on the encoder and the above loop repeats.

Each encoded chunk of video data is fed into the decoder as a raw blob of bytes plus a microsecond timestamp and a `type` set to "key" or "delta". The type is determined by parsing the NAL packets in the message and checking if this is an IDR frame or NDR. The timestamp is intended to be the Presentation Timestamp (PTS), and should be an integer value relative to the beginning of the video. We can either use the `startTime` of the data source as the zero point, or the timestamp of the first message fed into the decoder, as long as it's consistent.

By setting [optimizeForLatency](https://www.w3.org/TR/webcodecs/#dom-videodecoderconfig-optimizeforlatency) to true in the decoder config and feeding in NALUs that described exactly one frame, the decoder will trigger a callback with our `VideoFrame` shortly after calling `.decode()`. Although we can correlate decoder input with VideoFrame output using the `timestamp` field, there is no explicit coupling between calling decode and receiving a callback so we use a Promise with a timeout. If the frame is not produced in time it's not a major issue since the correct bytes were still fed into the decoder, it just means we ran out of time for the current frame and doesn't have image data to paint on the screen.

The primary challenge with decoding is seeking. To seek and immediately show the most recent image on screen, we need to start at the most recent keyframe (IDR) before the target frame and feed it into the VideoDecoder, followed by each P-frame (NDR) sequentially up to the target frame to display. This requires either fetching each previous message one by one and parsing the NAL packets, or keeping an index of where all the IDR frames are. It also requires feeding messages into the VideoDecoder and closing+discarding the intermediate VideoFrames up to the one we want. Seeking is out of scope for v1.

## Additional Considerations

Implementing this in the 3D panel was straightforward. A H264Decoder wrapper class handles the internal details of parsing bitstreams and feeding the decoder, it exposes two high-level methods `decode()` and `resetOnSeek()`. The Image panel implementation ran into two issues:

1. `useImagePanelMessages.ts` takes a batch of messages and discards all but the most recent message. This is incompatible with ingesting all messages between the most recent IDR and the current frame.
2. `downloadImage.ts` works by feeding a message back through the decoding pipeline to produce another image bitmap targeted at a non-screen canvas. This is incompatible with feeding each message exactly once into the video decoder.

### DIFF
--- a/packages/studio-base/src/panels/Plot/PlotLegendRow.tsx
+++ b/packages/studio-base/src/panels/Plot/PlotLegendRow.tsx
@@ -162,9 +162,9 @@ export function PlotLegendRow({
       if (timeToCompare == undefined || pt.x > timeToCompare) {
         break;
       }
-      value = pt.y;
+      value = pt.value;
     }
-    return value;
+    return value?.toString();
   }, [showPlotValuesInLegend, hoverValue?.value, currentTime, correspondingData]);
 
   return (


jon-chuang:
> To seek and immediately show the most recent image on screen, we need to start at the most recent keyframe (IDR) before the target frame and feed it into the VideoDecoder, followed by each P-frame (NDR) sequentially up to the target frame to display.

I think this may be possible to do by separating I-frame and and P-frame into two separate channels.

jhurliman:
I am currently looking at a slightly modified approach for this that puts the responsibility of SPS/PPS decoding on the MCAP writer to store as channel metadata, and using the least significant bit of the MCAP log_time to indicate if a message contains a keyframe or not. This would remove the need to do bitstream parsing in TypeScript, improving playback performance and supporting HEVC/AV1/etc without modification.

jhurliman:
@jon-chuang I agree manipulating the nanosecond timestamp of messages feels like a hack. It does have some compelling advantages, though:

- `log_time` is the only field in MCAP that is currently indexed
- There are two timestamps for every message, `publish_time` and `log_time`. This would only be modifying `log_time`, which is usually the timestamp when a message is written to the MCAP and there's no expectation of nanosecond alignment between messages (I think it actually introduces a poorly defined ordering if message across different channels share identical `log_time`).

Having said that, I can't think of a reason why splitting video across two separate channels for I frames and P frames wouldn't work.

I don't think B frame support will realistically land, it means feeding messages into the WebCodecs decoder until the frame we are looking for is eventually rendered instead of having a deterministic "feed this sequence of messages in and the frame I want will be rendered as a result". I imagine whatever spec arises from this work will explicitly forbid B frames.

The best way to move this work forward IMO is to generate sample MCAP files that can be shared and experimented with. Create a file using the two channel approach if that's what you're currently favoring. When the Studio side of this work is picked up again it will be very helpful to have multiple test files to experiment with.

jhurliman:
@jon-chuang the approach I've landed on for keyframe indexing is to write a separate channel (with a topic name such as `<mytopic>/keyframes`) with schema_id=0 encoding="keyframe_index" and reference this topic name from the video channel's channel metadata with keyframe_index=<mytopic>/keyframes. The index needs to be written into its own Chunk records to avoid having to decode all of the video Chunk records to read the index. But using this approach avoids modifying timestamps, avoids splitting the video bitstream into separate channels that have to be muxed back together for simple linear playback, and works nicely with the existing Studio architecture since the keyframe index topic can be preloaded into memory without Attachment or Metadata record support.

jhurliman:
Here's a command line utility to create video MCAP files: https://github.com/MetaverseIndustries/mcaptool

`mcaptool input.mp4 output.mcap`

jhurliman:
This has been superseded by <https://github.com/foxglove/studio/pull/5806>, feel free to close this PR.

PR #5402, opened 2023-02-21
Add `startTime` and `endTime` to panel extension API

jhurliman:
**User-Facing Changes**
- Adds `startTime: Time | undefined` and `endTime: Time | undefined` to the panel extension API for querying start/end time of the data source

**Description**
Generally useful thing to expose, and can potentially be useful for video playback

### DIFF
--- a/packages/studio-base/src/components/PanelExtensionAdapter/renderState.ts
+++ b/packages/studio-base/src/components/PanelExtensionAdapter/renderState.ts
@@ -342,6 +342,22 @@ function initRenderStateBuilder(): BuildRenderStateFn {
       }
     }
 
+    if (watchedFields.has("timeRange")) {
+      const startTime = activeData?.startTime;
+      const endTime = activeData?.endTime;
+
+      if (!renderState.timeRange) {
+        shouldRender = true;
+        renderState.timeRange = [startTime, endTime];
+      } else {
+        if (startTime !== renderState.timeRange[0] || endTime !== renderState.timeRange[1]) {
+          shouldRender = true;
+          renderState.timeRange[0] = startTime;
+          renderState.timeRange[1] = endTime;
+        }
+      }
+    }
+
     if (watchedFields.has("previewTime")) {
       const startTime = activeData?.startTime;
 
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/components/PanelExtensionAdapter/renderState.ts
```diff
@@ -342,6 +342,22 @@ function initRenderStateBuilder(): BuildRenderStateFn {
...
+      const endTime = activeData?.endTime;
+
+      if (!renderState.timeRange) {
+        shouldRender = true;
+        renderState.timeRange = [startTime, endTime];
+      } else {
+        if (startTime !== renderState.timeRange[0] || endTime !== renderState.timeRange[1]) {
+          shouldRender = true;
+          renderState.timeRange[0] = startTime;
+          renderState.timeRange[1] = endTime;
```
jhurliman, line 356: Should this be `timeRange = [startTime, endTime]` because of React?

packages/studio/src/index.ts
```diff
@@ -203,6 +203,13 @@ export interface RenderState {
    */
   currentTime?: Time;
 
+  /**
+   * The start and end timestamps of the playback range for the current data source. For offline
+   * files it is expected both values will be present. For live connections, the start time may
+   * or may not be present depending on the data source and the end time will be undefined.
```
jhurliman, line 209: Is this true, or can we tighten this and guarantee that `startTime` will always be provided when the data source is connected?

packages/studio/src/index.ts
```diff
@@ -203,6 +203,19 @@ export interface RenderState {
...
+  /**
+   * The start timestamp of the playback range for the current data source. For offline files it
+   * is expected to be present. For live connections, the start time may or may not be present
+   * depending on the data source.
+   */
+  startTime?: Time;
+
+  /**
+   * The end timestamp of the playback range for the current data source. For offline files it
+   * is expected to be present. For live connections, the end time will be undefined.
```
defunctzombie, line 215: Should we keep this more open ended and say it may or may not be defined depending on the data source like you have for startTime? If we say it will be `undefined` that prevents us from coming back to add this for live connections since folks will certainly start to use this as a nice indicator of something.

jhurliman: Updated

foxymiles (approved):

defunctzombie (approved):

PR #5411, opened 2023-02-23
Fix post-release workflow

defunctzombie:
**User-Facing Changes**
None

**Description**
Rather than checking out the target_commitish, checkout the release/v{tag_name} branch. This gives the version bump action a branch to push the bumped version whereas checking out a commit does not.

Example failed workflow: https://github.com/foxglove/studio/actions/runs/4245494198/jobs/7386039870

### DIFF
--- a/.github/workflows/post-release.yml
+++ b/.github/workflows/post-release.yml
@@ -4,12 +4,6 @@ name: Post Release
 on:
   release:
     types: [released]
-  workflow_dispatch:
-    inputs:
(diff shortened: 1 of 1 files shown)


defunctzombie:
I think the other option could be to make the release point at a branch. Right now I have draft releases pointing at a specific commit on the release branch. So we can go with this fix to make things work for the next release and change later.

esthersweon (approved):

PR #5413, opened 2023-02-23
Remove @mdi icons as dependency

2metres:
**User-Facing Changes**
None

**Description**
Absorb existing `@mdi` icons and remove as a dependency


packages/studio-base/src/components/TimeBasedChart/TimeBasedChartLegend.tsx
```diff
@@ -70,7 +71,7 @@ export default function TimeBasedChartLegend(props: Props): JSX.Element {
         }
 
         const CheckboxComponent =
-          linesToHide[label] === true ? CheckboxBlankOutlineIcon : CheckboxMarkedIcon;
+          linesToHide[label] === true ? CheckBoxOutlineBlankIcon : CheckBoxIcon;
```
2metres, line 74: This should just be a checkbox TBH

defunctzombie (dismissed):
Which icon package should we be using - mui? It might be nice to make a package to consolidate our icons into that and only ever use icons from that package.

jtbandes:
What's the point of doing this? Why would we maintain our own svgs vs importing them from a source where we can understand the origin of the icon and get updates for it?

2metres:
> Which icon package should we be using? 

Currently we have multiple options when it comes to which icon should I use:

`@mui/icons-material` - Only official material UI icons exported as react componets - 177 results, 65 files
`@mdi` Material Design Icons - official and unoffical Material Design Icons - 12 results, 6 files
`@fluentui/react-icons` - Fluent UI system icons (Personally I prefer these to MUI ones) - 39 results, 6 files 
+ a few custom icons we have created for pose, etc.

2metres:
> What's the point of doing this? Why would we maintain our own svgs vs importing them from a source where we can understand the origin of the icon and get updates for it?

The majority of the icons we are importing from `@mdi` are icons that look like Material Icons but aren't included in the main set for whatever reason. The intent of this PR is not to avocate for importing all the SVG icons we use into our app but to discourage the splintering of our icons. 

Long term I would prefer if we made an effort to use more of the Fluent UI icons as I feel like they are more polished than the Material UI ones. Potentially deleting the embedded ones as well with better alternatives

2metres:
Reading my own comment I've decided to replace some of the absorbed icons with Fluent UI analogues.

jtbandes:
I don't see any new fluent ui icons being used in the current version of the PR. Forgot to push some changes?

jtbandes (approved):

PR #5415, opened 2023-02-23
Add "Example CameraInfo" dropdown option to image renderables that synthesizes a plausible CameraInfo

jhurliman:
**User-Facing Changes**
- Add "Example CameraInfo" option to image topics in the 3D panel to project images into 3D without camera calibrations

**Description**
This adds a physically plausible (based on RealSense D435 sensor) CameraInfo option that can be selected when your 3D panel image topic does not have an associated CameraInfo topic.

### DIFF
--- a/packages/den/image/decodings.ts
+++ b/packages/den/image/decodings.ts
@@ -11,6 +11,17 @@
 //   found at http://www.apache.org/licenses/LICENSE-2.0
 //   You may not use this file except in compliance with the License.
 
+// The set of image formats that can natively be rendered in the browser
+export const BROWSER_IMAGE_FORMATS = new Set([
+  "apng",
+  "avif",
+  "gif",
+  "jpeg",
+  "png",
+  "svg+xml",
+  "webp",
+]);
+
 function yuvToRGBA8(
   y1: number,
   u: number,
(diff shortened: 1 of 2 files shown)


jtbandes (commented):
- What's the rationale behind making this opt-in instead of just always doing it when no CameraInfo/calibration is present?
- Wording-wise, "Example" seems like a weird phrase, since I think of "sample/example" as meaning something you'd use to demo a product feature, not rely on for everyday usage
- It makes a bit of sense for this synthesized camera info to be created in a function called `_updateImageRenderable`, less sense for it to be created in a function called `_updateImageBitmap`. I think it would make even more sense to update it in a function called `handleImage`.

defunctzombie:
> (based on RealSense D435 sensor)

Why is this the chosen sensor? My first reaction to seeing this feature is that it would allow me to configure the pixels -> meters (world space) unit multiplier and adjust the distance rather than pre-assigning a camera model.

When I select the example option in the sample data I still see this error and the image does not display:

image

defunctzombie:
We might pick this up after we finish more of the 3d scene in image panel work. Closing since we'll probably take a slightly different approach as the code changes.

PR #5419, opened 2023-02-24
Remove debounce from 3d settings update

foxymiles:
**User-Facing Changes**
None

**Description**
Removes the debounce from the 3d panel settings update to fix issues with cursor jumps in settings inputs.

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/SceneExtension.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/SceneExtension.ts
@@ -2,7 +2,7 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
-import { debounce, set, unset } from "lodash";
+import { set, unset } from "lodash";
 import * as THREE from "three";
 import { DeepPartial } from "ts-essentials";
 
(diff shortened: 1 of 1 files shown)


foxymiles:
cc @defunctzombie

defunctzombie:
@foxymiles what interaction should I try on this PR to make this a bad experience? I gave this a quick try with panning around the 3d scene with 3d panel settings open and could not say that I thought it was a bad experience. I also started playback and that seemed to run as it always does for me.

defunctzombie:
Tried this on my NUC 11 with the sample data and settings open. I think we should merge this. I am not seeing what specific downside this creates and we know that without it you cannot actually type into the settings fields.

foxymiles:
@defunctzombie Performance is the concern here. If you think it's acceptable maybe we merge it for now and at some point see if we can reduce the rerendering in the sidebar for 3d?

defunctzombie:
@foxymiles performance I'd say is always a concern but I am not seeing the difference with the debounce off. The challenge with saying performance is a concern here is that I don't see what number I should be measuring to say "oh yes this is 2x slower".

If we change the debounce time to 50ms does it fix the underlying bug? Is there a debounce on the overall speed of settings tree updates generally? Like if I call "updateSettingsTree" 10 times back-to-back (which I think does happen), does that result in 10 re-renders or do we have a debounce downstream limiting this to requestAnimationFrame speeds?

foxymiles:
@defunctzombie Sorry I should have been more specific here. The thing you want to compare is scrubbing numeric values in the sidebar. Panning around directly in the 3d panel is a different code path.

defunctzombie (approved):
:shipit:

Issue #5426, opened 2023-02-24
3D panel selection rendering issues


harip3rla:
**Description**
In the 3D panel window of Foxglove studio when the user selects an object to view it's properties, the 3D panel/view goes blank. Only way to bring it back is to switch to a different layout/panel and return to the 3D panel

- Version:1.42
- Platform: Studio

**Steps To Reproduce**

**Expected Behavior**
When selecting an object in the 3D panel, the object properties should be displayed and the screen should not go blank

jtbandes:
Thanks for the report. This regression seems to have been caused by https://github.com/foxglove/studio/pull/5045.

PR #5432, opened 2023-02-25
Add an error boundary to built-in panels

defunctzombie:
**User-Facing Changes**
Catch more types of errors for built-in panels and surface them as panel errors with the existing panel error boundary.

**Description**
When a built-in panel has an error in the react tree this error would bubble up, log to the console, but never result in a panel error boundary to show. This change updates the built-in panels with a new error boundary which relays this error back to the main app react tree where it can be handled by the panel error boundary we already have.

This change does not address 3rd-party panels which cause unhandled error. The solution for those is to move extension panels into an iframe.

Related issue which surfaced the gap in error handling: https://github.com/foxglove/studio/issues/5426 https://github.com/foxglove/studio/pull/5427

### DIFF
--- a/packages/studio-base/src/components/CaptureErrorBoundary.tsx
+++ b/packages/studio-base/src/components/CaptureErrorBoundary.tsx
@@ -1,0 +1,19 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+import { Component, PropsWithChildren, ReactNode } from "react";
+
+type Props = {
+  onError: (err: Error) => void;
+};
+
+/** An error boundary that calls an onError function when it captures an error */
+export class CaptureErrorBoundary extends Component<PropsWithChildren<Props>, unknown> {
+  public override componentDidCatch(error: Error): void {
+    this.props.onError(error);
+  }
+
+  public override render(): ReactNode {
+    return this.props.children;
+  }
+}
--- a/packages/studio-base/src/components/CaptureErrorBoundary.tsx
+++ b/packages/studio-base/src/components/CaptureErrorBoundary.tsx
@@ -7,13 +7,27 @@ type Props = {
   onError: (err: Error) => void;
 };
 
+type State = {
+  hadError: boolean;
+};
+
 /** An error boundary that calls an onError function when it captures an error */
-export class CaptureErrorBoundary extends Component<PropsWithChildren<Props>, unknown> {
+export class CaptureErrorBoundary extends Component<PropsWithChildren<Props>, State> {
+  public override state: State = {
+    hadError: false,
+  };
+
   public override componentDidCatch(error: Error): void {
+    this.setState({ hadError: true });
     this.props.onError(error);
   }
 
   public override render(): ReactNode {
+    // Avoid rendering children since the children are what caused the error
+    if (this.state.hadError) {
+      return <></>;
+    }
+
     return this.props.children;
   }
 }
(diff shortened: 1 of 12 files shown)


defunctzombie:
This PR is not ready to merge but shows my proposed approach to handle this for our built-in panels. If we are ok with this approach I would move the common elements into the components directory and update the other panels.

packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
```diff
@@ -390,6 +390,8 @@ function useRendererProperty<K extends keyof Renderer>(
 export function ThreeDeeRender({ context }: { context: PanelExtensionContext }): JSX.Element {
   const { initialState, saveState } = context;
 
+  throw new Error("foob");
```
foxymiles, line 393: There's no recovering from a `foob`.

packages/studio-base/src/panels/ThreeDeeRender/index.tsx
```diff
@@ -12,10 +13,48 @@ import { SaveConfig } from "@foxglove/studio-base/types/panels";
 
 import { ThreeDeeRender } from "./ThreeDeeRender";
 
-function initPanel(context: PanelExtensionContext) {
+type RethrowErrorBoundaryProps = {
+  onError: (err: Error) => void;
+};
+
+/** An error boundary that calls an onError function when it captures an error */
+class RethrowErrorBoundary extends Component<
```
foxymiles, line 21: This can only be done in a class component right?

foxymiles:
Seems sensible to me.

packages/studio-base/src/panels/ThreeDeeRender/index.tsx
defunctzombie: Right. That's why this is a class component... Do you think I've done something wrong here?

packages/studio-base/src/panels/ThreeDeeRender/index.tsx
```diff
@@ -12,10 +13,48 @@ import { SaveConfig } from "@foxglove/studio-base/types/panels";
...
+    (err: Error) =>
+      // Trick to throw in the set function will make the error catchable by an error boundary
+      setError(() => {
+        throw err;
+      }),
+    [],
+  );
+}
+
+function initPanel(crash: (err: Error) => void, context: PanelExtensionContext) {
```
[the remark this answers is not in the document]
defunctzombie: I did not want to make a public API decision on this one especially if the direction is towards iframe or some isolation - then we could handle uncaught errors for all panels at the document level (within the iframe).

packages/studio-base/src/panels/ThreeDeeRender/index.tsx
foxymiles: No just making sure I understand why it's a class component.

packages/studio-base/src/panels/ThreeDeeRender/index.tsx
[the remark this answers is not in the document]
jtbandes: 👍  seems alright then

packages/studio-base/src/components/CaptureErrorBoundary.tsx
```diff
@@ -0,0 +1,19 @@
...
+};
+
+/** An error boundary that calls an onError function when it captures an error */
+export class CaptureErrorBoundary extends Component<PropsWithChildren<Props>, unknown> {
+  public override componentDidCatch(error: Error): void {
+    this.props.onError(error);
+  }
+
+  public override render(): ReactNode {
+    return this.props.children;
```
jtbandes, line 17: I think technically you are supposed to stop rendering children when an error is caught, otherwise the error will be thrown again.

jtbandes (approved):

PR #5434, opened 2023-02-27
Output stable time values in extension panel renderState updates

foxymiles:
**User-Facing Changes**
None

**Description**
Fix logic in extension adapter renderState update function. Changes the semantics of currentTime, startTime and endTime in the render state to always contain a defined value if the value is available, instead of emitting alternating defined & undefined values to represent distinct snapshots of those values.

### DIFF
--- a/packages/studio-base/src/components/PanelExtensionAdapter/renderState.test.ts
+++ b/packages/studio-base/src/components/PanelExtensionAdapter/renderState.test.ts
@@ -34,6 +34,63 @@ describe("renderState", () => {
     });
   });
 
+  it("should provide stable time values", () => {
+    const buildRenderState = initRenderStateBuilder();
+    const initialState: Parameters<typeof buildRenderState>[0] = {
+      watchedFields: new Set(["currentTime", "endTime", "previewTime", "startTime"]),
+      appSettings: undefined,
+      currentFrame: [],
+      playerState: {
+        presence: PlayerPresence.PRESENT,
+        progress: {},
+        capabilities: [],
+        profile: "test",
+        playerId: "123",
+        activeData: {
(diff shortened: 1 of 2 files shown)


jtbandes:
I think part of the original idea of the renderState was that a field could be undefined if its value did not change. But now that I look at this code I'm not sure that was ever really implemented. Should we try to do that (which would require keeping track of the last-published values in a separate place) or, on the other hand, maybe we should reconsider whether these fields actually need to be nullable?

cc @defunctzombie as original author of renderState stuff

defunctzombie:
> I think part of the original idea of the renderState was that a field could be undefined if its value did not change.

Yea - this was the intent tho when I look back at the documentation in `studio/index.ts` I don't see any mention of that in the interface documentation. For example with `currentFrame` we don't send the user the existing data if there are no updates. 

My original thinking here was to avoid explaining "reference equality checking" to users and instead treat the presence of the field as the "event" that there is new data. It does mean that the absence of a field does not actually indicate the data is gone tho. So far we've not needed that at the field level. Maybe this is a poor choice in interface design? Or maybe we can do different things for different fields. I am open to feedback given the design goal I described.

foxymiles:
@defunctzombie I think that approach is going to make life hard for people doing panels in React, where a change in a field is meant to represent a change in the underlying state, not an instantaneous event.

defunctzombie:
We discussed in slack and came to the conclusion that many of our fields remain present in the render state so your change here would align with that behavior. We determined that `currentFrame` and `currentTime` currently don't exhibit that behavior but that is out of scope for this PR and we can revisit that separately.

defunctzombie (approved):

jtbandes (approved):

PR #5445, opened 2023-02-28
IterablePlayer: Fix setting of playback speed after playback has ended not working correctly

achim-k:
**User-Facing Changes**
- Fix setting of playback speed after playback has ended not working correctly

**Description**
When changing the playback speed after playing back an mcap/rosbag file, the new playback speed was not immediately taken into account as some internal state was not reseted. This PR fixes that.

Fixes FG-2170

### DIFF
--- a/packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
+++ b/packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
@@ -967,6 +967,10 @@ export class IterablePlayer implements Player {
     try {
       while (this._isPlaying && !this._hasError && !this._nextState) {
         if (compare(this._currentTime, this._end) >= 0) {
+          // Playback has ended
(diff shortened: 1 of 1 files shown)


jtbandes (approved):

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
```diff
@@ -967,6 +967,10 @@ export class IterablePlayer implements Player {
     try {
       while (this._isPlaying && !this._hasError && !this._nextState) {
         if (compare(this._currentTime, this._end) >= 0) {
+          // Playback has ended
```
defunctzombie, line 970:
```suggestion
          // Playback has ended. Reset internal trackers for maintaining the playback speed.
```

defunctzombie (approved):

PR #5446, opened 2023-02-28
Switch to @mcap/support

jhurliman:
**User-Facing Changes**
- None

**Description**
- Replace `@foxglove/mcap-support` with the npm-published `@mcap/support`. :hocho: duplicate code


defunctzombie:
We are not yet ready for this change because we don't know what form the support package will take. We had one support package in studio because it didn't matter to split the dependencies or organize anything better than whatever studio internals wanted. None of this has been designed or documented around external users.

PR #5452, opened 2023-03-01
Improve player state display while initializing a data platform datasource

foxymiles:
**User-Facing Changes**
Improve player state display while initializing a data platform datasource

**Description**
While initializing a data platform datasource there is a point in time where the `playerName` is undefined and the player state is `INITIALIZING`. This changes the value we display in this case from `unknown` to `Initializing...`.


jtbandes (approved):
Seems fine. I also wonder what it would be like if we just made playerName non-optional? I guess allowing it to be undefined gives us a chance at slightly more consistent handling during initialization...

PR #5453, opened 2023-03-01
Display user avatar image

foxymiles:
**User-Facing Changes**
None (feature flagged)

**Description**
Display the user's avatar image in the app bar.


packages/studio-base/src/components/AppBar/index.stories.tsx
```diff
@@ -74,6 +74,7 @@ export function UserPresent(): JSX.Element {
 
   const me = {
     id: "fake-userid",
+    avatarImageUrl: "https://loremflickr.com/64/64/animals?lock=56789",
```
jtbandes, line 77: Maybe a `data` url to avoid issues if this site goes down?

jtbandes (approved):
User-facing changes should probably be none since this is feature flagge.d

packages/studio-base/src/components/AppBar/index.stories.tsx
foxymiles: Good idea. Changed.

PR #5457, opened 2023-03-01
Fix user profile button showing add panel menu

jtbandes:
**User-Facing Changes**
None

**Description**
More of the same #5456


jtbandes:
#5469

Issue #5380, opened 2023-02-17
If there is a part of empty data in the bag file, Jumping from empty data to later pages gets stuck


Woodii1998:
**Description**
If there is a part of empty data in the bag file, Jumping from empty data to later pages gets stuck

- Version: foxglove online version
- Platform: web

**Steps To Reproduce**

https://user-images.githubusercontent.com/33593572/219618367-783cd30e-a736-4b2e-b2cd-a7fd677cef54.mp4

**Expected Behavior**
The bag will play normally after the jump

Woodii1998:
If you need this bag file:
https://drive.google.com/file/d/1jLV-jY9KeOdor6BbJSkBEwmX-AIluBdq/view?usp=sharing

jtbandes:
I'm not able to reproduce this issue. I tried uploading the bag file to foxglove data platform, and this is the behavior I see.
Can you provide more details or clearer reproduction steps?


https://user-images.githubusercontent.com/14237/220793714-b49c7d2c-c07e-47ab-805e-6b8f0c8247fd.mov

Woodii1998:
The fact that I can still reproduce it stably, I think the steps to reproduce it have the following points.
- Make sure the page has no data cache
- Play from the beginning
- Make sure you play to the blank data section before skipping (I play to about ```2023-02-06T02%3A47%3A51.769733341Z``` if you're not sure where you need to play to)


https://user-images.githubusercontent.com/33593572/220836138-0e5fb30e-4313-426f-a6fa-b3d05d941553.mp4

Woodii1998:
https://github.com/foxglove/studio/blob/main/packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts#L671

I tried to troubleshoot the issue and then found some clues, but probably didn't catch the essence of the problem

Normally when jumping this._abort.signal=true
When the problem occurs this._abort.signal=false

I'm not sure if this is the cause of the problem but I hope it helps you guys

defunctzombie:
I've been able to reproduce some of the undesirable behavior and am taking a look. Thank you for sharing the test bag and more detailed analysis.

defunctzombie:
@Woodii1998 Thanks for reporting this and providing the test data to reproduce. The fix for this is deployed to `studio.foxglove.dev`. Please give it a try and see if your issue is resolved.

PR #5466, opened 2023-03-02
Exit the producer loop when aborting during stamp wait

defunctzombie:
**User-Facing Changes**
Fix an issue where the playback would lock-up in certain circumstances.

**Description**
For some data sources, large gaps in data produce `stamp` results in their message iterators. These stamp results indicate where the underlying data source has moved its cursor. The BufferedIterableSource compares such stamp results to the current playhead location to determine whether it needs to buffer more data from the data source. The logic for this comparison sits in a loop until the playhead indicates it is time to read more data. This loop did not properly handle abort signals which resulted in the "producer" thread of the BufferedIterableSource never exiting.

This stall in the BufferedIterableSource was obfuscated by logic in IterablePlayer which did not wait on the return value of the message iterator from the BufferedIterableSource instance. This resulted in the stall occurring on transitions through the next seek, backfill, and play states of the player rather than during the initial transition to backfill. This change updates the IterablePlayer logic so it waits on the existing message iterator to exit before continuing to process the new state. This makes it more obvious if the player gets stuck trying to transition to the new state. It also removes an instance of promise handling that did not await the promise value which contributed to making it harder to track down the issue.

Fixes: #5380

### DIFF
--- a/packages/studio-base/src/players/IterablePlayer/BufferedIterableSource.ts
+++ b/packages/studio-base/src/players/IterablePlayer/BufferedIterableSource.ts
@@ -116,7 +116,15 @@ class BufferedIterableSource implements IIterableSource {
 
         // When receiving a stamp result we enqueue the result into the cache and notify a reader.
         if (result.type === "stamp" && compare(result.stamp, readUntil) >= 0) {
+          // Continue to wait until our stamp time surpasses the readUntil and we know that
+          // we should read more data.
           while (compare(result.stamp, readUntil) >= 0) {
+            // The producer may have aborted while we are waiting for the read head to progress
+            // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+            if (this.aborted) {
+              return;
+            }
+
             // enqueue the stamp and wakeup the reader
             // but the reader might not be reading - cause it is not being read from
             this.cache.enqueue(result);
@@ -164,6 +172,7 @@ class BufferedIterableSource implements IIterableSource {
   }
 
   public async stopProducer(): Promise<void> {
+    log.debug("Stopping producer");
     this.aborted = true;
     this.writeSignal.notifyAll();
     await this.producer;
--- a/packages/studio-base/src/players/IterablePlayer/BufferedIterableSource.test.ts
+++ b/packages/studio-base/src/players/IterablePlayer/BufferedIterableSource.test.ts
@@ -569,4 +569,55 @@ describe("BufferedIterableSource", () => {
     // We should have called the messageIterator method only once
     expect(messageIteratorCount).toEqual(1);
   });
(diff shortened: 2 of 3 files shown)


foxymiles (approved):
Not very familiar with this code but seems reasonable.

packages/studio-base/src/players/IterablePlayer/BufferedIterableSource.ts
```diff
@@ -116,7 +116,15 @@ class BufferedIterableSource implements IIterableSource {
 
         // When receiving a stamp result we enqueue the result into the cache and notify a reader.
         if (result.type === "stamp" && compare(result.stamp, readUntil) >= 0) {
+          // Continue to wait until our stamp time surpasses the readUntil and we know that
+          // we should read more data.
           while (compare(result.stamp, readUntil) >= 0) {
+            // The producer may have aborted while we are waiting for the read head to progress
+            // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+            if (this.aborted) {
+              return;
```
jtbandes, line 125: It looks like the `this.readSignal.notifyAll(); this.readDone = true;` later will happen in this case because it's in a `finally`. is that desired?

packages/studio-base/src/players/IterablePlayer/BufferedIterableSource.ts
```diff
@@ -116,7 +116,15 @@ class BufferedIterableSource implements IIterableSource {
 
         // When receiving a stamp result we enqueue the result into the cache and notify a reader.
         if (result.type === "stamp" && compare(result.stamp, readUntil) >= 0) {
+          // Continue to wait until our stamp time surpasses the readUntil and we know that
+          // we should read more data.
           while (compare(result.stamp, readUntil) >= 0) {
```
jtbandes, line 121: is this loop effectively busy-waiting in the case where there is a large gap?

jtbandes (approved):

packages/studio-base/src/players/IterablePlayer/BufferedIterableSource.ts
defunctzombie: "busy" waiting in that it waits on the read signal which allows us to compute a new readUntil and yes this fires constantly while playing. An alternative implementation would be to set the stamp and have the consumer thread notify when the readUntil reached the stamp and the producer thread only waits once.

packages/studio-base/src/players/IterablePlayer/BufferedIterableSource.ts
defunctzombie: Yep - those wakeup the waiting consumer thread and allow it to exit.

PR #5467, opened 2023-03-02
Improve raw messages panel render performance 

foxymiles:
**User-Facing Changes**

**Description**
Memoizing our `Value` component seems to significantly speed up rendering the tree.

### DIFF
--- a/packages/studio-base/src/panels/RawMessages/Value.tsx
+++ b/packages/studio-base/src/panels/RawMessages/Value.tsx
@@ -222,4 +222,6 @@ function Value(props: ValueProps): JSX.Element {
   );
 }
 
+// In practice this seems to be an expensive component to render.
+// Memoization provides a very noticeable performance boost.
 export default React.memo(Value);
--- a/packages/studio-base/src/panels/RawMessages/Value.tsx
+++ b/packages/studio-base/src/panels/RawMessages/Value.tsx
@@ -76,7 +76,7 @@ const emptyAction: ValueActionItem = {
 
 const MAX_ACTION_ITEMS = 4;
 
-export default function Value(props: ValueProps): JSX.Element {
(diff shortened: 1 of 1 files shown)


jtbandes (approved):

packages/studio-base/src/panels/RawMessages/utils.ts
```diff
@@ -56,10 +56,14 @@ export function generateDeepKeyPaths(obj: unknown, maxArrayLength: number): Set<
...
     }
 
-    for (const key of Object.getOwnPropertyNames(nestedObj)) {
+    const propertyNames = Object.getOwnPropertyNames(nestedObj);
+    if (propertyNames.length > maxArrayLength) {
+      return;
+    }
+
+    for (const key of propertyNames) {
       const newPath = [key, ...path];
```
jhurliman, line 65: The spread operator does not have great performance, and if `recurseMapKeys` can work with a readonly array this could be tweaked to use a global `[]` that is mutated each iteration instead of reallocated. Uglier code, but more performant.

foxymiles: Yeah that's a good point. My instinct here is that the object traversal here is not the bottleneck in most cases compared to actually rendering rows in the JSON tree but would could make some improvements here too.

packages/studio-base/src/panels/RawMessages/Value.tsx
```diff
@@ -221,3 +221,5 @@ export default function Value(props: ValueProps): JSX.Element {
     </Stack>
   );
 }
+
+export default React.memo(Value);
```
defunctzombie, line 225: We need a comment if this is an important memoization. Any performance tweaks need comments.

defunctzombie (approved):

PR #5468, opened 2023-03-02
Fix sign in prompt in layout browser when app bar is enabled

foxymiles:
**User-Facing Changes**
Fix sign in prompt in layout browser when app bar is enabled

**Description**
Update logic to either show the sign in modal directly if app bar is enabled or show the sign in left sidebar item if not.


esthersweon:
FYI - Not including this in changelog, as top nav is feature-flagged.

PR #5405, opened 2023-02-22
Move add panel list into top bar

2metres:
**User-Facing Changes**
None

**Description**
Moves Add Panel List to top nav behind feature flag

PR #5469, opened 2023-03-02
Show user menu on click (instead of Add Panel) 🐞

2metres:
User menu shown instead of Add Panel bug introduced in #5405


amacneil (approved):

esthersweon:
Not adding to changelog since this is behind feature flag, FYI.

Issue #5455, opened 2023-03-01
Custom panel - Cannot subscribe to certain topics and republish that same message


joshnewans:
**Description**

- Version: 1.42.0
- Platform: Ubuntu 20.04 (Pop OS) x64

For certain message types (tested only with ROS 2 `sensor_msgs/Joy`), an attempt to republish a message received from a subscription will fail silently. Put another way, it appears the serialise/deserialise (if you can call it that) is not quite compatible for sub/pub.

As I understand it, in this case the message contains an array of floats and an array of ints. Foxglove parses these from the subscriber as a `Float32Array` and `Int32Array` respectively, but the publisher will only accept raw arrays. A simple JSON wrap doesn't work, as it converts the special types into objects with numerical keys rather than actual arrays. Here is my current workaround.

```
const tmpMsg = {
    header: latestMsg.header,
    axes: Array.from(latestMsg.axes),
    buttons: Array.from(latestMsg.buttons),
};
```

**Steps To Reproduce**
An example repo is provided below with details and instructions in the README. Please note that this repo also demonstrates some other issues I am having with publisher timing (I suspect that one is user error) so if any assistance can be provided on that too that would be appreciated.
https://github.com/joshnewans/fox-pub-error-mwe

Essentially, subscribe to a `sensor_msgs/Joy` topic e.g. `/joy` (that is fed by external data), advertise a new topic of that same type (e.g. `/joy2`), and when receiving a message publish it directly.

**Expected Behavior**

One would expect to receive messages on the new topic, but they will not appear. Instead a conversion like the one described above is required.

jtbandes:
Thanks for the report and detailed writeup. I've made a PR https://github.com/foxglove/studio/pull/5473 which I believe addresses the other issues you reported in the repo readme, but not this one (about TypedArrays).

I also tweaked your usage of React to avoid side effects in setState updater functions. The changes I used for testing were:

```diff
diff --git a/src/ExamplePanel.tsx b/src/ExamplePanel.tsx
index 5785736..5bd1afc 100644
--- a/src/ExamplePanel.tsx
+++ b/src/ExamplePanel.tsx
@@ -32,8 +32,6 @@ function ExamplePanel({ context }: { context: PanelExtensionContext }): JSX.Elem
   const [messages, setMessages] = useState<readonly MessageEvent<unknown>[] | undefined>();
   const [rawMsg, setRawMsg] = useState<Joy | undefined>();
[...]

PR #5473, opened 2023-03-02
Fix advertise/publish race condition

jtbandes:
**User-Facing Changes**
Fixed an issue where extension panels were unable to advertise and immediately publish on the same channel.

**Description**
Addresses some of the issues discussed in https://github.com/foxglove/studio/issues/5455 & https://github.com/joshnewans/fox-pub-error-mwe

The MessagePipeline keeps a list of `publishersById` which holds the publishers for each panel. It also merges them into a single list of publishers, and calls `player.setPublishers()` with the merged list. The merging was being done in a reducer, and the `setPublishers` call was happening in a `useEffect`. However, the `useEffect` could occur "too late" and a panel (such as https://github.com/joshnewans/fox-pub-error-mwe) could call `publish()` before the advertisement had actually been set on the player, which would throw an error.

To address the timing issue, I moved `setPublishers` out of a `useEffect` and it now happens synchronously during the `advertise` call. Since the list of publishers was not used outside of the MessagePipeline for any other purpose, I've also moved it out of the "`public`" state.

The Rosbridge player would also throw an error if a publish() occurred before the advertisements had been fully set up (after a connection was established). However, the panels are not aware of the state of the underlying websocket connection, so they have no way to avoid publishing before it's fully established. So I added logic to avoid throwing an error in this case. I wasn't able to reproduce the same issue with the websocket player so I didn't make any change to it.


packages/studio-base/src/components/MessagePipeline/index.test.tsx
```diff
@@ -484,30 +482,48 @@ describe("MessagePipelineProvider/useMessagePipeline", () => {
...
-    const lastPublishers = result.current.publishers;
+    const lastPublishers = player.publishers;
     // cause the player to emit a frame outside the render loop to trigger another render
     await doubleAct(async () => await player.emit());
     // make sure publishers are reference equal when they don't change
-    expect(result.current.publishers).toBe(lastPublishers);
+    expect(player.publishers).toBe(lastPublishers);
+  });
+
+  it("keeps publishers when changing player", () => {
```
snosenzo, line 509: a comment to explain why this is tested against would be useful, not super evident from title.

snosenzo (commented):
not very familiar with the zustand change implications, so might want to get some other eyes on that

packages/studio-base/src/components/MessagePipeline/index.test.tsx
jtbandes: updated the title, hopefully it is more clear now

snosenzo (approved):

packages/studio-base/src/components/MessagePipeline/store.ts
```diff
@@ -93,14 +95,14 @@ export function createMessagePipelineStore({
...
       messageEventsBySubscriberId: new Map(),
       subscriptions: [],
-      publishers: [],
       sortedTopics: [],
       datatypes: new Map(),
       setSubscriptions(id, payloads) {
         get().dispatch({ type: "update-subscriber", id, payloads });
       },
       setPublishers(id, payloads) {
         get().dispatch({ type: "set-publishers", id, payloads });
```
defunctzombie, line 104: What's the purpose of this dispatch now that the publishers are set internally?

packages/studio-base/src/components/MessagePipeline/store.ts
```diff
@@ -93,14 +95,14 @@ export function createMessagePipelineStore({
...
+        get().player?.setPublishers(get().allPublishers);
```
defunctzombie, line 105: How is `allPublishers` updated prior to calling `setPublishers` on the player?

packages/studio-base/src/components/MessagePipeline/store.ts
jtbandes: We still need to keep track of the state somewhere.

packages/studio-base/src/components/MessagePipeline/store.ts
jtbandes: It's updated as part of the reducer call from the `dispatch` above

jtbandes:
> what's with the // foxglove-depcheck-used: babel-plugin-transform-import-meta ? was this causing an issue previously?

I added `babel-plugin-transform-import-meta` to the studio-base deps to ease development in the private repo. That for some reason triggered the unused dependencies linter where previously it was not triggered.

packages/studio-base/src/components/MessagePipeline/store.ts
defunctzombie: I did some reading on this and I think dispatch returns the new state? It wasn't entirely clear to me. But if it does that might be a little clearer here. Something to be aware of for the future.

PR #5476, opened 2023-03-03
New UI: Show new left sidebar and hide old sidebar

2metres:
**User-Facing Changes**
None

**Description**
Update left nav to use new style sidebar under feature flag

image


packages/studio-base/src/Workspace.tsx
```diff
@@ -521,6 +521,12 @@ export default function Workspace(props: WorkspaceProps): JSX.Element {
     };
   }, [enableNewTopNav]);
 
+  const TopicListSidebar = useMemo(() => {
+    return function TopicListSidebarImpl() {
+      return <TopicList onSelectDataSourceAction={() => setShowOpenDialog({ view: "start" })} />;
+    };
+  }, []);
```
jtbandes, line 528: might be a little cleaner as useCallback rather than useMemo

2metres: We have a mixture of `useMemo` and `useCallback` here. Should I prefer one over the other?

jtbandes: Usually useCallback when the thing is a function and useMemo otherwise

jtbandes (changes requested):
Spacing adjustments make storybook changes hard to review :(
image

These stories look wrong:
image

2metres:
> These stories look wrong: image

What looks wrong about them? Should we add test for the new UI?

amacneil:
Panel settings does not open when I click a panel's cog icon.

jtbandes (approved):
Code changes looking good but a couple of the stories are missing their sidebar.

packages/studio-base/src/components/Sidebars/NewSidebar.stories.tsx
```diff
@@ -0,0 +1,179 @@
...
+  label,
+  defaultLeftKey,
+  defaultRightKey,
+  enableAppBar,
+  height,
+}: {
+  label?: string;
+  defaultLeftKey?: string | undefined;
+  defaultRightKey?: string | undefined;
+  enableAppBar?: boolean;
```
jtbandes, line 70: Looks like this prop is unused now

packages/studio-base/src/components/Sidebars/NewSidebar.stories.tsx
```diff
@@ -133,7 +127,7 @@ LeftClosed.play = async () => {
 };
 
 // Right
-export const RightOpen = (): JSX.Element => <Story defaultRightKey="x" />;
+export const RightOpen = (): JSX.Element => <Story />;
```
jtbandes, line 130: Is this story doing anything at all?

jtbandes (approved):

amacneil:
Product review feedback:
- Remove left sidebar "open data source" button for now
- Right hand sidebar needs to be wider by default

Follow up meeting about panel settings position

Issue #5371, opened 2023-02-14
Last seen image marker persists when turned off during playback


vivek-rane:
**Description**
When a 2D image marker is turned off during playback, the last seen marker continues to be drawn on the screen. This finally disappears when the marker is turned back on. This does not happen if the playback is paused.

- Version: 1.39.1 (darwin)
- Platform: macOS Monterey (12.5.1)

**Steps To Reproduce**
1. Download the test mcap from https://github.com/foxglove/mcap/blob/main/testdata/mcap/demo.mcap
2. Add the following user script:
```
import { Input, Message } from "./types";

export const inputs = ["/image_color/compressed"];
export const output = "test marker";

type ImageMarker = Message<"foxglove.ImageAnnotations">;

var color_val = { r: 1, g: 0, b: 0.5, a: 1 };

export default function node(
  event: Input<"/image_color/compressed">
): ImageMarker {
  return {
    circles: [
      {
        timestamp: event.receiveTime,
        thickness: 2,
        position: { x: (event.message.header.seq - 100) * 10, y: 300 },
        outline_color: color_val,
        fill_color: color_val,
        diameter: 100,
      },
    ],
    points: [],
  };
}
```
3. Open an image panel and point it to `/image_color/compressed`
4. Turn on the marker called "test marker" from our script
5. Start playback, and turn the marker off. The marker will continue to be displayed in the position it was when turned off.
6. Turning the marker back on will update it to the right place.

**Expected Behavior**
When marker is turned off, it should disappear from the image.

https://user-images.githubusercontent.com/25373098/218605788-5686b20c-a94e-40e8-8029-ec6b8ca4985a.mov

defunctzombie:
@vivek-rane We've transferred this issue to the Studio repo as this is a studio bug not an mcap bug.

PR #5478, opened 2023-03-04
Fix some image panel state issues

jtbandes:
**User-Facing Changes**
Fixed an issue where the Image panel would sometimes continue displaying annotations after they were disabled in the panel's settings.

**Description**
Addresses two issues:
1. As reported in #5371, the panel would sometimes display annotations after they were disabled. This happened because of a race condition around `context.subscribe()` (quite similar to #5473, but for subscribing rather than publishing), which causes the panel to occasionally receive a message in `currentFrame` after it has already unsubscribed. To work around this I added a check that the annotation's topic is in the list of `annotationTopics` the panel expects to receive. Also filed #5479 to follow up on the issue.

2. I also found that the panel would sometimes display an image and the "waiting for images" message at the same time. This was due to some logic that would display the `lastImageMessageRef.current` if no image is available. Reading `lastImageMessageRef.current` during render was wrong, since the modification of the ref does not always trigger the panel to re-render. Since the `useImagePanelMessages` hook now has responsibility for keeping the latest image message around, I removed uses of `lastImageMessageRef` other than for the "download image" feature.

Fixes https://github.com/foxglove/studio/issues/5371
Fixes FG-1924

### DIFF
--- a/packages/studio-base/src/panels/Image/hooks/useImagePanelMessages.ts
+++ b/packages/studio-base/src/panels/Image/hooks/useImagePanelMessages.ts
@@ -28,7 +28,7 @@ type UseImagePanelMessagesParams = {
 export type ImagePanelState = UseImagePanelMessagesParams & {
   image?: NormalizedImageMessage;
   cameraInfo?: CameraInfo;
-  annotationsByTopic: Map<string, Annotation[]>;
+  annotationsByTopic: ReadonlyMap<string, Annotation[]>;
   tree: AVLTree<Time, SynchronizationItem>;
 
   actions: {
@@ -86,8 +86,15 @@ function addMessages(
     return synchronizedAddMessages(state, messageEvents);
   }
 
-  let newState: Pick<ImagePanelState, "image" | "annotationsByTopic" | "cameraInfo"> | undefined;
+  let newState:
+    | (Pick<ImagePanelState, "image" | "cameraInfo"> & {
+        annotationsByTopic: Map<string, Annotation[]>;
+      })
+    | undefined;
   for (const event of messageEvents) {
+    // We have to check `event.topic` in all cases because we may occasionally receive messages on
+    // topics that we have already unsubscribed from with context.subscribe().
+    // <https://github.com/foxglove/studio/issues/5371>
     const normalizedCameraInfo =
       event.topic === state.cameraInfoTopic
         ? normalizeCameraInfo(event.message, event.schemaName)
@@ -96,7 +103,9 @@ function addMessages(
       event.topic === state.imageTopic
         ? normalizeImageMessage(event.message, event.schemaName)
         : undefined;
-    const normalizedAnnotations = normalizeAnnotations(event.message, event.schemaName);
+    const normalizedAnnotations = state.annotationTopics.includes(event.topic)
+      ? normalizeAnnotations(event.message, event.schemaName)
+      : undefined;
     if (!normalizedCameraInfo && !normalizedImage && !normalizedAnnotations) {
       continue;
     }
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/panels/Image/hooks/useImagePanelMessages.ts
```diff
@@ -86,8 +86,15 @@ function addMessages(
...
-  let newState: Pick<ImagePanelState, "image" | "annotationsByTopic" | "cameraInfo"> | undefined;
+  let newState:
+    | (Pick<ImagePanelState, "image" | "cameraInfo"> & {
+        annotationsByTopic: Map<string, Annotation[]>;
+      })
+    | undefined;
   for (const event of messageEvents) {
+    // We have to check `event.topic` in all cases because we may occasionally receive messages on
+    // topics that we have already unsubscribed from with context.subscribe().
+    // <https://github.com/foxglove/studio/issues/5479>
```
defunctzombie, line 97: Would you expect that we'd remove this additional topic check after fixing https://github.com/foxglove/studio/issues/5479 or is it worth having such a check as part of the panel logic regardless?

packages/studio-base/src/panels/Image/hooks/useImagePanelMessages.ts
```diff
@@ -138,14 +147,45 @@ export function useImagePanelMessages(params: UseImagePanelMessagesParams): Publ
...
+                ...newParams,
+              };
+            }
+
+            // Otherwise, simply clear the image and any annotations when settings
+            // change. More precise/intelligent/complicated logic could be written to keep the image
+            // hidden if a new annotation is enabled but a synchronized set is not available, and
+            // similarly to show the image if an annotation being disabled results in a synchronized
+            // set being available. For now, we just clear everything.
+            return {
```
defunctzombie, line 181: We were previously clearing everything before but I'm not following why its ok to clear everything when we change one param. i.e. if I toggle an annotation off why does that clear the image?

packages/studio-base/src/panels/Image/hooks/useImagePanelMessages.ts
```diff
@@ -138,14 +147,45 @@ export function useImagePanelMessages(params: UseImagePanelMessagesParams): Publ
...
-            annotationsByTopic: new Map(),
-            tree: new AVLTree(compareTime),
-
-            ...newParams,
+          set((prevState) => {
+            // Optimize for the common case of toggling annotations on/off while the synchronize
+            // setting is disabled. As long as the image and camera info topics are the same, we can
+            // keep the existing image and need only rebuild the annotationsByTopic.
+            if (
+              !prevState.synchronize &&
```
defunctzombie, line 155: I struggle with reading multi-conditional if statements and would find this easier to read if the first two parts of this conditional were hoisted to a variable `const synchronizeEnabled = ...`.

packages/studio-base/src/panels/Image/hooks/useImagePanelMessages.ts
```diff
@@ -138,14 +147,45 @@ export function useImagePanelMessages(params: UseImagePanelMessagesParams): Publ
...
+              !newParams.synchronize &&
+              prevState.imageTopic === newParams.imageTopic &&
+              prevState.cameraInfoTopic === newParams.cameraInfoTopic
+            ) {
+              let newAnnotationsByTopic: Map<string, Annotation[]> | undefined;
+              if (prevState.annotationTopics !== newParams.annotationTopics) {
```
defunctzombie, line 161: Could this be flipped around to bail early and reduce the nesting in this logic and aid with following the code?

Or maybe some inner logic could be folded into a separate function like `rebuildAnnotationsByTopic` ?

defunctzombie (approved):

packages/studio-base/src/panels/Image/hooks/useImagePanelMessages.ts
jtbandes: It's not ideal but at least that now only happens when synchronization is enabled, previously it happened all the time.

packages/studio-base/src/panels/Image/hooks/useImagePanelMessages.ts
jtbandes: Could go either way, it is a bit weird that the panel was running `normalizeAnnotations()` without knowing for sure that the topic was one of the selected annotation topics – but the normalizeAnnotations function does a schemaName check so it should be alright without it.

PR #5480, opened 2023-03-04
Reorganize web package for reuse in private repo

jtbandes:
**User-Facing Changes**
None (internal package organization only)

**Description**
Moves most of the code and webpack configuration from `web` into a new workspace package `@foxglove/studio-web`. The new package is imported by `web` and can also be imported by the private repo that builds the production version of studio, to reduce duplication.


web/src/entrypoint.tsx
```diff
@@ -0,0 +1,7 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
defunctzombie, line 1: What's the purpose of this file relative to `main.ts` ?

defunctzombie (commented):
Feels to me like `<Root>` could possibly go away. It was originally meant as the top most component but looking at what is in main.tsx around it, doesn't seem like there's much purpose to it as a thin wrapper for `<App>`.

I don't follow how this change allows for re-use of anything tho? Nothing was hoisted out of the `web` folder into `studio-base`. I do not think our private package should depend on the `web` package. Maybe this is more obvious if I describe web as the _deployment_ package rather than a set of shared components.

web/src/entrypoint.tsx
jtbandes: This is the webpack entry point. `main.ts` no longer has side effects but just exports code so it can be invoked by the entry point.

web/src/entrypoint.tsx
```diff
@@ -0,0 +1,7 @@
...
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { main } from "@foxglove/studio-web";
```
defunctzombie, line 5: Are we really going to be able to share all of main? When we want to use different data sources this is not true. I think this PR is trying to consolidate too much into a "main".

packages/studio-web/src/webpackConfigs.ts
```diff
@@ -0,0 +1,153 @@
...
+    const plugins: WebpackPluginInstance[] = [];
+
+    if (isServe) {
+      plugins.push(new ReactRefreshPlugin());
+    }
+
+    // Source map upload if configuration permits
+    if (
+      !isDev &&
+      process.env.SENTRY_AUTH_TOKEN != undefined &&
```
defunctzombie, line 75: We can remove all this from this repo and move it to the other one.

packages/studio-web/src/webpackConfigs.ts
```diff
@@ -0,0 +1,153 @@
...
+        // Output filenames should include content hashes in order to cache bust when new versions are available
+        filename: isDev ? "[name].js" : "[name].[contenthash].js",
+
+        path: params.outputPath,
+      },
+
+      plugins: [
+        ...plugins,
+        ...(appWebpackConfig.plugins ?? []),
+        new EnvironmentPlugin(buildEnvironmentDefaults(argv.env?.FOXGLOVE_BACKEND ?? argv.mode)),
```
defunctzombie, line 117: We don't need this in this repo.

.eslintrc.yaml
```diff
@@ -151,7 +151,7 @@ overrides:
       no-restricted-imports: off
     files:
       # Ignore existing implementations of @mui/styled
-      - "web/src/VersionBanner.tsx"
+      - "packages/studio-web/src/VersionBanner.tsx"
```
defunctzombie, line 154: We have a new version of this somewhere that has been fixed do we not? We should copy that version here and remove this ignore.

defunctzombie (commented):
Trending in the right direction but still attempts to re-use more than I think we should. We can sync on this to land at the right place.

Stuff like the unavailable data sources should be moved out of this repo. And items like `IdbLayoutStorage` could easily go into `studio-base`.

web/src/entrypoint.tsx
jtbandes: That’s why I added an argument that allows you to pass in different data sources

packages/studio-web/src/webpackConfigs.ts
jtbandes: Would you rather I do that in this PR? Feels like it would add to the noise but I can do it.

packages/studio-web/src/index.tsx
```diff
@@ -58,7 +25,44 @@ function LogAfterRender(props: React.PropsWithChildren<unknown>): JSX.Element {
...
+};
+
+export async function main(params: MainParams = { appContext: {} }): Promise<void> {
+  log.debug("initializing");
+
+  window.onerror = (...args) => {
+    console.error(...args);
+  };
+
+  if (typeof process.env.SENTRY_DSN === "string") {
```
defunctzombie, line 41: delete sentry from here

defunctzombie (approved):

packages/studio-web/src/index.tsx
```diff
@@ -93,21 +97,15 @@ async function main() {
...
-      [AppSetting.SHOW_DEBUG_PANELS]: isDevelopment,
-    },
-  });
-
   ReactDOM.render(
     <StrictMode>
       <LogAfterRender>
         {banner}
-        <Root appConfiguration={appConfiguration} />
+        <AppContext.Provider value={params.appContext}>
```
defunctzombie, line 104: OSS doesn't need to do anything with this

packages/studio-web/src/webpackConfigs.ts
jtbandes: 👍 once desktop is deduped too, the buildEnvironmentDefaults function can be removed from oss

PR #5481, opened 2023-03-04
Added a First Draft of a Custom Map Max Zoom Option

blainemhuey:
**User-Facing Changes**
Adds a drop-down box for custom map tile sources to choose the maximum tile level.

**Description**
This pull request is for a way for the 'maxNativeZoom' parameter to be user-configurable for custom tile layers.  This feature would benefit the application since some mapping sources provide more high-resolution tiles than the current hard-coded zoom limit of 18 that is applied to the leafletjs TileLayer map object.  In my particular use case of Foxglove, the data for the mapping server I am using goes up to a zoom of 22, so this change would greatly improve the mapping detail.  This issue seems to have been brought up in different ways by other users in the past, such as with the foxglove discussion 296 linked below [1] and a previous merged pull request [2] where the max native zoom level was first set to 18. 

However, I do acknowledge that this commit is likely in need of many improvements and changes, since I am both not particularly familiar with TypeScript or the project's standards of practice, and so I welcome any advice for how to proceed.

1. https://github.com/orgs/foxglove/discussions/296
2. https://github.com/foxglove/studio/pull/610

### DIFF
--- a/packages/studio-base/src/panels/Map/MapPanel.tsx
+++ b/packages/studio-base/src/panels/Map/MapPanel.tsx
@@ -104,6 +104,7 @@ function MapPanel(props: MapPanelProps): JSX.Element {
       layer: initialConfig.layer ?? "map",
       topicColors: initialConfig.topicColors ?? {},
       zoomLevel: initialConfig.zoomLevel,
+      maxNativeZoom: initialConfig.maxNativeZoom ?? 18,
     };
   });
 
@@ -259,6 +260,13 @@ function MapPanel(props: MapPanelProps): JSX.Element {
       });
     }
 
+    if (path[1] === "maxNativeZoom" && input === "select") {
+      setConfig((oldConfig) => {
+        const zoom = parseInt(String(value));
+        return { ...oldConfig, maxNativeZoom: isNaN(zoom) ? oldConfig.maxNativeZoom : zoom };
+      });
+    }
+
     if (path[1] === "followTopic" && input === "select") {
       setConfig((oldConfig) => {
         return { ...oldConfig, followTopic: String(value) };
@@ -293,6 +301,12 @@ function MapPanel(props: MapPanelProps): JSX.Element {
       customLayer.setUrl(config.customTileUrl);
     }
   }, [config.layer, config.customTileUrl, customLayer]);
+
+  useEffect(() => {
+    if (config.layer === "custom") {
+      customLayer.options.maxNativeZoom = config.maxNativeZoom;
+    }
+  }, [config.layer, config.maxNativeZoom, customLayer]);
 
   // Subscribe to eligible and enabled topics
   useEffect(() => {
--- a/packages/studio-base/src/panels/Map/MapPanel.tsx
+++ b/packages/studio-base/src/panels/Map/MapPanel.tsx
@@ -263,7 +263,7 @@ function MapPanel(props: MapPanelProps): JSX.Element {
     if (path[1] === "maxNativeZoom" && input === "select") {
       setConfig((oldConfig) => {
         const zoom = parseInt(String(value));
-        return { ...oldConfig, maxNativeZoom: isNaN(zoom) ? oldConfig.maxNativeZoom : zoom };
+        return { ...oldConfig, maxNativeZoom: isFinite(zoom) ? zoom : oldConfig.maxNativeZoom };
       });
     }
 
--- a/packages/studio-base/src/panels/Map/config.ts
+++ b/packages/studio-base/src/panels/Map/config.ts
@@ -16,6 +16,7 @@ export type Config = {
   layer: string;
   topicColors: Record<string, string>;
   zoomLevel?: number;
+  maxNativeZoom?: number;
 };
 
 export function validateCustomUrl(url: string): Error | undefined {
@@ -110,6 +111,15 @@ export function buildSettingsTree(
       input: "string",
       value: config.customTileUrl,
       error,
+    };
+
+    generalSettings.maxNativeZoom = {
+      label: "Custom max zoom",
+      input: "select",
+      value: config.maxNativeZoom,
+      options: [18, 19, 20, 21, 22, 23, 24].map((num) => {
+        return { label: String(num), value: num };
+      }),
     };
   }
 
--- a/packages/studio-base/src/panels/Map/config.ts
+++ b/packages/studio-base/src/panels/Map/config.ts
@@ -114,12 +114,13 @@ export function buildSettingsTree(
     };
 
     generalSettings.maxNativeZoom = {
-      label: "Custom max zoom",
+      label: "Max source zoom",
       input: "select",
       value: config.maxNativeZoom,
       options: [18, 19, 20, 21, 22, 23, 24].map((num) => {
         return { label: String(num), value: num };
       }),
+      help: "Highest zoom supported by the custom map source. See https://leafletjs.com/examples/zoom-levels/ for more information.",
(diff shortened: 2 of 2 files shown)


packages/studio-base/src/panels/Map/config.ts
```diff
@@ -111,6 +112,15 @@ export function buildSettingsTree(
       value: config.customTileUrl,
       error,
     };
+
+    generalSettings.maxNativeZoom = {
```
defunctzombie, line 116: @foxymiles we probably want to only show this when the custom layer option is selected?

packages/studio-base/src/panels/Map/config.ts
```diff
@@ -111,6 +112,15 @@ export function buildSettingsTree(
...
+      label: "Custom max zoom",
```
defunctzombie, line 117: I'm thinking we can drop the word `Custom` from this.

Is there a help text we can add that can link to some rough explanation for users on what "zoom" means here or is this obvious to anyone selecting a custom layer?

We should default this value to whatever we have it set to by default today so when you select custom layer you see this setting show up with the `max zoom` pre-filled to our chosen default.

defunctzombie:
Thanks for the contribution! I left a few notes for us to discussed some approaches but overall what you have looks reasonable to me. Is it worth having a minZoomLevel for custom layers as well?

packages/studio-base/src/panels/Map/config.ts
blainemhuey: At least from my testing, 18 is selected as the default dropdown value because of the config change, which does align with the current hard-coded max native zoom.

As for the wording of the text, I agree that that could use some updating and clarification.  Would something like 'Max source zoom level' be more clear?  Additionally, if we want to provide some background on the meaning of zoom level, there are some sources we could reference to explain, such as:
1. The library used for the map tile panel: https://leafletjs.com/examples/zoom-levels/
2. OpenStreetMap's description with a handy table: https://wiki.openstreetmap.org/wiki/Zoom_levels

However, I am unclear about what you mean by "help text".  In terms of the UI, do you mean something like hover text for the zoom dropdown box with more info?

defunctzombie: > However, I am unclear about what you mean by "help text". In terms of the UI, do you mean something like hover text for the zoom dropdown box with more info?

Exactly. We have a `help` property on settings fields: https://github.com/foxglove/studio/blob/main/packages/studio/src/index.ts#L568 which shows up on hover of the field.

blainemhuey: Ah, I totally missed that property!  I've pushed an updated version with a brief help blurb.

blainemhuey:
> Thanks for the contribution! I left a few notes for us to discussed some approaches but overall what you have looks reasonable to me. Is it worth having a minZoomLevel for custom layers as well?

I'm not sure if this would be as helpful as being able to specify the max native zoom.  Since increasing zoom levels require exponential amounts of space on the map tile server for the same real-world area, including most or all of the lower zoom levels is comparatively easy.  It might just be an easy assumption to make that any lower zoom levels needed by the user  when using a custom map should be covered by the user.  However, I could imagine a circumstance where someone would have mapping information for only a small range of zoom levels would want this, but I don't think that is necessary to implement now unless other users have requested it.

packages/studio-base/src/panels/Map/config.ts
foxymiles: Looks like the code already handles this.

packages/studio-base/src/panels/Map/MapPanel.tsx
```diff
@@ -259,6 +260,13 @@ function MapPanel(props: MapPanelProps): JSX.Element {
       });
     }
 
+    if (path[1] === "maxNativeZoom" && input === "select") {
+      setConfig((oldConfig) => {
+        const zoom = parseInt(String(value));
+        return { ...oldConfig, maxNativeZoom: isNaN(zoom) ? oldConfig.maxNativeZoom : zoom };
```
foxymiles, line 266: I suggest using `isFinite` here and reversing the ternary. `isNan` will allow some values we don't want like `Infinity`.

foxymiles (approved):
Looks good to me. Left a minor comment on numeric value handling.

packages/studio-base/src/panels/Map/MapPanel.tsx
blainemhuey: Ok, I've made that change and updated the code in my most recent commit.

blainemhuey:
> Did another pass on the PR. Thoughts on changing the `maxNativeZoom` config option and setting to `maxZoom` or `maxZoomLevel`. I see we already have a `zoomLevel` and this would pair nicely with that terminology. Is there a specific meaning behind `Native` in this context?

The meaning of native is the level of zoom that is supported by the backend map tile server.  At least in leafletjs, the library used for the map, there is another parameter called 'maxZoom' which determines how far the user is allowed to zoom in.  In the current version of Foxglove studio, the maxNativeZoom is harcoded to 18 while maxZoom is harcoded to 24, so while the map supports zooming in beyond level 18, it will not fetch any additional, higher resolution data but instead interpolate/upsample the level 18 resolution data.

I initially stuck with the 'native' terminology for this PR, although I am not sure how clear it would be a user of foxglove, since maxZoom might be more clear even if it conflicts with the way leafletjs describes its parameters.

defunctzombie:
I see that leaflet calls the option maxNativeZoom which I can't say seems "obvious" to me as a user.

Here are the leaflet docs for this option:

> Maximum zoom number the tile source has available. If it is specified, the tiles on all zoom levels higher than maxNativeZoom will be loaded from maxNativeZoom level and auto-scaled.

Maybe these docs provide a clue to what a good name for our UI and config option could be. An idea that comes to mind is `Max tile level`.

Given that leaflet calls the config option `maxNativeZoom` I would be ok with that being the internal config name with a different user-facing label. But I would also be ok with the label and config name matching. It is easy for us to update the label and less easy to update the config field name.

blainemhuey:
> I see that leaflet calls the option maxNativeZoom which I can't say seems "obvious" to me as a user.
> 
> Here are the leaflet docs for this option:
> 
> > Maximum zoom number the tile source has available. If it is specified, the tiles on all zoom levels higher than maxNativeZoom will be loaded from maxNativeZoom level and auto-scaled.
> 
> Maybe these docs provide a clue to what a good name for our UI and config option could be. An idea that comes to mind is `Max tile level`.
> 
> Given that leaflet calls the config option `maxNativeZoom` I would be ok with that being the internal config name with a different user-facing label. But I would also be ok with the label and config name matching. It is easy for us to update the label and less easy to update the config field name.

I would lean towards using 'maxNativeZoom' for the internal config name since it is more specific and it leaves open the possibility of someday having a configurable maximum zoom level using the same naming convention (i.e 'maxZoom') separate from the tile level.  If that sounds good, I can push a commit updating the name in the UI to 'Max tile level' while leaving the internal config name the same.

defunctzombie:
> I would lean towards using 'maxNativeZoom' for the internal config name since it is more specific and it leaves open the possibility of someday having a configurable maximum zoom level using the same naming convention (i.e 'maxZoom') separate from the tile level. If that sounds good, I can push a commit updating the name in the UI to 'Max tile level' while leaving the internal config name the same.

Sounds good to me.

defunctzombie (approved):

defunctzombie:
Thank you for contributing!

PR #5483, opened 2023-03-06
Improve whitespace handling of TextMiddleTruncate component

foxymiles:
**User-Facing Changes**
Cosmetic improvements to the `TextMiddleTruncate` component.

**Description**
Improve whitespace handling of TextMiddleTruncate component by avoiding splits at or immediately after whitespace in the string.


jtbandes:
Could this be fixed by only adjusting the `white-space` CSS property rather than changing the JS logic?

For example, looks like `pre` could work: 
image


Here's another test case you can use, which seems to repro the problem even when the window is not narrow. The string is `ws://192.168.88.132:62338\niOS Foxglove Bridge`.

image
image

Although playing around with the `pre` value, it surprisingly turns into this in the title bar:
image

So might need to do something about splitting at newlines like we did in the sidebar...

jtbandes (approved):

foxymiles:
I tried removing the JS logic here and experimenting with various values for `white-space` but none of them gave the desired effect. Did you have a chance to try this yourself?

jtbandes:
Nope I did not spend much time with it. Change seems fine as is.

PR #5488, opened 2023-03-06
Apply background blur to tooltips

2metres:
**User-Facing Changes**
Apply background blur to tooltips for browsers that support it.

**Description**
Screen Shot 2023-03-07 at 9 52 13 am
Screen Shot 2023-03-07 at 9 53 35 am

Screen Shot 2023-03-07 at 9 52 21 am

### DIFF
--- a/packages/studio-base/src/theme/muiComponents.ts
+++ b/packages/studio-base/src/theme/muiComponents.ts
@@ -517,11 +517,11 @@ export default function muiComponents(theme: Theme): Theme["components"] & MuiLa
       styleOverrides: {
         arrow: {
           color: tinycolor(theme.palette.grey[700]).setAlpha(0.86).toRgbString(),
-          backdropFilter: "blur(10px)",
+          backdropFilter: "blur(3px)",
         },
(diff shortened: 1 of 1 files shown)


jtbandes (approved):
Sure why not. Maybe a (subtle) border & drop shadow would fit in nicely too?

defunctzombie:
Personally I find stuff like this makes readability worse. Even in the provided screenshots I am distracted.

jtbandes:
I don't get your complaint – it's somehow more distracting than showing the un-blurred content through the tooltips?

amacneil:
2c - I think it looks good on the plot tooltip. I'm not yet convinced that tooltip is the right solution for the data source info (and it is still behind feature flag), so I don't feel strongly either way on that one.

defunctzombie:
> I don't get your complaint – it's somehow more distracting than showing the un-blurred content through the tooltips?

Maybe I did not realize we were showing the un-blurred content? In my mind the tooltips had a solid background and none of the content bled through.

PR #5489, opened 2023-03-06
Fix handling of out of bounds urlState time values

foxymiles:
**User-Facing Changes**
Fix handling of out of bounds urlState time values.

**Description**
This makes two fixes:
1. Clamps the ranges supplied to the ProgressPlot component so it doesn't draw out of bounds.
2. Clamps the seek time in the `IterablePlayer` seekPlayback method within start & end time.

### DIFF
--- a/packages/studio-base/src/components/PlaybackControls/ProgressPlot.tsx
+++ b/packages/studio-base/src/components/PlaybackControls/ProgressPlot.tsx
@@ -4,6 +4,7 @@
 
 import { keyframes } from "@emotion/react";
 import { simplify } from "intervals-fn";
+import { clamp } from "lodash";
 import { useMemo } from "react";
 import tinycolor from "tinycolor2";
 import { makeStyles } from "tss-react/mui";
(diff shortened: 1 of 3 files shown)


jtbandes (approved):
I think this is a helpful change, but not entirely sufficient. Something in the app was allowing the `&time` in the url to overwrite the data source's actual time bounds, which shouldn't be allowed regardless of whether it's being displayed in a ProgressPlot or some other kind of UI element.

foxymiles:
@jtbandes I've also added a fix here to `IterablePlayer` which should handle the case of clamping seek time during initialization.

foxymiles:
Incorporated some suggestions from @defunctzombie to make this more robust.

jtbandes (approved):

jtbandes:
Forgot to ask earlier but are there some unit tests we can add around this for the player?

foxymiles:
> Forgot to ask earlier but are there some unit tests we can add around this for the player?

Yes we probably should. I'll try to work that in this week.

PR #5498, opened 2023-03-07
Respect opacity in model color override

jtbandes:
**User-Facing Changes**
Fixed a bug where the alpha component of the override color setting was not respected for ModelEntities.

**Description**
Fixes FG-2292

image

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableModels.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableModels.ts
@@ -289,6 +289,7 @@ export class RenderableModels extends RenderablePrimitive {
       }
       rgbToThreeColor(renderable.material.color, overrideColor);
       const transparent = overrideColor.a < 1;
+      renderable.material.opacity = overrideColor.a;
       renderable.material.transparent = transparent;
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableModels.ts
```diff
@@ -289,6 +289,7 @@ export class RenderableModels extends RenderablePrimitive {
       }
       rgbToThreeColor(renderable.material.color, overrideColor);
       const transparent = overrideColor.a < 1;
+      renderable.material.opacity = overrideColor.a;
       renderable.material.transparent = transparent;
```
defunctzombie, line 293: Should transparent be forced true if alpha is < 1 in overrideColor?

defunctzombie (approved):

snosenzo (approved):

packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableModels.ts
defunctzombie: 🤦 I only looked at the added line and missed the one above

PR #5499, opened 2023-03-07
Fix issue with Selected Point button positioning

2metres:
**User-Facing Changes**
Selected point now appears in the not incorrect place

**Description**
Fix issue with Selected Point button positioning and refactor to use makeStyles

Screen Shot 2023-03-07 at 3 29 44 pm

### DIFF
--- a/packages/studio-base/src/panels/Image/components/Toolbar.tsx
+++ b/packages/studio-base/src/panels/Image/components/Toolbar.tsx
@@ -3,9 +3,10 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
 import { Cursor24Regular } from "@fluentui/react-icons";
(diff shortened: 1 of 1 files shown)


jtbandes:
"refactor to use makeStyles" is not a user-facing change

defunctzombie (approved):
"correct place" has a pretty open meaning :D 

I personally find it "incorrect" that two icons are at separate corners.

2metres:
Updated comment to satify @defunctzombie

jtbandes:
@2metres User-facing changes is "cute" but please remember that this section of the PR is gathered up and put into our release notes that our customers will read. So it's better to put nothing at all than something that will confuse people. Keep the "cute" comments in the Description section 😄

PR #5502, opened 2023-03-07
Restyle tabs

2metres:
**User-Facing Changes**
Restyle tab panel tabs to be more minimal

**Description**
Screen Shot 2023-03-07 at 9 51 33 pm


packages/studio-base/src/panels/Tab/TabbedToolbar.tsx
```diff
@@ -25,25 +30,17 @@ import {
 } from "@foxglove/studio-base/panels/Tab/TabDndContext";
 import { TabConfig } from "@foxglove/studio-base/types/layouts";
 
-const STabbedToolbar = muiStyled("div")(({ theme }) => ({
```
jtbandes, line 28: 🔪

jtbandes (approved):
👏

packages/studio-base/src/panels/Tab/TabbedToolbar.tsx
2metres: 🗡️

PR #5503, opened 2023-03-07
Remove file info from top of DataSourceInfoPanel

2metres:
**User-Facing Changes**
Remove file info from top of DataSourceInfoPanel as it appears elsewhere in the app and takes up a significant amount of space.

| Before |. After |
| --- | --- |
|Screen Shot 2023-03-08 at 8 40 12 am | Screen Shot 2023-03-08 at 8 37 16 am |

Closes FG-1835

### DIFF
--- a/packages/studio-base/src/panels/DataSourceInfo/index.tsx
+++ b/packages/studio-base/src/panels/DataSourceInfo/index.tsx
@@ -2,11 +2,9 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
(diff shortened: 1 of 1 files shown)


jtbandes:
What's the rationale for this change?

amacneil:
Agree PR description needs motivation/justification.

I think this is a good change because (a) the top half of Data Source Info panel wastes a ton of space, and (b) that info is also in the sidebar. 

I would consider renaming the panel to `Topic List` if we make this change.

amacneil:
🚢

amacneil:
Add a 1px line between the panel toolbar and the table header

defunctzombie (approved):

PR #5508, opened 2023-03-08
Improve performance of map panel

r0gi:
**User-Facing Changes**
Map panel performance improved, especially when seeking through large bag files with many GPS points.

**Description**
I noticed bag files were playing back quite poorly when the MapPanel had many GPS points on it. This is due to the fact that [this setState call](https://github.com/foxglove/studio/blob/main/packages/studio-base/src/panels/Map/MapPanel.tsx#L414) actually creates a new object for `allMapMessages`, so [this useMemo ](https://github.com/foxglove/studio/blob/main/packages/studio-base/src/panels/Map/MapPanel.tsx#L143) is actually not doing anything, since the object reference changes every time. It then leads to [this expensive useEffect hook](https://github.com/foxglove/studio/blob/main/packages/studio-base/src/panels/Map/MapPanel.tsx#L507) getting called redundantly when `allMapMessages` contents haven't actually changed (very noticeable in large bag files with a lot of GPS data).

The change basically implements what's already done a few lines above for the `setTopics` call. Note that this could theoretically make performance slightly worse if `allMapMessages` is very large and the contents _do_ change every time (but by that point performance might be bottlenecked in other ways anyway).

Before:
Screenshot from 2023-03-08 16-33-36
After:
Screenshot from 2023-03-08 16-34-31


r0gi:
Here is a sample bag file with 8000 randomly generated GPS points (4000 in `/gps/fix` and 4000 in `/gps/fix2`). You can open the current studio https://studio.foxglove.dev/ and the vercel deployed one https://studio-qjhyiw7cx.foxglove.party/ side by side and compare for yourself. I had to zip the bag file because github doesn't support .bag files.

[map_panel_test.zip](https://github.com/foxglove/studio/files/10926581/map_panel_test.zip)

foxymiles:
Thanks for the fix. This makes sense. I wonder if we could also skip running this `isEqual` on the messages if we memoized the filtered messages in a separate `useMemo`?

foxymiles:
@r0gi For example something like this:

https://github.com/foxglove/studio/tree/miles/fix-map-panel-all-messages-performance

I think this should be even faster since it can skip the work of running `isEqual` on all the messages.

foxymiles:
Thanks very much for the extra testing and detailed report @r0gi. Since you did all the hard work of diagnosing and measuring this issue, would you consider incorporating my change into your branch and then we can merge this PR?

foxymiles (approved):

defunctzombie:
Thanks for the change here. I think performance enhancements like this are worth a code comment so we keep track of what to avoid in the future. It seems like the key insight here was that `setAll` should avoid setting an un-stable value even once `allFrames` is no longer changing by not running `.filter` to re-create a new array value.

packages/studio-base/src/panels/Map/MapPanel.tsx
```diff
@@ -411,7 +416,7 @@ function MapPanel(props: MapPanelProps): JSX.Element {
       }
 
       if (renderState.allFrames) {
```
r0gi, line 418:
```suggestion
      if (renderState.allFrames) {
      // use memoized value to avoid unnecessary calls of expensive dependent functions
```
@defunctzombie something like this?

defunctzombie:
```suggestion
      if (renderState.allFrames) {
      // use memoization to avoid re-filtering allFrames when it has not changed
```

packages/studio-base/src/panels/Map/MapPanel.tsx
```diff
@@ -411,7 +416,8 @@ function MapPanel(props: MapPanelProps): JSX.Element {
...
-        setAllMapMessages(renderState.allFrames.filter(isValidMapMessage));
+      // use memoization to avoid re-filtering allFrames when it has not changed
```
defunctzombie, line 419:
```suggestion
        // use memoization to avoid re-filtering allFrames when it has not changed
```

defunctzombie:
Thank you for surfacing this issue and providing a fix!

Issue #5487, opened 2023-03-06
Registering Multiple Message Converters Not Handled Correctly


turtle-operator:
**Description**

When registering two message conversions from the same input message type to two different output message types (`foxglove.SceneUpdate` and `tf2_msgs/TFMessage`), only the conversion registered last is being handled. This can be tested by changing the order of the conversion registrations.

- Version: 1.42.1
- Platform: Ubuntu

**Steps To Reproduce**

I have something like the following extension:

```
export function activate(extensionContext: ExtensionContext): void {
  extensionContext.registerMessageConverter({
   fromSchemaName: 'my_msgs/CustomMessage',
   toSchemaName: 'tf2_msgs/TFMessage',
   converter: (message: Any) => {
     console.log("Converting to TFMessage");
     return ... // some code to provide transforms
   },
  });

  extensionContext.registerMessageConverter({
    fromSchemaName: 'my_msgs/CustomMessage',
    toSchemaName: 'foxglove.SceneUpdate',
    converter: (message: Any) => {
      console.log("Converting to SceneUpdate");
      return {
        deletions: ...
        entities: ...
      };
    },
  });
}

```

In this example, only the conversion to `foxglove.SceneUpdate` is actually processed. This also becomes obvious by checking for the logs in the dev tools console.

**Expected Behavior**

Both conversions are being taken care of (i.e., I get transforms and scene updates, and also I see both log messages in the dev tools console).

PR #5510, opened 2023-03-08
Fix subscribing to multiple converters in single panel

defunctzombie:
**User-Facing Changes**
Fix a bug which prevented extension panels from subscribing to multiple converters for the same topic.

**Description**
`renderState` is responsible for figuring out the `currentFrame` and `allFrames` fields for extension panels. Part of this logic tracks which subscriptions need to leverage message converters and runs the converters. It would track this information by storing a map from the topic name to the target schema.

Unfortunately, this architecture meant that the panel could only subscribe to one converter per topic. If the user installed an extension which registered two converters, each for the same _from_ schema but different _to_ schemas, the panel could not subscribe to both _to_ schemas as one converter subscription would override the other.

This change fixes this behavior by tracking all of the subscribed converters for a specific input topic and schema. Now if a panel subscribes to the same topic with different convertTo parameters, both of these subscriptions are honored and the panel will receive both converted messages.

Fixes: #5487
Fixes: FG-2321

### DIFF
--- a/packages/studio-base/src/components/PanelExtensionAdapter/renderState.ts
+++ b/packages/studio-base/src/components/PanelExtensionAdapter/renderState.ts
@@ -80,6 +80,15 @@ function initRenderStateBuilder(): BuildRenderStateFn {
       watchedFields,
     } = input;
 
+    // Should render indicates whether any fields of render state are updated
+    let shouldRender = false;
+
+    // Hoisted active data to shorten some of the code below that repeatedly uses active data
+    const activeData = playerState?.activeData;
+
+    // The render state starts with the previous render state and changes are applied as detected
+    const renderState: RenderState = prevRenderState;
+
     // If the player has loaded all the blocks, the blocks reference won't change so our message
     // pipeline handler for allFrames won't create a new set of all frames for the newly
     // subscribed topic. To ensure a new set of allFrames with the newly subscribed topic is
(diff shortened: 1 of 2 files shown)


foxymiles (approved):
Makes sense. 

This function is becoming very hard to reason about. At some point it would be worth looking into breaking it up into some pure functional pieces IMO.

packages/studio-base/src/components/PanelExtensionAdapter/renderState.ts
```diff
@@ -94,42 +99,85 @@ function initRenderStateBuilder(): BuildRenderStateFn {
...
+        // We've already stored a converter for this topic to convertTo
+        const haveConverter = existingConverters?.find(
+          (conv) => conv.toSchemaName === subscription.convertTo,
+        );
+        if (haveConverter) {
+          continue;
+        }
+
+        // Find a converter that can go from the original topic schema to the target schema
+        const converter = messageConverters?.find(
```
snosenzo, line 169: why does it only look like we're looking for one converter rather than all of the ones that exist for this particular topic/subscription?

defunctzombie: We only allow a single converter for a given pair of from/to, you cannot register multiple converters for the _same_ from/to tuple. I could pre-cache this in a map where the key is `from + to` if that would be clearer.

snosenzo: oh it looks like each subscription just represents one topic and convertTo pair, so there would presumably be multiple subscriptions for a single topic with different `convertTo` values?

defunctzombie: I've decided to add a comment instead of a pre-cached map. There's no other place to use the map other than this lookup here and I generally expect the list of converters to be on the maximum order of 10s.

PR #5511, opened 2023-03-08
foxglove websocket: Add support for `schemaEncoding` field

achim-k:
**User-Facing Changes**
- foxglove websocket: Add support for `schemaEncoding` field

**Description**
Adds support for a channel's optional `schemaEncoding` field which was added in https://github.com/foxglove/ws-protocol/pull/385.

### DIFF
--- a/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
+++ b/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
@@ -318,26 +318,29 @@ export default class FoxgloveWebSocketPlayer implements Player {
         try {
           let schemaEncoding;
           let schemaData;
-          if (channel.encoding === "json") {
+          if (channel.schemaEncoding === "jsonschema" || channel.encoding === "json") {
             schemaEncoding = "jsonschema";
             schemaData = textEncoder.encode(channel.schema);
-          } else if (channel.encoding === "protobuf") {
+          } else if (channel.schemaEncoding === "protobuf" || channel.encoding === "protobuf") {
             schemaEncoding = "protobuf";
             schemaData = new Uint8Array(base64.length(channel.schema));
             if (base64.decode(channel.schema, schemaData, 0) !== schemaData.byteLength) {
               throw new Error(`Failed to decode base64 schema on channel ${channel.id}`);
             }
-          } else if (channel.encoding === "flatbuffer") {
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
```diff
@@ -318,26 +318,29 @@ export default class FoxgloveWebSocketPlayer implements Player {
         try {
           let schemaEncoding;
           let schemaData;
-          if (channel.encoding === "json") {
+          if (channel.schemaEncoding === "jsonschema" || channel.encoding === "json") {
```
jtbandes, line 321: I think we might want to update the logic to be more strict like we do inside of parseChannel, where we reject invalid combinations like `schemaEncoding: "jsonschema", encoding: "protobuf"` since we don't really support those. Or you could pass through the original schemaEncoding value and allow parseChannel to do the rejecting.

achim-k: It is already passed through, or am I missing something?

https://github.com/foxglove/studio/blob/c75b7bb358d78a9282621610623b09e231a88220/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts#L348-L351

jtbandes: Yes but this schemaEncoding is one chosen by the `if` statements here. It is not necessarily the original `channel.schemaEncoding`.

achim-k: So you propose to change the snippet that I linked above like so?
```ts
schema: { name: channel.schemaName, encoding: channel.schemaEncoding ?? schemaEncoding, data: schemaData },
```

jtbandes: Maybe, if you do that then you could get rid of the schemaEncoding variable. However, it seems wrong that the code would be processing `channel.schema` without looking at `channel.schemaEncoding` to confirm it has the expected value first.

I think the logic should become `if (channel.encoding === "json" && (channel.schemaEncoding == undefined || channel.schemaEncoding === "jsonschema")) ...` etc. Also, the error message thrown in the `else` case should probably include the schema encoding if set.

jtbandes (approved):

PR #5517, opened 2023-03-09
Fix spurious error icon in app bar

foxymiles:
**User-Facing Changes**
None (feature flagged)

**Description**
In certain cases switching from one layout to another can cause a transient player problem of `warn` severity to register:

```json
{
    "severity": "warn",
    "message": "Data went back in time",
    "error": {}
}
```

Since this is just a transient warning and not an error we don't need to flash the red error icon.

### DIFF
--- a/packages/studio-base/src/components/AppBar/DataSource.tsx
+++ b/packages/studio-base/src/components/AppBar/DataSource.tsx
@@ -90,10 +90,11 @@ export function DataSource({
   const playerName = useMessagePipeline(selectPlayerName);
   const playerPresence = useMessagePipeline(selectPlayerPresence);
   const playerProblems = useMessagePipeline(selectPlayerProblems) ?? [];
-
   const reconnecting = playerPresence === PlayerPresence.RECONNECTING;
   const initializing = playerPresence === PlayerPresence.INITIALIZING;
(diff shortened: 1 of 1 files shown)


foxymiles:
Some alternate fixes we could consider:
1. A separate yellow icon for problems that are only warnings.
2. Try to wait until the layout switch is complete and see if the player problem persists.

2metres (approved):

2metres:
🥇 for excellent use of the word spurious

jtbandes:
Do we have some idea of why this warning is being thrown in the first place?

foxymiles:
> Do we have some idea of why this warning is being thrown in the first place?

Not sure. I've noticed it only happens when switching between particular layouts. I can see it reliably switching from first to second:

[first.json.zip](https://github.com/foxglove/studio/files/10937301/first.json.zip)
[second.json.zip](https://github.com/foxglove/studio/files/10937302/second.json.zip)

foxymiles:
> Do we have some idea of why this warning is being thrown in the first place?

At least in this particular case with those two layouts on nuscenes it's happening on the `/diagnostics` topic.

jtbandes:
I’m not sure about this change — we don’t want to display the icon in the app bar

PR #5518, opened 2023-03-09
WIP: Darken dividers in dark theme

2metres:
**User-Facing Changes**
10% more darkness, light dividers are reserved for draggable borders and menus

**Description**

https://user-images.githubusercontent.com/924528/223985645-39fefac2-2675-4e1a-83ab-0b3c547460ee.mp4


jtbandes:
I'm a weak 👎 on this change. I think the dividers could get less bright, but I find it weird when they are darker than the background color.

PR #5520, opened 2023-03-09
Remove TF frames when removing URDF layer

jtbandes:
**User-Facing Changes**
Fixed an issue where transform frame axes from a URDF file would remain visible when the URDF was removed from the 3D panel.

**Description**
The `Urdfs` scene extension adds "static" transforms for URDFs with a timestamp of 0. When deleting a custom URDF layer, we now delete those transform history entries with a timestamp of 0, and clear out any newly-empty frames from the TF tree (as long as they don't have non-empty descendants).

Fixes FG-1538

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
@@ -259,6 +259,13 @@ export class Urdfs extends SceneExtension<UrdfRenderable> {
 
   public override removeAllRenderables(): void {
     // Re-add coordinate frames and transforms since the scene has been cleared
+    this._refreshTransforms();
+  }
+
+  /**
+   * Re-add coordinate frames and transforms corresponding to existing custom URDFs
+   */
+  private _refreshTransforms() {
     for (const [instanceId, frames] of this.framesByInstanceId) {
       this._loadFrames(instanceId, frames);
     }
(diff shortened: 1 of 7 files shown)


packages/studio-base/src/panels/ThreeDeeRender/transforms/TransformTree.ts
```diff
@@ -62,6 +62,71 @@ export class TransformTree {
...
+    }
+    if (child.parent()?.id !== parentFrameId) {
+      return;
+    }
+    child.removeTransformAt(stamp);
+    this._removeEmptyFrames();
+  }
+
+  /** Prune frames with no history entries and no children from the tree */
+  private _removeEmptyFrames() {
```
snosenzo, line 82: oo I just had a thought on how to do this differently-- still construct the child by parent map, but pass the transform id that was removed to this and then use it to crawl up the tree from the leave to see if it or its parents can be removed. I feel like this shouldn't remove frames that aren't part of the path to the frame that's being removed, and also it's doing more than it needs to by checking all of the frames

snosenzo (approved):
suggestion on cleaning up removeEmptyFrames, but other than that looks good

snosenzo (approved):

packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
```diff
@@ -259,6 +259,13 @@ export class Urdfs extends SceneExtension<UrdfRenderable> {
 
   public override removeAllRenderables(): void {
     // Re-add coordinate frames and transforms since the scene has been cleared
+    this._refreshTransforms();
+  }
+
+  /**
+   * Re-add coordinate frames and transforms corresponding to existing custom URDFs
```
defunctzombie, line 266: I don't see where this re-adds transforms? I do see how it removes transforms.

It would be helpful to add some more comments to this code around the logic that determines to remove the transforms and how that interacts with the transforms that were added. Mostly port what you wrote in the PR description over.

jtbandes: It re-adds them on lines 270 & 273

PR #5532, opened 2023-03-10
Add experimental "use as viewport" setting to Image topics in 3D panel

jtbandes:
**User-Facing Changes**

**Description**


defunctzombie:
Seems you've gone a different direction with this?

PR #5535, opened 2023-03-11
Move desktop code into studio-desktop package

jtbandes:
**User-Facing Changes**
None

**Description**
Move most code from desktop into studio-desktop package so it can be shared with internal repo.


defunctzombie (approved):
📿

PR #5539, opened 2023-03-13
Make panel titles editable for panels without custom settings

foxymiles:
**User-Facing Changes**
Make panel titles editable for panels without custom settings

**Description**
This shows the panel settings & title field even if a panel doesn't have any custom settings. However, it retains the restriction that panels with a custom toolbar do not have editable titles because we have no way to display them.

### DIFF
--- a/packages/studio-base/src/components/PanelToolbar/PanelToolbarControls.tsx
+++ b/packages/studio-base/src/components/PanelToolbar/PanelToolbarControls.tsx
@@ -15,7 +15,7 @@ import CloseIcon from "@mui/icons-material/Close";
 import SettingsIcon from "@mui/icons-material/Settings";
 import { IconButton, Tooltip } from "@mui/material";
 import produce from "immer";
-import { forwardRef, useCallback, useContext, useEffect } from "react";
+import { forwardRef, useCallback, useContext, useEffect, useMemo } from "react";
 import { useAsyncFn } from "react-use";
 import { makeStyles } from "tss-react/mui";
 
@@ -72,6 +72,11 @@ const PanelToolbarControlsComponent = forwardRef<HTMLDivElement, PanelToolbarCon
       [panelId],
     );
 
+    const panelInfo = useMemo(
+      () => (panelType != undefined ? panelCatalog?.getPanelByType(panelType) : undefined),
+      [panelCatalog, panelType],
+    );
+
     const hasSettings = usePanelStateStore(hasSettingsSelector);
 
     const userProfileStorage = useContext(UserProfileStorageContext);
@@ -91,6 +96,7 @@ const PanelToolbarControlsComponent = forwardRef<HTMLDivElement, PanelToolbarCon
         }
         return tooltip;
       }, [hasSettings, panelCatalog, panelType, userProfileStorage]);
+
     useEffect(() => {
       void loadOnboardingState();
     }, [loadOnboardingState]);
@@ -166,7 +172,7 @@ const PanelToolbarControlsComponent = forwardRef<HTMLDivElement, PanelToolbarCon
     return (
       <Stack direction="row" alignItems="center" paddingLeft={1} ref={ref}>
         {additionalIcons}
-        {hasSettings && settingsButton}
+        {panelInfo?.hasCustomToolbar !== true && settingsButton}
         <PanelActionsDropdown isUnknownPanel={isUnknownPanel} />
       </Stack>
     );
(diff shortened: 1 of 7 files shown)


packages/studio-base/src/components/PanelToolbar/PanelToolbarControls.tsx
```diff
@@ -166,7 +172,7 @@ const PanelToolbarControlsComponent = forwardRef<HTMLDivElement, PanelToolbarCon
     return (
       <Stack direction="row" alignItems="center" paddingLeft={1} ref={ref}>
         {additionalIcons}
-        {hasSettings && settingsButton}
+        {(panelInfo?.hasCustomToolbar !== true || hasSettings) && settingsButton}
```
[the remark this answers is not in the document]
defunctzombie: Personal preference is to hoist this more involved logic into a `showSettingsButton` variable. I find it gets harder to read the JSX portion and the intent the more we add this type of logic inline.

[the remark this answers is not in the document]
jtbandes: > How do we know if a panel has a custom toolbar?

The `hasCustomToolbar` property will be set to true. See https://github.com/foxglove/studio/blob/main/packages/studio-base/src/panels/index.ts

packages/studio-base/src/stories/PanelSetup.tsx
```diff
@@ -112,6 +112,21 @@ function setNativeValue(element: unknown, value: unknown) {
   }
 }
 
+export function makeMockPanelCatalog(): PanelCatalog {
+  const allPanels = [...panels.builtin, ...panels.debug, panels.legacyPlot, panels.urdfViewer];
+
+  const visiblePanels = [...panels.builtin];
```
jtbandes, line 118: I think it's generally nice to avoid storybook thrash when we add/remove/change panels, if the stories are not really dependent on those panels. Is there a reason we need to put all the real builtin panels in the catalog for stories?

packages/studio-base/src/components/SettingsTreeEditor/index.tsx
```diff
@@ -47,22 +52,25 @@ const useStyles = makeStyles()((theme) => ({
...
 const makeStablePath = memoizeWeak((key: string) => [key]);
 
+const EMPTY_NODES: SettingsTreeNodes = {};
+const NULL_ACTION_HANDLER: SettingsTree["actionHandler"] = () => undefined;
+
 export default function SettingsTreeEditor({
   settings,
 }: {
-  settings: DeepReadonly<SettingsTree>;
+  settings: undefined | DeepReadonly<SettingsTree>;
```
jtbandes, line 61: Seems like another option would be to just pass a default/dummy/empty tree instead of undefined, so we don't need so many `?.` inside this component.

jtbandes (approved):

packages/studio-base/src/components/SettingsTreeEditor/index.tsx
defunctzombie: Agree - empty tree would be nice.

packages/studio-base/src/components/PanelToolbar/PanelToolbarControls.tsx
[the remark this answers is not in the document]
defunctzombie: TIL

packages/studio-base/src/stories/PanelSetup.tsx
foxymiles: Unfortunately the stories are dependent on this data now. Without this even more stories break because the settings icon starts appearing on a lot of panel stories that previously didn't show it because we don't have an accurate `hasCustomToolbar` value.

packages/studio-base/src/components/SettingsTreeEditor/index.tsx
```diff
@@ -47,22 +52,25 @@ const useStyles = makeStyles()((theme) => ({
 
 const makeStablePath = memoizeWeak((key: string) => [key]);
 
+const EMPTY_NODES: SettingsTreeNodes = {};
```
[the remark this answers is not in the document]
foxymiles: This doesn't seem like we've done consistently in the codebase but I agree it's a nice defensive move.

packages/studio-base/src/components/SettingsTreeEditor/index.tsx
foxymiles: I've refactored this to pass an empty tree instead of undefined.

snosenzo (approved):

packages/studio-base/src/components/PanelToolbar/PanelToolbarControls.tsx
```diff
@@ -125,6 +131,9 @@ const PanelToolbarControlsComponent = forwardRef<HTMLDivElement, PanelToolbarCon
         <SettingsIcon />
       </ToolbarIconButton>
     );
+
+    const showSettingsButton = panelInfo?.hasCustomToolbar !== true || hasSettings;
```
jtbandes, line 135: Could probably use a comment as to why hasCustomToolbar!=true implies we should show the settings.

jtbandes (approved):

PR #5543, opened 2023-03-14
Avoid empty rendering of extension panels when frames are undefined

defunctzombie:
**User-Facing Changes**
None

**Description**
Fix a bug with extension panel handling which resulted in extension panel `onRender` function being called when there was no change to the render state.

The logic which handles `currentFrame` to determine whether to invoke a render on the extension panel used a `?.length !== 0` check when deciding whether to create the conditions for a render. This check is incorrect when `currentFrame` is undefined since `undefined !== 0` evaluates to true and the logic runs regardless of whether there is a change to the frame.

The issue is exposed because the IterablePlayer has a timer which publishes the player state to the message pipeline every second. We could remove this timer but similar logic could exist in other live players which might want to update the "clock" time but not produce any changes to frames. Extension panels don't need to render in these cases either.


jtbandes (approved):

snosenzo (approved):

defunctzombie:
The failing story is blocked by this bugfix: https://github.com/foxglove/studio/pull/5555

The 3d panel Images scene extension does not handle receiving the camera info before the image message in the same frame.

PR #5549, opened 2023-03-15
foxglove-websocket: Support byte array parameters

achim-k:
**User-Facing Changes**
- foxglove websocket: Add support for byte array parameters

**Description**
Adds support for displaying byte array parameters which are communicated as base64 encoded string (see https://github.com/foxglove/ws-protocol/pull/396). These parameters are now displayed as number array and can also be modified in that format.

**Before**:

Screenshot from 2023-05-08 17-21-50

**This PR**:

Screenshot from 2023-05-08 17-23-19

Fixes FG-2949

### DIFF
--- a/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
+++ b/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
@@ -53,6 +53,7 @@ import WorkerSocketAdapter from "./WorkerSocketAdapter";
 
 const log = Log.getLogger(__dirname);
 const textEncoder = new TextEncoder();
+const textDecoder = new TextDecoder();
 
 /** Suppress warnings about messages on unknown subscriptions if the susbscription was recently canceled. */
 const SUBSCRIPTION_WARNING_SUPPRESSION_MS = 2000;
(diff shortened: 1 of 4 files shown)


packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
```diff
@@ -800,7 +810,18 @@ export default class FoxgloveWebSocketPlayer implements Player {
     }
 
     log.debug(`FoxgloveWebSocketPlayer.setParameter(key=${key}, value=${value})`);
-    this._client.setParameters([{ name: key, value: value as Parameter["value"] }], uuidv4());
+    const isByteArray = value instanceof Uint8Array;
```
achim-k, line 813: This doesn't currently work correctly when setting parameters through the parameter panel. The value is an object in that case (e.g. `{"0": 1, "1": 2, "2": 3}` instead of `Uint8Array([1,2,3])`)

jtbandes: IMO the simplest way to allow displaying/editing byte arrays in the parameter panel would be to show them the same way as `number[]` (i.e. it should appear as `[1, 2, 3]`) Do you think there's a simple way to do that? I guess we need to detect the original type of the parameter and remember that when it's edited? Or do we need to change the protocol to allow `{type: "byte_array", value: [1, 2, 3] }` as well as base64?

achim-k: If I would use the extension API, I would set parameters with the same type as they were retrieved. So
```
const isByteArray = value instanceof Uint8Array;
```
should probably work, right?
So yes, I think it would be best if the parameter panel would remember the original type of the parameter.

> IMO the simplest way to allow displaying/editing byte arrays in the parameter panel would be to show them the same way as number[] (i.e. it should appear as [1, 2, 3])

Agreed.

jtbandes: Yep remembering the original type _should_ work. Though if you add a new param in the panel and type `[1,2,3]` then we'll either have to pick a type or ask the user what type it should be.

jtbandes (approved):
Any things we can do to update the Parameters panel stories to demonstrate the new functionality?

achim-k:
> Any things we can do to update the Parameters panel stories to demonstrate the new functionality?

What is currently not working in the storybook is editing of parameters, so the new behavior introduced with this PR is currently not demonstrated. Currently the only difference is how uint8Array parameters are displayed

Screenshot from 2023-05-08 17-55-38
Screenshot from 2023-05-08 17-55-04

achim-k:
> What is currently not working in the storybook is editing of parameters

I have now added this in my last commit

achim-k:
Merging as @jtbandes gave an OK on [Slack](https://foxgloveden.slack.com/archives/C036C8HGCGN/p1684169478345299?thread_ts=1684168669.166389&cid=C036C8HGCGN)

PR #5550, opened 2023-03-15
[3D Panel]  fix bad vertex normals for meshes after applying color

snosenzo:
**User-Facing Changes**

 - [3D Panel] fix bad vertex normals for meshes after applying color

**Description**
Just need to not override existing vertex normals on the mesh if they already exist. 

FG-2295


jtbandes (approved):
Seems good. Any thoughts on how we can get a simple mesh in a storybook to get test coverage on this?

snosenzo:
> Seems good. Any thoughts on how we can get a simple mesh in a storybook to get test coverage on this?

Potentially - will maybe need to find a simpler mesh to the car that would have different vertex normals than those that are computed by threejs. We could maybe test it with some of the existing meshes we have in `studio/packages/studio-base/src/panels/ThreeDeeRender/stories/common.ts` to see if any of them have no normals or different normals from those that threejs computes.

PR #5555, opened 2023-03-16
Fix 3d panel image scene extension

defunctzombie:
**User-Facing Changes**
- Fix 3d panel bug where visualizing an image would not work if there were multiple images for the same camera info.
- Fix 3d panel bug where changing the camera info of an image to a camera info on a different frame would not work.
- Fix 3d panel bug where receiving a camera info message before an image message would not display the image

**Description**

This change fixes several 3d panel bugs with the Images scene extension.

1. This change fixes a bug where the 3d panel would fail to subscribe to a camera info topic if there were multiple image topics related to the same camera info topic and the last image topic (alphabetically) was not visible but an earlier image topic was visible. This change corrects logic in `cameraInfoShouldSubscribe` to subscribe if any image topic needs the desired info topic.
2. This change fixes a bug where the 3d panel would use the `frame_id` from the `image` message rather than the `camera info` message. This bug resulted in the image being displayed with the correct camera information and distortion geometry from the selected camera info but in the wrong coordinate frame.
3. This change fixes a bug where the 3d panel would fail to display an image message when received after a camera info message. The bug happened because the `handleImage` call did not have logic to calculate a camera model from the latest camera info message. This logic only happened in `handleCameraInfo`. This change updates the `handleCameraInfo` logic to save the last camera info message on each topic so `handleImage` can use it to create a camera model when the image message is processed.

Fixes: FG-2500
Fixes: FG-2501

---
I moved the images stories under an `/images` path in the ThreeDeeRender stories. I found this easier to understand and find related stories rather than the previous ungrouped listing.

I added two stories for the explicit message receive ordering cases. I tested these on main and one of the newly added stories fails as expected.


packages/studio-base/src/panels/ThreeDeeRender/renderables/Images.ts
```diff
@@ -245,37 +238,83 @@ export class Images extends SceneExtension<ImageRenderable> {
...
+      return;
+    }
+
+    // We dispose the old renderable and setup a new one. It is easier to dispose the old
+    // renderable than clear only the specific camera topic fields.
+    renderable.dispose();
+    this.renderables.delete(imageTopic);
+
+    // Create a new renderable preserving the image from the previous one
+    const { image, receiveTime, frameId } = renderable.userData;
```
jtbandes, line 281: nit: maybe not great to access this after dispose()?

packages/studio-base/src/panels/ThreeDeeRender/renderables/Images.ts
```diff
@@ -336,46 +398,83 @@ export class Images extends SceneExtension<ImageRenderable> {
...
-            this.renderer.settings.errors.removeFromTopic(imageTopic, CAMERA_MODEL);
-          } catch (errUnk) {
-            const err = errUnk as Error;
-            this.renderer.settings.errors.addToTopic(imageTopic, CAMERA_MODEL, err.message);
-            renderable.userData.cameraModel = undefined;
-          }
-        }
+    this.cameraInfoByTopic.set(messageEvent.topic, cameraInfo);
+
+    // Lookup any image topics assigned to our camera info topic and determine if we need to update
```
jtbandes, line 406: nit: lookup is a noun, look up is a verb

packages/studio-base/src/panels/ThreeDeeRender/renderables/Images.ts
```diff
@@ -245,37 +238,83 @@ export class Images extends SceneExtension<ImageRenderable> {
...
+        imageTopic,
+        NO_CAMERA_INFO_ERR,
+        `No CameraInfo received on ${cameraInfoTopic}`,
+      );
+      return;
+    }
+
+    const cameraModel = this._recomputeCameraModel(newRenderable, cameraInfo);
+    if (image && cameraModel) {
+      this._updateImageRenderable(newRenderable, image, cameraModel, receiveTime, settings);
```
jtbandes, line 302: is there a world where maybe the `_recomputeCameraModel` can happen inside `_updateImageRenderable` so we don't have to do this if check every time?

jtbandes (approved):

packages/studio-base/src/panels/ThreeDeeRender/renderables/Images.ts
defunctzombie: Indeed - will fix that

packages/studio-base/src/panels/ThreeDeeRender/renderables/Images.ts
```diff
@@ -245,37 +238,83 @@ export class Images extends SceneExtension<ImageRenderable> {
...
     }
 
-    // Update the renderable
-    const renderable = this.renderables.get(imageTopic);
-    const image = renderable?.userData.image;
-    const cameraModel = renderable?.userData.cameraModel;
-    if (image && cameraModel) {
-      const receiveTime = renderable.userData.receiveTime;
+    const cameraInfoTopicChanged =
+      prevCameraInfoTopic != undefined && cameraInfoTopic !== prevCameraInfoTopic;
```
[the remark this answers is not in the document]
defunctzombie: Not entirely sure - it was there in the code before. My best evaluation of this was that it was a means of avoiding doing this code flow if the topic is defined already since that will happen when the image or camera info are received. This situation is likely to happen when starting the panel and settings are already present.

packages/studio-base/src/panels/ThreeDeeRender/renderables/Images.ts
```diff
@@ -336,46 +407,83 @@ export class Images extends SceneExtension<ImageRenderable> {
...
+    const dataEqual = cameraInfosEqual(renderable.userData.cameraInfo, newCameraInfo);
+    if (dataEqual && renderable.userData.cameraModel != undefined) {
+      return;
+    }
+
+    const imageTopic = renderable.userData.topic;
+
+    // clear the old model since that is no longer valid if the camera info changed
+    renderable.userData.cameraModel = undefined;
+    renderable.userData.geometry = undefined;
```
snosenzo, line 468: I believe this needs a `?.dispose` since it's a BufferGeometry

packages/studio-base/src/panels/ThreeDeeRender/renderables/Images.ts
```diff
@@ -245,36 +238,86 @@ export class Images extends SceneExtension<ImageRenderable> {
...
+        return;
+      }
 
       this._updateImageRenderable(renderable, image, cameraModel, receiveTime, settings);
+      return;
+    }
+
+    // The camera info topic changed for our renderable
+    // Remove the previous camera_info_topic -> image_topic mapping
+    this.cameraInfoToImageTopics.delete(prevCameraInfoTopic, imageTopic);
```
snosenzo, line 268: could a topic not be supplying multiple camera infos for different images? why can't different images receive camera info from the same topic?

snosenzo: looks like camera info's couldn't really target different images if they wanted to -- so you can ignore this

snosenzo (approved):
just need to add the ?.dispose

packages/studio-base/src/panels/ThreeDeeRender/renderables/Images.ts
defunctzombie: It can. You could as a user select the same camera info to different image topics. There is actually a bug this PR fixes around this. The cameraInfoToImageTopics supports this behavior. Are you seeing something that makes you think it does not?

packages/studio-base/src/panels/ThreeDeeRender/renderables/Images.ts
defunctzombie: added

PR #5559, opened 2023-03-16
Remove/extract logic dependent on `foxglove-data-platform`

jtbandes:
**User-Facing Changes**
None

**Description**
- Add `eventsSupported` and `deviceId` to EventsContext
- Add `currentUserRequired` to `IDataSourceFactory`

### DIFF
--- a/packages/studio-base/src/Workspace.tsx
+++ b/packages/studio-base/src/Workspace.tsx
@@ -59,6 +59,7 @@ import {
   useCurrentLayoutSelector,
 } from "@foxglove/studio-base/context/CurrentLayoutContext";
 import { useCurrentUser } from "@foxglove/studio-base/context/CurrentUserContext";
+import { EventsStore, useEvents } from "@foxglove/studio-base/context/EventsContext";
 import { useExtensionCatalog } from "@foxglove/studio-base/context/ExtensionCatalogContext";
 import { useNativeAppMenu } from "@foxglove/studio-base/context/NativeAppMenuContext";
 import {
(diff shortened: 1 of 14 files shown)


packages/studio-base/src/hooks/useInitialDeepLinkState.ts
```diff
@@ -127,17 +127,27 @@ function useSyncTimeFromUrl(targetUrlState: AppURLState | undefined) {
...
     [deepLinks],
   );
 
   // Maybe this should be abstracted somewhere but that would require a
   // more intimate interface with this hook and the player selection logic.
-  const currentUserRequiredParam = useMemo(
-    () => ({ currentUserRequired: targetUrlState?.ds === "foxglove-data-platform" }),
-    [targetUrlState?.ds],
-  );
+  const currentUserRequiredParam = useMemo(() => {
```
defunctzombie, line 138: Do we even need `useInitialDeepLinkState` to return this information?

defunctzombie: Reading the code here: https://github.com/foxglove/studio/blob/462ebd9ec4bde9a943af364feee1dffe49a5d2b4/packages/studio-base/src/Workspace.tsx#L204 I think we can have the selected source handled in the Workspace where the currentUserRequired is used. Or we can have this return the source that the url state has selected. That somehow feels more natural to me than this currentUserRequired return.

packages/studio-base/src/Workspace.tsx
```diff
@@ -628,7 +627,8 @@ export default function Workspace(props: WorkspaceProps): JSX.Element {
     currentUser,
   ]);
 
-  const showEventsTab = currentUser != undefined && playerSourceId === "foxglove-data-platform";
+  const eventsSupported = useEvents(selectEventsSupported);
```
defunctzombie, line 630: Somewhat surprised by this. Why would we have an events context at all if events are not supported? Don't feel strongly since this is all private and we can change it. This approach does align with a future notion of exposing events from the Player/DataSource which would probably report something like an eventsSupported value.

defunctzombie (approved):
I've always found the return value of `useInitialDeepLinkState` to be suspect so this does feel like a good time to make that more sane. Not a hard blocker tho if you think the return value is sane.

packages/studio-base/src/Workspace.tsx
jtbandes: The current way useEvents works is it requires the context to be present. Agree we could explore changing the APIs to make the context optional, but that might be a more invasive change everywhere that useEvents is used.

packages/studio-base/src/hooks/useInitialDeepLinkState.ts
jtbandes: You mean move the `foundSource` thing out into Workspace?

That would be ok, not sure I see how it's much better than this though?

PR #5560, opened 2023-03-16
ColorPicker input improvements

2metres:
**User-Facing Changes**
Color picker clear icon only appears on hover

**Description**

## Before 
Screen Shot 2023-03-16 at 1 12 19 pm 

## After
https://user-images.githubusercontent.com/924528/225493940-570ccda0-35ef-4069-a873-d282ac69fb55.mov


jtbandes:
Seems like a good idea. But why is the icon a circle now?

2metres:
I think what I was trying to do was give the clear icon a bit more mass.

jtbandes:
I guess that is kinda normal in Apple-land...

image

jtbandes (approved):

PR #5562, opened 2023-03-16
Restore legend toggle

2metres:
**User-Facing Changes**
Restore legend toggle in plot panel

**Description**

Changes circle checkboxes to squares (Radio vs Checkbox) and smallifies text in legend 

https://user-images.githubusercontent.com/924528/225771634-ded626b1-0396-4e26-aa85-eca55fa8d323.mov

### DIFF
--- a/packages/studio-base/src/panels/Plot/PlotLegendRow.tsx
+++ b/packages/studio-base/src/panels/Plot/PlotLegendRow.tsx
@@ -179,7 +179,7 @@ export function PlotLegendRow({
         <Typography
           noWrap
           flex="auto"
-          variant="body2"
+          variant="caption"
           className={cx({ [classes.disabledPathLabel]: !path.enabled })}
         >
           {plotPathDisplayName(path, index)}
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/panels/Plot/PlotLegendRow.tsx
```diff
@@ -63,14 +60,17 @@ const useStyles = makeStyles<void, "plotName">()((theme, _params, classes) => ({
...
     position: "sticky",
-    padding: theme.spacing(0, 0.25),
     height: ROW_HEIGHT,
     left: 0,
   },
   legendIconButton: {
-    padding: `${theme.spacing(0.75)} !important`,
-    marginLeft: theme.spacing(0.125),
-    fontSize: theme.typography.pxToRem(14),
+    padding: `${theme.spacing(0.625)} !important`,
```
defunctzombie, line 67: `!important` always looks suspect to me - why do we need to override that here rather than specifying it into the button styling?

defunctzombie:
Some thoughts on the current pr:

* I find the use of `!important` suspect given that we control all the components and their styling
* The button has a different elevation than the legend. The elevation feels too aggressive and the two should be the same elevation
image
* I find the button "appears" abnormally large compared to the legend elements. I think it is technically the same height as a legend row but with the elevation and the row text being smaller it looks outsized to me.

packages/studio-base/src/panels/Plot/PlotLegendRow.tsx
2metres: This might be inherited from before we sorted out the tss-react vs mui load order. I'll see if removing this works

defunctzombie:
I am 👎 on the smaller text - we have a text size in the app and I dislike when panels and components deviate cause they feel like it. Makes the app look inconsistent.

defunctzombie:
Separate from my comments I am 👍 on bringing this toggle back in some form.

defunctzombie (approved):
Approving the PR figuring that you'll come to some nice place that incorporates the comments.

2metres:
Closed FG-2502

PR #5563, opened 2023-03-16
Merge settings icons and built-in icons

2metres:
**User-Facing Changes**
None

**Description**
Remove duplication


packages/studio/src/index.ts
```diff
@@ -32,6 +32,66 @@ export interface Time {
...
+  | "Shapes"
+  | "Share"
+  | "SouthEast"
+  | "Star"
+  | "Timeline"
+  | "Topic"
+  | "Walk"
+  | "World"
+  | never; // never has no effect here other than keeping the semicolon on a separate line for easier conflict resolution
+
```
2metres, line 94: @foxymiles is there a way to avoid having to duplicate list?

packages/studio/src/index.ts
```diff
@@ -421,46 +481,6 @@ export interface ExtensionModule {
   activate: ExtensionActivate;
 }
 
-export type SettingsIcon =
```
jtbandes, line 424: Unfortunately this is now a public API that may be used by third-party extensions, so we can't change these names. If we want to change the icons, we'll need to map the old names to whatever new icons we choose.

jtbandes (changes requested):

packages/studio/src/index.ts
2metres: Can we map the SettingsIcon to the RegisteredIconNames?

jtbandes: Yep, see here

https://github.com/foxglove/studio/blob/main/packages/studio-base/src/components/SettingsTreeEditor/NodeEditor.tsx#L283

 https://github.com/foxglove/studio/blob/main/packages/studio-base/src/components/SettingsTreeEditor/icons.ts#L46

PR #5575, opened 2023-03-17
chore(mcap-support): rename method to `deserialize`

jon-chuang:
**User-Facing Changes**
None

**Description**
rename deserialization method to `deserialize`

### DIFF
--- a/packages/mcap-support/src/parseChannel.test.ts
+++ b/packages/mcap-support/src/parseChannel.test.ts
@@ -19,7 +19,7 @@ describe("parseChannel", () => {
         ),
       },
     });
-    expect(channel.deserializer(new TextEncoder().encode(JSON.stringify({ value: "hi" })))).toEqual(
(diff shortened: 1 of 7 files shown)


defunctzombie:
chatgpt has spoken the true words

image

jtbandes (approved):
Doesn't seem like a significant improvement, but not harmful 🤷

jon-chuang:
Hehe well, to me, it was a lot less clear before if calling deserializer would return a deserializer, or the deserialized object.


Chatgpt seems a little kooky about this.

jtbandes:
> calling `deserializer` would return a deserializer

One way of thinking about it is that the `deserializer` property holds a deserializer.

jon-chuang:
> One way of thinking about it is that the deserializer property holds a deserializer.

Yes, but it does not in fact, since the deserializer is actually either some external object referenced by the `deserialize` pointer...

The API may have been clearer, if more verbose, if `deserializer` stored an object with method `deserialize`...

PR #5578, opened 2023-03-17
Fix crash with "Do not know how to serialize a BigInt"

jtbandes:
**User-Facing Changes**
Fixed a crash in the Raw Messages panel in certain cases involving 64-bit integers nested inside objects nested inside arrays.

**Description**
Check that the nested value is a primitive before suggesting it as a filter. This avoids a crash when there was a field named e.g. `id` which contained a nested object that had a bigint somewhere inside it.

Test data:
[test_int64_0.mcap.zip](https://github.com/foxglove/studio/files/11006052/test_int64_0.mcap.zip)

Fixes FG-2528

### DIFF
--- a/packages/studio-base/src/panels/RawMessages/getValueActionForValue.ts
+++ b/packages/studio-base/src/panels/RawMessages/getValueActionForValue.ts
@@ -86,9 +86,10 @@ export function getValueActionForValue(
       // back to `/topic.object[10]` if necessary.
       let typicalFilterName;
       if (structureItem.structureType === "message") {
-        typicalFilterName = Object.keys(structureItem.nextByName).find((key) =>
-          isTypicalFilterName(key),
-        );
+        typicalFilterName = Object.entries(structureItem.nextByName).find(
+          ([key, nextStructureItem]) =>
+            nextStructureItem.structureType === "primitive" && isTypicalFilterName(key),
+        )?.[0];
       }
       if (
         typeof value === "object" &&
--- a/packages/studio-base/src/util/getItemString.tsx
+++ b/packages/studio-base/src/util/getItemString.tsx
@@ -5,10 +5,10 @@
 import { ReactNode } from "react";
 import tinycolor from "tinycolor2";
 
+import { filterMap } from "@foxglove/den/collection";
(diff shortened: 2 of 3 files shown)


packages/studio-base/src/panels/RawMessages/getValueActionForValue.ts
```diff
@@ -86,9 +86,10 @@ export function getValueActionForValue(
       // back to `/topic.object[10]` if necessary.
       let typicalFilterName;
       if (structureItem.structureType === "message") {
-        typicalFilterName = Object.keys(structureItem.nextByName).find((key) =>
-          isTypicalFilterName(key),
-        );
+        typicalFilterName = Object.entries(structureItem.nextByName).find(
```
defunctzombie, line 89: Feels ripe for a function `computeTypicalFilterName(structureItem)` or some such. Would help with reasoning about this code - at least for me.

packages/studio-base/src/util/getItemString.tsx
```diff
@@ -90,10 +91,16 @@ export function getItemString(
...
   // Surface typically-used keys directly in the object summary so the user doesn't have to expand it.
-  const filterKeys = keys
-    .filter(isTypicalFilterName)
-    .map((key) => `${key}: ${(data as { [key: string]: unknown })[key]}`)
-    .join(", ");
+  const filterKeys = filterMap(keys, (key) => {
+    const value = (data as Record<string, unknown>)[key];
+    if (
+      isTypicalFilterName(key) &&
+      (value == undefined || ["string", "number", "bigint", "boolean"].includes(typeof value))
```
defunctzombie, line 98: `["string", "number", "bigint", "boolean"]` could be hoisted to a file constant

defunctzombie (approved):

packages/studio-base/src/util/getItemString.tsx
jtbandes: I was hopeful that the js engine would do this for me. perhaps I should not pin my hopes on such things.

packages/studio-base/src/panels/RawMessages/getValueActionForValue.ts
jtbandes: unfortunately isTypicalFilterName is used in a lot of places where you are only working with keys, not values, so not sure how much use it would get

PR #5588, opened 2023-03-19
Enable minify for EsbuildMinifyPlugin

defunctzombie:
**User-Facing Changes**
None

**Description**
We want to production builds to be small and we don't need identifier names in our production builds. These can be part of source builds for OSS users.


jtbandes (approved):
A lot of these are likely the defaults and maybe we could remove some of the custom config now?

defunctzombie:
> A lot of these are likely the defaults and maybe we could remove some of the custom config now?

I was thinking that too - I was actually surprised that the code still remained as un-minified as it did. I'll do another pass to see if I can get this to only the defaults.

PR #5590, opened 2023-03-20
Persist sidebar state

foxymiles:
**User-Facing Changes**
Persist sidebar UI state.

**Description**
Persist sidebar UI state. To make this more manageable I've lifted all the sidebar related state up into a new Workspace zustand context. This should make it easier to expand the slice of workspace state we want to persist in the future.

### DIFF
--- a/packages/studio-base/src/App.tsx
+++ b/packages/studio-base/src/App.tsx
@@ -10,7 +10,6 @@ import GlobalCss from "@foxglove/studio-base/components/GlobalCss";
 import EventsProvider from "@foxglove/studio-base/providers/EventsProvider";
 import { StudioLogsSettingsProvider } from "@foxglove/studio-base/providers/StudioLogsSettingsProvider";
 import TimelineInteractionStateProvider from "@foxglove/studio-base/providers/TimelineInteractionStateProvider";
-import WorkspaceContextProvider from "@foxglove/studio-base/providers/WorkspaceContextProvider";
 
 import Workspace from "./Workspace";
 import { CustomWindowControlsProps } from "./components/AppBar/CustomWindowControls";
@@ -84,7 +83,6 @@ export function App(props: AppProps): JSX.Element {
 
   const providers = [
     /* eslint-disable react/jsx-key */
-    <WorkspaceContextProvider />,
     <StudioLogsSettingsProvider />,
     <StudioToastProvider />,
     <LayoutStorageContext.Provider value={layoutStorage} />,
--- a/packages/studio-base/src/App.tsx
+++ b/packages/studio-base/src/App.tsx
@@ -10,6 +10,7 @@ import GlobalCss from "@foxglove/studio-base/components/GlobalCss";
 import EventsProvider from "@foxglove/studio-base/providers/EventsProvider";
 import { StudioLogsSettingsProvider } from "@foxglove/studio-base/providers/StudioLogsSettingsProvider";
 import TimelineInteractionStateProvider from "@foxglove/studio-base/providers/TimelineInteractionStateProvider";
+import WorkspaceContextProvider from "@foxglove/studio-base/providers/WorkspaceContextProvider";
 
 import Workspace from "./Workspace";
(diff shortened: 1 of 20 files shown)


packages/studio-base/src/components/Sidebars/NewSidebar.stories.tsx
```diff
@@ -40,37 +53,30 @@ const C = () => <>{longText}</>;
...
 
-const LEFT_ITEMS = new Map<LeftKey, NewSidebarItem>([
-  ["a", { title: "A", component: A }],
-  ["b", { title: "B", component: B }],
-  ["c", { title: "C", component: C }],
+const LEFT_ITEMS = new Map<LeftSidebarItemKey, NewSidebarItem>([
+  ["topics", { title: "A", component: A }],
+  ["variables", { title: "B", component: B }],
+  ["studio-logs-settings", { title: "C", component: C }],
 ]);
```
jtbandes, line 64: In general I think the grouping of state into a zustand store is a good change. However one thing I'm not a fan of is when these stories become more dependent on the actual tabs/values we use in the app. Although technically this component lives inside the main app package, I feel that we should try to treat it as though it were part of a component _library_ and avoid coupling it directly to a specific use case. This can reduce the amount of state you have to hold in your head while working on the component (depending how you feel about generics). It also means that we don't have to come back and change the stories again if we decide to add/remove tabs in the workspace.

packages/studio-base/src/App.tsx
```diff
@@ -83,6 +84,7 @@ export function App(props: AppProps): JSX.Element {
 
   const providers = [
     /* eslint-disable react/jsx-key */
+    <WorkspaceContextProvider />,
```
jtbandes, line 87: I find it a bit confusing that in order to use the Workspace component, you have to render Workspace **and** WorkspaceContextProvider. Do you think it could make sense to have the Workspace render this provider itself? (I realize that would require a little restructuring so that the workspace itself could access the store)

packages/studio-base/src/components/Sidebars/NewSidebar.stories.tsx
foxymiles: I agree but the generics on the components make this this tricky. I can try to preserve them in their more general state but it makes this a lot harder for a generality we don't actually need today.

packages/studio-base/src/App.tsx
foxymiles: There could be a wrapper component in the Workspace.tsx file that provides the value if you think that's cleaner.

2metres:
Noticing in the Example Layout if I have a panel selected in panel settings then close/open panel settings it reverts back to the top level panel as the selected one. Should we persist which panel was selected last and restore it?

foxymiles:
> Noticing in the Example Layout if I have a panel selected in panel settings then close/open panel settings it reverts back to the top level panel as the selected one. Should we persist which panel was selected last and restore it?

We could try although that's layout specific. It's one of the things that's hard because our state is fragmented over many contexts.

packages/studio-base/src/Workspace.tsx
```diff
@@ -172,8 +178,22 @@ function WorkspaceContent(props: WorkspaceProps): JSX.Element {
   const playerProblems = useMessagePipeline(selectPlayerProblems);
 
   const sidebarItem = useWorkspaceStore(selectWorkspaceSidebarItem);
-
-  const { setLeftSidebarOpen, setRightSidebarOpen, selectSidebarItem } = useWorkspaceActions();
+  const leftSidebarItem = useWorkspaceStore(selectWorkspaceLeftSidebarItem);
+  const leftSidebarOpen = useWorkspaceStore(selectWorkspaceLeftSidebarOpen);
+  const leftSidebarSize = useWorkspaceStore(selectWorkspaceLeftSidebarSize);
```
foxymiles, line 183: Having this in a selector here might be costly.

jtbandes (approved):
Feel free to revert my latest changes if you desire

esthersweon:
Not adding to changelog, as this is for the new sidebar only (feature-flagged).

Issue #5592, opened 2023-03-20
Scrolling State Transitions panel conflicts with zoom


smbera:
**Description**

Here is the PR for this bug.  https://github.com/foxglove/studio/pull/5593

before repair：
**From this video, it can be seen that when I use the mouse wheel to scroll down and turn the page, the X-axis will shrink accordingly. This is because the operation of scrolling the mouse wheel down to turn the page conflicts with the operation of zooming the chart.**

https://user-images.githubusercontent.com/23401125/226315547-cae3b374-0096-4727-a4cd-c96dcefe4268.mp4

after repair：
**Here, I use "Ctrl" + mouse wheel to zoom the chart, but just use the mouse wheel to scroll down and turn pages.**

https://user-images.githubusercontent.com/23401125/226315608-9621cbbf-a7fc-47f5-8aeb-b5bb36391d9d.mp4

- Version: 1.44.0-dev
- Platform: Chrome 版本 110.0.5481.100（正式版本） （64 位）

**Steps To Reproduce**

**Expected Behavior**

smbera:
> The current version of the app is 1.44.0. Please try with the latest version. Also, please provide more information about what platform/browser you are using. I am unable to reproduce this issue on my system.

ok, I update the information

Version: 1.44.0-dev
Platform: Chrome 版本 110.0.5481.100（正式版本） （64 位）

jtbandes:
OK, I understand now that you are talking about the conflict between scrolling the state transitions list (1) and zooming (2). I was confused because your video also shows scrolling the whole page (3) which is unexpected.
<img width="830" alt="image" src="https://user-images.githubusercontent.com/14237/226770929-8fe1083b-9374-4911-b074-48e9a5bdc29b.png">

We have some work in progress (https://github.com/foxglove/studio/pull/5604) that will change the State Transitions panel a bit. However, I think this issue may still be relevant after that change.

PR #5593, opened 2023-03-20
fix: Page turning conflicts with scroll wheel zoom

smbera:
**User-Facing Changes**

fix: Page turning conflicts with scroll wheel zoom

**Description**
Here is the Issues for this PR. https://github.com/foxglove/studio/issues/5592


smbera:
> Hi, thanks for the contribution. From looking at this code change, I don't understand why it would result in different behavior. Can you please explain the change?
> 
> Additionally, based on your videos in #5592, I would add that the whole page **should not** be scrollable. So the real bug is that the page is scrollable, and I would rather see a PR to fix that underlying issue.

when the panel of "State Transitions" has many Topic，this will make the page is scrollable. When I use the mouse wheel to turn the page, I accidentally trigger the zooming of the chart, which is not the zooming operation I want

defunctzombie:
I tried this PR and one thing that it should note is this changes the scroll UX. You have to discover and press "ctrl" to scroll the chart. This is a change to the current behavior which does not require learning about "ctrl". This change also impacts the Plot panel and prevents it from being zoomable unless you hold ctrl.

We will need to find some other way to address the issue with state transitions that does not regress our Plot panel.

defunctzombie:
Closing this in favor of https://github.com/foxglove/studio/pull/5678 which preserves the existing chart scrolling behavior in the plot panel while still allowing scrolling of the state transitions panel via a scrollbar.

PR #5589, opened 2023-03-20
Remove analytics provider

defunctzombie:
**User-Facing Changes**
None

**Description**
Not relevant for OSS repo.

PR #5594, opened 2023-03-20
Replace Sentry.captureException with reportError()

jtbandes:
**User-Facing Changes**
None

**Description**
Remove `Sentry.captureException()` calls and replace them with a new `reportError()` function that can be overridden.
When https://github.com/foxglove/studio/pull/5589 merges we will be able to completely remove the `@sentry/core` dependency.
Relates to FG-2454

### DIFF
--- a/packages/studio-base/src/reportError.ts
+++ b/packages/studio-base/src/reportError.ts
@@ -1,0 +1,22 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+type ReportErrorHandler = (error: Error) => void;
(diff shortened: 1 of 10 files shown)


packages/studio-base/src/reportError.ts
```diff
@@ -0,0 +1,22 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+type ReportErrorHandler = (error: Error) => void;
+let reportFn: ReportErrorHandler = (_err: Error) => {};
```
defunctzombie, line 6: As a pedantic measure I would probably use something like `global.foxgloveStudioReportFn` for this to handle situations where (for some unlikely reason) two copies of reportError.ts end up existing in the generated bundle logic.

defunctzombie (approved):

PR #5601, opened 2023-03-21
Replace hard coded sign-in link with one provided by the context

2metres:
**User-Facing Changes**
| core | private-deployments |
|:---:|:---:|
| Screen Shot 2023-03-21 at 11 26 30 am | Screen Shot 2023-03-21 at 11 26 21 am |
| links to Data Platform homepage | triggers sign in flow | 

**Description**
Change hardcoded url to use user context and account for undefined state

### DIFF
--- a/packages/studio-base/src/components/OpenDialog/Start.tsx
+++ b/packages/studio-base/src/components/OpenDialog/Start.tsx
@@ -11,7 +11,10 @@ import FoxgloveLogoText from "@foxglove/studio-base/components/FoxgloveLogoText"
 import Stack from "@foxglove/studio-base/components/Stack";
 import TextMiddleTruncate from "@foxglove/studio-base/components/TextMiddleTruncate";
 import { useAnalytics } from "@foxglove/studio-base/context/AnalyticsContext";
-import { useCurrentUserType } from "@foxglove/studio-base/context/CurrentUserContext";
+import {
+  useCurrentUser,
+  useCurrentUserType,
+} from "@foxglove/studio-base/context/CurrentUserContext";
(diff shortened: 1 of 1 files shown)


jtbandes (approved):

packages/studio-base/src/components/OpenDialog/Start.tsx
```diff
@@ -310,36 +314,49 @@ function SidebarItems(props: { onSelectView: (newValue: OpenDialogViews) => void
...
                     });
+                    signIn();
                   }}
                 >
                   Sign in
                 </Button>
               </>
+            ) : (
+              <Button
+                href="https://foxglove.dev/data-platform"
```
defunctzombie, line 347: Do we not want to direct users to the signup page instead? https://console.foxglove.dev/signup

2metres: Great question. If I am a user of studio-core what can I signup for?

jtbandes: The whole point of this ticket is to remove hardcoded `console.foxglove.dev` since this may be deployed pointing to either dev or prod.

2metres: Where should this link point to?

defunctzombie: After second review what you have here is good. This is the fallback state to learn more.

What I would say is that the logEvent action should maybe be something other than "create-account"

defunctzombie (approved):

PR #5604, opened 2023-03-21
Migrate State Transitions panel to use Panel Settings

2metres:
**User-Facing Changes**
Message path input and time toggle on StateTransitions panel now in settings sidebar

Screen Shot 2023-03-24 at 2 18 20 pm

Screen Shot 2023-03-24 at 12 28 11 pm

**Description**
Migrates to use Panel Settings


defunctzombie (approved):
Maybe sprinkle a few why/noteworthy comments in a few places that you think would help future understanding.

Some ideas:
- Why memoizeWeak used rather than traditional react memoization
- Some of the magic numbers in path.slice calls
- types.ts has more than types now so maybe some comments on that or find a new home for those?

foxymiles:
I've pushed an update to address @defunctzombie's feedback.

PR #5607, opened 2023-03-22
Restyle data source info in AppBar

2metres:
**User-Facing Changes**
None (behind feature-flag)

**Description**
Update styling on data source in topnav.
- Remove border from initial "Open data source" button.
- Remove the down arrow


2metres:
Need help getting the Layout name. Currently it's fairly difficult

jtbandes:
Can we make the layout name also be present when no data source is selected?

2metres:
@foxymiles I've managed to retrieve the Layout ID but could you help to return the layout name?

amacneil:
- move layouts to the right
- make the text white
- keep the data source clickable for now (as you have in this pr)
- probably drop the external link icon for now

foxymiles (approved):

packages/studio-base/src/components/AppBar/DataSource.tsx
```diff
@@ -2,14 +2,16 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
-import ArrowDropDownIcon from "@mui/icons-material/ArrowDropDown";
-import ErrorIcon from "@mui/icons-material/Error";
-import { Button, ButtonBase, CircularProgress, Divider, Tooltip, Typography } from "@mui/material";
+import { ErrorCircle20Filled, Open16Filled } from "@fluentui/react-icons";
```
defunctzombie, line 5: What relationship does this change (styling the data source) have to the PR? If we can we should leave unrelated changes for separate PRs so when we review we know what change we are reviewing.

defunctzombie (changes requested):
Putting a hold on this PR from slack discussion. We'll use a different approach here.

defunctzombie:
I think we should replace this PR with https://github.com/foxglove/studio/pull/5687 and move the data source changes to a separate PR.

packages/studio-base/src/components/AppBar/DataSource.tsx
amacneil: That was actually the original project. It morphed into mostly a layout menu change and reverting most (I thought all but I guess not) of the data source name changes.

image

defunctzombie (approved):

PR #5609, opened 2023-03-22
Improve message path input interactions

foxymiles:
**User-Facing Changes**
Improves message path input autocomplete placement when the input is near the bottom of the screen.

**Description**
Replaces react-autocomplete with MUI autocomplete, which does a much better job of list placement at the edges of the window.


foxymiles:
We should complete the conversion of the state transition panel to settings before merging this.

defunctzombie:
> instead just showing all options on scroll when there has been no user input

I definitely expect this to be a problem and something we can try. I can point you in the direction of some larger datasets which we can use to experience the UX impact.

foxymiles:
> I definitely expect this to be a problem and something we can try. I can point you in the direction of some larger datasets which we can use to experience the UX impact.

Sure if you have some good datasets for this please link.

foxymiles:
@defunctzombie I've implemented the react-window based virtualization. Seems to work ok.

jtbandes (approved):

amacneil:
I did some basic testing and seems to work well.

Two UX complaints:
1/ There seems to be a ton of blank space at the end of the list:
image

2/ The vertical spacing/padding is much tighter than our other dropdowns. We should be consistent (imo our other dropdowns are much too padded so should probably change closer to what the autocomplete does, but they should both be identical). Also the other dropdowns have a much larger shadow than the autocomplete:
image

defunctzombie:
I've noticed that with this version the dropdown is shorter even when there is more screen space to use. I don't think that's strictly bad but something I will note.

foxymiles:
> I've noticed that with this version the dropdown is shorter even when there is more screen space to use. I don't think that's strictly bad but something I will note.

This is strictly configurable. Any idea what rough fraction of the screen we should shoot for?

foxymiles:
> Two UX complaints: 1/ There seems to be a ton of blank space at the end of the list:

I think I can fix this.

> 2/ The vertical spacing/padding is much tighter than our other dropdowns.

Do you want to try to address that as part of this PR or in a separate preliminary PR?

foxymiles:
> Two UX complaints: 1/ There seems to be a ton of blank space at the end of the list

Fixed.

> 2/ The vertical spacing/padding is much tighter than our other dropdowns

I've changed this to be the same as the existing message path input on main. If we want to change this app wide maybe we can handle that in a followup?

PR #5612, opened 2023-03-22
Improve contrast on plot tooltip

foxymiles:
**User-Facing Changes**
Improve contrast on plot tooltip

**Description**
Improve contrast on plot tooltip


2metres:
Closed in favor of #5612

PR #5617, opened 2023-03-23
Support GeoJSON feature names in map panel

foxymiles:
**User-Facing Changes**
Support GeoJSON feature names in map panel

**Description**
This establishes the convention that a `name` field in the `properties` object of a GeoJSON message will be used to show an on-hover tooltip on that feature in the map panel.

The GeoJSON spec allows the feature's `properties` object to contain arbitrary JSON:
https://www.rfc-editor.org/rfc/rfc7946

So we could call this `name` field anything, we just have to establish some convention. `name` seems to be often used in my searches online.

We should also document this in the GeoJSON section of the Map panel's docs.


packages/studio-base/src/panels/Map/MapPanel.tsx
```diff
@@ -468,12 +469,22 @@ function MapPanel(props: MapPanelProps): JSX.Element {
   const [filterBounds, setFilterBounds] = useState<LatLngBounds | undefined>();
 
   const addGeoFeatureEventHandlers = useCallback(
-    (message: MessageEvent<unknown>, layer: Layer) => {
+    (
+      feature: Feature<Geometry, GeoJsonProperties>,
```
defunctzombie, line 473: My general read of this is we can forgo adding the `geojson` dependency and instead make a type alias here for this basic type having an optional properties: `Record<string, unknown>`.

I don't know how large the geojson stuff is but seems like a lot for one type.

defunctzombie (approved):

packages/studio-base/src/panels/Map/MapPanel.tsx
foxymiles: I replaced this with our own typedef.

PR #5626, opened 2023-03-25
Create a bug issue template

defunctzombie:
**User-Facing Changes**
None

**Description**
Copy our generic bug issue template (https://github.com/foxglove/.github/blob/main/.github/ISSUE_TEMPLATE/bug.md) and add a line for "Data source".

For Studio issues it is helpful to know the data source; which is not relevant on other code repos.

Removed in: https://github.com/foxglove/studio/pull/4975


defunctzombie:
This might require us to bring back all the items of the ISSUE_TEMPLATE folder if we want to have the same list we currently have. Not sure if this will merge with the global list of issue templates.

PR #5628, opened 2023-03-27
Rearrange sidebar items

foxymiles:
**User-Facing Changes**
None (feature flagged)

**Description**
Rearranges the side bar items so that Settings & Topics are on the left and Events, Variables & Log Settings are on the right.

### DIFF
--- a/packages/studio-base/src/Workspace.tsx
+++ b/packages/studio-base/src/Workspace.tsx
@@ -571,7 +571,7 @@ function WorkspaceContent(props: WorkspaceProps): JSX.Element {
 
   const leftSidebarItems = useMemo(() => {
     const items = new Map<LeftSidebarItemKey, NewSidebarItem>([
-      ["panel-settings", { title: "Panel settings", component: PanelSettingsSidebar }],
+      ["panel-settings", { title: "Panel", component: PanelSettingsSidebar }],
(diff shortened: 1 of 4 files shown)


jtbandes:
User-facing changes should be none.

2metres (approved):
I think I prefer the Topics as the first item and the Panel Settings as the second. @amacneil thoughts?

packages/studio-base/src/Workspace.tsx
```diff
@@ -599,24 +599,24 @@ function WorkspaceContent(props: WorkspaceProps): JSX.Element {
 
   const leftSidebarItems = useMemo(() => {
     const items = new Map<LeftSidebarItemKey, NewSidebarItem>([
+      ["panel-settings", { title: "Panel settings", component: PanelSettingsSidebar }],
```
amacneil, line 602:
```suggestion
      ["panel-settings", { title: "Panel", component: PanelSettingsSidebar }],
```

@2metres wdyt? Seems cleaner imo, and still clear what it does

amacneil:
No panel settings should come first, you need to interact with it constantly. The topics list is (currently) useless.

amacneil:
Can you make the left sidebar open by default when the app is first launched.

PR #5632, opened 2023-03-27
Encode typed arrays as arrays instead of objects when publishing messages as JSON

### LINKED ISSUE #5455: Custom panel - Cannot subscribe to certain topics and republish that same message

achim-k:
**User-Facing Changes**
- Encode typed arrays as arrays instead of objects when publishing messages as JSON

**Description**
This PR adds a custom JSON.stringify replacer function to encode typed arrays as arrays rather than objects with numeric keys.

From https://github.com/foxglove/studio/issues/5455#issuecomment-1483075718:

> Default behavior, without replacer:
> 
> ```js
> JSON.stringify({ val: new Uint8Array([1,2,3])});
> // Results in 
> // '{"val":{"0":1,"1":2,"2":3}}'
> ```
> 
> With replacer:
> 
> ```js
> let replacer = (_key, value) => {
>     return ArrayBuffer.isView(value) ? Array.from(value) : value;
> };
> JSON.stringify({ val: new Uint8Array([1,2,3])}, replacer);
> // results in
> // '{"val":[1,2,3]}'
> ```
> 

Fixes #5455
Fixes FG-2243


jtbandes (approved):
Rather than "Encode typed arrays as arrays instead of objects when publishing messages as JSON", I think the user-facing changes for the changelog should be "The `publish()` API for extensions now supports publishing typed arrays"

PR #5633, opened 2023-03-27
foxglove-websocket: Add player problem for every status message

achim-k:
**User-Facing Changes**
- Foxglove websocket: Add player problems for every incoming status message

**Description**
- Maps status messages to player problems (unresolvable for now)
- Adds a special tip if the player message is `"Send buffer limit reached"` (see https://github.com/foxglove/ros-foxglove-bridge/pull/201)

Screenshot from 2023-03-27 15-26-46

Fixes: foxglove/ros-foxglove-bridge/issues/174
Fixes: FG-2202

### DIFF
--- a/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
+++ b/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
@@ -318,6 +318,17 @@ export default class FoxgloveWebSocketPlayer implements Player {
       } else {
         log.error(msg);
       }
+      
+      this._problems.addProblem(event.message, {
+        message: event.message,
+        severity:
+          event.level === StatusLevel.INFO
+            ? "info"
+            : event.level === StatusLevel.WARNING
+              ? "warn"
+              : "error",
+      });
+      this._emitState();
     });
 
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
```diff
@@ -318,6 +318,17 @@ export default class FoxgloveWebSocketPlayer implements Player {
       } else {
         log.error(msg);
       }
+      
+      this._problems.addProblem(event.message, {
+        message: event.message,
+        severity:
+          event.level === StatusLevel.INFO
```
defunctzombie, line 325: todo: make a function that will turn a StatusLevel into a problem level string

defunctzombie:
@jtbandes Achim and I paired on this and view this as a step in the direction of resolving the referenced foxglove bridge issue. There's no one size fits all solution for this so we are going with notifying the user this happened. We'll see what feedback we get and can iterate this approach.

Leaving the review for you since I was involved in the implementation.

jtbandes (approved):
👍 definitely better than nothing

PR #5636, opened 2023-03-27
Bump typescript to 5.0.3

jtbandes:
**User-Facing Changes**
None

**Description**

- Bumps typescript to 5.0
- Removes some type hacks now that the builtin lib has better OffscreenCanvas types (see https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/1480)


foxymiles (approved):
Thanks for doing this!

jtbandes:
Seems like this is blocked on https://github.com/cevek/ttypescript/issues/140, unless we can remove our usage of ttypescript.

packages/studio-base/package.json
```diff
@@ -190,7 +190,7 @@
     "ts-loader": "9.4.2",
     "tss-react": "4.7.7",
     "ttypescript": "1.5.15",
```
sunrabbit123, line 192:
```suggestion
    "ttsc": "0.0.5",
```
I think changing this would work nicely.

sunrabbit123: These libraries support TS@5 and use the
and provides the same functionality as ttypescript.

snosenzo (approved):

packages/studio-base/package.json
defunctzombie: This looks like a fork of ttypescript. Is there a PR for ttypescript that contributes these changes back so we can stay on ttypescript?

defunctzombie:
@jtbandes Not sure how to best drop ttypescript. My recollection is that we use it to re-write the paths in studio-base for the npm package so downstream users do not need to setup any aliases to work with the files. Tho I thought we also compiled those into bundles which should not require this - but its been a long time since I've looked at that flow.

packages/studio-base/package.json
sunrabbit123: Even with the PR, cevek has been quiet.
That's why we posted the new library.

PS.
The recommended library is ts-patch, but it doesn't support TS5 yet.
It looks like it will be supported soon.
The usage is pretty much the same for ttypescript, ttsc, and ts-patch.

PR #5644, opened 2023-03-28
Add toggle to disable zoom scrolling in StateTransitions panel

2metres:
**User-Facing Changes**
Allow user to disable Scroll to zoom behavior on State Transtition panel

Screen Shot 2023-03-28 at 9 40 13 pm

**Description**
Related to #5592 and #5593

### DIFF
--- a/packages/studio-base/src/components/SettingsTreeEditor/FieldEditor.tsx
+++ b/packages/studio-base/src/components/SettingsTreeEditor/FieldEditor.tsx
@@ -15,6 +15,7 @@ import {
   Tooltip,
   TextField,
   ListProps,
+  Button,
 } from "@mui/material";
 import { DeepReadonly } from "ts-essentials";
 import { makeStyles } from "tss-react/mui";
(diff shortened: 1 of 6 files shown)


defunctzombie:
Is the default config applied if there is already a config? I'm wondering if folks that have this panel will end up with a panel that has scroll zoom disabled after this change.

Is there something we can do that does not require the user to manually configure such a setting? Or surface this interaction to the panel as a sort of zoom or scroll lock?

It feels award to me that I'd need to adjust this setting depending on whether I've resized some of my other panels. Nothing in the failed user interaction reported in those tickets would make me think to go find this setting.

2metres:
This does not change the default behavior just introduces the option to disable zooming. Plot syncing and zooming still works, quite wll IMO because the chart only allows zoom not scroll. 

I would like to introduce a way to reset the zoom from within the panel/panel settings. However, it's entangled the TimeBasedChart component and chart syncing.

2metres:
> It feels award to me that I'd need to adjust this setting depending on whether I've resized some of my other panels. Nothing in the failed user interaction reported in those tickets would make me think to go find this setting.

🤷 We could introduce a button in the PanelToolbar?

2metres:
Closing based on feedback from team

Issue #5521, opened 2023-03-09
Vectors (possibly uint8 vectors) of enums don't show enum values


jameskuszmaul-brt:
**Description**

In a flatbuffer MCAP file with enums, ~all of my enums are `uint8`'s under the hood. When I have vectors of enums I don't see the name value translated in any pane, e.g.:
![image](https://user-images.githubusercontent.com/91565539/224185101-a22cb498-73df-4aa6-86c4-1a17d1e6770e.png)
(the faults list in the top panel is an enum).

I haven't tested with enums that are >1 byte.

Current studio.foxglove.dev; also an older version I have locally.

**Steps To Reproduce**
* Have a logfile with messages that contain a vector of 1-byte enums.
* Open a Raw Message panel with that panel; observe that only numbers are displayed.

I can try to produce a more detailed reproduction case later.

**Expected Behavior**

I can see enum names all the time.

jon-chuang:
The javascript object I return from the capnp transcoder is a Number corresponding to the enum. Is the datatypes map meant to translate the enum to a string?

PR #5645, opened 2023-03-28
Add support for showing enums names in arrays in raw message panel

snosenzo:
**User-Facing Changes**

 - Add support for showing enums names in arrays in raw message panel

**Description**
Element indices weren't linking back to parent field names that the enums were stored under. This checks the field name for whether it's a number and if it is uses the parent to grab the constant name instead. 

FG-2427 
Fixes: https://github.com/foxglove/studio/issues/5521 -- and confirmed that this is not specific to flatbuffers.

### DIFF
--- a/packages/studio-base/src/panels/RawMessages/index.tsx
+++ b/packages/studio-base/src/panels/RawMessages/index.tsx
@@ -341,7 +341,7 @@ function RawMessages(props: Props) {
               keyPath.slice(0, -1).reverse(),
             );
             if (childStructureItem) {
-              // if it's an array entry we need most root (first) array field name to get the enum annotation, not the element index
+              // if it's an array index (typeof number) then we want the nearest named array which will be typeof string 
 
               const keyPathIndex = keyPath.findIndex((key) => typeof key === "string");
               const field = keyPath[keyPathIndex];
--- a/packages/studio-base/src/panels/RawMessages/index.tsx
+++ b/packages/studio-base/src/panels/RawMessages/index.tsx
@@ -348,7 +348,7 @@ function RawMessages(props: Props) {
               if (typeof field === "string") {
                 const enumMapping = enumValuesByDatatypeAndField(datatypes);
                 const datatype = childStructureItem.datatype;
-                constantName = enumMapping[datatype]?.[field]![String(itemValue)];
+                constantName = enumMapping[datatype]?.[field]?.[String(itemValue)];
               }
             }
           }
--- a/packages/studio-base/src/panels/RawMessages/index.tsx
+++ b/packages/studio-base/src/panels/RawMessages/index.tsx
@@ -341,7 +341,7 @@ function RawMessages(props: Props) {
               keyPath.slice(0, -1).reverse(),
             );
             if (childStructureItem) {
-              // if it's an array index (typeof number) then we want the nearest named array which will be typeof string 
+              // if it's an array index (typeof number) then we want the nearest named array which will be typeof string
 
               const keyPathIndex = keyPath.findIndex((key) => typeof key === "string");
               const field = keyPath[keyPathIndex];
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/panels/RawMessages/index.tsx
```diff
@@ -341,7 +341,8 @@ function RawMessages(props: Props) {
               keyPath.slice(0, -1).reverse(),
             );
             if (childStructureItem) {
-              const field = keyPath[0];
+              // if it's an array entry we need the array field name to get the enum annotation, not the element index
+              const field = typeof keyPath[0] === "number" ? keyPath[1] : keyPath[0];
```
jtbandes, line 345: what if we have an array of arrays (or deeper)?

jtbandes (approved):

packages/studio-base/src/panels/RawMessages/index.tsx
snosenzo: ah, good catch. I'll update it to find the first non-numeric field in the keypath

packages/studio-base/src/panels/RawMessages/index.tsx
```diff
@@ -341,11 +341,14 @@ function RawMessages(props: Props) {
...
-              const field = keyPath[0];
+              // if it's an array entry we need most root (first) array field name to get the enum annotation, not the element index
+
+              const keyPathIndex = keyPath.findIndex((key) => typeof key === "string");
+              const field = keyPath[keyPathIndex];
               if (typeof field === "string") {
                 const enumMapping = enumValuesByDatatypeAndField(datatypes);
                 const datatype = childStructureItem.datatype;
-                constantName = enumMapping[datatype]?.[field]?.[String(itemValue)];
+                constantName = enumMapping[datatype]?.[field]![String(itemValue)];
```
jtbandes, line 351: Why `!`?

snosenzo: I guess prettier removed it somewhere along the way and I wrongly assumed that there was nothing there previously and thought asserting would fix, but that was not correct. fixed

packages/studio-base/src/panels/RawMessages/index.tsx
```diff
@@ -341,7 +341,10 @@ function RawMessages(props: Props) {
               keyPath.slice(0, -1).reverse(),
             );
             if (childStructureItem) {
-              const field = keyPath[0];
+              // if it's an array entry we need most root (first) array field name to get the enum annotation, not the element index
```
jtbandes, line 344: A little confused by this comment, you say "most root (first)" but it looks like the leaf is at [0], so most root would probably be last? But it seems correct to take the first string one, since we want the nearest array ancestor.

jtbandes (approved):

packages/studio-base/src/panels/RawMessages/index.tsx
snosenzo: yeah, nearest array ancestor is much better. why do I have to be so confusing 😅

snosenzo, line 344:
```suggestion
              // if it's an array index (typeof number) then we want the nearest named array which will be typeof string 
```

packages/studio-base/src/panels/RawMessages/index.tsx
```diff
@@ -341,7 +341,10 @@ function RawMessages(props: Props) {
               keyPath.slice(0, -1).reverse(),
             );
             if (childStructureItem) {
-              const field = keyPath[0];
+              // if it's an array index (typeof number) then we want the nearest named array which will be typeof string
```
snosenzo, line 344:
```suggestion
              // if it's an array index (typeof number) then we want the nearest named array which will be typeof string
```

PR #5647, opened 2023-03-28
Use @foxglove/react to restrict mui/styled

wimagguc:
**User-Facing Changes**

None.

**Description**

`mui/styled` imports are already restricted in https://github.com/foxglove/eslint-plugin/pull/44 so we can remove those rules from the local eslint. Note that the rule exceptions need to use the new `@foxglove/no-restricted-imports` name now.

`NOTE` this is a draft, https://github.com/foxglove/eslint-plugin/pull/44 needs to be merged first.


defunctzombie (commented):
This change needs to be bundled with an upgrade to the newer eslint package right? Otherwise these are not changes we can merge yet?

defunctzombie:
LGTM if this works as designed.

jtbandes (approved):

PR #5652, opened 2023-03-29
Fix initial state in layout browser multiple selection

foxymiles:
**User-Facing Changes**
Fix initial state in layout browser multiple selection

**Description**
Fix initial state in layout browser multiple selection. The multiple selection logic wasn't taking the currently selected layout as the initial state so it required two clicks to start multi selecting. 

This is an issue in both the existing layout browser and the new top nav layout menu.


jtbandes (approved):

amacneil:
👍 I don't fully understand the bug, but I did some testing and can observe on current main cmd-click works fine but shift-click is broken. That seems to be fixed on this branch.

PR #5653, opened 2023-03-29
Batch user node state diagnostics and improve layout switching performance

shuhaowu:
**User-Facing Changes**

Improved layout switching performance when user scripts are present

**Description**

After user nodes are created, the code runs through a loop of `_setUserNodeDiagnostics` for each user script. These in turn calls React set state functions. Since set state functions are not batched in an async function below React 18 (and I see the latest branch is using React 17), this results in synchronous component updates, which in a complex layout can result in a 100ms latency per call. In a complex layout with many scripts (I have 9), this can take about ~2 seconds of run time (see attached screenshot in PR). Note that I used a previous build which also had `_setNodeMetadata` in that loop. This means with the current code and my layout, that would cause a ~1s lag.

Moving this loop into a batched update makes the block only costs 200ms, see below. With the current code without `setNodeMetadata`, this will only be 100ms.

Moving to React 18 should remove the need for this, but it would need to be confirmed.

image

image

### DIFF
--- a/packages/studio-base/src/players/UserNodePlayer/index.ts
+++ b/packages/studio-base/src/players/UserNodePlayer/index.ts
@@ -13,6 +13,7 @@
 
 import { isEqual, uniq } from "lodash";
 import memoizeWeak from "memoize-weak";
+import ReactDOM from "react-dom";
 import shallowequal from "shallowequal";
 import { v4 as uuidv4 } from "uuid";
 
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/players/UserNodePlayer/index.ts
```diff
@@ -748,9 +749,11 @@ export default class UserNodePlayer implements Player {
       this._memoizedNodeDatatypes = nodeDatatypes;
     }
 
-    for (const nodeRegistration of state.nodeRegistrations) {
-      this._setUserNodeDiagnostics(nodeRegistration.nodeId, []);
-    }
+    ReactDOM.unstable_batchedUpdates(() => {
```
defunctzombie, line 752: Please copy some portion of your PR description here so we preserve the notes on performance and state setting for future readers.

defunctzombie (approved):
Once the comment is added and CLA signed we will merge.

snosenzo (approved):

shuhaowu:
> Once the comment is added and CLA signed we will merge.

Both of these are done.

shuhaowu:
Ah I had to add my work email here. Should be good now.

PR #5654, opened 2023-03-29
Save time by not prettifying the generated typesLib

shuhaowu:
**User-Facing Changes**
Improved layout switching performance in the presence of user scripts

**Description**

The writing of user scripts to consume topics in TypeScript is enhanced if the data type information from each topic is compiled into TypeScript  types. All types are compiled into a single file called  `generatedTypes.ts`, which is a read-only file that is likely never opened by the user. The `generatedTypes.ts` file is being passed through a prettifier, which can take upwards of 4-5s during initial loading with 662 topics and 1605 data types in our internal setup.

If we do not prettify the types, the resulting code is not a complete mess either. It looks like the following, which has somewhat improper indentations and that's about it:

```
    /**
     * MessageTypeBySchemaName enumerates the message types for for all the schema names
     * in the current data source.
     *
     * You probably want to use Message<...> instead.
     */
    export type MessageTypeBySchemaName = {

"action_msgs/GoalInfo": {
"goal_id": MessageTypeBySchemaName["unique_identifier_msgs/UUID"],
"stamp": Time,
},
"unique_identifier_msgs/UUID": {
"uuid": Uint8Array,
},
"action_msgs/GoalStatus": {
 // STATUS_UNKNOWN = 0
 // STATUS_ACCEPTED = 1
 // STATUS_EXECUTING = 2
```


jtbandes (approved):

packages/studio-base/src/players/UserNodePlayer/index.ts
```diff
@@ -184,7 +183,7 @@ export default class UserNodePlayer implements Player {
         datatypes: new Map([...basicDatatypes, ...args.datatypes]),
       });
 
-      return await getPrettifiedCode(lib);
```
defunctzombie, line 187: Is it worth setting some upper bound of this file where we return the prettified code (maybe below some size) and after that we don't prettify?

Doing that would also let us leave a comment here in the code around why we do and don't prettify cause I can easily see someone coming along and deciding "oh hey lets prettify" and now we've regressed this performance related change.

shuhaowu: An interesting idea, but at the moment I'm not sure where that threshold should be. My other concern is that we are also running this code on relatively low-end and resource constrained PCs (which is itself shared amongst many software components) which can make selecting an appropriate threshold even more difficult. What if I just put a comment here explaining the lack of code prettification?

defunctzombie:
@shuhaowu Are you able to complete the CLA step?

Did you happen to experiment with doing the prettier work lazily when the user does open the file? Or is the time to run prettier so bad that even in that case it is not worth doing?

defunctzombie:
Thank you for the contribution! I agree - if we want to improve the generated file formatting we have some options but we can revisit that in the future rather than holding up this improvement.

PR #5656, opened 2023-03-29
perf: improve playback performance with the presence of Autocomplete

shuhaowu:
**User-Facing Changes**

Improved playback performance with the presence of Autocomplete

**Description**

When there are many `Autocomplete` elements on screen, each of them needs their width measured. The `Autocomplete` element is being recreated on forward ref and this is happening for every frame during playback. Based on flamegraphs, this is taking about 8% CPU time in a layout with 5 autocomplete elements in a state transition panel, and a simple plot panel. This is enough to cause choppy rendering as the rest of the layout is already consistently taking up about 14-16ms of render time.

This change caches the width in a memo and improves playback performance such the playback is less choppy.

image

### DIFF
--- a/packages/studio-base/src/components/Autocomplete.tsx
+++ b/packages/studio-base/src/components/Autocomplete.tsx
@@ -412,6 +412,17 @@ export default React.forwardRef(function Autocomplete<T = unknown>(
     [blur, items],
   );
 
+  const inputWidth = useMemo(
+    () =>
+      autoSize
+        ? Math.max(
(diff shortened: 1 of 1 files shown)


foxymiles:
Thanks for the PR @shuhaowu!

We just replaced react-autocomplete with the MUI autocomplete in https://github.com/foxglove/studio/pull/5609. Could you please verify if you still see the same performance issues in your testing?

shuhaowu:
Good point. I looked through that code and it doesn't have the exact same problem as this, and this code is obsoleted by that change. Testing this is trickier for me as we haven't picked that change yet.

Issue #5500, opened 2023-03-07
No Plotting of Signals at Player Time When Buffer is Full


turtle-operator:
**Description**

The Plot panel does not load data for the current player time if the buffer is full. Thus, for large data files, there will never be data plotted starting at the time when the buffer reached its limits.

I think a change (in my opinion a regression) was introduced with https://github.com/foxglove/studio/pull/5212. Seems like [this issue](https://github.com/foxglove/studio/pull/5212#issuecomment-1402494604) has not eventually been correctly addressed.

- Version: 1.42.1
- Platform: Web
- Data source format: rosbag

**Steps To Reproduce**

Load a large data file in the web version of Foxglove Studio and add a Plot panel for an arbitrary signal.

**Expected Behavior**

The Plot panel always plots the required data around the current player time, potentially removing any "irrelevant" data (i.e., data far from the current player time) when the buffer reaches its limit.

![image](https://user-images.githubusercontent.com/109816930/223358125-a0e51d04-26b6-44c1-a5dd-8ee110399e67.png)

defunctzombie:
For panels which need to preload data we only have one set of logic and caching that handles all preloaded topics. This is true for preloading data for plots as well as preloading data for transform messages in the 3d panel. These different use-cases are sometimes at odds with each other - for example in the old logic when we had to evict data from preloading this would sometimes evict transforms the 3d panel needs. It would also sometimes cause plots to change from under you as you go through the data.

Another downside of the previous approach was that it would buffer from the current time to the end and then resume from the start which might result in missing plot data right _before_ a time you want to seek to. Arguably, if you know this behavior you can re-seek to an earlier time to re-buffer so there was an out that does not exist now.

Ultimately the answer to what is "irrelevant" data is not an easy one and so for now we've settled on an approach that is easier to reason about: "preload from the start until you fill up your buffer". Where I do think we need to be better is messaging when the buffer is full (something that is maybe not obvious today). We do revisit such changes and if we are able to come up with a strategy that balances the needs of 3d transform preloading with other panels we might change our approach.

> Also, I think it is quite annoying to have to wait until all data from the beginning of the file is loaded, even when directly jumping to a later point of the time series data.

This depends on the data source - for some data sources like our data platform we have separate paths for preloaded data which loads quicker.

defunctzombie:
Notes: https://github.com/foxglove/studio/blob/main/packages/studio-base/src/panels/Plot/index.tsx#L388-L389

[...]

PR #5657, opened 2023-03-29
Improve playback performance with state transition panel

shuhaowu:
**User-Facing Changes**

Improve playback performance with state transition panel

**Description**

The `StateTransitions` panel creates the chartjs datasets from either block data ("horizontal" data) or "vertical"/instantaneous data. This dataset is memo'ed in React based on both the horizontal data (`decodedBlocks`) and the vertical data (`itemsByPath`).  This memo is necessary because creating the dataset can be a time consuming operation (~10-15ms).  Since the vertical data can change every frame, this means the chartjs datasets are being recreated every frame in the memo.  This is not a good idea because most of the time, the memo doesn't even use `itemsByPath` as all data is coming from blocks (as it should? this is a horizontal panel).  This change is a quick "fix" for the problem where we do not recompute the datasets if no paths are using the vertical data and all paths are using the horizontal data. The fix is not ideal, and the reason is given in the code comments. 

I've tried to split the `useMemo` into two memos: one for `itemsByPath` and another for blocks, but the resulting code is even worse. A bigger rework may be needed here but I'm looking for advise on how to proceed.

Collectively, this is taking ~15-20% CPU time and delays every frame by 10-15ms:

image

### DIFF
--- a/packages/studio-base/src/panels/StateTransitions/index.tsx
+++ b/packages/studio-base/src/panels/StateTransitions/index.tsx
@@ -25,7 +25,10 @@ import { add as addTimes, fromSec, subtract as subtractTimes, toSec } from "@fox
 import * as PanelAPI from "@foxglove/studio-base/PanelAPI";
 import { useBlocksByTopic } from "@foxglove/studio-base/PanelAPI";
 import { getTopicsFromPaths } from "@foxglove/studio-base/components/MessagePathSyntax/parseRosPath";
-import { useDecodeMessagePathsForMessagesByTopic } from "@foxglove/studio-base/components/MessagePathSyntax/useCachedGetMessagePathDataItems";
+import {
+  MessageDataItemsByPath,
+  useDecodeMessagePathsForMessagesByTopic,
+} from "@foxglove/studio-base/components/MessagePathSyntax/useCachedGetMessagePathDataItems";
 import useMessagesByPath from "@foxglove/studio-base/components/MessagePathSyntax/useMessagesByPath";
 import {
   MessagePipelineContext,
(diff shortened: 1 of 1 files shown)


foxymiles:
Thanks for the detective work and the fix! What kind of machine are you doing these measurements on?

foxymiles:
I think your fix makes sense here. I've tidied up the logic a bit in this branch. Does this seem ok to you?

https://github.com/foxglove/studio/tree/miles/state-transition-memo-fix

foxymiles:
@shuhaowu Yes I think you're right this separation is not ideal. And I may have to break this separation to fix another issue that's come up so this might only be a stopgap fix.

defunctzombie:
> My impression is that, ideally, all data should be loaded via `decodedBlocks` so the entire history can be loaded all at once.

If the preloading buffers get full then you run into a situation where no more data is plotted. This depends on the preloading approach which we recently changed to preload from start rather than constantly change where we preload from.

See https://github.com/foxglove/studio/issues/5500 for a related issue this surfaced and some of the notes I left on this around what our reasoning for the change was.

foxymiles:
@shuhaowu I've updated my https://github.com/foxglove/studio/tree/miles/state-transition-memo-fix to incorporate your fix and also merge in the current main brain so it can be merged back to main. Do you want to pull this into your PR here so we can merge it or would you prefer I open a separate PR on my branch we can merge? In either case you'd get contribution credit on the commit.

foxymiles:
Incorporated into https://github.com/foxglove/studio/pull/6008. Thanks @shuhaowu !

PR #5659, opened 2023-03-29
Remove tabIndex from menu subheads

bryfox:
**User-Facing Changes**

This removes the explicit `tabindex` set on help menu subheaders, which cause them to be focusable and navigable by keyboard (arrow) navigation. Only menu items should be focusable.


jtbandes (approved):

esthersweon:
Will not include in changelog, just FYI.

PR #5661, opened 2023-03-29
Add experimental Image panel

jtbandes:
**User-Facing Changes**
None

**Description**
Adds `interfaceMode: "3d" | "image"` to ThreeDeeRender component, and a new "3D Image" panel under a feature flag.

For now the only difference is that the 3D/measure/publish tools are hidden:

image

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
+++ b/packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
@@ -144,9 +144,6 @@ function RendererOverlay(props: {
   const [interactionsTabType, setInteractionsTabType] = useState<TabType | undefined>(undefined);
   const renderer = useRenderer();
 
-  // Publish control is only available if the canPublish prop is true and we have a fixed frame in the renderer
-  const showPublishControl: boolean = props.canPublish && renderer?.fixedFrameId != undefined;
-
   // Toggle object selection mode on/off in the renderer
   useEffect(() => {
     if (renderer) {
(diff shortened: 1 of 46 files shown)


packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
```diff
@@ -255,92 +257,94 @@ function RendererOverlay(props: {
...
+              data-testid="measure-button"
+              className={classes.iconButton}
+              color={props.measureActive ? "info" : "inherit"}
+              title={props.measureActive ? "Cancel measuring" : "Measure distance"}
+              onClick={props.onClickMeasure}
+            >
+              <Ruler24Filled className={classes.rulerIcon} />
+            </IconButton>
+
+            {showPublishControl && (
```
snosenzo, line 280: I think i'd prefer a separate condition here for a little better readability and less nesting

packages/studio-base/src/panels/ThreeDeeRender/index.tsx
```diff
@@ -46,7 +54,16 @@ function ThreeDeeRenderAdapter(props: Props) {
   );
 }
 
-ThreeDeeRenderAdapter.panelType = "3D";
-ThreeDeeRenderAdapter.defaultConfig = {};
+export const ThreeDeePanel = Panel(
+  Object.assign(ThreeDeeRenderAdapter.bind(undefined, "3d"), {
+    panelType: "3D",
+    defaultConfig: {},
```
snosenzo, line 60: pros/cons on this just being part of config?

jtbandes: We still need to have an "Image" item in the "add panel" menu so I'm not totally sure how that would change based on the config. I guess we could make the Image menu item load a 3D panel with a different preset config?

jtbandes: I guess I'd be more inclined to add a panel setting that switches between old image and new image, rather than a switch between 3d and new image. But that felt like it would be significantly more complicated.

snosenzo: I meant more along the lines of instead of passing `interfaceMode` as a prop to the component having it be part of the `defaultConfig` / `config`

snosenzo (approved):

Issue #5484, opened 2023-03-06
Switching the layout will cause the reset view to be in a different position


Woodii1998:
**Description**
Switching the layout will cause the reset view to be in a different position

https://user-images.githubusercontent.com/33593572/223032121-3e2cb720-ab45-4622-b02c-08dec274b17d.mp4

layout: 
https://drive.google.com/file/d/1WGfS87CKpwmH5ER3e8I4uLwvjV6qa19z/view?usp=sharing
https://drive.google.com/file/d/1pz_9Dtlzh0JHLKDtwtMnr6K5HNHkFvpW/view?usp=sharing

bag file:
https://drive.google.com/file/d/1FC5J5EucTOQ35G0J1-UdhWQVL0JXP_Bu/view?usp=sharing

- Version: online version
- Platform: web

**Steps To Reproduce**

Like in the video

ps: Caching will affect the reproduction of the problem, to ensure that you can reproduce it every time you can use chrome's empty cache and hard reload like in the video

**Expected Behavior**
Consistent effect before and after switching layout

foxymiles:
It looks to me like something has gone wrong earlier in this flow. Your initial view is this:

<img width="1728" alt="Screenshot 2023-03-21 at 9 44 04 AM" src="https://user-images.githubusercontent.com/93935560/226500770-0d39cc38-8089-41d6-a0dc-342d6ae3194c.png">

But when you reload you have a different view like this:
<img width="1728" alt="Screenshot 2023-03-21 at 9 44 39 AM" src="https://user-images.githubusercontent.com/93935560/226500793-31e05449-0973-404a-bd24-7a9844969d8a.png">

Even though the camera settings are the same.

foxymiles:
Is it possible for you to privately share the layouts involved here?

libertychisasuro:
@Woodii1998 looks like the link to your bag file is invalid

foxymiles:
> Is it possible for you to privately share the layouts involved here?

Oh nevermind I see you've attached them I'll take a look.

Woodii1998:
> @Woodii1998 looks like the link to your bag file is invalid

I seem to have deleted my files by mistake, you can try with this link

https://drive.google.com/file/d/1FC5J5EucTOQ35G0J1-UdhWQVL0JXP_Bu/view?usp=sharing

Woodii1998:
> > Is it possible for you to privately share the layouts involved here?
> 
> Oh nevermind I see you've attached them I'll take a look.

[...]

PR #5662, opened 2023-03-29
Fix issue with "Fixed" follow mode when display frame is absent

snosenzo:
**User-Facing Changes**

 - fix fixed follow mode from centering on root frame when display frame does not exist.

**Description**
 - the unfollowPose Snapshot was being set to an identity pose because transforms would come in with the display frame as it's parent pose, which would create the frame, but it would have no transforms in it. This function only checked to see that the display frame exists and not that it has transforms in it before rendering. This was causing the camera to center on the root frame instead of the display frame. Checking for this before setting the unfollowPoseSnapshot corrects this behavior.

FG-2308
#5484

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
@@ -1472,7 +1472,12 @@ export class Renderer extends EventEmitter<RendererEvents> {
     }
 
     // Should only occur on reload when the saved followMode is not follow
-    if (this.followMode !== "follow-pose" && !this.unfollowPoseSnapshot) {
+    if (
+      this.followMode !== "follow-pose" &&
+      !this.unfollowPoseSnapshot &&
+      // only record snapshot if the frame has transforms, otherwise it will be identity pose
+      frame.transformsSize() > 0
+    ) {
       // Snapshot the current pose of the render frame in the fixed frame
       this.unfollowPoseSnapshot = makePose();
       fixedFrame.applyLocal(


packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -1471,8 +1471,15 @@ export class Renderer extends EventEmitter<RendererEvents> {
       this.fixedFrameId = fixedFrameId;
     }
 
+    const renderFrameIsRoot = this.fixedFrameId === this.renderFrameId;
     // Should only occur on reload when the saved followMode is not follow
-    if (this.followMode !== "follow-pose" && !this.unfollowPoseSnapshot) {
+    if (
+      this.followMode !== "follow-pose" &&
+      !this.unfollowPoseSnapshot &&
+      // only record snapshot if the frame has transforms, otherwise it will be identity pose
```
jtbandes, line 1479: I'm confused by this because I would expect the apply to fail if the frame has no transforms. Is this happening because we're not doing `this.unfollowPoseSnapshot = fixedFrame.applyLocal(...)` but just discarding the return value?

snosenzo: `fixedFrame.applyLocal(...)` returns the same pose. I believe this is supposed to work because frames without transforms are considered root frames and valid.

packages/studio-base/src/panels/ThreeDeeRender/Renderer.test.ts
```diff
@@ -115,6 +115,42 @@ describe("Renderer", () => {
...
+
+    // record to make sure it changes when there's a new fixed frame
+    let lastUnfollowPoseSnapshot = renderer.unfollowPoseSnapshot;
+
+    const tfWithDisplayChild = createTFMessageEvent("parentOfDisplay", "display", 1n, [1n]);
+    tfWithDisplayChild.message.transforms[0]!.transform.translation.x = 1;
+    renderer.addMessageEvent(tfWithDisplayChild);
+    renderer.animationFrame();
+    expect(renderer.fixedFrameId).toEqual("parentOfDisplay");
+    expect(renderer.unfollowPoseSnapshot).not.toEqual(lastUnfollowPoseSnapshot);
```
jtbandes, line 143: Would it make sense to assert on the specific values of the pose snapshot? It looks like you've chosen simple enough transforms that we should be able to hard-code the expected snapshot here.

jtbandes (approved):

PR #5665, opened 2023-03-30
Fix plot Y axis scale jittering

foxymiles:
**User-Facing Changes**
Fixes an issue with plot panel Y scale jittering.

**Description**
Update global bounds and rebuild datasets only when the chart reports an update to the X scale. We can skip this process when it automatically adjusts its Y scale and avoid this boundary condition feedback loop.

### DIFF
--- a/packages/studio-base/src/components/TimeBasedChart/index.tsx
+++ b/packages/studio-base/src/components/TimeBasedChart/index.tsx
@@ -681,11 +681,15 @@ export default function TimeBasedChart(props: Props): JSX.Element {
       }
 
       // If this is an update from the chart adjusting its own bounds and not a
-      // user interaction and no changes have been made to the X scale we can
+      // user interaction and the X scale is defined but hasn't changed we can
       // skip updating global bounds and downsampling. This avoids a feedback
       // loop on boundary conditions when the chart is adjusting its own Y axis
       // to fit the dataset.
-      if (isEqual(scales.x, currentScalesRef.current?.x) && !userInteraction) {
+      if (
+        scales.x != undefined &&
+        isEqual(scales.x, currentScalesRef.current?.x) &&
+        !userInteraction
+      ) {
         return;
       }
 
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/components/TimeBasedChart/index.tsx
```diff
@@ -679,6 +680,19 @@ export default function TimeBasedChart(props: Props): JSX.Element {
         setHasUserPannedOrZoomed(true);
       }
 
+      // If this is an update from the chart adjusting its own bounds and not a
+      // user interaction and the X scale is defined but hasn't changed we can
+      // skip updating global bounds and downsampling. This avoids a feedback
+      // loop on boundary conditions when the chart is adjusting its own Y axis
```
snosenzo, line 686: Could the X axis also be internally adjusted without user interaction (I think a live source would do this?)? Could there be an issue there?

foxymiles: Good question. I tested this also and it's working.

snosenzo (approved):

packages/studio-base/src/components/TimeBasedChart/index.tsx
```diff
@@ -679,6 +680,19 @@ export default function TimeBasedChart(props: Props): JSX.Element {
...
+      // to fit the dataset.
+      if (
+        scales.x != undefined &&
+        isEqual(scales.x, currentScalesRef.current?.x) &&
+        !userInteraction
+      ) {
+        return;
```
defunctzombie, line 693: Does this prevent updated `currentScalesRef` with the new _y_ value since `scales` contains `x` and `y` values?

defunctzombie (approved):

packages/studio-base/src/components/TimeBasedChart/index.tsx
foxymiles: Yes although this doesn't seem to break anything since it only happens in the case where the scale has auto-updated its Y scale and left the X scale untouched. Moving the `currentScalesRef` update ahead of this check also doesn't seem to break anything and is maybe a bit easier to reason about so that it always reflects the current scale value. I'll adjust this logic.

I think this code might be a bit cleaner if we make an explicit distinction between scales that we receive from the chart and scales we set ourselves but we could handle that in a future refactor.

PR #5670, opened 2023-03-30
Add fallback root frame when there are no transforms present

snosenzo:
**User-Facing Changes**

 - add fallback root frame when there are no transforms present

**Description**
This adds a fallback NONE_ROOT_FRAME to the transform tree to use when `renderFrameId` remains undefined after trying to find a default among options or is otherwise unset. This transform isn't visible anywhere in the UI. The fixed frame and render frame use this as a fallback which also allows the custom `Grid` layer to be rendered in it when using the `< Display Frame >` frame setting.

FG-1786


packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -1202,11 +1211,9 @@ export class Renderer extends EventEmitter<RendererEvents> {
     camera.layers.set(LAYER_DEFAULT);
     this.selectionBackdrop.visible = this.selectedRenderable != undefined;
 
-    const renderFrameId = this.renderFrameId;
-    const fixedFrameId = this.fixedFrameId;
-    if (renderFrameId == undefined || fixedFrameId == undefined) {
-      return;
-    }
+    // use the NONE_ROOT_FRAME_ID if renderFrame is undefined and there are no options for transforms
```
snosenzo, line 1214:
```suggestion
    // use the FALLBACK_FRAME_ID if renderFrame is undefined and there are no options for transforms
```

packages/studio-base/src/panels/ThreeDeeRender/renderables/FrameAxes.ts
```diff
@@ -502,7 +502,7 @@ function createLineGeometry(): LineGeometry {
 }
 
 function buildSettingsFields(
-  frame: CoordinateFrame | undefined,
+  frame: CoordinateFrame<string> | undefined,
```
jtbandes, line 505: might want to find/replace all `CoordinateFrame<string>` to `CoordinateFrame<UserFrameId>`?

packages/studio-base/src/panels/ThreeDeeRender/transforms/TransformTree.ts
```diff
@@ -144,12 +152,18 @@ export class TransformTree {
     }
   }
 
-  public hasFrame(id: string): boolean {
+  public hasFrame(id: CoordinateFrame<AnyFrameId>["id"]): boolean {
```
jtbandes, line 155:
```suggestion
  public hasFrame(id: AnyFrameId): boolean {
```

packages/studio-base/src/panels/ThreeDeeRender/transforms/TransformTree.ts
```diff
@@ -144,12 +152,18 @@ export class TransformTree {
...
+    if (id === CoordinateFrame.FALLBACK_FRAME_ID) {
+      return true;
```
jtbandes, line 157: I wonder if we should actually return false if there are some user frames present (and return undefined from `frame()`)? Not sure it would make a difference in practice though.

jtbandes (approved):

packages/studio-base/src/panels/ThreeDeeRender/transforms/TransformTree.ts
snosenzo: Hmmm. I think I'm going to leave it as is since if we're ever checking for a fallback frame's existence in the transform tree (which we shouldn't be), we will probably try to find another fallback frame or early return if this returns false which might hide the fact that we're checking it. I think showing something in the fallback frame will be a bit more obvious of an error because none of the other transforms will have a path to it in the tree

PR #5672, opened 2023-03-30
Added connection variable to PanelExtensionAdapter

err4o4:
**User-Facing Changes**
Added connection information to context in PanelExtensionContext (`context.connection`)

**Description**
This changes gives ability to access connection information from `PanelExtensionAdapter`. It is useful in custom extension development

### DIFF
--- a/packages/studio-base/src/components/PanelExtensionAdapter/PanelExtensionAdapter.tsx
+++ b/packages/studio-base/src/components/PanelExtensionAdapter/PanelExtensionAdapter.tsx
@@ -93,7 +93,7 @@ function PanelExtensionAdapter(props: PanelExtensionAdapterProps): JSX.Element {
   const { playerState, pauseFrame, setSubscriptions, seekPlayback, sortedTopics } =
     messagePipelineContext;
 
-  const { capabilities, profile: dataSourceProfile } = playerState;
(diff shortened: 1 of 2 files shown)


defunctzombie:
This will cause the entire extension to re-create whenever the urlState object changes which is not desired (from my perspective).

I am currently 👎 on this change since we consider the player url state something internal to the player and the PR description does not detail any _why_ this is useful for extension development. It would be helpful if you could provide some context on which parts of the urlState are useful since many parts of the url state are already exposed through other fields to extensions.

err4o4:
The main purpose for having connection info is to build custom panels which use this info. For example:
1. WebRTC panel which requires ws connection to the robot and it's not rosbridge or foxglove-ws
2. Custom control panel which requires REST API. This couldn't be done via rosbridge or foxglove-ws because it's used by other standalone software

defunctzombie:
Thank you for make a PR to propose this idea. Our current architecture in studio is that the data flows through studio data sources rather than external data sources. While things like what you have are possible and cool to experiment, we are not ready to expose what is effectively the unstructured "url state" to the panels. For connections that want to talk back to the robot or host we tend to recommend using the data sources publishing capabilities or service capabilities.

PR #5675, opened 2023-03-31
Add data source info in the playback bar

2metres:
**User-Facing Changes**
None (hidden behind feature flag)

**Description**
Adds icon that shows Data Source Info on hover in playback bar

Screen Shot 2023-04-01 at 8 21 08 am


defunctzombie:
👍 Since it is behind the feature flag we'll have time to re-visit this in future review sessions.

defunctzombie (approved):

PR #5676, opened 2023-03-31
Chinese translations for some panel settings

Woodii1998:
**User-Facing Changes**
Added more Chinese translations.

**Description**
Translated 3D, Plot, and Log panel settings, data source info, and open dialog.


packages/studio-base/src/components/PanelSettings/index.tsx
```diff
@@ -205,9 +216,9 @@ export default function PanelSettings({
               paddingX={enableNewTopNav ? 1 : 0}
             >
               <Typography variant="body2" color="text.secondary" align="center">
-                {`${
-                  panelInfo ? `The ${panelInfo.title}` : "This"
-                } panel does not have any settings`}
+                {t("panelNoAnySettings", {
+                  panelTitle: panelInfo ? t("panelTitle", { title: panelInfo.title }) : t("this"),
+                })}
```
jtbandes, line 221: We shouldn't break apart sentences like this for translation. Since it's not very important to use the panel title here, I will simplify by changing it to a single sentence.

packages/studio-base/src/components/PanelSettings/ActionMenu.tsx
```diff
@@ -50,15 +52,15 @@ export function ActionMenu({
             handleClose();
           }}
         >
-          Import/export settings...
+          {t("importOrExportSettings")}...
```
jtbandes, line 55: should the `...` be part of the string too?

packages/studio-base/src/providers/PanelStateContextProvider.tsx
```diff
@@ -197,7 +198,13 @@ const selectPanelTypesInUse = createSelector(selectLayoutConfigById, (config) =>
 });
 
 export function PanelStateContextProvider({ children }: { children?: ReactNode }): JSX.Element {
-  const [store] = useState(createPanelStateStore());
+  const [store, setStore] = useState(createPanelStateStore());
+
+  // recreate panel state store on language change
```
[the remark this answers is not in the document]
Woodii1998: > What's the reason for this?

Here, we need to reload the settings node when switching languages to ensure that the language in the settings is updated. You can try commenting out this code and then switching languages to understand what I mean. However, perhaps I am not very familiar with the code of the entire project, and maybe you have a better way to refresh the settings node?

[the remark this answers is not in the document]
jtbandes: I think that if `t` is a dependency of `useEffect`, then the panel should automatically call `updatePanelSettingsTree()` again when the language changes, which will update the settings nodes

packages/studio-base/src/i18n/zh/panelSettings.ts
```diff
@@ -0,0 +1,122 @@
...
+  labelScaleHelp: "应用于所有标签的比例因子",
+  ignoreColladaUpAxis: "忽略COLLADA的<up_axis>",
+  ignoreColladaUpAxisHelp: "通过忽略COLLADA文件中的<up_axis>标记，匹配rviz的行为",
+  takeEffectAfterReboot: "此设置需要重新启动以生效",
+  syncCamera: "同步相机",
+  syncCameraHelp: "将相机与其他启用此设置的面板同步。",
+  meshUpAxis: "网格上轴",
+  meshUpAxisHelp: "加载缺少方向信息的网格（STL和OBJ）时使用的“上”方向",
+  YUp: "Y-up",
+  ZUp: "Z-up",
```
jtbandes, line 38: Is it normal to say "up" in english?
```suggestion
  YUp: "Y上",
  ZUp: "Z上",
```

packages/studio-base/src/components/PanelSettings/index.tsx
jtbandes: You can change the english version to `This panel does not have any settings` and remove the panelInfo.title :)

packages/studio-base/src/components/SettingsTreeEditor/index.tsx
```diff
@@ -120,7 +122,7 @@ export default function SettingsTreeEditor({
             value={filterText}
             className={classes.textField}
             fullWidth
-            placeholder="Seach panel settings…"
```
jtbandes, line 123: Thanks for fixing this typo 🤣

packages/studio-base/src/i18n/zh/panelSettings.ts
Woodii1998: > Is it normal to say "up" in english?

In fact, for very simple English or nouns, the Chinese in general will also use the English name, and here the meaning is actually difficult to be accurately translated into Chinese, in Chinese '上' is a lot of meaning, so I think that rather than ambiguous forced translation, it is easier to use English directly to make people understand, in fact, many agreed vocabulary even in China we will also use English

packages/studio-base/src/i18n/zh/panelSettings.ts
```diff
@@ -0,0 +1,122 @@
...
+  displayFrameHelp: "放置相机用于显示坐标系。相机的位置和方向将相对于该坐标系的原点。",
+  followMode: "跟踪模式",
+  followModeHelp: "更改回放期间相机的行为，是否跟踪显示坐标系。",
+  scene: "场景",
+  reset: "重置",
+  renderStats: "渲染统计",
+  background: "背景",
+  labelScale: "标签比例",
+  labelScaleHelp: "应用于所有标签的比例因子",
+  ignoreColladaUpAxis: "忽略COLLADA的<up_axis>",
```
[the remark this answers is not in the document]
Woodii1998: > Thanks for translating these. I think it will be important to keep each panel (3D, Plot, etc.) in a separate file, because each panel is complex and they work as almost completely separate modules (some day they might be in a separate package).

I understand what you mean, I will split these translated texts later

packages/studio-base/src/providers/PanelStateContextProvider.tsx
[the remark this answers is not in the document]
Woodii1998: > I think that if `t` is a dependency of `useEffect`, then the panel should automatically call `updatePanelSettingsTree()` again when the language changes, which will update the settings nodes

I agree with you,  I will try to modify it as you want

packages/studio-base/src/components/PanelSettings/ActionMenu.tsx
Woodii1998: > should the `...` be part of the string too?

Well I'm being lazy here, my idea is to reuse the translations that are already available, so that there is no need for the 'Import/export settings' and 'Import/export settings...' As a non-native English speaking programmer I often struggle with how to name variables appropriately 🤣

packages/studio-base/src/providers/PanelStateContextProvider.tsx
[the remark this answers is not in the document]
Woodii1998: > I think that if `t` is a dependency of `useEffect`, then the panel should automatically call `updatePanelSettingsTree()` again when the language changes, which will update the settings nodes

It's been changed, but I still think that instead of monitoring `t` and calling `updatePanelSettingsTree` in each place, it would be better to update it uniformly as I wrote it in the beginning

Woodii1998:
@jtbandes I fixed the above issue and added the translation of the dataSource section, please help review it

packages/studio-base/src/providers/PanelStateContextProvider.tsx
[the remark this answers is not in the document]
jtbandes: cc @foxymiles - any thoughts on this?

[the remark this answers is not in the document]
foxymiles: What was the proposed change exactly? It looks like it's no longer in the PR.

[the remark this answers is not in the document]
jtbandes: It was basically doing `setStore(createPanelStateStore())` to reinitialize the store when the language changed. I've removed it for now since it seems like adding `t` as a useEffect dependency is generally working, but we can revisit in the future if we learn more.

packages/studio-base/src/i18n/zh/panelSettings.ts
```diff
@@ -0,0 +1,19 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { TypeOptions } from "i18next";
+
+export const panelSettings: TypeOptions["resources"]["panelSettings"] = {
+  importOrExportSettings: "导入/导出设置",
+  importOrExportSettingsWithEllipsis: "导入/导出设置...",
```
jtbandes, line 9: I was reading https://en.wikipedia.org/wiki/Chinese_punctuation and it says that it's actually common for Chinese to use 6 dots: `......` or `……` or `⋯⋯` (centered). Do you think it's a good idea to do that?

packages/studio-base/src/i18n/zh/plot.ts
```diff
@@ -0,0 +1,38 @@
...
+  deleteSeries: "删除数据系列",
+  messagePath: "消息地址",
+  label: "标签",
+  color: "颜色",
+  timestamp: "时间戳",
+  receiveTime: "接收时间",
+  headerStamp: "头时间戳",
+  series: "数据系列",
+  addSeries: "添加数据系列",
+  maxYError: "Y 的最大值必须大于 Y 的最小值。",
```
jtbandes, line 17: I noticed some inconsistent use of whitespace, e.g. `Y 的最大值` vs `ROS 2的本地连接`. Do you think it should be `Y的最大值`?

packages/studio-base/src/i18n/zh/threeDee.ts
```diff
@@ -0,0 +1,73 @@
...
+  scene: "场景",
+  renderStats: "渲染统计",
+  background: "背景",
+  labelScale: "标签比例",
+  labelScaleHelp: "应用于所有标签的比例因子",
+  ignoreColladaUpAxis: "忽略COLLADA的<up_axis>",
+  ignoreColladaUpAxisHelp: "通过忽略COLLADA文件中的<up_axis>标记，匹配rviz的行为",
+  takeEffectAfterReboot: "此设置需要重新启动以生效",
+  meshUpAxis: "网格上轴",
+  meshUpAxisHelp: "加载缺少方向信息的网格（STL和OBJ）时使用的“上”方向",
```
jtbandes, line 45: Do you think it's better to use `“”`, `「」`, `《》` etc.?

packages/studio-base/src/i18n/zh/openDialog.ts
```diff
@@ -0,0 +1,40 @@
...
+  viewOurDocs: "查看我们的文档",
+  joinOurCommunity: "加入我们的社区",
+  joinOurCommunityDescription: "在 Slack 上与我们的开发人员和用户交流，或在 GitHub 上提出问题。",
+  joinOurSlack: "加入我们的 Slack",
+  openAGitHubIssue: "打开 GitHub issue",
+  needHelp: "需要帮助?",
+  needHelpDescription: "查看我们的文档，或查看Foxglove博客上的教程。",
+  seeTutorials: "请参阅 '教程'。",
+  collaborateTitle: "使用 Foxglove 数据平台加速开发",
+  secureStorageOfData: "安全地存储PB级的ROS或自定义数据",
```
jtbandes, line 21: According to Google Translate, petabytes is "拍字节". Do you think it's better to use PB?

packages/studio-base/src/i18n/zh/openDialog.ts
```diff
@@ -0,0 +1,40 @@
...
+  newToFoxgloveStudioDescription: "从探索一个样本数据集开始，或查看我们的文档。",
+  exploreSampleData: "探索样本数据",
+  viewOurDocs: "查看我们的文档",
+  joinOurCommunity: "加入我们的社区",
+  joinOurCommunityDescription: "在 Slack 上与我们的开发人员和用户交流，或在 GitHub 上提出问题。",
+  joinOurSlack: "加入我们的 Slack",
+  openAGitHubIssue: "打开 GitHub issue",
+  needHelp: "需要帮助?",
+  needHelpDescription: "查看我们的文档，或查看Foxglove博客上的教程。",
+  seeTutorials: "请参阅 '教程'。",
```
jtbandes, line 19: I think this is wrong, it should be `教程` or `看教程`, or `阅读教程` according to Google Translate. What do you think?

packages/studio-base/src/i18n/zh/openDialog.ts
```diff
@@ -0,0 +1,40 @@
...
+  // Start
+  newToFoxgloveStudio: "第一次使用 Foxglove Studio?",
+  newToFoxgloveStudioDescription: "从探索一个样本数据集开始，或查看我们的文档。",
+  exploreSampleData: "探索样本数据",
+  viewOurDocs: "查看我们的文档",
+  joinOurCommunity: "加入我们的社区",
+  joinOurCommunityDescription: "在 Slack 上与我们的开发人员和用户交流，或在 GitHub 上提出问题。",
+  joinOurSlack: "加入我们的 Slack",
+  openAGitHubIssue: "打开 GitHub issue",
+  needHelp: "需要帮助?",
```
jtbandes, line 17: Should we use `？` U+FF1F?

packages/studio-base/src/i18n/zh/openDialog.ts
```diff
@@ -0,0 +1,40 @@
...
+  uploadToDataPlatform: "上传到数据平台",
+  shareLayouts: "分享布局",
+  openLocalFile: "打开本地文件",
+  openLocalFileDescription: "直接从本地文件系统可视化数据。",
+  openUrl: "上传和共享数据",
+  openUrlDescription: "使用 Foxglove 数据平台与您的团队共享数据。",
+  openConnection: "打开连接",
+  openConnectionDescription: "连接到实时机器人或服务器。",
+  openDataSource: "打开数据源",
+  recentDataSources: "最近的数据源",
```
jtbandes, line 39: What do you think about 最近的数据源 vs 最近使用的数据源? Is it normal to omit the verb?

jtbandes (commented):
@Woodii1998 I just pushed several updates! Can you address the "TODO" comments for missing translations? I also posted some questions about the translations in inline comments.

packages/studio-base/src/i18n/zh/plot.ts
Woodii1998: You reminded me that it is now more recommended to add a space between Chinese and English in the Chinese context. We even have a name for this space called "盘古之白" (”Pangu's blank space“ Pangu is the Creator of the universe in Chinese mythology). There is even a [Github project]('https://github.com/vinta/pangu.js') dedicated to this matter. So actually, there is an issue with the translation that follows. It should be translated as "ROS2 的本地连接." I seem to have forgotten this rule in some of my translations. I will double-check.

packages/studio-base/src/i18n/zh/threeDee.ts
Woodii1998: I think “” is more appropriate. All three symbols are used for emphasis, but 「」is rarely used in modern Chinese, while 《》 is more often used in book titles or movie titles.

packages/studio-base/src/i18n/zh/openDialog.ts
Woodii1998: See this message I just know for the first time that the original PB in Chinese can also be translated as "拍字节"😄, I believe I use PB here is the most appropriate, Chinese textbooks are also directly referred to with PB

packages/studio-base/src/i18n/zh/openDialog.ts
Woodii1998: You are right, adding the verb will make the translation seem more accurate, please forgive me for not being a professional

packages/studio-base/src/i18n/zh/openDialog.ts
Woodii1998: The translation here is correct. "请参阅" can be understood as "please refer to and read" in English

packages/studio-base/src/i18n/zh/openDialog.ts
jtbandes: The English version is also not very formal ("recent data sources" instead of "recently-used data sources"). So I think it's fine to use less formal language if appropriate. I just asked because I don't know what is normal for Chinese software 😄

packages/studio-base/src/i18n/zh/openDialog.ts
jtbandes: I don't understand why 请 is used here, but not for the other buttons, for example "请打开 GitHub issue"?

Why do we use `''` and `。` for this button?

packages/studio-base/src/i18n/zh/panelSettings.ts
Woodii1998: You're right. Because few people pay special attention to this matter in daily life, I ignored this issue. Using "……" is indeed a more official and correct way to write

jtbandes: Is …… commonly used in Chinese software?

packages/studio-base/src/i18n/zh/openDialog.ts
```diff
@@ -0,0 +1,40 @@
...
+  collaborateTitle: "使用 Foxglove 数据平台加速开发",
+  secureStorageOfData: "安全地存储PB级的ROS或自定义数据",
+  convenientWebInterface: "使用一个方便的网络界面，以闪电般的速度标记、搜索和检索数据",
+  canBeShared: "与队友分享数据文件、可视化布局和自定义扩展程序",
+  createAFreeAccount: "创建免费帐户",
+  signIn: "登录",
+  learnMore: "了解更多",
+  startCollaborating: "与您的 Foxglove 组织开始合作",
+  startCollaboratingDescription:
+    "充分利用你的Foxglove账户--无论你想深入研究你的数据还是与你的团队成员分享工具。",
```
jtbandes, line 29: should it be —— instead of -- (also from https://en.wikipedia.org/wiki/Chinese_punctuation)?

Woodii1998: I checked and both ways of writing are correct in the official regulations (https://www.w3.org/TR/clreq/#indication_punctuation_marks)

Woodii1998: Some Chinese materials
https://zh.wikipedia.org/wiki/%E7%A0%B4%E6%8A%98%E5%8F%B7
https://baike.baidu.com/item/%E7%A0%B4%E6%8A%98%E5%8F%B7/999304

packages/studio-base/src/i18n/zh/openDialog.ts
Woodii1998: You are right, reconsider the translation here

packages/studio-base/src/i18n/zh/openDialog.ts
jtbandes: It says:

> Two-em dashes [——] can be created using U+2E3A TWO-EM DASH [⸺] or two adjacent U+2014 EM DASH [—] characters. Both alternatives should take 2 em.

But "--"  is U+002D, so I think it's better to use "——" (2x U+2014) or "⸺" (1x U+2E3A)

packages/studio-base/src/i18n/zh/preferences.ts
```diff
@@ -35,4 +37,11 @@ export const preferences = {
   restartTheAppForChangesToTakeEffect: " 重新启动应用程序以使更改生效。",
   layoutDebugging: "布局调试",
   layoutDebuggingDescription: "显示用于开发和调试布局存储的额外控件。",
+  extensions: "扩展",
+  about: "关于",
+  noExperimentalFeatures: "目前没有实验性的功能。",
+  ros2NativeConnection: "ROS2 的本地连接",
+  ros2NativeConnectionDescription: "启用已废弃的 ROS2 本地连接器",
```
jtbandes, line 44:
```suggestion
  ros2NativeConnection: "ROS 2 的本地连接",
  ros2NativeConnectionDescription: "启用已废弃的 ROS 2 本地连接器",
```

😄 
https://twitter.com/OpenRoboticsOrg/status/1629286675649273856
https://twitter.com/OpenRoboticsOrg/status/1629215777525223424
https://twitter.com/OpenRoboticsOrg/status/1629208251563929600

jtbandes (approved):

PR #5678, opened 2023-03-31
Disable wheel events on state transition chart wrapper

defunctzombie:
**User-Facing Changes**
Fix zoom and scroll behavior for StateTransition panel when there are series outside of the visible panel area.

**Description**
See code comment for description.

Fixes: #5592


jtbandes (approved):
🤷 worth trying

2metres:
Not sure that this resolves the problem. We've fixed crazy scrolling with inaccessable rows outside the viewable area of the chart.

defunctzombie:
> We've fixed crazy scrolling with inaccessable rows outside the viewable area of the chart.

You can access the rows by using the scrollbar which is visible on the panel. With https://github.com/foxglove/studio/pull/5680 it becomes even more visible.

PR #5679, opened 2023-03-31
Fix Foxglove Websocket not reconnecting when server restarted

achim-k:
**User-Facing Changes**
- Fix Foxglove Websocket not reconnecting when server restarted

**Description**
- Fixes that the websocket worker adapter does not receive the `close` signal from the websocket when `close()` is called
- Abort unsuccessful connection attempts after 10 seconds

### DIFF
--- a/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
+++ b/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
@@ -161,7 +161,9 @@ export default class FoxgloveWebSocketPlayer implements Player {
     }
     log.info(`Opening connection to ${this._url}`);
 
-    // Set a timeout to abort the connection if we are still not connected by then
+    // Set a timeout to abort the connection if we are still not connected by then.
+    // This will abort hanging connection attempts that can for whatever reason not
+    // establish a connection with the server.
     this._connectionAttemptTimeout = setTimeout(() => {
       this._client?.close();
     }, 10000);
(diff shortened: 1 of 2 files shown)


jtbandes (approved):

packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
```diff
@@ -160,6 +161,11 @@ export default class FoxgloveWebSocketPlayer implements Player {
     }
     log.info(`Opening connection to ${this._url}`);
 
+    // Set a timeout to abort the connection if we are still not connected by then
```
defunctzombie, line 164: Worth adding the _why_ here. Something along the lines of us noticing that sometimes misbehaving network setups hang the connection attempt as if still attempting but will never actually complete the attempt. So instead we kill our attempt and try the entire connection again.

defunctzombie:
@achim-k did you find that this works with your devcontainer setup? Once you start the server in the devcontainer studio is able to connect on a retry attempt?

achim-k:
> @achim-k did you find that this works with your devcontainer setup? Once you start the server in the devcontainer studio is able to connect on a retry attempt?

Yes, that works now. But only since we abort the connection attempt after 10 seconds

packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
achim-k:
```suggestion
    // Set a timeout to abort the connection if we are still not connected by then.
    // This will abort hanging connection attempts that can for whatever reason not
    // establish a server with the server.
```

Open to other suggestions

defunctzombie (approved):

PR #5680, opened 2023-03-31
Increase scrollbar width to 6px

defunctzombie:
**User-Facing Changes**

**Description**
While working on https://github.com/foxglove/studio/pull/5678, I noticed that in some instances our highlighted panel border will protrude into the active panel.

Screenshot 2023-03-31 at 4 21 06 PM

Making it harder to select the scrollbar to scroll. Changing this border behavior is challenging and has other gotchas with react mosaic. To ease the pain of using a scrollbar in an active panel I am increasing the width to 6px from 4px. I've also removed the border radius since our app has a more boxy design.

### DIFF
--- a/packages/studio-base/src/components/CssBaseline.tsx
+++ b/packages/studio-base/src/components/CssBaseline.tsx
@@ -51,7 +51,7 @@ const useStyles = makeStyles()(({ palette, typography }) => ({
     },
     div: {
       "::-webkit-scrollbar": {
-        width: "6px",
+        width: 6,
         height: 0,
       },
       "::-webkit-scrollbar-corner": {
--- a/packages/studio-base/src/components/CssBaseline.tsx
+++ b/packages/studio-base/src/components/CssBaseline.tsx
@@ -51,8 +51,8 @@ const useStyles = makeStyles()(({ palette, typography }) => ({
     },
     div: {
       "::-webkit-scrollbar": {
-        width: 4,
-        height: 4,
+        width: "6px",
+        height: 0,
       },
       "::-webkit-scrollbar-corner": {
         background: "transparent",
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/components/CssBaseline.tsx
```diff
@@ -51,8 +51,8 @@ const useStyles = makeStyles()(({ palette, typography }) => ({
     },
     div: {
       "::-webkit-scrollbar": {
-        width: 4,
-        height: 4,
+        width: "6px",
```
jtbandes, line 54:
```suggestion
        width: 6,
```

jtbandes (approved):

packages/studio-base/src/components/CssBaseline.tsx
defunctzombie: Why do we prefer the unitless form over the one with units?

jtbandes: 🤷 consistency at this point? maybe a question for @2metres

packages/studio-base/src/components/CssBaseline.tsx
```diff
@@ -51,8 +51,8 @@ const useStyles = makeStyles()(({ palette, typography }) => ({
     },
     div: {
       "::-webkit-scrollbar": {
-        width: 4,
-        height: 4,
+        width: 6,
+        height: 0,
```
2metres, line 55: why zero? Shouldn't this be 6?

defunctzombie: 0 is correct. This hides the up/down buttons. The previous height was to have rounded corners. I removed those.

jtbandes: What about horizontal scrollbars?

PR #5681, opened 2023-03-31
Update StudioLogsChannel interval to 1 second

defunctzombie:
**User-Facing Changes**
None

**Description**
Rather than using a default interval, set the re-scan interval for internal studio logs to 100ms. This is sufficient for the desired functionality.


packages/studio-base/src/providers/StudioLogsSettingsProvider/StudioLogsSettingsProvider.tsx
```diff
@@ -28,14 +28,14 @@ function StudioLogsSettingsProvider(props: PropsWithChildren<unknown>): JSX.Elem
...
-  // newly added log chnanels.
+  // newly added log channels.
   useEffect(() => {
     const storeChannelsCount = studioLogsSettingsStore.getState().channels.length;
     const intervalHandle = setInterval(() => {
       if (storeChannelsCount !== Log.channels().length) {
         setStudioLogsSettingsStore(createStudioLogsSettingsStore(savedStateRef.current));
       }
-    });
+    }, 100);
```
defunctzombie, line 38:
```suggestion
    }, 1000);
```

jtbandes (approved):
Obviously better than what it was before, but we should try to find a way to do this without a timer.

PR #5687, opened 2023-04-03
Add layout menu to app context

defunctzombie:
**User-Facing Changes**
None

**Description**
Add a topnav element from AppContext.


foxymiles:
What's the motivation for putting this in app context vs workspace context?

packages/studio-base/src/providers/WorkspaceContextProvider.tsx
```diff
@@ -37,8 +36,7 @@ function createWorkspaceContextStore(
       {
         name: "fox.workspace",
         partialize: (value) => {
-          const { layoutMenuOpen: _, ...rest } = value;
-          return rest;
+          return value;
```
defunctzombie, line 39: @foxymiles do we still need this function if all it does is pass the value through?

foxymiles: No although it might be good to keep it there to remind people this is something they need to consider if they add stuff to this context.

foxymiles (approved):

packages/studio-base/src/context/WorkspaceContext.ts
```diff
@@ -118,8 +116,7 @@ export function useWorkspaceActions(): WorkspaceActions {
           ? set({ leftSidebarItem: "panel-settings", leftSidebarOpen: true })
           : set({ sidebarItem: "panel-settings" }),
 
-      openLayoutBrowser: () =>
-        enableNewTopNav ? set({ layoutMenuOpen: true }) : set({ sidebarItem: "layouts" }),
+      openLayoutBrowser: () => set({ sidebarItem: "layouts" }),
```
jtbandes, line 119: what are we going to do with this method when we remove the feature flag? just get rid of it?

jtbandes (approved):

packages/studio-base/src/context/WorkspaceContext.ts
defunctzombie: Yea - I think so. still unclear what behavior we'll want from empty layout state so we might need to revisit this but for now I removed it.

PR #5689, opened 2023-04-03
Use readahead message buffer in plot panel

foxymiles:
**User-Facing Changes**
Improve behavior of plot panel in memory constrained conditions

**Description**
This is a proof of concept PR that passes through the readahead buffer in `BufferedIterableSource` to the player and the plot panel.

Pros:
1. We already have the messages in the readahead buffer so this doesn't change our actual data loading pattern.
2. It doesn't allocate any more memory than the existing implementation.
3. The messages in the readahead are in a separate field in the player state so other panels should be able to ignore this safely and it shouldn't break any other assumptions about message ordering.

Cons:
1. The readahead buffer is arguably an implementation detail we don't want to commit to in the player state API.
2. The Y scale on the plot panel can change quite markedly while seeking around in the data.

***Before***:
https://user-images.githubusercontent.com/93935560/229448259-011898c4-5a1e-4163-a785-326d89ed5058.mov

***After***:
https://user-images.githubusercontent.com/93935560/229448329-867e5d2f-dc5f-4d3b-b34a-fb99a21b9755.mov

### DIFF
--- a/packages/studio-base/src/panels/Plot/index.tsx
+++ b/packages/studio-base/src/panels/Plot/index.tsx
@@ -13,17 +13,18 @@
 
 import DownloadIcon from "@mui/icons-material/Download";
 import { useTheme } from "@mui/material";
-import { compact, isNumber, uniq } from "lodash";
+import { compact, isNumber, last, transform, uniq } from "lodash";
 import memoizeWeak from "memoize-weak";
 import { ComponentProps, useCallback, useEffect, useMemo, useState } from "react";
 
 import { filterMap } from "@foxglove/den/collection";
 import { useShallowMemo } from "@foxglove/hooks";
 import {
+  Time,
   add as addTimes,
   fromSec,
+  isGreaterThan,
   subtract as subtractTimes,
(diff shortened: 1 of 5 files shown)


packages/studio-base/src/panels/Plot/index.tsx
```diff
@@ -205,6 +206,11 @@ function selectCurrentTime(ctx: MessagePipelineContext) {
   return ctx.playerState.activeData?.currentTime;
 }
 
+const EMPTY_READAHEAD_MESSAGES: MessageEvent<unknown>[] = [];
+function selectReadAheadMessages(ctx: MessagePipelineContext) {
+  return ctx.playerState.activeData?.readAheadMessages ?? EMPTY_READAHEAD_MESSAGES;
```
snosenzo, line 211: Given that the player is setting a new read ahead every frame because its immutable, this will update the plot every frame which can cause a significant performance impact. Would there be a way to separate this such that only when the time range of the readahead changes it updates the plot?

PR #5694, opened 2023-04-04
Use both blocks and current frame data in plots

foxymiles:
**User-Facing Changes**
Improve behavior of the plot panel in low memory conditions.

**Description**
Combine block and current frame / playback messages in plots to give better user experience in low memory conditions.

This is an alternative to https://github.com/foxglove/studio/pull/5689 and a proposed fix for https://github.com/foxglove/studio/issues/5500.

The idea here is to allow data from blocks/allFrames and accumulated `currentFrame` to be merged and plotted together, so in cases where the block cache is full we can still plot something. This requires playback to accumulate `currentFrame` data.

Pros:
* If the user is willing to wait for playback we can in theory plot more data than in #5689.

Cons:
* Playback is required to see the currentFrame data. I'm not sure how we'd create an affordance for that in the UI.

### DIFF
--- a/packages/studio-base/src/components/MessagePathSyntax/parseRosPath.ts
+++ b/packages/studio-base/src/components/MessagePathSyntax/parseRosPath.ts
@@ -50,7 +50,7 @@ const parseRosPath = memoize((path: string): RosPath | undefined => {
 });
 export default parseRosPath;
 
-export function getTopicsFromPaths(paths: string[]): string[] {
+export function getTopicsFromPaths(paths: readonly string[]): string[] {
   return uniq(
     filterMap(paths, (path) => {
       const rosPath = parseRosPath(path);
--- a/packages/studio-base/src/panels/Plot/datasets.tsx
+++ b/packages/studio-base/src/panels/Plot/datasets.tsx
@@ -2,6 +2,8 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
+import { Immutable } from "immer";
+
 import { filterMap } from "@foxglove/den/collection";
 import { isTime, subtract, Time, toSec } from "@foxglove/rostime";
 import { format } from "@foxglove/studio-base/util/formatTime";
@@ -27,8 +29,8 @@ function getXForPoint(
   xAxisVal: PlotXAxisVal,
   timestamp: number,
   innerIdx: number,
-  xAxisRanges: readonly (readonly PlotDataItem[])[] | undefined,
-  xItem: PlotDataItem | undefined,
+  xAxisRanges: Immutable<PlotDataItem[][]> | undefined,
+  xItem: undefined | Immutable<PlotDataItem>,
   xAxisPath: BasePlotPath | undefined,
 ): number | bigint {
   if (isCustomScale(xAxisVal) && xAxisPath) {
@@ -47,13 +49,13 @@ function getXForPoint(
 }
 
 function getDatumsForMessagePathItem(
-  yItem: PlotDataItem,
-  xItem: PlotDataItem | undefined,
+  yItem: Immutable<PlotDataItem>,
+  xItem: undefined | Immutable<PlotDataItem>,
   startTime: Time,
   timestampMethod: TimestampMethod,
   xAxisVal: PlotXAxisVal,
   xAxisPath?: BasePlotPath,
-  xAxisRanges?: readonly (readonly PlotDataItem[])[],
+  xAxisRanges?: Immutable<PlotDataItem[][]>,
 ): { data: Datum[]; hasMismatchedData: boolean } {
   const timestamp = timestampMethod === "headerStamp" ? yItem.headerStamp : yItem.receiveTime;
   if (!timestamp) {
@@ -116,11 +118,11 @@ function getDatasetsFromMessagePlotPath({
   invertedTheme = false,
 }: {
   path: PlotPath;
-  yAxisRanges: readonly (readonly PlotDataItem[])[];
+  yAxisRanges: Immutable<PlotDataItem[][]>;
   index: number;
   startTime: Time;
   xAxisVal: PlotXAxisVal;
-  xAxisRanges: readonly (readonly PlotDataItem[])[] | undefined;
+  xAxisRanges: Immutable<PlotDataItem[][]> | undefined;
   xAxisPath?: BasePlotPath;
   invertedTheme?: boolean;
 }): {
@@ -147,10 +149,10 @@ function getDatasetsFromMessagePlotPath({
   }
 
   for (const [rangeIdx, range] of yAxisRanges.entries()) {
-    const xRange: readonly PlotDataItem[] | undefined = xAxisRanges?.[rangeIdx];
+    const xRange = xAxisRanges?.[rangeIdx];
     let rangeData: Datum[] = [];
     for (const [outerIdx, item] of range.entries()) {
-      const xItem: PlotDataItem | undefined = xRange?.[outerIdx];
+      const xItem = xRange?.[outerIdx];
       const { data: datums, hasMismatchedData: itemHasMistmatchedData } =
         getDatumsForMessagePathItem(
           item,
@@ -234,7 +236,7 @@ function getDatasetsFromMessagePlotPath({
 
 type Args = {
   paths: PlotPath[];
-  itemsByPath: PlotDataByPath;
+  itemsByPath: Immutable<PlotDataByPath>;
   startTime: Time;
   xAxisVal: PlotXAxisVal;
   xAxisPath?: BasePlotPath;
--- a/packages/studio-base/src/panels/Plot/internalTypes.ts
+++ b/packages/studio-base/src/panels/Plot/internalTypes.ts
@@ -53,9 +53,7 @@ export type PlotDataItem = {
   headerStamp?: Time;
 };
 
-export type PlotDataByPath = {
-  [path: string]: PlotDataItem[][];
-};
+export type PlotDataByPath = Record<string, PlotDataItem[][]>;
 
 // A "reference line" plot path is a numeric value. It creates a horizontal line on the plot at the specified value.
 export function isReferenceLinePlotPathType(path: BasePlotPath): boolean {
--- a/packages/studio-base/src/panels/Plot/plotData.ts
+++ b/packages/studio-base/src/panels/Plot/plotData.ts
@@ -168,6 +168,8 @@ function mergeByPath(a: Im<PlotDataByPath>, b: Im<PlotDataByPath>): Im<PlotDataB
   );
 }
 
+// Sort by start time, then end time, so that folding from the left gives us the
+// right consolidated interval.
 function compare(a: Im<PlotDataByPath>, b: Im<PlotDataByPath>): number {
   const rangeA = findTimeRanges(a).all;
   const rangeB = findTimeRanges(b).all;
--- a/packages/studio-base/src/panels/Plot/plotData.ts
+++ b/packages/studio-base/src/panels/Plot/plotData.ts
@@ -30,26 +30,27 @@ type TimeRange = { start: Time; end: Time };
  * Find the earliest and latest times of messages in data, for all messages and
  * per-path.
  */
-export const findTimeRanges = memoizeWeak(
-  (data: Im<PlotDataByPath>): { all: TimeRange; byPath: Record<string, TimeRange> } => {
-    const byPath: Record<string, TimeRange> = {};
-    let start: Time = MAX_TIME;
-    let end: Time = MIN_TIME;
-    for (const path of Object.keys(data)) {
-      const thisPath = (byPath[path] = { start: MAX_TIME, end: MIN_TIME });
-      for (const item of data[path] ?? []) {
-        for (const datum of item) {
-          start = minTime(start, datum.receiveTime);
-          end = maxTime(end, datum.receiveTime);
-          thisPath.start = minTime(thisPath.start, datum.receiveTime);
-          thisPath.end = maxTime(thisPath.end, datum.receiveTime);
-        }
+export function findTimeRanges(data: Im<PlotDataByPath>): {
+  all: TimeRange;
+  byPath: Record<string, TimeRange>;
+} {
+  const byPath: Record<string, TimeRange> = {};
+  let start: Time = MAX_TIME;
+  let end: Time = MIN_TIME;
+  for (const path of Object.keys(data)) {
+    const thisPath = (byPath[path] = { start: MAX_TIME, end: MIN_TIME });
+    for (const item of data[path] ?? []) {
+      for (const datum of item) {
+        start = minTime(start, datum.receiveTime);
+        end = maxTime(end, datum.receiveTime);
+        thisPath.start = minTime(thisPath.start, datum.receiveTime);
+        thisPath.end = maxTime(thisPath.end, datum.receiveTime);
       }
     }
+  }
 
-    return { all: { start, end }, byPath };
-  },
-);
+  return { all: { start, end }, byPath };
+}
 
 /**
  * Fetch the data we need from each item in itemsByPath and discard the rest of
@@ -168,11 +169,13 @@ function mergeByPath(a: Im<PlotDataByPath>, b: Im<PlotDataByPath>): Im<PlotDataB
   );
 }
 
+const memoFindTimeRanges = memoizeWeak(findTimeRanges);
+
 // Sort by start time, then end time, so that folding from the left gives us the
 // right consolidated interval.
 function compare(a: Im<PlotDataByPath>, b: Im<PlotDataByPath>): number {
-  const rangeA = findTimeRanges(a).all;
-  const rangeB = findTimeRanges(b).all;
+  const rangeA = memoFindTimeRanges(a).all;
+  const rangeB = memoFindTimeRanges(b).all;
   const startCompare = compareTimes(rangeA.start, rangeB.start);
   return startCompare !== 0 ? startCompare : compareTimes(rangeA.end, rangeB.end);
 }
--- a/packages/studio-base/src/panels/Plot/plotData.ts
+++ b/packages/studio-base/src/panels/Plot/plotData.ts
@@ -1,0 +1,194 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { Immutable as Im } from "immer";
+import { assignWith, last, isEmpty } from "lodash";
+import memoizeWeak from "memoize-weak";
+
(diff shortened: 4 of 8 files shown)


defunctzombie:
I tried this PR and https://github.com/foxglove/studio/pull/5689.

5694: With this PR I can play data and see it accumulate on the plot panel as I play; however, once I seek back even a little, all my data is gone and I have to play again to see it. Overall I found the experience nice because the data I had played remained - until I seek at least.

5689: With https://github.com/foxglove/studio/pull/5689 a see some data ahead of my seek location, however once I click play the data prior to my playhead is constantly removed as the player moves forward. This is unpleasant since I lose data from the display after I play past it. This is countered by being able to seek and see the data ahead of the cursor.

Thoughts:
- During initial load and seeking to some new location, the behavior in 5698 which shows me a bunch of data ahead of the seek location is definitely nicer than having to play the data like in 5694.
- Having the data behind my cursor stick around as I play like in 5694 is nicer than what is currently happening in 5698.
- In the cases where "blocks" will have all of the data, we can avoid having to incorporate `currentFrame` or `readAhead` at all to avoid recompute costs.
- My high level understanding of how we might tackle merging the "blocks" data with external data in the plot panel leads me to think that for either 5694 and 5698 we'd need to keep track of what we have and have not seen and de-duplicate it between the various data streams to obtain acceptable performance?

Since the problem reported in #5500 only happens when you are plotting data from large messages and my last thought above, I'm currently inclined to go with 5694 since it involves no changes to player logic but will let us evolve the plot logic to properly incorporate a second, ever changing, set of messages. Once we have that in place as the first fix I do think there's an interesting world where we can expose more of what is already buffered to panels.

foxymiles:
@defunctzombie Thanks for the review. Yeah for this PR to be mergeable I need to add more logic to keep previously buffered messages and to handle potential overlaps between blocks data and different segments of messages accumulated through playback.

foxymiles:
I've added logic to retain and consolidate messages across multiple seeks and to handle overlaps between block data and currentFrame data.

snosenzo (commented):
a lot of questions mostly because I'm not as familiar with plot stuff as I could be, but overall looks good

packages/studio-base/src/panels/Plot/usePlotPanelMessageData.ts
```diff
@@ -0,0 +1,194 @@
...
+
+  const cachedGetMessagePathDataItems = useCachedGetMessagePathDataItems(allPaths);
+
+  const blocksTimeRange = PlotData.findTimeRanges(plotDataForBlocks);
+
+  // When restoring, keep only the paths that are present in allPaths. Without
+  // this, the reducer value will grow unbounded with new paths as users
+  // add/remove series. We use a timestamp tag so we can accumulate messages
+  // across multiple restore calls.
+  const restore = useCallback(
```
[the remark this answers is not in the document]
foxymiles: This is an argument to `useMessageReducer`, which calls the param `restore`.

packages/studio-base/src/panels/Plot/plotData.ts
```diff
@@ -0,0 +1,194 @@
...
+);
+
+/**
+ * Fetch the data we need from each item in itemsByPath and discard the rest of
+ * the message to save memory.
+ */
+const getByPath = (itemsByPath: MessageDataItemsByPath): PlotDataByPath => {
+  const ret: PlotDataByPath = {};
+  Object.entries(itemsByPath).forEach(([path, items]) => {
+    ret[path] = [
```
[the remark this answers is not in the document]
foxymiles: All these are 2d arrays as they are on current main as well. The existing comment which I just moved to this file:

```
            // PlotDataPaths have 2d arrays of items to accommodate blocks which
            // may have gaps so each continuous set of blocks forms one
            // continuous line. For streaming messages we treat this as one
            // continuous set of items and always add to the first "range"
```

packages/studio-base/src/panels/Plot/internalTypes.ts
```diff
@@ -53,9 +53,7 @@ export type PlotDataItem = {
   headerStamp?: Time;
 };
 
-export type PlotDataByPath = {
-  [path: string]: PlotDataItem[][];
-};
+export type PlotDataByPath = Record<string, PlotDataItem[][]>;
```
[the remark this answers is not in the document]
foxymiles: See comment above.

packages/studio-base/src/panels/Plot/usePlotPanelMessageData.ts
```diff
@@ -0,0 +1,194 @@
...
+      }
+
+      return { ...previous, data: pick(previous.data, allPaths) };
+    },
+    [allPaths],
+  );
+
+  const addMessages = useCallback(
+    (accumulated: TaggedPlotDataByPath, msgEvents: readonly MessageEvent<unknown>[]) => {
+      const lastEventTime = msgEvents[msgEvents.length - 1]?.receiveTime;
```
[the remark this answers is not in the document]
foxymiles: Existing code so I didn't want to import `last` just to change it.

packages/studio-base/src/panels/Plot/usePlotPanelMessageData.ts
```diff
@@ -0,0 +1,194 @@
...
+      let newAccumulated: TaggedPlotDataByPath | undefined;
+
+      for (const msgEvent of msgEvents) {
+        const paths = topicToPaths[msgEvent.topic];
+        if (!paths) {
+          continue;
+        }
+
+        for (const path of paths) {
+          const dataItem = cachedGetMessagePathDataItems(path, msgEvent);
```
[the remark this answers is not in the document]
foxymiles: The names of most of these variables come from existing code so I haven't changed them where it didn't seem necessary.

packages/studio-base/src/panels/Plot/usePlotPanelMessageData.ts
```diff
@@ -0,0 +1,194 @@
...
+
+  // Accumulate separate message playback sequences into distinct intervals we
+  // can later combine into a single combined set of message data.
+  const [accumulatedPathIntervals, setAccumulatedPathIntervals] = useState<
+    Record<string, PlotDataByPath>
+  >({});
+
+  useEffect(() => {
+    if (!isEmpty(plotDataByPath.data)) {
+      setAccumulatedPathIntervals((oldValue) => ({
```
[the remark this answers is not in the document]
foxymiles: Any suggestions? The accumulation happens inside the reducer so that intermediate state is not available to the component using it.

packages/studio-base/src/panels/Plot/usePlotPanelMessageData.ts
```diff
@@ -0,0 +1,194 @@
...
+import * as PlotData from "@foxglove/studio-base/panels/Plot/plotData";
+import { MessageEvent } from "@foxglove/studio-base/players/types";
+import { getTimestampForMessage } from "@foxglove/studio-base/util/time";
+
+type TaggedPlotDataByPath = { tag: string; data: PlotDataByPath };
+
+type Params = Immutable<{
+  allPaths: string[];
+  followingView: undefined | ChartDefaultView;
+  showSingleCurrentMessage: boolean;
```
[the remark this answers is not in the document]
foxymiles: This is just reorganizing the existing code since the component itself was getting to be quite complicated. Basically `showSingleCurrentMessage` is true for a following view (with a set width).

packages/studio-base/src/panels/Plot/usePlotPanelMessageData.ts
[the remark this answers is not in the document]
snosenzo: hmm maybe a `useRef` for the old value and a `useMemo` to return the new `accumulatedPathIntervals` value based off of `plotDataByPath`, then just assigning the ref to the new value after the memo? not sure if that feels sloppier, but that way it wouldn't have to go through an extra render loop at least

snosenzo (approved):
Thanks for answering my feedback. I think it might be good to update some of the old variable names for consistency and clarity. also if you want to try removing the extra render loop that could be good, but I think the PR looks good overall

packages/studio-base/src/components/MessagePathSyntax/parseRosPath.ts
```diff
@@ -50,7 +50,7 @@ const parseRosPath = memoize((path: string): RosPath | undefined => {
 });
 export default parseRosPath;
 
-export function getTopicsFromPaths(paths: string[]): string[] {
+export function getTopicsFromPaths(paths: readonly string[]): string[] {
```
defunctzombie, line 53: everything should be readonly by default. Rust got this right :(

packages/studio-base/src/panels/Plot/datasets.tsx
```diff
@@ -2,6 +2,8 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
+import { Immutable } from "immer";
```
defunctzombie, line 5: Is this a type annotation only or does it actually do something to the object? If it does something to the object what is our expected overhead on what it does? I ask because if it wraps it in a Proxy or something I think that's not really ok for us to use for the code I am seeing it used for (given my exposure to what proxies do to runtime perf)

defunctzombie: If this is a type only thing lets use `import type { ... }`. If we also find this intuitively useful to have this type we can consider making it globally available as if it was a built-in typescript utility type.

foxymiles: It's just an annotation. Equivalent to `DeepReadonly`.

+1 that this is an important concept that we should have available in our own codebase without an external dependency.

defunctzombie, line 5:
```suggestion
import type { Immutable } from "immer";
```

packages/studio-base/src/panels/Plot/usePlotPanelMessageData.ts
```diff
@@ -0,0 +1,194 @@
...
+
+/**
+ * Combines preloaded block data and messages accumulated during playback into a
+ * single PlotDataByPath object.
+ */
+export function usePlotPanelMessageData(params: Params): Immutable<PlotDataByPath> {
+  const { allPaths, followingView, showSingleCurrentMessage } = params;
+
+  // When iterating message events, we need a reverse lookup from topic to the
+  // paths that requested the topic.
```
defunctzombie, line 40: Why? (might be worth adding to the comment). IIRC it is because we need to add the event to the correct series. Path is synonymous of "series" here I think.

packages/studio-base/src/panels/Plot/usePlotPanelMessageData.ts
```diff
@@ -0,0 +1,194 @@
...
+  const plotDataForBlocks = useMemo(() => {
+    if (showSingleCurrentMessage) {
+      return {};
+    }
+    return PlotData.getBlockItemsByPath(decodeMessagePathsForMessagesByTopic, blocks);
+  }, [blocks, decodeMessagePathsForMessagesByTopic, showSingleCurrentMessage]);
+
+  const cachedGetMessagePathDataItems = useCachedGetMessagePathDataItems(allPaths);
+
+  const blocksTimeRange = PlotData.findTimeRanges(plotDataForBlocks);
```
defunctzombie, line 65: This made me do a double take. Its a computation inside of render. I do see we use `memoizeWeak` but that is not obvious when seeing this code. I would move this computation into the `useMemo` that returns `plotDataForBlocks` and have it instead return `{ plotDataForBlocks, blocksTimeRange }` So you can re-use react doing the memoize. I would then remove memoizeWeak from `findTimeRanges`.

packages/studio-base/src/panels/Plot/plotData.ts
```diff
@@ -0,0 +1,196 @@
...
+          receiveTime: messageAndData.messageEvent.receiveTime,
+          headerStamp,
+        };
+      }),
+    ];
+  });
+  return ret;
+};
+
+const getMessagePathItemsForBlock = memoizeWeak(
```
defunctzombie, line 75: Generally not feeling the vibe on `memoizeWeak`, feels easier to understand when I can see the memoization at the react/render level? Maybe this was already memoized in which case ignore.

packages/studio-base/src/panels/Plot/plotData.ts
```diff
@@ -0,0 +1,196 @@
...
+  (
+    decodeMessagePathsForMessagesByTopic: (_: MessageBlock) => MessageDataItemsByPath,
+    block: MessageBlock,
+  ): PlotDataByPath => {
+    return Object.freeze(getByPath(decodeMessagePathsForMessagesByTopic(block)));
+  },
+);
+
+/**
+ * Fetch all the plot data we want for our current subscribed topics from blocks.
```
defunctzombie, line 85: The comment is a bit misleading since there's nothing about this function related to "current subscribed". It has no concept of "subscribe" beyond being passed a list of `decodeMessagePathsForMessageByTopic`.

packages/studio-base/src/panels/Plot/plotData.ts
```diff
@@ -0,0 +1,196 @@
...
+      const newValues = filterMap(srcValue ?? [], (item) => {
+        const laterDatums = item.filter((datum) => isGreaterThan(datum.receiveTime, lastTime));
+        return laterDatums.length > 0 ? laterDatums : undefined;
+      });
+      return newValues.length > 0 ? objValue.concat(newValues) : objValue;
+    },
+  );
+}
+
+// Sort by start time, then end time, so that folding from the left gives us the
```
defunctzombie, line 171: This comment is a bit lackluster. It tells me maybe some implementation detail but I felt myself still wondering what it actually compares and returns.

packages/studio-base/src/panels/Plot/usePlotPanelMessageData.ts
```diff
@@ -0,0 +1,194 @@
...
+          const existingBlockRange = blocksTimeRange.byPath[path];
+          if (
+            existingBlockRange &&
+            isTimeInRangeInclusive(
+              msgEvent.receiveTime,
+              existingBlockRange.start,
+              existingBlockRange.end,
+            )
+          ) {
+            // Skip messages that fall within the range of our block data since
```
defunctzombie, line 113: Good comment - I'd move it above the _if_ statement. I found myself trying to grok the if statement's purpose only to land on this comment which told me exactly what I wanted to know.

packages/studio-base/src/panels/Plot/usePlotPanelMessageData.ts
```diff
@@ -0,0 +1,194 @@
...
+
+            newAccumulated.data[path] = plotDataPath;
+          }
+        }
+      }
+
+      return newAccumulated ?? accumulated;
+    },
+    [
+      blocksTimeRange,
```
defunctzombie, line 158: Note that in today's logic this `addMessages` callback is relatively stable: https://github.com/foxglove/studio/blob/main/packages/studio-base/src/panels/Plot/index.tsx#L442 whereas with this new change you are thrashing it whenever the blocks time range changes. This does tend to happen only on startup but it is a different behavior. Its another thing that's pushing me towards thinking that we are so close to a store-like solution here.

Maybe this dependency changing during load is not a problem in practice? Even for large data sets?

packages/studio-base/src/panels/Plot/datasets.tsx
foxymiles: Are we planning on doing this in general in our codebase? We import types directly without `import type` all over.

packages/studio-base/src/panels/Plot/usePlotPanelMessageData.ts
foxymiles: This is an old existing comment I just moved over as part of the refactor to this hook. I've mostly left that code alone except for moving it but we could use this as an opportunity to polish it up a bit too if you want.

packages/studio-base/src/panels/Plot/usePlotPanelMessageData.ts
foxymiles: We need `memoizeWeak` for `findTimeRanges` because it's also used to avoid recalculating time ranges over and over again in our sort function later. Using it here is basically free since this function will already have memoized the return value for the blocks during a previous `combine`.

packages/studio-base/src/panels/Plot/plotData.ts
foxymiles: Please see above comment. Block level memoization is a better fit here since `useMemo` is too coarse.

packages/studio-base/src/panels/Plot/usePlotPanelMessageData.ts
foxymiles: Yeah it will change every time new blocks are loaded, which is ~400 times at most (currently). We could refer to that by ref instead but this function really does need to have the current value of this range to function correctly so this seemed safer.

`useMessageReducer` actually already is using zustand under the hood. We could simplify the logic in this PR a fair bit if we're willing to adjust the semantics of `addMessages` in `useMessageReducer` to make it possible for the reducer function to also access messages across seeks. A lot of the complexity here is building up messages across seeks when `restore` in `ueMessageReducer` returns undefined.

packages/studio-base/src/panels/Plot/datasets.tsx
defunctzombie: Could be nice but without a lint rule hard to enforce. For this one place figured why not while I was in here doing the review and learned it is only a type.

packages/studio-base/src/panels/Plot/usePlotPanelMessageData.ts
defunctzombie: Gotcha - fine to do later since this is copy-paste currently.

packages/studio-base/src/panels/Plot/usePlotPanelMessageData.ts
defunctzombie: I find it nicer to drop the memoizeWeak and move it to the useMemo so I don't need to guess around the call of this in render as a reader.

packages/studio-base/src/panels/Plot/plotData.ts
defunctzombie: @foxymiles Any thoughts on what to improve with this comment?

packages/studio-base/src/panels/Plot/plotData.ts
```diff
@@ -0,0 +1,199 @@
...
+      const newValues = filterMap(srcValue ?? [], (item) => {
+        const laterDatums = item.filter((datum) => isGreaterThan(datum.receiveTime, lastTime));
+        return laterDatums.length > 0 ? laterDatums : undefined;
+      });
+      return newValues.length > 0 ? objValue.concat(newValues) : objValue;
+    },
+  );
+}
+
+const memoFindTimeRanges = memoizeWeak(findTimeRanges);
```
defunctzombie, line 172: Why do we need memoizeWeak here?

PR #5699, opened 2023-04-04
Fix ListItem styles when rendered in a MenuList

2metres:
**Description**
Adds 8px margin above and below Dividers when inside MUI Menu list.


defunctzombie (approved):
Lets have a story for this so we know when this changes.

2metres:
What are we testing for? MUI Component styling where elements are not inside the ThemeProvider?

defunctzombie:
> What are we testing for? MUI Component styling where elements are not inside the ThemeProvider?

These components _are_ inside a theme provider. Here you would be testing that a divider has the right spacing around ListItems in a menu. Literally the thing we needed this change for.

PR #5700, opened 2023-04-04
Remove global typography styles

2metres:
**User-Facing Changes**
None, hopefully

**Description**
In the olden times we had a lot of conflicting global styles. In the process of isolating MUI menu's with @defunctzombie we noticed that Menus, Dialogs and Modals were getting Typography styles from GlobalStyles being set on the body. This change removes that inheritence and makes the components less coupled to the App

### DIFF
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -8,7 +8,6 @@ import { useTranslation } from "react-i18next";
 
 import { Condvar } from "@foxglove/den/async";
 import CssBaseline from "@foxglove/studio-base/components/CssBaseline";
-// import GlobalCss from "@foxglove/studio-base/components/GlobalCss";
 import GlobalCss from "@foxglove/studio-base/components/GlobalCss";
 import MultiProvider from "@foxglove/studio-base/components/MultiProvider";
 import StudioToastProvider from "@foxglove/studio-base/components/StudioToastProvider";
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -8,7 +8,6 @@ import { useTranslation } from "react-i18next";
 
 import { Condvar } from "@foxglove/den/async";
 import CssBaseline from "@foxglove/studio-base/components/CssBaseline";
-import GlobalCss from "@foxglove/studio-base/components/GlobalCss";
 import MultiProvider from "@foxglove/studio-base/components/MultiProvider";
 import StudioToastProvider from "@foxglove/studio-base/components/StudioToastProvider";
 import AppConfigurationContext from "@foxglove/studio-base/context/AppConfigurationContext";
@@ -106,7 +105,6 @@ function StudioContextProviders({
             value={needsCombinedReadySignal ? readySignal1 : readySignal}
           >
             <ThemeProvider isDark={false}>
-              <GlobalCss />
               <CssBaseline>
                 <MultiProvider providers={providers}>{children}</MultiProvider>
               </CssBaseline>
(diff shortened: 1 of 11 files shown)


.storybook/preview.tsx
```diff
@@ -116,6 +106,7 @@ function StudioContextProviders({
             value={needsCombinedReadySignal ? readySignal1 : readySignal}
           >
             <ThemeProvider isDark={false}>
+              <GlobalCss />
```
defunctzombie, line 109: I don't think this should be here? It will set styles on the body will it not?

packages/studio-base/src/components/GlobalCss.tsx
```diff
@@ -25,13 +23,7 @@ export default function GlobalCss(): JSX.Element {
           boxSizing: "inherit",
         },
         body: {
-          background: theme.palette.background.default,
```
defunctzombie, line 28: I think we should keep the background and color and fonts as fallbacks. Background especially because it shows up before other things have loaded iirc.

defunctzombie (approved):
I would leave some of the body fallback items and remove <GlobalCss> from storybook.

.storybook/preview.tsx
jtbandes: The idea was to apply the same styles in storybook as we do in our app. Although now I do wonder how much change we would see if we removed this. I do think if we keep it, we should only keep 1 copy (like there was before).

packages/studio-base/src/components/GlobalCss.tsx
2metres: This comment confusing. Keeping the background, color, and fonts is the entireity of this change.

2metres: We can resolve the flash before load by adding some styles to https://github.com/foxglove/studio/blob/7a925b665da79e2533cbafad2ea09bbb7f4e47cf/packages/studio-web/src/webpackConfigs.ts#L101-L127

defunctzombie:
We can use storybook backgrounds to manage the background color: https://github.com/foxglove/studio/pull/5723/commits/0828aaae4a6868a6204dca695c5d179895c612d5

PR for viewing storybook: https://github.com/foxglove/studio/pull/5723

I think we should keep the component specific changes, keep GlobalCss unchanged, and use the backgrounds feature to get the color we want for screenshot testing.

jtbandes:
I don't really understand your recommendation – GlobalCss is currently setting the background color, so if we keep it, why would we need the storybook backgrounds feature?

defunctzombie:
> background color, so if we keep it, why would we need the storybook backgrounds feature?

It also sets the font. We want to avoid setting things on _body_ like the font so we can better identify which components are not styled properly when the font is not set.

jtbandes:
So to be clear you are proposing not changing GlobalCss, but removing it from the storybook?

defunctzombie:
> So to be clear you are proposing not changing GlobalCss, but removing it from the storybook?

image

My clarity needs work :D. Yes I am proposing we remove GlobalCss from storybook.

.storybook/styles.css
```diff
@@ -7,3 +7,26 @@
   animation: none !important;
   caret-color: transparent !important;
 }
+
+*,
+*:before,
+*:after {
+  box-sizing: inherit;
+}
```
2metres, line 15: @defunctzombie this one is a bit concerning. Without it 64% of our diffs failed

2metres: Its a pretty common global style so not surprising I guess

packages/studio-base/src/components/GlobalCss.tsx
```diff
@@ -2,15 +2,13 @@
...
+import { GlobalStyles } from "@mui/material";
 
 /** GlobalCss component configures html, body, and #root with theme elements */
 export default function GlobalCss(): JSX.Element {
-  const theme = useTheme();
-
   return (
     <GlobalStyles
-      styles={{
+      styles={(theme) => ({
```
defunctzombie, line 11: Why is this better than `useTheme`?

defunctzombie (approved):

PR #5701, opened 2023-04-04
Display `key` `value` shorthand in Raw message panel

defunctzombie:
**User-Facing Changes**
Add quick-view support to Raw Message panel for objects with `key` and `value` fields.

If the field is an object with a `key` field, display the key next to the field name. If a field is an object containing `key` and `value` fields, display those values next to the field name.

**Description**

| Before | After |
|--------|-------|
| image | image | 

Fixes: FG-2758


defunctzombie:
I am not sure this is the right approach. Instead of this being a typical filter value (which maybe we still want to do), it might be more useful to add logic to the raw message panel for `key/value` objects so we can show the `<key>:<value>` next to the object rather than `key: <key>`

packages/studio-base/src/components/MessagePathSyntax/isTypicalFilterName.test.ts
```diff
@@ -0,0 +1,18 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { isTypicalFilterName } from "./isTypicalFilterName";
+
+describe("isTypicalFilterName", () => {
+  it.each(["id", "trackID", "_id", "track_id", "ID", "Id", "key"])(
```
jtbandes, line 8: `trackId`?

packages/studio-base/src/components/MessagePathSyntax/isTypicalFilterName.ts
```diff
@@ -11,8 +11,18 @@
 //   found at http://www.apache.org/licenses/LICENSE-2.0
 //   You may not use this file except in compliance with the License.
 
-export const TypicalFilterNames = ["id", "_id", "ID", "Id"];
+/**
+ * List of typical filter names. Typical filter names are common names that are used for filters
+ * in message path queries. For example, "id" or "key".
+ */
+export const TypicalFilterNames = ["id", "_id", "ID", "Id", "key"];
```
jtbandes, line 18: Can we get rid of this array? Not sure why we have both an array and a regex?

jtbandes (approved):
Seems fine. I suppose we could try adding `value` as well and see if that feels helpful.

defunctzombie:
I'm kinda meh on this cause it only shows the _key_ and not the value. Maybe that's of some use? @amacneil can you follow up with whomever asked for this to see if this is at all helpful?

amacneil:
👍 adding `key` is very useful based on the conversation I had. let's start with that.

jhurliman:
https://github.com/orgs/foxglove/discussions/695

amacneil:
Any blockers to get this over the finish line?

defunctzombie:
> Any blockers to get this over the finish line?

Nope - I've not had time to come back to it. I've wanted to see if it would be easy to add `key : value` as per the linked discussion.

packages/studio-base/src/components/MessagePathSyntax/isTypicalFilterName.ts
defunctzombie: https://github.com/foxglove/studio/blob/main/packages/studio-base/src/components/MessagePathSyntax/useCachedGetMessagePathDataItems.ts#L315

defunctzombie:
image

packages/studio-base/src/components/MessagePathSyntax/isTypicalFilterName.test.ts
defunctzombie: Added

packages/studio-base/src/components/MessagePathSyntax/isTypicalFilterName.ts
defunctzombie: Handling this in a separate PR: https://github.com/foxglove/studio/pull/5821

jtbandes (approved):

PR #5702, opened 2023-04-04
Update User Scripts panel templates to use consistent naming for function

esthersweon:
**User-Facing Changes**

- N/A

**Description**
- Updated User Scripts panel templates to use consistent naming for function

### DIFF
--- a/packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/typescript/templates/locationFix.ts.template
+++ b/packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/typescript/templates/locationFix.ts.template
@@ -13,7 +13,7 @@ export const output = "/studio_script/my_gps";
 // Our node will output a LocationFix message
 type LocationFix = Message<"foxglove.LocationFix">;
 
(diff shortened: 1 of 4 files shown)


jtbandes (approved):

packages/studio-base/src/components/CreateEventDialog.tsx
```diff
@@ -194,13 +194,13 @@ export function CreateEventDialog(props: { onClose: () => void }): JSX.Element {
 
   return (
     <Dialog open onClose={onClose} fullWidth maxWidth="sm">
-      <Stack paddingX={3} paddingTop={2}>
-        <Typography variant="h2">Create event</Typography>
+      <Stack paddingX={3} paddingTop={2} paddingBottom={32}>
```
2metres, line 197: The padding bottom uses padding increments `32` seems like a lot of padding are you trying to get 32px of padding here?

esthersweon: Weird this file should've never been updated. Wonder if it was an old `git stash`. Removing.

2metres (approved):

PR #5704, opened 2023-04-05
Tint dark theme's greys to match paper color

2metres:
**User-Facing Changes**
None

**Description**
Shift grey palette from neutral grey to slightly more blue. Makes tooltip background more harmonious with backgrounds

### DIFF
--- a/packages/studio-base/src/components/PanelList/index.tsx
+++ b/packages/studio-base/src/components/PanelList/index.tsx
@@ -91,11 +91,14 @@ const useStyles = makeStyles<void, "dragIcon">()((theme, _params, classes) => {
       padding: theme.spacing(1.5),
       justifyContent: "stretch",
       backgroundImage: `linear-gradient(to top, transparent, ${
-        theme.palette.background.paper
+        theme.palette.background.menu
       } ${theme.spacing(1.5)}) !important`,
     },
     toolbarGrid: {
       padding: theme.spacing(2),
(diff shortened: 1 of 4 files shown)


jtbandes:
There's not really enough stories to show what the new colors look like in context, but this one looks worse (tooltip is basically the same color as the background):

image

2metres:
@jtbandes Do we want the the tooltip darker or lighter? I struggled to find the right contrast

jtbandes:
Hard to say. Is there a reason we went with a dark tooltip in dark mode rather than inverting? What does mui do by default?

jtbandes:
I guess that's what happens when you hold down the wrong modifier keys and hit enter

2metres:
> Hard to say. Is there a reason we went with a dark tooltip in dark mode rather than inverting? What does mui do by default?

I think MUI default is dark on dark, dark on light
```
alpha(theme.palette.grey[700], 0.92)
```

jtbandes (changes requested):
Not entirely sure what problem we are solving here. I played with the preview deploy and I find the new tooltips in dark mode to be basically invisible. I do agree that the hue is more accurate to the other colors we use, but it has the effect of making them completely blend in to the point where I cannot see the tooltip border without bringing my face really close to the screen.

Before
image
image

After
image
image

jtbandes (approved):

PR #5707, opened 2023-04-05
Add ability to toggle wireframes for markers

snosenzo:
**User-Facing Changes**

 - Add ability to toggle wireframes for markers

**Description**
All marker and scene entity topics now have the ability to toggle wireframes of all renderable types that have them. Initially tried giving people the ability to recolor to control opacity, but I think people generally will want it to be on or off~~but the `LineBasicMaterial` doesn't support opacity when rendering.~~

FG-2627


packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderablePrimitive.ts
```diff
@@ -15,7 +16,28 @@ export type EntityRenderableUserData = BaseUserData & {
...
+    renderer: Renderer,
+    userData: EntityRenderableUserData = {
+      receiveTime: -1n,
+      messageTime: -1n,
+      frameId: "",
+      pose: emptyPose(),
+      settings: PRIMITIVE_DEFAULT_SETTINGS,
+      settingsPath: [],
+      entity: undefined,
+    },
```
jtbandes, line 37: I've been wondering if we could generally do away with these made-up default values on user data...

jtbandes (approved):
Looks good but please add stories :)

jtbandes:
> `LineBasicMaterial` doesn't support opacity when rendering.


Thoughts? https://stackoverflow.com/questions/63647873/how-to-set-opacity-in-drawing-lines-use-three-js-use-linebasicmaterial

snosenzo:
> > `LineBasicMaterial` doesn't support opacity when rendering.
> 
> Thoughts? https://stackoverflow.com/questions/63647873/how-to-set-opacity-in-drawing-lines-use-three-js-use-linebasicmaterial

Really not sure how I got to that conclusion at the time, my mistake. I do think having a toggle makes sense given the original ask. My initial thinking was that they can just control the opacity via color, but in most cases I imagine a user would primarily want to just toggle the outline on/off. I have another branch with code for the other way if we decide to do it though.

jhurliman:
Too bad this wasn't implemented as an RGBA color where opacity could be used to hide the lines, but it would also allow coloring the wireframe. Maybe a future update @snosenzo?

PR #5709, opened 2023-04-05
Upgrade roslibjs to v0.0.3

defunctzombie:
**User-Facing Changes**
None

**Description**
Drops unused dependencies to fix dependabot alerts for critical vulnerabilities.


defunctzombie:
Tested with a local rosbridge and the connection was established and I received log messages.

PR #5711, opened 2023-04-05
Fix horizontal scrollbar visibility

defunctzombie:
**User-Facing Changes**
Not worth calling out

**Description**
Broken in: https://github.com/foxglove/studio/pull/5680


defunctzombie:
I broke this in #5680 cause I thought the height setting was for something else. Added a storybook to catch this in the future.

jtbandes (approved):

jtbandes:
Magical

defunctzombie:
Well the story does not work in chromatic. I merged this cause we need the fix. A story for this might be a lost cause.

PR #5712, opened 2023-04-05
[3D Panel] Fix backfilled messages from disappearing when changing subscription

snosenzo:
**User-Facing Changes**

 - Fix objects in 3D disappearing when playing after changing topic visibility

**Description**
Early return when the currentTime hasn't changed caused the seek to not be handled and set back to `false` in local state. This caused the behavior where the next time the current time change it would handle the seek, which clears renderables in the renderer. This would be after the backfill messages had been processed causing the previously backfilled messages to be cleared from the renderer. 

FG-2779

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
+++ b/packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
@@ -737,7 +737,7 @@ export function ThreeDeeRender(props: {
   // Keep the renderer currentTime up to date and handle seeking
   useEffect(() => {
     const newTimeNs = currentTime ? toNanoSec(currentTime) : undefined;
-    if (!renderer || newTimeNs == undefined || newTimeNs === renderer.currentTime) {
+    if (!renderer || newTimeNs == undefined) {
       return;
     }
     const oldTimeNs = renderer.currentTime;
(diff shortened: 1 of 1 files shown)


defunctzombie (approved):
Probably good to add some of your thinking as a code comment with this effect. The existing comment definitely is light on all of the learnings you have on when/why to update this effect and the cases to test for. If we can't easily make tests for this we can at least enumerate a few examples.

packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
```diff
@@ -737,7 +737,7 @@ export function ThreeDeeRender(props: {
   // Keep the renderer currentTime up to date and handle seeking
   useEffect(() => {
     const newTimeNs = currentTime ? toNanoSec(currentTime) : undefined;
-    if (!renderer || newTimeNs == undefined || newTimeNs === renderer.currentTime) {
```
defunctzombie, line 740: What was the original thinking for this part of the `if` statement? Without a comment we are a bit in the dark on whether removing this will break something else?

snosenzo: Was trying to gate `setCurrentTime` if the time hadn't changed if the renderer wanted to do something on the current time update.

jhurliman:
Confirmed this fixes the 3D model from disappearing in my scene when pausing then resuming playback!

PR #5714, opened 2023-04-05
Change default layout name to "Default layout"

defunctzombie:
**User-Facing Changes**
Not worth calling out

**Description**
Previously the name was "Default".


jtbandes (approved):

2metres (approved):

amacneil:
How do people feel about "Workspace"?

Would require changing in a few other places too.

jtbandes:
Workspace seems like a significant change for little benefit. Tons of code and docs would need updating, or code & docs would be perpetually different.

defunctzombie:
Closing. We have a small "LAYOUT" label in the dropdown now.

Issue #5640, opened 2023-03-28
`cameraInfoTopicFullObject` will never be convertible to `SUPPORTED_IMAGE_SCHEMAS`


james-rms:
```    if (
      cameraInfoTopicFullObject &amp;&amp;
      topicIsConvertibleToSchema(cameraInfoTopicFullObject, SUPPORTED_IMAGE_SCHEMAS)
    ) {
      subs.push({
        topic: cameraInfoTopicFullObject.name,
        preload: false,
        convertTo: pickConvertToSchema(cameraInfoTopicFullObject, SUPPORTED_IMAGE_SCHEMAS),
      });
    }```
this code doesn't work. `cameraInfoTopicFullObject` will never be convertible to `SUPPORTED_IMAGE_SCHEMAS`

[Slack Message](https://foxglove.slack.com/archives/C028UEY858S/p1679958776812259)

PR #5715, opened 2023-04-05
Fix camera info topic subscription in image panel

defunctzombie:
**User-Facing Changes**
Image panel is able to subscribe to camera calibration (info) topics again.

**Description**

The image panel had incorrect logic for the camera info subscription. It was checking the topic against supported image schemas rather than calibration schemas.

This change fixes the logic by adding the list of supported calibration schemas.

Fixes: #5640

### DIFF
--- a/packages/studio-base/src/panels/Image/lib/normalizeMessage.ts
+++ b/packages/studio-base/src/panels/Image/lib/normalizeMessage.ts
@@ -22,7 +22,7 @@ export const NORMALIZABLE_IMAGE_DATATYPES = [
   "foxglove_msgs/msg/RawImage",
   "foxglove_msgs/CompressedImage",
   "foxglove_msgs/msg/CompressedImage",
-];
+] as const;
 
 /**
  * Convert a message based on datatype to a NormalizedImageMessage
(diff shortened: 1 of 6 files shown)


packages/studio-base/src/panels/Image/lib/normalizeMessage.ts
```diff
@@ -32,7 +32,8 @@ export function normalizeImageMessage(
   message: unknown,
   datatype: string,
 ): NormalizedImageMessage | undefined {
-  switch (datatype) {
+  // Cast to the union of all supported datatypes to ensure we handle all cases
+  switch (datatype as (typeof NORMALIZABLE_IMAGE_DATATYPES)[number]) {
```
jtbandes, line 36: It's unfortunate that we have to lie to TS with a cast like this... in reality it is possible that the message comes in with some other datatype, right?

jtbandes (approved):

snosenzo (approved):

packages/studio-base/src/panels/Image/lib/normalizeMessage.ts
defunctzombie: Yes - but the function will return `undefined` for that situation. This _lie_ is about ensuring the switch statement covers all the strings in the array. We don't have to have this here but if we remove it we don't know that we subscribed to a datatype that we didn't write the switch logic to handle.

So tl;dr; we don't _have_ to do this but I found it a nice sanity check that if you add an item to the array of supported schema names you are forced to update the switch statement here.

PR #5721, opened 2023-04-06
prototype inline button to add panel between mosaic nodes

jtbandes:
**User-Facing Changes**

**Description**


jtbandes:
Product review: sounds like this is a no-go in its current form

PR #5722, opened 2023-04-06
WssErrorModal to be displayed with the new top nav

wimagguc:
**User-Facing Changes**

Since the data source sidebar is removed when the new top nav bar is switched on, the WssErrorModal is never opened in this view. This PR adds the new `WssErrorModal` to `AppBar/DataSource`

demo

### DIFF
--- a/packages/studio-base/src/components/AppBar/DataSource.tsx
+++ b/packages/studio-base/src/components/AppBar/DataSource.tsx
@@ -20,6 +20,7 @@ import {
 } from "@foxglove/studio-base/components/MessagePipeline";
 import Stack from "@foxglove/studio-base/components/Stack";
 import TextMiddleTruncate from "@foxglove/studio-base/components/TextMiddleTruncate";
+import WssErrorModal from "@foxglove/studio-base/components/WssErrorModal";
 import { PlayerPresence } from "@foxglove/studio-base/players/types";
 
 const LEFT_ICON_SIZE = 19;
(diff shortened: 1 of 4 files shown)


jtbandes (approved):
What do you think about moving the logic to check player problems inside of WssErrorModal itself? Then we would be able to avoid duplicating it? Just an idea..

defunctzombie (commented):
I do find it weird that this is at the AppBar/DataSource component level. This is specific to one kind of connection?

wimagguc:
> What do you think about moving the logic to check player problems inside of WssErrorModal itself? Then we would be able to avoid duplicating it? Just an idea..

I made these changes, much cleaner now. Thanks!

Re @defunctzombie's question:
> I do find it weird that this is at the AppBar/DataSource component level. This is specific to one kind of connection?

What would be a better place for this? It makes use of the `playerProblems` array which we'd have in `Ros1Player`, but more importantly if the dialog is open, it needs to be rendered in the document. Before it was removed it used this `renderToBody` function which isn't available (and probably for good reason):

```
        const modal = renderToBody(<WssErrorModal onRequestClose={() => modal.remove()} />);
```

defunctzombie:
👍 Good stuff. Yea I agree there is not really a better place for it.

I tried the preview deployment and it worked great.

esthersweon:
Will not include in changelog, as the top nav is still feature flagged.

PR #5724, opened 2023-04-06
Refactor camera handling into SceneExtension in 3D panel

snosenzo:
**User-Facing Changes**

 no changes

**Description**
Refactored to encapsulate all camera and displayframe/follow frame logic inside of the `CameraStateSettings` scene extension. This will enable us to completely take this extension out when using the `image` mode for the 3D panel.

 I also simplified some of the flow around the `followMode`. It should've never really been a member variable on the renderer since it is present in the config. So I removed the references from the `Renderer` and now it only has references from the config, which should be up to date with it. I also am overriding `startFrame` in order to update the cameras in the snapshot as well as recalculate or initialize the snapshot pose whenever the render or fixed frame id changes. Overall I think it's more simple and encapsulated and will more easily allow us to accomodate `image` mode


packages/studio-base/src/panels/ThreeDeeRender/renderables/CameraStateSettings.ts
```diff
@@ -146,10 +313,192 @@ export class CameraStateSettings extends SceneExtension {
     this.updateSettingsTree();
   };
 
+  // this extension has  NO RENDERABLES so the parent startFrame would do nothing
+  public override startFrame(
```
snosenzo, line 317: Most of the changes to flow are found here.
This function is aggregates necessary logic from these methods in `Renderer`
 `_updateFrames` - used to handle the setting of the posesnapshot on initial load
 `updateFollowMode` - no longer necessary do on-update changes because we're keeping track of frameId's used to create the snapshot
 `frameHandler` - used to be where the snapshot was being applied to the cameras. Now that this is a scene extension, we can just utilize `startFrame` which provides us with the necessary information and is already being called before rendering.

packages/studio-base/src/panels/ThreeDeeRender/renderables/CameraStateSettings.ts
```diff
@@ -4,113 +4,250 @@
...
+const tempVec3 = new THREE.Vector3();
+
+const tempSpherical = new THREE.Spherical();
+const tempEuler = new THREE.Euler();
+const FOLLOW_TF_PATH = ["general", "followTf"];
 export class CameraStateSettings extends SceneExtension {
-  public constructor(renderer: Renderer) {
+  // The pose of the render frame in the fixed frame when following was disabled
+
+  private unfollowSnapshotFrameIds:
```
jtbandes, line 40:
```suggestion
  /** The pose of the render frame in the fixed frame when following was disabled */
  private unfollowSnapshotFrameIds:
```

maybe add more on what it's used for?

packages/studio-base/src/panels/ThreeDeeRender/renderables/CameraStateSettings.ts
```diff
@@ -4,113 +4,250 @@
...
+    | {
+        render: UserFrameId;
+        fixed: UserFrameId;
+      }
+    | undefined;
+  public unfollowPoseSnapshot: Pose | undefined;
+
+  private controls: OrbitControls;
+  private _isUpdatingCameraState = false;
```
jtbandes, line 49: nit: inconsistent naming with `_` prefix compared to other private variables

jtbandes (commented):
haven't gotten through all the files yet, just posting some comments so far

jtbandes (commented):
Would it be easier now to extend the tests so we can actually test that all the follow modes do what they are supposed to do?

jtbandes (approved):

Issue #5725, opened 2023-04-06
images in 3d only show the first frame and don't update


jtbandes:
`_recomputeCameraModel` returns undefined if the camera model did not change between the current frame and previous, but undefined `cameraModel` return value causes the call to `this._updateImageRenderable(renderable, image, cameraModel, receiveTime, settings);` in `handleImage()` to be skipped so the new image data is discarded

[Slack Message](https://foxglove.slack.com/archives/C028UEY858S/p1680814696349569)

jhurliman:
This should be the fix (in handleImage):

`const cameraModel = this._recomputeCameraModel(renderable, cameraInfo) ?? renderable.userData.cameraModel;`

jhurliman:
It's desirable for _recomputeCameraModel to return undefined if nothing changed so the handleCameraInfo callback can be a mostly no-op if the camera model didn't change between messages, and it's uncommon for the camera model to change in real-time. That's why I think fixing it in this one place is better than changing _recomputeCameraModel

defunctzombie:
> It's desirable for _recomputeCameraModel to return undefined if nothing changed so the handleCameraInfo callback can be a mostly no-op if the camera model didn't change between messages

👍 Thanks for that extra context. I see that the function has that comment that it returns undefined but now I understand why that is the case and how it impacts the logic.

PR #5726, opened 2023-04-06
[3D] Don't skip image update when CameraInfo hasn't changed

jhurliman:
**User-Facing Changes**
- Fixes https://github.com/foxglove/studio/issues/5725

**Description**

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/Images.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/Images.ts
@@ -398,7 +398,8 @@ export class Images extends SceneExtension<ImageRenderable> {
       return;
     }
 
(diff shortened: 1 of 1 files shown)


jtbandes:
@defunctzombie looks like this line of code was touched in https://github.com/foxglove/studio/pull/5555 – do you think that is when this regressed? Thoughts on how we can add a regression test?

defunctzombie:
@jtbandes I'm pretty sure that is when it regressed. There are tests that I added in that PR which we should use to add coverage for this regression.

defunctzombie:
Thanks for surfacing this bug and the context on the related calls. Since some of the code paths have changed since the recompute calls was written I've created a different fix for the issue and included a test here: https://github.com/foxglove/studio/pull/5728

PR #5728, opened 2023-04-07
Fix image rendering in 3d panel

defunctzombie:
**User-Facing Changes**
Fix image rendering in the 3d panel.

**Description**
This change resolves a regression introduces in https://github.com/foxglove/studio/pull/5555. The change in that PR incorrectly skipped updating the image when the _recomputeCameraModel returned undefined - thinking that it would always return the latest camera model.

This fix modes the _recomputeCameraModel logic into _updateImageRenderable and changes the inputs to _updateImageRenderable to accept the CameraInfo rather than the camera model. It also updates _recomputeCameraModel to return the cached camera model if available.

_recomputeCameraModel was originally designed to return undefined so handleCameraInfo could avoid rendering the image for camera calibration messages which did not result in changes to the camera model. That logic changed in #5555 and the handleCameraInfo handler no longer renders images but instead saves the latest camera info and leaves image rendering to the handleImage handler.

Fixes: #5725


packages/studio-base/src/panels/ThreeDeeRender/renderables/Images.ts
```diff
@@ -502,7 +496,15 @@ export class Images extends SceneExtension<ImageRenderable> {
...
     renderable.userData.image = image;
-    renderable.userData.cameraModel = cameraModel;
+    const cameraModel = (renderable.userData.cameraModel = this._recomputeCameraModel(
+      renderable,
+      cameraInfo,
+    ));
+
+    // We need a valid camera model to render the image
+    if (!cameraModel) {
+      return;
```
jtbandes, line 506: Would we need to clear out/hide the existing image in this case?

jtbandes (approved):

packages/studio-base/src/panels/ThreeDeeRender/renderables/Images.ts
defunctzombie: 🤔 I don't think we did that before tho I suppose that would be correct to do.

packages/studio-base/src/panels/ThreeDeeRender/stories/ImageRender.stories.tsx
```diff
@@ -562,3 +562,152 @@ export const InfoThenImage: Story = () => {
   );
 };
 InfoThenImage.parameters = { colorScheme: "light" };
+
+// Sending an image message after an existing image should update the displayed image
+export const UpdateImage: Story = () => {
```
snosenzo, line 567: How does someone know that this story is broken? Can we make the failure or success mode more clear?

snosenzo (approved):

packages/studio-base/src/panels/ThreeDeeRender/stories/ImageRender.stories.tsx
defunctzombie: I struggle with this on stories all the time. Should we show some text in the story explaining what is expected? I suppose I could make the first image all red and second image green.

snosenzo: yeah that would work, then maybe put Green in the story name?

defunctzombie: 👍 done

PR #5734, opened 2023-04-09
Replace circular-dependency-plugin with eslint-plugin-import/no-cycle rule

defunctzombie:
**User-Facing Changes**
None

**Description**
This is a check that is better done during lint in CI rather than slowing down our build time. We were already using the eslint-plugin-import package so this change does not introduce any new packages and lets us remove the circular-dependency-plugin package.

The 3d panel is the main violator of this rule by passing the renderer into everything and also depending on everything in the renderer. This change makes an `IRenderer` interface to use as the type for all the SceneExtensions and Renderables to break the cycle.


defunctzombie:
Main downside is this takes lint time to 6 min :(

Might be able to make this better with the `maxDepth`.

jtbandes (approved):

PR #5735, opened 2023-04-10
Delete useCalloutDismissalBlocker

defunctzombie:
**User-Facing Changes**
None

**Description**
This hack is no longer relevant for MUI toolkit.

Originally added in: https://github.com/foxglove/studio/pull/2123. I've tested with a data source that produces log messages to make the log panel scroll. I was able to open the Panel context menu and navigate the "Change panel" lists without the context menu closing.

@foxymiles You did the original fix - is there some other workflow I should test?


foxymiles (approved):
That fix was targeting a CSS class specific to that fluent virtualized list so deleting this now should be harmless.

defunctzombie:
@foxymiles The css class is for react-virtualized which we still use? Tho seems not to cause issues with mui.

foxymiles:
Ah yeah you're right. I was misremembering that we removed the fluent ui virtual list but are still using react-virtualized.

But I think the callout dismissal behavior was a fluent specific issue so should not be a problem now that we've switched to MUI.

PR #5738, opened 2023-04-10
Pause playback when creating an event

esthersweon:
**User-Facing Changes**
Pauses playback when opening the event creation dialog

### DIFF
--- a/packages/studio-base/src/components/PlaybackControls/index.tsx
+++ b/packages/studio-base/src/components/PlaybackControls/index.tsx
@@ -158,8 +158,11 @@ export default function PlaybackControls(props: {
   );
 
   const toggleCreateEventDialog = useCallback(() => {
+    if (isPlaying) {
+      pause();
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/components/PlaybackControls/index.tsx
```diff
@@ -158,8 +158,11 @@ export default function PlaybackControls(props: {
   );
 
   const toggleCreateEventDialog = useCallback(() => {
+    if (isPlaying) {
+      pause();
+    }
     setCreateEventDialogOpen((open) => !open);
-  }, []);
+  }, [isPlaying, pause]);
```
defunctzombie, line 165: I'd remove `isPlaying` from this check and call `pause()` always. This way the callback itself does not change as often and I don't think there's any harm in calling this `pause()` if already paused.

jtbandes (approved):

PR #5727, opened 2023-04-07
patch protobufjs to support native bigints

jtbandes:
**User-Facing Changes**
Loading Protobuf data now preserves 64-bit integer values rather than rounding them to 53-bit integers.

**Description**

Since protobuf.js does not handle bigints natively (https://github.com/protobufjs/protobuf.js/issues/1838), I had to create a patch to use bigints instead of losing precision.

I tried switching to `@bufbuild/protobuf`, which does support bigints natively, but hit a roadblock because it modifies field names when parsing protobuf data into a message object (https://github.com/bufbuild/protobuf-es/issues/454).

Sample file: [int64.mcap.zip](https://github.com/foxglove/studio/files/11175108/int64.mcap.zip)

image

<details><summary>Script to produce sample data:</summary>

```proto
syntax = "proto3";

message Int64Test {
  int64 int64 = 1;
  uint64 uint64 = 2;
  sint64 sint64 = 3;
  fixed64 fixed64 = 4;
  sfixed64 sfixed64 = 5;
  map<int64, int64> int64map = 6;
  map<uint64, uint64> uint64map = 7;
  map<sint64, sint64> sint64map = 8;
  map<fixed64, fixed64> fixed64map = 9;
  map<sfixed64, sfixed64> sfixed64map = 10;
  repeated Nested nested = 11;
}

message Nested {
  int64 int64 = 1;
  uint64 uint64 = 2;
  sint64 sint64 = 3;
  fixed64 fixed64 = 4;
  sfixed64 sfixed64 = 5;
  map<int64, int64> int64map = 6;
  map<uint64, uint64> uint64map = 7;
  map<sint64, sint64> sint64map = 8;
  map<fixed64, fixed64> fixed64map = 9;
  map<sfixed64, sfixed64> sfixed64map = 10;
}
```

```py
from os import path
from tqdm import tqdm
from mcap_protobuf.writer import Writer
from Int64Test_pb2 import Int64Test
from google.protobuf.json_format import MessageToJson

with open(path.splitext(path.basename(__file__))[0] + ".mcap", "wb") as stream:
    writer = Writer(stream)

    frame_interval_ns = 1_000_000
    nframes = 10000
    for frame in tqdm(range(0, nframes)):
        t = frame * frame_interval_ns
        msg = Int64Test()
        msg.int64 = -999999999999999000 + frame
        msg.uint64 = 9999999999999999000 + frame
        msg.sint64 = -999999999999999000 + frame
[...]

PR #5757, opened 2023-04-12
Deserialize protobuf Timestamp/Duration seconds as number rather than bigint

jtbandes:
**User-Facing Changes**
Fixed an issue where `time` and `duration` types in Protobuf data were deserialized differently from other encodings, which could lead to errors in the app.

**Description**
This regressed in #5727, where we changed all int64 to be deserialized as bigint, without realizing that this also included Timestamp and Duration built-in types. Studio depends on those being `{sec: number, nsec: number}`, but after #5727 they were `{sec: bigint, nsec: number}`.

This change adds a postprocessing step to convert the `sec` field to a number, and throw an error if the value is too large to represent without rounding. This replaces some old logic which simply renamed the fields.

Fixes FG-2916


packages/mcap-support/src/parseProtobufSchema.ts
```diff
@@ -0,0 +1,74 @@
...
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import protobufjs from "protobufjs";
+import { FileDescriptorSet } from "protobufjs/ext/descriptor";
+
+import { protobufDefinitionsToDatatypes, stripLeadingDot } from "./protobufDefinitionsToDatatypes";
+import { MessageDefinitionMap } from "./types";
+
+export function parseProtobufSchema(
```
defunctzombie, line 11: Maybe this is an internal only function but a comment on its intended purpose/use/any gotchas would be helpful for future maintenance.

defunctzombie (approved):
Nice hack to tackle this prior to making the deserializer. Wonder what exciting edge cases this will unearth in a month.

PR #5771, opened 2023-04-13
Use CameraInfo to define camera projection in ImageMode

snosenzo:
**User-Facing Changes**

none (experimental feature)

**Description**
In the experimental image panel, the camera used to render the scene will now use the last received camera info message on the selected topic to define its projection matrix. Note that the corresponding image topic has to be turned on in the topics of the image panel in order to see the image in the panel. 

FG-2684


packages/studio-base/src/panels/ThreeDeeRender/IRenderer.ts
```diff
@@ -182,7 +182,7 @@ export interface IRenderer extends EventEmitter<RendererEvents> {
   publishClickTool: PublishClickTool;
 
   /** only public for testing - prefer to use `getCameraState` instead */
-  cameraStateSettings: CameraStateSettings;
+  cameraStateSettings: ICameraHandler;
```
jtbandes, line 185: time to rename to `cameraHandler`?

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -352,12 +352,18 @@ export class Renderer extends EventEmitter<RendererEvents> implements IRenderer
       preload: config.scene.transforms?.enablePreloading ?? true,
     });
 
+    let cameraHandler;
```
jtbandes, line 355: not sure this variable is necessary, or at least the `this.cameraStateSettings = cameraHandler` lines can be deduped

packages/studio-base/src/panels/ThreeDeeRender/renderables/ICameraHandler.ts
```diff
@@ -0,0 +1,30 @@
...
+
+import { CameraState } from "../camera";
+
+export interface ICameraHandler extends SceneExtension {
+  /**
+   * Gets the active camera to use for rendering the scene
+   * */
+  getActiveCamera(): THREE.PerspectiveCamera | THREE.OrthographicCamera;
+  /**
+   * Sets the camera state and should ensure that it isn't already being updated when set
```
jtbandes, line 16: what does "should ensure that it isn't already being updated when set" mean?

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode.ts
```diff
@@ -27,14 +38,58 @@ const CALIBRATION_TOPIC_PATH = ["imageMode", "calibrationTopic"];
...
+        info: CameraInfo;
+      }
+    | undefined;
 
-export class ImageMode extends SceneExtension {
-  public constructor(renderer: IRenderer) {
+  private cameraState: CameraState = DEFAULT_CAMERA_STATE;
+
+  private cameraInfoByTopic: Map<string, CameraInfo> = new Map();
+  private cameraImageByTopic: Map<string, AnyImage> = new Map();
```
jtbandes, line 70: not sure we still need these to be maps, since this extension only handles one camera info & one image at a time. Anyway, it looks like at least `cameraImageByTopic` is unused so can be deleted

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode.ts
```diff
@@ -3,13 +3,23 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
 import { set } from "lodash";
+import * as THREE from "three";
 
 import { filterMap } from "@foxglove/den/collection";
+import { CameraCalibration } from "@foxglove/schemas";
 import { SettingsTreeAction } from "@foxglove/studio";
+import { PinholeCameraModel } from "@foxglove/studio-base/../../den/image";
+import { AnyImage } from "@foxglove/studio-base/panels/ThreeDeeRender/renderables/Images";
```
jtbandes, line 12: import hygiene–wise, ImageMode shouldn't import anything from Images since they should be independent scene extensions

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode.ts
```diff
@@ -27,14 +38,58 @@ const CALIBRATION_TOPIC_PATH = ["imageMode", "calibrationTopic"];
...
+  perspective: true, // not used
+  phi: 60, // not used
+  target: [0, 0, 0],
+  targetOffset: [0, 0, 0], // not used
+  targetOrientation: [0, 0, 0, 1], // not used
+  thetaOffset: 45, // not used
+  fovy: 45, // not used because focal length from pinhole model is used
+  near: 0.001,
+  far: 1000,
+};
```
jtbandes, line 56: Feels like the ImageMode extension should almost not even know about the CameraState type. We currently show no settings for any of these (including near & far) in the panel settings. If we want near & far to be configurable we could add those back, but could consider storing them with the rest of the image settings rather than using the cameraState?

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode.ts
```diff
@@ -27,14 +38,58 @@ const CALIBRATION_TOPIC_PATH = ["imageMode", "calibrationTopic"];
...
+  private cameraState: CameraState = DEFAULT_CAMERA_STATE;
+
+  private cameraInfoByTopic: Map<string, CameraInfo> = new Map();
+  private cameraImageByTopic: Map<string, AnyImage> = new Map();
+  private zoom = new THREE.Vector2();
+  public constructor(renderer: IRenderer, aspect: number) {
     super("foxglove.ImageMode", renderer);
 
+    this.camera = new THREE.PerspectiveCamera();
+    // have it pointing in the correct direction
```
jtbandes, line 76: This comment says "what" but not "why" – is this something about images using a different coordinate convention? A more complete "why" explanation would be helpful here

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode.ts
```diff
@@ -215,6 +271,214 @@ export class ImageMode extends SceneExtension {
...
+  }
+
+  private getImageModeSettings(): {
+    readonly calibrationTopic?: string;
+    readonly imageTopic?: string;
+  } {
+    return this.renderer.config.imageMode;
+  }
+
+  private updateCamera(): void {
```
jtbandes, line 332: this one seems important enough to deserve a comment

jtbandes (commented):
This is great progress! Let's take this opportunity to start adding some stories that demonstrate the correct functionality (& ideally also show that resizing works with a `play` function)

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode.ts
```diff
@@ -3,13 +3,23 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
 import { set } from "lodash";
+import * as THREE from "three";
 
 import { filterMap } from "@foxglove/den/collection";
+import { CameraCalibration } from "@foxglove/schemas";
 import { SettingsTreeAction } from "@foxglove/studio";
+import { PinholeCameraModel } from "@foxglove/studio-base/../../den/image";
```
jtbandes, line 11:
```suggestion
import { PinholeCameraModel } from "@foxglove/den/image";
```

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode.ts
snosenzo: I mostly kept the maps because I thought backfill wouldn't happen when this scene extension selected a camera info topic that was already being used by the `Image` scene extension because it wouldn't trigger a subscription change. This lets us store them if one isn't selected in this panel but is being subscribed to elsewhere so we wouldn't need to rely on the backfill. Might be overly cautious since we might not be using the `Image` scene extension in `image` mode anyway

defunctzombie:
> Note that the corresponding image topic has to be turned on in the topics of the image panel in order to see the image in the panel.

Is this a temporary limitation with a later project goal being that this automatically happens when you select the "image" topic in the settings?

What happens if you don't have a camera calibration data? We still show the image and ImageAnnotations but cannot show any scene entities?

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode.ts
snosenzo: > > Note that the corresponding image topic has to be turned on in the topics of the image panel in order to see the image in the panel.
> 
> Is this a temporary limitation with a later project goal being that this automatically happens when you select the "image" topic in the settings?

This is a temporary limitation. This PR is just to have the camera using the right projection matrix in image mode based on the camera info. 

> What happens if you don't have a camera calibration data? We still show the image and ImageAnnotations but cannot show any scene entities?

That's a good question. I'm not sure what the expected behavior here would be. I imagine just having sensible defaults might be fine in that case, but I'm not sure if we'd then want people to be able to customize those camera settings.

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
snosenzo: Ah yeah initially I wasn't making the ICameraHandler extend SceneExtension so I needed another variable to prevent casting and it wouldn't let me declare a variable in the switch case without a closure. don't have that problem now that I'm extending though

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode.ts
snosenzo: The weird thing here is that there are public methods on the renderer that expose this (to sync camera state), hence the ICameraHandler. I'm not sure how we want to handle this.  I could just update the cameraHandler to just be either `CameraStateSettings | ImageMode` and just have ifs in those methods to direct what types to use or whether to use them at all, but that's part of what I was trying to avoid with the camera interface

packages/studio-base/src/panels/ThreeDeeRender/renderables/ICameraHandler.ts
jtbandes: bump

defunctzombie:
> That's a good question. I'm not sure what the expected behavior here would be. I imagine just having sensible defaults might be fine in that case, but I'm not sure if we'd then want people to be able to customize those camera settings.

For parity with the existing image panel there definitely needs to be some notion of being able to load a file with only image data and image annotations (no calibration data). The image annotation are relative to the image coordinates.

Without a calibration message I don't think we'd be able to render any of the 3d scene elements so would need to inform the user. A nice future enhancement would be the manual adjustment/setting of the calibration or picking a frame for the image if one is not available maybe. There are some related ideas to that concept here: https://github.com/foxglove/studio/pull/5415. The drawback in 5415 is that the "example camera info" is rather arbitrary to one user but this idea could be extended.

packages/studio-base/src/panels/ThreeDeeRender/renderables/ICameraHandler.ts
```diff
@@ -0,0 +1,30 @@
...
+   * Gets the active camera to use for rendering the scene
+   * */
+  getActiveCamera(): THREE.PerspectiveCamera | THREE.OrthographicCamera;
+  /**
+   * Sets the camera state
+   * @param state - The new camera state
+   */
+  setCameraState(state: CameraState | undefined): void;
+  /**
+   * Gets the state of the camera
```
jtbandes, line 21: Would be nice for this comment to explain when you would expect to see undefined.

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode.ts
```diff
@@ -215,6 +276,219 @@ export class ImageMode extends SceneExtension {
...
+    }
+    // Adapted from https://github.com/ros2/rviz/blob/ee44ccde8a7049073fd1901dd36c1fb69110f726/rviz_default_plugins/src/rviz_default_plugins/displays/camera/camera_display.cpp#L568
+    this.zoom.set(1.0, 1.0);
+
+    const { width: imgWidth, height: imgHeight } = model;
+
+    const fx = model.P[0]!;
+    const fy = model.P[5]!;
+    const windowAspect = this.aspect;
+    // prettier-ignore
```
jtbandes, line 463: let's maybe remove these prettier-ignores? I mostly used it above so the matrix could actually look like a rectangle

jtbandes (approved):
🎉

packages/studio-base/src/panels/ThreeDeeRender/stories/ImageMode.stories.tsx
```diff
@@ -0,0 +1,362 @@
...
+ImageModeResizeHandled.play = async (ctx) => {
+  const canvas = document.querySelector("canvas")!;
+  // Input attaches resize listener to parent element, so we need to resize that.
+  const parentEl = canvas.parentElement!;
+  await delay(30);
+  parentEl.style.width = "50%";
+  canvas.dispatchEvent(new Event("resize"));
+  await delay(30);
+
+  ctx.parameters.storyReady = true;
```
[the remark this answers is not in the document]
snosenzo: what is this usually used for? I thought I needed it for screenshot testing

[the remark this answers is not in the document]
snosenzo: nvm took a look -- doesn't seem like it needs the ready signal stuff

PR #5774, opened 2023-04-14
[3D panel] Fix `foxglove.Grid` and `OccupancyGrid` crash

snosenzo:
**User-Facing Changes**

- [3D Panel] Fix `foxglove.Grid` and `OccupancyGrid` topics causing a crash on opening

**Description**
The crash happens because both OccupancyGrid and foxglove.Grid use `colorMode`. So when you switch from a layout that was initialized with an OccupancyGrid and have a colorMode for topic `/map` that is 'raw', to a source with a `foxglove.Grid` on topic `/map`, that foxglove.Grid will attempt to use `colorMode=raw` from the existing layout. Both of these panels happen to throw an Exception when something outside of their specified colorModes is used so it results in a crash. I've simply changed this behavior to not throw an exception. The foxglove.Grid will leave the grid blank, and the OccupancyGrid will simply use its raw palette. In either case, the Color mode input on both topics will show as invalid.
image

FG-2774

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/OccupancyGrids.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/OccupancyGrids.ts
@@ -520,6 +520,12 @@ function paletteColorCached(
       }
       palette = rawPalette;
       break;
+    default:
+      // Default to raw palette if unknown colormode, the user will have an error already in the settings
+      if (!rawPalette) {
+        rawPalette = createRawPalette();
+      }
+      palette = rawPalette;
   }
 
   const colorRaw = palette[Math.trunc(unsignedValue)]!;
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/OccupancyGrids.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/OccupancyGrids.ts
@@ -481,7 +481,11 @@ let costmapPalette: [number, number, number, number][] | undefined;
 let mapPalette: [number, number, number, number][] | undefined;
 let rawPalette: [number, number, number, number][] | undefined;
 
-function paletteColorCached(output: ColorRGBA, value: number, color_mode: ColorModes) {
+function paletteColorCached(
+  output: ColorRGBA,
+  value: number,
+  color_mode: "costmap" | "map" | "raw",
+) {
   const unsignedValue = value >= 0 ? value : value + 256;
   if (unsignedValue < 0 || unsignedValue > 255) {
     output.r = 0;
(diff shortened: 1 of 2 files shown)


defunctzombie:
@foxymiles I added you for review here cause I think this could be educational for the work you are doing with multiple converters per topic in 3d panel. I see this as somewhat related since here it is the same topic that has two different schema names in two different data sources. Maybe the fix here could align or educate what happens for the multiple converter use case.

packages/studio-base/src/panels/ThreeDeeRender/renderables/OccupancyGrids.ts
```diff
@@ -501,13 +505,12 @@ function paletteColorCached(output: ColorRGBA, value: number, color_mode: ColorM
       mapPalette = createMapPalette();
     }
     palette = mapPalette;
-  } else if (color_mode === "raw") {
+  } else {
```
defunctzombie, line 508: This block of _if_ statements would benefit from being a switch, then you would get errors if you forget one of the blocks.

Aside from that - is it correct to assume raw if the color_mode is some other value? Will this cause the setting to update to the correct value?

packages/studio-base/src/panels/ThreeDeeRender/renderables/OccupancyGrids.ts
```diff
@@ -481,7 +481,11 @@ let costmapPalette: [number, number, number, number][] | undefined;
 let mapPalette: [number, number, number, number][] | undefined;
 let rawPalette: [number, number, number, number][] | undefined;
 
-function paletteColorCached(output: ColorRGBA, value: number, color_mode: ColorModes) {
+function paletteColorCached(
+  output: ColorRGBA,
+  value: number,
+  color_mode: "costmap" | "map" | "raw",
```
defunctzombie, line 487: Why's this not `colorMode` like all our other variable names and function arguments? Looks awkward to read when its like this.

packages/studio-base/src/panels/ThreeDeeRender/renderables/OccupancyGrids.ts
```diff
@@ -481,7 +481,11 @@ let costmapPalette: [number, number, number, number][] | undefined;
 let mapPalette: [number, number, number, number][] | undefined;
 let rawPalette: [number, number, number, number][] | undefined;
 
-function paletteColorCached(output: ColorRGBA, value: number, color_mode: ColorModes) {
+function paletteColorCached(
```
defunctzombie, line 484: While you are here and figured out this function could you add a comment so we know what this is for in the future?

defunctzombie (approved):
Something feels _off_ to me about this but I can't verbalize what so if it works :shipit: . Maybe with the multi-converter work we'll re-structure how we do this logic anyway.

jtbandes (approved):

packages/studio-base/src/panels/ThreeDeeRender/renderables/OccupancyGrids.ts
snosenzo: THis will not cause the settings to update to the correct value. This just won't cause a panel error if this function runs with an unsupported color value.

packages/studio-base/src/panels/ThreeDeeRender/renderables/OccupancyGrids.ts
snosenzo: It was a community contribution.

packages/studio-base/src/panels/ThreeDeeRender/renderables/OccupancyGrids.ts
snosenzo: Added comment explaining

PR #5775, opened 2023-04-14
Foxglove websocket: Clear state when serverInfo's sessionId changed

achim-k:
**User-Facing Changes**
- Foxglove websocket: Clear state when `serverInfo`'s `sessionId` changed

**Description**
Allows the server to clear state by sending a new `serverInfo` message with a different (or undefined) sessionId. Note that this will not reset the client's registered subscribers / publishers.

Fixes FG-2791

### DIFF
--- a/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
+++ b/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
@@ -186,7 +186,6 @@ export default class FoxgloveWebSocketPlayer implements Player {
       this._problems.clear();
       this._channelsById.clear();
       this._channelsByTopic.clear();
-      this._setupPublishers();
       this._servicesByName.clear();
       this._serviceResponseCbs.clear();
       this._parameters.clear();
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
```diff
@@ -263,6 +244,24 @@ export default class FoxgloveWebSocketPlayer implements Player {
...
+      this._publishedTopics = undefined;
+      this._subscribedTopics = undefined;
+      this._advertisedServices = undefined;
+      this._datatypes = new Map();
+
+      for (const topic of this._resolvedSubscriptionsByTopic.keys()) {
+        this._unresolvedSubscriptions.add(topic);
+      }
+      this._resolvedSubscriptionsById.clear();
+      this._resolvedSubscriptionsByTopic.clear();
```
achim-k, line 264: @jtbandes I guess this should only be done if the serverInfo's `sessionId` changed?

jtbandes: Yes, if you think the server might send serverInfo multiple times with the same sessionId, then I agree we probably need to track the latest sessionId and only reset when it changes.

jtbandes (approved):

PR #5781, opened 2023-04-15
Add warning to RemoteDataSourceFactory

defunctzombie:
**User-Facing Changes**
None (not worth noting)

**Description**
When starting a connection to a remote file, users will see this warning that makes them aware that remote files may have poor performance and links to data platform.

image


packages/studio-base/src/dataSources/RemoteDataSourceFactory.tsx
```diff
@@ -64,6 +65,15 @@ class RemoteDataSourceFactory implements IDataSourceFactory {
     ],
   };
 
+  public warning = (
+    <>
+      Loading large remote files can be slow. For better performance use{" "}
```
jtbandes, line 70:
```suggestion
      Loading large remote files can be slow. For better performance, use{" "}
```

jtbandes (approved):

amacneil:
As is this feels like more of an upsell than a technical warning. Could we add something to our docs (or a blog post) explaining the technical limitations here? 

Would also be good to give a specific indication of where we expect this to break down (e.g. I have been telling folks 1gb, but if we think it's something else let me know). I know this is going to be file dependent, but we should be able to come up with a sane recommended file size limit.

I was picturing the warning (player warning?) only appeariung after you try to open a file over the recommended limit. I guess having it in this dialog is ok too, but still think we should add an actual recommended limit and link to technical context.

amacneil (approved):

PR #5786, opened 2023-04-18
Patch chart.js to avoid providing invalid options to the NumberFormatter

defunctzombie:
**User-Facing Changes**
Fixes a crash in the plot panel when working with Float64 max values.

**Description**
When encountering datasets with large numbers (like Number.MAX_VALUE), the tick claculation
logic produces NaN. These NaN values would then get used to create NumberFormatter with min/max fraction digits.

DOM interfaces do not allow using NaN for these options and will throw. This change checks for NaN prior to building
the options for the formatter and uses a safe value instead of NaN to avoid throwing and crashing the panel.


defunctzombie:
Not sure how I feel about this patch but it does seem to do the trick to avoid a crash.

.gitignore
```diff
@@ -13,6 +13,7 @@ storybook-static/
 !.yarn/releases
 !.yarn/plugins
 !.yarn/sdks
+!.yarn/patches
```
jtbandes, line 16: Should we keep it in the same place with our other patches?

jtbandes (approved):

.gitignore
defunctzombie: Yes - somehow I thought this was the place but I see now that this is where yarn put it after I ran the command `yarn` told me to run.

Maybe we should move it here since this is where yarn put the patch after running `yarn commit-patch` ?

jtbandes: Yeah I suppose moving all of them here could make sense if that is the default location?

defunctzombie: Instead of changing our approach, I've updated the yarn config to set the patch directory to our existing directory: https://github.com/foxglove/studio/pull/5786/files#diff-88fbe28c4102501b94961511a0d70ff895bf39970b4d3fc11917794a239117c5R9

Now when you use `yarn commit-patch` you get a file in the existing patches directory.

.yarnrc.yml
```diff
@@ -6,6 +6,8 @@ defaultSemverRangePrefix: ""
 
 nodeLinker: node-modules
 
+patchFolder: ./patches
```
jtbandes, line 9: What does this do? Make it automatically put patches there when using `-s`?

defunctzombie: yep - when you do `yarn patch`, yarn gives you a command to run to finalize the patch. This is the folder where it puts the patch it creates from your changes.

PR #5787, opened 2023-04-18
Fix multi-topic playback for unindexed local mcap files

snosenzo:
**User-Facing Changes**

- Fix multi-topic playback for unindexed local mcap files

**Description**

Unindexed MCAP files use the `McapStreamingIterableSource`. The message iterator returned from this source yields message values by channel first, then time-sorted within the channel. This cause messages to be iterated through by the player to also be in that order. So it would only read messages into the current frame from the first subscribed topic and no other ones because they were after all of the messages in that first topic

FG-2863

### DIFF
--- a/packages/studio-base/src/players/IterablePlayer/Mcap/McapUnindexedIterableSource.ts
+++ b/packages/studio-base/src/players/IterablePlayer/Mcap/McapUnindexedIterableSource.ts
@@ -32,7 +32,8 @@ const DURATION_YEAR_SEC = 365 * 24 * 60 * 60;
 
 type Options = { size: number; stream: ReadableStream<Uint8Array> };
 
-export class McapStreamingIterableSource implements IIterableSource {
+/** Only efficient for small files */
+export class McapUnindexedIterableSource implements IIterableSource {
   private options: Options;
   private msgEventsByChannel?: Map<number, MessageEvent<unknown>[]>;
   private start?: Time;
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/players/IterablePlayer/Mcap/McapUnindexedIterableSource.ts
```diff
@@ -242,21 +243,32 @@ export class McapStreamingIterableSource implements IIterableSource {
     }
 
     const topicsSet = new Set(topics);
+    const messagesWithChannelId: { msgEvent: MessageEvent<unknown>; channelId: number }[] = [];
```
jtbandes, line 246: Maybe just make this an array of IteratorResult so we don't have to destructure and re-create objects in the `yield` loop?

jtbandes (approved):

packages/studio-base/src/players/IterablePlayer/Mcap/McapUnindexedIterableSource.ts
```diff
@@ -243,31 +243,28 @@ export class McapUnindexedIterableSource implements IIterableSource {
     }
 
     const topicsSet = new Set(topics);
-    const messagesWithChannelId: { msgEvent: MessageEvent<unknown>; channelId: number }[] = [];
+    const resultMessages = [];
```
jtbandes, line 246: Is TS really happy with this untyped array?

snosenzo: yeah there are no errors. initially its `any[]`, but then after pushing it takes on the pushed type, all I have to do is the `as const` so that the `yield` doesn't yell at me

PR #5791, opened 2023-04-19
Support multiple message converters per topic.

foxymiles:
**User-Facing Changes**
Support multiple message converters per topic.

**Description**
Support multiple message converters per topic.

The approach here depends on message converters being registered in a consistent order. In return we can use existing layouts without migration. 

The heuristic for picking a namespaced settings key under `/topics` for a topic is as follows:
1. If a handler handles a topic's original schema directly, that handler claims `/topic/$topicName`.
2. If no direct handler for the topic's original schema exists, but a handler can convert the first listed schema in a topic's `convertibleTo` array we also claim `/topic/$topicName`. This allows existing layouts to work without conversion alongside namespaced topics.
3. Finally, if no direct handler exists and a handler can convert the topic but is not the first listed converter we claim `/topic/$topicName$delimiter$convertedSchemaName`. This makes it possible to support multiple message converters for a topic as long as the restriction that the `[topic, convertedSchema]` pair remains unique.

Implementing this will require updating all SceneExtensions. As a proof of concept I've only updated Poses, OccupancyGrids and SceneEntities here.


jtbandes:
Decision from meeting is that it's preferable to store these in an array (like #5794) and it's okay to mark layouts as dirty when that migration happens. Merging this will be blocked by a version warning when an old app loads a new config (FG-3099 / FG-3100).

foxymiles:
Closing this in favor of an approach involving migrating the config to an array.

PR #5792, opened 2023-04-19
Move Data Source dialog state into workspace context (and rename)

2metres:
**User-Facing Changes**
None

**Description**
Moves OpenDialog state into workspace context so that it can be opened from within the AppBar and AppMenu. Rename to more accurately describe component's functionality in the global workspace.


packages/studio-base/src/Workspace.tsx
```diff
@@ -620,11 +611,14 @@ function WorkspaceContent(props: WorkspaceProps): JSX.Element {
...
       {showSignInForm && <SignInFormModal />}
-      {showOpenDialog != undefined && (
-        <OpenDialog
-          activeView={showOpenDialog.view}
-          activeDataSource={showOpenDialog.activeDataSource}
-          onDismiss={() => setShowOpenDialog(undefined)}
+      {dataSourceDialog.open && (
+        <DataSourceDialog
+          activeView={dataSourceDialog.item}
+          setActiveView={dataSourceDialogActions.selectItem}
```
defunctzombie, line 617: rename `selectItem` to `activeView`? Since we can pick the field names here it seems like we already referred to these as views?

packages/studio-base/src/Workspace.tsx
```diff
@@ -686,8 +679,33 @@ function WorkspaceContent(props: WorkspaceProps): JSX.Element {
...
+
+  const { currentUser } = useCurrentUser();
+
+  const { currentUserRequired } = useInitialDeepLinkState(props.deepLinks ?? DEFAULT_DEEPLINKS);
+
+  const playerPresence = useMessagePipeline(selectPlayerPresence);
+
+  const isPlayerPresent = playerPresence !== PlayerPresence.NOT_PRESENT;
+
+  const showSignInForm = currentUserRequired && currentUser == undefined;
```
defunctzombie, line 694: Not for this PR but we should figure out how we will strip this `showSignInForm` stuff from this repo.

packages/studio-base/src/Workspace.tsx
```diff
@@ -686,8 +679,33 @@ function WorkspaceContent(props: WorkspaceProps): JSX.Element {
 }
 
 export default function Workspace(props: WorkspaceProps): JSX.Element {
```
defunctzombie, line 681: Not something this PR did but reading this file is kind of out of hand - it now has two components? Workspace and WorkspaceContent? How did we get to that setup?

packages/studio-base/src/Workspace.tsx
```diff
@@ -686,8 +679,33 @@ function WorkspaceContent(props: WorkspaceProps): JSX.Element {
...
+  const [showOpenDialogOnStartup = true] = useAppConfigurationValue<boolean>(
+    AppSetting.SHOW_OPEN_DIALOG_ON_STARTUP,
+  );
+
+  const { currentUser } = useCurrentUser();
+
+  const { currentUserRequired } = useInitialDeepLinkState(props.deepLinks ?? DEFAULT_DEEPLINKS);
+
+  const playerPresence = useMessagePipeline(selectPlayerPresence);
```
defunctzombie, line 690: nit: The only use of this seems to be to calculate `isPlayerPresent` ? so the selector could itself do this logic and then you'll only trigger the value diff when the player presence changes rather than any presence value change.

packages/studio-base/src/Workspace.tsx
```diff
@@ -469,11 +458,13 @@ function WorkspaceContent(props: WorkspaceProps): JSX.Element {
       return (
         <DataSourceSidebar
           disableToolbar={enableNewTopNav}
-          onSelectDataSourceAction={() => setShowOpenDialog({ view: "start" })}
+          onSelectDataSourceAction={() =>
+            dataSourceDialogActions.set((oldValue) => ({ ...oldValue, open: true }))
```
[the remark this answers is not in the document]
foxymiles: Yeah I agree it would be better to have some symmetry here. We can adjust the naming.

packages/studio-base/src/Workspace.tsx
foxymiles: This was at @jtbandes request because he found it awkward that `Workspace` required the workspace store in a containing component instead of being self-contained so we moved the component that provides the store into this file as a simple wrapper.

packages/studio-base/src/context/WorkspaceContext.ts
```diff
@@ -113,6 +124,25 @@ export function useWorkspaceActions(): WorkspaceActions {
...
+        ) => {
+          set({
+            dataSourceDialog: {
+              activeDataSource: dataSource,
+              item: selectedDataSourceDialogItem,
+              open: true,
+            },
+          });
+        },
+      },
```
jtbandes, line 144: Feels like this is trying to be zustand but didn't quite get there.

jtbandes (approved):

packages/studio-base/src/context/WorkspaceContext.ts
foxymiles: Oh it's zustand alright.

foxymiles: I just pulled the actions out of the store. I tend to favor this approach now.

defunctzombie (commented):
👍 Lets be mindful to add some comments to action functions (especially those with arguments) so we can help ourselves understand what they do with the arguments.

PR #5794, opened 2023-04-19
[3D Image] Initial support for ImageAnnotations

jtbandes:
**User-Facing Changes**
None (feature flagged)

**Description**
- Add "Image annotations" section to settings
- Store settings in `config.imageMode.annotations` for now – might want to merge this with `config.topics` later, once we settle on the data structure
- Add `RenderableTopicAnnotations` to hold all renderables for a single topic
   - Add children `RenderablePointsAnnotation` and `RenderableLineListAnnotation`, which use `cameraModel.projectPixelTo3dPlane()` to place points in 3D space so they align with the image

Resolves FG-2683
Resolves FG-2685


packages/studio-base/src/panels/ThreeDeeRender/renderables/imageMode/ImageMode.ts
```diff
@@ -15,22 +15,23 @@ import {
...
+import { ImageAnnotations } from "./ImageAnnotations";
+import type { IRenderer } from "../../IRenderer";
+import { PartialMessageEvent, SceneExtension } from "../../SceneExtension";
+import { SettingsTreeEntry } from "../../SettingsManager";
 import {
   CAMERA_CALIBRATION_DATATYPES,
   COMPRESSED_IMAGE_DATATYPES,
   RAW_IMAGE_DATATYPES,
-} from "../foxglove";
+} from "../../foxglove";
```
defunctzombie, line 26: Can we use our absolute import syntax for these? I thought we had a lint rule to prevent `../` imports but I guess not.

jtbandes: Seems like we mostly use `../` imports to refer to other code in the same panel (ThreeDeeRender).

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/annotations/TopicAnnotationsRenderable.ts
```diff
@@ -0,0 +1,75 @@
...
+  #lineList?: LineListRenderable;
+  #scale = 0; // TODO: get correct initial scale
+
+  public dispose(): void {
+    this.#points?.dispose();
+  }
+
+  public updateScale(scale: number): void {
+    this.#scale = scale;
+    //TODO: re-render immediately with new scale
```
snosenzo, line 24: should put an updateScale method on the renderable that only changes what needs to be changed for scale

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/annotations/LineListRenderable.ts
```diff
@@ -0,0 +1,117 @@
...
+import { LineSegmentsGeometry } from "three/examples/jsm/lines/LineSegmentsGeometry";
+
+import { PinholeCameraModel } from "@foxglove/den/image";
+import { PointsAnnotation as NormalizedPointsAnnotation } from "@foxglove/studio-base/panels/Image/types";
+
+import { SRGBToLinear } from "../../../color";
+
+const tempVec3 = new THREE.Vector3();
+
+export class LineListRenderable extends THREE.Object3D {
```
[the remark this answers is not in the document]
jtbandes: I think that's a good idea and probably necessary for picking. What I don't like about it is that Renderable requires a reference to the IRenderer – I was hoping to keep that out of these classes.

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/annotations/RenderableTopicAnnotations.ts
```diff
@@ -0,0 +1,164 @@
...
+    this.#cameraModelNeedsUpdate ||= this.#cameraModel !== cameraModel;
+    this.#cameraModel = cameraModel;
+  }
+
+  public setAnnotations(annotations: NormalizedAnnotation[]): void {
+    this.#annotationsNeedsUpdate ||= this.#annotations !== annotations;
+    this.#annotations = annotations;
+  }
+
+  public update(): void {
```
snosenzo, line 57: Can we break this function up a bit or simplify this structure? I'm thinking it would be great if we could just always iterate through all annotations and just only call these functions on them if the needsupdate functions are true. Can we make the Annotation primitives extend a parent class or implement an interface?

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/annotations/ImageAnnotations.ts
```diff
@@ -0,0 +1,204 @@
...
+      renderable.setCameraModel(this.#cameraModel);
+      this.#renderablesByTopic.set(messageEvent.topic, renderable);
+      this.add(renderable);
+    }
+
+    renderable.setAnnotations(annotations);
+    renderable.update();
+  }
+
+  #handleTopicVisibilityChange(topic: Topic, action: SettingsTreeAction): void {
```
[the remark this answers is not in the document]
jtbandes: Currently we don't have any other settings besides topic visibility. I could still try to refactor it a bit though if you think it's a good idea.

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/annotations/RenderableLineListAnnotation.ts
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    const { points, outlineColor, outlineColors, thickness } = this.#annotation;
+    if (!outlineColor || outlineColor.a <= 0 || thickness <= 0 || points.length < 2) {
+      this.visible = false;
+      return;
+    }
+    this.visible = true;
+
+    if (this.#annotationNeedsUpdate || this.#scaleNeedsUpdate) {
```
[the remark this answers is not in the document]
jtbandes: This is here because the scale depends on `annotation.thickness`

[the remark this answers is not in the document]
snosenzo: ah, missed that :/

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/annotations/RenderableTopicAnnotations.ts
jtbandes: Skipping this to avoid having to destructure/assert things again inside the individual update functions.

snosenzo (approved):

PR #5796, opened 2023-04-19
Frame axes off by default in image mode

snosenzo:
**User-Facing Changes**

- none (experimental feature)

**Description**
Overall just passing the default visibility override for the renderables into the constructor of `FrameAxes`. Then created function to make sure defaults are used wherever they're needed. Also added short circuit in startFrame to prevent unnecessary calculations

FG-2786

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/FrameAxes.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/FrameAxes.ts
@@ -86,6 +86,7 @@ export class FrameAxes extends SceneExtension<FrameAxisRenderable> {
   private labelForegroundColor = 1;
   private labelBackgroundColor = new THREE.Color();
   private lineGeometry: LineGeometry;
+  private defaultAxisVisibility: boolean;
 
   public constructor(renderer: IRenderer) {
     super("foxglove.FrameAxes", renderer);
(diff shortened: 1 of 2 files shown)


jtbandes (approved):

packages/studio-base/src/panels/ThreeDeeRender/renderables/FrameAxes.ts
```diff
@@ -395,13 +397,19 @@ export class FrameAxes extends SceneExtension<FrameAxisRenderable> {
...
           | undefined;
-        renderable.userData.settings = { ...DEFAULT_SETTINGS, ...settings };
+        renderable.userData.settings = this.getRenderableSettingsWithDefaults(settings ?? {});
 
         this._updateFrameAxis(renderable);
       }
     }
   };
 
+  private getRenderableSettingsWithDefaults(
```
defunctzombie, line 407: Could be nice on the base renderable? Then `DEFAULT_SETTINGS` does need to be used everywhere?

snosenzo: Not sure that's possible because we want this to be used when generating the default values for the settings nodes as well, which would mean the renderables need to exist for the settings node to exist and I'm not sure that's always the case

PR #5799, opened 2023-04-20
Refactor ImageMode camera out of parent scene extension

snosenzo:
**User-Facing Changes**

- none

**Description**
Should be no experiential changes. Just thought I'd better encapsulate the image mode camera code, so that there are fewer conflicts if changes need to be made in parallel


packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageModelCamera.ts
```diff
@@ -0,0 +1,105 @@
...
+import * as THREE from "three";
+
+import { PinholeCameraModel } from "@foxglove/den/image";
+
+const DEFAULT_CAMERA_STATE = {
+  near: 0.001,
+  far: 1000,
+};
+
+export class ImageModelCamera extends THREE.PerspectiveCamera {
```
jtbandes, line 14: did you mean:
```suggestion
export class ImageModeCamera extends THREE.PerspectiveCamera {
```
or was this intentional?

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode.ts
```diff
@@ -39,25 +40,19 @@ const IMAGE_TOPIC_UNAVAILABLE = "IMAGE_TOPIC_UNAVAILABLE";
...
-  private aspect: number;
-  private camera: THREE.PerspectiveCamera;
-  private cameraState = DEFAULT_CAMERA_STATE;
+  private camera: ImageModelCamera;
   private cameraModel:
     | {
         model: PinholeCameraModel;
         info: CameraInfo;
       }
     | undefined;
```
jtbandes, line 53: Should this be owned by the new class and stay out of the ImageMode extension?

snosenzo: I initially had that but we need to share the PinholeCameraModel with the Image renderables to apply other transformations and projections, so I thought it best to keep here

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageModelCamera.ts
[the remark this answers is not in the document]
snosenzo: Ah yeah, great idea

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageModelCamera.ts
snosenzo: It was intentional, just to show its driven by the Image/Camera Model, but yeah it does look a bit confusing because the words are so close. What about `ModeledCamera`?

jtbandes (approved):

PR #5803, opened 2023-04-21
Only mount one copy of useInitialDeepLinkState

jtbandes:
**User-Facing Changes**
None (issue was not in a release)

**Description**
useInitialDeepLinkState calls selectSource. There were two copies of the hook, so two calls to selectSource, meaning we would create two players.

Fixes FG-3044
Regressed in #5792

### DIFF
--- a/packages/studio-base/src/hooks/useInitialDeepLinkState.ts
+++ b/packages/studio-base/src/hooks/useInitialDeepLinkState.ts
@@ -121,7 +121,10 @@ function useSyncTimeFromUrl(targetUrlState: AppURLState | undefined) {
   }, [playerPresence, seekPlayback, unappliedTime]);
 }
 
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/hooks/useInitialDeepLinkState.ts
```diff
@@ -121,12 +121,24 @@ function useSyncTimeFromUrl(targetUrlState: AppURLState | undefined) {
   }, [playerPresence, seekPlayback, unappliedTime]);
 }
 
+/** Ensure only one copy of the hook is mounted */
```
defunctzombie, line 124: A comment on _why_ would be nice since we don't typically do this for our hooks?

defunctzombie (approved):

PR #5808, opened 2023-04-23
Add touchend event handling in 3D panel to use object selection and m…

kqhasaki:
…easuring tool on touch screen device

**User-Facing Changes**

**Description**

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/Input.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/Input.ts
@@ -34,6 +34,11 @@ export type InputEvents = {
     worldSpaceCursorCoords: THREE.Vector3 | undefined,
     event: MouseEvent,
   ) => void;
+  touchend: (
+    cursorCoords: THREE.Vector2,
+    worldSpaceCursorCoords: THREE.Vector3 | undefined,
+    event: TouchEvent,
+  ) => void;
   keydown: (key: Key, event: KeyboardEvent) => void;
 };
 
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/panels/ThreeDeeRender/Input.ts
```diff
@@ -142,6 +147,24 @@ export class Input extends EventEmitter<InputEvents> {
 
   private onTouchEnd = (event: TouchEvent): void => {
     event.preventDefault();
+    const touch = event.changedTouches[0];
+    if (!touch || !this.startClientPos) {
+      return;
+    }
+
+    const dist = this.startClientPos.distanceTo(tempVec2.set(touch.clientX, touch.clientY));
+    this.startClientPos = undefined;
```
jtbandes, line 156: This seems wrong because `startClientPos` wasn't changed in the `touchstart` event. We might also need to check `touch.identifier` to see if it's equal to the identifier from `touchstart`.

jtbandes (commented):
Thanks for the PR!

It looks like the Input currently listens for `touchstart` & `touchend` and ignores those events. If I simply **remove** the existing listeners, touch events seem to work okay:

https://user-images.githubusercontent.com/14237/234997691-52d45653-ccfe-4390-9bb5-103111674be0.mov


Did you try this approach instead of changing the touchend handler?

kqhasaki:
Thanks for reviewing. I didn't try removing the existing listeners. Good approach!

jtbandes:
Feel free to make another PR with those changes if you're interested!

PR #5810, opened 2023-04-24
Fix chrome ResizeObserver error

foxymiles:
**User-Facing Changes**
None

**Description**
Calling the `ResizeObserver` more often than chrome can service it results in a ResizeObserver loop limit exceeded, so we need to debounce it.

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/Input.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/Input.ts
@@ -41,7 +41,7 @@ export type InputEvents = {
 export class Input extends EventEmitter<InputEvents> {
   private readonly canvas: HTMLCanvasElement;
   public canvasSize: THREE.Vector2;
-  private resizeObserver: ResizeObserver;
+  private readonly resizeObserver: ResizeObserver;
   private startClientPos?: THREE.Vector2;
   private cursorCoords = new THREE.Vector2();
   private worldSpaceCursorCoords?: THREE.Vector3;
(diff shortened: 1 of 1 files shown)


snosenzo (approved):

packages/studio-base/src/panels/ThreeDeeRender/Input.ts
```diff
@@ -57,7 +58,10 @@ export class Input extends EventEmitter<InputEvents> {
     this.canvas = canvas;
     this.canvasSize = new THREE.Vector2(canvas.width, canvas.height);
 
-    this.resizeObserver = new ResizeObserver(this.onResize);
+    // Calling the resize observer too often causes Chrome to throw an exception
+    // so we debounce it.
+    const debouncedOnResize = debounce(this.onResize);
```
defunctzombie, line 63: Reading this it was not obvious to me that debouncing this until next tick (which is what happens if you don't provide a duration) is ok against the comment of "calling too often causes Chrome". Maybe we can improve this comment with a note about next tick being ok and that is the behavior here? or link to some issue in chrome tracker?

defunctzombie (approved):

packages/studio-base/src/panels/ThreeDeeRender/Input.ts
foxymiles: To be honest I just left this at the default and it worked so I didn't try to set an explicit `wait`.

foxymiles: Without a wait it also coalesces multiple calls right? I would expect that's why this fixes it, not so much the delay to next tick.

PR #622, opened 2021-04-21
Replace bobjects with LazyMessages

defunctzombie:
Remove separate "bobjects" code paths and provide lazy messages using the existing messages code paths.

PR #5821, opened 2023-04-25
Remove unused TypicalFilterName code

defunctzombie:
**User-Facing Changes**
None

**Description**
Remove the automatic detection of filter path field from array items. Instead of creating a filter path with the field name heuristic, build a filter path with the array index.


packages/studio-base/src/components/MessagePathSyntax/useCachedGetMessagePathDataItems.ts
```diff
@@ -311,17 +310,11 @@ export function getMessagePathDataItems(
           // If we have a filter set after this, it will update the path appropriately.
           newPath = `${path}[:]`;
         } else if (typeof arrayElement === "object") {
-          // See if `arrayElement` has a property that we typically filter on. If so, show that.
-          const name = TypicalFilterNames.find((id) => id in arrayElement);
-          if (name != undefined) {
-            newPath = `${path}[:]{${name}==${arrayElement[name]}}`;
```
jtbandes, line 317: If this is not an observable change, what are we using the `newPath` variable for at all? Should we delete it entirely?

defunctzombie: There are two branches that set `newPath` to either `path[#]` or `path[:]`. No clear clue to me why since no tests fail. Maybe some storybook will fail.

jtbandes: With this change as-is, this comment would now be wrong: https://github.com/foxglove/studio/blob/5ab901d537b0a49425c0785ca46096ad3f58df39/packages/studio-base/src/components/MessagePathSyntax/useCachedGetMessagePathDataItems.ts#L40

defunctzombie: Yea - the code has not made that a reality for a long time. I'll update the comment to strip away some of the incorrect statements.

jtbandes: To provide a concrete example of where this code is used, if you have a Raw Message panel using a path with a `[:]` slice, and you click the button to open a corresponding plot panel, it will fill in the `id` as a filter when creating the plot panel.

image

🤦  It also appears that the id is not properly quoted, since it should be `{id=="c08b"}` and not `{id==c08b}`. This seems to have been an oversight because it works fine for numeric ids, but not strings.

jtbandes (changes requested):
We should either restore the functionality that was broken in #622, or delete the `newPath` stuff entirely. No reason to partially delete it when we know what the more correct code looked like and it would be easy to restore. (Or, at least, you have not provided a reason 😄)

To provide a concrete example of where this code is used, if you have a Raw Message panel using a path with a `[:]` slice, and you click the button to open a corresponding plot panel, it will fill in the `id` as a filter when creating the plot panel.

image

🤦  It also appears that the id is not properly quoted, since it should be `{id=="c08b"}` and not `{id==c08b}`. This seems to have been an oversight because it works fine for numeric ids, but not strings.


This is where it was previously used in the plot tooltip, but we've deleted that code in Studio:
image

defunctzombie:
> To provide a concrete example of where this code is used, if you have a Raw Message panel using a path with a `[:]` slice, and you click the button to open a corresponding plot panel, it will fill in the `id` as a filter when creating the plot panel.

I find this behavior completely arbitrary. Maybe in practice it works perfectly? But right now with our current logic you get the `[#]` index of the array item you selected. If you want to change that as a user you can.

I am 👎 On putting back this feature given its been missing for two years and I've not heard one single request around it.

jtbandes (approved):

jtbandes:
Anything more that needs to be done here besides resolving the merge conflict?

defunctzombie:
I followed your example above and with this change the value on the plot panel is `/markers/annotations.entities[18].frame_locked` whereas before it would be `/markers/annotations.entities[:]{id==85d7}.frame_locked`. Notably as we learned, the latter does not work with string _id_ values.

I am inclined to merge the behavior we have in this PR and see what (if any) downstream impact there is. Our SceneEntity messages use a `string` id and the current behavior would not work for them. I can see a user being impacted by this and if that surfaces we can re-visit the behavior and add a test for it if we want to add it back for some set of fields.

jtbandes (approved):

jtbandes:
Still not sure what “downstream impact” we are looking for here — we know what the impact is, don’t we? Is the question whether any users will care?

defunctzombie:
> Is the question whether any users will care?

Yea - that's the downstream impact. Whether there are literally any users of this aspect of the raw message panel - which is the only place this surfaces.

PR #5824, opened 2023-04-25
Improve panel settings text input interactivity

foxymiles:
**User-Facing Changes**
Improve panel settings text inputs interactivity.

**Description**
Our existing layout persistence code debounces layout writes but the debounce can happen at arbitrary times during the react event cycle and seems to occasionally delay the update enough to lose keystrokes. 

This moves the updates into a separate sync adapter that pushes the updates downstream to always happen in a `useEffect` (in the form of a `useAsync`) seems to fix it.

### DIFF
--- a/packages/studio-base/src/components/CurrentLayoutSyncAdapter.tsx
+++ b/packages/studio-base/src/components/CurrentLayoutSyncAdapter.tsx
@@ -3,48 +3,59 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
 import { enqueueSnackbar } from "notistack";
-import { useCallback, useEffect, useState } from "react";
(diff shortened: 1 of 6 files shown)


jtbandes (approved):
Not sure I quite understand what was wrong with the old version to cause this issue, and how the change fixes it, but it seems like a fine change/refactor.

foxymiles:
@jtbandes To be honest I'm not 100% clear on this myself. By trial and error I found that disabling the save fixed the problem and also that moving it always run in a useEffect instead of whenever the `setTimeout` ran also seemed to fix it.

foxymiles:
@defunctzombie Moved this into a sync adapter to make it more modular.

defunctzombie (approved):

PR #5827, opened 2023-04-25
Avoid updating the chart worker concurrently

defunctzombie:
**User-Facing Changes**

**Description**

### DIFF
--- a/packages/studio-base/src/components/Chart/index.tsx
+++ b/packages/studio-base/src/components/Chart/index.tsx
@@ -223,8 +223,13 @@ function Chart(props: Props): JSX.Element {
       return;
     }
 
+    console.log(data.datasets[0]?.data.length);
     return out;
   }, [data, height, options, isBoundsReset, width]);
+
+  const counter = useRef(0);
+
+  const running = useRef(false);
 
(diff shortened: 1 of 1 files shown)


defunctzombie:
There's a small nuance here that potentially the last update is not sent if there is already an update happening since no other update will trigger another send. I am not entirely sure if that happens in practice.

foxymiles (approved):
Fixes the issue I was seeing. We may want to add some logic here to do a final flush of a last message later but this seems like a good fix for now.

packages/studio-base/src/components/Chart/index.tsx
```diff
@@ -226,80 +226,105 @@ function Chart(props: Props): JSX.Element {
...
+        rpcSendRef.current = sendWrapperRef.current;
+
+        maybeUpdateScales(scales);
+        onFinishRender?.();
         return;
       }
 
-      if (canvasRef.current) {
-        throw new Error("Canvas has already been initialized");
+      if (!rpcSendRef.current) {
```
snosenzo, line 287: Can we swap this for an assert? I feel like it would read better that way

snosenzo (approved):

PR #5817, opened 2023-04-25
Allow dependabot-fix to upgrade monorepo packages in lockstep

jtbandes:
**User-Facing Changes**
None

**Description**
`yarn constraints` will now enforce:
- All `@storybook/*` packages (except `@storybook/testing-library`) are the same version
- All `@typescript-eslint/*` packages are the same version

`yarn constraints --fix` will upgrade these dependencies to the highest specified version.
The `dependabot-fix` action will run `yarn constraints --fix` to automatically apply fixes on Dependabot PRs.

Resolves FG-3001

PR #5828, opened 2023-04-25
constraints: add "includes" to match `storybook` version with `@storybook/*`

jtbandes:
**User-Facing Changes**
None

**Description**
Follow-up to #5817 
Forgot that the `storybook` package also needs to be the same version as the other `@storybook/*` packages.

### DIFF
--- a/constraints.pro
+++ b/constraints.pro
@@ -40,24 +40,24 @@ semver_max([First|Rest], Result) :-
 has_prefix(Prefix, Atom) :-
   atom_concat(Prefix, _, Atom).
 
-% Find all unique dependency versions across all workspaces that have a given Prefix, excluding any
-% dependencies in Excludes.
-all_versions_with_prefix(Prefix, Excludes, Result) :-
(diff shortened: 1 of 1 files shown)


constraints.pro
```diff
@@ -40,38 +40,38 @@ semver_max([First|Rest], Result) :-
...
     ResultList
   ),
   list_to_set(ResultList, Result).
 
 % These dependencies are released from monorepos, so we want each package to have the same version.
-% Arguments: requires_matching_versions(Prefix, Excludes)
-requires_matching_versions('@storybook/', ['@storybook/testing-library']).
-requires_matching_versions('@typescript-eslint/', []).
+% Arguments: requires_matching_versions(Prefix, Includes, Excludes)
+requires_matching_versions('@storybook/', ['storybook'], ['@storybook/testing-library']).
```
snosenzo, line 59: I don't really know prologue or whatever this language is, but where is this function defined? Why is it fine to just change like this?

snosenzo (approved):
I don't know prologue and it's very confusing but I trust you

PR #5829, opened 2023-04-25
[3D Image] add picking support for points & line lists, add stories

jtbandes:
**User-Facing Changes**
None

**Description**
Picking doesn't quite work yet because of camera issues, but the renderables are visible in the hitmap with DEBUG_PICKING enabled.


snosenzo (approved):
wondering if it would be possible to eventually auto-create assign picking material based off of the regular material used by the renderable

PR #5840, opened 2023-04-26
Refactor problems list to avoid modal

2metres:
**User-Facing Changes**

Screen Shot 2023-04-27 at 7 27 27 am

**Description**


amacneil:
👍

packages/studio-base/src/components/ProblemsList.stories.tsx
```diff
@@ -0,0 +1,87 @@
...
+          <ProblemsList />
+        </Box>
+      </MockMessagePipelineProvider>
+    );
+  },
+};
+export const DefaultChinese: StoryObj = {
+  ...Default,
+  parameters: { forceLanguage: "zh" },
+};
```
jtbandes, line 48: No need to add Chinese versions of all these since they're not localized.

jtbandes (approved):

packages/studio-base/src/components/ProblemsList.stories.tsx
2metres: Yeah thought about that. How do we go about getting these?

jtbandes: What do you mean by "these", are you talking about the error messages themselves? I don't think it's feasible for us to provide localized versions of all error messages since those can come from any various package we're using, such as `@mcap/core` or `@foxglove/wasm-zstd` or `roslibjs` or whatnot.

PR #5841, opened 2023-04-26
Check version stamp on layouts and warn if too new

foxymiles:
**User-Facing Changes**
Warn users when trying to load layouts created by newer, incompatible versions of studio.

**Description**
Adds a version stamp to layout data and displays an error dialog to the user when trying to load a layout created by a more recent version of studio with an incompatible layout format.

We want to deploy this now so that by the time we do need to do incompatible migrations to layouts the versions of the app in use will handle it gracefully.

Screenshot 2023-04-26 at 11 07 36 AM

### DIFF
--- a/packages/studio-base/src/providers/CurrentLayoutProvider/index.tsx
+++ b/packages/studio-base/src/providers/CurrentLayoutProvider/index.tsx
@@ -50,7 +50,7 @@ const log = Logger.getLogger(__filename);
 
 const SAVE_INTERVAL_MS = 1000;
 
-const MAX_SUPPORTED_LAYOUT_VERSION = 1;
+export const MAX_SUPPORTED_LAYOUT_VERSION = 1;
 
 /**
  * Concrete implementation of CurrentLayoutContext.Provider which handles automatically saving and
--- a/packages/studio-base/src/providers/CurrentLayoutProvider/index.tsx
+++ b/packages/studio-base/src/providers/CurrentLayoutProvider/index.tsx
@@ -1,6 +1,7 @@
 // This Source Code Form is subject to the terms of the Mozilla Public
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
+
 import { difference, isEqual } from "lodash";
 import { useSnackbar } from "notistack";
 import { useCallback, useEffect, useMemo, useRef, useState } from "react";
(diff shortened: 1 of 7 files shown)


packages/studio-base/src/providers/CurrentLayoutProvider/index.tsx
```diff
@@ -43,10 +44,14 @@ import { PanelConfig, UserNodes, PlaybackConfig } from "@foxglove/studio-base/ty
 import { windowAppURLState } from "@foxglove/studio-base/util/appURLState";
 import { getPanelTypeFromId } from "@foxglove/studio-base/util/layout";
 
+import { IncompatibleLayoutVersionAlert } from "./IncompatibleLayoutVersionAlert";
+
 const log = Logger.getLogger(__filename);
 
 const SAVE_INTERVAL_MS = 1000;
 
+const MAX_SUPPORTED_LAYOUT_VERSION = 1;
```
jtbandes, line 53: seems like this is not actually being written into layouts yet – should we also do that in this PR?

packages/studio-base/src/providers/CurrentLayoutProvider/index.tsx
```diff
@@ -368,5 +382,12 @@ export default function CurrentLayoutProvider({
     actions,
   });
 
-  return <CurrentLayoutContext.Provider value={value}>{children}</CurrentLayoutContext.Provider>;
+  return (
+    <CurrentLayoutContext.Provider value={value}>
+      {children}
```
defunctzombie, line 387: Do we want to render children if the layout is out of version? I think maybe not.

packages/studio-base/src/providers/CurrentLayoutProvider/index.tsx
defunctzombie: We could write it optimistically as the user themselves saves the layouts. And yes I think we do want to be writing this.

packages/studio-base/src/providers/CurrentLayoutProvider/index.tsx
foxymiles: My thinking was to keep this PR simple and get it out quickly we could wait until we get to version 2 and do it then.

packages/studio-base/src/providers/CurrentLayoutProvider/index.tsx
foxymiles: My thinking was to keep this PR simple and get it out quickly we could wait until we get to version 2 and do it then.

packages/studio-base/src/providers/CurrentLayoutProvider/index.tsx
foxymiles: Actually not rendering children pretty much blows away the entire UI including the layout selector so I think we'd need to be more targeted about this and make the error accessible to more nested components instead if we want to omit some UI.

defunctzombie (approved):
For web deployments the call to action taking you to the download page is wrong. We need some alternative - maybe asking the user to reload.

foxymiles:
> For web deployments the call to action taking you to the download page is wrong. We need some alternative - maybe asking the user to reload.

Added logic to show a different message for web.

PR #5842, opened 2023-04-26
Remove tooltip maps

defunctzombie:
**User-Facing Changes**
Improve plot and state transition panel performance

**Description**
Prior to this change, the chart hover logic would lookup the tooltip content in a `Map` keyed by a string of the `x` and `y` point. This required building a map of all the dataset points whenever the dataset was updated. This map would have the same size as the dataset which for large datasets means a large map build (and rebuild).

This change removes the tooltip maps and updates the hover logic to lookup the tooltip value from the datum in the hover event from the chart. This removes the need for building any tooltip maps leading to reduced CPU and memory use with no change in functionality.


defunctzombie:
Seeing an issue with one of the values in the state transition panel sample data showing as `null`
image

Not seeing the same on the current deployed production app.

packages/studio-base/src/panels/StateTransitions/messagesToDatasets.ts
```diff
@@ -34,23 +28,20 @@ type Args = {
...
-  let currentData: DatasetData = [];
+  let currentData: ChartDatum[] = [];
 
   const datasetIndexMap = new Map<unknown, number>();
   let lastDatasetIndex: undefined | number = undefined;
 
   for (const messages of blocks) {
     if (!messages) {
-      currentData.push({ x: NaN, y: NaN });
+      currentData.push({ x: NaN, y: NaN, value: NaN });
```
defunctzombie, line 44:
```suggestion
      currentData.push({ x: NaN, y: NaN });
```

jtbandes (approved):
My eyes glaze over when I read this chart code.

foxymiles:
> My eyes glaze over when I read this chart code.

It's some of the more difficult code in the app to understand for sure.

foxymiles (approved):
This is something I've wanted to do for a long time. Nice work!

PR #5843, opened 2023-04-26
ImageMode Render image without other Extensions

snosenzo:
**User-Facing Changes**

none (experimental)

**Description**
Removed both camera and image extensions. Rendering the image in `ImageMode` with an `ImageRenderable`. Needed to update subscription hook to recalculate for imageMode topic variables. Updated ImageMode internals to not need maps now that they can depend on backfill.

FG-3072


packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
```diff
@@ -720,7 +720,15 @@ export function ThreeDeeRender(props: {
     // Sort the list to make comparisons stable
     newSubscriptions.sort((a, b) => a.topic.localeCompare(b.topic));
     setTopicsToSubscribe((prev) => (isEqual(prev, newSubscriptions) ? prev : newSubscriptions));
-  }, [topics, config.topics, schemaHandlers, topicHandlers, config.imageMode.annotations]);
+  }, [
+    topics,
+    config.topics,
+    config.imageMode.calibrationTopic,
+    config.imageMode.imageTopic,
```
jtbandes, line 727: probably worth adding a comment here cause it is not obvious why these need to be here since they are not explicitly used in the body of the useMemo.

packages/studio-base/src/panels/ThreeDeeRender/renderables/Images/ImageRenderable.ts
```diff
@@ -102,6 +104,16 @@ export class ImageRenderable extends Renderable<ImageUserData> {
...
 
+  public renderBehindScene(): void {
+    this.#renderBehindScene = true;
+    this.#materialNeedsUpdate = true;
+  }
+
+  public renderInScene(): void {
+    this.#renderBehindScene = false;
+    this.#materialNeedsUpdate = true;
+  }
```
jtbandes, line 115: I think I would find these function names clearer if they were something more like `setRenderBehindScene(boolean)` since using "render" as the verb makes me think it is going to render something when called.

jtbandes: Maybe this can be removed for now and add later if/when we decide we want to support the image as an overlay.

packages/studio-base/src/panels/ThreeDeeRender/renderables/Images/ImageRenderable.ts
```diff
@@ -244,6 +256,14 @@ export class ImageRenderable extends Renderable<ImageUserData> {
     material.transparent = transparent;
     material.depthWrite = !transparent;
 
+    if (this.#renderBehindScene) {
+      material.depthTest = false;
```
jtbandes, line 260: should we disable depthWrite? Seems more important to modify depthWrite than depthTest, since if it comes first in the render order, there's definitely going to be nothing in the depth buffer already.

jtbandes (approved):

packages/studio-base/src/panels/ThreeDeeRender/renderables/Images/ImageRenderable.ts
snosenzo: I wanted to avoid a boolean parameter to a function so I thought this would be a good compromise, but we only need one for now so I'll use `setRenderBehindScene()`

packages/studio-base/src/panels/ThreeDeeRender/renderables/Images/ImageRenderable.ts
snosenzo: yeah I think disabling both makes sense - updated

PR #5856, opened 2023-04-27
Add panel config version guard wrapper component

foxymiles:
**User-Facing Changes**
None

**Description**
Adds a new highestSupportedConfigVersion property to PanelExtensionAdapter and renders an upgrade warning instead of the panel if the config format for the panel is too new for the panel to read.

Screenshot 2023-04-27 at 10 53 42 AM


packages/studio-base/src/components/PanelConfigVersionGuard.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+  const { classes } = useStyles();
+
+  if (isVersionedPanelConfig(config) && config.version > highestSupportedVersion) {
+    return (
+      <Paper className={classes.root}>
+        <Card>
+          <CardContent>
+            <Typography variant="subtitle1">
+              This layout requires a more recent version of this panel.
```
snosenzo, line 50: Should we also tell them they can clear the layout for this panel to continue using it? Or do we generally want to discourage that?

packages/studio-base/src/panels/ThreeDeeRender/index.tsx
```diff
@@ -46,11 +47,13 @@ function ThreeDeeRenderAdapter(interfaceMode: InterfaceMode, props: Props) {
   );
 
   return (
-    <PanelExtensionAdapter
-      config={props.config}
-      saveConfig={props.saveConfig}
-      initPanel={boundInitPanel}
-    />
+    <PanelConfigVersionGuard highestSupportedVersion={1} config={props.config}>
```
defunctzombie, line 50: What's the thinking with this wrapper vs making this an option to `<PanelExtensionAdater>` and having it check the version? I feel like we want this for all our panels and don't need to go one-by-one wrapping them.

packages/studio-base/src/components/PanelConfigVersionGuard.tsx
foxymiles: I'm not sure. If it were a shared layout that would probably not be what they want. For their own layout maybe?

packages/studio-base/src/panels/ThreeDeeRender/index.tsx
foxymiles: I can certainly move it inside the `PanelExtensionAdater`. Although in theory these breaking upgrades should be rare.

defunctzombie: Rare yes but we'll want this flexibility for any panel so either we wrap them all or we do it with the extension adapter. Extension adapter seems like the common place for all of them?

packages/studio-base/src/components/PanelConfigVersionGuard.tsx
foxymiles: Good question. I guess if they're going to that much trouble we should allow it? At least they will have been warned.

packages/studio-base/src/panels/ThreeDeeRender/index.tsx
foxymiles: Ok I've added this to the extension adapter itself. I guess third party panels will have to handle this on their own for now.

packages/studio-base/src/components/PanelConfigVersionGuard.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { ReactNode } from "react";
+import { makeStyles } from "tss-react/mui";
+
+type VersionedPanelConfig = Record<string, unknown> & { version: number };
+
+function isVersionedPanelConfig(config: unknown): config is VersionedPanelConfig {
+  return (
+    config != undefined &&
+    typeof config === "object" &&
+    "version" in config &&
```
jtbandes, line 15: Maybe we should make this a constant somewhere, like `VERSION_CONFIG_KEY`?

Might we want to prefix it since it's being added automatically, something like `foxgloveConfigVersion`?

Prior art is we have `foxglovePanelTitle` here: https://github.com/foxglove/studio/blob/f5ee78ca96de7a0af1df6ba6516a82c4766a1e1a/packages/studio-base/src/util/layout.ts#L44

jtbandes (approved):

packages/studio-base/src/components/PanelExtensionAdapter/PanelExtensionAdapter.tsx
```diff
@@ -56,14 +57,24 @@ import { initRenderStateBuilder } from "./renderState";
...
 };
 
+type PanelExtensionAdapterProps = PanelExtensionAdapterContentProps & {
+  /**
+   * If defined, the highest supported version of config the panel supports.
+   * Used to prevent older implementations of a panel from trying to access
+   * newer, incompatible versions of the panel's config. Panels should include a
+   * numbered version property in their config to control this.
+   */
+  highestSupportedConfigVersion?: number;
```
defunctzombie, line 75: Why is this a separate prop type? Add this prop to the existing type?

packages/studio-base/src/components/PanelExtensionAdapter/PanelExtensionAdapter.tsx
```diff
@@ -78,7 +89,7 @@ type RenderFn = NonNullable<PanelExtensionContext["onRender"]>;
  *
  * The adapter creates a PanelExtensionContext and invokes initPanel using the context.
  */
-function PanelExtensionAdapter(props: PanelExtensionAdapterProps): JSX.Element {
+function PanelExtensionAdapterContent(props: PanelExtensionAdapterContentProps): JSX.Element {
```
defunctzombie, line 92: I don't like the habit we have of putting multiple components in the same file - especially with different names. If you want to turn this into a separate component that's fine but we should use a separate file.

packages/studio-base/src/components/PanelExtensionAdapter/PanelExtensionAdapter.tsx
```diff
@@ -576,4 +587,21 @@ function PanelExtensionAdapter(props: PanelExtensionAdapterProps): JSX.Element {
   );
 }
 
+function PanelExtensionAdapter(
+  props: PanelExtensionAdapterProps & { highestSupportedConfigVersion?: number },
+): JSX.Element {
+  const { highestSupportedConfigVersion, ...otherProps } = props;
+
+  return highestSupportedConfigVersion != undefined ? (
+    <PanelConfigVersionGuard
```
defunctzombie, line 596: Seems this logic could be simpler by allowing the version guard to accept undefined - then you don't need this ternary at the use site.

packages/studio-base/src/panels/ThreeDeeRender/index.tsx
```diff
@@ -48,6 +48,7 @@ function ThreeDeeRenderAdapter(interfaceMode: InterfaceMode, props: Props) {
   return (
     <PanelExtensionAdapter
       config={props.config}
+      highestSupportedConfigVersion={1}
```
defunctzombie, line 51: Why are we not setting this for all our panels?

packages/studio-base/src/components/PanelExtensionAdapter/PanelExtensionAdapter.tsx
foxymiles: I agree with putting components in their own file in general but separating small wrapper components out like this just creates a proliferation of useless files IMO.

packages/studio-base/src/panels/ThreeDeeRender/index.tsx
foxymiles: Was trying to keep the scope of this small and get it done quickly so we can get it in the next release. I can add it to all of them if you wan to do it now.

packages/studio-base/src/components/PanelExtensionAdapter/PanelExtensionAdapter.tsx
```diff
@@ -52,18 +52,29 @@ import {
...
   config: unknown;
   saveConfig: SaveConfig<unknown>;
 };
 
+type PanelExtensionAdapterProps = PanelExtensionAdapterContentProps & {
+  /**
+   * If defined, the highest supported version of config the panel supports.
+   * Used to prevent older implementations of a panel from trying to access
+   * newer, incompatible versions of the panel's config. Panels should include a
+   * numbered version property in their config to control this.
```
jtbandes, line 73: comment is maybe out of date now that we have VERSION_CONFIG_KEY?

snosenzo (approved):

jtbandes:
Closed in favor of #5906

PR #5857, opened 2023-04-27
Add Git LFS Pull to contributing instructions

amacneil:
**User-Facing Changes**
N/A

**Description**
Should not be necessary if LFS is properly configured, but running this step explicitly will flush out any problems.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,11 +6,11 @@
 
 - [Node.js](https://nodejs.org/en/) v16.10+
 - [Git LFS](https://git-lfs.github.com/)
-- [Visual Studio Code](https://code.visualstudio.com/) – Recommended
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -6,11 +6,11 @@
 
 - [Node.js](https://nodejs.org/en/) v16.10+
 - [Git LFS](https://git-lfs.github.com/)
-- [Visual Studio Code](https://code.visualstudio.com/) – Recommended
```
amacneil, line 9: Removed this because it is not a dependency

jtbandes: Maybe we should keep this but move it up into the "supported development environments" section?

jtbandes (approved):

PR #5858, opened 2023-04-27
Add support for UI feature tours

2metres:
**User-Facing Changes**
Add sidebar button to trigger a tour of the new UI.

**Description**
* Add id's needed for feature tour in hosted deployment
* Add new `featureTours` slice to the workspace context to track what feature tour, if any, is active and which have already been shown to the user.
* Add new `workspaceExtensions` item to the app context for injecting items into the workspace.
* Add new `sidebarItems` item to the app context for injecting items into the sidebar.


2metres:
Working pretty well so far except a few issues: 

- [ ] Will need to use the `Wrapper` prop on the TourProvider to render the tour in a portal so that I overlays the MUI ones
- [ ] When opening a menu the menu doesn't get the correct element until window resize
- [ ] How to target Layouts Menu when it's not in the studio-base.
- [ ] Reset the index when clicking out of the tour.

packages/studio-base/src/components/AppBar/AddPanelMenu.tsx
```diff
@@ -51,14 +51,16 @@ export function AddPanelMenu(props: AddPanelProps): JSX.Element {
...
     >
-      <PanelList
-        // Close when a drag starts so the modal menu doesn't block the drop targets
-        onDragStart={handleClose}
-        onPanelSelect={(selection) => {
-          addPanel(selection);
-          handleClose();
-        }}
-      />
+      <div data-tourid="add-panel-menu">
```
2metres, line 54: This was an attempt to select the correct element

packages/studio-base/src/components/AppBar/UserMenu.tsx
```diff
@@ -130,6 +137,11 @@ export function UserMenu({
         onClose={handleClose}
         onClick={handleClose}
         MenuListProps={{ className: classes.menuList, dense: true }}
+        PaperProps={
+          {
+            "data-tourid": "account-menu",
+          } as Partial<PaperProps & { "data-tourid"?: string }>
+        }
```
2metres, line 144: Open to a better way of doing this. MUI rejects data-attributes otherwise.

packages/studio-base/src/providers/WorkspaceContextProvider.tsx
```diff
@@ -44,8 +49,7 @@ function createWorkspaceContextStore(
       {
         name: "fox.workspace",
         partialize: (value) => {
-          const { dataSourceDialog: _, ...rest } = value;
-          return rest;
+          return omit(value, ["dataSourceDialog", "prefsDialogState"]);
```
jtbandes, line 52: Does this mean we will also persist `featureTours.active`? Does it make sense to persist an in-progress tour?

jtbandes (approved):

packages/studio-base/src/Workspace.tsx
```diff
@@ -671,6 +680,7 @@ function WorkspaceContent(props: WorkspaceContentProps): JSX.Element {
           </div>
         )}
       </div>
+      {!props.showSignInForm && workspaceExtensions}
```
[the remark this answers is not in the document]
foxymiles: The goal here was to avoid showing the tour on top of the sign in form. Might need to tweak this further.

packages/studio-base/src/providers/WorkspaceContextProvider.tsx
foxymiles: Yes. Probably it would be better to omit it but it didn't seem worth complicating this logic further.

packages/studio-base/src/context/AppContext.ts
```diff
@@ -6,13 +6,15 @@ import { createContext, useContext } from "react";
...
 interface IAppContext {
   appBarLayoutButton?: JSX.Element;
-  syncAdapters?: readonly JSX.Element[];
   createEvent?: (args: {
     deviceId: string;
     timestamp: string;
     durationNanos: string;
     metadata: Record<string, string>;
   }) => Promise<void>;
+  sidebarItems?: readonly [[string, { iconName: string; title: string }]];
```
foxymiles, line 15: I've defined this just as a string to avoid pulling dependencies on the Sidebar & Workspace in here. I filter out things that don't conform to the types in the Workspace.

packages/studio-base/src/Workspace.tsx
```diff
@@ -529,6 +535,12 @@ function WorkspaceContent(props: WorkspaceContentProps): JSX.Element {
         });
       }
 
+      for (const [key, item] of appContextSidebarItems ?? []) {
```
foxymiles, line 538: This tightens up the loose typing in the AppContext itself.

packages/studio-base/src/providers/WorkspaceContextProvider.tsx
defunctzombie: Can we have this logic flipped around and selectively decide what to serialize rather than having to remember to omit things?

packages/studio-base/src/context/WorkspaceContext.ts
```diff
@@ -15,17 +16,20 @@ import { useAppConfigurationValue } from "@foxglove/studio-base/hooks";
...
-  | "help"
-  | "layouts"
-  | "panel-settings"
-  | "app-settings"
-  | "studio-logs-settings"
-  | "variables";
+export const SidebarItemKeys = [
+  "account",
+  "add-panel",
+  "app-bar-tour",
```
defunctzombie, line 22: Why is this a sidebar item when the entire feature hides the sidebar?

foxymiles: This is necessary for the button that allows a user to manually trigger the tour from the old sidebar.

packages/studio-base/src/providers/WorkspaceContextProvider.tsx
foxymiles: Yeah we have enough stuff in here now that opt in is probably more prudent.

packages/studio-base/src/Workspace.tsx
```diff
@@ -529,6 +535,12 @@ function WorkspaceContent(props: WorkspaceContentProps): JSX.Element {
...
+        if (SidebarItemKeys.some((itemKey) => itemKey === key) && item.iconName in ICONS) {
```
defunctzombie, line 539: Lets use a typescript type asserting function on these checks - then you don't need the "as" casts below.

defunctzombie (approved):
Feels to me like we could have tried harder not to make featureTours a thing here and instead use the extension point.

I also don't like the logic to persist the workspace - feels like we are constantly re-jiggering it and we should be saying what we want to save rather than remembering to exclude things.

packages/studio-base/src/providers/WorkspaceContextProvider.tsx
```diff
@@ -49,7 +49,17 @@ function createWorkspaceContextStore(
       {
         name: "fox.workspace",
         partialize: (value) => {
-          return omit(value, ["dataSourceDialog", "prefsDialogState"]);
+          return pick(value, [
```
defunctzombie, line 52: 👍 Thank you!

I do think this needs comment making it clear that this is about which fields to persist. Every time I read this I see "partialize" and it tells me nothing about the purpose of this function in the app. It is not a good name IMO.

foxymiles: Added a comment. I'd like to do a followup to move all the various sidebar related data under a common `sidebar` slice.

esthersweon:
Not adding to changelog, as it currently depends on a studio-private PR.

PR #5860, opened 2023-04-27
Update picker to work in image mode

snosenzo:
**User-Facing Changes**

- none (experimental)

**Description**
Now in ImageMode the picker does not use a smaller render target. Previously the `setViewOffset` was overriding the manually set projection matrix in the camera. The setViewOffset was used to render only a subset of the view from the camera, which improves performance when picking. From observation it looks like this makes the `clickHandler` in ImageMode take two times longer when picking (up from ~50ms to ~100ms) for the same size windows. I think this difference will be negligible for users when picking. If it does become a problem we can revisit the manual setting of the camera matrix in image mode. 

FG-3096

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/Picker.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/Picker.ts
@@ -8,6 +8,7 @@
 // - jfaust https://github.com/jfaust
 
 import * as THREE from "three";
+import { assert } from "ts-essentials";
 
 import type { Renderable } from "./Renderable";
 
@@ -29,6 +30,9 @@ export type PickedRenderable = {
 
 export type PickerOptions = {
   debug?: boolean;
+  // Disable the setting of the projection matrix in the picking pass.
+  // Use this if you are setting the projection matrix of the camera manually
+  disableSetViewOffset?: boolean;
 };
 
 /**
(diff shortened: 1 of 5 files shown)


packages/studio-base/src/panels/ThreeDeeRender/Picker.ts
```diff
@@ -29,6 +30,9 @@ export type PickedRenderable = {
 
 export type PickerOptions = {
   debug?: boolean;
+  // Disable the setting of the projection matrix in the picking pass.
+  // Use this if you are setting the projection matrix of the camera manually
```
jtbandes, line 34: nit: `/**` comment

packages/studio-base/src/panels/ThreeDeeRender/Picker.ts
```diff
@@ -163,6 +148,65 @@ export class Picker {
...
+    y: number,
+    options: PickerOptions,
+  ): { xInView: number; yInView: number } {
+    assert(this.camera, "camera must be set before updating for pick");
+    const w = this.gl.domElement.width;
+    const h = this.gl.domElement.height;
+    const pixelRatio = this.gl.getPixelRatio();
+
+    if (options.disableSetViewOffset !== true) {
+      const hw = PIXEL_WIDTH / 2;
```
jtbandes, line 162: you lost the `|0` – PIXEL_WIDTH is set to `31` so this won't be an integer. Not sure if that matters?

jtbandes (approved):
Is there a way we can start to get some stories that cover picking? Like showing the hitmap that we render?

packages/studio-base/src/panels/ThreeDeeRender/Picker.ts
snosenzo: doesn't seem to affect, but I put it back to keep it consistent

jtbandes (approved):

PR #5861, opened 2023-04-27
fix for TextAnnotations in a message without any circles or points

jtbandes:
**User-Facing Changes**
Fixed an issue where an ImageAnnotations message with text but no circles or points would not be displayed.

**Description**


snosenzo (approved):
my bad, thanks for fixing this! 
Would be ideal if we could just return an empty array instead of having to account for undefined though, but won't block on that

jtbandes:
Yeah I wasn't totally sure why we had undefined as an allowed return value, but didn't want to go through a whole refactor to fix this.

PR #5869, opened 2023-04-28
Rework the TimeBasedChart downsample code organization

defunctzombie:
**User-Facing Changes**
None

**Description**
Rather than having two layers of throttle and debounce, this change moves uses one layer of throttle and a Downsampler class to manage the state for the downsample. useEffects update the Downsampler instance with the latest state information and the debounced callback triggers a downsample which updates the state variable with the downsampled dataset.


foxymiles (approved):
+1 on factoring this out. I think I'd like the implementation better as a hook than a class.

defunctzombie:
> I think I'd like the implementation better as a hook than a class

I've soured a bit on hooks for some of this stuff. I suppose it could be a hook but what would be the interface? Seems like all that would do is introduce refs and more overhead during each render.

PR #5872, opened 2023-04-28
Fix memory leak with accumulated condvar callbacks in UserNodePlayer

defunctzombie:
**User-Facing Changes**
Improve memory use for user scripts

**Description**
UserNodePlayer message handling uses a Condvar to signal when the node should terminate (maybe due to code changes in other nodes or itself). The message processing logic calls `.wait()` on this condvar for every message to process and races the returned signal against the RPC call to process. During normal runtime (with no changes to user scripts), nothing notifies all of these signals and they continue to pile up in the callback queue of the condvar.

This change updates the logic to create one signal during the node registration for use by the message processor callback. This signal is re-created on terminate.

### DIFF
--- a/packages/studio-base/src/players/UserNodePlayer/index.ts
+++ b/packages/studio-base/src/players/UserNodePlayer/index.ts
@@ -385,6 +385,10 @@ export default class UserNodePlayer implements Player {
 
       // Prune the node registration cache so it doesn't grow forever.
       // We add one to the count so we don't have to recompile nodes if users undo/redo node changes.
+
+      // fixme - but if a user removes a user node and the keys length decreases this randomly un-cached some other userNode?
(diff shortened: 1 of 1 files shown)


jtbandes (approved):
Any tests we can add for this? I suppose that might be more complex than necessary and involve weird WeakMap usage?

defunctzombie:
> Any tests we can add for this? I suppose that might be more complex than necessary and involve weird WeakMap usage?

Amusingly there was already a test that failed with my first attempt so I think the desired "behavior" has test coverage and this did not break the desired behavior.

jtbandes:
👍 I guess I meant new tests to prove that we are not leaking promises. But that might be beyond the realm of things we can easily test.

PR #5873, opened 2023-04-28
Fix urdf collada mesh color being overwritten

achim-k:
**User-Facing Changes**
- Fix Collada meshes being displayed with wrong color

**Description**
Collada mesh materials were ignored if the URDF visual had a an optional material specified. In Rviz, the optional material is only used if the mesh itself does not contain texture information. This PR makes sure that for collada files, embedded mesh materials are always used. To my knowledge, Collada is the only mesh file type used for URDFs which contains texture information.

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
@@ -725,8 +725,10 @@ function createRenderable(
       return new RenderableSphere(name, marker, undefined, renderer);
     }
     case "mesh": {
-      // Use embedded materials only when no override material is defined in the URDF
-      const embedded = !visual.material ? EmbeddedMaterialUsage.Use : EmbeddedMaterialUsage.Ignore;
+      const isCollada = visual.geometry.filename.toLowerCase().endsWith(".dae");
+      // Use embedded materials if the mesh is a Collada file or if no material is defined in the URDF
+      const embedded =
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
```diff
@@ -725,8 +725,10 @@ function createRenderable(
       return new RenderableSphere(name, marker, undefined, renderer);
     }
     case "mesh": {
-      // Use embedded materials only when no override material is defined in the URDF
-      const embedded = !visual.material ? EmbeddedMaterialUsage.Use : EmbeddedMaterialUsage.Ignore;
+      const isCollada = visual.geometry.filename.toLowerCase().endsWith(".dae");
+      // Use embedded materials if the mesh is a Collada file or if no material is defined in the URDF
+      const embedded =
+        isCollada || !visual.material ? EmbeddedMaterialUsage.Use : EmbeddedMaterialUsage.Ignore;
```
snosenzo, line 731: The only part of this that's a little unclear to me is: where is the embedded material if not in `visual.material`? 
Also I imagine this is fine if the collada doesn't have a material?

achim-k: > The only part of this that's a little unclear to me is: where is the embedded material if not in `visual.material`?

What do you mean? `visual.material` is typically only specified if the mesh itself contains no material. 

> Also I imagine this is fine if the collada doesn't have a material?

I believe so, yes.

jtbandes:
@achim-k is this ready to merge?

PR #5875, opened 2023-04-28
[3D Image] Add message converter support for 2D annotations

jtbandes:
**User-Facing Changes**
None

**Description**
Adds support for message converters (including multiple converters) for image annotations. If multiple converters are available, or a topic's original schema and converted schema are both supported, multiple entries for the topic will be shown in the settings. They can all be toggled individually.

https://user-images.githubusercontent.com/14237/235260068-a8d6f7d4-051d-4a18-a528-b64b53499531.mov

Relates to FG-1040 / #5791

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/annotations/ImageAnnotations.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/annotations/ImageAnnotations.ts
@@ -6,6 +6,7 @@ import { t } from "i18next";
 import { Immutable } from "immer";
 import * as THREE from "three";
 
+import { TwoKeyMap } from "@foxglove/den/collection";
 import { PinholeCameraModel } from "@foxglove/den/image";
 import { ImageAnnotations as FoxgloveImageAnnotations } from "@foxglove/schemas";
 import { MessageEvent, SettingsTreeAction, Topic } from "@foxglove/studio";
@@ -21,6 +22,9 @@ import { SettingsTreeEntry } from "../../../SettingsManager";
 import { IMAGE_ANNOTATIONS_DATATYPES } from "../../../foxglove";
 import { IMAGE_MARKER_ARRAY_DATATYPES, IMAGE_MARKER_DATATYPES } from "../../../ros";
 import { topicIsConvertibleToSchema } from "../../../topicIsConvertibleToSchema";
+
+type TopicName = string & { __brand: "TopicName" };
+type SchemaName = string & { __brand: "SchemaName" };
 
 interface ImageAnnotationsContext {
   initialScale: number;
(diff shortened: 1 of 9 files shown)


packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/annotations/ImageAnnotations.ts
```diff
@@ -37,19 +37,38 @@ interface ImageAnnotationsContext {
...
+
 /**
  * This class handles settings and rendering for ImageAnnotations/ImageMarkers.
  */
 export class ImageAnnotations extends THREE.Object3D {
   #context: ImageAnnotationsContext;
 
-  /** FG-3065: support multiple converters per message */
-  #renderablesByTopic = new Map<string, RenderableTopicAnnotations>();
+  #renderablesByTopicAndSchemaName = new Map<string, Map<string, RenderableTopicAnnotations>>();
```
snosenzo, line 71: can we use type alias' to make it clearer that this is `topic -> schema` so have something like `new Map<TopicName, Map<SchemaName, RenderableTopicAnnotations>>`?

snosenzo: I wonder if it would be worthwhile to make a `TopicSchemaMap` class for iterating through elemenst, `get` and `set` so that we don't have to do repetitious checking for the layers of nesting.

packages/studio-base/src/panels/ThreeDeeRender/stories/ImageMode/ImageAnnotations.stories.tsx
```diff
@@ -267,3 +273,187 @@ export const Annotations: StoryObj = {
...
+                ],
+                fill_color: { r: 0, g: 0, b: 0, a: 0 },
+                thickness: 2,
+              },
+            ],
+          }),
+        },
+        {
+          fromSchemaName: "MyCustomSchema",
+          toSchemaName: "foxglove_msgs/msg/ImageAnnotations",
```
snosenzo, line 395: what's the difference between this schema and the `foxglove_msgs/ImageAnnotations`?

snosenzo (approved):

packages/studio-base/src/panels/ThreeDeeRender/stories/ImageMode/ImageAnnotations.stories.tsx
jtbandes: ROS 1 vs ROS 2, but the schemas themselves are basically the same. I used them in the story in order to show the panel can support multiple converters at once.

packages/studio-base/src/providers/ExtensionCatalogProvider.tsx
```diff
@@ -95,6 +95,7 @@ function activateExtension(
 
 export function createExtensionRegistryStore(
   loaders: readonly ExtensionLoader[],
+  mockMessageConverters: readonly RegisterMessageConverterArgs<unknown>[] | undefined,
```
jtbandes, line 98: cc @defunctzombie in case you have any thoughts on how I injected message converters for the storybook.

snosenzo (approved):

PR #5879, opened 2023-04-30
Use ES private fields instead of TS private accessibility modifier

jtbandes:
**User-Facing Changes**
None

**Description**
Add a lint rule to enforce the use of ES [private class fields](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields) rather than the TypeScript `private` modifier. The new lint rule also includes auto-fix support, which renames the property/method definition and the accesses it can see.

In contrast with `private x`, which is a compile-time-only feature of TypeScript, `#x` is _truly_ private (cannot be accessed at runtime from outside the class using subscript notation) and cannot accidentally interfere with superclass properties.

One downside worth noting is that private fields can interfere with `Proxy` behavior: https://lea.verou.me/2023/04/private-fields-considered-harmful/ / https://github.com/tc39/proposal-class-fields/issues/106

I had to change a couple of `delete this._x;` to `this.#x = undefined;` and a couple of `const { _x } = this;` to `const x = this.#x;`.

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -28,6 +28,7 @@ settings:
 rules:
   "@foxglove/license-header": error
   "@foxglove/studio/link-target": error
+  "@foxglove/studio/prefer-hash-private": error
 
   tss-unused-classes/unused-classes: error
 
--- a/benchmark/src/BenchmarkStats.ts
+++ b/benchmark/src/BenchmarkStats.ts
@@ -17,18 +17,18 @@ type RecordFrameTimesFn = (frameTimes: Sample[]) => void;
 class BenchmarkStats {
   private static instance: BenchmarkStats | undefined;
 
-  private frameTimesMs: Sample[] = [];
+  #frameTimesMs: Sample[] = [];
 
   private constructor() {}
 
   public recordFrameTime(durationMs: number): void {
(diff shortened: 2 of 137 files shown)


defunctzombie (commented):
When viewing instances in Chrome Devtools debugger do we have visibility into private variables?

defunctzombie:
Seems fine. I welcome our new private class member world.

foxymiles (approved):

foxymiles:
At some point I'd like to do a followup pass and mark as many class members `readonly` as possible too.

jtbandes:
> When viewing instances in Chrome Devtools debugger do we have visibility into private variables?

Yes, you can play with it in the preview build.

It seems like the field names in the debugger are minified:

image


However, this seems in line with the actual change in generated source code (left is preview deploy, right is studio.foxglove.party):

image image

This seems like something we might be able to control by configuring and/or changing our minifier, if we deem it to be important.

snosenzo (approved):

jtbandes:
It occurred to me that the non-private identifiers in classes were probably being left alone because minifying them would have had an observable effect. That’s not true with the private ones.

PR #5882, opened 2023-05-01
Allow ImageMode without CameraInfo

snosenzo:
**User-Facing Changes**

(none)

**Description**
Give option for `None` in calibration topics list. When this is selected the renderer goes into an Image-only render mode. This unsubscribes all 3D subscriptions (including transforms) and only keeps subscriptions for ImageMode (annotations,  and images).  While this mode is active there will be an error next to the calibration topic select to notify the user that no 3D topics or transforms will be available in this mode. 

- [x] Fix applying default camera model to annotations
- [x] create story to test imageonly subscription
- [x] fix race condition when seeking during playback where image throws GL error and only shows as black
- [x] ~Guarantee a camera model has the P matrix and can't return undefined~(going to do in another PR)

FG-2950


packages/studio-base/src/panels/ThreeDeeRender/renderables/Images/ImageRenderable.ts
```diff
@@ -91,6 +91,31 @@ export class ImageRenderable extends Renderable<ImageUserData> {
...
+    this.userData.material = undefined;
+    this.userData.geometry?.dispose();
+    this.userData.geometry = undefined;
+    this.userData.mesh = undefined;
+
+    this.userData.cameraInfo = undefined;
+    this.userData.cameraModel = undefined;
+    this.userData.image = undefined;
+
+    this.#bitmap = undefined;
```
jtbandes: do we need .close()?

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -614,6 +650,27 @@ export class Renderer extends EventEmitter<RendererEvents> implements IRenderer
...
+    this.#clearSubscriptions();
+    this.#addSubscriptionsFromSceneExtensions(
+      (extension) => extension === this.#imageModeExtension,
+    );
+  };
+
+  #disableImageOnlySubscriptionMode = (): void => {
+    this.#addSubscriptionsFromSceneExtensions(
+      (extension) => extension !== this.#imageModeExtension,
+    );
```
jtbandes, line 668: Hm, I think I would've expected to see `clear(); #addSubscriptionsFromSceneExtensions(()=>true)`, is there a reason to do it this way?

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageMode.ts
```diff
@@ -89,7 +111,28 @@ export class ImageMode extends SceneExtension<ImageRenderable> implements ICamer
...
+      config: () => renderer.config.imageMode,
+      updateConfig: (updateHandler) => {
+        renderer.updateConfig((draft) => updateHandler(draft.imageMode));
+      },
+      updateSettingsTree: () => {
+        this.updateSettingsTree();
+      },
+      addSchemaSubscriptions: (schemaNames, handler) => {
+        renderer.addSchemaSubscriptions(schemaNames, handler);
+      },
```
jtbandes, line 129: merge conflict fyi: I'm adding `labelPool` to this context in #5897

packages/studio-base/src/panels/ThreeDeeRender/SceneExtension.ts
```diff
@@ -80,6 +80,13 @@ export class SceneExtension<
     this.renderables.clear();
   }
 
+  /**
+   * Will add subscriptions from this scene extension to the renderer
+   * This will be called by the renderer when building topic and schema subscriptions on
+   * initialization and when imageOnlyMode becomes enabled
+   */
+  public addSubscriptionsToRenderer(): void {}
```
jtbandes, line 88: I think this method shouldn't mutate anything on renderer, but should return a list of subscriptions. Then the renderer can add them as it sees fit.
Something more like this:

```suggestion
  type Subscription = { type: "topic", topic: string, subscription: RendererSubscription<T> | MessageHandler<T> } | { type: "schemas", schemaNames: Iterable<string>, subscription: RendererSubscription<T> | MessageHandler<T> };
  public getSubscriptions(): Subscription[] {}
```

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageMode.ts
```diff
@@ -70,12 +79,25 @@ export class ImageMode extends SceneExtension<ImageRenderable> implements ICamer
...
    * @param canvasSize Canvas size in CSS pixels
    */
-  public constructor(renderer: IRenderer, canvasSize: THREE.Vector2) {
+  public constructor(
+    renderer: IRenderer,
+    canvasSize: THREE.Vector2,
+    {
+      enableImageOnlyMode,
+      disableImageOnlyMode,
+    }: { enableImageOnlyMode: () => void; disableImageOnlyMode: () => void },
```
jtbandes, line 94: In terms of the API of the ImageMode class, I think it'd be better to call these something like `setHasCameraCalibrationTopic(_: boolean)`. The "image only mode" is a concept that belongs to Renderer and doesn't need to leak into the scene extension.

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageMode.ts
```diff
@@ -549,3 +652,46 @@ export class ImageMode extends SceneExtension<ImageRenderable> implements ICamer
...
+  const cx = width / 2;
+  const cy = height / 2;
+  const fx = focalLength;
+  const fy = focalLength;
+  const cameraInfo = normalizeCameraInfo({
+    header: { seq: 0, stamp: { sec: 0, nsec: 0 }, frame_id: frameId },
+    height,
+    width,
+    distortion_model: "",
+    R: [1, 0, 0, 1, 0, 0, 1, 0, 0],
```
jtbandes, line 681:
```suggestion
    R: [1, 0, 0, 0, 1, 0, 0, 0, 1],
```

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageModeCamera.ts
```diff
@@ -24,6 +24,10 @@ export class ImageModeCamera extends THREE.PerspectiveCamera {
     this.#updateProjection();
   }
 
+  public clearModel(): void {
+    this.#model = undefined;
+  }
```
jtbandes, line 29: maybe just make updateCamera take `PinholeCameraModel | undefined`?

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageMode.ts
```diff
@@ -369,9 +432,53 @@ export class ImageMode extends SceneExtension<ImageRenderable> implements ICamer
...
+      .then((maybeBitmap) => {
+        const prevRenderable = renderable;
+        const currentRenderable = this.#imageRenderable;
+        if (currentRenderable !== prevRenderable) {
+          return;
+        }
+        this.renderer.settings.errors.removeFromTopic(topic, CREATE_BITMAP_ERR_KEY);
+        if (maybeBitmap instanceof ImageBitmap) {
+          renderable.setBitmap(maybeBitmap);
+        }
```
jtbandes, line 450: I don't get what we're doing in the case where it's not an ImageBitmap? Is the image still rendered somehow?

packages/studio-base/src/panels/ThreeDeeRender/stories/ImageMode/ImageAnnotations.stories.tsx
```diff
@@ -457,3 +459,13 @@ export const MessageConverterSupport: StoryObj = {
...
+
+export const Annotations: StoryObj = {
+  parameters: { colorScheme: "light" },
+  render: AnnotationsStory(),
+};
+
+export const ImageOnlyIsSameWithNo3d: StoryObj = {
+  parameters: { colorScheme: "light" },
+  render: AnnotationsStory({ calibrationTopic: UNSELECTED_CAMERA_CALIBRATION }),
+};
```
jtbandes, line 471: I think you can remove a level of nesting and use `render: AnnotationsStory`, then use `args: { calibrationTopic: ... }` to change the function args.

Let's make sure we also get a story with an _invalid_ CameraModel (e.g. invalid focal lengths or something that causes it to fail to load). You can pass an `includeSettings` prop to PanelSetup to make it show the settings UI, which could be good for testing that we display errors in the right place

packages/studio-base/src/panels/ThreeDeeRender/renderables/Images/ImageRenderable.ts
snosenzo: I believe it's taken care of in `texture.dispose()`, but I 've removed this function since regardless

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
snosenzo: no particular reason outside of avoiding unecessary clear of the image mode renderables.

packages/studio-base/src/panels/ThreeDeeRender/SceneExtension.ts
snosenzo: I know, I'm not happy with it either. that's what I originally wanted to do, but couldn't figure out how to safely type the Subscriptions without casting all of the handlers' messageEvent inputs to `unknown`

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageMode.ts
snosenzo: good point, and it's fine to have the boolean function in this case since it's only a single param and pretty clear how it's used?

packages/studio-base/src/panels/ThreeDeeRender/stories/ImageMode/ImageAnnotations.stories.tsx
snosenzo: I have a branch going to refactor some of the `PinholeCameraModel` methods after this and make it impossible to make one without a P matrix. I think an invalid PinholeCamera story would be a better fit with those changes. Does that sound reasonable?

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageMode.ts
snosenzo: `setImage` is called previously for Raw images. But yeah it's a little confusing I thought it would be better to treat the images more consistently through the decode function, but it doesn't make sense to send the data images through an async call. So I've updated it to only put compressed images through `decode`.

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageMode.ts
```diff
@@ -435,40 +430,57 @@ export class ImageMode extends SceneExtension<ImageRenderable> implements ICamer
...
+            `Error creating bitmap: ${err.message}`,
+          );
+        });
+    } else {
+      // Raw Images don't need to be decoded asynchronously
+      if (this.#fallbackCameraModelActive()) {
+        this.#updateFallbackCameraModel(image, getFrameIdFromImage(image));
+      }
+      renderable.update();
+    }
```
jtbandes, line 470: maybe flip this around and early return to make it easier to read?

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageMode.ts
```diff
@@ -79,24 +79,19 @@ export class ImageMode extends SceneExtension<ImageRenderable> implements ICamer
 
   #imageRenderable: ImageRenderable | undefined;
 
-  #enableImageOnly: () => void;
-  #disableImageOnly: () => void;
+  #setHasCameraInfo: ({ val }: { val: boolean }) => void;
```
jtbandes, line 82: `val` isn't a descriptive name for this variable - maybe `hasCameraInfo`? Or I think it's fine to keep it as a boolean parameter (not an object) add an eslint-disable to silence that warning.

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -325,10 +325,17 @@ export class Renderer extends EventEmitter<RendererEvents> implements IRenderer
...
+        this.#imageModeExtension = new ImageMode(
+          this,
+          this.input.canvasSize,
+          ({ val: hasCameraInfo }) => {
+            if (hasCameraInfo) {
+              this.#disableImageOnlySubscriptionMode();
+            } else {
+              this.#enableImageOnlySubscriptionMode();
+            }
+          },
```
jtbandes, line 337: feels like maybe this stuff should all move into an options object – it's not clear what the function is meant to be used for

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -614,6 +657,27 @@ export class Renderer extends EventEmitter<RendererEvents> implements IRenderer
...
+      "Image mode extension should be defined when calling enable Image only mode",
+    );
+    this.clear({ clearTransforms: true, resetAllFramesCursor: true });
+    this.#clearSubscriptions();
+    this.#addSubscriptionsFromSceneExtensions(
+      (extension) => extension === this.#imageModeExtension,
+    );
+  };
+
+  #disableImageOnlySubscriptionMode = (): void => {
```
jtbandes, line 672: will something go wrong if one of these functions is called twice in a row?

packages/studio-base/src/panels/ThreeDeeRender/SceneExtension.ts
jtbandes: Ah yeah, I mean realistically, if we want them in an array, TS doesn't have the ability to provide _separate_ strong types for all the items individually. I think we probably have this issue already with the renderer's internal storage of the list of subscriptions.

I'll file a separate followup ticket for myself to come back and look at alternatives to this after it merges

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -373,6 +364,15 @@ export class Renderer extends EventEmitter<RendererEvents> implements IRenderer
     this.#addSceneExtension(new VelodyneScans(this));
     this.#addSceneExtension(this.measurementTool);
     this.#addSceneExtension(this.publishClickTool);
+    if (
+      interfaceMode === "image" &&
+      config.imageMode.calibrationTopic === UNSELECTED_CAMERA_CALIBRATION
```
jtbandes, line 369: Can we use `undefined` for this so we don't leak the string `"UNSELECTED_CAMERA_CALIBRATION"` into the panel config? I believe the settings tree editor is set up to correctly handle undefined as a possible value.

jtbandes (commented):
Thanks for the changes, this is looking good. I think we are close but added a few more requested tweaks:

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
snosenzo: Currently the `add(X)Subscriptions` method on the renderer doesn't actually prevent duplicate handlers from being added. So in the `disableImageOnlyMode`it would create duplicate handlers and cause messages using those handlers to be handled twice. I've updated both of these to clear subscriptions and the renderer before performing the state change to ensure this doesn't happen.

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageMode.ts
```diff
@@ -70,12 +79,20 @@ export class ImageMode extends SceneExtension<ImageRenderable> implements ICamer
...
+
   /**
    * @param canvasSize Canvas size in CSS pixels
    */
-  public constructor(renderer: IRenderer, canvasSize: THREE.Vector2) {
+  public constructor(
+    renderer: IRenderer,
+    canvasSize: THREE.Vector2,
+    setHasCameraInfo: ({ val }: { val: boolean }) => void,
```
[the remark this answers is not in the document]
snosenzo: Yeah right now `hasCameraInfoTopic` would be more accurate, not sure what would happen if this was set while waiting for a camera info.

packages/studio-base/src/panels/ThreeDeeRender/renderables/Images.ts
```diff
@@ -274,8 +274,34 @@ export class Images extends SceneExtension<ImageRenderable> {
...
+          }
+          this.renderer.settings.errors.removeFromTopic(imageTopic, CREATE_BITMAP_ERR_KEY);
+          if (maybeBitmap instanceof ImageBitmap) {
+            renderable.setBitmap(maybeBitmap);
+          }
+          renderable.update();
+          this.renderer.queueAnimationFrame();
+        })
+        .catch((err) => {
+          this.renderer.settings.errors.addToTopic(
```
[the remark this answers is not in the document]
snosenzo: added it to `ImageMode` too

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -373,6 +364,12 @@ export class Renderer extends EventEmitter<RendererEvents> implements IRenderer
     this.#addSceneExtension(new VelodyneScans(this));
     this.#addSceneExtension(this.measurementTool);
     this.#addSceneExtension(this.publishClickTool);
+    if (interfaceMode === "image" && config.imageMode.calibrationTopic == undefined) {
+      this.#enableImageOnlySubscriptionMode();
+    } else {
+      this.#addTransformSubscriptions();
+      this.#addSubscriptionsFromSceneExtensions();
```
jtbandes, line 371: can this be `this.#disableImageOnlySubscriptionMode()`?

packages/studio-base/src/panels/ThreeDeeRender/stories/ImageMode/imageCommon.ts
```diff
@@ -0,0 +1,73 @@
...
+    receiveTime: { sec: 0, nsec: 0 },
+    message: {
+      timestamp: { sec: 0, nsec: 0 },
+      frame_id: frameId,
+      height,
+      width,
+      distortion_model: "rational_polynomial",
+      D: [],
+      K: [fx, 0, cx, 0, fy, cy, 0, 0, 1],
+      R: [1, 0, 0, 1, 0, 0, 1, 0, 0],
```
jtbandes, line 35:
```suggestion
      R: [1, 0, 0, 0, 1, 0, 0, 0, 1],
```

packages/studio-base/src/panels/ThreeDeeRender/stories/ImageMode/ImageAnnotations.stories.tsx
```diff
@@ -19,303 +16,262 @@ export default {
...
-                  settings: { visible: true },
-                },
-              ],
-            },
-          }}
-        />
-      </PanelSetup>
-    );
-  },
+export const ImageOnlyModeIsSameWithNo3d: StoryObj = {
```
jtbandes, line 271:
```suggestion
export const AnnotationsWithoutCalibration: StoryObj = {
```

jtbandes (approved):

packages/studio-base/src/panels/ThreeDeeRender/stories/ImageMode/imageCommon.ts
snosenzo: I'll update the rest of these in my camera model refactor:
image

PR #5883, opened 2023-05-01
Problems list cleanup

2metres:
**User-Facing Changes**
None

**Description**
Fixes missing key error and makes storybook errors consistent

Closes FG-3113 and FG-3120


defunctzombie (approved):

jtbandes (approved):

packages/studio-base/src/components/DataSourceSidebar/DataSourceSidebar.stories.tsx
```diff
@@ -135,6 +135,21 @@ export const PlayerIntializingJapanese: StoryObj = {
   parameters: { forceLanguage: "ja" },
 };
 
+const fakeError = (): Error =>
+  Object.assign(new Error("Fake Error"), {
+    stack: `Error: Original Error
+    at Story (https://603ec8bf7908b500231841e2-nozcuvybhv.capture.chromatic.com/studio-base-src-components-DataSourceSidebar-DataSourceSidebar-stories.f1dd4357.iframe.bundle.js:233:28)
```
jtbandes, line 141: We really do not need these to match the current strings in the snapshots. It's okay to accept a screenshot change one more time to make the strings whatever we want. We don't even need them to show any of this chromatic crap, we can just make up our own strings 😄  (but leaving it this way is fine too).

2metres: @jtbandes Just give me a realistic error message and I'll accept it

PR #5892, opened 2023-05-02
Stamp time of the user's first run of the app into the user profile

foxymiles:
**User-Facing Changes**
None

**Description**
Adds `firstSeenTime` and `firstSeenTimeIsFirstLoad` fields to `UserProfile` and sets them on app load. We treat the presence of a `currentLayoutId` in the UserProfile as a sign that this isn't the first time the user has loaded the app.


defunctzombie (commented):
I'd like us to not spread individual local storage access around the app and instead use a limited set of places that are reading/writing to stores.

I'm meh us having hooks like `useTimestampUserFirstSeen` doing direct storage access with the `WORKSPACE_CONTEXT_LOCAL_STORAGE_KEY` when this feels like something the workspace should be doing.

packages/studio-web/src/Root.tsx
```diff
@@ -28,6 +29,8 @@ export function Root(props: {
   extraProviders: JSX.Element[] | undefined;
   dataSources: IDataSourceFactory[] | undefined;
 }): JSX.Element {
+  useTimestampUserFirstSeen();
```
defunctzombie, line 32: This pattern seems weird to me - its like we want this hook simply for its side effect? I don't know why we need a hook for that?

defunctzombie:
Overall I feel like we could make a PR that better aligns us with the patterns we want to be using for this type of storage.

Related to the desired goal here - don't we already have storage for a user profile or possibly a layout? Somehow I feel like we already have something we store.

foxymiles:
> I'd like us to not spread individual local storage access around the app and instead use a limited set of places that are reading/writing to stores.
> 
> I'm meh us having hooks like `useTimestampUserFirstSeen` doing direct storage access with the `WORKSPACE_CONTEXT_LOCAL_STORAGE_KEY` when this feels like something the workspace should be doing.

What makes this tricky is it needs to run before all that other stuff has had a chance to touch local storage so it can guess if this is the first time the user has loaded the app. Open to ideas for other approaches here.

packages/studio-web/src/Root.tsx
foxymiles: How else would you like to see this implemented? It needs to remember it's run and not run again on every render.

foxymiles:
I guess we could also use the presence of a `currentLayoutId` in the user profile as an indicator it's not their first run.

foxymiles:
I've switched approaches here to store these values in the `UserProfile` and use the presence of a `currentLayoutId` in the existing profile as a sign that this is not the user's first session in the application.

packages/studio-base/src/context/UserProfileStorageContext.ts
```diff
@@ -9,6 +9,16 @@ import { LayoutID } from "@foxglove/studio-base/context/CurrentLayoutContext";
...
   currentLayoutId?: LayoutID;
+
+  /** Timestamp of the first time the user loaded the app. */
+  firstSeenTime?: string;
+
+  /**
+   * True if the at the time we assigned firstSeenTime it appeared to be the
+   * user's first load of the app.
+   */
+  firstSeenTimeIsFirstLoad?: boolean;
```
snosenzo, line 20: can this just be `isFirstLoad`, and it's set to `true` if undefined and false otherwise?

snosenzo (approved):

packages/studio-base/src/providers/UserProfileLocalStorageProvider.tsx
```diff
@@ -33,10 +34,23 @@ export default function UserProfileLocalStorageProvider({
...
   );
+
+  // On first load stamp firstSeenTime timestamp. We consider the time at which
+  // we stamp firstTime as the first time the user has opened the app if at that
+  // time there is no currentLayoutId already set in the profile.
+  useEffect(() => {
+    setUserProfile((old) => ({
+      ...old,
+      firstSeenTime: old.firstSeenTime ?? new Date().toISOString(),
+      firstSeenTimeIsFirstLoad: old.firstSeenTimeIsFirstLoad ?? old.currentLayoutId == undefined,
```
defunctzombie, line 45: Am I reading this correctly that this will keep `firstSeenTimeIsFirstLoad` to true once it gets set to true once?

packages/studio-base/src/providers/UserProfileLocalStorageProvider.tsx
```diff
@@ -33,10 +34,23 @@ export default function UserProfileLocalStorageProvider({
...
+    })).catch(console.error);
```
defunctzombie, line 46: Do we want to send this to sentry? It is anomalous for this to occur right?

packages/studio-base/src/providers/UserProfileLocalStorageProvider.tsx
defunctzombie: ok - I understand - this exists because we want to determine if we've already seen folks by checking whether there is a currentLayoutId

packages/studio-base/src/providers/UserProfileLocalStorageProvider.tsx
foxymiles: It happens every time the user loads the app for the very first time.

PR #5894, opened 2023-05-02
Remove clickable data source name

2metres:
**User-Facing Changes**
None (behind feature-flag)

**Description**
Removes clickable title from App Bar

Resolves FG-3179 and FG-3167

### DIFF
--- a/packages/studio-base/src/Workspace.tsx
+++ b/packages/studio-base/src/Workspace.tsx
@@ -595,15 +595,8 @@ function WorkspaceContent(props: WorkspaceContentProps): JSX.Element {
       },
       "[": () => setLeftSidebarOpen((oldValue) => !oldValue),
       "]": () => setRightSidebarOpen((oldValue) => !oldValue),
-      o: () => dataSourceDialogActions.open("start"),
     };
-  }, [
-    dataSourceDialogActions,
-    sidebarItem,
-    selectSidebarItem,
-    setLeftSidebarOpen,
-    setRightSidebarOpen,
-  ]);
+  }, [sidebarItem, selectSidebarItem, setLeftSidebarOpen, setRightSidebarOpen]);
 
   const play = useMessagePipeline(selectPlay);
   const playUntil = useMessagePipeline(selectPlayUntil);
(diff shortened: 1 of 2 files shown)


jtbandes:
The preview deploy is currently unusable once you select a data source since there's no way to get back to the open dialog. Does this need to be merged/rebased to show the app menu?

2metres:
Rebasing now

jtbandes:
Incidentally I feel like the spinner is too large:

image

jtbandes (approved):
Do we still need/want the max width?

packages/studio-base/src/Workspace.tsx
```diff
@@ -595,8 +595,15 @@ function WorkspaceContent(props: WorkspaceContentProps): JSX.Element {
       },
       "[": () => setLeftSidebarOpen((oldValue) => !oldValue),
       "]": () => setRightSidebarOpen((oldValue) => !oldValue),
+      o: () => dataSourceDialogActions.open("start"),
```
jtbandes, line 598: Should this be cmd/ctrl + O so it matches the normal "open" action in most apps?

packages/studio-base/src/components/AppBar/DataSource.tsx
```diff
@@ -22,39 +22,36 @@ import WssErrorModal from "@foxglove/studio-base/components/WssErrorModal";
 import { useWorkspaceActions } from "@foxglove/studio-base/context/WorkspaceContext";
 import { PlayerPresence } from "@foxglove/studio-base/players/types";
 
-const LEFT_ICON_SIZE = 19;
+const ICON_SIZE = 17;
```
jtbandes, line 25: can we use the theme instead of hard-coding?

packages/studio-base/src/Workspace.tsx
2metres: How do I do this with the KeyHandler?

2metres: Removing this for another PR

packages/studio-base/src/components/AppBar/DataSource.tsx
2metres: Not really…

PR #5895, opened 2023-05-02
Rename split panel text for sanity

amacneil:
**User-Facing Changes**
Rename:
- `Split horizontal` -> `Split down`
- `Split vertical` -> `Split right`

Despite years of computering, I am incapable of remembering the difference between vertical/horizontal splits in my head. I constantly click `Split vertical` when I want vertically stacked panels, but I end up with horizontally stacked panels.

I even got these backwards when I tried to make this PR, despite having fallen for it 30 seconds earlier in the UI.

No more! We shall follow VSCode into the glorious new world of `Split right` and `Split down`. Maybe one day we can even add `Split up` and `Split left`.

image


2metres (approved):
Makes sense to me

2metres:
Can we rename Fullscreen to Maximize while you're in there?

jtbandes:
I always just look at the icons so the words don't confuse me. No strong opinion.

PR #5896, opened 2023-05-02
Support for user defined custom Http headers

ruffsl:
**User-Facing Changes**

Adds an additional input field for users to specify custom Http headers to include for the websocket connection, such as headers including authentication tokens to access websocket backends behind reverse proxy servers that enforce access control restrictions for secure and private port forwarding, such as GitHub Codespaces.

-  https://docs.github.com/en/codespaces/developing-in-codespaces/forwarding-ports-in-your-codespace#using-command-line-tools-and-rest-clients-to-access-ports

**Description**

This adds a new `headers` input field in formConfig for FoxgloveWebSocketDataSourceFactory that accepts Http headers, as key value string pairs encoded in json, to include in the websocket connection request. After input validation, these headers are then passed along with the user specified websocket URL down to where they can be used in the call sight for the creation of the websocket connection.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,22 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/typescript-node
+{
+	"name": "Node.js & TypeScript",
+	// Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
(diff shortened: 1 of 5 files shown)


ruffsl:
I opened this draft PR for posterity, but upon later investigation I think the reuse of additional reverse proxies for websocket backends (that are behind corporate operated reverse proxies for session authentication) is the better, albeit more complex, solution. This is opposed to modifying the header request sent by the websocket client, given that doing so doesn't seem to be supported by the web standard, common libraries, nor supported web browsers. See this PR for example workaround:

- https://github.com/ros-planning/navigation2/pull/3551

PR #5898, opened 2023-05-02
Triangle list: use DynamicBufferGeometry

jtbandes:
**User-Facing Changes**
Fixed an issue where ROS Markers using the `TRIANGLE_LIST` type would fail to render when the number of triangles changed.

**Description**
Use DynamicBufferGeometry to handle resizing (similar to [RenderableTriangles](https://github.com/foxglove/studio/blob/main/packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableTriangles.ts))


snosenzo (approved):
Tested and looks to be working. What exactly was the root of this issue? using float32 for colors? not resizing the normals? looks like it still should've been resizing with the old code at first glace

jtbandes:
I believe it's because of one or both of the drawRange or attribute.count settings that DynamicBufferGeometry does. If you read the docs and forums, it sounds like BufferGeometry is really not meant to be resized. Initially I wrote a version of this where I just created a new geometry each time it needed to be resized, but then remembered we'd already got it working with DynamicBufferGeometry.

PR #5906, opened 2023-05-03
Panel version guard

defunctzombie:
**User-Facing Changes**
None

**Description**
In the future we are able to bump the panel configuration version. This will cause older apps to show a message informing the user their panel does not support the new version and they need to upgrade.

Based on: https://github.com/foxglove/studio/pull/5856


jtbandes (approved):

packages/studio-base/src/i18n/en/panelConfigVersionGuard.ts
```diff
@@ -0,0 +1,8 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+export const panelConfigVersionGuard = {
+  instructions: "Please update to the latest version.",
```
jtbandes, line 6:
```suggestion
  instructions: "Please update to the latest version of the app.",
```
Applying suggestion from @banisadr on chromatic.

PR #5913, opened 2023-05-04
Add text anchor point setting to image panel

snosenzo:
**User-Facing Changes**

 - Add option in image panel to show text anchor points 

**Description**
Add option in image panel to show text anchor points in the bottom left of the annotation using a circle of radius 2 with an outline of 1 using the background text color as the fill and the text color as the outline.
Example from story:
image

FG-2994

### DIFF
--- a/packages/studio-base/src/panels/Image/ImageView.tsx
+++ b/packages/studio-base/src/panels/Image/ImageView.tsx
@@ -114,7 +114,7 @@ export function ImageView({ context }: Props): JSX.Element {
     };
   });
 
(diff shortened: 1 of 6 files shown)


jtbandes (approved):
🤷  Doesn't seem terribly useful to me, but a pretty simple feature.

amacneil:
Launch review consensus: 👎 unclear customer need

snosenzo:
can reopen if there's more desire, but closing for now since it's not a big add.

PR #5914, opened 2023-05-04
Use EmptyState for the empty states in the sidebar

2metres:
**User-Facing Changes**
Make sidebar empty states consistent

### DIFF
--- a/packages/studio-base/src/components/DataSourceSidebar/TopicList.tsx
+++ b/packages/studio-base/src/components/DataSourceSidebar/TopicList.tsx
@@ -263,7 +263,7 @@ export function TopicList(): JSX.Element {
         <EmptyState>
           {playerPresence === PlayerPresence.PRESENT && filterText
             ? `No topics or datatypes matching \n “${filterText}”`
-            : "No topics available"}
+            : "No topics available. "}
           {playerPresence === PlayerPresence.RECONNECTING && "Waiting for connection"}
         </EmptyState>
       )}
--- a/packages/studio-base/src/components/DataSourceSidebar/TopicList.tsx
+++ b/packages/studio-base/src/components/DataSourceSidebar/TopicList.tsx
@@ -182,27 +182,12 @@ export function TopicList(): JSX.Element {
     [filterText, topics],
   );
 
-  const noTopicsMessage = useMemo(() => {
-    switch (playerPresence) {
-      case PlayerPresence.PRESENT:
-        return filterText
-          ? `No topics or datatypes matching \n “${filterText}”`
-          : "No topics available";
-      case PlayerPresence.ERROR:
-        return "An error occurred";
-      case PlayerPresence.NOT_PRESENT:
-        return "No data source selected";
-      case PlayerPresence.RECONNECTING:
-        return "No topics available \n Waiting for connection";
-      case PlayerPresence.BUFFERING:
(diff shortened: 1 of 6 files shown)


jtbandes:
Topics empty state is still bold while the others aren't.

2metres:
image

jtbandes (changes requested):
Not there yet. Topics empty state is still bold while the others aren't.

image

image

image

packages/studio-base/src/components/DataSourceSidebar/TopicList.tsx
```diff
@@ -181,24 +182,27 @@ export function TopicList(): JSX.Element {
...
+          ? `No topics or datatypes matching \n “${filterText}”`
+          : "No topics available";
+      case PlayerPresence.ERROR:
+        return "An error occurred";
+      case PlayerPresence.NOT_PRESENT:
+        return "No data source selected";
+      case PlayerPresence.RECONNECTING:
+        return "No topics available \n Waiting for connection";
+      case PlayerPresence.BUFFERING:
+        return "Loading topics";
```
jtbandes, line 198: Can we avoid changing the strings/logic in this PR? Off the top of my head I don't think BUFFERING should result in an EmptyState being shown, since it can happen even once topics are already known.

2metres: I think the current logic is wrong though. If you filter the text your end up with: 
Screen Shot 2023-05-05 at 3 08 03 pm

2metres: I can remove the BUFFERING string as it's not used

packages/studio-base/src/panels/Image/components/ImageEmptyState.tsx
```diff
@@ -3,46 +3,51 @@
...
+    return <EmptyState className={classes.root}>Select a topic to view images</EmptyState>;
   }
   return (
-    <Stack fullHeight fullWidth justifyContent="center" alignItems="center" position="absolute">
-      <EmptyState>
-        <Typography variant="body2" color="text.secondary" gutterBottom>
-          Waiting for images {markerTopics.length > 0 && "and markers"} on:
+    <EmptyState className={classes.root}>
+      <Typography variant="inherit" gutterBottom>
+        Waiting for images{markerTopics.length > 0 && "and markers"} on:
```
jtbandes, line 36: 😬 
```suggestion
        Waiting for images {markerTopics.length > 0 && "and markers"} on:
```

jtbandes (approved):

packages/studio-base/src/components/EmptyState.tsx
```diff
@@ -2,18 +2,15 @@
...
+
 const useStyles = makeStyles()((theme) => ({
   root: {
-    backgroundColor: theme.palette.background.default,
-    display: "flex",
-    flexDirection: "column",
-    flex: "auto",
-    alignItems: "center",
-    justifyContent: "center",
+    whiteSpace: "pre-wrap",
```
jtbandes, line 13:
```suggestion
    whiteSpace: "pre-line",
```

esthersweon:
Also think the copy should be consistent across all states (all complete sentences with a period, or all phrases without a period).

PR #5915, opened 2023-05-04
[3D Image] Add pan & zoom support

jtbandes:
**User-Facing Changes**
None

**Description**
Adds pan & zoom support to new Image panel.
Also adds a "Reset view" button that can be used to reset manually adjusted pan/zoom.

Moved RendererOverlay to its own file.

https://user-images.githubusercontent.com/14237/236349003-ca92b246-1b4e-4ac4-a5c5-50dda4881a4b.mov

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/stories/ImageMode/ImageMode.stories.tsx
+++ b/packages/studio-base/src/panels/ThreeDeeRender/stories/ImageMode/ImageMode.stories.tsx
@@ -335,14 +335,23 @@ export const ImageModeResizeHandled: StoryObj = {
   },
 };
 
-export const ImageModePanAndZoom: StoryObj = {
+export const ImageModePan: StoryObj = {
   render: () => <ImageModeFoxgloveImage imageType="raw" />,
   play: async () => {
     const canvas = document.querySelector("canvas")!;
     fireEvent.mouseDown(canvas, { clientX: 200, clientY: 200 });
     fireEvent.mouseMove(canvas, { clientX: 400, clientY: 200 });
     fireEvent.mouseUp(canvas, { clientX: 400, clientY: 200 });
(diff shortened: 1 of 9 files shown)


amacneil:
👍 Love it

Styling should be consistent with plot panel:
image

Imo we could explore an icon instead of text for the button (for both panels). But that could be a separate project. @2metres for thoughts.

jtbandes:
I tried styling it the same way. It did not work because the button was so transparent that the text became unreadable when the image or other stuff was underneath it.

packages/studio-base/src/panels/ThreeDeeRender/stories/ImageMode/ImageMode.stories.tsx
```diff
@@ -339,9 +335,27 @@ export const ImageModeResizeHandled: StoryObj = {
...
 };
 
+export const ImageModePanAndZoom: StoryObj = {
+  render: () => <ImageModeFoxgloveImage imageType="raw" />,
+  play: async () => {
+    const canvas = document.querySelector("canvas")!;
+    fireEvent.mouseDown(canvas, { clientX: 200, clientY: 200 });
+    fireEvent.mouseMove(canvas, { clientX: 400, clientY: 200 });
+    fireEvent.mouseUp(canvas, { clientX: 400, clientY: 200 });
+    fireEvent.wheel(canvas, { deltaY: 100, clientX: 200, clientY: 0 });
```
snosenzo, line 345: I feel like zooming in would be more obvious for this story. 
Or just update the story name to say `ZoomOut`, I just expected it to be zoomed in.
Maybe also add an event that moves after zooming to make sure that the scaled pan is working appropriately?

snosenzo (approved):
Got me wondering if we should better encapsulate the view state (including camera) and control handling in both 3D and image modes, this looks good for now though.

PR #5919, opened 2023-05-05
Offset fullscreen panel for app bar

foxymiles:
**User-Facing Changes**
None (feature flagged)

**Description**
Offset position of fullscreen panel by app bar height so it doesn't cover the app bar.

### DIFF
--- a/packages/studio-base/src/components/PanelRoot.tsx
+++ b/packages/studio-base/src/components/PanelRoot.tsx
@@ -8,6 +8,7 @@ import { TransitionStatus } from "react-transition-group";
 import { makeStyles } from "tss-react/mui";
 
 import { AppSetting } from "@foxglove/studio-base/AppSetting";
+import { APP_BAR_HEIGHT } from "@foxglove/studio-base/components/AppBar/constants";
 import { useAppConfigurationValue } from "@foxglove/studio-base/hooks";
 
 export const PANEL_ROOT_CLASS_NAME = "FoxglovePanelRoot-root";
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/components/PanelRoot.tsx
```diff
@@ -67,7 +67,7 @@ export const usePanelRootStyles = makeStyles<
     entered: {
       borderWidth: 4,
       position: "fixed",
-      top: 0,
+      top: sourceRect?.top ?? 0,
```
jtbandes, line 70: I don't think this is right, sourceRect should be where the panel is normally located on screen.

jtbandes: https://user-images.githubusercontent.com/14237/236363369-3b50348e-9350-4df2-bfaa-bd245c614805.mov

foxymiles: Yeah you're right that's only correct for a panel at the top of the layout.

foxymiles: Changed this to use the app config value explicitly. Unfortunately this meant hardcoding the height of the app bar but we're already also doing that for the playback controls.

jtbandes: I think we have a constant for it somewhere. We use it for the electron desktop window too.

foxymiles: Yep found it.

jtbandes (approved):

PR #5924, opened 2023-05-05
Upgrade monaco-editor and monaco-editor-webpack-plugin

jtbandes:
**User-Facing Changes**
None

**Description**
- Update the way we access the ICodeEditorService. In https://github.com/microsoft/vscode/commit/bac119e88e7615eeef39a62da4c72f4ddf1b0f60 the statically-exported services were replaced with a "registration" method, so we now use `StandaloneServices.get(ICodeEditorService)`.
- Update the way we are overriding the "jump to definition" functionality. In https://github.com/microsoft/vscode/pull/154352 they introduced a "`registerCodeEditorOpenHandler()`" API that seems a bit more future-proof than `doOpenEditor`, although it works the same way.
- Fixed a bug in our jump to definition functionality. The code comment said "If we are jumping to a definition within the user node, don't push to script override." However in practice, this logic was not working because the file path it was checking for didn't match the file paths we use when creating the model.

Manually tested the following functionality:
- Autocomplete `Input<"">`
- Autocomplete `Message<"">`
- Jump to definition of `Input` (a separate file)
- Jump to a definition within the same file

https://user-images.githubusercontent.com/14237/236585377-b1799ea1-4c52-48a5-83d5-4494f7529a60.mov


packages/studio-base/src/panels/NodePlayground/Editor.tsx
```diff
@@ -111,40 +113,52 @@ const Editor = ({
...
+          return editor;
+        }
+
+        // If we are jumping to a definition within the user node, don't push
+        // to script override.
+        if (
+          editor &&
+          script &&
+          requestedModel.uri.path ===
+            `${DEFAULT_STUDIO_NODE_PREFIX}${path.basename(script.filePath)}`
```
jtbandes, line 143: this path was wrong

packages/studio-base/src/panels/NodePlayground/Editor.tsx
```diff
@@ -30,9 +32,9 @@ import { mightActuallyBePartial } from "@foxglove/studio-base/util/mightActually
 
 import { themes } from "./theme";
 
-const codeEditorService = StaticServices.codeEditorService.get();
+const codeEditorService = StandaloneServices.get(ICodeEditorService);
 
-type CodeEditor = monacoApi.editor.IStandaloneCodeEditor;
+type CodeEditor = monacoApi.editor.ICodeEditor;
```
jtbandes, line 37: not necessary but seemed like perhaps more accurate for interoperating with some other types below?

packages/studio-base/src/panels/NodePlayground/Editor.tsx
```diff
@@ -265,7 +279,7 @@ const Editor = ({
     if (model && script && !script.readOnly) {
       // We have to use a ref for autoFormatOnSaveRef because of how monaco scopes the action callbacks
       if (autoFormatOnSaveRef.current) {
-        await editorRef.current?.getAction("editor.action.formatDocument").run();
+        await editorRef.current?.getAction("editor.action.formatDocument")?.run();
```
jtbandes, line 282: type changed, I guess

packages/studio-base/src/panels/NodePlayground/Editor.tsx
```diff
@@ -111,40 +113,52 @@ const Editor = ({
...
-      const requestedModel = monacoApi.editor.getModel(input.resource);
-      if (!requestedModel) {
-        return editor;
-      }
-
-      // If we are jumping to a definition within the user node, don't push
-      // to script override.
-      if (requestedModel.uri.path === script?.filePath) {
+  useEffect(() => {
+    const disposable = codeEditorService.registerCodeEditorOpenHandler(
```
jtbandes, line 123: new API

packages/studio-base/src/panels/NodePlayground/Editor.tsx
```diff
@@ -30,9 +32,9 @@ import { mightActuallyBePartial } from "@foxglove/studio-base/util/mightActually
 
 import { themes } from "./theme";
 
-const codeEditorService = StaticServices.codeEditorService.get();
+const codeEditorService = StandaloneServices.get(ICodeEditorService);
```
jtbandes, line 35: new API

defunctzombie (approved):
amazing

PR #5936, opened 2023-05-08
Support global variables in plot panel series paths

foxymiles:
**User-Facing Changes**
Support global variables in message paths.

**Description**
Support global variables in message paths. This is a POC limited to the plot panel only. I haven't done anything to make the message path input aware of this yet but you can enter paths with variables and plot them.

I've taken this path instead of changing anything in the message pipeline because there's no way to do this without having some awareness of a mapping in the panel so I decided to localize it to the panel, in a way we could generalize to other panels without too much difficulty.

I'm posting this in this state so we can play with it and see how we feel about the UX.

Screenshot 2023-05-08 at 8 04 45 AM


foxymiles:
One more thing to keep in mind here is that variables can also hold JSON or numbers so we need to figure out what to do in those cases.

foxymiles:
Closing in favor of https://github.com/foxglove/studio/pull/6080

PR #5939, opened 2023-05-08
Organize workspace context items into slices

foxymiles:
**User-Facing Changes**
None

**Description**
Group workspace context items by function for better organization.


packages/studio-base/src/context/Workspace/WorkspaceContext.ts
```diff
@@ -0,0 +1,94 @@
...
+    );
+  },
+};
+
+/**
+ * Fetches values from the workspace store.
+ */
+export function useWorkspaceStore<T>(
+  selector: (store: WorkspaceContextStore) => T,
+  equalityFn?: (a: T, b: T) => boolean,
```
jtbandes, line 90: what's the use case for a caller of useWorkspaceStore wanting to customize the equalityFn?

packages/studio-base/src/context/Workspace/migrations.ts
```diff
@@ -0,0 +1,85 @@
...
+  sidebarItem: undefined | SidebarItemKey;
+};
+
+export function migrateV0WorkspaceState(
+  oldState: unknown,
+  _version: number,
+): WorkspaceContextStore {
+  // Currently v0 is the only obsolete state. If we do more migrations this
+  // needs to consider the version number.
+  const v0State = oldState as WorkspaceContextStoreV0;
```
jtbandes, line 48: should we use DeepPartial here to make sure we don't crash when migrating? and/or does the `migrate` functionality handle exceptions?

packages/studio-base/src/context/Workspace/useWorkspaceActions.ts
```diff
@@ -0,0 +1,261 @@
...
+  );
+  const [initialEnableNewTopNav] = useState(currentEnableNewTopNav);
+  const enableNewTopNav = isDesktopApp() ? initialEnableNewTopNav : currentEnableNewTopNav;
+
+  const set = useCallback(
+    (setter: (draft: Draft<WorkspaceContextStore>) => void) => {
+      setState(produce<WorkspaceContextStore>(setter));
+    },
+    [setState],
+  );
```
jtbandes, line 98: is this the new recommended pattern, to separate the actions from the store? It seems like the zustand examples often make the actions part of the store. I haven't seen this pattern of exposing `setState` before.

jtbandes (approved):

packages/studio-base/src/context/Workspace/WorkspaceContext.ts
foxymiles: Probably the main one is to use a `shallow` comparison.

packages/studio-base/src/context/Workspace/migrations.ts
foxymiles: I found out the hard way during testing that an exception here will mean `persist` just loads the default state. It doesn't crash the app though.

packages/studio-base/src/context/Workspace/useWorkspaceActions.ts
foxymiles: Zustand isn't opinionated on this point so you can store the actions in the store or define them separately. The biggest advantage I've found of storing them separately is in a codebase like ours where actions often need to reference state from other contexts it keeps the store cleaner since that extra state doesn't need to be pulled into the store.

PR #5940, opened 2023-05-08
Alert that restart is required to revert UI on desktop

foxymiles:
**User-Facing Changes**
None (feature flagged)

**Description**
Show alert notifying the user that an app restart is required to complete reverting to the old UI.

### DIFF
--- a/packages/studio-base/src/components/AppBar/UserMenu.tsx
+++ b/packages/studio-base/src/components/AppBar/UserMenu.tsx
@@ -2,14 +2,7 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
-import {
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/components/AppBar/UserMenu.tsx
```diff
@@ -125,6 +121,19 @@ export function UserMenu({
...
   }, [analytics, currentUserType]);
 
+  const revertToOldUI = useCallback(async () => {
+    if (isDesktop) {
+      await confirm({
+        title: "Please restart the app to finish reverting to the old UI.",
+        ok: "OK",
+        cancel: false,
+      });
+      await setEnableNewTopNav(false);
```
defunctzombie, line 131: nit: you can have this line shared outside of the _if_ statement since it is the same for both cases

PR #5941, opened 2023-05-08
[Image Beta] auto-select image and calibration topics

jtbandes:
**User-Facing Changes**
None (beta)

**Description**
- Auto-select an image topic by default when none is selected (resolves FG-3224)
- Auto-select a calibration topic when changing image topics (resolves FG-3236)
- Sort matching calibration topics to the top of the list, like we do for annotation topics


packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageMode.ts
```diff
@@ -231,21 +239,57 @@ export class ImageMode
     super.removeAllRenderables();
   }
 
+  /** If no image topic is selected, automatically select the first available one from `renderer.topics` */
+  public autoSelectImageTopic(): void {
```
snosenzo, line 243: Is this code the same as in `Images`? should we pull these out into dedicated functions shared between the two that take a list of topics and an input topic and return the autoselected topic? We could then make tests for them

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -814,6 +814,8 @@ export class Renderer extends EventEmitter<RendererEvents> implements IRenderer
       for (const extension of this.sceneExtensions.values()) {
         this.settings.setNodesForKey(extension.extensionId, extension.settingsNodes());
       }
+
+      this.#imageModeExtension?.autoSelectImageTopic();
```
snosenzo, line 818: We weren't doing this previously for the `Images` scene extension? It seems like an unnecessary one-off. can we keep it contained to the extension?

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageMode.ts
```diff
@@ -231,21 +239,57 @@ export class ImageMode
...
+    if (this.#getImageModeSettings().imageTopic != undefined) {
+      return;
+    }
+
+    for (const topic of this.renderer.topics ?? []) {
+      if (topicIsConvertibleToSchema(topic, ALL_SUPPORTED_IMAGE_SCHEMAS)) {
+        this.renderer.updateConfig((draft) => {
```
snosenzo, line 250: is there any tradeoff for calling update config multiple times vs once with both potential updates?

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageMode.ts
```diff
@@ -254,18 +298,25 @@ export class ImageMode
...
-          )
-        ) {
+        if (!topicIsConvertibleToSchema(topic, ALL_SUPPORTED_CALIBRATION_SCHEMAS)) {
           return;
         }
         return { label: topic.name, value: topic.name };
       },
     );
 
+    // Sort calibration topics with prefixes matching the image topic to the top.
```
snosenzo, line 308: This would probably read better as `sortTopicsToMatchPrefix` or something along those lines

snosenzo (approved):

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
jtbandes: The extension needs to know when the topics have changed. Do you have suggestions for how else to trigger this happening? I didn’t want to put the side effect in the settingsNodes() function…

snosenzo: Maybe this is overkill, but it might not be a bad idea to add a `topicsChanged` method to the `SceneExtensions` API? `Images` would be able to use it to

jtbandes: done in https://github.com/foxglove/studio/pull/5941/commits/f0797b2d6bb8f442c1b02078c20542d968c749a5

snosenzo (approved):

PR #5942, opened 2023-05-08
Support math modifiers on settings panel message path inputs

foxymiles:
**User-Facing Changes**
Support math modifiers on settings panel message path inputs

**Description**
Support math modifiers on settings panel message path inputs. Adds a `supportsMathModifiers` field to the message path settings input type.


snosenzo (approved):
Can we get a story for this? or is there already one?

foxymiles:
> Can we get a story for this? or is there already one?

Good idea. I just updated one of the existing stories to use a math modifier.

jtbandes (approved):

PR #5951, opened 2023-05-09
Handle incremental subscriptions in message converters

foxymiles:
**User-Facing Changes**
Fixes a bug with message converters and extension panels.

**Description**
Fixes incremental updates to subscribed topics and message converters by extracting the calculation of the active conversions into a separate, memoized function and using changes in the results of that function to trigger currentFrame  and allFrames updates.

The strategy is as follows:
1. If we receive a new `currentFrame` or `allFrames`, emit all converted and unconverted messages.
2. If we receive a new set of converters, emit a new `allFrames` containing all converted and unconverted messages, as well as a s new `currentFrame` containing only the results of running new converters on the last message received on all topics.

### DIFF
--- a/packages/studio-base/src/components/PanelExtensionAdapter/messageProcessing.ts
+++ b/packages/studio-base/src/components/PanelExtensionAdapter/messageProcessing.ts
@@ -49,7 +49,7 @@ export function convertMessage(
 }
 
 /**
- * Returns a new map consisting of all items in a not present in b.
+ * Returns a new map consisting of all items in `a` not present in `b`.
  */
 export function mapDifference<K, V>(a: Map<K, V[]>, b: undefined | Map<K, V[]>): Map<K, V[]> {
   const result = new Map<K, V[]>();
--- a/packages/studio-base/src/components/PanelExtensionAdapter/messageProcessing.ts
+++ b/packages/studio-base/src/components/PanelExtensionAdapter/messageProcessing.ts
@@ -1,0 +1,199 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { difference } from "lodash";
+import { DeepReadonly } from "ts-essentials";
+
+import { MessageEvent, RegisterMessageConverterArgs, Subscription } from "@foxglove/studio";
+import { Topic as PlayerTopic } from "@foxglove/studio-base/players/types";
+
+// Branded string to ensure that users go through the `converterKey` function to compute a lookup key
+type Brand<K, T> = K & { __brand: T };
+type ConverterKey = Brand<string, "ConverterKey">;
+
+type MessageConverter = RegisterMessageConverterArgs<unknown>;
+
+type TopicSchemaConverterMap = Map<ConverterKey, MessageConverter[]>;
+
+// Create a string lookup key from a message event
+//
+// The string key uses a newline delimeter to avoid producting the same key for topic/schema name
+// values that might concatenate to the same string. i.e. "topic" "schema" and "topics" "chema".
+function converterKey(topic: string, schema: string): ConverterKey {
+  return (topic + "\n" + schema) as ConverterKey;
+}
+
+/**
+ * Convert message into convertedMessages using the keyed converters. Modifies
+ * convertedMessages in place for efficiency.
+ */
+export function convertMessage(
+  messageEvent: DeepReadonly<MessageEvent<unknown>>,
+  converters: DeepReadonly<TopicSchemaConverterMap>,
+  convertedMessages: MessageEvent<unknown>[],
+): void {
+  const key = converterKey(messageEvent.topic, messageEvent.schemaName);
+  const matchedConverters = converters.get(key);
+  for (const converter of matchedConverters ?? []) {
+    const convertedMessage = converter.converter(messageEvent.message);
+    convertedMessages.push({
+      topic: messageEvent.topic,
+      schemaName: converter.toSchemaName,
+      receiveTime: messageEvent.receiveTime,
+      message: convertedMessage,
+      originalMessageEvent: messageEvent,
+      sizeInBytes: messageEvent.sizeInBytes,
+    });
+  }
+}
+
+/**
+ * Returns a new map consisting of all items in a not present in b.
+ */
+export function mapDifference<K, V>(a: Map<K, V[]>, b: undefined | Map<K, V[]>): Map<K, V[]> {
+  const result = new Map<K, V[]>();
+  for (const [key, value] of a.entries()) {
+    const newValues = difference(value, b?.get(key) ?? []);
+    if (newValues.length > 0) {
+      result.set(key, newValues);
+    }
+  }
+  return result;
+}
+
+export type TopicSchemaConversions = {
+  // Topics which we are subscribed without a conversion, these are topics we
+  // want to receive the original message.
+  unconvertedSubscriptionTopics: Set<string>;
+
+  // When a subscription with a convertTo exists, we use this map to lookup a
+  // converter which can produce the desired output message schema. The keys for
+  // the map are `topic + input schema`.
+  //
+  // This allows the runtime message event handler logic which builds
+  // currentFrame and allFrames to lookup whether the incoming message event has
+  // converters to run by looking up the topic + schema of the message event in
+  // this map.
+  topicSchemaConverters: TopicSchemaConverterMap;
+};
+
+/**
+ * Builds a set of topics we can render without conversion and a map of
+ * converterKey -> converter arguments we use to produce converted messages.
+ *
+ * This will be memoized for performance so the inputs should be stable.
+ */
+export function collateTopicSchemaConversions(
+  subscriptions: readonly Subscription[],
+  sortedTopics: readonly PlayerTopic[],
+  messageConverters: undefined | readonly MessageConverter[],
+): TopicSchemaConversions {
+  const topicSchemaConverters: TopicSchemaConverterMap = new Map();
+  const unconvertedSubscriptionTopics: Set<string> = new Set();
+
+  // Bin the subscriptions into two sets: those which want a conversion and those that do not.
+  //
+  // For the subscriptions that want a conversion, if the topic schemaName matches the requested
+  // convertTo, then we don't need to do a conversion.
+  for (const subscription of subscriptions) {
+    if (!subscription.convertTo) {
+      unconvertedSubscriptionTopics.add(subscription.topic);
+      continue;
+    }
+
+    // If the convertTo is the same as the original schema for the topic then we don't need to
+    // perform a conversion.
+    const noConversion = sortedTopics.find(
+      (topic) => topic.name === subscription.topic && topic.schemaName === subscription.convertTo,
+    );
+    if (noConversion) {
+      unconvertedSubscriptionTopics.add(noConversion.name);
+      continue;
+    }
+
+    // Since we don't have an existing topic with out destination schema we need to find
+    // a converter that will convert from the topic to the desired schema
+    const subscriberTopic = sortedTopics.find((topic) => topic.name === subscription.topic);
+    if (!subscriberTopic) {
+      continue;
+    }
+
+    const key = converterKey(subscription.topic, subscriberTopic.schemaName ?? "<no-schema>");
+    let existingConverters = topicSchemaConverters.get(key);
+
+    // We've already stored a converter for this topic to convertTo
+    const haveConverter = existingConverters?.find(
+      (conv) => conv.toSchemaName === subscription.convertTo,
+    );
+    if (haveConverter) {
+      continue;
+    }
+
+    // Find a converter that can go from the original topic schema to the target schema
+    // Note: We only support one converter per unique from/to pair so this _find_ only needs to
+    //       find one converter rather than multiple converters.
+    const converter = messageConverters?.find(
+      (conv) =>
+        conv.fromSchemaName === subscriberTopic.schemaName &&
+        conv.toSchemaName === subscription.convertTo,
+    );
+
+    if (converter) {
+      existingConverters ??= [];
+      existingConverters.push(converter);
+      topicSchemaConverters.set(key, existingConverters);
+    }
+  }
+
+  return { unconvertedSubscriptionTopics, topicSchemaConverters };
+}
+
+/**
+ * Function to iterate and call function over multiple sorted arrays in sorted order across all items in all arrays.
+ * Time complexity is O(t*n) where t is the number of arrays and n is the total number of items in all arrays.
+ * Space complexity is O(t) where t is the number of arrays.
+ * @param arrays - sorted arrays to iterate over
+ * @param compareFn - function called to compare items in arrays. Returns a positive value if left is larger than right,
+ *  a negative value if right is larger than left, or zero if both are equal
+ * @param forEach - callback to be executed on all items in the arrays to iterate over in sorted order across all arrays
+ */
+export function forEachSortedArrays<Item>(
+  arrays: Item[][],
+  compareFn: (a: Item, b: Item) => number,
+  forEach: (item: Item) => void,
+): void {
+  const cursors: number[] = Array(arrays.length).fill(0);
+  if (arrays.length === 0) {
+    return;
+  }
+  for (;;) {
+    let minCursorIndex = undefined;
+    for (let i = 0; i < cursors.length; i++) {
+      const cursor = cursors[i]!;
+      const array = arrays[i]!;
+      if (cursor >= array.length) {
+        continue;
+      }
+      const item = array[cursor]!;
+      if (minCursorIndex == undefined) {
+        minCursorIndex = i;
+      } else {
+        const minItem = arrays[minCursorIndex]![cursors[minCursorIndex]!]!;
+        if (compareFn(item, minItem) < 0) {
+          minCursorIndex = i;
+        }
+      }
+    }
+    if (minCursorIndex == undefined) {
+      break;
+    }
+    const minItem = arrays[minCursorIndex]![cursors[minCursorIndex]!];
+    if (minItem != undefined) {
+      forEach(minItem);
+      cursors[minCursorIndex]++;
+    } else {
+      break;
+    }
+  }
+}
--- a/packages/studio-base/src/components/PanelExtensionAdapter/renderState.test.ts
+++ b/packages/studio-base/src/components/PanelExtensionAdapter/renderState.test.ts
@@ -52,7 +52,10 @@ function makeInitialState(): BuilderRenderStateInput {
     globalVariables: {},
     hoverValue: undefined,
     sharedPanelState: {},
-    sortedTopics: [{ name: "test", schemaName: "schema" }],
+    sortedTopics: [
+      { name: "test", schemaName: "schema" },
+      { name: "test2", schemaName: "schema2" },
+    ],
     subscriptions: [{ topic: "test" }, { topic: "test", convertTo: "otherSchema", preload: true }],
     messageConverters: [
       {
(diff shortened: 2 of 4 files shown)


packages/studio-base/src/components/PanelExtensionAdapter/renderState.test.ts
```diff
@@ -694,9 +694,240 @@ describe("renderState", () => {
...
+    expect(state2).toEqual({
+      topics: [
+        {
+          name: "test",
+          schemaName: "schema",
+          datatype: "schema",
+          convertibleTo: ["otherSchema", "anotherSchema"],
+        },
+      ],
+      currentFrame: [
```
defunctzombie, line 845: I don't think it is correct to produce the same message into `currentFrame` again without a seek or reset of some sort. I think the contract we have today is that current frame does not repeat data during playback.

packages/studio-base/src/components/PanelExtensionAdapter/renderState.ts
```diff
@@ -49,6 +51,95 @@ function converterKey(topic: string, schema: string): ConverterKey {
...
+      existingConverters ??= [];
+      existingConverters.push(converter);
+      topicSchemaConverters.set(key, existingConverters);
+    }
+  }
+
+  return { unconvertedSubscriptionTopics, topicSchemaConverters };
+}
+
+const memoCollateTopicSchemaConversions = memoizeWeak(collateTopicSchemaConversions);
```
defunctzombie, line 141: I am not a fan of `memoizeWeak` somehow I constantly feel like it hides important properties of referential equality (often with no comment) but typically the idea on memoization is to avoid an expensive calculation. memoization is the right idea but not clear in the ways we are often using it.

defunctzombie:
> @defunctzombie @jtbandes We should do this for `allFrames` too right?

Yep. `allFrames` should be using the same collated value.

packages/studio-base/src/components/PanelExtensionAdapter/renderState.ts
defunctzombie: Another gotcha with these memoizations - unless I am miss-reading - this is going to make a single  instance of `memoCollateTopicSchemaConversions` that is shared by all the callers of this function. This would make multiple extension panels potentially collide if the inputs were the same? Though probably not depending on how equality is checked by `memoizeWeak`. One thing it won't do well is release unused conversions when the extension panel is removed - or is there a facility for that.

foxymiles: What's the alternative?

packages/studio-base/src/components/PanelExtensionAdapter/renderState.test.ts
foxymiles: Yeah I see what you mean. I think the test isn't really set up correctly for this either but I guess we do want to make sure we only emit messages that originate from the changed active converter right?

packages/studio-base/src/components/PanelExtensionAdapter/renderState.ts
foxymiles: Don't we want those results to be shared? If the inputs are the same shouldn't the outputs be the same?

foxymiles: The nice thing about making this a pure function of the inputs is we don't need to worry about shared users of the function stomping on each other.

packages/studio-base/src/components/PanelExtensionAdapter/renderState.test.ts
foxymiles: Ok I think I'm now handling this correctly and only emitting new messages from changed converters.

packages/studio-base/src/components/PanelExtensionAdapter/renderState.ts
defunctzombie: There's no cleanup of this memoization - we build up global state and never release it. Granted this is a small amount of state but this is something I dislike about this approach to memoization. It is not hooked into any lifecycles.

foxymiles: It's all based on weak maps so it will be tossed if we're under memory pressure but we could lift the memoized version into the state builder closure if we wanted it to be scoped to panel lifetime.

foxymiles: I've switched these over to the `moize` library and lifted them into the render state builder closure so they are scoped to the life of the panel. Moize also by default only caches one hit so these will only use as much memory as it takes to cache one set of arguments.

https://github.com/planttheidea/moize

foxymiles: Generally speaking I think moize looks like a faster, better maintained and more featureful alternative to memoize-weak.

packages/studio-base/src/components/PanelExtensionAdapter/renderState.test.ts
```diff
@@ -694,9 +761,208 @@ describe("renderState", () => {
...
+          },
+          receiveTime: { nsec: 0, sec: 1 },
+          schemaName: "anotherSchema",
+          sizeInBytes: 1,
+          topic: "test",
+        },
+      ],
+    });
+  });
+
```
snosenzo, line 956: can message converters be removed? Should we have a test for that?

snosenzo (commented):
Sorry I missed reviewing this earlier. This breaks assumptions in `allFramesCursor` in the `Renderer` because when a message converter is added there could be new messages before the cursor location causing it to not add those messages and also re-add messages that it's already added. We'll need to either reset the cursor when converters are added or have it remember the last processed message and use referential equality to look back in `allFrames` until it reaches its last read message. Though if message converters can be removed then look-back won't be able to happen because the message won't be there - in which case it will go back to the beginning of allFrames.

packages/studio-base/src/components/PanelExtensionAdapter/renderState.ts
```diff
@@ -254,50 +187,43 @@ function initRenderStateBuilder(): BuildRenderStateFn {
...
-                  schemaName: converter.toSchemaName,
-                  receiveTime: messageEvent.receiveTime,
-                  message: convertedMessage,
-                  originalMessageEvent: messageEvent,
-                  sizeInBytes: messageEvent.sizeInBytes,
-                });
-              }
-            }
+      if (currentFrame) {
+        // If we have a new frame, emit that frame and process all messages on
```
defunctzombie, line 191: What if the currentFrame is the same as the previous frame? Previously, there was logic to prevent that from processing the currentFrame again.

defunctzombie:
> This breaks assumptions in `allFramesCursor` in the `Renderer` because when a message converter is added there could be new messages before the cursor location causing it to not add those messages and also re-add messages that it's already added.

@snosenzo I wonder if this means the cursor implementation is too fragile? I had imagined that the cursor detects if the thing it was pointing to in the allFrames array was not longer the same thing and invalidates itself in that case.

defunctzombie:
@foxymiles Is the gist of this change (as it relates to currentFrame), that it will only provide newly converted messages into currentFrame if new converters are activated? So any message already present in currentFrame will be ignored?

snosenzo:
> I wonder if this means the cursor implementation is too fragile? I had imagined that the cursor detects if the thing it was pointing to in the allFrames array was not longer the same thing and invalidates itself in that case.

I agree that it is fragile.  It does document the assumptions it makes [here](https://github.com/foxglove/studio/blob/main/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts#L515-L522). When we were using a more complex block eviction and loading method the cursor accounted for this, but since we updated our preloading strategy I trimmed it out since we no longer needed to account for it. I do realize now that  these assumptions for allFrames should probably be documented in `renderState` itself.

packages/studio-base/src/components/PanelExtensionAdapter/renderState.ts
foxymiles: I can add that back. I thought part of our player contract was that currentFrame shouldn't be present unless it's a new frame but it's easy enough to check for that.

foxymiles:
> @foxymiles Is the gist of this change (as it relates to currentFrame), that it will only provide newly converted messages into currentFrame if new converters are activated? So any message already present in currentFrame will be ignored?

The goal here is if we have a new currentFrame then we emit all those messages plus any converted messages. If converters change but we don't have a new currentFrame then we run only the converters that are new since the last currentFrame on the perviousFrame. Are our players consistent in only providing a non-undefined currentFrame if there are new messages or will they pass through the old currentFrame if something else in activeData has changed?

foxymiles:
@defunctzombie @snosenzo I've tweaked the current frame handling logic a bit:
1. If we have a new currentFrame, emit all messages + converted messages
2. If we don't have a new currentFrame but we do have different converters, run the converters on either the currentFrame if we have it or the previous frame if not and emit that.
3. If currentFrame has changed and is now undefined, emit an empty state

packages/studio-base/src/components/PanelExtensionAdapter/renderState.ts
foxymiles: Actually I think it's good enough to just hoist the memoized versions into the closure. We can consider adopting moize or something like it separately.

foxymiles:
Sounds like @snosenzo is going to handle the changes to the allFrames cursor in 3d panel in a followup.

foxymiles:
@snosenzo Thinking more about the expectations around `allFrames`. I think if converters change I should emit new blocks including all messages, both converted and unconverted right? I think our clients of allFrames expect blocks to be complete and not only contain new messages in each block.

foxymiles:
> @foxymiles Not sure what you're referring to when saying "emit new blocks", as panel extensions don't have a notion of blocks.

To put it another way, I think we have different expectations for currentFrame vs allFrames. I'm trying here to avoid sending duplicate messages in currentFrame. If the converters change I will only send new messages from the newly enabled converters in currentFrame. For allFrames, however, I think we want to include all messages, not only those that would result from newly enabled converters. I know plot panel, for example, expects allFrames to contain messages on all subscribed topics.

snosenzo:
> For allFrames, however, I think we want to include all messages, not only those that would result from newly enabled converters

@foxymiles I agree that `allFrames` should include all messages on subscribed, preloaded topics both from converters and the source regardless of whether they were just added. `currentFrame` should not contain messages that are preloaded for a given panel. Is this not the current implementation?

foxymiles:
> @foxymiles I agree that `allFrames` should include all messages on subscribed, preloaded topics both from converters and the source regardless of whether they were just added. `currentFrame` should not contain messages that are preloaded for a given panel. Is this not the current implementation?

The behavior I'm going for in this thread should only affect the case in which we do not have new messages but the active converters have changed. In that case this should happen:
1. allFrames gets updated to include all messages, both converted and unconverted
2. a new currentFrame is generated that only contains messages that came from newly activated converters

Does that make sense?

packages/studio-base/src/components/PanelExtensionAdapter/messageProcessing.ts
```diff
@@ -0,0 +1,199 @@
...
+      receiveTime: messageEvent.receiveTime,
+      message: convertedMessage,
+      originalMessageEvent: messageEvent,
+      sizeInBytes: messageEvent.sizeInBytes,
+    });
+  }
+}
+
+/**
+ * Returns a new map consisting of all items in a not present in b.
```
snosenzo, line 52: nit
```suggestion
 * Returns a new map consisting of all items in `a` not present in `b`.
```

foxymiles:
Feedback from @jtbandes: we really need to store the last message on each topic and process that when we get new converters, not the previous frame.

packages/studio-base/src/components/PanelExtensionAdapter/renderState.ts
```diff
@@ -254,93 +186,72 @@ function initRenderStateBuilder(): BuildRenderStateFn {
...
+
+      prevCurrentFrame = currentFrame;
     }
 
     if (watchedFields.has("allFrames")) {
-      // see comment for prevBlocksRef on why extended message store updates are gated this way
+      // Rebuild allFrames if we have new blocks or if our conversions have changed.
       const newBlocks = playerState?.progress.messageCache?.blocks;
-      if (newBlocks && prevBlocks !== newBlocks) {
+      if ((newBlocks && prevBlocks !== newBlocks) || conversionsChanged) {
```
snosenzo, line 223: Can we make this check `preloadedConversionsChanged` so that we don't update this for non-preloaded subscriptions?

snosenzo (approved):

packages/studio-base/src/components/PanelExtensionAdapter/renderState.ts
foxymiles: You're right that would be more correct. Not sure if it's worth the extra complexity since these won't be changing very often. I'll take a look at how much extra work this adds.

packages/studio-base/src/components/PanelExtensionAdapter/messageProcessing.ts
```diff
@@ -0,0 +1,199 @@
...
+  // convertTo, then we don't need to do a conversion.
+  for (const subscription of subscriptions) {
+    if (!subscription.convertTo) {
+      unconvertedSubscriptionTopics.add(subscription.topic);
+      continue;
+    }
+
+    // If the convertTo is the same as the original schema for the topic then we don't need to
+    // perform a conversion.
+    const noConversion = sortedTopics.find(
```
jtbandes, line 107: This logic seems to be written to handle multiple topics with the same name and different schemaNames (it will look through all topics to check both name and schemaName, rather than just checking name). However, the logic below looking for a topic that can be converted to the desired schema only gets the _first_ topic matching the requested name. This seems inconsistent and I think this code should stick with one assumption – either there might be topics with the same name and different schemas, or there might not. (I'm not sure how pervasive that assumption is in the code currently. We certainly haven't tested with much sample data that has this.)

packages/studio-base/src/components/PanelExtensionAdapter/messageProcessing.ts
```diff
@@ -0,0 +1,199 @@
...
+
+  // When a subscription with a convertTo exists, we use this map to lookup a
+  // converter which can produce the desired output message schema. The keys for
+  // the map are `topic + input schema`.
+  //
+  // This allows the runtime message event handler logic which builds
+  // currentFrame and allFrames to lookup whether the incoming message event has
+  // converters to run by looking up the topic + schema of the message event in
+  // this map.
+  topicSchemaConverters: TopicSchemaConverterMap;
```
jtbandes, line 78: I'd actually be interested in migrating this to use TwoKeyMap. It feels like it could make the code cleaner.

packages/studio-base/src/components/PanelExtensionAdapter/messageProcessing.ts
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    // Since we don't have an existing topic with out destination schema we need to find
+    // a converter that will convert from the topic to the desired schema
+    const subscriberTopic = sortedTopics.find((topic) => topic.name === subscription.topic);
+    if (!subscriberTopic) {
+      continue;
+    }
+
+    const key = converterKey(subscription.topic, subscriberTopic.schemaName ?? "<no-schema>");
```
jtbandes, line 122: If we need a sentinel key for no schema, it definitely seems like it should be stored in a constant somewhere rather than hard-coded where it needs to be used. (btw why is schemaName optional here? is that normal? I guess I've forgotten)

packages/studio-base/src/components/PanelExtensionAdapter/messageProcessing.ts
```diff
@@ -0,0 +1,199 @@
...
+/**
+ * Function to iterate and call function over multiple sorted arrays in sorted order across all items in all arrays.
+ * Time complexity is O(t*n) where t is the number of arrays and n is the total number of items in all arrays.
+ * Space complexity is O(t) where t is the number of arrays.
+ * @param arrays - sorted arrays to iterate over
+ * @param compareFn - function called to compare items in arrays. Returns a positive value if left is larger than right,
+ *  a negative value if right is larger than left, or zero if both are equal
+ * @param forEach - callback to be executed on all items in the arrays to iterate over in sorted order across all arrays
+ */
+export function forEachSortedArrays<Item>(
```
jtbandes, line 161: What's old is new again https://github.com/foxglove/rosbag/blob/802906f42eadf4b0565c51c5f3489ffce82b708a/src/nmerge.ts

I'd recommend moving this to a separate file, maybe under @foxglove/den/collection

packages/studio-base/src/components/PanelExtensionAdapter/renderState.ts
```diff
@@ -254,93 +191,73 @@ function initRenderStateBuilder(): BuildRenderStateFn {
...
-                  topic: messageEvent.topic,
-                  schemaName: converter.toSchemaName,
-                  receiveTime: messageEvent.receiveTime,
-                  message: convertedMessage,
-                  originalMessageEvent: messageEvent,
-                  sizeInBytes: messageEvent.sizeInBytes,
-                });
-              }
-            }
+      if (currentFrame && currentFrame !== prevCurrentFrame) {
```
jtbandes, line 194: Is there a way to do some early returns/breaks to reduce nesting? This could be a use case for labeling if statements, e.g. `currentFrame: if (watchedFields...) { ... break currentFrame; }`

packages/studio-base/src/components/PanelExtensionAdapter/renderState.ts
```diff
@@ -254,93 +191,73 @@ function initRenderStateBuilder(): BuildRenderStateFn {
...
+        for (const messageEvent of lastMessageByTopic.values()) {
+          convertMessage(messageEvent, newConverters, postProcessedFrame);
         }
+        renderState.currentFrame = postProcessedFrame;
+        shouldRender = true;
+      } else if (currentFrame !== prevCurrentFrame) {
+        // Otherwise if we're replacing a non-empty frame with an empty frame and
+        // conversions haven't changed, include the empty frame in the new render state.
+        renderState.currentFrame = currentFrame;
+        shouldRender = true;
```
jtbandes, line 220: what's the point of this section? Does "empty frame" mean `currentFrame: {}`? When would that ever happen and if it does, why is it important that we re-render? I thought an empty frame like this would not need to re-trigger any panel logic unless other fields besides currentFrame are changing.

packages/studio-base/src/components/PanelExtensionAdapter/messageProcessing.ts
```diff
@@ -0,0 +1,199 @@
...
+      const cursor = cursors[i]!;
+      const array = arrays[i]!;
+      if (cursor >= array.length) {
+        continue;
+      }
+      const item = array[cursor]!;
+      if (minCursorIndex == undefined) {
+        minCursorIndex = i;
+      } else {
+        const minItem = arrays[minCursorIndex]![cursors[minCursorIndex]!]!;
```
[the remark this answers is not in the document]
foxymiles: This is existing code I just relocated from renderState and tried to avoid changing since it seems to be working.

packages/studio-base/src/components/PanelExtensionAdapter/messageProcessing.ts
foxymiles: A bunch of this code is just moved over from renderState. Maybe we can do a followup PR to refine the logic?

packages/studio-base/src/components/PanelExtensionAdapter/renderState.ts
foxymiles: Just trying to preserve existing semantics:

https://github.com/foxglove/studio/blob/80adbe260c0076f2e9308f451649d9dfe433ebab/packages/studio-base/src/components/PanelExtensionAdapter/renderState.ts#L257-L293

packages/studio-base/src/components/PanelExtensionAdapter/renderState.ts
foxymiles: I think this whole renderState calculation thing could be simplified and made a bit less confusing but I've tried to keep this as much as possible like our current implementation on main since subtle changes here can break expectations downstream.

foxymiles:
@jtbandes I think you make a lot of good points but a lot of the changes here are just re-org of existing code.

foxymiles:
> Got it…hard to see what’s copied because GitHub doesn’t tell you. I miss that about phabricator.

Yes I agree that would be a great feature.

PR #5954, opened 2023-05-09
feat: marker datatypes support scale

smbera:
**User-Facing Changes**

**Description**
`visualization_msgs/Marker` and `visualization_msgs/MarkerArray` support scale.

Before: 
image

After: 
image


jtbandes:
Hi, thanks for your contribution!

The `scale` property can mean different things for different types of markers, as described [here](https://wiki.ros.org/rviz/DisplayTypes/Marker). For a `CUBE`, the scale represents the cube size; for an `ARROW`, scale means different things depending on whether `points` is also used. For text, only `scale.z` is used. For lines, only `scale.x` is used. As such, it seems like the proposed override feature could be confusing for users.

Could your provide some more context on what you are trying to accomplish with this feature? How have you handled this in other tools (like rviz)?

jtbandes:
Since we haven't heard back, I'm going to close out this PR for now. We might be open to a feature like this if we can figure out a path forward that avoids user confusion in the various cases I mentioned above!

PR #5956, opened 2023-05-09
Add @emotion/* packages to constraints

jtbandes:
**User-Facing Changes**
None

**Description**
Originally I didn't add these because I thought they weren't published at the same time, but based on https://github.com/foxglove/studio/pull/5945 & #5952  it seems like we want to keep them in sync.


snosenzo (approved):

constraints.pro
```diff
@@ -59,6 +59,8 @@ version_group(Dep, storybook) :-
   has_prefix('@storybook/', Dep), Dep \= '@storybook/testing-library'.
 version_group(Dep, typescript_eslint) :-
   has_prefix('@typescript-eslint/', Dep).
+version_group(Dep, emotion) :-
```
2metres, line 62: 😎  Was not aware of this feature

PR #5964, opened 2023-05-10
Fix Display Frame Error on refresh

snosenzo:
**User-Facing Changes**

 - [3D Panel] Fix Display Frame Error on refresh sometimes

**Description**

The issue was that the Renderer was setting `this.renderFrameId` using `getDefaultFrame` and it was going to render fine, but the settings nodes weren't getting regenerated so it wasn't clearing the error anymore. So any change to the settings (like moving the camera) would remove the error. 

I updated frame fallback conditions to hopefully be a bit clearer. The display frame settings will only ever display the value of followTf if it exists and if not it will show the followFrameId. The setting of the followFrameId is handled in CameraStateSettings so that it can be updated with the settings node and on transform tree changes. I've removed `renderFrameId` from the renderer to clear up 

FG-2985

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/IRenderer.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/IRenderer.ts
@@ -206,12 +206,12 @@ export interface IRenderer extends EventEmitter<RendererEvents> {
   transformTree: TransformTree;
   coordinateFrameList: SelectEntry[];
   currentTime: bigint;
-  /* Root frame of the follow frame */
+  /** Coordinate frame that transforms are applied through to the follow frame. Should be unchanging. */
   fixedFrameId: string | undefined;
   /**
    * The frameId that we _want_ to follow and render in if it exists.
    */
-  followFrameId: string | undefined;
+  readonly followFrameId: string | undefined;
 
   labelPool: LabelPool;
   markerPool: MarkerPool;
@@ -303,7 +303,7 @@ export interface IRenderer extends EventEmitter<RendererEvents> {
 
   addMessageEvent(messageEvent: Readonly<MessageEvent<unknown>>): void;
 
-  /* Set desired render/display frame, will render using fallback if id is undefined or frame does not exist */
+  /**  Set desired render/display frame, will render using fallback if id is undefined or frame does not exist */
   setFollowFrameId(frameId: string | undefined): void;
 
   /** Match the behavior of `tf::Transformer` by stripping leading slashes from
--- a/packages/studio-base/src/panels/ThreeDeeRender/IRenderer.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/IRenderer.ts
@@ -206,8 +206,11 @@ export interface IRenderer extends EventEmitter<RendererEvents> {
   transformTree: TransformTree;
   coordinateFrameList: SelectEntry[];
   currentTime: bigint;
+  /* Root frame of the follow frame */
   fixedFrameId: string | undefined;
-  renderFrameId: string | undefined;
+  /**
+   * The frameId that we _want_ to follow and render in if it exists.
+   */
   followFrameId: string | undefined;
 
   labelPool: LabelPool;
--- a/packages/studio-base/src/panels/ThreeDeeRender/IRenderer.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/IRenderer.ts
@@ -45,6 +45,7 @@ export type RendererEvents = {
     variables: ReadonlyMap<string, VariableValue> | undefined,
     renderer: IRenderer,
   ) => void;
+  /** Fired when the structure of the transform tree changes ie: new frame added/removed or frame assigned new parent */
   transformTreeUpdated: (renderer: IRenderer) => void;
   settingsTreeChange: (renderer: IRenderer) => void;
   configChange: (renderer: IRenderer) => void;
--- a/packages/studio-base/src/panels/ThreeDeeRender/IRenderer.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/IRenderer.ts
@@ -303,6 +303,9 @@ export interface IRenderer extends EventEmitter<RendererEvents> {
 
   addMessageEvent(messageEvent: Readonly<MessageEvent<unknown>>): void;
 
+  /* Set desired render/display frame, will render using fallback if id is undefined or frame does not exist */
+  setFollowFrameId(frameId: string | undefined): void;
+
   /** Match the behavior of `tf::Transformer` by stripping leading slashes from
    * frame_ids. This preserves compatibility with earlier versions of ROS while
    * not breaking any current versions where:
--- a/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
@@ -992,7 +992,7 @@ export class Renderer extends EventEmitter<RendererEvents> implements IRenderer 
 
   public setFollowFrameId(frameId: string | undefined): void {
     this.followFrameId = frameId;
-    log.debug(`Setting render frame to ${frameId}`);
+    log.debug(`Setting followFrameId to ${frameId}`);
   }
 
   #frameHandler = (currentTime: bigint): void => {
--- a/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
@@ -997,7 +997,7 @@ export class Renderer extends EventEmitter<RendererEvents> implements IRenderer 
 
   #frameHandler = (currentTime: bigint): void => {
     this.currentTime = currentTime;
-    this.#checkFollowFrame();
+    this.#updateFrameErrors();
     this.#updateFixedFrameId();
     this.#updateResolution();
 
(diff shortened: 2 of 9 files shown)


packages/studio-base/src/panels/ThreeDeeRender/IRenderer.ts
```diff
@@ -206,8 +206,11 @@ export interface IRenderer extends EventEmitter<RendererEvents> {
   transformTree: TransformTree;
   coordinateFrameList: SelectEntry[];
   currentTime: bigint;
+  /* Root frame of the follow frame */
```
jtbandes, line 209: Missed a `*`, also we might want to not say this is _always_ the root of the follow frame, in case we decide to add a setting to change the fixed frame.

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -1032,9 +990,15 @@ export class Renderer extends EventEmitter<RendererEvents> implements IRenderer
     }
   }
 
+  public setFollowFrameId(frameId: string | undefined): void {
+    this.followFrameId = frameId;
+    log.debug(`Setting render frame to ${frameId}`);
```
jtbandes, line 995:
```suggestion
    log.debug(`Setting followFrameId to ${frameId}`);
```

packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
```diff
@@ -187,7 +187,7 @@ export function ThreeDeeRender(props: {
   );
 
   // The frame we care about for syncing purposes can be either of these.
```
jtbandes, line 189: out of date comment

packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
```diff
@@ -187,7 +187,7 @@ export function ThreeDeeRender(props: {
...
-  const effectiveRendererFrameId = renderer?.followFrameId ?? renderer?.renderFrameId;
+  const effectiveRendererFrameId = renderer?.followFrameId;
```
jtbandes, line 190: maybe just remove this variable?

packages/studio-base/src/panels/ThreeDeeRender/renderables/CameraStateSettings.ts
```diff
@@ -2,7 +2,7 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
-import { t } from "i18next";
+import i18next, { t } from "i18next";
```
jtbandes, line 5: let's stay consistent within a file on whether we use `t` or `i18next.t`

packages/studio-base/src/panels/ThreeDeeRender/renderables/CameraStateSettings.ts
```diff
@@ -388,9 +390,59 @@ export class CameraStateSettings extends SceneExtension implements ICameraHandle
...
+    this.#checkFollowTf();
+
+    const followTfFrameExists = followTf != undefined && transformTree.hasFrame(followTf);
+    if (followTfFrameExists) {
+      this.renderer.setFollowFrameId(followTf);
+      return;
+    }
+
+    const newTransformFrameCount = transformTree.frames().size;
+    if (this.#lastTransformFrameCount === newTransformFrameCount) {
```
jtbandes, line 414: not ideal to just track size, in case the size stays the same but items change... would it work to use the `"transformTreeUpdated"` renderer event?

packages/studio-base/src/panels/ThreeDeeRender/transforms/TransformTree.ts
```diff
@@ -245,6 +248,35 @@ export class TransformTree {
 
     return output;
   }
+
+  /** Get heuristically most valid follow frame Id */
+  public getMostValidFollowFrameId(): string | undefined {
```
jtbandes, line 253: "most valid" is weird, something is either valid or it isn't. Maybe "getBest" or "getPreferred" or "getDefault"? Default would match the variable name of DEFAULT_FRAME_IDS.

packages/studio-base/src/panels/ThreeDeeRender/IRenderer.ts
```diff
@@ -206,8 +206,11 @@ export interface IRenderer extends EventEmitter<RendererEvents> {
...
   fixedFrameId: string | undefined;
-  renderFrameId: string | undefined;
+  /**
+   * The frameId that we _want_ to follow and render in if it exists.
+   */
   followFrameId: string | undefined;
```
jtbandes, line 214: should this be readonly now?

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -1252,79 +1239,37 @@ export class Renderer extends EventEmitter<RendererEvents> implements IRenderer
...
-            i18next.t("threeDee:noCoordinateFramesFound"),
-          );
-        }
-        this.fixedFrameId = undefined;
-        return;
-      } else {
-        log.debug(`Setting render frame to ${this.renderFrameId}`);
-        this.settings.errors.remove(FOLLOW_TF_PATH, NO_FRAME_SELECTED);
-      }
+  // update fixed frame to be the root of the render frame
```
jtbandes, line 1242: Comment does not appear to match what's happening in the function.

Also, maybe the function can have a better name than `checkFollowFrame`... at first glance I'm not clear why it needs to be "checked"? Maybe `updateFrameErrors`?

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -1252,79 +1239,37 @@ export class Renderer extends EventEmitter<RendererEvents> implements IRenderer
...
     }
 
-    const frame = this.transformTree.frame(this.renderFrameId);
+    this.settings.errors.remove(FOLLOW_TF_PATH, NO_FRAME_SELECTED);
+
+    const frame = this.transformTree.frame(this.followFrameId);
+
+    // The follow frame id should be chosen from a frameId that exists, but
+    // we still need to watch out for the case that the transform tree was
+    // cleared before that could be updated
```
jtbandes, line 1260: is this a rare/theoretical edge case or something that you can actually make it do in practice?

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -1066,6 +1033,26 @@ export class Renderer extends EventEmitter<RendererEvents> implements IRenderer
     this.gl.info.reset();
   };
 
+  #updateFixedFrameId(): void {
+    const frame =
```
jtbandes, line 1037: Seems like we are doing this on every frame. Is that something we could improve and do less often, like only when the root actually changes? Or maybe it's negligible compared to the tf tree updates we normally do on each frame?

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
snosenzo: Not dependably, but I had an assert here before and it ran into it on occasion.

packages/studio-base/src/panels/ThreeDeeRender/renderables/CameraStateSettings.ts
snosenzo: This is the logic we had before, though I know it's not ideal. I think it should be fine given that the only way we remove coordinate frames is when a URDF is removed and transforms are only ever added one at a time. I don't think that's going to change anytime soon. But yeah that's more context than one should need to be able to say that its safe, we could do a deeper equals on the frame list?

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
snosenzo: we used to be doing this on every frame in `#updateFrames`. We are doing this here because the follow frame could have been removed from the tree between setting the followframeId.

packages/studio-base/src/panels/ThreeDeeRender/renderables/CameraStateSettings.ts
snosenzo: I've updated this to remove this logic since it was causing an error in the URDF story which doesn't have a followTf assigned. This was happening because it wasn't recalculating the followFrameId when a frame got a new parent, which we should be accounting for since in the defaultFollowFrame function we want to get the root with the most children. So we should be calculating this whenever the transform tree changes, which is when the `updateTransformTree` is fired as it would happen, so we don't need any of this variable tracking 👍

packages/studio-base/src/panels/ThreeDeeRender/renderables/CameraStateSettings.ts
```diff
@@ -388,9 +390,49 @@ export class CameraStateSettings extends SceneExtension implements ICameraHandle
...
+      return;
+    }
+
+    // No valid renderFrameId set, or new frames have been added, fall back to selecting the
+    // heuristically most valid frame (if any frames are present)
+    const followFrameId = transformTree.getDefaultFollowFrameId();
+    this.renderer.setFollowFrameId(followFrameId);
+  }
+
+  #checkFollowTf = (): void => {
```
jtbandes, line 416: `updateFollowTfError`?

jtbandes (approved):

PR #5977, opened 2023-05-10
Mark `subscribe(topics[])` as deprecated.

defunctzombie:
**User-Facing Changes**
None

**Description**
Subscribing to an array of topics does not allow for controlling whether you want partial or full subscriptions and instead assumes full subscriptions which is often more than you want to subscribe to. By deprecating this form of subscribe we lead users to the more flexible form with subscriber options.

Related: https://github.com/foxglove/studio/pull/5267

### DIFF
--- a/packages/studio/src/index.ts
+++ b/packages/studio/src/index.ts
@@ -309,11 +309,19 @@ export type PanelExtensionContext = {
    * array will unsubscribe from all topics.
    *
    * Calling subscribe with an empty array of topics is analagous to unsubscribeAll.
+   *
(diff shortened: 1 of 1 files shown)


defunctzombie:
@jtbandes thoughts?

jtbandes (approved):
Sounds good. Though I guess we need to keep the old code around as long as we want to support old extensions.

defunctzombie:
> Though I guess we need to keep the old code around as long as we want to support old extensions

Yep - which is basically forever.

PR #5978, opened 2023-05-10
Use automerge-action to automerge Dependabot PRs

jtbandes:
**User-Facing Changes**
None

**Description**
Uses [@pascalgn/automerge-action](https://github.com/pascalgn/automerge-action) to auto-merge PRs from dependabot.


jtbandes:
Just remembered this won't work because we also need to auto-approve the PR 😄

PR #5988, opened 2023-05-10
Fix Renderables not clearing when seeking back to beginning of file

snosenzo:
**User-Facing Changes**

 - Fix Renderables not clearing when seeking back to beginning of file

**Description**
It wasn't just images that were the problem. We weren't rendering again after clearing. 

FG-3118


jtbandes (approved):
:wow:

PR #5989, opened 2023-05-10
Fix ImageAnnotation fill not updating after initial message

snosenzo:
**User-Facing Changes**

 - Fix ImageAnnotation fill not updating after initial message
**Description**

FG-3260

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/annotations/RenderableLineAnnotation.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/annotations/RenderableLineAnnotation.ts
@@ -180,7 +180,7 @@ export class RenderableLineAnnotation extends Renderable<BaseUserData, /*TRender
       if (
         this.#numPoints == undefined ||
         !this.#geometry ||
-        pointsLength !== this.#numPoints ||
+        pointsLength > this.#numPoints ||
         this.#style !== style
       ) {
         // Need to recreate the geometry when length changes: https://github.com/mrdoob/three.js/issues/21488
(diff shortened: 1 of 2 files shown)


jtbandes (approved):
Thanks!

jtbandes:
Should we try to test this in a storybook? Like send one message and then in the next tick send a 2nd annotations message?

snosenzo:
Found another issue where if the geometry line buffer was previously larger, then the next annotation would use points from the previous geometry. Caught it when iterating with update story.

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/annotations/RenderableLineAnnotation.ts
```diff
@@ -180,7 +180,7 @@ export class RenderableLineAnnotation extends Renderable<BaseUserData, /*TRender
       if (
         this.#numPoints == undefined ||
         !this.#geometry ||
-        pointsLength > this.#numPoints ||
+        pointsLength !== this.#numPoints ||
```
jtbandes, line 183: was this the source of a bug? I think we should be able to reuse the geometry if it's larger than necessary, right?

jtbandes: Maybe need to use drawRange?

snosenzo: Figured it out. I tried using drawRange and that didn't work. Then I realized the instanceCount stuff and we were calculating that incorrectly. Should represent the number of lines and our counts for polygons and line strips were wrong.

jtbandes: Thanks for fixing!

PR #5990, opened 2023-05-10
Add topic and details to selected ImageAnnotations

snosenzo:
**User-Facing Changes**

 - [Image Beta] Show a message's topic and details for Image Annotations on selection

**Description**

FG-3166


jtbandes (commented):
I think we should show the original message, not the normalized annotation.

Also note that these messages often come in with multiple annotations in a single ImageAnnotations or ImageMarkerArray message. Should we show the original array message or the single annotation that was clicked?

snosenzo:
> I think we should show the original message, not the normalized annotation.

I agree, but also I'm not seeing SceneEntities or Markers do this. Doing this would entail adding a `path` to each normalized annotation and using the path with the original message to get the annotation to show it. 

> Also note that these messages often come in with multiple annotations in a single ImageAnnotations or ImageMarkerArray message. Should we show the original array message or the single annotation that was clicked?

I think people would expect the single annotation that was clicked and the parent ImageAnnotation.

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/annotations/RenderableTopicAnnotations.ts
```diff
@@ -171,6 +180,7 @@ export class RenderableTopicAnnotations extends THREE.Object3D {
                 this.add(line);
               }
               this.#lines.push(line);
+              line.setOriginalMessage(this.#originalMessage);
               line.setAnnotation(
```
jtbandes, line 184: could also consider passing the original message and annotation as 2 arguments to the same function, since they should always change at the same time

jtbandes (approved):

jtbandes:
This doesn't fully fix FG-3166 because we also need to fix it so the topic name displays here: 
image

PR #5998, opened 2023-05-11
[POC] demonstrate topic mapping 

foxymiles:
**User-Facing Changes**
None

**Description**
POC of topic mapping approach to dynamic topic names. This hardcodes the following mappings:

```typescript
  readonly #mappings: Map<string, string> = new Map([
    ["/CAM_FRONT/image_rect_compressed", "/remapped_cam_front/image_rect_compressed"],
    ["/CAM_FRONT/camera_info", "/remapped_cam_front/camera_info"],
  ]);
```

It seems to work with 3d, image, plot and raw messages panels. A real implementation would allow a new type of extension to define these mappings dynamically. 

Also I've made no attempt to address performance here. We'd want to memoize.


foxymiles:
Closing in favor of actual PR.

PR #5999, opened 2023-05-11
Add layoutEmptyState to AppContext

defunctzombie:
**User-Facing Changes**
None

**Description**
Provides the fallback component if a layout is not available.


jtbandes (approved):
Seems nicer than swapping out the whole PanelLayout

PR #6000, opened 2023-05-11
Remove ROS2 native connection

achim-k:
**User-Facing Changes**
- Remove ROS2 native connector

**Description**
Removes the ROS2 native connector in favor of the `foxglove_bridge` / `rosbridge` connections.

See also #4900


yarn.lock
```diff
@@ -2490,7 +2477,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@foxglove/rtps@npm:1.6.0, @foxglove/rtps@npm:^1.6.0":
+"@foxglove/rtps@npm:1.6.0":
```
jtbandes, line 2480: I think we should be able to completely remove @foxglove/rtps as well?

jtbandes (approved):

yarn.lock
defunctzombie: 🔪

jtbandes: Theoretically lint-dependencies will catch this.

jtbandes: image

PR #6003, opened 2023-05-11
Refactor node playground stories to use play syntax

2metres:
**User-Facing Changes**
None

**Description**
Refactor NodePlayground stories to use `play: () => {}`

### DIFF
--- a/packages/studio-base/src/panels/NodePlayground/index.stories.tsx
+++ b/packages/studio-base/src/panels/NodePlayground/index.stories.tsx
@@ -12,6 +12,7 @@
 //   You may not use this file except in compliance with the License.
 
 import { StoryObj } from "@storybook/react";
+import { fireEvent, screen } from "@storybook/testing-library";
 
 import NodePlayground from "@foxglove/studio-base/panels/NodePlayground";
 import rawUserUtils from "@foxglove/studio-base/players/UserNodePlayer/nodeTransformerWorker/typescript/rawUserUtils";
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/panels/NodePlayground/index.stories.tsx
```diff
@@ -188,17 +186,14 @@ export const EditorShowsNewCodeWhenUserNodesChange: StoryObj = {
...
-          // Change the userNodes to confirm the code in the Editor updates
-          actions.setUserNodes({
-            nodeId1: {
-              name: "/studio_script/script",
-              sourceCode: utilsSourceCode,
-            },
-          });
-          el.querySelectorAll<HTMLElement>("[data-testid=node-explorer]")[0]?.click();
-        }, 500);
+      onMount={(_, actions) => {
```
jtbandes, line 189: Ideally we'd get rid of onMount. I think it's largely used for stories because play() didn't exist at the time it was added.

2metres: Can we get rid of it here? throw that into a useEffect or something?

packages/studio-base/src/panels/NodePlayground/index.stories.tsx
```diff
@@ -497,85 +452,40 @@ export const BottomBarErrorsOpen: StoryObj = {
...
-              source: "Source B",
-              code: 2304,
-            },
-            {
-              message: "This is a hint message (without line or column numbers).",
-              severity: 1,
-              source: "Source C",
-              code: 2304,
-            },
+            { message: Array(10).fill("Long error that might wrap.").join(" "), severity: 8, source: "Typescript", startLineNumber: 0, startColumn: 6, endLineNumber: 72, endColumn: 20, code: 2304 },
```
foxymiles, line 457: Not a fan of these really long lines.

foxymiles (approved):
Nice cleanup

PR #6004, opened 2023-05-11
Pass original message event to message converters

foxymiles:
**User-Facing Changes**
Pass the original message event to message converters to provide access to publishTime and other properties of the original event.

**Description**
Provide extra fields of the MessageEvent by passing the full MessageEvent as a second argument to message converters.

### DIFF
--- a/packages/studio/src/index.ts
+++ b/packages/studio/src/index.ts
@@ -411,7 +411,7 @@ export type ExtensionPanelRegistration = {
 export type RegisterMessageConverterArgs<Src> = {
   fromSchemaName: string;
   toSchemaName: string;
-  converter: ((msg: Src) => unknown) | ((msg: Src, event: Immutable<MessageEvent<Src>>) => unknown);
+  converter: (msg: Src, event: Immutable<MessageEvent<Src>>) => unknown;
 };
 
 export interface ExtensionContext {
--- a/packages/studio/src/index.ts
+++ b/packages/studio/src/index.ts
@@ -411,7 +411,7 @@ export type ExtensionPanelRegistration = {
 export type RegisterMessageConverterArgs<Src> = {
   fromSchemaName: string;
   toSchemaName: string;
-  converter: (msg: Src) => unknown;
+  converter: ((msg: Src) => unknown) | ((msg: Src, event: Immutable<MessageEvent<Src>>) => unknown);
 };
 
 export interface ExtensionContext {
--- a/packages/studio-base/src/components/PanelExtensionAdapter/messageProcessing.ts
+++ b/packages/studio-base/src/components/PanelExtensionAdapter/messageProcessing.ts
@@ -40,7 +40,7 @@ export function convertMessage(
   const key = converterKey(messageEvent.topic, messageEvent.schemaName);
   const matchedConverters = converters.get(key);
   for (const converter of matchedConverters ?? []) {
-    const convertedMessage = converter.converter(messageEvent.message);
+    const convertedMessage = converter.converter(messageEvent.message, messageEvent);
     convertedMessages.push({
       topic: messageEvent.topic,
       schemaName: converter.toSchemaName,
--- a/packages/studio-base/src/components/PanelExtensionAdapter/renderState.test.ts
+++ b/packages/studio-base/src/components/PanelExtensionAdapter/renderState.test.ts
@@ -535,6 +535,8 @@ describe("renderState", () => {
   // for two different converters.
   it("should run the correct converter", () => {
     const buildRenderState = initRenderStateBuilder();
+    const converter1 = jest.fn().mockImplementation(() => "srcschema-destschema");
+    const converter2 = jest.fn().mockImplementation(() => "srcschemade-stschema");
     const state = buildRenderState({
       watchedFields: new Set(["topics", "currentFrame"]),
       playerState: undefined,
@@ -545,7 +547,7 @@ describe("renderState", () => {
           schemaName: "srcschema",
           receiveTime: { sec: 0, nsec: 0 },
           sizeInBytes: 1,
-          message: {},
+          message: "raw-message",
         },
       ],
       colorScheme: undefined,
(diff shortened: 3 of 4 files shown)


packages/studio-base/src/providers/ExtensionCatalogProvider.tsx
```diff
@@ -2,25 +2,33 @@
...
+  VersionedMessageConverter,
 } from "@foxglove/studio-base/context/ExtensionCatalogContext";
 import { ExtensionLoader } from "@foxglove/studio-base/services/ExtensionLoader";
 import { ExtensionInfo, ExtensionNamespace } from "@foxglove/studio-base/types/Extensions";
 
 const log = Logger.getLogger(__filename);
 
+function extensionSupportsV2MessageConverters(extension: Immutable<ExtensionInfo>): boolean {
+  const studioBuildDependency = extension.devDependencies?.["@foxglove/studio"] ?? "0.0.0";
+  return semverCompare(studioBuildDependency, "1.54.0") >= 0;
```
jtbandes, line 26: what happens if devDependencies had something like `*` or `^1.52.0`?

packages/studio-base/src/context/ExtensionCatalogContext.ts
```diff
@@ -15,6 +15,12 @@ export type RegisteredPanel = {
   registration: ExtensionPanelRegistration;
 };
 
+export type VersionedMessageConverter<T> =
+  | ({ version: "1" } & RegisterMessageConverterArgs<T> & {
+        converter: (msg: T) => unknown;
+      })
+  | ({ version: "2" } & RegisterMessageConverterArgs<T>);
```
jtbandes, line 22: why strings instead of numbers for the version?

jtbandes (approved):
This approach seems reasonable to me. We should think about how to discuss this kind of change in our docs.

packages/studio-base/src/context/ExtensionCatalogContext.ts
foxymiles: Makes no difference right as long as they're distinct?

packages/studio-base/src/providers/ExtensionCatalogProvider.tsx
foxymiles: Good question. Looks like `semverCompare` throws on stuff like that so we should probably strip everything but [0-9.] from the version first?

packages/studio-base/src/providers/ExtensionCatalogProvider.tsx
```diff
@@ -2,25 +2,39 @@
...
   ExtensionCatalogContext,
   RegisteredPanel,
+  VersionedMessageConverter,
 } from "@foxglove/studio-base/context/ExtensionCatalogContext";
 import { ExtensionLoader } from "@foxglove/studio-base/services/ExtensionLoader";
 import { ExtensionInfo, ExtensionNamespace } from "@foxglove/studio-base/types/Extensions";
 
 const log = Logger.getLogger(__filename);
 
+function extensionSupportsV2MessageConverters(extension: Immutable<ExtensionInfo>): boolean {
```
defunctzombie, line 24: This needs a comment describing the desired flow and noting any gotchas.

packages/studio-base/src/providers/ExtensionCatalogProvider.tsx
```diff
@@ -66,11 +80,15 @@ function activateExtension(
       };
     },
 
-    registerMessageConverter<Src>(args: RegisterMessageConverterArgs<Src>) {
+    registerMessageConverter: <Src,>(args: RegisterMessageConverterArgs<Src>) => {
       log.debug(
         `Extension ${extension.qualifiedName} registering message converter from: ${args.fromSchemaName} to: ${args.toSchemaName}`,
       );
-      messageConverters.push(args as RegisterMessageConverterArgs<unknown>);
+      if (extensionSupportsV2MessageConverters(extension)) {
```
defunctzombie, line 87: Alternatively - you could return the version of message converter to use so if we introduce v3 then this code remains unchanged.

packages/studio-base/src/providers/ExtensionCatalogProvider.tsx
foxymiles: Yeah I think the main gotcha here is if for some reason their package.json doesn't contain the correct version of studio but they're expecting a full message event in their code then we'll send the wrong object. If they're using our create-foxglove-extension script then this shouldn't happen but of course we have no control over any changes they make to their package.json after that.

Also I suppose there's nothing stopping them from trying to build an extension from scratch without our setup script although I'm not sure why someone would do that.

defunctzombie: Should we make a statement that no-version = latest version? Or surface the lack of version somehow in the extension listing for installed extensions?

foxymiles: I think we have three choices in that case:
1. Assume the extension is using the old version.
2. Assume the extension is using the latest version.
3. Assume the extension is broken and disable it and surface that somehow to the user.

foxymiles: I think this is one area where the alternate proposal of passing the full MessageEvent as a second argument looks better because it sidesteps these concerns.

packages/studio-base/src/providers/ExtensionCatalogProvider.tsx
foxymiles: I've added some extra logic to strip extra characters from the version and to catch any errors and treat failed parses as a v1 extension.

foxymiles:
I've switched this to our approach that passes the original MessageEvent as a second argument to the message converter.

packages/studio/src/index.ts
```diff
@@ -411,7 +411,7 @@ export type ExtensionPanelRegistration = {
 export type RegisterMessageConverterArgs<Src> = {
   fromSchemaName: string;
   toSchemaName: string;
-  converter: (msg: Src) => unknown;
+  converter: ((msg: Src) => unknown) | ((msg: Src, event: Immutable<MessageEvent<Src>>) => unknown);
```
jtbandes, line 414:
```suggestion
  converter: (msg: Src, event: Immutable<MessageEvent<Src>>) => unknown;
```

I think it should work this way. The additional argument shouldn't cause TS errors for people who pass in functions that only take 1 argument.

jtbandes (approved):

packages/studio/src/index.ts
foxymiles: Yes you're right that also seems to work.

PR #6006, opened 2023-05-12
[Image Beta] Add missing setHasCalibrationTopic(true)

jtbandes:
**User-Facing Changes**
None (beta)

**Description**
Follow-up to #5941 
Ensures that 3D topics can be enabled immediately after a panel is added if there was an auto-selected calibration topic.

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageMode.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageMode.ts
@@ -268,6 +268,9 @@ export class ImageMode
         draft.imageMode.calibrationTopic = matchingCalibrationTopic.name;
       }
     });
+    if (matchingCalibrationTopic) {
+      this.#setHasCalibrationTopic(true);
+    }
   };
 
   /** Choose a calibration topic that best matches the given `imageTopic`. */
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageMode.ts
```diff
@@ -480,6 +483,7 @@ export class ImageMode
           this.renderer.updateConfig((draft) => {
             draft.imageMode.calibrationTopic = calibrationTopic.name;
           });
+          this.#setHasCalibrationTopic(true);
```
snosenzo, line 486: What about the case that someone has selected none as their calibration and is switching images, but now they have to set their calibration topic to None each time they switch images? Maybe it's an edge use-case we don't want to support 🤷

snosenzo (approved):

snosenzo:
Also story for this behavior?

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageMode.ts
jtbandes: probably not a big deal I think? We can change it if someone complains. But to fix that we might need to check the auto-selected topic of the old topic as well, and see if it matches the current config.

PR #6009, opened 2023-05-12
Add setCurrentLayoutData to CurrentLayoutContext

defunctzombie:
**User-Facing Changes**
None

**Description**
setCurrentLayoutData allows setting the data and replaces any existing data for the active layout.


foxymiles (commented):
What is this used for?

defunctzombie:
> What is this used for?

Nothing yet but it will be used for new code paths for layout storage.

jtbandes (approved):
💮

defunctzombie:
Based on my latest spike I don't think I want this as written.

PR #6015, opened 2023-05-12
Added possibility to download video from image panels and 3d panel

hannaSyrmia:
**Description**

Was added possibility to download video from image panels and 3d pane


defunctzombie:
@hannaSyrmia Thanks for sharing your use case and making an example. We are not ready to commit to this feature as implemented here at this time but have noted this in our overall product feedback.

PR #6025, opened 2023-05-15
Fix LinePrimitive memory usage and performance

snosenzo:
**User-Facing Changes**

 - Fix LinePrimitive memory usage and performance

**Description**
- renderable lines should now dispose properly and not make new buffer geometries every render. Potentially causing memory issues and crashes in chrome


packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableLines.ts
```diff
@@ -78,187 +134,244 @@ export class RenderableLines extends RenderablePrimitive {
...
+        rgbToThreeColor(color, singleColor);
+        // material.opacity = singleColor.a; // does not work for some reason
+        this.#material.uniforms.opacity!.value = singleColor.a;
+      }
     }
 
-    // Set an explicit instance count, because three.js ignores attribute offsets when
-    // automatically computing the instance count (and results differ across browsers because they
-    // depend on the key iteration order, since three.js derives the count from the first
-    // instanced interleaved attribute it sees).
```
jtbandes, line 145: I think this comment is useful to keep somewhere?

packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableLines.ts
```diff
@@ -20,54 +21,109 @@ import { LayerSettingsEntity } from "../../settings";
...
+      this.#updateLines(entity.lines);
+    }
+  }
+
+  public updateSettings(settings: LayerSettingsEntity): void {
+    this.update(this.userData.topic, this.userData.entity, settings, this.userData.receiveTime);
+  }
+}
+
+class LineGroup extends THREE.Group {
```
jtbandes, line 92: Any reason to extend Group instead of Object3D?

packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableLines.ts
```diff
@@ -20,54 +21,109 @@ import { LayerSettingsEntity } from "../../settings";
...
+
+  #material: LineMaterial;
+  #pickingMaterial: THREE.ShaderMaterial;
+  #transparent: boolean = true;
+  #line: LineSegments2 | Line2 | undefined;
+
+  #canvasSizeChanged: boolean = true;
+  #canvasSize: THREE.Vector2;
+
+  #primitiveChanged: boolean = true;
```
jtbandes, line 105: I think we have more of these named `NeedsUpdate` vs `Changed`? I guess we're not super consistent right now

packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableLines.ts
```diff
@@ -78,187 +134,244 @@ export class RenderableLines extends RenderablePrimitive {
...
         dashSize: { value: 1 },
         gapSize: { value: 1 },
       },
       defines: !primitive.scale_invariant ? { WORLD_UNITS: "" } : {},
     });
+  }
 
-    let line: LineSegments2;
+  public setSettings(settings: LayerSettingsEntity): void {
+    this.#colorChanged = this.#color !== settings.color;
```
jtbandes, line 148: Probably should use `||=` for these, otherwise two calls in a row with the same value will set it to `false`.

packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableLines.ts
```diff
@@ -78,187 +134,244 @@ export class RenderableLines extends RenderablePrimitive {
...
+        isLoop,
+      });
+
+      getPositions(this.#positionBuffer, this.#primitive);
+      this.#geometry.setPositions(this.#positionBuffer);
     }
 
-    const positions = getPositions(primitive);
-    // setPosition requires the position array to be >= 6 length or else it will error
-    // we skip primitives with empty points before calling this function
```
jtbandes, line 109: another comment that might be useful to keep - didn't we have some pain in debugging that led us to add this comment?

packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableLines.ts
```diff
@@ -78,187 +134,244 @@ export class RenderableLines extends RenderablePrimitive {
...
+      this.#pickingMaterial.defines = {};
     }
+    this.#pickingMaterial.uniformsNeedUpdate = true;
+    this.#pickingMaterial.needsUpdate = true;
   }
 
-  public updateSettings(settings: LayerSettingsEntity): void {
-    this.update(this.userData.topic, this.userData.entity, settings, this.userData.receiveTime);
+  public dispose(): void {
+    this.remove(this.#line!);
```
jtbandes, line 287: why `!` and not `if (this.#line) this.remove(`? Or `this.#line?.removeFromParent()`?

packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableLines.ts
```diff
@@ -78,187 +134,244 @@ export class RenderableLines extends RenderablePrimitive {
...
+function getLineInstanceCount({
+  pointsLength,
+  isLoop,
+  isSegments,
+}: {
+  pointsLength: number;
+  isLoop: boolean;
+  isSegments: boolean;
+}) {
+  return isSegments ? pointsLength >>> 1 : isLoop ? pointsLength : Math.max(pointsLength - 1, 0);
```
jtbandes, line 376: since this is now a separate function, might as well expand these to `if` statements

packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableLines.ts
```diff
@@ -78,187 +134,244 @@ export class RenderableLines extends RenderablePrimitive {
...
+  public setPrimitive(primitive: LinePrimitive): void {
+    this.#primitiveChanged = this.#primitive !== primitive;
+    this.#primitive = primitive;
+  }
+
+  public setCanvasSize(canvasSize: THREE.Vector2): void {
+    this.#canvasSizeChanged =
+      this.#canvasSize.x !== canvasSize.x || this.#canvasSize.y !== canvasSize.y;
+    this.#canvasSize = canvasSize;
+  }
```
jtbandes, line 161: I don't think this is necessary. The Renderer automatically updates the `resolution` uniform across the whole scene graph: 

https://github.com/foxglove/studio/blob/70177c7519a33153989eb99c935dc8ab65368f7e/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts#L1278

jtbandes (approved):
Thanks for fixing!

packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableLines.ts
snosenzo: I think that if we're keeping track of a variable in the instance, then "Changed" is better, but if we're coding those to update geometry, material or mesh then we'd use something like `materialNeedsUpdate` or `geometryNeedsUpdate`. This `update` function is a little less cut and dry than that unfortunately, so I'm just keeping track of variables that've changed.

packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableLines.ts
```diff
@@ -78,187 +129,246 @@ export class RenderableLines extends RenderablePrimitive {
       uniforms: {
         objectId: { value: [NaN, NaN, NaN, NaN] },
         linewidth: { value: primitive.thickness },
-        resolution: { value: this.renderer.input.canvasSize },
```
jtbandes, line 81: intentionally deleted? I think we might need to keep the initial value around since the renderer only updates it if it changes.

snosenzo: Oh yeah good catch! added back.

packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableLines.ts
```diff
@@ -78,187 +131,256 @@ export class RenderableLines extends RenderablePrimitive {
...
+
+  public update(): void {
+    if (this.#primitive == undefined) {
+      this.visible = false;
+      return;
+    }
+    this.visible = true;
+    const isLoop = this.#primitive.type === LineType.LINE_LOOP;
+    const isSegments = this.#primitive.type === LineType.LINE_LIST;
+    const pointsLength = isLoop ? this.#primitive.points.length + 1 : this.#primitive.points.length;
```
jtbandes, line 162: Do we need to account for `indices` here? It looks like that may have been dropped from the `getColors` logic where it used to allocate the Float32Array. (Let's make sure we get some test data/stories that use `indices` if we don't have them already.)

packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableLines.ts
```diff
@@ -78,187 +131,279 @@ export class RenderableLines extends RenderablePrimitive {
...
+          this.#colorBuffer.length < newPointsColorBufferLength
+        ) {
+          this.#colorBuffer = new Float32Array(newPointsColorBufferLength);
+        }
+        this.#material.vertexColors = true;
+        this.#material.opacity = 1;
+        this.#material.uniforms.opacity!.value = 1;
+        const colors = useIndices
+          ? serializeColorValuesWithIndices(this.#colorBuffer, this.#primitive)
+          : serializeColorValues(this.#colorBuffer, this.#primitive);
```
jtbandes, line 249: this should apply to positions as well

snosenzo: thanks for catching. Updated

packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableLines.ts
snosenzo: yeah I totally missed the indices logic : / . Thanks for catching. 

> Let's make sure we get some test data/stories that use indices if we don't have them already.
So this is accounted for in the story, and I think it catches everything except when the `points` and `colors` arrays differ from `indices`. Which is why it wasn't catching my changes before.

packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableLines.ts
```diff
@@ -78,187 +131,287 @@ export class RenderableLines extends RenderablePrimitive {
...
+      this.visible = false;
+      return;
+    }
+    this.visible = true;
+    const isLoop = this.#primitive.type === LineType.LINE_LOOP;
+    const isSegments = this.#primitive.type === LineType.LINE_LIST;
+    const useIndices = this.#primitive.indices.length > 0;
+    const indicesLength = isLoop
+      ? this.#primitive.indices.length + 1
+      : this.#primitive.indices.length;
```
jtbandes, line 165: nit: could move the `+ (isLoop?1:0)` onto the `numPointsForRendering` variable.

packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableLines.ts
```diff
@@ -78,187 +131,287 @@ export class RenderableLines extends RenderablePrimitive {
...
+    const pointsLength = isLoop ? this.#primitive.points.length + 1 : this.#primitive.points.length;
+
+    const numPointsForRendering = useIndices ? indicesLength : pointsLength;
```
jtbandes, line 168: naming: `numVertices`?

packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableLines.ts
```diff
@@ -78,187 +131,287 @@ export class RenderableLines extends RenderablePrimitive {
...
+
+        if (this.#colorBuffer == undefined || this.#colorBuffer.length < necessaryColorBufferSize) {
+          this.#colorBuffer = new Float32Array(necessaryColorBufferSize);
+        }
+        this.#material.vertexColors = true;
+        this.#material.opacity = 1;
+        this.#material.uniforms.opacity!.value = 1;
+        const colors = useIndices
+          ? serializeColorValuesWithIndices(this.#colorBuffer, this.#primitive)
+          : serializeColorValues(this.#colorBuffer, this.#primitive);
```
jtbandes, line 252: nit: don't really need these functions to return the same array

packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableLines.ts
```diff
@@ -78,187 +131,287 @@ export class RenderableLines extends RenderablePrimitive {
...
-    geometry.instanceCount = isSegments
-      ? primitive.points.length >>> 1
-      : isLoop
-      ? primitive.points.length
-      : Math.max(primitive.points.length - 1, 0);
-
-    line.userData.pickingMaterial = pickingMaterial;
-    return line;
+    if (this.#colorChanged || this.#primitiveChanged) {
+      this.#updateMaterial();
```
jtbandes, line 276: nit: looks like this condition is the same as the prior one, line 276 could be merged in there

packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableLines.ts
```diff
@@ -78,187 +131,287 @@ export class RenderableLines extends RenderablePrimitive {
...
+    this.#pickingMaterial.dispose();
   }
 }
 
-/**
- * Converts (x,y,z) values specified in `primitive.points` (and possibly `primitive.indices`) into
- * vertices for LineGeometry or LineSegmentsGeometry.
- */
-function getPositions(primitive: LinePrimitive): Float32Array {
+function serializePositionValues(
```
jtbandes, line 310: naming nit: maybe just serializePositions/serializeColors? I think "values" is not adding much clarity

packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableLines.ts
```diff
@@ -78,187 +131,287 @@ export class RenderableLines extends RenderablePrimitive {
...
-  if (isLoop && colors.length > 4) {
-    colors[colors.length - 4] = colors[0]!;
-    colors[colors.length - 3] = colors[1]!;
-    colors[colors.length - 2] = colors[2]!;
-    colors[colors.length - 1] = colors[3]!;
+  return colorsOut;
+}
+
+/** Add numItems elements from the beginning of the array onto the end at the specified index (in place)*/
+function loopArrayAtIndex(arrayOut: Float32Array, index: number, numItems: number): Float32Array {
```
jtbandes, line 406: Actually just remembered there's a `copyWithin` method you could use for this. You could probably even delete this function and just call `copyWithin` inline where it's needed.

jtbandes (approved):

PR #6030, opened 2023-05-15
[3D] Remove PointClouds and LaserScans renderables when setting visible=false

jtbandes:
**User-Facing Changes**
Fixed an issue where the last scan on a LaserScan topic would remain visible when the topic was hidden.

**Description**
The root cause was that the LaserScans extension did not set `visible=false` when the setting changed. It did call `clearHistory()`, but that method leaves the most recent entry in the array (intentionally).

Changed LaserScans and PointClouds to dispose hidden renderables in startFrame, instead of keeping them around.

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/pointExtensionUtils.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/pointExtensionUtils.ts
@@ -435,12 +435,10 @@ export class RenderObjectHistory<ParentRenderable extends Renderable<RenderObjec
 
   /** Removes all but the last renderable, which would be the current object used in rendering. */
   public clearHistory(): void {
-    for (let i = 1; i < this.#history.length; i++) {
-      const entry = this.#history[i]!;
+    for (const entry of this.#history.splice(0, this.#history.length - 1)) {
       entry.object3d.geometry.dispose();
       this.#renderable.remove(entry.object3d);
     }
-    this.#history.length = 1;
   }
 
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/panels/ThreeDeeRender/renderables/pointExtensionUtils.ts
```diff
@@ -426,22 +426,27 @@ export class RenderObjectHistory<ParentRenderable extends Renderable<RenderObjec
...
+      throw new Error("RenderObjectHistory is empty");
+    }
+    return this.#history[this.#history.length - 1]!;
   }
 
+  /** Removes all but the last renderable, which would be the current object used in rendering. */
   public clearHistory(): void {
-    // removes all but the last element of the array, which would be the current object used in rendering
-    for (const entry of this.history.splice(0, this.history.length - 1)) {
+    for (let i = 1; i < this.#history.length; i++) {
```
snosenzo, line 438: Shouldn't this keep the last point cloud instead of the first?

jtbandes: Oof thanks for catching. Adding a unit test 😀

jtbandes: Actually don't think this matters in practice because we always update the attribute buffers after doing this.

snosenzo (approved):

PR #6032, opened 2023-05-15
Notify clients of UserProfileStorageContext on changes

foxymiles:
**User-Facing Changes**
None

**Description**
Invalidate the `getUserProfile` callback on `UserProfileStorageContext` so that clients know when the profile has been updated.

This is needed for logic to determine when to default to the new UI.


2metres (approved):

defunctzombie (changes requested):
The force dependency update is sign that the structure here is wrong.

defunctzombie (changes requested):
The force dependency update is sign that the structure here is wrong.

foxymiles:
> The force dependency update is sign that the structure here is wrong.

Really the fix is to not make this async anymore. I'll handle that separately.

Issue #5935, opened 2023-05-06
3D crashed


263868161:
**Description**

When I use 3D, a crash occurs.Works fine on older versions（1.2.0）.
![image](https://user-images.githubusercontent.com/20036253/236619911-c00f2672-01bf-4ad4-95f7-d8b15d0b89e8.png)
![image](https://user-images.githubusercontent.com/20036253/236619925-9efac4ab-e24e-4567-9efc-297122d3393e.png)
![foxglove_error](https://user-images.githubusercontent.com/20036253/236620857-6f429fde-9528-4f64-b632-8e7a83462d59.gif)

- Version:1.53.0
- Platform:ubuntu20.04
- Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native):

**Steps To Reproduce**

**Expected Behavior**
[demo.zip](https://github.com/foxglove/studio/files/11412196/demo.zip)

defunctzombie:
@263868161 What type of computer are you running this on? What are the graphics drivers you are using? To me this looks like a limitation with whatever graphics drivers you are using on your specific computer.

263868161:
My GPU is nvidia  GEFORCE GTX 1060, gpaphics drivers is "nvidia-driver-525"

263868161:
@defunctzombie Hi,I have uploaded the demo program, can't it run?

jtbandes:
Hi again – we've made some performance improvements in #6034 that should help with your use case. I'm going to close out this issue for now since we haven't heard back recently, but feel free to report back or create a new issue if you are still having problems. (A sample bag/mcap file will be much easier for us to diagnose with than a sample program.)

PR #6034, opened 2023-05-16
Perf improvements for ROS line markers

jtbandes:
**User-Facing Changes**
[3D] Improved performance when rendering ROS MarkerArrays with LINE_STRIP or LINE_LIST markers.

**Description**
Reduce extra reallocation on each frame in RenderableLineList & RenderableLineStrip.
Relates to:
- #5935 
- #6025

Tested manually with this data: [line_perf.mcap.zip](https://github.com/foxglove/studio/files/11482738/line_perf.mcap.zip)

```py
import math
from os import path
from tqdm import tqdm
from foxglove_schemas_protobuf.SceneEntity_pb2 import SceneEntity


    writer = Writer(stream)

    frame_interval_ns = 10_000_000
    nframes = 1000
        t = frame * frame_interval_ns

        npoints = int(10 + 5 * math.sin(2*math.pi*frame/nframes))

        entity = SceneEntity()
        entity.timestamp.FromNanoseconds(t)
        line = entity.lines.add()
        line.thickness = 1
        line.pose.orientation.w = 1
        for i in range(npoints):
            pt = line.points.add()
            col = line.colors.add()
            pt.x = 0.2 * i
            pt.y = math.sin(2*math.pi * i/npoints * 2 * frame/nframes)
            pt.z = 0
            col.r = 1 - frame/nframes
            col.g = i/npoints
            col.b = frame/nframes
            col.a = 1

        writer.write_message("entity", entity, t)

    writer.finish()
    stream.close()
```

And this user script to convert to ROS marker types:

```ts
// The ./types module provides helper types for your Input events and messages.

// Your script can output well-known message types, any of your custom message types, or
// complete custom message types.
//
// Use `Message` to access your data source types or well-known types:
// type Twist = Message<"geometry_msgs/Twist">;
//
// Conventionally, it's common to make a _type alias_ for your script's output type
// and use that type name as the return type for your script function.
// Here we've called the type `Output` but you can pick any type name.
type Output = Message<"visualization_msgs/MarkerArray">;

// These are the topics your script "subscribes" to. Studio will invoke your script function
// when any message is received on one of these topics.
export const inputs = ["entity"];

// Any output your script produces is "published" to this topic. Published messages are only visible within Studio, not to your original data source.
export const output = "/studio_script/output_topic";

// This function is called with messages from your input topics.
// The first argument is an event with the topic, receive time, and message.
// Use the `Input<...>` helper to get the correct event type for your input topic messages.
export default function script(event: Input<"entity">): Output {
  return {
    markers: event.message.lines.map((line) => {
      let count = line.points.length;
      // const type: number = 4;
      const type: number = 5;
      if (count % 2 === 1 && type === 5) {
        count--;
      }
      return {
        header: { seq: 0, frame_id: "", stamp: { sec: 0, nsec: 0 } },
        ns: "",
        id: 0,
        type,
        action: 0,
        pose: {
          position: { x: 0, y: 0, z: 0 },
          orientation: { x: 0, y: 0, z: 0, w: 1 },
        },
        scale: { x: 0.1, y: 1, z: 1 },
        color: line.color,
        lifetime: { sec: 0, nsec: 0 },
        frame_locked: true,
        points: line.points.slice(0, count),
        colors: line.colors.slice(0, count),
        text: "",
        mesh_resource: "",
        mesh_use_embedded_materials: false,
      };
    }),
  };
}
```

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/markers/RenderableLineList.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/markers/RenderableLineList.ts
@@ -21,6 +21,8 @@ export class RenderableLineList extends RenderableMarker {
   #geometry: LineSegmentsGeometry;
   #linePrepass: LineSegments2;
   #line: LineSegments2;
+  #positionBuffer = new Float32Array();
+  #colorBuffer = new Uint8Array();
 
   public constructor(
     topic: string,
@@ -97,8 +99,8 @@ export class RenderableLineList extends RenderableMarker {
     const matLine = this.#line.material as LineMaterial;
     matLine.lineWidth = lineWidth;
 
-    const prevPointsLength = (this.#geometry.attributes.instanceStart?.count ?? 0) * 2;
-    if (pointsLength !== prevPointsLength) {
+    const prevPointsLength = this.#positionBuffer.length / 3;
+    if (pointsLength > prevPointsLength) {
       this.#geometry.dispose();
       this.#geometry = new LineSegmentsGeometry();
       this.#linePrepass.geometry = this.#geometry;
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/panels/ThreeDeeRender/renderables/markers/RenderableLineList.ts
```diff
@@ -114,38 +116,48 @@ export class RenderableLineList extends RenderableMarker {
...
-    this.#geometry.setPositions(linePositions);
+    this.#geometry.setPositions(positions);
+    this.#geometry.instanceCount = pointsLength >>> 1;
   }
 
   #setColors(marker: Marker, pointsLength: number): void {
     // Converts color-per-point to a flattened typed array
-    const rgbaData = new Float32Array(4 * pointsLength);
+    if (4 * pointsLength > this.#colorBuffer.length) {
+      this.#colorBuffer = new Uint8Array(4 * pointsLength);
```
snosenzo, line 138: good call on the Uint8Array

packages/studio-base/src/panels/ThreeDeeRender/renderables/markers/RenderableLineList.ts
```diff
@@ -114,38 +116,48 @@ export class RenderableLineList extends RenderableMarker {
...
+    }
+
+    const colorBuffer = this.#colorBuffer;
     this._markerColorsToLinear(marker, pointsLength, (color, i) => {
       const offset = i * 4;
-      rgbaData[offset + 0] = color[0];
-      rgbaData[offset + 1] = color[1];
-      rgbaData[offset + 2] = color[2];
-      rgbaData[offset + 3] = color[3];
+      colorBuffer[offset + 0] = Math.floor(255 * color[0]);
```
snosenzo, line 157: we don't have to worry about colorspaces or anything doing this right?

snosenzo (approved):

packages/studio-base/src/panels/ThreeDeeRender/renderables/markers/RenderableLineList.ts
jtbandes: Conversion happens inside `_markerColorsToLinear`

PR #6042, opened 2023-05-17
Reload main window when switching UI

foxymiles:
**User-Facing Changes**
None

**Description**
Reload main desktop window when toggling the new UI on and off.

### DIFF
--- a/packages/studio-desktop/src/main/StudioWindow.ts
+++ b/packages/studio-desktop/src/main/StudioWindow.ts
@@ -114,7 +114,7 @@ function getTitleBarOverlayOptions(): TitleBarOverlayOptions {
   return {};
 }
 
-function newStudioWindow(deepLinks: string[] = []): BrowserWindow {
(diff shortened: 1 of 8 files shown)


packages/studio-desktop/src/main/StudioWindow.ts
```diff
@@ -562,6 +533,60 @@ class StudioWindow {
     return StudioWindow.#windowsByContentId.get(id);
   }
 
+  #reloadMainWindow(): void {
```
jtbandes, line 536: It seems weird to me that reloading is done within the StudioWindow class and not outside it – I'd imagine that closing and opening a new window would involve another call to `new StudioWindow()`.

jtbandes (approved):

packages/studio-desktop/src/main/StudioWindow.ts
foxymiles: Yeah that might be cleaner. I think we can remove all this code in the near future.

PR #6045, opened 2023-05-17
Remove muiStyled from TextHighlight component

2metres:
**User-Facing Changes**
None

**Description**
Remove muiStyled from TextHighlight component in favor of tss-react

### DIFF
--- a/packages/studio-base/src/components/TextHighlight.tsx
+++ b/packages/studio-base/src/components/TextHighlight.tsx
@@ -12,9 +12,9 @@
 //   You may not use this file except in compliance with the License.
 
 import fuzzySort from "fuzzysort";
-import { withStyles } from "tss-react/mui";
+import { makeStyles } from "tss-react/mui";
 
-const STextHighlight = withStyles("span", (theme) => ({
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/components/TextHighlight.tsx
```diff
@@ -11,17 +11,17 @@
...
 
-const STextHighlight = muiStyled("span")(
-  ({ theme }) => `
-  .TextHighlight-highlight {
-    color: ${theme.palette.primary.main};
-    font-weight: bold;
-  }
-`,
-);
+const STextHighlight = withStyles("span", (theme) => ({
```
defunctzombie, line 17: Why'd this one go with `withStyles` rather that `useClasses` like you did for the other PRs I've reviewed? Something different about this component?

defunctzombie (approved):

packages/studio-base/src/components/TextHighlight.tsx
2metres: They are essentially the same except on returns a component. Going refactor this one to use makeStyles as I just noticed that there are issues using ref's with the component it returns

PR #6047, opened 2023-05-17
Migrate Tab panel stories to play function instead of onMount

2metres:
**User-Facing Changes**
None

### DIFF
--- a/packages/studio-base/src/panels/Tab/index.stories.tsx
+++ b/packages/studio-base/src/panels/Tab/index.stories.tsx
@@ -12,7 +12,7 @@
 //   You may not use this file except in compliance with the License.
 
 import { StoryObj, StoryFn } from "@storybook/react";
-import { fireEvent, screen, within } from "@storybook/testing-library";
+import { fireEvent, within } from "@storybook/testing-library";
 
 import Panel from "@foxglove/studio-base/components/Panel";
 import PanelLayout from "@foxglove/studio-base/components/PanelLayout";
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/panels/Tab/index.stories.tsx
```diff
@@ -309,206 +266,170 @@ export const WithChosenActiveTab: StoryObj = {
...
+    const canvas = within(canvasElement);
+    const tabs = await canvas.findAllByTestId("toolbar-tab");
+
+    try {
+      fireEvent.dragStart(tabs[0]!);
+      fireEvent.dragOver(tabs[2]!);
+    } catch (error) {
+      // supress the error to prevent Storybooks from failing
+      return;
+    }
```
2metres, line 407: Open to suggestions to do this better, but I think the error is a side effect of the behavior we want from the `fireEvent.dragOver` being rejected

defunctzombie (approved):

packages/studio-base/src/panels/Tab/index.stories.tsx
jtbandes: so this didn't previously fail because the error wasn't thrown out of the play function (it was just uncaught)?

jtbandes (approved):
🎉

packages/studio-base/src/panels/Tab/index.stories.tsx
2metres: I believe that is correct.

PR #6049, opened 2023-05-17
Eliminate white flash on dark mode on initial load

2metres:
**User-Facing Changes**
Removes white flash in Dark mode while app is loading

### DIFF
--- a/packages/studio-base/src/theme/ThemeProvider.tsx
+++ b/packages/studio-base/src/theme/ThemeProvider.tsx
@@ -28,6 +28,9 @@ export default function ThemeProvider({
   useEffect(() => {
     // Trick CodeEditor into sync with our theme
     document.documentElement.setAttribute("data-color-mode", isDark ? "dark" : "light");
+
+    // remove styles set to prevent browser flash on init
+    document.querySelector("#loading-styles")?.remove();
   }, [isDark]);
 
   const { i18n } = useTranslation();
--- a/packages/studio-desktop/src/main/StudioWindow.ts
+++ b/packages/studio-desktop/src/main/StudioWindow.ts
@@ -24,6 +24,7 @@ import {
   APP_BAR_HEIGHT,
   APP_BAR_FOREGROUND_COLOR,
 } from "@foxglove/studio-base/src/components/AppBar/constants";
+import * as palette from "@foxglove/studio-base/src/theme/palette";
 
 import StudioAppUpdater from "./StudioAppUpdater";
 import getDevModeIcon from "./getDevModeIcon";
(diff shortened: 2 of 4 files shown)


2metres:
The Desktop app still has a white flash. It [looks like electron](https://www.electronjs.org/docs/latest/tutorial/dark-mode) will allow us to set the window colorscheme but I'm out of my depth here

packages/studio-desktop/src/main/StudioWindow.ts
```diff
@@ -117,13 +117,17 @@ function getTitleBarOverlayOptions(): TitleBarOverlayOptions {
...
+  const isDark =
+    colorScheme === "dark" || (colorScheme === "system" && nativeTheme.shouldUseDarkColors);
 
   const preloadPath = path.join(app.getAppPath(), "main", "preload.js");
 
   const macTrafficLightInset =
     Math.floor((APP_BAR_HEIGHT - /*button size*/ 12) / 2) - /*for good measure*/ 1;
 
   const windowOptions: BrowserWindowConstructorOptions = {
+    backgroundColor: isDark ? "#15151a" : "#f4f4f5",
```
jtbandes, line 130: Can we get this from our theme palette rather than hard-coding?

packages/studio-web/src/webpackConfigs.ts
```diff
@@ -116,6 +116,22 @@ export const mainConfig =
...
+          margin: 0;
+        }
+        #root {
+          height: 100vh;
+          background-color: #15151a;
+          color: #e1e1e4;
+        }
+        @media (prefers-color-scheme: light) {
+          #root {
+            background-color: #f4f4f5;
```
jtbandes, line 130: Can we get these from our theme palette rather than hard-coding?

Did you consider checking the app preference to see if the user has overridden the color scheme?

2metres: This happens in the split seconda before the app has even initialized not if checking for localStorage values would happen quickly enough to match the flash

packages/studio-web/src/webpackConfigs.ts
```diff
@@ -116,6 +117,22 @@ export const mainConfig =
...
       <link rel="icon" type="image/png" sizes="16x16" href="favicon-16x16.png" />
       <title>Foxglove Studio</title>
+      <style type="text/css" id="loading-styles">
+        body {
+          margin: 0;
+        }
+        #root {
+          height: 100vh;
+          background-color: ${palette.light.background?.default};
+          color: ${palette.light.text?.primary};
```
jtbandes, line 127: Why assign these styles on `#root` instead of `body`?

jtbandes (approved):

packages/studio-web/src/webpackConfigs.ts
2metres: I think is mainly becuase some of our users embed foxglove within another app and setting it on body might interfere with other styles/

jtbandes: Folks who are embedding wouldn't be using this index.html I believe. I suppose it doesn't matter either way though.

packages/studio-base/src/theme/ThemeProvider.tsx
```diff
@@ -28,6 +28,9 @@ export default function ThemeProvider({
   useEffect(() => {
     // Trick CodeEditor into sync with our theme
     document.documentElement.setAttribute("data-color-mode", isDark ? "dark" : "light");
+
+    // remove styles set to prevent browser flash on init
+    document.querySelector("#loading-styles")?.remove();
```
defunctzombie, line 33: Why's it important to remove these?

PR #6059, opened 2023-05-18
Cleanup stories for StateTransition panel

2metres:
**User-Facing Changes**
None

**Description**
Removes onMount in favor of storybook.play


packages/studio-base/src/panels/StateTransitions/index.stories.tsx
```diff
@@ -122,171 +120,83 @@ export default {
...
-    const pauseFrame = useCallback(() => readySignal, [readySignal]);
-    return (
-      <PanelSetup
-        fixture={fixture}
-        pauseFrame={pauseFrame}
-        onMount={() => {
-          const mouseEnterContainer = document.querySelectorAll(
-            "[data-testid~=panel-mouseenter-container",
-          )[0]!;
-          TestUtils.Simulate.mouseEnter(mouseEnterContainer);
```
2metres, line 220: This was failing silently prior to this and was likely related to the old show buttons on hover behavior that we don't have anymore

packages/studio-base/src/panels/StateTransitions/index.stories.tsx
```diff
@@ -122,171 +120,83 @@ export default {
...
+        }}
+      />
+    </PanelSetup>
+  ),
 };
 
 export const WithSettings: StoryObj = {
-  render: function Story() {
-    const readySignal = useReadySignal({ count: 3 });
-    const pauseFrame = useCallback(() => readySignal, [readySignal]);
```
jtbandes, line 162: I don't think it's okay to remove the readySignal and pauseFrame stuff here. This helps ensure the chart is rendered before the screenshot is taken so we don't have flaky stories. If this happened in other recent story PRs I probably missed it when doing a quick review – if so we should probably bring it back in those places too.

PR #6060, opened 2023-05-18
Clean up RawMessages storybook to use play instead of onMount

2metres:
**User-Facing Changes**
None

**Description**
Clean up RawMessages storybook to use play instead of component.onMount

### DIFF
--- a/packages/studio-base/src/panels/RawMessages/index.stories.tsx
+++ b/packages/studio-base/src/panels/RawMessages/index.stories.tsx
@@ -14,6 +14,7 @@
 import { StoryObj } from "@storybook/react";
 
 import RawMessages, { PREV_MSG_METHOD } from "@foxglove/studio-base/panels/RawMessages";
+import { RawMessagesPanelConfig } from "@foxglove/studio-base/panels/RawMessages/types";
 import PanelSetup from "@foxglove/studio-base/stories/PanelSetup";
 
 import {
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/panels/RawMessages/index.stories.tsx
```diff
@@ -28,474 +29,367 @@ import {
...
   diffTopicPath: "/another/baz/enum_advanced",
   diffEnabled: true,
-};
+} as RawMessagesPanelConfig;
+
+const scrollToBottom: StoryObj["play"] = async ({ canvasElement }) => {
+  const scrollContainers = canvasElement.querySelectorAll("[data-testid=panel-scroll-container]");
 
-const scrollToBottom = () => {
-  const scrollContainer = document.querySelectorAll(".Flex-module__scroll___3l7to")[0]!;
```
2metres, line 45: Some dust old here. 🤣 This selector would have worked for 1 local build.

packages/studio-base/src/panels/RawMessages/index.stories.tsx
```diff
@@ -28,474 +29,367 @@ import {
 } from "./fixture";
 
 const noDiffConfig = {
-  expansion: "off",
```
2metres, line 31: Removing as the current diff is based on an incorrect setting
```ts
  expansion?: "all" | "none" | Record<string, "e" | "c">;;
```

defunctzombie (approved):

PR #6064, opened 2023-05-18
Migrate ImageView storybook to play functions

2metres:
**User-Facing Changes**
None

**Description**
Remove onMount in favor of storybook play and fix test


jtbandes (approved):
So much less code....nice

PR #6067, opened 2023-05-19
[Image Beta] Add rotation support

jtbandes:
**User-Facing Changes**
None

**Description**
Adds support for 90° image rotations.

https://github.com/foxglove/studio/assets/14237/8c18b049-66e6-4119-bc1e-34f9d9c33d1b


packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageModeCamera.ts
```diff
@@ -69,55 +92,96 @@ export class ImageModeCamera extends THREE.PerspectiveCamera {
...
     const fy = model.P[5];
+
     // (cx, cy) image center in pixel coordinates
     // for panning we can take offsets from this in pixel coordinates
     const scale = this.getEffectiveScale();
-    const cx = model.P[2] + this.#panOffset.x / scale;
-    const cy = model.P[6] + this.#panOffset.y / scale;
-    const { width, height } = model;
+    let panX, panY;
+    switch (this.#rotation) {
```
snosenzo, line 123: we could probably make a pretty simple math function to do these taking a rotation angle that would apply to both the pan and the top/left/bottom/right instead of having these case statements. not sure how much it's worth it tho

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageModeCamera.ts
```diff
@@ -131,16 +195,23 @@ export class ImageModeCamera extends THREE.PerspectiveCamera {
     if (!this.#model) {
       return 1;
     }
+    let { width: canvasWidth, height: canvasHeight } = this.#canvasSize;
+    if (this.#rotation === 90 || this.#rotation === 270) {
+      const width = canvasWidth;
```
snosenzo, line 200: the canvas isn't being rotated, the image/model is. semantically, shouldn't this be done with the `#model`?

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageModeCamera.ts
```diff
@@ -152,12 +223,18 @@ export class ImageModeCamera extends THREE.PerspectiveCamera {
...
-    const rendererAspect = this.#canvasSize.width / this.#canvasSize.height;
+    let rendererAspect = this.#canvasSize.width / this.#canvasSize.height;
     const imageAspect = imgWidth / fx / (imgHeight / fy);
-    // preserve the aspect ratio
-    const shrinkY =
-      this.#zoomMode === "fit" ? imageAspect > rendererAspect : imageAspect < rendererAspect;
-    if (shrinkY) {
+
+    if (this.#rotation === 90 || this.#rotation === 270) {
+      rendererAspect = 1 / rendererAspect;
```
snosenzo, line 230: related to above: why are we not doing this with image aspect?

snosenzo (approved):

snosenzo:
re: the stories we might want to use a clearer image than the grid of colors. with both zooming and panning it's hard to tell what's correct without looking at other stories

banisadr:
I prefer multi-button over dropdown because it's less clicks and you see all options up front. The only argument i see for drop-down is if its a drop down + optional type any value situation. I have trouble believing that users need to set the rotation to 31.045°, but maybe I'm wrong.

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageModeCamera.ts
jtbandes: Yea my feeling was that doing something clever with `sin`/`cos` would be slower and not really worth it.

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageModeCamera.ts
jtbandes: Basically because aspectZoom.y is applied to the image's y axis, not the canvas's y axis. That's a kinda arbitrary decision but it's how I used aspectZoom.

jtbandes:
I agree the stories are not super clear. I will create a follow-up ticket to add a better story.

PR #6070, opened 2023-05-19
Storybook cleanup Log panel

2metres:
**User-Facing Changes**
None

**Description**
Remove unneeded story and cleanup

### DIFF
--- a/packages/studio-base/src/panels/Log/index.stories.tsx
+++ b/packages/studio-base/src/panels/Log/index.stories.tsx
@@ -12,9 +12,8 @@
 //   You may not use this file except in compliance with the License.
 
 import { StoryObj } from "@storybook/react";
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/panels/Log/index.stories.tsx
```diff
@@ -129,220 +128,164 @@ function makeLongFixture(): Fixture {
...
-  render: () => {
-    return (
-      <PanelSetup fixture={fixture} includeSettings>
-        <Log />
-      </PanelSetup>
-    );
-  },
-};
-
-export const TopicToRender: StoryObj = {
```
jtbandes, line 167: Let's keep this story but add `withSettings`. I think the topicToRender setting used to be visible in the toolbar but now it's in the sidebar settings.

jtbandes (approved):

PR #6074, opened 2023-05-19
[Image Panel Beta] - Synchronize timestamp support

snosenzo:
**User-Facing Changes**

 - [Image Panel Beta] - Synchronize timestamp support

**Description**

Factored out message handling and normalization into `MessageHandler` class to be able to control synchronizations of messages better. This class passes a `state` off to the `ImageMode` and `ImageAnnotations` classes that they can use to update their renderables. This `state` represents current state of messages to render for a given config and set of received messages. The synchronization logic was mostly ported over from `synchronizedAddMessages.ts` with a few tweaks to support Topic & Schema for annotations. 

Still need to:

- [x]  ~add errors for absent subscribed annotations~ made ticket because there's some settings changes that need to be made: https://linear.app/foxglove/issue/FG-3596/add-errors-on-annotations-that-synchronization-is-waiting-for
- [x]   add tests

FG-2686


packages/den/collection/TwoKeyMap.ts
```diff
@@ -39,6 +44,15 @@ export class TwoKeyMap<K1, K2, V> {
     this.#map.clear();
   }
 
+  /** Returns the number of secondary maps ie: the number of maps that contain values*/
+  public size(): number {
+    let size = 0;
+    for (const map2 of this.#map.values()) {
+      size += map2.size;
+    }
+    return size;
```
jtbandes, line 53: Comment seems to not match the implementation.

Since this has to scan through each map2, maybe we should consider tracking size changes incrementally as inserts/deletes are done to avoid having to scan repeatedly?

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageMode.ts
```diff
@@ -132,6 +136,15 @@ export class ImageMode
     this.#camera.setCanvasSize(canvasSize.width, canvasSize.height);
     this.#camera.setZoomMode(renderer.config.imageMode.zoomMode ?? "fit");
 
+    const config = this.#getImageModeSettings();
+    this.#messageHandler = new MessageHandler({
+      calibrationTopic: config.calibrationTopic,
+      imageTopic: config.imageTopic,
+      synchronize: config.synchronize,
+      annotationSubscriptions: config.annotations ?? [],
```
jtbandes, line 144: nit: why rename 1 of the fields but leave the other 3 with the same name?

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageMode.ts
```diff
@@ -397,12 +410,11 @@ export class ImageMode
...
-    //   readonly: true,
-    //   input: "boolean",
-    //   label: "🚧 Synchronize timestamps",
-    //   value: synchronize,
-    // };
+    fields.synchronize = {
+      input: "boolean",
+      label: "Sync timestamps",
+      value: synchronize,
+    };
```
jtbandes, line 417: I've been wondering if this setting should move under the "Image annotations" section...since it mostly applies to those. On the other hand it's nice that being under General makes it more visible 🤷 

Maybe call it "Sync annotations" since it operates on the annotations? Not really sure what name would make this most clear. I guess we can bikeshed in product review later.

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageMode.ts
```diff
@@ -498,9 +510,17 @@ export class ImageMode
...
         this.resetViewModifications();
       }
+      if (config.synchronize !== prevImageModeConfig.synchronize) {
+        this.removeAllRenderables();
+      }
+      this.#messageHandler.setConfig({
+        synchronize: config.synchronize,
+        imageTopic: config.imageTopic,
+        calibrationTopic: config.calibrationTopic,
+      });
```
jtbandes, line 523: could this be setConfig(config)?

Could consider splitting these out to separate setter methods. Or merging the #camera.setZoomMode & other setters into one method. I just find it a little weird that the camera and messageHandler have different-looking APIs.

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageMode.ts
```diff
@@ -519,34 +539,27 @@ export class ImageMode
...
 
-  /** Processes camera info messages and updates state */
-  #handleCameraInfo = (
-    messageEvent: PartialMessageEvent<CameraInfo> & PartialMessageEvent<CameraCalibration>,
+  #updateFromMessageState = (
+    newState: Partial<MessageHandlerState>,
+    oldState: Partial<MessageHandlerState> | undefined,
   ): void => {
+    if (newState.image != undefined && newState.image !== oldState?.image) {
+      this.#handleImageChange(newState.image, newState.image.message as AnyImage);
```
jtbandes, line 547: why is a cast needed here?

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/annotations/ImageAnnotations.ts
```diff
@@ -42,6 +43,7 @@ interface ImageAnnotationsContext {
     handler: (messageEvent: MessageEvent<T>) => void,
   ): void;
   labelPool: LabelPool;
+  messageState: MessageHandler;
```
jtbandes, line 46: naming: messageHandler?

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/annotations/ImageAnnotations.ts
```diff
@@ -205,6 +215,9 @@ export class ImageAnnotations extends THREE.Object3D {
         draft.annotations.push(subscription);
       }
     });
+    this.#context.messageState.setConfig({
+      annotationSubscriptions: this.#context.config().annotations ?? [],
+    });
```
jtbandes, line 220: I can kinda see why this is set separately from the other config fields which are set in ImageMode, but it could be nice to keep them all in one place

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/MessageHandler.ts
```diff
@@ -0,0 +1,308 @@
...
+  ImageMarker as RosImageMarker,
+  ImageMarkerArray as RosImageMarkerArray,
+} from "@foxglove/studio-base/types/Messages";
+
+import { PartialMessageEvent } from "../../SceneExtension";
+import { CompressedImage as RosCompressedImage, Image as RosImage, CameraInfo } from "../../ros";
+
+type MessageEventAnnotationsPair = {
+  // required for setting the original message on the renderable
+  messageEvent: MessageEvent<RosImageMarkerArray | RosImageMarker | FoxgloveImageAnnotations>;
```
jtbandes, line 38:
```suggestion
type NormalizedAnnotations = {
  // required for setting the original message on the renderable
  originalMessage: MessageEvent<RosImageMarkerArray | RosImageMarker | FoxgloveImageAnnotations>;
```

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/MessageHandler.ts
```diff
@@ -0,0 +1,308 @@
...
+  synchronize: boolean;
+};
+
+export type MessageHandlerState = {
+  image?: PartialMessageEvent<AnyImage>;
+  cameraInfo?: CameraInfo;
+  annotationsByTopicSchema: TwoKeyMap<string, string, MessageEventAnnotationsPair>;
+};
+
+export class MessageHandler {
```
jtbandes, line 60: comment?

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/MessageHandler.ts
```diff
@@ -0,0 +1,308 @@
...
+  image?: PartialMessageEvent<AnyImage>;
+  annotationsByTopicSchema: TwoKeyMap<string, string, MessageEventAnnotationsPair>;
+};
+
+type ImageModeConfig = {
+  imageTopic: string | undefined;
+  calibrationTopic: string | undefined;
+  annotationSubscriptions: readonly ImageAnnotationSubscription[];
+  synchronize: boolean;
+};
```
jtbandes, line 52: seems like this could be a `Pick`? Since we already have a type named `ImageModeConfig` in IRenderer, would be nice to use a different name (or just reuse that type)

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/MessageHandler.ts
```diff
@@ -0,0 +1,308 @@
...
+  ) => void)[] = [];
+
+  public constructor(config: ImageModeConfig) {
+    this.#config = {
+      ...config,
+    };
+    this.#state = {
+      annotationsByTopicSchema: new TwoKeyMap(),
+    };
+    this.#tree = new AVLTree<Time, SynchronizationItem>(compareTime);
```
jtbandes, line 77: I found myself wondering why we used Time as the key rather than bigint. Not sure which would be faster – either way we'll probably have to convert back and forth at some point.

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/MessageHandler.ts
```diff
@@ -0,0 +1,308 @@
...
+    this.#listeners.push(listener);
+  }
+
+  public removeListener(
+    listener: (
+      newState: Partial<MessageHandlerState>,
+      oldState: Partial<MessageHandlerState> | undefined,
+    ) => void,
+  ): void {
+    this.#listeners.filter((fn) => fn !== listener);
```
jtbandes, line 94: missing a `this.#listeners =` ?

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/MessageHandler.ts
```diff
@@ -0,0 +1,308 @@
...
+  public handleCompressedImage = (messageEvent: PartialMessageEvent<CompressedImage>): void => {
+    this.#handleImage(messageEvent, normalizeCompressedImage(messageEvent.message));
+  };
+
+  #handleImage(message: PartialMessageEvent<AnyImage>, image: AnyImage) {
+    const normalizedImageMessage = {
+      ...message,
+      message: image,
+    };
+    this.#state.image = normalizedImageMessage;
```
jtbandes, line 120: should this be inside `if (!this.#config.synchronize)`? If synchronize is enabled it seems like we shouldn't update the image immediately.

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/MessageHandler.ts
```diff
@@ -0,0 +1,308 @@
...
+    messageEvent: MessageEvent<FoxgloveImageAnnotations | RosImageMarker | RosImageMarkerArray>,
+  ): void => {
+    const annotations = normalizeAnnotations(messageEvent.message, messageEvent.schemaName);
+
+    if (!annotations) {
+      return;
+    }
+
+    const { topic, schemaName } = messageEvent;
+    if (this.#config.synchronize) {
```
jtbandes, line 155: nit: flip / early return

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/MessageHandler.ts
```diff
@@ -0,0 +1,308 @@
...
+      annotationsByTopicSchema: new TwoKeyMap(),
+    };
+    this.#tree.clear();
+    this.#oldState = undefined;
+  }
+
+  #emitState() {
+    const state = this.#getState();
+    this.#listeners.forEach((fn) => fn(state, this.#oldState));
+    this.#oldState = { ...state };
```
jtbandes, line 241: should we make the states immutable? Not immediately clear why this needs to be shallow cloned.

Also not sure why oldState needs to be an instance variable – can it be a local `const oldState = this.#state`? For that matter I don't see why we are calling `#getState` but not assigning the result to `#state`. Generally I am confused by where the source of truth is supposed to live.

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/MessageHandler.ts
```diff
@@ -0,0 +1,308 @@
...
+
+  #numVisibleAnnotations(): number {
+    return this.#config.annotationSubscriptions.filter(({ settings }) => settings.visible).length;
+  }
+}
+
+/** Find the newest entry where we have everything synchronized */
+function findSynchronizedSetAndRemoveOlderItems(
+  tree: AVLTree<Time, SynchronizationItem>,
+  numActiveAnnotationSubscriptions: number,
```
jtbandes, line 276: Probably could use a `@param` in the above comment to explain what this is for

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/MessageHandler.ts
```diff
@@ -0,0 +1,308 @@
...
+    while (minKey && isLessThan(minKey, validEntry[0])) {
+      tree.shift();
+      minKey = tree.minKey();
+    }
+  }
+
+  return validEntry;
+}
+
+function getTimestampFromImage(image: AnyImage) {
```
jtbandes, line 302: I think we've got this logic in a few places, probably could move it to a shared file and use it elsewhere too.

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageMode.ts
```diff
@@ -95,6 +97,8 @@ export class ImageMode
 
   #imageRenderable: ImageRenderable | undefined;
 
+  #messageHandler: MessageHandler;
```
foxymiles, line 100: Can this be readonly?

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/MessageHandler.ts
```diff
@@ -0,0 +1,308 @@
...
+  image?: PartialMessageEvent<AnyImage>;
+  cameraInfo?: CameraInfo;
+  annotationsByTopicSchema: TwoKeyMap<string, string, MessageEventAnnotationsPair>;
+};
+
+export class MessageHandler {
+  #config: ImageModeConfig;
+  #oldState: Partial<MessageHandlerState> | undefined;
+  #state: MessageHandlerState;
+  #tree: AVLTree<Time, SynchronizationItem>;
```
foxymiles, line 64: Some of these could also be readonly?

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/MessageHandler.ts
snosenzo: technically `oldState` is the `RenderState` whereas `this.#state` is internal handler state of most recent messages. I've updated the names of functions to be more clear around this.

packages/den/collection/TwoKeyMap.ts
```diff
@@ -17,12 +18,14 @@ export class TwoKeyMap<K1, K2, V> {
     if (map2 == undefined) {
       map2 = new Map<K2, V>();
       this.#map.set(key1, map2);
+      this.#size++;
```
jtbandes, line 21: I think in the `else` branch you'll need to check whether `map2` size changes after the set call.

Let's add unit tests for size 😀

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageMode.ts
```diff
@@ -375,14 +383,14 @@ export class ImageMode
...
-      value: config.imageMode.calibrationTopic,
+      value: calibrationTopic,
       options: calibrationTopics,
       error: calibrationTopicError,
     };
     fields.zoomMode = {
       label: "Zoom mode",
       input: "toggle",
-      value: config.imageMode.zoomMode ?? DEFAULT_ZOOM_MODE,
+      value: zoomMode,
```
jtbandes, line 393: FYI conflict with #6085

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageModeCamera.ts
```diff
@@ -13,7 +13,7 @@ const DEFAULT_CAMERA_STATE = {
 
 const MIN_USER_ZOOM = 0.5;
 const MAX_USER_ZOOM = 50;
-export const DEFAULT_ZOOM_MODE = "fit";
+export const DEFAULT_ZOOM_MODE = "fit" as const;
```
jtbandes, line 16: FYI conflict with #6085

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/MessageHandler.ts
```diff
@@ -0,0 +1,306 @@
...
+  oldState: MessageRenderState | undefined,
+) => void;
+
+export class MessageHandler {
+  // settings that should reflect image mode config
+  #config: Immutable<Config>;
+  // last state passed to listeners
+  #oldRenderState: MessageRenderState | undefined;
+  // internal state of last received messages
+  #state: MessageHandlerState;
```
jtbandes, line 76: is there an advantage of having this grouped in a single variable vs having 3 variables `#image`, `#cameraInfo`, `#annotationsByTopicSchema`?

Reading this I find it unclear which variables are used when. I think it would help to rename variables or clarify comments to point out if some variables are only used in unsynchronized mode, or only used in synchronized mode.

Maybe a crazy idea, but one could imagine MessageHandler being an interface, with one concrete implementation for synchronized and one for unsynchronized, rather than having them live in the same class.

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageMode.ts
snosenzo: yeah we should bump this for product review, but I'll be out for this week's

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/annotations/ImageAnnotations.ts
snosenzo: I didn't want to mess with this too much. I think we could make the message handler paradigm something a little more generic that other scene extensions can use, maybe that's the time to move this, but I think having this separate here is fine.

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/MessageHandler.ts
jtbandes: Bump

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/MessageHandler.ts
```diff
@@ -0,0 +1,344 @@
...
+
+export type MessageRenderState = Readonly<Partial<MessageHandlerState>>;
+
+type RenderStateListener = (
+  newState: MessageRenderState,
+  oldState: MessageRenderState | undefined,
+) => void;
+
+export class MessageHandler {
+  // settings that should reflect image mode config
```
jtbandes, line 74: Prefer /** for all these comments

jtbandes (approved):

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageMode.ts
```diff
@@ -700,10 +700,13 @@ export class ImageMode
   }
 
   #getImageModeSettings(): Immutable<ConfigWithDefaults> {
-    return {
-      ...DEFAULT_CONFIG,
-      ...(this.renderer.config.imageMode as ImageModeConfig),
-    };
+    const config = { ...this.renderer.config.imageMode };
+    config.synchronize = config.synchronize ?? DEFAULT_CONFIG.synchronize;
```
jtbandes, line 704: can `??=` work here? I find it's a little shorter to read

jtbandes (commented):
:shipit:

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageMode.ts
```diff
@@ -690,8 +699,14 @@ export class ImageMode
...
-  #getImageModeSettings() {
-    return this.renderer.config.imageMode;
+  #getImageModeSettings(): Immutable<ConfigWithDefaults> {
+    const config = { ...this.renderer.config.imageMode };
+    config.synchronize ??= DEFAULT_CONFIG.synchronize;
+    config.rotation ??= DEFAULT_CONFIG.rotation;
+    config.flipHorizontal ??= DEFAULT_CONFIG.flipHorizontal;
+    config.flipVertical ??= DEFAULT_CONFIG.flipVertical;
+
+    return config as ConfigWithDefaults;
```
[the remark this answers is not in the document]
snosenzo: yeah i agree, will do in next settings PR.

PR #6078, opened 2023-05-21
Avoid showing the welcome dialog when using file -> open local file

defunctzombie:
**User-Facing Changes**
Fixes an issue where the user is first shown the welcome dialog when using File -> Open local file and then shown the file picker.

**Description**
Update the DataSourceDialog to hide itself when the active view is "file". This avoids showing the dialog when a user wants to select a file using the new app menu.

A quirk of the way we are leveraging the dialog for workspace actions is that if the user cancels the file picker they will be brought back to the welcome dialog.

Another quirk with this change is that if you click "Open local file" in the welcome dialog, the dialog goes away for a moment while the open file logic itself runs and you see the file picker. The real solution here is to stop using the dialog as the business logic and move the selection and state transition logic to workspace actions. I view this as a quick-fix to the currently annoying behavior with `File -> Open local file` for new nav users and not a reason to avoid doing the more appropriate code refactor.

### DIFF
--- a/packages/studio-base/src/components/DataSourceDialog/DataSourceDialog.tsx
+++ b/packages/studio-base/src/components/DataSourceDialog/DataSourceDialog.tsx
@@ -69,14 +69,19 @@ export function DataSourceDialog(props: DataSourceDialogProps): JSX.Element {
   useLayoutEffect(() => {
     if (activeView === "file") {
       openFile()
-        .catch((err) => {
-          console.error(err);
-        })
-        .finally(() => {
-          // set the view back to start so the user can click to open file again
-          if (isMounted()) {
+        .then((didOpen) => {
+          // If no file was selected we go back to the start screen.
+          // There is quirk here where if the user cancels the open file dialog they launched
+          // through the file -> open local file menu they will still see the start screen
+          //
+          // This is because we are using the dialog to handle logic which could move to workspace
+          // actions.
+          if (!didOpen && isMounted()) {
             dialogActions.dataSource.open("start");
           }
+        })
+        .catch((err) => {
+          console.error(err);
         });
     } else if (activeView === "demo" && firstSampleSource) {
       selectSource(firstSampleSource.id);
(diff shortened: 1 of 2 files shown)


jtbandes (commented):
My general feeling is that triggering the open file action from the menu should not need to involve the OpenDialog. Adding extra code paths inside the dialog to support this while not actually rendering the dialog seems much more awkward than just adding a direct flow from menu item->show file picker->create player.

defunctzombie:
@jtbandes I agree. Did we overlook this insight when the workspace actions were added? That would have been the best time to address this nuance rather than end up in the current situation. The second best time is now.

foxymiles:
> @jtbandes I agree. Did we overlook this insight when the workspace actions were added? That would have been the best time to address this nuance rather than end up in the current situation. The second best time is now.

All the workspace actions did was hoist a few state variables from the `Workspace` component itself into the store. Nothing inside the business logic of the dialog itself was changed.

foxymiles:
We've had this quirk as far back at least as 1.40 before any of this stuff was in the workspace store. You can see if you use the file menu in the desktop app:

https://github.com/foxglove/studio/assets/93935560/a484cb7e-259e-46bb-8a8a-6aa52cfcff02

I agree with @jtbandes that the longer term fix is to move this logic out of the DataSourceDialog but as a workaround we could add another flag to the data source dialog store state to tell it it's being used in this way and not reappear when the file dialog is canceled.

jtbandes:
I think if we're looking to solve this problem now, there's no reason to go with a quicker workaround and I'd rather see us implement the desired solution, whatever it may be.

jtbandes (changes requested):
see above

foxymiles:
Alternative solution: https://github.com/foxglove/studio/pull/6100

defunctzombie:
Closing in favor of alternatives. Agree that there's a better way to tackle this than this hack.

PR #6079, opened 2023-05-22
Expose our own Immutable type in @foxglove/studio

foxymiles:
**User-Facing Changes**
None

**Description**
Expose our own Immutable type in @foxglove/studio. Immutability is an important concept for correctness, performance, and for clarity of intentions in API definitions. It's worth having our own internal type for this.

Once this is merged we should take advantage of it in our public APIs to make it clearer what data we don't expect extensions to modify.

In this PR I'm just re-exporting the definition from immerjs but if we want it to be standalone I think we could also just copy over the definition itself with the MIT license and an attribution in the header or grab an equivalent type from ts-essentials.


jtbandes (approved):
No strong opinion on adding the dependency from me

PR #6080, opened 2023-05-22
Implement topic mapping extensions

foxymiles:
**User-Facing Changes**
Implement topic mapping extensions that allow topics to be renamed programatically. 

**Description**
Implement topic mapping extensions that allow topics to be renamed programatically. This adds a new extension type that extensions can register like this:

```typescript
type RegisterTopicMapperArgs = (
  topics: ReadonlyArray<{ name: string; schemaName?: string }>,
) => ReadonlyMap<string, string>;

registerTopicMapper: (args: RegisterTopicMapperArgs);
```

These extensions are single functions that take as input an array of topics and return a `Map<string,string>` that will be used to rename the topics after they are emitted from the core player but before any of the rest of the application uses them. This is accomplished by inserting a new `TopicMappingPlayer` between the core player and the `UserNodePlayer`. The advantage of this approach is that downstream users of player data can use the mapped topics without any changes or awareness that the mapping is going on. If the topic list changes extensions will run again on the new topic list and have a change to update they mappings dynamically.

I've tried hard here to have no overhead for users that don't have any mapping extensions installed.  This could also have been implemented as something attached to a layout but that would make the mappings difficult to update and reuse across an organization.

This PR could also use some design input for indications to the user that a topic was renamed. I'm showing the original topic name, if any, in small type in the data source info panel but we could probably improve this.

One somewhat unsatisfactory aspect of this implementation is that we have to wait until the player emits a topic list to calculate the mappings but in an existing layout we try to call `setSubscriptions` on the player before this happens. I've tried to solve this by storing the requested subscriptions and then applying them once we do have a topic list but this seems like the most fragile element of this approach.

### DIFF
--- a/packages/studio-base/src/components/PlayerManager.tsx
+++ b/packages/studio-base/src/components/PlayerManager.tsx
@@ -12,7 +12,14 @@
 //   You may not use this file except in compliance with the License.
 
 import { useSnackbar } from "notistack";
-import { PropsWithChildren, useCallback, useLayoutEffect, useMemo, useState } from "react";
+import {
+  PropsWithChildren,
+  useCallback,
+  useEffect,
+  useLayoutEffect,
+  useMemo,
+  useState,
+} from "react";
 import { useLatest, useMountedState } from "react-use";
 
 import { useShallowMemo } from "@foxglove/hooks";
@@ -24,6 +31,10 @@ import {
   useCurrentLayoutActions,
   useCurrentLayoutSelector,
 } from "@foxglove/studio-base/context/CurrentLayoutContext";
+import {
+  ExtensionCatalog,
+  useExtensionCatalog,
+} from "@foxglove/studio-base/context/ExtensionCatalogContext";
 import { useLayoutManager } from "@foxglove/studio-base/context/LayoutManagerContext";
 import { useNativeWindow } from "@foxglove/studio-base/context/NativeWindowContext";
 import PlayerSelectionContext, {
@@ -36,6 +47,7 @@ import { GlobalVariables } from "@foxglove/studio-base/hooks/useGlobalVariables"
 import useIndexedDbRecents, { RecentRecord } from "@foxglove/studio-base/hooks/useIndexedDbRecents";
 import useWarnImmediateReRender from "@foxglove/studio-base/hooks/useWarnImmediateReRender";
 import AnalyticsMetricsCollector from "@foxglove/studio-base/players/AnalyticsMetricsCollector";
+import { TopicMappingPlayer } from "@foxglove/studio-base/players/TopicMappingPlayer/TopicMappingPlayer";
 import UserNodePlayer from "@foxglove/studio-base/players/UserNodePlayer";
 import { Player } from "@foxglove/studio-base/players/types";
 import { UserNodes } from "@foxglove/studio-base/types/panels";
@@ -53,6 +65,7 @@ const userNodesSelector = (state: LayoutState) =>
   state.selectedLayout?.data?.userNodes ?? EMPTY_USER_NODES;
 const globalVariablesSelector = (state: LayoutState) =>
   state.selectedLayout?.data?.globalVariables ?? EMPTY_GLOBAL_VARIABLES;
+const selectTopicMappers = (catalog: ExtensionCatalog) => catalog.installedTopicMappers;
 
 export default function PlayerManager(props: PropsWithChildren<PlayerManagerProps>): JSX.Element {
   const { children, playerSources } = props;
@@ -83,6 +96,9 @@ export default function PlayerManager(props: PropsWithChildren<PlayerManagerProp
   const userNodes = useCurrentLayoutSelector(userNodesSelector);
   const globalVariables = useCurrentLayoutSelector(globalVariablesSelector);
 
+  const topicMappers = useExtensionCatalog(selectTopicMappers);
+  const [initialTopicMappers] = useState(topicMappers);
+
   const { recents, addRecent } = useIndexedDbRecents();
 
   // We don't want to recreate the player when the these variables change, but we do want to
@@ -93,15 +109,34 @@ export default function PlayerManager(props: PropsWithChildren<PlayerManagerProp
   // the message pipeline
   const globalVariablesRef = useLatest(globalVariables);
 
-  const player = useMemo(() => {
+  // Initialize the topic mapping player with the mappers we have at first load. Any
+  // changes in mappers caused by dynamically loaded extensions have to be set separately
+  // because we can only construct the wrapping player once since the underlying player
+  // doesn't allow us set a new listener after the initial listener is set.
+  const topicMapper = useMemo(() => {
     if (!basePlayer) {
       return undefined;
     }
 
-    const userNodePlayer = new UserNodePlayer(basePlayer, userNodeActions);
+    return new TopicMappingPlayer(basePlayer, initialTopicMappers ?? []);
+  }, [basePlayer, initialTopicMappers]);
+
+  // Topic mappers can change if we hot load a new extension with new mappers.
+  useEffect(() => {
+    if (topicMappers !== initialTopicMappers) {
+      topicMapper?.setMappers(topicMappers ?? []);
+    }
+  }, [initialTopicMappers, topicMapper, topicMappers]);
+
+  const player = useMemo(() => {
+    if (!topicMapper) {
+      return undefined;
+    }
+
+    const userNodePlayer = new UserNodePlayer(topicMapper, userNodeActions);
     userNodePlayer.setGlobalVariables(globalVariablesRef.current);
     return userNodePlayer;
-  }, [basePlayer, globalVariablesRef, userNodeActions]);
+  }, [globalVariablesRef, topicMapper, userNodeActions]);
 
   useLayoutEffect(() => void player?.setUserNodes(userNodes), [player, userNodes]);
 
--- a/packages/studio-base/src/panels/DataSourceInfo/index.tsx
+++ b/packages/studio-base/src/panels/DataSourceInfo/index.tsx
@@ -2,7 +2,7 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
-import { Divider } from "@mui/material";
+import { Divider, Typography } from "@mui/material";
 import { makeStyles } from "tss-react/mui";
 
 import CopyButton from "@foxglove/studio-base/components/CopyButton";
@@ -77,6 +77,11 @@ function TopicRow({ topic }: { topic: Topic }): JSX.Element {
           iconSize="small"
           getText={() => topic.name}
         />
+        {topic.mappedFromName && (
+          <Typography variant="subtitle2" fontSize="0.5rem">
+            from {topic.mappedFromName}
+          </Typography>
+        )}
       </td>
       <td>
         {topic.schemaName == undefined ? (
--- a/packages/studio-base/src/players/TopicMappingPlayer/TopicMappingPlayer.ts
+++ b/packages/studio-base/src/players/TopicMappingPlayer/TopicMappingPlayer.ts
@@ -1,0 +1,127 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { Immutable as Im } from "immer";
+
+import { Time } from "@foxglove/rostime";
+import { ParameterValue, RegisterTopicMapperArgs } from "@foxglove/studio";
+import { GlobalVariables } from "@foxglove/studio-base/hooks/useGlobalVariables";
+import {
+  AdvertiseOptions,
+  Player,
+  PlayerState,
+  PublishPayload,
+  SubscribePayload,
+} from "@foxglove/studio-base/players/types";
+
+import { MappingInputs, mapPlayerState, mapSubscriptions } from "./mapping";
+
+export class TopicMappingPlayer implements Player {
+  readonly #player: Player;
+
+  #inputs: Im<MappingInputs>;
+  #pendingSubscriptions: undefined | SubscribePayload[];
+  #skipMapping: boolean;
+
+  #listener?: (state: PlayerState) => Promise<void>;
+
+  public constructor(player: Player, mappers: readonly RegisterTopicMapperArgs[]) {
+    this.#player = player;
+    this.#skipMapping = mappers.length === 0;
+    this.#inputs = {
+      mappers,
+      topics: undefined,
+    };
+  }
+
+  public setListener(listener: (playerState: PlayerState) => Promise<void>): void {
+    this.#listener = listener;
+
+    this.#player.setListener(async (state) => await this.#onPlayerState(state));
+  }
+
+  public setMappers(mappers: readonly RegisterTopicMapperArgs[]): void {
+    this.#inputs = { mappers, topics: this.#inputs.topics };
+    this.#skipMapping = mappers.length === 0;
+  }
+
+  public close(): void {
+    this.#player.close();
+  }
+
+  public setSubscriptions(subscriptions: SubscribePayload[]): void {
+    if (this.#skipMapping) {
+      this.#player.setSubscriptions(subscriptions);
+    } else {
+      // If we have mappers but haven't yet recieved a topic list from an active state
+      // from the wrapped player yet we have to delay setSubscriptions until we have the
+      // topic list to set up the mappings.
+      if (this.#inputs.topics != undefined) {
+        const mappedSubscriptions = mapSubscriptions(this.#inputs, subscriptions);
+        this.#player.setSubscriptions(mappedSubscriptions);
+        this.#pendingSubscriptions = undefined;
+      } else {
+        this.#pendingSubscriptions = subscriptions;
+      }
+    }
+  }
+
+  public setPublishers(publishers: AdvertiseOptions[]): void {
+    this.#player.setPublishers(publishers);
+  }
+
+  public setParameter(key: string, value: ParameterValue): void {
+    this.#player.setParameter(key, value);
+  }
+
+  public publish(request: PublishPayload): void {
+    this.#player.publish(request);
+  }
+
+  public async callService(service: string, request: unknown): Promise<unknown> {
+    return await this.#player.callService(service, request);
+  }
+
+  public startPlayback?(): void {
+    this.#player.startPlayback?.();
+  }
+
+  public pausePlayback?(): void {
+    this.#player.pausePlayback?.();
+  }
+
+  public seekPlayback?(time: Time, backfillDuration?: Time | undefined): void {
+    this.#player.seekPlayback?.(time, backfillDuration);
+  }
+
+  public playUntil?(time: Time): void {
+    this.#player.playUntil?.(time);
+  }
+
+  public setPlaybackSpeed?(speedFraction: number): void {
+    this.#player.setPlaybackSpeed?.(speedFraction);
+  }
+
+  public setGlobalVariables(globalVariables: GlobalVariables): void {
+    this.#player.setGlobalVariables(globalVariables);
+  }
+
+  async #onPlayerState(playerState: PlayerState) {
+    if (this.#skipMapping) {
+      await this.#listener?.(playerState);
+      return;
+    }
+
+    if (playerState.activeData?.topics !== this.#inputs.topics) {
+      this.#inputs = { ...this.#inputs, topics: playerState.activeData?.topics };
+    }
+
+    const newState = mapPlayerState(this.#inputs, playerState);
+    await this.#listener?.(newState);
+
+    if (this.#pendingSubscriptions && this.#inputs.topics) {
+      this.setSubscriptions(this.#pendingSubscriptions);
+    }
+  }
+}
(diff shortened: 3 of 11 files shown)


packages/studio/src/index.ts
```diff
@@ -410,13 +410,19 @@ export type RegisterMessageConverterArgs<Src> = {
   converter: (msg: Src) => unknown;
 };
 
+export type RegisterTopicMapperArgs = (
+  topics: ReadonlyArray<{ name: string; schemaName?: string }>,
```
foxymiles, line 414: Note that I've used this inline definition of a topic because the `Topic` we define in this file includes message converter fields that are not present at this extension point.

defunctzombie:
> but in an existing layout we try to call setSubscriptions on the player before this happens

In what situation are we doing that? It seems like thats true of some panels but not generally true across all? So the statement here is that this creates even more of a two-step process for subscribing? Panels cannot subscribe without a known data source loaded since they don't know what topics to subscribe to yet?

packages/studio/src/index.ts
defunctzombie: Might be worth considering a `TopicBase` or similar type to cleanup the readability.

packages/studio/src/index.ts
```diff
@@ -410,13 +410,20 @@ export type RegisterMessageConverterArgs<Src> = {
...
+
 export interface ExtensionContext {
   /** The current _mode_ of the application. */
   readonly mode: "production" | "development" | "test";
 
   registerPanel(params: ExtensionPanelRegistration): void;
 
   registerMessageConverter<Src>(args: RegisterMessageConverterArgs<Src>): void;
+
+  registerTopicMapper(args: RegisterTopicMapperArgs): void;
```
defunctzombie, line 426: For some sad reason I failed to do this for the `regiserMessageConverter` and `registerPanel` but lets add a comment to this function that describes what it does - when it is called - etc. A user-facing documentation summary of your PR would be a good start. Notable things to call-out are when the mapper function will run.

Reading the types here. `RegisterTopicMapperArgs` had me tripped up cause its not actually _args_ like the other functions. You are actually registering a topic mapper (a function) - and that function accepts parameters.

I'm thinking a better type name is `TopicMapper`? Right now the type names used definitely made me do a double take. This is also a good place to document what the return type should be or any notable semantics.

defunctzombie:
Some questions around semantics and architecture:

- What's the argument for doing this before UserNodePlayer vs after (in extension adapter or message pipeline)?
- What happens if multiple topics are mapped to the same output topic? Across different mappers?
- If a mapped topic takes on the schema of its original topic, can I run converters on the mapped topic?
- If a mapped topic takes on the schema of its original topic, how obvious is it if we create situations where users map something like a transform topic and the 3d panel subscribes to all topics with transform schemas?

foxymiles:
> > but in an existing layout we try to call setSubscriptions on the player before this happens
> 
> In what situation are we doing that? It seems like thats true of some panels but not generally true across all? So the statement here is that this creates even more of a two-step process for subscribing? Panels cannot subscribe without a known data source loaded since they don't know what topics to subscribe to yet?

Yeah I think this is the most difficult aspect of this proposal. Plot panel at least will request subscriptions before the player has had a change to emit a state.

foxymiles:
> Some questions around semantics and architecture:
> 
> * What's the argument for doing this before UserNodePlayer vs after (in extension adapter or message pipeline)?

This is necessary for UserNodePlayer to be able to use the mapped topics.

> * What happens if multiple topics are mapped to the same output topic? Across different mappers?

I guess we probably want to flag this as an error.

> * If a mapped topic takes on the schema of its original topic, can I run converters on the mapped topic?

That should work.

> * If a mapped topic takes on the schema of its original topic, how obvious is it if we create situations where users map something like a transform topic and the 3d panel subscribes to all topics with transform schemas?

Do we have special logic that assumes transforms have names like `tf`? I think otherwise the mapped topic should be transparent to panels.

packages/studio-base/src/providers/ExtensionCatalogProvider.tsx
```diff
@@ -50,7 +58,7 @@ function activateExtension(
   const ctx: ExtensionContext = {
     mode: extensionMode,
 
-    registerPanel(params) {
+    registerPanel: (params) => {
```
defunctzombie, line 61: Are these changes syntactic or do they have a different meaning when written like this?

packages/studio-base/src/players/TopicMappingPlayer/mapping.ts
```diff
@@ -0,0 +1,249 @@
...
+}
+
+/**
+ * Maps an array of subscriptions to a new array with all topic mappings applied.
+ *
+ * @param inputs the inputs to the mapping function
+ * @param subscriptions the subscription payloads to map
+ * @returns a new array of subscription payloads with mapped topic names
+ */
+export const mapSubscriptions = memoizeWeak(
```
defunctzombie, line 227: I continue to not be a fan of memoizeWeak - maybe it is because of our inconsistent application or maybe because react already has a layer for this but I would prefer to see this logic with fewer memoizations on the individual functions or maybe a more obvious reason for why the memoization is happening.

packages/studio-base/src/players/TopicMappingPlayer/mapping.ts
```diff
@@ -0,0 +1,249 @@
...
+  variables: GlobalVariables;
+};
+
+function mapBlocks(blocks: MessageBlocks, mapping: Im<TopicMapping>): MessageBlocks {
+  if (mapping === EmptyMapping) {
+    return blocks;
+  }
+
+  return blocks.map((block) => {
+    return block
```
defunctzombie, line 35: Personal preference but I'd find this easier to read if it bailed early rather than used a ternary.

packages/studio-base/src/players/TopicMappingPlayer/TopicMappingPlayer.ts
```diff
@@ -0,0 +1,131 @@
...
+} from "@foxglove/studio-base/players/types";
+
+import { MappingInputs, mapPlayerState, mapSubscriptions } from "./mapping";
+
+export class TopicMappingPlayer implements Player {
+  readonly #player: Player;
+
+  #inputs: Immutable<MappingInputs>;
+  #pendingSubscriptions: undefined | SubscribePayload[];
+  #skipMapping: boolean;
```
defunctzombie, line 23: A comment on the behavior for this variable would make it clearer to readers in what situation and why this exists.

packages/studio-base/src/players/TopicMappingPlayer/TopicMappingPlayer.ts
```diff
@@ -0,0 +1,131 @@
...
+  AdvertiseOptions,
+  Player,
+  PlayerState,
+  PublishPayload,
+  SubscribePayload,
+} from "@foxglove/studio-base/players/types";
+
+import { MappingInputs, mapPlayerState, mapSubscriptions } from "./mapping";
+
+export class TopicMappingPlayer implements Player {
```
defunctzombie, line 18: Comment documenting the purpose of this class and some non-obvious call-outs if there are any.

packages/studio-base/src/players/TopicMappingPlayer/TopicMappingPlayer.ts
```diff
@@ -0,0 +1,131 @@
...
+    if (this.#skipMapping) {
+      await this.#listener?.(playerState);
+      return;
+    }
+
+    if (playerState.activeData?.topics !== this.#inputs.topics) {
+      this.#inputs = { ...this.#inputs, topics: playerState.activeData?.topics };
+    }
+
+    const newState = mapPlayerState(this.#inputs, playerState);
```
defunctzombie, line 124: My rough read of the flow you are going for with mapPlayerState is that you want it to return a stable reference if the input has not changed? Tho you already check for the same topic reference above but I'm thinking you want to also make this stable if the reference is different but the list of topics itself is the same? If thats the reason for the memo then I think its worth making that behavior more explicit - either with a new structure/wrapper/convention we have or explicitly doing a comparison before calling the mapPlayerState which itself will no longer memo and leaves it up to a caller to handle that behavior.

foxymiles: `onPlayerState` should only be called when the underlying player emits a new state so I'm not trying to handle a case here where it's the same state as the last call. I'm assuming if the source player has sent us a new state then _something_ in that state should be different from the last call.

packages/studio-base/src/players/TopicMappingPlayer/mapping.ts
foxymiles: The react memoization tools can't help us here because the player logic runs outside of the react render loop and can't use hooks. If your concern about memoizeWeak is memory leaks I think it's ok because the weakmaps it uses won't prevent GC if they're the only thing left pointing to data.

packages/studio-base/src/providers/ExtensionCatalogProvider.tsx
foxymiles: Just syntactic. I find that inline object style a bit harder to read.

jtbandes: Just a note that there is a slightly different meaning, `registerPanel(params) { ... }` is the same as `registerPanel: function (params) { ... }` (does not capture `this`)

foxymiles: Yeah that's another reason I tend to prefer this format for non-class objects. If I see the `foo()` syntax I tend to interpret that as something that needs properly scoped `this`.

packages/studio-base/src/players/TopicMappingPlayer/mapping.test.ts
jtbandes, line 1: I think we should also add a more end-to-end test that actually instantiates the TopicMappingPlayer and shows it does what it's supposed to do.

jtbandes (approved):

packages/studio-base/src/players/types.ts
```diff
@@ -211,6 +211,8 @@ export type Topic = {
   name: string;
   // Name of the datatype (see `type PlayerStateActiveData` for details).
   schemaName: string | undefined;
+  // Name of the topic before topic mapping, if any.
+  mappedFromName?: string;
```
defunctzombie, line 215: I don't see this being exposed to extension authors - do we think thats a follow-up? Or not something we want to offer downstream to extensions at all?

Notably - the DataSourceInfo panel is not yet converted to an extension but if it was you would not have been able to make the change there to show this.

Related - do we show the mapping in the topics sidebar listing? That would be a place I would expect to see such information.

packages/studio-base/src/components/PlayerManager.tsx
```diff
@@ -92,15 +107,44 @@ export default function PlayerManager(props: PropsWithChildren<PlayerManagerProp
...
+  }, [initialTopicMappers, topicMapper, topicMappers]);
+
+  // Topic mapper needs updated global variables.
+  useEffect(() => {
+    if (globalVariables !== initialGlobalVariables) {
+      topicMapper?.setGlobalVariables(globalVariables);
+    }
+  }, [globalVariables, initialGlobalVariables, topicMapper]);
+
+  const player = useMemo(() => {
```
defunctzombie, line 139: Can we drop the state for initial mappers and global variables if this returns `{ player, topicMapper }`?

packages/studio-base/src/panels/DataSourceInfo/index.tsx
```diff
@@ -77,6 +77,11 @@ function TopicRow({ topic }: { topic: Topic }): JSX.Element {
           iconSize="small"
           getText={() => topic.name}
         />
+        {topic.mappedFromName && (
```
foxymiles, line 80: @defunctzombie I'm showing the mapped from topic here. Probably ways we could improve the UX here.

packages/studio-base/src/players/types.ts
foxymiles: Good question. I suppose it would be better for extension panel authors to have some insight into this?

packages/studio-base/src/players/TopicMappingPlayer/mapping.ts
defunctzombie: my concern with memoizeWeak is that _to me_ it hide the state management behavior. Rather than explicitly being aware of the reference and update semantics the behavior becomes abstracted behind a "memoization" which can have many uses.

My leanings (can be concinved) is that 90% of the uses of memoizeWeak in this new code need to be dropped and replaced with a more explicit identification of "ok yea we need to do more work" that happens not by globally defined functions with memos but with an instance that is managing that state.

packages/studio-base/src/components/PlayerManager.tsx
foxymiles: Wouldn't we need to recreate the UserNodePlayer when variables or mappers change in that case?

defunctzombie: We'd still call the "setMappers" methods like you have in the effects (we do this for user node player elsewhere this file already).

packages/studio-base/src/players/TopicMappingPlayer/mapping.ts
foxymiles: > My leanings (can be convinced) is that 90% of the uses of memoizeWeak in this new code need to be dropped and replaced with a more explicit identification of "ok yea we need to do more work" that happens not by globally defined functions with memos but with an instance that is managing that state.

That's going to result in way more explicit tracking of state variables and a lot more ways to get it wrong. IMO the code we have in the app that manages large numbers of variables in 100+ line functions is the most difficult to reason about and the code I most often encounter subtle bugs in.

defunctzombie (commented):
I don't feel particularly happy about the _memoizeWeak_ stuff and to me feels like there is a better approach that is not "global memoize functions". Especially around your desire to have stateless functions :D I think my main gripe is that we already have an instance `TopicMappingPlayer` that is tied into lifecycle rules (i.e. torn down when players change) and these memos should also be tied into those lifecycle rules.

I'm also not seeing anything that shows the mapped topic information in the UI - if there's a linear for that and we do that as follow-up I can get behind that but I did not see it.

If we are sending the mapped topics to extensions it seems reasonable to also include the `mappedFrom` information in the topics going to extensions? Also ok if there's a linear for this to avoid jamming it all into one PR.

I would like to go through making an extension for this to experience the flow and UX - is that something you have done and have some sample code and dataset available?

foxymiles:
> I don't feel particularly happy about the _memoizeWeak_ stuff and to me feels like there is a better approach that is not "global memoize functions". Especially around your desire to have stateless functions :D I think my main gripe is that we 

That's the beauty of memoization. You don't have to handle all the caching and data dependencies yourself and inevitably get them wrong in subtle ways you only discover months later. Memoization is a fundamental technique we're already using all over our codebase so I'm not sure why it needs to be singled out here. The other class based state we have in the app in class files with 600+ lines is the least maintainable and testable code we have IMO so I'd really like to avoid more of that.

> I'm also not seeing anything that shows the mapped topic information in the UI - if there's a linear for that and we do that as follow-up I can get behind that but I did not see it.

I'm showing it in the sidebar (linked in comment above).

> If we are sending the mapped topics to extensions it seems reasonable to also include the `mappedFrom` information in the topics going to extensions? Also ok if there's a linear for this to avoid jamming it all into one PR.

Yeah that seems reasonable to me.

> I would like to go through making an extension for this to experience the flow and UX - is that something you have done and have some sample code and dataset available?

Yeah I have a sample extension that works with nuscenes. I can add that as an example in a create-foxglove-extension PR.

foxymiles:
> I would like to go through making an extension for this to experience the flow and UX - is that something you have done and have some sample code and dataset available?

https://github.com/foxglove/create-foxglove-extension/pull/107

This works with nuscenes.

defunctzombie:
Per slack discussion we need to show a player problem (error) if a user creates a topic mapping to an existing real topic or multiple real topics to the same output topic.

defunctzombie:
I tried the latest implementation as of [flag disallowed topic mappings](https://github.com/foxglove/studio/pull/6080/commits/e54f1ce0181af53e8c73801119b3b6913da7b749) and think we still have some gaps in functionality and surprising behavior.

I used the example [here](https://github.com/foxglove/create-foxglove-extension/pull/107) as a starting point and tweaked it. I added support for a "selected_camera" global variable that re-maps a specific camera's info topic to `/selected_camera/camera_info`:

This example extension is meant to pair with our sample data.

```typescript
const varVal = globalVariables["foo"] ?? "back";
const camera = globalVariables["selected_camera"];

return new Map([
  ["/imu", "/pose"], // this will be flagged as an problem by the player
  ["/odom", "/remapped_odom"],
  ["/map", "/remapped_odom"], // this will be flagged as an problem by the player

  ["/CAM_FRONT/image_rect_compressed", "/remapped_cam_front/image_rect_compressed"],
  ["/CAM_FRONT/camera_info", "/remapped_cam_front/camera_info"],
  ["/CAM_BACK/image_rect_compressed", `/remapped_cam_${varVal}/image_rect_compressed`],
  ["/CAM_BACK/camera_info", `/remapped_cam_${varVal}/camera_info`],

  [`/${camera}/image_rect_compressed`, "/selected_camera/image_rect_compressed"],
  [`/${camera}/camera_info`, "/selected_camera/camera_info"],
]);
```

In using this extension here are my thoughts:

1. I tried to make a raw message panel to view the `/odom` topic. I struggled to find my topic anywhere in the app, even though I am 100% certain the `/odom` topic exists in my data!!

> It seems that once a topic is re-mapped the original topic is no longer present. I was not expecting this and would maybe say it is incorrect behavior. I could see us offering an option to the user for controlling this behavior? but for now I expect that my original topics still remain.

2. When looking at topics in the sidebar, there is no indication for the source of a re-mapped topic:
Screenshot 2023-06-17 at 7 05 31 AM

> We should be showing the original topic with the info for a re-mapped topic.

3. To me, the problem text language, `Topic mapping extension requested disallowed mapping from original topic /imu to original topic /pose ` sounds awkward. `original topic` is not how I would describe these topics. I would probably call it a "data source topic". In user node player the error reads: `Output topic "${nodeData.outputTopic}" is already present in the data source`.

> Consider language like `Cannot map /imu to /pose . /pose is already present (in the data source).`

> Add the extension ID to the error message. If you have multiple extensions installed this could be helpful in tracking down which one is causing the problem.

3. `Topic mapping extension requested disallowed duplicate mapping from topic /map to topic /remapped_odom.` Same comments as above. Once we have language for the above we can align this error to follow whatever pattern we decide.

4. Do topic mappers run before message converters? So for my example, will I be able to write a message converter for `/selected_camera/camera_info` ?

----
Here are some more screenshots showing what I considered confusing behavior with "disappearing" topics. In the first screenshot I am inspecting my CAM_BACK_LEFT topic and see it in the topic list and raw message panel.

Screenshot 2023-06-17 at 7 10 42 AM

I then set the `selected_camera` global variable to `CAM_BACK_LEFT`. As expected, my `/selected_camera/camera_info` topic is now present BUT my `/CAM_BACK_LEFT/camera_info` topic is gone!

Screenshot 2023-06-17 at 7 10 58 AM

And I have no information telling me that `/selected_camera/camera_info` is a remapped topic.

----

P.S. I don't know that I feel strongly about this _yet_ but I did wonder whether topic _alias_ is a more natural description of what is happening? Or maybe what I _thought_ should be happening. A counter to this is that in ROS "mapping" is the name they've assigned to this concept. However, with ROS topic mapping, I think the original topic is _hidden_ from the node that you applied the mapping to or are publishing from. I don't think that same behavior is appropriate for Studio tho which one would use to inspect all the topics.

packages/studio-base/src/players/TopicAliasingPlayer/aliasing.ts
```diff
@@ -0,0 +1,360 @@
...
+  TopicStats,
+} from "@foxglove/studio-base/players/types";
+
+type TopicAliasMap = Map<string, string[]>;
+type MessageBlocks = readonly (undefined | MessageBlock)[];
+const EmptyAliasMap: Im<TopicAliasMap> = new Map();
+
+export type TopicAliasFunctions = Array<{ extensionId: string; aliasFunction: TopicAliasFunction }>;
+
+export type AlisingInputs = {
```
defunctzombie, line 26:
```suggestion
export type AliasingInputs = {
```

packages/studio-base/src/players/TopicAliasingPlayer/TopicAliasingPlayer.ts
```diff
@@ -0,0 +1,151 @@
...
+  #pendingSubscriptions: undefined | SubscribePayload[];
+  #subscriptions: SubscribePayload[] = [];
+
+  // True if no aliases are active and we can pass calls directly through to the
+  // underlying player.
+  #skipAliasing: boolean;
+
+  #listener?: (state: PlayerState) => Promise<void>;
+
+  public constructor(
```
defunctzombie, line 45: Does the constructor need to accept the alias functions and variables or is it sufficient to call the settings on those after constructing? If it is sufficient to call the setters I think its nicer to make the construction simpler.

defunctzombie (approved):
LGTM

@jtbandes fyi the user-facing extension api changed since you last approved

packages/studio-base/src/players/TopicAliasingPlayer/TopicAliasingPlayer.ts
foxymiles: IMO it's a bit more rigorous to initialize this with a complete state then piecemeal with the various parameters.

snosenzo (approved):

jtbandes (approved):

Issue #5877, opened 2023-04-29
Indicator panel not compatible with array message types


antrose99:
**Description**

This is a bit of a janky setup but I am using the sensor_msgs/msg/Imu message for INS data and storing the status message (which is just a number) in one of the covariances, which is defined as a Float64[9] message.

To visualize this status message in Foxglove I wanted to use the indicator panel to display the appropriate label and color for the given status. Entering all my rules and assigning the message path to, in my case, /imu/inspva.twist.covariance[0], led to the indicator falling back to the "Otherwise" rule (even if I set a rules that encapsulates all numbers e.g. one rule for >0 and one for <= 0). Using the raw panel with the same message path shows just the number (and not an array), as expected, and I can confirm that the indicator works fine for the other messages of the Imu message.

Version: 1.52.0
Platform: MCAP
Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native):
mcap file

**Steps To Reproduce**

1. Set the message path to one element of an array message in the indicator panel.

2. Add rules covering the entire space, e.g. for a float >0 and <= 0.

3. Notice that that the indicator falls back to the "Otherwise" case

Can provide an mcap containing messages with this problem if needed.

**Expected Behavior**

That it would use the data from the element in the array that was specified as the value the indicator compares against.

jtbandes:
> Can provide an mcap containing messages with this problem if needed.

Yes please!

antrose99:
Here is the mcap file: [indicator_array_mcap.zip](https://github.com/foxglove/studio/files/11360318/indicator_array_mcap.zip)

And here is a screenshot which perhaps better visualizes the behavior
![image](https://user-images.githubusercontent.com/93656450/235329473-915a8020-ac63-496f-ae53-3b010542857f.png)

pegadoflavian:
Thank you for fixing this :)

PR #6069, opened 2023-05-19
Default MessageEvent generic type to unknown

foxymiles:
**User-Facing Changes**
None

**Description**
Default the MessageEvent generic type to unknown to remove a lot of redundant explicit types.

PR #6083, opened 2023-05-22
fix: Indicator panel with TypedArrays

wimagguc:
**User-Facing Changes**

Fixes a bug where the Indicator panel didn't support TypedArrays.

Fixes: https://github.com/foxglove/studio/issues/5877

### DIFF
--- a/packages/studio-base/src/components/MessagePathSyntax/simpleGetMessagePathDataItems.test.ts
+++ b/packages/studio-base/src/components/MessagePathSyntax/simpleGetMessagePathDataItems.test.ts
@@ -9,7 +9,7 @@ import { simpleGetMessagePathDataItems } from "./simpleGetMessagePathDataItems";
 
 describe("simpleGetMessagePathDataItems", () => {
   it("returns root message if topic matches", () => {
-    const message: MessageEvent<unknown> = {
+    const message: MessageEvent = {
       topic: "/foo",
       receiveTime: { sec: 0, nsec: 0 },
       sizeInBytes: 0,
@@ -21,7 +21,7 @@ describe("simpleGetMessagePathDataItems", () => {
   });
 
   it("supports TypedArray messages", () => {
-    const message: MessageEvent<unknown> = {
+    const message: MessageEvent = {
       topic: "/foo",
       receiveTime: { sec: 0, nsec: 0 },
       sizeInBytes: 0,
@@ -37,7 +37,7 @@ describe("simpleGetMessagePathDataItems", () => {
   });
 
   it("returns correct nested values", () => {
-    const message: MessageEvent<unknown> = {
+    const message: MessageEvent = {
       topic: "/foo",
       receiveTime: { sec: 0, nsec: 0 },
       sizeInBytes: 0,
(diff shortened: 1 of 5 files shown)


packages/studio-base/src/components/MessagePathSyntax/simpleGetMessagePathDataItems.ts
```diff
@@ -32,7 +33,7 @@ export function simpleGetMessagePathDataItems(
     }
     switch (pathPart.type) {
       case "slice": {
-        if (!Array.isArray(value)) {
+        if (!Array.isArray(value) && !isTypedArray(value)) {
```
defunctzombie, line 36: https://github.com/foxglove/studio/blob/main/packages/studio-base/src/components/MessagePathSyntax/simpleGetMessagePathDataItems.test.ts add a test?

wimagguc:
The least wrong think I could think of was to move the raw message panel's `isTypedArray` util into under `types` — let me know how studio-esque this approach is @jtbandes

packages/studio-base/src/components/MessagePathSyntax/simpleGetMessagePathDataItems.test.ts
```diff
@@ -9,7 +9,7 @@ import { simpleGetMessagePathDataItems } from "./simpleGetMessagePathDataItems";
 
 describe("simpleGetMessagePathDataItems", () => {
   it("returns root message if topic matches", () => {
-    const message: MessageEvent = {
+    const message: MessageEvent<unknown> = {
```
jtbandes, line 12: shouldn't need the `<unknown>` since #6069

packages/studio-base/src/types/TypedArray.ts
jtbandes, line 1: I'd rename this file to `isTypedArray.ts` so it matches the export.

packages/studio-base/src/types/TypedArray.ts
```diff
@@ -0,0 +1,21 @@
...
+  | Int8Array
+  | Uint8Array
+  | Int16Array
+  | Uint16Array
+  | Int32Array
+  | Uint32Array
+  | Float32Array
+  | Float64Array
+  | BigInt64Array
+  | BigUint64Array;
```
jtbandes, line 17: nit: this should include Uint8ClampedArray as well

jtbandes (approved):

packages/studio-base/src/types/isTypedArray.ts
```diff
@@ -0,0 +1,22 @@
...
+  | Uint16Array
+  | Int32Array
+  | Uint32Array
+  | Float32Array
+  | Float64Array
+  | BigInt64Array
+  | BigUint64Array
+  | Uint8ClampedArray;
+
+export function isTypedArray(value: unknown): value is TypedArray {
```
defunctzombie, line 20: I find it weird to have functions in files called _types_ (and generally the stuff in this folder seems studio specific) - why'd this not go into den? @jtbandes ?

jtbandes: Where would it go inside den?

defunctzombie: An appropriate place.

PR #6095, opened 2023-05-23
Update loop icon to have clearer disabled state

2metres:
**User-Facing Changes**
Update loop icon to have clearer disabled state, also removes the component that was created to bold the loop button on hover

**Description**
| Before | After |
|---|---|
| Screen Shot 2023-05-23 at 10 44 15 am | Screen Shot 2023-05-23 at 10 43 28 am |

Closes FG-3571


amacneil (approved):
seems fine, I don't have strong feelings either way. others should chime in.

jtbandes (approved):

PR #6097, opened 2023-05-23
Revert "Expose our own Immutable type in @foxglove/studio (#6079)"

foxymiles:
This reverts commit e67769ad1c4a1d5a788f91b04748ab3822f9751c.

**User-Facing Changes**
None

**Description**
Revert the immutable PR. We like the concept but are not decided on the implementation of adding Immer as a dependency just for the type.

This assumes we go forward with https://github.com/foxglove/studio/pull/6098


defunctzombie (approved):
👍 I think not having the dependency is nice

snosenzo (approved):

PR #6098, opened 2023-05-23
Expose Immutable type from @foxglove/studio and mark message pipeline and renderState data as immutable

foxymiles:
**User-Facing Changes**
None

**Description**
Expose an Immutable type in @foxglove/studio and mark the message pipeline context and RenderState data as immutable. The implementation here uses the DeepReadonly type from the `ts-essentials` package and re-exposes it as `Immutable` but we could also just copy over the typedefs if we prefer not to have the dependency.

### DIFF
--- a/packages/studio/src/index.ts
+++ b/packages/studio/src/index.ts
@@ -88,7 +88,7 @@ export type Subscription = {
 /**
  * A message event frames message data with the topic and receive time
  */
-export type MessageEvent<T = unknown> = Readonly<{
+export type MessageEvent<T = unknown> = {
   /** The topic name this message was received on, i.e. "/some/topic" */
   topic: string;
   /**
@@ -103,14 +103,14 @@ export type MessageEvent<T = unknown> = Readonly<{
    * relative to another event such as system boot time or simulation start
    * time depending on the context.
    */
-  receiveTime: Readonly<Time>;
+  receiveTime: Time;
   /**
    * The time in nanoseconds this message was originally published. This is
    * only available for some data sources. The timestamp is often nanoseconds
    * since the UNIX epoch, but may be relative to another event such as system
    * boot time or simulation start time depending on the context.
    */
-  publishTime?: Readonly<Time>;
+  publishTime?: Time;
   /** The deserialized message as a JavaScript object. */
   message: T;
   /**
@@ -125,7 +125,7 @@ export type MessageEvent<T = unknown> = Readonly<{
    * un-converted message event.
    */
   originalMessageEvent?: MessageEvent;
-}>;
+};
 
 export interface LayoutActions {
   /** Open a new panel or update an existing panel in the layout.  */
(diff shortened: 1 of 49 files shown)


packages/studio/src/immutable.ts
```diff
@@ -0,0 +1,72 @@
...
+type AtomicObject = Function | Promise<any> | Date | RegExp;
+
+/**
+ * If the lib "ES2015.Collection" is not included in tsconfig.json,
+ * types like ReadonlyArray, WeakMap etc. fall back to `any` (specified nowhere)
+ * or `{}` (from the node types), in both cases entering an infinite recursion in
+ * pattern matching type mappings
+ * This type can be used to cast these types to `void` in these cases.
+ */
+export type IfAvailable<T, Fallback = void> =
```
foxymiles, line 44: We probably can get rid of this since we are targeting more recent JS versions.

packages/studio/src/immutable.ts
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * These should also never be mapped but must be tested after regular Map and
+ * Set
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type WeakReferences = IfAvailable<WeakMap<any, any>> | IfAvailable<WeakSet<any>>;
+
+/** Convert a mutable type into a readonly type */
+export type Immutable<T> = T extends PrimitiveType
```
defunctzombie, line 60: FWIW I think `DeepReadonly` while maybe more awkward also more conventional for typescript use. Even the type here uses plenty of `Readonly`. I would favor using a variant of `Readonly` over `Immutable` which is an "immerism"

foxymiles: IMO DeepReadonly is ungainly and conveys implementation more than intent but I can live with it.

foxymiles: When you see DeepReadonly several times in a complex type signature it becomes a bit of an eyesore.

jtbandes (approved):
No strong leanings on the name from me.

packages/studio/src/index.ts
```diff
@@ -84,7 +88,7 @@ export type Subscription = {
 /**
  * A message event frames message data with the topic and receive time
  */
-export type MessageEvent<T = unknown> = Readonly<{
+export type MessageEvent<T = unknown> = {
```
defunctzombie, line 91: Why drop the readonly?

defunctzombie (approved):
Seems legit. Do you think the "immutable" type from ts-essentials is better than the immer one?

packages/studio/src/index.ts
foxymiles: In places where this gets used its part of an Immutable parent. Marking core types like this just makes it harder to build them up. Better to do it at call sites I think.

foxymiles:
> Seems legit. Do you think the "immutable" type from ts-essentials is better than the immer one?

I think they're pretty much equivalent. The one in Immer does some extra work to handle older versions of JS we don't care about so this one is probably better for our needs.

PR #6099, opened 2023-05-23
Visibility toggle no longer hijacks spacebar

2metres:
**User-Facing Changes**
Fixes issues where toggling topic visibility hijacks spacebar

Closes FG-3333


foxymiles (approved):

jtbandes (approved):
Is there a version of this where we make it non-focusable? I guess that would interfere with legitimate keyboard use? (But so would calling blur())

PR #6100, opened 2023-05-23
Skip data source dialog when opening local file from app menu

foxymiles:
**User-Facing Changes**
None

**Description**
Present the file picker directly when choosing "open local file" from the app menu. This is exposed as a new workspace action under `dialogActions > openFile > open`.


packages/studio-base/src/components/AppBar/AppMenu.tsx
```diff
@@ -89,9 +94,9 @@ export function AppMenu(props: AppMenuProps): JSX.Element {
         label: t("openLocalFile"),
         key: "open-file",
         onClick: () => {
-          dialogActions.dataSource.open("file");
           handleAnalytics("open-file");
           handleNestedMenuClose();
+          openLocalFile().catch(console.error);
```
defunctzombie, line 99: Do we want to surface these errors to the user? Seems like this is a good way to end up with an error you cannot see.

packages/studio-base/src/components/DataSourceDialog/DataSourceDialog.tsx
```diff
@@ -17,11 +17,11 @@ import {
   useWorkspaceStore,
 } from "@foxglove/studio-base/context/Workspace/WorkspaceContext";
 import { useWorkspaceActions } from "@foxglove/studio-base/context/Workspace/useWorkspaceActions";
+import { useOpenFile } from "@foxglove/studio-base/hooks/useOpenFile";
```
defunctzombie, line 20: Why do we need to keep this? We have the workspace action. The dialog should use the workspace action.

packages/studio-base/src/Workspace.tsx
```diff
@@ -266,69 +266,7 @@ function WorkspaceContent(props: WorkspaceContentProps): JSX.Element {
...
-  useNativeAppMenuEvent(
-    "open-remote-file",
-    useCallback(() => dialogActions.dataSource.open("remote"), [dialogActions.dataSource]),
-  );
-
-  useNativeAppMenuEvent(
-    "open-sample-data",
-    useCallback(() => dialogActions.dataSource.open("demo"), [dialogActions.dataSource]),
-  );
+  useWorkspaceNativeAppMenuEvents();
```
defunctzombie, line 269: Personally I do not like this style of hook use. To me this is a component not a hook /shrug

Why are these moved at all?

packages/studio-base/src/components/AppBar/AppMenu.tsx
```diff
@@ -17,6 +17,7 @@ import {
   useWorkspaceStore,
 } from "@foxglove/studio-base/context/Workspace/WorkspaceContext";
 import { useWorkspaceActions } from "@foxglove/studio-base/context/Workspace/useWorkspaceActions";
+import { useOpenFile } from "@foxglove/studio-base/hooks/useOpenFile";
```
defunctzombie, line 20: This should be using the workspace action.

defunctzombie:
I'd like to see this implementation go all the way and actually use the workspace action which seems like that is what workspace actions were made for. This seems like we are kinda going half-way with our reworking of the logic and not quite getting to a nice end state.

packages/studio-base/src/Workspace.tsx
foxymiles: I moved them because the Workspace is already bloated and I didn't want to add more stuff into it in order to make this work. A think a component that only wraps a bunch of hooks is weird. Hooks are the compositional unit in non-gui react code. Wrapping a bunch of hooks in another hook is exactly how they're supposed to be used.

foxymiles:
> I'd like to see this implementation go all the way and actually use the workspace action which seems like that is what workspace actions were made for. This seems like we are kinda going half-way with our reworking of the logic and not quite getting to a nice end state.

We don't currently have a workspace action for this and this doesn't involve any state in the workspace store so to me it seems like a bit of an awkward fit. Which toplevel key would it go under? The native file picker is not really a dialog in the sense we're using it there now.

packages/studio-base/src/components/AppBar/AppMenu.tsx
foxymiles: Perhaps although we've been doing the same thing in DataSourceDialog for quite some time.

foxymiles:
I've moved this into workspace actions under `dialogActions > openFile > open`. IMO this is a slightly weird space for it to live because the native file picker isn't really a dialog in the sense we're using it here and it makes the workspace actions dependent on the player selection but it's also not that disruptive so I don't feel strongly about it.

packages/studio-base/src/context/Workspace/useWorkspaceActions.ts
```diff
@@ -24,13 +27,17 @@ import {
...
 
 export type WorkspaceActions = {
   dialogActions: {
     dataSource: {
       close: () => void;
       open: (item: DataSourceDialogItem, dataSource?: IDataSourceFactory) => void;
     };
+    openFile: {
+      open: () => Promise<void>;
+    };
```
jtbandes, line 40: I'm not sure why this is under `dialogActions` since it's not part of the open dialog anymore. Maybe a separate action/category at the top level?

jtbandes (approved):

packages/studio-base/src/context/Workspace/useWorkspaceActions.ts
foxymiles: Yeah it's kind of an awkward fit anywhere here I think.

PR #6101, opened 2023-05-23
[Image Beta] better stories for pan/zoom/rotate

jtbandes:
**User-Facing Changes**
None

**Description**
Improved stories to show where the mouse is being clicked and dragged, showing that the pan tracks the mouse correctly and the zoom is centered around the mouse.


foxymiles (approved):

snosenzo (approved):
I think a axis-colored pose marker or arrows in the middle might've been a bit more clear but this is definitely much better

jtbandes:
In the middle of the image you mean? Not sure I understand the suggestion.

PR #6111, opened 2023-05-24
Move non-studio-specific hooks to @foxglove/hooks package

jtbandes:
**User-Facing Changes**
None

**Description**
Move hooks under `studio-base/hooks` that aren't specific to studio into `packages/hooks`


packages/hooks/package.json
```diff
@@ -21,10 +21,13 @@
   },
   "devDependencies": {
     "@testing-library/react-hooks": "8.0.1",
+    "@types/foxglove__web": "workspace:*",
```
defunctzombie, line 24: What is this for?

defunctzombie (approved):
I don't understand the @types/foxglove__web dependency in the hooks package. It doesn't depend on @foxglove/web so why that type dependency? Seems suspect to me because @foxglove/web should sit _above_ the hooks package.

PR #6112, opened 2023-05-24
remove PanelSetup onMount/onFirstMount and clean up stories

jtbandes:
**User-Facing Changes**
None

**Description**
Removes last use of `onMount`. Also a little more `play` function cleanup.


defunctzombie (approved):
And so the future welcomes us and the past is left behind.

PR #6121, opened 2023-05-25
Fix point hover logic on map panel

foxymiles:
**User-Facing Changes**
Fix logic to detect hovered points on map panel.

**Description**
Fix logic to detect hovered points on map panel. When we have a preview time the existing logic looks for the most recent point _before_ the hover time, which causes it to pick the point before the actual hovered point. The fix is to look for the most precent point _before or at_ the current hover time which includes the currently hovered point.

### DIFF
--- a/packages/studio-base/src/panels/Map/MapPanel.tsx
+++ b/packages/studio-base/src/panels/Map/MapPanel.tsx
@@ -637,9 +637,10 @@ function MapPanel(props: MapPanelProps): JSX.Element {
       return;
     }
 
-    // get the point occuring most recently before preview time but not after preview time
+    // Find the point occuring most recently before or at preview time but not after
+    // preview time.
     const prevNavMessages = allNavMessages.filter(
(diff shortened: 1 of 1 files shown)


jtbandes (approved):

foxymiles:
@esthersweon Can you please confirm that this fixes the issue you observed?

jtbandes:
Can we add some kind of test for this?

foxymiles:
> Can we add some kind of test for this?

Difficult but I can try. It might be one of those tests that tends to be flaky.

esthersweon:
There are some instances where clicking on a point will still seek to the point before it – this feels like buggy behavior:

https://github.com/foxglove/studio/assets/6993359/e8f62c3d-8a88-477a-8eb2-87962ac3be52

foxymiles:
I'm going to merge this since it's definitely an improvement on the behavior on main but will keep the ticket open to track further fixes.

PR #6122, opened 2023-05-25
Avoid unnecessary readvertisments of foxglove websocket client channels

### LINKED ISSUE #6051: Making layout changes causes unadvertisement / advertisement of published topics
I added the `clientPublish` capability to my foxglove server (which is a great feature by the way 😄 )

I have two different publish panels on my layout. Now, when I make *some* kind of UI changes (e.g. add a new panel, delete a panel), the published channels are first unadvertised and then advertised again.
Shouldn't they be persistent? The same way that subscriptions are not changed when changes are made to unrelated tabs?

To reproduce:
Have a websocket server with the publish capability. Add a publisher panel. Then add a new publisher panel. The first panel gets unadvertised, and then both panels get advertised again.

Sometime (but not all the time) this is also the case when deleting an unrelated panel (e.g. Image panel).

achim-k:
**User-Facing Changes**
- Avoid unnecessary readvertisments of foxglove websocket client channels

**Description**
This PR makes several changes to avoid unnecessary advertisements / readvertisements of client channels when using the foxglove websocket connection:

- The player's `setPublishers` method is only called when topic or schemaName are valid (not empty)
- The player's datatypes are only updated if they actually changed
  - This is particular relevant when connecting to a foxglove_bridge instance, as client channels get also advertised as server channels.
- In the player's `setPublishers` method, we now determine which publications have been added / removed and only react on these changes rather than resetting all client publications.

Fixes #6051

### DIFF
--- a/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
+++ b/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
@@ -3,7 +3,7 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
 import * as base64 from "@protobufjs/base64";
-import { isEqual, uniqWith } from "lodash";
+import { isEqual, isMatch, uniqWith } from "lodash";
 import { v4 as uuidv4 } from "uuid";
 
 import { debouncePromise } from "@foxglove/den/async";
@@ -73,6 +73,7 @@ type ResolvedService = {
   parsedResponse: ParsedChannel;
   requestMessageWriter: MessageWriter;
 };
+type MessageDefinitionMap = Map<string, MessageDefinition>;
 
 export default class FoxgloveWebSocketPlayer implements Player {
   readonly #sourceId: string;
@@ -295,12 +296,12 @@ export default class FoxgloveWebSocketPlayer implements Player {
           ? CommonRosTypes.ros2galactic
           : CommonRosTypes.ros2humble;
 
+        const dataTypes: MessageDefinitionMap = new Map();
         for (const dataType in rosDataTypes) {
           const msgDef = (rosDataTypes as Record<string, MessageDefinition>)[dataType]!;
-          this.#datatypes.set(dataType, msgDef);
-          this.#datatypes.set(dataTypeToFullName(dataType), msgDef);
+          dataTypes.set(dataType, msgDef);
         }
-        this.#datatypes = new Map(this.#datatypes); // Signal that datatypes changed.
+        this.#updateDataTypes(dataTypes);
       }
 
       if (event.capabilities.includes(ServerCapability.clientPublish)) {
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
```diff
@@ -618,10 +618,18 @@ export default class FoxgloveWebSocketPlayer implements Player {
           : new JsonMessageWriter();
 
         // Add type definitions for service response and request
+        let datatypesNeedsUpdate = false;
         for (const [name, types] of [...parsedRequest.datatypes, ...parsedResponse.datatypes]) {
+          if (this.#datatypes.has(name)) {
+            // Assumption: Schemas do not change.
+            continue;
```
jtbandes, line 625: Do we need to make this assumption? Can we verify the assumption and raise an error to the user if it doesn't hold instead of silently failing?

packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
```diff
@@ -618,10 +618,18 @@ export default class FoxgloveWebSocketPlayer implements Player {
...
+          }
+          datatypesNeedsUpdate = true;
           this.#datatypes.set(name, types);
         }
-        this.#datatypes = new Map(this.#datatypes); // Signal that datatypes changed.
+        if (datatypesNeedsUpdate) {
+          this.#datatypes = new Map(this.#datatypes); // Signal that datatypes changed.
```
jtbandes, line 631: This pattern of mutating the existing map and then cloning later seems weird to me. Feels like we should clone as soon as we know we need to produce a new map, then add items to the new map.

packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
```diff
@@ -854,14 +873,45 @@ export default class FoxgloveWebSocketPlayer implements Player {
...
+    }
+
+    // Determine new & removed publications.
+    const currentPublications = Array.from(this.#publicationsByTopic.values());
+    const removedPublications = currentPublications.filter((channel) => {
+      return (
+        uniquePublications.find(
+          ({ topic, schemaName }) => channel.topic === topic && channel.schemaName === schemaName,
+        ) == undefined
+      );
```
jtbandes, line 892: Why do we use `isEqual` above but a different custom comparison here?

achim-k: We compare different types here: `AdvertiseOptions` and [Publication](https://github.com/foxglove/studio/blob/105f2404cfe9e2dd271af38b7e62d3e2883dd29e/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts#L70). Should I could change `Publication` to contain the entire `AdvertiseOptions`?

jtbandes: No worries, I didn't realize it was a different type

jtbandes (approved):

PR #6124, opened 2023-05-25
Rename PinholeCameraModel methods to be less confusing

jtbandes:
**User-Facing Changes**
None

**Description**
`rectifyPixel` -> `undistortPixel`
`unrectifyPixel` -> `distortPixel`

### DIFF
--- a/packages/den/image/PinholeCameraModel.test.ts
+++ b/packages/den/image/PinholeCameraModel.test.ts
@@ -127,52 +127,52 @@ describe("PinholeCameraModel", () => {
     });
   });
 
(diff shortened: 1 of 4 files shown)


jtbandes:
cc @jhurliman – can you confirm whether this new terminology seems accurate?

foxymiles:
Seems reasonable but not sure what's idiomatic for this stuff.

snosenzo (approved):

jtbandes:
OpenCV actually refers to undistortion and rectification as two separate things, undistortion being based on camera intrinsics. I think we may be doing both here although I'm not entirely sure.

PR #6127, opened 2023-05-26
[Image Beta] Move download to panel context menu

jtbandes:
**User-Facing Changes**
None

**Description**
Follow-up to https://github.com/foxglove/studio/pull/6073
image


defunctzombie (approved):

defunctzombie:
> Much less of a fan of this implementation vs the previous, but I assume it was requested.

Product review feedback was to put this back to the context menu behavior for now. No comment on the implementation desires - that's for you two to work out.

PR #6128, opened 2023-05-26
fix analytics context in extension panel

jtbandes:
**User-Facing Changes**
None

**Description**
Fix for #6125, which didn't actually work because the AnalyticsProvider was not exposed to the extension panel's inner React tree.


packages/studio-base/src/components/ForwardContextProviders.tsx
```diff
@@ -0,0 +1,68 @@
...
+): readonly [React.Context<T>, StoreApi<{ value: T }>] {
+  const value = useContext(context);
+  const [store] = useState(() => createStore(() => ({ value })));
+  useLayoutEffect(() => {
+    store.setState({ value });
+  }, [store, value]);
+  return useShallowMemo([context, store] as const);
+}
+
+function getContextValues(contexts: ForwardedContexts): ForwardedContextValues {
```
snosenzo, line 31:
```suggestion
function getContextValueMap(contexts: ForwardedContexts): ForwardedContextValues {
```
maybe also update `ForwardedContextValues`? feels a bit confusing since it's  a map but sounds like an array

packages/studio-base/src/components/ForwardContextProviders.tsx
```diff
@@ -0,0 +1,68 @@
...
+export function ForwardContextProviders({
+  /** Contexts to forward. A Map that should be built with entries from `useForwardContext()`. */
+  contexts,
+  children,
+}: React.PropsWithChildren<{ contexts: ForwardedContexts }>): JSX.Element {
+  useMustNotChange(contexts);
+  const [store] = useState(() =>
+    createStore<ForwardedContextValues>((set) => {
+      for (const contextStore of contexts.values()) {
+        contextStore.subscribe(() => set(getContextValues(contexts)));
```
snosenzo, line 57: I'm not super familiar with zustand, but what is this line doing exactly? Do we have to unsubscribe?

snosenzo (commented):
The innards are pretty confusing to me, being less familiar with zustand, but I appreciate its outward apparent simplicity. 
Feels like a little too much access to studio internals though, and wouldn't want it being used as an escape valve for things that should really have a better interface through the extension API. Would feel better if this only worked for analytics for now and then we can think harder about how we want to pass private state to extension panels.
The zustand usage definitely seems smelly, but i'll need to read up more before I really know better

defunctzombie:
Clever and tragic.

Since the context is an `IAnalytics` instance, could you accept an `IAnalytics` type instead of a `ForwardAnalytics` and then use a regular `AnalyticsContext.Provider` instead? I'm gonna guess no because the "instance" is actually hooked into the react machinery and its not "just a class instance" as one might image?

I agree with @snosenzo comment and the direction the PR has taken to limit this to the analytics context for now given its abstract art nature.

jtbandes:
> Since the context is an `IAnalytics` instance, could you accept an `IAnalytics` type instead of a `ForwardAnalytics` and then use a regular `AnalyticsContext.Provider` instead? I'm gonna guess no because the "instance" is actually hooked into the react machinery and its not "just a class instance" as one might image?

Actually the reason is different. It's because this inner `ReactDOM.render()` only happens once, inside of `initPanel()`. Even if the IAnalytics changed in the outer tree, we would not re-call initPanel. I worked around this using a store so I could `subscribe` to updates.

defunctzombie:
> Even if the IAnalytics changed in the outer tree, we would not re-call initPanel. I worked around this using a store so I could `subscribe` to updates.

Could we re-create the entire PanelExtensionAdapter machinery and re-call `initPanel` if that happens? If there is some way to re-init from above I do think it would be a win over this advanced "technique".

jtbandes:
That is possible but I think it would require PanelExtensionAdapter to be "aware" of the analytics value – and we would also have to expose it in the API for `initPanel`. I figured we do not want to expose this in our extension API.

snosenzo:
>  re-call initPanel

Also wouldn't calling `initPanel` again cause the panel to lose internal state not in the config? I imagine this analytics value shouldn't be changing, but it might be strange that the panel resets because it changed.

snosenzo (approved):
looks good to me with the scoped-down changes. Would be nice to test against update behavior, but since it's just analytics now it's probably fine.

PR #6129, opened 2023-05-26
Fix react error on button in playback controls tooltip

foxymiles:
**User-Facing Changes**
None

**Description**
Fix the react error caused by having a disabled button inside a `Tooltip`.

Screenshot 2023-05-26 at 10 24 32 AM

### DIFF
--- a/packages/studio-base/src/components/PlaybackControls/index.tsx
+++ b/packages/studio-base/src/components/PlaybackControls/index.tsx
@@ -24,7 +24,7 @@ import {
   ArrowRepeatAll20Regular,
   ArrowRepeatAllOff20Regular,
 } from "@fluentui/react-icons";
-import { Tooltip } from "@mui/material";
+import { Tooltip, useTheme } from "@mui/material";
 import { useCallback, useMemo, useState } from "react";
 import { makeStyles } from "tss-react/mui";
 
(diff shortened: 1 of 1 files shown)


jtbandes (commented):
Wouldn’t this also have the effect of making it so the button is still clickable? I imagine we want it to be actually disabled?

foxymiles:
> Wouldn’t this also have the effect of making it so the button is still clickable? I imagine we want it to be actually disabled?

There's no click action on this button anyway.

jtbandes (approved):
The only other change I'd consider making is whether there is a `theme` prop to get the disabled opacity, rather than hard-coding 50%.

packages/studio-base/src/components/PlaybackControls/index.tsx
```diff
@@ -206,7 +216,7 @@ export default function PlaybackControls(props: {
               >
                 <HoverableIconButton
                   className={classes.dataSourceInfoButton}
```
2metres, line 218: I'd prefer we use a className here something like 
```
className={cx(classes.dataSourceInfoButton, { [classes.disabled]: presence !== PlayerPresence.PRESENT })}
```

2metres: That way you don't need to call the theme inside

foxymiles: Yeah I agree that's a cleaner solution.

PR #6130, opened 2023-05-26
Stabilize the plot panel addMessages callback

foxymiles:
**User-Facing Changes**
None

**Description**
Stabilize the plot panel addMessages callback. This callback refers to our current loaded blocks time range but doesn't need to be invalidated when it changes so we can stabilize it by using a reference instead of a direct dependency.

### DIFF
--- a/packages/studio-base/src/panels/Plot/usePlotPanelMessageData.ts
+++ b/packages/studio-base/src/panels/Plot/usePlotPanelMessageData.ts
@@ -4,6 +4,7 @@
 
 import { groupBy, isEmpty, pick } from "lodash";
 import { useCallback, useEffect, useMemo, useState } from "react";
+import { useLatest } from "react-use";
 
 import { isLessThan, isTimeInRangeInclusive, subtract } from "@foxglove/rostime";
 import { Immutable } from "@foxglove/studio";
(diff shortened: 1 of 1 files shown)


jtbandes (commented):
Why don’t we need to re-process the messages when the blocks time range changes?

foxymiles:
> Why don’t we need to re-process the messages when the blocks time range changes?

Invalidating addMessages won't do that anyway. We'd need to toss the stuff we've accumulated across previous calls. The only case where this would be a problem is when blocks are loading slowly and the user hits playback in some part of the timeline not covered by a block though.

foxymiles:
@jtbandes I think we're going to have to overhaul the way we accumulate and process messages in this panel soon anyway to address some of the performance issues with larger datasets.

defunctzombie (approved):
:D

https://github.com/foxglove/studio/pull/5694#discussion_r1167069108

foxymiles:
> :D
> 
> [#5694 (comment)](https://github.com/foxglove/studio/pull/5694#discussion_r1167069108)

I should note that I haven't actually observed this to be a performance issue so the purpose of this PR is mainly to silence the warnings.

PR #6139, opened 2023-05-29
AppBar colors in theme

2metres:
**User-Facing Changes**
None

**Description**
Moves AppBar constants into the theme, motivation: Move theme colors into theme for re-use


packages/studio-desktop/src/main/StudioWindow.ts
```diff
@@ -101,14 +98,14 @@ const getTitleCase = (baseString: string): string =>
 
 type ClearableMenu = Menu & { clear: () => void };
 
+const theme = palette[nativeTheme.shouldUseDarkColors ? "dark" : "light"];
```
jtbandes, line 101: I don't think this will work when toggling between dark and light mode, will it?

2metres: Appears to from local testing.

jtbandes (changes requested):
main:

image
image


<br><br>
This branch:

image

(icon color is wrong)

image

(color does not change when toggling dark mode)

2metres:
@jtbandes I couldn't get my Windows version running can you check to see if this is fixed by calling that theme in a different place?

packages/studio-desktop/src/main/StudioWindow.ts
```diff
@@ -171,7 +167,7 @@ function newStudioWindow(deepLinks: string[] = [], reloadMainWindow: () => void)
   nativeTheme.on("updated", () => {
     if (isWindows) {
       // Although the TS types say this function is always available, it is undefined on non-Windows platforms
-      browserWindow.setTitleBarOverlay(getTitleBarOverlayOptions());
+      browserWindow.setTitleBarOverlay(getTitleBarOverlayOptions({ isDark }));
```
jtbandes, line 170: Nope, still doesn't auto-adjust. This line looks wrong because isDark is not updated to the latest value.

Also, the icons are still black: 
image

2metres: Hmm… I'm out of my depth here. Could use some help causing it to re-render

jtbandes: Is there a ticket for this task? Maybe you can assign to me or if there isn't one, create one and assign me

jtbandes (approved):

defunctzombie (commented):
nifty

PR #6148, opened 2023-05-30
[Image Beta] remove migration of legacy config within new panel code

jtbandes:
**User-Facing Changes**
None

**Description**
Since we are not swapping the old and new panels in-place, the new panel doesn't need to read config written by the old panel. We will move this into a separate migration step, like https://github.com/foxglove/studio/blob/main/packages/studio-base/src/services/migrateLegacyToNew3DPanels.ts (see also: #6145)

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
+++ b/packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
@@ -37,7 +37,7 @@ import type {
 } from "./IRenderer";
 import type { PickedRenderable } from "./Picker";
 import { SELECTED_ID_VARIABLE } from "./Renderable";
-import { LegacyImageConfig, Renderer } from "./Renderer";
+import { Renderer } from "./Renderer";
 import { RendererContext, useRendererEvent } from "./RendererContext";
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
```diff
@@ -137,7 +127,12 @@ export function ThreeDeeRender(props: {
       topics: partialConfig?.topics ?? {},
       layers: partialConfig?.layers ?? {},
       publish,
-      imageMode,
+      imageMode: {
+        ...partialConfig?.imageMode,
+        annotations: partialConfig?.imageMode?.annotations as
```
snosenzo, line 132: why is this casting necessary?

snosenzo (approved):

packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
jtbandes: The DeepPartial makes it so the array elements have optional properties, and I didn't feel like adding a filterMap to avoid that - realistically there shouldn't be any missing properties in there. I guess this will all change with #6120 anyway

PR #6158, opened 2023-05-31
Replace addSubscriptionsToRenderer with getSubscriptions

jtbandes:
**User-Facing Changes**
None

**Description**
Refactor to make the subscription APIs more declarative.


packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -614,7 +624,7 @@ export class Renderer extends EventEmitter<RendererEvents> implements IRenderer
     this.emit("schemaHandlersChanged", this);
   }
 
-  public addSchemaSubscriptions<T>(
+  #addSchemaSubscriptions<T>(
```
snosenzo, line 627: do we also want to update or remove the incorrect deduplication of handlers overwriting here? (in the `subscription instanceof Function` case)

snosenzo (approved):

snosenzo (approved):

PR #6159, opened 2023-05-31
Add items to Open welcome dialog and Explore sample data to AppMenu

2metres:
**User-Facing Changes**
Users can now get to Example Data and Open Dialog

Screen Shot 2023-05-31 at 1 37 54 pm

Closes FG-3599


defunctzombie:
I'm personally 👎 on this change. I do not view "Explore sample data" as such a common task that it requires this kind of call-out at the top level of our primary drop-down menu. The linear issue calls for adding it to the file menu which is still _meh_ to me. I'd probably consider throwing it into _help_? Anyhow - definitely not at the top level - I would revert this.

PR #6162, opened 2023-05-31
Display VersionBanner instead of crashing on Safari <16.4

jtbandes:
**User-Facing Changes**

**Description**
For older browsers such as Safari < 16.4 that don't support [static initialization blocks](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Static_initialization_blocks), we need to avoid importing the majority of Studio code or we'll hit a SyntaxError. By adding this check, we reduce the amount of code imported before we're sure the required features are supported by the browser. If the features are not supported, we display only the VersionBanner.

Additionally, the `params` object is changed to a `getParams` async function that's called after the compatibility checks, so the embedding code can similarly avoid importing most of its code until the checks have passed.

### DIFF
--- a/packages/studio-web/src/index.tsx
+++ b/packages/studio-web/src/index.tsx
@@ -10,6 +10,7 @@ import type { IDataSourceFactory } from "@foxglove/studio-base";
 import CssBaseline from "@foxglove/studio-base/components/CssBaseline";
 
 import VersionBanner from "./VersionBanner";
+import { canRenderApp } from "./canRenderApp";
 
 const log = Logger.getLogger(__filename);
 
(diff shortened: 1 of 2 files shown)


packages/studio-web/src/index.tsx
```diff
@@ -23,12 +24,26 @@ function LogAfterRender(props: React.PropsWithChildren<unknown>): JSX.Element {
...
   dataSources?: IDataSourceFactory[];
   extraProviders?: JSX.Element[];
 };
 
-export async function main(params: MainParams = {}): Promise<void> {
+/**
+ * Safari < 16.4 doesn't support `static{}` blocks in classes. TypeScript sometimes uses these when
+ * emitting code for decorators.
+ */
+function supportsClassStaticInitialization() {
```
defunctzombie, line 36: A nicer organization scheme would be to put these in a compat file rather than continuing to add them them into the index.tsx.

packages/studio-web/src/index.tsx
```diff
@@ -23,12 +24,26 @@ function LogAfterRender(props: React.PropsWithChildren<unknown>): JSX.Element {
...
+  try {
+    // eslint-disable-next-line no-new-func
+    new Function("class X { static { } }");
+    return true;
+  } catch (err) {
+    return false;
+  }
+}
+
+export async function main(getParams: () => Promise<MainParams> = async () => ({})): Promise<void> {
```
defunctzombie, line 46: Seems like it would make the calling code more awkward (is there an example PR?)

defunctzombie (approved):

PR #6163, opened 2023-05-31
Use theme from makeStyles instead of useTheme

2metres:
**User-Facing Changes**
None

**Description**
TIL that makeStyles exports the theme too


jtbandes (approved):
Sure, why not.

2metres:
How does one get a license?

2metres:
image

PR #6164, opened 2023-05-31
Fix player warning on global variable change

snosenzo:
**User-Facing Changes**

 - Fix player warning on global variable change

**Description**

- remove 3D panel listening to global variable changes
- add `messagesRecomputed` flag to ignore messages out of order error when recomputing current frame on global variable change

FG-3170

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
+++ b/packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
@@ -22,7 +22,6 @@ import {
   SettingsTreeNodes,
   Subscription,
   Topic,
-  VariableValue,
 } from "@foxglove/studio";
 import { AppSetting } from "@foxglove/studio-base/AppSetting";
 import ThemeProvider from "@foxglove/studio-base/theme/ThemeProvider";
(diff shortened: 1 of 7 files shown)


jtbandes (approved):
I guess this is fine. I do think a more proper solution would be a full resolution to FG-3749, but since we've filed that separately I guess we can address the warning message only for now.

The warning message was meant to flag that assumptions elsewhere in the app may be violated – which they indeed are  (hence FG-3749). But maybe the importance of those assumptions is not so critical these days.

Would be interested to get @defunctzombie's take on this.

packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
```diff
@@ -366,7 +361,6 @@ export function ThreeDeeRender(props: {
     context.watch("didSeek");
     context.watch("parameters");
     context.watch("sharedPanelState");
-    context.watch("variables");
```
defunctzombie, line 369: Does the 3d panel not use variables as an input to anything?

defunctzombie:
I'm a skeptical of this change almost to the point of saying we should probably re-consider it. My rough understanding of the data pipeline assumptions are that we should not re-deliver the same data to panels for the "current frame" unless it is marked as a "seek". If we break these assumptions panels which accumulate data themselves over-accumulate.

I read the original linear issue but did not connect that to how it resulted in triggering this player error? Why does clicking a marker do anything to messages in the message pipeline? Maybe there's a gap in the PR description here explaining how this happens and that gap is leading me to think this new player flag not the right "fix".

PR #6165, opened 2023-05-31
[Image Beta] Port min/max value settings for mono16 images

jtbandes:
**User-Facing Changes**
None

**Description**
Port the Min/Max settings which currently only apply to mono16 raw images.
image
image


packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageMode.ts
```diff
@@ -497,16 +504,24 @@ export class ImageMode
...
-
       if (config.flipVertical !== prevImageModeConfig.flipVertical) {
         this.#imageRenderable?.setFlipVertical(config.flipVertical);
         this.#camera.setFlipVertical(config.flipVertical);
       }
+      if (
+        config.minValue !== prevImageModeConfig.minValue ||
+        config.maxValue !== prevImageModeConfig.maxValue
+      ) {
+        this.#imageRenderable?.setSettings({
```
snosenzo, line 519: feels like all of these settings (flipVertical, flipHorizontal, rotation) could be consolidated into a single `setSettings`, then detect whether they've changed in the image renderable?

snosenzo (approved):
some consolidation of the setting of settings would be nice, but otherwise looks good

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageMode.ts
jtbandes: Removed the others as they're no longer needed after #6127

PR #6166, opened 2023-06-01
Show endTime in app bar

2metres:
**User-Facing Changes**
Screen Shot 2023-06-20 at 1 58 57 pm

**Description**
When connected to a live data source, show the _endTime_ in the top navbar. This provides a crude indicator on the status of a connection and the reported time (for simulated clock) for live connections since we do not show the playback bar in those cases.

FG-3581


packages/studio-base/src/components/AppBar/DataSource.tsx
```diff
@@ -59,30 +65,78 @@ const useStyles = makeStyles<void, "adornmentError">()((theme, _params, _classes
...
+  const durationRef = useRef<HTMLDivElement>(ReactNull);
+
+  const startTime = useMessagePipeline(selectStartTime);
+  const endTime = useMessagePipeline(selectEndTime);
+
+  const play = useMessagePipeline(selectPlay);
+  const pause = useMessagePipeline(selectPause);
+  const seek = useMessagePipeline(selectSeek);
+
+  const isConnection = !(play && pause && seek);
```
2metres, line 121: Is there a cleaner way to determine if the current player is a live connection?

jtbandes: Since you are using this to gate displaying the startTime, maybe just check presence of endTime?

jtbandes:
I'm not a fan of the "breadcrumbs"-style display for something that is not a navigation breadcrumb.

2metres:
Not sure how to fix it but If you aren't using any of the data in the panels the duration doesn't increment

packages/studio-base/src/components/Stack.tsx
```diff
@@ -291,6 +291,9 @@ export type StackProps = {
   /** CSS styles to apply to the component. */
   style?: CSSProperties;
 
+  /** HTML title attribute */
+  title?: string;
```
jtbandes, line 295: Should we be using mui tooltips instead of titles? I think we're not very consistent throughout the app.

packages/studio-base/src/components/AppBar/DataSource.tsx
```diff
@@ -59,30 +65,74 @@ const useStyles = makeStyles<void, "adornmentError">()((theme, _params, _classes
...
+  const playerSourceId = useMessagePipeline(selectPlayerSourceId);
   const playerName = useMessagePipeline(selectPlayerName);
   const playerPresence = useMessagePipeline(selectPlayerPresence);
   const playerProblems = useMessagePipeline(selectPlayerProblems) ?? [];
+
+  const isLiveConnection =
+    playerSourceId != undefined
+      ? playerSourceId.endsWith("socket") ||
+        playerSourceId.endsWith("lidar") ||
+        playerSourceId.endsWith("device")
```
jtbandes, line 122: It's not good to have a dependence on particular sources here. This will be brittle and potentially break if we change our data source IDs in the future. If we need more information from the source, it would be better to actually change the source/player APIs. On the other hand it's usually better to detect only the features that need to be used rather than checking something high-level like the source id.

How did the previous implementation of time/duration work?

jtbandes (commented):
I think the behavior you describe is "as designed" – we don't always live-update the current time if we haven't received new messages. There are many different possible time sources: message logTimes, explicit times sent by servers, and the system clock. If handling these all correctly is posing some difficulty I would recommend talking to @achim-k who is most familiar with the live sources.

I still feel like `some-connection / 123 / 0:00.00` is not a particularly _clear_ way of displaying this information. Having tooltips helps (though mui tooltip appears faster than `title`). But have we explored putting labels on them as well?

On the other hand, have we explored putting the connection info in the playback bar? I think we've gone back and forth on showing the playback bar for live connections. For example, YouTube's playback bar for a live stream:
image

packages/studio-base/src/components/AppBar/DataSource.tsx
2metres: https://github.com/foxglove/studio/blob/81eb1b49a1d64dacc2a583445ecbdc281bb3e455/packages/studio-base/src/components/DataSourceInfoView.tsx#L63-L66

2metres: I agree, this seems extremely britle. Ideally we should have something come from the MessagePipeline or PlayerCapabilities that say "isLiveSource"

packages/studio-base/src/components/Stack.tsx
2metres: > I still feel like `some-connection / 123 / 0:00.00` is not a particularly _clear_ way of displaying this information. Having tooltips helps (though mui tooltip appears faster than `title`). But have we explored putting labels on them as well?

I generally agree with that idea. However, I think what makes youtube live work really well is the ability to pause/scrub back on a live source. :scope_creep:

2metres:
Feedback from @defunctzombie favor endTime over startTime and duration. AppBar is fine for now. Live connection contols in playback bar save for future feature planning.

defunctzombie:
Gave this a spin - I would favor putting the `/` back to separate the data source from the time

image

packages/studio-base/src/components/AppBar/DataSource.tsx
defunctzombie: Let's simplify this to how we detect whether to show the playback controls. We should not have this "name" detection logic.

packages/studio-base/src/components/AppBar/Timestamp.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+export function Timestamp(props: { time?: Time }): JSX.Element | ReactNull {
+  const { time } = props;
+  const [timezone] = useAppConfigurationValue<string>(AppSetting.TIMEZONE);
+  const { timeFormat } = useAppTimeFormat();
+
+  const timeRef = useRef<HTMLDivElement>(ReactNull);
+
+  useEffect(() => {
+    if (timeRef.current) {
```
defunctzombie, line 22: This would read nicer if you bail early and reduce the levels of indentation.

packages/studio-base/src/components/AppBar/DataSource.tsx
```diff
@@ -59,29 +61,40 @@ const useStyles = makeStyles<void, "adornmentError">()((theme, _params, _classes
...
 export function DataSource(): JSX.Element {
   const { t } = useTranslation("appBar");
   const { classes, cx } = useStyles();
 
+  const playerSourceId = useMessagePipeline(selectPlayerSourceId);
   const playerName = useMessagePipeline(selectPlayerName);
   const playerPresence = useMessagePipeline(selectPlayerPresence);
   const playerProblems = useMessagePipeline(selectPlayerProblems) ?? [];
+
+  const endTime = useMessagePipeline(selectEndTime);
```
defunctzombie, line 89: We should not add this selector here. This will cause the render function for the entire DataSource to run. Instead this should be scoped to re-render the smallest set of components.

defunctzombie (commented):
What is the difference between `AppBar/Timestamp.tsx` and `components/Timestamp.tsx`? Adding some comments to those components would aid in answering that question and helping us maintain them going forward.

2metres:
> What is the difference between `AppBar/Timestamp.tsx` and `components/Timestamp.tsx`? Adding some comments to those components would aid in answering that question and helping us maintain them going forward.

`components/Timestamp.tsx` will likely be deleted as part of the new UI rollout. At that time `AppBar/Timestamp.tsx` should move to `/components`

packages/studio-base/src/components/AppBar/Timestamp.tsx
```diff
@@ -0,0 +1,47 @@
...
+  useMessagePipeline,
+} from "@foxglove/studio-base/components/MessagePipeline";
+import { useAppConfigurationValue, useAppTimeFormat } from "@foxglove/studio-base/hooks";
+import { format } from "@foxglove/studio-base/util/formatTime";
+import { fonts } from "@foxglove/studio-base/util/sharedStyleConstants";
+import { formatTimeRaw, isAbsoluteTime } from "@foxglove/studio-base/util/time";
+
+const selectEndTime = (ctx: MessagePipelineContext) => ctx.playerState.activeData?.endTime;
+
+export function Timestamp(): JSX.Element | ReactNull {
```
defunctzombie, line 19: Worth a comment describing what this component does. It is not a generic timestamp and is `endTime` specific.

There should also be a comment for this `timeRef` logic because to anyone reading it, it will be a bit of a wtf why didn't they put the time in the div rather than this ref stuff.

defunctzombie:
Maybe this feedback was missed but we should not have this detection logic:

```
playerSourceId.endsWith("socket") ||
        playerSourceId.endsWith("lidar") ||
        playerSourceId.endsWith("device")
```

2metres:
> Maybe this feedback was missed but we should not have this detection logic:
> 
> ```
> playerSourceId.endsWith("socket") ||
>         playerSourceId.endsWith("lidar") ||
>         playerSourceId.endsWith("device")
> ```

2metres:
> Maybe this feedback was missed but we should not have this detection logic:
> 
> ```
> playerSourceId.endsWith("socket") ||
>         playerSourceId.endsWith("lidar") ||
>         playerSourceId.endsWith("device")
> ```

@defunctzombie will need help from someone that understands the player to complete this

defunctzombie (approved):

packages/studio-base/src/components/AppBar/DataSource.tsx
```diff
@@ -105,6 +111,12 @@ export function DataSource(): JSX.Element {
           <div className={classes.textTruncate}>
             <TextMiddleTruncate text={playerDisplayName ?? `<${t("unknown")}>`} />
           </div>
+          {isLiveConnection && (
+            <>
+              <span>/</span>
+              <EndTimestamp />
+            </>
+          )}
```
jtbandes, line 119: Maybe an unpopular opinion but rather than coming up with some "crude" `isLive` detection it feels like showing or not showing the endTime should be gated on whether `endTime == undefined`.

PR #6167, opened 2023-06-01
Make MenuItems consistent heights

2metres:
**User-Facing Changes**
Make MenuItems consistent heights

FG-2700

### DIFF
--- a/packages/studio-base/src/components/PanelContextMenu.tsx
+++ b/packages/studio-base/src/components/PanelContextMenu.tsx
@@ -2,7 +2,7 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
-import { Divider, ListItemText, Menu, MenuItem } from "@mui/material";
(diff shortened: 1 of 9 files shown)


packages/studio-base/src/components/PanelContextMenu.tsx
```diff
@@ -116,7 +116,7 @@ export function PanelContextMenu(props: PanelContextMenuProps): JSX.Element {
               key={`item_${index}_${item.label}`}
               disabled={item.disabled}
             >
-              <ListItemText>{item.label}</ListItemText>
```
jtbandes, line 119: The examples at https://mui.com/material-ui/react-menu/ use `ListItemText`. Do we need to deviate from that? Is there a way we can fix this via the theme instead?

2metres: Introducing the ListItemText component inside of a menu is part of the issue we have with inconsistent sizing. The MUI docs use them interchangeably however they result in different styles. Typically I would try to avoid using ListItem components inside of MenuItems.

jtbandes (approved):

PR #6168, opened 2023-06-01
Separate PanelList and PanelGrid for clarity

2metres:
**User-Facing Changes**
None

**Description**
Trying to isolate the styling on the ChangePanelMenu and couldn't stand it any longer. Clear separation of styles and functionality.


packages/studio-base/src/context/PanelCatalogContext.ts
```diff
@@ -53,4 +55,27 @@ export function usePanelCatalog(): PanelCatalog {
...
+        const otherDisplayName = existingPanel.title ?? existingPanel.type ?? "<unnamed>";
+        throw new Error(
+          `Two components have the same panelType ('${type}') and no preset configs: ${otherDisplayName} and ${dispName}`,
+        );
+      }
+    }
+    panelTypes.set(type, panel);
+  }
+}
+
```
2metres, line 80: Not sure this is the right place for this but would like to put somewhere shared

packages/studio-base/src/components/PanelGrid/PanelGrid.tsx
```diff
@@ -0,0 +1,200 @@
...
+    }),
+    [getFilteredPanels, allRegularPanels, allPreconfiguredPanels],
+  );
+
+  const allFilteredPanels = useMemo(
+    () => [...filteredPreconfiguredPanels, ...filteredRegularPanels],
+    [filteredPreconfiguredPanels, filteredRegularPanels],
+  );
+
+  const noResults = allFilteredPanels.length === 0;
```
2metres, line 142: This code could all be potentially shared

jtbandes (changes requested):
Separating these seems like a reasonable goal but there's a lot of duplicated logic right now. Maybe we need the boundaries to be a bit different where the search field and filtering logic is shared and only the rendering of the list/grid is not shared.

jtbandes (approved):

PR #6174, opened 2023-06-01
[Image Beta] - foreground image with opacity

snosenzo:
**User-Facing Changes**

 - [Image Beta] - add image overlay with opacity setting

**Description**
Add foreground image with opacity control in image mode settings. This is to mirror RVIZ behavior where it renders a foreground and background image with alpha, it is not currently present in the old image panel. 

Also updated ImageAnnotations to always render on top of everything in the scene

- [x] add stories (waiting for @jtbandes to merge story changes from [this PR](https://github.com/foxglove/studio/pull/6172) so there aren't conflicts)

opacity @ 0.0 (default):
image

opacity @ 0.5:
image

opacity @ 1.0:
image

FG-2689

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/Images/ImageRenderable.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/Images/ImageRenderable.ts
@@ -26,7 +26,8 @@ export interface ImageRenderableSettings {
   color: string;
   minValue?: number;
   maxValue?: number;
-  opacity?: number;
+  /** Opacity of foreground image, not used when `renderBehindScene` set to true */
+  foregroundOpacity: number;
 }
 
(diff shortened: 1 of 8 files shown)


packages/studio-base/src/panels/ThreeDeeRender/renderables/Images/ImageRenderable.ts
```diff
@@ -285,6 +307,17 @@ export class ImageRenderable extends Renderable<ImageUserData> {
       transparent,
       depthWrite: !transparent,
     });
+    if (!this.#renderBehindScene) {
+      return;
+    }
+    this.#foregroundMaterial = new THREE.MeshBasicMaterial({
```
jtbandes, line 313: I'm confused by the naming, why do we want a foregroundMaterial if we are rendering "behind" the scene?

jtbandes (approved):

PR #6184, opened 2023-06-02
Simplify AppBar stories using storybook decorators

2metres:
**User-Facing Changes**
Adds translations for AppBar elements

**Description**
Simplify AppBar stories using storybook decorators


packages/studio-base/src/components/AppBar/index.stories.tsx
```diff
@@ -16,268 +18,254 @@ export default {
...
+export const SignInStates: Story = {
+  decorators: [
+    (Story: StoryFn): JSX.Element => {
+      return (
+        <>
+          <div style={{ padding: 8 }}>sign in undefined</div>
```
jtbandes, line 82: Why is this one separate?

jtbandes (approved):

packages/studio-base/src/components/AppBar/index.stories.tsx
2metres: Mainly because that story is designed to show what happens when the `CurrentUserContext.Provider` is not present

esthersweon:
Storybook changes are not user-facing changes, so will not include in changelog.

PR #6193, opened 2023-06-05
Fix built-in datatypes overwriting player datatypes

### LINKED ISSUE #6190: Values are being parsed wrong when publishing msg PoseWithCovarianceStamped

achim-k:
**User-Facing Changes**

- Fix built-in datatypes overwriting player datatypes

**Description**

From https://github.com/foxglove/studio/issues/6190#issuecomment-1576719837

> The reason for the shift is, because Studio uses the ROS1 std_msgs/Header definition instead of the ROS2 definition which should actually be used. The ROS1 definition contains an extra uint32 seq field which causes the shift.
>
> The ROS1 definitions are added [here](https://github.com/foxglove/studio/blob/54a1534574d97a5bcb60083d7065b507af0daa1d/packages/studio-base/src/util/basicDatatypes.ts#L18-L26) and communicated to the publish panel via the UserNodePlayer.

This patch fixes that player datatypes are being overwritten by builtin datatypes with the same name (e.g. `std_msgs/Header` in this specific case).

Fixes #6190

### DIFF
--- a/packages/studio-base/src/players/UserNodePlayer/index.ts
+++ b/packages/studio-base/src/players/UserNodePlayer/index.ts
@@ -212,7 +212,7 @@ export default class UserNodePlayer implements Player {
     (datatypes: RosDatatypes, nodeDatatypes: readonly RosDatatypes[]): RosDatatypes => {
       return nodeDatatypes.reduce(
         (allDatatypes, userNodeDatatypes) => new Map([...allDatatypes, ...userNodeDatatypes]),
-        new Map([...datatypes, ...basicDatatypes]),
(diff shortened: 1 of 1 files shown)


achim-k:
> Needs tests!

Added a test in https://github.com/foxglove/studio/pull/6193/commits/09964e7205d0212e01e9fafacfd57ec288f905e9 which fails without https://github.com/foxglove/studio/pull/6193/commits/5445224bdc528bf68e0889d6e4221197efb404ce.
Should I add more tests? Not sure how else to test this fix though

defunctzombie:
Good find and fix 👍

PR #6194, opened 2023-06-05
Fix calculation of number of points for plot data memory limit

foxymiles:
**User-Facing Changes**
None

**Description**
Fix calculation of number of points for plot data memory limit. The existing calculation undercounts the number of points.


jtbandes (approved):
🤔 I wonder why it was done this way.

defunctzombie:
That's quite the number of for loops. Loops upon loops.

foxymiles:
> That's quite the number of for loops. Loops upon loops.

Found a better way in a subsequent PR: https://github.com/foxglove/studio/blob/f2b8284537ed9511766134e20f134f9ef5cc62e8/packages/studio-base/src/panels/Plot/plotData.ts#L137

PR #6195, opened 2023-06-05
Inline immutable definition from ts-essentials

foxymiles:
**User-Facing Changes**
None

**Description**
Inline immutable definition from ts-essentials. Unfortunately it looks like yarn & npm don't install transitive dev dependencies  from @foxglove/studio in extensions so it seems cleaner to just inline the definition.

### DIFF
--- a/packages/studio/src/immutable.ts
+++ b/packages/studio/src/immutable.ts
@@ -3,6 +3,29 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
 // Immutable types borrowed from ts-essentials: https://github.com/ts-essentials/ts-essentials
+/**
+The MIT License
+
+Copyright (c) 2018-2019 Chris Kaczor (github.com/krzkaczor)
+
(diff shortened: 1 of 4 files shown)


packages/studio/src/immutable.ts
```diff
@@ -0,0 +1,58 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+// Immutable types borrowed from ts-essentials: https://github.com/ts-essentials/ts-essentials
```
jtbandes, line 5: I think we need to include license information here?

jtbandes (approved):

packages/studio/src/immutable.ts
foxymiles: Copied over the MIT license from ts-essentials.

PR #6196, opened 2023-06-05
Fix LinePrimitive rendering and selection issues

snosenzo:
**User-Facing Changes**

 - Fix LinePrimitive rendering and selection issues

**Description**
Selection issue was happening because scale_invariant = true makes the thickness of lines maintain an equivalent pixel value across zoom levels and screen sizes, so when we render to the picking target which is 31x31 pixels, the line is rendered such that the mouse click position is no longer over the line. 

The line width issue came from setting `linewidth` instead of `lineWidth` on our in-house LineMaterial.

FG-3424 FG-3742


packages/studio-base/src/panels/ThreeDeeRender/Picker.ts
```diff
@@ -287,11 +289,28 @@ export class Picker {
...
     const sizeAttenuation =
       (material as Partial<THREE.PointsMaterial>).sizeAttenuation === true ? 1 : 0;
     const pickingMaterial = renderItem.object.userData.pickingMaterial as
       | THREE.ShaderMaterial
+      | LineMaterial
       | undefined;
+    // line thickness scales on resolution. for smaller picking targets we need to make sure
+    // it is scaled, so that it can be picked correctly
+    if (pickingMaterial?.type === "LineMaterial") {
+      (pickingMaterial as LineMaterial).resolution = pickResolution;
```
jtbandes, line 306: Seems risky to save a reference to a tempResolution varaible

packages/studio-base/src/panels/ThreeDeeRender/Picker.ts
```diff
@@ -287,11 +289,28 @@ export class Picker {
...
+    );
     const sizeAttenuation =
       (material as Partial<THREE.PointsMaterial>).sizeAttenuation === true ? 1 : 0;
     const pickingMaterial = renderItem.object.userData.pickingMaterial as
       | THREE.ShaderMaterial
+      | LineMaterial
       | undefined;
+    // line thickness scales on resolution. for smaller picking targets we need to make sure
+    // it is scaled, so that it can be picked correctly
+    if (pickingMaterial?.type === "LineMaterial") {
```
jtbandes, line 305: Why do we care if it is a LineMaterial or some other material?

packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableLines.ts
```diff
@@ -284,18 +270,13 @@ class LinePrimitiveRenderable extends THREE.Object3D {
     if (this.#primitive == undefined) {
       return;
     }
-    this.#material.linewidth = this.#primitive.thickness;
+    this.#material.lineWidth = this.#primitive.thickness;
```
jtbandes, line 273: why is typescript happy with both of these? 🤔

jtbandes (commented):
I'm wondering if there is some way the Picker and Renderer can avoid knowing about LineMaterial. Seems not ideal to couple these.

packages/studio-base/src/panels/ThreeDeeRender/Picker.ts
snosenzo: ugh, yeah our custom `LineMaterial` has a custom `set` that does a `copy` internally.

packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableLines.ts
snosenzo: our custom `LineMaterial` extends `ShaderMaterial` which also has `.linewidth`, but it doesn't seem to update the uniform int he same way.  Updated the name of our custom line material to be less confusing

snosenzo:
> I'm wondering if there is some way the Picker and Renderer can avoid knowing about LineMaterial. Seems not ideal to couple these.

updated to make it only check `.uniforms.resolution` and update it if it's on the material.

packages/studio-base/src/panels/ThreeDeeRender/LineMaterialWithAlphaVertex.ts
```diff
@@ -471,7 +471,7 @@ export interface LineMaterialParameters extends MaterialParameters {
   worldUnits?: boolean | undefined;
 }
 
-export class LineMaterial extends ShaderMaterial {
+export class LineMaterialWithAlphaVertex extends ShaderMaterial {
```
jtbandes, line 474: Maybe add a comment describing what functionality is added by the subclass?

packages/studio-base/src/panels/ThreeDeeRender/renderables/markers/materials.ts
```diff
@@ -75,35 +75,41 @@ export function makeStandardInstancedMaterial(marker: Marker): THREE.MeshStandar
...
     linewidth: lineWidth,
-    resolution: options.resolution,
+    resolution: options.resolution.clone(),
 
     stencilWrite: true,
     stencilRef: 1,
     stencilZPass: THREE.ReplaceStencilOp,
   });
-  material.lineWidth = lineWidth; // Fix for THREE.js type annotations
+  material.linewidth = lineWidth; // Fix for THREE.js type annotations
```
jtbandes, line 96: is this comment still accurate? Not sure I quite understand what it was getting at...

jtbandes (approved):

PR #6199, opened 2023-06-06
Swap out nuscenes data with one including text annotations

esthersweon:
**User-Facing Changes**
- None

**Description**
- Swap out nuscenes data with one including text annotations 

image


jtbandes (approved):
@esthersweon Is it true that the layout doesn't need to change because the text annotations are on the same topic(s) that are already enabled in the current sample layout?

PR #6200, opened 2023-06-06
[Image Beta] Fix imprecise projection due to linear interpolation

jtbandes:
**User-Facing Changes**
None

**Description**
Images were wiggly due to grid segments being non-coplanar.

It appears there are still some rounding artifacts visible around straight lines in some stories. I'm not sure it's possible to completely avoid these except by making the mesh vertices a perfect rectangle in clip space without any other math (which is not a viable approach for images with distortion). I think this effect will be basically invisible in most real-world images that come from a real camera.

This is the golden story: https://www.chromatic.com/test?appId=603ec8bf7908b500231841e2&id=647fb028151c51e4d183bcaf

image


snosenzo (approved):
👏 😎 👏

PR #6208, opened 2023-06-07
Optimize plot panel blocks processing

foxymiles:
**User-Facing Changes**
Improve performance of plot panel on large datasets.

**Description**
Improve performance of plot panel on large datasets by debouncing blocks processing, optimizing `findTimeRanges`, and total message counting.

When loading blocks we rebuild our entire dataset every time more block data is loaded. Debouncing this eliminates a lot of wasted work. This speeds up loading large datasets by as much as 3x in my testing.

Improve performance of `findTimeRanges` in plotData.ts by taking advantage of the fact that ranges of messages are already sorted by receiveTime.

`findTimeRanges` has significant overhead on large datasets:
Screenshot 2023-06-07 at 6 58 38 AM

Replacing the per-message iteration with code that takes advantage of the fact that message ranges are already sorted seems to eliminate it:
Screenshot 2023-06-07 at 7 10 30 AM

### DIFF
--- a/packages/studio-base/src/panels/Plot/plotData.ts
+++ b/packages/studio-base/src/panels/Plot/plotData.ts
@@ -116,12 +116,6 @@ export function getBlockItemsByPath(
     }
 
     for (const [path, messagePathItems] of Object.entries(messagePathItemsForBlock)) {
-      for (const items of messagePathItems) {
-        for (const item of items) {
-          count += item.queriedData.length;
-        }
-      }
-
       const existingItems = ret[path] ?? [];
       // getMessagePathItemsForBlock returns an array of exactly one range of items.
       const [pathItems] = messagePathItems;
@@ -140,12 +134,14 @@ export function getBlockItemsByPath(
           existingItems.push(pathItems.slice());
         }
       }
+      count += pathItems?.length ?? 0;
       ret[path] = existingItems;
       lastBlockIndexForPath[path] = i;
     }
 
     i += 1;
   }
+
   return ret;
 }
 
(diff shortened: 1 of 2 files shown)


snosenzo (approved):

packages/studio-base/src/panels/Plot/plotData.ts
```diff
@@ -142,12 +134,14 @@ export function getBlockItemsByPath(
           existingItems.push(pathItems.slice());
         }
       }
+      count += pathItems?.length ?? 0;
```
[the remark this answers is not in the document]
foxymiles: Yeah it is a bit confusing. I double checked the count with the reported topic stats and it lines up.

packages/studio-base/src/panels/Plot/usePlotPanelMessageData.ts
```diff
@@ -63,6 +64,12 @@ export function usePlotPanelMessageData(params: Params): Immutable<PlotDataByPat
 
   const cachedGetMessagePathDataItems = useCachedGetMessagePathDataItems(allPaths);
 
+  // Debounce blocks data to improve handling of large blocks.
```
defunctzombie, line 67: It would be helpful to note how or why this improves handling of "large blocks".

How did you select the values of 500 and 2,000? That's worth nothing if those values are meaningful and we need to remember under what circumstances we should look to change them.

Side question: did you try debouncing on the plot panel worker thread render complete?

foxymiles: > How did you select the values of 500 and 2,000? That's worth nothing if those values are meaningful and we need to remember under what circumstances we should look to change them.

These are just empirical values that seem to improve rendering times of large datasets a lot without increasing the latency of smaller sets too much. I guess I could add a comment to that effect.

> Side question: did you try debouncing on the plot panel worker thread render complete?

No not yet. Debouncing earlier in the render cycle seemed more straightforward but there might be better places to do it. The overall result seems pretty good though.

PR #6209, opened 2023-06-07
Fix flaky stories

foxymiles:
**User-Facing Changes**
None

**Description**
Fix flaky stories.

Increase the delay on the chart zoom story and check for mounted state on Chart error update.


jtbandes (approved):

defunctzombie:
Story delays will continue until morale improves.

PR #6217, opened 2023-06-08
Prevent misalignment of autocomplete input and options in settings UI

foxymiles:
**User-Facing Changes**
Fix a bug with dynamic resizing of autocomplete options in the settings UI.

**Description**
Blur the input to hide the autocomplete options when the settings UI is resized to prevent the size of the input and the autocomplete box from getting out of sync.

### DIFF
--- a/packages/studio-base/package.json
+++ b/packages/studio-base/package.json
@@ -173,6 +173,7 @@
     "use-debounce": "9.0.4",
     "use-immer": "0.9.0",
     "use-pan-and-zoom": "0.6.5",
+    "use-resize-observer": "9.1.0",
     "uuid": "9.0.0",
     "webpack": "5.85.1",
     "zustand": "4.3.8"
--- a/packages/studio-base/package.json
+++ b/packages/studio-base/package.json
@@ -173,7 +173,6 @@
     "use-debounce": "9.0.4",
     "use-immer": "0.9.0",
     "use-pan-and-zoom": "0.6.5",
-    "use-resize-observer": "9.1.0",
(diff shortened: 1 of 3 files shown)


jtbandes (commented):
We already have `import { useResizeDetector } from "react-resize-detector";` that we use in a handful of places. Can we use that instead?

foxymiles:
> import { useResizeDetector } from "react-resize-detector";

For some reason this one seems prone to trigger resizeobserver loop limit errors.

foxymiles:
> We already have `import { useResizeDetector } from "react-resize-detector";` that we use in a handful of places. Can we use that instead?

Also using this I'm seeing some spurious callbacks that I'm not seeing with the other library.

foxymiles:
@jtbandes See https://github.com/maslianok/react-resize-detector/issues/241

foxymiles:
Actually with these options it seems to work:

```typescript
  useResizeDetector<HTMLInputElement>({
    handleHeight: false,
    onResize: () => inputRef.current?.blur(),
    refreshMode: "debounce",
    refreshRate: 0,
    skipOnMount: true,
    targetRef: inputRef,
  });```

jtbandes (approved):

defunctzombie:
Was there any hope of keeping the dropdown open and having it adjust its size to the newly available space?

foxymiles:
> Was there any hope of keeping the dropdown open and having it adjust its size to the newly available space?

Difficult since this is such a complex component. IMO it would make more sense to spend the effort on the new message path input dialog we've discussed a few times.

PR #6218, opened 2023-06-08
Fix react nth-child errors

foxymiles:
**User-Facing Changes**
None

**Description**
Fix devtools error on css nth-child selectors.

Screenshot 2023-06-08 at 8 18 47 AM


jtbandes (approved):
I find this particular error to be pretty ridiculous. Last time I looked into it my understanding was that `nth-child` sometimes doesn't work because `<style>` tags are inserted in random places in the dom. Banning `nth-child` because a better solution for style tags couldn't be found seems kinda crazy. On top of that, the error message mentions that this is only an issue for server-side rendering, which is not at all relevant to this application. It seems like ideally we would be able to inform the library that we don't intend to do SSR and it could silence this warning. Oh well, I guess we have to pick our battles.

PR #6219, opened 2023-06-08
Increase delay for chart stories

foxymiles:
**User-Facing Changes**
None

**Description**
Increase delay on chart stories to stabilize them in chromatic.


jtbandes (approved):
delay & pray

PR #6222, opened 2023-06-09
Allow extension panels to seek precisely

foxymiles:
**User-Facing Changes**
Fix an issue with time precision in seeking in extension panels.

**Description**
Fix an issue with time precision in seeking in extension panels.

The existing seek API takes a number as the seek target and then converts this back to a Time object, losing precision. Allow clients to also pass a Time object for greater seek precision to target specific messages.

### DIFF
--- a/packages/studio/src/index.ts
+++ b/packages/studio/src/index.ts
@@ -304,7 +304,7 @@ export type PanelExtensionContext = {
   /**
    * Seek playback to the given time. Behaves as if the user had clicked the playback bar to seek.
    */
-  seekPlayback?: (time: number) => void;
+  seekPlayback?: (time: number | Time) => void;
 
   /**
    * Subscribe to an array of topic names.
--- a/packages/studio/src/index.ts
+++ b/packages/studio/src/index.ts
@@ -302,7 +302,10 @@ export type PanelExtensionContext = {
   setPreviewTime: (time: number | undefined) => void;
 
   /**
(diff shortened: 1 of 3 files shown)


packages/studio/src/index.ts
```diff
@@ -302,9 +302,12 @@ export type PanelExtensionContext = {
   setPreviewTime: (time: number | undefined) => void;
```
jtbandes, line 302: Should we do the same for setPreviewTime?

jtbandes (approved):
Wonder why this was a number in the first place? Do you remember @defunctzombie?

packages/studio/src/index.ts
foxymiles: Maybe although that seems like it would be a bit more extensive change.

defunctzombie:
@jtbandes 

> Wonder why this was a number in the first place? Do you remember @defunctzombie?

Seem like you added it (https://github.com/foxglove/studio/pull/1398) so maybe you can tell me :D My vague guess is that it aligns with "setPreviewTime" which accepts a number and that API traces back to the internal hover-time apis that already existed iirc.

jtbandes:
I think I just did it for consistency with setPreviewTime:

https://github.com/foxglove/studio/pull/1398#discussion_r664805925

image

defunctzombie:
Guess we've learned that having precision was indeed important :D

PR #6225, opened 2023-06-12
Fix user script player recovery from missing input topic errors

foxymiles:
**User-Facing Changes**
Improve user scripts panel recovery from input topic errors.

**Description**
Update user node player to emit a new state when updated node registrations result in a changed topic list. This is necessary to correctly handle recovering from a initial error state that causes no subscriptions to be established and no messages to be emitted from the base player wrapped by UserNodePlayer.

### DIFF
--- a/packages/studio-base/src/players/UserNodePlayer/index.ts
+++ b/packages/studio-base/src/players/UserNodePlayer/index.ts
@@ -814,10 +814,13 @@ export default class UserNodePlayer implements Player {
         this.#playerState.activeData.datatypes,
         this.#memoizedNodeDatatypes,
       );
-      this.#playerState.activeData = {
-        ...this.#playerState.activeData,
-        topics: newTopics,
-        datatypes: newDatatypes,
+      this.#playerState = {
+        ...this.#playerState,
+        activeData: {
+          ...this.#playerState.activeData,
+          datatypes: newDatatypes,
+          topics: newTopics,
+        },
       };
       this.#emitState().catch(console.error);
     }
--- a/packages/studio-base/src/players/UserNodePlayer/index.ts
+++ b/packages/studio-base/src/players/UserNodePlayer/index.ts
@@ -810,9 +810,14 @@ export default class UserNodePlayer implements Player {
         this.#memoizedNodeTopics,
         (top) => top.name,
       );
+      const newDatatypes = this.#getDatatypes(
+        this.#playerState.activeData.datatypes,
+        this.#memoizedNodeDatatypes,
+      );
       this.#playerState.activeData = {
         ...this.#playerState.activeData,
         topics: newTopics,
+        datatypes: newDatatypes,
       };
       this.#emitState().catch(console.error);
     }
--- a/packages/studio-base/src/players/UserNodePlayer/index.ts
+++ b/packages/studio-base/src/players/UserNodePlayer/index.ts
@@ -822,7 +822,7 @@ export default class UserNodePlayer implements Player {
           topics: newTopics,
         },
       };
-      this.#emitState().catch(console.error);
+      void this.#emitState();
     }
   }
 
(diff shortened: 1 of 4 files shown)


packages/studio-base/src/players/UserNodePlayer/index.ts
```diff
@@ -144,6 +144,8 @@ export default class UserNodePlayer implements Player {
     inputsByOutputTopic: new Map(),
   });
 
+  readonly #emitLock = new MutexLocked(undefined);
```
jtbandes, line 147: Just use `import { Mutex } from "async-mutex";` – MutexLocked is for holding and operating on a value

packages/studio-base/src/players/UserNodePlayer/index.ts
```diff
@@ -794,6 +799,28 @@ export default class UserNodePlayer implements Player {
...
+      const newTopics = unionBy(
+        this.#playerState.activeData.topics,
+        this.#memoizedNodeTopics,
+        (top) => top.name,
+      );
+      const newDatatypes = this.#getDatatypes(
+        this.#playerState.activeData.datatypes,
+        this.#memoizedNodeDatatypes,
+      );
+      this.#playerState.activeData = {
```
jtbandes, line 817: Probably we should avoid mutating the player state? Or is this a common pattern within UserNodePlayer?

packages/studio-base/src/players/UserNodePlayer/index.ts
```diff
@@ -794,6 +799,28 @@ export default class UserNodePlayer implements Player {
...
+        ...this.#playerState.activeData,
+        topics: newTopics,
+        datatypes: newDatatypes,
+      };
+      this.#emitState().catch(console.error);
```
jtbandes, line 822: we use `void this.#emitState()` elsewhere in this file, would be nice to keep consistent

foxymiles: The `void promise` pattern seems undesirable to me. Shouldn't we at least log the error?

packages/studio-base/src/players/UserNodePlayer/index.ts
foxymiles: Agree on mutating but there's a lot of it going on in UserNodePlayer.

jtbandes: Does it not work to do `this.#playerState = { ...this.#playerState, activeData: {...` ?

foxymiles: That should work. I'll try that. Will be better for the day when we mark player state as `Immutable`.

packages/studio-base/src/players/UserNodePlayer/index.ts
foxymiles: Actually I guess `void promise` is better here since it will throw if it fails instead of logging to console.

jtbandes (approved):

packages/studio-base/src/players/UserNodePlayer/index.ts
```diff
@@ -794,6 +800,31 @@ export default class UserNodePlayer implements Player {
...
+      );
+      this.#playerState = {
+        ...this.#playerState,
+        activeData: {
+          ...this.#playerState.activeData,
+          datatypes: newDatatypes,
+          topics: newTopics,
+        },
+      };
+      void this.#emitState();
```
defunctzombie, line 826: This looks suspect - in other code paths we make sure to wait on and emit to finish before starting another one. It seems that here this is _ok_ because of the exclusive lock we have in emit state? The confusing thing about this is that you don't necessarily emit the playerState you set above this line since you might be blocked waiting for another emit to finish and by that time the player state is set to another value. If that behavior is acceptable here, then maybe changing the function name to `#queueEmitState` is more appropriate to convey that it is ok to use `void` here?

PR #6229, opened 2023-06-12
Allow assets with `package://` URIs to be fetched through foxglove websocket

achim-k:
**User-Facing Changes**
- Allow assets with `package://` URIs to be fetched through foxglove websocket

**Description**
- Adds a thin asset management layer (based on #5386) which allows to fetch assets through the data source. For now, only assets with `package://` URIs are attempted to be fetched through the data source. All other assets with e.g. http(s):// or file:// schemes are fetched using the builtin `fetch` function (as was the case so far).
- Implements the asset fetching capability for the foxglove websocket data source (see also https://github.com/foxglove/ws-protocol/pull/462)
- Adapts the URDF renderable and the ModelCache to use the asset management layer

Overall, this PR improves UX when working with URDF files which often reference mesh files that may not accessible by Studio e.g. when Studio is running in the browser or when the meshes are located on a remote server.


packages/studio-base/src/players/IterablePlayer/Mcap/McapIndexedIterableSource.ts
```diff
@@ -207,4 +207,20 @@ export class McapIndexedIterableSource implements IIterableSource {
     messages.sort((a, b) => compare(a.receiveTime, b.receiveTime));
     return messages;
   }
+
+  public async fetchAsset(name: string): Promise<Asset> {
+    const cleanName = name.replace("mcap-attachment://", "");
```
achim-k, line 212: This is a bit of a hack. The custom layer url below has to be a valid URL: https://github.com/foxglove/studio/blob/504c4511472d85e49bb139f10148f048ab25f732/packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts#L650

If that would be only `robot_description` (as the attachment is named), this would raise an exception, hence this workaround.

amacneil:
I would probably simplify the url scheme to just `mcap://attachmentname.urdf`, unless there is some other meaning planned for that.

packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
```diff
@@ -291,6 +291,13 @@ export function ThreeDeeRender(props: {
     }
   }, [context.dataSourceProfile, renderer]);
 
+  // Tell the renderer if the data source supports asset fetching
+  useEffect(() => {
+    if (renderer) {
+      renderer.fetchAsset = context.fetchAsset;
+    }
+  }, [context.fetchAsset, renderer]);
+
```
achim-k, line 300: Maybe we could do something similar here for telling renderables what kind of assets can be fetched. This could be e.g. be used in [isValidUrl](https://github.com/foxglove/studio/blob/c2323d05b0f49d244fd5f275f1f54a1f92a07324/packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts#L823-L832) to determine if `package://...` is a valid URDF URL

achim-k:
> I would probably simplify the url scheme to just `mcap://attachmentname.urdf`, unless there is some other meaning planned for that.

Or maybe just `asset://` to make it more data source independent?

defunctzombie:
Gave this a try and it did what it says it would do which is good. I did create this new issue in the project: https://linear.app/foxglove/issue/FG-3915/list-assets-in-sidebar as part of my experience.

Regarding the `fetchAsset` api surface area. I still think that `AssetInfo` type is missing an `id` and that asset fetching should be done by _id_ and it is up to the panel to lookup the correct asset by name - however I can see why this might be annoying and worth discussing. One thing to note is that there may be multiple assets with the same name for a given data source which is why I push for panels to make the decision on which _id_ to fetch as they might need to find an asset by id+timestamp.

packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
```diff
@@ -285,6 +285,13 @@ export function ThreeDeeRender(props: {
     }
   }, [context.dataSourceProfile, renderer]);
 
+  // Tell the renderer if the data source supports asset fetching
+  useEffect(() => {
+    if (renderer) {
+      renderer.fetchAsset = context.fetchAsset;
```
defunctzombie, line 291: Will this result in requests for `http://` assets also going to the `fetchAsset` function? If it does then I continue to be 👎 on this but willing to discuss cause I can see there being some interesting aspects like being able to intercept the http requests.

packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
```diff
@@ -487,6 +487,48 @@ export class Urdfs extends SceneExtension<UrdfRenderable> {
...
+            hasError ? `: ${err.message}` : ""
+          }`;
+          this.renderer.settings.errors.add(["layers", instanceId], FETCH_URDF_ERR, errMessage);
+        });
+    };
+
+    if (this.renderer.fetchAsset) {
+      let assetUrl: string | undefined;
+      this.renderer
+        .fetchAsset(url)
```
defunctzombie, line 514: Related to my http(s) handling comment above - this is something that I questioned in the first PR and needs architectural discussion. It also relates to whether the panel fetches by name or looks up an _id_ in the available asset lists and fetches that.

defunctzombie:
Is the idea that in our app you can specify either `package://ur_description/meshes/ur5e/visual/base.dae` or `mcap-attachment://ur_description/meshes/ur5e/visual/base.dae` ?

My broader question here is around how folks who are not using ROS are storing URDF and model references?

packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
```diff
@@ -285,6 +285,13 @@ export function ThreeDeeRender(props: {
...
+    }
+  }, [context.fetchAsset, renderer]);
```
snosenzo, line 293: do we need a `context.watch(fetchAsset)` down below?

packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
achim-k: I think the data source should be able to tell what kind of assets can be fetched. For MCAPs we know all available assets from the start and for foxglove websocket, maybe the server could tell Studio what kind of asset names it supports (e.g. `package://` or `file://`)?

packages/studio/src/index.ts
```diff
@@ -127,6 +127,28 @@ export type MessageEvent<T = unknown> = {
...
+ */
+export type AssetInfo = {
+  /** A unique identifier for this asset. */
+  name: string;
+  /** Asset size in bytes. */
+  sizeInBytes: number;
+  /** The type of asset. */
+  mediaType: string;
+  /** Last modified time of the asset. */
+  lastModified: Time;
```
achim-k, line 141: rm

achim-k:
This PR is ready for another round of review. Note that we have changed the scope to only add fetch-asset support for foxglove websocket. I have adapted the PR description accordingly.

packages/studio-base/src/players/IterablePlayer/Mcap/McapIterableSource.ts
```diff
@@ -114,4 +115,15 @@ export class McapIterableSource implements IIterableSource {
...
+  public async fetchAsset(name: string): Promise<Asset> {
+    if (this.#sourceImpl == undefined) {
+      throw new Error(`Source is not initialized`);
+    }
+    if (this.#sourceImpl.fetchAsset == undefined) {
+      throw new Error(`Source does not support asset fetching`);
+    }
+
+    return await this.#sourceImpl.fetchAsset(name);
+  }
```
jtbandes, line 128: Seems like ideally this method would be a complete pass-through of the `impl` method with no extra logic. Is there any way to achieve that?

packages/studio-base/src/players/IterablePlayer/WorkerIterableSourceWorker.ts
```diff
@@ -54,6 +54,14 @@ export class WorkerIterableSourceWorker implements IIterableSource {
     const cursor = new IteratorCursor(iter, abort);
     return Comlink.proxy(cursor);
   }
+
+  public async fetchAsset(name: string): Promise<Asset> {
```
jtbandes, line 58: also would ideally be a transparent pass-through without a `throw`

packages/studio/src/index.ts
```diff
@@ -375,6 +392,15 @@ export type PanelExtensionContext = {
    */
   callService?(service: string, request: unknown): Promise<unknown>;
 
+  /**
+   * Fetch an asset from Studio's asset management layer.
```
jtbandes, line 396: Should we say "asset manager"? "asset management layer" seem pretty wordy. Or are there any other proposed names for this concept?

packages/studio/src/index.ts
```diff
@@ -127,6 +127,23 @@ export type MessageEvent<T = unknown> = {
...
+  /** Binary asset data. */
+  data: Uint8Array;
+
+  mediaType?: string;
+};
+
+/**
+ * Function to fetch an asset from Studio's asset management layer.
+ */
+export type FetchAssetFn = (uri: string, options?: { signal: AbortSignal }) => Promise<Asset>;
```
jtbandes, line 145: Need docs on `options` as well - what can the user expect to happen if they call `signal.abort()`?

packages/studio/src/index.ts
```diff
@@ -127,6 +127,23 @@ export type MessageEvent<T = unknown> = {
...
+/**
+ * An asset loaded from Studio's asset management layer.
+ */
+export type Asset = {
+  /** A unique identifier for this asset. */
+  name: string;
+  /** Binary asset data. */
+  data: Uint8Array;
+
+  mediaType?: string;
```
jtbandes, line 139: needs comment

packages/studio/src/index.ts
```diff
@@ -127,6 +127,23 @@ export type MessageEvent<T = unknown> = {
   originalMessageEvent?: MessageEvent;
 };
 
+/**
+ * An asset loaded from Studio's asset management layer.
+ */
+export type Asset = {
+  /** A unique identifier for this asset. */
```
jtbandes, line 134: Do we want to claim this is a "unique identifier" or just that it corresponds with the requested name? We don't have this in the response message in the protocol, so I guess the name will equal whatever uri was passed in?

packages/studio-base/src/components/PanelExtensionAdapter/types.ts
```diff
@@ -0,0 +1,38 @@
...
+  /** Asset URI, corresponds to the URI with which the asset was retrieved. */
+  uri: string;
+  /** Binary asset data. */
+  data: Uint8Array;
+  /** Asset type. */
+  mediaType?: string;
+};
+
+/**
+ * PrivateExtensionContext adds additional built-in only functionality to the PanelExtensionContext.
```
defunctzombie, line 20:
```suggestion
 * BuiltinPanelExtensionContext adds additional built-in only functionality to the PanelExtensionContext.
```

defunctzombie:
Player changes LGTM - I'll let Jacob and Sam review and 👍 this since the majority of the changes are in the 3d code. Since we aren't exposing any public interfaces here we will be able to build on this to make sure the interfaces support the workflows we need for data platform and local files.

Make sure you give this a test with the desktop app and confirm it can still open `package://` and `file://` urls when using a local file data source.

achim-k:
> Make sure you give this a test with the desktop app and confirm it can still open `package://` and `file://` urls when using a local file data source.

Can confirm that this is still working.

packages/studio-base/src/components/PanelExtensionAdapter/types.ts
```diff
@@ -0,0 +1,38 @@
...
+   *
+   * The asset manager will determine how to fetch the asset. I.E. http(s) uris will use http requests
+   * while other schemes may fall back to the data source.
+   *
+   * @param uri URI identifying the asset
+   * @param options Optional abort signal that allows to abort fetching of the asset. Note that this
+   * might not be supported by all fetching methods.
+   * @returns
+   */
+  unstable_fetchAsset: (uri: string, options?: { signal: AbortSignal }) => Promise<Asset>;
```
snosenzo, line 37: Should we have a note on why it's "unstable"?

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -223,6 +229,7 @@ export class Renderer extends EventEmitter<RendererEvents> implements IRenderer
     this.interfaceMode = interfaceMode;
     this.#canvas = canvas;
     this.config = config;
+    this.#fetchAsset = fetchAsset;
```
snosenzo, line 232: can we also release this reference on `dispose`? just want to ensure we don't keep references that could create a memory leak.

packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
```diff
@@ -684,6 +695,17 @@ export default class FoxgloveWebSocketPlayer implements Player {
 
       this.#emitState();
     });
+
+    this.#client.on("fetchAssetResponse", (response) => {
```
snosenzo, line 699: does this need a corresponding `off`? or is it just cleared somewhere I can't see

snosenzo (approved):
looks good for the most part, just wondering about cleanup in some places. Also swapping the `throw Error`s for `assert`s would read a bit better to me, but not necessary

packages/studio-base/src/components/PanelExtensionAdapter/types.ts
defunctzombie: It is unstable cause it is not exposed in the public types for the `@foxglove/studio` package and for users that might stumble upon it by console logging the context we want to send a clear signal that it is not to be used.

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
achim-k: Could you explain how I can do that? Shouldn't this then also be done for `this.#canvas`?
I would assume that this member variable would get disposed automatically when the renderer instance is deleted?

packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
achim-k: We don't call any `off` in the player and I'm not sure if we have to, since we delete the entire client instance when the connection is closed:
https://github.com/foxglove/studio/blob/06891203b9975ea3b11eb1ffdae93103baaf1498/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts#L252-L253

packages/studio-base/src/panels/ThreeDeeRender/ModelCache.ts
```diff
@@ -105,7 +105,7 @@ export class ModelCache {
 
     // Check if this is a STL file based on content-type or file extension
     if (STL_MIME_TYPES.includes(contentType) || /\.stl$/i.test(url)) {
-      return this.#loadSTL(url, buffer, this.options.meshUpAxis);
+      return this.#loadSTL(url, buffer.buffer.slice(buffer.byteOffset), this.options.meshUpAxis);
```
achim-k, line 108: Would like to avoid this copy here, but not sure how. It otherwise doesn't work if the Uint8Array has a byteOffset > 0

packages/studio-base/src/components/PanelExtensionAdapter/types.ts
snosenzo: But does `unstable_` mean that it could be revealed in the future?
Is `private_`  a better descriptor here if we're not planning on releasing it?

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
achim-k: Would require me to change the type of `this.#fetchAsset` then as well. Not sure if it's worth it?

image

packages/studio-base/src/components/PanelExtensionAdapter/types.ts
jtbandes: I believe we are thinking that we'll eventually make it available to extensions if we determine that we're comfortable with the API and it works well.

packages/studio-base/src/panels/ThreeDeeRender/ModelCache.ts
jtbandes: Is that an error with the three.js STLLoader? Probably the "right" thing to do is make a PR upstream to fix it :)

jtbandes: At least for now we should add a comment to explain why this slice is needed

achim-k: > Is that an error with the three.js STLLoader? Probably the "right" thing to do is make a PR upstream to fix it :)

I wouldn't say it's an error, the API just expects you to pass in an ArrayBuffer instance where the first byte is already part of the STL. This wouldn't be the case if we have a byteOffset > 0. 

```suggestion
      // Create a copy of the array buffer to respect the `byteOffset` value as the
      //  underlying three.js STLLoader only accepts an ArrayBuffer instance.
      return this.#loadSTL(url, buffer.buffer.slice(buffer.byteOffset), this.options.meshUpAxis);
```

jtbandes: We probably need to respect `buffer.byteLength` then too, right?

achim-k: > We probably need to respect buffer.byteLength then too, right?

You are probably right :sweat_smile: 

```suggestion
      // Create a copy of the array buffer to respect the `byteOffset` and `byteLength` value as
      // the underlying three.js STLLoader only accepts an ArrayBuffer instance.
      return this.#loadSTL(
        url,
        buffer.buffer.slice(buffer.byteOffset, buffer.byteOffset + buffer.byteLength),
        this.options.meshUpAxis,
      );
```

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
snosenzo: don't worry about it then

packages/studio-base/src/components/PanelExtensionAdapter/types.ts
defunctzombie: Yep - once we believe the interface meets the various use cases we will make it public: https://linear.app/foxglove/issue/FG-2104/make-the-fetchasset-extension-interface-public

defunctzombie:
@achim-k There are 6 story changes that look potentially questionable. Do you have access to see what those are or need one of us to show you?

achim-k:
> @achim-k There are 6 story changes that look potentially questionable. Do you have access to see what those are or need one of us to show you?

Looking into it. Some of them look like a potential regression

achim-k:
> @achim-k There are 6 story changes that look potentially questionable. Do you have access to see what those are or need one of us to show you?

Should be fixed. The only remaining visual change seems to be a flaky story.

packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
```diff
@@ -240,10 +236,9 @@ export class Urdfs extends SceneExtension<UrdfRenderable> {
     for (const [instanceId, layerConfig] of Object.entries(this.renderer.config.layers)) {
       if (layerConfig?.layerId === LAYER_ID) {
         const config = layerConfig as Partial<LayerSettingsCustomUrdf>;
-        const placeholder = supportsPackageUrl ? "package://" : undefined;
-        const help = supportsPackageUrl
-          ? "package:// URL or http(s) URL pointing to a Unified Robot Description Format (URDF) XML file"
-          : "http(s) URL pointing to a Unified Robot Description Format (URDF) XML file";
+        const placeholder = "package://";
+        const help =
+          "package:// URL or http(s) URL pointing to a Unified Robot Description Format (URDF) XML file";
```
jtbandes, line 241: what will happen if a user enters a package:// url if the data source does not support it?

would it be right to check whether the player has the "assets" capability?

I guess even if the player doesn't support assets, in the desktop app, we could still support package urls...

packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
```diff
@@ -666,14 +660,27 @@ export class Urdfs extends SceneExtension<UrdfRenderable> {
       this.renderer.addTransform(parent, child, 0n, translation, rotation, settingsPath);
     }
   }
-}
 
-async function parseUrdf(text: string): Promise<ParsedUrdf> {
-  const fileFetcher = getFileFetch();
+  #getFileFetch(): (url: string) => Promise<string> {
```
jtbandes, line 664: why is this a function that returns a function instead of just a function?

packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
```diff
@@ -132,6 +136,9 @@ export default class FoxgloveWebSocketPlayer implements Player {
   #subscribedTopics?: Map<string, Set<string>>;
   #advertisedServices?: Map<string, Set<string>>;
   #nextServiceCallId = 0;
+  #nextAssetRequestId = 0;
+  #fetchAssetRequests = new Map<number, (response: FetchAssetResponse) => void>();
+  #fetchedAssets = new Map<string, Asset>();
```
jtbandes, line 141: might be nice to store `Promise<Asset>` as soon as we start the request, so if 2 panels ask for the same asset, we only send 1 request

jtbandes (approved):

packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
achim-k: > what will happen if a user enters a package:// url if the data source does not support it?

This is handled here: https://github.com/foxglove/studio/blob/18bcd4757b638a2ca2986400f13615cc748ea7e6/packages/studio-base/src/components/MessagePipeline/store.ts#L121-L144

The `fetch` will raise an Exception in the end that `package://` is not supported (if not the desktop app)

> would it be right to check whether the player has the "assets" capability?

I realized that this is missing in the above referenced function, I currently only check if the player has `fetchAsset` defined. Is there a way to access player capabilities in that place?

packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
achim-k: Initially tried this but resulted in a scope issue:
image

Will change it to use an arrow function, looks cleaner that way, I agree.

jtbandes: `this.#getFileFetch.bind(this)`?

packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
jtbandes: Not sure if it's right to check there or check on the caller side. If the API is not clear, maybe we should add comments to clarify it. It seems like this spot would have access to `get().lastCapabilities`...

jtbandes: You could follow this pattern: https://github.com/foxglove/studio/blob/e9b629d6739bf9dd60c4934d25e8d32912b5b4de/packages/studio-base/src/components/MessagePipeline/store.ts#L284-L299

jtbandes:
cc @defunctzombie for opinions on player/message pipeline APIs

defunctzombie (approved):

PR #6237, opened 2023-06-13
Move publish panel inputs into panel settings

2metres:
**User-Facing Changes**
Screen Shot 2023-06-15 at 10 06 46 am

Also add placeholder to Autocomplete fields in PanelSettings

**Description**
Moves Topic and Schema inputs into panel settings and renames datatypes to message schema

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,6 +78,7 @@
     "@storybook/addon-actions": "7.0.20",
     "@storybook/addon-essentials": "7.0.20",
     "@storybook/addon-interactions": "7.0.20",
+    "@storybook/csf-tools": "7.0.20",
     "@storybook/react": "7.0.20",
     "@storybook/react-webpack5": "7.0.20",
     "@types/babel__core": "^7.20.1",
--- a/package.json
+++ b/package.json
@@ -78,6 +78,7 @@
     "@storybook/addon-actions": "7.0.20",
     "@storybook/addon-essentials": "7.0.20",
     "@storybook/addon-interactions": "7.0.20",
+    "@storybook/jest": "0.1.0",
     "@storybook/react": "7.0.20",
     "@storybook/react-webpack5": "7.0.20",
     "@types/babel__core": "^7.20.1",
--- a/package.json
+++ b/package.json
@@ -78,7 +78,6 @@
     "@storybook/addon-actions": "7.0.20",
     "@storybook/addon-essentials": "7.0.20",
     "@storybook/addon-interactions": "7.0.20",
-    "@storybook/csf-tools": "7.0.20",
     "@storybook/react": "7.0.20",
     "@storybook/react-webpack5": "7.0.20",
     "@types/babel__core": "^7.20.1",
--- a/package.json
+++ b/package.json
@@ -78,7 +78,6 @@
     "@storybook/addon-actions": "7.0.20",
     "@storybook/addon-essentials": "7.0.20",
     "@storybook/addon-interactions": "7.0.20",
-    "@storybook/jest": "0.1.0",
     "@storybook/react": "7.0.20",
     "@storybook/react-webpack5": "7.0.20",
     "@types/babel__core": "^7.20.1",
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -80,6 +80,7 @@
     "@storybook/addon-actions": "7.0.20",
     "@storybook/addon-essentials": "7.0.20",
     "@storybook/addon-interactions": "7.0.20",
+    "@storybook/csf-tools": "7.0.20",
```
jtbandes, line 83: what's this for?

> An experimental library to read, analyze, transform, and write CSF programmatically.

packages/studio-base/src/panels/Publish/index.tsx
```diff
@@ -188,37 +198,9 @@ function Publish(props: Props) {
...
-  }, [actionHandler, props.config, updatePanelSettingsTree]);
-
-  const onChangeTopic = useCallback(
-    (_event: unknown, name: string) => {
-      saveConfig({ topicName: name });
-    },
-    [saveConfig],
-  );
-
-  // when a known topic is selected, also fill in its datatype
```
jtbandes, line 202: Was the deletion of this logic intentional?

package.json
2metres: Attempting to debug the broken build, will remove now

packages/studio-base/src/panels/Publish/index.tsx
2metres: I think so. Now that we handle the topic selection in the panel settings we don't need to update the config from within panel

jtbandes: I think it would still be nice to auto select the schema when selecting a topic.

jtbandes:
I guess it's not new, but there's a strange height difference of autocomplete compared to other settings field types:

image

jtbandes (commented):
Commented in Slack with some feedback

packages/studio-base/src/panels/Publish/index.tsx
2metres: I tried doing that in the panelSettings as well but caused a bug so removed it

2metres: I've almost got it working where it auto selects the correct schema based on the Topic, Its currently off by one. I could use some immer help here

defunctzombie:
This PR is also changing the public extension api and something with the FieldEditor but does not call that out in the description. Could you elaborate on these changes since they become our public interface.

packages/studio-base/src/panels/Publish/index.tsx
```diff
@@ -117,190 +89,95 @@ function parseInput(value: string): { error?: string; parsedObject?: unknown } {
...
-  );
-
-  const onSelectDatatype = useCallback(
-    (newDatatype: string, autocomplete: IAutocomplete) => {
-      saveConfig({ datatype: newDatatype });
-      autocomplete.blur();
-    },
-    [saveConfig],
-  );
+  usePublishPanelSettings(config, saveConfig, schemaNames, topics, datatypes);
```
jtbandes, line 113: seems like schemaNames is only used inside usePublishPanelSettings, it should move in there instead of being a parameter

jtbandes (changes requested):
Getting close. However the panel still crashes when I publish without a schema selected.

image

2metres:
@achim-k can you please test that this panel still publishes correctly?

jtbandes (approved):

packages/studio-base/src/panels/Publish/settings.ts
```diff
@@ -0,0 +1,145 @@
...
+  config: PublishConfig,
+  schemaNames: string[],
+  topics: readonly Topic[],
+): SettingsTreeNodes => ({
+  general: {
+    fields: {
+      topicName: {
+        label: "Topic",
+        input: "autocomplete",
+        placeholder: "/some_topic",
```
jtbandes, line 48: I think this placeholder and `std_msgs/Type` below are sorta arbitrary and might be confusing to non-ROS users. Could just remove them and have no placeholder.

PR #6238, opened 2023-06-13
Plot setting to turn off lines

wimagguc:
**User-Facing Changes**

Introduces a setting for the Plot panel, to show/hide lines:

demo

**Description**

Not 100% sure whether this switch is more a per-dataset or a per-panel setting, but to err on the side of more granular config, we can show/hide lines for each individual dataset:

https://github.com/foxglove/studio/assets/349687/96097d86-a1da-4161-9d6e-64adeea7c14d


foxymiles (approved):
Looks good. Maybe we could add a story or two?

jtbandes (approved):

PR #6239, opened 2023-06-13
Support style properties in geojson objects

foxymiles:
**User-Facing Changes**
Support style properties in geojson objects in the map panel.

**Description**
Support style properties in geojson objects in the map panel. 

We passthrough the `style` property but the supported values are documented here:
https://leafletjs.com/reference.html#path


packages/studio-base/src/panels/Map/index.stories.tsx
```diff
@@ -439,7 +461,16 @@ export const GeoJSON: StoryObj = {
             schemaName: "foxglove.GeoJSON",
             message: {
               geojson: JSON.stringify(
-                makeGeoJsonMessage({ lat: GeoCenter.lat - 0.1, lon: GeoCenter.lon - 0.1 }),
+                makeGeoJsonMessage({
+                  center: { lat: GeoCenter.lat - 0.1, lon: GeoCenter.lon - 0.1 },
+                  key: "/geo2",
```
snosenzo, line 466: can we name the keys/topics something that tells the viewer a bit about what they should see?

packages/studio-base/src/panels/Map/index.stories.tsx
```diff
@@ -46,12 +54,19 @@ function makeGeoJsonMessage(center: { lat: number; lon: number }) {
             [0.1 + center.lon, 0.1 + center.lat],
           ],
         },
+        properties: {
+          name: `Named Line ${key}`,
+          style: {
+            color: "#ff0000",
+            dashArray: "4 4",
```
snosenzo, line 61: are we using all possible geojson style properties? might be good to have examples of all of them just in case

snosenzo (approved):

packages/studio-base/src/panels/Map/index.stories.tsx
foxymiles: There's a bunch of these and they might add more at some point so since this is just a passthrough I don't think we need to test them all.

PR #6241, opened 2023-06-13
Increase the delay for the WithMinAndMaxYValues plot story

foxymiles:
**User-Facing Changes**
None

**Description**
Increase the delay for the WithMinAndMaxYValues plot story.

### DIFF
--- a/packages/studio-base/src/panels/Plot/index.stories.tsx
+++ b/packages/studio-base/src/panels/Plot/index.stories.tsx
@@ -790,6 +790,9 @@ export const WithMinAndMaxYValues: StoryObj = {
   },
 
   parameters: {
+    chromatic: {
+      delay: 500,
+    },
     colorScheme: "light",
     useReadySignal: true,
   },


jtbandes (approved):

defunctzombie:
What's with all these being flaky all of the sudden? We use the signal ready stuff for these do we not?

foxymiles:
> What's with all these being flaky all of the sudden? We use the signal ready stuff for these do we not?

Not sure. There may have been some changes on the Chromatic side lately or it may have been triggered by upgrading to newer versions of storybook. I don't think Chromatic is going to wait for our signal ready though.

The mechanism they suggest for controlling this more precisely is using `waitFor` in a play function:
https://www.chromatic.com/docs/delay

jtbandes:
I think we also made some changes to the Plot panel a little while ago, didn't we?

We normally use our `readySignal` to capture the specific number of renders we expect to happen while the panel settles. Maybe we need to bump those ready signal count numbers up?

foxymiles:
> We normally use our `readySignal` to capture the specific number of renders we expect to happen while the panel settles. Maybe we need to bump those ready signal count numbers up?

The readySignal isn't going to control when Chromatic decides it's waited long enough to take a snapshot right? All of these flaky stories are clearly chromatic taking a snapshot before some slow rendering component of ours is done. From the docs it looks to me like if we want to control these more precisely than just increasing delay we need to await things in the `play` function.

jtbandes:
> The readySignal isn't going to control when Chromatic decides it's waited long enough to take a snapshot right? ... we need to await things in the play function.

Yes, that's what this line does:

https://github.com/foxglove/studio/blob/62fd4868d594bafe17f5b67bea8dc700a6091042/packages/studio-base/src/components/Chart/index.stories.tsx#L159-L161

It connects the `play` function so it waits on the `readySignal`.

PR #6243, opened 2023-06-13
Distortion patch fix for raitonal-polynomial model

saching13:
**User-Facing Changes**

**Description**
This PR adds the capability to take account of k4, k5 and k6 parameters of the distortion coefficients in undisotrting and projecting the camera view on the 3D Panel and the new Image beta panel.

Refer [here](https://docs.opencv.org/3.4/d9/d0c/group__calib3d.html) for more details on the Distortion model

### DIFF
--- a/packages/den/image/PinholeCameraModel.ts
+++ b/packages/den/image/PinholeCameraModel.ts
@@ -133,8 +133,8 @@ export class PinholeCameraModel {
   // Question: Do I need this ut as input if I'm returning ?
   // I added it just to keep the similarity with other functions.
   // I don't see the use of it.
-  public undistort (out: Vector2, pixel: Readonly<Vector2>, iterations = 5): Vector2 {
-    const D  = this.D;
+  public undistort(out: Vector2, pixel: Readonly<Vector2>, iterations = 5): Vector2 {
+    const D = this.D;
     const [k1, k2, p1, p2, k3, k4, k5, k6] = D;
 
     let x0 = pixel.x;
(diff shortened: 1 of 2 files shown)


packages/den/image/PinholeCameraModel.ts
```diff
@@ -271,34 +328,20 @@ export class PinholeCameraModel {
...
-    const k1 = D[0]!;
-    const k2 = D[1]!;
-    const p1 = D[2]!;
-    const p2 = D[3]!;
-    const k3 = D[4]!;
-    let barrel_correction = 1 + k1 * r2 + k2 * r4 + k3 * r6;
-    barrel_correction /= 1.0 + D[5] * r2 + D[6] * r4 + D[7] * r6;
-    const xpp = xp * barrel_correction + p1 * a1 + p2 * (r2 + 2 * (xp * xp));
-    const ypp = yp * barrel_correction + p1 * (r2 + 2 * (yp * yp)) + p2 * a1;
+    // Question? This is not R^-1. R is in Row major order as per CameraInfo.
```
snosenzo, line 331: I think your change looks correct

snosenzo (commented):
Thanks so much for doing this! 
Was wondering though: do you have an example log file with distorted that we would be able to test this with?

saching13:
> Thanks so much for doing this! Was wondering though: do you have an example log file with distorted that we would be able to test this with?

It's in [this](https://foxglove.slack.com/archives/C01T10L5QJX/p1686174611410229?thread_ts=1686173837.831869&cid=C01T10L5QJX) slack thread. LMK if you want me to share it over G-drive or something

packages/den/image/PinholeCameraModel.ts
```diff
@@ -130,6 +130,38 @@ export class PinholeCameraModel {
     }
   }
 
+  // Question: Do I need this ut as input if I'm returning ?
```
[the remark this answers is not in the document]
saching13: Got it. Done.

FabienDanieau:
Hi, I'd love to have this in the next release. Let me know if I can help to close this PR

packages/den/image/PinholeCameraModel.ts
jtbandes: I agree. Confirmed that the [original source](http://docs.ros.org/en/diamondback/api/image_geometry/html/c++/pinhole__camera__model_8cpp_source.html) was expecting row-major storage (documented at https://docs.opencv.org/4.x/d3/d63/classcv_1_1Mat.html).

jtbandes: Also, `R^-1 = R^T`, source: https://math.stackexchange.com/a/2895897

packages/den/image/PinholeCameraModel.ts
```diff
@@ -271,34 +325,18 @@ export class PinholeCameraModel {
...
+
+    out = this.distort(out, normalizedPoint);
 
     // map_x(u,v) <- x''fx + cx
     // map_y(u,v) <- y''fy + cy
     // cx, fx, etc. come from original camera matrix K.
-    out.x = xpp * K[0] + K[2];
-    out.y = ypp * K[4] + K[5];
+    out.x = out.x * fx + cx;
+    out.y = out.y * fy + cy;
```
jtbandes, line 339: Thanks for the work. I think most of the changes make sense to me. I am not sure I understand why the `K` values were replaced with `P` values here. Can you explain?

saching13: you are right. my bad. Fixed it.

PR #6249, opened 2023-06-14
Update time/duration in protobufDefinitionsToDatatypes to matching deserialized object

jtbandes:
**User-Facing Changes**
Fixed the `google.protobuf.Duration` and `google.protobuf.Timestamp` generated type definitions in User Scripts so they match the way timestamps are represented in the rest of the app (`{sec, nsec}` instead of `{seconds, nanos}`).

**Description**
We modify the deserialization logic for Duration and Timestamp so we get `{sec, nsec}`, the same format as the rest of Studio expects. However, the generated types for User Scripts still showed the names as `{seconds, nanos}`.

### DIFF
--- a/packages/mcap-support/src/parseProtobufSchema.test.ts
+++ b/packages/mcap-support/src/parseProtobufSchema.test.ts
@@ -194,6 +194,39 @@ describe("parseProtobufSchema", () => {
       }
     `);
 
+    expect(sceneUpdateChannel.datatypes.get("google.protobuf.Timestamp")).toMatchInlineSnapshot(`
+      {
+        "definitions": [
+          {
+            "isArray": false,
+            "name": "sec",
+            "type": "int64",
+          },
+          {
+            "isArray": false,
+            "name": "nsec",
+            "type": "int32",
+          },
+        ],
+      }
+    `);
+    expect(sceneUpdateChannel.datatypes.get("google.protobuf.Duration")).toMatchInlineSnapshot(`
+      {
+        "definitions": [
+          {
+            "isArray": false,
+            "name": "sec",
+            "type": "int64",
+          },
+          {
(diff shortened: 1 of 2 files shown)


defunctzombie (approved):
I am sad that this mucks with _all_ instances of `google.protobuf.Timestamp` and `google.protobuf.Duration` regardless of it coming from a foxglove schema or not but I do understand how it simplifies some downstream behaviors in Studio.

Am I reading this correctly that it is scoped to only `google.protobuf.Timestamp` and `google.protobuf.Duration` rather than any object with field names "seconds" and "nanos"? It was not obvious to me what line actually limits to those built-in protobuf types.

Please make a follow-up linear issue to update our message schema docs to call-out this protobuf nuance. I believe a similar nuance happens with ROS 2 `builtin_interfaces/Time`? And would be worth mentioning as well. We can discuss on the docs PR where best to include this nuance but it does seem worth documenting so we can point to it when someone brings their custom protobuf message and it looks "wrong" in the raw message panel.

A consideration that should not hold up this PR: when writing a message converter a user will think their _input_ schema matches the field names they see in the .proto files but this will not actually be true. Since message converters are written outside of studio they do not benefit from the "auto generated" types of user-scripts (nor will user scripts once we allow writing outside of studio).

snosenzo (approved):
Want to echo roman's concern that it's not evident this is being run over only `Time` and `Duration` types. Would be good to maybe have a check in the function that asserts that

jtbandes:
Would yall have had a different opinion if this bit of the code had been visible in the diff?

https://github.com/foxglove/studio/blob/15edcf699c57c64868d76dc46d8117e877ab02f4/packages/mcap-support/src/parseProtobufSchema.ts#L64-L65

PR #6252, opened 2023-06-14
wip

defunctzombie:
**User-Facing Changes**

**Description**

### DIFF
--- a/packages/studio-base/src/PanelAPI/useBlocksByTopic.ts
+++ b/packages/studio-base/src/PanelAPI/useBlocksByTopic.ts
@@ -11,7 +11,7 @@
 //   found at http://www.apache.org/licenses/LICENSE-2.0
 //   You may not use this file except in compliance with the License.
 
-import memoizeWeak from "memoize-weak";
+// import memoizeWeak from "memoize-weak";
 import { useCallback, useEffect, useMemo, useState } from "react";
 import { v4 as uuidv4 } from "uuid";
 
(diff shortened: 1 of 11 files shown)


defunctzombie:
Next step is figuring out how to better reset the store in message pipeline when a player changes. The _bind_ calls are hanging into things they shouldn't be

packages/studio-base/src/PanelAPI/useBlocksByTopic.ts
```diff
@@ -34,27 +34,28 @@ export type MessageBlock = Immutable<{
...
 // That said, MessageBlock identity will change when the set of topics changes, so consumers should
 // prefer to use the identity of topic-block message arrays where possible.
-const filterBlockByTopics = memoizeWeak(
-  (block: Immutable<PlayerMessageBlock> | undefined, topics: readonly string[]): MessageBlock => {
-    if (!block) {
-      // For our purposes, a missing MemoryCacheBlock just means "no topics have been cached for
-      // this block". This is semantically different to an empty array per topic, but not different
-      // to a MemoryCacheBlock with no per-topic arrays.
-      return {};
+const filterBlockByTopics = (
```
defunctzombie, line 37: Dropping this memo reduces heap snapshot memory usage because it avoid retaining recent blocks. We already have a memo on block processing changing - why we do need this one?

packages/studio-base/src/components/MessagePipeline/index.tsx
```diff
@@ -84,13 +76,13 @@ export function MessagePipelineProvider({
...
   const promisesToWaitForRef = useRef<FramePromise[]>([]);
-  const [store] = useState(() =>
-    createMessagePipelineStore({ promisesToWaitForRef, initialPlayer: player }),
-  );
-  useEffect(() => {
-    store.getState().dispatch({ type: "set-player", player });
-    player?.setPublishers(store.getState().allPublishers);
-  }, [player, store]);
+
+  // fixme - seems one item to note is that allPublishers will be cleared when player is cleared?
```
defunctzombie, line 80: From @jtbandes 

> maybe the relevant thing to point out is that we need to make sure a panel that has registered a publisher still works when you reset the player?

defunctzombie: We re-mount the entire panel stack when we change players - so panels would need to re-register anyway?

defunctzombie: https://github.com/foxglove/studio/pull/5473

defunctzombie: The bugfix was about fixing a race between advertise/publish and not about preserving publishers across player changes.

PR #6256, opened 2023-06-15
flatbuffers: set readDefaults=true

jtbandes:
**User-Facing Changes**
FlatBuffer messages will now appear with all default values populated, even if they were not serialized explicitly in the binary data.

**Description**
Studio generally needs default values (such as `quaternion.w=1`) to be present in order to render things properly (at least, in panels that have a fixed set of supported schemas). Since the published flatbuffer schemas [here](https://github.com/foxglove/schemas/tree/main/schemas/flatbuffer) use [default values](https://github.com/foxglove/schemas/blob/a4fe2a8af0d837e7d413be6a325e47b8f0b81a0b/schemas/flatbuffer/Quaternion.fbs#L16-L17), a writer currently has to use `builder.ForceDefaults(true);` in order to ensure their values are written into the buffer if they match the defaults. With this change, defaults will always be populated based on the schema when Studio deserializes an object, even if the values were not present in the buffer.

This matches (roughly) what we do for protobuf: https://github.com/foxglove/studio/blob/f23e993460db86f9fe6de3ee59046fbea3211086/packages/mcap-support/src/parseProtobufSchema.ts#L68-L71

### DIFF
--- a/packages/mcap-support/src/parseFlatbufferSchema.test.ts
+++ b/packages/mcap-support/src/parseFlatbufferSchema.test.ts
@@ -352,7 +352,14 @@ describe("parseFlatbufferSchema", () => {
     Type.addIndex(builder, 123);
     builder.finish(Type.endType(builder));
 
-    expect(deserialize(builder.asUint8Array())).toEqual({ base_type: 7, index: 123 });
+    expect(deserialize(builder.asUint8Array())).toEqual({
+      base_size: 4n,
+      base_type: 7,
+      element: 0n,
+      element_size: 0n,
+      fixed_length: 0n,
+      index: 123,
+    });
   });
   it("converts uint8 vectors to uint8arrays", () => {
     const builder = new Builder();
--- a/packages/mcap-support/src/parseFlatbufferSchema.ts
+++ b/packages/mcap-support/src/parseFlatbufferSchema.ts
@@ -191,6 +191,10 @@ export function parseFlatbufferSchema(
     }
   }
   const parser = new Parser(rawSchema);
+  // We set readDefaults=true to ensure that the reader receives default values for unset fields, or
+  // fields that were explicitly set but with ForceDefaults(false) on the writer side. This is
+  // necessary because `datatypes` does not include information about default values from the
+  // schema. See discussion: <https://github.com/foxglove/studio/pull/6256>
   const toObject = parser.toObjectLambda(typeIndex, /*readDefaults=*/ true);
   const deserialize = (buffer: ArrayBufferView) => {
     const byteBuffer = new ByteBuffer(
--- a/packages/mcap-support/src/parseFlatbufferSchema.ts
+++ b/packages/mcap-support/src/parseFlatbufferSchema.ts
@@ -191,7 +191,7 @@ export function parseFlatbufferSchema(
     }
   }
   const parser = new Parser(rawSchema);
-  const toObject = parser.toObjectLambda(typeIndex);
+  const toObject = parser.toObjectLambda(typeIndex, /*readDefaults=*/ true);
   const deserialize = (buffer: ArrayBufferView) => {
     const byteBuffer = new ByteBuffer(
       new Uint8Array(buffer.buffer, buffer.byteOffset, buffer.byteLength),


jameskuszmaul-brt:
My thoughts, for the record:

### Background

First, in case anyone is unclear as to what `ForceDefaults()` does:

When writing a scalar value `x`, it will or won't be serialized into the output based on:
| `ForceDefaults(_)` | `x` is set to the default | `x` is set to a non-default value | `x` is never set |
| - | - | - | - |
| `true` | serialized | serialized | not present |
| `false` | not present | serialized | not present |

When you then attempt to read `x` when deserializing using the standard flatbuffers upstream library, you will observe no difference between the two settings--if you attempt to read `x` when it is not present in the serialized bytes, you will get the default value; if `x` is present, you will get the value of `x` (which may be the default).

However, if you use https://github.com/jkuszmaul/flatbuffers_reflection and attempt to read values, then you can specify `readDefaults` in the various function calls, which results in the following return values depending on if `x` is present in the serialized bytes:
| `readDefaults` | `x` is serialized | `x` is not present |
| - | - | - |
| `true` | `x` as it appears on the wire | the default value of `x` |
| `false` | `x` as it appears on the wire | `null` |

In general you need to have `readDefaults != ForceDefaults(_)` in order to have consistent behavior across the serialization/deserialization layers.

The fork of flatbuffers at https://github.com/frc971/971-Robot-Code/tree/master/third_party/flatbuffers which we use also adds a `has_*` function to the C++ generated code that can be used to detect whether `x` is present in the serialized flatbuffer.

### Opinions

This change would be consistent with the default behavior for the flatbuffers library, and thus is probably more friendly to someone trying to get started working with flatbuffers & studio. If there are any established code-bases using flatbuffers with `ForceDefaults(false)` who want to adopt studio, then this would be highly desirable for them.

However, the code-bases I am involved with actually have a local patch to the flatbuffers library to make it so that `ForceDefaults(true)` is the default behavior. This has the result that we can actually encode meaning into whether the field is present, e.g.:
* A sensor reading may not have a meaningful "default" value, and having it be absent may indicate that the sensor is disconnected.
* You might have a message that contains a bunch of settings for some configuration. When that message is sent, you may only populate the fields that you want to change/override, and want to have "not populated" mean "ignore this setting," not "set this setting back to the default value in the flatbuffer schema".

As such, there is real meaning encoded into a value not being present, and it would be very frustrating if, when I go to examine a message in the Raw Messages panel, I could not distinguish between a field not being populated and it being explicitly set to the defaults. At the same time, for some panels I don't care--if I am sending a Color to be consumed by some visualization panel, I'm never going to want to populate Red and Green but not Blue.

In the longer term, it seems like there would be utility in having the `MessageDefinitionField` capture the defaults and then have some internal interface that allows panels to select their behavior appropriately. In the shorter-term, it would be awesome if there were some way to expose a global setting to users (I do not know where would be appropriate) to allow selecting this behavior (and probably extend it to protobufs as well).

I would actually be somewhat surprised if no one has complained about this when using protobufs. Maybe that's a consequence of people being heavily used to ROS 1 serialization formats. Or maybe it's just that even in codebases like ours, _most_ fields end up just always being populated, so this isn't typically at the forefront of people's minds.

defunctzombie (approved):
@jameskuszmaul-brt Thank you for the detailed write up. After reading the issue and the related threads I am in agreement with this statement:

```
This change would be consistent with the default behavior for the flatbuffers library
```
To me it is a defensible position that when Studio de-serializes the message it applies the defaults expected from the schema definition and I would be _surprised_ as a flatbuffer user with any other behavior _especially_ now having learned about the `ForceDefaults(false)` behavior which I'm guessing exists to save serialization space when defaults are used.

Rather than "abusing" forcedefaults (https://github.com/google/flatbuffers/issues/6471) it seems that flatbuffers is exploring having true optional types: https://github.com/google/flatbuffers/issues/6014

A good direction to trend with MessageDefinitions is to introduce the concept of optional types.

@jtbandes Please add a comment to the change in parseFlatbufferSchema maybe referencing this PR and notes and/or copying some part of the reasoning so we have that recorded in the code.

jameskuszmaul-brt:
Not sure how I missed the introduction of optional scalars (and I don't *think* that we just saw them and decided not to use them), although they do have slightly different semantics than we are using currently. Migrating our code to that would be irritating, but might make sense.

jameskuszmaul-brt:
Is there anything that would be considered a tolerable mitigation to merge into `main` here rather than having to maintain a patched studio locally? E.g.,
> In the shorter-term, it would be awesome if there were some way to expose a global setting to users (I do not know where would be appropriate) to allow selecting this behavior (and probably extend it to protobufs as well).

jameskuszmaul-brt:
@jtbandes how do I go about getting an answer to the above question?

defunctzombie:
@jameskuszmaul-brt We chatted about this some internally and have not come up with something we'd consider tolerable. From everything I've been able to read about this - the behavior of using the defaults is a sane behavior and using the _lack_ of a field being set as an important signal of business logic is what I'd consider a-typical. There's definitely a set of workflows to debug serialization issues that benefit from this level of message introspection but that's not the level studio operates at today.

jameskuszmaul-brt:
So the official answer is that I should fork this and maintain a local copy of studio with the flag flipped?

defunctzombie:
> So the official answer is that I should fork this and maintain a local copy of studio with the flag flipped?

To any extent its an official answer, the official answer is that in Studio we apply the schema-defined defaults when de-serializing the data and are not planning to deviate from that architecture. I do think adding support for real _optional_ fields (with no defaults) is a sensible future but that doesn't sound like what you have happening with your use of flatbuffers.

PR #6263, opened 2023-06-16
Make settings in storybook actually work for non-extension panels

jtbandes:
**User-Facing Changes**
None

**Description**
With `<PanelSetup includeSettings>`, the settings weren't actually usable for non-extension panels, because the saveConfig call wouldn't do anything (childId was undefined).

For extension panels this doesn't matter, because the panel is responsible for managing its own copy of the config and does not rely on being re-rendered automatically with new config.

### DIFF
--- a/packages/studio-base/src/components/Panel.tsx
+++ b/packages/studio-base/src/components/Panel.tsx
@@ -146,6 +146,9 @@ type ComponentConstructorType<P> = { displayName?: string } & (
   | { (props: P): React.ReactElement<unknown> | ReactNull }
 );
 
+/** Used in storybook when panels are renered outside of a <PanelLayout/> */
+const FALLBACK_PANEL_ID = "$unknown_id";
+
 // HOC that wraps panel in an error boundary and flex box.
 // Gives panel a `config` and `saveConfig`.
 //   export default Panel(MyPanelComponent)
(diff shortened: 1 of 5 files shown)


packages/studio-base/src/panels/Image/index.stories.tsx
```diff
@@ -48,7 +48,7 @@ export const TopicButNoDataSourceHovered: StoryObj = {
   parameters: { colorScheme: "dark" },
   play: async ({ canvasElement }) => {
     const canvas = within(canvasElement);
-    userEvent.hover(await canvas.findByTestId("panel-mouseenter-container"));
+    userEvent.hover(await canvas.findByTestId(/panel-mouseenter-container/));
```
2metres, line 51: Why is this not a string?

jtbandes: It’s no longer an exact match, because the $unknown_id is in there too

2metres (approved):

2metres:
Was getting ready to merge and noticed 2 things 1 was that the panel title in the localized Plot storybooks have an issue with the panel title. Also when trying to interact witht the plot panel the settings don't change the panel.

jtbandes:
I believe the settings not working is due to the use of `overrideConfig` (`<Plot overrideConfig={props.config} />`). I think to make this work nicely we will want to move away from overrideConfig and figure out a way to initialize the settings store with a custom config but not permanently override it.

packages/studio-base/src/components/Panel.tsx
```diff
@@ -146,6 +146,9 @@ type ComponentConstructorType<P> = { displayName?: string } & (
   | { (props: P): React.ReactElement<unknown> | ReactNull }
 );
 
+/** Used in storybook when panels are renered outside of a <PanelLayout/> */
```
defunctzombie, line 149: Why would we render panels outside of PanelLayout? Somehow I thought that was not allowed.

jtbandes: A panel is just a component at the end of the day. In storybooks we have provided the props it needs as well as real or mock context providers.

PR #4514, opened 2022-09-29
improve playback scrubber precision using pointer events

jtbandes:
**User-Facing Changes**
None

**Description**
Subtle improvement to precision of playback scrubber. Scrubber and tooltip remain aligned.

Simplified a bit of code in the Slider component. This component is only used for the playback controls and does not need to be so generic.

<table><tr><th>Before</th><th>After</th></tr><tr><td>

https://user-images.githubusercontent.com/14237/193122525-d9643019-56f5-4747-91ca-7070c5f9fc21.mov

</td><td>

https://user-images.githubusercontent.com/14237/193122541-4662fc90-3096-4527-8c25-9324ef23e75f.mov

</td></tr></table>

PR #6265, opened 2023-06-16
Fix moving mouse off playback bar while scrubbing

jtbandes:
**User-Facing Changes**

**Description**

- Allows scrubbing to continue even if the mouse moves outside the playback bar or window
- Tooltip remains visible the entire time

I made this work in https://github.com/foxglove/studio/pull/2912, then I broke it again in https://github.com/foxglove/studio/pull/4514. The pendulum swings back again.

Additionally I believe the behavior of showing the tooltip even when the mouse is outside the bar changed in https://github.com/foxglove/studio/pull/4392.

Before

https://github.com/foxglove/studio/assets/14237/620511de-f16d-446f-a538-11a80373e186

After

https://github.com/foxglove/studio/assets/14237/98b41e2a-4732-4e77-8362-87683d385625

### DIFF
--- a/packages/studio-base/src/components/PlaybackControls/Scrubber.tsx
+++ b/packages/studio-base/src/components/PlaybackControls/Scrubber.tsx
@@ -31,7 +31,7 @@ import { EventsOverlay } from "./EventsOverlay";
 import PlaybackBarHoverTicks from "./PlaybackBarHoverTicks";
 import { PlaybackControlsTooltipContent } from "./PlaybackControlsTooltipContent";
 import { ProgressPlot } from "./ProgressPlot";
-import Slider from "./Slider";
+import Slider, { HoverOverEvent } from "./Slider";
 
 const useStyles = makeStyles()((theme) => ({
   marker: {
@@ -79,9 +79,12 @@ export default function Scrubber(props: Props): JSX.Element {
 
   const setHoverValue = useSetHoverValue();
 
-  const [hoverInfo, setHoverInfo] = useState<
-    { stamp: Time; position: { clientX: number; clientY: number } } | undefined
-  >();
+  type HoverInfo = {
+    stamp: Time;
+    clientX: number;
+    clientY: number;
+  };
+  const [hoverInfo, setHoverInfo] = useState<HoverInfo | undefined>();
   const latestHoverInfo = useLatest(hoverInfo);
 
   const latestStartTime = useLatest(startTime);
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/components/PlaybackControls/Scrubber.tsx
```diff
@@ -80,7 +79,10 @@ export default function Scrubber(props: Props): JSX.Element {
 
   const setHoverValue = useSetHoverValue();
 
-  const [hoverStamp, setHoverStamp] = useState<Time | undefined>();
+  const [hoverInfo, setHoverInfo] = useState<
+    { stamp: Time; position: { clientX: number; clientY: number } } | undefined
```
defunctzombie, line 83: Tempted to say that this type has evolved beyond what we should have inline and should hoist to a `type` in the file for readability.

packages/studio-base/src/components/PlaybackControls/Scrubber.tsx
```diff
@@ -147,57 +150,60 @@ export default function Scrubber(props: Props): JSX.Element {
...
+  const isHovered = hoverInfo != undefined;
+  const popperProps: Partial<PopperProps> = useMemo(
+    () => ({
+      open: isHovered, // Keep the tooltip visible while dragging even when the mouse is outside the playback bar
+      popperRef,
+      modifiers: [
+        {
+          name: "computeStyles",
+          options: {
+            gpuAcceleration: false, // Fixes hairline seam on arrow in chrome.
```
defunctzombie, line 162: 😮

packages/studio-base/src/components/PlaybackControls/Slider.tsx
```diff
@@ -12,7 +12,11 @@ type Props = {
   fraction: number | undefined;
   disabled?: boolean;
   onChange: (value: number) => void;
-  onHoverOver?: (value: number) => void;
+  /**
+   * @param value Hovered `fraction` value
+   * @param position Current hovered position in client coordinates
+   */
+  onHoverOver?: (value: number, position: { clientX: number; clientY: number }) => void;
```
defunctzombie, line 19: Once this starts to grow beyond one parameter maybe worth having a `HoverOverEvent`?

defunctzombie (approved):

packages/studio-base/src/components/PlaybackControls/Scrubber.tsx
jtbandes: just moved

PR #6268, opened 2023-06-17
Fix user script memory leak due to Promise.race

achim-k:
**User-Facing Changes**
- Fix memory leak in user scripts

**Description**
There is a potential memory leak when using `Promise.race` with long-running promises. This is very well described here: https://github.com/nodejs/node/issues/17469#issuecomment-685216777

> In actuality, when you call `Promise.race` with a long-running promise, the resolved value of the returned promise gets retained for as long as each of its promises do not settle.

This is the case for user scripts: User script outputs are retained in memory as long as `terminateSignal` does not resolve. For live connections, the promise `terminateSignal` never resolves and every user script output is retained until the app finally runs out of memory (memory grows faster with larger user script outputs). 

~This patch fixes the memory leak by ensuring that none of the promises in `Promise.race` is long-running. A cleaner fix might be to use a replacement for `Promise.race` which does not suffer from the leaking behavior. See the link above for a possible implementation that uses a `WeakMap` under the hood.~

Edit: This PR now ships a non-leaking re-implementation of `Promise.race`

There are other occurrences of `Promise.race` in the code base where memory might leak, but this one is by far the most severe with respect to retained memory.

### DIFF
--- a/packages/studio-base/src/players/UserNodePlayer/index.ts
+++ b/packages/studio-base/src/players/UserNodePlayer/index.ts
@@ -449,6 +449,7 @@ export default class UserNodePlayer implements Player {
       // These are lazily created in the first message we process and cleared on
       // terminate.
       let terminateCondvar: undefined | Condvar;
+      let terminateSignal: undefined | Promise<void>;
 
       const registration = async (msgEvent: MessageEvent, globalVariables: GlobalVariables) => {
         // Register the node within a web worker to be executed.
@@ -523,6 +524,7 @@ export default class UserNodePlayer implements Player {
         // processing that is in-flight.
         if (!terminateCondvar) {
           terminateCondvar = new Condvar();
+          terminateSignal = terminateCondvar.wait();
         }
 
         // To send the message over RPC we invoke maybePlainObject which calls toJSON on the message
@@ -538,7 +540,7 @@ export default class UserNodePlayer implements Player {
             },
             globalVariables,
           }),
-          terminateCondvar.wait(),
+          terminateSignal,
         ]);
 
         if (!result) {
@@ -547,9 +549,6 @@ export default class UserNodePlayer implements Player {
             severity: "warn",
           });
           return;
-        } else {
-          // Let terminateCondvar.wait() above resolve to avoid memory leaking in Promise.race()
-          terminateCondvar.notifyOne();
         }
 
         const allDiagnostics = result.userNodeDiagnostics;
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/players/UserNodePlayer/index.ts
```diff
@@ -549,6 +547,9 @@ export default class UserNodePlayer implements Player {
             severity: "warn",
           });
           return;
+        } else {
+          // Let terminateCondvar.wait() above resolve to avoid memory leaking in Promise.race()
+          terminateCondvar.notifyOne();
```
defunctzombie, line 552: This will notify _any_ single waiter of the condvar which might not be the one you created above with the `.wait()` call since the condvar is shared across registrations.

Also, the `terminateCondvar` is for when we want to terminate all the running scripts (resets, etc).

defunctzombie:
This is tragic and makes me sour on `Promise.race`. I'm tempted to say we should ban use of `Promise.race` and replace it with our own implementation (or the library referenced in the issue). I'm certainly surprised that the unresolved promise holds the entire memory because once the race is complete I'd not imagine that it matters what happens to the unresolved promise.

Alternatively, we could write this code in a way that resolves both promises but it does feel to me like that doesn't fix the general issue of it not being obvious that you can't have long-running promises in a `Promise.race`. Maybe this is obvious to others?

packages/studio-base/src/players/UserNodePlayer/index.ts
achim-k: Oh, you are right, this will not work with multiple user scripts. Then it's probably better to use our own implementation of `Promise.race`

packages/studio-base/src/util/safePromiseRace.ts
```diff
@@ -0,0 +1,74 @@
...
+ * from this drawback. It is an adaption of Brian Kim's [1] version including the improvement by
+ * Dan Bornstein [2].
+ *
+ * [1]: https://github.com/nodejs/node/issues/17469#issuecomment-685216777
+ * [2]: https://github.com/danfuzz/lactoserv/pull/173
+ *
+ */
+export default async function safePromiseRace<T extends readonly PromiseLike<unknown>[]>(
+  contenders: T,
+): Promise<Awaited<T[number]>> {
```
achim-k, line 21: The original `Promise.race` signature below allows to pass in arbitrary arguments. My typescript skills were not strong enough to get this working, hence I had to restrict this promises only :/


```
race<T extends readonly unknown[] | []>(values: T): Promise<Awaited<T[number]>>;
```

defunctzombie:
Thanks for digging into this tragic story. I've created a follow-on PR we can use once we purge the `Promise.race` evil.

I will pickup this PR and move the new safe race to our `den` package (We don't put things in `util` anymore).

defunctzombie:
Not sure yet the impact this has for us since typically when you race promises you are not CPU bound - but the proposed implementation queues fewer promises than the built-in one:

image

achim-k:
Superseded by #6303

PR #6269, opened 2023-06-19
Fall back to existing datatypes if not specified in advertise options

achim-k:
**User-Facing Changes**
- Foxglove websocket: Fall back to existing datatypes when advertising a ROS topic

**Description**
When calling `context.advertise?.("topicName", "schemaName")` without passing the schema's datatypes as options, advertisement should not fail if the player can resolve datatypes by the given schema name.

See https://foxglove.slack.com/archives/C01T10L5QJX/p1686859619483039 for context


jtbandes (approved):

jtbandes:
Sad we have no tests for this stuff.

achim-k:
> Sad we have no tests for this stuff.

Should I have added some?

jtbandes:
It sounds like a larger project, but if you have some ideas for how to do it, it would be nice to write them down somewhere :)

PR #6273, opened 2023-06-20
Align app menu and native menu

foxymiles:
**User-Facing Changes**
Bring desktop native menu and app menu closer to alignment.

**Description**
Bring desktop native menu and app menu closer to alignment.

This doesn't try to localize the menu items and omits the view menu items for toggling sidebar state. Both of those things could be accomplished with significantly more effort.

https://github.com/foxglove/studio/assets/93935560/cc8b8f24-d0a3-441c-8fe7-df5c998ceb02

### DIFF
--- a/packages/studio-desktop/src/main/StudioWindow.ts
+++ b/packages/studio-desktop/src/main/StudioWindow.ts
@@ -3,23 +3,23 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
 import {
-  app,
-  dialog,
   BrowserWindow,
   BrowserWindowConstructorOptions,
   Menu,
-  MenuItemConstructorOptions,
-  shell,
   MenuItem,
-  systemPreferences,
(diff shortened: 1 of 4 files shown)


packages/studio-desktop/src/main/StudioWindow.ts
```diff
@@ -299,16 +242,6 @@ function buildMenu(browserWindow: BrowserWindow): Menu {
...
-      ...(isMac
-        ? []
-        : [
-            { type: "separator" } as const,
-            {
-              label: "Settings…",
-              accelerator: "CommandOrControl+,",
-              click: () => browserWindow.webContents.send("open-app-settings"),
-            } as const,
-          ]),
```
2metres, line 311: Does this mean the `Cmd + ,` no longer opens the settings?

2metres (approved):

defunctzombie (approved):

PR #6276, opened 2023-06-21
Add padding to input when clear indicator is present

2metres:
**User-Facing Changes**
Prevents clear indicator from overlapping text

Screen Shot 2023-06-22 at 8 49 01 am

**Description**
Resolves FG-3940


defunctzombie (approved):
Any stories we can have for this or maybe there were stories already that caught this?

2metres:
> Any stories we can have for this or maybe there were stories already that caught this?

https://603ec8bf7908b500231841e2-hfquyihirm.chromatic.com/?path=/story/components-autocomplete--with-a-long-truncated-path-and-auto-size

Issue #6062, opened 2023-05-18
Constant error when connected to wifi without internet


bertaveira:
**Description**
We connect to our robots wifi where we use Foxglove Bridge. The problem is that this wifi does not have an internet connection. Because of this I get a constant error like this:
<img width="909" alt="image" src="https://github.com/foxglove/studio/assets/9644729/2e200ba0-df0f-4ee3-936c-866fe6b732b9">

It becomes quite disturbing to have to close it every few seconds

- Version: 1.53.1
- Platform: MacOS
- Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native): Foxglove Rosbridge

foxymiles:
Handing off to @2metres for UX inspiration here.

2metres:
I think we should persistant errors into the Problems sidebar instead of pushing to a toast

foxymiles:
These will show up as items in the problems sidebar instead of as a snackbar in the next studio release.

PR #6279, opened 2023-06-22
Offline avatar handling

2metres:
**User-Facing Changes**
Show default avatar when user is offline instead of missing image

**Description**
Detect when the user is offline and set the connection status in the WorkspaceContext. 

Resolves FG-3563


2metres:
I'm not 100% certain this is the right approach. But taking a stab anyways

foxymiles:
I think this is worth exploring but it won't help with issues like https://github.com/foxglove/studio/issues/6062 where the user has an active wifi connection but can't reach the internet. For that we'd need some kind of liveness check that actually hits an internet endpoint.

jtbandes:
> User-Facing Changes
> Detect when the user is offline and set the connection status in the WorkspaceContext.

The user-facing changes are meant to be used as an entry in the product release notes. A user who reads the release notes will have no idea what this means.

defunctzombie:
https://medium.com/@abhishekmicosoft/handling-img-fallback-307653b2f30

We can use a pure html solution and inline a base64 image resource as the fallback. This lets the browser handle what happens if the image can't be loaded.

```
<object data="https://www.gravatar.com/avatar/00000000000000000000000000000000" type="image/jpeg">
  <img src="https://www.gravatar.com/avatar/205e460b479e2e5b48aec07710c08d50" />
</object>
```

2metres:
Closed by #6308 6308

PR #6281, opened 2023-06-22
[Map Panel] Display covariance as ellipse

bryfox:
**User-Facing Changes**

Display NavSatFix covariances as ellipses on the Map Panel

**Description**

Currently, the map panel displays the accuracy of `NavSatFix` messages as a circle around the point, based on the max variance seen. There's enough information in the covariance matrix for us to display these as ellipses. We handle two cases of this, depending on the [covariance type](https://github.com/foxglove/studio/blob/50cf7b623ba307c5e5f5ccc77972ce7c5aeaecc4/packages/studio-base/src/panels/Map/types.ts#L17-L22).

This adds [leaflet-ellipse](https://www.npmjs.com/package/leaflet-ellipse) (Apache 2.0) as a dependency to render the ellipses.

The `getAccuracy` function has been moved to a separate file for testing, and extended in the following ways:

1. For the *diagonal known* type, we return both covariances as the radii for the ellipse, rather than the max of the two
2. For the *known* type, we use the two most significant eigenvalues from the matrix, and the eigenvector corresponding to the most significant value to determine the orientation.

FG-3873

### DIFF
--- a/packages/studio-base/src/panels/Map/getAccuracy.ts
+++ b/packages/studio-base/src/panels/Map/getAccuracy.ts
@@ -1,0 +1,92 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { MathNumericType, atan2, eigs, isNumber, unit } from "mathjs";
+
+import {
+  NavSatFixMsg,
+  NavSatFixPositionCovarianceType,
+} from "@foxglove/studio-base/panels/Map/types";
+
+type NumericPair = [MathNumericType, MathNumericType];
+
+/**
+ * Calculates the accuracy of a NavSatFix message, based on its type, and returns
+ * information suitable for display as a leaflet Ellipse.
+ *
+ * @param msg NavSatFix
+ * @returns radii and tilt (degrees from W)
+ */
+export function getAccuracy(
+  msg: NavSatFixMsg,
+): { radii: [number, number]; tilt: number } | undefined {
+  const covariance = msg.position_covariance;
+  if (!covariance) {
+    return undefined;
+  }
+
+  switch (msg.position_covariance_type) {
+    case undefined:
+      return undefined;
+    case NavSatFixPositionCovarianceType.COVARIANCE_TYPE_UNKNOWN:
+      return undefined;
+    case NavSatFixPositionCovarianceType.COVARIANCE_TYPE_DIAGONAL_KNOWN: {
+      // Tilt is degrees from west
+      const eastVariance = covariance[0];
+      const northVariance = covariance[4];
+      return { radii: [Math.sqrt(eastVariance), Math.sqrt(northVariance)], tilt: 0 };
+    }
+    case NavSatFixPositionCovarianceType.COVARIANCE_TYPE_APPROXIMATED:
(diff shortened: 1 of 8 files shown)


bryfox:
I'm not sure why these are not showing up in UI Tests, but the changed snapshots are here:

- [Diagonal](https://www.chromatic.com/component?appId=603ec8bf7908b500231841e2&csfId=panels-map--single-point-diagonal-covariance&buildNumber=15329&k=64948c5bb98c12b19639943c-1200-snapshot-true&h=12&b=-9)
- [Full](https://www.chromatic.com/component?appId=603ec8bf7908b500231841e2&csfId=panels-map--single-point-full-covariance&buildNumber=15329&k=64948c5bb98c12b19639943d-1200-snapshot-true&h=13&b=-10)

defunctzombie:
> I'm not sure why these are not showing up in UI Tests, but the changed snapshots are here:

Possibly turbosnap - I've asked Chromatic about this and referenced this PR

defunctzombie:
@bryfox the full covariance is not a circle in the updated story - is that intentional?

bryfox:
> the full covariance is not a circle in the updated story - is that intentional?

Yes; 'full' refers to the `COVARIANCE_TYPE_KNOWN` type, which provides a full covariance matrix. (This is mentioned in the ros docs, but there's a fuller explanation of these here: https://answers.ros.org/question/10310/calculate-navsatfix-covariance/?answer=15197#post-id-15197.)

Depending on the values, it's possible for this to be circular, though we'll still render using the ellipse type. I can add additional map points to this story if that would help.

bryfox:
I updated the stories to use a higher-contrast marker color; the diff now shows up in UI Tests.

The issue was threshold (https://www.chromatic.com/docs/threshold); a value of 0.05 would allow the previous colors to trigger a change if we want to keep those.

defunctzombie (approved):

jtbandes:
https://github.com/foxglove/studio/pull/6292

packages/studio-base/src/panels/Map/getAccuracy.ts
```diff
@@ -0,0 +1,92 @@
...
+          !isNumber(eigenvector[0]) ||
+          !isNumber(eigenvector[1]) ||
+          !isNumber(eigenvalues[0]) ||
+          !isNumber(eigenvalues[1])
+        ) {
+          return undefined;
+        }
+
+        // Ellipse `tilt` is defined as number of degrees from the negative x axis
+        const theta = unit(atan2(eigenvector[1], eigenvector[0]), "rad").toNumber("deg");
```
jtbandes, line 73: Probably would be a performance improvement to use `* 180 / Math.PI` instead of the `unit` stuff

packages/studio-base/src/panels/Map/getAccuracy.ts
```diff
@@ -0,0 +1,92 @@
...
+        if (isNaN(tilt) || isNaN(primaryRadius) || isNaN(secondaryRadius)) {
+          throw new Error("Unable to calculate accuracy");
+        }
+
+        return {
+          radii: [primaryRadius, secondaryRadius],
+          tilt,
+        };
+      } catch (err) {
+        return undefined;
```
jtbandes, line 88: if the error is being discarded, it'd be probably faster to just `return undefined` instead of `throw`ing in the first place

jtbandes (commented):
👍

PR #6282, opened 2023-06-22
Remove option to revert to old UI

defunctzombie:
**User-Facing Changes**
The "revert to old ui" option is removed from the user menu.

**Description**
Remove the user-menu option to revert to the old UI. Users can still revert using the experimental setting but we want to make it less obvious to revert now that we have made the new UI the default.


defunctzombie:
If a user had reverted to the old UI this change would not force them onto the new UI because it does not alter their local config from localstorage. We need to have a way of forcing that setting so we can separate go through delete code that runs for the old UI as a separate set of changes.

foxymiles (approved):

PR #6284, opened 2023-06-23
Recreate a new MessagePipeline store when the player changes

defunctzombie:
**User-Facing Changes**
None

**Description**
Rather than attempting a manual cleanup of the pipeline store, create a new store when the player changes. This avoids situations where downstream (or even the store itself) is holding references through player changes. We want to treat player changes like a "reset the world" action.

I stumbled upon this while doing some memory/heap analysis during player changes. I wouldn't say this change is _strictly_ necessary but I did find it easier to reason about a clean "reset" more than a reset that preserved some bits of old state.

### DIFF
--- a/packages/studio-base/src/components/MessagePipeline/index.tsx
+++ b/packages/studio-base/src/components/MessagePipeline/index.tsx
@@ -3,15 +3,7 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
 import { debounce } from "lodash";
-import {
-  createContext,
-  useCallback,
-  useEffect,
-  useLayoutEffect,
-  useMemo,
-  useRef,
-  useState,
-} from "react";
+import { createContext, useCallback, useEffect, useLayoutEffect, useMemo, useRef } from "react";
 import { StoreApi, useStore } from "zustand";
 
 import { useGuaranteedContext } from "@foxglove/hooks";
@@ -84,13 +76,10 @@ export function MessagePipelineProvider({
   globalVariables,
 }: ProviderProps): React.ReactElement {
   const promisesToWaitForRef = useRef<FramePromise[]>([]);
-  const [store] = useState(() =>
-    createMessagePipelineStore({ promisesToWaitForRef, initialPlayer: player }),
-  );
-  useEffect(() => {
-    store.getState().dispatch({ type: "set-player", player });
-    player?.setPublishers(store.getState().allPublishers);
-  }, [player, store]);
+
+  const store = useMemo(() => {
+    return createMessagePipelineStore({ promisesToWaitForRef, initialPlayer: player });
+  }, [player]);
 
   const subscriptions = useStore(store, selectSubscriptions);
 
--- a/packages/studio-base/src/components/MessagePipeline/index.tsx
+++ b/packages/studio-base/src/components/MessagePipeline/index.tsx
@@ -77,6 +77,13 @@ export function MessagePipelineProvider({
 }: ProviderProps): React.ReactElement {
   const promisesToWaitForRef = useRef<FramePromise[]>([]);
 
+  // We make a new store when the player changes. This throws away any state from the previous store
+  // and re-creates the pipeline functions and references. We make a new store to avoid holding onto
+  // any state from the previous store.
+  //
+  // Note: This throws away any publishers, subscribers, etc that panels may have registered. We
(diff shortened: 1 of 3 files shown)


defunctzombie:
@jtbandes Here's the set of tests that currently fail with this change. These tests relate to changes added here: https://github.com/foxglove/studio/pull/5473 and it makes sense that my proposed code change would cause them to fail. 

What I think the testing misses is that we reset all the panels when changing players so it is not reasonable that the message pipeline does not preserve this kind of state since the new player may be complete different from the old player or may not even support publishing.

I agree that this behavior might not be obvious from the POV of the message pipeline and am open to ideas on how to make it more obvious - even if that idea is a code comment.

```
 ● MessagePipelineProvider/useMessagePipeline › informs new player of existing prior advertisements

    expect(received).toEqual(expected) // deep equality

    Expected: [{"schemaName": "test", "topic": "/studio/test"}]
    Received: undefined

      522 |     setPlayer(player2);
      523 |     rerender();
    > 524 |     expect(player2.publishers).toEqual<typeof player2.publishers>([
          |                                ^
      525 |       { topic: "/studio/test", schemaName: "test" },
      526 |     ]);
      527 |   });

      at Object.toEqual (src/components/MessagePipeline/index.test.tsx:524:32)

  ● MessagePipelineProvider/useMessagePipeline › when changing the player › closes old player when new player is supplied and stops old player message flow

    expect(received).toBe(expected) // Object.is equality

    Expected: 4
    Received: 3

      636 |       rerender();
      637 |       expect(player.close).toHaveBeenCalledTimes(1); // eslint-disable-line jest/no-standalone-expect
    > 638 |       expect(all.length).toBe(4); // eslint-disable-line jest/no-standalone-expect
          |                          ^
      639 |     });
      640 |
      641 |     it("closes old player when new player is supplied and stops old player message flow", async () => {

      at Object.toBe (src/components/MessagePipeline/index.test.tsx:638:26)

  ● MessagePipelineProvider/useMessagePipeline › when changing the player › does not think the old player is the new player if it emits first

    expect(received).toBe(expected) // Object.is equality

    Expected: 4
    Received: 3

      636 |       rerender();
      637 |       expect(player.close).toHaveBeenCalledTimes(1); // eslint-disable-line jest/no-standalone-expect
    > 638 |       expect(all.length).toBe(4); // eslint-disable-line jest/no-standalone-expect
          |                          ^
      639 |     });
      640 |
      641 |     it("closes old player when new player is supplied and stops old player message flow", async () => {

      at Object.toBe (src/components/MessagePipeline/index.test.tsx:638:26)

  ● MessagePipelineProvider/useMessagePipeline › transfers subscriptions and publishers between players

    expect(received).toEqual(expected) // deep equality

    - Expected  - 8
    + Received  + 1

    - Array [
    -   Object {
    -     "topic": "/studio/test",
    -   },
    -   Object {
    -     "topic": "/studio/test2",
    -   },
    - ]
    + Array []

      698 |     rerender();
      699 |     await act(async () => await delay(1));
    > 700 |     expect(player2.subscriptions).toEqual([{ topic: "/studio/test" }, { topic: "/studio/test2" }]);
          |                                   ^
      701 |     expect(player2.publishers).toEqual<typeof player2.publishers>([
      702 |       { topic: "/studio/test", schemaName: "test" },
      703 |     ]);

      at Object.toEqual (src/components/MessagePipeline/index.test.tsx:700:35)
```

defunctzombie:
I've removed the above tests because the behavior they test is not an important property of the MessagePipeline given that we remount all the panels when players change. We don't need to remember advertisements as panels are responsible for calling advertise again.

packages/studio-base/src/components/MessagePipeline/index.tsx
```diff
@@ -84,13 +76,10 @@ export function MessagePipelineProvider({
...
   const promisesToWaitForRef = useRef<FramePromise[]>([]);
-  const [store] = useState(() =>
-    createMessagePipelineStore({ promisesToWaitForRef, initialPlayer: player }),
-  );
-  useEffect(() => {
-    store.getState().dispatch({ type: "set-player", player });
-    player?.setPublishers(store.getState().allPublishers);
-  }, [player, store]);
+
+  const store = useMemo(() => {
```
jtbandes, line 80: This seems like a good place to put a comment about why it is okay to throw away existing subscriptions/advertisements etc.

jtbandes (approved):

PR #6291, opened 2023-06-26
Optimize plot data accumulation

foxymiles:
**User-Facing Changes**
Improve plot panel performance on large datasets.

**Description**
Rework the plot panel data processing to use accumulating buffers and cursors instead of regenerating all datasets every time new messages are available.

Performance before this patch on a large dataset:

https://github.com/foxglove/studio/assets/93935560/8ca2265d-6246-4d87-b959-5a87c0b1e342

After:

https://github.com/foxglove/studio/assets/93935560/24e9b92e-1e89-4f54-b3fb-a98ebdb72e42

For live data I can see about a 2x improvement on frame rate on a layout with multiple plots.

Before:
Screenshot 2023-06-29 at 6 26 21 AM

After:
Screenshot 2023-06-29 at 6 29 26 AM

Note this reverts the enhancement in https://github.com/foxglove/studio/pull/5694. We will have to reconsider how to achieve that without undoing the performance gains in this PR.

This PR also doesn't try to preserve the existing logic that inserts discontinuities when there are no messages for a particular path in a block, since the unit of a `block` is somewhat arbitrary and the absence of messages in a particular block doesn't necessarily imply that the series should contain gaps in the chart.


defunctzombie:
> Note this reverts the enhancement in #5694. We will have to reconsider how to achieve that without undoing the performance gains in this PR.

Thoughts on how we would bring this back? It is a frustrating experience to hit the cache limit and not be able to see data at your current playback time.

> This PR also doesn't try to preserve the existing logic that inserts discontinuities when there are no messages for a particular path in a block, since the unit of a `block` is somewhat arbitrary and the absence of messages in a particular block doesn't necessarily imply that the series should contain gaps in the chart.

👍

packages/studio-base/src/PanelAPI/useSubscribeToTopicsForBlocks.ts
```diff
@@ -0,0 +1,39 @@
...
+import { SubscribePayload } from "@foxglove/studio-base/players/types";
+
+/**
+ * Maintains subscriptions to topics required for blocks.
+ */
+export function useSubscribeToTopicsForBlocks(topics: readonly string[]): void {
+  const [id] = useState(() => uuidv4());
+
+  const setSubscriptions = useMessagePipeline(
+    useCallback(
```
defunctzombie, line 21: I'd like us to avoid using callbacks with no dependencies (since they hold onto closures which themselves tend to hold on to more than they should in javascript).

I know you are moving code that was already there - so if you want to make a linear ticket to follow-up on this cleanup thats 👍 too or if you want to roll it into this change since this is already part of the diff I think we'd be able to follow along.

packages/studio-base/src/components/TimeBasedChart/downsample.ts
```diff
@@ -85,8 +85,8 @@ export function downsampleTimeseries(
       continue;
     }
 
-    const x = Math.round(datum.x * pixelPerXValue);
-    const y = Math.round(datum.y * pixelPerYValue);
+    const x = Math.trunc(datum.x * pixelPerXValue);
```
defunctzombie, line 88: Why this change? More a rhetorical question and a call out that a comment here seems appropriate so we educate our future selves when reading this code.

packages/studio-base/src/components/TimeBasedChart/index.tsx
```diff
@@ -204,9 +207,14 @@ export default function TimeBasedChart(props: Props): JSX.Element {
   // some callbacks don't need to re-create when the current scales change, so we keep a ref
   const currentScalesRef = useRef<RpcScales | undefined>(undefined);
 
-  // calculates the minX/maxX for all our datasets
-  // we do this on the unfiltered datasets because we need the bounds to properly filter adjacent points
+  // Calculates the minX/maxX for all our datasets. We do this on the unfiltered datasets
+  // because we need the bounds to properly filter adjacent points. Defers to precomputed
+  // dataBounds, if available.
   const datasetBounds = useMemo(() => {
+    if (dataBounds) {
```
defunctzombie, line 214: This looks suspect to me - I know the comment says it defers to precomputed bounds if available but since we control the callers of this code why wouldn't it be available?

packages/studio-base/src/types/Bounds.ts
```diff
@@ -0,0 +1,33 @@
...
+ */
+export function makeInitialBounds(): Bounds {
+  return {
+    x: { min: Number.MAX_SAFE_INTEGER, max: Number.MIN_SAFE_INTEGER },
+    y: { min: Number.MAX_SAFE_INTEGER, max: Number.MIN_SAFE_INTEGER },
+  };
+}
+
+/**
+ * Merges two bounds into the smallest area that encompasses them both.
```
defunctzombie, line 26: To me _merge_ implies combining (or in set terms - union). I think the operation described here is _intersection_.

packages/studio-base/src/types/Bounds.ts
```diff
@@ -0,0 +1,33 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { Immutable } from "@foxglove/studio";
+
+/**
+ * Describes a bounded rectangle in 2d space.
```
defunctzombie, line 8: I'm not feeling that this comment is on-point but its also not strictly wrong. In 2d graphics (which the 2d space) comment makes me think of specific coordinate values typically as points. And the type here doesn't seem quite that.

Maybe something like "Bounds represents the min and max values (inclusive?) for a 2d space (x and y)." Could even call the type InclusiveBounds (or Exclusive) depending on what you have.

note we already have a range: https://github.com/foxglove/studio/blob/main/packages/studio-base/src/util/ranges.ts as a thing which is not the same but I did want to surface this fyi

packages/studio-base/src/test/mocks/mockMessage.ts
```diff
@@ -0,0 +1,16 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { MessageEvent } from "@foxglove/studio";
+
+export function mockMessage<T>(message: T, fields?: Partial<MessageEvent<T>>): MessageEvent<T> {
```
defunctzombie, line 7: Add a comment if you hope for more use across more code.

packages/studio-base/src/types/Bounds.ts
```diff
@@ -0,0 +1,33 @@
...
+ */
+export type Bounds = {
+  x: { min: number; max: number };
+  y: { min: number; max: number };
+};
+
+/**
+ * Creates initial bounds with values set to extremes to simplify merging bounds.
```
defunctzombie, line 16: Unclear to me how it simplifies. Describe what the "extremes" are. Possibly a better name for this function like `makeInvertedBounds`? I dunno - but when I think of the output of this function (which is min > max) I think I've received some invalid initial bounds since I'd expect bounds to have min < max.

fwiw - I roughly understand why you'd make bounds like this and am suggesting there's a better naming to convey intent

packages/studio-base/src/panels/Plot/datasets.tsx
```diff
@@ -145,70 +144,36 @@ function getDatasetsFromMessagePlotPath({
...
+      if (maybeMathFn) {
+        rangeData.push(applyToDatum(datum, maybeMathFn));
       } else {
-        // If we have a scatter plot, we can't take the derivative, so instead show nothing
-        rangeData = [];
+        rangeData.push(datum);
       }
     }
 
-    // Messages are provided in receive time order but header stamps might be out of order
```
defunctzombie, line 188: This sort behavior seems lost in translation. I'd imagine this also fails a plot story or test?

packages/studio-base/src/panels/Plot/datasets.tsx
```diff
@@ -230,17 +195,20 @@ function getDatasetsFromMessagePlotPath({
...
   xAxisPath?: BasePlotPath;
   invertedTheme?: boolean;
-};
+}>;
 
-type ReturnVal = {
-  datasets: DataSet[];
+type DatasetWithPath = { path: PlotPath; dataset: DataSet };
+
+export type DataSets = {
```
defunctzombie, line 209: Maybe `GetDatasetsResult`?

I don't know that I'd think of what you've put in this type as "datasets". I don't feel strongly about this.

packages/studio-base/src/panels/Plot/datasets.tsx
```diff
@@ -273,13 +243,46 @@ export function getDatasets({
       if (res.hasMismatchedData) {
         pathsWithMismatchedDataLengths.push(path.value);
       }
-      return res.dataset;
+      for (const datum of res.dataset.data) {
```
defunctzombie, line 246: Why bother computing this here rather than downstream where we were already computing it? Seems that you could drop this entire loop if you don't compute it here. I see that downstream you merge bounds but it seems that since you had to loop over the data here anyway you could loop over it elsewhere if needed? Or is the accumulated case letting you avoid looping over the data by tracking the bounds live as you accumulate?

Could you avoid this loop again by building the bounds when building the dataset?

packages/studio-base/src/panels/Plot/datasets.tsx
```diff
@@ -273,13 +243,46 @@ export function getDatasets({
...
 
   return {
+    bounds,
     datasets,
     pathsWithMismatchedDataLengths,
   };
 }
+
+/**
+ * Merges two datasets into a single dataset containing all points from both.
```
defunctzombie, line 268: Notably containing only the "dataset" properties from _a_ though.

packages/studio-base/src/panels/Plot/plotData.ts
```diff
@@ -2,197 +2,51 @@
...
-
-  const reduced = sorted.reduce((acc, item) => {
-    if (isEmpty(acc)) {
-      return item;
-    }
-    return mergeByPath(acc, item);
-  }, {} as PlotDataByPath);
-
-  return reduced;
+  return getMessagePathItems(decodeMessagePathsForMessagesByTopic, messages);
```
defunctzombie, line 51: A function that only calls another function should be 🔪 and we call the function it calls directly.

packages/studio-base/src/panels/Plot/plotData.ts
```diff
@@ -2,197 +2,51 @@
...
+        receiveTime: messageAndData.messageEvent.receiveTime,
+        headerStamp,
+      };
+    });
   });
   return ret;
 };
 
-const getMessagePathItemsForBlock = memoizeWeak(
+const getMessagePathItems = memoizeWeak(
```
defunctzombie, line 31: personally I'd drop memoizeWeak here and rely on the react dependency memoization that already gates the calls to these functions

packages/studio-base/src/panels/Plot/plotData.ts
```diff
@@ -2,197 +2,51 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
```
defunctzombie, line 2: What's the purpose of keeping this file around? Seems like it use to have some bulky logic but has been cut down and most logic migrated elsewhere?

packages/studio-base/src/panels/Plot/useFlattenedBlocksByTopic.ts
```diff
@@ -0,0 +1,106 @@
...
+ * @param topics to load from blocks
+ * @returns flattened per-topic arrays of messages
+ */
+export function useFlattenedBlocksByTopic(
+  topics: readonly string[],
+): Immutable<Record<string, MessageEvent[]>> {
+  const [state, setState] = useState<State>(makeInitialState);
+
+  const blocks = useMessagePipeline(
+    useCallback((ctx) => ctx.playerState.progress.messageCache?.blocks ?? EmptyBlocks, []),
```
defunctzombie, line 44: Lets avoid callbacks with no deps and favor static selector functions that are not greedy in their closures

packages/studio-base/src/panels/Plot/useFlattenedBlocksByTopic.ts
```diff
@@ -0,0 +1,106 @@
...
+ * Flattens incoming blocks into per-topic allFrames arrays.
+ *
+ * Internally this is implemented via cursor and an accumulating arrays instead of
+ * returning new arrays on each invocation to improve performance.
+ *
+ * @param blocks blocks containing messages
+ * @param topics to load from blocks
+ * @returns flattened per-topic arrays of messages
+ */
+export function useFlattenedBlocksByTopic(
```
defunctzombie, line 38: Feels to me like this could be better as a class or store that you instantiate and pass in blocks to update and get back the new messages. You could even run it via a useMemo rather than state. I think it might even read better than this current react-like state update approach which makes it seem more involved than it really is (if I am reading it correctly).

packages/studio-base/src/panels/Plot/useFlattenedBlocksByTopic.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+    expect(result.current).toEqual({
+      topic_a: [
+        expect.objectContaining({ topic: "topic_a" }),
+        expect.objectContaining({ topic: "topic_a" }),
+      ],
+      topic_b: [expect.objectContaining({ topic: "topic_b" })],
+    });
+  });
+});
```
defunctzombie, line 70: What types of variations should we expect this logic to handle? It would be useful for the test names to call out those variations.

Some that come to mind:
- empty arrays (for some or all topics)
- missing topic, followed by update to some topics
- blocks with topics, then gap, then more blocks

packages/studio-base/src/panels/Plot/usePlotPanelDatasets.ts
```diff
@@ -0,0 +1,397 @@
...
+ *
+ * Messages are provided in receive time order but header stamps might be out of order
+ * This would create zig-zag lines connecting the wrong points. Sorting the header stamp values (x)
+ * results in the datums being in the correct order for connected lines.
+ *
+ * An example is when messages at the same receive time have different header stamps. The receive
+ * time ordering is undefined (could be different for different data sources), but the header stamps
+ * still need sorting so the plot renders correctly.
+ */
+function sortDatasetsByHeaderStamp(datasets: DataSets["datasets"]): DataSets["datasets"] {
```
defunctzombie, line 102: This use to happen earlier in the process but seems to have moved. Pro/cons of this new location?

packages/studio-base/src/panels/Plot/usePlotPanelDatasets.ts
```diff
@@ -0,0 +1,397 @@
...
+          pathsWithMismatchedDataLengths: [],
+        };
+      }
+    },
+    [allPaths],
+  );
+
+  const latestAllFramesDatasets = useLatest(state.datasets);
+
+  const addMessages = useCallback(
```
defunctzombie, line 267: Related to your PR description call-out about regressing on showing data once the buffers are full - I somewhat expected this PR to combine the per-message update logic here and the blocks processing into one class (I know you hate that word so I'll say store - but its the same thing 🤣 ). You would feed new blocks into it or new messages (kinda sounds action like a reducer 🤣 ) and it would output datasets? It would be able to use its internal knowledge of cursors, window sizes, header vs receive stamps to produce the dataset outputs.

defunctzombie (commented):
Took my first pass through this. My takeaway is it makes sense. Architecturally, I'd say the code is roughly the same and the main optimization is as discussed - using cursors to avoid re-processing seen block data over-and-over for the hooks that return the datasets from blocks.

I left some notes for thought and do think theres an interesting idea to explore here around using one "instance" (to avoid your hated C word) of a processor to manage the blocks and live data inputs under one roof.

packages/studio-base/src/types/Bounds.ts
defunctzombie: Thought reading the code it does seem like you are forming a union rather than an intersection. The comment "smallest area that encompasses them both" is what confused me. I think it means to say "largest" area?

packages/studio-base/src/components/TimeBasedChart/downsample.ts
foxymiles: In my testing there is a huge difference in performance between `Math.round` and `Math.trunc`. I'll add a comment to explain this.

packages/studio-base/src/components/TimeBasedChart/index.tsx
foxymiles: StateTransition panel also uses this component and doesn't (yet) precompute bounds.

packages/studio-base/src/types/Bounds.ts
foxymiles: Yeah it's a union.

packages/studio-base/src/panels/Plot/datasets.tsx
foxymiles: It's applied to the entire dataset as a postprocessing step the same way the `@derivative modifier is now.

packages/studio-base/src/panels/Plot/datasets.tsx
foxymiles: We need to evaluate this point by point either here or later and it seems simpler to do it here and always include it in the dataset.

packages/studio-base/src/panels/Plot/datasets.tsx
foxymiles: Yeah this is a bit awkward but this PR is already big and complex. We should probably at some point separate the code that computes the various metaproperties of the dataset and the points.

packages/studio-base/src/panels/Plot/useFlattenedBlocksByTopic.ts
foxymiles: Memo is not an option here since we're accumulating state into a growing per topic buffer.

packages/studio-base/src/panels/Plot/useFlattenedBlocksByTopic.ts
foxymiles: I prefer that too but was just reorganizing some existing code here.

packages/studio-base/src/panels/Plot/usePlotPanelDatasets.ts
foxymiles: It has to happen here because we're accumulating to a growing buffer piecewise instead of doing a sort pass every time we rebuild everything. This means sorting by headerstamp is going to come at a performance cost although no worse than the current implementation on main.

packages/studio-base/src/panels/Plot/usePlotPanelDatasets.ts
foxymiles: It's possible to bring back that logic but since I've already spent quite a bit of time on this and it's already a big PR I thought we could consider deferring that to a followup. If you think it's worth spending the time I can bring that back to this approach.

foxymiles: `usePlotPanelDatasets` is the one stop shop for accumulating data you describe. IMO we can do this just fine with react primitives and don't need a store or any custom classes. I've broken out `useFlattenedBlocksByTopic` into a sub-hook just because it's a self contained piece of the process and easier to reason about this way.

If we want to do better on header stamps we will need to move that to the block level and handle the case where one series wants a topic in recieveTime order and other wants it in headerStamp order.

packages/studio-base/src/types/Bounds.ts
foxymiles: Yeah we already have this as a repeated literal type in a few places. I was just trying to clean it up with a centralized definition.

packages/studio-base/src/panels/Plot/plotData.ts
foxymiles: Yeah I don't think we need this anymore since we're not rebuilding the same data over and over now.

packages/studio-base/src/panels/Plot/plotData.ts
foxymiles: Yeah this is where the logic to handle seeking in low memory conditions will come back if we decide to do that.

packages/studio-base/src/panels/Plot/useFlattenedBlocksByTopic.ts
```diff
@@ -0,0 +1,110 @@
...
+          return false;
+        }
+      } else {
+        return false;
+      }
+    }
+    return true;
+  }, [state.messages]);
+
+  // Reset cursors and buffers if the first block has changed.
```
snosenzo, line 69: Why only if the first block changed? Is this attached to a blockloader implementation detail?

packages/studio-base/src/panels/Plot/useFlattenedBlocksByTopic.ts
snosenzo: Yeah I agree with roman, we should use a `useMemo` with a `ref` to a class for accumulation instead of a `useState`. That way you also wouldn't have to wait for another render after setting state and things would be encapsulated in a more readable manner

packages/studio-base/src/panels/Plot/useFlattenedBlocksByTopic.ts
foxymiles: This is just the most straightforward way I could figure out to tell that we need to start over accumulating state. Open to suggestions.

foxymiles:
> Yeah I agree with roman, we should use a useMemo with a ref to a class for accumulation instead of a useState. That way you also wouldn't have to wait for another render after setting state and things would be encapsulated in a more readable manner

Some state needs to change so that downstream clients know they need to rerender. IMO a ref/memo/class hack is just going to make this more convoluted and less idiomatic react code. React isn't going to do two renders here that's the point of doing a `setState` inline.

packages/studio-base/src/panels/Plot/useFlattenedBlocksByTopic.ts
snosenzo: store a reference to the last message under the cursor index and recalculate if it's different like I did in the 3D panel.

foxymiles: I guess I'd need to do that on a per-topic basis right?

packages/studio-base/src/panels/Plot/plotData.ts
defunctzombie: I don't feel strongly that you need to remove it here since we are going to do the follow-up work immediately - but I suspect things should look quite different when we bring those logic paths back.

packages/studio-base/src/panels/Plot/useFlattenedBlocksByTopic.test.tsx
defunctzombie: This was a callout that I think you should add more tests for the above cases.

packages/studio-base/src/panels/Plot/usePlotPanelDatasets.ts
defunctzombie: I find the react primitives make things more confusing here. I am also happy to see this handled in a follow-up so we get these current improvements shipped.

packages/studio-base/src/types/Bounds.ts
defunctzombie: This comment is still poorly written and does not help me understand what kinds of bounds I am getting. I would consider the return value here as "invalid bounds" because the values are inverted. Give this a think and see if you can come up with some thing more clear. At a minimum the comment can describe why/how you use it to "simplify merging bounds". Like do you feed these bounds into `unionBounds()` with another set of bounds?

defunctzombie (approved):
Please improve the comment in the Bounds.ts file - the rest seems doable in a follow-up PR. I do think theres value in exploring a less "react-primitives" approach in the follow-up that will merge two input streams.

PR #6292, opened 2023-06-27
Set chromatic diff threshold to 0

jtbandes:
**User-Facing Changes**
None

**Description**
We use Chromatic as a change detector. We'd like to catch every change.


defunctzombie (approved):
📿

PR #6304, opened 2023-06-27
Simplify getAccuracy implementation

bryfox:
**User-Facing Changes**

None.

**Description**

This addresses some follow-up performance feedback from https://github.com/foxglove/studio/pull/6281.


jtbandes (approved):
Wonder if the try/catch is still needed at all? I suppose it doesn’t hurt

PR #6305, opened 2023-06-27
Remove Promise.race from IterablePlayer

defunctzombie:
**User-Facing Changes**
None

**Description**
When paused, the IterablePlayer would be in an idle state. In this state, we still want to emit updates if the buffered ranges have changed (to display in the playback bar). Prior to this change, the logic to do this would use an infinite loop that would emit an update every loop iteration. The loop iteration speed was controlled with a `Promise.race` between a 1 second delay and an abort signal that we want to change state.

We have recently learned of the evils of `Promise.race` with respect to long-running or un-resolved promises and leaking the return values of all contenders. For this particular case, there would be a "leak" when a user paused playback and leaves playback paused for a long time. The "leak" would be small since neither the delay promise or the abort promise return values that take up significant space.

However, since our goal is to banish `Promise.race` from the codebase and install a lint rule banning its use, this logic can be re-written to avoid the `Promise.race`. This change avoids the race and uses an event notification when there are updates to the cached ranges to queue an emit. The idle state function then waits on an abort signal (signaling the desire to change state) and removes the handler before moving to the next state.

### DIFF
--- a/packages/studio-base/src/players/IterablePlayer/BufferedIterableSource.ts
+++ b/packages/studio-base/src/players/IterablePlayer/BufferedIterableSource.ts
@@ -2,6 +2,8 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
+import EventEmitter from "eventemitter3";
+
 import { Condvar } from "@foxglove/den/async";
 import { VecQueue } from "@foxglove/den/collection";
 import Log from "@foxglove/log";
@@ -27,6 +29,10 @@ type Options = {
   readAheadDuration?: Time;
 };
 
+interface EventTypes {
+  rangeChange: () => void;
+}
+
 /**
  * BufferedIterableSource proxies access to IIterableSource. It buffers the messageIterator by
  * reading ahead in the underlying source.
@@ -35,7 +41,7 @@ type Options = {
  * is the consumer and reads messages from cache while the startProducer method produces messages by
  * reading from the underlying source and populating the cache.
  */
-class BufferedIterableSource implements IIterableSource {
+class BufferedIterableSource extends EventEmitter<EventTypes> implements IIterableSource {
   #source: CachingIterableSource;
 
   #readDone = false;
@@ -63,8 +69,13 @@ class BufferedIterableSource implements IIterableSource {
   #readAheadDuration: Time;
 
   public constructor(source: IIterableSource, opt?: Options) {
+    super();
+
     this.#readAheadDuration = opt?.readAheadDuration ?? DEFAULT_READ_AHEAD_DURATION;
     this.#source = new CachingIterableSource(source);
+
+    // pass-through the range change event
+    this.#source.on("rangeChange", () => this.emit("rangeChange"));
   }
 
   public async initialize(): Promise<Initalization> {
@@ -168,6 +179,7 @@ class BufferedIterableSource implements IIterableSource {
 
   public async terminate(): Promise<void> {
     this.#cache.clear();
+    this.#source.removeAllListeners("rangeChange");
     await this.#source.terminate();
   }
 
--- a/packages/studio-base/src/players/IterablePlayer/BufferedIterableSource.ts
+++ b/packages/studio-base/src/players/IterablePlayer/BufferedIterableSource.ts
@@ -30,7 +30,8 @@ type Options = {
 };
 
 interface EventTypes {
-  rangeChange: () => void;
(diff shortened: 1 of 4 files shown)


packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
```diff
@@ -932,35 +933,46 @@ export class IterablePlayer implements Player {
...
-        resolve();
-      });
-    });
+    // While idle, the buffered source might still be loading so we check it for range changes
+    // and queue and emit if there are changes.
+    //
+    // 100 ms second updates from the buffered source.
+    // The loadedRanges() call is memoized so this is cheap and will not queue state emit if
+    // there are no changes to the ranges.
+    const interval = setInterval(() => {
```
jtbandes, line 956: Not ideal that we have to repeatedly call this method to check if something has changed. Can the buffered source tell us when the thing has changed? Feels like we are burning (maybe not a lot of) CPU cycles on this

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
```diff
@@ -932,35 +933,46 @@ export class IterablePlayer implements Player {
...
-      await Promise.race([delay(1000), aborted]);
-      if (this.#nextState) {
-        break;
-      }
+    this.#queueEmitState();
+    try {
+      await aborted;
+    } finally {
+      clearInterval(interval);
     }
```
jtbandes, line 976: `aborted` never throws since you only `resolve` it
```suggestion
    await aborted;
    clearInterval(interval);
```

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
```diff
@@ -932,35 +933,46 @@ export class IterablePlayer implements Player {
...
-      throw new Error("Invariant: some other abort controller exists");
-    }
+    // set the latest value of the loaded ranges for the next emit state
+    this.#progress = {
+      fullyLoadedFractionRanges: this.#bufferedSource.loadedRanges(),
+      messageCache: this.#progress.messageCache,
+    };
+
     const abort = (this.#abort = new AbortController());
+    const aborted = new Promise((resolve) => abort.signal.addEventListener("abort", resolve));
```
jtbandes, line 948: Is it possible for this to never resolve? What will happen in that case?

defunctzombie: I do not think it is possible but one can never really know in software. When the close() method is called on the player, this calls "setState" which invokes abort on any active `#abort` so for the player to "close" the abort is called. Whenever state changes the abort is called and state waits for the current state to stop before starting the new state (so a failure to resolve this will keep the player stuck in idle and be obvious that something is wrong).

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
defunctzombie: Changed to use event emitter

packages/studio-base/src/players/IterablePlayer/BufferedIterableSource.ts
```diff
@@ -27,6 +29,10 @@ type Options = {
   readAheadDuration?: Time;
 };
 
+interface EventTypes {
+  rangeChange: () => void;
```
jtbandes, line 33: doc comment on what this event means?

packages/studio-base/src/players/IterablePlayer/CachingIterableSource.ts
```diff
@@ -52,14 +53,18 @@ type Options = {
   maxTotalSize?: number;
 };
 
+interface EventTypes {
+  rangeChange: () => void;
```
jtbandes, line 57: doc comment on what this event means?

jtbandes (approved):
Seems relatively sane. You know this code much better than me though.

Maybe call the event `loadedRangesChange` if `loadedRanges` is what is changing?

packages/studio-base/src/players/IterablePlayer/CachingIterableSource.ts
defunctzombie: done

packages/studio-base/src/players/IterablePlayer/BufferedIterableSource.ts
defunctzombie: done

snosenzo:
As you have this now, saving user scripts or changing global variables does not update subscribed panels.

PR #6306, opened 2023-06-27
Mx/svg export

maxpolzin:
**User-Facing Changes**
Added ability to save plot as SVG graphic in the plot panel

**Description**
Added ability to save plot as SVG graphic in the plot panel.
This code is somewhat working and could serve as a starting point to implement this functionality properly.
There are no tests implemented. 
Please feel free to close this pull request, I simply wanted to make you aware of the required changes to be made.


defunctzombie:
Thanks for sharing some ideas for how you leverage the plot panel but I don't think we are ready to take on maintaining this feature right now. We offer CSV export and figure folks can use one of the many other tools to make svgs or other visualizations from that data.

PR #6307, opened 2023-06-27
Remove stacktrace from Problems list error display

defunctzombie:
**User-Facing Changes**
Remove un-actionable stack traces from problems list sidebar.

Before:
Screenshot 2023-06-27 at 2 00 27 PM

After:
Screenshot 2023-06-27 at 2 00 04 PM

**Description**
Rather than displaying a mangled and un-actionable stacktrace in the problems sidebar, show the message from the error. This is cleaner and typically contains actionable information.

### DIFF
--- a/packages/studio-base/src/components/ProblemsList.tsx
+++ b/packages/studio-base/src/components/ProblemsList.tsx
@@ -111,7 +111,7 @@ function ProblemDetails(props: { details: DetailsType; tip?: string }): JSX.Elem
 
   const content = useMemo(() => {
     if (details instanceof Error) {
-      return <div className={classes.detailsText}>{details.stack}</div>;
+      return <div className={classes.detailsText}>{details.message}</div>;
     } else if (details != undefined && details !== "") {
       return (
         <Typography style={{ whiteSpace: "pre-line" /* allow newlines in the details message */ }}>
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/components/ProblemsList.tsx
```diff
@@ -111,7 +111,7 @@ function ProblemDetails(props: { details: DetailsType; tip?: string }): JSX.Elem
 
   const content = useMemo(() => {
     if (details instanceof Error) {
```
jtbandes, line 113: Seems like we should remove the ability to pass an Error in if we are never going to display it?

defunctzombie: We use the `message` property from the Error. On the next line we use `details.message`

jtbandes: My point is we might as well have said that a player problem should always be a `string`. That would simplify a bunch of code and types, since we are not really using any features of Error beyond its `message`.

defunctzombie: So is the ask to change the _details_ of player problems to always be a string?

defunctzombie: I looked into this and I don't think its better to remove the `error` field from the `PlayerProblem` type. In many code-paths this error field is populated by a _catch_ exception. That logic makes sense and preserves the error for downstream to display as appropriate. For us today, appropriate means displaying the error message.

jtbandes (approved):

PR #6308, opened 2023-06-27
Handle offline user avatar using HTML Object element

2metres:
**User-Facing Changes**
Fallback avatar displayed when user is offline


packages/studio-base/src/components/AppBar/index.tsx
```diff
@@ -375,15 +375,13 @@ export function AppBar(props: AppBarProps): JSX.Element {
...
-                      <img
-                        src={currentUser.avatarImageUrl}
-                        referrerPolicy="same-origin"
-                        className={classes.userIconImage}
-                      />
-                    ) : (
+                    <object
+                      data={currentUser?.avatarImageUrl ?? ""}
+                      className={classes.userIconImage}
+                      type="image/png"
```
defunctzombie, line 381: Are we sure it is always a png image? Maybe this doesn't care if we say png and it ends up being a jpeg?

defunctzombie:
@2metres how'd you test this?

2metres:
> @2metres how did you test this?

Launched the app, signed in, then turned off my wifi, relaunched the app.

packages/studio-base/src/components/AppBar/index.tsx
2metres: Google always returns a png, not sure about other types of Auth though

packages/studio-base/src/components/AppBar/index.tsx
```diff
@@ -375,15 +375,13 @@ export function AppBar(props: AppBarProps): JSX.Element {
...
                 >
                   <Avatar className={classes.avatar} variant="rounded">
-                    {currentUser?.avatarImageUrl ? (
-                      <img
-                        src={currentUser.avatarImageUrl}
-                        referrerPolicy="same-origin"
-                        className={classes.userIconImage}
-                      />
-                    ) : (
+                    <object
```
defunctzombie, line 378: I think it would be better for this logic and the fallback to be put in a component. Possibly even replace the `<Avatar` component with one that does this.

defunctzombie (approved):
I think the avatar stuff should move into a component rather than continuing to pile into the index.tsx file. It will be easier to read, can have a nice comment about this a-typical object tag and how it works. It can also have stories showing it working. Please do that and then this LGTM.

packages/studio-base/src/components/AppBar/index.tsx
2metres: I think that splitting this out into a separate component is novel idea however the act of doing so adds a lot of duplicated code for a component that is only used once

2metres:
Turns out the MUI Avatar component handles invalid images for us. We just have to use it: https://github.com/mui/material-ui/blob/283b1aa8c8b9d6db0eab842da472d4b7cab51122/packages/mui-material/src/Avatar/Avatar.js#L162C3-L180C4

defunctzombie:
> Turns out the MUI Avatar component handles invalid images for us.

Magical

PR #6310, opened 2023-06-27
Add Browse Data Platform link to AppMenu

2metres:
**User-Facing Changes**
Add Browse Data Platform link to AppMenu

Screen Shot 2023-06-28 at 9 02 25 am

**Description**
When the user is signed in and using the hosted version of the app link to recordings page


packages/studio-base/src/components/AppBar/AppMenu.tsx
```diff
@@ -80,6 +96,12 @@ export function AppMenu(props: AppMenuProps): JSX.Element {
     [analytics, currentUserType],
   );
 
+  const onConsoleClick = useCallback(() => {
+    handleAnalytics("console");
+    window.open("https://console.foxglove.dev/recordings/list", "_blank");
```
2metres, line 101: @defunctzombie how do get the correct deployment URL?

defunctzombie: This UI element needs to be something we add to via our deployed builds and not in studio-base. Those have the correct URL.

packages/studio-base/src/components/AppBar/AppMenu.tsx
```diff
@@ -243,6 +265,16 @@ export function AppMenu(props: AppMenuProps): JSX.Element {
           } as Partial<PaperProps & { "data-tourid"?: string }>
         }
       >
+        {showConsoleLink && (
+          <MuiMenuItem
+            onPointerEnter={() => setNestedMenu(undefined)}
+            onClick={onConsoleClick}
+            id="app-menu-console"
+          >
+            Browse data
```
defunctzombie, line 274: This should be using the translations system.

packages/studio-base/src/components/AppBar/AppMenu.tsx
```diff
@@ -80,6 +96,12 @@ export function AppMenu(props: AppMenuProps): JSX.Element {
     [analytics, currentUserType],
   );
 
+  const onConsoleClick = useCallback(() => {
```
defunctzombie, line 99: Why are we handling this via javascript? Are we not able to use a Mui Link element?

2metres: So that we can include the analytics tracking event

packages/studio-base/src/components/AppBar/AppMenu.tsx
2metres: Not sure how this should work? We add something to the `AppContext` to allow this to be passed through? Does it have to be a UI element or can it be just a URL?

packages/studio-base/src/components/AppBar/AppMenu.tsx
defunctzombie: Maybe time for an `<TrackedLink>` component :D

We need more components in our lives.

defunctzombie (approved):

packages/studio-base/src/components/AppBar/AppMenu.tsx
2metres: @foxymiles can you assist on passing the URL from console?

packages/studio-base/src/components/AppBar/AppMenu.tsx
```diff
@@ -80,6 +96,12 @@ export function AppMenu(props: AppMenuProps): JSX.Element {
...
+    handleAnalytics("console");
+    window.open(`${process.env.FOXGLOVE_CONSOLE_URL}/recordings/list`, "_blank");
```
jtbandes, line 101: What will this do in the open source build? It seems to me that this code does not belong in the open source repo but should be injected in the private build.

packages/studio-base/src/context/AppContext.ts
```diff
@@ -4,8 +4,11 @@
 
 import { createContext, useContext } from "react";
 
+import { AppBarMenuItem } from "@foxglove/studio-base/components/AppBar/types";
+
 interface IAppContext {
   appBarLayoutButton?: JSX.Element;
+  appBarMenuItems?: readonly AppBarMenuItem[];
```
jtbandes, line 11: naming: wonder if these should be called "extra" app bar menu items? Since they are not replacing the existing menu items.

jtbandes (approved):

packages/studio-base/src/context/AppContext.ts
foxymiles: Yes could be. In the interest of moving forward on this after a long delay I'll go with this but we can revisit later.

PR #6320, opened 2023-06-28
Remove Promise.race from promiseTimeout

defunctzombie:
**User-Facing Changes**
None

**Description**
See code comment

### DIFF
--- a/packages/den/async/promiseTimeout.ts
+++ b/packages/den/async/promiseTimeout.ts
@@ -4,10 +4,7 @@
 
 /** Error for promise timeouts from `promiseTimeout()` */
 class PromiseTimeoutError extends Error {
-  public constructor(message: string) {
-    super(message);
-    this.name = "PromiseTimeoutError";
-  }
+  public override name = "PromiseTimeoutError";
 }
 
 /**
@@ -15,6 +12,9 @@ class PromiseTimeoutError extends Error {
  *
  * If the promise takes longer than the specified timeout duration (in milliseconds), it will be
  * rejected with a timeout error.
+ *
+ * Note: Make sure the input promise resolves, rejects, or otherwise go out of scope. A long-lived
+ * promise that never resolves holds onto its resolution callbacks.
  *
  * @param promise The promise to execute
  * @param ms The timeout duration in milliseconds.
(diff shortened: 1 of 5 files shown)


packages/studio-base/src/util/promiseTimeout.ts
```diff
@@ -16,18 +16,20 @@ async function promiseTimeout<T>(
...
-      if (id != undefined) {
-        clearTimeout(id);
-      }
-      return result;
-    }),
-    new Promise<T>((_resolve, reject) => {
-      id = setTimeout(() => reject(new Error(`Promise timed out after ${ms}ms: ${reason} `)), ms);
-    }),
-  ]);
+  // We avoid using Promise.race here since it is succeptible to memory leaks for unresolved promises
```
jtbandes, line 19:
```suggestion
  // We avoid using Promise.race here since it is susceptible to memory leaks for unresolved promises
```

jtbandes (approved):

defunctzombie:
> Maybe time to add tests for this function?

Only if I also move it to den

jtbandes:
I don't really understand the logic, either way it would be a promiseTimeout.test.ts file alongside the original file right?

packages/den/async/promiseTimeout.ts
```diff
@@ -0,0 +1,41 @@
...
+    super(message);
+    this.name = "PromiseTimeoutError";
+  }
+}
+
+/**
+ * Executes a promise with a timeout.
+ *
+ * If the promise takes longer than the specified timeout duration (in milliseconds), it will be
+ * rejected with a timeout error.
```
jtbandes, line 17: Maybe add some comment here about what you need to avoid doing if you want to avoid leaks?

jtbandes (approved):

packages/den/async/promiseTimeout.ts
```diff
@@ -0,0 +1,41 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+/** Error for promise timeouts from `promiseTimeout()` */
+class PromiseTimeoutError extends Error {
+  public constructor(message: string) {
+    super(message);
+    this.name = "PromiseTimeoutError";
+  }
```
jtbandes, line 10:
```suggestion
  name = "PromiseTimeoutError";
```
Does this achieve the same thing?

PR #6321, opened 2023-06-28
Remove Promise.race from installChromeExtensions

defunctzombie:
**User-Facing Changes**
None

**Description**
Instead of using `Promise.race` to timeout `installChromeExtensions` we use a single promise wrapper with a timeout. Our goal is to discourage the use of `Promise.race` which has memory leaks for unresolved promises.

### DIFF
--- a/packages/studio-desktop/src/main/installChromeExtensions.ts
+++ b/packages/studio-desktop/src/main/installChromeExtensions.ts
@@ -4,29 +4,26 @@
 
 import installExtension, { REACT_DEVELOPER_TOOLS } from "electron-devtools-installer";
 
+import { PromiseTimeoutError, promiseTimeout } from "@foxglove/den/async";
 import Logger from "@foxglove/log";
 
 const log = Logger.getLogger(__filename);
 
 export default async function installChromeExtensions(): Promise<void> {
   log.info("Installing Chrome extensions for development...");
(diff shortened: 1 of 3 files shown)


packages/studio-desktop/src/main/installChromeExtensions.ts
```diff
@@ -12,21 +12,21 @@ export default async function installChromeExtensions(): Promise<void> {
...
-        }
+  return await new Promise((resolve, reject) => {
+    const id = setTimeout(() => {
+      console.warn(
+        "Warning: extension installation may be stuck; try relaunching electron or deleting its extensions directory. Continuing for now.",
+      );
+      resolve();
+    }, 5000);
+
+    Promise.allSettled([installExtension(REACT_DEVELOPER_TOOLS)])
```
jtbandes, line 23: I think we can remove the allSettled thing – we got rid of the 2nd array item here which was the redux dev tools.

Could we import & use `promiseTimeout` here?

defunctzombie: > Could we import & use `promiseTimeout` here?

I'll double check but I don't think studio-desktop/main is setup to import from studio-base. Note that `promiseTimeout` rejects while this does not when the timeout is hit. I could try/catch suppress that if we really want to use `promiseTimeout`.

jtbandes (approved):

PR #6322, opened 2023-06-28
Support orgSlug in data source url params

wimagguc:
**User-Facing Changes**

When accessing files from data-platform, the logged in user may not be signed in with the correct org for that resource.  Previously an error was thrown in case, and with this PR we'll help the user understand the problem and help take the necessary next step:

belongs-elsewhere

Also works in desktop and with the sidebar DataSource:

desktop

switch

(Keeping it a draft until the proposed solution is confirmed)


wimagguc:
Closing this PR here, in favor of a to-be-defined second version of this.

PR #6324, opened 2023-06-28
Re-process last active data when user scripts change

defunctzombie:
**User-Facing Changes**
When saving a user script, the updated script is run on its inputs to produce a new output message.

**Description**
While editing a user-script it is convenient to see the output message (either in the raw message panel or other downstream panels). The only way scripts re-run is with a new player state update. This relies on upstream players to trigger an update and is fragile to upstream player behavior changes (i.e. https://github.com/foxglove/studio/pull/6305 which no longer sends periodic updates while paused).

This change internalizes the desire to re-run scripts on save by moving the logic to trigger a re-processing of the messages into the UserNodePlayer. When scripts are updated, the user node player will take the last player state and active data input and re-process it. This uses existing internal logic which cached the last messages on each topic and re-processes these through the updated scripts and emits this downstream.


foxymiles:
We recently added some code to emit new messages when the output topic list changes. Does this supersede that?

https://github.com/foxglove/studio/blob/e8b477e659a88763507abd929892915462ef47ef/packages/studio-base/src/players/UserNodePlayer/index.ts#L793-L813

packages/studio-base/src/players/UserNodePlayer/index.ts
```diff
@@ -397,7 +397,29 @@ export default class UserNodePlayer implements Player {
...
+      // If we have previous player state and are paused, then we re-emit the last active data so
+      // any panels that want our output topic get the updated message.
+      //
+      // Note: Until we learn otherwise, we assume that if a player is playing, it will emit new
+      // player state that will output new messages so we don't emit here while playing.
+      if (playerState && lastActive?.isPlaying === false) {
+        return {
+          ...playerState,
+          activeData: {
+            ...lastActive,
```
snosenzo, line 412: Currently we would re-emit all of the previous messages when we really only should re-emit computed messages. we should set `messages:[]` so that panels don't receive duplicates. Also we should update `_getMessages` to only return computed messages when recomputing. This should eliminate duplicate messages on original topics.

snosenzo (changes requested):

defunctzombie:
@snosenzo good insight. I've updated the PR to avoid re-emitting previous message data when re-computing. I've also removed the caching of the previous output since we target the invariant that we don't emit the _same_ `messages` into player state on emits. If we determine we need to guard against this we can do that better with a reference check or fix upstream.

I also updated tests which behave differently under the new changes so take a look at those.

defunctzombie:
> We recently added some code to emit new messages when the output topic list changes

Not entirely sure I follow. Are you thinking we can remove that code with these changes?

packages/studio-base/src/players/UserNodePlayer/index.ts
```diff
@@ -410,6 +381,12 @@ export default class UserNodePlayer implements Player {
           ...playerState,
           activeData: {
             ...lastActive,
+            // We want to avoid re-emitting upstream data source messages into panels to maintain
+            // the invariant that the player emits a data-source message into "currentFrame" only once.
+            //
+            // Using an empty messages array will make user node player only emit the script output
+            // messages as a result of the updated script.
+            messages: [],
```
snosenzo, line 389: nit: is it worth making this a stable constant array for referential equality and not having to instantiate a new array every time?

snosenzo (approved):
Looks good for the most part but, can you remove the empty emits in the node player tests that used to be there to emulate the idle emit?  for example:
image

packages/studio-base/src/players/UserNodePlayer/index.ts
defunctzombie: I doubt this makes an impact given we are making new objects to host this array.

PR #6333, opened 2023-06-29
Remove Top Nav feature flag and old UI

2metres:
**User-Facing Changes**
Remove old UI


defunctzombie:
@2metres for the user facing changes - wasn't the new UI the default already? The change here is that you can no longer go back to the old UI?

defunctzombie:
https://github.com/foxglove/studio/blob/main/packages/studio-base/src/context/AppContext.ts#L15 this is unused now and can be removed?

packages/studio-base/src/components/ExtensionDetails.tsx
```diff
@@ -99,14 +106,21 @@ export function ExtensionDetails({ extension, onClose, installed }: Props): Reac
...
   return (
-    <SidebarContent
-      title={extension.name}
-      leadingItems={[
-        <IconButton key="back-arrow" onClick={onClose} size="small" edge="start">
-          <ChevronLeftIcon />
-        </IconButton>,
-      ]}
-    >
+    <Stack fullHeight flex="auto" gap={1}>
```
defunctzombie, line 109: How's this change related to removing the old sidebar code?

packages/studio-base/src/components/Sidebars/Sidebars.tsx
```diff
@@ -0,0 +1,190 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
defunctzombie, line 1: Seems fine now that its here but this move adds noise to the PR making it slightly more annoying to review.

packages/studio-base/src/components/SignInFormModal.tsx
```diff
@@ -6,7 +6,7 @@ import CloseIcon from "@mui/icons-material/Close";
 import { Dialog, IconButton } from "@mui/material";
```
defunctzombie, line 6: Not for this PR but this form should be moved our of this repo.

defunctzombie (approved):

packages/studio-base/src/components/ExtensionDetails.tsx
2metres: I've removed the SidebarContent component and tweaked the design to feel more at home in the modal

PR #6335, opened 2023-06-29
Fix flaky chart stories with ready signal

foxymiles:
**User-Facing Changes**
None

**Description**
Fix flaky chart stories with ready signal.

### DIFF
--- a/packages/studio-base/src/components/TimeBasedChart/index.stories.tsx
+++ b/packages/studio-base/src/components/TimeBasedChart/index.stories.tsx
@@ -11,6 +11,7 @@
 //   found at http://www.apache.org/licenses/LICENSE-2.0
 //   You may not use this file except in compliance with the License.
 import { StoryObj } from "@storybook/react";
+import { isEqual } from "lodash";
 import cloneDeep from "lodash/cloneDeep";
 import { useState, useCallback, useRef, useEffect } from "react";
 import TestUtils from "react-dom/test-utils";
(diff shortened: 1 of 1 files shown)


defunctzombie (approved):

packages/studio-base/src/components/TimeBasedChart/index.stories.tsx
```diff
@@ -106,9 +107,11 @@ export const SimpleLight: StoryObj = { ...Simple, parameters: { colorScheme: "li
 
 export const CanZoomAndUpdate: StoryObj = {
   render: function Story() {
-    const [chartProps, setChartProps] = useState(cloneDeep(commonProps));
+    const [chartProps, setChartProps] = useState(structuredClone(commonProps));
```
jtbandes, line 110: Should we be replacing all cloneDeep with structuredClone throughout the code? Do they have any differences?

foxymiles: I think `structuredClone` is superior since it's native to the platform and should be preferred in new code but in practical terms there's probably not much benefit in replacing `cloneDeep` in working code unless we think it might be a performance bottleneck.

foxymiles:
Also unfortunately this still doesn't seem to have fixed the stories. I guess we'll need some way to signal back from the chart worker that it's done?

packages/studio-base/src/components/TimeBasedChart/index.stories.tsx
foxymiles: Although it seems that `structuredClone` might actually be slower at least in some cases:

https://www.measurethat.net/Benchmarks/Show/16221/0/lodash-clonedeep-vs-structuredclone

jtbandes:
That's what we do here: https://github.com/foxglove/studio/blob/main/packages/studio-base/src/components/Chart/index.stories.tsx#L134

PR #6336, opened 2023-06-29
[Image] Decode raw images in worker thread

### LINKED ISSUE #5398: 3D panel image rendering reduces app frame rate compared to Image panel
**Description**

The Image panel and 3D panel both use `@foxglove/den/image/decodings` to convert binary blobs into renderable image data, but the image panel runs these methods on a web worker while the 3D panel runs on the main thread. This can cause significant app slowdown when rendering an image format not natively supported by the browser.

jtbandes:
**User-Facing Changes**
- Improved performance of raw images in the Image and 3D panels by doing more processing in a background thread.

**Description**
Resolves #5398

Each ImageRenderable now owns a Worker which handles raw image decoding.

This introduces a copy of each raw image's data when sending it to the worker. (The old image panel also did this copy.) It also introduces a new allocation of a buffer for the _decoded_ image data. This new buffer is _transferred_ from the worker back to the main thread.

As a future improvement, if the memory usage becomes an issue, it would be possible (though more complicated) to implement a pool of Uint8ClampedArrays that can be used for the decoded data. When the main thread is finished using one it would transfer it back to the worker for reuse in future decodes.


jtbandes:
cc @defunctzombie for comlink stuff

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageMode.ts
```diff
@@ -580,71 +576,28 @@ export class ImageMode
     }
 
     renderable.name = topic;
-    renderable.setImage(image);
-    const isCompressedImage = "format" in image;
-
-    if (!isCompressedImage) {
-      // Raw Images don't need to be decoded asynchronously
+    renderable.userData.receiveTime = receiveTime;
```
snosenzo, line 579: why is it important to se the `receiveTime` here instead of having it be read in with the header info from the image message?

packages/studio-base/src/panels/ThreeDeeRender/renderables/Images/ImageRenderable.ts
```diff
@@ -138,41 +151,82 @@ export class ImageRenderable extends Renderable<ImageUserData> {
...
+        }
+        // prevent displaying an image older than the one currently displayed
+        if (this.#displayedImageSequenceNumber > seq) {
+          return;
+        }
+        this.#displayedImageSequenceNumber = seq;
+        this.#decodedImage = result;
+        this.#textureNeedsUpdate = true;
+        this.update();
+        this.renderer.queueAnimationFrame();
```
snosenzo, line 191: reason not to put this after `onDecoded?.(result)`?

snosenzo (approved):

snosenzo:
an stats on performance improvement?

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageMode.ts
jtbandes: what do you mean by "why is it important to se the receiveTime here instead of having it be read in with the header info from the image message?"? I'm not sure I see where that is being done. I only added this because I found we had one in `Images.ts` and it seemed like we should be consistent.

packages/studio-base/src/panels/ThreeDeeRender/renderables/Images/ImageRenderable.ts
jtbandes: no reason

jtbandes:
Perf measurements from 2 image panels attempting to play a 1920x1080x3 60Hz image:

image

The playback performance is still not amazing but the playback bar does update a bit more smoothly in the worker branch.

PR #6342, opened 2023-06-30
Remove incorrect layout selection language from PanelLayout

defunctzombie:
**User-Facing Changes**
None

**Description**
Update the language in PanelLayout to remove the incorrect layout selection language since there's no layout selection fallback in studio-base.

This change removes LayoutManager from PanelLayout. CurrentLayoutProvider creates a default layout if there is not a layout when a user opens the app.


jtbandes (approved):
Might be better to have some string there rather than nothing at all?

defunctzombie:
> Might be better to have some string there rather than nothing at all?

Not sure yet. On the surface, I agree and am still going through some additional iteration on when this might even happen. My goal is that it should not be possible for it to happen.

PR #6344, opened 2023-06-30

foxymiles:
**User-Facing Changes**
Improve raw messages panel rendering performance.

**Description**
The `Tooltip` and `StyledIconButton` components in `Value` seem to be expensive to render. Render them conditionally only when the value element is hovered.

This brings the FPS of a 3 panel layout on `/imu` on nuscenes from about 13fps to 37.

Screenshot 2023-06-30 at 10 45 56 AM
Screenshot 2023-06-30 at 10 46 15 AM

### DIFF
--- a/packages/studio-base/src/panels/RawMessages/Value.tsx
+++ b/packages/studio-base/src/panels/RawMessages/Value.tsx
@@ -196,6 +196,9 @@ function Value(props: ValueProps): JSX.Element {
     };
   }, []);
 
+  // The Tooltip and StyledIconButton components seem to be expensive to render so we
+  // track our hover state and render them conditionally only when this component is
+  // hovered.
   const [pointerOver, setPointerOver] = useState(false);
 
   return (
--- a/packages/studio-base/src/panels/RawMessages/Value.tsx
+++ b/packages/studio-base/src/panels/RawMessages/Value.tsx
@@ -32,9 +32,6 @@ const StyledIconButton = withStyles(HoverableIconButton, (theme) => ({
     "&.MuiIconButton-root": {
       fontSize: theme.typography.pxToRem(16),
       padding: 0,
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/panels/RawMessages/Value.tsx
```diff
@@ -196,27 +196,39 @@ function Value(props: ValueProps): JSX.Element {
     };
   }, []);
 
+  const [pointerOver, setPointerOver] = useState(false);
```
defunctzombie, line 199: needs a nice juicy comment explaining why and warning others of danger

defunctzombie (commented):
How did we hide the icon button and tooltip content before? Was there some `:hover` style in use that we need to remove now?

If we are going to use JS for the hover detection then lets remove the css stuff so it is clear what powers hovering and displaying.

foxymiles:
> How did we hide the icon button and tooltip content before? Was there some `:hover` style in use that we need to remove now?

There were some `:hover` styles in `StyledIconButton`. I've removed them.

defunctzombie (approved):

defunctzombie:
> There were some `:hover` styles in `StyledIconButton`. I've removed them.

With those removed do we have the correct opacity on the elements that show on the manually controlled hover? Was that set with other styling?

foxymiles:
> > There were some `:hover` styles in `StyledIconButton`. I've removed them.
> 
> With those removed do we have the correct opacity on the elements that show on the manually controlled hover? Was that set with other styling?

No I guess we need to set those back to 0.6.

foxymiles:
@defunctzombie https://github.com/foxglove/studio/pull/6345

jtbandes (commented):
🎉 🏃

PR #6346, opened 2023-06-30
Remove layoutId from URLStateSyncAdapter

defunctzombie:
**User-Facing Changes**
None

**Description**
Update SyncAdapter to conditionally include URLStateSyncAdapter and CurrentLayoutSyncAdapter if there are no sync adapters from AppContext.

### DIFF
--- a/packages/studio-base/src/components/SyncAdapters.tsx
+++ b/packages/studio-base/src/components/SyncAdapters.tsx
@@ -2,17 +2,26 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
+import { useMemo } from "react";
+
 import { CurrentLayoutSyncAdapter } from "@foxglove/studio-base/components/CurrentLayoutSyncAdapter";
 import { URLStateSyncAdapter } from "@foxglove/studio-base/components/URLStateSyncAdapter";
 import { useAppContext } from "@foxglove/studio-base/context/AppContext";
 
 export function SyncAdapters(): JSX.Element {
-  const { syncAdapters = [] } = useAppContext();
-  return (
-    <>
-      {...syncAdapters}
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/components/SyncAdapters.tsx
```diff
@@ -2,17 +2,26 @@
...
-      <URLStateSyncAdapter />
-      <CurrentLayoutSyncAdapter />
-    </>
-  );
+  // Sync adapters from app context override any local sync adapters
+  const { syncAdapters } = useAppContext();
+
+  return useMemo(() => {
+    if (syncAdapters) {
+      return <>{...syncAdapters}</>;
```
jtbandes, line 17: I know it was already there, but what's the `...` for here? What if we just did `<>{syncAdapters}</>` or even `return syncAdapters`?

packages/studio-base/src/components/SyncAdapters.tsx
```diff
@@ -2,17 +2,26 @@
...
+    }
+
+    return (
+      <>
+        <URLStateSyncAdapter />
+        <CurrentLayoutSyncAdapter />
```
jtbandes, line 23: Will the CurrentLayoutSyncAdapter be going away?

jtbandes (approved):

packages/studio-base/src/components/SyncAdapters.tsx
defunctzombie: I think the types of `syncAdapter` are not quite right to use without the `...`

packages/studio-base/src/components/SyncAdapters.tsx
defunctzombie: It will be re-formed into a layout saver for studio-base so it won't be going away because studio-base still needs to save/restore the current layout state.

PR #6349, opened 2023-07-03

foxymiles:
**User-Facing Changes**
Improve raw messages panel rendering performance.

**Description**
Improve raw messages panel rendering performance.

Includes the following changes:
1. Factor out the toolbar into a separate memoized component.
2. Memoize the CopyButton component and stabilize its properties.
3. Optimize rendering of the `MaybeCollapsedValue` component for the non-truncated case.

This improves rendering framerate for a layout with 7 panels rendering `/imu` from nuscenes from ~18fps to ~40fps:

before:
Screenshot 2023-07-03 at 10 23 34 AM

after:
Screenshot 2023-07-03 at 10 22 30 AM

And the flame chart no longer shows emotion as a significant contributor to render time:
Screenshot 2023-07-03 at 10 23 01 AM

### DIFF
--- a/packages/studio-base/src/components/CopyButton.tsx
+++ b/packages/studio-base/src/components/CopyButton.tsx
@@ -17,7 +17,7 @@ import { useCallback, useState, PropsWithChildren } from "react";
 
 import clipboard from "@foxglove/studio-base/util/clipboard";
 
-export default function CopyButton(
+function CopyButtonComponent(
   props: PropsWithChildren<{
     getText: () => string;
     size?: "small" | "medium" | "large";
@@ -84,3 +84,5 @@ export default function CopyButton(
     </Button>
   );
 }
+
+export default React.memo(CopyButtonComponent);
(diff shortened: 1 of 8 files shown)


packages/studio-base/src/components/CopyButton.tsx
```diff
@@ -84,3 +84,5 @@ export default function CopyButton(
     </Button>
   );
 }
+
+export default React.memo(CopyButtonComponent);
```
defunctzombie, line 88: I find react.memo for components quite fragile since any one prop break it and it is the _caller_ that has to know to do the right thing and that is non-obvious so it becomes overhead. We should go with this change here but I don't think its a generally applicable change.

defunctzombie (approved):

packages/studio-base/src/components/CopyButton.tsx
foxymiles: What do you propose as an alternative?

defunctzombie: I don't think there is a good alternative and a deficiency in the react approach. Maybe I'd expect react to handle this itself rather than wrapping each component in memo.

foxymiles: I agree that it's too easy to break a memoization graph and nuke performance but in a rerender-the-world, vdom framework like React I think we're stuck with it.

PR #6351, opened 2023-07-03
Limit text width in cells in table panel

foxymiles:
**User-Facing Changes**
None

**Description**
Limit text width in cells in the table panel.

Set to `75vw` max width, which is somewhat arbitrary but seems like an ok compromise.


2metres:
I think the user that gave us feedback on this wanted the values to wrap not truncate. If I remember correctly

foxymiles:
The slack thread suggested truncating, I'm guessing because if we were to just wrap the messages they used as examples they'd take up a full screen height each.

2metres (approved):

PR #6354, opened 2023-07-04
Remove uses of tss-react/withStyles

2metres:
**User-Facing Changes**
None

**Description**
Remove uses of withStyles as it has been show to have performance issues


defunctzombie:
Why is `withStyles` bad? Seems like it does what you are doing yourself inline?

2metres:
`withStyles` inherently listens to a children prop. So if the child is updating withStyles gets called again, `useStyles` have this same issue because the style generation and component are separated

jtbandes (approved):
NO strong opinion

PR #6355, opened 2023-07-04
Remove backfillDuration from player interface

defunctzombie:
**User-Facing Changes**
None

**Description**
This argument is not used at any call-site.


jtbandes (approved):

defunctzombie:
Thanks for the test fix!

PR #6356, opened 2023-07-04
Support org url param

wimagguc:
**User-Facing Changes**

When accessing a url with an `org` parameter, the logged in user may be signed in with the incorrect org for that resource. With this PR we'll be able to recognize this and help the user take the necessary next step:

249537633-8412854e-1ea5-4d1e-9e43-052b94ae2b66

**Description**

Under the hood this approach only _extends_ on existing code paths, to sneak in the required changes:
* Besides `currentUserRequired`, we also return `userSwitchRequired` inside `useInitialDeepLinkState`
* The existing `SignInFormModal` component handles the signout confirmation dialog if needed


packages/studio-base/src/components/SignInFormModal.tsx
```diff
@@ -4,13 +4,34 @@
...
   const [open, setOpen] = useState(true);
+  const { signOut } = useCurrentUser();
+  const [confirm, confirmModal] = useConfirm();
+
+  useEffect(() => {
+    if (!userSwitchRequired) {
+      return;
+    }
+    void confirm({ title: "You need to switch users", ok: "Sign out" }).then((response) => {
+      if (response === "ok") {
```
defunctzombie, line 26: What happens if the response is not ok?

packages/studio-base/src/hooks/useInitialDeepLinkState.ts
```diff
@@ -150,21 +151,32 @@ export function useInitialDeepLinkState(deepLinks: readonly string[]): {
...
-  useSyncSourceFromUrl(targetUrlState, currentUserRequiredParam);
-  useSyncLayoutFromUrl(targetUrlState, currentUserRequiredParam);
+
+  useSyncSourceFromUrl(targetUrlState, { currentUserRequired });
+  useSyncLayoutFromUrl(targetUrlState, { currentUserRequired });
   useSyncTimeFromUrl(targetUrlState);
-  return currentUserRequiredParam;
+
+  const { currentUser } = useCurrentUser();
+  const userSwitchRequired = useMemo(
```
defunctzombie, line 173: I don't think you need a memo here - the memo check itself isn't gonna save you much over the boolean expression from what I can tell.

defunctzombie:
This makes sense to me. I'd like to see it in action.

packages/studio-base/src/components/SignInFormModal.tsx
wimagguc: The confirm dialog disappears, so the user is free to open another data source or do something else in there. (The dialog won't reappear unless the user tries to reopen the link.)

defunctzombie:
@wimagguc lets go with `org` as the url parameter and see how we feel about that

defunctzombie (approved):

PR #6358, opened 2023-07-05
Add import/export layout actions to View menu

defunctzombie:
**User-Facing Changes**
Users can import/export the current layout using the _View_ menu.

Screenshot 2023-07-05 at 10 50 46 AM

**Description**

This change adds two new items to the _View_ menu. These items allow a user to export the current layout to a file or import a file to load as the current layout.


2metres:
I think so. The only other place I could see adding it would be under a dedicated Layout/Workspace item

2metres:
I would also advocate for adding a keyboard shortcut like Cmd `I` or `L`

packages/studio-base/src/components/AppBar/AppMenu.tsx
```diff
@@ -178,9 +178,31 @@ export function AppMenu(props: AppMenuProps): JSX.Element {
           handleNestedMenuClose();
         },
       },
+      {
+        type: "divider",
+      },
+      {
+        type: "item",
+        label: rightSidebarOpen ? t("importLayoutFromFile") : t("importLayoutFromFile"),
```
foxymiles, line 186: What is this ternary for?

foxymiles (approved):

packages/studio-base/src/components/AppBar/AppMenu.tsx
defunctzombie: A sloppy copy. Removed.

PR #6359, opened 2023-07-05
Allow custom layer URDF to use param:// URLs, add frame prefix setting

achim-k:
**User-Facing Changes**
- Custom layer URDF: Allow `param://` URLs and add frame prefix setting

**Description**
Adds the possibility to specify `param://<param_name>` URLs in the custom layer URDF. This will make it easier to visualize multiple URDFs when using datasources that provide parameter support (foxglove websocket, rosbridge, ros1 native). Besides allowing to load URDFs from a parameter, this PR also adds support for the frame prefix (also commonly called tfPrefix) setting. This setting allows to visualize different robots that share the same URDF definition but the [robot_state_publisher](https://index.ros.org/p/robot_state_publisher/github-ros-robot_state_publisher) publishes their transforms with a individual prefix.

Fixes FG-3292
Fixes FG-4057

Related: 
- https://github.com/orgs/foxglove/discussions/407
- https://github.com/orgs/foxglove/discussions/759
- https://github.com/orgs/foxglove/discussions/712

image

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
@@ -490,6 +490,26 @@ export class Urdfs extends SceneExtension<UrdfRenderable> {
     }
     this.renderer.settings.errors.remove(renderable.userData.settingsPath, VALID_URL_ERR);
 
+    const { protocol } = new URL(url);
+
+    // Special case: Retrieve URDF from parameter store.
+    if (protocol === "param:") {
+      const paramName = url.slice("param://".length);
+      const urdfParam = this.renderer.parameters?.get(paramName) as string | undefined;
+
+      if (urdfParam) {
+        this.#loadUrdf(instanceId, urdfParam);
+        return;
+      } else {
+        this.renderer.settings.errors.add(
+          renderable.userData.settingsPath,
+          VALID_URL_ERR,
+          `Invalid parameter URL "${url}": The parameter "${paramName}" does not exist`,
+        );
+        return;
+      }
+    }
+
     // Check if this URL has already been fetched
     if (renderable.userData.url === url) {
       return;
@@ -811,7 +831,7 @@ function createMeshMarker(
   };
 }
 
-const VALID_PROTOCOLS = ["https:", "http:", "file:", "data:", "package:"];
+const VALID_PROTOCOLS = ["https:", "http:", "file:", "data:", "package:", "param:"];
 
 function isValidUrl(str: string): boolean {
   try {
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
@@ -245,7 +245,7 @@ export class Urdfs extends SceneExtension<UrdfRenderable> {
         const fields: SettingsTreeFields = {
           url: { label: "URL", input: "string", placeholder, help, value: config.url ?? "" },
           framePrefix: {
-            label: "frame prefix",
+            label: "Frame prefix",
             input: "string",
             help: "Prefix to apply to all frame names (also often called tfPrefix)",
             value: config.framePrefix ?? "",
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
```diff
@@ -247,6 +249,12 @@ export class Urdfs extends SceneExtension<UrdfRenderable> {
 
         const fields: SettingsTreeFields = {
           url: { label: "URL", input: "string", placeholder, help, value: config.url ?? "" },
+          framePrefix: {
+            label: "frame prefix",
```
snosenzo, line 253:
```suggestion
            label: "Frame prefix",
```

packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
```diff
@@ -456,10 +464,21 @@ export class Urdfs extends SceneExtension<UrdfRenderable> {
...
+    }
+
+    // Update custom layer URDFs that use param:// URLs.
+    for (const [instanceId, renderable] of this.renderables.entries()) {
+      const url = (renderable.userData.settings as Partial<LayerSettingsCustomUrdf>).url ?? "";
+      if (url.startsWith("param://")) {
+        const paramName = url.slice("param://".length);
+        const urdf = parameters?.get(paramName);
+        if (typeof urdf === "string") {
+          this.#loadUrdf(instanceId, urdf);
```
snosenzo, line 478: Do we need to do any cleanup of the old model when we load a new one? Are transforms cleaned up properly?

packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
```diff
@@ -668,13 +708,32 @@ export class Urdfs extends SceneExtension<UrdfRenderable> {
...
-async function parseUrdf(text: string): Promise<ParsedUrdf> {
+async function parseUrdf(text: string, framePrefix?: string): Promise<ParsedUrdf> {
   const fileFetcher = getFileFetch();
-
+  const applyFramePrefix = (name: string) => `${framePrefix}${name}`;
   try {
     log.debug(`Parsing ${text.length} byte URDF`);
     const robot = await parseRobot(text, fileFetcher);
 
+    if (framePrefix) {
```
snosenzo, line 718: why is it necessary to apply the frame prefix here?

snosenzo (approved):
not sure how easy it would be but it might be nice to get a story for this behavior as well.

packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
achim-k: You are right, that might be a problem. Especially when changing the frame prefix. Will look into this

achim-k:
> not sure how easy it would be but it might be nice to get a story for this behavior as well.

Will look into this. Will also need to create a story for #6409, so can combine this effort

packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
achim-k: https://github.com/foxglove/studio/pull/6359/commits/d7b4e2eb3c8e8819c69670f6fda6690ea8a625ac is now cleaning up old transforms. However, the list of coordinate frames will grow, there seems to be no way to clean them up?

snosenzo: That must mean that they still have child frames with transforms in them. In `removeTransforms` I believe we clean up empty frames.

snosenzo: Maybe that's for the case where you've received transforms from the source under one prefix but change the prefix to another? I expect the frames received from the source would be there and also the frames under the new prefix. However I would expect changing the prefix again to clear the frames under the second prefix given it hasn't received transforms from the source.

achim-k: I just confirmed that the size of `renderer.transformTree` and `renderer.coordinateFrameList` remains the same when changing prefix. So it seems that old frames / transforms are cleaned up correctly.

snosenzo (approved):

PR #6360, opened 2023-07-05
Replace yarn constraints with dependabot grouped version updates

jtbandes:
**User-Facing Changes**
None

**Description**
Use Dependabot's new native version groups feature (see https://github.com/dependabot/dependabot-core/issues/1296, https://github.blog/changelog/2023-06-30-grouped-version-updates-for-dependabot-public-beta/) instead of yarn constraints.

We no longer need to exclude packages such as `@storybook/testing-library` because dependabot does not require all versions within a group to be the same (which was a limitation of the yarn constraints-based approach). Instead it just ensures all dependencies are updated in one PR.


defunctzombie (approved):
👏 

@bryfox seems like something you'd be interested in for the console repo.

PR #6364, opened 2023-07-06
Delete LayoutManager

defunctzombie:
**User-Facing Changes**
None

**Description**
This change simplifies the CurrentLayoutProvider to manage the layout state without dealing with saving and loading. This responsibility is pushed to downstream components. The CurrentLayoutSyncAdapter is updated to save/restore the current layout to/from localStorage.

The PlayerManager is also simplified to avoid load/save behavior and instead this is pushed downstream into the CurrentLayoutSyncAdapter which is updated to load any sample layout into the CurrentLayoutProvider.

Resolves: FG-3348
Resolves: FG-3031
Resolves: FG-4151


foxymiles (approved):
👍 on making stores purely manage data and handling persistence in a separate layer.

defunctzombie:
`migratePanelState` needs to remain and be used to migrate layouts when opening

defunctzombie:
Updated the logic to call `migratePanelsState` when loading the layout from localstorage.

benchmark/src/layouts.ts
```diff
@@ -11,46 +11,22 @@ import PointcloudRawMessageAnd3d from "./layouts/PointcloudRawMessageAnd3d.json"
...
-    PointcloudMultipleThreeDee,
-  ),
-  panelStateJsonToLayout("TransformPreloading", "Transform Preloading", TransformPreloading),
-];
-
-const LAYOUTS = new Map<string, Layout>();
-for (const layout of layouts) {
-  LAYOUTS.set(layout.id, layout);
-}
+const LAYOUTS: Record<string, LayoutData> = {
```
jtbandes, line 23: nit: probably shouldn't need both `as const` and a type annotation on the variable?

packages/studio-base/src/App.tsx
```diff
@@ -99,10 +92,6 @@ export function App(props: AppProps): JSX.Element {
 
   if (extraProviders) {
     providers.unshift(...extraProviders);
-  } else {
-    // Extra providers have their own layout providers
-    providers.unshift(<LayoutManagerProvider />);
-    providers.unshift(<LayoutStorageContext.Provider value={layoutStorage} />);
```
jtbandes, line 105: magical

packages/studio-base/src/components/CurrentLayoutSyncAdapter.tsx
```diff
@@ -2,89 +2,80 @@
...
-/**
- * Observes changes in the current layout and asynchronously pushes them to the
- * layout manager.
- */
-export function CurrentLayoutSyncAdapter(): ReactNull {
-  const selectedLayout = useCurrentLayoutSelector(selectCurrentLayout);
+const KEY = "studio.layout";
 
-  const layoutManager = useLayoutManager();
+export function CurrentLayoutSyncAdapter(): JSX.Element {
```
jtbandes, line 29: nit: maybe name this something like `CurrentLayoutLocalStorageSyncAdapter` to indicate that it stores layouts in localStorage?

packages/studio-base/src/hooks/useInitialDeepLinkState.test.tsx
```diff
@@ -176,7 +170,5 @@ describe("Initial deep link state", () => {
       params: { bar: "baz" },
       type: "connection",
     });
-
-    expect(setSelectedLayoutId).toHaveBeenCalledWith("12345");
```
jtbandes, line 180: nit: there is still a `&layoutId=12345` above

packages/studio-base/src/providers/CurrentLayoutProvider/index.test.tsx
```diff
@@ -17,336 +16,88 @@ import {
...
-    mockUserProfile.getUserProfile.mockResolvedValue({ currentLayoutId: "example" });
-    mockUserProfile.setUserProfile.mockImplementation(async () => {
-      condvar.notifyAll();
     });
 
-    const { result, all } = renderTest({
-      mockLayoutManager,
-      mockUserProfile,
-    });
+    getByText("Incompatible layout version");
```
jtbandes, line 66: this looks strange - does it throw if not found? I'd wrap with a `expect(..).toBeDefined()` to make it clear that we expect it to return something

packages/studio-base/src/services/migrateLayout/migratePanelsState.ts
jtbandes, line 1: nit: should we rename the directory from migrateLayout to migratePanelsState?

packages/studio-base/src/components/CurrentLayoutSyncAdapter.tsx
```diff
@@ -2,89 +2,80 @@
...
+  useEffect(() => {
+    if (selectedSource?.sampleLayout) {
+      setCurrentLayoutState({
+        selectedLayout: {
+          id: "default" as LayoutID,
+          data: selectedSource.sampleLayout,
+        },
+      });
+    }
+  }, [selectedSource, setCurrentLayoutState]);
```
jtbandes, line 44: This seems like it could be a bit subtle and might deserve a comment. It seems like if someone calls `selectSource()` (with a source that has a sample layout) and then immediately calls `setCurrentLayoutState()`, the latter might get overwritten by this effect – is that right? I don't think it's a big problem, just trying to understand the possible interactions.

jtbandes (approved):

packages/studio-base/src/components/CurrentLayoutSyncAdapter.tsx
defunctzombie: What would call `setCurrentLayoutState` before this effect runs? I am not sure what comment I'd write here. I do agree the various effects at work in this component are confusing. Might be something I tackle on the cleanup pass unless there's a specific improvement you think would be better here.

packages/studio-base/src/services/migrateLayout/migratePanelsState.ts
defunctzombie: Not particularly drawn to making that change. I think it feels like a wash to me - its still a layout migration (from the POV of what a layout means in this repo.

packages/studio-base/src/providers/CurrentLayoutProvider/index.test.tsx
defunctzombie: It does throw. I wrapped it as suggested.

packages/studio-base/src/hooks/useInitialDeepLinkState.test.tsx
defunctzombie: Removed.

packages/studio-base/src/components/CurrentLayoutSyncAdapter.tsx
defunctzombie: Renamed.

benchmark/src/layouts.ts
defunctzombie: Removed

PR #6373, opened 2023-07-06
3D: reduce log spam

jtbandes:
**User-Facing Changes**
None

**Description**
Reduce `debug` logs from the 3D & Image panels.

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
@@ -1022,8 +1022,10 @@ export class Renderer extends EventEmitter<RendererEvents> implements IRenderer 
   }
 
   public setFollowFrameId(frameId: string | undefined): void {
+    if (this.followFrameId !== frameId) {
+      log.debug(`Setting followFrameId to ${frameId}`);
+    }
     this.followFrameId = frameId;
-    log.debug(`Setting followFrameId to ${frameId}`);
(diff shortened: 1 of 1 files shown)


snosenzo (approved):

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -1022,8 +1022,10 @@ export class Renderer extends EventEmitter<RendererEvents> implements IRenderer
   }
 
   public setFollowFrameId(frameId: string | undefined): void {
```
defunctzombie, line 1024: Should this function be getting called every frame? Maybe that itself points to something nefarious upstream? The debug statement itself didn't seem out of place.

defunctzombie: Seems to push the responsibility on the function to identify whether it should be responsible for tracking old state to avoid "doing work" on calls with new parameters. A passing observation.

jtbandes: Would the code/architecture be improved if every call site would need to check the previous value?

defunctzombie: unclear - In some instances that's already a by-product of the dependency array, effects, etc? The calls don't happen unless there are changes.

PR #6376, opened 2023-07-06
Expose foxglove schemas to user scripts

foxymiles:
**User-Facing Changes**
Expose foxglove schema definitions to user scripts.

**Description**
Expose foxglove schema definitions to user scripts. Scripts can import type definitions from `@foxglove/schemas` in the same way they can in code outside of the user script editor.

Typescript enums are mapped to uint32 ROS types since ROS doesn't have a native enum type.

Addresses https://github.com/orgs/foxglove/discussions/716


foxymiles:
@esthersweon We should probably mention this in the docs.

packages/studio-base/src/panels/NodePlayground/Editor.tsx
```diff
@@ -240,12 +241,25 @@ const Editor = ({
...
+          monaco.languages.typescript.typescriptDefaults.addExtraLib(
+            lib.sourceCode,
+            `file:///node_modules/${lib.fileName}`,
+          );
+        } else {
+          monaco.languages.typescript.typescriptDefaults.addExtraLib(
+            lib.sourceCode,
+            `file:///node_modules/@types/${lib.fileName}`,
+          );
+        }
```
jtbandes, line 255: Rather than `if/else` here, would it be more clear if we set the `fileName` to include the `@types/` for the other libs we generate?

defunctzombie:
> @esthersweon We should probably mention this in the docs.

Good opportunity to make a linear issue for this or next cycle so we don't forget :D

packages/studio-base/src/panels/NodePlayground/Editor.tsx
```diff
@@ -240,12 +241,25 @@ const Editor = ({
...
+      });
+      monaco.languages.typescript.typescriptDefaults.setCompilerOptions({
+        ...monaco.languages.typescript.typescriptDefaults.getCompilerOptions(),
+        // This is needed for @foxglove/schemas to resolve correctly in the editor.
```
defunctzombie, line 259: Why?

packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/typescript/foxgloveSchemaDeclarations.ts
```diff
@@ -0,0 +1,108 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
```
defunctzombie, line 4: Lets add a comment to this file letting us know its purpose. My guess is that this file provides the declarations for making the @foxglove/schemas import work?

packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/typescript/types.ts
```diff
@@ -0,0 +1,16 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+export type NodeProjectFile = {
```
defunctzombie, line 5: I know these didn't have comments before - but maybe you've learned their purpose and could help our future selves out by adding some learnings on how these connect into the larger user-script environment or config.

defunctzombie (commented):
Good stuff.

How do we help users discover this? Does it make sense to update our default template (or other templates) to make this obvious for first time and repeat users? I'm imagining that its a reasonably common activity to use a user-script to output a well-known schema.

packages/studio-base/src/panels/NodePlayground/Editor.tsx
foxymiles: Not sure. Without this it seems to think `@foxglove/schemas/blah` is an alias.

packages/studio-base/src/panels/NodePlayground/Editor.tsx
foxymiles: Maybe although that would expand the scope of this PR a fair bit.

foxymiles:
> How do we help users discover this? Does it make sense to update our default template (or other templates) to make this obvious for first time and repeat users?

Not sure. I guess at least some of our templates could use this.

defunctzombie:
> Not sure. I guess at least some of our templates could use this.

What do we need to get sure? Something good to add as follow-up tickets if we don't want to do it with this PR but its important that we actually make this discoverable by more than word of mouth.

defunctzombie:
I tried the preview deployment and got this error with the nuscenes data:

image

```typescript
import { Log } from "@foxglove/schemas";

type Output = Log;

export const output = "/studio_script/output_topic";

export default function script(event: Input<"/gps">): Output {
  return {
    file: "",
    message: JSON.stringify(event.message),
    name: "",
    line: 0,
    level: 1,
    timestamp: event.receiveTime,
  };
}
```

Were you able to get user scripts to work with this deployment?

foxymiles:
> Were you able to get user scripts to work with this deployment?

I've tried with a bunch of types which all worked. There seems to be an issue specifically with `Log`.

foxymiles:
> > Not sure. I guess at least some of our templates could use this.
> 
> What do we need to get sure? Something good to add as follow-up tickets if we don't want to do it with this PR but its important that we actually make this discoverable by more than word of mouth.

Somebody needs to spend more time in the bowels of the Monaco typescript configuration.

defunctzombie:
> There seems to be an issue specifically with `Log`.

I'm gonna claim the QA of the month for this one.

defunctzombie:
@foxymiles I am still seeing the same error when trying to use `import { Log } from "@foxglove/schemas";` How were you able to get this to work?

foxymiles:
> @foxymiles I am still seeing the same error when trying to use `import { Log } from "@foxglove/schemas";` How were you able to get this to work?

No there's still an obscure error with Log. Every other type I've tried seems to work fine.

defunctzombie (changes requested):
We need to resolved the error with using the `Log` schema.

foxymiles:
@defunctzombie This script with all types defined inline also generates the same error on main so I think we need to address this in the schemas package:

```typescript

type Time = {
  sec: number;
  nsec: number;
};

export enum LogLevel {
  UNKNOWN = 0,
  DEBUG = 1,
  INFO = 2,
  WARNING = 3,
  ERROR = 4,
  FATAL = 5,
}

type Log = {
  /** Timestamp of log message */
  timestamp: Time;

  /** Log level */
  level: LogLevel;

  /** Log message */
  message: string;

  /** Process or node name */
  name: string;

  /** Filename */
  file: string;

  /** Line number in the file */
  line: number;
};

type Output = {
  log: Log;
};

export const output = "/studio_script/output_topic";

export default function script(event: Input<"/gps">): Output {
  return {
    log: {
      file: "",
      message: JSON.stringify(event.message),
      name: "",
      line: 0,
      level: 1,
      timestamp: event.receiveTime,
    },
  };
}
```

foxymiles:
Perhaps related: https://github.com/cruise-automation/webviz/issues/613

defunctzombie:
@foxymiles When you open a data source that has a `foxglove.Log` message you can use `Message<"foxglove.Log">` and that works? What is different about the type we create from the data source schema? Is it the LogLevel type?

The issue you linked mentions union but that's not the same error here.

foxymiles:
> @foxymiles When you open a data source that has a `foxglove.Log` message you can use `Message<"foxglove.Log">` and that works? What is different about the type we create from the data source schema? Is it the LogLevel type?

Not sure but this is clearly an issue with our user script TS environment and not specific to our `@foxglove/schemas` integrtation and doesn't seem like something that should block this PR and keep this feature out of user's hands.

foxymiles:
@snosenzo did some good detective work on this in slack.

foxymiles:
I think we should open a ticket to track that something like this doesn't work today and generates the same error message:

```typescript
// The ./types module provides helper types for your Input events and messages.

enum MyEnum {
  A,
  B,
}

type Output = {
  hello: string;
  enum: MyEnum;
};

export const inputs = ["/imu"];

export const output = "/studio_script/output_topic";

export default function script(event: Input<"/imu">): Output {
  return {
    hello: "world!",
    enum: MyEnum.A,
  };
}
```

packages/studio-base/src/panels/NodePlayground/Editor.tsx
```diff
@@ -240,12 +241,25 @@ const Editor = ({
...
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        moduleResolution: ModuleResolutionKind.NodeNext as any,
```
jtbandes, line 261: why doesn't the moduleResolution param accept a ModuleResolutionKind? what does it accept and should we be using that type instead?

jtbandes (approved):
Do we need to think about what happens if a data source is loaded that uses a newer version of schemas? Will users be confused if the `@foxglove/schemas` import is out of date? Or maybe it will be clear that `@foxglove/schemas` comes from an embedded package and if you want to use your data source types, you import from the types lib.

What do you think it would take to get auto-import working? So e.g. if I start typing SceneEntity in my function, the editor would be able to add the import for me?

jtbandes (changes requested):
I just realized another issue that needs to be addressed for this to work. Although we are able to generate "datatypes" from these imported schemas, the `schemaName` going downstream is currently wrong. Since there is no logic to associate an import from `@foxglove/schemas` with a particular `schemaName`, it is using the fallback path of generating a schema name from the user script name: 
image

In order for panels to be able to use the outputs of these scripts properly, the schemaName will need to be correct.

foxymiles:
> I just realized another issue that needs to be addressed for this to work. Although we are able to generate "datatypes" from these imported schemas, the `schemaName` going downstream is currently wrong.

Good observation. Do you think we should block this until we can find a solution? It's no worse than using a user-defined type in a script today right?

defunctzombie:
> Do you think we should block this until we can find a solution? It's no worse than using a user-defined type in a script today right?

Yes. We can't ship it without this because without the correct schema name the panels that work on schema name (3d, log, map) will not display the messages you produce. Think about how someone might want to use their scripts. Without this feature this script does not actually serve the intended purpose.

snosenzo:
> the schemaName going downstream is currently wrong.

We should definitely create a story to test this too with a userscript and raw message panel side-by-side

jtbandes:
We can test it with a unit test showing the extracted datatypes, doesn’t need to be a story.

packages/studio-base/src/panels/NodePlayground/Editor.tsx
foxymiles: Seems like monaco is out of sync with typescript for this type.

packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/transform.test.ts
```diff
@@ -1135,6 +1160,22 @@ describe("pipeline", () => {
         datatypes: basicDatatypes,
         outputDatatype: "std_msgs/ColorRGBA",
       },
+      {
+        description: "Should detect output datatype from @foxglove/schemas",
+        sourceCode: `
+          import { Color } from "@foxglove/schemas";
```
jtbandes, line 1166: Include all the `generateFoxgloveSchemaDeclarations()` here so they're tested in the unit tests too and not just storybook?

packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/typescript/projectConfig.ts
```diff
@@ -11,41 +11,52 @@
...
+import { lib_dts, lib_filename } from "./lib";
+import { UserScriptProjectConfig, UserScriptProjectFile } from "./types";
 
-export type NodeProjectFile = {
-  fileName: string;
-  filePath: string;
-  sourceCode: string;
-};
+/**
+ * Generates virtual ts files for each type exported by the @foxglobve/schemas package.
```
jtbandes, line 28:
```suggestion
 * Generates virtual ts files for each type exported by the @foxglove/schemas package.
```

jtbandes (approved):

defunctzombie (approved):
Glad to see a bit of grit and teamwork got us to the result we wanted for this feature. Our users will be glad as well.

packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/typescript/ast.ts
```diff
@@ -289,6 +298,19 @@ export const constructDatatypes = (
     }
   }
 
+  // Check if the source file comes from our @foxglove/schemas package. If it does, we can
+  // use a well-known name for the datatype.
+  const schemasFileRegex = /^\/studio_script\/node_modules\/@foxglove\/schemas\/(\w+)\.ts$/;
+
+  const match = sourceFileName.match(schemasFileRegex);
+  const schemaName = match?.[1];
```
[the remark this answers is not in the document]
jtbandes: To be completely sure we can have a unit test for each of the foxglove schemas to ensure it can be correctly extracted if used as the return type of the function.

packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/transform.test.ts
foxymiles: Since we're just doing a regex on the filename I don't think that would add much value and we'd have to write a structurally correct test for each type.

PR #6378, opened 2023-07-06
perf: optimized useDeepMemo with dependencies

shuhaowu:
**User-Facing Changes**
None.

Analysis
--------

I'm looking at invocations of `useDeepMemo` inside `useCachedGetMessagePathDataItems` and found that `useDeepMemo` is doing a deep equal on basically every iteration in some cases where user scripts are registered. This is due to the following chain of events:

1. The deep memo inputs are `unmemoizedRelevantDatatypes` and `unmemoizedRelevantTopics`. These are computed via an `useMemo` with the effective dependencies of `[providerTopics, datatypes, memoizedPaths]`. So as long as the message path is not changing (it's not) and the provider topics&datatypes are not changing (not while during normal playback), the `unmemoizedRelevant*` references shouldn't be changing.
2. However, at the startup of the UI, the `providerTopics` and `datatypes` can change rapidly, due to the registration of user scripts. Let's say `useCachedGetMessagePathDataItems` is called first before the user-script-derived topics and datatypes are registered with the player. Let's use `A` to denote the references to the computed `unmemoizedRelevant*` at this point. `A` is passed to `useDeepMemo`, which internally sets `ref.current = A`. _Note: assume the message paths are for non-user-script topics._
3. If `useCachedGetMessagePathDataItems` is called multiple times during this window when user script has not been registered, the `useDeepMemo` will be "free" as the lodash `isEqual` call internally will basically check `if A === A` and return.
4. At some point, the user scripts registers, changing `providerTopics` and `datatypes`. When `useCachedGetMessagePathDataItems` gets called again, it will recompute the `unmemoizedRelevant*` values and their references change. Let's denote the references to these values at this point to be `B`. `B` is passed to `useDeepMemo`.
5. Inside `useDeepMemo`, `ref.current` is set to `A` . `B` is the same as `A` in value, but it's a different reference, which means `isEqual` is going to do a deep equal comparison. This comparison returns true and the `useDeepMemo` returns `A`.
6. Since the `providerTopics` and `datatypes` no longer change after all user scripts have been registered, the `unmemoizedRelevant*` references will always be `B`, which is passed to `useDeepMemo`, which retains a reference to `A`. This means a deep comparison will be on every frame!

Testing
-------

Note this problem only shows up occasionally and sometimes unpredictably, as this only shows up when the
`useCachedGetMessagePathDataItems` function wins the race with the user script topic registration. Depending on (1) the data type structures and (2) how often `useCachedGetMessagePathDataItems` is called, the deep comparison can have a significant performance impact on the UI. In our case, our data structure can be quite wide and nested and the function can be invoked very often as our custom panels are filtering for a relatively large number of user-specified message paths, this can be up to 10-15% of the sampled CPU stacks from the Chrome profiler.

With Foxglove studio, it's harder to replicate this bug's performance impact. Foxglove-native panels filters for less paths than our custom panels (Raw message, plot, and table), and that the race condition with the user scripts are more difficult to replicate. More testing is required here and I haven't had a chance to fully replicate this.

Solution
--------

I changed `useDeepMemo` to retain all deep equaled copies into a `WeakSet`. The hook first does a check to see if an object is in that `WeakSet` before performing the deep equal, which gets around this performance problem. `useDeepMemo` has to accept `undefined` as this is used in `useStateToURLSynchronization`.

### DIFF
--- a/packages/hooks/src/useDeepMemo.ts
+++ b/packages/hooks/src/useDeepMemo.ts
@@ -15,11 +15,35 @@ import { isEqual } from "lodash";
 import { useRef } from "react";
 
 // Continues to return the same instance as long as deep equality is maintained.
-export default function useDeepMemo<T>(value: T): T {
+export default function useDeepMemo<T extends object | undefined>(value: T): T {
   const ref = useRef(value);
-  if (isEqual(value, ref.current)) {
+  const deepEqualCopiesRef = useRef<WeakSet<object>>();
+
+  if (value == undefined) {
+    return value;
+  }
+
+  if (!deepEqualCopiesRef.current) {
+    deepEqualCopiesRef.current = new WeakSet();
+  }
+
+  // If A is passed to useDeepMemo first and get memoized, then B (which is
+  // equal to A in value but different in reference) is passed to useDeepMemo
+  // every time it is called henceforth, we need to ensure we don't run isEqual
+  // for every subsequent call. This is what the WeakSet allows us to do.
+  if (deepEqualCopiesRef.current.has(value)) {
     return ref.current;
   }
+
+  if (isEqual(value, ref.current)) {
+    deepEqualCopiesRef.current.add(value);
+    return ref.current;
+  }
+
+  // New value detected, so we need to remove all the deepEqualCopies weak set
+  // so we don't return the wrong value.
+  deepEqualCopiesRef.current = new WeakSet();
+  deepEqualCopiesRef.current.add(value);
   ref.current = value;
   return value;
 }
--- a/packages/hooks/src/useDeepMemo.test.ts
+++ b/packages/hooks/src/useDeepMemo.test.ts
@@ -18,7 +18,7 @@ import useDeepMemo from "./useDeepMemo";
 
 describe("useDeepMemo", () => {
   it("returns original object when deep equal", () => {
-    let obj: unknown = { x: 1 };
+    let obj: object = { x: 1 };
     const { result, rerender } = renderHook((val) => useDeepMemo(val), { initialProps: obj });
     expect(result.current).toBe(obj);
     rerender({ x: 1 });


packages/hooks/src/useDeepMemo.ts
```diff
@@ -15,11 +15,35 @@ import { isEqual } from "lodash";
...
+  if (deepEqualCopiesRef.current.has(value)) {
+    return ref.current;
+  }
+
   if (isEqual(value, ref.current)) {
+    deepEqualCopiesRef.current.add(value);
     return ref.current;
   }
+
+  // New value detected, so we need to remove all the deepEqualCopies weak map
```
jtbandes, line 43: nit: comment still refers to "weak map"

packages/studio-base/src/components/MessagePathSyntax/useCachedGetMessagePathDataItems.ts
```diff
@@ -66,6 +66,7 @@ export function useCachedGetMessagePathDataItems(
     }
     return filledInPaths;
   }, [globalVariables, memoizedPaths]);
+
```
jtbandes, line 69:
```suggestion
```

packages/hooks/src/useDeepMemo.ts
```diff
@@ -15,11 +15,35 @@ import { isEqual } from "lodash";
...
+  // so we don't return the wrong value.
+  deepEqualCopiesRef.current = new WeakSet();
+  deepEqualCopiesRef.current.add(value);
```
jtbandes, line 46: Why do we immediately `add(value)` here, but not on line 27?

shuhaowu: We can, but it's redundant.

1. `deepEqualCopiesRef.current` is only undefined at the first invocation (to my understand at least). So it gets created at line 27.
2. In the first invocation, we set `ref.current = value`.
3. The initial `deepEqualCopiesRef.current.has(value)` will fail, so that branch won't execute.
4. `isEqual(value, ref.current)` returns true and that branch is executed. So line 39 will then execute and add the value into the weak set.

The line here executes if the value passed to `useDeepMemo` changes. Naturally we need to clear out the WeakSet so we don't accidentally return the wrong value. We immediately add the value in so we don't wait until the next iteration to add the value into the Weak Set.

If it is clearer I could add this `.add` by line 27.

packages/hooks/src/useDeepMemo.ts
```diff
@@ -15,11 +15,35 @@ import { isEqual } from "lodash";
 import { useRef } from "react";
 
 // Continues to return the same instance as long as deep equality is maintained.
-export default function useDeepMemo<T>(value: T): T {
+export default function useDeepMemo<T extends object | undefined>(value: T): T {
   const ref = useRef(value);
+  const deepEqualCopiesRef = useRef<WeakSet<object>>();
```
jtbandes, line 20: can it be `WeakSet<T>`?

jtbandes (approved):
Makes sense to me, will also wait for @defunctzombie's opinion

packages/hooks/src/useDeepMemo.ts
shuhaowu: No. `T` can be `undefined` and it cannot be assigned to `object` as `WeakSet` is defined as `WeakSet<T extends object>`.

shuhaowu: And the reason we need undefined is because `useStateToURLSynchronization` has a line 

```
  const playerUrlState = useMessagePipeline(selectUrlState);
  const stablePlayerUrlState = useDeepMemo(playerUrlState);
```

`playerUrlState` can be `undefined`.

jtbandes: You could do:

```
useDeepMemo<T extends object>(value: T | undefined): T | undefined
```

Then `if (value == undefined)` should automatically refine the type to `value: T` and allow you to put it in the `WeakSet<T>`.

shuhaowu: I thought I tried this earlier and it didn't work, but it definitely works :shrug:. Probably because I didn't put `if (value == undefined)` in at that point... Anyways I updated this code to the suggestion above.

shuhaowu: Oh wait I spoke too soon! This doesn't quite work at another place where it is not expected for `useDeepMemo` to return `T | undefined` and I now get a bunch of "T is possibly undefined" error messages.

shuhaowu:
```
ERROR in ./packages/studio-base/src/components/MessagePathSyntax/useCachedGetMessagePathDataItems.ts:118:34
TS18048: 'relevantTopics' is possibly 'undefined'.
    116 |       }                                                                    
    117 |     }                                                                                                                                                            
  > 118 |     for (const { schemaName } of relevantTopics.values()) {
        |                                  ^^^^^^^^^^^^^^
    119 |       if (schemaName != undefined) {
    120 |         addRelevantDatatype(schemaName, []);
    121 |       }  

```

jtbandes: ah right. Well I think any solution to that would be more complicated than it's worth...`WeakSet<object>` is fine :)

defunctzombie:
Looking at the call-site where `useDeepMemo` is used and the `unmemoizedFilledInPaths` variable that it memoizes I wondered if we can drop the `useDeepMemo` altogether in favor of the memo already happening for computing `unmemoizedFilledInPaths`. I see that this logic is needed because variables may or may not be used in the paths and I'm thinking that what this logic tries to avoid is computing new `filledInPaths` instances for paths un-affected by variable changes.

Before I look too closely at the proposed `useDeepMemo` changes themselves I wonder if we can better improve the intent and remove the useDeepMemo? @foxymiles you working near this code recently and are actively working on more plot performance improvements.

On the code changes - @jtbandes I thought it was a react-anti-pattern to have side effects in render functions (and hook functions)? I see that we already had this for `useDeepMemo`.

---
Thinking more as I write this comment I am wondering if the hook to write here is a `useMemo` that does a deep comparison on its _output_. I guess that's what the use of these two back to back hooks is accomplishing. Amusingly this takes me back to something I've long wished I had with `useMemo` which was access to the existing memoized value (similar to useState).
```

**Comment by foxymiles** [comment]
This might be an opportunity to use a more full-featured memoization library like https://github.com/planttheidea/moize.

**Comment by foxymiles** [comment]
Yeah it seems to me like the root of the problem here is we do a useMemo then useDeepMemo on that. If we do a deep memoization within the `useMemo` that should fix it right?

It seems like the existing useDeepMemo implementation would handle more common cases.

File: packages/hooks/src/useDeepMemo.ts [hunk 5]
```diff
@@ -15,11 +15,35 @@ import { isEqual } from "lodash";
...
+
+  if (value == undefined) {
```
defunctzombie, line 22:
```suggestion
  if (value == undefined || value === ref.current) {
```

This will shortcut any weak set lookups or `isEqual` calls.

packages/hooks/src/useDeepMemo.ts
```diff
@@ -15,11 +15,35 @@ import { isEqual } from "lodash";
...
+    return value;
+  }
+
+  if (!deepEqualCopiesRef.current) {
+    deepEqualCopiesRef.current = new WeakSet();
+  }
+
+  // If A is passed to useDeepMemo first and get memoized, then B (which is
```
defunctzombie, line 30: To me a simpler explanation of what is happening and what you are adjusting this function to do is to _cache_ or _memoize_ when an input value reference had deep equality to avoid the expensive `isEqual` comparison. I drop this comment cause I personally find it confuses me more than what is actually happening here which is avoiding the expensive comparison.

I would add some form of my above statement to the hook documentation.

defunctzombie (commented):
I'm probably 50/50 on this but I see how it could be happening and how your proposed changes can help. After reading through the code what I see it doing is caching which input values were equal and remembering those references (weakly) with the statement that the WeakSet lookup is cheaper than the `isEqual` comparison which seems sensible for large-ish objects.

In a way, what this does is create a specialized form of this hook where the _compare_ function is a memoized form of `isEqual`:
https://github.com/react-hookz/web/blob/master/src/useCustomCompareMemo/index.ts#L13

I think you could achieve a similar result in our hook with:

```
import { memoize } from "memoize-weak";
const memoizedIsEqual = memoize(isEqual);
```

And using that within the hook. This will memoize the inputs and I believe achieve the same goal with fewer custom lines of logic. You could initialize the memoized `isEqual` within the hook if you wanted to scope its lifetime.

PR #6379, opened 2023-07-06
Move Image panel entry point to its own folder

jtbandes:
**User-Facing Changes**
None

**Description**
Add `panels/Image/index.tsx` as a more discoverable entry point for the Image panel.


packages/studio-base/src/panels/Image/index.tsx
```diff
@@ -0,0 +1,9 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+/* eslint-disable filenames/match-exported */
+
+import { ImagePanel } from "../ThreeDeeRender";
```
defunctzombie, line 7: We tend to avoid relative path imports that go _up_ the tree (i.e. `../`) in favor of referencing from @foxglove/studio-base.

packages/studio-base/src/panels/ThreeDeeRender/stories/ArrowMarkers.stories.tsx
```diff
@@ -17,7 +17,7 @@ import {
   SENSOR_FRAME_ID,
 } from "./common";
 import useDelayedFixture from "./useDelayedFixture";
-import { ThreeDeePanel } from "../index";
```
defunctzombie, line 20: Personally I am a fan of the `import { ThreeDeePanel }` style rather than the default export. I think default exports should die.

defunctzombie (approved):
Seems good. I filed the linear issue but leave it up to you on whether you think this was a net win for "where do I find the image panel code" or not.

packages/studio-base/src/panels/ThreeDeeRender/stories/ArrowMarkers.stories.tsx
jtbandes: In that case all the `module` functions in here would need to change: https://github.com/foxglove/studio/blob/main/packages/studio-base/src/panels/index.ts

jtbandes:
I do not particularly think it is a net win. I implemented it because it was requested in the issue 😄   @snosenzo thoughts?

packages/studio-base/src/panels/Image/index.tsx
jtbandes: Empirically we don't really avoid this 😆   We usually do relative imports when the files are part of the same logical "module".

Some day it would be nice if each panel were actually a separate package (an extension).

packages/studio-base/src/panels/ThreeDeeRender/stories/ArrowMarkers.stories.tsx
jtbandes: > I think default exports should die.

Until this becomes a linear issue or lint rule, it remains just a passing thought

snosenzo (approved):

snosenzo:
I think having the top level `Image` folder is a good win since people are expecting there to be a dedicated folder for it -- then having it link to `ThreeDeePanel` shows them where it's really at. Otherwise searching for `Image` in our application code is a bit tough

PR #6388, opened 2023-07-06
Image: fix alpha support for line annotations

jtbandes:
**User-Facing Changes**

**Description**
Fixes alpha for line annotations by using LineMaterialWithAlphaVertex.


snosenzo (approved):

defunctzombie:
@jtbandes fwiw I think this is worth calling out. It's a bugfix right?

jtbandes:
It’s a bug fix but it’s also restoring an existing feature (of the old image panel) that was only temporarily broken…

PR #6389, opened 2023-07-07
Add dependabot group for babel packages

jtbandes:
**User-Facing Changes**
None

**Description**
See #6360


defunctzombie (approved):

defunctzombie:
When can we rid ourselves of babel? I didn't realize we even depended on it directly.

jtbandes:
We use babel-jest, so currently "never".

PR #6390, opened 2023-07-07
Automatically approve Dependabot PRs

jtbandes:
**User-Facing Changes**
None

**Description**
Automatically approve Dependabot PRs when they are opened.

A small step towards FG-3299

### DIFF
--- a/.github/workflows/dependabot-approve.yml
+++ b/.github/workflows/dependabot-approve.yml
@@ -1,0 +1,27 @@
+name: Dependabot
+
+on:
+  pull_request_target:
+    types: [opened]
+
+jobs:
+  approve:
+    runs-on: ubuntu-latest
+    if: github.actor == 'dependabot[bot]'
+    steps:
+      - uses: actions/github-script@v6
+        name: Approve PR
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/dependabot-approve.yml
```diff
@@ -0,0 +1,20 @@
+name: Dependabot
```
defunctzombie, line 1: https://github.com/foxglove/studio/blob/main/.github/workflows/dependabot-fix.yml#L10 Is it ok to use the same name for two workflows?

Please add a comment describing the goals of this workflow.

jtbandes: https://stackoverflow.com/questions/70701626/github-actions-same-workflow-name-different-branches-which-takes-priority#:~:text=Two%20workflows%20with%20the%20same,appear%20using%20the%20same%20name.

defunctzombie (approved):
Not sure what I think about stack-overflow being the "citation" for this behavior but we can see how it goes. Also would be fine to call the workflows different names if we run into problems.

jtbandes:
📿

PR #6397, opened 2023-07-07
Add Font Settings for Raw Messages Panel

turtle-operator:
**User-Facing Changes**
- Add font setting for Raw Message panel
- Make single-value raw message display respect font setting

**Description**
Add option to overrule automatic large text display for primitive values in Raw Messages panel (as discussed in https://github.com/orgs/foxglove/discussions/618), as this takes a lot of screen real estate and might be counter-productive e.g. for long (multi-line) strings. For this, the font settings can be changed via the settings tree.

### DIFF
--- a/packages/studio-base/src/panels/RawMessages/index.tsx
+++ b/packages/studio-base/src/panels/RawMessages/index.tsx
@@ -19,7 +19,7 @@ import ReactHoverObserver from "react-hover-observer";
 import Tree from "react-json-tree";
 import { makeStyles } from "tss-react/mui";
 
-import { Immutable } from "@foxglove/studio";
+import { Immutable, SettingsTreeAction } from "@foxglove/studio";
 import { useDataSourceInfo } from "@foxglove/studio-base/PanelAPI";
 import EmptyState from "@foxglove/studio-base/components/EmptyState";
 import useGetItemStringWithTimezone from "@foxglove/studio-base/components/JsonTree/useGetItemStringWithTimezone";
@@ -44,6 +44,7 @@ import getDiff, {
   diffLabelsByLabelText,
 } from "@foxglove/studio-base/panels/RawMessages/getDiff";
 import { Topic } from "@foxglove/studio-base/players/types";
+import { usePanelSettingsTreeUpdate } from "@foxglove/studio-base/providers/PanelStateContextProvider";
 import { SaveConfig } from "@foxglove/studio-base/types/panels";
 import { enumValuesByDatatypeAndField } from "@foxglove/studio-base/util/enums";
 import { useJsonTreeTheme } from "@foxglove/studio-base/util/globalConstants";
@@ -92,11 +93,6 @@ const useStyles = makeStyles()((theme) => ({
     fontFamily: fonts.SANS_SERIF,
     fontFeatureSettings: `${theme.typography.fontFeatureSettings}, "zero"`,
   },
-  big: {
-    "&.MuiTypography-root": {
-      fontFeatureSettings: `${theme.typography.fontFeatureSettings}, "zero"`,
-    },
-  },
   hoverObserver: {
     display: "inline-flex",
     alignItems: "center",
@@ -113,6 +109,7 @@ function RawMessages(props: Props) {
   const { openSiblingPanel } = usePanelContext();
   const { topicPath, diffMethod, diffTopicPath, diffEnabled, showFullMessageForDiff } = config;
   const { topics, datatypes } = useDataSourceInfo();
+  const updatePanelSettingsTree = usePanelSettingsTreeUpdate();
 
   const defaultGetItemString = useGetItemStringWithTimezone();
   const getItemString = useMemo(
@@ -141,6 +138,7 @@ function RawMessages(props: Props) {
   }, [structures, topic, topicRosPath]);
 
   const [expansion, setExpansion] = useState(config.expansion);
+  const [customFontSize, setCustomFontSize] = useState<number | undefined>();
 
   // Pass an empty path to useMessageDataItem if our path doesn't resolve to a valid topic to avoid
   // spamming the message pipeline with useless subscription requests.
@@ -417,10 +415,9 @@ function RawMessages(props: Props) {
         />
         {shouldDisplaySingleVal ? (
           <Typography
-            className={classes.big}
             variant="h1"
-            fontWeight="bold"
-            whiteSpace="pre-line"
+            fontSize={customFontSize}
+            whiteSpace="pre-wrap"
             style={{ wordWrap: "break-word" }}
           >
             <MaybeCollapsedValue itemLabel={String(singleVal)} />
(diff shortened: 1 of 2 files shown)


defunctzombie:
Internal ticket: FG-4190

defunctzombie:
Could you remove the change for `Remove (maybe) collapsing values for single values.` to reduce the amount of things you are changing in one PR. We can focus this PR on adding font settings and respecting those font settings in the panel. Once that change is made we'll do a preview deployment on this PR and see how we like the UX.

defunctzombie:
@turtle-operator Is this PR ready for review or are you still working on the draft?

packages/studio-base/src/panels/RawMessages/types.ts
```diff
@@ -22,4 +22,6 @@ export type RawMessagesPanelConfig = {
 export const Constants = {
   CUSTOM_METHOD: "custom",
   PREV_MSG_METHOD: "previous message",
+  FONT_SIZE_OPTIONS: [8, 9, 10, 10.5, 11, 12, 14, 18, 20, 22, 24, 28, 32],
```
2metres, line 25: I would suggest updating size values to be more consistent with typography metrics 
```suggestion
  FONT_SIZE_OPTIONS: [8, 9, 10, 11, 12, 14, 16, 18, 24, 30, 36, 48, 60, 72],
```

packages/studio-base/src/panels/RawMessages/types.ts
```diff
@@ -22,4 +22,6 @@ export type RawMessagesPanelConfig = {
...
+  FONT_WEIGHT_OPTIONS: ["bold"],
```
2metres, line 26: Similarly with font weights we should call out the options we actually [provide we from the webfont](https://github.com/foxglove/studio/blob/8d68c29419b4c4c0d3354037b0bd5110cf261f0c/packages/studio-base/src/styles/assets/inter.css#L29-L55) otherwise we'll just get the browser's faux bolding

| Friendly name | CSS font weight |
|---|---|
| Regular | 400 |
| Medium | 500 |
| SemiBold (could be simplified to Bold)   | 600 |

packages/studio-base/src/panels/RawMessages/index.tsx
```diff
@@ -623,6 +614,59 @@ function RawMessages(props: Props) {
...
+            },
+            font_weight: {
+              label: "Font Weight",
+              input: "select",
+              options: [
+                { label: "default", value: undefined },
+                ...Constants.FONT_WEIGHT_OPTIONS.map((font_weight) => ({
+                  label: font_weight,
+                  value: font_weight,
+                })),
```
2metres, line 660: See comment [below](#diff-6dae59b1df06a111a3fdc2b0899401bcb9b1ce7402dea843240f35e0c312bc92R26)

packages/studio-base/src/panels/RawMessages/index.tsx
```diff
@@ -623,6 +614,59 @@ function RawMessages(props: Props) {
...
+  }, []);
+
+  useEffect(() => {
+    updatePanelSettingsTree({
+      actionHandler,
+      nodes: {
+        general: {
+          label: "General",
+          fields: {
+            font_size: {
```
2metres, line 640: 🍹 Nit: Elsewhere in the app we tend to use camelCase is used for setting field names.

packages/studio-base/src/panels/RawMessages/index.tsx
```diff
@@ -623,6 +614,59 @@ function RawMessages(props: Props) {
...
+              options: [
+                { label: "default", value: undefined },
+                ...Constants.FONT_SIZE_OPTIONS.map((front_size) => ({
+                  label: front_size.toString(),
+                  value: front_size,
+                })),
+              ],
+              value: customFontSize,
+            },
+            font_weight: {
```
2metres, line 652: Same feedback [as above](#diff-ca8a3a46914acc5156ff9c15ee6b85815754f4f4a02f2f2cd0e6d2c718cd2caaR640)

2metres (changes requested):

packages/studio-base/src/panels/RawMessages/index.tsx
```diff
@@ -623,6 +614,74 @@ function RawMessages(props: Props) {
...
+                {
+                  label: "Medium",
+                  value: theme.typography.fontWeightMedium as number,
+                },
+                {
+                  label: "Bold",
+                  value: theme.typography.fontWeightBold as number,
+                },
+              ],
+              value: customFontWeight,
```
snosenzo, line 668: should we have default placeholder values for these?

snosenzo: Right now it shows this by default because this `value` field is `undefined`: 
image
By doing something like this:
```suggestion
              value: customFontWeight ?? (theme.typography.fontWeightRegular as number),
```
Doing this we can get rid of the red border and show what we use by default when `customFontWeight` (and `customFontSize`) are undefined.

packages/studio-base/src/panels/RawMessages/index.tsx
```diff
@@ -623,6 +614,70 @@ function RawMessages(props: Props) {
...
+              options: [
+                ...Constants.FONT_SIZE_OPTIONS.map((value) => ({
+                  label: `${value} px` + (value === theme.typography.fontSize ? ` (Default)` : ``),
+                  value,
+                })),
+              ],
+              value: customFontSize,
+            },
+            fontWeight: {
+              label: "Font Weight",
```
esthersweon, line 648:
```suggestion
              label: "Font weight",
```

amacneil:
Feedback from internal discussion:
- Need to remove indentation when viewing a single value
- Add `Auto` at top of font size dropdown which keeps current behavior (large font for single values), and remove `Default` label. When a font size is selected, it overrides the automatic behavior for both single values and trees. Justification: we don't have any other data points from users finding the current automatic behavior painful, so we don't want to break existing user workflows
- Remove `Font Weight` option and make it regular weight always (even when in auto mode) - keep things simple.

defunctzombie (approved):
LGTM! Thanks for going through the feedback and helping us polish the product.

Please sign the CLA and I'll be able to merge this.

2metres (approved):

packages/studio-base/src/panels/RawMessages/index.tsx
```diff
@@ -623,6 +625,43 @@ function RawMessages(props: Props) {
...
+      actionHandler,
+      nodes: {
+        general: {
+          label: "General",
+          fields: {
+            fontSize: {
+              label: "Font size",
+              input: "select",
+              options: [
+                { label: "auto", value: undefined },
```
jtbandes, line 651: There are a few things about this that seem confusing to me.
- The "General > Font size" setting does not generally change the font size of the panel. It only changes it in the circumstance where you've drilled down to a single value — that seems like something we'd need to explain better in the UI, otherwise users will be confused when the dropdown doesn't do anything.
- There's nothing automatic about the "auto" setting — it just selects a particular size that happens to be inherited from the `h1` styling or something like that.
- The dropdown feels like it could be a numeric input field/slider? Not sure of the reasoning for using a dropdown.

We'll discuss these internally.

jtbandes: Never mind, seems like I was mistaken! 🙌  Sorry about that

PR #6399, opened 2023-07-07
Add lint rule to ban default exports

defunctzombie:
**User-Facing Changes**
None

**Description**


defunctzombie:
@jtbandes @snosenzo @foxymiles I've updated a bunch of files using the vscode refactor feature (and suppressed this rule for `.stories.tsx` files which require a default export. Before I go further, take a look and let me know if you think this is a sensible path forward given the changes you see.

Also - some components that use `export default React.memo(...);` inline will need updating to become named. I have not tackled that in any files here.

packages/studio-base/src/players/UserNodePlayer/nodeTransformerWorker/typescript/rawUserUtils.ts
```diff
@@ -19,3 +19,5 @@ export default [
   { fileName: "vectors.ts", sourceCode: vectors },
   { fileName: "markers.ts", sourceCode: markers },
 ];
+
+export { rawUserUtils };
```
jtbandes, line 23: Maybe I should make a separate lint rule to ban `export { }` in favor of `export const` 🔪

jtbandes (commented):
👍 excited for more consistency

PR #6400, opened 2023-07-07
Add ability to override default layout in Docker image

jtbandes:
**User-Facing Changes**
The Docker image now supports specifying a default layout via bind mount (`-v /path/to/custom_layout.json:/foxglove/default-layout.json`).

**Description**
Relates to: #4716, https://github.com/orgs/foxglove/discussions/217
Resolves FG-3318


banisadr (approved):
Looks good to me. There are two use cases this doesn't solve, but I don't think they should block this:
1. Scenario in which an user running the docker image has a static set of 2+ layouts they want to be available for switching between while running (i.e. a directory w/ multiple layouts in it).
2. Online layout sharing which - this we want to address with 1st class layout sharing.

defunctzombie (approved):
Quite the gymnastics with the array index but I understand. Code and Dockerfile could benefit from some explainer comments if nothing more than to indicate what this code powers so we know what related stuff to search for if we want to change it.

Guess we will commit to supporting this workflow or do we want to say anything about the stability of this volume mount for future versions?

jtbandes:
> Code and Dockerfile could benefit from some explainer comments if nothing more than to indicate what this code powers so we know what related stuff to search for

I guess I figured the giant `FOXGLOVE_STUDIO_DEFAULT_LAYOUT_PLACEHOLDER` would work well as a search keyword, but will add more comments

PR #6409, opened 2023-07-08
Allow to display URDF collisions

achim-k:
**User-Facing Changes**
Allow to display URDF collisions

**Description**
~Depends on #6359~

Adds a `display mode` configuration setting to the URDF visualization with the following options:
- `auto` - Display visual geometries. If link has no visuals, display collision geometries (current behavior)
- `visual` - Only display visual geometries
- `collision` - Only display collision geometries

Another option would be to do it like Rviz where there are two checkboxes `Visual enabled` and `Collision enabled`.

[Screencast from 08.07.2023 12:56:23.webm](https://github.com/foxglove/studio/assets/9250155/6c327015-1552-4aa7-8444-d8940baa22c5)

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
@@ -10,7 +10,12 @@ import * as THREE from "three";
 import { UrdfGeometryMesh, UrdfRobot, UrdfVisual, parseRobot, UrdfJoint } from "@foxglove/den/urdf";
 import Logger from "@foxglove/log";
 import { toNanoSec } from "@foxglove/rostime";
-import { SettingsTreeAction, SettingsTreeChildren, SettingsTreeFields } from "@foxglove/studio";
+import {
+  SettingsTreeAction,
+  SettingsTreeChildren,
+  SettingsTreeField,
+  SettingsTreeFields,
+} from "@foxglove/studio";
 import { eulerToQuaternion } from "@foxglove/studio-base/util/geometry";
 
 import { RenderableCube } from "./markers/RenderableCube";
@@ -65,6 +70,7 @@ const RPY_LABEL: [string, string, string] = ["R", "P", "Y"];
 
 export type LayerSettingsUrdf = BaseSettings & {
   instanceId: string; // This will be set to the topic name
+  displayMode: "auto" | "visual" | "collision";
 };
 
 export type LayerSettingsCustomUrdf = CustomLayerSettings & {
@@ -78,6 +84,7 @@ const DEFAULT_SETTINGS: LayerSettingsUrdf = {
   visible: false,
   frameLocked: true,
   instanceId: "invalid",
+  displayMode: "auto",
 };
 
 const DEFAULT_CUSTOM_SETTINGS: LayerSettingsCustomUrdf = {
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
```diff
@@ -612,7 +633,9 @@ export class Urdfs extends SceneExtension<UrdfRenderable> {
 
   #loadRobot(renderable: UrdfRenderable, { robot, frames, transforms }: ParsedUrdf): void {
     const renderer = this.renderer;
-    const instanceId = renderable.userData.settings.instanceId;
+    const settings = renderable.userData.settings;
+    const instanceId = settings.instanceId;
+    const displayMode = (settings as Partial<LayerSettingsCustomUrdf>).displayMode ?? "auto";
```
jtbandes, line 638: why is the cast needed?

jtbandes (approved):
Simpler code change than I expected 😄 
Now I guess we need a setting for collision color... (does rviz use any color from the urdf for collision elements, or a fixed color or what?)

packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
achim-k: Because I have added the `displayMode` parameter only to the custom layer settings. Will add this to the base settings so this setting is also available for the `/robot_description` URDF and then this cast can be removed.

jtbandes (commented):
Can we add stories for this?

achim-k:
> Now I guess we need a setting for collision color... (does rviz use any color from the urdf for collision elements, or a fixed color or what?)

Ok to do this in a follow up PR?

packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
```diff
@@ -424,8 +459,10 @@ export class Urdfs extends SceneExtension<UrdfRenderable> {
       // ["layers", instanceId, field]
       this.saveSetting(path, action.payload.value);
       const [_layers, instanceId, field] = path as [string, string, string];
-      if (path[1] === PARAM_KEY) {
+      if (instanceId === PARAM_KEY) {
         this.#loadUrdf(instanceId, this.renderer.parameters?.get(PARAM_NAME) as string | undefined);
+      } else if (instanceId === TOPIC_NAME) {
```
jtbandes, line 464: is this a bug fix?

jtbandes (approved):

jtbandes:
With the new story – what is it showing? I can see a cube but it's not necessarily clear whether the behavior is correct or not.

packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
achim-k: No, this is necessary as otherwise the URDF wouldn't be re-rendered since it [returns early here](https://github.com/foxglove/studio/blob/48fa86a1725a16c6fad6721a96bf5b9953645253/packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts#L664-L672) due to the topic-URDF not having a `url`

jtbandes:
Separate could be nice because then the story can have a name that describes what it's showing. Otherwise you could try to use the TF names or at least make it visually clear what a correct rendering should look like

PR #6410, opened 2023-07-10
Improve selection of thin LinePrimitives in the 3D panel

snosenzo:
**User-Facing Changes**

 - Improve selection of thin, scale_invaraint LinePrimitives in the 3D panel

**Description**
I made the picking material for scale_invariant LinePrimitives not go below a MIN_PICKING_WIDTH_PX of 5px. I only did scale_invariant because otherwise they are in world units and can be zoomed in on to be picked, and it would be hard to calculate what the min picking size of a line should be in world units. 

To make this all generic beyond LinePrimitives to all `Lines` I could check materials in `#processItem` in the Picker for `worldUnits = = false && typeof lineWidth == 'number'`, but I wanted to keep it scoped to just the `LinePrimitive` for now to avoid potentially unwanted side affects elsewhere. 

FG-3892

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableLines.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableLines.ts
@@ -88,6 +88,8 @@ export class RenderableLines extends RenderablePrimitive {
   }
 }
 
+const MIN_PICKING_WIDTH_PX = 5;
+
 class LinePrimitiveRenderable extends THREE.Object3D {
   #geometry: LineSegmentsGeometry | LineGeometry | undefined;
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableLines.ts
```diff
@@ -88,6 +88,8 @@ export class RenderableLines extends RenderablePrimitive {
   }
 }
 
+const MIN_PICKING_WIDTH_PX = 5;
```
defunctzombie, line 91: Add a comment - Maybe include `line width` in the name somewhere?

defunctzombie (approved):

jtbandes (approved):

PR #6418, opened 2023-07-11
[3D] Fix selection backdrop glitches

jtbandes:
**User-Facing Changes**

**Description**

The selection overlay had two problems:
- It can completely occlude some items that undergo depth sorting (see cube colors disappearing)
- It can be occluded by some items that are close to the camera (see axes & grid appearing in front of the overlay)

This fixes the issues by using a totally separate Scene for the overlay and disabling its use of the depth buffer, rather than relying on its sort order with other objects in the scene.

https://github.com/foxglove/studio/assets/14237/101da5a6-ec21-4b22-a6c7-6db7a6161d60

</td><td>

https://github.com/foxglove/studio/assets/14237/334fdf1b-6ef9-4e08-8b96-57c074ba9fff

</td></tr></table>


snosenzo (approved):
nice, elegant solution

PR #6420, opened 2023-07-11
Use currentFrame messages in low memory conditions in plot panel

foxymiles:
**User-Facing Changes**
Allow users to seek to reveal segments of messages when memory is exhausted and we can't load any more message blocks.

**Description**
Accumulate ranges of currentFrame messages during playback, allowing the user to reveal portions of the plot even if we don't have enough memory to load blocks from that timerange.

Restores functionality of https://github.com/foxglove/studio/pull/5694, preserving recent performance optimizations from https://github.com/foxglove/studio/pull/6291.

### DIFF
--- a/packages/studio-base/src/panels/Plot/internalTypes.ts
+++ b/packages/studio-base/src/panels/Plot/internalTypes.ts
@@ -46,6 +46,10 @@ export type Datum = ChartDatum & {
 
 export type DataSet = ChartDataset<"scatter", Datum[]>;
 
+// Key datasets by the full PlotPath instead of just the string value because we need to
+// generate a new dataset if the plot path is ordered by headerStamp.
+export type DatasetsByPath = Map<PlotPath, DataSet>;
+
 export type PlotDataItem = {
   queriedData: MessagePathDataItem[];
   receiveTime: Time;
--- a/packages/studio-base/src/panels/Plot/maps.ts
+++ b/packages/studio-base/src/panels/Plot/maps.ts
@@ -1,0 +1,65 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+/**
+ * Merges values from a & b into a new map, applying fn at each key.
+ */
+export function merge<K, V1, V2, V3>(
+  a: ReadonlyMap<K, V1>,
+  b: ReadonlyMap<K, V2>,
+  fn: (aval: V1, bval: V2, key: K) => V3,
+): Map<K, V1 | V2 | V3> {
+  const dest: Map<K, V1 | V2 | V3> = new Map();
+
+  for (const [key, aVal] of a) {
+    const bVal = b.get(key);
+    if (bVal == undefined) {
+      dest.set(key, aVal);
+    } else {
+      dest.set(key, fn(aVal, bVal, key));
+    }
+  }
+
+  for (const [key, bVal] of b) {
+    if (dest.has(key)) {
+      continue;
+    }
+
+    const aVal = a.get(key);
+    if (aVal == undefined) {
+      dest.set(key, bVal);
+    } else {
+      dest.set(key, fn(aVal, bVal, key));
+    }
+  }
+
+  return dest;
+}
+
+/**
+ * Generates a new map from input with the same keys but with fn applied to each value.
+ */
+export function mapValues<K, V, V2>(
+  input: ReadonlyMap<K, V>,
+  fn: (val: V, key: K) => V2,
+): Map<K, V2> {
+  const newEntries: [K, V2][] = [];
+  for (const [key, value] of input) {
+    newEntries.push([key, fn(value, key)]);
+  }
+  return new Map(newEntries);
+}
+
+/**
+ * Returns a new map containing only the keys in keys.
+ */
+export function pick<K, V>(input: ReadonlyMap<K, V>, keys: readonly K[]): Map<K, V> {
+  const newEntries: [K, V][] = [];
+  for (const [key, value] of input) {
+    if (keys.includes(key)) {
+      newEntries.push([key, value]);
+    }
+  }
+  return new Map(newEntries);
+}
--- a/packages/studio-base/src/panels/Plot/messageProcessing.ts
+++ b/packages/studio-base/src/panels/Plot/messageProcessing.ts
@@ -1,0 +1,52 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { Immutable } from "immer";
+
+import {
+  MessageDataItemsByPath,
+  useDecodeMessagePathsForMessagesByTopic,
+} from "@foxglove/studio-base/components/MessagePathSyntax/useCachedGetMessagePathDataItems";
+import { MessageEvent } from "@foxglove/studio-base/players/types";
+import { getTimestampForMessage } from "@foxglove/studio-base/util/time";
+
+import { PlotDataByPath } from "./internalTypes";
+
+type MessagePathDecoder = ReturnType<typeof useDecodeMessagePathsForMessagesByTopic>;
+
+/**
+ * Fetch the data we need from each item in itemsByPath and discard the rest of
+ * the message to save memory.
+ */
+const getByPath = (itemsByPath: MessageDataItemsByPath): PlotDataByPath => {
+  const ret: PlotDataByPath = {};
+  Object.entries(itemsByPath).forEach(([path, items]) => {
+    ret[path] = items.map((messageAndData) => {
+      const headerStamp = getTimestampForMessage(messageAndData.messageEvent.message);
+      return {
+        queriedData: messageAndData.queriedData,
+        receiveTime: messageAndData.messageEvent.receiveTime,
+        headerStamp,
+      };
+    });
+  });
+  return ret;
+};
+
+function getMessagePathItems(
+  decodeMessagePathsForMessagesByTopic: MessagePathDecoder,
+  messages: Immutable<Record<string, MessageEvent[]>>,
+): PlotDataByPath {
+  return Object.freeze(getByPath(decodeMessagePathsForMessagesByTopic(messages)));
+}
+
+/**
+ * Fetch all the plot data we want for our current subscribed topics from blocks.
+ */
+export function getBlockItemsByPath(
+  decodeMessagePathsForMessagesByTopic: MessagePathDecoder,
+  messages: Immutable<Record<string, MessageEvent[]>>,
+): PlotDataByPath {
+  return getMessagePathItems(decodeMessagePathsForMessagesByTopic, messages);
+}
--- a/packages/studio-base/src/panels/Plot/plotData.ts
+++ b/packages/studio-base/src/panels/Plot/plotData.ts
@@ -2,7 +2,7 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
-import { assignWith, isEmpty } from "lodash";
+import { isEmpty } from "lodash";
 import memoizeWeak from "memoize-weak";
 
 import { Time } from "@foxglove/rostime";
@@ -13,23 +13,24 @@ import { Range } from "@foxglove/studio-base/util/ranges";
 
 import {
   BasePlotPath,
-  DataSet,
+  DatasetsByPath,
   Datum,
   PlotDataByPath,
   PlotPath,
   PlotXAxisVal,
   isReferenceLinePlotPathType,
 } from "./internalTypes";
+import * as maps from "./maps";
 
 export type PlotData = {
   bounds: Bounds;
-  datasetsByPath: Record<string, DataSet>;
+  datasetsByPath: DatasetsByPath;
   pathsWithMismatchedDataLengths: string[];
 };
 
 export const EmptyPlotData: Im<PlotData> = Object.freeze({
   bounds: makeInvertedBounds(),
-  datasetsByPath: {},
+  datasetsByPath: new Map(),
   pathsWithMismatchedDataLengths: [],
 });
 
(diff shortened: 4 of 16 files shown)


packages/studio-base/src/panels/Plot/maps.ts
```diff
@@ -0,0 +1,65 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+/**
+ * Merges values from a & b into a new map, applying fn at each key.
+ */
+export function merge<K, V1, V2, V3>(
```
defunctzombie, line 8: Stuff like this feels like it belongs in `den` cause it looks generic and not specific to plot panel.

foxymiles: Eventually we should migrate this to den if we think it's useful in more than one place but I didn't want to hold up this PR on that.

packages/studio-base/src/panels/Plot/usePlotPanelData.ts
```diff
@@ -355,43 +333,67 @@ export function usePlotPanelDatasets(params: Params): {
...
+  const currentFrameData = useMessageReducer<TaggedPlotData>({
     topics: subscribeTopics,
     preloadType: "full",
     restore,
     addMessages,
   });
 
+  // Accumulate separate message playback sequences into distinct intervals we
+  // can later combine into a single combined set of message data.
+  const [accumulatedPathIntervals, setAccumulatedPathIntervals] = useState<
```
defunctzombie, line 345: Aren't you already accumulating in the `useMessageReducer` logic? Why this separate accumulation?

packages/studio-base/src/panels/Plot/usePlotPanelData.ts
```diff
@@ -355,43 +333,67 @@ export function usePlotPanelDatasets(params: Params): {
...
-    const sortedCurrentFrameWithDerivatives = sortDatasetsByHeaderStamp(
-      currentFrameWithDerivatives,
-    );
-    const allDatasets = Object.values(sortedStateWithDerivatives).concat(
-      Object.values(sortedCurrentFrameWithDerivatives),
-    );
-    const bounds = unionBounds(state.bounds, currentFrameDatasets.bounds);
+  const allData = useMemo(() => {
+    const combinedPlotData = reducePlotData([state.data, ...trimmedCurrentFrameData]);
+    const dataAfterDerivative = applyDerivativeToPlotData(combinedPlotData);
```
defunctzombie, line 388: Isn't it possible to apply derivative on a running set of points?

packages/studio-base/src/panels/Plot/usePlotPanelData.ts
```diff
@@ -355,43 +333,67 @@ export function usePlotPanelDatasets(params: Params): {
...
+      setAccumulatedPathIntervals((oldValue) => ({
+        ...oldValue,
+        [currentFrameData.tag]: currentFrameData.data,
+      }));
+    }
+  }, [currentFrameData.data, currentFrameData.tag]);
+
+  // Trim currentFrame data outside allFrames, assuming allFrames is contiguous from start
+  // time.
+  const trimmedCurrentFrameData = useMemo(() => {
```
defunctzombie, line 360: Why do you need to do this? earlier it looked like you already ignore currentFrame data that overlaps allFrames. The comment about trillimg data outside allFrames doesn't make sense to me.

packages/studio-base/src/panels/Plot/usePlotPanelData.ts
```diff
@@ -355,43 +333,67 @@ export function usePlotPanelDatasets(params: Params): {
...
+  >({});
+
+  useEffect(() => {
+    if (!isEmpty(currentFrameData.data.datasetsByPath)) {
+      setAccumulatedPathIntervals((oldValue) => ({
+        ...oldValue,
+        [currentFrameData.tag]: currentFrameData.data,
+      }));
+    }
+  }, [currentFrameData.data, currentFrameData.tag]);
```
defunctzombie, line 356: This effect seems out of place to me cause I'd have expected there to be memo's on the dataset merging layer rather than effects.

packages/studio-base/src/panels/Plot/useAllFramesByTopic.ts
```diff
@@ -89,12 +113,10 @@ export function useFlattenedBlocksByTopic(
...
-        for (const topic of topics) {
-          const blockMessages = block.messagesByTopic[topic];
-          if (blockMessages == undefined) {
-            continue;
-          }
-
+        // There is a delay between the time we set new subscriptions and the messages for
+        // those subscriptions appear in blocks so we load all topics we find in blocks
+        // here.
+        for (const [topic, blockMessages] of Object.entries(block.messagesByTopic)) {
```
defunctzombie, line 119: I don't understand this comment and logic. Why would we want to load all the topics in a block if those topics are not relevant to our plot panel?

packages/studio-base/src/panels/Plot/messageProcessing.ts
```diff
@@ -0,0 +1,52 @@
...
+} from "@foxglove/studio-base/components/MessagePathSyntax/useCachedGetMessagePathDataItems";
+import { MessageEvent } from "@foxglove/studio-base/players/types";
+import { getTimestampForMessage } from "@foxglove/studio-base/util/time";
+
+import { PlotDataByPath } from "./internalTypes";
+
+type MessagePathDecoder = ReturnType<typeof useDecodeMessagePathsForMessagesByTopic>;
+
+/**
+ * Fetch the data we need from each item in itemsByPath and discard the rest of
```
defunctzombie, line 19: How does this save memory? In live workflows? Cause in workflows where you pre-load data the message instance is already using memory.

packages/studio-base/src/panels/Plot/messageProcessing.ts
```diff
@@ -0,0 +1,52 @@
...
+  decodeMessagePathsForMessagesByTopic: MessagePathDecoder,
+  messages: Immutable<Record<string, MessageEvent[]>>,
+): PlotDataByPath {
+  return Object.freeze(getByPath(decodeMessagePathsForMessagesByTopic(messages)));
+}
+
+/**
+ * Fetch all the plot data we want for our current subscribed topics from blocks.
+ */
+export function getBlockItemsByPath(
```
defunctzombie, line 47: My take is that these single line wrappers don't add value - to me they are worse cause they add indirection. The call site could invoke the functions directly. Looking at where you call `getBlockItemsByPath` (in usePlotPanelData.ts) I think it would be fine to change that call site to invoke `getByPath(decodeMessagePathsForMessagesByTopic(messages))` directly and there'd be less indirection and function calls happening with the same result.

packages/studio-base/src/panels/Plot/messageProcessing.ts
```diff
@@ -0,0 +1,52 @@
...
+
+type MessagePathDecoder = ReturnType<typeof useDecodeMessagePathsForMessagesByTopic>;
+
+/**
+ * Fetch the data we need from each item in itemsByPath and discard the rest of
+ * the message to save memory.
+ */
+const getByPath = (itemsByPath: MessageDataItemsByPath): PlotDataByPath => {
+  const ret: PlotDataByPath = {};
+  Object.entries(itemsByPath).forEach(([path, items]) => {
```
defunctzombie, line 24: Probably a copy from code that existed but I would change the `forEach` to a regular `for` loop. One less closure and (my opinion easier) to read.

packages/studio-base/src/panels/Plot/messageProcessing.ts
```diff
@@ -0,0 +1,52 @@
...
+    ret[path] = items.map((messageAndData) => {
```
defunctzombie, line 25: This is making a closure in a tight loop. This closure only uses the array item so you should hoist the mapping function out so each iteration doesn't make a new function.

packages/studio-base/src/panels/Plot/plotData.ts
```diff
@@ -2,47 +2,200 @@
...
+}
 
-function getMessagePathItems(
-  decodeMessagePathsForMessagesByTopic: (
-    Record: Record<string, readonly MessageEvent[]>,
-  ) => MessageDataItemsByPath,
-  messages: Record<string, readonly MessageEvent[]>,
-): PlotDataByPath {
-  return Object.freeze(getByPath(decodeMessagePathsForMessagesByTopic(messages)));
+const memoFindXRanges = memoizeWeak(findXRanges);
```
defunctzombie, line 120: Please don't memoize this here - this is going to hang onto plot data references with no lifetime management. We already have layers of memoization in react for when we calculate ranges of things - why are we not able to leverage those?

packages/studio-base/src/panels/Plot/plotData.ts
```diff
@@ -2,47 +2,200 @@
...
+    }),
+  };
+}
+
+/**
+ * Merge two PlotData objects into a single PlotData object, discarding any overlapping
+ * messages between the two items. Assumes they represent non-contiguous segments of a
+ * chart.
+ */
+function mergePlotData(a: Im<PlotData>, b: Im<PlotData>): Im<PlotData> {
```
defunctzombie, line 92: Why do we need to do this? If we have cursors and also ignore currentFrame data that overlaps allFrames?

packages/studio-base/src/panels/Plot/plotData.ts
```diff
@@ -2,47 +2,200 @@
...
 
-import { MessageEvent } from "@foxglove/studio";
-import { MessageDataItemsByPath } from "@foxglove/studio-base/components/MessagePathSyntax/useCachedGetMessagePathDataItems";
-import { PlotDataByPath } from "@foxglove/studio-base/panels/Plot/internalTypes";
-import { getTimestampForMessage } from "@foxglove/studio-base/util/time";
+import { isEmpty } from "lodash";
+import memoizeWeak from "memoize-weak";
+
+import { Time } from "@foxglove/rostime";
+import { Immutable as Im } from "@foxglove/studio";
```
defunctzombie, line 9: I'm not particularly a fan of the inconsistent use of `Immutable` and `Im` in other places. Seems you used `Immutable` in other parts of this PR so why different here? Better to be consistent.

packages/studio-base/src/panels/Plot/plotData.ts
```diff
@@ -2,47 +2,200 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
```
defunctzombie, line 4: I dont follow the purpose of this file. It has no tests of its own and seems to be used only by `usePlotPanelData`? Should all this code live there? How would I determine why something goes into plotData.ts or messageProcessing.ts? Not obvious to me how these two files help organization or maintenance.

defunctzombie (approved):
Immutability is slowly kreeping its way in. Good stuff.

I confirmed that preloading a large buffer that fills up and playing data allows me to plot the new data even though preloading has filled. I did not notice a negative impact on loading a large dataset compared to main.

There is a noticeable delay when I have a large dataset and click the "new series" button to add an empty series. It would be nice to root that out and feels like memos should be accomplishing this but are not.

I think most of my comments are touchups or small moves/adjustments. Addressing those and merging this is the right move to resolve the known regression; however I also have an architectural read here that makes me think the number of dataset iterations, mangling, concats, etc could be vastly improved by consolidating the "currentFrame" flow (the reducer stuff) with the blocks flow rather than treating them separately. I think it will lead to fewer passes over the data and easier to reason logic - but it requires trying it out. So I would favor merging the feedback incorporated version of this and then take a spike to organize the data flows better.

packages/studio-base/src/panels/Plot/usePlotPanelData.ts
```diff
@@ -179,56 +163,35 @@ export function usePlotPanelDatasets(params: Params): {
     setState((oldState) => {
```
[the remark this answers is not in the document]
foxymiles: No it's recommended in some cases:

https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes

It's cheaper than doing a full render and then immediately triggering another render because of a `useEffect` that changes the state.

packages/studio-base/src/panels/Plot/usePlotPanelData.ts
foxymiles: We need to accumulate data across seeks. Each time you seek the reducer wipes out its own buffer.

This is logic I've just resurrected from the pre-optimization branch.

packages/studio-base/src/panels/Plot/plotData.ts
foxymiles: It's a weak memo. It won't hang onto anything.

packages/studio-base/src/panels/Plot/plotData.ts
foxymiles: PlotData is a type that bundles multiple datasets with their bounds and mismatched paths array.

packages/studio-base/src/panels/Plot/internalTypes.ts
```diff
@@ -46,6 +46,10 @@ export type Datum = ChartDatum & {
 
 export type DataSet = ChartDataset<"scatter", Datum[]>;
 
+// Key datasets by the full PlotPath instead of just the string value because we need to
+// generate a new dataset if the plot path is ordered by headerStamp.
+export type DatasetsByPath = Map<PlotPath, DataSet>;
```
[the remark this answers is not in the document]
foxymiles: When we pull data from the messages we get different results depending on the timestampMethod:

https://github.com/foxglove/studio/blob/0ede85438aff8b143c45775c3aba152e3165b68a/packages/studio-base/src/panels/Plot/datasets.tsx#L59

So the string path alone does not determine the dataset, it's a [path, timestampMethod] tuple. So we can't key datasets simply by path alone. On main I store these in an array but this means we have to be very careful to keep the paths and the datasets in the same order everywhere which seems asking for bugs to me. In fact I have an open ticket for an issue related to this on main right now.

So what I'm trying here is using a map and keying by the full object. I think this is going to be much less error prone.

packages/studio-base/src/panels/Plot/plotData.ts
foxymiles: There is only one case where we need to be careful with memoizeWeak and that's with using primitive arguments:
https://github.com/timkendrick/memoize-weak/issues/5

I think we should add a library like moize or memoizee to our toolkit for those cases.

PR #6421, opened 2023-07-11
Respect the show lines setting for each dataset

defunctzombie:
**User-Facing Changes**
When using `[:]` message path syntax to plot points from the current message, the `showLines` setting is respected.

**Description**

### DIFF
--- a/packages/studio-base/src/panels/Plot/datasets.tsx
+++ b/packages/studio-base/src/panels/Plot/datasets.tsx
@@ -126,7 +126,7 @@ function getDatasetsFromMessagePlotPath({
   dataset: DataSet;
   hasMismatchedData: boolean;
 } {
-  let showLine = path.showLine !== false;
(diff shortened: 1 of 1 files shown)


defunctzombie:
@foxymiles @jtbandes Any thoughts on why we forced `showLines` to false when using the array form of message path input for plots? One thought is that it could result in lines all over the place if your data has bad ordering? Though maybe we leave that up to the user since we have the setting?

foxymiles:
@defunctzombie We'd have to figure out how each point in one array slice maps to points in the subsequent array slice right?

defunctzombie:
> @defunctzombie We'd have to figure out how each point in one array slice maps to points in the subsequent array slice right?

Maybe I'm thinking about a different mode but I had thought there was a mode or path syntax where you have a scatter plot from a single message.

PR #6429, opened 2023-07-12
Remove SignInForm

defunctzombie:
**User-Facing Changes**

**Description**


packages/studio-base/src/Workspace.tsx
```diff
@@ -418,15 +413,55 @@ function WorkspaceContent(props: WorkspaceContentProps): JSX.Element {
     [getMessagePipeline],
   );
 
+  const targetUrlState = useMemo(() => {
```
foxymiles, line 416: This feels like it could be broken out into a sub-hook since Workspace is already such a bloated component.

foxymiles (approved):

packages/studio-base/src/Workspace.tsx
defunctzombie: This particular memo? Is there much value in that for this one memo?

foxymiles: The whole url state handling block.

defunctzombie: I'll think about it some more and I continue the quest around workspace - I'm kinda iffy on this "don't return anything hooks" pattern but I can see some value in it for closure isolation and readability - its kinda like a private function but the strangeness is I would not put it in the common hooks folder like we had done before cause that makes it look like it should be used in other places which is not valid.

My real play here is to move the "app state url" to a pattern like `useQueryParams` and have it handled in the respective providers.

foxymiles: I think there's value in using custom hooks strictly as a code organization tool. These 500+ line components are hard to work with.

Agree that longer term `useQueryParams` or something similar seems like a good path.

PR #6430, opened 2023-07-12
Add picking story for image annotations

jtbandes:
**User-Facing Changes**
None

**Description**
Adds a story showing the ImageAnnotations hitmap.

image

I also updated default picking material implementation to make distinct materials based on `depthTest` and `depthWrite`. I did this because the ImageRenderable uses these material flags to always render behind the annotations (see `renderBehindScene`). Without this change, the image would overlap some annotations in the hitmap even though it did not actually overlap them in a normal render. The multi-pass picking would be able to handle this and correctly show both the image and annotation in the dropdown menu, but I disabled multi-pass picking when `debugPicking==true` in #6419.

Blocked on #6419

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/Picker.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/Picker.ts
@@ -285,14 +285,13 @@ export class Picker {
       return;
     }
 
-    const sprite = material.type === "SpriteMaterial" ? 1 : 0;
+    const isSprite = material.type === "SpriteMaterial";
 
     const pickResolution = tempResolution.set(
       this.#pickingTarget.width,
       this.#pickingTarget.height,
     );
-    const sizeAttenuation =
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/panels/ThreeDeeRender/Picker.ts
```diff
@@ -344,18 +350,33 @@ export class Picker {
...
+    sizeAttenuation,
+    depthTest,
+    depthWrite,
+  }: {
+    isSprite: boolean;
+    sizeAttenuation: boolean;
+    depthTest: boolean;
+    depthWrite: boolean;
+  }): THREE.ShaderMaterial {
+    const index =
```
snosenzo, line 365: Why don't we just set the parameters and reuse the same material? It doesn't look like these new parameters affect shader compilation. Maybe a comment here clarifying why this is necessary

snosenzo (approved):

PR #6432, opened 2023-07-12
Consistent search box styling

2metres:
**User-Facing Changes**
None

**Description**
Consistent search box styling

Resolves FG-2219 and FG-4217


jtbandes (approved):
Wanna fix this bug while you're at it? https://linear.app/foxglove/issue/FG-4217/duplicate-form-field-id-in-the-same-form

Issue #5862, opened 2023-04-27
Incorrect visualization when plotting data with the `[:]` accessor


vhwanger:
**Description**
I have data of the form:

```
  1 syntax = "proto3";
  2
  3
  4 message ThreeTuple {
  5     float a = 1;
  6     float b = 2;
  7     float c = 3;
  8 }
  9 message DataCollection {
 10     repeated ThreeTuple data = 1;
 11 }
```

When I try and plot `/data.data[:].a` with the X-axis value using `Index` and I scrub around, the plot seemingly starts accumulating all the lines over the times I scrub over. The expected behavior is that it only plots the lines over the current timestamp of the log

![image](https://user-images.githubusercontent.com/2358630/235008675-eb3bf54f-56eb-4255-b288-78079e711faa.png)

- Version: 1.51.0
- Platform: Mac
- Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native):
mcap file

**Expected Behavior**
The expected behavior is shown below.
<img width="767" alt="Screenshot 2023-04-27 at 4 02 02 PM" src="https://user-images.githubusercontent.com/2358630/235008736-a005fd2d-2a4c-43e9-8da6-1dddaf0e7055.png">

vhwanger:
I have a layout and an mcap available for reproducing if needed.

foxymiles:
Thanks for the detailed bug report. I'm seeing some strange behavior too but a bit different from what you describe. Does what you're seeing resemble this?

https://user-images.githubusercontent.com/93935560/235024939-8d47b596-1586-45b7-b9ea-52455ca3a1bf.mov

vhwanger:
That's actually what the data looks like. Now if you scrub back and forth in the window, does it look the same?

foxymiles:
Yes I see what you mean. Scrubbing on the timeline back and forth results in these lines accumulating like this.

foxymiles:
@vhwanger Could you please try the deployment from this PR: https://github.com/foxglove/studio/pull/5868 and see if it fixes your problem? You can try it here: https://studio-8uik5ytns.foxglove.party/

vhwanger:
This looks like it fixes it on my fake data file, thank you so much! I'll have to wait for the desktop deployment to try it on the actual log I have.

foxymiles:
Looks like I inadvertently broke this as part of recent optimization work.

PR #6435, opened 2023-07-12
Fix handling of index-based x axis plots

foxymiles:
**User-Facing Changes**
Fixes an issue with x-axis index based plots.

**Description**
Fixes a regression introduced in https://github.com/foxglove/studio/pull/6291.

If we are displaying one message at a time we shouldn't append to previous accumulated datasets.

Fixes https://github.com/foxglove/studio/issues/5862


defunctzombie (commented):
I think it would be nice to write some non-visual stories for plot data. We tend to not realize when visual stories break since plot stories are often flaky.

foxymiles:
@defunctzombie Agreed although fixing those plot stories for good is definitely on my todo list.

PR #6442, opened 2023-07-13
Add support for `omgidl` schemas in foxglove studio

snosenzo:
**User-Facing Changes**

- add support for `omgidl`-encoded schemas for messages using `cdr` message encoding in foxglove studio

**Description**
Add support for cdr message encoding alongside omgidl schemas. 

FG-4304

### DIFF
--- a/packages/mcap-support/src/parseChannel.ts
+++ b/packages/mcap-support/src/parseChannel.ts
@@ -123,7 +123,7 @@ export function parseChannel(channel: Channel): ParsedChannel {
     };
   }
 
-  if (channel.messageEncoding === "cdr" || channel.messageEncoding === "xcdr2") {
+  if (channel.messageEncoding === "cdr") {
     if (
       channel.schema?.encoding !== "ros2msg" &&
       channel.schema?.encoding !== "ros2idl" &&
--- a/packages/mcap-support/src/parseChannel.ts
+++ b/packages/mcap-support/src/parseChannel.ts
@@ -3,6 +3,8 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
 import { MessageDefinition } from "@foxglove/message-definition";
+import { parseIdl } from "@foxglove/omgidl-parser";
+import { MessageReader as OmgidlMessageReader } from "@foxglove/omgidl-serialization";
 import { parse as parseMessageDefinition, parseRos2idl } from "@foxglove/rosmsg";
 import { MessageReader } from "@foxglove/rosmsg-serialization";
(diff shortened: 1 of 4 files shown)


packages/mcap-support/src/parseChannel.ts
```diff
@@ -132,17 +138,28 @@ export function parseChannel(channel: Channel): ParsedChannel {
...
-    const isIdl = channel.schema.encoding === "ros2idl";
+    if (channel.schema.encoding === "omgidl") {
+      const parsedDefinitions = parseIdl(schema);
+      const reader = new OmgidlMessageReader(channel.schema.name, parsedDefinitions);
+      const datatypes = parsedDefinitionsToDatatypes(parsedDefinitions);
+      return {
+        datatypes,
+        deserialize: (data) => reader.readMessage(data),
+      };
+    } else {
```
snosenzo, line 149: should ros2idl support xdr2 encoding?

defunctzombie: What do the ROS2 docs say? I'm not aware this is a thing in ROS - so if it isn't - then no.

defunctzombie (approved):
I wouldn't block merging this on the documentation. We'll need it but there's also time before the next release or whenever we go more public with this feature.

packages/mcap-support/src/parseChannel.ts
```diff
@@ -121,8 +123,12 @@ export function parseChannel(channel: Channel): ParsedChannel {
     };
   }
 
-  if (channel.messageEncoding === "cdr") {
-    if (channel.schema?.encoding !== "ros2msg" && channel.schema?.encoding !== "ros2idl") {
+  if (channel.messageEncoding === "cdr" || channel.messageEncoding === "xcdr2") {
```
jtbandes, line 126: We don't support `xcdr2` message encoding according to https://mcap.dev/spec/registry

defunctzombie: @snosenzo would it be correct to say `cdr` is the message encoding since `xcdr2` is implemented as `cdr` or do we want to be explicit that someone tell us they specifically used `xcdr2` vs `cdr`? If we do expect it is important the message encoding says one vs the other we should add it to the well known encodings.

snosenzo: Thanks for catching! Yeah CDR is the message encoding and are able to pick up on whether it's xcdr2 based off of the encapsulation type and (in the future) the extensibile type of the schema (ie @mutable etc.). Will remove this

PR #6445, opened 2023-07-14
Fix & update yuyv/uyvy encodings

### LINKED ISSUE #6271: Is this a mistake? there is no uyuy, does it mean uyvy?
**Description**
image
Is this a mistake? there is no uyuy, does it mean uyvy?

- Version:
- Platform:

**Steps To Reproduce**

**Expected Behavior**
uyvy

jtbandes:
**User-Facing Changes**
- Renamed raw image encoding `uyuv` to `uyvy` for accuracy.
- Added support for raw image encoding `yuv422_yuy2` as an alias for `yuyv`.

**Description**
- Fixes #6271, fixes FG-3953
- Schemas PR: https://github.com/foxglove/schemas/pull/121

All [YUV 4:2:2 formats](https://en.wikipedia.org/wiki/Chroma_subsampling) encode one U and one V value for every two Y values. The byte orders supported by Studio and by other apps in the ecosystem are `[U, Y1, V, Y2]` and `[Y1, U, Y2, V]`.

- Replaces `uyuv`, which seems to have been a typo introduced in https://github.com/foxglove/studio/pull/5292, with `uyvy`, which accurately represents the byte order we are decoding.
- Adds `yuv422_yuy2` as an encoding alias for `yuyv`, taken from ROS:
  - [sensor_msgs/image_encodings.hpp](https://github.com/ros2/common_interfaces/blob/366eea24ffce6c87f8860cbcd27f4863f46ad822/sensor_msgs/include/sensor_msgs/image_encodings.hpp)
  - https://github.com/ros2/common_interfaces/issues/76
  - https://github.com/ros2/common_interfaces/pull/214
  - `yuv422_yuy2` is the default encoding used by [usb_cam](https://github.com/ros-drivers/usb_cam/tree/ros2)
- Remove incorrect and useless `Int8Array` type casting
- Update YUV-to-RGB conversion formula to be faster and match RViz code
- Added story for both UYVY and YUYV
- Confirmed compatibility with rqt_image_view and rviz (although those two look slightly different from each other): 
  Screenshot 2023-07-13 at 5 02 16 PM

Other references:
- https://www.kernel.org/doc/html/latest/userspace-api/media/v4l/pixfmt-packed-yuv.html

### DIFF
--- a/packages/den/image/decodings.ts
+++ b/packages/den/image/decodings.ts
@@ -11,6 +11,10 @@
 //   found at http://www.apache.org/licenses/LICENSE-2.0
 //   You may not use this file except in compliance with the License.
 
+/**
+ * Adapted from:
+ * https://github.com/ros2/rviz/blob/e8838720e57b56cc0d50e05b00b28bee3c5dc9ee/rviz_default_plugins/src/rviz_default_plugins/displays/image/ros_image_texture.cpp#L332
+ */
 function yuvToRGBA8(
   y1: number,
   u: number,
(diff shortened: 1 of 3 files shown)


defunctzombie (approved):
Are we expecting any reports of folks that had recorded data with `uyuv`? Or is this unlikely as it is not a real format?

jtbandes:
Seems unlikely to me. `uyuv` was only added as an alias to the existing (and more common) `yuv422`, and only 5 months ago.

sgf201:
the YUV422P is useful, @jtbandes ,could you add the 420P and 422p support

PR #6446, opened 2023-07-14
Fix ordering of values in plot legend with disabled series

foxymiles:
**User-Facing Changes**
Fix ordering of values in plot legend with disabled series.

**Description**
Datasets sent to the plot must have the same order and count as the plot series in order for the legend to display the right values alongside the series path/label.

### DIFF
--- a/packages/studio-base/src/panels/Plot/usePlotPanelData.ts
+++ b/packages/studio-base/src/panels/Plot/usePlotPanelData.ts
@@ -392,7 +392,15 @@ export function usePlotPanelData(params: Params): Immutable<{
 
     return {
       bounds: sortedData.bounds,
-      datasets: filterMap(yAxisPaths, (path) => sortedData.datasetsByPath.get(path)),
+      // Return a dataset for all paths here so that the ordering of datasets corresponds
+      // to yAxisPaths as expected by downstream components like the legend.
+      //
+      // Label is needed so that TimeBasedChart doesn't discard the empty dataset and mess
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/panels/Plot/usePlotPanelData.ts
```diff
@@ -392,7 +392,15 @@ export function usePlotPanelData(params: Params): Immutable<{
 
     return {
       bounds: sortedData.bounds,
-      datasets: filterMap(yAxisPaths, (path) => sortedData.datasetsByPath.get(path)),
+      // Return a dataset for all paths here so that the ordering of datasets corresponds
+      // to yAxisPaths as expected by downstream components like the legend.
+      //
+      // Label is needed so that TimeBasedChart doesn't discard the empty dataset and mess
+      // up the ordering.
+      datasets: yAxisPaths.map(
```
defunctzombie, line 400: Seems like this logic to keep the empty dataset should be in the dataset building code and then we don't have to iterate on these yet again?

defunctzombie (approved):

packages/studio-base/src/panels/Plot/usePlotPanelData.ts
foxymiles: I actually originally had it there but moved it here since it seemed like the concerns here are closer to the plot panel and not the datasets themselves. We're talking on the order of 10s of datasets so the iteration shouldn't be costly.

PR #6449, opened 2023-07-14
Get typescript enums working in user script code

foxymiles:
**User-Facing Changes**
None

**Description**
Generate datatypes from typescript enums in user script code by treating them as uint32 types.

This of course doesn't cover all types of typescript enums but should enough to support the enums we use in @foxglove/schemas.


defunctzombie (approved):

foxymiles:
This was subsumed into https://github.com/foxglove/studio/pull/6376

PR #6455, opened 2023-07-14
Improve camera distortion and add support for rational_polynomial

jtbandes:
**User-Facing Changes**
- Fixed some correctness issues with support for camera distortion correction.
- Camera distortion correction now supports the `rational_polynomial` model with calibration parameters $k_4$, $k_5$, $k_6$.

**Description**
Thanks to @saching13 for working on this.
Closes #6243 
Resolves FG-3732

- Adds support for $k_4$, $k_5$, and $k_6$ parameters in distortion and undistortion, with code adapted from OpenCV
- Corrected the usage of $f_x,f_y,c_x,c_y$ (from $K$) vs $f_x',f_y',c_x',c_y'$ (from $K'$, i.e. the first 3 columns of $P$)
- Undistortion is now done in `projectPixelTo3dPlane`, so can be removed from the `projectPixel()` free function
- Fixed $R$ being applied instead of $R^{-1}$ in `distortPixel` (previously `rectifyPoint`), which seems to be a mistake [inherited from webviz](https://github.com/cruise-automation/webviz/blob/d13afdacaec0b48f983adcaca55b84c36c42a5f2/packages/webviz-core/src/panels/ImageView/CameraModel.js#L96-L99) when the code was initially ported from ROS.
- $t_x$ and $t_y$ are no longer used as we believe these are not necessary for this functionality (although we may be proven wrong someday if someone can provide good test data from a stereo camera system)


packages/den/image/PinholeCameraModel.ts
```diff
@@ -131,50 +133,105 @@ export class PinholeCameraModel {
...
-   * Projects a 2D image pixel to a point on a plane in 3D world coordinates a
-   * unit distance along the Z axis. This is equivalent to `projectPixelTo3dRay`
-   * before normalizing.
+   * Undoes camera distortion to map a given pixel coordinate from normalized raw image coordinates to normalized undistorted coordinates.
    *
-   * @param out - The output vector to receive the 3D point.
-   * @param pixel - The 2D image pixel coordinate.
-   * @returns `true` if the projection was successful, or `false` if the camera
-   *   projection matrix `P` is not set.
+   * @param out - The output rectified 2D normalized coordinate.
```
snosenzo, line 138: I think the old comment is more clear to me. I like the `to receive` word choice.
```suggestion
   * @param out - The output vector to receive the rectified 2D normalized coordinate.
```

packages/den/image/PinholeCameraModel.ts
```diff
@@ -131,50 +133,105 @@ export class PinholeCameraModel {
...
+   * @param point - The input distorted/unrectified 2D normalized coordinate.
+   * @param iterations - The number of iterations to use in the iterative optimization.
```
snosenzo, line 140: This comment doesn't tell a caller of the function  how to use this effectively without already knowing what `iterative optimization` is. Would be nice if I didn't have to look at the function to know. 
also should this comment also say that it's optional?

packages/den/image/PinholeCameraModel.ts
```diff
@@ -131,50 +133,105 @@ export class PinholeCameraModel {
...
-   * unit distance along the Z axis. This is equivalent to `projectPixelTo3dRay`
-   * before normalizing.
+   * Undoes camera distortion to map a given pixel coordinate from normalized raw image coordinates to normalized undistorted coordinates.
    *
-   * @param out - The output vector to receive the 3D point.
-   * @param pixel - The 2D image pixel coordinate.
-   * @returns `true` if the projection was successful, or `false` if the camera
-   *   projection matrix `P` is not set.
+   * @param out - The output rectified 2D normalized coordinate.
+   * @param point - The input distorted/unrectified 2D normalized coordinate.
```
snosenzo, line 139: Didn't we just go through a rename to change `rectify` to `undistort`? Will referring to this param as `distorted/unrectified` be confusing? or do they actually mean the same thing?

packages/den/image/PinholeCameraModel.ts
```diff
@@ -131,50 +133,105 @@ export class PinholeCameraModel {
...
+    out.y = y;
     return out;
   }
 
   /**
-   * Projects a 2D image pixel into a 3D ray in world coordinates. This is
-   * equivalent to normalizing the result of `projectPixelTo3dPlane` to get a
-   * direction vector.
+   * Applies camera distortion parameters to map a given pixel coordinate from normalized
+   * undistorted image coordinates to normalized raw coordinates.
```
snosenzo, line 196: specify normalized coordinates?

packages/den/image/PinholeCameraModel.ts
```diff
@@ -131,50 +133,105 @@ export class PinholeCameraModel {
   }
 
   /**
-   * Projects a 2D image pixel to a point on a plane in 3D world coordinates a
-   * unit distance along the Z axis. This is equivalent to `projectPixelTo3dRay`
-   * before normalizing.
+   * Undoes camera distortion to map a given pixel coordinate from normalized raw image coordinates to normalized undistorted coordinates.
```
snosenzo, line 136: specify `normalized coordinates` for all these comments? Could mean [-1, 1] or [0, 1] but these could also be outside of this range if they're distorted right? Or are these normalized coordinates based off of the distorted image? 😵‍💫

snosenzo (commented):
suggested clarity and wording changes, otherwise looks good to me

Is there a difference between `rectify` and `undistort`? should we keep their usage consistant?

packages/den/image/PinholeCameraModel.ts
jtbandes: I think optional is shown in the type signature so it's fine not to say it in words

packages/den/image/PinholeCameraModel.ts
```diff
@@ -254,51 +281,78 @@ export class PinholeCameraModel {
...
-    const ty = P[7];
+    const { P, K } = this;
+    const fx = K[0];
+    const fy = K[4];
+    const cx = K[2];
+    const cy = K[5];
 
-    // Formulae from:
-    // <https://docs.opencv.org/2.4/modules/calib3d/doc/camera_calibration_and_3d_reconstruction.html>
+    const fxp = P[0];
```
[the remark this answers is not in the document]
jtbandes: I guess [these docs](https://docs.ros.org/en/api/sensor_msgs/html/msg/CameraInfo.html) call them `fx'` rather than `f'x` but I can see an argument for both. Certainly LaTeX renders $f'_x$ (`$f'_x$`) and $f_x'$ (`$f_x'$`) the same :)

packages/den/image/PinholeCameraModel.ts
jtbandes: I'm still learning but I believe they are technically different things (distortion being more about camera intrinsics and lens distortion, and rectification being more about aligning stereo systems). However they seem to have been used interchangeably (or at least together) in some places. I'll try to skew these more towards only talking about "distort" since that's mostly what we've been testing with

snosenzo (approved):

PR #6461, opened 2023-07-17
Remove layoutId and org from AppURLState

defunctzombie:
**User-Facing Changes**
None

**Description**
These are no longer used in the app.


jtbandes (approved):
🎉

wimagguc (approved):

PR #6462, opened 2023-07-18
Create shared component for HighlightChars

2metres:
**User-Facing Changes**
None

**Description**
Now that we use this in multiple places it makes sense to have a single component

### DIFF
--- a/packages/studio-base/src/components/HighlightChars.tsx
+++ b/packages/studio-base/src/components/HighlightChars.tsx
@@ -12,13 +12,11 @@ type Props = {
 };
 
 const useStyles = makeStyles()({
-  root: {
-    whiteSpace: "pre",
-  },
+  root: { whiteSpace: "pre" },
 });
 
 /**
- * Renders the given text with the characters at the given indices wrapped in a
+ * Renders the given text with the characters highlighted text wrapped in a
  * <mark> component for Fzf results. The indices are the positions of the
  * matched characters in the original string.
  *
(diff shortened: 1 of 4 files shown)


2metres:
@foxymiles do we think we could refactor [components/HighlightedText.tsx](https://github.com/foxglove/studio/blob/main/packages/studio-base/src/components/HighlightedText.tsx) to also use this?

packages/studio-base/src/components/HighlightChars.tsx
```diff
@@ -0,0 +1,39 @@
...
+  const { str, indices, offset = 0 } = props;
+  const { classes } = useStyles();
+  const chars = str.split("");
+
+  const nodes = chars.map((char, i) => {
+    if (indices.has(i + offset)) {
+      return <mark key={i}>{char}</mark>;
+    }
+    return char;
+  });
```
jtbandes, line 36: Maybe `useMemo` this? Not sure but it feels like it could be a significant perf win.

packages/studio-base/src/components/HighlightChars.tsx
```diff
@@ -0,0 +1,39 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { makeStyles } from "tss-react/mui";
+
+// Renders the given text with the characters at the given indices wrapped in a
```
jtbandes, line 7: Please put this as a `/** ... */` comment on the HighlightChars component itself. Then VS Code will pick up the comment and show it when hovering on the component name from anywhere in the project

foxymiles (approved):
Makes sense. Agree with @jtbandes that we probably should `useMemo` the nodes.

defunctzombie:
FWIW I was confused by the PR description and the code changes I saw. It looks like we already had this component and this PR extracts it from Autocomplete.

PR #6463, opened 2023-07-18
Add extra capability check when advertising channels on foxglove websocket

achim-k:
**User-Facing Changes**
None

**Description**
Adds an extra check for the `advertise` capability when advertising channels for the foxglove websocket data source.


jtbandes (approved):
I guess this makes some sense. I do think it would be better if the upstream caller is unable to call this method given that it should also know the capabilities. But I suppose we need some last line of defense.

achim-k:
> I do think it would be better if the upstream caller is unable to call this method given that it should also know the capabilities.

Agreed. I tried investigating why this method is called in the first place but was unable to reproduce what I have noticed in FG-4323.

PR #6468, opened 2023-07-18
Fix copying of text from the TextMiddleTruncate component

### LINKED ISSUE #6403: Log names have a newline character in them
**Description**

Highlighting/copying the log name text always results in a newline character in the middle of the name, even though the logname is a filename with no newline character in it.

This behavior existed in previous versions of studio as well, for the log name in the left hand side panel. The log name in the titlebar on the latest foxglove has the same issue.

Screenshot 2023-07-07 at 11 19 26 AM

- Version: 1.58.0
- Platform: osx

**Steps To Reproduce**

Load an mcap file, select name in title, copy, paste

**Expected Behavior** 
name should paste exactly as the filename that was loaded

foxymiles:
**User-Facing Changes**
Fix copying of text from the TextMiddleTruncate component, used in the datasource name in the app bar.

**Description**
Overrides the `onCopy` handler of the wrapper div to copy the entire text content to the clipboard.

Fixes #6403

### DIFF
--- a/packages/studio-base/src/components/TextMiddleTruncate.tsx
+++ b/packages/studio-base/src/components/TextMiddleTruncate.tsx
@@ -11,7 +11,7 @@
 //   found at http://www.apache.org/licenses/LICENSE-2.0
 //   You may not use this file except in compliance with the License.
 
-import { CSSProperties } from "react";
+import { CSSProperties, ClipboardEvent, useCallback } from "react";
 import { makeStyles } from "tss-react/mui";
 
 const DEFAULT_END_TEXT_LENGTH = 16;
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/components/TextMiddleTruncate.tsx
```diff
@@ -79,6 +91,7 @@ export default function TextMiddleTruncate({
       className={cx(className, classes.root)}
       title={text}
       style={style}
+      onCopy={onCopy}
```
2metres, line 94: TIL about onCopy!

jtbandes:
This is kind of weird because you can now manually select one word or character from the text, but when you copy it you still get the full text. I'm not sure if that is better or worse than getting an unexpected newline – both seem like the app did not do what the user asked.

foxymiles:
> This is kind of weird because you can now manually select one word or character from the text, but when you copy it you still get the full text. I'm not sure if that is better or worse than getting an unexpected newline – both seem like the app did not do what the user asked.

I don't think there's any circumstance in which a user would want a newline. This fixes the original ask from the user. Selecting the text and copying it now provides the full text of the datasource.

foxymiles:
> Can we force selections on children to push to the parent so that we can avoid this:

IMO if this is an important workflow we want to support we should add an explicit copy button or roll it into the share dialog.

packages/studio-base/src/components/TextMiddleTruncate.tsx
foxymiles: Me too!

PR #6472, opened 2023-07-19
Use ready marker to wait for stories to complete

foxymiles:
**User-Facing Changes**
None

**Description**
Use a superimposed DOM element as a ready marker for stories. Hopefully this will be less flaky since it's relying on the storybook mechanism for waiting for stories to complete instead of a ready signal.

To make use of this stories can include the `StoryReadyMarker` component with a count like this:

```typescript
<StoryReadyMarker count={callCountRef.current} />
```

For chart stories `count` should generally be a count of the number of `pauseFrame` calls the chart has made.

And then in their play function wait for the expected text:

```typescript
play: async () => {
  await screen.findByText(storyReadyMarkerText(3));
}
```

### DIFF
--- a/packages/studio-base/src/components/TimeBasedChart/index.stories.tsx
+++ b/packages/studio-base/src/components/TimeBasedChart/index.stories.tsx
@@ -11,19 +11,22 @@
 //   found at http://www.apache.org/licenses/LICENSE-2.0
 //   You may not use this file except in compliance with the License.
 import { StoryObj } from "@storybook/react";
-import { isEqual } from "lodash";
 import cloneDeep from "lodash/cloneDeep";
-import { useState, useCallback, useRef, useEffect } from "react";
+import { useCallback, useEffect, useRef, useState } from "react";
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/components/TimeBasedChart/index.stories.tsx
```diff
@@ -138,34 +139,25 @@ export const CanZoomAndUpdate: StoryObj = {
...
         </MockMessagePipelineProvider>
+        <StoryReadyMarker count={callCountRef.current} />
       </div>
     );
   },
 
-  play: async (ctx) => {
-    await ctx.parameters.storyReady;
+  play: async () => {
+    await waitForStoryReadyMarker(3);
```
jtbandes, line 159: How is this different from `useReadySignal(3)` and the `await storyReady` mechanism we already had?

foxymiles: This is an experiment which unfortunately doesn't seem to work.

PR #6460, opened 2023-07-17
Remove LayoutID from CurrentLayoutContext

defunctzombie:
**User-Facing Changes**
None

**Description**
TBD

PR #6473, opened 2023-07-19
Remove `id` field from LayoutState

defunctzombie:
**User-Facing Changes**
None

**Description**
This change removes the _id_ field from the `LayoutState` used in `CurrentLayoutContext`. There is no concept of an `id` for a layout for `studio-base` and `CurrentLayoutContext` deals with setting and updating the layout displayed to the user in the workspace.

I've opted to land this change separately from re-working the `LayoutState` in https://github.com/foxglove/studio/pull/6460 to have a more focused discussion and smaller change to review.

### DIFF
--- a/packages/studio-base/src/providers/PanelStateContextProvider.tsx
+++ b/packages/studio-base/src/providers/PanelStateContextProvider.tsx
@@ -187,8 +187,6 @@ function hasSameKeys(
 
 const createHasSameKeySelector = createSelectorCreator(defaultMemoize, hasSameKeys);
 
-const selectCurrentLayoutId = (state: LayoutState) => state.selectedLayout?.id;
-
 const selectLayoutConfigById = createHasSameKeySelector(
   (state: LayoutState) => state.selectedLayout?.data?.configById,
   (config) => config,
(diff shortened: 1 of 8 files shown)


packages/studio-base/src/providers/PanelStateContextProvider.tsx
```diff
@@ -213,12 +211,5 @@ export function PanelStateContextProvider(props: Props): JSX.Element {
     store.setState((old) => ({ sharedPanelState: pick(old.sharedPanelState, panelTypesInUse) }));
   }, [panelTypesInUse, store]);
 
-  // clear shared panel state on layout change
```
defunctzombie, line 216: This is the only noteworthy change with the removal of _id_ from LayoutState. Since there's no _id_ there's no field to signal the entire layout changed. There is logic in an effect above this one that discards "shared panel state for panel types that are no longer in the layout" which also works as the user removes panels but does not change the layout.

I think we could drop this behavior but if we think it is important to keep this behavior - here are some alternative ideas:

1. Leave an opaque `string` _id_ field in the `LayoutState` for external use - sort of a user-data portion of the `LayoutState`.
2. Use the reference stability of the `LayoutState` itself as a signal that the state is reset (`data` is a field within the state and constantly changes but the state itself could be a stable object). I'm probably not particularly keen on this behavior since it feels fragile.
3. Move the `sharedPanelState` to `CurrentLayoutProvider` and clear it when someone calls `setCurrentLayoutState()`.

foxymiles (approved):

packages/studio-base/src/providers/PanelStateContextProvider.tsx
defunctzombie: Trying item 3 here: https://github.com/foxglove/studio/pull/6492

defunctzombie:
Replace by https://github.com/foxglove/studio/pull/6492

PR #6474, opened 2023-07-19
Allow duplicating custom layer URDFs

achim-k:
**User-Facing Changes**
- Allow duplicating custom layer URDFs

**Description**
Adds an additional menu action for duplicating the custom layer URDF

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
@@ -6,6 +6,7 @@ import { vec3 } from "gl-matrix";
 import i18next from "i18next";
 import { debounce, maxBy } from "lodash";
 import * as THREE from "three";
+import { v4 as uuidv4 } from "uuid";
 
 import { UrdfGeometryMesh, UrdfRobot, UrdfVisual, parseRobot, UrdfJoint } from "@foxglove/den/urdf";
 import Logger from "@foxglove/log";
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
```diff
@@ -410,12 +413,29 @@ export class Urdfs extends SceneExtension<UrdfRenderable> {
         // Re-add coordinate frames in case the deleted URDF shared frame names with other URDFs
         this.#refreshTransforms();
 
+        // Update the settings tree
+        this.updateSettingsTree();
+        this.renderer.updateCustomLayersCount();
+      } else if (action.payload.id === "duplicate") {
+        const newInstanceId = uuidv4();
```
jtbandes, line 420: Is there any situation where instanceId is expected to be a specific string instead of a random uuid?

jtbandes (approved):

packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
achim-k: Yes, when it is the instance for the `/robot_description` topic or parameter:
https://github.com/foxglove/studio/blob/cff1d868fb48218efa13fe71ee54085072c2682e/packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts#L655

PR #6475, opened 2023-07-19
Allow to give custom layer URDFs a label

achim-k:
**User-Facing Changes**
- Allow to give custom layer URDFs a label

**Description**
Allow to give custom layer URDFs a label which makes it easier to keep the overview when displaying multiple URDF instances.

image

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
@@ -71,6 +71,7 @@ const RPY_LABEL: [string, string, string] = ["R", "P", "Y"];
 export type LayerSettingsUrdf = BaseSettings & {
   instanceId: string; // This will be set to the topic name
   displayMode: "auto" | "visual" | "collision";
+  label: string;
 };
 
 export type LayerSettingsCustomUrdf = CustomLayerSettings & {
@@ -85,6 +86,7 @@ const DEFAULT_SETTINGS: LayerSettingsUrdf = {
   frameLocked: true,
   instanceId: "invalid",
   displayMode: "auto",
+  label: "URDF",
 };
 
 const DEFAULT_CUSTOM_SETTINGS: LayerSettingsCustomUrdf = {
@@ -285,6 +287,7 @@ export class Urdfs extends SceneExtension<UrdfRenderable> {
             value: config.framePrefix ?? "",
           },
           displayMode: { ...displayMode, value: config.displayMode ?? "auto" },
+          label: { label: "Label", input: "string", value: config.label ?? "URDF" },
         };
 
         entries.push({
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
```diff
@@ -285,6 +287,7 @@ export class Urdfs extends SceneExtension<UrdfRenderable> {
             value: config.framePrefix ?? "",
           },
           displayMode: { ...displayMode, value: config.displayMode ?? "auto" },
+          label: { label: "Label", input: "string", value: config.label ?? "URDF" },
```
jtbandes, line 290: I wonder if we should move this up to the top, or be 2nd after `url`. Looks like for the Plot panel we currently have it 2nd after the message path:

image

packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
```diff
@@ -637,6 +640,15 @@ export class Urdfs extends SceneExtension<UrdfRenderable> {
     const settingsPath = isTopicOrParam ? ["topics", instanceId] : ["layers", instanceId];
     const url = (settings as Partial<LayerSettingsCustomUrdf>).url;
     const framePrefix = (settings as Partial<LayerSettingsCustomUrdf>).framePrefix;
+    const label = (settings as Partial<LayerSettingsCustomUrdf>).label ?? "URDF";
+
+    if (label !== renderable?.userData.settings.label) {
+      // Label has changed, update the config
+      this.renderer.updateConfig((draft) => {
+        const config = draft.layers[instanceId];
+        draft.layers[instanceId] = { ...config, label };
```
jtbandes, line 649: I believe you can :just: do `draft.layers[instanceId].label = label`, since this is already inside immer produce.

packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
```diff
@@ -85,6 +86,7 @@ const DEFAULT_SETTINGS: LayerSettingsUrdf = {
   frameLocked: true,
   instanceId: "invalid",
   displayMode: "auto",
+  label: "URDF",
```
jtbandes, line 89: Instead of having a default label called "URDF", should we actually use the URL like `"URDF (http://...)"` ? Maybe the URL is usually so long that it won't be useful because it's always truncated?

jtbandes (approved):

packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
achim-k: This is what I originally started with. But then I realized that some instances might share the same URDF (url) and only the `frame prefix` is different. And as you said, the URL being usually quite long doesn't help either. In the end I decided for adding the additional `label` setting as I think it gives the user the most flexibility.

PR #6483, opened 2023-07-20
Prompt to refresh web app when chunk fails to load

jtbandes:
**User-Facing Changes**

**Description**
- Rename VersionBanner to CompatibilityBanner
- Add AppOutdatedBanner
- Override `__webpack_chunk_load__` in order to show AppOutdatedBanner when a webpack chunk fails to load.
- Add `manifest.json` output file with app version
- Inside AppOutdatedBanner, attempt to load the latest `manifest.json` and compare the version with the current app version. If the version has changed, we tell the user the app is out of date. If not, we display a more generic "_encountered an error_" message.

IMG_1906
IMG_1907


packages/studio-web/src/index.tsx
```diff
@@ -37,6 +40,18 @@ export async function main(getParams: () => Promise<MainParams> = async () => ({
...
 
+  // Detect when webpack chunks fail to load and show the AppOutdatedBanner
+  const originalChunkLoad = __webpack_chunk_load__;
+  __webpack_chunk_load__ = async (chunkId) => {
+    try {
+      return await originalChunkLoad(chunkId);
+    } catch (err) {
+      log.error("Failed to load chunk", chunkId, "- app may be outdated");
+      useAppOutdatedState.setState({ bannerShown: true });
+      throw err;
```
jtbandes, line 51: maybe this needs to be something other than a rethrow to avoid sending these to sentry? But this function is expected to return the loaded module – so if it returns something else, more things will probably break downstream anyway.

jtbandes:
Decided not to put this in the open-source app since it depends on some details of the deployment process.

PR #6484, opened 2023-07-20
Fix LaunchPreference dialog's remember preference behavior

wimagguc:
**User-Facing Changes**

Fixes a bug where unchecking the "Remember my preference" button would make the LaunchPreference dialog disappear, and open the link in the browser:

web

**Description**

Besides toggling the `rememberPreference` variable, the checkbox used to also call `setGlobalPreference` to clear the globally stored value ahead of a button click. This then caused the "side-useEffect" whereby `useDefaultWebLaunchPreference` called `setLaunchPreference(LaunchPreferenceValue.WEB)`, causing the confusing user experience.
This PR changes the checkbox to only controls rememberPreference, and leave it to the web/desktop buttons to explicitly store the setting globally depending on what the user decided to do.

### DIFF
--- a/packages/studio-base/src/screens/LaunchPreferenceScreen.tsx
+++ b/packages/studio-base/src/screens/LaunchPreferenceScreen.tsx
@@ -44,22 +44,15 @@ export function LaunchPreferenceScreen(): ReactElement {
 
   async function launchInWeb() {
     setSessionPreference(LaunchPreferenceValue.WEB); // always set session preference to allow overriding the URL param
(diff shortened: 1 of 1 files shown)


jtbandes (commented):
I wonder how this used to work and why it broke? Something about how we use sessionStorage?

wimagguc:
> I wonder how this used to work and why it broke? Something about how we use sessionStorage? Or was it always broken?

Seems like it was broken for at least a while. I looked back until January this year and it's broken all the way.

PR #6486, opened 2023-07-20
Try another fix for the TimeBasedChart zoom story

foxymiles:
**User-Facing Changes**
None

**Description**
Try another fix for the TimeBasedChart zoom story.

Changes here include slowing down the zoom events slightly since response time for these seems to vary from platform to platform and to also wait to trigger the readySignal until we've received `pauseFrame` the expected number of times.

### DIFF
--- a/packages/studio-base/src/components/TimeBasedChart/index.stories.tsx
+++ b/packages/studio-base/src/components/TimeBasedChart/index.stories.tsx
@@ -11,19 +11,19 @@
 //   found at http://www.apache.org/licenses/LICENSE-2.0
 //   You may not use this file except in compliance with the License.
 import { StoryObj } from "@storybook/react";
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/components/TimeBasedChart/index.stories.tsx
```diff
@@ -119,43 +119,37 @@ export const CanZoomAndUpdate: StoryObj = {
       }
 
       // Zoom is a continuous event, so we need to simulate wheel multiple times
-      for (let i = 0; i < 5; i++) {
+      for (const _ of range(2)) {
```
defunctzombie, line 122: Personally I like the for loop better /shrug - to my mind it reads more idiomatically but can live with this.

defunctzombie (approved):

PR #6492, opened 2023-07-20
Move sharedPanelState to CurrentLayoutContext

defunctzombie:
**User-Facing Changes**
None

**Description**
This change moves the `sharedPanelState` from the `PanelStateContext` into the `CurrentLayoutContext` alongside `selectedLayout`. This reduces an awkward coupling that had to happen between the two contexts by moving the shared state tracking for panels alongside the logic that tracks current layout updates. Architecturally the `CurrentLayoutContext` is a place for tracking the layout workspace and panel related config and state. By moving this state into the `CurrentLayoutContext`, the logic for determining when to clear the `sharedPanelState` runs when specific actions on the layout require this state to change.

---
I could see the `PanelStateContext` and `PanelStateStore` being rolled into the layout context as a single place to manage panel layout and state but I've not tackled that in this PR. The focus of this PR was to move the PanelStateProvider logic that relied on tracking the layout ID via `useEffect` and `configsById` via `useEffect` out of this provider and into the layout context where the layout events directly responsible for influencing that state could be applied to manage the state.


foxymiles (approved):
Logically I think it makes sense to consolidate these. 

Going forward if we want to keep adding stuff to this context we just need to keep throughput in mind because the initial discussion around shared panel state involved some scenarios with high frequency updates.

defunctzombie:
> Going forward if we want to keep adding stuff to this context we just need to keep throughput in mind because the initial discussion around shared panel state involved some scenarios with high frequency updates.

👍 I was not able to reproduce adverse effects in the app with this move - but if we do discover them we can re-visit the structure and this time leave comments around what falls over when you try doing it a different way

PR #6493, opened 2023-07-21
Attempt to auto-merge dependabot PRs using pascalgn/automerge-action

jtbandes:
**User-Facing Changes**
None

**Description**
New version of https://github.com/foxglove/studio/pull/5978 now that #6390 is done


.github/workflows/dependabot-automerge.yml
```diff
@@ -0,0 +1,41 @@
...
+        if: ${{ github.actor == 'dependabot[bot]' && (github.event.action == 'opened' || github.event.action == 'reopened' || github.event.action == 'ready_for_review') }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.FOXGLOVEBOT_GITHUB_TOKEN }}
+        with:
+          route: POST /repos/{owner_and_repo}/pulls/{pull_number}/reviews
+          owner_and_repo: ${{ github.repository }}
+          pull_number: ${{ github.event.pull_request.number }}
+          event: "APPROVE"
+
+      - uses: pascalgn/automerge-action@v0.15.6
```
defunctzombie, line 35: Let's fork this action and run from our forked version. I'm a little meh about sending our private github token to a 3rd party action like this. Maybe we can pin to a specific commit hash so it is harder to swap out from under us?

.github/workflows/dependabot-automerge.yml
```diff
@@ -0,0 +1,41 @@
+# Automatically approve and merge PRs made by Dependabot
+
+name: Dependabot
+
+on:
+  pull_request_target:
```
defunctzombie, line 6: Why do we need all these types?

defunctzombie (approved):
Seems good - would be nice to test on some sample repo first and make sure it works.

defunctzombie:
I do wonder whether we live in a world of constant dependabot updates for no real gains. I guess if the bots approve their own work its fine.

jtbandes:
Superseded by #6496

PR #6495, opened 2023-07-21
Try fix for flaky plot legend story

foxymiles:
**User-Facing Changes**
None

**Description**
Try fix to flaky plot legend story by explicitly waiting for the ready signal instead of increasing delay.

### DIFF
--- a/packages/studio-base/src/panels/Plot/PlotLegend.stories.tsx
+++ b/packages/studio-base/src/panels/Plot/PlotLegend.stories.tsx
@@ -3,6 +3,7 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
 import { StoryObj, StoryFn } from "@storybook/react";
+import { waitFor } from "@storybook/testing-library";
 import { useCallback } from "react";
 
 import Stack from "@foxglove/studio-base/components/Stack";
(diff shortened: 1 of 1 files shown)


achim-k (approved):

packages/studio-base/src/panels/Plot/PlotLegend.stories.tsx
```diff
@@ -152,10 +153,11 @@ export const LimitWidth: StoryObj = {
     );
   },
 
+  play: async (ctx) => {
+    await waitFor(() => ctx.parameters.storyReady);
```
jtbandes, line 157: What's the difference between `await waitFor(() => ctx.parameters.storyReady);` and `await ctx.parameters.storyReady;`? Any reason for this to differ from some other stories we have?

foxymiles: This will register as an error if the promise doesn't resolve. A bit stronger signal than a snapshot mismatch.

jtbandes: What would `await ctx.parameters.storyReady;` do? It would probably show some kind of timeout error if it hits the 15 sec chromatic limit?

foxymiles: At least when running locally it just seems to silently not do anything.

jtbandes: Yeah I think the 15 sec timeout is a Chromatic only concept. So maybe it's better to use waitFor with all our stories because it makes the error visible locally too?

foxymiles: Yeah maybe so. We could wait and see if this seems to be really fixed finally and if so generalize it.

PR #6500, opened 2023-07-24
Auto-save when leaving unsaved userscript

snosenzo:
**User-Facing Changes**

- Autosave when leaving unsaved userscript

**Description**
Auto save user script when the panel unmounts, like when switching tabs in the layout.

FG-4161

### DIFF
--- a/packages/studio-base/src/panels/NodePlayground/index.tsx
+++ b/packages/studio-base/src/panels/NodePlayground/index.tsx
@@ -320,26 +320,22 @@ function NodePlayground(props: Props) {
     [scriptBackStack],
   );
 
-  const askSaveOnLeave = useCallback(() => {
+  const saveOnLeave = useCallback(() => {
     if (isNodeSaved) {
       return;
     }
-    const response = confirm(
-      `You have unsaved changes in your user script: ${currentScript.filePath}. Would you like to save?`,
-    );
-    if (response) {
-      saveCurrentNode();
-    }
-  }, [isNodeSaved, saveCurrentNode, currentScript?.filePath]);
+    // automatically save script on panel leave
+    saveCurrentNode();
+  }, [isNodeSaved, saveCurrentNode]);
 
   // The cleanup function below should only run when this component unmounts.
   // We're using a ref here so that the cleanup useEffect doesn't run whenever one of the callback
   // dependencies changes, only when the component unmounts and with the most up-to-date callback.
-  const askSaveOnLeaveRef = useRef(askSaveOnLeave);
-  askSaveOnLeaveRef.current = askSaveOnLeave;
+  const saveOnLeaveRef = useRef(saveOnLeave);
+  saveOnLeaveRef.current = saveOnLeave;
   useEffect(() => {
     return () => {
-      askSaveOnLeaveRef.current();
+      saveOnLeaveRef.current();
     };
   }, []);
 
--- a/packages/studio-base/src/panels/NodePlayground/index.tsx
+++ b/packages/studio-base/src/panels/NodePlayground/index.tsx
@@ -24,7 +24,7 @@ import {
   Typography,
   inputClasses,
 } from "@mui/material";
-import { Suspense, useCallback, useEffect, useState } from "react";
+import { Suspense, useCallback, useEffect, useRef, useState } from "react";
 import { makeStyles } from "tss-react/mui";
 import { v4 as uuidv4 } from "uuid";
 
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/panels/NodePlayground/index.tsx
```diff
@@ -320,6 +320,24 @@ function NodePlayground(props: Props) {
...
+    if (isNodeSaved) {
+      return;
+    }
+    const response = confirm("You have unsaved changes. Would you like to save before leaving?");
+    if (response) {
+      saveCurrentNode();
+    }
+  }, [isNodeSaved, saveCurrentNode]);
+
+  const cleanupRef = useRef(askSaveOnLeave);
```
defunctzombie, line 333: We need a comment here describing why we are using a ref.

I also think a nicer name for this is `askSaveOnLeafRef`. Something that aligns with the name of the thing you are referring to.

packages/studio-base/src/panels/NodePlayground/index.tsx
```diff
@@ -320,6 +320,24 @@ function NodePlayground(props: Props) {
     [scriptBackStack],
   );
 
+  const askSaveOnLeave = useCallback(() => {
+    if (isNodeSaved) {
+      return;
+    }
+    const response = confirm("You have unsaved changes. Would you like to save before leaving?");
```
defunctzombie, line 327: _leaving_ sounds strange to me here but I also can't think of a better word choice. Maybe drop the "before leaving"? So `You have unsaved changes. Would you live to save?`

defunctzombie (approved):
Do we think it would be fine to save the changes automatically for the user on unmount instead of prompting? I suppose in the case they didn't want the changes saved it would be annoying but I can't particularly think of why I'd not want the changes saved? We don't save while you edit cause it would re-run nodes.

snosenzo:
@defunctzombie 

> Do we think it would be fine to save the changes automatically for the user on unmount instead of prompting?

I think for shared layouts that might be dangerous. Since we have an explicit save button I also think that behavior would be a departure from our current pattern where we save things explicitly. If users don't like the prompt I'm sure they'll say so and we can move to saving automatically if desired, or having a setting, but right now I think saving automatically would be a bit preemptive.

packages/studio-base/src/panels/NodePlayground/index.tsx
snosenzo: I've updated this to remove the `before leaving` and also to include the script file path in case there are multiple user scripts in the tab.

2metres (commented):
Should we batch up the confirm into a single modal instead of 2? 

I would also be in favor of adding an autosave on blur as a panel setting. I personally haven't saved a file in a code editor in at least 5 years.

defunctzombie:
> I think for shared layouts that might be dangerous. Since we have an explicit save button I also think that behavior would be a departure from our current pattern where we save things explicitly. If users don't like the prompt I'm sure they'll say so and we can move to saving automatically if desired, or having a setting, but right now I think saving automatically would be a bit preemptive.

@snosenzo this would only save the script changes into the active layout - rather than saving the layout. The layout is still dirty until the user explicitly saves it.

defunctzombie:
> Should we batch up the confirm into a single modal instead of two modals? I feel like repetitive popups could become really annoying.

Where are you seeing two modals?

2metres:
> > Should we batch up the confirm into a single modal instead of two modals? I feel like repetitive popups could become really annoying.
> 
> Where are you seeing two modals?

In the attached video

snosenzo:
@defunctzombie in the screen recording in the description there are two sequential popups for each unsaved script.

snosenzo:
> Should we batch up the confirm into a single modal instead of two modals? I feel like repetitive popups could become really annoying.

@2metres  I'm not sure this is that feasible given these dialogues are coming from two different panels. If this is a dealbreaker maybe we should just do onBlur. But at that point should we also remove the `Save` button?

defunctzombie:
> But at that point should we also remove the `Save` button?

We need the save button (unless we autosave with a debounce) because the user may want to save without having to click off the panel.

I see these as possible future enhancements rather than requirements to tangle into this change. For this change we should either ship this confirm or auto-save the script changes when you navigate away.

amacneil:
review feedback: use autosave instead of dialog

packages/studio-base/src/panels/NodePlayground/index.tsx
```diff
@@ -320,6 +320,25 @@ function NodePlayground(props: Props) {
...
+      return;
+    }
+    // automatically save script on panel leave
+    saveCurrentNode();
+  }, [isNodeSaved, saveCurrentNode]);
+
+  // The cleanup function below should only run when this component unmounts.
+  // We're using a ref here so that the cleanup useEffect doesn't run whenever one of the callback
+  // dependencies changes, only when the component unmounts and with the most up-to-date callback.
+  const saveOnLeaveRef = useRef(saveOnLeave);
```
defunctzombie, line 334: There's a convenience hook for this we have called `useLatest`. look around for that in the codebase and you'll see it does these two lines for you.

defunctzombie (approved):

PR #6501, opened 2023-07-24
Fix state transition panel zooming behavior

foxymiles:
**User-Facing Changes**
Fix an issue with state transition panel zooming behavior.

**Description**
ChartJS seems to get confused when zooming into datasets with segmented and discontinuous datasets. This simplifies our state transition dataset generation by putting everything into a single dataset and using per-point labels and segment colors to distinguish different states. This should also help performance.

This requires applying a per-segment coloring function in `ChartJSManager` because functions can't be serialized when sending data to the worker.

### DIFF
--- a/packages/studio-base/src/components/Chart/worker/ChartJSManager.ts
+++ b/packages/studio-base/src/components/Chart/worker/ChartJSManager.ts
@@ -23,6 +23,8 @@ import { RpcElement, RpcScales } from "@foxglove/studio-base/components/Chart/ty
 import { maybeCast } from "@foxglove/studio-base/util/maybeCast";
 import { fonts } from "@foxglove/studio-base/util/sharedStyleConstants";
 
+import { lineSegmentLabelColor } from "./lineSegments";
+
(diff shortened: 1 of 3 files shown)


defunctzombie:
> single dataset and using per-point labels and segment colors to distinguish different states. This should also help performance.

Awesome! When we first started using chartjs it didn't have this feature.

packages/studio-base/src/components/Chart/worker/ChartJSManager.ts
```diff
@@ -197,6 +199,14 @@ export default class ChartJSManager {
       return {};
     }
 
+    for (const ds of data?.datasets ?? []) {
+      // Apply a line segment coloring function, if the label color is present in the data points.
+      // This has to happen here because functions can't be serialized to the chart worker.
```
defunctzombie, line 204: Add a comment referencing the state transition panel so future selves can know why this has to happen here and what other logic is connects to.

defunctzombie (approved):
So happy to see that weird multi-series code banished

PR #6509, opened 2023-07-25
Fix problem title only fully visible when resizing sidebar

achim-k:
**User-Facing Changes**
Not worthy to be mentioned in release notes

**Description**
Adds a `title` attribute to the problem accordion which allows to see the full error text without having to make the sidebar wider.

@2metres, for copying the error message one would still have to resize the sidebar. I could add a copy icon button for this? WDYT? 

[Screencast from 25.07.2023 11:43:31.webm](https://github.com/foxglove/studio/assets/9250155/8b8c0b80-ff0c-4efa-bda5-1c6176b384d7)


2metres (approved):
Looks good, I think we should still think about how we use the problem description vs title. I could see adding a copy button to the description

PR #6515, opened 2023-07-25
[FG-3927] Image flickering in adjacent Image panels

cfoust:
**User-Facing Changes**
Before, users experienced periodic "flashes" under certain circumstances in the Image panel. This was caused by coincident calls to `Renderer.#frameHandler` which could cause two frames to be rendered on top of each other without the frame being cleared.

Before:

https://github.com/foxglove/studio/assets/4588553/4fdb354a-ad56-4fea-8244-220bef69fdb1

After:

https://github.com/foxglove/studio/assets/4588553/40875a99-e427-44fc-9e02-70e13a73977f

**Description**
This change prevents `#frameHandler` calls from interfering with each other. It's stupid, but it works. (I didn't bother investigating why or how several `#frameHandler` calls could overlap in the way they were, but at any rate this forces it to be consistent. Perhaps it's worth deep-diving our rendering pipeline timing at some point.)

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
@@ -212,6 +212,7 @@ export class Renderer extends EventEmitter<RendererEvents> implements IRenderer 
 
   #prevResolution = new THREE.Vector2();
   #pickingEnabled = false;
+  #rendering = false;
   #animationFrame?: number;
   #cameraSyncError: undefined | string;
   #devicePixelRatioMediaQuery?: MediaQueryList;
@@ -1024,7 +1025,10 @@ export class Renderer extends EventEmitter<RendererEvents> implements IRenderer 
 
   public animationFrame = (): void => {
     this.#animationFrame = undefined;
-    this.#frameHandler(this.currentTime);
+    if (!this.#rendering) {
+      this.#frameHandler(this.currentTime);
+      this.#rendering = false;
+    }
   };
 
   public queueAnimationFrame(): void {
@@ -1045,6 +1049,7 @@ export class Renderer extends EventEmitter<RendererEvents> implements IRenderer 
   }
 
   #frameHandler = (currentTime: bigint): void => {
+    this.#rendering = true;
     this.currentTime = currentTime;
     this.#updateFrameErrors();
     this.#updateFixedFrameId();


packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -212,6 +212,7 @@ export class Renderer extends EventEmitter<RendererEvents> implements IRenderer
 
   #prevResolution = new THREE.Vector2();
   #pickingEnabled = false;
+  #rendering = false;
```
snosenzo, line 215: Do we need this extra variable? Can we use `#animationFrame != undefined` to determine whether it's rendering? It feels like that might've been that variable's original intention.

snosenzo: talked through on slack that the using `#animationFrame` along with making the component use `renderer.queueAnimationFrame` brings back the flicker.

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -1024,7 +1025,10 @@ export class Renderer extends EventEmitter<RendererEvents> implements IRenderer
 
   public animationFrame = (): void => {
     this.#animationFrame = undefined;
-    this.#frameHandler(this.currentTime);
+    if (!this.#rendering) {
```
snosenzo, line 1028: would be good to have a comment for why we want/need to prevent calling `#frameHandler` while rendering

snosenzo (approved):
great find!

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -1024,7 +1025,10 @@ export class Renderer extends EventEmitter<RendererEvents> implements IRenderer
...
+      this.#frameHandler(this.currentTime);
+      this.#rendering = false;
```
defunctzombie, line 1030: The miss-match with `#rendering = false` and true is a bit hard to follow. Maybe worth having this logic all in `#frameHandler`? Or all here if this is the only place frameHandler is called?

I am suspicious of why animation frame is not better gated higher up or maybe this is the appropriate place to do that gating. It does feel like there was some attempt with `#animationFrame` but as noted above it is not working?

cfoust:
It seems like the `animationFrame`, `#animationFrame` and `queueAnimationFrame` logic was intended to ensure that the canvas is rerendered once, no matter how many times `queueAnimationFrame` was called, but the `Renderer` code has a bunch of invocations of `animationFrame` directly, which evidently are causing some (rare) issues where the scene is rendering twice without a `clear()` call in between. I asked @snosenzo whether it was worth the time to explore why this is the case specifically, but opted to get a fix out first and judge whether that exploration was worthwhile at a later time.

In general, this rendering pipeline could use some work. Random parts of the code should not _directly_ (= in some subroutine, here `animationFrame`) issue draw calls. But we can fix that later.

PR #6516, opened 2023-07-25
[Map Panel] Fix rendering of points with NO_FIX status

bryfox:
**User-Facing Changes**

Fix rendering of map points having NO_FIX status.

**Description**

This loosens the map panel's definition of a valid message. In response to https://github.com/foxglove/studio/issues/4763, we began filtering points which had invalid coordinates or no fix. However, per https://github.com/orgs/foxglove/discussions/775, a device may still estimate good coordinates through other means, and we should continue to display those coordinates even when status is NO_FIX.

### DIFF
--- a/packages/studio-base/src/panels/Map/index.stories.tsx
+++ b/packages/studio-base/src/panels/Map/index.stories.tsx
@@ -330,47 +330,6 @@ export const MultipleTopics: StoryObj = {
   },
 };
 
-export const SinglePointNoFix: StoryObj = {
-  render: function Story() {
-    return <MapPanel />;
-  },
-
-  decorators: [Wrapper],
-
-  parameters: {
-    chromatic: {
-      delay: 1000,
-    },
-    decorators: [Wrapper],
-    panelSetup: {
-      fixture: {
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/panels/Map/index.stories.tsx
```diff
@@ -292,47 +330,6 @@ export const MultipleTopics: StoryObj = {
   },
 };
 
-export const SinglePointNoFix: StoryObj = {
```
bryfox, line 295: This story is valid, but I deleted it because it duplicates `SinglePointWithNoFix` above.

foxymiles (approved):
I seem to recall there was some disagreement at the time about what we should do with these messages. If we agree that we should show them even without a fix then this seems like a reasonable fix.

bryfox:
> some disagreement at the time about what we should do with these messages

I see. I don't have a strong opinion, but the linked discussion seems to be a case where I'd want to see the points. And since we render no-fix points differently, there's a way for the user to distinguish them. If it proves to be confusing, we could update that style and/or document the behavior.

foxymiles:
I guess if this turns out to be controversial we can give users a setting to control the behavior.

PR #6522, opened 2023-07-26
Add check for fetch response status when fetching assets

### LINKED ISSUE #6519: Mesh Error
**Description**

- Version: 1.62+
- Platform: Ubuntu 20
- Data source: MCAP file/websocket

**Steps To Reproduce**
We've been sending geometry messages using ROS Markers that have a path to a local mesh. As of 1.62, Foxglove is consistently registering those filepaths as non-existant. To reproduce, send ROS marker messages with the mesh filename as a path to a local file. You may need to send several different meshes; it seems to *sometimes* work, but not consistently.

**Expected Behavior**
ROS Marker messages with a local filepath will be usable in Foxglove. We've verified that 1.62+ is what breaks this; rolling back to 1.61.1 restores functionality.

achim-k:
**User-Facing Changes**
Not worth being mentioned in release notes

**Description**
~We were rejecting any asset uri that was not a valid URL. However, the Desktop app is able to resolve assets given as file path (e.g. `/home/foo/mesh.dae`). This patch changes the behavior to only throw an invalid URL error if it's not a desktop app.~

~Fixes #6519~

Update: We agreed on leaving it as it was and not support invalid URIs. This PR now only adds an additional check for the fetch response status.

### DIFF
--- a/packages/studio-base/src/components/MessagePipeline/store.ts
+++ b/packages/studio-base/src/components/MessagePipeline/store.ts
@@ -8,7 +8,6 @@ import shallowequal from "shallowequal";
 import { createStore, StoreApi } from "zustand";
 
 import { Condvar } from "@foxglove/den/async";
-import Logger from "@foxglove/log";
 import { MessageEvent } from "@foxglove/studio";
 import {
   AdvertiseOptions,
@@ -23,8 +22,6 @@ import isDesktopApp from "@foxglove/studio-base/util/isDesktopApp";
 
 import { FramePromise } from "./pauseFrameForPromise";
 import { MessagePipelineContext } from "./types";
-
-const log = Logger.getLogger(__filename);
 
 export function defaultPlayerState(): PlayerState {
   return {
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/components/MessagePipeline/store.ts
```diff
@@ -118,9 +121,20 @@ export function createMessagePipelineStore({
...
-        const { protocol } = new URL(uri);
-        const player = get().player;
+        let protocol: string | undefined = undefined;
+        try {
+          protocol = new URL(uri).protocol;
+        } catch (err) {
+          if (isDesktopApp()) {
+            // We only log the error here and continue with a normal _fetch_ call as the Desktop
+            // app may still be able to resolve the "URI" to a local file.
+            log.error("Invalid URI passed to fetchAsset", err);
```
snosenzo, line 131: why are we logging this as an invalid URI if it could still resolve to a local file? Seems strange to log an error for something that might not be an error.

achim-k: removed it

snosenzo (approved):

packages/studio-base/src/components/MessagePipeline/store.ts
```diff
@@ -118,9 +118,18 @@ export function createMessagePipelineStore({
...
-        const { protocol } = new URL(uri);
-        const player = get().player;
+        let protocol: string | undefined = undefined;
+        try {
+          protocol = new URL(uri).protocol;
+        } catch (err) {
+          // Bail out if this is not a desktop app. If this is a local file path, the Desktop app may
+          // still be able to resolve that to a local file with the _fetch_ call.
+          if (!isDesktopApp()) {
+            throw new Error(`${uri} is not a valid URI`);
```
defunctzombie, line 128:
```suggestion
            throw new Error(`Invalid URI: ${uri}`);
```

achim-k:
Update: We agreed on leaving it as it was and not support invalid URIs. This PR now only adds an additional check for the fetch response status.

snosenzo (approved):

PR #6523, opened 2023-07-26
fix seek forward for players that don't implement playUntil

shuhaowu:
** User-facing changes **
Fix a bug where "Seek forward" would start to play instead of stopping after seeking.

** Description **

Since the introduction of `TopicAliasingPlayer` between `UserNodePlayer` and the underlying player, the seek forward button is broken for underlying players that do not implement the `playUntil` method and relied on the fallback `seekPlayback` method call. The `PlaybackControls` and `UserNodePlayer` both has the fallback code paths for calling `seekPlayback` in case `playUntil` is undefined. This is not implemented in `TopicAliasingPlayer`. This patch introduces this fallback code in `TopicAliasingPlayer` to fix this issue.

Fixes: FG-4440

### DIFF
--- a/packages/studio-base/src/players/TopicAliasingPlayer/TopicAliasingPlayer.ts
+++ b/packages/studio-base/src/players/TopicAliasingPlayer/TopicAliasingPlayer.ts
@@ -120,7 +120,11 @@ export class TopicAliasingPlayer implements Player {
   }
 
   public playUntil?(time: Time): void {
-    this.#player.playUntil?.(time);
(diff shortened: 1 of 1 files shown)


foxymiles:
This makes sense to me but maybe @defunctzombie is a bit more familiar with the expected semantics of `playUntil`?

defunctzombie (approved):

defunctzombie:
Thank you for the bugfix!

PR #6524, opened 2023-07-26
[FG-4201] Update color attribute when new triangles arrive

cfoust:
**User-Facing Changes**

Fix rendering bug with vertex-colored triangle meshes.

**Description**

`RenderableTriangles.#updateTriangleMeshes` only made ThreeJS aware of updates to a mesh's colors if the material had not been initialized. This seems to be a mistake; we do all of this work to build `colorChanged` and then [throw out the result](https://github.com/foxglove/studio/blob/main/packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableTriangles.ts?plain=1#L127). If `colorChanged` is true at all we know that (a) `geometry.colors` is defined and (b) we need to flag it for update. That's what this PR does.

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableTriangles.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableTriangles.ts
@@ -120,6 +120,9 @@ export class RenderableTriangles extends RenderablePrimitive {
         geometry.computeBoundingSphere();
         geometry.attributes.position!.needsUpdate = true;
       }
+      if (colorChanged) {
+        geometry.attributes.color!.needsUpdate = true;
+      }
 
       // covers the case where a geometry went from being defined by a single color to vertex colors
       // but there was no difference in the vertex colors that already existed and the new ones


packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableTriangles.ts
```diff
@@ -120,6 +120,9 @@ export class RenderableTriangles extends RenderablePrimitive {
         geometry.computeBoundingSphere();
         geometry.attributes.position!.needsUpdate = true;
       }
+      if (colorChanged) {
```
snosenzo, line 123: Hmm yeah I think this `colorChanged` variable is a little overused. 
We should split it up into one flag that represents whether the vertex colors have changed which will handle setting this `needsUpdate` on the geometry attribute, and one that represents whether vertex colors were just turned on after being off (`!material.vertexColors`) which will handle the setup of the material (on line 131).

cfoust: In general, I think this function could use a cleanup, particularly this section with the material--notice that one part already uses a variable called `materialNeedsUpdate`. It seems like the previous state of the objects used for rendering influences subsequent calls in hard-to-analyze ways; as a result I'm hesitant to play with the logic here too much without more confidence that I fully understand the different states here.

cfoust: Though my bias is towards not doing a cleanup right now if (a) I don't have a good understanding of the state space of this function and (b) I don't have data that captures most of the edge cases

snosenzo: Fair enough, I do think we should at least change the `colorChanged = !material.vertexColors...` variable below to be it's own `const changedToVertexColors` so that it's less confusing for the next folks who come across this.

PR #6529, opened 2023-07-26
Fix comparison of message definitions

achim-k:
**User-Facing Changes**
Fix comparison of message definitions

**Description**
The existing message definition comparison would wrongly flag definitions as unequal when parsed from a `IDL` and `MSG` file.

See also https://github.com/foxglove/message-definition/pull/1
Fixes https://github.com/foxglove/ros-foxglove-bridge/issues/253

### DIFF
--- a/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
+++ b/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
@@ -3,13 +3,13 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
 import * as base64 from "@protobufjs/base64";
-import { isEqual, isMatch, uniqWith } from "lodash";
+import { isEqual, uniqWith } from "lodash";
 import { v4 as uuidv4 } from "uuid";
 
 import { debouncePromise } from "@foxglove/den/async";
 import Log from "@foxglove/log";
 import { parseChannel, ParsedChannel } from "@foxglove/mcap-support";
-import { MessageDefinition } from "@foxglove/message-definition";
+import { MessageDefinition, isMsgDefEqual } from "@foxglove/message-definition";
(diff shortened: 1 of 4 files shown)


packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
```diff
@@ -1198,7 +1198,7 @@ export default class FoxgloveWebSocketPlayer implements Player {
     const maybeRos = ["ros1", "ros2"].includes(this.#profile ?? "");
     for (const [name, types] of datatypes) {
       const knownTypes = this.#datatypes.get(name);
-      if (knownTypes && !isMatch(types, knownTypes)) {
+      if (knownTypes && !isMsgDefEqual(types, knownTypes as MessageDefinition)) {
```
defunctzombie, line 1201: Why the cast here? This looks suspicious to me because `#datatypes.get(name)` should have `knownTypes` be typed as a MessageDefinition from previous parsing.

`#datatypes: RosDatatypes` seems like this is an outdated type description. Let's make a followup linear issue to get this updated to `MessageDefinition` instead of `RosDatatypes` or whatever the correct type is to remove this _rosism_ at the player level.

defunctzombie (approved):

packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
achim-k: > Why the cast here? This looks suspicious to me because `#datatypes.get(name)` should have `knownTypes` be typed as a MessageDefinition from previous parsing.

Turns out the cast is not necessary. 

> `#datatypes: RosDatatypes` seems like this is an outdated type description. Let's make a followup linear issue to get this updated to `MessageDefinition` instead of `RosDatatypes` or whatever the correct type is to remove this _rosism_ at the player level.

-> FG-4446

PR #6530, opened 2023-07-26
[FG-4246] Don't add new point cloud history entries unnecessarily

cfoust:
**User-Facing Changes**

Fix for bug where adjusting point cloud settings created copies of point cloud

**Description**

Adjusting the render settings for a point cloud with decay enabled created lots of duplicate history entries, in effect causing the point cloud to render many more times than necessary.

This fix is naive and tentative: when we attempt to add a new history entry whose `receiveTime` matches the history entry at the top of the stack, we ignore it. In practice this should never occur otherwise in production workloads, though I'll be the first to admit that this doesn't feel very clean.

Fixed:
https://github.com/foxglove/studio/assets/4588553/83939d1d-a5c7-4aa5-9b0b-b93580cf8b13

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/PointClouds.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/PointClouds.ts
@@ -274,53 +274,6 @@ export class PointCloudHistoryRenderable extends Renderable<PointCloudHistoryUse
       return;
     }
 
-    const topic = this.userData.topic;
-    const isDecay = settings.decayTime > 0;
-    if (isDecay) {
-      // Push a new (empty) entry to the history of points
-      const geometry = createGeometry(topic, THREE.StaticDrawUsage);
-      const points = new PointCloudRenderable(
-        topic,
-        {
-          receiveTime: -1n, // unused
-          messageTime: -1n, // unused
-          frameId: "", //unused
-          pose: getPose(pointCloud),
-          settingsPath: [], //unused
-          settings: { visible: true }, //unused
-          topic,
-          pointCloud,
-          originalMessage,
-        },
-        geometry,
-        material,
-        this.userData.pickingMaterial,
-        this.userData.instancePickingMaterial,
-      );
-      pointsHistory.addHistoryEntry({ receiveTime, messageTime, renderable: points });
-      this.add(points);
-
-      if (settings.stixelsEnabled) {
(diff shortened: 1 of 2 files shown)


snosenzo (changes requested):
So I can see that `handleSettingsAction` calls `updatePointCloud` with the last `renderable.userData.latestPointCloud` which is why `addHistoryEntry` is being called every time the settings change. At the point that `addHistoryEntry` is being called a `new PointCloudRenderable` has been made with a new `geometry`, and this has been added to the scene. It doesn't look like this renderable is disposed of either. I think it would be good to find a solution where these duplicate `PointCloudRenderable` aren't created at all when the settings change.

packages/studio-base/src/panels/ThreeDeeRender/renderables/PointClouds.ts
```diff
@@ -888,6 +904,7 @@ export class PointClouds extends SceneExtension<PointCloudHistoryRenderable> {
       this.renderables.set(topic, renderable);
     }
 
+    renderable.pushHistory(pointCloud, originalMessage, renderable.userData.settings, receiveTime);
```
snosenzo, line 907:
```suggestion
    if(renderable.userData.settings.decayTime > 0) {
      renderable.pushHistory(pointCloud, originalMessage, renderable.userData.settings, receiveTime);
    }  
```
Looks a little strange to see `pushHistory` if we don't know that decay is active.

packages/studio-base/src/panels/ThreeDeeRender/renderables/PointClouds.ts
```diff
@@ -373,6 +327,68 @@ export class PointCloudHistoryRenderable extends Renderable<PointCloudHistoryUse
...
+    receiveTime: bigint,
+  ): void {
+    const messageTime = toNanoSec(getTimestamp(pointCloud));
+    const pointsHistory = this.#pointsHistory;
+    const stixelsHistory = this.#stixelsHistory;
+    const material = this.userData.material;
+    const stixelMaterial = this.userData.stixelMaterial;
+    const topic = this.userData.topic;
+    const isDecay = settings.decayTime > 0;
+    if (!isDecay) {
```
snosenzo, line 344: I think it'd be a bit more readable if this check was moved outside of the function call as I suggest below

snosenzo (approved):
minor readability nit, but looks good

snosenzo (approved):

PR #6533, opened 2023-07-27
Reduce plot legend rerendering

foxymiles:
**User-Facing Changes**
None (small increase in plot panel performance)

**Description**
The plot legend is an expensive component to rerender, partly because it uses complex dynamic styles. On a M1 Pro it seems to take ~5ms per legend. On a complex layout with many plots this can be a significant performance hit. When we're not displaying values there's no need to update the legend so this PR memoizes the legend component and stabilizes its properties and the hover values the legend rows receive to eliminate this overhead.

Screenshot 2023-07-27 at 6 18 46 AM

### DIFF
--- a/packages/studio-base/src/context/TimelineInteractionStateContext.tsx
+++ b/packages/studio-base/src/context/TimelineInteractionStateContext.tsx
@@ -88,11 +88,12 @@ export function useSetHoverValue(): TimelineInteractionStateStore["setHoverValue
 const undefinedSelector = () => undefined;
 
 /**
- * Encapsulates logic for selecting the current hover value depending
- * on which component registered the hover value.
+ * Encapsulates logic for selecting the current hover value depending on which component registered
+ * the hover value.
  *
  * @param componentId the component to scope updates to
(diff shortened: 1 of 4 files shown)


packages/studio-base/src/context/TimelineInteractionStateContext.tsx
```diff
@@ -85,24 +85,26 @@ export function useSetHoverValue(): TimelineInteractionStateStore["setHoverValue
...
 }
 
+const undefinedSelector = () => undefined;
+
 /**
  * Encapsulates logic for selecting the current hover value depending
  * on which component registered the hover value.
+ *
+ * @param componentId the component to scope updates to
+ * @param disableUpdates if true then return undefined, for performance reasons
```
defunctzombie, line 95: Any comment that says "for performance reasons" should be paired with an explanation on what those reasons are so that we can appropriately re-visit those assumptions as architecture or system characteristics changes.

defunctzombie (approved):

PR #6534, opened 2023-07-27
Foxglove weboscket: Avoid emitting unnecessary parameter & topic stats changes

achim-k:
**User-Facing Changes**
Small performance improvement for Foxglove weboscket connections

**Description**
Avoids falsely flagging parameters & topic stats as changed even though they did not change at all. This was causing unnecessary function calls / re-rendering in e.g. components that were subscribing to parameter changes.

### DIFF
--- a/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
+++ b/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
@@ -198,19 +198,20 @@ export default class FoxgloveWebSocketPlayer implements Player {
       this.#channelsByTopic.clear();
       this.#servicesByName.clear();
       this.#serviceResponseCbs.clear();
-      this.#parameters = new Map();
-      this.#profile = undefined;
-      this.#publishedTopics = undefined;
-      this.#subscribedTopics = undefined;
-      this.#advertisedServices = undefined;
       this.#publicationsByTopic.clear();
-      this.#datatypes = new Map();
-
       for (const topic of this.#resolvedSubscriptionsByTopic.keys()) {
         this.#unresolvedSubscriptions.add(topic);
       }
       this.#resolvedSubscriptionsById.clear();
       this.#resolvedSubscriptionsByTopic.clear();
+
+      // Re-assign members that are emitted as player state
+      this.#profile = undefined;
+      this.#publishedTopics = undefined;
+      this.#subscribedTopics = undefined;
+      this.#advertisedServices = undefined;
+      this.#datatypes = new Map();
+      this.#parameters = new Map();
     });
 
     this.#client.on("error", (err) => {
@@ -532,7 +533,7 @@ export default class FoxgloveWebSocketPlayer implements Player {
           topicStats.set(topic, stats);
         }
         stats.numMessages++;
-        this.#topicsStats = new Map(topicStats);
+        this.#topicsStats = topicStats;
       } catch (error) {
         this.#problems.addProblem(`message:${chanInfo.channel.topic}`, {
           severity: "error",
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
```diff
@@ -198,7 +198,7 @@ export default class FoxgloveWebSocketPlayer implements Player {
       this.#channelsByTopic.clear();
       this.#servicesByName.clear();
       this.#serviceResponseCbs.clear();
-      this.#parameters.clear();
+      this.#parameters = new Map();
```
snosenzo, line 201: maybe a comment as to why we need a new map instead of clearing? Is there a reason we don't create new maps for everything?

packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
```diff
@@ -525,12 +525,14 @@ export default class FoxgloveWebSocketPlayer implements Player {
...
-        let stats = this.#topicsStats.get(topic);
+        const topicStats = new Map(this.#topicsStats);
+        let stats = topicStats.get(topic);
         if (!stats) {
           stats = { numMessages: 0 };
-          this.#topicsStats.set(topic, stats);
+          topicStats.set(topic, stats);
         }
         stats.numMessages++;
+        this.#topicsStats = new Map(topicStats);
```
snosenzo, line 535: I think we can do `this.#topicStats = topicStats`

packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
```diff
@@ -573,7 +575,9 @@ export default class FoxgloveWebSocketPlayer implements Player {
         this.#parameters = new Map(mappedParameters.map((param) => [param.name, param.value]));
       } else {
         // Update params
-        mappedParameters.forEach((param) => this.#parameters.set(param.name, param.value));
+        const updatedParameters = new Map(this.#parameters);
+        mappedParameters.forEach((param) => updatedParameters.set(param.name, param.value));
+        this.#parameters = new Map(updatedParameters);
```
snosenzo, line 580: also here
`this.#parameters = updatedParameters`

packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
```diff
@@ -717,12 +721,14 @@ export default class FoxgloveWebSocketPlayer implements Player {
...
-    for (const topic of this.#topicsStats.keys()) {
+    const topicStats = new Map(this.#topicsStats);
+    for (const topic of topicStats.keys()) {
       if (!topicsSet.has(topic)) {
-        this.#topicsStats.delete(topic);
+        topicStats.delete(topic);
       }
     }
 
+    this.#topicsStats = new Map(topicStats);
```
snosenzo, line 731: same for here -- I don't think we need another new map

packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
```diff
@@ -839,14 +845,16 @@ export default class FoxgloveWebSocketPlayer implements Player {
...
-        this.#topicsStats.delete(topic);
+        topicStats.delete(topic);
 
         setTimeout(
           () => this.#recentlyCanceledSubscriptions.delete(subId),
           SUBSCRIPTION_WARNING_SUPPRESSION_MS,
         );
       }
     }
+    this.#topicsStats = new Map(topicStats);
```
snosenzo, line 856: here too

snosenzo (commented):
Not sure these extra `new Map(...)`'s are necessary

defunctzombie:
> Foxglove weboscket: Avoid emitting unnecessary parameter & topic stats changes

FWIW - This would not be something actionable for a user to understand for a changelog. Might be better worded as "small performance improvement" or something generic like that to lump into a grab-bag changelog item of "performance improvements"

packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
achim-k: Not all members are emitted via `emitState`, we only have to re-assign the member if it is emitted

snosenzo (approved):

PR #6537, opened 2023-07-27
Preserve distinctly labeled chart segments during downsampling

foxymiles:
**User-Facing Changes**
Fix an issue with ordering of state transitions when zooming out on the state transitions panel.

**Description**
We downsample the data in the state transitions panel but this can throw away points that represent state changes. This changes the downsampling logic to consider a point with a new label as a distinct segment in the downsampled data.

If we don't want to make this a universal change we could also pass it down as an option to the downsampler through TimeBasedChart.

### DIFF
--- a/packages/studio-base/src/components/TimeBasedChart/downsample.ts
+++ b/packages/studio-base/src/components/TimeBasedChart/downsample.ts
@@ -90,8 +90,12 @@ export function downsampleTimeseries(
     const x = Math.trunc(datum.x * pixelPerXValue);
     const y = Math.trunc(datum.y * pixelPerYValue);
 
-    // interval has ended, we determine whether to write additional points for min/max/last
-    if (intFirst?.xPixel !== x) {
+    // interval has ended, we determine whether to write additional points for min/max/last. Always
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/components/TimeBasedChart/downsample.ts
```diff
@@ -90,8 +90,12 @@ export function downsampleTimeseries(
     const x = Math.trunc(datum.x * pixelPerXValue);
     const y = Math.trunc(datum.y * pixelPerYValue);
 
-    // interval has ended, we determine whether to write additional points for min/max/last
-    if (intFirst?.xPixel !== x) {
+    // interval has ended, we determine whether to write additional points for min/max/last. Always
+    // create a new interval when encountering a new label.
```
defunctzombie, line 94:
```suggestion
    // create a new interval when encountering a new label to preserve the transition from one label to another
```

defunctzombie (approved):
Since we already have a condition here I think this fix seems reasonable. The TimeBasedChart serves two panels in our codebase so until that changes 👍

PR #6538, opened 2023-07-27
Allow custom layer URDFs to subscribe to a topic

achim-k:
**User-Facing Changes**
Allow custom layer URDFs to subscribe to a topic

**Description**
- Allows custom layer URDFs to subscribe to a topic. This was not possible so far, and only the `/robot_description` topic could be visualized.
- Adds a select box for selecting the custom layer source (`url`, `filePath`, `param`, `topic`)
  - The `filePath` option is only available on the desktop app

[Screencast from 27.07.2023 18:58:15.webm](https://github.com/foxglove/studio/assets/9250155/1ab285f2-0c8d-4162-a999-f627570a4595)

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
@@ -174,6 +174,7 @@ export class Urdfs extends SceneExtension<UrdfRenderable> {
   #transformsByInstanceId = new Map<string, TransformData[]>();
   #jointStates = new Map<string, JointPosition>();
   #textDecoder = new TextDecoder();
+  #urdfsByTopic = new Map<string, string>();
 
   public constructor(renderer: IRenderer) {
     super("foxglove.Urdfs", renderer);
@@ -306,7 +307,7 @@ export class Urdfs extends SceneExtension<UrdfRenderable> {
             value: config.sourceType ?? DEFAULT_CUSTOM_SETTINGS.sourceType,
             options: [
               {
-                label: "From URL",
+                label: "URL",
                 value: "url",
               },
               {
@@ -375,6 +376,7 @@ export class Urdfs extends SceneExtension<UrdfRenderable> {
             label: "Frame prefix",
             input: "string",
             help: "Prefix to apply to all frame names (also often called tfPrefix)",
+            placeholder: "Frame prefix",
             value: config.framePrefix ?? "",
           },
           displayMode: {
@@ -524,7 +526,8 @@ export class Urdfs extends SceneExtension<UrdfRenderable> {
         });
 
         // Add the URDF renderable
-        this.#loadUrdf({ instanceId: newInstanceId, urdf: undefined });
+        const renderable = this.renderables.get(instanceId);
+        this.#loadUrdf({ instanceId: newInstanceId, urdf: renderable?.userData.urdf });
 
         // Update the settings tree
         this.updateSettingsTree();
@@ -581,11 +584,28 @@ export class Urdfs extends SceneExtension<UrdfRenderable> {
         this.#debouncedLoadUrdf({ instanceId, urdf: undefined });
       } else if (field === "parameter") {
         urdf = this.renderer.parameters?.get(action.payload.value as string) as string | undefined;
-        this.#debouncedLoadUrdf({ instanceId, urdf });
+        this.#debouncedLoadUrdf({ instanceId, urdf, forceReload: true });
       } else if (field === "framePrefix") {
         this.#debouncedLoadUrdf({ instanceId, urdf, forceReload: true });
-      } else if (field === "displayMode") {
+      } else if (field === "displayMode" || field === "visible") {
         this.#loadUrdf({ instanceId, urdf, forceReload: true });
+      } else if (field === "sourceType") {
+        const sourceType = action.payload.value as LayerSettingsCustomUrdf["sourceType"];
+        if (sourceType === "topic") {
+          urdf = renderable?.userData.topic
+            ? this.#urdfsByTopic.get(renderable.userData.topic)
+            : undefined;
+        } else if (sourceType === "param") {
+          urdf = renderable?.userData.parameter
+            ? (this.renderer.parameters?.get(renderable.userData.parameter) as string | undefined)
+            : undefined;
+        } else {
+          urdf = undefined;
+        }
+        this.#loadUrdf({ instanceId, urdf, forceReload: true });
+      } else if (field === "topic") {
+        urdf = this.#urdfsByTopic.get(action.payload.value as string);
+        this.#loadUrdf({ instanceId, urdf });
       } else {
         this.#loadUrdf({ instanceId, urdf });
       }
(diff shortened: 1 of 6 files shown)


packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
```diff
@@ -77,7 +77,10 @@ export type LayerSettingsUrdf = BaseSettings & {
 
 export type LayerSettingsCustomUrdf = CustomLayerSettings & {
   layerId: "foxglove.Urdf";
+  sourceType: "url" | "param" | "topic";
   url: string;
+  parameter: string;
```
defunctzombie, line 82: Should these be optional? `parameter?: string`? And should `url` become optional?

packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
```diff
@@ -275,12 +293,59 @@ export class Urdfs extends SceneExtension<UrdfRenderable> {
...
+                    .map((t) => t.name),
+                }
+              : undefined,
+          parameter:
+            config.sourceType === "param"
+              ? {
+                  label: "Parameter",
+                  input: "autocomplete",
+                  value: config.parameter ?? "",
+                  items: Array.from(this.renderer.parameters ?? [])
```
defunctzombie, line 344: We have a function called `filterMap` that will do the map and filter in one pass over the array.

packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
```diff
@@ -275,12 +293,59 @@ export class Urdfs extends SceneExtension<UrdfRenderable> {
...
+                  value: config.url ?? "",
+                }
+              : undefined,
+          topic:
+            config.sourceType === "topic"
+              ? {
+                  label: "Topic",
+                  input: "autocomplete",
+                  value: config.topic ?? "",
+                  items: Array.from(this.renderer.topics ?? [])
```
defunctzombie, line 333: We have a function called `filterMap` that will do the filter and map with one pass over the array.

packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
```diff
@@ -275,12 +293,59 @@ export class Urdfs extends SceneExtension<UrdfRenderable> {
...
+                    .filter((t) => URDF_TOPIC_SCHEMAS.has(t.schemaName))
+                    .map((t) => t.name),
```
defunctzombie, line 335: nit: I don't like single letter variable names unless we are doing math

packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
```diff
@@ -275,12 +293,59 @@ export class Urdfs extends SceneExtension<UrdfRenderable> {
...
+              {
+                label: "Topic",
+                value: "topic",
+              },
+            ],
+          },
+          url:
+            config.sourceType === "url"
+              ? {
+                  label: "URL",
```
defunctzombie, line 320: Do you think `URL` will be sufficiently obvious for folks using files and needing to do `file://` or should be separate that out as a dropdown option and let you enter a file path while handling turning it into a file:// url separately?

Is this where we might let you select asset? Or do we still expect urls for that as well?

achim-k: > Do you think `URL` will be sufficiently obvious for folks using files and needing to do `file://` or should be separate that out as a dropdown option and let you enter a file path while handling turning it into a file:// url separately?

Sounds like a good idea, I will do that

> Is this where we might let you select asset? Or do we still expect urls for that as well?

For now we still expect an URL as there is currently no way to list available assets

packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
```diff
@@ -653,31 +793,66 @@ export class Urdfs extends SceneExtension<UrdfRenderable> {
...
+        return;
+      } else if (sourceType === "filePath" && filePath != undefined) {
+        // Fetch the URDF from a local file
+        this.#fetchUrdf(instanceId, `file://${filePath}`);
+        return;
+      }
+    }
+
+    // At this point we have to have a URDF.
+    if (!urdf) {
```
snosenzo, line 831: any way we can consolidate and early return with these `!urdf` checks?

packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
```diff
@@ -653,31 +793,66 @@ export class Urdfs extends SceneExtension<UrdfRenderable> {
...
-    if (!urdf) {
+    if (!urdf || forceReload) {
       renderable.removeChildren();
+    }
 
-      // Fetch the URDF from the URL if we have one
-      if (url != undefined) {
+    if (!urdf) {
+      if (sourceType === "url" && url != undefined) {
+        // Fetch the URDF from the URL
```
snosenzo, line 820: nit: these comments (for url and filepath) don't really add anything imo

snosenzo (approved):
would be good to put errors on the urdf settings for the ones that are only enabled in desktop mode. 

Also we need a docs update for this right?

defunctzombie:
> Also we need a docs update for this right?

Documentation issue: https://linear.app/foxglove/issue/FG-4278/documentation-for-urdf-over-websockets-and-dropdown

defunctzombie:
Good stuff 👍 I played around with this and have some feedback and some things that didn't work:

---

I'd change this to "URL" and drop "From". "From" is implied by the "Source" label.
image

---

When I had selected "Topic" and entered a topic that had a bad URDF string I got a URDF error as expected; however, when I switched the type to "Parameter" I now somehow have two errors:
image

---
When I set a topic that is valid and the URDF loaded I then tried typing in an invalid topic name. I was able to type the name but the typical URDF error of "invalid topic" did not show up. When I refreshed the page the invalid topic was there and the error was there.

---

Probably separate PR feedback but I find the "Frame prefix" item with a blank box somehow confusing. Maybe it needs some placeholder so I realize it will adopt a default.
image

amacneil:
LR: 👍

packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
```diff
@@ -275,20 +298,89 @@ export class Urdfs extends SceneExtension<UrdfRenderable> {
...
+            value: config.sourceType ?? DEFAULT_CUSTOM_SETTINGS.sourceType,
+            options: [
+              {
+                label: "From URL",
+                value: "url",
+              },
+              {
+                label: "File Path",
+                value: "filePath",
+                disabled: !isDesktopApp(),
```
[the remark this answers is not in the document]
snosenzo: Was there a conclusion reached about whether to put errors on invalid fields? I don't see them added yet

defunctzombie (approved):

jtbandes:
This PR title says "subscribe to a topic" but the demo video shows using a Parameter 🤔

PR #6543, opened 2023-07-28
[WIP] Video support (rev 2)

jhurliman:
Alternative to: https://github.com/foxglove/studio/pull/5806

This approach uses a message structure identical to foxglove.CompressedImage. It prompts the question of whether a new message definition is needed, since either foxglove.CompressedVideo or foxglove.CompressedImage can be used with no change to the Studio code.

Only H264 video is currently supported, using a format string of "h264". More codec support can be added in a straightforward way.

The bitstream parsing code does not do any allocations or copies, and does a very lightweight pass per-frame. The first-time initialization when the first keyframe is encountered is slightly more involved but should still be very fast.

---
To test this PR - write [`foxglove.CompressedVideo`](https://github.com/foxglove/schemas/pull/124) messages into an mcap file with format `h264`.


defunctzombie:
@jhurliman Do you have a sample file that works with this PR?

defunctzombie:
Here's a sample file I created that doesn't seem to work reliably when playing or seeking:
[compressed_video.mcap.zip](https://github.com/foxglove/studio/files/12208251/compressed_video.mcap.zip)

When I first play the file it starts to play correctly. After some time (10 seconds ish) playback goes blank and never shows me the picture again. Sometimes when I see a picture show up again after passing a message with a keyframe and other times I see an error: "Failed to execute 'reset' on 'VideoDecoder': Cannot call 'reset' on a closed codec.".

The mcap file was created using: https://github.com/foxglove/mcap-h264-testing

defunctzombie:
Closing in favor of: https://github.com/foxglove/studio/pull/6590

Thanks for putting this together!

PR #6545, opened 2023-07-28
Remove unneeded message item lookup cache

foxymiles:
**User-Facing Changes**
Performance improvement for plot panel on some datasets

**Description**
Remove the per-message caching in `useCachedGetMessagePathDataItems`. Now that we no longer re-process the same messages over and over again when generating plot datasets this cache is pure overhead.

Screenshot 2023-07-28 at 7 44 57 AM


defunctzombie (approved):
🔪 love to see it

PR #6546, opened 2023-07-28
Memoize enumValuesByDatatypeAndField in RawMessage panel

defunctzombie:
**User-Facing Changes**
None

**Description**
This function only needs calling when the datatypes change.


foxymiles:
Do you want to do this in addition to https://github.com/foxglove/studio/pull/6544?

defunctzombie:
> Do you want to do this in addition to #6544?

Yes - I think its more obvious what is happening. I continue to dislike "hidden" memoization.

foxymiles (approved):
I think this is redundant given our other memoization but it shouldn't cause any issues.

PR #6547, opened 2023-07-28
Smarter expansion logic in Raw Message panel

cfoust:
**User-Facing Changes**
[FG-4165] More intuitive node expansion in the Raw Message panel

**Description**
This one is tricky. The expansion logic in the Raw Message panel had some strange behavior. User actions would lead to unexpected results. Namely:

Scenario 1:
1. Collapse all nodes.
2. Expand one root node by clicking on it.
3. **Result:** All nodes expanded.

The inverse was also true.

Scenario 2:
1. Expand all nodes.
2. Collapse one root node by clicking on it.
3. **Result:** All nodes collapsed.

We fix this by cleaning up the transition from "all collapsed" and "all expanded" to the user-defined state by applying the previous state to all nodes when the user toggles a node when the tree is in one of those states.

This is much easier to understand visually:

https://github.com/foxglove/studio/assets/4588553/7ccfc156-089c-4d0e-82d1-f8b1a9deea54

### DIFF
--- a/packages/studio-base/src/panels/RawMessages/index.stories.tsx
+++ b/packages/studio-base/src/panels/RawMessages/index.stories.tsx
@@ -26,7 +26,8 @@ import {
   topicsWithIdsToDiffFixture,
   withMissingData,
 } from "./fixture";
-import { Constants, RawMessagesPanelConfig } from "./types";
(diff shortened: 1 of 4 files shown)


foxymiles (approved):
Makes sense. Thanks for the fix!

PR #6554, opened 2023-07-30
Hoist memoization handling to react components

defunctzombie:
**User-Facing Changes**
None

**Description**


foxymiles (approved):

packages/den/monads/unwrap.ts
```diff
@@ -0,0 +1,22 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+// eslint-disable-next-line no-restricted-syntax
+type Null = null;
+
+/**
+ * Convenience function to check that _val_ is defined and return the value if it is defined. Throw
+ * if it is undefined or null.
```
snosenzo, line 10: I think you need to update this comment or update the if statement below

packages/studio-base/src/util/enums.ts
```diff
@@ -0,0 +1,117 @@
...
+// Foxglove Studio enum annotations are of the form: "Foo__foxglove_enum" (notice double underscore)
+// This method returns type name from "Foo" or undefined name doesn't match this format
+export function extractTypeFromStudioEnumAnnotation(name: string): string | undefined {
+  const match = extractTypeFromStudioEnumAnnotationRegex.exec(name);
+  return match?.[1];
+}
+
+// Returns a nested record of the form {datatype -> {field -> {value -> name}}}.
+//
+// We need memoizeWeak here because this function is called by multiple callers, each with their own
```
snosenzo, line 53: no longer using `memoizeWeak` update comment

packages/den/monads/unwrap.ts
defunctzombie: I don't follow? The if statement _does_ check if the value is undefined or null? `value == undefined` will be true if `val` is `undefined` or `null`.

defunctzombie: I added tests to ensure the behavior does not regress.

snosenzo: 🤦 forgetting and relearning javascript quirks. Thanks for reminding me

snosenzo (approved):

PR #6555, opened 2023-07-31
Fix rerenders on PlaybackTimeDisplay

foxymiles:
**User-Facing Changes**
None (minor performance enhancement)

**Description**
PlaybackTimeDisplay selects too much from the message pipeline, selecting all of `activeState`. This causes it to re-render more than necessary when the fields it needs aren't changing:

Screenshot 2023-07-31 at 6 53 34 AM

Using more specific selectors for the fields it needs eliminates this.

### DIFF
--- a/packages/studio-base/src/components/PlaybackControls/PlaybackTimeDisplay.tsx
+++ b/packages/studio-base/src/components/PlaybackControls/PlaybackTimeDisplay.tsx
@@ -26,13 +26,18 @@ type Props = {
   onPause: () => void;
 };
 
-const selectActiveData = (ctx: MessagePipelineContext) => ctx.playerState.activeData;
+const selectIsPlaying = (ctx: MessagePipelineContext) => ctx.playerState.activeData?.isPlaying;
+const selectStartTime = (ctx: MessagePipelineContext) => ctx.playerState.activeData?.startTime;
+const selectEndTime = (ctx: MessagePipelineContext) => ctx.playerState.activeData?.endTime;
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/components/PlaybackControls/PlaybackTimeDisplay.tsx
```diff
@@ -26,13 +26,18 @@ type Props = {
   onPause: () => void;
 };
 
-const selectActiveData = (ctx: MessagePipelineContext) => ctx.playerState.activeData;
+const selectIsPlaying = (ctx: MessagePipelineContext) => ctx.playerState.activeData?.isPlaying;
```
snosenzo, line 29: All of this boilerplate is unfortunate :/
Is there no better way around this?

snosenzo (approved):

packages/studio-base/src/components/PlaybackControls/PlaybackTimeDisplay.tsx
foxymiles: We could also select a subset of the message and use a `shallow` comparison on the selector. We'd have to plumb that through `useMessagePipeline` to the underlying zustand selector.

foxymiles: Yeah this is one downside of the store/selector model. We could probably use shallow selectors in more places although anything that looks at currentTime or other fields that update at high frequency needs to be very deliberate about how it accesses the message pipeline store especially since it seems like a lot of MUI components are pretty expensive to render.

PR #6558, opened 2023-07-31
Use binary search for merging {all,current}Frames in plot data

cfoust:
**User-Facing Changes**
Use a more efficient strategy for incorporating blocks of plot messages.

**Description**
Before, filtering old data was O(n); now we do a binary search to decide what data to keep.

I'm not sure whether this is the best solution, but it _greatly_ decreased this operation's contributions to slow rendering time (from several hundred milliseconds almost every frame to <1ms.)

### DIFF
--- a/packages/studio-base/src/panels/Plot/usePlotPanelData.ts
+++ b/packages/studio-base/src/panels/Plot/usePlotPanelData.ts
@@ -9,7 +9,12 @@ import { useLatest } from "react-use";
 
 import { filterMap } from "@foxglove/den/collection";
 import { useShallowMemo } from "@foxglove/hooks";
-import { isGreaterThan, isLessThan, isTimeInRangeInclusive, subtract } from "@foxglove/rostime";
+import {
+  isLessThan,
+  isTimeInRangeInclusive,
+  compare as compareTime,
+  subtract,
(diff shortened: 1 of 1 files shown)


defunctzombie:
This is an appropriate change for what is happening with `currentFrame` - but _why_ is `currentFrame` so large that this was necessary?

foxymiles (approved):
Makes sense. Good find!

Can we use the sortedIndexBy stuff from lodash instead of writing our own search?

cfoust:
interesting, it appears to be undocumented but it actually does use binary search: https://github.com/lodash/lodash/blob/4.17.15/lodash.js#L4131
I had looked into whether lodash already had one, but their documentation did not mention it

foxymiles:
They do document it here: https://lodash.com/docs/4.17.15#sortedIndex

But yeah they should really call this out more in the docs.

cfoust:
I'm new to lodash (I prefer ramda) but it was unclear whether `sortedIndexBy` had the same semantics as `sortedIndex`; in my experience the `*By` can change behavior in unexpected ways

foxymiles:
I prefer ramda too. We use it in console.

cfoust:
made unnecessary by pending plot pipeline changes

Issue #6275, opened 2023-06-21
Unable to Select Colour Using Keyboard


camchambers:
**Description**
Typing a hex colour in the colour picker and pressing enter does not select the colour. 

![color-selection](https://github.com/foxglove/studio/assets/26011644/cd06649c-1be1-493e-8d2f-ed8e9e2668be)

- Version:  1.54.0
- Platform: Ubuntu Linux
- Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native): rosbridge

**Steps To Reproduce**
- Configure a topic containing a colour picker
- Enter a hex value into the colour field
- Press enter

**Expected Behavior**
- The colour entered is updated in the UI

jtbandes:
Hi, I believe this is happening because the color picker requires 8 digits, `#RRGGBBAA`. However, I agree that it should be possible to enter 6 values and automatically set the alpha to FF.

camchambers:
Right. The field requires an alpha value (suggested by the placeholder value). Defaulting the alpha would be a nice feature.

foxymiles:
I'm going to reclassify this as an enhancement request.

camchambers:
Absolutely.

PR #6575, opened 2023-08-01
Allow more flexible color entry for ColorPicker settings item

defunctzombie:
**User-Facing Changes**
Allow more color string input variants for ColorPicker panel setting input type.

**Description**
Prior to this change, a user had to entry all of the hex digits into the color picker text input before the picker would accept their color. For RGB colors this meant entering all 6 hex characters for `rrggbb` and, for RGBA, 8 hex characters `rrggbbaa`.  When using RGBA it is common to want no opacity and default the last two characters to `ff`.

This change updates the hex text input to accept any of the following for RGB inputs: `rrggbb` and the common shorthand `rgb` which expands to repeat the characters -> `rrggbb`.

And for RGBA inputs this change updates the logic to accept any of the valid RGB inputs (defaulting alpha to `FF`) and also supports: `rrggbbaa` and the shorthand `rgba`.

As long as the user is editing the text input field they continue to see the string they input (if using the shorthand form) while the actual setting value is expanded to the long form. When the user exits the color picker popup and re-opens it they will see the long form of their color in the text field.

Updates the font style on the display field to use monospace. This makes the hex display visually match the input entry.

Fixes: #6275

### DIFF
--- a/packages/studio-base/src/components/SettingsTreeEditor/inputs/ColorPickerControl.tsx
+++ b/packages/studio-base/src/components/SettingsTreeEditor/inputs/ColorPickerControl.tsx
@@ -44,8 +44,59 @@ type ColorPickerInputProps = {
   onEnterKey?: () => void;
 };
 
-const hexMatcher = /^#?([0-9A-F]{3,8})$/i;
+export function ColorPickerControl(props: ColorPickerInputProps): JSX.Element {
+  const { alphaType, onChange, value, onEnterKey } = props;
 
+  const { classes } = useStyles();
+
+  const {
+    swatchColor,
+    updatePrefixedColor,
+    editedValueIsInvalid,
+    editedValue,
+    updateEditedValue,
+    onInputBlur,
+  } = useColorPickerControl({
+    alphaType,
+    onChange,
+    value,
+  });
+
+  return (
+    <Stack className={classes.container} gap={1}>
+      {alphaType === "alpha" ? (
+        <HexAlphaColorPicker
+          className={classes.picker}
+          color={swatchColor}
+          onChange={updatePrefixedColor}
+        />
+      ) : (
+        <HexColorPicker
+          className={classes.picker}
+          color={swatchColor}
+          onChange={updatePrefixedColor}
+        />
+      )}
+      <TextField
+        size="small"
+        error={editedValueIsInvalid}
+        InputProps={{
+          onFocus: (event) => event.target.select(),
+          role: "input",
+          startAdornment: <TagIcon fontSize="small" />,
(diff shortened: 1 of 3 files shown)


defunctzombie:
@2metres please take a look at the visual diff and 👍 if you are good with this text field using monospace font while other text fields use the regular font.

packages/studio-base/src/components/SettingsTreeEditor/inputs/ColorPickerControl.tsx
```diff
@@ -54,46 +61,55 @@ export function ColorPickerControl(props: ColorPickerInputProps): JSX.Element {
...
 
+      // if it is a valid color then we can emit the new value
       if (isValidHexColor(newValue, alphaType)) {
         onChange(`#${newValue}`);
       }
     },
     [alphaType, onChange],
   );
 
-  useEffect(() => {
```
foxymiles, line 80: I think we need this if the value is updated by the source after we receive the first value.

foxymiles: Yeah that's why I haven't done a PR on this myself yet. I think if you want to do this you need to add some on-blur logic or something so that the edited value doesn't get stuck there forever.

snosenzo (commented):
Should we have a story for the acceptance of these new values and their expansion to their longer hex counterparts?

packages/studio-base/src/components/SettingsTreeEditor/inputs/ColorPickerControl.tsx
defunctzombie: 👍 I'll add onBlur - fwiw the edited value is not stuck forever. When the user hits enter or closes the color picker the component is re-created next time the picker is opened and the actual stored value is shown.

defunctzombie:
> Should we have a story for the acceptance of these new values and their expansion to their longer hex counterparts?

I'll take a stab at a hook for the logic so we can avoid a screenshot story for it since that might not test the logic so well.

defunctzombie:
I've added `onBlur` handling to update the input field value to the latest `value` prop when the input loses focus.

I've added tests for the behavior we expect by writing a hook for the business logic of the component and testing the hook. I don't know how we feel about this style of testing vs storybooks; we don't use this style a lot but I think it lets you test more scenarios faster and has easier to reason about assertions. Stories are a bit tricky to do that kind of asserting. This testing could be done with stories though if we think that is better at capturing the true behavior that a user will experience.

foxymiles (approved):

amacneil:
LR: 👌

PR #6576, opened 2023-08-02
Extract parsedPaths into its own memo

defunctzombie:
**User-Facing Changes**
None

**Description**
useCachedGetMessagePathDataItems had two places where it was parsing the paths. This change hoists this parsing to a single memo and uses the result in the two other memos.


defunctzombie:
No specific reason brought me to this change other than it being a tiny thing I noticed while reviewing the reasoning behind: https://github.com/foxglove/studio/pull/6378

foxymiles (approved):

PR #6577, opened 2023-08-02
Improve interactivity of plot series path editing

foxymiles:
**User-Facing Changes**
Improve interactivity of plot series path editing.

**Description**
Currently when a user is actively editing a series path in the plot panel the paths that are inputs to `usePlotPanelData` are rapidly changing and this causes a complete rebuild of the datasets on each keystroke which can be very slow on larger datasets. This patch tries to eliminate most of that work in the cases where it's not useful by filtering out changes to series paths that will not result in any new messages being processed and included in the dataset.

This involves pre-filtering paths that either are not valid ros paths or that parse to a topic that is not present in our dataset. And the logic that builds datasets from these filtered paths tries to eliminate any changes that wouldn't result in any new messages.

### DIFF
--- a/packages/studio-base/src/panels/Plot/index.tsx
+++ b/packages/studio-base/src/panels/Plot/index.tsx
@@ -58,7 +58,7 @@ export type { PlotConfig } from "./types";
 const defaultSidebarDimension = 240;
 
 const EmptyDataSets: Immutable<DataSet[]> = Object.freeze([]);
-const EmtpyDataSet: Immutable<DataSet> = Object.freeze({ data: [] });
+const EmtpyDataSet: Immutable<DataSet> = Object.freeze({ data: [], label: "" });
 
 export function openSiblingPlotPanel(openSiblingPanel: OpenSiblingPanel, topicName: string): void {
   openSiblingPanel({
--- a/packages/studio-base/src/panels/Plot/index.tsx
+++ b/packages/studio-base/src/panels/Plot/index.tsx
@@ -16,6 +16,7 @@ import { ComponentProps, useCallback, useEffect, useMemo, useState } from "react
 import { useLatest } from "react-use";
 import { DeepWritable } from "ts-essentials";
 
+import { useShallowMemo } from "@foxglove/hooks";
 import {
   Time,
   add as addTimes,
@@ -23,6 +24,7 @@ import {
   subtract as subtractTimes,
   toSec,
 } from "@foxglove/rostime";
+import { Immutable } from "@foxglove/studio";
 import {
   MessagePipelineContext,
   useMessagePipeline,
@@ -55,7 +57,8 @@ export type { PlotConfig } from "./types";
 
 const defaultSidebarDimension = 240;
 
-const EmptyDatasets: DataSet[] = [];
+const EmptyDataSets: Immutable<DataSet[]> = Object.freeze([]);
+const EmtpyDataSet: Immutable<DataSet> = Object.freeze({ data: [] });
 
 export function openSiblingPlotPanel(openSiblingPanel: OpenSiblingPanel, topicName: string): void {
   openSiblingPanel({
@@ -240,6 +243,13 @@ function Plot(props: Props) {
 
   const onClickPath = useCallback((index: number) => setFocusedPath(["paths", String(index)]), []);
 
+  // Stabilize datasets we send to the chart by first replacing empty datasets with a constant value
+  // and then shallow memoizing the whole set. This prevents a lot of unnecessary and expensive
+  // rendering while the user is interactively editing paths that don't result in any messages.
+  const stableDatasets = useShallowMemo(
+    datasets.map((ds) => (ds.data.length === 0 ? EmtpyDataSet : ds)),
+  );
+
   return (
     <Stack
       flex="auto"
@@ -259,7 +269,7 @@ function Plot(props: Props) {
         {legendDisplay !== "none" && (
           <PlotLegend
             currentTime={showPlotValuesInLegend ? currentTimeSinceStart : undefined}
-            datasets={showPlotValuesInLegend ? datasets : EmptyDatasets}
+            datasets={showPlotValuesInLegend ? datasets : EmptyDataSets}
             legendDisplay={legendDisplay}
             onClickPath={onClickPath}
             paths={yAxisPaths}
@@ -274,7 +284,7 @@ function Plot(props: Props) {
           <PlotChart
             currentTime={currentTimeSinceStart}
             datasetBounds={datasetBounds}
-            datasets={castWritable(datasets)}
+            datasets={castWritable(stableDatasets)}
             defaultView={defaultView}
             isSynced={xAxisVal === "timestamp" && isSynced}
             maxYValue={parseFloat((maxYValue ?? "").toString())}
--- a/packages/studio-base/src/panels/Plot/index.tsx
+++ b/packages/studio-base/src/panels/Plot/index.tsx
@@ -11,7 +11,7 @@
 //   found at http://www.apache.org/licenses/LICENSE-2.0
 //   You may not use this file except in compliance with the License.
 
-import { compact, isNumber, uniq } from "lodash";
+import { isNumber, uniq } from "lodash";
 import { ComponentProps, useCallback, useEffect, useMemo, useState } from "react";
 import { useLatest } from "react-use";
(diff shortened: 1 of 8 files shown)


defunctzombie:
When I hover over the sample dataset plot value I see this strange tooltip:
image

Notably, this does not happen in the other plot panel in the sample dataset. However, I do not see the same behavior on main...

packages/studio-base/src/panels/Plot/index.tsx
```diff
@@ -240,6 +238,13 @@ function Plot(props: Props) {
 
   const onClickPath = useCallback((index: number) => setFocusedPath(["paths", String(index)]), []);
 
+  // Stabilize datasets we send to the chart by first replacing empty datasets with a constant value
+  // and then shallow memoizing the whole set. This prevents a lot of unnecessary and expensive
+  // rendering while the user is interactively editing paths that don't result in any messages.
+  const stableDatasets = useShallowMemo(
```
defunctzombie, line 244: Are we not able to stabilize this upstream in the hook?

defunctzombie:
I think `PlotDataByPath` is now unused from `internalTypes.ts`.

packages/studio-base/src/panels/Plot/useAllFramesByTopic.ts
```diff
@@ -128,5 +131,8 @@ export function useAllFramesByTopic(
     });
   }
 
-  return state.messages;
+  // Stablize the flattened messages by shallow memoing the whole set after excluding empty topics.
+  const stableMessagesWithData = useShallowMemo(pickBy(state.messages, (v) => v.length > 0));
```
defunctzombie, line 135: Lets use a better variable name than _v_?

packages/studio-base/src/panels/Plot/useStableValidPathsForDatasourceTopics.ts
```diff
@@ -0,0 +1,40 @@
...
+import { useShallowMemo } from "@foxglove/hooks";
+import { Immutable } from "@foxglove/studio";
+import parseRosPath from "@foxglove/studio-base/components/MessagePathSyntax/parseRosPath";
+import {
+  MessagePipelineContext,
+  useMessagePipeline,
+} from "@foxglove/studio-base/components/MessagePipeline";
+import { PlotPath } from "@foxglove/studio-base/panels/Plot/internalTypes";
+
+export const selectKeyedTopics = createSelector(
```
defunctzombie, line 17: Why is this exported?

What is special about `createSelector` over a regular selector?

packages/studio-base/src/panels/Plot/useStableValidPathsForDatasourceTopics.ts
```diff
@@ -0,0 +1,40 @@
...
+ * Returns a referentially stable list of paths that parse as a valid path and reference a topic
+ * that exists in our datasource.
+ */
+export function useStableValidPathsForDatasourceTopics(
+  paths: Immutable<PlotPath[]>,
+): Immutable<PlotPath[]> {
+  const keyedTopics = useMessagePipeline(selectKeyedTopics);
+
+  const stableValidPaths = useShallowMemo(
+    filterMap(paths, (path) => {
```
defunctzombie, line 32: This filterMap should be in a memo - otherwise it is going to run every time render runs to parse the paths. We render during playback even if paths are not changing

packages/studio-base/src/panels/Plot/usePlotPanelData.ts
```diff
@@ -166,58 +187,86 @@ export function usePlotPanelData(params: Params): Immutable<{
...
-    xAxisPath !== state.xAxisPath ||
+    xAxisPathAsFullPlotPath !== state.xAxisPath ||
     globalVariables !== state.globalVariables;
 
-  if (allFrames !== state.allFrames || resetDatasets) {
-    // use setState directly instead of useEffect to skip an extra render.
+  if (allFrames !== state.allFrames || validAllPaths !== state.allPaths || resetDatasets) {
+    // Derive a new state based on the old state & new message data. Note that we maintain a
+    // separate cursor into allFrames for each path. This is necessary because as new series are
+    // added we need to read all available messages even if the new series matches the path of an
```
defunctzombie, line 204: Why do we need to read all available messages?

packages/studio-base/src/panels/Plot/usePlotPanelData.ts
```diff
@@ -166,58 +187,86 @@ export function usePlotPanelData(params: Params): Immutable<{
...
+    // existing series.
+    //
+    // We try here to provide a minimal update for downstream referential integrity purposes.
+    // Paths that recieve no new messages should be represented unchanged in the new state.
+    //
+    // Uses setState directly instead of useEffect to skip an extra render.
```
defunctzombie, line 210: I don't think I ever fully understood how this doesn't result in constantly re-rendering since every call to `setState` is making a new object for the state.

defunctzombie (commented):
What's a good way to test this behavior so we don't regress it again?

packages/studio-base/src/panels/Plot/index.tsx
foxymiles: No because the labels are embedded in the dataset so we need to include the changed dataset with the new label even if they don't have any data.

packages/studio-base/src/panels/Plot/useStableValidPathsForDatasourceTopics.ts
foxymiles: Actually it doesn't need to be exported.

`createSelector` creates a memoized selector so all panels can benefit from the same memoization.

packages/studio-base/src/panels/Plot/useStableValidPathsForDatasourceTopics.ts
foxymiles: This is a short list of stuff and `parseRosPath` is already memoized so I doubt this is going to be much slower than a memoized version but I can add an extra memo here.

packages/studio-base/src/panels/Plot/usePlotPanelData.ts
foxymiles: Imagine this scenario:
1. You load the entire dataset with a plot of `/imu.x`, your cursor for `/imu.x` is pointing to the end of allframes for `/imu`.
2. You add a new series also with `/imu.x`. 

We can't use the previous cursor for this new series, we need a new one that points to the start of `/imu` so we can load the whole series and generate a complete dataset for this new series.

packages/studio-base/src/panels/Plot/usePlotPanelData.ts
foxymiles: Only if the conditions change and force it to generate a new state, otherwise this gets skipped.

foxymiles:
> I think `PlotDataByPath` is now unused from `internalTypes.ts`.

Yeah you're right I wonder why the linter didn't catch that.

foxymiles:
> What's a good way to test this behavior so we don't regress it again?

I can add some more hook tests with incremental inputs.

foxymiles:
> When I hover over the sample dataset plot value I see this strange tooltip

Good catch. Looks like the downsampler was throwing away the empty dataset so the indices were off.

packages/studio-base/src/panels/Plot/usePlotPanelData.ts
defunctzombie: I'd add something like this to the comment. I'll get better at being explicit in my _why_ asks but typically if I'm asking _why_ that's an opportunity to enhance a comment that leaves me wanting more info.

defunctzombie (approved):
Tests for the behavior that I caught in the regression would be good or something that flags this more loudly than manual testing.

packages/studio-base/src/panels/Plot/usePlotPanelData.test.tsx
```diff
@@ -35,9 +33,9 @@ const testDataTypes: RosDatatypes = new Map(
   }),
 );
 
-const fixtureMessages1: MessageEvent[] = [mockMessage({ value: 1 })];
+// const fixtureMessages1: MessageEvent[] = [mockMessage({ value: 1 }, { topic: "topic_a" })];
```
defunctzombie, line 36: Did you mean to leave this commented out?

defunctzombie (approved):

PR #6583, opened 2023-08-03
Use @foxglove/schemas in user script templates and examples

foxymiles:
**User-Facing Changes**
Use @foxglove/schemas in user script templates and examples.

**Description**
Demonstrate use of the LocationFix type from @foxglove/schemas in our location fix template and update the comments in the standard template to differentiate between the `Message` types and @foxglove/schema types.

### DIFF
--- a/packages/studio-base/src/panels/NodePlayground/index.tsx
+++ b/packages/studio-base/src/panels/NodePlayground/index.tsx
@@ -58,8 +58,11 @@ import { Input, Message } from "./types";
 // Your script can output well-known message types, any of your custom message types, or
 // complete custom message types.
 //
-// Use \`Message\` to access your data source types or well-known types:
+// Use \`Message\` to access types from the schemas defined in your data source:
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/panels/NodePlayground/index.tsx
```diff
@@ -58,9 +58,12 @@ import { Input, Message } from "./types";
 // Your script can output well-known message types, any of your custom message types, or
 // complete custom message types.
 //
-// Use \`Message\` to access your data source types or well-known types:
+// Use \`Message\` to access types from the schemas defined in your data source:
 // type Twist = Message<"geometry_msgs/Twist">;
 //
+// Import from the @foxglove/schemas package to use standard schema types:
```
defunctzombie, line 64:
```suggestion
// Import from the @foxglove/schemas package to use foxglove schema types:
```

defunctzombie (approved):

PR #6584, opened 2023-08-03
Revamp of pipeline for generating plots

cfoust:
**User-Facing Changes**
Move all aspects of generating data for plot panels to a Web Worker and reduce plot generation memory usage.

**Description**
This diff is really scary, but it's divided up into two main sections:
* 70% is related to adding support for `Float32Array`-backed datasets to our charting components (namely: `Chart` and `TimeBasedChart`). This greatly reduces the memory overhead of plots with lots of data points.
* 30% is moving all dataset generation and downsampling to a Web Worker


jtbandes:
> * 70% is related to adding support for `Float32Array`-backed datasets to our charting components (namely: `Chart` and `TimeBasedChart`). This greatly reduces the memory overhead of plots with lots of data points.
> * 30% is moving all dataset generation and downsampling to a Web Worker


This sounds like it's two fairly separate features and should be (at least) two PRs, does that seem reasonable?

Re ramda: since we already make pretty significant use of lodash, I think it would be worth trying to stick with one rather than having both. It seems like most of the things ramda is being used for here could be replaced with either lodash functions or builtin methods (e.g. Array#flatMap). I'll admit I don't have personal experience with ramda but it seems like [one of the maintainers admits that lodash is better for performance](https://stackoverflow.com/a/71403954/23649), so might be best to continue on the lodash path.

cfoust:
`ramda` is already in our dependency tree (it was relied on by some descendant package, as far as I could discern) and Miles is also a fan, so I just went for it. The packages have different fundamental philosophies, as that StackOverflow comment implies, and ramda generally maps better onto functional programming.

As far as splitting goes, I could do it if you think it's worthwhile, though I suspect it won't make much difference in terms of reviewing time; even separate, they're still rather big diffs.

cfoust (commented):
Instead of just adjusting the description I decided to format this overview as a review--that way I can make comments inline on things that I think merit further explanation.

As mentioned in the PR description, this PR does two things:
1. **It adds support for `Float32Array`-backed plot datasets, greatly reducing the memory footprint of plot data.** Theoretically this could have been done in the old setup even without introducing a worker, but because of how I went about this (I realized how big of a problem memory usage was about halfway through) it would not be easy to just slice off this portion of the diff. For example, I completely wiped away the former method of aggregating data, `usePlotPanelData.tsx`, which makes a lot of assumptions about the shape of the data.
2. **It moves all data processing and downsampling for plots to a Web Worker.** Downsampling in the main thread imposed a significant burden on our rendering pipeline, which resulted in a fairly low upper bound on the amount of data we could handle. With this new approach, we send _all_ incoming messages on the relevant topics to the worker along with the visual area of the plots and let it decide when and how to update the datasets.

#1 is not very complicated, but may look strange at first glance. Plot datasets are now represented as `TypedData[]`, where `TypedData` is defined as follows:
```ts
export type TypedData = {
  x: Float32Array;
  y: Float32Array;
  value: (string | number | bigint | boolean | undefined)[];
  constantName?: string[];
};
```
It is important to notice that a plot dataset is _an array_ of `TypedData`. This is because appending a new element to a `Float32Array` is an expensive operation: once declared, for understandable reasons the size of a `TypedArray` cannot be expanded. Once we have built a range of datapoints, we want to modify it as little as possible. A single `TypedData` is typically equivalent in size to the data points extracted from a single block (in the case of block data) or from a sequence of just a few messages (in the case of current data).

As you can imagine, #2 is the real source of complexity in this PR, though the previous version (`usePlotPanelData.tsx`) was also wildly complex, mostly because of all of the hacks we had to employ to ensure that object references were as stable as possible--a common problem with data-intensive React code. There are two basic strategies we employ to render datasets depending on the data source we are using:
* **Recorded data**: As blocks are decoded, we send the relevant messages to the worker in their entirety and let it build new, renderable datasets at some cadence (as of writing, at most every 100ms.) When a plot's viewport changes as the result of a pan or zoom, we downsample and send a new dataset.
* **Live data**: We send new data to be rendered _as fast as possible without downsampling_ and then, every couple of seconds, cull, downsample, and replace the rendered dataset.

The latter strategy is questionable because at least as far as I am aware, we don't do any intentional data culling elsewhere in the app for live sources. For now, my approach throws away data points older than 30 seconds and any data points that bring the total for any one dataset above 10,000 points, which is an arbitrary number I chose. But frankly we don't have a choice: ChartJS is face-achingly slow for datasets beyond 10-20,000 points and we need to cull data to keep rendering fast and not run out of memory (though admittedly, without message slicing this PR still will, eventually, because of the cache the data source keeps in the main thread.)

Errata I noticed while doing this refactor:
* We need to be a lot more conscious about the kinds of things we build as React hooks versus pure functions, which compose better and can be used everywhere. A particularly egregious example is `useCachedGetMessagePathDataItems`, a hook with (in addition to a 30-character name) contains useful logic for accessing paths that use global variables, but its logic is extremely complex, all to avoid reference changes. To move things to the worker, which lacks anything resembling hooks, I had to reimplement much of the functionality we bury in hooks from scratch.
* Several components in the Plot panel still rely on having complete `PlotData`, such as the legend (for showing current values), which meant that in practice I had to send the downsampled dataset to the main thread twice, which is inconvenient.
* ChartJS is woefully slow. It is embarrassing that it drops below 60FPS with only a few thousand points. Our needs are simple; we really could accomplish everything we want on much larger datasets (hundreds of thousands (?) of points) at 60FPS with just a couple of WebGL draw calls.

packages/studio-base/src/panels/Plot/useDatasets.ts
```diff
@@ -0,0 +1,249 @@
...
+
+type Client = {
+  topics: readonly string[];
+  setter: (topics: string[]) => void;
+};
+let clients: Record<string, Client> = {};
+
+// Calculate the list of unique topics that _all_ of the plots need and
+// nominate one panel to subscribe to the topics on behalf of the rest.
+function chooseClient() {
```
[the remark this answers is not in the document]
cfoust: Update: on further investigation it seems that `useBlocksByTopic` is responsible for the new reference, at least for blocks (I hadn't looked into it too deeply.) Presumably relying on a raw `useMessagePipeline` instead would improve things, but not by much. It's probably worth another look, though it would not diminish the benefits of subscription pooling.

packages/studio-base/src/components/PanelContextMenu.tsx
```diff
@@ -36,6 +36,11 @@ type PanelContextMenuProps = {
   getItems: () => Immutable<PanelContextMenuItem[]>;
 };
 
+function isContextEvent(event: MouseEvent): boolean {
```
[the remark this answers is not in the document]
jtbandes: The PR is big enough – let's remove this and review it separately. We might want to explore actually deleting this custom mouse event handling (which I added in #6127) and just go back to using the contextmenu event. I added it when making the 3D-based image panel but it may not be necessary if there is no right-click-drag interaction in the image panel.

foxymiles:
Splitting an existing plot panel into a new panel doesn't seem to work:


https://github.com/foxglove/studio/assets/93935560/495e3a67-7505-4257-be9e-6916336df2a9

foxymiles:
Also there seems to be an issue with indexed x axis plots. Load this mcap and plot `/data.data[:].a` with x-axis set to `Index` and nothing renders.

[array multiple lines.mcap.zip](https://github.com/foxglove/studio/files/12420178/array.multiple.lines.mcap.zip)

foxymiles:
When connected to a rosbridge source from https://github.com/foxglove/ros1-bench-workspace and plotting `/sample1.data` nothing is rendered in the plot.

foxymiles:
The bounds calculation seems to have changed. If I create a layout with a plot and a state transition panel, both with the path `/diagnostics.status[:].values[:]{key=="brake_switch"}.value`, I get different default bounds. 

Before:
before

After:
after

foxymiles:
This layout when applied to nuscenes doesn't seem to render anything:

```
{"configById":{"Plot!1t9d2uf":{"paths":[{"value":"/CAM_BACK_LEFT/camera_info.height","enabled":true,"timestampMethod":"receiveTime"}],"showXAxisLabels":true,"showYAxisLabels":true,"showLegend":true,"legendDisplay":"floating","showPlotValuesInLegend":false,"isSynced":true,"xAxisVal":"custom","sidebarDimension":240,"xAxisPath":{"value":"/CAM_BACK_LEFT/camera_info.timestamp.sec"}}},"globalVariables":{},"userNodes":{},"playbackConfig":{"speed":1},"layout":"Plot!1t9d2uf"}
```

packages/studio-base/src/components/TimeBasedChart/useDownsampler.tsx
```diff
@@ -0,0 +1,99 @@
...
+  downsampler: Provider<ObjectData>;
+  setScales: (scales: RpcScales) => void;
+} {
+  const [view, setView] = React.useState<View | undefined>();
+  const [setter, setSetter] = React.useState<ProviderStateSetter<ObjectData> | undefined>();
+
+  const register = React.useCallback((newSetter: ProviderStateSetter<ObjectData>) => {
+    // lol
+    setSetter(newSetter);
+  }, []);
```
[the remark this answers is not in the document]
cfoust: I don't totally remember what I was doing here, but I used to perform some intermediate step on `newSetter` before passing it to `setSetter`; once I removed that bit, I never took another look at this

packages/studio-base/src/panels/Plot/plotData.ts
```diff
@@ -216,7 +256,95 @@ export function buildPlotData(
...
   };
 }
+
+export function resolvePath(
+  metadata: MetadataEnums,
+  messages: readonly MessageEvent[],
+  path: RosPath,
+): PlotDataItem[] {
+  const { structures, enumValues } = metadata;
+  const topics = R.pipe(
```
foxymiles, line 270: Can this be replaced by a single `indexBy`?

packages/studio-base/src/panels/Plot/useDatasets.ts
```diff
@@ -0,0 +1,249 @@
...
+      if (numClients === 0) {
+        worker?.terminate();
+        worker = service = undefined;
+      }
+    };
+  }, []);
+
+  useMetadata();
+
+  // Allow the worker direct access to `setState`.
```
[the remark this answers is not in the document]
cfoust: aha, the vagaries of comments. changed the code but not the comment

packages/studio-base/src/panels/Plot/PlotLegendRow.tsx
```diff
@@ -138,8 +138,8 @@ export function PlotLegendRow({
     const timeToCompare = hoverValue?.value ?? currentTime;
 
     let value;
-    for (const pt of correspondingData) {
-      if (timeToCompare == undefined || pt == undefined || pt.x > timeToCompare) {
+    for (const pt of iterateTyped(correspondingData as TypedData[])) {
```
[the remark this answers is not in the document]
cfoust: `iterateTyped` takes a mutable reference, which (though not ideal) is because of some technical debt I discovered, so this is just going from `readonly` to mutable

packages/studio-base/src/components/TimeBasedChart/useDownsampler.tsx
```diff
@@ -0,0 +1,99 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+//
+// This file incorporates work covered by the following copyright and
+// permission notice:
+//
+//   Copyright 2018-2021 Cruise LLC
+//
+//   This source code is licensed under the Apache License, Version 2.0,
+//   found at http://www.apache.org/licenses/LICENSE-2.0
+//   You may not use this file except in compliance with the License.
```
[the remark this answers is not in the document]
cfoust: for the new code I wrote, I just blindly copied in the first license header I could find in adjacent code and did not inspect it too closely, still not sure what our practices are on that

packages/studio-base/src/components/TimeBasedChart/downsample.ts
```diff
@@ -92,93 +87,90 @@ export function downsampleTimeseries(
...
 }
 
-export function downsampleScatter(dataset: ChartDataset, bounds: DownsampleBounds): ChartDataset {
-  // datasets of length 1 don't need downsampling
-  if (dataset.data.length <= 1) {
-    return dataset;
-  }
+export function downsampleScatter<T>(
+  iterate: (dataset: T) => Generator<Point>,
+  dataset: T,
```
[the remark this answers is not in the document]
cfoust: it's written this way to accommodate arbitrary data types, not just objects that satisfy the `Iterable` interface
in general this is a common pattern in FP. your function asks you to provide the functions it needs in order to manipulate your (arbitrary) data type, then the data type itself. that way the data type is completely up to you

packages/studio-base/src/components/Chart/datasets.ts
```diff
@@ -0,0 +1,116 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import * as R from "ramda";
+
+import { TypedData, ObjectData } from "./types";
+
+export type Point = { index: number; x: number; y: number; label: string | undefined };
```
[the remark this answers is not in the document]
cfoust: It is similar to but not the same as other `ObjectData`-adjacent types (like `Datum`), since it has fewer fields and only exists so that we have a consistent interface for mapping over plot datasets

packages/studio-base/src/panels/Plot/plotData.ts
cfoust: Indeed it can! Nice catch; I believe (as happened elsewhere in this PR) I had some intermediate logic that I removed and didn't revisit this

packages/studio-base/src/components/TimeBasedChart/downsample.ts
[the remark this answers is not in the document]
jtbandes: I don't feel strongly about changing this but want to explain my thinking:

The reason this raised an eyebrow for me is the unconstrained `<T>` type parameter — as the type signature is written the function body cannot do anything whatsoever with `dataset` _except_ passing it to `iterate()`, so that's why it feels redundant to me to have both parameters.

> it's written this way to accommodate arbitrary data types, not just objects that satisfy the Iterable interface

But it doesn't accommodate totally arbitrary data types, it only accommodates types for which you can write an `iterate()` function that returns a `Generator<Point>`... which is almost exactly what the `Iterable`/`Iterator` interfaces are for 🙂  Using one of them would not really restrict the container types this function could be used with.

Point being, as a reader I would find `downsample(iterateObjects(dataset), view)` more clear than `downsample(iterateObjects, dataset, view)`.

packages/studio-base/src/components/Chart/worker/proxy.ts
```diff
@@ -0,0 +1,84 @@
...
+import { ChartData } from "chart.js";
+import * as R from "ramda";
+
+import { findIndices } from "../datasets";
+import { TypedChartData, TypedData } from "../types";
+
+type TypedDataSet = TypedChartData["datasets"][0];
+type NormalDataSet = ChartData<"scatter">["datasets"][0];
+
+function proxyDataset(dataset: TypedDataSet): NormalDataSet {
```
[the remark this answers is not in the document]
jtbandes: ^ this should be a code comment

Also, I found that there was some work done in chart.js to support typedarrays [here](https://github.com/chartjs/Chart.js/pull/5905) — why doesn't that work for us?

packages/studio-base/src/components/Chart/datasets.ts
```diff
@@ -0,0 +1,137 @@
...
+  return R.pipe(
+    R.map((v: TypedData) => v.x.length),
+    R.sum,
+  )(data);
+}
+
+/**
+ * iterateObjects iterates over ObjectData, yielding a `Point` for each entry.
+ */
+export function* iterateObjects(dataset: ObjectData): Generator<Point> {
```
snosenzo, line 22: Also this name is too generic, I keep thinking that it's a util that could be used outside of plots. `iterateDatumObjects` or something similar would help keep it's context in the greater repo

packages/studio-base/src/components/TimeBasedChart/useProvider.tsx
```diff
@@ -0,0 +1,204 @@
...
+
+import { iterateObjects, iterateTyped } from "@foxglove/studio-base/components/Chart/datasets";
+import type { ObjectData, TypedData } from "@foxglove/studio-base/components/Chart/types";
+
+import { PlotDataProvider, ProviderState, Bounds, PlotViewport } from "./types";
+
+type Datasets<T> = ChartDataset<"scatter", T>[];
+type Data<T> = ChartData<"scatter", T>;
+
+export function getBounds(data: Datasets<ObjectData>): Bounds | undefined {
```
snosenzo, line 26: prefer `getDatumBounds` to parallel `getTypedBounds`

packages/studio-base/src/panels/Plot/datasets.tsx
```diff
@@ -205,10 +234,219 @@ export function getDatasetsFromMessagePlotPath({
...
-    data: plotData,
+    data: [datumToTyped(plotData)],
   };
   return {
     dataset,
     hasMismatchedData,
   };
 }
+
+export function resolveTypedIndices(data: TypedData[], indices: number[]): TypedData[] | undefined {
```
snosenzo, line 245: Why does this return an array instead of a single `TypedData`?

packages/studio-base/src/panels/Plot/index.tsx
```diff
@@ -222,29 +240,29 @@ function Plot(props: Props) {
...
-  const latestDatasets = useLatest(datasets);
-
   const getPanelContextMenuItems = useCallback(() => {
     const items: PanelContextMenuItem[] = [
       {
         type: "item",
         label: "Download plot data as CSV",
-        onclick: () => downloadCSV(latestDatasets.current, xAxisVal),
+        onclick: async () => {
+          const data = await getFullData();
```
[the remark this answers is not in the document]
snosenzo: I think this comment could be a comment in the actual code.

packages/studio-base/src/panels/Plot/index.tsx
```diff
@@ -183,18 +174,45 @@ function Plot(props: Props) {
...
+      datasets: yAxisPaths.map((path) => {
+        for (const [otherPath, dataset] of data.datasets.entries()) {
+          if (
+            otherPath.value === path.value &&
+            otherPath.timestampMethod === path.timestampMethod
+          ) {
+            return dataset;
+          }
+        }
+        return { label: path.label ?? path.value, data: [EmptyData] };
```
snosenzo, line 211: Why `[EmptyData]` and not `EmptyDataset`?

packages/studio-base/src/panels/Plot/plotData.ts
```diff
@@ -200,14 +229,25 @@ export function buildPlotData(
...
-          bounds.x.max = Math.max(bounds.x.max, datum.x);
-        }
-        if (isFinite(datum.y)) {
-          bounds.y.min = Math.min(bounds.y.min, datum.y);
-          bounds.y.max = Math.max(bounds.y.max, datum.y);
+
+      const {
+        dataset: { data: subData },
+      } = res;
+      for (const dataset of subData) {
```
snosenzo, line 236: Can't we use one of the `getBounds` or `getTypedBounds` here?

packages/studio-base/src/components/TimeBasedChart/downsample.ts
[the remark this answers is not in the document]
cfoust: I see where you're coming from--it would be great if this were possible--but `Iterable`/`Iterator` require setting or overriding the existing implementation for `Symbol.iterator` on the underlying value, which is a no-go. It would be really weird to pass around an array (`TypedData[]`) whose iterator no longer does what you expect. I might be missing something but I'm pretty confident that's what this would entail.

packages/studio-base/src/panels/Plot/datasets.tsx
cfoust: While it will only ever return an array with a single item, it's to make it obvious that it's still just `TypedData[]` and should be used as such. Forcing the caller to do `[result]` is undesirable

packages/studio-base/src/panels/Plot/index.tsx
cfoust: a full dataset is the outer type (`{ label...`), this is just the `data` field inside of a dataset. admittedly this is confusing, but just a quirk of the chartJS typing

packages/studio-base/src/components/Chart/worker/proxy.ts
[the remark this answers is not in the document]
cfoust: It turns out that you can indeed pass `TypedArray` data, something that did not appear in their documentation. That still does not save us much, though. Our tooltips, for example, expect objects; I left this in place so that ChartJS would provide data in the shape that main thread components are expecting. I didn't want to bother taking the time to look into how to patch the differences up.

cfoust:
Just to summarize the thinking on the `Datum`/`object` and `TypedData[]` duality for posterity: I think that given the imminent deadline it probably doesn't make sense to spend the time it would take to shift these around, rename them, or validating inputs in the `TimeBasedChart`.

If we were making a long-term commitment to having this API surface in our codebase I'd say agree. Coming up with a sensible and self-describing abstraction would be worth it. But I think that it's obvious to all involved that, minimally, we should come back soon and remove the `Datum` logic altogether, because the only reason that's sticking around in this PR is due to (a) technical debt and (b) I didn't want to touch the State Transition panel, which still uses it. "Soon" is a dangerous word in engineering, but we have at least made verbal plans to "fix plots" as soon as we can, preferably out from under the shadow of a deadline.

So in absence of anyone taking a harder line on this, I'm just going to do the following:
* [ ] Write tests for `sliceTyped`, `mergeTyped`, `resolveTypedIndices`
* [ ] Break out and clarify the "client" logic both in `useDatasets` and in the worker (respectively)

packages/studio-base/src/panels/Plot/useDatasets.ts
```diff
@@ -0,0 +1,269 @@
...
+
+// topic -> isSent
+type BlockStatus = Record<string, boolean>;
+let blockStatus: BlockStatus[] = [];
+
+type Client = {
+  topics: readonly string[];
+  setter: (topics: string[]) => void;
+};
+let clients: Record<string, Client> = {};
```
foxymiles, line 51: We've generally tried to avoid module level state like this. Can we provide these via something we inject? That would make testing easier too I think.

packages/studio-base/src/panels/Plot/useDatasets.ts
```diff
@@ -0,0 +1,269 @@
...
+    return () => {
+      const { [id]: _, ...rest } = clients;
+      clients = rest;
+      chooseClient();
+    };
+  }, [id, topics]);
+
+  const isLive = useMessagePipeline<boolean>(getIsLive);
+  useEffect(() => {
+    void (async () => {
```
foxymiles, line 93: Maybe we should at least log these to console.error instead of `void` ing them?

packages/studio-base/src/panels/Plot/useDatasets.ts
```diff
@@ -0,0 +1,269 @@
...
+
+  useCurrent<number>({
+    topics: subscribed,
+    restore: React.useCallback((state: number | undefined): number => {
+      if (state == undefined) {
+        void service?.clearCurrent();
+      }
+      return 0;
+    }, []),
+    addMessages: React.useCallback(
```
foxymiles, line 107: Agree that it would be nice to omit the `React.` prefixes for consistency with the rest of the app.

packages/studio-base/src/panels/Plot/useDatasets.worker.ts
```diff
@@ -0,0 +1,596 @@
...
+function getFullData(id: string): PlotData | undefined {
+  const client = clients[id];
+  if (client == undefined) {
+    return;
+  }
+
+  return getClientData(client);
+}
+
+export const service = {
```
foxymiles, line 582: I like that this file is broken up into smaller, more focused functions. Can we write tests for some of this?

packages/studio-base/src/panels/Plot/useDatasets.worker.ts
```diff
@@ -0,0 +1,596 @@
...
+
+    const accumulated = accumulate(initAccumulated(client.topics), params, current);
+    mutateClient(client.id, {
+      ...client,
+      current: accumulated,
+    });
+    client.setProvided?.(getProvidedData(accumulated.data));
+  }
+}
+setInterval(compressClients, 2000);
```
foxymiles, line 544: Having this at module level feels a bit off. Can this be managed in a more controlled way?

packages/studio-base/src/components/Chart/worker/proxy.ts
foxymiles, line 23: This feels like something we'd want to have some tests for.

packages/studio-base/src/panels/Plot/maps.ts
```diff
@@ -37,20 +37,6 @@ export function merge<K, V1, V2, V3>(
   return dest;
 }
 
-/**
- * Generates a new map from input with the same keys but with fn applied to each value.
- */
-export function mapValues<K, V, V2>(
```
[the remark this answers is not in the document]
foxymiles: That use of object keys was deliberate and we make sure to maintain reference equality in the places we used it.

foxymiles (commented):
The overall approach makes sense here. I think Sam & Jacob have done a good job addressing naming/organization concerns. It's a bit difficult to review a change this big in a single PR but I know this is a hard change to make incrementally.

Despite being a big fan of functional style in general I'm not immediately sold on the usage of Ramda here. Maybe over time it would grow on me.

foxymiles:
Also it would be nice to see some quantitative before/after examples of performance improvements from these changes.

snosenzo:
Found an issue: If local data is loaded without a plot panel originally in the layout, the plot panel doesn't update until it seeks into the file (or receives a message?). Tested with nuscenes mcap.

https://github.com/foxglove/studio/assets/10187776/e8dc5a94-fbab-43d9-93de-cace88b906fa

jtbandes:
Minor issue I noticed on this branch: while I am in the process of zooming out, I see lines extending beyond the active viewport:

image

They disappear very quickly but if you look carefully at the right side of the plot, it is noticeable. This video compares this preview deploy with studio.foxglove.dev where the issue is not present:

https://github.com/foxglove/studio/assets/14237/82614e9f-dd98-49c6-8ac8-78b1467efd01

This is the sample dataset with a single plot panel showing `/imu.linear_accel.x`.

packages/studio-base/src/components/Chart/datasets.ts
```diff
@@ -0,0 +1,138 @@
...
+ * }
+ * It is used to go from `TypedData`'s various incarnations to what a single
+ * point would look like as a `Datum`.
+ *
+ * These `any`s do not introduce anything unsafe; they are necessary for
+ * specifying the type (which is ultimately type-checked at point of use.)
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type ExtractPoint<T extends { [key: string]: Array<any> | Float32Array }> = {
+  [P in keyof T]-?: T[P] extends Float32Array ? number : NonNullable<T[P]>[0];
```
jtbandes, line 63: I think you can simplify this, because `Float32Array[0]` is `number.

```suggestion
  [P in keyof T]-?: NonNullable<T[P]>[0];
```

cfoust: You're correct! Nice catch!

foxymiles (approved):

PR #6585, opened 2023-08-03
Eliminate unproductive subscription requests from raw messages panel path editing

foxymiles:
**User-Facing Changes**
Improve performance of raw message panel message path editing.

**Description**
Check that the path in the raw messages panel message path resolves to a known topic before passing it on to the message pipeline to avoid spamming pointless subscription requests for non-existing topics.


defunctzombie (approved):
👏

PR #6586, opened 2023-08-03
Use fast fuzzy find algorithm on long item lists in autocomplete

foxymiles:
**User-Facing Changes**
Performance improvement on message path editing with long autocomplete lists.

**Description**
There seems to be a very large performance difference on the fzf matching in the autocomplete on with long lists of items.

This is the performance of the slower v2 algorithm on a list of about 8k items:
Screenshot 2023-08-03 at 12 46 55 PM

Switching to the v1 algorithm makes this much faster:
Screenshot 2023-08-03 at 12 46 33 PM

This patch switches to the faster but less precise algorithm with more than 1,000 items.

https://github.com/junegunn/fzf/blob/4c9cab3f8ae7b55f7124d7c3cf7ac6b4cc3db210/src/algo/algo.go#L5

To quote:

```
FuzzyMatchV1 finds the first "fuzzy" occurrence of the pattern within the given
text in O(n) time where n is the length of the text. Once the position of the
last character is located, it traverses backwards to see if there's a shorter
substring that matches the pattern.

    a_____b___abc__  To find "abc"
    *-----*-----*>   1. Forward scan
             <***    2. Backward scan

The algorithm is simple and fast, but as it only sees the first occurrence,
it is not guaranteed to find the occurrence with the highest score.

    a_____b__c__abc
    *-----*--*  ***

FuzzyMatchV2 implements a modified version of Smith-Waterman algorithm to find
the optimal solution (highest score) according to the scoring criteria. Unlike
the original algorithm, omission or mismatch of a character in the pattern is
not allowed.

Performance
-----------

The new V2 algorithm is slower than V1 as it examines all occurrences of the
pattern instead of stopping immediately after finding the first one. The time
complexity of the algorithm is O(nm) if a match is found and O(n) otherwise
where n is the length of the item and m is the length of the pattern. Thus, the
performance overhead may not be noticeable for a query with high selectivity.
However, if the performance is more important than the quality of the result,
you can still choose v1 algorithm with --algo=v1.
```

### DIFF
--- a/packages/studio-base/src/components/Autocomplete.tsx
+++ b/packages/studio-base/src/components/Autocomplete.tsx
@@ -32,6 +32,9 @@ import { ReactWindowListboxAdapter } from "@foxglove/studio-base/components/Reac
 
 const MAX_FZF_MATCHES = 200;
 
+// Above this number of items we fall back to the faster fuzzy find algorithm.
+const FAST_FIND_ITEM_CUTOFF = 1_000;
+
 type AutocompleteProps<T> = {
   autoSize?: boolean;
   disableAutoSelect?: boolean;
@@ -198,7 +201,8 @@ export default React.forwardRef(function Autocomplete<T = unknown>(
   const fzf = useMemo(() => {
     // @ts-expect-error Fzf selector TS type seems to be wrong?
     return new Fzf(items, {
-      fuzzy: "v2",
+      // v1 algorithm is significantly faster on long lists of items.
+      fuzzy: items.length > FAST_FIND_ITEM_CUTOFF ? "v1" : "v2",
       sort: sortWhenFiltering,
       limit: MAX_FZF_MATCHES,
       selector: getItemText,


packages/studio-base/src/components/Autocomplete.tsx
```diff
@@ -32,6 +32,9 @@ import { ReactWindowListboxAdapter } from "@foxglove/studio-base/components/Reac
 
 const MAX_FZF_MATCHES = 200;
 
+// Above this number of items we fall back to the faster fuzzy find algorithm.
+const FAST_FIND_ITEM_CUTOFF = 1_000;
```
defunctzombie, line 36: For randomly picked cutoff numbers like this I like to indicate that it is an arbitrarily picked number. If you can also link to any specific process or write a thought on how we should evaluate adjusting this number that will help for future maintenance. Stuff like this scares me a bit because it turns into a gamble of why not 500, 2000?

defunctzombie: Also worth calling out at a high level the difference in algorithm choice that this will make on result quality.

defunctzombie (approved):

defunctzombie:
Is it worth always using the v1 algo for our use case?

packages/studio-base/src/components/Autocomplete.tsx
foxymiles: Yeah it's an educated guess. The matching on my machine is very slow with 8k items but it will also depend on the string length of the provided potential matches.

Nuscenes has about 1,740 potential matches.

foxymiles:
> Is it worth always using the v1 algo for our use case?

Worth considering. I think it's still pretty good at finding items in a long list and it's much faster.

defunctzombie:
> Worth considering.

How would we make the decision? Somehow I recall maybe we used v1 and switched to v2? What are some examples of queries where v1 is worse?

foxymiles:
> How would we make the decision? Somehow I recall maybe we used v1 and switched to v2? What are some examples of queries where v1 is worse?

I haven't really tried to do a survey of a variety of topic/fields + queries to gauge the scope of it. I'd argue v2 is so slow as to be functionally broken on larger datasets though.

defunctzombie:
@foxymiles Do you need more input on this or is it ready to go?

foxymiles:
@defunctzombie You sounded like you wanted a bit more quantitative breakdown of the difference between the two algorithms. If not then this is ready to go.

foxymiles:
> I've read the above blurb and think I understand the difference in descriptions at a high level. It would have been better for your PR to provide an example where using the new algorithm returned different results so we could compare the tradeoff. Right now it says "this algorithm is faster" but yea - it would be faster to do a worse search so what's the impact to our users? It is hard to review this PR from a product perspective otherwise since this change _does_ change the search behavior. For our use case it might result in the same overall experience but I see nothing in the PR to indicate that or justifying a worse experience if there is on.

The justification is vastly improved performance. I think the user can still pretty easily narrow down what they want but it might take a few more keystrokes. I can provide some concrete examples of the difference if you want.

foxymiles:
Looks like we already made a similar but more drastic optimization for the TopicList:

https://github.com/foxglove/studio/blob/4b18ba9c8d20ece085c23aab137c4dadc0edf9cb/packages/studio-base/src/components/TopicList.tsx#L164

defunctzombie (approved):
What does the performance look like if you cut over to v2 after 2 (or 3) characters like we do in topic list? Is it still bad?

foxymiles:
> What does the performance look like if you cut over to v2 after 2 (or 3) characters like we do in topic list? Is it still bad?

Performance at that point is ok but there's a perceptible one-time hiccup as it transitions.

PR #6590, opened 2023-08-04
CompressedVideo support for Image Panel

defunctzombie:
**User-Facing Changes**
Add support for viewing H264 encoded data to image panel

**Description**

Modern robotics stacks can use many high-resolution cameras. On such systems, storing and sending each individual frame, even when using compressed image formats like jpeg, can be costly. Video codecs provide an alternative approach for recording discrete image frames with greatly reduced storage, bandwidth, and CPU. Robotics systems developers can tune their video codecs to tradeoff system resources for image fidelity and vice-versa.

This PR adds support for viewing H264 encoded data in the image panel. To support H264 video data, we create a new schema `foxglove.CompressedVideo` that stores compressed video data. The message schema fields are identical to `foxglove.CompressedImage` but are used for video. A separate schema is introduced to better differentiate messages for single-frame image topics vs messages with video data. Messages with video data are stateful and require different kinds of handling to display. For example a delta frame cannot be displayed on its own without the keyframe and entire sequence of delta frames proceeding it.

```typescript
type CompressedVideo = {
  timestamp: Time;
  frame_id: string;
  data: Uint8Array;
  format: string;
};
```

For this PR, only `format: "h264"` is supported. With this format, the `data` must contain NAL packets in annex-b format. Each message must contain enough packets to decode _exactly_ one image (either a keyframe or a delta frame).

This PR uses [WebCodecs](https://www.w3.org/TR/webcodecs/#video-decoder-config) to decode the video stream. We look for the first keyframe message and parse it to build a VideoDecoderConfig. Once the VideoPlayer is initialized, each `foxglove.CompressedVideo` message is fed into the VideoDecoder and a decoded frame is read and displayed.

When a user seeks, they may land on any `CompressedVideo` message. If the message contains `delta` frame data, we show a "waiting for keyframe" message and the user must play forward until a message with a `keyframe` is read.

Resolves: FG-2214

---
Test file:
[compressed_video.mcap.zip](https://github.com/foxglove/studio/files/12296529/compressed_video.mcap.zip)


packages/studio-base/src/panels/ThreeDeeRender/renderables/Images/ImageRenderable.ts
```diff
@@ -48,6 +57,7 @@ export const IMAGE_RENDERABLE_DEFAULT_SETTINGS: ImageRenderableSettings = {
 export type ImageUserData = BaseUserData & {
   topic: string;
   settings: ImageRenderableSettings;
+  firstMessageTime: bigint | undefined;
```
snosenzo, line 60: comment for what this is used for?

packages/studio-base/src/panels/ThreeDeeRender/renderables/Images/ImageRenderable.ts
```diff
@@ -79,8 +92,29 @@ export class ImageRenderable extends Renderable<ImageUserData> {
...
+    this.#waitingForKeyframeLabel.setColor(1, 1, 1);
+    this.#waitingForKeyframeLabel.setOpacity(1);
+    this.#waitingForKeyframeLabel.setBackgroundColor(0, 0, 0);
+    this.#waitingForKeyframeLabel.setLineHeight(20);
+    this.#waitingForKeyframeLabel.setBillboard(true);
+    this.#waitingForKeyframeLabel.setSizeAttenuation(false);
+    this.#waitingForKeyframeLabel.visible = false;
+
+    this.#waitingForKeyframeLabel.quaternion.set(0, 0, 0, 1);
+    this.#waitingForKeyframeLabel.position.set(0, 0, 0.1);
```
snosenzo, line 115: I think this will only work for cameras that face up

packages/studio-base/src/panels/ThreeDeeRender/renderables/Images/decodeImage.ts
```diff
@@ -19,18 +19,67 @@ import {
...
+  switch (frameMsg.format) {
+    case "h264": {
+      // Search for an IDR NAL unit to determine if this is a keyframe
+      return H264.IsKeyframe(frameMsg.data);
+    }
+  }
+  return false;
+}
+
+export function getVideoDecoderConfig(frameMsg: CompressedVideo): VideoDecoderConfig | undefined {
```
snosenzo, line 47: Would be nice to have comments for these functions. Also hard to tell how performance intensive this is

packages/studio-base/src/panels/ThreeDeeRender/renderables/Images/ImageRenderable.ts
```diff
@@ -165,18 +203,98 @@ export class ImageRenderable extends Renderable<ImageUserData> {
...
+      // Initialize the video player if needed
+      if (!videoPlayer.isInitialized()) {
+        const decoderConfig = getVideoDecoderConfig(frameMsg);
+        if (decoderConfig) {
+          await videoPlayer.init(decoderConfig);
+        }
+      }
+
+      // If still not initialized then we can't process the frame
+      if (!videoPlayer.isInitialized()) {
```
snosenzo, line 263: nit: this double checking of `isInitialized()` looks a little awkward to me, wonder if there's simpler way to do this logic and maybe consolidate the `emptyVideoFrame` behavior.

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageMode.ts
```diff
@@ -626,9 +637,13 @@ export class ImageMode
...
       planarProjectionFactor: 1,
     };
+    const messageTime = image
+      ? toNanoSec("header" in image ? image.header.stamp : image.timestamp)
+      : 0n;
     renderable = new ImageRenderable(topicName, this.renderer, {
       receiveTime,
-      messageTime: image ? toNanoSec("header" in image ? image.header.stamp : image.timestamp) : 0n,
+      messageTime,
+      firstMessageTime: messageTime,
```
snosenzo, line 646: fyi this won't be re-set when the image topic changes. You'll have to set that `#this.handleImageChange`

snosenzo (commented):
took a look at the 3D stuff so far, will review the rest a bit later

packages/studio-base/src/panels/ThreeDeeRender/renderables/Images/ImageRenderable.ts
defunctzombie: What does "cameras that face up" mean? Does every camera in `ImageMode` face up? In what situation would a camera not face up in `ImageMode`? Or is the statement that to support video in 3d mode we need to support other camera orientations?

packages/studio-base/src/panels/ThreeDeeRender/renderables/Images/ImageRenderable.ts
defunctzombie: I had added this because it _seemed_ possible for the video player to get initialized, then a seek happens which resets the initialization but maybe that's too pedantic and not actually possible for this to occur between the `await` on `videoPlayer.init()` and the code that follows so I can remove this extra check.

packages/studio-base/src/panels/ThreeDeeRender/renderables/Images/ImageRenderable.ts
defunctzombie: So I should leave the code as I have it?

snosenzo: yeah you can leave it as is 👍

packages/den/video/h264/H264.ts
```diff
@@ -0,0 +1,182 @@
...
+export enum H264NaluType {
+  NDR = 1,
+  IDR = 5,
+  SEI = 6,
+  SPS = 7,
+  PPS = 8,
+  AUD = 9,
+}
+
+export class H264 {
```
snosenzo, line 16: might be worth putting a link to the spec in a comment  for some of these classes that have them

packages/den/video/h264/Bitstream.ts
```diff
@@ -0,0 +1,185 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+/**
+ * Bitstream reader for H264 data, handling emulation prevention bytes and
+ * decoding exponential Golomb format integers.
```
snosenzo, line 7: Is there a reference for this Golomb format that was used for this? would be helpful for future folks in case there's an issue

packages/den/video/h264/SPS.ts
```diff
@@ -0,0 +1,367 @@
...
+  [18, 11],
+  [15, 11],
+  [64, 33],
+  [160, 99],
+  [4, 3],
+  [3, 2],
+  [2, 1],
+];
+
+export class SPS {
```
snosenzo, line 38: comment would help

amacneil:
LR feedback:
- Capitalize `Video` panel name
- `Waiting for keyframe` text should look more like our normal blank panel format:

   image
   image
- If there is no camera calibration (topic error), don't show anything in 3d panel:
   
   image
- In 3d panel, if we are waiting for keyframe, show a black "Waiting for keyframe" placeholder image _in place_ of the normal video (i.e. not always facing the camera)

PR #6593, opened 2023-08-04
Allow to set URDF fallback color

achim-k:
**User-Facing Changes**
Allow to set URDF fallback color which is used when link do not contain color information

**Description**
Adds a color setting to the URDF renderable, which allows users to specify a fallback color. The fallback color is only used when the link itself does not contain any color information (no `<material>` element, no collada file)

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
@@ -231,7 +231,7 @@ export class Urdfs extends SceneExtension<UrdfRenderable> {
 
   public override settingsNodes(): SettingsTreeEntry[] {
     const entries: SettingsTreeEntry[] = [];
-    const baseDisplayMode: SettingsTreeField = {
+    const baseDisplayModeField: SettingsTreeField = {
       label: "Display mode",
       input: "select",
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
```diff
@@ -1012,18 +1051,20 @@ async function parseUrdf(
...
+  renderer: IRenderer;
+  baseUrl?: string;
+  fallbackColor?: ColorRGBA;
+}): Renderable {
+  const { visual, robot, id, frameId, renderer, baseUrl, fallbackColor } = args;
   const name = `${frameId}-${id}-${visual.geometry.geometryType}`;
   const orientation = eulerToQuaternion(visual.origin.rpy);
   const pose = { position: visual.origin.xyz, orientation };
-  const color = getColor(visual, robot);
+  const color = getColor(visual, robot, fallbackColor ?? DEFAULT_COLOR);
```
defunctzombie, line 1067: Maybe nicer to make `getColor` return undefined and apply the fallback color on top of that since you need to provide the fallback color anyway?

```
getColor() ?? fallbackColor;
```

defunctzombie (approved):

PR #6594, opened 2023-08-04
Pause playback until a fixed amount is buffered

achim-k:
**User-Facing Changes**
Pause playback until a fixed amount is buffered

**Description**
Changes playback behavior to only resume playback when a fixed amount of time (1 second by default) is buffered. This results in a smoother, YouTube-like playback experience. So far when the read head reached the end of the buffered time range, we kept reading from the buffer which made visualizations appear jittery if buffering could not keep up with playback speed.

| Current behavior                                                                                       | This PR                                                                                                |
|--------------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------|
| <video src="https://github.com/foxglove/studio/assets/9250155/2610247f-3ecf-474b-aaee-f0b716ced2e3" /> | <video src="https://github.com/foxglove/studio/assets/9250155/6e8ae2f9-145b-46df-b841-2bf98358c3fc" /> |

### DIFF
--- a/packages/studio-base/src/players/IterablePlayer/BufferedIterableSource.ts
+++ b/packages/studio-base/src/players/IterablePlayer/BufferedIterableSource.ts
@@ -23,10 +23,13 @@ import {
 const log = Log.getLogger(__filename);
 
 const DEFAULT_READ_AHEAD_DURATION = { sec: 10, nsec: 0 };
+const DEFAULT_MIN_READ_AHEAD_DURATION = { sec: 1, nsec: 0 };
 
 type Options = {
   // How far ahead to buffer
   readAheadDuration?: Time;
+  // The minimum duration to buffer before playback resumes
+  minReadAheadDuration?: Time;
 };
 
 interface EventTypes {
@@ -69,10 +72,14 @@ class BufferedIterableSource extends EventEmitter<EventTypes> implements IIterab
   // How far ahead of the read head we should try to keep buffering
   #readAheadDuration: Time;
 
+  // The minimum duration to buffer before playback resumes
+  #minReadAheadDuration : Time;
+
   public constructor(source: IIterableSource, opt?: Options) {
     super();
 
     this.#readAheadDuration = opt?.readAheadDuration ?? DEFAULT_READ_AHEAD_DURATION;
+    this.#minReadAheadDuration = opt?.minReadAheadDuration ?? DEFAULT_MIN_READ_AHEAD_DURATION;
     this.#source = new CachingIterableSource(source);
 
     // pass-through the range change event
@@ -149,6 +156,12 @@ class BufferedIterableSource extends EventEmitter<EventTypes> implements IIterab
         }
 
         this.#cache.enqueue(result);
+
+        // Make sure that we have buffered enough ahead before telling the consumer to try reading again.
+        const rangeEnd = addTime(this.#readHead, this.#minReadAheadDuration);
+        if (!this.#source.isRangeBuffered(this.#readHead, rangeEnd)) {
+          continue;
+        }
 
         // Indicate to the consumer that it can try reading again
         this.#readSignal.notifyAll();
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/players/IterablePlayer/BufferedIterableSource.ts
```diff
@@ -150,6 +161,12 @@ class BufferedIterableSource extends EventEmitter<EventTypes> implements IIterab
 
         this.#cache.enqueue(result);
 
+        // Make sure that we have buffered enough ahead before telling the consumer to try reading again.
+        const rangeEnd = addTime(this.#readHead, this.#minReadAheadDuration);
+        if (!this.#source.isRangeBuffered(this.#readHead, rangeEnd)) {
```
defunctzombie, line 166: This looks like an expensive operation to do for every single message. The logic loops over all the cache blocks. Mentally I'm thinking we can track where we've buffered within the BufferedIterableSource without dropping down to the CachingIterableSource.

defunctzombie (commented):
I like the results in the video. I propose we find an implementation that works within the BufferedIterableSource and avoid looping over the blocks for every message.

packages/studio-base/src/players/IterablePlayer/BufferedIterableSource.ts
achim-k: I have now changed it to communicate the last received message time stamp via the `IteratorResult`. Is this acceptable?

defunctzombie:
@achim-k Your changes were close. We don't need the `CachingIterableSource` involved in this logic at all. I've pushed a commit to this PR that isolates the logic to `BufferedIterableSource`. Your new test still passes unchanged. Take a look and if it behaves as you intended it LGTM.

defunctzombie (approved):

achim-k:
> @achim-k Your changes were close. We don't need the `CachingIterableSource` involved in this logic at all. I've pushed a commit to this PR that isolates the logic to `BufferedIterableSource`. Your new test still passes unchanged. Take a look and if it behaves as you intended it LGTM.

Much more elegant like that! Works as expected, will only change the name of the test as I saw that one is incorrect

packages/studio-base/src/players/IterablePlayer/BufferedIterableSource.test.ts
```diff
@@ -459,6 +460,69 @@ describe("BufferedIterableSource", () => {
     expect(messageIteratorCount).toEqual(1);
   });
 
+  it("should wait to buffer more messages until reading moves forward (with min. read ahead)", async () => {
```
achim-k, line 463:
```suggestion
  it("should buffer minimum duration ahead before messages can be read", async () => {
```

packages/studio-base/src/players/IterablePlayer/BufferedIterableSource.ts
```diff
@@ -150,14 +161,22 @@ class BufferedIterableSource extends EventEmitter<EventTypes> implements IIterab
 
         this.#cache.enqueue(result);
 
+        // We tend to expect message revents (not problems) so optimistically grab the receive time
```
jtbandes, line 164: typo: revents

packages/studio-base/src/players/IterablePlayer/BufferedIterableSource.ts
```diff
@@ -150,14 +161,22 @@ class BufferedIterableSource extends EventEmitter<EventTypes> implements IIterab
...
+        // and minReadAheadUntil
+        const receiveTime =
+          result.type === "message-event" ? result.msgEvent.receiveTime : undefined;
+
+        // Make sure that we have buffered enough ahead before telling the consumer to try reading again.
+        const minReadAheadUntil = addTime(this.#readHead, this.#minReadAheadDuration);
+        if (receiveTime && compare(receiveTime, minReadAheadUntil) < 0) {
```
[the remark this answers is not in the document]
achim-k: good catch, I will test this. We might lack a `clampTime` here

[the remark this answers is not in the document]
achim-k: This is not an issue as we exit the for loop when the underlying source iterator does not yield any messages anymore.

PR #6597, opened 2023-08-06
Convert Log panel to an extension panel

defunctzombie:
**User-Facing Changes**
Message Converter support for Log panel.

Visually:

The filter bar is moved under the panel toolbar. The toolbar shows the topic name.
image
**Description**
Resolves FG-4013


banisadr:
Copy button at right of search box is cut off:
image

2metres:
> Copy button at right of search box is cut off: image

Apparently that was a feature? https://github.com/foxglove/studio/blob/cd6c2cf8f00122ba1981fa95b940a1cdb710859b/packages/studio-base/src/panels/Log/FilterBar.tsx#L38-L40

PR #6598, opened 2023-08-07
Stop reading messages into cache when cache is full

achim-k:
**User-Facing Changes**
Stop reading messages into cache when cache is full

**Description**
Currently, we only stop reading messages into the cache after having read a sufficient time ahead (10 seconds by default). The cache size limit is not taken into account, which is often not a problem when dealing with lightweight messages. However, when reading large messages (images / pointclouds), the cache may be full before having read the entire read-ahead duration. This PR changes the behavior to also take into account the cache size when determining if more messages should be read into the cache. A positive side effect of this implementation is a smarter strategy when choosing the most evictable cache block

**Current behavior**
Only stops when read-ahead duration has been reached
current_buffering_behavior

Always evicts oldest cache block which can cause unnecessary re-reads
Peek 2023-08-11 19-31

**New behavior**
Stops when either read-ahead duration has been read or when cache is full 
new_buffering_behavior

uses a smart strategy for choosing the best evictable cache block
Peek 2023-08-11 19-32

Fixes FG-4631

### DIFF
--- a/packages/studio-base/src/players/IterablePlayer/CachingIterableSource.ts
+++ b/packages/studio-base/src/players/IterablePlayer/CachingIterableSource.ts
@@ -154,6 +154,9 @@ class CachingIterableSource extends EventEmitter<EventTypes> implements IIterabl
     };
 
     this.#currentReadHead = readHead;
+
+    // Compute evictable block candiates such that canReadMore() returns a correct result as the
+    // callee may stop iterating when canReadMore() returns false.
     this.#evictableBlockCandidates = this.#findEvictableBlockCandidates(this.#currentReadHead);
 
     for (;;) {
@@ -528,7 +531,7 @@ class CachingIterableSource extends EventEmitter<EventTypes> implements IIterabl
    * Update the current read head, such that the source can determine which blocks are evictable.
    * @param readHead current read head
    */
-  public updateCurrentReadHead(readHead: Time): void {
+  public setCurrentReadHead(readHead: Time): void {
     this.#currentReadHead = readHead;
   }
 
@@ -575,6 +578,8 @@ class CachingIterableSource extends EventEmitter<EventTypes> implements IIterabl
 
     if (readHeadIdx === -1) {
       // No block contains current read head, return the oldest cache block.
+      // This can only happen when seeked to a new time and no message has been read yet, as
+      // reading a new message that does not fit in any block will always create a new cache block.
       const oldestBlockIdx = minIndexBy(this.#cache, (a, b) => a.lastAccess - b.lastAccess);
       return [this.#cache[oldestBlockIdx]!.id];
     }
--- a/packages/studio-base/src/players/IterablePlayer/CachingIterableSource.ts
+++ b/packages/studio-base/src/players/IterablePlayer/CachingIterableSource.ts
@@ -584,7 +584,7 @@ class CachingIterableSource extends EventEmitter<EventTypes> implements IIterabl
       return [this.#cache[oldestBlockIdx]!.id];
     }
 
-    // Blocks that are before the read head can be recycled.
+    // Blocks that are before the read head can be evicted.
     const blockIdsBeforeReadHead = mappedCache.splice(0, readHeadIdx).map((item) => item.id);
 
     // Iterate through remaining blocks until we find a gap in the block chain
--- a/packages/studio-base/src/players/IterablePlayer/CachingIterableSource.ts
+++ b/packages/studio-base/src/players/IterablePlayer/CachingIterableSource.ts
@@ -581,7 +581,12 @@ class CachingIterableSource extends EventEmitter<EventTypes> implements IIterabl
       // This can only happen when seeked to a new time and no message has been read yet, as
       // reading a new message that does not fit in any block will always create a new cache block.
       const oldestBlockIdx = minIndexBy(this.#cache, (a, b) => a.lastAccess - b.lastAccess);
-      return [this.#cache[oldestBlockIdx]!.id];
+      const oldestBlock = this.#cache[oldestBlockIdx];
+      if (!oldestBlock) {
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/players/IterablePlayer/CachingIterableSource.ts
```diff
@@ -505,6 +505,42 @@ class CachingIterableSource extends EventEmitter<EventTypes> implements IIterabl
...
+   */
+  public cacheLimitReached(readHead: Time): boolean {
+    if (this.#totalSizeBytes + this.#maxBlockSizeBytes < this.#maxTotalSizeBytes) {
+      // Still room for reading new messages from the source.
+      return false;
+    }
+
+    // Find the oldest cache block as this is the cache block that will be evicted first
+    // in #maybePurgeCache.
+    const oldestBlockWithIdx = this.#getOldestCacheItem();
```
defunctzombie, line 522: If the cache is full, this is going to do a search whenever we receive another message. Once you fill the cache and are playing, every message is going to incur a search.

defunctzombie (commented):
I like the visual result and its the visual we want but I am skeptical of the search on every received message. Ideas on what else we can do? Nothing obvious stood out to me.

achim-k:
> I like the visual result and its the visual we want but I am skeptical of the search on every received message. Ideas on what else we can do? Nothing obvious stood out to me.

I have changed it now a little (updated PR description as well). You think it's OK like that? 
(will look into failing tests next week)

packages/studio-base/src/players/IterablePlayer/CachingIterableSource.ts
```diff
@@ -505,7 +524,88 @@ class CachingIterableSource extends EventEmitter<EventTypes> implements IIterabl
...
+    }));
+    // Sort blocks by time (earlier blocks first).
+    mappedCache.sort((a, b) => compare(a.end, b.end));
+    // Find the index of the block that contains the current read head.
+    const readHeadIdx = mappedCache.findIndex(
+      (block) => compare(block.start, readHead) <= 0 && compare(block.end, readHead) >= 0,
+    );
+
+    if (readHeadIdx === -1) {
+      // No block contains current read head, return the oldest cache block.
```
snosenzo, line 577: If the readAhead is before all of the cache blocks (for example: after a seek), should we be removing the first block? Or are we assuming that the cache is cleared on seek? If so, is that stated somewhere?

packages/studio-base/src/players/IterablePlayer/CachingIterableSource.ts
```diff
@@ -517,16 +617,22 @@ class CachingIterableSource extends EventEmitter<EventTypes> implements IIterabl
       return false;
     }
 
-    // Find the oldest cache item
-    const oldestIdx = minIndexBy(this.#cache, (a, b) => a.lastAccess - b.lastAccess);
+    // Find evictable block candidates
+    this.#evictableBlockCandidates = this.#findEvictableBlockCandidates(this.#currentReadHead);
```
snosenzo, line 621: shouldn't this be cleared on seek? Also is it possible that this could contain stale block id's that are under the cursor? We should try to have some tests around this

packages/studio-base/src/players/IterablePlayer/CachingIterableSource.ts
achim-k: > If the readAhead is before all of the cache blocks (for example: after a seek), should we be removing the first block? 

I thought so initially as well, but I just tested it and noticed that it behaves correctly. The reason being that when seeking to a non-cached time, we always create a new block. The above if block is only executed when [calculating the disposable blocks right after the seek ](https://github.com/foxglove/studio/blob/c1b1b7ed396ccdd15fcbfcb7f1c212ee73b546e4/packages/studio-base/src/players/IterablePlayer/CachingIterableSource.ts#L156-L157) (before the first message is read). I will add comments for that.

> Or are we assuming that the cache is cleared on seek? If so, is that stated somewhere?

No, the cache is not cleared.

packages/studio-base/src/players/IterablePlayer/CachingIterableSource.ts
achim-k: When seeking, the message iterator is reinstantiaed and the evictable candidates are calculated at the start of it: https://github.com/foxglove/studio/blob/c1b1b7ed396ccdd15fcbfcb7f1c212ee73b546e4/packages/studio-base/src/players/IterablePlayer/CachingIterableSource.ts#L156-L157

> Also is it possible that this could contain stale block id's that are under the cursor? We should try to have some tests around this

Will add some tests

snosenzo (approved):

packages/studio-base/src/players/IterablePlayer/CachingIterableSource.ts
```diff
@@ -505,7 +527,90 @@ class CachingIterableSource extends EventEmitter<EventTypes> implements IIterabl
...
+    const readHeadIdx = mappedCache.findIndex(
+      (block) => compare(block.start, readHead) <= 0 && compare(block.end, readHead) >= 0,
+    );
+
+    if (readHeadIdx === -1) {
+      // No block contains current read head, return the oldest cache block.
+      // This can only happen when seeked to a new time and no message has been read yet, as
+      // reading a new message that does not fit in any block will always create a new cache block.
+      const oldestBlockIdx = minIndexBy(this.#cache, (a, b) => a.lastAccess - b.lastAccess);
+      return [this.#cache[oldestBlockIdx]!.id];
```
foxymiles, line 584: I see you check the cache length earlier in the function but maybe best to avoid the `!` for people that are maintaining this later and don't realize that that check makes this safe?

foxymiles (approved):

PR #6608, opened 2023-08-08
Support subscribing to slices of topics

foxymiles:
**User-Facing Changes**
None (performance improvments)

**Description**
Proof of concept of subscribing to slices of topics containing specific fields instead of all fields of topics.


packages/studio-base/src/players/types.ts
```diff
@@ -336,3 +341,29 @@ export interface PlayerMetricsCollectorInterface {
   recordUncachedRangeRequest(): void;
   recordTimeToFirstMsgs(): void;
 }
+
+/**
+ * Builds a SubscribePayload from a ros path, requesting a specific field of the message if the
+ * message path resolves to a field name.
+ */
+export function subscribePayloadFromRosPath(
```
snosenzo, line 349: How is this specific to `ros`?

packages/studio-base/src/players/types.ts
```diff
@@ -276,9 +281,9 @@ export type SubscriptionPreloadType =
 
 // Represents a subscription to a single topic, for use in `setSubscriptions`.
 export type SubscribePayload = {
-  // The topic name to subscribe to
-  topic: string;
+  fields?: string[];
```
snosenzo, line 284: this should have comment

snosenzo (commented):
we're only supporting top-level fields right now it looks like. Is this because offering field depth and merging them would be too hard? Or is there a larger reason not to support a greater field depth?

packages/studio-base/src/components/MessagePipeline/subscriptions.ts
```diff
@@ -0,0 +1,65 @@
...
+
+import { Immutable } from "@foxglove/studio";
+import { SubscribePayload } from "@foxglove/studio-base/players/types";
+
+/**
+ * Create a deep equal memoized identify function. Used for stabilizing the subscription payloads we
+ * send on to the player.
+ */
+export function makeSubscriptionMemoizer(): (val: SubscribePayload) => SubscribePayload {
+  return moize((val: SubscribePayload) => val, { isDeepEqual: true, maxSize: Infinity });
```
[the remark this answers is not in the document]
foxymiles: The entire memoizer is cleared in the `reset` function of the store.

foxymiles:
> we're only supporting top-level fields right now it looks like. Is this because offering field depth and merging them would be too hard? Or is there a larger reason not to support a greater field depth?

This seems like it solves the problem well enough without getting too complex.

packages/studio-base/src/players/types.ts
foxymiles: It calls the `parseRosPath` function. Maybe that name is an anachronism now?

packages/studio-base/src/components/MessagePipeline/subscriptions.ts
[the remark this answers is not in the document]
snosenzo: would be good to have a comment here explaining that

snosenzo:
Also we have a _before_ performance snapshot in the description, can we get an _after_ snapshot? also would be nice to have a video of the behavior change around when topics change it has to reload

defunctzombie:
> This seems like it solves the problem well enough without getting too complex.

What happens if I select a nested field? Do I still benefit from the top field that my nested field is under being sent rather than the entire message? Or does selecting a nested field entire disable this optimization?

packages/studio-base/src/players/types.ts
defunctzombie: It is an anachronism - we should not introduce more instances of it.

yarn.lock
```diff
@@ -2743,6 +2743,7 @@ __metadata:
     mathjs: 11.9.1
     memoize-weak: 1.0.2
     meshoptimizer: 0.19.0
+    moize: 6.1.6
```
defunctzombie, line 2746: Why are we adding _yet another_ memoization library? What is deficient in our current library? Are you going to remove the current one in favor of this one?

foxymiles: This is a more featureful and better maintained library than the ones we're using now. It also gives you much more control including inspecting the current memo cache. None of the memo tools we have in studio now will work because I need per-instance deep equals memoziation.

foxymiles:
> > This seems like it solves the problem well enough without getting too complex.
> 
> What happens if I select a nested field? Do I still benefit from the top field that my nested field is under being sent rather than the entire message? Or does selecting a nested field entire disable this optimization?

You'd still benefit because the parent slice would be included.

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
```diff
@@ -592,7 +595,7 @@ export class IterablePlayer implements Player {
 
     log.debug("Initializing forward iterator from", this.#start);
     this.#playbackIterator = this.#bufferedSource.messageIterator({
-      topics: Array.from(this.#allTopics),
+      topics: this.#allTopics,
```
snosenzo, line 598: Should this be wrapped in a `new Map(this.#allTopics)` so that it can't be mutated?

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
```diff
@@ -681,12 +684,10 @@ export class IterablePlayer implements Player {
       this.#queueEmitState();
     }, 100);
 
-    const topics = Array.from(this.#allTopics);
-
     try {
       this.#abort = new AbortController();
       const messages = await this.#bufferedSource.getBackfillMessages({
-        topics,
+        topics: this.#allTopics,
```
snosenzo, line 690: Same here for wrapping this in `new Map`

packages/studio-base/src/players/types.ts
```diff
@@ -336,3 +353,29 @@ export interface PlayerMetricsCollectorInterface {
   recordUncachedRangeRequest(): void;
   recordTimeToFirstMsgs(): void;
 }
+
+/**
+ * Builds a SubscribePayload from a message path, requesting a specific field of the message if the
+ * message path resolves to a field name.
+ */
+export function subscribePayloadFromMessagePath(
```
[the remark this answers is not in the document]
foxymiles: Dunno it seemed like putting a single function in its own file was a bit to granular. And utils is out of fashion.

packages/studio-base/src/components/MessagePipeline/subscriptions.test.ts
```diff
@@ -0,0 +1,33 @@
...
+
+    const result = simplifySubscriptionsById(subs);
+
+    expect(result).toEqual(
+      expect.arrayContaining([
+        { topic: "a", preloadType: "full" },
+        { topic: "a", preloadType: "partial" },
+
+        { topic: "b", preloadType: "full", fields: ["one", "two"] },
+        { topic: "b", preloadType: "partial", fields: ["one", "two", "three"] },
```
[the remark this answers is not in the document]
foxymiles: Currently I'm interpreting a subscription with fields === [] as requesting nothing.

packages/studio-base/src/test/mocks/mockTopicSelection.ts
```diff
@@ -0,0 +1,9 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { TopicSelection } from "@foxglove/studio-base/players/types";
+
+export function mockTopicSelection(...topics: readonly string[]): TopicSelection {
```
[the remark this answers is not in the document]
foxymiles: It's just a test helper.

packages/studio-base/src/panels/Plot/usePlotPanelData.ts
```diff
@@ -212,15 +218,29 @@ export function usePlotPanelData(params: Params): Immutable<{
...
         const topic = parseRosPath(path.value)?.topicName;
         if (topic == undefined) {
           continue;
         }
 
+        // True if messages contain the same fields as our previous state. If so we can reuse the
+        // previous cursor.
+        const isSameMessageContent =
+          allFramesByTopic[topic]?.[0] === newState.allFrames[topic]?.[0];
+
```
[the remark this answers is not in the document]
foxymiles: Because when the subscriptions change these messages will be reloaded. We can't check fields here because this will contain everything all subscribers have requested, not just our fields.

packages/studio-base/src/players/subscribePayloadFromMessagePath.test.ts
```diff
@@ -0,0 +1,17 @@
...
+  it("handles whole topic paths", () => {
+    const result = subscribePayloadFromMessagePath("topic", "partial");
+    expect(result).toEqual({ topic: "topic", preloadType: "partial" });
+  });
+
+  it("handles specific field paths", () => {
+    const result = subscribePayloadFromMessagePath("topic.field");
+    expect(result).toEqual({ topic: "topic", fields: ["field"] });
+  });
+});
```
jtbandes, line 17: Add a test for a message path with more than 1 level of nesting (e.g. `topic.field.subfield`)?

And message paths with other features (filters, array slices) – like `topic{x==1}.field[:].subfield`? How is it supposed to behave in this case?

packages/studio-base/src/panels/Plot/useAllFramesByTopic.ts
```diff
@@ -114,6 +111,11 @@ export function useAllFramesByTopic(
           break;
         }
 
+        // Only include fully loaded blocks.
```
jtbandes, line 114: could use more information about why this is necessary

packages/studio-base/src/players/IterablePlayer/Mcap/McapIndexedIterableSource.ts
```diff
@@ -146,14 +149,20 @@ export class McapIndexedIterableSource implements IIterableSource {
         continue;
       }
       try {
+        const msg = channelInfo.parsedChannel.deserialize(message.data) as Record<string, unknown>;
+        const spec = args.topics.get(channelInfo.channel.topic);
+        const payload = spec?.fields != undefined ? pick(msg, spec.fields) : msg;
```
jtbandes, line 154: I think you might want to change this to `pick(msg, spec.fields.map(field => [field]))`, because `pick` will treat these as "paths" so if there happens to be a field with a `.` in it, things will break. (Although you probably wouldn't want to be executing that `map` each time here.)

jtbandes: We could also just replace lodash pick with our own simpler `for` loop.

jtbandes (approved):
I'm ok with trying out this approach, it seems not too invasive. I added some comments around `pick` correctness and tests that it would be good to address. I think @snosenzo should also take another look since he previously had a bunch of comments here.

It sounds like there are some known limitations where the `fields` optimization will be defeated (like with user scripts) but we can start here and evolve the solution over time.

packages/studio-base/src/players/IterablePlayer/Mcap/McapIndexedIterableSource.ts
foxymiles: Yes good catch! I'll replace this with a simpler pick that just takes top level fields.

packages/studio-base/src/components/MessagePipeline/subscriptions.test.ts
```diff
@@ -0,0 +1,58 @@
...
+  it("combines full and partial subscriptions", () => {
+    const subs: SubscribePayload[] = [
+      { topic: "a", preloadType: "full" },
+      { topic: "a", preloadType: "partial" },
+    ];
+
+    const result = mergeSubscriptions(subs);
+
+    expect(result).toEqual(
+      expect.arrayContaining([
```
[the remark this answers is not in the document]
foxymiles: Yeah you're right I think this is leftover from a previous iteration of these tests. They can just be `toEqual`.

packages/studio-base/src/players/subscribePayloadFromMessagePath.test.ts
foxymiles: Updated the logic to look for the first element of `type = "name"`.

snosenzo (approved):

Issue #6559, opened 2023-07-31
Setting Parameter of Type Double to Integer w/Trailing 0 Throws Type Error 


memtech3:
When attempting to set a parameter with type double to values like 1.0 or 2.0 from the parameters panel, foxglove removes the trailing zero causing an error `Failed to set one or more parameters for node '{NODE_NAME}': Wrong parameter type, parameter {PARAMETER_NAME} is of type {double}, setting it to {integer} is not allowed.`

- Version: 1.63
- Platform: Web (Chrome), Windows (Electron)
- Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native): foxglove_bridge

**Steps To Reproduce**
Try setting a double typed parameter to 0.0, 1.0, or 2.0 in foxglove studio

**Expected Behavior**
When attempting to pass a value with a trailing 0 into a double typed parameter the value should retain its trailing 0. When attempting to pass an integer into a double typed parameter foxglove should add a trailing 0.

PR #6613, opened 2023-08-08
foxglove websocket: Remember parameter types

achim-k:
**User-Facing Changes**
Not worth being mentioned in release notes

**Description**
When setting a float parameter to e.g. `1.0`, Javascript automatically makes it to `1` and the server would deduce this as a integer parameter. Remembering the parameter type that is communicated by the server, and sending it along with a `setParameter` request fixes this.

See also https://github.com/foxglove/ws-protocol/pull/519 and https://github.com/foxglove/ros-foxglove-bridge/pull/256

Fixes https://github.com/foxglove/studio/issues/6559

### DIFF
--- a/packages/studio-base/package.json
+++ b/packages/studio-base/package.json
@@ -45,7 +45,7 @@
     "@foxglove/velodyne-cloud": "1.0.1",
     "@foxglove/wasm-bz2": "0.1.1",
     "@foxglove/wasm-lz4": "1.0.2",
-    "@foxglove/ws-protocol": "0.7.0",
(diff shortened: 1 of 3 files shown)


snosenzo (commented):
I imagine this is related to the `float64` and `float64_array` type additions, but from the description and the code it's hard to see how just passing the type with the parameter allows the receiver to properly cast `1` to `1.0`. Where does this conversion happen? Is something already looking for `float64` to cast it?

snosenzo (approved):
Thanks for clarifying. I think I see now

jtbandes (approved):

PR #6618, opened 2023-08-09
Fix imu chart in sample data

foxymiles:
**User-Facing Changes**
Fix imu diagnostics chart in the sample data layout.

**Description**
The imu series in the sample layout is set to order by header stamp, which results in an empty series:
Screenshot 2023-08-09 at 8 28 40 AM

Changing it to receiveTime fixes it:
Screenshot 2023-08-09 at 8 29 01 AM

### DIFF
--- a/packages/studio-base/src/dataSources/SampleNuscenesLayout.json
+++ b/packages/studio-base/src/dataSources/SampleNuscenesLayout.json
@@ -324,7 +324,7 @@
         {
           "value": "/imu.linear_accel.x",
           "enabled": true,
-          "timestampMethod": "headerStamp"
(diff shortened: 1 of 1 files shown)


snosenzo (approved):
when did this change? Did it work with header stamp at some point?

foxymiles:
@snosenzo Good question. It looks like imu doesn't have a header? Maybe it did in a previous version of our sample data?

Screenshot 2023-08-09 at 8 55 52 AM

jtbandes:
Maybe in https://github.com/foxglove/nuscenes2mcap/pull/19?

PR #6626, opened 2023-08-10
Allow to skip messages in 3D/Image panel if possible

achim-k:
**User-Facing Changes**
Allow to skip messages in 3D/Image panel if possible

**Description**
Allows the 3D/Image scene extensions to decide if all messages have to be processed or if some messages can be skipped. Skipping messages will have positive impact on performance.

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
@@ -573,7 +573,7 @@ export class Renderer extends EventEmitter<RendererEvents> implements IRenderer 
 
     for (const message of currentFrameMessages) {
       const remainingMsgsOnSameTopic = msgCountByTopic.get(message.topic)! - 1;
-      this.addMessageEvent(message, remainingMsgsOnSameTopic);
+      this.addMessageEvent(message, { isLastMsgOnTopic: remainingMsgsOnSameTopic === 0 });
       msgCountByTopic.set(message.topic, remainingMsgsOnSameTopic);
     }
   }
@@ -907,7 +907,7 @@ export class Renderer extends EventEmitter<RendererEvents> implements IRenderer 
 
   public addMessageEvent(
     messageEvent: Readonly<MessageEvent>,
-    remainingMessagesOnSameTopic?: number,
+    options?: { isLastMsgOnTopic: boolean },
   ): void {
     const { message } = messageEvent;
 
@@ -942,16 +942,8 @@ export class Renderer extends EventEmitter<RendererEvents> implements IRenderer 
       this.addCoordinateFrame(maybeHasFrameId.frame_id);
     }
 
-    handleMessage(
-      messageEvent,
-      this.topicHandlers.get(messageEvent.topic),
-      remainingMessagesOnSameTopic,
-    );
-    handleMessage(
-      messageEvent,
-      this.schemaHandlers.get(messageEvent.schemaName),
-      remainingMessagesOnSameTopic,
-    );
+    handleMessage(messageEvent, this.topicHandlers.get(messageEvent.topic), options);
+    handleMessage(messageEvent, this.schemaHandlers.get(messageEvent.schemaName), options);
   }
 
   /** Match the behavior of `tf::Transformer` by stripping leading slashes from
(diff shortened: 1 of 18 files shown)


packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -887,7 +901,10 @@ export class Renderer extends EventEmitter<RendererEvents> implements IRenderer
     }
   }
 
-  public addMessageEvent(messageEvent: Readonly<MessageEvent>): void {
+  public addMessageEvent(
+    messageEvent: Readonly<MessageEvent>,
+    remainingMessagesOnSameTopic?: number,
```
snosenzo, line 906: This addition feels a bit awkward to me, but not sure what better way there would be. Maybe handling messages queues them, then another method flushes them to their handlers? Would be good to have a comment though because its usage is unclear

snosenzo (commented):
Overall I think this approach looks good. Wish the `remainingMessagesOnSameTopic` was a little cleaner, but not a big deal. Also it should be straightforward for H264 to not have active. 
Would be good to add to other scene extensions as well

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -564,6 +564,20 @@ export class Renderer extends EventEmitter<RendererEvents> implements IRenderer
     return true;
   }
 
+  public handleCurrentFrameMessages(currentFrameMessages: readonly MessageEvent[]): void {
```
snosenzo, line 567: Comment for this function would be good

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -887,7 +905,10 @@ export class Renderer extends EventEmitter<RendererEvents> implements IRenderer
     }
   }
 
-  public addMessageEvent(messageEvent: Readonly<MessageEvent>): void {
+  public addMessageEvent(
```
snosenzo, line 908: Should have comment describing parameters

achim-k:
> I have some small concerns about this being relatively hidden logic, and when debugging some folks might be wondering why some messages aren't being processed.

Do you think we should add a setting that lets users opt out of the skipping behavior if that causes problems?

packages/studio-base/src/panels/ThreeDeeRender/renderables/PointClouds.ts
```diff
@@ -707,12 +707,22 @@ export class PointClouds extends SceneExtension<PointCloudHistoryRenderable> {
       {
         type: "schema",
         schemaNames: ROS_POINTCLOUD_DATATYPES,
-        subscription: { handler: this.#handleRosPointCloud },
+        subscription: {
+          handler: this.#handleRosPointCloud,
+          canSkipMessages: () =>
+            this.userData.settings?.decayTime != undefined &&
```
snosenzo, line 713: I don't think the scene extension uses `this.userData` only the `renderables`. For these (Point clouds and laser scans) I think you'll need to do something more complex. You'll need to use the topic to get the renderable and check the `.settings.decayTime` on the renderable

snosenzo (changes requested):

KerstinKeller:
I appreciate that you are tackeling this issue, as it is really blocking us 👍 And happy to test out the changes.

I understand too little about the Foxglove architecture, to really make informed comments here.
However, to me, the current realization by providing a bool if you may / may not skip a data element if a newer one is already in the queue feels a bit unintuitive.

When we designed plugins (for algorithms, or monitoring solutions) in the past that handle visualization, not the data callbacks triggered preparing the data for rendering, but rather rendering triggers did.
Depending on what you wanted to render (or compute for that matter), you would create a queue of the data you had received, or you would keep just the last message.
Then only during cyclic compute/rendering calls you would identifiy which was the correct data to be used for computations / drawn into the scene.

achim-k:
The following screencast shows a comparison between current Studio (left) and Studio as of https://github.com/foxglove/studio/pull/6626 (right):

[Screencast from 21.08.2023 13:36:39.webm](https://github.com/foxglove/studio/assets/9250155/1a44cc13-6faf-4d82-a8ca-8d7b790572b1)

Oberservations:
- Raw message panel shows latest message and immediately stops when the image publisher is stopped (left and right)
- The image panel on the right side is up to date whereas the left side is lagging behind
- When the image publisher is stopped, the left side keeps on rendering buffered messages

achim-k:
Closing this PR, superseded by https://github.com/foxglove/studio/pull/7064

PR #6634, opened 2023-08-14
Inject workspace store creator via AppContext

foxymiles:
**User-Facing Changes**
None

**Description**
Inject a workspace store creation function via AppContext to allow for greater flexibility in workspace context setup.

### DIFF
--- a/packages/studio-base/src/providers/WorkspaceContextProvider.tsx
+++ b/packages/studio-base/src/providers/WorkspaceContextProvider.tsx
@@ -16,7 +16,6 @@ import { migrateV0WorkspaceState } from "@foxglove/studio-base/context/Workspace
 /**
  * Creates the default initial state for the workspace store.
  */
-// ts-unused-exports:disable-next-line
 export function makeWorkspaceContextInitialState(): WorkspaceContextStore {
   return {
     dialogs: {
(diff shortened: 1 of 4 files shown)


packages/studio-base/src/providers/WorkspaceContextProvider.tsx
```diff
@@ -13,44 +13,53 @@ import {
 } from "@foxglove/studio-base/context/Workspace/WorkspaceContext";
 import { migrateV0WorkspaceState } from "@foxglove/studio-base/context/Workspace/migrations";
 
+/**
+ * Creates the default initial state for the workspace store.
+ */
+// ts-unused-exports:disable-next-line
+export function makeWorkspaceContextInitialState(): WorkspaceContextStore {
```
snosenzo, line 20: Do we need the `WorkspaceContextStore` moved to `.d.ts` file so that we can extend the type?

snosenzo (approved):
this looks fine to me but you might want another set of eyes on it since I haven't worked in this area too much

packages/studio-base/src/providers/WorkspaceContextProvider.tsx
```diff
@@ -13,44 +13,53 @@ import {
 } from "@foxglove/studio-base/context/Workspace/WorkspaceContext";
 import { migrateV0WorkspaceState } from "@foxglove/studio-base/context/Workspace/migrations";
 
+/**
+ * Creates the default initial state for the workspace store.
+ */
+// ts-unused-exports:disable-next-line
```
jtbandes, line 19: Is this imported by the private repo? I think it'd be nicer to add it to studio-base/index and import it from there so that you don't need this disable rule

jtbandes (approved):
Seems fine to me as a direction we can try and easily tweak later if desired.

PR #6644, opened 2023-08-15
Plot x axis series data

alex-luci:
**User-Facing Changes**

**Description**
Add options to plot panel
- change point size
- plot current series x axis data

### DIFF
--- a/packages/studio-base/src/i18n/en/plot.ts
+++ b/packages/studio-base/src/i18n/en/plot.ts
@@ -33,5 +33,6 @@ export const plot = {
   index: "Index",
   currentPath: "Path (current)",
   accumulatedPath: "Path (accumulated)",
+  currentSeries: "Series (current)",
   secondsRange: "Range (seconds)",
 };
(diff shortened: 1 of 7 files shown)


alex-luci:
I would love the ability to be able to increase the size of the point for visual purposes, and need to be able to plot x axis data for multiple series. For example I'm plotting 2 joysticks, I need to be able to plot joystick1's x and y axis as well as joystick2's x and y axis. I added the settings options here and would love some input on suggestions for implementation

jtbandes:
Hi, thanks for the contribution!

First of all I would suggest separating the point size control from the other change so that we can review it individually.

Can you provide screenshots/example data to demonstrate the feature being added around the "current series x axis data"? From reading the description I'm not sure I understand what it is doing.

Also, just want to point out FYI that we have some currently ongoing work around plot dataset creation, #6584, which will probably cause conflicts with this work.

jtbandes:
Since we haven't heard back for a couple weeks I'm going to close this PR. However, please feel free to respond later or submit a new PR when you are ready.

Issue #6655, opened 2023-08-16
Plot setting to turn off line Invalid


smbera:
**Description**

- Version: 1.66.0-dev
- Platform: Google Chrome(115.0.5790.110)
- Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native):

**Steps To Reproduce**

https://github.com/foxglove/studio/assets/23401125/ad532921-c1b5-403a-b4f0-9b85f1fabd89

**Expected Behavior**

https://github.com/foxglove/studio/assets/23401125/10598b3c-e79c-4ae6-996e-1c3296787853

jtbandes:
This is the expected behavior. The button in the legend toggles visibility of the entire series, which is the same as the 👁️ button.

<img width="50" alt="image" src="https://github.com/foxglove/studio/assets/14237/56bea6b8-1042-405d-bf22-6c4935bb9d5b"> & <img width="67" alt="image" src="https://github.com/foxglove/studio/assets/14237/354302c6-424e-44cc-a7c2-38c6f5a6ae7e">

This is different from the "show lines" setting which only toggles visibility of the _line_, not all the data points.
<img width="195" alt="image" src="https://github.com/foxglove/studio/assets/14237/345c1733-b739-4ede-8bab-c8bb279dad0d"> <img width="222" alt="image" src="https://github.com/foxglove/studio/assets/14237/6bc9ad0a-2456-4e39-a115-f4aef6738ab3">

PR #6656, opened 2023-08-16
<fix>: Plot setting to turn off line invalid

smbera:
**User-Facing Changes**

Plot setting to turn off line invalid

**Description**

Here is the Issues for this PR. https://github.com/foxglove/studio/issues/6655


jtbandes:
This change is incorrect. The button in the legend toggles visibility of the entire series, which is the same as the 👁️ button.

image & image

This is different from the "show lines" setting which only toggles visibility of the _line_, not all the data points.
image image

PR #6657, opened 2023-08-16
foxglove websocket: Transfer arrayBuffer messages from worker

achim-k:
**User-Facing Changes**
None

**Description**
From https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Transferable_objects

> For example, when an [ArrayBuffer](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer) is transferred between threads, the memory resource that it points to is literally moved between contexts in a fast and efficient zero-copy operation

### DIFF
--- a/packages/studio-base/src/players/FoxgloveWebSocketPlayer/worker.ts
+++ b/packages/studio-base/src/players/FoxgloveWebSocketPlayer/worker.ts
@@ -15,9 +15,9 @@ export type FromWorkerMessage =
 
 let ws: WebSocket | undefined = undefined;
 
-const send = (msg: FromWorkerMessage, transfer?: Transferable[]): void => {
-  self.postMessage(msg, transfer ?? []);
-};
+const send: (message: FromWorkerMessage) => void = self.postMessage;
+const sendWithTransfer: (message: FromWorkerMessage, transfer: Transferable[]) => void =
+  self.postMessage;
 
 self.onmessage = (event: MessageEvent<ToWorkerMessage>) => {
   const { type, data } = event.data;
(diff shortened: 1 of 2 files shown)


jtbandes (approved):
You could also stick with the same `"message"` op for both, just don't pass a list of transferables in the case of text messages.

packages/studio-base/src/players/FoxgloveWebSocketPlayer/worker.ts
```diff
@@ -42,10 +42,13 @@ self.onmessage = (event: MessageEvent<ToWorkerMessage>) => {
...
-            type: "message",
-            data: wsEvent.data,
-          });
+          send(
+            {
+              type: "message",
+              data: wsEvent.data,
+            },
+            wsEvent.data instanceof ArrayBuffer ? [wsEvent.data] : [],
+          );
```
achim-k, line 51: @jtbandes is this what you suggesteded? Is there a way to test if objects are actually transferred rather than copied?

jtbandes: After the `send` you can check whether you are still able to access the arraybuffer data or not. I believe it will appear as an 0-length buffer after the transfer.

packages/studio-base/src/players/FoxgloveWebSocketPlayer/worker.ts
```diff
@@ -15,8 +15,8 @@ export type FromWorkerMessage =
 
 let ws: WebSocket | undefined = undefined;
 
-const send = (msg: FromWorkerMessage): void => {
-  self.postMessage(msg);
+const send = (msg: FromWorkerMessage, transfer?: Transferable[]): void => {
+  self.postMessage(msg, transfer ?? []);
```
jtbandes, line 19: is `?? []` necessary? just wondering if we can avoid the extra array allocation since this is called frequently

jtbandes (approved):

packages/studio-base/src/players/FoxgloveWebSocketPlayer/worker.ts
achim-k: Was necessary due to the overloaded `self.postMessage` function. Created a new `sendWithTransfer` function for that

PR #6661, opened 2023-08-16
Improve warning message for empty MCAP files

jtbandes:
**User-Facing Changes**
Improved warning message when loading an empty MCAP file.

**Description**

Before | After
--- | ---
image image | image image

Since McapUnindexedIterableSource already reads all the messages up front, we can easily check whether there are actually zero messages.

For the Quick Look preview, we can check for a Statistics record. If it's present, and it says `message_count: 0`, then we can trust the file is empty.


wkalt (approved):

packages/mcap-support/src/TempBuffer.ts
```diff
@@ -0,0 +1,39 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { IWritable, McapTypes } from "@mcap/core";
+
+export class TempBuffer implements McapTypes.IReadable, IWritable {
```
wkalt, line 7: if this is meant to be used as a test utility, might be worth marking it so, or if not maybe worth a comment what it is/how it grows the buffer.

packages/studio-base/src/players/IterablePlayer/Mcap/McapUnindexedIterableSource.ts
```diff
@@ -209,11 +211,18 @@ export class McapUnindexedIterableSource implements IIterableSource {
...
-    });
+    if (messageCount === 0) {
+      problems.push({
+        message: "This file contains no messages.",
+        severity: "warn",
+      });
+    } else {
+      problems.push({
+        message: "This file is unindexed. Unindexed files may have degraded performance.",
+        tip: "See the mcap spec: https://mcap.dev/specification/index.html#summary-section",
```
wkalt, line 222: do we use mcap or MCAP?

packages/studio-base/src/players/IterablePlayer/Mcap/McapUnindexedIterableSource.ts
```diff
@@ -209,11 +211,18 @@ export class McapUnindexedIterableSource implements IIterableSource {
       });
     }
 
-    problems.push({
-      message: "This file is unindexed. Unindexed files may have degraded performance.",
-      tip: "See the mcap spec: https://mcap.dev/specification/index.html#summary-section",
-      severity: "warn",
-    });
+    if (messageCount === 0) {
+      problems.push({
```
snosenzo, line 215: So `McapUnindexedIterableSource` handles all empty mcap files? That seems a little unintuitive and feels like an unspecified requirement coupled with `tryCreateIndexedReader` behavior. Should we add a comment there that it should return undefined for empty files?

jtbandes: added comment

snosenzo (approved):

PR #6665, opened 2023-08-17
Support `step` for raw images

jtbandes:
**User-Facing Changes**
The `step` field of `foxglove.RawImage` and `sensor_msgs/Image` is now respected.

**Description**
Adds support for `step` to all the decoder functions and adds unit tests.

Also adds sanity checks for the `step` value to ensure that it is large enough.

### DIFF
--- a/packages/den/image/decodings.ts
+++ b/packages/den/image/decodings.ts
@@ -50,8 +50,9 @@ export function decodeUYVY(
 
   // populate 2 pixels at a time
   for (let row = 0; row < height; row++) {
+    const rowStart = row * step;
     for (let col = 0; col < width; col += 2) {
-      const off = row * step + col * 2;
+      const off = rowStart + col * 2;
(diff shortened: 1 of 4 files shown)


packages/den/image/decodings.ts
```diff
@@ -37,177 +37,221 @@ function yuvToRGBA8(
...
-    const b = rgba[inIdx++]!;
-    const a = rgba[inIdx++]!;
-
-    output[outIdx++] = r;
-    output[outIdx++] = g;
-    output[outIdx++] = b;
-    output[outIdx++] = a;
+  for (let row = 0; row < height; row++) {
+    for (let col = 0; col < width; col++) {
+      const inIdx = row * step + col * 4;
```
foxymiles, line 132: Do you think there would be any performance benefit in hoisting `row * step` up to the outer loop?

foxymiles (approved):

PR #6667, opened 2023-08-18
Plot support custom function editor

smbera:
**User-Facing Changes**

Introduces a setting for the Plot panel, to custom function editor:
image

**Description**
For curve data, support function input (using Javascript interpreted language), for example: multiplying the curve by a constant coefficient, binary operation on the curve.

For example:
The topic name of the curve data is /hal/sensor/imu_with_odometry, and the imu.angular_velocity_x data in it is binary processed. The goal is to find the time period when the angular velocity in the x-axis is greater than 0.2 rad/s. To achieve this function, you can enter the following code in the function box:
```
return value > 0.2 ? 1 : 0;
```

For example:
Since the values of curve 1 and curve 2 differ greatly, it is difficult to observe and analyze them in the same window. It is necessary to reduce the values of curve 1 by a factor of 10. To achieve this function, you can enter the following code in the function box:
```
return value / 10;
```

https://github.com/foxglove/studio/assets/23401125/4fde1c7d-b8a9-4d90-8dfb-88ae94ea090e

### DIFF
--- a/packages/studio-base/src/components/SettingsTreeEditor/FieldEditor.tsx
+++ b/packages/studio-base/src/components/SettingsTreeEditor/FieldEditor.tsx
@@ -241,6 +241,7 @@ function FieldInput({
           placeholder={field.placeholder}
           InputProps={{
             readOnly: field.readonly,
+            multiline: field.multiline,
           }}
           onChange={(event) =>
             actionHandler({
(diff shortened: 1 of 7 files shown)


jtbandes:
Thank you for the proposal. We are aware that many users are interested in custom plot expressions. However, we are not currently prepared to commit to a solution and to maintain this feature as proposed. (Additionally, we are currently in the middle of some significant performance work on the Plot panel that needs to settle before we will consider such a feature.) The current recommended workflow for custom functions is to create them as [User Scripts](https://foxglove.dev/docs/studio/panels/user-scripts).

PR #6375, opened 2023-07-06
Add Service Panel

itzhaolin:
**User-Facing Changes**

**Description**
This is helpful for foxglove-bridge.
image

PR #6669, opened 2023-08-18
Add panel to call services

achim-k:
**User-Facing Changes**
Add panel to call services

**Description**
Adds a panel to call services (if the data source supports it). Look and feel is similar to the publish panel

Closes #6375
Fixes FG-4155


packages/studio-base/src/panels/CallService/index.tsx
```diff
@@ -0,0 +1,229 @@
...
+  };
+});
+
+function parseInput(value: string): { error?: string; parsedObject?: unknown } {
+  let parsedObject;
+  let error = undefined;
+  try {
+    const parsedAny: unknown = JSON.parse(value);
+    if (Array.isArray(parsedAny)) {
+      error = "Message content must be an object, not an array";
```
jtbandes, line 80: nit: all these "Message content" might want to rename them to "Request content"?

packages/studio-base/src/panels/CallService/settings.ts
```diff
@@ -0,0 +1,123 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { produce } from "immer";
+import { isEqual, set } from "lodash";
+import { useCallback, useEffect, useMemo } from "react";
+
+import { Immutable, SettingsTreeAction, SettingsTreeNodes } from "@foxglove/studio";
+import buildSampleMessage from "@foxglove/studio-base/panels/Publish/buildSampleMessage";
```
jtbandes, line 10: let's avoid importing from the Publish panel – we can move this function to some common/shared location

packages/studio-base/src/panels/CallService/settings.ts
```diff
@@ -0,0 +1,123 @@
...
+        help: "Optional request schema, used to pre-populate the request",
+        items: schemaNames,
+        value: config.datatype ?? "",
+      },
+      layout: {
+        label: "Layout",
+        input: "toggle",
+        options: [
+          { label: "vertical", value: "vertical" },
+          { label: "horizontal", value: "horizontal" },
```
jtbandes, line 54:
```suggestion
          { label: "Vertical", value: "vertical" },
          { label: "Horizontal", value: "horizontal" },
```

packages/studio-base/src/panels/CallService/settings.ts
```diff
@@ -0,0 +1,123 @@
...
+}
+
+const buildSettingsTree = (
+  config: CallServiceConfig,
+  schemaNames: string[],
+): SettingsTreeNodes => ({
+  general: {
+    fields: {
+      serviceName: {
+        label: "Service",
```
jtbandes, line 37: Maybe "Service name"?

packages/studio-base/src/panels/CallService/types.ts
```diff
@@ -0,0 +1,13 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+export type CallServiceConfig = {
+  serviceName?: string;
+  datatype?: string;
```
jtbandes, line 7: let's use `schemaName` or `requestSchemaName` or something like that

jtbandes (changes requested):
We should structure the panel as an extension panel, using PanelExtensionAdapter. Then you can use the panel extension API to implement the settings, instead of `usePanelSettingsTreeUpdate`.

packages/studio-base/src/panels/CallService/CallService.tsx
```diff
@@ -0,0 +1,247 @@
...
+  const [config, setConfig] = useState<Config>(() => ({
+    ...defaultConfig,
+    ...(context.initialState as Partial<Config>),
+  }));
+  const { classes } = useStyles({ buttonColor: config.buttonColor });
+
+  useEffect(() => {
+    context.saveState(config);
+    context.setDefaultPanelTitle(
+      config.serviceName === "" ? undefined : `Call service ${config.serviceName}`,
```
jtbandes, line 101: Want to avoid "Call service undefined":
image

```suggestion
      config.serviceName ? `Call service ${config.serviceName}` : undefined,
```

packages/studio-base/src/panels/index.ts
```diff
@@ -107,6 +107,13 @@ export const getBuiltin: (t: TFunction<"panels">) => PanelInfo[] = (t) => [
     thumbnail: publishThumbnail,
     module: async () => await import("./Publish"),
   },
+  {
+    title: t("callService"),
+    type: "CallService",
+    description: t("callServiceDescription"),
+    thumbnail: publishThumbnail,
```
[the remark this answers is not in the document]
jtbandes: I think it's ok to reuse the same one since the panels are similar. @2metres is the one who created the others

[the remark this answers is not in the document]
achim-k: @2metres, we decided to merge this PR as is with the publish thumbnail. Feel free to open a follow-up PR that adjusts the thumbnail

PR #6680, opened 2023-08-21
Improve performance of the state transitions panel on large messages

foxymiles:
**User-Facing Changes**
Improve performance of the state transitions panel on large messages

**Description**
Improve performance of the state transitions panel on large messages by taking advantage of our new sliced topic subscriptions.

### DIFF
--- a/packages/studio-base/src/PanelAPI/useBlocksSubscriptions.test.tsx
+++ b/packages/studio-base/src/PanelAPI/useBlocksSubscriptions.test.tsx
@@ -16,13 +16,14 @@ import { renderHook } from "@testing-library/react-hooks";
 import { cloneDeep } from "lodash";
 
 import MockMessagePipelineProvider from "@foxglove/studio-base/components/MessagePipeline/MockMessagePipelineProvider";
+import { SubscribePayload } from "@foxglove/studio-base/players/types";
 
 import * as PanelAPI from ".";
 
-describe("useBlocksByTopic", () => {
+describe("useBlocksSubscriptions", () => {
   it("returns an empty structure when there are no blocks", async () => {
-    const { result } = renderHook(({ topics }) => PanelAPI.useBlocksByTopic(topics), {
-      initialProps: { topics: ["/foo"] },
+    const { result } = renderHook(({ topics }) => PanelAPI.useBlocksSubscriptions(topics), {
+      initialProps: { topics: [{ topic: "/foo" }] },
       wrapper: ({ children }) => (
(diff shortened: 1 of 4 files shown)


packages/studio-base/src/PanelAPI/useBlocksSubscriptions.test.tsx
```diff
@@ -16,13 +16,14 @@ import { renderHook } from "@testing-library/react-hooks";
 import { cloneDeep } from "lodash";
 
 import MockMessagePipelineProvider from "@foxglove/studio-base/components/MessagePipeline/MockMessagePipelineProvider";
+import { SubscribePayload } from "@foxglove/studio-base/players/types";
 
 import * as PanelAPI from ".";
 
-describe("useBlocksByTopic", () => {
+describe("useBlocksSubscriptions", () => {
```
snosenzo, line 23: should we add some tests for `fields` subscriptions as well? Not seeing that represented here

packages/studio-base/src/panels/StateTransitions/index.tsx
```diff
@@ -196,7 +196,20 @@ const StateTransitions = React.memo(function StateTransitions(props: Props) {
 
   const decodeMessagePathsForMessagesByTopic = useDecodeMessagePathsForMessagesByTopic(pathStrings);
 
-  const blocks = useBlocksByTopic(subscribeTopics);
+  const subscriptions: SubscribePayload[] = useMemo(
+    () =>
+      filterMap(pathStrings, (path) => {
+        const payload = subscribePayloadFromMessagePath(path, "full");
+        // Include the header in case we are ordering by header stamp.
+        if (payload?.fields != undefined) {
```
snosenzo, line 204: In the plot panel do we always request header as well? or only when ordering by header stamp is active?

snosenzo (approved):
couple questions/consideration, otherwise looks good

packages/studio-base/src/PanelAPI/useBlocksSubscriptions.test.tsx
foxymiles: Yes good idea. I'll add some.

packages/studio-base/src/panels/StateTransitions/index.tsx
foxymiles: In plot we're a bit smarter and only add it if header stamp is requested. I'll rearrange this a bit to use the same logic.

PR #6686, opened 2023-08-22
Table panel: Add support for displaying typed arrays

achim-k:
**User-Facing Changes**
None

**Description**
Adds support for displaying typed arrays as single paginated column in the table panel. Before, the table implementation created a column for every array element which would crash the app for large arrays (e.g. image data)

~Depending on the array size, expanding the row can take some time. This should ideally not depend on the array size since the values are paginated :thinking:~

[Screencast from 23.08.2023 13:59:42.webm](https://github.com/foxglove/studio/assets/9250155/bbce5e62-6849-4b5a-b632-796e2229afc5)

fixes FG-4575

### DIFF
--- a/packages/studio-base/src/panels/Table/Table.tsx
+++ b/packages/studio-base/src/panels/Table/Table.tsx
@@ -37,6 +37,31 @@ import TableCell from "./TableCell";
 import { sanitizeAccessorPath } from "./sanitizeAccessorPath";
 import { CellValue } from "./types";
 
+type TypedArray =
+  | Int8Array
+  | Uint8Array
+  | Uint8ClampedArray
+  | Int16Array
+  | Uint16Array
+  | Int32Array
+  | Uint32Array
+  | Float32Array
+  | Float64Array;
+
+function isTypedArray(value: unknown): value is TypedArray {
+  return (
+    value instanceof Uint8Array ||
+    value instanceof Uint8ClampedArray ||
+    value instanceof Int8Array ||
+    value instanceof Uint16Array ||
+    value instanceof Int16Array ||
+    value instanceof Uint32Array ||
+    value instanceof Int32Array ||
+    value instanceof Float32Array ||
+    value instanceof Float64Array
+  );
+}
+
 const useStyles = makeStyles<void, "tableData" | "tableHeader">()((theme, _params, classes) => ({
   table: {
     border: "none",
@@ -111,7 +136,7 @@ const columnHelper = createColumnHelper<CellValue>();
 
 function getColumnsFromObject(val: CellValue, accessorPath: string, iconButtonClasses: string) {
   const obj = val.toJSON?.() ?? val;
-  if (ArrayBuffer.isView(obj) && !(obj instanceof DataView)) {
+  if (isTypedArray(obj)) {
     return [
       columnHelper.display({
         id: "typedArray",
@@ -119,7 +144,7 @@ function getColumnsFromObject(val: CellValue, accessorPath: string, iconButtonCl
         enableSorting: false,
         enableColumnFilter: false,
         cell: ({ row }) => {
-          return <span>{`${(obj as Uint8Array)[row.index]}`}</span>;
+          return <span>{`${obj[row.index]}`}</span>;
         },
       }),
     ];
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/panels/Table/Table.tsx
```diff
@@ -111,6 +111,19 @@ const columnHelper = createColumnHelper<CellValue>();
...
   const obj = val.toJSON?.() ?? val;
+  if (ArrayBuffer.isView(obj) && !(obj instanceof DataView)) {
+    return [
+      columnHelper.display({
+        id: "typedArray",
+        header: "",
+        enableSorting: false,
+        enableColumnFilter: false,
+        cell: ({ row }) => {
+          return <span>{`${(obj as Uint8Array)[row.index]}`}</span>;
```
achim-k, line 122: Didn't find away to avoid the cast, also had to do the cast below. Is there a way to avoid it?

foxymiles: You could try a type predicate function: https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

snosenzo: Does this assume it's always a `Uint8Array`? Couldn't it be any other typedArray?

snosenzo:
> Depending on the array size, expanding the row can take some time. This should ideally not depend on the array size since the values are paginated 🤔

if you do a perf trace on what's taking so long what does it reveal?

achim-k:
> > Depending on the array size, expanding the row can take some time. This should ideally not depend on the array size since the values are paginated 🤔
> 
> if you do a perf trace on what's taking so long what does it reveal?

Most of the time is spent [creating rows for every array element](https://github.com/TanStack/table/blob/9ef186f53dbad3047e8cc968a665794de7c44314/packages/table-core/src/utils/getCoreRowModel.ts#L24-L74). I'm not sure how to avoid this, ideally these rows should only be created lazily when in the current page view

snosenzo (approved):
> Most of the time is spent [creating rows for every array element](https://github.com/TanStack/table/blob/9ef186f53dbad3047e8cc968a665794de7c44314/packages/table-core/src/utils/getCoreRowModel.ts#L24-L74). I'm not sure how to avoid this, ideally these rows should only be created lazily when in the current page view

There's nothing in the API that we can tweak to prevent this?

foxymiles:
Thanks for adding this. On my machine this is extremely slow for something like `/CAM_BACK_LEFT/image_rect_compressed`. Is there any way we can truncate this by default and click to expand the full message? Considering practical use cases people probably aren't going to want to page way past the first page of binary elements right?

achim-k:
> Thanks for adding this. On my machine this is extremely slow for something like `/CAM_BACK_LEFT/image_rect_compressed`. Is there any way we can truncate this by default and click to expand the full message? Considering practical use cases people probably aren't going to want to page way past the first page of binary elements right?

I'm working on this right now and I got something working with "controlled pagination". In the end this means that you only pass in a subset of the data (the one that is on the current page) to the `table` api

achim-k:
This is now working smoother after changing the pagination to controlled pagination:

[Screencast from 23.08.2023 13:59:42.webm](https://github.com/foxglove/studio/assets/9250155/bbce5e62-6849-4b5a-b632-796e2229afc5)

packages/studio-base/src/panels/Table/Table.tsx
```diff
@@ -194,22 +228,40 @@ export default function Table({
     return getColumnsFromObject(maybeMessage as CellValue, accessorPath, classes.iconButton);
   }, [accessorPath, classes.iconButton, value]);
 
-  const data = React.useMemo(() => (Array.isArray(value) ? value : [value]), [value]);
+  const [{ pageIndex, pageSize }, setPagination] = React.useState<PaginationState>({
+    pageIndex: 0,
+    pageSize: 30,
```
snosenzo, line 233: Do we want to set this lower (5) by default so that a single element doesn't take up the size of a page by default?

snosenzo (approved):

packages/studio-base/src/panels/Table/Table.tsx
achim-k: I also think `30` is too big. We use `10` in the stories, I'll change it to that

achim-k:
```suggestion
    pageSize: 10,
```

PR #6687, opened 2023-08-22
Support x axis min, max and range in state transitions panel

foxymiles:
**User-Facing Changes**
Support x axis min, max and range in state transitions

**Description**
The interface and implementation is analogous to the same feature in the plot panel.

Screenshot 2023-08-22 at 8 41 58 AM

### DIFF
--- a/packages/studio-base/src/panels/StateTransitions/index.tsx
+++ b/packages/studio-base/src/panels/StateTransitions/index.tsx
@@ -168,6 +168,10 @@ function selectCurrentTime(ctx: MessagePipelineContext) {
   return ctx.playerState.activeData?.currentTime;
 }
 
+function selectEndTime(ctx: MessagePipelineContext) {
+  return ctx.playerState.activeData?.endTime;
+}
+
 type Props = {
   config: StateTransitionConfig;
   saveConfig: SaveConfig<StateTransitionConfig>;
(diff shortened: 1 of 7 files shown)


packages/studio-base/src/panels/StateTransitions/index.tsx
```diff
@@ -222,6 +228,32 @@ const StateTransitions = React.memo(function StateTransitions(props: Props) {
...
   }, [decodedBlocks, itemsByPath]);
 
+  const { minTime, maxTime } = useMemo(() => {
+    if (config.xAxisRange != undefined) {
+      return endTime
+        ? {
+            minTime: subtractTimes(endTime, fromSec(config.xAxisRange)),
+            maxTime: endTime,
+          }
+        : {};
```
snosenzo, line 238: What's the reason for counting back from endTime instead of counting forward from startTime?

snosenzo (approved):

packages/studio-base/src/panels/StateTransitions/index.tsx
foxymiles: In the fixed range case we want the X latest seconds of data.

snosenzo: oh right, that makes sense 👍

jtbandes (approved):

PR #6688, opened 2023-08-22
Fix handling of schemaless JSON messages

foxymiles:
**User-Facing Changes**
Fix handling of schemaless JSON messages. Regressed in https://github.com/foxglove/studio/pull/5817

**Description**
Revert the code removal and fix all resulting issues.

### DIFF
--- a/packages/studio-base/src/components/MessagePathSyntax/useCachedGetMessagePathDataItems.ts
+++ b/packages/studio-base/src/components/MessagePathSyntax/useCachedGetMessagePathDataItems.ts
@@ -230,6 +230,8 @@ export function getMessagePathDataItems(
     }
     const pathItem = filledInPath.messagePath[pathIndex];
     const nextPathItem = filledInPath.messagePath[pathIndex + 1];
+    const structureIsJson =
+      structureItem.structureType === "primitive" && structureItem.primitiveType === "json";
     if (!pathItem) {
       // If we're at the end of the `messagePath`, we're done! Just store the point.
       let constantName: string | undefined;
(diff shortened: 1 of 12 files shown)


jtbandes (commented):
This doesn't seem like the right approach to me. Most of the code here is referring to `primitiveType: "json"` in a schema – which was removed because it was a cruise-ism leftover from webviz code – but the issue we're trying to fix is for data with _no schema_. Is there a place that these code paths also handle the lack of a schema?

foxymiles:
> This doesn't seem like the right approach to me. Most of the code here is referring to `primitiveType: "json"` in a schema – which was removed because it was a cruise-ism leftover from webviz code – but the issue we're trying to fix is for data with _no schema_ (something webviz was never built to handle). Is there a place that these code paths also handle the lack of a schema?

Not sure. There's a lot of complicated logic here so I opted for just undoing the change that broke it.

foxymiles:
BTW this does fix the originally reported customer issue.

packages/studio-base/src/components/MessagePathSyntax/useCachedGetMessagePathDataItems.ts
```diff
@@ -243,12 +245,21 @@ export function getMessagePathDataItems(
...
       // If the `pathItem` is a name, we're traversing down using that name.
       const next = structureItem.nextByName[pathItem.name];
-      traverse(value[pathItem.name], pathIndex + 1, `${path}.${pathItem.repr}`, next);
-    } else if (pathItem.type === "slice" && structureItem.structureType === "array") {
+      const nextStructIsJson = next?.structureType === "primitive" && next.primitiveType === "json";
+
+      const actualNext: MessagePathStructureItem =
+        !nextStructIsJson && next
+          ? next
+          : { structureType: "primitive", primitiveType: "json", datatype: "" };
```
jtbandes, line 253: I think this is the critical part of the change that makes the plotting work. If the `next` is not found (which it won't be when there are no schemas), we fall back to this fake `primitiveType: "json"` thing.

I would look at trying to keep this traversal working without the fallback to `"json"`. That would mean revisiting the `if (... structureItem == undefined) { return }` at the top of the `traverse` function.

foxymiles: I think @cfoust has major changes to this code coming in https://github.com/foxglove/studio/pull/6584 so maybe I'll hold off on trying to change this until that lands.

foxymiles: We don't expose this `json` type as an actual schema item and the code doesn't require a `json` schema to work. It's just an internal representation of an unstructured object. If we remove it we'll have to replace it with some similar concept for the other code to work anyway.

PR #6689, opened 2023-08-22
Fix playback ui not updating during player seek backfill

snosenzo:
**User-Facing Changes**

 - Fix playback UI not updating while loading

**Description**

Looks like the `isPlaying` variable change when play/pause is called is never being emitted because the player is in `seek-backfill` state. I've updated play/pause to `queueEmitState` when they don't call `setState`. This will update the UI while the Player is in `seek-backfill`. 

FG-4474


foxymiles (approved):

jtbandes (approved):
What could possibly go wrong?

jtbandes:
Can this be added to IterablePlayer unit tests?

snosenzo:
> Can this be added to IterablePlayer unit tests?

@jtbandes Added one for setting play, but struggling to get the pause one working. For the final `store.reset` if I put `reset(2)` it gives me an error that it's getting 3 and if i put `reset(3)` it's getting a timeout. Any thoughts/suggestions?

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
```diff
@@ -1047,6 +1063,7 @@ export class IterablePlayer implements Player {
     await this.#playbackIterator?.return?.();
     this.#playbackIterator = undefined;
     await this.#iterableSource.terminate?.();
+    this.#resolveIsClosed(true);
```
jtbandes, line 1066: maybe leave out the `true` i.e. use `Promise<void>`?

jtbandes (approved):

PR #6690, opened 2023-08-23
replace ramda usage with lodash or vanilla JS

jtbandes:
This is an attempt to reduce the scope of #6584 so we don't have to take on a new dependency and coding style while also making other large changes (since the rest of this repo isn't currently using ramda). Most of the functions used had direct equivalents in lodash or could be replaced by vanilla JS (either built-in methods or for loops). Some notable things to call out are: different `zip` semantics, and ~lack of an equivalent for `mergeWith` as far as I could find~.


jtbandes:
Closing this, however I think we can learn some patterns from it:

- JS builtins like Math.min/max and Object.keys should be preferred over R.min/max/keys when possible
- `R.chain` should be replaced with `filterMap` when possible

PR #6692, opened 2023-08-23
Subscribe only to the fields necessary for plotting

cfoust:
**User-Facing Changes**
None.

**Description**
Integrate message field slicing into the dataset PR.

### DIFF
--- a/packages/studio-base/src/PanelAPI/useBlocksSubscriptions.ts
+++ b/packages/studio-base/src/PanelAPI/useBlocksSubscriptions.ts
@@ -34,7 +34,10 @@ export type MessageBlock = Immutable<{
 // That said, MessageBlock identity will change when the set of topics changes, so consumers should
 // prefer to use the identity of topic-block message arrays where possible.
 const filterBlockByTopics = memoizeWeak(
-  (block: Immutable<PlayerMessageBlock> | undefined, topics: readonly string[]): MessageBlock => {
+  (
+    block: Immutable<PlayerMessageBlock> | undefined,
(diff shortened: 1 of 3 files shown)


foxymiles (commented):
This seems to overlap quite a bit with the code we have here:

https://github.com/foxglove/studio/blob/caleb/23-08-23/feat/use-message-slicing/packages/studio-base/src/components/MessagePipeline/subscriptions.ts

https://github.com/foxglove/studio/blob/caleb/23-08-23/feat/use-message-slicing/packages/studio-base/src/players/subscribePayloadFromMessagePath.ts

Can we consolidate this logic?

snosenzo:
Only reading from the current playhead when using a plot panel that wasn't in the original loaded layout:

https://github.com/foxglove/studio/assets/10187776/7983ea43-cf19-4a44-87f0-5dbcedfbd984

Seeking after playback just clears the plot again

snosenzo:
Seeing some rubberbanding happening when switching between tabs with plots.
Studio Prod: 
https://github.com/foxglove/studio/assets/10187776/396a89f0-d8d1-44b4-b253-61bf224abd27

This deployment:
https://github.com/foxglove/studio/assets/10187776/9bae9472-ed76-455e-ac7b-c6b589fc05ce

Might be not such a big issue, but thought I'd mention

jtbandes:
It seems like switching tabs breaks preloading, and even reopening the data source does not fix it, only a full page refresh fixes it:

https://github.com/foxglove/studio/assets/14237/45ce1ce1-8653-4098-be1b-2c4f237149f9

PR #6693, opened 2023-08-24
Improve memory usage on large messages in the state transitions panel.

foxymiles:
**User-Facing Changes**
Improve memory usage on large messages in the state transitions panel.

**Description**
During playback the StateTransition panel accumulates currentFrame messages via the useMessagesByPath hook. This PR expands topic slicing to that hook so that we only accumulate the portions of messages we actually need.

### DIFF
--- a/packages/studio-base/src/PanelAPI/useMessagesByTopic.test.tsx
+++ b/packages/studio-base/src/PanelAPI/useMessagesByTopic.test.tsx
@@ -34,6 +34,20 @@ describe("useMessagesByTopic", () => {
     expect(result.current).toEqual({ "/foo": [] });
   });
 
+  it("accepts SubscribePayloads", async () => {
+    const { result } = renderHook(
+      ({ topics, historySize }) => PanelAPI.useMessagesByTopic({ topics, historySize }),
+      {
+        initialProps: { topics: [{ topic: "/foo", fields: ["a"] }], historySize: 1 },
+        wrapper: ({ children }) => (
+          <MockMessagePipelineProvider>{children}</MockMessagePipelineProvider>
+        ),
+      },
+    );
+
+    expect(result.current).toEqual({ "/foo": [] });
+  });
+
   it("add messages to their respective arrays", () => {
     const message1: MessageEvent = {
       topic: "/foo",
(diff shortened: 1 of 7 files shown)


packages/studio-base/src/PanelAPI/useMessagesByTopic.test.tsx
```diff
@@ -34,6 +34,20 @@ describe("useMessagesByTopic", () => {
...
+      {
+        initialProps: { topics: [{ topic: "/foo", fields: ["a"] }], historySize: 1 },
+        wrapper: ({ children }) => (
+          <MockMessagePipelineProvider>{children}</MockMessagePipelineProvider>
+        ),
+      },
+    );
+
+    expect(result.current).toEqual({ "/foo": [] });
+  });
```
jtbandes, line 49: What is this test testing? It would be nice to actually see a "sliced" object show up in the results to be sure it works

packages/studio-base/src/players/subscribePayloadFromMessagePath.ts
```diff
@@ -28,12 +28,12 @@ export function subscribePayloadFromMessagePath(
...
   if (!firstField) {
-    return { topic: parsedPath.topicName, preloadType };
+    return { topic: parsedPath.topicName, preloadType: preloadType ?? "partial" };
   }
 
   return {
     topic: parsedPath.topicName,
-    preloadType,
+    preloadType: preloadType ?? "partial",
     fields: [firstField.name],
```
jtbandes, line 37: What's the reason for this change? Is there a problem with passing an `undefined` value downstream?

jtbandes (approved):

packages/studio-base/src/PanelAPI/useMessagesByTopic.test.tsx
foxymiles: The actual slicing happens at the datasource level so the responsibility of this hook is just to pass through the subscription correctly.

packages/studio-base/src/players/subscribePayloadFromMessagePath.ts
foxymiles: We initially made `preloadType` optional just to limit the scope of the PR but I'd like to start taking steps towards making it required and explicit.

PR #6653, opened 2023-08-16
Add fields to TopicList

2metres:
**User-Facing Changes**
The topic list is now a searchable tree with fields under the Topic

PR #6695, opened 2023-08-24
Add drag & drop support to Topics sidebar

jtbandes:
**User-Facing Changes**
None (will announce later after #6653 is merged)

**Description**

Adds ability to drag & drop topics from the Topics sidebar onto the Image, Raw Message, and Table panels.

https://github.com/foxglove/studio/assets/14237/32c4861d-061a-4a17-909e-1eeda9333fac

### DIFF
--- a/packages/studio-base/src/components/StatsChip.tsx
+++ b/packages/studio-base/src/components/StatsChip.tsx
@@ -1,0 +1,55 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { Divider, Paper } from "@mui/material";
+import { makeStyles } from "tss-react/mui";
+
+const useStyles = makeStyles()((theme) => ({
+  root: {
+    display: "flex",
+    borderColor: theme.palette.action.selected,
+    borderRadius: "1em",
+    color: theme.palette.text.secondary,
+    backgroundColor: theme.palette.background.paper,
+
+    [`@container (max-width: 320px)`]: {
+      display: "none",
+    },
+  },
+  stat: {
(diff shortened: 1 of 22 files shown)


packages/studio-base/src/components/StatsChip.tsx
```diff
@@ -0,0 +1,55 @@
...
+      paddingInlineEnd: theme.spacing(0.75),
+    },
+  },
+  divider: {
+    borderColor: theme.palette.action.selected,
+    marginInline: theme.spacing(0.5),
+  },
+}));
+
+export function StatsChip({ topicName }: { topicName: string }): JSX.Element {
```
snosenzo, line 42: Why `Stats`?
Also should this have a story? I know it's in workspace.stories, but wondering if it could be helpful to have one on its own

packages/studio/src/index.ts
```diff
@@ -238,6 +238,35 @@ export type RenderState = {
...
+};
+
+export type MessagePathDropStatus = {
+  /** True if the panel would be able to accept this dragged message path. */
+  canDrop: boolean;
+  /**
+   * Indicate the type of operation that would occur if this path were dropped. Used to change the
+   * mouse cursor.
+   */
+  effect?: "replace" | "add";
```
snosenzo, line 255: Why not rename this to `cursor` to expose the mouse cursor style directly?

jtbandes: I think a bit of abstraction away from the cursor allows us to change our design style later (maybe showing some text or color instead of or in addition to a different cursor) without changing the API

snosenzo (approved):
Looks good to me

packages/studio-base/src/components/StatsChip.tsx
jtbandes: Covered generally under TopicList stories

PR #6700, opened 2023-08-25
Refurbish panel action overlays to match new ones

2metres:
**User-Facing Changes**
None

**Description**
Clean up panel action overlays to look like new drop targets

### When multiple panels are selected

| dark | light |
|---|---|
| Screen Shot 2023-08-28 at 3 30 04 pm | Screen Shot 2023-08-28 at 3 29 59 pm |

### When <code>` or  ~</code> is pressed

| dark | light |
|---|---|
| Screen Shot 2023-08-28 at 3 29 42 pm | Screen Shot 2023-08-28 at 3 29 54 pm |


packages/studio-base/src/components/PanelOverlay.tsx
```diff
@@ -0,0 +1,247 @@
...
+  isValidTarget: boolean;
+  quickActionsKeyPressed: boolean;
+  quickActionsOverlayRef: React.MutableRefObject<HTMLElement | ReactNull>;
+  selectedPanelCount: number;
+  connectOverlayDragSource: (el: HTMLElement | ReactNull) => void;
+  createTabs: () => void;
+  groupPanels: () => void;
+  removePanel: () => void;
+  splitPanel: () => void;
+};
```
jtbandes, line 139: This is a huge number of props and makes the overlay component highly coupled to the panel component. I think it would be better to make the overlay a "dumb" component that just displays whatever buttons or text you tell it to display. i.e. give it a prop for `buttons: [{title: "Split", action: () => {}} ...]` or something like that. This would also make it easier to add stories for the overlay component.

amacneil:
LR feedback:
- Looks good
- Button corners could use less aggressive rounding

2metres:
@foxymiles can you take a look at this failing test. It looks like JSDOM is choking on css somewhere

packages/studio-base/src/components/Panel.tsx
```diff
@@ -598,73 +643,18 @@ export default function Panel<
...
-                    message={dropMessage}
-                  />
-                )}
-                {type !== TAB_PANEL_TYPE && quickActionsKeyPressed && !fullscreen && (
-                  <div
-                    className={classes.actionsOverlay}
+                {!fullscreen && type !== TAB_PANEL_TYPE && (
+                  <PanelOverlay
+                    {...panelOverlayProps}
+                    dropMessage={dropMessage}
```
jtbandes, line 649: any reason for dropMessage to be out here all alone and not part of panelOverlayProps?

jtbandes (approved):
Nice!

PR #6704, opened 2023-08-29
Allow arbitrary precision in settings number inputs

foxymiles:
**User-Facing Changes**
Allow arbitrary precision in number inputs in settings fields.

**Description**
Instead of clamping these fields to 2 digits of precision by default allow arbitrary precision but remove trailing zeroes.

### DIFF
--- a/packages/studio-base/src/components/SettingsTreeEditor/inputs/NumberInput.tsx
+++ b/packages/studio-base/src/components/SettingsTreeEditor/inputs/NumberInput.tsx
@@ -81,7 +81,7 @@ export function NumberInput(
     onChange,
     disabled,
     readOnly,
-    precision = 2,
+    precision = 100,
   } = props;
 
   const inputRef = useRef<HTMLInputElement>(ReactNull);
@@ -143,7 +143,7 @@ export function NumberInput(
   );
 
   const displayValue =
-    inputRef.current === document.activeElement ? value : value?.toFixed(precision);
+    inputRef.current === document.activeElement ? value : Number(value?.toFixed(precision));
 
   return (
     <TextField
--- a/packages/studio-base/src/panels/Plot/index.stories.tsx
+++ b/packages/studio-base/src/panels/Plot/index.stories.tsx
@@ -493,7 +493,13 @@ export const LineGraphWithSettings: StoryObj = {
     return (
       <PlotWrapper
         pauseFrame={pauseFrame}
-        config={{ ...exampleConfig, minYValue: -1, maxYValue: 1, minXValue: 0, maxXValue: 3 }}
+        config={{
+          ...exampleConfig,
+          minYValue: -3.1415,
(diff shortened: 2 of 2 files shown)


foxymiles:
> From reading the code I’m not really sure how this is supposed to behave; can we add unit tests with examples?

I think the stories cover the possibilities pretty well but I'll add some tests too.

packages/studio-base/src/components/SettingsTreeEditor/inputs/formatNumber.ts
```diff
@@ -0,0 +1,20 @@
...
+ */
+export function formatNumber(n: undefined, precision: number): undefined;
+export function formatNumber(n: number, precision: number): number;
+export function formatNumber(n: undefined | number, precision: number): undefined | number;
+export function formatNumber(n: undefined | number, precision: number): undefined | number {
+  if (n == undefined) {
+    return undefined;
+  }
+
+  return Number(n.toFixed(precision));
```
jtbandes, line 19: How does this differ from lodash's `round(number, precision)`?

jtbandes (commented):
I notice that when I use the mouse to adjust the ymin/max values, I end up with a lot of random decimal places, is there any way to avoid this? Maybe limit the precision when scrubbing with the mouse?

https://github.com/foxglove/studio/assets/14237/9fca8127-764b-41b5-bdfd-2b04274ddae9

packages/studio-base/src/components/SettingsTreeEditor/inputs/formatNumber.ts
foxymiles: I guess we could also use `_.round(n, 100)` here to the same effect.

foxymiles:
> I notice that when I use the mouse to adjust the ymin/max values, I end up with a lot of random decimal places, is there any way to avoid this? Maybe limit the precision when scrubbing with the mouse?

Yeah we could try limiting the precision. Not sure what would be a good value for that though.

foxymiles:
The lodash implementation of round is a bit 🤯 :
https://github.com/lodash/lodash/blob/2da024c3b4f9947a48517639de7560457cd4ec6c/.internal/createRound.js

foxymiles:
I've rounded to 4 digits of precision on scrubbing. Subjectively feels ok I think.

packages/studio-base/src/components/SettingsTreeEditor/inputs/NumberInput.tsx
```diff
@@ -135,15 +135,19 @@ export function NumberInput(
           0.1 *
           step *
           scale;
-        scrubValue.current += delta;
+        scrubValue.current = round(scrubValue.current + delta, 4);
```
snosenzo, line 138: nit: Pull `4` out into named constant?

snosenzo (approved):

PR #6710, opened 2023-08-30
Log panel: Allow to filter by name

achim-k:
**User-Facing Changes**
Log panel: Allow to filter by name

**Description**
Allows to filter log messages by publisher name

image

Resolves FG-4172


packages/studio-base/src/panels/Log/index.tsx
```diff
@@ -113,11 +94,67 @@ const LogPanel = React.memo(({ config, saveConfig }: Props) => {
...
+      ),
+    });
+  }, [
+    actionHandler,
+    availableTopics,
+    topicToRender,
+    nameFilter,
+    updatePanelSettingsTree,
+    seenNodeNames,
+    seenNodeNames.size, // Needed as seenNodeNames doesn't
```
snosenzo, line 144: unfinished comment

snosenzo (approved):
The behavior of accumulating the node names is a little strange, but I'm happy to see it remembers their status after reloading. 
Wondering if we also want the Log Settings panel to have a search-for-settings filter (like the 3D panel) at the top in case the name list is really long?
image

Otherwise I think it looks good, but it sounds like design/other folks might also want a pass

packages/studio-base/src/panels/Log/filterMessages.test.ts
```diff
@@ -121,8 +129,19 @@ describe("filter", () => {
...
         }),
       ).toEqual(ros1msgs);
     });
+
+    it("should filter messages by name", () => {
+      expect(
+        filterMessages(ros1msgs, {
+          minLogLevel,
+          searchTerms: [],
+          nameFilter: { "/some_topic": { visible: false } },
```
snosenzo, line 142: another test with visible:true?

amacneil:
LR feedback:
- Drop node autocomplete from filter box

Feedback on filter:
- Move log level filter dropdown to sidebar
- Move search filter to top of panel (below panel titlebar)
- Remove copy button

packages/studio-base/src/i18n/en/log.ts
```diff
@@ -5,4 +5,7 @@
 export const log = {
   topicError: "Topic {{topic}} is not available",
   topic: "Topic",
+  nameFilter: "Name filter",
```
amacneil, line 8:
```suggestion
  nameFilter: "Namespaces",
```

achim-k:
Addressed all comments, this one is ready for another review

image

packages/studio-base/src/panels/Log/FilterBar.tsx
```diff
@@ -53,85 +19,21 @@ type Filter = {
...
-          onClick={() => {
-            void clipboard.copy(JSON.stringify(props.messages, undefined, 2) ?? "");
-          }}
-          title="Copy log to clipboard"
-        >
-          <CopyAllIcon />
-        </ToolbarIconButton>
-      </Stack>
-    </Stack>
+    <FilterTagInput
```
snosenzo, line 28: now that its only this, would it make more sense to pull this out and use this directly in the parent container?

snosenzo (approved):
Tested, looks good to me 👍

packages/studio-base/src/panels/Log/FilterBar.tsx
achim-k: Yes, I can do that

snosenzo (approved):

packages/studio-base/src/panels/Log/index.stories.tsx
```diff
@@ -319,3 +319,22 @@ export const FoxgloveLog: StoryObj = {
     </PanelSetup>
   ),
 };
+
+export const FilterByName: StoryObj = {
+  render: () => (
+    <PanelSetup fixture={fixture}>
```
jtbandes, line 325: It would be nice to show what the settings tree looks like in these stories too (with some filters enabled/disabled). You can add `includeSettings` to the PanelSetup

jtbandes (approved):
This story seems either broken or no longer relevant:
image

PR #6711, opened 2023-08-30
3d Panel: Editable Transforms now compose quaternions instead of adding them

snosenzo:
**User-Facing Changes**

 - "Rotation offset" fields in transform editing now compose with the original rotation instead of adding

**Description**

Double checked that the RPY changes are affecting the same axes as they did previously. Open to considering better names for "Rotation offset" in the settings panel and `offsetEulerDegrees` variable names since they're not an additive "offset" anymore

Related: FG-2355


jtbandes (approved):
Would it make sense to try and test this in storybook or maybe even a unit test since this lives in CoordinateFrame?

cc @banisadr to help evaluate whether we are ok with this as a breaking change. I think it's probably not a big deal since I believe the offset is only a runtime/transient thing and not saved to the layout, right?

snosenzo:
> I believe the offset is only a runtime/transient thing and not saved to the layout

I think this is incorrect. If you edit the transform and reload the page, the information is persisted across reloads

jtbandes:
In that case, should we change the name of the settings field so that we don't reinterpret existing settings values with a different meaning but rather just drop them?

PR #6715, opened 2023-08-31
Make user script bottom bar resizable

foxymiles:
**User-Facing Changes**
Make user script bottom bar resizable

**Description**
Uses the [react-resizable-panels](https://github.com/bvaughn/react-resizable-panels) library to make the bottom bar resizable.

https://github.com/foxglove/studio/assets/93935560/9304be6c-c93c-4914-97cb-1ce453a62251

@2metres Can you help with styling? When the user can resize to reveal the bottom bar at any time it doesn't make sense to have no tab selected IMO. Also the resize handle could be better.

FG-3816

### DIFF
--- a/packages/studio-base/package.json
+++ b/packages/studio-base/package.json
@@ -154,6 +154,7 @@
     "react-monaco-editor": "0.54.0",
     "react-mosaic-component": "6.0.1",
     "react-refresh-typescript": "2.0.9",
+    "react-resizable-panels": "0.0.55",
     "react-resize-detector": "9.1.0",
     "react-transition-group": "4.4.5",
     "react-use": "17.4.0",
--- a/packages/studio-base/src/panels/NodePlayground/BottomBar/LogsSection.tsx
+++ b/packages/studio-base/src/panels/NodePlayground/BottomBar/LogsSection.tsx
@@ -19,6 +19,7 @@ const useStyles = makeStyles()({
 });
 
 const LogsSection = ({ logs }: { logs: readonly UserNodeLog[] }): JSX.Element => {
+  // Manage auto-scroll behavior when user is also manually scrolling the list.
   const [autoScroll, setAutoScroll] = useState(true);
 
   const { classes } = useStyles();
--- a/packages/studio-base/src/panels/NodePlayground/BottomBar/LogsSection.tsx
+++ b/packages/studio-base/src/panels/NodePlayground/BottomBar/LogsSection.tsx
@@ -3,13 +3,56 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
 import { Link, List, ListItem, ListItemButton, ListItemText, Typography } from "@mui/material";
+import { useEffect, useRef, useState } from "react";
 import Tree from "react-json-tree";
+import { makeStyles } from "tss-react/mui";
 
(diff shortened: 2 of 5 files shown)


2metres:
Looks good. Gave it a bit of a styling pass. We should use this on the plot legend too!

packages/studio-base/src/panels/NodePlayground/BottomBar/index.tsx
```diff
@@ -132,7 +121,7 @@ const BottomBar = ({ nodeId, isSaved, save, diagnostics, logs }: Props): ReactEl
               }
               value="diagnostics"
               data-testid="np-errors"
-              onClick={() => handleClick("diagnostics")}
+              onClick={() => handleClick()}
```
snosenzo, line 124: nit: 
```suggestion
              onClick={handleClick} 
```
(or alternatively `onChangeTab` since handleClick just wraps that)
and below

packages/studio-base/src/panels/NodePlayground/BottomBar/LogsSection.tsx
```diff
@@ -3,13 +3,56 @@
...
 
+const useStyles = makeStyles()({
+  list: {
+    height: "100%",
+    overflowY: "auto",
+  },
+});
+
 const LogsSection = ({ logs }: { logs: readonly UserNodeLog[] }): JSX.Element => {
+  const [autoScroll, setAutoScroll] = useState(true);
```
snosenzo, line 22: Can we get some comments around the expected behavior of autoscroll? I think that might better clarify what this is trying to do

snosenzo (approved):

packages/studio-base/package.json
```diff
@@ -151,6 +151,7 @@
     "react-monaco-editor": "0.54.0",
     "react-mosaic-component": "6.0.1",
     "react-refresh-typescript": "2.0.9",
+    "react-resizable-panels": "0.0.55",
```
jtbandes, line 154: We've already got `react-mosaic` which we use in a few places already to provide resize functionality – why not use that here too?

packages/studio-base/src/panels/NodePlayground/BottomBar/index.tsx
```diff
@@ -76,35 +67,33 @@ const useStyles = makeStyles()((theme) => ({
...
-    }
-  }, [autoScroll, logs]);
-
   return (
     <>
-      <Paper elevation={0}>
+      <Paper
+        elevation={0}
+        style={{ height: "100%", overflowY: "hidden", display: "flex", flexDirection: "column" }}
+      >
```
2metres, line 96: Can you move these out of inline styles into the useStyles

2metres (approved):

packages/studio-base/package.json
jtbandes: I think it creates a weird inconsistency when we use multiple resizing libraries throughout the app.

2metres: I don't think this is the same thing as react-mosaic though. When it's just a sidebar/drawer within a panel we use something like this. when it's part of the global layout is when mosaic gets used. In the case of the PlotLegend we have rolled our own drawer.

PR #6716, opened 2023-08-31
Changes for supporting new omgidl versions with unions

snosenzo:
**User-Facing Changes**

 - `omgidl` schema encoding now supports,   XCDR1 encoding,  and schemas that use unions, and enumerator overrides

**Description**

These dependency updates bring all the changes described [here](https://github.com/foxglove/omgidl/pull/67)

Resolves: FG-4635


packages/mcap-support/src/parseChannel.ts
```diff
@@ -25,6 +25,36 @@ export type ParsedChannel = {
...
+    if (definition.aggregatedKind === "union") {
+      const innerDefs: MessageDefinitionField[] = definition.cases.map((caseDefinition) => ({
+        ...caseDefinition.type,
+        predicates: caseDefinition.predicates,
+      }));
+
+      if (definition.defaultCase != undefined) {
+        innerDefs.push(definition.defaultCase);
+      }
+      const { cases: _c, defaultCase: _dC, ...rest } = definition;
```
jtbandes, line 45: I generally think it's nicer to destructure & use explicitly the properties that we are interested in rather than doing this underscore thing to omit some stuff.

jtbandes (approved):

PR #6717, opened 2023-09-01
Allow searching and drag & drop of schema fields in Topics sidebar

jtbandes:
**User-Facing Changes**
Added the ability to search for message fields in the Topics sidebar, and drag & drop fields onto panels to visualize them. Currently, drag & drop is supported for the Raw Message, Image, Plot, State Transitions, and Table panels.

**Description**
Resolves FG-4348
Resolves FG-3941

https://github.com/foxglove/studio/assets/14237/4aa2fc22-ed68-4f08-9c73-bb1020170bfb


2metres:
Looking good. Truncation is a bit of a pain on long topics and message names.

Should we ignore non alphanumeric characters in the search?

snosenzo:
Searching for a topic name that  also overlaps with a lot fields can be tedious (like searching for a `point`-named topic in a data source with many annotations), should we allow collapsing/expanding of topics with a lot of field matches on them? or alternatively have a toggle for also searching fields?

packages/studio-base/src/panels/Plot/index.tsx
```diff
@@ -112,6 +113,32 @@ function Plot(props: Props) {
...
+        }
+        return { canDrop: true, effect: "add" };
+      },
+      handleDrop(path) {
+        saveConfig((prevConfig) => ({
+          ...prevConfig,
+          paths: [
+            // If there was only a single series and its path was empty (the default state of the
+            // panel), replace the series rather than adding to it
+            ...(prevConfig.paths.length === 1 && prevConfig.paths[0]?.value === ""
```
snosenzo, line 132: Worth pulling out into a function? this feels like it would be useful elsewhere to know that the panel is in the default "empty" state

packages/studio/src/index.ts
```diff
@@ -243,6 +243,10 @@ export type DraggedMessagePath = {
   path: string;
   /** The schema name of the top-level topic being dragged */
   rootSchemaName: string | undefined;
+  /** True if the path represents a whole topic (no message path component). */
+  isTopic: boolean;
+  /** True if the path represents a primitive value inside a message. */
+  isLeaf: boolean;
```
snosenzo, line 249: `isLeaf` feels like it's referring to an implementation detail -- maybe just `isPrimitive`?

snosenzo (approved):
Code looks good, just some thoughts about behavior and naming

packages/studio-base/src/panels/Plot/index.tsx
jtbandes: I'd actually be interested in revisiting this default state, see FG-4841

snosenzo (approved):

PR #6718, opened 2023-09-01
Fix topic stats disappearing when filtering the topic list

jtbandes:
**User-Facing Changes**

**Description**

Fixes FG-4839

The DirectTopicStatsUpdater didn't handle when rows were dynamically added or removed from the topic list. This PR fixes the issue by using a MutationObserver to see when nodes are added/removed from the parent element to trigger updates.

This is also a must-have for #6717 because we're going to virtualize the list using `react-window`.

https://github.com/foxglove/studio/assets/14237/8d0bb595-34c7-4691-8edd-64ea4db335d6

</td><td>

https://github.com/foxglove/studio/assets/14237/85b3c054-3863-49c7-bd2d-b71363d149b8

</td></tr></table>

### DIFF
--- a/packages/studio-base/src/components/DirectTopicStatsUpdater.tsx
+++ b/packages/studio-base/src/components/DirectTopicStatsUpdater.tsx
@@ -80,6 +80,26 @@ export function DirectTopicStatsUpdater({ interval = 1 }: { interval?: number })
     });
   }, [latestFrequenciesByTopic, latestStats]);
 
+  // Update when new "data-topic" nodes are added, to support virtualized lists and filtering.
+  useEffect(() => {
+    if (!rootRef.current?.parentElement) {
+      return;
+    }
+    const observer = new MutationObserver((mutations) => {
+      for (const mutation of mutations) {
+        for (const node of mutation.addedNodes) {
+          // updateStats() triggers mutations of text nodes, so only update if HTMLElements are added to avoid infinite loops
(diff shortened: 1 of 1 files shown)


foxymiles:
Do you think it's worth trying this instead of the DirectTopicStatsUpdater?

https://docs.pmnd.rs/zustand/recipes/recipes#transient-updates-(for-frequent-state-changes)

packages/studio-base/src/hooks/useDirectTopicStatsUpdate.ts
```diff
@@ -0,0 +1,59 @@
...
+  const latestFrequenciesByTopic = useLatest(frequenciesByTopic);
+
+  useMessagePipelineSubscription(
+    useCallback(
+      (state) => {
+        const topicStats = state.playerState.activeData?.topicStats;
+        if (!topicStats) {
+          return;
+        }
+        if (updateCount.current++ % UPDATE_INTERVAL_FRAMES !== 0) {
```
snosenzo, line 38: This feels really strange, what are we trying to avoid here? Would a debounce suffice?

jtbandes: Debounce might work indeed. This was already the approach being used so I just copied it.

foxymiles (approved):

PR #6719, opened 2023-09-01
Fix visibility filter padding in panel settings

achim-k:
**User-Facing Changes**
None

**Description**
Fixes the top padding for the visibility filter as remarked in https://github.com/foxglove/studio/pull/6710#issuecomment-1701816475

**Before**:

   image

**After**:

image


foxymiles:
Is that screenshot from before or after the fix?

achim-k:
> Is that screenshot from before or after the fix?

Was before the fix. Have adapted the PR description and added a screenshot after the fix

foxymiles (approved):

PR #6723, opened 2023-09-01
Upgrade Yarn to v3.6.3

amacneil:
**User-Facing Changes**
N/A

**Description**
- Upgrade Yarn to v3.6.3
- Upgrade Yarn plugins

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,8 +1,8 @@
+.yarn/**/*.cjs filter=lfs diff=lfs merge=lfs -text
 * text=auto eol=lf
 *.afdesign filter=lfs diff=lfs merge=lfs -text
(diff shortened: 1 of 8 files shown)


.gitattributes
```diff
@@ -1,3 +1,5 @@
+.yarn/plugins/**/*.cjs filter=lfs diff=lfs merge=lfs -text
+.yarn/releases/**/*.cjs filter=lfs diff=lfs merge=lfs -text
```
jtbandes, line 2: FYI we already have a `*.cjs` rule below.

jtbandes (approved):

.gitattributes
amacneil: Oh interesting. I will remove that in favor one more explicitly scoped to `.yarn`.

PR #6732, opened 2023-09-05
Add delete button to legend row on plot panel

2metres:
**User-Facing Changes**
Adds delete button to plot panel legend

Screen Shot 2023-09-05 at 12 33 12 pm

### DIFF
--- a/packages/studio-base/src/panels/Plot/PlotLegendRow.tsx
+++ b/packages/studio-base/src/panels/Plot/PlotLegendRow.tsx
@@ -2,8 +2,13 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
-import { ErrorCircle16Filled, Square24Filled, Square24Regular } from "@fluentui/react-icons";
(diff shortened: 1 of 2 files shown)


jtbandes:
- I find it weird that the button has no hover/active styling (same applies to the toggle button on the left)
- The button is present but doesn't work when there is only 1 series left (see also FG-4841)

amacneil:
👍

packages/studio-base/src/panels/Plot/PlotLegendRow.tsx
```diff
@@ -211,6 +232,22 @@ export function PlotLegendRow({
...
         </div>
       )}
+      <div>
+        <ButtonBase
+          title="Delete series"
+          aria-label="Delete series"
+          className={classes.removeButton}
+          onClick={() => {
+            const newPaths = paths.slice();
+            if (newPaths.length > 1) {
```
jtbandes, line 242: Seems like we should allow deleting the last one – at least it should work the same as the X button in the settings does.

jtbandes (approved):

PR #6733, opened 2023-09-05
Fix for single-message plots

cfoust:
**User-Facing Changes**
Fix a small issue affecting single-message plots.

**Description**
When a plot only uses a single message (ie its x-axis is either a custom data source or indices) we should not use our debounced rebuilding mechanism.


snosenzo (approved):

jtbandes (approved):
What's wrong with downsampling a single-message plot? Is that something the `downsample` function would be expected to handle or is it expected behavior that it doesn't work and that's why we want to avoid it? More detail on this could be helpful to add to that code comment for future readers.

cfoust:
Downsampling single-message plots should work just fine, but doing so has some architectural implications that I'll have to come back and adjust for later.

PR #6734, opened 2023-09-05
Don't track other pointer events in NumberInput

cfoust:
**User-Facing Changes**
One less bug.

**Description**
The mouse event handlers in `NumberInput` did not track whether a click and drag had initiated in the input element or not. This fixes that.

https://github.com/foxglove/studio/assets/4588553/7c787585-30cc-4b46-ad30-3b4eb4d13564

### DIFF
--- a/packages/studio-base/src/components/SettingsTreeEditor/inputs/NumberInput.tsx
+++ b/packages/studio-base/src/components/SettingsTreeEditor/inputs/NumberInput.tsx
@@ -130,10 +130,6 @@ export function NumberInput(
     event.currentTarget.releasePointerCapture(event.pointerId);
   }, []);
 
-  const onPointerLeave = useCallback(() => {
-    isDragging.current = false;
-  }, []);
-
   const onPointerMove = useCallback(
     (event: React.PointerEvent<HTMLInputElement>) => {
       if (event.buttons !== 1 || !isDragging.current) {
@@ -172,7 +168,6 @@ export function NumberInput(
         onPointerDown,
         onPointerUp,
         onPointerMove,
-        onPointerLeave,
       }}
       InputProps={{
         readOnly,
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/components/SettingsTreeEditor/inputs/NumberInput.tsx
```diff
@@ -124,24 +126,26 @@ export function NumberInput(
   );
 
   const onPointerUp = useCallback((event: React.PointerEvent) => {
+    isDragging.current = false;
     event.currentTarget.releasePointerCapture(event.pointerId);
   }, []);
 
+  const onPointerLeave = useCallback(() => {
+    isDragging.current = false;
```
jtbandes, line 134: Good find, thanks for fixing this issue!

How come dragging still works when I move the mouse outside of the element? Looking at this line of code I would've expected it to stop as soon as the mouse leaves the element's bounding box, but it works. Maybe there's something I don't understand about pointer events.

https://github.com/foxglove/studio/assets/14237/d7b0e8c2-f6ca-4587-b6ac-01a1a02913a6

cfoust: I just played with this a little bit and at least to me, it feels more natural this way. (Note: this does not differ from the behavior in `main`, I just tried it out!) Admittedly the browser behavior here is strange: why are we getting `pointermove` events from outside the bounds of the element?

cfoust: But you're right, the code suggests this will stop dragging after the pointer leaves, but it doesn't; that's misleading

jtbandes (approved):

PR #6735, opened 2023-09-05
Bump Prettier to v3.0.3

amacneil:
**User-Facing Changes**
N/A

**Description**
Bump Prettier to v3.0.3

Replaces https://github.com/foxglove/studio/pull/6720


.prettierrc.yaml
```diff
@@ -1,2 +1 @@
 printWidth: 100
-trailingComma: all
```
amacneil, line 2: This is the [default now](https://prettier.io/docs/en/options.html#trailing-commas)

.vscode/settings.json
```diff
@@ -16,7 +16,7 @@
   "files.insertFinalNewline": true,
   "files.trimFinalNewlines": true,
   "files.trimTrailingWhitespace": true,
-  "prettier.prettierPath": "./node_modules/prettier",
+  "prettier.prettierPath": "./node_modules/prettier/index.cjs",
```
amacneil, line 19: https://github.com/prettier/prettier-vscode/issues/3100#issuecomment-1660829933

jtbandes:
Let's wait for #6724

packages/studio-base/src/panels/NodePlayground/getPrettifiedCode.ts
```diff
@@ -15,9 +15,15 @@
 async function getPrettifiedCode(code: string): Promise<string> {
   // use dynamic imports to avoid loading prettier unless the function is invoked
   const prettier = await import("prettier/standalone");
-  const parserPlugin = await import("prettier/parser-typescript");
+  const estreePlugin = (await import(
+    "prettier/plugins/estree"
+  )) as typeof import("prettier/plugins/estree");
```
amacneil, line 20: `as typeof` was necessary to make typescript happy here, it seemed to incorrectly assume this plugin was under a `.default` property (e.g. cjs import?). But the runtime tests did not pass when using `.default`.

Feels like this should not be needed, but I'm not sure what I did wrong.

amacneil:
Sounds good. Just need this to align with private repos.

packages/studio-base/src/panels/NodePlayground/getPrettifiedCode.ts
jtbandes: What's the story behind the new plugins? This is just how it's done now in v3?

jtbandes (approved):
Thanks!

Manually tested User Script editor to make sure prettier still works in there.

packages/studio-base/src/panels/NodePlayground/getPrettifiedCode.ts
amacneil: Yeah this is just the v3 syntax. Looking at the package.json in the [npm module](https://www.npmjs.com/package/prettier?activeTab=code), the old parser-typescript path was returning cjs so I had to update it to the esm path. Then I got an error about estree plugin not included by default anymore, so I had to manually add it.

After that the tests ran fine (and correctly updated the snapshot with a trailing comma) so I'm fairly sure it is working as expected.

packages/studio-base/webpack.ts
```diff
@@ -181,7 +181,7 @@ export function makeConfig(
           // bundle it for the renderer.
           // https://github.com/microsoft/TypeScript/issues/39436
           // Prettier's TS parser also bundles the same code: https://github.com/prettier/prettier/issues/11076
-          test: /[\\/]node_modules[\\/]typescript[\\/]lib[\\/]typescript\.js$|[\\/]node_modules[\\/]prettier[\\/]parser-typescript\.js$/,
+          test: /[\\/]node_modules[\\/]typescript[\\/]lib[\\/]typescript\.js$|[\\/]node_modules[\\/]prettier[\\/]plugins[\\/]typescript\.m?js$/,
```
amacneil, line 184: @jtbandes do you know if having the prettier typescript plugin here is still required, or whether we test this string replace stuff anywhere? The issue referenced in that comment is marked as fixed.

jtbandes: I don't think we have a test as the problem being addressed here was some webpack build warnings (not errors).

PR #6756, opened 2023-09-05
Use `_.` for all lodash imports

jtbandes:
**User-Facing Changes**
None

**Description**
Since #6584 added Ramda with `R.*` imports, to improve consistency and avoid confusion we can use `_.` for all lodash imports.

This adds a new lint rule within the repo to enforce this import style (which supports auto-fixing).

Since using `import * as _ from "lodash"` increased the bundle size, I've also changed us over to `lodash-es` which results in a net _decrease_ in bundle size.

Relates to FG-4805

- [Before (8af796ff5)](https://github.com/foxglove/studio/actions/runs/6097415144/job/16545007074): `Total production build size: 131998171 bytes`
- [After](https://github.com/foxglove/studio/actions/runs/6101498121/job/16557952926?pr=6756): `Total production build size: 130341402 bytes `
- Difference: 130341402 - 131998171 = -1,656,769bytes


snosenzo (commented):
Looks like `_` isn't being picked up as an unused import if it's unused. I think we might have to add an exception for it to the unused variable starting with `_` rule.

snosenzo (approved):

PR #6765, opened 2023-09-06
Updated Chinese translation

Woodii1998:
**User-Facing Changes**

Updated Chinese translation

**Description**


Woodii1998:
@yujingz Can you proofread the Chinese translation?

packages/studio-base/src/panels/ThreeDeeRender/renderables/OccupancyGrids.ts
```diff
@@ -109,25 +110,25 @@ export class OccupancyGrids extends SceneExtension<OccupancyGridRenderable> {
...
-          label: "Color mode",
+          label: i18n.t("threeDee:colorMode"),
           input: "select",
           value: configWithDefaults.colorMode,
           options: [
-            { label: "Custom", value: "custom" },
-            { label: "Map", value: "map" },
+            { label: i18n.t("threeDee:custom"), value: "custom" },
+            { label: i18n.t("threeDee:map"), value: "map" },
             { label: "Costmap", value: "costmap" },
```
yujingz, line 119: I think we should add a comment here as why we don't translate costmap

yujingz (approved):

packages/studio-base/src/panels/ThreeDeeRender/renderables/OccupancyGrids.ts
jtbandes: I think that maybe "map" and "costmap" should both be untranslated, or both be translated. These are pre-defined color schemes that come from RViz: https://github.com/ros2/rviz/blob/rolling/docs/FEATURES.md#possible-color-schemes

jtbandes: I also think `映射` might be an incorrect translation — in English it's more literal like 地图. However I'm going to just change it to "Map".

packages/studio-base/src/panels/ThreeDeeRender/renderables/PoseArrays.ts
```diff
@@ -188,23 +183,40 @@ export class PoseArrays extends SceneExtension<PoseArrayRenderable> {
       const gradient = config.gradient ?? DEFAULT_GRADIENT_STR;
 
       const fields: SettingsTreeFields = {
-        type: { label: "Type", input: "select", options: TYPE_OPTIONS, value: displayType },
+        type: {
+          label: i18next.t("threeDee:type"),
+          input: "select",
+          options: [
+            { label: i18next.t("threeDee:poseDisplayTypeAxis"), value: "axis" },
```
snosenzo, line 190: can we use just `t(` here like we do elsewhere?

packages/studio-base/src/panels/ThreeDeeRender/renderables/OccupancyGrids.ts
```diff
@@ -109,25 +110,40 @@ export class OccupancyGrids extends SceneExtension<OccupancyGridRenderable> {
...
-          label: "Color mode",
+          label: i18n.t("threeDee:colorMode"),
           input: "select",
           value: configWithDefaults.colorMode,
           options: [
-            { label: "Custom", value: "custom" },
-            { label: "Map", value: "map" },
-            { label: "Costmap", value: "costmap" },
-            { label: "Raw", value: "raw" },
+            { label: i18n.t("threeDee:colorModeCustom"), value: "custom" },
```
snosenzo, line 117: just `t(` for consistency

packages/studio-base/src/panels/ThreeDeeRender/renderables/colorMode.ts
```diff
@@ -2,6 +2,7 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
+import i18n from "i18next";
```
snosenzo, line 5:
```suggestion
import  { t } from "i18next";
```

snosenzo (commented):
We should also update the PR title and description to reflect that japanese translations were updated as well

jtbandes:
The Japanese translations mostly aren’t updated, it just adds `undefined` to keep the formatting the same

packages/studio-base/src/panels/ThreeDeeRender/renderables/PoseArrays.ts
jtbandes: Some of these files already had variables named `t`, I guess we could rename those though

snosenzo:
> The Japanese translations mostly aren’t updated, it just adds `undefined` to keep the formatting the same

ah thanks for clarifying, skimmed over it maybe a little too hard 😅

jtbandes (approved):

PR #6767, opened 2023-09-06
Mark diagnostic messages as stale after a certain timeout

### LINKED ISSUE #6705: diagnostics won't go stale
**Description**

foxglove's diagnostics panel won't register a node as stale if it doesn't publish anymore. 

rqt_runtime_monitor works as intended

image

- Version: 1.67.0
- Platform: Windows Desktop & Browser

**Steps To Reproduce**

Stop publishing a diagnostic.

That node doesnt go to stale

**Expected Behavior**

That node should go to stale. Like in rqt_runtime_monitor

achim-k:
**User-Facing Changes**
Mark diagnostic messages as stale after a certain timeout

**Description**
Marks diagnostic messages as stale when no new messages with the same ID has been received within the configured timeout (default is 5 seconds)

Fixes #6705 
Resolves FG-4771

For reviewing purposes, you can use the following bag file that contains some diagnostic messages: [diagnostics_bag.zip](https://github.com/foxglove/studio/files/12540758/diagnostics.zip)

### DIFF
--- a/packages/studio-base/src/panels/diagnostics/DiagnosticStatusPanel.tsx
+++ b/packages/studio-base/src/panels/diagnostics/DiagnosticStatusPanel.tsx
@@ -14,11 +14,17 @@
 import { Autocomplete, TextField } from "@mui/material";
 import { produce } from "immer";
 import * as _ from "lodash-es";
-import { useCallback, useMemo, useEffect } from "react";
+import { useCallback, useMemo, useEffect, useState } from "react";
+import { useInterval } from "react-use";
 
+import { Time, subtract, compare } from "@foxglove/rostime";
 import { SettingsTreeAction } from "@foxglove/studio";
 import { useDataSourceInfo } from "@foxglove/studio-base/PanelAPI";
 import EmptyState from "@foxglove/studio-base/components/EmptyState";
+import {
+  MessagePipelineContext,
+  useMessagePipeline,
+} from "@foxglove/studio-base/components/MessagePipeline";
 import Panel from "@foxglove/studio-base/components/Panel";
 import { usePanelContext } from "@foxglove/studio-base/components/PanelContext";
 import PanelToolbar from "@foxglove/studio-base/components/PanelToolbar";
@@ -30,7 +36,12 @@ import DiagnosticStatus from "./DiagnosticStatus";
 import { buildStatusPanelSettingsTree } from "./settings";
 import useAvailableDiagnostics from "./useAvailableDiagnostics";
 import useDiagnostics from "./useDiagnostics";
-import { DiagnosticStatusConfig as Config, getDisplayName } from "./util";
+import {
+  DiagnosticStatusConfig as Config,
+  DEFAULT_SECONDS_UNTIL_STALE,
+  LEVELS,
+  getDisplayName,
+} from "./util";
 
 type Props = {
   config: Config;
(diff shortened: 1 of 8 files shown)


packages/studio-base/src/panels/diagnostics/DiagnosticStatusPanel.tsx
```diff
@@ -43,13 +54,32 @@ const ALLOWED_DATATYPES: string[] = [
...
+
+  // The stale time marks the point in time a diagnostic message is considered stale when its timestamp is older.
+  const newStaleTime =
+    currentTime && secondsUntilStale >= 1
+      ? subtract(currentTime, { sec: Math.floor(secondsUntilStale), nsec: 0 })
+      : undefined;
+  const [staleTime, setStaleTime] = useState<Time | undefined>(newStaleTime);
+  useInterval(() => {
+    setStaleTime(newStaleTime);
+  }, 1000);
```
achim-k, line 82: I introduced the `staleTime` state variable to avoid having to pass the frequently changing variable `currentTime` as dependency in the useMemo below. Not sure if it's better here to use a `useEffect` hook that sets the interval?

jtbandes: The `useMessagePipeline(selectCurrentTime)` will cause the component to re-render frequently which is not ideal for performance, so maybe you could use `useMessagePipelineGetter` instead. You might also want to wrap this logic in a custom hook like `useStaleTime(secondsUntilStale)` that will take care of this logic in a reusable way so it can be in both panels.

achim-k: Thanks for that hint! That's exactly what I want

packages/studio-base/src/panels/diagnostics/useStaleTime.ts
```diff
@@ -0,0 +1,39 @@
...
+  secondsUntilStale: number,
+  updateIntervalMillis?: number,
+): Time | undefined {
+  const messagePipeline = useMessagePipelineGetter();
+  const getCurrentTime = useCallback(() => {
+    const { playerState } = messagePipeline();
+    return playerState.activeData?.currentTime;
+  }, [messagePipeline]);
+
+  const [currentTime, setCurrentTime] = useState<Time | undefined>(getCurrentTime());
```
jtbandes, line 23:
```suggestion
  const [currentTime, setCurrentTime] = useState<Time | undefined>(() => getCurrentTime());
```

packages/studio-base/src/panels/diagnostics/useStaleTime.ts
```diff
@@ -0,0 +1,39 @@
...
+      setCurrentTime(getCurrentTime());
+    }, updateIntervalMillis ?? DEFAULT_UPDATE_INTERVAL_MS);
+    return () => {
+      clearInterval(interval);
+    };
+  }, [getCurrentTime, updateIntervalMillis]);
+
+  return useMemo(() => {
+    return currentTime && secondsUntilStale >= 1
+      ? subtract(currentTime, { sec: Math.floor(secondsUntilStale), nsec: 0 })
```
jtbandes, line 36: Could this throw an error if `currentTime < 0:00:01`?

packages/studio-base/src/panels/diagnostics/DiagnosticSummary.tsx
```diff
@@ -42,19 +42,22 @@ import { usePanelContext } from "@foxglove/studio-base/components/PanelContext";
...
   DiagnosticStatusConfig,
-  getDiagnosticsByLevel,
-  filterAndSortDiagnostics,
-  LEVEL_NAMES,
+  DiagnosticSummaryConfig,
   KNOWN_LEVELS,
+  LEVEL_NAMES,
+  filterAndSortDiagnostics,
+  getDiagnosticsByLevel,
+  getDiagnosticsWithStales as markOldDiagnosticsAsStales,
```
jtbandes, line 60: change these names to be the same?

jtbandes (approved):
I think this looks pretty good. I see now that updating periodically makes the results a little nondeterministic since it depends on when you play/pause, rather than depending completely on currentTime like your original version. However I think this is a reasonable tradeoff.

jtbandes:
Some more comments on user experience:

- The stale color is different between summary (gray) & detail (blue):  
   image

- In the detail view, when I seek the color doesn't immediately change back to green, it only changes after a delay:

   https://github.com/foxglove/studio/assets/14237/a097fa9f-5da5-4b41-9ed6-c64fd69f4d34

PR #6768, opened 2023-09-06
Fix broken Table panel for live data

cfoust:
**User-Facing Changes**
The Table panel did not work when using a live data source. Now it will.

**Description**
`react-table` requires stable cell rendering functions; for the library, a new function is equivalent to a new component, and it is duly remounted. React components that are unmounted and mounted too quickly do not have working `onclick` handlers (TIL).

This PR stabilizes the cell rendering functions so that we're not returning a new one on every render.

https://github.com/foxglove/studio/assets/4588553/bdcfd35e-2bd2-46c8-94c2-c8fd294bc389

### DIFF
--- a/packages/studio-base/src/panels/Table/Table.tsx
+++ b/packages/studio-base/src/panels/Table/Table.tsx
@@ -27,7 +27,9 @@ import {
   getExpandedRowModel,
   getSortedRowModel,
   useReactTable,
+  CellContext,
 } from "@tanstack/react-table";
+import memoizeWeak from "memoize-weak";
 import { makeStyles } from "tss-react/mui";
 
 import EmptyState from "@foxglove/studio-base/components/EmptyState";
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/panels/Table/Table.tsx
```diff
@@ -134,7 +136,46 @@ const useStyles = makeStyles<void, "tableData" | "tableHeader">()((theme, _param
...
+const memoizedCellRenderer = memoizeWeak((accessorPath: string, id: string) => {
+  return function ValueCell(info: CellContext<CellValue, unknown>) {
+    const value = info.getValue();
+    const row = info.row;
+    if (Array.isArray(value) && typeof value[0] !== "object") {
+      return JSON.stringify(value);
+    }
+
+    // eslint-disable-next-line no-restricted-syntax
+    if (typeof value === "object" && value != null) {
```
jtbandes, line 148:
```suggestion
    if (typeof value === "object" && value != undefined) {
```

jtbandes (approved):
Seems like this could've also been in a useMemo?

cfoust:
Pulling this out into a `useMemo` looked like it was nontrivial; this needs to be able to memoize the cell renderer in trees of arbitrary depth. it's conceptually similar but I didn't want to mess with the tree mapping too much; the `getColumnsFromObject` function is not actually a React component

PR #6769, opened 2023-09-07
Fix issue with gaps that can appear in plots under certain circumstances

cfoust:
**User-Facing Changes**
None.

**Description**
After seeking to the beginning of an unbounded plot, gaps would appear while playing in addition to errant lines going back to the origin.
https://github.com/foxglove/studio/assets/4588553/34c7ef8e-b058-4139-825f-d9fcd8de8372

It turned out that there were two reasons for this:
* There was a bug in the `sliceSingle` function that was resulting in the addition of erroneous datapoints at (0, 0).
* In my refactor of plot generation, I neglected to include support for the `{min,max}{X,Y}` and `followingViewWidth` config properties, which meant that the plot would fail to update when those parameters were changed. This is related because we should adjust the datasets (ie "current" or "block" data) we're using depending on the needs of the plot.

This PR fixes both of these things. Bounded and unbounded plots are now working correctly without artifacts.

### DIFF
--- a/packages/studio-base/src/panels/Plot/datasets.tsx
+++ b/packages/studio-base/src/panels/Plot/datasets.tsx
@@ -390,7 +390,7 @@ export function sliceTyped(dataset: TypedData[], start: number, end?: number): T
   return [sliceSingle(startSlice, offset0), ...between, sliceSingle(endSlice, 0, offset1)];
 }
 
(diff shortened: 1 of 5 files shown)


jtbandes (approved):
Can we add unit tests for these things?

cfoust:
Indeed we can!

PR #6770, opened 2023-09-07
Fix block data for user scripts

cfoust:
**User-Facing Changes**
None.

**Description**
We were incorrectly ignoring block subscriptions to topics defined in user scripts. Now we merge `SubscriberPayload`s before calculating the topics we need.

### DIFF
--- a/packages/studio-base/src/components/MessagePipeline/subscriptions.ts
+++ b/packages/studio-base/src/components/MessagePipeline/subscriptions.ts
@@ -2,10 +2,9 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
-import * as _ from "lodash-es";
 import moize from "moize";
+import * as R from "ramda";
 
-import { filterMap } from "@foxglove/den/collection";
 import { Immutable } from "@foxglove/studio";
 import { SubscribePayload } from "@foxglove/studio-base/players/types";
 
@@ -20,6 +19,66 @@ export function makeSubscriptionMemoizer(): (val: SubscribePayload) => Subscribe
 }
 
 /**
+ * Merge two SubscribePayloads, using either all of the fields or the union of
+ * the specific fields requested.
+ */
+function mergeSubscription(
+  a: Immutable<SubscribePayload>,
+  b: Immutable<SubscribePayload>,
+): Immutable<SubscribePayload> {
+  const isAllFields = a.fields == undefined || b.fields == undefined;
+  const fields = R.pipe(
+    R.chain((payload: Immutable<SubscribePayload>): readonly string[] => payload.fields ?? []),
+    R.map((v) => v.trim()),
+    R.filter((v: string) => v.length > 0),
+    R.uniq,
+  )([a, b]);
+
+  return {
+    ...a,
+    ...(fields.length > 0 && !isAllFields ? { fields } : {}),
+  };
+}
+
+/**
+ * Merge subscriptions that subscribe to the same topic, paying attention to
+ * the fields they need. This ignores `preloadType`.
+ */
+function denormalizeSubscriptions(
+  subscriptions: Immutable<SubscribePayload[]>,
+): Immutable<SubscribePayload[]> {
+  return R.pipe(
+    R.groupBy((v: Immutable<SubscribePayload>) => v.topic),
+    R.values,
+    // Filter out any set of payloads that contains _only_ empty `fields`
+    R.filter((payloads: Immutable<SubscribePayload[]> | undefined) => {
+      // Handle this later
+      if (payloads == undefined) {
+        return true;
+      }
+
+      return !R.all(
+        (v: Immutable<SubscribePayload>) => v.fields != undefined && v.fields.length === 0,
+        payloads,
+      );
+    }),
+    // Now reduce them down to a single payload for each topic
+    R.map((payloads: Immutable<SubscribePayload[]> | undefined): Immutable<SubscribePayload> => {
+      const first = payloads?.[0];
+      if (first == undefined) {
+        // by definition, we can only get here iff there is at least one
+        // payload
+        return {
+          topic: "/impossible",
+        };
+      }
+
+      return R.reduce(mergeSubscription, first, payloads ?? []);
+    }),
+  )(subscriptions);
+}
+
+/**
  * Merges individual topic subscriptions into a set of subscriptions to send on to the player.
  *
  * If any client requests a "whole" subscription to a topic then all fields will be fetched for that
@@ -28,40 +87,9 @@ export function makeSubscriptionMemoizer(): (val: SubscribePayload) => Subscribe
  */
 export function mergeSubscriptions(
   subscriptions: Immutable<SubscribePayload[]>,
-): Immutable<SubscribePayload>[] {
-  const fullSubsByTopic = new Map<string, Immutable<SubscribePayload>>();
-  const partialSubsByTopic = new Map<string, Immutable<SubscribePayload>>();
-  for (const sub of subscriptions) {
-    const target = sub.preloadType === "full" ? fullSubsByTopic : partialSubsByTopic;
-    const existing = target.get(sub.topic);
-    if (existing) {
-      if (existing.fields == undefined) {
-        // Nothing to do, already subscribed to the whole topic.
-      } else if (sub.fields == undefined) {
-        // Replace any slice subscription with a subscription to the whole topic.
-        target.set(sub.topic, sub);
-      } else {
-        // Skip slice subscriptions with no non-empty fields selected.
-        const nonEmptyFields = filterMap(sub.fields, (field) => {
-          const trimmed = field.trim();
-          return trimmed.length > 0 ? trimmed : undefined;
-        });
-        if (nonEmptyFields.length > 0) {
-          target.set(sub.topic, { ...sub, fields: _.union(existing.fields, nonEmptyFields) });
-        }
-      }
-    } else {
-      if (sub.fields == undefined) {
-        // If no subscription for this topic exists, register a whole topic subscription.
-        target.set(sub.topic, sub);
-      } else {
-        // Only register a slice sub if some fields are selected.
-        const hasNonEmptyField = sub.fields.some((field) => field.trim().length > 0);
-        if (hasNonEmptyField) {
-          target.set(sub.topic, sub);
-        }
-      }
-    }
-  }
-  return [...fullSubsByTopic.values(), ...partialSubsByTopic.values()];
+): Immutable<SubscribePayload[]> {
+  return R.pipe(
+    R.partition((v: Immutable<SubscribePayload>) => v.preloadType === "full"),
+    ([full, partial]) => [...denormalizeSubscriptions(full), ...denormalizeSubscriptions(partial)],
+  )(subscriptions);
 }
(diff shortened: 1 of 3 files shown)


snosenzo:
Can we use this `mergeSubscriptions` function and update the logic? https://github.com/foxglove/studio/blob/main/packages/studio-base/src/components/MessagePipeline/subscriptions.ts#L29

packages/studio-base/src/components/MessagePipeline/subscriptions.ts
```diff
@@ -19,6 +18,66 @@ export function makeSubscriptionMemoizer(): (val: SubscribePayload) => Subscribe
...
+      return !R.all(
+        (v: Immutable<SubscribePayload>) => v.fields != undefined && v.fields.length === 0,
+        payloads,
+      );
+    }),
+    // Now reduce them down to a single payload for each topic
+    R.map((payloads: Immutable<SubscribePayload[]> | undefined): Immutable<SubscribePayload> => {
+      const first = payloads?.[0];
+      if (first == undefined) {
+        // by definition, we can only get here iff there is at least one
```
snosenzo, line 69: Should probably throw an error

packages/studio-base/src/players/UserNodePlayer/index.test.ts
```diff
@@ -601,21 +601,7 @@ describe("UserNodePlayer", () => {
       });
       userNodePlayer.setSubscriptions(topicNames.map((topic) => ({ topic })));
       await delay(10); // wait for subscriptions to take effect
-      expect(fakePlayer.subscriptions).toEqual([{ topic: "/np_input" }]);
-    });
-
-    it("passes through sliced subscriptions", async () => {
```
snosenzo, line 607: You were correct that this is not the behavior that we're looking for, but instead of deleting it can we correct it?
This should be a test for `it("upgrades field-specific subscriptions to whole subscriptions if a user node is also subscribed to it")`, and there should be one combinations of both partial and full preload subscription types if there aren't already

packages/studio-base/src/players/UserNodePlayer/index.ts
```diff
@@ -1044,13 +1046,37 @@ export default class UserNodePlayer implements Player {
     const nodeSubscriptions: Record<string, SubscribePayload> = {};
     const realTopicSubscriptions: SubscribePayload[] = [];
 
+    // We need a list of all of the topics used and their preload mode,
```
snosenzo, line 1049: kinda feel like that we should take this whole `setSubscriptionsUnlocked` function out and test against it

packages/studio-base/src/components/MessagePipeline/subscriptions.ts
cfoust: Pure functions ideally should not throw errors--it's a side effect, after all

snosenzo: I'd rather have this fail loudly than silently. If there's an invariant here it's better we know about it sooner rather than have it be silent

jtbandes: If the side effect is that we find a bug, it’s a good side effect 😀

jtbandes:
> We were incorrectly ignoring block subscriptions to topics defined in user scripts.

We need to include a regression test for this. I haven't taken the time to do a full read through of the code changes yet, but I don't see any tests that prove we've fixed this bug.

packages/studio-base/src/players/UserNodePlayer/index.test.ts
```diff
@@ -641,15 +641,19 @@ describe("UserNodePlayer", () => {
...
       ]);
 
       // Wait for subscriptions to take effect.
       await delay(10);
 
       // The underlying player subscription should not be sliced since we don't know which fields of
       // the message the script will use.
-      expect(fakePlayer.subscriptions).toEqual([{ topic: "/np_input", preloadType: "partial" }]);
+      expect(fakePlayer.subscriptions).toEqual([
+        { topic: "/np_input", preloadType: "full" },
```
snosenzo, line 654: shouldn't this only have a full subscription? There's no player behavior change by also having the preloadType: "partial" description right?

packages/studio-base/src/players/UserNodePlayer/index.ts
```diff
@@ -1041,39 +1042,86 @@ export default class UserNodePlayer implements Player {
...
-      if (!inputs) {
-        nodeSubscriptions[subscription.topic] = subscription;
-        realTopicSubscriptions.push(subscription);
-        continue;
-      }
+    type InputMapping = [SubscribePayload, readonly string[] | undefined];
+
+    // Check all subscriptions against the list of topics this UserNodePlayer
+    // can provide
+    const inputs = R.pipe(
```
snosenzo, line 1049:
```suggestion
    const incomingSubsAndInputTopicsPair = R.pipe(
```

packages/studio-base/src/players/UserNodePlayer/index.ts
```diff
@@ -1041,39 +1042,86 @@ export default class UserNodePlayer implements Player {
...
+
+    // An array of all of the input topics used by the user nodes referenced by
+    // `subscriptions`
+    const inputTopics = R.pipe(
+      R.chain(([, v]: InputMapping): readonly string[] => v ?? []),
+      R.uniq,
+    )(inputs);
+
+    this.#nodeSubscriptions = R.pipe(
+      // Ignore all subscriptions that resolved to empty (but not undefined)
```
snosenzo, line 1062:
```suggestion
      // Ignore subscriptions to output topics of nodes with empty input arrays
```

packages/studio-base/src/players/UserNodePlayer/index.ts
```diff
@@ -1041,39 +1042,86 @@ export default class UserNodePlayer implements Player {
...
+    )(subscriptions);
+
+    // An array of all of the input topics used by the user nodes referenced by
+    // `subscriptions`
+    const inputTopics = R.pipe(
+      R.chain(([, v]: InputMapping): readonly string[] => v ?? []),
+      R.uniq,
+    )(inputs);
+
+    this.#nodeSubscriptions = R.pipe(
```
snosenzo, line 1061: add a comment at `#nodeSubscriptions` about how it's used to check whether output topics of user nodes are being subscribed to and whether those subs are preloaded. 

Could be helpful to add another comment here saying that `#nodeSubscriptions` should be a map of incoming subscriptions to consolidated preload values

packages/studio-base/src/players/UserNodePlayer/index.ts
```diff
@@ -1041,39 +1042,86 @@ export default class UserNodePlayer implements Player {
...
-        nodeSubscriptions[subscription.topic] = subscription;
-        realTopicSubscriptions.push(subscription);
-        continue;
-      }
+    type InputMapping = [SubscribePayload, readonly string[] | undefined];
+
+    // Check all subscriptions against the list of topics this UserNodePlayer
+    // can provide
+    const inputs = R.pipe(
+      mergeSubscriptions,
```
snosenzo, line 1050: is it necessary to mergeSubscriptions here? isn't this being done in the message pipeline store before getting here?

packages/studio-base/src/players/UserNodePlayer/index.ts
```diff
@@ -1041,39 +1042,86 @@ export default class UserNodePlayer implements Player {
...
-      for (const inputTopic of inputs) {
-        realTopicSubscriptions.push({
-          topic: inputTopic,
-          preloadType: subscription.preloadType ?? "partial",
-        });
-      }
-    }
+    // Merge subscriptions we pass on to the underlying player.
+    this.#player.setSubscriptions(
+      R.pipe(
```
snosenzo, line 1085: would prefer this to be assigned to a constant with a descriptive name (like `realTopicSubscriptions`) before being passed to the underlying player's `setSubscriptions`

packages/studio-base/src/players/UserNodePlayer/index.ts
```diff
@@ -1041,39 +1042,86 @@ export default class UserNodePlayer implements Player {
...
+          // If the inputs array is empty then we don't have anything to
+          // subscribe to for this output
+          if (topics.length === 0) {
+            return [];
+          }
+
+          // Subscribe to all fields for all topics used by this user script
+          // because we can't know what fields the user script actually uses
+          // (for now)
+          return R.map(
```
snosenzo, line 1114: prefer normal TS `topics.map`

packages/studio-base/src/players/UserNodePlayer/index.ts
```diff
@@ -1041,39 +1042,86 @@ export default class UserNodePlayer implements Player {
...
-    this.#nodeSubscriptions = nodeSubscriptions;
+            return [subscription];
+          }
 
-    // Merge subscriptions we pass on to the underlying player.
-    const mergedSubscriptions = mergeSubscriptions(realTopicSubscriptions);
-    this.#player.setSubscriptions(mergedSubscriptions);
+          // If the inputs array is empty then we don't have anything to
+          // subscribe to for this output
+          if (topics.length === 0) {
```
snosenzo, line 1107: Maybe makes more sense to filter this out earlier so that we don't have to account for it in both `nodeSubscriptions` and `realTopicSubscriptions`

snosenzo (commented):
would also be good to have a regression test that catches this bug

packages/studio-base/src/players/UserNodePlayer/index.test.ts
cfoust: We do not currently have any behavior here that gets rid of matching pairs of `full` and `partial` subscriptions (if `fields` matches) even though a `full` subscription implies a `partial` subscription. A plot panel, for example, subscribes to both. In theory we could adjust `mergeSubscriptions` to be smarter about this but it doesn't change much in practice.

packages/studio-base/src/players/UserNodePlayer/index.ts
```diff
@@ -1041,39 +1042,86 @@ export default class UserNodePlayer implements Player {
...
-    // the map of output topics -> inputs. Add these required input topics to the set of topic
-    // subscriptions to the underlying player.
-    for (const subscription of subscriptions) {
-      const inputs = state.inputsByOutputTopic.get(subscription.topic);
-      if (!inputs) {
-        nodeSubscriptions[subscription.topic] = subscription;
-        realTopicSubscriptions.push(subscription);
-        continue;
-      }
+    type InputMapping = [SubscribePayload, readonly string[] | undefined];
```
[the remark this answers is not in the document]
cfoust: I find this name harder to understand -- I'd prefer to keep the shorter one

packages/studio-base/src/players/UserNodePlayer/index.ts
cfoust: I also don't find this to be clearer 😓

packages/studio-base/src/players/UserNodePlayer/index.ts
cfoust: Personally I find consistency in function application to be easier to read than the `Array` properties. In other words, the function always comes first (e.g. `R.map`) and is followed by its arguments, rather than `[arg0] [function] [arg1]` e.g. `someArray.map(someFunction)`.

packages/studio-base/src/players/UserNodePlayer/index.ts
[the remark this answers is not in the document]
snosenzo: I think we should keep `input` reserved for referring to input topics to user nodes. I'd really prefer to use a different word

packages/studio-base/src/players/UserNodePlayer/index.ts
snosenzo: I find `inputs` to be only confusing and non descriptive😅

snosenzo (commented):
couple more clarity comments

PR #6775, opened 2023-09-08
fix: minor bug

pezy:
**User-Facing Changes**

**Description**

After reading the code, it is obvious that it should be the judgment of y.


jtbandes:
Could you add a unit test for this?

cfoust (approved):

PR #6780, opened 2023-09-08
Don't subscribe to block data unless we actually need it for plotting

cfoust:
**User-Facing Changes**
None.

**Description**
One of the challenges of plotting large `.mcap` datasets is that we currently subscribe to both block and current data, even if the plot only needs current data. This uses a lot of unnecessary resources. This PR makes it so that when a plot is either bounded (has a limited time range) or plots a single message, we only subscribe to current data.

### DIFF
--- a/packages/studio-base/src/panels/Plot/useDatasets.ts
+++ b/packages/studio-base/src/panels/Plot/useDatasets.ts
@@ -136,12 +136,21 @@ function chooseClient() {
       const isOnlyCurrent = isBounded(params) || isSingleMessage(params);
       return R.pipe(
         getPayloadsFromPaths,
-        R.map(
-          (v): SubscribePayload => ({
+        R.chain((v): SubscribePayload[] => {
+          const partial: SubscribePayload = {
             ...v,
-            preloadType: isOnlyCurrent ? "partial" : "full",
-          }),
-        ),
+            preloadType: "partial",
+          };
+
+          if (isOnlyCurrent) {
+            return [partial];
+          }
+
+          // Subscribe to both "partial" and "full" when using "full" In
+          // theory, "full" should imply "partial" but not doing this breaks
+          // MockMessagePipelineProvider
+          return [partial, { ...partial, preloadType: "full" }];
+        }),
       )(getPaths(yAxisPaths, xAxisPath));
     }),
     (v) => mergeSubscriptions(v) as SubscribePayload[],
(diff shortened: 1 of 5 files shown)


jtbandes:
Based on the description I've tried creating a plot panel with "x axis > value" set to "index", or with "x axis > range" set to something other than auto, however I can't really observe a difference in the "buffered" status of the playback bar (when using a local mcap file). Is there anything I can do to observe the behavior this PR changes other than checking memory usage?

We also need to start adding unit tests for this code to prove that the changes are having the intended effect and the behavior doesn't regress in the future.

On another note, I'm not sure if this was intentional but it seems like the behavior of the "Range (seconds)" changed since #6584. Previously, the whole plot would still be shown but the x-axis would be shortened to the number of seconds specified ([example deploy](https://studio-ic0wticfq.foxglove.party/) from before that PR merged), and when seeking you'd simply be looking at a different subset of the data (but could pan around):


https://github.com/foxglove/studio/assets/14237/1e9180c9-aa42-4aca-b565-7ee19f31aae5


Now (on studio.foxglove.dev and still in this PR), the data clears out whenever you seek:

https://github.com/foxglove/studio/assets/14237/ef7943d9-6339-4e15-9d62-718b3e14346e

I suspect it may have been intentional to keep memory usage down, but it might be worth thinking about whether there's a way to have the best of both worlds (i.e. if it can show all the data unless it would exceed some memory limit).

cfoust:
> Is there anything I can do to observe the behavior this PR changes other than checking memory usage?
Nope, that's it.

> We also need to start adding unit tests for this code [...]
I agree--indeed, that was a strong motivation for breaking this up into little, testable pieces. We can test the whole state machine in a vacuum.

> the behavior of the "Range (seconds)" changed since #6584
Yup, that's an intentional design decision. It's probably worth discussing this in detail on Slack--this is admittedly a tricky balance to strike.

packages/studio-base/src/panels/Plot/useDatasets.ts
```diff
@@ -133,7 +116,7 @@ function getPayloadsFromPaths(paths: readonly string[]): SubscribePayload[] {
       return [payload];
     }),
     // Then simplify
-    normalizePaths,
+    (v: SubscribePayload[]) => mergeSubscriptions(v) as SubscribePayload[],
```
jtbandes, line 119: why's the cast needed here?

jtbandes (approved):

PR #6785, opened 2023-09-09
Fix protobufDefinitionsToDatatypes for Timestamp/Duration

### LINKED ISSUE #6783: FrameTransform object incompatible between Protobuf and Typescript
I'm trying to write a user script in Foxglove Studio that saves `FrameTransform` messages into global scope. The source messages are of type https://github.com/foxglove/schemas/blob/main/schemas/proto/foxglove/FrameTransform.proto. However, when I try to make the assignment, I get the following error: 

```
Type '{ timestamp: { sec: bigint; nsec: number; }; parent_frame_id: string; child_frame_id: string; translation: { x: number; y: number; z: number; }; rotation: { x: number; y: number; z: number; w: number; }; }' is not assignable to type 'FrameTransform'.
  The types of 'timestamp.sec' are incompatible between these types.
    Type 'bigint' is not assignable to type 'number'.
```

- Version: 1.67.0
- Platform: MacOS

**Steps To Reproduce**
Try the following in a Foxglove Studio user script:

```
import {Input} from "./types";
import {FrameTransform} from "@foxglove/schemas";

export const inputs = ["/tf_dynamic"];

let last_tf: FrameTransform | null = null;

export default function script(event: Input<"/tf_dynamic">): void {
  last_tf = event.message;
}
```

Where `/tf_dynamic` is a topic that contains `FrameTransform` protos.

**Expected Behavior**
I should be able to manipulate `FrameTransform` objects in Typescript even if they originated from a protobuf message.

jtbandes:
**User-Facing Changes**
Fixed an incompatibility between timestamp/duration types in Protobuf messages and User Scripts.

**Description**
Fixes #6783 
Fixes FG-4764

Background:

Protobuf's standard Timestamp and Duration types use int64 `seconds` + int32 `nanos`, but most of Studio works with timestamps in `{sec,nsec}` format with `number` values.

Prior work:

- In #3443 I added code to modify these definitions to use `sec`/`nsec` instead for compatibility with the rest of Studio — however, I neglected to change the types in User Scripts.

- In #5727 I added `bigint` support to our Protobuf deserialization, which accidentally changed these timestamps to use `bigint` for `sec`, which I fixed in #5757.

- In #6249 I changed the User Script generated types to use the `sec`/`nsec` field names, but I didn't realize that they still used a `bigint` type for the `sec` field.

This PR finally fixes the User Script types so that they have `sec: number` instead of `sec: bigint`.

### DIFF
--- a/packages/mcap-support/src/protobufDefinitionsToDatatypes.ts
+++ b/packages/mcap-support/src/protobufDefinitionsToDatatypes.ts
@@ -74,6 +74,15 @@ export function protobufDefinitionsToDatatypes(
         throw new Error("Repeated bytes are not currently supported");
       }
       definitions.push({ type: "uint8", name: field.name, isArray: true });
+    } else if (
+      type.fullName === ".google.protobuf.Timestamp" ||
+      type.fullName === ".google.protobuf.Duration"
+    ) {
+      definitions.push({
+        type: "int32",
+        name: field.name === "seconds" ? "sec" : "nsec",
+        isArray: field.repeated,
+      });
     } else {
       definitions.push({
         type: protobufScalarToRosPrimitive(field.type),
--- a/packages/mcap-support/src/parseProtobufSchema.test.ts
+++ b/packages/mcap-support/src/parseProtobufSchema.test.ts
@@ -200,7 +200,7 @@ describe("parseProtobufSchema", () => {
           {
             "isArray": false,
             "name": "sec",
-            "type": "int64",
+            "type": "int32",
           },
           {
             "isArray": false,
@@ -216,7 +216,7 @@ describe("parseProtobufSchema", () => {
           {
             "isArray": false,
             "name": "sec",
-            "type": "int64",
+            "type": "int32",
           },
           {
             "isArray": false,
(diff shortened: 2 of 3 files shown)


packages/mcap-support/src/protobufDefinitionsToDatatypes.ts
```diff
@@ -74,6 +74,15 @@ export function protobufDefinitionsToDatatypes(
         throw new Error("Repeated bytes are not currently supported");
       }
       definitions.push({ type: "uint8", name: field.name, isArray: true });
+    } else if (
+      type.fullName === ".google.protobuf.Timestamp" ||
+      type.fullName === ".google.protobuf.Duration"
+    ) {
+      definitions.push({
+        type: "int32",
+        name: field.name === "seconds" ? "sec" : "nsec",
```
snosenzo, line 83: Is this operating on the not-yet-processed protobuf definitions? should it be looking for `sec` if not?

snosenzo (approved):

packages/mcap-support/src/protobufDefinitionsToDatatypes.ts
jtbandes: It's called after processing but I am no longer changing the names in the protobufjs type object (see deleted lines in parseProtobufSchema.ts)

PR #6786, opened 2023-09-09
Add multi-select and multi-drag support to topic list

jtbandes:
**User-Facing Changes**
Added support for selecting and dragging multiple topics and message paths from the Topics sidebar.

**Description**
Resolves FG-4849

Also fixes a `@container` query from #6653 that I broke in #6717 when moving styles between branches.

https://github.com/foxglove/studio/assets/14237/46130405-2164-4386-9601-a10eb4791cb2


packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
```diff
@@ -166,33 +166,39 @@ export function ThreeDeeRender(props: {
...
-        if (!path.isTopic || path.rootSchemaName == undefined) {
-          return { canDrop: false };
-        }
-        if (ALL_SUPPORTED_IMAGE_SCHEMAS.has(path.rootSchemaName)) {
-          return { canDrop: true, effect: "replace" };
-        }
-        if (ALL_SUPPORTED_ANNOTATION_SCHEMAS.has(path.rootSchemaName)) {
-          return { canDrop: true, effect: "add" };
+        let effect: "add" | "replace" = "add";
+        for (const path of paths) {
```
snosenzo, line 174: won't these override their effect based on the order of `paths`?
For example when you have an image topic and an annotation topic selected and being dragged, what's the expected result?

packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
```diff
@@ -166,33 +166,39 @@ export function ThreeDeeRender(props: {
...
+          if (!path.isTopic || path.rootSchemaName == undefined) {
+            return { canDrop: false };
+          }
+          if (ALL_SUPPORTED_IMAGE_SCHEMAS.has(path.rootSchemaName)) {
```
snosenzo, line 178: also shouldn't this only `replace` if it's an image panel?

packages/studio-base/src/components/TopicList/TopicList.tsx
```diff
@@ -89,12 +46,77 @@ export function TopicList(): JSX.Element {
...
 
   useEffect(() => {
     // Discard cached row heights when the filter results change
     listRef.current?.resetAfterIndex(0);
   }, [treeItems]);
 
+  const itemData = useMemo(() => ({ treeItems, selectedIndexes }), [selectedIndexes, treeItems]);
+
+  const renderRow: React.FC<ListChildComponentProps<typeof itemData>> = useCallback(
+    ({ index, style, data }) => {
```
snosenzo, line 85: Where are these params being supplied where this is used below? are they being injected by  `VariableSizeList`? should you comment that here?

packages/studio-base/src/services/messagePathDragging/index.ts
```diff
@@ -103,7 +108,35 @@ export function useMessagePathDrag({
     }
   }
 
-  return { connectDragSource, connectDragPreview, cursor, isDragging };
+  // To allow the draggedItemCount to be shown in the drag preview, we have to breifly render with
+  // the actual count, then re-render with the count at 0 to avoid showing the item count in the
+  // page itself.
```
snosenzo, line 113: Is this just a quirk of the library we're using? Would we ever be able to remove this hack in the future? if so, when?

packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
jtbandes: Yes, the regular 3D panel doesn't support drag/drop yet, this is the first conditional inside getDropStatus

packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
jtbandes: The logic I went with for now was "if there is any image schema, use `replace`, otherwise use `add`"

packages/studio-base/src/services/messagePathDragging/index.ts
jtbandes: I don't think we would be able to remove it unless we create a totally custom drag preview image that does not come from the element itself. If the element is used as a drag preview, the timing has to be such that the badge is rendered when the browser takes a snapshot, then it is un-rendered so you don't see the badge still on the original element while you drag the preview around.

packages/studio-base/src/components/TopicList/TopicList.tsx
jtbandes: `data` comes from the `itemData` we pass to the list... added comment

snosenzo (approved):

PR #6800, opened 2023-09-12
Fix backquote overlay shortcut

foxymiles:
**User-Facing Changes**
Fix backquote overlay shortcut with key combinations.

**Description**
When the meta option key is pressed and the backquote key is released we receive a `key` of `Dead` on a Mac. We can get around this by looking at the `code` instead.

Screenshot 2023-09-12 at 7 00 34 AM

I tried but wasn't able to get a story for this working. Because we use css `:hover` classes to control the visibility of the overlay the `userEvent.hover` function from the testing library doesn't seem to work.

FG-4825

### DIFF
--- a/packages/studio-base/src/components/KeyListener.tsx
+++ b/packages/studio-base/src/components/KeyListener.tsx
@@ -29,10 +29,12 @@ function callHandlers(handlers: KeyHandlers | undefined, event: KeyboardEvent): 
     return;
   }
 
-  if (typeof handlers[event.key] === "function") {
+  const handler = handlers[event.key] ?? handlers[event.code];
+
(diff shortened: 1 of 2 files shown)


jtbandes (approved):
This definitely seems to be an improvement – I verified that the "press `` ` ``, press option, release `` ` ``, release option" flow works now. However if I use command instead of option, the same thing still happens.

foxymiles:
> This definitely seems to be an improvement – I verified that the "press `` ` ``, press option, release `` ` ``, release option" flow works now. However if I use command instead of option, the same thing still happens.

Interesting. Looks like in this case at least on a Mac we never get the keyup from the backquote if the command key is down.

PR #6801, opened 2023-09-12
IterablePlayer: Fix exception from source not being handled

achim-k:
**User-Facing Changes**
None

**Description**
Fixes exceptions stemming from the iterable source (comlink proxies exceptions from the worker) not being handled.

Resolves FG-4762

### DIFF
--- a/packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
+++ b/packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
@@ -1067,20 +1067,24 @@ export class IterablePlayer implements Player {
   }
 
   async #startBlockLoading() {
-    await this.#blockLoader?.startLoading({
-      progress: async (progress) => {
-        this.#progress = {
-          fullyLoadedFractionRanges: this.#progress.fullyLoadedFractionRanges,
-          messageCache: progress.messageCache,
-        };
+    try {
+      await this.#blockLoader?.startLoading({
+        progress: async (progress) => {
+          this.#progress = {
+            fullyLoadedFractionRanges: this.#progress.fullyLoadedFractionRanges,
+            messageCache: progress.messageCache,
+          };
 
-        // If we are in playback, we will let playback queue state updates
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
```diff
@@ -1067,20 +1067,24 @@ export class IterablePlayer implements Player {
...
+          }
 
-        this.#queueEmitState();
-      },
-    });
+          this.#queueEmitState();
+        },
+      });
+    } catch (err) {
+      this.#setError((err as Error).message, err);
```
jtbandes, line 1087: I wonder if this should be put outside this function, in the caller of `#startBlockLoading` instead? I guess it's a matter of style/opinion. I think I would slightly lean towards putting it [here](https://github.com/foxglove/studio/blob/1b53ddcea1f890b19682a447ac439e82f13a2fc0/packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts#L550) instead.

jtbandes (approved):

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
```diff
@@ -546,8 +546,12 @@ export class IterablePlayer implements Player {
 
       this.#blockLoader?.setTopics(this.#preloadTopics);
 
-      // Block loadings is constantly running and tries to keep the preloaded messages in memory
-      this.#blockLoadingProcess = this.#startBlockLoading();
+      try {
+        // Block loadings is constantly running and tries to keep the preloaded messages in memory
+        this.#blockLoadingProcess = this.#startBlockLoading();
```
jtbandes, line 551: I think this would need to be a `.catch` instead of a try/catch

achim-k: Oops, you are absolutely right. Good `catch ` ;)

jtbandes (approved):

PR #6802, opened 2023-09-12
Add lint rule to replace R.min, R.max, & R.map with builtins

jtbandes:
**User-Facing Changes**
None

**Description**
Relates to FG-4805
Adds a couple lint rules to prefer usage of JS builtins over Ramda functions:
- Replace `R.min`/`R.max` with `Math.min`/`Math.max` when they are used with 2 arguments
- Replace `R.map(fn, array)` with `array.map(fn)`

### DIFF
--- a/packages/eslint-plugin-studio/ramda-usage.test.ts
+++ b/packages/eslint-plugin-studio/ramda-usage.test.ts
@@ -1,0 +1,73 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { RuleTester } from "@typescript-eslint/rule-tester";
+import { TSESLint } from "@typescript-eslint/utils";
+import path from "path";
+
+// eslint-disable-next-line @typescript-eslint/no-var-requires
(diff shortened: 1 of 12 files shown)


packages/eslint-plugin-studio/ramda-usage.test.ts
```diff
@@ -0,0 +1,73 @@
...
+        import * as R from "ramda";
+        Math.min(1, 2);
+        Math.max(1, 2);
+      `,
+    },
+
+    {
+      code: /* ts */ `
+        import * as R from "ramda";
+        R.map((x) => x + 1, [1, 2]);
```
snosenzo, line 53: should we have a test for a curried version of this function or does the rule not try to change those?

snosenzo (approved):

packages/eslint-plugin-studio/ramda-usage.test.ts
cfoust: +1 to a test for a curried version 👀

jtbandes: That would be here: https://github.com/foxglove/studio/pull/6802/files/1b53ddcea1f890b19682a447ac439e82f13a2fc0#diff-6bcc63f4e6b7f47fe23284d73bcc093301a5b47b48f6c27bd433e6f1545fc4f4R27

cfoust (approved):

PR #6803, opened 2023-09-12
Handle top-level topic filters

cfoust:
**User-Facing Changes**
None.

**Description**
Plots paths that use top-level filters (e.g. `/diagnostics{header.frame_id==""}`) are broken right now, which is causing issues for at least a couple of users. This PR fixes it and adds tests for paths like these to ensure we're always subscribing to them appropriately.

### DIFF
--- a/packages/studio-base/src/panels/Plot/useDatasets.ts
+++ b/packages/studio-base/src/panels/Plot/useDatasets.ts
@@ -86,8 +86,8 @@ export function pathToPayload(path: RosPath): SubscribePayload | undefined {
 
   // We want to take _all_ of the filters that start the path, since these can
   // be chained
-  const [filters, rest] = R.partition((part: MessagePathPart) => part.type === "filter", parts);
-  const firstField = rest[0];
+  const filters = R.takeWhile((part: MessagePathPart) => part.type === "filter", parts);
+  const firstField = R.find((part: MessagePathPart) => part.type === "name", parts);
   if (firstField == undefined || firstField.type !== "name") {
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/panels/Plot/useDatasets.ts
```diff
@@ -73,30 +76,58 @@ function normalizePaths(topics: SubscribePayload[]): SubscribePayload[] {
   )(topics);
 }
 
+/**
+ * Get the SubscribePayload for a single path by subscribing to all fields
+ * referenced in leading MessagePathFilters and the first field of the
+ * message.
+ */
+export function getPayload(path: RosPath): SubscribePayload | undefined {
```
jtbandes, line 84: naming: maybe getSubscribePayload would be clearer?

packages/studio-base/src/panels/Plot/useDatasets.ts
```diff
@@ -73,30 +76,58 @@ function normalizePaths(topics: SubscribePayload[]): SubscribePayload[] {
...
+        const { path: filterPath } = part;
+        const field = filterPath[0];
+        if (field == undefined) {
+          return [];
+        }
+
+        return [field];
+      }),
+      // Always subscribe to the header field
+      (v) => [...v, firstField.name, "header"],
```
jtbandes, line 111: Is there a more descriptive name that would work instead of `v`, like `filterField`?

jtbandes (approved):

PR #6804, opened 2023-09-12
Fixes locally installed message converters not being preferred over org provided ones

### LINKED ISSUE #6787: Local installation of extensions overwritten by organisation extension
**Description**

- Version: 1.69.0
- Platform: Ubuntu

**Steps To Reproduce**
1. `foxglove-extension install` a plugin being developed locally
2. Refresh foxglove, the new plugin should be working :green_circle: 
3. `foxglove extensions publish` the package to an org
4. Refresh foxglove, the published plugin will now be working :green_circle: 
5. Make changes to the local plugin (eg a message converter) and `foxglove-extension install`. The local changes to the message converter will not be used, the published plugin converter will be used. :red_circle:

**Expected Behavior**
I had expected the locally installed plugin to take precedence over the published plugin. I am not sure if this effects panels also. I suspect the issue lies with:
https://github.com/foxglove/studio/blob/2d7bb15b8571e5ee531d103a6e4568dae5c56559/packages/studio-base/src/components/PanelExtensionAdapter/messageProcessing.ts#L140 

The extensions could be ordered by locally installed first to resolve this?

achim-k:
**User-Facing Changes**
Fixes locally installed message converters not being preferred over org provided ones

**Description**
Rather than just changing the [extension loader ordering](https://github.com/foxglove/studio/blob/2d7bb15b8571e5ee531d103a6e4568dae5c56559/packages/studio-desktop/src/renderer/Root.tsx#L67-L70), this patch is slightly more explicit by selecting a message converter based on the extension namespace. Message converters with `local` namespace are preferred over `org` provided ones.

Fixes #6787
Resolves FG-4948

### DIFF
--- a/packages/studio-base/src/components/PanelExtensionAdapter/messageProcessing.ts
+++ b/packages/studio-base/src/components/PanelExtensionAdapter/messageProcessing.ts
@@ -11,12 +11,15 @@ import {
   Subscription,
 } from "@foxglove/studio";
 import { Topic as PlayerTopic } from "@foxglove/studio-base/players/types";
+import { ExtensionNamespace } from "@foxglove/studio-base/types/Extensions";
 
 // Branded string to ensure that users go through the `converterKey` function to compute a lookup key
 type Brand<K, T> = K & { __brand: T };
 type ConverterKey = Brand<string, "ConverterKey">;
 
-type MessageConverter = RegisterMessageConverterArgs<unknown>;
+type MessageConverter = RegisterMessageConverterArgs<unknown> & {
+  extensionNamespace?: ExtensionNamespace;
+};
 
 type TopicSchemaConverterMap = Map<ConverterKey, MessageConverter[]>;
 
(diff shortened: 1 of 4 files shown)


packages/studio-base/src/components/PanelExtensionAdapter/messageProcessing.ts
```diff
@@ -136,12 +136,16 @@ export function collateTopicSchemaConversions(
...
-    );
+    //       find one converter rather than multiple converters. Local message converters are
+    //       preferred over org provided converters.
+    const converters = (messageConverters ?? [])
+      .filter(
+        (conv) =>
+          conv.fromSchemaName === subscriberTopic.schemaName &&
+          conv.toSchemaName === subscription.convertTo,
+      )
+      .sort((conv) => (conv.extensionNamespace === "local" ? -1 : 0));
```
jtbandes, line 147: I think this may violate some requirements of the comparison function if `compare(a,b)` and `compare(b,a)` are not opposites – if you want to use only a 1-argument version then it's better to use a `sortBy`.

image

packages/studio/src/index.ts
```diff
@@ -454,6 +454,7 @@ export type RegisterMessageConverterArgs<Src> = {
   fromSchemaName: string;
   toSchemaName: string;
   converter: (msg: Src, event: Immutable<MessageEvent<Src>>) => unknown;
+  extensionNamespace?: string;
```
jtbandes, line 457: I'm not sure if it makes sense to add this to the public types because then users might think they can pass in some meaningful value here, when it seems like we are trying to keep the value of this property internally controlled.

packages/studio-base/src/components/PanelExtensionAdapter/messageProcessing.ts
```diff
@@ -135,13 +138,14 @@ export function collateTopicSchemaConversions(
...
-    //       find one converter rather than multiple converters.
-    const converter = messageConverters?.find(
+    let converters = (messageConverters ?? []).filter(
       (conv) =>
         conv.fromSchemaName === subscriberTopic.schemaName &&
         conv.toSchemaName === subscription.convertTo,
     );
+    // Sort by extension namespace to prefer 'local' converters over 'org' provided ones
+    converters = _.sortBy(converters, (conv) => conv.extensionNamespace ?? "unknown");
+    const converter = converters[0];
```
foxymiles, line 148: I think the intent might be a bit clearer here with lodash `minBy`.

foxymiles (approved):

packages/studio-base/src/components/PanelExtensionAdapter/messageProcessing.ts
achim-k: Makes it a lot clearer indeed. Thanks for the suggestion!

PR #6805, opened 2023-09-12
Fix extra lines rendering for LinePrimitives

snosenzo:
**User-Facing Changes**

- Fix occasional extra lines in SceneUpdate line primitives

**Description**

Fixed instanceCount for `LINE_LOOP` primitive types. Because we add one to the `numVertices` when we have a line loop, the  number of lines in a line loop should be `numVertices - 1`, or the original number of points on the primitive. The story I added has one square defined with 4 points (not closed) and the other defined with 5 points (already closed). We then add one in calculating `numVertices` because we close the loop in `serializePositions` even if the loop is already closed. This makes the number of lines (`instanceCount`) equal to 4 for the first and 5 for the second, whereas previously it was 5 and 6. This would cause the renderable to read additional points from existing, larger `positionBuffers` that were used after an update.

Fixes: FG-4944

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableLines.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableLines.ts
@@ -161,8 +161,6 @@ class LinePrimitiveRenderable extends THREE.Object3D {
         this.#geometry == undefined ||
         this.#lineType !== this.#primitive.type ||
         !this.#positionBuffer ||
-        // Because `LineGeometry.setPosition` iterates through the positionBuffer to create an interleaved buffer of positions,
-        // we can't simply reuse the float32array with a larger length, we need to recreate it or we will get segments beyond the new length
         this.#positionBuffer.length < necessaryPositionBufferSize;
       if (geometryNeedsRecreated) {
         if (this.#geometry != undefined) {
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableLines.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableLines.ts
@@ -163,7 +163,7 @@ class LinePrimitiveRenderable extends THREE.Object3D {
         !this.#positionBuffer ||
         // Because `LineGeometry.setPosition` iterates through the positionBuffer to create an interleaved buffer of positions,
         // we can't simply reuse the float32array with a larger length, we need to recreate it or we will get segments beyond the new length
-        this.#positionBuffer.length !== necessaryPositionBufferSize;
+        this.#positionBuffer.length < necessaryPositionBufferSize;
       if (geometryNeedsRecreated) {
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/panels/ThreeDeeRender/stories/SceneEntities.stories.tsx
```diff
@@ -436,3 +438,425 @@ export const BasicEntities: StoryObj = {
...
+];
+
+const frame1: MessageEvent<Partial<SceneUpdate>> = {
+  topic: "scene",
+  schemaName: "foxglove.SceneUpdate",
+  sizeInBytes: 0,
+  receiveTime: { sec: 10, nsec: 0 },
+  message: lineLoopSampleData[0]!,
+};
+const frame2: MessageEvent<Partial<SceneUpdate>> = {
```
jtbandes, line 774: Maybe a comment somewhere as to what this is testing?

packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableLines.ts
```diff
@@ -161,7 +161,9 @@ class LinePrimitiveRenderable extends THREE.Object3D {
         this.#geometry == undefined ||
         this.#lineType !== this.#primitive.type ||
         !this.#positionBuffer ||
-        this.#positionBuffer.length < necessaryPositionBufferSize;
+        // Because `LineGeometry.setPosition` iterates through the positionBuffer to create an interleaved buffer of positions,
+        // we can't simply reuse the float32array with a larger length, we need to recreate it or we will get segments beyond the new length
+        this.#positionBuffer.length !== necessaryPositionBufferSize;
```
jtbandes, line 166: hm... isn't the custom value of `this.#geometry.instanceCount` (below) supposed to take care of this?

snosenzo: Yeah I address that in the PR description, it doesn't look like it's working for LineGeometry, but that would probably be worth mentioning here too

snosenzo: InstanceCount was working, we were calculating it wrong. I updated the code and PR description to reflect that

packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableLines.ts
```diff
@@ -161,6 +161,8 @@ class LinePrimitiveRenderable extends THREE.Object3D {
         this.#geometry == undefined ||
         this.#lineType !== this.#primitive.type ||
         !this.#positionBuffer ||
+        // Because `LineGeometry.setPosition` iterates through the positionBuffer to create an interleaved buffer of positions,
```
jtbandes, line 164: Stale comment?

jtbandes (approved):

PR #6812, opened 2023-09-13
UserScript Tabs are actually tabs

2metres:
**User-Facing Changes**
Make the UserScripts Tabs a reality

**Description**


amacneil:
LR feedback:
- Just kill the tab bar entirely
- Find a new place to put the "add script" button

PR #6814, opened 2023-09-13
Handle updated block data in plots

cfoust:
**User-Facing Changes**
None.

**Description**
The plot data pipeline was not detecting changes to block data and clearing out the plot. This can occur when the user makes a change to a user script, the output of which they're rendering in a plot.

### DIFF
--- a/packages/studio-base/src/panels/Plot/useDatasets.ts
+++ b/packages/studio-base/src/panels/Plot/useDatasets.ts
@@ -44,8 +44,11 @@ async function waitService(): Promise<Service> {
 
 const getIsLive = (ctx: MessagePipelineContext) => ctx.seekPlayback == undefined;
 
-// topic -> number of fields
-type BlockStatus = Record<string, number>;
+// We need to keep track of the block data we've already sent to the worker and
+// detect when it has changed, which can happen when the user changes a user
+// script or they trigger a subscription to different fields.
+// mapping from topic -> the first message on that topic in the block
+type BlockStatus = Record<string, unknown>;
 let blockStatus: BlockStatus[] = [];
 
 const getPayloadString = (payload: SubscribePayload): string =>
@@ -152,15 +155,6 @@ function chooseClient() {
   blockStatus = R.map((block) => R.pick(R.map(getPayloadString, topics), block), blockStatus);
 }
 
-function getNumFields(events: readonly MessageEvent[]): number {
-  const message = events[0]?.message;
-  if (message == undefined) {
-    return 0;
-  }
-
-  return Object.keys(message).length;
-}
-
 // Subscribe to "current" messages (those near the seek head) and forward new
 // messages to the worker as they arrive.
 function useData(id: string, topics: SubscribePayload[]) {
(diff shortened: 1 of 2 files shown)


cfoust:
The Storybook changes appear to be the result of previous `readySignal` counts that were inaccurate. I verified the changes locally and they both match CI and seem correct.

packages/studio-base/src/panels/Plot/useDatasets.ts
```diff
@@ -210,13 +204,18 @@ function useData(id: string, topics: SubscribePayload[]) {
     React.useMemo(() => R.map((v) => ({ ...v, preloadType: "full" }), subscribed), [subscribed]),
   );
   useEffect(() => {
+    const wasReset: Record<string, boolean> = {};
```
jtbandes, line 207: Why a `Record<string,boolean>` rather than a `Set<string>`? A Set would also let you remove the `R.keys` below

packages/studio-base/src/panels/Plot/useDatasets.ts
```diff
@@ -44,8 +44,11 @@ async function waitService(): Promise<Service> {
 
 const getIsLive = (ctx: MessagePipelineContext) => ctx.seekPlayback == undefined;
 
-// topic -> number of fields
-type BlockStatus = Record<string, number>;
+// We need to keep track of the block data we've already sent to the worker and
+// detect when it has changed, which can happen when the user changes a user
+// script or they trigger a subscription to different fields.
+// mapping from topic -> the first message on that topic in the block
+type BlockStatus = Record<string, unknown>;
```
jtbandes, line 51: Why do we track the message itself here and not the MessageEvent? The `unknown` makes it seem like just about anything could be in here – it might be more readable if it were `Record<string, MessageEvent<unknown>>`.

packages/studio-base/src/panels/Plot/useDatasets.ts
```diff
@@ -225,19 +224,29 @@ function useData(id: string, topics: SubscribePayload[]) {
           continue;
         }
 
-        const numFields = getNumFields(topicMessages);
-        if (status[ref] === numFields) {
+        const first = topicMessages[0]?.message;
+        const existing = status[ref];
+        if (R.equals(existing, first)) {
```
jtbandes, line 229: I'm not sure if we need deep equality here – I think `===` would be sufficient to check if the message is the same? (In fact, I think there could even truly be two deep-equal messages at different timestamps (or even the same timestamp) in the data source)

packages/studio-base/src/panels/Plot/useDatasets.worker.ts
```diff
@@ -391,20 +391,31 @@ function evictCache() {
...
 }
 
-function addBlock(block: Messages): void {
+function addBlock(block: Messages, resetTopics: string[]): void {
   const topics = R.keys(block);
-  blocks = R.mergeWith(R.concat, blocks, block);
+
+  blocks = R.pipe(
+    // Remove data for any topics that have been reset
+    R.pick(R.difference(R.keys(blocks), resetTopics)),
```
jtbandes, line 399: I think this would be equivalent to just `R.omit(resetTopics)` – or am I missing something?

packages/studio-base/src/panels/Plot/useDatasets.ts
cfoust: For reasons I could not discern, we update properties on the `MessageEvent` after the fact (namely, `sizeInBytes`). We don't want that to trigger a full resend of the data

packages/studio-base/src/panels/Plot/useDatasets.ts
cfoust: I tried this--something, somewhere was updating the message references.

packages/studio-base/src/panels/Plot/useDatasets.ts
cfoust: I tend not to use `Set`; it's really just personal preference I suppose

jtbandes (approved):
Re: the Set thing, now I think it looks strange to have `wasReset` being a `Record` and `resetData` being a `Set` – maybe both can be Sets?

PR #6816, opened 2023-09-13
Tests for #nodeSubscriptions and subscriptions in UserNodePlayer

cfoust:
**User-Facing Changes**
None.

**Description**
I promised to write a comprehensive set of tests for this complicated transformation with a lot of edge cases, so here we are.

### DIFF
--- a/packages/studio-base/src/players/UserNodePlayer/index.ts
+++ b/packages/studio-base/src/players/UserNodePlayer/index.ts
@@ -58,7 +58,7 @@ import { UserNode, UserNodes } from "@foxglove/studio-base/types/panels";
 import Rpc from "@foxglove/studio-base/util/Rpc";
 import { basicDatatypes } from "@foxglove/studio-base/util/basicDatatypes";
 
-import { simplifySubscriptions } from "./subscriptions";
+import { remapVirtualSubscriptions, getPreloadTypes } from "./subscriptions";
 
 const log = Log.getLogger(__filename);
 
@@ -1042,13 +1042,10 @@ export default class UserNodePlayer implements Player {
   }
 
   #setSubscriptionsUnlocked(subscriptions: SubscribePayload[], state: ProtectedState): void {
-    const [nodeSubscriptions, resolvedSubscriptions] = simplifySubscriptions(
-      subscriptions,
-      state.inputsByOutputTopic,
+    this.#nodeSubscriptions = getPreloadTypes(subscriptions);
+    this.#player.setSubscriptions(
+      remapVirtualSubscriptions(subscriptions, state.inputsByOutputTopic),
     );
-
-    this.#nodeSubscriptions = nodeSubscriptions;
-    this.#player.setSubscriptions(resolvedSubscriptions);
   }
 
   public close = (): void => {
--- a/packages/studio-base/src/players/UserNodePlayer/index.ts
+++ b/packages/studio-base/src/players/UserNodePlayer/index.ts
@@ -14,7 +14,6 @@
 import { Mutex } from "async-mutex";
 import * as _ from "lodash-es";
 import memoizeWeak from "memoize-weak";
-import * as R from "ramda";
 import ReactDOM from "react-dom";
 import shallowequal from "shallowequal";
 import { v4 as uuidv4 } from "uuid";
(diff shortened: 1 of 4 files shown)


packages/studio-base/src/players/UserNodePlayer/subscriptions.test.ts
```diff
@@ -0,0 +1,157 @@
...
+        },
+        {
+          topic: "/test2",
+          preloadType: "partial",
+        },
+      ],
+    ]);
+  });
+
+  it("upgrades to a full subscription from fields", () => {
```
jtbandes, line 92: Not sure I fully understand this scenario – why doesn't the output have the `fields` in it anywhere?

cfoust: any inputs to user scripts receive subscriptions to all fields. we also don't allow for subscribing to individual fields of a user node topic

packages/studio-base/src/players/UserNodePlayer/index.ts
```diff
@@ -1042,76 +1042,12 @@ export default class UserNodePlayer implements Player {
...
-              {
-                topic: subscription.topic,
-                preloadType,
-              },
-            ];
-          }
-
-          return [subscription];
-        }
+    const [nodeSubscriptions, resolvedSubscriptions] = simplifySubscriptions(
```
snosenzo, line 1045: I think this and the tests might be easier to read if this function was split into two
```
// this actually doesn't need inputsByOutputTopic
this.#nodeSubscriptions = simplifySubscriptions(subscriptions); 

this.#playerSubscriptions.setSubscriptions(
  resolveToRealSubs(subscriptions, state.inputsByOutputTopic)
);
```
I think this function name hides too much and it's hard to see what's being passed where. Splitting the two of these functions up will also make the tests a bit easier to read.

packages/studio-base/src/players/UserNodePlayer/subscriptions.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  const resolvedSubscriptions = R.pipe(
+    R.chain(([subscription, topics]: SubscriberInputs): SubscribePayload[] => {
+      const preloadType = subscription.preloadType ?? "partial";
+
+      // Leave the subscription unmodified if it is not a user script topic
+      if (topics == undefined) {
+        // If this is an input to a user script, we need to upgrade it to a
+        // subscription of all the fields
+        if (neededInputTopics.includes(subscription.topic)) {
```
snosenzo, line 70: actually coming back to this now -- we only need to upgrade this subscription to read all of the fields if the output of the user script is subscribed to

cfoust:
Bump on this!

packages/studio-base/src/players/UserNodePlayer/subscriptions.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  return R.pipe(
+    R.chain(([subscription, topics]: SubscriberInputs): SubscribePayload[] => {
+      const preloadType = subscription.preloadType ?? "partial";
+
+      // Leave the subscription unmodified if it is not a user script topic
+      if (topics == undefined) {
+        // If this is an input to a user script, we need to upgrade it to a
+        // subscription of all the fields
+        if (neededInputTopics.includes(subscription.topic)) {
```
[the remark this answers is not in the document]
cfoust: I might be misunderstanding what you're saying, but if a topic is an input to a user script we necessarily need a subscription to all of its fields because we can't know what fields the script uses. If that's not what you're referring to, could you clarify what you mean here?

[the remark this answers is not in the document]
snosenzo: Those input subscriptions to all fields will be produced below where you iterate over the inputs to an output topic and create whole-message subscriptions with the corresponding preloadType. Since those input topic subscriptions are already being produced, mergeSubscriptions should be able to resolve fields subs to whole-message subs as we need.

snosenzo (approved):
up to you whether you want to include the changes from my branch or not:
https://github.com/foxglove/studio/compare/caleb/fg-4952-tests-for-nodesubscriptions-and-subscriptions-in...no-coalesce-input-subs?expand=1

PR #6825, opened 2023-09-14
Fix crash in eslint-plugin-studio

amacneil:
**User-Facing Changes**
N/A

**Description**
Fix unhandled undefined in eslint-plugin-studio

```
Oops! Something went wrong! :(

ESLint: 8.49.0

TypeError: Cannot read properties of undefined (reading 'name')
Occurred while linting /packages/console/components/AppLink.tsx:96
Rule: "@foxglove/studio/link-target"
    at /studio/packages/eslint-plugin-studio/link-target.js:14:61
    at Array.some (<anonymous>)
```

### DIFF
--- a/packages/eslint-plugin-studio/link-target.js
+++ b/packages/eslint-plugin-studio/link-target.js
@@ -11,7 +11,7 @@ module.exports = {
   create: (context) => {
     return {
       [`JSXElement > JSXOpeningElement > JSXAttribute[name.name="href"]`]: (node) => {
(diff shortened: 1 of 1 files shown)


jtbandes (approved):
How is there an attribute without a name?

Do we need to run the studio lint rules on console?

jtbandes:
It’d be nice to add a test case for this if possible!

jtbandes:
OK, I learned that an attribute like `{...other}` is what triggers this.

PR #6828, opened 2023-09-14
fix TypeError when custom message header.stamp is not object

pezy:
**User-Facing Changes**

None

**Description**

When a custom message has a header.stamp but does not follow the ros style, it will cause errors such as `TypeError: Cannot use 'in' operator to search for 'sec' in 1694712977`.

Add necessary judgments and test cases.


jtbandes (approved):
Makes sense, thanks!

PR #6833, opened 2023-09-15
Prevent hoisting dependencies above workspace

amacneil:
**User-Facing Changes**
N/A

**Description**
Experimental\


jtbandes:
Doesn't seem like a viable approach, doing #6834 instead

PR #6846, opened 2023-09-18
Fix broken client behavior on plots

cfoust:
**User-Facing Changes**
None.

**Description**
Due to my own inaccurate understanding of `useEffect` (which I have used since its release and whose idiosyncrasies I have defended to the death--pride truly cometh before the fall) there was a bug that could cause repeated client registrations, which manifested in some odd plot behavior, such as in stories.

Fixes FG-5031

### DIFF
--- a/packages/studio-base/src/panels/Plot/useDatasets.ts
+++ b/packages/studio-base/src/panels/Plot/useDatasets.ts
@@ -37,7 +37,7 @@ type Service = Comlink.Remote<(typeof import("./useDatasets.worker"))["service"]
 // mapping from topic -> the first message on that topic in the block
 type BlockStatus = Record<string, unknown>;
 type Client = {
-  params: PlotParams;
+  params: PlotParams | undefined;
   setter: (topics: SubscribePayload[]) => void;
 };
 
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/panels/Plot/useDatasets.ts
```diff
@@ -184,6 +188,18 @@ function useData(id: string, params: PlotParams) {
...
+    const { [id]: client } = clients;
+    if (client == undefined) {
+      return;
+    }
+
+    clients = {
+      ...clients,
+      [id]: { ...client, params },
+    };
   }, [id, params]);
```
jtbandes, line 203: Did you consider potentially keeping a single useEffect and just rewriting it to handle both the case of initial registration and param updates? I think either way is fine, that just occurred to me as a possible refactoring of this.

I think it would help readability to add a comment above these useEffects to help understand their purpose – something like "register client when the panel mounts and unregister when it remounts" + "update registration when params change"

jtbandes (approved):

packages/studio-base/src/panels/Plot/useDatasets.ts
```diff
@@ -184,6 +188,18 @@ function useData(id: string, params: PlotParams) {
...
+  useEffect(() => {
+    const { [id]: client } = clients;
+    if (client == undefined) {
+      return;
+    }
+
+    clients = {
+      ...clients,
+      [id]: { ...client, params },
+    };
```
jtbandes, line 202:
```suggestion
    };
    chooseClient();
```

This seems to fix some of the newly-blank stories locally. But I'm not sure it is a totally correct change.

PR #6848, opened 2023-09-18
Renderer uses `SceneExtensionConfig`

snosenzo:
**User-Facing Changes**

- none

**Description**

In order to support adding or overriding scene extensions in the panel I have added a `sceneExtensionsConfigOverride` prop to the `ThreeDeePanel` component, which is merged with a default config and passed to the `Renderer`. The config contains `init`ializers that are callbacks which return the new `SceneExtension` instance, these are treated generically with exceptions for scene extensions that are assigned to render variables on the `Renderer` itself.

Extensions are placed in the config by their extensionId. The extensionId has been made a static member on the class so that it can be more easily subclassed and overriden by `sceneExtensionOverrides`.

`ImageMode` input props were reduced to make more consistent with other scene extensions. This also required `enableImageOnlyMode` (and disable) to be made a public so that `ImageMode` can utilize it directly. I know this is an additional public method, but because the initializers are outside of `Renderer` they no longer have access to these methods to begin with. 

FG-5027


packages/studio-base/src/panels/ThreeDeeRender/SceneExtensionConfig.ts
```diff
@@ -0,0 +1,122 @@
...
+  imageMode: ExtensionOverride<ImageMode>;
+  measurementTool: ExtensionOverride<MeasurementTool>;
+  publishClickTool: ExtensionOverride<PublishClickTool>;
+  //   cameraHandler: ExtensionOverride<ICameraHandler>;
+};
+
+export type ExtensionOverride<ExtensionType extends SceneExtension> = {
+  init: (renderer: IRenderer) => ExtensionType;
+  /** Which interfaceModes this extension is supported in. If undefined, will default to both '3d' and 'image' modes */
+  supportedModes?: InterfaceMode[];
```
jtbandes, line 46:
```suggestion
  supportedInterfaceModes?: InterfaceMode[];
```

packages/studio-base/src/panels/ThreeDeeRender/SceneExtensionConfig.ts
```diff
@@ -0,0 +1,122 @@
...
+      init: (renderer: IRenderer) => new Cameras(renderer),
+      supportedModes: ["3d"],
+    },
+    [PublishSettings.extensionId]: {
+      init: (renderer: IRenderer) => new PublishSettings(renderer),
+      supportedModes: ["3d"],
+    },
+    [FrameAxes.extensionId]: {
+      init: (renderer: IRenderer) =>
+        // only show frame axes labels when in 3d mode
```
jtbandes, line 79:
```suggestion
        // only show frame axes and labels by default when in 3d mode
```

packages/studio-base/src/panels/ThreeDeeRender/SceneExtensionConfig.ts
```diff
@@ -0,0 +1,122 @@
...
+  reserved: ReservedSceneExtensionConfig;
+  extensionsById: Record<AvailableNames, ExtensionOverride<SceneExtension>>;
+};
+/** Reserved because the Renderer has members that reference them specifically */
+
+export type ReservedSceneExtensionConfig = {
+  imageMode: ExtensionOverride<ImageMode>;
+  measurementTool: ExtensionOverride<MeasurementTool>;
+  publishClickTool: ExtensionOverride<PublishClickTool>;
+  //   cameraHandler: ExtensionOverride<ICameraHandler>;
```
jtbandes, line 40: delete?

packages/studio-base/src/panels/ThreeDeeRender/SceneExtensionConfig.ts
```diff
@@ -0,0 +1,122 @@
...
+import { SceneExtension } from "./SceneExtension";
+import { MeasurementTool } from "./renderables/MeasurementTool";
+import { PublishClickTool } from "./renderables/PublishClickTool";
+import { InterfaceMode } from "./types";
+
+export type SceneExtensionConfig = {
+  reserved: ReservedSceneExtensionConfig;
+  extensionsById: Record<AvailableNames, ExtensionOverride<SceneExtension>>;
+};
+/** Reserved because the Renderer has members that reference them specifically */
```
jtbandes, line 34: nit: seems to have gotten detached from its code

packages/studio-base/src/panels/ThreeDeeRender/SceneExtensionConfig.ts
```diff
@@ -0,0 +1,122 @@
...
+};
+
+export type ExtensionOverride<ExtensionType extends SceneExtension> = {
+  init: (renderer: IRenderer) => ExtensionType;
+  /** Which interfaceModes this extension is supported in. If undefined, will default to both '3d' and 'image' modes */
+  supportedModes?: InterfaceMode[];
+};
+
+type ReservedNames = keyof ReservedSceneExtensionConfig;
+type AvailableNames = Exclude<string, ReservedNames>;
```
jtbandes, line 50: You can't actually exclude things from `string` like that, I think Exclude only works for unions:
```suggestion
type AvailableNames = Exclude<string, ReservedNames>;
```

packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
```diff
@@ -147,7 +153,18 @@ export function ThreeDeeRender(props: {
   const rendererRef = useRef<IRenderer | undefined>(undefined);
   useEffect(() => {
     const newRenderer = canvas
-      ? new Renderer({ canvas, config: configRef.current, interfaceMode, fetchAsset, debugPicking })
+      ? new Renderer({
+          canvas,
+          config: configRef.current,
+          interfaceMode,
+          fetchAsset,
+          sceneExtensionConfig: _.merge(
```
jtbandes, line 161: Do we really want a recursive merge? That seems like it would allow someone to override the `init` but not the `supportedModes` or vice versa, which seems strange

packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
```diff
@@ -102,10 +106,12 @@ export function ThreeDeeRender(props: {
   interfaceMode: InterfaceMode;
   /** Override default downloading behavior, used for Storybook */
   onDownloadImage?: (blob: Blob, fileName: string) => void;
+  sceneExtensionConfigOverrides?: DeepPartial<SceneExtensionConfig>;
```
jtbandes, line 109: not convinced on the naming here, it seems like this could just be named `sceneExtensions` or something simpler like that

packages/studio-base/src/panels/ThreeDeeRender/renderables/FoxgloveGrid.ts
```diff
@@ -385,10 +385,11 @@ export class FoxgloveGridRenderable extends Renderable<FoxgloveGridUserData> {
 }
 
 export class FoxgloveGrid extends SceneExtension<FoxgloveGridRenderable> {
+  public static extensionId = "foxglove.Grid";
   #fieldsByTopic = new Map<string, string[]>();
 
-  public constructor(renderer: IRenderer) {
-    super("foxglove.Grid", renderer);
+  public constructor(renderer: IRenderer, name: string = FoxgloveGrid.extensionId) {
```
jtbandes, line 391: what are these names used for again? I wonder if we could get rid of them some day...seems to just add noise most of the time

packages/studio-base/src/panels/ThreeDeeRender/renderables/Markers.ts
```diff
@@ -35,8 +35,9 @@ const DEFAULT_SETTINGS: LayerSettingsMarker = {
 };
 
 export class Markers extends SceneExtension<TopicMarkers> {
-  public constructor(renderer: IRenderer) {
-    super("foxglove.Markers", renderer);
+  public static extensionId = "foxglove.Markers";
+  public constructor(renderer: IRenderer, name: string = Markers.extensionId) {
+    super(name, renderer);
   }
```
jtbandes, line 41: can we make the superclass constructor use `this.constructor.extensionId`/`new.target.extensionId` as the name by default so we don't have to do this in each extension?

packages/studio-base/src/panels/ThreeDeeRender/renderables/FoxgloveGrid.ts
snosenzo: they're used as a unique identifier for the SceneExtension in renderer.#addSceneExtension, which throws an error if more than one scene extension is added with the same `.name`.

packages/studio-base/src/panels/ThreeDeeRender/renderables/Markers.ts
snosenzo: So the reason why I did it this way is so that someone who's trying to subclass `Markers` could easily change the name if they wanted to but use the old class's static member to place in the `sceneExtensionConfig`.
For Example:
```ts

class MarkersExtra extends Markers {
  constructor(renderer: IRenderer) {
     super(renderer, "my.MarkersExtra");
   }
   ...
}

const sceneExtensionConfigOverride = {
  extensionsById: {
    [Markers.extensionId]: {
      init: (renderer) => {
        return new MarkersExtra(renderer);
      },
    },
  },
} ;

```
It feels natural since they'll be importing the class to begin with and wanting to override that scene extension in the config. Open to alternatives though if this feels clunky.

packages/studio-base/src/panels/ThreeDeeRender/SceneExtensionConfig.ts
snosenzo: Really had me going when it didn't error :(

packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
snosenzo: Yeah I think we do. As described below, someone should be able to override a default scene extension with their own instance.

packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
snosenzo: renamed to `customSceneExtensions` with comment saying that it is for overriding or injecting. thoughts?

jtbandes (approved):

PR #6852, opened 2023-09-18
Update depcheck and remove patch

jtbandes:
**User-Facing Changes**
None

**Description**
https://github.com/depcheck/depcheck/pull/834 was merged and released, so we don't need this patch anymore.


amacneil (approved):
hooray for fewer patches

PR #6863, opened 2023-09-19
Scene extensions can hook into drag and drop

snosenzo:
**User-Facing Changes**

- n/a

**Description**
Drag and drop capability now driven by scene extensions in 3D panel. This will allow custom scene extensions to be injected into the 3D panel and utilize this functionality.

FG-5025


packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -1356,6 +1358,47 @@ export class Renderer extends EventEmitter<RendererEvents> implements IRenderer
...
+      }
+      effects.push(effect);
+    }
+    // prioritize replace effect over add
+    const finalEffect = effects.reduce((acc, cur) => {
+      if (cur === "replace") {
+        return "replace";
+      }
+      return acc;
+    }, "add");
```
jtbandes, line 1385: might be easier to read as `effects.includes("replace") ? "replace" : "add"`

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -1356,6 +1358,47 @@ export class Renderer extends EventEmitter<RendererEvents> implements IRenderer
...
+      canDrop: true,
+      effect: finalEffect,
+    };
+  };
+
+  public handleDrop = (paths: readonly DraggedMessagePath[]): void => {
+    this.updateConfig((draft) => {
+      for (const path of paths) {
+        for (const extension of this.sceneExtensions.values()) {
+          extension.handleDropForPath(draft, path);
```
jtbandes, line 1397: I guess this assumes that 2 extensions won't handle the same path?

packages/studio-base/src/panels/ThreeDeeRender/SceneExtension.ts
```diff
@@ -158,6 +158,18 @@ export class SceneExtension<
     void backgroundColor;
   }
 
+  /** Returns a drop status if the Scene Extension can handle a message path drop, undefined if it cannot */
+  public getDropStatusForPath = (path: DraggedMessagePath): "add" | "replace" | undefined => {
```
jtbandes, line 162: I think this should either be renamed to say "effect" instead of "status", or it should return the MessagePathDropStatus type; it's confusing to use the "drop status" name but only return a part of the status (the effect).

packages/studio-base/src/panels/ThreeDeeRender/SceneExtension.ts
```diff
@@ -158,6 +158,18 @@ export class SceneExtension<
...
+    void path;
+    return undefined;
+  };
+
+  /** Called when a Message Path is dropped on the panel */
+  public handleDropForPath = (draft: Writable<RendererConfig>, path: DraggedMessagePath): void => {
```
jtbandes, line 168: Maybe name this `updateConfigForDroppedPath` or something? Otherwise when I see a name like "handleDrop" I'm not expecting there to be a `draft` passed in – alternatively the scene extension itself could be responsible for calling `updateConfig`.

jtbandes (approved):

packages/studio-base/src/panels/ThreeDeeRender/SceneExtension.ts
snosenzo: I wanted to avoid having to call `updateConfig` multiple times across scene extensions. Happy to update the name though

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
snosenzo: How does this assume that? I think multiple extensions can handle the same path and update their respective slices of the config

jtbandes: Ah yeah I just meant that if the drag&drop was meant for one of them but both handled it, you could end up adding the same topic twice or something like that. I guess it's not likely to cause a problem

jtbandes (approved):

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
snosenzo: ah yeah I mean ideally the scene extensions are operating on entirely different slices of state, but I guess that's not really documented anywhere.

PR #6865, opened 2023-09-19
Lift problems state into separate store

foxymiles:
**User-Facing Changes**
None

**Description**
Lift problems state into a separate store.

I'd prefer to keep this in the workspace store and not add yet one more context but components like the layout mananager that sit above the workspace need to be able to use this.

### DIFF
--- a/packages/studio-base/src/providers/ProblemsContextProvider.tsx
+++ b/packages/studio-base/src/providers/ProblemsContextProvider.tsx
@@ -1,0 +1,44 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { Immutable } from "immer";
+import { ReactNode, useState } from "react";
+import { StoreApi, create } from "zustand";
+
+import {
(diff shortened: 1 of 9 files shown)


packages/studio-base/src/providers/ProblemsContextProvider.tsx
```diff
@@ -0,0 +1,44 @@
...
+    return {
+      problems: [],
+      actions: {
+        clearProblem: (tag: string) => {
+          set({
+            problems: get().problems.filter((prob) => prob.tag !== tag),
+          });
+        },
+        setProblem: (tag: string, problem: Immutable<SessionProblem>) => {
+          const newProblems = get().problems.filter((prob) => prob.tag !== tag);
```
snosenzo, line 26: nit: isn't this technically the `oldProblems`? 😏

snosenzo (approved):

jtbandes (approved):
RIP workspace megastore dreams

packages/studio-base/src/providers/ProblemsContextProvider.tsx
foxymiles: Yeah I originally was doing an `unshift` on it but switched to a spread and didn't change the name.

PR #6872, opened 2023-09-20
Allow scene extensions to hook into context menu

snosenzo:
**User-Facing Changes**

- n/a

**Description**

Added `getContextMenuItems` to `SceneExtension` and `Renderer`. Moved the existing `DownloadImage` option into the `ImageMode` scene extension. 

Consolidated `debugPicking` and `onDownloadImage` into a `TestOptions` type that's passed to `ThreeDeeRender` and the `Renderer`.

FG-5026


packages/studio-base/src/panels/ThreeDeeRender/SceneExtension.ts
```diff
@@ -103,6 +105,14 @@ export class SceneExtension<
     this.updateSettingsTree();
   }
 
+  /** Allows SceneExtensions to add options to the context menu. Provides function to add pop up information via enqueueSnackbar */
+  public getContextMenuItems = (
+    enqueueSnackbar: EnqueueSnackbar,
```
jtbandes, line 110: I don't like having a dependence on EnqueueSnackbar here, what do you think of adding a method on Renderer like "displayTemporaryError" or something?

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -1337,6 +1343,11 @@ export class Renderer extends EventEmitter<RendererEvents> implements IRenderer
 
     this.settings.errors.remove(FOLLOW_TF_PATH, FOLLOW_FRAME_NOT_FOUND);
   }
+  public getContextMenuItems = (enqueueSnackbar: EnqueueSnackbar): PanelContextMenuItem[] => {
+    return _.flatMap(Array.from(this.sceneExtensions.values()), (extension) =>
```
jtbandes, line 1347: Array actually has a built-in flatMap these days. I feel another lint rule coming on

packages/studio-base/src/panels/ThreeDeeRender/RendererOverlay.tsx
```diff
@@ -297,83 +285,9 @@ export function RendererOverlay(props: {
...
-      {
-        type: "item",
-        label: "Download image",
-        onclick: doDownloadImage,
-        disabled: renderer?.getCurrentImage() == undefined,
-      },
-    ],
-    [doDownloadImage, renderer],
-  );
+  const getContextMenuItems = useCallback((): PanelContextMenuItem[] => {
```
foxymiles, line 288: Is there any other context an extension might need to generate its menu or is access to the renderer enough?

foxymiles (approved):

packages/studio-base/src/panels/ThreeDeeRender/RendererOverlay.tsx
snosenzo: Access to the renderer should be enough for now. We can always add more context as needed later

PR #6883, opened 2023-09-21
Remove uses on React.PropsWithChildren where there are no props

2metres:
**User-Facing Changes**
None

**Description**
Stops lint errors


jtbandes:
I don’t think we need to change this, there are no lint errors on main. Try running yarn and restarting your editor.

PR #6884, opened 2023-09-21
Fix issue with disappearing preloaded plots

cfoust:
**User-Facing Changes**
None.

**Description**
Plots that used block data were cleared when switching tabs. This PR fixes that.

The problem was that `addBlock` could come before `register` had completed, and thus `evictCache` would clear out the new block that just got added. (Theoretically this is a timing problem but it's more complicated than it seems at first glance; waiting for `register` to complete before doing `addBlock` would not make sense, since the `addBlock` mechanism in the rendering thread has already done the work to determine that a client actually needs the block data.)


snosenzo (approved):
would be good to have a story/test for this at some point

PR #6886, opened 2023-09-21
Fix TextAnnotations not showing separate alpha channels

snosenzo:
**User-Facing Changes**

 - Fix TextAnnotations not showing specified alpha channels

**Description**
Update `three-text` so that it now takes foreground and background alpha channels. 
Remove use of setOpacity (no longer exists).

FG-5007


packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/annotations/RenderableTextAnnotation.ts
```diff
@@ -108,20 +108,23 @@ export class RenderableTextAnnotation extends Renderable<BaseUserData, /*TRender
     }
 
     if (this.#annotationNeedsUpdate) {
+      // Need to keep it transparent so that other transparent objects aren't rendered over it
+      const colorAlpha = Math.min(textColor.a, 0.999);
```
jtbandes, line 112: Looks like this trick was introduced with the "foreground image opacity" feature that we've since removed...is it still necessary?

jtbandes (approved):
🙌

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/annotations/RenderableTextAnnotation.ts
```diff
@@ -109,22 +109,20 @@ export class RenderableTextAnnotation extends Renderable<BaseUserData, /*TRender
 
     if (this.#annotationNeedsUpdate) {
       // Need to keep it transparent so that other transparent objects aren't rendered over it
```
jtbandes, line 111: outdated comment?

PR #6892, opened 2023-09-22
Use preload data for all plots again

cfoust:
**User-Facing Changes**
None.

**Description**
Now we attempt to load all preload data for a plot regardless of whether it's bounded.


jtbandes (approved):
I think it’s worth calling out the user facing change in release notes — e.g. “restored preloading behavior for plot panels with bounded x axes”

PR #6894, opened 2023-09-22
Inject custom scene extensions from `injectedFeatures`

snosenzo:
**User-Facing Changes**

 - n/a

**Description**

Added `InjectedFeatures` which is an entrypoint that allows features to be injected from an outside context.

Modified `ImageMode` and `Images` scene extensions and the `ImageRenderable` class that they use to be more friendly to subclassing to add support for additional schema types.


packages/studio-base/src/context/AppContext.ts
```diff
@@ -26,6 +30,15 @@ interface IAppContext {
     initialState?: Partial<WorkspaceContextStore>,
   ) => StoreApi<WorkspaceContextStore>;
 }
+export type VerifiedFeatureMap = {
+  "ThreeDeeRender.customSceneExtensions"?: {
```
foxymiles, line 34: Can we avoid duplicating this string?

packages/studio-base/src/context/AppContext.ts
```diff
@@ -26,6 +30,15 @@ interface IAppContext {
...
+    customSceneExtensions: DeepPartial<SceneExtensionConfig>;
+  };
+};
+
+export interface VerifiedFeatureStore {
```
foxymiles, line 39: Why interface here and type elsewhere?

foxymiles: Also elsewhere in the code we tend to reserve `store` for things like zustand stores, not plain maps/objects.

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/constants.ts
```diff
@@ -0,0 +1,14 @@
...
+export const IMAGE_TOPIC_PATH = ["imageMode", "imageTopic"];
+export const CALIBRATION_TOPIC_PATH = ["imageMode", "calibrationTopic"];
+export const IMAGE_TOPIC_UNAVAILABLE = "IMAGE_TOPIC_UNAVAILABLE";
+export const CALIBRATION_TOPIC_UNAVAILABLE = "CALIBRATION_TOPIC_UNAVAILABLE";
+export const MISSING_CAMERA_INFO = "MISSING_CAMERA_INFO";
+export const IMAGE_TOPIC_DIFFERENT_FRAME = "IMAGE_TOPIC_DIFFERENT_FRAME";
+export const CAMERA_MODEL = "CameraModel";
+export const DEFAULT_FOCAL_LENGTH = 500;
+export const REMOVE_IMAGE_TIMEOUT_MS = 50;
+export const DECODE_IMAGE_ERR_KEY = "CreateBitmap";
```
foxymiles, line 14: Would it be cleaner to bundle these up into some wrapper object instead of as a bunch of free values?

foxymiles (approved):
Not familiar with the functionality here so just left some comments on implementation.

packages/studio-base/src/panels/ThreeDeeRender/index.tsx
```diff
@@ -64,15 +69,34 @@ function ThreeDeeRenderAdapter(interfaceMode: InterfaceMode, props: Props) {
...
+  const { verifiedFeatures: verifiedFeatureStore } = useAppContext();
+  const sceneExtensionConfig = useMemo(() => {
+    if (verifiedFeatureStore == undefined) {
+      return undefined;
+    }
+    const extensionConfigOverride =
+      verifiedFeatureStore.availableFeatures["ThreeDeeRender.customSceneExtensions"]
+        ?.customSceneExtensions;
+    return extensionConfigOverride;
+  }, [verifiedFeatureStore]);
```
jtbandes, line 81: nit: I find it difficult to read when there are 3 separate names referring to the same thing: `sceneExtensionConfig`, `extensionConfigOverride`, and `customSceneExtensions` – maybe some of them can be removed or renamed to match? (I know that the no-shadow lint rule sometimes prevents this)

packages/studio-base/src/panels/ThreeDeeRender/index.tsx
```diff
@@ -64,15 +69,34 @@ function ThreeDeeRenderAdapter(interfaceMode: InterfaceMode, props: Props) {
   const crash = useCrash();
 
   const forwardedAnalytics = useForwardAnalytics();
+  const { verifiedFeatures: verifiedFeatureStore } = useAppContext();
```
jtbandes, line 72:
```suggestion
  const { verifiedFeatures } = useAppContext();
```

packages/studio-base/src/context/AppContext.ts
```diff
@@ -27,6 +31,16 @@ interface IAppContext {
   ) => StoreApi<WorkspaceContextStore>;
 }
 
+export type VerifiedFeatureMap = {
+  "ThreeDeeRender.customSceneExtensions"?: {
+    customSceneExtensions: DeepPartial<SceneExtensionConfig>;
+  };
+};
+
+export type VerifiedFeatures = {
```
jtbandes, line 40: Not crazy about the "verified" features name, it's alright but here are some other ideas:
- InjectedFeatures
- ExtraFeatures
- FeatureOverrides

packages/studio-base/src/context/AppContext.ts
```diff
@@ -27,6 +31,16 @@ interface IAppContext {
...
+  availableFeatures: VerifiedFeatureMap;
```
jtbandes, line 41: Feels like there is an extra level of nesting that I'm not sure we need. Are you thinking of adding more members here in the future?

packages/studio-base/src/context/AppContext.ts
```diff
@@ -3,12 +3,15 @@
...
 
 import { createContext, useContext } from "react";
+import { DeepPartial } from "ts-essentials";
 import { StoreApi } from "zustand";
 
 import { AppBarMenuItem } from "@foxglove/studio-base/components/AppBar/types";
 import { LayoutData } from "@foxglove/studio-base/context/CurrentLayoutContext";
 import { WorkspaceContextStore } from "@foxglove/studio-base/context/Workspace/WorkspaceContext";
 
+import { SceneExtensionConfig } from "../panels/ThreeDeeRender/SceneExtensionConfig";
```
jtbandes, line 13: Kinda sketchy that we are importing from within a specific panel here. However in some ways it does make sense. The main problem would be if the panel also imports from AppContext creating a circular import.

Let's at least use an absolute rather than relative path here since it's referencing a totally separate component

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageMode.ts
```diff
@@ -291,38 +298,29 @@ export class ImageMode
...
   }
 
-  // eslint-disable-next-line @foxglove/no-boolean-parameters
-  #setHasCalibrationTopic = (hasCalibrationTopic: boolean): void => {
-    if (hasCalibrationTopic) {
-      this.renderer.disableImageOnlySubscriptionMode();
-    } else {
-      this.renderer.enableImageOnlySubscriptionMode();
-    }
-  };
```
jtbandes, line 313: was it needed to remove this or is this just cleanup? In some ways I think having a separate method for this can make it clearer - to help understand _why_ we are enabling/disabling image-only subscription mode

packages/studio-base/src/panels/ThreeDeeRender/renderables/Images/ImageRenderable.ts
```diff
@@ -369,6 +387,22 @@ export class ImageRenderable extends Renderable<ImageUserData> {
   }
 }
 
+export interface IImageRenderable extends Renderable<ImageUserData> {
```
jtbandes, line 390: maybe move to a separate `types` file?

Actually I'm wondering how necessary this interface is. It seems like you intend for `initRenderable` to return a subclass rather than a potentially whole separate class, so maybe the type annotation should just say that it returns `ImageRenderable`?

packages/studio-base/src/panels/ThreeDeeRender/renderables/Images/ImageRenderable.ts
```diff
@@ -83,12 +81,21 @@ export class ImageRenderable extends Renderable<ImageUserData> {
     super(topicName, renderer, userData);
   }
 
+  protected isDisposed(): boolean {
```
jtbandes, line 84: could also just be `protected isDisposed: boolean`?

jtbandes (approved):
Overall seems fine. It doesn't feel like the injected extension/renderable slots in "cleanly" but the bar for clean interfaces in the 3D panel is not high...and I suspect we won't need to change this much in the future. I commented on some opportunities for cleanup/simplification.

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageMode.ts
snosenzo: hm I'm conflicted on this. On one hand I think keeping this adds indirection for what this function actually does, but yeah I do also agree that it does help with the _why_. I'll add it back

packages/studio-base/src/context/AppContext.ts
snosenzo: yeah I'm thinking there might be more there in the future

packages/studio-base/src/context/AppContext.ts
snosenzo: This felt like a name that allowed it to be found more easily and was more descriptive (maybe too descriptive?). I'm open to using `InjectedFeatures` though

packages/studio-base/src/panels/ThreeDeeRender/renderables/Images/ImageRenderable.ts
snosenzo: I didn't want to allow access to assigning to the underlying variable so I made this a function instead to keep the variable `#disposed` private

packages/studio-base/src/context/AppContext.ts
jtbandes: @snosenzo did you have thoughts on this?

jtbandes (approved):

packages/studio-base/src/context/AppContext.ts
snosenzo: oop updated to use const object

PR #6903, opened 2023-09-25
Jump to settings button for selected object in 3d panel

foxymiles:
**User-Facing Changes**
Add button to jump to per-topic settings in the 3d panel object inspector.

**Description**
Put this alongside the existing link to open a raw messages panel:
Screenshot 2023-09-25 at 9 43 50 AM


snosenzo (approved):
I think this looks good. Can we get a story if it's easy enough? we have a couple of stories that already show the info overlay after picking an object, would be nice to have this as a continuation

foxymiles:
> I think this looks good. Can we get a story if it's easy enough? we have a couple of stories that already show the info overlay after picking an object, would be nice to have this as a continuation

Do you mean an interactive story? There are several stories that show the new button already.

snosenzo:
> Do you mean an interactive story? There are several stories that show the new button already.
@foxymiles  yeah I do. Looking for a story that clicks the button and opens the topic in the settings panel

snosenzo:
@jtbandes I added stories if you wouldn't mind checking their code to make sure they look fine. Had some strange issues where I had to send two point events and not sure if there's anything I can do about it

jtbandes (approved):

PR #6926, opened 2023-09-27
Test suite for plot worker state machine

cfoust:
**User-Facing Changes**
None--it should work exactly the way it did before.

**Description**
I broke this changeset into two PRs for clarity while reviewing.
1. This one, which breaks up the monolithic `useDatasets.worker.ts` into smaller, functionally related pieces.
2. #6927 , which adds a test suite covering the entire state machine's functionality.

There was a lot going on in the plot dataset worker; the entirety of the state machine was in one file and it made extensive use of global state. This PR breaks it into a pure, modular state machine that can be tested more easily, and then adds a complete test suite.

This PR **does not change the existing state machine at all.** It works exactly the same way that it does in `main`. The only notable difference is that as a result of making all state mutations pure transformations, side effects are produced and handled separately.


packages/studio-base/src/panels/Plot/useDatasets.worker.ts
```diff
@@ -260,314 +123,92 @@ function rebuild(id: string) {
...
-function addBlock(block: Messages, resetTopics: string[]): void {
-  const topics = R.keys(block);
-
-  blocks = R.pipe(
-    // Remove data for any topics that have been reset
-    R.omit(resetTopics),
-    // Merge the new block into the existing blocks
-    (newBlocks) => R.mergeWith(R.concat, newBlocks, block),
-  )(blocks);
+function handleEffects([newState, effects]: StateAndEffects): void {
```
jtbandes, line 132: nice, this is much cleaner & easier to reason about 🎉

packages/studio-base/src/panels/Plot/processor/clients.ts
```diff
@@ -0,0 +1,189 @@
...
+
+  if (params == undefined || view == undefined) {
+    return undefined;
+  }
+
+  const { bounds: blockBounds } = blockData;
+  const { bounds: currentBounds } = currentData;
+
+  let datasets: PlotData[] = [];
+  if (blockBounds.x.min <= currentBounds.x.min && blockBounds.x.max > currentBounds.x.max) {
```
jtbandes, line 180: I don't think I realized until now that we are using `x` comparisons to decide whether to include current data – are these using a float32 precision and if so, do you think there's a risk of merging data incorrectly if message receiveTimes are sufficiently large & close together?

jtbandes (approved):
New organization seems much cleaner 👍

snosenzo (approved):

packages/studio-base/src/panels/Plot/processor/state.ts
```diff
@@ -0,0 +1,126 @@
...
+  ([state, effects]: StateAndEffects): StateAndEffects => {
+    return [mutator(state), effects];
+  };
+
+export const findClient = (state: State, id: string): Client | undefined =>
+  R.find((client) => client.id === id, state.clients);
+
+export const mutateClient = (state: State, id: string, newClient: Client): State => ({
+  ...state,
+  clients: state.clients.map((client) => (client.id === id ? newClient : client)),
```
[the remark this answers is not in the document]
cfoust: Exceptions/errors are side effects, and nasty ones at that. In pure code this is sometimes handled with a `Result<S, T>` where `T` is any error type, but that's almost certainly overkill here--we just silently ignore unknown ids.

packages/studio-base/src/panels/Plot/processor/accumulate.ts
```diff
@@ -0,0 +1,97 @@
...
+}
+
+export function buildPlot(
+  metadata: MetadataEnums,
+  globalVariables: GlobalVariables,
+  params: PlotParams,
+  messages: Messages,
+): PlotData {
+  const { paths, invertedTheme, startTime, xAxisPath, xAxisVal } = params;
+  return buildPlotData({
```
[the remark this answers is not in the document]
cfoust: Yeah, cleaning this up is on my to-do list and I address this in another diff (that I have not made into a PR quite yet) with this PR as a base. It's a little weird but it's a holdover from how the code used to be set up and I've resisted touching it because of how delicate this code tree is.

packages/studio-base/src/panels/Plot/processor/clients.ts
```diff
@@ -0,0 +1,189 @@
...
+      ...client,
+      topics,
+      blocks: accumulate(metadata, globalVariables, initialState, params, blocks),
+      current: accumulate(metadata, globalVariables, initialState, params, current),
+    },
+    [rebuildClient(id)],
+  ];
+}
+
+export function receiveVariables(variables: GlobalVariables, state: State): StateAndEffects {
```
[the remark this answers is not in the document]
cfoust: This is a great call-out--I don't recall why I named it this way

packages/studio-base/src/panels/Plot/useDatasets.worker.ts
cfoust: 🎉 it does make it easier to see what's happening and when!

packages/studio-base/src/panels/Plot/processor/clients.ts
cfoust: yes, for the same reasons we discussed in that thread. (the bounds are calculated based on the values in a `Float32Array`, so presumably suffer from the same issue). it's worth taking a deeper look at this

packages/studio-base/src/panels/Plot/processor/state.ts
```diff
@@ -0,0 +1,126 @@
...
+  clients: Client[];
+  globalVariables: GlobalVariables;
+  blocks: Messages;
+  current: Messages;
+  metadata: MetadataEnums;
+};
+
+export enum SideEffectType {
+  Rebuild = "rebuild",
+  Data = "data",
```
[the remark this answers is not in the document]
cfoust: that's a great idea!

packages/studio-base/src/panels/Plot/processor/state.ts
```diff
@@ -0,0 +1,126 @@
...
+  Partial = "partial",
```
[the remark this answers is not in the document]
cfoust: yup, vestigial, removed it!

cfoust:
All stories rendered correctly on my local machine. I have a task to fix these flaky stories (once and for all, we can hope) soon.

packages/studio-base/src/panels/Plot/processor/clients.ts
```diff
@@ -0,0 +1,186 @@
...
+
+  return mapClients((client) => {
+    const { params } = client;
+    if (params == undefined) {
+      return noEffects(client);
+    }
+
+    // We only want to rebuild clients whose paths actually change when global
+    // variables do
+    const changedPaths = R.pipe(
```
defunctzombie, line 67: Can't say I am a fan of this R stuff - what makes this better than regular iteration and mapping?

cfoust: This is a can of worms and probably easier to discuss in person but:
1. Breaks up complicated transformations into a sequence of pure (and, ideally, comprehensible) steps.
2. Encourages the composition of small functions that do just one thing.

PR #6927, opened 2023-09-27
Test suite for plot worker state machine (just the tests)

cfoust:
**User-Facing Changes**
None--see other PR.

**Description**
Adds tests for the plot worker state machine introduced in #6926.

Test coverage is comprehensive. The state machine itself (`clients.ts`, `messages.ts`, `state.ts`) has 100% coverage; getting `jest`'s numbers to be perfect is a little annoying so I won't bother with that for now.
image

### DIFF
--- a/packages/studio-base/src/panels/Plot/processor/clients.test.ts
+++ b/packages/studio-base/src/panels/Plot/processor/clients.test.ts
@@ -1,0 +1,265 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { PlotViewport } from "@foxglove/studio-base/components/TimeBasedChart/types";
+
+import {
+  unregister,
+  compressClients,
+  MESSAGE_CULL_THRESHOLD,
+  register,
+  refreshClient,
+  receiveVariables,
+  updateParams,
+  updateView,
+  getClientData,
+} from "./clients";
+import { init, initClient, State, rebuildClient } from "./state";
+import {
+  FAKE_PATH,
+  CLIENT_ID,
+  createClient,
+  createData,
+  createPath,
+  createState,
+  createMessages,
+  createParams,
+} from "./testing";
+import { DatasetsByPath, PlotPath, TypedDataSet } from "../internalTypes";
+
+describe("refreshClient", () => {
+  it("ignores client without params", () => {
+    const client = initClient(CLIENT_ID, undefined);
+    const initial: State = { ...init(), clients: [client] };
+    const [newClient, effects] = refreshClient(client, initial);
+    expect(newClient).toEqual(client);
+    expect(effects).toEqual([]);
+  });
+
+  it("regenerates plots and triggers a rebuild", () => {
+    const initial = createState(FAKE_PATH);
+    const {
+      clients: [client],
+    } = initial;
+    if (client == undefined) {
+      throw new Error("client missing somehow");
+    }
+    const [newClient, effects] = refreshClient(client, initial);
+    expect(effects).toEqual([rebuildClient(client.id)]);
+    // we aren't testing accumulate(); just check whether the reference changed
+    expect(newClient.blocks).not.toBe(client.blocks);
+    expect(newClient.current).not.toBe(client.current);
+  });
+});
+
+describe("receiveVariables", () => {
+  const vars = {
+    foo: "bar",
+  };
+  it("does nothing when client does not use variables", () => {
+    const before = createState(FAKE_PATH);
+    const [after] = receiveVariables(vars, before);
+    expect(after.clients[0]).toEqual(before.clients[0]);
+  });
+  it("does nothing when client has no params", () => {
+    const client = initClient(CLIENT_ID, undefined);
+    const before: State = { ...init(), clients: [client] };
+    const [after] = receiveVariables(vars, before);
+    expect(after.clients[0]).toEqual(before.clients[0]);
+  });
+  it("does nothing when client has invalid path", () => {
+    const before = createState("你好");
+    const [after] = receiveVariables(vars, before);
+    expect(after.clients[0]).toEqual(before.clients[0]);
+  });
+  it("refreshes the client when it does use variables", () => {
+    const before = createState(`/topic.field[:]{id==$foo}`);
+    const [after, effects] = receiveVariables(vars, before);
+    expect(after.clients[0]).not.toBe(before.clients[0]);
+    expect(effects).toEqual([rebuildClient(CLIENT_ID)]);
+  });
+});
+
+describe("updateParams", () => {
+  it("ignores missing id", () => {
+    const before = createState();
+    const [after] = updateParams("123", createParams(), before);
+    expect(after.clients[0]).toEqual(before.clients[0]);
+  });
+  it("updates client", () => {
+    const before = createState("/foo.bar");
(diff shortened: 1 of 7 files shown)


packages/studio-base/src/panels/Plot/processor/clients.test.ts
```diff
@@ -0,0 +1,265 @@
...
+    expect(effects).toEqual([rebuildClient(CLIENT_ID)]);
+  });
+});
+
+describe("unregister", () => {
+  it("removes an existing client", () => {
+    const after = unregister(CLIENT_ID, createState());
+    expect(after.clients.length).toEqual(0);
+  });
+});
```
jtbandes, line 140: I find it a bit confusing that `init()` creates a state with no clients and `createState()` creates a state with one client. Maybe there is some better naming that could make this clearer, e.g. `createEmptyState()` / `createSingleClientState()`?

packages/studio-base/src/panels/Plot/processor/clients.test.ts
```diff
@@ -0,0 +1,265 @@
...
+
+describe("compressClients", () => {
+  it("does nothing if not live", () => {
+    const before = { ...createState(), isLive: false };
+    const [after, effects] = compressClients(before);
+    expect(effects.length).toEqual(0);
+    expect(after).toEqual(before);
+  });
```
jtbandes, line 148: Since a blank state is being used here, would `after` and `before` actually differ if `isLive` had been true? (in other words, should we add another test that's identical but with `isLive: true` to show that something different happens? intuitively I'm not sure I would expect `compressClients` to do anything on an empty state)

packages/studio-base/src/panels/Plot/processor/clients.test.ts
```diff
@@ -0,0 +1,265 @@
...
+          ...createData(path, 2),
+          bounds: {
+            x: { min: 2, max: 8 },
+            y: { min: 0, max: 0 },
+          },
+        },
+      },
+    };
+    const result = getClientData(client);
+    expect(getDataset(result?.datasets)?.data.length).toEqual(1);
```
jtbandes, line 223: Not sure how noisy this would be but I think it could be illustrative to actually just use `expect(result?.datasets).toEqual({ the actual data here })` so you could visually see what the resulting data looks like

packages/studio-base/src/panels/Plot/processor/messages.test.ts
```diff
@@ -0,0 +1,164 @@
...
+} from "./testing";
+import { EmptyPlotData } from "../plotData";
+
+describe("receiveMetadata", () => {
+  it("updates metadata", () => {
+    const before = init();
+    const after = receiveMetadata(FAKE_TOPICS, FAKE_DATATYPES, before);
+    expect(after.metadata).not.toEqual(before);
+  });
+});
```
jtbandes, line 28: If there are no side effects to receiving new metadata, what does it get used for?

packages/studio-base/src/panels/Plot/processor/messages.test.ts
```diff
@@ -0,0 +1,164 @@
...
+describe("addBlock", () => {
+  it("resets the requested topics", () => {
+    const [after] = addBlock({}, [FAKE_TOPIC], {
+      ...init(),
+      blocks: {
+        [FAKE_TOPIC]: [],
+      },
+    });
+    expect(Object.entries(after.blocks).length).toEqual(0);
+  });
```
jtbandes, line 51: Testing the reset logic might be clearer if there were some messages before the reset and are none after, instead of having empty blocks before & after

packages/studio-base/src/panels/Plot/processor/messages.test.ts
```diff
@@ -0,0 +1,164 @@
...
+          },
+        },
+      ],
+    };
+    const [after, effects] = addBlock({}, [], before);
+    expect(after.clients[0]).toEqual(before.clients[0]);
+    expect(effects.length).toEqual(0);
+  });
+  it("ignores client with no related topics", () => {
+    const before: State = createState("/bar.baz");
```
jtbandes, line 86: why do some of these tests use `receiveMetadata` and others don't? Are we implicitly testing some dependence on metadata and not just on the client's list of paths?

packages/studio-base/src/panels/Plot/processor/messages.test.ts
```diff
@@ -0,0 +1,164 @@
...
+  it("clears out the plot data for a client", () => {
+    const before: State = receiveMetadata(FAKE_TOPICS, FAKE_DATATYPES, createState(FAKE_PATH));
+    const [after, effects] = addBlock(
+      createMessages(FAKE_TOPIC, FAKE_SCHEMA, 1),
+      [FAKE_TOPIC],
+      before,
+    );
+    expect(effects).toEqual([rebuildClient(CLIENT_ID)]);
+    expect(after.clients[0]?.blocks).not.toEqual(before.clients[0]?.blocks);
+    expect(after.clients[0]?.blocks.cursors?.[FAKE_TOPIC]).toEqual(1);
```
jtbandes, line 106: This is the only place I see `cursors` mentioned in the tests – is there more we can verify about how they behave?

jtbandes (approved):
Great stuff, happy to see more test coverage, and I think reading the tests has helped me understand the logic a bit better too.

packages/studio-base/src/panels/Plot/processor/clients.test.ts
cfoust: `createState` is only used in testing 😄

packages/studio-base/src/panels/Plot/processor/clients.test.ts
cfoust: The state here is not empty (which perhaps is indicative of naming issues with `createState`) and does include one client, so this test correctly ensures that no compression happens in that scenario as far as I can determine

packages/studio-base/src/panels/Plot/processor/clients.test.ts
cfoust: I avoided this for the simple reason that I didn't want tests to break if the underlying representation changed, since that's not what this test is checking

packages/studio-base/src/panels/Plot/processor/messages.test.ts
cfoust: Theoretically `receiveMetadata` not resulting in rebuilds could lead to issues, but in practice it's hard to cook up scenarios in which that might happen. It's worth revisiting.

packages/studio-base/src/panels/Plot/processor/messages.test.ts
cfoust: it's the same issue you identified above with `receiveMetadata` -- I think ideally we would fix this since there is a required order and `receiveMetadata` should have side effects

packages/studio-base/src/panels/Plot/processor/messages.test.ts
cfoust: this is intentionally blackboxed as a part of `accumulate.ts`, which really could use more tests on its own, but this is a bit of a philosophical question: if the functions in `accumulate.ts` are covered 100% by the tests and the other tests are reliant on these functions behaving correctly, presumably we could just ignore the implementation details of `accumulate.ts`. For example, I have a diff sitting around that rips up everything in `accumulate.ts` but leaves the API intact--tests still pass.

cfoust:
Merging since this is just going into the main branch.

PR #6933, opened 2023-09-30
Improve Plot empty state

jtbandes:
**User-Facing Changes**

**Description**
Basically the same as #6860, but for the Plot panel. Also includes some minor cleanup/fixes leftover from #6860.

- Improve the empty state so the panel doesn't just say "Series 1", but teaches you how to add a series.
- Hide the "X" delete button when there are no series configured.

image

### DIFF
--- a/packages/studio-base/src/panels/Plot/PlotLegendRow.tsx
+++ b/packages/studio-base/src/panels/Plot/PlotLegendRow.tsx
@@ -4,12 +4,14 @@
 
 import {
   Dismiss12Regular,
+  Add12Regular,
   ErrorCircle16Filled,
   Square12Filled,
   Square12Regular,
 } from "@fluentui/react-icons";
 import { ButtonBase, Checkbox, Tooltip, Typography } from "@mui/material";
 import { useMemo, useState } from "react";
+import { useTranslation } from "react-i18next";
 import { makeStyles } from "tss-react/mui";
 import { v4 as uuidv4 } from "uuid";
 
@@ -136,6 +138,7 @@ export function PlotLegendRow({
   const { id: panelId } = usePanelContext();
   const { setSelectedPanelIds } = useSelectedPanels();
   const { classes, cx } = useStyles();
+  const { t } = useTranslation("plot");
 
   const correspondingData = useMemo(() => {
     if (!showPlotValuesInLegend) {
@@ -211,7 +214,7 @@ export function PlotLegendRow({
           variant="body2"
           className={cx({ [classes.disabledPathLabel]: !path.enabled })}
         >
-          {plotPathDisplayName(path, index)}
+          {index === paths.length ? t("clickToAddASeries") : plotPathDisplayName(path, index)}
         </Typography>
         {hasMismatchedDataLength && (
           <Tooltip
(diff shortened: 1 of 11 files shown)


packages/studio-base/src/panels/Plot/PlotLegendRow.tsx
```diff
@@ -211,7 +214,7 @@ export function PlotLegendRow({
           variant="body2"
           className={cx({ [classes.disabledPathLabel]: !path.enabled })}
         >
-          {plotPathDisplayName(path, index)}
+          {index === paths.length ? t("clickToAddASeries") : plotPathDisplayName(path, index)}
```
snosenzo, line 217: I thought that this would make the last legend row an "add series" button it doesn't appear to do that. So what is this doing? would taking extracting out `index === path.length` to be equal to another named variable make it a bit more clear?

packages/studio-base/src/panels/Plot/settings.ts
```diff
@@ -17,18 +17,27 @@ import { lineColors } from "@foxglove/studio-base/util/plotColors";
...
+export const DEFAULT_PATH: PlotPath = Object.freeze({
+  timestampMethod: "receiveTime",
+  value: "",
+  enabled: true,
+});
+
 const makeSeriesNode = memoizeWeak(
-  (path: PlotPath, index: number, t: TFunction<"plot">): SettingsTreeNode => {
+  // eslint-disable-next-line @foxglove/no-boolean-parameters
+  (path: PlotPath, index: number, canDelete: boolean, t: TFunction<"plot">): SettingsTreeNode => {
```
snosenzo, line 28: why can't we do object param props?

snosenzo (commented):
would be nice if clicking the `Click to add a series` button highlighted the place in the settings to fill in, because otherwise I half expect it to show something on the plot. It feels a bit misleading because the button itself isn't adding a series, just opening the settings

packages/studio-base/src/panels/Plot/settings.ts
jtbandes: I believe it would defeat the purpose of `memoizeWeak`

jtbandes:
> would be nice if clicking the `Click to add a series` button highlighted the place in the settings to fill in

It does:

https://github.com/foxglove/studio/assets/14237/a832baf0-74fb-4505-872c-6e1567fa459a

Clicking it a 2nd time doesn't work but I think that's an existing issue with the way the highlighting is being done.

packages/studio-base/src/panels/Plot/PlotLegendRow.tsx
jtbandes: Added new variable + comment

snosenzo (approved):
I think I'd prefer it highlight the message path rather than the section heading, but otherwise looks good

packages/studio-base/src/panels/Plot/PlotLegendRow.tsx
```diff
@@ -167,6 +170,9 @@ export function PlotLegendRow({
     return value;
   }, [showPlotValuesInLegend, hoverValue?.value, currentTime, correspondingData]);
 
+  // When there are no series configured we render an extra row to show an "add series" button.
+  const isAddSeriesRow = index === paths.length;
```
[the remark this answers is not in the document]
jtbandes: Yeah that would be equivalent, the only time this would happen would be when paths.length===0 but we render an extra row anyway.

jtbandes:
I'll leave the highlighting behavior as is for now since it was the pre-existing highlight behavior

PR #6841, opened 2023-09-17
Remove tabs from User Scripts panel

2metres:
**User-Facing Changes**
Remove tab bar from user script panel as it was confusing and restyle Sidebar Tabs to be more aligned with our styles

Screen Shot 2023-09-17 at 12 42 53 pm

PR #6936, opened 2023-10-03
Reduce ready signal count for failing Plot story

jtbandes:
**User-Facing Changes**
None

**Description**
This story has been timing out in a couple of my branches. Unfortunately that requires re-running the whole storybook job each time to get past it.


jtbandes:
Actually I'm just going to fold this into https://github.com/foxglove/studio/pull/6841 because I can't get the stories to run successfully otherwise, even with a few rerun attempts.

PR #6937, opened 2023-10-03
Fix state transition tooltip not showing label / path value

achim-k:
**User-Facing Changes**
Fix state transition tooltip not showing label / path value

**Description**
The tooltip was showing the stringified series index instead of the appropriate label / path value.

| Before                                                                                           | After (this PR)                                                                                  |
|--------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------|
| image | image |

### DIFF
--- a/packages/studio-base/src/panels/StateTransitions/messagesToDatasets.ts
+++ b/packages/studio-base/src/panels/StateTransitions/messagesToDatasets.ts
@@ -45,12 +45,12 @@ type Args = {
  * dataset with different labels and colors applied per-point.
  */
 export default function messagesToDatasets(args: Args): ChartDatasets {
-  const { path, pathIndex, startTime, y, blocks } = args;
+  const { path, startTime, y, blocks } = args;
 
   const dataset: ChartDataset = {
     borderWidth: 10,
     data: [],
-    label: pathIndex.toString(),
+    label: path.label ?? "" ? path.label : path.value,
     pointBackgroundColor: "rgba(0, 0, 0, 0.4)",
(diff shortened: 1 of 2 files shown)


jtbandes:
Any idea why the color is black? Is it because of the `pointBackgroundColor: "rgba(0, 0, 0, 0.4)",`? I'm not sure if that is a regression but it seems strange.

jtbandes (approved):

packages/studio-base/src/panels/StateTransitions/messagesToDatasets.ts
```diff
@@ -45,12 +45,12 @@ type Args = {
...
  */
 export default function messagesToDatasets(args: Args): ChartDatasets {
-  const { path, pathIndex, startTime, y, blocks } = args;
+  const { path, startTime, y, blocks } = args;
 
   const dataset: ChartDataset = {
     borderWidth: 10,
     data: [],
-    label: pathIndex.toString(),
+    label: path.label ?? "" ? path.label : path.value,
```
snosenzo, line 53: I think this should be
```suggestion
    label: path.label ? path.label : path.value,
```

snosenzo (approved):
one small change, otherwise looks good

packages/studio-base/src/panels/StateTransitions/messagesToDatasets.ts
jtbandes: I suspect eslint forced the issue here... but yeah this looked weird to me too.

PR #6938, opened 2023-10-03
Add support for coloring pointclouds by distance

IvDmNe:
**User-Facing Changes**
Added option to color points in 3D and Image panels by norm of their coordinates.

**Description**

image
image

I added a ColorFieldOption "\<distance\>" and a new reader function for packedColorReader that uses 3 another readers for X, Y, Z coordinates and returns norm of these values.

I am not sure about one moment - should "\<distance\>" option be added in such hardcoded way or the name should be some variable to support translation to another languages?


jtbandes (commented):
Thanks for the PR and sorry for the delay in response. I think we are in favor of adding this feature. My main concern is that the `"<distance>"` string is being appended to a list of `field` names that come from user data, so in the rare event there is a conflict, that would cause problems. I would suggest that you actually add a new field to the config, maybe something like `config.colorFieldComputed: "distance" | undefined`. Then it becomes possible to disambiguate this special computed value from any user-defined point cloud field. The `colorFieldOptions` can still be a combination of user-defined and computed options, but the `value` of each option can be prefixed to avoid conflicting with a value of `"distance"`.

Re translation, the option `label` can be translated, e.g. `t("colorFieldDistance")` and then add `colorFieldDistance: "<distance>"` to `i18n/en/threeDee.ts`.

You should also be able to add an entry to `foxglove.PointCloud.stories.tsx` to demonstrate the new functionality in a screenshot test.

IvDmNe:
I have done as you suggested: automatically computed distance option now has prefix ```_auto_```. In ```updatePointCloud``` function it is checked if specified value in config has this prefix. If so, it sets ```config.ColorFieldComputed``` to ```distance```, which is used to create ```PackedColorReader```. Is this what you meant?

If everything is alright, I will start to add the story.

amacneil:
:+1: great feature

snosenzo (approved):
Thanks so much for the contribution!

PR #3883, opened 2022-07-20
Rename Node Playground to User Scripts

foxymiles:
**User-Facing Changes**
Node Playground is now called `User Scripts`.

**Description**
This just changes the name displayed to the user. We could also do a bigger PR and change the names in the codebase.

PR #6945, opened 2023-10-04
Finish renaming "user nodes" to "user scripts"

jtbandes:
**User-Facing Changes**
None

**Description**
In https://github.com/foxglove/studio/pull/3883 we renamed the app feature from "Node Playground" to "User Scripts". However, many references to "node playground" & "user nodes" remained in the code. This PR renames everything that remained, except for the strings which are actually persisted in layouts.


snosenzo (approved):
thanks for doing this! will be good to have a common language around this. Took a glance through most of this and looks good, just have some conflicts from my subs pr that I merged earlier today

defunctzombie:
👏

PR #5819, opened 2023-04-25
Delete some unused code around json primitives

defunctzombie:
**User-Facing Changes**
None

**Description**
JSON primitives are not a thing in ROS. This was a bespoke webviz-ism that we removed long ago. This cleans up some remnants that are no longer used.

FG-3297

PR #6950, opened 2023-10-05
Allow getMessagePathDataItems and getValueActionForValue to work with schemaless messages

jtbandes:
**User-Facing Changes**
Fixed the ability to plot values from topics that have no specified schema.

**Description**
Alternative to / closes #6688 
Fixes FG-4596

In #5819 we removed some code related to a Webviz/Cruise-specific feature where the `json` keyword could be used as a ROS primitive type. This inadvertently broke the ability to plot values from _schemaless_ topics. This PR restores the desired behavior by adjusting the traversal logic in `getMessagePathDataItems` & `getValueActionForValue` so that nested values can be extracted even if the schema does not contain information about them.

image


packages/studio-base/src/components/MessagePathSyntax/useCachedGetMessagePathDataItems.ts
```diff
@@ -236,15 +236,21 @@ export function getMessagePathDataItems(
...
         const fieldName = prevPathItem.name;
-        const enumMap = enumValues[structureItem.datatype];
+        const enumMap = structureItem != undefined ? enumValues[structureItem.datatype] : undefined;
         constantName = enumMap?.[fieldName]?.[value];
       }
       queriedData.push({ value, path, constantName });
-    } else if (pathItem.type === "name" && structureItem.structureType === "message") {
+    } else if (
+      pathItem.type === "name" &&
+      (structureItem == undefined || structureItem.structureType === "message")
```
snosenzo, line 245: This check looks a bit strange because it passes for undefined or a specific structureType, but then the next `else if` also checks `structureItem == undefined`. It looks like the more important thing is the `pathItem.type`, but why does it also require `structureType`? 
Is there a way we could make this flow a bit more clear?

snosenzo (approved):
why do we use the `structureItem.structureType` at all?

packages/studio-base/src/components/MessagePathSyntax/useCachedGetMessagePathDataItems.ts
jtbandes: It checks structureType because it's trying to resolve a path segment `[...].fieldName` and it needs to know that the current object is actually a "message" i.e. Record with named fields (not an array). The change being introduced is to also allow traversing through unknown structures. I suppose we could have similar logic to the `isObjectElement` thing from the getValueActionForValue code...

Ultimately I am unhappy with all of the message path traversal / structures code and I'd love to delete it someday. However it does technically provide a small benefit of being able to "validate" message paths (and show them red if invalid) when we know the topic's schema.

PR #6956, opened 2023-10-05
Generalized fix for sending block data to plot worker

cfoust:
**User-Facing Changes**
Fixes for a range of issues affecting plot preloading.

**Description**
This PR does the following:
* Breaks out the logic we use to decide when to send block data to the plot worker.
* Adds a test suite for it.

This is a complicated state machine with a few nasty edge cases. The existing logic did not handle them very well and was hard to reason about. To learn more about why this is hard, read [this comment](https://github.com/foxglove/studio/blob/d85325f895fec865bfb6c8fb8bada542224e665d/packages/studio-base/src/panels/Plot/blocks.test.ts#L45) in the test file.

I added tests for all of the edge cases I could come up with and ensured that this PR fixes the three ([#1](https://github.com/foxglove/studio/pull/6928) [#2](https://linear.app/foxglove/issue/FG-5215/strange-plot-behavior-when-used-alongside-map-panel) [#3](https://github.com/foxglove/studio/issues/6957)) recent bugs users experienced related to this functionality.

### DIFF
--- a/packages/studio-base/src/panels/Plot/blocks.test.ts
+++ b/packages/studio-base/src/panels/Plot/blocks.test.ts
@@ -42,12 +42,56 @@ describe("refreshBlockTopics", () => {
   });
 });
 
+// This functionality is far more complex than it appears at first glance and
+// it is worth documenting why this is the case here rather than in some
+// external document that will probably not be updated.
+//
+// I came up with a concise way of describing this problem to make the state
+// machine's behavior easier to understand. Message blocks can be represented
+// as a 1D array of cells, each of which can be in one of four states:
+// s = same
+// c = changed
+// n = new
+// e = empty
+//
+// Blocks are accumulated and sent using a cursor system, here represented by
+// the pipe character ("|"). This indicates that all of the data prior to the
+// cursor has been sent.
+//
+// For example:
+// ssss|nnneee
+// |    |  |
+// |    |  empty data that has not been loaded
+// |    new data
+// unchanged
+//
+// processBlocks does the following:
+// * moves the cursor forward according to a set of rules and returns the new
+//   data as a set of bundles that can be forwarded to the plot worker
+// * indicates whether the existing data already sent should be cleared
+// * stores the first message of each block that was sent so we can detect
+//   changed blocks
+//
+// In this case, processBlocks would return a new state that looks like this:
+// sssssss|eee
+//
+// In instances where the data should be cleared, we append an asterisk ("*")
+// to the end of the state string.
+//
+// Some examples:
+// ssss|nnneee -> sssssss|eee (data accumulating normally)
+// s|nene -> sses|e (skipping empty data)
+// cs|nnneee -> s|snnneee* (a block was changed before the cursor)
+// sss|ccsss -> sssss|sss (a block was changed after the cursor)
+//
+// We use this system for describing each test case.
 describe("processBlocks", () => {
   const subscriptions: SubscribePayload[] = [createSubscription(FAKE_TOPIC)];
   const initial = refreshBlockTopics(subscriptions, initBlockState());
   const block = createBlock(1);
 
   it("should send data as it arrives", () => {
+    // |ne -> s|e
     const first = processBlocks([block, {}], subscriptions, initial);
     {
       const {
@@ -61,6 +105,7 @@ describe("processBlocks", () => {
       expect(newData).toEqual([block]);
     }
 
+    // s|n -> ss|
     const second = processBlocks([block, block], subscriptions, first.state);
     {
       const {
@@ -76,6 +121,7 @@ describe("processBlocks", () => {
   });
 
   it("should skip empty blocks", () => {
+    // |nene -> ses|e
     const {
       state: { messages, cursors },
       resetTopics,
@@ -91,9 +137,11 @@ describe("processBlocks", () => {
     const newBlock = createBlock(2);
 
     // we have loaded a full range of data
+    // |nnn -> sss|
     const before = processBlocks([block, block, block], subscriptions, initial);
 
     // change some of it
+    // ccs| -> ss|s
     const first = processBlocks([newBlock, newBlock, block], subscriptions, before.state);
     {
       const {
@@ -107,6 +155,7 @@ describe("processBlocks", () => {
       expect(newData).toEqual([newBlock, newBlock]);
     }
 
+    // ss|c -> sss|
     const second = processBlocks([newBlock, newBlock, newBlock], subscriptions, first.state);
     {
       const {
@@ -124,8 +173,10 @@ describe("processBlocks", () => {
   it("should resend all data up to and including change if there is a change in the middle", () => {
     const newBlock = createBlock(2);
 
+    // |nnn -> sss|
     const before = processBlocks([block, block, block], subscriptions, initial);
 
+    // scs| -> ss|s*
     const first = processBlocks([block, newBlock, block], subscriptions, before.state);
     {
       const {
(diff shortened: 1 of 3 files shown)


cfoust:
Still investigating a couple of bugs (one seems unrelated, one probably is) but this might be worth an initial look.

cfoust:
Scratch that. We're still clearing out the cache too aggressively; I'm investigating.

cfoust:
I verified that all of the failing stories were still rendered correctly locally.

packages/studio-base/src/panels/Plot/useDatasets.ts
```diff
@@ -254,57 +237,24 @@ function useData(id: string, params: PlotParams) {
...
-      void service?.addBlock(messages, Array.from(resetData));
+    const {
+      state: newState,
+      resetTopics,
+      newData,
+    } = processBlocks(blocks, blockSubscriptions, blockState);
+
+    blockState = newState;
+
+    void service?.addBlock(
```
cfoust, line 248: This is a hack so we can avoid changing the plot worker's API until after #6926 lands. After it does, we can make a `clearBlocks` function that takes a topic and clears plots that use it rather than just using the `resetTopics` parameter of `addBlocks`.

packages/studio-base/src/panels/Plot/blocks.test.ts
```diff
@@ -0,0 +1,208 @@
...
+// I came up with a concise way of describing this problem to make the state
+// machine's behavior easier to understand. Message blocks can be represented
+// as a 1D array of cells, each of which can be in one of four states:
+// s = same
+// c = changed
+// n = new
+// e = empty
+//
+// Blocks are accumulated and sent using a cursor system, here represented by
+// the pipe character ("|"). This indicates that all of the data prior to the
```
snosenzo, line 58: I see this, but I also see examples below with `n`'s and `s`'s after the cursor. Why wouldn't the cursor consume these? Shouldn't the cursor always push to the end or the latest empty block?

packages/studio-base/src/panels/Plot/blocks.test.ts
```diff
@@ -0,0 +1,208 @@
...
+// sssssss|eee
+//
+// In instances where the data should be cleared, we append an asterisk ("*")
+// to the end of the state string.
+//
+// Some examples:
+// ssss|nnneee -> sssssss|eee (data accumulating normally)
+// s|nene -> sses|e (skipping empty data)
+// cs|nnneee -> s|snnneee* (a block was changed before the cursor)
+// sss|ccsss -> sssss|sss (a block was changed after the cursor)
```
snosenzo, line 85: When would a case like this ever happen? For the user script case, wouldn't all of the blocks be changed or none of them?

cfoust: yup, it happens regularly, both when a user script changes and when slicing does. the blocks are not all changed at once.

packages/studio-base/src/panels/Plot/blocks.ts
```diff
@@ -0,0 +1,175 @@
...
+import { MessageBlock } from "@foxglove/studio-base/PanelAPI/useBlocksSubscriptions";
+import { SubscribePayload } from "@foxglove/studio-base/players/types";
+
+import { Messages } from "./internalTypes";
+
+// We need to keep track of the block data we've already sent to the worker and
+// detect when it has changed, which can happen when the user changes a user
+// script or they trigger a subscription to different fields.
+// mapping from topic -> the first message on that topic in the block
+type FirstMessages = Record<string, unknown>;
```
snosenzo, line 16: `FirstMessagesByTopic` would let you remove these `topic->  ` clarifying comments, same for `Cursors`

packages/studio-base/src/panels/Plot/blocks.test.ts
cfoust: Any `n` after the cursor will be consumed regardless. Since we only consume up to and including the last changed block in the event that changes occur after the cursor, it is not desirable for us to consume all `s`.

The reason for this is for the case where a block is updated before the cursor, but blocks after the changed block are untouched. This occurs when a user script is nondeterministic (i.e. what occurred in issue [#1](https://github.com/foxglove/studio/pull/6928)). We can't know whether blocks will continue to change after the changed blocks, so we have to wait until the next iteration to be sure.

To illustrate:
```
sscs| -> sss|s*
sss|s -> ssss|
```

packages/studio-base/src/panels/Plot/blocks.test.ts
snosenzo: What do we gain by using cursors then if changes can happen anywhere in the array? Why not just use array status diffs for blocks if they need to be iterated through anyways?

packages/studio-base/src/panels/Plot/blocks.test.ts
snosenzo: seems strange to me that the cursor position doesn't always represent that data before the cursor has already be sent. Why can't/shouldn't it resolve that over one iteration?

cfoust: The cursor position always represents the data that has already been sent. That's not violated anywhere. In the case I described above, we cannot know whether data will change following the cursor, so we don't consume it until the next iteration.

packages/studio-base/src/panels/Plot/blocks.test.ts
cfoust: We need to keep track of the data we've already sent to handle (a) changes that occur before the cursor and (b) data that is still streaming in.

packages/studio-base/src/panels/Plot/blocks.ts
cfoust: Hmm, perhaps--personally I prefer specifying the mapping separately for associative arrays to keep identifiers short.

snosenzo: The short identifiers are sacrificing readability and incur more mental overhead when reviewing imo

cfoust: I see why this is important to you, but everyone brings a different perspective on naming. Unless a name is blatantly incorrect or misleading, I think we should concentrate our efforts on examining the correctness of the methodology of PRs, not their idiosyncrasies

packages/studio-base/src/panels/Plot/blocks.test.ts
snosenzo: What triggers these iterations? Is it safe to assume that there will always be another iteration that will consume the whole array?

cfoust: Any reference change to `blocks` returned [here](https://github.com/foxglove/studio/pull/6956/files#diff-7372e3a2054c7c5fd4edb8e904c174c725b3ed43c034938df26f5a442918c881R218) will trigger an iteration. In practice the player (or more specifically, the hook) will always bump the reference if data is being added or changed.

packages/studio-base/src/panels/Plot/blocks.test.ts
snosenzo: I think something that wasn't really clear to me before reading this was that we can only really add or reset blocks as an outcome to this process. This context is important to know for why we're using a cursor. Those requirements aren't really said anywhere.

snosenzo (approved):
would be good if @jtbandes to give a look over too. tested and works.

I still think that the cursor behavior that assumes there will be another block update trigger after not consuming the entire array of blocks in this example (see below) is a bit fragile and over-complex, but didn't run into any related issues testing against user scripts. 
```
sscs| -> sss|s*
sss|s -> ssss|
```

PR #6960, opened 2023-10-06
Build incoming plot data in the rendering thread

cfoust:
**User-Facing Changes**
Greatly reduce memory usage for plots.

**Description**
Still WIP.


cfoust:
this doesn't actually work and we shouldn't pursue it--it's too slow on big datasets

PR #6968, opened 2023-10-09
Fix data platform topics sometimes not being streamed

achim-k:
**User-Facing Changes**
Fix data platform topics sometimes not being streamed

**Description**
Fixes that a subscription change is ignored when the iterable player is in the `start-play` state and when there is no [seek target](https://github.com/foxglove/studio/blob/c648303a3d017a7fa32afcc278bb145c5692f641/packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts#L596-L600). This mostly happened when loading a recording from data platform,  as the `start-play` state seems to take longer compared to when loading a local recording (which also uses the iterable player under the hood)

Resolves FG-5159

### DIFF
--- a/packages/studio-base/src/players/IterablePlayer/IterablePlayer.test.ts
+++ b/packages/studio-base/src/players/IterablePlayer/IterablePlayer.test.ts
@@ -805,4 +805,41 @@ describe("IterablePlayer", () => {
     player.close();
     await player.isClosed;
   });
+
+  it("should allow changing subscriptions when player in start-play state", async () => {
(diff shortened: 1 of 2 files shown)


cfoust (approved):

cfoust:
disregard my approval, thought it was a critical fix and it looked fine to me but it seems Jacob has some concerns!

jtbandes (approved):
Seems like we should be able to have a unit test for this. Is that possible?

jtbandes (approved):

PR #6969, opened 2023-10-09
Add wider support for image downloading

snosenzo:
**User-Facing Changes**

**Description**
Update download image feature to use already-decoded image instead of image message itself. This simplifies download image and allows support for downloading any form of decoded image bitmap or image data. 

Resolves: FG-4138


achim-k (approved):
Tested it, worked nicely :+1:

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageMode.ts
```diff
@@ -849,48 +844,28 @@ export class ImageMode
     this.updateSettingsTree();
   };
 
-  protected getLatestImage(): DownloadImageInfo | undefined {
```
jtbandes, line 852: I think this `DownloadImageInfo` type can be removed now

jtbandes (approved):
Don't remember why I didn't do it this way in the first place...

PR #6972, opened 2023-10-10
<fix>: fix cancel dragged panel

smbera:
**User-Facing Changes**

**Description**
When the user wants to cancel a dragged panel, he will let go at an invalid place, but there is a bug here, and a new panel will still be added that the user just canceled.

**Before**

https://github.com/foxglove/studio/assets/23401125/faa8f111-86e4-4e3e-a5f0-89329c51027b

**After**

https://github.com/foxglove/studio/assets/23401125/709cd715-8900-44a8-8ec9-cc409898f481

### DIFF
--- a/packages/studio-base/src/components/PanelCatalog/PanelListItem.tsx
+++ b/packages/studio-base/src/components/PanelCatalog/PanelListItem.tsx
@@ -76,10 +76,10 @@ export function PanelListItem(props: Props): JSX.Element {
     },
     options: { dropEffect: "copy" },
     end: (_item, monitor) => {
-      const dropResult = monitor.getDropResult() ?? {};
+      const dropResult = monitor.getDropResult();
 
       // do nothing when the user wants to cancel a dragged panel
-      if (_.isEmpty(dropResult)) {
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/components/PanelCatalog/PanelListItem.tsx
```diff
@@ -76,6 +77,12 @@ export function PanelListItem(props: Props): JSX.Element {
     options: { dropEffect: "copy" },
     end: (_item, monitor) => {
       const dropResult = monitor.getDropResult() ?? {};
+
+      // do nothing when the user wants to cancel a dragged panel
+      if (_.isEmpty(dropResult)) {
```
jtbandes, line 82: Maybe we could just remove this `??` and use `dropResult == undefined` below?

```suggestion
      const dropResult = monitor.getDropResult();

      // do nothing when the user wants to cancel a dragged panel
      if (dropResult == undefined) {
```

jtbandes:
Thanks, I merged the changes in #6975

PR #6989, opened 2023-10-13
Fix bug with subscription fields

cfoust:
**User-Facing Changes**
Fix an issue where plotting a signal used by a map panel caused the map panel to fail.

**Description**
An `SubscriptionPayload` with `preloadType="full"` but no `fields` property implies a subscription to `"partial"` as well. We were not handling this correctly, which meant that we were dropping subscriptions to all message fields under certain circumstances.

### DIFF
--- a/packages/studio-base/src/components/MessagePipeline/subscriptions.test.ts
+++ b/packages/studio-base/src/components/MessagePipeline/subscriptions.test.ts
@@ -69,7 +69,6 @@ describe("mergeSubscriptions", () => {
     const subs: SubscribePayload[] = [
       { topic: "a", preloadType: "partial", fields: ["one", "two"] },
       { topic: "a", preloadType: "full" },
-      { topic: "a", preloadType: "partial" },
     ];
 
     const result = mergeSubscriptions(subs);
--- a/packages/studio-base/src/components/MessagePipeline/subscriptions.test.ts
+++ b/packages/studio-base/src/components/MessagePipeline/subscriptions.test.ts
@@ -64,4 +64,19 @@ describe("mergeSubscriptions", () => {
       { topic: "a", preloadType: "partial" },
     ]);
   });
+
+  it("switches to subscribing to all fields across preloadType", () => {
+    const subs: SubscribePayload[] = [
+      { topic: "a", preloadType: "partial", fields: ["one", "two"] },
+      { topic: "a", preloadType: "full" },
+      { topic: "a", preloadType: "partial" },
+    ];
+
+    const result = mergeSubscriptions(subs);
+
+    expect(result).toEqual([
+      { topic: "a", preloadType: "full" },
+      { topic: "a", preloadType: "partial" },
+    ]);
+  });
 });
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/components/PanelExtensionAdapter/PanelExtensionAdapter.tsx
```diff
@@ -402,7 +402,6 @@ function PanelExtensionAdapter(
 
           return {
             topic: item.topic,
-            convertTo: item.convertTo,
```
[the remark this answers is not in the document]
jtbandes: Really weird that TS didn't catch this error! I notice it catches it if you change the map function to be `map((item): SubscribePayload => {`, but it does not catch it when using `map<SubscribePayload>((item) => {`.... kinda seems like a TS bug


...did some digging and found this:
- https://github.com/microsoft/TypeScript/issues/45979

Which was closed as a duplicate of:
- https://github.com/microsoft/TypeScript/issues/241

So maybe there's hope of it being fixed some day 😐

packages/studio-base/src/components/MessagePipeline/subscriptions.test.ts
```diff
@@ -64,4 +64,19 @@ describe("mergeSubscriptions", () => {
...
+      { topic: "a", preloadType: "partial" },
+    ];
+
+    const result = mergeSubscriptions(subs);
+
+    expect(result).toEqual([
+      { topic: "a", preloadType: "full" },
+      { topic: "a", preloadType: "partial" },
+    ]);
+  });
```
jtbandes, line 81: I copied & pasted this test into my working copy (on main) and it passes. Could you update the test to include a case that fails without your fix?

packages/studio-base/src/components/MessagePipeline/subscriptions.ts
```diff
@@ -86,6 +86,16 @@ export function mergeSubscriptions(
...
 ): Immutable<SubscribePayload[]> {
   return R.pipe(
+    R.chain((v: Immutable<SubscribePayload>): Immutable<SubscribePayload>[] => {
+      const { preloadType } = v;
+      if (preloadType !== "full") {
+        return [v];
+      }
+
+      // a "full" subscription to all fields implies a "partial" subscription
+      // to those fields, too
```
jtbandes, line 96: not sure if I understand this premise — can't subscribers sometimes use full subscriptions without also using partial subscriptions? Are you saying there is something special that makes this true when using `fields` but not otherwise?

packages/studio-base/src/components/PanelExtensionAdapter/PanelExtensionAdapter.tsx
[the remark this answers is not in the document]
jtbandes: => https://github.com/foxglove/studio/pull/6995

packages/studio-base/src/components/MessagePipeline/subscriptions.ts
snosenzo: I think it's just trying to say that preloadType: full subscriptions also cause messages to appear in currentframe messages (as well as block subscriptions)

snosenzo: The `IterablePlayer` uses `allTopics` to read messages from the source, which includes preload topics.

cfoust:
Added a test that fails in main but succeeds in this PR.
Снимок экрана 2023-10-16 в 9 59 12 AM

jtbandes (approved):

PR #6992, opened 2023-10-13
Select a default sidebar item if the selected one is not present

jtbandes:
**User-Facing Changes**

**Description**
Some sidebar items (such as "Studio Logs" / "Events") are conditionally included in the sidebar. If one was selected but no longer present then the sidebar could not be expanded. Now we auto-select one of the available items instead.

Fixes FG-5315

### DIFF
--- a/packages/studio-base/src/components/Sidebars/Sidebars.tsx
+++ b/packages/studio-base/src/components/Sidebars/Sidebars.tsx
@@ -101,16 +101,10 @@ export function Sidebars<LeftKey extends string, RightKey extends string>(
   // Select an available item if the selected one is not actually available
   useEffect(() => {
     if (leftSidebarOpen && !leftItems.has(selectedLeftKey)) {
-      const anyLeftKey = [...leftItems.keys()][0];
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/components/Sidebars/Sidebars.tsx
```diff
@@ -95,8 +95,33 @@ export function Sidebars<LeftKey extends string, RightKey extends string>(
...
+      if (anyLeftKey != undefined && anyLeftKey !== selectedLeftKey) {
+        onSelectLeftKey(anyLeftKey);
+      }
+    }
+    if (rightSidebarOpen && !rightItems.has(selectedRightKey)) {
+      const anyRightKey = [...rightItems.keys()][0];
+      if (anyRightKey != undefined && anyRightKey !== selectedRightKey) {
+        onSelectRightKey(anyRightKey);
+      }
+    }
```
snosenzo, line 114: What happens when `anyLeft/RightKey` is still undefined? 
should we call `onSelectLefKey(undefined)` if `anyLeftKey` is undefined?

snosenzo (approved):
wondering about an edge case but otherwise looks good

PR #6993, opened 2023-10-13
Fix point cloud messagetime

snosenzo:
**User-Facing Changes**

- n/a

**Description**
Point cloud message time was being assigned to the receivetime instead of the headerstamp.


jtbandes (approved):
What was the impact of this?

snosenzo:
@jtbandes 
> What was the impact of this?

When getting the pose of a pointcloud we look at messageTime to grab the transform state at that time. This should be the headerstamp and not the receiveTime

jtbandes:
Is that something we could have a test for then?

snosenzo:
@jtbandes I think this is something that might be better enforced with types. i.e. using different types for headerstamp and receivetime and only explicitly widening in certain places.

PR #6997, opened 2023-10-17
New downsampling algorithm for line plots

cfoust:
**User-Facing Changes**
Implement a new downsampling algorithm for plots and optimize its usage.

**Description**
WIP


amacneil:
- deleting the last series doesn't clear the plot
- sometimes the downsampling seems to just stop half way through and leave things in a broken state. I think it as something to do with either being seeked to a non-default position, or having the plot viewport be a non-default position

image

packages/studio-base/src/panels/Plot/processor/downsample.ts
```diff
@@ -0,0 +1,677 @@
...
+      return [];
+    }
+    return [state];
+  }, paths);
+  if (shouldResetViewport(pathStates, downsampledView, view, previousBounds)) {
+    return updateDownsample(view, blocks, current, initDownsampled());
+  }
+
+  const numDatasets = Math.max(blocks.datasets.size, current.datasets.size);
+  // We don't have any data
```
defunctzombie, line 625: comment is not useful - unless you say _why_ having no data means we need to return `initDownsampled`

defunctzombie:
A meta comment on this note before diving deeper into the PR:

> Try to bear in mind before commenting that a lot of thought went into every line of this PR, and often there is a good explanation for why it looks the way it does, even if that explanation may not yet be represented in the code itself. 

To me this reads as both an attempt to ward-off actual comments and a pass on adding better comments to the code. I'm giving you the benefit of doubt on this but recommend against such statement in the future - a lot of code we all write has a good explanation - and if that is not obvious from the code, and we become aware of it, the correct action is to comment the code. If you already know of such parts in the code please point them out. Otherwise what this PR does is create some untouchable set of code that we can never know why parts were written the way they were or change them for fear of breaking a meticulously crafted file.

packages/studio-base/src/components/TimeBasedChart/lttb.ts
```diff
@@ -0,0 +1,146 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
defunctzombie, line 1: This file (by name) seems to be a math algorithm implementation yet the file has no tests. How do we know that it does what it says it does? How will we feel confident changing it or fixing bugs specific to this implementation? Please add tests for this that show it working in the ways you say it should work in this PR (limits on output data) and other notable edge cases you have discovered.

packages/studio-base/src/components/Chart/datasets.ts
```diff
@@ -135,3 +134,57 @@ export function findIndices(
 
   return undefined;
 }
+
+/**
+ * lookupIndices returns a faster version of findIndices in exchange for doing
+ * some compute ahead of time.
+ */
+export const lookupIndices = (dataset: TypedData[]): ((offset: number) => Indices | undefined) => {
+  const offsets: number[] = R.pipe(
```
defunctzombie, line 143: I find this style of code extremely hard to read without any sort of comments

packages/studio-base/src/components/Chart/datasets.ts
```diff
@@ -135,3 +134,57 @@ export function findIndices(
...
+    R.map(({ x: { length } }: TypedData) => length),
+    R.reduce(
+      (lengths: number[], length: number): number[] => [
+        ...lengths,
+        (R.last(lengths) ?? 0) + length,
+      ],
+      [],
+    ),
+    // remove the last one (can't resolve to greater than end of dataset)
+    (v) => v.slice(0, -1),
```
defunctzombie, line 153: What is `v`? Using a longer variable name here could make this easier to read

packages/studio-base/src/panels/Plot/datasets.tsx
```diff
@@ -346,8 +347,8 @@ function sliceSingle(slice: TypedData, start: number, end?: number): TypedData {
 
   return {
     receiveTime,
-    headerStamp: headerStamp.length > 0 ? headerStamp : undefined,
-    constantName: constantName.length > 0 ? constantName : undefined,
+    ...(headerStamp.length > 0 ? { headerStamp } : {}),
```
defunctzombie, line 350: What's this change for?

packages/studio-base/src/panels/Plot/plotData.ts
```diff
@@ -185,6 +185,22 @@ export function reducePlotData(data: PlotData[]): PlotData {
   return reduced;
 }
 
+export function mergeAllData(blockData: PlotData, currentData: PlotData): PlotData {
```
defunctzombie, line 188: function needs a comment

packages/studio-base/src/panels/Plot/processor/downsample.ts
```diff
@@ -0,0 +1,677 @@
...
+      }
+      return [bounds];
+    }),
+    combineBounds,
+  )([...datasets.values()]);
+};
+
+const getBoundsRange = ({ max, min }: Bounds1D): number => Math.abs(max - min);
+
+const MAX_POINTS = 3_000;
```
defunctzombie, line 146: Is this constant important? Seems important yet it is buried well into this file and has no comment. Please add a comment and consider surfacing this constant higher - possibly make it something a caller can provide as an argument but I feel less strongly about that in this first version than I do about commenting and discovery.

packages/studio-base/src/panels/Plot/processor/downsample.ts
```diff
@@ -0,0 +1,677 @@
...
+
+/**
+ * Get the visual scale of the `PlotViewport`, or the ratio of a point in
```
defunctzombie, line 149: I don't follow this comment - does this get the visual scale or does it get the ratio of a point in viewport space to pixels? Is the `or` a typo and the second part of this sentence is describing the visual scale?

packages/studio-base/src/panels/Plot/processor/downsample.ts
```diff
@@ -0,0 +1,677 @@
...
+ * viewport space to pixels.
+ */
+const getScale = ({ width, height, bounds: { x, y } }: PlotViewport): { x: number; y: number } => ({
+  x: (x.max - x.min) / width,
+  y: (y.max - y.min) / height,
+});
+
+const ZOOM_THRESHOLD_PERCENT = 0.2;
```
defunctzombie, line 157: What is this constant? Please add a comment and consider surfacing earlier in the file.

packages/studio-base/src/panels/Plot/processor/downsample.ts
```diff
@@ -0,0 +1,677 @@
...
+  R.pipe(
+    isDerivative(path) ? derivative : noop,
+    isHeaderStamp(path) ? sortDataByHeaderStamp : noop,
+  )(data);
+
+/**
+ * updateSource processes new points for one path and one source (either block
+ * or current data), doing a partial downsample of any new points.
+ */
+export function updateSource(
```
defunctzombie, line 246: Personally I find functions with more than two arguments pretty unwieldy at the call site (as an example several call sites use a raw number for the `minSize` argument and reading it I have no idea what that argument is). I know this is sometimes done for performance but in the places I see this called it doesn't seem that this is critical here (given all the other object and array handling that all the code does). I would have turned this into an args object but do not feel that is a strong requirement to change in this PR but offer my thoughts for consideration.

packages/studio-base/src/panels/Plot/processor/downsample.ts
```diff
@@ -0,0 +1,677 @@
...
+  }
+  // the input data regressed for some reason, handle this gracefully
+  if (newCursor < oldCursor) {
+    return updateSource(path, raw, view, viewBounds, maxPoints, minSize, initSource());
+  }
+  if (newCursor === oldCursor) {
+    return state;
+  }
+
+  // it's game over, these do not yet have optimizations
```
defunctzombie, line 273: Could you elaborate on why this is "game over" and what the optimizations are that these don't have? Maybe some notes on why its hard or what might need to be done while your mind is fresh with the info. Feel free to link to the linear issue as a shortcut since I think you put some thoughts there. It will provide a good path towards remembering what was hard.

packages/studio-base/src/panels/Plot/processor/downsample.ts
```diff
@@ -0,0 +1,677 @@
...
+    return {
+      ...initSource(),
+      dataset: {
+        ...raw,
+        data: downsampled,
+      },
+    };
+  }
+
+  // same with this, but for different reasons
```
defunctzombie, line 288: What are those reasons? We are gonna have a hard time reading your mind when reading the code. Please elaborate while the ideas are fresh.

packages/studio-base/src/panels/Plot/processor/downsample.ts
```diff
@@ -0,0 +1,677 @@
...
+    isPartial: true,
+    dataset: {
+      ...blockData,
+      pointRadius: 0,
+      data: downsampled,
+    },
+  };
+}
+
+export function updatePath(
```
defunctzombie, line 477: Function is missing a comment. Also suffers from the above described many input arguments affliction.

packages/studio-base/src/panels/Plot/processor/downsample.ts
```diff
@@ -0,0 +1,677 @@
...
+    newBounds.min <= dataBounds.min &&
+    newBounds.max >= dataBounds.max
+  ) {
+    return false;
+  }
+
+  return didZoom;
+}
+
+export function updateDownsample(
```
defunctzombie, line 596: Function is missing a comment.

packages/studio-base/src/panels/Plot/datasets.tsx
```diff
@@ -390,18 +391,18 @@ export function sliceTyped(dataset: TypedData[], start: number, end?: number): T
   return [sliceSingle(startSlice, offset0), ...between, sliceSingle(endSlice, 0, offset1)];
 }
 
-export function getXBounds(dataset: TypedData[]): [min: number, max: number] | undefined {
+export function getXBounds(dataset: TypedData[]): Bounds1D | undefined {
```
defunctzombie, line 394: Was an object with named fields easier to use than the tuple or no difference just happened to be a type change to use a named type?

packages/studio-base/src/components/TimeBasedChart/lttb.ts
cfoust: I don't think it's in our best interest (nor necessarily useful) to write tests for an algorithm we implemented from a paper unless we have specific needs for its behavior. In this case we really just want an algorithm that, given N points and M buckets, returns a list of indices strictly equal to `min(N, M)`. We care to the extent that the choices the algorithm makes about which points to include produces a decent approximation of the dataset, but not the particulars of how it gets there.

To illustrate, consider what writing a test suite would even involve: presumably, doing the original algorithm by hand and comparing the output to that of our own implementation. It's hard to imagine that that brings us any confidence.

packages/studio-base/src/panels/Plot/datasets.tsx
cfoust: Including the key causes issues downstream in the way we proxy dataset access to ChartJS. I can explain this in more detail over Slack if you'd like, it's a bit nuanced and annoying, but this was causing runtime errors.

packages/studio-base/src/panels/Plot/processor/downsample.ts
cfoust: This is a good call-out, you're completely right. Having lots of parameters makes it hard to know what's happening in each of these calls.

packages/studio-base/src/panels/Plot/processor/downsample.ts
cfoust: 👍

packages/studio-base/src/panels/Plot/datasets.tsx
cfoust: Making it an object with named fields allowed me to unify this with the existing `Bounds` type. It felt weird that we had two different types representing the same thing in files with similar functionality.

cfoust:
I have fixed both of @amacneil 's bugs and (as far as I can discern) replied to or resolved all comments.

packages/studio-base/src/panels/Plot/datasets.tsx
defunctzombie: If its nuanced and annoying its worth calling out as a code comment because Slack won't be available when we come back to this code.

packages/studio-base/src/components/TimeBasedChart/lttb.ts
defunctzombie: I disagree with this reasoning. I think it is definitely in our interest to have tests on a math algorithm that we expect to produce specific results. Good things to test for might be that all outputs existed in the input. Maybe that it limited the number of points as expected by the interface. The test doesn't have to dictate the algorithm but should be a valuable check on assumptions and help us evolve the code. If we discover bugs in our implementation of the algorithm we would want to capture those with tests so we don't regress. Additionally, tests allow us to evolve the implementation while knowing we adhered to the interface and output expectations or create a trip-wire if assumptions are changing.

cfoust: It actually seems that we agree. You're saying that we could test the properties of the output but not the implementation. We do not expect specific results; we expect the property that I described to hold and that's all. Unless you're suggesting we do something more granular than that?

defunctzombie: I don't see any test files that show this property tested (unless the tests are not in an lttb.test.ts file)? Do we have tests for these functions showing that outputs only contain data present in the input?

packages/studio-base/src/panels/Plot/datasets.tsx
cfoust: We do this throughout this file; where do you think the best place to document this would be?

packages/studio-base/src/components/TimeBasedChart/lttb.ts
cfoust: I added [a few tests](https://github.com/foxglove/studio/pull/6997/commits/a1c1b7426a7695cd8fad64d80b01e80a6c594ed9) that verify the properties I describe. I want to be clear that they tell us very little about whether the algorithm is working correctly--someone could happily come in and hack away without breaking them.

My thinking is that every test we write should have a confident answer to the question "what behavior should this prevent?" In other words, they should be a living record of the constraints and assumptions of our system and allow us to catch accidental deviations.

But the choice of a downsampling algorithm is an aesthetic choice--the rest of our system does not depend on the output _beyond the properties described above._ Put differently, we don't care that the algorithm chooses point `i` vs `i+1`.

jtbandes: It does seem that the paper describes at least one way in which we do care about the algorithm choosing specific points:

> The first and last buckets however contain only the first and last data points of the original data as shown in figure 2.3 on page 8. This is to ensure that those points will be included in the downsampled data.


That seems worth testing for, to ensure the implementation is correct.

I think we could also include a test case with exactly 3 buckets where the number of points is quite small and we can actually assert that a specific point is supposed to be chosen in the middle bucket. Something like this:

image

cfoust: My broader point was that we don't actually care if the implementation is 100% identical to the paper and/or makes those specific point choices. A downsampling algorithm is "correct" when it appeals to our aesthetic preferences, and that's about it, it can just be a black box beyond that

packages/studio-base/src/components/TimeBasedChart/lttb.ts
```diff
@@ -0,0 +1,146 @@
...
+    maxIndex = -1;
+    maxArea = 0;
+    // Choose the triangle with the maximum area
+    for (let i = bucketStart; i < bucketEnd; i++) {
+      if (xBuffer == undefined || yBuffer == undefined) {
+        return undefined;
+      }
+      x = xBuffer[sliceOffset]!;
+      y = yBuffer[sliceOffset]!;
+      const area = Math.abs((prevX - avgX) * (y - prevY) - (prevX - x) * (avgY - prevY)) * 0.5;
```
jtbandes, line 126: can we drop the `* 0.5`? it seems like it would not affect the results of the algorithm and probably helps to shave a little bit off this hot path

packages/studio-base/src/components/TimeBasedChart/lttb.ts
```diff
@@ -0,0 +1,146 @@
...
+      sliceOffset = 0;
+      sliceLength = data[sliceIndex]?.x.length;
+      xBuffer = data[sliceIndex]?.x;
+      yBuffer = data[sliceIndex]?.y;
+    }
+  };
+
+  let prevX: number = 0;
+  let prevY: number = 0;
+  let points: number[] = [0];
```
jtbandes, line 70: Are these indices into the input data? if so, might help to name the variable `indices` or `pointIndices`. I think the function's doc comment should also be updated to describe what the parameters do and what it returns.

packages/studio-base/src/components/TimeBasedChart/lttb.ts
cfoust: ooo yeah, you're right, that should not change the outcome

packages/studio-base/src/components/Chart/datasets.ts
```diff
@@ -135,3 +134,67 @@ export function findIndices(
 
   return undefined;
 }
+
+/**
+ * lookupIndices returns a faster version of findIndices in exchange for doing
+ * some compute ahead of time.
+ */
+export const lookupIndices = (dataset: TypedData[]): ((index: number) => Indices | undefined) => {
```
jtbandes, line 142: The names of these two functions are similar but different in a way that makes me think their functionality is probably different – not just their performance. I would recommend naming them something more similar and obviously related, e.g. `createOptimizedFindIndices()`

packages/studio-base/src/panels/Plot/processor/downsample.ts
```diff
@@ -0,0 +1,757 @@
...
+  sliceTyped,
+  resolveTypedIndices,
+} from "../datasets";
+import { PlotPath, DatasetsByPath, TypedDataSet, TypedData } from "../internalTypes";
+import { EmptyPlotData, PlotData, sortDataByHeaderStamp } from "../plotData";
+
+type PathMap<T> = Map<PlotPath, T>;
+
+type SourceState = {
+  cursor: number;
```
jtbandes, line 33: these fields could use comments – what does the cursor indicate? And in what sense is this a "source" (is that related to players)?

packages/studio-base/src/panels/Plot/processor/downsample.ts
```diff
@@ -0,0 +1,757 @@
...
+};
+type PathState = {
+  blocks: SourceState;
+  current: SourceState;
+  dataset: TypedDataSet | undefined;
+  isPartial: boolean;
+};
+
+export const initSource = (): SourceState => ({
+  cursor: 0,
```
jtbandes, line 46: nit: is a 0 cursor invalid on an empty dataset, or would it work to use something like -1 or NaN here to avoid potentially confusing it with a valid 0 value?

packages/studio-base/src/panels/Plot/processor/downsample.ts
```diff
@@ -0,0 +1,757 @@
...
+    isValid: false,
+    view: undefined,
+    paths: new Map(),
+    data: EmptyPlotData,
+  };
+};
+
+// This is the desired number of data points for each plot across all signals
+// and data sources. Beyond this threshold, ChartJS can no longer render at
+// 60FPS.
```
jtbandes, line 78: …on a 2023 MBP 🤣

packages/studio-base/src/panels/Plot/processor/downsample.ts
```diff
@@ -0,0 +1,757 @@
...
+
+export const initPath = (): PathState => ({
+  blocks: initSource(),
+  current: initSource(),
+  dataset: undefined,
+  isPartial: false,
+});
+
+export type Downsampled = {
+  isValid: boolean;
```
jtbandes, line 60: why would downsampled data not be valid? could use comment

packages/studio-base/src/panels/Plot/processor/downsample.ts
```diff
@@ -0,0 +1,757 @@
...
+  return [x, y];
+};
+
+const combineBounds = (bounds: Bounds1D[]): Bounds1D | undefined => {
+  if (bounds.length === 0) {
+    return undefined;
+  }
+
+  let min = Number.MAX_SAFE_INTEGER;
+  let max = Number.MIN_SAFE_INTEGER;
```
jtbandes, line 136: Not sure if there is was a reason to use these specific values – but just a nit that I prefer to use -Infinity & +Infinity for things like this because they're more clearly invalid numbers, whereas MAX_SAFE_INTEGER is actually a finite number that _might_ get confused with a valid value

packages/studio-base/src/panels/Plot/processor/downsample.ts
```diff
@@ -0,0 +1,757 @@
...
+ * occupies.
+ */
+const getScale = ({ width, height, bounds: { x, y } }: PlotViewport): { x: number; y: number } => ({
+  x: (x.max - x.min) / width,
+  y: (y.max - y.min) / height,
+});
+
+const isPartialState = (state: PathState) => state.isPartial;
+
+const findPointBinary = (
```
jtbandes, line 178: Could use a comment on what the parameters are for and what the return value is. Sometimes these things are self-explanatory but when 4 out of 6 types are :just: `number`, I find it can be useful to clarify what units we're talking about.

packages/studio-base/src/panels/Plot/processor/downsample.ts
```diff
@@ -0,0 +1,757 @@
...
+};
+
+/**
+ * Get the portion of TypedData[] that falls within `bounds`.
+ */
+const sliceBounds = (data: TypedData[], bounds: Bounds1D): TypedData[] => {
+  const lookup = lookupIndices(data);
+  const length = getTypedLength(data);
+  const start = findPointBinary(lookup, data, 0, length, bounds.min);
+  const end = findPointBinary(lookup, data, 0, length, bounds.max);
```
jtbandes, line 216: more a question for my own learning: what happens if bounds.min and bounds.max do not have _exact_ matches in the `data`? maybe this could be part of the comment on `findPointBinary` – its behavior in the event there is no exact match

packages/studio-base/src/panels/Plot/processor/downsample.ts
```diff
@@ -0,0 +1,757 @@
...
+  if (start == undefined) {
+    return data;
+  }
+
+  return sliceTyped(data, start, end);
+};
+
+const getVisibleBounds = (
```
jtbandes, line 224: What does "visible" mean in this case? Based on the arguments it looks like we just have two data sets – no information about what is visible or not. Could probably use either a comment or a new name, or both

packages/studio-base/src/panels/Plot/processor/downsample.ts
```diff
@@ -0,0 +1,757 @@
...
+  const blockBounds = getXBounds(blockData.data);
+  const currentBounds = getXBounds(currentData.data);
+  if (blockBounds == undefined || currentBounds == undefined) {
+    return undefined;
+  }
+
+  return combineBounds([blockBounds, currentBounds]);
+};
+
+const isDerivative = (path: PlotPath): boolean => path.value.endsWith(".@derivative");
```
jtbandes, line 249: Not something to tackle in this PR but just a note that I feel there is some disconnect around these `.@derivative` paths – we have code to actually parse them in `parseRosPath`, but we seem to end up ignoring that and using `endsWith` in a few places instead. Maybe something for future cleanup.

https://github.com/foxglove/studio/blob/afe02317ee9a74ac906b2a4dac120a6d77a2acfa/packages/studio-base/src/components/MessagePathSyntax/parseRosPath.test.ts#L42

packages/studio-base/src/panels/Plot/processor/downsample.ts
jtbandes, line 1: Tangential thought: it would be super cool if this stuff could be architected in a way that's independent of the plot panel and TypedData types and things – and could be a separate package that we'd be able to import and maybe even publish on NPM.

packages/studio-base/src/panels/Plot/processor/downsample.ts
```diff
@@ -0,0 +1,757 @@
...
+  // machine currently cannot handle.
+  // 2. When a path uses header stamps, we must sort the data points by their
+  // header stamps before we can present them. This means that new points can
+  // appear out-of-order and must be merged correctly into an existing, sorted
+  // dataset that we hold on to, similar to what we would have to do for
+  // derivatives.
+  //
+  // Both present serious drawbacks for memory, since we would have to store an
+  // additional copy of the entire dataset with these transformations applied.
+  if (isDerivative(path) || isHeaderStamp(path)) {
```
jtbandes, line 311: wondering why these checks are not higher up in the function (e.g. before the `newCursor === oldCursor` check) – does the state machine assume that the path does not change between calls?

packages/studio-base/src/panels/Plot/processor/downsample.ts
```diff
@@ -0,0 +1,757 @@
...
+  //
+  // this is just a guess, but mostly works, and if the plot gets too dense we
+  // will downsample again anyway
+  const newRange = getBoundsRange(newBounds);
+  if (previous == undefined || chunkSize === 0) {
+    const proportion = newRange / viewportRange;
+    if (proportion < minSize) {
+      // Before we can infer a reasonable bucket size, we still want to show
+      // data, so we fall back to the previous algorithm
+      const indices = downsampleTimeseries(iterateTyped(raw.data), view);
```
jtbandes, line 364: I don't fully know the consequences of this, but it sounds like the minSize thing could potentially be replaced by better bucket size guessing – or by _not_ downsampling at all until we reach a certain number of points. How did you settle on the minSize approach?

I don't love the fact that we are keeping the old algorithm around and using it as a fallback. I would hope this new algorithm will allow us to eventually delete the old one.

packages/studio-base/src/panels/Plot/processor/downsample.ts
```diff
@@ -0,0 +1,757 @@
...
+  // in order for the downsampled signal to maintain the same visual density
+  // over the entire plot, the number of points we downsample needs to stay the
+  // same; this is what `chunkSize` does.
+  //
+  // most of the time, however, we receive new points in quantities far smaller
+  // than `chunkSize`, so to get around this (but still retain visual density)
+  // we reuse raw points that already exist in the dataset and start our
+  // downsample a few buckets _before_ the points we're adding. this is because
+  // the point we choose for each bucket depends on the point we chose in the
+  // previous bucket, and so on.
```
jtbandes, line 411: so in this sense, is it possible that the result of running the `update` logic with incrementally-loaded data might not end up being _exactly_ the same as the results you'd get by downsampling all the data in one run of the algorithm?

packages/studio-base/src/panels/Plot/processor/downsample.ts
```diff
@@ -0,0 +1,757 @@
...
+      cursor: oldCursor + chunkSize,
+      dataset: concatDataset(previous, { ...previous, data: downsampled }),
+    });
+  }
+
+  const numOldPoints = chunkSize - numNewPoints;
+  const rawStart = newCursor - chunkSize;
+  const pointsPerBucket = Math.trunc(chunkSize / numBuckets);
+  const lastRawBucket = Math.max(
+    (numOldPoints - (numOldPoints % pointsPerBucket)) / pointsPerBucket - 2,
```
jtbandes, line 436: would this give the same result?
```suggestion
    Math.floor(numOldPoints / pointsPerBucket) - 2,
```

packages/studio-base/src/panels/Plot/processor/downsample.ts
```diff
@@ -0,0 +1,757 @@
...
+  const lastPoint = getLastPoint(previous.data);
+  if (downsampled == undefined || lastPoint == undefined) {
+    return state;
+  }
+
+  const [lastX] = lastPoint;
+  let firstNew = -1;
+  for (const { index, x } of iterateTyped(downsampled)) {
+    if (x > lastX) {
+      firstNew = index;
```
jtbandes, line 449: would this be a candidate for binary search? I see you've used binary search in a handful of other places

packages/studio-base/src/panels/Plot/processor/downsample.ts
```diff
@@ -0,0 +1,757 @@
...
+  }
+
+  if (current != undefined) {
+    return current;
+  }
+
+  return undefined;
+}
+
+type PathParameters = {
```
jtbandes, line 493: nit: I think it would help readability to keep most of the type definitions toward the top of the file rather than interspersed with functions

packages/studio-base/src/panels/Plot/processor/downsample.ts
```diff
@@ -0,0 +1,757 @@
...
+  view: PlotViewport;
+  viewBounds: Bounds1D;
+  maxPoints: number;
+};
+
+/**
+ * When the dataset is only partially visible, we cut off the occluded portions
+ * and downsample what's left.
+ */
+function updatePartialView(path: PlotPath, params: PathParameters, state: PathState): PathState {
```
jtbandes, line 505: nit: I saw a lot of `const fn = (...)=>` – I was gonna say I prefer the `function` style for module-level functions, but decided not to say anything because I could just go write a lint rule if I really cared about it. But then I noticed this  `function fn(...)` here so I figured maybe I'll at least ask for consistency within one file :)

packages/studio-base/src/panels/Plot/processor/downsample.ts
cfoust: 0 is valid on an empty dataset--there is no difference, in the state machine, between a `SourceState` with no data and a newly initialized `SourceState`

packages/studio-base/src/panels/Plot/processor/downsample.ts
cfoust: Haha, yup. But at least now we could adjust the max on the fly and even according to the user agent :)

packages/studio-base/src/panels/Plot/processor/downsample.ts
cfoust: ooo, I was just copying the pattern elsewhere in our other bounds checks, but this makes sense

packages/studio-base/src/panels/Plot/processor/downsample.ts
cfoust: they don't need exact matches-- it just finds the closest (from below), the base case (`start >= end`) covers the rest

also maybe i'm opening a can of worms but the correctness of this specific helper function doesn't matter that much; we just use it as a quick and dirty way of only slicing the portion of the dataset that fits in the viewport window; we don't need point-level accuracy for that 🤷

packages/studio-base/src/panels/Plot/processor/downsample.ts
cfoust: 👍

packages/studio-base/src/panels/Plot/processor/downsample.ts
cfoust: hmm, you're right, this is interesting

packages/studio-base/src/panels/Plot/processor/downsample.ts
cfoust: yeah it seems generally useful tbh

packages/studio-base/src/panels/Plot/processor/downsample.ts
cfoust: if a path changes, the downsample state for the client is reset completely (not that this is perfect, but it's what we have)

packages/studio-base/src/panels/Plot/processor/downsample.ts
cfoust: yes

packages/studio-base/src/panels/Plot/processor/downsample.ts
cfoust: yes, but in practice this costs us very little (N is quite small)

packages/studio-base/src/components/TimeBasedChart/lttb.test.ts
```diff
@@ -0,0 +1,29 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import * as R from "ramda";
+
+import { createDataset } from "@foxglove/studio-base/panels/Plot/processor/testing";
```
defunctzombie, line 7: A quirk I noticed but these tests import from the Plot panel even though the lttb stuff is in TimeBasedChart which is a dependency of the plot panel. A bit of a backwards dependency import.

packages/studio-base/src/components/TimeBasedChart/lttb.test.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  it("does not downsample if numBuckets == 0", () => {
+    const result = downsampleLTTB(createData(10), 10, 0);
+    expect(result?.length).toEqual(10);
+  });
+
+  it("downsamples", () => {
+    const result = downsampleLTTB(createData(100), 100, 10);
+    expect(result?.length).toEqual(10);
+    expect(R.all((index) => index >= 0 && index < 100, result ?? [-1])).toEqual(true);
```
defunctzombie, line 27: Good additional test ideas:

1. Ensure that the output index set is strictly monotonic. We should never see reversions in index where a lower index comes after a higher one in the output array.
2. Make `createData` generate some random y-values. Your current data is strictly monotonic y-values and could be interesting to not have that property.
3. Make `createData` generate some variance in the x-value which increases by exactly 1 each time.

packages/studio-base/src/components/Chart/datasets.ts
```diff
@@ -135,3 +134,67 @@ export function findIndices(
 
   return undefined;
 }
+
+/**
+ * fastFindIndices returns a faster version of findIndices in exchange for
```
defunctzombie, line 139: What compute does it do ahead of time? Or where is this compute? I didn't understand the "compute ahead of time" parts of this comment. The compute looks like it is happening within the function? So why have a version that is not "fast"? When would I use one vs the other?

PR #7005, opened 2023-10-19
state transitions panel support display topic

smbera:
**User-Facing Changes**

**Description**
Supports displaying the time when a topic appears, making it easy to locate the time when a specific event occurs and start playing from the time of occurrence. ( same as RVIZ. )

rviz

**After**
image


jtbandes:
Could you provide some details on how to see this feature in RViz? I've never seen it there before. Or is it a separate program?

jtbandes:
What is the name of that program?

defunctzombie:
Thanks for proposing the idea and thinking about some workflows however this is a feature we want to offer at the app level rather than in the state transition panel. The state transitions panel is for viewing state transitions rather than dataset overview and we'd like to keep its behavior specific to that workflow rather than overload it with this behavior.

PR #7010, opened 2023-10-23
Control error clearing in ImageRenderable

snosenzo:
**User-Facing Changes**

- n/a

**Description**

If an image sets an error while decoding but still resolves successfully in the promise, then we don't want to clear the errors set on the renderable.


snosenzo:
no longer necessary

PR #7011, opened 2023-10-23
AppBar needs no currentUser

wimagguc:
**User-Facing Changes**

Removes `currentUser` and the signin button from the top navigation menu. These are not useful in public Studio.


packages/studio-base/src/components/PublicAppBar/index.tsx
jtbandes, line 1: I don't love the "PublicAppBar" name...maybe just "AppBar" or "SimpleAppBar"/"BasicAppBar" or something like that?

packages/studio-base/src/components/PublicAppBar/UserMenu.tsx
jtbandes, line 1: This should probably be renamed to SettingsMenu or something like that, since it has nothing to do with users anymore. Instead of displaying a 👤 icon we should use a ⚙️ or something like that.

We might also want to explore completely merging the two dropdowns (left & right) since there's a lot of redundancy already. cc @2metres for thoughts.

wimagguc: I like the `SettingsMenu` name, will make that change in a sec. Thanks!

jtbandes: Well if there's no user/login concept in this repo at all, it seems weird to me to use a 👤. I guess we can always do it in a separate PR though.

jtbandes (approved):

PR #7012, opened 2023-10-23
Add `user-select: all` to TextMiddleTruncate

### LINKED ISSUE #6878: Title copying still broken
**Description**
Similar to https://github.com/foxglove/studio/issues/6403, but now the behavior is that when you click on title text it highlights the first half, but ctrl+c gives you the full title text. There is no way to select a portion of the text and copy it, like normal text.

- Version: 1.68.0 (darwin)
- Platform: osx

**Steps To Reproduce**
Open an mcap file. Try to select part of the file name in the title and copy.

**Expected Behavior**
You can select text to copy it. Instead you get the full title/filename always

jtbandes:
**User-Facing Changes**

**Description**
In #6468 we changed the behavior of the "copy" command on TextMiddleTruncate components to copy the entire string. This was done to avoid issues where a [stray newline character](https://github.com/foxglove/studio/issues/6403) was copied. However, it introduced confusion because the user-selected text was no longer the text that is copied. This PR reduces confusion by adding `user-select: all` which makes it clear that the full string is implicitly selected and will be copied.

This is probably not an ideal user experience, but it's a trivial change to decrease confusion a bit, and we haven't come up with a good proposal for a more complete middle-truncation solution and copying experience that would be intuitive as well as consistent with the rest of the app.

Closes #6878

### DIFF
--- a/packages/studio-base/src/components/TextMiddleTruncate.tsx
+++ b/packages/studio-base/src/components/TextMiddleTruncate.tsx
@@ -21,6 +21,7 @@ const useStyles = makeStyles()(() => ({
     display: "flex",
     justifyContent: "flex-start",
     overflow: "hidden",
+    userSelect: "all",
   },
   start: {
     overflow: "hidden",


snosenzo (approved):
seems reasonable. don't see a deployment to test this out on though, which is curious

PR #7013, opened 2023-10-23
Add size setting for plot lines

alexern14:
**User-Facing Changes**

Add line size setting option to plot panel.

**Description**

The following image shows the blue series with the auto line size which is unchanged, and the orange series with an increased line size.
image


jtbandes:
Internal tracking issue: FG-5377

alexern14:
@jtbandes @cfoust Is there any update on this?

packages/studio-base/src/panels/Plot/settings.ts
```diff
@@ -58,6 +58,14 @@ const makeSeriesNode = memoizeWeak(
           label: t("color"),
           value: path.color ?? lineColors[index % lineColors.length],
         },
+        size: {
```
jtbandes, line 61: Maybe we can rename this (and the settings key) to `lineSize`? Seems better to be a bit more explicit for future sanity.

packages/studio-base/src/panels/Plot/settings.ts
```diff
@@ -58,6 +58,14 @@ const makeSeriesNode = memoizeWeak(
...
+          input: "number",
+          label: "Line Size",
```
jtbandes, line 63: This should be `t("lineSize")` and added to `i18n/en/plot.ts` so it can be localized. Also we generally use sentence case, i.e. `"Line size"` instead of `"Line Size"`.

jtbandes (approved):

alexern14:
@jtbandes This is my first time putting up a PR for you guys. What's the process for getting the UI reviewed and tests approved?

jtbandes:
We can take care of the approval/merging step. Could you also add another entry to `Plot/index.stories.tsx` to create a new screenshot test demonstrating the line size functionality? You should be able to copy/paste one of the simpler entries there (e.g. `HiddenConnectingLines`) and just add a `lineSize` value to the hard-coded config.

packages/studio-base/src/panels/Plot/index.stories.tsx
```diff
@@ -1670,3 +1670,44 @@ export const PreloadedDataAndItsAbsoluteValue: StoryObj = {
...
+        pauseFrame={pauseFrame}
+        config={{
+          ...exampleConfig,
+          paths: [
+            {
+              value: "/some_topic/location.pose.velocity",
+              enabled: true,
+              showLine: false,
+              timestampMethod: "receiveTime",
+              lineSize: 1.4,
```
jtbandes, line 1690: Making the difference more obvious
```suggestion
              timestampMethod: "receiveTime",
              lineSize: 2.5,
```

packages/studio-base/src/panels/Plot/index.stories.tsx
```diff
@@ -1670,3 +1670,44 @@ export const PreloadedDataAndItsAbsoluteValue: StoryObj = {
...
+    return (
+      <PlotWrapper
+        pauseFrame={pauseFrame}
+        config={{
+          ...exampleConfig,
+          paths: [
+            {
+              value: "/some_topic/location.pose.velocity",
+              enabled: true,
+              showLine: false,
```
jtbandes, line 1688: Please remove the `showLine: false` too since otherwise we see only the points in the screenshot. The new 2.5 size looks good though! They're obviously bigger than the other points

jtbandes:
Thank you!

PR #7014, opened 2023-10-23
util: replace overwriteFetch with a wrapper

james-rms:
**User-Facing Changes**
None.

**Description**

Context: I'm trying to update `@node/types` to `20.8.4` in foxglove/app. That update includes types for the `fetch` API in node, see this recently resolved issue: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/66824 . These allow the fetch API to be used in typescript without using the `lib: ["dom"]` in `tsconfig.json`.

In doing this, I've found that `overwriteFetch` conflicts with the type definitions given for `fetch`. The error is:

```
  ERROR in ../../studio/packages/studio-base/src/util/overwriteFetch.ts:18:3
  TS2322: Type '(url: RequestInfo | URL, init?: RequestInit) => Promise<Response>' is not assignable to type '{ (input: RequestInfo | URL, init?: RequestInit | undefined): Promise<Response>; (input: RequestInfo | URL, init?: RequestInit | undefined): Promise<...>; (input: RequestInfo, init?: RequestInit | undefined): Promise<...>; }'.
    Types of parameters 'url' and 'input' are incompatible.
      Type 'RequestInfo' is not assignable to type 'RequestInfo | URL'.
        Type 'Request' is not assignable to type 'RequestInfo | URL'.
          Property 'referrer' is missing in type 'import("/home/runner/work/app/app/node_modules/undici-types/fetch").Request' but required in type 'Request'.
      16 | export default function overwriteFetch(): void {
      17 |   const originalFetch = global.fetch;
    > 18 |   global.fetch = async (url: RequestInfo | URL, init?: RequestInit) => {
         |   ^^^^^^^^^^^^
      19 |     // Use this replacement error instead of the original one, because this one will have the correct stack trace.
      20 |     // eslint-disable-next-line @typescript-eslint/no-base-to-string
      21 |     const replacementError = new TypeError(`Failed to fetch: ${url}`);
```

As I understand it (very roughly) the `@node/types` definition gives a few overloads for `fetch`, and we can't satisfy all of them by overwriting fetch with an implementation of only one of those overloads. 

To solve this, I figured the cleanest solution then is to not overwrite the global `fetch` object, and instead just provide a function wrapper for our own use. This seems like a less surprising way to implement the functionality we want, and doesn't involve replicating a bunch of type definitions to appease the type checker.

Note: Before this PR I did try replacing the argument type definitions of the replacement fetch with `input: Parameters<typeof fetch>[0], init?: Parameters<typeof fetch>[1]` but ended up with the same errors.


jtbandes:
I don't think this is terribly risky but it does have the downside that it won't handle cases where third-party/imported code uses the global `fetch` (examples I can think of but haven't verified manually: three.js loaders, wasm wrappers that fetch their .wasm files).

james-rms:
Found a better solution, closing.

PR #7019, opened 2023-10-24
Use user-facing script name in user script errors

cfoust:
**User-Facing Changes**
Use user-defined user script names in errors instead of the script's ID.

**Description**
This seems to have just been a mistake; as far as I'm aware we don't actually show this to the user anywhere else.


jtbandes (approved):
🎉

PR #7020, opened 2023-10-24
Fix collada textures not being loaded over websocket

achim-k:
**User-Facing Changes**
Fix collada textures not being loaded over websocket

**Description**
Studio loads collada meshes through the asset management layer which, in case of a foxglove-websocket connection, may fetch the asset over the websocket connection. However, this was not the case for texture assets that are referenced by the collada mesh, resulting in the entire collada mesh not being loaded.

This PR fixes this by also loading the referenced texture assets through Studio's asset management layer.

Resolves FG-5340

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/ModelCache.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/ModelCache.ts
@@ -233,6 +233,10 @@ export class ModelCache {
     const dae = daeLoader.parse(xmlText, baseUrl(url));
     manager.itemEnd(url);
 
+    for (const objectUrl of textureUrls.values()) {
+      URL.revokeObjectURL(objectUrl);
+    }
+
     // If the <up_axis> is Y_UP, rotate to the Studio convention of Z-up following
     // ROS [REP-0103](https://www.ros.org/reps/rep-0103.html)
     if (upAxis === "Y_UP") {
--- a/packages/studio-base/src/panels/ThreeDeeRender/ModelCache.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/ModelCache.ts
@@ -203,8 +203,30 @@ export class ModelCache {
     });
     const xmlText = xml.documentElement.outerHTML;
 
+    // Preload textures. We do this here since we can't pass in an async function in LoadingManager.setURLModifier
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/panels/ThreeDeeRender/ModelCache.ts
```diff
@@ -203,8 +203,30 @@ export class ModelCache {
...
+      if (!node.textContent) {
+        continue;
+      }
+
+      try {
+        const textureUrl = new URL(node.textContent, baseUrl(url)).toString();
+        const textureAsset = await this.#fetchAsset(textureUrl);
+        const objectUrl = URL.createObjectURL(
+          new Blob([textureAsset.data], { type: textureAsset.mediaType }),
+        );
```
achim-k, line 220: There is currently no way to call [URL.revokeObjectURL()](https://developer.mozilla.org/en-US/docs/Web/API/URL/revokeObjectURL_static) on this...

jtbandes: It wouldn't work to do it after the `itemEnd`?

jtbandes (approved):

packages/studio-base/src/panels/ThreeDeeRender/ModelCache.ts
achim-k: Oh yes, that actually works :+1:

packages/studio-base/src/panels/ThreeDeeRender/ModelCache.ts
```diff
@@ -203,14 +203,40 @@ export class ModelCache {
...
+        textureUrls.set(textureUrl, objectUrl);
+      } catch (e) {
+        log.error(e);
+        onError(node.textContent);
+      }
+    }
+
     const manager = new THREE.LoadingManager(undefined, undefined, onError);
-    manager.setURLModifier(rewriteUrl);
+    manager.setURLModifier((u) => textureUrls.get(u) ?? rewriteUrl(u));
```
defunctzombie, line 229: small stylistic nit: I dislike the use of single letter variable names (unless they are math constants). Even in limited settings such as this function definition having a few more characters helps reading by removing ambiguity on what _u_ is suppose to be.

PR #7022, opened 2023-10-25
Save Font Settings for Raw Messages Panel

pezy:
**User-Facing Changes**

Font settings of the RawMessage panel can be saved with the layout

**Description**

#6397 thoughtfully added the font setting function, but unfortunately, every time the data is switched or refreshed after the setting is completed, it will be reset to `auto` and needs to be set again.

I hope this setting can be remembered.

### DIFF
--- a/packages/studio-base/src/panels/RawMessages/index.tsx
+++ b/packages/studio-base/src/panels/RawMessages/index.tsx
@@ -96,7 +96,8 @@ function RawMessages(props: Props) {
   const jsonTreeTheme = useJsonTreeTheme();
   const { config, saveConfig } = props;
   const { openSiblingPanel } = usePanelContext();
-  const { topicPath, diffMethod, diffTopicPath, diffEnabled, showFullMessageForDiff } = config;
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/panels/RawMessages/index.tsx
```diff
@@ -145,7 +145,7 @@ function RawMessages(props: Props) {
   }, [structures, topic, topicRosPath]);
 
   const [expansion, setExpansion] = useState(config.expansion);
-  const [customFontSize, setCustomFontSize] = useState<number | undefined>();
+  const [customFontSize, setCustomFontSize] = useState<number | undefined>(config.fontSize);
```
jtbandes, line 148: You shouldn't need `useState` anymore, you can add fontSize to the `const { ... } = config` on line 99 instead.

pezy: You're right, it's definitely cleaner...

jtbandes (approved):

jtbandes:
Thanks!

PR #7023, opened 2023-10-25
foxglove websocket: Fix service parsing errors not being surfaced to user

### LINKED ISSUE #6967: Service calls silently fail when messageEncoding is JSON and schemaEncoding is not jsonschema
**Description**

We have a non-ROS robot that is advertising services over ws-protocol to Studio. The robot wants to receive service calls with JSON serialization and advertises "json" as a supported encoding, but the "message definition" for the services are published as protobuf definitions. This setup silently fails since there is no way to specify the schema encoding for services in ws-protocol.

```
if (this.#serviceCallEncoding === "json") {
  schemaEncoding = "jsonschema";
}
```

- Version: latest
- Platform: all
- Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native): Custom ws-protocol server

achim-k:
**User-Facing Changes**
foxglove websocket: Fix service parsing errors not being surfaced to user

**Description**
Fixes errors occurred while parsing service definitions not being made visible to the user. 

Fixes #6967
Resolves FG-5274

### DIFF
--- a/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
+++ b/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
@@ -615,46 +615,60 @@ export default class FoxgloveWebSocketPlayer implements Player {
       }
 
       for (const service of services) {
-        const requestType = `${service.type}_Request`;
-        const responseType = `${service.type}_Response`;
-        const parsedRequest = parseChannel({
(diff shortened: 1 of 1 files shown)


jtbandes (approved):
What happened to this error before? It was just uncaught?

achim-k:
> What happened to this error before? It was just uncaught?

Yes, they were only logged to console and never surfaced to the user

PR #7024, opened 2023-10-25
Fix inconsistent shadow and min height on plot legend

2metres:
**User-Facing Changes**
None

**Description**
Screen Shot 2023-10-26 at 9 34 40 am

Floating legend is now consistent with other floating buttons and the rows.

### DIFF
--- a/packages/studio-base/src/panels/Plot/PlotLegend.tsx
+++ b/packages/studio-base/src/panels/Plot/PlotLegend.tsx
@@ -52,14 +52,14 @@ const useStyles = makeStyles<void, "container" | "toggleButton" | "toggleButtonF
     rootFloating: {
       padding: spacing(0.75), // pad the container to prevent shadow from being clipped
       pointerEvents: "none",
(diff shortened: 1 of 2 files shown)


defunctzombie (approved):
This fixes the visual inconsistency which I suppose resolves the bug but I find that this button and the whitespace is different than the panel titlebar height? Maybe that is a separate issue and we need to bring panel height inline with our ROW_HEIGHT?

2metres:
> This fixes the visual inconsistency which I suppose resolves the bug but I find that this button and the whitespace is different than the panel titlebar height? Maybe that is a separate issue and we need to bring panel height inline with our ROW_HEIGHT?

Both should be 30 px from what I can tell

defunctzombie:
> Both should be 30 px from what I can tell

I checked with inspector and indeed they are both 30px. I guess the drop shadow makes the button somehow "look" like it is bigger than the toolbar. Not worth tackling now if they are both the same but maybe something interesting to consider with floating vs flat elements.

PR #7025, opened 2023-10-25
Remove corners from floating overlay buttons

2metres:
**User-Facing Changes**
None

**Description**
Remove corners from floating overlay buttons


defunctzombie:
Are there no user facing changes here? This sounds like a change that would have some sort of visual change for a user?

jtbandes (approved):
There is a visible change but only if you are squinting at pixels. I wouldn’t put it in the release notes.

PR #7026, opened 2023-10-25
Add border bottom to topic list row

2metres:
**User-Facing Changes**
| Before | After |
| --- | --- |
| image | image |

**Description**
Add border to bottom of topic list items based on feedback

### DIFF
--- a/packages/studio-base/src/components/TopicList/useTopicListStyles.ts
+++ b/packages/studio-base/src/components/TopicList/useTopicListStyles.ts
@@ -12,7 +12,6 @@ import { makeStyles } from "tss-react/mui";
 
 type TreeClasses = "dragHandle" | "row" | "isDragging" | "selected";
 
-/* eslint-disable tss-unused-classes/unused-classes */
 export const useTopicListStyles = makeStyles<void, TreeClasses>()((theme, _, classes) => ({
   isDragging: {},
   selected: {},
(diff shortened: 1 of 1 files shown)


defunctzombie:
What does this accomplish? Please elaborate a bit more in the PR description (screenshots if helpful) on some motivation for the change; the current description repeats the title.

packages/studio-base/src/components/TopicList/useTopicListStyles.ts
```diff
@@ -69,6 +65,11 @@ export const useTopicListStyles = makeStyles<void, TreeClasses>()((theme, _, cla
       opacity: 1,
     },
   },
+  /* eslint-disable tss-unused-classes/unused-classes */
```
jtbandes, line 68: This disable comment disables the error for the rest of the file. You should either leave it at the top of the file, or add a corresponding eslint-enable below, or change it to a "next-line" disable.

```suggestion
  // eslint-disable-next-line tss-unused-classes/unused-classes
```

jtbandes (approved):

packages/studio-base/src/components/TopicList/useTopicListStyles.ts
2metres: > This disable comment disables the error for the rest of the file 

That's why I've reordered and added it so that the classes I want to ignore are below the comment.

defunctzombie: @2metres I don't think that's how it works - once that comment is present I think it disables it for the entire file. Either way this is fragile and we should disable the lint rules specifically where needed and with a comment on why. Looking at this I don't see why we need to disable so lets re-enable or add a comment so others that maintain this code can understand.

2metres: It appears to me that the linter is working with the change. But if this is actually not I can make the change so that the lint rule is a comment fence instead.

**Why we need to disable:**
This file is shared between a few different files. The lint rule is fairly crude in how it checks for uses of classes. In the cases where the we call the output classes it pases and everything is cool. In the cases where we call it something else because another const with the same name exists, it fails.

jtbandes: https://github.com/foxglove/studio/pull/7082

Issue #6984, opened 2023-10-12
Each Grid message causes two calls to message converter


henrik-kaijser:
**Description**
Using an mcap file that contains foxglove.Grid messages and a message converter that converts "foxglove.Grid -> another type", this causes two calls with the same parameters from the framework to my custom message converter. I think this is a bug, can't see why there should be a second identical call? This only seems to happen when foxglove.Grid is the source type, although I have only tested some of the other types.

- Version: Foxglove studio 1.72.0+8aa80e1.fdaac58, browser version
- Platform: Ubuntu 22.04.3 LTS
- Data source mcap file

**Steps To Reproduce**
1. Write a message converter that use foxglove.Grid as source type, see attached file index.zip for index.ts file as an example
2. Use Foxglove studio with a layout that only has a single 3D panel
3. Use attached mcap file (see only_ogrid.zip) as input. Note: This mcap has no transform messages, but you get the same behavior even if it does.
4. View the console output window in the browser
5. There are two lines for each foxGlove.grid message in the mcap-file, i.e. console output: 
```
index.ts:12 new call: topic = /env/output/ogrid t = 102 202478650
index.ts:12 new call: topic = /env/output/ogrid t = 102 202478650
index.ts:12 new call: topic = /env/output/ogrid t = 102 262478649
index.ts:12 new call: topic = /env/output/ogrid t = 102 262478649
```

**Expected Behavior**
Only one call per message, i.e. console output
index.ts:12 new call: topic = /env/output/ogrid t = 102 202478650
index.ts:12 new call: topic = /env/output/ogrid t = 102 262478649

[only_ogrid.zip](https://github.com/foxglove/studio/files/12878751/only_ogrid.zip)
[index.zip](https://github.com/foxglove/studio/files/12878775/index.zip)

PR #7027, opened 2023-10-26
Avoid adding the same ID twice to subscriberIdsByTopic

defunctzombie:
**User-Facing Changes**
Fix duplicate messages being sent to a panel if it subscribed twice to the same topic.

**Description**
This change updates the `MessagePipeline` store `updateSubscriberAction` function to avoid adding a subscriber ID multiple times to the `subscriberIdsByTopic` array for a particular topic. Including the ID multiple times results in other logic in the store which handles message distribution to erroneously provide the message to the panel multiple times here: https://github.com/foxglove/studio/blob/main/packages/studio-base/src/components/MessagePipeline/store.ts#L289-L300

A panel might subscribe to the same topic multiple times if it an array of subscription requests and includes the same topic multiple times and does not itself de-duplicate the array entries based on topic name. A panel is not required to do such de-duplication.

This issue was uncovered by #6984  which created a situation where the 3d panel thought it could subscribe to the same topic twice because the original topic was a `foxglove.Grid` schema and there was a converter installed which provided a conversion from `foxglove.Grid` to `foxglove.PosesInFrame`. Both of these schemas are 3d renderable and the panel wanted to subscribe for both.

Fixes: #6984

### DIFF
--- a/packages/studio-base/src/components/MessagePipeline/store.ts
+++ b/packages/studio-base/src/components/MessagePipeline/store.ts
@@ -239,7 +239,12 @@ function updateSubscriberAction(
       const topic = subscription.topic;
 
       const ids = subscriberIdsByTopic.get(topic) ?? [];
-      ids.push(id);
+      // If the id is already present in the array for the topic then we should not add it again.
+      // If we add it again it will be given frame messages again when bucketing incoming messages
+      // by subscriber id.
+      if (!ids.includes(id)) {
+        ids.push(id);
+      }
       subscriberIdsByTopic.set(topic, ids);
     }
   }
--- a/packages/studio-base/src/components/MessagePipeline/store.ts
+++ b/packages/studio-base/src/components/MessagePipeline/store.ts
@@ -48,17 +48,19 @@ export type MessagePipelineInternalState = {
 
   /** used to keep track of whether we need to update public.startPlayback/playUntil/etc. */
   lastCapabilities: string[];
-  // Preserves reference equality of subscriptions to minimize player subscription churn.
+  /** Preserves reference equality of subscriptions to minimize player subscription churn. */
   subscriptionMemoizer: (sub: SubscribePayload) => SubscribePayload;
   subscriptionsById: Map<string, Immutable<SubscribePayload[]>>;
   publishersById: { [key: string]: AdvertiseOptions[] };
   allPublishers: AdvertiseOptions[];
-  // A map of topic name to the IDs that are subscribed to that topic. Incoming messages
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/components/MessagePipeline/store.ts
```diff
@@ -239,7 +239,12 @@ function updateSubscriberAction(
       const topic = subscription.topic;
 
       const ids = subscriberIdsByTopic.get(topic) ?? [];
-      ids.push(id);
+      // If the id is already present in the array for the topic then we should not add it again.
+      // If we add it again it will be given frame messages again when bucketing incoming messages
+      // by subscriber id.
+      if (!ids.includes(id)) {
```
snosenzo, line 245: would it make more sense to use a set instead of an array in `subscriberIdsByTopic`? It would be better if we required deduplication in the type

defunctzombie: I think yes it could (though arrays are typically cheaper to work with). I know some time ago I benchmarked Set and the behavior was poor for small sets and then more recently benchmarked again and it was good behavior so I think its ok to update this to a set now.

defunctzombie: I thought about this some more and an array is the correct data structure. I've updated the PR with a comment on why. We want to optimize for the hot path use case and an array is better for the hot path than a set with our current logic.

jtbandes (commented):
Can we add a unit test that covers this fix?

packages/studio-base/src/components/MessagePipeline/store.ts
```diff
@@ -53,6 +53,12 @@ export type MessagePipelineInternalState = {
   subscriptionsById: Map<string, Immutable<SubscribePayload[]>>;
   publishersById: { [key: string]: AdvertiseOptions[] };
   allPublishers: AdvertiseOptions[];
+  // A map of topic name to the IDs that are subscribed to that topic. Incoming messages
```
snosenzo, line 56: nit: can we make this comment use `/** */ ` so it shows up in tips?

snosenzo (approved):
agree w/ jacob that it would be nice if we had a unit test for this, but I do see that it's internal state, so I'll leave it up to you

packages/studio-base/src/components/MessagePipeline/store.ts
defunctzombie: Yep - my oversight.

defunctzombie:
Test should be added so I've added a test.

jtbandes (approved):

PR #7030, opened 2023-10-26
Allow ability to toggle frame lock on more 3d topics

snosenzo:
**User-Facing Changes**

**Description**
In draft for testing.
This would let things like images in the 3D scene to use their headerstamp to get transform data rather than the transform at the current time according to studio. When there is a mismatch of clocks between studio and the websocket server, this can cause issues becuase studio uses the transform at the currentTime (Date.now())  to access the corresponding transform with a headerstamp at that time. This can make it look like new messages are using old transforms if the studio clock is behind the websocket server clock that's writing the headerstamps

Potentially fixes this issue: FG-5229


defunctzombie:
@snosenzo is this still something we want? We closed the linear issue so maybe this is no longer relevant?

snosenzo:
Yeah I don't believe it is relevant given our conclusions about proper handling of time differences over websocket being to use a clock signal.

PR #7037, opened 2023-10-27
Move NodeEditor error icon position

defunctzombie:
**User-Facing Changes**
When showing an error on a topic, the error dot appears in place of the topic icon.

Before:
Screenshot 2023-11-01 at 8 06 34 PM
Screenshot 2023-11-01 at 8 06 24 PM

After:
Screenshot 2023-11-01 at 8 05 13 PM
Screenshot 2023-11-01 at 8 05 06 PM

**Description**
This change moves the error icon for a NodeEditor section to the left of the title. Previously this icon would be to the right of the visibility indicator. This would create a poor UX experience when the user would toggle a topic on which could result in an error on the topic immediately displacing the visibility icon they clicked and placing the error icon under their cursor. After this change the visibility icon remains under the cursor even if there is a topic error.


jtbandes (commented):
Is there a corresponding change to make for FieldEditor as well?

It seemed like the original ticket description was saying that the icon should be to the left _of the visibility toggle_, not necessarily on the _far_ left of the row.

defunctzombie:
> It seemed like the original ticket description was saying that the icon should be to the left _of the visibility toggle_, not necessarily on the _far_ left of the row.

I forgot to add commentary on this to my description. I originally tried this but found it looked awkward when the user was not hovering on the row and when they did hover on the row. The error icon would appear between the visibility icon and the edit icon (if editing was allowed). When not hovering the error icon would appear with a gap (which could be ok but seemed nicer if I could find a way to avoid it). And finally - the user can add custom icons which seem to appear _after_ the visibility icon. So if the user has custom icons there would be a gap between those and the error icon when not hovering and that too looked awkward.

All of this led me towards exploring a location for the error dot that did not position itself in a place that it was influenced by hover state icons.

jtbandes (approved):

2metres:
Instead of being to the left could we instead replace the specified icon (in the screenshot the cog) with the error icon?

defunctzombie:
> Instead of being to the left could we instead replace the specified icon (in the screenshot the cog) with the error icon?

Good idea - I'll give that a try.

defunctzombie:
I've updated the PR with logic to replace the topic icon with the error icon. I do think that provides a good visual indicator the specific topic is in error.

@jtbandes FieldEditor does not use the same type of error display and does not need an update.

2metres (approved):

amacneil:
:+1:

PR #7038, opened 2023-10-27
Remove camera info missing errors when calibration is `None`

snosenzo:
**User-Facing Changes**

- Remove occasional camera info missing errors when calibration is `None`

**Description**
Also consolidated the setting of the image topic so that we can take advantage of it in the subclass of imagemode.

Fixes: FG-5458


packages/studio-base/src/panels/ThreeDeeRender/stories/ImageMode/ImageMode.stories.tsx
```diff
@@ -696,8 +696,8 @@ export const UnsupportedEncodingError: StoryObj = {
   play: async () => {
     const errorIcon = await waitFor(async () => {
       const icons = await screen.findAllByTestId("ErrorIcon");
-      if (icons.length !== 2) {
-        throw new Error("Expected 2 error icons");
+      if (icons.length !== 1) {
+        throw new Error("Expected l error icon. (unsupported encoding)");
```
jtbandes, line 700:
```suggestion
        throw new Error("Expected 1 error icon. (unsupported encoding)");
```

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageMode.ts
```diff
@@ -750,7 +755,7 @@ export class ImageMode
    */
   #updateViewAndRenderables(): void {
     const cameraInfo = this.#cameraModel?.info;
-    if (!cameraInfo) {
+    if (!this.#fallbackCameraModelActive() && !cameraInfo) {
```
jtbandes, line 758: So I'm guessing this is the main fix... why did this only happen sometimes and not all the time?

jtbandes (approved):

jtbandes:
Another question: did this regress at some point? I seem to remember there was a time when these errors weren't showing up...

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageMode.ts
snosenzo: I think there's a race condition. I'm not able to reproduce it consistently

snosenzo:
> Another question: did this regress at some point? I seem to remember there was a time when these errors weren't showing up...

Also unsure about that, it does have to do with timing and how we decode the images.

PR #7039, opened 2023-10-27
Improve estimation of per message memory usage

achim-k:
**User-Facing Changes**
Improve estimation of per message memory usage

**Description**
When loading a recording, Studio attempts to buffer 10 seconds ahead (for a smoother playback experience) and preload messages used in plots or 3D panels (tf) for the entire recording duration. This is however not always possible, especially when dealing with larger messages. Studio will stop reading messages ahead if 1GB worth of messages have been read (both for preloading and playback buffer). 

For determining how many messages can still be read into the buffer, Studio uses the MessageEvent's [`sizeInBytes`](https://github.com/foxglove/studio/blob/b18467986cda61a73eb098da4c6a26cf59dd118a/packages/studio/src/index.ts#L116-L120) which is the size of the serialized message (and which obviously depends on the message encoding). Unfortunately, the deserialized message javascript object can take up a lot more space, especially when the message contains lots of fields. This underestimation of per-message memory size causes Studio to read ahead more messages than it should which might eventually lead to the app running out of memory :headstone: 

This PR attempts to improve the estimation of the memory a deserialized message will take up. The approach taken is rather simple: When parsing the schema, we count the number of primitive fields (e..g bool, int8, int32, ...) and use that to guesstimate the approximate memory size of a deserialized message object

### DIFF
--- a/packages/studio-base/src/players/IterablePlayer/Mcap/McapIndexedIterableSource.ts
+++ b/packages/studio-base/src/players/IterablePlayer/Mcap/McapIndexedIterableSource.ts
@@ -179,10 +179,6 @@ export class McapIndexedIterableSource implements IIterableSource {
           channelInfo.approxDeserializedMsgSize,
         );
 
-        log.info(
-          `${channelInfo.schemaName}, ${message.data.byteLength}, ${channelInfo.approxDeserializedMsgSize}`,
-        );
-
         yield {
           type: "message-event",
           msgEvent: {
--- a/packages/studio-base/src/players/IterablePlayer/Mcap/McapIndexedIterableSource.ts
+++ b/packages/studio-base/src/players/IterablePlayer/Mcap/McapIndexedIterableSource.ts
@@ -16,6 +16,7 @@ import {
   IteratorResult,
   MessageIteratorArgs,
 } from "@foxglove/studio-base/players/IterablePlayer/IIterableSource";
+import { guesstimateDeserializedMsgSize } from "@foxglove/studio-base/players/messageMemoryEstimation";
 import { PlayerProblem, Topic, TopicStats } from "@foxglove/studio-base/players/types";
 import { RosDatatypes } from "@foxglove/studio-base/types/RosDatatypes";
 
@@ -25,7 +26,13 @@ export class McapIndexedIterableSource implements IIterableSource {
   #reader: McapIndexedReader;
   #channelInfoById = new Map<
     number,
-    { channel: McapTypes.Channel; parsedChannel: ParsedChannel; schemaName: string | undefined }
+    {
+      channel: McapTypes.Channel;
+      parsedChannel: ParsedChannel;
+      schemaName: string | undefined;
+      // Guesstimate of the memory size in bytes of a deserialized message object
+      approxDeserializedMsgSize: number;
+    }
   >();
   #start?: Time;
   #end?: Time;
@@ -63,8 +70,13 @@ export class McapIndexedIterableSource implements IIterableSource {
       }
 
       let parsedChannel;
+      let approxDeserializedMsgSize;
       try {
         parsedChannel = parseChannel({ messageEncoding: channel.messageEncoding, schema });
+        approxDeserializedMsgSize =
+          schema?.name != undefined
+            ? guesstimateDeserializedMsgSize(parsedChannel.datatypes, schema.name)
+            : 0;
       } catch (error) {
         problems.push({
           severity: "error",
@@ -73,7 +85,12 @@ export class McapIndexedIterableSource implements IIterableSource {
         });
         continue;
       }
-      this.#channelInfoById.set(channel.id, { channel, parsedChannel, schemaName: schema?.name });
+      this.#channelInfoById.set(channel.id, {
+        channel,
+        parsedChannel,
+        schemaName: schema?.name,
+        approxDeserializedMsgSize,
+      });
 
       let topic = topicsByName.get(channel.topic);
       if (!topic) {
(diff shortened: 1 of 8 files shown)


defunctzombie (commented):
Seems like a good band-aid. I have nothing else to add that @snosenzo is not already surfacing.

We'll need to update some of our closed source data sources to use this method as well.

packages/studio-base/src/players/messageMemoryEstimation.ts
```diff
@@ -0,0 +1,148 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { MessageDefinitionMap } from "@foxglove/mcap-support/src/types";
+
+const OBJECT_BASE_SIZE = 12;
+const TYPED_ARRAY_BASE_SIZE = 64; // byteLenght, byteOffset, ...
```
snosenzo, line 8:
```suggestion
const TYPED_ARRAY_BASE_SIZE = 64; // byteLength, byteOffset, ...
```

packages/studio-base/src/players/messageMemoryEstimation.ts
```diff
@@ -0,0 +1,148 @@
...
+
+    // In return, properties are no longer stored in the properties array
+    sizeInBytes -= 4 * nonConstantFields.length;
+  }
+
+  for (const field of nonConstantFields) {
+    if (field.isComplex ?? false) {
+      const count =
+        field.isArray === true
+          ? 0 // We are conservative and assume an empty array
```
snosenzo, line 81: why? also if it is fixed length, shouldn't we use that at least?

packages/studio-base/src/players/messageMemoryEstimation.ts
```diff
@@ -0,0 +1,148 @@
...
+          break;
+        case "int32":
+        case "uint32":
+        case "float32":
+          sizeInBytes += TYPED_ARRAY_BASE_SIZE + arrayLength * 4;
+          break;
+        case "float64":
+        case "int64":
+        case "uint64":
+          sizeInBytes += TYPED_ARRAY_BASE_SIZE + arrayLength * 8;
```
snosenzo, line 123: nit: this could be reduced a bit
```suggestion
        case "int8":
        case "uint8":
        case "int16":
        case "uint16":
        case "int32":
        case "uint32":
        case "float32":
        case "float64":
        case "int64":
        case "uint64":
          sizeInBytes += TYPED_ARRAY_BASE_SIZE + arrayLength * FIELD_SIZE_BY_PRIMITIVE[field.type];
          break;
```
or have a set of typedarray types and check to see if the field.type is in it

packages/studio-base/src/players/messageMemoryEstimation.test.ts
```diff
@@ -0,0 +1,98 @@
...
+          {
+            definitions: [
+              ...new Array(numInts).fill({ type: "int16", name: "field" }),
+              ...new Array(numFloats).fill({ type: "float64", name: "field" }),
+            ],
+          },
+        ],
+      ]);
+      const sizeInBytes = estimateMessageObjectSize(datatypes, "SomeType", new Map());
+      expect(sizeInBytes).toBeGreaterThanOrEqual(measuredSize * ((100 - tolerancePercent) / 100));
```
snosenzo, line 94: can probably reduce these checks to one with:
```suggestion
      expect(Math.abs(sizeInBytes-measuredSize).toBeLessThanOrEqual(measuredSize * .01 * tolerancePercent);
```

packages/studio-base/src/players/messageMemoryEstimation.ts
achim-k: Oh you are right, I do this below for arrays, but not here

snosenzo: should we also conservatively assume one element in the array instead of none?

snosenzo (approved):

defunctzombie (approved):

packages/studio-base/src/players/messageMemoryEstimation.ts
```diff
@@ -0,0 +1,152 @@
...
+
+import { MessageDefinitionMap } from "@foxglove/mcap-support/src/types";
+
+const OBJECT_BASE_SIZE = 12;
+const TYPED_ARRAY_BASE_SIZE = 64; // byteLength, byteOffset, ...
+const MAX_NUM_FAST_PROPERTIES = 1020;
+const FIELD_SIZE_BY_PRIMITIVE: Record<string, number> = {
+  bool: 4,
+  int8: 4,
+  uint8: 4,
```
snosenzo, line 13: I didn't take a closer look at these before -- where did these numbers come from?

packages/studio-base/src/players/messageMemoryEstimation.ts
snosenzo: oop the field size by primitive probably doesn't apply here.

packages/studio-base/src/players/messageMemoryEstimation.ts
achim-k: Mostly by inspecting object sizes in a memory snapshot, I wouldn't say that these numbers are 100% correct.

packages/studio-base/src/players/IterablePlayer/Mcap/McapIndexedIterableSource.ts
```diff
@@ -152,6 +174,15 @@ export class McapIndexedIterableSource implements IIterableSource {
...
         const payload = spec?.fields != undefined ? pickFields(msg, spec.fields) : msg;
+        const sizeInBytes = Math.max(
+          message.data.byteLength,
+          channelInfo.approxDeserializedMsgSize,
+        );
+
+        log.info(
+          `${channelInfo.schemaName}, ${message.data.byteLength}, ${channelInfo.approxDeserializedMsgSize}`,
+        );
+
```
achim-k, line 185: rm debug print

PR #7040, opened 2023-10-27
Update topic list search to include schema results when searching for topicname+'.'

jtbandes:
**User-Facing Changes**
Improved search matching algorithm in the Topics list.

**Description**

Originally I'd explicitly excluded field results from the search if the search string only matched `topic.` – I'm not totally sure why I did that; didn't have a test covering it, and it sounds like including the results in this case is more desirable.

Before | After
--- | ---
image<br>image | imageimage

### DIFF
--- a/packages/studio-base/src/components/TopicList/useTopicListSearch.test.ts
+++ b/packages/studio-base/src/components/TopicList/useTopicListSearch.test.ts
@@ -59,4 +59,20 @@ describe("useTopicListSearch", () => {
     );
     expect(result.current.map(itemToString)).toEqual(["xyz", "xyz.foobar", "footballer"]);
   });
+
+  it("includes topic matches when there's a trailing dot", () => {
+    const topics: UseTopicListSearchParams["topics"] = [
+      { name: "abc", schemaName: "ABCD" },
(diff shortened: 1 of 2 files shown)


defunctzombie (approved):
Magical. Feels much nicer to use.

I think your comment on line 58 is now outdated?

> // the path `suffix`. In this case we show only the topic row and not the message path rows.

achim-k:
Personally, I think that there are too many results now (might have been the case before already). The examples below show some matches which are somewhat far fetched. I would prefer if e.g. `imu` matches only with topics that contain the exact sequence


| Example 1                                                                                        | Example 2                                                                                        |
|--------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------|
| image | image |

jtbandes:
> I think your comment on line 58 is now outdated?
> 
> > // the path `suffix`. In this case we show only the topic row and not the message path rows.

After pondering this some more, I believe the change actually made the comment more accurate. The comment said "…exclude results if the query matched only the topic name and not the path `suffix`…" but previously were were also excluding results if the query matched the topic name + the beginning `.` which is part of the suffix. Now we are only excluding them if the match range does not extend into the suffix at all.

PR #7041, opened 2023-10-27
Fix a range of issues with user scripts and plots

cfoust:
**User-Facing Changes**
Fixed a handful of issues where plotting user script topics could lead to broken plots.

**Description**
This one is a doozy. It fixes three separate bugs that combined to create the poor experience described in [this GitHub ~issue~ discussion](https://github.com/orgs/foxglove/discussions/855):
1. If block data already exists on a topic a plot panel needs when it's mounted, the main thread delivers block data to the plot worker before the client is registered. The plot worker would drop the data.
  a. **Solution:** Now we hold on to incoming block data in anticipation of the registration of a client.
3. The plot panel block processor treated empty arrays on a topic as valid message data, not as being empty. This can occur if the `UserScriptPlayer` is partway through processing data when a subscription to a user script topic is cancelled.
  a. **Solution:** Empty message arrays now indicate an empty block.
5. The `BlockLoader` does not handle changes in subscriptions gracefully, particularly for user script topics.
  a. **Solution:** If topics change while we're loading data, we return from `load()` and wait until the next iteration to continue.

### DIFF
--- a/packages/studio-base/src/players/IterablePlayer/BlockLoader.ts
+++ b/packages/studio-base/src/players/IterablePlayer/BlockLoader.ts
@@ -97,7 +97,16 @@ export class BlockLoader {
     // Update all the blocks with any missing topics
     for (const block of this.#blocks) {
       if (block) {
-        block.needTopics = new Map(topics);
+        const blockTopics = Object.keys(block.messagesByTopic);
+        const needTopics = new Map(topics);
+        for (const topic of blockTopics) {
+          // We need the topic unless the subscription is identical to the subscription for this
+          // topic at the time the block was loaded.
+          if (this.#topics.get(topic) === topics.get(topic)) {
+            needTopics.delete(topic);
+          }
+        }
+        block.needTopics = needTopics;
       }
     }
 
@@ -249,12 +258,15 @@ export class BlockLoader {
         const untilTime = clampTime(this.#blockIdToEndTime(currentBlockId), this.#start, this.#end);
 
         const results = await cursor.readUntil(untilTime);
-        if (!R.equals(topics, this.#topics)) {
-          return;
-        }
         // No results means cursor aborted or eof
         if (!results) {
           await cursor.end();
+          return;
+        }
+
+        // When topics change while loading, we need to wait for the next
+        // iteration
+        if (!R.equals(topics, this.#topics)) {
           return;
         }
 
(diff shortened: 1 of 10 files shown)


defunctzombie:
Could you elaborate in the PR description _how_ this PR fixes the above? Or at least the notable bits of those fixes.

cfoust:
I added short summaries in the PR description. The diff is also rather short; let me know if there are any confusing bits.

jtbandes:
> a. Solution: Now we hold on to incoming block data in anticipation of the registration of a client.

Does this make some of the timing changes I'd added to fix broken stories unnecessary?

cfoust:
it does indeed, i stand corrected

jtbandes:
> a. Solution: Empty message arrays now indicate an empty block.

@defunctzombie probably knows more about this but I believe there was an intentional distinction between undefined and [] in the block loader. Something about knowing where we have gaps in the loaded data and that we should not draw a line connecting two plot points if there is a gap in the blocks?

cfoust:
I could be wrong but I don't think that's true. If it is, there's a bug in the block processor, because that violates an assumption I made (and I have never seen that behavior.)

jtbandes:
This is what I was talking about: https://github.com/foxglove/studio/blob/f9d74dccd829f0831256bfdc2dc7c9afa4e877e2/packages/studio-base/src/players/types.ts#L267

cfoust:
Yup, now that I had to spend a while debugging this I'm reasonably familiar with that portion of the code (to say nothing of the downstream use of blocks), and I don't think the `BlockLoader` uses `[]` to represent anything

packages/studio-base/src/players/IterablePlayer/BlockLoader.ts
```diff
@@ -263,6 +264,12 @@ export class BlockLoader {
           return;
         }
 
+        // When topics change while loading, we need to wait for the next
+        // iteration
```
jtbandes, line 268: Next iteration of what? When thinking of "next iteration" I would expect `continue` rather than `return` ... and _why_ do we need to wait for the next iteration?

jtbandes (approved):

packages/studio-base/src/players/IterablePlayer/BlockLoader.ts
cfoust: The next iteration of [this loop](https://github.com/foxglove/studio/blob/0cee7008df16b8ded249c948feb507d26f04e601/packages/studio-base/src/players/IterablePlayer/BlockLoader.ts#L163). We cannot be sure what work we calculated at the beginning is still accurate, so we just wait for the next "iteration" entirely

PR #7048, opened 2023-10-31
Convert UserScriptStateContext to zustand store

defunctzombie:
**User-Facing Changes**
Performance improvements when scrubbing or seeking and user scripts are present.

**Description**
When scrubbing or seeking, the user scripts are reset which causes the `UserScriptStateContext` _state_ to get updated. Because the entire context is not selectable this invalidates the _value_ of the context which causes it to render and any downstream users of the context to re-render. This re-renders a lot of things downstream that don't need to render because they only need the stable actions functions and not the state.

This change updates the context to use zustand and allow for separately selecting the actions and the state.

---

Here is a before and after showing scrubbing behavior with a layout that has some user scripts. In this recording I did not have sidebars open. The effect is even more dramatic if I open sidebars.

Before:

https://github.com/foxglove/studio/assets/84792/771c070f-ecbc-44ec-98a9-8fcafa9d8165

After:

https://github.com/foxglove/studio/assets/84792/200e7d3a-339d-4547-9f02-6e0d2fcaa451

---
I think there's a poor behavior here where a user script that _logs_ could eat up memory if no one is viewing the logs but have not confirmed that.

### DIFF
--- a/packages/studio-base/src/StudioApp.tsx
+++ b/packages/studio-base/src/StudioApp.tsx
@@ -55,10 +55,10 @@ export function StudioApp(): JSX.Element {
   const providers = [
     /* eslint-disable react/jsx-key */
     <TimelineInteractionStateProvider />,
-    <UserScriptStateProvider />,
     <CurrentLayoutProvider />,
     <ExtensionMarketplaceProvider />,
     <ExtensionCatalogProvider loaders={extensionLoaders} />,
+    <UserScriptStateProvider />,
     <PlayerManager playerSources={dataSources} />,
     <EventsProvider />,
     /* eslint-enable react/jsx-key */
--- a/packages/studio-base/src/components/PlayerManager.tsx
+++ b/packages/studio-base/src/components/PlayerManager.tsx
@@ -22,7 +22,7 @@ import {
 } from "react";
 import { useLatest, useMountedState } from "react-use";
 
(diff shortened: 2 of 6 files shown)


snosenzo:
> When scrubbing or seeking, the user scripts are reset 

Why is this necessary to begin with?

packages/studio-base/src/StudioApp.tsx
```diff
@@ -55,10 +55,10 @@ export function StudioApp(): JSX.Element {
   const providers = [
     /* eslint-disable react/jsx-key */
     <TimelineInteractionStateProvider />,
-    <UserScriptStateProvider />,
     <CurrentLayoutProvider />,
     <ExtensionMarketplaceProvider />,
     <ExtensionCatalogProvider loaders={extensionLoaders} />,
+    <UserScriptStateProvider />,
```
jtbandes, line 61: Why did this move?

jtbandes (approved):

packages/studio-base/src/StudioApp.tsx
defunctzombie: The other providers don't need the UserScriptState provider so I moved it to contain its re-render scope. Not strictly necessary with the new zustand changes.

defunctzombie:
> Why is this necessary to begin with?

I miss-spoke. I don't think they are reset in the complete sense but something in their state is cleared (possibly the logs).

PR #7050, opened 2023-10-31
Move Playback Performance panel into the right sidebar

jtbandes:
**User-Facing Changes**
Moved the Playback Performance debugging panel into the right sidebar.

**Description**
- Renamed "Studio debug panels" experimental feature to "Debug mode"
- Renamed "Studio Logs" sidebar to "Logs"
- Added a "Performance" tab to the sidebar that contains the PlaybackPerformance component

image


snosenzo (approved):
Looks good, just wondering if we need any kind of migration for this if people have this panel in their layouts? 
All i see is this: 
image

jtbandes:
I don't feel strongly about a migration path – this has always been a debug panel behind an experimental feature flag.

PR #7055, opened 2023-11-01
Use a more specific workspace selector in AppBar

defunctzombie:
**User-Facing Changes**
Not worth calling out individually; could group into a small performance improvement item.

**Description**
When selecting from the workspace context using a selector, it is important
to avoid having the selector value change if the value you want to select is not changing. This helps reduce
renders of unrelated components when other aspects of the selectable context change.

This change updates the AppBar and AppMenu selection of the left/right sidebar open state to be more stable
thus avoiding re-rendering those components when other aspects of the workspace context change like sidebar sizes.


wimagguc (approved):

snosenzo (approved):
Should we remove the shallow selector entirely in favor of more specific selectors?

defunctzombie:
> Should we remove the shallow selector entirely in favor of more specific selectors?

I did that in the PR.

PR #7056, opened 2023-11-01
Move state transition open sibling function to separate file

defunctzombie:
**User-Facing Changes**
None

**Description**

Better separation of the panel and all the imports for the panel and the open sibling function which does
not need all the logic of the panel. This avoids bringing in the imports of the StateTransition panel into other panels and since this
logic is not specific to anything else in the panel I find it can be nicely separated.


packages/studio-base/src/panels/StateTransitions/openSibling.ts
jtbandes, line 1: Would be nice for the filename to match the export, e.g. `openSiblingStateTransitionsPanel.ts`

jtbandes (approved):

PR #7058, opened 2023-11-01
Inject Performance sidebar and memory use indicator from AppContext

jtbandes:
**User-Facing Changes**
None

**Description**
Removes the PlaybackPerformance component (see #7050) and MemoryUseIndicator, and replaces it with a component injected from AppContext.

Also moving around some utility functions/types that were used by the PlaybackPerformance component.


packages/den/format/formatByteSize.ts
```diff
@@ -0,0 +1,14 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+export function formatByteSize(size: number): string {
```
snosenzo, line 5: amazing -- surprised we didn't have something like this earlier.

snosenzo (approved):

packages/den/format/formatByteSize.ts
jtbandes: We did! It moved from the open source repo to the private repo and is now moving back

PR #7059, opened 2023-11-01
Set maximum byte size on messages in current frame in websocket player

snosenzo:
**User-Facing Changes**

- Mitigates crashing on returning to inactive tabs that were viewing live data by dropping messages after a certain size limit

**Description**

When the tab is inactive `setTimeout`'s are throttled to at most once per second. Because the MessagePipeline listener uses timeouts to resolve its promises, it throttles our ability to emit a frame more than once per second. In the websocket player this was causing an accumulation of messages that were waiting to be emitted, this could keep growing indefinitely if the rate at which we emit a frame is low enough for inactive tabs. To mitigate this, I've incorporated a maximum cache size for single frames in the WebSocketPlayer. Once it reaches this limit it will drop the oldest messages until it comes down to 80% cache capacity.

Here is an example of the number of messages that can appear in a single player emit while the tab is inactive. Normally while this is running in an active tab it is 1 message per frame (this is a single pointcloud topic at ~30hz)
image

Fixes: FG-4666
Fixes: FG-4578

### DIFF
--- a/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
+++ b/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
@@ -78,7 +78,14 @@ type ResolvedService = {
 };
 type MessageDefinitionMap = Map<string, MessageDefinition>;
 
-// 1500MB
+/**
+ * When the tab is inactive setTimeout's are throttled to at most once per second.
+ * Because the MessagePipeline listener uses timeouts to resolve its promises, it throttles our ability to
+ * emit a frame more than once per second. In the websocket player this was causing
+ * an accumulation of messages that were waiting to be emitted, this could keep growing
+ * indefinitely if the rate at which we emit a frame is low enough.
+ * 1500MB
+ */
 const CURRENT_FRAME_MAXIMUM_SIZE_BYTES = 15e8;
 
 export default class FoxgloveWebSocketPlayer implements Player {
--- a/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
+++ b/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
@@ -78,8 +78,8 @@ type ResolvedService = {
 };
 type MessageDefinitionMap = Map<string, MessageDefinition>;
 
-// 500MB
-const CURRENT_FRAME_MAXIMUM_SIZE_BYTES = 5.0e8;
+// 1500MB
+const CURRENT_FRAME_MAXIMUM_SIZE_BYTES = 15e8;
 
 export default class FoxgloveWebSocketPlayer implements Player {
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
```diff
@@ -524,6 +528,28 @@ export default class FoxgloveWebSocketPlayer implements Player {
...
+        this.#parsedMessagesBytes += data.byteLength;
+        if (this.#parsedMessagesBytes > CURRENT_FRAME_MAXIMUM_SIZE_BYTES) {
+          this.#problems.addProblem(`webSocketPlayer:parsedMessageCacheFull`, {
+            severity: "error",
+            message: `WebSocketPlayer maximum frame size (${(
+              CURRENT_FRAME_MAXIMUM_SIZE_BYTES / 1_000_000
+            ).toFixed(
+              2,
+            )}MB) reached. Dropping old messages. This accumulation can occur if the browser tab has been inactive.`,
+          });
```
achim-k, line 540: Should we clear this problem at some point again?

achim-k (commented):
> Because the MessagePipeline listener uses timeouts to resolve its promises, it throttles our ability to emit a frame more than once per second.

Is the problem really on the websocket player side if the player emits all messages at least once per second? It's hard to imagine that within one second the websocket player receives so many messages that would cause a OOM crash.

snosenzo:
> Is the problem really on the websocket player side if the player emits all messages at least once per second? It's hard to imagine that within one second the websocket player receives so many messages that would cause a OOM crash.

It emits all messages at _most_ once per second according to [this](https://stackoverflow.com/questions/6032429/chrome-timeouts-interval-suspended-in-background-tabs). With some simple point cloud test data I see this evicting multiple times within a single emit

packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
snosenzo: I'm not sure when a good time to clear this error would be though.

achim-k (approved):

packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
```diff
@@ -78,6 +78,9 @@ type ResolvedService = {
 };
 type MessageDefinitionMap = Map<string, MessageDefinition>;
 
+// 1500MB
+const CURRENT_FRAME_MAXIMUM_SIZE_BYTES = 15e8;
```
defunctzombie, line 82: Please copy parts of your PR description as a code comment here. It is helpful when reading the code to understand the history and _why_ we are doing this.

defunctzombie (approved):

Issue #7007, opened 2023-10-20
Flickering bugs for GLTF models


henrik-kaijser:
**Description**
I have implemented a message converter that produces a height map as a GLTF in a ModelPrimitive message, which is sent in a SceneUpdate Message. When the model is displayed in the studio 3D panel there's substantial flickering. 

- Version: latest studio
- Platform: ubuntu 22.04
- Data source: mcap

**Studio Code Analysis**
The `#updateModels` method in `ThreeDeeRender/renderables/primitives/RenderableModels.ts` begins by doing
`this.clear() `
which wipes the currently loaded model from Three. Then it starts loading the next version of the model  asynchronously. Since this may take quite some time, a new frame that is rendered before the model has been loaded will this not contain the model, which causes flickering.
Similarly, after the model has been successfully loaded towards the end of  `#updateModels` it gets rid of the model by doing
```
renderable.model.removeFromParent();
this.#disposeModel(renderable);
```
which creates a very similar problem.

**Workaround**  
As a quick hack I removed the three lines listed above, and instead cleared the previous model version after the new version has been loaded, but before
_this.add(renderable.model);_ 
This fixes the problem but you may want to fix it in some other way.

defunctzombie:
Internal tracking ticket: https://linear.app/foxglove/issue/FG-5384/flickering-bugs-for-gltf-models

snosenzo:
The lines that are disposing of the renderable in the `.finally` should be kept to dispose of unused previous renderables. The renderables that are used across updates should be taken out of that `prevRenderables` list by the time it gets to the finally. I moved some code around and renamed some things to hopefully make that a bit more clear. 

I tested with some sample data of my own and didn't have a flicker. Let me know if this fix works for your case.

PR #7060, opened 2023-11-02
Fix ModelPrimitive flickering

snosenzo:
**User-Facing Changes**

- Fix ModelPrimitive flickering

**Description**

Updated to do as much before the animation frame as possible. So now old renderables are cleared or kept before the await loop for loading new renderables.

Below are videos of stepping through animation frames to show that there is less flickering.

Before:

https://github.com/foxglove/studio/assets/10187776/ee84b955-1310-4aa4-9415-04df161e5d70

After:

https://github.com/foxglove/studio/assets/10187776/e0aba817-3fa4-475b-ae2e-a8d69e1b350b

Sample data file: [flamingo.mcap.zip](https://github.com/foxglove/studio/files/13284693/flamingo.mcap.zip)

Fixes: FG-5363
Fixes:  https://github.com/foxglove/studio/issues/7007


packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableModels.ts
```diff
@@ -55,48 +55,60 @@ export class RenderableModels extends RenderablePrimitive {
...
+   * Uses matching function to find, remove and return the first renderable from the list that matches
+   * @param renderables - (MUTABLE) list of RenderableModel objects
+   * @param primitivesMatch - Comparison function that returns true if the two primitives are a match
+   * @param primitive - Primitive to match against
+   * @returns - matching renderable if found, otherwise undefined
+   */
+  #removeMatchFromList(
+    renderables: RenderableModel[] | undefined,
+    primitivesMatch: (a: ModelPrimitive, b: ModelPrimitive) => boolean,
+    primitive: ModelPrimitive,
```
defunctzombie, line 94: Does this parameter become _a_ or _b_ in the match function? If the caller knows the primitive already why wouldn't they make a match function that closes over it rather than use the parameter?

Separately - I would put the match function parameter after the `primitive` parameter argument. having the function param last feels more natural to me with other APIs I've seen.

packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableModels.ts
```diff
@@ -55,48 +55,60 @@ export class RenderableModels extends RenderablePrimitive {
...
+  ) {
+    let renderable: RenderableModel | undefined;
+    if (renderables) {
+      const idx = renderables.findIndex((prev) => primitivesMatch(prev.primitive, primitive));
+      if (idx >= 0) {
+        // remove from previous renderables so that it doesn't get disposed
+        renderable = renderables.splice(idx, 1)[0]!;
+      }
+    }
+    return renderable;
```
defunctzombie, line 104: I think this is the same logic but drops the intermediate variable.
```
    if (renderables) {
      const idx = renderables.findIndex((prev) => primitivesMatch(prev.primitive, primitive));
      if (idx >= 0) {
        // remove from previous renderables so that it doesn't get disposed
        return renderables.splice(idx, 1)[0];
      }
    }
    return undefined;
```

packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableModels.ts
```diff
@@ -332,3 +365,9 @@ function cloneAndPrepareModel(cachedModel: LoadedModel) {
   removeLights(model);
   return new THREE.Group().add(model);
 }
+
+const crcPrimitivesMatch = (model1: ModelPrimitive, model2: ModelPrimitive) =>
```
defunctzombie, line 369: comments on these helper functions so we can know what they are intending to do.

defunctzombie (approved):

packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableModels.ts
snosenzo: updated to pass single `isMatch` function with `primitive` being closed within

packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableModels.ts
```diff
@@ -183,22 +216,24 @@ export class RenderableModels extends RenderablePrimitive {
...
-            this.#disposeModel(renderable);
-          }
-        }
-        for (const renderables of prevRenderablesByDataCrc.values()) {
-          for (const renderable of renderables) {
-            renderable.model.removeFromParent();
-            this.#disposeModel(renderable);
-          }
-        }
         this.#updateOutlineVisibility();
```
jtbandes, line 219: seems like we have two of these now – intentional?

packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableModels.ts
```diff
@@ -332,3 +367,11 @@ function cloneAndPrepareModel(cachedModel: LoadedModel) {
   removeLights(model);
   return new THREE.Group().add(model);
 }
+
+/** Used to check that crc-data primitives are using the same model data */
+const crcPrimitivesMatch = (model1: ModelPrimitive, model2: ModelPrimitive) =>
```
jtbandes, line 372: I find this name slightly misleading because the function is comparing the data byte-for-byte, not by crc. But I see it is being called a "crc primitive" because it's one that we store keyed by its crc.

packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableModels.ts
snosenzo: yes, intentional. Commented here: https://github.com/foxglove/studio/pull/7101

packages/studio-base/src/panels/ThreeDeeRender/renderables/primitives/RenderableModels.ts
snosenzo: Good call. updated here: https://github.com/foxglove/studio/pull/7101

PR #7061, opened 2023-11-02
Add state transition panel setting to hide intermediate values

defunctzombie:
**User-Facing Changes**

**Description**

### DIFF
--- a/packages/studio-base/src/panels/StateTransitions/messagesToDatasets.ts
+++ b/packages/studio-base/src/panels/StateTransitions/messagesToDatasets.ts
@@ -16,28 +16,13 @@ import { StateTransitionPath } from "./types";
 
 const baseColors = [grey, ...expandedLineColors];
 
-/**
- * Returns a function that can be used to assign unique indexes to distinct values. Returns the
- * previous index if the value has been seen before otherwise a new index.
- */
-function makeValueIndexer() {
-  const seenValues = new Map<unknown, number>();
-  return (val: unknown) => {
-    const seen = seenValues.get(val);
-    if (seen != undefined) {
-      return seen;
-    }
-    seenValues.set(val, seenValues.size);
-    return seenValues.size - 1;
-  };
-}
-
 type Args = {
   path: StateTransitionPath;
   startTime: Time;
   y: number;
   pathIndex: number;
   blocks: readonly (readonly MessageAndData[] | undefined)[];
+  showIntermediate: boolean;
 };
 
 /**
@@ -45,7 +30,7 @@ type Args = {
  * dataset with different labels and colors applied per-point.
  */
 export default function messagesToDatasets(args: Args): ChartDatasets {
-  const { path, startTime, y, blocks } = args;
+  const { path, startTime, y, blocks, showIntermediate } = args;
 
   const dataset: ChartDataset = {
     borderWidth: 10,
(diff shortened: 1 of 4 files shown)


amacneil:
- color state to end of recording

options:
- invert setting to "Show intermediate", off by default (i.e. performance by default)
- remove setting, show points based on heuristic (e.g. max half as many points as there are horizontal pixels)

defunctzombie:
This PR is ready for another round of product review and can get its first round of code review.

Regarding the first round feedback feedback; I've punted "color state to end of recording". I've decided that is better served by a separate follow-up discussion that does not block this change. Instead - I have updated this change to include the last input state value in the output state even if it is the same as before. This avoids "cutting off" the last bar prematurely since the last bar will always have the same state throughout.

Tooltips are always shown (even when between datums) as of: https://github.com/foxglove/studio/pull/7086

I have opted to avoid the heuristic approach in this PR to provide this incremental quality-of-life improvement and gather feedback. The heuristic approach (and doing more dataset truncation) will take more time to implement. The issue is more than hiding the points. Yes the points are slower to draw than not having the points - but even without drawing the points we have to avoid creating a large dataset for the worker architecture we currently have for this panel. That downsampling work will take more thinking.

packages/studio-base/src/panels/StateTransitions/messagesToDatasets.ts
```diff
@@ -15,37 +15,23 @@ import positiveModulo from "./positiveModulo";
...
+  showIntermediate: boolean;
 };
 
 /**
  * Processes messages into datasets. For performance reasons all values are condensed into a single
  * dataset with different labels and colors applied per-point.
  */
-export default function messagesToDatasets(args: Args): ChartDatasets {
-  const { path, startTime, y, blocks } = args;
+export default function messagesToDatasets(args: Args): ChartDataset {
```
snosenzo, line 33: Can we update the function name and comment to have singular `Dataset` as well?

packages/studio-base/src/panels/StateTransitions/messagesToDatasets.ts
```diff
@@ -82,18 +66,6 @@ export default function messagesToDatasets(args: Args): ChartDatasets {
 
       const { constantName, value } = queriedData;
 
-      const datasetIndex = indexer(value);
-
-      // Skip duplicates.
```
snosenzo, line 87: Why did we skip duplicates previously? and why don't we need to now?

snosenzo (approved):
code looks good to me. Might want another person to look it over who has more context on charts though

packages/studio-base/src/panels/StateTransitions/messagesToDatasets.ts
defunctzombie: I tried to think about why and could not figure it out nor come up with a reason for it. Maybe an optimization? @jtbandes do you recall?

jtbandes: I do not know – seems to have been around since webviz days. Maybe it caused some chartjs glitch? Or maybe it was a premature optimization? The only way to know for sure is 🔪 and find out

amacneil:
how much additional work is it to add the heuristic instead of toggle? it feels a bit weird adding a setting that we plan to delete, and it feels weird having a setting that makes the app explode.

defunctzombie:
Changed the setting to "Show points" and updated tooltip and behavior based on discussion that decided a more sensible behavior is to always hide all the points and have a setting to toggle points.

PR #7062, opened 2023-11-02
Style headings like console and fix up dialog titles

2metres:
**User-Facing Changes**
None

**Description**
Restyle typography to be more like console and fix up issues with DialogTItle

### DIFF
--- a/packages/studio-base/src/components/ShareJsonModal.tsx
+++ b/packages/studio-base/src/components/ShareJsonModal.tsx
@@ -10,8 +10,8 @@ import {
   Dialog,
   DialogContent,
   DialogActions,
+  DialogTitle,
   TextField,
-  Typography,
   outlinedInputClasses,
 } from "@mui/material";
 import { useCallback, useMemo, useState } from "react";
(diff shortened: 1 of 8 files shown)


wimagguc (approved):

packages/studio-base/src/components/ShareJsonModal.tsx
```diff
@@ -39,6 +39,11 @@ const useStyles = makeStyles()((theme) => ({
       padding: theme.spacing(0.25),
     },
   },
+  dialogTitle: {
+    display: "flex",
+    justifyContent: "space-between",
+    alignItems: "center",
+  },
```
jtbandes, line 46: what's the reason for these overrides?

jtbandes (approved):

packages/studio-base/src/components/ShareJsonModal.tsx
2metres: In places where we want to have a close button present we need to change the styles on the dialog title. Across the app we were doing it in various different ways. This was me attempting to make it consistent

PR #7064, opened 2023-11-02
[3D Panel] Process messages on Render frame and skip processing messages that will be overriden

snosenzo:
**User-Facing Changes**

 - 3D performance improved by skipping the processing of some unecessary messages

**Description**

Added a `queue` and `processQueue` fields to the `RendererSubscription`. `addMessages` now adds messages to that subscription queue, and then on `handleFrame` we process and clear the queue. `processQueue` lets extensions opt into filtering the queue to only take the last message per topic, or apply whatever SceneExtension specific logic that will skip processing unecessary messages.

Only updated some of the SceneExtensions that this would be relevant for, I can update more if we settle on this as the way we want to do this.
I didn't update `ImageMode` because I didn't want to affect video yet. I'll do that in a separate PR

Newer version of https://github.com/foxglove/studio/pull/6626

alternate approach here: https://github.com/foxglove/studio/pull/7079


packages/studio-base/src/panels/ThreeDeeRender/SceneExtension.ts
```diff
@@ -241,3 +241,13 @@ export class SceneExtension<
     }
   }
 }
+
+export function onlyLastByTopicMessage<T>(msgs: MessageEvent<T>[]): MessageEvent<T>[] {
```
defunctzombie, line 245: Please add a comment on what this is suppose to do or be used for.

defunctzombie (commented):
Some thoughts as I figure out what the logic is doing:

1. "processMessageQueue" seems more like a "filterQueue" function based on how I am seeing it used.
2. There seems to be another layer of topic grouping that happens in each extension in the proposed process functions. Why is that? Isn't an extension created per topic? This grouping then flattens the list so seems like pure overhead?

My mostly un-educated sense is that I see why the queueMessage function exists - to avoid calling the handle logic for each extension until after you've gotten a sense of the messages to send to each extension. Where I think it could be simpler is the queueMessage function could invoke a `shouldQueue` function on the extension which would return whether messages on that extension should queue or whether it should be dropped (looking at how you've handled the decay time).

Another option is to queue up the messages as you are doing and then update the handle functions of the extensions to accept a batch of messages for the extension rather than one at a time? The extension logic would decide if it will skip items from the incoming batch?

defunctzombie:
> @defunctzombie
> 
> > "processMessageQueue" seems more like a "filterQueue" function based on how I am seeing it used.
> 
> I chose a wider scoped name to allow for more flexibility. Can rename to `filterQueue` if it's not as clear as it could be.
> 
> > There seems to be another layer of topic grouping that happens in each extension in the proposed process functions. Why is that? Isn't an extension created per topic? This grouping then flattens the list so seems like pure overhead?
> 
> An extension is not created per topic. The grouping by topic in the process function is because we create/update _renderables_ by topic in most cases.
> 
> > Another option is to queue up the messages as you are doing and then update the handle functions of the extensions to accept a batch of messages for the extension rather than one at a time? The extension logic would decide if it will skip items from the incoming batch?
> 
> I thought about that, but decided against it because I didn't want to encourage message handling functions with added responsibility and thought it might be cleaner to have it as an explicit subscription step and to not change the message handlers, but maybe it's more confusing to go back and forth with the subscription logic calling a `processQueue` function. Also combining this inside of a `handleQueue` function could let us consolidate this separating by topics into a single part of the pipeline, so maybe that's the better choice. It will require me to update every SceneExtension though :(

Replied in slack with my thoughts. Seems like we will try the handleQueue approach or similar to avoid this layer of indirection.

packages/studio-base/src/panels/ThreeDeeRender/renderables/PointClouds.ts
```diff
@@ -707,16 +708,47 @@ export class PointClouds extends SceneExtension<PointCloudHistoryRenderable> {
...
+        },
       },
     ];
   }
 
+  #processMessageQueue<T>(msgs: MessageEvent<T>[]): MessageEvent<T>[] {
+    if (msgs.length === 0) {
+      return msgs;
+    }
+    const msgsByTopic = _.groupBy(msgs, (msg) => msg.topic);
```
defunctzombie, line 731: Did something else previously group these by topic? Its not clear to me why the new logic requires grouping by topic before processing.

snosenzo: the grouping by topic happens when we update/create a renderable associated with the topic from the `renderables` map in the scene extension

snosenzo: I do agree that it is a hidden coupling here. I will add a comment. If we stored renderables by topic + schemaname or some other key this would need to be updated as well.

packages/studio-base/src/panels/ThreeDeeRender/IRenderer.ts
```diff
@@ -176,6 +176,10 @@ export type RendererSubscription<T = unknown> = {
   shouldSubscribe?: (topic: string) => boolean;
   /** Callback that will be fired for each matching incoming message */
   handler: (messageEvent: MessageEvent<T>) => void;
+  /** Queue of messages to be handled since last frame. Will be reassigned to new empty array each frame. */
+  queue?: MessageEvent<T>[] | undefined;
```
jtbandes, line 180: I find it strange to have a mutable array as part of the subscribe payload

jtbandes (approved):

PR #7065, opened 2023-11-03
Use a fixed fallback window for state transition

defunctzombie:
**User-Facing Changes**
When loading a pre-recorded data-source, the state transition panel no longer "scrolls" the end position on the chart and instead uses the data source end time. This keeps the chart end position stable similar to the plot panel and the transition data loads from the start.

Before:

https://github.com/foxglove/studio/assets/84792/910986eb-55cd-4567-8943-b494f9591cf3

After:

https://github.com/foxglove/studio/assets/84792/3e67df09-3a3a-4c63-889d-93a78a540ab4

**Description**
If there is no configured range or min/max for the x-axis, set the x-axis bounds to `0, end-time`. This avoids the chart constantly re-computing the end time when new data arrives.

I think this behavior is also ok for live data sources since the end time constantly updates. Having some tire kicking on that would be nice.

### DIFF
--- a/packages/studio-base/src/panels/StateTransitions/index.tsx
+++ b/packages/studio-base/src/panels/StateTransitions/index.tsx
@@ -326,9 +326,18 @@ const StateTransitions = React.memo(function StateTransitions(props: Props) {
         x: { min: currentTimeSinceStart - config.xAxisRange, max: currentTimeSinceStart },
         y: { min: Number.MIN_SAFE_INTEGER, max: Number.MAX_SAFE_INTEGER },
       };
-    } else {
-      return undefined;
+    } else if (endTimeSinceStart != undefined) {
+      // If we have no configured xAxis min/max or range, then we set the x axis max to end time
+      // This will mirror the plot behavior of showing the full x-axis for data time range rather
+      // than constantly adjusting the end time to the latest loaded state transition while data
+      // is loading.
+      return {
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/panels/StateTransitions/index.tsx
```diff
@@ -326,9 +326,18 @@ const StateTransitions = React.memo(function StateTransitions(props: Props) {
...
       };
-    } else {
-      return undefined;
+    } else if (endTimeSinceStart != undefined) {
+      // If we have no configured xAxis min/max or range, then we set the x axis max to end time
+      // This will mirror the plot behavior of showing the full x-axis for data time range rather
+      // than constantly adjusting the end time to the latest loaded state transition while data
+      // is loading.
+      return {
+        x: { min: 0, max: endTimeSinceStart },
```
jtbandes, line 335: how do we know `0` is a correct min value here? Is it relative to the start time of the data source?

jtbandes (approved):
Any way we could add a story for this?

defunctzombie:
> Any way we could add a story for this?

🤔 yea - I think we can. Will add to this PR or note why it was impossible.

defunctzombie:
Actually - the stories have changed because of the new logic. I will review the changed stories to confirm they appear with the correct new intent.

PR #7070, opened 2023-11-06
Fix SceneEntity primitives from disappearing / only showing outlines after seeking

snosenzo:
**User-Facing Changes**

 - Fix SceneEntity primitives from disappearing / only showing outlines after seeking

**Description**

The issue here is that the primitive Renderables are using outdated/incorrect bounding volumes, which is causing them to be culled when they shouldn't be. 

In more detail, after seeking primitive renderables are released to the `PrimitivePool`, and then re-acquired on the next message. These renderables are `acquired` in reverse of the order that they were `released` from the pool. This caused entities to get renderables that were formerly part of other entities, which isn't inherently wrong except for the fact that we don't recalculate the bounding sphere on the mesh being used by the primitive, so it is being culled based off of an old bounding sphere. This PR sets `frustumCulled` on `InstancedMesh` to be false. 
This issue came from a change that happened in threejs between v150 and v151 that makes the `InstancedMesh.frustumCulled` default to true. 

Fixes: FG-5211
Fixes: FG-5460

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/DynamicInstancedMesh.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/DynamicInstancedMesh.ts
@@ -25,6 +25,7 @@ export class DynamicInstancedMesh<
     super(geometry, material, 0);
 
     this.#capacity = initialCapacity;
+    this.frustumCulled = false;
     this.#resize();
   }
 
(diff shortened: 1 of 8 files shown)


jtbandes (approved):
Approving [e2087a7](https://github.com/foxglove/studio/pull/7070/commits/e2087a795ff049262434c4fff41d0a15ec345f49) 😄

PR #7074, opened 2023-11-07
Re-run topic alias functions when global variables change

defunctzombie:
**User-Facing Changes**
Fix an issue when using topic aliasing extensions that rely on global variables not updating the list of topics.

**Description**
This change updates the `TopicAliasingPlayer` to reprocess the last player state when global variables lead to new topic aliases. Prior to this change, updating the global variables while paused would not produce any feedback to the UI that the alias functions had run. It would only update the state when new state was emitted from the wrapped player (i.e. during playback or preloading).

This change overhauls the internal architecture of `TopicAliasingPlayer`. Prior to this change, there was a single state processing entry point `aliasPlayerState` which was responsible for creating a set of aliases and then applying those to the player state. I have changed this to separate the creating of aliases and the applying into two distinct workflows each with their own memoization. The processing of alias functions (by providing topics and global variables as inputs) is now the responsibility of the `StateProcessorFactory`. Instances of this class memoize the previously processed aliases and avoid making new processors if the set of aliases is unchanged. This is important because `setGlobalVariables` can get called often while topic alias functions may not use any of the updated variables. In these instances we do not want to be invaliding our state processing or re-computing state modifications.

The second class is `StateProcessor`. This instance is responsible for applying a set of aliases to player state and subscriptions. It manages memoization on the various fields of player state to avoid re-processing those if they are unchanged in the player state.

With these two classes, the `TopicAliasingPlayer` can opportunistically evaluate changes to alias inputs and determine if they produced a new `StateProcessor`. If there is a new state processor (because aliases changed), then we can re-process the previous player state.

Memoization which used to be global is now isolated to instances of the two classes specific for their workflows.


packages/studio-base/src/players/TopicAliasingPlayer/StateProcessorFactory.ts
```diff
@@ -0,0 +1,106 @@
...
+      aliases: mapper.aliasFunction({
+        topics: input.topics ?? [],
+        globalVariables: input.variables,
+      }),
+    }));
+
+    const anyMappings = mappings.some((map) => [...map.aliases].length > 0);
+    if (!anyMappings) {
+      this.#aliases = new Map();
+      return (this.#stateProcessor = new NoopStateProcessor());
```
jtbandes, line 53: would this happen repeatedly (and cause downstream to re-run the state processor since it's a new instance) since it is not guarded by the isEqual?

packages/studio-base/src/players/TopicAliasingPlayer/TopicAliasingPlayer.ts
```diff
@@ -33,15 +37,23 @@ import {
...
 
   // True if no aliases are active and we can pass calls directly through to the
   // underlying player.
   #skipAliasing: boolean;
 
-  #listener?: (state: PlayerState) => Promise<void>;
+  #stateProcessorFactory: StateProcessorFactory = new StateProcessorFactory();
+  #stateProcessor: IStateProcessor = new NoopStateProcessor();
+
+  #lastPlayerstate?: PlayerState;
```
jtbandes, line 51:
```suggestion
  #lastPlayerState?: PlayerState;
```

packages/studio-base/src/players/TopicAliasingPlayer/TopicAliasingPlayer.ts
```diff
@@ -135,7 +148,32 @@ export class TopicAliasingPlayer implements Player {
...
+
+    // If we have a new state processor, it means something about the aliases has changed and we
+    // need to re-process the existing player state
+    const shouldReprocess = stateProcessor !== this.#stateProcessor;
+    this.#stateProcessor = stateProcessor;
+
+    // Re-process the last player state if the processor has changed since we might have new downstream topics
+    // for panels to subscribe or get new re-mapped messages.
+    //
+    // Skip this if we are playing and allow the next player state update to handle this.
```
jtbandes, line 173: is this important or just a small optimization?

defunctzombie: Added more notes

PR #7075, opened 2023-11-07
Fix some Tab panel bugs

jtbandes:
**User-Facing Changes**
Fixed issues where splitting tab panels would sometimes result in panels losing their config, or multiple panels sharing the same config.

**Description**
- The logic that "clones" panels residing in tabs when the tab is split lives in `getSaveConfigsPayloadForAddedPanel`. This function was only being invoked when the panel being split was not inside a tab layout. It needs to be invoked regardless of whether the panel being split is in nested in another tab.
- When cloning panels while splitting, the `relatedConfigs` was being used as a source for the newly cloned panels' configs. This `relatedConfigs` was a subset of the overall `configById`, which led to newly cloned panels sometimes having no config (because it could not be found in `relatedConfigs`). I changed this to pass in the whole `configById`.
- Removed the `relatedConfigs` concept entirely since we do not use this in the app (it was related to a defunct "presets" feature).
- Removed some test helpers which were no longer used.


packages/studio-base/src/providers/CurrentLayoutProvider/reducers.ts
```diff
@@ -205,36 +202,19 @@ const splitPanel = (
...
-      newPanelsState,
-      getSaveConfigsPayloadForAddedPanel({ id: newId, config, relatedConfigs }),
-    );
   }
+
+  // Save the new panel's config and clone any panels in tabs if necessary
+  newPanelsState = savePanelConfigs(
+    newPanelsState,
+    getSaveConfigsPayloadForAddedPanel({ id: newId, config, savedProps }),
+  );
```
jtbandes, line 211: Moving this outside the `if` and passing in `savedProps` instead of `relatedConfigs` is the main bug fix.

packages/studio-base/src/providers/CurrentLayoutProvider/reducers.test.tsx
```diff
@@ -234,37 +228,6 @@ describe("layout reducers", () => {
         foo: "bar",
       });
     });
-    it("drops nested Tab panel into main layout", () => {
```
jtbandes, line 237: No longer relevant since it tested dropping in a pre-configured Tab which already had children (the children were not already present in `configById`). This scenario would only occur with "preset" panels in the panel catalog, which we do not have.

packages/studio-base/src/providers/CurrentLayoutProvider/reducers.test.tsx
```diff
@@ -170,9 +168,7 @@ describe("layout reducers", () => {
       expect(tabs.length).toEqual(1);
       expect(tabs[0]?.title).toEqual("A");
 
-      const newAudioId = tabs[0]?.layout as string;
-      expect(getPanelTypeFromId(newAudioId)).toEqual("Audio");
-      expect(configById[newAudioId]).toEqual({ foo: "baz" });
```
jtbandes, line 175: No longer relevant since it tested dropping in a pre-configured Tab which already had children (the children were not already present in `configById`). This scenario would only occur with "preset" panels in the panel catalog, which we do not have.

defunctzombie (approved):
Magical

PR #7076, opened 2023-11-07
Improve Log message normalization

jtbandes:
**User-Facing Changes**
None

**Description**
The Log panel would crash if it received `timestamp: null`. Improved the normalization functions to catch this case (and some others).

I did not try to go overboard here with normalization because we could spend a ton of effort doing this in all our panels... some day it would be nice to have more centralized schema validation (see FG-4917)

Fixes FG-5578

### DIFF
--- a/packages/studio-base/src/panels/Log/conversion.tsx
+++ b/packages/studio-base/src/panels/Log/conversion.tsx
@@ -3,8 +3,8 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
 import { fromNanoSec } from "@foxglove/rostime";
+import { Log as FoxgloveLog } from "@foxglove/schemas";
 import { Time } from "@foxglove/studio";
-import { FoxgloveMessages } from "@foxglove/studio-base/types/FoxgloveMessages";
 
 import {
   Ros1RosgraphMsgs$Log,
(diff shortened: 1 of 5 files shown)


packages/studio-base/src/panels/Log/conversion.tsx
```diff
@@ -15,23 +15,26 @@ import {
 } from "./types";
 
 // Get the log message string from the log message
-export function getNormalizedMessage(logMessage: LogMessageEvent["message"]): string {
+export function getNormalizedMessage(logMessage: Partial<LogMessageEvent["message"]>): string {
```
defunctzombie, line 18: Could make a type for `Partial<LogMessageEvent["message"]>` in this file since you use it in a few places.

packages/studio-base/src/panels/Log/filterMessages.ts
```diff
@@ -11,6 +11,8 @@
 //   found at http://www.apache.org/licenses/LICENSE-2.0
 //   You may not use this file except in compliance with the License.
 
+import { mightActuallyBePartial } from "@foxglove/studio-base/util/mightActuallyBePartial";
```
defunctzombie, line 14: Should live in den :(

defunctzombie: Also - could this be a type only thing rather than a runtime function?

packages/studio-base/src/panels/Log/conversion.tsx
```diff
@@ -40,17 +43,17 @@ export function getNormalizedLevel(datatype: string, raw: LogMessageEvent["messa
...
     case "foxglove_msgs/msg/Log":
     case "foxglove::Log":
     case "foxglove.Log": {
-      const timestamp = (raw as FoxgloveMessages["foxglove.Log"]).timestamp;
+      const timestamp = (raw as Partial<FoxgloveLog>).timestamp;
       if (typeof timestamp === "bigint") {
         return fromNanoSec(timestamp);
       }
-      return timestamp;
+      return timestamp ?? { sec: 0, nsec: 0 };
```
defunctzombie, line 56: I guess the log panel does not have a provision for the timetamp being undefined? would `NaN` work or will that crash other things? Guess 0/0 is a reasonable sentinel though it is also a real value.

defunctzombie (approved):

packages/studio-base/src/panels/Log/conversion.tsx
jtbandes: Yeah we already used 0/0 as a fallback in other places here so I didn't change that.

packages/studio-base/src/panels/Log/filterMessages.ts
jtbandes: Is there a way to do such a thing in TS?

PR #7077, opened 2023-11-07
Fix deserialization errors hiding errors from other topics

jtbandes:
**User-Facing Changes**
None

**Description**

The IterablePlayer uses the `connectionId` as part of the key in the problem manager. The source was not sending any `connectionId`, so errors on multiple channels would override each other and only one was visible at a time. With this change, each channel's latest error is visible.

https://github.com/foxglove/studio/blob/3f87e3009acfb94020ad8270492225179ca6a5d3/packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts#L652

Fixes FG-5631

### DIFF
--- a/packages/studio-base/src/players/IterablePlayer/BagIterableSource.ts
+++ b/packages/studio-base/src/players/IterablePlayer/BagIterableSource.ts
@@ -220,7 +220,6 @@ export class BagIterableSource implements IIterableSource {
 
         yield {
           type: "message-event",
-          connectionId,
           msgEvent: {
             topic: bagMsgEvent.topic,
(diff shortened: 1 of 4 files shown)


snosenzo (approved):
should we make `connectionId` on the `IteratorResult` type required? could other sources fulfill it in a similar way?

defunctzombie (approved):

jtbandes:
Yep it seems like we were now setting this everywhere, so I made it non-optional. Also, it appears we were not using the connectionId in the case of `message-event`s, so I removed it – @defunctzombie please take another look

snosenzo (approved):

PR #7079, opened 2023-11-07
use queueHandler instead of processQueue

snosenzo:
**User-Facing Changes**

**Description**

alternative approach to #7064


packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -1106,11 +1106,14 @@ export class Renderer extends EventEmitter<RendererEvents> implements IRenderer
...
+        const { queue, handler, queueHandler } = subscription;
         subscription.queue = undefined;
-        for (const messageEvent of processedQueue) {
-          subscription.handler(messageEvent);
+        if (queueHandler) {
+          queueHandler(queue);
+          continue;
+        }
+        for (const messageEvent of queue) {
+          handler(messageEvent);
```
jtbandes, line 1116: does this mean `handler` could now be optional (if `queueHandler` is provided)?

PR #7081, opened 2023-11-08
Add some minor console styles to theme

2metres:
**User-Facing Changes**
None

**Description**
Move some more minor styles from console into main theme


packages/theme/src/components/MuiFormControl.ts
```diff
@@ -2,10 +2,21 @@
...
   defaultProps: {
     variant: "standard",
   },
+  styleOverrides: {
+    root: {
+      [`label[data-shrink=false] + .${inputBaseClasses.formControl}`]: {
+        [`.${inputBaseClasses.input}::placeholder`]: {
+          opacity: "0.6 !important",
+        },
+      },
```
2metres, line 19: This selector is messed up but this is the reverse of what MUI material does for the form label

jtbandes (commented):
Can we add stories for whatever these changes are doing?

2metres:
They are mainly to address issues in the storybooks on console

jtbandes (approved):

PR #7084, opened 2023-11-08
Don't auto-merge electron updates

jtbandes:
**User-Facing Changes**
None

**Description**
The idea of auto-merging electron updates makes me a bit nervous.


defunctzombie (approved):
Seems fine - though worth a comment in the action cause when I read this again in a year I am unlikely to remember why this matters

If this makes you nervous do we need better integration tests? What will make you less nervous?

jtbandes:
Upon further thought this PR is misguided because we removed the electron dependency from this repo. 🙃

PR #7085, opened 2023-11-08
Address TopicList feedback

2metres:
**User-Facing Changes**
Hide the topic stats as a last resort and tighten spacing around drag handle.

FG-5411

### DIFF
--- a/packages/studio-base/src/components/TopicList/TopicStatsChip.tsx
+++ b/packages/studio-base/src/components/TopicList/TopicStatsChip.tsx
@@ -13,8 +13,9 @@ const useStyles = makeStyles<void, "selected">()((theme, _props, classes) => ({
     color: theme.palette.action.selected,
     borderColor: "currentColor",
     backgroundColor: theme.palette.background.paper,
+    cursor: "grab",
 
-    [`@container (max-width: 320px)`]: {
+    [`@container (max-width: 180px)`]: {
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/components/TopicList/TopicStatsChip.tsx
```diff
@@ -13,8 +13,9 @@ const useStyles = makeStyles<void, "selected">()((theme, _props, classes) => ({
     color: theme.palette.action.selected,
     borderColor: "currentColor",
     backgroundColor: theme.palette.background.paper,
+    cursor: "grab",
 
-    [`@container (max-width: 320px)`]: {
+    [`@container (max-width: 180px)`]: {
```
jtbandes, line 18: I am not the one with strong feelings on this design. But is there a world where we move back to the vertically stacked stats and remove `display: none` because it works even at narrow sizes (with default text truncation)?

jtbandes (approved):

packages/studio-base/src/components/TopicList/TopicStatsChip.tsx
defunctzombie: There seems to be some aversion to the vertically stacked stat /shrug. Personally I find the pill out-of-place but at least with this change my primary report is resolved.

PR #7086, opened 2023-11-08
Always show state transition tooltip

defunctzombie:
**User-Facing Changes**

**Description**

### DIFF
--- a/packages/studio-base/src/components/Chart/worker/ChartJSManager.ts
+++ b/packages/studio-base/src/components/Chart/worker/ChartJSManager.ts
@@ -11,7 +11,16 @@
 //   found at http://www.apache.org/licenses/LICENSE-2.0
 //   You may not use this file except in compliance with the License.
 
-import { Chart, ChartData, ChartOptions, ChartType } from "chart.js";
+import {
+  Chart,
+  ChartData,
+  ChartOptions,
+  ChartType,
+  Interaction,
+  InteractionModeFunction,
+  InteractionItem,
+} from "chart.js";
+import { getRelativePosition } from "chart.js/helpers";
 import type { Context as DatalabelContext } from "chartjs-plugin-datalabels";
 import DatalabelPlugin from "chartjs-plugin-datalabels";
 import { type Options as DatalabelsPluginOptions } from "chartjs-plugin-datalabels/types/options";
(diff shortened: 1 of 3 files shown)


snosenzo (approved):
don't have much context on chartjs, but I tested and it seems to work well with even large datasets

defunctzombie:
> Do we know if this has any performance implications for larger datasets?

From what I can tell by reading the chartjs code of the existing interaction mode we were using this is no worse. The existing code would also iterate through all of the data points in the dataset. So broadly yea there's a not-so-great performance thing happening here cause there is no index to make this lookup faster but its not a change compared to what we had before.

packages/studio-base/src/components/Chart/worker/ChartJSManager.ts
```diff
@@ -69,6 +78,39 @@ type ZoomableChart = Chart & {
...
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-argument, @typescript-eslint/no-explicit-any
+  const position = getRelativePosition(event, chart as any);
+
+  const datasetIndexToLastItem: InteractionItem[] = [];
+  Interaction.evaluateInteractionItems(chart, "x", position, (element, datasetIndex, index) => {
+    const center = element.getCenterPoint(useFinalPosition);
+    if (center.x <= position.x) {
+      datasetIndexToLastItem[datasetIndex] = { element, datasetIndex, index };
+    }
+  });
```
jtbandes, line 108: if you say so 😵‍💫

packages/studio-base/src/components/Chart/worker/ChartJSManager.ts
```diff
@@ -69,6 +78,39 @@ type ZoomableChart = Chart & {
...
+const lastX: InteractionModeFunction = (chart, event, _options, useFinalPosition) => {
+  // Suppress the type error on the _chart_ type. Chartjs types are broken for the
+  // `getRelativePosition` function which seems to use a different declaration of the Chart type
+  // than what is exported from chart.js.
+  //
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-argument, @typescript-eslint/no-explicit-any
+  const position = getRelativePosition(event, chart as any);
+
+  const datasetIndexToLastItem: InteractionItem[] = [];
+  Interaction.evaluateInteractionItems(chart, "x", position, (element, datasetIndex, index) => {
```
jtbandes, line 103: so is this a linear scan through all data points?

jtbandes (approved):

packages/studio-base/src/components/Chart/worker/ChartJSManager.ts
defunctzombie: yep - and my understanding of the existing interaction modes we use is that they behave similarly: https://github.com/chartjs/Chart.js/blob/master/src/core/core.interaction.js#L332

PR #7089, opened 2023-11-08
Change selected panel visual

defunctzombie:
**User-Facing Changes**
Highlight the toolbar for the selected panel.

Before:
Screenshot 2023-11-08 at 8 19 58 AM
Screenshot 2023-11-08 at 8 19 51 AM

After:
Screenshot 2023-11-08 at 8 18 50 AM
Screenshot 2023-11-08 at 8 18 42 AM

**Description**
This change alters the visual feedback for a selected panel. Prior to this change the visual feedback for a selected panel was a purple 1px border. This border went around the entire panel. This change updates the visual to highlight the panel toolbar instead of a border.

My opinion is that the toolbar highlight is less disruptive and maintains a cleaner appearence. The panel border would often result in a 1px boundary around some parts of the panel (possibly due to mosaic interactions) which has long been a visual annoyance for me.

### DIFF
--- a/packages/studio-base/src/components/Panel.tsx
+++ b/packages/studio-base/src/components/Panel.tsx
@@ -615,6 +615,7 @@ export default function Panel<
             id: childId,
             title,
             config: panelComponentConfig,
+            isSelected: isSelected || (isDragging && isValidTarget && isOver),
             saveConfig: saveConfig as SaveConfig<PanelConfig>,
             updatePanelConfigs,
             openSiblingPanel,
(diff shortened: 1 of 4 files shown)


2metres:
Can we do border and top bar?

jtbandes:
I agree this seems pretty subtle and I liked having a visual indicator of the whole panel's bounding box. With many panels in a layout, if you are editing a panel's settings I think it wouldn't be terribly clear which panel is selected because there's not much visual separation between adjacent panels.

amacneil:
@2metres and @defunctzombie to discuss

defunctzombie:
Summary from conversation: Without a border the title highlight does not adequately capture the highlighted panel so we don't like it as much as the border outline.

PR #7090, opened 2023-11-08
Fix panels not receiving latest messages on newly subscribed topic that is already subscribed to by another panel.

snosenzo:
**User-Facing Changes**

 - Fix panels not receiving latest messages on newly subscribed topic that is already subscribed to by another panel.

**Description**

The cause of this issue is that we are doing `mergeSubscriptions` in both the MessagePipeline and the `UserScriptPlayer` before it gets to the `IterablePlayer`  comparing against new and old subscriptions to determine whether to backfill. The fix was to make `IterablePlayer` compare subscription arrays before `mergeSubscriptions` in order to determine backfill. This will allow it to take into consideration duplicate subscriptions that could come from multiple panels.`mergeSubscriptions` is now only done to populate preload and allTopics. 

Fixes: FG-5574


jtbandes:
> Fix panels not receiving backfill when subscribing to a topic that another panel is subscribed to

I don't think this "user-facing change" will make sense to users – for the release notes let's describe the specific (or not too specific) symptoms a user of the app might actually have experienced with this bug.

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
```diff
@@ -300,27 +301,31 @@ export class IterablePlayer implements Player {
...
-    this.#metricsCollector.setSubscriptions(newSubscriptions);
+    // get stable array of all subscriptions (including duplicates)
+    const sortedSubs = newSubscriptions.sort((a, b) => a.topic.localeCompare(b.topic));
+    // If there are no changes to topics there's no reason to perform a "seek" to trigger loading
+    if (_.isEqual(sortedSubs, this.#subscriptions)) {
+      return;
+    }
+    this.#subscriptions = sortedSubs;
+    this.#metricsCollector.setSubscriptions(sortedSubs);
+    const mergedSubscriptions = mergeSubscriptions(sortedSubs) as SubscribePayload[];
```
jtbandes, line 312: can we figure out a way to make this work without `as`? like making mergeSubscriptions generic or having two overloads? Or can we change the downstream uses of this to accept Immutable subscriptions assuming they don't actually need them to be mutable (since probably none of this code does)?

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
```diff
@@ -300,27 +301,31 @@ export class IterablePlayer implements Player {
 
   public setSubscriptions(newSubscriptions: SubscribePayload[]): void {
     log.debug("set subscriptions", newSubscriptions);
-    this.#subscriptions = newSubscriptions;
-    this.#metricsCollector.setSubscriptions(newSubscriptions);
+    // get stable array of all subscriptions (including duplicates)
+    const sortedSubs = newSubscriptions.sort((a, b) => a.topic.localeCompare(b.topic));
+    // If there are no changes to topics there's no reason to perform a "seek" to trigger loading
```
jtbandes, line 306:
```suggestion
    // If there are no changes to subscriptions there's no reason to perform a "seek" to trigger loading
```

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
```diff
@@ -300,27 +301,31 @@ export class IterablePlayer implements Player {
...
+    }
+    this.#subscriptions = sortedSubs;
+    this.#metricsCollector.setSubscriptions(sortedSubs);
+    const mergedSubscriptions = mergeSubscriptions(sortedSubs) as SubscribePayload[];
 
-    const allTopics: TopicSelection = new Map(
-      this.#subscriptions.map((subscription) => [subscription.topic, subscription]),
+    this.#allTopics = new Map(
+      mergedSubscriptions.map((subscription) => [subscription.topic, subscription]),
     );
```
jtbandes, line 316: It looks like `mergeSubscriptions` can return multiple items for the same topic (because it handles full and partial separately). So how does this work if we put it into a single Map by topic?

jtbandes (commented):
Would it be feasible to create a pseudo "integration" test that creates a MessagePipeline with an IterablePlayer, adds two panels in order and asserts that the second panel gets the latest message?

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
defunctzombie: Agreed. We should not have an _as_ here. We control all the functions and types for this entire system so an _as_ here means we've messed up some assumption.

defunctzombie:
I'm not sure I fully understand or agree with the reasoning given with the merged subscriptions. Is the issue that upstream we've lost information about whether a topic needs or does not needs preloading? It makes me nervous to relax the idea that the subscriptions have not been merged prior to arriving at the player because that's a large purpose of the message pipeline - to take a single player listener callback (players only allow one) and multiplex this to panels. Reading deeper into the issue maybe what I am not clear on is what "merge" means for subscriptions and you are saying that we need to avoid merging subscriptions for preloading and those without preloading for the same topic?

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
snosenzo: It does keep preload and partial loading separate. This keeps the same behavior that `#allTopics` had previously. `#allTopics` is only used grab messages on topics from the underlying source for the current frame or backfill. Thinking about it more, I think this might not be ideal because it could use a preloading subscriptions fields to read in for the current frame if it overrides the partial subscription passed to the Map. I wonder why we haven't had issues with that previously. However, this fix might be a bit out of scope for this bug. What I'm doing in this PR is keeping the same behavior for `#allTopics` that we had previously.

snosenzo:
@defunctzombie 
> I'm not sure I fully understand or agree with the reasoning given with the merged subscriptions. Is the issue that upstream we've lost information about whether a topic needs or does not needs preloading? It makes me nervous to relax the idea that the subscriptions have not been merged prior to arriving at the player because that's a large purpose of the message pipeline - to take a single player listener callback (players only allow one) and multiplex this to panels. Reading deeper into the issue maybe what I am not clear on is what "merge" means for subscriptions and you are saying that we need to avoid merging subscriptions for preloading and those without preloading for the same topic?

The crux of this issue was that we were merging duplicate partial and preloaded subscriptions before they got to the `IterablePlayer.setSubscriptions`. This means that the `IterablePlayer` was not able to know that duplicate subscriptions were added from new panels in order to perform backfill. The `IterablePlayer` should know that duplicate subscriptions were added because it should perform backfill to emit messages to subscribers of that duplicate subscription. Indeed, `mergeSubscriptions` does not merge subscriptions across preload/partial boundaries, but keeping those separate is not really relevant to this issue. The fact is that `mergeSubscriptions` was hiding that subscriptions had changed from the player which was preventing backfill.

jtbandes:
> This means that the `IterablePlayer` was not able to know that duplicate subscriptions were added from new panels in order to perform backfill. The `IterablePlayer` should know that duplicate subscriptions were added because it should perform backfill to emit messages to subscribers of that duplicate subscription. 

Isn't this what the MessagePipeline is trying to do here?

https://github.com/foxglove/studio/blob/0b7e2ad056a85be9dbe08c016a260ef4d623dbf1/packages/studio-base/src/components/MessagePipeline/store.ts#L297-L299

snosenzo:
> > This means that the `IterablePlayer` was not able to know that duplicate subscriptions were added from new panels in order to perform backfill. The `IterablePlayer` should know that duplicate subscriptions were added because it should perform backfill to emit messages to subscribers of that duplicate subscription.
> 
> Isn't this what the MessagePipeline is trying to do here?
> 
> https://github.com/foxglove/studio/blob/0b7e2ad056a85be9dbe08c016a260ef4d623dbf1/packages/studio-base/src/components/MessagePipeline/store.ts#L297-L299

I think you are correct that this should be working, however because "subscriptions" haven't changed according to the underlying player it does not re-emit state. Which means that `updatePlayerStateAction` on the message pipeline store is never called to set the lastMessages on the newly subscribed topics. I think the solution here is to either have the player re-emit an unchanged state after `setSubscriptions` is called, or to have the `updateSubscriberAction` on the messagepipeline store immediately apply the `lastMessageEventsByTopic` to new subscribers in the new state. I'm leaning towards the latter, however I would probably abandon this PR in favor of a new one for that route.

snosenzo:
Closing this PR in favor of this approach: https://github.com/foxglove/studio/pull/7113

PR #7091, opened 2023-11-08
Fix relative mesh URLs of URDF loaded from file not working

achim-k:
**User-Facing Changes**
Fix relative mesh URLs of URDF loaded from file not working

**Description**
Fixes that meshes with relative paths were not working for URDFs that were loaded from file as the [baseURL](https://github.com/foxglove/studio/blob/e3c981bf595731b7e20c41d6873daae4aff15857/packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts#L941) was not defined.

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
@@ -127,8 +127,6 @@ const tempEuler = new THREE.Euler();
 export type UrdfUserData = BaseUserData & {
   settings: LayerSettingsUrdf | LayerSettingsCustomUrdf;
   fetching?: { url: string; control: AbortController };
-  url: string | undefined;
-  filePath: string | undefined;
   urdf: string | undefined;
   sourceType: LayerSettingsCustomUrdf["sourceType"] | undefined;
   parameter: string | undefined;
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
```diff
@@ -861,6 +856,14 @@ export class Urdfs extends SceneExtension<UrdfRenderable> {
     renderable.userData.settings = settings;
     renderable.userData.fetching = undefined;
 
+    if (urdf != undefined && sourceType === "url") {
```
snosenzo, line 859: what is the significance of urdf being undefined here and why don't we want to set this if it is undefined? Can you add some comments about this intention of this logic?

snosenzo (approved):
would be good to get some comments around this logic and a story would be good to have too if possible

packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts
```diff
@@ -742,15 +740,6 @@ export class Urdfs extends SceneExtension<UrdfRenderable> {
     }
     this.renderer.settings.errors.remove(renderable.userData.settingsPath, VALID_SRC_ERR);
 
-    // Check if this URL has already been fetched
```
snosenzo, line 745: where are we checking that the url hasn't been fetched now?

achim-k: We call `#fetchUrdf` only when the urdf is not defined. Hence this check was unnecessary

https://github.com/foxglove/studio/blob/037a042d295e6d95dcc865a13ac253ccff0fc36a/packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts#L852-L865

PR #7092, opened 2023-11-08
Add PerformanceContext and record time spent in UserScriptPlayer onPlayerState

jtbandes:
**User-Facing Changes**

None

**Description**

Adds PerformanceContext and logs data from UserScriptPlayer to the context.


packages/studio-base/src/Workspace.tsx
```diff
@@ -356,13 +356,13 @@ function WorkspaceContent(props: WorkspaceProps): JSX.Element {
       ["variables", { title: t("variables"), component: VariablesList }],
     ]);
     if (enableDebugMode) {
-      items.set("studio-logs-settings", { title: t("studioLogs"), component: StudioLogsSettings });
```
defunctzombie, line 359: I see where your priorities are 🤣

packages/studio-base/src/context/PerformanceContext.ts
```diff
@@ -0,0 +1,56 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { createContext, useContext } from "react";
+
+// Ensure Symbol.dispose and Symbol.asyncDispose are defined
+// https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-2.html
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+(Symbol as any).dispose ??= Symbol("Symbol.dispose");
```
defunctzombie, line 10: Can we do this in https://github.com/foxglove/studio/blob/main/packages/studio-base/src/typings/overrides.d.ts or one of the other files there?

packages/studio-base/src/context/PerformanceContext.ts
```diff
@@ -0,0 +1,56 @@
...
+  /** Name, such as "Execution time" */
+  name: string;
+  /** Human-readable unit, such as "ms per frame" */
+  unit: string;
+};
+
+/** An object that records performance timing information. */
+export interface IPerformanceRegistry {
+  /** Register a new metric with the given name and unit */
+  registerMetric(metric: Omit<PerformanceMetric, "id">): PerformanceMetricID;
```
defunctzombie, line 27: Do I need to register a metric before adding measurements? Might be handy to call out this behavior or appropriate call order either here or the `addMeasurement` method.

defunctzombie (approved):

packages/studio-base/src/context/PerformanceContext.ts
```diff
@@ -0,0 +1,56 @@
...
+  /** Unregister a previously registered metric */
+  unregisterMetric(metricId: PerformanceMetricID): void;
+
+  /** Record data for a metric */
+  addMeasurement(metricId: PerformanceMetricID, timestamp: number, value: number): void;
```
defunctzombie, line 32: Should timestamp be our `Time` type? To be clear what kind of units it is? What is `value`?

packages/studio-base/src/players/UserScriptPlayer/index.ts
```diff
@@ -842,6 +853,17 @@ export default class UserScriptPlayer implements Player {
...
+      if (this.#totalTimeMetric == undefined) {
+        this.#totalTimeMetric = this.#perfRegistry?.registerMetric({
+          name: "User scripts (total)",
+          unit: "ms per frame",
+        });
+      }
+      using $timer = this.#totalTimeMetric
+        ? this.#perfRegistry?.scopeTimer(this.#totalTimeMetric)
+        : undefined;
+      void $timer;
```
defunctzombie, line 865: Any hope of removing this _void_ requirement when setting up the scope timer?

packages/studio-base/src/context/PerformanceContext.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  /**
+   * Starts measuring time and returns a Disposable that finishes & records the measurement when it
+   * is disposed. Use with `using timer = scopeTimer(...);`.
```
snosenzo, line 36: I feel we should note that the default behavior of this would be to dispose of itself at the end of the scope of the closure that it was used in. I know that's implicit in `using` but it is a bit of a new paradigm and would be nice to have a reminder.

packages/studio-base/src/context/PerformanceContext.ts
```diff
@@ -0,0 +1,56 @@
...
+   */
+  scopeTimer(metricId: PerformanceMetricID): Disposable;
```
snosenzo, line 38: Not sure how I feel about having implicit disposing of these timers. I think you lose readability of where the timer precisely ends, and it's easy to add things to what it's measuring without the intention to do so. I'd prefer having an explicit end so that it's less likely that these metrics are modified unintentionally ie: are measuring more or less than what's intended.

defunctzombie: The definition of a scope timer is that it ends when the scope ends - I do think that it could provide a function to end the scope earlier for cases where you want to do that without terminating scope. In practice its a super handy concept.

snosenzo: Yeah I agree it's a useful concept, and it accounts for errors being thrown in the middle of what it's measuring.I think I was a little thrown off by it being used in the middle of a closure of a rather large function below. So maybe this is more of a usage problem for me. As for ending the scope early, technically they could just call the dispose function early I believe? 
Anyway, I'm fine with this now. :shipit:

packages/studio-base/src/players/UserScriptPlayer/index.ts
```diff
@@ -842,6 +853,17 @@ export default class UserScriptPlayer implements Player {
         return;
       }
 
+      if (this.#totalTimeMetric == undefined) {
```
snosenzo, line 856: why not move this to the beginning of the `try` block or function?

packages/studio-base/src/context/PerformanceContext.ts
```diff
@@ -0,0 +1,56 @@
...
+import { createContext, useContext } from "react";
+
+// Ensure Symbol.dispose and Symbol.asyncDispose are defined
+// https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-2.html
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+(Symbol as any).dispose ??= Symbol("Symbol.dispose");
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+(Symbol as any).asyncDispose ??= Symbol("Symbol.asyncDispose");
+
+export type PerformanceMetricID = number & { __brand: "PerformanceMetricID" };
```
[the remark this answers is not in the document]
jtbandes: Filed FG-5696 for this. It does not seem we currently use `Opaque` anywhere so what I have here is more consistent for now.

packages/studio-base/src/context/PerformanceContext.ts
jtbandes: This is a runtime polyfill, not a type definition – the type definitions are included by adding the `disposable` thing to tsconfig `lib`.

packages/studio-base/src/context/PerformanceContext.ts
jtbandes: IMO the fact that you have to pass a `PerformanceMetricID` into `addMeasurement` is sufficient documentation :)

packages/studio-base/src/players/UserScriptPlayer/index.ts
jtbandes: The alternative is:

```ts
      // eslint-disable-next-line no-underscore-dangle
      using _timer = ...
```

but I'm not sure that's better?

packages/studio-base/src/players/UserScriptPlayer/index.ts
jtbandes: Which part – the `if` or the scope timer? The scope timer has to be here so that it does not wrap around the `await this.#queueEmitState();` in the function's `finally`, and I also felt like wrapping around the `if (!activeData)` case wouldn't be that interesting because the user script player isn't really doing anything in that case.

For the `if` I just felt like keeping it close to where the `#totalTimeMetric` was actually used 🤷

packages/studio-base/src/context/PerformanceContext.ts
jtbandes: Added comments

PR #7100, opened 2023-11-09
Tolerate missing properties key in JSON schema

bryfox:
**User-Facing Changes**

Not significant.

**Description**

This allows a simpler representation of a JSON schema representing "any object".

When parsing a JSON schema, we assume the presence of `.properties`, but this is not a required field. Currently, if I open an mcap file with a JSON schema of `{"type":"object"}`, I see a confusing error. We could produce a better error message. However, I'd like to document this workaround for empty schemas, and accepting the above schema seems preferable to requiring `{"type": "object", "properties": {}}`.

err

### DIFF
--- a/packages/mcap-support/src/parseJsonSchema.test.ts
+++ b/packages/mcap-support/src/parseJsonSchema.test.ts
@@ -285,4 +285,18 @@ describe("parseJsonSchema", () => {
   it.each(Object.values(foxgloveMessageSchemas))("handles Foxglove schema '$name'", (schema) => {
     expect(() => parseJsonSchema(generateJsonSchema(schema), schema.name)).not.toThrow();
   });
+
+  it("tolerates an 'any object' schema", () => {
+    const { datatypes } = parseJsonSchema({ type: "object" }, "Any object");
+    expect(datatypes).toEqual(
+      new Map([
+        [
+          "Any object",
+          {
+            definitions: [],
+          },
+        ],
+      ]),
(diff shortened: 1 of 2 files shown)


bryfox:
> This change seems fine, but as for documenting the workaround, why would a `jsonschema` with `{type:"object"}` be preferred over no schema at all?

Data platform does yet not support schemaless MCAPs because of relational constraints.

Supporting the [empty schema](https://mcap.dev/spec/registry#empty-string) here would be preferable to the "any-object" schema. It looks like we might be able to support that by treating `undefined` and empty schemas the same through the following block, but I was going for a low-touch change.

https://github.com/foxglove/studio/blob/480e8222affc9fcef029ffcb46dccffccdae4e0e/packages/mcap-support/src/parseChannel.ts#L80-L108

If it seems reasonable to support empty-string schemas above, I can update this PR.

jtbandes:
Yeah, I think technically according to the spec we should support them and treat them the same as missing schemas. It's fine to support that only for `json` though (since the app doesn't generally support mix & match between any message encoding & schema encoding).

image

Also this PR is totally viable as-is, so if you want to merge it and do the empty string thing separately that's fine too!

bryfox:
I wrote some code to deal with the empty schema (encoding="", data.length=0), but while it doesn't report any Problems, I don't see raw messages appearing from my test file like I'd expect. I'll file a follow-up issue for the empty string support.

PR #7102, opened 2023-11-09
Remove ROS assumptions when using services with foxglove websocket

### LINKED ISSUE #6970: Service calls make ROS assumptions with _Request and _Response type suffixes
```
requestType = `${service.type}_Request`;
responseType = `${service.type}_Response`;
```

This is a ROS-specific assumption in FoxgloveWebSocketPlayer/index.ts. Services that don't follow ROS conventions can't easily be mapped to the Foxglove WebSocket protocol without modifying Studio.

achim-k:
**User-Facing Changes**
Remove ROS assumptions when using services with foxglove websocket

**Description**
Updates the websocket player to comply with the latest `ws-protocol` spec changes made in https://github.com/foxglove/ws-protocol/pull/563. This allows to remove some ROS specific assumptions which makes it easier to advertise services from non-ROS systems.

Fixes #6970
Resolves FG-5280

### DIFF
--- a/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
+++ b/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
@@ -648,6 +648,8 @@ export default class FoxgloveWebSocketPlayer implements Player {
 
       for (const service of services) {
         const serviceProblemId = `service:${service.id}`;
+        // If not explicitly given, derive request / response type name from the service type
+        // (according to ROS convention).
         const requestType = service.request?.schemaName ?? `${service.type}_Request`;
         const responseType = service.response?.schemaName ?? `${service.type}_Response`;
         const requestMsgEncoding = service.request?.encoding ?? this.#serviceCallEncoding;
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
```diff
@@ -637,44 +637,63 @@ export default class FoxgloveWebSocketPlayer implements Player {
...
-        schemaEncoding = "ros2msg";
-      } else {
-        throw new Error(`Unsupported encoding "${this.#serviceCallEncoding}"`);
+        defaultSchemaEncoding = "ros2msg";
       }
 
       for (const service of services) {
         const serviceProblemId = `service:${service.id}`;
+        const requestType = service.request?.schemaName ?? `${service.type}_Request`;
+        const responseType = service.response?.schemaName ?? `${service.type}_Response`;
```
jtbandes, line 652: maybe add a comment that these fallbacks are from ROS? Should we also gate this on the encoding being one of the ROS encodings?

jtbandes (approved):

packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
```diff
@@ -637,44 +637,63 @@ export default class FoxgloveWebSocketPlayer implements Player {
...
             },
           });
           const requestMsgDef = rosDatatypesToMessageDefinition(
             parsedRequest.datatypes,
             requestType,
           );
-          const requestMessageWriter = ROS_ENCODINGS.includes(this.#serviceCallEncoding)
-            ? this.#serviceCallEncoding === "ros1"
+          const requestMessageWriter = ROS_ENCODINGS.includes(requestMsgEncoding)
+            ? requestMsgEncoding === "ros1"
```
jtbandes, line 696: nit: seems unnecessary to have ROS_ENCODINGS and also an explicit check for `ros1`... we could instead have an explicit check for `ros1` and one for `cdr`

packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
achim-k: > Should we also gate this on the encoding being one of the ROS encodings?

I wouldn't do this. Someone might have adapted their server implementation to be compatible with this assumption here. I'll add the comment though.

PR #7105, opened 2023-11-10
Fix legend styles so remove button is sticky (…again)

2metres:
**User-Facing Changes**
None

**Description**

https://github.com/foxglove/studio/assets/924528/ada41813-e2ab-4bb1-97c2-fa2543d342c5

Bug fix where plot legend remove button was not sticking

Resolves FG-5484


jtbandes (approved):
Seems to work and does not break plot mouse interaction 😄

PR #7106, opened 2023-11-10
Move reset zoom button so it only overlaps StateTransition chart when it has to.

2metres:
**User-Facing Changes**
None

**Description**
Move reset zoom button so it only overlaps StateTransition chart when it has to.

### DIFF
--- a/packages/studio-base/src/components/TimeBasedChart/index.tsx
+++ b/packages/studio-base/src/components/TimeBasedChart/index.tsx
@@ -76,6 +76,9 @@ const useStyles = makeStyles()((theme) => ({
       pointerEvents: "auto",
     },
   },
+  stateTransition: {
+    padding: theme.spacing(0, 1, 2),
+  },
   tooltip: {
     maxWidth: "none",
   },
@@ -116,6 +119,7 @@ export type Props = {
   xAxes?: ScaleOptions<"linear">;
   yAxes: ScaleOptions<"linear">;
   annotations?: AnnotationOptions[];
+  isStateTransition?: boolean;
   isSynced?: boolean;
   linesToHide?: {
     [key: string]: boolean;
@@ -148,6 +152,7 @@ export default function TimeBasedChart(props: Props): JSX.Element {
     datasetId,
     defaultView,
     height,
+    isStateTransition = false,
     isSynced = false,
     showXAxisLabels,
     type,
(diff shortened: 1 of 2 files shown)


defunctzombie:
Is this ready for review or is it a draft?

2metres:
Sorry, It's a draft. I found a bug after reviewing the stories

2metres:
Ready to go now

jtbandes:
This change does seem overall a lot better. However when I add many series to a state transitions panel, and scroll down, I get this which looks weird (probably because of inconsistent bottom/right whitespace):

image

Maybe just make the bottom and right margins the same?

2metres:
This same button is present on the plot panel as well, hence the margin bottom to avoid the X axis

defunctzombie:
> This same button is present on the plot panel as well, hence the margin bottom to avoid the X axis

As in it is provided by the `TimeBasedChart` component which is shared between the two?

Question unrelated to PR: did this panel always scroll? Not super obvious if your series is not in the visible panel area?

defunctzombie:
https://github.com/foxglove/studio/assets/84792/79f0cf21-c214-43c7-99ba-d98ee95017b7

Something about these interactions still feels off. The scrollbar appearing when there doesn't seem to be a need for one is strange.

2metres:
Yeah I agree the position of the reset button and drag targets are not great. Overall this PR feels like a bit of a hack. Not sure how we might fix this more holistically

2metres:
> Question unrelated to PR: did this panel always scroll? Not super obvious if your series is not in the visible panel area?

Yeah it has always scrolled however the scroll target is only 6px on the right side of the panel. Not vary obvious

amacneil:
:+1:

2metres:
Feedback: pass prop through to adjust button margin

packages/studio-base/src/components/TimeBasedChart/index.tsx
```diff
@@ -116,6 +119,7 @@ export type Props = {
   xAxes?: ScaleOptions<"linear">;
   yAxes: ScaleOptions<"linear">;
   annotations?: AnnotationOptions[];
+  isStateTransition?: boolean;
```
defunctzombie, line 122: a better name for this is something like "resetButtonPadding" or similar that does not imply some knowledge of the parent but instead exposes a property that lets you control behavior of this component. To the TimeBasedChart component the idea of "isStateTransition" does not mean anything.

jtbandes (approved):

PR #7112, opened 2023-11-13
Fix plot story flakiness

cfoust:
**User-Facing Changes**
None; this is just an improvement to developer experience.

**Description**
Our plot stories have been flaky for a while, even predating the changes I made to move data processing into a worker. This PR attempts to fix these issues.

There were two main causes for flaky stories:
1. The mechanism we were using to indicate that a story was ready was error-prone, particularly after moving plot data processing to a worker, which meant that differences in timing could cause us to render an inconsistent number of times. This broke the traditional `useReadySignal({ count: N })` strategy.
2. There were (and are) significant problems with timing in the `Chart` component, which sends rendering jobs to a Web Worker that uses ChartJS.

I fixed #1 by switching to a strategy where, instead of counting the number of rerenders, we wait a certain amount of time after the last time the Chart component renders before declaring the story "ready". This may not actually fix this problem definitively because Chromatic's workers are extremely weak, but it's less error-prone than hoping we get the same number of renders every time.

For #2 I added what can only be described as a hack because that component is woefully, woefully tangled. (I'm getting flashbacks to our layout bug.) We have several interacting layers of React state (mostly refs), `async` code that can complete at any time, and then, if that weren't enough, the whole `Rpc` layer that also seems to respond to requests whenever it feels like.

The core problem is that we were [waiting for a bunch of calls to finish](https://github.com/foxglove/studio/pull/7112/files#diff-c83b85fde0ce9eb98a2d3d36cc869873cad467fc1fb3aa3f4cc48a8fe73e6b8fR286) asynchronously and then setting some state on the component, which in a compute-constrained environment could mean that several seconds have already passed. In other words, the requests in flight elsewhere were being randomly overwritten.

In my opinion, we really just need to rewrite this whole section with `getNewUpdateMessage`, the queued updates stuff, and initialization. I swear, at least half of the problems we've had with rendering and timing have been because of this code block.

### DIFF
--- a/packages/studio-base/src/components/Chart/index.tsx
+++ b/packages/studio-base/src/components/Chart/index.tsx
@@ -284,8 +284,14 @@ function Chart(props: Props): JSX.Element {
         const { current: queued } = queuedUpdates;
         queuedUpdates.current = [];
         for (const queuedUpdate of queued) {
-          const newScales = await sendWrapperRef.current<RpcScales>("update", queuedUpdate);
-          maybeUpdateScales(newScales);
+          void (async () => {
+            const { current } = sendWrapperRef;
+            if (current == undefined) {
+              return;
+            }
+            const newScales = await current<RpcScales>("update", queuedUpdate);
+            maybeUpdateScales(newScales);
+          });
         }
 
         // once we are initialized, we can allow other handlers to send to the rpc endpoint
--- a/packages/studio-base/src/components/Chart/index.tsx
+++ b/packages/studio-base/src/components/Chart/index.tsx
@@ -286,14 +286,7 @@ function Chart(props: Props): JSX.Element {
         for (const queuedUpdate of queued) {
           // We do not want to block the rest of this function on these calls,
           // which can take a while (particularly on weak devices)
-          void (async () => {
-            const { current } = sendWrapperRef;
-            if (current == undefined) {
-              return;
-            }
-            const newScales = await current<RpcScales>("update", queuedUpdate);
-            maybeUpdateScales(newScales);
-          })();
+          void sendWrapperRef.current<RpcScales>("update", queuedUpdate).then(maybeUpdateScales);
         }
 
         // once we are initialized, we can allow other handlers to send to the rpc endpoint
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/components/Chart/index.tsx
```diff
@@ -284,8 +284,16 @@ function Chart(props: Props): JSX.Element {
         const { current: queued } = queuedUpdates;
         queuedUpdates.current = [];
         for (const queuedUpdate of queued) {
-          const newScales = await sendWrapperRef.current<RpcScales>("update", queuedUpdate);
-          maybeUpdateScales(newScales);
+          // We do not want to block the rest of this function on these calls,
+          // which can take a while (particularly on weak devices)
```
jtbandes, line 288: What happens if you don't make this change and only do the debouncing stuff?

packages/studio-base/src/components/Chart/index.tsx
```diff
@@ -284,8 +284,16 @@ function Chart(props: Props): JSX.Element {
...
+          void (async () => {
+            const { current } = sendWrapperRef;
+            if (current == undefined) {
+              return;
+            }
+            const newScales = await current<RpcScales>("update", queuedUpdate);
+            maybeUpdateScales(newScales);
+          })();
```
jtbandes, line 296: Little hard to follow this logic but it kinda looks like it can be replaced with `void sendWrapperRef.current<RpcScales>("update", queuedUpdate).then(maybeUpdateScales);`.

packages/studio-base/src/components/Chart/index.tsx
cfoust: we would still have flaky stories

packages/studio-base/src/components/Chart/index.tsx
cfoust: Personally I find `async` code easier to read than promise chaining--I haven't seen a `then` for a long time!

packages/studio-base/src/components/Chart/index.tsx
cfoust: this code is broken in its current form, which is particularly obvious in the Chromatic runtime environment

packages/studio-base/src/components/Chart/index.tsx
defunctzombie: I agree with Jacob on the style here - the many lines of code and iife closure made in each loop do seem easier to read with his proposed line of code.

I've found `.then` more common when `void`ing awaits since you still have something to handle the response.

defunctzombie (approved):
After discussion we think this is a safe short-term fix until we get an explicit queue in place we can cull for unnecessary updates.

Code style wise I like the proposed one-liner over the current iife.

packages/studio-base/src/components/Chart/index.tsx
jtbandes: You also get to take advantage of the narrowing where typescript knows that `sendWrapperRef.current` is not undefined.

PR #7113, opened 2023-11-13
Fix panels not receiving latest messages on newly subscribed topic that is already subscribed to by another panel

snosenzo:
**User-Facing Changes**

 - Fix panels not receiving latest messages on newly subscribed topic that is already subscribed to by another panel

**Description**
Supercedes: https://github.com/foxglove/studio/pull/7090

Because "subscriptions" haven't changed according to the underlying player it does not re-emit state. Which means that updatePlayerStateAction on the message pipeline store is never called to set the lastMessages on the newly subscribed topics. 

This PR moves the logic for applying lastMessages to subscribed topics to `updateSubscriberAction` to emit these messages immediately instead of waiting for another underlying player emit state. I've removed the need to store `newTopics` across the actions because handling these topics happens entirely in the `updateSubscriberAction`. I've also added some logic to remove messages on topics that are no longer subscribed to from `lastMessageByTopic` so that stale messages are not passed when resubscribing to that topic before backfill resolves. 

Some notes about reading the PR:
I've removed some useDelayedFixture since they were causing strange behavior with some stories. These shouldn't be necessary anymore since the `MockMessagePipelineProvider` should now handle "Player" lifecycle and doing backfill for initial subscriptions better. 

I had to add a `pauseFrame` functionality that more closely mirrored player listener behavior to the MMPP so that we wait until all components finish rendering before updating the store. 

Fixes: FG-5574


packages/studio-base/src/components/MessagePipeline/store.ts
```diff
@@ -62,7 +62,8 @@ export type MessagePipelineInternalState = {
    * for dispatching messages needs to iterate over the array of IDs.
    */
   subscriberIdsByTopic: Map<string, string[]>;
-  newTopicsBySubscriberId: Map<string, Set<string>>;
+  /** This holds the last message emitted by the player on each topic. This should attempt to be used before
```
defunctzombie, line 65: before...what?

packages/studio-base/src/PanelAPI/useMessageReducer.ts
```diff
@@ -30,18 +30,37 @@ import {
 
 const log = Log.getLogger(__filename);
 
-type MessageReducer<T> = (arg0: T, message: MessageEvent) => T;
-type MessagesReducer<T> = (arg0: T, messages: readonly MessageEvent[]) => T;
+type MessageReducer<T> = (state: T, message: MessageEvent) => T;
+type MessagesReducer<T> = (state: T, messages: readonly MessageEvent[]) => T;
 
 type Params<T> = {
+  /**
```
snosenzo, line 37: commented this because I had to debug the tests and wanted to have a better understanding of the parameters

packages/studio-base/src/components/MessagePipeline/index.test.tsx
```diff
@@ -358,11 +358,7 @@ describe("MessagePipelineProvider/useMessagePipeline", () => {
     });
     expect(result.current.subscriptions).toEqual([{ topic: "/input/foo" }]);
 
-    // Emit empty player state to process new subscriptions
```
snosenzo, line 361: no longer need to wait for a player to emit again to receive the last messages on a topic

packages/studio-base/src/PanelAPI/useMessageReducer.test.tsx
```diff
@@ -436,53 +436,18 @@ describe("useMessageReducer", () => {
 
     // restore call with undefined, then add message called with our subscribed message
     expect(restore.mock.calls).toEqual([[undefined], [undefined]]);
-    expect(addMessage.mock.calls).toEqual([[0, message2]]);
+    expect(addMessage.mock.calls).toEqual([[0, messageBar]]);
     expect(result.current).toEqual(2);
 
     rerender({ topics: ["/bar", "/foo"] });
 
-    // no additional calls
```
snosenzo, line 444: no longer need player emit to push lastMessagesOnTOpic

packages/studio-base/src/components/MessagePipeline/MockMessagePipelineProvider.tsx
```diff
@@ -216,49 +254,34 @@ export default function MockMessagePipelineProvider(
...
             return { ...newState, dispatch: state.dispatch };
           });
         }
       };
       const reset = () => {
         throw new Error("not implemented");
       };
-      const initialPublicState = getPublicState(undefined, props, dispatch);
+
+      // exclude messages from initial state because there are no subscribers yet
```
snosenzo, line 275: will remove

packages/studio-base/src/panels/ThreeDeeRender/stories/Urdfs.stories.tsx
```diff
@@ -164,39 +163,42 @@ export const Urdfs: StoryObj = {
...
+    const fixture = {
       topics,
       frame: {
-        "/robot_description": [robot_description],
-      },
-      capabilities: [],
-      activeData: {
-        currentTime: { sec: 0, nsec: 0 },
-        parameters: new Map([[urdfParamName, URDF3]]),
-        messages: [
```
jtbandes, line 176: was this just broken if these were getting ignored?

packages/studio-base/src/panels/ThreeDeeRender/stories/UrdfDisplayMode.stories.tsx
```diff
@@ -112,13 +111,11 @@ export const UrdfDisplayMode: StoryObj = {
       },
     };
 
-    const fixture = useDelayedFixture({
```
jtbandes, line 115: did having this break something?

jtbandes (approved):

packages/studio-base/src/panels/ThreeDeeRender/stories/UrdfDisplayMode.stories.tsx
snosenzo: This wasn't breaking anything as of my final implementation, but I believe it was responsible for the failing story in a previous iteration. Since it doesn't do anything anymore I'll keep it as is

packages/studio-base/src/components/MessagePipeline/store.ts
```diff
@@ -227,36 +228,20 @@ function updateSubscriberAction(
   action: UpdateSubscriberAction,
```
defunctzombie, line 228: Github is terrible at letting me comment outside of changed lines but the comment a few lines above this that mentions `newTopicsBySubscriberId` is wrong now. You can probably write a better comment on what this function does given the new implementation.

packages/studio-base/src/components/MessagePipeline/store.ts
```diff
@@ -271,17 +256,55 @@ function updateSubscriberAction(
...
+
+  const prevSubsForId = previousSubscriptionsById.get(action.id);
+  const prevTopics = new Set(prevSubsForId?.map((sub) => sub.topic) ?? []);
+  for (const { topic: newTopic } of action.payloads) {
+    if (!prevTopics.has(newTopic)) {
+      newTopicsForId.add(newTopic);
+    }
+  }
+
+  const messagesBySubscriberId = new Map<string, MessageEvent[]>();
```
defunctzombie, line 270: Is `SubscriberId` the same ID as `newTopicsForId` a few lines up? Both are the subscriber id? How come we have a new map of messages by subscriberId here but the action only has one `id`?

defunctzombie: If my reading of this is correct - the downsteam impact of this change is that whenever a panel subscribes, all panels will get notified with empty frame data? There is no carry-over of the `messagesBySubscriberId` from the existing state. Is that a problem? No test seemed to think it was a problem but it does seem like there's _some_ gap here.

packages/studio-base/src/panels/ThreeDeeRender/stories/Urdfs.stories.tsx
snosenzo: These weren't being ignored, but they weren't going through the normal message pipeline. They were emitted on the initialization of the store, but because there are no subscribers, nothing would handle them. They would not be emitted again because subsequent dispatching to the MP would overwrite it. putting them in the `frame` lets them go through the normal `messages` mock message pipeline prop

packages/studio-base/src/PanelAPI/useMessageReducer.ts
```diff
@@ -30,18 +30,37 @@ import {
...
-  // Functions called when the reducers change and for each newly received message.
-  // The object is assumed to be immutable, so in order to trigger a re-render, the reducers must
-  // return a new object.
-  restore: (arg: T | undefined) => T;
+  /**
+   * Called on intialization, seek, and when reducers change.
+   * @param state - Immutable. `undefined` when called on initialization or seek. Otherwise, the current state.
+   * @returns - New state. Must be new reference to trigger rerender.
+   */
+  restore: (state: T | undefined) => T;
```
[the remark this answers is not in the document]
snosenzo: Was just trying to portray the original comment, will try to apply `Immutable<T>` but might complain in a bunch of places that I don't want to change in this PR

packages/studio-base/src/components/MessagePipeline/store.ts
```diff
@@ -62,7 +62,8 @@ export type MessagePipelineInternalState = {
    * for dispatching messages needs to iterate over the array of IDs.
    */
   subscriberIdsByTopic: Map<string, string[]>;
-  newTopicsBySubscriberId: Map<string, Set<string>>;
+  /** This holds the last message emitted by the player on each topic. Attempt to use this before falling back to player backfill.
```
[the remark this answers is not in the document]
snosenzo: player backfill reads the last message on all active subscriptions from the file and triggers a `seek`

packages/studio-base/src/components/PanelExtensionAdapter/PanelExtensionAdapter.test.tsx
```diff
@@ -124,7 +124,7 @@ describe("PanelExtensionAdapter", () => {
       await Promise.resolve();
     });
     expect(renderStates).toEqual([
-      { currentFrame: [message], didSeek: false },
+      { currentFrame: [], didSeek: false }, // first frame is empty because there are no subscribers yet
```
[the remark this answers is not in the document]
snosenzo: because it would use the custom logic in MMPP that would add all of the existing messages to `messagesBySubscriberId` when `update-subscriber` was called. This now happens in a separate `update-player` action after `update-subscribers` was called.

packages/studio-base/src/components/MessagePipeline/store.ts
snosenzo: responded in slack, but `useMessageReducer` and `PanelExtensionAdapter` are memoized by the message array within `messagesBySubscriberId` and will not read in a new frame unless that array has changed

PR #7114, opened 2023-11-13
Attempt to resolve `package://` asset URLs relative to URDF url

achim-k:
**User-Facing Changes**
attempt to resolve `package://` asset URLs relative to URDF url

**Description**
Often, a URDF contains `package://` references to e.g. mesh files or xacro includes (converted to package:// URLs by [xacro-parser](https://www.npmjs.com/package/xacro-parser)). package:// resources can be loaded in 2 different ways:
- Through the player's fetchAsset method (if defined) - e.g. used for foxglove_websocket
- Through the Desktop app (if ROS_PACKAGE_PATH is defined)

This PR adds a third method in case the first two are not applicable: 

```
As last resort to load the package://<pkgName>/<pkgPath> URL, we resolve the package URL to
be relative of the base URL (which contains <pkgName> and is not a package:// URL itself).
Example:
   base URL: https://example.com/<pkgName>/urdf/robot.urdf
   resolved: https://example.com/<pkgName>/<pkgPath>
```

This allows loading entire URDFs from Github URLs, for instance:

- https://raw.githubusercontent.com/LeoRover/leo_common-ros2/humble/leo_description/urdf/leo.urdf.xacro
- https://raw.githubusercontent.com/LeoRover/leo_common-ros2/de35d0ee7/leo_description/urdf/leo.urdf.xacro

### DIFF
--- a/packages/studio-base/src/components/MessagePipeline/store.ts
+++ b/packages/studio-base/src/components/MessagePipeline/store.ts
@@ -166,29 +166,56 @@ export function createMessagePipelineStore({
         const { protocol } = new URL(uri);
         const player = get().player;
 
-        if (player?.fetchAsset && protocol === "package:") {
-          try {
-            return await player.fetchAsset(uri);
-          } catch (err) {
-            // Bail out if this is not a desktop app. For the desktop app, package:// is registered
-            // as a supported schema for builtin _fetch_ calls. Hence we fallback to a normal
-            // _fetch_ call if the asset couldn't be loaded through the player.
-            if (!isDesktopApp()) {
-              throw err;
+        const builtinFetch = async (url: string, opts?: { signal?: AbortSignal }) => {
+          const response = await fetch(url, opts);
+          if (!response.ok) {
+            const errMsg = response.statusText;
+            throw new Error(`Error ${response.status}${errMsg ? ` (${errMsg})` : ``}`);
+          }
+          return {
+            uri,
+            data: new Uint8Array(await response.arrayBuffer()),
+            mediaType: response.headers.get("content-type") ?? undefined,
+          };
+        };
+
+        if (protocol === "package:") {
+          // For the desktop app, package:// is registered as a supported schema for builtin _fetch_ calls.
+          const canBuiltinFetchPkgUri = isDesktopApp();
+          const pkgPath = uri.slice("package://".length);
+          const pkgName = pkgPath.split("/")[0];
+
+          if (player?.fetchAsset) {
+            try {
+              return await player.fetchAsset(uri);
+            } catch (err) {
+              if (canBuiltinFetchPkgUri) {
+                // Fallback to a builtin _fetch_ call if the asset couldn't be loaded through the player.
+                return await builtinFetch(uri, options);
+              }
+              throw err; // Bail out otherwise.
             }
+          } else if (canBuiltinFetchPkgUri) {
+            return await builtinFetch(uri, options);
+          } else if (
+            pkgName &&
+            options?.baseUrl != undefined &&
+            !options.baseUrl.startsWith("package://") &&
+            options.baseUrl.includes(pkgName)
+          ) {
+            // As last resort to load the package://<pkgName>/<pkgPath> URL, we resolve the package URL to
+            // be relative of the base URL (which contains <pkgName> and is not a package:// URL itself).
+            // Example:
+            //   base URL: https://example.com/<pkgName>/urdf/robot.urdf
+            //   resolved: https://example.com/<pkgName>/<pkgPath>
+            const resolvedUrl =
+              options.baseUrl.slice(0, options.baseUrl.lastIndexOf(pkgName)) + pkgPath;
+            return await builtinFetch(resolvedUrl, options);
           }
         }
 
-        const response = await fetch(uri, options);
-        if (!response.ok) {
-          const errMsg = response.statusText;
-          throw new Error(`Error ${response.status}${errMsg ? ` (${errMsg})` : ``}`);
-        }
-        return {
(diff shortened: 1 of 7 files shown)


defunctzombie:
How does this determine the `<pkg_name>` relative to the URDF url? When I have a urdf url? the referenced files don't have `package://` urls but instead use a `$(find)` directive which is a `xacro` thing and not a URDF thing right?

achim-k:
> How does this determine the `<pkg_name>` relative to the URDF url? When I have a urdf url? 

I take advantage of the fact, that the URDF URL often includes the package name, e,g,
`https://example.com/../<pkg_name>/urdf/robot.urdf`. So we can kind of defer the package root URL by removing everything after `<pkg_name>`  and resolve `package://<pkg_name>/<pkg_path>` URLs by appending `<pkg_path>` to the package root URL.

> the referenced files don't have `package://` urls but instead use a `$(find)` directive which is a `xacro` thing and not a URDF thing right?

Correct, but [xacro-parser](https://www.npmjs.com/package/xacro-parser) (which we use to parse urdf / xacro) substitutes `$(find ...)` to `package://` URLs.

defunctzombie:
> I take advantage of the fact, that the URDF URL often includes the package name, e,g,
> `https://example.com/../<pkg_name>/urdf/robot.urdf`. So we can kind of defer the package root URL by removing everything after `<pkg_name>` and resolve `package://<pkg_name>/<pkg_path>` URLs by appending `<pkg_path>` to the package root URL.

How do you know _where_ in the url the <pkg_name> is? Or are you doing this the other way around - you take a `package://<pkg_name>/<pkg_path>` url and you look for the <pkg_name> in the original url? Seems like a decent trick - though if you find it twice maybe that should be an error? Or we can deal with that when someone encounters it?

achim-k:
> How do you know _where_ in the url the <pkg_name> is? Or are you doing this the other way around - you take a `package://<pkg_name>/<pkg_path>` url and you look for the <pkg_name> in the original url?

Yes, that's how I do it.

> Seems like a decent trick - though if you find it twice maybe that should be an error? Or we can deal with that when someone encounters it?

When <pkg_name> occurs multiple times in the original URL the position of the last occurrence is used for determining the package root URL. Throwing an error could also be an option, although this would be done anyway if the fetch fails. I think we should also only do this if we are sure that we could otherwise not fetch the package:// url.

defunctzombie:
Sounds reasonable - let me know once the PR is ready for review.

Are you still concerned about this statement:

> Might cause misbehavior if the ROS_PACKAGE_PATH is set, so we might not want to do that in this case. Could also be disabled/enabled with an additional URDF setting

achim-k:
> Are you still concerned about this statement:
> 
> > Might cause misbehavior if the ROS_PACKAGE_PATH is set, so we might not want to do that in this case. Could also be disabled/enabled with an additional URDF setting

A little bit. I think we should only do this if we know that the package:// url cannot be fetched through either the player's fetchAsset method or through the desktop app (with ROS_PACKAGE_PATH). To achieve that, I would have to add this piece of code a few levels lower to the asset manager. Would this be acceptable?

defunctzombie:
> To achieve that, I would have to add this piece of code a few levels lower to the asset manager. Would this be acceptable?

I think so? Certainly we can try it and give that a review. I do think that moving more "asset loading" behavior out of the panel and into the asset layer makes sense in theory. I could see how maybe the panel could want some control over such things but in this instance am not coming up why moving it would be a bad thing.

We should add comments to make it clear what behavior we are supporting and why though cause once we do start supporting it it will be hard(er) to undo if people rely on it.

packages/studio-base/src/components/MessagePipeline/store.ts
```diff
@@ -166,29 +166,56 @@ export function createMessagePipelineStore({
...
+          return {
+            uri,
+            data: new Uint8Array(await response.arrayBuffer()),
+            mediaType: response.headers.get("content-type") ?? undefined,
+          };
+        };
+
+        if (protocol === "package:") {
+          // For the desktop app, package:// is registered as a supported schema for builtin _fetch_ calls.
+          const canBuiltinFetchPkgUri = isDesktopApp();
```
achim-k, line 184: Ideally, I would like to access the `ROS_PACKAGE_PATH` setting here so `canBuiltinFetchPkgUri` will be false if it is not set.

defunctzombie (approved):
I felt like `baseUrl` is somewhat misleading because it is actually the full url? At least the comment here seems to provide an example of a baseUrl that is the full url - maybe that's an error in the comment but the actual baseUrl is still a base without the final `file.urdf` part?

defunctzombie:
> Yes, this actually is the full URL for the current use case. I could not come up with a better name. Suggestions?

`url`?

achim-k:
> > Yes, this actually is the full URL for the current use case. I could not come up with a better name. Suggestions?
> 
> `url`? `assetUrl`? (though asset might be redundant when in-context)

Changed it now to `referenceUrl`

packages/studio-base/src/panels/ThreeDeeRender/ModelCache.ts
```diff
@@ -29,7 +29,8 @@ export type ModelCacheOptions = {
 
 type LoadModelOptions = {
   overrideMediaType?: string;
-  baseUrl?: string;
+  /** A URL to e.g. a URDf which may be used to resolve mesh package:// URLs */
```
defunctzombie, line 32: URDf -> URDF

nit: this comment says "which may be used to resolve mesh package:// URLs" in some ways keeps the reader guessing on what that means and in which situation it may be used. Something to think about for future commenting.

PR #7117, opened 2023-11-14
Add delay to image download stories to reduce flakiness

jtbandes:
**User-Facing Changes**
None

**Description**
Attempting to reduce flakiness


2metres (approved):
Worth a shot

PR #7121, opened 2023-11-15
Clear data on path change in state transition panel and show error for array data

cfoust:
**User-Facing Changes**
In the state transition panel, changing a path to an invalid one will correctly remove the existing data.
We also now show an error when the user's path refers to array data.

**Description**
These tasks were very similar, so I combined them.

image

### DIFF
--- a/packages/studio-base/src/components/TimeBasedChart/useProvider.tsx
+++ b/packages/studio-base/src/components/TimeBasedChart/useProvider.tsx
@@ -191,12 +191,12 @@ export default function useProvider<T>(
     }
 
     const bounds = getDatasetBounds(data.datasets);
-    if (bounds == undefined) {
-      return undefined;
-    }
 
     return {
-      bounds,
+      bounds: bounds ?? {
+        x: { min: 0, max: 0 },
+        y: { min: 0, max: 0 },
+      },
       data,
     };
   }, [data, state, getDatasetBounds, mergeState]);
--- a/packages/studio-base/src/panels/StateTransitions/index.tsx
+++ b/packages/studio-base/src/panels/StateTransitions/index.tsx
@@ -15,6 +15,7 @@ import { Add16Filled, Edit16Filled } from "@fluentui/react-icons";
 import { Button, Typography } from "@mui/material";
 import { ChartOptions, ScaleOptions } from "chart.js";
 import * as _ from "lodash-es";
+import * as R from "ramda";
 import { useCallback, useEffect, useMemo, useState } from "react";
 import { useResizeDetector } from "react-resize-detector";
 import tinycolor from "tinycolor2";
@@ -251,6 +252,17 @@ const StateTransitions = React.memo(function StateTransitions(props: Props) {
       // negative makes each path render below the previous
       const y = (pathIndex + 1) * 6 * -1;
       outMinY = Math.min(outMinY ?? y, y - 3);
+
+      const isEmpty = R.all(
+        (block) => R.all(({ queriedData }) => queriedData.length === 0, block[path.value] ?? []),
+        decodedBlocks,
+      );
+      if (isEmpty) {
+        outDatasets = outDatasets.concat({
+          data: [],
+        });
+        return;
+      }
 
       const blocksForPath = decodedBlocks.map((decodedBlock) => decodedBlock[path.value]);
 
--- a/packages/studio-base/src/panels/StateTransitions/index.tsx
+++ b/packages/studio-base/src/panels/StateTransitions/index.tsx
@@ -27,6 +27,7 @@ import { add as addTimes, fromSec, subtract as subtractTimes, toSec } from "@fox
 import * as PanelAPI from "@foxglove/studio-base/PanelAPI";
 import {
   MessageDataItemsByPath,
+  MessageAndData,
   useDecodeMessagePathsForMessagesByTopic,
 } from "@foxglove/studio-base/components/MessagePathSyntax/useCachedGetMessagePathDataItems";
 import useMessagesByPath from "@foxglove/studio-base/components/MessagePathSyntax/useMessagesByPath";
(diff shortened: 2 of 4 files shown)


cfoust:
fixing stories

cfoust:
done!

packages/studio-base/src/components/TimeBasedChart/useProvider.tsx
```diff
@@ -191,12 +191,12 @@ export default function useProvider<T>(
...
     const bounds = getDatasetBounds(data.datasets);
-    if (bounds == undefined) {
-      return undefined;
-    }
 
     return {
-      bounds,
+      bounds: bounds ?? {
+        x: { min: 0, max: 0 },
+        y: { min: 0, max: 0 },
```
jtbandes, line 198: what makes this a reasonable fallback? Seems like it would cause issues downstream like something trying to render a 0x0 viewport?

packages/studio-base/src/panels/StateTransitions/index.tsx
```diff
@@ -262,12 +266,34 @@ const StateTransitions = React.memo(function StateTransitions(props: Props) {
...
+
+      outIsArrayData.push(R.all((numPoints) => numPoints > 1, dataCounts));
+
+      const isEmpty = dataCounts.length === 0;
+      if (isEmpty) {
+        outDatasets = outDatasets.concat({
+          data: [],
+        });
+        return;
+      }
```
jtbandes, line 293: Is this the fix for the clearing data issue? Why does this fix the issue (why wasn't the data being cleared before)?

jtbandes (commented):
Is there a way to add a unit test for the clearing bug?

jtbandes:
@defunctzombie is more familiar with this panel than I am

packages/studio-base/src/components/TimeBasedChart/useProvider.tsx
cfoust: the bounds of the data do not determine the viewport
the problem is that our ChartJS worker uses `undefined` to mean that it should not change the currently rendered chart; providing an empty chart with zero bounds causes it to behave correctly

defunctzombie: In what instances does `getDatasetBounds` not return bounds? I think we want the chart to determine the bounds in those cases? It might be that some of these assumptions have changed and we always return bounds? Or should?

I'm thinking this could be related to live connections?

packages/studio-base/src/panels/StateTransitions/index.tsx
```diff
@@ -262,12 +266,34 @@ const StateTransitions = React.memo(function StateTransitions(props: Props) {
...
       });
 
-      outDatasets = outDatasets.concat(newBlockDataSets);
-
       // We have already filtered out paths we can find in blocks so anything left here
       // should be included in the dataset.
       const items = newItemsByPath[path.value];
-      if (items) {
+
+      const dataCounts = R.pipe(
```
defunctzombie, line 273: Personally I am not a fan of the new ramda code and still find myself struggling to understand what it is doing. Does pipe somehow chain the map functions so you only need to loop over the data once but can provide different transformations? How is this different to `Array.reduce`?

packages/studio-base/src/panels/StateTransitions/index.tsx
```diff
@@ -282,6 +308,7 @@ const StateTransitions = React.memo(function StateTransitions(props: Props) {
     return {
       datasets: outDatasets,
       minY: outMinY,
+      isArrayData: outIsArrayData,
```
defunctzombie, line 311: This should be per-dataset? Not for all datasets (see screenshot)

packages/studio-base/src/panels/StateTransitions/settings.ts
```diff
@@ -22,8 +22,14 @@ import { StateTransitionConfig, StateTransitionPath } from "./types";
 // the NodeEditor component is wrapped in a React.memo.
 
 const makeSeriesNode = memoizeWeak(
-  // eslint-disable-next-line @foxglove/no-boolean-parameters
-  (path: StateTransitionPath, index: number, canDelete: boolean): SettingsTreeNode => {
+  (
+    path: StateTransitionPath,
+    index: number,
+    // eslint-disable-next-line @foxglove/no-boolean-parameters
```
defunctzombie, line 28: IMO perfect example of why boolean parameters are bad - now that there are two of them reading the call site is worse. Can we pass an options object instead?

packages/studio-base/src/panels/StateTransitions/settings.ts
```diff
@@ -126,13 +144,14 @@ function buildSettingsTree(
...
     },
-    paths: makeRootSeriesNode(config.paths),
+    paths: makeRootSeriesNode(config.paths, isArrayData),
   };
 }
 
 export function useStateTransitionsPanelSettings(
   config: StateTransitionConfig,
   saveConfig: SaveConfig<StateTransitionConfig>,
+  isArrayData: boolean[],
```
defunctzombie, line 154: This could use a comment

defunctzombie:
I understand what `isArrayData` is doing but it does feel like it would be nicer if it traveled with its dataset to the settings builder rather than a sidecar field that has to match the number of datasets. What would it take to make that happen?

packages/studio-base/src/panels/StateTransitions/index.tsx
cfoust: you are correct--this is no longer necessary after the bounds fix. I had originally written this independently of the other fix, then combined them, and did not revisit this

packages/studio-base/src/panels/StateTransitions/settings.ts
cfoust: yup

packages/studio-base/src/components/TimeBasedChart/useProvider.tsx
cfoust: This is a nastier corner of the plot pipeline but `getDatasetBounds` returns `undefined` when the dataset is empty in lieu of representing this in some other way, so that downstream we can differentiate between a plot being empty vs. not having data yet.

cfoust:
@defunctzombie I followed your suggestion and made it so that the state of the path lives alongside `isArray` in its own `PathState` object. I agree that this was a strange pattern

packages/studio-base/src/panels/StateTransitions/index.tsx
cfoust: I [updated this code block](https://github.com/foxglove/studio/pull/7121/files#diff-00d14a5a16edac91e72a18185daded3b0f24abf3cb912c3162dd20e278ce291bR273-R280) to remove the subsequent `map` -- it can all be done in one pass. That should make this a little simpler. Regarding `reduce`, we're not aggregating any state here until the `uniq`, which does what it says on the tin.

As far as coding style and philosophy goes, that's probably a conversation better suited for a different medium (and one I'd be happy to have!)

defunctzombie: Please put a comment above this block - or separate it from the line above. I was reading the code and thought the comment about filtering items somehow also applied to this block of code.

defunctzombie (approved):

packages/studio-base/src/panels/StateTransitions/index.tsx
cfoust: done, added comment!

packages/studio-base/src/panels/StateTransitions/settings.ts
jtbandes: I think passing an object here defeats the memoizeWeak.

cfoust: Probably, but the state transition panel's perf is such a trash fire I can't imagine it makes much of a difference right now--we may have to revisit this soon anyway

PR #7126, opened 2023-11-16
Inject custom player via app context and delete UserScriptPlayer

defunctzombie:
**User-Facing Changes**
None

**Description**
Adds a `wrapPlayer` function to AppContext which is used by the PlayerManager to wrap the new player.

The User Script feature in our official releases (https://studio.foxglove.dev/ and https://foxglove.dev/download) is provided by a closed source extension, and Foxglove is no longer maintaining the open source version.


packages/studio-base/src/context/AppContext.ts
```diff
@@ -36,6 +38,7 @@ interface IAppContext {
   ) => StoreApi<WorkspaceContextStore>;
   PerformanceSidebarComponent?: React.ComponentType;
   extraPanels?: PanelInfo[];
+  buildPlayer: (child: Player, perfRegistry: IPerformanceRegistry) => Player;
```
jtbandes, line 41: maybe `wrapPlayer`?

jtbandes (approved):

PR #7128, opened 2023-11-16
Fix incorrect removal of unused topics from preload cache

achim-k:
**User-Facing Changes**
Fix removal of unused topics of the preloaded topics cache 

**Description**
When the block loader cache exceeds its limit, we attempt to remove already preloaded topics that are not needed anymore from existing blocks. We were however only checking for topic name equality and did not take into account if the subscription payload (e.g. message slicing fields) are different. This PR fixes it.

### DIFF
--- a/packages/studio-base/src/players/IterablePlayer/BlockLoader.ts
+++ b/packages/studio-base/src/players/IterablePlayer/BlockLoader.ts
@@ -340,6 +340,19 @@ export class BlockLoader {
         }
 
         const existingBlock = this.#blocks[currentBlockId];
+
+        // Calculate size of messages in the existing block that will be overridden.
+        // These have to be taken into account when calculating the size of the new block.
+        let overridenBlockMessagesSize = 0;
+        for (const topic of Object.keys(messagesByTopic)) {
+          const messages = existingBlock?.messagesByTopic[topic];
+          if (messages) {
+            overridenBlockMessagesSize += messages.reduce((acc, msg) => acc + msg.sizeInBytes, 0);
+          }
+        }
+        const newBlockSizeInBytes =
+          (existingBlock?.sizeInBytes ?? 0) - overridenBlockMessagesSize + sizeInBytes;
+
         this.#blocks[currentBlockId] = {
           needTopics: new Map(),
           messagesByTopic: {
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/players/IterablePlayer/BlockLoader.ts
```diff
@@ -127,8 +128,9 @@ export class BlockLoader {
         };
         const blockTopics = Object.keys(newMessagesByTopic);
         for (const topic of blockTopics) {
-          // remove topics that are no longer requested to be preloaded.
-          if (!topics.has(topic) && newMessagesByTopic[topic]) {
+          // remove topics that are no longer requested to be preloaded and topics that will
+          // be re-loaded (due to different subscription parameters)
+          if ((!topics.has(topic) || block.needTopics.has(topic)) && newMessagesByTopic[topic]) {
```
snosenzo, line 133: This looks risky, could block.needTopics get into a cycle where it's evicting and re-loading itself? Alternatively should we clear blocks with only specific fields when we receive a subscription that requires more fields?

we should definitely have a test case for this if possible.

defunctzombie:
> We were however only checking for topic name equality and did not take into account if the subscription payload (e.g. message slicing fields) are different.

How does this PR address that? I see the change to consult the _needs topics_ array for the topic before evicting but what does that have to do with message slicing?

achim-k:
Suppose that you have preloaded a certain topic up until the point that the cache is full. Now, your subscription to that topic changes and you only subscribe to a sub-field of that topic (message slicing). What do you want to do?
- Evict existing messages of that topic and repopulate the cache with the sliced messages instead (can load more messages). This is what this PR does
- Keep the messages in the cache since they already contain the sub-field. Will not require to re-read messages into the cache, but we won't be able to fully-preload the topic

---

Just found another problem:

https://github.com/foxglove/studio/blob/0977e3f553d0fc7bed19eabfad477536c32c6b96/packages/studio-base/src/players/IterablePlayer/BlockLoader.ts#L343-L348

When calculating `sizeInBytes`, we do not take into account the size of messages of the existing block that are overridden

defunctzombie (approved):
Approving this after talking in slack. I understand this fix here. The naming/terminology was not updated well when slicing was added which makes it confusing to understand why we want to remove a topic that we think we need. That has happened is that the topic is actually stale.

PR #7130, opened 2023-11-16
Raise error if channel schema is empty

achim-k:
**User-Facing Changes**
Raise error if channel schema is empty for a ROS channel.

**Description**
Raises an error if the channel schema data is empty and the schema name is not one of the well-known empty schemas ( `std_msgs/Empty` or `std_msgs/msg/Empty`). An empty schema is often a sign that something went wrong when recording messages (e.g. message definition could not be found). This change makes sure that this problem is surfaced to the user.

Resolves FG-5451

### DIFF
--- a/packages/mcap-support/src/parseChannel.ts
+++ b/packages/mcap-support/src/parseChannel.ts
@@ -25,6 +25,8 @@ export type ParsedChannel = {
   datatypes: MessageDefinitionMap;
 };
 
+const KNOWN_EMPTY_SCHEMA_NAMES = ["std_msgs/Empty", "std_msgs/msg/Empty"];
+
 function parseIDLDefinitionsToDatatypes(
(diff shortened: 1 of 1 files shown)


jtbandes (approved):
Seems alright – I could imagine a world where someday somebody comes back to us saying this should not be an error. But that probably won't happen for a while.

defunctzombie:
> I could imagine a world where someday somebody comes back to us saying this should not be an error

I'll be interested to hear their reasoning.

defunctzombie (commented):
Do we want to limit this behavior specifically to ROS encoded channels? protobuf encoded channels? I'm wondering if this will be an issue for someone's existing json channels?

defunctzombie (approved):

jtbandes:
Related: https://linear.app/foxglove/issue/FG-5754/support-empty-mcap-schemas

PR #7131, opened 2023-11-16
Do not assume 0 bytes for size of sliced messages

achim-k:
**User-Facing Changes**
Not worth being called out

**Description**
Instead of assuming the size of sliced messages to be 0, we use an estimate that is closer to the actual memory usage. This makes OOM crashes slightly less likely. 

Resolves FG-5897

### DIFF
--- a/packages/studio-base/src/players/IterablePlayer/Mcap/McapIndexedIterableSource.ts
+++ b/packages/studio-base/src/players/IterablePlayer/Mcap/McapIndexedIterableSource.ts
@@ -18,7 +18,7 @@ import {
 } from "@foxglove/studio-base/players/IterablePlayer/IIterableSource";
 import {
   OBJECT_BASE_SIZE,
-  estimateMessageObjectSize,
+  estimateMessageFieldSizes,
 } from "@foxglove/studio-base/players/messageMemoryEstimation";
 import { PlayerProblem, Topic, TopicStats } from "@foxglove/studio-base/players/types";
 import { RosDatatypes } from "@foxglove/studio-base/types/RosDatatypes";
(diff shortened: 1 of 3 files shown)


jtbandes (dismissed):
Better than 0 but still not very accurate because the field could be a nested message that contains a ton of sub-fields 😓

packages/studio-base/src/players/IterablePlayer/Mcap/McapIndexedIterableSource.ts
```diff
@@ -175,10 +179,10 @@ export class McapIndexedIterableSource implements IIterableSource {
         const msg = channelInfo.parsedChannel.deserialize(message.data) as Record<string, unknown>;
         const spec = args.topics.get(channelInfo.channel.topic);
         const payload = spec?.fields != undefined ? pickFields(msg, spec.fields) : msg;
-        const sizeInBytes = Math.max(
-          message.data.byteLength,
-          channelInfo.approxDeserializedMsgSize,
-        );
+        const sizeInBytes =
+          spec?.fields == undefined
```
defunctzombie, line 183: This has problably reached a point where it could be a function in the `messageMemoryEstimation` file? Something about estimating the size of a slice or some such?

defunctzombie (approved):

achim-k:
> Better than 0 but still not very accurate because the field could be a nested message that contains a ton of sub-fields 😓

We could do a more involved memory estimation which would take sub-fields into account. The result could be stored in a map and used for the next message on the same topic

packages/studio-base/src/players/IterablePlayer/Mcap/McapIndexedIterableSource.ts
```diff
@@ -153,6 +156,31 @@ export class McapIndexedIterableSource implements IIterableSource {
 
     const topicNames = Array.from(topics.keys());
 
+    // Estimate memory size of sliced schemas.
+    const slicedMessageMemSize = new Map<number, number>();
```
snosenzo, line 160: This definitely shouldn't be done for every message. I think this should be a function that should be memoized by schemaName and fieldName since it really only needs to be done once per field.

snosenzo (approved):

PR #7132, opened 2023-11-16
More theme consolidation work

2metres:
**User-Facing Changes**
None

**Description**
Set sticky headers off by default and address styling


jtbandes (approved):

jtbandes (commented):
As a follow-up can you add a story that covers this in this repo?

PR #7140, opened 2023-11-17
Skip Chart frames if we fall behind

cfoust:
**User-Facing Changes**
Minor improvements to rendering performance for the Plot and State Transition panels.

**Description**
The Chart component, which is used by the State Transition and Plot panels, does not currently have a way to drop frames if it starts to fall behind. This PR adds that functionality by coalescing rendering jobs together if we're not keeping up.

### DIFF
--- a/packages/studio-base/src/components/Chart/index.stories.tsx
+++ b/packages/studio-base/src/components/Chart/index.stories.tsx
@@ -187,6 +187,12 @@ export const AllowsClickingOnDatalabels: StoryObj = {
       }
     }, [clickedDatalabel, readySignal]);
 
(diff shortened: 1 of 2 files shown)


cfoust:
I rebased this PR and some of its stories are failing in CI:
image
I can't reproduce them locally, even with the method I devised, so something else must be going on.

cfoust:
I inspected each story and while this is not perfect, it does appear that this PR fixes far more stories than it breaks and given the fact that it's strictly a timing fix, it should improve things. I only see one questionable story (the index-based one) that we may need to revisit.

PR #7143, opened 2023-11-17
A few more Ramda lint rules

jtbandes:
**User-Facing Changes**
None

**Description**
- `R.keys/values` -> `Object.keys/values`
- `R.find` -> `Array#find`
- `R.all/any` -> `Array#every/some`

Relates to FG-4805

### DIFF
--- a/packages/studio-base/src/components/Chart/datasets.ts
+++ b/packages/studio-base/src/components/Chart/datasets.ts
@@ -85,13 +85,13 @@ export function* iterateTyped<T extends { [key: string]: Array<any> | Float32Arr
   let index = 0;
   for (const slice of dataset) {
     // Find a property for which we can check the length
-    const first = R.head(R.values(slice));
+    const first = R.head(Object.values(slice));
     if (first == undefined) {
       continue;
     }
 
     for (let j = 0; j < first.length; j++) {
-      for (const key of R.keys(slice)) {
+      for (const key of Object.keys(slice) as (keyof typeof slice)[]) {
         point[key] = slice[key]?.[j];
       }
 
--- a/packages/eslint-plugin-studio/ramda-usage.js
+++ b/packages/eslint-plugin-studio/ramda-usage.js
@@ -12,8 +12,9 @@ module.exports = {
     type: "problem",
     fixable: "code",
     messages: {
-      useMath: `Use built-in Math.{{name}} instead of R.{{name}} when applying arguments directly`,
-      useArrayMap: `Use built-in Array#map instead of R.map`,
+      useMath: `Use built-in Math.{{name}} instead of R.{{name}}`,
+      useObject: `Use built-in Object.{{name}} instead of R.{{name}}`,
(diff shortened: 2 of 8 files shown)


packages/studio-base/src/components/Chart/datasets.ts
```diff
@@ -85,13 +85,13 @@ export function* iterateTyped<T extends { [key: string]: Array<any> | Float32Arr
...
     // Find a property for which we can check the length
-    const first = R.head(R.values(slice));
+    const first = R.head(Object.values(slice));
     if (first == undefined) {
       continue;
     }
 
     for (let j = 0; j < first.length; j++) {
-      for (const key of R.keys(slice)) {
+      for (const key of Object.keys(slice) as (keyof typeof slice)[]) {
```
defunctzombie, line 94: Why the cast?

defunctzombie (approved):
What was the genesis of this? Did we decide that using the builtins is more idiomatic?

packages/studio-base/src/components/Chart/datasets.ts
jtbandes: `Object.keys` has poor TS types :( https://www.mattstobbs.com/object-keys-typescript/

jtbandes:
It is my small attempt to hold back the flood of Ramda everywhere  (and I don't believe there's any reason to use the Ramda ones over the builtin ones? seems like they could only be slower?)

packages/studio-base/src/components/Chart/datasets.ts
cfoust: lol, this begs the question of why we're banning R.keys then, right?

cfoust:
Is this something we've decided on?

jtbandes:
Nope it was just one of the things I noted down on my wishlist in FG-4805. Feel free to debate it here

packages/studio-base/src/components/Chart/datasets.ts
jtbandes: We could exclude this one if it seems like a significant benefit (or we could make the lint rule go the other way and enforce `R.keys` over `Object.keys`). Will note that the repo currently has 46 uses of `Object.keys` and 3 uses of `R.keys`.

defunctzombie: We could override the TS types for Object.keys like we do for json stringify.

defunctzombie:
I like the idea of using the builtin language functions where available and don't know why I'd turn to the ramda variants for that. I would turn to ramda for its other features. Though I would also ask what it does that lodash does not? Maybe it has more functions so we should use it instead of lodash? Or do the two offer different things?

jtbandes:
Merging as-is for now since this is a path to consistency with relatively minimal changes. If someone else wants to champion some Object.keys type overrides or migrating to R.keys everywhere, I would be open to that as well. We've learned that both Ramda and TS are kind of lying with their type definitions here – pick your poison.

PR #7149, opened 2023-11-20
Allow service request/response schemas to be empty

achim-k:
**User-Facing Changes**
Not worth mentioning

**Description**
Fix falsely raising an error on empty ROS service request / response schemas. Since #7130 we are raising an error when a channel's schema is empty, but we shouldn't do this for service schemas as they can be often empty.

### DIFF
--- a/packages/mcap-support/src/parseChannel.ts
+++ b/packages/mcap-support/src/parseChannel.ts
@@ -74,6 +74,9 @@ function parsedDefinitionsToDatatypes(
  * Process a channel/schema and extract information that can be used to deserialize messages on the
  * channel, and schemas in the format expected by Studio's RosDatatypes.
  *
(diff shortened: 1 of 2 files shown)


packages/mcap-support/src/parseChannel.ts
```diff
@@ -78,10 +78,14 @@ function parsedDefinitionsToDatatypes(
  * - https://github.com/foxglove/mcap/blob/main/docs/specification/well-known-message-encodings.md
```
defunctzombie, line 78: Add a comment that empty schemas (except std_msgs/Empty) are treated as errors. If you want to allow empty schemas then use the `allowEmptySchemas` option.

defunctzombie (approved):

jtbandes:
What do we do with the return value of parseChannel for service request/response schemas? It is still true that if the schema is empty we're not going to be able to serialize/deserialize the messages.

PR #7153, opened 2023-11-21
Update subscriptions on topic alias changes

defunctzombie:
**User-facing changes**
Topic alias subscriptions are updated when changing global variables.

**Description**
This change fixes a behavior where changes to global variables would not cause updated subscriptions for the wrapped player. This would happen because aliased subscriptions were only updated in the `setSubscriptions` call. However, for some panel workflows they would not re-subscribe when the topics list changed because they were already subscribed to the topic they wanted (even though it was not yet present in the topic list). Plot panel is an example of such a panel. Because they had already specified their subscriptions there was no additional call to `setSubscriptions` and no re-calculating of aliased subscriptions for the original topic.

This change updates the `TopicAliasingPlayer` logic to re-process aliased subscriptions and call `setSubscriptions` on the wrapped player for all three workflows: setGlobalVariables, setSubscriptions, and player state updates. Now if a panel had previously subscribed that subscription will be properly aliased.

I've also changed the `setSubscriptions` behavior. Previously it would not set subscriptions until topics were received from the underlying player. I see no reason to keep this restriction since alias functions could be hard-coded and not require any input topics.

### DIFF
--- a/packages/studio-base/src/players/IterablePlayer/BlockLoader.ts
+++ b/packages/studio-base/src/players/IterablePlayer/BlockLoader.ts
@@ -91,8 +91,7 @@ export class BlockLoader {
 
     this.#abortController.abort();
     this.#activeChangeCondvar.notifyAll();
-    // fixme - this join has [object Object] - this is no longer an array of topics...?
-    log.debug(`Preloaded topics: ${[...topics].join(", ")}`);
+    log.debug(`Preloaded topics: ${Array.from(topics.keys()).join(", ")}`);
 
     // Update all the blocks with any missing topics
     for (const block of this.#blocks) {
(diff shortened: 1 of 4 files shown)


packages/studio-base/src/players/IterablePlayer/BlockLoader.ts
```diff
@@ -91,7 +91,7 @@ export class BlockLoader {
 
     this.#abortController.abort();
     this.#activeChangeCondvar.notifyAll();
-    log.debug(`Preloaded topics: ${[...topics].join(", ")}`);
+    log.debug(`Preloaded topics: ${Array.from(topics.keys()).join(", ")}`);
```
defunctzombie, line 94: topics is a map so this prints the topic names (keys). Before it would print the name + [object Object]

jtbandes (approved):
If you say so

PR #7162, opened 2023-11-23
Fix automatic change in point cloud settings not being made visible to user

achim-k:
**User-Facing Changes**
Fix automatic change in point cloud settings not being made visible to user

**Description**
We were updating the settings but the change was not immediately made visible to the user

Resolves FG-5832


snosenzo (approved):
looks good, but definitely call the fix out in user facing changes. also not sure if there's a ticket attached to this, but would be good to include that in the description as well.

jtbandes (approved):

PR #7167, opened 2023-11-27
Limit the maximum number of points returned by downsampleTimeseries

cfoust:
**User-Facing Changes**
Improve rendering performance for plots with large datasets.

**Description**
One of our discoveries while testing [the downsampling PR](https://github.com/foxglove/studio/pull/6997) was that it was practical to add an upper bound to the original downsampling algorithm, which appears to do a better job of capturing the outliers of a plot. This backports that insight to `main`.

It's worth noting that in an effort to keep this diff as minimal as possible, I omitted the change that hides the dots at each data point if any downsampling occurred.

### DIFF
--- a/packages/studio-base/src/components/TimeBasedChart/downsample.test.ts
+++ b/packages/studio-base/src/components/TimeBasedChart/downsample.test.ts
@@ -65,9 +65,12 @@ describe("downsampleTimeseries", () => {
         { x: 1, y: 4, value: 4 },
         { x: 2, y: 5, value: 5 },
       ]),
-      bounds,
+      {
+        ...bounds,
+        width: bounds.width * 4,
+      },
     );
-    expect(result).toEqual([0, 2, 4]);
+    expect(result).toEqual([0, 1, 2, 3, 4]);
   });
 });
 
--- a/packages/studio-base/src/components/TimeBasedChart/downsample.test.ts
+++ b/packages/studio-base/src/components/TimeBasedChart/downsample.test.ts
@@ -67,7 +67,7 @@ describe("downsampleTimeseries", () => {
       ]),
       bounds,
     );
-    expect(result).toEqual([0, 1, 2, 3, 4]);
+    expect(result).toEqual([0, 2, 4]);
   });
 });
 
--- a/packages/studio-base/src/components/TimeBasedChart/downsample.test.ts
+++ b/packages/studio-base/src/components/TimeBasedChart/downsample.test.ts
@@ -4,7 +4,7 @@
 
 import { iterateObjects } from "@foxglove/studio-base/components/Chart/datasets";
 
-import { downsampleTimeseries, downsampleScatter } from "./downsample";
+import { MINIMUM_PIXEL_DISTANCE, downsampleTimeseries, downsampleScatter } from "./downsample";
 
 describe("downsampleTimeseries", () => {
   const bounds = {
@@ -67,7 +67,7 @@ describe("downsampleTimeseries", () => {
       ]),
       {
         ...bounds,
-        width: bounds.width * 4,
+        width: bounds.width * MINIMUM_PIXEL_DISTANCE,
       },
     );
     expect(result).toEqual([0, 1, 2, 3, 4]);
--- a/packages/studio-base/src/components/TimeBasedChart/downsample.ts
+++ b/packages/studio-base/src/components/TimeBasedChart/downsample.ts
@@ -15,10 +15,21 @@ type Dataset<T> = ChartDataset<"scatter", T>;
 // 60FPS.
 export const MAX_POINTS = 5_000;
 
+// Each interval can produce up to this many points
+const POINTS_PER_INTERVAL = 4;
+
+// Points that appear within this threshold are visually indistinguishable
+export const MINIMUM_PIXEL_DISTANCE = 3;
+
 /**
- * Downsample a timeseries dataset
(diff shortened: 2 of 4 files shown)


packages/studio-base/src/components/TimeBasedChart/downsample.test.ts
```diff
@@ -67,7 +67,7 @@ describe("downsampleTimeseries", () => {
       ]),
       bounds,
     );
-    expect(result).toEqual([0, 1, 2, 3, 4]);
+    expect(result).toEqual([0, 2, 4]);
```
defunctzombie, line 70: This looks like we lost the min/max at x=1?

packages/studio-base/src/components/TimeBasedChart/downsample.ts
```diff
@@ -32,13 +37,19 @@ type Dataset<T> = ChartDataset<"scatter", T>;
...
   const { bounds, width, height } = view;
 
-  const pixelPerXValue = width / (bounds.x.max - bounds.x.min);
+  // Each interval can produce up to four points
+  const numIntervals = (numPoints ?? width) / 4;
+  const pixelPerXValue = numIntervals / (bounds.x.max - bounds.x.min);
   const pixelPerYValue = height / (bounds.y.max - bounds.y.min);
 
-  const downsampled: number[] = [];
+  const indices: number[] = [];
```
defunctzombie, line 52: 👍 took me a while to realize that's what this output is now

packages/studio-base/src/components/TimeBasedChart/downsample.ts
```diff
@@ -32,13 +37,19 @@ type Dataset<T> = ChartDataset<"scatter", T>;
...
+export function downsampleTimeseries(
+  points: Iterable<Point>,
+  view: PlotViewport,
+  numPoints?: number,
+): number[] {
   const { bounds, width, height } = view;
 
-  const pixelPerXValue = width / (bounds.x.max - bounds.x.min);
+  // Each interval can produce up to four points
+  const numIntervals = (numPoints ?? width) / 4;
```
defunctzombie, line 48: This favors numPoints rather than the width for capping the number of intervals which means we use more intervals than we should when the width is smaller.

My original PR had this:
```
const cappedWidth = Math.min(width, 5000 / numSeries);
```

Which capped the number of intervals per series but preferred letting the width cap it first.

Am I reading this new logic wrong?

packages/studio-base/src/components/TimeBasedChart/downsample.test.ts
cfoust: I had typed up a comment directed towards you and forgot to submit it--I did not adjust the algorithm beyond these top-level parameters, going to investigate why this is different

packages/studio-base/src/panels/Plot/useDatasets.worker.ts
```diff
@@ -6,7 +6,7 @@ import * as Comlink from "comlink";
 
 import { Immutable } from "@foxglove/studio";
 import { iterateTyped } from "@foxglove/studio-base/components/Chart/datasets";
-import { downsample } from "@foxglove/studio-base/components/TimeBasedChart/downsample";
+import { downsample, MAX_POINTS } from "@foxglove/studio-base/components/TimeBasedChart/downsample";
```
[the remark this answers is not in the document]
cfoust: We use it [in the incremental downsampling PR](https://github.com/foxglove/studio/pull/7168/files#diff-eeabcedb24ddfb4cfb7c29dc3204b550b55655e878ce3433fcbe4cb3b841bda1R15) so it's useful to have the maximum in one specific place (but I'm open to putting it wherever)

packages/studio-base/src/components/TimeBasedChart/downsample.ts
cfoust: There was a [part of our conversation](https://foxglovehq.slack.com/archives/C036C8HGCGN/p1700674346455709?thread_ts=1700242039.583979&cid=C036C8HGCGN) where I explain why I had to change this-- my preference with this line is to show as much detail as possible since we're sort of violating the N pixels = M points relationship, but I might be missing something

packages/studio-base/src/components/TimeBasedChart/downsample.test.ts
cfoust: I realized that this was because we were not dividing the viewport width by 4 in the old version, so to fix the test (that is, to ensure the number of intervals is the same as it used to be) I just multiply the width by 4

packages/studio-base/src/components/TimeBasedChart/downsample.ts
defunctzombie: Shouldn't we prefer width because the width may already limit the number of output points under our max per series?

packages/studio-base/src/components/TimeBasedChart/downsample.test.ts
defunctzombie: Why would we need to divide the viewport width by 4 if we are already under the point cap?

packages/studio-base/src/components/TimeBasedChart/downsample.ts
cfoust: ie I understand this to mean that we might do slightly more work than we have to in some circumstances (specifically when `width` < `numPoints / 4`, but it seems negligible?

defunctzombie (approved):

PR #7171, opened 2023-11-27
Add Image Panel empty state

snosenzo:
**User-Facing Changes**

- Image panel will now display over canvas that it is waiting for messages

**Description**
Allows SceneExtensions to display heads-up messages to the DOM over the canvas. 

Empty state (3 cases, all have stories): 
image

Notice state: for video keyframes (private), sync annotations, waiting for image messages when calibration is present
image

FG-5459


2metres:
I feel like we've tangled up 2 different types of messages here: 
- **Information messages** that appear when some of the content is not displaying for a reason
- **Empty States** that are displated when there isn't content to show.

### Examples of information messages
“Awaiting synced annotations...” Sync annotations is turned on and the panel hasn’t received messages on the active annotation topics that match the the timestamp of the last image.
“Waiting for video keyframe...” This is quirk of programming

In the case of information messages I feel like using the toast/chip style makes sense but am cautious to only show these 1 at a time and make them dismissable

Screen Shot 2023-11-28 at 4 06 38 pm

### Examples of empty states
No data source: Layout/config present but no data “Selected topics do not exist on data source”
No data present on selected topic “Waiting for data on selected topics: /CAM_BACK_LEFT/image_raw”

In the case of empty states I feel like matching the other empty states seen in other panels make sense.

Screen Shot 2023-11-29 at 11 50 24 am

packages/studio-base/src/panels/ThreeDeeRender/HUD.tsx
```diff
@@ -2,23 +2,35 @@
...
     position: "absolute",
-    alignContent: "center",
     top: 0,
-    padding: theme.spacing(1),
-    border: "1px solid black",
-    background: "grey",
-    select: "none",
+    left: "50%",
+    transform: "translateX(-50%)",
+    pointerEvents: "none",
```
2metres, line 22: Not sure why but is interfering with pointerEvents on the three.js canvas

jtbandes: @snosenzo was this comment addressed yet?

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/MessageHandler.ts
```diff
@@ -74,6 +76,26 @@ type RenderStateListener = (
   oldState: MessageRenderState | undefined,
 ) => void;
 
+const WAITING_FOR_BOTH_HUD_ITEM: HUDItem = {
+  id: "WAITING_FOR_BOTH_MESSAGES",
+  group: IMAGE_MODE_HUD_GROUP_ID,
+  message: "Waiting for messages…",
```
jtbandes, line 82: consider adding these to `i18n/*/threeDee.ts` allowing them to be translated – since we already have a lot of 3D panel translated

packages/studio-base/src/panels/ThreeDeeRender/HUDManager.ts
```diff
@@ -0,0 +1,63 @@
...
+    }
+  }
+
+  public removeHUDItem(id: string): void {
+    if (this.#HUDItemsById.delete(id)) {
+      this.#onChange();
+    }
+  }
+
+  public removeGroup(group: string): void {
```
jtbandes, line 25: what do you envision the groups being used for? it seems like we only have one group today?

packages/studio-base/src/panels/ThreeDeeRender/HUDManager.ts
```diff
@@ -0,0 +1,63 @@
...
+    return Array.from(this.#HUDItemsById.values());
+  }
+
+  public clear(): void {
+    this.#HUDItemsById.clear();
+    this.#onChange();
+  }
+}
+
+export type HUDItem = {
```
jtbandes, line 54: nit: move to top

jtbandes (commented):
General approach seems fine. I'm not a huge fan of the generic "HUD" name since this seems targeted at being more of an "error overlay" for now – but I don't really have better suggestions.

I'm still able to see a completely blank panel sometimes:

https://github.com/foxglove/studio/assets/14237/0292716a-b9ed-4f1f-a388-b0824f3de6fb

packages/studio-base/src/panels/ThreeDeeRender/HUD.tsx
snosenzo: When I tested it it was working. What specifically wasn't working @2metres ?

packages/studio-base/src/panels/ThreeDeeRender/HUDManager.ts
snosenzo: So that scene extensions can clean up their own or related HUD items

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageMode.ts
```diff
@@ -83,6 +88,12 @@ const DEFAULT_FOCAL_LENGTH = 500;
 
 const REMOVE_IMAGE_TIMEOUT_MS = 50;
 
+const NO_IMAGE_TOPICS_HUD_ITEM: HUDItem = {
+  id: "NO_IMAGE_TOPICS",
+  group: IMAGE_MODE_HUD_GROUP_ID,
+  message: t3D("noImageTopicsAvailable"),
```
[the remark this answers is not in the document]
snosenzo: I changed the HUDItem `message` field to `getMessage: () => string` so that we can call the message function only when we need to access it. I think that should work for t(), hope it's not too weird, but I didn't want to change these to not be constants so that they're relatively central.

packages/studio-base/src/panels/ThreeDeeRender/t3D.ts
```diff
@@ -0,0 +1,9 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { t } from "i18next";
+
+import { threeDee } from "@foxglove/studio-base/i18n/en";
+
+export const t3D = (threeDeeKey: keyof typeof threeDee): string => t(`threeDee:${threeDeeKey}`);
```
jtbandes, line 9: this feels like a lot of effort for a very minimal abstraction - wouldn't hurt much to just use `t("threeDee:foo")` in other places instead of `t3D`

packages/studio-base/src/panels/ThreeDeeRender/LayerErrors.ts
```diff
@@ -111,6 +111,23 @@ export class LayerErrors extends EventEmitter<LayerErrorEvents> {
...
   }
 
+  /**
+   * If value is falsy then add error to path, otherwise remove error from settings path
+   * @param value - value to check, if false, add error, if true, remove error
+   * @param path  - path to add/remove error
+   * @param errorId - id unique to error
+   * @param errorMessage - error message
+   */
+  public errorIfFalsey(value: unknown, path: Path, errorId: string, errorMessage: string): void {
```
jtbandes, line 121: seems like only booleans are being passed in here

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/MessageHandler.test.ts
```diff
@@ -34,42 +43,62 @@ function wrapInMessageEvent<T>(
 }
 
 describe("MessageHandler: synchronized = false", () => {
+  let hud = new HUDItemManager(() => {});
+  beforeEach(() => {
+    hud = new HUDItemManager(() => {});
+  });
```
jtbandes, line 49: could the variable just be moved into the test body instead of being global?

jtbandes (approved):
Seems pretty good, but I think this should be an empty state:

image

Also not sure how it should interact with the background color setting:

image

some bg colors make it invisible:

image

bg color setting ignored for empty state:

image


Not sure there is anything else we need to do to address this, just calling it out as a possible edge case.

packages/studio-base/src/panels/ThreeDeeRender/t3D.ts
snosenzo: I like that it gives me options for `keyof typeof threeDee` while typing out though. I don't think `t(` did that

snosenzo: `t` does give completion options, but it gets a little tough after typing the `:`

packages/studio-base/src/panels/ThreeDeeRender/LayerErrors.ts
snosenzo: I'm going to leave this as is because I wanted it to kinda work how assert works (though I know `assert` has more purposes), and if I type it as a boolean I have to add a lint rule

2metres:
Seems strange to have 2 messages like this when really only one makes sense: 
Screen Shot 2023-12-12 at 9 32 44 am

packages/studio-base/src/panels/ThreeDeeRender/LayerErrors.ts
snosenzo: nevermind I changed it

snosenzo:
@2metres Updated to include priority list of empty states so that the only highest priority empty state in the HUD is shown.

PR #6221, opened 2023-06-09
[Image Beta] Improve flickering while scrubbing

jtbandes:
**User-Facing Changes**
None

**Description**
Reverts #6175. Instead, remove image renderable only after a short delay, canceling the removal if a new image message is received. Canceling the removal allows an old bitmap to be displayed when it's done decoding, even if a new decode has already started.

Before vs after:

https://github.com/foxglove/studio/assets/14237/d97ea048-64b8-4621-9f9c-f594efe466ea

With and without synchronization:

https://github.com/foxglove/studio/assets/14237/b46d6299-8bd3-4ba8-a880-8d07979371f6

PR #7173, opened 2023-11-27
Fix Image panel images not clearing when looping playback

snosenzo:
## User-facing changes

 - Fix Image panel images not clearing when looping playback

## Description

Image renderables weren't being cleared on seek for two reasons. The first being that we weren't queuing an animationframe once the timeout is up, and the second being that if the next image message would fail decoding, then we would clear the timeout so the renderable would still have a stale final image being displayed. To fix these, I've added a queueAnimation frame in the timeout and now when erroring we show an error image instead of the previous frame. 

Fixes: FG-5772


jtbandes (approved):
Hm... I wonder if we would need to change the `REMOVE_IMAGE_TIMEOUT_MS` now? The decoding delay will mean we are closer to hitting the timeout on a regular basis. It would be good to do some manual testing to confirm the flickering while scrubbing has not regressed — see https://github.com/foxglove/studio/pull/6221

jtbandes:
Can the new error image be shown in a story?

jtbandes (approved):

PR #7179, opened 2023-11-28
avoid unnecessary plot legend re-render

achim-k:
**User-Facing Changes**
not worth calling out

**Description**

Before:

[Screencast from 27.11.2023 20:57:49.webm](https://github.com/foxglove/studio/assets/9250155/ff7406ec-f8c8-42bb-b7d6-491856ec6ffb)

After (this PR):

[Screencast from 27.11.2023 21:06:06.webm](https://github.com/foxglove/studio/assets/9250155/162dab0f-3bb9-4ab4-a9a5-c58ee0d874a5)

### DIFF
--- a/packages/studio-base/src/panels/Plot/index.tsx
+++ b/packages/studio-base/src/panels/Plot/index.tsx
@@ -15,6 +15,7 @@ import { useTheme } from "@mui/material";
 import * as _ from "lodash-es";
 import { ComponentProps, useCallback, useEffect, useMemo, useState } from "react";
 
+import { useDeepMemo } from "@foxglove/hooks";
 import {
   Time,
   add as addTimes,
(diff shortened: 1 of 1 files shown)


defunctzombie (approved):
Any easy hope of whatever produces `pathsWithMismatchedDataLengths` with empty lists using a stable object for empty? We do that in some other places where if the list is empty we have some constant like `EMPTY_ARRAY` or some such at the file level that is used.

achim-k:
> Any easy hope of whatever produces `pathsWithMismatchedDataLengths` with empty lists using a stable object for empty? We do that in some other places where if the list is empty we have some constant like `EMPTY_ARRAY` or some such at the file level that is used.

Tried that, but the problem is the `structuredClone` when copying the object from the worker

```js
const empty = [];
const a = { empty };
const b = { empty };
const c = structuredClone(a);
console.log(Object.is(a.empty, b.empty));  // --> true
console.log(Object.is(a.empty, c.empty));  // --> false
```

PR #7181, opened 2023-11-28
Update @node/types, handle errors that follow

james-rms:
**User-Facing Changes**

**Description**

Updates our @types/node dependency resolution to 20.10.* . This includes type definitions for `fetch`, which conflict for the `dom.d.ts` definitions for fetch, so some wrangling needs to be done in overwriteFetch.

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     ]
   },
   "resolutions": {
-    "@types/node": "16.18.40",
+    "@types/node": "20.10.0",
     "react-use": "patch:react-use@17.4.0#./patches/react-use.patch",
     "react-dnd": "patch:react-dnd@npm:16.0.1#./patches/react-dnd.patch"
   },
--- a/packages/studio-base/src/players/IterablePlayer/Mcap/McapIterableSource.test.ts
+++ b/packages/studio-base/src/players/IterablePlayer/Mcap/McapIterableSource.test.ts
@@ -17,7 +17,14 @@ describe("McapIterableSource", () => {
     await writer.start({ library: "", profile: "" });
     await writer.end();
 
-    const source = new McapIterableSource({ type: "file", file: new Blob([tempBuffer.get()]) });
+    const source = new McapIterableSource({
+      type: "file",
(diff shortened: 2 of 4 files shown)


packages/studio-base/src/players/IterablePlayer/Mcap/McapIterableSource.test.ts
```diff
@@ -17,7 +17,10 @@ describe("McapIterableSource", () => {
     await writer.start({ library: "", profile: "" });
     await writer.end();
 
-    const source = new McapIterableSource({ type: "file", file: new Blob([tempBuffer.get()]) });
+    const source = new McapIterableSource({
+      type: "file",
+      file: new Blob([tempBuffer.get()]) as unknown as globalThis.Blob,
```
james-rms, line 22: this is another weird conflict between dom types and node types. This code is meant to run in a node context. In that context, `Blob` is not defined globally at runtime (but it is in @types/node!?).

package.json
```diff
@@ -51,7 +51,7 @@
     ]
   },
   "resolutions": {
-    "@types/node": "16.18.40",
+    "@types/node": "^20.10.0",
```
defunctzombie, line 54:
```suggestion
    "@types/node": "20.10.0",
```

defunctzombie: I'm not a fan of `^` in our dependencies and since we our not a library and have a lockfile this also doesn't do anything for us.

packages/studio-base/src/players/IterablePlayer/Mcap/McapIterableSource.test.ts
defunctzombie: Worth putting some comment about this in the code? Or will a build failure be enough to ward off curious explorers?

defunctzombie (approved):
The things we do to appease the compilers. I like that you were able to use `typeof` to figure out the types. Seems simpler than the before version too.

defunctzombie (approved):

jtbandes (approved):

PR #7183, opened 2023-11-28
Flaky plots IV: A New Hope

cfoust:
**User-Facing Changes**
N/A

**Description**
I'm continuing my quest to fix our flaky stories. In this attempt I:
1. Changed all Plot and State Transition stories to just use single light mode plots (where applicable), since it seems that some of our problems stem from timing discrepancies between the two. In general it's nice to see both themes but in practice this seems to still cause us problems. This could just be temporary.
2. Fixed an ordering issue in the `Chart` component. It turns out that the previous `queueUpdate` logic was incorrect for specifically the first `update` call: if updates were queued while we were still initializing, we would incorrectly push the first `update` on top of them. Instead, we insert the first update at the _beginning_ of the queued updates.

### DIFF
--- a/packages/studio-base/src/components/Chart/index.tsx
+++ b/packages/studio-base/src/components/Chart/index.tsx
@@ -241,8 +241,8 @@ function Chart(props: Props): JSX.Element {
     return out;
   }, [data, typedData, height, options, isBoundsReset, width]);
 
-  // Flush all new to the worker, coalescing them together if there is more
-  // than one.
+  // Flush all new updates to the worker, coalescing them together if there is
+  // more than one.
   const flushUpdates = useCallback(
     async (send: RpcSend | undefined) => {
       if (send == undefined || isSending.current) {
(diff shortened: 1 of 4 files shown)


defunctzombie (approved):
Seems strange that we have updates prior to initialization being complete? Wouldn't those be dropped (or cause errors) since initialization has not set everything up yet?

Do we have some stories left that do test plots in both color schemes to make sure we get good visual coverage so we don't break one of the color modes with respect to axis colors or other theme changes?

cfoust:
Even before my changes this code had a quasi-mechanism for queuing rendering jobs that occur before the ChartJS instance is started; there's nothing new there. Updates to the component happen independently of the initialization process, naturally, so there needs to be some kind of queue. The bug occurred because we were pushing an old update (the first one, which kicked off initialization) onto the top of the queue.

I don't recall us having any stories with that explicitly in mind, but there's also a lot of stories and I'm not sure what the initial intent was for all of them

PR #7184, opened 2023-11-28
Publish panel: Conditinally add common ROS datatypes

achim-k:
**User-Facing Changes**
Not worth calling out

**Description**
Adds well-known ROS datatypes to the publish panel if the datasource profile is ROS1 or ROS2. Allows these types to be selected from the message type dropdown in case the player does not already provide these datatypes.

### DIFF
--- a/packages/studio-base/src/panels/Publish/index.tsx
+++ b/packages/studio-base/src/panels/Publish/index.tsx
@@ -16,7 +16,13 @@ import { useEffect, useMemo } from "react";
 import { makeStyles } from "tss-react/mui";
 import { useDebounce } from "use-debounce";
 
+import { MessageDefinition } from "@foxglove/message-definition";
+import CommonRosTypes from "@foxglove/rosmsg-msgs-common";
 import { useDataSourceInfo } from "@foxglove/studio-base/PanelAPI";
+import {
+  MessagePipelineContext,
+  useMessagePipeline,
+} from "@foxglove/studio-base/components/MessagePipeline";
 import Panel from "@foxglove/studio-base/components/Panel";
 import PanelToolbar from "@foxglove/studio-base/components/PanelToolbar";
 import Stack from "@foxglove/studio-base/components/Stack";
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/panels/Publish/index.tsx
```diff
@@ -93,11 +100,35 @@ function parseInput(value: string): { error?: string; parsedObject?: unknown } {
...
+    if (commonTypes == undefined) {
+      return datasourceDatatypes;
+    }
+
+    const additionalTypes: RosDatatypes = new Map();
+    for (const dataType in commonTypes) {
+      additionalTypes.set(dataType, commonTypes[dataType]!);
+    }
+
+    return new Map([...additionalTypes, ...datasourceDatatypes]);
```
defunctzombie, line 130:
```suggestion
    // datasourceDatatypes is added after commonTypes to take precedence (override) any commonTypes of the same name
    return new Map([...Object.entries(commonTypes), ...datasourceDatatypes]);
```

packages/studio-base/src/panels/Publish/index.tsx
```diff
@@ -93,11 +100,35 @@ function parseInput(value: string): { error?: string; parsedObject?: unknown } {
...
+    const commonTypes: Record<string, MessageDefinition> | undefined = {
+      ros1: CommonRosTypes.ros1,
+      ros2: CommonRosTypes.ros2galactic,
+    }[datasourceProfile ?? ""];
+
+    if (commonTypes == undefined) {
+      return datasourceDatatypes;
+    }
+
+    const additionalTypes: RosDatatypes = new Map();
```
defunctzombie, line 125: You can delete this with the change I proposed below.

defunctzombie (approved):

packages/studio-base/src/panels/Publish/index.tsx
```diff
@@ -93,11 +100,35 @@ function parseInput(value: string): { error?: string; parsedObject?: unknown } {
...
+    for (const dataType in commonTypes) {
+      additionalTypes.set(dataType, commonTypes[dataType]!);
+    }
```
achim-k, line 128:
```suggestion
```

PR #7189, opened 2023-11-29
[Image Panel] Fix unmatched frameId error when switching to none from non-existing topic

snosenzo:
## User-facing changes

 - [Image Panel] Fix unmatched frameId error when switching to `None` from non-existing topic

## Description

When loading a config where the image topic is defined but the calibration topic is not, the image is read and decoded, but a fallback camera model is not created onDecode because the calibration topic is not `None`. Then when the user switches to `None`, a new fallback is not created because it does not

This issue existed because we were only creating fallback calibrations on image decode. If an image had been received when there was a non-existing calibration topic, a fallback is not created. When switching to `None` as a calibration topic we were getting an error because the fallback calibration had not been created. This fixes the issue by creating a fallback calibration when switching to `None` if an image renderable already exists.

Fixes: FG-5796

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageMode.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageMode.ts
@@ -505,59 +505,73 @@ export class ImageMode
     const path = action.payload.path;
     const category = path[0]!;
     const value = action.payload.value;
-    if (category === "imageMode") {
-      const prevImageModeConfig = this.getImageModeSettings();
-      this.saveSetting(path, value);
-      const config = this.getImageModeSettings();
-      const calibrationTopicChanged =
-        config.calibrationTopic !== prevImageModeConfig.calibrationTopic;
-      if (calibrationTopicChanged) {
-        const changingToUnselectedCalibration = config.calibrationTopic == undefined;
-        if (changingToUnselectedCalibration) {
-          this.renderer.enableImageOnlySubscriptionMode();
-        }
 
-        const changingFromUnselectedCalibration = prevImageModeConfig.calibrationTopic == undefined;
-        if (changingFromUnselectedCalibration) {
-          this.renderer.disableImageOnlySubscriptionMode();
-        }
-      }
-      const imageTopicChanged = config.imageTopic !== prevImageModeConfig.imageTopic;
-      if (imageTopicChanged && config.imageTopic != undefined) {
-        const imageTopic = this.renderer.topics?.find((topic) => topic.name === config.imageTopic);
-        if (imageTopic) {
-          this.setImageTopic(imageTopic);
-        }
-      }
-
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageMode.ts
```diff
@@ -505,59 +505,73 @@ export class ImageMode
...
+        this.renderer.enableImageOnlySubscriptionMode();
+        const decodedImage = this.imageRenderable?.getDecodedImage();
+        const lastImageMessage = this.imageRenderable?.userData.image;
+        // if we've already received an image, use it to create a fallback camera model
+        // otherwise we would need to wait for the next image
+        if (decodedImage && lastImageMessage) {
+          this.#updateFallbackCameraModel(
+            { width: decodedImage.width, height: decodedImage.height },
+            getFrameIdFromImage(lastImageMessage),
+          );
```
snosenzo, line 532: real change here.
the rest is taking everything out of the if

jtbandes (commented):
This seems ok, but I think it's kind of weird to do `#updateFallbackCameraModel` inside of `handleSettingsAction` – outside the scope of responsibility of a function that is supposed to be doing settings. 

Is there a way this could be folded into `updateViewAndRenderables` so it happens "automatically" after the settings change?

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageMode.ts
```diff
@@ -645,23 +652,35 @@ export class ImageMode
...
-      focalLength: DEFAULT_FOCAL_LENGTH,
-    });
-    this.#updateCameraModel(cameraInfo);
-    this.#updateViewAndRenderables();
-  };
+  /** Creates a fallback camera model based off of the renderable with a decoded image and updates the camera.
+   * Will no-op if there is not a decodedImage on the renderable.
+   * Be sure to call `#updateViewAndRenderables` after calling this method to update the camera and renderable.
+   */
+  #updateFallbackCameraModelFor(renderable: ImageRenderable) {
```
jtbandes, line 667:
```suggestion
  #updateFallbackCameraModel(renderable: ImageRenderable) {
```

the `For` suffix is weird – reads like Obj-C or Swift 🤣 

```swift
func updateFallbackCameraModel(for renderable: ImageRenderable) {
```

jtbandes (approved):

packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageMode.ts
snosenzo: ah yeah, I should just use the old method name. I had them both originally, but now they're consolidated it probably makes sense not to name it this way

PR #7192, opened 2023-11-30
Add defer function to foxglove/den/disposable

defunctzombie:
**User-facing changes**
None

**Description**

Add a `defer` function which behaves like the `defer` function from golang and defers the execution of the passed function until the scope exists. This makes use of the new `Disposables` spec. See comment in code for more details and usage.

This function is useful if you want to ensure some logic runs regardless of how the execution of the scope ends. Previously you would accomplish this via a `try/finally` which would add indentation levels and move the logic you wanted to defer away from the place where you initialized the resource you want to defer.


snosenzo (approved):

defunctzombie:
@jtbandes you expressed some sadness about this living in monads. I am happy to move it to another place within `den` - let me know where you would prefer it lives.

From wikipedia https://en.wikipedia.org/wiki/Monad_(functional_programming):

image

monads seemed reasonable based on my reading

jtbandes:
`den/disposable`?

defunctzombie:
`den/disposable` it is 🗑️

PR #7196, opened 2023-11-30
Avoid opening plot settings which deleting a plot series via the legend

defunctzombie:
**User-Facing Changes**
When clicking the "x" on a plot legend, the series is deleted without the plot settings opening.

**Description**
When searching a list of topics and using the drag & drop feature with the plot panel it is convenient to quickly add a signal and then remove it. When removing the signal using the legend "x" button, the panel settings would open and lose the topic search sidebar context.

This change updates the "x" button behavior on the legend to stop the click event from triggering the settings to open and leaves the sidebar state unchanged.


amacneil (approved):
👍 from me

jtbandes (approved):

2metres (approved):

PR #7199, opened 2023-12-01
Disable ligatures in default datagrid styles

bryfox:
**User-Facing Changes**

None.

**Description**

This disables ligatures for type inside a data grid, which is intended for tabular data. We don't display any data grids in Studio today.

FG-5926


2metres:
Closing in favor of [adding changes to global theme typography](https://github.com/foxglove/studio/pull/7219)

PR #7205, opened 2023-12-05
Restore state transition panel downsampling

cfoust:
**User-Facing Changes**
Marginally speed up the rendering performance of the state transition panel.

**Description**
This had broken at some point, certainly when I introduced `useProvider`. Honestly though, downsampling doesn't do very much for the state transition panel right now because distinct states are always represented by at least one point; the maximum number of points is not respected for some nuanced reasons. We will address this soon.

In the mean time though, this PR helps because when we zoom in, we will (correctly) cull points outside of the viewport. We are not doing that right now.

### DIFF
--- a/packages/studio-base/src/components/TimeBasedChart/Downsampler.ts
+++ b/packages/studio-base/src/components/TimeBasedChart/Downsampler.ts
@@ -5,7 +5,7 @@
 import { iterateObjects } from "@foxglove/studio-base/components/Chart/datasets";
 import { RpcScales } from "@foxglove/studio-base/components/Chart/types";
 
-import { downsample } from "./downsample";
+import { downsample, MAX_POINTS } from "./downsample";
 import { ChartDatasets, PlotViewport } from "./types";
 
 type UpdateParams = {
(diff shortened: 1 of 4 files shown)


packages/studio-base/src/components/TimeBasedChart/Downsampler.ts
```diff
@@ -66,19 +61,13 @@ export class Downsampler {
...
     }
 
-    const { bounds: dataBounds } = this.#datasetBounds;
-    const view: PlotViewport = {
-      width: 0,
-      height: 0,
-      bounds: dataBounds,
-    };
-
+    const numPoints = MAX_POINTS / Math.max(this.#datasets.length, 1);
```
defunctzombie, line 64: Amusing way to avoid divide by 0 :D

packages/studio-base/src/components/TimeBasedChart/useDownsampler.tsx
```diff
@@ -85,7 +85,7 @@ export default function useDownsampler(datasets: ChartDataset[]): {
         // setSetter cannot take two arguments, so we can't just write
         // `register: setSetter`
         register: (newSetter) => {
-          setSetter(newSetter);
+          setSetter(() => newSetter);
```
defunctzombie, line 88: What happened here?

defunctzombie (approved):
What is the actual fix? The PR description doesn't call that out - looks to me like it is either the setter or the missing `getDatasetBounds` call?

packages/studio-base/src/components/TimeBasedChart/useDownsampler.tsx
jtbandes: Classic footgun. Just ask your past self: https://github.com/foxglove/studio/pull/3964

PR #7208, opened 2023-12-06
Relax width on nested menus to prevent clipping

2metres:
**User-Facing Changes**
Bug fix

**Description**
[Bug reported in storybook](https://www.chromatic.com/test?appId=603ec8bf7908b500231841e2&id=6564e09e2f132656d829b88a)

Screen Shot 2023-12-06 at 1 00 37 pm

Resolves FG-5917

### DIFF
--- a/packages/studio-base/src/components/AppBar/NestedMenuItem.tsx
+++ b/packages/studio-base/src/components/AppBar/NestedMenuItem.tsx
@@ -35,7 +35,7 @@ const useStyles = makeStyles<void, "endIcon">()((theme, _params, classes) => ({
   },
   menuList: {
     minWidth: 180,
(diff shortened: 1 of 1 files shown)


jtbandes (commented):
This seems like a pretty fragile solution – is there a way to either enable truncation on these or remove the max width entirely? Do you know why we have it in the first place?

2metres:
Mainly because the FIle menu shows recent file names which can get quite long which are [middle truncated](https://www.chromatic.com/test?appId=603ec8bf7908b500231841e2&id=656fd6a59a198426c9ceae7f) But we dont' apply it to other items in the menu.

jtbandes (approved):

PR #7212, opened 2023-12-06
Better downsampling for the State Transition panel

cfoust:
**User-Facing Changes**
Introduce a new downsampling algorithm for the State Transition panel to greatly improve its performance.

**Description**
This PR introduces a downsampling algorithm specifically for state transition data called `downsampleStates`. It is conceptually similar to our existing `downsampleTimeseries` algorithm, which divides the viewport into a set of equally-sized intervals.

The nature of state transitions lets us do something a bit more sophisticated: when an interval contains more points than are in our "budget" (put differently: it contains more than one distinct state transition), we render it in gray. We can work on the UX of this, but it does work, includes a suite of tests, and resolves the high-level issue with the panel's rendering performance.

Here is an example from nuScenes. This scenario intentionally contains a lot of state transitions, more than might actually occur in everyday use; it's just a good way of demonstrating how this might work.

https://github.com/foxglove/studio/assets/4588553/a9b3e86b-b673-4529-bd8d-c126a8eb34c4

When the user mouses over a segment with multiple states, we render them in the tooltip:
image

### DIFF
--- a/packages/studio-base/src/components/Chart/types.ts
+++ b/packages/studio-base/src/components/Chart/types.ts
@@ -18,6 +18,10 @@ type Datum = ScatterDataPoint & {
   value?: string | number | bigint | boolean;
   // Constant name for the datum (used by state transitions)
   constantName?: string | undefined;
+
+  // Contains all of the distinct states present in this line segment. Only for
+  // state transition data.
+  states?: string[];
 };
 export type ObjectData = (Datum | typeof ChartNull)[];
 export type ChartData = ChartJsChartData<"scatter", ObjectData>;
--- a/packages/studio-base/src/components/TimeBasedChart/downsampleStates.ts
+++ b/packages/studio-base/src/components/TimeBasedChart/downsampleStates.ts
@@ -1,0 +1,148 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { Point } from "@foxglove/studio-base/components/Chart/datasets";
+
+import type { PlotViewport } from "./types";
+
+import { MINIMUM_PIXEL_DISTANCE } from "./downsample";
+
+type StatePoint = {
+  x: number;
+  index: number | undefined;
+};
+
+type Label = {
+  index: number;
+  value: string;
+};
+
+function addLabel(label: Label, labels: Label[]): Label[] {
+  const last = labels.at(-1);
+  if (last != undefined && label.value === last.value) {
+    return labels;
+  }
+
+  return [...labels, label];
+}
+
+type Interval = { x: number; xPixel: number; labels: Label[]; index: number };
+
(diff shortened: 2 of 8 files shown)


amacneil:
- performance is great
- change label name to `[...]`
- change tooltip: if possible, add a list entry for every available state under the cursor
- slightly darker grey

cfoust:
This is complete and includes all of the fixes for issues identified in product review. It also shows all coalesced states in the tooltip when applicable.

packages/studio-base/src/components/Chart/types.ts
```diff
@@ -18,6 +18,9 @@ type Datum = ScatterDataPoint & {
   value?: string | number | bigint | boolean;
   // Constant name for the datum (used by state transitions)
   constantName?: string | undefined;
+
+  // Used only for showing multiple states in state transition
```
defunctzombie, line 22:
```suggestion
  // The states in a state transition
```

I tend to prefer docs that say what the thing is or what it should contain because the use might evolve over time.

packages/studio-base/src/components/TimeBasedChart/downsampleStates.ts
```diff
@@ -0,0 +1,209 @@
...
+  index: number;
+  // The value of the label
+  value: string;
+};
+
+/**
+ * Add a Label to a list of Labels, but not if the top of the new Label's value
+ * matches the value of the Label at the top of the stack.
+ */
+function addLabel(label: Label, labels: Label[]): Label[] {
```
defunctzombie, line 34:
```suggestion
function addLabel(label: Label, labels: Immutable<Label[]>): Label[] {
```

packages/studio-base/src/components/TimeBasedChart/downsampleStates.ts
```diff
@@ -0,0 +1,209 @@
...
+  // All of the labels that appeared in this interval
+  labels: Label[];
+  // The index of the point that started the interval
+  index: number;
+};
+
+/**
+ * Downsample state transition data by breaking the visible bounds into a set
+ * of evenly-sized intervals and making a note of the number of state
+ * transitions (between distinc states) that occur in each.
```
defunctzombie, line 60:
```suggestion
 * transitions (between distinct states) that occur in each.
```

packages/studio-base/src/components/TimeBasedChart/downsampleStates.ts
```diff
@@ -0,0 +1,209 @@
...
+  // connected to their peers and available for pan/zoom.
+  // Points outside this buffer window are dropped.
+  const xRange = bounds.x.max - bounds.x.min;
+  const minX = bounds.x.min - xRange * 0.5;
+  const maxX = bounds.x.max + xRange * 0.5;
+
+  let firstPastBounds: number | undefined = undefined;
+
+  /**
+   * Conclude the current interval, producing StatePoint[] as necessary.
```
defunctzombie, line 99: What does "necessary" mean in this context?

defunctzombie (approved):
Good stuff

PR #7216, opened 2023-12-07
Prevent flickering or clearing of image in image panel when switching between selected and unselected calibration topics

snosenzo:
**User-Facing Changes**

 - Prevent flickering or clearing of image in image panel when switching between selected and unselected calibration topics

**Description**

We don't need to clear the `ImageMode` scene extension when switching between Image 3D and Only modes. 

Depends on #7171 because it can get into a bad state when switching from None to selected if another panel is subscribed to the same calibration topic. That PR will properly show a waiting for calibration empty state instead of warped camera settings


defunctzombie (commented):
Maybe I am late to the review party here but as someone who only knows about the "image mode" concept at a high level, the "clearImageMode" argument was a bit surprising. Does it actually clear the panel's image mode? Does image mode mean something else for this context?

I also found it somewhat strange that the "Renderer" is involved in this. Is that because the two topics are separate renderables?

snosenzo:
@defunctzombie the renderer is responsible for calling `removeAllRenderables` on all scene extensions when clearing. `ImageMode` in this context is merely the name of the scene extension that handles the main settings and renderable for the Image panel. So we are clearing the ImageMode extension here, it seems like that's a little unclear so I can rename the variable `clearImageModeExtension`

jtbandes (approved):

PR #7219, opened 2023-12-08
Disable contextual alternatives on theme typography

2metres:
**User-Facing Changes**
None

**Description**
Resolves FG-5926


2metres:
Context on [what is going on here from Inter repo](https://github.com/rsms/inter/issues/222)

PR #7221, opened 2023-12-11
add support for ws protocol's status message supplying tips

superyuuki:
**User-Facing Changes**
Adds support for an optional 'tip' feature that websocket servers can send alongside status broadcasts

**Description**
I use the status type message from foxglove ws-protocol to send server status information. sometimes I want to send tips when the server errors, but the client doesn't support it even if the Problem type fully does. I have 0 clue how typescript's optional?s work but hopefully this should make it so that if the event supplies a tips field in json, that field is given to the problem, so that it renders.

### DIFF
--- a/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
+++ b/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
@@ -389,9 +389,11 @@ export default class FoxgloveWebSocketPlayer implements Player {
         log.error(msg);
       }
 
+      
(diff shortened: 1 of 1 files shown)


superyuuki:
use case: 


I'm sending tips alongside my status messages in my java implementation of ws-protocol
image
image

sadly i don't get to send any tips alongside :(
image

superyuuki:
okay i may need to fix this

superyuuki:
oh StatusMessage is part of the protocol :D

PR #7224, opened 2023-12-12
Unbolden button text and headings

2metres:
**User-Facing Changes**
None

**Description**
Follow on to feedback from @amacneil that the headings and button text look too bold since merging with the console theme.

So reducing the boldness on both slightly


jtbandes:
Sure – I can buy that buttons should not be bold. However, I am concerned if the default "normal" font weight is 400 (is that right?), then having both 400 and 500 is such a subtle difference it is not worth it. Maybe just set buttons to 400?

2metres:
We assign 400 | 500 | 600 to specific variants of Inter

https://github.com/foxglove/studio/blob/9875db161bf5818b47b76edfa0d8b6be0d865501/packages/studio-base/src/styles/assets/inter.css#L29-L56

So 500 means using Medium and 600 means using SemiBold

defunctzombie (commented):
What's the justification for this change? Is that called out anywhere and I missed it? Something opinion based like stating "I visually thing this looks nicer" would be helpful to add context on how to review this.

Weird that chromatic is showing this as a new story for typography but we already had a story.

Your comment about how 500 and 600 are treated could be worth adding to the code so we remember this behavior.

What is the visual difference for this change? No stories are changed.

Will we need to make a similar change for app theme?

jtbandes:
The visual changes can be seen here: https://www.chromatic.com/review?appId=603ec8bf7908b500231841e2&number=7224&type=linked&view=changes

I still find it questionable that we would be using both 400 and 500 regularly in the app – seems like we should focus on fewer more distinct weights rather than having two weights which are so similar to each other.

> We specifically assign 400 | 500 | 600 to specific variants of Inter so they are less subtle.

I don't really know what this means but I don't think that the CSS file you linked is changing the appearance of the font weights. It is just saying which weight corresponds to which font file. This mapping is normal and not something we hand-picked. This is what they look like on Google Fonts, confirming the mapping of 400->regular, 500->medium, 600->semibold:

image

And this from MDN docs on font-weight:

image

My concern is just that fonts which look very similar, but slightly different, will contribute to the app looking a bit messy. Like Roman I would be interested in seeing more motivation/background for what we are trying to achieve with these changes since just looking at the stories doesn't tell the whole story.

2metres:
My overall motivation was to make the headings less bold as they feel quite heavy and reduce the boldness of the button text as it feels a bit too heavy.

I personally prefer this change but open to discussing further.

PR #7226, opened 2023-12-12
Don't keep current messages in the plot worker if block data already has them

cfoust:
**User-Facing Changes**
Slight improvements to plot memory usage in certain circumstances.

**Description**
We used to do this but it got removed when we migrated to a worker. It does not seem to make a big difference, though.

Still needs a test, probably.


cfoust:
Closing in favor of https://github.com/foxglove/studio/pull/7235

PR #7227, opened 2023-12-13
Add experimental memory saver option

achim-k:
**User-Facing Changes**
Add experimental memory saver app setting

**Description**
Adds a new `Memory saver`  app setting (experimental section) which, when enabled, reduces memory consumption of the buffered iterable source by
- Evicting old blocks that are before the current read head, even if the cache is not yet full
- Limiting the max. buffered size to 50% of the default limit of 1GB

We may use some other memory saving strategies later, but for now I only started with these two

### DIFF
--- a/packages/studio-base/src/i18n/en/appSettings.ts
+++ b/packages/studio-base/src/i18n/en/appSettings.ts
@@ -20,6 +20,8 @@ export const appSettings = {
   layoutDebugging: "Layout debugging",
   layoutDebuggingDescription: "Show extra controls for developing and debugging layout storage.",
   light: "Light",
+  memorySaver: "Memory saver",
+  memorySaverDescription: "Reduces memory consumption",
   messageRate: "Message rate",
   newAppMenu: "Enable unified navigation",
   newAppMenuDescription: "Show the new menu and navigation.",
--- a/packages/studio-base/src/AppSetting.ts
+++ b/packages/studio-base/src/AppSetting.ts
@@ -20,6 +20,7 @@ export enum AppSetting {
 
   // Experimental features
   SHOW_DEBUG_PANELS = "showDebugPanels",
+  MEMORY_SAVER_ENABLED = "memorySaverEnabled",
 
   // Miscellaneous
   HIDE_SIGN_IN_PROMPT = "hideSignInPrompt",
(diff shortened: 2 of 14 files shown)


packages/studio-base/src/i18n/en/appSettings.ts
```diff
@@ -20,6 +20,8 @@ export const appSettings = {
   layoutDebugging: "Layout debugging",
   layoutDebuggingDescription: "Show extra controls for developing and debugging layout storage.",
   light: "Light",
+  memorySaver: "Memory saver",
+  memorySaverDescription: "Reduces memory consumption",
```
achim-k, line 24: Needs a better description :sweat_smile:

defunctzombie:
> * Evicting old blocks that are before the current read head, even if the cache is not yet full

The thing we lose with this behavior is the ability to seek back to something you played through recently and want to re-visit. I don't have metrics to back this up but my intuition is that more often after seeing something interesting you might seek right before it and play it again. This is less important for plot workflows where you can see all the data but more important for 3d panel, raw messages, etc where you only see current frame data. By evicting this we will make it "expensive" to seek back even a little bit by having to re-fetch the data from the data source.

defunctzombie:
> * Limiting the max. buffered size to 50% of the default limit of 1GB

I'm inclined to say this is the setting we should go with - It does indicate to me that it should be a setting where you can configure the available memory in the buffer.

More broadly - I am trying to understand how I as a user would know to tick this and what impact it will have on my workflow. Right now as presented I don't think that is clear.

packages/studio-base/src/i18n/en/appSettings.ts
jtbandes: The description should definitely indicate reasons that you wouldn't want this always on – otherwise it will seem like it should be always on

snosenzo: label: "Limit memory usage"
description: "Lower in-app memory limit to reduce out-of-memory crashes and tab freezes"
?

jtbandes: Still needs to explain what the downside is

snosenzo: "Lower in-app memory limit to reduce out-of-memory crashes and tab freezes. This will reduce the amount of data that can be preloaded and buffered at a given time."

defunctzombie:
I'm not feeling this setting as presented. There's too much "assumption" in turning this on and I noted above some downsides to seeking behavior. It sounds like we have some other promising PRs around memory estimates and use and I'd like to explore those before coming back to this setting.

achim-k:
Closing this due to above raised concerns

PR #7229, opened 2023-12-13
Fix a handful of plot issues

cfoust:
**User-Facing Changes**
Fix an issue where plotting a signal with the header stamp would result in misleading output.
Revert a "fix" that was intended to handle missing block data more safely.
Fix an issue where downsampled plot data points could appear out of order.

**Description**

This PR fixes ~three~ four separate issues:
1. #7193 caused a regression for topic aliases. I'm just going to revert it for now.
2. We were incorrectly showing the dots for signals that use header stamps. I centralized the logic that downsamples entire signals at once and ensured they correctly detect when we actually can just render the raw points.
3. (Unrelated to the customer issues) Our downsampler occasionally reversed the order of points inside of a single interval. We now sort the points inside of an interval.
4. It turns out that `showLine` can also appear on `TypedDataset`s and not just on the path, which is news to me. This caused an issue where we were hiding points (with `pointRadius=0`) that were not connected by a line.

### DIFF
--- a/packages/studio-base/src/components/TimeBasedChart/downsample.test.ts
+++ b/packages/studio-base/src/components/TimeBasedChart/downsample.test.ts
@@ -100,7 +100,7 @@ describe("downsampleTimeseries", () => {
       ]),
       bounds,
     );
-    expect(result).toEqual([0, 2, 1, 3]);
+    expect(result).toEqual([0, 1, 2, 3]);
   });
 
   it("should keep entry/exit datum to an interval", () => {
--- a/packages/studio-base/src/components/TimeBasedChart/downsample.test.ts
+++ b/packages/studio-base/src/components/TimeBasedChart/downsample.test.ts
@@ -119,6 +119,21 @@ describe("downsampleTimeseries", () => {
     );
     expect(result).toEqual([0, 1, 2, 3, 4]);
   });
+
+  it("should not produce out-of-order points", () => {
+    const result = downsampleTimeseries(
+      iterateObjects([
+        { x: 0, y: 15, value: 0 },
+        { x: 1, y: 100, value: 0 },
+        { x: 2, y: 10, value: 0 },
+        { x: 3, y: 15, value: 0 },
+      ]),
+      bounds,
+      6, // two intervals
+    );
+    expect(result).toEqual([...result].sort((a, b) => a - b));
+    expect(result).toEqual([0, 1, 2]);
+  });
 });
 
 describe("downsampleScatter", () => {
--- a/packages/studio-base/src/components/TimeBasedChart/downsample.ts
+++ b/packages/studio-base/src/components/TimeBasedChart/downsample.ts
@@ -129,22 +129,25 @@ export function continueDownsample(
     // create a new interval when encountering a new label to preserve the transition from one label to another
     if (intFirst?.xPixel !== x || (intLast?.label != undefined && intLast.label !== datum.label)) {
       // add the min value from previous interval if it doesn't match the first or last of that interval
+      const newPoints: number[] = [];
       if (intMin && intMin.yPixel !== intFirst?.yPixel && intMin.yPixel !== intLast?.yPixel) {
-        indices.push(intMin.index);
+        newPoints.push(intMin.index);
       }
 
       // add the max value from previous interval if it doesn't match the first or last of that interval
       if (intMax && intMax.yPixel !== intFirst?.yPixel && intMax.yPixel !== intLast?.yPixel) {
-        indices.push(intMax.index);
+        newPoints.push(intMax.index);
       }
 
       // add the last value if it doesn't match the first
       if (intLast && intFirst?.yPixel !== intLast.yPixel) {
-        indices.push(intLast.index);
+        newPoints.push(intLast.index);
       }
 
       // always add the first datum of an new interval
-      indices.push(index);
+      newPoints.push(index);
+
+      indices.push(...newPoints.sort((a, b) => a - b));
 
       intFirst = { xPixel: x, yPixel: y, index, label };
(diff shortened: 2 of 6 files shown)


packages/studio-base/src/components/TimeBasedChart/downsample.ts
```diff
@@ -129,22 +129,25 @@ export function continueDownsample(
...
       if (intLast && intFirst?.yPixel !== intLast.yPixel) {
-        indices.push(intLast.index);
+        newPoints.push(intLast.index);
       }
 
       // always add the first datum of an new interval
-      indices.push(index);
+      newPoints.push(index);
+
+      indices.push(...newPoints.sort((a, b) => a - b));
```
james-rms, line 150:
```suggestion
      indices.push(...newPoints.sort();
```
sad we can't do an in-place sort on a range of an array, but I can't figure out an obvious way without using a JS sort implementation, which might be slower.

cfoust: at any rate, N is strictly <= 4 here so it could be worse?

james-rms: if N is strictly <= 4, it seems likely that an in-place simple (insertion?) sort on the end of `indices` would be faster than creating a new array, sorting it, then pushing it onto `indices`.

packages/studio-base/src/panels/Plot/blocks.ts
```diff
@@ -122,7 +76,38 @@ export function processBlocks(
...
+    R.map((v: SubscribePayload): Update => {
+      const { topic } = v;
+      const currentCursor = cursors[topic] ?? 0;
+
+      // 1. check for any new, non-empty unsent blocks
+      const lastNew = R.findLastIndex((block) => {
+        const data = block[topic];
+        return data != undefined && data.length !== 0;
+      }, blocks.slice(currentCursor));
+      const newCursor = lastNew === -1 ? currentCursor : lastNew + currentCursor + 1;
```
james-rms, line 88: why the ` + 1` here?

james-rms (approved):

cfoust:
Added a test that fails on `main`:
image

I also find it tragicomic that we straight up had a test with out-of-order points:
image

packages/studio-base/src/panels/Plot/processor/downsample.ts
```diff
@@ -97,15 +97,25 @@ const ZOOM_RESET_FACTOR = 0.2;
...
+
+  const indices = downsampleTimeseries(iterateTyped(data), view, maxPoints);
   const resolved = resolveTypedIndices(data, indices);
   if (resolved == undefined) {
     return undefined;
   }
 
-  return resolved;
+  return {
+    pointRadius: 0,
```
defunctzombie, line 116: Good to add a comment here that the UX we want is that when downsampling we do not show the datapoints

packages/studio-base/src/components/TimeBasedChart/downsample.ts
defunctzombie: I agree with James here - this looks somewhat "wasteful" to the reader without an additional comment - either way a comment would be useful here on what behavior matters

packages/studio-base/src/components/TimeBasedChart/downsample.ts
```diff
@@ -96,7 +96,7 @@ export function finishDownsample(state: DownsampleState): number[] {
     indices.push(intLast.index);
   }
 
-  return indices;
+  return indices.sort((a, b) => a - b);
```
defunctzombie, line 99: comment on the _why_ for future maintainers

packages/studio-base/src/components/TimeBasedChart/downsample.test.ts
```diff
@@ -119,6 +119,21 @@ describe("downsampleTimeseries", () => {
     );
     expect(result).toEqual([0, 1, 2, 3, 4]);
   });
+
+  it("should not produce out-of-order points", () => {
```
defunctzombie, line 123: Arguably this test is no longer needed since the other test is updated (in a later commit) with the correct index order? The testing gap was that the other test was wrong till now?

packages/studio-base/src/panels/Plot/processor/downsample.ts
```diff
@@ -417,9 +417,12 @@ function updatePartialView(path: PlotPath, params: PathParameters, state: PathSt
   };
 
   const mergedData = mergeTyped(blockData?.data ?? [], currentData?.data ?? []);
+  const data = applyTransforms(sliceBounds(mergedData, partialBounds), path);
+
+  // Scatter plots also need special treatment here
```
defunctzombie, line 422: What does "special treatment" mean?

defunctzombie:
> 4\. It turns out that `showLine` can also appear on `TypedDataset`s and not just on the path, which is news to me. This caused an issue where we were hiding points (with `pointRadius=0`) that were not connected by a line.

How do we test these last two commits you added?

packages/studio-base/src/components/TimeBasedChart/downsample.ts
james-rms: this could be clean actually if every `push` onto `indices` was a done with a `pushAndBubbleDown` function. you'd be doing an online insertion sort into `indices`.

packages/studio-base/src/components/TimeBasedChart/downsample.test.ts
cfoust: You're right that theoretically the other tests implicitly cover this property, but I wanted to add a test that explicitly refers to it (and includes the logic that checks the result against the sorted result).

defunctzombie (approved):

PR #7230, opened 2023-12-13
New method for estimating message memory size

achim-k:
**User-Facing Changes**
Improve memory estimation to make OOM crashes less likely

**Description**
For estimating the size of a deserialized message, we currently use a schema-based approach where the message schema definition is analyzed to get an idea on what fields are to be expected in the message and how big the message might be. This however falls short on dynamic messages with arrays or strings as we cannot know the array or string size upfront. For these messages, we conservatively assumed an empty array size and a fixed string size, which in some cases leads to a large underestimation of the actual message size.

This PR introduces a new estimation method that estimates the memory size of a deserialized message by iterating over its properties. Unlike the schema-based estimation, this new method iterates on the properties of a deserialized message and can thus more accurately estimate the size of dynamic messages. The size estimates are cached for each topic as it would be quite expensive to do estimate the size for every message. 
Caching also has its disadvantages however, namely that a topic's first message size estimate will be used for subsequent messages on that topic which might not reflect reality (e.g. consider two publishers publishing pointclouds of various sizes to the same topic). 

Experimental results have shown that the proposed new size estimation method produces more accurate size estimates than the current schema-based estimation method, especially when the message schema contains dynamic types (arrays or strings).

### DIFF
--- a/packages/studio-base/src/players/IterablePlayer/Mcap/McapIndexedIterableSource.ts
+++ b/packages/studio-base/src/players/IterablePlayer/Mcap/McapIndexedIterableSource.ts
@@ -35,6 +35,7 @@ export class McapIndexedIterableSource implements IIterableSource {
   #start?: Time;
   #end?: Time;
   #messageSizeEstimates: Record<string, number> = {};
+  #slicedMessageSizeEstimate: Record<string, number> = {};
 
   public constructor(reader: McapIndexedReader) {
     this.#reader = reader;
@@ -140,10 +141,8 @@ export class McapIndexedIterableSource implements IIterableSource {
     }
 
     const topicNames = Array.from(topics.keys());
-
-    // Clear the cache for existing size estimates such that the message size per topic is
-    // re-evaluated again (as subscription options might have changed).
-    this.#messageSizeEstimates = {};
+    // Clear the cache for existing size estimate of sliced topics such that these topics are re-evaluated again.
+    this.#slicedMessageSizeEstimate = {};
 
     for await (const message of this.#reader.readMessages({
       startTime: toNanoSec(start),
@@ -167,7 +166,11 @@ export class McapIndexedIterableSource implements IIterableSource {
         const msg = channelInfo.parsedChannel.deserialize(message.data) as Record<string, unknown>;
         const spec = args.topics.get(channelInfo.channel.topic);
         const payload = spec?.fields != undefined ? pickFields(msg, spec.fields) : msg;
-        const estimatedMemorySize = this.#estimateMessageSize(channelInfo.channel.topic, payload);
+        const estimatedMemorySize = this.#estimateMessageSize(
+          channelInfo.channel.topic,
+          spec?.fields == undefined ? "full" : "sliced",
+          payload,
+        );
         const sizeInBytes =
           spec?.fields == undefined
             ? Math.max(message.data.byteLength, estimatedMemorySize)
@@ -222,7 +225,7 @@ export class McapIndexedIterableSource implements IIterableSource {
           const deserializedMessage = channelInfo.parsedChannel.deserialize(message.data);
           const sizeInBytes = Math.max(
             message.data.byteLength,
-            this.#estimateMessageSize(channelInfo.channel.topic, deserializedMessage),
+            this.#estimateMessageSize(channelInfo.channel.topic, "full", deserializedMessage),
           );
           messages.push({
             topic: channelInfo.channel.topic,
@@ -243,14 +246,20 @@ export class McapIndexedIterableSource implements IIterableSource {
     return messages;
   }
 
-  #estimateMessageSize(topic: string, msg: unknown): number {
-    const cachedSize = this.#messageSizeEstimates[topic];
+  #estimateMessageSize(topic: string, type: "full" | "sliced", msg: unknown): number {
+    const cachedSize =
+      type === "full" ? this.#messageSizeEstimates[topic] : this.#slicedMessageSizeEstimate[topic];
     if (cachedSize != undefined) {
       return cachedSize;
     }
 
     const sizeEstimate = estimateObjectSize(msg);
-    this.#messageSizeEstimates[topic] = sizeEstimate;
+    if (type === "full") {
+      this.#messageSizeEstimates[topic] = sizeEstimate;
+    } else {
+      this.#slicedMessageSizeEstimate[topic] = sizeEstimate;
+    }
+
     return sizeEstimate;
   }
 }
--- a/packages/studio-base/src/players/IterablePlayer/Mcap/McapIndexedIterableSource.ts
+++ b/packages/studio-base/src/players/IterablePlayer/Mcap/McapIndexedIterableSource.ts
@@ -34,8 +34,7 @@ export class McapIndexedIterableSource implements IIterableSource {
   >();
   #start?: Time;
   #end?: Time;
-  #messageSizeEstimates: Record<string, number> = {};
-  #slicedMessageSizeEstimate: Record<string, number> = {};
+  #messageSizeEstimates: Record<string /* subscription hash */, number> = {};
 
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/players/IterablePlayer/Mcap/McapIndexedIterableSource.ts
```diff
@@ -163,20 +140,7 @@ export class McapIndexedIterableSource implements IIterableSource {
...
-      const fields = args.topics.get(channelInfo.channel.topic)?.fields;
-      if (fields != undefined) {
-        const sizeInBytes = fields.reduce(
-          (acc, field) => acc + (channelInfo.msgSizeByField[field] ?? 0),
-          OBJECT_BASE_SIZE,
-        );
-        slicedMsgSizeByChannelId[channelId] = sizeInBytes;
-      }
-    }
+    this.#messageMemoryEstimator = new MessageMemoryEstimator(); // Reset cache for now as slicing options might have changed
```
defunctzombie, line 143: nit: drop the "for now" unless there's a situation you want to describe when we would be able to change this.

Maybe elaborate on the _why_ for a cache reset when slicing options change

packages/studio-base/src/players/messageMemoryEstimation.ts
```diff
@@ -186,3 +186,69 @@ export function estimateMessageFieldSizes(
...
+    }
+    case "symbol":
+    case "undefined":
+    case "function": {
+      throw new Error(`Can't estimate size for type ${typeof obj}`);
+    }
+  }
+}
+
+export class MessageMemoryEstimator {
```
defunctzombie, line 241: nit: consider putting this into its own file since the rest of this file seems to be exported free functions. Code structure wise I like keeping "class" files separate from free function files.

less of a nit: this class needs a comment that guides us in understanding the caching properties. Maybe consider calling it `CachingMessageMemoryEstimator`. I don't feel as strongly about that but we do need good docs on the expected behaviors and any notable why/how

packages/studio-base/src/players/messageMemoryEstimation.ts
```diff
@@ -186,3 +186,69 @@ export function estimateMessageFieldSizes(
...
+      return SMALL_INTEGER_SIZE;
+    }
+    case "bigint": {
+      return HEAP_NUMBER_SIZE;
+    }
+    case "string": {
+      return obj.length;
+    }
+    case "object": {
+      if (Array.isArray(obj)) {
```
defunctzombie, line 203: Set and Map support?

packages/studio-base/src/players/messageMemoryEstimation.ts
```diff
@@ -186,3 +186,69 @@ export function estimateMessageFieldSizes(
...
+    case "function": {
+      throw new Error(`Can't estimate size for type ${typeof obj}`);
+    }
+  }
+}
+
+export class MessageMemoryEstimator {
+  #cachedSizeEstimateByTopic: Record<string, number> = {};
+
+  public estimateMsgObjectSizeInBytes(msg: unknown, topic: string): number {
```
defunctzombie, line 244: Why is topic the second argument?

defunctzombie:
This seems good directionally - I'd like to see some examples on where we previously under-estimated and now better estimate. It seemed like you had identified some of those so maybe tests can capture the better estimation.

The new class doesn't look to me like it adds all that much but if it stays I would be clear about the interface and assumptions.

snosenzo (approved):
Overall I think this looks fine, however I think the `MessageMemoryEstimator` shouldn't be dependent on `messageIterator` calls.  I think all `Source` classes could have their own `MessageSizeCache` instance variable that stores the the size of the first seen message for a given subscription (topic, fields). this cache can hold the logic for whether it needs to update the cache based off of the subscription, so the only thing the Source needs to do is `messageSizeCache.getSizeBytes(deserializedMessage, subscription)`

packages/studio-base/src/players/messageMemoryEstimation.ts
achim-k: I have included it now, even though I also would not expect a message to contain a map or set

packages/studio-base/src/players/IterablePlayer/Mcap/McapIndexedIterableSource.ts
```diff
@@ -245,4 +245,21 @@ export class McapIndexedIterableSource implements IIterableSource {
     messages.sort((a, b) => compare(a.receiveTime, b.receiveTime));
     return messages;
   }
+
+  #estimateMessageSize(topic: string, type: "full" | "sliced", msg: unknown): number {
```
snosenzo: I'd avoid `full` terminology since it feels a bit to close to preloading. I think `whole` is better here just to keep them separate

achim-k: I have removed that logic again as it didn't work for topics that were previously fully subscribed and then only partially subscribed.

I think I'll have to include the subscription payload (sliced fields) into the cache hash. The current approach falls short when a topic is subscribed multiple times with various subscription options.

packages/studio-base/src/players/IterablePlayer/Mcap/McapIndexedIterableSource.ts
```diff
@@ -34,7 +34,8 @@ export class McapIndexedIterableSource implements IIterableSource {
   >();
   #start?: Time;
   #end?: Time;
-  #messageMemoryEstimator = new MessageMemoryEstimator();
+  #messageSizeEstimates: Record<string, number> = {};
```
defunctzombie, line 37: I'd find it helpful to add a comment noting that the _string_ here is a topic. So we know this is a map of estimates by topic. Alternatively a variable name of `messageSizeEstimatesByTopic` would do the trick.

packages/studio-base/src/players/IterablePlayer/Mcap/McapIndexedIterableSource.ts
defunctzombie: When you update this - throw in a comment for the method that says something about how this method uses the cache or caches results or some such giving us a hint that it does a form of memoization and is not calculating the size. Alternatively - you could throw `cache` in the name of the method somewhere - like `#cachedMessageSizeEstimate`. Some ideas, you can pick which makes sense.

defunctzombie: Another thing that feels somewhat off about the API is that its not quite obvious what the "type" parameter is for. It seems to only control which cache is used. Maybe this parameter is going away per your other comment above?

packages/studio-base/src/players/IterablePlayer/Mcap/McapIndexedIterableSource.ts
```diff
@@ -139,11 +139,22 @@ export class McapIndexedIterableSource implements IIterableSource {
...
-    const topicNames = Array.from(topics.keys());
+    // Determine the subscription hash which is used to lookup message size estimates.
+    // This is done here to avoid doing this repeatedly when iterating over messages.
+    const topicsWithSubscriptionHash = new Map(
+      Array.from(topics, ([topic, subscribePayload]) => [
+        topic,
+        {
+          ...subscribePayload,
+          // The subscription hash is the topic name appended by + seperated message slicing fields (if any).
+          subscriptionHash: subscribePayload.fields
```
defunctzombie, line 150: Making a function like `computeSubscriptionHash` that you can pass a topic + payload to would make this easier to read.

defunctzombie (approved):
👍 

nit: I like when Map variable names tell you what the key field is semantically. Since keys are most often "strings" having this makes it easier to grok the source code.
`messageSizeEstimates` -> `messageSizeEstimateByHash`

PR #7231, opened 2023-12-13
Show transform section settings error when there is an error adding a transform

snosenzo:
User-facing changes:

 - Fix: error when adding transform will not cause 3D panel to crash

Description:

Wrapped `addTransform` methods along with `toNanoSec` in try/catch and show error on transform section.

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
@@ -109,6 +109,7 @@ const NO_FRAME_SELECTED = "NO_FRAME_SELECTED";
 const TF_OVERFLOW = "TF_OVERFLOW";
 const CYCLE_DETECTED = "CYCLE_DETECTED";
 const FOLLOW_FRAME_NOT_FOUND = "FOLLOW_FRAME_NOT_FOUND";
+const ADD_TRANSFORM_ERROR = "ADD_TRANSFORM_ERROR";
 
 // An extensionId for creating the top-level settings nodes such as "Topics" and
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -978,21 +979,37 @@ export class Renderer extends EventEmitter<RendererEvents> implements IRenderer
...
+      const t = tf.transform.translation;
+      const q = tf.transform.rotation;
 
-    this.addTransform(normalizedParentId, normalizedChildId, stamp, t, q);
+      this.addTransform(normalizedParentId, normalizedChildId, stamp, t, q);
+    } catch (err) {
+      this.settings.errors.add(
+        ["transforms"],
+        ADD_TRANSFORM_ERROR,
+        `Error adding transform for frame ${normalizedChildId}: ${err.message}`,
```
jtbandes, line 1010: so I guess this is never cleared? do you think there would be a time when it makes sense to clear it?

jtbandes (approved):

defunctzombie (approved):
Does it make sense to wrap the calls to `#addFrameTransform` in try/catch instead? Not sure if that would be more natural for the call sites or not compared to try/catch in the functions.

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
snosenzo: all errors are cleared on seek and when transform preloading is toggled, so I think it's fine.

PR #7235, opened 2023-12-15
Do not hold on to block data in the plot worker

cfoust:
**User-Facing Changes**
Drastically reduce the amount of memory required for plotting.

**Description**
In the beginning, I moved the plot pipeline to a Web Worker in order to reduce the burden that dataset creation and downsampling put on the main thread. This was a big change, so to reduce the number of variables I kept the logic for aggregating messages the same. As a result, we had two copies of the message data used for plotting: one in the main thread and another in the worker. This approach cost a lot of memory, but interacting with plots was much faster.

This PR finally closes the loop and makes it so that the main thread sends the messages that plots need on demand; the worker only holds on to plot datasets and not the messages necessary to produce them. This greatly reduces the memory usage of plots, with the downside that the relevant raw messages must be transmitted to the worker in full when plot parameters change.

This costs a fair bit of complexity, so to make this easier to test and reason about I broke out the logic for sending messages to clients into a standalone state machine that's used by `useDatasets` and added a suite of tests.

**Known issues**
~There is still a timing issue affecting plot stories that I am investigating.~ Update: fixed!


cfoust:
I noticed a discrepancy that appears most obviously in stories: parameter updates require that we clear out the aggregated plot data and in the case of blocks, resend the underlying message data. `useDatasets` does not hold on to current data, however, which means that we cannot currently resend it on changes. In real workloads this doesn't matter that much because we either have access to block data or live data is flowing in, but this still needs to be solved.

cfoust:
I fixed the issue with current data. We now retransmit it to a specific client when plot params change. Right now we transmit _all_ current data subscribed to by plots, not just what the client actually needs, so it's a little wasteful but I'm still polishing.

packages/studio-base/src/panels/Plot/useDatasets.ts
```diff
@@ -11,36 +11,37 @@ import { useDeepMemo } from "@foxglove/hooks";
...
-  setter: (topics: SubscribePayload[]) => void;
-};
 
 let worker: Worker | undefined;
 let service: Service | undefined;
 let numClients: number = 0;
-let blockState = initBlockState();
-let clients: Record<string, Client> = {};
+let datasetsState: DatasetsState = initDatasets();
+let callbacks: Record<string, (topics: SubscribePayload[]) => void> = {};
```
defunctzombie, line 44: Not a fan of all these globals - why do we have globals here?

cfoust: It's module-level state that exists outside of the React tree. I prefer not to connect state to React that doesn't need to be there (and, for example, does not on its own trigger component updates.)

cfoust:
Just to recap:
1. I realized that we also needed to retransmit current data to the worker when plot parameters changed, though fixing this is much simpler than the approach necessary for transmitting blocks.
2. I made it so that when a client is reset, we only send the current data it needs rather than all of the accumulated current messages.

packages/studio-base/src/panels/Plot/processor/messages.ts
```diff
@@ -32,106 +40,155 @@ export function receiveMetadata(
...
+    pending: [...pending, ...(unused.length > 0 ? [{ messages, updates: unused }] : [])],
+  });
+}
+
+/**
+ * Distribute new current data to all clients, or (optionally) just a single
+ * client specified by `clientId`.
+ */
+export function addCurrentData(
+  events: readonly MessageEvent[],
```
achim-k, line 144: Is this guaranteed to be at max only one message per topic? Because otherwise we might drop some messages with `R.map((messages) => messages.slice(-1), current)` below.

packages/studio-base/src/panels/Plot/processor/messages.ts
```diff
@@ -32,106 +40,155 @@ export function receiveMetadata(
...
-    mapClients((client): [Client, SideEffects] => {
-      const { current, metadata, globalVariables } = newState;
-      const { id, params } = client;
-      if (params == undefined) {
-        return noEffects(client);
-      }
-
-      if (isSingleMessage(params)) {
-        const plotData = buildPlot(
+        current: accumulate(
```
achim-k, line 175: If all of these new current messages are already preloaded, do we still have to recompute the plot data and send it to the client? I would assume that the plot data wouldn't/shouldn't change in that case?

packages/studio-base/src/panels/Plot/useDatasets.ts
```diff
@@ -54,134 +56,57 @@ async function waitService(): Promise<Service> {
...
-  )(clientList);
-  clientList[0]?.setter(subscriptions);
-  blockState = refreshBlockTopics(subscriptions, blockState);
+  // Only send the current events that the client can actually use. This also
+  // saves us from having to `structuredClone` unused data
+  const topics = new Set(R.uniq(getClientPayloads(client).map(({ topic }) => topic)));
+  void service?.addCurrentData(
+    current.filter(({ topic }) => topics.has(topic)),
+    id,
+  );
```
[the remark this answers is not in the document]
cfoust: That's a bit tricky and even potentially undesirable--as a user, I would not expect that to happen. For example, users seem to rely on the fact that data stays around (even if it's not visible) for the CSV export functionality

packages/studio-base/src/panels/Plot/processor/messages.ts
cfoust: That's only for a single-message plot--which necessarily will ignore anything but the most recent message

packages/studio-base/src/panels/Plot/processor/messages.ts
cfoust: Theoretically yes, but it's a pretty negligible amount of compute, worth fixing later only if we really think it's a problem

packages/studio-base/src/panels/Plot/processor/messages.ts
```diff
@@ -32,106 +40,155 @@ export function receiveMetadata(
...
+ * Distribute new block data to all clients.
+ */
+export function addBlockData(update: BlockUpdate, state: State): StateAndEffects {
+  const { pending } = state;
+  const { updates, messages } = update;
+
+  // If we get updates for clients that haven't registered yet, we've got to
+  // keep that data around and use it when they register
+  const clientIds = state.clients.map(({ id }) => id);
+  const unused = updates.filter(({ id }) => !clientIds.includes(id));
```
jtbandes, line 132: Random drive-by thought as I'm reading this – if we are usually accessing `updates` by specific client ids, would it make sense to store them grouped by id instead of in a single array?

packages/studio-base/src/panels/Plot/processor/messages.ts
```diff
@@ -32,106 +40,155 @@ export function receiveMetadata(
...
+
+            const [start, end] = range;
+            const topicMessages = messages[topic];
+            if (topicMessages == undefined) {
+              return a;
+            }
+
+            return {
+              ...a,
+              [topic]: topicMessages.slice(start, end).flatMap((v) => v),
```
jtbandes, line 93: nit: I believe `.flatMap((v) => v)` is just `.flat()`

packages/studio-base/src/panels/Plot/blocks.ts
```diff
@@ -52,12 +50,161 @@ export function refreshBlockTopics(
...
+
+/**
+ * An Update describes the set of data that should be ingested by a client in
+ * the plot worker.
+ */
+export type Update = {
   topic: string;
+  // The range of blocks that should be ingested in the form [start index, end
+  // index).
   range: Range;
```
jtbandes, line 62: I think it would be helpful to name this something like `blockRange` – so the association with blocks doesn't get lost when reading code elsewhere that simply refers to `range`.

packages/studio-base/src/panels/Plot/processor/messages.ts
```diff
@@ -32,106 +40,155 @@ export function receiveMetadata(
...
+
+        const shouldReset = updates.some(
+          ({ update: { shouldReset: updateShouldReset } }) => updateShouldReset,
+        );
 
-export function addBlock(block: Messages, resetTopics: string[], state: State): StateAndEffects {
-  const { blocks, pending, metadata, globalVariables } = state;
+        // Build a consolidated set of messages by combining all of the updates
+        // together
+        const clientMessages = R.reduce(
```
jtbandes, line 79: This block feels like it's complex enough that it should be broken out to a separate function so it doesn't clutter up the main flow of `applyBlockUpdate`

packages/studio-base/src/panels/Plot/processor/messages.ts
```diff
@@ -32,106 +40,155 @@ export function receiveMetadata(
...
+            } = clientUpdate;
+
+            const [start, end] = range;
+            const topicMessages = messages[topic];
+            if (topicMessages == undefined) {
+              return a;
+            }
+
+            return {
+              ...a,
```
jtbandes, line 92: I'm mildly concerned that a proliferation of `reduce` with object spreads will lead to a lot more CPU work being done than it would be with a loop more like `const clientMessages = {}; for (const clientUpdate of updates) { clientMessages[clientUpdate.update.topic] = ... }`. In less-hot code paths or when N is small I'm sure it's not a big deal but I am gathering that this will be happening rapidly many times in a row for each topic during preloading?

packages/studio-base/src/panels/Plot/processor/messages.ts
jtbandes: Just for my understanding: in the case where we do have both blocks & current data, where is the logic that chooses to use the block data instead of the current data?

cfoust: For generating the final dataset? That happens [here](https://github.com/foxglove/studio/blob/685e3b34b55c9935116c02fbcbb90b438488a213/packages/studio-base/src/panels/Plot/processor/downsample.ts#L384)

packages/studio-base/src/panels/Plot/processor/clients.ts
```diff
@@ -115,45 +112,51 @@ export function updateParams(id: string, params: PlotParams, state: State): Stat
...
+        ...client,
+        params,
+        topics: getParamTopics(params),
+        downsampled: initDownsampled(),
+      });
     }),
-    keepEffects(evictCache),
     concatEffects((newState: State): StateAndEffects => {
-      const { pending, blocks } = newState;
+      const { pending } = newState;
```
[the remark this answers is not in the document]
cfoust: Yup, exactly. It was weird and complicated to have two different code paths for ingesting block data, one that used messages and one that used `BlockUpdate`s.

packages/studio-base/src/panels/Plot/processor/clients.ts
```diff
@@ -115,45 +112,51 @@ export function updateParams(id: string, params: PlotParams, state: State): Stat
...
-        return [newState, []];
-      }
+      const newPending: BlockUpdate[] = allUpdates
+        .filter(([unused]) => unused.updates.length > 0)
+        .map(([unused]) => unused);
 
-      const newTopics = Object.keys(newData);
+      const updatesToApply: BlockUpdate[] = allUpdates
+        .filter(([, used]) => used.updates.length > 0)
+        .map(([, used]) => used);
```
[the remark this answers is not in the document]
cfoust: While this section could use some improvement--no denying that--your code snippet copies the `BlockUpdate` (including all of its updates) on every iteration. The subtlety is that `ClientUpdate`s refer to data inside of the `messages` field on the `BlockUpdate`; you cannot just combine `BlockUpdate`s as it seems you're suggesting in the parenthetical at the end of your comment.

I can think of ways to clean this up, however! It's a good call-out that this is hard to understand.

packages/studio-base/src/panels/Plot/processor/messages.ts
```diff
@@ -32,106 +40,155 @@ export function receiveMetadata(
...
+ * Consolidate block data updates for each client and build new plots.
+ */
+export function applyBlockUpdate(update: BlockUpdate, state: State): StateAndEffects {
+  const { metadata, globalVariables } = state;
+  const { messages, updates: clientUpdates } = update;
+
+  // We aggregate all of the updates for each client and then apply them as a
+  // group. This is because we don't want the `shouldReset` field, which can
+  // reset the plot data, to throw away data we aggregated from an update we
+  // just applied.
```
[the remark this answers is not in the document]
cfoust: Maybe I could state this more clearly, but a single `BlockUpdate` contains a single "step" forward in the block state machine. In other words, this code cannot result in out-of-order application. The scenario it's avoiding is where you apply `ClientUpdate`s from the `updates` field one at a time. If one update contains `shouldReset=true`, all of them will for a given client, but if you apply them one by one, you might reset the client's data entirely (with `shouldReset`) several times.

ex, this goes from:
```
shouldReset update shouldReset update
```

to
```
shouldReset update update
```

packages/studio-base/src/panels/Plot/processor/messages.ts
cfoust: lol wow no idea what I was thinking

packages/studio-base/src/panels/Plot/blocks.ts
cfoust: 👍

packages/studio-base/src/panels/Plot/processor/messages.ts
cfoust: Great idea! I agree completely.

packages/studio-base/src/panels/Plot/processor/messages.ts
cfoust: In the worst case this code block runs N*M times while preloading data, where N is the number of blocks and M is the number of topics, so usually this is pretty small. In relative terms the compute cost of this is dwarfed by dataset generation and downsampling, where runtime complexity is obviously a function of the number of messages.

packages/studio-base/src/panels/Plot/processor/messages.ts
cfoust: While this is totally subjective and I'm not going to die on this hill, I have a mild preference in favor of lists over associative arrays, mostly because it makes mapping and filtering more ergonomic. In this case, however, it may make sense given that we just `groupBy` in `applyBlockUpdate` anyway, but I will think about it

packages/studio-base/src/panels/Plot/processor/clients.ts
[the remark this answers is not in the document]
jtbandes: What do you mean by copy? I don’t see anything in my snippet that would be creating new objects, just organizing them into new arrays right?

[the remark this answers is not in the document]
cfoust: “copy” here meaning “duplicate”, not in terms of memory layout. You are appending the entire BlockUpdate to the list for every ClientUpdate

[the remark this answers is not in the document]
cfoust: Which, though it’s clearly just copying the reference, is a little counterintuitive since then you have to reaggregate them when you apply all the pending updates for a client that joins later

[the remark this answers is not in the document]
jtbandes: I see, I think I get your point now. Would this be more accurate?
```ts
const newPending: BlockUpdate[] = [];
const updatesToApply: BlockUpdate[] = [];
for (const blockUpdate of pending) {
  const [used, unused] = R.partition(
    ({ id: clientId }) => clientIds.includes(clientId),
    blockUpdate.updates,
  );
  if (used.length > 0) {
    updatesToApply.push({ ...blockUpdate, updates: used });
  }
  if (unused.length > 0) {
    newPending.push({ ...blockUpdate, updates: unused });
  }
}
```

jtbandes (approved):

packages/studio-base/src/panels/Plot/processor/messages.ts
```diff
@@ -24,6 +25,9 @@ import { BlockUpdate, ClientUpdate } from "../blocks";
 import { Messages } from "../internalTypes";
 import { isSingleMessage } from "../params";
 
+// Maximum number of accumulated current messages before triggering a cull
+const ACCUMULATED_CURRENT_MESSAGE_CULL_THRESHOLD = 50_000;
```
achim-k, line 29: Should we also have an upper limit for memory size? Depending on the message size, 50k messages might be way more than the allowed JS heap size, so keeping on to the full message objects should be avoided.

defunctzombie: Possibly - though I think we use typedarrays for storing this data so it won't be the same kind of impact on the heap?

achim-k: Ah sorry, you are right. I was referring to the wrong code section, I actually meant https://github.com/foxglove/studio/blob/b31eca452057caaa24283d76fd7a79d6df112589/packages/studio-base/src/panels/Plot/useDatasets.ts#L137 where we [concatenate](https://github.com/foxglove/studio/blob/main/packages/studio-base/src/panels/Plot/clients.ts#L207-L216) new current message events to existing ones, and there seems to be no culling in place like it is here. I'm pretty sure that this is one of the main reasons for OOM crashes.

achim-k: -> https://github.com/foxglove/studio/pull/7298

PR #7236, opened 2023-12-15
Add memory info to player state

achim-k:
**User-Facing Changes**
None

**Description**
Adds a new `memoryInfo` object to `PlayerState` which holds information of the current amount of buffered and preloaded messages. This mainly applies for non-live data sources where messages can be buffered (for smoother playback) or preloaded (e.g. for plots).

Part of FG-6099

### DIFF
--- a/packages/studio-base/src/players/IterablePlayer/BlockLoader.ts
+++ b/packages/studio-base/src/players/IterablePlayer/BlockLoader.ts
@@ -25,6 +25,8 @@ import { IIterableSource, MessageIteratorArgs } from "./IIterableSource";
 
 const log = Log.getLogger(__filename);
 
+export const MEMORY_INFO_PRELOADED_MSGS = "Preloaded messages";
+
 type BlockLoaderArgs = {
   cacheSizeBytes: number;
   source: IIterableSource;
@@ -42,7 +44,7 @@ type CacheBlock = MessageBlock & {
 type Blocks = (CacheBlock | undefined)[];
 
 type LoadArgs = {
-  progress: (progress: Progress, cacheSize: number) => void;
+  progress: (progress: Progress) => void;
 };
 
 /**
@@ -185,7 +187,7 @@ export class BlockLoader {
 
     // Ignore changing the blocks if the topic list is empty
     if (topics.size === 0) {
-      args.progress(this.#calculateProgress(topics), this.#cacheSize());
+      args.progress(this.#calculateProgress(topics, this.#cacheSize()));
       return;
     }
 
@@ -334,7 +336,7 @@ export class BlockLoader {
             });
             // We need to emit progress here so the player will emit a new state
             // containing the problem.
-            progress(this.#calculateProgress(topics), totalBlockSizeBytes);
+            progress(this.#calculateProgress(topics, totalBlockSizeBytes));
             return;
           }
         }
@@ -367,7 +369,7 @@ export class BlockLoader {
         // (The size of new messages is already added above).
         totalBlockSizeBytes -= overridenBlockMessagesSize;
 
-        progress(this.#calculateProgress(topics), totalBlockSizeBytes);
+        progress(this.#calculateProgress(topics, totalBlockSizeBytes));
       }
 
       await cursor.end();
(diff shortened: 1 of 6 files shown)


packages/studio-base/src/players/IterablePlayer/BlockLoader.ts
```diff
@@ -42,7 +42,7 @@ type CacheBlock = MessageBlock & {
 type Blocks = (CacheBlock | undefined)[];
 
 type LoadArgs = {
-  progress: (progress: Progress) => void;
+  progress: (progress: Progress, cacheSize: number) => void;
```
snosenzo, line 45: why not include it in progress?

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
```diff
@@ -1075,12 +1087,15 @@ export class IterablePlayer implements Player {
...
   async #startBlockLoading() {
     await this.#blockLoader?.startLoading({
-      progress: async (progress) => {
+      progress: async (progress, cacheSize) => {
         this.#progress = {
           fullyLoadedFractionRanges: this.#progress.fullyLoadedFractionRanges,
           messageCache: progress.messageCache,
         };
-
+        this.#memoryInfo = {
```
snosenzo, line 1095: seems to be coupled with `#progress` update locations, so might as well leverage that type and add this info to it?

achim-k: So you basically propose to include the new `memoryInfo` object into the PlayerState's `progress` prop? I can do that

defunctzombie: Agree with @snosenzo that I'd probably throw it into "progress". Progress is a bit of a sad name itself but since its there might as well use it. Otherwise the event you emit is somewhat awkward in it being a `progress` event and randomly has this second argument that some cache size?

packages/studio-base/src/players/types.ts
```diff
@@ -94,6 +94,14 @@ export type PlayerURLState = Immutable<{
   parameters?: Record<string, string>;
 }>;
 
+// Contains details about a Player's memory usage
```
defunctzombie, line 97: Switch these over to use `/** */` style comments and we will get nicer vscode integration by seeing the comments when hovering over a field where it is used in the code.

packages/studio-base/src/players/types.ts
```diff
@@ -94,6 +94,14 @@ export type PlayerURLState = Immutable<{
...
+export type PlayerMemoryInfo = Readonly<{
+  // Size in bytes of all buffered messages.
```
defunctzombie, line 99: You use `size` here but `amount` in the api call of the source. Maybe `amount` makes more sense in that context but it does seem like you use `size` everywhere and `amount` only on that one method. Consider using `size` instead of `amount`.

packages/studio-base/src/players/types.ts
```diff
@@ -94,6 +94,14 @@ export type PlayerURLState = Immutable<{
...
+  bufferedMsgsSize?: number;
+  // Size in bytes of all preloaded messages.
+  preloadedMsgsSize?: number;
```
defunctzombie, line 102: not for this PR - but a thought I'm having as I read this is that our use of `preloaded` is rather confusing if you consider that `buffered` data is also in some ways preloaded. I don't have better terms but I do feel like it makes me have to think harder about the code than I'd hope

defunctzombie (approved):
👍 I would drop the extra argument to the _progress_ event and put the block cache size  in the existing progress structure.

Another thought is to make the memory info a Map. Then a player could add arbitrary memory tracking items and performance could show that. This comes to mind as something that would allow the websocket player to expose its worker buffer using this same scheme and the performance sidebar could show that.

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
```diff
@@ -983,6 +985,10 @@ export class IterablePlayer implements Player {
       this.#progress = {
         fullyLoadedFractionRanges: this.#bufferedSource.loadedRanges(),
         messageCache: this.#progress.messageCache,
+        memoryInfo: {
+          ...this.#progress.memoryInfo,
+          [MEMORY_INFO_KEY_BUFFERED_MSGS_SIZE]: this.#bufferedSource.getCacheSize(),
```
achim-k, line 990: Instead of calling `getCacheSize()` here, I could change the buffered source's `loadedRanges()` to a `getProgress()` which includes the memory info object. WDYT?

snosenzo (approved):

PR #7237, opened 2023-12-16
Change dependabot updates to monthly

defunctzombie:
### User-Facing Changes
None

### Description

Dependabot is out of control - we don't need daily updates of our npm packages. If there's an actionable bug we will update quicker - otherwise monthly is sufficient.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,8 +3,10 @@ updates:
   - package-ecosystem: npm
     directory: "/"
     schedule:
-      interval: daily
-    open-pull-requests-limit: 10
+      # We found daily updates to be too noisy. They consumed too much CI time and engineer review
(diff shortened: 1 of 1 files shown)


jtbandes (approved):
That de-escalated quickly.

snosenzo (approved):
Why monthly and not weekly or bi-weekly as a next step?

defunctzombie:
> Why monthly and not weekly or bi-weekly as a next step?

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#scheduleinterval

There is only daily, weekly, and monthly. Maybe my opinion will change after giving this a try but I think even weekly is too aggressive and distracts us from our regular work to garden what are non-critical updates.

PR #7256, opened 2023-12-18
Improvements to message path autocomplete

jtbandes:
**User-Facing Changes**
Improved autocomplete suggestions when typing message path syntax.

**Description**
The logic that generated the initial combined (topics+paths) list of suggestions was bespoke and did not account for array fields. This PR changes it to use the existing `messagePathsForStructure` function (with some additional info returned). It also updates `messagePathsForStructure` to extend the typicalFilterName logic to strings, and fix a bug where quotes in string filters (`...{field=="value"}`) were being deleted.

Follow-up to #1971 & #3190

Before | After
--- | ---
I am not sure why the header sub-fields were not shown: <br>image | image
image | Filter and array indexing suggestions are now shown:<br> image

### DIFF
--- a/packages/studio-base/src/components/MessagePathSyntax/MessagePathInput.tsx
+++ b/packages/studio-base/src/components/MessagePathSyntax/MessagePathInput.tsx
@@ -16,15 +16,11 @@ import * as _ from "lodash-es";
 import { CSSProperties, useCallback, useMemo } from "react";
 import { makeStyles } from "tss-react/mui";
 
-import { MessageDefinitionField } from "@foxglove/message-definition";
-import { Immutable } from "@foxglove/studio";
 import * as PanelAPI from "@foxglove/studio-base/PanelAPI";
 import Autocomplete, { IAutocomplete } from "@foxglove/studio-base/components/Autocomplete";
 import useGlobalVariables, {
   GlobalVariables,
 } from "@foxglove/studio-base/hooks/useGlobalVariables";
-import { Topic } from "@foxglove/studio-base/players/types";
-import { RosDatatypes } from "@foxglove/studio-base/types/RosDatatypes";
 
 import { RosPath, RosPrimitive } from "./constants";
 import {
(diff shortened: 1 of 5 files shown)


packages/studio-base/src/components/MessagePathSyntax/MessagePathInput.tsx
```diff
@@ -188,6 +117,24 @@ const useStyles = makeStyles()({
...
+ *
+ *  <MessagePathInput path={this.state.path} onChange={path => this.setState({ path })} />
+ *
+ * To limit the autocomplete items to only certain types of values, you can use
+ *
+ *  <MessagePathInput types={["message", "array", "primitives"]} />
+ *
+ * Or use actual ROS primitive types:
+ *
+ *  <MessagePathInput types={["uint16", "float64"]} />
```
defunctzombie, line 131: Is this _types_ feature used anywhere in the codebase?

packages/studio-base/src/components/MessagePathSyntax/MessagePathInput.tsx
```diff
@@ -207,7 +154,37 @@ export default React.memo<MessagePathInputBaseProps>(function MessagePathInput(
     variant = "standard",
   } = props;
   const { classes } = useStyles();
-  const topicFields = useMemo(() => getFieldPaths(topics, datatypes), [datatypes, topics]);
+
+  const messagePathStructuresForDataype = useMemo(
+    () => messagePathStructures(datatypes),
+    [datatypes],
+  );
+  const allStructureItemsByPath = useMemo(
```
defunctzombie, line 162: A quick blurb on what this chunk of code is doing would aid reading this later or any important bits about this implementation.

packages/studio-base/src/components/MessagePathSyntax/MessagePathInput.tsx
```diff
@@ -408,10 +381,12 @@ export default React.memo<MessagePathInputBaseProps>(function MessagePathInput(
                   validTypes,
                   noMultiSlices,
                   messagePath: rosPath.messagePath,
-                }).filter(
-                  // .header.seq is pretty useless but shows up everryyywhere.
-                  (msgPath) => msgPath !== "" && !msgPath.endsWith(".header.seq"),
-                ),
+                })
+                  .filter(
+                    // .header.seq is pretty useless but shows up everryyywhere.
```
defunctzombie, line 386:
```suggestion
                    // .header.seq is pretty useless so we explicitly filter it out of results
```

packages/studio-base/src/components/MessagePathSyntax/messagePathsForDatatype.ts
```diff
@@ -164,12 +164,12 @@ export function messagePathsForStructure(
     noMultiSlices?: boolean;
     messagePath?: MessagePathPart[];
   } = {},
-): string[] {
+): { path: string; terminatingStructureItem: MessagePathStructureItem }[] {
```
defunctzombie, line 167: The return type does not match the function comment anymore. What's this new thing you are adding?

defunctzombie (commented):
What's the gist of the change here? what was happening wrong before and what has been fixed?

I see some logic that is suppose to filter out `header.seq` but I see that in your screenshot. Is that logic no longer useful and should be removed? Or is the screenshot exposing a bug?

defunctzombie (approved):

packages/studio-base/src/components/MessagePathSyntax/MessagePathInput.tsx
jtbandes: It's been renamed to `validTypes`, but yes, we actually pipe it through the extension API: https://github.com/foxglove/studio/blob/c404fae2540c0eb45e42dc94af238f751177185b/packages/studio/src/index.ts#L558

jtbandes:
> I see some logic that is suppose to filter out `header.seq` but I see that in your screenshot. Is that logic no longer useful and should be removed? Or is the screenshot exposing a bug?

I believe that is a different code path that runs when you have already started typing a partially valid path and drilled down to a topic. You can see header seq is not included here:

image

image


That said, I feel like the difference between the two code paths is kind of weird so I'm happy to remove the logic that filters out seq.

jtbandes:
> What's the gist of the change here? what was happening wrong before and what has been fixed?

Updated PR description:

> The logic that generated the initial combined (topics+paths) list of suggestions was bespoke and did not account for array fields. This PR changes it to use the existing `messagePathsForStructure` function (with some additional info returned). It also updates `messagePathsForStructure` to extend the typicalFilterName logic to strings, and fix a bug where quotes in string filters (`...{field=="value"}`) were being deleted.

PR #7258, opened 2023-12-18
Fix error memory estimation handling of null values

snosenzo:
**User-Facing Changes**

- Fix errors decoding messages with null values

**Description**

We weren't handling `typeof null === "object"` in `estimateObjectSize`.

### DIFF
--- a/packages/studio-base/src/players/messageMemoryEstimation.test.ts
+++ b/packages/studio-base/src/players/messageMemoryEstimation.test.ts
@@ -134,6 +134,18 @@ describe("memoryEstimationByObject", () => {
     expect(sizeInBytes).toBeGreaterThan(0);
   });
 
+  it("estimates size of null object to be greater than 0", () => {
+    // eslint-disable-next-line no-restricted-syntax
+    const sizeInBytes = estimateObjectSize(null);
+    expect(sizeInBytes).toBeGreaterThan(0);
+  });
+
+  it("estimates size of undefined object to be greater than 0", () => {
+    // eslint-disable-next-line no-restricted-syntax
+    const sizeInBytes = estimateObjectSize(null);
+    expect(sizeInBytes).toBeGreaterThan(0);
+  });
+
   it("correctly estimates the size for a simple object", () => {
     const sizeInBytes = estimateObjectSize({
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/players/messageMemoryEstimation.ts
```diff
@@ -251,12 +256,14 @@ export function estimateObjectSize(obj: unknown): number {
...
-      const valuesSize = Object.values(obj!).reduce((acc, val) => acc + estimateObjectSize(val), 0);
+      const valuesSize = Object.values(obj).reduce((acc, val) => acc + estimateObjectSize(val), 0);
       return OBJECT_BASE_SIZE + propertiesSize + valuesSize;
     }
     case "symbol":
     case "function": {
       throw new Error(`Can't estimate size of type '${typeof obj}'`);
     }
   }
+  // Should never happen
```
defunctzombie, line 267: Should this throw? Or maybe leave a comment something like: while we don't think it should happen we want to be safe and return 0 size?

defunctzombie (approved):

packages/studio-base/src/players/messageMemoryEstimation.ts
snosenzo: Updated to do a `log.error` and return `SMALL_INTEGER_SIZE` to be safe, so that we don't prevent decoding of the message like we did previously

packages/studio-base/src/players/messageMemoryEstimation.test.ts
```diff
@@ -134,6 +134,18 @@ describe("memoryEstimationByObject", () => {
...
 
+  it("estimates size of null object to be greater than 0", () => {
+    // eslint-disable-next-line no-restricted-syntax
+    const sizeInBytes = estimateObjectSize(null);
+    expect(sizeInBytes).toBeGreaterThan(0);
+  });
+
+  it("estimates size of undefined object to be greater than 0", () => {
+    // eslint-disable-next-line no-restricted-syntax
+    const sizeInBytes = estimateObjectSize(null);
```
achim-k, line 145: This should have been `undefined`?

achim-k (commented):
Thanks for fixing!

PR #7259, opened 2023-12-19
Add background color to chip so it's visible on light theme with dark background

2metres:
**User-Facing Changes**
None

**Description**
Resolve bug where messages were not visible on dark backgrounds


defunctzombie (approved):
And this will be correctly flipped in dark mode?

snosenzo:
Left is after, right is before.
image

snosenzo:
For dark mode with light background: 
image

snosenzo (approved):

PR #7263, opened 2023-12-19
Fix phantom error after hot reloading

snosenzo:
**User-Facing Changes**

n/a

**Description**
This line was causing a type error after hot reloading, or if the file was changed the vscode ts server would catch it, but not initially in either case.


achim-k (approved):

defunctzombie (approved):
Weird but ok. fyi we tend to use `import { Immutable } from "@foxglove/studio"` for consistency but this is also fine.

PR #7264, opened 2023-12-19
Reduce limit of buffered messages from 1GB to 600 MB

achim-k:
**User-Facing Changes**
Reduce limit of buffered messages to mitigate OOM crashes

**Description**
This PR reduces the limit of buffered messages from 1GB to 600 MB. Previously, with the cache size for preloaded and buffered messages both being limited to 1GB, the app could go well beyond the 2GB JS heap limit at which the app eventually gets killed by the browser (or electron). By reducing the cache size of buffered messages, we make these OOM crashes less likely. The downside of this approach is that less messages can be buffered which can be noticed mainly by the amount of messages buffered behind the current playback cursor.

### DIFF
--- a/packages/studio-base/src/players/IterablePlayer/CachingIterableSource.ts
+++ b/packages/studio-base/src/players/IterablePlayer/CachingIterableSource.ts
@@ -101,7 +101,7 @@ class CachingIterableSource extends EventEmitter<EventTypes> implements IIterabl
     super();
 
     this.#source = source;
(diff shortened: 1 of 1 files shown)


snosenzo (approved):

defunctzombie (approved):
I think its worth adding some shorter version of your PR description as a code comment above the change `maxTotalSizeBytes` variable so we can remember some justification on 600MB.

I am 👍 to this change in the short term to allow us time to explore better caching strategies - primarily for streaming data where we want to avoid re-fetching if we can.

PR #7266, opened 2023-12-20
Add tooltip to 3d floating buttons

2metres:
**User-Facing Changes**
None

**Description**
Use MUI tooltip instead of native title element

https://github.com/foxglove/studio/assets/924528/b14006a0-e979-4a03-b578-f9673f0d0c36

### DIFF
--- a/packages/studio-base/src/components/ExpandingToolbar.tsx
+++ b/packages/studio-base/src/components/ExpandingToolbar.tsx
@@ -3,7 +3,7 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
 import { ArrowMinimize20Filled } from "@fluentui/react-icons";
-import { Paper, IconButton, Tabs, Tab, tabClasses, tabsClasses } from "@mui/material";
+import { Paper, IconButton, Tabs, Tab, tabClasses, tabsClasses, Tooltip } from "@mui/material";
 import { ReactElement, ReactNode } from "react";
 import { makeStyles } from "tss-react/mui";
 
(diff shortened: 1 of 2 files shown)


defunctzombie:
The reason we hadn't done this before is because the mui tooltip has very poor performance for re-rendering when we last profiled it and I am hesitant to add another use of it in a location that renders at 60fps. Have we fixed these performance concerns?

defunctzombie (dismissed):
See comment about performance concerns. Convince me this is ok to add to a panel rendering at 60fps.

2metres:
Do we have something documenting the how to test the performance benchmarks that we need to pass for this?

2metres:
Throwing a tooltip around a button in the ui-benchmarking tool certainly doesn't bode well. 

Screen Shot 2023-12-21 at 9 09 40 am

jtbandes:
One would hope that the `RendererOverlay` is not re-rendering at 60fps, during panning or playback. I am not saying I know for sure it doesn't – just that I would _hope_ it doesn't 😅

2metres:
> One would hope that the `RendererOverlay` is not re-rendering at 60fps, during panning or playback. I am not saying I know for sure it doesn't – just that I would _hope_ it doesn't 😅

From memory this should be true. But I also have no way of testing this

2metres:
From what I can tell the RendererOverlay does not re-render at 60fps. I think this PR is safe to merge.

defunctzombie (approved):
👍

Unrelated to the PR but it would be nice to find ways of using smaller component files for some of these things. It would make it more obvious which components use what state and selectors and what will cause them to re-render. I think we should be doing that across the codebase more.

PR #7267, opened 2023-12-21
Store buffered messages as typed arrays

achim-k:
**User-Facing Changes**
Store buffered messages as raw bytes to reduce javascript heap memory pressure

**Description**

Summary from @defunctzombie:

> Our current caching strategy for recently played messages (and upcoming messages) is to keep the deserialized messages in memory. The upside is that we have immediate access to the deserialized messages on seek and playback without having to deserialize the messages again. The downside is this caching uses valuable JS heap space. Instead of caching deserialized messages we can look at caching the message bytes (i.e. mcap message records) in typed arrays which are not subject to the same js heap limits.

While this PR reduces heap memory consumption, the downside is that buffered messages are now deserialized on the main thread which may cause a noticeable performance degradation. A possible solution for this would be to move the buffering layer to the worker thread where the iterable source is being read.

### DIFF
--- a/packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
+++ b/packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
@@ -9,6 +9,7 @@ import { v4 as uuidv4 } from "uuid";
 import { debouncePromise } from "@foxglove/den/async";
 import { filterMap } from "@foxglove/den/collection";
 import Log from "@foxglove/log";
+import { parseChannel } from "@foxglove/mcap-support";
 import {
   Time,
   add,
@@ -175,6 +176,8 @@ export class IterablePlayer implements Player {
   readonly #sourceId: string;
 
   #untilTime?: Time;
+
+  #deserializersByTopic: Record<string, (data: ArrayBufferView) => unknown> = {};
 
   /** Promise that resolves when the player is closed. Only used for testing currently */
   public readonly isClosed: Promise<void>;
@@ -504,6 +507,18 @@ export class IterablePlayer implements Player {
         }
 
         uniqueTopics.set(topic.name, topic);
+
+        if (topic.messageEncoding && topic.schemaData && topic.schemaEncoding && topic.schemaName) {
+          const { deserialize } = parseChannel({
+            messageEncoding: topic.messageEncoding,
+            schema: {
+              data: topic.schemaData,
+              encoding: topic.schemaEncoding,
+              name: topic.schemaName,
+            },
+          });
+          this.#deserializersByTopic[topic.name] = deserialize;
+        }
       }
 
       this.#providerTopics = Array.from(uniqueTopics.values());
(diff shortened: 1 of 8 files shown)


packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
```diff
@@ -504,6 +507,24 @@ export class IterablePlayer implements Player {
...
+          const { deserialize } = parseChannel({
+            messageEncoding: topic.messageEncoding,
+            schema: {
+              data: topic.schemaData,
+              encoding: topic.schemaEncoding,
+              name: topic.schemaName,
+            },
+          });
+          this.#deserializersBySchema[topic.schemaName] = deserialize;
+        }
```
achim-k, line 527: This is not necessary when the underlying source does not support yielding raw, non-deserialized messages. It would make sense to detect this and skip channel parsing in that case.

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
```diff
@@ -663,9 +684,21 @@ export class IterablePlayer implements Player {
         }
 
         if (iterResult.type === "message-event") {
+          let msgEvent = iterResult.msgEvent;
+          if (iterResult.msgEvent.rawBytes === true) {
```
achim-k, line 688: Instead of adding the `rawBytes` field, I considered adding another iterator result type `raw-message-event` (with `Message<Uint8Array>`) which would not require the casting below. However, I decided against it as it would require a lot more code changes.

packages/studio/src/index.ts
```diff
@@ -125,6 +125,11 @@ export type MessageEvent<T = unknown> = {
    * un-converted message event.
    */
   originalMessageEvent?: MessageEvent;
+
+  /**
+   * Message is not deserialized and contains the raw bytes.
+   */
+  rawBytes?: boolean;
```
snosenzo, line 132: `rawBytes` doesn't really feel like a boolean. can we do something like `needsDeserialization` or `isSerialized`?

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
snosenzo: isn't `parseChannel` already done somewhere else? 
I feel like we shouldn't be parsing the channel twice and it should be shared from the source

snosenzo (commented):
implementation looks fine for the most part, but I'm curious what the final performance and memory differences are and how much longer it takes to deserialize on the main thread.

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
achim-k: Sharing is hard, as a deserialize function is not structured cloneable (to send it from worker to main thread). While not impossible (I guess we would have to write a comlink transfer handler for that), it doesn't seem like the cleanest solution.

achim-k:
> I'm curious what the final performance and memory differences are and how much longer it takes to deserialize on the main thread.

This is hard to answer, it mostly depends on what topics are subscribed and their message and schema encoding. However, from testing various layouts + datasets it seems that the additional main thread load is < 5%

defunctzombie:
> A possible solution for this would be to move the buffering layer to the worker thread where the iterable source is being read.

Any insights on what it would take to do this? It's been a while since I looked into the iterable player stuff - last I recall only the lower level source was in the worker and the Buffered and Caching source were not?

packages/studio/src/index.ts
defunctzombie: Could we have this be a `Uint8Array` (or similar) and store the message bytes in that if available? The `message` field would be undefined until parsed into it.

Could be an interesting thing to make available to other parts of the app.

defunctzombie:
One thing that makes me uneasy is the re-use of the _message_ field for raw data. We already abuse this _message_ field for slices where we don't have a full set of fields even though it is suppose to be the message. Having the _bytes_ in that field is more accurate than some subset of fields - but that makes it more likely that they somehow "leak" out as the message to other parts of the app if we are not diligent; at least that's my worry. A separate field for the message bytes feels more robust and less prone to error. The downside is now we are likely to send along the bytes with each MessageEvent and maybe panels end up storing that as well (when they want to cache a message event) and that seems not so ideal - an example would be user scripts - would this separate field be sent over postMessage as well? Seems like in many cases you don't need to be dealing with the raw message data and it is extra overhead to send it and make it available.

packages/studio/src/index.ts
snosenzo: Are you suggesting letting panels deserialize messages?

defunctzombie: nope - but it would open the door for a "hex viewer" panel. Not sure if I think that's worth the exposure to all the questions about "how can my panel deserialize the data" though

Issue #7278, opened 2023-12-28
The plot component has a problem displaying the chart when adding a curve.


Woodii1998:
**Description**
Plot component curves are lost when adding curves
The plot component loses some curves when adding curves.

- Version: 1.82.0+dcf781b.f63d3af
- Platform: web
- Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native): mcap

**Steps To Reproduce**

https://github.com/foxglove/studio/assets/33593572/cac96841-799c-4d4c-a160-ba1068b9c7f5

**Expected Behavior**
When you add a curve, you can show all the curves.

Woodii1998:
mcap file is here, if you want

https://drive.google.com/file/d/1K042_0shfzOBv4FkieIsbqL-LMIIJ-mp/view?usp=sharing

defunctzombie:
@Woodii1998 could you give this preview deployment a try: https://studio-hdnbgt93s.foxglove.party/ I think it will fix your issue

Woodii1998:
> @Woodii1998 could you give this preview deployment a try: https://studio-hdnbgt93s.foxglove.party/ I think it will fix your issue

Looking good, looks like the issue is fixed!

PR #7279, opened 2024-01-01
Plots!

defunctzombie:
**User-Facing Changes**
Plot panel bugfixes

**Description**
Over the past few months there's been a lot of work on the plot panel to improve performance for large datasets. The performance has improved (yet is still not meeting expectations) and the number of bug reports as skyrocketed. Additionally, the code changes introduced thousands of lines of code that no current team member understands and can effectively debug to resolve the outstanding plot bugs.

Rather than bandaid the plot and continue with a complex and fragile architecture, I've taken a different tactic and propose a complete overhaul of the architecture while maintaining the important performance characteristics.

In this PR I do away with the following concepts from the Plot panel code:
- Processor: the recently added code that is large, complex, fragile, and hard to debug). It is responsible for some form of coordination between downsampling and rendering.
- PlotChart (and the many hooks): another layer wrapping the TimeBasedChart component
- TimeBasedChart: the react layer(s) wrapping the chartjs interface (and ultimately the worker that renders the chart data)
- useDataset (and other hooks): the layers of hooks responsible for figuring out datasets? unclear what it does now with the processor
- many other support files related to powering the above

To replace the above I've implemented:
- Plot.tsx: The plot panel component. This sets up interfaces between the "react" layer (message pipeline, DOM elements, mouse events, config, etc) and the chart logic (dataset building, rendering) which does not use react and does away with the many many hooks for dataset fetching.
- PlotCoordinator.ts: Receives inputs from the "react" layer (DOM events, hover updates, player state, config, tooltips) and dispatches them to the the DatasetsBuilder or the ChartRenderer depending on the update. It gets downsampled datasets from the Builder to the the ChartRenderer. The builder and renderer run in workers so the main thread UX interactions remain fast while downsampling and rendering.
- OffscreenCanvasRenderer & ChartRenderer.ts: Draws chart datasets onto an offscreen canvas.
- IDatasetsBuilder.ts: Accumulates raw datasets and provides an interface to query for downsampled chart datasets.
- BlockTopicCursor: Tracks the processed block location for a topic. Each series has a cursor which identifies the _next_ blocks of data that need sending to the dataset builder. When the cursor detects the underlying blocks changed it resets and we can re-accumulate the updated series data.

Note: The panel does not pause/resume frame

Fixes: #7278
Fixes: FG-5381
Fixes: FG-6190
Fixes: FG-6181
Fixes: FG-6179
Fixes: FG-6171
Fixes: FG-6184
Fixes: FG-6218
Fixes: FG-6192
Fixes: FG-6189
Fixes: FG-6151
Fixes: https://github.com/orgs/foxglove/discussions/549

---
Showstoppers:

Errata:

- [x] story preloaded data in binary blocks (issue is same message path different series w/ receiveTime vs headerStamp) https://github.com/foxglove/studio/pull/7324
- [ ] When using "show value" in the legend - if you remove a series the value of the series that remains disappears 
- [ ] two series with the same message path and same timestamp method? (roman: if they use the same timestamp mode they will appear as the same series - we can disambiguate them by their _index_ in the paths list but if one of them is removed then it will appear like the remaining one is a _new_ series (different index) and we will need to re-send the data to the worker. 
- [ ] `grid.data[:]` crashes with the tf grid dataset

TODO perf/cleanup:
- [ ] use partial subscriptions for index mode and current msg path mode
- [x] one version of getChartValue, readMessagePathItems, isChartValue
- [ ] "split down" on mcnasty-5m causes the UI to lock up for 0.5-3s. if I try to split many panels too quickly it can crash the tab (roman: split current processing from blocks to dispatch render updates between blocks when all data is already preloaded)
- [x] more testing with data platform
- [ ] avoid dataset re-downsample when playing or hovering if no dataset changes
- [ ] a path like `/mcnasty/position.` resulted in a subscription with an "" empty field name when really this path is not yet valid and we should not update subscriptions
- [ ] probably don't need the useEffect to set renderer and can add a canvas element directly to the react via jsx and get a ref on that instead of the canvasDiv flow (noted by roman)
- [x] normalize config before sending downstream (maybe add the computed key and apply global variables as well)
- [ ] deduplicate mock Worker code for unit tests
- [ ] fillInGlobalVariablesInPath is done in each datasets builder - would be nice to do once at top level
- [ ] have renderer request viewport dataset directly instead of hop through main thread
- [ ] `void this.#datasetsBuilderRemote.setConfig(this.#seriesConfigs);` needs error handling - maybe move the sending of the config before sending the datasets since that is already in an async function
- [ ] "RosPath" stuff needs to be renamed as MessagePath and put in a new _package_ in the repo.
- [ ] incremental downsampling
- [x] Add _key_ to IndexDatasetsBuilder series and CurrentCustomDatasetsBuilder for lookup of series that changes after global variables change.
- [ ] This code to build subscriptions from series could be moved into a separate function/file and given unit tests
- [ ] https://github.com/foxglove/studio/pull/7279#discussion_r1445530771
- [ ] https://github.com/foxglove/studio/pull/7279#discussion_r1451030540

---
When plotting an enum value the tooltip can show the "text"
image


amacneil:
image

amacneil:
Note from our call: I'm able to crash the tab fairly easily with mcnasty-5m by splitting and panning tabs. Happy to test more after the next round of code lands.

packages/studio-base/src/panels/Plot/builders/TimeseriesDatasetsBuilderImpl.ts
```diff
@@ -0,0 +1,324 @@
...
+        // Limit the total current datums for any series so they do not grow indefinitely
+        const cullSize = Math.max(
+          0,
+          series.current.length + action.items.length - MAX_CURRENT_DATUMS,
+        );
+        series.current.splice(0, cullSize);
+
+        const sorted =
+          series.config.timestampMethod === "headerStamp"
+            ? action.items.slice().sort(compareDatum)
```
jtbandes, line 268: could we perhaps say that this function takes "ownership" of `action.items` and is therefore allowed to sort in-place? or is the calling code actually still referencing this array?

packages/studio-base/src/panels/Plot/builders/TimeseriesDatasetsBuilderImpl.ts
```diff
@@ -0,0 +1,324 @@
...
+        if (series.config.timestampMethod === "headerStamp") {
+          series.full.sort(compareDatum);
+        }
+
+        // trim current data to remove values present in the full data
+        const lastX = series.full[series.full.length - 1]?.x;
+        if (lastX != undefined) {
+          let idx = 0;
+          for (const item of series.current) {
+            if (item.x > lastX) {
```
jtbandes, line 317: this could probably done with a `findIndex` or `findLastIndex`?

packages/studio-base/src/panels/Plot/csv.ts
```diff
@@ -2,44 +2,52 @@
...
-  const { x, value, receiveTime, headerStamp } = data;
+function getCSVRow(label: string, data: CsvDataset["data"][0]) {
+  const { x, y, receiveTime, headerStamp } = data;
   const receiveTimeFloat = formatTimeRaw(receiveTime);
   const stampTime = headerStamp ? formatTimeRaw(headerStamp) : "";
-  return [x, receiveTimeFloat, stampTime, label, value];
+  return [x, receiveTimeFloat, stampTime, label, y];
 }
 
 const getCVSColName = (xAxisVal: PlotXAxisVal): string => {
```
jtbandes, line 19:
```suggestion
const getCSVColName = (xAxisVal: PlotXAxisVal): string => {
```

packages/studio-base/src/panels/Plot/builders/CustomDatasetsBuilder.ts
```diff
@@ -0,0 +1,324 @@
...
+        value: chartValue,
+        receiveTime: event.receiveTime,
+      });
+    }
+  }
+
+  return out;
+}
+
+function getChartValue(value: unknown): number | undefined {
```
jtbandes, line 285: this function needs to be deduped

packages/studio-base/src/panels/Plot/builders/CustomDatasetsBuilder.ts
```diff
@@ -0,0 +1,324 @@
...
+    this.#pendingDataDispatch.push({
+      type: "reset-current-x",
+    });
+
+    this.#pendingDataDispatch.push({
+      type: "reset-full-x",
+    });
+  }
+
+  public setConfig(config: Immutable<PlotConfig>, globalVariables: GlobalVariables): void {
```
jtbandes, line 191: Seems like some of this stuff could maybe be deduped/shared between the different builders? or do you want to keep them totally separate?

packages/studio-base/src/panels/Plot/builders/CustomDatasetsBuilder.ts
```diff
@@ -0,0 +1,324 @@
...
+        this.#pendingDataDispatch.push({
+          type: "append-current",
+          series: seriesConfig.messagePath,
+          items: pathItems,
+        });
+      }
+    }
+
+    const blocks = state.progress.messageCache?.blocks;
+    if (blocks) {
```
jtbandes, line 118: early return? there's a lot of nesting going on here  -- or maybe the answer is to move this part into a separate testable function?

packages/studio-base/src/panels/Plot/builders/CurrentCustomDatasetsBuilder.ts
```diff
@@ -0,0 +1,227 @@
...
+    if (!activeData || !this.#parsedPath) {
+      return;
+    }
+
+    const msgEvents = activeData.messages;
+    if (msgEvents.length === 0) {
+      return;
+    }
+
+    for (let i = msgEvents.length - 1; i >= 0; --i) {
```
jtbandes, line 54: `findLast`

packages/studio-base/src/panels/Plot/builders/CurrentCustomDatasetsBuilder.ts
```diff
@@ -0,0 +1,227 @@
...
+
+    if (this.#xValues.length === 0) {
+      return;
+    }
+
+    const range: Bounds1D = { min: 0, max: 0 };
+    for (const series of this.#seriesByMessagePath.values()) {
+      // loop over the events backwards and once we find our first matching topic
+      // read that for the path items
+      for (let i = msgEvents.length - 1; i >= 0; --i) {
```
jtbandes, line 85: `findLast`

packages/studio-base/src/panels/Plot/builders/BlockTopicCursor.ts
```diff
@@ -0,0 +1,68 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import { Immutable } from "@foxglove/studio";
+import { MessageEvent } from "@foxglove/studio";
+import { MessageBlock } from "@foxglove/studio-base/players/types";
+
+/**
+ * BlockCursor tracks the last seen block messages for a given topic and can produce the next
```
jtbandes, line 10:
```suggestion
 * BlockTopicCursor tracks the last seen block messages for a given topic and can produce the next
```

packages/studio-base/src/panels/Plot/builders/CurrentCustomDatasetsBuilder.ts
```diff
@@ -0,0 +1,227 @@
...
+
+type SeriesItem = {
+  enabled: boolean;
+  messagePath: string;
+  parsed: RosPath;
+  dataset: ChartDataset<"scatter", DatumWithReceiveTime[]>;
+};
+
+export class CurrentCustomDatasetsBuilder implements IDatasetsBuilder {
+  #parsedPath?: Immutable<RosPath>;
```
jtbandes, line 35: seems like this is the X path - would be helpful to name it as such

```suggestion
  #parsedXPath?: Immutable<RosPath>;
```

packages/studio-base/src/panels/Plot/builders/CurrentCustomDatasetsBuilder.ts
```diff
@@ -0,0 +1,227 @@
...
+        if (msgEvent.topic !== series.parsed.topicName) {
+          continue;
+        }
+
+        const items = simpleGetMessagePathDataItems(msgEvent, series.parsed);
+        const pathItems = items.map((item, idx) => {
+          const chartValue = getChartValue(item);
+
+          return {
+            x: this.#xValues[idx] ?? NaN,
```
jtbandes, line 96: I am not too familiar with the existing "custom x" code paths but how do we know that `xValues[idx]` is the right corresponding x value? What if the number of messages on the X topic was different from the number of messages on the Y topic?

packages/studio-base/src/panels/Plot/HoverValue.tsx
```diff
@@ -0,0 +1,40 @@
...
+  enabled: boolean;
+};
+
+/**
+ * Apply the hover value to the chart.
+ *
+ * Since the hover value can update frequently on hover we use this component to contain the scope
+ * of what will re-render when the value updates.
+ */
+export function HoverValue(props: Props): JSX.Element {
```
jtbandes, line 22: maybe call it an "Adapter" since it follows our "adapter" pattern of not rendering anything?

packages/studio-base/src/components/MessagePipeline/index.tsx
```diff
@@ -67,6 +67,21 @@ export function useMessagePipeline<T>(selector: (arg0: MessagePipelineContext) =
   );
 }
 
+export function useMessagePipelineSubscribe(): (
```
jtbandes, line 70: wow - a new era in hacking around the original design of the panel APIs 😂😭

packages/studio-base/src/panels/Plot/builders/CustomDatasetsBuilder.ts
```diff
@@ -0,0 +1,324 @@
...
+  color: string;
+  timestampMethod: TimestampMethod;
+  showLine: boolean;
+  lineSize: number;
+  enabled: boolean;
+  blockCursor: BlockTopicCursor;
+};
+
+export class CustomDatasetsBuilder implements IDatasetsBuilder {
+  #parsedPath?: Immutable<RosPath>;
```
jtbandes, line 45:
```suggestion
  #parsedXPath?: Immutable<RosPath>;
```

packages/studio-base/src/panels/Plot/builders/CustomDatasetsBuilder.ts
```diff
@@ -0,0 +1,324 @@
...
+  parsed: RosPath;
+  color: string;
+  timestampMethod: TimestampMethod;
+  showLine: boolean;
+  lineSize: number;
+  enabled: boolean;
+  blockCursor: BlockTopicCursor;
+};
+
+export class CustomDatasetsBuilder implements IDatasetsBuilder {
```
jtbandes, line 44: maybe "CustomX" rather than just "Custom" to be more clear about what is custom?

packages/studio-base/src/panels/Plot/internalTypes.ts
```diff
@@ -98,3 +73,10 @@ export type PlotParams = {
   minYValue: string | number | undefined;
   maxYValue: string | number | undefined;
 };
+
+/**
+ * @returns true if the series config is a reference line
+ */
+export function isReferenceLinePlotPathType(path: Immutable<PlotPath>): boolean {
+  return !isNaN(+path.value);
```
jtbandes, line 81: why change the implementation of this? I'm not sure offhand what the effect will be but maybe we should leave it alone to avoid inviting new demons?

packages/studio-base/src/panels/Plot/subscription.ts
```diff
@@ -0,0 +1,46 @@
...
+import type { Immutable } from "@foxglove/studio";
+import type {
+  MessagePathPart,
+  RosPath,
+} from "@foxglove/studio-base/components/MessagePathSyntax/constants";
+import type { SubscribePayload } from "@foxglove/studio-base/players/types";
+
+const typeIsName = (part: Immutable<MessagePathPart>) => part.type === "name";
+
+export function pathToPayload(path: Immutable<RosPath>): SubscribePayload | undefined {
```
jtbandes, line 14:
```suggestion
export function pathToSubscribePayload(path: Immutable<RosPath>): SubscribePayload | undefined {
```

maybe rename the file to match?

packages/studio-base/src/panels/Plot/Plot.tsx
```diff
@@ -0,0 +1,666 @@
...
+        pipelineSetSubscriptions,
+      [],
+    ),
+  );
+  const subscribeMessasagePipeline = useMessagePipelineSubscribe();
+
+  const { globalVariables } = useGlobalVariables();
+
+  useEffect(() => {
+    coordinator?.handleConfig(config, globalVariables);
```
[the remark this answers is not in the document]
defunctzombie: Separate which two? the config and global variables? The global variables are needed to create the "filled in paths". I could do that here as a memo at the react layer and depend only on the config.paths + global variables. Though that doesn't really change the nature of the update.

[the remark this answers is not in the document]
jtbandes: 👍 I missed that dependency and thought this was just updating two things at once.

packages/studio-base/src/panels/Plot/ChartRenderer.worker.ts
```diff
@@ -0,0 +1,77 @@
...
+
+import { ChartRenderer } from "./ChartRenderer";
+
+type InitArgs = {
+  canvas: OffscreenCanvas;
+  devicePixelRatio: number;
+};
+
+export type Service<T> = {
+  init(args: InitArgs): Promise<T>;
```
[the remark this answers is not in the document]
defunctzombie: The generic is so the `Service` type can be used for Comlink `expose` and `wrap` since they need to accept different types.

In PlotCoordinator we provide this type to wrap: `Comlink.wrap<Service<Comlink.RemoteObject<ChartRenderer>>>` and lower in this file we expose using `satisfies Service<Comlink.LocalObject<ChartRenderer>>`

packages/studio-base/src/panels/Plot/PlotCoordinator.ts
```diff
@@ -0,0 +1,299 @@
...
+    // twice.
+    this.#renderer = remote.init(
+      Comlink.transfer(
+        {
+          canvas: this.#canvas,
+          devicePixelRatio: window.devicePixelRatio,
+        },
+        [this.#canvas],
+      ),
+    );
```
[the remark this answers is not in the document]
defunctzombie: We need to provide the offscreen canvas before we initialize chartjs - this could be done as an update/call after but there is still some sort of "initialization" that needs to happen.

packages/studio-base/src/panels/Plot/Plot.tsx
```diff
@@ -0,0 +1,666 @@
...
+    if (!canvasDiv || !datasetsBuilder) {
+      return;
+    }
+    const canvas = document.createElement("canvas");
+    canvas.style.width = "100%";
+    canvas.style.height = "100%";
+    canvasDiv.appendChild(canvas);
+
+    if (typeof canvas.transferControlToOffscreen !== "function") {
+      throw new Error("Offscreen rendering is not supported");
```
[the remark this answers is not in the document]
defunctzombie: MDN shows this as supported in all the browsers: https://developer.mozilla.org/en-US/docs/Web/API/OffscreenCanvas

I would vote we say it is a requirement and update `canRenderApp`. If we do not want to make it a requirement then we can support it with the current ChartRenderer.ts since that is designed to work on the main thread as well.

packages/studio-base/src/panels/Plot/Plot.tsx
```diff
@@ -0,0 +1,666 @@
...
+  // Migrate legacy Plot-specific title setting to new global title setting
+  // https://github.com/foxglove/studio/pull/5225
+  useEffect(() => {
+    if (legacyTitle && (customTitle == undefined || customTitle === "")) {
+      saveConfig({
+        title: undefined,
+        [PANEL_TITLE_CONFIG_KEY]: legacyTitle,
+      } as Partial<PlotConfig>);
+    }
+  }, [customTitle, legacyTitle, saveConfig]);
```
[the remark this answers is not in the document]
defunctzombie: Sure - I had wanted to avoid making behavior changes but I would agree that this can go since its been a year.

packages/studio-base/src/panels/Plot/ChartRenderer.worker.ts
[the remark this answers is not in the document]
jtbandes: Is `Service<Remote<ChartRenderer>>` different from `Remote<Service<ChartRenderer>>`? I think the ChartRenderer type should be baked into the Service definition because realistically you do not change what init() returns. Also I think something is wrong / not ideal with how the comlink types are being used. The docs for `Local<T>` say "It takes a `Remote<T>` and returns its original input `T`". ChartRenderer is already the "original" object so it should not need to be passed to Local.

packages/studio-base/src/components/MessagePipeline/index.tsx
defunctzombie: The one true way

packages/studio-base/src/panels/Plot/builders/TimeseriesDatasetsBuilderImpl.ts
```diff
@@ -0,0 +1,392 @@
...
+  items: DataItem[];
+};
+
+export type UpdateDataAction =
+  | ResetSeriesFullAction
+  | ResetSeriesCurrentAction
+  | UpdateSeriesCurrentAction
+  | UpdateSeriesFullAction;
+
+const MAX_CURRENT_DATUMS = 50_000;
```
snosenzo, line 90: mention per series?

packages/studio-base/src/panels/Plot/ChartRenderer.worker.ts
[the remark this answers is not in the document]
defunctzombie: Spoke in slack and we determined that this needs to remain how it is because the `Remote<T>` comlink generic does not apply _deeply_

packages/studio-base/src/panels/Plot/PlotCoordinator.ts
```diff
@@ -0,0 +1,299 @@
...
+  public setTimeseriesBounds(bounds: Immutable<Partial<Bounds1D>>): void {
+    this.#timeseriesBounds = bounds;
+    this.#pendingRenderActions.push({
+      type: "range",
+      bounds: {
+        min: bounds.min ?? this.#datasetBounds.min,
+        max: bounds.max ?? this.#datasetBounds.max,
+      },
+    });
+    this.#queueDispatchRender();
```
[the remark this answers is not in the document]
defunctzombie: pendingRenderActions is no more

packages/studio-base/src/panels/Plot/builders/CurrentCustomDatasetsBuilder.ts
defunctzombie: changed to a function that finds the last item without making a closure `lastMatchingTopic`

packages/studio-base/src/panels/Plot/builders/CurrentCustomDatasetsBuilder.ts
defunctzombie: Same as above comment

packages/studio-base/src/panels/Plot/builders/CurrentCustomDatasetsBuilder.ts
defunctzombie: The number of messages is expected to be the same. We don't do any logic to check that `xValues[idx]` is the same original message. I do think it could be more robust to do that - we would build up an array of arrays so in the case where your message path itself is an array you still find the same message. Right now its assumed the user "knows what they are doing" and need to pick array values that line up. I would put this as future work if we ever learn that it is confusing to users.

packages/studio-base/src/panels/Plot/builders/CustomDatasetsBuilder.ts
defunctzombie: I opted to make these match the xAxisVal (mode) names.

packages/studio-base/src/panels/Plot/builders/CustomDatasetsBuilder.ts
defunctzombie: The answer is to move to separate function. One of the follow-up items is to split current frame and blocks processing into two functions so that when you split a panel and blocks are already loaded we can show the user rendering progress more iteratively.

packages/studio-base/src/panels/Plot/builders/CustomDatasetsBuilder.ts
defunctzombie: done

packages/studio-base/src/panels/Plot/PlotLegend.tsx
```diff
@@ -213,6 +213,42 @@ function PlotLegendComponent(props: Props): JSX.Element {
...
+
+  const [currentValuesBySeriesIndex, setCurrentValuesBySeriesIndex] = useState<
+    unknown[] | undefined
+  >();
+  useEffect(() => {
+    if (!coordinator || !showValues) {
+      return;
+    }
+    const handler = (values: readonly unknown[]) => {
+      setCurrentValuesBySeriesIndex([...values]);
```
achim-k, line 240: Probably not an issue here, but using the spread operator can lead to stack size exceeded errors when `values` would be large enough

packages/studio-base/src/panels/Plot/builders/TimeseriesDatasetsBuilderImpl.ts
defunctzombie: we can say that since this is on the receiving side of the post-message - though this function is currently accepting an Immutable action and we'd need to change that.

packages/studio-base/src/panels/Plot/PlotLegend.tsx
defunctzombie: 👍 changed to slice

packages/studio-base/src/panels/Plot/Plot.tsx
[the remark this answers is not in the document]
defunctzombie: This has been removed.

packages/studio-base/src/panels/Plot/Plot.tsx
```diff
@@ -0,0 +1,666 @@
...
+  const subscribeMessasagePipeline = useMessagePipelineSubscribe();
+
+  const { globalVariables } = useGlobalVariables();
+
+  useEffect(() => {
+    coordinator?.handleConfig(config, globalVariables);
+  }, [coordinator, config, globalVariables]);
+
+  const datasetsBuilder = useMemo(() => {
+    switch (xAxisVal) {
```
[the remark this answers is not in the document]
defunctzombie: renamed

packages/studio-base/src/panels/Plot/Plot.tsx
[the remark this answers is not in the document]
defunctzombie: updated `canRenderApp` to require offscreen canvas support

packages/studio-base/src/panels/Plot/internalTypes.ts
defunctzombie: reverted to the old implementation

packages/studio-base/src/panels/Plot/Plot.tsx
```diff
@@ -0,0 +1,666 @@
...
+    const unsub = subscribeMessasagePipeline((state) => {
+      plotCoordinator.handlePlayerState(state.playerState);
+    });
+
+    // Subscribing only gets us _new_ updates, so we feed the latest state into the chart
+    plotCoordinator.handlePlayerState(getMessagePipelineState().playerState);
+
+    const resizeObserver = new ResizeObserver((entries) => {
+      for (const entry of entries) {
+        if (entry.target !== canvasDiv) {
```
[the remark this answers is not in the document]
defunctzombie: updated to use findLast from lodash - our ts config target settings don't have `findLast` on arrays yet

packages/studio-base/src/panels/Plot/Plot.tsx
```diff
@@ -0,0 +1,666 @@
...
+    }
+  }, [datasetsBuilder, globalVariables, xAxisPath]);
+
+  useEffect(() => {
+    if (!canvasDiv || !datasetsBuilder) {
+      return;
+    }
+    const canvas = document.createElement("canvas");
+    canvas.style.width = "100%";
+    canvas.style.height = "100%";
```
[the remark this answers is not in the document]
defunctzombie: done

packages/studio-base/src/panels/Plot/Plot.tsx
```diff
@@ -0,0 +1,666 @@
...
+    return () => {
+      unsub();
+      resizeObserver.disconnect();
+      plotCoordinator.destroy();
+      canvasDiv.removeChild(canvas);
+    };
+  }, [canvasDiv, datasetsBuilder, getMessagePipelineState, subscribeMessasagePipeline]);
+
+  const onWheel = useCallback(
+    async (event: React.WheelEvent<HTMLElement>) => {
```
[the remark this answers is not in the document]
defunctzombie: removed

packages/studio-base/src/panels/Plot/Plot.tsx
```diff
@@ -0,0 +1,666 @@
...
+    [coordinator],
+  );
+
+  const [buildTooltip, setBuildTooltip] = useState<
+    ((args: ElementAtPixelArgs) => void) | undefined
+  >(undefined);
+
+  const mousePresentRef = useRef(false);
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
defunctzombie: updated to useMemo and renamed `moveHandler` - much nicer now

packages/studio-base/src/panels/Plot/PlotCoordinator.ts
```diff
@@ -0,0 +1,299 @@
...
+ */
+export class PlotCoordinator extends EventEmitter<EventTypes> {
+  #renderingWorker: Worker;
+  #canvas: OffscreenCanvas;
+  #renderer?: Promise<Comlink.RemoteObject<ChartRenderer>>;
+
+  #datasetsBuilder: IDatasetsBuilder;
+
+  #timeseriesBounds?: Immutable<Partial<Bounds1D>>;
+  #datasetBounds: Bounds1D = { min: 0, max: 1 };
```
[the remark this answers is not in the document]
defunctzombie: renamed to global bounds

packages/studio-base/src/panels/Plot/PlotCoordinator.ts
```diff
@@ -0,0 +1,299 @@
...
+      type: "range",
+      bounds: {
+        min: bounds.min ?? this.#datasetBounds.min,
+        max: bounds.max ?? this.#datasetBounds.max,
+      },
+    });
+    this.#queueDispatchRender();
+  }
+
+  public resetBounds(): void {
```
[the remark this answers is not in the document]
defunctzombie: `resetBounds` resets all the bounds (including interaction). `setGlobalBounds` sets the global bounds _and_ resets the interaction bounds. It has a specific behavior around how global bounds are handled by the coordinator.

packages/studio-base/src/panels/Plot/IDatasetsBuilder.ts
```diff
@@ -0,0 +1,49 @@
...
+  };
+  size: Size;
+};
+
+export type CsvDataset = {
+  label: string;
+  data: CsvDatum[];
+};
+
+interface IDatasetsBuilder {
```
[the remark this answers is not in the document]
defunctzombie: added a comment

packages/studio-base/src/panels/Plot/IDatasetsBuilder.ts
```diff
@@ -0,0 +1,49 @@
...
+  handlePlayerState(state: Immutable<PlayerState>): Bounds1D | undefined;
```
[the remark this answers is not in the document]
defunctzombie: I want to split this into current and blocks functions - that would be a good time to also simplify the input requirements. 👍

packages/studio-base/src/panels/Plot/IndexDatasetsBuilder.ts
```diff
@@ -0,0 +1,181 @@
...
+      if (isReferenceLinePlotPathType(path)) {
+        continue;
+      }
+
+      const parsed = parseRosPath(path.value);
+      if (!parsed) {
+        continue;
+      }
+
+      const filledParsed = fillInGlobalVariablesInPath(parsed, globalVariables);
```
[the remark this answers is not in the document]
defunctzombie: You need to fill in the global variables to then lookup the existing series. This is important because if global variables change the path then we need to consider that a new series.

snosenzo (approved):

packages/studio-base/src/panels/Plot/builders/TimeseriesDatasetsBuilder.ts
```diff
@@ -0,0 +1,264 @@
...
+
+      // fixme - filter data dispatch to remove series that are reset
+      // and series that are no longer present
+      // we don't need to send
+      // go backwards from end to start
+
+      await this.#datasetsBuilderRemote.updateData(dispatch);
+    }
+
+    return await this.#datasetsBuilderRemote.getViewportDatasets(viewport);
```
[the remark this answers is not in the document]
defunctzombie: Which are always together?

packages/studio-base/src/panels/Plot/Plot.tsx
```diff
@@ -0,0 +1,743 @@
...
+
+    // Subscribing only gets us _new_ updates, so we feed the latest state into the chart
+    coordinator.handlePlayerState(getMessagePipelineState().playerState);
+    return unsub;
+  }, [coordinator, getMessagePipelineState, subscribeMessasagePipeline]);
+
+  const datasetsBuilder = useMemo(() => {
+    switch (xAxisMode) {
+      case "timestamp":
+        return new TimeseriesDatasetsBuilder();
```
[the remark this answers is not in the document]
defunctzombie: 👍 Will do in this PR - the names are supposed to match the modes

packages/studio-base/src/panels/Plot/builders/CurrentCustomDatasetsBuilder.ts
```diff
@@ -0,0 +1,232 @@
...
+    if (!activeData || !this.#xParsedPath) {
+      return;
+    }
+
+    const msgEvents = activeData.messages;
+    if (msgEvents.length === 0) {
+      return;
+    }
+
+    {
```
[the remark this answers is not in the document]
defunctzombie: So the `msgEvent` variable is localized since I make another one below and lint complained about variable shadowing

packages/studio-base/src/panels/Plot/builders/CurrentCustomDatasetsBuilder.ts
```diff
@@ -0,0 +1,232 @@
...
+};
+
+type SeriesItem = {
+  enabled: boolean;
+  messagePath: string;
+  parsed: RosPath;
+  dataset: ChartDataset<"scatter", DatumWithReceiveTime[]>;
+};
+
+export class CurrentCustomDatasetsBuilder implements IDatasetsBuilder {
```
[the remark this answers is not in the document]
defunctzombie: The names are meant to match the modes. I will add a comment to the class describing what it does.

packages/studio-base/src/panels/Plot/builders/BlockTopicCursor.test.ts
```diff
@@ -0,0 +1,199 @@
...
+
+    const blocks = [block0, block1];
+
+    expect(cursor.nextWillReset(blocks)).toEqual(true);
+    expect(cursor.next(blocks)).toEqual(block0["messagesByTopic"]["/foo"]);
+    expect(cursor.next(blocks)).toEqual(block1["messagesByTopic"]["/foo"]);
+    expect(cursor.next(blocks)).toEqual(undefined);
+  });
+
+  it("should continue from last location when new blocks arrive", () => {
```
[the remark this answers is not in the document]
defunctzombie: Why's that? Do we not like that pattern? It's pretty typical language use for the describe/it API style. `it` _should_. I don't feel strongly about it though.

packages/studio-base/src/panels/Plot/builders/BlockTopicCursor.test.ts
```diff
@@ -0,0 +1,199 @@
...
+    };
+
+    expect(cursor.nextWillReset([block0])).toEqual(true);
+    expect(cursor.next([block0])).toEqual(block0["messagesByTopic"]["/foo"]);
+    expect(cursor.next([block0])).toEqual(undefined);
+    expect(cursor.next([block0, block1])).toEqual(block1["messagesByTopic"]["/foo"]);
+    expect(cursor.next([block0, block1])).toEqual(undefined);
+  });
+
+  it("should restart if either the first or last block of the topic change", () => {
```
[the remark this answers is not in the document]
defunctzombie: Yep - will change to `last iterated`

packages/studio-base/src/panels/Plot/HoverValue.tsx
defunctzombie: This is now `VerticalBars` component

packages/studio-base/src/panels/Plot/builders/TimeseriesDatasetsBuilder.ts
```diff
@@ -0,0 +1,264 @@
...
+  public async getCsvData(): Promise<CsvDataset[]> {
+    return await this.#datasetsBuilderRemote.getCsvData();
+  }
+
+  public destroy(): void {
+    this.#datasetsBuilderWorker.terminate();
+  }
+}
+
+function readMessagePathItems(
```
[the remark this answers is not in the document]
defunctzombie: opted to test at the outer layer (the class itself) for now

packages/studio-base/src/panels/Plot/builders/TimeseriesDatasetsBuilder.ts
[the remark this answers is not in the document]
jtbandes: `updateData` & `getViewportDatasets` – neither of these seem to happen anywhere else, so we could save on the number of postMessages by combining them.

packages/studio-base/src/panels/Plot/builders/BlockTopicCursor.test.ts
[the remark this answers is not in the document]
jtbandes: yeah, I've seen both "it should work" and "it works" styles of writing test names – I don't think we're consistent today in our code

packages/studio-base/src/panels/Plot/Plot.tsx
```diff
@@ -0,0 +1,732 @@
...
+        throw new Error(`unsupported mode: ${xAxisMode}`);
+    }
+
+    return undefined;
+  }, [xAxisMode]);
+
+  useEffect(() => {
+    if (
+      datasetsBuilder instanceof CurrentCustomDatasetsBuilder ||
+      datasetsBuilder instanceof CustomDatasetsBuilder
```
jtbandes, line 277: maybe Plot.tsx shouldn't have to care too much about specific builder classes?

packages/studio-base/src/panels/Plot/PlotCoordinator.ts
```diff
@@ -0,0 +1,331 @@
...
+    });
+    this.#currentValues = [];
+    this.emit("currentValuesChanged", this.#currentValues);
+
+    this.#updateAction.showXAxisLabels = config.showXAxisLabels;
+    this.#updateAction.showYAxisLabels = config.showYAxisLabels;
+    this.#updateAction.referenceLines = referenceLines;
+
+    // Config changes to yBounds always takes precedence over user interaction changes like pan/zoom
+    this.#updateAction.yBounds = this.#configBounds.y;
```
jtbandes, line 198: Currently this resets y bounds whenever _any_ config change is made, even if you are not changing y bounds

packages/studio-base/src/components/MessagePathSyntax/stringifyRosPath.ts
```diff
@@ -0,0 +1,79 @@
...
+    }
+    return String(slicePart);
+  }
+
+  return `$${slicePart.variableName}`;
+}
+
+function filterToString(filter: Immutable<MessagePathFilter>): string {
+  if (typeof filter.value === "object") {
+    const start = filter.path[0] ?? "";
```
jtbandes, line 74: needs fix - working on this

jtbandes (approved):
📿

packages/studio-base/src/panels/Plot/Plot.tsx
```diff
@@ -0,0 +1,666 @@
...
+  const onResetView = useCallback(() => {
+    setShowReset(false);
+    coordinator?.resetBounds();
+
+    if (shouldSync) {
+      setGlobalBounds(undefined);
+    }
+  }, [coordinator, setGlobalBounds, shouldSync]);
+
+  const valuesBySeriesIndex = useMemo(() => {
```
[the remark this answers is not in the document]
defunctzombie: I think all of OffscreenCanvas setup will move but I will tackle it when we do state transitions

packages/studio-base/src/panels/Plot/Plot.tsx
```diff
@@ -0,0 +1,666 @@
...
+    return () => {
+      hammerManager.destroy();
+    };
+  }, [canvasDiv, coordinator]);
+
+  // We could subscribe in the chart renderer, but doing it with react effects is easier for
+  // managing the lifecycle of the subscriptions. The renderer will correlate input message data to
+  // the correct series.
+  useEffect(() => {
+    const subscriptions = filterMap(series, (item): SubscribePayload | undefined => {
```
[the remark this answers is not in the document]
defunctzombie: added to cleanup items

packages/studio-base/src/panels/Plot/builders/TimeseriesDatasetsBuilder.ts
```diff
@@ -0,0 +1,264 @@
...
+        color: getLineColor(path.color, idx),
+        lineSize: path.lineSize ?? 1.0,
+        timestampMethod: path.timestampMethod,
+        showLine: path.showLine ?? true,
+        enabled: path.enabled,
+        blockCursor: existing?.blockCursor ?? new BlockTopicCursor(parsed.topicName),
+      };
+    });
+
+    void this.#datasetsBuilderRemote.setConfig(this.#seriesConfigs);
```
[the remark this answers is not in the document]
defunctzombie: Added to follow-up

packages/studio-base/src/panels/Plot/builders/TimeseriesDatasetsBuilder.ts
```diff
@@ -0,0 +1,264 @@
...
+    if (!activeData) {
+      return;
+    }
+
+    const didSeek = activeData.lastSeekTime !== this.#lastSeekTime;
+    this.#lastSeekTime = activeData.lastSeekTime;
+
+    const msgEvents = activeData.messages;
+    if (msgEvents.length > 0) {
+      for (const seriesConfig of this.#seriesConfigs) {
```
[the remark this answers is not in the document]
defunctzombie: 📿

packages/studio-base/src/panels/Plot/builders/CustomDatasetsBuilder.ts
defunctzombie: done!

packages/studio-base/src/panels/Plot/PlotCoordinator.ts
defunctzombie: fixed

packages/studio-base/src/panels/Plot/Plot.tsx
defunctzombie: Probably - then all of them would need a setXPath and some would ignore or throw

packages/studio-base/src/panels/Plot/builders/CurrentCustomDatasetsBuilder.ts
```diff
@@ -0,0 +1,232 @@
...
+  #xParsedPath?: Immutable<RosPath>;
+
+  #xValues: number[] = [];
+
+  #seriesByMessagePath = new Map<string, SeriesItem>();
+  #pathsWithMismatchedDataLengths = new Set<string>();
+
+  public handlePlayerState(state: Immutable<PlayerState>): Bounds1D | undefined {
```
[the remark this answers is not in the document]
defunctzombie: added a comment

PR #7298, opened 2024-01-02
Prune current plot data

achim-k:
**User-Facing Changes**
Fix memory leak in plot panel

**Description**
Prunes current plot data if the size of accumulated messages exceeds 20 MB.

### DIFF
--- a/packages/studio-base/src/panels/Plot/clients.ts
+++ b/packages/studio-base/src/panels/Plot/clients.ts
@@ -41,14 +41,20 @@ export type Client = {
 export type DatasetsState = {
   clients: Record<string, Client>;
   blocks: readonly MessageBlock[];
-  current: MessageEvent[];
+  current: {
+    messages: MessageEvent[];
+    sizeInBytes: number;
+  };
 };
 
 export function initDatasets(): DatasetsState {
   return {
     clients: {},
     blocks: [],
-    current: [],
+    current: {
+      messages: [],
+      sizeInBytes: 0,
+    },
   };
 }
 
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/panels/Plot/clients.ts
```diff
@@ -211,7 +217,34 @@ export function updateCurrent(
...
+
+/**
+ * Removes older current messages so that current data does not exceed the given upper byte size limit.
+ * Ensures that at least one message remains.
+ */
+export function pruneCurrent(state: DatasetsState, upperByteSizeLimit: number): DatasetsState {
+  const { current: oldCurrent } = state;
+  const minNumBytesToBeFreed = oldCurrent.sizeInBytes - upperByteSizeLimit;
+  let numBytesToBeFreed = 0;
+  let idx = 0;
```
defunctzombie, line 235: You could also use a for loop here:

```(for let idx=0; idx < oldCurrent.messages.length - 1 && numBytesToBeFreed < minNumBytesToBeFreed; ++idx) { ...```

Did you mean to use `oldCurrent.messages.length - 1` rather than `oldCurrent.messages.length`? You always want to leave at least one old current message?

defunctzombie (approved):

packages/studio-base/src/panels/Plot/clients.ts
achim-k: Yes, at least one message should remain. I'm uncertain what will happen if all messages are pruned

achim-k: About the for loop approach: Wouldn't `idx`, which I use after the loop, be out of scope then?

defunctzombie: Yea - missed that bit. The loop is to calculate idx for the later slice. 👍

PR #7299, opened 2024-01-03
Bug fixes to ColorPickerInput

2metres:
**User-Facing Changes**
Input now can recieve and validate color strings + minor styling pass to bring into alignment with our UI

https://github.com/foxglove/studio/assets/924528/9746662b-53c1-4568-948f-19e429f8ea90

**Description**
Previously the input displayed in the sidebar was readonly and only clicking the swatch would let you change the color. Now the inputs and UI picker are all connected

Resolves FG-4534

### DIFF
--- a/packages/studio-base/src/components/SettingsTreeEditor/inputs/ColorPickerInput.tsx
+++ b/packages/studio-base/src/components/SettingsTreeEditor/inputs/ColorPickerInput.tsx
@@ -108,7 +108,7 @@ export function ColorPickerInput(props: ColorPickerInputProps): JSX.Element {
         placeholder={props.placeholder}
         size="small"
         variant="filled"
-        value={`#${editedValue.replace("#", "")}`}
+        value={editedValue ? `#${editedValue.replace("#", "")}` : editedValue}
         onKeyDown={(event) => event.key === "Enter" && handleClose}
         onChange={(event) => {
           updateEditedValue(event.target.value);
--- a/packages/studio-base/src/components/SettingsTreeEditor/inputs/ColorPickerInput.tsx
+++ b/packages/studio-base/src/components/SettingsTreeEditor/inputs/ColorPickerInput.tsx
@@ -3,14 +3,13 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
 import CancelIcon from "@mui/icons-material/Cancel";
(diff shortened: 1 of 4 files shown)


2metres:
🤔  Fix one thing break another

packages/studio-base/src/components/SettingsTreeEditor/inputs/ColorPickerInput.tsx
```diff
@@ -100,10 +107,18 @@ export function ColorPickerInput(props: ColorPickerInputProps): JSX.Element {
         disabled={disabled}
         placeholder={props.placeholder}
         size="small"
-        value={displayValue ?? ""}
         variant="filled"
+        value={`#${editedValue.replace("#", "")}`}
```
2metres, line 111: This is less than ideal. Could use help unwravelling the `useColorPickerControl` hook

packages/studio-base/src/components/SettingsTreeEditor/inputs/ColorPickerInput.tsx
```diff
@@ -100,10 +107,18 @@ export function ColorPickerInput(props: ColorPickerInputProps): JSX.Element {
         disabled={disabled}
         placeholder={props.placeholder}
         size="small"
-        value={displayValue ?? ""}
         variant="filled"
+        value={editedValue ? `#${editedValue.replace("#", "")}` : editedValue}
```
2metres, line 111: This is less than ideal. Could use help unwravelling the useColorPickerControl hook

snosenzo (approved):

packages/studio-base/src/components/SettingsTreeEditor/inputs/ColorPickerInput.tsx
2metres: So we need to return undefined when empty so using starts with doesn't work. I tried moving this into the reducer but i couldn't get it to work

bryfox: > we need to return undefined when empty

It looks like `editedValue` is typed as a string, so I don't understand this bit.

I see you're also now returning `displayValue` (which makes sense if the intent is to offload the hex display to the calling component); that sounds to me like the value you'd want to display as a text label?

PR #7300, opened 2024-01-03
Match panel overlay to panel menu

2metres:
**User-Facing Changes**
The panel actions overlay now has separate buttons to split right and split down, rather than a single split button that chooses a split direction automatically.

**Description**
Update PanelOverlay actions to match PanelDropdown with "Split right" and "Split down" instead of "Split panel"

Screen Shot 2024-01-03 at 2 18 31 pm


2metres:
@jtbandes in testing the panel does not respect the requested split direction can you take a look and help me understand why?

defunctzombie:
Visual changes LGTM. What's with the text behind the buttons?

2metres:
That is the panel empty state showing underneath the overlay

jtbandes (approved):

PR #7305, opened 2024-01-08
Better reuse of transforms and intermediate structures for transforms

snosenzo:
**User-Facing Changes**

- Better transform tree memory management and performance

**Description**

To keep new transforms from being created for every new transform past when transforms are evicted from their CoordinateFrame history buffers, I've implemented a TransformPool which will hold transforms that are no longer being used so that they can be used again by future transform messages. This cuts down garbage collection costs for cleaning up unused transforms. Also I removed some intermediate array declarations created when instantiating transforms.

As a cleanup Item I also reduced the complexity of transform history eviction but maintaining the true cap and cutting down by a quarter of the maximum when it goes over instead of the more complicated overfillPercentage method I used before.

To support the `release` for TransformPool I've updated the ArrayMap methods to return the items that were removed so that they can be released back to the pool.

FG-6240


packages/studio-base/src/panels/ThreeDeeRender/transforms/TransformPool.ts
```diff
@@ -0,0 +1,33 @@
...
+    return transform;
+  }
+
+  public release(transform: Transform): void {
+    if (!this.#isDisposed) {
+      this.#transforms.push(transform);
+    }
+  }
+
+  public dispose(): void {
```
achim-k, line 29: where is this called?

snosenzo: ah doesn't look like it is, will update

packages/den/collection/ArrayMap.ts
```diff
@@ -54,25 +62,27 @@ export class ArrayMap<K, V> {
   }
 
   /** Removes the element with the given key, if it exists */
```
defunctzombie, line 64: Update comment to say what this returns.

packages/den/collection/ArrayMap.ts
```diff
@@ -27,11 +31,14 @@ export class ArrayMap<K, V> {
   /**
    * Store a key/value tuple in the sorted list. If the key already exists, the
    * previous entry is overwritten.
+   * Returns previous entry if it exists
    */
-  public set(key: K, value: V): void {
+  public set(key: K, value: V): [K, V] | undefined {
```
defunctzombie, line 36: This feels a bit strange. I don't think I've encountered an API for Map that returns the previous entry when you call "set". Not sure I would even find that intuitive if I saw:

```
const prev = map.set("foo", "bar");
```

packages/den/collection/ArrayMap.ts
```diff
@@ -54,25 +62,27 @@ export class ArrayMap<K, V> {
...
-  public removeAfter(key: K): void {
+  public removeAfter(key: K): [K, V][] {
     const index = this.binarySearch(key);
     const greaterThanIndex = index >= 0 ? index + 1 : ~index;
-    this.#list.length = greaterThanIndex;
+    const removed = this.#list.splice(greaterThanIndex);
+    return removed;
   }
 
   /** Removes all elements with keys less than the given key. */
```
defunctzombie, line 81: Update comment to say this returns the removed elements.

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -1042,7 +1047,9 @@ export class Renderer extends EventEmitter<RendererEvents> implements IRenderer
     const t = translation;
     const q = rotation;
 
-    const transform = new Transform([t.x, t.y, t.z], [q.x, q.y, q.z, q.w]);
+    const transform = this.#transformPool.acquire();
+    transform.setPositionValues(t.x, t.y, t.z);
```
defunctzombie, line 1051: Maybe drop the `Values` part of the name?

```suggestion
    transform.setPosition(t.x, t.y, t.z);
    transform.setRotation(q.x, q.y, q.z, q.w);
```

packages/studio-base/src/panels/ThreeDeeRender/transforms/CoordinateFrame.ts
```diff
@@ -15,6 +16,7 @@ import { Duration, interpolate, percentOf, Time } from "./time";
 type TimeAndTransform = [time: Time, transform: Transform];
 
 export const MAX_DURATION: Duration = 4_294_967_295n * BigInt(1e9);
+export const MAX_CAPACITY_EVICT_AMOUNT = 0.25;
```
defunctzombie, line 19: Is this an amount or percent?

packages/studio-base/src/panels/ThreeDeeRender/transforms/CoordinateFrame.ts
```diff
@@ -118,6 +118,9 @@ export class CoordinateFrame<ID extends AnyFrameId = UserFrameId> {
    */
   public setParent(parent: CoordinateFrame): void {
     if (this.#parent && this.#parent !== parent) {
+      for (const [__, tf] of this.#transforms) {
+        this.#transformPool.release(tf);
+      }
       this.#transforms.clear();
```
defunctzombie, line 124: Should clear return the removed items so you can release those back into the pool? If you forget to call clear on `#transforms` after releasing then you end up having released transforms. Any other way you can think of to avoid that?

I'd also considered whether a finalization registry could be useful but maybe that has more overhead than we want?

packages/den/collection/ArrayMap.ts
defunctzombie: I take that back. Seems you are in good company:

https://doc.rust-lang.org/std/collections/hash_map/struct.HashMap.html#method.insert

packages/studio-base/src/panels/ThreeDeeRender/transforms/Transform.ts
```diff
@@ -6,6 +6,9 @@ import { mat4, vec3, quat, ReadonlyMat4, ReadonlyVec3, ReadonlyQuat } from "gl-m
 
 import { Pose, getRotationNoScaling, mat4Identity, quatIdentity, vec3Identity } from "./geometry";
 
+const tempVec3: vec3 = [0, 0, 0];
```
defunctzombie, line 9: comment on purpose/use because this seems to be related to performance and is somewhat a-typical

packages/studio-base/src/panels/ThreeDeeRender/transforms/Transform.ts
```diff
@@ -51,12 +59,27 @@ export class Transform {
...
     mat4.fromRotationTranslation(this.#matrix, this.#rotation, this.#position);
     return this;
   }
 
+  public setRotationValues(x: number, y: number, z: number, w: number): this {
+    tempQuat[0] = x;
+    tempQuat[1] = y;
+    tempQuat[2] = z;
+    tempQuat[3] = w;
+    return this.setRotation(tempQuat);
```
defunctzombie, line 80: Do you not need to normalize like the above `setRotation` does?

packages/studio-base/src/panels/ThreeDeeRender/transforms/Transform.ts
```diff
@@ -51,12 +59,27 @@ export class Transform {
     return this;
   }
 
+  public setPositionValues(x: number, y: number, z: number): this {
```
defunctzombie, line 62: Earlier - I had mentioned calling this `setPosition` but I see now there is still a `setPosition`. Maybe a good name is `updatePosition`? `setPositionValues` is ok too if that feels natural to you. I do think a comment would help clarify how this method differs from `setPosition`. It seems to still call `setPosition` so I am not sure why the caller wouldn't do that themselves.

packages/studio-base/src/panels/ThreeDeeRender/transforms/TransformPool.ts
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * An object pool for Transforms
+ */
+export class TransformPool {
+  #transforms: Transform[] = [];
+  #isDisposed = false;
+
+  public acquire(): Transform {
+    let transform = this.#transforms.pop();
```
defunctzombie, line 15: `return this.#transforms.pop() ?? new Transform()`

Also add a comment to the function making it clear that if the pool is empty it will allocate a new transforms. Some pools will not allocate beyond some capacity. Maybe you want to also consider a capacity so when you release transforms you don't continue to store 100k transforms in the pool but limit it to some number. You can still allocate new objects but releasing them will let them be garbage collected instead of put back into the pool.

If you feel empowered - there's nothing specific to "Transforms" in this pool code and you could make a generic ObjectPool<T> in foxglove/den/collections :D

packages/studio-base/src/panels/ThreeDeeRender/transforms/TransformPool.ts
```diff
@@ -0,0 +1,33 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
defunctzombie, line 1: Some tests for TransformPool? (Or ObjectPool if you feel empowered to make a nice re-usable pool for us all)

packages/studio-base/src/panels/ThreeDeeRender/transforms/TransformTree.ts
```diff
@@ -26,19 +28,24 @@ export class TransformTree {
   #frames = new Map<string, CoordinateFrame>();
   #maxStorageTime: Duration;
   #maxCapacityPerFrame: number;
+  #transformPool: TransformPool;
+
   public defaultRootFrame: CoordinateFrame<FallbackFrameId>;
 
   public constructor(
+    transformPool: TransformPool = new TransformPool(),
```
defunctzombie, line 36: What's the purpose of `new TransformPool()` here? You have to specify this argument so the caller will always provide a pool for you.

defunctzombie:
> This cuts down garbage collection costs for cleaning up unused transforms.

Do you expect these transforms to get used again? So the number of unused transforms will typically be small and steady state? Something to consider might be a cap on the pool size so if there is some logic that allocates 100k transforms from the pool, releases them back, and never has to use more than 1k transforms you don't end up hanging on to 100k objects in the pool forever.

packages/studio-base/src/panels/ThreeDeeRender/transforms/CoordinateFrame.ts
```diff
@@ -141,45 +144,56 @@ export class CoordinateFrame<ID extends AnyFrameId = UserFrameId> {
...
       const startTime = endTime - this.maxStorageTime;
-      // at the very least we remove the overfill, but if the maxStorageTime enforces a  larger trim we take that
-      // we can't afford to check maxStorageTime every time we add a transform, so we only check it when overfill is full
+      // at the very least we trim a quarter, but if the maxStorageTime enforces a larger trim we take that
+      // we can't afford to check maxStorageTime every time we add a transform, so we only check it when capacity is full
       removeBeforeTime = startTime > removeBeforeTime ? startTime : removeBeforeTime;
 
-      this.#transforms.removeBefore(removeBeforeTime);
+      const entriesRemoved = this.#transforms.removeBefore(removeBeforeTime);
+      for (const [__, tf] of entriesRemoved) {
```
jtbandes, line 177:
```suggestion
      for (const [, tf] of entriesRemoved) {
```

packages/studio-base/src/panels/ThreeDeeRender/transforms/CoordinateFrame.ts
```diff
@@ -141,45 +144,56 @@ export class CoordinateFrame<ID extends AnyFrameId = UserFrameId> {
...
+        this.#transformPool.release(tf);
+      }
     }
   }
 
   /** Remove all transforms with timestamps greater than the given timestamp. */
   public removeTransformsAfter(time: Time): void {
-    this.#transforms.removeAfter(time);
+    const removed = this.#transforms.removeAfter(time);
+    for (const [__, tf] of removed) {
```
jtbandes, line 186:
```suggestion
    for (const [, tf] of removed) {
```

packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
```diff
@@ -811,3 +813,7 @@ export function ThreeDeeRender(props: {
     </ThemeProvider>
   );
 }
+
+function callFn(done: () => void): React.SetStateAction<(() => void) | undefined> {
```
jtbandes, line 817: maybe call this `wrapFn` or `makeGetter` or something like that?

If this really is a memory leak then we'll need to do this in several other panels too.

jtbandes (approved):

packages/studio-base/src/panels/ThreeDeeRender/ThreeDeeRender.tsx
snosenzo: Gonna remove, I don't think it's an issue

packages/studio-base/src/panels/ThreeDeeRender/transforms/CoordinateFrame.ts
snosenzo: Unfortunate about the extra array allocation, but I suppose it's fine since it's a relatively infrequent operation

packages/studio-base/src/panels/ThreeDeeRender/transforms/Transform.ts
snosenzo: not sure what you mean, setRotation does the normalization, which this function uses

packages/den/collection/ObjectPool.ts
```diff
@@ -0,0 +1,40 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+/**
+ * An object pool for reusing objects.
```
defunctzombie, line 6: Elaborate on this with how `maxCapacity` is used and any other notable behaviors you believe users of this need to understand to use it correctly.

packages/den/collection/ObjectPool.ts
```diff
@@ -0,0 +1,40 @@
...
+    if (!this.#isDisposed && this.#objects.length < this.#maxCapacity) {
+      this.#objects.push(obj);
+    }
+  }
+
+  /** Disables releasing items to the pool and clears the pool array.
+   * This should be called when the pool is no longer needed.
+   */
+  public dispose(): void {
+    this.#isDisposed = true;
```
defunctzombie, line 37: Why do you need to set this flag? Is it a problem to dispose/flush the pool and then start using it again? I don't see that `acquire` is gated on this flag so even after disposing it looks like you can keep getting new objects.

packages/den/collection/ObjectPool.ts
```diff
@@ -0,0 +1,40 @@
...
+    this.#objects.length = 0;
```
defunctzombie, line 38: I learned and have since forgotten this trick - I think it will allow for release of all the array items?

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -196,7 +197,10 @@ export class Renderer extends EventEmitter<RendererEvents> implements IRenderer
   #selectedRenderable: PickedRenderable | undefined;
   public colorScheme: "dark" | "light" = "light";
   public modelCache: ModelCache;
-  public transformTree = new TransformTree();
+
+  #transformPool = new ObjectPool(initTransform, 30000);
```
defunctzombie, line 201:
```suggestion
  #transformPool = new ObjectPool(initTransform, 30_000);
```

How did you pick 30k? Even if picked randomly add a comment so if we need to revisit this in the future we are not guessing how this number was picked. If there are some considerations at deciding on a "good" number for this note those as well.

packages/den/collection/ObjectPool.test.ts
```diff
@@ -0,0 +1,31 @@
...
+    const obj1 = { a: 1 };
+    objectPool.release(obj1);
+    const acq = objectPool.acquire();
+    // should have same object reference
+    expect(acq === obj1).toBe(true);
+  });
+  it("does not release past the maximum capacity", () => {
+    const list = [{ a: 1 }, { a: 1 }];
+    const objectPool = new ObjectPool(() => ({ a: 1 }), 1);
+    list.forEach((obj) => {
```
defunctzombie, line 24: nit: unless specific reason for a forEach callback (re-use, index value, etc) I find for loop easier to read and nice benefit of not making closures when we don't need them.

```suggestion
   for (const obj of list) {
```

packages/den/collection/ObjectPool.ts
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * An object pool for reusing objects.
+ */
+export class ObjectPool<T> {
+  #init: () => T;
+  #maxCapacity: number;
+  #objects: T[] = [];
+  #isDisposed = false;
+
+  public constructor(init: () => T, maxCapacity: number = 8000) {
```
defunctzombie, line 14: I would change the `maxCapacity` argument to an options object `options: { maxCapacity?: number }`. This gives us the ability to add future options easier and IMO makes the call-site easier to understand:

```
const pool = new ObjectPool(init, { maxCapacity: 30 });
```

defunctzombie: The default max capacity seems arbitrary to me.

defunctzombie (approved):
👍 because ObjectPool is something that others might want to rely on we need to be more clear with documenting it around assumptions, behavior, etc. Think of the things in _den_ like core/standard libraries and the level of documentation and notes you like to see on such libraries when you want to use them.

packages/den/collection/ObjectPool.ts
snosenzo: added jsdoc comment to constructor and mentioned here.

packages/den/collection/ObjectPool.ts
snosenzo: decided to remove `dispose` for now in lieu of a `clear` function that returns all of the elements in the pool for outside disposal. My only real reason for having the dispose method to begin with was to mirror the other `Pool` classes we have.

packages/den/collection/ObjectPool.ts
snosenzo: We use it in several places and it's done good by us so far

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
snosenzo: Added a comment about considerations and also made it a multiple of the max coordinate frame capacity.

packages/den/collection/ObjectPool.ts
snosenzo: removed default maxCapacity in favor of `undefined`.

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -198,7 +204,15 @@ export class Renderer extends EventEmitter<RendererEvents> implements IRenderer
   public colorScheme: "dark" | "light" = "light";
   public modelCache: ModelCache;
 
-  #transformPool = new ObjectPool(initTransform, 30000);
+  /**
+   * Max capacity should be chosen to be at least a several multiples of
```
defunctzombie, line 208:
```suggestion
   * Max capacity should be chosen to be at least several multiples of
```

packages/studio-base/src/panels/ThreeDeeRender/Renderer.ts
```diff
@@ -198,7 +204,15 @@ export class Renderer extends EventEmitter<RendererEvents> implements IRenderer
...
+   * the CoordinateFrame transform max capacity. So that it can store a
```
defunctzombie, line 209:
```suggestion
   * the CoordinateFrame transform max capacity. So that it can store
```

packages/studio-base/src/panels/ThreeDeeRender/transforms/TransformTree.ts
```diff
@@ -9,7 +9,7 @@ import { Transform } from "./Transform";
 import { Pose } from "./geometry";
 import { Duration, Time } from "./time";
 
-const DEFAULT_MAX_CAPACITY_PER_FRAME = 50_000;
+export const DEFAULT_MAX_CAPACITY_PER_FRAME = 50_000;
```
defunctzombie: Your other PR changes this value, right?

packages/studio-base/src/panels/ThreeDeeRender/transforms/Transform.ts
```diff
@@ -138,3 +170,7 @@ export class Transform {
     return out;
   }
 }
+
+export function initTransform(): Transform {
```
defunctzombie, line 174: Maybe a nicer way to expose this is a static on the Transform class rather than a free function?

https://www.typescriptlang.org/play?#code/MYGwhgzhAEBiD29oG8BQ0PQgFzNglsNAKIC2ADtgJ4AUAlCupswE4Cm2AriwHbQ9sA7nET0A3EwwBfVDNTB4PHNABmfALwj4AOjKUqEhUuzQw0TWvFA

defunctzombie (approved):

snosenzo:
gonna hold off on merging this, finding some small issues with transforms in further testing. looking into it now

snosenzo:
> gonna hold off on merging this, finding some issues with transforms in further testing. looking into it now

Fixed. The issue was that `set` on the `ArrayMap` was returning the referentially equal [K, V] tuple as the one that was in the `#list` causing a transform that was in use to be `released`. I've fixed this to always make a new entry and place a new entry so that these aren't entangled. I also added a test to prevent this mistake from being made if there are any changes to the util.

packages/den/collection/ArrayMap.ts
```diff
@@ -32,16 +32,16 @@ export class ArrayMap<K, V> {
   /**
    * Store a key/value tuple in the sorted list. If the key already exists, the
    * previous entry is overwritten.
-   * Returns previous entry if it exists
+   * Returns previous value if one already exists at that time.
```
defunctzombie, line 35:
```suggestion
   * Returns previous value if one already exists.
```

PR #7306, opened 2024-01-08
Throttle frame axes settings updates

snosenzo:
**User-Facing Changes**

- Better performance in 3D and image panels when settings are open

**Description**

From testing with the settings open and transforms section open FPS with the file from websocket I'm using goes from ~5 fps before to ~33fps after. I tested this after the mcap looped in websocket. This should deliver much better performance generally.

FG-6241


achim-k (approved):

packages/studio-base/src/panels/ThreeDeeRender/renderables/FrameAxes.ts
```diff
@@ -227,6 +228,10 @@ export class FrameAxes extends SceneExtension<FrameAxisRenderable> {
     ];
   }
 
+  #throttledUpdateSettingsTree = _.throttle(() => {
```
defunctzombie, line 231: Why 500ms? Typically we've used 100ms as the "still feels responsive" update time.

https://www.nngroup.com/articles/response-times-3-important-limits/

How does this affect users who manually update the transform tree items?

packages/studio-base/src/panels/ThreeDeeRender/renderables/FrameAxes.ts
```diff
@@ -237,7 +242,7 @@ export class FrameAxes extends SceneExtension<FrameAxisRenderable> {
 
     // Update all the transforms settings nodes each frame since they contain
     // fields that change when currentTime changes
```
jtbandes, line 244: This whole concept of updating in startFrame seems kinda wasteful. I mean I get why it is happening because it depends on currentTime... but it's a shame that this code has to run even if you don't have the settings sidebar open.

jtbandes (approved):

packages/studio-base/src/panels/ThreeDeeRender/renderables/FrameAxes.ts
snosenzo: I think this is a little different in terms of responsiveness since these changes don't always happen as a result of user input. Sure in the seek example it does and the response will be sub .1 seconds, but during playback I don't think that users expect updates this frequently. Even if they are watching the transform values change during playback, I don't think they're expecting to see a continuous stream of all of the values because we can't show that because we don't have a frame for every message, we group them up into a frame which has a duration. I think we just need to show that the UI is doing _something_ to keep up with the values changing so that the user has some idea what these values represent. I think .1s is a bit too frequent for what we need here and not worth the 5x extra function calls for updating the UI. I think 1s a bit too much such that it will feel delayed, but I'm open to it. 

> How does this affect users who manually update the transform tree items?

This does not affect users who manually update transform tree items because those are separate fields on the transform that appear when `Editable` is set to true and are meant to be modifiers on the other values. These values do not change with current time.

packages/studio-base/src/panels/ThreeDeeRender/renderables/FrameAxes.ts
snosenzo: I agree. FWIW I don't think this is doing _as much_ when the settings panel isn't open. It's just updating the settings object but not rendering the UI, which isn't too bad, but definitely still has some impact

defunctzombie (approved):
I'd give your FPS test a go at 100ms and see how that compares to 500 since visually it the field updates will look less jumpy at 100ms but if you settle on 500 - include some shorter variant of your justification as a code comment so we have the thought process for number selection recorded alongside the logic.

snosenzo:
@defunctzombie 

> I'd give your FPS test a go at 100ms and see how that compares to 500 since visually it the field updates will look less jumpy at 100ms but if you settle on 500 - include some shorter variant of your justification as a code comment so we have the thought process for number selection recorded alongside the logic.

Here are the results:
 - .1s
	- just transforms open: 30fps
	- individual tf open:  24fps
- .5s
	- just transforms open: settles around 40 fps
	- individual tf open:  35-40 fps

I think the extra 30% fps from .5s is worth it and doesn't feel sluggish, whereas the .1s moves a bit too fast to read even. Going to keep it at .5s

snosenzo:
@jtbandes There seemed to be an issue with the `useFixtureQueue` function  and I felt it was confusing to use `setSubscriptions` to update the fixture so I removed it and replaced it with the same pattern we have for ImageAnnotations updating. I think generally we probably want a `useFixtureQueue` function that takes a series of fixtures and applies them after each update is flushed, but what was there made little sense to me because the `queue` part of it was inherent in the zustand fixture and not what was being passed to the hook.

PR #7307, opened 2024-01-09
Add tooltips with keyboard shortcuts to playback controls

2metres:
**User-Facing Changes**
Add tooltips to Playback controls.

### DIFF
--- a/packages/studio-base/src/components/PlaybackControls/index.tsx
+++ b/packages/studio-base/src/components/PlaybackControls/index.tsx
@@ -221,43 +221,67 @@ export default function PlaybackControls(props: {
             <PlaybackTimeDisplay onSeek={seek} onPause={pause} />
           </Stack>
           <Stack direction="row" alignItems="center" gap={1}>
-            <HoverableIconButton
-              disabled={disableControls}
-              size="small"
-              title="Seek backward"
-              icon={<Previous20Regular />}
-              activeIcon={<Previous20Filled />}
-              onClick={() => {
-                seekBackwardAction();
-              }}
-            />
-            <HoverableIconButton
-              disabled={disableControls}
-              size="small"
-              title={isPlaying ? "Pause" : "Play"}
-              onClick={togglePlayPause}
-              icon={isPlaying ? <Pause20Regular /> : <Play20Regular />}
-              activeIcon={isPlaying ? <Pause20Filled /> : <Play20Filled />}
-            />
-            <HoverableIconButton
-              disabled={disableControls}
-              size="small"
-              title="Seek forward"
-              icon={<Next20Regular />}
-              activeIcon={<Next20Filled />}
-              onClick={() => {
-                seekForwardAction();
-              }}
-            />
+            <Tooltip
+              title={
+                <>
+                  Seek backward <kbd>←</kbd>
+                </>
+              }
+            >
+              <HoverableIconButton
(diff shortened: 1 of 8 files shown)


defunctzombie:
We need to profile this to make sure this thing is not rendering at 60fps - similar to the 3d panel where we convinced ourselves this is ok. Alternatively find a way to make the tooltip component use less scary. We need to be mindful of the impact MUI's lack of performance care has on our app.

2metres:
From my observation I think the Play Loop and Seek buttons are safe. 

Unrelated to the tooltip though
I am seeing an issue when the loop playback option is active where the play button changes to the pause momentarily at when the loop start/end happens. Which causes the component to rerender as the icon has changed

2metres:
Can you direct me to the issue where we determined that the MUI tooltip has performance issues? So that I can better understand the cases where I **has** caused issues.

amacneil:
:+1:

packages/studio-base/src/components/PlaybackControls/index.tsx
```diff
@@ -221,43 +221,67 @@ export default function PlaybackControls(props: {
...
+                size="small"
+                onClick={togglePlayPause}
+                icon={isPlaying ? <Pause20Regular /> : <Play20Regular />}
+                activeIcon={isPlaying ? <Pause20Filled /> : <Play20Filled />}
+              />
+            </Tooltip>
+            <Tooltip
+              title={
+                <>
+                  Seek foreward <kbd>→</kbd>
```
amacneil, line 260:
```suggestion
                  Seek forward <kbd>→</kbd>
```

packages/theme/src/components/MuiTooltip.ts
```diff
@@ -21,6 +23,14 @@ export const MuiTooltip: OverrideComponentReturn<"MuiTooltip"> = {
       backdropFilter: "blur(3px)",
       fontWeight: "normal",
       fontSize: theme.typography.caption.fontSize,
+
+      kbd: {
+        background: tc(theme.palette.common.white).darken(45).toString(),
```
defunctzombie, line 28: Does mui not have color utilities we can use to darken/lighten colors?

2metres: It does but we use tinycolor2 elsewehere as well. For color manipulation I'm trying to stick with tinycolor2

defunctzombie: Why?

packages/theme/src/components/MuiTooltip.ts
```diff
@@ -2,25 +2,35 @@
...
 export const MuiTooltip: OverrideComponentReturn<"MuiTooltip"> = {
   defaultProps: {
     arrow: true,
     TransitionComponent: Fade,
+    enterDelay: 500,
   },
   styleOverrides: {
     arrow: ({ theme }) => ({
-      color: alpha(theme.palette.grey[700], 0.92),
+      color: tc(theme.palette.grey[700]).setAlpha(0.92).toString(),
```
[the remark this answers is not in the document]
2metres: I was recently asked a question about which color manipulation library we are using consistently across the app. Noticed that this was using the MUI alpha function and moved to use tinycolor2 instead

2metres:
Closing in favor of #7364

PR #7308, opened 2024-01-09
variable slider: Add debounce for applying global variable update

achim-k:
**User-Facing Changes**
None

**Description**
When adjusting the slider value, use a debounced function for applying the new value to the global variable. This will prevent the global variable from changing frequently while the slider is being moved which in turn might prevent some heavy processing depending on how the variable is being used.

### DIFF
--- a/packages/studio-base/src/panels/VariableSlider/index.tsx
+++ b/packages/studio-base/src/panels/VariableSlider/index.tsx
@@ -35,11 +35,19 @@ function VariableSliderPanel(props: Props): JSX.Element {
   const { min = 0, max = 10, step = 1 } = sliderProps;
   const globalVariableValue = globalVariables[globalVariableName];
   const theme = useTheme();
+  const [sliderValue, setSliderValue] = React.useState<number | number[]>(
+    typeof globalVariableValue === "number" ? globalVariableValue : 0,
+  );
 
   useVariableSliderSettings(config, saveConfig);
 
-  const sliderOnChange = useCallback(
-    (_event: Event, value: number | number[]) => {
+  const sliderOnChange = useCallback((_event: Event, value: number | number[]) => {
+    setSliderValue(value);
+  }, []);
+
+  const sliderOnChangeCommitted = useCallback(
+    (_event: unknown, value: number | number[]) => {
+      setSliderValue(value);
       if (value !== globalVariableValue) {
         setGlobalVariables({ [globalVariableName]: value });
       }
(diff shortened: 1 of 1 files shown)


snosenzo (approved):
might be worth mentioning in the user-facing section that the variable slider only changes the value on mouse-up. 
Also, did we consider using `debounce` instead? wondering if that might be a better UX for people who don't necessarily want to trigger up.

packages/studio-base/src/panels/VariableSlider/index.tsx
```diff
@@ -35,11 +35,19 @@ function VariableSliderPanel(props: Props): JSX.Element {
   const { min = 0, max = 10, step = 1 } = sliderProps;
   const globalVariableValue = globalVariables[globalVariableName];
   const theme = useTheme();
+  const [sliderValue, setSliderValue] = React.useState<number | number[]>(
```
defunctzombie, line 38: nit: we tend to use the the react builtin hooks without the "React." prefix.
```suggestion
  const [sliderValue, setSliderValue] = useState<number | number[]>(
```

defunctzombie (commented):
I agree with @snosenzo that we should give debounce a try. If you don't have a large dataset or if applying the variable change is otherwise a cheap operation it could be a nicer UX to be able to "move" the slider to a value and see the display update while still holding down the mouse to maybe move the slider more.

Whichever method we ultimately choose we should leave a code comment on the tradeoff choice with ux and performance for future maintainers.

jtbandes (commented):
The original use case for this panel was allowing the user to scrub through a series of predicted time steps while watching the 3D view to see where the robot was at that time. I believe this change would make that workflow a lot worse. That said, I don't know how most folks are using the panel in reality these days.

One option could be to make a setting for it – e.g. `continuous: boolean` (cf. https://developer.apple.com/documentation/appkit/nscontrol/1428952-iscontinuous)

defunctzombie:
@jtbandes with a debounce we could still support scrubbing. The thing we want to avoid is triggering updates as you move through the various values on your way towards the one you want (which I think currently happens)?

defunctzombie (approved):

jtbandes:
> The thing we want to avoid is triggering updates as you move through the various values on your way towards the one you want

This behavior of showing intermediate is useful to make the workflow feel more interactive. We will just have to see if anyone complains after this is shipped.

defunctzombie:
> This behavior of showing intermediate is useful to make the workflow feel more interactive. We will just have to see if anyone complains after this is shipped.

I played with this PR and with the debounce logic it feels "interactive". At 250ms until the change in variable I could scan over to the new variable and I saw it update in the variable sidebar.

PR #7310, opened 2024-01-09
Fix Image panel memory leak when disposing while decoding an image

snosenzo:
**User-Facing Changes**

- Fix Image panel memory leak when unmounting while decoding an image

**Description**

Terminating the worker while the promise hasn't been resolved results in the promise never getting resolved nor rejected. This promise is referenced within the renderable itself and has callbacks within the renderable. This promise being linked to the global worker object and the renderable cause the entire renderer to not be disposed, leaving multiple instances in memory. This abort lets the promise resolve and clean itself up so that it doesn't keep it's associations in memory.

sam/fg-6148-dataset-live-stream-raw-images-leads-to-oom

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/Images/WorkerImageDecoder.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/Images/WorkerImageDecoder.ts
@@ -20,6 +20,8 @@ import { Image as RosImage } from "../../ros";
 export class WorkerImageDecoder {
   #worker: Worker;
   #remote: Comlink.Remote<(typeof import("./WorkerImageDecoder.worker"))["service"]>;
+  /** Aborts the current decode promise. */
+  #abort?: () => void;
 
   public constructor() {
     this.#worker = new Worker(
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/panels/ThreeDeeRender/renderables/Images/WorkerImageDecoder.ts
```diff
@@ -36,10 +38,23 @@ export class WorkerImageDecoder {
     image: RosImage | RawImage,
     options: Partial<RawImageOptions>,
   ): Promise<ImageData> {
-    return await this.#remote.decode(image, options);
+    return await new Promise((resolve, reject) => {
+      this.#abort = reject;
+      void this.#remote.decode(image, options).then((decodedImage) => {
```
defunctzombie, line 43:
```suggestion
      void this.#remote.decode(image, options).then(resolve);
```

packages/studio-base/src/panels/ThreeDeeRender/renderables/Images/WorkerImageDecoder.ts
```diff
@@ -36,10 +38,23 @@ export class WorkerImageDecoder {
     image: RosImage | RawImage,
     options: Partial<RawImageOptions>,
   ): Promise<ImageData> {
-    return await this.#remote.decode(image, options);
+    return await new Promise((resolve, reject) => {
+      this.#abort = reject;
```
defunctzombie, line 42: What happens if you make a new promise? It will replace the #abort instance and you will never abort the old one...

snosenzo: Good catch. Not sure if it makes more sense to do this inside or outside of the promise. I put it inside.

defunctzombie (approved):
We will need to fix comlink - we need this behavior everywhere we use it and terminate workers.

packages/studio-base/src/panels/ThreeDeeRender/renderables/Images/WorkerImageDecoder.ts
```diff
@@ -36,10 +38,25 @@ export class WorkerImageDecoder {
     image: RosImage | RawImage,
     options: Partial<RawImageOptions>,
   ): Promise<ImageData> {
-    return await this.#remote.decode(image, options);
+    return await new Promise((resolve, reject) => {
+      // abort previous request
+      if (this.#abort) {
+        this.#abort();
+      }
```
jtbandes, line 45: nit
```suggestion
      this.#abort?.();
```

jtbandes (approved):

PR #7312, opened 2024-01-09
Move currentTime & hover bar rendering back to <div>s

jtbandes:
**User-Facing Changes**
N/A - feature branch

**Description**

- Remove `annotations` and use `<div>`s instead
- Restore ability to hover when in non-timeseries mode (blue line, does not update playback bar or other plot panels)
- Add `xScaleChanged` event to PlotCoordinator to allow tracking scale updates


packages/studio-base/src/panels/Plot/VerticalBars.tsx
```diff
@@ -0,0 +1,130 @@
...
+  if (xValue < scale.min || xValue > scale.max) {
+    return undefined;
+  }
+
+  // Linear interpolation to place the xValue within min/max
+  return scale.left + ((xValue - scale.min) / (scale.max - scale.min)) * pixelRange;
+}
+
+/**
+ * Display vertical bars at the currentTime & the hovered time
```
defunctzombie, line 67: add a note that this is done as a component to limit the re-rendering that happens.

I'd also consider getting a reference to the <divs> and updating the style position manually - thus bypassing even more react re-renders and temporary functions, etc for what we know is a code path running at 60fps.

packages/studio-base/src/panels/Plot/PlotCoordinator.ts
```diff
@@ -30,6 +29,9 @@ import type { PlotConfig } from "./types";
 
 type EventTypes = {
   timeseriesBounds(bounds: Immutable<Bounds1D>): void;
+
+  /** X scale changed. Call `getXScale()` to get the latest scale. */
```
defunctzombie, line 33: Pass the scale as an arg to the event would be more natural and the receiver does not need to call anything to get the scales

defunctzombie (approved):

PR #7314, opened 2024-01-10
Add delete to State Transition legend

2metres:
**User-Facing Changes**
Update state transition legend to look more like PlotLegend and include quick delete

Screen Shot 2024-01-10 at 1 58 15 pm


jtbandes (approved):

jtbandes:
Definitely looks better but can we converge these two panels even more? I'm not sure why we would want them to look different:

image
image

PR #7317, opened 2024-01-10
Fix Image panel memory leak when unmounting while decoding an image

snosenzo:
**User-Facing Changes**

- Fix Image panel memory leak when unmounting while decoding an image

**Description**
Undoes the revert from: https://github.com/foxglove/studio/pull/7316
And fixes the crash and not show error images if there is a decode request received while the previous request is processing.

My previous attempt used closures for both this.#abort = () =>  and .finally in the promise, which used the prevAbort . When both of these use closures it creates a promise chain linking each promise to the next promise such that they are never garbage collected unless all promises have been resolved.


packages/studio-base/src/panels/ThreeDeeRender/renderables/Images/WorkerImageDecoder.ts
```diff
@@ -39,12 +39,19 @@ export class WorkerImageDecoder {
...
     return await new Promise((resolve, reject) => {
+      const oldAbort = this.#abort;
       // abort previous request
-      if (this.#abort) {
-        this.#abort();
-      }
-      this.#abort = reject;
-      void this.#remote.decode(image, options).then(resolve).catch(reject);
+      this.#abort = () => {
+        reject("Decode aborted.");
```
defunctzombie, line 45: We should reject using error instances rather than free strings so we have stack traces.

```suggestion
        reject(new Error("Decode aborted."));
```

packages/studio-base/src/panels/ThreeDeeRender/renderables/Images/WorkerImageDecoder.ts
```diff
@@ -39,12 +39,19 @@ export class WorkerImageDecoder {
     options: Partial<RawImageOptions>,
   ): Promise<ImageData> {
     return await new Promise((resolve, reject) => {
+      const oldAbort = this.#abort;
```
defunctzombie, line 42:
```suggestion
      const prevAbort = this.#abort;
```

defunctzombie (approved):
Its not obvious to me why the caller doesn't wait to decode another but I guess nothing requires that order of operations. Might be worth a comment here that you could have two decodes in-flight at once so if the new one happens to finish before the old it aborts the old.

packages/studio-base/src/panels/ThreeDeeRender/renderables/Images/WorkerImageDecoder.ts
```diff
@@ -36,10 +37,39 @@ export class WorkerImageDecoder {
...
+     * Renderer to never be garbage collected because it's linked to this ongoing
+     * promise and worker.
+     */
+    this.#abort?.();
+    this.#abort = undefined;
     this.#worker.terminate();
   }
 }
+
+/** Creates a function that calls the reject function with an abort error */
```
defunctzombie, line 70: The purpose of this is to avoid any possibility of closing over other parts of the promise or options passed to the promise?

defunctzombie (approved):
I'd like us to fix comlink to handle this and then re-visit testing this behavior and see if we can drop this bespoke abort handling. I think we've created a linear issue for it so I am hopeful this workaround will be short-lived 🤞 and we have a more robust solution soon.

packages/studio-base/src/panels/ThreeDeeRender/renderables/Images/WorkerImageDecoder.ts
```diff
@@ -36,10 +37,39 @@ export class WorkerImageDecoder {
...
+       * Always test that these promises are not keeping images in memory over time.
+       */
+
+      /** More decode requests can be made while the last one is being processed
+       * so we need to keep track of the previous abort function and abort if the
+       * next one finishes before the first. If it's already resolved, then the abort is a noop.
+       */
+      const prevAbort = this.#abort;
+      this.#abort = makeAbort(reject);
+      void this.#remote.decode(image, options).then(resolve).catch(reject).finally(prevAbort);
```
jtbandes, line 53: Putting question here for posterity - how does this not end up creating a large chain where each promise keeps a reference to the previous one via prevAbort?

snosenzo: In theory I agree with that logic, however I've triple checked that unmounting the panel during playback does not cause a memory leak from this.

packages/studio-base/src/panels/ThreeDeeRender/renderables/Images/WorkerImageDecoder.ts
snosenzo: that's correct.

PR #7318, opened 2024-01-10
Use object key + isEqual instead of JSON.stringify

jtbandes:
**User-Facing Changes**
N/A - feature branch

**Description**
Remove `JSON.stringify` and replace with object key and `isEqual`


defunctzombie (approved):
LGTM - not for this PR but I think we can hoist this to the coordinator or the panel and normalize the settings and fill the paths before sending downstream

PR #7324, opened 2024-01-12
use a string key for plot series

defunctzombie:
**User-Facing Changes**

**Description**
Disambiguates two series with the same path but different header stamps so they appear as two separate series.

See the "preloaded data in binary blocks" story for an example of the fix compared to the base branch


defunctzombie:
@jtbandes merged this to the other PR to keep iterating. If you end up not liking this stringifyRosPath approach we can change it in the other PR.

PR #7326, opened 2024-01-16
Prevent body from scrolling

2metres:
**User-Facing Changes**
None

**Description**
Resolves bug where window frame scrolls when tooltip is present inside PlotPanel

### DIFF
--- a/packages/studio-base/src/components/PlaybackControls/index.tsx
+++ b/packages/studio-base/src/components/PlaybackControls/index.tsx
@@ -64,6 +64,13 @@ const useStyles = makeStyles()((theme) => ({
     backgroundColor: theme.palette.background.paper,
     borderTop: `1px solid ${theme.palette.divider}`,
     zIndex: 100000,
+    overflowX: "auto",
+  },
+  scrubberWrapper: {
+    position: "sticky",
+    top: 0,
+    right: 0,
+    left: 0,
   },
   disabled: {
     opacity: theme.palette.action.disabledOpacity,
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/components/PlaybackControls/index.tsx
```diff
@@ -183,8 +190,10 @@ export default function PlaybackControls(props: {
       <RepeatAdapter play={play} seek={seek} repeatEnabled={repeat} />
       <KeyListener global keyDownHandlers={keyDownHandlers} />
       <div className={classes.root}>
-        <Scrubber onSeek={seek} />
-        <Stack direction="row" alignItems="center" flex={1} gap={1} overflowX="auto">
+        <div className={classes.scrubberWrapper}>
```
jtbandes, line 193: What's the purpose of adding a wrapper here?

2metres: So that the scrollbar appears outside the padded container instead of inside.

Before
Screen Shot 2024-01-17 at 1 47 46 pm

After
Screen Shot 2024-01-17 at 1 47 15 pm

2metres: It allows the scrubber to remain full width while the other elements scroll

jtbandes: I’m confused — I thought the point of the overflow:hidden was to disable scrolling. Why is there still a scrollbar?

2metres: We've disabled scrolling on the body but we still allow scrolling on the playback bar controls

jtbandes: My gut feeling would be that we don't want to allow scrolling on the playback bar since it's part of the app "chrome" that is supposed to remain fixed. I think it could make more sense to start wrapping or hiding the controls if they don't fit. (However I also don't think we have any particularly strong goals right now about making this work nicely on tiny screens?)

jtbandes: I'm not sure how to make this work in chrome (it seems to limit the width when I use the dev tools in responsive mode) but in Safari this is a pretty weird experience, especially since the buttons scroll but the scrubber does not:

https://github.com/foxglove/studio/assets/14237/572d9825-9949-4d90-8fd8-0282b2868d1f

2metres: > since the buttons scroll but the scrubber does not

The scrubber is position sticky and full width and we are not truncating it. The buttons scrolling has been the behavior since we redsigned the playback bar ages ago. I guess the hope is that we at least make the content accessible even if it's not an ideal UX

defunctzombie:
LR: visually does the good stuff

jtbandes (approved):

PR #7327, opened 2024-01-16
Fix for settings changes causing another decoding of images unnecessarily

snosenzo:
**User-Facing Changes**

- Fix changing image settings causing extra decode of image. Notably this was a problem for changing settings with video topics.
- remove colormode settings for non-raw images in ImageMode settings

**Description**

The cause of the issue was the emitState image change logic comparing a non-referentially stable value, and also the gradient in the settings being a non-referentially stable value. Both of these paths are activated by the chancing of the setting and can cause the same image to be decoded again.

FG-6127

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageMode.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageMode.ts
@@ -110,6 +110,8 @@ export const ALL_SUPPORTED_IMAGE_SCHEMAS = new Set([
   ...COMPRESSED_IMAGE_DATATYPES,
 ]);
 
+const SUPPORTED_RAW_IMAGE_SCHEMAS = new Set([...RAW_IMAGE_DATATYPES, ...ROS_IMAGE_DATATYPES]);
+
 const ALL_SUPPORTED_CALIBRATION_SCHEMAS = new Set([
   ...CAMERA_INFO_DATATYPES,
   ...CAMERA_CALIBRATION_DATATYPES,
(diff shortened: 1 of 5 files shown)


packages/studio-base/src/panels/ThreeDeeRender/renderables/ImageMode/ImageMode.ts
```diff
@@ -418,21 +426,21 @@ export class ImageMode
...
+      sortPrefixMatchesToFront(calibrationTopics, imageTopicName, (option) => option.label);
     }
 
     // add unselected camera calibration option
     calibrationTopics.unshift({ label: "None", value: undefined });
 
     const imageTopicExists = !(
-      imageTopic && !imageTopics.some((topic) => topic.value === imageTopic)
+      imageTopicName && !imageTopics.some((topic) => topic.value === imageTopicName)
     );
```
jtbandes, line 438: `!(A && !B)` => `!A || B`

packages/studio-base/src/panels/ThreeDeeRender/renderables/Images/ImageRenderable.ts
```diff
@@ -21,6 +22,8 @@ import { CameraInfo } from "../../ros";
 import { DECODE_IMAGE_ERR_KEY, IMAGE_TOPIC_PATH } from "../ImageMode/constants";
 import { ColorModeSettings } from "../colorMode";
 
+const EMPTY_ARRAY: unknown[] = [];
```
defunctzombie, line 25: Object.freeze maybe to ensure it does not get changed?

defunctzombie (approved):

packages/studio-base/src/panels/ThreeDeeRender/renderables/Images/ImageRenderable.ts
```diff
@@ -154,7 +157,7 @@ export class ImageRenderable extends Renderable<ImageUserData> {
     if (
       prevSettings.colorMode !== newSettings.colorMode ||
       prevSettings.flatColor !== newSettings.flatColor ||
-      prevSettings.gradient !== newSettings.gradient ||
+      !vecEqual(prevSettings.gradient ?? EMPTY_ARRAY, newSettings.gradient ?? EMPTY_ARRAY) ||
```
jtbandes, line 160: This function doesn't seem to check that `a.length === b.length` 😕 Maybe we should use isEqual from lodash instead (and/or fix the vecEqual function)?

jtbandes (approved):

PR #7332, opened 2024-01-17
Remove outline on datagrid cell header

2metres:
**User-Facing Changes**
None

**Description**
Add `:focus` to selector as well as `:focus-within`


amacneil (approved):

defunctzombie:
> Add `:focus` to selector as well as `:focus-within`

PR meta feedback - this doesn't tell me anything as a reviewer about why we want this change or motivation for this change. I can read the PR to see that this line changed - even copy/pasting some context from the linear issue would help guide the review.

PR #7335, opened 2024-01-17
Disable playback loop when loop happens too frequently

jtbandes:
**User-Facing Changes**
Not worth calling out

**Description**
To avoid bad UX and log spam, disable looping when the automatic loop happens more than once per 250ms. Also log loop events separately from regular "play" events.

https://github.com/foxglove/studio/assets/14237/515b1554-d768-47be-9a3c-8bede6f0f901

### DIFF
--- a/packages/studio-base/src/components/MessagePathSyntax/MessagePathInput.tsx
+++ b/packages/studio-base/src/components/MessagePathSyntax/MessagePathInput.tsx
@@ -159,6 +159,7 @@ export default React.memo<MessagePathInputBaseProps>(function MessagePathInput(
     () => messagePathStructures(datatypes),
     [datatypes],
   );
+  /** A map from each possible message path to the corresponding MessagePathStructureItem */
   const allStructureItemsByPath = useMemo(
     () =>
       new Map(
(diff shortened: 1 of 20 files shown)


defunctzombie:
🤔 This seems weird to me. I can't quite articulate why - but its not specifically obvious why looping quickly is something we have to disable automatically for the user? I do find the "dataset is too small" somewhat awkward wording as well.

My initial reaction to this was that if you do want to restrict looping that could be a property of the dataset duration - but I see that it could depend on how fast you play rather than the dataset duration since you could play a short dataset slowly.

My current feels on this are 👎 - if the user wants to loop quickly they can loop quickly and instead we should be smarter about our event collection if that's what this is trying to resolve.

jtbandes:
> be smarter about our event collection

What would "smarter" look like? Would it look different than the time-based check here? Are you suggesting that the time-based check could gate the event logging only and not gate the actual looping?

defunctzombie:
LR: 👎 on the notification. 👍 to fixing the play/pause button behavior. Open to considering a popup that is closer to the speed setting and pushes the user into the ball pit of success.

PR #7336, opened 2024-01-17
Reduce instances of erroneous "reset view" on the plot

defunctzombie:
**User-Facing Changes**
None - not worth calling out

**Description**
The plot and state transition panels use different approaches to "global bounds" when using the "sync with other plots" feature. The plot panel only sets the global bounds when a user interacts with the panel (zoom or pan), otherwise it leaves the bounds unset. The state transition panel sets the bounds when it loads. When both panels are used on a layout, this causes the existing plot panel logic for when to show the reset button to think the bounds are set and you could reset them. However, the bounds already match what the panel is displaying so there's no need to reset.

This change updates the "show reset" logic to account for bounds being set and the panel already matching those bounds. When that happens, there's no need to show the reset button.


packages/studio-base/src/panels/Plot/PlotCoordinator.ts
```diff
@@ -312,6 +312,34 @@ export class PlotCoordinator extends EventEmitter<EventTypes> {
     return await this.#datasetsBuilder.getCsvData();
   }
 
+  /**
+   * Get the xBounds exluding the global bounds. This returns the xBounds that would be used
+   * if the global bounds were not overriding.
+   */
+  public getXBoundsWithoutGlobalBounds(): Partial<Bounds1D> {
```
jtbandes, line 319: Rather than exposing an almost-identical function and doing the equality check in Plot.tsx, what about exposing a function like `isUsingGlobalBounds()` which basically does `return this.#interactionBounds == undefined && this.#globalBounds != undefined` (capturing the situations where globalBounds would have been used)?

packages/studio-base/src/panels/Plot/Plot.tsx
```diff
@@ -508,10 +508,10 @@ export function Plot(props: Props): JSX.Element {
         deltaX: event.deltaX,
         boundingClientRect: boundingRect.toJSON(),
       });
+      setShowReset(true);
```
jtbandes, line 511: Feels like it could help to rename this `showReset` variable to something more directly related to how we use it – such as `interacted` (since it seems to be set only on mouse interactions)

defunctzombie:
Good feedback - I've updated the logic in Plot.tsx and PlotCoordinator to do away with `showReset` and `getXBoundsWithoutGlobalBounds`. Instead I have one `canReset` state variable and the coordinator has a `canReset` function that handles the logic for determining if the plot could be reset or not. This keeps the logic contained within the coordinator instead of splitting it between the react layer and the coordinator.

packages/studio-base/src/panels/Plot/Plot.tsx
```diff
@@ -508,10 +512,10 @@ export function Plot(props: Props): JSX.Element {
         deltaX: event.deltaX,
         boundingClientRect: boundingRect.toJSON(),
       });
+      setCanReset(coordinator.canReset());
```
jtbandes, line 515: should we do this after every interaction event?

packages/studio-base/src/panels/Plot/Plot.tsx
```diff
@@ -247,6 +249,8 @@ export function Plot(props: Props): JSX.Element {
 
     const unsub = subscribeMessagePipeline((state) => {
       coordinator.handlePlayerState(state.playerState);
+
+      setCanReset(coordinator.canReset());
```
jtbandes, line 253: Maybe instead of remembering to call `.canReset()` again after every change, we should add a new event to the coordinator EventTypes?

packages/studio-base/src/panels/Plot/PlotCoordinator.ts
```diff
@@ -312,6 +312,44 @@ export class PlotCoordinator extends EventEmitter<EventTypes> {
...
+    const xMinIfFollowMode =
+      this.#isTimeseriesPlot && this.#followRange != undefined && xMax != undefined
+        ? xMax - this.#followRange
+        : undefined;
+    const xMin = xMinIfFollowMode ?? this.#configBounds.x.min ?? this.#datasetRange?.min;
+
+    return { min: xMin, max: xMax };
+  }
+
   #getXBounds(): Partial<Bounds1D> {
```
jtbandes, line 353: It seems like this could now be implemented as `xMax = this.#interactionBounds?.x.max ?? this.#globalBounds?.max ?? this.#getXResetBounds().max` etc?

jtbandes (approved):

packages/studio-base/src/panels/Plot/PlotCoordinator.ts
```diff
@@ -312,31 +315,50 @@ export class PlotCoordinator extends EventEmitter<EventTypes> {
...
         : undefined;
+    const xMin = xMinIfFollowMode ?? this.#configBounds.x.min ?? this.#datasetRange?.min;
+
+    return { min: xMin, max: xMax };
+  }
+
+  #getXBounds(): Partial<Bounds1D> {
+    // Interaction, synced global bounds override the config and data source bounds in precedence
+    const xMax =
+      this.#interactionBounds?.x.max ?? this.#globalBounds?.max ?? this.#getXResetBounds().max;
```
jtbandes, line 359: nit: call the function only once

jtbandes (approved):

PR #7338, opened 2024-01-17
Keep referentially stable blocks in TopicAliasingPlayer

defunctzombie:
**User-Facing Changes**
None

**Description**
The AliasingStateProcessor (part of the TopicAliasingPlayer machinery) processes blocks to add the aliased MessageEvent arrays to `messageByTopic`. The existing behavior would receive a new set of blocks and update the aliased topic messages in all of them. This resulted in a new messages array reference for all aliased topics any time a new block is added. Changing the message array reference signals to downstream that the specific messages in the block for that topic have changed and can trigger work on those messages. In the case of plots, it causes the internal block cursors to think the data is changed and re-send all the block data for the series.

This change updates the `AliasingStateProcessor` with a new assistant - the `BlockTopicProcessor`. The BlockTopicProcessor remembers the aliased message arrays for the aliased topics and returns those if the source topic messages are unchanged (referentially stable). This allows downstream consumers to continue using the referential equality of message arrays to identify if the data is changed.

### DIFF
--- a/packages/studio-base/src/players/TopicAliasingPlayer/AliasingStateProcessor.ts
+++ b/packages/studio-base/src/players/TopicAliasingPlayer/AliasingStateProcessor.ts
@@ -46,7 +46,7 @@ export class AliasingStateProcessor implements IStateProcessor {
     this.#inverseMapping = invertAliasMap(mapping);
 
     this.#blockProcessors = Array.from(mapping.entries()).map(
-      (entry) => new BlockTopicProcessor(entry[0], entry[1]),
+      ([topic, aliases]) => new BlockTopicProcessor(topic, aliases),
     );
   }
 
@@ -117,19 +117,16 @@ export class AliasingStateProcessor implements IStateProcessor {
         return undefined;
       }
 
-      const accumulatedMessagesByTopic = this.#blockProcessors.reduce((acc, processor) => {
+      const messagesByTopic = this.#blockProcessors.reduce((acc, processor) => {
         return {
           ...acc,
           ...processor.aliasBlock(block, idx),
         };
-      }, {});
+      }, block.messagesByTopic);
 
       return {
         ...block,
-        messagesByTopic: {
-          ...block.messagesByTopic,
-          ...accumulatedMessagesByTopic,
-        },
+        messagesByTopic,
       };
     });
   });
--- a/packages/studio-base/src/players/TopicAliasingPlayer/AliasingStateProcessor.ts
+++ b/packages/studio-base/src/players/TopicAliasingPlayer/AliasingStateProcessor.ts
@@ -16,6 +16,7 @@ import {
   TopicStats,
 } from "@foxglove/studio-base/players/types";
 
+import { BlockTopicProcessor } from "./BlockTopicProcessor";
 import { IStateProcessor } from "./IStateProcessor";
 
 export type TopicAliasMap = Map<string, string[]>;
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/players/TopicAliasingPlayer/AliasingStateProcessor.ts
```diff
@@ -30,13 +31,23 @@ type MessageBlocks = readonly (undefined | MessageBlock)[];
...
 
+  #blockProcessors: BlockTopicProcessor[] = [];
+
   public constructor(mapping: Im<TopicAliasMap>, problems?: PlayerProblem[]) {
     this.#mapping = mapping;
     this.#problems = problems ?? [];
     this.#inverseMapping = invertAliasMap(mapping);
+
+    this.#blockProcessors = Array.from(mapping.entries()).map(
+      (entry) => new BlockTopicProcessor(entry[0], entry[1]),
```
jtbandes, line 49:
```suggestion
      ([topic, aliases]) => new BlockTopicProcessor(topic, aliases),
```

packages/studio-base/src/players/TopicAliasingPlayer/AliasingStateProcessor.ts
```diff
@@ -101,29 +112,24 @@ export class AliasingStateProcessor implements IStateProcessor {
...
-      if (block == undefined) {
+    return blocks.map((block, idx) => {
+      if (!block) {
         return undefined;
       }
 
+      const accumulatedMessagesByTopic = this.#blockProcessors.reduce((acc, processor) => {
+        return {
+          ...acc,
+          ...processor.aliasBlock(block, idx),
```
jtbandes, line 123: do we have any validation that ensures there aren't two separate topics that alias to the same destination topic?

packages/studio-base/src/players/TopicAliasingPlayer/BlockTopicProcessor.ts
```diff
@@ -0,0 +1,62 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import type { Immutable as Im, MessageEvent } from "@foxglove/studio";
+import type { MessageBlock } from "@foxglove/studio-base/players/types";
+
+type BlockItem = { original: Im<MessageEvent[]>; aliased: Record<string, MessageEvent[]> };
```
jtbandes, line 8:
```suggestion
type BlockItem = { inputEvents: Im<MessageEvent[]>; aliased: Record<string, MessageEvent[]> };
```

packages/studio-base/src/players/TopicAliasingPlayer/AliasingStateProcessor.ts
```diff
@@ -101,29 +112,24 @@ export class AliasingStateProcessor implements IStateProcessor {
...
-              }
-            }
-            acc[topic] = messages;
-          },
-          {} as Record<string, MessageEvent[]>,
-        ),
+        messagesByTopic: {
+          ...block.messagesByTopic,
+          ...accumulatedMessagesByTopic,
+        },
```
jtbandes, line 132: could we have passed `block.messagesByTopic` as the initial value for the `reduce` instead of doing another spread here?

jtbandes (approved):

packages/studio-base/src/players/TopicAliasingPlayer/AliasingStateProcessor.ts
defunctzombie: https://github.com/foxglove/studio/blob/main/packages/studio-base/src/players/TopicAliasingPlayer/StateProcessorFactory.ts#L103

PR #7339, opened 2024-01-17
Don't log play events when playback loops

jtbandes:
**User-Facing Changes**
None

**Description**

Alternative to https://github.com/foxglove/studio/pull/7335
Don't log the "played" event when playback loops automatically.

### DIFF
--- a/packages/studio-base/src/components/PlaybackControls/RepeatAdapter.tsx
+++ b/packages/studio-base/src/components/PlaybackControls/RepeatAdapter.tsx
@@ -12,8 +12,6 @@ import {
 
 type RepeatAdapterProps = {
   repeatEnabled: boolean;
-  // eslint-disable-next-line @foxglove/no-boolean-parameters
-  setRepeat: (repeat: boolean) => void;
   play: (opts: { looped: boolean }) => void;
   seek: (to: Time) => void;
 };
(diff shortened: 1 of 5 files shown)


packages/studio-base/src/components/PlaybackControls/RepeatAdapter.tsx
```diff
@@ -29,27 +29,25 @@ function activeDataSelector(ctx: MessagePipelineContext) {
 export function RepeatAdapter(props: RepeatAdapterProps): JSX.Element {
   const { play, seek, repeatEnabled } = props;
 
-  const activeData = useMessagePipeline(activeDataSelector);
+  const currentTime = useMessagePipeline(selectCurrentTime);
+  const startTime = useMessagePipeline(selectStartTime);
+  const endTime = useMessagePipeline(selectEndTime);
```
jtbandes, line 34: I noticed that this effect seemed to run multiple times. I am guessing there were other things in activeData that were changing, so I reduced the effect dependencies to only the time values rather than the whole activeData.

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
```diff
@@ -209,26 +209,30 @@ export class IterablePlayer implements Player {
...
-    if (opt?.untilTime) {
-      if (this.#currentTime && compare(opt.untilTime, this.#currentTime) <= 0) {
+    if (untilTime) {
+      if (this.#currentTime && compare(untilTime, this.#currentTime) <= 0) {
         throw new Error("Invariant: playUntil time must be after the current time");
       }
-      this.#untilTime = clampTime(opt.untilTime, this.#start, this.#end);
+      this.#untilTime = clampTime(untilTime, this.#start, this.#end);
+    }
+    // don't log "play" events when playback is looping automatically
```
defunctzombie, line 232: feels awkward that we pass in this "looped" parameter to `startPlayback` and the caller needs to tell us this happened...

What would it take to not have this happen? Does this mean dropping the adapter? Maybe it is time.

jtbandes: > the caller needs to tell us this happened...

yeah – currently the player is unaware of looping. That is why it needs to be told that it looped.

One option would be to rename this flag to something like `logAnalyticsEvent: false` to make it clearer what it is used for?

PR #7340, opened 2024-01-17
Abort blockloader when player is closing

snosenzo:
**User-Facing Changes**

- fix memory leak across data sources when blockloading was incomplete

**Description**

`#abortController.abort()` was not being called in `Blockloader.stopLoading` so the blockloader would need to finish loading before the player was closed.

We also might want to consider adding a timeout for a warning or error when `stateClose` is called on the player so that we know if the last player didn't close in a reasonable amount of time.


defunctzombie (approved):
needs more Finalization registry 🤣

achim-k (approved):

PR #7342, opened 2024-01-18
Fix issues with disabled plot series

jtbandes:
**User-Facing Changes**
Fixed issues with disabled series in the Plot panel.

**Description**
- Add `PlotCoordinator.destroy()` to fix an issue where switching from Timestamp to Index mode would still show the timestamp data (because the old coordinator rendered after the new one)
- Ensure all builders include an empty dataset in `getViewportDatasets` for disabled series
- Ensure all builders keep track of the new `enabled` setting in `setSeries` (even if the series did not otherwise change)

### DIFF
--- a/packages/studio-base/src/panels/Plot/Plot.tsx
+++ b/packages/studio-base/src/panels/Plot/Plot.tsx
@@ -344,6 +344,7 @@ export function Plot(props: Props): JSX.Element {
 
     return () => {
       resizeObserver.disconnect();
+      plotCoordinator.destroy();
     };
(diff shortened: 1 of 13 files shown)


defunctzombie (approved):
We need to destroy the plot coordinator because it is doing async work that might complete after we've setup a different coordinator? Should we be aborting that work? That might be harder to do than the booleans you've added.

jtbandes:
I’m not sure the aborting is so important - I suppose it would be the most thorough thing to do but I didn’t feel compelled.

PR #7345, opened 2024-01-18
Periodically trigger render when processing blocks

defunctzombie:
**User-Facing Changes**
TBD
**Description**

### DIFF
--- a/packages/studio-base/src/panels/Plot/PlotCoordinator.ts
+++ b/packages/studio-base/src/panels/Plot/PlotCoordinator.ts
@@ -7,7 +7,7 @@ import EventEmitter from "eventemitter3";
 import { debouncePromise } from "@foxglove/den/async";
 import { filterMap } from "@foxglove/den/collection";
 import { toSec, subtract as subtractTime } from "@foxglove/rostime";
-import { Immutable } from "@foxglove/studio";
+import { Immutable, Time } from "@foxglove/studio";
 import { RosPath } from "@foxglove/studio-base/components/MessagePathSyntax/constants";
 import parseRosPath from "@foxglove/studio-base/components/MessagePathSyntax/parseRosPath";
 import { simpleGetMessagePathDataItems } from "@foxglove/studio-base/components/MessagePathSyntax/simpleGetMessagePathDataItems";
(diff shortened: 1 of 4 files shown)


packages/studio-base/src/panels/Plot/builders/TimestampDatasetsBuilder.ts
```diff
@@ -139,10 +140,53 @@ export class TimestampDatasetsBuilder implements IDatasetsBuilder {
         }
       }
     }
+    */
 
     return { min: 0, max: toSec(subtractTime(activeData.endTime, activeData.startTime)) };
   }
 
+  public async handleBlocks(
```
snosenzo, line 148: Do we need a way to abort or cancel this?

defunctzombie: 🤔 I suppose when we get new blocks we can re-start the process without finishing the existing process. Since we use cursors this would pick up right where it left off. That would also make some UX better when you change series since there's no need to finish the old processing.

packages/studio-base/src/panels/Plot/PlotCoordinator.ts
```diff
@@ -379,4 +381,18 @@ export class PlotCoordinator extends EventEmitter<EventTypes> {
...
+    startTime: Immutable<Time>,
+    blocks: Immutable<(MessageBlock | undefined)[]>,
+  ): Promise<void> {
+    if (!this.#datasetsBuilder.handleBlocks) {
+      return;
+    }
+
+    await this.#datasetsBuilder.handleBlocks(startTime, blocks, async () => {
+      this.#queueDispatchDatasets();
+      await delay(0);
```
jtbandes, line 395: Needs a comment to explain why it shouldn’t be deleted by MarieKondoBot

packages/studio-base/src/panels/Plot/PlotCoordinator.ts
```diff
@@ -379,4 +381,18 @@ export class PlotCoordinator extends EventEmitter<EventTypes> {
     this.emit("xScaleChanged", this.#latestXScale);
     this.emit("pathsWithMismatchedDataLengthsChanged", [...result.pathsWithMismatchedDataLengths]);
   }
+
+  async #dispatchBlocks(
+    startTime: Immutable<Time>,
+    blocks: Immutable<(MessageBlock | undefined)[]>,
+  ): Promise<void> {
+    if (!this.#datasetsBuilder.handleBlocks) {
```
jtbandes, line 389: Feels a little suboptimal that we check this in two places. Not sure I have a better suggestion though.

jtbandes (approved):
Glorious

packages/studio-base/src/panels/Plot/builders/TimestampDatasetsBuilder.ts
snosenzo: Also wondering if you unmount the panel while it's loading does this continue?

defunctzombie:
@jtbandes @snosenzo I've added logic to bail processing when the coordinator is destroyed or when the input blocks change. This avoids having to finish processing old blocks when we have new ones. This comes up when adding a new field to an existing topic and we should start to show that field as soon as blocks have that data. Finishing processing for the existing field is not useful because it will be re-done again on the next loop through blocks since the underlying blocks change with the addition of the new field.

packages/studio-base/src/panels/Plot/builders/TimestampDatasetsBuilder.ts
```diff
@@ -163,7 +163,10 @@ export class TimestampDatasetsBuilder implements IDatasetsBuilder {
           items: pathItems,
         });
 
-        await progress();
+        const bail = await progress();
```
snosenzo, line 166: nit: I feel like we usually use `abort` terminology in these cases?

snosenzo (approved):

PR #7346, opened 2024-01-18
Keep current values in legend when editing plot config

jtbandes:
**User-Facing Changes**
Not worth calling out

**Description**
In `handleConfig`, rather than completely resetting `currentValues`, keep values that corresponded to the same path in the old config.


packages/studio-base/src/panels/Plot/PlotCoordinator.ts
```diff
@@ -198,7 +199,14 @@ export class PlotCoordinator extends EventEmitter<EventTypes> {
 
       return fillInGlobalVariablesInPath(parsed, globalVariables);
     });
-    this.#currentValues = [];
+    // Keep current values for paths that match existing ones
+    this.#currentValues = newSeriesPaths.map((path) => {
```
defunctzombie, line 203: There is logic further down in this same method that parses and fills in the global variables. That logic also assigns a key to each series. That key could be the lookup. Also we would avoid parsing and filling variables twice.

I guess the semantic is that `currentValues` is an array with index corresponding to the index of the path? I think you can also do that in the `filterMap` lower in the same method.

defunctzombie (approved):
I think we can drop the two iterations we have on `config.paths`.

snosenzo (approved):
not much to add except that I agree with roman's comments

packages/studio-base/src/panels/Plot/PlotCoordinator.ts
```diff
@@ -124,28 +124,29 @@ export class PlotCoordinator extends EventEmitter<EventTypes> {
...
+      this.#currentValuesByConfigIndex = [];
       this.#lastSeekTime = lastSeekTime;
     }
 
-    this.#seriesPaths.forEach((path, idx) => {
-      if (!path) {
-        return;
+    for (const seriesItem of this.#series) {
+      if (seriesItem.timestampMethod === "headerStamp") {
+        // We currently do not support showing current values in the legend for header.stamp mode
```
defunctzombie, line 133: Why is that?

packages/studio-base/src/panels/Plot/PlotCoordinator.ts
```diff
@@ -247,9 +231,17 @@ export class PlotCoordinator extends EventEmitter<EventTypes> {
       // using a key instead of the path index lets us preserve loaded data when a path is removed
       const key = `${path.timestampMethod}:${stringifyRosPath(filledParsed)}` as SeriesConfigKey;
 
+      // Keep current values for paths that match existing ones
+      const existingSeries = this.#series.find((series) => _.isEqual(series.key, key));
```
defunctzombie, line 235: consider using `ArrayMap` for `#series` then you can use the map functions to locate by key.

nit: key is a string so you can use `===` comparison (which I guess lodash does for you first?)

defunctzombie (approved):
Still feels like _index_ is being overloaded in a somewhat fragile way when really we want to track an _id_ and provide values by _id_ but this seems fine. Certainly better than before 👍

packages/studio-base/src/panels/Plot/PlotCoordinator.ts
jtbandes: We wouldn’t be able to iterate over message events in the current frame the way we do now. It would require a totally different approach for finding current values. Btw — how does header stamp mode work with live playback?

defunctzombie: The datums are sorted when making the downsampled dataset: https://github.com/foxglove/studio/blob/main/packages/studio-base/src/panels/Plot/builders/TimestampDatasetsBuilderImpl.ts#L368

packages/studio-base/src/panels/Plot/PlotCoordinator.ts
jtbandes: not changing this now because it would require the builder `setSeries` API to change too (or do some kind of map() before passing series to setSeries)

packages/studio-base/src/panels/Plot/PlotCoordinator.ts
jtbandes: updated comment to reference this logic

PR #7347, opened 2024-01-18
Remove play, pause, and close player metrics

jtbandes:
**User-Facing Changes**
None

**Description**
Alternative to https://github.com/foxglove/studio/pull/7339 & #7335
These events are not used and are noisy.


defunctzombie:
@amacneil I brought this up as something that seems of limited app analytics value. We know folks play and pause players so there's no insight to glean from this happening (that I can think of).

amacneil (approved):
let's do it

PR #7348, opened 2024-01-18
Deploy Studio on CF Pages

foxglovenes:
**User-Facing Changes**

None

**Description**

Deploy Studio on CF Pages


.github/workflows/cloudflare-pages.yml
```diff
@@ -0,0 +1,65 @@
...
+jobs:
+  deploy:
+    if: github.event.pull_request || github.event.workflow_run.conclusion == 'success'
+    name: Deploy to Cloudflare Pages
+    runs-on: ubuntu-20.04
+
+    env:
+      # Environment variables that are injected into the build
+      ENABLE_EXPERIMENTAL_COREPACK: 1
+      FOXGLOVE_API_URL: https://api-dev.foxglove.party
```
bryfox, line 26: Should this be part of an OSS deployment? I don't think API config would be needed.

web/.cloudflare/_headers
```diff
@@ -0,0 +1,7 @@
+/*
```
bryfox, line 1: Maybe add a comment with a link to the [format docs](https://developers.cloudflare.com/pages/configuration/headers/) at the top of this file; I didn't register this as a path pattern.

We may want to extend this in the future (e.g., tighter CSP on some pages).

.github/workflows/cloudflare-pages.yml
foxglovenes: I wasn't sure either. They are in the other deployment so I copied them over.

web/.cloudflare/_headers
```diff
@@ -0,0 +1,11 @@
+# Cloudflare Plain Text Headers File
+# Documentation: https://developers.cloudflare.com/pages/configuration/headers/
+
+# Apply to all pages
+/*
+    X-Frame-Options: DENY
+    X-Content-Type-Options: nosniff
+    Referrer-Policy: origin
```
jtbandes, line 8: Looks like this was strict-origin on vercel, any particular reason to change it?

jtbandes (approved):

web/.cloudflare/_headers
foxglovenes: Looks like origin to me? https://github.com/foxglove/studio/blob/main/vercel.json#L10

.github/workflows/cloudflare-pages.yml
foxglovenes: Removed.

PR #7350, opened 2024-01-19
Fix community Slack link

amacneil:
**User-Facing Changes**
None

**Description**
`/join-slack` was an old link but no longer works, `/slack` is correct.


jtbandes (approved):
Shouldn’t we keep the redirect around so old links continue to work?

amacneil:
We had the redirect for like 2 years, at some point it got deleted because I thought nothing used it any more. Turns out I was wrong.

jtbandes:
How about all of these? https://github.com/search?q=org%3Afoxglove%20join-slack&type=code

amacneil:
😭 Fixed

PR #7353, opened 2024-01-22
Message path autocomplete uses an unconstrained popup

defunctzombie:
**User-Facing Changes**
Improve UX for message path autocomplete by allowing the popup to expand wider than the input field to show long completion paths.

**Description**
The current autocomplete component for message path sets the width of the suggestion list to the width of the input. When used on the settings sidebar with long paths this will truncate the topic names and suggestions. Instead of limiting the width of the input, we can allow it to use more space on the page.

This change uses a custom popper component for the MuiAutocomplete to set a `minWidth` property rather than `width` as the built-in behavior does. I've also updated the `renderOption` logic to create a tuple rather than rendering the `MenuItem` component. Since we use a virtualized list, we let the virtualization logic determine when to render the component since it only renders some of the elements in the list avoid building a MenuItem instance for all items since many won't be visible.

Here are some examples with the nuscenes sample data:

| Input | Before | After |
| --- | --- | --- |
| empty | image | image|
| `/imu` | image| image|
| `image_rect` |image |image |

One side-effect of this approach is that it causes the popup to be disjoint from the input field depending on where the input field is in relation to the window. This happens because the width of the popup is such that it is wide enough to fit the longest item in the entire list so when you scroll the width does not change.

Example input popup for a raw message panel on the side of the window.

| Before | After |
| --- | --- |
| image |image|


packages/theme/src/components/MuiAutocomplete.tsx
```diff
@@ -32,7 +32,7 @@ export const MuiAutocomplete: OverrideComponentReturn<"MuiAutocomplete"> = {
         },
       },
       [`.${autocompleteClasses.input}.${inputClasses.inputAdornedEnd}`]: {
-        paddingRight: theme.spacing(0.75),
+        paddingRight: theme.spacing(2),
```
2metres, line 35: Prevent the clear icon from covering the string

packages/studio-base/src/components/Autocomplete/ReactWindowListboxAdapter.tsx
```diff
@@ -2,71 +2,131 @@
...
-  (children as React.ReactChild[]).forEach(
-    (item: React.ReactChild & { children?: React.ReactChild[] }) => {
-      itemData.push(item);
-      itemData.push(...(item.children ?? []));
-    },
+  const { className, ...rest } = other;
+
+  const options = children as ListboxAdapterChild[];
+
+  const longestChild = useMemo(
```
jtbandes, line 62: this kinda assumes you are using a monospace font / no ligatures, right?

packages/studio-base/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -131,13 +85,31 @@ function itemToFzfResult<T>(item: T): FzfResultItem<T> {
...
+// disables the internal filtering of MuiAutocomplete
+//
+// https://mui.com/material-ui/react-autocomplete/#search-as-you-type
+const filterOptions = (options: FzfResultItem[]) => options;
+
+const getOptionLabel = (item: string | FzfResultItem) =>
+  typeof item === "string" ? item : item.item;
+
+// The builtin Popper in MuiAutocomplete uses the width hint from the parent Autocomplete to set
+// the width. We want to set the minWidth to allow the popper to grow wider than the input field width.
```
jtbandes, line 99:
```suggestion
// the width. We want to set the minWidth to allow the popper to grow wider than the input field width,
```

jtbandes (approved):

packages/studio-base/src/components/Autocomplete/ReactWindowListboxAdapter.tsx
defunctzombie: yea - in theory - its definitely "best effort" but works well in practice to avoid the width fluctuating.

jtbandes: If an item does go beyond the width is it scrollable or hoverable to see the full title?

defunctzombie: When you scroll to it - the width will adjust to include it so you will see the full item.

PR #7354, opened 2024-01-22
Custom window controls with initialZoomFactor

wimagguc:
**User-Facing Changes**

None.

**Description**

Allows configuring a custom zoom factor, and piping it all the way down to the AppBars.


packages/studio-base/src/components/AppBar/CustomWindowControls.tsx
```diff
@@ -14,6 +14,7 @@ import Stack from "@foxglove/studio-base/components/Stack";
 export type CustomWindowControlsProps = {
   showCustomWindowControls?: boolean;
   isMaximized?: boolean;
+  initialZoomFactor?: number;
```
jtbandes, line 17: Needs a comment – what is this used for?

jtbandes (approved):
This seems like a fine approach. I am assuming the motivation here is to make the zoom compensation work across restarts of the app?

Did you research how VS Code handles this problem?

wimagguc:
> Did you research how VS Code handles this problem?

Ha thanks for the pointers! I haven't until now, but this was a great idea. Seems like they do a similar counter-zoom trick (this is nicely described in their context menu code: https://github.com/microsoft/vscode/blob/c1cbd4909947fd3b3ba5212ea22f550be8d7cc67/src/vs/workbench/services/contextmenu/electron-sandbox/contextmenuService.ts#L115); and they run a similar even as the on in this PR, if you look for `vscode:notifyZoomLevel`

PR #7355, opened 2024-01-22
Use foxglove fork of comlink

defunctzombie:
**User-Facing Changes**
None

**Description**
Point at the foxglove fork of comlink. This will let us incorporate our changes to comlink without waiting for upstream to accept and release them.

Currently our fork does not have any changes and matches upstream 4.4.1.


achim-k (approved):

defunctzombie:
> will note that having `github:` dependencies has caused vercel build issues in the past, so we may eventually end up needing to publish.

Yea - we will see. I had to force use of `lf` line endings in the tsconfig.json file in the comlink fork to get the yarn windows build to play nice.

PR #7356, opened 2024-01-22
Fix issues with invalid plot series

jtbandes:
**User-Facing Changes**
Not worth calling out

**Description**
Fixed bugs with series that are invalid (fail to parse) causing tooltips and current values to display on a different series. (Similar to #7342 but for invalid series)

Approach: instead of each builder handling its own `data: []` defaults (added in #7342), the builders use the series' `configIndex` to output `datasetsByConfigIndex`. The coordinator fills in missing entries with `data: []` before passing to the renderer.

https://github.com/foxglove/studio/assets/14237/c8f6ba28-8eb9-46d8-a3cc-6b914c7aab93

### DIFF
--- a/packages/studio-base/src/panels/Plot/PlotCoordinator.ts
+++ b/packages/studio-base/src/panels/Plot/PlotCoordinator.ts
@@ -19,7 +19,7 @@ import { Bounds } from "@foxglove/studio-base/types/Bounds";
 import delay from "@foxglove/studio-base/util/delay";
 import { getContrastColor, getLineColor } from "@foxglove/studio-base/util/plotColors";
 
-import { InteractionEvent, Scale, UpdateAction } from "./ChartRenderer";
+import { Dataset, InteractionEvent, Scale, UpdateAction } from "./ChartRenderer";
 import { OffscreenCanvasRenderer } from "./OffscreenCanvasRenderer";
 import {
   CsvDataset,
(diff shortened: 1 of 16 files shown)


packages/studio-base/src/panels/Plot/PlotCoordinator.ts
```diff
@@ -431,7 +431,11 @@ export class PlotCoordinator extends EventEmitter<EventTypes> {
     if (this.#isDestroyed()) {
       return;
     }
-    this.#latestXScale = await this.#renderer.updateDatasets(result.datasets);
+    this.#latestXScale = await this.#renderer.updateDatasets(
+      // Use Array.from to fill in any `undefined` entries with an empty dataset (`map` would not
+      // work for sparse arrays)
+      Array.from(result.datasetsByConfigIndex, (dataset) => dataset ?? { data: [] }),
```
defunctzombie, line 437: hoist the closure since it doesn't need to actually close over anything?

packages/studio-base/src/panels/Plot/builders/TimestampDatasetsBuilder.test.ts
```diff
@@ -618,12 +630,62 @@ describe("TimestampDatasetsBuilder", () => {
...
+
+    builder.setSeries([
+      {
+        configIndex: 3,
+        parsed: parseRosPath("/foo.val")!,
+        color: "red",
+        contrastColor: "blue",
+        enabled: true,
+        timestampMethod: "receiveTime",
+        key: "x" as SeriesConfigKey,
```
defunctzombie, line 653: It is somewhat awkward that we have both a _key_ and a config index. The key was meant to act like the identifier.

jtbandes: The key does not _uniquely_ identify series because you could have multiple series in the config with the same timestampMethod and path.

PR #7357, opened 2024-01-22
Move comlink worker mock to den

defunctzombie:
**User-Facing Changes**
None

**Description**
This same pattern can be applied when testing any class that uses a worker interface via comlink. By putting this in den we can re-use this pattern instead of copy/pasting it.


packages/den/testing/ComlinkWorkerMock.ts
```diff
@@ -0,0 +1,75 @@
...
+ * invoke methods on the underling class.
+ *
+ * ```
+ * Object.defineProperty(global, "Worker", {
+ *   writable: true,
+ *   value: ComlinkWorkerMock(() => new SomeClassThatYouExpose())
+ * });
+ * ```
+ */
+export function ComlinkWorkerMock(makeInstance: () => unknown): ComlinkWorkerConstructor {
```
jtbandes, line 22:
```suggestion
export function makeComlinkWorkerMock(makeInstance: () => unknown): ComlinkWorkerConstructor {
```

jtbandes (approved):

PR #7358, opened 2024-01-22
Use partial subscriptions for index and currentCustom

defunctzombie:
**User-Facing Changes**
None - not worth calling out

**Description**
These modes only need the latest message to build the series. We avoid expensive preloading by using partial subscriptions.


jtbandes (approved):
👏

PR #7359, opened 2024-01-22
Fix dragging topic names with special characters to the Image panel

jtbandes:
**User-Facing Changes**
Fixed an issue with dragging and dropping topics into the Image panel if the topic names contain special characters.

**Description**
The `path` field is quoted, but the Image panel uses an unquoted topic name in its config, so drag & drop would result in an invalid config if the topic name required quotes. This fixes the issue by adding a new `topicName` field to pass through the unmodified topic name.

### DIFF
--- a/packages/studio-base/src/components/PanelExtensionAdapter/types.ts
+++ b/packages/studio-base/src/components/PanelExtensionAdapter/types.ts
@@ -25,6 +25,8 @@ export type DraggedMessagePath = {
   isTopic: boolean;
   /** True if the path represents a primitive value inside a message. */
   isLeaf: boolean;
+  /** The name of the top-level topic being dragged */
+  topicName: string;
 };
 
 export type MessagePathDropStatus = {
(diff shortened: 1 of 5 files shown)


defunctzombie (approved):
It kinda looks like we've already parsed the path somewhere to get the `topicName`? Or maybe that's from the list of topics itself?

Would `rootTopicName` make more sense to align with `rootSchemaName`?

Does seem somewhat awkward to see "isTopic: false" and then you have a topic name - I guess that tells you if the path itself is a topic? Why doesn't the receiver figure that out itself? Should the parsed path AST be sent instead?

jtbandes:
> Should the parsed path AST be sent instead?

Might be nice but the thought was to eventually publish this in the extension API and I believe we haven’t reworked our “RosPath” to a place where we’re comfortable with publishing it.

jtbandes:
> Would rootTopicName make more sense to align with rootSchemaName?

I don't feel too strongly either way. My thinking on this was that `rootSchemaName` makes some sense because you could imagine a schema for each sub-message, and this one refers to the root. On the other hand, with topics, there are no "nested topics" – there is just one topic at the root level.

PR #7360, opened 2024-01-22
Plot: don't subscribe to empty field names

defunctzombie:
**User-Facing Changes**
None

**Description**
The underlying logic would likely ignore these but it is good to avoid creating a subscription for an empty field name.


jtbandes (approved):
Interesting that this needs to be handled separately from the case where the message path is completely empty. I suppose it makes some sense though.

PR #7361, opened 2024-01-22
Use `ComlinkWrap` instead of `Comlink.wrap` to expose explicit `dispose`

snosenzo:
**User-Facing Changes**

n/a

**Description**

The purpose of this change is to attempt to reduce memory-leaks from comlink usage, which can hold onto references if the worker is terminated before async calls have been resolved as seen here: https://github.com/GoogleChromeLabs/comlink/issues/655.

Testing has been completed and I wasn't able to find any obvious memory leaks from switching between plot and image tabs during playback and switching file sources. 

FG-6301


packages/studio-base/src/players/IterablePlayer/WorkerIterableSource.ts
```diff
@@ -28,29 +29,35 @@ type ConstructorArgs = {
...
-  #worker?: Comlink.Remote<WorkerIterableSourceWorker>;
+  #sourceWorkerRemote?: Comlink.Remote<WorkerIterableSourceWorker>;
+  #disposeRemote?: () => void;
 
   public constructor(args: ConstructorArgs) {
     this.#args = args;
   }
 
   public async initialize(): Promise<Initalization> {
+    if (this.#sourceWorkerRemote != undefined && this.#disposeRemote != undefined) {
```
defunctzombie, line 40: nit: Change this if statement and body to `this.#disposeRemote?.()`

defunctzombie (approved):
Once this merges I think we'll also want to update the user-script use of comlink as well.

nit: `den/worker` might be a better home than `den/async` but I don't feel too strongly about it if async felt more natural to you.

PR #7362, opened 2024-01-22
Reduce downsamples and re-renders for plot

defunctzombie:
**User-Facing Changes**
Plot performance improvements

**Description**
Prior to this change, any plot panel would re-downsample its dataset on any player state update regardless of whether it had new data to render or not. This change updates the `handlePlayerState` method of `IDatasetsBuilder` to return the number of datums added in addition to the range. The coordinator can avoid downsampling if there are no datums read.

This change also splits the downsample dispatch from the rendering of the downsampled datasets. Since these are powered by separate workers we can start work on a new downsample request while rendering the old.


defunctzombie:
@jtbandes @snosenzo I have to port the remaining DatasetsBuilders to the new interface but I wanted to get your first impressions of this approach for some light optimization to reduce plot rendering. It felt natural to return the number of datums that would be added via the `handlePlayerState` call but maybe you will see some other way that feels nicer.

I could also see a path here where the datasets builder dispatches calls to render itself with the latest viewport or downsampled datasets. That would require changing how the communication happens but would avoid a hop through the main thread. I see this change as a small improvement towards that larger change - but if we want to tackle the larger change I can see that happening too.

packages/studio-base/src/panels/Plot/PlotCoordinator.ts
```diff
@@ -159,8 +153,34 @@ export class PlotCoordinator extends EventEmitter<EventTypes> {
...
+    const newRange = handlePlayerStateResult.range;
+
+    // If the range has changed we will trigger a render to incorporate the new range
+    if (!_.isEqual(this.#datasetRange, newRange)) {
+      this.#datasetRange = handlePlayerStateResult.range;
+      this.#queueDispatchRender();
+    }
+
+    // The underlying dataset may have changed so we downsample
+    if (handlePlayerStateResult.numDatums > 0) {
```
snosenzo, line 181: I think since numDatums isn't actually used except for its existence that it might be more worth it to return `needsDownsample` from `handlePlayerState`

defunctzombie: `needsDownsample` implies an action by the caller but I understand the suggestion. I've changed this to `datasetsChanged` boolean.

packages/studio-base/src/panels/Plot/builders/IDatasetsBuilder.ts
```diff
@@ -71,6 +71,13 @@ export type GetViewportDatasetsResult = {
   pathsWithMismatchedDataLengths: ReadonlySet<string>;
 };
 
+export type HandlePlayerStateResult = {
+  /** The x-axis range of the dataset if it is known */
+  range?: Bounds1D;
```
snosenzo, line 76: should probably be `readonly`

snosenzo (approved):
a little worried about the passing of mutable bounds references that are being set to private member variables, but otherwise looks good

jtbandes (approved):

packages/studio-base/src/panels/Plot/PlotCoordinator.ts
```diff
@@ -159,8 +153,34 @@ export class PlotCoordinator extends EventEmitter<EventTypes> {
...
+      this.#datasetRange = undefined;
+      this.#queueDispatchRender();
+      return;
+    }
+
+    const newRange = handlePlayerStateResult.range;
+
+    // If the range has changed we will trigger a render to incorporate the new range into the chart
+    // axis
+    if (!_.isEqual(this.#datasetRange, newRange)) {
```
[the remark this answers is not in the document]
defunctzombie: Good call-out but in this case I am ok with the strict equality. What I am checking for is whether the exact values are already set on #datasetRange. Any deviation and I do want to update the coordinator's range.

PR #7371, opened 2024-01-23
Move parseRosPath to @foxglove/message-path

defunctzombie:
**User-Facing Changes**
None

**Description**
Move `parseRosPath.ts` and `constants.ts` to a new module `@foxglove/message-path`.

- Rename `parseRosPath` -> `parseMessagePath`
- Rename `RosPath` type -> `MessagePath`
- Rename `RosPrimitive` type -> `PrimitiveType`
- Rename `constants.ts` -> `types.ts` in the new module

Note: I've removed the _memoize_ wrapping for `parseMessagePath`. I don't think this is the appropriate level to memoize things but I can put it back if we want to minimize change. I don't know how we will prove or disprove this is important to memoize.


packages/studio-base/src/components/MessagePathSyntax/messagePathsForDatatype.ts
```diff
@@ -40,8 +38,24 @@ const STRUCTURE_ITEM_INTEGER_TYPES = [
...
+    case "uint8":
+    case "int16":
+    case "uint16":
+    case "int32":
+    case "uint32":
+    case "int64":
+    case "uint64":
+    case "float32":
+    case "float64":
+    case "string":
```
jtbandes, line 54: how will we remember to keep this up to date with the PrimitiveType definition?

jtbandes (approved):

packages/studio-base/src/components/MessagePathSyntax/messagePathsForDatatype.ts
defunctzombie: 🤔 if I do `switch (val as PrimitiveType)` typescript will complain about missing case statements? Does that seem good or should I move this back to the package?

jtbandes: I suppose that's fine as well

jtbandes (approved):

PR #7372, opened 2024-01-23
CompressedVideo support

longgliveopensource:
**User-Facing Changes**

**Description**
After reading the announcement of h264 support I couldn't find any of that code in the studio side of the project. After searching a lot I found this PR https://github.com/MetaverseIndustries/foxglove-studio/pull/1 but strangely enough the branches were removed :thinking: Then I found this commit https://github.com/foxglove/studio/commit/4975e65041299cc4cf04305f254c65eb12e48b5e with a (almost) working code.
 
Don't know what happened but here it is the backport of that code to the current codebase. Hope it helps!

I tested this with the following rosbag:
[compressed_video_rosbag.zip](https://github.com/foxglove/studio/files/14019555/compressed_video_rosbag.zip)

<!-- add relevant metric tracking for experimental / new features --

### DIFF
--- a/packages/den/package.json
+++ b/packages/den/package.json
@@ -17,6 +17,7 @@
     "xacro-parser": "0.3.9"
   },
   "devDependencies": {
-    "@foxglove/tsconfig": "2.0.0"
+    "@foxglove/tsconfig": "2.0.0",
+    "@types/dom-webcodecs": "0.1.7"
   }
 }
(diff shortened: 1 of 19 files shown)


amacneil:
Thanks for the write up. This is an early version of the code - we changed it quite a bit since then, but we since decided to ship h264 support as a closed source add-on (more info [here](https://github.com/foxglove/studio#open-source)). You can use h264 support for free in the official builds available at https://studio.foxglove.dev/ and https://foxglove.dev/download.

You are of course free to maintain your own fork, but we do not plan to add this to open source.

PR #7375, opened 2024-01-24
Move setSeries to an update action

defunctzombie:
**User-Facing Changes**
None

**Description**
The TimestampDatasetsBuilder and CustomDatasetsBuilder need to send the series configuration to their respective workers. Prior to this change this was done with a call to `setSeries` on the exposed RPC interface with a _void_ since this call happened in an async function that did not wait. This change moves the call into an update action so it can be dispatched with an await like the other pending update actions and be captured by whatever error mechanism captures the async function containing the dispatched calls.

---
I think this is a good change given our current dispatch structure but it did make me realize that we could use the browser postMessage calls as our queue manager rather than async/await these dispatches over RPC. Practically we don't care about the response value from the receiver - only that we sent it to the receiver. I suppose there is an aspect around getting the latest downsampled datasets that cares that the receiver has gotten all the latest updates (at least for series config). Nothing I would change in our structure yet but something to think about with respect to how we leverage workers.

### DIFF
--- a/packages/studio-base/src/panels/Plot/builders/CustomDatasetsBuilder.ts
+++ b/packages/studio-base/src/panels/Plot/builders/CustomDatasetsBuilder.ts
@@ -213,7 +213,10 @@ export class CustomDatasetsBuilder implements IDatasetsBuilder {
       };
     });
 
-    void this.#datasetsBuilderRemote.setSeries(series);
+    this.#pendingDataDispatch.push({
+      type: "update-series-config",
+      seriesItems: series,
+    });
   }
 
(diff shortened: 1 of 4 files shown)


snosenzo (approved):
Seems fine. The only potential issue that came to mind was that maybe other `pendingDataDispatch`s depend on that series needing to exist  or have been updated before getting there, but now they come out of order

jtbandes (approved):
I'm not quite sure why this is called `pendingDataDispatch` & `UpdateDataAction` since it doesn't seem to be only for "data"

PR #7376, opened 2024-01-24
Buffer messages from mcap sources as raw Uint8Arrays 

achim-k:
**User-Facing Changes**
Store buffered messages as raw typed arrays to reduce memory pressure

**Description**
This PR aims to reduce OOM crashes by changing the playback buffer to store messages as raw Uint8Arrays. Storing messages as Uint8arrays reduces heap memory pressure as they do not count towards the heap memory size compared to how deserialized messages (plain JS objects) would. A positive consequence of this is, that we can buffer much more messages than we were previously able to.

Whereas previously, messages would be deserialized in the worker, the worker now sends (or rather transfers) undeserialized messages to the main thread where they are stored in the playback buffer and deserialized on the fly. The only downside of this approach is, that the main thread now has to pay the cost of message deserialization. However, this cost might be compensated by more efficient webworker communication (due to the transferring of array buffers) and is worth to pay in return for less OOM crashes and a larger playback buffer.

Note that for now the above was only implemented for MCAP related sources (local/remote mcap files, nuscenes)

Supersedes #7267
Resolves FG-6094


packages/studio-base/src/players/IterablePlayer/IBufferedIterableSource.ts
```diff
@@ -0,0 +1,43 @@
...
+  ranges: Range[];
+  cacheSizeInBytes: number;
+};
+
+export interface IBufferedIterableSource extends IIterableSource {
+  initialize(options?: InitalizationOptions): Promise<Initalization>;
+
+  stopProducer(): Promise<void>;
+  getCacheSize(): Promise<number>;
+  loadedRanges(): Promise<Range[]>;
```
achim-k, line 37: Not sure if these two methods are really necessary. I think the `onLoadedRanges` listener is maybe sufficient. Also, these two methods could probably be merged into a single one.

packages/studio-base/src/players/IterablePlayer/IBufferedIterableSource.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export type BufferedRanges = {
+  ranges: Range[];
+  cacheSizeInBytes: number;
+};
+
+export interface IBufferedIterableSource extends IIterableSource {
+  initialize(options?: InitalizationOptions): Promise<Initalization>;
+
+  stopProducer(): Promise<void>;
```
achim-k, line 35: I added this because this is called by the IterablePlayer. I don't think that this is really needed, creating a new message iterator or terminating the source will implicitly do this.

packages/studio-base/src/players/IterablePlayer/IteratorCursor.ts
```diff
@@ -72,22 +77,28 @@ class IteratorCursor implements IMessageCursor {
...
+          transfers.push(result.msgEvent.message.buffer);
+        }
+        if (compare(result.msgEvent.receiveTime, cutoffTime) > 0) {
+          break;
+        }
       }
     }
-    return results;
+
+    return Comlink.transfer(results, transfers); // ACHIM: This has to be done differently. If this is not used in a worker then the transfer cache will explode at some point -> OOM
```
achim-k, line 90: I have to get rid of comlink in this class. The class musn't care in what context it is being used.

defunctzombie:
Is this a replacement for https://github.com/foxglove/studio/pull/7267 ?

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
```diff
@@ -1076,7 +1084,8 @@ export class IterablePlayer implements Player {
     await this.#blockLoader?.stopLoading();
     await this.#blockLoadingProcess;
     await this.#bufferedSource.stopProducer();
-    await this.#bufferedSource.terminate();
+    await this.#bufferedSource.terminate?.();
+    await this.#iterableSource.terminate?.();
```
defunctzombie, line 1088: The source is terminated a few lines lower.

defunctzombie (commented):
This is great - I am amused at how the IIterableSource was made generic and could serve the purpose here. It does feel like it might be entering _too clever_ territory but I can see us being ok with this approach to get this shipped since it does slot into the machinery. I like that the places that expect a "raw" source have a type of `IIterableSource<Uint8Array>` but the constraint does not go the other way - i.e. the BlockLoader can't say it needs a "Parsed" source rather than a raw source. So it might be worth thinking about how to make the intent more clear.

I am a little less clear on the IBufferedIterableSource and how that wraps/semi-wraps but still is a source? I think there's something to re-visit there with maybe naming or semantics.

The Worker boundary is another place where maybe the re-use of the same class is re-used for raw and parsed is maybe a bit much since the implementation is a very different path for the two and you don't want to mix them up.

An upside of this approach with the sources here is that the diff is quite minimal - a source can be ported to produce raw output and wrapped in the deserialization layer. So that seems like a good way to get to where we want to be. Have a think of some of the feedback - none of which seems like a blocker or hard to cleanup in follow-up.

packages/studio-base/src/players/IterablePlayer/IIterableSource.ts
```diff
@@ -71,10 +71,10 @@ export type MessageIteratorArgs = {
  * The source may return stamp results to indicate to callers that it has read through some time
  * when there are no message events available to indicate the time is reached.
  */
-export type IteratorResult =
+export type IteratorResult<MessageType = unknown> =
   | {
       type: "message-event";
-      msgEvent: MessageEvent;
+      msgEvent: MessageEvent<MessageType>;
```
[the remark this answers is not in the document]
achim-k: Not sure, I think it's better if the consumer has the guarantee that it only has to deal with one kind of message event.

packages/studio-base/src/players/IterablePlayer/Mcap/McapIndexedIterableSource.ts
```diff
@@ -16,18 +15,12 @@ import {
...
-  Topic,
-  TopicStats,
-} from "@foxglove/studio-base/players/types";
+import { PlayerProblem, Topic, TopicStats } from "@foxglove/studio-base/players/types";
 import { RosDatatypes } from "@foxglove/studio-base/types/RosDatatypes";
 
 const log = Logger.getLogger(__filename);
 
-export class McapIndexedIterableSource implements IIterableSource {
+export class McapIndexedIterableSource implements IIterableSource<Uint8Array> {
```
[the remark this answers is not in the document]
achim-k: One way to make it less fragile would be to use stricter types for deserialized message sources, e.g. instead of `IIterableSource<unknown>` we could use `IIterableSource<Record<string, unknown>>`

packages/studio-base/src/players/IterablePlayer/IBufferedIterableSource.ts
```diff
@@ -0,0 +1,31 @@
...
+  /** exclusive */
+  end: number;
+};
+
+export type BufferedRanges = {
+  ranges: Range[];
+  cacheSizeInBytes: number;
+};
+
+export interface IBufferedIterableSource<MessageType = unknown>
```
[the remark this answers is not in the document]
achim-k: > Or maybe it has a strange name. It isn't really a "source" like the other sources and I think led to some of my initial confusion when reading the PR cause I had a preconceived notion of what a source is.

In what way is it different from other sources? We did have the `BufferedIterableSource` class before, that's where this interface more or less stems from.

packages/studio-base/src/dataSources/McapLocalDataSourceFactory.ts
```diff
@@ -35,9 +43,18 @@ class McapLocalDataSourceFactory implements IDataSourceFactory {
       initArgs: { file },
     });
 
+    const rawBufferedSource = new BufferedIterableSource<Uint8Array>(workerSource, {
+      readAheadDuration: { sec: 120, nsec: 0 },
+      maxCacheSizeBytes: 2 * GIGABYTE_IN_BYTES,
+    });
+
+    const source = new DeserializingIterableSource(workerSource);
+    const bufferedSource = new DeserializingBufferedIterableSource(rawBufferedSource);
```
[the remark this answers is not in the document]
achim-k: I basically need two deserializing sources here as we cannot have one message iterator for both block loader and playback at the same time. This is also the reason why I have to call `initialized` on both deserializing sources (as they have to initialize decoders, etc.)
The `DeserializingBufferedIterableSource` forwards all methods exposed by the `IbufferedIterableSource` interface to the underlying buffered source. 

```
┌────────────┐   ┌────────────────────┐
│ BlockLoader├──►│deserializing source├────────────────────────┐
│            │   └────────────────────┘                        │   ┌──────────┐
│ It. player │                                                 ├──►│raw source│
│            │   ┌────────────────────┐  ┌───────────────────┐ │   └──────────┘
│   Playback ├──►│deserializing source├─►│raw buffered source├─┘
└────────────┘   └────────────────────┘  └───────────────────┘
```

packages/studio-base/src/players/IterablePlayer/IBufferedIterableSource.ts
```diff
@@ -0,0 +1,31 @@
...
+export type Range = {
+  /** inclusive */
+  start: number;
+  /** exclusive */
+  end: number;
+};
+
+export type BufferedRanges = {
+  ranges: Range[];
+  cacheSizeInBytes: number;
```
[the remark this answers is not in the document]
achim-k: The IterablePlayer uses this for populating the `memoryInfo` field in its progress updates (for showing cache size in the perf side bar).

`BufferInfo` is a better name, agreed.

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
```diff
@@ -458,6 +460,8 @@ export class IterablePlayer implements Player {
     this.#queueEmitState();
 
     try {
+      const initResult = await this.#iterableSource.initialize();
+      this.#bufferedSource.init(initResult);
```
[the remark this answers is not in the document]
achim-k: Agreed, the double init is weird. I currently need this when two deserializing sources are being used. I will see if I can avoid that.

packages/studio-base/src/players/IterablePlayer/WorkerIterableSourceWorker.ts
```diff
@@ -17,11 +18,15 @@ import type {
 } from "./IIterableSource";
 import { IteratorCursor } from "./IteratorCursor";
 
-export class WorkerIterableSourceWorker implements IIterableSource {
-  protected _source: IIterableSource;
+export class WorkerIterableSourceWorker<MessageType = unknown>
```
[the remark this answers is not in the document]
achim-k: > Like it would be "wrong" to use isRawSource: true without the Uint8Array message type?

Yeah it would be wrong. Not sure how I could make this better with typescript, there is no such think as template specialization AFAIK

packages/studio-base/src/players/types.ts
```diff
@@ -210,6 +210,12 @@ export type Topic = {
   schemaName: string | undefined;
   // Name of the topic before topic aliasing, if any.
   aliasedFromName?: string;
+  /** Schema data */
```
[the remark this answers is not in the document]
defunctzombie: I think a different type is public: https://github.com/foxglove/studio/blob/main/packages/studio/src/index.ts#L42 and your use of the Topic here is specific to player interfaces and is ok.

packages/studio-base/src/players/IterablePlayer/WorkerIterableSourceWorker.ts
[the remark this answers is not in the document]
defunctzombie: Amusingly I think there is.

packages/studio-base/src/players/IterablePlayer/BufferedIterableSource.ts
```diff
@@ -75,27 +72,29 @@ class BufferedIterableSource extends EventEmitter<EventTypes> implements IIterab
...
-    this.#source.on("loadedRangesChange", () => this.emit("loadedRangesChange"));
   }
 
   public async initialize(): Promise<Initalization> {
     this.#initResult = await this.#source.initialize();
     return this.#initResult;
   }
 
-  async #startProducer(args: MessageIteratorArgs): Promise<void> {
+  public init(initResult: Initalization): void {
```
snosenzo, line 92: unclear of the difference between `init()` and `initialize()` methods. and seems strange that both are happening? can we consolidate or better differentiate here?

packages/studio-base/src/players/IterablePlayer/IBufferedIterableSource.ts
```diff
@@ -0,0 +1,31 @@
...
+};
+
+export interface IBufferedIterableSource<MessageType = unknown>
+  extends IIterableSource<MessageType> {
+  // Initialize the source without initializing the underyling source which may already have been initialized.
+  init(initResult: Initalization): void;
```
snosenzo, line 22: Would `sourceInitialized` or `onSourceInitialized` be a better name here?

packages/studio-base/src/players/IterablePlayer/BufferedIterableSource.ts
snosenzo: oop just read roman's comment about the double init elsewhere -- I think that mostly captures what I was getting at

packages/studio-base/src/players/IterablePlayer/ComlinkTransferIteratorCursor.ts
```diff
@@ -0,0 +1,57 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+import * as Comlink from "comlink";
+
+import { Time } from "@foxglove/studio";
+
+import type { IMessageCursor, IteratorResult } from "./IIterableSource";
+
+class ComlinkTransferIteratorCursor implements IMessageCursor<Uint8Array> {
```
defunctzombie, line 10: Comment on what this is please and any notable things for callers/users to be aware of and maintainers to keep in mind.

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
```diff
@@ -184,7 +187,18 @@ export class IterablePlayer implements Player {
     const { metricsCollector, urlParams, source, name, enablePreload, sourceId } = options;
 
     this.#iterableSource = source;
-    this.#bufferedSource = new BufferedIterableSource(source);
+
+    const GIGABYTE_IN_BYTES = 1024 * 1024 * 1024;
+    this.#bufferedSource =
+      source.sourceType === "raw"
```
defunctzombie, line 193: nit: "serialized" is maybe the opposite of "deserialized"?

"raw" seems fine too but is not a word I would typically use to describe the serialized message data. I would say it is "serialized".

packages/studio-base/src/players/IterablePlayer/IIterableSource.ts
```diff
@@ -205,3 +213,8 @@ export type IterableSourceInitializeArgs = {
     auth?: string;
   };
 };
+
+export type IRawIterableSource = IIterableSource<Uint8Array> & { sourceType: "raw" };
```
defunctzombie, line 217: Comment on what these are and the context in which one might use raw vs deserialized.

nit: same nit comment as I made in IterablePlayer about "serialized" maybe being a nice opposite to "deserialized" even if it is a longer word :D

packages/studio-base/src/players/IterablePlayer/CachingIterableSource.ts
```diff
@@ -112,7 +116,7 @@ class CachingIterableSource extends EventEmitter<EventTypes> implements IIterabl
 
   public async terminate(): Promise<void> {
     this.#cache.length = 0;
-    this.#cachedTopics.clear();
+    this.#cachedTopics = new Map();
```
defunctzombie, line 119: Anything notable about making a new map vs clearing the old?

packages/studio-base/src/players/IterablePlayer/CachingIterableSource.ts
```diff
@@ -681,6 +687,34 @@ class CachingIterableSource extends EventEmitter<EventTypes> implements IIterabl
...
+    unsubscribe: () => void;
+  } {
+    const handler = () => {
+      bufferInfoChangeHandler({
+        cacheSizeInBytes: this.getCacheSize(),
+        loadedRanges: this.loadedRanges(),
+      });
+    };
+
+    this.on("loadedRangesChange", handler);
```
defunctzombie, line 710: Since this internally uses an event emitter how come external callers can't use it and instead go through a `subscribeToBufferingChanges` interface?

packages/studio-base/src/players/IterablePlayer/CachingIterableSource.ts
achim-k: This was a consequence of introducing `Immutable`. Maybe that is no longer the case and I can revert it to keep the diff smaller.

packages/studio-base/src/players/IterablePlayer/CachingIterableSource.ts
achim-k: They still could. Maybe I remove the `IBufferedIterableSource` interface from this file or I remove the event emitter

achim-k:
In the latest changes I managed to remove the `IBufferedIterableSource` interface again and keep the buffered / caching iterable source more or less as they were (minus the message type template changes).

defunctzombie:
👍 Would be nice to also rename `IRawIterableSource` to `ISerializedIterableSource` https://github.com/foxglove/studio/blob/main/packages/studio-base/src/players/IterablePlayer/IIterableSource.ts#L221 to match the new `serialized` terminology.

achim-k:
> 👍 Would be nice to also rename `IRawIterableSource` to `ISerializedIterableSource` https://github.com/foxglove/studio/blob/main/packages/studio-base/src/players/IterablePlayer/IIterableSource.ts#L221 to match the new `serialized` terminology.

Ah, good point :sweat_smile: I'll do that when adapting the other sources.

PR #7404, opened 2024-01-26
Move wrapPlayer out of useMemo in PlayerManager

defunctzombie:
**User-Facing Changes**
None

**Description**
The `wrapPlayer` call runs in a memo which executes during the _render_ phase in react. However, `wrapPlayer` may run logic which invokes a state update. React does not like state updates of a component to occur when another component is rendering. This moves the construction of the wrapped player to a callback that builds the topicAliasPlayer and the wrapped player and then sets the state.

This avoids the error for updating a component while rendering.


jtbandes (commented):
> wrapPlayer may run logic which invokes a state update

Why is that?

defunctzombie:
> Why is that?

See AppContextProvider line 110.

packages/studio-base/src/components/PlayerManager.tsx
```diff
@@ -53,17 +53,28 @@ export default function PlayerManager(props: PropsWithChildren<PlayerManagerProp
   const analytics = useAnalytics();
   const metricsCollector = useMemo(() => new AnalyticsMetricsCollector(analytics), [analytics]);
 
-  const [basePlayer, setBasePlayer] = useState<Player | undefined>();
+  const [playerState, setPlayerState] = useState<
```
jtbandes, line 56: Would recommend renaming this since "PlayerState" is already a core concept in the app.

jtbandes (approved):
Approach seems fine. Another thing that I wondered about is whether we could move the side effects out of `wrapPlayer` and into `useEffect`s? Maybe this is more straightforward though.

PR #7405, opened 2024-01-26
Suppress webpack client overlay for webworker importScript errors

defunctzombie:
**User-Facing Changes**
None

**Description**
When a webworker is terminated, any pending `importScript` calls are cancelled by the browser. These appear in the devtools network tab as "(cancelled)" and bubble up to the parent page as errors which trigger `window.onerror`.

Webpack devserver sees these uncaught errors and shows the error overlay. This is a false positive since the worker is already terminated. This change suppresses the overlay in this scenario.

**Why are we seeing this so often now?**

We are seeing more instances of this lately due to the plot panel launching workers for each panel and launching the worker in a `useMemo()` hook: https://github.com/foxglove/studio/blob/main/packages/studio-base/src/panels/Plot/Plot.tsx#L264 In react strict mode, `useMemo` is run twice. Since we only store a reference to output of the memo, the first run of the memo is discarded with nothing having a reference. The internal finalization registry of the `TimestampDatastsBuilder` disposes of the un-used instance and terminates the worker. If this happens to race with the worker loading it's scripts, the termination of the worker will trigger such `importScript` errors.

### DIFF
--- a/packages/studio-web/src/webpackConfigs.ts
+++ b/packages/studio-web/src/webpackConfigs.ts
@@ -64,6 +64,38 @@ export const devServerConfig = (params: ConfigParams): WebpackConfiguration => (
       "cross-origin-opener-policy": "same-origin",
       "cross-origin-embedder-policy": "credentialless",
     },
+
+    client: {
+      overlay: {
+        runtimeErrors: (error) => {
+          // Suppress overlays for importScript errors from terminated webworkers.
+          //
+          // When a webworker is terminated, any pending `importScript` calls are cancelled by the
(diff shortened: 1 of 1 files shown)


jtbandes (approved):
Do other browsers have the same behavior and do they use the same error message?

defunctzombie:
> Do other browsers have the same behavior

What are these other browsers you speak of?!

defunctzombie:
Merging this to get it resolved for Chrome - our primary dev browser. I will review what happens on Safari and Firefox.

PR #7407, opened 2024-01-29
Fix collada textures not being loaded

achim-k:
**User-Facing Changes**
Fix collada textures not being loaded

**Description**
We were revoking temporary object URLs to early, resulting in errors when loading these URLs. This PR ties the lifetime of these temporary object URLs to the lifetime of the ModelCache.

### DIFF
--- a/packages/studio-base/src/panels/ThreeDeeRender/ModelCache.ts
+++ b/packages/studio-base/src/panels/ThreeDeeRender/ModelCache.ts
@@ -50,7 +50,7 @@ export class ModelCache {
   #models = new Map<string, Promise<LoadedModel | undefined>>();
   #edgeMaterial: THREE.Material;
   #fetchAsset: BuiltinPanelExtensionContext["unstable_fetchAsset"];
-
+  #colladaTextureObjectUrls = new Map<string, string>();
   #dracoLoader?: DRACOLoader;
 
   public constructor(public readonly options: ModelCacheOptions) {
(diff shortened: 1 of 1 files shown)


snosenzo (approved):
Seems fine, though I do want to make sure: there's no problem with these models only being cleaned up on panel unmount? They shouldn't grow over time otherwise?

achim-k:
> Seems fine, though I do want to make sure: there's no problem with these models only being cleaned up on panel unmount? They shouldn't grow over time otherwise?

No I don't think that there is an issue, especially since models are cached, so temporary object URLs are created only once for every model URL

packages/studio-base/src/panels/ThreeDeeRender/ModelCache.ts
```diff
@@ -220,25 +219,21 @@ export class ModelCache {
         const objectUrl = URL.createObjectURL(
           new Blob([textureAsset.data], { type: textureAsset.mediaType }),
         );
-        textureUrls.set(textureUrl, objectUrl);
+        this.#colladaTextureObjectUrls.set(textureUrl, objectUrl);
```
jtbandes, line 222: Should we check if one already exists before creating a new one here?

achim-k: Good remark, I'll create a follow-up PR for this

PR #7408, opened 2024-01-29
Fix URDF meshes not loading when connected to rosbridge

achim-k:
**User-Facing Changes**
Fix URDF meshes not loading when connected to rosbridge

**Description**
Instead of only trying one method for fetching assets with `package://` URLs, other valid methods are also tried before an error is raised.

### DIFF
--- a/packages/studio-base/src/components/MessagePipeline/store.ts
+++ b/packages/studio-base/src/components/MessagePipeline/store.ts
@@ -176,16 +176,16 @@ export function createMessagePipelineStore({
           if (player?.fetchAsset) {
             try {
               return await player.fetchAsset(uri);
-            } catch (err) {
-              if (canBuiltinFetchPkgUri) {
-                // Fallback to a builtin _fetch_ call if the asset couldn't be loaded through the player.
-                return await builtinFetch(uri, options);
-              }
-              throw err; // Bail out otherwise.
+            } catch (_err) {
+              // Do nothing here as one of the fallback methods below might work.
             }
-          } else if (canBuiltinFetchPkgUri) {
+          }
+
+          if (canBuiltinFetchPkgUri) {
             return await builtinFetch(uri, options);
-          } else if (
+          }
+
+          if (
             pkgName &&
             options?.referenceUrl != undefined &&
             !options.referenceUrl.startsWith("package://") &&
@@ -200,6 +200,8 @@ export function createMessagePipelineStore({
               options.referenceUrl.slice(0, options.referenceUrl.lastIndexOf(pkgName)) + pkgPath;
             return await builtinFetch(resolvedUrl, options);
           }
+
+          throw new Error(`Failed to load asset ${uri}`);
         }
 
         // Use a regular fetch for all other protocols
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/components/MessagePipeline/store.ts
```diff
@@ -176,16 +176,16 @@ export function createMessagePipelineStore({
...
             try {
               return await player.fetchAsset(uri);
-            } catch (err) {
-              if (canBuiltinFetchPkgUri) {
-                // Fallback to a builtin _fetch_ call if the asset couldn't be loaded through the player.
-                return await builtinFetch(uri, options);
-              }
-              throw err; // Bail out otherwise.
+            } catch (_err) {
+              // Do nothing here as one of the fallback methods below might work.
```
achim-k, line 180: In this specific case, the presence of `player.fetchAsset` cannot be used to determine if the player is able to fetch assets as the TopicAliasing player might be inbetween which always implements `fetchAsset`. One option might be to add `fetchAsset` as an additional player capability.

snosenzo: I thought this was already the case?

achim-k: It's a capability of the foxglove websocket player, but we haven't added it as a [possible player capability](https://github.com/foxglove/studio/blob/b6376c6789c5255d01e108cb93a7c48eeb14fa2d/packages/studio-base/src/players/types.ts#L321-L342). Should I do that?

snosenzo: I think that makes sense. Is it an accurate way to represent that other `Player`s have this capability? Maybe someone else has differing opinions on adding this to the API.

achim-k: @jtbandes @defunctzombie thoughts?

jtbandes: > the presence of player.fetchAsset cannot be used to determine if the player is able to fetch assets

this seems kinda sad... was it marked optional in the interface in the hopes that it would be a way of determining whether the player can fetch assets?

achim-k: Not sure, I think John implemented this

defunctzombie: yea - I think that was the original thinking but we should use the player capability.

FWIW - it would not be hard to update the topic aliasing player (and user script player) to avoid having this function if the wrapped player does not have it. That's another thing we can do.

Even if we did that - the proposed fall-through changes here seem ok? What would be a reason why we would not want to let them run?

achim-k: Update: I just realized that the [`asset` player capability](https://github.com/foxglove/studio/blob/544fbc0239b2d14bd1261585f623c55e33768f24/packages/studio-base/src/players/types.ts#L325-L326) was already there :man_facepalming:

achim-k: Update: I just realized that the [`asset` player capability](https://github.com/foxglove/studio/blob/544fbc0239b2d14bd1261585f623c55e33768f24/packages/studio-base/src/players/types.ts#L325-L326) was already there :man_facepalming:

snosenzo (approved):

PR #7411, opened 2024-01-31
Updated Chinese translation

Woodii1998:
**User-Facing Changes**

- Added some Chinese translations
- Adding [localization](https://mui.com/material-ui/guides/localization/) to mui components 

**Description**


packages/theme/src/createMuiTheme.ts
```diff
@@ -19,11 +21,26 @@ declare module "@mui/material/styles" {
...
+    case "en":
+      return enUS;
+    default:
+      return enUS;
+  }
+}
+
+export const createMuiTheme = (themePreference: ThemePreference, lang: string): Theme =>
+  createTheme(
+    _.merge(getMaterialLocale(lang), {
```
defunctzombie, line 39: Why do we want to merge the theme with the translation string map/dictionary objects?

jtbandes: If you look at what these objects contain they seem to be some component prop overrides for specific languages' strings: https://github.com/mui/material-ui/blob/master/packages/mui-material/src/locale/index.ts#L4045

packages/studio-base/src/i18n/zh/topicList.ts
```diff
@@ -6,11 +6,11 @@ import { TypeOptions } from "i18next";
...
   anErrorOccurred: "发生错误",
   clearFilter: "清除筛选",
-  copyMessagePath: undefined,
-  copyMessagePaths: undefined,
-  copySchemaName: undefined,
-  copyTopicName: undefined,
-  copyTopicNames: undefined,
+  copyMessagePath: "复制消息路径",
+  copyMessagePaths: "复制所选的消息路径",
+  copySchemaName: "复制 schema 名称",
```
jtbandes, line 11: I wonder if `schema` should be translated the same as 数据类型 ("datatype") below? Maybe there is a better word for this that could be used in both places?

packages/theme/src/createMuiTheme.ts
jtbandes: `createTheme` can take multiple arguments and merge them together, but I believe this order of merging first makes more sense so that our app's custom theme doesn't get overridden by the locale overrides.

jtbandes (approved):
Thanks!

packages/studio-base/src/i18n/en/general.ts
```diff
@@ -6,4 +6,7 @@
 export const general = {
   foxglove: "Foxglove",
   learnMore: "Learn more",
+  on: "On",
+  off: "Off",
+  auto: "Auto",
```
jtbandes, line 11:
```suggestion
  auto: "auto",
```

PR #7413, opened 2024-01-31
Prevent unnecessary re-renders in 3D panel when selecting source

snosenzo:
**User-Facing Changes**

- Prevent unnecessary re-renders in 3D panel after selecting a source

**Description**
The ThreeDeePanel would render 3 times when a source was selected when it only needs to rerender once. This is because after the player is set for the source, a new messagepipleine store is created and the player along with its source need to get initialized. To limit rerenders in Extension panels while initializing a new player I've added some logic to not call `initPanel` when the player `presence` is `INITIALIZING`.  I also had to update `defaultPlayerState` in the messagepipeline store to take an optional parameter of whether it was initialized with the player so that it can initialize the default player state to `INITIALIZING` when there is a player present for the store or `NOT_PRESENT` when there is not a player (on startup).

A drawback of this solution is that it only applies to extension panels, but might be useful for other panels as well. One other consideration might be to prevent rendering the mosaic layout while the player is initializing to prevent thrashing and show a loading indicator instead. I wasn't sure what the right place to leverage this would be or if it would be desirable. 

FG-6410


jtbandes:
> **User-Facing Changes**
> 
> * Prevent unnecessary re-renders in 3D panel after selecting a source

IMO a user reading this in release notes would not know what it means. Also, looking at the diff, it has nothing to do with the 3D panel in particular. So maybe we just exclude from release notes, or call it something more generic like "performance improvements when switching data sources"?

packages/studio-base/src/components/PanelExtensionAdapter/PanelExtensionAdapter.tsx
```diff
@@ -547,6 +548,10 @@ function PanelExtensionAdapter(
     );
   }, [initialState, highestSupportedConfigVersion]);
 
+  const playerPresenceShouldBlockRender = useMemo(() => {
+    return playerPresence === PlayerPresence.INITIALIZING;
+  }, [playerPresence]);
```
jtbandes, line 553: A simple `=== (string)` does not need to be in `useMemo` which will actually add more overhead in this case

packages/studio-base/src/components/MessagePipeline/store.ts
```diff
@@ -27,9 +27,11 @@ import isDesktopApp from "@foxglove/studio-base/util/isDesktopApp";
 import { FramePromise } from "./pauseFrameForPromise";
 import { MessagePipelineContext } from "./types";
 
-export function defaultPlayerState(): PlayerState {
+export function defaultPlayerState(player?: Player): PlayerState {
   return {
-    presence: PlayerPresence.NOT_PRESENT,
+    // when there is a player we default to initializing, to prevent less thrashing in the UI when
```
jtbandes, line 32: > to prevent less thrashing

:philosoraptor:

jtbandes (commented):
I am a bit concerned that players which spend a long time in INITIALIZING (e.g. when establishing a websocket connection) will lead to the app looking broken. What does it look like if you try to connect to a websocket and the server is not running?

packages/studio-base/src/components/PanelExtensionAdapter/PanelExtensionAdapter.tsx
defunctzombie: More context: It will add more overhead because the memo itself has to iterate the dependency list to see if all are referentially equal so when the thing you memo is a cheap equality check you are better off avoiding the additional memo dependencies check as well.

defunctzombie:
> I am a bit concerned that players which spend a long time in INITIALIZING (e.g. when establishing a websocket connection) will lead to the app looking broken.

In what way? Because we don't initialize the panels?

---

Related question - but if re-rendering the panel is cheap why is it a problem to re-render it in the rare case of source selection? Or is re-render also referring to re-initialize?

packages/studio-base/src/components/PanelExtensionAdapter/PanelExtensionAdapter.tsx
snosenzo: Good catch. will update

snosenzo:
> I am a bit concerned that players which spend a long time in INITIALIZING (e.g. when establishing a websocket connection) will lead to the app looking broken. What does it look like if you try to connect to a websocket and the server is not running?

@jtbandes  it looks like for the websocket player it starts initializing and then goes into `PlayerPresence.RECONNECTING` so  it will remember the panel. Looking at the other Players it looks like they're only INITIALIZING until the sources are set up and /or the listener is set. 

> Related question - but if re-rendering the panel is cheap why is it a problem to re-render it in the rare case of source selection? Or is re-render also referring to re-initialize?

@defunctzombie Right, in saying `re-render` I was actually referring to re-running `initPanel`. This action forces the previous panel to completely unmount and a new panel to be initialized with a new `Renderer` resubscriptions, etc. This is especially heavyweight with URDF models being initialized via the layout, along with atlas textures being initialized and disposed. Should overall reduce time to get into the app from the splash screen w/ 3D/image panels.

packages/studio-base/src/components/MessagePipeline/store.ts
```diff
@@ -27,9 +27,11 @@ import isDesktopApp from "@foxglove/studio-base/util/isDesktopApp";
 import { FramePromise } from "./pauseFrameForPromise";
 import { MessagePipelineContext } from "./types";
 
-export function defaultPlayerState(): PlayerState {
+export function defaultPlayerState(player?: Player): PlayerState {
   return {
-    presence: PlayerPresence.NOT_PRESENT,
+    // when there is a player we default to initializing, to prevent thrashing in the UI when
```
defunctzombie, line 32: This comment is somewhat confusing because it doesn't really explain why or what setting a presence of INITIALIZING has to do with UI thrashing. I can see this change being useful by being an indicator that when a player instance is available we start in the initializing presence until we get an update from the player so downstream components (ui or otherwise) are aware that a player is present.

packages/studio-base/src/components/PanelExtensionAdapter/PanelExtensionAdapter.tsx
```diff
@@ -547,6 +548,8 @@ function PanelExtensionAdapter(
     );
   }, [initialState, highestSupportedConfigVersion]);
 
+  const playerPresenceShouldBlockRender = playerPresence === PlayerPresence.INITIALIZING;
```
defunctzombie, line 551: nit: `playerIsInitializing` makes more sense to me as the variable name - and the fact that it blocks `init` (not rendering) is noted in the logic below.

defunctzombie (approved):
A few suggestions on style/comments that will help future us understand the reasoning.

Issue #7414, opened 2024-01-31
Plots with set time window x-axis range don't automatically scroll


davidzimmerjr:
**Description**

- Version: 1.86.0
- Platform: Mac OS
- Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native): bag file

**Steps To Reproduce**

When plotting data and setting up the plot window with the below configuration where X axis has a set Range (seconds) [screenshot shows 15], in 1.84.0 the plot would scroll as the bag file would play.
![image](https://github.com/foxglove/studio/assets/58760631/3be8f65e-bcb9-417a-8930-de91680d8be9)

**Expected Behavior**
Plot scrolls as the bag file plays.

defunctzombie:
Thanks for reporting this. It is fixed in 1.86.1

davidzimmerjr:
Yes! Thank you :) Just noticed it working!

PR #7415, opened 2024-01-31
Revert "Reduce downsamples and re-renders for plot"

defunctzombie:
Reverts foxglove/studio#7362

I suspect this caused https://github.com/foxglove/studio/issues/7414


jtbandes (approved):
💮

PR #7417, opened 2024-01-31
(fix) handle empty min/maxYValue strings

HillaryManning:
**User-Facing Changes**

- FIX: Empty plot configuration string values for minYValue and maxYValue led to an initially rendered Y scale of 0-0

**Description**
The minYValue and maxYValue type currently supports number _or_ string, but empty strings were cast to 0 when creating the Y-bounds on the Plot for demo data. This led to y bounds of min: 0 and max: 0 (rather than defaulting to auto scale). I removed the empty string values from the test data and added a Storybook case to cover backwards compatibility.

### DIFF
--- a/packages/studio-base/src/panels/Plot/PlotCoordinator.ts
+++ b/packages/studio-base/src/panels/Plot/PlotCoordinator.ts
@@ -184,12 +184,13 @@ export class PlotCoordinator extends EventEmitter<EventTypes> {
       },
       y: {
         max:
-          config.maxYValue == undefined || (typeof config.maxYValue === 'string' && config.maxYValue.length === 0)
-
+          config.maxYValue == undefined ||
+          (typeof config.maxYValue === "string" && config.maxYValue.length === 0)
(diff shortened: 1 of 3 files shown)


packages/studio-base/src/panels/Plot/PlotCoordinator.ts
```diff
@@ -183,8 +183,16 @@ export class PlotCoordinator extends EventEmitter<EventTypes> {
         min: config.minXValue,
       },
       y: {
-        max: config.maxYValue == undefined ? undefined : +config.maxYValue,
-        min: config.minYValue == undefined ? undefined : +config.minYValue,
+        max:
+          config.maxYValue == undefined ||
+          (typeof config.maxYValue === "string" && config.maxYValue.length === 0)
```
jtbandes, line 188: nit: this could be slightly shorter as `config.maxYValue === ""`

jtbandes (approved):
👍  nice!

PR #7419, opened 2024-02-01
Fix plot panel X axis precision flicker

wimagguc:
**User-Facing Changes**

Once again, X-axis tick marks in the Plot panel use 2 decimal places, reducing jitter for live-updating data.

**Description**

The new Plot panel undid the fix in https://github.com/foxglove/studio/pull/2933 — this PR merely reapplies the same change to the new chart renderer.

Before:

https://github.com/foxglove/studio/assets/349687/03eb7abd-6873-488e-8f54-bce8f0700786

After:

https://github.com/foxglove/studio/assets/349687/e7d93b96-0bb3-4fac-a0fa-fd33469f7e1a


jtbandes (commented):
I think we need to do this for Y too (see the comment thread on the issue)

wimagguc:
> I think we need to do this for Y too (see the comment thread on the issue)

I've added custom formatters to the y axis ticks, where the number of digits displayed depends on the min/max value displayed. For example, between `(100, 110)` the ticks will be all integers `101, 102...`, and for small numbers such as `(0.1, 0.2)` each tick will display two digits `(0.11, 0.12...)`.

To be fair, this is more or less what chartjs does already, and the custom ticks take a bit more calculating than I'd like — while I could never actually replicate the original bug or observe flickering / odd behavior myself.

jtbandes:
It's definitely an improvement. I can no longer reproduce the really long "47.99999" problem. However there is still some flickering because trailing zeros are not shown (e.g. it switches from "-45.1" to "-45" instead of "-45.0"):

<table><tr><th>Main</th><th>This PR</th></tr>
<tr><td>

https://github.com/foxglove/studio/assets/14237/a74d2496-3b16-443a-84dd-e321caae70c6

</td><td>

https://github.com/foxglove/studio/assets/14237/7598fc49-59c1-4b24-bf06-a251d901e880

</td>
</tr></table>

wimagguc:
> Is there a reason to use different logic for X vs Y ticks?

The most important difference is that with the X ticks we only need to modify the last one, because the changing the middle ones' width doesn't change the chart's. This also means that losing precision is not such a big deal, because in the worst case people can look at the other ticks and/or hover over the chart.

But for the Y axis all ticks' bounding boxes have to be around the same width, because if any one tick has too many digits, it will push the Y axis towards the right. And if we limit the number of digits to an arbitrary number, we'll end up in situations where all of the ticks are the same, for example between `0.123` and `0.124` the Y axis will show 10 ticks, all of them being the same: `0.12`

indirectlylit:
I can't find it anywhere in chart.js docs, but if we could simply fix the position of the y-axis instead of having it move with text, it would solve a lot of problems:

image

If we can't do that, it seems that setting the first and last tick label to be a long string hacks it in and prevents flicker:

```js
scaleOption.ticks.callback = function (value, index, ticks) {
  // hide first/last ticks
  if (index === 0 || index === ticks.length - 1) {
    return "       ";
  }
  // otherwise use chart.js's default formatter
  return Ticks.formatters.numeric.apply(this, [value as number, index, ticks]);
};
```



The labeled axis are a bit odd anyway because they don't occur at regularly-spaced tick marks as the user pans around the data. There actually appears to be a chart.js bug or configuration issue where vertical grid-lines and tick labels disappear long before they should:

image

wimagguc:
Decision from eng standup: we'll ship the X axis work and decide separately whether Y is actually bug.

indirectlylit:
Thanks guys sounds good!

Some quick updates to my comment above based on standup for future work:

- Hadn't considered fixed min/max axis values
  - In this case it _does_ make sense to label plot boundaries. In this case there would never be a flicker.
- Hadn't considered deep zoom values
  - In this case a fixed-position y-axis wouldn't work because you might need arbitrarily long decimals.
  - This resulted in a state where the _only_ labels were the boundaries, with no gridlines/ticks. The spacing between tick marks / gridlines would optimally vary with the size of the chart and the min/max values

Seem like there are important differences between fixed-min/max axes and variable ones.

PR #7420, opened 2024-02-01
Buffer messages of bag & db3 sources as raw uint8arrays

achim-k:
**User-Facing Changes**
Store buffered messages as raw bytes to reduce javascript heap memory pressure (bag + db3)

**Description**
Similar to #7376  this PR converts the bag (ROS1) and db3 (ROS2) sources to raw sources which allows to buffer messages as Uint8Arrays. This Improves worker communication performance and reduces heap memory pressure.

Resolves FG-6464

### DIFF
--- a/packages/studio-base/src/dataSources/McapLocalDataSourceFactory.ts
+++ b/packages/studio-base/src/dataSources/McapLocalDataSourceFactory.ts
@@ -7,7 +7,7 @@ import {
   IDataSourceFactory,
 } from "@foxglove/studio-base/context/PlayerSelectionContext";
 import { IterablePlayer } from "@foxglove/studio-base/players/IterablePlayer";
-import { WorkerRawIterableSource } from "@foxglove/studio-base/players/IterablePlayer/WorkerRawIterableSource";
+import { WorkerSerializedIterableSource } from "@foxglove/studio-base/players/IterablePlayer/WorkerSerializedIterableSource";
 import { Player } from "@foxglove/studio-base/players/types";
 
 class McapLocalDataSourceFactory implements IDataSourceFactory {
@@ -23,7 +23,7 @@ class McapLocalDataSourceFactory implements IDataSourceFactory {
       return;
     }
 
-    const source = new WorkerRawIterableSource({
+    const source = new WorkerSerializedIterableSource({
       initWorker: () => {
         return new Worker(
           // foxglove-depcheck-used: babel-plugin-transform-import-meta
--- a/packages/studio-base/src/dataSources/McapLocalDataSourceFactory.ts
+++ b/packages/studio-base/src/dataSources/McapLocalDataSourceFactory.ts
@@ -7,7 +7,7 @@ import {
   IDataSourceFactory,
 } from "@foxglove/studio-base/context/PlayerSelectionContext";
 import { IterablePlayer } from "@foxglove/studio-base/players/IterablePlayer";
-import { WorkerSerializedIterableSource } from "@foxglove/studio-base/players/IterablePlayer/WorkerSerializedIterableSource";
+import { WorkerRawIterableSource } from "@foxglove/studio-base/players/IterablePlayer/WorkerRawIterableSource";
 import { Player } from "@foxglove/studio-base/players/types";
 
 class McapLocalDataSourceFactory implements IDataSourceFactory {
@@ -23,7 +23,7 @@ class McapLocalDataSourceFactory implements IDataSourceFactory {
       return;
     }
 
-    const source = new WorkerSerializedIterableSource({
+    const source = new WorkerRawIterableSource({
       initWorker: () => {
         return new Worker(
           // foxglove-depcheck-used: babel-plugin-transform-import-meta
--- a/packages/studio-base/src/dataSources/Ros2LocalBagDataSourceFactory.ts
+++ b/packages/studio-base/src/dataSources/Ros2LocalBagDataSourceFactory.ts
@@ -6,7 +6,10 @@ import {
   IDataSourceFactory,
   DataSourceFactoryInitializeArgs,
 } from "@foxglove/studio-base/context/PlayerSelectionContext";
-import { IterablePlayer, WorkerIterableSource } from "@foxglove/studio-base/players/IterablePlayer";
+import {
+  IterablePlayer,
+  WorkerRawIterableSource,
+} from "@foxglove/studio-base/players/IterablePlayer";
 import { Player } from "@foxglove/studio-base/players/types";
 
 class Ros2LocalBagDataSourceFactory implements IDataSourceFactory {
@@ -25,7 +28,7 @@ class Ros2LocalBagDataSourceFactory implements IDataSourceFactory {
       return;
     }
 
-    const source = new WorkerIterableSource({
+    const source = new WorkerRawIterableSource({
       initWorker: () => {
         return new Worker(
           // foxglove-depcheck-used: babel-plugin-transform-import-meta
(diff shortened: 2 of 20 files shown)


packages/studio-base/src/dataSources/Ros2LocalBagDataSourceFactory.ts
```diff
@@ -25,7 +28,7 @@ class Ros2LocalBagDataSourceFactory implements IDataSourceFactory {
       return;
     }
 
-    const source = new WorkerIterableSource({
+    const source = new WorkerRawIterableSource({
```
snosenzo, line 31: I know this was already like this but since you also updated verbage elsewhere to be consistent: `WorkerSerializedIterableSource`?

achim-k: Will change, more consistent indeed.

packages/studio-base/src/players/IterablePlayer/rosdb3/RosDb3IterableSource.ts
```diff
@@ -86,14 +85,30 @@ export class RosDb3IterableSource implements IIterableSource {
...
+              const newComplexType = datatypes.get(type);
+              if (!newComplexType) {
+                // Should in theory never happen as these are all well-known types
+                throw new Error(`invariant: Subtype ${type} of type ${rosType.name} not found.`);
+              }
+              typesToProcess.push(newComplexType);
+              typesForMessage.set(type, newComplexType);
+            }
+          }
+        }
```
achim-k, line 105: This is necessary for parseChannel to parse the schema correctly. Could also adapt parseChannel to take a message definition array as an input

packages/studio-base/src/dataSources/McapLocalDataSourceFactory.ts
```diff
@@ -23,7 +23,7 @@ class McapLocalDataSourceFactory implements IDataSourceFactory {
       return;
     }
 
-    const source = new WorkerRawIterableSource({
+    const source = new WorkerSerializedIterableSource({
```
defunctzombie, line 26: Why are we going back to a serialized worker? Not sure I understand why the McapLocal source is changing in a PR for bag and db3.

defunctzombie: Oh nvm - serialized is the original byte data - this is the rename from "raw". Somewhat confusing. Would have been easier to see in a separate PR.

packages/studio-base/src/players/IterablePlayer/rosdb3/RosDb3IterableSource.ts
```diff
@@ -86,14 +85,30 @@ export class RosDb3IterableSource implements IIterableSource {
           message: `Topic "${topicDef.name}" has unsupported datatype "${topicDef.type}"`,
           tip: "ROS 2 .db3 files do not contain message definitions, so only well-known ROS types are supported in Foxglove Studio. As a workaround, you can convert the db3 file to mcap using the mcap CLI. For more information, see: https://docs.foxglove.dev/docs/connecting-to-data/frameworks/ros2",
         });
-        continue;
+      } else {
+        // Create the full gendeps-like message definition so that parseChannel() can parse it.
+        const typesToProcess = [parsedMsgdef];
+        const typesForMessage: RosDatatypes = new Map();
+        while (typesToProcess.length > 0) {
```
defunctzombie, line 92: A bit scared of the infinite loop possibilities here.

packages/studio-base/src/players/IterablePlayer/rosdb3/RosDb3IterableSource.ts
defunctzombie: What does `parseChannel` have to do with DB3 files? That's an `mcap` thing is it not?

packages/studio-base/src/players/IterablePlayer/rosdb3/RosDb3IterableSource.ts
```diff
@@ -133,29 +148,25 @@ export class RosDb3IterableSource implements IIterableSource {
       startTime: start,
       endTime: inclusiveEndTime,
       topics: Array.from(topics.keys()),
+      rawMessages: true,
```
defunctzombie, line 151: `serializedMessages` to align with other conventions?

defunctzombie (commented):
The "Create the full gendeps-like message definition so that parseChannel() can parse it." logic is still lost on me ... how did the db3 source previous de-serialize the data if we didn't have this expanded message definition?

packages/studio-base/src/players/IterablePlayer/rosdb3/RosDb3IterableSource.ts
achim-k: `parseChannel` - which is used in the `DeserializingIterableSource` - can build decoders for several schema & message encodings including `ros2msg/cdr`. However, `parseChannel` expects the schema to be given as a raw schema definition string (gendeps like) and not an array of `MessageDefinitions`. That's why I have to construct the raw schema definition here.

packages/studio-base/src/dataSources/McapLocalDataSourceFactory.ts
achim-k: will revert that commit again and submit a new PR for it. The change introduced too many changes

packages/studio-base/src/players/IterablePlayer/rosdb3/RosDb3IterableSource.ts
achim-k: This stems from the `@foxglove/rosbag2` dependency, so not trivial to change

achim-k:
> The "Create the full gendeps-like message definition so that parseChannel() can parse it." logic is still lost on me ... how did the db3 source previous de-serialize the data if we didn't have this expanded message definition?

Before, deserialization happened within `@foxglove/rosbag2`'s `Rosbag2.readMessages(...)`. Now message deserialization happens in the `DeserializingIterableSource` which is the reason why we have to communicate decoding information, including the gendeps-like message definition, to that source.

packages/studio-base/src/players/IterablePlayer/rosdb3/RosDb3IterableSource.ts
defunctzombie: How can we make this more flexible to such data sources so we don't need to construct a fake schema? What I am not so keen on is that we are opening up the door for some bugs in our schema reconstruction. Does this impact IDL schemas? I think you can have those in a db3 file? Or maybe this is not true since this is only well-known schemas that we support for db3.

Generally I could be "ok" with this since db3 is not the preferred format on Iron and newer but we do have users on humble where it is the default so might get some reports if something weird breaks.

packages/studio-base/src/players/IterablePlayer/BagIterableSource.ts
```diff
@@ -214,49 +211,29 @@ export class BagIterableSource implements IIterableSource {
...
-            severity: "error",
-            message: `Cannot deserialize message for missing connection id ${connectionId}`,
-            tip: `Check that your bag file is well-formed. It should have a connection record for every connection id referenced from a message record.`,
-          },
-        };
-      }
+      // bagMsgEvent.data is a view on top of the entire chunk. To avoid keeping references for
+      // chunks (which will fill up memory space when we cache messages) when make a copy of the
+      // data.
+      const dataCopy = bagMsgEvent.data.slice();
```
defunctzombie, line 217: Do we still need this slice? Maybe yes because we need to transfer the buffer and for that we do need a copy so we don't transfer the entire chunk?

defunctzombie (approved):

packages/studio-base/src/players/IterablePlayer/BagIterableSource.ts
achim-k: I am not entirely sure but for now I would leave it as it. We only transfer the underlying array buffer, not the Uint8Array object.

PR #7423, opened 2024-02-01
(fix) prevent play button flicker

HillaryManning:
**User-Facing Changes**

- FIX: Prevent play button flicker on rapidly repeating playback

**Description**
The controls for repeating playback have been moved from the RepeatAdaptor to the iterablePlayer. This allows us to more directly determine the order of operations for changing the isPlaying/idle states as they relate to repeating the data. This change means the 'isPlaying' state remains consistent while the data loops, which prevents the odd play button flashing behavior.


jtbandes (commented):
Nice, this does seem to fix the flickering. But I notice that the button only works about 75% of the time (with almost_empty.mcap) or if I turn up the playback speed to 3x, the button seems to not really work at all. I think it might be worth some deeper investigation to see why that's happening

packages/studio-base/src/components/PlaybackControls/index.tsx
```diff
@@ -111,8 +121,17 @@ export default function PlaybackControls(props: {
...
   const toggleRepeat = useCallback(() => {
-    setRepeat((old) => !old);
-  }, [setRepeat]);
+    // toggle repeat on the workspace
+    setRepeat(!repeat);
+  }, [setRepeat, repeat]);
+
+  useEffect(() => {
+    // if workspace has a preference stored that is not reflected in the iterable player...
+    if (repeat !== repeatEnabled) {
```
jtbandes, line 130: nit: I bet you could even leave out the `if` and the dependency on `repeatEnabled` here

packages/studio-base/src/components/PlaybackControls/index.tsx
```diff
@@ -111,8 +121,17 @@ export default function PlaybackControls(props: {
   } = useWorkspaceActions();
 
   const toggleRepeat = useCallback(() => {
-    setRepeat((old) => !old);
```
jtbandes, line 114: nit: I'd leave this as-is if possible to avoid creating a new toggleRepeat function (and therefore recreating event listeners) when `repeat` changes

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
```diff
@@ -1049,7 +1071,11 @@ export class IterablePlayer implements Player {
 
     try {
       while (this.#isPlaying && !this.#hasError && !this.#nextState) {
-        if (compare(this.#currentTime, this.#end) >= 0) {
+        if (this.#repeatEnabled && compare(this.#currentTime, this.#end) === 0) {
```
jtbandes, line 1074: I would suggest hoisting the `compare(...)` result to a local variable to avoid computing it twice (this loop runs very frequently)

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
```diff
@@ -275,6 +276,19 @@ export class IterablePlayer implements Player {
...
+  public repeatPlayback(): void {
+    if (!this.#isPlaying) {
+      return;
+    }
+    this.#untilTime = undefined;
+    this.#lastTickMillis = undefined;
+    this.#lastRangeMillis = undefined;
+    this.#currentTime = this.#start;
+
+    this.#setState("reset-playback-iterator");
```
jtbandes, line 289: In order to trigger downstream panels to reset their state when playback loops, we will need `lastSeekTime` to be updated when the repeat happens. From a quick glance I believe the `reset-playback-iterator` flow will not do that — it might be better to piggyback on the seek functionality instead. cc @defunctzombie for any better advice on how to do this within the player

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
```diff
@@ -275,6 +276,19 @@ export class IterablePlayer implements Player {
     }
   }
 
+  // Repeating is a continuation of the playing state/we should already be playing
+  public repeatPlayback(): void {
```
jtbandes, line 280: nit: doesn't need to be public

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
defunctzombie: You want to use the seek behavior because it will trigger the "backfill" messages to load. This is important for a source like the data platform where we might fetch a backfill message prior to the start time of playback (i.e. a less frequently published image or map topic).

HillaryManning: Got it, on it: thank you both!

packages/studio-base/src/components/PlaybackControls/index.tsx
HillaryManning: Probably: was being paranoid in favor of minimizing unnecessary calls. But removing the dependency is a fair argument.

jtbandes (approved):
Nice! The play/pause button seems to work great even with really fast repeats now. One minor thing I noticed is that when playback is at the end and repeat is off, toggling repeat on used to automatically start playback again, and now it doesn't. But I don't think that's a particularly important behavior

PR #7439, opened 2024-02-02
Adjust settings tree behavior to allow horizontal scrolling

2metres:
**User-Facing Changes**
None

#### Before

https://github.com/foxglove/studio/assets/924528/e1d067fe-bd73-453d-a487-e33f7a648df3

#### After

https://github.com/foxglove/studio/assets/924528/60498d34-19ac-4314-87d6-f950a2c11c52

**Description**
Instead of allowing settings sidebar to be infinitely collapsed until the label/form is not visible set a min width on them and allow horizontal scrolling

### DIFF
--- a/packages/studio-base/src/components/Autocomplete/Autocomplete.tsx
+++ b/packages/studio-base/src/components/Autocomplete/Autocomplete.tsx
@@ -71,6 +71,9 @@ const useStyles = makeStyles()((theme) => ({
       color: theme.palette.error.main,
     },
   },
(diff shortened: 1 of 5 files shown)


2metres:
Also resolves FG-6493

jtbandes:
I'm having trouble seeing what I am supposed to be testing here. Could you update the PR description with some before/after screenshots? The original linked ticket was about a max width, but I see min width and several other things changing in this PR>

PR #7441, opened 2024-02-02
Make ModelCache shared across 3D panels

snosenzo:
**User-Facing Changes**

- better mesh load times with multiple 3D or image panels

**Description**

Turned ModelCache into a global singleton. It is only cleared on page refresh to maintain loaded models across switching tabs, layouts and sources. This aligns with what we were previously doing. 
This means that models can build up indefinitely, however I don't think it's all that common to load many different meshes across a session, and historically we haven't had issues with it. 

Also FWIW, I did also look into threejs' support for sharing 3D objects across scenes and renderables. For the most part this is discouraged and sometime not possible, however, it works in foxglove because we run `clone` all models coming from the `ModelCache`. This unfortunately does cause duplication in the data for each usage, but this is what we were doing previously to keep the different usages of models separate from one another.

Tested:
 -  Have two 3D panels in a layout both using the same model. Change the model URL in one and make sure the other isn't affected. 
 - unmounting one panel doesn't clear another.
 - ensured multiple Renderers aren't being kept in memory on unmount by modelCache

Assists problem in FG-6409


packages/studio-base/src/panels/ThreeDeeRender/ModelCache.ts
```diff
@@ -414,3 +449,30 @@ function unrewriteUrl(url: string): string {
 function baseUrl(url: string): string {
   return url.slice(0, url.lastIndexOf("/") + 1);
 }
+
+let modelCacheInstance: ModelCache | undefined = undefined;
```
achim-k, line 453: I don't think that it will work with only one instance, I believe that you need a instance for every possible configuration option

achim-k (commented):
I think there is an issue with different configuration options, e.g. the mesh up axis setting. The layout below yields different URDF visualizations on your PR and on current main.

[foxglove-layout (9).json](https://github.com/foxglove/studio/files/14170117/foxglove-layout.9.json)

packages/studio-base/src/panels/ThreeDeeRender/ModelCache.ts
snosenzo: Good catch. I think I can keep the single instance, but I'll have to apply the up-axes after taking from the cache and cloning them, which should be possible.

snosenzo: Updated to have the cache key use ignoreColladaUpAxis. It's unfortunate it's applied to all models, but I thought it would keep it simpler and faster than only applying it to the cache if it's definitely not a different type of model (ie: test against override mediatype and url not containing non-collada media types). I don't think that there are many folks out there not using 3D panels with intermixed collada up axis options.

packages/studio-base/src/panels/ThreeDeeRender/ModelCache.ts
```diff
@@ -156,14 +177,10 @@ export class ModelCache {
     const gltf = await gltfLoader.loadAsync(url);
     manager.itemEnd(url);
 
-    // THREE.js uses Y-up, while Studio follows the ROS
```
snosenzo, line 151: This has the same comment as the conditional applications of `rotateX` elsewhere so it made me think that this was meant to also be wrapped in the same conditional that the others were wrapped in. If someone has more context on this and I should make it unconditionally rotateX again, please do flag.

snosenzo:
> Not sure if related to the build failure, but the following layout does not work on the PR deployment:

I think you've actually found a very niche layout bug for layouts built from the default layout. I think this happens because the first 3D panel has the same panel ID as the one from the default layout, so it ends up not getting changed by the imported layout somehow?

packages/studio-base/src/panels/ThreeDeeRender/ModelCache.ts
```diff
@@ -47,46 +51,72 @@ const OBJ_MIME_TYPES = ["model/obj", "text/prs.wavefront-obj"];
...
+      // [REP-0103](https://www.ros.org/reps/rep-0103.html) convention of Z-up
+      if (modelCacheOptions.meshUpAxis === "y_up") {
+        model.rotateX(Math.PI / 2);
+      }
+      return model;
+    };
   }
 
   public async load(
+    config: ModelCacheOptions,
```
achim-k, line 89: can/should this method now be made private?

packages/studio-base/src/panels/ThreeDeeRender/ModelCache.ts
```diff
@@ -301,16 +311,54 @@ export class ModelCache {
...
-  public dispose(): void {
+  public incrementUsage(): void {
+    this.#usageCount += 1;
+  }
+
+  public decrementUsage(): void {
+    this.#usageCount -= 1;
+    if (this.#usageCount <= 0) {
+      this.reset();
+    }
```
achim-k, line 322: Would this also happen when e.g. loading a new recording because AFAIK all panels are reloaded?

achim-k (approved):

packages/studio-base/src/panels/ThreeDeeRender/ModelCache.ts
snosenzo: good callout, will fix

packages/studio-base/src/panels/ThreeDeeRender/ModelCache.ts
snosenzo: That's correct, it would reset when the panels are reloaded on source change. This definitely isn't ideal for performance, but we should clear the cache at some point. Maybe after a reasonable timeout if its usageCount is not incremented? My qualm with that is that it could still hold onto models for a while if the 3D panel is not unmounted for a certain period of time, but maybe this is less of a concern because it wouldn't be too common for different models to build up in the cache

achim-k: How are models stored? I guess model vertices mostly as array buffers so the JS heap shouldn't be affect to much? Maybe for now I think it's OK to keep all models in the cache until the app is closed or refreshed.

snosenzo: Yeah, majority is geometry array buffers, so it shouldn't add too much to the heap.

packages/studio-base/src/panels/ThreeDeeRender/ModelCache.ts
```diff
@@ -33,6 +35,8 @@ type LoadModelOptions = {
   referenceUrl?: string;
 };
 
+type CacheKey = string & { __brand: "CacheKey" };
```
defunctzombie, line 38:
```suggestion
type CacheKey = Opaque<string, "CacheKey">;
```

https://github.com/foxglove/studio/blob/aeae9afbc04a15cc6a854d4ea1e138f00296e835/packages/studio-base/src/context/PerformanceContext.ts#L15

packages/studio-base/src/panels/ThreeDeeRender/ModelCache.ts
```diff
@@ -45,48 +49,74 @@ const STL_MIME_TYPES = ["model/stl", "model/x.stl-ascii", "model/x.stl-binary",
 const DAE_MIME_TYPES = ["model/vnd.collada+xml"];
 const OBJ_MIME_TYPES = ["model/obj", "text/prs.wavefront-obj"];
 
-export class ModelCache {
+class ModelCache {
```
defunctzombie, line 52: This class could benefit from a class comment and a comment on some of the public interfaces. Maybe mentioning some gotchas or expectations/behaviors of the cache.

packages/studio-base/src/panels/ThreeDeeRender/ModelCache.ts
```diff
@@ -218,7 +237,7 @@ export class ModelCache {
         if (this.#colladaTextureObjectUrls.has(textureUrl)) {
           continue;
         }
-        const textureAsset = await this.#fetchAsset(textureUrl);
+        const textureAsset = await config.fetchAsset(textureUrl);
```
defunctzombie, line 240: As an outsider this looked a bit awkward - I wouldn't have expected to call `fetchAsset` on an instance of `ModelCacheOptions`.

packages/studio-base/src/panels/ThreeDeeRender/ModelCache.ts
```diff
@@ -414,3 +459,32 @@ function unrewriteUrl(url: string): string {
...
+  load: (
+    url: string,
+    opts: LoadModelOptions,
+    reportError: ErrorCallback,
+  ) => Promise<LoadedModel | undefined>;
+  release: () => void;
+  options: ModelCacheOptions;
+};
+
+export function getConfiguredModelCache(
```
defunctzombie, line 477: nit: stylistically I would make this a static on the ModelCache. I've found it more natural to have the class expose static methods on different ways to construct it (when you want to make the constructor itself private even).

packages/studio-base/src/panels/ThreeDeeRender/ModelCache.ts
```diff
@@ -414,3 +459,32 @@ function unrewriteUrl(url: string): string {
 function baseUrl(url: string): string {
   return url.slice(0, url.lastIndexOf("/") + 1);
 }
+
+let modelCacheInstance: ModelCache | undefined = undefined;
+
+export type ConfiguredModelCache = {
```
defunctzombie, line 465: Seems more like an interface to me.

packages/studio-base/src/panels/ThreeDeeRender/ModelCache.ts
snosenzo: updated the name of the interface/type to be `ModelCacheConfiguration`, that seems semantically a bit better than `Options`. I also considered `Context`, but felt that excluded the "options" present

PR #7442, opened 2024-02-02
Fix memory estimation not differintiating between normal and sliced messages

achim-k:
**User-Facing Changes**
None (as #7376 wasn't released yet)

**Description**
Fixes a regression introduced with #7376 regarding message size estimation. The message size estimation did not take into account potentially sliced messages which could lead to the message size estimate being far from reality.


jtbandes (approved):
Is there a way we can add unit tests for this?

PR #7443, opened 2024-02-02
(enhance) move playback speed to workspace

HillaryManning:
**User-Facing Changes**

- Playback speed setting moved from layout level to workspace: playback speed setting will persist across multiple layouts

**Description**
Followed the playback 'repeat' property as a model for moving playback speed from layout to workspace. Made the playback speed type more specific (let me know if too much: not sure if we support custom speeds somehow).


packages/studio-base/src/components/PlaybackSpeedControls.tsx
```diff
@@ -37,18 +36,18 @@ export default function PlaybackSpeedControls(): JSX.Element {
...
-      setPlaybackConfig({ speed: newSpeed });
-      setPlaybackSpeed?.(newSpeed);
+  const speed = useWorkspaceStore(selectPlaybackSpeed);
+
+  // Speed setting lives on the Workspace/persists accross layouts
+  const {
+    playbackControlActions: { setSpeed },
+  } = useWorkspaceActions();
+
+  const assignSpeed = useCallback(
```
jtbandes, line 46: is there a purpose for this function instead of directly calling `setSpeed`?

packages/studio-base/src/types/panels.ts
```diff
@@ -28,7 +28,7 @@ export type PanelConfig = {
 export type TimeDisplayMethod = "SEC" | "TOD";
 
 export type PlaybackConfig = {
```
jtbandes, line 30: I think we might want to explore deleting this PlaybackConfig type. It existed to be part of LayoutData, and `SET_PLAYBACK_CONFIG` actions etc. If those are no longer relevant then we can probably delete quite a bit of code.

HillaryManning: Just pushed my work from this morning where I removed this from the panel/created a type for the workspace. There are two playback controls on the Workspace: repeat, and speed.

packages/studio-base/src/components/PlaybackSpeedControls.tsx
HillaryManning: There was earlier on but not now/will clean up 👍

packages/studio-base/src/context/CurrentLayoutContext/actions.ts
jtbandes, line 1: There's still an `export type SET_PLAYBACK_CONFIG` in here that can be deleted

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
```diff
@@ -39,6 +39,7 @@ import {
   TopicStats,
 } from "@foxglove/studio-base/players/types";
 import { RosDatatypes } from "@foxglove/studio-base/types/RosDatatypes";
+import { PlaybackConfig } from "@foxglove/studio-base/types/Workspace";
```
jtbandes, line 42: I don't think we should have the players themselves importing Workspace-related stuff. I am not sure the player needs to care what the specific numbers that we expose in the UI are; it should work regardless of the number passed in. What do you think?

packages/studio-base/src/types/Workspace.ts
```diff
@@ -0,0 +1,8 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+export type PlaybackConfig = {
+  repeat: boolean;
```
jtbandes, line 6: is this used? I saw `PlaybackConfig["speed"]` in a few places but not the whole `PlaybackConfig` type itself.

packages/studio-base/src/components/MessagePipeline/types.ts
```diff
@@ -29,10 +30,10 @@ export type MessagePipelineContext = Immutable<{
   fetchAsset: BuiltinPanelExtensionContext["unstable_fetchAsset"];
   startPlayback?: () => void;
   pausePlayback?: () => void;
+  setPlaybackSpeed?: ((speedFraction: PlaybackSpeed) => void) | undefined;
```
jtbandes, line 33:
```suggestion
  setPlaybackSpeed?: (speedFraction: PlaybackSpeed) => void;
```

jtbandes (approved):

Issue #7406, opened 2024-01-27
Studio plot window starts from epoch 0 on restart when using "use_sim_time" in ROS1


calumblair:
**Description**
When using foxglove studio with foxglove bridge, if I playback a ROSbag with `<param name="/use_sim_time" value="true"/>`, and `--clock` (i.e. the ROSbag playback is controlling timing), when I restart playback (i.e. rerun the launch file), the Foxglove studio plot axis starts at epoch 0, and all the data is squashed at the right-hand side of the plot window.

Possibly related to https://github.com/foxglove/studio/issues/5244. 

- Version: Studio (desktop version) 1.8.5, foxglove-bridge 0.7.5
- Platform: Linux, x86_64
- Data source (e.g. bag file, mcap file, rosbridge, ROS 1/2 native):
Bagfile from [here](https://github.com/cra-ros-pkg/robot_localization/blob/noetic-devel/test/test1.bag) 

**Steps To Reproduce**

1. Checkout and build foxglove-bridge with the following docker-compose file:
```
services:
  foxglove-ros-bridge:
    build:
      context: .
      dockerfile: Dockerfile.ros1
    volumes:
      - .:/ros1_ws/src/ros-foxglove-bridge
    tty: true
    ports:
      - 8765:8765
    command: /bin/bash
    restart: always
```
and the  following launch file:
```
<launch>
    <include file="$(find foxglove_bridge)/ros1_foxglove_bridge/launch/foxglove_bridge.launch">
    </include>

    <param name="/use_sim_time" value="true"/> 

     <node pkg="rosbag" type="play" name="rosbagplay" output="screen" args=
        "--clock --bags /ros1_ws/src/ros-foxglove-bridge/data/test1.bag" />

</launch>
```
2. Copy the bagfile into `ros-foxglove-bridge/data/test1.bag`
3. Launch the foxglove-bridge container
4. From the container, run `roslaunch foxglove_bridge launch_wrapper.launch`
5. Open foxglove-studio, open a plot window and plot anything (e.g. `/fix/latitude`)
6. Observe that the plot window looks fine, especially the x-axis:
![firstlaunch_ok](https://github.com/foxglove/studio/assets/11029297/0176e96e-313f-4f40-8d78-070fe57822c3)
7. Stop and restart the launch file: Ctrl-C then re-run `roslaunch foxglove_bridge launch_wrapper.launch`
8. Observe that the plot window x-axis now starts at epoch 0 and that all the plot data is squashed to the right hand side
[...]

PR #7449, opened 2024-02-05
fix: foxglove websocket player incorrect startTime

defunctzombie:
**User-Facing Changes**
Fix the startTime behavior for foxglove websocket player when a connection is reset.

**Description**
The foxglove websocket player produces an incorrect _startTime_ (of 0) when a new session is started that uses "clock" time leading to incorrect plot displays like this: https://github.com/foxglove/studio/issues/7406

This behavior occurs because the state emit updates the `startTime` to the `currentTime` when the currentTime is less than the startTime:
https://github.com/foxglove/studio/blob/main/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts#L869

During a session reset, the currentTime is cleared and the #clockTime is cleared (https://github.com/foxglove/studio/blob/main/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts#L1290) so when the emit computes the currentTime it uses `ZERO_TIME` (https://github.com/foxglove/studio/blob/main/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts#L1186) to set startTime and never updates it again.

The reason this does _not_ happen for the first connection to the server is because topics are not yet available and the player emit bails early:
https://github.com/foxglove/studio/blob/main/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts#L853

On subsequent re-connections, topics are already available and the logic to set `startTime` runs before any clock message is received.

This PR updates the client `time` event handler to re-set the `startTime` and `endTime` on the first setting of `clockTime`. Since `clockTime` will be used for subsequent calls to `getCurrentTime` this results in a `startTime` that is properly updated to a server specified clock time instead of `ZERO_TIME`.

Fixes: #7406

### DIFF
--- a/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
+++ b/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
@@ -203,6 +203,7 @@ export default class FoxgloveWebSocketPlayer implements Player {
       if (this.#closed) {
         return;
       }
+      log.info("opened connection");
       if (this.#connectionAttemptTimeout != undefined) {
         clearTimeout(this.#connectionAttemptTimeout);
       }
@@ -299,6 +300,11 @@ export default class FoxgloveWebSocketPlayer implements Player {
       this.#serverPublishesTime = this.#serverCapabilities.includes(ServerCapability.time);
       this.#supportedEncodings = event.supportedEncodings;
       this.#datatypes = new Map();
+
+      // Clear the thopics which makes us request them again
+      // and has the side-effect that we don't emit without a time update
+      // fixme - this is annoying because it will thrash the display
+      this.#topics = undefined;
 
       // If the server publishes the time we clear any existing clockTime we might have and let the
       // server override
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
```diff
@@ -846,10 +854,17 @@ export default class FoxgloveWebSocketPlayer implements Player {
   // Potentially performance-sensitive; await can be expensive
   // eslint-disable-next-line @typescript-eslint/promise-function-async
   #emitState = debouncePromise(() => {
+    // When the connection closes, we should not emit anoyher state update because it will override the startTime?
```
snosenzo, line 857: unfinished comment w/ typo?

packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
```diff
@@ -846,10 +854,17 @@ export default class FoxgloveWebSocketPlayer implements Player {
...
+    //
+
     if (!this.#listener || this.#closed) {
       return Promise.resolve();
     }
 
+    // the reason this does not break us on startup is because this #topics gate
```
snosenzo, line 864: It feels like this comment is meant to be in the context of the PR? it's a bit confusing as to what "this" is referring to

snosenzo (approved):
Overall seems fine, but I'm wondering if it's worth emitting state without topics at all to avoid extra rerenders.

packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts
defunctzombie: stale comment - removed

achim-k (approved):

defunctzombie:
> but I'm wondering if it's worth emitting state without topics at all to avoid extra rerenders

We want to emit a state with a presence of "RECONNECTING" to show that feedback to the user.

PR #7452, opened 2024-02-06
avoid creating temporary comlink proxies

achim-k:
**User-Facing Changes**
None

**Description**
When using Comlink to wrap a remote interface, there is a subtle difference in usage which can lead to [new temporary proxy being created](https://github.com/foxglove/comlink/blob/9181fa505671b35b1e66e0a8361a6fc1bdd03307/src/comlink.ts#L499) for every function call:

```ts
const remote = comlink.wrap(new Worker("..."));
for (...) {
  await remote.someFunc();  // creates a temporary proxy on every call
}
```

This can be avoided by creating the proxy only once:
```ts
const remoteSomeFunc = remote.someFunc;  // create the proxy once
for (...) {
  await remoteSomeFunc();  // no temporary proxy created
}
```
The second usage is generally to be preferred as it avoids creating temporary proxies. While the performance overhead of creating temporary proxies is probably negligible, it does have some impact on garbage collection time as more objects have to be freed.

This PR changes all Comlink usage to avoid creation of temporary proxies. Since this pitfall is everything but obvious, we could also (or instead) consider to patch our Comlink fork to avoid temporary proxy creation.

Internal tracking issue: FG-6370

### DIFF
--- a/packages/studio-base/src/panels/Plot/OffscreenCanvasRenderer.ts
+++ b/packages/studio-base/src/panels/Plot/OffscreenCanvasRenderer.ts
@@ -21,6 +21,9 @@ const registry = new FinalizationRegistry<() => void>((dispose) => {
 export class OffscreenCanvasRenderer {
   #canvas: OffscreenCanvas;
   #remote: Promise<Comlink.RemoteObject<ChartRenderer>>;
+  #remoteUpdate?: Comlink.Remote<ChartRenderer["update"]>;
+  #remoteGetElementsAtPixel?: Comlink.Remote<ChartRenderer["getElementsAtPixel"]>;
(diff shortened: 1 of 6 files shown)


snosenzo (commented):
It's a bit unclear what we lose by not having the proxies. 

Also since it's not immediately evident why we do things this way, would it be possible to create a wrapper function that does this for us safely, or should we update our comlink branch to not create proxies instead?

defunctzombie:
> we could also (or instead) consider to patch our Comlink fork to avoid temporary proxy creation

I would strongly prefer we look into this approach so that every use of comlink does not have to remember to hoist the function all in this somewhat unidiomatic manner.

achim-k:
Superseded by https://github.com/foxglove/comlink/pull/4

PR #7453, opened 2024-02-07
Fix uncaught exception resulting in player error

achim-k:
**User-Facing Changes**
None (fixes a non-released regression)

**Description**
Fixes that an uncaught exception when deserializing messages in `getBackfillMessages` causes the player to go into error state, displaying `An error occurred` in the topic list.

### DIFF
--- a/packages/studio-base/src/players/IterablePlayer/DeserializingIterableSource.ts
+++ b/packages/studio-base/src/players/IterablePlayer/DeserializingIterableSource.ts
@@ -168,16 +168,22 @@ export class DeserializingIterableSource implements IDeserializedIterableSource 
       ]),
     );
 
-    const deserialize = (rawMessages: MessageEvent<Uint8Array>[]) => {
-      return rawMessages.map((rawMsg) => {
+    const rawMessages = await this.#source.getBackfillMessages(args);
+    const deserializedMsgs: MessageEvent[] = [];
+    for (const rawMsg of rawMessages) {
+      try {
         const subscription = subscribePayloadWithHashByTopic.get(rawMsg.topic);
         if (!subscription) {
           throw new Error(`Received message on topic ${rawMsg.topic} which was not subscribed to.`);
         }
-        return this.#deserializeMessage(rawMsg, subscription);
-      });
-    };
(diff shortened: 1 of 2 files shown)


packages/studio-base/src/players/IterablePlayer/DeserializingIterableSource.ts
```diff
@@ -168,16 +168,22 @@ export class DeserializingIterableSource implements IDeserializedIterableSource
...
         if (!subscription) {
           throw new Error(`Received message on topic ${rawMsg.topic} which was not subscribed to.`);
         }
-        return this.#deserializeMessage(rawMsg, subscription);
-      });
-    };
-    return await this.#source.getBackfillMessages(args).then(deserialize);
+        deserializedMsgs.push(this.#deserializeMessage(rawMsg, subscription));
+      } catch (err) {
+        // We simply ignore errors here as there is no way to pass errors/problems to the caller.
```
defunctzombie, line 181: At a minimum we need to log this error - maybe send it to sentry?

Why is there no way to report this to the caller? Something called this and there was an error with our handling of the call. This could manifest as some behavior where a panel does not display a message and we would be confused as to what went wrong.

Typically I am not a fan of try/catch going unsurfaced in some form unless we are _expecting_ a specific error to occur as normal behavior.

achim-k: > At a minimum we need to log this error - maybe send it to sentry?

I will add error logging, this is also how it was before:
https://github.com/foxglove/studio/blob/dd4add4e8fa353ff07ed014ba2bb6fdb2779c2c3/packages/studio-base/src/players/IterablePlayer/Mcap/McapIndexedIterableSource.ts#L251-L253

> Why is there no way to report this to the caller? Something called this and there was an error with our handling of the call. This could manifest as some behavior where a panel does not display a message and we would be confused as to what went wrong.

The function signature expects only `MessageEvents`, so we cannot return a `problem` as it is done in `messageIterator`. We would have to change the return value to properly handle errors. But it's very likely, that these errors are already surfaced to the user through normal message iteration

PR #7455, opened 2024-02-07
Fix rosbag2 (db3) deserialization errors

achim-k:
**User-Facing Changes**
None (regression hasn't been released yet)

**Description**
Fixes a regression introduced in #7420 which changed the way rosbag2 message are deserialized. The gendep-like message definition that we had to generate for `parseChannel` contained ROS1 message definition which can be quite different than their ROS2 counterparts (e.g. std_msgs/Header contains an additional `seq uint32` field). This PR makes sure that only ROS2 message definitions are part of the generated message definition string.


achim-k:
This feels still somewhat brittle, I will check if we can generating a gendeps-like message definition. But I would propose to merge this PR already as some users have already noticed the regression.

defunctzombie (approved):

PR #7456, opened 2024-02-07
unindexed mcap: Fix double transfer of array buffers

achim-k:
**User-Facing Changes**
None (regression hasn't been released yet)

**Description**
Fixes message array buffers being invalidated as they are transferred. The unindexed mcap iterable source stores messages in memory, so we cannot transfer these buffers away. Solution: `structureClone` these message events in the message iterator.

### DIFF
--- a/packages/studio-base/src/players/IterablePlayer/Mcap/McapUnindexedIterableSource.ts
+++ b/packages/studio-base/src/players/IterablePlayer/Mcap/McapUnindexedIterableSource.ts
@@ -284,7 +284,7 @@ export class McapUnindexedIterableSource implements ISerializedIterableSource {
           resultMessages.push({
             type: "message-event" as const,
             connectionId: channelId,
-            msgEvent,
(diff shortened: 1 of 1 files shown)


packages/studio-base/src/players/IterablePlayer/Mcap/McapUnindexedIterableSource.ts
```diff
@@ -284,7 +284,7 @@ export class McapUnindexedIterableSource implements ISerializedIterableSource {
           resultMessages.push({
             type: "message-event" as const,
             connectionId: channelId,
-            msgEvent,
+            msgEvent: structuredClone(msgEvent),
```
defunctzombie, line 287: Please add a comment on why we structure clone here (and the other change in this PR) so we remember why this needs to be here. It is not obvious here because this transferring of the underlying message event happens elsewhere.

defunctzombie (approved):

jtbandes (approved):

PR #7460, opened 2024-02-07
Call terminate on the buffered source to clear its caches when closing the player

snosenzo:
**User-Facing Changes**

- mitigate memory leak when selecting a new source

**Description**

Calls `terminate` on the `BufferedIterableSource` when closing the player, so that it can clear its child caches. This is important to do because the previous player remains in react state after a new source is selected. It is difficult to fix this more directly, so in the meantime its important for the footprint of that previous player to be as small as possible. 

Partially addresses FG-6539

### DIFF
--- a/packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
+++ b/packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
@@ -1106,6 +1106,7 @@ export class IterablePlayer implements Player {
     await this.#blockLoader?.stopLoading();
     await this.#blockLoadingProcess;
     await this.#bufferImpl.stopProducer();
+    await this.#bufferImpl.terminate();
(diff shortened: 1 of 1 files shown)


achim-k (approved):

packages/studio-base/src/players/IterablePlayer/IterablePlayer.ts
```diff
@@ -1106,6 +1106,7 @@ export class IterablePlayer implements Player {
     await this.#blockLoader?.stopLoading();
     await this.#blockLoadingProcess;
     await this.#bufferImpl.stopProducer();
+    await this.#bufferImpl.terminate();
```
defunctzombie, line 1109: Do we need both `stopProducer` and `terminate` calls? Does terminate not imply `stopProducer`?

PR #7487, opened 2024-02-13
Propagate Plot worker errors to React

jtbandes:
**User-Facing Changes**
Not worth calling out

**Description**
Instead of silently failing with an error in the console, propagate worker errors to the React layer so they can crash the panel:

Screenshot 2024-02-13 at 3 26 22 PM


jtbandes:
I tested it by hard-coding a bad url in the `new Worker()` call 😕

PR #7491, opened 2024-02-14
Use deep merge for workspace persistence

jtbandes:
**User-Facing Changes**
None

**Description**
Relates to https://github.com/foxglove/studio/pull/7443. I noticed that sometimes I would get an `undefined` value for speed, which led to this problem:
image

You could reproduce this by editing the local storage value to have `"playbackControls":{"repeat":false}`.

This deep merge should result in using the default 1.0 if the speed was not present in local storage. (if I am using zustand correctly...)


defunctzombie (approved):
comment on why we need a custom merge? Maybe it is obvious to those that use this store that this should have existed?

PR #7494, opened 2024-02-14
Fix message iterator skipping messages at source start time

achim-k:
**User-Facing Changes**
Fix messages at the very start of the file being skipped

**Description**
When seeking back to the start of the file, we were falsely adding 1ns to the playback message iterator resulting in messages at the very start of the file being skipped. This PR changes the behavior to not add the additional 1ns when seeking to the start of the file.


jtbandes (approved):
👍 🔧

